PR #4192, opened 2023-10-03
feat: refactor how warnings are displayed in onboarding

### LINKED ISSUE #4174: Onboarding: show button to fix a failing check if command is available

In the current onboarding implementation we only show the list of checks (both successful and failed) and a description/link next to the failed ones, to guide the user to a solution.

image

In some cases, when we can provide support to fix the problem, we should display a button so that the user can click on it and make Desktop do the job. 

Example of mockup 

image

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR updates a bit how the warnings are displayed during the onboarding. 
This will be useful for follow-up PRs when we add buttons that automatically update/fix the system.

Following the new mockup that Mo made
image

Let's say i miss 3 requirements, we can tell to Desktop that requirement 3 has a command that can fix it and 1,2 only have documentation and it's up to the user to fix them (e.g if you are on an older version of windows or you need more memory).

It can also happen that a warning has both (command + links)

image

it resolves #4174 

### How to test this PR?

1. run tests (when i'll push them), open for discussion for the moment

### DIFF
--- a/extensions/podman/src/base-check.ts
+++ b/extensions/podman/src/base-check.ts
@@ -21,11 +21,28 @@ export abstract class BaseCheck implements extensionApi.InstallCheck {
   abstract title: string;
   abstract execute(): Promise<extensionApi.CheckResult>;
 
-  protected createFailureResult(description?: string, title?: string, url?: string): extensionApi.CheckResult {
+  protected createFailureResult(
+    description?: string,
+    linksDescription?: string,
+    linkTitle?: string,
+    linkUrl?: string,
+    fixFailedCheckCommandId?: string,
+    fixFailedCheckCommandTitle?: string,
+  ): extensionApi.CheckResult {
     const result: extensionApi.CheckResult = { successful: false, description };
-    if (title && url) {
-      result.docLinks = [{ url, title }];
+    if (linkTitle && linkUrl) {
+      result.docLinks = [{ url: linkUrl, title: linkTitle }];
     }
+    if (linksDescription) {
+      result.docLinksDescription = linksDescription;
+    }
+    if (fixFailedCheckCommandId) {
+      result.fixFailedCheckCommand = {
+        id: fixFailedCheckCommandId,
+        title: fixFailedCheckCommandTitle,
+      };
+    }
+
     return result;
   }
 
--- a/extensions/podman/src/base-check.ts
+++ b/extensions/podman/src/base-check.ts
@@ -17,32 +17,31 @@
  ***********************************************************************/
 import type * as extensionApi from '@podman-desktop/api';
 
+export interface FailureObject {
+  description: string;
+  docLinksDescription?: string;
+  docLinks?: extensionApi.CheckResultLink;
+  fixCommand?: extensionApi.CheckResultFixCommand;
+}
+
 export abstract class BaseCheck implements extensionApi.InstallCheck {
   abstract title: string;
   abstract execute(): Promise<extensionApi.CheckResult>;
 
-  protected createFailureResult(
-    description?: string,
-    linksDescription?: string,
-    linkTitle?: string,
-    linkUrl?: string,
-    fixFailedCheckCommandId?: string,
-    fixFailedCheckCommandTitle?: string,
-  ): extensionApi.CheckResult {
-    const result: extensionApi.CheckResult = { successful: false, description };
-    if (linkTitle && linkUrl) {
-      result.docLinks = [{ url: linkUrl, title: linkTitle }];
+  protected createFailureResult(failureObject: FailureObject): extensionApi.CheckResult {
+    const result: extensionApi.CheckResult = { successful: false, description: failureObject.description };
+    if (failureObject.docLinksDescription) {
+      result.docLinksDescription = failureObject.docLinksDescription;
     }
-    if (linksDescription) {
-      result.docLinksDescription = linksDescription;
+    if (failureObject.docLinks) {
+      result.docLinks = [{ url: failureObject.docLinks.url, title: failureObject.docLinks.title }];
     }
-    if (fixFailedCheckCommandId) {
-      result.fixFailedCheckCommand = {
-        id: fixFailedCheckCommandId,
-        title: fixFailedCheckCommandTitle,
+    if (failureObject.fixCommand) {
+      result.fixCommand = {
+        id: failureObject.fixCommand.id,
+        title: failureObject.fixCommand.title,
       };
     }
-
     return result;
   }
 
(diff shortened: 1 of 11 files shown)


extensions/podman/src/base-check.ts
```diff
@@ -21,11 +21,28 @@ export abstract class BaseCheck implements extensionApi.InstallCheck {
...
+    if (linkTitle && linkUrl) {
+      result.docLinks = [{ url: linkUrl, title: linkTitle }];
     }
+    if (linksDescription) {
+      result.docLinksDescription = linksDescription;
+    }
+    if (fixFailedCheckCommandId) {
+      result.fixFailedCheckCommand = {
+        id: fixFailedCheckCommandId,
+        title: fixFailedCheckCommandTitle,
```
jeffmaury: What if fixFailedCheckCommandTitle is undefined ?

packages/renderer/src/lib/markdown/micromark-warnings-directive.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (item.command) {
+      // Make this a button
+      this.tag(
+        '<button class="px-4 py-[6px] rounded-[4px] text-white text-[13px] whitespace-nowrap bg-purple-600 hover:bg-purple-500 no-underline" data-command="' +
+          this.encode(item.command.id) +
+          '">',
+      );
+
+      // Add the "progress" spinner
+      // TODO: Remove cloudfoundry to tailwind
```
benoitf: cloudfoundry ?

benoitf: I guess we can't introduce patternfly as we're trying to get rid of it

packages/extension-api/src/extension-api.d.ts
```diff
@@ -277,10 +277,17 @@ declare module '@podman-desktop/api' {
...
+    id: string;
+    title: string;
+  }
+
   export interface CheckResult {
     successful: boolean;
     description?: string;
+    docLinksDescription?: string;
     docLinks?: CheckResultLink[];
+    fixFailedCheckCommand?: CheckResultFixCommand;
```
benoitf: IMHO as we are in a CheckResult object we don't need to add failed or check inside the attribute
failed as it depends on the successful response and check as we're inside a CheckResult

so maybe shorter like `fixCommand?`

extensions/podman/src/base-check.ts
```diff
@@ -21,11 +21,28 @@ export abstract class BaseCheck implements extensionApi.InstallCheck {
...
 
-  protected createFailureResult(description?: string, title?: string, url?: string): extensionApi.CheckResult {
+  protected createFailureResult(
+    description?: string,
+    linksDescription?: string,
+    linkTitle?: string,
+    linkUrl?: string,
+    fixFailedCheckCommandId?: string,
+    fixFailedCheckCommandTitle?: string,
+  ): extensionApi.CheckResult {
```
benoitf: I think it's starting to become a nightmare with all these parameters

so probably need to switch to a JSON like object/interface

it'll be also more readable from client/consumer

instead of

```ts
this.createFailureResult(
          'To be able to run WSL2 you need Windows 10 Build 18362 or later.',
          'Learn about WSL requirements:',
          'WSL2 Install Manual',
          'https://docs.microsoft.com/en-us/windows/wsl/install-manual#step-2---check-requirements-for-running-wsl-2',
        );
```

if it's more like

```ts
this.createFailureResult({
description:  'To be able to run WSL2 you need Windows 10 Build 18362 or later.',
descriptionLink: 'Learn about WSL requirements:',
linkTitle: 'WSL2 Install Manual',
linkUrl: 'https://docs.microsoft.com/en-us/windows/wsl/install-manual#step-2---check-requirements-for-running-wsl-2'
});
```

extensions/podman/src/podman-install.ts
```diff
@@ -410,18 +414,24 @@ class WinVersionCheck extends BaseCheck {
...
-          'To be able to run WSL2 you need Windows 10 Build 18362 or later.',
-          'WSL2 Install Manual',
-          'https://docs.microsoft.com/en-us/windows/wsl/install-manual#step-2---check-requirements-for-running-wsl-2',
-        );
+        return this.createFailureResult({
+          description: 'To be able to run WSL2 you need Windows 10 Build 18362 or later.',
+          docLinksDescription: 'Learn about WSL requirements:',
+          docLinks: {
+            url: 'https://docs.microsoft.com/en-us/windows/wsl/install-manual#step-2---check-requirements-for-running-wsl-2',
+            title: 'WSL2 Install Manual',
```
benoitf: I know it's not related to this PR, just wondering if `WSL2 Install Manual` is correct or it should be `WSL2 Manual Installation Steps`

benoitf (approved):

extensions/podman/src/podman-install.ts
lstocchi: Renamed. I don't have a strong opinion on this. Both works fine on my end but if you have a preference for the second, it's ok for me.

mairin:
@lstocchi this is looking great! I wonder what is the relationship between the "Install WSL2" button and the command to the right of it... is it that you can press the button OR run that command to install WSL2? Or is it that you have to press the button, then run that command?

If it's the first one (you can do either, do not have to do both) I would maybe put the command under the button left aligned to it, maybe with a label (Alternative:) or something like that? So it's clear why the two things are there?

lstocchi:
> @lstocchi this is looking great! I wonder what is the relationship between the "Install WSL2" button and the command to the right of it... is it that you can press the button OR run that command to install WSL2? Or is it that you have to press the button, then run that command?
> 
> If it's the first one (you can do either, do not have to do both) I would maybe put the command under the button left aligned to it, maybe with a label (Alternative:) or something like that? So it's clear why the two things are there?

The button 'Install WSL2' starts the actual execution of the action. In this case we install/enable WSL2 in the user machine, the user just needs to click it. 
The link on the right redirects to the official documentation. So if you want to install it manually, you can go there, read it, open a terminal and do the job yourself.

So it's like ... if you imagine that each row is splitted in 3 parts.
Left -> status icon + description
Center -> action button, if available
Right -> any documentation + links, if available

Maybe i can add `Alternative` on the top of the third column if we have command + documentation/link, and I omit if there is no command? Or something else? WDYT?

cdrage (approved):
LGTM!

Was able to test it as part of the compose onboarding sequence by using the example JSON you provided:

```json
    const warnings = [
      {
        state: 'failed',
        description: 'description',
      },
    ];
```

Was able to appear correctly, tests pass locally / no errors.

cdrage (approved):

Issue #4120, opened 2023-09-27
Add a CLI provider extension point


benoitf:
### Is your feature request related to a problem? Please describe

We should be able to list all CLIs installed by 3rd party extensions and have an unified way of updating it, remove it, setup it, etc

### Describe the solution you'd like

Add a Command Line provider

These CLIs will be displayed in a CLI Tools section in settings

https://github.com/containers/podman-desktop/issues/3781

For example, it should include compose, kubectl, oc, etc

### Additional context

This feature starts with presentation only for registered tools and continue with implementation of 

* https://github.com/containers/podman-desktop/issues/4648
* https://github.com/containers/podman-desktop/issues/4650
* https://github.com/containers/podman-desktop/issues/4651

Related issues

* https://github.com/containers/podman-desktop/issues/3781
* https://github.com/containers/podman-desktop/issues/4118

benoitf:
it differs from https://github.com/containers/podman-desktop/issues/4059 as here the CLI can be embedded in the extension as well

BinaryRegistry is an helper, not the way to register CLIs

dgolovin:
@benoitf podman can be installed using brew on macOS, chocolate on Windows, dnf on Fedora aptitude on Debian and etc. Should CLI provider extension point cover different ways to install tools on different platforms or we just provide implementation similar to what docker-compose does?

benoitf:
hi @dgolovin yes I think there are a lot of unknown in the current description

1. First, I think the idea is to bring awareness to the user that some CLI commands exists and are useful. So here the idea is that if `kubectl` is already found in the PATH (and that we have an extension supporting it), it'll display that `kubectl` entry in the CLI page in the UI. From here it'll display its version.

2. Then, comes the awareness of the installation of the tool. For example as you said, it can be useful to know who was responsible to install the tool. So on macOS we have already some commands checking if podman is installed by brew (to not let upgrade path continue). So it is very helpful to see the `source` or `origin` of the tool. So on macOS, to know if podman CLI that is in the PATH comes from the `podman installer` (in `/opt/podman/bin` or from `/opt/homebrew/etc` if it's coming from brew). (For example we might say it's not the recommended way, etc). If the system is Linux, for now we won't support all Linux distribution to check with apt-get, dnf, rpm-ostree, yum, etc) so source can be `N/A` or `unknown`

(The Provided by in top right corner of each tile I suppose)

[...]

PR #4092, opened 2023-09-25
draft: BinaryRegistry and UpdateProviderRegistry

axel7083:
### Why this PR

I started developing an extension aiming to add the [kompose](https://kompose.io/) project into podman-desktop. While doing so, I discovered that a lot of code/logic is being duplicated in each extension implementation (downloading, updating etc.). 

### What does this PR do?

This PR provide helpers/methods to handle binaries files in a way to ease the updating process without relying on each extensions to handle the job of searching for update, updating the file, and keeping it system wide.

This PR is mainly related to https://github.com/containers/podman-desktop/issues/4059

But extend and offer a better way later to catch other errors such as https://github.com/containers/podman-desktop/issues/4079

### Screenshot/screencast of this PR

Here is a minimal example realised with the code available in the first comment. 

> The extension is not responsible of showing the picker, nor managing the installation.

https://github.com/containers/podman-desktop/assets/42176370/7814d98a-4d2f-4028-baaf-93074e7fcc05

Issue #3781, opened 2023-09-05
Feature: Add Command-Line Tools section in Settings


cdrage:
### Is your feature request related to a problem? Please describe

Mockup:

![Screenshot 2023-09-05 at 2 55 48 PM](https://github.com/containers/podman-desktop/assets/6422176/fd4870bd-225b-4e53-a156-921c60a7c863)

### Describe the solution you'd like

Add a Command-Line Tools section to Podman Desktop in order to show and integrate CLI versioning as well as update indications for all CLI tools as part of the provider such as `kind`, `compose` and `podman`.

### Tools to integrate

- [ ] podman
- [ ] kind
- [ ] compose cli
- [ ] lima
- [ ] minikube
- [ ] docker

### Additional context

Issues:

* https://github.com/containers/podman-desktop/issues/3708
* https://github.com/containers/podman-desktop/issues/3707

afbjorklund:
The list seems to be missing the Kubernetes client, `kubectl` ? (but it is in the screenshot)

Using something like `kuberlr` might help with the crazy k8s version update schedule...

afbjorklund:
Note: the name of the binary is `limactl`

The name of the plug-in is Lima, though.

deboer-tim:
Like the Resources page this is primarily an issue to implement the overall design and some initial CLIs. Each extension would to contribute known CLIs, how to detect versions/update, etc. So, even if some are 'missing' here we can add to continue them later (and once this is far enough along we should track via independent/another issue anyway so this issue doesn't stay open forever)

afbjorklund:
These were just the programs used by the current providers

axel7083:
@cdrage to continue the discussion on the implementation, I drafted an example of implementation to back the command-line tool section.

![image](https://github.com/containers/podman-desktop/assets/42176370/edae9cf9-aa5d-4da6-bee0-be4d3ea9fd97)

The cli tools, and other binaries could be stored in a folder, that would be managed by a BinaryRegistry. Having a dedicated folder, maybe split in two (one added to path for system wide tools) and another for local usage.

> It would also allow podman-desktop to handle the cleanup, and prevents issues with manual installation made by each extensions.

Extensions could register BinaryProvider, which could includes some kind of updater, here I defined an abstract class of `UpdaterProvider` and a `GithubUpdater` could be one of the implementation, handling many things that is often coded mutliple times in the extensions (fetching the releases etc.).

[...]

PR #4197, opened 2023-10-04
feat: CLI Tools provider API implementation for tools registration

dgolovin:
WIP implementation to let an extension to declare required CLI tools.

### What does this PR do?

Adds extension point, simplified API implementation, UI implementation and extension API call from docker-compose.

https://github.com/containers/podman-desktop/assets/620330/0840e5d4-62ac-48b2-bae4-e0ec7fde31b3

https://github.com/containers/podman-desktop/assets/620330/4afe850d-b4aa-425a-a1d9-85ccc24ef183

#4120

Dependency of 
* https://github.com/containers/podman-desktop/pull/4387
* https://github.com/containers/podman-desktop/pull/4388


packages/extension-api/src/extension-api.d.ts
```diff
@@ -2295,4 +2295,16 @@ declare module '@podman-desktop/api' {
...
+  export interface CliToolOptions {
+    id: string;
+    name: string;
+    displayName: string;
+    description: string;
+    images: ProviderImages;
+  }
+
+  export namespace cli {
+    export function createCliTool(options: CliToolOptions): Disposable;
```
benoitf: I think we should go with the `registerCliToolProvider` approach and providing a `CliTool`
then we can return a Disposable

if we have a `create***()` method, an object should be returned and not a Disposable (return CliTool object)

but it looks more to be a `provider` thing to me

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2295,4 +2295,16 @@ declare module '@podman-desktop/api' {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     export function setValue(key: string, value: any, scope?: 'onboarding'): void;
   }
+
+  export interface CliToolOptions {
+    id: string;
+    name: string;
+    displayName: string;
+    description: string;
+    images: ProviderImages;
```
benoitf, line 2437: this type is not defined in the .d.ts file

dgolovin:
@benoitf This PR is seems to be pursuing the same goal as https://github.com/containers/podman-desktop/pull/4092.

Some entities names and UI implementation are different, but goals are the same. 
Should we agree on how we name the first class objects and just merge the efforts.

We can basically abandon CliToolsRegistry and use Binary Registry form @4092. I would have to redo UI to use BinaryInfo classes and merge it in #4092. 

We would need to implement different plugins to support download, github download, brew and choc update/install but it should fit nicely in that nice API and implementation developed by @axel7083.

@benoitf WDYT?

benoitf:
yes there is a dependency. But I think we need a clear separation. One is about to fetch, list, get notifications from 3rd party repositories

Here it's to provide CLI information, lifecycle, updates.

packages/extension-api/src/extension-api.d.ts
dgolovin: It is here https://github.com/containers/podman-desktop/blob/312dcd2bf988f001c52c6ae2ff73a0cf1c1deeca/packages/extension-api/src/extension-api.d.ts#L316

dgolovin:
@benoitf that is what I have for now. Ideally extensions should declare tools in package.json without any coding.
image

packages/extension-api/src/extension-api.d.ts
dgolovin: I am going to stick to create method returning CliTool, the same pattern used for Providers.

benoitf: @dgolovin providers use a registration mechanism, not a create method

benoitf: for example registerProvider vs createStatusbarItem

when we provide something, it's a provider (then a register thing), when we ask the engine to create object for us like a StatusBar, a Messagebox, etc it's a create

but here it's a registerCliToolProvider where you provide the (cliToolProvider object)

benoitf: I'm a bit lost on 
> I am going to stick to create method returning CliTool, the same pattern used for Providers.

but on https://github.com/containers/podman-desktop/pull/4197#issuecomment-1752429026
image

it's using register approach so I would say you'll stick to a register method returning a Disposable object as shown in the image ?

dgolovin: I meant to use approach similar to provider module here:
https://github.com/containers/podman-desktop/blob/efbe62aef78467ac79644839d69d7a3a9a89c327/packages/extension-api/src/extension-api.d.ts#L426

Right now stripped diagram looks like this:


Currently implemented on UI side:
1. Simple UI
2. UI update on 'extension-loaded' event
3. UI update on 'cli-tool-create' or 'cli-tool-remove'

On main side:
1. Simple registry above
2. Notification on create and dispose of CliTool
3. Generating of the Info objects for UI
4. Cli Tool declaration in package.json through extension

You can ran it from branch and you will see docker-compose in CLI Tools preferences
If you install/uninstall `ghcr.io/dgolovin/podman-desktop-kubectl-ext:latest` you will see kubectl appears and disappears in UI. Compose is using cli module to define cli tool form activation method. kubctl does it in package.json extension manifest.

https://github.com/containers/podman-desktop/assets/620330/0840e5d4-62ac-48b2-bae4-e0ec7fde31b3

packages/renderer/src/stores/cli-tools.ts
```diff
@@ -0,0 +1,52 @@
...
+    return (extensionsStarted = true);
+  }
+
+  return (extensionsStarted && eventName === 'cli-tool-create') || eventName === 'cli-tool-remove';
+}
+
+export const cliToolInfos: Writable<CliToolInfo[]> = writable([]);
+
+const eventStore = new EventStore<CliToolInfo[]>(
+  'providers',
```
benoitf: I think it's a copy/paste, here we should have like 'cliTools' (it's a display thing but better to use cliTool wording)

packages/renderer/src/stores/cli-tools.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const windowEvents: string[] = ['extensions-started', 'cli-tool-create', 'cli-tool-remove'];
+const windowListeners = ['system-ready'];
+
+let extensionsStarted = false;
+
+export async function checkForUpdate(eventName: string): Promise<boolean> {
+  if (eventName === 'extensions-started') {
+    return (extensionsStarted = true);
+  }
```
benoitf:
```
 if ('extensions-already-started' === eventName) {
```

packages/renderer/src/stores/cli-tools.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { writable, type Writable } from 'svelte/store';
+import { EventStore } from './event-store';
+import type { CliToolInfo } from '../../../main/src/plugin/api/cli-tool-info';
+
+const windowEvents: string[] = ['extensions-started', 'cli-tool-create', 'cli-tool-remove'];
+const windowListeners = ['system-ready'];
```
benoitf:
```suggestion
const windowListeners = ['system-ready', 'extensions-already-started'];
```

I think it should track also when an extension is stopped or started

```ts
windowEvents = ['extension-stopped', 'extensions-started'];
```

extensions/compose/compose.png
benoitf, line 1: the file should not be part of the implementation of the provider

but also AFAIK it's not the expected icon for compose (compose is an octopus)

packages/main/src/plugin/extension-loader.ts
```diff
@@ -548,6 +550,20 @@ export class ExtensionLoader {
...
+      }
+      const disposable = this.cliToolRegistry.createCliTool(
+        extension.id,
+        extension.name,
+        extension.manifest.displayName,
+        extensionCliTools,
+      );
+      extension.subscriptions.push(disposable);
+    }
+
```
benoitf: maybe we need to defer this part as it might evolve as I'm not sure about the usage of the extension name and the displayName of the extension.

I mean, probably you want to register multiple CLIS like multiple commands so it might be multiple createCliTool

packages/main/src/plugin/cli-tool-registry.ts
```diff
@@ -0,0 +1,115 @@
...
+  ) {}
+
+  private sequence = new IdSequence();
+
+  private cliTools = new Map<number, CliToolContext>();
+
+  createCliTool(
+    extensionId: string,
+    extensionName: string,
+    extensionDisplayName: string,
```
benoitf: I'm wondering why we need to store the extension name and extension display name as you can find details about an extension from its extension Id

packages/main/src/plugin/cli-tool-registry.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  getCliToolInfos(): CliToolInfo[] {
+    return Array.from(this.cliTools.values()).map(context => ({
+      id: context.cliTool.internalId,
+      name: context.cliTool.name,
+      displayName: context.cliTool.displayName,
+      description: context.cliTool.markdownDescription,
+      state: 'unknown',
+      images: context.cliTool.images,
+      providedBy: context.extensionDisplayName,
```
benoitf: it is confusing that you get as id an internal thing

so it means if you register 'docker-compose' from an extension you don't have a way to find it back ?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2424,4 +2424,19 @@ declare module '@podman-desktop/api' {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     export function setValue(key: string, value: any, scope?: 'onboarding'): void;
   }
+
```
benoitf, line 2428: probably missing a list or get methods to list all cli tools registered ?

axel7083:
@benoitf is there still room for #4092 or should I close it ?

benoitf:
@axel7083 there is room for it.

It's not targeting the same goal. 4092 should be about having ways to be notified for updates, etc from repositories (including Github repository)

so extensions could use 4092 when wanting to download data.

axel7083:
> @axel7083 there is room for it.
> 
> It's not targeting the same goal. 4092 should be about having ways to be notified for updates, etc from repositories (including Github repository)
> 
> so extensions could use 4092 when wanting to download data.

Okey, I will wait for this PR to be merged to adapt #4092 based on the new requirements. Keep me updated !

packages/renderer/src/stores/cli-tools.ts
dgolovin: in current implementation there are specific events for creating and removing cli tools from store. When extension started/stopped the store items are already populated/disposed.

packages/main/src/plugin/extension-loader.ts
dgolovin: Good catch on registering multiple CLIs. I just copy/pasted the code to test the idea and missed the point there could be many CLIs.

extensions/compose/compose.png
dgolovin: where should the icon come from then? Should extension icon be used?

packages/main/src/plugin/cli-tool-registry.ts
dgolovin: it just to pass less data from main to UI, but we sure can get rid of it.

packages/main/src/plugin/extension-loader.ts
dgolovin: I'll remove this code for now.

extensions/compose/compose.png
benoitf: @mairin do we have a dedicated icon for compose cli?

dgolovin: compose.png removed, I used icon.png instead

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2424,4 +2424,47 @@ declare module '@podman-desktop/api' {
...
+    name: string;
+    displayName: string;
+    markdownDescription: string;
+    images: ProviderImages;
+    location?: string;
+    version?: string;
+    detection?: {
+      versionOptions: string[]; // run command by name using these options
+      versionParser: (string) => string;
+    };
```
benoitf: I would remove this part for now as I'm not sure if it's the way to move forward or why we need this.

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2424,4 +2424,47 @@ declare module '@podman-desktop/api' {
...
+    | 'uninstalling'
+    | 'unknown';
+
+  export interface CliTool extends CliToolOptions, Disposable {
+    extensionInfo: {
+      id: string;
+      label: string;
+    };
+
+    detect(): Promise<void>;
```
benoitf: remove for now

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2424,4 +2424,47 @@ declare module '@podman-desktop/api' {
...
+    state: CliToolState;
+  }
+
+  export namespace cli {
+    export function createCliTool(options: CliToolOptions): CliTool;
+    export function listCliTools(): CliTool[];
```
benoitf: should it return CliTool objects or some 'info' view ? (like a view, not the real object)

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2424,4 +2424,47 @@ declare module '@podman-desktop/api' {
...
+    | 'installed-unknown'
+    | 'uninstall-pending'
+    | 'uninstalling'
+    | 'unknown';
+
+  export interface CliTool extends CliToolOptions, Disposable {
+    extensionInfo: {
+      id: string;
+      label: string;
+    };
```
benoitf, line 2452: why it contains a label + an id there ? and not only the extensionId ?

packages/main/src/plugin/api/cli-tool-info.ts
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { CliToolState, ProviderImages } from '@podman-desktop/api';
+
+export type BinaryProviderName = 'download' | 'brew' | 'chocolate' | string;
```
benoitf: I'm not sure we can have 'brew' or 'chocolate' here (BTW I think you're speaking about chocolatey ?)

packages/main/src/plugin/api/cli-tool-info.ts
```diff
@@ -0,0 +1,41 @@
...
+  state: CliToolState;
+  images?: ProviderImages;
+  providedBy: string;
+  binary?: CliToolBinaryInfo;
+}
+
+export interface CliToolBinaryInfo {
+  location?: string;
+  version: string;
+  binaryManagerName?: BinaryProviderName;
```
benoitf: remove for now probably

packages/main/src/plugin/api/cli-tool-info.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export interface CliToolInfo {
+  id: string;
+  name: string;
+  description: string;
+  displayName: string;
+  state: CliToolState;
+  images?: ProviderImages;
+  providedBy: string;
+  binary?: CliToolBinaryInfo;
```
benoitf: it's missing extensionId ?

packages/main/src/plugin/api/cli-tool-info.ts
```diff
@@ -0,0 +1,41 @@
...
+export type ButtonName = 'detect' | 'setup';
+
+export interface CliToolInfo {
+  id: string;
+  name: string;
+  description: string;
+  displayName: string;
+  state: CliToolState;
+  images?: ProviderImages;
+  providedBy: string;
```
benoitf: is that providedBy is the extensionId ?

packages/main/src/plugin/api/cli-tool-info.ts
benoitf: why do we need a separate field for binary info ?

for example why the version or location is not already in the upper info object ?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2424,4 +2424,47 @@ declare module '@podman-desktop/api' {
...
+    | 'installing'
+    | 'update-pending'
+    | 'updating'
+    | 'installed'
+    | 'installed-unknown'
+    | 'uninstall-pending'
+    | 'uninstalling'
+    | 'unknown';
+
+  export interface CliTool extends CliToolOptions, Disposable {
```
benoitf: why CLiTool is extending CliToolOptions ?

benoitf:
let's try to have only the simple state being merged. (no detection, etc)

packages/extension-api/src/extension-api.d.ts
dgolovin: I am not sure ether, but the idea is to be able to configure detection without coding, so you can just add it to package.json or call api directly. We can have multiple predefined version parsers that can be referenced by id. Usually version is a sub-command for a cli tool, it also might have options to control output. It could be `--short` that just prints version in a single line (docker-compose) or `-o json` to print json version (kubectl and oc). 

We can create passthrough and json version parser and confiure detection this way in package.json.

for docker-compose

```
detection?: {
    versionOptions: 'version', '--short',
    versionParser: 'passthrough'
}
```

for kubectl 


```
detection?: {
    versionOptions: 'version', '-o', 'json',
    versionParser: 'json',
    versionParserConfig: {
       path: 'clientVersion.gitVersion'
    }
}
```

and so on.

packages/extension-api/src/extension-api.d.ts
dgolovin: Should it return proxy, that would let call lifecycle methods like install (if we decide to put them in CliTool), but block any other modification for non-owner extensions?

Another option is just expose current internal method that returns Infos used in UI. Is that what you getting at?

packages/extension-api/src/extension-api.d.ts
dgolovin: It is a cache to avoid ipc call to main process to get the name

packages/main/src/plugin/api/cli-tool-info.ts
dgolovin: Extension name to avoid call form UI to main thread.

packages/main/src/plugin/api/cli-tool-info.ts
dgolovin: It could be that way, I was overthinking, that we might have multiple versions available form PATH or from binary storage for downloaded files.

packages/extension-api/src/extension-api.d.ts
dgolovin: because the same properties should be on CliTool anyway, I can copy/paste it if that is better.

packages/extension-api/src/extension-api.d.ts
benoitf: for now it's ok as we have mostly 'properties' but yes I would expect that an extension wouldn't be able to call methods of another CLI registered elsewhere

so don't know if proxy works or just different objects that provides only 'properties' (like is there a lifecycle: yes/no or what are the lifecycle methods but there is no direct call possible on these objects)


if it's easier for now, we can remove the list method from the API

packages/main/src/plugin/api/cli-tool-info.ts
benoitf: ok so probably it should be extensionName or extensionInfo ?

in the API we have
```
 extensionInfo: {
      id: string;
      label: string;
    };
```

and here it looks like we have a different structure. I would probably unify (like switch here to extensionInfo as well then ?)

packages/main/src/plugin/api/cli-tool-info.ts
```diff
@@ -0,0 +1,33 @@
...
+  location?: string;
+  version?: string;
+  systemWide?: boolean;
+  installedOn?: Date;
```
benoitf: I don't think we have that information easily so I would postpone it for the first iteration

packages/main/src/plugin/cli-tool-registry.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { test } from 'vitest';
+
+test('CLI Tool registry sends "cli-tool-create" event when new tool is created', () => {
+  // TBD
```
benoitf: ok tests are missing

packages/renderer/src/lib/preferences/PreferencesCliToolsRendering.svelte
```diff
@@ -0,0 +1,44 @@
...
+import EngineIcon from '../ui/EngineIcon.svelte';
+import EmptyScreen from '../ui/EmptyScreen.svelte';
+</script>
+
+<SettingsPage title="CLI Tools">
+  <div class="h-full">
+    <EmptyScreen
+      aria-label="no-resource-panel"
+      icon="{EngineIcon}"
+      title="No CLI Tool description found"
```
benoitf: should we add that it is not registered ?

```suggestion
      title="No CLI tool has been registered"
```

packages/renderer/src/lib/preferences/PreferencesCliToolsRendering.svelte
```diff
@@ -0,0 +1,44 @@
...
+      message="Start an extension that add CLI Tool description"
```
benoitf:
```suggestion
      message="Start an extension that registers a CLI"
```

packages/renderer/src/lib/preferences/PreferencesCliToolsRendering.svelte
```diff
@@ -0,0 +1,44 @@
...
+                {/if}
+              {/if}
+              <span class="my-auto text-gray-400 ml-3 break-words">{cliTool.name}</span>
+            </div>
+          </div>
+        </div>
+        <!-- cli-tools columns -->
+        <div class="grow flex-column divide-gray-900 ml-2">
+          <span class="my-auto text-gray-400 ml-3 break-words">{cliTool.displayName}</span><span
+            class="float-right text-sm">Provided by {cliTool.providedBy}</span>
```
benoitf:
```suggestion
            class="float-right text-sm">Registered by {cliTool.providedBy}</span>
```

I think provided is misleading as for example I can have kubectl binary being installed by myself but if an extension find it installed globally it would say that it's provided by this extension while it's more "registered" (but not for example managed or provided)

benoitf (commented):
overall, LGTM, I added few minor remarks
Of course unit tests need to be added

packages/extension-api/src/extension-api.d.ts
benoitf: agreed we can just proceed with extensionInfo

packages/extension-api/src/extension-api.d.ts
dgolovin: Let's remove it.

packages/main/src/plugin/api/cli-tool-info.ts
dgolovin: I'll remove them.

packages/main/src/plugin/api/cli-tool-info.ts
dgolovin: Let's unify.

packages/main/src/plugin/api/cli-tool-info.ts
dgolovin: Should I remove location and version form extension-api.ts as well?

dgolovin: State is also seems to be not relevant in this iteration. Should it be removed from first iteration?

dgolovin: I removed not easily available info and reduced states to the only one we have 'registered'.

extensions/compose/src/compose-extension.ts
```diff
@@ -135,6 +135,21 @@ export class ComposeExtension {
       this.statusBarItem.show();
       this.extensionContext.subscriptions.push(this.statusBarItem);
     }
+    const description = `Compose is a tool for defining and running multi-container Docker applications. 
+    With Compose, you use a YAML file to configure your application's services. Then, with a single command, 
+    you create and start all the services from your configuration.
+    Compose works in all environments; production, staging, development, testing, as well as CI workflows. 
+    It also provides commands for managing the whole lifecycle of your application.`;
+
+    extensionApi.cli.createCliTool({
```
cdrage: Should we add a note like.. "TODO: Remove before merging?" so we don't merge this in by mistake?

extensions/compose/src/compose-extension.ts
```diff
@@ -135,6 +135,21 @@ export class ComposeExtension {
       this.statusBarItem.show();
       this.extensionContext.subscriptions.push(this.statusBarItem);
     }
+    const description = `Compose is a tool for defining and running multi-container Docker applications.
```
cdrage: `compose-extension.ts` will be removed in the near future after we 100% get compose onboarding in, the reason being is that `compose-extension.ts` is mainly for the compose status bar on the bottom of Podman Desktop which will be removed in favourite for notifications.

cdrage: can you move it to `extension.ts`?

you can also re-use the same description for `extension.ts`

cdrage: you'll be able to use the same description once https://github.com/containers/podman-desktop/pull/4179 is added since it'll add the compose "provider" to the list.

packages/renderer/src/lib/preferences/PreferencesCliToolsRendering.svelte
```diff
@@ -0,0 +1,44 @@
...
+  <div class="h-full">
+    <EmptyScreen
+      aria-label="no-resource-panel"
+      icon="{EngineIcon}"
+      title="No CLI tool has been registered"
+      message="Start an extension that registers a CLI"
+      hidden="{$cliToolInfos.length > 0}" />
+
+    {#each $cliToolInfos as cliTool}
+      <div class="bg-charcoal-600 mb-5 rounded-md p-3 divide-x divide-gray-900 flex">
```
cdrage: I feel like the below section can be put into a different modal?

it'll also be easier when writing the tests for the .svelte

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2425,4 +2425,31 @@ declare module '@podman-desktop/api' {
...
+    id: string;
+    name: string;
+    displayName: string;
+    markdownDescription: string;
+    images: ProviderImages;
+  }
+
+  export type CliToolState = 'registered';
+
+  export interface CliTool extends Disposable {
```
cdrage, line 2442: Are you able to add comments similar to other parts of extension-api.d.ts? This will help others when implementing their own extension on how to use the CLI functionality and this functionality is regarding creating update / download / management sections for podman desktop.

cdrage (changes requested):
The implementation looks great and testing it against Compose works! Getting the initial functionality is great.

@benoitf should we hide this behind an experimental flag so we don't show the "Cli Tools" section which wouldn't have full functionality yet?

dgolovin:
@benoitf @cdrage we can move UI part into different PR for #3781, because this PR should be about API only. WDYT?

packages/extension-api/src/extension-api.d.ts
dgolovin: done.

benoitf:
I'm ok @dgolovin to split the pr

packages/main/src/plugin/cli-tool-registry.spec.ts
dgolovin: added tests for cli API

packages/renderer/src/lib/preferences/PreferencesCliToolsRendering.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { test } from 'vitest';
+
+test('CLI Tool info presented correctly in UI', () => {
+  // TBD
+});
```
benoitf: probably should not belong to this PR ?

benoitf (commented):
AFAIK some files are moved to https://github.com/containers/podman-desktop/pull/4387

so here we should remove renderer/ part ?

packages/renderer/src/lib/preferences/PreferencesPage.svelte
```diff
@@ -54,7 +54,7 @@ onMount(async () => {
   <Route path="/extension/:extensionId/*" breadcrumb="Extensions" let:meta>
     <PreferencesExtensionRendering extensionId="{meta.params.extensionId}" />
   </Route>
-  <Route path="/provider/:providerInternalId/*" breadcrumb="Resources" let:meta navigationHint="details">
+  <Route path="/provider/:providerInternalId/*" breadcrumb="Resources" let:meta>
```
benoitf: I suppose we need to remove as well this change ?

benoitf (approved):
LGTM (removing the renderer change)

extensions/compose/src/compose-extension.ts
dgolovin: moved to separate PR https://github.com/containers/podman-desktop/pull/4388

extensions/compose/src/compose-extension.ts
dgolovin: Moved to PR https://github.com/containers/podman-desktop/pull/4388

packages/renderer/src/lib/preferences/PreferencesCliToolsRendering.spec.ts
dgolovin: removed

cdrage (approved):
LGTM. This is awesome 🚀  thanks for addressing the comments.

PR #4209, opened 2023-10-04
feat: add action to install wsl if missing during onboarding (#4193)

lstocchi:
### What does this PR do?

This PR adds the action/button to install wsl in modern Windows OS and show it during the onboarding if wsl2 is not installed.

The button has like two modes. The first is the normal one (purple with the spinner on the left when executing) and the "error" mode if the command fails its execution.
As there is no much space in the onboarding view if the command fail somehow, the button gets red and an expandable section is displayed that allow to see the error logs.

action_onboarding_error

it is part of #4193 

### How to test this PR?

1. uninstall wsl from your machine and run the onboarding
2. install it by using the new button

### DIFF
--- a/tailwind.config.cjs
+++ b/tailwind.config.cjs
@@ -21,7 +21,7 @@ const tailwindColors = require('tailwindcss/colors')
 module.exports = {
   content: [
     'packages/renderer/index.html',
-    'packages/renderer/src/**/*.{svelte,ts,css}',    
+    'packages/renderer/src/**/*.{svelte,ts,css}',
   ],
   theme: {
     extend: {
--- a/tailwind.config.cjs
+++ b/tailwind.config.cjs
@@ -21,7 +21,7 @@ const tailwindColors = require('tailwindcss/colors')
 module.exports = {
   content: [
     'packages/renderer/index.html',
-    'packages/renderer/src/**/*.{svelte,css}',
+    'packages/renderer/src/**/*.{svelte,ts,css}',    
   ],
   theme: {
     extend: {
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -876,9 +876,14 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
   // provide an installation path ?
   if (podmanInstall.isAbleToInstall()) {
+    // init all install checks
+    const installChecks = podmanInstall.getInstallChecks();
+    for (const check of installChecks) {
+      await check.init?.();
+    }
     provider.registerInstallation({
       install: () => podmanInstall.doInstallPodman(provider),
-      preflightChecks: () => podmanInstall.getInstallChecks(),
+      preflightChecks: () => installChecks,
     });
   }
 
--- a/extensions/podman/src/podman-install.spec.ts
+++ b/extensions/podman/src/podman-install.spec.ts
@@ -38,6 +38,9 @@ vi.mock('node:os', async () => {
 
 vi.mock('@podman-desktop/api', async () => {
   return {
+    commands: {
+      registerCommand: vi.fn(),
+    },
     window: {
       withProgress: vi.fn(),
       showNotification: vi.fn(),
@@ -430,4 +433,13 @@ test('expect winWSL2 preflight check return failure result if it fails when chec
   expect(result.description).equal('Could not detect WSL2');
   expect(result.docLinks[0].url).equal('https://learn.microsoft.com/en-us/windows/wsl/install');
   expect(result.docLinks[0].title).equal('WSL2 Manual Installation Steps');
+});
+
+test('expect winWSL2 init to rgister WSLInstall command', async () => {
+  const registerCommandMock = vi.spyOn(extensionApi.commands, 'registerCommand');
+  const installer = new WinInstaller();
+  const preflights = installer.getPreflightChecks();
+  const winWSLCheck = preflights[4];
+  await winWSLCheck.init();
+  expect(registerCommandMock).toBeCalledWith('podman.onboarding.installWSL', expect.any(Function));
 });
--- a/extensions/podman/src/podman-install.ts
+++ b/extensions/podman/src/podman-install.ts
@@ -480,6 +480,13 @@ class VirtualMachinePlatformCheck extends BaseCheck {
 
 class WSL2Check extends BaseCheck {
   title = 'WSL2 Installed';
+  installWSLCommandId = 'podman.onboarding.installWSL';
+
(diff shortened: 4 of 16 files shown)


lstocchi:
This PR is not ready yet but i opened it to have some suggestion how to improve the experience.
My 2 problems are... where do i show the error raised during the execution (if any) and what i can show to the users to let them know what's going on in a better way.

This is the video demo (when it gets black it's because i got displayed some system dialog for permissions).
https://drive.google.com/file/d/10IM1XGDwC2lGPWoPNiqsI5oDLgr5BOWL/view

1. Basically if there is an error, i don't know where to show it. Do i have to show a new red icon and when hovering the user see the error? or by clicking it opens up a dialog? or i just show a warning dialog? Something else?

2. If you look at the video you can see that for 30 sec i just show the spinner looping. Do we want to add some more info to explain what's going on?

@mairin questions for you 😆 and whoever wants to comment

mairin:
> 1. Basically if there is an error, i don't know where to show it. Do i have to show a new red icon and when hovering the user see the error? or by clicking it opens up a dialog? or i just show a warning dialog? Something else?

See below 😁

> 2. If you look at the video you can see that for 30 sec i just show the spinner looping. Do we want to add some more info to explain what's going on?

Yeh! I would also just have the one spinner... excuse the awkward screenshot scribbling here but:

Screenshot_20231004_205955_Gallery

You have a spinner at (1) and at (2) and they mark the same thing, right? I'd drop the one at (1) and just use (2) for the spinner. I would also gray out (4) until the spinner is stopped. 

Now the bit where I marked (3) - I know we have that discussion ongoing in another PR. That bit, I know sometimes you have it, sometimes you dont. It feels like it should go under the button, not to the right. Maybe under a disclosure triangle (patternfly calls them expandable sections 🤷‍♀️ https://www.patternfly.org/components/expandable-section/html/#disclosure-variation-hidden ). We can refine that in the other PR maybe. But! If you move that below, then you'll have space to the right of the button for more detailed status, and to throw errors.

Do you get more detailed status from the WSL creation process? You know, "Reticulating splines..." or whatnot? Do you get error messages if it fails? Or is it just a raw log? What does the structure look like?

I would also say to ideally treat WSL install as a task since it's a longer running process, so maybe if there's an error and the user navigated away from this screen, flag it in the task manager?

mairin:
> Now the bit where I marked (3) - I know we have that discussion ongoing in another PR. That bit, I know sometimes you have it, sometimes you dont. It feels like it should go under the button, not to the right. Maybe under a disclosure triangle (patternfly calls them expandable sections 🤷‍♀️ https://www.patternfly.org/components/expandable-section/html/#disclosure-variation-hidden ). We can refine that in the other PR maybe. 

Ah found the other one, it's merged already:

https://github.com/containers/podman-desktop/pull/4192#issuecomment-1746715703

I'll do a quick mockup

mairin:
OK here is one idea:

Screenshot from 2023-10-04 21-58-28

I put the extra info / command line in a disclosure triangle under the button. That goes away if you fail - maybe that's weird? - and turns into an error log display.

lstocchi:
> You have a spinner at (1) and at (2) and they mark the same thing, right? I'd drop the one at (1) and just use (2) for the spinner. I > would also gray out (4) until the spinner is stopped.

👍  

> Do you get more detailed status from the WSL creation process? You know, "Reticulating splines..." or whatnot? Do you get 
> error messages if it fails? Or is it just a raw log? What does the structure look like?

If i recall correctly we receive logs about the status of the download (1,5,8,20 ... 99%)... then the completion message.
If i stop the execution because i don't give the right permission i receive an error message. I still haven't faced any other situation so idk what kind of errors i can receive.

But keep in mind that we're creating a ui for every onboarding, so maybe WSL install action has some structured error message and some other command don't. So we should have something that can potentially work with a variety of outputs.

>I would also say to ideally treat WSL install as a task since it's a longer running process, so maybe if there's an error and the 
> user navigated away from this screen, flag it in the task manager?

Yes, i forgot to mention but I already planned to do it 👍 

> I put the extra info / command line in a disclosure triangle under the button. That goes away if you fail - maybe that's weird? - and turns into an error log display.

Oooh that's nice. I'm just afraid of the length of the error logs. Maybe we should open a dialog with the logs view?

I'll start working on your suggestion so we can see how it looks 👍 
Thank you so much!!

lstocchi:
@mairin i still have to fix some stuff but this could give you an idea of how it is... WDYT?
Removing the `More info` section when the `error log` is visible can be problematic as they are 2 different components which do not have a direct relationship. I kept both for this reason. If it's a problem i can figure out something.

When the gif freezes a bit it's because the system dialog pop up and i click on cancel.
action_onboarding_error

mairin:
> @mairin i still have to fix some stuff but this could give you an idea of how it is... WDYT? Removing the `More info` section when the `error log` is visible can be problematic as they are 2 different components which do not have a direct relationship. I kept both for this reason. If it's a problem i can figure out something.

I'm really liking this @lstocchi!!! I think having both the log and more info available is a good idea too! What do you think?

lstocchi:
> I'm really liking this @lstocchi!!! I think having both the log and more info available is a good idea too! What do you think?

+1 to have both. I think it's useful to always see the `more info` toggle. If the action keeps failing maybe the official documentation can give some hints too. 
Aaaand ... we would not have the problem to decide when to show the "more info" label again if the command fails 😆

tailwind.config.cjs
```diff
@@ -21,7 +21,7 @@ const tailwindColors = require('tailwindcss/colors')
 module.exports = {
   content: [
     'packages/renderer/index.html',
-    'packages/renderer/src/**/*.{svelte,css}',
+    'packages/renderer/src/**/*.{svelte,ts,css}',
```
benoitf, line 24: hello, could you say why we added `ts` ?

benoitf (approved):
LGTM

tailwind.config.cjs
lstocchi: I really don't know the reasoning behind it but it's like we need to tell tailwind where we use its classes even if we just generate html strings.

By looking at the documentation (https://tailwindcss.com/docs/configuration#content) they clearly state "any file that contain Tailwind classes" 

`The content section is where you configure the paths to all of your HTML templates, JS components, and any other files that contain Tailwind class names.`

And by testing it it was true because the style was not applied otherwise
image

The fact is that i see this effect only on some classes so not sure what's the exact behavior. But by adding the ts file everything works fine.

benoitf: Ok I think we didn't had TS classes before using TailWindCSS only svelte components

PR #4211, opened 2023-10-04
chore: donut design and fixes

deboer-tim:
### What does this PR do?

Updates the Donut component to match the design better, and fixes a few minor bugs:

- Grey circle and arc are thinner (closer to the design).
- Switched to using colors from the palette.
- Since x,y=radius and starting angle=0 always I simplified the function a bit.
- Made the viewbox square, reduced padding, and made font sizes relative to reduce scaling issues.
- Added end limits (before if the client incorrectly requested 125% the arc would fail, now it looks like 100%).
- Colors weren't changing if the live value crossed thresholds (e.g. green to red), made it dynamic.
- Added tooltip.

There are still several issues compared to the design, e.g.:
- The titles doesn't match.
- Unit font should be smaller.
- Values should be rounded more (e.g. should be 99Gb not 99.25Gb).

Fixing the first is likely better done in the calling code, e.g.: changing the property names or some custom code to override 'Disk space' with 'DISK'. Unless we want the Donut to be really clever the second likely should be fixed in the caller too. Either way, it felt like there are more than enough changes in one PR and better to stop here for now.

Design:

251246170-37fe06e9-47ad-45de-b5cf-691536196176

Before:

Screenshot 2023-10-04 at 5 11 40 PM

After:

Screenshot 2023-10-04 at 5 11 10 PM

N/A

### How to test this PR?

Go to Settings > Resources and look at a running Podman Machine.

### DIFF
--- a/packages/renderer/src/lib/donut/Donut.svelte
+++ b/packages/renderer/src/lib/donut/Donut.svelte
@@ -1,50 +1,43 @@
 <script lang="ts">
+import Tooltip from '../ui/Tooltip.svelte';
+
 export let percent = 0;
-
 export let size = 64;
-
 export let title = '';
-
 export let value: unknown;
 
-function polarToCartesian(centerX: number, centerY: number, radius: number, angleInDegrees: number) {
+// describes an arc with the given radius, centered at an x,y position matching the radius
+function describeArc(radius: number, endAngle: number) {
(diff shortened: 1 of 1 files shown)


mairin:
This looks great!! 😍 Agreed with all the noted issues for further improvement in other PR(s)!

jeffmaury (approved):
LGTM

benoitf (approved):

PR #4222, opened 2023-10-05
feat: create Docker context for Podman machines

### LINKED ISSUE #2630: Docker compatibility on macOS doesn't work anymore
### Bug description

Before upgrading to newer versions of Docker Desktop and Podman Desktop, the Docker compatibility layer on my MacBook was working correctly (macOS Ventura on M1). I'm currently using Docker Desktop 4.19.0 and Podman Desktop 1.0.1, and I can't run Docker commands anymore.

From what I understood, recent versions of Docker Desktop moved the default socket location from `/var/run/docker.sock` to `/Users/<username>/.docker/run/docker.sock`, and a link exists from the first to the second.

The result is that whenever I type Docker commands to interact with Podman transparently (e.g. `docker ps`), the command fails with the following error because Podman Desktop knows how to intercept the first socket location and not the second one.

`Cannot connect to the Docker daemon at unix:///Users/<user>/.docker/run/docker.sock. Is the docker daemon running?`

I found a solution for Linux in https://github.com/containers/podman-desktop/issues/2388#issuecomment-1538389686, but I didn't find a way to make it work on macOS. Do you have any suggestion?

### Operating system

macOS Ventura 13.3.1 (M1)

### Version

1.0.x

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #2630

### What does this PR do?

Create a Docker context if Docker is installed and supports context when creating a Podman machine.
If the default Docker context is not reachable, then this context become the default one.

docker-context

#2630 

### How to test this PR?

1. Create a Podman machine
2. Check the docker contexts: `docker context list`

### DIFF
--- a/extensions/podman/src/docker-cli.spec.ts
+++ b/extensions/podman/src/docker-cli.spec.ts
@@ -1,0 +1,126 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { expect, test, vi } from 'vitest';
+import type { RunResult } from '@podman-desktop/api';
+import * as extensionApi from '@podman-desktop/api';
+import { getDockerCli } from './docker-cli';
+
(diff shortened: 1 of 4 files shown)


benoitf (commented):
Hello, I think it should be handled by the Podman CLI, not podman desktop as users can still use CLi to create machines

jeffmaury:
> Hello, I think it should be handled by the Podman CLI, not podman desktop as users can still use CLi to create machines

Do we have an issue on the CLI side ?

benoitf:
I think it's linked there https://github.com/containers/podman-desktop/issues/2630#issuecomment-1562847263
https://github.com/containers/podman/issues/16548

benoitf (commented):
I'm wondering if we keep it in Podman Desktop (as you did a lot of work so better to keep) if we should not ask at some point the user like when we detect that 'podman system connection' is not matching the one that we started.

Like: 'We detected you've docker CLI configured to use a local context of `$HOME/.docker/run/docker.sock` but docker is not running. Do you want to set docker CLI  context to the podman machine. ? (or something like that)

here is what we do for podman system connection

https://github.com/containers/podman-desktop/blob/main/extensions/podman/src/extension.ts#L267-L285

benoitf (commented):
also, I would recommend to move out code refactoring as a separate PR (like how we compute socket path, name of the is**Alive method, etc

Issue #4226, opened 2023-10-05
Inconsistencies in DetailsPage


deboer-tim:
### Is your enhancement related to a problem? Please describe

The DetailsPage page is slightly inconsistent with other/similar pages like FormPage, which make it look slightly off/things move when switching pages:
- The breadcrumb/close button had differences (fixed in PR #4182).
- The close button is offset on the details page vs breadcrumb/other pages.
- Status icons were too small (fixed via PR #4167).
- 'shadow-pageheader' is missing vs the design, subheader font is too big.

### Describe the solution you'd like

The close button should be moved down to the correct location. One thing that's blocking this is the ContainerStatistics when a container is running. Since this component is unique and reasonably large, you can't keep the same height without the statistics overlapping with tab titles or causing layout issues when the window is narrow.

We discussed on the UX call last week whether the container stats should be shown 'globally' at all or moved to somewhere like the Summary, or perhaps be in a dropdown so it is not always visible. Without making more drastic changes, the simpler solution for now is to switch ContainerStatistics to use the new Donuts component, since that's more consistent and is much narrower.

PR #4227, opened 2023-10-05
chore: use donuts for container stats

deboer-tim:
### What does this PR do?

Switches the container statistics from bars to use the Donut component. This is more consistent, will match what we do in Settings > Resources, visually nice, and takes up far less horizontal space vs today (which is necessary for #4226 - details page consistency with design and other forms).

Minor benefit: the colors are now from the palette.

The donuts are taller than the previous component, but still visually look a little small. To me this is ok and still legible, and I can't go any bigger and solve #4226.

Two minor fixes for Donut: move the tooltip below instead of right (since container stats are shown near the right edge of the window) and do not show the value until it is set. This avoids showing 'undefined' before the container stats are available.

Before:

Screenshot 2023-10-04 at 11 22 37 PM

After:

Screenshot 2023-10-05 at 4 01 37 PM

Partial fix for #4226.

### How to test this PR?

Start a container and go to its details page.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerStatistics.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerStatistics.spec.ts
@@ -1,0 +1,174 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


dgolovin (approved):
Works for me, but it needs a component test before merging.

benoitf (commented):
as suggested by @dgolovin a component test is missing

deboer-tim:
Rebased and tests added. @dgolovin or @cdrage ok to merge now?

dgolovin (approved):
LGTM

PR #4242, opened 2023-10-07
feat: adding copy button to details page

### LINKED ISSUE #4241: feat: add copy button next to title in details page

I don't like selecting manually the image name especially when it is very long

Adding a copy button next to the title on details page

_No response_

### Additional context

_No response_

axel7083:
fixes: https://github.com/containers/podman-desktop/issues/4241

https://github.com/containers/podman-desktop/assets/42176370/beb1df00-5d4c-47d9-b768-8a53d22bfe36

### DIFF
--- a/packages/renderer/src/lib/ui/DetailsPage.svelte
+++ b/packages/renderer/src/lib/ui/DetailsPage.svelte
@@ -1,7 +1,10 @@
 <script lang="ts">
 import { lastPage, currentPage } from '../../stores/breadcrumb';
+import { faCopy } from '@fortawesome/free-solid-svg-icons';
+import Fa from 'svelte-fa';
 import { router } from 'tinro';
 import Link from './Link.svelte';
+import Tooltip from './Tooltip.svelte';
 
 export let title: string;
 export let titleDetail: string | undefined = undefined;
@@ -15,6 +18,17 @@ function handleKeydown(e: KeyboardEvent) {
   if (e.key === 'Escape') {
     close();
     e.preventDefault();
+  }
+}
+
+let copyToolTip = 'copy';
+
+async function handleCopyTitle() {
+  try {
+    await navigator.clipboard.writeText(title);
+    copyToolTip = 'copied';
+  } catch (err) {
+    console.error('Failed to copy title: ', err);
   }
 }
 </script>
@@ -40,6 +54,16 @@ function handleKeydown(e: KeyboardEvent) {
         <div class="flex flex-col grow pr-2">
           <div class="flex flex-row">
             <h1 aria-label="{title}" class="text-xl leading-tight">{title}</h1>
+            <button
+              title="Copy To Clipboard"
+              class="mr-2 p-1"
+              on:click="{handleCopyTitle}"
+              on:mouseenter="{() => (copyToolTip = 'copy')}">
+              {' '}
+              <Tooltip tip="{copyToolTip}" top>
+                <Fa icon="{faCopy}" size="xs" />
+              </Tooltip>
+            </button>
             <div class="text-violet-400 ml-2 leading-normal" class:hidden="{!titleDetail}">{titleDetail}</div>
           </div>
           <div class="pt-1">
--- a/packages/renderer/src/lib/ui/DetailsPage.svelte
+++ b/packages/renderer/src/lib/ui/DetailsPage.svelte
@@ -2,6 +2,7 @@
 import { lastPage, currentPage } from '../../stores/breadcrumb';
 import { faCopy } from '@fortawesome/free-solid-svg-icons';
 import { router } from 'tinro';
+import Fa from 'svelte-fa';
 import Link from './Link.svelte';
 import Tooltip from './Tooltip.svelte';
 
--- a/packages/renderer/src/lib/ui/DetailsPage.svelte
+++ b/packages/renderer/src/lib/ui/DetailsPage.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
 import { lastPage, currentPage } from '../../stores/breadcrumb';
 import { faCopy } from '@fortawesome/free-solid-svg-icons';
-import Fa from 'svelte-fa';
 import { router } from 'tinro';
 import Link from './Link.svelte';
 import Tooltip from './Tooltip.svelte';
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/ui/DetailsPage.svelte
```diff
@@ -17,6 +20,17 @@ function handleKeydown(e: KeyboardEvent) {
     e.preventDefault();
   }
 }
+
+let copyToolTip = 'copy';
```
lstocchi, line 24: Should it mention what you're going to copy? `Copy image name`

lstocchi (commented):
LGTM, missing a test?

packages/renderer/src/lib/ui/DetailsPage.svelte
axel7083: The details page is shared between many elements, such as the container, images, volumes. how am i suppose to know where is it being used ? Adding a new property that the parent should give ?

lstocchi: you could use the breadcrumb info or maybe just add a generic `copy name`

deboer-tim:
I really like this capability & convenience, but I don't love the copy button on every page, and I think there are other things that should be copyable (e.g. ids or tags) which would mean the buttons show up beside a lot of fields.

In the Mac password dialog the user or password become links when you hover over them, and there's a popup menu to copy it. It's horrible for discovery (yet I found it easily enough) but also means you can click anywhere on the field. Maybe we don't even need a popup and copy whenever you click on the hover? I don't know, just some ideas to reduce visual clutter and make copying as easy as possible.

@mairin any thoughts on the best way to enable this?

axel7083:
> I really like this capability & convenience, but I don't love the copy button on every page, and I think there are other things that should be copyable (e.g. ids or tags) which would mean the buttons show up beside a lot of fields.
> 
> In the Mac password dialog the user or password become links when you hover over them, and there's a popup menu to copy it. It's horrible for discovery (yet I found it easily enough) but also means you can click anywhere on the field. Maybe we don't even need a popup and copy whenever you click on the hover? I don't know, just some ideas to reduce visual clutter and make copying as easy as possible.
> 
> @mairin any thoughts on the best way to enable this?

From my point of view, I am always copying either images  names or container names, and the current way is to select the text then copy past, for very long names, it always takes a few seconds and makes it pretty repetitive. 

I do not like the personally the idea of having a popup for copying elements, the hover tooltip is pretty self explanatory in this context, and the icon used here is also pretty common, therefore confirmation for clicking on a copy icon is probably not useful.

The idea of making the title a link, with a hover message copy, thus doing the same, is also a good solution. In the same way as the button, if a hover tooltip show "Copy content" or something similar, it should not show a popup making it a procedure as long as selecting and copying manually.

Just tell me which solution you would prefer, and I will adapt the code

axel7083:
> LGTM, missing a test?

Done!

mairin:
> > @mairin any thoughts on the best way to enable this?

Sorry, I had it on my list to do a quick mockup for this but time got away from me -

my thinking was to have the clipboard icon appear on hover / kb focus - so you wont have the visual clutter and it'll scale visually. and if its something to copy your instinct might ve to try to select + copy, so you'll naturally discover the hover action enroute to that.

what do you think

cdrage:
> > > @mairin any thoughts on the best way to enable this?
> 
> Sorry, I had it on my list to do a quick mockup for this but time got away from me -
> 
> my thinking was to have the clipboard icon appear on hover / kb focus - so you wont have the visual clutter and it'll scale visually. and if its something to copy your instinct might ve to try to select + copy, so you'll naturally discover the hover action enroute to that.
> 
> what do you think

Sums up from me too, @axel7083 sorry for the long wait, are you able to update the PR with @mairin 's suggestion and let's get this merged in?

axel7083:
> > > @mairin any thoughts on the best way to enable this?
> 
> Sorry, I had it on my list to do a quick mockup for this but time got away from me -
> 
> my thinking was to have the clipboard icon appear on hover / kb focus - so you wont have the visual clutter and it'll scale visually. and if its something to copy your instinct might ve to try to select + copy, so you'll naturally discover the hover action enroute to that.
> 
> what do you think

I am not quick sure to get what you mean by appear on hover:
- Do you mean on the tooltip ? (Then the current Tooltip implementation only support String so it cannot be added).
- Do you mean having literally the mouse icon to be the copy icon ? (Very complicated to do, so i'm guessing this is really not that)
- Adding the copy icon on the right of the title while it is hovering the title, with a tooltip with "Copy" text.

I went with the last option (see video in the Edit section above)

trmendes:
Not planing on changing the flow of this conversation, but I would like to share my feelings on the topic.

I spent some time checking what is the docker desktop approach for providing the user a way to copy content, and it same we do. They have “copy to clipboard” buttons. Difference is that the copy button is not so bright as ours. They usually make it more gray.

I personally think we could just be more careful choosing the places to have such copy button and, find a more soft way (can't find a better word for it) to present such icon. 

The option to have a different approach is also nice, but I also struggle to fully grasp the concept.

What comes to my mind is a pop-up (tooltip) with a copy button on it, just like the ones we have on mobile platforms.

If that is that case, I would be confused on what is the content to be copied: The hovered content or the text that is shown in the tooltip, together with the copy button.

@deboer-tim maybe you could post a screenshot or a video? :D

deboer-tim:
Hi @trmendes. I tried to throw a quick sample together twice and it didn't quite work, so no video. :) My fallback thought was that we should make a /ui component that puts a copy button on the right for now, with clean enough API that if we ever came up with another design we could easily change it in the future without changing all the places that would use it.

I just looked at what you did in #5687 (thank you, btw), and CopyToClipboard is basically that. 👍🏼 We need to use it in a few more places to see if there are any tweaks necessary and it works well anywhere, but that's a great start.

trmendes:
@deboer-tim Thank you for your answer. Do we have an issue listing the places we would like to make usagge of ```CopyToClipboard``` component?

A quick search for 'Clipboard' on main:

```
TroubleshootingDevToolsConsoleLogs.svelte
EmptyScreen.svelte
```

deboer-tim:
I'm only aware of these two issues open (#4241 and #5608), but I think I've seen a couple others in messages/forums, and IMHO any sha/url could be requested at some point. So we can start with these two and get some feedback/use and see if others are requested.

PR #4244, opened 2023-10-07
feat: StatusBarItem notification icon

### LINKED ISSUE #4243: feat: adding badge to status bar item such as tasks on specific events

When a task is created, nothing is indicating it. This can be frustrating because we have no idea something is happening (E.g. an image is being copying to kind cluster or something is being download.)

A simple `badge` on the task icon when a task is created.

_No response_

### Additional context

_No response_

axel7083:
fixes: https://github.com/containers/podman-desktop/issues/4243

### What does this PR do?

When registering the `Tasks` status bar item, we subscribe using the api sender to the task creation event, and we add the badge to the StatusBarItem entry of tasks, when we open the tasks manager, we remove the badge using the api sender event.

https://github.com/containers/podman-desktop/assets/42176370/e0deb3c7-3c82-4c95-841b-2e8a7cdfbd2b

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1195,10 +1195,6 @@ declare module '@podman-desktop/api' {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     commandArgs?: any[];
     /**
-     * Marks an item to have a badge.
-     */
-    badged: boolean;
-    /**
      * Shows the entry in the status bar.
      */
     show(): void;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1195,6 +1195,10 @@ declare module '@podman-desktop/api' {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     commandArgs?: any[];
     /**
+     * Marks an item to have a badge.
+     */
+    badged: boolean;
+    /**
      * Shows the entry in the status bar.
      */
     show(): void;
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -490,7 +490,7 @@ export class PluginSystem {
     );
 
     commandRegistry.registerCommand('show-task-manager', () => {
-      statusBarRegistry.setBadged('tasks', false);
+      statusBarRegistry.setDotted('tasks', false);
       apiSender.send('toggle-task-manager', '');
     });
 
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -395,8 +395,8 @@ export class PluginSystem {
 
     const iconRegistry = new IconRegistry(apiSender);
     const directories = new Directories();
-
-    const taskManager = new TaskManager(apiSender);
+    const statusBarRegistry = new StatusBarRegistry(apiSender);
+    const taskManager = new TaskManager(apiSender, statusBarRegistry);
     const notificationRegistry = new NotificationRegistry(apiSender, taskManager);
 
     const configurationRegistry = new ConfigurationRegistry(apiSender, directories, notificationRegistry);
@@ -422,7 +422,6 @@ export class PluginSystem {
     const cancellationTokenRegistry = new CancellationTokenRegistry();
     const providerRegistry = new ProviderRegistry(apiSender, containerProviderRegistry, telemetry);
     const trayMenuRegistry = new TrayMenuRegistry(this.trayMenu, commandRegistry, providerRegistry, telemetry);
-    const statusBarRegistry = new StatusBarRegistry(apiSender);
     const inputQuickPickRegistry = new InputQuickPickRegistry(apiSender);
     const fileSystemMonitoring = new FilesystemMonitoring();
     const customPickRegistry = new CustomPickRegistry(apiSender);
@@ -477,22 +476,6 @@ export class PluginSystem {
     });
 
     statusBarRegistry.setEntry('help', false, 0, undefined, 'Help', 'fa fa-question-circle', true, 'help', undefined);
-
-    statusBarRegistry.setEntry(
-      'tasks',
-      false,
-      0,
-      undefined,
-      'Tasks',
-      'fa fa-bell',
-      true,
-      'show-task-manager',
-      undefined,
-    );
-
-    apiSender.receive('task-created', () => {
-      statusBarRegistry.setBadged('tasks', true);
-    });
 
     statusBarRegistry.setEntry(
       'troubleshooting',
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -476,18 +476,7 @@ export class PluginSystem {
       }
     });
 
-    statusBarRegistry.setEntry(
-      'help',
-      false,
-      0,
-      undefined,
-      'Help',
-      'fa fa-question-circle',
-      true,
-      'help',
-      undefined,
-      false,
-    );
+    statusBarRegistry.setEntry('help', false, 0, undefined, 'Help', 'fa fa-question-circle', true, 'help', undefined);
 
     statusBarRegistry.setEntry(
       'tasks',
@@ -499,7 +488,6 @@ export class PluginSystem {
       true,
       'show-task-manager',
       undefined,
-      false,
     );
 
     apiSender.receive('task-created', () => {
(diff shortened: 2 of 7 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -1103,6 +1103,10 @@ declare module '@podman-desktop/api' {
      */
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     commandArgs?: any[];
+    /**
+     * Marks an item to have a badge.
+     */
+    badged: boolean;
```
benoitf: it should not be exposed (we're mimic VS Code extension there) and to avoid Xmas tree effect, it should be limited to internal part

axel7083:
@benoitf is the macOS pr-check flaky ? I'm not getting the error

packages/extension-api/src/extension-api.d.ts
jeffmaury: Should be optional otherwise looks like an API breakage ?

axel7083: definitely, I will fix that later

benoitf (commented):
forgot to submit the review I did...

packages/extension-api/src/extension-api.d.ts
axel7083: I see !

packages/renderer/src/lib/statusbar/StatusBarItem.svelte
```diff
@@ -1,3 +1,20 @@
...
+  display: inline-block;
+}
+
+.status-bar-item .badge {
+  position: absolute;
+  top: -4px;
+  right: -4px;
+  padding: 4px 4px;
+  border-radius: 50%;
+  background-color: var(--pf-global--primary-color--100);
```
cdrage: Is there any way to not have the hardcoded style / CSS here? What do you think @benoitf 

Otherwise, this PR LGTM.

cdrage (changes requested):

packages/renderer/src/lib/statusbar/StatusBarItem.svelte
axel7083: I updated, and use the class when possible!

axel7083:
CC @cdrage

packages/renderer/src/lib/statusbar/StatusBarItem.spec.ts
```diff
@@ -44,6 +46,7 @@ test('check iconClass with font awesome icons and spinning', () => {
   const statusBarEntry: StatusBarEntry = {
     enabled: true,
     activeIconClass: 'fas fa-sync~spin',
+    badged: false,
   };
 
   const icon = iconClass(statusBarEntry);
```
benoitf, line 54: there is no test checking that the badged is displayed or not

packages/renderer/src/lib/statusbar/StatusBarItem.svelte
```diff
@@ -35,12 +35,15 @@ async function executeCommand(entry: StatusBarEntry) {
...
+  class="{opacity(entry)} px-1 flex items-center {hoverBackground(entry)} {hoverCursor(entry)} relative inline-block"
   title="{tooltipText(entry)}">
   {#if iconClass(entry)}
     <span class="{iconClass(entry)}" aria-hidden="true"></span>
   {/if}
   {#if entry.text}
     <span class="ml-1">{entry.text}</span>
   {/if}
+  {#if entry.badged}
+    <span class="absolute bg-purple-500" style="border-radius: 50%; padding: 4px 4px;top: -4px; right: -2px;"></span>
```
benoitf: AFAIK style should not be used but their tailwind equivalent should be used

packages/main/src/plugin/index.ts
```diff
@@ -483,6 +483,10 @@ export class PluginSystem {
       undefined,
     );
 
+    apiSender.receive('task-created', () => {
+      statusBarRegistry.setBadged('tasks', true);
+    });
+
```
benoitf: here, it feels to me that we're moving logic to a place where it shouldn't (it would be harder to maintain when some logic code is spread across different classes)

It should be the taskManager that updates its statusBar entry

packages/main/src/plugin/statusbar/statusbar-item.ts
```diff
@@ -74,6 +75,15 @@ export class StatusBarItemImpl implements StatusBarItem {
...
 
+  public set badged(badged: boolean) {
+    this._badged = badged;
+    this.update();
+  }
+
+  public get badged(): boolean {
+    return this._badged;
+  }
+
```
benoitf: I would suggest to use another term than badge as badge would be more the overall entry element (also we have a BadgeUI element which is displaying small letters inside so it might be confusing to have "badges" and "badged"

maybe dot ? as we display a dot/circle ?

packages/main/src/plugin/index.ts
benoitf: we initialize StatusBarRegistry and TaskManager in index.ts

maybe move initialization of StatusBarRegistry earlier, then in TaskManager we do the 
```
statusBarRegistry.setEntry(
      'tasks',
      false,
      0,
      undefined,
      'Tasks',
      'fa fa-bell',
      true,
      'show-task-manager',
      undefined,
    );
```

this method could return the entry object

then instead of doing somewhere

```ts
 apiSender.receive('task-created', () => {
      statusBarRegistry.setBadged('tasks', true);
    });
```

we do in Taskmanager:

```ts
statusBarTaskEntry.dotted = true;
```

Of course, I would split that enhancement in two:

first, we change order of initialization, return entry object when creating an entry, etc, without introducing any badged/dot/circle

it also means that we could move

```

    commandRegistry.registerCommand('show-task-manager', () => {
      statusBarRegistry.setBadged('tasks', false);
      apiSender.send('toggle-task-manager', '');
    });
```

to the TaskManager instance

so everything managing the icon would be in that class

benoitf (changes requested):

packages/main/src/plugin/statusbar/statusbar-item.ts
axel7083: Good idea, go for dotted

packages/main/src/plugin/index.ts
axel7083: Yes, did that

packages/renderer/src/lib/statusbar/StatusBarItem.svelte
axel7083: Yes, this PR is old, I wasn't aware of it, fixed it

packages/renderer/src/lib/statusbar/StatusBarItem.spec.ts
axel7083: added

deboer-tim (commented):
Could you use a different name than 'dotted' and 'badged: boolean' in main? First, when I saw the name I thought you were going to make the badge have a dotted outline or something. :) Second, it's mixing presentation (the fact that it's shown as a badge) with API (what the purpose of the boolean is - it's a new task) - ideally we'd be able to change the UI completely and API wouldn't need to change.

This is notification for new tasks, so maybe a name like 'isNew' or 'userNotified'?

benoitf (commented):
please do separate PRs to ease the merge
small PRs = small review = fast merge

-one small PR for example to return the entry if you create it
- one  small PR to have TaskManager creating the statusbar entry and register the command
(no new feature, just changing how code is organized)
- one  small PR that brings the 'fact that you want to give awareness' option
- one PR that use the feature in the task manager

benoitf:
> This is notification for new tasks, so maybe a name like 'isNew' or 'userNotified'?

@deboer-tim  The status bar entry is not related to tasks. The role is to add an icon and call a command if you click on it. So you have the same usage for the notification icon or for the bulb icon or feedback icon, etc

As it's generic, if you add a `isNew` I don't know what it would mean if for example if we do that on the bulb icon or on the feedback icon.

The renderer part is the same whatever entry is, so name might be generic but should not be meaningful only for tasks

axel7083:
> please do separate PRs to ease the merge small PRs = small review = fast merge
> 
> -one small PR for example to return the entry if you create it
> 
> * one  small PR to have TaskManager creating the statusbar entry and register the command
>   (no new feature, just changing how code is organized)
> * one  small PR that brings the 'fact that you want to give awareness' option
> * one PR that use the feature in the task manager

Thanks for the last-minute advice. I'll split the PRs as you now recommend.

axel7083:
Closing in favor of

https://github.com/containers/podman-desktop/pull/5689
https://github.com/containers/podman-desktop/pull/5690
https://github.com/containers/podman-desktop/pull/5691

PR #4257, opened 2023-10-09
chore(ui,tests): include ARIA attributes into Create a Podman machine page and Resources page

xbabalov:
### What does this PR do?
Provides ARIA labels for the Create a Podman machine page and the Settings Resources page.

https://github.com/containers/podman-desktop/issues/3935

### How to test this PR?
yarn test

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
@@ -144,7 +144,7 @@ function getLoggerHandler(
   {#if connection.lifecycleMethods && connection.lifecycleMethods.length > 0}
     <div class="mt-2 relative">
       <!-- TODO: see action available like machine infos -->
-      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto">
+      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto" role="group" aria-label="ConnectionActions">
         {#if connection.lifecycleMethods.includes('start')}
           <div class="ml-2">
             <LoadingIconButton
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
@@ -144,7 +144,7 @@ function getLoggerHandler(
   {#if connection.lifecycleMethods && connection.lifecycleMethods.length > 0}
     <div class="mt-2 relative">
       <!-- TODO: see action available like machine infos -->
-      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto" role="group" aria-label="ConnectionActions">
+      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto" role="group" aria-label="Connection Actions">
         {#if connection.lifecycleMethods.includes('start')}
           <div class="ml-2">
             <LoadingIconButton
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationRendering.svelte
@@ -431,7 +431,7 @@ function closePage() {
             class="p-2 space-y-7 h-fit"
             on:submit|preventDefault="{handleOnSubmit}"
             bind:this="{formEl}"
-            aria-label="ResourceProperties">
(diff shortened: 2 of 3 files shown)


packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
```diff
@@ -144,7 +144,7 @@ function getLoggerHandler(
   {#if connection.lifecycleMethods && connection.lifecycleMethods.length > 0}
     <div class="mt-2 relative">
       <!-- TODO: see action available like machine infos -->
-      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto">
+      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto" role="group" aria-label="ConnectionActions">
```
jeffmaury: Just wondering is the role attribute required ? Shouldn't the label be enough ?

xbabalov: It's not strictly required but it allows for more precise locating in playwright in testing, since getByRole()  has more options to work with than getByLabel().

odockal (approved):

packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
odockal: @jeffmaury I would say that if we got an option to use a specific role that fits into a HTML structure and is missing (could be region, form, etc.) we should add it. Otherwise we could be alright with a HTML provided role. but since we got a lot of "containers" as `div`s we need to add role ourselves.

packages/renderer/src/lib/preferences/PreferencesConnectionCreationRendering.svelte
```diff
@@ -426,7 +426,12 @@ function closePage() {
           {#if connectionAuditResult && (connectionAuditResult.records?.length || 0) > 0}
             <AuditMessageBox auditResult="{connectionAuditResult}" />
           {/if}
-          <form novalidate class="p-2 space-y-7 h-fit" on:submit|preventDefault="{handleOnSubmit}" bind:this="{formEl}">
+          <form
+            novalidate
+            class="p-2 space-y-7 h-fit"
+            on:submit|preventDefault="{handleOnSubmit}"
+            bind:this="{formEl}"
+            aria-label="ResourceProperties">
```
benoitf: I think we can use space in the label name. Label should be more "human readable" than camelCase IMHO

"Properties information"

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -459,7 +462,10 @@ function isOnboardingEnabled(provider: ProviderInfo, globalContext: ContextUI):
 
               {#if providerContainerConfiguration.has(provider.internalId)}
                 {@const providerConfiguration = providerContainerConfiguration.get(provider.internalId) || []}
-                <div class="flex mt-3 {container.status !== 'started' ? 'text-gray-900' : ''}">
+                <div
+                  class="flex mt-3 {container.status !== 'started' ? 'text-gray-900' : ''}"
+                  role="group"
+                  aria-label="ProviderConfiguration">
```
benoitf: maybe more human readable ?
```suggestion
                  aria-label="Configuration of the provider">
```

packages/renderer/src/lib/preferences/PreferencesConnectionCreationRendering.svelte
odockal: that is a good point. For now we stick to camel case since I am not sure that AT can make a difference, or how it is read by such tool if there is space missing. 

I have checked existing aria-labels in the codebase, it is true that most cases are standard text with spaces. So let's stick to that variant if possible.

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
odockal: `Provider configuration` is the same but shorter, right?

packages/renderer/src/lib/preferences/PreferencesConnectionCreationRendering.svelte
xbabalov: It should be fixed, I'll keep it in mind for future labels.

PR #4266, opened 2023-10-10
feat: check if reboot is needed after wsl install (#4193)

lstocchi:
### What does this PR do?

This PR adds a check to see if the system must be restarted when WSL is installed.
In older version of Windows it is mandatory. 

N/A

it is part of #4193 

### How to test this PR?

1. you need to be on win 10
2. uninstall wsl
3. restart the machine
4. install wsl and do not reboot
5. launch Desktop
6. start the onboarding, you should see the warning telling you to restart the system

### DIFF
--- a/extensions/podman/src/podman-install.ts
+++ b/extensions/podman/src/podman-install.ts
@@ -576,7 +576,7 @@ class WSL2Check extends BaseCheck {
       throw new Error(message);
     }
   }
-  
+
   private async isRebootNeeded(): Promise<boolean> {
     try {
       await extensionApi.process.exec('wsl', ['-l'], {
--- a/extensions/podman/src/podman-install.ts
+++ b/extensions/podman/src/podman-install.ts
@@ -583,12 +583,16 @@ class WSL2Check extends BaseCheck {
         env: { WSL_UTF8: '1' },
       });
     } catch (error) {
-      console.log(error);
       // we only return true for the WSL_E_WSL_OPTIONAL_COMPONENT_REQUIRED error code
       // as other errors may not be connected to a reboot, like
       // WSL_E_DEFAULT_DISTRO_NOT_FOUND = wsl was installed without the default distro
       if (error.stdout.includes('Wsl/WSL_E_WSL_OPTIONAL_COMPONENT_REQUIRED')) {
         return true;
+      } else if (error.stdout.includes('Wsl/WSL_E_DEFAULT_DISTRO_NOT_FOUND')) {
+        // treating this log differently as we install wsl without any distro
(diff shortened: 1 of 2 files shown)


extensions/podman/src/podman-install.ts
```diff
@@ -550,4 +556,21 @@ class WSL2Check extends BaseCheck {
...
     }
   }
+
+  private async isRebootNeeded(): Promise<boolean> {
+    try {
+      await extensionApi.process.exec('wsl', ['-l'], {
+        env: { WSL_UTF8: '1' },
+      });
+    } catch (error) {
+      console.log(error);
```
benoitf: do we need to log the error like that or we need to log why we have this error and log only if the component is not required ?

benoitf (approved):

extensions/podman/src/podman-install.ts
lstocchi: i just added it for testing and forgot it. It should also be `console.error`. Fixing

extensions/podman/src/podman-install.ts
```diff
@@ -550,4 +556,25 @@ class WSL2Check extends BaseCheck {
...
+  private async isRebootNeeded(): Promise<boolean> {
+    try {
+      await extensionApi.process.exec('wsl', ['-l'], {
+        env: { WSL_UTF8: '1' },
+      });
+    } catch (error) {
+      // we only return true for the WSL_E_WSL_OPTIONAL_COMPONENT_REQUIRED error code
+      // as other errors may not be connected to a reboot, like
+      // WSL_E_DEFAULT_DISTRO_NOT_FOUND = wsl was installed without the default distro
+      if (error.stdout.includes('Wsl/WSL_E_WSL_OPTIONAL_COMPONENT_REQUIRED')) {
```
jeffmaury, line 589: Just wondering if we cannot rely on the returned error code for better reliability ?

lstocchi: I didn't feel it was very reliable. I received the error code `4294967295` but by doing a search i found that it is used in other cases (like if the distro fails to start https://github.com/microsoft/WSL/issues/5092) so i'm not sure if when executing `wsl -l` the error code is only associated to the reboot or it's just a case.

lstocchi: @jeffmaury ping. What do you suggest here?

jeffmaury: If they don't use a specific error code then you can simply ignore my remark :)

jeffmaury (approved):
LGTM

PR #4282, opened 2023-10-11
feat: add WSL version pre-flight check (#3659)

### LINKED ISSUE #3659: Improve pre-flight checks on Windows for podman

On Windows we are doing some checks

but it seems that we could do other checks around WSL and the windows version

here are some Windows WSL prerequisite
https://github.com/containers/podman-desktop/discussions/3151#discussioncomment-6754515

they should be converted as "pre-flight" checks

it could be like warnings (you can proceed but you have a notice) or as errors if we know it won't work

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR adds a new pre-flight check to verify is the WSL version is >= 1.2.5.0.

N/A

it resolves #3659 

### How to test this PR?

1. run tests

### DIFF
--- a/extensions/podman/src/podman-install.ts
+++ b/extensions/podman/src/podman-install.ts
@@ -37,6 +37,7 @@ import {
   START_NOW_MACHINE_INIT_SUPPORTED_KEY,
   isStartNowAtMachineInitSupported,
 } from './extension';
+import { WslHelper } from './wsl-helper';
 
 const readFile = promisify(fs.readFile);
 const writeFile = promisify(fs.writeFile);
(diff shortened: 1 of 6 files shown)


extensions/podman/src/podman-install.ts
```diff
@@ -576,3 +566,48 @@ class WSL2Check extends BaseCheck {
...
+          docLinksDescription: `Call 'wsl --version' in a terminal to check your wsl version.`,
+        });
+      }
+      const version = firstLine.substring(colonPosition + 1).trim();
+      if (compare(version, this.minVersion, '>=')) {
+        return this.createSuccessfulResult();
+      } else {
+        return this.createFailureResult({
+          description: `Your WSL version is ${version} but it should be >= ${this.minVersion}.`,
+          docLinksDescription: `Call 'wsl --update' to update your WSL installation. If you do not have access to the Windows store you can run 'wsl --update --web-download'. If you still receive an error please contact your IT administator as 'Windows Store Applications' may have been disabled.`,
```
benoitf: maybe you should use https://github.com/containers/podman-desktop/blob/main/extensions/podman/src/wsl-helper.ts ?

lstocchi: 😱 i didn't notice it. It won't work though as it only matches the english lang. I'll fix and use that 👍

benoitf: thanks 👍

lstocchi: done 👍 and thanks for letting me know of `WSLHelper` 🙏

benoitf (approved):

Issue #4113, opened 2023-09-27
Make cpu/memory/disk of Podman machine being editable


benoitf:
### Is your enhancement related to a problem? Please describe

If I create a podman machine with low CPU and want to increase it after seeing the usage done in https://github.com/containers/podman-desktop/issues/3063

I can't use the UI

### Describe the solution you'd like

I should be able to edit the cpu/memory/disk, it'll restart the podman machine when applying the changes

It could be done in the details page of the machine or somewhere else

jeffmaury:
For Podman, only a few options are available:

On Windows, only rootful and user-mode-networking are supported but requires a stopped machine

On MacOS, everything is supported but requires a stopped machine.

So I suggest:

- allow edit on stopped machine
- declare properties with a new field editable (that should be an expression)

@benoitf @deboer-tim WDYT ?

deboer-tim:
Sorry, delayed 'yes'. I think it's really unfortunate that we have to use 'editable-when' expressions and I think we should keep it as simple (logic and code) as possible, but it's probably inevitable here. This isn't super-obvious to users either, but I think per-platform people will figure out you can edit stopped machines and what you can change.

PR #4307, opened 2023-10-12
refactor: extract value retrieval from PreferencesRenderingItemFormat

jeffmaury:
### What does this PR do?

While preparing #4113 I noticed there was already a form to edit machine properties (the one used when creating a machine) but as the scope is different a sub component (PreferencesRenderingItemFormat) was retrieving the value with code embedded into the component. This refactoring just extract value retrieval outside of the component

N/A

N/A

### How to test this PR?

Check tests

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
@@ -90,7 +90,7 @@ $: resetToDefault = false;
         updateResetButtonVisibility="{updateResetButtonVisibility}"
         resetToDefault="{resetToDefault}"
         enableAutoSave="{true}"
-        initialValue="{() => getInitialValue(recordUI.original)}" />
+        initialValue="{getInitialValue(recordUI.original)}" />
     {/if}
   </div>
   {#if recordUI.original.type !== 'string' || (recordUI.original.enum && recordUI.original.enum.length > 0)}
@@ -99,6 +99,6 @@ $: resetToDefault = false;
       updateResetButtonVisibility="{updateResetButtonVisibility}"
       resetToDefault="{resetToDefault}"
       enableAutoSave="{true}"
-      initialValue="{() => getInitialValue(recordUI.original)}" />
+      initialValue="{getInitialValue(recordUI.original)}" />
   {/if}
 </div>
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.svelte
```diff
@@ -21,6 +20,8 @@ export let setRecordValue = (_id: string, _value: string) => {};
 export let enableSlider = false;
 export let record: IConfigurationPropertyRecordedSchema;
 
+export let initialValue: any;
```
benoitf: it's not clear to me why type is any as it looks like you want to have an async function so it should at least say that it's a function

packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
```diff
@@ -88,14 +89,16 @@ $: resetToDefault = false;
         record="{recordUI.original}"
         updateResetButtonVisibility="{updateResetButtonVisibility}"
         resetToDefault="{resetToDefault}"
-        enableAutoSave="{true}" />
+        enableAutoSave="{true}"
+        initialValue="{async () => await getInitialValue(recordUI.original)}" />
```
benoitf: I'm not sure to follow why we need to add await here and why it couldn't return the promise ?

packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.svelte
jeffmaury: The refactoring extracts the computation of the initial value out of the component and the initialValue is provided thus the any type. And because the method that now computes the value is async this is how it can be called in a Svelte loop according to the Svelte tutorial

benoitf: if initialValue is a function, then instead of `any` I would expect `() => Promise<any>` no ?

benoitf: or probably better `() => Promise<unknown>`

benoitf (approved):
LGTM

still wondering if `Promise<any>` should be `Promise<unknown>`

PR #4317, opened 2023-10-12
feat: provider creation form

### LINKED ISSUE #2853: Clicking ESC on Create new machine/cluster, etc is not closing the window

I would expect that using the <kbd>ESC</kbd> would close the 'Create new..." option

image

<kbd>ESC</kbd>close the 'Create new..." window (as if I could hit the X icon)

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Creating an instance of a Podman Machine or Kubernetes provider had a UI that didn't quite match other places. Switching it to be a FormPage aligns the look and feel with the rest of our UI (header, breadcrumbs, etc.) and fixes things that we've already solved once in all the form pages, e.g. clicking Esc to close (#2853).

Since this page was done a while ago there was some minor restructuring to make this happen:
- The Resources page has to be marked as a details navigation hint so that the backlink/close button goes back to the correct page.
- The code to find the provider display name and image (the header info) is moved up from the connection creation page to the provider rendering form.
- This makes the provider image 'always hidden' on the component, so the property could be removed.

Before:

Screenshot 2023-10-12 at 2 05 53 PM

Screenshot 2023-10-12 at 2 11 54 PM

After:

Screenshot 2023-10-12 at 1 56 15 PM

Onboarding unchanged.

Fixes #2853.

### How to test this PR?

Create a Podman machine or Kubernetes provider.

### DIFF
--- a/packages/renderer/src/lib/onboarding/OnboardingComponent.spec.ts
+++ b/packages/renderer/src/lib/onboarding/OnboardingComponent.spec.ts
@@ -93,7 +93,7 @@ test('Expect to find PreferencesConnectionCreationRendering component if step in
   });
 
   const title = screen.getAllByRole('heading', { name: 'title' });
-  expect(title[0].textContent).equal('Create a Podman machine ');
+  expect(title[0].textContent).equal('Create a Podman machine');
 });
 
 test('Expect to find "not supported" message if step includes a component not supported by the provider', async () => {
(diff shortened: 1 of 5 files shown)


lstocchi (approved):
LGTM

My only doubt is about the header color. It looks to me a bit too dark compared to the rest. I know it depends on the background of the resources section ... it's just an impression.

deboer-tim:
Noted. I confirmed it is picking up the 'correct' color and it is the same as the top was before, so I think we review the design with Mo later.

PR #4330, opened 2023-10-13
fix: hide cog icon button if provider has no onboarding/configurations (#4320)

### LINKED ISSUE #4320: Don't show Setup button when there are no preferences
### Bug description

The Setup buttons in Settings > Resources make sense when there is onboarding to do (e.g. creating a Podman machine) or when there are preferences (e.g. OpenShift Local). For other cases though, it just goes to a preference page that says `No Settings Found` and nothing is selected in the navigation (because it is trying to show a preferences page for an extension that doesn't have any).

If an extension has no onboarding and no preferences (or no preference page), hide or disable the Setup button.

### Operating system

macOS 14.0

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

Go to Settings > Resources and click on the Setup for Kind.

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR hides the cog icon button if the provider has neither onboarding nor configurations.

Below the podman extension is in `not-installed` status and has an onboarding workflow.
Openshift local has no onboarding but it has some custom configurations, reachable with the cog icon button
Kind has no onboarding/config so we only display the `create new ...` button
image

it resolves #4320 

### How to test this PR?

1. go to resources or run tests

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.spec.ts
@@ -26,6 +26,8 @@ import userEvent from '@testing-library/user-event';
 import { router } from 'tinro';
 import { onboardingList } from '/@/stores/onboarding';
 import type { OnboardingInfo } from '../../../../main/src/plugin/api/onboarding';
+import { configurationProperties } from '/@/stores/configurationProperties';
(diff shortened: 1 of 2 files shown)


benoitf (approved):

jeffmaury (approved):
LGTM

Issue #4125, opened 2023-09-27
Onboarding: Add notification about available onboarding flows to dashboard


mairin:
### Epic
This is part of the Dashboard Revamp epic (https://github.com/containers/podman-desktop/issues/4020).

For the initial release of the onboarding feature, we should have notifications for each of the onboarding flows that are available and have not been entered / set up.

### Describe the solution you'd like

Use https://github.com/containers/podman-desktop/issues/1229 (if only one to display) or https://github.com/containers/podman-desktop/issues/1230 (if multiple to display) to display the available onboarding flows on the dashboard.

The mockups in those issues show podman and docker onboarding examples.

Issue #1230, opened 2023-01-24
UI Component: Dashboard Message (Multi)


mairin:
### Epic
This is part of the Dashboard Revamp epic (https://github.com/containers/podman-desktop/issues/4020).

### Related Components
https://github.com/containers/podman-desktop/issues/1229 -  UI Component: Dashboard Message (Simple) 

### Preview
![Screenshot from 2023-10-10 11-48-50](https://github.com/containers/podman-desktop/assets/799683/6c5293f3-a08e-4fef-bfb1-3ec53ea0b225)

### Link to Penpot file with inspectable fonts/colors/css/etc:
https://design.penpot.app/#/view/c1a43240-198d-11ed-817f-f940e3befd53?page-id=f89df008-0194-80f6-8001-9a648b96f00b&section=handoff&index=4&share-id=2a21f5ae-60ee-8151-8001-f3985b0d35fe

[Note: You don't need a Penpot username to view the mockups if you don't have one, you can log in using your GitHub account. It should then show you a code inspector for the mockup.]

## Specification

There are main parts to this dashboard message component:
* **Overall Message Headline** [optional]: Should be a slug that ideally unifies the messages contained within the dashboard multi message block (if they are thematically related.) If the messages aren't related, this could just default to reading "Notices" or could be omitted. It can use an emoji, ideally short, one-line, to the point. 
* **Individual Message Headline** [required]: Each message has its own headline. It should be a slug. Here you can see it's using an emoji. Ideally this would be one-line, short, to the point: What does the user need to do?
* **Individual Message Icon** [optional]: Per message contained in the multi message block. Icon or emoji that relates to the action item or content of the message. Here in the example we have green checkboxes to indicate something has been detected, and a red empty checkbox to indicate something is missing.
* **Individual Message Description** [required]: Per message contained in the multi message block. Explain what the user needs to do in more detail, and why, and if applicable provide pointers to relevant areas in the UI. (The underlined "Settings > Resources > Podman" is meant to be a link to that specific screen.)
* **Individual Message Call-to-Action Button** [optional]: Per message. A button that describes - ideally as a verb or verb + object - the action (if any) that the user needs to take with a button to bring them to where they need to do it. [Note this mockup has no example buttons.]
[...]

Issue #1229, opened 2023-01-24
UI Component: Dashboard Message (Simple)


mairin:
### Epic
This is part of the Dashboard Revamp epic (https://github.com/containers/podman-desktop/issues/4020).

### Related Components
https://github.com/containers/podman-desktop/issues/1230 UI Component: Dashboard Message (Multi)

### Preview
![Screenshot from 2023-01-23 20-43-15](https://user-images.githubusercontent.com/799683/214195922-7a1fdc23-b9aa-40d8-9b31-aa83e7f50c83.png)

### Link to Penpot file with inspectable fonts/colors/css/etc:
https://design.penpot.app/#/view/c1a43240-198d-11ed-817f-f940e3befd53?page-id=f89df008-0194-80f6-8001-9a648b96f00b&section=handoff&index=3&share-id=2a21f5ae-60ee-8151-8001-f3985b0d35fe

[Note: You don't need a Penpot username to view the mockups if you don't have one, you can log in using your GitHub account. It should then show you a code inspector for the mockup.]

## Specification

There are four main parts to this dashboard message component:

* **Headline** [required]: Should be a slug. Here you can see it's using an emoji. Ideally this would be one-line, short, to the point: What does the user need to do?
* **Description** [required]: Explain what the user needs to do in more detail, and why, and if applicable provide pointers to relevant areas in the UI. (The underlined "Settings > Resources" is meant to be a link to that screen.)
* **Call-to-Action Button** [optional]: A button that describes - ideally as a verb or verb + object - the action (if any) that the user needs to take with a button to bring them to where they need to do it.
* **Dismiss button** [required]: X in the upper right. Once the user clicks this, the dashboard message should go away and not reappear [unless triggered by a new instance of the action / status / situation that triggered the one that was dismissed.]

mairin:
I am adding this to sprint 247 / planned since it is needed for Onboarding.

nichjones1:
Decided we do not need this.  Closed

PR #4333, opened 2023-10-13
feat: add notification registry (#4125)

lstocchi:
### What does this PR do?

This is the base for #4125 . The notification registry keeps a queue with all notifications that will be displayed into the dashboard.

A notification consists of a title, description and type (for the icon and other ui features).

A notification can be removed either by using the id (click the X icon in the notification car), or by using the extensionId + title ( maybe for the backend to update a state) or by removing all (the clear all button somewhere)

N/A

it is part of #4125 

### How to test this PR?

run tests


packages/extension-api/src/extension-api.d.ts
```diff
@@ -2425,4 +2425,28 @@ declare module '@podman-desktop/api' {
...
+  /**
+   * The context provides write access to the Podman Desktop Notifications system.
+   */
+  export namespace notifications {
+    /**
+     * it pushes a notification that will be displayed in the dashboard/frontend
+     * @param notificationInfo notification to push
+     */
+    export function pushNotification(notificationInfo: NotificationInfo): void;
+  }
```
benoitf: I have mitigated feeling here as we do have the `window.showNotification` method 

so I would expect to reuse `window.showNotification` existing stuff

lstocchi: To me they are two completely different things. `showNotification` pushes an OS notification while this one is something custom to Desktop which will be handled differently (it will have a page where to list all of them, it supports markdown with actions ...).

I could change the name to mark the differentiation but they are not the same imo.

benoitf: I think using OS notification is not what was designed (looks like a quick implementation at the time it was implemented Also it mostly only works on Windows.

In fact it should be mapped to the notifications / tasks area where you can be notified inside Podman Desktop

Then, what is the difference between "Podman needs to resume setup" and "You have finished to build an image" notifications. it's still a notification so you should see "Podman needs to resume setup in the list of notifications using the bell icon" And be able to resume from there.

If I want to display a notification only in the dashboard then it should be a dedicated type in the option but I think it'll confusing it we proceed using two notification systems in parallel.

Also using Operating System or not should be an implementation details. As a extension author you ask for a thing with option, but it does not contains too much specifics

Going back to the option, it should be like a "highlight" option so it means it'll be hightlighted
Then an implementation of the "highlight" will display it inside the dashboard. But later, let say we discover it should be move to another place, it doesn't change the client code)

so I would use the showNotification with a custom option to highlight/promote this notification

We shouldn't provide any `extensionId: string` attribute. the extension id is provided by the caller. (if I call showNotification I have a custom API that knows from which extension I'm calling then I can call the method using the extensionId (but extension ID is not exposed to the public API, it's only used in the local/implementation)

deboer-tim: I'm still struggling on exactly what I think the UI/API should look like, but I was about to post something similar. I think the difference is between just briefly telling you about something that happened (e.g. image built successfully), something that you need to acknowledge/dismiss (e.g. maybe a new version was installed and we tell you about new features) and something that you can take action on or dismiss (e.g. click here to setup podman). I don't think these are fundamentally that different, and having a single API will make it easier for us to decide where/how to display them, or change UI as necessary.

slemeur:
Based on https://github.com/containers/podman-desktop/issues/1230 and https://github.com/containers/podman-desktop/issues/1229, it should also be possible to customize the notification's icon.

deboer-tim:
> Based on #1230 and #1229, it should also be possible to customize the notification's icon.

It's not clear to me if the design is allowing every notification to have a custom icon or just saying we should (e.g.) use a common icon for onboarding, or warnings, etc. Another option is to use the extension's icon, which is what we do for dialogs so that there is some indication/variation but it isn't open. So I'd say we should leave options open and ensure we have enough metadata, but this may not actually need icon support in the registry.

lstocchi:
> Based on #1230 and #1229, it should also be possible to customize the notification's icon.

Do you mean each developer can add a custom icon to any notification? Is there any strong reason for this? In my mind each icon was linked to a different state. So it was up to us to decide which type of notification to support (like now a notification with `info` type was associated with the I icon)

packages/extension-api/src/extension-api.d.ts
lstocchi: Updated to "re-use" the old method. It still accepts the old options object to not break the old code.
Is this what you meant? @benoitf 

So, the bell icon list will eventually be fed by a queue with both tasks and notifications, correct?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -1235,10 +1252,10 @@ declare module '@podman-desktop/api' {
     ): Promise<R>;
 
     /**
-     * Show OS desktop notification
-     * @param options
+     * Show notification on different area of Podman Desktop based on its options (Dashboard, bell icon list, OS notification)
+     * @param options define how the notification must be created.
      */
-    export function showNotification(options: NotificationOptions): Disposable;
+    export function showNotification(options: NotificationCardOptions | NotificationOptions): Disposable;
```
benoitf: hello, I'm wondering if we need another interface or we just enhance NotificationOptions ?

for example missing type can default to 'info' type

benoitf: and only if it's added with higlight that we display it in dashboard as well

lstocchi: the problem is that notificationOptions has all properties optionals, so by keeping them separated you have a check while coding. I expect that if a notification has to be shown in the bell list or in the dashboard it has a title and a description. But to keep it compatible with the old NotificationOptions i have to keep all props optional and it would be ugly to return an error if we find that the highlight is set but it has no title or description, or to not show it in the bell list bc it misses some props, no?

benoitf: if there is no title it can add the extension's name

so we never throw any error, we just display it with defaults. If it looks odd it's the extension's responsibility.

benoitf: I'm +0 on the change. I would say let's keep it simple with only extra attributes but if other reviewers/yourself think it's better to use two interfaces I'm ok.

lstocchi: removed the second interface

packages/extension-api/src/extension-api.d.ts
```diff
@@ -1042,6 +1042,18 @@ declare module '@podman-desktop/api' {
...
      */
     silent?: boolean;
+    /**
+     * The type of the notification. Default value: info
+     */
+    type?: NotificationType;
+    /**
+     * displayed below the description, centered in the notification card. It contains actions (like markdown commands/buttons and links)
+     */
+    actions?: string;
```
benoitf: I think in the API we should not say where it'll be or if it'll be centered, etc as layout may change later

it's not clear to me what's the content of the actions field. Is it markdown content or name of a command that I want to invoke ?
Probably need to use `markdown` prefix as in other place to make sure we can provide markdown content there and maybe rename it to `markdownSomething` (like `markdownDetail`) as actions might look too specific on the purpose, you might just want to display something without any action

packages/extension-api/src/extension-api.d.ts
```diff
@@ -1042,6 +1042,18 @@ declare module '@podman-desktop/api' {
...
+    /**
+     * the notification will be visible in the dashboard
```
```suggestion
     * the notification will be visible in the dashboard
```

I think we might not say it'll go in the dashboard, it's how it's implemented now but it could change
We might say it in a generic way: this notification will be highlighted to the user so it draws attention.

benoitf (commented):
overall looks OK
do you have an extension published using the API (with both highlighted/not highlighted ?)

I'll try to setup a new one tomorrow if there is none to test

lstocchi:
> overall looks OK do you have an extension published using the API (with both highlighted/not highlighted ?)
> 
> I'll try to setup a new one tomorrow if there is none to test

no, i used the podman extension here https://github.com/containers/podman-desktop/pull/4348 but it only has the highlighted option.

packages/extension-api/src/extension-api.d.ts
lstocchi: Idk honestly. Based on Mo's mockup a notification could have a button associated to it, ( look at the third item https://github.com/containers/podman-desktop/issues/1230 )
But i think that it could be a link, or a couple of buttons. This is why i called them actions.

lstocchi: renamed it to markdownActions for the moment. Not sure if we want to make it an array or rename it in another way. markdownDetail does not seem too correct but i don't have any other idea

packages/renderer/src/stores/notifications.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { get } from 'svelte/store';
+import type { Mock } from 'vitest';
+import { beforeAll, expect, test, vi } from 'vitest';
+import { fetchNotifications, notificationEventStore, notificationQueue } from './notifications';
+import type { NotificationCard } from '../../../main/src/plugin/api/notification';
+
+// first, path window object
```
benoitf:
```suggestion
// first, patch window object
```

benoitf (approved):
nice enhancement 👍

PR #4334, opened 2023-10-13
fix: do not register the same registry twice

### LINKED ISSUE #3908: Every re-enablement of podman extension adds copy of registry record that was added manually
### Bug description

When I have a custom registry added to a registries in podman desktop, and I disable and enable podman extension, the registry record is copied and I have two same record. It behaves as one record -> deleting one will delete all redundant records. 

### Operating system

Windows 10

### Installation Method

### Version

1.4.0

### Steps to reproduce

1. Have podman installed (might not be necessary)
2. Insert new registry
3. Extensions -> Podman - disable, enable
4. Check registries

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

This PR checks that a Registry is not already registered by a source before to register it.

This is useful as the podman extension (at least) re-registers registries every time it is enabled again.

Fixes #3908 

### How to test this PR?

- Have podman installed (might not be necessary)
- Insert new registry
- Extensions -> Podman - disable, enable
- Check registries: they should not be duplicated

### DIFF
--- a/packages/main/src/plugin/image-registry.ts
+++ b/packages/main/src/plugin/image-registry.ts
@@ -142,6 +142,15 @@ export class ImageRegistry {
   }
 
   registerRegistry(registry: containerDesktopAPI.Registry): Disposable {
+    const found = this.registries.find(
+      reg =>
+        reg.source === registry.source && reg.serverUrl === registry.serverUrl && reg.username === registry.username,
+    );
+    if (found) {
+      // Ignore and don't register - extension may register registries every time it is restarted
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      return Disposable.create(() => {});
+    }
     this.registries = [...this.registries, registry];
     this.telemetryService.track('registerRegistry', {
       serverUrl: this.getRegistryHash(registry),
--- a/packages/main/src/plugin/image-registry.ts
+++ b/packages/main/src/plugin/image-registry.ts
@@ -148,8 +148,8 @@ export class ImageRegistry {
     );
     if (found) {
       // Ignore and don't register - extension may register registries every time it is restarted
-      // eslint-disable-next-line @typescript-eslint/no-empty-function
-      return Disposable.create(() => {});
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/image-registry.ts
```diff
@@ -142,6 +142,15 @@ export class ImageRegistry {
...
 
   registerRegistry(registry: containerDesktopAPI.Registry): Disposable {
+    const found = this.registries.find(
+      reg =>
+        reg.source === registry.source && reg.serverUrl === registry.serverUrl && reg.username === registry.username,
+    );
+    if (found) {
+      // Ignore and don't register - extension may register registries every time it is restarted
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      return Disposable.create(() => {});
```
benoitf: maybe use Disposable.from(); so you don't need to have a no-empty linter rule
(or add a noop/empty() in Disposable)

packages/main/src/plugin/image-registry.ts
```diff
@@ -142,6 +142,15 @@ export class ImageRegistry {
   }
 
   registerRegistry(registry: containerDesktopAPI.Registry): Disposable {
+    const found = this.registries.find(
+      reg =>
+        reg.source === registry.source && reg.serverUrl === registry.serverUrl && reg.username === registry.username,
+    );
+    if (found) {
+      // Ignore and don't register - extension may register registries every time it is restarted
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
```
benoitf: it should probably log an entry in console ?

```ts
console.log('Registry already registered, skipping registration')
```

benoitf (approved):

benoitf (approved):
@feloy if you can add a unit test to ensure that we receive a disposable and not an error when  we register twice a registry calling registerRegistry method

benoitf (approved):

PR #4335, opened 2023-10-13
docs: using tabs for the proxy section 

### LINKED ISSUE #4297: Use OS tabs for proxies documentation

The proxies documentation has Mac and Windows on one page and Linux on another one, and they aren't sorted near each other in the nav.

FYI @themr0c

Put the content of these two pages directly in the 'Using behind a proxy' page, using OS tabs like the proposed Installation section in #4280. This would put the relevant information right on the main page and reduce the amount of navigation.

_No response_

### Additional context

_No response_

themr0c:
### What does this PR do?

- Using [tabs with query strings](https://docusaurus.io/docs/markdown-features/tabs#query-string) for the proxy section.
- Outcome: simplifified navigation: everything in two page (to avoid nested tabs).
- Integrated optional steps as [details](https://docusaurus.io/docs/markdown-features#details) into the procedure flow.
- Adapted the procedures steps for consistency.
- Added screenshot.
- Using [reusable snippets](https://docusaurus.io/docs/markdown-features/react#importing-markdown)

image

fixes #4297


deboer-tim:
The intention of #4297 was that we shouldn't have multiple pages in the nav when the same topic applies to different OSes, so the user doesn't have to navigate to the correct one or see info from the other platforms. To me this PR feels like trying to stuff all proxy related topics (which was 6 pages before) into one.

I also don't love multiple rows of tabs. We're using the top row b/c it hides the other platforms by default and the user doesn't really have to think about it, but now there's a second row that they do need to look at and understand to find all the relevant topics.

IMHO we should keep separate pages for each topic and only use tabs for platform filtering.

themr0c:
@deboer-tim so you would be happy with 2 pages?
- behind a proxy 
- behind a VPN (Windows only AFAIK)

themr0c:
@deboer-tim What do you think about adding foldable optional steps? Does it enhance clarity and navigation?

themr0c:
@deboer-tim Updated with two pages to avoid nested tabs. WDYT?

PR #3679, opened 2023-08-25
chore: fix svelte checks

benoitf:
### What does this PR do?

Turn on the strict mode in tsconfig.json
650errors were reported, trying to target all of them

### What issues does this PR fix or reference?

Fixes https://github.com/containers/podman-desktop/issues/3665

### How to test this PR?

Test lot of use-cases, pages

PR #4336, opened 2023-10-13
fix: breadcrumb history

### LINKED ISSUE #4288: Exiting the "Pull Image from Registry" screen brings me to dashboard instead of Images screen
### Bug description

I had a confusing moment and then noticed if you click the close button on the "Pull Image From a Registry" screen, you are taken to the Dashboard screen instead of back to the Images screen.

### Operating system

RHEL 9.4 installed via tarball

### Installation Method

Other

### Version

1.4.0

### Steps to reproduce

1. Click on the images tab
2. Click on "Pull Image"
3. Click "X" in the upper right just above the manage registries button
4. Find yourself on the Dashboard screen instead of the Images screen

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Breadcrumb history is going back to Home incorrectly in some cases because after PR #3679 the detailsPage does not get cleared.

Instead of just clearing the details page, I'm replacing list page + details page with a history array. This is more generic code, a more common pattern, more easily extendable if our needs change, and less likely to be broken by other changes. I probably should have done this originally, but at that point I was trying to keep it simple and didn't foresee us using history in Settings.

While I was at it I moved mockBreadcrumb() out of the .ts as promised a while ago in https://github.com/containers/podman-desktop/pull/3388#pullrequestreview-1558267032

N/A

Fixes #4288.

### How to test this PR?

Check that the breadcrumb correctly lists the previous page and works when going to the lists, details pages, forms from either, and in Settings > Resources for both machine creation and machine details.

### DIFF
--- a/packages/renderer/src/Route.svelte
+++ b/packages/renderer/src/Route.svelte
@@ -2,7 +2,7 @@
 import { createRouteObject } from 'tinro/dist/tinro_lib';
 import type { TinroBreadcrumb, TinroRouteMeta } from 'tinro';
 import { TelemetryService } from './TelemetryService';
-import { lastPage, currentPage, listPage, detailsPage } from './stores/breadcrumb';
+import { lastPage, currentPage, history } from './stores/breadcrumb';
 import type { NavigationHint } from './Route';
 import { onDestroy } from 'svelte';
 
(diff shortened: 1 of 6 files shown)


packages/renderer/src/Route.svelte
```diff
@@ -38,23 +38,23 @@ function processMetaBreadcrumbs(breadcrumbs?: Array<TinroBreadcrumb>) {
     if (!curPage) return;
 
     if (navigationHint === 'root') {
-      listPage.set(curPage);
-      detailsPage.set({ name: 'Home', path: '/' });
+      history.set([curPage]);
     } else if (navigationHint === 'details') {
-      detailsPage.set(curPage);
-      lastPage.set($listPage);
+      history.set([$history[0], curPage]);
```
jeffmaury, line 43: Out of curiosity what is this $history([0] notation for ?

jeffmaury (approved):
LGTM

packages/renderer/src/Route.svelte
deboer-tim: history is an array which could be [list] or [list, some page], and we want to make sure it is now [list, current page]. So this is just setting it to two elements, the first is whatever was already there.

PR #4342, opened 2023-10-14
fix: progressImpl properly middleware tasks to set the proper result status

### LINKED ISSUE #4341: ProgressImpl: impossible to fail a task
### Bug description

The following lines indicates, that the task created with the `withWidget` can never reach the `catch` handle since an await is located on the `then` handle. Thus preventing from failing a task properly.

https://github.com/containers/podman-desktop/blob/d0cafacb0c92300897b03a6e97586c8fcb6b71c4/packages/main/src/plugin/progress-impl.ts#L94-L105

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

Create a simple progress task and try to fail it.

```
progressImpl.withProgress<void>(
  { location: ProgressLocation.TASK_WIDGET, title: `Pulling ${imageName}.` },
  async  progress => {
    throw new Error('error');
});
```

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Fixes #4341

### DIFF
--- a/packages/main/src/plugin/progress-impl.ts
+++ b/packages/main/src/plugin/progress-impl.ts
@@ -35,6 +35,13 @@ export enum ProgressLocation {
 export class ProgressImpl {
   constructor(private taskManager: TaskManager) {}
 
+  /**
+   * Execute a task with progress, based on the provided options and task function.
+   * @template R - The type of the result of the task.
+   * @param {extensionApi.ProgressOptions} options - The options for the progress.
+   * @param {Function} task - The task function to be executed with progress.
+   * @returns {Promise<R>} - A promise that resolves to the result of the task.
+   */
   withProgress<R>(
     options: extensionApi.ProgressOptions,
     task: (
@@ -94,16 +101,21 @@ export class ProgressImpl {
         new CancellationTokenImpl(),
       )
         .then(value => {
+          // Middleware to capture the success of the task
           t.status = 'success';
           t.state = 'completed';
+          // We propagate the result to the caller, so he can use the result
           resolve(value);
         })
         .catch((err: unknown) => {
+          // Middleware to set to error the task
           t.status = 'failure';
           t.state = 'completed';
+          // We propagate the error to the caller, so it can handle it if needed
           reject(err);
         })
         .finally(() => {
+          // Ensure the taskManager is updated properly is every case
           this.taskManager.updateTask(t);
         });
     });
--- a/packages/main/src/plugin/progress-impl.ts
+++ b/packages/main/src/plugin/progress-impl.ts
@@ -85,39 +85,37 @@ export class ProgressImpl {
   ): Promise<R> {
     const t = this.taskManager.createTask(options.title);
 
-    return new Promise<R>((resolve, reject) => {
-      task(
-        {
-          report: value => {
-            if (value.message) {
-              t.error = value.message;
-            }
-            if (value.increment) {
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/progress-impl.ts
```diff
@@ -91,18 +91,22 @@ export class ProgressImpl {
...
 
-    task_.catch((err: unknown) => {
-      t.status = 'failure';
-      t.state = 'completed';
-      t.error = String(err);
-      this.taskManager.updateTask(t);
-    });
+    await Promise.any([
+      task_.then(async () => {
+        console.log('progressImpl THEN');
```
benoitf: looks like you forgot to remove traces

packages/main/src/plugin/progress-impl.ts
```diff
@@ -91,18 +91,22 @@ export class ProgressImpl {
...
+        t.status = 'success';
+        t.state = 'completed';
+        this.taskManager.updateTask(t);
+      }),
+      task_.catch((err: unknown) => {
+        console.log('progressImpl on failure');
```
benoitf: trace to remove

benoitf (commented):
thanks @axel7083 could you add a unit test to ensure that the current code is failing and with the updated code we're reaching the expected behaviour ?

jeffmaury (commented):
Looks barely readable to me.
Could be rewritten either through method chaining: task.`then(...).catch(....)` or as:

```
try {
  await task_;
  t.status = 'success';
} catch (err) {
  t.status = 'failure';
  t.error = String(err);
} finally {
  t.state = 'completed';
  this.taskManager.updateTask(t);
}
```

axel7083:
> Looks barely readable to me. Could be rewritten either through method chaining: task.`then(...).catch(....)` or as:
> 
> ```
> try {
>   await task_;
>   t.status = 'success';
> } catch (err) {
>   t.status = 'failure';
>   t.error = String(err);
> } finally {
>   t.state = 'completed';
>   this.taskManager.updateTask(t);
> }
> ```

The return type of `async withWidget<R>(...)` is `Promise<R>`, if we wait for the completion of the task don't we prevent caller from handling the exception of the task ?

I added a unit test to ensure two important element
- The error is not handled by the withWidged, it is propagated to the caller
- The state of the task is set to **completed** and status to **failure**

I changed the implementation to return a new Promise<R>

benoitf:
@jeffmaury could you check again ?

packages/main/src/plugin/progress-impl.ts
```diff
@@ -77,32 +77,35 @@ export class ProgressImpl {
...
-      t.state = 'completed';
-      this.taskManager.updateTask(t);
-    });
 
-    task_.catch((err: unknown) => {
-      t.status = 'failure';
-      t.state = 'completed';
-      t.error = String(err);
-      this.taskManager.updateTask(t);
+    return new Promise<R>((resolve, reject) => {
```
jeffmaury: As `then`, `catch` and `finally` methods allow methods chaining, I think there is an extra Promise object created that is not required.

axel7083: I remove the encapsulating promise ! Thanks for noticing

axel7083:
CC @jeffmaury could you take a look ?

jeffmaury (approved):
LGTM

PR #4348, opened 2023-10-16
feat: add notifications box in the dashboard (#4125)

### LINKED ISSUE #4125: Onboarding: Add notification about available onboarding flows to dashboard

lstocchi:
### What does this PR do?

This PR is based on #4333 . If you want to look the code related to this PR only just pick the last commit. Otherwise i'll rebase it when #4333 is merged. This can be used for testing.

It adds the notificationsBox component which displays the latest 3 notifications in the dashboard.

notification_dashboard_2

it resolves #4125 

### How to test this PR?

1. have podman not installed
2. launch desktop

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -78,7 +78,7 @@ const podmanInfoHelper = new PodmanInfoHelper();
 
 let shouldNotifySetup = true;
 const setupPodmanNotification: extensionApi.NotificationOptions = {
-  title: 'Podman needs to be set up!',
+  title: 'Podman needs to be set up',
   body: 'The Podman extension is installed, yet requires configuration. Some features might not function optimally.',
   type: 'info',
   markdownActions: ':button[Set up]{href=/preferences/onboarding/podman-desktop.podman title="Set up Podman"}',
@@ -542,6 +542,9 @@ async function monitorProvider(provider: extensionApi.Provider) {
       if (!installedPodman) {
         provider.updateStatus('not-installed');
         extensionApi.context.setValue('podmanIsNotInstalled', true, 'onboarding');
+        // if podman is not installed and the OS is linux we show the podman onboarding notification (if it has not been shown earlier)
+        // this should be limited to Linux as in other OSes the onboarding workflow is enabled based on the podman machine existance
+        // and the notification is handled by checking the machine
         if (isLinux() && shouldNotifySetup) {
           // push setup notification
           extensionApi.window.showNotification(setupPodmanNotification);
@@ -554,6 +557,7 @@ async function monitorProvider(provider: extensionApi.Provider) {
           provider.updateStatus('installed');
         }
         extensionApi.context.setValue('podmanIsNotInstalled', false, 'onboarding');
+        // if podman has been installed, we reset the notification flag so if podman is uninstalled in future we can show the notification again
         if (isLinux()) {
           shouldNotifySetup = true;
         }
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -81,7 +81,7 @@ const setupPodmanNotification: extensionApi.NotificationOptions = {
   title: 'Podman needs to be set up!',
   body: 'The Podman extension is installed but needs to be set up. Some features may not work as expected.',
   type: 'info',
-  actions: ':button[Set up]{href=/preferences/onboarding/podman-desktop.podman title="Set up Podman"}',
+  markdownActions: ':button[Set up]{href=/preferences/onboarding/podman-desktop.podman title="Set up Podman"}',
   highlight: true,
   silent: true,
 };
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -79,7 +79,7 @@ const podmanInfoHelper = new PodmanInfoHelper();
 let shouldNotifySetup = true;
 const setupPodmanNotification: extensionApi.NotificationOptions = {
   title: 'Podman needs to be set up!',
-  body: 'The Podman extension is installed but needs to be set up. Some features may not work as expected.',
+  body: 'The Podman extension is installed, yet requires configuration. Some features might not function optimally.',
   type: 'info',
   markdownActions: ':button[Set up]{href=/preferences/onboarding/podman-desktop.podman title="Set up Podman"}',
   highlight: true,
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -76,6 +76,16 @@ const qemuHelper = new QemuHelper();
 const podmanBinaryHelper = new PodmanBinaryLocationHelper();
 const podmanInfoHelper = new PodmanInfoHelper();
 
+let shouldNotifySetup = true;
+const setupPodmanNotification: extensionApi.NotificationOptions = {
+  title: 'Podman needs to be set up!',
+  body: 'The Podman extension is installed but needs to be set up. Some features may not work as expected.',
+  type: 'info',
+  actions: ':button[Set up]{href=/preferences/onboarding/podman-desktop.podman title="Set up Podman"}',
+  highlight: true,
+  silent: true,
+};
+
 export type MachineJSON = {
   Name: string;
   CPUs: number;
@@ -108,11 +118,26 @@ export type MachineInfo = {
 
 async function updateMachines(provider: extensionApi.Provider): Promise<void> {
   // init machines available
-  const machineListOutput = await getJSONMachineList();
+  let machineListOutput: string;
+  try {
+    machineListOutput = await getJSONMachineList();
+  } catch (error) {
+    if (shouldNotifySetup) {
+      // push setup notification
+      extensionApi.window.showNotification(setupPodmanNotification);
+      shouldNotifySetup = false;
+    }
+    throw error;
+  }
 
   // parse output
   const machines = JSON.parse(machineListOutput) as MachineJSON[];
   extensionApi.context.setValue('podmanMachineExists', machines.length > 0, 'onboarding');
+  if (shouldNotifySetup && machines.length === 0) {
+    // push setup notification
+    extensionApi.window.showNotification(setupPodmanNotification);
+    shouldNotifySetup = false;
+  }
 
   // update status of existing machines
   for (const machine of machines) {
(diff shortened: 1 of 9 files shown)


lstocchi:
@mairin by looking at your new dahsboard mockup it looks to me you use different charcoal colors compared to what we have right now - charcoal-700 for the background and charcoal-800 for the cards background. 
In this PR I kept using the same old palette but I can change all of them in a separate PR if you confirm we need to use lighter colors.

extensions/podman/src/extension.ts
```diff
@@ -76,6 +76,15 @@ const qemuHelper = new QemuHelper();
 const podmanBinaryHelper = new PodmanBinaryLocationHelper();
 const podmanInfoHelper = new PodmanInfoHelper();
 
+let shouldNotifySetup = true;
+const setupPodmanNotification: extensionApi.NotificationInfo = {
+  extensionId: 'podman-desktop.podman',
+  title: 'Podman needs to be setup!',
+  description: 'The Podman extension is installed but needs to be setup. Some features may not work as expected.',
```
mairin: This is so pedantic 😅 but I would do

s/setup/set up - in both the title and the description. It just comes across more clearly (I think usually "setup" is treated as a noun whereas "set up" is a verb / verbal phrase.)

extensions/podman/src/extension.ts
```diff
@@ -76,6 +76,15 @@ const qemuHelper = new QemuHelper();
...
+  type: 'info',
+  actions: ':button[Setup]{href=/preferences/onboarding/podman-desktop.podman title="Setup Podman"}',
```
mairin: Another s/Setup/Set up

benoitf:
I'm wondering if we need to display something when it's empty : we have here "New notifications widget displaying no notification to display"

packages/renderer/src/lib/dashboard/NotificationCard.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+import '@testing-library/jest-dom/vitest';
+import { test, expect, vi, beforeAll } from 'vitest';
+import { fireEvent, render, screen } from '@testing-library/svelte';
+import type { Notification } from '../../../../main/src/plugin/api/notification';
+import NotificationCard from './NotificationCard.svelte';
+
+const removeNotificatioMock = vi.fn();
+
+beforeAll(() => {
+  (window as any).removeNotification = removeNotificatioMock;
```
mairin: Is there a typo here in removeNotificatioMock => should it be removeNotificatio**n**Mock?

packages/renderer/src/lib/dashboard/NotificationCard.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  expect(descriptionDiv).toBeInTheDocument();
+  expect(titleDiv.textContent).toEqual('title');
+  expect(descriptionDiv.textContent).toContain('description');
+
+  const deleteButton = screen.getByRole('button', { name: 'Delete notification 1' });
+  expect(deleteButton).toBeInTheDocument();
+
+  await fireEvent.click(deleteButton);
+
+  expect(removeNotificatioMock).toBeCalled();
```
mairin: this does not have the "n" in "Notificatio" either so it's intentional?

packages/renderer/src/lib/dashboard/NotificationsBox.svelte
```diff
@@ -0,0 +1,35 @@
...
+    notificationsUnsubscribe();
+  }
+});
+</script>
+
+<div class="bg-charcoal-800 m-5 px-5 py-4 rounded-lg">
+  <div class="flex flex-col items-center justify-content space-y-3">
+    <span class="text-lg text-white mb-1"> New Notifications </span>
+    {#if notifications.length === 0}
+      <div class="pt-3 pb-5" aria-label="No notification to display">No notification to display</div>
```
mairin: I would suggest the following classes on the no notification to display message:

text-gray-700
text-sm

packages/renderer/src/lib/dashboard/NotificationsBox.svelte
```diff
@@ -0,0 +1,35 @@
...
+onDestroy(() => {
+  if (notificationsUnsubscribe) {
+    notificationsUnsubscribe();
+  }
+});
+</script>
+
+<div class="bg-charcoal-800 m-5 px-5 py-4 rounded-lg">
+  <div class="flex flex-col items-center justify-content space-y-3">
+    <span class="text-lg text-white mb-1"> New Notifications </span>
```
mairin: I would suggest scaling down to text-md here.

lstocchi:
> I'm wondering if we need to display something when it's empty : we have here "New notifications widget displaying no notification to display"

The idea that Mo had was to show an empty message and add a link to an archive page (when we implement it).

mairin (commented):
This looks great! I put in a few comments of small things I noticed and some tweaks to font size & color.

You did a great job picking colors here - I apologize for the mockups coloring being wrong, they were based on the old UI coloring 🤦‍♀️

mairin:
> > I'm wondering if we need to display something when it's empty : we have here "New notifications widget displaying no notification to display"
> 
> The idea that Mo had was to show an empty message and add a link to an archive page (when we implement it).

@benoitf is the issue here that the title says "New Notifications" even though there is nothing? Maybe better to have the card title be, "Notifications" ?

benoitf:
> @benoitf is the issue here that the title says "New Notifications" even though there is nothing? Maybe better to have the card title be, "Notifications" ?

I would say both. Why display something if it's to say "nothing to be displayed". Just do not display it ! :-)

and yes, seeing "New notifications" while there is nothing is also confusing so Notifications might be better.

I also commented elsewhere that these notifications should be part of the global notification area (bell icon) so if I click on the bell icon I can see the notification and resume it from there as well.
It's just that on the dashboard it will filter out items

mairin:
> I would say both. Why display something if it's to say "nothing to be displayed". Just do not display it ! :-)

Ah yeh. I would say let's do that, except it would be nice to have a place to get a link to all the past notifications (in the case where you have a notification, you click dismiss and then immediately have remorse lol. it would be easy to click the "view past notifications" link there - when available.)

i don't know i don't have a strong opinion on this, so i would be ok if it disappears totally for now if there are 0 messages, and if we want to bring back an empty list message later when we have an archive link, that might work?

> and yes, seeing "New notifications" while there is nothing is also confusing so Notifications might be better.

+1

> I also commented elsewhere that these notifications should be part of the global notification area (bell icon) so if I click on the bell icon I can see the notification and resume it from there as well. It's just that on the dashboard it will filter out items

+1 agreed

lstocchi:
Updated the PR with Mo's suggestions and i made the notifications box hidden if there are no notifications. Updated the gif in the first post

mairin (approved):
Looks great!!!

lstocchi:
Rebased as #4125 has been merged. Ready to be reviewed

extensions/podman/src/extension.ts
```diff
@@ -76,6 +76,16 @@ const qemuHelper = new QemuHelper();
 const podmanBinaryHelper = new PodmanBinaryLocationHelper();
 const podmanInfoHelper = new PodmanInfoHelper();
 
+let shouldNotifySetup = true;
+const setupPodmanNotification: extensionApi.NotificationOptions = {
+  title: 'Podman needs to be set up!',
```
cdrage:
```suggestion
  title: 'Podman needs to be set up',
```

Such a nitpick, but should we have exclamations / shouting in important notifications like this? WDYT @mairin

extensions/podman/src/extension.ts
```diff
@@ -76,6 +76,16 @@ const qemuHelper = new QemuHelper();
...
+  body: 'The Podman extension is installed but needs to be set up. Some features may not work as expected.',
```
cdrage:
```suggestion
  body: 'The Podman extension is installed, yet requires configuration. Some features might not function optimally.',
```

May be better?

extensions/podman/src/extension.ts
```diff
@@ -108,11 +118,26 @@ export type MachineInfo = {
 
 async function updateMachines(provider: extensionApi.Provider): Promise<void> {
   // init machines available
-  const machineListOutput = await getJSONMachineList();
+  let machineListOutput: string;
+  try {
+    machineListOutput = await getJSONMachineList();
```
cdrage, line 123: I find it weird that we do a try / catch around this?

Shouldn't we keep it like:
```ts
const machineListOutput = await getJSONMachineList();
if (machineListOutput === undefined) {
    if (shouldNotifySetup) {
      // push setup notification
      extensionApi.window.showNotification(setupPodmanNotification);
      shouldNotifySetup = false;
    }
}

?

I find it odd we also throw an error but it doesn't appear anywhere?

Comment (cdrage) [reply extensions/podman/src/extension.ts:123]: try / catch also prevents the rest of the activate() to proceed?

File: extensions/podman/src/extension.ts [hunk 10]
```diff
@@ -517,13 +542,21 @@ async function monitorProvider(provider: extensionApi.Provider) {
       if (!installedPodman) {
         provider.updateStatus('not-installed');
         extensionApi.context.setValue('podmanIsNotInstalled', true, 'onboarding');
+        if (isLinux() && shouldNotifySetup) {
```
cdrage, line 548: imho i think we should have comments here regarding why we are not notifying on linux. it's a bit ambiguous on here.

cdrage (changes requested):

extensions/podman/src/extension.ts
lstocchi: the problem is that if you don't put the `getJSONMachineList` inside a try/catch and it fails for some reason (like if it tries to fetch machines when podman is not installed) you never display the notification. This way if there is an error, it checks if the notification needs to be shown and throw the error as it would have done before.

extensions/podman/src/extension.ts
lstocchi: onboarding for podman works differently based on the OS.
On linux we just check if podman is installed, on mac/win if there is atleast a machine. This is the reason why this part is dedicated to linux.
I'm gonna add a comment to explain it

extensions/podman/src/extension.ts
lstocchi: yeah, Thanks!!

lstocchi:
@cdrage should be ready for another review

cdrage (approved):

PR #4356, opened 2023-10-16
feat: retrieve proxy settings from system

### LINKED ISSUE #3769: Retrieve proxy configuration from the system

The proxy configuration has to be manually configured. It may be retrieved from the system (Windows only ?). That may prevent errors reported by users.

On Windows, the Network configuration can be retrieved bu running the following command:

`reg query "HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Internet Settings"`

The output is the following:

```
HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Internet Settings
    CertificateRevocation    REG_DWORD    0x1
    DisableCachingOfSSLPages    REG_DWORD    0x0
    IE5_UA_Backup_Flag    REG_SZ    5.0
    PrivacyAdvanced    REG_DWORD    0x1
    SecureProtocols    REG_DWORD    0x800
    User Agent    REG_SZ    Mozilla/4.0 (compatible; MSIE 8.0; Win32)
    ZonesSecurityUpgrade    REG_BINARY    A3AACCDDB7CCD901
    WarnonZoneCrossing    REG_DWORD    0x0
    EnableNegotiate    REG_DWORD    0x1
    ProxyEnable    REG_DWORD    0x1
    MigrateProxy    REG_DWORD    0x1
    SecureProtocolsUpdated    REG_DWORD    0x1
    LockDatabase    REG_QWORD    0x1d9cc6d40de1d54
    ProxyHttp1.1    REG_DWORD    0x1
    EnableHttp1_1    REG_DWORD    0x1
    ProxyServer    REG_SZ    http=127.0.0.1:8888;https=127.0.0.1:8888
    ProxyOverride    REG_SZ    <-loopback>
    _FiddlerDirectWrite    REG_SZ    BuggyWinShutdown at 12:23:12 2023-09-05
```

Retrieve the configuration from the system

_No response_

### Additional context

_No response_

### LINKED ISSUE #4765: Proxy environment code prepends "http://"
### Bug description

Ad noticed by @win5do in https://github.com/containers/podman-desktop/issues/4549#issuecomment-1806010216

```
podman machine ssh
env | grep -i proxy

ouput:
http_proxy=http://host.containers.internal:8080
https_proxy=http://host.containers.internal:8080
HTTP_PROXY=http://http://host.containers.internal:8080
HTTPS_PROXY=http://http://host.containers.internal:8080

```

The proxy env code is prepending "http://" here:
https://github.com/containers/podman-desktop/blob/2d2e5e8102d7141a2fc95ce85320478d0e3b0cb6/packages/main/src/plugin/util/exec.ts#L59

It should not be hardcoded to "http" since proxy servers may use other schemes (e.g. https). It might be ok to assume and append it if not already present, but this should be done earlier so it is consistent with the other code path which sets containers.conf (source of the lower case values). Note It might also be worth dropping the container.conf write path if this is now being set at the env point. 

### Operating system

Windows

### Installation Method

None

### Version

1.5.0

jeffmaury:
Fixes #3769
Fixes #4765

### What does this PR do?

Get the proxy settings from the system. On Windows it uses the registry. On MacOS networksetup tool is used. On Linux HTTP_PROXY, HTTPS_PROXY and NO_PROXY are used. There are no 3 options for proxy: system (default), manual and disabled.

#3769 
#4765

### How to test this PR?

Set proxy settings on your system and verify there are use if system is used inside Podman Desktop

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -105,7 +105,7 @@ import type { WebviewInfo } from '/@api/webview-info.js';
 
 import { securityRestrictionCurrentHandler } from '../security-restrictions-handler.js';
 import type { TrayMenu } from '../tray-menu.js';
-import { isMac } from '../util.js';
+import { isLinux, isMac } from '../util.js';
 import type { ApiSenderType } from './api.js';
 import type { PodInfo, PodInspectInfo } from './api/pod-info.js';
 import { AppearanceInit } from './appearance-init.js';
@@ -1920,16 +1920,16 @@ export class PluginSystem {
       },
     );
 
-    this.ipcHandle('proxy:setState', async (_listener: Electron.IpcMainInvokeEvent, state: boolean): Promise<void> => {
-      return proxy.setState(state);
+    this.ipcHandle('proxy:setState', async (_listener: Electron.IpcMainInvokeEvent, state: number): Promise<void> => {
+      return proxy.setState(state as ProxyState);
     });
 
     this.ipcHandle('proxy:getSettings', async (): Promise<containerDesktopAPI.ProxySettings | undefined> => {
       return proxy.proxy;
     });
 
-    this.ipcHandle('proxy:isEnabled', async (): Promise<boolean> => {
-      return proxy.isEnabled();
+    this.ipcHandle('proxy:getState', async (): Promise<number> => {
+      return proxy.getState();
     });
 
     this.ipcHandle(
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -105,7 +105,7 @@ import type { WebviewInfo } from '/@api/webview-info.js';
 
 import { securityRestrictionCurrentHandler } from '../security-restrictions-handler.js';
 import type { TrayMenu } from '../tray-menu.js';
-import { isLinux, isMac } from '../util.js';
+import { isMac } from '../util.js';
 import type { ApiSenderType } from './api.js';
 import type { PodInfo, PodInspectInfo } from './api/pod-info.js';
 import { AppearanceInit } from './appearance-init.js';
@@ -157,7 +157,7 @@ import { NavigationItemsInit } from './navigation-items-init.js';
 import { OnboardingRegistry } from './onboarding-registry.js';
 import { OpenDevToolsInit } from './open-devtools-init.js';
 import { ProviderRegistry } from './provider-registry.js';
-import { Proxy } from './proxy.js';
+import { Proxy, type ProxyState } from './proxy.js';
 import { RecommendationsRegistry } from './recommendations/recommendations-registry.js';
 import { SafeStorageRegistry } from './safe-storage/safe-storage-registry.js';
 import type { StatusBarEntryDescriptor } from './statusbar/statusbar-registry.js';
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -1920,9 +1920,12 @@ export class PluginSystem {
       },
     );
 
-    this.ipcHandle('proxy:setState', async (_listener: Electron.IpcMainInvokeEvent, state: ProxyState): Promise<void> => {
-      return proxy.setState(state as ProxyState);
-    });
+    this.ipcHandle(
+      'proxy:setState',
+      async (_listener: Electron.IpcMainInvokeEvent, state: ProxyState): Promise<void> => {
+        return proxy.setState(state);
+      },
+    );
 
     this.ipcHandle('proxy:getSettings', async (): Promise<containerDesktopAPI.ProxySettings | undefined> => {
       return proxy.proxy;
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -98,6 +98,7 @@ import type {
   ProviderInfo,
   ProviderKubernetesConnectionInfo,
 } from '/@api/provider-info.js';
+import type { ProxyState } from '/@api/proxy.js';
 import type { PullEvent } from '/@api/pull-event.js';
 import type { ViewInfoUI } from '/@api/view-info.js';
 import type { VolumeInspectInfo, VolumeListInfo } from '/@api/volume-info.js';
@@ -157,7 +158,7 @@ import { NavigationItemsInit } from './navigation-items-init.js';
 import { OnboardingRegistry } from './onboarding-registry.js';
 import { OpenDevToolsInit } from './open-devtools-init.js';
 import { ProviderRegistry } from './provider-registry.js';
-import { Proxy, type ProxyState } from './proxy.js';
+import { Proxy } from './proxy.js';
 import { RecommendationsRegistry } from './recommendations/recommendations-registry.js';
 import { SafeStorageRegistry } from './safe-storage/safe-storage-registry.js';
 import type { StatusBarEntryDescriptor } from './statusbar/statusbar-registry.js';
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -1920,7 +1920,7 @@ export class PluginSystem {
       },
     );
 
-    this.ipcHandle('proxy:setState', async (_listener: Electron.IpcMainInvokeEvent, state: number): Promise<void> => {
+    this.ipcHandle('proxy:setState', async (_listener: Electron.IpcMainInvokeEvent, state: ProxyState): Promise<void> => {
       return proxy.setState(state as ProxyState);
     });
 
@@ -1928,7 +1928,7 @@ export class PluginSystem {
       return proxy.proxy;
     });
 
-    this.ipcHandle('proxy:getState', async (): Promise<number> => {
+    this.ipcHandle('proxy:getState', async (): Promise<ProxyState> => {
       return proxy.getState();
     });
 
--- a/packages/main/src/plugin/proxy-system.spec.ts
+++ b/packages/main/src/plugin/proxy-system.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/main/src/plugin/proxy-system.spec.ts
+++ b/packages/main/src/plugin/proxy-system.spec.ts
@@ -57,14 +57,8 @@ describe('Windows platform tests', () => {
         cb(new Error('execution error'), undefined);
       }),
     });
-    let error;
-    try {
-      await getProxySettingsFromSystem({} as Proxy);
-    } catch (err) {
-      error = err;
-    }
-    expect(error).toBeDefined();
-    expect(error).toBeInstanceOf(Error);
+    const settings = await getProxySettingsFromSystem({} as Proxy);
+    expect(settings).toBeUndefined();
   });
 
   test('No state returned in case of proxy disabled', async () => {
@@ -75,10 +69,7 @@ describe('Windows platform tests', () => {
       }),
     });
     const settings = await getProxySettingsFromSystem({} as Proxy);
-    expect(settings).toBeDefined();
-    expect(settings.httpProxy).toBeUndefined();
-    expect(settings.httpsProxy).toBeUndefined();
-    expect(settings.noProxy).toBeUndefined();
+    expect(settings).toBeUndefined();
   });
 
   test('Empty state returned in case of proxy enabled only', async () => {
@@ -142,10 +133,7 @@ describe('Linux platform test', () => {
       delete process.env['HTTPS_PROXY'];
       delete process.env['NO_PROXY'];
       const settings = await getProxySettingsFromSystem({} as Proxy);
-      expect(settings).toBeDefined();
-      expect(settings.httpProxy).toBeUndefined();
-      expect(settings.httpsProxy).toBeUndefined();
-      expect(settings.noProxy).toBeUndefined();
+      expect(settings).toBeUndefined();
     } finally {
       process.env['HTTP_PROXY'] = previousHttpProxy;
       process.env['HTTPS_PROXY'] = previousHttpsProxy;
@@ -221,34 +209,22 @@ describe('MacOS platform tests', () => {
   test('No state returned in case of execution error', async () => {
     setupPlatform(Platform.MACOS);
     vi.spyOn(Exec.prototype, 'exec').mockRejectedValue(new Error('execution error'));
-    let error;
-    try {
-      await getProxySettingsFromSystem({} as Proxy);
-    } catch (err) {
-      error = err;
-    }
-    expect(error).toBeDefined();
-    expect(error).toBeInstanceOf(Error);
+    const settings = await getProxySettingsFromSystem({} as Proxy);
+    expect(settings).toBeUndefined();
   });
 
   test('No state returned if no network connections', async () => {
     setupPlatform(Platform.MACOS);
     vi.spyOn(Exec.prototype, 'exec').mockResolvedValue({ stdout: '' } as RunResult);
     const settings = await getProxySettingsFromSystem({} as Proxy);
-    expect(settings).toBeDefined();
-    expect(settings.httpProxy).toBeUndefined();
-    expect(settings.httpsProxy).toBeUndefined();
-    expect(settings.noProxy).toBeUndefined();
+    expect(settings).toBeUndefined();
   });
 
   test('No state returned if network connection is disabled', async () => {
     setupPlatform(Platform.MACOS);
     vi.spyOn(Exec.prototype, 'exec').mockResolvedValue({ stdout: '\n*Connection' } as RunResult);
     const settings = await getProxySettingsFromSystem({} as Proxy);
-    expect(settings).toBeDefined();
-    expect(settings.httpProxy).toBeUndefined();
-    expect(settings.httpsProxy).toBeUndefined();
-    expect(settings.noProxy).toBeUndefined();
+    expect(settings).toBeUndefined();
   });
 
   test('State returned with http proxy if network connection', async () => {
--- a/packages/main/src/plugin/proxy-system.spec.ts
+++ b/packages/main/src/plugin/proxy-system.spec.ts
@@ -124,86 +124,86 @@ describe('Windows platform tests', () => {
 
 describe('Linux platform test', () => {
   test('No state returned in case of proxy disabled', async () => {
-    const previousHttpProxy = process.env.HTTP_PROXY;
-    const previousHttpsProxy = process.env.HTTPS_PROXY;
-    const previousNoProxy = process.env.NO_PROXY;
+    const previousHttpProxy = process.env['HTTP_PROXY'];
+    const previousHttpsProxy = process.env['HTTPS_PROXY'];
+    const previousNoProxy = process.env['NO_PROXY'];
     try {
       setupPlatform(false, false, true);
-      delete process.env.HTTP_PROXY;
-      delete process.env.HTTPS_PROXY;
-      delete process.env.NO_PROXY;
+      delete process.env['HTTP_PROXY'];
+      delete process.env['HTTPS_PROXY'];
+      delete process.env['NO_PROXY'];
       const settings = await getProxySettingsFromSystem({} as Proxy);
       expect(settings).toBeDefined();
       expect(settings.httpProxy).toBeUndefined();
       expect(settings.httpsProxy).toBeUndefined();
       expect(settings.noProxy).toBeUndefined();
     } finally {
-      process.env.HTTP_PROXY = previousHttpProxy;
-      process.env.HTTPS_PROXY = previousHttpsProxy;
-      process.env.NO_PROXY = previousNoProxy;
+      process.env['HTTP_PROXY'] = previousHttpProxy;
+      process.env['HTTPS_PROXY'] = previousHttpsProxy;
+      process.env['NO_PROXY'] = previousNoProxy;
     }
   });
 
   test('State returned in case of http proxy', async () => {
-    const previousHttpProxy = process.env.HTTP_PROXY;
-    const previousHttpsProxy = process.env.HTTPS_PROXY;
-    const previousNoProxy = process.env.NO_PROXY;
+    const previousHttpProxy = process.env['HTTP_PROXY'];
+    const previousHttpsProxy = process.env['HTTPS_PROXY'];
+    const previousNoProxy = process.env['NO_PROXY'];
     try {
       setupPlatform(false, false, true);
-      process.env.HTTP_PROXY = 'http://127.0.0.1:8888';
-      delete process.env.HTTPS_PROXY;
-      delete process.env.NO_PROXY;
+      process.env['HTTP_PROXY'] = 'http://127.0.0.1:8888';
+      delete process.env['HTTPS_PROXY'];
+      delete process.env['NO_PROXY'];
       const settings = await getProxySettingsFromSystem({} as Proxy);
       expect(settings).toBeDefined();
       expect(settings?.httpProxy).toBe('http://127.0.0.1:8888');
       expect(settings?.httpsProxy).toBeUndefined();
       expect(settings?.noProxy).toBeUndefined();
     } finally {
-      process.env.HTTP_PROXY = previousHttpProxy;
-      process.env.HTTPS_PROXY = previousHttpsProxy;
-      process.env.NO_PROXY = previousNoProxy;
+      process.env['HTTP_PROXY'] = previousHttpProxy;
+      process.env['HTTPS_PROXY'] = previousHttpsProxy;
+      process.env['NO_PROXY'] = previousNoProxy;
     }
   });
 
   test('State returned in case of https proxy', async () => {
-    const previousHttpProxy = process.env.HTTP_PROXY;
-    const previousHttpsProxy = process.env.HTTPS_PROXY;
-    const previousNoProxy = process.env.NO_PROXY;
+    const previousHttpProxy = process.env['HTTP_PROXY'];
+    const previousHttpsProxy = process.env['HTTPS_PROXY'];
+    const previousNoProxy = process.env['NO_PROXY'];
     try {
       setupPlatform(false, false, true);
-      process.env.HTTPS_PROXY = 'http://127.0.0.1:8888';
-      delete process.env.HTTP_PROXY;
-      delete process.env.NO_PROXY;
+      process.env['HTTPS_PROXY'] = 'http://127.0.0.1:8888';
+      delete process.env['HTTP_PROXY'];
+      delete process.env['NO_PROXY'];
       const settings = await getProxySettingsFromSystem({} as Proxy);
       expect(settings).toBeDefined();
       expect(settings?.httpProxy).toBeUndefined();
       expect(settings?.httpsProxy).toBe('http://127.0.0.1:8888');
       expect(settings?.noProxy).toBeUndefined();
     } finally {
-      process.env.HTTP_PROXY = previousHttpProxy;
-      process.env.HTTPS_PROXY = previousHttpsProxy;
-      process.env.NO_PROXY = previousNoProxy;
+      process.env['HTTP_PROXY'] = previousHttpProxy;
+      process.env['HTTPS_PROXY'] = previousHttpsProxy;
+      process.env['NO_PROXY'] = previousNoProxy;
     }
   });
 
   test('State returned in case of no proxy', async () => {
-    const previousHttpProxy = process.env.HTTP_PROXY;
-    const previousHttpsProxy = process.env.HTTPS_PROXY;
-    const previousNoProxy = process.env.NO_PROXY;
+    const previousHttpProxy = process.env['HTTP_PROXY'];
+    const previousHttpsProxy = process.env['HTTPS_PROXY'];
+    const previousNoProxy = process.env['NO_PROXY'];
     try {
       setupPlatform(false, false, true);
-      delete process.env.HTTP_PROXY;
-      delete process.env.HTTPS_PROXY;
-      process.env.NO_PROXY = '*.internal';
+      delete process.env['HTTP_PROXY'];
+      delete process.env['HTTPS_PROXY'];
+      process.env['NO_PROXY'] = '*.internal';
       const settings = await getProxySettingsFromSystem({} as Proxy);
       expect(settings).toBeDefined();
       expect(settings?.httpProxy).toBeUndefined();
       expect(settings?.httpsProxy).toBeUndefined();
       expect(settings?.noProxy).toBe('*.internal');
     } finally {
-      process.env.HTTP_PROXY = previousHttpProxy;
-      process.env.HTTPS_PROXY = previousHttpsProxy;
-      process.env.NO_PROXY = previousNoProxy;
+      process.env['HTTP_PROXY'] = previousHttpProxy;
+      process.env['HTTPS_PROXY'] = previousHttpsProxy;
+      process.env['NO_PROXY'] = previousNoProxy;
     }
   });
 });
--- a/packages/main/src/plugin/proxy-system.spec.ts
+++ b/packages/main/src/plugin/proxy-system.spec.ts
@@ -25,6 +25,18 @@ import { Exec } from '/@/plugin/util/exec.js';
 
 import * as util from '../util.js';
 
+const mocks = vi.hoisted(() => {
+  return {
+    WinRegMock: vi.fn(),
+  };
+});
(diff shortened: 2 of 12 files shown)


lstocchi (commented):
Tried but it doesn't work as expected. 

-  If i set a proxy manually, then i switch to system, it keeps using the manual proxy.

- Maybe we should display the message "restart your container engine" even when switching from one proxy state to another.

- Then i think we should make it more clear when someone changes state.
Maybe hide all the inputs if manual is not selected?
When i first got into te proxy settings page it seemed that everything was disabled, even the system option.

lstocchi (commented):
Tried again but it still act weird.
I started Desktop, went to proxy -> manual, added a fake proxy to see if Desktop fails. But i can configure a registry, start a podman machine, pull an image.

jeffmaury:
> Tried again but it still act weird. I started Desktop, went to proxy -> manual, added a fake proxy to see if Desktop fails. But i can configure a registry, start a podman machine, pull an image.

Cannot reproduce. My scenario:
1. Stop machine
2. Configure dummy manual proxy
3. Start machine
4. pull image

image

But using a CLI I noticed that it did not try to contact 127.0.0.1 but http so seems there is a format issue here.

Can you check the content of `$HOME/AppData/Roaming/containers/containers.conf`

lstocchi:
> Can you check the content of `$HOME/AppData/Roaming/containers/containers.conf`

Just noticed it was broken. I had some "already defined" key
To be precise the extra fields were

```
secrets]
[configmaps]

[farms]
```

but i'm sure it worked before bc, now, when opening Desktop i saw the notification to setup Podman but i already had it installed. Desktop was not able to detect it though bc all commands were failing

jeffmaury:
> > Tried again but it still act weird. I started Desktop, went to proxy -> manual, added a fake proxy to see if Desktop fails. But i can configure a registry, start a podman machine, pull an image.
> 
> Cannot reproduce. My scenario:
> 
>     1. Stop machine
> 
>     2. Configure dummy manual proxy
> 
>     3. Start machine
> 
>     4. pull image
> 
> 
> image
> 
> But using a CLI I noticed that it did not try to contact 127.0.0.1 but http so seems there is a format issue here.
> 
> Can you check the content of `$HOME/AppData/Roaming/containers/containers.conf`

proxy

lstocchi:
> proxy 

Finishing one thing and i'll record a video to show what i do. Any reason why you add both http and https proxy? I expect to only have one. In my case i just added http, maybe that's the difference

jeffmaury:
> > proxy
> >       
> >         [
> >           
> >         
> >             proxy
> >           ](https://user-images.githubusercontent.com/695993/278611392-ed15f4ae-657d-4d65-86c8-cd5777e673eb.gif)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/695993/278611392-ed15f4ae-657d-4d65-86c8-cd5777e673eb.gif)
> 
> Finishing one thing and i'll record a video to show what i do. Any reason why you add both http and https proxy? I expect to only have one. In my case i just added http, maybe that's the difference

Yes http proxy is for http urls https one for https urls

packages/main/src/plugin/proxy.ts
```diff
@@ -73,56 +100,71 @@ export class Proxy {
...
-
-    this.proxySettings = {
-      httpProxy,
-      httpsProxy,
-      noProxy,
-    };
-    this.proxyState = isEnabled;
+    const isEnabled = proxyConfiguration.get<unknown>('enabled');
+    let proxyState: ProxyState = PROXY_SYSTEM;
+    if (typeof isEnabled === 'boolean') {
```
lstocchi, line 128: It looks to me that `enabled` is only updated in one place with ProxyState so why do we need to check if it's boolean?

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -37,78 +38,70 @@ async function updateProxySettings() {
...
           bind:value="{proxySettings.noProxy}"
           placeholder="Example: *.domain.com, 192.168.*.*"
           class="w-full p-2 outline-none text-sm bg-charcoal-800 rounded-sm text-gray-700 placeholder-gray-700"
           required />
       </div>
       <div class="my-2 pt-4">
-        <Button on:click="{() => updateProxySettings()}" disabled="{!proxyState}" class="w-full" icon="{faPen}">
-          Update
-        </Button>
+        <Button on:click="{() => updateProxySettings()}" class="w-full" icon="{faPen}">Update</Button>
```
lstocchi: I find the update button a bit confusing. Mainly because of the proxy mode dropdownlist position (which looks to be out of the form) and also bc all the "manual proxy" inputs are always visible, even when i'm on disabled/system mode.

By looking at it the first time i was not sure if after switching from system to disabled i had to click on `update` to save the state. The same from manual to system/disabled.

Can we add a warning to remind to click on update after the user switch mode or edit any value?

lstocchi (commented):
I cannot say that it does not work bc after knowing that i had to set the https proxy (as podman only relies on it for https calls) everything worked fine but i'm still a bit puzzled about how proxies work in Desktop.

BTW the PR is about retrieving the settings from system and that works fine.

Just a couple of comments. For the UI part, maybe @mairin can give better ideas but if everyone is ok with the current page, you can skip my comment. It can also be enhanced in a separate PR.

packages/main/src/plugin/index.ts
```diff
@@ -1607,16 +1607,16 @@ export class PluginSystem {
       },
     );
 
-    this.ipcHandle('proxy:setState', async (_listener: Electron.IpcMainInvokeEvent, state: boolean): Promise<void> => {
-      return proxy.setState(state);
+    this.ipcHandle('proxy:setState', async (_listener: Electron.IpcMainInvokeEvent, state: number): Promise<void> => {
```
lstocchi: I wonder why don't you use ProxyState instead of number?

dgolovin (commented):
We can also use https://github.com/felicienfrancois/node-electron-proxy-agent approach and there would be not need to support platform specific code to fetch proxy settings.

lstocchi:
i was able to replicate the error i had above about duplicate stuff in config file, maybe it's a problem on podman though.

duplicate_config

Another problem is that it looks like it keeps using manual proxy if set even if i select system.
What i do:
1. set non working manual proxy
2. stop and start machine
3. pull an image
4. fails as expected
5. go to proxy and change dropdown to system
6. stop and start machine
7. same error as before. However i switched to system which does not have any proxy so i would expect it to work.

For the rest, same comments as above. Hope there will be some follow-up PRs to notify the users who only set http proxy that podman only uses https proxy. Or have a way that an extension can tell if it supports proxies and which one (so let's say i only set http, i see a message - extensions podman, x, y, z only works with https proxies)

packages/main/src/plugin/index.ts
axel7083:
```suggestion
    this.ipcHandle('proxy:setState', async (_listener: Electron.IpcMainInvokeEvent, state: ProxyState): Promise<void> => {
```

packages/main/src/plugin/index.ts
```diff
@@ -1651,16 +1651,16 @@ export class PluginSystem {
...
+      return proxy.setState(state as ProxyState);
     });
 
     this.ipcHandle('proxy:getSettings', async (): Promise<containerDesktopAPI.ProxySettings | undefined> => {
       return proxy.proxy;
     });
 
-    this.ipcHandle('proxy:isEnabled', async (): Promise<boolean> => {
-      return proxy.isEnabled();
+    this.ipcHandle('proxy:getState', async (): Promise<number> => {
```
axel7083:
```suggestion
    this.ipcHandle('proxy:getState', async (): Promise<ProxyState> => {
```

packages/main/src/plugin/proxy-system.ts
```diff
@@ -0,0 +1,143 @@
...
+    const httpsProxy = process.env.HTTPS_PROXY;
+    const noProxy = process.env.NO_PROXY;
+    return { httpProxy, httpsProxy, noProxy };
+  }
+  return {} as ProxySettings;
+}
+
+async function getWindowsProxySettings(exec: Exec): Promise<ProxySettings> {
+  try {
+    const result = await exec.exec('reg', [
```
axel7083: I am never a big fan of running register query, in AI Lab to get the GPU I used https://www.npmjs.com/package/winreg, not sure if we want to include it, as it is the only place where we read the windows registry, but parsing output is always a bit of a mess for readability

packages/main/src/plugin/proxy.ts
```diff
@@ -22,6 +22,13 @@ import { ProxyAgent } from 'undici';
 import type { ConfigurationRegistry, IConfigurationNode } from './configuration-registry.js';
 import { Emitter } from './events/emitter.js';
 import { getProxyUrl } from './proxy-resolver.js';
+import { getProxySettingsFromSystem } from './proxy-system.js';
+
+export const PROXY_SYSTEM = 0;
+export const PROXY_MANUAL = 1;
+export const PROXY_DISABLED = 2;
+
+export type ProxyState = 0 | 1 | 2;
```
axel7083:
```
export enum ProxyState {
  PROXY_SYSTEM = 0,
  PROXY_MANUAL = 1,
  PROXY_DISABLED = 2,
}
```

having an enum is probably easier to understand ?

packages/main/src/plugin/proxy.ts
```diff
@@ -87,9 +94,9 @@ export class Proxy {
           hidden: true,
         },
         ['proxy.enabled']: {
-          description: 'Toggle to enable',
-          type: 'string',
-          default: false,
+          description: 'System(0)/Manual(1)/Disabled(2)',
+          type: 'number',
```
axel7083:
```suggestion
          description: 'Configure proxy',
          type: 'number',
          maximum: 2,
          minimum: 0,
          placeholder: 'System(0)/Manual(1)/Disabled(2)',
```

packages/main/src/plugin/util/exec.ts
```diff
@@ -146,6 +146,8 @@ export class Exec {
       cwd = options.cwd;
     }
 
+    console.log(`Calling ${command} HTTP_PROXY='${env.HTTP_PROXY}' HTTPS_PROXY='${env.HTTPS_PROXY}`);
```
axel7083:
```suggestion
```

to remove ?

packages/preload/src/index.ts
```diff
@@ -718,10 +718,10 @@ export function initExposure(): void {
     return ipcInvoke('proxy:getSettings');
   });
 
-  contextBridge.exposeInMainWorld('isProxyEnabled', async (): Promise<boolean> => {
-    return ipcInvoke('proxy:isEnabled');
+  contextBridge.exposeInMainWorld('getProxyState', async (): Promise<number> => {
```
axel7083: ProxyState instead of number ?

packages/preload/src/index.ts
```diff
@@ -718,10 +718,10 @@ export function initExposure(): void {
...
+    return ipcInvoke('proxy:getState');
   });
-  contextBridge.exposeInMainWorld('setProxyState', async (state: boolean): Promise<void> => {
+  contextBridge.exposeInMainWorld('setProxyState', async (state: number): Promise<void> => {
```
axel7083: ProxyState instead of number ?

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -4,21 +4,21 @@ import type { ProxySettings } from '@podman-desktop/api';
 import { Button, ErrorMessage, Input } from '@podman-desktop/ui-svelte';
 import { onMount } from 'svelte';
 
-import SlideToggle from '../ui/SlideToggle.svelte';
 import SettingsPage from './SettingsPage.svelte';
 import { validateProxyAddress } from './Util';
 
 let proxySettings: ProxySettings;
-let proxyState: boolean;
+let proxyState: number;
```
axel7083: ProxyState instead of number ?

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -63,19 +59,29 @@ function validate(event: any) {
...
-          class="mb-2 text-sm font-medium {proxyState ? 'text-[var(--pd-invert-content-card-text)]' : 'text-gray-900'}"
-          >Web Proxy (HTTP):</label>
+          class="mb-2 text-sm font-medium {proxyState === 1
+            ? 'text-[var(--pd-invert-content-card-text)]'
+            : 'text-gray-900'}">Web Proxy (HTTP):</label>
         <Input
           name="httpProxy"
           id="httpProxy"
-          disabled="{!proxyState}"
+          disabled="{proxyState !== 1}"
```
axel7083: proxyState !== PROXY_MANUAL

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -87,13 +93,13 @@ function validate(event: any) {
       <div class="space-y-2">
         <label
           for="httpsProxy"
-          class="pt-4 mb-2 text-sm font-medium {proxyState
+          class="pt-4 mb-2 text-sm font-medium {proxyState === 1
```
axel7083: proxyState === PROXY_MANUAL

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -87,13 +93,13 @@ function validate(event: any) {
...
             ? 'text-[var(--pd-invert-content-card-text)]'
             : 'text-gray-900'}">Secure Web Proxy (HTTPS):</label>
         <Input
           name="httpsProxy"
           id="httpsProxy"
-          disabled="{!proxyState}"
+          disabled="{proxyState !== 1}"
```
axel7083: proxyState !== PROXY_MANUAL

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -105,21 +111,19 @@ function validate(event: any) {
       <div class="space-y-2">
         <label
           for="httpProxy"
-          class="pt-4 mb-2 text-sm font-medium {proxyState
+          class="pt-4 mb-2 text-sm font-medium {proxyState === 1
```
axel7083: proxyState === PROXY_MANUAL

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -105,21 +111,19 @@ function validate(event: any) {
...
             ? 'text-[var(--pd-invert-content-card-text)]'
             : 'text-gray-900'}">Bypass proxy settings for these hosts and domains:</label>
         <Input
           name="noProxy"
           id="noProxy"
-          disabled="{!proxyState}"
+          disabled="{proxyState !== 1}"
```
axel7083: proxyState !== PROXY_MANUAL

lstocchi:
@jeffmaury is this ready to review?

jeffmaury:
> @jeffmaury is this ready to review?

One extra commit is to be added

packages/main/src/plugin/proxy.ts
```diff
@@ -146,10 +171,10 @@ export class Proxy {
   }
 
   isEnabled(): boolean {
-    return this.proxyState;
+    return this.proxyState === ProxyState.PROXY_SYSTEM || this.proxyState === ProxyState.PROXY_MANUAL;
```
axel7083:
```suggestion
    return this.proxyState !== ProxyState.PROXY_DISABLED;
```

packages/main/src/plugin/proxy.ts
```diff
@@ -22,6 +22,13 @@ import { ProxyAgent } from 'undici';
 import type { ConfigurationRegistry, IConfigurationNode } from './configuration-registry.js';
 import { Emitter } from './events/emitter.js';
 import { getProxyUrl } from './proxy-resolver.js';
+import { getProxySettingsFromSystem } from './proxy-system.js';
+
+export enum ProxyState {
```
axel7083: If we need this to be shared between renderer and main, we might place it in a dedicated file in `packages/api/src`, therefore both renderer and main could access it using /@api/...

lstocchi (approved):
Same concern i had before for the update button https://github.com/containers/podman-desktop/pull/4356#discussion_r1374680285 and as Axel suggested we could use the enum instead of number. 
I played with it on Windows and it works fine for me :+1:

axel7083 (approved):

packages/main/src/plugin/proxy-system.spec.ts
```diff
@@ -0,0 +1,292 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/main/src/plugin/proxy.ts
```diff
@@ -98,36 +104,51 @@ export class Proxy {
...
+      const noProxy = proxyConfiguration.get<string>('no');
+      this.proxySettings = {
+        httpProxy,
+        httpsProxy,
+        noProxy,
+      };
+    } else if (this.proxyState === ProxyState.PROXY_SYSTEM) {
+      this.proxySettings = await getProxySettingsFromSystem(this);
+    } else {
+      this.proxySettings = {} as ProxySettings;
```
benoitf, line 146: why do we use an empty object as ProxySettings

I would expect that it's undefined

benoitf: because we expect some fields from a ProxySettings object if defined

packages/main/src/plugin/proxy-system.ts
```diff
@@ -0,0 +1,141 @@
...
+    return getWindowsProxySettings();
+  } else if (isMac()) {
+    return getMacOSProxySettings(new Exec(proxy));
+  } else if (isLinux()) {
+    const httpProxy = process.env.HTTP_PROXY;
+    const httpsProxy = process.env.HTTPS_PROXY;
+    const noProxy = process.env.NO_PROXY;
+    return { httpProxy, httpsProxy, noProxy };
+  }
+  return {} as ProxySettings;
```
benoitf: should be undefined

packages/main/src/plugin/proxy-system.ts
```diff
@@ -0,0 +1,141 @@
...
+import { promisify } from 'node:util';
+
+import { type ProxySettings } from '@podman-desktop/api';
+import WinReg from 'winreg';
+
+import { isLinux, isMac, isWindows } from '../util.js';
+import type { Proxy } from './proxy.js';
+import { Exec } from './util/exec.js';
+
+export async function getProxySettingsFromSystem(proxy: Proxy): Promise<ProxySettings> {
```
benoitf: should be

```ts
Promise<ProxySettings | undefined>
```

packages/main/src/plugin/proxy-system.ts
```diff
@@ -0,0 +1,141 @@
...
+          }
+        });
+      } else if (key.name === 'ProxyOverride') {
+        noProxy = key.value;
+      }
+    });
+    return enabled ? { httpProxy, httpsProxy, noProxy } : ({} as ProxySettings);
+  } catch (err) {
+    console.warn(`Error while getting Windows internet settings from registry`, err);
+    return {} as ProxySettings;
```
benoitf: undefined

packages/main/src/plugin/proxy-system.ts
```diff
@@ -0,0 +1,141 @@
...
+            httpsProxy = `http://${item.substring(6)}`;
+          } else {
+            httpProxy = httpsProxy = `http:${item}`;
+          }
+        });
+      } else if (key.name === 'ProxyOverride') {
+        noProxy = key.value;
+      }
+    });
+    return enabled ? { httpProxy, httpsProxy, noProxy } : ({} as ProxySettings);
```
benoitf: no {}

packages/main/src/plugin/proxy-system.ts
```diff
@@ -0,0 +1,141 @@
...
+        }
+        if (!noProxy) {
+          noProxy = await getMacOSConnectionProxyByPass(exec, lines[index]);
+        }
+      }
+    }
+    return httpProxy || httpsProxy || noProxy ? { httpProxy, httpsProxy, noProxy } : ({} as ProxySettings);
+  } catch (err) {
+    console.warn(`Error while getting MacOS network services`, err);
+    return {} as ProxySettings;
```
benoitf: undefined

packages/main/src/plugin/proxy-system.ts
```diff
@@ -0,0 +1,144 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/main/src/plugin/proxy-system.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+    });
+    const settings = await getProxySettingsFromSystem({} as Proxy);
+    expect(settings).toBeDefined();
+    expect(settings?.httpProxy).toBeUndefined();
+    expect(settings?.httpsProxy).toBe('http://127.0.0.1:8888');
+    expect(settings?.noProxy).toBeUndefined();
+  });
+
+  test('State returned with no proxy if network connection', async () => {
+    setupPlatform(false, true, false);
```
benoitf: just tiny thing, but it might be easier to have the name of the platform in the method being set rather than 3 boolean (also with 3 booleans, we might set like both windows and macOS)

so something like setupMacPlatform(), setupWindowsPlatform() or setupPlatform('macOS')

packages/main/src/plugin/proxy-system.ts
benoitf: here it's still returning {} and not undefined ?

benoitf (commented):
I still see {} and not undefined values ?

packages/main/src/plugin/proxy-system.spec.ts
jeffmaury: Fixed

packages/main/src/plugin/proxy-system.ts
benoitf: and here, it's still {} being casted to ProxySettings ?

packages/main/src/plugin/proxy-system.ts
benoitf: here as well

packages/main/src/plugin/proxy-system.ts
benoitf: here as well

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -87,12 +95,13 @@ function validate(event: any) {
...
-          class="pt-4 mb-2 font-medium {proxyState ? 'text-[var(--pd-invert-content-card-text)]' : 'text-gray-900'}"
-          >Secure Web Proxy (HTTPS):</label>
+          class="pt-4 mb-2 font-medium {proxyState === 1
+            ? 'text-[var(--pd-invert-content-card-text)]'
+            : 'text-gray-900'}">Secure Web Proxy (HTTPS):</label>
         <Input
           name="httpsProxy"
           id="httpsProxy"
-          disabled={!proxyState}
+          disabled={proxyState !== 1}
```
axel7083:
```suggestion
          disabled={proxyState !== ProxyState.PROXY_MANUAL}
```

Copy paste from line 86

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -104,20 +113,19 @@ function validate(event: any) {
...
-          class="pt-4 mb-2 font-medium {proxyState ? 'text-[var(--pd-invert-content-card-text)]' : 'text-gray-900'}"
-          >Bypass proxy settings for these hosts and domains:</label>
+          class="pt-4 mb-2 font-medium {proxyState === 1
+            ? 'text-[var(--pd-invert-content-card-text)]'
+            : 'text-gray-900'}">Bypass proxy settings for these hosts and domains:</label>
         <Input
           name="noProxy"
           id="noProxy"
-          disabled={!proxyState}
+          disabled={proxyState !== 1}
```
axel7083:
```suggestion
          disabled={proxyState !== ProxyState.PROXY_MANUAL}
```

Copy paste from line 86

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -104,20 +113,19 @@ function validate(event: any) {
...
           bind:value={proxySettings.noProxy}
           placeholder="Example: *.domain.com, 192.168.*.*"
           required />
       </div>
       <div class="my-2 pt-4">
-        <Button on:click={() => updateProxySettings()} disabled={!proxyState} class="w-full" icon={faPen}>
-          Update
-        </Button>
+        <Button on:click={() => updateProxySettings()} class="w-full" icon={faPen}>Update</Button>
```
axel7083:
```suggestion
        <Button on:click={updateProxySettings} class="w-full" icon={faPen}>Update</Button>
```

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -104,20 +113,19 @@ function validate(event: any) {
       <div class="space-y-2">
         <label
           for="httpProxy"
-          class="pt-4 mb-2 font-medium {proxyState ? 'text-[var(--pd-invert-content-card-text)]' : 'text-gray-900'}"
-          >Bypass proxy settings for these hosts and domains:</label>
+          class="pt-4 mb-2 font-medium {proxyState === 1
```
axel7083:
```suggestion
          class="pt-4 mb-2 font-medium {proxyState === ProxyState.PROXY_MANUAL
```

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -87,12 +95,13 @@ function validate(event: any) {
       <div class="space-y-2">
         <label
           for="httpsProxy"
-          class="pt-4 mb-2 font-medium {proxyState ? 'text-[var(--pd-invert-content-card-text)]' : 'text-gray-900'}"
-          >Secure Web Proxy (HTTPS):</label>
+          class="pt-4 mb-2 font-medium {proxyState === 1
```
axel7083:
```suggestion
          class="pt-4 mb-2 font-medium {proxyState === ProxyState.PROXY_MANUAL
```

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -63,19 +61,29 @@ function validate(event: any) {
...
-      >Proxy configuration {proxyState ? 'enabled' : 'disabled'}</SlideToggle>
+    <label for="toggle-proxy" class="inline-flex relative items-center mt-1 mb-4 cursor-pointer"
+      >Proxy configuration
+      <select
+        class="p-2 outline-none text-sm bg-charcoal-600 rounded-sm text-gray-700 placeholder-gray-700"
+        id="toggle-proxy"
+        bind:value={proxyState}>
+        <option value={0}>System</option>
+        <option value={1}>Manual</option>
+        <option value={2}>Disabled</option>
```
axel7083:
```suggestion
        <option value={ProxyState.PROXY_SYSTEM}>System</option>
        <option value={ProxyState.PROXY_MANUAL}>Manual</option>
        <option value={ProxyState.PROXY_DISABLED}>Disabled</option>
```
No sure the linter would be happy but could give it a try

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
axel7083: Reopeing as resolved without comment

packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -4,21 +4,23 @@ import type { ProxySettings } from '@podman-desktop/api';
...
 
-import SlideToggle from '../ui/SlideToggle.svelte';
+import { ProxyState } from '/@api/proxy';
+
 import SettingsPage from './SettingsPage.svelte';
 import { validateProxyAddress } from './Util';
 
 let proxySettings: ProxySettings;
-let proxyState: boolean;
+let proxyState: number;
```
axel7083:
```suggestion
let proxyState: ProxyState;
```

packages/preload/src/index.ts
axel7083: Re-opening because close without comment

packages/preload/src/index.ts
```diff
@@ -748,10 +748,10 @@ export function initExposure(): void {
...
```
axel7083:
```suggestion
  contextBridge.exposeInMainWorld('setProxyState', async (state: ProxyState): Promise<void> => {
```

packages/preload/src/index.ts
```diff
@@ -748,10 +748,10 @@ export function initExposure(): void {
     return ipcInvoke('proxy:getSettings');
   });
 
-  contextBridge.exposeInMainWorld('isProxyEnabled', async (): Promise<boolean> => {
-    return ipcInvoke('proxy:isEnabled');
+  contextBridge.exposeInMainWorld('getProxyState', async (): Promise<number> => {
```
axel7083:
```suggestion
  contextBridge.exposeInMainWorld('getProxyState', async (): Promise<ProxyState> => {
```

packages/main/src/plugin/proxy-system.ts
jeffmaury: Yes this is for the case the system information tells proxy is disabled so an empty structure is returned (just like manual and not data specified)

axel7083:
Argos is not happy, might need a rebase!

packages/main/src/plugin/proxy.spec.ts
```diff
@@ -33,7 +34,7 @@ const certificates: Certificates = {
 } as unknown as Certificates;
 
 function getConfigurationRegistry(
-  enabled: boolean,
+  enabled: number,
```
axel7083, line 37: number ? or ProxyState ?

axel7083 (approved):
LGTM, just have some checks failing

packages/main/src/plugin/proxy.ts
```diff
@@ -104,58 +109,76 @@ export class Proxy {
...
-    this.proxySettings = {
-      httpProxy: ensureURL(httpProxy),
-      httpsProxy: ensureURL(httpsProxy),
-      noProxy,
-    };
-    this.proxyState = isEnabled;
+    const isEnabled = proxyConfiguration.get<unknown>('enabled');
+    let proxyState: ProxyState = ProxyState.PROXY_SYSTEM;
+    if (typeof isEnabled === 'boolean') {
+      proxyState = isEnabled ? ProxyState.PROXY_MANUAL : ProxyState.PROXY_SYSTEM;
```
dgolovin, line 129: Default to PROXY_SYSTEM will make podman desktop to fail to load after restart in corner cases like:
1. Running on unsupported platform
2. Detection fails on supported platform

While (1) should be unlikely, (2) will happen because of Murphy Law.

Proxy.init() called in PluginSystem.initExtensions() outside of try catch and that will end with error below

```
UnhandledPromiseRejectionWarning: Error: Unsupported environment
    at getProxySettingsFromSystem (/Users/eskimo/Sources/podman-desktop/packages/main/dist/index.cjs:122228:9)
    at Proxy2.updateFromConfiguration (/Users/eskimo/Sources/podman-desktop/packages/main/dist/index.cjs:122326:34)
    at Proxy2.init (/Users/eskimo/Sources/podman-desktop/packages/main/dist/index.cjs:122303:16)
    at PluginSystem.initExtensions (/Users/eskimo/Sources/podman-desktop/packages/main/dist/index.cjs:127287:17)
    at async /Users/eskimo/Sources/podman-desktop/packages/main/dist/index.cjs:129959:23
```

jeffmaury: Added protection against 2. 1. is unlikely

dgolovin (approved):

benoitf (approved):
thanks for the changes
make sure to rebase before merging as new lint rules were added

axel7083:
Yeahhhhhh 🎉

benoitf:
almost one year old PR 🥳🎂

PR #4358, opened 2023-10-16
chore(tests): adding extension installation verification

xbabalov:
### What does this PR do?
Integrates e2e tests for general Extension extension installation and verification in Podman Desktop, covering OpenShift Sandbox and OpenShift Local installations in one parametrized test file

https://github.com/containers/podman-desktop/issues/3340

### How to test this PR?
yarn test:e2e:smoke


tests/src/extension-installation-smoke.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    test('Verify Dashboard components', async () => {
+      await goToDashboard();
+      const dashboardInstallButton = extensionDashboardBox.getByRole('button', { name: installButtonLabel });
+      await playExpect(dashboardInstallButton).toBeVisible();
+    });
+  });
+});
+
+async function initializeLocators(extensionType: string) {
```
jeffmaury: async not required here

jeffmaury (commented):
I'm wondering if these tests should not be into the extension repo

xbabalov:
> I'm wondering if these tests should not be into the extension repo

To avoid duplicating and migrating all test files, we could setup a pipeline to run these tests in extension repositories. I created a follow up issue for that matter for OpenShift https://github.com/containers/podman-desktop/issues/4474, and we could merge the test in here. I also added new npm target so the extension test runs separately.

tests/src/model/pages/dashboard-page.ts
```diff
@@ -35,7 +40,22 @@ export class DashboardPage extends BasePage {
...
     this.devSandboxProvider = page.getByLabel('Developer Sandbox Provider');
     this.devSandboxBox = this.featuredExtensions.getByLabel('Developer Sandbox');
+    this.devSandboxEnabledStatus = page.getByText('Developer Sandbox is running');
+
+    this.openshiftLocalProvider = page.getByLabel('OpenShift Local Provider');
+    this.openshiftLocalBox = this.featuredExtensions.getByLabel('OpenShift Local');
+    this.openshiftLocalEnabledStatus = this.getOpenShiftStatus(page);
+  }
+
+  getOpenShiftStatus(page: Page) {
```
odockal: missing return type. (it does work though, but please use it.)

odockal: Please also add "Locator" suffix, so it is clear we are not returning some enum or string value. Which could be maybe the case here. If we are aware of the various status the extension can get, we can make it an enum, or at least uses constants to holds variable in.

tests/src/model/pages/dashboard-page.ts
```diff
@@ -27,6 +28,10 @@ export class DashboardPage extends BasePage {
   readonly featuredExtensions: Locator;
   readonly devSandboxProvider: Locator;
   readonly devSandboxBox: Locator;
+  readonly devSandboxEnabledStatus: Locator;
+  readonly openshiftLocalProvider: Locator;
+  readonly openshiftLocalBox: Locator;
+  openshiftLocalEnabledStatus: Locator;
```
odockal: Why not readonly?

tests/src/extension-installation.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      extensionDashboardProvider = dashboardPage.openshiftLocalProvider;
+      extensionSettingsBox = settingsExtensionsPage.openshiftLocalBox;
+      installButtonLabel = 'Install redhat.openshift-local Extension';
+      settingsLink = 'Red Hat OpenShift Local';
+      break;
+    }
+  }
+}
+
+async function goToDashboard() {
```
odockal, line 190: Cannot this be part of more general functionality util files? I guess we can make us of this elsewhere.

odockal (changes requested):

odockal:
@xbabalov Please also create an follow up issue for adding a check for resources page. (I did not notice them in the test).

tests/src/extension-installation.spec.ts
xbabalov: We can move them when they are used again in another test, maybe?

odockal: ok!

odockal (approved):

PR #4364, opened 2023-10-16
feat: adding optional abort controller to dockerode api

axel7083:
Related: https://github.com/containers/podman-desktop/issues/4363

### What does this PR do?

Adding optional AbortController argument to functions with potentially long duration

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -1003,7 +1003,6 @@ export class ContainerProviderRegistry {
       if (engine) {
         await this.attachToContainer(engine, container);
       }
-      // see https://github.com/containers/podman-desktop/issues/4363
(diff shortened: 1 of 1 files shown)


lstocchi (approved):
LGTM

packages/main/src/plugin/container-registry.ts
```diff
@@ -993,8 +1003,8 @@ export class ContainerProviderRegistry {
       if (engine) {
         await this.attachToContainer(engine, container);
       }
-
-      return container.start();
+      // see https://github.com/containers/podman-desktop/issues/4363
```
benoitf: IMHO we can remove the link to the podman desktop issue, we only reference issues in the commit message / PR description

each time we do changes in the codebase we don't add backlinks to the repository
we can track the links by looking at the git changes and get back to the commit

benoitf (approved):
checked with https://github.com/containers/podman-desktop/pull/4538

nice contribution @axel7083 

just a minor nit

PR #4369, opened 2023-10-16
feat: making task cancellable

axel7083:
Related to https://github.com/containers/podman-desktop/issues/4363

### What does this PR do?

This PR adds a `cancellableTokenId` property to Task object. This PR only adds the backend logic and a test.

In this PR I am adding the cancelable action to the Pull Image as a demonstration of how to use the cancelable tasks logic.

https://github.com/containers/podman-desktop/assets/42176370/de76aa80-d781-494e-b03c-6fd4a9c82771

### How to test this PR?

- unit tests have been added
- Go to image page
- Pull an image

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -708,12 +708,9 @@ declare module '@podman-desktop/api' {
     title?: string;
 
     /**
-     * Controls if a cancel button should show to allow the user to
-     * cancel the long running operation.  Note that currently only
-     * `ProgressLocation.Notification` is supporting to show a cancel
-     * button.
+     * Correspond to a cancel token that can be used to cancel the task
      */
-    cancellable?: boolean;
+    cancellableTokenId?: number;
   }
 
   /**
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1494,7 +1494,7 @@ declare module '@podman-desktop/api' {
 
     export function withProgress<R>(
       options: ProgressOptions,
-      task: (progress: Progress<{ message?: string; increment?: number }>, token: CancellationToken) => Promise<R>,
+      task: (progress: Progress<{ message?: string; increment?: number }>, token?: CancellationToken) => Promise<R>,
     ): Promise<R>;
 
     /**
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -154,6 +154,7 @@ import { OpenDevToolsInit } from './open-devtools-init.js';
 import { NavigationManager } from '/@/plugin/navigation/navigation-manager.js';
 import { WebviewRegistry } from './webview/webview-registry.js';
 import type { IDisposable } from './types/disposable.js';
+import type { Task } from '/@/plugin/api/task.js';
 
 import { KubernetesUtils } from './kubernetes-util.js';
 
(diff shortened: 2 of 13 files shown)


axel7083:
Requires https://github.com/containers/podman-desktop/pull/4342 to show an error on the TASKS panel when the task is cancelled.

axel7083:
CC @lstocchi I saw you where the one to introduce the cancellable action on the `Build Image` page. I tried to follow the same logic as you did for the `Pull Image` action page.

packages/main/src/plugin/index.ts
```diff
@@ -2153,6 +2187,26 @@ export class PluginSystem {
...
+      if (tokenSource) {
+        // We cancel it
+        tokenSource.cancel();
+        // Update the task accordinaly
+        if (taskManager.isStatefulTask(task)) {
+          task.state = 'completed';
+          task.status = 'cancelled';
+          task.error = 'task cancelled';
+        }
+        taskManager.updateTask(task);
```
lstocchi: you could call the cancelTask function from here to avoid code duplication. And maybe renaming it as it does not really cancel a task but it just update the task state

lstocchi (commented):
Overall LGTM, just a tiny comment.
I also saw an issue when you cancel the task from the taskmanager, its status seems to be updated but to actually see it you have to close/reopen the taskmanger.
task_state

packages/main/src/plugin/index.ts
axel7083: Agreed !

axel7083:
> Overall LGTM, just a tiny comment. I also saw an issue when you cancel the task from the taskmanager, its status seems to be updated but to actually see it you have to close/reopen the taskmanger. 

The code selecting the icon was missing a check on the task ! Thanks @lstocchi

axel7083:
cc @benoitf :)

lstocchi (approved):
LGTM

axel7083:
Wait for 1.7 to be release

packages/extension-api/src/extension-api.d.ts
```diff
@@ -1494,7 +1491,7 @@ declare module '@podman-desktop/api' {
 
     export function withProgress<R>(
       options: ProgressOptions,
-      task: (progress: Progress<{ message?: string; increment?: number }>, token: CancellationToken) => Promise<R>,
+      task: (progress: Progress<{ message?: string; increment?: number }>, token?: CancellationToken) => Promise<R>,
```
benoitf, line 1494: token is mandatory

packages/extension-api/src/extension-api.d.ts
```diff
@@ -708,12 +708,9 @@ declare module '@podman-desktop/api' {
...
 
     /**
-     * Controls if a cancel button should show to allow the user to
-     * cancel the long running operation.  Note that currently only
-     * `ProgressLocation.Notification` is supporting to show a cancel
-     * button.
+     * Correspond to a cancel token that can be used to cancel the task
      */
-    cancellable?: boolean;
+    cancellableTokenId?: number;
```
benoitf, line 713: it should be kept as it is

1. we can't change things like that on public API
2. it's to control a display thing, it's not related to token ids

benoitf (changes requested):

packages/extension-api/src/extension-api.d.ts
axel7083: What if the options set cancellable to false ?

packages/extension-api/src/extension-api.d.ts
axel7083: Then how extension can provide a way to cancel themself a task ?

Inside the `task: (progress: Progress<{ message?: string; increment?: number }>, token?: CancellationToken` the CancellationToken, does not allow to cancel the request.

Therefore extensions can only completed it, or throw an error inside.

packages/extension-api/src/extension-api.d.ts
benoitf: a token is to know if the task has been canceled or not, if cancellable is false, it'll just always return that task was never canceled

packages/extension-api/src/extension-api.d.ts
benoitf: the idea is that it's the framewor/user that can do the cancellation, not the extension directly. The extension monitor the token to check if it has been canceled and then decide to abort if user asked for cancellation

axel7083: Yes but for example, I have a download button, on my webview, I want to create a Task so the user can know if he leave the page, that it is still running, but I also want a cancel button on my webview.

The pulling is a good example, I want a cancel button on the page, and also a cancel button inside the Task-Manager

axel7083: Another solution, we add in the Progress object a cancel method (so we ensure backward compatibility), but it makes no sense that the creator of the task do not have controls over it

axel7083:
Closing as will be spited in smaller PR

PR #4370, opened 2023-10-16
fix: details page consistency

### LINKED ISSUE #4226: Inconsistencies in DetailsPage

deboer-tim:
### What does this PR do?

Moves the close button and actions of the details page to be consistent with form pages. It also makes the subtitle font a little smaller, and adds a missing shadow-pageheader that was in the design.

The overall page structure & layout is much closer to form pages after this. I didn't go full shared-component because there are still a few differences between the two, but this is a step towards a common page component or breadcrumb + close header.

Shouldn't really be released before PR #4227. It's not awful, but one of the reasons for switching to donuts is that it takes up less horizontal space and won't go over the tab titles.

Before:

Screenshot 2023-10-16 at 4 25 28 PM

After:

Screenshot 2023-10-16 at 4 23 48 PM

Fixes #4226.

### How to test this PR?

Just go to details pages.

### DIFF
--- a/packages/renderer/src/lib/ui/DetailsPage.svelte
+++ b/packages/renderer/src/lib/ui/DetailsPage.svelte
@@ -24,11 +24,11 @@ function handleKeydown(e: KeyboardEvent) {
 <div class="flex flex-col w-full h-full shadow-pageheader">
   <div class="flex flex-row w-full h-fit px-5 py-4">
     <div class="flex flex-col w-full h-fit">
-      <div class="flex flew-row items-center">
+      <div class="flex flew-row items-center text-sm text-gray-700">
         <Link class="text-sm" aria-label="back" internalRef="{$lastPage.path}" title="Go back to {$lastPage.name}"
           >{$lastPage.name}</Link>
-        <div class="text-md mx-2 text-gray-700">></div>
-        <div class="grow text-sm font-extralight text-gray-700" aria-label="name">{$currentPage.name}</div>
+        <div class="mx-2">&gt;</div>
+        <div class="grow font-extralight" aria-label="name">{$currentPage.name}</div>
         <a href="{$lastPage.path}" title="Close" class="justify-self-end text-gray-900">
           <i class="fas fa-times" aria-hidden="true"></i>
         </a>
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/ui/DetailsPage.svelte
```diff
@@ -21,46 +21,49 @@ function handleKeydown(e: KeyboardEvent) {
...
-            >{$lastPage.name}</Link>
-          <div class="mx-2">&gt;</div>
-          <div class="font-extralight" aria-label="name">{$currentPage.name}</div>
+<div class="flex flex-col w-full h-full shadow-pageheader">
+  <div class="flex flex-row w-full h-fit px-5 py-4">
+    <div class="flex flex-col w-full h-fit">
+      <div class="flex flew-row items-center">
+        <Link class="text-sm" aria-label="back" internalRef="{$lastPage.path}" title="Go back to {$lastPage.name}"
+          >{$lastPage.name}</Link>
+        <div class="text-md mx-2 text-gray-700">></div>
```
benoitf: any reason why `&gt;` is replaced with `>` ?

deboer-tim: Purely due to using multiple workspaces and accidentally undoing my own changes... sorry, fixed now.

jeffmaury (approved):
LGTM

benoitf (approved):

PR #4371, opened 2023-10-17
fix: use encodeURIComponent to encode engineId

### LINKED ISSUE #3868: Empty screen when clicking on pod on ocp sandbox
### Bug description

empty_screen_pod

### Operating system

Windows 11

### Installation Method

### Version

1.3.1

### Steps to reproduce

1. create a pod on your ocp sandbox cluster
2. click on the pod from the pods list

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

- Use encode/decodeURIComponent instead of encode/decodeURI to encode/decode the engineId, which can contain / characters (in ocp sandbox for example).

Fixes #3868

### How to test this PR?

See steps in #3868


packages/renderer/src/routes.ts
```diff
@@ -0,0 +1,21 @@
+export class PodDetailsRoute {
```
benoitf: Hello, I think we should not add code in src/ folder, as it's dedicated to Pods probably it can go as new method of the pod-utils


https://github.com/containers/podman-desktop/blob/main/packages/renderer/src/lib/pod/pod-utils.ts

packages/renderer/src/App.svelte
```diff
@@ -145,11 +146,11 @@ window.events?.receive('display-troubleshooting', () => {
...
         </Route>
-        <Route path="/pods/:kind/:name/:engineId/*" breadcrumb="Pod Details" let:meta navigationHint="details">
+        <Route path="{PodDetailsRoute.path()}" breadcrumb="Pod Details" let:meta navigationHint="details">
           <PodDetails
-            podName="{decodeURI(meta.params.name)}"
-            engineId="{decodeURI(meta.params.engineId)}"
-            kind="{decodeURI(meta.params.kind)}" />
+            podName="{PodDetailsRoute.getName(meta.params)}"
+            engineId="{PodDetailsRoute.getEngineId(meta.params)}"
+            kind="{PodDetailsRoute.getKind(meta.params)}" />
```
benoitf: Maybe it will be easier to have a getRouteDetails on utility class and then just pass the argument

like

```svelte
<Route path="podUtils.path" breadcrumb="Pod Details" let:meta navigationHint="details">
          {@const routeDetails = podUtils.getRouteDetails(meta.params)}
          <PodDetails
               podName="{routeDetails.podName}"
               engineId="{routeDetails.engineId}"
```

packages/renderer/src/lib/pod/PodsList.svelte
```diff
@@ -126,7 +127,15 @@ async function deleteSelectedPods() {
 }
 
 function openDetailsPod(pod: PodInfoUI) {
-  router.goto(`/pods/${encodeURI(pod.kind)}/${encodeURI(pod.name)}/${encodeURI(pod.engineId)}/logs`);
+  const route = PodDetailsRoute.getRoute(
```
benoitf: we can ask podUtils object instead of adding a new PodDetailsRoute

also we can provide podInfoUi object to the method instead of providing each individual parameter

packages/renderer/src/routes.ts
feloy: My idea was to add new class/methods for each route defined in `App.svelte` (at least the ones with parameters), which are not all related to pods, and have these methods at the same level of `App.svelte`

benoitf: I would say we need to split "code refactoring" and "fixes"

so do not try to change all routes of App.svelte and fix the encoding at the same time.

We need separate iteration/steps
one PR = one goal (can be refactoring (no-op) or bug fixes)

I don't see value in just adding a dedicated object just for routes as routes are specific to some area (like pods, containers, images, etc and we already have utility there)

in PodList we already have the object there

packages/renderer/src/lib/pod/PodsList.svelte
feloy: I would not want to make the (generic) Route methods coupled with the podInfoUi. For example in the following line, it would need to build such an podInfoUi object from the ContainerGroupInfoUI object and the `podman` constant:
https://github.com/containers/podman-desktop/blob/main/packages/renderer/src/lib/ContainerList.svelte#L349

I would prefer to simply call it as:
```
PodDetailsRoute.getRoute({
  kind: 'podman',
  name: containerGroup.name,
  engineId: containerGroup.engineId
})
```

benoitf: but you can have two methods, one with PodInfoUI and one with meta.params so it's convenient on client side

benoitf: in the example you shown https://github.com/containers/podman-desktop/blob/main/packages/renderer/src/lib/ContainerList.svelte#L349-L351  you can provide containerGrou/ContainerGroupInfoUI object instead of building a new object

packages/renderer/src/routes.ts
feloy: > I would say we need to split "code refactoring" and "fixes"

You are right. I've created a new PR with the encoding fix only. I can work on the refactoring later, when I have a better idea of the source code.

benoitf (commented):
hello @feloy could you add unit tests to check that we're fixing the problem

?

feloy:
> hello @feloy could you add unit tests to check that we're fixing the problem

@benoitf  I'm not sure how to test this.

This involves the Route, which will analyze the passed `path` and split it to extract the engineId. 

But the Route is defined at the top level, in `App.svelte`, which means the test would need to instantiate the whole app.

What do you think?

benoitf (commented):
@feloy I agree it might be difficult without a refactoring to test the current design of App.svelte as it's not so modular/delegating stuff (the decode part)

But we can I think test the part on encoding like on ContainerList, PodActions, PodDetailsLogs PodList to see if the router is correctly calling with the expected route

for example if we call in PodList the openDetails we can check the parameter on router.goto and see the behaviour with encodeURI vs encodeURIComponent

benoitf (approved):
thanks @feloy

lstocchi (approved):
LGTM. Tested, works fine 🚀

Issue #3793, opened 2023-09-06
Telemetry requests for Onboarding feature


mairin:
### Is your enhancement related to a problem? Please describe

This is a ticket to list UX requests for telemetry on the onboarding feature so we can see if the design was effective and identify issues with the design as users begin to make use of it.

### Describe the solution you'd like

Here's my requested telemetry data for onboarding... I'll pose it in terms of the data I would like, and maybe we can rephrase each one in terms of the specific needed hooks?

## First-run (not Settings) / First page

- [ ] How many extensions does the user select? 
- [ ] How many users select $EXTENSION? (E.g. how many pick Podman / Docker / Lima / etc)
- [ ] How many users skip onboarding (how many hit "skip this entire setup" in the upper left + how many hit "cancel" in the bottom right.)

## Podman onboarding

- [x] How many users have podman already installed vs need to install it?
- [x] How many users meet system requirements vs. don't meet system requirements?
- [x] How many users hit an error setting up podman vs successfully complete?
- [ ] How many users skip Podman setup past the first onboarding menu (where you pick which extensions you want) screen?

## First-run (not Settings) / General

- [ ] Once the user has completed all onboarding screens and is brought to the dashboard... what is the first screen they navigate to?

(will edit later with more)

PR #4377, opened 2023-10-17
feat: add telemetry for podman onboarding (#3793)

lstocchi:
### What does this PR do?

This PR adds telemetry to the podman onboarding 

N/A

it is part of #3793

### How to test this PR?

N/A

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -1004,11 +1004,13 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
       extensionApi.context.setValue('podmanIsNotInstalled', !installed, 'onboarding');
       if (installed) {
         extensionApi.context.setValue('podmanInstalledTitle', 'Podman already installed', 'onboarding');
-        telemetryLogger.logUsage('podman.onboarding.podman-already-installed');
       } else {
         extensionApi.context.setValue('podmanInstalledTitle', 'Podman successfully installed', 'onboarding');
-        telemetryLogger.logUsage('podman.onboarding.podman-not-installed');
       }
+      telemetryLogger.logUsage('podman.onboarding.checkPodmanInstalled', {
+        status: installed,
+        version: installation.version || '',
+      });
     },
   );
 
@@ -1047,6 +1049,7 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
       const warnings = [];
       const telemetryRecords: Record<string, unknown> = {};
+      telemetryRecords.successful = successful;
 
       for (const res of result) {
         const warning = {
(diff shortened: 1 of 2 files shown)


extensions/podman/src/extension.ts
```diff
@@ -1004,8 +1004,10 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
       extensionApi.context.setValue('podmanIsNotInstalled', !installed, 'onboarding');
       if (installed) {
         extensionApi.context.setValue('podmanInstalledTitle', 'Podman already installed', 'onboarding');
+        telemetryLogger.logUsage('podman.onboarding.podman-already-installed');
       } else {
         extensionApi.context.setValue('podmanInstalledTitle', 'Podman successfully installed', 'onboarding');
+        telemetryLogger.logUsage('podman.onboarding.podman-not-installed');
```
jeffmaury: I'm wondering about the title that is set in the line above ?

extensions/podman/src/extension.ts
```diff
@@ -1004,8 +1004,10 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
       extensionApi.context.setValue('podmanIsNotInstalled', !installed, 'onboarding');
       if (installed) {
         extensionApi.context.setValue('podmanInstalledTitle', 'Podman already installed', 'onboarding');
+        telemetryLogger.logUsage('podman.onboarding.podman-already-installed');
```
jeffmaury: Can we also record the version ?

extensions/podman/src/extension.ts
```diff
@@ -1069,9 +1080,17 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
+        if (installed) {
+          telemetryLogger.logUsage('podman.onboarding.podman-installed', {
+            version: installation.version,
+          });
+        }
       } catch (e) {
         console.error(e);
         extensionApi.context.setValue('podmanIsNotInstalled', true, 'onboarding');
+        telemetryLogger.logError('podman.onboarding.podman-installation', {
+          error: String(e),
```
jeffmaury: Why don't you pass e as it may leads to [object Object] being recorded

extensions/podman/src/extension.ts
```diff
@@ -1069,9 +1080,17 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
         extensionApi.context.setValue('podmanIsNotInstalled', !installed, 'onboarding');
+        if (installed) {
+          telemetryLogger.logUsage('podman.onboarding.podman-installed', {
+            version: installation.version,
+          });
+        }
       } catch (e) {
         console.error(e);
         extensionApi.context.setValue('podmanIsNotInstalled', true, 'onboarding');
+        telemetryLogger.logError('podman.onboarding.podman-installation', {
```
jeffmaury: I think we should record the action (here installPodman) with data: version, installed (false/true) and error the second one being the only mandatory but in all case the event name should be the same

extensions/podman/src/extension.ts
jeffmaury: Also the event name should be checkPodmanInstalled and we should report the status

jeffmaury (changes requested):

extensions/podman/src/extension.ts
lstocchi: it is just to customize the title shown during the onboarding. If when starting it podman is discovered we show `podman already installed` otherwise it means that is going to be installed and if everything succeed we show "podman successfully installed" in the successful step

lstocchi:
@jeffmaury i guess i got how to write them now lol. Please check it again

jeffmaury (approved):
LGTM

extensions/podman/src/extension.ts
```diff
@@ -1054,24 +1060,39 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
+          installed: installed,
+        });
       } catch (e) {
         console.error(e);
         extensionApi.context.setValue('podmanIsNotInstalled', true, 'onboarding');
+        telemetryLogger.logError('podman.onboarding.installPodman', {
+          version: installation.version || '',
+          installed: installed,
+          error: e,
+        });
```
benoitf: we should use logUsage in both case (see some previous comments/issue)
https://github.com/containers/podman-desktop/pull/4407
https://github.com/containers/podman-desktop/issues/4214

benoitf: also probably need to use the pattern try / catch / finally (and do telemetry in finally block)

benoitf (changes requested):

benoitf (approved):

PR #2990, opened 2023-06-23
fix: fail extension load when manifest is incomplete

deboer-tim:
### What does this PR do?

This changes analyzeExtension() to reject the Promise if there is an error loading an extension (no manifest or missing required attributes), instead of returning undefined. The main reason for the change is to allow error messages to be propagated back to the client (e.g. UI).

The remaining changes are handling this in all the places that call analyzeExtension(). IMHO most are more readable and handle error cases cleaner after this change, but for lines 267/272/286 I tried several options and couldn't find anything better than adding '.catch(() => undefined))).'

There is also a related issue: if an extension is corrupt or fails to install, the extension installer doesn't clean it up. This doesn't cause any direct issues (on each startup it would quietly be ignored before; now it will log) but I was going to defer this to another PR.

### Screenshot/screencast of this PR

Screenshot 2023-06-23 at 2 17 03 PM

### What issues does this PR fix or reference?

Fixes #2871.

### How to test this PR?

See issue #2871 - try installing this extension: quay.io/lstocchi/sample-podman-extension:v1

PR #4380, opened 2023-10-17
fix: report errors when analyzing extensions

### LINKED ISSUE #2871: When installing an extension that has an undefined displayName the extensions section gets broken
### Bug description

If an extension does not have a display name the extensions section is broken.
 
extension-tree-broken

It fails at 

```
export async function fetchExtensions() {
  const result = await window.listExtensions();
  result.sort((a, b) => a.displayName.localeCompare(b.displayName));
  extensionInfos.set(result);
}
```

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. install quay.io/lstocchi/sample-podman-extension:v1
2. check that your extensions section is broken

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Since PR #2990 is very old I'm creating a new one.

Adds an error property to analyzed extensions, and analyzeExtension() sets it whenever there is a problem loading a folder instead of returning undefined. Added error for missing manifest entries.

Updated callers to check for error instead of undefined, and send an error when trying to install an extension that has an error.

Tests updated to confirm an error is being created.

Screenshot 2023-10-17 at 1 26 31 PM

Fixes #2871.

### How to test this PR?

See issue https://github.com/containers/podman-desktop/issues/2871 - try installing this extension: quay.io/lstocchi/sample-podman-extension:v1

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -325,23 +325,36 @@ export class ExtensionLoader {
   async analyzeExtension(extensionPath: string, removable: boolean): Promise<AnalyzedExtension> {
     // do nothing if there is no package.json file
     let error = undefined;
+    const disposables: Disposable[] = [];
     if (!fs.existsSync(path.resolve(extensionPath, 'package.json'))) {
       error = `Ignoring extension ${extensionPath} without package.json file`;
       console.warn(error);
-      return { error } as AnalyzedExtension;
+      const analyzedExtension: AnalyzedExtension = {
+        id: '<unknown>',
+        name: '<unknown>',
+        path: extensionPath,
+        manifest: undefined,
+        api: <typeof containerDesktopAPI>{},
+        removable,
+        subscriptions: disposables,
+        dispose(): void {
+          disposables.forEach(disposable => disposable.dispose());
+        },
+        error,
+      };
+      return analyzedExtension;
     }
 
     // log error if the manifest is missing required entries
     const manifest = await this.loadManifest(extensionPath);
     if (!manifest.name || !manifest.displayName || !manifest.version || !manifest.publisher || !manifest.description) {
-      error = `Extension ${extensionPath} missing required manifest entries in package.json`;
+      error = `Extension ${extensionPath} missing required manifest entry in package.json (name, displayName, version, publisher, description)`;
       console.warn(error);
     }
 
     // create api object
     const api = this.createApi(extensionPath, manifest);
 
-    const disposables: Disposable[] = [];
     const analyzedExtension: AnalyzedExtension = {
       id: `${manifest.publisher}.${manifest.name}`,
       name: manifest.name,
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -325,7 +325,6 @@ export class ExtensionLoader {
   async analyzeExtension(extensionPath: string, removable: boolean): Promise<AnalyzedExtension> {
     // do nothing if there is no package.json file
     let error = undefined;
-    const disposables: Disposable[] = [];
     if (!fs.existsSync(path.resolve(extensionPath, 'package.json'))) {
       error = `Ignoring extension ${extensionPath} without package.json file`;
       console.warn(error);
@@ -336,10 +335,8 @@ export class ExtensionLoader {
         manifest: undefined,
         api: <typeof containerDesktopAPI>{},
         removable,
-        subscriptions: disposables,
-        dispose(): void {
-          disposables.forEach(disposable => disposable.dispose());
-        },
+        subscriptions: [],
+        dispose(): void {},
         error,
       };
       return analyzedExtension;
@@ -355,6 +352,7 @@ export class ExtensionLoader {
     // create api object
     const api = this.createApi(extensionPath, manifest);
 
+    const disposables: Disposable[] = [];
     const analyzedExtension: AnalyzedExtension = {
       id: `${manifest.publisher}.${manifest.name}`,
       name: manifest.name,
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -93,6 +93,8 @@ export interface AnalyzedExtension {
   missingDependencies?: string[];
   circularDependencies?: string[];
 
+  error?: string;
+
   readonly subscriptions: { dispose(): unknown }[];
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/extension-loader.ts
```diff
@@ -310,22 +312,29 @@ export class ExtensionLoader {
...
-  async analyzeExtension(extensionPath: string, removable: boolean): Promise<AnalyzedExtension | undefined> {
+  async analyzeExtension(extensionPath: string, removable: boolean): Promise<AnalyzedExtension> {
     // do nothing if there is no package.json file
+    let error = undefined;
     if (!fs.existsSync(path.resolve(extensionPath, 'package.json'))) {
-      console.warn(`Ignoring extension ${extensionPath} without package.json file`);
-      return undefined;
+      error = `Ignoring extension ${extensionPath} without package.json file`;
+      console.warn(error);
+      return { error } as AnalyzedExtension;
```
benoitf: here I think it may lead to client's error as the object they'll get is not compliant to the typescript contract.
So while I see no problem of using that pattern in tests, we should not in main code as it may lead to runtime errors.

packages/main/src/plugin/extension-loader.ts
```diff
@@ -310,22 +312,29 @@ export class ExtensionLoader {
...
     }
 
+    // log error if the manifest is missing required entries
     const manifest = await this.loadManifest(extensionPath);
+    if (!manifest.name || !manifest.displayName || !manifest.version || !manifest.publisher || !manifest.description) {
+      error = `Extension ${extensionPath} missing required manifest entries in package.json`;
```
benoitf: maybe add the required attributes in the error message (name, displayName, version, publisher & description ?) so it might help the user

benoitf (commented):
overall LGTM

probably add tests to check specific error message returned

packages/main/src/plugin/extension-loader.ts
deboer-tim: Understood, and I agree that would be better. I've changed it to return a regular AnalyzedExtension now. I didn't want to change the API contract, so added a couple 'fake' values like id and name (felt like these could cause problems if empty), and tried to give reasonable defaults/undefined elsewhere. I wasn't sure if there's a simpler way to do the subscription/dispose(), please review again.

packages/main/src/plugin/extension-loader.ts
deboer-tim: Good point, added.

packages/main/src/plugin/extension-loader.ts
```diff
@@ -312,27 +314,47 @@ export class ExtensionLoader {
...
+        id: '<unknown>',
+        name: '<unknown>',
+        path: extensionPath,
+        manifest: undefined,
+        api: <typeof containerDesktopAPI>{},
+        removable,
+        subscriptions: disposables,
+        dispose(): void {
+          disposables.forEach(disposable => disposable.dispose());
+        },
```
benoitf: I'm not sure why we dispose the disposables. Do we register subscriptions in case of errors ?

benoitf (commented):
overall, LGTM

I'll need to test with an extension

packages/main/src/plugin/extension-loader.ts
deboer-tim: Because we still have to return the subscriptions array from here I felt like we should correctly dispose it... but had forgot it was readonly anyway. Simplified this in new commit.

deboer-tim:
@benoitf could you take another look?

lstocchi (approved):
LGTM 🚀

packages/main/src/plugin/extension-loader.ts
```diff
@@ -293,12 +295,12 @@ export class ExtensionLoader {
...
     const analyzedFoldersExtension = (
       await Promise.all(folders.map(folder => this.analyzeExtension(folder, false)))
-    ).filter(extension => extension !== undefined) as AnalyzedExtension[];
+    ).filter(extension => !extension.error) as AnalyzedExtension[];
     analyzedExtensions.push(...analyzedFoldersExtension);
 
     const analyzedExternalExtensions = (
       await Promise.all(externalExtensions.map(folder => this.analyzeExtension(folder, false)))
-    ).filter(extension => extension !== undefined) as AnalyzedExtension[];
+    ).filter(extension => !extension.error) as AnalyzedExtension[];
```
```suggestion
    ).filter(extension => !extension.error);
```

packages/main/src/plugin/extension-loader.ts
```diff
@@ -312,22 +314,40 @@ export class ExtensionLoader {
       // collect all extensions from the pluginDirectory folders
       const analyzedPluginsDirectoryExtensions: AnalyzedExtension[] = (
         await Promise.all(pluginDirectories.map(folder => this.analyzeExtension(folder, true)))
-      ).filter(extension => extension !== undefined) as AnalyzedExtension[];
+      ).filter(extension => !extension.error) as AnalyzedExtension[];
```
```suggestion
      ).filter(extension => !extension.error);
```

packages/main/src/plugin/extension-loader.ts
```diff
@@ -293,12 +295,12 @@ export class ExtensionLoader {
 
     const analyzedFoldersExtension = (
       await Promise.all(folders.map(folder => this.analyzeExtension(folder, false)))
-    ).filter(extension => extension !== undefined) as AnalyzedExtension[];
+    ).filter(extension => !extension.error) as AnalyzedExtension[];
```
dgolovin: Here and below `as AnalyzedExtension[]` is not needed.
```suggestion
    ).filter(extension => !extension.error) as AnalyzedExtension[];
```

dgolovin (changes requested):

packages/main/src/plugin/extension-loader.ts
deboer-tim: Good catch, fixed all three instances in latest commit.

packages/main/src/plugin/extension-loader.ts
deboer-tim: Fixed.

packages/main/src/plugin/extension-loader.ts
deboer-tim: Fixed.

benoitf (approved):

deboer-tim:
@dgolovin good to remove your request (fixed) and merge?

dgolovin (approved):

PR #4387, opened 2023-10-18
feat: settings page for CLI Tools

dgolovin:
### What does this PR do?

Adds `Settings/CLI Tools` page to show cli tools registered by extensions using cli API

### Screenshot/screencast of this PR

image

### What issues does this PR fix or reference?

related to #3781
depends on #4197

PR #4388, opened 2023-10-18
feat: register docker-compose cli tool through podman-desktop/cli API

dgolovin:
### What does this PR do?

Uses cli API to register docker-compose CLI tool in podman-desktop

Screenshot 2023-10-18 at 12 29 29 AM

related to https://github.com/containers/podman-desktop/issues/3781
depends on https://github.com/containers/podman-desktop/pull/4197

### DIFF
--- a/extensions/compose/src/compose-extension.spec.ts
+++ b/extensions/compose/src/compose-extension.spec.ts
@@ -114,6 +114,9 @@ vi.mock('@podman-desktop/api', () => {
         return telemetryLogger;
       },
     },
+    cli: {
+      createCliTool: vi.fn(),
+    },
   };
 });
 
--- a/extensions/compose/src/compose-extension.ts
+++ b/extensions/compose/src/compose-extension.ts
@@ -135,6 +135,20 @@ export class ComposeExtension {
       this.statusBarItem.show();
       this.extensionContext.subscriptions.push(this.statusBarItem);
     }
+    const description = `Compose is a tool for defining and running multi-container Docker applications. 
+    With Compose, you use a YAML file to configure your application's services. Then, with a single command, 
+    you create and start all the services from your configuration.
+    Compose works in all environments; production, staging, development, testing, as well as CI workflows. 
+    It also provides commands for managing the whole lifecycle of your application.`;
+
+    extensionApi.cli.createCliTool({
+      name: 'docker-compose',
+      displayName: 'Compose',
+      markdownDescription: description.replace('\n', ''),
+      images: {
+        icon: 'icon.png',
+      },
+    });
 
     // run init checks
     await this.runChecks(true);


deboer-tim:
That description will be way too long by the time we get multiple tools. Can we have the UI control crop long descriptions to 2-3 lines, with some way to show the full text (e.g. twisty or "more" link)?

benoitf:
what is looking odd to me is that we somehow want to promote `podman compose` and not directly `docker-compose` so maybe it should be highlighted somehow that's an enabler for 'podman cli tool' or that we display 'podman compose' instead of 'docker-compose' I don't know

benoitf:
the icon is not the one from the compose extension also ? https://github.com/containers/podman-desktop/blob/main/extensions/compose/icon.png

cdrage (changes requested):
Some of my review comments may have been lost, I see a few things that we should do:

- Use the description and icon added from: https://github.com/containers/podman-desktop/pull/4179
- Defining the createCliTool should be done within`extension.ts` not `compose-extension.ts` as we are removing the `compose-extension.ts` file in the future / not having Compose on the status bar anymore.

cdrage:
> what is looking odd to me is that we somehow want to promote `podman compose` and not directly `docker-compose` so maybe it should be highlighted somehow that's an enabler for 'podman cli tool' or that we display 'podman compose' instead of 'docker-compose' I don't know

We should use the description that is being added via https://github.com/containers/podman-desktop/pull/4179

dgolovin:
> @deboer-tim: That description will be way too long by the time we get multiple tools. Can we have the UI control crop long descriptions to 2-3 lines, with some way to show the full text (e.g. twisty or "more" link)?

This should be implemented in #3781.

dgolovin:
> @benoitf: what is looking odd to me is that we somehow want to promote podman compose and not directly docker-compose so maybe it should be highlighted somehow that's an enabler for 'podman cli tool' or that we display 'podman compose' instead of 'docker-compose' I don't know

Description should be definitely updated to be about 'podman compose'. On the left side it should still say 'docker-compose' because users should know exactly what is going to be installed.

cdrage:
Hi @dgolovin I've opened up https://github.com/containers/podman-desktop/pull/4437 to replace this PR with some changes.

dgolovin:
@cdrage Do you want to replace #4387 and #4388 with this PR?

mairin:
I don't think we can use that icon. We have permission to use the Docker icon for referring to Docker but only the one we were given (we use it for the Docker provider on the Dashboard and in Settings > Resources.) I would use the Compose icon because that's affiliated with the tool.

I agree with @deboer-tim that the text is too long. Where is it coming from? Is it pulling from a file included with docker-compose? This is what I came up with for the mockup:

"The Compose extension provides optional command line support for compose files with Podman. More details, including usage info, are available in Compose Settings."

cdrage:
> I don't think we can use that icon. We have permission to use the Docker icon for referring to Docker but only the one we were given (we use it for the Docker provider on the Dashboard and in Settings > Resources.) I would use the Compose icon because that's affiliated with the tool.
> 
> I agree with @deboer-tim that the text is too long. Where is it coming from? Is it pulling from a file included with docker-compose? This is what I came up with for the mockup:
> 
> "The Compose extension provides optional command line support for compose files with Podman. More details, including usage info, are available in Compose Settings."

Im going to close this PR as https://github.com/containers/podman-desktop/pull/4437 replaces it.

Let's continue the discussion in https://github.com/containers/podman-desktop/pull/4437 but can we use the already existing compose logo. I believe this icon.png was added just for testing by @dgolovin

PR #4419, opened 2023-10-19
fix: calc width of bottom toolbar in onboarding (#4418)

### LINKED ISSUE #4418: Bottom onboarding toolbar does not show the next/cancel buttons when switching to/from an external program
### Bug description

This thing happened when testing the onboarding on Mac.

When installing podman, the Podman Desktop window disappears to display the podman wizard. At the end of the installation, Desktop gets the focus back, the correct step is displayed but i don't see the next button.

It looks like that the resize function is not triggered and the div is too large because of  `w-full` class. 

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

this PR calculate the width of the bottom toolbar by subtract the width of the left nav (54px) and left sidebar (225px)

As before

it resolves #4418 

### How to test this PR?

1. run the onboarding on mac. after the podman installation wizard finishes and you are in the step where you can set the autostart setting you should see the next/cancel buttons correctly

### DIFF
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -91,7 +91,9 @@ export let meta: TinroRouteMeta;
 </script>
 
 <svelte:window />
-<nav class="group w-[54px] min-w-[54px] flex flex-col justify-between hover:overflow-y-none" aria-label="AppNavigation">
+<nav
+  class="group w-leftnavbar min-w-leftnavbar flex flex-col justify-between hover:overflow-y-none"
+  aria-label="AppNavigation">
   <NavItem href="/" tooltip="Dashboard" bind:meta="{meta}">
     <div class="flex items-center w-full h-full">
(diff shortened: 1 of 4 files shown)


benoitf (commented):
My worry is that if someone increase/decrease the width of navbar or settings sidebar I'm pretty sure we'll omit the change here

so, 

could you add in `tailwind.config.js` 

```js
 width: {
      'navbar': '54px',
       ....
    },
    minWidth: {
      'navbar': '54px',
       ....
    },
```

and then use `w-navbar` / `w-min-navbar` where we define the bars

and for onboarding use theme(value) to get the constants:

```
 w-[calc(<something>-theme(width.navbar)-theme(...)]
```

benoitf (approved):
image

Issue #4214, opened 2023-10-05
Log extension errors with action


deboer-tim:
### Is your feature request related to a problem? Please describe

See #3788 for background. Several of our extensions (e.g. kind, compose) currently send telemetry events for actions, and then a separate error if those actions fail. This makes it really hard to track failure rates, or understand what the failure was caused by if some of the context was in the 'action event'.

We should save the telemetry error logging for asynchronous events, and only send one event for an actions (with different/additional properties when there is an error).

### Describe the solution you'd like

Switch kind and compose extensions to each send a single event per action, with additional properties on error.

We should start to set a pattern for what properties we use for CLI commands (e.g. error.code, error.message) vs REST calls (e.g. http.status).

Execution time property should be added.

Bonus points if for each of these we can do some minimal cleanup or figuring out which properties are going to be useful for understanding why it fails in some cases for that specific command - e.g. are we tracking the version of the cli? is there another useful HTTP header?

deboer-tim:
Kind and Compose error cases were updated and released, and telemetry added for podman machine. We're starting to see some data coming in, but too soon to tell if there's anything else we should do. Guidance is being added to a USAGE_LOG.md via PR #4619. Since this addresses all we can do on this topic for now I'm closing it, and will create a new feature if we identify more work.

PR #4427, opened 2023-10-19
fix: log kind cluster creation failure with usage

deboer-tim:
### What does this PR do?

Minimal change to send one event when we create a kind cluster instead of 1 event + different event on failure.

Added the duration while I was at it, but nothing else. There is likely more work to do here to classify failures or add more information to tell what is different about failure cases, but this will solve one core problem that it isn't easy to correlate failure with the usage.

N/A

Fixes part of #4214.

### How to test this PR?

Make a kind cluster, look for dev telemetry in Segment with duration (and maybe error).

### DIFF
--- a/extensions/kind/src/create-cluster.ts
+++ b/extensions/kind/src/create-cluster.ts
@@ -174,6 +174,7 @@ export async function createCluster(
       await setupIngressController(clusterName);
     }
   } catch (error) {
+    telemetryOptions.error = error;
     let errorMessage: string;
 
     if (typeof error === 'object' && 'message' in error) {
@@ -181,8 +182,6 @@ export async function createCluster(
     } else if (typeof error === 'string') {
       errorMessage = error;
     }
-
-    telemetryOptions.error = errorMessage;
 
     throw new Error(`Failed to create kind cluster. ${errorMessage}`);
   } finally {
--- a/extensions/kind/src/create-cluster.ts
+++ b/extensions/kind/src/create-cluster.ts
@@ -157,14 +157,22 @@ export async function createCluster(
   // update PATH to include kind
   env.PATH = getKindPath();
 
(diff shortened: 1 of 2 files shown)


extensions/kind/src/create-cluster.ts
```diff
@@ -174,16 +182,15 @@ export async function createCluster(
...
-    telemetryLogger.logError('createCluster', {
-      provider,
-      httpHostPort,
-      httpsHostPort,
-      ingressController,
-      error: errorMessage,
-      stdErr: errorMessage,
-    });
+    telemetryOptions.error = errorMessage;
+    telemetryOptions.stdErr = errorMessage;
```
benoitf: not sure why we have two fields with the same object ?

extensions/kind/src/create-cluster.ts
```diff
@@ -174,16 +182,14 @@ export async function createCluster(
...
-    telemetryLogger.logError('createCluster', {
-      provider,
-      httpHostPort,
-      httpsHostPort,
-      ingressController,
-      error: errorMessage,
-      stdErr: errorMessage,
-    });
+    telemetryOptions.error = errorMessage;
+
```
benoitf: in compose, we're logging the error object (that will include exit code, etc)
https://github.com/containers/podman-desktop/pull/4407/files

here just the message. is it wanted ?

extensions/kind/src/create-cluster.ts
deboer-tim: Solely because 'it was there before' :) and I was trying to keep the PR to the smallest change. Fixed to only add one attribute (and change to error as per next comment).

extensions/kind/src/create-cluster.ts
deboer-tim: Fixed to use the same pattern as compose here.

benoitf (approved):

PR #4446, opened 2023-10-20
docs: modify pull request template to document release notes section

jeffmaury:
### What does this PR do?

Update the pull request template to document the release notes section

N/A

N/A

### How to test this PR?

Only review the template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,10 +7,10 @@ explaining what is doing this PR -->
 
 ### Release notes
 
+<!-- Add text between the tags below to have it automatically included
+in a release note draft -->
+
 // release-notes
-Please insert markdown section between those tags to get them automatically included
-into a release notes draft. As this section is between those tags, remove it if you
-don't want to get included into the release notes
 // release-notes
 
 ### What issues does this PR fix or reference?
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,12 +2,20 @@
 
 ### Screenshot/screencast of this PR
 
-<!-- Please include a screenshot or a screencast 
+<!-- Please include a screenshot or a screencast
 explaining what is doing this PR -->
+
+### Release notes
+
+// release-notes
+Please insert markdown section between those tags to get them automatically included
+into a release notes draft. As this section is between those tags, remove it if you
+don't want to get included into the release notes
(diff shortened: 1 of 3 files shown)


deboer-tim:
I'm mildly worried about leaving a section that will get included if the person does nothing. Does the release note tool ignore empty sections, e.g.:
```
### Release notes

<!-- Add text between the tags below to have it automatically included
in a release note draft -->
// release-notes
// release-notes
```

or comments?
```
// release-notes
<!--- Insert markdown section between these tags to have it automatically included
in a release notes draft --->
// release-notes
```

jeffmaury:
> I'm mildly worried about leaving a section that will get included if the person does nothing. Does the release note tool ignore empty sections, e.g.:
> 
> ```
> ### Release notes
> 
> 
> // release-notes
> // release-notes
> ```
> 
> or comments?
> 
> ```
> // release-notes
> 
> // release-notes
> ```

I think it's better that the current situation that no one uses the generator tool and every releaser pings me about nothing being generated. And PR author are supposed to review the PR description (like screen cast,...). Will update the generator to handle empty content in this PR

benoitf:
could we assume that by default the generator uses the title of the pull request and the text inside the 

```
What does this PR do?

```

section

and have a work in progress PR or a view of what would look like the current PR so we can just adjust content of the PR without the need of adding an extra section

if you want to select / promote a given PR we could just give an extra label 'release-notes-included'

benoitf:
it's because having a label is an easy filter rather than a specific body description

with a label I can use github tools for displaying quickly what issues or pr will be included or not, remove it, add it

here, without running a specific tool we can't leverage GitHub tools/views

jeffmaury:
> could we assume that by default the generator uses the title of the pull request and the text inside the
> 
> ```
> What does this PR do?
> ```
> 
> section
> 
> and have a work in progress PR or a view of what would look like the current PR so we can just adjust content of the PR without the need of adding an extra section
> 
> if you want to select / promote a given PR we could just give an extra label 'release-notes-included'

I think this needs to be dicussed because when the tool was done there was a request for having a section specific to the release notes that is different from the description.

deboer-tim:
> I think this needs to be discussed because when the tool was done there was a request for having a section specific to the release notes that is different from the description.

I think the reasoning was that often the PR description is too low-level. e.g. description might talk about why certain files changed and include before/after images with the change circled in red, but the release note could be one line and a pretty screenshot.

Personally I'm even fine if description was picked up by default, but would want something like `### Release notes` to allow you to override.

PR #4457, opened 2023-10-20
chore: add close button to troubleshooting and help page

### LINKED ISSUE #2878: The Help screen does not the option to "x" out of it
### Bug description

After clicking on the help option in the bottom right I was able to view the information however I was not presented with the option to close it out.  I had to make another selection to remove it from view rather than x'ing it out.

### Operating system

macOS

### Installation Method

### Version

1.1.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

jannikbertram:
### What does this PR do?

- Add an X button to pages which don't have breadcrumbs

image

image

Closes #2878 

### How to test this PR?

Visit the "Troubleshooting" and "Help" page

### DIFF
--- a/packages/renderer/src/lib/help/HelpPage.svelte
+++ b/packages/renderer/src/lib/help/HelpPage.svelte
@@ -17,7 +17,7 @@ $: contributedLinks = $providerInfos
   }, new Map<string, ProviderLinks[]>());
 </script>
 
-<FormPage title="Help" showBreadcrumb="{false}">
+<FormPage title="Help" showBreadcrumb="{true}">
   <svelte:fragment slot="icon">
     <i class="fas fa-question-circle fa-2x" aria-hidden="true"></i>
   </svelte:fragment>
--- a/packages/renderer/src/lib/help/HelpPage.svelte
+++ b/packages/renderer/src/lib/help/HelpPage.svelte
@@ -17,7 +17,7 @@ $: contributedLinks = $providerInfos
   }, new Map<string, ProviderLinks[]>());
 </script>
 
-<FormPage title="Help" showBreadcrumb="{true}">
+<FormPage title="Help" showBreadcrumb="{false}" showClose="{true}">
   <svelte:fragment slot="icon">
     <i class="fas fa-question-circle fa-2x" aria-hidden="true"></i>
   </svelte:fragment>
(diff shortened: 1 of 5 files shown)


deboer-tim:
@mairin The original design was no breadcrumbs on these two pages, but it sounds like it's throwing people off. Thoughts?

@jannikbertram If we do go down this route we should remove the property entirely instead of setting it true everywhere, and likely rename Home to Dashboard since I assume that's where it goes 'back' to.

jannikbertram:
Another option would be to only include the ❌ icon and remove the breadcrumbs. 
I just thought I propose this obvious approach first. 
Just let me know which design is preferred

mairin:
@deboer-tim @jannikbertram Help and Troubleshooting didn't have breadcrumbs because they sort of belong to the status bar... Neither has a presence on the dashboard (not that they shouldn't, they just don't) so semantically linking back to the Dashboard as a parent in a breadcrumb feels weird.

I do think the X in the upper right is a good idea and may have even been in the original design. I really like how you have it going to the last previously-viewed screen when the user hits the X.

I would say let's try a light touch first and just do the X, and try to side step the 🤯-ness of the breadcrumbs for now?

jannikbertram:
Thanks for clarifying this @mairin. I added a new commit and updated the PR description.

This is how it looks when you use the X button together with actions and breadcrumbs:

image

I'm not sure if it should be possible like this. I feel like this really is about small details, so please let me know if you want this to behave differently. I can keep updating this until we find a consistent solution

deboer-tim:
That looks odd to me. I don't think this PR should change the location of the close button on all the other pages (it should remain on the 'breadcrumb bar'), only add a close button to the ones that don't have breadcrumbs.

jannikbertram:
I updated the PR again

> only add a close button to the ones that don't have breadcrumbs.

now it simply does this ☝️

mairin:
@jannikbertram this looks great to me, just tested it out :)

jeffmaury (changes requested):
Fix unit tests

jannikbertram:
@jeffmaury can you trigger the unit tests again? They should be working now

benoitf (approved):
thanks @jannikbertram for your contribution

deboer-tim (approved):
LGTM. @jeffmaury, assuming you're 👍🏼 now.

jeffmaury (approved):
LGTM. Thanks @jannikbertram for this

PR #4459, opened 2023-10-20
feat: use directory name as default container image name

jannikbertram:
### What does this PR do?

When creating a new container, the container image name currently defaults to `my-custom-image`. 
With this PR, the container image defaults to the directory name of the chose container file.

image

Fix for https://github.com/containers/podman-desktop/issues/2713

Inspired by https://github.com/containers/podman-desktop/pull/2897 (added support for Windows style paths)

### How to test this PR?

Create a custom container, choose a file and observe the image name

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -122,10 +122,20 @@ async function getContainerfileLocation() {
   const result = await window.openFileDialog('Select Containerfile to build');
   if (!result.canceled && result.filePaths.length === 1) {
     containerFilePath = result.filePaths[0];
+    hasInvalidFields = false;
+
+    // normalize the file path to posix style
+    // eslint-disable-next-line no-useless-escape
+    const posixContainerDirPath = containerFilePath.replace(/\\/g, '/').replace(/\/[^\/]*$/, '');
+    const containerDirName = posixContainerDirPath.split('/').slice(-1)[0];
+    if (containerDirName) {
+      // use the directory as default image name
+      containerImageName = containerDirName;
+    }
+
     if (!containerBuildContextDirectory) {
       // select the parent directory of the file as default
-      // eslint-disable-next-line no-useless-escape
-      containerBuildContextDirectory = containerFilePath.replace(/\\/g, '/').replace(/\/[^\/]*$/, '');
+      containerBuildContextDirectory = posixContainerDirPath;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


jannikbertram:
I'm going to add some unit tests for this as well.

✅ done

jannikbertram:
will take a look at the unit test failures

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -155,7 +165,8 @@ async function getContainerBuildContextDirectory() {
               class="w-full p-2 mr-3 outline-none text-sm bg-charcoal-600 rounded-sm text-gray-700 placeholder-gray-700"
               required />
 
-            <Button on:click="{() => getContainerfileLocation()}">Browse...</Button>
+            <Button dataTestId="containerFilePathBrowse" on:click="{() => getContainerfileLocation()}"
```
lstocchi: you could use the aria-label in the button component without adding a dataTestId.
Like
`<Button aria-label="containerFilePathBrowse"  ...`
then in the test
`screen.getByRole('button', { name: 'containerFilePathBrowse' `

lstocchi (commented):
Thanks @jannikbertram . Could you rebase it?

And sorry for the super slow response 😞 🙏

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
jannikbertram: Nice, I didn't know this way of reference a component. Updated the code accordingly and rebased :)

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -122,10 +122,20 @@ async function getContainerfileLocation() {
   const result = await window.openFileDialog('Select Containerfile to build');
   if (!result.canceled && result.filePaths.length === 1) {
     containerFilePath = result.filePaths[0];
+    hasInvalidFields = false;
+
+    // normalize the file path to posix style
+    // eslint-disable-next-line no-useless-escape
+    const posixContainerDirPath = containerFilePath.replace(/\\/g, '/').replace(/\/[^\/]*$/, '');
+    const containerDirName = posixContainerDirPath.split('/').slice(-1)[0];
+    if (containerDirName) {
```
benoitf, line 131: should we set the image name only if it's not already defined ?

I see pro/cons here

deboer-tim: +1 from me, but I would change it slightly to 'was edited by the user'. i.e. if the user hasn't edited the field and it was auto-generated, then they pick a different path, we should auto-generate it again. But if the user typed in the name, or edited one we auto-generated, we shouldn't change it again. It is probably safe enough to say 'focussed on the field' = 'user edited' for coding simplicity.

I've seen this in other products and it strikes a good balance between being very helpful but not overwriting anything the user has done.

jannikbertram: hey guys, thanks for the input. Will try to implement it like you described when I find the time 👍

Issue #4365, opened 2023-10-16
Reusable table component


deboer-tim:
### Is your enhancement related to a problem? Please describe

We've discussed future improvements to our 'list pages' like Containers and Pods to be able to:
- be sortable
- allow grouping by container engine/Kubernetes context
- allow the user to pick a different subset of columns, or completely customize the columns or ordering
- click anywhere to go to Details page vs current single column
etc.

Currently, each of these pages has a hardcoded set of columns, and implementing any of these would require changes to every page. As we discuss adding more pages for Kubernetes objects it would be a good time to make sure we don't extend this debt.

### Describe the solution you'd like

Create a basic table component that covers the current column layout requirements of any of these pages. Each page just says 'I have X items and can provide Y sets of columns for each one' and the component is responsible for iterating over the items to create the layout.

For example the list might say it has a name column that should be left justified, a size column that should be right justified, provide 'renderers' for both, and then 5 objects. The component would iterate over these, create the header, trs, tds, and call the renderers at the appropriate places.

This initial component doesn't need to support any of the new features listed, it just needs to allow us to do that in the future in one place. For instance, hiding a column would just mean it leaves out the header and doesn't use the renderer for it.

feloy:
Did we consider using a specialized library, for example AG grid Community (MIT license)? 
https://www.ag-grid.com/
https://ag-grid-svelte.michael.kim/guide/overview/

slemeur:
That's a great suggestion!

jeffmaury:
Good candidate for POC during DevTools week

deboer-tim:
FWIW, I looked at the top >10 tables that came up when I googled variations of 'svelte table component'. Most were very basic, and even the more advanced ones didn't support most of our current requirements (e.g. custom cell rendering w/ actions, custom padding, expandable sections, etc), let alone future requirements like sorting (correctly) or customizing columns, etc.

AG Grid definitely is feature rich and might be an option, but that is a bigger change along with the implications of that - dependencies, applying our UI theme, etc.

@slemeur , @mairin I think we need to make a choice. Continuing down the path of this PR it is totally doable to implement all of the future designs we've seen so far (e.g. sorting, different column sets) but it would continue to be custom and one PR at a time, so there's a limit to how far we'd go. Switching to an existing component would be a bigger initial change and might restrict us in some ways, but hopefully would give us access to a longer list of features. So, what else might we need our tables to do in a year?

[...]

PR #4470, opened 2023-10-23
Refactor containers list component

feloy:
### What does this PR do?

Refactor the Containers List component, so we can later insert Tabs inside it, to show running/stopped containers

Related to https://github.com/containers/podman-desktop/issues/707

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -93,7 +93,7 @@ window.events?.receive('display-troubleshooting', () => {
         <Route path="/" breadcrumb="Dashboard Page">
           <DashboardPage />
         </Route>
-        <Route path="/containers" breadcrumb="Containers" navigationHint="root">
+        <Route path="/containers/*" breadcrumb="Containers" navigationHint="root">
           <ContainerList searchTerm="{meta.query.filter || ''}" />
         </Route>
         <Route path="/containers/:id/*" breadcrumb="Container Details" let:meta navigationHint="details">
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -115,7 +115,7 @@ export let meta: TinroRouteMeta;
         ></svg>
     </div>
   </NavItem>
-  <NavItem href="/containers" tooltip="Containers{containerCount}" ariaLabel="Containers" bind:meta="{meta}">
+  <NavItem href="/containers/all" tooltip="Containers{containerCount}" ariaLabel="Containers" bind:meta="{meta}">
     <ContainerIcon size="24" />
   </NavItem>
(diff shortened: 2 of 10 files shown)


deboer-tim:
I am worried that this will make the Containers page more complex, when we will almost certainly be asked to do similar things to other pages (e.g. running/stopped pods, or used vs unused images or volumes). I think we need to figure out how to extract some of the commonality and implement once vs doing it in each page. I don't know if it's possible yet, but see issue #4365 for some of my thoughts.

feloy:
> I am worried that this will make the Containers page more complex, when we will almost certainly be asked to do similar things to other pages (e.g. running/stopped pods, or used vs unused images or volumes). I think we need to figure out how to extract some of the commonality and implement once vs doing it in each page. I don't know if it's possible yet, but see issue #4365 for some of my thoughts.

Can it help if I refactor the other List components (Pods, Images, ...) the same way, so it will be easier to detect common patterns between them?

benoitf (commented):
from my understanding it will help if when splitting into components, if it also becomes less "container" centric so component could be used elsewhere

deboer-tim:
Ditto. Separating each of the existing components into 4 just gives us 4x the number of files ;-) unless we see a common part we could extract & share to reduce the complexity.

feloy:
By splitting into 4 components each List, I expect the related sub-components will be easier to merge together, for example the ContainerListContent and ImageListContent, as @deboer-tim describes in #4365.

feloy:
To be sure we are talking about the same thing, I have pushed a commit adding the Tabs (it can be removed and cherry-picked to another PR when we agree, to have separate PRs for fixes / features). 

This commit shows IMO the necessity to move the ContainerListContent to a separate component, to be able to include it in different tabs, with differents containersLists as parameter.

The first 2 refactorings (TopActions and BottomActions) are not necessary, I could remove them if you think it is not worth it doing them.

deboer-tim:
After a very rough look through the changes:
- The top vs bottom actions I don't think is really saving us much, to me we're losing more through the amount of code duplication.
- Personally, I'm not convinced tabs are the right approach here. First, because we don't use tabs for filtering elsewhere (ie I think we should use a different UI control), and second b/c we're essentially creating three separate container lists instead of having one container list with a filter.
- To me this is deferring the issue in #4365... it doesn't make it better or worse, we're just doing one more thing for containers that would have to be refactored out or done twice when we get the next request.

benoitf:
@deboer-tim  tabs is what has been designed: https://github.com/containers/podman-desktop/issues/707#issuecomment-1773952917

feloy:
Closing in favor of https://github.com/containers/podman-desktop/pull/4483

I think I have a better way to use the Tabs: in #4483, the tabs are ONLY used as a selector to select the running/stopped filter. This way, we can change easily from using tab to using another kind of selector (radio button, ...).

PR #4471, opened 2023-10-23
fix: allow editing of build containerfile

### LINKED ISSUE #3824: Make input fields that open system native file explorer to pick up a file or path to accept actual user input and add browse button

It is not possible for user to insert a path into input fields that open only file explorer dialog on `onClick` event. It is also hard to work with these dialogs when writing e2e tests.

Example:
Screenshot_20230907_162534

Make such fields accepting text from user input, also please add Browse button so the explorer can be opened and option preserved.

For example this might it look like (although the input here is set to readonly):
Screenshot_20230907_162642

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Changes the build image form from a readonly text field that you click on to browser, to a more common editable text field with a Browse button. Likewise the build directory was hidden until you entered a containerfile before, now it is always visible and uses the same text field + browse button.

This change makes the form use a more common pattern, and allows e2e tests to more easily enter text in either field.

The drawback to this approach is that there is no more validation than before - the build button enables whenever there is text in both fields. This happens automatically when you use the Browse button like before, but if you edit or copy/paste and have a typo you won't find out until you click Build. We could add more validation in the future, but IMHO the better usability/form layout is enough of a benefit to make this change in the meantime.

Screenshot 2023-10-23 at 9 52 25 AM

Fixes #3824.

### How to test this PR?

Just go to Images > Build Image.

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.spec.ts
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.spec.ts
@@ -88,9 +88,12 @@ test('Expect Build button is enabled', async () => {
   render(BuildImageFromContainerfile, {});
 
   const containerFilePath = screen.getByRole('textbox', { name: 'Containerfile Path' });
-
   expect(containerFilePath).toBeInTheDocument();
-  await userEvent.click(containerFilePath);
+  await userEvent.type(containerFilePath, '/somepath/containerfile');
+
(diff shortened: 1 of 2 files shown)


benoitf:
hello, I'm fine with the change but I think we need to postpone the merge to be after the release

benoitf (approved):
probably need form component (not asked by this PR) but we do have several ways to submit files in the code base

PR #4483, opened 2023-10-24
feat: tabs in containers list to filter running/stopped

### LINKED ISSUE #707: "Only show running containers" filter for containers

Docker desktop has a feature where you can choose to filter the list of containers to show only the running ones. If it has none, it will show "No running containers". I'd like to see this feature for Podman desktop.

Add a toggle to the right of the search bar in the containers tab to allow users to filter on the running status. Only show running containers in the list.

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

This PR adds tabs to the Containers List page, to filter on Running / Stopped containers.

When the user moves to a Running / Stopped tab, the filter "is:running" / "is:stopped" is added to the filter field.

When the user adds / removes the "is:running" / "is:stopped" text from the filter field, the UI changes the tab accordingly.

Tje filter field content is kept when the user swicthes from one tab to another (the filters is:running / is:stopped being modified, the rest of the filter being kept).

Groups of containers:
- After filtering containers, only groups with remaining containers are displayed
- The number of all containers (before filtering) for a group is displayed, and if some containers are filtered, the number of filtered containers is also displayed. For example `2 containers (1 filtered)` indicates that the group contains 2 containers, but only 1 is displayed because the other one is filtered.

Fixes #707

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -94,7 +94,10 @@ window.events?.receive('display-troubleshooting', () => {
           <DashboardPage />
         </Route>
         <Route path="/containers/*" firstmatch>
-          <Route path="/list/*" breadcrumb="Containers" navigationHint="root">
+          <Route path="/">
+            <ContainerList searchTerm="{meta.query.filter || ''}" />
+          </Route>
+          <Route path="/list/:id" breadcrumb="Containers" navigationHint="root">
             <ContainerList searchTerm="{meta.query.filter || ''}" />
           </Route>
           <Route path="/:id/*" breadcrumb="Container Details" let:meta navigationHint="details">
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -95,10 +95,10 @@ window.events?.receive('display-troubleshooting', () => {
         </Route>
         <Route path="/containers/*" firstmatch>
           <Route path="/">
-            <ContainerList searchTerm="{meta.query.filter || ''}" />
+            <ContainerList filter="{meta.query.filter || ''}" />
           </Route>
           <Route path="/list/:id" breadcrumb="Containers" navigationHint="root">
-            <ContainerList searchTerm="{meta.query.filter || ''}" />
+            <ContainerList filter="{meta.query.filter || ''}" />
           </Route>
           <Route path="/:id/*" breadcrumb="Container Details" let:meta navigationHint="details">
             <ContainerDetails containerID="{meta.params.id}" />
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -93,7 +93,7 @@ window.events?.receive('display-troubleshooting', () => {
         <Route path="/" breadcrumb="Dashboard Page">
           <DashboardPage />
         </Route>
-        <Route path="/containers/*" breadcrumb="Containers" navigationHint="root">
+        <Route path="/containerslist/*" breadcrumb="Containers" navigationHint="root">
           <ContainerList searchTerm="{meta.query.filter || ''}" />
         </Route>
         <Route path="/containers/:id/*" breadcrumb="Container Details" let:meta navigationHint="details">
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -93,7 +93,7 @@ window.events?.receive('display-troubleshooting', () => {
         <Route path="/" breadcrumb="Dashboard Page">
           <DashboardPage />
         </Route>
-        <Route path="/containers" breadcrumb="Containers" navigationHint="root">
+        <Route path="/containers/*" breadcrumb="Containers" navigationHint="root">
           <ContainerList searchTerm="{meta.query.filter || ''}" />
         </Route>
         <Route path="/containers/:id/*" breadcrumb="Container Details" let:meta navigationHint="details">
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -93,13 +93,14 @@ window.events?.receive('display-troubleshooting', () => {
         <Route path="/" breadcrumb="Dashboard Page">
           <DashboardPage />
         </Route>
-        <Route path="/containerslist/*" breadcrumb="Containers" navigationHint="root">
-          <ContainerList searchTerm="{meta.query.filter || ''}" />
+        <Route path="/containers/*" firstmatch>
+          <Route path="/list/*" breadcrumb="Containers" navigationHint="root">
+            <ContainerList searchTerm="{meta.query.filter || ''}" />
+          </Route>
+          <Route path="/:id/*" breadcrumb="Container Details" let:meta navigationHint="details">
+            <ContainerDetails containerID="{meta.params.id}" />
+          </Route>
         </Route>
-        <Route path="/containers/:id/*" breadcrumb="Container Details" let:meta navigationHint="details">
-          <ContainerDetails containerID="{meta.params.id}" />
-        </Route>
-
         <Route path="/kube/play" breadcrumb="Play Kubernetes YAML">
           <KubePlayYAML />
         </Route>
--- a/packages/renderer/src/lib/ContainerList.svelte
+++ b/packages/renderer/src/lib/ContainerList.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 import { onDestroy, onMount } from 'svelte';
-import { filtered, searchPattern, containersInfos, runningFilter, stoppedFilter } from '../stores/containers';
+import { containersInfos } from '../stores/containers';
 import { viewsContributions } from '../stores/views';
 import { context } from '../stores/context';
 
@@ -41,6 +41,7 @@ import Button from './ui/Button.svelte';
 import StateChange from './ui/StateChange.svelte';
 import SolidPodIcon from './images/SolidPodIcon.svelte';
 import Tab from './ui/Tab.svelte';
+import { findMatchInLeaves } from '../stores/search-util';
 
 const containerUtils = new ContainerUtils();
 let openChoiceModal = false;
@@ -52,7 +53,16 @@ let viewContributions: ViewInfoUI[] = [];
 let globalContext: ContextUI;
 let containersInfo: ContainerInfo[] = [];
 export let searchTerm = '';
-$: searchPattern.set(searchTerm);
+let runningFilter = false;
+let stoppedFilter = false;
+
+interface FilterOptions {
+  searchTerm: string;
+  runningFilter: boolean;
+  stoppedFilter: boolean;
+}
+
+$: updateContainers($containersInfos, $context, $viewsContributions, { searchTerm, runningFilter, stoppedFilter });
 
 function fromExistingImage(): void {
   openChoiceModal = false;
@@ -60,12 +70,6 @@ function fromExistingImage(): void {
 }
 
 let multipleEngines = false;
-
-router.subscribe(route => {
-  const f = route.path.substring(route.path.lastIndexOf('/') + 1);
-  runningFilter.set(f === 'running');
-  stoppedFilter.set(f === 'stopped');
-});
 
 $: providerConnections = $providerInfos
   .map(provider => provider.containerConnections)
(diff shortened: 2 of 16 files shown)


packages/renderer/src/stores/containers.ts
```diff
@@ -65,7 +65,21 @@ const containersEventStore = new EventStore<ContainerInfo[]>(
...
+      .filter(containerInfo => findMatchInLeaves(containerInfo, $searchPattern.toLowerCase()))
+      .filter(containerInfo => {
+        if ($runningFilter) {
+          return containerInfo.State.toUpperCase() === 'RUNNING';
+        }
+        if ($stoppedFilter) {
+          return containerInfo.State.toUpperCase() !== 'RUNNING';
+        }
+        return true;
+      }),
```
benoitf: to me it's strange that the state of the filter is in the store ?
shouldn't it belong to container list display thing ?

feloy: Is it different from the searchPattern, which belongs to the state?

benoitf: hello @feloy, you're correct no it's not so different from the search pattern but I'm not sure search pattern is at the right place as well ?!

I mean, do we consider the stores as the only data being exposed in all the UIs of Podman Desktop or it's just a representation state of the current state on the server

for example, is that if I order by Age in container list I want that in another view containers will be sorted in the same way or it depends on the view

so, is that filtering is part of the store or the view

it would be ok for me if there is a allContainers (what we have) a runningContainers and a stoppedContainers objects
but do we want to display all running containers in more than one location ?

feloy: Yes, I also think that the state should contain the representation of what exists in the machine, and this information should be transformed depending on where it is displayed.

For the moment, I don't think the list of containers is displayed in several places, but I'm pretty sure we will want to add some counters somewhere in the UI, and the state won't be helpful for that.

I can look for moving the searchPattern and the running state out of the state.

benoitf: you could also for the running/all/stopped, have a filter that is being applied on the route

all : noop filter, running = filter out what is not running, stopped = keep what is not running

feloy: > you could also for the running/all/stopped, have a filter that is being applied on the route
> 
> all : noop filter, running = filter out what is not running, stopped = keep what is not running

Hi @benoitf sorry I don't understand completely what you mean. Could you elaborate please?

benoitf: I mean that we could have different fields in the store where it represents different view (not saying that we must)

For now we have a field with all containers (containerInfos)

But we could have in the container store other objects representing different states like all the running containers and another one with all non-running containers.

Sso it's still accurate data, it does not depends on a user filter, etc

feloy: ok, got it. It would be interesting if these infos were used in different places. I would prefer to have them located in the component for now, and move them to the state when we use it in different places.

packages/renderer/src/App.svelte
```diff
@@ -93,7 +93,7 @@ window.events?.receive('display-troubleshooting', () => {
         <Route path="/" breadcrumb="Dashboard Page">
           <DashboardPage />
         </Route>
-        <Route path="/containers" breadcrumb="Containers" navigationHint="root">
+        <Route path="/containerslist/*" breadcrumb="Containers" navigationHint="root">
```
benoitf: could you explain why we need to have a dedicated url for containerslist rather than having it part of /containers/ ? or `/containers/list` for example

benoitf (commented):
I've tested it

https://github.com/containers/podman-desktop/assets/436777/b6e36775-e0c7-44f8-aab8-76d2232208b2

it looks that some space is missing at the top between title and the tabs

I have a question around the tabs as well

I see that some mockups have Running tab first, some have All tab first so I'm not sure which one to follow

cc @mairin (what is the expected tab order ? All, Running, Stopped or Running, Stopped, All


image

benoitf:
also it looks like that the filtering is done using `is:running` has a keyword in the search field

mairin:
@benoitf @feloy Good spot on the mockup inconsistency. When making these way back, I had weighed both options and decided starting with "All containers" being the first tab / default was the best approach, because you might get in a few situations where you would miss things if it wasn't. For example, you have a pod with a few stopped containers in it and not realize how many you had total because the running tab filters them out.

packages/renderer/src/App.svelte
feloy: There is another route `/containers/:id`, and it seems it will have priority over `/containers/list`, or `/containers/anything`.

I tried to move the `/containers/*` before `/containers/:id` in the list of Reoutes, but the priority didn't change.

Is there a way to change this priority?

benoitf: based on https://github.com/AlexxNB/tinro#show-first-matched-nested-route-only

looks like `/containers/list`. would work but maybe you want to have `/containers/list/*` and `/containers/details/...`

might be better if all routes for containers are in a single route prefix

feloy:
> also it looks like that the filtering is done using `is:running` has a keyword in the search field

@mairin @benoitf 
- do we want to add this `is:running` text in the search field when we switch to the `Running Containers` tab?
-  what would be the text for `Stopped containers` tab? `is:stopped`, `is:not:running`, `is:!running`, etc?
- what should happen if the user manually adds/removes this text from the text search field?

mairin:
@feloy My thought when mocking this up was to make search filters based on the status, yeh, and have them prepopulate based on the tab to kind of teach the user in context if they prefer keyboard how to do it. It might make sense to match whatever podman does on the command line if it provides filters. If not, I'd match the filter names to the names of the tabs so it's sort of self-documented.

https://www.redhat.com/sysadmin/container-information-podman

Looks like for podman it's:

`podman ps` - lists only running containers
`podman ps --all` - lists all containers

Since we can't really be exactly analogous with podman then I'd do:

- nothing: show all containers
- `is:running` only running containers
- `is:stopped` only stopped containers

If the user clears manually with the keyboard, then it should go back to the all tab maybe? Does that feel weird?

deboer-tim:
I noticed in Florent's video that the compose group is 'stopped' in the first tab, but then appears in the Running tab as 'running' and again as 'stopped' in the Stopped tab. The status of the overall compose group isn't changing and this is giving incorrect information. Its status shouldn't change between the tabs.

Likewise, the compose group says '2 containers' in the first tab, then '1 container' in the other two. The compose group always has 2 containers and should say so... but then there's the inevitable question of why it is saying 2 but I only see one. I think there needs to be something to indicate that there are more/hidden/filtered containers (which will also help explain cases like the first paragraph where the state 'does not make sense' based on the filtered info you can see). e.g. something like '2 containers (1 stopped)'.

Everything ditto for Pods of course.

deboer-tim:
> ...
> is:running only running containers
> is:stopped only stopped containers
> If the user clears manually with the keyboard, then it should go back to the all tab maybe? Does that feel weird?

IMHO Github PR/issue filtering is a very good example of what to do here. It's pretty minimal and natural filtering, and easy to type. The header is a bit different, but switching between tabs is similar to the Open / Closed links.

feloy:
> > ...
> > is:running only running containers
> > is:stopped only stopped containers
> > If the user clears manually with the keyboard, then it should go back to the all tab maybe? Does that feel weird?
> 
> IMHO Github PR/issue filtering is a very good example of what to do here. It's pretty minimal and natural filtering, and easy to type. The header is a bit different, but switching between tabs is similar to the Open / Closed links.

As discussed in UX Check In, I'll move the search field above the tabs, to have the same experience as with GH

deboer-tim:
> '2 containers (1 stopped)'.

I realized a much better way to do this would be something like '4 containers (1 filtered)' - then the text doesn't need to change between tabs and it would work if we do different kinds of filtering in the future.

feloy:
> > > ...
> > > is:running only running containers
> > > is:stopped only stopped containers
> > > If the user clears manually with the keyboard, then it should go back to the all tab maybe? Does that feel weird?
> > 
> > 
> > IMHO Github PR/issue filtering is a very good example of what to do here. It's pretty minimal and natural filtering, and easy to type. The header is a bit different, but switching between tabs is similar to the Open / Closed links.
> 
> As discussed in UX Check In, I'll move the search field above the tabs, to have the same experience as with GH

This has been implemented in latest commits, feel free to test / review

feloy:
> > '2 containers (1 stopped)'.
> 
> I realized a much better way to do this would be something like '4 containers (1 filtered)' - then the text doesn't need to change between tabs and it would work if we do different kinds of filtering in the future.

I have made some changes to reflect this. Feel free to test / review

packages/renderer/src/lib/ContainerList.svelte
```diff
@@ -40,18 +40,48 @@ import type { ContextUI } from './context/context';
...
+let runningFilter: boolean;
+let stoppedFilter: boolean;
+
+$: filterObj = new Filter(decodeURI(filter));
+
+$: checkFilter(filterObj.rawFilter);
+
+function checkFilter(_: string): void {
+  if (filterObj.isRunning() && !runningFilter) {
+    router.goto(`/containers/list/running?filter=${filterObj.setState(true, false)}`);
```
lstocchi: reading all these setState(true/false) is not really meaningful without looking at the function itself. I think it could be made more readable by using some specific function (`setRunningState`), or as you have the information already in the url "running/stopped" you can make a function that create it (`createRunningURL(/containers/list/)`) .

packages/renderer/src/lib/ContainerList.svelte
```diff
@@ -40,18 +40,48 @@ import type { ContextUI } from './context/context';
...
+
+let tab: string;
+
+export let filter = '';
+let runningFilter: boolean;
+let stoppedFilter: boolean;
+
+$: filterObj = new Filter(decodeURI(filter));
+
+$: checkFilter(filterObj.rawFilter);
```
lstocchi: if you don't use `filterObj.rawFilter` inside checkFilter maybe there is no reason to add it as a arg.
To trigger the call i think it also work something like `filterObj.rawFilter && checkfilter()`

lstocchi (commented):
Tried on Windows, works fine!! GJ 🚀

mairin:
Tried this on Fedora 38 this morning - this is *so freaking awesome*!!!

slemeur:
Tried on Mac OS! Works fine and is absolutely a pleasure to use :) ! 
Well done @feloy !

slemeur:
and @mairin for the UX!

deboer-tim (commented):
Tested, agree with previous comments that the behaviour is quite nice.

Is there any reason the URL to the page has to change? I was expecting the 'all tab' to remain the default at /containers and the respective tabs would highlight based on the query, but now it is at /containers/list/all and the other tabs are at different urls. This means callers like RunImage have to know about the tabs (and that they're implemented differently than other filters).

feloy:
> Is there any reason the URL to the page has to change? I was expecting the 'all tab' to remain the default at /containers and the respective tabs would highlight based on the query, but now it is at /containers/list/all and the other tabs are at different urls. This means callers like RunImage have to know about the tabs (and that they're implemented differently than other filters).

Using a `<Tab>` element to implement the tabs is probably not a good idea: I first used it to mimic the graphical Tabs, but it implies to involve the router, and so to have different routes for different tabs. In the other hand, it is necessary to pass the running/stopped filter to the query, and, by this, to synchronize this filter between route and query. 

I can try to change the implementation with radio buttons instead, to keep the route of the page as before, and have the running/stopped filter only in the query. This will probably need some work to graphically mimic tabs with radio buttons. @deboer-tim WDYT?

feloy:
> > Is there any reason the URL to the page has to change? I was expecting the 'all tab' to remain the default at /containers and the respective tabs would highlight based on the query, but now it is at /containers/list/all and the other tabs are at different urls. This means callers like RunImage have to know about the tabs (and that they're implemented differently than other filters).
> 
> Using a `<Tab>` element to implement the tabs is probably not a good idea: I first used it to mimic the graphical Tabs, but it implies to involve the router, and so to have different routes for different tabs. In the other hand, it is necessary to pass the running/stopped filter to the query, and, by this, to synchronize this filter between route and query.
> 
> I can try to change the implementation with radio buttons instead, to keep the route of the page as before, and have the running/stopped filter only in the query. This will probably need some work to graphically mimic tabs with radio buttons. @deboer-tim WDYT?

Here is a new PR, not using urls to pass the filter: https://github.com/containers/podman-desktop/pull/4600
I still need to cherry-pick tests and other changes to this new PR.

feloy:
Closing in favour of #4600

PR #4491, opened 2023-10-24
chore(ui,tests): Add ARIA accessibility to the Registries page

xbabalov:
### What does this PR do?
Provides ARIA labels for the Registries settings page for better (testing) accessibility. 

https://github.com/containers/podman-desktop/issues/4376

### How to test this PR?
yarn test

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
@@ -252,18 +252,21 @@ const processPasswordElement = (node: HTMLInputElement, registry: containerDeskt
 
   <div class="container bg-charcoal-600 rounded-md p-3">
     <!-- Registries table start -->
-    <div class="w-full border-t border-b border-gray-900" role="region" aria-label="Registries table">
-      <div class="flex w-full">
-        <div class="flex-1 text-left py-4 pl-5 text-sm font-bold w-auto">Registry Location</div>
-        <div class="text-left py-4 text-sm font-bold w-1/4">Username</div>
-        <div class="text-left py-4 text-sm font-bold w-2/5">Password</div>
+    <div class="w-full border-t border-b border-gray-900" role="table" aria-label="Registries">
+      <div class="flex w-full" role="rowgroup" aria-label="header">
+        <div class="flex-1 text-left py-4 pl-5 text-sm font-bold w-auto" role="columnheader">Registry Location</div>
+        <div class="text-left py-4 text-sm font-bold w-1/4" role="columnheader">Username</div>
+        <div class="text-left py-4 text-sm font-bold w-2/5" role="columnheader">Password</div>
       </div>
 
       {#each $registriesInfos as registry}
         <!-- containerDesktopAPI.Registry row start -->
-        <div class="flex flex-col w-full border-t border-gray-900" aria-label="{registry.serverUrl}">
+        <div
+          class="flex flex-col w-full border-t border-gray-900"
+          role="row"
+          aria-label="{registry.name ? registry.name : registry.serverUrl}">
           <div class="flex flex-row items-center pt-4 pb-3">
-            <div class="flex-1 pl-5 pr-5 text-sm w-auto m-auto">
+            <div class="flex-1 pl-5 pr-5 text-sm w-auto m-auto" role="cell">
               <div class="flex w-full h-full">
                 <div class="flex items-center">
                   <!-- Only show if a "suggested" registry icon has been added -->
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
```diff
@@ -252,7 +252,7 @@ const processPasswordElement = (node: HTMLInputElement, registry: containerDeskt
 
   <div class="container bg-charcoal-600 rounded-md p-3">
     <!-- Registries table start -->
-    <div class="w-full border-t border-b border-gray-900">
+    <div class="w-full border-t border-b border-gray-900" role="region" aria-label="Registries table">
```
odockal: I would use the description on ARIA table suggestions as per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/table_role.

In this div, it would be right to have also `role="table"` and then we can have `aria-label="Registries"`.

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
```diff
@@ -252,7 +252,7 @@ const processPasswordElement = (node: HTMLInputElement, registry: containerDeskt
...
       <div class="flex w-full">
```
odockal: Here we could probably use `role="rowgroup"` or maybe even better, just a `row` role since there will be only one row. although, please take a look at specification, you might get better idea. Then, propably `aria-label="header"` attribute.

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
```diff
@@ -252,7 +252,7 @@ const processPasswordElement = (node: HTMLInputElement, registry: containerDeskt
...
         <div class="flex-1 text-left py-4 pl-5 text-sm font-bold w-auto">Registry Location</div>
```
odockal: Here is could be a role `columnheader`. Please see the docs.

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
```diff
@@ -261,7 +261,7 @@ const processPasswordElement = (node: HTMLInputElement, registry: containerDeskt
 
       {#each $registriesInfos as registry}
         <!-- containerDesktopAPI.Registry row start -->
-        <div class="flex flex-col w-full border-t border-gray-900">
+        <div class="flex flex-col w-full border-t border-gray-900" aria-label="{registry.serverUrl}">
```
odockal: For each div, that denounce the row, please use probably `row` role and `aria-label="registry.NAME?"` the property of the registry object that reference the Registry Name ( ie. "Docker Hub", "Red Hat Quay", etc.). We could then use `aria-labeledby="{registry.serverUrl}"`.

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
```diff
@@ -294,6 +294,7 @@ const processPasswordElement = (node: HTMLInputElement, registry: containerDeskt
                   <input
                     type="text"
                     placeholder="Username"
+                    aria-label="Username"
```
odockal, line 300: We could also probably use `role="cell"` on a proper level - probably on the div that is the most outer for given cell.

odockal: This also counts for other cells we need to operate over.

packages/renderer/src/lib/ui/DropdownMenu.svelte
```diff
@@ -41,7 +41,8 @@ function onWindowClick(e: any) {
       toggleMenu();
     }}"
     bind:this="{outsideWindow}"
-    class="mr-2 text-gray-400 hover:bg-charcoal-800 hover:text-purple-400 font-medium rounded-md inline-flex items-center px-2 py-2 text-center">
+    class="mr-2 text-gray-400 hover:bg-charcoal-800 hover:text-purple-400 font-medium rounded-md inline-flex items-center px-2 py-2 text-center"
+    aria-label="Dropdown menu">
```
odockal: How is this change related to registries?

odockal (changes requested):

packages/renderer/src/lib/ui/DropdownMenu.svelte
xbabalov: For each registry row there's a dropdown menu to offer the edit and delete options.

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
xbabalov: The newly added registries dont have the name attribute but are labeled by their url, so I kept serverUrl for the new registries and name for default registries. I can further separate by adding another rowgroup in between the new and default registries if that's desired.

packages/renderer/src/lib/ui/DropdownMenu.svelte
odockal: @xbabalov ok, thanks!

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
odockal: I would not introduce another separator, instead, can we use ternary operator and use name if the property exists and url if it does not?

odockal (approved):

PR #4497, opened 2023-10-24
docs: using standard OS tabs for registries docs

### LINKED ISSUE #4297: Use OS tabs for proxies documentation

themr0c:
### What does this PR do?

* Using standard "Windows, macOS, Linux" OS tabs to display options.
* Slight refactoring: using snippets for repeated content.
* Moved the content without rewriting it.
* Minimal content fixes.

Screenshot_20231025_140902

Standardize navigation.

### DIFF
--- a/website/docs/working-with-containers/registries/_insecure-registry-linux.md
+++ b/website/docs/working-with-containers/registries/_insecure-registry-linux.md
@@ -1,6 +1,6 @@
 #### Prerequisites
 
-- Podman.
+- Podman is installed.
 
 #### Procedure
 
--- a/website/docs/working-with-containers/registries/_insecure-registry-linux.md
+++ b/website/docs/working-with-containers/registries/_insecure-registry-linux.md
@@ -1,0 +1,37 @@
+#### Prerequisites
+
+- Podman.
+
+#### Procedure
+
+1. Add your insecure registry within **<icon icon="fa-solid fa-cog" size="lg" /> Settings > Registries**.
+
+   ![Adding a custom registry](img/adding-a-custom-registry.png)
+
+1. Click "Yes" to the insecure registry warning.
+
+   ![Podman Desktop Registry Warning](img/registry-warning-insecure.png)
+
+1. Open `registries.conf`.
+
+   ```shell-session
+   $ sudo vi /etc/containers/registries.conf
+   ```
+
+1. Add the insecure registry: Add a new `[[registry]]` section for the URL of the insecure registry you want to use. For example, if your insecure registry is located at `http://registry.example.com`, add the following lines:
+
+   ```toml
+   [[registry]]
+   location = "registry.example.com"
+   insecure = true
+   ```
+
+   If you have multiple registries, you can add one `[[registry]]` block per registry:
+
+1. Save and exit the file.
+
+1. Restart Podman.
+
+   ```shell-session
+   $ sudo systemctl restart podman
+   ```
--- a/website/docs/working-with-containers/registries/_insecure-registry-windows-macos.md
+++ b/website/docs/working-with-containers/registries/_insecure-registry-windows-macos.md
@@ -1,0 +1,44 @@
+#### Prerequisites
+
+- [A running Podman machine](/docs/onboarding-for-containers/installing-podman).
+
+#### Procedure
+
+1. Add your insecure registry within **<icon icon="fa-solid fa-cog" size="lg" /> Settings > Registries**.
+
+   ![Adding a custom registry](img/adding-a-custom-registry.png)
+
+1. Click "Yes" to the insecure registry warning.
+
+   ![Podman Desktop Registry Warning](img/registry-warning-insecure.png)
+
+1. SSH into the Podman Machine to edit `registries.conf`.
+
+   ```shell-session
+   $ podman machine ssh [optional-machine-name]
+   ```
+
+1. Open `registries.conf`.
+
+   ```shell-session
+   $ sudo vi /etc/containers/registries.conf
+   ```
+
+1. Add the insecure registry: Add a new `[[registry]]` section for the URL of the insecure registry you want to use. For example, if your insecure registry is located at `http://registry.example.com`, add the following lines:
+
+   ```toml
+   [[registry]]
+   location = "registry.example.com"
+   insecure = true
+   ```
+
+   If you have multiple registries, you can add one `[[registry]]` block per registry.
+
+1. Save and exit the file.
+
+1. Restart Podman by the CLI or through Podman Desktop.
+
+   ```shell-session
+   $ podman machine stop
+   $ podman machine start
+   ```
--- a/website/docs/working-with-containers/registries/_verification-private-registry.md
+++ b/website/docs/working-with-containers/registries/_verification-private-registry.md
@@ -1,0 +1,8 @@
+#### Verification
+
+1. Go to **<icon icon="fa-solid fa-cloud" size="lg" /> Images**.
+1. You can pull a private image from the registry.
+1. You can push an image to the registry:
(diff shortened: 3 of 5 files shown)


themr0c:
This check is no good. Starting references to images with `/docs...` unfortunately does not work.

```
[vale] reported by reviewdog 🐶
[PodmanDesktop.Links] Consider starting the link with https:// or /, rather than ](img/adding-a-custom-registry.png)
```

website/docs/working-with-containers/registries/_insecure-registry-linux.md
```diff
@@ -0,0 +1,37 @@
+#### Prerequisites
+
+- Podman.
```
deboer-tim: I assume it needs to be running? Before the change that was required, and there is a restart in the steps.

deboer-tim (approved):

benoitf:
I don't see why we add 2 new files (linux / windows) that are only included by one file.

why not keep the existing file (not extracting the content)

themr0c:
@benoitf  for clarity and readability, I find it preferable to decouple the presentation layer (tabs) from the content (snippets). I plan to generalize this approach in every page containing tabs. I have not done that before only because previously, I was not familiar with the concept of "snippets" in Docusaurus.

website/docs/working-with-containers/registries/_insecure-registry-linux.md
themr0c: It only needs to be installed. You need to restart any podman process if it exists. I will clarify.

themr0c:
```suggestion
- Podman is installed.
```

website/docs/working-with-containers/registries/_insecure-registry-linux.md
```diff
@@ -0,0 +1,37 @@
...
+
+   If you have multiple registries, you can add one `[[registry]]` block per registry:
+
+1. Save and exit the file.
+
+1. Restart Podman.
+
+   ```shell-session
+   $ sudo systemctl restart podman
+   ```
```
themr0c, line 37: The "restart Podman" steps in various Linux procedures (here, proxy) are inconsistent and mostly irrelevant. It deserves its own issue.

website/docs/working-with-containers/registries/_insecure-registry-linux.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. Add the insecure registry: Add a new `[[registry]]` section for the URL of the insecure registry you want to use. For example, if your insecure registry is located at `http://registry.example.com`, add the following lines:
+
+   ```toml
+   [[registry]]
+   location = "registry.example.com"
+   insecure = true
+   ```
+
+   If you have multiple registries, you can add one `[[registry]]` block per registry:
```
cdrage, line 29: `:` added by mistake

website/docs/working-with-containers/registries/_insecure-registry-linux.md
```diff
@@ -0,0 +1,37 @@
...
+   ![Adding a custom registry](img/adding-a-custom-registry.png)
+
+1. Click "Yes" to the insecure registry warning.
+
+   ![Podman Desktop Registry Warning](img/registry-warning-insecure.png)
+
+1. Open `registries.conf`.
+
+   ```shell-session
+   $ sudo vi /etc/containers/registries.conf
```
cdrage, line 18: Isn't it in openshift docs / standards that sudo commands use `#` / comment out feature so it's not accidently copied-and-pasted? unsure if we should apply this here / other parts of PD documentation.

website/docs/working-with-containers/registries/_insecure-registry-windows-macos.md
```diff
@@ -0,0 +1,44 @@
...
+
+#### Procedure
+
+1. Add your insecure registry within **<icon icon="fa-solid fa-cog" size="lg" /> Settings > Registries**.
+
+   ![Adding a custom registry](img/adding-a-custom-registry.png)
+
+1. Click "Yes" to the insecure registry warning.
+
+   ![Podman Desktop Registry Warning](img/registry-warning-insecure.png)
```
cdrage, line 13: false positive on vale issue?

website/docs/working-with-containers/registries/_insecure-registry-windows-macos.md
```diff
@@ -0,0 +1,44 @@
...
+   $ podman machine ssh [optional-machine-name]
+   ```
+
+1. Open `registries.conf`.
+
+   ```shell-session
+   $ sudo vi /etc/containers/registries.conf
+   ```
+
+1. Add the insecure registry: Add a new `[[registry]]` section for the URL of the insecure registry you want to use. For example, if your insecure registry is located at `http://registry.example.com`, add the following lines:
```
cdrage, line 27: Accidental :?

website/docs/working-with-containers/registries/_insecure-registry-windows-macos.md
```diff
@@ -0,0 +1,44 @@
...
+
+   ```toml
+   [[registry]]
+   location = "registry.example.com"
+   insecure = true
+   ```
+
+   If you have multiple registries, you can add one `[[registry]]` block per registry.
+
+1. Save and exit the file.
```
cdrage, line 37: I don't think we need this step as it's implied by editing the file you save and exit.

cdrage:
I still have some comments on this PR @themr0c

website/docs/working-with-containers/registries/_insecure-registry-linux.md
themr0c: `# command` for a command by root. But usually people don't understand.


`$ sudo command` is correct, as you execute in the user context (sudo does the privilege escalation). Less prone to error because user forget to get root.

website/docs/working-with-containers/registries/_insecure-registry-windows-macos.md
themr0c: Bad rule. I opened an issue. And am currently lost in regex: do not match strings starting by img/. The `/docs...` notation breaks the build...

website/docs/working-with-containers/registries/_insecure-registry-windows-macos.md
themr0c: That was already in the content. I changed as little content as possible. This might require a new pr with new scope.

deboer-tim:
> @benoitf for clarity and readability, I find it preferable to decouple the presentation layer (tabs) from the content (snippets). I plan to generalize this approach in every page containing tabs. I have not done that before only because previously, I was not familiar with the concept of "snippets" in Docusaurus.

Just to be upfront, I was also uncomfortable with the separation but let it go b/c I thought 'it's just one doc' and I like the user-side of the change. When I saw Florent's comments I realized I was wrong, but by then this was already merged.

I definitely do not think it is a good idea to use imports for every page with tabs - imports should be rare, and only for cases where one file is getting unmanageable or it actually makes sense for the same content to appear in two places. Why? First, it will make it harder for anyone else to contribute to docs, because it's harder to find things without searching, and having to spend extra time to understand the structure and how and where the content is going to appear, follow links, etc. Second, with 4x the files it would be harder to maintain, rename/move topics, slower to review PRs, etc for the same reasons.

Issue #4456, opened 2023-10-20
Error when installing compose system wide on ubuntu


lstocchi:
### Bug description

![error_onboarding_compose](https://github.com/containers/podman-desktop/assets/49404737/6555722f-0d05-43d9-9e5a-1c51d1183274)

### Operating system

Ubuntu 22.04

### Installation Method

Installer from website/GitHub releases

### Version

next (development version)

cdrage:
Any output from the dev console?

odockal:
@cdrage erros from dev console:
```
VM5:2 main ↪️ [compose] Failed to install 'docker-compose' binary: Command execution failed with exit code 127: Command execution failed with exit code 127
console.<computed> @ VM5:2
(anonymous) @ VM5:2
emit @ VM4 sandbox_bundle:2
onMessage @ VM4 sandbox_bundle:2
```

```
VM5:2 main ↪️ [compose] Command execution failed with exit code 127: Command execution failed with exit code 127
    at ChildProcess.<anonymous> (/home/odockal/git/podman-desktop/dist/linux-unpacked/resources/app/packages/main/dist/index.cjs:382:12995)
    at ChildProcess.emit (node:events:514:28)
    at Process.onexit (node:internal/child_process:291:12)
console.<computed> @ VM5:2
(anonymous) @ VM5:2
emit @ VM4 sandbox_bundle:2
onMessage @ VM4 sandbox_bundle:2
```

PR #4501, opened 2023-10-24
fix: handle flatpak usecase for admin commands

benoitf:
### What does this PR do?
we need to use flatpak-spawn only after command is set and not at the beginning

else we may try to execute commands with incorrect order of params

related to https://github.com/containers/podman-desktop/issues/4456

one error of #4456 

### How to test this PR?

unit test provided


cdrage (approved):
LGTM! Patched this with the other PR and it worked on Fedora 38. 👍

mairin:
Just tested on F38 baremetal using `yarn watch` and PR #4502, worked great.

PR #4503, opened 2023-10-25
feat: show CLI versions

### LINKED ISSUE #3707: Compose settings: Show CLI version number

As per https://github.com/containers/podman-desktop/issues/2523:

We should be able to see the CLI version number of compose.

Implement the "version" part of the UI

**August 31st update:** Awaiting finalized UI/UX to show update mechanism

_No response_

### Additional context

_No response_

cdrage:
feat: show CLI versions

### What does this PR do?

* Shows the CLI version number by executing the `--version` or `-v`,
  etc. command via the command line. This is done by adding the
  "helpCommand" parameter to createCliTool as well as the storage path
  of the binary from extensionContext.storagePath
* If it succeeds, it will show the version number, if it fails, it will
  not appear (but show the error via the console). This would only
  happen if the extension was registered but the binary removed from the
  extension folder.

Screenshot 2023-10-24 at 8 17 42 PM

Closes https://github.com/containers/podman-desktop/issues/3707

### How to test this PR?

1. Install Compose / go through the onboarding
2. View the CLI Tools via preferences and you'll see the version number.

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2447,6 +2447,10 @@ declare module '@podman-desktop/api' {
     displayName: string;
     markdownDescription: string;
     images: ProviderImages;
+    storagePath: string;
+    // This is needed since every binary handles getting the version differently, for example some tools use
+    // --version, some use -v, some use -V, etc.
+    helpCommand: string;
   }
 
   export type CliToolState = 'registered';
--- a/packages/main/src/plugin/cli-tool-registry.ts
+++ b/packages/main/src/plugin/cli-tool-registry.ts
@@ -21,6 +21,8 @@ import type { CliToolExtensionInfo, CliToolInfo } from './api/cli-tool-info.js';
 import type { ApiSenderType } from './api.js';
 import type { Telemetry } from './telemetry/telemetry.js';
 import type { Exec } from './util/exec.js';
+import path from 'node:path';
+import { isWindows } from '../util.js';
 
 export class CliToolImpl implements CliTool, Disposable {
   readonly id: string;
@@ -56,6 +58,11 @@ export class CliToolImpl implements CliTool, Disposable {
     return Object.freeze(this._options.images);
   }
 
+  // getVersion will call getCliVersion which will check via running the binary with `--version` or `-v`, etc (from _options.helpCommand)
+  async getVersion(): Promise<string> {
+    return await getCliVersion(this._exec, this._options.name, this._options.storagePath, this._options.helpCommand);
+  }
+
   dispose(): void {
     this.registry.disposeCliTool(this);
   }
@@ -82,9 +89,11 @@ export class CliToolRegistry {
     this.apiSender.send('cli-tool-remove', cliTool.id);
   }
 
-  getCliToolInfos(): CliToolInfo[] {
-    return Array.from(this.cliTools.values()).map(cliTool => {
-      return {
+  async getCliToolInfos(): Promise<CliToolInfo[]> {
+    const cliToolInfos: CliToolInfo[] = [];
+    for (const cliTool of this.cliTools.values()) {
+      const cliVersion = await cliTool.getVersion();
+      cliToolInfos.push({
         id: cliTool.id,
         name: cliTool.name,
         displayName: cliTool.displayName,
@@ -92,7 +101,40 @@ export class CliToolRegistry {
         state: cliTool.state,
         images: cliTool.images,
         extensionInfo: cliTool.extensionInfo,
-      };
-    });
+        version: cliVersion,
+      });
+    }
+    return cliToolInfos;
   }
+}
+
+// This function will take in the exec, name and storage path of the binary.
+// it will also take in the "command" that achieves getting the version of the binary as
+// each tool has a different way of getting it (for example, -v, --version, version, --help, etc.)
+// NOTE: We only care what's in the ~/.local/share/containers/podman-desktop/extensions-storage/ folder
+// not the binary that is installed on the system (since we have the ability to install system-wide)
+async function getCliVersion(exec: Exec, name: string, storagePath: string, helpCommand: string): Promise<string> {
+  // Make sure we resolve if it is a Windows binary or not as well.
+  const destFile = path.resolve(storagePath, 'bin', isWindows() ? name + '.exe' : name);
+  try {
+    const result = await exec.exec(destFile, [helpCommand]);
+    return extractVersion(result.stdout);
+  } catch (e) {
+    // If unable to get binary, just return blank, don't bother with returning an error, just output it to console.
+    console.log(`Error getting version of binary ${name}, error:`, e);
+  }
+  return '';
+}
+
+// Extracts a version from an output string. This is a best-effort attempt to
+// extract a version from the output of a command. It is not guaranteed to
+// succeed.
+function extractVersion(output: string): string {
+  // This regex captures:
+  // - A leading "v" (optional)
+  // - Major, minor, and patch numbers
+  // - Optional suffixes like -alpha, -beta, -rc1, etc.
+  const regex = /v?(\d+\.\d+(\.\d+)?(-[a-zA-Z0-9]+)?)/;
+  const match = output.match(regex);
+  return match ? match[1] : '';
 }
--- a/extensions/compose/src/extension.ts
+++ b/extensions/compose/src/extension.ts
@@ -31,6 +31,11 @@ import { ComposeInstallation } from './installation';
 let composeExtension: ComposeExtension | undefined;
 let composeVersionMetadata: ComposeGithubReleaseArtifactMetadata | undefined;
 
+const composeCliName = 'docker-compose';
+const composeDisplayName = 'Compose';
+const composeDescription = `Compose is a specification for defining and running multi-container applications. We support both [podman compose](https://docs.podman.io/en/latest/markdown/podman-compose.1.html) and [docker compose](https://github.com/docker/compose) commands.\n\nMore information: [compose-spec.io](https://compose-spec.io/)`;
+const imageLocation = './icon.png';
+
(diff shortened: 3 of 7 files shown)


cdrage:
Some notes:

Requires https://github.com/containers/podman-desktop/pull/4437 to be merged before this PR.

@mairin @benoitf @deboer-tim 

There may be a case where the user removes the binary from the `~/.local/containers/podman-desktop` extensions storage folder and we'll error out when trying to get the version number (file not found). Right now I simply return blank and then output to the console. Obviously if they were to remove the local binary, onboarding Setup button will re-appear so at least the user will know something is wrong.

My question is, should I keep the current behaviour of simply not showing the version as it would be solved by the user running onboarding again (since it'll re-appear), or should I have it show a warning of "Binary not found in ~/.local/containers etc... folder)?

mairin:
@cdrage Could you do something like, "Binary not found, run set up again" ?

deboer-tim:
> @cdrage Could you do something like, "Binary not found, run set up again" ?

+1. I haven't looked at the API, but sounds like a 'binary not found' error that could be returned by any cli provider and UI would render text like this. (ie not compose specific)

cdrage:
Thank you for the suggestions! I've updated the PR to add this if the version is undetectable:

Screenshot 2023-10-24 at 10 23 44 PM

cdrage:
Tests have been added, PR is ready for a review 🎉

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2447,6 +2447,10 @@ declare module '@podman-desktop/api' {
     displayName: string;
     markdownDescription: string;
     images: ProviderImages;
+    storagePath: string;
+    // This is needed since every binary handles getting the version differently, for example some tools use
+    // --version, some use -v, some use -V, etc.
+    helpCommand: string;
```
lstocchi, line 2453: I find the name a bit confusing. Wouldn't it be better to be named `versionCommand` ?

packages/main/src/plugin/cli-tool-registry.ts
```diff
@@ -82,17 +89,52 @@ export class CliToolRegistry {
...
 }
+
+// This function will take in the exec, name and storage path of the binary.
+// it will also take in the "command" that achieves getting the version of the binary as
+// each tool has a different way of getting it (for example, -v, --version, version, --help, etc.)
+// NOTE: We only care what's in the ~/.local/share/containers/podman-desktop/extensions-storage/ folder
+// not the binary that is installed on the system (since we have the ability to install system-wide)
+async function getCliVersion(exec: Exec, name: string, storagePath: string, helpCommand: string): Promise<string> {
+  // Make sure we resolve if it is a Windows binary or not as well.
+  const destFile = path.resolve(storagePath, 'bin', isWindows() ? name + '.exe' : name);
```
lstocchi, line 118: Not sure what's the correct behavior but if we consider the compose set up to be completed when the compose is installed system-wide, why do we display the version from the cli in the storagePath?
 
Maybe there is something that still has to be implemented and i don't know but i find it weird. In future, Desktop compose will execute commands by using the compose installed system-wide or that in the storagePath? Because if it's the latter, i don't get why we consider the set up completed if it's installed system-wide. If it's the former, we should show the version of the system-wide.

lstocchi (commented):
Works fine, just a couple of comments

packages/extension-api/src/extension-api.d.ts
benoitf: I don't think we need to provide this information, it's the responsibility of the provider to provide the version.

Version can have multiple lines, be with tons of options, so it should not be managed by the extension API

benoitf (changes requested):
the version needs to be provided, not by providing some special flags, etc.

cdrage:
> I don't think we need to provide this information, it's the responsibility of the provider to provide the version.
> 
> Version can have multiple lines, be with tons of options, so it should not be managed by the extension API

Do you mind pointing me where we should implement getting the version? 

My only concern is that we are implementing update mechanisms into the API via: https://github.com/containers/podman-desktop/pull/4092/files#diff-37880cdf58174a1e52a3b42dd17e3ce8bff6b4f60a85588048c23417fe0c5de8

We already retrieve the file, wouldn't it make sense to simply execute `--version / -v`  on it too to retrieve the version / managed by the extension?

My concern is that a provider could:
1. Point to /usr/local/bin and that the versioning is coming from there which would be different than our ~/.local/share install
2. Won't be able to actively refresh / update on command since a provider will be managing it.

IMHO I think PD should retrieve the versioning since the binary is in our ~/.local/share folder anyways.

cdrage:
> I don't think we need to provide this information, it's the responsibility of the provider to provide the version.
> 
> Version can have multiple lines, be with tons of options, so it should not be managed by the extension API

Do you mind pointing me where we should implement getting the version? 

My only concern is that we are implementing update mechanisms into the API via: https://github.com/containers/podman-desktop/pull/4092/files#diff-37880cdf58174a1e52a3b42dd17e3ce8bff6b4f60a85588048c23417fe0c5de8

We already retrieve the file, wouldn't it make sense to simply execute `--version / -v`  on it too to retrieve the version / managed by the extension?

My concern is that a provider could:
1. Point to /usr/local/bin and that the versioning is coming from there which would be different than our ~/.local/share install
2. Won't be able to actively refresh / update on command since a provider will be managing it.

IMHO I think PD should retrieve the versioning since the binary is in our ~/.local/share folder anyways. 

LMK your thoughts @benoitf / @lstocchi that's the only part I'm concerned about.

packages/extension-api/src/extension-api.d.ts
cdrage: You're right, should of been `versionCommand` not help. Was from a previous iteration.

packages/main/src/plugin/cli-tool-registry.ts
cdrage: It's because the ~/.local directory is universal between OS' and we provide the option to system-install. The .local directory 

Checking the version via $PATH may give different results depending on where the binary is located (could be located in three different areas... for example, the `kompose` binary may be located in the $GOBIN directory, or /usr/local/ or maybe even in brew).

So we would check the ~/.local folder first and then we can provide a way to install system-wide.

cdrage:
From the call regarding implementation:

- [x] Extension should handle checking the cli version number via running the local binary and then pushing that property to the PD API. We can use the `exec` functionality of the PD API in order to execute `docker-compose --version`. This is because every extension may be a bit different (regex, etc.) in order to get the version number.
- [x] Podman desktop API should have cliVersion value which is managed by the extension
- [x] PD API should maybe have a 'cliBinaryLocation' value which is pushed / managed by the extension
- [x] Documentation added regarding this functionality into the PD api
- [x] Within the UI we should (in small font) tell the user where the binary is actually located / where we are checking

Future:
- [ ] PD should ask each extension / trigger an update to check the binary install so we can check github releases, etc.

packages/main/src/plugin/cli-tool-registry.ts
```diff
@@ -82,17 +89,52 @@ export class CliToolRegistry {
...
+    return cliToolInfos;
   }
 }
+
+// This function will take in the exec, name and storage path of the binary.
+// it will also take in the "command" that achieves getting the version of the binary as
+// each tool has a different way of getting it (for example, -v, --version, version, --help, etc.)
+// NOTE: We only care what's in the ~/.local/share/containers/podman-desktop/extensions-storage/ folder
+// not the binary that is installed on the system (since we have the ability to install system-wide)
+async function getCliVersion(exec: Exec, name: string, storagePath: string, helpCommand: string): Promise<string> {
```
dgolovin, line 116: IMO it should be declared in CliTool interface and implemented in CliToolImpl. The reason is it would need access to CliToolImpl internal properties to change state of CliToolImpl instance. It has currently only one state `registered`. It means CliTool added podman desktop and nothing was done with it. As we introduce detection there are going to be additional states 'detection is in progress`, 'not detected', 'detected'. Those can be calculated based on values of internal properties (if version property is present then state is detected and etc) or set explicitly by detection function as it goes through different detection stages.

packages/main/src/plugin/cli-tool-registry.ts
dgolovin: I agree with @lstocchi that at this point storagePath should not be involved in detection, because if we provide binaries inside extension then we now the location and version. We also can say for sure if version installed system wide is the same we have bundled with extension

packages/extension-api/src/extension-api.d.ts
dgolovin: name property holds command name without file extension, all you need is 'options' to run command to produce stdout with version and parser function that gets stdout text and figure out what piece of text is the version.

For version parser function there are several variations to support when command prints:
- only version 
- some text and version somewhere in between
- when output is json or yaml and you have to know name of the property to extract form parsed object

I suggested that before but it did not get into first push see my comment here https://github.com/containers/podman-desktop/pull/4197#discussion_r1358600751.

packages/main/src/plugin/cli-tool-registry.ts
```diff
@@ -82,17 +89,52 @@ export class CliToolRegistry {
...
   }
 
-  getCliToolInfos(): CliToolInfo[] {
-    return Array.from(this.cliTools.values()).map(cliTool => {
-      return {
+  async getCliToolInfos(): Promise<CliToolInfo[]> {
+    const cliToolInfos: CliToolInfo[] = [];
+    for (const cliTool of this.cliTools.values()) {
+      const cliVersion = await cliTool.getVersion();
+      cliToolInfos.push({
```
dgolovin, line 96: There should be not so many calls go get all CliToolInfo's. Only at start or if manual reload requested (not if it is going to be needed). Mostly events are going to be sent out about added/removed/changed items.

cdrage:
Sorry @dgolovin forgot to put this in draft / pending for review after the comments. I've refactored and opened up: https://github.com/containers/podman-desktop/pull/4541 instead.

Issue #3538, opened 2023-08-15
Pushing local image to the OpenShift Local Cluster does nothing


rostalan:
### Bug description

Trying to deploy  pod created from image to Kubernetes using crc-admin context. For that I assume, I need to push the local image to the OpenShift Local cluster? When trying to do this with Kind/Dev Sandbox, it pops a message, that the push has been succesfull. With OS Local cluster it just doe nothing. As a result, cannot deploy pods or containers created from that image.
![Screenshot_20230815_131356](https://github.com/containers/podman-desktop/assets/75624842/9aabbc83-1a85-434b-81e0-2504f246bb1e)
![Screenshot_20230815_131251](https://github.com/containers/podman-desktop/assets/75624842/57447200-6fff-4a0e-a840-dc5e38994bf5)

### Operating system

Fedora 38

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

Have an image, select drop down menu on the right, click Push image to OpenShift Local cluster.
There is no pop-up or message unlike for Kind/Dev Sanbox.

### Relevant log output

```shell
main ↪️ event is 
{status: 'save', id: 'bc80818f28f0d5acc6b4abf1c08d531506e94e1bb9e99366b4e7c52f17989e3a', Type: 'image', Action: 'save', Actor: {…}, …}
Action
: 
"save"
Actor
: 
{ID: 'bc80818f28f0d5acc6b4abf1c08d531506e94e1bb9e99366b4e7c52f17989e3a', Attributes: {…}}
Type
: 
"image"
id
: 
"bc80818f28f0d5acc6b4abf1c08d531506e94e1bb9e99366b4e7c52f17989e3a"
scope
: 
"local"
status
: 
"save"
time
: 
1692098024
timeNano
: 
1692098024497994500
[[Prototype]]
: 
Object
```

### Additional context

There are several issues opened in relation to this one:
1. https://github.com/crc-org/crc/issues/3897 does not allow to push to the OpenShift Local using podman/docker push command we use to push to OpenShift Sandbox cluster
2. https://github.com/crc-org/crc-extension/issues/185 OS specific notification that is not noticeable right away should be replaced by information message used by other extensions and there should be task added in task view to show progress on the push
3. https://github.com/crc-org/crc-extension/issues/186 there is a OpenShift Local status detection issues on linux (Fedora 38 specifically) when `crc setup --check-only` returns error code 2 when crc set up correctly but not started
[...]

PR #4523, opened 2023-10-26
fix: check for self signed cert message and use insecure param when editing registry password

### LINKED ISSUE #4522: Adding OpenShift Local docker registry does not trigger insecure registry confirmation request
### Bug description

I have OpenShift Local running and want to add docker registry form it to list of the registries in podman desktop. The registry is using self signed certificate and I always get an error message about 'self signed certificate in certificate chain'. There should be a request if certificate is not trusted and if I still want to use this specific registry, but it never shows up.

https://github.com/containers/podman-desktop/assets/620330/a9d0b09c-1ad8-4857-9e04-b727b960bc28

The same problem is happening if I would try to update password for the registry using self singed certificate.

https://github.com/containers/podman-desktop/assets/620330/bdda7399-9222-434c-bc3f-38535f7d7ba9

### Operating system

macOS Sonoma 14.0

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

When creating new registry:
1. Install OpenShift Local extension
2. Start OpenShift Local cluste
3. Switch to crc-admin kubernetes context
4. Open terminal and run oc whoami --show-token
5. Copy token to clipboard
6. Open Settings/Registries
7. Add new registry  url: default-route-openshift-image-registry.apps-crc.testing, user: kubeadmin, password: paste token

When editing OpenShift Local registry password
1. Select edit 'Edit Password'
2. Change password to incorrect one
3. Press Login
4. Error appears about 'self signed certificate in certificate chain' not 'Incorrect username or password'

### Relevant log output

_No response_

### Additional context

related to testing and looking for solution of #3538.

dgolovin:
### What does this PR do?

Checking for another message related to self signed certificate when adding OpenShift Local image registry to trigger insecure registry confirmation request.

Use insecure parameter to verify new credentials when editing password for insecure registry.

https://github.com/containers/podman-desktop/assets/620330/a6d13e9f-0048-4c8c-8c9c-223eda46e911

fix #4522.
releated to testing of #3538.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
@@ -212,9 +212,9 @@ async function loginToRegistry(registry: containerDesktopAPI.Registry) {
 
   try {
     if (newRegistry) {
-      await window.createImageRegistry(registry.source, registry);
+      await window.createImageRegistry(registry.source, { ...registry });
     } else {
-      await window.updateImageRegistry(registry);
+      await window.updateImageRegistry({ ...registry });
     }
   } catch (error: any) {
     setErrorResponse(registry.serverUrl, error.message);
--- a/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
@@ -189,7 +189,11 @@ async function loginToRegistry(registry: containerDesktopAPI.Registry) {
   try {
     await window.checkImageCredentials(registry);
   } catch (error) {
-    if (error instanceof Error && error.message.includes('unable to verify the first certificate')) {
+    if (
+      error instanceof Error &&
+      (error.message.includes('unable to verify the first certificate') ||
+        error.message.includes('self signed certificate in certificate chain'))
+    ) {
       const result = await window.showMessageBox({
         title: 'Invalid Certificate',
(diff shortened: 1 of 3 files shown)


jeffmaury (commented):
Add unit test

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
```diff
@@ -208,9 +212,9 @@ async function loginToRegistry(registry: containerDesktopAPI.Registry) {
 
   try {
     if (newRegistry) {
-      await window.createImageRegistry(registry.source, registry);
+      await window.createImageRegistry(registry.source, { ...registry });
```
jeffmaury, line 215: Sounds weird to me

dgolovin: It is just pass a shallow copy of the object to function rather than shared object.
Another option is create new shard object when page cleaning new registry form.

dgolovin: This is required or the hasBeenCalledWith assertion is going to fail in test. It happens because newRegistryForm is going to be cleaned of entered values right after calling window.createNewRegistry.

jeffmaury (approved):
LGTM

cdrage (approved):
LGTM! Was able to test this locally and I was able to successfully edit the password (and have the prompt come up again fine). Code LGTM too.

PR #4539, opened 2023-10-27
feat: make it possible for lima to provide both

### LINKED ISSUE #5206: Use the same Lima for both containers and Kubernetes

Currently you can only create a "podman" or "docker" instance, **or** a "k3s" or "k8s" instance (with containerd).

It is not possible to start Kubernetes on the container engine, nor to access the container runtime of the cluster.

https://podman-desktop.io/docs/lima/customizing

Even if you make a custom template, Podman Desktop currently makes the connections mutually exclusive.

So that when you have the type "kubernetes", you are not able to also have it provide access to the containers.

Make it possible for the same (Lima) provider, to have both a connection endpoint* and a kubernetes endpoint*.

* a unix socket, like: docker.sock

* a tcp socket, like: port 6443

pd-resources-lima-both

pd-preferences-lima-both

In theory you could even run both Docker and Podman on the same instance, but that would be less useful...

You can run "kind", using the podman or docker instance.

We could add another instance field, and run two machines.

The current workaround is to change the provider type in Preferences: Lima, and then restart the Lima provider.

Of course, then you can't access both containers and Kubernetes at the same time (but only one type at a time).

### Additional context

_No response_

afbjorklund:
Allow installing a container engine on your kubernetes node, and use the same lima instance for **both** containers and kubernetes.

For instance by installing "nerdctld" for containerd, or by running with Podman/CRI-O or Docker/CRI-Docker custom setup.

The same VM, running both "containers" and "kubernetes":

lima-provider-both

You configure what Docker socket filename you want to use:

lima-extension-sock

Closes #5206

### How to test this PR?

The default templates don't have any such socket (only containerd), so it would require a custom installation.

Podman: `lima sudo apt install -y podman`. This will install podman version 3.4.4 (or [updates](https://packages.ubuntu.com/jammy-updates/podman))

<details>
<pre>
$ lima sudo podman --remote version
Client:
Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.18.1
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
Server:
Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.18.1
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
</pre>
</details>

Docker: `lima sudo apt install -y docker.io`. This will install docker version 20.10 (or [updates](https://packages.ubuntu.com/jammy-updates/docker.io))

<details>
<pre>
$ lima sudo docker version
Client:
 Version:           24.0.5
 API version:       1.43
 Go version:        go1.20.3
 Git commit:        24.0.5-0ubuntu1~22.04.1
 Built:             Mon Aug 21 19:50:14 2023
 OS/Arch:           linux/amd64
 Context:           default
Server:
 Engine:
  Version:          24.0.5
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.3
  Git commit:       24.0.5-0ubuntu1~22.04.1
  Built:            Mon Aug 21 19:50:14 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.7.7
  GitCommit:        8c087663b0233f6e6e2f4515cee61d49f14746a8
 runc:
  Version:          1.1.9
  GitCommit:        v1.1.9-0-gccaecfcb
 docker-init:
  Version:          0.19.0
  GitCommit:        
</pre>
</details>

Then the unix socket for the container engine needs to be forwarded (see "podman" or "docker" templates).

The VM would run both the containers and the Kubernetes cluster, but they would **not** be sharing images.

So you would have to download the images (from podman/docker), and upload them again (to k3s or k8s)...

### (Advanced) To share image storage:

You would install either Podman + CRI-O or Docker + CRI-Docker, and then set up both the socket and the config.

But the default container runtime is containerd, and it does not _have_ a Docker API socket - only the `nerdctl` CLI.

The support for nerdctld is experimental, it can be found here: https://github.com/afbjorklund/nerdctld

You will need to install nerdctl and buildkit, in addition to the already running containerd (for k3s or k8s).

```
Client:
 Version:           24.0.5
 API version:       1.42 (downgraded from 1.43)
 Go version:        go1.20.3
 Git commit:        24.0.5-0ubuntu1~20.04.1
 Built:             Mon Aug 21 19:50:14 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: 🤓
 nerdctl:
  Version:          1.6.2
 containerd:
  Version:          1.7.6-k3s1.27
 Engine:
  Version:          1.6.2
  API version:      1.42 (minimum version 1.24)
  Go version:       go1.21.1
  Git commit:       e3dc23be348efded17d2cd244397b4f7018e0794
  Built:            
  OS/Arch:          linux/amd64
  Experimental:     true

```

You could also start with the container runtime, and add the Kubernetes installation (complete with CRI/CNI).

For podman you would need to add cri-o, and for docker you would need to add cri-dockerd. And Kubernetes.

* https://docs.k3s.io/installation

* https://kubernetes.io/docs/setup/

### DIFF
--- a/extensions/lima/package.json
+++ b/extensions/lima/package.json
@@ -34,6 +34,11 @@
           "type": "string",
           "default": "",
           "description": "Instance name (default is same name as type)"
+        },
+        "lima.socket": {
+          "type": "string",
+          "default": "",
+          "description": "Socket name (default is equal to type.sock)"
         }
       }
     },
--- a/extensions/lima/src/extension.ts
+++ b/extensions/lima/src/extension.ts
@@ -34,10 +34,10 @@ const imageHandler = new ImageHandler();
 function registerProvider(
   extensionContext: extensionApi.ExtensionContext,
   provider: extensionApi.Provider,
+  providerType: limaProviderType,
   providerPath: string,
 ): void {
   let providerState: extensionApi.ProviderConnectionStatus = 'unknown';
-  const providerType: limaProviderType = configuration.getConfiguration('lima').get('type');
   const instanceName: string = configuration.getConfiguration('lima').get('name') || providerType;
   if (providerType === 'podman' || providerType === 'docker') {
     const connection: extensionApi.ContainerProviderConnection = {
@@ -83,8 +83,9 @@ function registerProvider(
 export async function activate(extensionContext: extensionApi.ExtensionContext): Promise<void> {
   const engineType = configuration.getConfiguration('lima').get('type') || 'podman';
   const instanceName = configuration.getConfiguration('lima').get('name') || engineType;
+  const socketName = configuration.getConfiguration('lima').get('socket') || engineType + '.sock';
   const limaHome = 'LIMA_HOME' in process.env ? process.env['LIMA_HOME'] : os.homedir() + '/.lima';
-  const socketPath = path.resolve(limaHome, instanceName + '/sock/' + engineType + '.sock');
+  const socketPath = path.resolve(limaHome, instanceName + '/sock/' + socketName);
   const configPath = path.resolve(limaHome, instanceName + '/copied-from-guest/kubeconfig.yaml');
 
   let provider;
@@ -104,22 +105,23 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
     extensionContext.subscriptions.push(provider);
   }
 
-  switch (engineType) {
-    case 'podman':
-    case 'docker':
-      if (fs.existsSync(socketPath)) {
-        registerProvider(extensionContext, provider, socketPath);
-      } else {
-        console.debug(`Could not find socket at ${socketPath}`);
-      }
-      break;
-    case 'kubernetes':
-      if (fs.existsSync(configPath)) {
-        registerProvider(extensionContext, provider, configPath);
-      } else {
-        console.debug(`Could not find config at ${configPath}`);
-      }
-      break;
+  // container provider
+  if (socketName !== 'kubernetes.sock') {
+    const providerType = engineType === 'kubernetes' ? 'docker' : (engineType as limaProviderType);
+    if (fs.existsSync(socketPath)) {
+      registerProvider(extensionContext, provider, providerType, socketPath);
+    } else {
+      console.debug(`Could not find socket at ${socketPath}`);
+    }
+  }
+  // kubernetes provider
+  if (engineType === 'kubernetes') {
+    const providerType = engineType as limaProviderType;
+    if (fs.existsSync(configPath)) {
+      registerProvider(extensionContext, provider, providerType, configPath);
+    } else {
+      console.debug(`Could not find config at ${configPath}`);
+    }
   }
 }
 


afbjorklund:
Alternatively one could install cri-o on the podman machine, but I don't know how to do that.

Since Fedora CoreOS is read-only, it is more effort compared to doing it on Ubuntu or CentOS.

`limactl start template://ubuntu-lts`

`limactl start template://centos-stream`

----

Something like: (i.e. merge the templates)
```yaml
copyToHost:
- guest: "/etc/kubernetes/admin.conf"
  host: "{{.Dir}}/copied-from-guest/kubeconfig.yaml"
  deleteOnStop: true
portForwards:
- guestSocket: "/var/run/docker.sock"
  hostSocket: "{{.Dir}}/sock/docker.sock"
message: |
  To run `kubectl` on the host (assumes kubectl is installed), run the following commands:
  ------
  export KUBECONFIG="{{.Dir}}/copied-from-guest/kubeconfig.yaml"
  kubectl ...
  ------
  To run `docker` on the host (assumes docker-cli is installed), run the following commands:
  ------
  export DOCKER_HOST=unix://{{.Dir}}/sock/docker.sock
  docker ...
  ------
```

Ready to run examples are available here:

https://github.com/lima-vm/lima/pull/1991/files

afbjorklund:
It should probably have said "Docker" and "Kubernetes" for the connection name, instead of "Lima" and "Lima"

```diff
-      name: 'Lima',
+      name: providerType,
       type: providerType,
       status: () => providerState,
```

deboer-tim:
> It should probably have said "Docker" and "Kubernetes" for the connection name, instead of "Lima" and "Lima"

Could the user give these a name when they create, similar to Podman machines or Kind instances? If not, then differentiating them would be great but I think we need to keep Lima in the title - otherwise you'll see these in places like 
the system tray or other UI controls (e.g. see the 'environment' in some of the Kube mockups) and without the context it could be confusing.

afbjorklund:
They have names (the instances/VMs), but the issue here is that it is the same name for both connections...

But it is quite obvious, the only issue is that "Docker endpoint" is not shown - only "Kubernetes endpoint"
If you go to the up-right-arrow page, then it _will_ show the details about each connection (name and url)

There is some [other issue](https://github.com/containers/podman-desktop/pull/2689) about where would be a good place to post the Lima instance name, in the status.
One alternative would be to just show "lima-k8s" like the "podman-machine-default", but then it's "too long"

So I think I will just leave it as-is, maybe revisit when lima instances work like podman machines do today.

----

```console
$ podman-remote-static machine ls
NAME                    VM TYPE     CREATED        LAST UP            CPUS        MEMORY      DISK SIZE
podman-machine-default  qemu        6 minutes ago  Currently running  1           2GiB        100GiB
$ limactl ls k8s
NAME    STATUS     SSH            VMTYPE    ARCH      CPUS    MEMORY    DISK      DIR
k8s     Stopped    127.0.0.1:0    qemu      x86_64    4       4GiB      100GiB    ~/.lima/k8s
```

```go
    name: prettyMachineName(machineInfo.name),
    type: 'podman',
    status: () => podmanMachinesStatuses.get(machineInfo.name),
```
```go
function prettyMachineName(machineName: string): string {
  let name;
  if (machineName === 'podman-machine-default') {
    name = 'Podman Machine';
  } else if (machineName.startsWith('podman-machine-')) {
    const sub = machineName.substring('podman-machine-'.length);
    name = `Podman Machine ${sub}`;
  } else {
    name = machineName;
  }
  return name;
}
```

For now, the GUI has a single text field with the name of _the_ instance... So you need to know the name anyway.

afbjorklund:
* https://github.com/containers/podman-desktop/pull/4488#issuecomment-1784058295

afbjorklund:
If the endpoint _had_ been showing, it should probably show the real `ssh://` url and not the "fake" `unix://` url.

Like it is showing the apiserver url, instead of the kubeconfig file... Or like `podman connection`/`docker context`.

afbjorklund:
If you change dockerd to use containerd for storage, then the images will be shared between them:

https://docs.docker.com/storage/containerd/

Similar to how it works with podman and cri-o

Similarly, you can configure docker to use buildkitd for building, instead of the old docker builder.

benoitf (approved):
miss that one

benoitf (commented):
would be better with unit tests

PR #4541, opened 2023-10-27
feat: show registered CLI versions

### LINKED ISSUE #3707: Compose settings: Show CLI version number

cdrage:
feat: show registered CLI versions

### What does this PR do?

* Shows the CLI version within the CLI Tools section of preferences
* The extension handles the updating / checking of the version and
  updates the CLI page accordingly.

Screenshot 2023-10-27 at 11 45 07 AM

Closes #3707

### How to test this PR?

1. Install Compose through onboarding
2. Open the CLI Tools section and you should now see the version number
   of the tool.

### DIFF
--- a/extensions/compose/src/extension.ts
+++ b/extensions/compose/src/extension.ts
@@ -24,8 +24,11 @@ import { ComposeGitHubReleases } from './compose-github-releases';
 import { OS } from './os';
 import * as handler from './handler';
 import { ComposeDownload } from './download';
+import * as path from 'path';
 
 let composeVersionMetadata: ComposeGithubReleaseArtifactMetadata | undefined;
+let composeCliTool: extensionApi.CliTool | undefined;
+const os = new OS();
 
 // Telemetry
 let telemetryLogger: extensionApi.TelemetryLogger | undefined;
@@ -51,7 +54,6 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
   // Create new classes to handle the onboarding sequence
   const octokit = new Octokit();
-  const os = new OS();
   const detect = new Detect(os, extensionContext.storagePath);
 
   const composeGitHubReleases = new ComposeGitHubReleases(octokit);
@@ -195,13 +197,50 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
   const provider = extensionApi.provider.createProvider(providerOptions);
   extensionContext.subscriptions.push(provider);
 
-  // Register the CLI extension, this will allow us to show the CLI version, update the CLI, installation, etc.
-  extensionApi.cli.createCliTool({
+  // Push the CLI tool as well (but it will do it postActivation so it does not block the activate() function)
+  // Post activation
+  setTimeout(() => {
+    postActivate(extensionContext).catch((error: unknown) => {
+      console.error('Error activating extension', error);
+    });
+  }, 0);
+}
+
+// Activate the CLI tool (check version, etc) and register the CLi so it does not block activation.
+async function postActivate(extensionContext: extensionApi.ExtensionContext): Promise<void> {
+  // The location of the binary (local storage folder)
+  const binaryPath = path.join(
+    extensionContext.storagePath,
+    'bin',
+    os.isWindows() ? composeCliName + '.exe' : composeCliName,
+  );
+  let binaryVersion = '';
+
+  // Retrieve the version of the binary by running exec with --short
+  try {
+    const result = await extensionApi.process.exec(binaryPath, ['--version', '--short']);
+    binaryVersion = result.stdout;
+  } catch (e) {
+    console.error(`Error getting compose version: ${e}`);
+  }
+
+  // Register the CLI tool so it appears in the preferences page. We will detect which version is being ran by
+  // checking the local storage folder for the binary. If it exists, we will run `--version` and parse the information.
+  composeCliTool = extensionApi.cli.createCliTool({
     name: composeCliName,
     displayName: composeDisplayName,
     markdownDescription: composeDescription,
     images: {
       icon: imageLocation,
     },
+    version: binaryVersion,
+    path: binaryPath,
   });
+}
+
+export async function deactivate(): Promise<void> {
+  // Dispose the CLI tool
+  if (composeCliTool) {
+    composeCliTool.dispose();
+  }
 }
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2447,6 +2447,16 @@ declare module '@podman-desktop/api' {
     displayName: string;
     markdownDescription: string;
     images: ProviderImages;
+
+    /**
+     * Within your extension, it is reccommended to implement your own functionality to check the current
+     * version number of the CLI tool. For example, parsing the information from the CLI tool's `--version` flag.
+     * Passing in path will also help to show where the CLI tool is expected to be installed.
+     * This is usually the ~/.local/share/containers/podman-desktop/extensions-storage directory.
+     * Note: The expected value should not include 'v'.
+     */
+    version: string;
+    path: string;
   }
 
   export type CliToolState = 'registered';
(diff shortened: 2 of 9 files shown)


extensions/compose/src/extension.ts
```diff
@@ -138,16 +145,45 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
+    extensionContext.storagePath,
+    'bin',
+    os.isWindows() ? composeCliName + '.exe' : composeCliName,
+  );
+  let binaryVersion = '';
+
+  // Use getCliVersion to try and get the version of the binary.
+  // If this fails, return output to console the error.
+  try {
+    binaryVersion = await getCliVersion(binaryPath, '--version');
```
dgolovin: Version detection should not be a part of extension activation function. All long running tasks should not be slowing down extension activation.

dgolovin: For `docker-compose` cli `version --short` options should be used, so we do not need to parse output and just use `stdout.trim()` as a version.
```
eskimo@degolovi-mac xterm.js % docker-compose version --short
2.20.1
eskimo@degolovi-mac xterm.js %
```

dgolovin (changes requested):

extensions/compose/src/extension.ts
cdrage: Do you have a recommendation where we should have it separate from activate? 

Just looking through some other extensions (kind, podman) and I see that they also call the CLI before activating.

cdrage: Maybe add a callback in the podman desktop API to go through and "activate" each CLITool if the provider has it?

cdrage:
I've made the code much simpler as well as have this issue opened to make it more reactively: https://github.com/containers/podman-desktop/issues/4543 I believe in the call we were going to incrementally add new features since there's a few others we need to work on (update mechanism as well).

extensions/compose/src/extension.ts
jeffmaury: The pre onboarding compose extension did everything in a delayed `postActivate` method launched through setTimeout

cdrage: Thanks @jeffmaury I'll modify the code that we update it via postActivate instead.

cdrage: Updated and moved to postActivation.

deboer-tim (approved):
As per chat, there should be a clear way for extensions to register when the cli is not installed and allow the UI to trigger install. Since that's not directly part of this I opened issue #4574.

Otherwise, this worked on my Mac and correctly picked up version #s.

cdrage:
> As per chat, there should be a clear way for extensions to register when the cli is not installed and allow the UI to trigger install. Since that's not directly part of this I opened issue #4574.
> 
> Otherwise, this worked on my Mac and correctly picked up version #s.

You're right, let's follow up in a separate PR for #4574 since it wasn't discovered until the PR that are missed in the scope of adding the installation detection / installation through CLI Tools section.

packages/main/src/plugin/api/cli-tool-info.ts
```diff
@@ -31,4 +31,6 @@ export interface CliToolInfo {
   state: CliToolState;
   images?: ProviderImages;
   extensionInfo: CliToolExtensionInfo;
+  version: string;
```
dgolovin:
```suggestion
  version?: string;
```

packages/main/src/plugin/api/cli-tool-info.ts
```diff
@@ -31,4 +31,6 @@ export interface CliToolInfo {
...
+  path: string;
```
dgolovin:
```suggestion
  path?: string;
```

dgolovin: Nit: Registration of new tool not necessarily have location and version, because it could be unknown at the time of registration. That would save couple lines when registering a tool, because there would be no path and version fields initialized with empty string.

dgolovin (changes requested):

packages/renderer/src/lib/preferences/PreferencesCliToolsRendering.svelte
```diff
@@ -44,6 +44,15 @@ import EmptyScreen from '../ui/EmptyScreen.svelte';
...
           <div role="region" class="ml-3 mt-2 text-sm text-gray-300">
             <Markdown markdown="{cliTool.description}" />
+            {#if cliTool.version !== ''}
+              <span
+                class="text-white-400 font-bold font-mono text-xs bg-charcoal-900 p-2 rounded-lg"
+                aria-label="cli-version">{cliTool.name} v{cliTool.version}</span>
+            {:else}
+              <span
+                class="text-white-400 font-bold font-mono text-xs bg-charcoal-900 p-2 rounded-lg"
+                aria-label="cli-version">Unable to retrieve binary version, try running the setup again.</span>
```
dgolovin: Nit: It should distinguish, `not installed` from 'installed, but cannot detect the version`, because error can happen at any stage of detection running  version command and parsing version command output.

extensions/compose/src/extension.ts
```diff
@@ -138,16 +145,63 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
+  }, 0);
+}
+
+// Activate the CLI tool (check version, etc) and register the CLi so it does not block activation.
+async function postActivate(extensionContext: extensionApi.ExtensionContext): Promise<void> {
+  // The location of the binary (local storage folder)
+  const binaryPath = path.join(
+    extensionContext.storagePath,
+    'bin',
+    os.isWindows() ? composeCliName + '.exe' : composeCliName,
```
dgolovin, line 215: there is no need to add '.exe', windows will look for it automatically.

packages/main/src/plugin/api/cli-tool-info.ts
dgolovin: image

There also should be `* SETUP NEEDED` or `* ACTIVE` on the left side of divider to show if CLI tool is detected or need installation.

dgolovin (changes requested):

extensions/compose/src/extension.ts
cdrage: I tried this without `.exe`  being added and it does not work on Windows (I'm using Windows 10 Enterprise).

cdrage:
Updated the PR:
- Do not show version if we error out, we will add more error reporting in: https://github.com/containers/podman-desktop/issues/4543
- Active / Inactive status: https://github.com/containers/podman-desktop/issues/4595
- Made version / path optional as per @dgolovin suggestion.

@dgolovin The scope of the PR is to add just the version, we will add the setup button in #4543 and we'll update the status in #4595 so we can incrementally add changes / implement CLI tools.

jeffmaury (approved):
LGTM

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2447,6 +2447,15 @@ declare module '@podman-desktop/api' {
...
     markdownDescription: string;
     images: ProviderImages;
+
+    /**
+     * Within your extension, it is reccommended to implement your own functionality to check the current
+     * version number of the CLI tool. For example, parsing the information from the CLI tool's `--version` flag.
+     * Passing in path will also help to show where the CLI tool is expected to be installed.
+     * This is usually the ~/.local/share/containers/podman-desktop/extensions-storage directory.
+     */
+    version: string;
```
feloy, line 2458: You could document the expected value does not include the `v`.

extensions/compose/src/extension.ts
dgolovin: Work fine on mine.

dgolovin (approved):

cdrage:
Fixed the rebasing + added the comment that @feloy suggested, thanks for the new reviews!

PR #4545, opened 2023-10-27
feat: table component

### LINKED ISSUE #4365: Reusable table component

deboer-tim:
### What does this PR do?

This is a clean, rebased branch of the table component after previous review comments in this PR.

We have impending technical debt where our expanding number of table components (images, volumes, pods, containers, soon adding Kubernetes objects) will cross with our future design ideas (tables should be sortable, columns configurable, environments groupable, etc). We can't copy/paste these features into every page, and deal with differences between them.

Each page passes an array of objects and Columns to the Table component, which is responsible for creating the basic layout and calling back to the Column renderer to render each cell in the table. The Table uses grid layout and provides enough abstraction that we can implement additional features later. (e.g. optional columns, column reordering, or resizing. grouping. etc)

The Table provides sorting & sort indicators, and tests are included. Support for 'object containers/grouping' in order to support Container list will be done separately since this is already a big PR.

Before:

Screenshot 2023-11-16 at 11 04 40 AM

After:

Screenshot 2023-11-15 at 12 40 46 PM

Fixes #4365.

### How to test this PR?

PR tests; separate PRs for Images and Volumes coming shortly.


packages/renderer/src/lib/ImageRow.svelte
```diff
@@ -0,0 +1,101 @@
...
+  renameImageModal = true;
+}
+
+function closeModals() {
+  pushImageModal = false;
+  renameImageModal = false;
+}
+</script>
+
+{#if column === 0}
```
lstocchi: I don't like much this check. If we add a new column between 0 and 1, we need to edit all *Row components.
Wouldn't it be better to use `slots`?

deboer-tim: Good point. I thought of using slots earlier, the problem is that slots are fixed and named - i.e. we need to pick how many and what they're called up front. I'm trying other options, so far I think the best is just using names instead of numbers here.

deboer-tim: Pushed another update, columns are named now (via the column header) instead of numbered. This is makes the code a little more clearer and adding/removing columns in the future is straightforward.

deboer-tim:
Updated commit to make columns named instead of numbered, and a little cleanup and renaming. Still not entirely sure how the column/row styling should work in the Column interface, still iterating there.

Added sorting just to prove it was possible. There is no sort indicator on the column header yet and a few columns should sort by value instead of alphabetically, but it works.

benoitf:
hi, from what I've seen, and I think @cdrage did some investigations long time ago, we should use grid system and not table

deboer-tim:
Pinged Charlie, he's checked all his old branches and forks and can't find any of the code from that investigation.

I'm somewhat hesitant to switch to grid here since that will hold this up and do 'multiple things' in one PR... but adopting this on a page is also the ideal time to switch, so I'll give it a try.

benoitf:
https://github.com/containers/podman-desktop/issues/885

deboer-tim:
Updated description, added video, and force-pushed the latest rebased on main. A lot has changed since the last update and would appreciate feedback so that I can address issues before this moves out of draft state.

packages/renderer/src/lib/image/ImagesList.svelte
```diff
@@ -214,6 +169,50 @@ function computeInterval(): number {
...
+let table: Table;
+
+const columns: Column[] = [
+  new Column('Status', 'center', '70px'),
+  new Column('Name', 'left', '3fr'),
+  new Column('Age', 'left', '1fr'),
+  new Column('Size', 'right', '1fr'),
+];
+
+const tableHelper = new (class extends TableHelper {
```
benoitf: As early feedback I would say I'm not finding convenient this kind of declaration from consumer POV

packages/renderer/src/lib/volume/VolumesList.svelte
```diff
@@ -185,6 +163,48 @@ async function fetchUsageData() {
...
+const tableHelper = new (class extends TableHelper {
+  selectable(volume: any): boolean {
+    return volume.inUse;
+  }
+
+  disabledText(): string {
+    return 'Volume is used by a container';
+  }
+
+  compare(column: string): ((object1: any, object2: any) => number) | undefined {
```
benoitf: should we have dedicated separate objects instead of one helper ?

like one comparator, one selectable

packages/renderer/src/lib/volume/VolumesList.svelte
```diff
@@ -185,6 +163,48 @@ async function fetchUsageData() {
...
+      return (a, b) => a.shortName.localeCompare(b.shortName);
+    } else if (column === 'Age') {
+      return (a, b) => {
+        const au = moment().diff(a.created);
+        const bu = moment().diff(b.created);
+        return au - bu;
+      };
+    } else if (column === 'Size') {
+      // TODO: should sort by value, not alphabetically (e.g. 1 Gb > 2 Mb)
+      //return (a, b) => a.humanSize.localeCompare(b.humanSize);
```
benoitf: I would expect to not have any, any in signature but a typed field so I know what I can get from there

packages/renderer/src/lib/volume/VolumesList.svelte
```diff
@@ -185,6 +163,48 @@ async function fetchUsageData() {
 function gotoCreateVolume(): void {
   router.goto('/volumes/create');
 }
+
+let selectedItemsNumber: number;
+let table: Table;
+
+const columns: Column[] = [
+  new Column('Status', 'center', '70px'),
```
benoitf: should we not add the sort comparator on the column object so we have dedicated type there ?

benoitf:
```ts
const statusColumn = new Column<ExpectedType>('Status', 'center', '70px');
statusColumn.setSortComparator( and here we can sort using ExpectedType)
```

cdrage (changes requested):
Amazing! I'm able to actually organize it by age / alphabetical whenever I want. so big 👍 

With regards to table helper, I'm assuming we have to make a new helper for each list scenario for Pods / Containers / etc?

I'm good with the concept of using TableHelper to make it generic and then "extending" it for different use cases. As honestly it seems like the best way to make it as "generic" as possible considering how difficult it is with other tables / etc.

I'm assuming this will scale well with the other lists we are implementing too

packages/renderer/src/lib/volume/VolumesList.svelte
```diff
@@ -215,76 +235,14 @@ function gotoCreateVolume(): void {
...
-        {/each}
-      </tbody>
-    </table>
+    <Table
+      bind:this="{table}"
+      bind:selectedItemsNumber="{selectedItemsNumber}"
+      objects="{volumes}"
+      columns="{columns}"
+      row="{VolumeRow}"
+      helper="{tableHelper}">
```
benoitf: I'm not sure about the helper name, it looks too generic

packages/renderer/src/lib/image/ImagesList.svelte
```diff
@@ -214,6 +169,50 @@ function computeInterval(): number {
...
+  selectable(image: any): boolean {
+    return image.inUse;
+  }
+
+  disabledText(): string {
+    return 'Image is used by a container';
+  }
+
+  compare(column: string): ((object1: any, object2: any) => number) | undefined {
+    if (column === 'Status') {
```
cdrage: switch statement instead of what if may be better / cleaner?

packages/renderer/src/lib/volume/VolumesList.svelte
```diff
@@ -215,76 +235,14 @@ function gotoCreateVolume(): void {
...
-            </td>
-          </tr>
-          <tr><td class="leading-[8px]">&nbsp;</td></tr>
-        {/each}
-      </tbody>
-    </table>
+    <Table
+      bind:this="{table}"
+      bind:selectedItemsNumber="{selectedItemsNumber}"
+      objects="{volumes}"
```
benoitf: data instead of objects ?

packages/renderer/src/lib/volume/VolumesList.svelte
```diff
@@ -185,6 +163,48 @@ async function fetchUsageData() {
...
+  selectable(volume: any): boolean {
+    return volume.inUse;
+  }
+
+  disabledText(): string {
+    return 'Volume is used by a container';
+  }
+
+  compare(column: string): ((object1: any, object2: any) => number) | undefined {
+    if (column === 'Status') {
```
cdrage: switch statement instead?

cdrage (dismissed):

packages/renderer/src/lib/volume/VolumeRow.svelte
```diff
@@ -0,0 +1,34 @@
+<script lang="ts">
+import { router } from 'tinro';
+import StatusIcon from '../images/StatusIcon.svelte';
+import VolumeIcon from '../images/VolumeIcon.svelte';
+import VolumeActions from './VolumeActions.svelte';
+import type { VolumeInfoUI } from './VolumeInfoUI';
+
+export let object: any;
```
benoitf: I don't see why it should be with any type

packages/renderer/src/lib/volume/VolumesList.svelte
```diff
@@ -215,76 +235,14 @@ function gotoCreateVolume(): void {
...
-          <tr><td class="leading-[8px]">&nbsp;</td></tr>
-        {/each}
-      </tbody>
-    </table>
+    <Table
+      bind:this="{table}"
+      bind:selectedItemsNumber="{selectedItemsNumber}"
+      objects="{volumes}"
+      columns="{columns}"
+      row="{VolumeRow}"
```
benoitf: if name is row, I would expect that I provide a row entry to the table and then my table has only one row
here it looks it's more like a template here that we use against all the data

--> rowComponent or rowTemplate

benoitf: thinking more on the topic, I'm still wondering if we need this template per row where we need to go through columns
if we add a column we should not forget to customize the row template --> it'll lead to errors

as inside the row, you're checking the column information

so, probably we could set the component per column with proper type as well

ageColumn.setRenderingTemplate(AgeTemplate)

(and it's in charge of only rendering the age)

deboer-tim:
Added new commit:

Lots of changes based on review comments:
- Added a Type to Column and moved the comparator to it. This makes the Column[] creation a little longer, but the typing works well and keeps the TableHelper minimal.
- Renamed 'objects' to 'data'.
- Added the type to ImageRow/VolumeRow - but these are about to change drastically in the next step...
- Florent had a comment about trying ageColumn.setRenderingTemplate() similar to setComparator(). I had thought about this before and was worried about lots of tiny components, but gave it a shot. Many of these are indeed tiny, but it does provide nice componentization (possible to share some of these between types later?) and gets rid of the big if/switch block.
- Every column needs a renderer - so removed the function and just made it part of the constructor for simpler code.
- At this point it seemed like actions should just be a regular column too. Making this change simplified things a little.

Overall I like these changes; simpler implementation and easier/safer to use for consumers. I am still having trouble with the grid layout misbehaving though.

packages/renderer/src/lib/image/ImagesList.svelte
deboer-tim: I don't love this syntax either. I had an interface before, but I was worried that adding capability in the future (like the previous comparator function) meant that all consumers would be required to change even if they don't care about the functionality yet.

We could stop it at the current capability and extend in other ways. Or maybe switch it to be more like Column where you create an instance and customize via functions? I will keep trying different ways.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: Selection is per object/row so it needs to stay at the 'helper' level, but I've moved comparators to the columns.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: Fixed in latest commit.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: Fixed in latest commit.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: Fixed in latest commit.

packages/renderer/src/lib/image/ImagesList.svelte
deboer-tim: Used alternate suggestion to move comparator (and renderer) to column level.

packages/renderer/src/lib/volume/VolumeRow.svelte
deboer-tim: Fixed in latest commit.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: Moved renderer to each column in latest commit. This leads to a bunch of small components, but it removes a big if/switch statement and IMHO makes the code a little cleaner and more obvious. Thanks for the suggestion.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: Moved the comparator to each Column as suggested in another comment, no more if/switch.

packages/renderer/src/lib/image/ImagesList.svelte
deboer-tim: Pushed another commit to change TableHelper to be more like Column definition. Much simpler for clients to create, please take a look.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: Not sure what else to call it? 🤷🏻‍♂️ Let me know if you have any suggestions.

benoitf: maybe some dedicated methods instead of a generic one. (one entry per helper method/goal) ?

packages/renderer/src/lib/image/ImagesList.svelte
```diff
@@ -214,6 +173,32 @@ function computeInterval(): number {
...
+
+const columns: Column<ImageInfoUI>[] = [
+  statusColumn,
+  nameColumn,
+  ageColumn,
+  new Column<ImageInfoUI>('Size', 'right', '1fr', ImageColumnSize),
+  new Column<ImageInfoUI>('Actions', 'right', '150px', ImageColumnActions),
+];
+
+const tableHelper = new TableHelper<ImageInfoUI>('image', image => image.inUse, 'Image is used by a container');
```
benoitf: as commented before, maybe here we would need a method dedicated to rows (like you had for columns)

so instead of helper, you apply information on the "rowRenderer"

const imageListRowRender = new RowRender<ImageInfoUI>();
imageListRowRender.setDisable(image => image.inUse, 'Image is used by a container');

(it can be any name but usually a helper is something that we're using before giving the result to the final object, not a parameter)

it will avoid to have constructors with 5-10+ items as we'll probably add multiple options

packages/renderer/src/lib/image/ImagesList.svelte
```diff
@@ -214,6 +173,32 @@ function computeInterval(): number {
   // every day
   return 60 * 60 * 24 * SECOND;
 }
+
+let selectedItemsNumber: number;
+let table: Table;
+
+let statusColumn = new Column<ImageInfoUI>('Status', 'center', '70px', ImageColumnStatus);
```
benoitf: I think we'll reach soon a limit on constructor parameters

so while the name of the column looks quite important, I think the other parameters are kind of optional

so either we move to dedicated method or we provide JSON like parameters with an interface so it'll easier to grow/add mandatory parameters/optional parameters

```ts
new Column<ImageInfoUI>('Status', {align: 'center', width: '70px'});
```
or 

```ts
new Column<ImageInfoUI>('Status')
column.width = '70px';
column.align = 'center';
```

or use fluent API like
```ts
Column.name('Status').width('70px').align('center').create();
```

benoitf (commented):
much more simple, nice work
I still believe we might simplify again as number of parameters will grow and we won't be able to scale easily or without like lots of `param1, undefined, undefined, param2` like in parameters at some points

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  return columnWidths.join(',');
+}
+</script>
+
+<div class="w-full" class:hidden="{data.length === 0}" role="table">
+  <!-- Table header -->
+  <div
+    class="grid grid-flow-row grid-cols-[{gridCol}] gap-x-0.5 mx-5 h-7 sticky top-0 bg-charcoal-700 text-xs text-gray-600 font-bold uppercase z-[2]"
```
lstocchi: You cannot do `grid-cols-[{gridCol}]` or the style will not be applied. You always have to use static class https://tailwindcss.com/docs/content-configuration#dynamic-class-names
You have to pass the class already defined or think some other way to do it

before
image
after
image

lstocchi (commented):
I tried it and looks fine. Waiting to being ready to play with it more.

packages/renderer/src/lib/table/TestTable.svelte
```diff
@@ -13,21 +13,23 @@ const people: object[] = [
...
 nameCol.setComparator((a, b) => a.name.localeCompare(b.name));
 
-const ageCol: Column<any> = new Column('Age', 'right', '1fr', TestColumnAge);
+const ageCol: Column<any> = new Column('Age', { align: 'right', renderer: TestColumnAge });
 ageCol.setComparator((a, b) => a.age - b.age);
 
 const columns: Column<any>[] = [nameCol, ageCol];
 
-const tableHelper = new TableHelper<any>('people', person => person.age < 50, 'People over 50 cannot be deleted');
+const row = new Row<any>();
```
benoitf: shouldn't be type to {name: string; age: string} ?

deboer-tim: Added Person type in latest commit.

packages/renderer/src/lib/table/Table.svelte
deboer-tim: Thank you! I was getting really confused because the behaviour was inconsistent, and had forgotten this should not work. Latest commit has a fix setting the style directly on the element.

packages/renderer/src/lib/image/ImagesList.svelte
deboer-tim: I went with the first option in the latest commit. It does complicate things a bit because it makes sense to have most of the parameters optional, but that makes more sense and keeps the use simpler.

packages/renderer/src/lib/image/ImagesList.svelte
deboer-tim: I went with a Row class instead of TableHelper in the second latest commit, and moved the 'kind' as a standalone property. The Row is only for setting selected/disabled text for now. I debated whether selection should be a completely standalone thing, but I think using Row will lead to simple use long term as we add other options.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: As per other comment, TableHelper is gone in latest commit.

deboer-tim:
Clean branch (without Images/Volumes) pushed, PR comment updated, and draft status removed. Ready for final review.

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -0,0 +1,165 @@
...
+  let columnWidths: string[] = ['20px', '32px'];
+
+  // custom columns
+  for (const column of columns) {
+    if (column.info.width) {
+      columnWidths.push(column.info.width);
+    } else {
+      columnWidths.push('1fr');
+    }
+  }
```
feloy: You could use a map here:

```
const columnWidths = ['20px', '32px', ...columns.map(c => c.info.width ?? '1fr'), '5px'];
```

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -0,0 +1,165 @@
...
+  ? data.filter(object => row.selectable?.(object)).every(object => object.selected)
+  : false;
+
+function toggleAll(checked: boolean) {
+  if (!row.selectable) {
+    return;
+  }
+  const toggleData = data;
+  toggleData.filter(object => row.selectable?.(object)).forEach(object => (object.selected = checked));
+  data = toggleData;
```
feloy: What is the purpose of using the `toggleData` temp variable? As toggleData is a reference to data, it does not seem necessary to me

packages/renderer/src/lib/table/TestTable.svelte
```diff
@@ -0,0 +1,40 @@
...
+  age: number;
+};
+
+const people: Person[] = [
+  { name: 'John', age: 57 },
+  { name: 'Henry', age: 27 },
+  { name: 'Charlie', age: 43 },
+];
+
+const nameCol: Column<Person> = new Column('Name', { width: '3fr', renderer: TestColumnName });
```
feloy, line 21: Is it really Column&lt;Person>, or Column&lt;string>, for the name only?

packages/renderer/src/lib/table/TestTable.svelte
```diff
@@ -0,0 +1,40 @@
...
+nameCol.setComparator((a, b) => a.name.localeCompare(b.name));
+
+const ageCol: Column<Person> = new Column('Age', { align: 'right', renderer: TestColumnAge });
```
feloy, line 24: Column&lt;number>?

packages/renderer/src/lib/table/TestTable.svelte
```diff
@@ -0,0 +1,40 @@
...
+ageCol.setComparator((a, b) => a.age - b.age);
+
+const columns: Column<any>[] = [nameCol, ageCol];
```
feloy:
```suggestion
const columns: Column<Person>[] = [nameCol, ageCol];
```

packages/renderer/src/lib/table/table.ts
```diff
@@ -0,0 +1,83 @@
...
+  ) {}
+
+  /**
+   * Set a comparator used to sort the data by the values in this column.
+   *
+   * @param comparator
+   */
+  setComparator(comparator: (object1: Type, object2: Type) => number) {
+    this.comparator = comparator;
+  }
```
feloy: Did you consider passing this `comparator` as part of the `ColumnInformation`?
For the user, it would be more *functional*.

packages/renderer/src/lib/table/table.ts
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Set a function to be used to determine which objects in the data can be selected.
+   *
+   * @param selectable a function that returns false when the object should not be selectable
+   * @param disabledText text to display as a tooltip when selection is disabled
+   */
+  setSelectable(selectable: (object: Type) => boolean, disabledText: string) {
+    this.selectable = selectable;
+    this.disabledText = disabledText;
+  }
```
feloy: same here, `selectable` could be part of some `RowInformation`. `disabledText` still being set during the construction, depending on the `selectable` option.

packages/renderer/src/lib/table/Table.svelte
deboer-tim: Yup, updated in latest commit.

packages/renderer/src/lib/table/Table.svelte
deboer-tim: You can see this was actually unchanged from before. Svelte only detects changed objects, not internal state, so sometimes you have to explicitly assign something to see the change visually. I did some testing and this doesn't seem to be required anymore, so I removed it.

The same thing existed further down on line 64 with sorting. This one is required, but I can see the same question arising so I changed it to an explicit 'data = data.sort()', and added the required no-mutation comment. I don't think this is any better code-wise, but at least it's obvious it was intentional.

packages/renderer/src/lib/table/TestTable.svelte
deboer-tim: Fixed in commit.

packages/renderer/src/lib/table/table.ts
deboer-tim: tbh I thought passing the Type down through ColumnInformation would be an issue, but seems to be fine in TypeScript so I did this in latest commit - everything goes into ColumnInfo and can be done in the constructor.

packages/renderer/src/lib/table/table.ts
deboer-tim: Yes, since Type isn't an issue and it makes sense to have everything the same, I've added RowInformation and done this. The one point I don't love is that it means selectable can be set and not disabledText or vice versa, but that's extremely minimal and doesn't really matter.

Since selection is optional I finally went ahead and made the entire column disappear if selection isn't allowed. (previously the blank column would be there but no checkboxes)

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -0,0 +1,161 @@
...
+
+export let kind: string;
+export let data: any[];
+export let columns: Column<any>[];
+export let row: Row<any>;
+
+// number of selected items in the list
+export let selectedItemsNumber: number = 0;
+$: selectedItemsNumber = row.info.selectable
+  ? data.filter(object => row.info.selectable?.(object)).filter(object => object.selected).length
```
lstocchi, line 24: any reason why you don't merge the two filters? ` data.filter(object => row.info.selectable?.(object) && object.selected)`

lstocchi (approved):
LGTM

packages/renderer/src/lib/table/Table.svelte
deboer-tim: None at all, I just copied from ImagesList and VolumesList and didn't think to fix. :) I'll do it in follow-up PR.

packages/renderer/src/lib/table/table.ts
```diff
@@ -0,0 +1,82 @@
...
+   * Tooltip text to show when row selection is disabled.
+   */
+  readonly disabledText?: string;
+}
+
+/**
+ * A table row.
+ */
+export class Row<Type> {
+  constructor(readonly info: RowInformation<Type>) {}
```
feloy, line 81: You could raise an error in the constructor (intended for the developer) if selectable and disabledText are both not set.

packages/renderer/src/lib/table/Table.svelte
feloy: Now that you tell me the reason, I remember I effectively had to write some `data = [...data]` in some other PR, to make the refresh happen. Not sure what is the recommended way to handle this

feloy: In this article, it is explained that `data = data` is enough to trigger the Svelte changes.

https://javascript.plainenglish.io/update-arrays-and-objects-in-svelte-40c6a84f8a28

packages/renderer/src/lib/table/table.ts
deboer-tim: Actually, I retract my previous point :) b/c I think PodsList is the reason not to do this: all pods are selectable, so forcing PodsList to provide disabledText actually doesn't make sense.

Technically it still doesn't make sense if someone provides disabledText and not selectable (i.e. nothing can be selected but you provide text for that case), but that seems very marginal and not something we need to protect against.

packages/renderer/src/lib/table/Table.svelte
deboer-tim: I've seen it elsewhere too so it seem to be a common way of handling it. For now though sort is the only place that needs it and eslint comment make it clear the code is intentional.

feloy (approved):

PR #4546, opened 2023-10-27
fix: reduce website footer

### LINKED ISSUE #4299: Reduce website footer

The website footer has a Documentation column with 8 links, a Links column with 7, and then a copyright and license. In the window I was loading the website that was 40% of the page height, which to me looks too big and unpolished.

I think we should cut it down to ~4 links per side. Docs we should be able to reduce to the more common pages, and for links we could remove the Blog (already on the header), Github (also on the header), and maybe cut down on the duplicate text or merge two onto one line.

Put copyright and license on one line?

Move some of this to a resources or community page.

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Ok, here's my pass at reducing the vertical height of the footer. Long term I think we can and should do more, this is just a quick initial cleanup pass:
- Left: removed three links (so both columns are down to 5). These are a judgement call, e.g. I felt like 'Docker migration' is a bit of marketing we'd want to leave for now.
- Right: remove Blog (it's already at the top of the page).
- Right: remove planning and roadmap, both not really useful link and accessible from the Github link above.
- Put copyright and license on one line.

Overall, this reduces the footer height from 11 lines of text to 7.

Before:

Screenshot 2023-10-27 at 4 30 37 PM

After (updated to latest commit):

Screenshot 2023-10-30 at 2 43 53 PM

Fixes #4299.

### How to test this PR?

`yarn website:dev`

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -351,18 +351,6 @@ const config = {
                 to: '/docs/installation',
               },
               {
-                label: 'Onboarding for containers',
-                to: '/docs/onboarding/containers',
-              },
-              {
-                label: 'Onboarding for Kubernetes',
-                to: '/docs/onboarding-for-kubernetes',
-              },
-              {
-                label: 'Working with containers',
-                to: '/docs/working-with-containers',
-              },
-              {
                 label: 'Migrating from Docker',
                 to: '/docs/migrating-from-docker',
               },
@@ -384,10 +372,6 @@ const config = {
             title: 'Links',
             items: [
               {
-                label: 'Blog',
-                to: '/blog',
-              },
-              {
                 label: 'GitHub',
                 href: 'https://github.com/containers/podman-desktop',
               },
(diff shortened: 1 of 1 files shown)


benoitf (commented):
IMHO I think roadmap is interesting as a direct link

but probably for Chat, we should reduce the numbers of links as it looks like it's too many

jeffmaury:
> IMHO I think roadmap is interesting as a direct link
> 
> but probably for Chat, we should reduce the numbers of links as it looks like it's too many

Keep the discord one and replace the others chat links by a link to https://github.com/containers/podman-desktop#communication

deboer-tim:
@benoitf I didn't think the ([previous link](https://github.com/containers/podman-desktop/projects?type=beta)) was very useful. How about pointing them to the [planning board](https://github.com/orgs/containers/projects/4) or [current sprint](https://github.com/orgs/containers/projects/4/views/8) instead?

@jeffmaury Makes sense, will do.

benoitf:
@deboer-tim yes as you suggested, we need a redirect link, current link is not good 👍  (I would prefer 'current sprint' link as default to backlog is not really great as well (you need to click on a link again)

I agree that footer links are like direct calls and as Jeff pointed out, when there are several links, just take one

deboer-tim:
Pushed another commit:
- Reduced four chat links to just Discord and a link to Communication section in the readme. Added back in a link to the sprint planning board.
- Removed Compose from the doc links to even things out again.

slemeur:
Good simplification!

slemeur (approved):

benoitf (approved):

cdrage (approved):

PR #4556, opened 2023-10-30
fix: report error if container engine action fails in details page (#3796)

### LINKED ISSUE #3796: No errors reported within Container Engine details page
### Bug description

If a container engine fails to start within the container engine details page (Settings > Resources, click on Podman details on a machine, then Start), no error is reported within the UI, it just goes back to stopped state. If you go back to the Resources page the error isn't listed there either (it only shows if you try to start from within the Resources page).

Yes, the user could look at the Logs or open the Task Manager, but the error on the action should be similar to the Resources page.

### Operating system

macOS 13.5.1

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR adds the error button that get shown if a container action fails

error_container-action

it resolves #3796

### How to test this PR?

1. execute any action that you know it will fail from the details page of a container engine
2. you should see the error button

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionActions.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionActions.spec.ts
@@ -25,7 +25,6 @@ import type {
   ProviderKubernetesConnectionInfo,
 } from '../../../../main/src/plugin/api/provider-info';
 import type { IConnectionStatus } from './Util';
-import { getProviderConnectionName } from './Util';
 import PreferencesConnectionActions from './PreferencesConnectionActions.svelte';
 
 const containerProviderInfo: ProviderInfo = {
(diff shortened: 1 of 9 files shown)


benoitf (commented):
it looks like the major part of this PR is about a code refactoring, then the report of the error is mainly about only adding
`<ConnectionErrorInfoButton status="{connectionStatus}" />` but it's like hard to see what is new and what is refactoring
so I'm wondering if it should be split at least into two commits

lstocchi:
@benoitf done 👍

benoitf (approved):
thanks @lstocchi

deboer-tim (approved):

PR #4558, opened 2023-10-30
feat: reorganize doc navigation by provider

### LINKED ISSUE #4557: Main doc navigation section for each provider

Currently the topics for a provider like Lima are spread out in several places (at least three in this case - onboarding for containers, onboarding for Kubernetes, and Kubernetes). Ditto for Podman, Kind, Minikube, and OpenShift Local.

As a user I know if I'm using provider X (or not) and if I'm looking for information on using X within Podman Desktop, vs how to use Podman Desktop. I expect to see an item in the main navigation for the provider I'm using.

Add a section to the main doc navigation for every provider (Podman, Lima, Kind, Minikube, OpenShift), with an intro on the main page and a common sub-nav to view installation instructions.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Switches to give every provider (Podman, Lima, Kind, Minikube, OpenShift) its own root folder in the doc navigation (and the child '/installing' is normally where installation instructions are).

Why? Currently the topics for several of these are spread out between multiple sections, e.g. there is Lima information in onboarding for containers, onboarding for Kubernetes, and Kubernetes; installing Podman is under Onboarding for containers but setting the default connection is in Working with containers. As a user I know if I'm using provider X and if I'm looking for information on using X within Podman Desktop, vs how to use Podman Desktop.

After this move, there was only one page in each of the 'onboarding' sections - the onboarding containers overview and working with an existing Kubernetes. These two pages were moved into the respective /containers and /kubernetes folders so that the section could be removed. /working-with-containers was reduced to /containers to match all the other folders.

As much as possible (except for links) I've tried not to edit the content of any page. In a couple cases I had to merge pages that were near duplicates, and likewise there are a couple places where we didn't have a single landing page before (e.g. OpenShift) where I added an initial page.

Added redirects in the docusaurus config, and as much as possible updated links in existing files to point to the new locations.

Before:

Screenshot 2023-10-30 at 2 48 13 PM

After:

Screenshot 2023-10-30 at 2 48 29 PM

Fixes #4557.

### How to test this PR?

`yarn website:dev`


cdrage:
This looks great, the only thing that bothers me is that "Migrating from Docker" is located in between Working with Compose and Working with Kubernetes. Could we change the order so all the "Working with" is grouped together?

deboer-tim:
Yeah, I think that makes sense. Moved Migration from Docker up so that it is immediately under the Podman section.

afbjorklund:
I don't understand why most information should be duplicated by provider, or why it wasn't enough with two topics...
Thought it was bad enough with podman/docker (in "containers") and kind/minikube (in "kubernetes"), **not** adding Lima.

slemeur:
@afbjorklund As we are supporting multiple tools, like Kind, Lima, Minikube, Podman, it offers versatility for users, but also presents challenge for organizing the content effectively. 

There should not be duplicated content, each section should be tailored to the specific nuances and use cases of the respective tool. But they should try to be as consistent as possible in term of structure, so it would help the users who might switch between tools or looking to compare them. 

By organizing the sections this way, we can provide a comprehensive, easy-to-navigate resource that caters to users’ different needs and preferences regarding Kubernetes (or containers) tools. But still, this structure allows for both depth and breadth in covering Kind, Minikube, and Lima, while keeping the content accessible and practical.

afbjorklund:
You are free to organize the documentation, as you see best for the Podman Desktop product and users...
But I will defer the details from the Lima provider, to the upstream projects: podman/docker/k3s/kubeadm

deboer-tim:
> You are free to organize the documentation, as you see best for the Podman Desktop product and users... But I will defer the details from the Lima provider, to the upstream projects: podman/docker/k3s/kubeadm

Case in point Anders, as a user I thought the two docs on "Custom Lima instance" and "Creating a Lima instance" were duplicates until this PR brought them in one place and it dawned on me. I extracted the common cli install bits here and tried to make the difference a little more obvious, but IMHO this still needs improvement - but this change gives a logical place to put that 'Lima is special and can do both' kind of thing.

What's here should just be a quick overview, info on what's possible and how to use it within Podman Desktop. Wherever it makes sense we should defer information on Lima (or Podman, or ...) to the actual docs for that provider.

slemeur (approved):
LGTM

afbjorklund:
Just know that I copied a fair amount of text, only to do a replacement of "kind" for "minikube" or for "lima"

The actual differences were only in a couple of sentences (or commands), the rest was just basic Kubernetes

afbjorklund:
For the Lima project, the adaptions for Docker or for Kubernetes will be pushed to the downstream projects.

The wrappers and the templates will probably remain though, so the information in the PD docs should be valid.
But there won't be any custom preloads or dashboards, or whatever other features that the distributions* have...

\* the Kubernetes distributions such as minikube or kind, or what the Podman Machine is doing for podman

jeffmaury (approved):
LGTM

PR #4560, opened 2023-10-30
feat: show kubernetes context details in settings

### LINKED ISSUE #3919: KubeUI: Introduce a screen to list the context from a Kube config file

I've repeatedly created Kind servers and never cleaned up their kube config. Likewise, Sandbox has an issue where I need to recreate it and the old kube config is stale. Over time I've ended up with content that Podman Desktop created but has no way to clean up.

We should provide a basic UI to see the list of clusters and contexts, and allow deletion. It should be easy to tell what's local vs remote, and which point to Kubernetes providers within Podman Desktop (i.e. show me which contexts are my Kind or Sandbox or OpenShift Local instance vs random other contexts).

I've been meaning to open this before, but was triggered by #3914 (provide a way to overwrite a context when creating a sandbox instance) since this would be another way of getting out of that situation.

_No response_

### Additional context

Design is ready in https://design.penpot.app/#/workspace/76981dd7-ec3b-802f-8001-9a56ef3fbc9d/18f3baa5-c6db-81d1-8002-9f756ec487eb?page-id=2419dd66-911a-807e-8003-43a20ad0ffc1

cdrage:
feat: show kubernetes context details in settings

### What does this PR do?

* The first implementation for the Kubernetes context screen
* Shows Kubernetes context details (name, cluster, server URL)

In future PR's:
* Ways to determine the name / assign correct icon's to the provider
* Reactively update details (amount of pods, deployments, etc.)

Screenshot 2023-10-30 at 7 04 07 PM

Closes: https://github.com/containers/podman-desktop/issues/3919

Completes first phase of https://github.com/containers/podman-desktop/issues/4416

### How to test this PR?

* Have a ~/.kube/config
* Click on Kubernetes in the settings


cdrage:
UI / implementation notes @mairin @benoitf 

UI:
* Implemented the design based upon Providers / CLI Tools section (for now!)
* Will improve as we complete other parts of the epic: https://github.com/containers/podman-desktop/issues/4416 as new connection information is added

Implementation:
* Implemented all that can be done from the config file. The rest of the information (online status, amount of pods, deployments, etc.) needs to be determined in the background / "loading" as some kubernetes clusters take a while to query, etc. This will be done in: 

Follow up issues after this PR:
- https://github.com/containers/podman-desktop/issues/4561
- https://github.com/containers/podman-desktop/issues/4562

deboer-tim:
Looking good 👍🏼, will do full review tomorrow. The only things that are standing out quickly are:
- The charcoal-900 on cluster/server/user looks like black, wondering if it should be backed off to be closer to design.
- No vertical line in the design? Curious why this was added.

cdrage:
> Looking good 👍🏼, will do full review tomorrow. The only things that are standing out quickly are:
> 
> * The charcoal-900 on cluster/server/user looks like black, wondering if it should be backed off to be closer to design.
> * No vertical line in the design? Curious why this was added.

Thank you! I changed it to 800 instead.

The reason for the vertical line is so that it's consistent with the Providers page / CLI Tools / etc since the format is quite similar.

When #4562 is added the design will change as we add more information to the left since such as connection information, pods, deployments, etc.

slemeur:
This is looking great!

@mairin @cdrage : In my experience the names given to the kubecontext are always quite long. I'm a bit worry about how it will scale if the name is awfully long.

Here, with one context created by connecting to DevSandbox:
Screenshot 2023-10-31 at 14 26 09

I'm wondering if the name of the context shouldn't be better as a title of the card, so it would benefit from the full width card, be more readable than split on multiple lines? If it requires more space than available, then we could a text-overflow and a tooltip?

deboer-tim:
Yes, +1 to making it a full-width title please. :)

Screenshot 2023-10-31 at 10 04 13 AM

cdrage:
> Yes, +1 to making it a full-width title please. :)
> 
> Screenshot 2023-10-31 at 10 04 13 AM

I'll make that change (and remove the line too while I'm at it..) Thanks for pointing it out. Yay long cluster names

mairin:
Screenshot from 2023-10-31 06-46-24
Screenshot from 2023-10-31 06-47-37

I love seeing this become real :) I noticed some scaling issues on my screens (this is on Fedora 38.)

deboer-tim:
A few more concerns:
- What is the current context and how do I change it? I found a mockup with tiles that had a blue left border, not sure if it should be the same with these rows (and if it should be purple).
- The server link should be purple (maybe link button?), not blue.
- I could only see 3ish of my current 5 contexts on a small window, and the last screen cap above only has 2. I think we need a full-width title, but that might make the vertical space worse in some cases. I think we need to keep an eye on vertical space so more fit on screen at a time, e.g. reduce the top/bottom padding of those values and maybe use ellipsis with tooltips so long cluster/server/user values don't go two lines.

cdrage:
> A few more concerns:
> 
> * What is the current context and how do I change it? I found a mockup with tiles that had a blue left border, not sure if it should be the same with these rows (and if it should be purple).

Will be added in https://github.com/containers/podman-desktop/issues/4417 this PR is only implementing the page to display the context (for now).

> * The server link should be purple (maybe link button?), not blue.

Updated! Thanks for pointing that out, I've actually opened up https://github.com/containers/podman-desktop/pull/4577 when implementing this as well. So if someone is testing this PR, you'll notice a very subtle unaligned link, which will be fixed in https://github.com/containers/podman-desktop/pull/4577

> * I could only see 3ish of my current 5 contexts on a small window, and the last screen cap above only has 2. I think we need a full-width title, but that might make the vertical space worse in some cases. I think we need to keep an eye on vertical space so more fit on screen at a time, e.g. reduce the top/bottom padding of those values and maybe use ellipsis with tooltips so long cluster/server/user values don't go two lines.

Agreed! Added the full-width title now. 

Screenshot 2023-10-31 at 5 19 09 PM

Ready for another review 👍 

To note, the cluster information is centered until we implement this section:
 
Screenshot 2023-10-31 at 5 24 15 PM

lstocchi (commented):
LGTM, just a tiny issue with the hover background that is a bit smaller with long names

hover-background

cdrage:
> LGTM, just a tiny issue with the hover background that is a bit smaller with long names
> 
> hover-background [ hover-background ](https://user-images.githubusercontent.com/49404737/279945135-f4c53f53-c0d8-4d18-b3cb-6566d3d62136.gif) [ ](https://user-images.githubusercontent.com/49404737/279945135-f4c53f53-c0d8-4d18-b3cb-6566d3d62136.gif)

Thank you for the review! I'll fix it, it looks like we have an odd text-wrapping problem too.

cdrage:
@deboer-tim Updated! I had the columns set incorrectly and `ml-3` added in the wrong place. It now shows the full hover.

https://github.com/containers/podman-desktop/assets/6422176/e045cfff-2f80-4c5e-8313-459dcd1a8ac7

deboer-tim (approved):
LGTM, worked well, and scaled correctly.

lstocchi (approved):
LGTM, thanks Charlie!! 🚀

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -0,0 +1,75 @@
...
+
+onMount(async () => {
+  // Retrieve both the contexts and clusters
+  let k8sContexts = await window.kubernetesGetContexts();
+  let k8sClusters = await window.kubernetesGetClusters();
+
+  // Convert them to KubeContextUI so we can safely render them.
+  contexts = getKubeUIContexts(k8sContexts, k8sClusters);
+});
+</script>
```
benoitf, line 22: From what I see we collect only the contexts/clusters when we display the page

so if I create a cluster, delete, etc while being on the page it won't be refreshed.

I think we need to use the store pattern and events to correctly have the UI being refreshed from changes and not from when we display the UI

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -0,0 +1,75 @@
...
+
+<SettingsPage title="Kubernetes Contexts">
+  <div class="h-full" role="table" aria-label="contexts">
+    <!-- Use KubernetesIcon in the future / not EngineIcon -->
+    <EmptyScreen
+      aria-label="no-resource-panel"
+      icon="{EngineIcon}"
+      title="No Kubernetes contexts found"
+      message="Check that ~/.local/config exists or KUBECONFIG environment variable has been set correctly."
```
benoitf: IMHO it's `$HOME/.kube/config` not `$HOME/.local/config`

We also have a property in Podman Desktop so probably ask user to check the value of this property as well (KUBECONFIG won't work AFAIK)

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
cdrage: Yes, we will be adding this as part of the second phase since we need to add actions such as delete / create. All of it which would be in a store so it's updating the UI in real time: https://github.com/containers/podman-desktop/issues/4416

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
cdrage: Updated! Good catch, it's suppose to be .kube/ not .local/

cdrage:
Pushed the changes to the text output. For the responsive / store, we'll implement that when we add the create / delete functionality as we will need to have that within a store / responsive to changes.

benoitf (approved):
approving as @cdrage said he'll do the store implementation in the next step

PR #4572, opened 2023-10-31
chore: add telemetry to compose onboarding

### LINKED ISSUE #4571: Add Compose onboarding telemetry

Lack of telemetry for onboarding process at the moment.

Code added for telemetry.

_No response_

### Additional context

_No response_

cdrage:
chore: add telemetry to compose onboardin

### What does this PR do?

Adds telemetry to the functions of the compose onboarding.

N/A

Closes https://github.com/containers/podman-desktop/issues/4571

### How to test this PR?

Should run as normal

### DIFF
--- a/extensions/compose/src/extension.ts
+++ b/extensions/compose/src/extension.ts
@@ -27,7 +27,16 @@ import { ComposeDownload } from './download';
 
 let composeVersionMetadata: ComposeGithubReleaseArtifactMetadata | undefined;
 
+// Telemetry
+let telemetryLogger: extensionApi.TelemetryLogger | undefined;
+
+export function initTelemetryLogger(): void {
+  telemetryLogger = extensionApi.env.createTelemetryLogger();
+}
+
 export async function activate(extensionContext: extensionApi.ExtensionContext): Promise<void> {
+  initTelemetryLogger();
+
   // Check docker-compose binary has been downloaded and update both
   // the configuration setting and the context accordingly
   await handler.updateConfigAndContextComposeBinary(extensionContext);
@@ -73,6 +82,12 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
           extensionApi.context.setValue('composeDownloadVersion', composeVersionMetadata.tag, 'onboarding');
         }
       }
+
+      // Log if it's downloaded and what version is being selected for download (can be either latest, or chosen by user)
+      telemetryLogger.logUsage('compose.onboarding.checkComposeDownloaded', {
+        downloaded: isDownloaded === '' ? false : true,
+        version: composeVersionMetadata?.tag,
+      });
     },
   );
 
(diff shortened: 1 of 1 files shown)


cdrage:
First time adding telemetry so may need a more thorough review @deboer-tim @slemeur

extensions/compose/src/extension.ts
```diff
@@ -73,6 +82,12 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
           extensionApi.context.setValue('composeDownloadVersion', composeVersionMetadata.tag, 'onboarding');
         }
       }
+
+      // Log if it's downloaded and what version is being selected for download (can be either latest, or chosen by user)
+      telemetryLogger.logUsage('compose.onboarding.checkComposeInstalled', {
```
lstocchi: AFAIK we should use the name of the command `compose.onboarding.checkComposeDownloaded`

deboer-tim (commented):
Ideally for something like installComposeBinary() we would log the error or why it failed as a property of the event. Otherwise, we might see that 90% of people are successful, but not know if the 10% failure was due to a bug we can fix, a usability issue, or something out of our control.

Ditto for download() - but doesn't look like the library gives us any info?

This PR would give us separate events for when someone picked a version, what version we tried to download and if is succeeded, and whether they tried to install system wide and if it succeeded. Although it's possible to correlate events in a flow to find out (e.g.) what % of people successfully installed compose, this is a bit harder to do and I don't think we particularly need separate events.

All we really want to know is what choices they made (did they exit immediately after seeing compose? try to download? install system wide?), and whether it succeeded or where it failed along they way. Is it possible to always send a single event after, with optional attributes based on what they tried to do? e.g. something like:

```
  version:
  download:
```

lstocchi (commented):
LGTM, just a comment about the naming based on my previous experience with telemetry ... and I agree with Tim about errors

cdrage:
> Ideally for something like installComposeBinary() we would log the error or why it failed as a property of the event. Otherwise, we might see that 90% of people are successful, but not know if the 10% failure was due to a bug we can fix, a usability issue, or something out of our control.
> 
> Ditto for download() - but doesn't look like the library gives us any info?
> 
> This PR would give us separate events for when someone picked a version, what version we tried to download and if is succeeded, and whether they tried to install system wide and if it succeeded. Although it's possible to correlate events in a flow to find out (e.g.) what % of people successfully installed compose, this is a bit harder to do and I don't think we particularly need separate events.
> 
> All we really want to know is what choices they made (did they exit immediately after seeing compose? try to download? install system wide?), and whether it succeeded or where it failed along they way. Is it possible to always send a single event after, with optional attributes based on what they tried to do? e.g. something like:
> 
> ```
> .logUsage('compose.onboarding.checkComposeInstalled', {
>   version:
>   version-default: (true/false if they modified the default value)
>   download:
>   download-error: (error object, or -success true/false if we can't get that)
>   system-wide: (only added if they got to this part)
>   system-wide-error: (only added if they got to this part. error object, or true/false if we can't get that)
> ```

I do agree for all of the above and we should do this as well for the Podman extension: https://github.com/cdrage/podman-desktop/blob/aaa75128266787f1cf710df4e8799ca93b6f425c/extensions/podman/src/extension.ts#L1047

I modelled this telemetry based on what we already do within `podman/extension.ts` 

We should refactor this after this release a better way to push telemetry that gives an entire summary of the onboarding process (version check error, installation check error, system wide error) rather than how we do it right now with regards to within `registerCommand`). 

I've opened up issue: https://github.com/containers/podman-desktop/issues/4590

cdrage:
Updated the incorrect name I used @lstocchi ready for another review.

deboer-tim (approved):
Given the chat discussion and issue opened for future work, I am approving.

PR #4577, opened 2023-10-31
nit: remove padding from link

cdrage:
nit: remove padding from link

### What does this PR do?

There is a little bit of padding in the link which makes it difficult to
implement columns / tables with lining up the imported link to other
elements.

It also fixes the very subtle spacing in sentences where link is being
used.

https://github.com/containers/podman-desktop/assets/6422176/be2594aa-221e-4d9c-bfa8-8016cbd4b271

Found it while implementing: https://github.com/containers/podman-desktop/pull/4560

### How to test this PR?

Go to "Deploy To Kube" page and see the link (it should be more inline
with the text)

### DIFF
--- a/packages/renderer/src/lib/ui/Link.svelte
+++ b/packages/renderer/src/lib/ui/Link.svelte
@@ -35,7 +35,7 @@ function click() {
 <!-- svelte-ignore a11y-no-redundant-roles -->
 <!-- svelte-ignore a11y-interactive-supports-focus -->
 <a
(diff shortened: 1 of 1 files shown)


deboer-tim:
The padding is intentional, so that the rounded hover extends on either side of the text. Without padding there will be cases where the text aligns with the edge or a sharp corner ('W'?) might even go outside of the hover.

Not saying there aren't other ways we might fix this, but it is intentional and done with Mo.

Issue #4521, opened 2023-10-25
Group pods by environment


deboer-tim:
### Is your feature request related to a problem? Please describe

Issue #4401 will move the current environment 'chiclet' into its own column. We could sort by environment, but this still leaves them in a list where it can be hard to tell where the differences are.

### Describe the solution you'd like

Instead of sorting, it would be nice to see pods grouped by environment. This will provide some visual separation, and can be used to provide more identification/information for each grouping.

mairin:
Just for reference this is the grouping mockups penpot page:
https://design.penpot.app/#/view/18f3baa5-c6db-81d1-8002-9f756ec487eb?page-id=44a14610-3a47-80a0-8003-45811485cb0e&section=interactions&index=0&share-id=bbc09113-e424-8083-8003-5f4111fc4004

deboer-tim:
Still valid.

PR #4583, opened 2023-11-01
feat: add environment column to pods / containers

### LINKED ISSUE #4401: KubeUI: Add environment column to containers and pods lists

The environment chiclets in the pods list are not clear enough and we have feedback that a column should be used instead.

Create a new environment column in the pods list.

_No response_

### Additional context

_No response_

### LINKED ISSUE #4579: Make (pod) or (compose) grouping better in Container view

Unsure where to put it yet, but the whole (compose) and (pod) in the screen doesn't look great, where should we put it instead.

Screenshot 2023-10-31 at 9 32 01 PM

Perhaps put it in environment?

_No response_

### Additional context

_No response_

cdrage:
feat: add environment column to pods / containers

### What does this PR do?

* Adds environment column for pods / containers
* Changes the engine from k8s to Kubernetes to match Podman and Docker
  engine type names
* Reworks the columns a bit
* Moves (Pod) and (Compose) to Environment when displaying a group

Screenshot 2023-10-31 at 10 38 06 PM
Screenshot 2023-10-31 at 10 38 11 PM

Closes https://github.com/containers/podman-desktop/issues/4401
Closes https://github.com/containers/podman-desktop/issues/4579

### How to test this PR?

View Pods / Containers section with multiple engines

### DIFF
--- a/packages/renderer/src/lib/ui/ProviderInfo.svelte
+++ b/packages/renderer/src/lib/ui/ProviderInfo.svelte
@@ -1,0 +1,39 @@
+<script lang="ts">
+import Tooltip from './Tooltip.svelte';
+
+// Name of the provider (e.g. podman, docker, kubernetes)
+export let provider = '';
+
+// Only used for Kubernetes-like distros
+export let context = '';
+
+// Each provider has a colour associated to it within tailwind, this is a map of those colours.
+// bg-purple-600 = podman
+// bg-sky-300 = docker
+// bg-sky-600 = kubernetes
+// bg-gray-900 = unknown
+function getProviderColour(providerName: string): string {
+  switch (providerName) {
+    case 'Podman':
+      return 'bg-purple-600';
+    case 'Docker':
+      return 'bg-sky-400';
+    case 'Kubernetes':
+      return 'bg-sky-600';
+    default:
+      return 'bg-gray-900';
+  }
+}
+</script>
+
+<div class="flex items-center bg-charcoal-500 p-1 rounded-md">
+  <div class="w-2 h-2 {getProviderColour(provider)} rounded-full mr-1"></div>
+  <span class="text-xs capitalize">
+    <!-- If Kubernetes, show the context via the tooltip / hover, else just provider the name.-->
+    {#if provider === 'Kubernetes'}
+      <Tooltip tip="{context}" top>{provider}</Tooltip>
+    {:else}
+      {provider}
+    {/if}
+  </span>
+</div>
--- a/packages/renderer/src/lib/ui/ProviderInfo.svelte
+++ b/packages/renderer/src/lib/ui/ProviderInfo.svelte
@@ -1,5 +1,7 @@
 <script lang="ts">
 import Tooltip from './Tooltip.svelte';
+import { ContainerGroupInfoTypeUI } from '../container/ContainerInfoUI';
+import { PodGroupInfoTypeUI } from '../pod/PodInfoUI';
 
 // Name of the provider (e.g. podman, docker, kubernetes)
 export let provider = '';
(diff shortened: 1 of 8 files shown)


cdrage:
Ready for review!

I have UI questions though @deboer-tim @benoitf @mairin 
* Should we hide the column if only one container engine is active? 
* Should (Pod) and (Compose) live in the environment column? If so, what colour / how should we show it? Makes IMO sense to move it to that column section. 

This was encountered when implementing since (pod) being displayed by the name looked very out-of-place with the Environment column there.

cdrage:
We should prioritize working on this after: https://github.com/containers/podman-desktop/issues/3865 so we can reduce the clutter from the container image name length as adding a new column (Environment) adds more text to the row horizontally.

lstocchi (approved):
LGTM

jeffmaury (approved):
LGTM

slemeur:
This is a great improvement! great work @cdrage and @mairin !

packages/renderer/src/lib/ui/ProviderInfo.svelte
```diff
@@ -0,0 +1,39 @@
...
+    case 'Podman':
+      return 'bg-purple-600';
+    case 'Docker':
+      return 'bg-sky-400';
+    case 'Kubernetes':
+      return 'bg-sky-600';
+    default:
+      return 'bg-gray-900';
+  }
+}
```
benoitf: We can't hardcode values based on the name of the engine. It's unknown from the core part

you could use the engine type docker vs podman but not the name

benoitf (dismissed):

packages/renderer/src/lib/ui/ProviderInfo.svelte
cdrage: Sure, I'll make that change. For container side, it will have to be engineType and use that instead of engineName.

But for pods side, we'll have to use kind: https://github.com/cdrage/podman-desktop/blob/b59dfc2a26156b053ace35475e351a110274a015/packages/renderer/src/lib/pod/PodInfoUI.ts#L38-L39

benoitf: yes relying on kind for k8s and engineType for container side will be more generic 🎉

deboer-tim:
> I have UI questions though @deboer-tim @benoitf @mairin

Sorry, I must have missed the notification.

> * Should we hide the column if only one container engine is active?

I'm going to say no for now. I'm sure there will be some users with only a single environment that find it annoying, but grouping is coming (e.g. #4521) which I'm assuming would also remove the column it's grouping by. We've also talked about customizing the columns in the future. So, I'd rather see if other features address this vs what would have to be very custom code now.

> * Should (Pod) and (Compose) live in the environment column? If so, what colour / how should we show it? Makes IMO sense to move it to that column section.
> 
> This was encountered when implementing since (pod) being displayed by the name looked very out-of-place with the Environment column there.

I was confused for a second when I looked at the screenshot - e.g. to me it looks like the pod is running on some environment called 'pod'. I would expect the pod environment to be the same as all of its children.

Personally I'd rather defer whether the container group type moves columns to a separate PR, but I would at least say it should not use the same UI representation as the environment.

cdrage:
> > I have UI questions though @deboer-tim @benoitf @mairin
> 
> Sorry, I must have missed the notification.
> 
> > * Should we hide the column if only one container engine is active?
> 
> I'm going to say no for now. I'm sure there will be some users with only a single environment that find it annoying, but grouping is coming (e.g. #4521) which I'm assuming would also remove the column it's grouping by. We've also talked about customizing the columns in the future. So, I'd rather see if other features address this vs what would have to be very custom code now.
> 
> > * Should (Pod) and (Compose) live in the environment column? If so, what colour / how should we show it? Makes IMO sense to move it to that column section.
> > 
> > This was encountered when implementing since (pod) being displayed by the name looked very out-of-place with the Environment column there.
> 
> I was confused for a second when I looked at the screenshot - e.g. to me it looks like the pod is running on some environment called 'pod'. I would expect the pod environment to be the same as all of its children.
> 
> Personally I'd rather defer whether the container group type moves columns to a separate PR, but I would at least say it should not use the same UI representation as the environment.

Thank you for the feedback! We'll continue to have (Pod) and (Compose) be shown as part of the group / not in the environment variable and take this up in: https://github.com/containers/podman-desktop/issues/4579

cdrage:
Pushed the changes! Updated it back to the original area for (Pods) and (Compose), and included the changes requested by @benoitf regarding using type /kind instead of engineName

benoitf:
looks like it's not compiling

cdrage:
> looks like it's not compiling

Fixed! Ready for another review.

benoitf (approved):

PR #4600, opened 2023-11-02
feat: add tabs to containers list page

### LINKED ISSUE #707: "Only show running containers" filter for containers

feloy:

Replaces #4483 

### What does this PR do?

This PR adds tabs to the Containers List page, to filter on Running / Stopped containers.

- When the user moves to a Running / Stopped tab, the filter "is:running" / "is:stopped" is added to the filter field.

- When the user adds / removes the "is:running" / "is:stopped" text from the filter field, the UI changes the tab accordingly.

- The filter field content is preserved when the user switches from one tab to another (the filters is:running / is:stopped being modified, the rest of the filter being preserved).

- Groups of containers:
  - After filtering containers, only groups with remaining containers are displayed
  - The number of all containers (before filtering) for a group is displayed, and if some containers are filtered, the number of filtered containers is also displayed. For example `2 containers (1 filtered)` indicates that the group contains 2 containers, but only 1 is displayed because the other one is filtered.

- The empty pages are adapted depending on the tab.

- On an empty page with a filter, the button "Clear filter" removes the search term from the searchbar, but preserves the is:running / is:stopped

#### TODO
- [x] unit tests

Fixes #707 

### How to test this PR?

Have standalone containers and grouped containers (by pod, by compose), stop some of the containers, and switch from tab to tab.

### Demo

https://youtu.be/LRGmCsDBq0M


packages/renderer/src/lib/ui/TabButton.svelte
```diff
@@ -0,0 +1,13 @@
...
+</script>
+
+<button
+  on:click
+  class="pb-2 border-b-[3px] border-charcoal-700 whitespace-nowrap hover:cursor-pointer px-4 py-2 text-gray-600 no-underline"
+  class:border-purple-500="{selected}"
+  class:hover:border-charcoal-100="{!selected}"
+  class:text-white="{selected}"
+  >{title}
+</button>
```
benoitf: hello, I have a question

should it not be based on Podman Desktop `<Button>` or add selected flag on `<Button>` ?

feloy: I wanted to keep a Tab graphical aspect. If we use a `<Button>`, we will get a Button aspect?

benoitf: so I'm wondering if it's not another type of button as we have for example 'Link' type that displays it in a different manner

https://github.com/containers/podman-desktop/blob/71bad3bef2bc53d19792c70c3957e9f1436259e5/packages/renderer/src/lib/ui/Button.ts#L25

so if I say `<Button type="tab">` it'll be displayed as a Tab button ?

It's a suggestion, just to make sure that usage of `<button>` would be only inside Podman Desktop `<Button>` component

I'll see what others think about it

deboer-tim: TBH I like the GitHub Open/Closed equivalent where these would just be link Buttons, but I suppose we should follow the design. :)
We already have the Tab component, it's just hardwired to links. I guess ideally I agree there would be a Button type tab and then Tab would just use it and add link but I haven't looked at the code to see how feasible that would be.

feloy: hello @benoitf , any update?

benoitf: @feloy I think `<Button type="tab"` looks like the straightforward way

feloy: I wonder if we want to add some assertion on the `selected` param, to be used only for type='tab', of if we may want to use for other types?

benoitf: I think it may be ok to keep it simple like a new parameter but only used by tab mode

benoitf (approved):
image

works fine, good job @feloy

deboer-tim (approved):
LGTM. Thanks for all the changes @feloy, this is looking really nice now!

PR #4613, opened 2023-11-02
fix: shorter doc nav section titles

deboer-tim:
### What does this PR do?

A few of the section titles in the doc nav are long, which makes it harder to scan for a topic. Worse, some of these wrap to two lines making the layout not flow very well.

Before:

Screenshot 2023-11-02 at 8 52 18 AM

After:

Screenshot 2023-11-02 at 8 56 01 AM

N/A

### How to test this PR?

`yarn website:dev`

### DIFF
--- a/website/docs/compose/index.md
+++ b/website/docs/compose/index.md
@@ -1,6 +1,6 @@
 ---
 sidebar_position: 35
-title: Working with Compose
+title: Compose
 description: With Podman Desktop, you can install a Compose engine and manage multi-container applications defined in Compose files.
 keywords: [compose]
 tags: [compose]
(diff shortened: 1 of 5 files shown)


slemeur (approved):
+1

benoitf:
I like short names but now I have a hard time to figure out the sorting used in the sidebar

I don't know if we can have some separators or something else but let say I want to look quickly as it's not sorted in alphabetical order or it's not in some main sections I still think navigation is harder

benoitf (approved):

feloy (approved):

deboer-tim:
+1 to @benoitf comments. Separate PR ;-), but I think some subtle dividers or something to break up the list/show sections would be helpful.

Issue #4395, opened 2023-10-18
Dashboard Revamp - Increase awareness of Podman Desktop new capabilities


slemeur:
## Epic Description

The dashboard is the critical place for providing the user a great first impression. After getting through the onboarding process of configuring the different tools needed by Podman Desktop - the user will land onto the dashboard → that’s our opportunity to help the user in the getting started experience with the tool and the technology.

It’s also the place users will see after restarting the tool, or upgrading to a newer version. That’s another opportunity to call out the new features and to entice users to give them a try.

Finally, as user, it’s also the place where I’m expecting to get a glance of my system, so I can quickly free resources, stop intensive operations and tasks. 

### Paris F2F

The "Getting started" button on the dashboard is static even after learning is complete. Nichole noted that dashboard learning never ends.

### Describe the solution you'd like

In this epic, we want to focus on the awareness of the new capabilities provided by Podman Desktop: new releases, new featured extensions. 

We will want as part of this epic to introduce a new section to display the latest release note of Podman Desktop. We would also like to reorder elements so that more attention is driven to the featured extensions.

nichjones1:
Sprint 247 focus will be on reducing the size of cards.  Will make the UI a bit smaller

mairin:
updated mockup (as discussed in https://github.com/containers/podman-desktop/pull/5013#issuecomment-1837001821 )

![287415511-620aaf37-f847-4723-a60d-3c5185a25a30](https://github.com/containers/podman-desktop/assets/799683/36837c26-2156-4060-bfa5-50c8a0ca7683)

Goals with new design:
- have a link to settings > resources (the "manage" button) - helps solve issues identified in https://github.com/containers/podman-desktop/issues/4283
- puts the resource links down a level in the hierarchy - i'm guessing they aren't clicked on too often - under a kebab menu

ekidneyrh:
![Screenshot from 2024-03-25 20-52-55](https://github.com/containers/podman-desktop/assets/87311656/8894af0c-c86a-496f-b536-054b5bf73891)

Added this change to the current mock-up of the dashboard. Just wondering if the 'container engines & orchestrators' is a different section than the providers section, or are they the same thing? 

TIA

deboer-tim:
Same.

ekidneyrh:
Had a chat with Florent - looks like that sections is to be named 'Providers'. 

[...]

PR #4615, opened 2023-11-02
fix: reduce size of provider cards on the dashboard

deboer-tim:
### What does this PR do?

The current provider cards on the dashboard are massive, and it restricts us from adding other things to the dashboard, e.g. notifications or tutorials. There is a longer term redesign of these cards in the works, but in the meantime we can easily start by reducing their size.

Made the following changes:
- Reduce logo size from max-h-24 to max-h-12.
- mt-10 is excessive, reduce all to margins to mt-5.
- provider.updateInfo?.version can be undefined when there is no update available, so adding a check removes the section.
- ProviderWarnings section was always there. Change it to be like the other sections and only appear when there is content.
- Remove extraneous <div> in ProviderInstalled.

Before:

Screenshot 2023-11-02 at 12 26 24 PM

After:

Screenshot 2023-11-02 at 12 21 59 PM

Fixes part of #4395.

### How to test this PR?

Install several providers, go to the dashboard and look for visual anomalies.

### DIFF
--- a/packages/renderer/src/lib/dashboard/ProviderConfigured.svelte
+++ b/packages/renderer/src/lib/dashboard/ProviderConfigured.svelte
@@ -89,12 +89,12 @@ onMount(() => {
       <ErrorMessage class="flex flex-col mt-2 my-2 text-sm" error="{runError}" />
     {/if}
   </div>
-  {#if provider.version !== provider.updateInfo?.version}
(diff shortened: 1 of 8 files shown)


cdrage (approved):
Looks way better with it being small, didn't notice any major visual changes / anything the images affects even scaling it worked well.

benoitf (approved):

Issue #3907, opened 2023-09-14
Container created with an error can be supossedly started and is shown running in Containers page


odockal:
### Bug description

there are multiple wrong information when starting container (that cannot be stared due to upstream bug) from Containers Page. See the attachment.

![rootless-container-error-containers-page](https://github.com/containers/podman-desktop/assets/19164299/e1fd262c-c0cd-4e5c-ac4a-2ef7b8d6ca54)

### Operating system

Windows 10

### Installation Method

Installer from website/GitHub releases

### Version

1.4.0

### Steps to reproduce

Follow up of #3899.
1. rootless podman machine
2. pull image alpine
3. start container (expected state is Created because of error)
4. Start container on Containers page -
5. check its status should get back to Created, shows an error, and shows right AGE.

alexojegu:
Same here on Arch Linux with rootless podman and Podman Desktop installed from Flatpak.

Can be related with the update of `conmon` or `containers-common` or maybe `slirp4netns`?
> upgraded conmon (1:2.1.7-1 -> 1:2.1.8-1)
> upgraded containers-common (1:0.55.2-1 -> 1:0.55.3-1)
> upgraded slirp4netns (1.2.0-1 -> 1.2.1-1)

alexojegu:
For me it was fixed after the last update of my system:

> podman (4.6.2-1 -> 4.7.1-1)
> containers-common (1:0.55.4-1 -> 1:0.56.0-1)
> netavark (1.7.0-1 -> 1.8.0-1)
> catatonit (0.1.7-2 -> 0.2.0-1)

lstocchi:
I'm still able to replicate it but i wonder about the correct behavior.
It also happen if you try to create 2 containers that use the same port. The first one succeeds, the second is created but not started.

So the question is, if we create the container but it does not get started, what should we do? We redirect the user to the containers page anyway? We do not redirect and we show an error "the container has been created but not started"? Anything else? @benoitf

benoitf:
@lstocchi I would say there are two issues if we take one approach or the second one

In case on some error, let say the port number is incorrect, I would like to be kept in the creation form because I can quickly change the port number

If I go directly to the details of a container because it's created but not started then it's nice as I can go the logs page and see error, delete the container but I can't recreate it with different value (like the port number)

for the specific use case of the port number, maybe we shouldn't let people click on 'create' as we know it won't be available
like 'fail fast' pattern

I don't know what others think

PR #4616, opened 2023-11-02
fix: find a free port

lstocchi:
### What does this PR do?

I found out that, atleast on my machine, the `getFreePort` function was not actually returning a free port, so when trying to create/start a container the form was always pre-filled with port 9000 even if it was busy.

I saw that we didn't add any hostname to specify to the "fake" net.server where to listen to and that was my problem. By adding the localhost ip, it worked as expected.

Now the question is, there was any reason why we never added a hostname? I was expecting that localhost was the default but that is not true apparently. 
Without specifying the hostname it tried to connect to
`6::::9000`
by adding the localhost
`4:127.0.0.1:9001`

get_free_port

This PR is part of #3907

### How to test this PR?

1. create and start a container
2. repeat to see if the port number has been updated

### DIFF
--- a/packages/main/src/plugin/util/port.ts
+++ b/packages/main/src/plugin/util/port.ts
@@ -21,19 +21,19 @@ import * as net from 'net';
 /**
  * Find a free port starting from the given port
  */
-export function getFreePort(port = 0): Promise<number> {
+export async function getFreePort(port = 0): Promise<number> {
   if (port < 1024) {
     port = 9000;
   }
-  const server = net.createServer();
-  return new Promise((resolve, reject) =>
-    server
-      .on('error', (error: NodeJS.ErrnoException) =>
-        error.code === 'EADDRINUSE' ? server.listen(++port) : reject(error),
-      )
-      .on('listening', () => server.close(() => resolve(port)))
-      .listen(port),
-  );
+  let isFree = false;
+  while (!isFree) {
+    isFree = await isFreePort(port);
+    if (!isFree) {
+      port++;
+    }
+  }
+
+  return port;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


benoitf (approved):

packages/main/src/plugin/util/port.ts
```diff
@@ -21,19 +21,19 @@ import * as net from 'net';
...
-  return new Promise((resolve, reject) =>
-    server
-      .on('error', (error: NodeJS.ErrnoException) =>
-        error.code === 'EADDRINUSE' ? server.listen(++port) : reject(error),
-      )
-      .on('listening', () => server.close(() => resolve(port)))
-      .listen(port),
-  );
+  let isFree = false;
+  while (!isFree) {
```
feloy, line 29: I would add an upper limit, in case detection fails every time for some system failure, to avoid looping infinitely

lstocchi: What do you mean by upper limit? 65353? or like try 100 times and stop?

feloy: A port upper limit, like 65353

lstocchi: In that case it already fails bc when it goes over the upper range, the promise in `isFreePort` rejects. 
So it will not run infinitely. It just resolve to false if the error code is `EADDRINUSE`

lstocchi: I can return the starting port if that happens. Updating

lstocchi: Thinking again, the function should return a free port so if that's not possible it makes sense it returns an error

lstocchi: added a couple of tests to verify that it fails with an out of range port

feloy (approved):

PR #4630, opened 2023-11-03
feat: command palette: add enablement property

### LINKED ISSUE #4107: command palette: add enablement property

Add support for enablement in contributed commands

an extension could decide to disable a command from the command palette using the enablement clause.

```json
      {
        "command": "my.command",
        "title": "My Command",
        "category": "Podman",
        "enablement": "!operationInProgress"
      },
```

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
as reported in comments https://github.com/containers/podman-desktop/pull/4081#pullrequestreview-1642374073
all commands should not appear in the command palette

N/A

fixes #4107

### How to test this PR?

Unit test added

can add enablement in package.json and see if it's disabled based on the property

### DIFF
--- a/EXTENSIONS.md
+++ b/EXTENSIONS.md
@@ -93,6 +93,37 @@ Upon a successful start up via the `activate` function within your extension, `P
 - `extension-loader.ts`: Attempts to load the extension and sets the status accordingly (either `started`, `stopped`, `starting` or `stopping`). If an unknown error has occurred, the status is set to `unknown`. `extension-loader.ts` also sends an API call to Podman Desktop to update the UI of the extension.
 - `tray-menu.ts`: If `extensionApi.tray.registerMenuItem(item);` API call has been used, a tray menu of the extension will be created. When created, Podman Desktop will use the `ProviderConnectionStatus` to indicate the status within the tray menu.
 
+#### Commands
+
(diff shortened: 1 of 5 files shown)


deboer-tim (changes requested):
I think there are going to be several extenders who don't notice this flag and unknowingly add their commands to the palette when they didn't want to... but I agree it's the right choice, as is making it optional. Code looks good. 👍🏼

Two concerns though:
- This is basically a way for extenders to say 'no, don't make my command accessible from elsewhere'. Is `enablement` the right name? Wondering if it should be `visible` or `externally-accessible` (ok, that is horrible), or maybe even something like `hidden`.
- There's no way for developers to understand what this property does without finding this PR or following the code. Could you add a comment in the API or something short in the documentation for writing extensions?

benoitf:
about your first concern (the enablement name), I followed VS Code API

deboer-tim:
I thought that might be the reason. :) Could you just add something short to the API or docs so developers can tell what it does?

benoitf:
@deboer-tim documentation updated

lstocchi (approved):
LGTM

deboer-tim (approved):

PR #4637, opened 2023-11-03
fix: disable create/start container if any port is busy

### LINKED ISSUE #3907: Container created with an error can be supossedly started and is shown running in Containers page

lstocchi:
### What does this PR do?

This PR adds a check everytime a port is added/updated in the create/start container page so the button gets disabled if one of the ports is busy

show_used_ports

it resolves #3907 

### How to test this PR?

1. go to the create a container from an image page
2. add a port that is already in use
3. you should not be able to start the container

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -52,7 +52,7 @@ import type { PullEvent } from './api/pull-event.js';
 import type { ExtensionInfo } from './api/extension-info.js';
 import type { ImageInspectInfo } from './api/image-inspect-info.js';
 import type { TrayMenu } from '../tray-menu.js';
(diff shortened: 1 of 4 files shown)


benoitf (approved):
I thought I already review it but it seems not

PR #4646, opened 2023-11-03
feat: add status dots for pods

### LINKED ISSUE #3925: Show containers as little green/yellow/orange dots in Pods list.

No indication what containers are not working in the Pods list in the UI. 

As part of devtools week, a UI design was implemented to show containers as little green dots as part of the Pod.

This gives good indication of which ones are starting up / not working / working sucessfully.

Screenshot 2023-09-15 at 12 04 20 PM

_No response_

### Additional context

There is some PoC code implemented in: https://github.com/containers/podman-desktop/compare/main...cdrage:podman-desktop:kubernetes-dashboard

_No response_

### LINKED ISSUE #4612: Clicking on "2 containers" / "1 container" in Pods screen on a Kubernetes Pod is broken
### Bug description

https://github.com/containers/podman-desktop/assets/6422176/b4405257-0a17-4b61-8f52-ea1169dd7e37

When clicking on "1 container" / "2 containers" on a k8s pod, it will take you to the Containers list.. where the containers wouldn't be listed because this is k8s related.

### Operating system

mac M1

### Installation Method

None

### Version

1.5.0

### Steps to reproduce

1. Deploy a Kubernetes Pod
2. Click on the "1 container" text.

### Relevant log output

_No response_

### Additional context

Will be fixed when I've implemented #3925

cdrage:
feat: add status dots for pods

### What does this PR do?

* Adds dots for the containers when listing a pod (podman or kubernetes)
* Hover to see the tooltip of the container name as well as the running
  status
* Dots will change when the container status has changed.
* Fixes a small redirection error bug when clicking a kubernetes pod

https://github.com/containers/podman-desktop/assets/6422176/7fbee0ef-9a2a-42e7-9087-7ab08c29b425

Closes https://github.com/containers/podman-desktop/issues/3925
Closes https://github.com/containers/podman-desktop/issues/4612

### DIFF
--- a/packages/renderer/src/lib/ui/StatusDot.svelte
+++ b/packages/renderer/src/lib/ui/StatusDot.svelte
@@ -1,35 +1,30 @@
 <!-- StatusDot.svelte -->
 <script lang="ts">
-import type { PodInfoContainerUI } from '../pod/PodInfoUI';
 import Tooltip from './Tooltip.svelte';
+import { getStatusColor } from './Dots';
+import { capitalize } from './Util';
 
-export let container: PodInfoContainerUI;
+export let status: string;
+export let name: string = '';
+export let tooltip: string = '';
+export let number: number = 0;
 
-// All the possible statuses that will appear for both Pods and Kubernetes
-const colors: Record<string, string> = {
-  // Podman & Kubernetes
-  running: 'bg-green-500', // Green for running
-
-  // Kubernetes-only
-  terminated: 'bg-red-500', // Red for terminated
-  waiting: 'bg-amber-500', // Amber for waiting
-
-  // Podman-only
-  stopped: 'bg-gray-500', // Gray indicated dormancy / inactivity
-  paused: 'bg-amber-500', // Equivilant to "waiting" in Kubernetes
-  exited: 'bg-red-300', // Light red for exited, but dark red for dead
-  dead: 'bg-red-500', // Dark red for dead
-  created: 'bg-green-300', // Blue indicates "freshness", so would be good for 'created'
-  degraded: 'bg-amber-700', // Dark amber indicating "degrading" state / "unhealthy"
-};
-
-// Returns the color for the status and if it cant, it'll use gray
-const getColor = (status: string): string => colors[status] || 'bg-gray-500';
+// If the tooltip is blank, use the container name and status
+// as the tooltip / title
+if (tooltip === '' && name !== '' && status !== '') {
+  tooltip = `${name}: ${capitalize(status)}`;
+}
+// if text is not blank, make class 'mt-3' to keep it center
 </script>
 
-<Tooltip tip="{container.Names}: {container.Status}" top
-  ><div
-    class="w-2 h-2 rounded-full mr-1 text-center {getColor(container.Status)}"
+<Tooltip tip="{tooltip}" top>
+  <div
+    class="w-2.5 h-2.5 mr-0.5 rounded-full text-center {getStatusColor(status)} {number ? 'mt-3' : ''}"
     data-testid="status-dot"
-    title="{container.Names}: {container.Status}">
-  </div></Tooltip>
+    title="{tooltip}">
+  </div>
+  <!-- If text -->
+  {#if number}
+    <div class="text-xs text-bold text-gray-600 mr-0.5">{number}</div>
+  {/if}
+</Tooltip>
--- a/packages/renderer/src/lib/ui/StatusDot.svelte
+++ b/packages/renderer/src/lib/ui/StatusDot.svelte
@@ -1,0 +1,35 @@
+<!-- StatusDot.svelte -->
+<script lang="ts">
+import type { PodInfoContainerUI } from '../pod/PodInfoUI';
+import Tooltip from './Tooltip.svelte';
+
+export let container: PodInfoContainerUI;
+
+// All the possible statuses that will appear for both Pods and Kubernetes
+const colors: Record<string, string> = {
+  // Podman & Kubernetes
+  running: 'bg-green-500', // Green for running
+
+  // Kubernetes-only
+  terminated: 'bg-red-500', // Red for terminated
+  waiting: 'bg-amber-500', // Amber for waiting
+
+  // Podman-only
+  stopped: 'bg-gray-500', // Gray indicated dormancy / inactivity
+  paused: 'bg-amber-500', // Equivilant to "waiting" in Kubernetes
+  exited: 'bg-red-300', // Light red for exited, but dark red for dead
+  dead: 'bg-red-500', // Dark red for dead
+  created: 'bg-green-300', // Blue indicates "freshness", so would be good for 'created'
+  degraded: 'bg-amber-700', // Dark amber indicating "degrading" state / "unhealthy"
+};
+
+// Returns the color for the status and if it cant, it'll use gray
+const getColor = (status: string): string => colors[status] || 'bg-gray-500';
+</script>
+
+<Tooltip tip="{container.Names}: {container.Status}" top
+  ><div
+    class="w-2 h-2 rounded-full mr-1 text-center {getColor(container.Status)}"
+    data-testid="status-dot"
+    title="{container.Names}: {container.Status}">
+  </div></Tooltip>
--- a/packages/renderer/src/lib/ui/StatusDot.svelte
+++ b/packages/renderer/src/lib/ui/StatusDot.svelte
@@ -14,12 +14,19 @@ export let number: number = 0;
 if (tooltip === '' && name !== '' && status !== '') {
   tooltip = `${name}: ${capitalize(status)}`;
 }
-// if text is not blank, make class 'mt-3' to keep it center
+
+// Get the color class for the status
+// that could be either an outline or a fill
+let dotClass = getStatusColor(status);
+
+// If dotClass contains "outline", then we will use 'outline-1 outline-offset-[-2px]
 </script>
 
 <Tooltip tip="{tooltip}" top>
   <div
-    class="w-2.5 h-2.5 mr-0.5 rounded-full text-center {getStatusColor(status)} {number ? 'mt-3' : ''}"
+    class="w-2.5 h-2.5 mr-0.5 rounded-full text-center {dotClass.includes('outline')
+      ? 'outline-2 outline-offset-[-2px] outline'
+      : ''} {getStatusColor(status)} {number ? 'mt-3' : ''}"
     data-testid="status-dot"
     title="{tooltip}">
   </div>
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -59,14 +59,14 @@ import type { Telemetry } from '/@/plugin/telemetry/telemetry.js';
 function toContainerStatus(state: V1ContainerState | undefined): string {
   if (state) {
     if (state.running) {
-      return 'Running';
+      return 'running';
     } else if (state.terminated) {
-      return 'Terminated';
+      return 'terminated';
     } else if (state.waiting) {
-      return 'Waiting';
+      return 'waiting';
     }
   }
-  return 'Unknown';
+  return 'unknown';
 }
 
 function toPodInfo(pod: V1Pod, contextName?: string): PodInfo {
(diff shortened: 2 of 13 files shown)


cdrage:
Ready for review! 💯 

Need some help however with regards to what colours to use as there are 9 different types of statuses that can be shown.

https://github.com/containers/podman-desktop/pull/4646/files#diff-122a4449594e476ae1551ff015e3547ecec1c30d999c77295b5938a329d495f7R17

packages/renderer/src/lib/ui/StatusDot.svelte
```diff
@@ -0,0 +1,35 @@
...
+  // Podman & Kubernetes
+  running: 'bg-green-500', // Green for running
+
+  // Kubernetes-only
+  terminated: 'bg-red-500', // Red for terminated
+  waiting: 'bg-amber-500', // Amber for waiting
+
+  // Podman-only
+  stopped: 'bg-gray-500', // Gray indicated dormancy / inactivity
+  paused: 'bg-amber-500', // Equivilant to "waiting" in Kubernetes
```
deboer-tim: -600. and Equivalent :)

packages/renderer/src/lib/ui/StatusDot.svelte
```diff
@@ -0,0 +1,35 @@
...
+  exited: 'bg-red-300', // Light red for exited, but dark red for dead
+  dead: 'bg-red-500', // Dark red for dead
+  created: 'bg-green-300', // Blue indicates "freshness", so would be good for 'created'
+  degraded: 'bg-amber-700', // Dark amber indicating "degrading" state / "unhealthy"
```
deboer-tim: Should be -600

packages/renderer/src/lib/ui/StatusDot.svelte
```diff
@@ -0,0 +1,35 @@
...
+<script lang="ts">
+import type { PodInfoContainerUI } from '../pod/PodInfoUI';
+import Tooltip from './Tooltip.svelte';
+
+export let container: PodInfoContainerUI;
+
+// All the possible statuses that will appear for both Pods and Kubernetes
+const colors: Record<string, string> = {
+  // Podman & Kubernetes
+  running: 'bg-green-500', // Green for running
```
deboer-tim: The only existing colours we have are in StatusIcon, so we should start be matching those, e.g. -400 here (or arguing that we should change both). It doesn't have all these states though, so picking reasonable ones for things that aren't covered should be ok.

packages/renderer/src/lib/ui/StatusDot.svelte
benoitf: what about adding a `status` color in the palette (tailwind.config.js) with something like (colors codes are pure example) 

```js
'status': {
        'running': tailwindColors.green[500],
        'stopped': tailwindColors.amber[500],
        etc..
      }
```

so when we need to display a running state we do `bg-status-running` and we see values only at one location

benoitf: or an utility class that provides/share the value across StatusDot and StatusIcon

jeffmaury (changes requested):
We should handle large numbers of containers better 

dots

deboer-tim:
lol, I want to know what Jeff is trying to run ;-)

Could wrap to 2 or 3 lines when there are too many, and if it's still too many add something like `+3`.

packages/renderer/src/lib/ui/StatusDot.svelte
deboer-tim: +1 to tailwind cfg, I think we could use these for things like machine status too.

benoitf:
if too many I don't think we should wrap anyway as we'll have rows with different size
if we cut, I don't know if it's possible to group them by status like `3+ 🟢 ` and `4+ 🔴`

cdrage:
As a result of some last minute discussions, I'll be implementing the following in the PR:

- [x] Fix misrepresented colours (some are -700 which should be -400)
- [x] Add colours representing statuses to the tailwind configuration so we can change the colours there rather than in the svelte file
- [x] For more than 10 containers, have them separate by colour dot and a number beside them, example: 🔴 +4, with the tooltip saying: "Terminated: 4".
- [x] The containers should be arranged by terminated / running, for example: running first, then paused / degraded / waiting then terminated / stopped / etc.

cdrage:
Ready for another review / test! @jeffmaury @benoitf @deboer-tim @mairin 

I have implemented all the above suggestions. Below are screenshots / recordings of the new UI.

I have also decided to put the numbers below the dots, it can encompass 2 digit numbers (I really really doubt that a cluster would have over 100 containers in 1 pod). 

Which makes the UI more uniform in terms of looks. Potentially we could have 1 of every status (total of 9), and numbers underneath it would still fit under the '10 container' limit we've imposed.

https://github.com/containers/podman-desktop/assets/6422176/5a64ba5d-329e-4787-9d54-5c42eca00ca5

Screenshot 2023-11-06 at 9 49 01 PM

After a preliminary review, I can add more tests for the new files.

benoitf:
good work @cdrage 

based on your screenshot, I think the number being displayed below the circle shouldn't move the circles
I mean, circles should always be at the same location on the height of the row (with or without numbers below)


and to my personal preference I think numbers are too big (or the circles are too small) and too bright (but you can ignore my preference :-)

mairin:
> Which makes the UI more uniform in terms of looks. Potentially we could have 1 of every status (total of 9), and numbers underneath it would still fit under the '10 container' limit we've imposed.

Hey 👋 That was fast! 😁 I wanna give you my honest opinion Charlie but plz feel free to take or leave it - 

I think the benefit of this approach is you can scale for anything. But a potential issue / downside is that uniformity - just glancing at the list of pods, you no longer get the sense of the comparative "weights" of them... the benefit of the 1:1 icon:container ratio is at a glance you can tell which pod is a biggie and which is small. Now to figure that out, you gotta pause and process the #'s.

I think > 10 containers (and it seems maybe  greater than 3 or 4 really) in a pod is pretty edge case, so you lose the advantage of at l-a-glance relative size to scale to a size you might not ever need to scale to for most users at most times.

The UX approach I take is to optimize for the majority case, so that's why if it were my call I'd say make the relative sizes scannable for < 10... it makes it a little easier for the user to not have to read and thus the overall experience a bit better. But I also don't know for sure and dont have your k8s expertise to know if scanning against relative sizes is important, just a gut feeling.

One way to better enable easier to perceive relative sizing with this approach would be to maybe eventually make the container count a sortable field. You'd probably need to sort by the total # of containers despite the breakdown by type (maybe that would feel weird / awkward to users mechanically but im ot so sure actually.) But then if the relative sizes were important to me, I could sort from most to least containers or vice-versa. 

I hope this helps, it's just what struck me looking this over!

cdrage:
> > Which makes the UI more uniform in terms of looks. Potentially we could have 1 of every status (total of 9), and numbers underneath it would still fit under the '10 container' limit we've imposed.
> 
> Hey 👋 That was fast! 😁 I wanna give you my honest opinion Charlie but plz feel free to take or leave it -
> 
> I think the benefit of this approach is you can scale for anything. But a potential issue / downside is that uniformity - just glancing at the list of pods, you no longer get the sense of the comparative "weights" of them... the benefit of the 1:1 icon:container ratio is at a glance you can tell which pod is a biggie and which is small. Now to figure that out, you gotta pause and process the #'s.
> 
> I think > 10 containers (and it seems maybe greater than 3 or 4 really) in a pod is pretty edge case, so you lose the advantage of at l-a-glance relative size to scale to a size you might not ever need to scale to for most users at most times.
> 
> The UX approach I take is to optimize for the majority case, so that's why if it were my call I'd say make the relative sizes scannable for < 10... it makes it a little easier for the user to not have to read and thus the overall experience a bit better. But I also don't know for sure and dont have your k8s expertise to know if scanning against relative sizes is important, just a gut feeling.
> 
> One way to better enable easier to perceive relative sizing with this approach would be to maybe eventually make the container count a sortable field. You'd probably need to sort by the total # of containers despite the breakdown by type (maybe that would feel weird / awkward to users mechanically but im ot so sure actually.) But then if the relative sizes were important to me, I could sort from most to least containers or vice-versa.
> 
> I hope this helps, it's just what struck me looking this over!

Hmmm we are getting there!

Okay so from @benoitf feedback as well as yours, I propose we:

- Make the circles a tad larger since they are quite hard to read to begin with even with less than 10 containers.
- When >10 keep circles still uniform in the middle of the row, but add numbers on bottom (with the current same font size). The larger circles will make it more pleasing to the eye

To be honest, over 10 containers is like you said going to be more of an "astronautical edge case", but we should obviously still have it there for those who do have more than 10 / something odd happened to their pod cluster.

Let me know if this is good with you both!

cdrage:
Converted to draft as I rebase and make some changes

benoitf:
I'm wondering if instead of adding numbers we should not change the shape like to be ellipsis instead of circles (for multiple items) or picking up another form (I won't block the merge of PR anyway, just sharing thoughts)

feloy:
A non-blocking remark, I'm not sure if it is due to my mild color blindness of if it is a common issue, but I can hardly differentiate between green and red dots when they are not close enough, for example if all dots in one pod are red and all dots in the following pod are green, I hardly differentiate which one is running, which one is stopped

cdrage:
> A non-blocking remark, I'm not sure if it is due to my mild color blindness of if it is a common issue, but I can hardly differentiate between green and red dots when they are not close enough, for example if all dots in one pod are red and all dots in the following pod are green, I hardly differentiate which one is running, which one is stopped

I agree, I added some spacing to fit the numbers in, I'll remove the spacing to make it smaller and more accessible.

cdrage:
Ready for review!

Updates:
- When more than 10 containers, the dots are now centered / does not change the view
- Dots are slightly larger now
- No more spacing between dots / they are closer together
- A lot more tests have been added

deboer-tim (commented):
I really like this, and it's scaling well now. Honestly, the only thing that threw me off was 'why does pink mean stopped'. IMHO stopped/exited should be an outlined circle (no fill) to try to match what we have in the other pages (or the pod itself).

Will need some follow-up to make sure we open issues to use these newly defined colors in other places.

Tests 👍🏼.

cdrage:
> I really like this, and it's scaling well now. Honestly, the only thing that threw me off was 'why does pink mean stopped'. IMHO stopped/exited should be an outlined circle (no fill) to try to match what we have in the other pages (or the pod itself).
> 
> Will need some follow-up to make sure we open issues to use these newly defined colors in other places.
> 
> Tests 👍🏼.

I agree, I will add a second set of dots we can add which instead have "outlines" with no fills (example red outline no fill, grey outline no fill, etc.). Which would very beneficial instead of having them confused with a small pink dot vs red, etc.

The second set of dots will allow us to add a grey dot with an outline that's similar as the current icon's we have

cdrage:
> I really like this, and it's scaling well now. Honestly, the only thing that threw me off was 'why does pink mean stopped'. IMHO stopped/exited should be an outlined circle (no fill) to try to match what we have in the other pages (or the pod itself).
> 
> Will need some follow-up to make sure we open issues to use these newly defined colors in other places.
> 
> Tests 👍🏼.

Ready for another review! Great suggestion with regards to outline, I've now added it. Having outlines opens up a greater set of possibilities for different colours / representing statuses, even being able to possible integrate this into the cluster UI.

Feel free to go over the colors in tailwind.config.cjs to make sure it's the correct ones!

I've realized that we use the same colour / gray outline for stopped and exited, so I've updated that accordingly in this PR. 

Screenshot 2023-11-09 at 10 45 32 AM

deboer-tim (approved):
Outline works much better and looks consistent!

I still see a slight difference in that the exited status icon is currently using gray-700 for the outline (#aaabac) while the dot outline is showing up as #d1d5db, but we may be into the realm of 'color differences that only Tim would notice' so I am approving and I can investigate some time later.

deboer-tim:
(need to fix failing tests though)

cdrage:
> Outline works much better and looks consistent!
> 
> I still see a slight difference in that the exited status icon is currently using gray-700 for the outline (#aaabac) while the dot outline is showing up as #d1d5db, but we may be into the realm of 'color differences that only Tim would notice' so I am approving and I can investigate some time later.

I used -700 instead of -300 by mistake! Fixed it in the config and good to go now!

jeffmaury (approved):

cdrage:
Fixeed the test (was testing against the incorrect class), enabling auto-merge.

mairin:
There's still issues here with relying on color, this is how they look in the mockups:

Screenshot from 2023-11-09 18-49-49

So the orange has an X in it, so you can rely on the shape without the color. The colors from the mockup are:

- green-500 (filled)
- amber-600 (outline x)
- gray-900 (outline)

mairin:
gah i missed this will move to an issue

PR #4662, opened 2023-11-06
feat: image checker extension API

### LINKED ISSUE #4697: Image Checker extension API specification

Images should be checked against specific contraints. The Podman Desktop API should be extension to allow providers to implement such checks and reports results to the UI

An extension to the Podman Desktop API

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

This PR defines and implements the API for Image Checker providers.

An extension registers as an Image checker, by calling the following function, passing a Provider object and optional metadata:

```ts
  registerImageCheckerProvider(
      imageCheckerProvider: ImageCheckerProvider,
      metadata?: ImageCheckerProviderMetadata,
    )
```

The provider object needs to define a `check` method, which will be called by the UI:

```ts
  export interface ImageCheckerProvider {
    check(image: ImageInfo, token?: CancellationToken): ProviderResult<ImageChecks>;
  }
```

The `check` method returns a list of image checks results.

```ts
  export interface ImageCheck {
    name: string;
    status: 'success' | 'failed';
    severity?: 'low' | 'medium' | 'high' | 'critical';
    markdownDescription?: string;
  }

  export interface ImageChecks {
    checks: ImageCheck[];
  }
```

Fixes #4697


packages/extension-api/src/extension-api.d.ts
```diff
@@ -2488,4 +2488,24 @@ declare module '@podman-desktop/api' {
...
+  }
+
+  export interface ImageCheckerProvider {
+    readonly name: string;
+    readonly categories: string[];
+    checkImage(image: string): Promise<ImageCheckResult>;
+  }
+
+  export namespace imageChecker {
+    export function registerImageCheckerProvider(id: string, imageCheckerProvider: ImageCheckerProvider): Disposable;
```
benoitf: Hello, what is the goal of the id ?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2488,4 +2488,24 @@ declare module '@podman-desktop/api' {
...
+
+  export interface ImageCheckResult {
+    status: string;
+    partialResults: ImageCheckPartialResult[];
+  }
+
+  export interface ImageCheckerProvider {
+    readonly name: string;
+    readonly categories: string[];
+    checkImage(image: string): Promise<ImageCheckResult>;
```
benoitf: AFAIK the provider goal is to provide only functions so I don't see name and categories there ?

benoitf: I think checkImage is missing a `cancellationToken: CancellationToken` parameter to cancel the check

packages/extension-api/src/extension-api.d.ts
feloy: To make the provider unique (between extensions, but an extension could also declare several image checker providers). I got inspiration from other providers (`registerAuthenticationProvider`), but it could be created automatically, like `clitool`?

benoitf: got it, I think it's because for Authentication Provider it's one of the few where it needs to provide a public identifier so that other consumers can ask for a 'github' provider

We took inspiration of providers from https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts (you'll see that only few of them provides an identifier, the identifier mostly being the matching part.


Here, as you said, as your goal is for unicity  it needs to be managed internally by the core part.

feloy: ok, I'll change to compute it internally, thanks

packages/extension-api/src/extension-api.d.ts
feloy: The `name` is what the UI can display for this provider, for example "Image OpenShift Analyzer". The `categories` indicates to the UI which are the different categories of tests the checker will run, so the results can be classified in these categories (for example `USER directive`, `RUN Directive`, `Expose directive`)

benoitf: it looks that from design perspective, the static fields are passed on the registration method

example of contract:

```
registerCodeActionsProvider(selector: DocumentSelector, provider: CodeActionProvider, metadata?: CodeActionProviderMetadata): Disposable;
```

From end user perspective, I don't really understand at all the goals of `USER`, `RUN` or `Expose directive`

I mean, I launch a check on the image so if the provider is there it'll do the check, else it won't.

then, if the checker is not able to analyze some images it's more a selector/filter parameter

About the name, it's part of the Result no ?

feloy: design-checker

In this example, the name would be `OpenShift Checker`, the UI needs to know it before to start the tests, to be able to display it in the tab.

For the categories, they would be `Container User` and `Exposed Ports`, so the UI can display them without any failed / passed mark, the time the checks are executed, then they are updated with passed / failed. But I perhaps misunderstood the design doc.

benoitf: I read again draft of the spec

For categories, something like 'Container User' or 'Exposed Ports' are only known by the checker and to me it's part of the result of the scan, not part of the registration.

Until we didn't launch a scan, nothing is displayed except that we're processing.

> In this example, the name would be OpenShift Checker, the UI needs to know it before to start the tests, to be able to display it in the tab.

what we can do for that is to give an extra argument when registering the provider with metadata

```ts
registerImageCheckerProvider(imageCheckerProvider: ImageCheckerProvider, metadata?: ImageCheckerProviderMetadata): Disposable;
```

and in the metadata you can specify a readonly name

without the name, we can assume the provider will take the name of the extension

so if your extension is named 'OpenShift Checker' then it'll use that name as well



Now if I look at the signature of the provider, it should return a ProviderResult<ImageCheck> result (or ImageCheck as the type we want is ImageCheck, result is just from what we call)

then if I look at Trivy extension result
image

I got them classified/sorted by severity


but if I look at

```ts
  export interface ImageCheckPartialResult {
    category: string;
    description: string;
  }

  export interface ImageCheckResult {
    status: string;
    partialResults: ImageCheckPartialResult[];
  }
```

>  (a partial result indicates a failing check on the image).

I don't get the why a partial result would be a failed result. I assume partial result is that I don't have all the result, not that it failed

so looking back at trivy results and openshift checker, it looks like I need a severity (like using a wrong EXPOSE stuff might be less critical than using an image with root user)


I would probably need some markdown result like if it passes the test

image


and then probably a score or grade if it fails.

my image is Grade A, grade B, grade C etc depending on the check

feloy: thanks @benoitf for your feedback. I'll make changes based on this

packages/extension-api/src/extension-api.d.ts
```diff
@@ -136,6 +136,37 @@ declare module '@podman-desktop/api' {
...
+   *    return new ImageChecks();
+   *  }
+   *
+   * let b: ImageCheckerProvider = {
+   *  check(image: ImageInfo, token?: CancellationToken): ProviderResult<ImageChecks> {
+   * 		return new Promise(resolve => {
+   * 			resolve(new ImageChecks());
+   * 	 	});
+   * 	}
+   * }
```
```suggestion
   * let b: ImageCheckerProvider = {
   *  async check(image: ImageInfo, token?: CancellationToken): ProviderResult<ImageChecks> {
   * 		return new ImageChecks();
   * 	}
   * }
```

use async instead of new Promise(...) 

Also we could do `return Promise.resolve()` but I think it's better to use async methods

packages/extension-api/src/extension-api.d.ts
```diff
@@ -136,6 +136,37 @@ declare module '@podman-desktop/api' {
...
+   *
+   * let c: ImageCheckerProvider = {
+   *  check(image: ImageInfo, token?: CancellationToken): ProviderResult<ImageChecks> {
+   * 		return; // undefined
+   * 	}
+   * }
+   * ```
+   */
+  export type ProviderResult<T> = T | undefined | null | Promise<T | undefined | null>;
```
```suggestion
  export type ProviderResult<T> = T | undefined | Promise<T | undefined>;
```

In Podman Desktop we've banned `null`

feloy:
@benoitf how do we want to proceed for integrating this API into `main`? Do I continue to work on the implementation of the API and the UI on this same PR?

benoitf:
@feloy you can use a branch in your forked repository but to move forward, we should have one PR about the API d.ts and its implementation and other PR for UI side.
Also other PRs for the extension itself but it'll be in a separate repository

benoitf (approved):

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1015,6 +1017,17 @@ export class ExtensionLoader {
       },
     };
 
+    const imageCheckerProvider = this.imageCheckerProvider;
+    console.log('imageCheckerProvider', this.imageCheckerProvider);
```
benoitf, line 1021: it should be removed

PR #4674, opened 2023-11-07
feat: add extension/kubectl-cli to register kubectl CLI Tool

### LINKED ISSUE #4118: Provide a way to install `kubectl` CLI

We're providing a lot of CLI like compose, kind, etc

but we should also provide a way to setup `kubectl` binary

Have a nice way to install `kubectl`

_No response_

### Additional context

_No response_

dgolovin:
### What does this PR do?

Adds internal extension for kubectl registration and tasks to package.json to build/test/type check.

image

Partial fix #4118


extensions/kubectl-cli/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "license": "Apache-2.0",
+  "engines": {
+    "podman-desktop": "^0.0.1"
+  },
+  "main": "./dist/extension.js",
+  "contributes": {
+    "cliTools": {
+      "name": "kubectl",
+      "displayName": "kubectl",
+      "markdownDescription": "A command line tool used to run commands against Kubernetes clusters. It does this by authenticating with the Master Node of your cluster and making API calls to do a variety of management actions. If you're just getting started with Kubernetes, prepare to be spending a lot of time with kubectl",
```
afbjorklund: "Control Plane Node"

https://kubernetes.io/blog/2022/04/07/upcoming-changes-in-kubernetes-1-24/

https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint

https://github.com/kubernetes/community/blob/master/sig-architecture/naming/recommendations/001-master-control-plane.md

afbjorklund (changes requested):
The terms master and minion should not be used with Kubernetes anymore.

extensions/kubectl-cli/icon.png
afbjorklund, line 1: The logo looks inverted and square, and has a fairly low resolution

afbjorklund (changes requested):

extensions/kubectl-cli/.extfiles
```diff
@@ -0,0 +1,5 @@
+dist/**
+node_modules/**
+icon.png
+LICENSE
+package.json
```
benoitf: hello, what is it for ?

extensions/kubectl-cli/.gitignore
```diff
@@ -0,0 +1,130 @@
+# Logs
```
benoitf: AFAIK this should be reduced

there is a global .gitignore at the root folder that include many of the ignore

this file also contains some reference to old tools that we don't use (bower, grunt), etc.

extensions/kubectl-cli/package.json
dgolovin: Updated text and added link to additional info similar to compose.

benoitf: yes I would avoid `master` here

extensions/kubectl-cli/package.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "scripts": {
+    "build": "vite build && node ./scripts/build.js",
+    "test": "vitest run --coverage",
+    "watch": "vite build -w",
+    "format:check": "prettier --check \"**/*.ts\" \"scripts/*.js\"",
+    "format:fix": "prettier --write \"**/*.ts\" \"scripts/*.js\""
+  },
+  "dependencies": {
+    "@podman-desktop/api": "^1.2.0"
```
benoitf: not sure about this version as it's in the repo 0.0.1

extensions/kubectl-cli/icon.png
dgolovin: @mairin could you help to create proper Icon? I used this one https://kubernetes.io/images/wheel.svg.

extensions/kubectl-cli/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "@types/node": "^18.15.11",
+    "7zip-min": "^1.4.3",
+    "byline": "^5.0.0",
+    "copyfiles": "^2.4.1",
+    "mkdirp": "^2.1.3",
+    "prettier": "^2.8.8",
+    "typescript": "^5.0.4",
+    "vite": "^4.3.9",
+    "vitest": "^0.34.6",
+    "zip-local": "^0.3.5"
```
benoitf: it looks like we have more dev dependencies that expected as some of them are inherited from the root folder and AFAIK some here are not used by the extension

extensions/kubectl-cli/src/extension.ts
```diff
@@ -0,0 +1,33 @@
...
+  cli.createCliTool({
+    markdownDescription:
+      'A command line tool used to run commands against Kubernetes clusters. It does this by authenticating with the Control Plane Node of your cluster and making API calls to do a variety of management actions. If you are just getting started with Kubernetes, prepare to be spending a lot of time with kubectl.\n\nMore information: [kubernetes.io](https://kubernetes.io/docs/reference/kubectl/)',
+    name: 'kubectl',
+    displayName: 'kubectl',
+    images: {
+      icon: 'icon.png',
+    },
+    version: '',
+    path: '',
```
benoitf: there is something missing here no ? like it should be registered only if we found it for now ?
And if we found it we set the version that we found ?

extensions/kubectl-cli/icon.png
benoitf: you can take this existing one: https://github.com/containers/podman-desktop/blob/main/extensions/kube-context/icon.png

extensions/kubectl-cli/.extfiles
dgolovin: This is just a config file with a list of files included into cdix archive. It is used in scripts/build.js.

extensions/kubectl-cli/.gitignore
dgolovin: That is default generated by github.com, will clean it up.

extensions/kubectl-cli/src/extension.ts
dgolovin: I think I missed that one when we pushed API. Those to fields should be optional. It should possible to register without version and location, then podman-desktop would do detection and fill that fields up.

extensions/kubectl-cli/.gitignore
dgolovin: It is obviously not needed at all, because one in the root of the repo works fine

dgolovin: removed.

extensions/kubectl-cli/.extfiles
dgolovin: LICENSE file does not exists. I removed it form the list.

extensions/kubectl-cli/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "description": "Install and update kubectl CLI Tools without leaving Podman Desktop",
+  "version": "0.0.1",
+  "icon": "icon.png",
+  "publisher": "redhat",
+  "license": "Apache-2.0",
+  "engines": {
+    "podman-desktop": "^0.0.1"
+  },
+  "main": "./dist/extension.js",
+  "contributes": {
```
dgolovin: This is not implemented and should be removed. Only cli.registerCliTool() call should be used to register tools

extensions/kubectl-cli/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "kubectl-cli-tool",
+  "displayName": "kubectl CLI Tool for Podman Desktop",
```
deboer-tim: Can we remove 'tool' from the name and shorten the displayName to something like `kubectl CLI`? It is redundant and long, and doesn't match other extensions (e.g. we have `Docker`, not `Docker Tool for Podman Desktop`.

dgolovin:
```suggestion
  "displayName": "kubectl CLI",
```

extensions/kubectl-cli/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "kubectl-cli-tool",
```
benoitf, line 2: I guess we can shorter the name as well
```suggestion
  "name": "kubectl-cli",
```

extensions/kubectl-cli/src/extension.ts
benoitf: no, we only register for now if it has been found

not found : nothing
found: provides the path of the CLI and provides the version

benoitf: once installation would be available, it could provide an optional installation process but it's not there for now

mairin:
I think the text description is too long. This is what the mockup has:

Screenshot from 2023-11-09 07-49-14

I think the more info line you added is fine, but seriously if users wanna know more than just that once sentence, they can look it up :)

benoitf:
code is LGTM but tests are failing

if you can apply @mairin suggestion by dropping some text

dgolovin:
Updated description. I decided to keep `More information' part, so it is easier to get additional info about kubectl.

extensions/kubectl-cli/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "engines": {
+    "podman-desktop": "^0.0.1"
+  },
+  "main": "./dist/extension.js",
+  "scripts": {
+    "build": "vite build && node ./scripts/build.js",
+    "test": "vitest run --coverage",
+    "watch": "vite build -w",
+    "format:check": "prettier --check \"**/*.ts\" \"scripts/*.js\"",
+    "format:fix": "prettier --write \"**/*.ts\" \"scripts/*.js\""
```
lstocchi, line 18: We already have the format scripts in the main package.json. Any reason to have it here as well?

extensions/kubectl-cli/src/extension.ts
```diff
@@ -0,0 +1,85 @@
...
+  return { version, path };
+}
+
+const markdownDescription = `A command line tool for communicating with a Kubernetes cluster's control plane, using the Kubernetes API.\n\nMore information: [kubernetes.io](https://kubernetes.io/docs/reference/kubectl/)`;
+
+async function registerTool(cliInfo: KubectlInfo) {
+  cli.createCliTool({
+    markdownDescription,
+    name: 'kubectl',
+    displayName: 'kubectl',
```
lstocchi, line 78: Should it have the first char uppercase? I see that compose has both the title and name in `registered by` label with the first capital letter `Compose`
image

lstocchi (approved):
LGTM, just a couple of nits

extensions/kubectl-cli/src/extension.ts
```diff
@@ -0,0 +1,85 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { CliToolOptions } from '@podman-desktop/api';
+import { cli, process as podmanProcess } from '@podman-desktop/api';
```
deboer-tim, line 20: I didn't notice this until below, where I started wondering why we were exec-ing something on podman to find kubectl. Can we use something more generic or just not rename?

extensions/kubectl-cli/src/extension.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  return deferred.then(() => {
+    expect(console.log).toBeCalled();
+    expect(console.log).toBeCalledWith(
+      expect.stringContaining('Cannot detect kubectl CLI tool: Error: Cannot extract version from stdout'),
+    );
+  });
+});
+
+test('kubectl CLI tool not registered when path cannot be found', async () => {
```
deboer-tim, line 161: This and the above case are a bit odd to me, because we know kubectl _is_ installed but we're reporting it's not because we either can't tell what version it is or where it is. Shouldn't we be reporting partial data or giving the error with the data we do know? Or maybe the API doesn't support that yet?

dgolovin: > @benoitf: 
>
> no, we only register for now if it has been found
> 
> not found : nothing
> found: provides the path of the CLI and provides the version

@deboer-tim it was a request from @benoitf (see resolved conversations above). Compose extension seems to register it no matter if it found or not.

extensions/kubectl-cli/src/extension.ts
dgolovin: @deboer-tim It is just a name for process module so it is not conflicting with global var process. I use podmanProcess.exec('kubectl`, ['--version']) to test if kubectl is available on users machine. It has nothing to do with podman executable.

extensions/kubectl-cli/src/extension.ts
dgolovin: On Kubernetes web site they use it the same way, no uppercase.

image

extensions/kubectl-cli/package.json
dgolovin: It just works much faster when it checks only extension.

benoitf (approved):

PR #4683, opened 2023-11-07
feat: add editable item for cpu/memory/disk (#2319)

### LINKED ISSUE #2319: Sliders for adjusting the settings of a podman-machine are not convenient to use

While the sliders are quite easy way to interact with the settings of the podman machine, it's super tricky to adjust the value to a precise one.
If somebody, just want to create a podman-machine with 4GB of RAM, most likely, the user will get "around it": 3.96, 4.07 ...

We should make the sliders having better "graduations".
For Disk: I believe a single digit unique would be enough 
For Ram: Maybe 0.5 increments

Also it would be nice to display the scales (from which min value it starts, to which max value it goes)

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR enhances the slider item. It increases the step value and allow to edit the value manually

editable_item_float

it resolves #2319 

### How to test this PR?

1. create a podman machine and adds memory/cpu/disk by using the editable input

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationRendering.svelte
@@ -3,7 +3,7 @@ import type { IConfigurationPropertyRecordedSchema } from '../../../../main/src/
 import type { ProviderInfo } from '../../../../main/src/plugin/api/provider-info';
 import PreferencesRenderingItemFormat from './PreferencesRenderingItemFormat.svelte';
 import TerminalWindow from '../ui/TerminalWindow.svelte';
-import { getNormalizedDefaultNumberValue, writeToTerminal, isPropertyValidInContext, getInitialValue } from './Util';
+import { writeToTerminal, isPropertyValidInContext, getInitialValue } from './Util';
 import ErrorMessage from '../ui/ErrorMessage.svelte';
 import {
   clearCreateTask,
@@ -19,7 +19,6 @@ import { get, type Unsubscriber } from 'svelte/store';
 import { onDestroy, onMount } from 'svelte';
 /* eslint-enable import/no-duplicates */
 import { createConnectionsInfo } from '/@/stores/create-connections';
-import { filesize } from 'filesize';
 import { router } from 'tinro';
 import LinearProgress from '../ui/LinearProgress.svelte';
 import Spinner from '../ui/Spinner.svelte';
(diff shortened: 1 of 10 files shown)


lstocchi:
@mairin here's the editable input with the pencil icon. WDYT?

mairin:
> @mairin here's the editable input with the pencil icon. WDYT?

I love this!!! I think the stairstep alignment of the icons looks natural and functionally makes it easier to target the right one to click. The animation in/out of the field looks slick. A++++++ WOULD BUY AGAIN

🫡

deboer-tim (changes requested):
I love it! But one nit and one bug:

Nit: when you edit it shifts the rest of the UI down 1 pixel, and it moves up again 1px when you finish. We should make the original control slightly bigger or do something else so it doesn't 'expand' on opening and the rest of the UI doesn't move. 

Bug: One of the main reasons for editing is that we're making the slider have nice jump points like 3, 3.5, or 4Gb, so if you want a more unique memory value like 3.75Gb you would edit instead. The problem is that I can't use `.` in the edit control, so I can only enter a whole value like 3 or 4.

deboer-tim (commented):
Another issue: there is some weird behaviour going on with the green checkmark enablement. Sometimes when I open it up I can immediately click it to close; other times you can only click the check if you change the value first. Likewise if I return to the original value sometimes it is enabled and sometimes it isn't. IMHO it should be enabled whenever the value is valid (and never when it isn't).

And FWIW I found a workaround to enter a value like 14.75Gb. If you start at `14.5` and try to delete the `5` it (incorrectly) deletes the `.`, but if you start at `14.5` and insert a `7` it works. ;-)

I also saw that you split parts of this out to other PRs, please consider my comments here on #4685 as appropriate.

lstocchi:
@deboer-tim updated the code and the gif. Please give it another look 👍

deboer-tim (changes requested):
Regular editing flow is now awesome. 👍

Still some concerns though:
- The left side of the error tooltip is getting cut off for me. It either needs a higher z-order or move to another position so it doesn't go off the left of the form.
- The green check is still disabled when you're at the original value (EditableItem:17). I find this confusing, esp. if you edit a value twice and don't realize you've edited it back to the original value. IMHO it's better if we leave it enabled (unless there's an error).

Had these ideas too, although they feel more like polish/bonus:
- It would be nice if the check was grey instead of green when disabled, it looks clickable until you mouse over it.
- It would be really slick if the slider changed as you edit, so you get an idea of where it is in the range. (and reverts back if you cancel)

lstocchi:
@deboer-tim Updated 
slider_edit

deboer-tim (approved):
All concerns addressed. Thank you for sticking through with this, it's really slick and easy to use now!

benoitf (approved):
I don't know if you were waiting for another review

lstocchi:
> I don't know if you were waiting for another review

Yeah :) Thanks!! 🚀

PR #4684, opened 2023-11-07
feat: add slider step for memory/disk when creating a podman machine (#2319)

lstocchi:
### What does this PR do?

This PR adds a step to the podman machine memory/disk sliders.
The code has been extracted from https://github.com/containers/podman-desktop/pull/4683 to simplify the review.

slider_step

it is part of #2319 

### How to test this PR?

1. create a new podman machine and check the sliders value change at a 0.5 step

### DIFF
--- a/extensions/podman/package.json
+++ b/extensions/podman/package.json
@@ -101,6 +101,7 @@
           "default": 4000000000,
           "maximum": "HOST_TOTAL_MEMORY",
           "scope": "ContainerProviderConnectionFactory",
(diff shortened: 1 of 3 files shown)


deboer-tim (approved):
I am so happy to see that when running code it doesn't leave artifacts on the slider bar like in the video. ;-)

Looks much better when editing, and far easier to set a whole value. Allowing .5 steps worked well for me too - makes sense for memory granularity and wasn't too fine grained for disk.

PR #4685, opened 2023-11-07
feat: add EditableItem component to edit numeric value manually

lstocchi:
### What does this PR do?

This adds an editableItem component which will be used by #4683 

This has been extracted from there to make the review easier.

N/A
you can check it in #4683 

it is part of #2913 

### How to test this PR?

1. run tests

### DIFF
--- a/packages/renderer/src/lib/ui/Button.svelte
+++ b/packages/renderer/src/lib/ui/Button.svelte
@@ -14,7 +14,7 @@ export let selected: boolean | undefined = undefined;
 $: if (selected !== undefined && type !== 'tab') {
   console.error('property selected can be used with type=tab only');
 }
-export let padding: string | undefined = 'px-4 py-[6px]';
+export let padding: string = 'px-4 py-[6px]';
 
 let iconType: string | undefined = undefined;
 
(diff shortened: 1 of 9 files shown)


packages/renderer/src/lib/ui/Button.svelte
```diff
@@ -42,7 +43,8 @@ $: {
 
 <button
   type="button"
-  class="relative px-4 py-[6px] rounded-[4px] box-border text-[13px] whitespace-nowrap select-none transition-all {classes} {$$props.class ||
+  class="relative {padding ||
+    'px-4 py-[6px]'} rounded-[4px] box-border text-[13px] whitespace-nowrap select-none transition-all {classes} {$$props.class ||
```
benoitf: I'm wondering if instead of having `{padding || 'px-4 py-[6px]'}` we could not initialize padding to `px-4 py-[6px]` instead of undefined so we always use what is inside `{padding}`

lstocchi:
Updated based on @deboer-tim's review

benoitf (approved):

jeffmaury (approved):
LGTM

PR #4692, opened 2023-11-07
fix: avoid to send telemetry usage as this method is called every 5s

### LINKED ISSUE #4687: Telemetry overload from configuration registry
### Bug description

PR #3111 added telemetry to the configuration-registry.ts info() function, which is called every 5s and produces way too many telemetry events. We need to remove this call or trigger it only on user action to reduce the load on our systems.

### Operating system

macOS 14

### Installation Method

None

### Version

1.5.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Do not send telemetry usage

fixes https://github.com/containers/podman-desktop/issues/4687

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -1987,37 +1987,29 @@ export class ContainerProviderRegistry {
   }
 
   async info(engineId: string): Promise<containerDesktopAPI.ContainerEngineInfo> {
-    let telemetryOptions = {};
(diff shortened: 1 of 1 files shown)


benoitf:
should be backported to v1.5.x

deboer-tim (approved):
LGTM

jeffmaury (approved):
LGTM

PR #4733, opened 2023-11-09
chore: send bundled telemetry at the end of onboarding session

### LINKED ISSUE #4590: Improve telemetry for onboarding process (send a bundled log)

Reference: https://github.com/containers/podman-desktop/pull/4572#pullrequestreview-1706816125

We should send a bundled telemetry event rather than separate ones (one for each registerCommand), so we can see a summary of the entire process rather than each individual event.

Send a single onboarding event on failure / success, as per Tim's example:

```
  version:
  download:
```

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

This PR makes changes to the onboarding process, by collecting telemetry data from the commands executed during the onboarding (through the response of commands), and sending the bundled data at the end of the onboarding session.

Fixes #4590

### DIFF
--- a/extensions/compose/src/extension.ts
+++ b/extensions/compose/src/extension.ts
@@ -95,6 +95,12 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
         downloaded: isDownloaded === '' ? false : true,
         version: composeVersionMetadata?.tag,
       });
+      return {
+        telemetry: {
+          downloaded: isDownloaded === '' ? false : true,
+          version: composeVersionMetadata?.tag,
+        },
+      };
     },
   );
 
--- a/packages/renderer/src/lib/onboarding/Onboarding.svelte
+++ b/packages/renderer/src/lib/onboarding/Onboarding.svelte
@@ -67,6 +67,8 @@ let displayCancelSetup = false;
 
 let executedCommands: string[] = [];
 
+let telemetry: any = {};
+
 /*
 $: enableNextButton = false;*/
 let onboardingUnsubscribe: Unsubscriber;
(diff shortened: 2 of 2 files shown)


benoitf (commented):
AFAIK compose should not know anything about telemetry

extensions/compose/src/extension.ts
```diff
@@ -95,6 +95,12 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
         downloaded: isDownloaded === '' ? false : true,
         version: composeVersionMetadata?.tag,
       });
+      return {
```
benoitf, line 98: compose should not return special objects I would say

feloy:
> AFAIK compose should not know anything about telemetry

The current code for the compose extension is using the `createTelemetryLogger` function available from the API to send telemetry directly, how is it different? 

Which part of the app is expected to deal with telemetry, `renderer`?

benoitf:
I mean, tracking what has been done by the user is independent of the wish of the extension wanting to send telemetry events


There should not be any telemetry contract on callbacks/signature except calling explicitly telemetry methods

lstocchi:
I wonder if we should allow to create a sort of transaction for telemetry.

So, from an extension, you can call a method begin transaction, which returns an id. You add content to that, and when you want to push it you call merge transaction. So we could reuse it in other places if needed.

For the onboarding we can say if the workflow is cancelled/skipped/completed, if there is any open transaction it gets closed and merged automatically, otherwise a generic telemetry with the steps name or anything else is sent

feloy:
closing in favour of #4736

PR #4736, opened 2023-11-09
chore: send bundled onboarding telemetry

### LINKED ISSUE #4590: Improve telemetry for onboarding process (send a bundled log)

feloy:

### What does this PR do?

This PR makes Podman Desktop send bundled telemetry when an onboarding session terminates, either correctly or canceled.

The data sent to the telemetry is the list of steps done during the onboarding.

When an error occurs during a step, the Error is added to the step information.

The duration for each step and for the entire onboarding session is given.

Fixes #4590 

### How to test this PR?

Run any onboarding, and check the data sent to telemetry at the end of the onboarding session.

### DIFF
--- a/packages/renderer/src/lib/onboarding/Onboarding.svelte
+++ b/packages/renderer/src/lib/onboarding/Onboarding.svelte
@@ -54,6 +54,7 @@ import Button from '../ui/Button.svelte';
 import Link from '../ui/Link.svelte';
 import OnboardingComponent from './OnboardingComponent.svelte';
 import Spinner from '../ui/Spinner.svelte';
+import { TelemetrySession } from './telemetry';
 
 export let extensionIds: string[] = [];
 
@@ -66,6 +67,8 @@ let globalContext: ContextUI;
 let displayCancelSetup = false;
 
 let executedCommands: string[] = [];
+
+let telemetrySession = new TelemetrySession();
 
 /*
 $: enableNextButton = false;*/
@@ -111,6 +114,7 @@ async function startOnboarding(): Promise<void> {
   if (!started && globalContext && onboardings.length > 0) {
     started = true;
     if (!isOnboardingsSetupCompleted(onboardings)) {
+      telemetrySession.restart();
       await restartSetup();
     }
   }
@@ -141,6 +145,7 @@ async function setActiveStep() {
             whenDeserialized = ContextKeyExpr.deserialize(when);
           }
           if (!step.when || whenDeserialized?.evaluate(globalContext)) {
+            telemetrySession.startStep(i, step.id, step.title);
             activeStep = {
               onboarding,
               step,
@@ -154,7 +159,13 @@ async function setActiveStep() {
                 });
               }) || [];
             if (step.command) {
-              await doExecuteCommand(step.command);
+              try {
+                await doExecuteCommand(step.command);
+              } catch (error: unknown) {
+                if (error instanceof Error) {
+                  telemetrySession.setStepError(i, step.id, error);
+                }
+              }
               // after command has been executed, we check if the step must be marked as completed
               await assertStepCompletedAfterCommandExecution();
             }
@@ -167,6 +178,7 @@ async function setActiveStep() {
       }
     }
   }
+  telemetrySession.send(onboardings.map(o => o.extension).join(','), false);
   // if it reaches this point it means that the onboarding is fully completed and the user is redirected to the dashboard
   router.goto($lastPage.path);
 }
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/onboarding/Onboarding.svelte
```diff
@@ -167,6 +189,10 @@ async function setActiveStep() {
       }
     }
   }
+  telemetryData.extensions = onboardings.map(o => o.extension).join(',');
+  telemetryData.skipped = false;
+  console.log('telemetry: onboarding.setupDone', telemetryData);
```
feloy: To be removed before to merge

packages/renderer/src/lib/onboarding/Onboarding.svelte
```diff
@@ -262,9 +288,10 @@ async function cancelSetup() {
   // TODO: it cancels all running commands
   // it redirect the user to the dashboard
   await cleanSetup(onboardings, globalContext);
-  window.telemetryTrack('onboarding.cancelSetup', {
-    extensions: onboardings.map(o => o.extension).join(','),
-  });
+  telemetryData.extensions = onboardings.map(o => o.extension).join(',');
+  telemetryData.skipped = true;
+  console.log('telemetry: onboarding.cancelSetup', telemetryData);
```
feloy: To be removed before to merge

extensions/compose/src/extension.ts
```diff
@@ -89,12 +80,6 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
           extensionApi.context.setValue('composeDownloadVersion', composeVersionMetadata.tag, 'onboarding');
         }
       }
-
-      // Log if it's downloaded and what version is being selected for download (can be either latest, or chosen by user)
-      telemetryLogger.logUsage('compose.onboarding.checkComposeDownloaded', {
-        downloaded: isDownloaded === '' ? false : true,
-        version: composeVersionMetadata?.tag,
-      });
```
benoitf: should we remove the current events being done ?

I mean, it's interesting to know the workflow of users but for a detailed event it might be good to know more specifics like the version of compose being installed by users ?

feloy:
The array of steps is displayed in Amplitude as a string, and not separate properties. Is it exploitable?
amplitude-onboarding

extensions/compose/src/extension.ts
feloy: @deboer-tim WDYT?

feloy: We will keep both

feloy:
Steps are sent both in a single `steps` property and in a series of numbered propreties:

telemetry-steps-numbered

packages/renderer/src/lib/onboarding/telemetry.ts
```diff
@@ -0,0 +1,133 @@
...
+
+function telemetryAddStep(data: TelemetryData, i: number, id: string, title: string): TelemetryData {
+  const completeId = getStepCompleteId(i, id);
+  data.steps.push({ id: completeId, title, durationMs: 0, error: undefined });
+  return data;
+}
+
+function telemetrySetStepDuration(data: TelemetryData, i: number, stepId: string, ms: number): TelemetryData {
+  const completeId = getStepCompleteId(i, stepId);
+  console.log('step duration', completeId, ms);
```
benoitf:
```suggestion
  console.log('step duration', completeId, ms);
```

probably remove the debug console output

packages/renderer/src/lib/onboarding/telemetry.ts
```diff
@@ -0,0 +1,133 @@
...
+      };
+    }
+    data.stepsDetails[step.id].count++;
+    data.stepsDetails[step.id].durationMs.push(step.durationMs);
+    if (step.error) {
+      data.stepsDetails[step.id].errors.push(step.error);
+      data.errors.push(`${step.error} [${step.title}]`);
+    }
+  }
+  console.log('data', data);
```
benoitf:
```suggestion
```

packages/renderer/src/lib/onboarding/telemetry.ts
```diff
@@ -0,0 +1,134 @@
...
+  extension: string;
+  skipped: boolean;
+  steps: TelemetryStep[];
+  stepsDetails?: any;
+  durationMs: number;
+  skipAtStep?: string;
+  errors?: string[];
+}
+
+export class TelemetrySession {
```
benoitf: just a minor thing but as it's scoped only for onboarding, maybe we should prefix it with Onboarding so we're not tempted to use it elsewhere than the onboarding.

TelemetrySession --> OnboardingTelemetrySession

benoitf (approved):

PR #4745, opened 2023-11-09
chore: add stores to kubernetes contexts

### LINKED ISSUE #4688: Kubernetes Context UI: Add store to make it reactive to configuration changes / updates.

Right now the Kubernetes configuration section will only update when you restart Podman Desktop.

Implement using store values it changes whenever `.kube/config` has been changed. 

_No response_

### Additional context

_No response_

cdrage:
chore: add stores to kubernetes contexts

### What does this PR do?
* Adds apiSender events for Kubernetes context changes
* Switches kubernetes contexts to a store
* Changes are reactive / reflective when changes have been made to the
  ~/.kube/config

https://github.com/containers/podman-desktop/assets/6422176/84e440d8-2d2d-4a41-88e5-8149cba4e2b6

Closes #4688

### How to test this PR?

1. View your Kubernetes context via Podman Desktop
2. Make changes to your .kube/config and you should see them change on
   PD.

### DIFF
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -185,6 +185,9 @@ export class KubernetesClient {
     });
   }
 
+  // The below methods (update, create, delete)
+  // help send information to the renderer process to notify any
+  // stores (in particular kube context store) as well as extensions (extensions/kube-context)
   setupWatcher(kubeconfigFile: string): void {
     // cancel the previous one (if any)
     this.kubeConfigWatcher?.dispose();
@@ -198,6 +201,7 @@ export class KubernetesClient {
     this.kubeConfigWatcher.onDidChange(async () => {
       this._onDidUpdateKubeconfig.fire({ type: 'UPDATE', location });
       await this.refresh();
+      this.apiSender.send('kubernetes-context-update');
     });
 
     this.kubeConfigWatcher.onDidCreate(async () => {
--- a/packages/renderer/src/stores/kubernetes-contexts.ts
+++ b/packages/renderer/src/stores/kubernetes-contexts.ts
@@ -1,0 +1,56 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { writable, type Writable } from 'svelte/store';
+import { EventStore } from './event-store';
+import { getKubeUIContexts, type KubeContextUI } from '../lib/kube/KubeContextUI';
+
+const windowEvents: string[] = ['extensions-started', 'kubernetes-context-update'];
+const windowListeners = ['extensions-already-started'];
+
+// Do not update until all extensions are started (since we want to make sure kube-context has been loaded)
+let readyToUpdate = false;
+export async function checkForUpdate(eventName: string): Promise<boolean> {
+  if ('extensions-already-started' === eventName) {
+    readyToUpdate = true;
+  }
+
+  // do not fetch until extensions are all started
+  return readyToUpdate;
+}
+
+export const kubernetesContexts: Writable<KubeContextUI[]> = writable([]);
+
+const eventStore = new EventStore<KubeContextUI[]>(
+  'kubernetes contexts',
+  kubernetesContexts,
+  checkForUpdate,
+  windowEvents,
+  windowListeners,
+  grabKubernetesContexts,
+);
+eventStore.setup();
+
+export async function grabKubernetesContexts(): Promise<KubeContextUI[]> {
+  // Retrieve both the contexts and clusters
+  const k8sContexts = await window.kubernetesGetContexts();
+  const k8sClusters = await window.kubernetesGetClusters();
+
+  // Convert them to KubeContextUI so we can safely render them.
+  return getKubeUIContexts(k8sContexts, k8sClusters);
+}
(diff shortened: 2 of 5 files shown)


packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -568,6 +573,7 @@ export class KubernetesClient {
     await this.refresh();
     // notify change
     this._onDidUpdateKubeconfig.fire({ type: 'UPDATE', location });
+    this.apiSender.send('kubernetes-context-update');
```
benoitf: I'm wondering here, if we should not add a listener to onDidUpdateKubeConfig as there is already a listener

and then we send the event

Also maybe we could just throw the same event, providing the type as an argument ?

it would avoid to miss some locations where we might fire the event

so, subscribe too onDidUpdateKubeConfig and send to apiSender on this listener

jeffmaury (approved):
LGTM

packages/renderer/src/stores/kubernetes-contexts.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { writable, type Writable } from 'svelte/store';
+import { EventStore } from './event-store';
+import { getKubeUIContexts, type KubeContextUI } from '../lib/kube/KubeContextUI';
+
+const windowEvents: string[] = [
+  'extensions-started',
+  'kubernetes-context-update',
+  'kubernetes-context-create',
+  'kubernetes-context-delete',
```
benoitf: if we have only one event it looks fine no ?

packages/renderer/src/stores/kubernetes-contexts.ts
```diff
@@ -0,0 +1,61 @@
...
+  windowEvents,
+  windowListeners,
+  grabKubernetesContexts,
+);
+eventStore.setup();
+
+export async function grabKubernetesContexts(): Promise<KubeContextUI[]> {
+  // Retrieve both the contexts and clusters
+  const k8sContexts = await window.kubernetesGetContexts();
+  const k8sClusters = await window.kubernetesGetClusters();
```
benoitf, line 52: do we know how costly it is to call these ?

cdrage: Not costly at all, as we previously used this in the render. 

I did make them into a "combined" window call but noticed no difference in speed.

packages/main/src/plugin/kubernetes-client.ts
cdrage: I noticed that too, but I wanted to keep this similar to how we use Pods + Containers for stores and rely on api sender requests.

We do not need a `onDidUpdateKubeConfig` listener as there already is one in the kube context extension.

The reason why I do not want to add another `onDidUpdateKubeConfig` is that it's handled by the kube context extension, so if we disable the context extension it would also disable us updating the settings as well if the user disabled the kube context extension.

packages/renderer/src/stores/kubernetes-contexts.ts
cdrage: Changed to only check kubeneretes-context-update! Thanks for pointing that out.

cdrage:
Updated! Had to change the Kubernetes context delete function a little bit, but ready for another review.

I addressed all of your comments @benoitf (see resolved comments).

packages/renderer/src/stores/kubernetes-contexts.ts
benoitf: my question is that we're calling kubernetes API so, it may take time to grab these informations if clusters are remote (just a thought, maybe it's not true)

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -198,16 +201,19 @@ export class KubernetesClient {
     this.kubeConfigWatcher.onDidChange(async () => {
       this._onDidUpdateKubeconfig.fire({ type: 'UPDATE', location });
       await this.refresh();
+      this.apiSender.send('kubernetes-context-update');
```
benoitf, line 204: as reported before I think we would need to subscribe to the onDidUpdateKubeconfig to fire the event when it's there
so it avoids to add apiSender on kubeConfigWatcher as we've already a listener that can call our apiSender method

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -198,16 +201,19 @@ export class KubernetesClient {
...
     });
 
     this.kubeConfigWatcher.onDidCreate(async () => {
       this._onDidUpdateKubeconfig.fire({ type: 'CREATE', location });
       await this.refresh();
+      this.apiSender.send('kubernetes-context-create');
```
benoitf: so it seems we're never looking at these events so maybe we should not send them instead ?

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -198,16 +201,19 @@ export class KubernetesClient {
...
     });
 
     this.kubeConfigWatcher.onDidDelete(() => {
       this._onDidUpdateKubeconfig.fire({ type: 'DELETE', location });
       this.kubeConfig = new KubeConfig();
+      this.apiSender.send('kubernetes-context-delete');
```
benoitf: so it seems we're never looking at these events so maybe we should not send them instead ?

or you send kubernetes-context-event with the type 'DELETE', 'UPDATE' or 'CREATE' by being a client of onDidUpdateKubeconfig

packages/main/src/plugin/kubernetes-client.ts
cdrage: No need to add another onDidUpdateKubeconfig as we do that in kube context: https://github.com/containers/podman-desktop/blob/d64d20c7c4b50b525e10280148364505da5dff9c/extensions/kube-context/src/extension.ts#L158 

When I added an onDidUpdateKubeconfig outside of the extension, it was firing twice / refreshing twice all the time. 

Since kube context is in an extension that can be disable / enabled, this makes sense.

packages/renderer/src/stores/kubernetes-contexts.ts
cdrage: I've updated this in https://github.com/containers/podman-desktop/pull/4921 and refactored it so it's less costly (we now do the conversion in the plugin rather than the renderer), since we need to get the current kube context name anyways.

cdrage:
Updated @benoitf 

I've added some changes to this PR too where the kubernetes calls are refactored when implementing the "get current context" change: https://github.com/containers/podman-desktop/pull/4921

See https://github.com/containers/podman-desktop/pull/4745#discussion_r1402078690 for the reason why I do not add another watcher (since we already watch in the kube-context extension).

benoitf (approved):

PR #4762, opened 2023-11-10
chore: drop patternfly

### LINKED ISSUE #1617: UI: Drop patternfly and use only tailwind (technical debt)

When implementing patternfly + tailwindcss you have to bounce between both tailwind as well as patternfly to implement a feature.

This causes technical debt with maintaining two different frameworks.

Remove patternfly and use only tailwind.

But "components" and UI templates into the `/ui` folder, etc.

_No response_

### Additional context

_No response_

### LINKED ISSUE #4319: PatternFly cleanup
### Epic domain

We no longer have any direct dependency on PatternFly, but removing it breaks three things:

1. Missing RedHatText font.
2. Missing Font Awesome 5 font, along with the fa and fas classes.
3. Some global CSS properties. The only two that I've found make any difference can be replaced by:
```
  color-scheme: dark;
  line-height: 0.9rem;
```

In order to remove PatternFly completely we'll need to remove it from App.svelte (and eventually the packaging) at the same time we release PR(s) to fix all of these things.

The only immediate benefits will be a slightly smaller package/load time, but this would make it easier in the future for us to do light mode/themes, and have only one copy of Font Awesome (6 vs current 5 + 6).

### Additional context

_No response_

benoitf:
### What does this PR do?
All usage of patternfly have been dropped so it's time to be able to remove patternfly

I also remove the overrides that were done on top of patternfly as we no longer them

I changed default colour of text to be white and bring font-awesome font/css before it was being provided through patternfly but as it is removed it was not there anymore

we might have some little glitches this is why I wanted to get this merge long time before a new release so we have to identify them/fix them

We've lost the Redhat Font also, I don't know if we want to pickup another default font.

fixes https://github.com/containers/podman-desktop/issues/1617
fixes https://github.com/containers/podman-desktop/issues/4319

### How to test this PR?

Look at the UI


jeffmaury (approved):
LGTM although I see some light differences in font on some screens (eg Create container dialog):

After:
image

1.5.3:
image

benoitf:
@jeffmaury 
so how do we want to proceed there ?

Try to fix all issues coming from dropping patternfly before the merge or merge early and do follow-up PRs to get all breakages being fixed.

Also, do you prefer with the new font or with older font size

jeffmaury:
> @jeffmaury so how do we want to proceed there ?
> 
> Try to fix all issues coming from dropping patternfly before the merge or merge early and do follow-up PRs to get all breakages being fixed.
> 
> Also, do you prefer with the new font or with older font size

No for me it's not a blocker just recording for later issues

deboer-tim (approved):
The CSS override was setting a line height of 0.9rem, we might want to keep that for now just for less change and smaller screens.

I've been running with most of these changes or variants of them already and LGTM. I think we're close enough that it's better to switch early - more time for everyone to catch regressions, no risk of new dependencies on PF, benefits like single version of FA, etc.

benoitf:
ok let's merge it now then

afbjorklund:
Here is what I got, on Linux:

Before: before

After: after

The text is now invisible (until selected)

benoitf:
thanks @afbjorklund

PR #4782, opened 2023-11-13
feat: add purple dot when new content is available in dashboard (#4043)

### LINKED ISSUE #4043: Add a purple dot when new content available in Dashboard
### Epic
This is part of the Dashboard Revamp epic (https://github.com/containers/podman-desktop/issues/4020).

## Summary:

Add a purple notification dot to the dashboard icon when new content is available on the dashboard.

### User perspective:

Signalling that new content is available on the dashboard via a small notification dot on the dashboard tab will let the user know there's something new having to do with their environment - e.g., we detected a new container engine - with a convenient direction to address that change - e.g., click here to set up this new container engine. It also can flag to them new content related to their container work / Podman Desktop has been posted that they might want to evaluate taking a look at - e.g. the latest release notes for their auto-updated version of Podman Desktop.

In many cases users may see the purple dot and click on the dashboard icon to quickly dismiss it (like popping bubbles in bubble wrap - it's satisfying to remove the dot, and less irritating - however mildly - to not have it in view. Understanding this, the design of the dashboard should be such that the most important content - the notifcation block - is on the top of the screen, so even "bubble wrap poppers" have a chance to see it. The example / tutorial content should be lower on the screen since it's less critical.

## Rough specification:

When we add more content to the dashboard, including new notifications (e.g., we detected a new Podman install) or new examples / tutorials (e.g. new HOWTO on using compose) we put a purple notification dot to the upper right of the dashboard icon in the global nav, to signal there is something to see there.

The dot should clear as soon as the user clicks on the dashboard icon.

It might be nice to highlight the net new content so it's easy to find, and have its highlight fade out after 15s or so. (Suggestion documented in issue # [TBF] )

If a new piece of content was added to the dashboard at the same time the app was started up - because you're placed on the dashboard to start, don't bother with the dot. (Still can do the background highlight.) 

The dot would only practically appear if new content is surfaced while the Podman Desktop application is open AND the user is on any tab or screen other than the dashboard tab/screen.

## Design specification

The dot is purple-500 and s 6px x 6px. It has a 3px space between its left bounding box and the icon's right bounding box. Its top bounding box is flush with the icon's top bounding box.

Screenshot from 2023-09-22 07-57-09

Here is what the dot would look like in situ:

Screenshot from 2023-09-22 07-57-41

### Logic for visibility

IF no dot is visible AND IF a new piece of content [ new notification in the notification block | new example / tutorial in the example / tutorial block ] is available AND the user is NOT on the Dashboard screen at the time, THEN display the dot.

ELSE IF the dot is visible AND IF the user clicks on the dashboard icon, THEN the dot disappears.

lstocchi:
### What does this PR do?

It adds a component that check if new content is available in the dashboard and show a purple dot accordingly.

dot_new_content

it closes #4043 

### How to test this PR?

1. open desktop
2. remove all podman machines so a new notification is pushed or uninstall/install an extension that adds a provider
3. you should see a dot next to the dashboard icon
4. it should be removed once you visit the dashboard


packages/renderer/src/AppNavigation.svelte
```diff
@@ -95,24 +96,29 @@ export let meta: TinroRouteMeta;
...
+              ></path
+              ></g
             ></g
-          ></g
-        ></svg>
+          ></svg>
+      </div>
+      <div class="absolute top-0 right-[-9px]">
```
deboer-tim: This should probably go in the component. Or like Tooltip, it could wrap the thing (default slot) it's adding the dot to?

packages/renderer/src/lib/dashboard/NewContentOnDashboardBadge.svelte
```diff
@@ -0,0 +1,71 @@
...
+let hasNew: boolean;
+$: hasNew = !isInDashBoard && (hasNewProviders || hasNewNotification);
+
+let providersUnsubscribe: Unsubscriber;
+let notificationsUnsubscribe: Unsubscriber;
+let routerUnsubscribe: Unsubscriber;
+
+onMount(() => {
+  // if there is a new provider we display the dot
+  providersUnsubscribe = providerInfos.subscribe(updatedProviders => {
```
deboer-tim, line 19: I don't want to turn this into a big componentization effort... but I'm pretty sure that we're going to want similar purple notification badges in other places too, e.g. if you pull an image the Images tab would have a badge until you go to it. We don't need to implement that here or make this a full component, but maybe we can extract the purple dot and cleanup based on router path from the dashboard-specific logic of when to add it and specific path?

deboer-tim (changes requested):

packages/renderer/src/lib/dashboard/NewContentOnDashboardBadge.svelte
lstocchi: I get your point, I also thought about it while implementing bc i wanted to create something customizable that could be reused but I didn't find any easy way to do it. The problem is that each route could have different and multiple ways to detect new content. And instead of adding multiple `ifs` - if dashboard subscribe to, if images subscribe to, ... - I thought that creating a custom component for each place could be easier to maintain in the future. So if we add new things to the dashboard and we want to change when the purple dot is visible, we only work on this.

The thing that multiple dot implementation can have in common is the router.subscribe and reset of the dot state, so i can create a baseComponent with it. WDYT?

deboer-tim: > The thing that multiple dot implementation can have in common is the router.subscribe and reset of the dot state, so i can create a baseComponent with it. WDYT?

Yeah, that's all I was thinking - a NotificationBadge component with 'I will disappear when user gets here' path property and route logic, outer dashboard component sets '/' and has logic of when to trigger. It's not much but stops us from replicating router code and UI in other instances.

packages/renderer/src/lib/dashboard/NewContentOnDashboardBadge.svelte
```diff
@@ -0,0 +1,63 @@
...
+}
+
+function onHide() {
+  hasNewProviders = false;
+  hasNewNotifications = false;
+}
+</script>
+
+<div class="absolute top-0 right-[-9px]">
+  <NewContentBadge pagePath="/" show="{hasNew}" onHide="{onHide}" />
```
deboer-tim, line 62: I'd prefer if there was a way to poke the badge when there was new content and it reset itself without the callback... but I already appreciate separating it out and think this is fine for now. Thanks for the changes!

deboer-tim (approved):

feloy (approved):

PR #4787, opened 2023-11-13
fix: report metrics for stopped machines

### LINKED ISSUE #4224: Missing CPU/disk/memory info on secondary machines
### Bug description

Issue #3111 changed the CPU/disk/memory for Podman machines to Donuts. This looks great on the primary machine, but when you have multiple machines the CPU/disk/memory which used to be grey are now missing. (not entirely sure that's due to #3111, but seems likely?) Not only would it be nice to still have this info on stopped machines, but the missing components lead to layout issues.

Screenshot 2023-10-05 at 3 40 46 PM

### Operating system

macOS 14.0

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #4224

### What does this PR do?

Reports machine info in donuts when machine is stopped

Fixes #4224

### How to test this PR?

1. Create a machine (do not ask to start it)
2. Go to the resources pages
3. There should be donuts for cpu, memory and disk

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -169,15 +169,19 @@ export async function updateMachines(provider: extensionApi.Provider): Promise<v
     const userModeNetworking = isWindows() ? machine.UserModeNetworking : true;
     podmanMachinesInfo.set(machine.Name, {
       name: machine.Name,
-      memory: machineInfo ? machineInfo?.memory : Number(machine.Memory),
-      cpus: machineInfo ? machineInfo?.cpus : machine.CPUs,
-      diskSize: machineInfo ? machineInfo?.diskSize : Number(machine.DiskSize),
+      memory: machineInfo ? machineInfo.memory : Number(machine.Memory),
+      cpus: machineInfo ? machineInfo.cpus : machine.CPUs,
+      diskSize: machineInfo ? machineInfo.diskSize : Number(machine.DiskSize),
       userModeNetworking: userModeNetworking,
-      cpuUsage: machineInfo?.cpuIdle ? 100 - machineInfo?.cpuIdle : 0,
+      cpuUsage: machineInfo?.cpuIdle !== undefined ? 100 - machineInfo?.cpuIdle : 0,
       diskUsage:
-        machineInfo?.diskUsed && machineInfo?.diskSize ? (machineInfo?.diskUsed * 100) / machineInfo?.diskSize : 0,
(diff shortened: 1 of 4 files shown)


extensions/podman/src/extension.ts
```diff
@@ -169,9 +169,9 @@ async function updateMachines(provider: extensionApi.Provider): Promise<void> {
     const userModeNetworking = isWindows() ? machine.UserModeNetworking : true;
     podmanMachinesInfo.set(machine.Name, {
       name: machine.Name,
-      memory: machineInfo?.memory,
-      cpus: machineInfo?.cpus,
-      diskSize: machineInfo?.diskSize,
+      memory: machineInfo ? machineInfo?.memory : Number(machine.Memory),
+      cpus: machineInfo ? machineInfo?.cpus : machine.CPUs,
+      diskSize: machineInfo ? machineInfo?.diskSize : Number(machine.DiskSize),
```
lstocchi: no need for the `?`

lstocchi (commented):
I tried and i only see the Disk size for a new created/not started machine.
is that correct or i should also see memory 0 cpu 0?
image

lstocchi:
Tried again, i guess i missed the last commit. Now i see all three values
image

lstocchi:
@jeffmaury i wonder if we can round the value. In my case 677.38 MB goes outside the circle but is it really necessary to know i have .38? Could we just display 677 MB? or round to the first decimal digit, 677.4MB?

jeffmaury:
> @jeffmaury i wonder if we can round the value. In my case 677.38 MB goes outside the circle but is it really necessary to know i have .38? Could we just display 677 MB? or round to the first decimal digit, 677.4MB?

Yes that maybe an option but I don't think this is related to this PR maybe we should have a look at the filesize utility that we use

lstocchi (approved):
Works fine for me. Just the above remark for the `?` operator

deboer-tim:
> > @jeffmaury i wonder if we can round the value. In my case 677.38 MB goes outside the circle but is it really necessary to know i have .38? Could we just display 677 MB? or round to the first decimal digit, 677.4MB?
> 
> Yes that maybe an option but I don't think this is related to this PR maybe we should have a look at the filesize utility that we use

100% agreed. Since it is a separate issue I opened #4802.

benoitf:

Issue #3234, opened 2023-07-14
UI Tweak: Consistent text inputs


mairin:
### Is your enhancement related to a problem? Please describe

There's some variants in how we use text inputs and we have some standalone ones. (For instance, right now when active on the text input on Desktop Extensions is a white border, whereas on the preferences search text input the active version has a purple border.) The color palette for these I don't think is 100% consistent either. 

### Describe the solution you'd like

Here is a spec for creating a consistent component for text inputs (including search text inputs.) 

https://design.penpot.app/#/view/8ec95363-4e2c-80a4-8002-cb67c874d0be?page-id=42e69059-b409-807e-8002-d0b98cda8446&section=interactions&index=0&share-id=c65f3168-697c-808c-8002-d0c926187cdf

Via the link above, make sure you click on the </> icon at the top/center of the page, and you can inspect individual elements like so:

![Screenshot from 2023-07-14 16-31-33](https://github.com/containers/podman-desktop/assets/799683/6a928e7b-509a-4602-8a69-e46b4f6ec703)

Posting screenshot of the spec above, but the penpot link above will always have the most updated/latest, these are a snapshot in time (updated 17 Jul 2023):

![Text Input](https://github.com/containers/podman-desktop/assets/799683/072c14ad-321f-4bd2-952b-f0cfb1908905)

deboer-tim:
Could you give an example of user entered text that is not-active/disabled, or confirm the color? Not clear to me if it should remain white, dim to the same level as default (placeholder) text, or stay somewhere between.

mairin:
> Could you give an example of user entered text that is not-active/disabled, or confirm the color? Not clear to me if it should remain white, dim to the same level as default (placeholder) text, or stay somewhere between.

Oh sure. I updated the screenshot in the issue description with color names (from top to bottom, (1) accent/border line, (2) text /icon color, (3) input background color).

Let me try to think through the case... so the user navigates to a text field:

![Screenshot from 2023-07-17 15-10-05](https://github.com/containers/podman-desktop/assets/799683/87d49fa2-0111-4a68-b218-43a93a2faaec)

They hover over it:
![Screenshot from 2023-07-17 15-10-37](https://github.com/containers/podman-desktop/assets/799683/fe8495a6-df98-4f44-900f-8cb48cad5cb3)

They click into it which focuses it (border color is slightly darker):
[...]

PR #4803, opened 2023-11-14
fix: fix background colours after patternfly removal

benoitf:
### What does this PR do?
background color was missing on some input
apply colors

you can see old look&feel in the referenced issue

how it looks in this PR:
image

fixes https://github.com/containers/podman-desktop/pull/4762#issuecomment-1810069867

### DIFF
--- a/packages/renderer/src/lib/preferences/item-formats/FileItem.svelte
+++ b/packages/renderer/src/lib/preferences/item-formats/FileItem.svelte
@@ -29,7 +29,7 @@ function handleCleanValue(
 
 <div class="w-full flex">
   <input
-    class="grow {!value ? 'mr-3' : ''} py-1 px-2 outline-0 text-sm placeholder-gray-900"
+    class="grow {!value ? 'mr-3' : ''} py-1 px-2 outline-0 text-sm placeholder-gray-900 bg-zinc-700"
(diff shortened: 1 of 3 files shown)


deboer-tim (approved):
I'm good with this.

Makes me wonder if there's an issue with the dark mode setting, but also that we will need to do Text component at some point (#3234).

benoitf:
so for now there is no "light/dark" mode in Podman Desktop so it takes default colour and assume it'll take light colours.
patternfly was set to dark

from now we'll be able to introduce context switching for colours and add specifics for light or dark

deboer-tim:
> so for now there is no "light/dark" mode in Podman Desktop so it takes default colour and assume it'll take light colours. patternfly was set to dark

My bad, I had seen this in the website and thought I was looking at the main tailwind cfg:
https://github.com/containers/podman-desktop/blob/cedd28b8b2b4cf9676a4e722d8dfff17a62809d0/website/tailwind.config.js#L6C1-L6C1

Re: https://tailwindcss.com/docs/dark-mode

> from now we'll be able to introduce context switching for colours and add specifics for light or dark

👍🏼. I think we should figure out how we're going to do the toggle and introduce a private feature flag for light mode soon, so that we can start exploring.

PR #4809, opened 2023-11-14
fix: reconnect to /events if disconnected

### LINKED ISSUE #4712: Connection to /events drops sometimes
### Bug description

Podman Desktop use events to refresh containers/images

But sometimes the `/events` connection is dropping

### Operating system

macOS

### Installation Method

None

### Version

1.5.0

### Steps to reproduce

long live running session / hivernate/suspend/restore computer

### Relevant log output

_No response_

### Additional context

If I do a curl on the endpoint, I receive a failure, so Podman Desktop should be able to handle the error/reconnect/notify that there is a problem

```
 curl --unix-socket /var/run/docker.sock "http://v1.47.0/events"

...
curl: (18) transfer closed with outstanding read data remaining
```

benoitf:
### What does this PR do?
When a provider is added, we establish a connection to `/events` REST API handler.
but if provider is not restarted and ping is still replying, we could have been disconnected from this stream.
It might happen if machine is restarting or some network issues.
In that case, re-establish the connection to the events so UI is properly refreshed. Because without events, we do not refresh the UI automatically

N/A

fixes https://github.com/containers/podman-desktop/issues/4712

### How to test this PR?

hard to test but I've added a unit test

you can for example to test, drop the update of the status in podman extension so you can do 'podman machine stop' wait and then 'podman machine start' and it'll reconnect
you should see mesages in the dev console that event connection has been dropped

### DIFF
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -16,6 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import type { Mock } from 'vitest';
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 import type { InternalContainerProvider } from '/@/plugin/container-registry.js';
 import { ContainerProviderRegistry } from '/@/plugin/container-registry.js';
@@ -143,6 +144,10 @@ class TestContainerProviderRegistry extends ContainerProviderRegistry {
 
   setStreamsPerContainerId(id: string, data: NodeJS.ReadWriteStream) {
     this.streamsPerContainerId.set(id, data);
+  }
+
+  setRetryDelayEvents(delay: number): void {
+    super.retryDelayEvents = delay;
   }
 }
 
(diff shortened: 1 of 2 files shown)


benoitf:

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -2210,3 +2215,87 @@ test('createNetwork', async () => {
...
+      Type: 'container',
+      id: fakeId,
+    }),
+  );
+  // check apiSender if we have a message 'container-started-event' with the right id
+  await new Promise(resolve => setTimeout(resolve, 1000));
+  expect(internalContainerProvider.api).toBeDefined();
+
+  // last call should be with the 'container-started-event' message
+  const allCalls = (apiSender.send as Mock).mock.calls;
```
dgolovin:
```suggestion
  const allCalls = vi.mocked(apiSender.send).mock.calls;
```

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -16,6 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import type { Mock } from 'vitest';
```
dgolovin:
```suggestion
```

dgolovin (approved):
nit: no need to import Mock.
LGTM

jeffmaury (approved):
LGTM

packages/main/src/plugin/container-registry.spec.ts
benoitf: thanks ! I always forget this mocked method !

fixed

Issue #4618, opened 2023-11-02
Kube Config UI: View current context


deboer-tim:
### Is your enhancement related to a problem? Please describe

We are imminently adding a list of Kube contexts to the Settings via #4560, but currently you can't see which context is the current one, and have no ability to change it.

### Describe the solution you'd like

The new tiles should show what the current context is, and there should be an action for selecting a different one. (yes, I could use the status bar context switcher too, but it's more natural to click on the one I can see in the list)

Switching the context via the editor or status bar should stay in sync, i.e. change in it one place and the other should update.

cdrage:
Split this up into two issues as I'll push this in two PRs:
- https://github.com/containers/podman-desktop/issues/4618 view which one is current context
- https://github.com/containers/podman-desktop/issues/4920 add the button

Reason being is that we need a quick UI mockup on where we should put the button as we've changed it to use a "trash can" for deleting, but unsure for selection (should we just select any area?).

PR #4812, opened 2023-11-14
feat: delete kube context

### LINKED ISSUE #4417: Kube context delete

As Kube contexts are going to be first class citizen in Podman Desktop and displayed nicely, we should be able to delete them as Kubeconfig may have a lot of obsoleted contexts

On the Kubecontext UI we should be able to see a delete button/icon

_No response_

### Additional context

_No response_

feloy:

#### Settings > Kubernetes

This PR adds a Trash button for each item in the list. 
Clicking the button deletes the context from the list and from the kubeconfig file.

/!\ Any user or cluster which were referenced by the context and not referenced anymore in the kubeconfig file is also removed from the file.

https://github.com/containers/podman-desktop/assets/9973512/7f90f800-2791-4d62-9732-6b0fe0e8ae5f

TODO
- [ ] do we want to ask for confirmation before to delete the context from the file?
- [ ] how do we want to display an error when one occurs (for example if the kubeconfig file is not writable)?
- [ ] do we want to update the current-context?
- [ ] unit tests

Fixes #4417 

### How to test this PR?

First make a backup copy of your kubeconfig file, as it will be modified by the test.

Press the Trash button on any of the entries.

### DIFF
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -274,12 +274,9 @@ export class KubernetesClient {
       users: this.kubeConfig.users.filter(user => newContexts.some(ctx => ctx.user === user.name)),
       currentContext: this.kubeConfig.currentContext,
     });
-    try {
-      await this.saveKubeConfig(newConfig);
-      this.kubeConfig = newConfig;
-    } catch {
-      return this.getContexts();
-    }
+    await this.saveKubeConfig(newConfig);
+    // the config is saved back only if saving the file succeeds
+    this.kubeConfig = newConfig;
     return this.getContexts();
   }
 
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -104,7 +104,7 @@ const DEFAULT_NAMESPACE = 'default';
  * Handle calls to kubernetes API
  */
 export class KubernetesClient {
-  private kubeConfig;
+  protected kubeConfig;
 
   private static readonly DEFAULT_KUBECONFIG_PATH = resolve(homedir(), '.kube', 'config');
 
@@ -112,7 +112,7 @@ export class KubernetesClient {
   private kubeconfigPath: string = KubernetesClient.DEFAULT_KUBECONFIG_PATH;
 
   protected currentNamespace: string | undefined;
-  private currentContextName: string | undefined;
+  protected currentContextName: string | undefined;
 
   private kubeConfigWatcher: containerDesktopAPI.FileSystemWatcher | undefined;
 
@@ -266,12 +266,22 @@ export class KubernetesClient {
   }
 
   async deleteContext(contextName: string): Promise<Context[]> {
+    const previousContexts = this.kubeConfig.contexts;
     const newContexts = this.kubeConfig.contexts.filter(ctx => ctx.name !== contextName);
     const newConfig = new KubeConfig();
     newConfig.loadFromOptions({
       contexts: newContexts,
-      clusters: this.kubeConfig.clusters.filter(cluster => newContexts.some(ctx => ctx.cluster === cluster.name)),
-      users: this.kubeConfig.users.filter(user => newContexts.some(ctx => ctx.user === user.name)),
+      clusters: this.kubeConfig.clusters.filter(cluster => {
+        // remove clusters not referenced anymore, except if there were already not referenced before
+        return (
+          newContexts.some(ctx => ctx.cluster === cluster.name) ||
+          !previousContexts.some(ctx => ctx.cluster === cluster.name)
+        );
+      }),
+      users: this.kubeConfig.users.filter(user => {
+        // remove users not referenced anymore, except if there were already not referenced before
+        return newContexts.some(ctx => ctx.user === user.name) || !previousContexts.some(ctx => ctx.user === user.name);
+      }),
       currentContext: this.kubeConfig.currentContext,
     });
     await this.saveKubeConfig(newConfig);
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -55,6 +55,7 @@ import { PassThrough } from 'node:stream';
 import type { ApiSenderType } from './api.js';
 import { parseAllDocuments } from 'yaml';
 import type { Telemetry } from '/@/plugin/telemetry/telemetry.js';
+import * as jsYaml from 'js-yaml';
 
 function toContainerStatus(state: V1ContainerState | undefined): string {
   if (state) {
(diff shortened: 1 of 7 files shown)


afbjorklund:
After deleting the kubernetes context, the current-context still references it.

```
apiVersion: v1
kind: Config
clusters: []
users: []
contexts: []
preferences: {}
current-context: kubernetes-admin@kubernetes
```

After that, `kubectl` starts throwing errors: (and you can't really "deselect" it, in GUI)

`invalid configuration: [context was not found for specified context: kubernetes-admin@kubernetes, cluster has no server defined]`


Have to use the terminal, to fix it:

```
kubectl config unset current-context
```

But `delete-context` does the same thing: (only difference is that it warns you that it got broken)

```
kubectl config delete-context kubernetes-admin@kubernetes
```

`warning: this removed your active context, use "kubectl config use-context" to select a different one`

feloy:
Based on @deboer-tim and @afbjorklund feedback, we could block the deletion for the current context (but we will need to implement before the change of current context from the GUI), except when this is the latest context in the file: in this case we would unet the current context.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -264,6 +265,30 @@ export class KubernetesClient {
...
   }
 
+  async deleteContext(contextName: string): Promise<Context[]> {
+    const newContexts = this.kubeConfig.contexts.filter(ctx => ctx.name !== contextName);
+    const newConfig = new KubeConfig();
+    newConfig.loadFromOptions({
+      contexts: newContexts,
+      clusters: this.kubeConfig.clusters.filter(cluster => newContexts.some(ctx => ctx.cluster === cluster.name)),
+      users: this.kubeConfig.users.filter(user => newContexts.some(ctx => ctx.user === user.name)),
+      currentContext: this.kubeConfig.currentContext,
```
jeffmaury, line 285: What if the current context if deleted

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -6,19 +6,32 @@ import { onMount } from 'svelte';
...
+  k8sClusters = await window.kubernetesGetClusters();
 
   // Convert them to KubeContextUI so we can safely render them.
   contexts = getKubeUIContexts(k8sContexts, k8sClusters);
 });
+
+async function handleDeleteContext(contextName: string) {
+  try {
+    const newK8sContexts = await window.kubernetesDeleteContext(contextName);
+    contexts = getKubeUIContexts(newK8sContexts, k8sClusters);
```
jeffmaury, line 45: That should not be necessary once store is there, don't know how to handle it once it's been merged

benoitf:
hello, I'm wondering why I could not remove the current context ?

let say my context is an old kind cluster that I have deleted, why should not be able to get rid of it ?

benoitf:

deboer-tim:
> hello, I'm wondering why I could not remove the current context ?
> 
> let say my context is an old kind cluster that I have deleted, why should not be able to get rid of it ?

We talked about #4417 on a call yesterday but maybe we weren't thinking about some cases. My recommendation would be a confirmation dialog if you're deleting the current context, but no confirmation dialog otherwise. (i.e. you can delete any context, you just get prompted before current because it's 'in use'/special, and I'm assuming we would unset it).

We don't have the ability to see or change the current context yet (issue #4618 later this sprint) so this part is a little cart-horse.

feloy:
I have made the changes according to the recommendations from @deboer-tim, by using a confirmation dialog before to delete the current context.

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -4,21 +4,51 @@ import EngineIcon from '../ui/EngineIcon.svelte';
...
+
+async function handleDeleteContext(contextName: string) {
+  if (currentContextName === contextName) {
+    const result = await window.showMessageBox({
+      title: 'Delete Context',
+      message:
+        'You will delete the current context. If you delete it, you will need to switch to another context. Continue?',
+      buttons: ['Yes', 'No'],
+    });
+    if (!result || result.response !== 0) {
```
jeffmaury:
```suggestion
    if (result?.response !== 0) {
```

jeffmaury: Not sure why we check for !result given the method signature

jeffmaury (commented):
Maybe we should have a unit test on the kubernetes client ?

packages/main/src/plugin/kubernetes-client.ts
feloy: This is my question above (`do we want to update the current-context?`). 

As it is now, it is the same behaviour as kubectl:
```
$ kubectl config current-context 
my-current-context
$ kubectl config delete-context my-current-context
warning: this removed your active context, use "kubectl config use-context" to select a different one
deleted context my-current-context from /home/phmartin/.kube/config
$ kubectl config current-context 
my-current-context
```

Do we want to keep this same behaviour, or have a different one?

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: Yes, this will change once #4745 is merged. I would prefer to wait for #4745 to be merged before to continue on this PR, so I can handle this accordingly.

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: yes, right, it should not be necessary. Thanks

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
jeffmaury: I think that would be the rebase of #4745 to remove what needs to be removed and this PR should not be blocked

packages/main/src/plugin/kubernetes-client.ts
jeffmaury: Ignore that was kubectl behavior, thus this is ok for me

jeffmaury (approved):
LGTM

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: When using the store, I think we won't have to get the contexts as result of ` window.kubernetesDeleteContext` (as it will come from the store), but we will only need to catch the error. But we will probably need to have a separate variable holding the errors, something like:

```
try {
    await window.kubernetesDeleteContext(contextName);
  } catch (e: unknown) {
    if (e instanceof Error) {
+      errors[contextName] = e;
-      contexts = setKubeUIContextError(contexts, contextName, e);
    }
  }
```

deboer-tim (approved):
I guess I hadn't manually cleaned up my contexts for a month, this was incredibly useful and far easier/faster than command line or editing. :)

Confirmation before deleting my current context worked well.

PR #4825, opened 2023-11-15
fix: use URL for proxy specification and add validation

### LINKED ISSUE #4810: Allow URLs for proxy specification and implement validation

Today proxies are specified usin hostname:port which is quite limiting and values are not validating

In addition to hostname:port we should also support URLS and perform validations

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #4810

### What does this PR do?

Ensure proxy addresses are now URLs; convert previous values and add validation in the settings screen

Fixes #4810

### How to test this PR?

Go to the proxy settings page

### DIFF
--- a/packages/main/src/plugin/proxy.ts
+++ b/packages/main/src/plugin/proxy.ts
@@ -22,7 +22,7 @@ import { Emitter } from './events/emitter.js';
 import { getProxyUrl } from './proxy-resolver.js';
 import { ProxyAgent } from 'undici';
 
-function ensureURL(urlstring: string | undefined): string | undefined {
+export function ensureURL(urlstring: string | undefined): string | undefined {
   if (urlstring) {
     try {
       const url = new URL(urlstring);
--- a/packages/main/src/plugin/proxy.ts
+++ b/packages/main/src/plugin/proxy.ts
@@ -22,6 +22,30 @@ import { Emitter } from './events/emitter.js';
 import { getProxyUrl } from './proxy-resolver.js';
 import { ProxyAgent } from 'undici';
 
+function ensureURL(urlstring: string | undefined): string | undefined {
+  if (urlstring) {
+    try {
+      const url = new URL(urlstring);
+      if (url.hostname) {
+        return urlstring;
+      }
+    } catch (err) {
+      /* empty */
+    }
+    return `http://${urlstring}`;
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function asURL(url: any): URL {
+  if (url instanceof URL) {
+    return url;
+  } else if (typeof url === 'string') {
+    return new URL(url);
+  }
+  return new URL((url as Request).url);
+}
+
 /**
  * Handle proxy settings for Podman Desktop
  */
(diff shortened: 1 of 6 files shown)


packages/main/src/plugin/proxy.ts
```diff
@@ -22,6 +22,30 @@ import { Emitter } from './events/emitter.js';
 import { getProxyUrl } from './proxy-resolver.js';
 import { ProxyAgent } from 'undici';
 
+function ensureURL(urlstring: string | undefined): string | undefined {
```
lstocchi: Missing some tests?

packages/main/src/plugin/proxy.ts
```diff
@@ -22,6 +22,30 @@ import { Emitter } from './events/emitter.js';
...
+      }
+    } catch (err) {
+      /* empty */
+    }
+    return `http://${urlstring}`;
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function asURL(url: any): URL {
```
lstocchi, line 40: It looks to me that url can only be URL, string or Request, no? Why using any?

jeffmaury: > It looks to me that url can only be URL, string or Request, no? Why using any?

It's related to the signature of the `fetch` method a few lines below

lstocchi: but why we use any? we are overwriting the fetch function which accept only those three types 
image

lstocchi: it's not a change made by this PR, just noticed it now but i think we can remove the any type

jeffmaury: > it's not a change made by this PR, just noticed it now but i think we can remove the any type

any was introduced because of the recent changes in Node types library

lstocchi: mmm, saw it. The problem is that we update the types/node without updating the minimum node version and it creates discrepancy.
So, nevermind, nothing we can do here

packages/main/src/plugin/proxy.ts
```diff
@@ -22,6 +22,30 @@ import { Emitter } from './events/emitter.js';
 import { getProxyUrl } from './proxy-resolver.js';
 import { ProxyAgent } from 'undici';
 
+function ensureURL(urlstring: string | undefined): string | undefined {
+  if (urlstring) {
+    try {
+      const url = new URL(urlstring);
+      if (url.hostname) {
```
dgolovin, line 29: What `urlstring' should be passed to URL constructor, to get falsy `url.hostname`?
It always has `hostname` if URL is created out of string or it throws exception.

packages/renderer/src/lib/preferences/Util.ts
```diff
@@ -145,3 +145,16 @@ export function uncertainStringToNumber(value: string | number): number {
   if (typeof value === 'number') return value;
   return parseFloat(value);
 }
+
+export function validateProxyAddress(value: string): string | undefined {
+  if (value) {
+    try {
+      const u = new URL(value);
+      if (!u.hostname) {
```
dgolovin, line 153: Is there a test that gets inside this if block?

dgolovin (changes requested):

packages/main/src/plugin/proxy.ts
jeffmaury: Yes but hostname is then empty so section is executed

packages/main/src/plugin/proxy.ts
jeffmaury: Added

packages/main/src/plugin/proxy.ts
jeffmaury: Added unit tests to check

packages/renderer/src/lib/preferences/Util.ts
jeffmaury: Just added

lstocchi (approved):
LGTM

dgolovin (approved):

PR #4826, opened 2023-11-15
fix: light scrollbars and line height

deboer-tim:
### What does this PR do?

After the PatternFly removal I noticed that the scrollbars on Mac are in light-mode. PF-dark was setting the color-scheme before, setting it again fixes the problem and means we can also remove the explicit text-white. (maybe some of the changes in #4803 too?)

PF was also setting the line height to 0.9. I'm not tied to it, but it makes buttons, multi-line text, and preferences nav items all less tall. Maybe I'm just used to it, but I prefer a smaller window and 0.95 gets most of the way there & uses less space.

1.5.3:
Screenshot 2023-11-15 at 6 55 43 PM

Post PatternFly:
Screenshot 2023-11-15 at 4 15 15 PM

This PR:
Screenshot 2023-11-15 at 4 14 40 PM

N/A

### How to test this PR?

Open dashboard, preferences.

### DIFF
--- a/packages/renderer/index.html
+++ b/packages/renderer/index.html
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, height=device-height, initial-scale=1.0" />
     <title>Podman Desktop</title>
   </head>
-  <body class="overflow-hidden text-white">
+  <body class="overflow-hidden">
     <div id="app"></div>
     <script type="module" src="/src/main.ts"></script>
   </body>
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -1,3 +1,10 @@
+<style>
+:root {
+  color-scheme: dark;
+  line-height: 0.95rem;
+}
+</style>
+
 <script lang="ts">
 import './app.css';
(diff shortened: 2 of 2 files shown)


jeffmaury (approved):
LGTM on Windows

benoitf (changes requested):
I would not opt-in for this change (color-scheme), it will delay the work to have proper light/dark theme.

For the line-height, I would not follow css approach as well, it's breaking the tailwind flow
as elements are interconnected. It's making harder to have clean stuff.

deboer-tim:
color-scheme sets the defaults (light=white background, black text, scrollbars colors, etc, dark=black background, white text, etc). I was assuming we'd set it now to fix scrollbars (and anything else like text-white where the default is fine), then future light mode would toggle it. What is the other option? I don't think we want to custom style scrollbars or to use `text-black dark:text-white` in the future when those could already be the default colors.

AFAIK Tailwind doesn't have an equivalent option to globally scale line height. Do you think we should override all the font sizes in config, change scaling/padding in all the affected components individually, or something else?

deboer-tim:
Light mode fixed via #4910. I will submit separate issues and/or PRs for components affected by change in line height.

PR #4827, opened 2023-11-16
fix: merged compose deploy to kube page in UI

### LINKED ISSUE #4723: Pressing "Deploy to Kubernetes" on a Compose example adds extra section to bottom
### Bug description

https://github.com/containers/podman-desktop/assets/6422176/72decace-759b-4b98-ae4e-25dff17e593a

Toolbar section being shown for some reason at the bottom of the screen.

### Operating system

mac m1

### Installation Method

None

### Version

1.5.0

### Steps to reproduce

Deploy any Compose example:

```yaml
services:

  redis-master:
    #container_name: redis-master
    image: redis:latest
    ports:
      - "6379"

  redis-replica:
    #container_name: redis-replica
    image: redis:latest
    ports:
      - "6379"
    command: redis-server --replicaof redis-master 6379
    depends_on:
      - redis-master

  frontend:
    #container_name: frontend
    image: cdrage/frontend
    #build: ./frontend
    ports:
      - "8080:8080"
```

Click on `Deploy to Kubernetes`

Scroll to the bottom.

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

When used on a compose app, the Deploy to Kube form incorrectly shows the compose details at the bottom of the page. You can also see it in the breadcrumb that says 'Container > Compose Details' when it should say 'Container > Deploy to Kubernetes'.

The problem is the route path of these two pages can overlap:
 - /compose/deploy-to-kube/:composeGroupName/:engineId/*
 - /compose/:name/:engineId/* i.e. deploy to kube path also resolves to compose details page with name 'deploy-to-kube' and engineId of the compose group name.

Simple fix to just rename the path so they don't overlap.

Same as before, but with correct breadcrumb and no details page at the bottom!

Fixes #4723.

### How to test this PR?

Create a compose app and use the Deploy to Kubernetes action on the Containers page.

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -132,13 +132,13 @@ window.events?.receive('display-troubleshooting', () => {
             type="container" />
         </Route>
         <!-- Same DeployPodToKube route, but instead we pass in the compose group name, then redirect to DeployPodToKube -->
-        <Route path="/compose-deploy-to-kube/:composeGroupName/:engineId/*" breadcrumb="Deploy to Kubernetes" let:meta>
+        <Route path="/compose/deploy-to-kube/:composeGroupName/:engineId/*" breadcrumb="Deploy to Kubernetes" let:meta>
           <DeployPodToKube
             resourceId="{decodeURI(meta.params.composeGroupName)}"
             engineId="{decodeURI(meta.params.engineId)}"
             type="compose" />
         </Route>
-        <Route path="/compose/:name/:engineId/*" breadcrumb="Compose Details" let:meta navigationHint="details">
+        <Route path="/compose/details/:name/:engineId/*" breadcrumb="Compose Details" let:meta navigationHint="details">
           <ComposeDetails composeName="{decodeURI(meta.params.name)}" engineId="{decodeURI(meta.params.engineId)}" />
         </Route>
         <Route path="/pods/:kind/:name/:engineId/*" breadcrumb="Pod Details" let:meta navigationHint="details">
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -132,7 +132,7 @@ window.events?.receive('display-troubleshooting', () => {
             type="container" />
         </Route>
         <!-- Same DeployPodToKube route, but instead we pass in the compose group name, then redirect to DeployPodToKube -->
-        <Route path="/compose/deploy-to-kube/:composeGroupName/:engineId/*" breadcrumb="Deploy to Kubernetes" let:meta>
+        <Route path="/compose-deploy-to-kube/:composeGroupName/:engineId/*" breadcrumb="Deploy to Kubernetes" let:meta>
           <DeployPodToKube
             resourceId="{decodeURI(meta.params.composeGroupName)}"
             engineId="{decodeURI(meta.params.engineId)}"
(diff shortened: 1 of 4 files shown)


packages/renderer/src/App.svelte
```diff
@@ -132,7 +132,7 @@ window.events?.receive('display-troubleshooting', () => {
             type="container" />
         </Route>
         <!-- Same DeployPodToKube route, but instead we pass in the compose group name, then redirect to DeployPodToKube -->
-        <Route path="/compose/deploy-to-kube/:composeGroupName/:engineId/*" breadcrumb="Deploy to Kubernetes" let:meta>
+        <Route path="/compose-deploy-to-kube/:composeGroupName/:engineId/*" breadcrumb="Deploy to Kubernetes" let:meta>
```
benoitf: I would keep the `/compose` namespace to all the compose routing and use other differentiators instead of having separate routes at the beginning
-> change routing scheme after `/compose` rather than at the root level

if we have common prefix we can route or use components for the routing (I think @feloy wanted to introduce routing scheme/helpers at some point) but if we break at the root level, it'll be harder to introduce Routing components per scope

deboer-tim: Ok, second commit reverts the deploy-to-kube path and changes the details path instead. I just don't love this option because it makes the compose details different from image/volumes/pods/etc details instead, but we have to change one of them.

benoitf: we could change the route for image/volumes/pods/etc to be consistent 👍

benoitf: I liked the approach of @feloy to have helpers for the routes so we don't duplicate everywhere the route path

benoitf: I found the related PR https://github.com/containers/podman-desktop/pull/4371

benoitf:
is it missing details there: https://github.com/containers/podman-desktop/blob/a9e69b6e0af750a0d71b2169fbb96e43cd81289a/packages/renderer/src/lib/compose/ComposeDetailsLogs.svelte#L38 ?

packages/renderer/src/App.svelte
feloy: I finally reverted it in this PR, as I didn't have enough perspective at this time. But it is visible in the first commit of the PR

benoitf:
@deboer-tim 

ping about the last comment
isn't it missing a change there https://github.com/containers/podman-desktop/blob/a9e69b6e0af750a0d71b2169fbb96e43cd81289a/packages/renderer/src/lib/compose/ComposeDetailsLogs.svelte#L38 ?

deboer-tim:
Sorry, was distracted by other PRs. Yes, I missed that reference. Added in new commit, and I did a scan again to confirm there are no further references.

cdrage (commented):
LGTM!

Confirmed it no longer shows.

https://github.com/containers/podman-desktop/assets/6422176/1869518c-7e87-4204-bd3e-3650437337e6

cdrage (approved):

packages/renderer/src/App.svelte
deboer-tim: 100% agreed on the solution, but I don't think we should hold up a minor fix to solve the broader problem. I've opened #4966 to track solving this based on what was in #4371.

PR #4841, opened 2023-11-16
Volume table component

### LINKED ISSUE #4839: Move Volumes page to table component

We have a new table component, Volumes page should switch.

Volumes page should be based on the new table component, and most columns should be sortable.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Switches the Volume page to the new table component. Columns are extracted as expected, and I've pulled out the engine into an Environment column to match the Pods page.

I made all the columns sortable. The raw size had to be made visible from VolumeInfoUI because the regular column is humanized and not sortable. Status and size are 'reverse sorted' because you typically want to see used volumes and bigger things first.

Before:

Screenshot 2023-11-16 at 2 51 56 PM

After:

Screenshot 2023-11-16 at 1 44 12 PM

Fixes #4839.

### How to test this PR?

Use the Volumes page as normal. Try sorting by each column.


deboer-tim:
Rebased on main.

feloy:
The order seems reverted, at least on Name and Age columns.

When a `^` is displayed, I expect to have lowest values first.

sort-age
sort-name

feloy:
With content larger than the cell, the content is overlapping through the other columns.

column-overlap

deboer-tim:
I've set a max width so that table cells do not overlap anymore; they will be cropped. Now that we're not based on html table I think there is a lot more that we could do here in the future (e.g. correctly use '...' in the cell, or setting minimum widths), but this initial change fixes the ugliness.

The sorting direction felt natural to me at the time, but you are correct. However, it is mostly independent to this PR, the image PR, and the PR for sort indicators, and I'm starting to have trouble with all of them touching the same file. :) Are you ok splitting it off into a separate issue and fixing it 'later'? I've opened #4881 to track this, and would be able to PR quickly as soon as one or two of these others get merged.

mairin:
copy/pasta from dev chat, re: the arrows and ordering:

ooh this is a perennial problem 🙂 does the arrow (or any other icon that also can be read as status) mark the action to be taken by clicking on it, or the current state of things. I would argue in this case, the up arrow marks the action to be taken, and you can make that clear by keeping the downward facing error but greying it out (it's disabled since you can't sort the other way as its already sorted that way)

deboer-tim:
I’ve tried to cover improved sorting for the component in #4881 so that this can focus just on moving to the component for the Volume page.

packages/renderer/src/lib/volume/VolumeColumnAge.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+import type { VolumeInfoUI } from './VolumeInfoUI';
+
+export let object: VolumeInfoUI;
```
feloy, line 4: Could we pass the value only (age), so we could have a more generic and reusable AgeColumn, or DurationColumn, or NumberColumn?

packages/renderer/src/lib/volume/VolumeColumnEnvironment.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+import ProviderInfo from '../ui/ProviderInfo.svelte';
+import type { VolumeInfoUI } from './VolumeInfoUI';
+
+export let object: VolumeInfoUI;
```
feloy, line 5: Same here, could we pass the useful info only (engineName, engineId), and have a generic and resusable ProviderColumn?

packages/renderer/src/lib/volume/VolumeColumnName.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
```
feloy, line 1: Here, we could have a ClickableStringColumn? Passing the url and string value

packages/renderer/src/lib/volume/VolumeColumnAge.svelte
deboer-tim: I would like to be able to do this, and share code, but I am worried about the scope and co-dependencies of these PRs and how much we continue to refine before anything is usable and more widely available in main.

To me we always need to support the current behaviour (since columns like name are often custom or include multiple properties) and we could support this later through something like a column type mapping function, e.g. from:
```
  renderer: ContainerColumnAge
```
to:
```
  property: container => container.age,
  renderer: AgeColumn
```

packages/renderer/src/lib/volume/VolumeColumnEnvironment.svelte
deboer-tim: Same as above.

packages/renderer/src/lib/volume/VolumeColumnName.svelte
deboer-tim: Yes, but again I think this is a feature we could add later. We have a bug for the fact that you can only click on one column today, we need to decide how you get to details vs other table actions.

feloy (approved):
LGTM

packages/renderer/src/lib/volume/VolumeColumnActions.svelte
```diff
@@ -0,0 +1,8 @@
+<script lang="ts">
+import VolumeActions from './VolumeActions.svelte';
+import type { VolumeInfoUI } from './VolumeInfoUI';
+
+export let object: VolumeInfoUI;
```
jeffmaury, line 5: Not sure object is well choosen

jeffmaury (approved):
LGTM

packages/renderer/src/lib/volume/VolumeColumnActions.svelte
deboer-tim: I would rather not change it atm since it spans at least 3 PRs ;-), but if you can think of a better generic property name I'd be glad to PR it early next week.

Issue #3875, opened 2023-09-12
Cont'd of improving the accessibility of Main pages content (Images, containers) in podman desktop


odockal:
### Is your enhancement related to a problem? Please describe

When writing e2e tests, I have noticed that sometimes it was very hard to get specific element. For example html table with images was hard to handle. Adding some aria - based features could improve things to be more readable and handle by playwright as well.

Improve areas (looking at html side) in Image/Container/etc. pages' main content - table of images, containers, pods.

### Describe the solution you'd like

Use accessibility functionality (ARIA roles and attributes) to better describe the table rows and columns. 

We are interested in easy identification of the image/container/pods/etc row in the table. So it is easy to locate immediately. Also, the row contains specific information we would like to easily identify and retrieve.

use appropriate ARIA roles and attributes. 
![MainPageContentTableStructure](https://github.com/containers/podman-desktop/assets/19164299/0a273635-f813-4fa7-ad89-1e68e3bb495e)

### Additional context

Follow up to: https://github.com/containers/podman-desktop/issues/3338.

Issue #3109, opened 2023-07-04
When clicking to see details of container, I expect entire row to initialize details.


cdrage:
### Is your enhancement related to a problem? Please describe

Whenever I am trying to access the details of a container / pod I explicitly have to click on the name.

I expect that the *entire* row be clickable except for the obvious buttons (start, stop, restart, etc.)

Right now I have to hover over the name / sometimes image?

### Describe the solution you'd like

Being able to click anywhere in the row to be able to access the details of the container.

### Additional context

https://github.com/containers/podman-desktop/assets/6422176/d8778640-cdcb-45e8-9b19-491e4fd9a02f

cdrage:
Same thing happens on the pod page too

PR #4844, opened 2023-11-16
Image table component

### LINKED ISSUE #4840: Move Images page to table component

We have a new table component, Images page should switch.

The Images page should be based on the new table component, and most columns should be sortable.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Switches the Images page to the new table component. Columns are extracted as expected, and I've pulled out the engine into an Environment column to match the Pods page.

I made all the columns sortable. The raw size had to be made visible from ImageInfoUI because the regular column is humanized and not sortable. Status and size are 'reverse sorted' because you typically want to see used images and bigger ones first.

Before:

Screenshot 2023-11-16 at 2 52 08 PM

After:

Screenshot 2023-11-16 at 1 44 04 PM

Fixes #4840.

### How to test this PR?

Use Images page as normal, try sorting.


deboer-tim:
Rebased on main, fixed one test failure and found another (forgot to set role on first two cells in table) but Playwright is still await-ing indefinitely looking for the table row cells that I can see onscreen. May need some help next week to understand why e2e is failing.

feloy:
> Rebased on main, fixed one test failure and found another (forgot to set role on first two cells in table) but Playwright is still await-ing indefinitely looking for the table row cells that I can see onscreen. May need some help next week to understand why e2e is failing.

As there is no `role="cell"` for the first row (the header), the `zeroCell` and `thirdCell` are not found in the first row. By skipping the first row, it works:

```
diff --git a/tests/src/model/pages/images-page.ts b/tests/src/model/pages/images-page.ts
index 4d2003617..e0b0a7753 100644
--- a/tests/src/model/pages/images-page.ts
+++ b/tests/src/model/pages/images-page.ts
@@ -58,7 +58,12 @@ export class ImagesPage extends MainPage {
     }
     const table = await this.getTable();
     const rows = await table.getByRole('row').all();
+    let i=-1;
     for (const row of rows) {
+      i++;
+      if (i==0) {
+        continue;
+      }
       // test on empty row - contains on 0th position &nbsp; character (ISO 8859-1 character set: 160)
       const zeroCell = await row.getByRole('cell').nth(0).innerText();
       if (zeroCell.indexOf(String.fromCharCode(160)) === 0) {
```

deboer-tim:
Ahh, that makes total sense. Thank you @feloy!

Based on this [1] I think the current behaviour is now correct, i.e. I've accidentally 'fixed' the table and the correct fix is to the test, so I've made a similar change as you proposed. The test is now passing.

It still failed elsewhere, but since it isn't failing there locally and doesn't seem to be related to the change I've rerun the test.

[1] https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/table/

feloy:
> Ahh, that makes total sense. Thank you @feloy!
> 
> Based on this [1] I think the current behaviour is now correct, i.e. I've accidentally 'fixed' the table and the correct fix is to the test, so I've made a similar change as you proposed. The test is now passing.
> 
> It still failed elsewhere, but since it isn't failing there locally and doesn't seem to be related to the change I've rerun the test.
> 
> [1] https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/table/

I can see the tests fail locally when I open the developer tools part and increase it, the same size as in the CI, see video artifact below (problem starting at 0:11): 

[622b2819432213f69980b45949f60570.webm](https://github.com/containers/podman-desktop/assets/9973512/f09cb554-631d-4a86-8b20-b965a9b35f16)

But I cannot figure out for now why it fails

odockal:
@deboer-tim @feloy I would like to mention #3875 which should aim the issue of the ARIA functionality integration and its use in e2e tests.

odockal:
@deboer-tim I am really looking into redo table components, let's also consider broader context. I will take a look at PR and test it. (I haven't yet)

deboer-tim:
@odockal I am hoping that the component can give some consistency between the pages, and maybe there is something general we can do to make the particular cells easier to identity (e.g. each cell have an aria label with the column name so we're not doing 'nth cell'?).

Short term we need to get pages using the component though, and I could use some help figuring out why the e2e test is failing here :).

cdrage (changes requested):
The code looks good to me, but I'm getting this odd issue I noticed, I'm unsure if this is related to the PR however:


https://github.com/containers/podman-desktop/assets/6422176/8b8f26d5-48e3-4c5d-97cd-cac3b410fd25

When I first load the images page, I have to click the Age twice to sort, for some reason? I realize it's already sorted by age, but seems odd I have to press it twice

deboer-tim:
> The code looks good to me, but I'm getting this odd issue I noticed, I'm unsure if this is related to the PR however:
> 
> When I first load the images page, I have to click the Age twice to sort, for some reason? I realize it's already sorted by age, but seems odd I have to press it twice

This was fixed in #4860 by adding a 'defaultSortColumn' to tell the table component what column the data is already sorted by, then ImageList table just needs `defaultSortColumn="Age"`. Now that #4860 is merged I can rebase and add that.

deboer-tim:
Added default sort column (age).

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -143,9 +143,9 @@ function setGridColumns() {
       class="grid grid-table gap-x-0.5 mx-5 h-12 bg-charcoal-800 hover:bg-zinc-700 rounded-lg mb-2"
       animate:flip="{{ duration: 300 }}"
       role="row">
-      <div class="whitespace-nowrap justify-self-start"></div>
+      <div class="whitespace-nowrap justify-self-start" role="cell"></div>
       {#if row.info.selectable}
-        <div class="whitespace-nowrap place-self-center">
+        <div class="whitespace-nowrap place-self-center" role="cell">
```
odockal: It would be really helpful, if every cell element, would have associated `aria-label` that would help to identify what kind of information/object does this particular cell contains.

In this case: `role="cell" aria-label="Toggle image"`
Although this behaviour would be more visible in other cells.

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -143,9 +143,9 @@ function setGridColumns() {
       class="grid grid-table gap-x-0.5 mx-5 h-12 bg-charcoal-800 hover:bg-zinc-700 rounded-lg mb-2"
       animate:flip="{{ duration: 300 }}"
       role="row">
```
odockal: Adding a `aria-label="{object.name}"` into a row element would be also super helpful.

deboer-tim:
Rebased to pick up #4969 as we believe that was the cause of the e2e test issue (thanks @odockal, excellent debugging !).

#4967 opened to improve aria roles for tables. I believe all of the improvements are either things that can be done in the component (for all tables) or general improvements we should do to every table, either way unrelated to this PR (moving Image page to table component) and would like to handle separately.

@cdrage please re-review since your issue was fixed (by support in the table component and adding `defaultSortColumn="Age"`).

cdrage (approved):

cdrage:
> Rebased to pick up #4969 as we believe that was the cause of the e2e test issue (thanks @odockal, excellent debugging !).
> 
> #4967 opened to improve aria roles for tables. I believe all of the improvements are either things that can be done in the component (for all tables) or general improvements we should do to every table, either way unrelated to this PR (moving Image page to table component) and would like to handle separately.
> 
> @cdrage please re-review since your issue was fixed (by support in the table component and adding `defaultSortColumn="Age"`).

Thank you! Tested it and I see it's fixed / it's sorted by default now. Code LGTM.

benoitf (commented):
I see no component test for most of the .svelte component ?

packages/renderer/src/lib/image/ImageColumnName.svelte
```diff
@@ -0,0 +1,20 @@
...
+
+export let object: ImageInfoUI;
+
+function openDetailsImage(image: ImageInfoUI) {
+  router.goto(`/images/${image.id}/${image.engineId}/${image.base64RepoTag}/summary`);
+}
+</script>
+
+<!-- svelte-ignore a11y-click-events-have-key-events -->
+<!-- svelte-ignore a11y-no-static-element-interactions -->
```
benoitf: should we use `<button>` below ? AFAIK it makes the expected behaviour rather than applying on:click on a div

deboer-tim: Initially I was a little hesitant because 'this is the minimal change to what we were doing before', I wasn't trying to change this, and I think we need to look at how this should work / #3109 / maybe a table column url property - but I tried, it works, and is a little simpler. Updated in latest commit.

deboer-tim:
> I see no component test for most of the .svelte component ?

IMHO the individual column components are so small that there's limited benefit to testing separately from the List component, and one of my initial goals was switching to this component with minimal breakage to existing (List) tests (i.e. this should be a change of implementation with no functional and limited visible change). That said, the existing tests for images are pretty limited... What area are you most concerned about?

benoitf:
> IMHO the individual column components are so small that there's limited benefit to testing separately from the List component, and one of my initial goals was switching to this component with minimal breakage to existing (List) tests (i.e. this should be a change of implementation with no functional and limited visible change). That said, the existing tests for images are pretty limited... What area are you most concerned about?

The idea is that someone can decide to move the component elsewhere at some point. Then, we would benefit from specific tests.

For example, checking that the click on the details of an image is simpler on ImageColumnName.svelte component rather on the image list. So we have very small scoped tests.
 (https://github.com/containers/podman-desktop/pull/4844/files#diff-2399065e5eeac2c889853bc5b5dcf0cffa288fd0d25384d3ad3a9efb2d261d65)

when it goes to the image list, usually we're not testing heavily all the details/deep stuff as we rely on the component's test.

deboer-tim:
Ok. I hope to remove some of these soon through things like #5011, but tests added for each of the columns.

benoitf (approved):

slemeur:
For release note 

images-sorting

PR #4856, opened 2023-11-17
chore: rename <icon component to <Icon

benoitf:
### What does this PR do?
React components should start with an uppercase else it is considered as a default HTML component

fixed issue with docusaurus v3 enforcing that rule (React 18 vs 17)

https://github.com/containers/podman-desktop/pull/4764

### How to test this PR?
should not have any visual change


themr0c (approved):

benoitf:
argos is reporting no screenshot difference so I think we're good !

PR #4859, opened 2023-11-17
Image Checker UI

### LINKED ISSUE #4707: Image Checker: reporting UI

As provider may report image checks thanks to #4697 the checks should be reported in the Images page

Report checks in a new tab

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

This PR adds a UI for image checkers.

https://github.com/containers/podman-desktop/assets/9973512/d409b774-b92d-45bc-815f-a4f575d45ebf

Fixes #4707 

### How to test this PR?

You can install a fake image checker from the image at `quay.io/phmartin/fake-image-checker:v0.0.3` (sources: https://github.com/feloy/podman-desktop-extension-fake-image-checker)

Go to the details page of an image, and click the Check tab. Use can use the Abort button to ask the image checker to stop processing.


packages/renderer/src/lib/image/ImageDetailsCheck.svelte
```diff
@@ -0,0 +1,149 @@
...
+});
+
+onDestroy(() => {
+  // unsubscribe from the store
+  if (providersUnsubscribe) {
+    providersUnsubscribe();
+  }
+});
+
+async function callProviders(_providers: readonly ImageCheckerInfo[]) {
```
jeffmaury, line 40: I'm wondering why the aggregation is done on the UI side (which needs more exchanges with the main process) and why it's not done in the main process (single call)

jeffmaury (commented):
I also have a concern about UI as I think we should have a way to focus on a specific severity so each severity should either be in a expandable component or be filterable @mairin may help

feloy:
> I also have a concern about UI as I think we should have a way to focus on a specific severity so each severity should either be in a expandable component or be filterable @mairin may help

For now, the results are sorted by status (failed before passed), then severity (critical ones first), then by provider name. This can help focusing on a specific severity.

packages/renderer/src/lib/image/ImageDetailsCheck.svelte
feloy: It helps differentiate the results of each provider (the user can Abort before waiting all providers to return results, and it will get the results of the first providers).

packages/renderer/src/lib/image/ImageDetails.svelte
```diff
@@ -55,6 +65,13 @@ onMount(() => {
     }
   });
 });
+
+onDestroy(() => {
+  // unsubscribe from the store
+  if (providersUnsubscribe) {
+    providersUnsubscribe();
+  }
```
benoitf:
```suggestion
providersUnsubscribe?.();
```

packages/renderer/src/lib/image/ImageDetailsCheck.svelte
```diff
@@ -0,0 +1,149 @@
...
+  providersUnsubscribe = imageCheckerProviders.subscribe(providers => {
+    callProviders(providers);
+  });
+});
+
+onDestroy(() => {
+  // unsubscribe from the store
+  if (providersUnsubscribe) {
+    providersUnsubscribe();
+  }
```
benoitf:
```suggestion
  // unsubscribe from the store
providersUnsubscribe?.();
```

packages/renderer/src/lib/image/ImageDetailsCheck.svelte
```diff
@@ -0,0 +1,149 @@
...
+          const providerA = orderProvider.indexOf(a.provider.id);
+          const providerB = orderProvider.indexOf(b.provider.id);
+          if (providerA !== providerB) {
+            return providerA - providerB;
+          }
+          return 0;
+        });
+        results = [...results];
+      })
+      .catch((e: unknown) => console.log('check error', e));
```
benoitf: it looks like user is not notified that there was an error there
results should contain the error no ?

benoitf (commented):
Nice demo 🎉 

I know the spec was about an image checker but I'm wondering if we need to call the tab  'Check' or we should call it 'Analysis' or 'Analytics' etc

(can be done in follow-up PR, etc, I'm just raising the question)



On the result layout, I would expect like for trivy:

image

A search to be able to filter elements (like we have in containers, pods, images, volumes, etc)

then for the filtering of severity, reuse kind of layout used for filtering All containers, Running containers and stopped containers

and for the display, should we use the table component instead ?

severity being on the left, and a way to open/collapse the details

feloy:
> Nice demo 🎉
> 
> 
> On the result layout, I would expect like for trivy:
> 
> image
> 
> A search to be able to filter elements (like we have in containers, pods, images, volumes, etc)
> 
> then for the filtering of severity, reuse kind of layout used for filtering All containers, Running containers and stopped containers
> 

Is it expected for this issue, or for a following one?


> and for the display, should we use the table component instead ?
> 
> severity being on the left, and a way to open/collapse the details

benoitf:
Can be follow-up, it's like open questions but goal IMHO is to match some of our patterns when we have list like for containers, pods, images, volumes, etc. (search, toggle, filter, table)

benoitf:
the only 'blocker' for me is to be able to see that there was an error during a checker (unexpected error)

I think our buttons are named 'Cancel' rather than 'Abort' 

we have like 'infinite spinner line' when we create a podman machine
image

I would expect probably the same design there below the 'Analysis/check in progress'

benoitf:
and my last comment. Maybe that we keep a title, space always on top of the page because it'll make a blink/scroll effect when action is finished. (Let say I start to look at the result and we have one checker still running)
when it's finished, it'll remove the top banner/cancel steps then the content will move upper

benoitf:
ok it wasn't my last comment.

Maybe a "Retry" button when we have aborted. (for example in your workflow you cancel and see no results, when you want to try again you need to switch to another tab and then come back to the tab again to launch again)

Maybe a button to retry/spawn again this could be interesting even in case of success (if for example you've updated the image in a terminal, you just want to launch the analysis again without going elsewhere)

feloy:
> and my last comment. Maybe that we keep a title, space always on top of the page because it'll make a blink/scroll effect when action is finished. (Let say I start to look at the result and we have one checker still running) when it's finished, it'll remove the top banner/cancel steps then the content will move upper

The result being sorted, the results of the second checker will be inserted anywhere in the list, which will also move the existing results of the first checker

feloy:
> Nice demo 🎉
> 
> I know the spec was about an image checker but I'm wondering if we need to call the tab 'Check' or we should call it 'Analysis' or 'Analytics' etc
> 
> (can be done in follow-up PR, etc, I'm just raising the question)
> 
> On the result layout, I would expect like for trivy:
> 
> image
> 

> and for the display, should we use the table component instead ?
> 

There is this arrow at the right end of the row, to see the details (it will be necessary for displaying the description of the error, which can be quite long). Do we (want to) support this in our table implementation? @deboer-tim @mairin

feloy:
> 
> then for the filtering of severity, reuse kind of layout used for filtering All containers, Running containers and stopped containers
> 

This would introduce a second level of tabs, would it be ok? @mairin 

image-check-double-tabs

mairin:
Board (1)


I played with this for a couple hours or so... maybe this is me tweaking too far. But a few ideas:

- maybe toggles for severity?  
- ability to turn checkers on or off (while a checker is still loading, the toggle could be a spinner)

feloy:
I've added telemetry, to send imagecheck events every time a check is started for a provider, with provider and error properties. (I've created related charts in Amplitude)

jeffmaury (approved):
LGTM but still have the check executed when the tab is clicked

feloy:
> LGTM but still have the check executed when the tab is clicked

What do you mean? To abort the checks when the user leaves the page?
I have added a commit to cover this.

extensions/image-checker/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "image-checker",
+  "displayName": "Image Checker *",
+  "description": "Analyze a Dockerfile and highlight the directives and commands which could cause an unexpected behavior when running on an OpenShift cluster.",
```
lstocchi: Should be `Analyze` a `Containerfile?`

packages/renderer/src/lib/image/ImageDetailsCheck.svelte
```diff
@@ -0,0 +1,184 @@
...
+        _result?.checks.forEach(check => {
+          results.push({
+            provider,
+            check,
+          });
+        });
+        results.sort((a, b) => {
+          const orderStatus = ['failed', 'success'];
+          const orderSeverity = ['critical', 'high', 'medium', 'low', undefined];
+          const orderProvider = providers.map(p => p.id).sort();
```
lstocchi: they can be moved outside

packages/renderer/src/stores/image-checker-providers.ts
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { writable, type Writable } from 'svelte/store';
+import type { ImageCheckerInfo } from '../../../main/src/plugin/api/image-checker-info';
+import { EventStore } from './event-store';
+
+const windowEvents = ['image-checker-provider-update'];
```
lstocchi: I wonder if there should be an event for when an extension is removed.
If i uninstall the image-checker extension, do the registered imageCheckers get removed?

extensions/image-checker/package.json
feloy: This is not really part of the PR, as this fake extension will be removed before to merge, I note it for the next PR

packages/renderer/src/stores/image-checker-providers.ts
feloy: Good catch! Without `image-checker-provider-remove` here, when disabling the image checker, the Check tab was still displayed.

When I set `image-checker-provider-remove` in the list, the Check tab disappears when the user disables the Image Checker, and re-appears when the extension is restarted.

jeffmaury:
> > LGTM but still have the check executed when the tab is clicked
> 
> What do you mean? To abort the checks when the user leaves the page? I have added a commit to cover this.

No my concern is that if I go to the check tab, provider are called, wait for all of them, then go to another tab and go back and providers are called again

feloy:
> > > LGTM but still have the check executed when the tab is clicked
> > 
> > 
> > What do you mean? To abort the checks when the user leaves the page? I have added a commit to cover this.
> 
> No my concern is that if I go to the check tab, provider are called, wait for all of them, then go to another tab and go back and providers are called again

Ah, ok. I understood from the design doc that we wanted to make this better for a next milestone only (by running the checks in the background or similar)

feloy:
I reverted the commit with the fake extension, to prepare the PR to be merged. Feel free to get rid of this revert commit locally if you want to test the PR again with the extension

lstocchi (approved):
LGTM, i also agree with Jeff. We could enhance it by saving the result based on the id/sha, maybe in a follow-up PR?

benoitf:
> I reverted the commit with the fake extension, to prepare the PR to be merged. Feel free to get rid of this revert commit locally if you want to test the PR again with the extension

hi, maybe it would be easier to publish the extension example to quay.io (or another registry) so people wanted to try can pull the extension and try it out

I can't get back the commit that is no longer in the branch (let say I want to get back the source code of the extension)

feloy:
> > I reverted the commit with the fake extension, to prepare the PR to be merged. Feel free to get rid of this revert commit locally if you want to test the PR again with the extension
> 
> hi, maybe it would be easier to publish the extension example to quay.io (or another registry) so people wanted to try can pull the extension and try it out
> 
> I can't get back the commit that is no longer in the branch (let say I want to get back the source code of the extension)

Right, I will publish the fake extension.

I didn't delete the initial commit, but created a new one reverting the initial one (see the latest commit `Revert "feat: fake extension"`. By deleting this latest one, you can find again the original one

benoitf:
> I didn't delete the initial commit, but created a new one reverting the initial one (see the latest commit Revert "feat: fake extension". By deleting this latest one, you can find again the original one

I have the sha but it's orphan, so I can't retrieve it

NB: using a local git command, not trying to browse online

feloy:
> > I didn't delete the initial commit, but created a new one reverting the initial one (see the latest commit Revert "feat: fake extension". By deleting this latest one, you can find again the original one
> 
> I have the sha but it's orphan, so I can't retrieve it
> 
> NB: using a local git command, not trying to browse online

What if you do
```
$ git checkout feat-4454/image-checker-ui
$ git pull 
$ git show
commit 4f47d7685346ca12ebb250179460c671795d71c5 (HEAD -> feat-4454/image-checker-ui, origin/feat-4454/image-checker-ui)
Author: Philippe Martin <phmartin@redhat.com>
Date:   Thu Nov 23 13:40:49 2023 +0100

    Revert "feat: fake extension"
    
    This reverts commit 69962c6319fba7186c01e8a5f793b7eb21072735.
$ git reset HEAD~
$ git checkout .
$ ls extensions/image-checker/
icon.png  package.json  src  tsconfig.json  vitest.config.js
```
```

benoitf:
oh ok you didn't remove the commit, revert it 👍

benoitf:
image


on my side it's still looking like this

benoitf (changes requested):
IMHO it requires a quick polish before landing to main

design:
image

implementation:
image

benoitf:
I think toggle can be a step 2 but it requires quick polishing before merging:

- static header (stethoscope icon) (it should never move to the top, disappear etc. when we have result coming, etc)
- left/right page with left = name of checkers (for now toggle can be later and in progress icon if still processing?)
- width of result should leave some space on the right here with the scrollbar it's overlapping with content behind
- left border border-l tailwind with severity
- rounded only on the right of the tile
- use color on the icon rather than on the text of the problem

toggles in checkers, filters and dropdown to minimize/maximize can be a follow-up

feloy:
> hi, maybe it would be easier to publish the extension example to quay.io (or another registry) so people wanted to try can pull the extension and try it out

I have published an image at `quay.io/phmartin/fake-image-checker:v0.0.3` (see details in the PR description https://github.com/containers/podman-desktop/pull/4859#issue-1999638319)

feloy:
> I think toggle can be a step 2 but it requires quick polishing before merging:
> 
> * static header (stethoscope icon) (it should never move to the top, disappear etc. when we have result coming, etc)
> * left/right page with left = name of checkers (for now toggle can be later and in progress icon if still processing?)
> * width of result should leave some space on the right here with the scrollbar it's overlapping with content behind
> * left border border-l tailwind with severity
> * rounded only on the right of the tile
> * use color on the icon rather than on the text of the problem
> 
> toggles in checkers, filters and dropdown to minimize/maximize can be a follow-up

I'm working on it.

Where would we like to display the error when a provider fails to run the analysis?

benoitf:
image

maybe below the title ?

feloy:
> image
> 
> maybe below the title ?

As the error is attached to the provider, I wonder if we could display it below the name of the provider, in the left column, or, because the text can be long, only display an alert icon, and have the error message when the user hovers the icon (but I'm not sure it is accessible)

feloy:
I've added a commit with the new design. See demo in PR description

benoitf (approved):
thanks @feloy I think all other feedback, requests can come in follow-up PRs now
Ui is now in a better shape

PR #4860, opened 2023-11-17
feat: default table sorting

### LINKED ISSUE #4838: Table sorting improvements

I know the Table component hasn't been released yet, but I've already found two things that annoy me, I have a fix ready, and don't want to hold up the main PR. The two issues:

- I added table sorting, optional for each column. The sort icons are always visible so that you know which columns are sortable, but they stand out and are too noticeable.
- The initial data is already sorted by one of the columns. There's no indication of this, and if you click to sort that column nothing happens.

Suggested fixes for both issues:
- Sort icons should be dimmed by default and only the current sort should be the normal color.
- The Table should support a 'default sort column' so that when you open it the correct column already has a sort indicator.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds support for a 'default sort column' so that you can tell if the data was initially sorted by a particular column (and don't try to sort that column and have nothing change).

Dims the sort indicator on non-active columns so that you can still tell which columns are sortable, but they don't stand out as much.

Also includes a minor '&&' simplification suggested in previous PR review.

sort2

Fixes #4838.

### How to test this PR?

It will change the column sort indicator if used in conjunction with the Image or Volume PRs; to set a default sort column you need to add something like `defaultSortColumn="Name"` to one of those tables.

### DIFF
--- a/packages/renderer/src/lib/table/Table.svelte
+++ b/packages/renderer/src/lib/table/Table.svelte
@@ -7,7 +7,7 @@
 <script lang="ts">
 /* eslint-disable import/no-duplicates */
 // https://github.com/import-js/eslint-plugin-import/issues/1479
-import { afterUpdate, tick } from 'svelte';
+import { afterUpdate, onMount, tick } from 'svelte';
 import Checkbox from '../ui/Checkbox.svelte';
 import type { Column, Row } from './table';
 import { flip } from 'svelte/animate';
@@ -17,11 +17,12 @@ export let kind: string;
 export let data: any[];
 export let columns: Column<any>[];
 export let row: Row<any>;
+export let defaultSortColumn: string | undefined = undefined;
 
 // number of selected items in the list
 export let selectedItemsNumber: number = 0;
 $: selectedItemsNumber = row.info.selectable
-  ? data.filter(object => row.info.selectable?.(object)).filter(object => object.selected).length
+  ? data.filter(object => row.info.selectable?.(object) && object.selected).length
   : 0;
 
 // do we need to unselect all checkboxes if we don't have all items being selected ?
@@ -65,6 +66,17 @@ function sort(column: Column<any>) {
   // eslint-disable-next-line etc/no-assign-mutated-array
   data = data.sort(comparator);
 }
+
+onMount(async () => {
+  if (defaultSortColumn) {
+    for (const column of columns) {
+      if (column.title === defaultSortColumn) {
+        sortCol = column;
+        sortAscending = true;
+      }
+    }
+  }
+});
 
 afterUpdate(async () => {
   await tick();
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/table/Table.svelte
```diff
@@ -66,6 +67,17 @@ function sort(column: Column<any>) {
...
+  if (defaultSortColumn) {
+    for (const column of columns) {
+      if (column.title === defaultSortColumn) {
+        sortCol = column;
+        sortAscending = true;
+      }
+    }
+  }
+});
+
```
benoitf: it does the job but I'm wondering if we could not use some filter/find/map/etc keywords to to the same without having 3 sub-blocks

benoitf (approved):
probably missing a test to check that

packages/renderer/src/lib/table/Table.svelte
deboer-tim: Changed to use .find() in new commit.

deboer-tim:
Added tests.

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -66,6 +67,16 @@ function sort(column: Column<any>) {
   data = data.sort(comparator);
 }
 
+onMount(async () => {
+  if (defaultSortColumn) {
+    const column: Column<any> | undefined = columns.find(column => column.title === defaultSortColumn);
```
benoitf: I'm wondering if we need the if (defaultSortColumn) as if it's not defined it won't match as it looks like that title is not optional

benoitf: (also I'm still approving, my comments were not "required for changes" request)

deboer-tim: Understood. I figured I may as well make the changes while I waited for another approval. I've removed the if, and if there are no further comments in the next hour I think it has been long enough/is small enough that I'll merge.

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -66,6 +67,14 @@ function sort(column: Column<any>) {
   data = data.sort(comparator);
 }
 
+onMount(async () => {
+  const column: Column<any> | undefined = columns.find(column => column.title === defaultSortColumn);
+  if (column) {
+    sortCol = column;
+    sortAscending = true;
+  }
```
dgolovin, line 75:
```suggestion
    sort(column);
  }
```

dgolovin: it is not sorting default column after opening the page.

https://github.com/containers/podman-desktop/assets/620330/e2542138-f0ce-4da8-8627-e16954d80624

dgolovin (changes requested):

packages/renderer/src/lib/table/Table.svelte
deboer-tim: The intention (for now) wasn't to provide a way to sort the initial data by any column, but just to set the initial sort indicator _when the initial data is already sorted_.

The suggested change doesn't work because most of our existing pages are dynamic / lazy-loading - data is empty when the page is shown (so onMount would sort nothing) and then data is set. To be able to set an arbitrary sort column we'd need to sort after the data is received.

We may need to support this at some point, but since our current pages don't need it I would prefer to release what's here and defer arbitrary default sorting until later.

dgolovin (approved):
LGTM

PR #4861, opened 2023-11-17
fix: use vi.waitUtnil instead of cycles with awaiting promises

dgolovin:
### What does this PR do?
this PR replaces:
* awaits for promises resolved by using timers
* cycles with awaits for promises

to `vi.waitUntil`


benoitf (approved):

benoitf (approved):

deboer-tim (approved):

jeffmaury (commented):
There a lots of other test using this pattern, we should fix them as well ?

dgolovin:
IMO we should fix it whenever we add or update tests, it is not critical.

PR #4864, opened 2023-11-19
chore: use font color rather than hardcoding fff as a color

benoitf:
### What does this PR do?
use the color in the svg rather than hardcoding the value

should not cause any regression

### How to test this PR?

Check the dashboard icon and settings icon in the left navbar, it should be as before


deboer-tim (approved):
Looks like `color` can be removed and just set `fill`? I'm good either way since neither should be hardcoded.

benoitf:
@deboer-tim I thought as well that it was only for fill but it didn't worked so maybe fill was not used. Will investigate later 👍 
(probably need to extract the svg from AppNavigation and move it as a *Icon component as well)

Issue #4866, opened 2023-11-20
Fix Podman Extension E2E test method descriptions that use invalid character when saving the artifacts


odockal:
### Bug description

Failed test on GHA caused the artifact archive to be invalid.

```
Artifact name is valid!
Error: Artifact path is not valid: /screenshots/Disable_Podman_extension_from_Settings_->_Extensions_->_Podman_page_and_verify_status_failure.png. Contains the following character:  Greater than >
          
Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n
          
The following characters are not allowed in files that are uploaded due to limitations with certain file systems such as NTFS. To maintain file system agnostic behavior, these characters are intentionally not allowed to prevent potential problems with downloads on different file systems.
```

https://github.com/containers/podman-desktop/actions/runs/6901490632/job/18777821429

### Operating system

Ubuntu 22.04

### Installation Method

Other

### Version

next (development version)

### Additional context

Archiving the artifacts has some restriction as of file names, and so we need to avoid using these.

PR #4867, opened 2023-11-20
fix(ci,tests): remove using of invalid characters in test method desc…

odockal:
…ription

### What does this PR do?
Fixes test method names which uses invalid character when the screenshot with such description is used during GHA steps of archiving the artifacts.

#4866

### DIFF
--- a/tests/src/podman-extension-smoke.spec.ts
+++ b/tests/src/podman-extension-smoke.spec.ts
@@ -66,13 +66,13 @@ describe('Verification of Podman extension', async () => {
   test('Podman is enabled and present', async () => {
     await verifyPodmanExtensionStatus(true);
   });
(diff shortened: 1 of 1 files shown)


benoitf (approved):

benoitf:
it fixes the cause thanks but should we add a replace on invalid characters before taking screenshots so it avoids other potential problems ?

feloy (approved):

jeffmaury (approved):
LGTM but I'm afraid it's going to happen again. Is there a way to fix it when test name is mapped to a file path ?

odockal:
@jeffmaury @benoitf Here is an follow up issue: https://github.com/containers/podman-desktop/issues/4905.

PR #4879, opened 2023-11-20
fix: retain autostart setting (#4647)

### LINKED ISSUE #4647: Podman Desktop doesn't retain autostart setting
### Bug description

I have installed Podman on my MacOS M1 Macbook using `brew install podman` and then installed Podman Desktop using `brew install podman-desktop`. I then ran through the initial setup and created the machine/VM. When running through the initial setup, I ensured the autostart option was selected. However, the machine/VM does not autostart.

I've gone into Preferences and enabled Autostart in the Autostart Podman engine section. It shows as enabled. But if I quit Podman Desktop or reboot my computer, the setting value is not retained. And the machine/VM does not autostart.

I am currently running v1.5.2.

### Operating system

macOS Sonoma 14.0 (Arm)

### Installation Method

Brew (macOS)

### Version

1.5.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR cleans the code that defines the autostart setting and fixes a problem when storing its value.

N/A

it fixes #4647 

### How to test this PR?

1. check that the autostart setting is actually saved
2. set autostart, stop machine, restart podman desktop, check that the autostart setting is true and the podman machine is started
3. set autostart to false, stop machine, restart desktop, check autostart setting is false and the podman machine is still stopped.

### DIFF
--- a/packages/main/src/plugin/autostart-engine.spec.ts
+++ b/packages/main/src/plugin/autostart-engine.spec.ts
@@ -50,17 +50,11 @@ beforeAll(() => {
   providerRegistry.runAutostart = mockRunAutostart;
 });
 
-test('Check that default value is false if provider autostart setting is undefined and old global setting is false', async () => {
-  vi.spyOn(configurationRegistry, 'getConfiguration').mockImplementation((section?: string) => {
-    if (section === `preferences.${extensionId}`) {
-      return {
-        get: (_section: string) => undefined,
-      } as Configuration;
-    } else {
-      return {
-        get: (_section: string) => false,
-      } as Configuration;
-    }
+test('Check that default value is false if provider autostart setting is false', async () => {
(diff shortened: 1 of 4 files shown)


benoitf:
I have a question, I see lot of people still using old versions

here it looks like if we remove all properties, users migrating from an old version will have their autostart setting being ignored ?

lstocchi:
@benoitf ppl that uses the old version will end up to a true value. 
Having both complicates the code and we agreed on moving on specific provider autostart setting. 
As the autostart is not working as expected i think it's a good time to remove old code and start from 0 again. The worst scenario is that people will see the autostart setting enabled if they set it to false using the old version, but we can say there was a bug and it was fixed. If you disable it again, it will work fine.

benoitf (approved):

jeffmaury (approved):
LGTM

Issue #576, opened 2022-10-11
Light mode button?


cdrage:
### Is your enhancement related to a problem? Please describe

I can't seem to find a button / setting to switch Podman Desktop interface to light mode?

### Describe the solution you'd like

I see code such as: https://github.com/containers/podman-desktop/commit/ad6f4fc4c41e3fd6102203f5b5b83c53698149bc

But I'm unable to find a button or documentation describing how to switch the GUI to light mode.

benoitf:
It's not yet supported ( waiting UI mockups, etc.)

so in fact the default mode is dark but to turn it to dark mode, all styles should be prefixed by dark:


Website is implementing both light and dark mode (+ a toggle)

SubOptimal:
I would appreciate light theme support, as dark themes are not suitable for everyone.
I have a problem with colour vision, and the dark themes make it even more difficult for me to use the applications.

gotmax23:
> I would appreciate light theme support, as dark themes are not suitable for everyone. I have a problem with colour vision, and the dark themes make it even more difficult for me to use the applications.

Indeed. Not having a light mode poses significant accessibility issues.

nozzlegear:
I'd also love to see light mode support. It's difficult for me to use this in dark mode due to my own vision problems, and I'd consider them relatively mild.

markenzie:
> It's not yet supported ( waiting UI mockups, etc.)

I would be fine even just passing a flag to the startup script or a hidden key combination. 

Here is a good article detailing and demonstrating (with _slightly exaggerated_ pictures) how someone with astigmatism views light on dark:

https://bootcamp.uxdesign.cc/why-dark-mode-isnt-a-ux-panacea-1590226e5ba4

![image](https://github.com/containers/podman-desktop/assets/3488874/5a5f08d4-a0ac-4697-a2fe-93234bce8db4)

Viltersten:
To be honest (and I'm sorry to sound harsh), I find it nonchalant to fail acting on this request that's been reported 1.5 years ago. No, I don't think it's that hard to create a bright mode version. (It may be suboptimized, possibly distributed separately or such, as long as it's present.) Imagine someone in a wheelchair who can't access a building but no ramp or other workaround is provided. It's legally dubious and definitely wrong ethically.

[...]

PR #4887, opened 2023-11-20
feat: introduce property for appearance but for now only dark is supported

benoitf:
### What does this PR do?
add a new setting which is hidden to toggle between light or dark mode
it's hidden because if the user is asking for light mode it will be still displayed with dark colours.

It's set to `system` mode but even if all the logic is there, at the end we force 'dark' mode for the system mode to not break users having wrong display of items until the light mode is completed

But it toggles on or off the dark theme support then we can use dark: prefix in the codebase to make the style being specific for the dark theme

it's like the beginning of more upcoming PRs

N/A as setting is hidden

related to #576 

### How to test this PR?

you can replace `hidden: true` by `hidden:false` and then you'll see the preference in preferences page

default is system so it means it'll use mode of the underlying operating system
(but default will still be mapped now to dark to not let display something broken to the user as it's the default property)

then you can inspect with 'light' or 'dark' toggle the class of the html root element (it should add `dark` or remove `dark` if you change the property)

unit tests have been added

### DIFF
--- a/packages/main/src/plugin/appearance-init.ts
+++ b/packages/main/src/plugin/appearance-init.ts
@@ -1,0 +1,43 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { IConfigurationNode, IConfigurationRegistry } from './configuration-registry.js';
+import { AppearanceSettings } from './appearance-settings.js';
+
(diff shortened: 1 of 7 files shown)


lstocchi:
So from now on, we should go and update all `bg`/`text` color classes with `dark`: .... and the default value will be used in light mode?

benoitf:
@lstocchi yes that is the plan unless using a radical different approach like https://github.com/L-Blondy/tw-colors but it won't work for what we have now as we're not using 'primary' color but a color from a palette

deboer-tim (approved):
I had a very similar change that I was starting to test, and I'm not ashamed to say this is more thorough and sets us up well for opening the preference later. :)

My only concern is that I think we should also set the default color scheme in addition to the class, i.e.:
`html.style.setProperty('color-scheme','light');`
and
`html.style.setProperty('color-scheme','dark');`
in Appearance.svelte.

This would set the default scheme so that we only need to override where we don't match, instead of having to set styles for everything. If we do this then it's also a fix for #4888 (and we can close that part of #4826). If not, then we need to create our own dark scrollbar styling and always do things like "text-black dark:text-white" even though that would be the default.

Approving since this is an extra step beyond and I agree with what's here.

deboer-tim:
Yes, I think to support light we'll be changing every 'bg-x' to 'bg-<new value> dark:bg-x'. We'll need to slowly work our way through each of these anyway, so not hard to prefix existing classes.

lstocchi (approved):

benoitf:
I played over the week end to see some issues (and there are ;-)

it's something that we could render:
image

https://github.com/containers/podman-desktop/assets/436777/0e320b46-0daf-44b0-a961-e6f64d144684

Issue #4905, opened 2023-11-21
Transfer characters in test method names  that are marked as invalid by GHA archive artifacts action


odockal:
### Is your enhancement related to a problem? Please describe

Find a better way to avoid this kind of problems (https://github.com/containers/podman-desktop/issues/4866) in the tests.

### Describe the solution you'd like

Translate or transfer problematic characters in test method names used as a file names for taking a screenshot during test failure, that are then uploaded by GHA archive artifact action.

PR #4908, opened 2023-11-21
fix(ci,tests): improve test failure screenshot files naming

odockal:
#4905

### How to test this PR?
Include some invalid character (as defined by filesystem or third party program) into test method name, make the test fail and check failure screenshot file name in `tests/output/screenshots` folder.

### DIFF
--- a/tests/src/setupFiles/extended-hooks.ts
+++ b/tests/src/setupFiles/extended-hooks.ts
@@ -23,11 +23,7 @@ import * as path from 'node:path';
 
 afterEach(async (context: RunnerTestContext) => {
   context.onTestFailed(async () => {
-    const normalizedFilePath = context.task.name
-      .replace(/'/g, '')
-      .replace(/\//g, '_')
-      .replace(/:/g, '_')
-      .replace(/ /g, '_');
+    const normalizedFilePath = context.task.name.replace(/([/: ])/g, '_').replace(/(['"><|*?\r\n])/g, '');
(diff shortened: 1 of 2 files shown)


tests/src/setupFiles/extended-hooks.ts
```diff
@@ -23,11 +23,7 @@ import * as path from 'node:path';
 
 afterEach(async (context: RunnerTestContext) => {
   context.onTestFailed(async () => {
-    const normalizedFilePath = context.task.name
-      .replace(/'/g, '')
-      .replace(/\//g, '_')
-      .replace(/:/g, '_')
-      .replace(/ /g, '_');
+    const normalizedFilePath = context.task.name.replace(/([/: ])/g, '_').replace(/(['"><|*?\r\n])/g, '');
```
benoitf: I was wondering if we should not like replace all non letters/digits characters ?

something like `replace(/[^a-zA-Z0-9]/g, '')`;

jeffmaury: Seems more readable

odockal: I have included a bit more sophisticated replacement that will remove most characters that are not letters/numbers or underscore, which I want to use as a delimiter to a words in the filename to keep better readability.

jeffmaury (approved):
LGTM

cdrage (approved):
LGTM hopefully this fixes some of the flakes

PR #4918, opened 2023-11-21
fix: unable to read wsl version when using chinese as syslang on Windows (#4863)

### LINKED ISSUE #4863: Unable to read the version of the WSLcorrectly

When executing entire setup, the following error message is prompted
image

But my local WSL version is greater than required.
image

Can't recognize the WSL version of the machine, or is my WSL version incorrect?

Sorry, I couldn't find the feedback email, so I went to Github to create an issue. If need more detailed machine version information, I can provide it here

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR fixes an issue when detecting wsl version on a chinese lang system.
There were 2 problems, the first we encoded using utf8 when wsl returns an output encoded utf16, and then the output contains a fullwidth colon which is different from the one we use in ASCII.

N/A 

it fixes #4863

### How to test this PR?

1. switch to chinese on windows or just run on your usual syste mto see if it works as before

### DIFF
--- a/extensions/podman/src/wsl-helper.spec.ts
+++ b/extensions/podman/src/wsl-helper.spec.ts
@@ -59,7 +59,7 @@ Windows version: 10.0.22621.2134
   expect(windowsVersion).toBe('10.0.22621.2134');
 
   // expect called with wsl --version
-  expect(extensionApi.process.exec).toHaveBeenCalledWith('wsl', ['--version']);
+  expect(extensionApi.process.exec).toHaveBeenCalledWith('wsl', ['--version'], { encoding: 'utf16le' });
 });
 
 test('should grab correct versions even with an output with a language different from english', async () => {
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/util/exec.ts
```diff
@@ -180,8 +180,8 @@ export class Exec {
         reject(errResult);
       });
 
-      childProcess.stdout.setEncoding('utf8');
-      childProcess.stderr.setEncoding('utf8');
+      childProcess.stdout.setEncoding(options?.encoding || 'utf8');
+      childProcess.stderr.setEncoding(options?.encoding || 'utf8');
```
benoitf: it's minor but linter recommends nullish coalescing rather than the OR operator

benoitf: `??`

benoitf (approved):
probably need to split in two, one part for extension api, encoding, etc, tests

and another PR for just podman

benoitf (commented):
interesting reading and fix 👍

feloy (approved):
works on my Windows 10 machine

jeffmaury (approved):
LGTM. Maybe test should be enhanced to exhibit the difference values for :

lstocchi:
just added a test for the small colon symbol so now all symbols are covered

PR #4921, opened 2023-11-21
feat: show what the current context is

### LINKED ISSUE #4618: Kube Config UI: View current context

cdrage:
feat: show what the current context is

### What does this PR do?

* Refactors the Kubernetes window calls a bit so we can do it in one
  call that returns the current context, cluster information, etc.
* Now shows what the current context is / slight shade / different
  background

https://github.com/containers/podman-desktop/assets/6422176/e061fdbd-c512-40c9-b4eb-92aed3558bfb

Closes https://github.com/containers/podman-desktop/issues/4618

### How to test this PR?

1. Have a .kube/config
2. Go to Kubernetes view in settings
3. Edit the current context to a different value and you will see the
   current context change

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -145,6 +145,7 @@ import { NotificationRegistry } from './notification-registry.js';
 import { ImageCheckerImpl } from './image-checker.js';
 import type { ImageCheckerInfo } from './api/image-checker-info.js';
 import { AppearanceInit } from './appearance-init.js';
+import type { KubeContext } from './kubernetes-context.js';
 
 type LogType = 'log' | 'warn' | 'trace' | 'debug' | 'error';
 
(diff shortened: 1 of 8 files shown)


cdrage:
Some notes:

This PR is based upon #4745 so we can either merge #4745 or we can just merge this PR too.

@mairin I added the "Current Context" text because it was hard to see which one was the current context from the current mockups (just a different background shade). 

@benoitf I have refactored the retrieval of context information a bit as per your note on #4745

deboer-tim:
Just my feeling, but since #4745 already has one approval it's probably quicker to wait and merge that first (and just review the additions here) vs restarting approval.

I agree the current background color change is almost unnoticeable, but even if we improve contrast I think I'd be confused without the label. I don't like the placement and moving title though, I think a badge to the right of the card title (like to our Environment columns) would work better.

lstocchi (approved):
LGTM

dgolovin (approved):
Even though it works correctly (LGTM), it took time to realize what is changed on the page.

Imagine if current context is somewhere down below not visible, how that would work?

cdrage:
> Even though it works correctly (LGTM), it took time to realize what is changed on the page.
> 
> Imagine if current context is somewhere down below not visible, how that would work?

Shown a lot better in https://github.com/containers/podman-desktop/pull/4927 as we're adding the icon in there too to show how to switch between them.

PR #4922, opened 2023-11-21
feat: table sorting II

### LINKED ISSUE #4881: Improve table sort indicators

The current sort indicators are causing confusion and are 'incorrect' in some cases [1]. I put that in quotes, because I did what I thought was correct at the time and googling I found that neither the internet nor user research seem to agree with itself in this area. :)

That said, the comments in #4841 seem to match what more people prefer and match what MacOS is doing, so we should change:
- Sort indicator should point up for ascending, and down for descending.
- Most columns (e.g. names) sort ascending on first click, but some (e.g. sizes) should sort descending.

I've also found that there are some aria indicators that we can use [2].

[1] https://github.com/containers/podman-desktop/pull/4841#issuecomment-1818815145

[2] https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/

Change indicators should be as described above. We'll need to add a Column property to allow some columns (e.g. Sizes) set their initial sort to descending.

Add aria sort properties.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Fixes the sort issues raised in issue #4881:

- Fixed sort direction: ascending sort should be 'up', descending 'down'.
- Added a new 'initialSortDir' for Columns like Size to specify that they should be sorted in descending order by default. This could have been a boolean flag, but seeing "initialSortDir: 'descending'" in Column definitions seems more natural and obvious to me.

This also fixes the sorting gap raised in #4860: constant sorting by any column.

Prior to this change sorting only happens once, when the user clicks on the column header. If the data changes it becomes unsorted again, and the column header is then wrong/out of sync. e.g. if you're in the Containers view, sort by Status, and then start a new container, data will revert back to the initial order and the new container will appear where it would have, even though the Status column still says it is sorted.

Now, sorting is also applied _whenever the data changes_. This means that if you sort by Status and start a container it would move to the top, or if you've sorted Age by descending it would move to the bottom. i.e. the sorting is always applied. This also means that any view can set the Table defaultSortColumn to any arbitrary column.

Note changed sort indicators (ascending is 'up'), Containers sort descending the first time you click on it (initialSortDir), and once sorted by Status the started pod moves to the top: 
https://github.com/containers/podman-desktop/assets/19958075/bc3c69ee-150b-44a5-b1eb-d48575a64d47

Fixes #4881.

### How to test this PR?

Use along with the Images or Volumes PR.

### DIFF
--- a/packages/renderer/src/lib/table/Table.svelte
+++ b/packages/renderer/src/lib/table/Table.svelte
@@ -40,6 +40,10 @@ function toggleAll(checked: boolean) {
 let sortCol: Column<any>;
 let sortAscending: boolean;
 
+if (data) {
+  sortImpl();
+}
+
 function sort(column: Column<any>) {
   if (!column) {
     return;
@@ -55,8 +59,23 @@ function sort(column: Column<any>) {
     sortAscending = !sortAscending;
   } else {
     sortCol = column;
-    sortAscending = true;
+    sortAscending = column.info.initialOrder ? column.info.initialOrder !== 'descending' : true;
   }
+  sortImpl();
+}
+
+function sortImpl() {
+  // confirm we're sorting
+  if (!sortCol) {
+    return;
+  }
+
+  let comparator = sortCol.info.comparator;
+  if (!comparator) {
+    // column is not sortable
+    return;
+  }
+
   if (!sortAscending) {
     // we're already sorted, switch to reverse order
     let comparatorTemp = comparator;
@@ -69,9 +88,9 @@ function sort(column: Column<any>) {
 
 onMount(async () => {
   const column: Column<any> | undefined = columns.find(column => column.title === defaultSortColumn);
-  if (column) {
+  if (column?.info.comparator) {
     sortCol = column;
-    sortAscending = true;
+    sortAscending = column.info.initialOrder ? column.info.initialOrder !== 'descending' : true;
   }
 });
 
@@ -84,7 +103,9 @@ function setGridColumns() {
   // section and checkbox columns
   let columnWidths: string[] = ['20px'];
 
-  if (row.info.selectable) columnWidths.push('32px');
+  if (row.info.selectable) {
+    columnWidths.push('32px');
+  }
 
   // custom columns
   columns.map(c => c.info.width ?? '1fr').forEach(w => columnWidths.push(w));
@@ -129,8 +150,8 @@ function setGridColumns() {
         {column.title}{#if column.info.comparator}<i
             class="fas pl-0.5"
             class:fa-sort="{sortCol !== column}"
-            class:fa-sort-up="{sortCol === column && !sortAscending}"
-            class:fa-sort-down="{sortCol === column && sortAscending}"
+            class:fa-sort-up="{sortCol === column && sortAscending}"
+            class:fa-sort-down="{sortCol === column && !sortAscending}"
             class:text-charcoal-200="{sortCol !== column}"
             aria-hidden="true"></i
           >{/if}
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/table/Table.svelte
```diff
@@ -40,6 +40,10 @@ function toggleAll(checked: boolean) {
 let sortCol: Column<any>;
 let sortAscending: boolean;
 
+$: if (data) {
+  sortImpl();
+}
+
```
benoitf: I would move the condition to the sort but it's only my feeling

packages/renderer/src/lib/table/table.ts
```diff
@@ -47,6 +47,19 @@ export interface ColumnInformation<Type> {
...
+   * naturally sorted in ascending order and do not need to
+   * specify this value - e.g. names are sorted alphabetically.
+   *
+   * Columns that are naturally sorted in descending order -
+   * e.g. file sizes or 'number of children' by biggest first -
+   * can set this value to change the initial sort direction.
+   *
+   * Defaults to 'ascending'.
+   */
+  readonly initialSortDir?: 'ascending' | 'descending';
```
benoitf: could be just be named `order` ?  (I mean, of course user can change it later but when we register objects we specify an order) (or then `initialOrder`

deboer-tim: Renamed to initialOrder in new commit.

packages/renderer/src/lib/table/Table.svelte
deboer-tim: I would too, but it is using svelte reactivity here: sortImpl() is called whenever data changes to make sure the sort is maintained. e.g. if we get an update from provider that replaces all the objects the current sort is still applied.

benoitf: I mean, to do:

```
$: sortImpl();
```

and in sortImpl if we don't have data, we abort

benoitf (approved):
I guess it requires a test ?

packages/renderer/src/lib/table/Table.svelte
deboer-tim: Yes, but if you remove `if (data)` then sortImpl() never gets called. Including data in the `$:` statement triggers svelte to register data as reactive and call sortImpl() whenever it gets changed.

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -55,8 +59,21 @@ function sort(column: Column<any>) {
     sortAscending = !sortAscending;
   } else {
     sortCol = column;
-    sortAscending = true;
+    sortAscending = column.info.initialSortDir ? column.info.initialSortDir !== 'descending' : true;
```
benoitf:
```suggestion
    sortAscending = column.info.initialOrder ? column.info.initialOrder !== 'descending' : true;
```

deboer-tim:
Rebased since the changes couldn't be auto-merged on main due to other PRs. While I was at it I added a couple brackets for consistent indenting, and added the sortOrder to the Volume page Size column so that there's actually something to test. Clicking on Size (once you've pulled them) will sort by largest first and now set the column sort indicator correctly (down).

I take back the comment I made in the description about this change _always_ keeping data sorted, as I've found a case in the Volumes page where data changes and yet there is no svelte trigger. I'll investigate separately, but this change is still at least a partial change in that direction: it wasn't the primary focus of this PR but more pages will stay sorted when data changes after this PR.

feloy (approved):

deboer-tim:
Additional test added.

PR #4931, opened 2023-11-21
feat: Deployments page

### LINKED ISSUE #4366: Kubernetes Deployments page

Deployments are the main way workloads are provisioned on Kubernetes, but we have no way to view them.

Add a new main nav item for Kubernetes Deployments, a list page, and a details page - similar to what we have for Containers or Pods. Users should be able to delete deployments.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Initial Deployments page. Creates a new route and icon in the main nav, plus
a regular page/table with deployment details. All columns from the design are
included, along with delete action and multiple deletion.

Notes:
- Deployments are not refreshing. It is likely unrelated to this PR, but
  should be resolved before this PR is merged. If you have existing
  deployments or Force Refresh everything else is working. (Fixed via #4989)
- In the design the Pods column are dots like the Pod list containers. This
  is deferred to a separate PR.
- We could sort by more columns... but not critical/can be done later.
- We should hide the nav item when there are no active Kubernetes providers. (PR updated to hide Deployments until you have a Kubernetes context)

Screenshot 2023-11-23 at 3 16 18 PM

Final major part of #4366.

### How to test this PR?

Set kube context to a cluster with a deployment.


deboer-tim:
Pushed new branch based on main with the final changes, ready for regular review.

lstocchi:
Played with it and it looks nice. I'll give my LGTM after we have https://github.com/containers/podman-desktop/issues/4974 otherwise it looks like it doesn't work. If you delete a deploy the table does not get refreshed, ... I'll give it a look for the ingress part

lstocchi:
Other thing is, should we display some message if there are no deployments? This is what i see now
image

deboer-tim:
Added new commit. Here's my view on where things stand:
- We should not merge until there is a fix for #4974. I've put up PR #4989 that solves the problem for deployments, but we need to merge that or a broader solution.
- The new commit hides the Deployment icon from the main nav when there are no Kubernetes contexts. I think this is a reasonable simple test for now, since we want to keep the UI simple when you aren't using Kubernetes, and when you are you could have a provider within Podman Desktop but could just have created a context externally. 
- The commit adds the missing empty screens for no deployments or all deployments filtered by search. There is no need
for a 'no provider' screen like other list pages because of the nav hiding.
- The commit also adds tests that were missing - basic UI and empty screens.

packages/renderer/src/lib/deployments/DeploymentColumnNamespace.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
```
jeffmaury: I'm wondering why we don't have a generic NamespaceColumn as it's likely that we will have a lot of Kubernetes objects that are namespace aware

deboer-tim: I meant to open an issue after your comment in the Volume (?) PR, and did now: #5002. I was not trying to support reusing/sharing columns in the first pass of the table component, and right now they're still sort of typed to the object. I think this a very logical thing to do (soon), but the code/complexity reduction is fairly small so I have been focusing on getting at least two tables merged so that we 'have this problem' first.

packages/renderer/src/lib/deployments/DeploymentUI.ts
```diff
@@ -18,9 +18,17 @@
...
   ready: number;
+  created: Date | undefined;
   age: string;
   selected: boolean;
+  conditions: DeploymentCondition[];
+}
+
+export interface DeploymentCondition {
+  type: string;
+  message: string | undefined;
```
benoitf: I think you need to review other places where we need to use the optional `?` vs `| undefined`
```
message? : string
```

packages/renderer/src/lib/deployments/DeploymentUI.ts
```diff
@@ -18,9 +18,17 @@
 
 export interface DeploymentUI {
   name: string;
+  status: string;
   namespace: string;
   replicas: number;
   ready: number;
+  created: Date | undefined;
```
benoitf:
```suggestion
  created?: Date;
```

packages/renderer/src/lib/deployments/DeploymentUI.ts
benoitf:
```suggestion
  message?: string;
```

packages/renderer/src/lib/deployments/deployment-utils.ts
```diff
@@ -29,16 +29,47 @@ export class DeploymentUtils {
...
+    const conditions = [];
+    if (deployment.status?.conditions) {
+      for (const con of deployment.status.conditions) {
+        const c: DeploymentCondition = {
+          type: con.type,
+          message: con.message,
+        };
+        conditions.push(c);
+      }
+    }
```
benoitf: should be simplified using map like approach or reduce like approach
```suggestion
conditions = (deployment.status.conditions ?? []).map(c => {
          type: con.type,
          message: con.message,
        }
);
```

packages/renderer/src/lib/deployments/deployment-utils.ts
```diff
@@ -29,16 +29,47 @@ export class DeploymentUtils {
...
     return {
       name: deployment.metadata?.name || '',
+      status: status,
       namespace: deployment.metadata?.namespace || '',
-      age: this.humanizeAge((deployment.metadata?.creationTimestamp || '').toString()),
+      created: deployment.metadata?.creationTimestamp,
+      age: deployment.metadata?.creationTimestamp
+        ? this.humanizeAge(deployment.metadata.creationTimestamp.toString())
```
benoitf: can use maybe the same approach then previous lines with ?? or || if there is no value
or we let humanizeAge accept undefined and it returns '' in that case

packages/renderer/src/lib/deployments/DeploymentUI.ts
deboer-tim: Sorry, one of the drawbacks of having multiple branches and PR reviews atm is that I thought this had been merged as part of the original deployment store and I'd go back and fix it later based on the service store review. 🤦🏼‍♂️ Fixed via new commit.

packages/renderer/src/lib/deployments/DeploymentUI.ts
deboer-tim: Fixed in new commit.

packages/renderer/src/lib/deployments/DeploymentUI.ts
deboer-tim: Fixed in new commit.

packages/renderer/src/lib/deployments/deployment-utils.ts
deboer-tim: Fixed in new commit.

packages/renderer/src/lib/deployments/deployment-utils.ts
deboer-tim: I tried a couple ways and went with humanizeAge() accepting undefined since it seemed like the shortest/simplest code to me.

deboer-tim:
I think I've addressed all the issues raised, but this PR is stagnating. Could I please get another review or comment on what else this needs?

cdrage (approved):
No issues with the code, this looks absolutely amazing.

The only feedback that I have is with the "Pods" column, I expect that if I click on the numbers, I'm taking to the Pods screen / be shown which Pods may be unavailable / not working / etc.

LGTM though, but maybe we should open up an issue to improve that column / UI design.

cdrage (changes requested):
After my review, I realized while I was playing more with it that there is no sorting for Pods or Age (I understand why we cant sort Conditions)?

Screenshot 2023-11-30 at 10 32 49 PM

deboer-tim:
Sorry, I should have explained - I was worried about both the size of this PR and it dragging on, so I was trying to find a reasonable spot to stop and why it says "Final major part of 4366" and not "Fixes".

You've listed exactly what I still planned to do: we should have pod linking and column sorting, and the design has the pods column like the pod view's container dots. If you think this isn't a good place to merge I can withdraw and include more of that here.

packages/renderer/src/lib/deployments/DeploymentColumnAge.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+import type { DeploymentUI } from './DeploymentUI';
+
+export let object: DeploymentUI;
+</script>
+
+<div class="text-xs text-gray-500">
+  {object.age}
+</div>
```
benoitf: now it seems you can use your custom rendering / simple column

packages/renderer/src/lib/deployments/DeploymentColumnNamespace.svelte
benoitf: as PR has been merged, now you can use the custom rendering and drop some .svelte files

packages/renderer/src/lib/deployments/DeploymentColumnStatus.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+import StatusIcon from '../images/StatusIcon.svelte';
+import DeploymentIcon from '../images/DeploymentIcon.svelte';
+import type { DeploymentUI } from './DeploymentUI';
+
+export let object: DeploymentUI;
+</script>
+
+<StatusIcon icon="{DeploymentIcon}" status="{object.status}" />
```
benoitf, line 9: missing a component test

packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+import EmptyScreen from '../ui/EmptyScreen.svelte';
+import DeploymentIcon from '../images/DeploymentIcon.svelte';
+</script>
+
+<EmptyScreen
```
benoitf, line 6: missing a component test

packages/renderer/src/lib/deployments/deployment-utils.ts
```diff
@@ -22,23 +22,48 @@ import type { DeploymentUI } from './DeploymentUI';
...
     // make it human friendly
     return humanizeDuration(uptimeInMs, { round: true, largest: 1 });
   }
 
+  refreshAge(deployment: DeploymentUI): string {
+    if (!deployment.created) {
+      return '';
+    }
+    // make it human friendly
+    return this.humanizeAge(deployment.created.toString());
```
benoitf: I'm wondering why we use the toString and do not send the `.created` field directly

humanizeAge should be smarter

packages/renderer/src/lib/deployments/deployment-utils.ts
```diff
@@ -22,23 +22,48 @@ import type { DeploymentUI } from './DeploymentUI';
...
+        status = 'DEGRADED';
+      }
+    }
     return {
       name: deployment.metadata?.name || '',
+      status: status,
       namespace: deployment.metadata?.namespace || '',
-      age: this.humanizeAge((deployment.metadata?.creationTimestamp || '').toString()),
+      created: deployment.metadata?.creationTimestamp,
+      age: this.humanizeAge(deployment.metadata?.creationTimestamp?.toString()),
```
benoitf: here I would not use `.toString`, send `.creationTImestamp`

packages/renderer/src/lib/deployments/DeploymentColumnAge.svelte
deboer-tim: Rebased on main to pick this up, and replaced Age and Namespace columns.

packages/renderer/src/lib/deployments/DeploymentColumnStatus.svelte
deboer-tim: Added tests for all columns.

packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
deboer-tim: Added tests for all columns.

packages/renderer/src/lib/deployments/deployment-utils.ts
deboer-tim: Yes, this was poor. Updated humanizeAge to accept Date.

packages/renderer/src/lib/deployments/deployment-utils.ts
deboer-tim: Fixed.

deboer-tim:
Rebased on main and responded to feedback. Added default name sorting and replaced Age/Namespace columns now that there are new options. Added tests, fixed humanizeAge(), and made the Age column sortable.

As per my previous comment, I felt like this was a reasonable place to merge. In a future PR I'd make the Pods column into dots and allow sorting there. Please re-review or let me know if you disagree. Thanks.

cdrage (approved):
LGTM! Amazing. Tested against a 2 node cluster and I was able to view deployments, sort correctly, have it update, etc. Didn't notice any issues with the UI or code.

Obviously we should merge after 1.6 / finalization of Icon?

deboer-tim:
The deployment icon is finalized (unless we get feedback in future), it's just the Ingress/Route icon which Conor was working on (and he pinged me a final one overnight).

Yes, given the size I would expect to wait for another approval + 1.6.

deboer-tim:
@jeffmaury or @benoitf ?

benoitf (approved):
I thought it was waiting the Informer stuff

deboer-tim:
A deployment store with a regular Kube watcher (same as we do for pods) was merged before Luca came up with the informer idea. So it works fine, but I've started a separate PR to refactor deployment store to informer.

Issue #4911, opened 2023-11-21
Apply naming pattern for onboarding ids


deboer-tim:
### Is your enhancement related to a problem? Please describe

The ids used during onboarding aren't especially obvious when you look at logs or telemetry data, e.g. picking on compose:

- checkComposeDownloaded - IMHO it checks for 'local install'
- startComposeDownload - doesn't actually download, it just tells the user it will
- composeDownloaded is the success equivalent of composeFailedDownload

### Describe the solution you'd like

We should define some simple patterns that make onboarding ids consistent and fairly obvious, and apply it to both podman and compose extensions. It should have:

- clear differentiation between (a pattern for) ids for content vs command
- similar pattern for success vs failure screens (e.g. '.success' and '.failure')
- pattern for welcome or summary content pages

Do we need the extension in the id? (e.g. do compose ids need to include 'compose' or is that always obvious in context and could be removed?)

It should be obvious from the ids what the content shows or command does.

## Rules
- steps displaying information are suffixed with `View`
- steps executing a command are suffixed with `Command`
- steps with a state='failed' are suffixed with `Failure`
- steps with a state='completed' are suffixed with `Success`

## Acceptance criteria

- [x] modify Podman onboarding (#4935)
- [x] Modify Compose onboarding (#4948)
- [ ] Update documentation to describe the rules (blocked by PR 4837)
- [x] Raise warnings when extension is loaded if rules are not respected (#4951)

PR #4837, opened 2023-11-16
docs: explain how to create an onboarding workflow for an extension (#3440)

lstocchi:
### What does this PR do?

It adds the documentation to show how to add an onboarding workflow to an extension

### Screenshot/screencast of this PR

N/A

### What issues does this PR fix or reference?

it resolves #3440 

### How to test this PR?

N/A

PR #4935, opened 2023-11-22
refactor: human readable podman onboarding ids

feloy:

### What does this PR do?

This PR changes the ids on the Podman Onboarding steps (and the ids of their associated commands and context values), so they are easily readable in the telemetry tool when they are sent to the telemetry.

- steps displaying information are suffixed with `View`
- steps executing a command are suffixed with `Command`
- steps with a state='failed' are suffixed with `Failure`
- steps with a state='completed' are suffixed with `Success`

Fixes partially #4911 

### How to test this PR?

Play the onboarding for Podman extension.

### DIFF
--- a/extensions/podman/package.json
+++ b/extensions/podman/package.json
@@ -17,11 +17,11 @@
         "title": "Podman: Socket Compatibility Mode"
       },
       {
-        "command": "podman.onboarding.checkPodmanInstalled",
+        "command": "podman.onboarding.checkInstalledCommand",
         "title": "Podman: Check podman installation"
       },
       {
-        "command": "podman.onboarding.checkPodmanRequirements",
+        "command": "podman.onboarding.checkRequirementsCommand",
         "title": "Podman: Check system requirements to install podman"
       },
       {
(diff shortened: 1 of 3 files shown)


extensions/podman/package.json
```diff
@@ -162,29 +162,29 @@
...
         },
         {
-          "id": "startPodmanInstallation",
+          "id": "welcomeInstallationView",
           "title": "We could not find Podman. Let's install it!",
           "when": "onboardingContext:podmanIsNotInstalled"
         },
         {
-          "id": "checkPodmanRequirements",
+          "id": "checkRequirementsCommand",
```
jeffmaury, line 178: Shouldn't end with View like following one ?

benoitf:
hello, I'm wondering if the suffix could not be added automatically ?

if step has a command --> add Command in telemetry, if it's upon a failure, append Failed, if a success append Success, etc
so the extension does not really care about a pattern

or if it's a recommended pattern/guidelines, if when loading an extension we should report these warnings

feloy:
> hello, I'm wondering if the suffix could not be added automatically ?
> 

> if step has a command --> add Command in telemetry, if it's upon a failure, append Failed, if a success append Success, etc so the extension does not really care about a pattern

This is what is done for step index in #4736 (adding the step index at the beginning of the id). But if a developer wants to match telemetry ids with the ones in package.json, I think it would be harder if we remove these suffixes.

> 
> or if it's a recommended pattern/guidelines, if when loading an extension we should report these warnings

Yes, it could be interesting to do this IMO.

extensions/podman/package.json
feloy: This step executing a command (even if there is a title and some description), it should be considered as a command.

benoitf (approved):
would be nice to have a follow-up issue/guidelines in the onboarding documentation

https://github.com/containers/podman-desktop/pull/4837

and to raise warnings when loading the extension

feloy:
> would be nice to have a follow-up issue/guidelines in the onboarding documentation
> 
> #4837
> 
> and to raise warnings when loading the extension

I've added acceptance criteria to #4911 4911 to track these points

deboer-tim:
Are steps with 'failed' or 'success' always the result of a command? I'm wondering if there should be another pattern where the step after xCommand is either xCommandSuccess or xCommandFailure.

feloy:
> Are steps with 'failed' or 'success' always the result of a command? I'm wondering if there should be another pattern where the step after xCommand is either xCommandSuccess or xCommandFailure.

This pattern can be seen for the Compose onboarding (see other PR: installSystemWideCommand / installSystemWideFailure / installSystemWideSuccess). But here, because the installation process is different in Linux and in !Linux, we cannot have this pattern (install is done either with installOnLinuxInstructionsView or with autoInstallCommand).

deboer-tim:
Ah, that makes sense, and why it can't be a forced/general pattern.

deboer-tim (approved):
LGTM, much easier to understand what each command does.

+1 on the followup issues opened, that will help as well.

PR #4937, opened 2023-11-22
chore: remove patternfly dependency

benoitf:
### What does this PR do?
remove the patternfly library

usage was removed but somehow the dependency was still there

N/A

https://github.com/containers/podman-desktop/issues/1617

### How to test this PR?

check that everything is still working
(as we're not importing the library anymore it should be easy)


deboer-tim (approved):
LGTM

PR #4945, opened 2023-11-22
chore: replace zip-local by adm-zip 

### LINKED ISSUE #4938: replace zip-local library

This library is used in Podman Desktop but now with vite5 we have an error like

```
Cannot read properties of null (reading 'module')
```

it appears that the culprit is coming from this module.
As it is no longer widely downloaded and maintained, I suggest to use another library

we need both unzip and zip (for example for extensions)

so while `unzipper` library was interesting it does not do the 'zip' part

https://www.npmjs.com/package/adm-zip looks like still maintained and has a lot of downloads each week

unzipper, decompress, jszip, pure nodejs etc

### Additional context

_No response_

benoitf:
### What does this PR do?
replace zip-local broken with vite5 with another more popular library

2 commits:
one for main package
another one for the extensions

fixes https://github.com/containers/podman-desktop/issues/4938

### How to test this PR?

build the cdix file of DevSandbox extension for example, copy it to podman desktop folder:

```bash
cp redhat-sandbox.cdix ~/.local/share/containers/podman-desktop/plugins-scanning
```

then see if sandbox extension has been unpacked in `~/.local/share/containers/podman-desktop/unpacked/` and is loaded

and for extensions:
check that in production mode or in development mode you're able to use/see the extensions

### DIFF
--- a/extensions/compose/package.json
+++ b/extensions/compose/package.json
@@ -160,11 +160,10 @@
     "shell-path": "^3.0.0"
   },
   "devDependencies": {
-    "7zip-min": "^1.4.3",
+    "adm-zip": "^0.5.10",
     "@podman-desktop/api": "^0.0.1",
     "mkdirp": "^3.0.1",
     "vite": "^4.5.0",
-    "vitest": "^0.34.6",
-    "zip-local": "^0.3.5"
+    "vitest": "^0.34.6"
   }
 }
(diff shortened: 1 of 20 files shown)


feloy (approved):

cdrage (approved):

benoitf:
rebased due to yarn.lock conflict

PR #4948, opened 2023-11-22
refactor: human readable compose onboarding ids

feloy:

### What does this PR do?

This PR changes the ids on the Compose Onboarding steps (and the ids of their associated commands), so they are easily readable in the telemetry tool when they are sent to the telemetry.

Partially #4911

### How to test this PR?

Play the onboarding for Compose extension.

### DIFF
--- a/extensions/compose/package.json
+++ b/extensions/compose/package.json
@@ -21,11 +21,11 @@
         "title": "Compose: Checks..."
       },
       {
-        "command": "compose.onboarding.checkComposeDownloaded",
+        "command": "compose.onboarding.checkDownloadedCommand",
         "title": "Compose: Check Compose Downloaded"
       },
       {
-        "command": "compose.onboarding.downloadCompose",
+        "command": "compose.onboarding.downloadCommand",
         "title": "Compose: Download Compose"
       }
     ],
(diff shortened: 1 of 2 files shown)


extensions/compose/package.json
```diff
@@ -68,23 +68,23 @@
...
-          "command": "compose.onboarding.downloadCompose",
+          "command": "compose.onboarding.downloadCommand",
           "completionEvents": [
-            "onCommand:compose.onboarding.downloadCompose"
+            "onCommand:compose.onboarding.downloadCommand"
           ]
         },
         {
-          "id": "composeFailedDownload",
+          "id": "downloadFailure",
```
benoitf, line 81: is it not a View ?

extensions/compose/package.json
```diff
@@ -68,23 +68,23 @@
...
           "title": "Failed Downloading Compose",
           "when": "onboardingContext:composeIsNotDownloaded == true",
           "state": "failed"
         },
         {
-          "id": "composeDownloaded",
+          "id": "downloadSuccessfulView",
```
benoitf, line 87: here it looks strange
it should not have a state of "completed" ?

and be only "downloadSuccessful" but not a view ?

I mean, it looks like either both of them are views or not a view

feloy: I don't think so, because this is an intermediary step. The onboarding first downloads the binary locally, then copies it system-wide. If downloading the binary fails, this is a failure (we will retry...), but it it succeeds, other steps are following. If we were setting a `completed` state here, the onboarding would stop immediately, and it would never copy it system-wide.

benoitf (approved):

cdrage (approved):

PR #4949, opened 2023-11-22
nit: table column overlapping and aria roles

deboer-tim:
### What does this PR do?

These changes are already proposed in PRs #4844 and #4841 respectively, but since those PRs are taking a while in review and these changes are common and required for other PRs I'm breaking them out here:
- Add missing aria roles to the first two cells in the column. (Not only is this correct, but some e2e tests are finding columns by counting aria cells).
- Add max-w-full to each cell, so that the overflow-hidden applies correctly and cells that are too narrow are cropped instead of overlapping.

N/A

Issues raised during reviews of the mentioned PRs.

### How to test this PR?

See issues in the PRs. e.g. to test overflow try #4841 with a long volume name and narrow columns, with or without this change.

### DIFF
--- a/packages/renderer/src/lib/table/Table.spec.ts
+++ b/packages/renderer/src/lib/table/Table.spec.ts
@@ -18,7 +18,7 @@
 
 import '@testing-library/jest-dom/vitest';
 import { test, expect } from 'vitest';
-import { fireEvent, render, screen } from '@testing-library/svelte';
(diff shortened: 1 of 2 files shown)


benoitf (approved):

benoitf:
unit test to grab the cell role ?

feloy (approved):

deboer-tim:
Added aria role to first column (future section expander) to keep things consistent, and added tests.

PR #4951, opened 2023-11-22
chore: warning when onboarding ids do not respect specific rules

feloy:

### What does this PR do?

Checks if the ids of the steps of onboardings (sent to Telemetry) respect some specific rules, to make them more easily readable. In case of non respect, warnings are displayed in the backend console.

Partially #4911

### DIFF
--- a/packages/main/src/plugin/onboarding-registry.spec.ts
+++ b/packages/main/src/plugin/onboarding-registry.spec.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { beforeEach, describe, expect, expectTypeOf, test, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, expectTypeOf, test, vi } from 'vitest';
 import { OnboardingRegistry } from './onboarding-registry.js';
 import type { ConfigurationRegistry } from './configuration-registry.js';
 import type { AnalyzedExtension } from './extension-loader.js';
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/onboarding-registry.ts
```diff
@@ -142,4 +143,34 @@ export class OnboardingRegistry {
...
+   *
+   * In case of a rule not respected, a warning is displayed in the console.
+   */
+  checkIdsReadability(extension: AnalyzedExtension, onboarding: Onboarding) {
+    const warn = (msg: string) => {
+      console.warn(`${extension.name} Onboarding: ${msg}`);
+    };
+    onboarding.steps.forEach(step => {
+      const id = step.id;
+      const isCommand = !!step.command || !!step.component;
```
feloy, line 159: Here, I'm not sure if it is reasonable to consider a screen with a "component" to be a command.

benoitf (commented):
image


I got reports when trying 👍 

about the format; maybe pickup the same than when an extension is performing a log

```
main ↪️ [redhat-sandbox] starting extension openshift-sandbox
main ↪️ [redhat-sandbox] Configfile location /Users/benoitf/.kube/config
```
so put the extension with square brackets and use the full name like `podman-desktop.compose`


Then on the log message, I think I would rework with the 'what I'm missing first' (and explanation later)

-->
`Missing 'Command' suffix for step checkComposeDownloaded that defines a command`

instead of 
`step checkComposeDownloaded defines a command, its id should end with 'Command'`

`step ${id} should end with 'View' had a different warning message template than the other


you should add unit tests so in case we tweak the isCommand, isFailedState, etc we've no-regression tests

packages/main/src/plugin/onboarding-registry.ts
benoitf: hard to tell but as long as we define tests, we'll able to sort it out

benoitf (approved):
overall approving

deboer-tim (approved):
Ditto, would be good to just change the log format as Florent suggested.

packages/main/src/plugin/onboarding-registry.spec.ts
```diff
@@ -47,131 +48,213 @@ const context = new Context(apiSender);
...
+        label: 'Installation successful',
+        title: 'Installation successful',
+        state: 'completed' as OnboardingState,
+      },
+    ],
+    enablement: 'true',
+  };
+
+  const consoleWarnMock = vi.fn();
+  console.warn = consoleWarnMock;
```
benoitf: IMHO it should be part of a beforeEach, afterEach where in after step we reset to the previous console.warn object to not mess up further tests

benoitf (approved):
approving again

PR #4953, opened 2023-11-22
Refactor List UI components

feloy:
### What does this PR do?

- moves layout information into NavPage, instead of having to duplicate the TailWind classes in places where the NavPage component is used.

It takes advantage of `$$slots` (https://github.com/sveltejs/svelte/pull/4602) to move layout into component when the content is optional.


benoitf (approved):
always 👍 when we remove lines of code

dgolovin (approved):
💯 for removing additional lines

PR #4955, opened 2023-11-22
feat: deployment store

deboer-tim:
### What does this PR do?

Follows the same pattern as pods, images, volumes, etc to define a deployment store that's reactive to the expected events, a UI type, and a util class that maps to the UI type.

Re: index.ts - I copied from Luca, who likely copied from elsewhere, and we both missed updating one type. :)

We will definitely need additional properties mapped from V1Deployment to DeploymentUI when we add the UI, this PR is purely to get the store, UI object, and basic scaffolding in place.

N/A

Part of #4366.

### How to test this PR?

Run test.

### DIFF
--- a/packages/preload/src/index.ts
+++ b/packages/preload/src/index.ts
@@ -68,6 +68,7 @@ import type {
   Cluster,
   Context,
   V1ConfigMap,
+  V1Deployment,
   V1Ingress,
   V1NamespaceList,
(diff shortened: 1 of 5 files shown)


benoitf (approved):

cdrage (approved):
LGTM! Minor linting issues, but code looks good.

Issue #4367, opened 2023-10-16
Kubernetes Ingress / Routes page


deboer-tim:
### Is your feature request related to a problem? Please describe

The podify flow creates ingress or Routes on Kubernetes, but there is no way to view them or clean them up from within Podman Desktop, and hence rerunning the flow doesn't work.

### Describe the solution you'd like

Add a new main nav item for Kubernetes Ingress & Routes, a list page, and details page - similar to what we have for Containers or Pods. Users should be able to delete either.

deboer-tim:
Design issue: #4404

PR #4958, opened 2023-11-23
feat: ingresses/routes store

lstocchi:
### What does this PR do?

Adds ingress and route store needed for #4367 

N/A

it is part of #4367 

### How to test this PR?

run tests


deboer-tim (approved):
Code is as expected, LGTM, and test pass, so approved.

benoitf (approved):
Looks like stores are untested

Issue #4968, opened 2023-11-23
Close Dev Tools window for all E2E tests


odockal:
### Is your enhancement related to a problem? Please describe

It happens that when running the e2e tests, the electron application is opened with dev tools window opened by default. Sometime it prevents to properly find elements or interact with them during the testing.

### Describe the solution you'd like

Close the dev tools as part of `podman-desktop-runner.ts` `start` method, as this functionality is useful for all e2e tests as we can get better screenshots as well.

### Additional context

Failing Smoke e2e test on PR check: https://github.com/containers/podman-desktop/actions/runs/6950062212/job/18971018336?pr=4844

Example of the test failure screenshot:
![Check_image_details_failure](https://github.com/containers/podman-desktop/assets/19164299/d4fefd1a-a2ea-4cd1-88dc-6593d3287df0)

odockal:
@deboer-tim FYI. PR will follow asap.

benoitf:
I would not close it, just keep it in a separate window

odockal:
@benoitf Proper issue to handle this enhancement: https://github.com/containers/podman-desktop/issues/5031.

PR #4969, opened 2023-11-23
chore(tests,ci): close the Dev Tools window when starting app. during…

odockal:
… E2E tests

### What does this PR do?
It closes the dev tools that are opened during the e2e tests. Electron app is opened with dev tools on the side by default, I guess, because of the development mode the tests run in. It removes the redundancy in welcome-page tests file and add one more check for devtools window state.

welcome-page-display

#4968 

### How to test this PR?
Run the E2E tests (`yarn test:e2e:smoke`) and checks the `tests/output/screenshots` folder for `png` files

### DIFF
--- a/tests/src/runner/podman-desktop-runner.ts
+++ b/tests/src/runner/podman-desktop-runner.ts
@@ -21,6 +21,12 @@ import { _electron as electron } from '@playwright/test';
 import { join } from 'node:path';
 import type { BrowserWindow } from 'electron';
 
+type WindowState = {
+  isVisible: boolean;
+  isDevToolsOpened: boolean;
+  isCrashed: boolean;
+};
+
 export class PodmanDesktopRunner {
   private _options: object;
   private _running: boolean;
@@ -60,6 +66,14 @@ export class PodmanDesktopRunner {
       console.log(`STDERR: ${data}`);
     });
 
+    // Evaluate that the main window is visible
+    // at the same time, the function also makes sure that event 'ready-to-show' was triggered
+    const windowState = await this.getBrowserWindowState();
+    console.log(`Application Browser's window is visible: ${windowState.isVisible}`);
+
+    // close dev tools windows
+    await this.closeDevTools();
+
     return this._page;
   }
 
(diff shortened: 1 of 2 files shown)


benoitf:
we should have a dedicated window for devtools instead no ? 

For example on my side I'm always using a separate window in development mode

https://github.com/containers/podman-desktop/blob/6339a6ffed6379097f0c9e9e0f5229aee03687b7/packages/main/src/mainWindow.ts#L98

if we provide `{mode: undocked}` it should work (we would have two windows, record stuff from the two windows, etc so better for debugging/troubleshooting)

odockal:
> we should have a dedicated window for devtools instead no ?
> 
> For example on my side I'm always using a separate window in development mode
> 
> https://github.com/containers/podman-desktop/blob/6339a6ffed6379097f0c9e9e0f5229aee03687b7/packages/main/src/mainWindow.ts#L98
> 
> if we provide `{mode: undocked}` it should work (we would have two windows, record stuff from the two windows, etc so better for debugging/troubleshooting)

Yes, I agree... but... 
1. Even if the dev tools part is opened. What information can we take out of it and use? Dev console, right? It is not opened by default and we do not know how to do it. So we would need to figure that out. On the other hand, isn't exactly that information in the troubleshooting? 
2. I am not sure about how overlapping in the CI env., taking of the screenshots and recording of the docked window works right now. Would have to investigate. It takes times. 

I would be for creating an follow up issue right now. @benoitf wdyt?

benoitf:
I would not open another issue, it's just that for debugging CI jobs I need the devtools output, so I would not close/hide it

> Even if the dev tools part is opened. What information can we take out of it and use? Dev console, right? It is not opened by default and we do not know how to do it. So we would need to figure that out. On the other hand, isn't exactly that information in the troubleshooting?

it is opened by default 

You need to go to troubleshooting explicitely while here you have the information steaming continously

> I am not sure about how overlapping in the CI env., taking of the screenshots and recording of the docked window works right now. Would have to investigate. It takes times.

I guess it will work the same way, just that we will have two maximized window rather than two small width windows

odockal:
> > Even if the dev tools part is opened. What information can we take out of it and use? Dev console, right? It is not opened by default and we do not know how to do it. So we would need to figure that out. On the other hand, isn't exactly that information in the troubleshooting?

> it is opened by default

I mean, Console tab is not opened by default. It is Elements that are open, and so you can just see DOM...

benoitf:
you're correct 👍 

ok so let's try this PR but really try to make it a separate window with focus being on the console tab

benoitf (approved):

deboer-tim (approved):
This is working for me and potentially fixes the e2e test problem in #4844, so I am approving. We still need to hear back from @benoitf or ... ok, I got the notification email, let's merge. :)

odockal:
@benoitf It seems that there might be much better options: https://playwright.dev/docs/api/class-tracing. I am looking swiftly into this as it could be better alternative than manipulating with various browser windows (not saying it wont work).

benoitf:
we already send the log to the window so we see output in the log file/stdout

odockal:
@benoitf that is true.

PR #4981, opened 2023-11-24
fix: add path prop for route object

lstocchi:
### What does this PR do?

a route can use a port but also a path https://docs.openshift.com/container-platform/4.14/networking/routes/route-configuration.html#nw-path-based-routes_route-configuration
this PR adds support for the path property

N/A

it is part of #4367 

### How to test this PR?

1. run tests

### DIFF
--- a/packages/renderer/src/lib/ingresses-routes/ingress-route-utils.spec.ts
+++ b/packages/renderer/src/lib/ingresses-routes/ingress-route-utils.spec.ts
@@ -90,7 +90,7 @@ test('expect basic UI conversion for route with port', async () => {
   expect(routeUI.to.name).toEqual(route.spec.to.name);
 });
 
-test('expect basic UI conversion for route with port', async () => {
+test('expect basic UI conversion for route with path', async () => {
   const route = {
     metadata: {
       name: 'my-route',
--- a/packages/renderer/src/lib/ingresses-routes/ingress-route-utils.spec.ts
+++ b/packages/renderer/src/lib/ingresses-routes/ingress-route-utils.spec.ts
@@ -63,7 +63,7 @@ test('expect basic UI conversion for ingress', async () => {
   expect(ingressUI.rules).equal(ingress.spec?.rules);
 });
 
-test('expect basic UI conversion for route', async () => {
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/ingresses-routes/ingress-route-utils.spec.ts
```diff
@@ -84,7 +84,33 @@ test('expect basic UI conversion for route', async () => {
...
   expect(routeUI.namespace).toEqual('test-namespace');
   expect(routeUI.host).toEqual(route.spec.host);
-  expect(routeUI.port).toEqual(route.spec.port.targetPort);
+  expect(routeUI.port).toEqual(route.spec.port?.targetPort);
+  expect(routeUI.path).toBeUndefined();
+  expect(routeUI.to.kind).toEqual(route.spec.to.kind);
+  expect(routeUI.to.name).toEqual(route.spec.to.name);
+});
+
+test('expect basic UI conversion for route with port', async () => {
```
feloy:
```suggestion
test('expect basic UI conversion for route with path', async () => {
```

packages/renderer/src/lib/pod/DeployPodToKube.svelte
```diff
@@ -569,7 +569,7 @@ function updateKubeResult() {
               <ul class="list-disc list-inside">
                 {#each createdRoutes as createdRoute}
                   <li class="pt-2">
-                    Port {createdRoute.spec.port.targetPort} is reachable with route
+                    Port {createdRoute.spec.port?.targetPort} is reachable with route
```
feloy, line 572: I think you will need to wrap this text with a `{#if createdRoute.spec.port}`, or the text will be meaningless

lstocchi: we are the one that creates the route when deploying a pod so i expect that a port is always there as before. This is just a fix bc lint was complaining.

I added support for the path because when listing all routes from a cluster  there could be some resources not created by us which just uses path.

feloy (approved):

benoitf (approved):

Issue #4368, opened 2023-10-16
Kubernetes Services page


deboer-tim:
### Is your feature request related to a problem? Please describe

The podify flow creates Kubernetes services, but there is no way to view them or clean them up from within Podman Desktop, and hence rerunning the flow doesn't work.

### Describe the solution you'd like

Add a new main nav item for Kubernetes Services, a list page, and details page - similar to what we have for Containers or Pods. Users should be able to delete services.

cdrage:
I humbly request that if I can take this up for this upcoming sprint! I'd like to finally see Services on there / port mapping. 

I've been working with https://github.com/containers/podman-desktop/issues/4576 so would like to see some of the design mockup implemented.

PR #4984, opened 2023-11-24
chore: service icon

deboer-tim:
### What does this PR do?

Adding icon for services, in both outline and solid form, as required by upcoming issue.

Mockup for context:

Screenshot 2023-11-24 at 9 12 11 AM

Part of #4368.

### How to test this PR?

Screenshot above, or replace an existing icon in dev mode.

### DIFF
--- a/packages/renderer/src/lib/images/ServiceIcon.svelte
+++ b/packages/renderer/src/lib/images/ServiceIcon.svelte
@@ -4,7 +4,7 @@ export let solid = false;
 
 let boxStyle = 'fill: currentColor;';
 if (!solid) {
-  boxStyle = 'stroke-miterlimit: 10; stroke: currentColor;';
+  boxStyle = 'stroke: currentColor;';
 }
 </script>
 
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/images/ServiceIcon.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+export let size = '1em';
+export let solid = false;
+
+let boxStyle = 'fill: currentColor;';
+if (!solid) {
+  boxStyle = 'stroke-miterlimit: 10; stroke-width: 1; stroke: currentColor; stroke-opacity: 1;';
```
benoitf: can't we use tailwind classes ?  `stroke-1` `stroke-current`

benoitf (approved):

packages/renderer/src/lib/images/ServiceIcon.svelte
deboer-tim: Updates:
- I realized stroke-width and stroke-opacity were just set to the defaults, so I removed those (and I bet we could in several other svgs too).
- Changed "fill: none" to fill-none class.
- Realized miterlimit wasn't having any visible effect even at huge sizes, so removed (again, could probably do in other icons too).
- Using fill-current and stroke-current don't work (goes to black), so I left those as styles.

I suspect the 'current' styles need to be defined in our palette or something, so IMHO this is as far as we can go for now.

PR #4986, opened 2023-11-24
feat: show canceled analyzers + filter on analyzers

feloy:

### What does this PR do?

This PR adds information on the Image Checker page, to display which provider has been canceled when the user clicks the Cancel button before the providers finish.

The PR also adds toggle button to Providers, to filter the results on providers.

https://github.com/containers/podman-desktop/assets/9973512/ab2d2096-9d35-4c1c-84f3-d792934ddbe9

References #4707

### How to test this PR?

You can install a fake image checker from the image at quay.io/phmartin/fake-image-checker:v0.0.3 (sources: https://github.com/feloy/podman-desktop-extension-fake-image-checker)

### DIFF
--- a/packages/renderer/src/lib/ui/ProviderResultPage.svelte
+++ b/packages/renderer/src/lib/ui/ProviderResultPage.svelte
@@ -11,6 +11,7 @@ import {
 } from '@fortawesome/free-solid-svg-icons';
 import type { ProviderUI } from './ProviderResultPage';
 import Spinner from './Spinner.svelte';
+import SlideToggle from './SlideToggle.svelte';
 
 interface CheckUI {
   provider: ImageCheckerInfo;
@@ -20,6 +21,9 @@ interface CheckUI {
 export let providers: ProviderUI[] = [];
 
 export let results: CheckUI[] = [];
+
+$: checkedProviders = getCheckedProviders(providers);
+$: filteredResults = getFilteredResults(results, checkedProviders);
 
 function getIcon(check: ImageCheck): IconDefinition {
   if (check.status === 'success') {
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/ui/ProviderResultPage.svelte
```diff
@@ -35,6 +39,23 @@ function getIcon(check: ImageCheck): IconDefinition {
...
+    if (p.info.id === id) {
+      p.checked = checked;
+    }
+    return p;
+  });
+}
+
+function getFilteredResults(results: CheckUI[], checkedProviders: string[]): CheckUI[] {
+  return results.filter(r => checkedProviders.includes(r.provider.id));
+}
```
benoitf, line 58: it makes me feel that maybe we should have used a store on the renderer side, so when we have results, we update the store or we use a derived storer rather than calling getFilteredResults

and we use `$object` rather than `object` in the loops so it's in reactive mode (--> less code, no subscribe/unsubscribe)

benoitf (approved):

jeffmaury:
Can't test it as I can't build the fake extension: I had to change the podman-desktop API version to next but then watch does not work:
```

[09:38:24] Starting compilation in watch mode...

src/extension.ts:19:1 - error TS1202: Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} from "mod"', 'import d from "mod"', or another module format instead.

19 import EventEmitter = require('events');
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

[09:38:25] Found 1 error. Watching for file changes.
```

feloy:
> Can't test it as I can't build the fake extension: I had to change the podman-desktop API version to next but then watch does not work:

You can deploy the extension from the image at quay.io/phmartin/fake-image-checker:v0.0.3, without having to build it with the desktop

jeffmaury (approved):
LGTM

PR #4987, opened 2023-11-24
feat: service store

deboer-tim:
### What does this PR do?

Follows the same pattern as pods, images, volumes, etc to define a service store that's reactive to the expected events, a UI type, and a util class that maps to the UI type.

We will likely need additional properties mapped from V1Service to ServiceUI when we add the UI, this PR is to get the store, UI object, and basic scaffolding in place.

N/A

Part of #4368.

### How to test this PR?

`yarn test:renderer`

### DIFF
--- a/packages/renderer/src/lib/service/ServiceUI.ts
+++ b/packages/renderer/src/lib/service/ServiceUI.ts
@@ -20,7 +20,7 @@ export interface ServiceUI {
   name: string;
   status: string;
   namespace: string;
-  created: Date | undefined;
+  created?: Date;
   age: string;
   conditions: Condition[];
   selected: boolean;
@@ -28,5 +28,5 @@ export interface ServiceUI {
 
 export interface Condition {
   type: string;
-  message: string | undefined;
+  message?: string;
 }
--- a/packages/renderer/src/lib/service/ServiceUI.ts
+++ b/packages/renderer/src/lib/service/ServiceUI.ts
@@ -1,0 +1,32 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface ServiceUI {
+  name: string;
+  status: string;
+  namespace: string;
+  created: Date | undefined;
+  age: string;
+  conditions: Condition[];
+  selected: boolean;
+}
+
+export interface Condition {
+  type: string;
+  message: string | undefined;
+}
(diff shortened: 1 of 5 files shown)


benoitf (approved):

packages/renderer/src/lib/service/ServiceUI.ts
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface ServiceUI {
+  name: string;
+  status: string;
+  namespace: string;
+  created: Date | undefined;
```
benoitf: Should be with a ?

created? : Date

packages/renderer/src/lib/service/ServiceUI.ts
```diff
@@ -0,0 +1,32 @@
...
+  age: string;
+  conditions: Condition[];
+  selected: boolean;
+}
+
+export interface Condition {
+  type: string;
+  message: string | undefined;
```
benoitf: message?: string

packages/renderer/src/lib/service/service-utils.ts
```diff
@@ -0,0 +1,62 @@
...
+        const c: Condition = {
+          type: con.type,
+          message: con.message,
+        };
+        conditions.push(c);
+      }
+    }
+
+    return {
+      name: service.metadata?.name || '',
```
benoitf: just a FYi should use nullish coalescing ?? Rather than || ( OR)

packages/renderer/src/lib/service/ServiceUI.ts
deboer-tim: Fixed.

packages/renderer/src/lib/service/ServiceUI.ts
deboer-tim: Fixed.

packages/renderer/src/lib/service/service-utils.ts
deboer-tim: Fixed.

packages/renderer/src/lib/service/ServiceUI.ts
benoitf: I don't see the fix ?

deboer-tim: Forgot to actually hit push... there now

packages/renderer/src/lib/service/service-utils.ts
```diff
@@ -0,0 +1,62 @@
...
+  refreshAge(service: ServiceUI): string {
+    if (!service.created) {
+      return '';
+    }
+    // make it human friendly
+    return this.humanizeAge(service.created.toString());
+  }
+
+  getServiceUI(service: V1Service): ServiceUI {
+    const conditions = [];
```
benoitf: conditions = (service.status.conditions ?? []).map(c => { type : c.type, message: c.message})

deboer-tim: Added in new commit.

packages/renderer/src/stores/services.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  // do not fetch until extensions are all started
+  return readyToUpdate;
+}
+
+export const services: Writable<V1Service[]> = writable([]);
+
+export const searchPattern = writable('');
+
+export const filtered = derived([searchPattern, services], ([$searchPattern, $serviceInfos]) =>
```
feloy, line 44: For containers, we decided to move the filtering of the state. For pods, it is still done in the state. Do we want to do filtering in a consistent way? Which one?

deboer-tim: FYI I am intentionally letting this chill for a couple days to see where #5029 goes, since I think the informer pattern is a better approach for Kubernetes objects long term. Assuming that gets merged I would keep the files here in /services but rebase the store.

deboer-tim:
Withdrawing for now. To me the informer pattern in #5029 looks better and we should make it consistent for Kubernetes objects. Assuming that gets approved I would wait for it to merge and then put up new PR(s) for Deployment and Service stores.

Issue #4983, opened 2023-11-24
E2E tests instability: electronApplication.browserWindow: Execution context was destroyed, most likely because of a navigation


odockal:
### Bug description

The patch for #4968 was merged and it introduced probably a regression. E2E main tests started to fail. It is very interesting that Smoke E2E suite is still passing without any issues. 

Error differs a bit when run locally, run the nature could be the same.

Playwright test log (this one differs)
```
Error: electronApplication.browserWindow: Execution context was destroyed, most likely because of a navigation.
```
Console output:
```
main ↪️ System ready. Loading extensions...
STDERR: Error creating window Error: ERR_FAILED (-2) loading 'file:///home/odockal/git/podman-desktop/packages/renderer/dist/index.html'
    at rejectAndCleanup (node:electron/js2c/browser_init:2:84559)
    at WebContents.stopLoadingListener (node:electron/js2c/browser_init:2:84946)
    at WebContents.emit (node:events:514:28) {
  errno: -2,
  code: 'ERR_FAILED',
  url: 'file:///home/odockal/git/podman-desktop/packages/renderer/dist/index.html'
}
```

### Operating system

Linux

### Installation Method

Other

### Version

next (development version)

### Relevant log output

```shell
main ↪️ System ready. Loading extensions...
STDERR: Error creating window Error: ERR_FAILED (-2) loading 'file:///home/odockal/git/podman-desktop/packages/renderer/dist/index.html'
    at rejectAndCleanup (node:electron/js2c/browser_init:2:84559)
    at WebContents.stopLoadingListener (node:electron/js2c/browser_init:2:84946)
    at WebContents.emit (node:events:514:28) {
  errno: -2,
  code: 'ERR_FAILED',
  url: 'file:///home/odockal/git/podman-desktop/packages/renderer/dist/index.html'
}
```

odockal:
The problem seems to be introduced with [remove playwright dep]
and eventually partially fixed for e2e smoke tests with [add playwright install] step in gha file. Now @ScrewTSW is working on the fix for all e2e tests.

[remove playwright dep]: https://github.com/containers/podman-desktop/commit/56d7d555c2ff1851223ee6ba81a6b14e8a74ac9b
[add playwright install]: https://github.com/containers/podman-desktop/commit/466340a8fbb6bace9d16f02bf320a1c5e1c2723f

odockal:
It might be that the fix mentioned above is not directly solving this issue. 
[...]

PR #4988, opened 2023-11-24
chore: env. CLOSE_DEVTOOLS can enforce devtools closed during test

odockal:
### What does this PR do?
Introduces an option how to enforce podman desktop not to open dev tools window. Especially during testing.
This bring only a condition in `packages/main/src/mainWindow.ts` that checks for env. var. CLOSE_DEVTOOLS, if it is defined and is `'true'`, then `browserWindow?.webContents.openDevTools();` is not called during application start up. This applies only in DEV mode which can be used mainly during `yarn watch` and during running the e2e tests. next pr will bring default setup of this env. var in `tests/src/runner/podman-desktop-runner.ts` during application start up. Resulting in a testing window without half of the space covered with dev tools.

#4983 #4968 

### How to test this PR?
you can run tests without any change. `yarn test:e2e:smoke`. Check for the app during test run and verify that there is a dev tools opened. Then add `cross-env CLOSE_DEVTOOLS=true` at a beginning of a npm task `test:e2e:smoke:ruin` in `package.json`, task: `. If you then execute the tests, you should obeserve that dev tools were closed during testing.

### DIFF
--- a/packages/main/src/mainWindow.ts
+++ b/packages/main/src/mainWindow.ts
@@ -95,7 +95,9 @@ async function createWindow(): Promise<BrowserWindow> {
     }
 
     if (import.meta.env.DEV) {
-      browserWindow?.webContents.openDevTools();
+      if (!process.env.CLOSE_DEVTOOLS || process.env.CLOSE_DEVTOOLS !== 'true') {
+        browserWindow?.webContents.openDevTools();
+      }
     }
   });
 
--- a/packages/main/src/mainWindow.ts
+++ b/packages/main/src/mainWindow.ts
@@ -94,10 +94,11 @@ async function createWindow(): Promise<BrowserWindow> {
       browserWindow.show();
     }
 
-    if (import.meta.env.DEV) {
-      if (!process.env.CLOSE_DEVTOOLS || process.env.CLOSE_DEVTOOLS !== 'true') {
-        browserWindow?.webContents.openDevTools();
-      }
+    // Only in DEV mode, if OPEN_DEVTOOLS is not set or is true, we want to open Developer tools window
+    // When running the E2E tests, to avoid problems with small application window, we are settings
+    // OPEN_DEVTOOLS=false.
+    if (import.meta.env.DEV && (!process.env.OPEN_DEVTOOLS || process.env.OPEN_DEVTOOLS === 'true')) {
+      browserWindow?.webContents.openDevTools();
     }
   });
 


packages/main/src/mainWindow.ts
```diff
@@ -95,7 +95,9 @@ async function createWindow(): Promise<BrowserWindow> {
     }
 
     if (import.meta.env.DEV) {
-      browserWindow?.webContents.openDevTools();
+      if (!process.env.CLOSE_DEVTOOLS || process.env.CLOSE_DEVTOOLS !== 'true') {
```
jeffmaury: That does not sound logic: I would rather have OPEN_DEVTOOLS=false to prevent opening DevTools. And test should be merged with previous test

benoitf:
here I would opt-in for a Podman Desktop preference where in development mode, I can choose between : no dev tools, or all the options of devtools like the mode:

openDevTools has the parameter option:
 
```
options Object (optional)
mode string - Opens the devtools with specified dock state, can be left, right, bottom, undocked, detach. Defaults to last used dock state. In undocked mode it's possible to dock back. In detach mode it's not.
```

odockal:
> here I would opt-in for a Podman Desktop preference where in development mode, I can choose between : no dev tools, or all the options of devtools like the mode:
> 
> openDevTools has the parameter option:
> 
> ```
> options Object (optional)
> mode string - Opens the devtools with specified dock state, can be left, right, bottom, undocked, detach. Defaults to last used dock state. In undocked mode it's possible to dock back. In detach mode it's not.
> ```

@benoitf  Can you please clarify more about how this can be put into preferences and how it should work? Right now I am just assuming and doing something that will not show up on preferences since when in development mode, it is too late for e2e tests, and for `yarn watch` why should we be putting this into preferences as this option is already included in dev tools itself, developer can simply open dev tools and choose whatever suits him. right?

I can only see a benefit in using configuration properties as a way how to configure dev tools in dev mode on application start up but I am not sure how pick up two different options other than using env. vars. (Yes, @jeffmaury is right about wrong names).

benoitf:
> @benoitf Can you please clarify more about how this can be put into preferences and how it should work? Right now I am just assuming and doing something that will not show up on preferences since when in development mode, it is too late for e2e tests, and for yarn watch why should we be putting this into preferences as this option is already included in dev tools itself, developer can simply open dev tools and choose whatever suits him. right?

for example today, if I'm not interested to see devtools, I need to close the window each time I run it.
with a property I can say that I don't want to see it 'by default' on every startup and keep it only on demand.

so while it can fit with developer experience, I would rather add the setting as a setting rather than using an environment variable.

odockal:
@benoitf Here is an issue to properly handle the problem: https://github.com/containers/podman-desktop/issues/5031

benoitf:
superseded by https://github.com/containers/podman-desktop/pull/5280

PR #4999, opened 2023-11-27
feat: make Podman machine editable

### LINKED ISSUE #4113: Make cpu/memory/disk of Podman machine being editable

jeffmaury:
Fixes #4113

### What does this PR do?

Allow to edit Podman machine settings: memory, cpu, disk

Fixes #4113

### How to test this PR?

1. Have a Podman machine
2. Go to Resources pages
3. On the page, click the edit icon and change the values, machine is updated and restarted if running


benoitf:
nice enhancement !

it looks like when I create a machine, sliders have 'small steps" like memory is like every 0.5GB
but in edit mode, I'm never reaching a fixed value

in creation mode, I can use left and right arrows to decrease/increase the value but it seems it's not working on edit mode ?


https://github.com/containers/podman-desktop/assets/436777/3ae194c8-c85d-48c8-9d22-361b24cc89df

jeffmaury:
> nice enhancement !
> 
> it looks like when I create a machine, sliders have 'small steps" like memory is like every 0.5GB but in edit mode, I'm never reaching a fixed value
> 
> in creation mode, I can use left and right arrows to decrease/increase the value but it seems it's not working on edit mode ?
> Screen.Recording.2023-11-27.at.11.58.36.mov

Will check but looks strange to me as create and edit are sharing the same component. Maybe related to max value.

jeffmaury:
> nice enhancement !
> 
> it looks like when I create a machine, sliders have 'small steps" like memory is like every 0.5GB but in edit mode, I'm never reaching a fixed value
> 
> in creation mode, I can use left and right arrows to decrease/increase the value but it seems it's not working on edit mode ?
> Screen.Recording.2023-11-27.at.11.58.36.mov

Fixed by ceeba2d

deboer-tim (changes requested):
Confirmed that the fix in second commit is good, the slider stops are working well for me.

A few concerns:
- If I've made changes the Close button isn't clear - you can only infer it won't save changes because the other button is Update. I think it should be 'Cancel' instead. I'd prefer Cancel in the creation flow too, but it seems more confusing to me in this path since in most places (e.g. Preferences) changes are live and you don't need to Update.
- I can edit a non-running Podman machine, and when I click Update I get exit code 125, and the message `Error: podman-machine-default: VM does not exist`.
- When it fails, Show Logs and Close are added along the top but there is no other indication that something failed - I have to expand the Logs to see that there was actually a failure. There should be an error message and/or the Logs should auto-expand.
 
And one more likely unrelated? I couldn't copy the error from the command window.

jeffmaury:
> Confirmed that the fix in second commit is good, the slider stops are working well for me.
> 
> A few concerns:
> 
>     * If I've made changes the Close button isn't clear - you can only infer it won't save changes because the other button is Update. I think it should be 'Cancel' instead. I'd prefer Cancel in the creation flow too, but it seems more confusing to me in this path since in most places (e.g. Preferences) changes are live and you don't need to Update.
> 
>     * I can edit a non-running Podman machine, and when I click Update I get exit code 125, and the message `Error: podman-machine-default: VM does not exist`.
> 
>     * When it fails, Show Logs and Close are added along the top but there is no other indication that something failed - I have to expand the Logs to see that there was actually a failure. There should be an error message and/or the Logs should auto-expand.
> 
> 
> And one more likely unrelated? I couldn't copy the error from the command window.

I fixed the error issue. I left the renaming from Close to Cancel to a follow up PR because it is common with Create machine

benoitf (approved):

benoitf:
I think we need this for 1.6 no ?

jeffmaury:
> I think we need this for 1.6 no ?

I don't remember being listed as a strong requirement but can make it

jeffmaury:
> > I think we need this for 1.6 no ?
> 
> I don't remember being listed as a strong requirement but can make it

But we need @deboer-tim to approve as he requested changes

slemeur:
I'd like to include it into 1.6 yes.

deboer-tim (approved):
Sorry, I didn't notice the review request. Bug is gone, would still like the rename but as you said that's not strictly part of this PR. LGTM, thanks for changes.

slemeur:
Adding recording for release note:
resize-podman-machine

Issue #4985, opened 2023-11-24
Button Toggle Group


feloy:
### Is your feature request related to a problem? Please describe

Related to #4707

To implement the toggle buttons to select the severity to display, we need to have a "Button Toggle Group" component (similar to https://material.angular.io/components/button-toggle/overview).

![image-checker-design](https://github.com/containers/podman-desktop/assets/9973512/3fd15cbf-257e-4f8b-a4ce-70afb48d57ba)

### Describe the solution you'd like

The interface for the component would be:

```
const buttons: ToggleButton[] = [
{
  icon: faExclamationCircle,
  iconClass: 'text-red-600',
  name: 'Critical (5)',
  checked: true,
  disabled: false
}, {
  icon: faExclamationTriangle,
  iconClass: 'text-amber-500',
  name: 'High (2)',
  checked: true,
  disabled: false
}, {
  icon: faExclamationTriangle,
  iconClass: 'text-gray-800',
  name: 'Medium (0)',
  checked: false,
  disabled: true
}, ...
];

function onChange(newState: ToggleButton[]) {
 // newState contains updated checked values for buttons
}

<ToggleButtonGroup buttons={buttons} on:change={(event) => onChange(event.details)} />
```

benoitf:
should we have the declarative button definition or just something more in line with svelte components

```svelte
<ToggleButtonGroup>
  <ToggleButton icon={faExclamationCircle} class="text-red-600">Critical ${data.length}</ToggleButton>
  <ToggleButton icon={...} class="text-red-600">Critical ${data.length}</ToggleButton>
</ToggleButtonGroup>
```svelte

feloy:
Yes, it seems a more elegant way.

But I don't have a clue how to implement it this way. For example, how the group component can know about the children buttons, which one is the first one, and which one the last one, to draw them differently (corners being rounded only at beginning and end of the group)?

benoitf:
maybe it's not possible :-)

if it might help:
in tailwindcss you have `first` and `last` directive to know if you're the first or last child
https://tailwindcss.com/docs/hover-focus-and-other-states#first-last-odd-and-even

you can bind children using `bind:this={yourArray[i]} `

but I'm not sure about why, etc

you can also displatch stuff 

in your case you can use setContext/getContext https://svelte.dev/docs/svelte#setcontext between component and children, I mean I have not looked deeply of what you need but between svelte and tailwind there are some capabilities.

feloy:
[...]

PR #5000, opened 2023-11-27
feat: toggle button group to select severity results to display

feloy:

### What does this PR do?

On the Image Checker screen, adds toggle buttons to filter results by severity.

TODO
- [x] add unit tests

https://github.com/containers/podman-desktop/assets/9973512/33c17817-11cb-4816-bb56-df41535a5688

#4985

### How to test this PR?

You can install a fake image checker from the image at quay.io/phmartin/fake-image-checker:v0.0.3 (sources: https://github.com/feloy/podman-desktop-extension-fake-image-checker)

### DIFF
--- a/packages/renderer/src/lib/ui/ProviderResultPage.svelte
+++ b/packages/renderer/src/lib/ui/ProviderResultPage.svelte
@@ -156,7 +156,7 @@ function onSeverityClicked(severity: 'critical' | 'high' | 'medium' | 'low' | 's
   <div class="h-full flex flex-row space-x-8">
     <div class="h-full overflow-y-auto w-1/3">
       {#each providers as provider}
-        <div class="rounded-lg bg-charcoal-700 mb-4 p-4 flex flex-col">
+        <div role="row" class="rounded-lg bg-charcoal-700 mb-4 p-4 flex flex-col">
           <div class="flex flex-row items-center">
             <span class="grow">{provider.info.label}</span>
             {#if provider.state === 'running'}
@@ -190,6 +190,7 @@ function onSeverityClicked(severity: 'critical' | 'high' | 'medium' | 'low' | 's
     <div class="h-full w-full pr-4 overflow-y-scroll pb-16">
       {#each filtered as result}
         <div
+          role="row"
           class="rounded-r-lg bg-charcoal-700 mb-4 mr-4 p-4 border-l-2"
           class:border-l-red-600="{result.check.severity === 'critical'}"
           class:border-l-amber-500="{result.check.severity === 'high'}"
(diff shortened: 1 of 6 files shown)


benoitf (approved):
hi, my only comment would be about the appearance of the toggles

we see toggle even when we're still processing checkers and we don't have any result.

so I don't know if we should hide toggles until we have checkers/results or disable toggles (like greyed colors) so we see them but it's not really enabled or if there is a missing loading/spinner/horizontal bar in the right part of the screen let me think that there is something that might appear.

As a side note, if all checkers run successfully but do not report anything (no warning, no success), do we see something on the right or it's blank ?

feloy:
> hi, my only comment would be about the appearance of the toggles
> 
> we see toggle even when we're still processing checkers and we don't have any result.
> 
> so I don't know if we should hide toggles until we have checkers/results or disable toggles (like greyed colors) so we see them but it's not really enabled or if there is a missing loading/spinner/horizontal bar in the right part of the screen let me think that there is something that might appear.
> 
> As a side note, if all checkers run successfully but do not report anything (no warning, no success), do we see something on the right or it's blank ?

When the processing is still running, the toggles are actually disabled (as a side effect that the count is zero for the different severities). 

When a first provider has finished running, the buttons start to be enabled, depending on the severities available in the result.

benoitf:
in the recorded video we still see
image

and toolbar does not render like 'disabled' to me

feloy:
I accentuated the disabled state, and added the colors on icons I has forgotten (only when not disabled)

benoitf:
thanks @feloy looks like much better

benoitf (approved):
I thought I approved it this morning but it seems no...

cdrage (approved):
Screenshot 2023-11-27 at 9 15 38 PM

Sorry was late with the review. But the code LGTM and I was able to test it. The toggling looks good.

One small comment, but I honestly don't think it matters.

packages/renderer/src/lib/ui/ProviderResultPage.svelte
```diff
@@ -53,20 +88,75 @@ function onProviderChecked(id: string, checked: boolean) {
...
+      return selectedSeverities['success'];
+    }
+    if (r.check.severity) {
+      return selectedSeverities[r.check.severity];
+    }
+    return true;
+  });
+}
+
+function onSeverityClicked(severity: 'critical' | 'high' | 'medium' | 'low' | 'success', clicked: boolean) {
```
cdrage, line 107: Should `'critical' | 'high' | 'medium' | 'low' | 'success'` be a const / type instead?

But honestly not a big deal if we change it. Everything LGTM code-wise!

PR #5003, opened 2023-11-27
chore(test,ci): fix broken e2e test execution

ScrewTSW:
### What does this PR do?

* Fixes missing playwright dependencies installation

#4983
#5032 

### How to test this PR?

* yarn --frozen-lockfile
* yarn run test:e2e:smoke

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "compile:current": "cross-env MODE=production npm run build && electron-builder build --config .electron-builder.config.cjs",
     "test": "npm run test:unit && npm run test:e2e",
     "test:unit": "npm run test:main && npm run test:preload && npm run test:preload-docker-extension && npm run test:renderer && npm run test:tools && npm run test:extensions",
-    "test:e2e": "cross-env NODE_ENV=development MODE=development DEBUG=pw:browser npm run build && npm run test:e2e:run",
+    "test:e2e": "npm run test:e2e:build && npm run test:e2e:run",
     "test:e2e:build": "cross-env NODE_ENV=development MODE=development DEBUG=pw:browser npm run build",
     "test:e2e:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/src/ --threads false",
     "test:e2e:smoke": "npm run test:e2e:build && npm run test:e2e:smoke:run",
@@ -92,7 +92,8 @@
     "website:prod": "cd website && yarn run docusaurus build && yarn serve",
     "website:dev": "cd website && yarn run docusaurus start",
     "website:screenshots": "cd website-argos && yarn run screenshot",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "postinstall": "playwright install"
   },
   "lint-staged": {
     "*.{js,ts,tsx,svelte}": [
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -291,9 +291,6 @@ jobs:
       - name: Execute yarn
         run: yarn --frozen-lockfile
 
-      - name: setup playwright
-        run: npx playwright install
-
       - name: Run E2E smoke tests
         run: yarn test:e2e:smoke
 


jeffmaury (approved):
LGTM

benoitf (changes requested):
it only needs to be done once, so it should not be done at each run of the test command

odockal:
@benoitf It does not force overwrite the artifacts, so if it is called twice, it does not do anything if already up to date. 
Before it was part of `yarn install`, now it would be part of `yarn test:e2e:XXX`, which is truly called more than once when running tests, but it is better, imho, than adding this steps to the every workflow.

benoitf:
> so if it is called twice, it does not do anything if already up to date
yes but it's still not something that we should do there



move it to `postinstall` where it means it should be only done 'once' when dependencies are updated/changed

odockal:
@benoitf Nice, neat.

package.json
```diff
@@ -43,8 +43,8 @@
     "compile:current": "cross-env MODE=production npm run build && electron-builder build --config .electron-builder.config.cjs",
     "test": "npm run test:unit && npm run test:e2e",
     "test:unit": "npm run test:main && npm run test:preload && npm run test:preload-docker-extension && npm run test:renderer && npm run test:tools && npm run test:extensions",
-    "test:e2e": "cross-env NODE_ENV=development MODE=development DEBUG=pw:browser npm run build && npm run test:e2e:run",
-    "test:e2e:build": "cross-env NODE_ENV=development MODE=development DEBUG=pw:browser npm run build",
+    "test:e2e": "npm run test:e2e:build && npm run test:e2e:run",
```
odockal, line 46: As per discussion on PR, please revert running the install part in tests, and put it into a postinstall.

```
...
"prepare": "husky install",
"postinstall": "npx playwright install webkit ffmpeg"
```

odockal (changes requested):

package.json
odockal: @benoitf Like this?

ScrewTSW: fixed, tested.

odockal (approved):

package.json
```diff
@@ -92,7 +92,8 @@
     "website:prod": "cd website && yarn run docusaurus build && yarn serve",
     "website:dev": "cd website && yarn run docusaurus start",
     "website:screenshots": "cd website-argos && yarn run screenshot",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "postinstall": "npx playwright install webkit ffmpeg"
```
benoitf: why do we need these extra items ( WebKit and ffmpeg) and not only install ?

odockal: `npx playwright install` downloads all default browser, (including chromium, firefox) and it is not necessary, imho.

benoitf: @odockal it does require webkit for electron ? AFAIK electron is based on chromium

odockal: @benoitf I am not sure about what exactly electron requires, but I tried `webkit + ffmpeg` and `chromium` only (it pulls ffmpeg by default) and both combination works. So I guess we can stick to defaults which could be more general for any OS we will run workflow in.

benoitf: @odockal and only ffmpeg ?

odockal: @benoitf Well.. maybe, I will let @ScrewTSW to verify.

odockal:
It is all messy. I have created new issue that this PR directly addresses. It is #5032.
#4983 which I though was to be addressed by this PR will probably not be fixed by this.

package.json
ScrewTSW: @benoitf hello, we have reset the requirements to the default ones as they were being downloaded previously

Would that be an acceptable solution?

benoitf (approved):

package.json
```diff
@@ -92,7 +92,8 @@
     "website:prod": "cd website && yarn run docusaurus build && yarn serve",
     "website:dev": "cd website && yarn run docusaurus start",
     "website:screenshots": "cd website-argos && yarn run screenshot",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "postinstall": "npx playwright install"
```
benoitf: remove npx before merging

ScrewTSW: done.

PR #5004, opened 2023-11-27
feat: add local version retrieval for cli API

### LINKED ISSUE #5008: Podman Desktop API: Should be able to call a "get local binary version" function.

See implementation here: https://github.com/containers/podman-desktop/pull/5004

We should be able to call a function such as `getVersion` or `getLocalVersion` that will retrieve the version of the local binary and report the version back.

This is required so we can correctly check a local version and then check for updates and compare.

Will also allow PD to check / retrieve the binary version periodically rather than only working on start-up (current functionality).

Modify registering a CLI to something similar to this:

```ts
  kubectlCliTool = cli.createCliTool({
    markdownDescription,
    name: 'kubectl',
    displayName: 'kubectl',
    images: {
      icon: 'icon.png',
    },
    path: binaryPath,
    getVersion: async () => {
      return detectVersion('kubectl', ['version', '--client=true', '-o=json']);
    },
  });
}
```

_No response_

### Additional context

See: https://github.com/containers/podman-desktop/pull/5004

Part of implementing: https://github.com/containers/podman-desktop/issues/4689

cdrage:
feat: add local version retrieval for cli API

### What does this PR do?

* Adds the ability for the Podman Desktop API to retrieve the local
  binary for the extension via a defined function.
* Fixes the problem with Compose extension only showing the local
  storage binary version rather than the system-wide installed one.
  
This is required so that we can update the version reactively / see the changes locally so that when we update the binary, it's reactively updated within Podman Desktop.

N/A

Part of https://github.com/containers/podman-desktop/issues/4689

Closes https://github.com/containers/podman-desktop/issues/5008

### How to test this PR?

Go to CLI screen in settings and you should see the appropriate CLI
version.

You can confirm it is the correct version shown by typing `podman
compose version` into the CLI, or `kubectl version`.

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2481,16 +2481,25 @@ declare module '@podman-desktop/api' {
     displayName: string;
     markdownDescription: string;
     images: ProviderImages;
+    path: string;
 
     /**
-     * Within your extension, it is reccommended to implement your own functionality to check the current
+     * Within your extension, it is recommended to implement your own functionality to check the current
      * version number of the CLI tool. For example, parsing the information from the CLI tool's `--version` flag.
      * Passing in path will also help to show where the CLI tool is expected to be installed.
-     * This is usually the ~/.local/share/containers/podman-desktop/extensions-storage directory.
+     *
+     * You can either pass in the current version in the `version` field or implement `getLocalVersion` function to
+     * return the current version of the CLI tool.
+     *
      * Note: The expected value should not include 'v'.
      */
-    version: string;
-    path: string;
+    version?: string;
+
+    /**
+     * Since it's difficult to determine how to retrieve the exact version of the local binary, we recommend the
+     * extension use the getLocalVersion function to retrieve the version of the CLI tool.
+     */
+    getLocalVersion: () => Promise<string>;
   }
 
   export type CliToolState = 'registered';
@@ -2506,6 +2515,7 @@ declare module '@podman-desktop/api' {
       id: string;
       label: string;
     };
+    getLocalVersion: () => Promise<string>;
   }
 
   /**
--- a/extensions/compose/src/extension.ts
+++ b/extensions/compose/src/extension.ts
@@ -24,7 +24,7 @@ import { ComposeGitHubReleases } from './compose-github-releases';
 import { OS } from './os';
 import * as handler from './handler';
 import { ComposeDownload } from './download';
-import * as path from 'path';
+import { process as podmanProcess } from '@podman-desktop/api';
 
 let composeVersionMetadata: ComposeGithubReleaseArtifactMetadata | undefined;
 let composeCliTool: extensionApi.CliTool | undefined;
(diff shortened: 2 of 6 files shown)


cdrage:
@benoitf 

Need some advice on the API implementation.

I have removed `version` from the API and replaced with `getVersion` as IMO I don't see why we specify the CLI version when we register it as it may change in the future as it's being updated / downloaded / a new update.

`getVersion` is added so that we can call it within podman desktop / get the local version of the CLI. This is needed so that we can then get versions from GitHub and compare the two. When the binary is updated locally, it will update reactively / update has been successful.

cdrage:
TODO: 

- [ ] Fix extension tests after an initial review

benoitf:
@cdrage in short, we can't remove any field that we had in the API as it'll break clients

benoitf (commented):
as a workaround, you can make version becoming optional

and add a n ew method on CliTool like we did on the provider (updateVersion)

so you register without a version, but can later update the version using updateVersion call

cdrage:
> as a workaround, you can make version becoming optional
> 
> and add a n ew method on CliTool like we did on the provider (updateVersion)
> 
> so you register without a version, but can later update the version using updateVersion call

Ah I get it.

So keep `version`, but when you register you can make it optional and change `getVersion` to `updateVersion` which will check the local binary and update the `version` within?

Only thing is we'll have to a `set` method to `version` since it's currently read-only within the class.

Thanks for noting that stuff added to the API is generally not-removeable, I'll make sure run anything by you first so we can correctly implement.

cdrage:
Ready for review! This adds the ability for the podman desktop UI to now be able to use the API and call a "getLocalVersion" function without having to activate / reactivate the extension.

This is the work towards #5008 

What I was thinking on how to implement #5008 is:

- Having a "update provider" extension would be difficult in handling assets as each github project has different ways of naming release assets, I believe that each extension could handle it's own way of retrieving release information (for example... you can write an extension that "gets" a release from a custom release server if you wanted to... rather than relying on Podman Desktop maintaining different providers (gitlab, github, http, etc.)

- Add `getRemoteVersions` to CliTool returns an array of the latest releases. This would be handled by the extension (similar to `getLocalVersion`).

- Add `hasUpdate` that returns a true / false boolean. 


Then within the Podman Desktop it's as easy as executing `hasUpdate` which will show the notification in the UI.

What do you think @benoitf

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2481,16 +2481,27 @@ declare module '@podman-desktop/api' {
...
-    path: string;
+
+    // Version is optional if getLocalVersion is implemented
+    version?: string;
+
+    /**
+     * Since it's difficult to determine how to retrieve the exact version of the local binary, we recommend the
+     * extension use the getLocalVersion function to retrieve the version of the CLI tool.
+     */
+    getLocalVersion: () => Promise<string>;
```
lstocchi, line 2502: The comment above `version` says `// Version is optional if getLocalVersion is implemented` but getLocalVersion is always implemented. Should it be `getLocalVersion?:`?

cdrage: Yes as pointed out by @benoitf if we want to "remove" something, we make it optional and going forward just have `getLocalVersion` since every CLI provider should implement a way to retrieve the local binary version. I'll fix the comment.

lstocchi (approved):

jeffmaury (commented):
A CliTool was registered with a version specified in the options. What is the purpose of moving that information to a method ?

benoitf (changes requested):
I see no value of having a getLocalVersion method as it's always the extension that is provide g the value anyway ?

it's only missing the update version method like on providers to notify that a version has now been found ( how we don't care it's the responsibility for now of the extension)

Issue #4967, opened 2023-11-23
Improve aria roles for table component


deboer-tim:
### Is your enhancement related to a problem? Please describe

The table component already has more correct and consistent aria roles than the previous tables, but as per comments in #4844 we can do more, both for accessibility and to make playwright tests simpler.

### Describe the solution you'd like

Make the following changes:
- [ ] Each row should have an aria-label="{object.name}" or equivalent so that tests can find a row via `table.getByRole('row', { name: name })` instead of iterating over every row.
- [ ] Do we add a label to every cell as well? See comments in PR #4844.
- [ ] The table div is missing aria-label and aria-describedby as defined at [1].
- [x] The outer div should have rowgroup.
- [ ] Make sure we have a plan for how tests will find cells in the right columns if the user can switch order of or remove columns.

[1] https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/table/

deboer-tim:
Took another pass at this and putting up a PR. The PR will cover the first item and first part of item 3 (aria-label). I won't do aria-describedby for now since the only logical thing to use is typically the page title, which is redundant both for accessibility and e2e tests so not clear it would help.

For these two items:

> * [ ]  Each row should have an aria-label="{object.name}" or equivalent so that tests can find a row via `table.getByRole('row', { name: name })` instead of iterating over every row.
> * [ ]  Make sure we have a plan for how tests will find cells in the right columns if the user can switch order of or remove columns.

I have seen several e2e tests that look up the column name first, then use this as an index into the row. IMHO this is a best practice - it's defensive coding, matches what users do, and avoids a huge amount of boilerplate aria that has limited use otherwise, so I do not plan to do anything for these.

PR #5005, opened 2023-11-27
fix: add rowgroup to tables

deboer-tim:
### What does this PR do?

Lint re-formatting makes it look like a big change, but it's just the addition of <div role="rowgroup"> elements around header row and data rows as described here: https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/table/

N/A, no change.

Part of #4967.

### How to test this PR?

`yarn test:renderer` or try Volumes page to confirm no change.

### DIFF
--- a/packages/renderer/src/lib/table/Table.spec.ts
+++ b/packages/renderer/src/lib/table/Table.spec.ts
@@ -175,3 +175,22 @@ test('Expect correct aria roles', async () => {
     expect(cells.length).toBe(5);
   }
 });
+
+test('Expect rowgroups', async () => {
+  render(TestTable, {});
+
+  // there should be two role groups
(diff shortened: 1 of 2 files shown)


benoitf:
LGTM reviewed using https://github.com/containers/podman-desktop/pull/5005/files?diff=unified&w=1

I suppose we need a test to test this role, see if we.re able to correctly use them  to fetch items/data

deboer-tim:
TBH I expect our e2e tests will use the other parts of #4967 (e.g. lookup row by name) more than this role, I was just ticking it off as an easy thing we can fix to be correct for Aria. We should have tests though - added.

cdrage (commented):
Confirm that there is no change UI wise! Code LGTM.

cdrage (approved):

benoitf (approved):

PR #5011, opened 2023-11-27
Table column mapping and reuse

### LINKED ISSUE #5002: Improve table column sharing

The new table component allows you to create columns for a given type (e.g. VolumeInfoUI) and apply a renderer for that type, e.g.:
`new Column<VolumeInfoUI>('Age', { renderer: VolumeColumnAge, ...`
and:
```
export let object: VolumeInfoUI;
...
  {object.age}
```

This encapsulates the volume/age combination in a nicely typed way, but it doesn't easily support sharing column renderers between types.

There are at least a couple ways we could support sharing:
1. Decide the typing is less important, and use the same pattern with `let object: any` and render whatever age property exists, e.g. VolumeInfoUI.age, ImageInfoUI.age, anything with a `.age`.
2. Let the column provide a mapping function, e.g.:
`new Column<VolumeInfoUI>('Age', { renderer: SimpleStringColumn, map: object => object.age, ...`
and:
```
export let object: string;
...
  {object}
```

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

It has come up a couple times so I thought I should make a proposal for how I think we should share column renderers between tables. IMHO we have three cases:
1. Type-specific column rendering like Image name. These should continue to be type-specific as today.
2. Generic columns that really just need to (e.g.) render a string in a certain way.
3. The middle ground where it's a little more complex, or maybe you need to render a couple properties.

This PR adds a simple type-mapping function to directly support #2, and applies it to VolumeList.

For the middle ground, I think we're going to have convenient cases where objects used in two tables have the same property types & names (e.g. { id: string, some-prop: number}) and we can just render those, and other cases where maybe the renderer should use another type (e.g. Condition) or one of the objects has a different property name. Again, this PR provides a simple way for tables to do whatever mapping is required.

Based on top of #5005 assuming that would merge soon.

N/A

Fixes #5002. We would share more columns as we add other tables that would use them.

### How to test this PR?

`yarn test:renderer`

### DIFF
--- a/packages/renderer/src/lib/table/SimpleColumn.svelte
+++ b/packages/renderer/src/lib/table/SimpleColumn.svelte
@@ -2,4 +2,6 @@
 export let object: any;
 </script>
 
-{object.name}
+<div class="text-sm text-gray-700">
+  {object}
+</div>
--- a/packages/renderer/src/lib/table/SimpleColumn.svelte
+++ b/packages/renderer/src/lib/table/SimpleColumn.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-export let object: any;
+export let object: string;
 </script>
 
 <div class="text-sm text-gray-700">
--- a/packages/renderer/src/lib/table/Table.svelte
+++ b/packages/renderer/src/lib/table/Table.svelte
@@ -186,7 +186,9 @@ function setGridColumns() {
                 : 'justify-self-start'} self-center overflow-hidden max-w-full"
             role="cell">
             {#if column.info.renderer}
-              <svelte:component this="{column.info.renderer}" object="{object}" />
+              <svelte:component
+                this="{column.info.renderer}"
+                object="{column.info.renderMapping ? column.info.renderMapping(object) : object}" />
             {/if}
           </div>
         {/each}
--- a/packages/renderer/src/lib/table/Table.svelte
+++ b/packages/renderer/src/lib/table/Table.svelte
@@ -188,7 +188,7 @@ function setGridColumns() {
             {#if column.info.renderer}
               <svelte:component
                 this="{column.info.renderer}"
-                object="{column.info.renderMapping ? column.info.renderMapping(object) : object}" />
+                object="{column.info.renderMapping?.(object) ?? object}" />
             {/if}
           </div>
         {/each}
--- a/packages/renderer/src/lib/table/TestTable.svelte
+++ b/packages/renderer/src/lib/table/TestTable.svelte
@@ -18,29 +18,29 @@ const people: Person[] = [
   { id: 3, name: 'Charlie', age: 43 },
 ];
 
-const idCol: Column<Person> = new Column('Id', {
+const idCol: Column<Person, string> = new Column('Id', {
   align: 'right',
-  renderMapping: obj => obj.id,
+  renderMapping: obj => obj.id.toString(),
   renderer: SimpleColumn,
   comparator: (a, b) => a.id - b.id,
 });
 
-const nameCol: Column<Person> = new Column('Name', {
+const nameCol: Column<Person, string> = new Column('Name', {
   width: '3fr',
   renderMapping: obj => obj.name,
   renderer: SimpleColumn,
   comparator: (a, b) => a.name.localeCompare(b.name),
 });
 
-const ageCol: Column<Person> = new Column('Age', {
+const ageCol: Column<Person, string> = new Column('Age', {
   align: 'right',
-  renderMapping: obj => obj.age,
+  renderMapping: obj => obj.age.toString(),
   renderer: SimpleColumn,
   comparator: (a, b) => a.age - b.age,
   initialOrder: 'descending',
 });
 
-const columns: Column<Person>[] = [idCol, nameCol, ageCol];
+const columns: Column<Person, string>[] = [idCol, nameCol, ageCol];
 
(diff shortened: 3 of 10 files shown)


packages/renderer/src/lib/table/SimpleColumn.svelte
```diff
@@ -2,4 +2,6 @@
 export let object: any;
```
benoitf: I suppose it should be 'string' type

packages/renderer/src/lib/table/table.ts
```diff
@@ -34,6 +34,14 @@ export interface ColumnInformation<Type> {
    */
   readonly width?: string;
 
+  /**
+   * Map the source object to another type for rendering. Allows
+   * easier reuse and sharing of renderers by converting to simple
+   * types (e.g. rendering 'string' instead of 'type.name') or
+   * converting to a different type.
+   */
+  readonly renderMapping?: (object: Type) => any;
```
benoitf: it should not be `any` but a type (could be a generic or another stuff but we should not drop the type at the end)

packages/renderer/src/lib/table/TestTable.svelte
```diff
@@ -22,19 +20,22 @@ const people: Person[] = [
...
+  renderMapping: obj => obj.name,
+  renderer: SimpleColumn,
   comparator: (a, b) => a.name.localeCompare(b.name),
 });
 
 const ageCol: Column<Person> = new Column('Age', {
   align: 'right',
-  renderer: TestColumnAge,
+  renderMapping: obj => obj.age,
+  renderer: SimpleColumn,
```
benoitf: should we provide the function as a parameter of renderer parameter ? as they're tied

packages/renderer/src/lib/table/SimpleColumn.svelte
deboer-tim: I will fix in next commit.

packages/renderer/src/lib/table/table.ts
deboer-tim: It needs to be generic since the column could be converting to whatever type the renderer supports, but tbh I am still learning typescript and struggling with how you could declare this, help or pointers appreciated.

packages/renderer/src/lib/table/TestTable.svelte
deboer-tim: Likewise I agree this would be good (especially if there was a way to tie in what type the renderer supports, should be either Type or this NewType ...) but I can't find similar examples, and my attempts at what I think might be valid syntax aren't working.

packages/renderer/src/lib/table/table.ts
jeffmaury: Add a second type parameter which defaults to T and make the renderer accept this type and the mapper convert from T to this type ?

packages/renderer/src/lib/table/SimpleColumn.svelte
deboer-tim: Fixed.

packages/renderer/src/lib/table/table.ts
deboer-tim: Figured it out, and now I'm very pleased with Typescript. :) Columns now have a second type, which defaults to the first if you're not using this feature. If you are, the renderMapping must convert to this second type.

packages/renderer/src/lib/table/TestTable.svelte
deboer-tim: Did you mean like changing it to `renderer: { component: SimpleColumn, mapping: obj => obj.name }` where second is optional? I was hoping there was something simpler when you don't need the mapping but haven't found any other options.

deboer-tim:
Tests never ran, so I rebased on main and force pushed.

packages/renderer/src/lib/table/TestTable.svelte
benoitf: I'm fine for now as it's not an exposed API we can refactor whenever we want and see if we would really benefit from having a single entry rather than two

packages/renderer/src/lib/table/Table.svelte
```diff
@@ -186,7 +186,9 @@ function setGridColumns() {
                 : 'justify-self-start'} self-center overflow-hidden max-w-full"
             role="cell">
             {#if column.info.renderer}
-              <svelte:component this="{column.info.renderer}" object="{object}" />
+              <svelte:component
+                this="{column.info.renderer}"
+                object="{column.info.renderMapping ? column.info.renderMapping(object) : object}" />
```
benoitf: just wondering if we could use `column.info.renderMapping?.(object) ?? object` but not a required change

benoitf (approved):

packages/renderer/src/lib/table/Table.svelte
deboer-tim: Sorry, I am still missing out on obvious better syntax. Updated.

packages/renderer/src/lib/table/TestTable.svelte
deboer-tim: I don't think it makes a huge difference, but since they are related it probably makes sense. I'd prefer not to touch it while there are two PRs in review it would 'break', so I've opened #5071 for me to follow up on this later.

cdrage (approved):

Issue #4760, opened 2023-11-10
Provide ARIA functionality into Extensions page


odockal:
Is your enhancement related to a problem? Please describe

Enhance ARIA accessibility of the Extensions Page. Settings -> Extensions.
There is a shared code for all settings pages: `pacakges/renderer/src/lib/prefereces/SettingsPage.svelte`

Describe alternatives you've considered

Additional context
https://www.w3.org/WAI/standards-guidelines/aria/

PR #5023, opened 2023-11-28
chore(ui,test): add aria functionality to SettingsPage

ScrewTSW:
### What does this PR do?

* Adds ARIA functionality to SettingsPage and PreferencesExtensionList

#### Left side is modified, right side is current main
Screenshot_20231128_174533c
Screenshot_20231128_174508c

#4760 

### How to test this PR?
* run `yarn --frozen-lockfile`
* add `"test:e2e:smoke:podman": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/src/podman-extension-smoke.spec.ts --threads false",` to package.json
* run `yarn run test:e2e:smoke:podman`

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
@@ -83,7 +83,10 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
       <h1 class="text-lg mb-2">Install a new extension from OCI Image</h1>
 
       <div class="flex flex-col w-full">
-        <div class="flex flex-row mb-2 w-full space-x-8 items-center">
+        <div
+          class="flex flex-row mb-2 w-full space-x-8 items-center"
+          role="region"
+          aria-label="Install Extension From OCI">
           <input
             name="ociImage"
             id="ociImage"
@@ -117,11 +120,11 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
       </div>
     </div>
     <div class="bg-charcoal-600 mt-5 rounded-md p-3">
-      <table class="min-w-full">
+      <table class="min-w-full" aria-label="Installed Extensions">
         <tbody>
           {#each $extensionInfos as extension}
-            <tr class="border-y border-gray-900">
-              <td class="px-6 py-2">
+            <tr class="border-y border-gray-900" aria-label="{extension.name}">
+              <td class="px-6 py-2" aria-label="Extension Details">
                 <div class="flex items-center">
                   <div class="flex-shrink-0 h-10 w-10 py-3" title="Extension {extension.name} is {extension.state}">
                     <Fa
@@ -132,7 +135,7 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
                       icon="{faPuzzlePiece}" />
                   </div>
                   <div class="ml-4">
-                    <div class="flex flex-row">
+                    <div class="flex flex-row" aria-label="Extension Details">
                       <div class="text-sm text-gray-300">
                         {extension.displayName}
                         {extension.removable ? '(user)' : '(default extension)'}
@@ -142,13 +145,16 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
                           <button
                             class="mx-2 px-2 text-xs font-medium text-center text-white bg-violet-600 rounded-sm hover:bg-dustypurple-800 focus:ring-2 focus:outline-none focus:ring-dustypurple-700"
                             title="Update to {extension.update.version}"
+                            aria-label="Extension Action Update"
                             on:click="{() => updateExtension(extension, extensionUpdate.ociUri)}">
                             Update</button>
                         {/if}
                       </div>
                     </div>
-                    <div class="flex flex-row">
-                      <div class="text-sm text-gray-700 italic">{extension.description}</div>
+                    <div class="flex flex-row" aria-label="Extension Description">
+                      <div class="text-sm text-gray-700 italic">
+                        {extension.description}
+                      </div>
                     </div>
                     <div class="flex">
                       <ConnectionStatus status="{extension.state}" />
@@ -156,15 +162,17 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
                   </div>
                 </div>
               </td>
-              <td class="px-2 py-2 whitespace-nowrap">
+              <td class="px-2 py-2 whitespace-nowrap" aria-label="Extension Actions">
                 <div class="flex flex-row justify-end">
                   <button
                     title="Start extension"
+                    aria-label="Extension Action Start"
                     on:click="{() => startExtension(extension)}"
                     class="{buttonClass}"
                     class:hidden="{extension.state !== 'stopped'}"><Fa class="h-4 w-4" icon="{faPlay}" /></button>
                   <button
                     title="Stop extension"
+                    aria-label="Extension Action Stop"
                     class="{buttonClass}"
                     on:click="{() => stopExtension(extension)}"
                     hidden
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
```diff
@@ -83,7 +83,10 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
       <h1 class="text-lg mb-2">Install a new extension from OCI Image</h1>
 
       <div class="flex flex-col w-full">
-        <div class="flex flex-row mb-2 w-full space-x-8 items-center">
+        <div
+          class="flex flex-row mb-2 w-full space-x-8 items-center"
+          role="region"
+          aria-label="InstallExtensionFromOCI">
```
odockal: Use spaces between words, everywhere in aria-label.

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
```diff
@@ -116,11 +119,11 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
         </div>
       </div>
     </div>
-    <div class="bg-charcoal-600 mt-5 rounded-md p-3">
+    <div class="bg-charcoal-600 mt-5 rounded-md p-3" role="region" aria-label="InstalledExtensions">
```
odockal: I would just put `aria-label` into table element a bit down.

packages/renderer/src/lib/preferences/SettingsPage.svelte
```diff
@@ -6,8 +6,10 @@ export let title: string;
   <div class="min-w-full px-5 py-4" role="region" aria-label="header">
     <div class="flex flex-row">
       <div class="grow">
-        <p class="capitalize text-xl">{title}</p>
-        <p class="text-sm text-gray-700"><slot name="subtitle"><br /></slot></p>
+        <p class="capitalize text-xl" role="heading" aria-level="1" aria-label="title">{title}</p>
```
odockal: Imho using `p` for headings is wrong. On the one side, you have a paragraph, which is role dedicated for some purpose, and then you have basically a content in it, that is heading. This should be fixed. 
It should be rewritten into either (and I am not sure what is better) into `h1` or `h2` elements, or into `div` or `span` with a dedicated role `heading` and the level as you put it here.

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
```diff
@@ -116,11 +119,11 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
...
       <table class="min-w-full">
         <tbody>
           {#each $extensionInfos as extension}
-            <tr class="border-y border-gray-900">
+            <tr class="border-y border-gray-900" aria-label="{extension.name}">
```
odockal, line 126: this is nice

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
```diff
@@ -157,7 +160,7 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
                 </div>
               </td>
               <td class="px-2 py-2 whitespace-nowrap">
-                <div class="flex flex-row justify-end">
+                <div class="flex flex-row justify-end" role="region" aria-label="ExtensionManagement">
```
odockal: Region is not right role here. We are just fine with a `td` element which is `cell` role, we just need to let the user know, what information/action this cell offers. So that would be something like `aria-label="Extension Actions"`.

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
odockal: But I am missing a descriptions/labels/roles in each cell that the row contains. if we would have a way to get the information from each cell, that would be nice.

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
```diff
@@ -116,11 +119,11 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
...
       </div>
     </div>
-    <div class="bg-charcoal-600 mt-5 rounded-md p-3">
+    <div class="bg-charcoal-600 mt-5 rounded-md p-3" role="region" aria-label="InstalledExtensions">
       <table class="min-w-full">
         <tbody>
           {#each $extensionInfos as extension}
-            <tr class="border-y border-gray-900">
+            <tr class="border-y border-gray-900" aria-label="{extension.name}">
               <td class="px-6 py-2">
```
odockal: Here I can imagine `aria-label="Extension state or information"`

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
```diff
@@ -116,11 +119,11 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
...
                 <div class="flex items-center">
```
odockal, line 128: It is a pity that we got only `div` elements here. Though, it would be nice, to capture the most important information or a general labels to be added to this vague categorization in HTML, so that we can easily get information:
* default vs. user
* extension version
* if the update is available
* state of the extension

odockal (changes requested):

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
ScrewTSW: fixed

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
ScrewTSW: fixed

packages/renderer/src/lib/preferences/SettingsPage.svelte
ScrewTSW: fixed, please review and approve, reference screenshots in PR description

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
ScrewTSW: extended

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
ScrewTSW: fixed

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
ScrewTSW: I wanted to be able to easily reference this element by using exact name of the extension as a label to search for this element

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
ScrewTSW: not sure how to easily differentiate the user extension from a default extension besides this:
```
                  <div class="ml-4">
                    <div class="flex flex-row">
                      <div class="text-sm text-gray-300">
                        {extension.displayName}
                        {extension.removable ? '(user)' : '(default extension)'}
```
for other stuff, we can check the new aria-labels presence on the page based on the labels I just added

odockal:
@ScrewTSW Change subject, feat is dedicated for more complex changes. This is rather chore(ui,tests).

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
ScrewTSW: done.

odockal (approved):

feloy (approved):

PR #5025, opened 2023-11-28
fix: refresh list of kube contexts immediately after delete / change current

### LINKED ISSUE #5014: deleting kube contexts is not working correctly
### Bug description

The deletion of Kubernetes contexts is not working correctly, when the cluster of the current context is not accessible.

Every time a context is deleted, the desktop checks if the cluster of the current context is accessible. If not, this can take a (very) long time. During this time, nothing is displayed in the screen and the user doesn't know what is happening.

Do we really want to do this check every time? In any case, it should not be blocking for the user, as the main purpose of the operation is to edit a file, not to test the connectivity to clusters.

https://github.com/containers/podman-desktop/assets/9973512/afd510eb-6577-4a8b-9bb7-6597f508dfe0

### Operating system

Linux Fedora

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

This PR makes the kubrnetes client send an update event immediately after a context is deleted or the current context is changed.

Another event is sent later, after the .config file change is detected and the content is analyzed, but this event can come very late, because the network connectivity is tested for the current context, which can get some time.

Fixes #5014 

### How to test this PR?

Have a .kube file referencing non accessible clusters, and set the current context for a context with a non accesisble cluster. Then try to delete another context. The deleted context should be deleted immediately from the list.

### DIFF
--- a/packages/main/src/plugin/kubernetes-client-kubeconfig.spec.ts
+++ b/packages/main/src/plugin/kubernetes-client-kubeconfig.spec.ts
@@ -40,6 +40,8 @@ describe('context tests', () => {
 
   let client: TestKubernetesClient;
 
+  const apiSendMock = vi.fn();
+
   function createClient(): TestKubernetesClient {
     const configurationRegistry: ConfigurationRegistry = {} as unknown as ConfigurationRegistry;
     const fileSystemMonitoring: FilesystemMonitoring = new FilesystemMonitoring();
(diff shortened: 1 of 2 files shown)


benoitf (approved):
probably missing a test ?

benoitf (approved):

deboer-tim (approved):

PR #5029, opened 2023-11-28
feat: add store backed by informer

lstocchi:
### What does this PR do?

This PR adds a new type of store which leverages an informer to being updated. 
If the kubernetes context/namespace/user changes, the informer gets updated to listen to the new context.
If there is an error in the connection, it retries automatically and informs the UI.
I still have to add tests, probably clean something but you can play with it.

In the video you can see i'm connected to my kind cluster (default namespace). I see the ingresses, then add/delete one, Switch namespace, stop/start my cluster and the ui reacts to it. 
The top red bar is just an example of what we can do (that part is not in this PR)

https://drive.google.com/file/d/1ZYwavjcvwRHeNLooyTePuer9haiJpm9U/view?usp=drive_link

it is part of #4367 

### How to test this PR?

1. I used this simple component for the demo https://gist.github.com/lstocchi/79b906595f716d6937873e80a1cc45ab

### DIFF
--- a/packages/main/src/plugin/api/informer-info.ts
+++ b/packages/main/src/plugin/api/informer-info.ts
@@ -1,0 +1,26 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import type { Context, Informer, KubernetesObject } from '@kubernetes/client-node';
+
+export type InformerResourcesType = 'INGRESS';
+
+export interface InformerInfo {
+  informer: Informer<KubernetesObject>;
+  context: Context;
+  resourcesType: InformerResourcesType;
+}
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -61,8 +61,8 @@ import { parseAllDocuments } from 'yaml';
 import type { Telemetry } from '/@/plugin/telemetry/telemetry.js';
 import * as jsYaml from 'js-yaml';
 import type { KubeContext } from './kubernetes-context.js';
-import type { InformerManager } from './informer-registry.js';
-import type { InformerResourcesType } from './api/informer-info.js';
+import type { KubernetesInformerManager } from './kubernetes-informer-registry.js';
+import type { KubernetesInformerResourcesType } from './api/kubernetes-informer-info.js';
 
 function toContainerStatus(state: V1ContainerState | undefined): string {
   if (state) {
@@ -142,7 +142,7 @@ export class KubernetesClient {
     private readonly apiSender: ApiSenderType,
     private readonly configurationRegistry: ConfigurationRegistry,
     private readonly fileSystemMonitoring: FilesystemMonitoring,
-    private readonly informerManager: InformerManager,
+    private readonly informerManager: KubernetesInformerManager,
     private readonly telemetry: Telemetry,
   ) {
     this.kubeConfig = new KubeConfig();
@@ -1066,7 +1066,7 @@ export class KubernetesClient {
     throw new Error('no active namespace');
   }
 
-  async startInformer(resourcesType: InformerResourcesType, id?: number): Promise<number> {
+  async startInformer(resourcesType: KubernetesInformerResourcesType, id?: number): Promise<number> {
     switch (resourcesType) {
       case 'INGRESS': {
         return this.createIngressesInformer(id);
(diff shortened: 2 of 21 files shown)


deboer-tim (commented):
I have not done a thorough code review, but overall this looks like a really good step in the right direction. The informer pattern + local cache + future correct disconnection warnings is all goodness.

Given that, this PR is already setting up a good pattern for extending to other kube types, and I was trying to figure out if there are any other ways we could further simplify things like informerListener in the store to put more of it in the informer-event-store. Nothing useful to suggest, and we can always do this in future PRs.

I did notice that the store is still calling window.kubernetesListIngresses(), which uses k8sNetworkingApi.listNamespacedIngress(). I would have expected that code that only needs to get the list once would call window.kubernetesListIngresses(), but anything like the store that needs updates would only rely on the informer (which already calls listNamespacedIngress() as well).

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1006,4 +1012,86 @@ export class KubernetesClient {
       );
     }
   }
+
+  async createIngressesInformer(id?: number): Promise<number> {
```
cdrage, line 1055: Instead of having to add a function for each class,

Should we instead having a map constant of values so it makes it easier for the calls?

Ex. If PODS is passed in, we map and check the table, get /apis/networking.k8s.io/v1/*/pods 

For ingresses, do the same, etc.





For example,

Instead of createIngressesInformer 

It's more generic / uses createInformer.

packages/preload/src/index.ts
```diff
@@ -1490,6 +1490,18 @@ function initExposure(): void {
     return ipcInvoke('kubernetes-client:listIngresses');
   });
 
+  contextBridge.exposeInMainWorld('kubernetesWatchIngresses', async (): Promise<number> => {
```
cdrage: see other comment, should we make it more "generic" and do WatchInformer instead and pass in a type?

cdrage (changes requested):

packages/main/src/plugin/kubernetes-client.ts
lstocchi: Sorry i was busy on other things that had higher priority. Now i'm back to this.
The problem is that we also need to provide a ListPromise object to the informer which has type `() => Promise<{response, body}>`. To create this, we need to set the apiCLient and call the according listNamespaced function with the current namespace. 

Even though we could create a map that has a function that returns what we need as a value (we need to pass the current namespace), I wonder if it's worth it, just to try to save some couple of rows each time we lose readibility. I would skip this if there is not a strong will.

packages/preload/src/index.ts
lstocchi: I already have a generic `createInformer` function that accepts a type as a param, i can expose that function

lstocchi:
ready for a review. Followed Charlie's suggestion and added some tests/cleaned the code

deboer-tim (approved):
Thorough, concerns were addressed, tested, and code LGTM. Thanks Luca.

cdrage (approved):
Thanks for implementing the changes requested, LGTM! Tested and works.

packages/main/src/plugin/api/informer-info.ts
```diff
@@ -0,0 +1,26 @@
...
+ ***********************************************************************/
+import type { Context, Informer, KubernetesObject } from '@kubernetes/client-node';
+
+export type InformerResourcesType = 'INGRESS';
+
+export interface InformerInfo {
+  informer: Informer<KubernetesObject>;
+  context: Context;
+  resourcesType: InformerResourcesType;
+}
```
benoitf: I don't see why InformerResourcesType or InformerInfo would container Kubernetes or Ingress types

IMHO it looks like such generic name can't be used with such specific data added (should be generic and do not contain any Kubernetes or INGRESSES values there)

benoitf: should it be a KubernetesInformer* to be more accurate ? / filename changed as well

packages/renderer/src/stores/informer-event-store.ts
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+// eslint-disable-next-line import/no-duplicates
+import type { InformerWritable } from './informerWritable';
+
+// Helper to manage store updated from events
+export class EventStoreWithInformer<T> {
```
benoitf: should container Kubernetes there as it's not generic

benoitf (commented):
I'm late in the party but I think `Kubernetes` or `k8s`words or something like that should be added to Informer objects to make sure/let people know that they're not generic but stick to Kubernetes Informer pattern (as they're linked/including some Kubernetes anyway)

packages/renderer/src/stores/informerWritable.ts
```diff
@@ -0,0 +1,39 @@
+import type { Invalidator, StartStopNotifier, Subscriber, Unsubscriber, Writable } from 'svelte/store';
+import { writable } from 'svelte/store';
+
+/** Writable interface for both updating and subscribing. */
+export interface InformerWritable<T> extends Writable<T> {
```
benoitf: maybe rename this one too ?

packages/renderer/src/stores/informerWritable.ts
```diff
@@ -0,0 +1,39 @@
...
+  startInformer?: () => Promise<number>,
+  start?: StartStopNotifier<T>,
+): InformerWritable<T> {
+  let informer: number | undefined;
+  const origWritable = writable(value, start);
+
+  function subscribe(this: void, run: Subscriber<T>, invalidate?: Invalidator<T>): Unsubscriber {
+    startInformer?.()
+      .then(id => (informer = id))
+      .catch((e: unknown) => console.error(e));
```
benoitf: might be good to say in the error message some context to better understand the error (let say we only see in the console `invalid name` or `parsing error` it's hard to know what happened

benoitf (approved):
thanks @lstocchi 

I added just minor comments, could be follow-up/ignored

PR #5054, opened 2023-11-29
feat: allow to show new/update cli version

### LINKED ISSUE #4690: CLI Update Mechanism: UI update button & version checking

Part of https://github.com/containers/podman-desktop/issues/3708

We are missing an "Update now" button for updating the binary.

Be able to check what binary is installed, and press "Update now" and go through the appropriate steps (downloading, installing and system-wide)

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR adds the logic to show new/update cli version. Probably to be splitted in multiple parts.

if succeed

update_version_succeed

if fails

update_version_error

it resolves #4690

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
@@ -74,7 +74,7 @@ function getLoggerHandler(_cliToolId: string): ConnectionCallback {
         </div>
         <div class="">
           <div class="p-0.5 rounded-lg bg-charcoal-900 w-fit">
-            {#if cliToolStatus}
+            {#if cliTool.version && cliToolStatus}
               <LoadingIconButton
                 action="update"
                 clickAction="{() => {
@@ -85,7 +85,8 @@ function getLoggerHandler(_cliToolId: string): ConnectionCallback {
                 icon="{faCircleArrowUp}"
                 leftPosition="left-[0.4rem]"
                 state="{cliToolStatus}"
-                color="primary" />
+                color="primary"
+                tooltip="{!cliTool.newVersion ? 'No updates' : `Update to v.${cliTool.newVersion}`}" />
             {/if}
           </div>
         </div>
--- a/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
@@ -1,0 +1,135 @@
+<script lang="ts">
+import { faCircleArrowUp, faCircleXmark } from '@fortawesome/free-solid-svg-icons';
+import Button from '../ui/Button.svelte';
+import Fa from 'svelte-fa';
+import { startTask, type ConnectionCallback, eventCollect } from './preferences-connection-rendering-task';
+import type { CliToolInfo } from '../../../../main/src/plugin/api/cli-tool-info';
+import LoadingIconButton from '../ui/LoadingIconButton.svelte';
+
+import Markdown from '../markdown/Markdown.svelte';
+import type { ILoadingStatus } from './Util';
+
+export let cliTool: CliToolInfo;
+let showError = false;
+let cliToolStatus: ILoadingStatus = {
+  inProgress: false,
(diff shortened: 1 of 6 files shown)


packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
```diff
@@ -0,0 +1,135 @@
...
+  } finally {
+    cliToolStatus.inProgress = false;
+    cliToolStatus = cliToolStatus;
+  }
+}
+
+function getLoggerHandler(_cliToolId: string): ConnectionCallback {
+  return {
+    log: () => {},
+    warn: () => {},
```
benoitf, line 43: should we be able to see these logs in case of errors ?

packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
```diff
@@ -0,0 +1,135 @@
...
+}
+</script>
+
+<div role="row" class="bg-charcoal-600 mb-5 rounded-md p-3 flex flex-col">
+  <div class="divide-x divide-gray-900 flex flex-row">
+    <div>
+      <!-- left col - cli-tool icon/name + "create new" button -->
+      <div class="min-w-[170px] max-w-[200px] h-full flex flex-col justify-between">
+        <div class="flex flex-row">
+          {#if cliTool?.images?.icon}
```
benoitf: should it display the extension icon if there is none on the clitool part ?

let say for kubectl extension I would assume that we can use extension icon

feloy:
I'm testing on Linux Fedora, the Update buttons for docker-compose (v2.23.1) and kubectl (v1.24.1) are disabled. Is it expected?

lstocchi:
> I'm testing on Linux Fedora, the Update buttons for docker-compose (v2.23.1) and kubectl (v1.24.1) are disabled. Is it expected?

yes, there is no code on the extensions atm. This covers just the UI part, Jeff will work on the kubectl extension in a follow-up PR.

packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
lstocchi: WDYM? bc when there is an error i show a message to open the task manager. Not sure where to display warn and log messages.

benoitf: I mean, if the update command is displaying stuff on stdout/sterr (or in log method) should we have this log somehow available ?

lstocchi: No idea. More a question for @mairin . Maybe it's enough the task manager?

lstocchi: But i can enhance it on a separate PR so Jeff can work on the kubectl if it must be ready for the release

packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
lstocchi: it could make sense. I wonder if we should make the image mandatory so the extension is forced to provide an icon or we should ask for additional extensionInfo

benoitf (approved):

packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
```diff
@@ -0,0 +1,141 @@
...
+      getLoggerHandler(cliTool.id),
+    );
+    await window.updateCliTool(cliTool.id, loggerHandlerKey, eventCollect);
+    showError = false;
+    cliToolStatus.status = 'unknown';
+  } catch (e) {
+    showError = true;
+  } finally {
+    cliToolStatus.inProgress = false;
+    cliToolStatus = cliToolStatus;
```
jeffmaury, line 36: Is that required ?

lstocchi: the loadingButton does not get refreshed otherwise

jeffmaury (approved):
LGTM

PR #5055, opened 2023-11-29
fix: update current context reactively

### LINKED ISSUE #5053: After switching current kube context, popup is not raised when deleting new current one
### Bug description

When deleting the current kube context, a popup is raised to ask the user to confirm the deletion.

But if the current context is changed to another context, deleting this new context does not raise the popup.

### Operating system

Linux Fedora

### Installation Method

### Version

next (development version)

### Steps to reproduce

- go to Settings > Kubernetes
- switch the current context to another context
- delete this new current context

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

On the Kubernetes Context page, the current context is updated reactively, so it has the good value when the suer switches the current context.

Fixes #5053

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
@@ -2,7 +2,6 @@
 import SettingsPage from './SettingsPage.svelte';
 import EngineIcon from '../ui/EngineIcon.svelte';
 import EmptyScreen from '../ui/EmptyScreen.svelte';
-import { onMount } from 'svelte';
 import Link from '../ui/Link.svelte';
 import { faTrash, faRightToBracket } from '@fortawesome/free-solid-svg-icons';
 import ListItemButtonIcon from '../ui/ListItemButtonIcon.svelte';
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -2,19 +2,14 @@
...
 import ErrorMessage from '../ui/ErrorMessage.svelte';
 import { kubernetesContexts } from '../../stores/kubernetes-contexts';
 import { clearKubeUIContextErrors, setKubeUIContextError } from '../kube/KubeContextUI';
 
-let currentContextName: string | undefined;
-
-onMount(async () => {
-  currentContextName = await window.kubernetesGetCurrentContextName();
-});
+$: currentContextName = $kubernetesContexts.filter(c => c.currentContext).map(c => c.name)[0];
```
benoitf: what if array is empty ?

feloy: the test I have made indicates the value of `currentContextName` is `undefined`, which is ok

benoitf: what about using instead `find()` method instead that returns the first matching element ?

`$kubernetesContexts.find(c => c.currentContext)?.name`

benoitf: b/c using `[0]` to access the first element of an empty array looks like an error prone case to me

feloy: ok, I'll change this to make an explicit check

benoitf (approved):

lstocchi (approved):

PR #5064, opened 2023-11-30
feat: add api to register cli updater (#4690)

lstocchi:
### What does this PR do?

It adds the logic to register an cli updater which will be called on request to update a cli.

For gif to see how it will be used/work check https://github.com/containers/podman-desktop/pull/5054

this is a part of #4690. 

### How to test this PR?

1. run tests

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2493,6 +2493,22 @@ declare module '@podman-desktop/api' {
     path: string;
   }
 
+  /**
+   * Options to update CliTool instance
+   */
+  export interface CliToolUpdateOptions {
+    version: string;
+    displayName?: string;
+    markdownDescription?: string;
+    images?: ProviderImages;
+    path?: string;
+  }
+
+  export interface CliToolUpdate {
+    version: string;
+    doUpdate: (logger: Logger) => Promise<void>;
+  }
+
   export type CliToolState = 'registered';
 
   export interface CliTool extends Disposable {
@@ -2506,6 +2522,12 @@ declare module '@podman-desktop/api' {
       id: string;
       label: string;
     };
+
+    updateVersion(version: CliToolUpdateOptions): void;
+    onDidUpdateVersion: Event<string>;
+
+    // register cli update flow
+    registerUpdate(update: CliToolUpdate): Disposable;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -2506,6 +2522,12 @@ declare module '@podman-desktop/api' {
       id: string;
       label: string;
     };
+
+    updateVersion(version: CliToolUpdateOptions): void;
```
benoitf, line 2526: hello, I'm wondering why we need to pass so many things in the update function like displayName, images, path, etc ?

lstocchi: for me a new version could have a different description, display name, icon.

lstocchi: maybe in a new version there is something super which is supported or not supported anymore and you want to mention it in the description, or mark it in the name, or the icon has different color. Idk

benoitf: in that case, maybe the extension unregister the current CliTool and register a new one ?
for the images, I would say that as they come with the extension, probably if CLI has a new icon we will update the new icon in a newest extension but it won't be different in all the calls (like between register and update)

I'll wait other reviews, but it looks a little bit over-engineering to me.

benoitf (approved):
LGTM

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2493,6 +2493,22 @@ declare module '@podman-desktop/api' {
...
 
+  /**
+   * Options to update CliTool instance
+   */
+  export interface CliToolUpdateOptions {
+    version: string;
+    displayName?: string;
+    markdownDescription?: string;
+    images?: ProviderImages;
+    path?: string;
```
cdrage, line 2504: Should we also include the asset name, as when retrieving the / installing the new update, you'll have to pass in the asset name determined by the extension (example, hey, install arm64 darwin). 

Or should we just do quickpick / prompt for now?

cdrage (changes requested):

packages/extension-api/src/extension-api.d.ts
benoitf: it's the extension that is doing the update so extension is aware on asset name to pick-up

lstocchi: the extension has already a reference of the cli that it wants to update as it registered it during the activation. So you're going to do `cli.update(new data)` from the extension when the user asks for it. There is no need for the asset name. 
I didn't make it updatable as I see it like an id. You can change the display name, description, version but the asset name should be the same, otherwise you're most probably registering a new cli tool.

jeffmaury (approved):
LGTM

cdrage (approved):

PR #5066, opened 2023-11-30
chore(tests,ui): added compose onboarding aria accessibility

xbabalov:
Signed-off-by: Tamara Babalova <tbabalov@redhat.com>

### What does this PR do?
Adds aria accessibility into the compose onboarding process. 

https://github.com/containers/podman-desktop/issues/4716

### DIFF
--- a/packages/renderer/src/lib/onboarding/Onboarding.svelte
+++ b/packages/renderer/src/lib/onboarding/Onboarding.svelte
@@ -282,7 +282,11 @@ async function restartSetup() {
     class="flex flex-col bg-charcoal-500 h-full overflow-y-auto w-full overflow-x-hidden"
     class:bodyWithBar="{!activeStep.step.completionEvents || activeStep.step.completionEvents.length === 0}">
     <div class="flex flex-col h-full">
-      <div class="flex flex-row justify-between h-[100px] p-5 z-20 fixed w-full bg-opacity-90 bg-charcoal-700">
+      <div
+        class="flex flex-row justify-between h-[100px] p-5 z-20 fixed w-full bg-opacity-90 bg-charcoal-700"
+        role="heading"
+        aria-level="{2}"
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/onboarding/Onboarding.svelte
```diff
@@ -414,7 +420,9 @@ async function restartSetup() {
 {#if displayCancelSetup}
   <!-- Create overlay-->
   <div class="fixed top-0 left-0 right-0 bottom-0 bg-black bg-opacity-60 bg-blend-multiply h-full grid z-50">
-    <div class="flex flex-col place-self-center w-[550px] rounded-xl bg-charcoal-800 shadow-xl shadow-black">
+    <div
+      class="flex flex-col place-self-center w-[550px] rounded-xl bg-charcoal-800 shadow-xl shadow-black"
+      aria-label="Skip setup popup">
```
benoitf: here it's not using Upper Case on every words like the others `Onboarding Status Message` or `Step Buttons`


but still I'm wondering why we need the aria-label there ?

should it be on the button that don't have an aria-label for now ?
```
<Button type="secondary" aria-label="Cancel" on:click="{() => setDisplayCancelSetup(false)}">Cancel</Button>
<Button type="primary" class="mr-2" on:click="{() => cancelSetup()}">Ok</Button>
```

xbabalov: Thanks for the comment, I fixed the Upper Case for the rest of the labels in the file as well. 

I think it's good to label the popup message for testing purposes so it's clear what we're working with. 

It's also not needed to label buttons when their name is visible in the ui, because equally the aria-label and the button name can be used for the name in getByRole("button", {name:xxx}).

odockal (approved):

Issue #4990, opened 2023-11-26
feat: configured preferred namespace(s) for Kubernetes contexts


axel7083:
### Is your feature request related to a problem? Please describe

When using external cluster, we might want to use a specific namespace by default or we might not have access to the default namespace. 

### Describe the solution you'd like

In the Kubernetes settings, being able to set the preferred namespace.

This settings _could_ be reflected when generating kube.

### Additional context

I do not know if podman-desktop have some kind of internal persistency system ?

afbjorklund:
I think each context has a namespace

```
Usage:
  kubectl config set-context [NAME | --current] [--cluster=cluster_nickname] [--user=user_nickname]
[--namespace=namespace] [options]
```

You can view the current value in CLI:

```console
$ kubectl config get-contexts
CURRENT   NAME        CLUSTER     AUTHINFO    NAMESPACE
*         kind-kind   kind-kind   kind-kind   
```

axel7083:
> I think each context has a namespace
> 
> ```
> Usage:
>   kubectl config set-context [NAME | --current] [--cluster=cluster_nickname] [--user=user_nickname]
> [--namespace=namespace] [options]
> ```
> 
> You can view the current value in CLI:
> 
> ```
> $ kubectl config get-contexts
> CURRENT   NAME        CLUSTER     AUTHINFO    NAMESPACE
> *         kind-kind   kind-kind   kind-kind   
> ```

I don't know how the context would be used specifically in the use case I am seeing. I was thinking for instance, when I am downloading a kubeconfig from rancher for example, it looks like this
```
apiVersion: v1
kind: Config
clusters:
- name: "<name>"
  cluster:
    server: <server>
    certificate-authority-data: <cert>
users:
- name: "<name>"
  user:
    token:  <token>

contexts:
- name: "<name>"
  context:
    user: "<name>"
    cluster: "<name>"

current-context: "<name>"
```

There is no preferred namespace configured, but I do only have access to one namespace, thus, leading to error when importing the config in some Kubernetes dashboards. Some of those dashboard offer a way to configured the "Accessible namespaces"

afbjorklund:
You should be able to add one with `kubectl config set-context <name> --namespace=<name>`

```yaml
contexts:
- context:
    cluster: kind-kind
    namespace: kind-kind
    user: kind-kind
  name: kind-kind
```

The default is empty (and omitted)

[...]

PR #5088, opened 2023-12-01
feat: show the k8s namespace

afbjorklund:
### What does this PR do?

Show the namespace for the context, if any

```console
$ kubectl config get-contexts
CURRENT   NAME        CLUSTER     AUTHINFO    NAMESPACE
*         kind-kind   kind-kind   kind-kind   
$ kubectl config set-context --namespace kind-namespace kind-kind
Context "kind-kind" modified.
$ kubectl config get-contexts
CURRENT   NAME        CLUSTER     AUTHINFO    NAMESPACE
*         kind-kind   kind-kind   kind-kind   kind-namespace
```

* #4990

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
@@ -113,6 +113,14 @@ async function handleDeleteContext(contextName: string) {
             <span class="my-auto font-bold col-span-1 text-right">USER</span>
             <span class="my-auto col-span-5 text-left pl-0.5 ml-3" aria-label="context-user">{context.user}</span>
           </div>
+
+          {#if context.namespace}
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -116,6 +116,14 @@ async function handleDeleteContext(contextName: string) {
             <span class="my-auto font-bold col-span-1 text-right">USER</span>
             <span class="my-auto col-span-5 text-left pl-0.5 ml-3" aria-label="context-user">{context.user}</span>
           </div>
+
+          {#if context.namespace !== undefined}
```
benoitf:
```suggestion
          {#if context.namespace}
```

benoitf (approved):
could you add a component test for the svelte component ? (should render namespace if defined)

cdrage (approved):

Issue #4413, opened 2023-10-19
Pod creation button should say "Create Pod" as small icon is hard to understand.


cdrage:
### Is your enhancement related to a problem? Please describe

When creating a Pod from a group of containers, it's hard to understand what the icon means. 

The trash-can however is explanatory. 

So it's inline with other buttons (Create for container, Volume, Image, etc.), we should add text to the side as well saying "Create Pod".

<img width="289" alt="Screenshot 2023-10-18 at 10 19 24 PM" src="https://github.com/containers/podman-desktop/assets/6422176/1197984f-eb5f-4ca4-9d8a-d2a38ba7e5bf">

### Describe the solution you'd like

Text to be shown when creating Pod.

ecrookshanks-rh:
Hi.  Since this is labeled as a good first issue I would like to be assigned this enhancement.

benoitf:
OK it's now assigned to you @ecrookshanks-rh

cdrage:
Closed via #5095

PR #5095, opened 2023-12-01
fix: added text beside icon for create pods

ecrookshanks-rh:
### What does this PR do?
Adds the text "Create Pod" beside the icon in the Containers List screen.

Screenshot from 2023-11-29 21-59-32

Issue #4413

### How to test this PR?
UI test only.  Go to the Containers tab and select a container.  When the buttons appear near the top of the screen the words "Create Pod" will display on the button next to the pods icon.


benoitf:
on the screenshot it looks like buttons don't have the same height (trash icon and for example 'create pods')

ecrookshanks-rh:
Issue appears to be different size buttons when icons are alone and when they are combined with text.  I will investigate a way to make them align.

ecrookshanks-rh:
Updated Screenshot:

Screenshot from 2023-12-01 13-11-30

deboer-tim (commented):
This code is compensating for the fact that our Buttons have likely never had the same height when there is just an icon vs when there is icon vs text, and currently it is quite noticeable. The only change that _should_ have to be made here is the "Create Pod" line.

Someone is going to have to fix the Button at some point, so I opened #5102. I think it is a separate issue and not fair to @ecrookshanks-rh to have to do that 'correct' Button fix here... but ugly to either make it a prerequisite or accept the current code and remember to clean it up later.

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -467,16 +467,22 @@ function setStoppedFilter() {
...
-        on:click="{() => deleteSelectedContainers()}"
-        aria-label="Delete selected containers and pods"
-        title="Delete {selectedItemsNumber} selected items"
-        bind:inProgress="{bulkDeleteInProgress}"
-        icon="{faTrash}" />
-      <Button
-        on:click="{() => createPodFromContainers()}"
-        title="Create Pod with {selectedItemsNumber} selected items"
-        icon="{SolidPodIcon}" />
+      <div style="display: inline-flex;">
```
benoitf: Hello, if you want to apply style, as we're using tailwind for CSS you should use `inline-flex`
```suggestion
      <div class="inline-flex">
```

https://tailwindcss.com/docs/display

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -467,16 +467,22 @@ function setStoppedFilter() {
...
+        <Button
+          on:click="{() => deleteSelectedContainers()}"
+          aria-label="Delete selected containers and pods"
+          title="Delete {selectedItemsNumber} selected items"
+          bind:inProgress="{bulkDeleteInProgress}"
+          icon="{faTrash}">
+        </Button>
+        &nbsp;
```
benoitf: if you want to add space between element, maybe `space-x-2` on the parent div

https://tailwindcss.com/docs/space

ecrookshanks-rh: I've added the class (nbsp did feel funky) and pushed.

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -467,16 +467,22 @@ function setStoppedFilter() {
...
-        on:click="{() => deleteSelectedContainers()}"
-        aria-label="Delete selected containers and pods"
-        title="Delete {selectedItemsNumber} selected items"
-        bind:inProgress="{bulkDeleteInProgress}"
-        icon="{faTrash}" />
-      <Button
-        on:click="{() => createPodFromContainers()}"
-        title="Create Pod with {selectedItemsNumber} selected items"
-        icon="{SolidPodIcon}" />
+      <div style="display: inline-flex;" class="space-x-2">
```
benoitf:
```suggestion
      <div class="inline-flex space-x-2">
```

benoitf (approved):
thanks @ecrookshanks-rh for the contrib

jeffmaury (approved):
LGTM

PR #5098, opened 2023-12-03
feat: show container connection type and endpoint

### LINKED ISSUE #4727: Show connection endpoint instead of provider information for containers

Currently the provider information is shown for containers, where the connection information is shown for Kubernetes.

This makes it somewhat confusing, when they are shown side by side or when connection name does not match provider.

Like here:

* https://github.com/containers/podman-desktop/pull/4539

lima-provider-both

* the version is missing: https://github.com/containers/podman-desktop/pull/3748
  (it is only showing the CLI version anyway, so it is [obsolete](https://github.com/containers/podman-desktop/pull/4503))

It would be better if the resources showed the endpoints in a similar matter (with the URL), for both types of connections.

And the provider information should probably move to the "left hand side", and not be repeated for every connection ?

podman-desktop-provider-endpoint

Note: this is just showing the raw values (type and socketPath), and has some additional features like name and version

The provider type should be Capitalized, and one might consider truncating long URLs (or only show the path.basepath)

_No response_

### Additional context

_No response_

----

PoC:

```html
              <div class="mt-2">
                <div class="text-gray-700 text-xs">{container.type} endpoint</div>
                <div class="mt-1">
                  <span class="my-auto text-xs" class:text-gray-900="{container.status !== 'started'}"
                    >unix://{container.endpoint.socketPath}</span>
                </div>
```
```html
              <div class="mt-2">
                <div class="text-gray-700 text-xs">Kubernetes endpoint</div>
                <div class="mt-1">
                  <span class="my-auto text-xs" class:text-gray-900="{kubeConnection.status !== 'started'}"
                    >{kubeConnection.endpoint.apiURL}</span>
                </div>
              </div>
```

afbjorklund:
Make the container and the kubernetes endpoint look more similar.

Show the connection type, since it might not match the provider.

pd-endpoints

Fixes #4727

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
@@ -473,6 +473,15 @@ function hasAnyConfiguration(provider: ProviderInfo) {
                   <ConnectionErrorInfoButton status="{status}" />
                 {/if}
               </div>
+              <div class="mt-2">
+                <div class="text-gray-700 text-xs">
+                  {#if container.type === 'docker'}Docker{:else if container.type === 'podman'}Podman{/if} endpoint
+                </div>
+                <div class="mt-1">
+                  <span class="my-auto text-xs" class:text-gray-900="{container.status !== 'started'}"
+                    >{container.endpoint.socketPath}</span>
+                </div>
+              </div>
 
               {#if providerContainerConfiguration.has(provider.internalId)}
                 {@const providerConfiguration = providerContainerConfiguration.get(provider.internalId) || []}
--- a/packages/renderer/src/lib/preferences/PreferencesContainerConnectionDetailsSummary.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesContainerConnectionDetailsSummary.svelte
@@ -63,6 +63,11 @@ $: providerContainerConfiguration = tmpProviderContainerConfiguration.filter(
         </div>
       {/each}
       <div class="flex flex-row mt-5">
+        <span class="font-semibold min-w-[150px]">Type</span>
+        <span aria-label="{containerConnectionInfo.type}"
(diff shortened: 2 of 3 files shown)


packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -473,6 +473,15 @@ function hasAnyConfiguration(provider: ProviderInfo) {
                   <ConnectionErrorInfoButton status="{status}" />
                 {/if}
               </div>
+              <div class="mt-2">
+                <div class="text-gray-700 text-xs">
+                  {#if container.type === 'docker'}Docker{:else if container.type === 'podman'}Podman{/if} endpoint
+                </div>
+                <div class="mt-1">
+                  <span class="my-auto text-xs" class:text-gray-900="{container.status !== 'started'}"
+                    >unix://{container.endpoint.socketPath}</span>
```
benoitf: I think it won't work on windows as it's using `npipe:`

afbjorklund: Seems to be copied everywhere, no function.

```
extensions/compose/src/templates/podman-compose.bat.mustache:set DOCKER_HOST=npipe://{{{ socketPath }}}
extensions/compose/src/templates/podman-compose.sh.mustache:export DOCKER_HOST=unix://{{{ socketPath }}}
```

Will probably make it harder to add `ssh:` etc.

afbjorklund: Probably would be good with an abstraction for type as well

afbjorklund: Changed back from URL to path, until the interface changes.

benoitf (approved):
probably need to have component tests to see if we're displaying correct container type and socketPath based on the status ?

jeffmaury (approved):
LGTM

slemeur:
Merging for being part of 1.6.0 release.
Expecting follow up PR with tests.

PR #5101, opened 2023-12-04
feat: extension icons

### LINKED ISSUE #5100: Use extension icons in Settings > Extensions

The puzzle pieces have always been a noticeable placeholder to me. After seeing PR #5089 I was reminded, and although I've seen an early draft for redoing the extension page that included icons I couldn't find any issue tracking it.

Use each extension's icon instead of the puzzle piece.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

After seeing Luca's PR #5089 I felt like I had to do this. :) Uses the extension's icon in the Settings > Extensions page, using an almost identical component to ProviderLogo. If the extension has no icon, the puzzle piece is still used.

I've put the component in the /preferences folder since that's the only place it is being used for now.

The puzzle pieces used purple vs gray to show whether the extension is running, I've switched to using a brightness filter instead, since it is consistent between the two. Tests included.

Before:

Screenshot 2023-12-03 at 9 28 21 PM

After:

Screenshot 2023-12-03 at 8 56 47 PM

(I've stopped the Kube CLI extension in both for comparison)

Fixes #5100.

### How to test this PR?

Go to Settings > Extensions and start/stop some extensions.

### DIFF
--- a/packages/renderer/src/lib/preferences/ExtensionIcon.svelte
+++ b/packages/renderer/src/lib/preferences/ExtensionIcon.svelte
@@ -1,0 +1,26 @@
+<script lang="ts">
+import Fa from 'svelte-fa';
+import type { ExtensionInfo } from '../../../../main/src/plugin/api/extension-info';
+import { faPuzzlePiece } from '@fortawesome/free-solid-svg-icons';
+
+export let extension: ExtensionInfo;
+
+let icon: string | undefined = undefined;
+$: {
+  if (extension.icon) {
+    if (typeof extension.icon === 'string') {
+      icon = extension.icon;
+    } else if (extension.icon.dark) {
+      // for now use dark theme
+      icon = extension.icon.dark;
+    }
+  }
+}
+$: fade = extension.state !== 'started' ? ' brightness-50' : '';
+</script>
+
+{#if icon}
+  <img src="{icon}" alt="{extension.name}" class="max-w-10 max-h-10 {fade}" />
+{:else}
+  <Fa alt="{extension.name}" class="h-10 w-10 rounded-full text-violet-600 {fade}" size="25" icon="{faPuzzlePiece}" />
+{/if}
--- a/packages/main/src/plugin/api/extension-info.ts
+++ b/packages/main/src/plugin/api/extension-info.ts
@@ -38,6 +38,7 @@ export interface ExtensionInfo {
   state: string;
   error?: ExtensionError;
   path: string;
+  icon?: string | { light: string; dark: string };
   update?: {
     version: string;
     ociUri: string;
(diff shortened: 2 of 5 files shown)


dgolovin (approved):
LGTM

feloy (approved):

packages/renderer/src/lib/preferences/ExtensionIcon.svelte
```diff
@@ -0,0 +1,26 @@
...
+
+let icon: string | undefined = undefined;
+$: {
+  if (extension.icon) {
+    if (typeof extension.icon === 'string') {
+      icon = extension.icon;
+    } else if (extension.icon.dark) {
+      // for now use dark theme
+      icon = extension.icon.dark;
+    }
```
benoitf, line 16: as we'll introduce light theme soon, I would think that it has the logic for light and dark

packages/renderer/src/lib/preferences/ExtensionIcon.svelte
```diff
@@ -0,0 +1,26 @@
...
+  }
+}
+$: fade = extension.state !== 'started' ? ' brightness-50' : '';
+</script>
+
+{#if icon}
+  <img src="{icon}" alt="{extension.name}" class="max-w-10 max-h-10 {fade}" />
```
benoitf, line 23: should it contain the `aria-label` ?

packages/renderer/src/lib/preferences/ExtensionIcon.svelte
deboer-tim: Agreed, but I think the right solution is that we need an appearance-util.js that'll tell you if you should use light or dark, maybe with a utility to pull the right thing from `string | { light: string; dark: string }`, then used here + ProviderLogo (maybe others?). Are you ok doing that as follow-up PR?

packages/renderer/src/lib/preferences/ExtensionIcon.svelte
deboer-tim: I've found a couple references like the following that say you don't need aria-label (as well) when image has alt text:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

`If you give your <iframe>s a title, your images an alt attributes, and your input's associated <label>s, aria-label is not necessary.`

packages/renderer/src/lib/preferences/ExtensionIcon.svelte
deboer-tim: Ok, I caved and threw my thoughts in PR #5117 and will wait until tomorrow for initial feedback. If that approach makes sense I can wait and rebase this; otherwise I would prefer to merge and use #5116 to solve this along with the other places we already have this problem.

benoitf: can be follow-up

cdrage (commented):
Works amazing and LGTM code-wise, no issues there! 🚀

cdrage (approved):

Issue #4880, opened 2023-11-20
Link has no style in onboarding


lstocchi:
### Bug description

![image](https://github.com/containers/podman-desktop/assets/49404737/2df2a5c4-2547-4427-83c0-22d74b3d4e40)

In this old PR you can see the link text had a different color before https://github.com/containers/podman-desktop/pull/4455

### Operating system

Linux

### Installation Method

Installer from website/GitHub releases

### Version

next (development version)

lstocchi:
closed by https://github.com/containers/podman-desktop/pull/5108

PR #5108, opened 2023-12-04
fix: add style to link (#4880)

lstocchi:
### What does this PR do?

it adds style to links as it was before

purple_links_2

it closes 4880

### How to test this PR?

1. open the resources page or any other page with a link

### DIFF
--- a/packages/renderer/src/lib/markdown/Markdown.svelte
+++ b/packages/renderer/src/lib/markdown/Markdown.svelte
@@ -1,6 +1,6 @@
 <!-- The markdown rendered has it's own style that you'll have to customize / check against podman desktop 
 UI guidelines -->
-<style>
+<style lang="postcss">
 .markdown > :global(p) {
   line-height: normal;
   padding-bottom: 8px;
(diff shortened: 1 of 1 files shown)


jeffmaury (changes requested):
I don't see the cursor being updated when the link is in a onboading view

jeffmaury (approved):
LGTM

lstocchi:
I have to remove the underline :/
underline

deboer-tim:
If we're fixing this, any chance we could make the links purple too? I always thought they were supposed to be like buttons, i.e. purple-600 with a -500 hover.

FYI to @mairin

benoitf (commented):
should we not add the style in the Markdown component ?

https://github.com/containers/podman-desktop/blob/417fc4bb2b29946c095a30e5b2cffc1258dc6ee9/packages/renderer/src/lib/markdown/Markdown.svelte#L4-LL36

lstocchi:
Updated based on your suggestions. Not sure if purple-600 is too dark or it's ok. Do we use it somewhere else?
@mairin

deboer-tim:
> Updated based on your suggestions.

Thank you, that's already looking more consistent to me, and code change is better too.

> Not sure if purple-600 is too dark or it's ok. Do we use it somewhere else?

I pulled those colors from the Button for consistency (which is on the left in the video), but I agree it's too dark when it's not a full box. From the video it looks like maybe bumping it up one level to -500 and -400 would work? I would +1 this after a minor tweak.

lstocchi:
@deboer-tim updated. Idk why the gif i record is always a bit darker of what i actually see on my screen

deboer-tim (approved):
LGTM, thanks for the changes.

cdrage (approved):

benoitf (approved):
my approval was not required, so many approvals already but here is mine

thanks @lstocchi

Issue #5116, opened 2023-12-04
Light icon/logo support


deboer-tim:
### Is your enhancement related to a problem? Please describe

We have support for extensions to provide simple icons, or different ones for light vs dark mode. However, since we only have dark mode we have hardcoded many of these to only use light mode: ProviderLogo, PreferencesAuthentication, possibly ExtensionIcon (#5100). 

### Describe the solution you'd like

Simple component for displaying icons in either mode, which picks up the correct theme, and is reused in these places.

cdrage:
![Screenshot 2024-06-11 at 8 18 39 AM](https://github.com/containers/podman-desktop/assets/6422176/c5059a93-d8fc-455e-a5fa-b640fd128dc5)

Kubernetes icon is still white / no dark mode for it.

deboer-tim:
> Kubernetes icon is still white / no dark mode for it.

This issue was created for extension icons/pngs in the files listed at the top. Light mode has been applied to the svgs in the main nav but my guess is a poor choice of color or wrong ref to color palette. But either way, can you open a separate issue to track?

PR #5117, opened 2023-12-04
feat: icon image component

deboer-tim:
### What does this PR do?

Creates a new IconImage component (couldn't think of a better name) that can display our normal icon/logo pattern of "string | { light: string; dark: string } | undefined" as an image, with the default slot as a fallback.

The logic for which image to use could have been directly in the component, but in order to start sharing code with Appearance.svelte (and in case we have other places that will need to know which theme we're using) I extracted the basic function to appearance-utils.js.

This component would allows us to basically delete components like ProviderLogo and ExtensionIcon and replace with single lines like: <IconImage image="{provider?.images?.logo}" class=".." alt="{provider.name}"/>

### Screenshot / video of UI

N/A, unused so far.

First half of #5116.

### How to test this PR?

`yarn test:renderer`


benoitf (approved):

jeffmaury (approved):
LGTM

Issue #4842, opened 2023-11-16
Move Pods page to table component


deboer-tim:
### Is your enhancement related to a problem? Please describe

We have a new table component, Pods page should switch.

### Describe the solution you'd like

The Pods page should be based on the new table component, and most columns should be sortable.

PR #5141, opened 2023-12-06
chore: simplify pod actions & callbacks

deboer-tim:
### What does this PR do?

It has been bugging me for a while that PodActions has a callback when an action affects a pod, and both PodList and PodDetails have essentially the same response of updating the pod UI object, and then PodList needs to trigger the UI to update. PodActions also passes a podInfoUI to functions when it already has the pod as a property.

When trying to extract columns to use the Table component this gets uglier because there is an intermediate component that needs to pass the callbacks through as well, so I figure it's a good time to simplify this a bit first.

This change just makes the PodAction update the pod UI object itself, and then uses a standard dispatch event to notify parents that something changed. Both PodList and PodDetails just need an on:update to trigger a Svelte update. Simpler/less code all around and less chance of PodList and PodDetails having different behaviour.

Note that there's one thing that still bothers me but is not fixed in this change: PodList and PodDetails use their own copies of PodInfoUI objects, so if you click Start in one view and then switch quickly, the pod is still stopped for a couple seconds in the other view. That is a separate problem but at least this change might make it marginally easier to fix in the future.

### Screenshot / video of UI

N/A

Cleanup prior to #4842.

### How to test this PR?

`yarn test:renderer` or use PodList/PodDetails. As a code simplification chore, there should be no change to behaviour or tests.

### DIFF
--- a/packages/renderer/src/lib/pod/PodActions.spec.ts
+++ b/packages/renderer/src/lib/pod/PodActions.spec.ts
@@ -30,13 +30,16 @@ const pod: PodInfoUI = {
   kind: 'podman',
 } as PodInfoUI;
 
-const errorCallback = vi.fn();
 const listContainersMock = vi.fn();
 const getContributedMenusMock = vi.fn();
+const updateMock = vi.fn();
 
 beforeEach(() => {
   (window as any).kubernetesDeletePod = vi.fn();
   (window as any).listContainers = listContainersMock;
+  (window as any).startPod = vi.fn();
+  (window as any).stopPod = vi.fn();
+  (window as any).restartPod = vi.fn();
   (window as any).removePod = vi.fn();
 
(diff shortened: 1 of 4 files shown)


benoitf (approved):
LGTM but should be merged after 1.6

benoitf (changes requested):
it is missing tests to cover the update

deboer-tim:
Added tests (and noticed/fix a bug, which new tests cover).

benoitf (approved):
main branch is open for new changes and tests have been added 👍

deboer-tim:
@jeffmaury @feloy could I please get a review?

deboer-tim:
(forgot I had a merge conflict, rebased and pushed now)

feloy (approved):

PR #5151, opened 2023-12-06
Revert "feat: update to podman v4.8.1"

benoitf:
### What does this PR do?
This reverts commit 361900642e73746e5be3c885c880d6dcc17ecaa6.

it seems it does not work on Windows with user networking mode

### DIFF
--- a/extensions/podman/src/podman.json
+++ b/extensions/podman/src/podman.json
@@ -1,18 +1,18 @@
 {
-  "version": "4.8.1",
+  "version": "4.8.0",
   "platform": {
     "win32": {
-      "version": "v4.8.1",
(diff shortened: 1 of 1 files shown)


cdrage (approved):

deboer-tim (approved):

gbraad:
According to https://github.com/containers/podman/issues/20921 the functionality wouldn't work, just not error out: https://github.com/containers/podman/issues/20921#issuecomment-1843699520 This soudns to me that reverting is not enough to deal with user expectation until this is fixed.

The issue will be fixed with PR: https://github.com/containers/podman/pull/20929

PR #5155, opened 2023-12-06
Extension installation through OCI image

xbabalov:
### What does this PR do?
Adds the option of installation through OCI image in case the original installation button is not available.

https://github.com/containers/podman-desktop/issues/5067

### How to test this PR?
yarn test:e2e:extension

### DIFF
--- a/packages/renderer/src/lib/docker-extension/PreferencesPageDockerExtensions.svelte
+++ b/packages/renderer/src/lib/docker-extension/PreferencesPageDockerExtensions.svelte
@@ -72,7 +72,8 @@ function deleteContribution(extensionName: string) {
         <input
           name="ociImage"
           id="ociImage"
-          aria-label="ociImage"
+          aria-label="OCI Image Name"
+          type="text"
           bind:value="{ociImage}"
           placeholder="Name of the Image"
           class="text-sm rounded-sm focus:ring-purple-500 focus:border-purple-500 block p-2.5 bg-charcoal-800 border-gray-900 placeholder-gray-400 text-white"
--- a/packages/renderer/src/lib/docker-extension/PreferencesPageDockerExtensions.svelte
+++ b/packages/renderer/src/lib/docker-extension/PreferencesPageDockerExtensions.svelte
@@ -72,6 +72,7 @@ function deleteContribution(extensionName: string) {
         <input
           name="ociImage"
           id="ociImage"
+          aria-label="ociImage"
           bind:value="{ociImage}"
           placeholder="Name of the Image"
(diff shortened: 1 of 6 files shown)


packages/renderer/src/lib/docker-extension/PreferencesPageDockerExtensions.svelte
```diff
@@ -72,6 +72,7 @@ function deleteContribution(extensionName: string) {
         <input
           name="ociImage"
           id="ociImage"
+          aria-label="ociImage"
```
odockal: I would choose: `OCI image name`

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
```diff
@@ -79,14 +79,15 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
...
-    <div class="bg-charcoal-700 mt-5 rounded-md p-3">
+    <div class="bg-charcoal-700 mt-5 rounded-md p-3" role="region" aria-label="OCI image installation box">
       <h1 class="text-lg mb-2">Install a new extension from OCI Image</h1>
 
       <div class="flex flex-col w-full">
         <div class="flex flex-row mb-2 w-full space-x-8 items-center">
           <input
             name="ociImage"
             id="ociImage"
+            aria-label="ociImage"
```
odockal: Same as above.

packages/renderer/src/lib/docker-extension/PreferencesPageDockerExtensions.svelte
odockal: Another note is that this input does not have a type. Since we are passing a text, we should add attribute `type="text"`. This is something ARIA spec. suggests (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role), but also we might consider the input of a `url` type which *warning* could introduce also a validation. So it accepts url form of the image name only. Now I would be very careful with `type="url"` as it probably also must accept local file (file:///path/to/image). Probably text suits better here.

odockal (changes requested):
I tried to run e2e-tests-main from your branch, but it is still failing: https://github.com/containers/podman-desktop/actions/runs/7119925938/job/19386080318

odockal:
I have run the e2e-tests-main [workflow](https://github.com/containers/podman-desktop/actions/runs/7250055764/job/19749383267) and except expected failing test due to an upstream issue (https://github.com/containers/podman-desktop/issues/5303) the extension tests are now passing!

Screenshot_20231218_170649

Thus I am approving this PR.

odockal (approved):

Issue #4843, opened 2023-11-16
Move Containers to table component


deboer-tim:
### Is your enhancement related to a problem? Please describe

We have a new table component, Containers page should switch.

### Describe the solution you'd like

The Containers page should be based on the new table component, and most columns should be sortable. Unlike the other pages we'll need some new capability in the Table component to be able to group (by pods and compose).

deboer-tim:
Still valid.

PR #5160, opened 2023-12-06
chore: simplify container/compose actions & callbacks

deboer-tim:
### What does this PR do?

The same change as #5141, but for containers & compose.

This change just makes the ContainerAction and ComposeAction update their UI objects directly, and then uses a standard dispatch event to notify parents that something changed. ContainerList, ContainerDetails, and ComposeDetails just need an on:update to trigger a Svelte update. Simpler/less code all around and less chance of the UI components having different behaviour.

Compared with pods I noticed and fixed a couple things:
- ComposeActions is updating the containers but not the compose group. Added this so now the group also goes into states like 'starting'.
- ComposeDetails was not listening for changes, so even after the change above there was no change in the UI. Fixed this.
- Both *Actions were frequently using the finally block to set the state after the action is completed. e.g. containers are set to RUNNING after start. In the best case this matches the state update we get from the container engine, so just setting twice. In the worst case the action fails, and we set the state anyway... Fixed.
- The ContainerList wasn't listening for changes to pods [*].

There is still one difference between pods and compose:
- When you start a compose group we set all containers (plus now the group) to STARTING, but for pods we only set the pod. I don't know if this is intentional, but I left it / unrelated for now.

and one unrelated gap [*]:
- The ContainerList passes newly created UI objects to PodActions and ComposeActions, so when those objects change the ContainerList doesn't update until the next notification through the respective store. This is the same problem listed in #5141 - if you start a container and then quickly go to Details it is still stopped. This problem is unrelated to / unaffected by this change.

Like #5141 this shouldn't merge until after 1.6.

### Screenshot / video of UI

N/A

Cleanup prior to #4843.

### How to test this PR?

yarn test:renderer or use ContainerList/ContainerDetails/ComposeDetails. As a code simplification chore, there should be no change to behaviour or tests. (although note some minor bugs fixed, e.g. compose groups will show 'starting' when started from compose details)

### DIFF
--- a/packages/renderer/src/lib/compose/ComposeActions.svelte
+++ b/packages/renderer/src/lib/compose/ComposeActions.svelte
@@ -5,18 +5,16 @@ import { router } from 'tinro';
 import ListItemButtonIcon from '../ui/ListItemButtonIcon.svelte';
 import DropdownMenu from '../ui/DropdownMenu.svelte';
 import FlatMenu from '../ui/FlatMenu.svelte';
-import type { ContainerInfoUI } from '../container/ContainerInfoUI';
 import type { Menu } from '../../../../main/src/plugin/menu-registry';
 import ContributionActions from '/@/lib/actions/ContributionActions.svelte';
-import { onMount } from 'svelte';
+import { createEventDispatcher, onMount } from 'svelte';
 import { MenuContext } from '../../../../main/src/plugin/menu-registry';
 
 export let compose: ComposeInfoUI;
 export let dropdownMenu = false;
 export let detailed = false;
 
-export let inProgressCallback: (containers: ContainerInfoUI[], inProgress: boolean, state?: string) => void = () => {};
-export let errorCallback: (erroMessage: string) => void = () => {};
+const dispatch = createEventDispatcher<{ update: ComposeInfoUI }>();
 
 const composeLabel = 'com.docker.compose.project';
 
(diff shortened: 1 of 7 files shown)


packages/renderer/src/lib/compose/ComposeActions.svelte
```diff
@@ -25,46 +23,76 @@ onMount(async () => {
...
+  compose.actionInProgress = inProgress;
+  // reset error when starting task
+  if (inProgress) {
+    compose.actionError = '';
+  }
+  if (state) {
+    compose.status = state;
+  }
+
+  compose.containers.forEach(container => {
```
benoitf:
```suggestion
  for (const container of compose.containers) {
```

benoitf: (not a required change)

benoitf (approved):
note sure if we're not missing some component tests

benoitf (changes requested):
we're missing some tests there

ContainerActions has no tests
ComposeActions has no tests

packages/renderer/src/lib/compose/ComposeActions.svelte
deboer-tim: Yes, this is better, updated when I added tests.

deboer-tim:
Added tests in new commit.

benoitf (approved):

benoitf:
PR needs rebase/solving conflicts

benoitf:
I think it can be merged if conflict is solved

deboer-tim:
Rebased and merge conflict fixed.

benoitf (approved):
approving again

feloy (approved):

PR #5168, opened 2023-12-07
fix: reset loggerhandlerKey after restarting machine (#5150)

### LINKED ISSUE #5150: Success icon displayed after podman machine restart failure
### Bug description

When restarting a podman machine with the tasks area open, the operation is shown as succesfull even after it fails later in the execution. This is only shown for this operation afaik and is corrected once the tasks pop-up is reopened.

image

### Operating system

Windows 10

### Installation Method

### Version

next (development version)

### Steps to reproduce

1. open resources page
2. create a podman machine
3. open its details
4. open tasks bar on the bottom right
5. click the restart button
6. the task shown as succesfull even after it fails

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

it resets the loggerHandlerKey which is used when restarting a podman machine. As that was not reset the startAction was called twice, resulting in a failure the second time as the machine was already running

### Screenshot / video of UI

N/A

it resolves #5150 

### How to test this PR?

1. restart a podman machine from its details page

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesContainerConnectionRendering.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesContainerConnectionRendering.spec.ts
@@ -288,3 +288,80 @@ test('Expect to see error message if action fails', async () => {
   // expect to see the delete failed button
   expect(deleteFailedButton).toBeInTheDocument();
 });
+
+test('Expect startContainerProvider to only be called once when restarting', async () => {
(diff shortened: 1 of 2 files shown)


benoitf (changes requested):
thanks

could you please add a test where we reproduce the bug so if someone do some a refactoring, we won't have again this problem ?

check we don't have the two actions loop

benoitf (approved):
thanks @lstocchi

PR #5174, opened 2023-12-07
feat: restart kubernetes pod

### LINKED ISSUE #4402: K8s pods actions

Pods running on Kubernetes do not have appropriate menu items for common actions.

Make sure the following actions exist in the pods list for pods running on Kubernetes:

- Open terminal
- Open browser
- Open logs
- Restart
- pause/stop/start if possible

_No response_

### Additional context

_No response_

vzhukovs:
### What does this PR do?

The purpose of these changes is to provide users with the ability to restart pods that are deployed in a Kubernetes environment. The restart functionality supports configurations such as Deployment, ReplicaSet, StatefulSet, Job, CronJob, as well as regular pods that are created without using a controller or how they called, manually created.

Manually created Pods restart through deleting and then recreating pods configuration. Controller-based Pods (Deployment, ReplicaSet, StatefulSet) perform restart through changing scale configuration to 0 and then to initial scale value. Job based Pods restart through re-creating the Job configuration.

### Screenshot / video of UI

Снимок экрана 2024-04-05 в 22 24 50

More screencasts you can find in How to test section.

partially #4402 

### How to test this PR?

Below are test configurations that can be deployed in Kubernetes and used to test the pod restart capability. It's important to note one key aspect: in the OpenShift environment, you can deploy a CronJob, but it cannot be triggered manually as you can do in the Minikube dashboard.

1. **Manually created Pods:**

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox-pod
spec:
  containers:
  - name: busybox-container
    image: busybox
    args:
    - /bin/sh
    - -c
    - "sleep 3600"
```
_Description: this and below configurations will contain the only one Pod and three containers in each such Pod. All containers are based on the busybox image._

**Minikube Demo:**

https://github.com/containers/podman-desktop/assets/1968177/f32d2c1b-95b9-4305-ad7f-34da9ceab200

**OpenShift Local Demo:**

https://github.com/containers/podman-desktop/assets/1968177/bf8f2e70-facc-4679-9c66-fcbbb34021da

2. **Deployment Configuration:**

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: busybox-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: busybox
  template:
    metadata:
      labels:
        app: busybox
    spec:
      containers:
      - name: busybox-container-1
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      - name: busybox-container-2
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      - name: busybox-container-3
        image: busybox
        command: ["sleep"]
        args: ["3600"]
```

**Minikube Demo:**

https://github.com/containers/podman-desktop/assets/1968177/5e67ce5d-1806-4843-88fa-3f43590743e6

**OpenShift Local Demo:**

https://github.com/containers/podman-desktop/assets/1968177/9b5adfb5-9863-464b-9bf9-24632b04b7ee

3. **ReplicaSet Configuration:**

```yaml
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: busybox-replicaset
spec:
  replicas: 1
  selector:
    matchLabels:
      app: busybox
  template:
    metadata:
      labels:
        app: busybox
    spec:
      containers:
      - name: busybox-container-1
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      - name: busybox-container-2
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      - name: busybox-container-3
        image: busybox
        command: ["sleep"]
        args: ["3600"]
```

**Minikube Demo:**

https://github.com/containers/podman-desktop/assets/1968177/3f328dd1-d512-4895-952f-23fff3adc23a

**OpenShift Local Demo:**

https://github.com/containers/podman-desktop/assets/1968177/538c4209-0001-40d2-b7cd-9db73b0d50e5

4. **StatefulSet Configuration:**

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: busybox-statefulset
spec:
  serviceName: "busybox"
  replicas: 1
  selector:
    matchLabels:
      app: busybox
  template:
    metadata:
      labels:
        app: busybox
    spec:
      containers:
      - name: busybox-container-1
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      - name: busybox-container-2
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      - name: busybox-container-3
        image: busybox
        command: ["sleep"]
        args: ["3600"]
```

**Minikube Demo:**

https://github.com/containers/podman-desktop/assets/1968177/c2bb67f7-5233-47c0-877c-3c572bee66ad

**OpenShift Local Demo:**

https://github.com/containers/podman-desktop/assets/1968177/5b05c69d-f9fc-4365-ae9b-e073a517ee74

5. **Job Configuration:**

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: busybox-job
spec:
  template:
    spec:
      containers:
      - name: busybox-container-1
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      - name: busybox-container-2
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      - name: busybox-container-3
        image: busybox
        command: ["sleep"]
        args: ["3600"]
      restartPolicy: Never
  backoffLimit: 4
```

**Minikube Demo:**

https://github.com/containers/podman-desktop/assets/1968177/14cf511f-e112-4cdc-a02c-fc0cf2e1c1d6

**OpenShift Local Demo:**

https://github.com/containers/podman-desktop/assets/1968177/34fa687b-f000-471e-a856-5b8b167d08d2

6. **CronJob Configuration:**

```yaml
apiVersion: batch/v1
kind: CronJob
metadata:
  name: busybox-cronjob
spec:
  schedule: "0 * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: busybox-container-1
            image: busybox
            command: ["sleep"]
            args: ["3600"]
          - name: busybox-container-2
            image: busybox
            command: ["sleep"]
            args: ["3600"]
          - name: busybox-container-3
            image: busybox
            command: ["sleep"]
            args: ["3600"]
          restartPolicy: OnFailure
```

**Minikube Demo:**

https://github.com/containers/podman-desktop/assets/1968177/acd6bc1a-f833-4b4e-b600-9b0f2ad3cf6d


deboer-tim:
Worked for me, code looks good, but wondering if we should really call this 'reset' or something.

Multiple failed tests but they don't appear to be related to the code, so rerunning them.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -657,6 +657,46 @@ export class KubernetesClient {
...
+          const coreApi = this.kubeConfig.makeApiClient(CoreV1Api);
+          // To restart pod we need to delete existed pod and create the new one
+          await coreApi.deleteNamespacedPod(name, ns);
+
+          pod.metadata.resourceVersion = undefined;
+          pod.metadata.creationTimestamp = new Date();
+
+          // Wait the pod to be successfully deleted
+          // eslint-disable-next-line no-constant-condition
+          while (true) {
```
jeffmaury: I think we should put a time limit

vzhukovs: I've updated the while statement with a timed out request. Set the time out to one minute by default, not 100% sure if this value is reasonable, but a good starting point.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -664,6 +664,56 @@ export class KubernetesClient {
...
+      if (ns && connected) {
+        const pod = await this.readNamespacedPod(name, ns);
+        if (pod?.metadata) {
+          const coreApi = this.kubeConfig.makeApiClient(CoreV1Api);
+          // To restart pod we need to delete existed pod and create the new one
+          await coreApi.deleteNamespacedPod(name, ns);
+
+          pod.metadata.resourceVersion = undefined;
+          pod.metadata.creationTimestamp = new Date();
+
```
feloy: I would recreate a metadata field, with only the expected values: name, namespace, labels, annotations. The other ones (including creationTimestamp) are not expected to be set by the user, but by the system (https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#ObjectMeta)

You can also delete the status field.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -664,6 +664,56 @@ export class KubernetesClient {
...
+    }
+  }
+
+  private async waitForPodDeletion(coreApi: CoreV1Api, name: string, ns: string, timeoutMs: number) {
+    try {
+      await coreApi.readNamespacedPod(name, ns);
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
+      if (error.response.statusCode === 404) {
+        return; // Pod deleted
```
feloy: Here, you may want to use a DELETE instead of a GET, so the api won't have to compute the data to return. Also, you could use a Precondition with the uid of the pod, so you are sure to not consider another pod with the same name created in the meantime by someone else. 

Here is described how `kubectl delete` has been implemented: https://github.com/kubernetes/kubernetes/issues/42594

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: @feloy I've updated the PR, now new pod configuration doesn't contain status field and the only necessary metadata information

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: Unfortunately from the code we can not use `--wait-for-deletion` option, so I tried to workaround with a bit different approach, in the code the only pod status is checking which also avoids the computation of the pod data. Also added additional checking to avoid creating the pod with the same name which is restarting at the moment.

vzhukovs:
had some issues due to updating the current PR, that's why there is a few events about closing and opening, but it ready to review

jeffmaury (commented):
I'm wondering why it's not implemented as scale to 0 and scale to original number ?

vzhukovs:
> I'm wondering why it's not implemented as scale to 0 and scale to original number ?

@jeffmaury, this also a correct way to restart the pod. During the investigation, I've found that there are four legal approach to restart the running pod:

1. To update the scale configuration.
2. To delete the pod and then to create the same one.
3. To update the environment variables in Deployment (requires to have existed deployment).
4. To use `kubectl rollout` command to sequentially restart the each container in the pod. (The less popular approach, tbh).

Using scale command, we need to be sure that we have deployment, although it's possible to create the pod without creating deployment. That will increase the kube api calls. The main idea is to have less api calls here.

jeffmaury:
> > I'm wondering why it's not implemented as scale to 0 and scale to original number ?
> 
> @jeffmaury, this also a correct way to restart the pod. During the investigation, I've found that there are four legal approach to restart the running pod:
> 
>     1. To update the scale configuration.
> 
>     2. To delete the pod and then to create the same one.
> 
>     3. To update the environment variables in Deployment (requires to have existed deployment).
> 
>     4. To use `kubectl rollout` command to sequentially restart the each container in the pod. (The less popular approach, tbh).
> 
> 
> Using scale command, we need to be sure that we have deployment, although it's possible to create the pod without creating deployment. That will increase the kube api calls. The main idea is to have less api calls here.

Asking @feloy to comment if there one recommended over the other ones

feloy:
> > > I'm wondering why it's not implemented as scale to 0 and scale to original number ?
> > 
> > 
> > @jeffmaury, this also a correct way to restart the pod. During the investigation, I've found that there are four legal approach to restart the running pod:
> > ```
> > 1. To update the scale configuration.
> > 
> > 2. To delete the pod and then to create the same one.
> > 
> > 3. To update the environment variables in Deployment (requires to have existed deployment).
> > 
> > 4. To use `kubectl rollout` command to sequentially restart the each container in the pod. (The less popular approach, tbh).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Using scale command, we need to be sure that we have deployment, although it's possible to create the pod without creating deployment. That will increase the kube api calls. The main idea is to have less api calls here.
> 
> Asking @feloy to comment if there one recommended over the other ones

Scaling the pod to 0 involves that we know the kind of its manager (ReplicaSet, StatefulSet, etc, or any other 3rd party operator), and access its scale subresource. For 3rd-party operators, there is a chance this scale subresource is not implemented. That would be a similar problem if we want to restart it by updating the podTemplate on its manager (point 3).

Killing the pod seems to me the most generic approach, but for it to be restarted, we need to be check if it is managed by some controller (ReplicaSet, statefulset, etc) or not. If it is managed, I don't think necessary to restart it ourselves, as the manager will take care of that (and it is the manager, so I would not interfere with it).

If it has been created as a standalone pod, it will never be restarted automatically (metadata.ownerReferences would help determining if its standalone of not), and in this case, we would need to recreate it.

benoitf:
moving to draft

vzhukovs:
@jeffmaury @feloy @benoitf I've updated the PR and description, now it's ready to be reviewed again

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -120,6 +123,11 @@ const DEFAULT_NAMESPACE = 'default';
 
 const FIELD_MANAGER = 'podman-desktop';
 
+export interface PodCreationSource {
+  isManuallyCreated: boolean;
+  controllerType?: 'Deployment' | 'ReplicaSet' | 'StatefulSet' | 'Job' | 'DaemonSet' | 'CronJob';
+}
```
```suggestion
const SCALABLE_CONTROLLER_TYPES = ['Deployment', 'ReplicaSet', 'StatefulSet'];

export type ScalableControllerType = typeof SCALABLE_CONTROLLER_TYPES[number];

export type ControllerType = ScalableControllerType | 'Job' | 'DaemonSet' | 'CronJob' | undefined;

function isScalableControllerType(string: unknown): string is ScalableControllerType {
  return typeof string === 'string' && string in SCALABLE_CONTROLLER_TYPES;
}

export interface PodCreationSource {
  isManuallyCreated: boolean;
  controllerType: ControllerType;
}
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+
+        const controller = this.getPodController(pod.metadata);
+        const controllerName = controller!.name;
+
+        if (['Deployment', 'ReplicaSet', 'StatefulSet'].includes(creationSource.controllerType)) {
+          const controllerType: 'Deployment' | 'ReplicaSet' | 'StatefulSet' = creationSource.controllerType as
+            | 'Deployment'
+            | 'ReplicaSet'
+            | 'StatefulSet';
+          await this.scaleControllerToRestartPods(ns, controllerName, controllerType);
```
```suggestion
       ```suggestion
        if (isScalableControllerType(creationSource.controllerType)) {
          await this.scaleControllerToRestartPods(ns, controllerName, creationSource.controllerType)
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+      }
+    }
+
+    return false;
+  }
+
+  protected async scaleControllerToRestartPods(
+    namespace: string,
+    controllerName: string,
+    controllerType: 'Deployment' | 'ReplicaSet' | 'StatefulSet',
```
```suggestion
    controllerType: ScalableControllerType,
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+      return {
+        isManuallyCreated: false,
+        controllerType: controller.kind as
+          | 'Deployment'
+          | 'ReplicaSet'
+          | 'StatefulSet'
+          | 'Job'
+          | 'DaemonSet'
+          | 'CronJob'
+          | undefined,
```
```suggestion
        controllerType: controller.kind
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+      };
+    }
+
+    return {
+      isManuallyCreated: true,
+      controllerType: undefined,
```
dgolovin, line 1724: The way it is used right now if controller type is undefined it means it is manually created. Might be better to use ScalableControllerType value not PodCreationSource interface.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+    };
+  }
+
+  protected getPodController(podMetadata: V1ObjectMeta): V1OwnerReference | undefined {
+    // possible check is also in pod-template-hash label:
+    // pod.metadata?.labels && 'pod-template-hash' in pod.metadata.labels
+    if (podMetadata.ownerReferences && podMetadata.ownerReferences.length > 0) {
+      return podMetadata.ownerReferences.find((ref: V1OwnerReference) => ref.controller === true);
+    }
```
```suggestion
    return podMetadata.ownerReferences?.find((ref: V1OwnerReference) => ref.controller === true);
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+
+    return undefined;
```
```suggestion
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+        namespace,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        selector,
+      );
+      if (podList.items.length === 0) {
+        return true;
+      }
```
dgolovin, line 1705: Should be using watch with resourceVersion, to avoid use case when pod deleted before while loop started (see https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes).

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+      );
+    }
+
+    delete existingJob.metadata!.creationTimestamp;
+    delete existingJob.metadata!.resourceVersion;
+    delete existingJob.metadata!.selfLink;
+    delete existingJob.metadata!.uid;
+    delete existingJob.metadata!.ownerReferences;
+    delete existingJob.status;
+    delete existingJob.spec!.selector;
```
dgolovin, line 1642:
```suggestion
  _.omit(existingJob.metadata!, ['creationTimestamp', 'resourceVersion', 'selfLink', 'uid', 'ownerReferences', 'status', 'selector']);
```

you have to install @types/lodash package as `yarn -D add @types/lodash -W` for code above to work.
the same applies to the code below.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+        if (typeof error === 'object' && 'response' in error) {
+          const axiosError = error as { response: { statusCode: number } };
+          if (axiosError.response.statusCode === 404) {
+            return true;
+          } else {
+            throw e;
+          }
+        } else {
+          throw e;
+        }
```
```suggestion
        if (typeof error === 'object' && 'response' in error) {
          const axiosError = error as { response: { statusCode: number } };
          if (axiosError.response.statusCode === 404) {
            return true;
          }
        }
        throw e;
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+  protected async restartManuallyCreatedPod(name: string, namespace: string, pod: V1Pod): Promise<void> {
+    const coreApi = this.kubeConfig.makeApiClient(CoreV1Api);
+    await coreApi.deleteNamespacedPod(name, namespace);
+
+    const isDeleted = await this.waitForPodDeletion(coreApi, name, namespace);
+    if (!isDeleted) {
+      throw new Error(`pod "${name}" in namespace "${namespace}" was not deleted within the expected timeframe`);
+    }
+
+    delete pod.metadata?.resourceVersion;
```
dgolovin, line 1506: maybe use lodash.omit here

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+    timeout: number = 60000,
+  ): Promise<boolean> {
+    const startTime = Date.now();
+
+    while (Date.now() - startTime < timeout) {
+      try {
+        await coreApi.readNamespacedPodStatus(name, namespace);
+        await new Promise(resolve => setTimeout(resolve, 1000));
+      } catch (e) {
+        const error = e ?? {};
```
dgolovin, line 1529: Can it be `null` here?

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1319,304 @@ export class KubernetesClient {
...
+        if (typeof error === 'object' && 'response' in error) {
+          const axiosError = error as { response: { statusCode: number } };
+          if (axiosError.response.statusCode === 404) {
+            return true;
+          } else {
+            throw e;
+          }
+        } else {
+          throw e;
+        }
```
```suggestion
        if (typeof error === 'object' && 'response' in error) {
          const axiosError = error as { response: { statusCode: number } };
          if (axiosError.response.statusCode === 404) {
            return true;
          }
        } 
        throw e;
```

dgolovin (changes requested):

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated, but instead using `in` operator we should use `.includes()`

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: `e` by default has `unknown`, checking for nullability is denied by es-lint, but when we're checking for `undefined` we're getting `TS18047: error is possibly null`, closed loop in the end. providing empty object is the workaround for typescript to follow the es-lint rules

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: not really sure, that we need to install this library just for one stylish usecase

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: you mean, that if Deployment has two or more pods run simultaneously?

> Why not just delete the pod and let controller to get it back according to replicas number set in descriptor?

there was an agreement to operate with scale property only and left approach that deletes pod to manually created ones

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: if the pod deleted before while loop started, the method would just return true, indicating, that pod had been already deleted. should we refactor this method to use watch api?

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1311,4 +1326,283 @@ export class KubernetesClient {
...
+        await new Promise(resolve => setTimeout(resolve, 1000));
+      } catch (e) {
+        const error = e ?? {};
+        if (typeof error === 'object' && 'response' in error) {
+          const axiosError = error as { response: { statusCode: number } };
+          if (axiosError.response.statusCode === 404) {
+            return true;
+          }
+        }
+        throw e;
```
dgolovin, line 1536: nit: you are correct, but still can be shorter and without additional constant. It is up to you just to resolve this conversation without changes.

```suggestion
        if (e && typeof e === 'object' && 'response' in e) {
          const axiosError = e as { response: { statusCode: number } };
          if (axiosError.response.statusCode === 404) {
            return true;
          }
        }
        throw e;
```

packages/main/src/plugin/kubernetes-client.ts
dgolovin: Works for me as it is, watch would be cleaner, but it is optional.

dgolovin (approved):
Works for me, but I would wait for another approval before pushing it to main.

I was able to run Deployment from example and restart the pod.
I had other problems, probably not related to this PR. I was running on latest OpenShift Local and I had first import the busybox image to cluster to make Deployment to work.
Containers are not visible in Podman Desktop Containers view.

vzhukovs:
@benoitf @jeffmaury can we proceed with merge of the current PR?

nichjones1:
@benoitf @jeffmaury following up on this issue.  Can Vlad proceed with this PR?

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -120,6 +123,18 @@ const DEFAULT_NAMESPACE = 'default';
 
 const FIELD_MANAGER = 'podman-desktop';
 
+const SCALABLE_CONTROLLER_TYPES = ['Deployment', 'ReplicaSet', 'StatefulSet'];
+export type ScalableControllerType = (typeof SCALABLE_CONTROLLER_TYPES)[number];
+export type ControllerType = ScalableControllerType | 'Job' | 'DaemonSet' | 'CronJob' | undefined;
+function isScalableControllerType(string: unknown): string is ScalableControllerType {
```
jeffmaury, line 133: Maybe rename string parameter to type ?

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1303,4 +1318,283 @@ export class KubernetesClient {
...
+
+      const creationSource = this.checkPodCreationSource(pod.metadata);
+      if (creationSource.isManuallyCreated) {
+        await this.restartManuallyCreatedPod(name, ns, pod);
+      } else {
+        if (!creationSource.controllerType) {
+          throw new Error('unable to restart controlled pod');
+        }
+
+        const controller = this.getPodController(pod.metadata);
```
jeffmaury, line 1480: What if controller is undefined ?

jeffmaury (approved):
LGTM

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1303,4 +1318,283 @@ export class KubernetesClient {
...
+      currentReplicas = currentDeployment.spec?.replicas || 1;
+    } else if (controllerType === 'ReplicaSet') {
+      const { body: currentReplicaSet } = await appsApi.readNamespacedReplicaSet(controllerName, namespace);
+      currentReplicas = currentReplicaSet.spec?.replicas || 1;
+    } else if (controllerType === 'StatefulSet') {
+      const { body: currentStatefulSet } = await appsApi.readNamespacedStatefulSet(controllerName, namespace);
+      currentReplicas = currentStatefulSet.spec?.replicas || 1;
+    }
+
+    await this.scaleController(appsApi, namespace, controllerName, controllerType, 0);
```
feloy, line 1563: This works for scalable resource with only 1 replica, right? If there are more than 1 replica, all will be killed, but this is not what we want

vzhukovs: Yes, it should scale the replica to 0 and rescale to the original number. I think, we can provide enhancement for this in the next PR as the next step.

PR #5179, opened 2023-12-07
feat: add cleanup for podman provider

### LINKED ISSUE #5016: Implement clean API for podman provider

as part of the global epic to be able to cleanup "old state", handle cleanup of data for podman.

Remove all folders used by podman, stop/kill processes, etc.

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Implements cleanup for podman provider on macOS and Windows (for podman machines)

depends on
- [x] https://github.com/containers/podman-desktop/pull/5139

fixes https://github.com/containers/podman-desktop/issues/5016

### DIFF
--- a/extensions/podman/src/podman-cleanup-macos.ts
+++ b/extensions/podman/src/podman-cleanup-macos.ts
@@ -1,0 +1,58 @@
+import type { ProviderCleanupExecuteOptions } from '@podman-desktop/api';
+import { process } from '@podman-desktop/api';
+import psList from 'ps-list';
+import { homedir } from 'node:os';
+import { resolve as pathResolve } from 'node:path';
+import { AbsPodmanCleanup } from './podman-cleanup-abstract';
+
+// Handle cleanup of Podman on macOS
+export class PodmanCleanupMacOS extends AbsPodmanCleanup {
+  async killProcess(processId: number): Promise<void> {
+    await process.exec('/bin/kill', ['-9', String(processId)]);
+  }
+
+  async stopPodmanProcesses(options: ProviderCleanupExecuteOptions): Promise<void> {
+    // grab process
+    const pidsToKill = await this.getProcessesToKill(['gvproxy', 'qemu', 'vfkit']);
+
+    // remove any duplicates in PID
+    const uniquePidsToKill = [...new Set(pidsToKill)];
+
+    // kill each process
+    for (const pid of uniquePidsToKill) {
+      try {
+        options.logger.log(`Killing process ${pid}`);
+        await this.killProcess(pid);
+      } catch (error) {
+        options.logger.error('Error killing process', error);
+      }
+    }
+  }
+
+  // folders to remove:
+  // ~/.config/containers
+  // ~/.local/share/containers/podman
+  getFoldersToDelete(): string[] {
+    const configContainersFolder = pathResolve(homedir(), '.config/containers');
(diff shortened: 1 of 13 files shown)


benoitf:

extensions/podman/src/podman-cleanup-macos.ts
```diff
@@ -0,0 +1,58 @@
...
+import { process } from '@podman-desktop/api';
+import psList from 'ps-list';
+import { homedir } from 'node:os';
+import { resolve as pathResolve } from 'node:path';
+import { AbsPodmanCleanup } from './podman-cleanup-abstract';
+
+// Handle cleanup of Podman on macOS
+export class PodmanCleanupMacOS extends AbsPodmanCleanup {
+  async killProcess(processId: number): Promise<void> {
+    await process.exec('/bin/kill', ['-9', String(processId)]);
```
jeffmaury: Would be better to use SIGTERM or SIGSTOP instead of SIGKILL as it allow cleanup

extensions/podman/src/podman-cleanup-windows.ts
```diff
@@ -0,0 +1,75 @@
+import type { ProviderCleanupExecuteOptions } from '@podman-desktop/api';
+import { process } from '@podman-desktop/api';
+
+import { homedir } from 'node:os';
+import { resolve as pathResolve } from 'node:path';
+import { AbsPodmanCleanup } from './podman-cleanup-abstract';
+
+export class PodmanCleanupWindows extends AbsPodmanCleanup {
+  async stopPodmanProcesses(options: ProviderCleanupExecuteOptions): Promise<void> {
+    // stop all running podman machine if there are some
```
jeffmaury, line 10: You should kill win-sshproxy.exe as it owns the pipes and I noticed that after I did cleanup, Docker compatibilty does not work anymore

jeffmaury (changes requested):

extensions/podman/src/podman-cleanup-macos.ts
benoitf: does it really matter as we do the cleanup ?

feloy: I also think that it would be better to try to use SIGTERM/SIGSTOP first, and use SIGKILL after a grace period. The process may want to cleanup some temporary resources it has created, that we don't know about

benoitf: I'm not sure you meant SIGSTOP ?

feloy: Effectively, SIGSTOP doesn't seem appropriate. SIGTERM (kill) first, then SIGKILL (kill -9) later.

extensions/podman/src/podman-cleanup-windows.ts
benoitf: added kill of the processes on Windows side

extensions/podman/src/podman-cleanup-macos.ts
benoitf: added usage of SIGTERM instead of SIGKILL which I still believe is a bad idea in this context as it should be fast, robust and anyway we remove data after that

benoitf:
@jeffmaury you should be able to test again with my new changes

I've checked on Windows and it works as expected

jeffmaury (approved):
LGTM. The only concern I have is that processes are identified by they short names so other installations may use the same name

PR #5180, opened 2023-12-07
feat: add cleanup UI

### LINKED ISSUE #5022: Adds clean/purge data button in the troubleshooting page

Add cleanup operation for each provider supporting the handling.

I would say that we can select which provider we might want to cleanup (only the one supporting may be listed ?)

_No response_

### Additional context

part of epic https://github.com/containers/podman-desktop/issues/4994

benoitf:

### What does this PR do?
Add cleanup section in Troubleshooting page

depends on:
- [x] https://github.com/containers/podman-desktop/pull/5139
- [x] https://github.com/containers/podman-desktop/pull/5177

Note:  while UI is working, after UX meetup I'll need to include it as part of a tab but it allows to test the whole workflow for now

### Screenshot / video of UI

image

fixes https://github.com/containers/podman-desktop/issues/5022

### How to test this PR?

Go to troubleshooting page and click on 'cleanup'

all podman machine stuff should be gone, creating a fresh machine after this step should work

### DIFF
--- a/packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
+++ b/packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
@@ -46,7 +46,7 @@ async function cleanup() {
 <div class="flex flex-row items-center">
   <div>
     <div class="text-gray-700 flex flex-row items-center">Clean / Purge data</div>
-    <div class="text-gray-900 text-xs flex flex-row items-center">
+    <div class="text-gray-900 text-sm flex flex-row items-center pt-1">
       <Fa class="pr-1" size="9" icon="{faWarning}" />Proceeding with this action may result in data loss, including
       existing volumes, containers, images, etc.
     </div>
@@ -56,10 +56,9 @@ async function cleanup() {
     <Button
       type="danger"
       on:click="{() => openCleanupDialog()}"
-      disabled="{cleanupInProgress}"
       inProgress="{cleanupInProgress}"
       aria-label="Cleanup"
-      icon="{faBroom}">Cleanup</Button>
+      icon="{faBroom}">Cleanup / Purge data</Button>
   </div>
 
   <div>
--- a/packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
+++ b/packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
@@ -1,0 +1,72 @@
+<script lang="ts">
+import { faBroom, faWarning } from '@fortawesome/free-solid-svg-icons';
+import type { ProviderInfo } from '../../../../main/src/plugin/api/provider-info';
+import Button from '../ui/Button.svelte';
+import Fa from 'svelte-fa';
+
+export let providers: ProviderInfo[] = [];
+let providerIdsWithCleanup: string[] = [];
+$: providerIdsWithCleanup = providers.filter(provider => provider.cleanupSupport).map(provider => provider.internalId);
+
+let cleanupInProgress = false;
+
+let cleanupFailures: string[] = [];
+
+async function openCleanupDialog(): Promise<void> {
+  let message = 'This action may delete data. Proceed ?';
+
+  const result = await window.showMessageBox({
+    title: 'Cleanup',
+    message: message,
+    buttons: ['Yes', 'No'],
+  });
+
+  if (result?.response === 0) {
+    cleanup();
+  }
+}
+
+async function cleanup() {
+  try {
+    cleanupInProgress = true;
+    cleanupFailures = [];
+    await window.cleanupProviders(providerIdsWithCleanup, Symbol('cleanup-provider'), (_key, _eventName, args) => {
+      // log into the console as no UI is available to display logs
+      console.log('cleanup event', args);
+    });
+  } catch (e) {
+    console.error(e);
+    cleanupFailures.push(String(e));
+  } finally {
+    cleanupInProgress = false;
+  }
+}
+</script>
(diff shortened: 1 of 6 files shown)


benoitf:

packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
```diff
@@ -0,0 +1,72 @@
...
+  } finally {
+    cleanupInProgress = false;
+  }
+}
+</script>
+
+<div class="flex flex-row items-center">
+  <div>
+    <div class="text-gray-700 flex flex-row items-center">Clean / Purge data</div>
+    <div class="text-gray-900 text-xs flex flex-row items-center">
```
deboer-tim: Maybe text-sm? This is a really important message and it's very small. Could also maybe use a pt-1 to separate from the title.

packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
```diff
@@ -0,0 +1,72 @@
...
+      <Fa class="pr-1" size="9" icon="{faWarning}" />Proceeding with this action may result in data loss, including
```
deboer-tim, line 50: The genericnmessage and using 'may' leaves me a bit uneasy, and it feels vague. What is this 'data' that it's going to be cleaned up? Are my podman/minikube/compose installs ok? Containers/images on kube providers like Dev Sandbox? I think we need to be clearer on what this will (or won't) purge.

packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
```diff
@@ -0,0 +1,72 @@
...
+      existing volumes, containers, images, etc.
+    </div>
+  </div>
+
+  <div class="flex flex-1 justify-end">
+    <Button
+      type="danger"
+      on:click="{() => openCleanupDialog()}"
+      disabled="{cleanupInProgress}"
```
deboer-tim: You don't need to set disabled here, Buttons are automatically disabled while in progress.

packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
```diff
@@ -0,0 +1,72 @@
...
+    cleanupFailures.push(String(e));
+  } finally {
+    cleanupInProgress = false;
+  }
+}
+</script>
+
+<div class="flex flex-row items-center">
+  <div>
+    <div class="text-gray-700 flex flex-row items-center">Clean / Purge data</div>
```
deboer-tim, line 48: Since the button and dialog are called 'cleanup', maybe this should be too? i.e. "Cleanup / Purge data"

deboer-tim (changes requested):
UI issues are minor and could be done later, but the message is a little too vague/I don't know what it's going to do.

feloy:
Could we be more explicit on what exactly will be deleted? As a user, I would be afraid to click without knowing exactly what will be deleted

benoitf:
@feloy it depends on each plug-in contributing something to delete

it may be possible to list everything but I'm not sure the overall experience would be better.

deboer-tim:
> @feloy it depends on each plug-in contributing something to delete
> 
> it may be possible to list everything but I'm not sure the overall experience would be better.

That's two of us who don't even want to click it to test because we're worried about what it might delete :). Could each contributing extension provide a string saying what they'll do (and in the future we could provide UI to select individual ones)? I think it either needs to give enough of a hint (e.g. it is only going to delete all podman machines) or it's going to be more of a 'you should only click this when you're specifically asked to do so, and the person who told you to will tell you what it'll do'.

benoitf:
I'm fine if you don't click.

Do you will click on Docker ?
image

benoitf:
I think it's where we're trying to overengineering stuff. When you hit a red button, it's like the last thing you do if you're not able to have something working.

It's not a button like 'remove unused containers'

packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
benoitf: I think first step was to do cleanup on supported extensions. So if devsanbox is not supporting it, it won't do anything.

it looks like enhancements

packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.svelte
benoitf: I think it was the first title but then moved to a shorter one
so going back to `Cleanup / Purge data`

benoitf:
PR updated on UI side

deboer-tim:
> I think it's where we're trying to overengineering stuff. When you hit a red button, it's like the last thing you do if you're not able to have something working.
> 
> It's not a button like 'remove unused containers'

I totally get this and don't want to go overboard, but we already have extensions for several things - so if I click it tomorrow it could delete all my perfectly working podman machines and not fix the OpenShift Local install that I was actually having problems with. Except I won't click it in the first place, because I have something running in Minikube that I'm not willing to risk atm...

I'll reset my review because the technical side is fine and this will indeed help someone, I'm just worried it's a vague extreme desperation button that would be used more if there was a little scoping.

deboer-tim (approved):

benoitf:
yes I'll create an issue to improve the cleanup

overall idea was that you could review each extension performing a cleanup and select/deselect

view what it will do, etc before confirming

just need a design

PR #5182, opened 2023-12-08
fix: avoid to crash if configuration is invalid

benoitf:
### What does this PR do?
do not stop if config file is corrupted

could be related to https://github.com/containers/podman-desktop/issues/4649

### How to test this PR?

unit test provided

manual testing:
Edit podman settings file and use invalid values ( ~/.local/share/containers/podman-desktop/configuration/settings.json)

before the patch: unable to start
after the patch: able to start

### DIFF
--- a/packages/main/src/plugin/autostart-engine.spec.ts
+++ b/packages/main/src/plugin/autostart-engine.spec.ts
@@ -25,6 +25,7 @@ import { AutostartEngine } from './autostart-engine.js';
 import type { Configuration } from '@podman-desktop/api';
 import { CONFIGURATION_DEFAULT_SCOPE, CONFIGURATION_ONBOARDING_SCOPE } from './configuration-registry-constants.js';
 import type { ApiSenderType } from '/@/plugin/api.js';
+import type { NotificationRegistry } from './notification-registry.js';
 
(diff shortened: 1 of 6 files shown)


cdrage (approved):
Works and I was able to start PD fine,

Should we warn somewhere within PD we have a corrupted configuration file so we can save / read correctly from it? 

When saving a new configuration, won't it override the "corrupt" one?

For example:

1. Start PD with corrupt
2. PD sets it to `configData = {};`
3. PD "clears" the corrupted one rather than warn.

benoitf:
I think that if you do changes it'll override the file with a valid content

jeffmaury (approved):
LGTM. The only concer I have is that the old content is lost, maybe we should backup it and inform user but could be a follow up PR

benoitf:
@jeffmaury 

I've added changes to the PR, it will display a notification + do a backup of the corrupted file

benoitf:
@jeffmaury @cdrage you approved but I added stuff so probably good to review it again

jeffmaury (approved):
LGTM

cdrage:
Tested and viewed the new changes. LGTM!

PR #5191, opened 2023-12-08
extension installation checks architecture and os

### LINKED ISSUE #5190: extension installation checks architecture but no os

When an extension contains a binary it may be packaged using platform ie os and architecture.
The installation is only done using architecture and not os

See https://github.com/containers/podman-desktop/blob/ae2c861f014512522602e6739d45a4426ea7abde/packages/main/src/plugin/image-registry.ts#L630

The os should be taken into account

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

This PR makes the extension installation check for the OS in addition to the architecture of the platform, so binaries in the image can be used in the platform.

fixes #5190 

### How to test this PR?

- Test with existing extensions that the extension is correctly installed
- Test with the extension image: quay.io/phmartin/openshift-image-checker:v0.0.32

### DIFF
--- a/packages/main/src/plugin/image-registry.spec.ts
+++ b/packages/main/src/plugin/image-registry.spec.ts
@@ -496,3 +496,92 @@ describe('expect checkCredentials', async () => {
     expect(registries2.length).toBe(1);
   });
 });
+
+test('getBestManifest returns the expected manifest', () => {
+  const manifests = {
+    'linux-amd64': {
+      platform: {
+        architecture: 'amd64',
+        os: 'linux',
+      },
+      name: 'linux-amd64',
+    },
+    'linux-arm64': {
+      platform: {
+        architecture: 'arm64',
+        os: 'linux',
+      },
+      name: 'linux-arm64',
+    },
+    'windows-amd64': {
+      platform: {
+        architecture: 'amd64',
+        os: 'windows',
+      },
+      name: 'windows-amd64',
+    },
+    'windows-arm64': {
+      platform: {
+        architecture: 'arm64',
+        os: 'windows',
+      },
+      name: 'windows-arm64',
+    },
+    'darwin-amd64': {
+      platform: {
+        architecture: 'amd64',
+        os: 'darwin',
+      },
+      name: 'darwin-amd64',
+    },
+  };
+
+  // Exact matches
+  expect(
+    imageRegistry.getBestManifest(
+      [manifests['linux-amd64'], manifests['linux-arm64'], manifests['windows-amd64'], manifests['windows-arm64']],
+      'amd64',
+      'linux',
+    ),
+  ).toHaveProperty('name', 'linux-amd64');
+  expect(
+    imageRegistry.getBestManifest(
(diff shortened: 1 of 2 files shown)


jeffmaury (commented):
We probably need unit tests

benoitf (changes requested):
It requires tests

Also I'm not sure we need to match OS

I explain, if there is we could use it else it should fetch the Linux version as I assume most of the extensions are built using linux/amd64 or linux/arm64 depending on the system the user has

feloy:
> It requires tests
> 
> Also I'm not sure we need to match OS
> 
> I explain, if there is we could use it else it should fetch the Linux version as I assume most of the extensions are built using linux/amd64 or linux/arm64 depending on the system the user has

ok, so we first find the version with the matching os and arch. If there is no version found, we find the one with 'linux' and arch. Right?

Do we continue with arch, if nothing is found with 'linux' and current arch, we get version for linux/amd64?

benoitf:
yes I think the pattern is 'exact platform/arch' then exact arch then default image but I think last two are already handle ( as basically it's multi arch image vs direct image)

jeffmaury:
> It requires tests
> 
> Also I'm not sure we need to match OS
> 
> I explain, if there is we could use it else it should fetch the Linux version as I assume most of the extensions are built using linux/amd64 or linux/arm64 depending on the system the user has

Our usecase is when extension embeds a binary so we would like selection based on bin

> It requires tests
> 
> Also I'm not sure we need to match OS
> 
> I explain, if there is we could use it else it should fetch the Linux version as I assume most of the extensions are built using linux/amd64 or linux/arm64 depending on the system the user has

Our use case is when extension embeds a binary so we need os selection

packages/main/src/plugin/image-registry.spec.ts
```diff
@@ -496,3 +496,92 @@ describe('expect checkCredentials', async () => {
...
+      'amd64',
+      'darwin',
+    ),
+  ).toHaveProperty('name', 'linux-amd64');
+  expect(
+    imageRegistry.getBestManifest([manifests['windows-amd64'], manifests['linux-amd64']], 'amd64', 'darwin'),
+  ).toHaveProperty('name', 'linux-amd64');
+
+  // Only one os by default
+  expect(
```
benoitf, line 568: I think here it's the blurry lines

Most of people will do Linux images so I would think it's ok to only have partial match when OCI image has Linux as platform. But for windows and Darwin I would think that it doesn't make sense to have this kind of partial matching

Probably that if you do a Darwin image then probably it means that it should not work on windows or Linux

If you do a linux/arm64 or linux/amd64 then yes here probably you don't mean the exact os but mainly the arch

WDYT ?

feloy: Yes, I completely agree. I'll make the change

benoitf: 👍

benoitf (approved):

jeffmaury (approved):
LGTM

Issue #5157, opened 2023-12-06
Details page for Deployments


deboer-tim:
### Is your feature request related to a problem? Please describe

#4366 will deliver a Deployments list, but like other lists we should have a details page.

### Describe the solution you'd like

Provide a basic details page for Deployments with Summary and Kube (YAML) tabs.

PR #5195, opened 2023-12-08
feat: reading deployments, ingress, routes, and services

deboer-tim:
### What does this PR do?

When we do DetailPages for the four Kubernetes objects that we're building nav pages for, we'll need the ability to read each of the objects the same way we do for pods today.

These are fairly boilerplate, so I'm just doing them all at once. Extracted createTestClient() since we're calling that a lot.

### Screenshot / video of UI

N/A

Part of #5157/#5158/#5159.

### How to test this PR?

`yarn test:main`

### DIFF
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -741,11 +741,7 @@ export class KubernetesClient {
     const k8sApi = this.kubeConfig.makeApiClient(CoreV1Api);
     try {
       const res = await k8sApi.readNamespacedPod(name, namespace);
-      if (res?.body) {
-        return res.body;
-      } else {
-        return undefined;
-      }
+      return res?.body;
     } catch (error) {
       telemetryOptions = { error: error };
       throw this.wrapK8sClientError(error);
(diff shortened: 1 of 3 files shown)


deboer-tim:
Rebased and force-pushed to kick SonarCloud.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -747,6 +747,83 @@ export class KubernetesClient {
...
+  async readNamespacedDeployment(name: string, namespace: string): Promise<V1Deployment | undefined> {
+    let telemetryOptions = {};
+    const k8sAppsApi = this.kubeConfig.makeApiClient(AppsV1Api);
+    try {
+      const res = await k8sAppsApi.readNamespacedDeployment(name, namespace);
+      if (res?.body) {
+        return res.body;
+      } else {
+        return undefined;
+      }
```
```suggestion
return res.body;
```

here and everywhere else

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -747,6 +747,83 @@ export class KubernetesClient {
...
+  async readNamespacedIngress(name: string, namespace: string): Promise<V1Ingress | undefined> {
+    let telemetryOptions = {};
+    const k8sNetworkingApi = this.kubeConfig.makeApiClient(NetworkingV1Api);
+    try {
+      const res = await k8sNetworkingApi.readNamespacedIngress(name, namespace);
+      if (res?.body) {
+        return res.body;
+      } else {
+        return undefined;
+      }
```
```suggestion
return res.body;
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -747,6 +747,83 @@ export class KubernetesClient {
...
+        'v1',
+        namespace,
+        'routes',
+        name,
+      );
+      if (res?.body) {
+        return res.body as V1Route;
+      } else {
+        return undefined;
+      }
```
```suggestion
    return res.body;
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -747,6 +747,83 @@ export class KubernetesClient {
...
+  async readNamespacedService(name: string, namespace: string): Promise<V1Service | undefined> {
+    let telemetryOptions = {};
+    const k8sApi = this.kubeConfig.makeApiClient(CoreV1Api);
+    try {
+      const res = await k8sApi.readNamespacedService(name, namespace);
+      if (res?.body) {
+        return res.body;
+      } else {
+        return undefined;
+      }
```
```suggestion
return res.body;
```

dgolovin (approved):
LGTM

nit: you could just return `res.body` rather than do full if/else block.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Had to rebase, so I fixed this at the same time, along with the place I copy/pasted from.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Had to rebase, so I fixed this at the same time, along with the place I copy/pasted from.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Had to rebase, so I fixed this at the same time, along with the place I copy/pasted from.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Had to rebase, so I fixed this at the same time, along with the place I copy/pasted from.

lstocchi (approved):
LGTM

PR #5219, opened 2023-12-11
fix: long socket path

### LINKED ISSUE #5218: Socket paths overflowing UI
### Bug description

PR #5098 added socket paths to the UI, but mine aren't short. With no break they are very ugly in the UI and both overlapping and going outside of the natural space for providers.

Screenshot 2023-12-11 at 11 17 00 AM

We need to either revert #5098 or take a step to crop/make these fit within the UI.

### Operating system

macOS 14.1

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Fixes the long socket path by truncating, and then adding a tooltip.

Removed unnecessary parent div and changed span to div while I was editing, as that's what we'd normally do here.

### Screenshot / video of UI

Before:
Screenshot 2023-12-11 at 11 17 00 AM

After:

Screenshot 2023-12-11 at 11 54 34 AM

Fixes #5218.

### How to test this PR?

Go to Settings > Resources on any platform with long socket paths.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
@@ -473,16 +473,15 @@ function hasAnyConfiguration(provider: ProviderInfo) {
                   <ConnectionErrorInfoButton status="{status}" />
                 {/if}
               </div>
-              <div class="mt-2">
-                <div class="text-gray-700 text-xs" aria-label="{container.name} type">
-                  {#if container.type === 'docker'}Docker{:else if container.type === 'podman'}Podman{/if} endpoint
-                </div>
-                <div class="mt-1">
-                  <span
-                    class="my-auto text-xs"
-                    class:text-gray-900="{container.status !== 'started'}"
-                    aria-label="{container.name} endpoint">{container.endpoint.socketPath}</span>
-                </div>
+              <div class="mt-2 text-gray-700 text-xs" aria-label="{container.name} type">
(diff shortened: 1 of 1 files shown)


benoitf (approved):

dgolovin (approved):
LGTM on macOS
image

afbjorklund:
It would be better if it also showed the ".../podman.sock" part?

Another trick would be to change the $HOME into `~`, for display.

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -473,16 +473,15 @@ function hasAnyConfiguration(provider: ProviderInfo) {
...
-                <div class="mt-1">
-                  <span
-                    class="my-auto text-xs"
-                    class:text-gray-900="{container.status !== 'started'}"
-                    aria-label="{container.name} endpoint">{container.endpoint.socketPath}</span>
-                </div>
+              <div class="mt-2 text-gray-700 text-xs" aria-label="{container.name} type">
+                {#if container.type === 'docker'}Docker{:else if container.type === 'podman'}Podman{/if} endpoint
+              </div>
+              <div
```
afbjorklund, line 479: One alternative is to remove the socketPath from the summary, and only show the type.

benoitf: could be a follow-up PR/improvement

PR #5222, opened 2023-12-11
fix: allow table columns to specify overflow

### LINKED ISSUE #5220: Popup menus not appearing in Images and Volumes lists
### Bug description

The new table component sets overflow-hidden and max-width on each cell to avoid long content (e.g. >50 character volume or image names) from overlapping other columns (or as in the past, tables that are ridiculously wide). However, this means that popup menus or tooltips can't appear outside of the cell either.

### Operating system

macOS 14.1

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

With the previous table (<1.6), column widths were variable and this caused problems:
- sometimes a long image name would cause the tables to require horizontal scrolling
- sometimes a change in one column would cause other columns to shift E.g. starting a container can cause the start/stop button to jump right, and stopping the container the button jumps left, so you might click the wrong button (e.g. delete).

With the new grid-based table component, column widths are fixed so content doesn't affect table width. This solves the previous issues, but it means that simple columns with long content are more likely to have overflow. This was 'fixed' in #4841 by setting a maximum column width and overflow-hidden to limit rendering to within the cell, but now things like popup menus can't go outside of the cells.

I've struggled deciding what the correct default behaviour should be, but I still think #4841 is correct and means the average renderer can stay simple. This change adds a new 'overflow' property to allow renderers that are more interesting to control their own behaviour and render 'outside' of their cell.

Property added to both image and volume action columns.

### Screenshot / video of UI

Self-explanatory; popup menus don't appear.

Fixes #5220.

### How to test this PR?

Test popup menus in Images list.

### DIFF
--- a/packages/renderer/src/lib/image/ImagesList.svelte
+++ b/packages/renderer/src/lib/image/ImagesList.svelte
@@ -213,7 +213,7 @@ const columns: Column<ImageInfoUI>[] = [
   envColumn,
   ageColumn,
   sizeColumn,
-  new Column<ImageInfoUI>('Actions', { align: 'right', width: '150px', renderer: ImageColumnActions }),
+  new Column<ImageInfoUI>('Actions', { align: 'right', width: '150px', renderer: ImageColumnActions, overflow: true }),
 ];
 
 const row = new Row<ImageInfoUI>({ selectable: image => !image.inUse, disabledText: 'Image is used by a container' });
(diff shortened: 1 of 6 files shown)


lstocchi (approved):
LGTM

jeffmaury (approved):
LGTM

benoitf:
it is missing a test to check that overflow is added (or not)

odockal:
Works for me locally on Linux.

deboer-tim:
> it is missing a test to check that overflow is added (or not)

I spent a few minutes yesterday trying to think of a way to test something visually without thinking of this more obvious answer. :(. Test for overflow-hidden class added.

benoitf (approved):

PR #5231, opened 2023-12-12
chore: allow to use index manifest without recommended mediaType field

### LINKED ISSUE #33: Unable to install demo extension
I'm not 100% sure, but I thought I had installed the demo extension from quay since #27 was merged. Quay was also updated this weekend so something might have changed.

Either way, I now receive this error trying to install the extension:
`Error while analyzing image: Error: No config manifest found for [quay.io/podman-desktop-demo/podman-desktop-extension](http://quay.io/podman-desktop-demo/podman-desktop-extension)`

benoitf:
### What does this PR do?
mediaType SHOULD be included but it's not mandatory (but it's not a REQUIRED one)
https://github.com/opencontainers/image-spec/blob/main/image-index.md

so we can't rely on the presence of this field as it may not be there

but add check on schemaVersion REQUIRED field as well

fixes https://github.com/redhat-developer/podman-desktop-demo/issues/33

### DIFF
--- a/packages/main/src/plugin/image-registry.spec.ts
+++ b/packages/main/src/plugin/image-registry.spec.ts
@@ -585,3 +585,78 @@ test('getBestManifest returns the expected manifest', () => {
     imageRegistry.getBestManifest([manifests['windows-amd64'], manifests['darwin-amd64']], 'amd64', 'linux'),
   ).toBeUndefined();
 });
+
+test('getManifestFromUrl returns the expected manifest with mediaType', async () => {
+  const fakeManifest = {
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM. Was able to install the extension and pull the demo images on Win11 Pro

lstocchi (approved):
Works fine with that extension, i wonder if we should check something in the `parsedManifest.manifests` instead of just look if it's an array. Maybe a future enhancement

deboer-tim (approved):
LGTM

Issue #4763, opened 2023-11-10
push image to cluster should not use kebab menu icon


benoitf:
### Bug description

![image](https://github.com/containers/podman-desktop/assets/436777/33d69c46-b5da-4833-bf35-b7d13204d8b9)

this icon is being used for pushing an image but it feels that I could get a drop down menu as it's the icon for a kebab menu so it is confusing

https://github.com/containers/podman-desktop/assets/436777/d2ae73fc-2ac5-4fa1-ac73-20ac82526c05

### Operating system

macOS

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

use kind and OpenShift local extension

salaheddine-zemmouri:
Hello , I am interested in contributing to this project and i would love to fix this issue. Is it still open ?

jeffmaury:
> Hello , I am interested in contributing to this project and i would love to fix this issue. Is it still open ?

It was scheduled to be closed based on priorities but feel free to contribute, you're more than welcomed

myfear:
This seems to be fixed now @jeffmaury .

deboer-tim:
Yes, this was fixed: #5236 (and extensions can customize the icon now too).

PR #5236, opened 2023-12-12
fix: improve default contribution action icon

deboer-tim:
### What does this PR do?

We've had several people get confused with actions contributed by extensions that show up as an ellipsis icon in popup menus and Detail pages. The correct/full fix will require us to allow commands to provide their own icons, and to do this for all of the current extensions. This is not that fix. :)

The other two things we should do are:
- switch the default icon so that it doesn't use the same icon as our popup menus
- pass the detailed property through the ContributionActions component so that these icons show up on buttons on the Details pages like the other actions.

This is that part of the fix!

I went through FA trying to find a more suitable icon and this is the best I could come up with. I preferred it to icons like the puzzle piece and I think it still gives some indication that the action comes from an extension.

### Screenshot / video of UI

Before:

Screenshot 2023-12-12 at 2 25 34 PM
Screenshot 2023-12-12 at 2 25 13 PM

After:

Screenshot 2023-12-12 at 2 26 37 PM
Screenshot 2023-12-12 at 2 01 15 PM

First step toward fixing #4763.

### How to test this PR?

Go to contribution actions, e.g. Image Details page. Use PR #5222 as well to see the popup on Images list.

### DIFF
--- a/packages/renderer/src/lib/actions/ContributionActions.spec.ts
+++ b/packages/renderer/src/lib/actions/ContributionActions.spec.ts
@@ -43,6 +43,41 @@ test('Expect one ListItemButtonIcon', async () => {
   expect(item).toBeInTheDocument();
 });
 
+test('Expect one ListItemButtonIcon without detail', async () => {
+  render(ContributionActions, {
+    args: [],
+    contributions: [
+      {
+        command: 'dummy.command',
+        title: 'dummy-title',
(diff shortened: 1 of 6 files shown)


benoitf:
rebasing to trigger again sonar check

benoitf (approved):
image
works for me

PR #5240, opened 2023-12-13
fix: handle event when loading images from archives

### LINKED ISSUE #5221: podman load does not refresh images interface
### Bug description

When using podman load < image.tar or podman load -i image.tar the images view on podman desktop does not update.

### Operating system

Windows 10

### Installation Method

### Version

1.6.0

### Steps to reproduce

Save an image to a tar file: podman save -o test.tar
Remove the image: podman image rm test
Observe that the image is gone from Podman Desktop
Load the image: podman load < test.tar

UI is not updated to reflect image

### Relevant log output

_No response_

### Additional context

'podman pull' and 'podman image rm' both update the image view properly 

Tried on multiple machines. 
All Windows 10 with latest updates. 
Combination of internet connected and not
Bug is present in 1.5.3 and 1.6.0.
I did not try any versions in between

benoitf:
### What does this PR do?
Refresh the UI when we load images from archive

fixes https://github.com/containers/podman-desktop/issues/5221

### How to test this PR?

unit test or follow test case of the issue

### DIFF
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -2351,3 +2351,37 @@ test('setupConnectionAPI with errors after machine being removed', async () => {
     'Aborting reconnect due to error as connection has been removed (probably machine has been removed)',
   );
 });
+
+test('check handleEvents with loadArchive', async () => {
+  const getEventsMock = vi.fn();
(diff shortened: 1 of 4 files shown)


jeffmaury (approved):
LGTM. Was able to load an image from CLI and see it in image list

Issue #4229, opened 2023-10-06
If creating new resource fails when form has vertical scrolling the error is not visible


dgolovin:
### Bug description

The form below just failed to create podman VM and I am wondering what happened.

![image](https://github.com/containers/podman-desktop/assets/620330/9ef75225-42b4-4a7d-94bc-3b2d18c3fc51)

I have to scroll the form up to see the error.

![image](https://github.com/containers/podman-desktop/assets/620330/5f0b249d-fe34-4bf9-94f5-88df2023bac0)

The error message `Error: [object Object]` is reported here #4230.

### Operating system

macOS Sonoma 14.0

### Installation Method

Other

### Version

next (development version)

deboer-tim:
Progress is shown up there too, so for a second yesterday I thought it hadn't done anything when I clicked Create.

A simple solution would be to auto-scroll to the top when we click Create, but I don't love it when forms move themselves. Alternately we could move progress/error/logs to the bottom, have some kind of floating control, or clear the properties.

FYI to @mairin, I've added this to the next UX discussion.

deboer-tim:
Still a valid problem.

PR #5253, opened 2023-12-13
feat: form progress

deboer-tim:
### What does this PR do?

I'm pretty sure this was an idea from Mo. Adds a linear progress bar at the bottom of the FormPage header to indicate that something is 'in progress', and uses this when building and pulling images, playing kube, deploying pods, creating pods from containers, creating volumes, and creating resources (container or Kubernetes).

This helps to indicate that there's 'something happening', especially in cases like #4229 where you can't see another progress monitor in the view.

This 'duplicates' the progress monitor when creating a container/Kubernetes instance, but I don't want to remove that as these pages are also used during onboarding and the original/smaller one helps draw your eye to the Show Logs expander. IMHO having two here is still an improvement.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/19958075/7c9a5619-1137-45a7-a6d1-a3825feb7f7d

N/A

### How to test this PR?

Perform any of these actions.

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -146,7 +146,7 @@ async function abortBuild() {
 }
 </script>
 
-<FormPage title="Build Image from Containerfile">
+<FormPage title="Build Image from Containerfile" inProgress="{buildImageInfo?.buildRunning}">
   <svelte:fragment slot="icon">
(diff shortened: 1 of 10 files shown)


benoitf (commented):
LGTM but there is no test of the LinearProgress in FormPage

benoitf (approved):
👍

deboer-tim:
Lying in bed last night I was trying to remember if I actually included the test I wrote in the PR. 😂 There now.

PR #5261, opened 2023-12-14
chore: fix GitHub template for versions

benoitf:
### What does this PR do?
1.5.3 was the published release
1.6.0 is not published, 1.6.2 will probably

### Screenshot / video of UI

N/A

Drop down when reporting issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -49,9 +49,9 @@ body:
       label: Version
       description: What version of the software are you running?
       options:
-        - "1.6.0"
+        - "1.6.2"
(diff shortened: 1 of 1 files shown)


lstocchi (approved):

deboer-tim (approved):
Maybe just do 1.5.x/1.6.x/... or 1.5/1.6/...? We've only released one of each, might be simpler, and if we had to put out another fix I'm assuming we wouldn't want to update this.

benoitf:
we usually put the version numbers of releases that went out (not the pre-releases)

benoitf:
and it is important when people report to say for example it's 1.6.2 or 1.6.3 (if we do a bugfix release after it's out for all people)

PR #5267, opened 2023-12-14
fix: don't refresh image list when age updates

### LINKED ISSUE #5263: Whole Image table is refreshed (reset) every 2 seconds if there is a fresh image (with age counted in seconds)
### Bug description

As reported on the slack, the effect of the Age being refreshed every two seconds is that scrolling is also reset and move back to the top. Also if the context menu of the image is opened, it is closed after every event change.

Happens when the image is built and the age is counted in seconds. Similar issue will happen with minutes, but the number of occurrences will get lower and lower eventually. 

https://github.com/containers/podman-desktop/assets/19164299/93a7e587-9113-4e8d-92a6-b79a03f5552b

### Operating system

Windows, Linux, Mac

### Installation Method

### Version

1.6.0

### Steps to reproduce

1. Open Images table, ideally have many images available (not necessary)
2. Build an image
3. Assert: Image's age is in seconds and changes every 2 seconds
4. Scroll down
5. Assert: Scroll is not being changed
Actual result: Position of the scroll is moved back to the top
6. Alternatively, open context menu on the image (ie. kebab menu with more options/actions)
7. Assert: Context menu is opened
Actual: It closes after age is changed. (on any image row)

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

We have been replacing the entire image list whenever the age is updated. This worked (?) for the <table> implementation, but for the new table component it is triggering a full refresh/reset of the table. Just updating the image age properties and then setting images to itself is enough to trigger Svelte to update the UI without a full replacement.

### Screenshot / video of UI

See #5263, this will still update age without the forced reset each 2s.

Fixes #5263.

### How to test this PR?

Build an image and make sure the age increases on the list without causing everything to scroll.

### DIFF
--- a/packages/renderer/src/lib/image/ImagesList.svelte
+++ b/packages/renderer/src/lib/image/ImagesList.svelte
@@ -133,9 +133,10 @@ async function deleteSelectedImages() {
 let refreshTimeouts: NodeJS.Timeout[] = [];
 const SECOND = 1000;
 function refreshAge() {
-  images = images.map(imageInfo => {
-    return { ...imageInfo, age: imageUtils.refreshAge(imageInfo) };
+  images.forEach(imageInfo => {
+    imageInfo.age = imageUtils.refreshAge(imageInfo);
   });
+  images = images;
 
   // compute new interval
   const newInterval = computeInterval();
--- a/packages/renderer/src/lib/image/ImagesList.svelte
+++ b/packages/renderer/src/lib/image/ImagesList.svelte
@@ -133,9 +133,9 @@ async function deleteSelectedImages() {
 let refreshTimeouts: NodeJS.Timeout[] = [];
 const SECOND = 1000;
 function refreshAge() {
-  images.forEach(imageInfo => {
+  for (const imageInfo of images) {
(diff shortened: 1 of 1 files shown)


lstocchi (approved):
LGTM

packages/renderer/src/lib/image/ImagesList.svelte
```diff
@@ -133,9 +133,10 @@ async function deleteSelectedImages() {
 let refreshTimeouts: NodeJS.Timeout[] = [];
 const SECOND = 1000;
 function refreshAge() {
-  images = images.map(imageInfo => {
-    return { ...imageInfo, age: imageUtils.refreshAge(imageInfo) };
+  images.forEach(imageInfo => {
```
benoitf: might be better to use for of rather than forEach in case at some point we need async stuff, etc

benoitf: (and it's faster)

deboer-tim: Switched to for loop.

benoitf (approved):

PR #5287, opened 2023-12-15
chore: Update electron to v28.1.4

### LINKED ISSUE #5285: Electron modules may not have ids now
### Bug description

in Electron 28 and support of ESM some modules may not have id/children properties

we should handle that else startup is failing with Electron 28

### Operating system

macOS 14

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Includes a fixes for handling new module format in Electron

fixes #5285 
https://github.com/containers/podman-desktop/pull/5124

### How to test this PR?

PR check should be ✅

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -111,6 +111,10 @@ export interface ActivatedExtension {
 
 const EXTENSION_OPTION = '--extension-folder';
 
+export interface RequireCacheDict {
+  [key: string]: NodeModule | undefined;
+}
+
 export class ExtensionLoader {
   private overrideRequireDone = false;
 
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/extension-loader.ts
```diff
@@ -1092,18 +1096,23 @@ export class ExtensionLoader {
...
+  protected getRequireCache(): RequireCacheDict {
+    return require.cache;
+  }
+
   loadRuntime(extension: AnalyzedExtension): NodeRequire | undefined {
     // cleaning the cache for all files of that plug-in.
-    Object.keys(require.cache).forEach(function (key): void {
-      const mod: NodeJS.Module | undefined = require.cache[key];
+    Object.keys(this.getRequireCache()).forEach(key => {
+      const mod: NodeJS.Module | undefined = this.getRequireCache()[key];
```
```suggestion
      const mod: NodeModule | undefined = this.requireCache[key];
```

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1092,18 +1096,23 @@ export class ExtensionLoader {
     return require(path);
   }
 
+  // helper function to get require cache
+  protected getRequireCache(): RequireCacheDict {
+    return require.cache;
+  }
```
```suggestion
  // helper function to get require cache
  protected get requireCache(): RequireCacheDict {
    return require.cache;
  }
```

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1092,18 +1096,23 @@ export class ExtensionLoader {
...
+
   loadRuntime(extension: AnalyzedExtension): NodeRequire | undefined {
     // cleaning the cache for all files of that plug-in.
-    Object.keys(require.cache).forEach(function (key): void {
-      const mod: NodeJS.Module | undefined = require.cache[key];
+    Object.keys(this.getRequireCache()).forEach(key => {
```
```suggestion
    Object.keys(this.requireCache).forEach(key => {
```

packages/main/src/plugin/extension-loader.ts
dgolovin: Would be easier to read.

dgolovin (commented):
Some nitpicking

packages/main/src/plugin/extension-loader.ts
benoitf: fixed

benoitf:
new playwright fixed the issue with electron v28 so now e2e tests can run

deboer-tim (approved):
Code LGTM, tested for several minutes and couldn't see any difference/no problems.

PR #5289, opened 2023-12-15
Display files into Image layers

feloy:

### What does this PR do?

This Pr started as a PoC realized during devtools week 23Q4. 

A new "Files" tab is added to the Image details. This new page lists the layers of the image, and for each layer, the files contained in this layer, in a way similar to  `dive` does.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/9973512/9ba11327-ddc6-4ea7-9ad7-74f5ece6437f

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -70,6 +70,7 @@ import datejs from 'date.js';
 import { isWindows } from '../util.js';
 import { EnvfileParser } from './env-file-parser.js';
 import type { ProviderRegistry } from '/@/plugin/provider-registry.js';
+import type { ImageLayer } from './image-layers.js';
 
 export interface InternalContainerProvider {
   name: string;
@@ -2070,27 +2071,44 @@ export class ContainerProviderRegistry {
     }
   }
 
-  async getImageLayers(engineId: string, id: string): Promise<Map<string, string[]>> {
+  async getImageLayers(engineId: string, id: string): Promise<ImageLayer[]> {
     const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), 'podman-desktop'));
     const tarFile = path.join(tmpdir, id + '.tar');
     await this.saveImage(engineId, id, tarFile);
     await nodeTar.extract({ file: tarFile, cwd: tmpdir });
+
     const manifest = JSON.parse(fs.readFileSync(path.join(tmpdir, 'manifest.json'), 'utf-8'));
     const layers = manifest[0].Layers;
-    const layersMap = new Map<string, string[]>();
+
+    const configFile = manifest[0].Config;
+    const config = JSON.parse(fs.readFileSync(path.join(tmpdir, configFile), 'utf-8'));
+    const history = config.history;
+    const layersResult: ImageLayer[] = [];
     for (const layer of layers) {
-      const entries: string[] = [];
+      const entries: nodeTar.ReadEntry[] = [];
       const layerTar = path.join(tmpdir, layer);
       await nodeTar.list({
         file: layerTar,
         onentry: (entry: nodeTar.ReadEntry) => {
-          entries.push(entry.path);
+          entries.push(entry);
         },
       });
-      layersMap.set(layer, entries);
+      let layerHistory: string | undefined;
+      for (;;) {
+        const hist = history.shift();
+        if (!hist) {
+          break;
+        }
+        if (hist.empty_layer) {
+          continue;
+        }
+        layerHistory = hist.created_by.replace(/^\/bin\/sh -c /, '');
+        break;
+      }
+      layersResult.push({ id: layer, files: entries, history: layerHistory });
     }
     fs.rmSync(tmpdir, { force: true, recursive: true });
-    return layersMap;
+    return layersResult;
   }
 
   async getPodInspect(engineId: string, id: string): Promise<PodInspectInfo> {
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -70,7 +70,7 @@ import datejs from 'date.js';
 import { isWindows } from '../util.js';
 import { EnvfileParser } from './env-file-parser.js';
 import type { ProviderRegistry } from '/@/plugin/provider-registry.js';
-import type { ImageLayer } from './image-layers.js';
+import { getLayersFromImageArchive, type ImageLayer } from './image-layers.js';
 
 export interface InternalContainerProvider {
   name: string;
@@ -2076,39 +2076,9 @@ export class ContainerProviderRegistry {
     const tarFile = path.join(tmpdir, id + '.tar');
     await this.saveImage(engineId, id, tarFile);
     await nodeTar.extract({ file: tarFile, cwd: tmpdir });
-
-    const manifest = JSON.parse(fs.readFileSync(path.join(tmpdir, 'manifest.json'), 'utf-8'));
-    const layers = manifest[0].Layers;
-
-    const configFile = manifest[0].Config;
-    const config = JSON.parse(fs.readFileSync(path.join(tmpdir, configFile), 'utf-8'));
-    const history = config.history;
-    const layersResult: ImageLayer[] = [];
-    for (const layer of layers) {
-      const entries: nodeTar.ReadEntry[] = [];
-      const layerTar = path.join(tmpdir, layer);
-      await nodeTar.list({
-        file: layerTar,
-        onentry: (entry: nodeTar.ReadEntry) => {
-          entries.push(entry);
-        },
-      });
-      let layerHistory: string | undefined;
-      for (;;) {
-        const hist = history.shift();
-        if (!hist) {
-          break;
-        }
-        if (hist.empty_layer) {
-          continue;
-        }
-        layerHistory = hist.created_by.replace(/^\/bin\/sh -c /, '');
-        break;
-      }
-      layersResult.push({ id: layer, files: entries, history: layerHistory });
-    }
+    const result = await getLayersFromImageArchive(tmpdir);
     fs.rmSync(tmpdir, { force: true, recursive: true });
-    return layersResult;
+    return result;
   }
 
   async getPodInspect(engineId: string, id: string): Promise<PodInspectInfo> {
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -60,6 +60,9 @@ import type { NetworkInspectInfo } from './api/network-info.js';
 import type { Event } from './events/emitter.js';
 import { Emitter } from './events/emitter.js';
 import fs from 'node:fs';
+import os from 'node:os';
+import nodeTar from 'tar';
+import path from 'node:path';
 import { pipeline } from 'node:stream/promises';
 import type { ApiSenderType } from './api.js';
 import { type Stream, Writable } from 'node:stream';
@@ -2065,6 +2068,29 @@ export class ContainerProviderRegistry {
     } finally {
       this.telemetryService.track('imageSave', telemetryOptions);
     }
+  }
+
+  async getImageLayers(engineId: string, id: string): Promise<Map<string, string[]>> {
+    const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), 'podman-desktop'));
+    const tarFile = path.join(tmpdir, id + '.tar');
+    await this.saveImage(engineId, id, tarFile);
+    await nodeTar.extract({ file: tarFile, cwd: tmpdir });
+    const manifest = JSON.parse(fs.readFileSync(path.join(tmpdir, 'manifest.json'), 'utf-8'));
+    const layers = manifest[0].Layers;
+    const layersMap = new Map<string, string[]>();
+    for (const layer of layers) {
+      const entries: string[] = [];
+      const layerTar = path.join(tmpdir, layer);
+      await nodeTar.list({
+        file: layerTar,
+        onentry: (entry: nodeTar.ReadEntry) => {
+          entries.push(entry.path);
+        },
+      });
+      layersMap.set(layer, entries);
+    }
+    fs.rmSync(tmpdir, { force: true, recursive: true });
+    return layersMap;
   }
 
   async getPodInspect(engineId: string, id: string): Promise<PodInspectInfo> {
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -2073,12 +2073,14 @@ export class ContainerProviderRegistry {
 
   async getImageLayers(engineId: string, id: string): Promise<ImageLayer[]> {
     const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), 'podman-desktop'));
-    const tarFile = path.join(tmpdir, id + '.tar');
-    await this.saveImage(engineId, id, tarFile);
-    await nodeTar.extract({ file: tarFile, cwd: tmpdir });
-    const result = await getLayersFromImageArchive(tmpdir);
-    fs.rmSync(tmpdir, { force: true, recursive: true });
-    return result;
+    try {
+      const tarFile = path.join(tmpdir, id + '.tar');
+      await this.saveImage(engineId, id, tarFile);
+      await nodeTar.extract({ file: tarFile, cwd: tmpdir });
+      return await getLayersFromImageArchive(tmpdir);
+    } finally {
+      fs.rmSync(tmpdir, { force: true, recursive: true });
+    }
   }
 
   async getPodInspect(engineId: string, id: string): Promise<PodInspectInfo> {
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -60,6 +60,7 @@ import type { NetworkInspectInfo } from './api/network-info.js';
 import type { Event } from './events/emitter.js';
 import { Emitter } from './events/emitter.js';
 import fs from 'node:fs';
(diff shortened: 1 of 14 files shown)


deboer-tim:
This addition is awesome, I love it! However, I think it needs a bit more UI work (and fixed tests) before merging.

Some initial reaction:
- Maybe the tab should be called Layers so that we don't need a title within the tab?
- Can't tell which layer is selected. Scrollbar at bottom isn't great either; maybe we use ... and full name on tooltip?
- Can we all overall layer size or reference to which line in the containerfile caused it? I haven't looked at what the API is but those would be really nice to add.
- Initial layer root should be initially expanded.
- File attributes and size are nice, but then folder/files don't line up nor even files within a folder. Maybe needs a design suggestion here, either columns or a fixed font.
- One of the other tools I looked at used color to indicate when a layer added a new file vs removed or modified. Not accessible by itself, but would be a really nice addition to this.

As long as the core UI issues are addressed I would be fine if some of these were tracked via future enhancements, they were just top of mind as I reviewed.

packages/renderer/src/lib/ui/TreeView.svelte
```diff
@@ -0,0 +1,130 @@
...
+.no-arrow {
+  padding-left: 1rem;
+}
+.arrow {
+  cursor: pointer;
+  display: inline-block;
+}
+.arrowDown {
+  transform: rotate(90deg);
+}
```
benoitf: I think we should remove this style directive

for example instead of adding no-arrow class, just add `pl-4`

instead of adding `arrow`, add `cursor-pointer` and `inline-block` classes

etc

packages/renderer/src/lib/ui/TreeView.svelte
```diff
@@ -0,0 +1,130 @@
...
+  }
+  if (data.type === 'SymbolicLink') {
+    return 'text-sky-300';
+  }
+  if (data.type === 'Directory') {
+    return 'text-sky-500';
+  }
+  if (isExecutable(data.mode)) {
+    return 'text-green-500';
+  }
```
benoitf: it does not look like a generic TreeView component

AFAIK the component should not be tied to 'SymbolicLink' or 'Directory' or '.wh' or some executables

color should be provided by the consumer of the component, not the component
component should take leafs/tree as input

packages/renderer/src/lib/ui/TreeView.svelte
```diff
@@ -0,0 +1,130 @@
...
+  if (size > 1024) {
+    size = Math.floor(size / 100) / 10;
+    u = 'M';
+  }
+
+  if (size > 99) {
+    size = Math.floor(size);
+  }
+  return size + u;
+}
```
benoitf: it can't go into a generic TreeView component, it looks like a specific TreeImageLayer

packages/main/src/plugin/container-registry.ts
```diff
@@ -1822,6 +1826,16 @@ export class ContainerProviderRegistry {
...
   }
 
+  async getImageLayers(engineId: string, id: string): Promise<ImageLayer[]> {
+    const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), 'podman-desktop'));
+    const tarFile = path.join(tmpdir, id + '.tar');
+    await this.saveImage(engineId, id, tarFile);
+    await nodeTar.extract({ file: tarFile, cwd: tmpdir });
+    const result = await getLayersFromImageArchive(tmpdir);
+    fs.rmSync(tmpdir, { force: true, recursive: true });
+    return result;
```
benoitf: probably need a try/finally for deletion so in case of failure we try at least to delete

benoitf: or if we ask two times in 2mn maybe we don't redo the same steps, so maybe it should be scheduled for deletion after some time and use a cache (but can be follow-up)

packages/main/src/plugin/filetree.ts
```diff
@@ -0,0 +1,49 @@
+export class fileNode<T> {
```
benoitf: it is missing copyright headers

benoitf: probably rename it to file-tree.ts

packages/main/src/plugin/filetree.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  addChild(name: string): fileNode<T> {
+    const child = new fileNode<T>(name);
+    this.children.set(name, child);
+    return child;
+  }
+}
+
+export class fileTree<T> {
```
benoitf: I would expect an uppercase on the first letter : fileTree --> FileTree

packages/main/src/plugin/image-layers.ts
```diff
@@ -0,0 +1,48 @@
+import nodeTar from 'tar';
```
benoitf, line 19: it is missing the headers

packages/main/src/plugin/image-layers.ts
```diff
@@ -0,0 +1,48 @@
...
+import { fileTree } from './filetree.js';
+
+export interface ImageLayer {
+  id: string;
+  history?: string;
+  files: nodeTar.ReadEntry[];
+  tree: fileTree<nodeTar.ReadEntry>;
+}
+
+export async function getLayersFromImageArchive(tmpdir: string): Promise<ImageLayer[]> {
```
benoitf, line 71: I think it would be easier with a class for testing (as it's missing unit tests there)

packages/main/src/plugin/image-layers.ts
```diff
@@ -0,0 +1,48 @@
...
+    let layerHistory: string | undefined;
+    for (;;) {
+      const hist = history.shift();
+      if (!hist) {
+        break;
+      }
+      if (hist.empty_layer) {
+        continue;
+      }
+      layerHistory = hist.created_by.replace(/^\/bin\/sh -c /, '');
```
benoitf: is it covering all the cases ?

packages/main/src/plugin/file-tree.spec.ts
```diff
@@ -25,35 +25,42 @@ interface typ {
 
 test('add paths to filetree', () => {
   const tree = new FileTree<typ>('tree1');
-  tree.addPath('A', { path: 'A-path' });
-  tree.addPath('a/', { path: 'a/-path' });
-  tree.addPath('a/b/c/d.txt', { path: 'a/b/c/d.txt-path' });
-  tree.addPath('a/b/c/e.txt', { path: 'a/b/c/e.txt-path' });
+  tree.addPath('A', { path: 'A-path' }, 5);
```
dgolovin: Method chaining can be used here

packages/renderer/src/lib/ui/TreeView.svelte
```diff
@@ -0,0 +1,79 @@
+<script context="module" lang="ts">
+const _expansionState = new Map<string, boolean>();
+</script>
+
+<script lang="ts">
+import type { FileNode } from '../../../../main/src/plugin/file-tree';
+import { ImageUtils } from '../image/image-utils';
+
+export let tree: FileNode<any>;
+export let margin = 0;
```
benoitf: IMHO it can't be in ui folder as it's not a generic component

it includes ImageUtils and file-tree

packages/renderer/src/lib/ui/TreeView.svelte
```diff
@@ -0,0 +1,79 @@
...
+}
+
+$: expanded = _expansionState.get(label) || false;
+const toggleExpansion = () => {
+  expanded = !expanded;
+  _expansionState.set(label, expanded);
+};
+$: arrowDown = expanded;
+
+function getLink(file: any): string {
```
benoitf: why the type is any ? (and how linter is not reporting an issue there)

packages/renderer/src/lib/ui/TreeView.svelte
benoitf: either it needs to be a fully generic `TreeView` or its location is in `renderer/src/lib/image` folder

packages/renderer/src/lib/ui/TreeView.svelte
```diff
@@ -0,0 +1,79 @@
+<script context="module" lang="ts">
+const _expansionState = new Map<string, boolean>();
```
benoitf: why using _ prefix here ?

packages/renderer/src/lib/ui/TreeView.svelte
benoitf: and to avoid confusion, it should be named ImageFileTreeView (or something like that)

packages/main/src/plugin/container-registry.ts
```diff
@@ -1888,6 +1892,18 @@ export class ContainerProviderRegistry {
...
+  async getImageLayers(engineId: string, id: string): Promise<ImageLayer[]> {
+    const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), 'podman-desktop'));
+    try {
+      const tarFile = path.join(tmpdir, id + '.tar');
+      await this.saveImage(engineId, id, tarFile);
+      await nodeTar.extract({ file: tarFile, cwd: tmpdir });
+      return await getLayersFromImageArchive(tmpdir);
+    } finally {
+      fs.rmSync(tmpdir, { force: true, recursive: true });
+    }
```
benoitf: any reason to use the sync version rather than the promises one ?
```
import { mkdtemp } from 'node:fs/promises';
```

benoitf: fsPromises.rm instead of rmSync

packages/main/src/plugin/file-tree.spec.ts
```diff
@@ -0,0 +1,66 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

it will start in 2024 when it's merged

packages/main/src/plugin/file-tree.ts
```diff
@@ -0,0 +1,71 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export class FileNode<T> {
+  name: string;
+  isRemoved: boolean;
+  data: T | undefined;
```
benoitf: it looks like an optional field
```suggestion
  data?: T;
```

packages/main/src/plugin/file-tree.ts
```diff
@@ -0,0 +1,71 @@
...
+  children: Map<string, FileNode<T>>;
+  size: number;
+
+  constructor(name: string) {
+    this.isRemoved = name.startsWith('.wh.');
```
benoitf: should it be a constant '.wh.'

packages/main/src/plugin/image-layers.ts
```diff
@@ -0,0 +1,123 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/main/src/plugin/image-layers.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import nodeTar from 'tar';
+import fs from 'node:fs';
+import path from 'node:path';
+import { FileTree } from './file-tree.js';
+
+export interface Colorable {
```
benoitf: probably need to be FileColorable as it is related to files

it will avoid name collision as it's exported

packages/main/src/plugin/image-layers.ts
```diff
@@ -0,0 +1,123 @@
...
+  isLink: boolean;
+  isFifo: boolean;
+  isBlock: boolean;
+  isChar: boolean;
+  isExec: boolean;
+  isSUID: boolean;
+  isSGID: boolean;
+}
+
+export interface File extends Colorable {
```
benoitf: ImageLayerFile ?

packages/renderer/src/lib/image/ImageDetailsFiles.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+});
+
+beforeEach(() => {
+  vi.resetAllMocks();
+});
+
+vi.mock('../ui/TreeView.svelte');
+
+test('when the image does not contain layers', async () => {
+  mocks.getImageLayersMock.mockResolvedValue([]);
```
benoitf:
```
const getImageLayersMock = vi.fn()
```

remove the hoisted part

then
```
beforeAll(() => {
  (window as any).getImageLayers = getImageLayersMock;
});
```

packages/renderer/src/lib/ui/TreeView.svelte
```diff
@@ -0,0 +1,79 @@
...
+  if (tree.data.isDir) {
+    return 'text-sky-500';
+  }
+  if (tree.data.isExec) {
+    return 'text-green-500';
+  }
+  return '';
+}
+
+$: expanded = _expansionState.get(label) || false;
```
benoitf:
```suggestion
$: expanded = _expansionState.get(label) ?? false;
```

packages/renderer/src/lib/ui/TreeView.svelte
```diff
@@ -0,0 +1,79 @@
+<script context="module" lang="ts">
```
benoitf: why do we have an empty file in __mocks__ folder TreeView.svelte ?

feloy: I wanted to mock TreeView when testing the parent component, but I don't need to mock it anymore, I'm removing it

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -0,0 +1,69 @@
...
+let currentRoot: FileNode<ImageLayerFile> | undefined; // keep this variable declaration first (for the tests to access it at $$.ctx[1] - layers, its dependency, is at 0)
+
+export let image: ImageInfoUI;
+let loading = true;
+let layers: ImageLayer[];
+let currentLayerId: string;
+$: currentRoot = layers?.length ? layers[0].tree.root : undefined;
+
+onMount(async () => {
+  layers = await window.getImageLayers(image.engineId, image.id);
```
axel7083, line 19: I have an image, which take forever on the screen `Loading layers, please wait...`, so I quite the page, come back, and do that 2-3 times, when looking at the code I understand that each time I opened the container tab files I re-start a full extraction. 

I think we should have either inside the preload or the container registry a place where we know which one are still running, as some may take very long time (1GB image took nearly 3 minutes)

In the future we might want also to cache this information

axel7083 (commented):
I love this feature!

image

The scrollbar are a bit weird in my opinion. The entire page should be scrollable in my opinion, not only the layers, and filesystem.

I feel also that the spacing between the information is a bit too big,
image

I get it mimic the result of `ls` but I feel like there are too much space between the permissions and the rest

## Edit

My opinion is that it is an amazing feature, that really would help developers. But there are maybe a lot of work to do yet, so maybe hiding it being an experimental flag ? This PR is already a decent size, and should probably be kept small and re-work in follow-up issues ?

packages/main/src/plugin/image-layers.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+  isSGID?: boolean;
+  isSUID?: boolean;
+  linkTarget?: string;
+  modeString?: string;
+  size?: number;
+  typeChar?: string;
+  uid?: number;
+};
+
+function newFile(val: FileOptions): ImageLayerFile {
```
axel7083: Instead of creating a new type, you could have use `Partial<ImageLayerFile>`

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
feloy: Yes, there are many things I would like to add after this first version is in. For example, I would like to show from the beginning the different layers (we can get them very quickly), then show which layer is being "analyzed" (this taks time), so the user has an idea of how many time it will need to wait

feloy:
> I love this feature!
> 
> image
> The scrollbar are a bit weird in my opinion. The entire page should be scrollable in my opinion, not only the layers, and filesystem.
> 
> I feel also that the spacing between the information is a bit too big, image
> 
> I get it mimic the result of `ls` but I feel like there are too much space between the permissions and the rest
> 

Here, the user and group are 0 and 0, but they could be 1000:1000, which could take more space

feloy:
I'm wondering ft it would be better to have this functionality as an extension.

What is missing today in the API is that an extension can register a new tab in a Details page (here image, but I have other ideas of extensions on container, pod, deployment, etc). The extension would get as entry the information of the object (image, etc), and would have access to a webview, which would be placed in the tab.

@benoitf wdyt? Do you think it would be reasonably possible?

benoitf:
Every screen could be provided by extensions. All the kubernetes stuff could be done with extensions, etc. It's where we are putting the cursor.

I'm fine having it as an extension. But where I'm struggling is that if it needs to be a webview or we could have the generic TreeView component (not tied to directories/files). I mean, if we could have something like for the checkers, it could be better if it could be only backend extensions.

So I would prefer if the extension, if possible, should not provide a webview, but just a TreeView data provider that would be displayed by the core of Podman Desktop

And then extension could say: Register a TreeView provider for details tab

I think it would be beneficial for both Podman Desktop and other extensions if we don't need specific webviews as it's more complicated to maintain. (on the extension side) and will always be less integrated

feloy:
Closing for now as there is no clear indication that users would be interested by the feature

PR #5295, opened 2023-12-18
feat: deployment, route, service informer stores

deboer-tim:
### What does this PR do?

Kubernetes-backed informer stores for deployments, routes, and services
based on Luca's work in #5029:
- Add kubernetes-client.ts plumbing for deployment, route, and service
  informers.
- Switch existing deployment and route stores to the informer pattern.
- Adds a new services store.
- Adds filtering to all stores (including ingress) so that lists can
  filter as with previous stores.

### Screenshot / video of UI

N/A

Core support towards several Kube UI issues.

### How to test this PR?

`yarn test:main`, `yarn test:renderer`


feloy:
I'm a little concerned about how the informers are handled on the renderer side, the informers are started in main and events (add/delete/update) are sent from main to renderer.

I think it would be simpler to completely manage the informers and the events in the main side, compute (from events) and store the current state also in the main side, and only send the current state at the renderer, through a store, so the components get the updates listening at the store.

The `kubernetes-client` seems a good place to manage these informers and these states, specifically from its `refresh` method, which is a method called every time the `kube/config` file changes, which is, IMO, the only condition necessary to change informers from one context to another.

To illustrate, I'm testing such an implementation for displaying the information about the kube contexts: https://github.com/containers/podman-desktop/pull/5329

deboer-tim:
@feloy It's not clear to me how you're going to sync with the renderer in #5329 yet, maybe just needs the next commit?

We already have 3 Kubernetes stores (or 4 if you count pods). Two (deployments, routes) are using the same pattern as most stores + a Kubernetes watch, and one (ingress) is using informer pattern from #5029. This PR is just switching deployments and routes to the same informer pattern, and adding one for services. How would you like to proceed? If we can find a better v2 informer pattern IMHO it would still be easier to apply that to 4 consistent v1 informer stores than closing this, or holding and trying to do everything here.

+1 to using kube/config change to trigger informer change. Also see #5311 - if the cluster is unreachable when they start I don't think the informers are automatically connecting today.

benoitf (approved):

jeffmaury (commented):
I see code duplicated between stores (try to match items in array that can be refactored and shared

feloy:
> @feloy It's not clear to me how you're going to sync with the renderer in #5329 yet, maybe just needs the next commit?
> 
> We already have 3 Kubernetes stores (or 4 if you count pods). Two (deployments, routes) are using the same pattern as most stores + a Kubernetes watch, and one (ingress) is using informer pattern from #5029. This PR is just switching deployments and routes to the same informer pattern, and adding one for services. How would you like to proceed? If we can find a better v2 informer pattern IMHO it would still be easier to apply that to 4 consistent v1 informer stores than closing this, or holding and trying to do everything here.
> 
> +1 to using kube/config change to trigger informer change. Also see #5311 - if the cluster is unreachable when they start I don't think the informers are automatically connecting today.

I have made progress on the PR #5329: All the logic is limited to the kubernetes-client, and the store is just a pass-through, getting data through events. 

It seems it would be easy to extend the logic to start informers for the current context, and stop previously started informers on previous current context.

But I'm ok to move to this pattern later, after you have moved all existing ones to a single one.

deboer-tim:
The Services UI #5363 was based on the services store in this PR / this branch and the 'do not merge before' comment likely wasn't obvious enough - anyway, I've confirmed that all the changes here are now merged so closing this PR.

Post this PR we have working/consistent informer stores for the first batch of Kubernetes objects, and UI is unblocked. Let's continue discussion on #5329 if we can improve the pattern. 👍🏼

benoitf:
yes I thought some PR was expecting a rebase but it wasn't this one (I didn't see quickly the comment)

Issue #4972, opened 2023-11-23
Refactor the refreshing of age in tables


deboer-tim:
### Is your enhancement related to a problem? Please describe

While building the Deployments page I was reminded of how much near-boilerplate code exists in each page to determine when to refresh each objects age.

### Describe the solution you'd like

We should refactor or componentize how we determine when to refresh each row in a table (potentially something in the table component itself) and the timer for it, so that we don't have this copy/paste code inside each page.

For new objects I don't love the 'refresh every 2s' - but when I tried to 'fix' it a few months ago I had to grudgingly agree with the intelligence of not refreshing every second since the numbers almost look blurred and it is visually distracting. :) Maybe this is a chance to think of something different as well, although I wouldn't hold this issue up for it.

deboer-tim:
Fixed for Kubernetes objects, still valid for others.

PR #5306, opened 2023-12-18
feat: duration column

deboer-tim:
### What does this PR do?

Creates a new duration column that takes a date and displays a humanized time, then automatically refreshes exactly when there should be an update.

Since the timeouts and UI are handled by the column itself there's no need for the UI object to store and update the age string, for the table to have timers or do anything related to the update. List components are much simpler, better encapsulation, and less overall code.

### Screenshot / video of UI

N/A

Core fix for #4972, would still need to be used by other tables.

### How to test this PR?

`yarn test:renderer`, or bring up Deployments page and see no difference.


benoitf (approved):

lstocchi (approved):
LGTM

benoitf:

PR #5307, opened 2023-12-18
fix: empty screen reset filter by default

### LINKED ISSUE #5288: Clear filter button is useless in Images page
### Bug description

If you click on the Clear filter button and you have a filter set, it does nothing

clear

### Operating system

Win11Pro

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Fixes the regression caused by #4600. Prior to this the FilteredEmptyScreen would reset all filters to ''. The container list needed special processing to not remove the filters like 'is:running' so it was changed to an event, but none of the other lists respond to this event so none of the filters get cleared.

This sets a default behaviour of clearing the list so that none of the other lists need to respond to the event if they are fine just clearing the filter. If you do your own special processing, you just call e.preventDefault(). The search term is included as an event property because we need to set cancelable in the third argument.

The pods list has the tabs/is:running so I added the event handling there as well. Without this change clearing the filter would always return you to the first tab.

### Screenshot / video of UI

N/A

Fixes #5288.

### How to test this PR?

Enter random text in container, pods, and volumes/images view, and confirm that clearing the filter works (and leaves the is:running/stopped for containers/pods list).

### DIFF
--- a/packages/renderer/src/lib/container/ContainerList.svelte
+++ b/packages/renderer/src/lib/container/ContainerList.svelte
@@ -674,7 +674,10 @@ function setStoppedFilter() {
         <FilteredEmptyScreen
           icon="{ContainerIcon}"
           kind="containers"
-          on:resetFilter="{() => (searchTerm = containerUtils.filterResetSearchTerm(searchTerm))}"
+          on:resetFilter="{e => {
+            searchTerm = containerUtils.filterResetSearchTerm(searchTerm);
+            e.preventDefault();
+          }}"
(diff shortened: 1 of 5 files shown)


benoitf (commented):
bugfix would require a non regression test

slemeur:
👍

deboer-tim:
Rebased to latest and added regression and non-regression tests.

axel7083 (approved):
LGTM

PR #5312, opened 2023-12-19
fix: spontaneous zoom after navigation + resize

### LINKED ISSUE #5265: Flickering Zoom levels
### Bug description

Zoom/resize problem when bouncing between chrome and podman desktop. You'll see in the video, I define a zoom level, I switch to chrome, I close podman desktop and sometime the zoom level is reset (either bigger or smaller).
I also have the same behavior happening when I resize the podman desktop window.
Happens with or without an external monitor.
Reproduced on Linux too

https://github.com/containers/podman-desktop/assets/1636769/eb2d7588-dfd3-4b2a-8d52-e6ad8939c22b

(see in the steps to reproduce the explanations about the video)

### Operating system

macOS 14.2.1
windows 10

### Installation Method

### Version

1.6.0

### Steps to reproduce

- I start at normal zoom level (just the one when starting the app)
- I bounce between chrome and PD
- Then I close with the "closing app button" on mac
- _BUG1 (0:11 --> 0:24)_ Then I reopen it by going to the tray menu, dashboard --> **Here it automatically Zoomed In (probably 2 or 3 levels)**
- (0:24 --> 0:29) Then I'm zooming in with keyboard shortcut --> working fine
- (0:29 --> 0:40) Then I'm moving around in the app and it works fine
- _BUG2 (0:40 --> 0:42)_ I'm zooming out. **On the first press on the keyboard short cut, it is zooming out by 2 or 3 levels**
- (0:42 --> 1:12) I'm again bouncing between chrome and my environment. Things are going fine
- _BUG3 (1:12 --> 1:15)_ I'm now resizing my podman desktop window. And you can see that **the zoom level change while I'm resizing, but I'm not changing anything manually.**

### Relevant log output

_No response_

### Additional context

_No response_

dgolovin:
### What does this PR do?

This fix creates internal http server with OS dfined random port and serves podman-desktop html through it. That fixes spontaneous zoom change in app. In the case when there are problems to start server it resorts to serving from file.

### Screenshot / video of UI

Before fix:

https://github.com/containers/podman-desktop/assets/620330/be2f270b-9cfa-4b0a-9aa0-07ef473749e5

After fix:

https://github.com/containers/podman-desktop/assets/620330/03f9d2cc-1a6e-4fb9-a05f-fd739c79b560

Fix #5265

### How to test this PR?

* pull PR branch
* run `yarn compile:next`
* install podman-desktop from dist folder
* repeat steps from screen-casts above

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,6 +168,7 @@
     "got": "^13.0.0",
     "hpagent": "^1.2.0",
     "moment": "^2.29.4",
+    "node-static": "^0.7.11",
     "os-locale": "^6.0.2",
     "stream-json": "^1.8.0",
     "sudo-prompt": "^9.2.1",
--- a/packages/main/src/mainWindow.ts
+++ b/packages/main/src/mainWindow.ts
@@ -20,11 +20,13 @@ import type { BrowserWindowConstructorOptions, FileFilter } from 'electron';
 import { autoUpdater, Menu, BrowserWindow, ipcMain, app, dialog, screen, nativeTheme } from 'electron';
 import contextMenu from 'electron-context-menu';
 import { aboutMenuItem } from 'electron-util';
-import { join } from 'path';
+import path, { join } from 'path';
 import { URL } from 'url';
 import type { ConfigurationRegistry } from './plugin/configuration-registry.js';
 import { isLinux, isMac, stoppedExtensions } from './util.js';
 import { OpenDevTools } from './open-dev-tools.js';
+import http from 'http';
+import nodeStatic = require('node-static');
 
 const openDevTools = new OpenDevTools();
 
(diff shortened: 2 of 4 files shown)


dgolovin:
@slemeur FYI

benoitf (commented):
I'm kind of -1 to change the way we load the html files

package.json
```diff
@@ -168,6 +168,7 @@
     "got": "^13.0.0",
     "hpagent": "^1.2.0",
     "moment": "^2.29.4",
+    "node-static": "^0.7.11",
```
benoitf, line 171: we can't introduce this library. It is not maintained and has security issues

benoitf (changes requested):
node-static library can't be used (5years since last release)

while it helps for investigation, I think other ways should be explored

slemeur:
I'm indeed not able to reproduce the issue with this PR

dgolovin:
It can be done with any other server implementation module and for macOS only.

cdrage (changes requested):
I'm unable to reproduce the original issue (what a weird bug) so in this PR is has been fixed, but I agree that we should not use node-static.

Is it a common way to do this? I'm trying to find examples online with regards to serving an electron app through node-static / http, but I'm unable.

dgolovin:
After playing with it some more time it seems can be fixed, the way it keep 1.0 ratio when I am switching between pages, but it fails as soon as I change zoom.

Vscode and slack seems to work fine. I have one more trick to test and then the only option what @cdrage suggested.

Another option is create simple app that replicates the bug and open issue for electron.

Issue #4562, opened 2023-10-30
Kubernetes context UI: Add connection information, API, # of Pods, Deployments, etc.


cdrage:
### Is your enhancement related to a problem? Please describe

After https://github.com/containers/podman-desktop/issues/3919 we should add non-blocking retrieval information that will load when viewing the context UI. For example.. pinging and checking the connection as well as the amount of pods, deployments, etc.

### Describe the solution you'd like

Connection / cluster information being shown other than what's provided by .kube/config

afbjorklund:
~~Not version ?~~

Never mind, the version _was_ there - it was listed in grey-on-grey under the table with other information.

feloy:
Fixed by #5912 and #5917

PR #5329, opened 2023-12-19
feat: create informers for kube contexts

feloy:

### What does this PR do?

- adds a new boolean Preference entry `Get information about clusters declared in Kubeconfig`, to enable / disable the running of informers watching pods/deployments/replicasets for different kube contexts
- if enabled, the kubernetes client singleton manages a state of informers for the different contexts (stops informers of delete contexts, starts new informers for new contexts)
- if enabled, displays in the list of kube contexts if the cluster is reachable, and the number of current pods/deployemnts/replicasets for each kube context

#4562

### How to test this PR?

- Enable the feature in the Preferences
- Switch contexts and check information is updated
- add/remove resources (pods, deployments) from context and check information is updated
-

### DIFF
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -35,6 +35,7 @@ import type {
   KubernetesListObject,
   Informer,
   ObjectCache,
+  V1ReplicaSet,
 } from '@kubernetes/client-node';
 import {
   ApisApi,
@@ -66,16 +67,8 @@ import * as jsYaml from 'js-yaml';
 import type { KubeContext } from './kubernetes-context.js';
 import type { KubernetesInformerManager } from './kubernetes-informer-registry.js';
 import type { KubernetesInformerResourcesType } from './api/kubernetes-informer-info.js';
+import type { ContextInternalState, ContextState } from './kubernetes-context-state.js';
 import type { IncomingMessage } from 'node:http';
-
-interface ContextInternalState {
-  podInformer?: Informer<V1Pod> & ObjectCache<V1Pod>;
-}
-
-export interface ContextState {
-  reachable: boolean;
-  podsCount: number;
-}
 
 function toContainerStatus(state: V1ContainerState | undefined): string {
   if (state) {
@@ -480,10 +473,15 @@ export class KubernetesClient {
       // Add informers for new contexts
       for (const context of this.kubeConfig.contexts) {
         if (this.contextsInternalState.get(context.name) === undefined) {
-          this.contextsState.set(context.name, { reachable: false, podsCount: 0 });
-          const informer = this.createKubeContextInformer(context);
-          if (informer) {
-            this.contextsInternalState.set(context.name, { podInformer: informer });
+          this.contextsState.set(context.name, {
+            reachable: false,
+            podsCount: 0,
+            deploymentsCount: 0,
+            replicasetsCount: 0,
+          });
+          const informers = this.createKubeContextInformers(context);
+          if (informers) {
+            this.contextsInternalState.set(context.name, informers);
           }
         }
       }
@@ -491,6 +489,8 @@ export class KubernetesClient {
       for (const [name, state] of this.contextsInternalState) {
         if (!this.kubeConfig.contexts.find(c => c.name === name)) {
           await state.podInformer?.stop();
+          await state.deploymentInformer?.stop();
+          await state.replicasetInformer?.stop();
           this.contextsInternalState.delete(name);
           this.contextsState.delete(name);
         }
@@ -498,6 +498,8 @@ export class KubernetesClient {
     } else {
       for (const state of this.contextsInternalState.values()) {
         await state.podInformer?.stop();
+        await state.deploymentInformer?.stop();
+        await state.replicasetInformer?.stop();
       }
       this.contextsInternalState.clear();
       this.contextsState.clear();
@@ -1301,7 +1303,7 @@ export class KubernetesClient {
     }
   }
 
-  createKubeContextInformer(context: KubeContext): (Informer<V1Pod> & ObjectCache<V1Pod>) | undefined {
+  createKubeContextInformers(context: KubeContext): ContextInternalState | undefined {
     const kc = new KubeConfig();
     const cluster = this.kubeConfig.clusters.find(c => c.name === context.cluster);
     const user = this.kubeConfig.users.find(u => u.name === context.user);
@@ -1315,8 +1317,16 @@ export class KubernetesClient {
       currentContext: context.name,
     });
 
-    const k8sApi = kc.makeApiClient(CoreV1Api);
     const ns = context.namespace ?? 'default';
+    return {
+      podInformer: this.createPodInformer(kc, ns, context),
+      deploymentInformer: this.createDeploymentInformer(kc, ns, context),
+      replicasetInformer: this.createReplicasetInformer(kc, ns, context),
+    };
+  }
+
+  createPodInformer(kc: KubeConfig, ns: string, context: KubeContext): Informer<V1Pod> & ObjectCache<V1Pod> {
+    const k8sApi = kc.makeApiClient(CoreV1Api);
     const listFn = () => k8sApi.listNamespacedPod(ns);
     const informer = makeInformer(kc, `/api/v1/namespaces/${ns}/pods`, listFn);
 
@@ -1359,7 +1369,75 @@ export class KubernetesClient {
     return informer;
   }
 
-  restartInformer(informer: Informer<V1Pod> & ObjectCache<V1Pod>, context: KubeContext) {
+  createDeploymentInformer(
+    kc: KubeConfig,
+    ns: string,
+    context: KubeContext,
+  ): Informer<V1Deployment> & ObjectCache<V1Deployment> {
+    const k8sApi = kc.makeApiClient(AppsV1Api);
+    const listFn = () => k8sApi.listNamespacedDeployment(ns);
+    const informer = makeInformer(kc, `/apis/apps/v1/namespaces/${ns}/deployments`, listFn);
+
+    informer.on('add', (_obj: V1Deployment) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.deploymentsCount++;
+      }
+      this.dispatchContextsState();
+    });
+
+    informer.on('delete', (_obj: V1Deployment) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.deploymentsCount--;
+      }
+      this.dispatchContextsState();
+    });
+    informer.on('error', (_err: unknown) => {
+      // Restart informer after 5sec
+      setTimeout(() => {
+        this.restartInformer(informer, context);
+      }, 5000);
+    });
+    this.restartInformer(informer, context);
+    return informer;
+  }
+
+  createReplicasetInformer(
+    kc: KubeConfig,
+    ns: string,
+    context: KubeContext,
+  ): Informer<V1ReplicaSet> & ObjectCache<V1ReplicaSet> {
+    const k8sApi = kc.makeApiClient(AppsV1Api);
+    const listFn = () => k8sApi.listNamespacedReplicaSet(ns);
+    const informer = makeInformer(kc, `/apis/apps/v1/namespaces/${ns}/replicasets`, listFn);
+
+    informer.on('add', (_obj: V1ReplicaSet) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.replicasetsCount++;
+      }
+      this.dispatchContextsState();
+    });
+
+    informer.on('delete', (_obj: V1ReplicaSet) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.replicasetsCount--;
+      }
+      this.dispatchContextsState();
+    });
+    informer.on('error', (_err: unknown) => {
+      // Restart informer after 5sec
+      setTimeout(() => {
+        this.restartInformer(informer, context);
+      }, 5000);
+    });
+    this.restartInformer(informer, context);
+    return informer;
+  }
+
+  restartInformer(informer: Informer<KubernetesObject> & ObjectCache<KubernetesObject>, context: KubeContext) {
     const kubernetesConfiguration = this.configurationRegistry.getConfiguration('kubernetes');
     const liveContextsInfo = kubernetesConfiguration.get<boolean>('LiveContextsInfo');
     if (!liveContextsInfo) {
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -33,9 +33,6 @@ import type {
   KubernetesObject,
   ListPromise,
   KubernetesListObject,
-  Informer,
-  ObjectCache,
-  V1ReplicaSet,
 } from '@kubernetes/client-node';
 import {
   ApisApi,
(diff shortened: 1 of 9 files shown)


packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+      this.contextsInternalState.clear();
+      this.contextsState.clear();
+      this.dispatchContextsState();
+    }
+    console.log('==> # informers', this.contextsInternalState.size, this.contextsState.size);
+  }
+
+  private createKubeContextInformers(context: KubeContext): ContextInternalState | undefined {
+    const kc = new KubeConfig();
+    const cluster = this.kubeConfig.clusters.find(c => c.name === context.cluster);
```
dgolovin, line 81:
```suggestion
    const cluster = this.kubeConfig.clusters.find(c => c.name === context.cluster);
    if (!cluster) {
      retrun;
    }
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+    const user = this.kubeConfig.users.find(u => u.name === context.user);
+    if (!cluster || !user) {
```
dgolovin:
```suggestion
    if (!user) {
```

dgolovin: Why calculate user if there is no cluster?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    informer.on('delete', (_obj: V1Pod) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.podsCount--;
+      }
+      this.dispatchContextsState();
+    });
+    informer.on('error', (err: unknown) => {
+      console.error(`==> ${context.name}: error`, err);
```
dgolovin:
```suggestion
      console.error(`Pod informer error for context '${context.name}':`, String(err));
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+        previous.reachable = err === undefined;
+      }
+      this.dispatchContextsState();
+      // Restart informer after 5sec
+      setTimeout(() => {
+        this.restartInformer(informer, context);
+      }, 5000);
+    });
+    informer.on('connect', (err: unknown) => {
+      console.error(`==> ${context.name}: connect`, err);
```
dgolovin:
```suggestion
      console.error(`Pod informer connection error for context '${context.name}':`, String(err));
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    informer.on('add', (_obj: V1Pod) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.podsCount++;
+      }
+      this.dispatchContextsState();
+    });
+
+    informer.on('delete', (_obj: V1Pod) => {
```
dgolovin:
```suggestion
    informer.on('delete', () => {
```
My understanding the 'underscore' would be good if you have first parameter that is not used and then some more that is used. If it is the only parameter, you can just skip it.

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+  private createReplicasetInformer(
+    kc: KubeConfig,
+    ns: string,
+    context: KubeContext,
+  ): Informer<V1ReplicaSet> & ObjectCache<V1ReplicaSet> {
+    const k8sApi = kc.makeApiClient(AppsV1Api);
+    const listFn = () => k8sApi.listNamespacedReplicaSet(ns);
+    const informer = makeInformer(kc, `/apis/apps/v1/namespaces/${ns}/replicasets`, listFn);
+
+    informer.on('add', (_obj: V1ReplicaSet) => {
```
dgolovin:
```suggestion
    informer.on('add', () => {
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.replicasetsCount++;
+      }
+      this.dispatchContextsState();
+    });
+
+    informer.on('delete', (_obj: V1ReplicaSet) => {
```
dgolovin:
```suggestion
    informer.on('delete', () => {
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+    });
+
+    informer.on('delete', (_obj: V1Deployment) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.deploymentsCount--;
+      }
+      this.dispatchContextsState();
+    });
+    informer.on('error', (_err: unknown) => {
```
dgolovin:
```suggestion
    informer.on('error', () => {
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    informer.on('add', (_obj: V1Deployment) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.deploymentsCount++;
+      }
+      this.dispatchContextsState();
+    });
+
+    informer.on('delete', (_obj: V1Deployment) => {
```
dgolovin:
```suggestion
    informer.on('delete', () => {
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+  private createDeploymentInformer(
+    kc: KubeConfig,
+    ns: string,
+    context: KubeContext,
+  ): Informer<V1Deployment> & ObjectCache<V1Deployment> {
+    const k8sApi = kc.makeApiClient(AppsV1Api);
+    const listFn = () => k8sApi.listNamespacedDeployment(ns);
+    const informer = makeInformer(kc, `/apis/apps/v1/namespaces/${ns}/deployments`, listFn);
+
+    informer.on('add', (_obj: V1Deployment) => {
```
dgolovin:
```suggestion
    informer.on('add', () => {
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+    });
+
+    informer.on('delete', (_obj: V1ReplicaSet) => {
+      const previous = this.contextsState.get(context.name);
+      if (previous) {
+        previous.replicasetsCount--;
+      }
+      this.dispatchContextsState();
+    });
+    informer.on('error', (_err: unknown) => {
```
dgolovin:
```suggestion
    informer.on('error', () => {
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+      replicasetInformer: this.createReplicasetInformer(kc, ns, context),
+    };
+  }
+
+  private createPodInformer(kc: KubeConfig, ns: string, context: KubeContext): Informer<V1Pod> & ObjectCache<V1Pod> {
+    const k8sApi = kc.makeApiClient(CoreV1Api);
+    const listFn = () => k8sApi.listNamespacedPod(ns);
+    const informer = makeInformer(kc, `/api/v1/namespaces/${ns}/pods`, listFn);
+
+    informer.on('add', (_obj: V1Pod) => {
```
dgolovin:
```suggestion
    informer.on('add', () => {
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+      for (const state of this.contextsInternalState.values()) {
+        await state.podInformer?.stop();
+        await state.deploymentInformer?.stop();
+        await state.replicasetInformer?.stop();
+      }
+      this.contextsInternalState.clear();
+      this.contextsState.clear();
+      this.dispatchContextsState();
+    }
+    console.log('==> # informers', this.contextsInternalState.size, this.contextsState.size);
```
dgolovin: Replace `==>` with meaningful, human readable message.

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+  private restartInformer(informer: Informer<KubernetesObject> & ObjectCache<KubernetesObject>, context: KubeContext) {
+    const kubernetesConfiguration = this.configurationRegistry.getConfiguration('kubernetes');
+    const liveContextsInfo = kubernetesConfiguration.get<boolean>('LiveContextsInfo');
+    if (!liveContextsInfo) {
+      return;
+    }
+
+    informer
+      .start()
+      .then(() => {})
```
dgolovin: Why is that required?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,243 @@
...
+      .catch((err: unknown) => {
+        console.log('==> catched err', err);
```
dgolovin: There should be readable message that would explain where the error has happened.

dgolovin:
```suggestion
        console.log('==> catched err', String(err));
```

dgolovin (changes requested):

packages/main/src/plugin/kubernetes-context-state.ts
feloy: I'm using this `==>` sign to grep into it, to specifically get my logs during debug sessions. Most of these logs will probably disappear before the final revision

packages/main/src/plugin/kubernetes-context-state.ts
feloy: Not sure it is. Removing it

packages/main/src/plugin/kubernetes-context-state.ts
feloy: This should be fixed with latest commit

feloy:
Thanks @dgolovin for your review. I've added unit tests, the PR is ready for review again

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -157,7 +162,7 @@ export class KubernetesClient {
     // add configuration
     const kubeconfigConfigurationNode: IConfigurationNode = {
       id: 'preferences.Kubernetes.Kubeconfig',
-      title: 'Path to the kubeconfig file',
+      title: 'Kubernetes Contexts',
```
deboer-tim: I'm really sorry about the merge conflict I (inadvertently) created. I still think the title should be just 'Kubernetes' here though, at least until we have several other Kubernetes preferences.

deboer-tim (changes requested):
I'm seeing some minor UI problems while testing this:
- 'Replicasets' extends underneath the cluster details on the right.
- When the window width is < 950px, the namespace and its value collide. If you keep shrinking it, cluster has the same problem. There is enough width to show these without overlapping, it just isn't using the right side of the UI.
- My Minikube cluster is old/doesn't exist. However, every ~5s it flashes and I can see 0 pods/deployments/replicasets for a split second before it resets to unreachable.

Screenshot 2024-01-16 at 5 12 44 PM

It also feels a bit odd that we chose Deployments/Ingress/Routes/Services as the first set of objects to add nav pages for, but are showing Deployments/Replicasets here. I wonder if we should use a different set here, or even just not show replicasets for now (which would also avoid the first problem).

feloy:
> * My Minikube cluster is old/doesn't exist. However, every ~5s it flashes and I can see 0 pods/deployments/replicasets for a split second before it resets to unreachable.

The problem is that, when the informer fails starting, we receive a `connect` event with a 
 null error. And we are checking if we a correctly connected to a cluster by checking this same error in the `connect` event. Due to this inconsistency, we have two choices:

- have a small delay (<1s), and change a state from connected<>disconnect after this delay
- wait for a add/update/delete/change event to declare a cluster reachable. But in this case, a cluster with no resource will be declared as unreachable. To fix this we could watch for nodes, for examples, for which we know we will always have a resource, and then always receive at least an `add` event (but the problem is that many users won't have permissions to watch this `node` resource)

feloy:
> > * My Minikube cluster is old/doesn't exist. However, every ~5s it flashes and I can see 0 pods/deployments/replicasets for a split second before it resets to unreachable.
> 
> The problem is that, when the informer fails starting, we receive a `connect` event with a null error. And we are checking if we a correctly connected to a cluster by checking this same error in the `connect` event. Due to this inconsistency, we have two choices:
> 
> * have a small delay (<1s), and change a state from connected<>disconnect after this delay
> * wait for a add/update/delete/change event to declare a cluster reachable. But in this case, a cluster with no resource will be declared as unreachable. To fix this we could watch for nodes, for examples, for which we know we will always have a resource, and then always receive at least an `add` event (but the problem is that many users won't have permissions to watch this `node` resource)

I've added a new commit with the first solution (waiting a short delay before sending the value of the reachable state, and do not send it if a new one it sent later), to avoid flashing

deboer-tim:
I looked through this again - the work is really good and UI just has a couple minor issues, but we've let it stagnate. We already have other PRs like #5850 that are solving some of the same problems. Assuming @feloy is still tied up in other work, I'd like to find someone else who can take over this PR soon and wrap up the (hopefully) last few things.

feloy:
> I looked through this again - the work is really good and UI just has a couple minor issues, but we've let it stagnate. We already have other PRs like https://github.com/containers/podman-desktop/pull/5850 that are solving some of the same problems. Assuming @feloy is still tied up in other work, I'd like to find someone else who can take over this PR soon and wrap up the (hopefully) last few things.

I should be able to finish this PR, could you sumarize the UI changes that would be needed?

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -95,32 +96,72 @@ async function handleDeleteContext(contextName: string) {
           {/if}
         </div>
         <div class="grow flex-column divide-gray-900 text-gray-400">
-          <div class="text-xs bg-charcoal-800 p-2 rounded-lg mt-1 grid grid-cols-6">
-            <span class="my-auto font-bold col-span-1 text-right">CLUSTER</span>
-            <span class="my-auto col-span-5 text-left pl-0.5 ml-3" aria-label="context-cluster">{context.cluster}</span>
-          </div>
+          <div class="flex flex-row">
+            {#if $kubernetesContextsState.get(context.name)}
```
dgolovin, line 100: I have current context pointing to sandbox instance and I can query resources using `kubectl`. That means my context state should be reachable, but for some reason I don't see part of UI below.

dgolovin (changes requested):
The part with counters does not show up for active reachable context pointing to sandbox instance.

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: Did you enable the feature in the preferences, as the feature is disabled by default?

feloy:
As discussed during daily meeting, I've ushed a new commit removing the replicaset information.

I'll split this PR into several smaller ones.

feloy:
Closing in favor of #5912  and following

PR #5333, opened 2023-12-20
fix: stop provider card overflow

### LINKED ISSUE #5316: Dashboard provider card is garbage
### Bug description

While testing OpenShift Local extension, I noticed a garbled display:

image

If the window is maximized, then it disappear

image

### Operating system

Win11Pro

### Installation Method

None

### Version

1.6.3

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Small tweaks to ProviderCard to make sure the icon margin isn't growing and right column is growing correctly. Uses max-w-xs to limit the width of the right column wherever text was not wrapping correctly. Removes redundant provider version from the configured state.

Even with these changes it is still possible to get overlap when the window is between 640px (the minimum) and 690px wide. We're at the point where the required content does not fit the current design. If anyone else has a suggestion I'm willing to try it, but IMHO it may not be worth it as this is more of an extreme/edge case now, and we've been waiting for a couple sprints now on a design for moving some of the actions to other places like Resources.

Since this is visual overlap of elements not staying within their flex layout I can't think of any useful tests.

### Screenshot / video of UI

Before
Screenshot 2023-12-19 at 10 42 23 PM

After:
Screenshot 2023-12-19 at 10 42 04 PM

Fixes #5316 (mostly).

### How to test this PR?

Just make the window narrow.

### DIFF
--- a/packages/renderer/src/lib/dashboard/ProviderCard.svelte
+++ b/packages/renderer/src/lib/dashboard/ProviderCard.svelte
@@ -10,7 +10,7 @@ export let provider: ProviderInfo;
 <div class="bg-charcoal-800 mb-5 rounded-md p-3 flex-nowrap" role="region" aria-label="{provider.name} Provider">
   <div class="flex flex-row">
     <div class="flex flex-row">
-      <IconImage image="{provider?.images?.icon}" class="mx-auto max-h-12" alt="{provider.name}"></IconImage>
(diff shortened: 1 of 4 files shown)


dgolovin (changes requested):
It seems does not work with zoom
image

benoitf:
move to draft as no update

PR #5345, opened 2023-12-20
feat: table update event passthrough

deboer-tim:
### What does this PR do?

The pods and containers list use an update event (previously callbacks) to let the parent page know that one column has triggered a change that requires a UI refresh in other columns. As we move to the table component there is an additional level of components, so we just need this event to pass through the table component.

This change just allows the table to pass through update events. To test the age column was changed to trigger a callback whenever it is sorted.

### Screenshot / video of UI

N/A

N/A, required by #4842.

### How to test this PR?

`yarn test:renderer`


benoitf (approved):

jeffmaury (approved):
LGTM

PR #5357, opened 2023-12-21
chore: replace CustomIcon by IconImage

benoitf:
### What does this PR do?
we have like 2 components CustomIcon and IconImage doing almost the same thing
IconImage handle light icons so switch to this one$

N/A


deboer-tim (approved):
LGTM

Issue #4576, opened 2023-10-31
Design for Services page


deboer-tim:
### Is your enhancement related to a problem? Please describe

We need a place to view and edit Kubernetes Services.

### Describe the solution you'd like

We need a design for an initial Kubernetes Services page, to be implemented by #4368.

ConorC117:
Link to Services list page: 
https://design.penpot.app/#/workspace/76981dd7-ec3b-802f-8001-9a56ef3fbc9d/18f3baa5-c6db-81d1-8002-9f756ec487eb?page-id=28aff1ad-ce4d-807a-8003-7157ad663023 

Screenshot of initial draft: 
![image](https://github.com/containers/podman-desktop/assets/99751994/0f8fe63e-5f82-4c14-999f-c25d3998374f)

ConorC117:
Screenshot of second draft (changed global nav icon): 

![image](https://github.com/containers/podman-desktop/assets/99751994/4af4f890-2993-4c5f-85b2-77f75d525c51)

cdrage:
> Screenshot of second draft (changed global nav icon):
> 
> ![image](https://user-images.githubusercontent.com/99751994/284385036-4af4f890-2993-4c5f-85b2-77f75d525c51.png)

This looks awesome! Here's my suggestions:
- To keep with the theme of "labels" for types, I think Type should be a label similar to environment / your other mockup with showing Progressing, etc.
- Let's remove the "selectors"? I know other graphical tools have it, but I think it clutters is up and is not very useful for "overall" graphical viewing. I believe we are going to eventually add information on if you "click" on the service it will show the YAML / details, so we can easily show it there instead what the selectors are. 
- We should add a way to show "external IP". In your example ClusterIP wont have an external IP, but any type that has a LoadBalancer will, see: https://kubernetes.io/docs/tutorials/stateless-application/expose-external-ip-address/ IMHO clusterIP will be the most common as well as nodeport. I think we should it that if we 'hover' over the LoadBalancer label we can use Tooltip.svelte to show the IP address of the external IP.


@deboer-tim @feloy

ConorC117:
Current screenshot of Services List : 

![image](https://github.com/containers/podman-desktop/assets/99751994/82ddb289-8f3f-4e1a-9854-6cd5fe20e454)

I included colors for the 'Type' labels based on the color blind safe palette [here](https://lospec.com/palette-list/ibm-color-blind-safe) but can of course swap them around or use different colors if preferred

[...]

PR #5363, opened 2023-12-21
Kubernetes Services Page

### LINKED ISSUE #4368: Kubernetes Services page

deboer-tim:
### What does this PR do?

Adds a UI for Kubernetes Services, similar to the Deployments page. All
design elements are included, columns are sortable, and there is a
delete action.

Depends on service store (PR #5295), so ignore everything except the
last commit (i.e. only review /service folder and App/AppNavigation).
Do not merge before #5295.

### Screenshot / video of UI

Screenshot 2023-12-21 at 5 00 23 PM

Fixes #4368.

### How to test this PR?

Like the Deployments page, the Services page will appear whenever you have a kube context.

### DIFF
--- a/packages/main/src/plugin/api/kubernetes-informer-info.ts
+++ b/packages/main/src/plugin/api/kubernetes-informer-info.ts
@@ -17,7 +17,7 @@
  ***********************************************************************/
 import type { Context, Informer, KubernetesObject } from '@kubernetes/client-node';
 
-export type KubernetesInformerResourcesType = 'INGRESS';
+export type KubernetesInformerResourcesType = 'DEPLOYMENT' | 'INGRESS' | 'ROUTE' | 'SERVICE';
 
 export interface KubernetesInformerInfo {
   informer: Informer<KubernetesObject>;
--- a/packages/main/src/plugin/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes-client.spec.ts
@@ -59,6 +59,20 @@ class TestKubernetesClient extends KubernetesClient {
   }
 }
 
+function createTestClient(namespace?: string): TestKubernetesClient {
+  const client = new TestKubernetesClient(
+    apiSender,
+    configurationRegistry,
+    fileSystemMonitoring,
+    informerManager,
+    telemetry,
+  );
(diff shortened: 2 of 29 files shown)


benoitf (approved):
I was wondering if we should have named services objects KubernetesServices to make it less confusing ( in the code)

not a required change

slemeur:
pretty cool ! Nice work @deboer-tim  
cc @axel7083 ;)

axel7083:
This is amazing job !

feloy:
It seems to me there are missing some important information: the workloads served by these services. The 2 information important to me would be:
- the selector (what the user has defined)
- the list of endpoints, or at least the number of endpoints (what is the current state)

deboer-tim:
> It seems to me there are missing some important information: the workloads served by these services. The 2 information important to me would be:
> 
> * the selector (what the user has defined)
> * the list of endpoints, or at least the number of endpoints (what is the current state)

@feloy the design discussion was in #4576 and this matches that. I'm fine changing, but there are already a lot of columns so I think we're at the point where every suggested addition should say which column it would replace - i.e. we only get to show <= 6 key things on the summary list, everything else would go in details like it does for pods or containers.

feloy:
> > It seems to me there are missing some important information: the workloads served by these services. The 2 information important to me would be:
> > 
> > * the selector (what the user has defined)
> > * the list of endpoints, or at least the number of endpoints (what is the current state)
> 
> @feloy the design discussion was in #4576 and this matches that. I'm fine changing, but there are already a lot of columns so I think we're at the point where every suggested addition should say which column it would replace - i.e. we only get to show <= 6 key things on the summary list, everything else would go in details like it does for pods or containers.

That's ok for me. Just a note for future discussions

slemeur:
It's actually a good idea, let's follow up on https://github.com/containers/podman-desktop/issues/5367

Issue #4403, opened 2023-10-18
Design for Deployments page


deboer-tim:
### Is your feature request related to a problem? Please describe

Kubernetes!

### Describe the solution you'd like

We need a design for an initial Kubernetes Deployments page, to be implemented by #4366.

ConorC117:
Link to Deployments List page: 

https://design.penpot.app/#/workspace/76981dd7-ec3b-802f-8001-9a56ef3fbc9d/18f3baa5-c6db-81d1-8002-9f756ec487eb?page-id=88a72dea-ea00-80c5-8003-65badc968c16

deboer-tim:
Current screenshot:
<img width="810" alt="Screenshot 2023-11-07 at 8 38 14 AM" src="https://github.com/containers/podman-desktop/assets/19958075/b2c7e829-4494-4df8-a836-b11e0c708c61">

deboer-tim:
@mairin I just noticed that both this design and Ingress/Routes don't have any nav icon on the left. We don't need a mockup for that, but we do need svgs to use. Any chance you can round one up? (and may as well prep Services too, since that's in the pipe)

mairin:
@ConorC117 can you create an icon for deployments based on the kube icon here with the loopy arrow? (upper right)

https://uploads-ssl.webflow.com/5e5a5c35a623e45f7f701878/5e99435c7ab3fdfb624ac319_Screen-Shot-2020-03-11-at-12.50.59-AM-768x616.png

cdrage:
> Current screenshot: <img alt="Screenshot 2023-11-07 at 8 38 14 AM" width="810" src="https://user-images.githubusercontent.com/19958075/281058244-b2c7e829-4494-4df8-a836-b11e0c708c61.png">

My feedback:

- The "Reasons" column is odd as "Reasons" are shown as an array of events for errors. Instead, if it is has a warning, we should hover over the "Warning" symbol and then use Tooltip.svelte to show what the warning / error is.

![Screenshot 2023-11-09 at 8 21 25 AM](https://github.com/containers/podman-desktop/assets/6422176/72437877-63ae-4334-8009-a0d44d94b486)


- Pods we could possibly work in the dots there? One thing too is that having it as 1/2 Pods may be difficult to figure out which pod is not working correctly.


https://github.com/containers/podman-desktop/assets/6422176/8321dbfa-e3e8-4551-96df-1b6a58869771

[...]

PR #5405, opened 2024-01-02
feat: display Kubernetes connection status in relevant pages

axel7083:
### What does this PR do?

This PR is adding a badge to the Service and Deployment page indicating the status ('connected' | 'error') to the current context (when exist).

### Screenshot / video of UI

**error display**
image

**on hover**
image

**connected**
image

https://github.com/containers/podman-desktop/issues/5401

### How to test this PR?

- Create / delete kind cluster
- Switch between contexts

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -149,6 +149,7 @@ import type { KubeContext } from './kubernetes-context.js';
 import { KubernetesInformerManager } from './kubernetes-informer-registry.js';
 import type { KubernetesInformerResourcesType } from './api/kubernetes-informer-info.js';
 import { OpenDevToolsInit } from './open-devtools-init.js';
+import type { KubernetesConnection } from '/@/plugin/kubernetes-connection.js';
 
 type LogType = 'log' | 'warn' | 'trace' | 'debug' | 'error';
 
@@ -1996,6 +1997,12 @@ export class PluginSystem {
 
     this.ipcHandle('kubernetes-client:getDetailedContexts', async (): Promise<KubeContext[]> => {
       return kubernetesClient.getDetailedContexts();
+    });
+
+    this.ipcHandle('kubernetes-client:getConnectionStatus', async (): Promise<KubernetesConnection | undefined> => {
+      const value = kubernetesClient.getConnectionStatus();
+      console.log('getConnectionStatus', value);
+      return value;
     });
 
     this.ipcHandle('kubernetes-client:getClusters', async (): Promise<Cluster[]> => {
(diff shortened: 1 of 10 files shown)


deboer-tim:
Out of context the badges look a bit like buttons to me (they've only been used in table columns so far), I'm wondering if we should use a different UI element or just not have the badge background.

The original designs for all these pages (e.g. #4403) is to allow multiple contexts or namespaces, so a 'global' status wouldn't work long term, but until we get there there isn't an obvious other place to put this.

For both reasons I think it's worth a quick design discussion.

axel7083:
> Out of context the badges look a bit like buttons to me (they've only been used in table columns so far), I'm wondering if we should use a different UI element or just not have the badge background.
> 
> The original designs for all these pages (e.g. #4403) is to allow multiple contexts or namespaces, so a 'global' status wouldn't work long term, but until we get there there isn't an obvious other place to put this.
> 
> For both reasons I think it's worth a quick design discussion.

I agree for the button, as you said it is the one used in the column. I read the #4403 original design page, and I have some questions
- Does all contexts will be automatically connected and displayed ?
- Does the page will not have a overload look if we display many contexts in the same page ?
> I worked with some Kubernetes cluster where we had hundreds of pods in a single namespace, I am a bit afraid for the scaling 
- Does the search bar search for all the pods/services in all contexts ?

slemeur:
At the moment, Podman Desktop displays the objects from the selected kubecontext and the objects from Podman - pods and deployments at least.  

As the connectivity information is related to the kubecontext, it would make sense to me to display it along with the kubecontext in the status bar. This way as a user, I have a single place to see the information about my current context: which one is selected, what is that connectivity status.

On the other hand, I think the current empty screen is misleading. Right now, we are saying there are no services to display, while in fact, there is a connectivity issue... That error should be reported to the user in a more obvious way and I would not hide the error behind a tooltip. We could probably, add in the empty screen message, the connectivity error.

axel7083:
> At the moment, Podman Desktop displays the objects from the selected kubecontext and the objects from Podman - pods and deployments at least.
> 
> As the connectivity information is related to the kubecontext, it would make sense to me to display it along with the kubecontext in the status bar. This way as a user, I have a single place to see the information about my current context: which one is selected, what is that connectivity status.

Adding the connection status to the status bar along it could be interesting, but would require some internal changes

> On the other hand, I think the current empty screen is misleading. Right now, we are saying there are no services to display, while in fact, there is a connectivity issue... That error should be reported to the user in a more obvious way and I would not hide the error behind a tooltip. We could probably, add in the empty screen message, the connectivity error.

I agree that the `empty screen is misleading` as we do not see the difference between nothing to display vs an error.

PR #5406, opened 2024-01-02
:fix: typo in shell examples

tjajt:
add missing semi colon between setting shell variables in podman verifying-your-tools-are-using-podman.

### What does this PR do?

fixes documentation guidance for those movinf from docker.

### How to test this PR?

copy and execute the commands in macOS or unix shell environment, the previous examples were missing a ';' between the setting of the CONTAINER_HOST variable and the command to execute afterward. Without it a posix shell would interpret the entire command as a whole, producing no result from the 'podman ps' command, this might confuse users.

### DIFF
--- a/website/docs/migrating-from-docker/verifying-your-tools-are-using-podman.md
+++ b/website/docs/migrating-from-docker/verifying-your-tools-are-using-podman.md
@@ -55,21 +55,21 @@ When you have configured your host to use Podman rather then Docker, consider ve
      <TabItem value="win" label="Windows">
 
    ```shell-session
-   $ CONTAINER_HOST=npipe:////./pipe/docker_engine podman ps
+   $ CONTAINER_HOST=npipe:////./pipe/docker_engine; podman ps
    ```
 
      </TabItem>
      <TabItem value="mac" label="macOS">
 
    ```shell-session
-   $ CONTAINER_HOST=/var/run/docker.sock podman ps
+   $ CONTAINER_HOST=/var/run/docker.sock; podman ps
    ```
 
      </TabItem>
      <TabItem value="linux" label="Linux">
 
    ```shell-session
-   $ CONTAINER_HOST=/var/run/docker.sock podman ps
+   $ CONTAINER_HOST=/var/run/docker.sock; podman ps
    ```
 
      </TabItem>


axel7083 (approved):
LGTM Thanks for the fix!

website/docs/migrating-from-docker/verifying-your-tools-are-using-podman.md
```diff
@@ -55,21 +55,21 @@ When you have configured your host to use Podman rather then Docker, consider ve
      <TabItem value="win" label="Windows">
 
    ```shell-session
-   $ CONTAINER_HOST=npipe:////./pipe/docker_engine podman ps
+   $ CONTAINER_HOST=npipe:////./pipe/docker_engine; podman ps
```
jeffmaury, line 58:
```suggestion
   $ set CONTAINER_HOST=npipe:////./pipe/docker_engin && podman ps
```

website/docs/migrating-from-docker/verifying-your-tools-are-using-podman.md
```diff
@@ -55,21 +55,21 @@ When you have configured your host to use Podman rather then Docker, consider ve
...
    ```
 
      </TabItem>
      <TabItem value="mac" label="macOS">
 
    ```shell-session
-   $ CONTAINER_HOST=/var/run/docker.sock podman ps
+   $ CONTAINER_HOST=/var/run/docker.sock; podman ps
```
jeffmaury, line 65: Not required on MacOS

website/docs/migrating-from-docker/verifying-your-tools-are-using-podman.md
```diff
@@ -55,21 +55,21 @@ When you have configured your host to use Podman rather then Docker, consider ve
...
    ```
 
      </TabItem>
      <TabItem value="linux" label="Linux">
 
    ```shell-session
-   $ CONTAINER_HOST=/var/run/docker.sock podman ps
+   $ CONTAINER_HOST=/var/run/docker.sock; podman ps
```
jeffmaury, line 72: Not required on Linux

jeffmaury (changes requested):

website/docs/migrating-from-docker/verifying-your-tools-are-using-podman.md
afbjorklund: It actually breaks the example, since it is no longer setting the CONTAINER_HOST environment variable.

But it was broken on many podman versions anyway, since it doesn't use podman-remote or --remote.

afbjorklund:
The examples were working fine in a normal shell, if you want to make them run in something odd like `csh` you could add an `env` in front of the command. But (almost) nobody runs those old and crufty shells anymore, with `setenv` et al.

cdrage:
I don't believe this fix is necessary as it's for older shells as noted by @afbjorklund this actually will now break examples as it's running the command separately. Similar to passing in `DOCKER_HOST=/etc/foobar docker-compose up`, this is the same concept

website/docs/migrating-from-docker/verifying-your-tools-are-using-podman.md
cdrage: this breaks examples similar to how we pass in DOCKER_HOST for compose examples, I don't believe we should implement this change.

afbjorklund:
For the record the automatic `--remote` flag was added in Podman v4, which is why it is not in Ubuntu.

commit https://github.com/containers/podman/commit/4f857bc10627d950b3363e47cee02ef07618f0ae

A workaround is to call the `podman-remote` program instead, like is being done on Mac and Windows.

Or perhaps even "podman-remote-static".

----

EDIT: The  "podman-remote" package is missing, from all releases before Debian 13 and Ubuntu 24.04

https://packages.debian.org/search?keywords=podman-remote

https://packages.ubuntu.com/search?keywords=podman-remote

But as long as you set CONTAINER_HOST (or CONTAINER_CONNECTION), the flag should not be needed

themr0c:
The absence of `;` is intentional and required. However, using the `--url=` flag instead might bring more clarity.
The absence of the `--remote` flag is a mistake.

afbjorklund:
> The absence of the `--remote` flag is a mistake.

Podman v4 automatically adds the --remote flag, but it was still needed in Podman v3 (only when running on Linux)

It shouldn't _hurt_ anything, so I guess it could be added for as long as Podman v3 is the default in some setups?

Unfortunately `podman-remote` seems to be missing from several distributions (again, mostly for older versions), and the binary from the Podman releases is called `podman-remote-static` (for some reason) so it has a different name.

Always adding `--remote` is probably the safest option, and maybe it can be separated by OS (or templated perhaps).
This what we ended up doing in Minikube and in Lima, but the easiest option is to just use the `docker` client. :-P

----

minikube (one tab, for each OS)

https://minikube.sigs.k8s.io/docs/handbook/pushing/#3-pushing-directly-to-in-cluster-cri-o-podman-env

Note: it explicitly uses `podman-remote`, since you **have** to download an older binary to work with Ubuntu.

```
Error: unable to connect to Podman socket: server API version is too old. Client "4.0.0" server "3.4.2"
```

lima (templated, with .HostOS)

```yaml
message: |
  To run `podman` on the host (assumes podman-remote is installed), run the following commands:
  ------
  podman system connection add lima-{{.Name}} "unix://{{.Dir}}/sock/podman.sock"
  podman system connection default lima-{{.Name}}
  podman{{if eq .HostOS "linux"}} --remote{{end}} run quay.io/podman/hello
  ------

```

benoitf:
closing as not updated

Issue #5140, opened 2023-12-05
extensions should be able to register web views


benoitf:
### Is your enhancement related to a problem? Please describe

A webview will provide a dedicated UI (like Docker Desktop extensions) so you'll get a full right page UI

### Describe the solution you'd like

add methods

```ts
export function createWebviewPanel(viewType: string, title: string): WebviewPanel;
```

```ts
interface WebviewPanel {
    readonly viewType: string;
    title: string;
    readonly webview: Webview;
}
```

```ts
interface Webview {
     html: string;
     postMessage(message: unknown): Promise<boolean>;
     readonly onDidReceiveMessage: Event<unknown>;
}
```

depends on:
 - [x] https://github.com/containers/podman-desktop/issues/5359

axel7083:
To follow up, here is the list of things that we noticed

- Connection is lost between the front and backend of the extension if we leave the webview and come back
- States is lost, everytime the webview is open it is reloaded and loses the state (E.g. current page)
> We might want to be able to configure this aspect, as keeping all the extension loaded in background would be expensive (?)

benoitf:
I've checked and with tinro Router, component is destroyed when it's not displayed anymore, so to get it back to the location where we are we'll need to rely on the state object (for now it is not restored as you pointed it out)

benoitf:
implemented

PR #5410, opened 2024-01-03
feat: creating interface for subviews

axel7083:
### What does this PR do?

Related to #5140

This PR is trying to abstract and reuse the existing implementation used for Docker Extension UI.

> The word `subview` is used instead of `webview` (we can change it, it was ), but it was to emphasis with the relation with the parent page.
- Creating a SubviewRegistry 
- The ContributionManager will register each contribution having an UI to the SubviewRegistry
- A Subviews store is created and used by the AppNavigation (instead of Contributions)
- Rename and abstract the DockerExtension.svelte component to Subview.svelte
- Introducing `extensionApi.subview.registerSubview` method 

**warning:** The docker-extension-preload script will be used. Meaning, extension loading a custom page will have access to Docker client through `window.ddClient`.

## Future consideration

Should we allow the extension to register they own preload script ?
> Me personal thought on the subject is **not** to allow extension to register their own preload as it is runned with full node capabilities, therefore leading to potential security issue.

How subview should interact with podman apis ? Does the `extensionApi` should be available, therefore requiring to load it in the extension preload (more complex that it seems) ?

Should the Podman extension have access to the Docker Client adapter provided by the `package/docker-extension-preload` ?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2651,4 +2651,19 @@ declare module '@podman-desktop/api' {
       metadata?: ImageCheckerProviderMetadata,
     ): Disposable;
   }
+
+  export interface SubviewOption {
+    // unique identifier
+    id: string;
+    // name to display
+    name: string;
+    // the path to index.html or url
+    source: string;
+    // base64 icon
+    icon: string;
+  }
+
+  export namespace subview {
+    export function registerSubview(option: SubviewOption): Disposable;
+  }
 }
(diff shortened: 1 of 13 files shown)


benoitf:
hello, FYI I worked on the related issue so I'm not sure about the goal of this PR ?

axel7083:
> hello, FYI I worked on the related issue so I'm not sure about the goal of this PR ?

Hey! Since it has the `planned ` label I thought I could help by taking a look and start digging on the potential implementations, I can close if you already have related work

benoitf:
@axel7083 you should look if there is an assignee

or https://github.com/orgs/containers/projects/4/views/8 

some unassigned issues that are in the current sprint

PR #5421, opened 2024-01-04
fix: using window.showMessageBox instead of custom modal

### LINKED ISSUE #5290: An error dialog that appears when image is pushed to unavailable cluster is shown only when image's actions context menu is opened
### Bug description

Expected error dialog is shown only withing context menu from image's actions.

pushimagerror_badly_shown

### Operating system

Windows, Linux

### Installation Method

### Version

1.6.3

### Steps to reproduce

1. Pull an image
2. Push image to a Kind (for example) which does not exist
3. Assert: Error popup is shown
Actual result: Popup is part of context menu over image's actions

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Replace the existing error modal implemented in the ImageAction svelte component by the `window.showMessageBox`.
- Improve consistency in application
- Avoid to maintain individual implementation of error message

### Screenshot / video of UI

| Before | After |
| --- | --- |
| image  | image | 

Fixes https://github.com/containers/podman-desktop/issues/5290 https://github.com/containers/podman-desktop/issues/5420

CC @lstocchi  :)

### DIFF
--- a/packages/renderer/src/lib/image/ImageActions.svelte
+++ b/packages/renderer/src/lib/image/ImageActions.svelte
@@ -55,7 +55,7 @@ async function deleteImage(): Promise<void> {
   try {
     await imageUtils.deleteImage(image);
   } catch (error) {
-    onError(`Error while deleting image: ${error}`);
+    onError(`Error while deleting image: ${String(error)}`);
   }
 }
 
--- a/packages/renderer/src/lib/image/ImageActions.svelte
+++ b/packages/renderer/src/lib/image/ImageActions.svelte
@@ -1,13 +1,5 @@
 <script lang="ts">
-import {
-  faArrowUp,
-  faLayerGroup,
-  faPlay,
-  faTrash,
-  faEdit,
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/image/ImageActions.svelte
```diff
@@ -64,8 +52,7 @@ async function deleteImage(): Promise<void> {
   try {
     await imageUtils.deleteImage(image);
   } catch (error) {
-    errorTitle = 'Error while deleting image';
-    errorMessage = String(error);
+    onError(`Error while deleting image: ${error}`);
```
lstocchi: Before it was `String(error)` so i would keep it the same -> `Error while deleting image:  ${String(error)}`

lstocchi (commented):
LGTM, just one comment

lstocchi (approved):
Thanks!! LGTM

dgolovin (approved):
LGTM

PR #5423, opened 2024-01-04
chore(tests): adds build image test to image smoke suite

### LINKED ISSUE #3553: Add build image test case to the e2e image smoke test

I'd like to add the e2e test case for building image scenario to expand the existing e2e image smoke test. 

Implement test case covering build image scenario and verification.

_No response_

### Additional context

Issue https://github.com/containers/podman-desktop/issues/3824 is desired to be implemented for this test case functionality.

cbr7:
### What does this PR do?

Adds e2e test to cover the image build functionality

Closes #3553


tests/src/fixtures/test-Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM alpine
```
odockal: After some discussion and experiences, it is better practice to use different image registry than implicit (docker). With this image being pulled on every GHA run we will soon reach API rate limit.
I would go with `ghcr.io/linuxcontainers/alpine`.
Please also choose different folder. I would use `tests/resources/` for now.

tests/src/model/pages/build-image-page.ts
```diff
@@ -0,0 +1,62 @@
...
+  readonly buildContextDirectoryInput: Locator;
+  readonly imageNameInput: Locator;
+  readonly buildButton: Locator;
+  readonly doneButton: Locator;
+  readonly containerFilePathButton: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.heading = page.getByRole('heading', { name: 'Build Image from Containerfile' });
+    this.containerFilePathInput = page.locator(`[id='containerFilePath']`);
```
odockal: We are trying to follow playwright's best practices and prefer to use roles + accessible names (https://playwright.dev/docs/locators#locate-by-role). Sometimes it might mean that we need to include ARIA functionality into the codebase of the page in order to have it properly tested. 
Since I am not sure about particular DOM structure of the page from the top of my head, it there is a way to achieve using of `getByRole` use that one, or please also add particular aria functionality (ARIA role, aria-label). I can explain more via chat.

tests/src/model/pages/build-image-page.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    await this.containerFilePathInput.fill(containerFilePath);
+
+    if (contextDirectory) await this.buildContextDirectoryInput.fill(contextDirectory);
+    if (imageName) {
+      await this.imageNameInput.clear();
+      await this.imageNameInput.fill(imageName);
+    }
+
+    await playExpect(this.buildButton).toBeEnabled();
```
odockal: In the Page Object models it would be nice to omit assertions, if we can. We can make use of `.waitFor` for example, right?

odockal (changes requested):

tests/src/fixtures/test-Dockerfile
cbr7: done

tests/src/model/pages/build-image-page.ts
cbr7: I wasn't able to get it to work with getByRole for some reason, but I got it working with getByPlaceholder method, should look better now

tests/src/model/pages/build-image-page.ts
cbr7: done

tests/src/model/pages/build-image-page.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    await this.containerFilePathInput.fill(containerFilePath);
+
+    if (contextDirectory) await this.buildContextDirectoryInput.fill(contextDirectory);
+    if (imageName) {
+      await this.imageNameInput.clear();
+      await this.imageNameInput.fill(imageName);
+    }
+
+    await this.buildButton.waitFor({ state: 'visible', timeout: 3000 });
```
odockal: There is an implementation of a custom wait conditions in the utility class (`tests/src/utility/wait.ts`) - `waitWhile` and `waitUntil`. These can be used for passing any callback/statement to check for proper DOM's elements and their states. 

Here the snippet.:
```
    try {
      await waitUntil(
        async () => await this.buildButton.isEnabled(),
        5000,
        900,
        true,
        'Cannot build an image',
      );
    } catch (err) {
      await this.errorAlert.waitFor({ state: 'visible', timeout: 2000 });
      const errMessage = await this.errorAlert.innerText({ timeout: 1000 });
      throw new Error(`Error starting a container: ${errMessage}`);
    }
```

cbr7: done

odockal (approved):

PR #5424, opened 2024-01-04
feat: improve Authentication Providers page

### LINKED ISSUE #2274: Improve logout action, sync
### Bug description

The logout button on the Settings > Authentication page (faRightFromBracket) is not obvious and does not have a tooltip.

Clicking on it logs out, but the system tray menu (at least for RH auth) still thinks you're logged in.

### Operating system

macOS 13.3 M1

### Version

next (development version)

### Steps to reproduce

Click the icon in Settings > Authentication to log out, check the system tray to see it still thinks you're logged in.

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

This PR introduce small change to the Authentification Providers page

### Screenshot / video of UI

| Before | After |
| --- | --- |
| image | image | 
| image | image | 
| image | image | 

Related to https://github.com/containers/podman-desktop/issues/2274

- [x] Display all accounts instead of only one
- [x] Allow to sign out of individual account instead of only the first one
- [x] Add tooltip for sign out  

### How to test this PR?

You can use the https://github.com/containers/podman-desktop/pull/5400 to test an authentification provider

CC @mairin :)

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.svelte
@@ -7,6 +7,7 @@ import KeyIcon from '../images/KeyIcon.svelte';
 import EmptyScreen from '../ui/EmptyScreen.svelte';
 import DropdownMenu from '../ui/DropdownMenu.svelte';
 import DropdownMenuItem from '../ui/DropDownMenuItem.svelte';
+import Tooltip from '/@/lib/ui/Tooltip.svelte';
 </script>
 
 <SettingsPage title="Authentication">
@@ -19,9 +20,13 @@ import DropdownMenuItem from '../ui/DropDownMenuItem.svelte';
       hidden="{$authenticationProviders.length > 0}" />
     {#each $authenticationProviders as provider}
       <!-- Registered Authentication Provider row start -->
-      <div class="flex flex-col w-full">
-        <div class="flex rounded-md border-0" style="background-color: rgb(39 39 42 / var(--tw-bg-opacity))">
+      <div class="flex flex-col w-full mb-5">
+        <div
+          class="flex rounded-md border-0 justify-between"
+          style="background-color: rgb(39 39 42 / var(--tw-bg-opacity))">
+          <!-- Icon + status -->
           <div class="ml-4 flex items-center">
+            <!-- Icon -->
             <div class="flex">
               {#if provider?.images?.icon}
                 {#if typeof provider.images.icon === 'string'}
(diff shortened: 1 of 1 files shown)


dgolovin (approved):
LGTM

deboer-tim:
👍🏼 to change. 2 nits though:
- Line 26 - since we're changing it, can you replace with the class instead of style?
- The drop down menu looks out of place now (doesn't look like a button next to the accounts) and 'moves' the accounts left when it exists. For the first it might help to give it a bg, but that might not be necessary if you (e.g.) put the accounts in a middle column so that they're always aligned.

We should really switch this to use IconImage too, but that's a different PR.

axel7083:
> * Line 26 - since we're changing it, can you replace with the class instead of style?

Can you make a commit on the line, I do not get the issue on line 26 ?

axel7083:
> * The drop down menu looks out of place now (doesn't look like a button next to the accounts) and 'moves' the accounts left when it exists. For the first it might help to give it a bg, but that might not be necessary if you (e.g.) put the accounts in a middle column so that they're always aligned.

@deboer-tim I updated the mockups and the code

packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.svelte
```diff
@@ -45,48 +50,54 @@ import DropdownMenuItem from '../ui/DropDownMenuItem.svelte';
...
+              <!-- Authentication Provider Session label start -->
+              <div class="pt-3 pb-3 text-sm">
+                {#each provider.accounts as account}
+                  <div class="flex flex-row">
+                    <div class="flex items-center w-full">
+                      <div class="text-xs bg-charcoal-800 p-2 rounded-lg mt-1">
+                        <span class="my-auto font-bold col-span-1 text-right">
+                          {account.label}
+                        </span>
+                        <Tooltip tip="Sign out of {account.label}" left>
```
deboer-tim, line 85: You can just give the button a title (which is used as a tooltip) instead of using a Tooltip.

deboer-tim (approved):
Thanks for the changes. One minor nit but LGTM.

packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.svelte
axel7083: image

The button tooltip is a bit _unstyled_ (ugly), do we want to use that ?

deboer-tim: We don't really have a standard, and have used button/other tooltip for most simple cases and Tooltip for cases where it isn't supported or needs more styling. I'm fine leaving it, let's merge.

PR #5445, opened 2024-01-05
feat: add confirmation dialog when deleting objects

### LINKED ISSUE #5444: Confirmation when deleting for all critical components

This was brought up here: https://github.com/containers/podman-desktop/issues/4993#issuecomment-1873988206

Similar to other tools, we should be confirming that you really want to delete something when pressing the trash-can option.

Especially since we now have Deployments, Services, etc. which could be critical parts of ones infrastructure.

Clear modal / confirmation when deleting a pod / service / etc.

_No response_

### Additional context

_No response_

cdrage:
feat: add confirmation dialog when deleting objects

### What does this PR do?

Adds confirmation dialogs to all critical components of Podman Desktop
where you are deleting a container, pod, deployment, service, etc.

This makes sure you do not accidently delete something (such as a
deployment) that could be critical infrastructure.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/6422176/b094db26-d9d6-4d8d-8891-16a05c39b4a8

Closes https://github.com/containers/podman-desktop/issues/5444

### How to test this PR?

Try to delete anything (volume, pod, container, etc.) and you will now
see a dialog.

### DIFF
--- a/packages/renderer/src/lib/compose/ComposeActions.spec.ts
+++ b/packages/renderer/src/lib/compose/ComposeActions.spec.ts
@@ -18,7 +18,7 @@
 
 import '@testing-library/jest-dom/vitest';
 import { test, expect, vi, beforeEach, afterEach } from 'vitest';
-import { fireEvent, render, screen } from '@testing-library/svelte';
+import { fireEvent, render, screen, waitFor } from '@testing-library/svelte';
 import ComposeActions from './ComposeActions.svelte';
 import type { ComposeInfoUI } from './ComposeInfoUI';
 import type { ContainerInfoUI } from '../container/ContainerInfoUI';
@@ -41,8 +41,10 @@ const compose: ComposeInfoUI = {
 
 const getContributedMenusMock = vi.fn();
 const updateMock = vi.fn();
+const showMessageBoxMock = vi.fn();
 
 beforeEach(() => {
+  (window as any).showMessageBox = showMessageBoxMock;
   (window as any).startContainersByLabel = vi.fn();
   (window as any).stopContainersByLabel = vi.fn();
   (window as any).restartContainersByLabel = vi.fn();
@@ -103,12 +105,17 @@ test('Expect no error and status restarting compose', async () => {
 });
 
 test('Expect no error and status deleting compose', async () => {
+  // Mock the showMessageBox to return 0 (yes)
+  showMessageBoxMock.mockResolvedValue({ response: 0 });
   const { component } = render(ComposeActions, { compose });
   component.$on('update', updateMock);
 
   // click on delete button
   const deleteButton = screen.getByRole('button', { name: 'Delete Compose' });
   await fireEvent.click(deleteButton);
+
+  // Wait for confirmation modal to disappear after clicking on delete
+  await waitFor(() => expect(screen.queryByRole('dialog')).not.toBeInTheDocument());
 
   expect(compose.status).toEqual('DELETING');
   expect(compose.actionError).toEqual('');
--- a/packages/renderer/src/lib/dialogs/MessageBox.svelte
+++ b/packages/renderer/src/lib/dialogs/MessageBox.svelte
@@ -139,6 +139,8 @@ function getButtonType(b: boolean): ButtonType {
   <div class="fixed top-0 left-0 right-0 bottom-0 bg-black bg-opacity-60 bg-blend-multiply h-full grid z-50">
     <div
       class="flex flex-col place-self-center w-[550px] rounded-xl bg-charcoal-800 shadow-xl shadow-black"
+      role="dialog"
+      aria-labelledby="{title}"
       bind:this="{messageBox}">
       <div class="flex items-center justify-between pl-4 pr-3 py-3 space-x-2 text-gray-400">
         {#if type === 'error'}
@@ -153,7 +155,7 @@ function getButtonType(b: boolean): ButtonType {
         {:else if type === 'question'}
           <Fa class="h-4 w-4" icon="{faCircleQuestion}" />
         {/if}
-        <h1 class="grow text-lg font-bold capitalize">{title}</h1>
+        <h1 aria-label="{title}" class="grow text-lg font-bold capitalize">{title}</h1>
 
         <button
           class="p-2 hover:text-gray-300 hover:bg-charcoal-500 rounded-full cursor-pointer"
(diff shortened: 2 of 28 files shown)


cdrage:
TODO:

- [x] Fix e2e tests (modal comes up to confirm deletion, need to click on it).

deboer-tim (commented):
I've been thinking about this for a while, and agree we need to do it. +1 to approach with ListItemButtonIcon where we can just add a confirmation attribute.

I think we've already heard from a couple users that don't always want this on for every object. I'd still say we do this first and be ready to have a discussion about a 'do not show again' and whether preference should be per object, per environment, etc. This approach would work, could always do something like `export let confirm: boolean | string` (preference key).

I don't love how the tests change, especially ones less related to the actions like *Details and *ListCompose. Instead of having to hook all message boxes to respond with 0 it feels like the tests should just say 'Say yes to confirmation dialog(s)'. For now maybe it would be better to at least put the message box mocking in a test helper util?

cdrage:
> I've been thinking about this for a while, and agree we need to do it. +1 to approach with ListItemButtonIcon where we can just add a confirmation attribute.
> 
> I think we've already heard from a couple users that don't always want this on for every object. I'd still say we do this first and be ready to have a discussion about a 'do not show again' and whether preference should be per object, per environment, etc. This approach would work, could always do something like `export let confirm: boolean | string` (preference key).
> 
> I don't love how the tests change, especially ones less related to the actions like *Details and *ListCompose. Instead of having to hook all message boxes to respond with 0 it feels like the tests should just say 'Say yes to confirmation dialog(s)'. For now maybe it would be better to at least put the message box mocking in a test helper util?

I agree!

Do we happen to already have a utility folder for svelte front-end-related tests? I wasn't able to find a helper file / directory that we could use. If not, I can throw it in a new folder. 

@cbr7 @odockal I'm unsure what I'm doing wrong, but I'm trying to create a helper function to click the confirmation dialog for the e2e tests, if you're able to point out a fix that'd be awesome

packages/renderer/src/lib/dialogs/MessageBox.svelte
```diff
@@ -148,7 +150,7 @@ function getButtonType(b: boolean): ButtonType {
         {:else if type === 'question'}
           <Fa class="h-4 w-4" icon="{faCircleQuestion}" />
         {/if}
-        <h1 class="grow text-lg font-bold capitalize">{title}</h1>
+        <h1 id="{title}" class="grow text-lg font-bold capitalize">{title}</h1>
```
deboer-tim: What is this used for? Couldn't find a reference, maybe can be removed now.

packages/renderer/src/lib/compose/ComposeActions.spec.ts
```diff
@@ -103,13 +105,18 @@ test('Expect no error and status restarting compose', async () => {
 });
 
 test('Expect no error and status deleting compose', async () => {
+  // Mock the showMessageBox to return 0 (yes)
+  showMessageBoxMock.mockResolvedValue({ response: 0 });
```
deboer-tim, line 109: I really like what handleConfirmationDialog() does for the e2e tests, and that's kind of what I was hoping for in the regular tests, i.e. is there any way to change the pattern of:
   showMessageBoxMock/`await fireEvent.click(deleteButton)`
to something like::
   `await utility.clickButtonAndConfirm(deleteButton)`
that would do the click and respond yes (or mock/disable the confirmation dialog completely)? (and no, I don't know of/think we have a common place for things like this yet)

deboer-tim (commented):
Tried this on volumes and services, it is working great. 👍🏼 Just one comment inline.

Nit: it says 'Do you want to delete volume", feels like it should say "Do you want to delete the volume" but this is not required.

packages/renderer/src/lib/dialogs/MessageBox.svelte
cdrage: Used for e2e tests to run correctly :)

packages/renderer/src/lib/compose/ComposeActions.spec.ts
cdrage: Due to https://vitest.dev/guide/mocking.html#mocking-pitfalls

The 'screen' / rendering portion stays within the test() function, so we're unable to move this to a separate file for mocking / using screen to do clicking.

I've tried multiple ways to implement this into a separate function, but the screen does not "transfer" over, even when passing in `screen: any` to the function.

```ts
export async function clickAndConfirm(clickAction: () => void) {
  const showMessageBoxMock = vi.fn();
  (window as any).showMessageBox = showMessageBoxMock;
  showMessageBoxMock.mockResolvedValue({ result: 1 });
  clickAction();
  await waitFor(() => expect(screen.queryByRole('dialog')).not.toBeInTheDocument());
}
```

cdrage:
> Tried this on volumes and services, it is working great. 👍🏼 Just one comment inline.
> 
> Nit: it says 'Do you want to delete volume", feels like it should say "Do you want to delete the volume" but this is not required.

We could update that in a separate PR as we'd have to update the titles in most places: https://github.com/containers/podman-desktop/pull/5445/files#diff-1fb3450fd535b8bfd2e357f46b03ba71a318831bcb449a065d509b27ebd53c05R16

odockal:
@cdrage I am afraid that you have overwritten commit that fixed e2e tests, so it fails again.

cdrage:
> @cdrage I am afraid that you have overwritten commit that fixed e2e tests, so it fails again.

Ah sorry, do you happen to have your commit still available? For some reason with the rebase it did overwrite.

odockal:
@cdrage Seems that we are all set for the merge?

cdrage:
Yup 💯 just need a re-review!

odockal:
I have had a chance to test the PR locally on Linux due to e2e tests updates, so it is +1 from me as well.

odockal (approved):

packages/renderer/src/lib/compose/ComposeActions.spec.ts
deboer-tim: ok, thanks for investigating.

deboer-tim (approved):
LGTM

packages/renderer/src/lib/dialogs/MessageBox.svelte
benoitf: shouldn't be an id but an aria-label

benoitf: then `getByRole('heading', { name: '....' }`

benoitf (changes requested):

cdrage:
Ready for a re-review 👍 

Got e2e tests working thanks to help from @odockal was having difficulty why aria-label wasn't working.. then realized it was getting it from the wrong dialog.

benoitf (approved):

PR #5448, opened 2024-01-07
fix: service type colors

### LINKED ISSUE #5443: Kubernetes service page: NodePort should have green not orange indicator

The colouring for NodePort is Orange when its should be green to indicate that there is nothing wrong:
Screenshot 2024-01-05 at 2 07 07 PM

This would follow other conventions such as how on deployments "Available" is green, while "Progressing" is blue, etc.

Screenshot 2024-01-05 at 2 11 53 PM

Green instead of orange / other way of indicating that it's available / working correctly.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

The original design #4576 had a few different colors for port types, this switches to use the closest palette colors I could find. Green/amber are not used, so hopefully this is enough to show them as different types without a started/stopped association.

### Screenshot / video of UI

Screenshot 2024-01-07 at 1 32 29 PM

Fixes #5443.

### How to test this PR?

`yarn test:renderer`, check if colors are distinct enough and avoid

### DIFF
--- a/packages/renderer/src/lib/service/ServiceColumnType.spec.ts
+++ b/packages/renderer/src/lib/service/ServiceColumnType.spec.ts
@@ -42,7 +42,7 @@ test('Expect basic column styling', async () => {
 
   const dot = text.parentElement?.children[0].children[0];
   expect(dot).toBeInTheDocument();
-  expect(dot).toHaveClass('bg-gray-900');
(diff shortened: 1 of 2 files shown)


deboer-tim:
@cdrage the original design had colors for each type and avoided green/amber, let me know if this works for you.

benoitf (approved):

PR #5458, opened 2024-01-08
chore(tests): play kubernetes yaml e2e test

### LINKED ISSUE #5440: Implement E2E  test case using Playwright to verify Play Kubernetes Yaml scenario
On containers or pods page, there is an option to `Play Kubernetes Yaml` which can bring for example pods to life using a yaml file.
Is is a simply form/page that we need to test. It takes a yaml file path, let use pick up runtime/engine and then "build" it. The output could be bunch of containers in a pod. 

Screenshot_20240105_161214

We are restricted with a native OS file manager dialog that will be opened when browse button is pressed. There are workarounds, though. (This is not the best option), but we could do it for now and fill an issue to make the field an editable textbox. 
Workaround Snippet:
```
const buildContextDirectory = labelParent.getByRole('textbox');
await buildContextDirectory.evaluate(node => node.removeAttribute('readonly'));
```

Example yamls: https://github.com/odockal/podman-desktop-qe/blob/main/examples/manifests/primary-podify-demo.yaml
Resources: https://www.redhat.com/sysadmin/podman-play-kube-updates

cbr7:
Signed-off-by: Vladimir Lazar <vlazar@redhat.com>

### What does this PR do?

Creates e2e test for Play Kubernets Yaml functionality

Closes #5440


tests/resources/primary-podify-demo.yaml
```diff
@@ -0,0 +1,38 @@
+# Save the output of this file and use kubectl create -f to import
```
odockal, line 1: Probably a license might be here? But I am not sure. It is generated from podman-desktop using some podman feature. it is a question then...

tests/src/model/pages/play-kube-yaml-page.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.playButton = page.getByRole('button', { name: 'Play' });
+    this.doneButton = page.getByRole('button', { name: 'Done' });
+  }
+
+  async playYaml(pathToYaml: string): Promise<PodsPage> {
+    if (!pathToYaml) {
+      throw Error(`Path to Yaml file is incorrect or not provided!`);
+    }
+
+    await this.yamlPathInput.evaluate(node => node.removeAttribute('readonly'));
```
odockal, line 42: Please provide some comment and todo part as we know this should change after a proper input handling update happens. Also, please fill an issue for this in the repo to have a reference (if it does not exist).

tests/src/pod-smoke.spec.ts
```diff
@@ -206,5 +207,22 @@ describe.skipIf(process.env.GITHUB_ACTIONS && process.env.RUNNER_OS === 'Linux')
       playExpect(await podsPage.podExists(podToRun)).toBeFalsy();
       await pdRunner.screenshot('pods-pod-deleted.png');
     });
+
+    test('Playing yaml', async () => {
+      const navigationBar = new NavigationBar(page);
```
odockal: I am not sure if we want to have this test as it is part of pod-smoke. First problem could be the yaml file used, it uses the same images (not a problem) and starts up the same app (name of the pod/container conflict, port conflict ?). 
We definitely want to test this kube. manifest example, just probably in other test class/file. We can keep this test here as well, just please use some different image/app to test - something more lightweight.

odockal (changes requested):

tests/src/pod-smoke.spec.ts
cbr7: @odockal I could move the test into a new test file as such it would run independently of any other test, would that be better?

tests/resources/primary-podify-demo.yaml
cbr7: Which or from where would I get that license?

odockal: basically any file in this codebase should have it. But that is more of a question I am not sure about the answer. :)
We can probably keep what we got now there.

tests/src/pod-smoke.spec.ts
odockal: Yes that can work! Except I am curious about if it will be working on ubuntu runner on GHA since the pod-smoke does not run properly due to a bug in kernel. Please try it then. We could also check for the image pulled during pod creation as part of that test.

tests/resources/primary-podify-demo.yaml
cbr7: ok, left that part unchanged.

tests/src/pod-smoke.spec.ts
cbr7: moved the test to a new test file

tests/src/model/pages/play-kube-yaml-page.ts
cbr7: Created issue https://github.com/containers/podman-desktop/issues/5479 and added TODO to code for when the change is made.

tests/src/model/pages/play-kube-yaml-page.ts
```diff
@@ -0,0 +1,51 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
odockal: New year :)

tests/resources/primary-podify-demo.yaml
```diff
@@ -0,0 +1,27 @@
...
+      value: d6ee99a0905d
+    - name: TERM
+      value: xterm
+    - name: HOME
+      value: /home/redis
+    image: quay.io/podman-desktop-demo/podify-demo-backend:v1
+    name: backend-podified
+    ports:
+    - containerPort: 5000
+      hostPort: 5000
```
odockal: It works locally for you I guess, but if someone else will run it on mac, it will fail. Default behavior is that mac has 5000 port dedicated to some system application, we must take that into account. Please change host port to something else.

tests/src/yaml-smoke.spec.ts
```diff
@@ -0,0 +1,77 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
odockal: happy new year :)

tests/resources/primary-podify-demo.yaml
```diff
@@ -0,0 +1,27 @@
+# Save the output of this file and use kubectl create -f to import
+# it into Kubernetes.
+#
+# Created with podman-4.8.1
```
odockal, line 4: Since you have moved the testing part into single file, I would then be for a testing full podify example. backend+frontend images as it was before.
Maybe it was not clear, but If we would just added play yaml test within pod-smoke, then the example could be something else then primary-podify-demo, like alpine pod, or something lightweight. 
since we are adding whole new test, let's test the same thing as in pod-smoke. wdyt?

tests/src/yaml-smoke.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+beforeEach<RunnerTestContext>(async ctx => {
+  ctx.pdRunner = pdRunner;
+});
+
+afterAll(async () => {
+  await pdRunner.close();
+});
+
+describe.skipIf(process.env.GITHUB_ACTIONS && process.env.RUNNER_OS === 'Linux')(
+  'Play yaml file to pull images and create pods',
```
odockal: Please also state a name for pod app we are creating -> ie. testing primary-podify-demo. It is part of demos presented and it is in testing template as well.

tests/src/yaml-smoke.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      await playExpect(podsPage.heading).toBeVisible();
+
+      const playYamlPage = await podsPage.openPlayKubeYaml();
+      await playExpect(playYamlPage.heading).toBeVisible();
+
+      const yamlFilePath = path.resolve(__dirname, '..', 'resources', 'primary-podify-demo.yaml');
+      podsPage = await playYamlPage.playYaml(yamlFilePath);
+      await playExpect(podsPage.heading).toBeVisible();
+
+      playExpect(await podsPage.podExists('podify-demo-pod')).toBeTruthy();
```
odockal: here I would include test method:
so it would be like this;
```
test('Play yaml'){}
test('Primary-podify-pod is created') {}
test('Pod's images are present/pulled'){}
test('can delete pods and images'){}

File: tests/src/yaml-smoke.spec.ts [hunk 10]
```diff
@@ -0,0 +1,77 @@
...
+
+  await new WelcomePage(page).handleWelcomePage(true);
+});
+
+beforeEach<RunnerTestContext>(async ctx => {
+  ctx.pdRunner = pdRunner;
+});
+
+afterAll(async () => {
+  await pdRunner.close();
```
odockal, line 52: I would also include cleanup method call. take a look at `podm-smoke`

odockal (changes requested):

tests/resources/primary-podify-demo.yaml
cbr7: @odockal wait... so you want me to add back the entire yaml file as it was at the beginning?

tests/src/yaml-smoke.spec.ts
cbr7: I'm not sure I understand what you mean, do you want to split the test in four parts?

tests/src/yaml-smoke.spec.ts
cbr7: But it does clean up, it just does it inside the test, not in the afterAll hook

odockal: What if the test fails? then it will be kept there for next test.

odockal: Cleanups are designed to not fail if pod/image/container does not exist. So nothing happen if test passes.

tests/src/yaml-smoke.spec.ts
odockal: Yes. It is mainly for readability. Without change -> if deleting pods fails -> on first sight it will show that whole test failed, and you need to go deep to find out where. 
We can also divide the play yaml scenario into multiple logical parts. First you go through the form "Play Kube yaml" and verifies you can pass the file, let it process the file and got Done button. But it also pulls the images and creates and start pods and also containers are shown. So it implies we should also check these and the are individuals parts, so it would be nice to have them "divided".

tests/resources/primary-podify-demo.yaml
odockal: Yes.

tests/src/yaml-smoke.spec.ts
cbr7: But if we do that then we're just testing the delete pods and images functionality and we're already doing that in another test. I'm not sure I see the point of such duplication. The way I understand it is that creating the pods and pulling the images are part of the "play yaml" functionality so a test should check that, but if you want to split it down further you are just retesting functionality that is already tested in another test. Deleting a pod from play yaml is the same as deleting a pod that was created in some other way.

Ultimately if we go with such atomization, we're going to end up with many very similar/identical tests in different test files.

tests/resources/primary-podify-demo.yaml
cbr7: got it, will add it back

tests/src/yaml-smoke.spec.ts
odockal: I agree with deleting part. It might be also part of clean up in after hook.
But as for checking images, pods and containers, I would stick to what was said. Since we are passing just a file with all inputs and it behaves as a blackbox with how is it processing the kube manifest (it just calls alternative of `podman play kube`). We need to make sure that correct images were pulled, containers in pod are named correctly (pod as well).

tests/src/model/pages/play-kube-yaml-page.ts
cbr7: done

tests/resources/primary-podify-demo.yaml
cbr7: switched to port 5007, hopefully that is fine

tests/src/yaml-smoke.spec.ts
cbr7: done

tests/resources/primary-podify-demo.yaml
cbr7: done

tests/src/yaml-smoke.spec.ts
cbr7: I've split it into three tests:
1. play yaml
2. validate pod and delete pod
3. validate images and delete images

tests/src/yaml-smoke.spec.ts
cbr7: added the relevant method calls into the afterAll hook.

tests/src/yaml-smoke.spec.ts
cbr7: done

odockal (approved):

PR #5494, opened 2024-01-10
feat: image status & deletion

### LINKED ISSUE #5474: No feedback when deleting an image

When you delete an image, there's no feedback. Fine for small images which delete quickly, but when I delete a 1Gb image you can't tell if it's still trying to do it.

Status icon should change to a spinner (deleting state) like on other pages. Should work in both the list and the details pages.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Images had an inUse flag that we used to determine which status icon to show or whether the image is delete-able. If you delete the image and there is any delay, there is no feedback in the UI and you can delete it again (or accidentally click on the next image delete if the first one disappears at the wrong time).

This changes images to have a status the same as most other types (e.g. pods, containers, deployments), so once the user clicks to delete it is marked for deletion (spinner in UI) and you can't delete again.

### Screenshot / video of UI

Self-explanatory, now same as other lists.

Fixes #5474.

### How to test this PR?

Delete an image!

### DIFF
--- a/packages/renderer/src/lib/image/ImageActions.spec.ts
+++ b/packages/renderer/src/lib/image/ImageActions.spec.ts
@@ -54,6 +54,7 @@ test('Expect showMessageBox to be called when error occurs', async () => {
     onRenameImage: vi.fn(),
     image: {
       name: 'dummy',
+      status: 'UNUSED',
     } as unknown as ImageInfoUI,
   });
   const button = screen.getByTitle('Delete Image');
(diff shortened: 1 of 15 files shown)


benoitf (commented):
I think there should be a commit around moving inUse to status USED/UNUSED and another one being about the image deletion
b/c here I think it's hard to see what is really about deletion and about status change

deboer-tim:
> I think there should be a commit around moving inUse to status USED/UNUSED and another one being about the image deletion b/c here I think it's hard to see what is really about deletion and about status change

Ok 👍🏼. Retiring this in favour of #5528 + a future PR.

PR #5503, opened 2024-01-11
feat: add a way to choose platform when building image

### LINKED ISSUE #5492: allow to provide platform parameter when doing builds

When building image it is now providing the default platform (the one used by the podman machine)

we should be able to select the platform

allow to select the platform

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Allow to build for a different platform when building image within Podman Desktop

### Screenshot / video of UI

image

fixes https://github.com/containers/podman-desktop/issues/5492

### How to test this PR?

unit tests added

you can also go to build panel and choose a platform

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
@@ -70,6 +70,7 @@ onMount(() => {
   class="rounded-md p-2 min-w-32 w-32 cursor-pointer hover:bg-charcoal-700 {checked
     ? 'border-dustypurple-700'
     : 'border-gray-700'} border-2 flex flex-row"
+  aria-label="{value}"
   on:click|preventDefault="{() => handleClick()}">
   <div>
     {#if !additionalItem}
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
@@ -11,11 +11,10 @@ export let isDefault: boolean = false;
 export let checked: boolean = false;
 export let value: string = '';
 export let icon: any;
+let iconType: 'fontAwesome' | 'unknown' | undefined = undefined;
 
 export let additionalItem: boolean = false;
 let additionalValue: string = '';
-
-export let fontAwesomeIcon = false;
 
 let displayValueFieldInput = false;
 
@@ -55,6 +54,12 @@ function handleClick() {
 }
 
 onMount(() => {
+  if (icon?.prefix?.startsWith('fa')) {
+    iconType = 'fontAwesome';
+  } else {
+    iconType = 'unknown';
+  }
+
   if (isDefault) {
     dispatch('card', { mode: 'add', value: value });
   }
@@ -76,9 +81,9 @@ onMount(() => {
   <div class="mr-2 text-gray-700">
     <div class="flex flex-row">
       <div class="ml-1">
-        {#if fontAwesomeIcon}
+        {#if iconType === 'fontAwesome'}
           <Fa class="text-gray-700 cursor-pointer" icon="{icon}" size="24" />
-        {:else}
+        {:else if iconType === 'unknown'}
           <svelte:component this="{icon}" class="text-gray-700 cursor-pointer" size="24" />
         {/if}
       </div>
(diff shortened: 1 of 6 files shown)


packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
```diff
@@ -0,0 +1,107 @@
...
+    {#if !additionalItem}
+      <Checkbox bind:checked="{checked}" on:click="{() => handleClick()}" />
+    {:else}
+      <div>&nbsp;</div>
+    {/if}
+  </div>
+  <div class="mr-2 text-gray-700">
+    <div class="flex flex-row">
+      <div class="ml-1">
+        {#if fontAwesomeIcon}
```
deboer-tim: Could just use 'startsWith('fa')' since we essentially do this elsewhere.

deboer-tim (approved):
Code LGTM, and nice improvement to have. My only concern is that the architecture cards are large and very noticeable - i.e. drawing your eyes and putting focus on an option that we only expect to be used in some cases. I think I'd prefer if they were closer to badges (e.g. maybe just icon + amd) and you used tooltips if you didn't recognize it. But, we could always do this later.

benoitf:
cc @mairin for design review

packages/renderer/src/lib/image/BuildImageFromContainerfileCards.svelte
```diff
@@ -0,0 +1,190 @@
...
+    value: item.value,
+    icon: faLayerGroup,
+    fontAwesomeIcon: true,
+  };
+  advancedCards.push(card);
+  advancedCards = advancedCards;
+}
+</script>
+
+<div class="flex flex-col">
```
odockal: I would consider also adding ARIA functionality from the start of this PR. 
adding `role="region" aria-label="Build Platform Options"` or something similar would be nice.

packages/renderer/src/lib/image/BuildImageFromContainerfileCards.svelte
```diff
@@ -0,0 +1,190 @@
...
+  <div class="flex flex-row gap-x-4 gap-y-4 flex-wrap">
+    {#each sortedCards as card}
+      <BuildImageFromContainerfileCard
+        title="{card.title}"
+        isDefault="{card.isDefault}"
+        checked="{card.checked}"
+        badge="{card.badge}"
+        value="{card.value}"
+        icon="{card.icon}"
```
odockal, line 137: `aria-label="{card.value}"`

packages/renderer/src/lib/image/BuildImageFromContainerfileCards.svelte
```diff
@@ -0,0 +1,190 @@
...
+    <div class="flex flex-col pt-4">
+      <div class="flex flex-row gap-x-4 flex-wrap gap-y-4">
+        {#each advancedCards as card}
+          <BuildImageFromContainerfileCard
+            title="{card.title}"
+            isDefault="{card.isDefault}"
+            checked="{card.checked}"
+            badge="{card.badge}"
+            value="{card.value}"
+            icon="{card.icon}"
```
odockal, line 160: `aria-label="{card.value}"`

packages/renderer/src/lib/image/BuildImageFromContainerfileCards.svelte
```diff
@@ -0,0 +1,190 @@
...
+            fontAwesomeIcon="{card.fontAwesomeIcon}"
+            on:card="{item => handleCard(item)}" />
+        {/each}
+        <BuildImageFromContainerfileCard
+          title="New platform"
+          isDefault="{false}"
```
odockal, line 165: `aria-label="New Platform"`

odockal (changes requested):

benoitf:
@odockal while I understand the first aria stuff around `role="region" aria-label="Build Platform Options"` I don't understand the other one about aria-labels on Cards object.

The card element already have a div with the value

packages/renderer/src/lib/image/BuildImageFromContainerfileCards.svelte
benoitf: added

packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
benoitf: I changed it.
I think in other places it's checking prefix being exactly 'fa' and for Linux icon as a brand it was 'fab'

odockal (approved):

benoitf:
@deboer-tim discussed with @mairin and she'll provide a new UI based on your feedback

PR #5528, opened 2024-01-12
chore: image status

deboer-tim:
### What does this PR do?

Images have an inUse flag that we use to determine which status icon to show or whether the image is delete-able. However, that means we cannot have other states like deleting or new.

This changes images to have a status the same as most other types (e.g. pods, containers, deployments). This will allow future PRs to add additional states (e.g. deleting) and UI updates, e.g. so that you see a spinner if an image takes a second to delete.

### Screenshot / video of UI

N/A

Step toward #5474.

### How to test this PR?

`yarn test:renderer`


packages/renderer/src/lib/image/ImageInfoUI.ts
```diff
@@ -31,5 +31,5 @@ export interface ImageInfoUI {
   // no tag, we encode <none>
   base64RepoTag: string;
   selected: boolean;
-  inUse: boolean;
+  status: string;
```
benoitf: should the status be `'USED' | 'UNUSED'` rather than string ?

deboer-tim: Good idea (we should probably do this with other types too).

Updated, and force-pushed to rebase.

jeffmaury: I don't see the type change, did I miss something ?

benoitf: I think first 'force-push' had the right fix but then the second one replaced the fix

fist one: https://github.com/containers/podman-desktop/commit/92b339f5635ee540646289f3dbacbd4db9d6b4b8

deboer-tim: Doh, must have lost it in the rebase. Most of the files now have merge conflicts too, I'll do another rebase and make sure I pick this up.

deboer-tim: Rebased and force-pushed, ready for (re)review.

benoitf (approved):

benoitf (approved):

PR #5533, opened 2024-01-15
feat: adding BuildOption

### LINKED ISSUE #5507: buildImage need to have an `BuildOption` object instead of many arguments
### Bug description

The signature of `containerEngine.buildImage` should not be using many arguments, but have some kind of `BuildOption` in the same way that `createContainer` with its `ContainerCreateOptions`.

### Operating system

All

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Fix https://github.com/containers/podman-desktop/issues/5507

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1919,6 +1919,14 @@ declare module '@podman-desktop/api' {
     id: string;
   }
 
+  export interface BuildImageOptions {
+    containerFile?: string;
+    target?: string;
+    platform?: string;
+    provider?: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection;
+    abortController?: AbortController;
+  }
+
   export interface NetworkCreateOptions {
     Name: string;
   }
@@ -1953,13 +1961,9 @@ declare module '@podman-desktop/api' {
     export function stopContainer(engineId: string, id: string): Promise<void>;
     export function deleteContainer(engineId: string, id: string): Promise<void>;
     export function buildImage(
-      containerBuildContextDirectory: string,
-      relativeContainerfilePath: string,
-      imageName: string,
-      platform: string,
-      selectedProvider: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection,
+      context: string,
       eventCollect: (eventName: 'stream' | 'error' | 'finish', data: string) => void,
-      abortController?: AbortController,
+      option?: BuildImageOptions,
     );
     export function saveImage(engineId: string, id: string, filename: string): Promise<void>;
     export function listImages(): Promise<ImageInfo[]>;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1963,7 +1963,7 @@ declare module '@podman-desktop/api' {
     export function buildImage(
       context: string,
       eventCollect: (eventName: 'stream' | 'error' | 'finish', data: string) => void,
-      option?: BuildImageOptions,
+      options?: BuildImageOptions,
     );
     export function saveImage(engineId: string, id: string, filename: string): Promise<void>;
     export function listImages(): Promise<ImageInfo[]>;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1920,10 +1920,15 @@ declare module '@podman-desktop/api' {
   }
 
   export interface BuildImageOptions {
+    // Specifies a Containerfile which contains instructions for building the image
     containerFile?: string;
-    target?: string;
+    // Specifies the name which is assigned to the resulting image if the build process completes successfully.
+    tag?: string;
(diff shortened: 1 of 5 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -1953,13 +1961,9 @@ declare module '@podman-desktop/api' {
...
     export function buildImage(
-      containerBuildContextDirectory: string,
-      relativeContainerfilePath: string,
-      imageName: string,
-      platform: string,
-      selectedProvider: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection,
+      context: string,
       eventCollect: (eventName: 'stream' | 'error' | 'finish', data: string) => void,
-      abortController?: AbortController,
+      option?: BuildImageOptions,
```
benoitf: I have a remark

you designed option being an optional parameter but some of the parameters are mandatory (like the provider)

so probably the option should only contain the optional parameters (then we can omit it) or it's a mandatory parameter

```suggestion
      options?: BuildImageOptions,
```

axel7083: > you designed option being an optional parameter but some of the parameters are mandatory (like the provider)

Why does the provider should be mandatory ? 

Inside the container registry we can get the default one following our own logic (in favor of podman) `getFirstRunningConnection`. 

I feel like it is easier to make is optional, so extension can set it only if they have a requirements on a specific provider. Otherwise most of extension will simply get all providers and take the first one (as our implementation with @lstocchi ).

I added the logic to get use the getFirstRunningConnection method if provider is undefined (I also added a test to ensure it is properly called)

> so probably the option should only contain the optional parameters (then we can omit it) or it's a mandatory parameter

I agree that the option should only be the optional elements, that's why the first arguments in the buildImage is the context dir, from a user point of view the "podman build ." is the simplest method it can use.

packages/main/src/plugin/container-registry.ts
```diff
@@ -1946,17 +1946,23 @@ export class ContainerProviderRegistry {
...
     try {
-      // grab all connections
-      const matchingContainerProviderApi = this.getMatchingEngineFromConnection(selectedProvider);
+      let matchingContainerProviderApi: Dockerode;
+      if (selectedProvider !== undefined) {
+        // grab all connections
+        matchingContainerProviderApi = this.getMatchingEngineFromConnection(selectedProvider);
+      } else {
+        // Get the first running connection (preference for podman)
+        matchingContainerProviderApi = this.getFirstRunningConnection()[1];
```
benoitf, line 1964: I don't see how this line can work properly

packages/extension-api/src/extension-api.d.ts
```diff
@@ -1953,13 +1961,9 @@ declare module '@podman-desktop/api' {
...
     export function buildImage(
-      containerBuildContextDirectory: string,
-      relativeContainerfilePath: string,
-      imageName: string,
-      platform: string,
-      selectedProvider: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection,
+      context: string,
       eventCollect: (eventName: 'stream' | 'error' | 'finish', data: string) => void,
-      abortController?: AbortController,
+      options?: BuildImageOptions,
```
benoitf: it requires at least a comment that if provider is not provided, it will try with the first running provider and with a preference of a provider using podman engine type

packages/extension-api/src/extension-api.d.ts
```diff
@@ -1953,13 +1961,9 @@ declare module '@podman-desktop/api' {
     export function stopContainer(engineId: string, id: string): Promise<void>;
     export function deleteContainer(engineId: string, id: string): Promise<void>;
     export function buildImage(
-      containerBuildContextDirectory: string,
-      relativeContainerfilePath: string,
-      imageName: string,
-      platform: string,
-      selectedProvider: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection,
+      context: string,
```
benoitf, line 1970: IDK why rename `containerBuildContextDirectory` to `context` ?

axel7083: We are calling the method `buildImage`, that seems redundant to say again build, and container, and the context is always a directory when building an image.

In the same way:
- `selectedProvider` is renamed to `provider`, since we are providing it, it seems logic it is the one we want to _select_
- `imageName` is renamed to `target`, to match `podman | docker build` arguments
- `relativeContainerfilePath` is renamed to `containerfile` because using `file` (--file) is not really explicit enough

axel7083: i've just notice that `--target` is not the equivalent `-t`. The shortcut `-t` is for tag. I will then rename the target property to `tag`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -1953,13 +1966,9 @@ declare module '@podman-desktop/api' {
...
```
benoitf, line 1970:
```suggestion
      // build context directory
      context: string,
```

benoitf (approved):

deboer-tim (approved):

Issue #3934, opened 2023-09-17
Cont'd of improving the accessibility of Resources Page


odockal:
### Is your enhancement related to a problem? Please describe

![Screenshot_20230917_172008](https://github.com/containers/podman-desktop/assets/19164299/a23622dd-025f-4dac-86d0-7e5f4399d9d1)

### Describe the solution you'd like

Include proper ARIA roles and attributes or label to better work with the content of the Resources Page.

PR #5546, opened 2024-01-15
chore(ui): add ARIA functionality into Podman Machine resources

odockal:
### What does this PR do?
It adds ARIA functionality in Resources Connection side in podman desktop.
#3934

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
@@ -153,9 +153,9 @@ function getLoggerHandler(
 
 {#if connectionStatus}
   {#if connection.lifecycleMethods && connection.lifecycleMethods.length > 0}
-    <div class="mt-2 relative" role="group" aria-label="Connection Actions">
+    <div class="mt-2 relative">
       <!-- TODO: see action available like machine infos -->
-      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto">
+      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto" role="group" aria-label="Connection Actions">
         {#if connection.lifecycleMethods.includes('start')}
           <div class="ml-2">
             <LoadingIconButton
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
@@ -153,9 +153,9 @@ function getLoggerHandler(
 
 {#if connectionStatus}
   {#if connection.lifecycleMethods && connection.lifecycleMethods.length > 0}
-    <div class="mt-2 relative">
+    <div class="mt-2 relative" role="group" aria-label="Connection Actions">
       <!-- TODO: see action available like machine infos -->
-      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto" role="group" aria-label="Connection Actions">
+      <div class="flex bg-charcoal-800 w-fit rounded-lg m-auto">
         {#if connection.lifecycleMethods.includes('start')}
           <div class="ml-2">
             <LoadingIconButton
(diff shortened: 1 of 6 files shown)


packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
```diff
@@ -153,9 +153,9 @@ function getLoggerHandler(
 
 {#if connectionStatus}
   {#if connection.lifecycleMethods && connection.lifecycleMethods.length > 0}
-    <div class="mt-2 relative">
+    <div class="mt-2 relative" role="group" aria-label="Connection Actions">
```
benoitf: it's not clear why we need to move some group definition at an upper level

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -441,17 +441,17 @@ function hasAnyConfiguration(provider: ProviderInfo) {
...
           <PreferencesConnectionsEmptyRendering
             message="{provider.emptyConnectionMarkdownDescription}"
             hidden="{provider.containerConnections.length > 0 || provider.kubernetesConnections.length > 0}" />
           {#each provider.containerConnections as container}
             {@const peerProperties = new PeerProperties()}
-            <div class="px-5 py-2 w-[240px]">
+            <div class="px-5 py-2 w-[240px]" role="region" aria-label="{container.name}">
               <div class="float-right">
-                <Tooltip tip="{provider.name} details" bottom>
+                <Tooltip tip="{provider.name} Details" bottom>
```
benoitf: AFAIK this change is not related to ARIA

benoitf (commented):
some changes in the label name (like details --> Details or the opposite, etc) are failing unit tests that rely on these current definition

Title might be confusing as it says 'add ARIA functionality' but it
- move some existing definition
- rename some labels

to me, adding ARIA should only include new stuff else it's a refactoring, a renaming etc

packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
odockal: So they would be on the same level.

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
odockal: that is right. will fix.

packages/renderer/src/lib/preferences/PreferencesConnectionActions.svelte
odockal: From testing point of view it does not matter after all. will revert.

odockal:
> to me, adding ARIA should only include new stuff else it's a refactoring, a renaming etc

For me it seems that former labels, especially aria-labels were not really intended to provide complex ARIA functionality, more just like randomly placed into UI into some parts, so I was thinking that I am actually adding it. How about `ARIA functionality expansion` or `Enhancement and refactoring of ARIA functionality` would that fits better what I am doing her and helps to identify kind of changes?

packages/renderer/src/lib/ui/ConnectionStatus.svelte
```diff
@@ -58,5 +58,5 @@ $: statusStyle = statusesStyle.get(status) || {
 };
 </script>
 
-<div aria-label="Connection Status Icon" class="{roundIconStyle} {statusStyle.bgColor}"></div>
-<span aria-label="Connection Status Label" class="{labelStyle} {statusStyle.txtColor}">{statusStyle.label}</span>
+<div aria-label="connection-status-icon" class="{roundIconStyle} {statusStyle.bgColor}"></div>
+<span aria-label="connection-status-label" class="{labelStyle} {statusStyle.txtColor}">{statusStyle.label}</span>
```
benoitf, line 62: to me labels are human labels not with `-` no ?

odockal: This is the state before, I am reverting it in order to do not mix adding - refactoring unless we agree on how to handle that. Otherwise, I would definitely go and fix such label as I encounter them when doing changes. Addressing them at once could be hard. So if it is ok to make also these changes even though they are on different place than is PR about, I would change it then.

odockal: This specific might require standalone pr, since also unit tests require updates...

benoitf (approved):

PR #5550, opened 2024-01-15
feat: merge extension preferences

### LINKED ISSUE #5549: nit: merge extensions preferences

When updating a different preference I noticed that we have both an 'Extensions' and 'Extensions Updates' preferences. When you click on the first nav item they both appear; when you click on the second only that section is visible.

Screenshot 2024-01-15 at 1 50 40 PM

These really seem like these should be on the same page, to reduce clutter and avoid the issue when clicking on the first.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Sets the same id and title on the extension update preferences as the activation pref. We could merge the two, but having two contributions with the same id doesn't seem to have any side effects and this keeps each bit of code separate.

Yes, this means that anyone who has set these two properties will have them reset when they update to this PR, but that seems inevitable when we fix this (the sooner the better) and doesn't seem worth additional code to automatically migrate the setting.

### Screenshot / video of UI

Before:

Screenshot 2024-01-15 at 1 50 40 PM

After:

Screenshot 2024-01-15 at 1 51 28 PM

Fixes #5549.

### How to test this PR?

Check Settings > Preferences.

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -231,7 +231,7 @@ export class ExtensionLoader {
     this.moduleLoader.overrideRequire();
     // register configuration for the max activation time
     const maxActivationTimeConfiguration: IConfigurationNode = {
-      id: 'preferences.extension',
+      id: 'preferences.extensions',
       title: 'Extensions',
       type: 'object',
       properties: {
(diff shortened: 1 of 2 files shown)


benoitf (commented):
what about doing the opposite ?

keeping `preferences.extensions` and move the default activation time to `preferences.extensions`

it's because the property being saved are:
```
"extensions.maxActivationTime": ...,
"extensions.autoCheckUpdates": ...
```

deboer-tim:
> what about doing the opposite ?
> ...

Yes, that's a better idea, will do.

benoitf (approved):

PR #5558, opened 2024-01-16
feat: adding navigation api

### LINKED ISSUE #5556: feat: adding navigation api

Extension need to be able to use the navigation

adding possibility to navigate 

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Fixes #5556

### How to test this PR?

You can use the extensionApi in an extension `await extensionApi.navigate.container.viewContainers();`


benoitf (changes requested):
backend side should not be aware of any routing path as it's a 'client side only'

so backend should never send ` apiSender.send('navigate', '/troubleshooting/repair-connections');` or this kind of stuff as you're linking backend to frontend

it can use constants, etc but routing path should only be known by frontend

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1079,49 +1080,101 @@ export class ExtensionLoader {
     const navigate: typeof containerDesktopAPI.navigate = {
       container: {
         viewContainers: async function (): Promise<void> {
-          apiSender.send('navigate', '/containers');
+          apiSender.send('navigate', {
+            page: Page.CONTAINERS,
+          } as NavigateRequest);
```
benoitf: wondering why we add `as NavigateRequest` as it's basically casting any object to be NavigateRequest even if the object is not compliant at all
If you're searching for type-safety, I would argue that you should call a function taking a NavigateRequest parameter that do the apiSender.send('navigate') instead of calling apiSender.send('navigate' everywhere in extension-loader.js so you've type-safety on calls and don't need the usage of `as NavigateRequest`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2692,4 +2692,60 @@ declare module '@podman-desktop/api' {
...
+    // Navigate to a specific pod referenced by kind, name and engineId in Podman-Desktop
+    viewPod(kind: string, name: string, engineId: string): Promise<void>;
+  }
+
+  export interface ContributionNavigation {
+    // Navigate to a specific contribution (aka extension page) referenced by name in Podman-Desktop
+    viewContribution(name: string): Promise<void>;
+  }
+
+  export namespace navigate {
```
benoitf: namespace should be a noun so `navigation` rather than `navigate`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2692,4 +2692,60 @@ declare module '@podman-desktop/api' {
...
+    export const image: ImageNavigation;
+
+    // Navigation methods related to volumes
+    export const volume: VolumeNavigation;
+
+    // Navigation methods related to pods
+    export const pod: PodNavigation;
+
+    // Navigation methods related to contributions
+    export const contribution: ContributionNavigation;
```
benoitf: do we need these elements as each method contains the name of the item

I mean by that, would it be more convenient to do

```
navigation.viewContainer(containerId)
```

instead of
```
navigation.container.viewContainer(containerId)
```

I would assume than navigateTo might be better than `view`

```
navigation.navigateToContainer(containerId)
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2692,4 +2692,60 @@ declare module '@podman-desktop/api' {
...
+    // Navigate to the Containers page in Podman-Desktop
+    viewContainers(): Promise<void>;
+    // Navigate to the Container page in Podman-Desktop
+    viewContainer(id: string): Promise<void>;
+    // Navigate to the Container logs tab in Podman-Desktop
+    viewContainerLogs(id: string): Promise<void>;
+    // Navigate to the Container inspect tab in Podman-Desktop
+    viewContainerInspect(id: string): Promise<void>;
+    // Navigate to the Container terminal tab in Podman-Desktop
+    viewContainerTerminal(id: string): Promise<void>;
```
benoitf: I think id might not be enough as you can have the same container's name between docker engine and podman engine (it requires engineId) as well

As we handle ContainerInfo or ImageInfo object etc we could accept this object directly ?
```
navigateToContainerDetails(containerInfo: ContainerInfo)
navigateToContainerDetails(engineId: string, id: string)
```

packages/main/src/plugin/extension-loader.ts
axel7083: Thanks for the tips, I haven't thought directly about this solution!

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1074,6 +1076,109 @@ export class ExtensionLoader {
       },
     };
 
+    const _navigate = (navigateRequest: NavigateRequest) => {
```
```suggestion
    const navigateTo = (navigateRequest: NavigateRequest) => {
```
do we need to use _ prefix ? usually this prefix is more to say that we ignore a variable from parameter list

packages/main/src/plugin/routes.ts
```diff
@@ -0,0 +1,21 @@
+export enum Page {
```
benoitf: I think we need to pickup a common pattern

filename is called routes
but there is a NavigateRequest object inside and a Page object

so it might be hard to know from these types come from

so might be interesting to use filename matching one of the object

like navigation.ts with NavigationRequest and NavigationPage objects (or better idea)

packages/extension-api/src/extension-api.d.ts
axel7083: > I think id might not be enough as you can have the same container's name between docker engine and podman engine (it requires engineId) as well
> 
> As we handle ContainerInfo or ImageInfo object etc we could accept this object directly ?
> 
> ```
> navigateToContainerDetails(containerInfo: ContainerInfo)
> navigateToContainerDetails(engineId: string, id: string)
> ```

How does this would translate in terms of path ? I used the `App.svelte` as a source of requirements for the routes

https://github.com/containers/podman-desktop/blob/caf13ef86dbed5a3f775963827ea2bdfc4eb54ee/packages/renderer/src/App.svelte#L95-L97

benoitf: it depends on what we're calling 'id' from consumer POV

for example on CLI you can do : `podman stop container-name` or `podman stop container-id`

so it means client won't be able to call `navigateToContainer('container-name')` but always use the `('container-shaID')`


if we accept ContainerInfo object it means we don't care much on what we authorize from id

also, what happens on the API if the id does not exists, should it return a boolean flag ?

axel7083: > it depends on what we're calling 'id' from consumer POV

For examples the method `createContainer` returns `ContainerCreateResult` which only contains an `id: string` property. On the other side, the ContainerInfo is only returned by the `listContainers` methods. If we only support from ContainerInfo an extension would have first to createContainer, then listContainers, find on id, navigateToContainer then.

Do we want to support both ? `navigateToContainer(shaId: string)` and `navigateToContainer(info: ContainerInfo)` ?

> Supporting both _seems_ to need different method name.

axel7083: > also, what happens on the API if the id does not exists, should it return a boolean flag ?

We might either throw an error ?

benoitf: for sure we can use same method name

but for now we can use only 'id' being the sha id (you can't give a container's name)

packages/main/src/plugin/navigation-page.ts
```diff
@@ -1,4 +1,4 @@
-export enum Page {
+export enum NavigationPage {
```
benoitf: FYI it is missing copyright headers

packages/main/src/plugin/navigation-request.ts
```diff
@@ -0,0 +1,6 @@
+import type { NavigationPage } from '/@/plugin/navigation-page.js';
```
benoitf: FYI it is missing copyright headers

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2693,59 +2693,34 @@ declare module '@podman-desktop/api' {
     ): Disposable;
   }
 
-  export interface ContainerNavigation {
+  export namespace navigation {
     // Navigate to the Containers page in Podman-Desktop
```
benoitf: I think you can remove all the ` in Podman-Desktop` as we're anyway inside Podman Desktop

packages/extension-api/src/extension-api.d.ts
axel7083: > > also, what happens on the API if the id does not exists, should it return a boolean flag ?
> 
> We might either throw an error ?

Or do we even want to check that ? This might be the extension responsibility ? And we simply display not found on the renderer if the id does not exist

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2692,4 +2692,35 @@ declare module '@podman-desktop/api' {
...
+    // Navigate to a specific volume in Podman-Desktop
+    export function navigateToVolume(name: string, engineId: string): Promise<void>;
+
+    // Navigate to the Pods page in Podman-Desktop
+    export function navigateToPods(): Promise<void>;
+    // Navigate to a specific pod referenced by kind, name and engineId in Podman-Desktop
+    export function navigateToPod(kind: string, name: string, engineId: string): Promise<void>;
+
+    // Navigate to a specific contribution (aka extension page) referenced by name in Podman-Desktop
+    export function navigateToContribution(name: string): Promise<void>;
```
benoitf: I'm not sure about this one yet as it's mainly a Docker Desktop extension

I suppose we would need to navigate to a webview but not sure about Docker Desktop contributions

packages/main/src/plugin/navigation-request.ts
```diff
@@ -0,0 +1,6 @@
...
+
+export interface NavigationRequest {
+  page: NavigationPage;
+  parameters?: Record<string, string>;
```
benoitf:
```suggestion
  parameters?: { [key: string]: string };
```

packages/renderer/src/Route.ts
```diff
@@ -24,3 +27,53 @@
...
  */
 export type NavigationHint = 'root' | 'details' | 'tab';
+
+export const navigationHandle = (page: NavigationPage, parameters?: Record<string, string>) => {
+  switch (page) {
+    case NavigationPage.CONTAINERS:
+      router.goto('/containers');
+      break;
+    case NavigationPage.CONTAINER:
+      parameters && router.goto(`/containers/${parameters['id']}/`);
```
benoitf: either we say that checking only parameters is not enough as we can come with parameters and another field that id
or if we don't care then you can use optional field ? and remove `parameters &&`

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1074,6 +1076,99 @@ export class ExtensionLoader {
       },
     };
 
+    const navigateTo = (navigateRequest: NavigationRequest) => {
+      this.apiSender.send('navigate', navigateRequest);
+    };
+    const navigation: typeof containerDesktopAPI.navigation = {
+      navigateToContainers: async function (): Promise<void> {
+        navigateTo({ page: NavigationPage.CONTAINERS });
+      },
```
benoitf: we should not use function keywords there but use arrow functions
```suggestion
      navigateToContainers: async : Promise<void> => {
        navigateTo({ page: NavigationPage.CONTAINERS });
      },
```

benoitf (commented):
looks almost OK

the only thing odd is the return type of functions if we do not redirect to the expected resource of if the resource is not available

let say I call: `navigateToContainer('foo')`

axel7083:
> looks almost OK
> 
> the only thing odd is the return type of functions if we do not redirect to the expected resource of if the resource is not available
> 
> let say I call: `navigateToContainer('foo')`

I would argue that the renderer would be responsible of handling the error and display a `Not found`, once called `navigateToX` you are not responsible of the result as we are navigating to another page.

### Some examples 
- (tinro) `router.goto('/foo');` does not return anything, even if the page is not available, they could check for 404 but they don't
- (react-router-dom) `navigate('hello')` have the same behavior

packages/extension-api/src/extension-api.d.ts
axel7083: I was thinking about the webview mainly when making this one

benoitf:
> I would argue that the renderer would be responsible of handling the error and display a Not found, once called navigateToX you are not responsible of the result as we are navigating to another page.


if I call `navigateToContainer('unknownId')` I want to receive an error (or a promise with a result) so from the extension I know the resource I'm asking for does not exist

if the resource exists then it's ok to let the renderer handle the display

but if I have an extension that asks to navigate to ('00000') then I see a '00000' does not exists in the UI I would have preferred to not navigate to this page and either extension handling the error or reporting it so we know who is asking for that.

From the signature POV of the method, if it returns a Promise<void> I expect that it ends when I'm redirected and that I receive an error if it failed else it's not an asynchronous call if I'm not notified when redirect happens or if redirect failed so no need of async

axel7083:
> if I call `navigateToContainer('unknownId')` I want to receive an error (or a promise with a result) so from the extension I know the resource I'm asking for does not exist

I added the implementation of the following
- [x] `assertContainerIdExist`
- [x] `assertImageExist`
- [x] `assertVolumeExist`
- [x] `assertPodExist`
- [x] `assertContributionExist`

They throw an error if the resource does not exist, which is propagated to the extension for it to handle it.

_I will start tomorrow writing tests._

packages/main/src/plugin/extension-loader.spec.ts
```diff
@@ -803,3 +819,458 @@ test('Verify extension uri', async () => {
...
+          parameters: {
+            id: 'valid',
+          },
+        },
+      },
+    ];
+
+    for (const testCase of tests) {
+      // Call the method provided
+      await testCase.method('valid');
```
benoitf: I think you should use a parameterized test that would better fit in the test model

with test.each() / https://vitest.dev/api/#test-each

example in Podman Desktop https://github.com/containers/podman-desktop/blob/4767272a1562e9b4cfc85674578228abccfbfd63/packages/main/src/plugin/types/uri.spec.ts#L63-L69

packages/main/src/plugin/extension-loader.spec.ts
```diff
@@ -803,3 +819,458 @@ test('Verify extension uri', async () => {
...
+      },
+      {
+        name: 'navigateToContainerTerminal',
+        method: api.navigation.navigateToContainerTerminal,
+      },
+    ];
+
+    for (const testCase of tests) {
+      // Call the method provided
+      let error = undefined;
```
benoitf: same remark here, IMHO use parameterized test

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1074,6 +1078,145 @@ export class ExtensionLoader {
...
+      }
+      if (imageInfo.RepoTags?.find(repoTag => repoTag === tag) === undefined) {
+        throw new Error(`Image with id ${id}, engine id ${engineId} and tag ${tag} cannot be found.`);
+      }
+    };
+    const assertVolumeExist = async (name: string, engineId: string) => {
+      const volumes = await containerProviderRegistry.listVolumes();
+      const allVolumes = volumes.map(volume => volume.Volumes).flat();
+      const matchingVolume = allVolumes.find(volume => volume.Name === name && volume.engineId === engineId);
+      if (matchingVolume === undefined) {
```
```suggestion
      if (!matchingVolume) {
```

or use some()

```ts
if (!allVolumes.some(....)
```

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1074,6 +1078,145 @@ export class ExtensionLoader {
...
+        throw new Error(`Volume with name ${name} and engine id ${engineId} cannot be found.`);
+      }
+    };
+    const assertPodExist = async (kind: string, name: string, engineId: string) => {
+      const pods = await containerProviderRegistry.listPods();
+      const matchingPod = pods.find(
+        podInPods => podInPods.Name === name && podInPods.engineId === engineId && kind === podInPods.kind,
+      );
+      if (matchingPod === undefined) {
```
benoitf: same remark there

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1074,6 +1078,145 @@ export class ExtensionLoader {
...
+        throw new Error(`Pod with kind ${kind}, name ${name} and engine id ${engineId} cannot be found.`);
+      }
+    };
+    const assertContributionExist = async (name: string) => {
+      const contribs = this.contributionManager.listContributions();
+      if (contribs.find(contrib => contrib.name === name) === undefined) {
```
benoitf: same remark there

packages/main/src/plugin/extension-loader.ts
benoitf: I think all assert should be part of where it's used

like assertVolumeExist should be moved to containerProviderRegistry class, etc.

extensionLoader should be as small as possible

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1074,6 +1078,145 @@ export class ExtensionLoader {
...
+      navigateToContribution: async (name: string): Promise<void> => {
+        await assertContributionExist(name);
+        navigateTo({
+          page: NavigationPage.CONTRIBUTION,
+          parameters: {
+            name: name,
+          },
+        });
+      },
+    };
```
benoitf: it make me feels wondering if we should not delegate / move these lines to a dedicated file

usually we're just delegating but here you're adding logic (all assert calls)

packages/main/src/plugin/navigation-page.ts
```diff
@@ -0,0 +1,34 @@
+/**********************************************************************
```
benoitf: create a subfolder navigation ?

plugin/navigation/navigation-page, navigation-request

packages/renderer/src/App.svelte
```diff
@@ -53,14 +55,8 @@ router.subscribe(function (navigation) {
...
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
-window.events?.receive('display-help', () => {
-  router.goto('/help');
-});
-
-window.events?.receive('display-troubleshooting', () => {
-  router.goto('/troubleshooting/repair-connections');
+window.events?.receive('navigate', (navigationRequest: NavigationRequest) => {
+  navigationHandle(navigationRequest.page, navigationRequest.parameters);
```
benoitf: nice to also take care of that

axel7083: I do not get what you want ?

benoitf: just saying that it's nice to have included existing troubleshooting and help navigation into the new system (but as I commented on a line and not the right block you could not understand)

packages/main/src/plugin/container-registry.ts
```diff
@@ -2063,4 +2063,27 @@ export class ContainerProviderRegistry {
...
+    const imageInfo = images.find(c => c.Id === id && c.engineId === engineId);
+    return imageInfo?.RepoTags !== undefined && imageInfo.RepoTags.some(repoTag => repoTag === tag);
+  }
+
+  async volumeExist(name: string, engineId: string): Promise<boolean> {
+    const volumes = await this.listVolumes();
+    const allVolumes = volumes.map(volume => volume.Volumes).flat();
+    return allVolumes.some(volume => volume.Name === name && volume.engineId === engineId);
+  }
+  async podExist(kind: string, name: string, engineId: string) {
```
benoitf: it is missing the return type 
```suggestion
  async podExist(kind: string, name: string, engineId: string): Promise<boolean> {
```

benoitf (approved):
👍 just one minor comment

PR #5572, opened 2024-01-17
chore(tests): add prune containers e2e test

### LINKED ISSUE #5545: Implement E2E test case using Playwright to verify Prune images, containers, pods, volumes

I'd like to add test cases for pruning images, containers, pods and volumes.

Implement e2e test cases that check the respective scenarios and verifies the results.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?

Adds e2e tests for pruning functionality.

Closes https://github.com/containers/podman-desktop/issues/5545


tests/src/container-smoke.spec.ts
```diff
@@ -158,4 +167,35 @@ describe('Verification of container creation workflow', async () => {
...
+      let containersPage = new ContainersPage(page);
+      const containersDetails = await containersPage.openContainersDetails(container);
+      await playExpect(containersDetails.heading).toBeVisible();
+      await playExpect(containersDetails.heading).toContainText(container);
+      playExpect(await containersDetails.getState()).toContain(ContainerState.Running);
+      await containersDetails.stopContainer();
+      await playExpect(await containersDetails.getStateLocator()).toHaveText(ContainerState.Exited, { timeout: 20000 });
+      containersPage = await navigationBar.openContainers();
+      await playExpect(containersPage.heading).toBeVisible();
+      await containersPage.pruneContainers();
```
odockal, line 189: I would expect prune called outside of the iteration, right?

tests/src/container-smoke.spec.ts
```diff
@@ -153,9 +162,40 @@ describe('Verification of container creation workflow', async () => {
...
+    await playExpect(containersPage.heading).toBeVisible();
+    await playExpect.poll(async () => await containersPage.containerExists(containerToRun)).toBeFalsy();
     await pdRunner.screenshot('containers-container-deleted.png');
   });
+
+  test('Prune containers', async () => {
+    const navigationBar = new NavigationBar(page);
+
+    for (const container of containerList) {
+      const images = await navigationBar.openImages();
```
odockal, line 175: I can imagine this exact functionality to be extracted with assert in some helper class or utility, because this exact code can be definitely reuse on other places.

tests/src/container-smoke.spec.ts
```diff
@@ -153,9 +162,40 @@ describe('Verification of container creation workflow', async () => {
...
+      const containersPage = await runImage.startContainer(container);
+      await playExpect(containersPage.heading).toBeVisible();
+      await playExpect
+        .poll(async () => await containersPage.containerExists(container), { timeout: 15000 })
+        .toBeTruthy();
+    }
+
+    for (const container of containerList) {
+      let containersPage = new ContainersPage(page);
+      const containersDetails = await containersPage.openContainersDetails(container);
```
odockal: For the purpose of stopping container I am not sure we need to go into container details, right? We could stay on Containers page and simply stop it from there?
Also extraction could be nice.

tests/src/image-smoke.spec.ts
```diff
@@ -120,4 +114,33 @@ describe('Image workflow verification', async () => {
...
   });
+
+  test('Prune images', async () => {
+    const imageList = ['quay.io/podman/image1', 'quay.io/podman/image2'];
+
+    let imagesPage = await navBar.openImages();
+    await playExpect(imagesPage.heading).toBeVisible();
+
+    for (const image of imageList) {
+      const pullImagePage = await imagesPage.openPullImage();
```
odockal: also extraction could be used?

tests/src/model/pages/containers-page.ts
```diff
@@ -47,25 +49,38 @@ export class ContainersPage extends MainPage {
...
-      const index = thirdCell.indexOf(name);
-      if (index >= 0) {
-        return row;
+    } catch (err) {
+      console.log(`Exception caught on containers page with message: ${err}`);
+    }
+    return undefined;
+  }
+
+  async uncheckAllContainers(): Promise<void> {
```
odockal, line 82: this is ok, but there is also an check/uncheck all option in table's header row I think.

tests/src/model/pages/images-page.ts
```diff
@@ -66,40 +75,33 @@ export class ImagesPage extends MainPage {
...
-      }
-      const thirdCell = await row.getByRole('cell').nth(3).innerText();
-      const index = thirdCell.indexOf(name);
-      if (index >= 0) {
-        return row;
+    try {
+      const table = await this.getTable();
+      const rows = await table.getByRole('row').all();
+      for (let i = rows.length - 1; i > 0; i--) {
+        const thirdCell = await rows[i].getByRole('cell').nth(3).getByText(name).count();
```
odockal: Just a note: I would be careful with getByText (same it was before with checking indexOf innerText) which also can find substring of the image name. 
so If we got two images (`alpine` and `alpine-hello`) we might end up simply believing we are working with the right image. 
It always returns first best match... 
I can imagine throwing an error in case when we received ambiguous results.

odockal (changes requested):

tests/src/container-smoke.spec.ts
cbr7: no, because this way we're also checking that containers that are still in use don't get pruned. The test stops the containers one by one and calls each time prune command and checks that only the correct container is removed.

tests/src/container-smoke.spec.ts
cbr7: What functionality? This part only creates some containers, same as previous tests in the file.

tests/src/image-smoke.spec.ts
cbr7: I'm not sure what you mean by extraction?

tests/src/container-smoke.spec.ts
```diff
@@ -27,12 +27,14 @@ import { waitUntil, waitWhile } from './utility/wait';
...
 import { ContainerState } from './model/core/states';
 import type { ImagesPage } from './model/pages/images-page';
+import { ContainersPage } from './model/pages/containers-page';
 
 let pdRunner: PodmanDesktopRunner;
 let page: Page;
 const imageToPull = 'ghcr.io/linuxcontainers/alpine';
 const imageTag = 'latest';
 const containerToRun = 'alpine-container';
+const containerList = ['first', 'second', 'third'];
```
ScrewTSW, line 37: Are the names in this list arbitrary? Where does the initialization of said containers happen?

tests/src/model/pages/containers-page.ts
cbr7: There is a bulk selector, but it has multiple states and in general it's much harder to manage for tests that don't actually need bulk actions. To be honest this was created mostly to bypass the selection box caching that happens in containers, which is not present in the other pages.

tests/src/image-smoke.spec.ts
```diff
@@ -120,4 +114,33 @@ describe('Image workflow verification', async () => {
     imagesPage = await imageDetailsPage.deleteImage();
     expect(await imagesPage.waitForImageDelete('build-test-image')).toBeTruthy();
   });
+
+  test('Prune images', async () => {
+    const imageList = ['quay.io/podman/image1', 'quay.io/podman/image2'];
```
ScrewTSW: I would prefer for this to be declared outside of the test case at the top of the smoke test

tests/src/model/pages/container-details-page.ts
```diff
@@ -79,17 +79,9 @@ export class ContainerDetailsPage extends BasePage {
...
     await deleteButton.click();
-    await waitWhile(
-      async () => await this.heading.isVisible(),
-      timeout,
-      1000,
-      true,
-      `Container was not deleted in ${timeout / 1000}s`,
-    );
-    // after delete is successful we expect to see containers page
     return new ContainersPage(this.page);
```
ScrewTSW, line 87: Won't this call fail unless the expected page is present in DOM?

tests/src/model/pages/images-page.ts
ScrewTSW: Also, might be a good idea to label the row inside .svelte providing ARIA functionality and also simplifying the test logic.

tests/src/model/pages/pods-page.ts
```diff
@@ -62,26 +59,19 @@ export class PodsPage extends BasePage {
...
-      }
-      const nameCell = await row.getByRole('cell').nth(3).innerText();
-      const index = nameCell.indexOf(name);
-      if (index >= 0) {
-        return row;
+    try {
+      const podsTable = await this.getTable();
+      const rows = await podsTable.getByRole('row').all();
+      for (let i = rows.length - 1; i > 0; i--) {
+        const nameCell = await rows[i].getByRole('cell').nth(3).getByText(name).count();
```
ScrewTSW: Same for here, ARIA functionality might be a more elegant and stable solution

tests/src/model/pages/run-image-page.ts
```diff
@@ -81,22 +80,7 @@ export class RunImagePage extends BasePage {
...
-        true,
-        'Error starting a container',
-      );
-    } catch (err) {
-      await this.errorAlert.waitFor({ state: 'visible', timeout: 2000 });
-      const errMessage = await this.errorAlert.innerText({ timeout: 1000 });
-      throw new Error(`Error starting a container: ${errMessage}`);
-    }
-    return containers;
+    return new ContainersPage(this.page);
```
ScrewTSW, line 83: Same question here about failures if the page is not already in DOM

tests/src/model/pages/images-page.ts
cbr7: That's a fair point, will change it to getByText with exact param enabled.

tests/src/model/pages/pods-page.ts
cbr7: I agree, but this is mostly a refactor of current functionality, I don't want to start changing every locator in the framework in one PR otherwise it will never get merged.

tests/src/model/pages/run-image-page.ts
cbr7: The failure would happen regardless, the refactor of this code did not change that. Navigation validation should be done outside of these functional methods, preferably inside the tests themselves, via assertions.

tests/src/model/pages/images-page.ts
cbr7: Actually checking exactly will no be possible for now on images without further refactoring, since when we're checking for an image on the page we're looking for a substring inside the tests, not the entire name, but I have changed it to exact for containers and pods.

tests/src/container-smoke.spec.ts
cbr7: I'm not sure what you're asking, yes the container names are just made up.

tests/src/image-smoke.spec.ts
cbr7: ok, but is there a specific reason?

tests/src/model/pages/container-details-page.ts
cbr7: Not anymore than it was failing previously, validating navigation should be done outside of functional methods.

tests/src/container-smoke.spec.ts
cbr7: Potentially, I just followed the template from the previous tests that did the same action.

tests/src/image-smoke.spec.ts
cbr7: To be honest I just looked at the other tests inside the test file and all of them have the image names defined inside the test, not globally. Is there a particular benefit to having them defined globally?

tests/src/model/pages/images-page.ts
cbr7: ok changed images, containers and pods to be matched exactly when looking in the table

tests/src/model/pages/pods-page.ts
cbr7: Created https://github.com/containers/podman-desktop/issues/5639 to track this effort.

tests/src/container-smoke.spec.ts
ScrewTSW: ah, I see, thank you for the explanation

tests/src/image-smoke.spec.ts
ScrewTSW: Well, turning things like these into a const readonly at the top of the test file improves readability, you know right-away what image is being used, constants are a good practice in general. If other tests do it like that, you can get bonus points for refactoring a bit :D

cbr7: Moved the image names to global context for all tests in the file.

tests/src/container-smoke.spec.ts
odockal: Exactly, I was thinking that we can extract the logic for creating a container into a helper or utility class so that the code does not need to be duplicated.

tests/src/image-smoke.spec.ts
odockal: This whole code could be extracted into a function with parameters. You could also have a part where you pull an image and rename it, or it could be two individual functions.

tests/src/runner/podman-desktop-runner.ts
```diff
@@ -51,29 +51,50 @@ export class PodmanDesktopRunner {
...
     }
-
-    // start the app with given properties
-    this._app = await electron.launch({
-      ...this._options,
-    });
-    // setup state
-    this._running = true;
-    this._page = await this.getElectronApp().firstWindow();
+    let retry = true;
```
odockal: I would prefer this kind of changes to be done outside of a PR introducing new tests.

odockal (changes requested):

tests/src/image-smoke.spec.ts
cbr7: done

tests/src/container-smoke.spec.ts
cbr7: done

tests/src/runner/podman-desktop-runner.ts
cbr7: reverted this part, will make another PR for this

tests/src/model/pages/containers-page.ts
```diff
@@ -84,4 +110,11 @@ export class ContainersPage extends MainPage {
     await this.page.getByRole('button', { name: `Create Pod` }).click();
     return new CreatePodsPage(this.page);
   }
+
+  async pruneContainers(): Promise<ContainersPage> {
+    await this.pruneContainersButton.click();
+    await this.pruneConfirmationButton.waitFor({ state: 'visible', timeout: 3000 });
+    await this.pruneConfirmationButton.click();
```
odockal: Can you please use `handleConfirmationDialog` as is used here: https://github.com/containers/podman-desktop/pull/5572/files#diff-4b11f9fa58190b54db2662ae7038e089037f9df7d2fa8aa6aa8104572920d64eR69 ?

tests/src/model/pages/image-edit-page.ts
```diff
@@ -33,4 +35,18 @@ export class ImageEditPage extends BasePage {
     this.name = name;
     this.imageTag = page.getByLabel('imageTag');
   }
+
+  async renameImage(name: string): Promise<ImagesPage> {
```
odockal, line 39: I am looking at the page/dialog dom: 
```
<div class="modal svelte-ipux5s" role="dialog" aria-label="Edit Image" aria-modal="true">
```

We can easily improve the edit-image to use this root element as a container locator. Can yu please fill an issue that would follow up to this? no need to tackle that in this PR.

tests/src/model/pages/images-page.ts
```diff
@@ -57,6 +80,13 @@ export class ImagesPage extends MainPage {
     return new ImageDetailsPage(this.page, name);
   }
 
+  async pruneImages(): Promise<ImagesPage> {
+    await this.pruneImagesButton.click();
+    await this.pruneConfirmationButton.waitFor({ state: 'visible', timeout: 3000 });
+    await this.pruneConfirmationButton.click();
```
odockal: same as above-we can use handleconfirmationdialog function from operations utility file.

tests/src/model/pages/pods-page.ts
```diff
@@ -92,4 +82,11 @@ export class PodsPage extends BasePage {
     await this.playKubernetesYAMLButton.click();
     return new PlayKubeYamlPage(this.page);
   }
+
+  async prunePods(): Promise<PodsPage> {
+    await this.prunePodsButton.click();
+    await this.pruneConfirmationButton.waitFor({ state: 'visible', timeout: 3000 });
+    await this.pruneConfirmationButton.click();
```
odockal: also using the dialog handling would be nice.

odockal (changes requested):
small changes, please take a look.

tests/src/model/pages/image-edit-page.ts
cbr7: Created issue here https://github.com/containers/podman-desktop/issues/5752

tests/src/model/pages/containers-page.ts
cbr7: done

tests/src/model/pages/pods-page.ts
cbr7: done

tests/src/model/pages/images-page.ts
cbr7: done

tests/src/model/pages/image-edit-page.ts
odockal: thx

odockal (approved):

PR #5578, opened 2024-01-17
chore: unused palette entries

deboer-tim:
### What does this PR do?

I had to look at the palette for another PR and noticed that three of the old legacy colors are no longer in use just due to natural evolution. Removing these so they don't accidentally get used again.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Confirm no test failures or grep to confirm no use of these colors.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
@@ -70,7 +70,7 @@ function validate(event: any) {
         id="toggle-proxy"
         class="sr-only peer" />
       <div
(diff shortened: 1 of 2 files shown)


cdrage (approved):

benoitf (changes requested):
I see a reference to violet-800 there
https://github.com/containers/podman-desktop/blob/main/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte#L73

benoitf (approved):

deboer-tim:
Doh. Sorry, I guess I had a change to that file in my workspace that had already removed it.

I've tried changing the class to a bright color and confirmed that it has no effect no matter what I do/click on the proxies page - either it hasn't been doing anything for a while or maybe removing PF had an impact. Either way, I've just removed the class for now.

PR #5579, opened 2024-01-17
chore: cleanup API sender object

benoitf:
### What does this PR do?
replace any type by unknown type
(and remove all eslint ignore/disable lines like @typescroipt-eslint/no-explicit-any)
also I had to fixes issues where we can't use async methods when sync methods are expected (before it was silently hidden due to the lack of signature/usage of any type)

make it consistent across all the packages and ensure it returns a disposable object for receiving events

### Screenshot / video of UI

N/A

consistency

### How to test this PR?

I've added a unit test on the implementation of Api Sender

But you could check that communication is working (UI is displaying correct objects ;-)

### DIFF
--- a/packages/main/src/plugin/contribution-manager.spec.ts
+++ b/packages/main/src/plugin/contribution-manager.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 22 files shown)


packages/main/src/plugin/contribution-manager.spec.ts
```diff
@@ -42,14 +43,17 @@ const portNumber = 10000;
...
-const send = (channel: string, data?: any) => {
+const send = (channel: string, data?: unknown) => {
   eventEmitter.emit(channel, data);
 };
 
-const receive = (channel: string, func: any) => {
-  eventEmitter.on(channel, data => {
-    func(data);
-  });
+const receive = (channel: string, func: any): IDisposable => {
```
lstocchi: forgot to convert this any?

benoitf: ah yes it was in a test, will fix it

benoitf: fixed in a new commit

lstocchi (approved):
LGTM

PR #5619, opened 2024-01-22
fix: fix UI not being refreshed if container is only created

### LINKED ISSUE #5617: missing container in container list in case of startup issue
### Bug description

Reported by a user:

I created two containers, which bind to the same port.
```
podman run -d  -p 5984:5984 --name my-couchdb -e COUCHDB_USER=admin -e COUCHDB_PASSWORD=password apache/couchdb:3.3.3
```

```
podman run -d  -p 5984:5984 --name my-couchdb-client -e COUCHDB_USER=admin -e COUCHDB_PASSWORD=password apache/couchdb:3.3.3
```

This is an invalid configuration, so the second container crashed with the message:
Error: cannot listen on the TCP port: listen tcp4 :5984: bind: address already in use
This is expected.

I then change the ports, and run again.

I get the message:

Error: creating container storage: the container name "my-couchdb-client" is already in use by 8df9472abe3bd4cf0e0c6450c719e5447ae521a6e3b58e5a5cdd4ec540253d8f. You have to remove that container to be able to reuse that name: that name is already in use

But `my-couchdb-client` is not visible in the GUI.

### Operating system

Windows

### Installation Method

None

### Version

1.6.3

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Ensure pods, containers and volumes are updated when containers are just created (not initialized or started)

### Screenshot / video of UI

N/A

fixes https://github.com/containers/podman-desktop/issues/5617

### How to test this PR?

Unit tests

(can also use test case from linked issue for manual check)

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -129,6 +129,9 @@ export class ContainerProviderRegistry {
       } else if (jsonEvent.status === 'init' && jsonEvent?.Type === 'container') {
         // need to notify that a container has been started
         this.apiSender.send('container-init-event', jsonEvent.id);
+      } else if (jsonEvent.status === 'create' && jsonEvent?.Type === 'container') {
+        // need to notify that a container has been created
+        this.apiSender.send('container-created-event', jsonEvent.id);
       } else if (jsonEvent.status === 'start' && jsonEvent?.Type === 'container') {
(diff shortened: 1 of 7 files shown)


deboer-tim (approved):
LGTM

PR #5641, opened 2024-01-22
chore: improve nav sections

### LINKED ISSUE #5618: nav extension for docker desktop extensions issues
### Bug description

- 1. it's displayed even if I don't have any Docker Desktop extensions (so I click on it and I have an empty state)

IMHO if there is none, we shouldn't display it at all

https://github.com/containers/podman-desktop/assets/436777/bf8c6910-5405-4eec-81a8-4dacafcc3a4d

- 2. if there is only extension, it should be in expanded state (why click on an item that is taking the same height)
image

- 3. Clicking on the item is moving the height of the component

https://github.com/containers/podman-desktop/assets/436777/474bf713-6813-4c17-8849-ca80b2b2fc03

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Improving a number of issues with the new nav sections. Here are the changes in behaviour/UI:
- Webviews are added to the same section as the Docker extensions, and it is renamed to just Extensions. (This reduces the point of separating Webviews.svelte a bit - maybe future chore to have an an Extensions.svelte for both?)
- If there are no items, the section is not shown.
- Sections are now expanded by default, but collapsed by default if there are more than 3 items. (this is arbitrary, but 4+ felt like a lot to me)
- If there is only one item, the button to collapse is greyed out and disabled. I tried it without this, but even after you fix the height issue IMHO it was really confusing what it is doing ('why are you giving me a button to hide the nav item??') This felt like the right balance between showing that there is a section and what it is (you can still hover over the collapse button to see the tooltip) without confusing users.

Technical notes:
- The only way to count the number of items in a section is using the context. The nice side benefit of this is that we no longer need to tell NavItems that they are inSection - if the context exists, they are. It makes testing a little harder though.
- Why 'expanded && $count < 2'? Two reasons: 1) when you collapse the item count (correctly) goes to zero and you always want to be able to expand. 2) if there is one item and a second is added dynamically. the collapse button/styling will be removed.

### Screenshot / video of UI

Shown with a fake Kubernetes section that only has one item so you can see what it looks like:

https://github.com/containers/podman-desktop/assets/19958075/edafafee-89e7-455c-8887-60dbcb331f47

Fixes #5618.

### How to test this PR?

Automated tests, plus try sections out with different number of extensions installed

### DIFF
--- a/packages/renderer/src/lib/ui/NavSection.svelte
+++ b/packages/renderer/src/lib/ui/NavSection.svelte
@@ -1,10 +1,13 @@
 <script lang="ts">
+/* eslint-disable import/no-duplicates */
+// https://github.com/import-js/eslint-plugin-import/issues/1479
 import { faChevronDown, faChevronUp } from '@fortawesome/free-solid-svg-icons';
 import Tooltip from './Tooltip.svelte';
 import Fa from 'svelte-fa';
 import { onMount, setContext } from 'svelte';
 import { cubicOut } from 'svelte/easing';
 import { writable, type Writable } from 'svelte/store';
+/* eslint-disable import/no-duplicates */
 
 export let expanded: boolean = true;
 export let tooltip: string;
--- a/packages/renderer/src/lib/ui/NavSection.svelte
+++ b/packages/renderer/src/lib/ui/NavSection.svelte
@@ -2,10 +2,14 @@
 import { faChevronDown, faChevronUp } from '@fortawesome/free-solid-svg-icons';
 import Tooltip from './Tooltip.svelte';
 import Fa from 'svelte-fa';
+import { onMount, setContext } from 'svelte';
 import { cubicOut } from 'svelte/easing';
+import { writable, type Writable } from 'svelte/store';
 
-export let expanded: boolean = false;
+export let expanded: boolean = true;
 export let tooltip: string;
+
+const count: Writable<number> = setContext('nav-items', writable(0));
 
 function fadeSlide(node: any, { delay = 0, duration = 400, easing = cubicOut }) {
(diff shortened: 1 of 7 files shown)


mairin:
LGTM!

mairin (approved):
LGTM

packages/renderer/src/lib/ui/NavSection.svelte
```diff
@@ -26,6 +33,13 @@ function fadeSlide(node: any, { delay = 0, duration = 400, easing = cubicOut })
       `padding-bottom: ${t * paddingBottom}px;`,
   };
 }
+
+onMount(() => {
+  // collapse by default if > 3 items
+  if ($count > 3) {
```
benoitf, line 39: maybe a property ( defaulted to 3) but we could have different values for kube or extensions

deboer-tim: Originally callers could just set expanded based on their requirements (e.g. could do `expanded="{x > 4}"`), but this line overrides whatever you've set so you can only bind to check the status afterward.

There's no way to set a dynamic default value in svelte, so if we want to both check the current state and set a default behaviour that means 2 properties (e.g. `initialCollapseThreshold` and `expanded` [current state at any time]). Ok if we defer this requirement to another issue so can discuss how best to expose the new property?

benoitf: sure

benoitf (approved):

PR #5643, opened 2024-01-23
feat: allow to create container without starting it

### LINKED ISSUE #5475: Podman Desktop API always starts container after creating
### Bug description

Within the API there is no way to create a container without it immediately starting after, see:

https://github.com/containers/podman-desktop/blob/f75dcfd8bccbd17525afb5e5ff54850cd4f5ddb0/packages/main/src/plugin/extension-loader.ts#L888-L889

```ts
      createContainer(
        engineId: string,
        containerCreateOptions: containerDesktopAPI.ContainerCreateOptions,
      ): Promise<containerDesktopAPI.ContainerCreateResult> {
        return containerProviderRegistry.createAndStartContainer(engineId, containerCreateOptions);
      },
```

createContainer should only createContainer and not start it right after.

we should be able to pass in an option to be able to start it immediately or not.

### Operating system

mac m1

### Installation Method

None

### Version

1.6.3

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR just gives the chance to create a container without it is actually started. 
It is needed when we create containers that have to be added to a pod in ai-studio.

### Screenshot / video of UI

N/A

it resolves #5475 

### How to test this PR?

1. run test

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2140,7 +2140,7 @@ declare module '@podman-desktop/api' {
     OpenStdin?: boolean;
     StdinOnce?: boolean;
     Detach?: boolean;
-    startOnCreation?: boolean;
+    start?: boolean;
   }
 
   export interface ContainerCreateResult {
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2140,6 +2140,7 @@ declare module '@podman-desktop/api' {
     OpenStdin?: boolean;
     StdinOnce?: boolean;
     Detach?: boolean;
+    startOnCreation?: boolean;
   }
 
   export interface ContainerCreateResult {
@@ -2234,7 +2235,6 @@ declare module '@podman-desktop/api' {
     export function createContainer(
       engineId: string,
       containerCreateOptions: ContainerCreateOptions,
-      startOnCreation?: boolean,
     ): Promise<ContainerCreateResult>;
     export function startContainer(engineId: string, id: string): Promise<void>;
     export function logsContainer(
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2234,6 +2234,7 @@ declare module '@podman-desktop/api' {
     export function createContainer(
       engineId: string,
       containerCreateOptions: ContainerCreateOptions,
+      startOnCreation?: boolean,
     ): Promise<ContainerCreateResult>;
     export function startContainer(engineId: string, id: string): Promise<void>;
     export function logsContainer(
--- a/packages/main/src/plugin/api/container-info.ts
+++ b/packages/main/src/plugin/api/container-info.ts
@@ -95,6 +95,7 @@ export interface ContainerCreateOptions {
   OpenStdin?: boolean;
   StdinOnce?: boolean;
   Detach?: boolean;
+  startOnCreation?: boolean;
 }
 
 export interface NetworkCreateOptions {
--- a/packages/main/src/plugin/api/container-info.ts
+++ b/packages/main/src/plugin/api/container-info.ts
@@ -95,7 +95,7 @@ export interface ContainerCreateOptions {
   OpenStdin?: boolean;
   StdinOnce?: boolean;
   Detach?: boolean;
-  startOnCreation?: boolean;
+  start?: boolean;
 }
 
 export interface NetworkCreateOptions {
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -2147,7 +2147,7 @@ describe('createAndStartContainer', () => {
       api: fakeDockerode,
     } as InternalContainerProvider);
 
-    const container = await containerRegistry.createAndStartContainer('podman1', {});
+    const container = await containerRegistry.createAndStartContainer('podman1', {}, true);
 
     expect(container.id).toBe(createdId);
     expect(createContainerMock).toHaveBeenCalled();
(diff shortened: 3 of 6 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -2234,6 +2234,7 @@ declare module '@podman-desktop/api' {
     export function createContainer(
       engineId: string,
       containerCreateOptions: ContainerCreateOptions,
+      startOnCreation?: boolean,
```
benoitf: I would put startOnCreation  inside `ContainerCreateOptions`

packages/main/src/plugin/container-registry.ts
```diff
@@ -1692,7 +1692,11 @@ export class ContainerProviderRegistry {
     }
   }
 
-  async createAndStartContainer(engineId: string, options: ContainerCreateOptions): Promise<{ id: string }> {
+  async createAndStartContainer(
+    engineId: string,
+    options: ContainerCreateOptions,
+    doStartContainer: boolean,
+  ): Promise<{ id: string }> {
```
benoitf: Probably need to rename the method createContainer() (as it is not starting all the time)

(we can still have aliases createAndStart that does the start)

benoitf:
there is an issue for your PR: https://github.com/containers/podman-desktop/issues/5475

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2140,6 +2140,7 @@ declare module '@podman-desktop/api' {
     OpenStdin?: boolean;
     StdinOnce?: boolean;
     Detach?: boolean;
+    startOnCreation?: boolean;
```
benoitf: sorry, but maybe we could just have `start` as we're inside the create Operation

```suggestion
    start?: boolean;
```

packages/main/src/plugin/api/container-info.ts
```diff
@@ -95,6 +95,7 @@ export interface ContainerCreateOptions {
   OpenStdin?: boolean;
   StdinOnce?: boolean;
   Detach?: boolean;
+  startOnCreation?: boolean;
```
benoitf:
```suggestion
  start?: boolean;
```

packages/main/src/plugin/index.ts
```diff
@@ -1077,7 +1077,8 @@ export class PluginSystem {
     this.ipcHandle(
       'container-provider-registry:createAndStartContainer',
       async (_listener, engine: string, options: ContainerCreateOptions): Promise<{ id: string }> => {
-        return containerProviderRegistry.createAndStartContainer(engine, options);
+        options.startOnCreation = true;
```
```suggestion
        options.start = true;
```

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -2147,14 +2147,14 @@ describe('createAndStartContainer', () => {
       api: fakeDockerode,
     } as InternalContainerProvider);
 
-    const container = await containerRegistry.createAndStartContainer('podman1', {});
+    const container = await containerRegistry.createContainer('podman1', { startOnCreation: true });
```
benoitf: should it be 
```suggestion
    const container = await containerRegistry.createContainer('podman1', { start: true });
```

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -2204,6 +2204,42 @@ describe('createAndStartContainer', () => {
...
+    containerRegistry.addInternalProvider('podman1', {
+      name: 'podman1',
+      id: 'podman1',
+      connection: {
+        type: 'podman',
+      },
+      api: fakeDockerode,
+    } as InternalContainerProvider);
+
+    const container = await containerRegistry.createContainer('podman1', { startOnCreation: false });
```
benoitf:
```suggestion
    const container = await containerRegistry.createContainer('podman1', { start: false });
```

packages/main/src/plugin/container-registry.ts
```diff
@@ -1718,13 +1718,15 @@ export class ContainerProviderRegistry {
 
       const container = await engine.api.createContainer(options);
       await this.attachToContainer(engine, container, options.Tty, options.OpenStdin);
-      await container.start();
+      if (options.startOnCreation === true || options.startOnCreation === undefined) {
```
```suggestion
      if (options.start === true || options.start === undefined) {
```

axel7083:
What is the default behavior ? if I run ` await containerRegistry.createContainer('podman1')` does it start or not ?

lstocchi:
> What is the default behavior ? if I run ` await containerRegistry.createContainer('podman1')` does it start or not ?

the container gets started

packages/main/src/plugin/container-registry.spec.ts
lstocchi: I thought you added all changes. I just committed them and went to have lunch. Checking everything now

benoitf: yes it's my fault I didn't suggested changes everywhere

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2140,6 +2140,7 @@ declare module '@podman-desktop/api' {
     OpenStdin?: boolean;
     StdinOnce?: boolean;
     Detach?: boolean;
+    start?: boolean;
```
cdrage, line 2143: if we add this to the API, doesn't this mean that we are now not 1-1 with the dockerode api?

Unsure if this is okay @benoitf

cdrage (changes requested):

packages/extension-api/src/extension-api.d.ts
benoitf: it's ok dockerode will ignore the flag

benoitf: I think from user POV, it's better to have one option argument rather than two block of arguments, b/c they're all around creating a container

benoitf (approved):
👍

cdrage (approved):
LGTM. Thanks for addressing my concerns @benoitf

PR #5648, opened 2024-01-23
feat: expose create/start Pod and replicatePodmanContainer

lstocchi:
### What does this PR do?

It just exposes the createPod, startPod and replicatePodmanContainer functions

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

N/A

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -270,6 +270,8 @@ declare module '@podman-desktop/api' {
   export interface PodCreateOptions {
     name: string;
     portmappings?: PodCreatePortOptions[];
+    // Set the provider to use, if not we will try select the first one available (sorted in favor of Podman).
+    provider?: ProviderContainerConnectionInfo | ContainerProviderConnection;
   }
 
   export interface KubernetesProviderConnectionEndpoint {
@@ -2294,10 +2296,7 @@ declare module '@podman-desktop/api' {
     export function createVolume(options?: VolumeCreateOptions): Promise<VolumeCreateResponseInfo>;
     export function deleteVolume(volumeName: string, options?: VolumeDeleteOptions): Promise<void>;
 
-    export function createPod(
-      podOptions: PodCreateOptions,
-      selectedProvider?: ProviderContainerConnectionInfo | ContainerProviderConnection,
-    ): Promise<{ engineId: string; Id: string }>;
+    export function createPod(podOptions: PodCreateOptions): Promise<{ engineId: string; Id: string }>;
     export function replicatePodmanContainer(
       source: { engineId: string; id: string },
       target: { engineId: string },
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -259,6 +259,19 @@ declare module '@podman-desktop/api' {
     status(): ProviderConnectionStatus;
   }
 
+  export interface PodCreatePortOptions {
+    host_ip: string;
+    container_port: number;
+    host_port: number;
+    protocol: string;
+    range: number;
+  }
+
+  export interface PodCreateOptions {
+    name: string;
+    portmappings?: PodCreatePortOptions[];
+  }
+
   export interface KubernetesProviderConnectionEndpoint {
     apiURL: string;
   }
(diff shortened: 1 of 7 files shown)


packages/main/src/plugin/container-registry.ts
```diff
@@ -1132,7 +1132,7 @@ export class ContainerProviderRegistry {
   }
 
   async createPod(
-    selectedProvider: ProviderContainerConnectionInfo,
+    selectedProvider: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection,
```
benoitf: I think this one is missing a unit test

benoitf: also in the code we should use if `getMatchingEngineFromConnection(selectedProvider)`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2279,6 +2292,17 @@ declare module '@podman-desktop/api' {
     export function listVolumes(): Promise<VolumeListInfo[]>;
     export function createVolume(options?: VolumeCreateOptions): Promise<VolumeCreateResponseInfo>;
     export function deleteVolume(volumeName: string, options?: VolumeDeleteOptions): Promise<void>;
+
+    export function createPod(
+      selectedProvider: ProviderContainerConnectionInfo | ContainerProviderConnection,
+      podOptions: PodCreateOptions,
+    ): Promise<{ engineId: string; Id: string }>;
```
benoitf: it looks like we should have the provider being an optional field of the options like for createVolume

benoitf: it would take the default running engine if omitted

packages/main/src/plugin/container-registry.ts
lstocchi: `getMatchingEngineFromConnection` returns the api but i also need the id of the matching provider

lstocchi: Do i update that to return the `ProviderContainerConnectionInfo`?

benoitf: you can introduce another method that would return the provider like `getMatchingProviderFromConnection`
and `getMatchingEngineFromConnection` would call this method and return `.api`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2280,6 +2293,17 @@ declare module '@podman-desktop/api' {
     export function listVolumes(): Promise<VolumeListInfo[]>;
     export function createVolume(options?: VolumeCreateOptions): Promise<VolumeCreateResponseInfo>;
     export function deleteVolume(volumeName: string, options?: VolumeDeleteOptions): Promise<void>;
+
+    export function createPod(
+      podOptions: PodCreateOptions,
+      selectedProvider?: ProviderContainerConnectionInfo | ContainerProviderConnection,
```
benoitf: `provider` should be inside PodCreateOptions (maybe my previous comment has been lost)

lstocchi: ops i missed it. Done :+1:

packages/main/src/plugin/container-registry.ts
```diff
@@ -891,6 +902,21 @@ export class ContainerProviderRegistry {
     return matchingContainerProvider.api;
   }
 
+  protected getMatchingContainerProvider(
```
benoitf, line 907: I thought this method should use code from getMatchingEngineFromConnection and getMatchingEngineFromConnection should call this new method and just return .api no ?

packages/main/src/plugin/container-registry.ts
```diff
@@ -876,6 +875,18 @@ export class ContainerProviderRegistry {
     ];
   }
 
+  public getFirstRunningPodmanContainerProvider(): InternalContainerProvider {
```
benoitf, line 884: I think it should be using getFirstRunningConnection code except that it does not return .api at the end


my worries are that here for example you're not sorting provider so podman might not be first while if we call getFirstRunningConnection it would have been first

AFAIK the only difference is returning `.api` or not

benoitf (commented):
`getFirstRunningConnection` and `getFirstRunningPodmanContainerProvider` should call them each other and the same for `getMatchingEngineFromConnection` that should call `getMatchingProviderFromConnection`
so we don't duplicate lines but calls each other

packages/main/src/plugin/container-registry.ts
lstocchi: AFAIK podman is the only one that has a defined libpodApi and i was using this one to find the running container provider.

lstocchi: In this case i'm only interested in a podman one so if you are using docker it throws

packages/main/src/plugin/container-registry.ts
lstocchi: done 👍

packages/main/src/plugin/container-registry.ts
benoitf: Ok I see but then it should contain some comments saying that the other method can't be used or then we name method with Podman to be sure that we check podmanApi and not podApi

benoitf: Ok I see, maybe some comments

benoitf (approved):
Thanks 🙏

lstocchi:
@benoitf thanks to you 👍

benoitf (approved):

PR #5663, opened 2024-01-24
fix: image usage by containers

### LINKED ISSUE #5583: All copies of an image are flagged as in use if only one of them is in use
### Bug description

If the user has multiple copies of the same image and one of the copies is used in a container then all the copies will be flagged as in use by the system. This prevents the user from deleting any of the redundant copies.

https://github.com/containers/podman-desktop/assets/106525396/6a3fac73-403c-4a26-b100-6a3992ed8735

### Operating system

macOS

### Installation Method

None

### Version

1.6.3

### Steps to reproduce

1. Pull an image.
2. Rename the image pulled previously.
3. Pull the image again.
4. Start a container with either image copy.
5. Notice both image copies are flagged as in use.

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Image usage should only match if the image tag is matching as well

### Screenshot / video of UI

image

fixes https://github.com/containers/podman-desktop/issues/5583

### How to test this PR?

Unit tests

try the case reported by the issue (or another usage)

### DIFF
--- a/packages/renderer/src/lib/image/image-utils.spec.ts
+++ b/packages/renderer/src/lib/image/image-utils.spec.ts
@@ -21,6 +21,7 @@ import { ImageUtils } from './image-utils';
 import type { ImageInfo } from '../../../../main/src/plugin/api/image-info';
 import { ContextUI } from '../context/context';
 import type { ViewInfoUI } from '../../../../main/src/plugin/api/view-info';
+import type { ContainerInfo } from '../../../../main/src/plugin/api/container-info';
 
 let imageUtils: ImageUtils;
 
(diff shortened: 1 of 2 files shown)


cbr7:
@benoitf I checked the fix on this branch and the original issue is indeed fixed, however I've noticed a somewhat related problem if the user uses the Prune functionality on the duplicate images, while individual copies can now be deleted, the prune bulk action still sees the copies as being in use. I've attached a video of the behavior.


https://github.com/containers/podman-desktop/assets/106525396/6ecff6e3-ca06-48fc-a1f7-3d584dae786d

benoitf:
@cbr7 we delegate the prune operation to the container engine (using the REST API)

it won't prune the image as it's used. From the container engine POV, there is only one image with multiple tags.
We display one line per tag for easiness but pruning the image is either removing the image or not, not untagging some unused tags

cbr7 (approved):

PR #5689, opened 2024-01-25
refactor: making TaskManager responsible of creating its entry in StatusBarRegistry

axel7083:
### What does this PR do?

Related to https://github.com/containers/podman-desktop/issues/4243 (part 1)

> See https://github.com/containers/podman-desktop/pull/4244#pullrequestreview-1843650350 for details

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -398,8 +398,6 @@ export class PluginSystem {
     const iconRegistry = new IconRegistry(apiSender);
     const directories = new Directories();
     const statusBarRegistry = new StatusBarRegistry(apiSender);
-    const taskManager = new TaskManager(apiSender, statusBarRegistry);
-    const notificationRegistry = new NotificationRegistry(apiSender, taskManager);
 
     const configurationRegistry = new ConfigurationRegistry(apiSender, directories);
     notifications.push(...configurationRegistry.init());
@@ -413,6 +411,9 @@ export class PluginSystem {
     const exec = new Exec(proxy);
 
     const commandRegistry = new CommandRegistry(apiSender, telemetry);
+    const taskManager = new TaskManager(apiSender, statusBarRegistry, commandRegistry);
+
+    const notificationRegistry = new NotificationRegistry(apiSender, taskManager);
     const menuRegistry = new MenuRegistry(commandRegistry);
     const kubeGeneratorRegistry = new KubeGeneratorRegistry();
     const certificates = new Certificates();
@@ -494,10 +495,6 @@ export class PluginSystem {
       'troubleshooting',
       undefined,
     );
-
-    commandRegistry.registerCommand('show-task-manager', () => {
-      apiSender.send('toggle-task-manager', '');
-    });
 
     statusBarRegistry.setEntry(
       'feedback',
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -484,18 +484,6 @@ export class PluginSystem {
     statusBarRegistry.setEntry('help', false, 0, undefined, 'Help', 'fa fa-question-circle', true, 'help', undefined);
 
     statusBarRegistry.setEntry(
-      'tasks',
-      false,
-      0,
-      undefined,
-      'Tasks',
-      'fa fa-bell',
-      true,
-      'show-task-manager',
-      undefined,
-    );
-
-    statusBarRegistry.setEntry(
       'troubleshooting',
       false,
       0,
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -397,8 +397,8 @@ export class PluginSystem {
 
     const iconRegistry = new IconRegistry(apiSender);
     const directories = new Directories();
-
-    const taskManager = new TaskManager(apiSender);
+    const statusBarRegistry = new StatusBarRegistry(apiSender);
+    const taskManager = new TaskManager(apiSender, statusBarRegistry);
     const notificationRegistry = new NotificationRegistry(apiSender, taskManager);
 
     const configurationRegistry = new ConfigurationRegistry(apiSender, directories);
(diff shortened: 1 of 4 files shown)


benoitf (commented):
would you include the command in this PR or another one

as TaskManager should be responsible as well for the command that it's registering in the status bar

'show-task-manager'
```

    commandRegistry.registerCommand('show-task-manager', () => {
      apiSender.send('toggle-task-manager', '');
    });

```

benoitf (commented):
need a test checking we create a status bar entry


also based on some other patterns, probably add the setEntry in a init() method rather than in the constructor

it's easier for mocking/testing

axel7083:
> as TaskManager should be responsible as well for the command that it's registering in the status bar

This would create a circular dependency

`CommandRegistry` requires `Telemetry`
`Telemetry `requires `ConfigurationRegistry`
`ConfigurationRegistry` requires `NotificationRegistry`
`NotificationRegistry` requires `TaskManager`

Therefore, `TaskManager` cannot requires `CommandRegistry`

benoitf:
> > as TaskManager should be responsible as well for the command that it's registering in the status bar
> 
> This would create a circular dependency
> 
> `CommandRegistry` requires `Telemetry` `Telemetry `requires `ConfigurationRegistry` `ConfigurationRegistry` requires `NotificationRegistry` `NotificationRegistry` requires `TaskManager`
> 
> Therefore, `TaskManager` cannot requires `CommandRegistry`

ok we need to break the dependency ConfigurationRegistry -> NotificationRegistry

axel7083:
> ok we need to break the dependency ConfigurationRegistry -> NotificationRegistry

How ? We have a powerful direct usage 
https://github.com/containers/podman-desktop/blob/1b370f67f13a5938e025644a35e1fea0e10c1205/packages/main/src/plugin/configuration-registry.ts#L161-L168

benoitf:
one proposal could be: (I'm sure there are many others)

- Create an empty array of notifications in index.ts
- we call init method of ConfigurationRegistry that returns a notification (or none) and we add that to the array
- when we create NotificationRegistry, we give all existing notifications to the object (and clear the array / make it invalid, whatever for subsequent push/addcalls)

axel7083:
> one proposal could be: (I'm sure there are many others)
> 
> * Create an empty array of notifications in index.ts
> * we call init method of ConfigurationRegistry that returns a notification (or none) and we add that to the array
> * when we create NotificationRegistry, we give all existing notifications to the object (and clear the array / make it invalid, whatever for subsequent push/addcalls)

I feel this fall out of the scope of this PR

axel7083:
> one proposal could be: (I'm sure there are many others)
> 
> * Create an empty array of notifications in index.ts
> * we call init method of ConfigurationRegistry that returns a notification (or none) and we add that to the array
> * when we create NotificationRegistry, we give all existing notifications to the object (and clear the array / make it invalid, whatever for subsequent push/addcalls)

rebased with dependency broken

packages/main/src/plugin/index.ts
```diff
@@ -413,6 +411,9 @@ export class PluginSystem {
     const exec = new Exec(proxy);
 
     const commandRegistry = new CommandRegistry(apiSender, telemetry);
+    const taskManager = new TaskManager(apiSender, statusBarRegistry, commandRegistry);
+
```
benoitf, line 416: here we should call the init method

packages/main/src/plugin/task-manager.ts
```diff
@@ -28,7 +30,32 @@ export class TaskManager {
 
   private tasks = new Map<string, Task>();
 
-  constructor(private apiSender: ApiSenderType) {}
+  constructor(
+    private apiSender: ApiSenderType,
+    private statusBarRegistry: StatusBarRegistry,
+    private commandRegistry: CommandRegistry,
+  ) {
+    this.init();
```
benoitf: constructor should not call methods to match existing patterns

packages/main/src/plugin/task-manager.spec.ts
```diff
@@ -164,7 +174,7 @@ test('return true if notificationTask', async () => {
 });
 
 test('return false if it is not a notificationTask', async () => {
-  const taskManager = new TaskManager(apiSender);
+  const taskManager = new TaskManager(apiSender, statusBarRegistry, commandRegistry);
   const task = taskManager.createTask('title');
   const result = taskManager.isNotificationTask(task);
   expect(result).toBeFalsy();
```
benoitf, line 185: missing tests that init method is registering a command and registering a status bar entry

packages/main/src/plugin/task-manager.ts
```diff
@@ -19,6 +19,8 @@
 import type { ApiSenderType } from './api.js';
 import type { NotificationInfo } from './api/notification.js';
 import type { NotificationTask, StatefulTask, Task } from './api/task.js';
+import type { StatusBarRegistry } from './statusbar/statusbar-registry.js';
```
benoitf, line 22: here you didn't use `/@/` import but for the next line yes so it's looking odd

benoitf (commented):
did a review yesterday but forgot to hit the 'submit' button...

packages/main/src/plugin/index.ts
axel7083: Done

packages/main/src/plugin/task-manager.ts
axel7083: Agreed

packages/main/src/plugin/task-manager.ts
axel7083: My IDE is a crazy guy, I removed it

packages/main/src/plugin/task-manager.spec.ts
axel7083: Added

packages/main/src/plugin/task-manager.ts
```diff
@@ -28,7 +30,30 @@ export class TaskManager {
...
-  constructor(private apiSender: ApiSenderType) {}
+  constructor(
+    private apiSender: ApiSenderType,
+    private statusBarRegistry: StatusBarRegistry,
+    private commandRegistry: CommandRegistry,
+  ) {}
+
+  public init() {
+    // The TaskManager is responsible for creating the entry he will be using
+    this.statusBarRegistry.setEntry(
```
benoitf, line 41: not for this PR but probably would be a nice enhancement on statusBarRegistry to return the Entry when calling setEntry method

benoitf (approved):

PR #5691, opened 2024-01-25
feat: make TaskManager using active state of StatusBarItem

axel7083:
### What does this PR do?

Requires https://github.com/containers/podman-desktop/pull/5690

Related to https://github.com/containers/podman-desktop/issues/4243 (part 3)

> See https://github.com/containers/podman-desktop/pull/4244#pullrequestreview-1843650350 for details

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -490,6 +490,7 @@ export class PluginSystem {
     );
 
     commandRegistry.registerCommand('show-task-manager', () => {
+      taskManager.clearDotted();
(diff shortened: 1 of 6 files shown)


deboer-tim:
My comment from #4244 still stands :), I would like 'dotted' renamed to what it is (a flag that the user should be notified), not tied to what the current UI does.

i.e. if tomorrow UX decided that notifications should pulse the icon instead, we shouldn't need to change anything in /main.

axel7083:
Closing as outdated

PR #5702, opened 2024-01-26
refactor: dialog box moving from 'no' to 'cancel'

### LINKED ISSUE #5446: Dialogs should always use "Cancel" and "Yes" rather than "Yes" and "No".

When a dialog is using Cancel, it will show a different style, example:

Screenshot 2024-01-05 at 5 07 12 PM

However, throughout PD we have "No" which has a different style:

Screenshot 2024-01-05 at 5 07 31 PM

Go through everything that uses showMessageBox and update it to use "Cancel" instead of "No".

_No response_

### Additional context

_No response_

trmendes:
Move from No to Cancel to keep a standard approach when showing users a dialog

### What does this PR do?

The PR proposes using the word "cancel" instead of the word "no".

### Screenshot / video of UI

no screenshots

Fixes #5446

### How to test this PR?

1) Check the dialogs boxes and verify if they contains "Cancel" instead of "No"

### DIFF
--- a/packages/renderer/src/lib/engine/Prune.svelte
+++ b/packages/renderer/src/lib/engine/Prune.svelte
@@ -20,7 +20,7 @@ async function openPruneDialog(): Promise<void> {
   const result = await window.showMessageBox({
     title: 'Prune',
     message: message,
-    buttons: ['Yes', 'No'],
+    buttons: ['Yes', 'Cancel'],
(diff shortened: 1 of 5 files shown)


benoitf (approved):

benoitf:
thanks @trmendes

PR #5703, opened 2024-01-26
fix: removing dependency between ConfigurationRegistry and NotificationRegistry

axel7083:
### What does this PR do?

Remove the direct dependencies between the ConfigurationRegistry and NotificationRegistry.

Reference: https://github.com/containers/podman-desktop/pull/5689#issuecomment-1910333126

### How to test this PR?

- [x] Unit tests have been adapted

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -377,6 +377,8 @@ export class PluginSystem {
 
   // initialize extension loader mechanism
   async initExtensions(): Promise<ExtensionLoader> {
+    const notifications: NotificationCardOptions[] = [];
+
     this.isReady = false;
     this.uiReady = false;
     this.ipcHandle('extension-system:isReady', async (): Promise<boolean> => {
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/index.ts
```diff
@@ -445,6 +447,9 @@ export class PluginSystem {
       await trayIconColor.init();
     }
 
+    // Add all notifications to notification registry
+    notifications.forEach(notification => notificationRegistry.addNotification(notification));
+
```
```suggestion
notifications.length = 0;
Object.freeze(notifications);
```

probably need to make notifications immutable after we've used the array content

packages/main/src/plugin/index.ts
```diff
@@ -445,6 +447,10 @@ export class PluginSystem {
       await trayIconColor.init();
     }
 
+    // Add all notifications to notification registry
+    notifications.forEach(notification => notificationRegistry.addNotification(notification));
+notifications.length = 0;
+Object.freeze(notifications);
```
axel7083:
```suggestion
    notifications.forEach(notification => notificationRegistry.addNotification(notification));
    notifications.length = 0;
    Object.freeze(notifications);
```

benoitf (approved):

PR #5600, opened 2024-01-19
docs: accessing Podman from another WSL instance (install, group, uri, test)

themr0c:
### What does this PR do?

fixes #5342 

On Windows, access your Podman Desktop containers from another WSL distribution.

The central user story focuses on the most common context:

- Default Ubuntu WSL
- Default Podman Machine.

Variations are displayed as foldable details rather than sections, to minimize visual distractions for users in the "default" scenario:

- Custom WSL (possible differences: group name, Podman client version).
- Custom Podman Machine (possible differences: rootless, socket path).

### Screenshot / video of UI

Firefox_Screenshot_2024-01-24T09-23-31 057Z

### What issues does this PR fix or reference?

fixes #5342 

### How to test this PR?

Run the procedures on a Windows host.

PR #5706, opened 2024-01-26
docs: added _Accessing Podman from another WSL instance_ (config, verify)

### LINKED ISSUE #5342: Doc: Adding documentation to how to access podman from other WSL instances

In some cases the user will have multiple WSL instances.
It is possible to make podman available in the other WSL instances: https://github.com/containers/podman/issues/16660 

Document how the solution works and what the user has to do.

This would have to be checked:
https://github.com/containers/podman/issues/15190#issuecomment-1661548692
https://github.com/containers/podman/pull/19705

_No response_

### Additional context

_No response_

themr0c:
### What does this PR do?

docs: added _Accessing Podman from another WSL instance_

Alternate outline than #5600

fixes #5342

### DIFF
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -33,23 +33,26 @@ In foldable details, you can find alternative steps for least common contexts:
 1. To communicate with the remote Podman Machine, you need a Podman client.
 
    To benefit from the latest features, such as `podman kube` subcommands, use a recent Podman version rather than the `podman` package from the distribution.
-   
+
    The Podman client is available with a full `podman` installation or with the `podman-remote` version 4.x and higher. On Ubuntu it is generally easier to install `podman-remote`.
-   
+
    With `podman-remote` you also enable the remote mode by default.
 
    Check for the latest release which includes the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
-   
+
    Download and unpack the binary:
-   
+
    ```shell-session
    $ wget https://github.com/containers/podman/releases/download/v4.9.1/podman-remote-static-linux_amd64.tar.gz
    $ sudo tar -C /usr/local -xzf podman-remote-static-linux_amd64.tar.gz
+   ```
 
    Make this executable as `podman` with the following addition to `.bashrc`:
 
-
+   ```shell-session
    $ export PATH="$PATH:/usr/local/bin"
+   ```
+
 1. Configure the Podman client in your WSL distribution to communicate with the remote Podman machine defined by Podman Desktop.
 
    This will ensure consistency when you are working with Podman from all your different environments
@@ -69,11 +72,11 @@ Two parameters can change:
 - The machine name might differ from `podman-machine-default`.
 - The socket name is different when the Podman machine has root privileges disabled (rootless mode).
 
-   Find your Podman Machine name and connection path:
+  Find your Podman Machine name and connection path:
    </summary>
    <div>
 
-   1. Identify the sockets available in your WSL distribution.
+  1.  Identify the sockets available in your WSL distribution.
 
       The Podman machine shares sockets in a `/mnt/wsl/podman-sockets/` subdirectory named after the Podman machine name.
 
@@ -95,7 +98,7 @@ Two parameters can change:
       /mnt/wsl/podman-sockets/podman-machine-default/podman-user.sock
       ```
 
-   2. Identify the socket that Podman Desktop uses.
+  2.  Identify the socket that Podman Desktop uses.
 
       Podman Desktop defaults to rootful Podman.
       However, consider identifying the active socket.
@@ -124,7 +127,7 @@ Two parameters can change:
       podman-machine-default-root ssh://root@127.0.0.1:59292/run/podman/podman.sock C:\Users\Podman Desktop User\.ssh\podman-machine-default true
       ```
 
-   3. To define the Podman machine remote destination, prepend with `unix://` the socket path that is available in your WSL, and corresponds to the Podman Desktop active socket:
+  3.  To define the Podman machine remote destination, prepend with `unix://` the socket path that is available in your WSL, and corresponds to the Podman Desktop active socket:
 
       For the default Podman machine:
 
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -46,6 +46,10 @@ In foldable details, you can find alternative steps for least common contexts:
    $ wget https://github.com/containers/podman/releases/download/v4.9.1/podman-remote-static-linux_amd64.tar.gz
    $ sudo tar -C /usr/local -xzf podman-remote-static-linux_amd64.tar.gz
 
+   Make this executable as `podman` with the following addition to `.bashrc`:
+
+
+   $ export PATH="$PATH:/usr/local/bin"
 1. Configure the Podman client in your WSL distribution to communicate with the remote Podman machine defined by Podman Desktop.
 
    This will ensure consistency when you are working with Podman from all your different environments
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -115,15 +115,6 @@ In foldable details, you can find alternative steps for least common contexts:
    </div>
    </details>
 
-1. Podman requires you to actively confirm that you want to use a remote connection with the `--remote` flag.
-
-   To always enable the remote connection, edit the `~/.config/containers/containers.conf`, and add the section:
-
-   ```toml
-   [engine]
-   remote=true
-   ```
-
 1. The communication channel between your WSL distribution and the Podman Machine is a special file (a socket).
    The Podman Machine creates this file with specific permissions.
    To communicate with the Podman Machine from your WSL distribution your user must have write permissions for the socket.
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -64,6 +64,11 @@ In foldable details, you can find alternative steps for least common contexts:
    <summary>
    On a custom Podman Machine, the remote Podman Machine destination might be different.
 
+Two parameters can change:
+
+- The machine name might differ from `podman-machine-default`.
+- The socket name is different when the Podman machine has root privileges disabled (rootless mode).
+
    Find your Podman Machine name and connection path:
    </summary>
    <div>
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -32,28 +32,9 @@ In foldable details, you can find alternative steps for least common contexts:
 
 1. To communicate with the remote Podman Machine, you need a Podman client.
 
-   Install the Podman client:
+   To benefit from the latest features, such as `podman kube` subcommands, and to enable the remote mode by default, consider installing the `podman-remote` binary rather than the `podman` package from the distribution.
 
-   ```shell-session
-   $ sudo apt update
-   $ sudo apt install podman
-   ```
-
-   The package might set up local Podman resources.
-   However, you plan not to use them.
-
-   :::tip
-
-   When the Podman client is already installed, you might skip this step.
-
-   :::
-
-   :::info
-
-   On a custom WSL distribution, the installation step might be different.
-   See [Installing Podman on Linux](https://podman.io/docs/installation#installing-on-linux).
-
-   :::
+   Download the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
 
 1. By default, the Podman client in your WSL distribution communicates with the local Podman.
    It should rather communicate with the remote Podman machine defined by Podman Desktop.
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -54,7 +54,7 @@ In foldable details, you can find alternative steps for least common contexts:
 
    This will ensure consistency when you are working with Podman from all your different environments
 
-   Set the default Podman system connection to your Podman Machine:
+   Set the default Podman system connection to your Podman Machine (assuming Podman Desktop is configured with the default of Podman Machine enabled with root privileges):
 
    ```shell-session
    $ podman system connection add --default podman-machine-default-root unix:///mnt/wsl/podman-sockets/podman-machine-default/podman-root.sock
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -148,7 +148,7 @@ In foldable details, you can find alternative steps for least common contexts:
 
 1. The communication channel between your WSL distribution and the Podman Machine is a special file (a socket).
    The Podman Machine creates this file with specific permissions.
-   On your WSL distribution, to communicate with the Podman Machine, your user must have the write permissions on this file.
+   To communicate with the Podman Machine from your WSL distribution your user must have write permissions for the socket.
 
    To give access to the remote Podman machine to your user: create the group if necessary, assign group membership, and exit your session on the WSL distribution to apply the new group membership:
 
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -32,9 +32,19 @@ In foldable details, you can find alternative steps for least common contexts:
 
 1. To communicate with the remote Podman Machine, you need a Podman client.
 
-   To benefit from the latest features, such as `podman kube` subcommands, and to enable the remote mode by default, consider installing the `podman-remote` binary rather than the `podman` package from the distribution.
+   To benefit from the latest features, such as `podman kube` subcommands, use a recent Podman version rather than the `podman` package from the distribution.
+   
+   The Podman client is available with a full `podman` installation or with the `podman-remote` version 4.x and higher. On Ubuntu it is generally easier to install `podman-remote`.
+   
+   With `podman-remote` you also enable the remote mode by default.
 
-   Download the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
+   Check for the latest release which includes the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
+   
+   Download and unpack the binary:
+   
+   ```shell-session
+   $ wget https://github.com/containers/podman/releases/download/v4.9.1/podman-remote-static-linux_amd64.tar.gz
+   $ sudo tar -C /usr/local -xzf podman-remote-static-linux_amd64.tar.gz
 
 1. Configure the Podman client in your WSL distribution to communicate with the remote Podman machine defined by Podman Desktop.
 
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -202,13 +202,13 @@ Verify that, on your WSL distribution, the Podman CLI communicates with your Pod
    - Rootful Podman:
 
      ```shell-session
-     $ grep :10: /etc/group
+     $ getent group 10
      ```
 
    - Rootless Podman:
 
      ```shell-session
-     $ grep :1000: /etc/group
+     $ getent group 1000
      ```
 
    </div>
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -55,28 +55,6 @@ In foldable details, you can find alternative steps for least common contexts:
 
    :::
 
-1. The communication channel between your WSL distribution and the Podman Machine is a special file (a socket).
-   The Podman Machine creates this file with specific permissions.
-   On your WSL distribution, to communicate with the Podman Machine, your user must have the write permissions on this file.
-
-   To give access to the remote Podman machine to your user, assign group membership:
-
-   ```shell-session
-   $ sudo usermod --append --groups 10 your_user
-   ```
-
-1. Exit your session on the WSL distribution to apply the new group membership:
-
-   ```shell-session
-   $ exit
-   ```
-
-1. Start a new WSL session:
-
-   ```shell-session
-   > wsl
-   ```
-
 1. By default, the Podman client in your WSL distribution communicates with the local Podman.
    It should rather communicate with the remote Podman machine defined by Podman Desktop.
 
@@ -159,19 +137,26 @@ In foldable details, you can find alternative steps for least common contexts:
    </div>
    </details>
 
-1. Podman requires you to actively confirm that you want to use a remote connection.
+1. Podman requires you to actively confirm that you want to use a remote connection with the `--remote` flag.
 
-   To always enable the remote connection, set an alias:
+   To always enable the remote connection, edit the `~/.config/containers/containers.conf`, and add the section:
 
-   ```shell-session
-   $ alias podman='podman --remote'
+   ```toml
+   [engine]
+   remote=true
    ```
 
-   :::tip
+1. The communication channel between your WSL distribution and the Podman Machine is a special file (a socket).
+   The Podman Machine creates this file with specific permissions.
+   On your WSL distribution, to communicate with the Podman Machine, your user must have the write permissions on this file.
 
-   On a custom WSL distribution with a Podman client version 4, you might skip this step.
+   To give access to the remote Podman machine to your user: create the group if necessary, assign group membership, and exit your session on the WSL distribution to apply the new group membership:
 
-   :::
+   ```shell-session
+   $ sudo groupadd --gid 10 --force podman
+   $ sudo usermod --append --groups 10 $(whoami)
+   $ exit
+   ```
 
 ## Testing the connection
 
@@ -233,12 +218,6 @@ Verify that, on your WSL distribution, the Podman CLI communicates with your Pod
 
    ```shell-session
    $ podman system connection list
-   ```
-
-1. Verify that your session has an `alias` for `podman --remote`:
-
-   ```shell-session
-   $ alias | grep podman
    ```
 
 1. Verify that Podman has a `Server` version corresponding to your Podman Machine version:
--- a/website/docs/podman/accessing-podman-from-another-wsl-instance.md
+++ b/website/docs/podman/accessing-podman-from-another-wsl-instance.md
@@ -1,0 +1,288 @@
+---
+sidebar_position: 60
+title: Accessing Podman from another WSL distribution
+description: On Windows, access your Podman Desktop containers from another Windows Subsystem for Linux (WSL) distribution.
+---
+
+# Accessing Podman from another WSL distribution
+
+On Windows, [Podman Desktop configures Podman in a dedicated Windows Subsystem for Linux (WSL) distribution: the Podman Machine](/docs/podman/creating-a-podman-machine.md).
+It also configures the Windows Podman client to communicate with the Podman Machine.
+However, it does not configure your other WSL distributions.
+
+You might have other WSL distributions running, and want to access from there to your Podman Desktop containers.
+
+This tutorial focuses on the most common context to walk you through the steps to configure your WSL distribution:
+
+- Ubuntu distribution of Linux.
+- Default Podman Machine.
+
+In foldable details, you can find alternative steps for least common contexts:
+
+- Custom WSL distribution.
+- Custom Podman Machine.
+
+## Configuring your WSL distribution
+
+1. Start a session in your WSL distribution:
+
+   ```shell-session
+   > wsl
+   ```
+
+1. To communicate with the remote Podman Machine, you need a Podman client.
+
+   Install the Podman client:
+
+   ```shell-session
+   $ sudo apt update
+   $ sudo apt install podman
+   ```
+
+   The package might set up local Podman resources.
+   However, you plan not to use them.
+
+   :::tip
+
+   When the Podman client is already installed, you might skip this step.
+
+   :::
+
+   :::info
+
+   On a custom WSL distribution, the installation step might be different.
+   See [Installing Podman on Linux](https://podman.io/docs/installation#installing-on-linux).
+
+   :::
+
+1. The communication channel between your WSL distribution and the Podman Machine is a special file (a socket).
+   The Podman Machine creates this file with specific permissions.
+   On your WSL distribution, to communicate with the Podman Machine, your user must have the write permissions on this file.
+
+   To give access to the remote Podman machine to your user, assign group membership:
+
+   ```shell-session
+   $ sudo usermod --append --groups 10 your_user
+   ```
+
+1. Exit your session on the WSL distribution to apply the new group membership:
+
+   ```shell-session
+   $ exit
+   ```
+
+1. Start a new WSL session:
+
+   ```shell-session
+   > wsl
+   ```
+
+1. By default, the Podman client in your WSL distribution communicates with the local Podman.
+   It should rather communicate with the remote Podman machine defined by Podman Desktop.
+
+   Configure the Podman client in your WSL distribution to communicate with the remote Podman machine defined by Podman Desktop.
+
+   This will ensure consistency when you are working with Podman from all your different environments
+
+   Set the default Podman system connection to your Podman Machine:
+
+   ```shell-session
+   $ podman system connection add --default podman-machine-default-root unix:///mnt/wsl/podman-sockets/podman-machine-default/podman-root.sock
+   ```
+
+   <details>
+   <summary>
+   On a custom Podman Machine, the remote Podman Machine destination might be different.
+
+   Find your Podman Machine name and connection path:
+   </summary>
+   <div>
+
+   1. Identify the sockets available in your WSL distribution.
+
+      The Podman machine shares sockets in a `/mnt/wsl/podman-sockets/` subdirectory named after the Podman machine name.
+
+      In your WSL session, list the available sockets:
+
+      ```shell-session
+      $ find /mnt/wsl/podman-sockets/ -name '*.sock'
+      ```
+
+      Each Podman Machine has a socket for:
+
+      - Rootful Podman: `podman-root.sock`
+      - Rootless Podman: `podman-user.sock`
+
+      Sample output:
+
+      ```shell-session
+      /mnt/wsl/podman-sockets/podman-machine-default/podman-root.sock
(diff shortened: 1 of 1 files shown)


themr0c:
@fabricepipart1a Does it work for you?

fabricepipart1a:
What do you do here compared to other PR? Did you change something? Is there a staging website tat I can use to review the change?

themr0c:
@fabricepipart1a In the other pull request, we explored in all directions to reach technical accuracy, and some steps were repeated. In this pull request, I reorganized the content to be more user centric:

- the first section explains how to configure your WSL distribution. You will probably execute it only once

- the second section explains how to verify that everything is working. You might reuse it later for troubleshooting, to rule out the fact that something is wrong with this part of your setup.

themr0c:
You can view a documentation build for each PR: 

1. To view all checks list, click: "Show all checks".
2. To open the website, in the "netlify — See pull request website online" line, click: "Details". (Note: this line appears only after the website has been built).

themr0c:
For the current commit, the page preview is: https://65c2388aa2bbe90a991eace1--podman-desktop-pr.netlify.app/docs/podman/accessing-podman-from-another-wsl-instance

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,288 @@
...
+
+   To always enable the remote connection, set an alias:
+
+   ```shell-session
+   $ alias podman='podman --remote'
+   ```
+
+   :::tip
+
+   On a custom WSL distribution with a Podman client version 4, you might skip this step.
```
fabricepipart1a: I think this is incorrect. I had to do it. I think it would be preferable to propose in the tip:

Alternatively you can edit your `~/.config/containers/containers.conf` and add the section:
```
[engine]
remote=true
```

fabricepipart1a:
I added a comment to remind you that the alias was still needed in podman v4 according to me.
In the configuration section, could you also add a _details_ or change the _tip_ section that would propose to edit the `containers.yaml` as a possible other option instead of creating an alias?

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: I am trying to reproduce this to see if there is a more precise version where it starts working as intended. I will first test with Debian bookworm that has Podman 4.3.1.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,288 @@
...
+   ```shell-session
+   $ exit
+   ```
+
+1. Start a new WSL session:
+
+   ```shell-session
+   > wsl
+   ```
+
```
themr0c: I will move these steps at the end to:

- Have one step less.
- Avoid one exit/reconnect on WSL.

themr0c: Additionally this will require less attention from the user: 


```shell-session
$ sudo usermod --append --groups 10 $(whoami)
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: Indeed, on Debian, with Podman 4.3.1, you need an explicit `--remote` option:

```shell-session
$ groups
debian adm uucp cdrom sudo dip plugdev users
$ podman system connection list
Name                         URI                                                                     Identity    Default
podman-machine-default-root  unix:///mnt/wsl/podman-sockets/podman-machine-default/podman-root.sock              true
$ podman version
Client:       Podman Engine
Version:      4.3.1
API Version:  4.3.1
Go Version:   go1.19.8
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
$ alias podman="podman --remote"
$ podman version
Client:       Podman Engine
Version:      4.3.1
API Version:  4.3.1
Go Version:   go1.19.8
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64

Server:       Podman Engine
Version:      4.9.0
API Version:  4.9.0
Go Version:   go1.21.6
Built:        Wed Jan 24 11:07:27 2024
OS/Arch:      linux/amd64
```

themr0c: On Podman 4.3.1, when you set the default connection to a remote connection, you need to explicitly set the remote flag (--remote or in the configuration file). On Podman 4.9.0 the remote connection is automatically set (implicitly). When I dig in Podman release notes, I don't see a mention about when this feature has been introduced. I know for sure that it is between 4.3.1 and 4.9.0, but I would like to know the exact version.

themr0c: Same thing with Podman 4.9.0 on Kali Linux: you need to explicitly set the `--remote` flag.

computator: I was looking into this recently, and as far as I can tell the only options to trigger remote mode are the config file, the `--remote` flag, the `--url` flag, or the `CONTAINER_HOST|CONTAINER_CONNECTION` environment variables.

https://github.com/containers/podman/blob/caea1f43ed0fdec3fa8785cdb48add99068aceae/cmd/podman/registry/remote.go#L31-L56

(the other flags in the file are hidden ones for docker compatibility afaict)

computator: oh, and also the `--connection` option of course

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: On openSUSE Leap, we need to also create the group:

```
$ sudo usermod --append --groups 10 your_user
usermod: group '10' does not exist
```

themr0c:
```
$ sudo groupadd --gid 10 --force podman
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: On openSUSE Leap, Podman is 4.8.3 and requires the `--remote` flag too.

themr0c: On Oracle Linux, I see we have other issues. Group is already set, but:

```
$ podman system connection list
Name                         URI                                                                     Identity    Default
podman-machine-default-root  unix:///mnt/wsl/podman-sockets/podman-machine-default/podman-root.sock              true
$ podman version
WARN[0000] "/" is not a shared mount, this could cause issues or missing mounts with rootless containers
ERRO[0000] running `/usr/bin/newuidmap 170 0 1000 1 1 100000 65536`: newuidmap: write to uid_map failed: Operation not permitted
Error: cannot set up namespace using "/usr/bin/newuidmap": should have setuid or have filecaps setuid: exit status 1
```

themr0c: Decision in standup: do not document the Oracle Linux edge case.

themr0c:
@fabricepipart1a @computator I implemented your suggestions. can you have a look?

fabricepipart1a (approved):
This is perfect according to me. Concise yet precise.
That will greatly help users.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
+---
+sidebar_position: 60
+title: Accessing Podman from another WSL distribution
+description: On Windows, access your Podman Desktop containers from another Windows Subsystem for Linux (WSL) distribution.
+---
+
+# Accessing Podman from another WSL distribution
+
+On Windows, [Podman Desktop configures Podman in a dedicated Windows Subsystem for Linux (WSL) distribution: the Podman Machine](/docs/podman/creating-a-podman-machine.md).
```
cdrage: Run off sentence, see other comment on the other PR regarding this information.

I feel like this is copy and pasted a lot everywhere

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+However, it does not configure your other WSL distributions.
+
+You might have other WSL distributions running, and want to access from there to your Podman Desktop containers.
+
+This tutorial focuses on the most common context to walk you through the steps to configure your WSL distribution:
+
+- Ubuntu distribution of Linux.
+- Default Podman Machine.
+
+In foldable details, you can find alternative steps for least common contexts:
```
cdrage, line 20: kinda implied by just scrolling down? no need to mention as we have the TOC

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+
+# Accessing Podman from another WSL distribution
+
+On Windows, [Podman Desktop configures Podman in a dedicated Windows Subsystem for Linux (WSL) distribution: the Podman Machine](/docs/podman/creating-a-podman-machine.md).
+It also configures the Windows Podman client to communicate with the Podman Machine.
+However, it does not configure your other WSL distributions.
+
+You might have other WSL distributions running, and want to access from there to your Podman Desktop containers.
+
+This tutorial focuses on the most common context to walk you through the steps to configure your WSL distribution:
```
cdrage, line 15: I dont think this would be needed

cdrage (changes requested):

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+   > wsl
+   ```
+
+1. To communicate with the remote Podman Machine, you need a Podman client.
+
+   Install the Podman client:
+
+   ```shell-session
+   $ sudo apt update
+   $ sudo apt install podman
```
computator: It could be mentioned that you can also use the `podman-remote` binary from the releases page. These instructions are also assuming you are using Ubuntu 22.04 (as opposed to an older version or to another distro), which should be noted somewhere.

If someone is using `podman-remote` then the `[engine]` `remote` parameter is unnecessary and they also get newer features of the CLI released since Podman 3.4.4 such as some of the `podman kube` subcommands.

computator: Actually I see your mention of a custom WSL distro below, but the mention of this being based on Ubuntu could maybe be a little more obvious. Also, when I hear "custom distro" I think of some manually created WSL image instead of just a non-default distro.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+   To always enable the remote connection, edit the `~/.config/containers/containers.conf`, and add the section:
+
+   ```toml
+   [engine]
+   remote=true
+   ```
+
+1. The communication channel between your WSL distribution and the Podman Machine is a special file (a socket).
+   The Podman Machine creates this file with specific permissions.
+   On your WSL distribution, to communicate with the Podman Machine, your user must have the write permissions on this file.
```
computator:
```suggestion
   To communicate with the Podman Machine from your WSL distribution your user must have write permissions for the socket.
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+
+   To give access to the remote Podman machine to your user: create the group if necessary, assign group membership, and exit your session on the WSL distribution to apply the new group membership:
+
+   ```shell-session
+   $ sudo groupadd --gid 10 --force podman
```
computator: This will not work correctly, per the man page it will create a group with a new unique GID if GID 10 already exists. It will not replace it. You would not want to replace it anyway because that might cause issues with other software in the WSL distro that uses that group.
> -f, --force
>           This option causes the command to simply exit with success status if the specified group already exists. When used with -g, **and the specified GID already exists, another (unique) GID is chosen
>           (i.e.  -g is turned off)**.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+   - Rootful Podman:
+
+     ```shell-session
+     $ grep :10: /etc/group
+     ```
+
+   - Rootless Podman:
+
+     ```shell-session
+     $ grep :1000: /etc/group
```
computator: For any distro with libc (I.E. pretty much all WSL distros) you can also use `getent` which works like this:
```
$ getent group 10
uucp:x:10:
$
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+   OS/Arch:      linux/amd64
+   ```
+
+   :::info
+
+   On your environment, the Podman version might be different.
+
+   :::
+
+1. Verify that you can list running containers.
```
computator, line 246: This step doesn't hurt but is unnecessary because the `podman version` command already tests whether it can talk to the remote Podman Machine

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+
+      - Rootful Podman: `unix:///mnt/wsl/podman-sockets/podman-machine-default/podman-root.sock`
+      - Rootless Podman: `unix:///mnt/wsl/podman-sockets/podman-machine-default/podman-user.sock`
+
+   </div>
+   </details>
+
+1. Podman requires you to actively confirm that you want to use a remote connection with the `--remote` flag.
+
+   To always enable the remote connection, edit the `~/.config/containers/containers.conf`, and add the section:
```
computator: A few notes that may or may not be worth mentioning:
- If you are using `podman-remote` this step is unnecessary
- As an alternative to setting this value, you can just set the `CONTAINER_CONNECTION` environment variable to `podman-machine-default-root` (maybe in `.profile` or something)

computator (commented):
I left some comments, but overall this looks fairly good.

It may be worth mentioning that when using this method that any bind mounts created with `-v` (or `--mount` etc) refer to paths in the remote Podman Machine and not paths in the local WSL instance. This means that binds like this one will not work as expected:
```
podman run --rm -it -v /my/project/code:/app/code some-image-name
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
fabricepipart1a: I did nt know about this `podman-remote` alternative. Honestly I find it a bit confusing to have to change of binary. People already have a tendency to do `alias docker=podman`... From a UX point of view, I think I love if the connection had a remote attribute in order to specify it when setting up the connection once for all.

Probably something like
```podman system connection add --remote <name> <socket-path>```

Comment (computator) [reply website/docs/podman/accessing-podman-from-another-wsl-instance.md]: Yeah I agree that having a way to set remote as the default would be nice. 

`podman-remote` is a thin client version of Podman that doesn't have any of the code to run containers locally and is just for remote connections or Podman machine.

https://github.com/containers/podman/blob/main/docs/tutorials/remote_client.md
Although the docs describe using SSH, it doesn't actually need SSH for connections on the same host (such as in WSL) and can just use the sockets directly.

File: website/docs/podman/accessing-podman-from-another-wsl-instance.md (hunk 4 above)
Comment (themr0c) [reply website/docs/podman/accessing-podman-from-another-wsl-instance.md:20]: In docs, never leave something implicit.

File: website/docs/podman/accessing-podman-from-another-wsl-instance.md (hunk 5 above)
Comment (themr0c) [reply website/docs/podman/accessing-podman-from-another-wsl-instance.md:15]: We need to be explicit about the scope.

File: website/docs/podman/accessing-podman-from-another-wsl-instance.md (hunk 6 above)
Comment (themr0c) [reply website/docs/podman/accessing-podman-from-another-wsl-instance.md]: Reading these comments, it seems indeed that the features in `podman-remote` make the case strong enough to cover `podman-remote` rather than the (old) `podman` package.

We might keep mention of the installation with the package for environments where	installing binaries must be avoided.

File: website/docs/podman/accessing-podman-from-another-wsl-instance.md (hunk 5 above)
Comment (themr0c) [reply website/docs/podman/accessing-podman-from-another-wsl-instance.md:15]: See: https://github.com/containers/podman-desktop/pull/5706/files#r1487194745

File: website/docs/podman/accessing-podman-from-another-wsl-instance.md (hunk 8 above)
Comment (themr0c) [reply website/docs/podman/accessing-podman-from-another-wsl-instance.md]: My mistake, I assumed the `groupadd` was a standard and mature package that doesn't change, and therefore that the options would be the same as on a recent Fedora. It is not the case. I know: never assume.

```
  -f, --force                   exit successfully if the group already exists,
                                and cancel -g if the GID is already used
```

File: website/docs/podman/accessing-podman-from-another-wsl-instance.md (hunk 9 above)
Comment (themr0c) [reply website/docs/podman/accessing-podman-from-another-wsl-instance.md]: Oh nice indeed! Way better.

File: website/docs/podman/accessing-podman-from-another-wsl-instance.md [hunk 12]
```diff
@@ -0,0 +1,267 @@
...
+
+   - Rootful Podman: GID `10` name is `uucp`.
+   - Rootless Podman: GID `1000` name is the same as the user name you chose when creating the WSL machine.
+
+   On a custom WSL distribution, find the group name for:
+
+   - Rootful Podman:
+
+     ```shell-session
+     $ grep :10: /etc/group
```
themr0c:
```suggestion
     $ getent group 10
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,267 @@
...
+     ```
+
+   - Rootless Podman:
+
+     ```shell-session
+     $ grep :1000: /etc/group
```
themr0c:
```suggestion
     $ getent group 1000
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: There is a slight uncertainty that subsists. With the previous test, you know that you can communicate with a Podman server that has the desired version, but if you have multiple Podman machines (and a wrong configuration), you might get a reply for the wrong machine.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: This step is not necessary on Ubuntu (I found it necessary only on openSUSE). Let's remove it then. We already do no document additional SELinux steps required for openSUSE.


```suggestion
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c:
```suggestion
On Windows, [Podman Desktop creates a Windows Subsystem for Linux (WSL) virtual machine: the Podman Machine](/docs/podman/creating-a-podman-machine.md).
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: To check:

Is it posssible to run:

```
podman-remote system connection add
```

Or do we still need `podman` to define the connection?

computator: Yes you can define the connection with `podman-remote` the same way as with the full version of `podman`. That's how it's configured in the doc I linked to above.

themr0c: Recent features + consistent across distros + one step less. It might be enough to convince people who would prefer using the package.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,266 @@
...
+   When the Podman client is already installed, you might skip this step.
+
+   :::
+
+   :::info
+
+   On a custom WSL distribution, the installation step might be different.
+   See [Installing Podman on Linux](https://podman.io/docs/installation#installing-on-linux).
+
+   :::
```
themr0c:
```suggestion
   To benefit from the latest features, such as `podman kube` subcommands, and to enable the remote mode by default, consider installing the `podman-remote` binary rather than the `podman` package from the distribution.

   Download the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,247 @@
...
+1. To communicate with the remote Podman Machine, you need a Podman client.
+
+   To benefit from the latest features, such as `podman kube` subcommands, and to enable the remote mode by default, consider installing the `podman-remote` binary rather than the `podman` package from the distribution.
+
+   Download the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
+
+1. By default, the Podman client in your WSL distribution communicates with the local Podman.
+   It should rather communicate with the remote Podman machine defined by Podman Desktop.
+
+   Configure the Podman client in your WSL distribution to communicate with the remote Podman machine defined by Podman Desktop.
```
themr0c:
```suggestion
1. Configure the Podman client in your WSL distribution to communicate with the remote Podman machine defined by Podman Desktop.
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,244 @@
...
+
+1. Podman requires you to actively confirm that you want to use a remote connection with the `--remote` flag.
+
+   To always enable the remote connection, edit the `~/.config/containers/containers.conf`, and add the section:
+
+   ```toml
+   [engine]
+   remote=true
+   ```
+
```
themr0c: Removing. Obsolete with `podman-remote`.

```suggestion
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: Perfect, removed.

themr0c:
@fabricepipart1a @computator @cdrage Adapted to use `podman-remote` rather than `podman`. And I believe  I adressed all commments. Can you do another round of review?

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,235 @@
...
+- Custom WSL distribution.
+- Custom Podman Machine.
+
+## Configuring your WSL distribution
+
+1. Start a session in your WSL distribution:
+
+   ```shell-session
+   > wsl
```
MarkCheshire: Recommend `wsl -d your-distro-name` because likely if people are doing this they want it configured for more than one distro

cdrage (approved):
I'm unable to verify the steps (if someone else can that'd be awesome).

But the textual representation / formatting / flow of the tutorial LGTM!

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,235 @@
...
+   Download the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
+
+1. Configure the Podman client in your WSL distribution to communicate with the remote Podman machine defined by Podman Desktop.
+
+   This will ensure consistency when you are working with Podman from all your different environments
+
+   Set the default Podman system connection to your Podman Machine:
+
+   ```shell-session
+   $ podman system connection add --default podman-machine-default-root unix:///mnt/wsl/podman-sockets/podman-machine-default/podman-root.sock
```
MarkCheshire, line 64: This part is super confusing when you come from Docker. First time through I thought the integration was broken because containers were not visible in Podman Desktop. Then I thought that it was wrong to use Podman rootful as the default. Finally when going through this new version of the docs, I realized that the default Podman Desktop installation is for rootful, but I had changed that to rootless (figuring that is a big part of the benefit of Podman), and I had forgotten.

I recommend an addition in line 43: "Podman Machine (assuming Podman Desktop is configured with the default of Podman Machine enabled with root privileges):"

Then later make it more explicit in troubleshooting and that you can switch between rootful and rootless

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,235 @@
...
+
+   ```shell-session
+   > wsl
+   ```
+
+1. To communicate with the remote Podman Machine, you need a Podman client.
+
+   To benefit from the latest features, such as `podman kube` subcommands, and to enable the remote mode by default, consider installing the `podman-remote` binary rather than the `podman` package from the distribution.
+
+   Download the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
```
MarkCheshire: The experience to complete this step when running on WSL/Ubuntu leaves a lot to be desired. First going to the link in the latest release, it does not show podman-remote as an asset, you have to scroll down and open the fold before you even find it (I had assumed it was a bad link). Then there are multiple binaries with no explanation on which to use nor how to install. I had seen multiple comments in forums stating that `podman-remote-static` was not compatible with Podman Desktop, so I thought it would be simpler to install the full Podman. However this is a waste of time on Ubuntu because 3.x does not support the necessary commands. I spent ages trying to find a way to install Podman 4.x on Ubuntu, but I wish someone had warned me just not to bother. So finally I went back to explore podman-remote. I recommend the following changes to make this is smoother experience:

The Podman client is available with a full `podman ` installation or with the `podman-remote` version 4.x and higher. On Ubuntu it is generally easier to install `podman-remote`

Check for the latest release which includes the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest)
Download and unpack the binary:
`wget https://github.com/containers/podman/releases/download/v4.9.1/podman-remote-static-linux_amd64.tar.gz`
`sudo tar -C /usr/local -xzf podman-remote-static-linux_amd64.tar.gz`
Make this executable as `podman` with the following additions to `.basrc`:
`export PATH="$PATH:/usr/local/bin"`

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,235 @@
...
+
+   :::info
+
+   On your environment, the Podman version might be different.
+
+   :::
+
+1. Verify that you can list running containers.
+
+   [On the Podman Desktop graphical interface, start a container such as `quay.io/podman/hello`](/docs/containers/starting-a-container).
```
MarkCheshire: IMO is it easier to start a container from the command line. Then confirm with `podman ps -a`. Finally verify that the container is visible in Podman Desktop. If it is not visible, then verify that everything is either configured as rootful or as rootless.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,235 @@
...
+
+   On your WSL distribution, list running containers:
+
+   ```shell-session
+   $ podman ps -a
+   ```
+
+   The output lists the same containers that you see on Podman Desktop graphical interface.
+
```
MarkCheshire, line 290: In order to reinforce the behavior change coming from WSL/Docker I recommend finishing with some additional context:

e.g. insert a comment that podman desktop only has visibility to either rootless or rootful containers but not both at the same time. This can be switched with `podman machine set --rootful=false` and `podman machine set --rootful=true` and a corresponding change in the `podman system connection` configuration.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
computator: An additional advantage to starting with user instead of root is that the steps for adding permissions to the socket for GID 10 are no longer necessary (at least for Ubuntu)

themr0c: Following previous suggestions, I removed mention to rootful/rootless Podman in the default user story. This distinction is implicitly present in the details right after. I understand I should add it again there.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,235 @@
...
+   Set the default Podman system connection to your Podman Machine:
+
+   ```shell-session
+   $ podman system connection add --default podman-machine-default-root unix:///mnt/wsl/podman-sockets/podman-machine-default/podman-root.sock
+   ```
+
+   <details>
+   <summary>
+   On a custom Podman Machine, the remote Podman Machine destination might be different.
+
```
themr0c, line 71: @MarkCheshire Would it address your concern with the fact that the default story is for a rootful Podman?

```suggestion

Two parameters can change:

- The machine name might differ from `podman-machine-default`.
- The socket name is different when the Podman machine has root privileges disabled (rootless mode).

```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c: Unfortunately, the default changed from rootless to rootful to address issues specific to the rootless mode, notably when using Kind. It hopefully might change back later.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
themr0c:
```suggestion
   To benefit from the latest features, such as `podman kube` subcommands, use a recent Podman version rather than the `podman` package from the distribution.
   
   The Podman client is available with a full `podman` installation or with the `podman-remote` version 4.x and higher. On Ubuntu it is generally easier to install `podman-remote`.
   
   With `podman-remote` you also enable the remote mode by default.

   Check for the latest release which includes the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
   
   Download and unpack the binary:
   
   ```shell-session
   $ wget https://github.com/containers/podman/releases/download/v4.9.1/podman-remote-static-linux_amd64.tar.gz
   $ sudo tar -C /usr/local -xzf podman-remote-static-linux_amd64.tar.gz
   ```

   Make this executable as `podman` with the following addition to `.bashrc`:

   ```shell-session
   $ export PATH="$PATH:/usr/local/bin"
   ```
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,245 @@
...
+   With `podman-remote` you also enable the remote mode by default.
+
+   Check for the latest release which includes the `podman-remote` binary from the [Podman releases page](https://github.com/containers/podman/releases/latest).
+   
+   Download and unpack the binary:
+   
+   ```shell-session
+   $ wget https://github.com/containers/podman/releases/download/v4.9.1/podman-remote-static-linux_amd64.tar.gz
+   $ sudo tar -C /usr/local -xzf podman-remote-static-linux_amd64.tar.gz
+
```
themr0c, line 49:
```suggestion

   Make this executable as `podman` with the following addition to `.bashrc`:


   $ export PATH="$PATH:/usr/local/bin"
   ```
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,249 @@
...
```
themr0c, line 48:
```suggestion
   ```
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,249 @@
...
+   Make this executable as `podman` with the following addition to `.bashrc`:
+
+
+   $ export PATH="$PATH:/usr/local/bin"
+1. Configure the Podman client in your WSL distribution to communicate with the remote Podman machine defined by Podman Desktop.
+
+   This will ensure consistency when you are working with Podman from all your different environments
+
+   Set the default Podman system connection to your Podman Machine:
```
themr0c:
```suggestion
   Set the default Podman system connection to your Podman Machine (assuming Podman Desktop is configured with the default of Podman Machine enabled with root privileges):
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,254 @@
...
+
+- Custom WSL distribution.
+- Custom Podman Machine.
+
+## Configuring your WSL distribution
+
+1. Start a session in your WSL distribution:
+
+   ```shell-session
+   > wsl
```
themr0c:
```suggestion
   > wsl --distribution your-distribution-name
```

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,290 @@
...
+
+   ```shell-session
+   $ wget https://github.com/containers/podman/releases/download/v4.9.1/podman-remote-static-linux_amd64.tar.gz
+   $ sudo tar -C /usr/local -xzf podman-remote-static-linux_amd64.tar.gz
+   ```
+
+   Make this executable as `podman` with the following addition to `.bashrc`:
+
+   ```shell-session
+   $ export PATH="$PATH:/usr/local/bin"
```
MarkCheshire, line 53: It is missing the alias:
`alias podman='podman-remote-static-linux_amd64'`

themr0c:
```suggestion
   $ export PATH="$PATH:/usr/local/bin"
   $ alias podman='podman-remote-static-linux_amd64'
```

themr0c:
Waiting for a final approval from @MarkCheshire and @fabricepipart1a, and I believe we can merge.

website/docs/podman/accessing-podman-from-another-wsl-instance.md
```diff
@@ -0,0 +1,293 @@
...
+   On your environment, the Podman version might be different.
+
+   :::
+
+1. Verify that you can list running containers.
+
+   On your WSL distribution, start a container such as `quay.io/podman/hello`, and list the name of the last running container:
+
+   ```shell-session
+   $ podman quay.io/podman/hello
```
glsee, line 251: This should be `podman run quay.io/podman/hello` instead. `run` is missing

glsee:
I happened to be looking for solution to accessing podman from another WSL instance and it led me to this PR.

I followed the instructions in the doc file pending review for this PR. Apart from the typo at L251 as commented above, I can verify that it works on my machine.

EDIT: One minor thing though, there is a warning from running `podman run quay.io/podman/hello`. I'm not certain if this is just me or if we need to address this in the doc.
```
WARN[0000] "/run/user/1000/" directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.: stat /run/user/1000/: no such file or directory: Trying to pull image in the event that it is a public image.
```

lordrip:
Hi @themr0c, just out of curiosity, what would be the URL to see these docs?

fabricepipart1a:
Hi @lordrip 
I guess you are looking for that URL?
https://podman-desktop.io/docs/podman/accessing-podman-from-another-wsl-instance

lordrip:
> Hi @lordrip I guess you are looking for that URL? https://podman-desktop.io/docs/podman/accessing-podman-from-another-wsl-instance

Yes, thanks a lot @fabricepipart1a

PR #5709, opened 2024-01-26
feat: image deletion

### LINKED ISSUE #5474: No feedback when deleting an image

deboer-tim:
### What does this PR do?

Adds the 'DELETING' state to images so that a spinner appears while they are being deleted. Follows the identical pattern to pods, deployments, etc:
- Add new status.
- Set the status when deleting and fire an event.
- Propagate event through ColumnActions.
- List/Details trigger UI change.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/19958075/f17c9939-d39e-4c32-af8d-ba1bdb0bf9e7

Fixes #5474.

### How to test this PR?

Create and delete an image.

### DIFF
--- a/packages/renderer/src/lib/image/ImageActions.spec.ts
+++ b/packages/renderer/src/lib/image/ImageActions.spec.ts
@@ -51,13 +51,15 @@ test('Expect showMessageBox to be called when error occurs', async () => {
   showMessageBoxMock.mockResolvedValue({ response: 0 });
   getContributedMenusMock.mockImplementation(() => Promise.resolve([]));
 
+  const image: ImageInfoUI = {
+    name: 'dummy',
+    status: 'UNUSED',
+  } as ImageInfoUI;
+
   render(ImageActions, {
     onPushImage: vi.fn(),
     onRenameImage: vi.fn(),
-    image: {
-      name: 'dummy',
-      status: 'UNUSED',
-    } as unknown as ImageInfoUI,
+    image,
   });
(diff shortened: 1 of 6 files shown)


packages/renderer/src/lib/image/ImageActions.spec.ts
```diff
@@ -19,8 +19,8 @@
 import '@testing-library/jest-dom/vitest';
 import { test, expect, vi, beforeAll } from 'vitest';
 import { render, screen, waitFor, fireEvent } from '@testing-library/svelte';
-import ImageActions from '/@/lib/image/ImageActions.svelte';
-import type { ImageInfoUI } from '/@/lib/image/ImageInfoUI';
+import ImageActions from './ImageActions.svelte';
+import type { ImageInfoUI } from './ImageInfoUI';
```
benoitf: note sure about the why

benoitf (approved):

packages/renderer/src/lib/image/ImageActions.spec.ts
deboer-tim: tbh I just instinctively did a 'that's different and longer than it needs to be' and changed it. Re-pushed without this change.

PR #5712, opened 2024-01-26
fix: improper task state

### LINKED ISSUE #5711: podman-desktop api withProgress not have failure status when report message
### Bug description

When running 

```ts
void podmanDesktopApi.window.withProgress<void>(
  { location: podmanDesktopApi.ProgressLocation.TASK_WIDGET, title: `Pulling ${recipe.name}.` },
  progress => {
    progress.report({message: `Something went wrong.`});
  },
);
```

But inside the TaskManager the Task is marked as completed

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Set the state to failure if user report a message (aka error)

Fixes https://github.com/containers/podman-desktop/issues/5711

### How to test this PR?

- [x] regression tests has been added

### DIFF
--- a/packages/main/src/plugin/progress-impl.spec.ts
+++ b/packages/main/src/plugin/progress-impl.spec.ts
@@ -122,8 +122,8 @@ test('Should create a task and have it in state failure', async () => {
 
   const progress = new ProgressImpl(taskManager);
 
-  await progress.withProgress({ location: ProgressLocation.TASK_WIDGET, title: 'My task' }, async (progress) => {
-    progress.report({ message: 'something wrong'});
+  await progress.withProgress({ location: ProgressLocation.TASK_WIDGET, title: 'My task' }, async progress => {
+    progress.report({ message: 'something wrong' });
   });
 
   expect(updateTaskMock).toHaveBeenLastCalledWith({
@@ -132,4 +132,3 @@ test('Should create a task and have it in state failure', async () => {
     error: 'something wrong',
   });
 });
-
--- a/packages/main/src/plugin/progress-impl.spec.ts
+++ b/packages/main/src/plugin/progress-impl.spec.ts
@@ -109,3 +109,27 @@ test('Should create a task and propagate the result', async () => {
     status: 'success',
   });
 });
+
+test('Should create a task and have it in state failure', async () => {
+  const createTaskMock = vi.fn();
+  const updateTaskMock = vi.fn();
+  const taskManager = {
+    createTask: createTaskMock,
+    updateTask: updateTaskMock,
+  } as unknown as TaskManager;
+
+  createTaskMock.mockImplementation(() => ({}));
+
+  const progress = new ProgressImpl(taskManager);
+
+  await progress.withProgress({ location: ProgressLocation.TASK_WIDGET, title: 'My task' }, async (progress) => {
+    progress.report({ message: 'something wrong'});
+  });
+
+  expect(updateTaskMock).toHaveBeenLastCalledWith({
+    state: 'completed',
+    status: 'failure',
+    error: 'something wrong',
+  });
+});
+
--- a/packages/main/src/plugin/progress-impl.ts
+++ b/packages/main/src/plugin/progress-impl.ts
@@ -101,7 +101,7 @@ export class ProgressImpl {
     )
       .then(value => {
         // Middleware to capture the success of the task
-        t.status = 'success';
+        t.status = t.error?'failure':'success';
         t.state = 'completed';
         // We propagate the result to the caller, so he can use the result
         return value;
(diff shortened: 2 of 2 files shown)


benoitf (commented):
AFAIK from the API POV progress.report is just to report progress on a task, not to report a failure or a success

axel7083:
> AFAIK from the API POV progress.report is just to report progress on a task, not to report a failure or a success

The `message` is already used to display the error message inside the ui, it makes sense that if we set a message, when completed to define the state based on yes or not it has been defined

axel7083:
cc @benoitf I would like to have a final statement on the expected behaviour we want to follow for failing a Task, and display the error message, as _some extensions_ in development (cc @feloy) are currently facing the issue of displaying an error message for the user.

The fix I made in this PR allow an extension to set the state with the error message, _this is easy and nice_. This would make the following code working:

```ts
 void podmanDesktopApi.window.withProgress<void>(
      { location: podmanDesktopApi.ProgressLocation.TASK_WIDGET, title: `Running stuff` },
      async progress => {
        try {
          await complicatedTask(recipe, model);
          // mark the task as completed
          progress.report({ increment: -1 });
        } catch (error) {
          setError(error); // our own error system must be alerted
          progress.report({ message: `Error: ${String(error)}` }); // putting the task in error state
        }
      },
    );
```

Otherwise with the current implementation we need to do something like this

```ts
 void podmanDesktopApi.window.withProgress<void>(
      { location: podmanDesktopApi.ProgressLocation.TASK_WIDGET, title: `Running stuff` },
      async progress => {
        try {
          await complicatedTask(recipe, model);
          // mark the task as completed
          progress.report({ increment: -1 });
        } catch (error) {
          progress.report({ message: `Error: ${String(error)}` }); // putting the task in error state
          throw error; // re-throw the error, as the task need to raise an error to be set as failure and display the error message
        }
      },
    ).catch((error) => {
         setError(error); // our own error system must be alerted
    });
```

The difference is mainly, that we need to throw an error inside the Task to make it fail, this is a bit ugly, I would rather manually set it to error. Because of that behavior I must add a catch clause to avoid long complaining NodeJS traces 

> I would also suggest to add a code example in the docs to keep track of that (I can include in this PR or another when the decision is taken)

benoitf:
Hello @axel7083 

the root problem is that we can't assume that `progress.report` is to report an error
The API clearly state that it's to report a progress not to report an error.

So from an extension I can add messages like
```
progress.report({ message: 'Initializing...'});
progress.report({ message: 'Building..'});
```

it should just update the notification text.

I'm not sure as a user you should write as much as you copied in the two samples

if there is an error during the task, it needs to throw an Error (for example complicatedTask could report the error)

I think what user is expected to write should be something like this instead
```ts
await window.withProgress<void>(
      { location: podmanDesktopApi.ProgressLocation.TASK_WIDGET, title: `Running stuff` },
      async progress => {
          // might throw an error
          await complicatedTask(recipe, model);
          // mark the task as completed
          progress.report({ increment: -1 });
      },
    );
```
then it could be captured by upper error handling that would add the error to your 'own' error handling system

or if you really need for each call:
```ts
try {
podmanDesktopApi.window.withProgress<void>(
      { location: podmanDesktopApi.ProgressLocation.TASK_WIDGET, title: `Running stuff` },
      async progress => {
          // might throw an error
          await complicatedTask(recipe, model);
          // mark the task as completed
          progress.report({ increment: -1 });
      },
    );
} catch (error) {
   setError(error); // our own error system must be alerted
}
```




(error message should be extracted from the error reported by complicatedTask for example)

axel7083:
> Hello @axel7083
> 
> the root problem is that we can't assume that `progress.report` is to report an error The API clearly state that it's to report a progress not to report an error.
> 
> So from an extension I can add messages like
> 
> ```
> progress.report({ message: 'Initializing...'});
> progress.report({ message: 'Building..'});
> ```
> 
> it should just update the notification text.

Thanks @benoitf for the details, however the current implementation _does not_ follow the behaviour you describe, the `message` property is not used to update the task title, but the error message. We need to change that to match the behaviour you describe, do you agree ? (I can open an issue if we agree on the matter)

https://github.com/containers/podman-desktop/blob/1f427287f9221562f18aba54dd61ca920f76d36a/packages/main/src/plugin/progress-impl.ts#L90-L98

benoitf:
yes I think we should change the implementation that is not doing what the contract is instead of trying to introduce an interpretation/change of the API contract. (reporting a message being an error)

axel7083:
Closing as will not be merged

PR #5716, opened 2024-01-27
feat: make localhost a valid domain for env.openExternal calls

dgolovin:
http://localhost is considered a secure context. VS Code link protections always trusts localhost urls.

### What does this PR do?

Adds localhost to allowed domains in security restriction on links. That would let to avoid extra click between login request and login form opened in browser.

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -344,7 +344,7 @@ export class PluginSystem {
 
       // if url is a known domain, open it directly
       const urlObject = new URL(url);
-      const validDomains = ['podman-desktop.io', 'podman.io'];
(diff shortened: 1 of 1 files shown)


deboer-tim (commented):
Should probably do 127.0.0.1 then too?

benoitf (approved):

dgolovin:
@deboer-tim adding 127.0.0.1 looks good to me. @benoitf WDYT?

benoitf:
I'm fine by adding also `127.0.0.1`. It's harmless if we add localhost

feloy (approved):

deboer-tim (approved):

PR #5717, opened 2024-01-28
feat: multi-delete animation

### LINKED ISSUE #5615: No feedback when deleting multiple instances

When you delete 1 of something (except for images, volumes - separate issues open) there is a spinner in the UI to show you that it's in progress. This stops you from trying to delete it again or wondering if your click worked.

When you delete multiple objects, there's no feedback. It's just pause, pause, pause, gone.

Deleting multiple objects should set the same deleting status that deleting a single object does, so that you know it's still in progress.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

When you delete multiple objects at once, the delete button would show as in progress, but each object you're deleting didn't have any spinner like it does when you click the object's delete button (except the containers list, which marked containers for deletion but not pods).

This makes the bulk delete actions a little more consistent:
- exit if there's nothing being deleted
- mark everything as being deleted
- do the deletion

Also fixes a case in container list where the bulk delete wouldn't start (so the button wouldn't go in-progress) until after deleting pods.

### Screenshot / video of UI

Before:

https://github.com/containers/podman-desktop/assets/19958075/9c25d78c-6ed3-483d-8613-a6da4a2c7ce4

After:

https://github.com/containers/podman-desktop/assets/19958075/fc476482-bb19-432a-9d45-30f3336d92e3

Fixes #5615.

### How to test this PR?

Create and delete lots of objects.

### DIFF
--- a/packages/renderer/src/lib/deployments/DeploymentsList.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentsList.svelte
@@ -52,7 +52,7 @@ async function deleteSelectedDeployments() {
       try {
         await window.kubernetesDeleteDeployment(deployment.name);
       } catch (e) {
-        console.log('error while deleting deployment', e);
+        console.error('error while deleting deployment', e);
       }
     }),
   );
--- a/packages/renderer/src/lib/deployments/DeploymentsList.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentsList.svelte
@@ -38,20 +38,25 @@ onMount(() => {
 let bulkDeleteInProgress = false;
 async function deleteSelectedDeployments() {
   const selectedDeployments = deployments.filter(deployment => deployment.selected);
-
-  if (selectedDeployments.length > 0) {
-    bulkDeleteInProgress = true;
-    await Promise.all(
-      selectedDeployments.map(async deployment => {
-        try {
-          await window.kubernetesDeleteDeployment(deployment.name);
-        } catch (e) {
-          console.log('error while deleting deployment', e);
-        }
-      }),
-    );
-    bulkDeleteInProgress = false;
+  if (selectedDeployments.length === 0) {
+    return;
   }
+
+  // mark deployments for deletion
+  bulkDeleteInProgress = true;
+  selectedDeployments.forEach(image => (image.status = 'DELETING'));
+  deployments = deployments;
+
+  await Promise.all(
+    selectedDeployments.map(async deployment => {
+      try {
+        await window.kubernetesDeleteDeployment(deployment.name);
+      } catch (e) {
+        console.log('error while deleting deployment', e);
+      }
+    }),
+  );
+  bulkDeleteInProgress = false;
 }
 
 let selectedItemsNumber: number;
--- a/packages/renderer/src/lib/pod/PodsList.svelte
+++ b/packages/renderer/src/lib/pod/PodsList.svelte
@@ -94,24 +94,29 @@ onDestroy(() => {
 let bulkDeleteInProgress = false;
 async function deleteSelectedPods() {
   const selectedPods = pods.filter(pod => pod.selected);
-
-  if (selectedPods.length > 0) {
-    bulkDeleteInProgress = true;
-    await Promise.all(
-      selectedPods.map(async pod => {
-        try {
-          if (pod.kind === 'podman') {
-            await window.removePod(pod.engineId, pod.id);
-          } else {
-            await window.kubernetesDeletePod(pod.name);
-          }
-        } catch (e) {
-          console.log('error while removing pod', e);
-        }
-      }),
-    );
-    bulkDeleteInProgress = false;
+  if (selectedPods.length === 0) {
+    return;
   }
+
+  // mark pods for deletion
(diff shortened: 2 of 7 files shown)


packages/renderer/src/lib/deployments/DeploymentsList.svelte
```diff
@@ -38,20 +38,25 @@ onMount(() => {
...
+  bulkDeleteInProgress = true;
+  selectedDeployments.forEach(image => (image.status = 'DELETING'));
+  deployments = deployments;
+
+  await Promise.all(
+    selectedDeployments.map(async deployment => {
+      try {
+        await window.kubernetesDeleteDeployment(deployment.name);
+      } catch (e) {
+        console.log('error while deleting deployment', e);
```
axel7083:
```suggestion
        console.error('error while deleting deployment', e);
```

packages/renderer/src/lib/pod/PodsList.svelte
```diff
@@ -94,24 +94,29 @@ onDestroy(() => {
...
+          await window.removePod(pod.engineId, pod.id);
+        } else {
+          await window.kubernetesDeletePod(pod.name);
         }
-      }),
-    );
-    bulkDeleteInProgress = false;
-  }
+      } catch (e) {
+        console.log('error while removing pod', e);
```
axel7083:
```suggestion
        console.error('error while removing pod', e);
```

packages/renderer/src/lib/service/ServicesList.svelte
```diff
@@ -36,20 +36,25 @@ onMount(() => {
...
+  bulkDeleteInProgress = true;
+  selectedServices.forEach(service => (service.status = 'DELETING'));
+  services = services;
+
+  await Promise.all(
+    selectedServices.map(async service => {
+      try {
+        await window.kubernetesDeleteService(service.name);
+      } catch (e) {
+        console.log('error while deleting service', e);
```
axel7083:
```suggestion
        console.error('error while deleting service', e);
```

packages/renderer/src/lib/volume/VolumesList.svelte
```diff
@@ -94,19 +94,25 @@ let bulkDeleteInProgress = false;
...
+  bulkDeleteInProgress = true;
+  selectedVolumes.forEach(volume => (volume.status = 'DELETING'));
+  volumes = volumes;
+
+  await Promise.all(
+    selectedVolumes.map(async volume => {
+      try {
+        await window.removeVolume(volume.engineId, volume.name);
+      } catch (e) {
+        console.log('error while removing volume', e);
```
axel7083:
```suggestion
        console.error('error while removing volume', e);
```

axel7083 (approved):
LGTM! Love this feature, tested locally and works perfectly !

benoitf (approved):
I guess we could have common code to handle such status updates

like instead of duplicating on all list/table pages

```
// mark services for deletion
bulkDeleteInProgress = true;
selectedServices.forEach(service => (service.status = 'DELETING'));
services = services;
```

I guess it could be applied to the Table component itself (for simple cases)
if not, I guess that we could have a function handling the `.status = 'DELETING'`

jeffmaury (approved):
LGTM

packages/renderer/src/lib/deployments/DeploymentsList.svelte
deboer-tim: Fixed in new commit.

packages/renderer/src/lib/pod/PodsList.svelte
deboer-tim: Fixed in new commit.

packages/renderer/src/lib/service/ServicesList.svelte
deboer-tim: Fixed in new commit.

packages/renderer/src/lib/volume/VolumesList.svelte
deboer-tim: Fixed in new commit.

deboer-tim:
> I guess we could have common code to handle such status updates
> 
> like instead of duplicating on all list/table pages
> 
> ```
> // mark services for deletion
> bulkDeleteInProgress = true;
> selectedServices.forEach(service => (service.status = 'DELETING'));
> services = services;
> ```
> 
> I guess it could be applied to the Table component itself (for simple cases) if not, I guess that we could have a function handling the `.status = 'DELETING'`

Yes. I wanted to do a 'bulk action helper' that would set a given state and then callback for the action, but there is still a lot of variation between the objects (e.g. containers use 'state' instead of 'status' [should we change this?], and has to do two passes) that I wasn't quite sure what level to do it at and decided to just make the code consistent to start. I am fine taking another pass at this, but ok if we handle as a separate issue?

benoitf:
sure I approved

cdrage (approved):
I wanted to test this out with about 40+ volumes.

Deleted beautifully!

Since I'm the last one awaiting a review, LGTM! I think 4 reviews are good enough for merge.

PR #5729, opened 2024-01-29
feat: introduce a version field for 3rd party extensions

### LINKED ISSUE #5728: expose podman desktop version to extensions

Extensions might want to report or behave differently based on the Podman Desktop version

introduce a `version` field on the extension API so extensions can know which Podman Desktop version is running

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
add a way for extensions to know the Podman Desktop version

fixes https://github.com/containers/podman-desktop/issues/5728

### How to test this PR?

unit test provided

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -18,6 +18,11 @@
 
 declare module '@podman-desktop/api' {
   /**
+   * The version of Podman Desktop.
+   */
+  export const version: string;
+
+  /**
    * Represents a reference to a command. Provides a title which
    * will be used to represent a command in the UI and, optionally,
    * an array of arguments which will be passed to the command handler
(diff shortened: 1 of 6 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -17,6 +17,11 @@
  ***********************************************************************/
 
 declare module '@podman-desktop/api' {
+  /**
+   * The version of Podman Desktop.
+   */
+  export const version: string;
```
jeffmaury, line 23: I'm wondering if we respect SemVer if we should not expose the SemVer components so that extensions have a better way to implement checks ?

benoitf: it can be any string (as it relies on the definition of package.json version definition)

As some people may override, build a custom version, it may not be a semVer at all

so we should not probably restrict here

feloy (approved):

packages/main/src/plugin/extension-loader.spec.ts
```diff
@@ -1305,3 +1314,22 @@ test('check listWebviews', async () => {
...
+    name: 'name',
+    publisher: 'publisher',
+    version: '1',
+    displayName: 'dname',
+  });
+
+  const readPodmanVersion = api.version;
+
+  // check we called method
+  expect(readPodmanVersion).toBeDefined();
```
dgolovin: nit: IMO toBe() call verifies both: to be defined and to be the expected value.

benoitf: @dgolovin amended

PR #5731, opened 2024-01-29
fix: message property used to update task name

### LINKED ISSUE #5730: ProgressImpl report `message` properties should not be used for error but title
### Bug description

The current implementation uses the `message` property to set the error message. 

https://github.com/containers/podman-desktop/blob/1f427287f9221562f18aba54dd61ca920f76d36a/packages/main/src/plugin/progress-impl.ts#L90-L98

Which is wrong according to https://github.com/containers/podman-desktop/pull/5712#issuecomment-1914384759

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Tweak the implementation of the `withProgress` method, to comply with the goal of the method.

Fixes https://github.com/containers/podman-desktop/issues/5730

### How to test this PR?

- [x] Regression unit tests have been added
- [x] Previous unit tests have been adapted

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1512,18 +1512,33 @@ declare module '@podman-desktop/api' {
      * If the promise complete without raising an exception the task will be marked as completed, otherwise it will be marked as failed.
      *
      * ```ts
-     * try {
-     *  void podmanDesktopApi.window.withProgress<void>(
-     *       { location: podmanDesktopApi.ProgressLocation.TASK_WIDGET, title: `Running task` },
+     * await window.withProgress(
+     *       { location: ProgressLocation.TASK_WIDGET, title: `Running task` },
      *       async progress => {
-     *           // might throw an error
-     *           await complicatedTask();
+     *          progress.report({message: 'task1' });
+     *          await task1();
+     *          progress.report({increment: 50, message: 'task2' });
+     *          await task2();
+     *       },
+     *     );
+     * ```
+    * // in case of error during the callback, the error is rethrown so you'll need to use try/catch around the call to withProgress
+     * ```ts
+     * try {
+     *   await window.withProgress(
+     *       { location: ProgressLocation.TASK_WIDGET, title: `Running task` },
+     *       async progress => {
+     *          throw new Error('error when running a task');
      *       },
      *     );
      * } catch (error) {
-     *    // to something with the error
+     *    // do something with the error
      * }
      * ```
+     
+     
+     
+     
      *
      * @param options The option to use
      * @param task The promise to execute
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1507,42 +1507,6 @@ declare module '@podman-desktop/api' {
      */
     export function showErrorMessage(message: string, ...items: string[]): Promise<string | undefined>;
 
-    /**
-     * Create a Task inside the TaskManager with a progress bar, by default indeterminate for the time of the promise you gave.
-     * If the promise complete without raising an exception the task will be marked as completed, otherwise it will be marked as failed.
-     *
-     * ```ts
-     * await window.withProgress(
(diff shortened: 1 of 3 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -1507,6 +1507,27 @@ declare module '@podman-desktop/api' {
...
+     *       { location: podmanDesktopApi.ProgressLocation.TASK_WIDGET, title: `Running task` },
+     *       async progress => {
+     *           // might throw an error
+     *           await complicatedTask();
+     *       },
+     *     );
+     * } catch (error) {
+     *    // to something with the error
+     * }
+     * ```
```
benoitf: I would have the version without error first and another example with catch


```suggestion
     * ```ts
     * await window.withProgress(
     *       { location: ProgressLocation.TASK_WIDGET, title: `Running task` },
     *       async progress => {
     *          progress.report({message: 'task1' });
     *          await task1();
     *          progress.report({increment: 50, message: 'task2' });
     *          await task2();
     *       },
     *     );
     * ```
    * // in case of error during the callback, the error is rethrown so you'll need to use try/catch around the call to withProgress
     * ```ts
     * try {
     *   await window.withProgress(
     *       { location: ProgressLocation.TASK_WIDGET, title: `Running task` },
     *       async progress => {
     *          throw new Error('error when running a task');
     *       },
     *     );
     * } catch (error) {
     *    // do something with the error
     * }
     * ```
     
     
     
     
```

packages/main/src/plugin/progress-impl.spec.ts
```diff
@@ -109,3 +110,29 @@ test('Should create a task and propagate the result', async () => {
...
+  const updateTaskMock = vi.fn();
+  const taskManager = {
+    createTask: createTaskMock,
+    updateTask: updateTaskMock,
+  } as unknown as TaskManager;
+
+  createTaskMock.mockImplementation(() => ({}));
+  const progress = new ProgressImpl(taskManager);
+
+  void (await progress.withProgress<void>(
```
benoitf: why are you adding these void ?
you can just await and not specify <void>

```ts
await progress.withProgress(...)
```

packages/main/src/plugin/progress-impl.spec.ts
```diff
@@ -78,6 +78,7 @@ test('Should create a task and propagate the exception', async () => {
 
   expect(updateTaskMock).toHaveBeenCalledTimes(1);
   expect(updateTaskMock).toHaveBeenCalledWith({
```
benoitf, line 76: I think the test 
```
  try {
    await progress.withProgress({ location: ProgressLocation.TASK_WIDGET, title: 'My task' }, async () => {
      throw new Error('dummy error');
    });
    // Should NEVER be here.
    expect(true).toBe(false);
  } catch (e: unknown) {
    expect((e as Error).message).toBe('dummy error');
  }

```

should use await expect(....).rejects.toThrow('dummy error') form

axel7083: Yes, I fixed it !

packages/main/src/plugin/progress-impl.spec.ts
```diff
@@ -109,3 +110,26 @@ test('Should create a task and propagate the result', async () => {
...
+  const updateTaskMock = vi.fn();
+  const taskManager = {
+    createTask: createTaskMock,
+    updateTask: updateTaskMock,
+  } as unknown as TaskManager;
+
+  createTaskMock.mockImplementation(() => ({}));
+  const progress = new ProgressImpl(taskManager);
+
+  await progress.withProgress<void>({ location: ProgressLocation.TASK_WIDGET, title: 'My task' }, async progress => {
```
benoitf, line 121:
```suggestion
  await progress.withProgress<void>({ location: ProgressLocation.TASK_WIDGET, title: 'My task' }, progress => {
```

benoitf (approved):
you can ignore my minor comment on extra async

Issue #5670, opened 2024-01-24
truncate long usernames in registries page


vrothberg:
### Is your enhancement related to a problem? Please describe

Usernames of registry users can be long, especially when machine-generated such as the registry-service accounts used by the Red Hat auth plugin (see https://github.com/redhat-developer/podman-desktop-redhat-account-ext/pull/26).  Since names are displayed entirely, it can mess up the interface a bit.

### Describe the solution you'd like

Truncate long usernames (maybe even long registries as well) and expand them on demand.

### Describe alternatives you've considered

I haven't considered any other alternative.

### Additional context

![image](https://github.com/redhat-developer/podman-desktop-redhat-account-ext/assets/11679875/9eeec119-a729-45b1-874d-05d8d5756b84)

vrothberg:
Wow, that was super fast.  Thanks a ton, @deboer-tim !

deboer-tim:
You made Podman stable on my machine, fixing a minor UI issue is the least I can do. ;-)

PR #5737, opened 2024-01-29
chore: long usernames in auth page

deboer-tim:
### What does this PR do?

While testing #5670 I noticed that the auth page has an identical issue with long usernames. This just uses a flex row and the same 'text-ellipsis overflow-hidden max-w-64' to ensure long usernames don't cause UI issues.

### Screenshot / video of UI

Before:

Screenshot 2024-01-29 at 10 19 32 AM

After:

Screenshot 2024-01-29 at 10 10 59 AM

N/A

### How to test this PR?

Have a very long username, or edit line 83 with junk to simulate one.


axel7083 (approved):
LGTM

benoitf (approved):
probably it's missing a test to check it's cut to avoid it'll be reproduced

cdrage (approved):

PR #5763, opened 2024-01-30
chore: fix style of download button when we can't detect OS

deboer-tim:
### What does this PR do?

When we can determine the user's OS, we have a nice purple Download Now button on the front page.

When we can't determine the user's OS, it's a full-width blue button linking to the Download Page instead. This button doesn't match our current styling at all.

I copied from the Download Now button to make it match:
- Added outer div (makes it normal size and centered).
- Changed indigo to purple.
- Added 'mt-6 mb-1' to space it from above and below.

### Screenshot / video of UI

Before:
Screenshot 2024-01-30 at 2 34 32 PM

After:
Screenshot 2024-01-30 at 2 39 30 PM

N/A

### How to test this PR?

`yarn website:dev`, set `operatingSystem = ''` to undo OS detection.

### DIFF
--- a/website/src/pages/index.tsx
+++ b/website/src/pages/index.tsx
@@ -67,11 +67,13 @@ function DownloadClientLinks() {
     );
   } else {
     mainButton = (
-      <Link
-        className="no-underline hover:no-underline inline-flex text-white hover:text-white bg-indigo-500 border-0 py-2 px-6 focus:outline-none hover:bg-indigo-600 rounded text-lg"
-        to="/downloads">
-        Download Page
-      </Link>
+      <div>
+        <Link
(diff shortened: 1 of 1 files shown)


benoitf (commented):
just a FYI, I think this PR is doing too many things at once

if there is a regression, it's hard to track

- one PR to remove a comment
- one PR to change the style of a button

and then the PR that gets rid of indigo by just replacing the color

deboer-tim:
Ok, I'll reduce this one to start.

deboer-tim:
Force pushed just the one change instead, and updated title and description to match.

benoitf (approved):

benoitf:
thanks for the updated PR @deboer-tim

PR #5773, opened 2024-01-31
feat: add statusbar item highlight property

### LINKED ISSUE #5814: Status Bar item should have a highlighted/badged/dotted elements to bring attention to it

Some status bar item should have a visible mark when they requires/want user's attention.

A dotted visual elements on top/right corner of the text, to bring a light, and nice attention to the item

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

This PR is adding an `important` property to the StatusBarItem. This can be useful to bring attention to the user to a specific item. For example here the `Update available` StatusBar item.

### Screenshot / video of UI

image

Fixes https://github.com/containers/podman-desktop/issues/5814

### How to test this PR?

- [x] Unit tests have been added

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -570,6 +570,7 @@ export class PluginSystem {
           true,
           'update',
           undefined,
+          true,
         );
       });
 
--- a/packages/main/src/plugin/statusbar/statusbar-item.ts
+++ b/packages/main/src/plugin/statusbar/statusbar-item.ts
@@ -34,6 +34,7 @@ export class StatusBarItemImpl implements StatusBarItem {
   private isVisible = false;
   private _iconClass: string | { active: string; inactive: string } | undefined;
   private _enabled = true;
+  private _important = false;
 
   private _command: string | undefined;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -46,6 +47,15 @@ export class StatusBarItemImpl implements StatusBarItem {
     this._alignment = alignment;
     this._priority = priority;
     this._id = StatusBarItemImpl.nextId();
+  }
+
+  public get important(): boolean {
+    return this._important;
+  }
+
+  public set important(important: boolean) {
+    this._important = important;
+    this.update();
   }
 
   public get alignment(): StatusBarAlignment {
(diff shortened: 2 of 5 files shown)


axel7083:
cc @deboer-tim I followed your advise and change the name `dotted` to have something agnostic of the frontend implementation, and I went with `important`.

deboer-tim:
Thanks for changing it, but I'm not sure important makes sense in this context? It kind of says that everything is important until you look at it, and it's not. Could you change it to something more like isNew, or userNotified, or ...?

axel7083:
> Thanks for changing it, but I'm not sure important makes sense in this context? It kind of says that everything is important until you look at it, and it's not. Could you change it to something more like isNew, or userNotified, or ...?

@deboer-tim I am willing to change, but maybe we could discuss which one we should use ? 
- I would argue that `isNew` could be interpret as a new status bar item has been created.
- For `userNotified`, we could think, that it is true when the user has been notified, but in fact it is factually not true, the user has not been notified, the ui has changed to _hightlight_ the status bar item
- To continue, the `hightlight` is also not really something I like, as it would let think that it is hightlight in some manner, than the frontend implementation should follow, which looks like similar to `dotted`.

I choose `important` as it let the frontend choose the way to display it, an update is important, you cannot remove the `dot` of the update by just clicking on it, you need to perform the update.

But I'm not very good at wording, so here is a list of proposition given by our dear friend ChatGPT.

image

cc @benoitf any suggestion ?

benoitf:
I would go with `highlight` if it works for @deboer-tim as well

deboer-tim:
The point of the function is notifying the user that something is new/has changed, and the status goes away once they've acknowledged it, so I still prefer something like `userNotified` or `acknowledged` or whatever 🤷🏻‍♂️, but we're probably already bike shedding here so `highlight` would be fine too! (thank you @axel7083)

axel7083:
> The point of the function is notifying the user that something is new/has changed, and the status goes away once they've acknowledged it, so I still prefer something like `userNotified` or `acknowledged` or whatever 🤷🏻‍♂️, but we're probably already bike shedding here so `highlight` would be fine too! (thank you @axel7083)

@deboer-tim @benoitf I updated to `highlight`

deboer-tim (approved):
Thanks for all the review on this.

PR #5775, opened 2024-01-31
Disable actions when deleting

### LINKED ISSUE #5739: Disable actions when deleting is on process

When I click to delete an image and the deleting process starts, I notice that Podman-Dektop will gray out the delete button to avoid double click, but the action buttons are still clickable. 

In that case if, I click on play button while an image is being deleted Podman-Desktop will take me to a blank screen.

https://github.com/containers/podman-desktop/assets/16507629/1646cde5-8f9d-4a47-aaf9-2b8ff59d6caf

When I click to delete an image and the delete process starts, I would like to have all actions and buttons for that specific image to be disabled.

Can't think of any other alternative to the delete process but, I would love to get more ideas :)

### Additional context

_No response_

trmendes:
### What does this PR do?

To avoid user to perform any action on an object that is being deleted, this PR disables some actions buttons when user click on delete.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/16507629/327419fb-e1c4-4e52-9526-fd586a5e61d2

Fixes #5739

### How to test this PR?

Option 1) Unit Tests
Option 2) Play around with creating and deleting. Expect buttons to be disabled.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerActions.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerActions.spec.ts
@@ -100,3 +100,26 @@ test('Expect no error and status deleting container', async () => {
   expect(container.actionError).toEqual('');
   expect(updateMock).toHaveBeenCalled();
 });
+
+test('Expect buttons to be disabled when deleting', async () => {
+  render(ContainerActions, { container });
+  container.state = 'DELETING';
+
+  const deleteButton = screen.getByRole('button', { name: 'Delete Container' });
+  expect(deleteButton).toBeDisabled();
+
+  const startButton = screen.getByRole('button', { name: 'Start Container' });
+  expect(startButton).toBeDisabled();
+
+  const stopButton = screen.getByRole('button', { name: 'Stop Container' });
+  expect(stopButton).toBeDisabled();
+
+  const generateKubeButton = screen.getByRole('button', { name: 'Generate Kube' });
+  expect(generateKubeButton).toBeDisabled();
+
+  const deployToKubeButton = screen.getByRole('button', { name: 'Deploy to Kubernetes' });
+  expect(deployToKubeButton).toBeDisabled();
+
+  const restartContainerButton = screen.getByRole('button', { name: 'Restart Container' });
+  expect(restartContainerButton).toBeDisabled();
+});
--- a/packages/renderer/src/lib/deployments/DeploymentActions.spec.ts
+++ b/packages/renderer/src/lib/deployments/DeploymentActions.spec.ts
@@ -64,3 +64,11 @@ test('Expect no error and status deleting deployment', async () => {
   expect(updateMock).toHaveBeenCalled();
   expect(deleteMock).toHaveBeenCalled();
 });
+
+test('Expect buttons to be disabled when deleting', async () => {
+  render(DeploymentActions, { deployment });
+  deployment.status = 'DELETING';
+
+  const deleteButton = screen.getByRole('button', { name: 'Delete Deployment' });
+  expect(deleteButton).toBeDisabled();
+});
--- a/packages/renderer/src/lib/service/ServiceActions.spec.ts
+++ b/packages/renderer/src/lib/service/ServiceActions.spec.ts
@@ -63,3 +63,11 @@ test('Expect no error and status deleting service', async () => {
   expect(updateMock).toHaveBeenCalled();
   expect(deleteMock).toHaveBeenCalled();
 });
+
+test('Expect buttons to be disabled when deleting', async () => {
+  render(ServiceActions, { service });
+  service.status = 'DELETING';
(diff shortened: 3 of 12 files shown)


feloy:
When I delete images and open the contextual menu on the image actions, I can see that some entries are still enabled (most probably entries added by extensions). Do you think it could be posisble to disable them, or disable the three dots instead?

trmendes:
> When I delete images and open the contextual menu on the image actions, I can see that some entries are still enabled (most probably entries added by extensions). Do you think it could be posisble to disable them, or disable the three dots instead?

Hi @feloy. I have noticed it too! Tomorrow I will try to check how to disable the entries added by extensions.

deboer-tim:
I'm a little uneasy about this - for our own actions we're changing their enablement based on state, but for extensions we're just saying 'no you can't do that here'. I suppose it is ok for deletion b/c the object is going away anyway, but I worry it's removing the onus on extension developers to correctly enable their actions based on state (status in this case).

trmendes:
> I'm a little uneasy about this - for our own actions we're changing their enablement based on state, but for extensions we're just saying 'no you can't do that here'. I suppose it is ok for deletion b/c the object is going away anyway, but I worry it's removing the onus on extension developers to correctly enable their actions based on state (status in this case).

I agree with @deboer-tim . 

What I have in mind for trying tomorrow is to have all other actions disabled for the deletion only.

packages/renderer/src/lib/container/ContainerActions.spec.ts
```diff
@@ -100,3 +100,28 @@ test('Expect no error and status deleting container', async () => {
   expect(container.actionError).toEqual('');
   expect(updateMock).toHaveBeenCalled();
 });
+
+test('Expect buttons to be disabled when deleting', async () => {
+  const { component } = render(ContainerActions, { container });
+  component.$on('update', updateMock);
```
deboer-tim: Shouldn't need this line (since we're not doing anything with the mock).

packages/renderer/src/lib/deployments/DeploymentActions.spec.ts
```diff
@@ -64,3 +64,13 @@ test('Expect no error and status deleting deployment', async () => {
   expect(updateMock).toHaveBeenCalled();
   expect(deleteMock).toHaveBeenCalled();
 });
+
+test('Expect buttons to be disabled when deleting', async () => {
+  const { component } = render(DeploymentActions, { deployment });
+  component.$on('update', updateMock);
```
deboer-tim: Shouldn't need this line either.

packages/renderer/src/lib/service/ServiceActions.spec.ts
```diff
@@ -63,3 +63,12 @@ test('Expect no error and status deleting service', async () => {
   expect(updateMock).toHaveBeenCalled();
   expect(deleteMock).toHaveBeenCalled();
 });
+
+test('Expect buttons to be disabled when deleting', async () => {
+  const { component } = render(ServiceActions, { service });
+  component.status = 'DELETING';
```
deboer-tim: I thought it was suspicious that the test still worked when the status was set on `component` instead of `service` ;-) and indeed this test is still passing because the button is already in progress from the previous test - if you reorder the tests, it fails.

I didn't have enough time to figure out why the tests aren't independent, but needs some cleanup between. To confirm the test is working it's probably good to confirm that it fails when this line is deleted and not otherwise.

packages/renderer/src/lib/service/ServiceActions.spec.ts
```diff
@@ -63,3 +63,12 @@ test('Expect no error and status deleting service', async () => {
...
+  component.$on('update', updateMock);
```
deboer-tim: Shouldn't need this line either.

deboer-tim (changes requested):
Thanks for the contribution - the action changes all look good, just needs some test cleanup.

packages/renderer/src/lib/service/ServiceActions.spec.ts
trmendes: Ops! Thank you for noticing that :D.

trmendes:
Hi @deboer-tim. Thanks for the hints. I wonder if I'm missing something.

trmendes:
> Hi @deboer-tim. Thanks for the hints. I wonder if I'm missing something.

@deboer-tim, @feloy: I'm leaving the 'disable menu' functionality for further discussions and, to another Issue/PR. What are your comments about?

The PR is open for another review :D. 

I hope we are able to merge it soon.

benoitf:
@feloy @deboer-tim looks like you didn't reply to @trmendes comments

switching to draft

benoitf:
not updated, closing

PR #5797, opened 2024-02-01
chore(tests):bootc e2e installation test

### LINKED ISSUE #117: Create basic e2e tests to validate extension installation and functionality

Currently there are no e2e tests to check bootc extension.

Create e2e tests to check bootc extension installation and basic functionality.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?

Adds e2e test for the installation for bootc extension.

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/117

### How to test this PR?

`yarn test:e2e:bootc`


benoitf (commented):
just a side note but these tests should not be committed into this repository
as it's a 3rd party extension

this repository should only contain test from this repository

cbr7:
@benoitf that is the long term plan however for now it has to be like this since the automation framework cannot be exported currently and putting the tests in the extension repo would mean that we'd have to create a second test automation framework.

The plan for right now is to pull the PD repo and run the tests like that when doing the CICD pipeline of the bootc extension.

benoitf:
> The plan for right now is to pull the PD repo and run the tests like that when doing the CICD pipeline of the bootc extension.


the test class should be part of the 3rd party extension repository


1. clone Podman Desktop repository to a fixed revision like v1.7.0
1. add extension class test
1. run the e2e test from there

cbr7:
@benoitf I'm not sure I understand, the test class relies on the tst automation framework as a dependency, and right now we cannot export the test framework as an npm package, so creating the test class in another repo would not work.

benoitf:
> @benoitf I'm not sure I understand, the test class relies on the tst automation framework as a dependency, and right now we cannot export the test framework as an npm package, so creating the test class in another repo would not work.

I don't see any difference at running time if the class is copied after we clone the repository or it was part of the repository

I understand that you don't want for now export/publish to a npm package

but if you clone the full podman desktop repository  when launching the test, if you just add a new file in the tests, it will execute the tests from this file. This file can contain specific 3rd party extension tests

cbr7:
The difference is that the test class will show up in the code as unbuildable, since it would be just undefined objects being used that don't have the definitions. I'm not even sure the code would compile, am I missing something?

benoitf:
yes it won't compile on its own but at least it's on the right repository

odockal:
@benoitf @cbr7 We have a test we need. It is in the wrong place or it will not be simply possible to develop such test. 
There are two options from my POV. 
* We accept the fact we are in the wrong repo and still have running E2E tests, at the same time we will propose and start executing a plan to alleviate this issue to make it right
* We will address the moving and refactoring the test-framework first so it will be possible to write podman-desktop e2e tests anywhere but the needed e2e tests will be delivered later

@jeffmaury @deboer-tim wdyt? Maybe other option? Of course, if the necessity of a right placement of the tests due to various reasons untold here is of a higher priority, we have no other option atm and we need to start moving forward with the second point asap.

cbr7:
@odockal has there been any agreement on how to proceed on this?

cbr7:
@odockal @benoitf I've moved the test file for here into the extension repo as requested, the file will be copied into PD repo via npm script before executing the test run. I've opened https://github.com/containers/podman-desktop-extension-bootc/pull/163 for the test file in the other repo.

odockal (approved):

odockal:
@cbr7 @deboer-tim @benoitf From my point of view and the way we agreed for this to be done before we will be able make the testing framework modular/reusable, it is ok. I would like to see this in action since we cannot really see the test results here.

PR #5799, opened 2024-02-01
chore: remove indigo from palette

deboer-tim:
### What does this PR do?

Indigo is one of the historic colors in our palette that we're not supposed to use, and it is finally out of the codebase. Removing from the palette so that it doesn't get accidentally used.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Confirm there are no instances of `indigo` anywhere else in the codebase.


axel7083 (approved):
LGTM

PR #5809, opened 2024-02-01
feat: adding extensionId optional property to Task interface

axel7083:
### What does this PR do?

This PR is adding a `extensionId` property to the Task interface.

> The `NotificationCardOptions` already have such property, as it makes sense to keep track of which extension created notifcations, this PR is expanding this idea to the Tasks.

Related to https://github.com/containers/podman-desktop/issues/5804

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/api/task.ts
+++ b/packages/main/src/plugin/api/task.ts
@@ -23,6 +23,7 @@ export interface Task {
   id: string;
   name: string;
   started: number;
+  readonly extensionId?: string;
 }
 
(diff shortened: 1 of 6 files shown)


benoitf (changes requested):
I would not add it for the cleanup part. It's unrelated as cleanup operation can be achieved without extensionIDs

axel7083:
Closing as solution offered in https://github.com/containers/podman-desktop/issues/5804#issuecomment-1922329043 is really nice

PR #5811, opened 2024-02-01
feat: adding pull image to task manager

### LINKED ISSUE #5810: User should be able to follow progress of pulling image even after leaving the page

When pulling large image, we might leave the pull image page, user should be able to see that something is happening in the background

Using TaskManager

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Pull image action should create a task visible in the Task Manager.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/42176370/0eadc9b6-80ed-4031-b8d4-14a9a0453ebd

Fixes https://github.com/containers/podman-desktop/issues/5810

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -57,7 +57,7 @@ import { getFreePort, getFreePortRange, isFreePort } from './util/port.js';
 import { isLinux, isMac } from '../util.js';
 import type { MessageBoxOptions, MessageBoxReturnValue } from './message-box.js';
 import { MessageBox } from './message-box.js';
-import { ProgressImpl } from './progress-impl.js';
+import { ProgressImpl, ProgressLocation } from './progress-impl.js';
 import type { ContributionInfo } from './api/contribution-info.js';
 import { ContributionManager } from './contribution-manager.js';
 import { DockerDesktopInstallation } from './docker-extension/docker-desktop-installation.js';
@@ -742,6 +742,8 @@ export class PluginSystem {
     const kubernetesUtils = new KubernetesUtils(configurationRegistry);
     kubernetesUtils.init();
 
+    const progressImpl = new ProgressImpl(taskManager);
+
     this.extensionLoader = new ExtensionLoader(
       commandRegistry,
       menuRegistry,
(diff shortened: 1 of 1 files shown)


deboer-tim (commented):
I really appreciate this fix for a longstanding and annoying gap!

However, using withProgress() in main highlighted another inconsistency to me:
- doing it here means that even if an extension triggers the pull via API, we'll still see the progress (awesome!).
- doing it in the renderer/UI we'd create a task that included a link back to the form so that you could go back and see the output, but doing it here there's no returning to see the output, even for pulls started in the form (not so awesome).

I think we need a quick architecture chat with @benoitf or others to confirm how this should work long term. My initial reaction is that this PR should go forward as-is and we add a pattern for associating execution output during withProgress(), and there's a way to view that in UI without going back to the original form. I think this could actually simplify some of our forms and provide a better experience.

benoitf (changes requested):
AFAIK it's the system call in this method and it should stay like that
we should not bring a direct dependency to the task

axel7083:
> I really appreciate this fix for a longstanding and annoying gap!
> 
> However, using withProgress() in main highlighted another inconsistency to me:
> 
> * doing it here means that even if an extension triggers the pull via API, we'll still see the progress (awesome!).

I don't think we want that, do we ? I think like extensions should be responsible of displaying the things they want ?

> * doing it in the renderer/UI we'd create a task that included a link back to the form so that you could go back and see the output, but doing it here there's no returning to see the output, even for pulls started in the form (not so awesome).

The pulling page is not stateless, therefore we cannot go back to it, as after leaving it, it is lost forever, we need to refactor/update it a bit to make it being able to return to it with a link. But I would love that!

> I think we need a quick architecture chat with @benoitf or others to confirm how this should work long term. My initial reaction is that this PR should go forward as-is and we add a pattern for associating execution output during withProgress(), and there's a way to view that in UI without going back to the original form. I think this could actually simplify some of our forms and provide a better experience.

axel7083:
> AFAIK it's the system call in this method and it should stay like that we should not bring a direct dependency to the task

@benoitf I'm sorry but could you clarify, I am not sure understanding what you mean

axel7083:
> AFAIK it's the system call in this method and it should stay like that we should not bring a direct dependency to the task

@benoitf I still think that we should have some kind of tasks created for this action, what solution do you prefer
- Adding the TaskDependency as a dependencies in the ContainerRegistry, and making it call the withProgress
> This would lead to another question, do we want to expand that to other features ?
- Only having the tasks on the renderer side, as the `build-image-task.ts` 

_Other solution?_

PR #5825, opened 2024-02-02
fix: check route tls to use either http or https

### LINKED ISSUE #5766: Kubernetes UI is always assuming OpenShift routes are using https
### Bug description

In the UI, in Ingresses & Routes section, you can see your OpenShift routes for examples but links are always using https

it should not assume it's always https, but check the tls spec of the Route object

### Operating system

macOS

### Installation Method

None

### Version

1.7.0

### Steps to reproduce

create a simple dummy Route, and see it'll be displayed using https

but it should be based on what is defined by the route
https://docs.openshift.com/container-platform/4.14/networking/routes/secured-routes.html

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Ensure protocol http or https depends on the tls part of the route

fixes https://github.com/containers/podman-desktop/issues/5766

### How to test this PR?

unit test

or deploy some routes without tls and check that it's either http or https

### DIFF
--- a/packages/renderer/src/lib/ingresses-routes/IngressRouteActions.spec.ts
+++ b/packages/renderer/src/lib/ingresses-routes/IngressRouteActions.spec.ts
@@ -69,6 +69,7 @@ test('Expect no error and status deleting route', async () => {
       name: 'service',
     },
     selected: false,
+    tlsEnabled: false,
   };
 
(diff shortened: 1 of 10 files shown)


deboer-tim (approved):
LGTM

PR #5828, opened 2024-02-02
feat: adds more summary information for kubernetes

### LINKED ISSUE #5658: Kubernetes UI Details pages should be unified / similar format across all details pages.

Within Pods we have this layout:

Screenshot 2024-01-23 at 5 14 34 PM

But in Deployments we have:

Screenshot 2024-01-23 at 5 15 22 PM

We should have it follow the same "design" pattern that we have for container / pod details. 

_No response_

### Additional context

_No response_

cdrage:
feat: adds more summary information for kubernetes

### What does this PR do?

* Refactors and unifies some of the parts that were used for summary
  pages
* Adds a lot more information with regards to details on each Kubernetes
  artifact (ingress, routes, services, etc.)

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/6422176/faf5f465-c829-43dc-8052-c8a8e5849108

Closes https://github.com/containers/podman-desktop/issues/5658

### How to test this PR?

Click on the Summary for each section.


cdrage:
This is up for review!

I do need to **write more tests** before this is merged.

But since this is a lot more information added, I'll wait until we get a final review before some simple tests are added.

axel7083:
IMHO I feel it is hard to read, they are a lot of information, and not enough contrast in certain information.

**containers**

E.g. The containers in the pods are not contrasted enough, they are the core of the resource, and as a developers, I might want an indicator related to their status, (are they all running ? Did one is still pending?)

image

**IPs**

I also feel like _other tools_ are highlighting some informations, we might want to have some kind of `code block` for some information, and colors for status

image

cdrage:
> IMHO I feel it is hard to read, they are a lot of information, and not enough contrast in certain information.
> 
> 
> 
> **containers**
> 
> 
> 
> E.g. The containers in the pods are not contrasted enough, they are the core of the resource, and as a developers, I might want an indicator related to their status, (are they all running ? Did one is still pending?)
> 
> 
> 
> image
> 
> 
> 
> **IPs**
> 
> 
> 
> I also feel like _other tools_ are highlighting some informations, we might want to have some kind of `code block` for some information, and colors for status
> 
> 
> 
> image
> 
> 
> 
> 

I 100% and totally agree. This PR is meant to add the initial information. 

I am hoping in subsequent PRs we can work with the UI team to define some good ways to contrast all the new information once we have more resources.

packages/renderer/src/lib/deployments/DeploymentDetailsSummary.svelte
```diff
@@ -1,11 +1,10 @@
 <script lang="ts">
 import type { V1Deployment } from '@kubernetes/client-node';
 import ErrorMessage from '../ui/ErrorMessage.svelte';
-import DeploymentColumnConditions from './DeploymentColumnConditions.svelte';
-import type { DeploymentUI } from './DeploymentUI';
-import DeploymentColumnPods from './DeploymentColumnPods.svelte';
+import ObjectMeta from '../kube/details/ObjectMeta.svelte';
```
benoitf: I would prefix ObjectMeta with something like KubernetesDetailsObjectMeta to avoid to have like 5 ObjectMeta classes in Podman Desktop

packages/renderer/src/lib/deployments/DeploymentDetailsSummary.svelte
```diff
@@ -1,11 +1,10 @@
...
+import Deployment from '../kube/details/Deployment.svelte';
```
benoitf: please add prefix /suffix with Kubernetes and details so with the className you see in one sec if it's a generic, specific, etc class

packages/renderer/src/lib/deployments/DeploymentDetailsSummary.svelte
```diff
@@ -1,11 +1,10 @@
...
+import DeploymentStatus from '../kube/details/DeploymentStatus.svelte';
```
benoitf: add kube/Kubernetes prefix

packages/renderer/src/lib/ingresses-routes/IngressRouteDetailsSummary.svelte
```diff
@@ -2,14 +2,18 @@
 import type { V1Ingress } from '@kubernetes/client-node';
 import type { V1Route } from '../../../../main/src/plugin/api/openshift-types';
 import ErrorMessage from '../ui/ErrorMessage.svelte';
-import type { IngressUI } from './IngressUI';
-import type { RouteUI } from './RouteUI';
-import IngressRouteColumnHostPath from './IngressRouteColumnHostPath.svelte';
-import IngressRouteColumnBackend from './IngressRouteColumnBackend.svelte';
+import ObjectMeta from '../kube/details/ObjectMeta.svelte';
+import Ingress from '../kube/details/Ingress.svelte';
+import IngressStatus from '../kube/details/IngressStatus.svelte';
```
benoitf: need to find a new home for this objects

it looks strange to import from `../kube/details` when we're inside `ingresses-routes`

benoitf: also we should favor the `/@/ notation`

packages/renderer/src/lib/kube/details/ObjectMeta.svelte
```diff
@@ -0,0 +1,122 @@
...
+let internalAnnotations: [string, string][] = [];
+if (artifact?.annotations) {
+  Object.entries(artifact.annotations).forEach(([key, value]) => {
+    if (internalKubernetesKeys.includes(key)) {
+      internalAnnotations.push([key, value]);
+    } else {
+      annotations.push([key, value]);
+    }
+  });
+}
```
benoitf: it feels weird that it's not inside onMount method ?

packages/renderer/src/lib/kube/details/Volume.svelte
```diff
@@ -0,0 +1,227 @@
+<script lang="ts">
+/* There are many types of volumes that need to be shown, all with different 'name' values, etc.
```
benoitf: this class needs a prefix, Volume is too generic

packages/renderer/src/lib/kube/details/utils.ts
```diff
@@ -0,0 +1,58 @@
+/**********************************************************************
+ * Copyright (C) 2022-2023 Red Hat, Inc.
```
benoitf: 2024

benoitf (commented):
nice improvement

I'm finding a bit strange the pattern
```
<tr>
    <td class="{rowStyle}">Name</td>
    <td class="{rowStyle}">{artifact.name}</td>
  </tr>
```

we're like duplicating <tr><td with the style 'rowStyle' everywhere

we have the table component maybe should use it

or we avoid all the <tr><td> stuff

I mean, you could have a <Row> or <WhateverComponent> object instead of duplicating all the lines

or just give an array like
```
 [["Metadata"], ['Name', artifact.name]]
 ```

 like a structure to be displayed

File: packages/renderer/src/lib/kube/details/ObjectMeta.svelte (hunk 5 above)
Comment (cdrage) [reply packages/renderer/src/lib/kube/details/ObjectMeta.svelte]: So because it is simple sorting, I did not put it in the onMount method since I get this error:

```
if_block0.p is not a function
```

When svelte tries to check: `  {#if annotations.length > 0}`

It seems to stem from this issue: https://github.com/sveltejs/svelte/issues/5789

File: packages/renderer/src/lib/kube/details/KubeIngressArtifact.svelte [hunk 8]
```diff
@@ -0,0 +1,35 @@
...
+        {#each artifact.rules || [] as rule}
+          <!-- Here we use || [] to ensure it's always an array -->
+          {#if rule.http}
+            {#each rule.http.paths as path}
+              Path: {path.path} • Link:
+              <Link externalRef="{artifact.tls && artifact.tls.length > 0 ? 'https' : 'http'}://{rule.host}{path.path}">
+                {artifact.tls && artifact.tls.length > 0 ? 'https' : 'http'}://{rule.host}{path.path}
+              </Link>
+              • Backend: {path.backend.service?.name}{path.backend.service?.port?.number ? ':' : ''}{path.backend
+                .service?.port?.number}
```
lstocchi: Maybe we should check the values here.
I tried to check the summary of the resources i had on my cluster and i saw 
image

lstocchi (commented):
Nice and impressive work really. Everything is showed in a clear way.

Looking at the code i don't have much to say other than the thought Florent had about the repeating <tr>, if there could be a way to simplify the table stuff would be awesome.

Just noticed a missing check on the ingress using my existing cluster.

cdrage:
> Nice and impressive work really. Everything is showed in a clear way.
> 
> Looking at the code i don't have much to say other than the thought Florent had about the repeating , if there could be a way to simplify the table stuff would be awesome.
> 
> Just noticed a missing check on the ingress using my existing cluster.

Updated!

@lstocchi @benoitf @deboer-tim 

The reason for the "repeating" and the fact we don't implement a generic table is that we plan to implement some of the more "prettier" parts of the summary back. For example, how we show the columns for NodePort, etc:

Screenshot 2024-02-14 at 8 41 44 AM

I've gone ahead and implemented the `Cell.svelte` to avoid repeating the styles however.

Another reason is that the API for Kubernetes is quite... finnicky.

There's empty arrays (on purpose), some elements are undefined and others are straight up empty strings, it's an absolute mess at times especially dealing with custom resources too.

So that's why we're trying to keep the "meta" stuff generic (using objectmeta, etc.), but the more larger svelte files such as Deployment, would have to be custom.

I hope that helps!

This PR is just the first stepping stone to improved summary pages.

cdrage:
Ready for another review 👍

odockal:
@cdrage Do we need some E2E tests to cover functionality or are unit tests enough?
Or maybe something in between?

cdrage:
> @cdrage Do we need some E2E tests to cover functionality or are unit tests enough? Or maybe something in between?

Unit tests are fine :) these are just taking the artifacts and parsing tghem.

lstocchi:
@cdrage thanks charlie, saw your update. 
The missing piece is still the backend from my cluster
image

cdrage:
> @cdrage thanks charlie, saw your update. The missing piece is still the backend from my cluster image

Odd you don't have a backend for the ingress! Or it may be a resource instead of a service?

I'm curious what your YAML is. 

I've gone ahead and updated the PR to also include showing resource as well as checking to see that 'service' and 'resource' is available before showing.

cdrage:
Since I've gotten a few reviews, I'm going to go ahead and add the rest of the tests now so it's ready for final review 👍

cdrage:
Tests have been added 🎉 

I also missed adding "WilcardPolicy" for OpenShift Routes as that's a mandatory field, so modified the details screen for that.

Ready for review!

jeffmaury:
If got an empty details, then the section should be hidden ?

image

cdrage:
> If got an empty details, then the section should be hidden ?
> 
> image

Good catch!

Today I learned that you can have an ingress with zero rules and just a defaultBackend. 

I've updated the PR to accurately show the default resource (should appear as not blank for you anymore):
Screenshot 2024-02-22 at 9 16 24 AM

And this made me open up issue https://github.com/containers/podman-desktop/issues/6107

I tested the ingress (screenshot above) with this file:

```yaml
kind: Ingress
metadata:
  name: test
  namespace: default
spec:
  defaultBackend:
    resource:
      name: 'foobar'
      kind: 'foobar'
```

feloy:
I find it unfortunate to hardcode the list of information displayed in these Summaries, in which order, with which titles, etc.

I would prefer to have a kind of template (remember gotemplates ;) ) and the resource would be applied to the template before to display it. This way, we can imagine in the future that users (or companies) may be able to provide their own templates.

> There's empty arrays (on purpose), some elements are undefined and others are straight up empty strings, it's an absolute mess at times especially dealing with custom resources too.

For this, we could compute the value that would be displayed, but let the template decide where/if to display this data. That would help separate data and presentation.

But, anyway, that's a great work

cdrage:
> I find it unfortunate to hardcode the list of information displayed in these Summaries, in which order, with which titles, etc.
> 
> I would prefer to have a kind of template (remember gotemplates ;) ) and the resource would be applied to the template before to display it. This way, we can imagine in the future that users (or companies) may be able to provide their own templates.
> 
> > There's empty arrays (on purpose), some elements are undefined and others are straight up empty strings, it's an absolute mess at times especially dealing with custom resources too.
> 
> For this, we could compute the value that would be displayed, but let the template decide where/if to display this data. That would help separate data and presentation.
> 
> But, anyway, that's a great work

Yeah :( I find it unfortunate too with regards to making it more generic / template-like. 

At least we do have some generic sections such as metadata.

We made an early decision to "manually" create the summary pages and not have them automated, as we will be eventually adding UI mockups such as:

Screenshot 2024-02-22 at 9 54 18 AM

back into the summary pages. 

As well as the fact that the information is "curated" and not all information would be helpful to summarize / show to the user, such as managed fields, or other fields which are more kubernetes-internal-only.

If a user wanted the "raw" information, they can have a look at Inspect tab, but for Summary we try to make it as useful as possible / being able to look at all the information immediatley.

lstocchi (approved):
Tried again with many random resources and everything is displayed correctly and as expected for me. LGTM, great work!! 🚀

packages/renderer/src/lib/kube/details/KubeIngressArtifact.svelte
```diff
@@ -0,0 +1,87 @@
...
+              Path: {path.path} •
+              {#if rule.host}
+                Link:
+                <Link
+                  externalRef="{artifact.tls && artifact.tls.length > 0 ? 'https' : 'http'}://{rule.host}{path.path}">
+                  {artifact.tls && artifact.tls.length > 0 ? 'https' : 'http'}://{rule.host}{path.path}
+                </Link>
+              {:else}
+                <!-- If rule.host is undefined, we only show the path without constructing a link -->
+                Path only: {path.path}
```
feloy: Here, the path.path will be displayed twice (already on line 65)

packages/renderer/src/lib/kube/details/KubeVolumeArtifact.svelte
```diff
@@ -0,0 +1,228 @@
...
+  {/if}
+  {#if artifact.downwardAPI}
+    <tr>
+      <Cell>Downward API</Cell>
+    </tr>
+  {/if}
+  {#if artifact.emptyDir}
+    <tr>
+      <Cell>Empty Directory</Cell>
+      <Cell>Medium: {artifact.emptyDir.medium || 'Default'}</Cell>
```
feloy: The sizeLimit information would be helpful IMO

feloy (commented):
A couple of nits, LGTM otherwise

feloy (approved):

cdrage:
Thank you! 

I've added the two requests (path being shown twice and volumeSize) being added.

Reviewed by @lstocchi and @feloy 

I'm going to enable auto-merge and start to get this in as this will be the initial "base" implementation to improve on.

Any further improvements can be in new PR's related to https://github.com/containers/podman-desktop/pull/5828#issuecomment-1927117690 comment with regards to further improving it (showing IP, etc in code font, etc.).

EDIT: Also rebasing before merge.

PR #5832, opened 2024-02-05
Copy button - Details page

### LINKED ISSUE #4241: feat: add copy button next to title in details page

trmendes:
### What does this PR do?

It makes usage of the “Copy To Clipboard” component, so the users can copy the title of the details page to the clipboard.

### Screenshot / video of UI
copy_button_details

Fixes #4241

### How to test this PR?

1) Open the detail page
2) Click on the copy button
3) Check if the title is present on your clipboard as expected

### DIFF
--- a/packages/renderer/src/lib/ui/CopyToClipboard.svelte
+++ b/packages/renderer/src/lib/ui/CopyToClipboard.svelte
@@ -12,16 +12,16 @@ function copyTextToClipboard() {
 </script>
 
 <div class="float-right">
-  <Tooltip tip="Copy to Clipboard" bottom>
+  <Tooltip tip="Copy to Clipboard" bottom class="{$$props.class || ''}">
     <button
       title="Copy To Clipboard"
-      class="ml-5 {$$props.class || ''}"
+      class="ml-3 {$$props.class || ''}"
       aria-label="Copy To Clipboard"
       on:click="{() => copyTextToClipboard()}">
       <Fa icon="{faPaste}" />
     </button>
   </Tooltip>
 </div>
-<div class="mt-1 my-auto text-xs truncate {$$props.class || ''}" aria-label="{title} copy to clipboard" title="{title}">
+<div class="mt-1 my-auto truncate {$$props.class || ''}" aria-label="{title} copy to clipboard" title="{title}">
   {title}
 </div>
--- a/packages/renderer/src/lib/ui/DetailsPage.svelte
+++ b/packages/renderer/src/lib/ui/DetailsPage.svelte
@@ -2,6 +2,7 @@
 import { lastPage, currentPage } from '../../stores/breadcrumb';
 import { router } from 'tinro';
 import Link from './Link.svelte';
+import CopyToClipboard from './CopyToClipboard.svelte';
 
 export let title: string;
 export let titleDetail: string | undefined = undefined;
@@ -39,8 +40,8 @@ function handleKeydown(e: KeyboardEvent) {
         </div>
         <div class="flex flex-col grow pr-2">
           <div class="flex flex-row">
-            <h1 aria-label="{title}" class="text-xl leading-tight">{title}</h1>
-            <div class="text-violet-400 ml-2 leading-normal" class:hidden="{!titleDetail}">{titleDetail}</div>
+            <div><CopyToClipboard title="{title}" clipboardData="{title}" class="text-xl inline-block mt-1" /></div>
+            <div class="text-violet-400 ml-2 leading-normal mt-2" class:hidden="{!titleDetail}">{titleDetail}</div>
           </div>
           <div>
             <span class="text-sm leading-none text-gray-900" class:hidden="{!subtitle}">{subtitle}</span>
--- a/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
@@ -478,7 +478,7 @@ function hasAnyConfiguration(provider: ProviderInfo) {
                 {#if container.type === 'docker'}Docker{:else if container.type === 'podman'}Podman{/if} endpoint
               </div>
               <PreferencesResourcesRenderingCopyButton
-                class="{container.status !== 'started' ? 'text-gray-900' : ''}"
+                class="text-xs {container.status !== 'started' ? 'text-gray-900' : ''}"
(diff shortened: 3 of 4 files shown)


trmendes:
Not very happy with the layout of it. Any suggestions and hints for the CSS?

benoitf:
I think it's around the vertical alignment (I have not looked at the html elements being currently used to give you the proper answer but hints would be around `items-center`, `align-middle`, etc

packages/renderer/src/lib/ui/CopyToClipboard.svelte
```diff
@@ -22,6 +22,6 @@ function copyTextToClipboard() {
     </button>
   </Tooltip>
 </div>
-<div class="mt-1 my-auto text-xs truncate {$$props.class || ''}" aria-label="{title} copy to clipboard" title="{title}">
+<div class="mt-1 my-auto truncate {$$props.class || ''}" aria-label="{title} copy to clipboard" title="{title}">
```
dgolovin, line 25: You can modify line 15 to
```
<Tooltip tip="Copy to Clipboard" bottom class="{$$props.class || ''}">
```
so you can control `<span/>` that wraps the `<slot />`

packages/renderer/src/lib/ui/DetailsPage.svelte
```diff
@@ -39,7 +40,7 @@ function handleKeydown(e: KeyboardEvent) {
         </div>
         <div class="flex flex-col grow pr-2">
           <div class="flex flex-row">
-            <h1 aria-label="{title}" class="text-xl leading-tight">{title}</h1>
+            <div><CopyToClipboard title="{title}" clipboardData="{title}" class="text-xl" /></div>
```
dgolovin: Then update classes for button to
```suggestion
            <div><CopyToClipboard title="{title}" clipboardData="{title}" class="text-xl inline-block mt-1" /></div>
```

That change wont affect style of the copy button in resources page and fix `Details page` layout

image

dgolovin (changes requested):

trmendes:
@dgolovin Thank you do much for your guidance with CSS :)

It looks much better now <3

copy1

copy2

benoitf:
I find confusing the clipboard button being in the middle of the line as I'm not sure what it could copy

image

I don't know if we need to move the sha to a newline or think in a different way or just copy the sha and not the image name (so button is at the right)

trmendes:
@benoitf I completely agree, and I experienced the same feeling of confusion when opening the panel to take the screenshot.

I might be a bit busy with an interview this week, but I will come back to it, do some experiments and post screenshots here as soon as possible. 

Thanks for your comment :)

benoitf:
@trmendes good stuff anyway

maybe to be less confusing it's to have the icon appearing only on hover and a different background including both the thing to copy and the copy button so we see that it's the 'same unit'

I'll let others comment here

trmendes:
@benoitf I took a quick look at docker desktop to see how they handle that sort of information, and they actually have a copy button for the hash and not for the object name. They also print the tag along with the name.

What is the reason(s) we choose copying the name over hash?

benoitf:
cc @mairin

axel7083:
> cc @mairin

I had something similar open _long time ago_ see @mairin comment on the subject https://github.com/containers/podman-desktop/pull/4242#issuecomment-1765216485

mairin:
@axel7083 Yep exactly what I was gonna suggest :) Here's a prototype, it's a little hairy but you get the idea:

https://design.penpot.app/#/view/43442e9d-d45f-8169-8003-daff3cb3e0b8?page-id=43442e9d-d45f-8169-8003-daff3cb3e0b9&section=interactions&index=0&share-id=93d0ad32-dfe5-8194-8003-db0444b2c3bf

[Screencast from 2024-02-06 10-34-52.webm](https://github.com/containers/podman-desktop/assets/799683/0bd9da83-716d-462b-9a90-5e7c376fcf9f)

feloy:
> I find confusing the clipboard button being in the middle of the line as I'm not sure what it could copy
> 
> image
> 
> I don't know if we need to move the sha to a newline or think in a different way or just copy the sha and not the image name (so button is at the right)

I would move the icon closer to the text it copies. Here it is exactly in the middle, the reason why IMO it is confusing.
302550209-975bb917-3a7f-4e28-90e9-cbf496c128a8

deboer-tim:
+1 to design from @mairin. That's a much cleaner look, which means we could use it in far more places (e.g. we don't need to decide between name or hash, both could allow copying without any UI clutter or confusion).

CopyToClipboard is already used in Preferences > Resources page (#5687), I think the question is whether we use it here basically as-is or refactor it to the design first. Since there are multiple comments on placement I'd prefer the latter, also so that we wouldn't add buttons and then 'remove' them later.

feloy:
> +1 to design from @mairin. That's a much cleaner look, which means we could use it in far more places (e.g. we don't need to decide between name or hash, both could allow copying without any UI clutter or confusion).
> 
> CopyToClipboard is already used in Preferences > Resources page (#5687), I think the question is whether we use it here basically as-is or refactor it to the design first. Since there are multiple comments on placement I'd prefer the latter, also so that we wouldn't add buttons and then 'remove' them later.

My only concern about these popup on which we need to click (not specifically here, but in different apps/websites), is related to accessibility and that you have to be quite precise with your mouse and don't move too much, or the popup disappears before you can click on it. And is it usable with keyboard only?

deboer-tim:
> My only concern about these popup on which we need to click (not specifically here, but in different apps/websites), is related to accessibility and that you have to be quite precise with your mouse and don't move too much, or the popup disappears before you can click on it.

Agreed, we need to make sure it is not finicky and mouse-over area is generous. Could be other things we could do too, e.g. maybe clicking on the text keeps the popup open?

> And is it usable with keyboard only?

FWIW, I just checked what to me is the 'Mac equivalent': copying a password from the password manager. When you're using the keyboard you never see the popup - when you tab in it auto-selects all the text and a normal Ctrl-C works fine.

mairin:
@deboer-tim Ah that is what I would suggest for keyboard... just enable ctrl+c on copyable fields on focus.

benoitf:
converting do draft

benoitf:
not updated, closing

PR #5844, opened 2024-02-05
feat: making donut having a loading state

### LINKED ISSUE #5837: ContainerStatistics displaying `-1` before displaying real values

When opening a container info page, the values for RAM and CPU can take some time to be displayed, (3-4 seconds), we might want to show something like `Collection` or else instead of `-1`

image

Either a loading animation, or a default message

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Adding a loading state for the `Donut.svelte` component. This is backward compatible, as the loading state is `false` by default.

> This PR only add the support to the Donut, no component are using it for now, it will be in a follow up PR

### Screenshot / video of UI

See comments 

Related to https://github.com/containers/podman-desktop/issues/5837

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/donut/Donut.spec.ts
+++ b/packages/renderer/src/lib/donut/Donut.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/renderer/src/lib/donut/Donut.spec.ts
+++ b/packages/renderer/src/lib/donut/Donut.spec.ts
@@ -30,11 +30,11 @@ test('Expect the tooltip to be in document', async () => {
 });
 
 test('Expect the tooltip to display loading', async () => {
-  render(Donut, { value: undefined, loading: true });
+  render(Donut, { value: undefined, title: 'potatoes', loading: true });
 
   const tooltip = screen.getByLabelText('tooltip');
   expect(tooltip).toBeInTheDocument();
-  expect(tooltip.textContent).toBe('Loading...');
+  expect(tooltip.textContent).toBe('Loading potatoes...');
 });
 
 test('Expect text being displayed when not loading', async () => {
@@ -44,9 +44,9 @@ test('Expect text being displayed when not loading', async () => {
   expect(text).toBeInTheDocument();
 });
 
-test('Expect text not being displayed when loading', async () => {
+test('Expect text being displayed when loading', async () => {
   render(Donut, { value: 'dummy-text', loading: true });
 
-  const text = screen.queryByText('dummy-text');
-  expect(text).toBeNull();
+  const text = screen.getByText('dummy-text');
+  expect(text).toBeInTheDocument();
 });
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/donut/Donut.spec.ts
```diff
@@ -0,0 +1,52 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
axel7083:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

deboer-tim:
I think we should still show the title during loading, otherwise I wonder why there are two spinners and have to hover over to see their purpose.

axel7083:
> I think we should still show the title during loading, otherwise I wonder why there are two spinners and have to hover over to see their purpose.

@deboer-tim Here is an updated version displaying the provided `value` even when loading

> This mean the parent is responsible of displaying an appropriate value when loading.

https://github.com/containers/podman-desktop/assets/42176370/71983da6-5e7b-4929-91b4-bc9bb136d206

packages/renderer/src/lib/donut/Donut.svelte
```diff
@@ -21,21 +22,40 @@ function describeArc(radius: number, endAngle: number) {
...
   <svg viewBox="-4 -4 {size + 8} {size + 8}" height="{size}" width="{size}">
-    <circle fill="none" class="stroke-charcoal-300" stroke-width="1" r="{size / 2}" cx="{size / 2}" cy="{size / 2}"
-    ></circle>
-    <path
-      fill="none"
-      class="{stroke}"
-      stroke-width="3.5"
-      d="{describeArc(size / 2, (percent * 360) / 100)}"
-      data-testid="arc"></path>
+    <g class:animate-spin="{loading}" class:origin-[50%_50%]="{loading}" style="transform-box: fill-box;">
```
axel7083, line 54: the `transform-box: fill-box;` is not yet supported by tailwinds, ref https://github.com/tailwindlabs/tailwindcss/pull/10982

deboer-tim:
Note that I was just suggesting the title was shown while loading, not value. :)

This got me thinking though: why introduce a new exported loading property, where we could end up with cases like 'loading but value shown' or 'not loading but no value' vs just making the loading animation appear whenever the value isn't set (or perhaps is set to an invalid value)?

axel7083:
> Note that I was just suggesting the title was shown while loading, not value. :)
> 
> This got me thinking though: why introduce a new exported loading property, where we could end up with cases like 'loading but value shown' or 'not loading but no value' vs just making the loading animation appear whenever the value isn't set (or perhaps is set to an invalid value)?

Because I wanted to be backward compatible as some elements are using the Donut component. It is honestly pretty complicated for nothing, for example, why the donut is responsible of building the tooltip (`${percent}% ${title} usage`), it is not generic, nor easily reusable. 

My personal thought:

(1) It should act as a progress bar
(2) It should have 4 properties `label: string`, `value: number`, `loading: boolean`, `color: string`
- label should be used in the tooltip + for the text
- the value should be used for the progress
- loading to make the animation or not
- color for the progress indicator

Then, a wrapper component could be created to be used for resource usage etc.

deboer-tim:
Donut isn't external API, so we only need to make sure we don't break existing places where it's used inside Podman Desktop - but we can always fix those references when we're improving the component. +1 to improving the way the tooltip works or changing title to label.

The original issue didn't talk about using Donut as a general progress widget, and our designs normally use Spinner or LinearProgress for that. Can you point me at where/how it would be used? IMHO some kind of 'if (!value) then loading = true' would be a good improvement to make either way for most of our current use.

axel7083:
Closing for now, might want to come back later -

PR #5848, opened 2024-02-05
feat: allow to create containers within a pod

lstocchi:
### What does this PR do?

This PR exposes the libPodApi createContainer function which allows to create a container inside a pod (useful in ai-studio) 

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

run tests

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -1843,10 +1843,13 @@ export class ContainerProviderRegistry {
       for (const [key, value] of Object.entries(options.HostConfig?.PortBindings)) {
         const keyAsNumber = parseInt(key);
         if (Array.isArray(value) && 'HostPort' in value[0] && !isNaN(keyAsNumber)) {
-          portmappings.push({
-            container_port: keyAsNumber,
-            host_port: value[0].HostPort,
-          });
+          const valueAsNumber = parseInt(value[0].HostPort);
+          if (!isNaN(valueAsNumber)) {
+            portmappings.push({
+              container_port: keyAsNumber,
+              host_port: valueAsNumber,
+            });
+          }
         }
       }
     }
@@ -1902,28 +1905,28 @@ export class ContainerProviderRegistry {
     if (bindItems.length < 2) {
       return undefined;
     }
-    let flags: string[] = [];
-    if (bindItems.length === 3) {
-      flags = bindItems[2].split(',');
-    }
     let mode = '';
     let propagation = 'rprivate';
-    for (const flag of flags) {
-      switch (flag) {
-        case 'Z':
-        case 'z':
-          mode = flag;
-          break;
-        case 'private':
-        case 'rprivate':
-        case 'shared':
-        case 'rshared':
-        case 'slave':
-        case 'rslave':
-          propagation = flag;
-          break;
+    if (bindItems.length === 3) {
+      const flags = bindItems[2].split(',');
+      for (const flag of flags) {
+        switch (flag) {
+          case 'Z':
+          case 'z':
+            mode = flag;
+            break;
+          case 'private':
+          case 'rprivate':
+          case 'shared':
+          case 'rshared':
+          case 'slave':
+          case 'rslave':
+            propagation = flag;
+            break;
+        }
       }
     }
+
     return {
       Destination: bindItems[1],
       Source: bindItems[0],
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -49,7 +49,9 @@ import type {
   PlayKubeInfo,
   ContainerCreateOptions as PodmanContainerCreateOptions,
   PodInfo as LibpodPodInfo,
-  ContainerCreateMountOptions,
+  ContainerCreateMountOption,
+  ContainerCreateNetNSOption,
+  ContainerCreatePortMappingOption,
 } from './dockerode/libpod-dockerode.js';
 import { LibpodDockerode } from './dockerode/libpod-dockerode.js';
 import type { ContainerStatsInfo } from './api/container-stats-info.js';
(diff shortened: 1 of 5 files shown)


benoitf:
hello, I'm worried to introduce specific Podman functions, because we tried to have a common layer and translation pattern.

For example if we try to list all containers, we have an information if it's inside a pod or not.

So could we add the pod information in the default `ContainerCreateOptions` ?

```
createContainer({pod: ...})
```


if there is a pod info we just call libPodApi or fails if no libPodApi, etc

but as a consumer you just have a single API if you need to create a container


Probably it means that `replicatePodmanContainer` should be deprecated/removed as well as we should not have any 'podman' functions
For some very specific podman feature, it means it should be provided by the extension itself than the common API

lstocchi:
> hello, I'm worried to introduce specific Podman functions, because we tried to have a common layer and translation pattern.
> 
> For example if we try to list all containers, we have an information if it's inside a pod or not.
> 
> So could we add the pod information in the default `ContainerCreateOptions` ?
> 
> ```
> createContainer({pod: ...})
> ```
> 
> if there is a pod info we just call libPodApi or fails if no libPodApi, etc
> 
> but as a consumer you just have a single API if you need to create a container
> 
> Probably it means that `replicatePodmanContainer` should be deprecated/removed as well as we should not have any 'podman' functions For some very specific podman feature, it means it should be provided by the extension itself than the common API

Updated 👍

packages/main/src/plugin/container-registry.ts
```diff
@@ -1757,6 +1736,80 @@ export class ContainerProviderRegistry {
...
+      options.Env = env;
+      // remove EnvFiles from options
+      delete options.EnvFiles;
+    }
+
+    const container = await engine.api.createContainer(options);
+    await this.attachToContainer(engine, container, options.Tty, options.OpenStdin);
+    if (options.start === true || options.start === undefined) {
+      await container.start();
+    }
```
benoitf: it looks like there is a difference then between if I submit a pod or not, 

in one case I'll be attaching streams and eventually start it but not using podman (if I say I want to have a pod) ?

packages/main/src/plugin/container-registry.ts
```diff
@@ -1757,6 +1736,80 @@ export class ContainerProviderRegistry {
...
+      entrypoint: options.Entrypoint,
+      env: updatedEnv,
+      image: options.Image,
+      pod: options.pod,
+      hostname: options.Hostname,
+      mounts: updatedMounts,
+    };
+
+    const container = await libPod.createPodmanContainer(podmanOptions);
+    return { id: container.Id };
```
benoitf: it looks like we're handling only a subset

if I look at what RunImage.svelte is giving

Ports, networks, restart policy, security policy, tty/std*,

lstocchi: To me this seems overoptimization. Nobody will use it atm, just ai-studio and this covers its needs. 
When we'll want to support the run inside a pod in the runImage.svelte page we will refine it.
I mean i can also write the code to support ports, network, ... but i will not have the ability to effectively test it without updating runImage as well

jeffmaury:
I think code should be refactored as starting a container in a pod should be similar to starting a container expect the api objects used are different so code should be shared

lstocchi:
> I think code should be refactored as starting a container in a pod should be similar to starting a container expect the api objects used are different so code should be shared

I didn't get it. You can start the container in a pod the same way of one outside a pod. You just need its id which is returned.
Or you mean sharing the logic for the start flag?

jeffmaury:
> > I think code should be refactored as starting a container in a pod should be similar to starting a container expect the api objects used are different so code should be shared
> 
> I didn't get it. You can start the container in a pod the same way of one outside a pod. You just need its id which is returned. Or you mean sharing the logic for the start flag?

No I mean the code start that's a container either in a pod or not should be very similar and shared. Here the methods `createContainerLibPod `and `createContainerDockerode` should be verify similar and share common logic

packages/main/src/plugin/container-registry.ts
lstocchi: Unified in `createContainer` so it is called in both cases

packages/main/src/plugin/container-registry.ts
lstocchi: Done

lstocchi:
@benoitf @jeffmaury updated to share the common code and extends the properties supported on libpod

benoitf (commented):
hello, could you extract the introduction of `getMatchingPodmanEngine` and `getMatchingPodmanEngineLibPod` to a separate PR

lstocchi:
> hello, could you extract the introduction of `getMatchingPodmanEngine` and `getMatchingPodmanEngineLibPod` to a separate PR

Done https://github.com/containers/podman-desktop/pull/6160

benoitf (approved):
tried to start containers using podman engine or docker engine and it was working 👍

packages/main/src/plugin/container-registry.ts
```diff
@@ -1766,6 +1757,187 @@ export class ContainerProviderRegistry {
...
+      const envFileContent = await this.getEnvFileParser().parseEnvFiles(envFiles);
+
+      const env = options.Env || [];
+      env.push(...envFileContent);
+      options.Env = env;
+      // remove EnvFiles from options
+      delete options.EnvFiles;
+    }
+
+    return await engine.api.createContainer(options);
```
benoitf, line 1785: wondering why we need to do return await and not return without await ?

benoitf, line 1785:
```suggestion
    return engine.api.createContainer(options);
```

lstocchi:
As i'm PTO next week, i'm going to merge this before the end of the day if there are no extra requests. This should be in 1.8.0 for ai-studio. Fixing the lint/format issues now

jeffmaury (approved):
LGTM

Issue #5661, opened 2024-01-24
Ability to apply a Kube YAML file to the cluster


deboer-tim:
### Is your enhancement related to a problem? Please describe

We have a Play Kubernetes YAML button in the Containers and Pods views, but nothing similar in the Kubernetes nav pages.

### Describe the solution you'd like

When you're in the new Kubernetes pages it would be really handy to have a button to quickly select a YAML file and apply it to the cluster.

odockal:
@deboer-tim Do we need to make sure to have this covered with e2e tests? I am not quite sure I understand where this new feature will land, new Kubernetes page? It is play kube yaml page or somewhere else?

deboer-tim:
@odockal the only thing that this will surface is an 'Apply YAML' button on Deployment/Service/Ingress pages that will do the equivalent of `kubectl apply -f`. Both the button (PR #5901) and underlying logic (#5849) will have their own unit tests so I'm not worried directly about the function, but it would be good to have an E2E test that would use this button and confirm that the resources appear back in the UI.

Ideally we could include this in any cluster (kind/minikube/etc) test flow - e.g. use the button to create the Service/Deployment that you want to test next.

PR #5849, opened 2024-02-05
feat: ability to apply kube yaml

deboer-tim:
### What does this PR do?

Adds a function to apply a Kubernetes yaml file, identical to how 'kubectl apply -f <file>' works. Based on the API's example but adapted to our codebase.

Exposes the new function to the renderer as window.kubernetesApply().

### Screenshot / video of UI

N/A

First step toward #5661.

### How to test this PR?

`yarn test:main` or hardcode a call to try window.kubernetesApply().

### DIFF
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -1074,8 +1074,7 @@ export class KubernetesClient {
    * @param namespace the namespace to use for any resources that don't include one
    * @return an array of resources created
    */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  async applyResources(context: string, manifests: any[], namespace?: string): Promise<KubernetesObject[]> {
+  async applyResources(context: string, manifests: unknown[], namespace?: string): Promise<KubernetesObject[]> {
     return this.syncResources(context, manifests, 'apply', namespace);
   }
 
@@ -1091,16 +1090,13 @@ export class KubernetesClient {
    * Heavily influenced by the API example:
    * https://github.com/kubernetes-client/javascript/blob/0fbfd8fc2dcc7f4ec3e6fcd64a5c55169b6ef0b8/examples/typescript/apply/apply-example.ts
    */
-
   async syncResources(
     context: string,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    manifests: any[],
+    manifests: unknown[],
     action: 'create' | 'apply',
     namespace?: string,
   ): Promise<KubernetesObject[]> {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const telemetryOptions: Record<string, any> = {
+    const telemetryOptions: Record<string, unknown> = {
       manifestsSize: manifests?.length,
       action: action,
     };
@@ -1113,7 +1109,7 @@ export class KubernetesClient {
       ctx.loadFromFile(this.kubeconfigPath);
       ctx.currentContext = context;
 
-      const validSpecs = manifests.filter(s => s?.kind);
+      const validSpecs = manifests.filter(s => (s as { kind: unknown })?.kind) as KubernetesObject[];
 
       const client = ctx.makeApiClient(KubernetesObjectApi);
       const created: KubernetesObject[] = [];
@@ -1132,7 +1128,8 @@ export class KubernetesClient {
         try {
           // try to get the resource, if it does not exist an error will be thrown and we will
           // end up in the catch block
-          await client.read(spec);
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          await client.read(spec as any);
           // we got the resource, so it exists: patch it
           //
           // Note that this could fail if the spec refers to a custom resource. For custom resources
@@ -1150,15 +1147,16 @@ export class KubernetesClient {
         }
       }
       return created;
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    } catch (error: any) {
-      console.log(error);
+    } catch (error: unknown) {
       telemetryOptions.error = error;
-      if (error?.response?.body) {
-        if (error.response.body.message) {
-          throw new Error(error.response.body.message);
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if ((error as any)?.response?.body) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const err: any = error as { response: { body: unknown } };
+        if (err.response.body.message) {
+          throw new Error(err.response.body.message);
         }
-        throw new Error(error.response.body);
+        throw new Error(err.response.body);
       }
       throw error;
     } finally {
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -1017,7 +1017,7 @@ export class KubernetesClient {
 
   // load yaml file and extract manifests
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  async loadManifestsFromFile(file: string): Promise<any[]> {
+  async loadManifestsFromFile(file: string): Promise<KubernetesObject[]> {
     // throw exception if file does not exist
     if (!fs.existsSync(file)) {
       throw new Error(`File ${file} does not exist`);
@@ -1123,8 +1123,7 @@ export class KubernetesClient {
       const client = ctx.makeApiClient(KubernetesObjectApi);
       const created: KubernetesObject[] = [];
       for (const spec of validSpecs) {
-        // this is to convince TypeScript that metadata exists even though we already filtered specs
-        // without metadata out
+        // this is to convince TypeScript that metadata exists
         spec.metadata = spec.metadata || {};
         spec.metadata.annotations = spec.metadata.annotations || {};
 
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -45,6 +45,7 @@ import {
   Watch,
   VersionApi,
   makeInformer,
+  KubernetesObjectApi,
 } from '@kubernetes/client-node';
 import type { V1Route } from './api/openshift-types.js';
 import type * as containerDesktopAPI from '@podman-desktop/api';
@@ -987,46 +988,6 @@ export class KubernetesClient {
     return tags;
   }
 
-  // load yaml file and extract manifests
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  async loadManifestsFromFile(file: string): Promise<any[]> {
-    // throw exception if file does not exist
-    if (!fs.existsSync(file)) {
-      throw new Error(`File ${file} does not exist`);
-    }
-
-    // load file and create resources
-    const content = await fs.promises.readFile(file, 'utf-8');
-
-    const manifests = parseAllDocuments(content, { customTags: this.getTags });
-    // filter out null manifests
-    return manifests.map(manifest => manifest.toJSON()).filter(manifest => !!manifest);
-  }
-
-  async createResourcesFromFile(context: string, filePath: string, namespace: string): Promise<void> {
-    let telemetryOptions = {};
-    try {
-      const manifests = await this.loadManifestsFromFile(filePath);
-      try {
-        await this.createResources(context, manifests, namespace);
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      } catch (error: any) {
-        if (error?.response?.body) {
-          if (error.response.body.message) {
-            throw new Error(error.response.body.message);
-          }
-          throw new Error(error.response.body);
-        }
-        throw error;
-      }
-    } catch (error) {
-      telemetryOptions = { error: error };
-      throw error;
-    } finally {
-      this.telemetry.track('kubernetesCreateResourcesFromFile', telemetryOptions);
-    }
-  }
-
   async getAPIResource(
     client: CustomObjectsApi,
     apiGroup: { group: string; version: string },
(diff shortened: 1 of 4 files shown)


benoitf (commented):
hello, I was wondering what would be the need of this PR

We can already apply a yaml file to a Kubernetes cluster
image
so it means everything is already there on the 'backend' part

was the difference with the new `apply` method vs existing methods like

https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/kubernetes-client.ts#L1006

or
https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/kubernetes-client.ts#L1059

as we're already handling more complex objects with custom resources and AFAIK it's doing the job of applying resources on a kubernetes cluster.

Here `apply` looks only a partial implementation so I'm lost.

deboer-tim:
createResourcesFromFile() is a rough equivalent (or createResources() if you've already loaded the file/created it programmatically). create will take a set of resources and iterate over them trying to create each one in turn. If any resources already exist or there's a problem creating it, it skips the rest and exits. This is why Play YAML doesn't work twice; even if it fails early there can be manual cleanup you need to do before trying again.

The apply() function is written more generically, but the big difference is that the contract (like kubectl apply -f) is to create _or patch_ all resources. If any exist it replaces them; if any fail it keeps trying the rest.

For a general purpose Apply YAML button the latter is what a Kube developer would expect, because it's more likely to work and force the state you want - even though it could replace existing resources. It could be that this would be better for Play YAML in the long run too, but that is a big behaviour change and I do not want to do that as part of this issue.

FWIW I could rename to applyFromFile() to allow for a non-file apply() later, or we could create a more generic function with some arguments to allow either patch/create or fail-fast/keep-trying behaviours?

cdrage:
> hello, I was wondering what would be the need of this PR
> 
> We can already apply a yaml file to a Kubernetes cluster image so it means everything is already there on the 'backend' part
> 
> was the difference with the new `apply` method vs existing methods like
> 
> https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/kubernetes-client.ts#L1006
> 
> or https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/kubernetes-client.ts#L1059
> 
> as we're already handling more complex objects with custom resources and AFAIK it's doing the job of applying resources on a kubernetes cluster.
> 
> Here `apply` looks only a partial implementation so I'm lost.

As what Tim said, there's a big difference for k8s on `kubectl create` and `kubectl apply`.

`kubectl apply` is more commonly used and is imperative management. See differences here: https://spacelift.io/blog/kubectl-apply-vs-create and apply any changes to something that has been previously created.

`kubectl create` is used more in terms of initially setting up your cluster and does not save your previous history / information. It's declarative management and better suited for non-development Im-never-changing-settings environments.

I opened up this issue a while back referencing that we should actually use `kubectl apply` like functionality for the "Play Kubernetes YAML" part: https://github.com/containers/podman-desktop/issues/4411

Honestly, after this implementation in this PR, I believe we should be replacing the code that's in "Play Kubernetes YAML" with @deboer-tim 's code for `apply`.

benoitf:
my point is that we should have only one code dealing with the creation of kubernetes objects.

Erroring if it exists or updating if it exists is like a parameter of the function.

Else we're like ending up in two different methods with different implementations, then we would have different bugs/etc

cdrage:
> my point is that we should have only one code dealing with the creation of kubernetes objects.
> 
> Erroring if it exists or updating if it exists is like a parameter of the function.
> 
> Else we're like ending up in two different methods with different implementations, then we would have different bugs/etc

I agree, however Kubernetes functions are different in that the API for simplicity, and doesn't implement apply. So erroring out if it exists and updating is common, the code that @deboer-tim has added is taken from the official kubernetes example: https://github.com/kubernetes-client/javascript/blob/master/examples/typescript/apply/apply-example.ts

I believe the two functions for Kubernetes should be separate (create vs apply) rather than using one with an added parameter as the way they error out will be similar / same functionality as `kubectl create` and `kubectl apply`

benoitf (dismissed):
we should unify/share/refactor code here
not introduce a new specific code

cdrage:
> we should unify/share/refactor code here not introduce a new specific code

Maybe we rename to applyFromFile? Intention of this is to take a straight-up YAML file from the user and apply it without manipulating anything within the file. Similar to other k8s GUI tool functionality.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1110,6 +1111,65 @@ export class KubernetesClient {
...
+      const client = ctx.makeApiClient(KubernetesObjectApi);
+      const created: KubernetesObject[] = [];
+      for (const spec of validSpecs) {
+        // this is to convince TypeScript that metadata exists even though we already filtered specs
+        // without metadata out
+        spec.metadata = spec.metadata || {};
+        spec.metadata.annotations = spec.metadata.annotations || {};
+
+        delete spec.metadata.annotations['kubectl.kubernetes.io/last-applied-configuration'];
+        spec.metadata.annotations['kubectl.kubernetes.io/last-applied-configuration'] = JSON.stringify(spec);
```
cdrage, line 1131: Since we applying this from podman desktop, we should be adding our own history / ability to "roll back" to the previous config. I believe we may have talked about this before with @feloy in a different meeting.

There are some well known annotations / labels that k8s reccommends: https://kubernetes.io/docs/reference/labels-annotations-taints/

I forget which one we used in other kubernetes tools, but there's a certain label for any tooling with k8s can use.

cdrage (changes requested):

cdrage:
We also have createResourcesFromFile: https://github.com/containers/podman-desktop/blob/420a802f462639c368efcf5072774076ea3e36ca/packages/main/src/plugin/kubernetes-client.ts#L1007 so we could just add `applyResourcesFromFile`.

packages/main/src/plugin/kubernetes-client.ts
feloy: What would be the interest of adding this? To exploit it ourselves to be able to make a rollback later? 

But the strategy for updating Kubernetes resources seems to go into another direction, to let different participants update the same resource, without conflicts. As a matter of facts, resources contain many fields, and no one is supposed to managed them all, but they are distributed between different roles (developer, operators, ...)..

The other way to be sure that what the user has sent the first time is kept as is, would be to watch the resource, and when it is modified, to make a 'server-side apply' on the resource with the original manifest, which will ensure that the fields defined in the original manifest are still the same, but won't touch the fields not specified originally, and which could have been modified by another participant.

When discussing the other day about this long annotation, I mixed up with the `metadata.managedFields` field, which contains the history of all the changes by the different participants, and the "owner" of each field.

deboer-tim:
I've been taking a look at the code. Renaming to applyResourcesFromFile() is of course easy, and/or I can make the function generic enough to do create or apply. I don't think the new function should replicate the 'fail early' behaviour though, as that's both not common in kube commands and leaves things in an odd state.

Wherever I see we've used createResources* in the code we _probably_ should have used apply and that would increase the chances of success. However, I am concerned that we've already exposed both createResource() and createResourceFromFile() in the API, and both are already used by kind cluster creation and kube play respectively. They also expose a namespace option that we don't have for apply and couldn't do the same way.

So, I'm ok with exposing a generic function that could be reused in the future and an applyResourcesFromFile() based on it - but not so comfortable with changing all the existing use and immediate testing that implies.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: I don't see any well-known labels that fit what kubectl does (mark the json last-applied from kubectl), and IMHO the benefit is dubious for the reasons @feloy pointed out, unless kubectl is literally the only tool you know is being used. For the same reason I don't think we should be creating any Podman Desktop-specific equivalent, or making any assumption that some other tool hasn't changed the resource in the meantime.

I was on the fence about removing this (and would if others felt we should) because we aren't kubectl, but thought it was relatively harmless and might help when also using kubectl or related-tools.

deboer-tim:
Rebased, renamed main method to applyResourcesFromFile(), and extracted the implementation to a applyResources() that supports both creation-only or patching behaviour.

The behaviour of applyResources() is still slightly different than the existing create* functions (it won't fail-fast if one of the resources exists), but that is intentional. It also doesn't support the optional namespace attribute yet. However, I think this could be a replacement for the logic in our existing create* methods (and/or the users could switch to apply instead).

I'm hoping this meets the middle ground where we have a path to one shared function/code path, but aren't trying to replace the create methods and all usage today.

packages/main/src/plugin/kubernetes-client.ts
feloy: Yes, I agree, it should not harm to add it, and that would make podman desktop compatible with `kubectl apply`, which it "emulates" with this functionality

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1110,6 +1111,88 @@ export class KubernetesClient {
...
+        } catch (error) {
+          // we did not get the resource, so it does not exist: create it
+          const response = await client.create(spec);
+          created.push(response.body);
+        }
+      }
+      return created;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
+      console.log(error);
```
benoitf: should it be removed ?

benoitf:
We should not have duplicated things when handling kubernetes resource saying that one day it will eventually merge

There is no 'this implementation is better', we need to pick up only one
For example I see something more generic in the resources creation in apply but I also see in the create some handling that are missing in apply.

So let say I take the same yaml file on an empty cluster, why one would work using apply and not if I use create. We'll have endless discussions finding the root cause or asking questions

Stepping back on the issue, I don't see any current usage implying a "create" mode by the extensions.
kind extension is using the method to create the ingress controller. But apply mode would fit the needs.

This is why I'm saying that 'play yaml' should use the 'apply' behaviour, but for backward compliance, we still need a create mode where it will throw an error if it exists.

And apply still require namespace handling anyway

cdrage:
> We should not have duplicated things when handling kubernetes resource saying that one day it will eventually merge
> 
> There is no 'this implementation is better', we need to pick up only one For example I see something more generic in the resources creation in apply but I also see in the create some handling that are missing in apply.
> 
> So let say I take the same yaml file on an empty cluster, why one would work using apply and not if I use create. We'll have endless discussions finding the root cause or asking questions
> 
> Stepping back on the issue, I don't see any current usage implying a "create" mode by the extensions. kind extension is using the method to create the ingress controller. But apply mode would fit the needs.
> 
> This is why I'm saying that 'play yaml' should use the 'apply' behaviour, but for backward compliance, we still need a create mode where it will throw an error if it exists.
> 
> And apply still require namespace handling anyway

We could get this PR in and in a subsequent PR, change the "Play Kubernetes YAML" functionality to use this function, so we can solve https://github.com/containers/podman-desktop/issues/4411 you are right in the way that it is incorrect how it is being used, and underneath it should of been apply from the start.

To answer your question of why would one use apply instead of create, it's because apply is imperative management and very common way of applying Kubernetes resources. It searches if anything was created already, and if it has, it "updates" it to the applied config, with roll-back enablement (last-applied-config annotation) and accurate patches.

Honestly, I rarely see documentation using `kubectl create -f file.yaml` anymore, and it's more for something like: `kubectl create pod` or something similar to that. Here is some documentation to back it up: **apply manages applications through files defining Kubernetes resources. It creates and updates resources in a cluster through running kubectl apply. This is the recommended way of managing Kubernetes applications on production. See [Kubectl Book](https://kubectl.docs.kubernetes.io/).** https://kubernetes.io/docs/reference/kubectl/quick-reference/#kubectl-apply the examples in the quick reference also use apply even for "first creation" resources.

There is also a table on declarative and imperative: https://kubectl.docs.kubernetes.io/guides/introduction/kubectl/#command-families for `create` it's for Deployment and Operations (e.g. GitOps), for `apply`, it's for `Development Only`

`kind` is correct in it's way of using create, it's "deployment and operations" and is a way to check to see if anything has been deployed before and subsequently erroring out. We should still keep createResources.

Apply is more needed for development / continued work, such as "apply YAML" or "play yaml".

apply and create are two separate paradigms / ways of thinking, and the functionalities should be in separate functions. One for deployment and operations / the other for development.

benoitf:
> apply and create are two separate paradigms / ways of thinking, and the functionalities should be in separate functions. One for deployment and operations / the other for development.

in reality I see 90% of the same lines between the 2 functions with 10% being different where it's mostly handling some bugs differently
So it can be done in the same function.

deboer-tim:
Ok, here's a pass with 4 create/readResources() with fromFile() variants, all using the same common syncResources() function (any better name?), which now supports a default namespace.

Updating all of the existing tests was a pain because they expected a specific implementation of functions and specific telemetry, but after fixing that they pass fine. I've been able to use Kube Play without issues. I haven't been able to test kind cluster creation because today it is failing for me before it gets to calling createResource(). Kube Apply button working fine with this code too.

IMHO this is a better path and cleaner going forward... but this is a fairly fundamental change and I don't know if anything else is touching this code or any alternate paths I should test.

@benoitf is this what you were looking for?

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1010,159 @@ export class KubernetesClient {
...
       const ctx = new KubeConfig();
       ctx.loadFromFile(this.kubeconfigPath);
       ctx.currentContext = context;
-      for (const manifest of manifests) {
-        // https://github.com/kubernetes-client/javascript/issues/487
-        if (manifest?.metadata?.creationTimestamp) {
-          manifest.metadata.creationTimestamp = new Date(manifest.metadata.creationTimestamp);
+
+      const validSpecs = manifests.filter(s => s?.kind);
+
```
benoitf, line 1122: it looks like we had a fix before
```
        // https://github.com/kubernetes-client/javascript/issues/487
        if (manifest?.metadata?.creationTimestamp) {
          manifest.metadata.creationTimestamp = new Date(manifest.metadata.creationTimestamp);
```

that is no longer present in the new method.
The ticket being still opened I suppose we still need the fix on manifests ?

benoitf: by finding when we added the code either we can have a test case probably

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1010,159 @@ export class KubernetesClient {
...
     try {
       const ctx = new KubeConfig();
       ctx.loadFromFile(this.kubeconfigPath);
       ctx.currentContext = context;
-      for (const manifest of manifests) {
-        // https://github.com/kubernetes-client/javascript/issues/487
-        if (manifest?.metadata?.creationTimestamp) {
-          manifest.metadata.creationTimestamp = new Date(manifest.metadata.creationTimestamp);
+
+      const validSpecs = manifests.filter(s => s?.kind);
```
benoitf: in the referenced link they were also filtering with ` && s.metadata` but it has been dropped

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1010,159 @@ export class KubernetesClient {
...
+   * @return an array of resources created
+   *
+   * Heavily influenced by the API example:
+   * https://github.com/kubernetes-client/javascript/blob/0fbfd8fc2dcc7f4ec3e6fcd64a5c55169b6ef0b8/examples/typescript/apply/apply-example.ts
+   */
+
+  async syncResources(
+    context: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    manifests: any[],
```
benoitf: avoid the any usage, use what is used inside the function like  `{kind?: unknown, metadata?:.....}[]`

packages/main/src/plugin/kubernetes-client.ts
benoitf: add cast `as {kind: unknown}[]` at the end so you don't need it

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1010,159 @@ export class KubernetesClient {
...
-          manifest.metadata.creationTimestamp = new Date(manifest.metadata.creationTimestamp);
+
+      const validSpecs = manifests.filter(s => s?.kind);
+
+      const client = ctx.makeApiClient(KubernetesObjectApi);
+      const created: KubernetesObject[] = [];
+      for (const spec of validSpecs) {
+        // this is to convince TypeScript that metadata exists even though we already filtered specs
+        // without metadata out
+        spec.metadata = spec.metadata || {};
```
benoitf: you remove the filtering on metadata so I'm not sure it's applying there
don't know why we had to remove the filtering

and if we keep the filtering we wouldn't need that instruction

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1010,159 @@ export class KubernetesClient {
...
+          const response = await client.create(spec);
+          created.push(response.body);
         }
       }
-      return undefined;
-    } catch (error) {
-      telemetryOptions = { error: error };
+      return created;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
```
benoitf: do not use any but unknown

benoitf (commented):
hi @deboer-tim yes it looks far better 👍 

I left some comments, it's no longer 'architectural issue'

I'll need to test the PR as well

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Removed.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: The problem here was due to type conversion between the Kubernetes Typescript API (timestamp is Date | undefined) vs Kubernetes REST API (date-time string | undefined). In the old path functions like createNamespacedDeploymen() were parsing the JSON into Kube API objects locally and then serializing them to send the data, so this conversion was necessary. With the new path we're doing a lower-level patch/create which AFAICT isn't trying to convert (and the apply example wasn't doing this), so shouldn't be necessary.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: I struggled with this. The reason for filtering out objects without kind or metadata is 'don't bother to send objects that aren't valid Kubernetes YAML and likely to fail in patch/create', but our old create methods did this, and many of the tests are written w/o manifests.

If I add it back then the API of createResources() changes more, as some % of calls could 'succeed' (to do nothing) where it would have failed before. Alternately, applyResources() returns the successfully created resources so you can at least compare the count, but changing createResources() to return an array and updating all callers to check is yet another big PR/change.

Thoughts? I would like to add `&& s.metadata` back but was thinking that should be future PR.

Casting as `{kind:unknown}[]` caused lots of issues further down, so I eventually went with `KubernetesObject[]` instead. client.read() still doesn't like this and couldn't find another solution so any moved to there.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Changed this (in sync and apply).

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: I think KubernetesObject is the right cast above, but it has every property optional so this is still required. I don't know any way to do 'KubernetesObject but kind and metadata are definitely there'?

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Changed, but this led to cludgy any code for checking response that I'm not sure how to deal with.

deboer-tim:
Code updated, please take another look @benoitf or @cdrage.

packages/main/src/plugin/kubernetes-client.ts
benoitf: you need to add a type that extends KubernetesObject but with kind not being optional

like KubernetesWithKindObject

when doing the filtering, you will do as MyCustomType[]; then after this line it'll assume all objects have kind being set and not being optional

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1010,157 @@ export class KubernetesClient {
     throw new Error(`Unable to find API resource for ${apiGroup.group}/${apiGroup.version}/${kind}`);
   }
 
+  // load yaml file and extract manifests
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf, line 1019: I guess it should be unknown as well ?

benoitf: or KubernetesObject or something else

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1010,157 @@ export class KubernetesClient {
...
     try {
       const ctx = new KubeConfig();
       ctx.loadFromFile(this.kubeconfigPath);
       ctx.currentContext = context;
-      for (const manifest of manifests) {
-        // https://github.com/kubernetes-client/javascript/issues/487
-        if (manifest?.metadata?.creationTimestamp) {
-          manifest.metadata.creationTimestamp = new Date(manifest.metadata.creationTimestamp);
+
+      const validSpecs = manifests.filter(s => (s as { kind: unknown })?.kind) as KubernetesObject[];
```
benoitf: I think the `(s as { kind: unknown })`should be moved to line 1095 where we defined manifests

if it's KubernetesObject then probably it's `manifests: KubernetesObject[]`

then you don't need to say s is from type {kind:unknown}

the ending cast should be a valid cast like refining a type but not goes straight from unknown to KubernetesObject (it's the manifests object that needs to be of KubernetesObject type)

packages/main/src/plugin/kubernetes-client.ts
benoitf: ok 👍

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1010,157 @@ export class KubernetesClient {
...
-      return undefined;
-    } catch (error) {
-      telemetryOptions = { error: error };
+      return created;
+    } catch (error: unknown) {
+      telemetryOptions.error = error;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if ((error as any)?.response?.body) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const err: any = error as { response: { body: unknown } };
```
benoitf: I think that if we avoid any line 1150 it's not to introduce it line 1152

we should probably do
```ts
if (error instanceOf HttpError) {
  const httpError = error as HttpError;
  if (httpError.response.body.message) {
  ...
```

deboer-tim: Fixed. Thanks, that is much better.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Agreed, fixed.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Fixed. Thanks, this does work better.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1015,159 @@ export class KubernetesClient {
...
-        if (manifest?.metadata?.creationTimestamp) {
-          manifest.metadata.creationTimestamp = new Date(manifest.metadata.creationTimestamp);
+
+      const validSpecs = manifests.filter(s => s?.kind) as KubernetesObjectWithKind[];
+
+      const client = ctx.makeApiClient(KubernetesObjectApi);
+      const created: KubernetesObject[] = [];
+      for (const spec of validSpecs) {
+        // this is to convince TypeScript that metadata exists even though we already filtered specs
+        // without metadata out
```
benoitf: need to remove this comment as it's not filtered out
```suggestion
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1049,64 +1015,159 @@ export class KubernetesClient {
...
+        spec.metadata = spec.metadata || {};
```
benoitf:
```suggestion
        spec.metadata = spec.metadata ?? {};
```

benoitf (commented):
code is LGTM (small comment to remove and replace `||` by `??`)
I'll need to test before approving but if others tested I'm no longer requested changes

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Changed to KubernetesObject.

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Fixed.

cdrage (approved):
LGTM! I was able to use the apply functionality perfectly via `kubernetesApplyResourcesFromFile`.

Amazing work 👍 

Tested this with a few "production-like" files and had no problems

benoitf:
@cdrage should probably test the startup of kind as well as this is where we create the ingress stuff

cdrage:
> @cdrage should probably test the startup of kind as well as this is where we create the ingress stuff

Checked and contour deployed fine using kind / was able to use kind fine, pods were up as well as deployment.

```
▶ k get deployments,pods -A
NAMESPACE            NAME                                     READY   UP-TO-DATE   AVAILABLE   AGE
kube-system          deployment.apps/coredns                  2/2     2            2           6m59s
local-path-storage   deployment.apps/local-path-provisioner   1/1     1            1           6m57s
projectcontour       deployment.apps/contour                  2/2     2            2           6m56s

NAMESPACE            NAME                                                     READY   STATUS      RESTARTS   AGE
kube-system          pod/coredns-787d4945fb-m655j                             1/1     Running     0          6m46s
kube-system          pod/coredns-787d4945fb-nsdvl                             1/1     Running     0          6m46s
kube-system          pod/etcd-kind-cluster-control-plane                      1/1     Running     0          7m1s
kube-system          pod/kindnet-9ztxf                                        1/1     Running     0          6m46s
kube-system          pod/kube-apiserver-kind-cluster-control-plane            1/1     Running     0          7m
kube-system          pod/kube-controller-manager-kind-cluster-control-plane   1/1     Running     0          6m59s
kube-system          pod/kube-proxy-bqffv                                     1/1     Running     0          6m46s
kube-system          pod/kube-scheduler-kind-cluster-control-plane            1/1     Running     0          6m59s
local-path-storage   pod/local-path-provisioner-75f5b54ffd-9kp7v              1/1     Running     0          6m46s
projectcontour       pod/contour-74866bdd99-th2lm                             1/1     Running     0          6m46s
projectcontour       pod/contour-74866bdd99-zkvfv                             1/1     Running     0          6m46s
projectcontour       pod/contour-certgen-v1.24.2-wmh2k                        0/1     Completed   0          6m47s
projectcontour       pod/envoy-78n6c                                          2/2     Running     0          6m44s
```

packages/main/src/plugin/kubernetes-client.ts
deboer-tim: Fixed.

benoitf:
then probably you're good to merge here as Charlie tested

deboer-tim:
Thanks!

PR #5850, opened 2024-02-05
feat: kubernetes connection store

axel7083:
### What does this PR do?

The connection status of the current context must be accessible to the frontend. This PR is adding a `kubernetesConnection` store.

Related to https://github.com/containers/podman-desktop/issues/5401

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -155,6 +155,7 @@ import { WebviewRegistry } from './webview/webview-registry.js';
 import type { IDisposable } from './types/disposable.js';
 
 import { KubernetesUtils } from './kubernetes-util.js';
+import type { KubernetesConnection } from './kubernetes-connection.js';
 
 type LogType = 'log' | 'warn' | 'trace' | 'debug' | 'error';
 
@@ -2007,6 +2008,12 @@ export class PluginSystem {
 
     this.ipcHandle('kubernetes-client:getDetailedContexts', async (): Promise<KubeContext[]> => {
       return kubernetesClient.getDetailedContexts();
+    });
+
+    this.ipcHandle('kubernetes-client:getConnectionStatus', async (): Promise<KubernetesConnection | undefined> => {
+      const value = kubernetesClient.getConnectionStatus();
+      console.log('getConnectionStatus', value);
+      return value;
     });
 
     this.ipcHandle('kubernetes-client:getClusters', async (): Promise<Cluster[]> => {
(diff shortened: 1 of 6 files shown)


packages/main/src/plugin/kubernetes-connection.ts
```diff
@@ -0,0 +1,26 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
deboer-tim, line 2: Nit: 2024

deboer-tim (changes requested):
+1 to this function, we really need it.

My main concern is that to get the list of contexts or what the current one is you use stores/kubernetes-contexts, but then to get the status of the current context you use another store that also says what the context is. When changing current context it's not clear how in sync these are or whether/how clients would deal with when the connection context isn't the current context.

There's also PR #5329 which overlaps on connection status. The one nice aspect there is that it's a map, so it doesn't duplicate what the current context is and can store connection info for multiple contexts.

Maybe a group chat would help, or you could discuss with @feloy?

packages/renderer/src/stores/kubernetes-connection.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { writable, type Writable } from 'svelte/store';
+import { EventStore } from './event-store';
+import { checkForUpdate } from './kubernetes-contexts';
+import type { KubernetesConnection } from '../../../main/src/plugin/kubernetes-connection';
+
+const windowEvents: string[] = ['kubernetes-context-update', 'kubernetes-connection-update'];
```
jeffmaury, line 24: I don't see where kubernetes-connection-update is being sent

packages/main/src/plugin/index.ts
```diff
@@ -2009,6 +2010,12 @@ export class PluginSystem {
       return kubernetesClient.getDetailedContexts();
     });
 
+    this.ipcHandle('kubernetes-client:getConnectionStatus', async (): Promise<KubernetesConnection | undefined> => {
+      const value = kubernetesClient.getConnectionStatus();
+      console.log('getConnectionStatus', value);
```
jeffmaury, line 2015: To be removed

axel7083:
Closing in favour of https://github.com/containers/podman-desktop/pull/5917/files#diff-ea47798d4fd89a4b135b3c7c7e1be18d4ba792cb8dd900425946693f0e7a53c9

PR #5880, opened 2024-02-06
fix: removing skipLibCheck option in `package/extension-api/tsconfig.json`

### LINKED ISSUE #5879: typescript configuration issue lead to problems with certain interface/types not being defined
### Bug description

Some elements in the extension-api are not defined, and there are not error where it should be. Leading to potential definition missed

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Adds the missing interfaces/types definition in the `extension-api.d.ts` file.

Fixes https://github.com/containers/podman-desktop/issues/5879

### How to test this PR?

- [x] compile without the skipLibCheck option

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -526,7 +526,7 @@ declare module '@podman-desktop/api' {
      */
     export function getProviderLifecycleContext(
       providerId: string,
-      providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+      providerConnectionInfo: ContainerProviderConnection,
     ): LifecycleContext;
   }
 
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -520,13 +520,13 @@ declare module '@podman-desktop/api' {
      * If no context is found it throws an error
      *
      * @param providerId the provider id
-     * @param providerConnectionInfo the connection to retrieve the lifecycle context for
+     * @param containerProviderConnection the connection to retrieve the lifecycle context for
      * @returns the lifecycle context
      * @throws if no provider with the id has been found or there is no context associate to it.
      */
     export function getProviderLifecycleContext(
       providerId: string,
-      providerConnectionInfo: ContainerProviderConnection,
+      containerProviderConnection: ContainerProviderConnection,
     ): LifecycleContext;
   }
 
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2846,12 +2846,12 @@ declare module '@podman-desktop/api' {
     // name of the volume to create
     Name: string;
     // Set the provider to use, if not we will try select the first one available (sorted in favor of Podman).
-    provider?: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection;
+    provider?: ContainerProviderConnection;
   }
 
   export interface VolumeDeleteOptions {
     // Set the provider to use, if not we will try select the first one available (sorted in favor of Podman).
-    provider?: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection;
+    provider?: ContainerProviderConnection;
   }
 
   export interface VolumeCreateResponseInfo {
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -1146,10 +1146,7 @@ describe('buildImage', () => {
         containerFile: 'file',
         tag: 'name',
         platform: '',
-        provider: {
-          ...connection,
-          status: () => connection.status,
-        },
+        provider: connection,
       }),
     ).rejects.toThrow('no running provider for the matching container');
   });
@@ -1235,10 +1232,7 @@ describe('buildImage', () => {
         containerFile: 'file',
         tag: 'name',
         platform: '',
-        provider: {
-          ...connection,
-          status: () => connection.status,
-        },
+        provider: connection,
       }),
     ).rejects.toThrow('human error message');
   });
@@ -1323,10 +1317,7 @@ describe('buildImage', () => {
       containerFile: '\\path\\file',
       tag: 'name',
       platform: '',
-      provider: {
-        ...connection,
-        status: () => connection.status,
-      },
+      provider: connection,
     });
 
     expect(dockerAPI.buildImage).toBeCalledWith({} as NodeJS.ReadableStream, {
@@ -1425,10 +1416,7 @@ describe('buildImage', () => {
       containerFile: '/dir/dockerfile',
       tag: 'name',
       platform: '',
-      provider: {
-        ...connection,
-        status: () => connection.status,
-      },
+      provider: connection,
       ...extraArgs,
     });
 
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -1146,7 +1146,10 @@ describe('buildImage', () => {
         containerFile: 'file',
         tag: 'name',
         platform: '',
-        provider: connection,
+        provider: {
+          ...connection,
+          status: () => connection.status,
+        },
       }),
     ).rejects.toThrow('no running provider for the matching container');
   });
@@ -1232,7 +1235,10 @@ describe('buildImage', () => {
         containerFile: 'file',
         tag: 'name',
         platform: '',
-        provider: connection,
+        provider: {
+          ...connection,
+          status: () => connection.status,
+        },
       }),
     ).rejects.toThrow('human error message');
   });
@@ -1317,7 +1323,10 @@ describe('buildImage', () => {
       containerFile: '\\path\\file',
       tag: 'name',
       platform: '',
-      provider: connection,
+      provider: {
+        ...connection,
+        status: () => connection.status,
+      },
     });
 
     expect(dockerAPI.buildImage).toBeCalledWith({} as NodeJS.ReadableStream, {
@@ -1416,7 +1425,10 @@ describe('buildImage', () => {
       containerFile: '/dir/dockerfile',
       tag: 'name',
       platform: '',
-      provider: connection,
+      provider: {
+        ...connection,
+        status: () => connection.status,
+      },
       ...extraArgs,
     });
 
@@ -3210,7 +3222,10 @@ test('check createPod uses running podman connection if ProviderContainerConnect
 
   const result = await containerRegistry.createPod({
     name: 'pod',
-    provider: containerProviderConnection,
+    provider: {
+      ...containerProviderConnection,
+      status: () => containerProviderConnection.status,
+    },
   });
   expect(result.Id).equal('id');
   expect(result.engineId).equal('podman1');
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -3222,10 +3222,7 @@ test('check createPod uses running podman connection if ProviderContainerConnect
 
   const result = await containerRegistry.createPod({
     name: 'pod',
-    provider: {
-      ...containerProviderConnection,
-      status: () => containerProviderConnection.status,
-    },
+    provider: containerProviderConnection,
   });
   expect(result.Id).equal('id');
   expect(result.engineId).equal('podman1');
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -726,7 +726,7 @@ export class ExtensionLoader {
       getProviderLifecycleContext(
         providerId: string,
         providerConnectionInfo:
-          containerDesktopAPI.ContainerProviderConnection
+          | containerDesktopAPI.ContainerProviderConnection
           | ProviderContainerConnectionInfo
           | ProviderKubernetesConnectionInfo,
       ): containerDesktopAPI.LifecycleContext {
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -726,10 +726,10 @@ export class ExtensionLoader {
         providerId: string,
         containerProviderConnection: containerDesktopAPI.ContainerProviderConnection,
       ): containerDesktopAPI.LifecycleContext {
-        return providerRegistry.getMatchingProviderLifecycleContextByProviderId(providerId, {
-          ...containerProviderConnection,
-          status: containerProviderConnection.status(),
-        });
+        return providerRegistry.getMatchingProviderLifecycleContextByProviderId(
+          providerId,
+          containerProviderConnection,
+        );
       },
     };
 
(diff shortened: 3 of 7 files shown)


axel7083:
cc @benoitf the `extensionApi.provider.getProviderLifecycleContext` is kinda annoying (see linter), and I don't really know how to fix it.

> To summarize, before this PR, the linter was not able to know the types, so it was ignoring it. Now the argument `providerConnectionInfo` of the `getProviderLifecycleContext` takes `ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo` **but** in the podman extension, we provide `ContainerProviderConnection` which has some differences, and make it incompatible.

dgolovin (changes requested):
@axel7083 It looks like ProviderContainerConnectionInfo and ProviderKuberneterConnectionInfo are DAO objects leaked from internal API https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/api/provider-info.ts.

Have not tested it all the way, but if you remove them form API definition file nothing is affected.

benoitf:
looks like it's stale. AFAIK what @dgolovin said is accurate

axel7083:
> looks like it's stale. AFAIK what @dgolovin said is accurate

I removed `ProviderContainerConnectionInfo` and `ProviderKubernetesConnectionInfo`, but we are facing many other issues

packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
```diff
@@ -86,7 +86,14 @@ async function doCreatePodFromContainers() {
...
   // first create pod
-  const { Id, engineId } = await window.createPod({ name: podCreation.name, portmappings, provider: selectedProvider });
+  const { Id, engineId } = await window.createPod({
+    name: podCreation.name,
+    portmappings,
+    provider: {
+      ...selectedProvider,
+      status: () => selectedProvider?.status ?? 'unknown',
+    },
+  });
```
benoitf: AFAIK we should not have to do this

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -1416,7 +1425,10 @@ describe('buildImage', () => {
       containerFile: '/dir/dockerfile',
       tag: 'name',
       platform: '',
-      provider: connection,
+      provider: {
+        ...connection,
+        status: () => connection.status,
+      },
```
benoitf: IMHO the test is correct without this change

benoitf (changes requested):
some changes are not expected.

Client should not have to provide status() method

axel7083:
> some changes are not expected.
> 
> Client should not have to provide status() method

The question is deeper. `ProviderContainerConnectionInfo` status field is a string, where `ContainerProviderConnection` is a function returning a string.

_We were lucky in the past that it did not raised an error_ but we were passing the wrongs objects, and the typing definitions are different. We need to discuss how we want to solve that

benoitf:
> The question is deeper. ProviderContainerConnectionInfo status field is a string, where ContainerProviderConnection is a function returning a string.

yes it's expected

benoitf:
one object is created by the providers (on the backend side) with methods
the other object is the one that UI can use and they're 'plain' object, no functions

axel7083:
@benoitf this PR need to be divided, and what we want to do need to be cleared up because I am not seeing what we expect here. We need to change some functions signature to be able to handle some cases here

benoitf:
yes 👍

packages/main/src/plugin/extension-loader.ts
```diff
@@ -725,9 +724,12 @@ export class ExtensionLoader {
...
       getProviderLifecycleContext(
         providerId: string,
-        providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+        containerProviderConnection: containerDesktopAPI.ContainerProviderConnection,
       ): containerDesktopAPI.LifecycleContext {
-        return providerRegistry.getMatchingProviderLifecycleContextByProviderId(providerId, providerConnectionInfo);
+        return providerRegistry.getMatchingProviderLifecycleContextByProviderId(providerId, {
+          ...containerProviderConnection,
+          status: containerProviderConnection.status(),
+        });
```
benoitf: I think it's inside providerRegistry.getMatchingProviderLifecycleContextByProviderId that we should add the 
providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo

benoitf:
@axel7083 if you rebase and include https://github.com/containers/podman-desktop/pull/6203 I think it'll reduce the number of changes and remove the need to provide status() method for at least createPod

packages/main/src/plugin/extension-loader.ts
axel7083: @benoitf currently the `getMatchingProviderLifecycleContextByProviderId` is the following

```ts
getMatchingProviderLifecycleContextByProviderId(
    providerId: string,
    providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
  ): LifecycleContextImpl
```

you want to make the following ?
```diff
- providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+ providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo | containerDesktopAPI.ContainerProviderConnection,
```

packages/main/src/plugin/container-registry.spec.ts
axel7083: No. Inside containerDesktopAPI.BuildImageOptions we have the following
```
```

Which is **not** compatible with ProviderContainerConnectionInfo.

benoitf: will need another change in the BuildImageOptions like for createPod
the test should not be changed

packages/main/src/plugin/extension-loader.ts
benoitf: I guess so

axel7083:
> @axel7083 if you rebase and include #6203 I think it'll reduce the number of changes and remove the need to provide status() method for at least createPod

@benoitf I have a proposal. The `ProviderContainerConnectionInfo` (internally) and `ContainerProviderConnection` (api) share 3 identical properties 
- `name: string`
- `type: 'docker' | 'podman'`
- `endpoint: ContainerProviderConnectionEndpoint`

Most of the methods `getMatchingContainerProvider`, `getMatchingEngineFromConnection`, `buildImage` etc. would have enough with an interface with only those properties.
Let's create the following inside the extension-api.d.ts

```ts
export interface BaseContainerConnectionProvider {
    name: string;
    type: 'docker' | 'podman';
    endpoint: ContainerProviderConnectionEndpoint;
  }
```

> If you have a better name I would be willing to change it

and make the `ProviderContainerConnectionInfo` and `ContainerProviderConnection` **extend** containerDesktopAPI.BaseContainerConnectionProvider.

If you agree with this idea, I can make another PR for this

packages/extension-api/src/extension-api.d.ts
```diff
@@ -827,7 +827,7 @@ declare module '@podman-desktop/api' {
     /**
      * A description of the field to be show (Markdown format)
      */
-    markdownDescription?;
+    markdownDescription?: string;
```
benoitf: maybe issue a small PR with the changes related to adding the missing types as it should work out of the box

this one, new NotificationType, IPAM interface, DeviceRequest etc

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -3210,7 +3222,10 @@ test('check createPod uses running podman connection if ProviderContainerConnect
 
   const result = await containerRegistry.createPod({
     name: 'pod',
-    provider: containerProviderConnection,
+    provider: {
+      ...containerProviderConnection,
+      status: () => containerProviderConnection.status,
+    },
```
benoitf: should be removed with the new changes

packages/extension-api/src/extension-api.d.ts
axel7083: Yes, this PR was suppose to be small enough for those changes, but it start to get too big. Agreed

packages/main/src/plugin/index.ts
```diff
@@ -1069,7 +1069,10 @@ export class PluginSystem {
             containerFile: relativeContainerfilePath,
             tag: imageName,
             platform,
-            provider: selectedProvider,
+            provider: {
+              ...selectedProvider,
+              status: () => selectedProvider.status,
+            },
```
benoitf: here as well we should remove this change and make the method accepting the info object

benoitf:
> and make the ProviderContainerConnectionInfo and ContainerProviderConnection extend containerDesktopAPI.BaseContainerConnectionProvider.

I disagree internal types should not extend extension API types

packages/main/src/plugin/extension-loader.ts
axel7083: Done

axel7083:
@benoitf linter say yes ✅ !

packages/main/src/plugin/provider-registry.ts
```diff
@@ -812,7 +815,10 @@ export class ProviderRegistry {
 
   getMatchingProviderLifecycleContextByProviderId(
     providerId: string,
-    providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+    providerConnectionInfo:
+      | ProviderContainerConnectionInfo
```
benoitf, line 819:
```suggestion
      ProviderContainerConnectionInfo
```
should we remove this first `|`

packages/main/src/plugin/provider-registry.ts
```diff
@@ -789,7 +789,10 @@ export class ProviderRegistry {
 
   getMatchingConnectionLifecycleContext(
     internalId: string,
-    providerContainerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+    providerContainerConnectionInfo:
+      | ProviderContainerConnectionInfo
```
benoitf, line 793:
```suggestion
      ProviderContainerConnectionInfo
```

should we remove this first `|`

packages/main/src/plugin/extension-loader.ts
```diff
@@ -725,7 +725,10 @@ export class ExtensionLoader {
       },
       getProviderLifecycleContext(
         providerId: string,
-        providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+        providerConnectionInfo:
+          | containerDesktopAPI.ContainerProviderConnection
```
benoitf: should we remove this first `|`
```suggestion
          containerDesktopAPI.ContainerProviderConnection
```

packages/main/src/plugin/provider-registry.ts
```diff
@@ -921,7 +927,10 @@ export class ProviderRegistry {
 
   getMatchingConnectionFromProvider(
     internalProviderId: string,
-    providerContainerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+    providerContainerConnectionInfo:
+      | ProviderContainerConnectionInfo
```
benoitf, line 931:
```suggestion
      ProviderContainerConnectionInfo
```

axel7083:
> should we remove this first |

@benoitf prettier do not want it to be removed :(

benoitf (approved):
nice enhancement 🎉

axel7083:
@dgolovin waiting on your `change requested`

packages/main/src/plugin/extension-loader.ts
```diff
@@ -725,7 +725,10 @@ export class ExtensionLoader {
...
+          | ProviderContainerConnectionInfo
+          | ProviderKubernetesConnectionInfo,
```
```suggestion
        providerConnectionInfo: containerDesktopAPI.ContainerProviderConnection
```

IMO there is no need for union type here, because you cannot use internal types from client extension.

packages/main/src/plugin/provider-registry.ts
```diff
@@ -812,7 +815,10 @@ export class ProviderRegistry {
 
   getMatchingProviderLifecycleContextByProviderId(
     providerId: string,
-    providerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+    providerConnectionInfo:
+      | ProviderContainerConnectionInfo
+      | ProviderKubernetesConnectionInfo
+      | ContainerProviderConnection,
```
dgolovin, line 821: This method called only from extension loader and only with objects that is instance of `ContainerProviderConnection', so no union type is needed as well.

```suggestion
    providerConnectionInfo: ContainerProviderConnection,
```

packages/main/src/plugin/provider-registry.ts
```diff
@@ -789,7 +789,10 @@ export class ProviderRegistry {
 
   getMatchingConnectionLifecycleContext(
     internalId: string,
-    providerContainerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+    providerContainerConnectionInfo:
+      | ProviderContainerConnectionInfo
+      | ProviderKubernetesConnectionInfo
+      | ContainerProviderConnection,
```
dgolovin, line 795: This union type is fine, because it is called from renderer with info objects and from extension-loader.

packages/main/src/plugin/provider-registry.ts
```diff
@@ -921,7 +927,10 @@ export class ProviderRegistry {
 
   getMatchingConnectionFromProvider(
     internalProviderId: string,
-    providerContainerConnectionInfo: ProviderContainerConnectionInfo | ProviderKubernetesConnectionInfo,
+    providerContainerConnectionInfo:
+      | ProviderContainerConnectionInfo
+      | ProviderKubernetesConnectionInfo
+      | ContainerProviderConnection,
```
dgolovin, line 933: This is required too, because it is called from ```getMatchingProviderLifecycleContextByProviderId```

dgolovin (approved):
LGTM
nit: some union types are not required.
Compiled and run locally to test suggested changes. Works for me.

packages/main/src/plugin/extension-loader.ts
```diff
@@ -725,10 +725,7 @@ export class ExtensionLoader {
       },
       getProviderLifecycleContext(
         providerId: string,
-        providerConnectionInfo:
-          | containerDesktopAPI.ContainerProviderConnection
-          | ProviderContainerConnectionInfo
-          | ProviderKubernetesConnectionInfo,
+        providerConnectionInfo: containerDesktopAPI.ContainerProviderConnection
```
axel7083:
```suggestion
        providerConnectionInfo: containerDesktopAPI.ContainerProviderConnection,
```

axel7083:
🚀

benoitf:
looks like this PR has broken the website build

when looking at the PR check there were failures 
https://github.com/containers/podman-desktop/actions/runs/8137643006/job/22236482437

but it was reported as ✅ 

skipLibCheck might be related there

axel7083:
> looks like this PR has broken the website build
> 
> when looking at the PR check there were failures https://github.com/containers/podman-desktop/actions/runs/8137643006/job/22236482437
> 
> but it was reported as ✅
> 
> skipLibCheck might be related there

I will take a look

benoitf:
https://github.com/containers/podman-desktop/pull/6251 in the meantime as it broke the website

axel7083:
> #6251 in the meantime as it broke the website

I opened https://github.com/tgreyuk/typedoc-plugin-markdown/issues/583 to warn about the problem of the exit code being 0 when the Typedoc fails. 
I also open a PR to fix the issue on their side https://github.com/tgreyuk/typedoc-plugin-markdown/pull/584

benoitf:
thanks @axel7083

Issue #5335, opened 2023-12-20
Docs: Improve introduction page


slemeur:
### Is your enhancement related to a problem? Please describe

The intro documentation is not welcoming the user and enticing to user the product: https://podman-desktop.io/docs/intro

### Describe the solution you'd like

The introduction is the documentation should provide a good summary of the project and entice the user to read more. From a quick scan of the page, the user might caught interest on the product. 

For this, it’s important that we provide a short introduction of the Project:
- What it is, what it does 
- why we created it
It’s important to differentiate ourselves from the beginning and we should be reusing the talking points we have on the website.

I would add, that we could eventually show a demo of the product, by attaching a link to the youtube short overview from Aine.

Then, we want to guide the user on how to start

We should also provide another set of links:
- Issue tracker
- Troubleshooting
- Community

themr0c:
This is a very unusual request for a documentation landing page.

Usually, the purpose of the documentation landing page is to provide a quick overview of the docs catalog, and guide the user to find quickly the documentation he needs.

Some examples:

- https://docs.docker.com/
- https://docs.ansible.com/
- https://stripe.com/docs
- https://www.twilio.com/docs
- https://quarkus.io/guides/

themr0c:
What seems quite usual is to have a series of cards, each card representing a major user story in the documentation.

I was searching for a way to display things as cards in docusaurus docs, could not find it, until I bumped on a screenshot in #5881. Convergent minds :)

So it seems there is a way in docusaurus to define "categories" and display them as cards.

1. Define the categories and their summary https://docusaurus.io/docs/sidebar/autogenerated#category-item-metadata
2. Use this metatata in the general intro, and in each category index.

@cdrage I need to sync with you on this, you are more advanced on the topic.

cdrage:
> What seems quite usual is to have a series of cards, each card representing a major user story in the documentation.
> 
> I was searching for a way to display things as cards in docusaurus docs, could not find it, until I bumped on a screenshot in #5881. Convergent minds :)
> 
> So it seems there is a way in docusaurus to define "categories" and display them as cards.
> 
> 1. Define the categories and their summary https://docusaurus.io/docs/sidebar/autogenerated#category-item-metadata
> 2. Use this metatata in the general intro, and in each category index.
> 
> @cdrage I need to sync with you on this, you are more advanced on the topic.

I think this would be a bad way of showing the introduction as cards create clutter / creates a sort of "table of contents".

[...]

PR #5881, opened 2024-02-06
docs: sidebar improvements / expansion / bolded css

### LINKED ISSUE #5338: Doc: Improve left-sidebar navigation

When you arrive on the documentation, we have a lot of sections but it's hard to know what's in each of the section. It forces the reader to "unflod" each of the section to have an idea of what's into it. 

Screenshot 2023-12-20 at 10 18 44

We would like to have 2 levels of sections unfolded by default.
Also, we would like to visually differentiate the levels of navigation.

Example:
Screenshot 2023-12-20 at 10 21 16
(link: https://ionicframework.com/docs)

_No response_

### Additional context

_No response_

cdrage:
docs: sidebar improvements / expansion / bolded css

### What does this PR do?

* Adds "2 levels deep" expansion by default
* Reorganizes the categories in a better way. For example, setting up a
  proxy environment and troubleshooting is more at the bottom of the
  documentation rather than the top
* Adds bold css for the category so it's easier to look at / navigate /
  read.

### Screenshot / video of UI

Screenshot 2024-02-06 at 10 13 26 AM

Screenshot 2024-02-06 at 10 14 52 AM

Closes [#5388](https://github.com/containers/podman-desktop/issues/5338)

### How to test this PR?

View the website and expand the pages


cdrage:
Note that we still auto-generate the "children" files / folders. But we have to make sure that when we add a new category, we've got to update the sidebar config too.

deboer-tim:
Can you break this up/focus on just the non-generated sidebar here? I don't love having to keep it in sync, but I think it changes rarely enough and there is minimal duplication, so I'd be ok with it given the benefit.

For styling, notice that the top Introduction title isn't styled, b/c it's not a section. :) I figured this out and could apply the style in another PR.

Reordering might be contentious, I thought someone had said they wanted Troubleshooting last - easier to review and debate if it's separate, and won't hold this up.

cdrage:
> Can you break this up/focus on just the non-generated sidebar here? I don't love having to keep it in sync, but I think it changes rarely enough and there is minimal duplication, so I'd be ok with it given the benefit.
> 
> For styling, notice that the top Introduction title isn't styled, b/c it's not a section. :) I figured this out and could apply the style in another PR.
> 
> Reordering might be contentious, I thought someone had said they wanted Troubleshooting last - easier to review and debate if it's separate, and won't hold this up.

Done! I've removed the CSS change and moved troubleshooting to the bottom.

deboer-tim:
Expanding only to second level: working great!

Unfortunately there's something going wrong with the categories: every category that has content is now showing up 'within itself'. e.g. on the live website I can select Installation to see one doc, and it has 3 children (one for each OS). When testing this PR I can't click on Installation, and it has 4 children: one for each OS and a 'second' Installation topic that contains what was in the parent before. Another example: we already had a Containers > Containers entry, but now there are two of them named Containers.

Personally I'd be ok if the categories weren't selectable, but that's likely a far bigger change since it goes down to every level, and we'd have to move each existing topic to be the first child and have a better name...

I also noticed Restricted Environments was the third option before, now it is called Proxy Environment and third-last.

cdrage:
> Expanding only to second level: working great!
> 
> Unfortunately there's something going wrong with the categories: every category that has content is now showing up 'within itself'. e.g. on the live website I can select Installation to see one doc, and it has 3 children (one for each OS). When testing this PR I can't click on Installation, and it has 4 children: one for each OS and a 'second' Installation topic that contains what was in the parent before. Another example: we already had a Containers > Containers entry, but now there are two of them named Containers.
> 
> Personally I'd be ok if the categories weren't selectable, but that's likely a far bigger change since it goes down to every level, and we'd have to move each existing topic to be the first child and have a better name...
> 
> I also noticed Restricted Environments was the third option before, now it is called Proxy Environment and third-last.

Okay, so after some digging and back and forth I was able to get this to work. The problem was that we were using `index.md` files as a makeshift table of contents / summary page, which ended up showing within the categories.

Docusaurus has one actually built-in with:

```
      link: {
        type: 'generated-index',
      },
```

I noticed that when updating this, some of our documentation isn't consistent with the "generated TOC / index section". 

So instead of this:

Screenshot 2024-02-06 at 3 32 15 PM

We could now have this:

Screenshot 2024-02-06 at 3 19 46 PM

It's _sort of_ out of scope of this PR, but these generated parts gets rid of the "index.md" problem as well as make the categories more consistent.

There was one document that for some reason had installation instructions in the "index" page, which I had moved, but that was the only caveat.

Let me know what you think.

themr0c:
After reading the comments, I understand why all the `index.md` are gone. I would like to see the build!! It might change drastically how to work on #5335, as my first questions for this issue were: 
- How could I display the sections as "cards" rather than a bare list?
- How can it be generated automatically?

website/sidebars.js
```diff
@@ -4,9 +4,188 @@
 const sidebars = {
   // Generate sidebar from the docs folder (or versioned_docs/<version>)
   mySidebar: [
+    'intro',
     {
-      type: 'autogenerated',
-      dirName: '.',
```
themr0c, line 9: Does this mean that we would need to manually add entries to the navigation bar? I'd prefer not :/

cdrage:
@themr0c Updated! It was indeed an issue, I had to add `slug: `/installation'` for the sidebar config.

As per our meeting, let's get this merged in and then further improve the "category" pages after the PR.

For others, the reasoning why we cannot use a 100% auto-generated sidebar is because docusaurus doesn't allow "expanding only 2 levels deep", and thus we have to create categories and specifically mark them as collapsed: false.

website/sidebars.js
cdrage: Marking as resolved as fabrice and I talked about this in a meeting, but TLDR is yes we do as docusaurus doesn't allow specifying how "many levels deep" to only expand.

deboer-tim (changes requested):
I have to be honest - I don't like these pages at all:
Screenshot 2024-02-13 at 4 18 15 PM

To me they don't fit in with our website or provide any real visual improvement, and as much as many of the previous index pages were minimal I like them better. I'd prefer we stick with one-change-per-PR, as I'd approve the sidebar change w/o the rest (here, or use a different PR for that part).

cdrage:
I'm closing this for now as @themr0c is looking to add some new changes to the intro pages. This PR was a nice way to figure out how to get 2-depth expanding working and works well as a reference PR.

Issue #5812, opened 2024-02-02
Document the API elements


feloy:
### Is your enhancement related to a problem? Please describe

Now there is a generated API documentation available online (https://podman-desktop.io/api), the elements of the API need to be documented.

### Describe the solution you'd like

Add source documentation

axel7083:
We should probably have some kind of coverage, or some metrics to track the documentation of the apis. In the meantime I made a quick program using `typedoc` to list the objects missing documentations

<details>
  <summary>Documentation converage</summary>

## Namespaces (11/112)
 - [x] authentication: `Namespace` ✅
   - [ ] getSession: `Function` ❌
   - [ ] onDidChangeSessions: `Function` ❌
   - [ ] registerAuthenticationProvider: `Function` ❌
 - [x] cli: `Namespace` ✅
   - [ ] createCliTool: `Function` ❌
 - [ ] commands: `Namespace` ❌
   - [ ] executeCommand: `Function` ❌ (#6353)
   - [ ] registerCommand: `Function` ❌ (#6353)
 - [ ] configuration: `Namespace` ❌
   - [ ] getConfiguration: `Function` ❌
   - [ ] onDidChangeConfiguration: `Function` ❌
 - [ ] containerEngine: `Namespace` ❌
   - [x] buildImage: `Function` ❌ (#5891)
   - [x] createContainer: `Function` ❌ (#5891)
   - [ ] createNetwork: `Function` ❌
   - [ ] createPod: `Function` ❌
   - [ ] createVolume: `Function` ❌
   - [x] deleteContainer: `Function` ❌ (#5891)
   - [x] deleteImage: `Function` ❌ (#5891)
   - [ ] deleteVolume: `Function` ❌
   - [x] getImageInspect: `Function` ❌ (#5891)
   - [ ] info: `Function` ❌
   - [x] inspectContainer: `Function` ❌ (#5891)
   - [x] listContainers: `Function` ❌ (#5891)
   - [x] listImages: `Function` ❌ (#5891)
   - [ ] listNetworks: `Function` ❌
   - [ ] listVolumes: `Function` ❌
   - [x] logsContainer: `Function` ❌ (#5891)
   - [ ] onEvent: `Function` ❌
   - [x] pullImage: `Function` ❌ (#5891)
   - [x] pushImage: `Function` ❌ (#5891)
   - [ ] replicatePodmanContainer: `Function` ❌
   - [x] saveImage: `Function` ❌ (#5891)
[...]

PR #5891, opened 2024-02-07
docs: container and image related methods of containerEngine api

feloy:

### What does this PR do?

Document Container and Image-related methods on the containerEngine namespace of the API

More text formatting to come...

Part of #5812

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2200,12 +2200,16 @@ declare module '@podman-desktop/api' {
     Type?: string;
   }
 
+  /**
+   * Authentication credentials, used when pushing an image to a registry
+   */
   interface ContainerAuthInfo {
     username: string;
     password: string;
     serveraddress: string;
     email?: string;
   }
+
   interface PullEvent {
     stream?: string;
     id?: string;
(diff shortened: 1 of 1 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -2350,30 +2350,113 @@ declare module '@podman-desktop/api' {
...
   export namespace containerEngine {
+    /**
+     * Returns the list of containers across all container engines.
+     *
+     * @return A promise resolving to an array of containers information
+     */
     export function listContainers(): Promise<ContainerInfo[]>;
+
+    /**
+     * Get the Inspect information of a specific container, similar to what a `podman inspect` command returns
```
benoitf: I'm not sure to use 'podman inspect' as it's more behaving like a 'docker REST API' and then 'docker inspect' result

benoitf: maybe just avoid to use a specific engine name

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2382,12 +2465,34 @@ declare module '@podman-desktop/api' {
...
+    /**
+     * Delete a container image from a local engine
+     *
+     * @param engineId the id of the engine managing the image
+     * @param id the id of the image on this engine
+     */
     export function deleteImage(engineId: string, id: string): Promise<void>;
+
+    /**
+     * Get Inspect information for a local image, similarly to the `podman image inspect` command
```
benoitf: get rid of 'podman'

axel7083 (approved):
LGTM

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2350,30 +2350,113 @@ declare module '@podman-desktop/api' {
...
+     */
     export function buildImage(
       // build context directory
       context: string,
       eventCollect: (eventName: 'stream' | 'error' | 'finish', data: string) => void,
       options?: BuildImageOptions,
     );
+
+    /**
+     * Save a container image on disk, similarly to the `podman image save` command
```
benoitf: get rid of 'podman'

benoitf (commented):
it may require some `{@link NameOfTheClass}`

many methods are self-explanatory so I would guess addition of typedoc would be to provide examples or explain stuff

for example an engineID, how to grab it or to know it

for id, that it can be the name of a container (or image or etc) or the sha or a partial sha, etc

feloy:
> it may require some `{@link NameOfTheClass}`
> 
> many methods are self-explanatory so I would guess addition of typedoc would be to provide examples or explain stuff
> 
> for example an engineID, how to grab it or to know it
> 

I wanted to document this, but I'm not sure yet what is the way to have it. Only from listImages(), listContainers(), or is there a way to list the engines (I cannot find any)?


> for id, that it can be the name of a container (or image or etc) or the sha or a partial sha, etc

benoitf:
> I wanted to document this, but I'm not sure yet what is the way to have it. Only from listImages(), listContainers(), or is there a way to list the engines (I cannot find any)?

yes it's from the list operations for now

probably let say something like for deleteContainer (could be fully reworked, just a sample)

`When getting all containers using the {@link engine.listContainers} function to enumerate all containers, you can retrieve the IDs and engine IDs of each container. Then you can use them as parameters to deleteContainer`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2276,89 +2359,154 @@ declare module '@podman-desktop/api' {
...
+    /**
+     * Set the os/arch of the built image (and its base image, when using one) to the provided value instead of using the current operating system and architecture of the host
+     */
     platform?: string;
-    // Set the provider to use, if not we will try select the first one available (sorted in favor of Podman).
+
+    /**
+     * Set the provider to use, if not we will try select the first one available (sorted in favor of Podman)
+     */
     provider?: ProviderContainerConnectionInfo | containerDesktopAPI.ContainerProviderConnection;
```
feloy, line 2380: These 2 types are not declared on the file

benoitf: I think only one should be there

benoitf (approved):
Thanks now the typedoc is providing real value 👏 and not only describing the name of the functions/parameters

packages/extension-api/src/extension-api.d.ts
feloy: `ContainerProviderConnection` is defined, but `containerDesktopAPI` makes it undiscoverable

Do you prefer I make the change as part of this PR? It should be this, right?

```
```

benoitf: I think it needs a separate PR
there is already one on this topic

benoitf: https://github.com/containers/podman-desktop/pull/5880

PR #5901, opened 2024-02-07
Kube apply button

### LINKED ISSUE #5661: Ability to apply a Kube YAML file to the cluster

deboer-tim:
### What does this PR do?

Based on PR #5849, only /renderer files (latest commit) is changed here.

Creates an 'Apply YAML' button that prompts for a file, applies it to
the current context, and reports how many resources have been created.

I've added the button to the Deployments, Services, and Ingress/Routes
pages, as these are the most useful pages for a Kubernetes user. I
didn't add it to the Pods page b/c this is less likely used for a
Kube user and it might create confusion with the 'Play Kubernetes YAML'
button.

Created a SolidKubeIcon identical to how we created a SolidPodIcon
for the 'create pod from containers' action.

### Screenshot / video of UI

Screenshot 2024-02-14 at 4 16 49 PM

Fixes #5661.

### How to test this PR?

Unit tests included. Pick your favourite yaml and apply!

### DIFF
--- a/packages/main/src/plugin/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes-client.spec.ts
@@ -1218,22 +1218,27 @@ test('Expect ingress refreshInformer should stop and start the informer again', 
 
 test('Expect apply with invalid file should error', async () => {
   const client = createTestClient('default');
+  let expectedError: unknown;
   try {
     await client.applyResourcesFromFile('default', 'missing-file.yaml');
-    throw Error('should never get here');
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  } catch (err: any) {
-    expect(err).to.be.a('Error');
-    expect(err.message).equal('File missing-file.yaml does not exist');
+  } catch (err: unknown) {
+    expectedError = err;
   }
+  expect(expectedError).to.be.a('Error');
+  expect((expectedError as Error).message).equal('File missing-file.yaml does not exist');
 });
 
-test('Expect apply with empty yaml should return no objects', async () => {
+test('Expect apply with empty yaml should throw error', async () => {
   const client = createTestClient('default');
   vi.spyOn(client, 'loadManifestsFromFile').mockReturnValue(Promise.resolve([]));
-
-  const objects = await client.applyResourcesFromFile('default', 'missing-file.yaml');
-  expect(objects).toEqual([]);
+  let expectedError: unknown;
+  try {
+    await client.applyResourcesFromFile('default', 'missing-file.yaml');
+  } catch (err: unknown) {
+    expectedError = err;
+  }
+  expect(expectedError).to.be.a('Error');
+  expect((expectedError as Error).message).equal('No valid Kubernetes resources found in file');
 });
 
 test('Expect apply should create if object does not exist', async () => {
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -1048,6 +1048,9 @@ export class KubernetesClient {
    */
   async createResourcesFromFile(context: string, filePath: string, namespace?: string): Promise<void> {
     const manifests = await this.loadManifestsFromFile(filePath);
+    if (manifests.filter(s => s?.kind).length === 0) {
+      throw new Error('No valid Kubernetes resources found in file');
+    }
     await this.syncResources(context, manifests, 'create', namespace);
   }
 
@@ -1074,6 +1077,9 @@ export class KubernetesClient {
    */
   async applyResourcesFromFile(context: string, filePath: string, namespace?: string): Promise<KubernetesObject[]> {
     const manifests = await this.loadManifestsFromFile(filePath);
+    if (manifests.filter(s => s?.kind).length === 0) {
+      throw new Error('No valid Kubernetes resources found in file');
+    }
     return this.syncResources(context, manifests, 'apply', namespace);
   }
 
--- a/packages/renderer/src/lib/kube/KubeApplyYAMLButton.svelte
+++ b/packages/renderer/src/lib/kube/KubeApplyYAMLButton.svelte
@@ -1,0 +1,59 @@
+<script lang="ts">
+import Button from '../ui/Button.svelte';
+import type { KubernetesObject } from '@kubernetes/client-node';
+import SolidKubeIcon from '../images/SolidKubeIcon.svelte';
+
+let inProgress = false;
+
+async function kubeApply(): Promise<void> {
+  let contextName = await window.kubernetesGetCurrentContextName();
+  if (!contextName) {
+    return;
+  }
+
+  const result = await window.openFileDialog('Select a .yaml file to apply', {
+    name: 'YAML files',
+    extensions: ['yaml', 'yml'],
+  });
+  if (result.canceled || result.filePaths.length !== 1) {
+    return;
+  }
+
+  let file = result.filePaths[0];
+  if (!file) {
+    return;
+  }
+
+  inProgress = true;
+  try {
+    const namespace = await window.kubernetesGetCurrentNamespace();
+    let objects: KubernetesObject[] = await window.kubernetesApplyResourcesFromFile(contextName, file, namespace);
+    if (objects.length === 0) {
+      await window.showMessageBox({
+        title: 'Kubernetes',
+        type: 'warning',
+        message: `No resource(s) were applied`,
+        buttons: ['OK'],
+      });
+    } else {
+      await window.showMessageBox({
+        title: 'Kubernetes',
+        type: 'info',
+        message: `Successfully applied ${objects.length} resource(s)`,
+        buttons: ['OK'],
+      });
+    }
+  } catch (error) {
+    await window.showMessageBox({
+      title: 'Kubernetes',
+      type: 'error',
+      message: 'Could not apply Kubernetes YAML: ' + error,
+      buttons: ['OK'],
+    });
+  }
+  inProgress = false;
+}
+</script>
+
+<Button on:click="{() => kubeApply()}" title="Apply YAML" icon="{SolidKubeIcon}" inProgress="{inProgress}"
+  >Apply YAML</Button>
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -2010,8 +2010,8 @@ export class PluginSystem {
 
     this.ipcHandle(
       'kubernetes-client:applyResourcesFromFile',
-      async (_listener, context: string, file: string): Promise<KubernetesObject[]> => {
-        return kubernetesClient.applyResourcesFromFile(context, file);
+      async (_listener, context: string, file: string, namespace?: string): Promise<KubernetesObject[]> => {
+        return kubernetesClient.applyResourcesFromFile(context, file, namespace);
       },
     );
 
--- a/packages/preload/src/index.ts
+++ b/packages/preload/src/index.ts
@@ -1829,8 +1829,8 @@ export function initExposure(): void {
 
   contextBridge.exposeInMainWorld(
     'kubernetesApplyResourcesFromFile',
-    async (context: string, file: string): Promise<KubernetesObject[]> => {
-      return ipcInvoke('kubernetes-client:applyResourcesFromFile', context, file);
+    async (context: string, file: string, namespace?: string): Promise<KubernetesObject[]> => {
+      return ipcInvoke('kubernetes-client:applyResourcesFromFile', context, file, namespace);
     },
   );
 
--- a/packages/renderer/src/lib/deployments/DeploymentsList.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentsList.svelte
@@ -21,6 +21,7 @@ import FilteredEmptyScreen from '../ui/FilteredEmptyScreen.svelte';
 import SimpleColumn from '../table/SimpleColumn.svelte';
 import DurationColumn from '../table/DurationColumn.svelte';
 import KubernetesCurrentContextConnectionBadge from '/@/lib/ui/KubernetesCurrentContextConnectionBadge.svelte';
+import KubeApplyYamlButton from '../kube/KubeApplyYAMLButton.svelte';
 
 export let searchTerm = '';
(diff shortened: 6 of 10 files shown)


deboer-tim:
Rebased on latest Kube apply branch, updated to use new Kube icon (same as nav), and added tests. Ready for review.

packages/renderer/src/lib/kube/KubeApplyYAMLButton.svelte
```diff
@@ -0,0 +1,49 @@
...
+    return;
+  }
+
+  inProgress = true;
+  try {
+    let objects: KubernetesObject[] = await window.kubernetesApplyResourcesFromFile(contextName, file);
+    await window.showMessageBox({
+      title: 'Kubernetes',
+      type: 'info',
+      message: `${objects.length} resource(s) applied`,
```
cdrage: Could we say something like: 'Successfully applied {objects.length} resource(s)'?

Just to indicate it's successful :)

packages/renderer/src/lib/kube/KubeApplyYAMLButton.svelte
```diff
@@ -0,0 +1,49 @@
...
+      buttons: ['OK'],
+    });
+  } catch (error) {
+    await window.showMessageBox({
+      title: 'Kubernetes',
+      type: 'error',
+      message: 'Could not apply Kubernetes yaml: ' + error,
```
cdrage:
```suggestion
      message: 'Could not apply Kubernetes YAML: ' + error,
```

cdrage (changes requested):
Success is working, but for some reason not erroring out anything unvalidated / working correctly, and I'm unsure if this is something we didn't check more thoroughly in the other PR:

Working:
https://github.com/containers/podman-desktop/assets/6422176/e22cbdbf-c998-47a0-accf-bae9c396b4c6


When using a random YAML (I just used `helloworld123) in a `test.yaml`):


https://github.com/containers/podman-desktop/assets/6422176/473a9954-e121-4ccc-864b-7fd6eafd2330

packages/renderer/src/lib/kube/KubeApplyYAMLButton.svelte
deboer-tim: Fixed.

packages/renderer/src/lib/kube/KubeApplyYAMLButton.svelte
deboer-tim: Fixed.

deboer-tim:
Other comments fixed, and rebased on main.

> Success is working, but for some reason not erroring out anything unvalidated / working correctly, and I'm unsure if this is something we didn't check more thoroughly in the other PR:
> 
> When using a random YAML (I just used `helloworld123) in a `test.yaml`):

I debugged what's happening here: loadManifestsFromFile() loads and parses the file as a JSON object `helloworld123)`, then syncResources() filters it out (b/c it's not a valid manifest with a kind) and doesn't attempt to apply, and returns 0 resources pushed. Even if had tried it would have failed and returned 0 too.

This is kind of an odd case because the caller doesn't know how many valid resource(s) are supposed to be in the file to begin with, i.e. the file could have 5 resources and only 4 were valid/successfully applied, or 3 and 0. I'm thinking there should be two fixes to handle different cases:

1. applyResourcesFromFile() could throw an error if there are no valid resources in the file.
2. The button should special-case 0 and give a warning message that nothing was applied.

I'll do this in my next commit.

deboer-tim:
New commit added, as above. Please verify.

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1119,6 +1125,7 @@ export class KubernetesClient {
       ctx.currentContext = context;
 
       const validSpecs = manifests.filter(s => s?.kind) as KubernetesObjectWithKind[];
+      console.log(validSpecs);
```
cdrage:
```suggestion
```

Just need to remove this

deboer-tim:
Looks like I didn't actually push last week, tests fixed now. @cdrage, please re-review/see if the prior comment works for you.

cdrage:
https://github.com/containers/podman-desktop/assets/6422176/1af30bf3-c834-4c45-b497-8f85a75fcf86

Awesome work! LGTM! Tested via a faulty YAML as well as a valid one.

cdrage (approved):

lstocchi (commented):
I see this error everytime i try to apply a yaml file (no matter if it's valid or not). 
I'm connected to the sandbox cluster.

image

cdrage:
I've opened up https://github.com/containers/podman-desktop/pull/6086

I can confirm that this PR patched on this one, will result on correct error outputs.

lstocchi (commented):
The patch made by @cdrage allowed me to read the error.
The problem is that it tries to push the resource to the default namespace even if i'm connected to lstocchi-dev.

image

image

cdrage:
> The patch made by @cdrage allowed me to read the error.
> 
> The problem is that it tries to push the resource to the default namespace even if i'm connected to lstocchi-dev.
> 
> 
> 
> image
> 
> 
> 
> image
> 
> 

Does your test YAML has namespace default in it?

If so, that's a normal error. 

If not, then yes it should be applying it to the current selected namespace.

deboer-tim:
I was going to say the same thing. Maybe you could share the yaml (offline)? If metadata/namespace exists and points to 'default' then this is expected behaviour, otherwise the bug is not directly in this PR but the underlying API (#5849) and I will fix.

lstocchi:
This is the yaml
```
kind: Ingress
metadata:
  name: i-17
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
spec:
  ingressClassName: nginx-example
  rules:
  - http:
      paths:
      - path: /testpath1
        pathType: Prefix
        backend:
          resource:
            name: test
            kind: 'sink'
```

deboer-tim:
My bad - underlying function supports applying namespace now but I never exposed or hooked it up to the button. Rebased on top of latest to get Charlie's error fix, exposed and added namespace, and updated tests. Working well for me now against sandbox with YAML above.

packages/renderer/src/lib/kube/KubeApplyYAMLButton.svelte
```diff
@@ -0,0 +1,59 @@
...
+
+  let file = result.filePaths[0];
+  if (!file) {
+    return;
+  }
+
+  inProgress = true;
+  try {
+    const namespace = await window.kubernetesGetCurrentNamespace();
+    let objects: KubernetesObject[] = await window.kubernetesApplyResourcesFromFile(contextName, file, namespace);
```
cdrage, line 30: 🎉

cdrage (changes requested):
Tested it 3 ways with the following YAML:

```yaml
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: helloworld-pod
  name: helloworld-pod
spec:
  containers:
  - image: docker.io/cdrage/helloworld:latest
    name: helloworld
    ports:
    - containerPort: 8080
      hostPort: 8080
 ```
 
 
 1. 🎉  `namespace: kube-system` added, even though kubernetes context is `default`. Deploys to `kube-system` which is intentional and awesome.
 2. 🎉 `namespace: default` added, while being on default namespace.
 3. 🛑 No namespace: No Pod is deployed despite it saying "Resources applied". If you apply the same yaml via `kubectl apply` it works fine.

File: packages/main/src/plugin/kubernetes-client.spec.ts [hunk 5]
```diff
@@ -1221,19 +1221,23 @@ test('Expect apply with invalid file should error', async () => {
...
   const client = createTestClient('default');
   vi.spyOn(client, 'loadManifestsFromFile').mockReturnValue(Promise.resolve([]));
 
-  const objects = await client.applyResourcesFromFile('default', 'missing-file.yaml');
-  expect(objects).toEqual([]);
+  try {
+    await client.applyResourcesFromFile('default', 'missing-file.yaml');
+    throw Error('should never get here');
+  } catch (err: unknown) {
+    expect(err).to.be.a('Error');
```
dgolovin: Same problem.

packages/main/src/plugin/kubernetes-client.spec.ts
```diff
@@ -1221,19 +1221,23 @@ test('Expect apply with invalid file should error', async () => {
   try {
     await client.applyResourcesFromFile('default', 'missing-file.yaml');
     throw Error('should never get here');
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  } catch (err: any) {
+  } catch (err: unknown) {
     expect(err).to.be.a('Error');
```
dgolovin:
```suggestion
```

packages/main/src/plugin/kubernetes-client.spec.ts
```diff
@@ -1221,19 +1221,23 @@ test('Expect apply with invalid file should error', async () => {
...
-    expect(err.message).equal('File missing-file.yaml does not exist');
+    expect((err as Error).message).equal('File missing-file.yaml does not exist');
```
dgolovin:
```suggestion
expectedError = err;
```

packages/main/src/plugin/kubernetes-client.spec.ts
```diff
@@ -1221,19 +1221,23 @@ test('Expect apply with invalid file should error', async () => {
   try {
     await client.applyResourcesFromFile('default', 'missing-file.yaml');
     throw Error('should never get here');
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  } catch (err: any) {
+  } catch (err: unknown) {
```
dgolovin, line 1224: This test does not fail if apply does not throw an error, but test name says it 'should'.

packages/main/src/plugin/kubernetes-client.spec.ts
```diff
@@ -1221,19 +1221,23 @@ test('Expect apply with invalid file should error', async () => {
   try {
```
dgolovin, line 1222:
```suggestion
  let expectedError: unknown;
  try {
```

packages/main/src/plugin/kubernetes-client.spec.ts
```diff
@@ -1221,19 +1221,23 @@ test('Expect apply with invalid file should error', async () => {
...
     await client.applyResourcesFromFile('default', 'missing-file.yaml');
     throw Error('should never get here');
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  } catch (err: any) {
+  } catch (err: unknown) {
     expect(err).to.be.a('Error');
-    expect(err.message).equal('File missing-file.yaml does not exist');
+    expect((err as Error).message).equal('File missing-file.yaml does not exist');
   }
 });
```
dgolovin, line 1229:
```suggestion
  expect(expectedError).to.be.a('Error');
  expect((err as Error).message).equal('File missing-file.yaml does not exist');
});
```

packages/renderer/src/lib/kube/KubeApplyYAMLButton.svelte
```diff
@@ -0,0 +1,59 @@
...
+        type: 'info',
+        message: `Successfully applied ${objects.length} resource(s)`,
+        buttons: ['OK'],
+      });
+    }
+  } catch (error) {
+    await window.showMessageBox({
+      title: 'Kubernetes',
+      type: 'error',
+      message: 'Could not apply Kubernetes YAML: ' + error,
```
dgolovin, line 50: Better to use coercion here
```suggestion
      message: 'Could not apply Kubernetes YAML: ' + String(error),
```

dgolovin (changes requested):

cdrage:
> Tested it 3 ways with the following YAML:
> 
> ```yaml
> apiVersion: v1
> kind: Pod
> metadata:
>   labels:
>     app: helloworld-pod
>   name: helloworld-pod
> spec:
>   containers:
>   - image: docker.io/cdrage/helloworld:latest
>     name: helloworld
>     ports:
>     - containerPort: 8080
>       hostPort: 8080
> ```
> 
> 1. 🎉  `namespace: kube-system` added, even though kubernetes context is `default`. Deploys to `kube-system` which is intentional and awesome.
> 2. 🎉 `namespace: default` added, while being on default namespace.
> 3. 🛑 No namespace: No Pod is deployed despite it saying "Resources applied". If you apply the same yaml via `kubectl apply` it works fine.

Scratch that...

I found the problem and it was a lingering "Terminating" helloworld-pod that was not allowing the YAML to complete, and that's why it was giving the false "0 Resources Applied" being shown.

My "hello world" container takes too long to shut down.

Intentional feature of applying it looks!

cdrage (approved):
LGTM! Just updating the tests of @dgolovin 's comments and we should be good to go 🚀 🚀 🚀

deboer-tim:
@dgolovin  tests updated, please review

jeffmaury:
The Pod page still have Play Kubernetes YAML and not Apply YAML

deboer-tim:
> The Pod page still have Play Kubernetes YAML and not Apply YAML

Play Kube is fundamentally different from Apply: it works against Podman (hence the name), or can target Kubernetes but does a create (i.e. won't change existing resources). Apply matches 'kubectl apply' which does 'create or patch', and only works on Kubernetes.

Personally I think we should reduce Play Kube to only work on Podman and use Apply for Kube, but this is not the only area where Pods fall in both camps, having two buttons on the Pods page would be even more confusing, and this gets into some political questions. My goal here was to add Apply Kube to pages like Deployments and Services which are clearly Kube pages and it is expected, and defer/avoid disrupting/wait for feedback on the Pods page.

deboer-tim:
Rebased to fix conflicts.

deboer-tim:
@dgolovin another review please?

dgolovin (approved):
LGTM

cdrage (approved):
After your changes, I've tested this again against a "error out YAML" as well as a functional yaml with multiple resources.

Everything LGTM! Error-wise too.

Hitting the merge button 🎉

PR #5904, opened 2024-02-08
feat: input component

deboer-tim:
### What does this PR do?

Creates a new Input component with the same basic support as an HTML input, but with our styling. Supports four types so far:
- normal text
- search (adds a search icon on the left)
- clear (adds an X on the right, which clears the field unless the parent event handler overrides it)
- password (with a show/hide button on the right ( unless the parent event handler disables it)

I don't want to give any impression that this component is fully done or supports all cases, my intention is to get a good base component that does enough to support the initial use, then expand it as necessary to support different uses or additional style requirements.

In order to give something to test, the first few uses of Input are included:
- Main nav page search bar
- Settings > Preferences search bar
- Settings > Preferences string and file items

### Screenshot / video of UI

Before:
Screenshot 2024-02-07 at 10 12 59 AM

After:
https://github.com/containers/podman-desktop/assets/19958075/6578e8b4-a58f-4c86-9011-861f2704e4e3

https://github.com/containers/podman-desktop/assets/19958075/11fec031-6859-44a9-a4bf-79fa49e19172

First step of #3234.

### How to test this PR?

Go to places noted and use the inputs as normal, look for odd style or behaviour.

### DIFF
--- a/packages/renderer/src/lib/ui/SearchInput.svelte
+++ b/packages/renderer/src/lib/ui/SearchInput.svelte
@@ -1,0 +1,32 @@
+<script lang="ts">
+import Input from './Input.svelte';
+
+export let title: string;
+export let searchTerm: string = '';
+</script>
+
+<Input
+  class="{$$props.class || ''}"
+  id="search-{title}"
+  name="search-{title}"
+  placeholder="Search {title}..."
+  bind:value="{searchTerm}"
+  aria-label="search"
+  clearable
+  on:input>
+  <svelte:fragment slot="left">
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      class="w-4 h-4 mx-1 text-gray-700"
+      fill="none"
+      viewBox="0 0 24 24"
+      stroke="currentColor"
+      role="img">
+      <path
+        stroke-linecap="round"
+        stroke-linejoin="round"
+        stroke-width="2"
+        d="M21 21l-6-6m2-5a7 7 0 11-14 0 7 7 0 0114 0z"></path>
+    </svg>
+  </svelte:fragment>
+</Input>
--- a/packages/renderer/src/lib/ui/SearchInput.svelte
+++ b/packages/renderer/src/lib/ui/SearchInput.svelte
@@ -11,7 +11,7 @@ export let searchTerm: string = '';
   name="search-{title}"
   placeholder="Search {title}..."
   bind:value="{searchTerm}"
-  aria-label="search"
+  aria-label="search {title}"
   clearable
   on:input>
   <svelte:fragment slot="left">
--- a/packages/renderer/src/lib/container/ContainerList.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerList.spec.ts
@@ -141,7 +141,7 @@ test('Expect is:running / is:stopped is added to the filter field', async () => 
   }
   await waitRender({});
 
-  const searchField = screen.getByPlaceholderText('Search containers....');
+  const searchField = screen.getByPlaceholderText('Search containers...');
   expect(searchField).toBeInTheDocument();
   expect(searchField).not.toHaveDisplayValue(/is:running/);
   expect(searchField).not.toHaveDisplayValue(/is:stopped/);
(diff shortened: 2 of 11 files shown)


benoitf (commented):
hello, I won't discuss the style as it can be amended easily

I'm wondering if we need different type for input component like search or clear or if we need composition.

Let me explain.

For example, to me, search is more a custom component than an input type
Like there is a generic input component where you specify an icon on the left or on the right, etc and Search is reusing that generic component with 'search' not being a given type

We could say that search input also have a way to clear the section as well but then it's either a 'search' or a 'clear' type so you're stuck and maybe sometimes you want one and not the other so it would end up with `search` `searchWithClear` etc types

if I look at the usage I see for example
```html
 <Input
      slot="header"
      on:input="{e => updateSearchValue(e)}"
      class="mt-4"
      name="search"
      type="search"
      placeholder="Search preferences"
      id="input-search"
      aria-label="input-search" />
```

so it means I'll have to duplicate all these lines for each search field

I'm wondering then if I should not have a `<SearchInput name="preferences">` using the generic <Input component where SearchInput will provide a Left icon with a magnifying glass that will add the placeHolder for me, etc

because I could omit the placeHolder in ImageList page, forget the aria-label somewhere else, etc but I would expect that when using the Search component, it's always added

We already have a <Checkbox> component defined (and not for example a <Input type="checkbox" being defined) so maybe we should go with dedicated component

`<PasswordInput />` `<TextInput />`, `<SearchInput>`, `<AnotherKindOfInput>`

or cut Input like for `Checkbox` that is not a `CheckboxInput`

deboer-tim:
I'm not opposed to that, but I think the main point is that 'clear' is not a 'type' since we may want it on search inputs (or maybe even password or other?) If we had separate components, it would be an attribute on the parent Input (likely TextInput as parent).

We could make it a separate attribute or default things like aria-label now though, so I'm not sure separate components buys us much (it becomes the difference between using 'Input type=text or search' vs 'TextInput or SearchInput'), but if you feel strongly I can change.

benoitf:
so from 'componentization' perspective, yes the input component should have lot of customization attributes (as you said,  yes we could move clear to be an attribute instead of a type)

but then I think that we could go one step further to reduce the copy/paste over all the code

if each time we do a search input for example we need to provide icon={faMagnifyGlass}, placeholder, aria label, etc, I think we're like in the middle. We have the colours being set but we still copy a lot of "common code" everywhere

so I would do "ready-to-use" components where basically you almost do nothing when consuming it. Just the 'logic'

like:

```
<Search title="images" />
```

vs

```
<Input leftIcon={faMagnifyGlass} placeHolder="Search images" aria-label="search icons" .... />
```

It's almost what was done for `<Checkbox>` where you just use the component without customization, just use the bindings

deboer-tim:
Ok, I think the last example is compelling and tried it in the latest commit. I left Input as the basic/fully-featured component and created SearchInput and PasswordInput that add icons/actions and expose fewer options b/c they simplify/default them.

I had to make some choices of what to expose from Search/PasswordInput and tried to keep them fairly minimal. SearchInput has the X icon now and I updated the ContainerList test b/c I went with a more standard '...'. Let me know what you think.

packages/renderer/src/lib/ui/SearchInput.svelte
```diff
@@ -0,0 +1,32 @@
...
+export let searchTerm: string = '';
+</script>
+
+<Input
+  class="{$$props.class || ''}"
+  id="search-{title}"
+  name="search-{title}"
+  placeholder="Search {title}..."
+  bind:value="{searchTerm}"
+  aria-label="search"
```
benoitf: I think here we're missing title as well no ?

benoitf (commented):
thanks, it looks like in the current usage
We're reducing more lines of copy/pasting stuff

I think @mairin wanted to review the style

benoitf (approved):
forgot to approve yesterday

packages/renderer/src/lib/ui/SearchInput.svelte
deboer-tim: Updated aria-label (for search and password).

deboer-tim:
Talked to @mairin and we likely want to make some tweaks, but do not want to hold up getting the component in, and starting to use it in other places. So, I'm going to merge and opened #5932 (targeted to this sprint) to follow up.

PR #5912, opened 2024-02-08
feat: add a KubernetesContextState component

feloy:

### What does this PR do?

This PR adds a KubernetesContextState to the backend. 

This component watches pods and deployments (using informers) on all the contexts defined in the user's kubeconfig file.

Every time a change happens in any of the context (number of pods/deployment, context is reachable or not), the new context is sent with a message `kubernetes-contexts-state-update`

Part of #4562 

The PR #5917, based on this one, can be used to test with the information displayed on the frontend


packages/main/src/plugin/testutils/fake-informer.ts
```diff
@@ -0,0 +1,64 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ErrorCallback, KubernetesObject, ObjectCallback } from '@kubernetes/client-node';
+
+export class FakeInformer {
```
benoitf: hello, is that this class will be used in other places than kubernetes-context-state ?

feloy: For this feature no, but it is very generic and could be useful for any other code using informers

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,223 @@
...
+    });
+
+    const ns = context.namespace ?? 'default';
+    return {
+      podInformer: this.createPodInformer(kc, ns, context),
+      deploymentInformer: this.createDeploymentInformer(kc, ns, context),
+    };
+  }
+
+  private createPodInformer(kc: KubeConfig, ns: string, context: KubeContext): Informer<V1Pod> & ObjectCache<V1Pod> {
```
deboer-tim, line 186: There is a lot of duplication between createPodInformer and createDeploymentInformer. Could you refactor out a common function sort of like kubernetes-client.ts makeKubernetesInformer?

deboer-tim (commented):
Sorry for this long review. I'm not sure why it wasn't fully dawning on me before (sorry Philip, I think you've been telling me), but after this PR we will have:

1. pod data pulled using list + watcher.
2. deployments/services/ingress/routes list + informers for current context, with state managed in the store via add/remove events.
3. pod/deployment counts pulled for all contexts by these informers, with future store.

To me #1 is unchanged technical debt; we didn't want to risk a failed release while trying informers/Kubernetes UI, but we've been using them long enough now to know they're better in all respects.

I like that #2 gives us add/remove events in the renderer, but we may never need it and could replicate. Our long-term design mockups had the ability to see resources in multiple contexts at once, and #3 is the only multi-context pattern.

So accepting this PR will create some overlapping function for the current context and technical debt, but I also think it gives us a better way to get to multiple contexts without any disruption, e.g. a path something like:

- Merge this PR (which would unblock Settings > Kubernetes UI improvements, and ability to put current connection status on Deployments/Services/etc pages or statusbar)
- Update ContextsState to keep arrays for pods & deployments instead of counts.
- Add services/ingress/routes to ContextsState.
- Change kube resource stores to use ContextsState.
- Delete 'old' informers.
- Update pods to use informer store, delete old pod watcher.

Thoughts? My only real concern with the existing code was to reduce duplicate code in createXInformer functions.

packages/main/src/plugin/kubernetes-context-state.ts
dgolovin: There are many smaller duplications like:

```
      const previous = this.contextsState.get(context.name);
      if (previous) {
        /*
           do something with previous context state
        */
      }
      this.dispatchContextsState();
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,230 @@
...
+    ns: string,
+    context: KubeContext,
+  ): Informer<V1Deployment> & ObjectCache<V1Deployment> {
+    const k8sApi = kc.makeApiClient(AppsV1Api);
+    const listFn = () => k8sApi.listNamespacedDeployment(ns);
+    const path = `/apis/apps/v1/namespaces/${ns}/deployments`;
+    return this.createInformer<V1Deployment>(kc, context, path, listFn, 'deploymentsCount', { checkReachable: false });
+  }
+
+  private createInformer<T extends KubernetesObject>(
```
feloy, line 215: This method may be moved to kubernetes-client.ts sooner or later

feloy:
> Sorry for this long review. I'm not sure why it wasn't fully dawning on me before (sorry Philip, I think you've been telling me), but after this PR we will have:
> 
> 1. pod data pulled using list + watcher.
> 2. deployments/services/ingress/routes list + informers for current context, with state managed in the store via add/remove events.
> 3. pod/deployment counts pulled for all contexts by these informers, with future store.
> 
> To me #1 is unchanged technical debt; we didn't want to risk a failed release while trying informers/Kubernetes UI, but we've been using them long enough now to know they're better in all respects.
> 
> I like that #2 gives us add/remove events in the renderer, but we may never need it and could replicate. Our long-term design mockups had the ability to see resources in multiple contexts at once, and #3 is the only multi-context pattern.
> 
> So accepting this PR will create some overlapping function for the current context and technical debt, but I also think it gives us a better way to get to multiple contexts without any disruption, e.g. a path something like:
> 
> * Merge this PR (which would unblock Settings > Kubernetes UI improvements, and ability to put current connection status on Deployments/Services/etc pages or statusbar)
> * Update ContextsState to keep arrays for pods & deployments instead of counts.
> * Add services/ingress/routes to ContextsState.
> * Change kube resource stores to use ContextsState.
> * Delete 'old' informers.
> * Update pods to use informer store, delete old pod watcher.
> 
> Thoughts? My only real concern with the existing code was to reduce duplicate code in createXInformer functions.

I have refactored the createXInformer with a single generic createInformer. Handlers are passed to this method, to make it completely agnostic of its context, to make it (hopefully) reusable in other parts of the code. The function may be moved to kubernetes-client.ts sooner or later, when it is reused elsewhere

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,227 @@
...
+      options.onReachable?.(true);
+    });
+
+    informer.on('delete', (obj: T) => {
+      options.onDelete?.(obj);
+      options.onReachable?.(true);
+    });
+    informer.on('error', (err: unknown) => {
+      if (err !== undefined) {
+        console.error(`pod informer error for context ${context.name}: `, String(err));
```
deboer-tim: 'pod' needs to be replaced here, maybe use path in the message instead

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,227 @@
...
+      // Restart informer after 5sec
+      setTimeout(() => {
+        this.restartInformer<T>(informer, context, options);
+      }, 5000);
+    });
+
+    if (options.onReachable) {
+      informer.on('connect', (err: unknown) => {
+        if (err !== undefined) {
+          console.error(`pod informer connect error for context ${context.name}: `, String(err));
```
deboer-tim: Ditto here.

deboer-tim (approved):
Thanks for the changes, just a couple minor issues noted.

feloy:
> Thanks for the changes, just a couple minor issues noted.

Fixed, thanks

deboer-tim (approved):
Adding back my approval, changes were addressed.

feloy:
In my first PR (https://github.com/containers/podman-desktop/pull/5329), I had added a Preference to enable/disable running informers on all clusters of the kubeconfig files, and disabled by default, because I was afraid that user with many old/unexisting clusters on the kubeconfig would retry connecting to all of them infinitely.

I removed this preference in this PR, because I understand that we would like to reuse the information, at least fot the current context.

But I'm still wondering if we could have this preference for the non-current context, to avoid trying to connect to these non-current contexts.

@deboer-tim @benoitf WDYT?

feloy:
> In my first PR (#5329), I had added a Preference to enable/disable running informers on all clusters of the kubeconfig files, and disabled by default, because I was afraid that user with many old/unexisting clusters on the kubeconfig would retry connecting to all of them infinitely.
> 
> I removed this preference in this PR, because I understand that we would like to reuse the information, at least fot the current context.
> 
> But I'm still wondering if we could have this preference for the non-current context, to avoid trying to connect to these non-current contexts.
> 
> @deboer-tim @benoitf WDYT?

As discussed in daily meeting, I'll add an exponential backoff with jitter on a follow up PR

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,227 @@
...
+  podInformer?: Informer<V1Pod> & ObjectCache<V1Pod>;
+  deploymentInformer?: Informer<V1Deployment> & ObjectCache<V1Deployment>;
+}
+
+// ContextState stores information for the user about a kube context: is the cluster reachable, the number
+// of instances of different resources
+export interface ContextState {
+  reachable: boolean;
+  podsCount: number;
+  deploymentsCount: number;
```
benoitf, line 42: it's just questions as it's an internal object we can changes later, when we want

should it be podsCount or `{pods: {count : number}}` if we add other stuff related to pods later

about reachable, I think we should add the last time it was reachable as well

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,227 @@
...
+          this.contextsInternalState.set(context.name, informers);
+        }
+      }
+    }
+    // Delete informers for removed contexts
+    let removed = false;
+    for (const [name, state] of this.contextsInternalState) {
+      if (!this.kubeConfig.contexts.find(c => c.name === name)) {
+        await state.podInformer?.stop();
+        await state.deploymentInformer?.stop();
```
benoitf: should we have only a stop method that does that for us, instead of the need to call sub parts

in addition, It's not also mapping overall arch that is using disposable pattern

so when I register something I have a disposable object that will do stop, delete, etc
cleaning is just calling the dispose() method on the disposable object

so I would have only called dispose() on the state and it would have self stopped/removed the resource

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,227 @@
...
+  private dispatchContextsState() {
+    // Debounce: send only the latest value if several values are sent in a short period
+    if (this.timeoutId) {
+      clearTimeout(this.timeoutId);
+    }
+
+    this.timeoutId = setTimeout(() => {
+      this.apiSender.send(`kubernetes-contexts-state-update`, this.contextsState);
+    }, 100);
+  }
```
benoitf: the timeout looks like too low (or high depending on how we look)
I guess we could be like 1second

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,227 @@
...
+    });
+    informer.on('error', (err: unknown) => {
+      if (err !== undefined) {
+        console.error(`informer error on path ${path} for context ${context.name}: `, String(err));
+      }
+      options.onReachable?.(err === undefined);
+      // Restart informer after 5sec
+      setTimeout(() => {
+        this.restartInformer<T>(informer, context, options);
+      }, 5000);
```
benoitf: where is kept that setTimeout reference ?

if there is like several errors, would it add two times restartInformer ?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  private restartInformer<T>(
+    informer: Informer<KubernetesObject> & ObjectCache<KubernetesObject>,
+    context: KubeContext,
+    options: CreateInformerOptions<T>,
+  ) {
+    informer.start().catch((err: unknown) => {
+      if (err !== undefined) {
+        console.log('informer start error: ', String(err));
```
benoitf:
```suggestion
        console.error('informer start error: ', String(err));
```
(or warn but probably not console.log)

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,227 @@
...
+      }
+      options.onReachable?.(err === undefined);
+      // Restart informer after 5sec
+      setTimeout(() => {
+        this.restartInformer<T>(informer, context, options);
+      }, 5000);
```
benoitf: same question than before on the reference to the timeout reference

packages/main/src/plugin/testutils/fake-informer.ts
benoitf: I would keep the objet in the kubernetes-context-state.spec.ts and only if we need it elsewhere we would move it rather than creating a `testutils` directory with an hypothetical reuse of this object.

Also

packages/main/src/plugin/testutils/fake-informer.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+  off(
+    verb: 'change' | 'add' | 'update' | 'delete' | 'error' | 'connect',
+    cb: ErrorCallback | ObjectCallback<KubernetesObject>,
+  ) {
+    this.offCb.set(verb, cb);
+  }
+  get() {}
+  list() {}
+}
```
benoitf: I think I would have picked up mocking style here rather than declaring empty method

instead of `get()` --> `get: vi.fn()`

for off, usage of mockImplementation etc

packages/main/src/plugin/kubernetes-context-state.ts
deboer-tim: IMHO it should be changed to `pods: PodInfo[]` in a future PR, ditto for deployments. Easy enough to get the count from there and can replace current-context informer.

Good call on reachable. We could track timestamp on every interaction so that when disconnected we know the last time it was connected, and when connected we know the last time we had an update or when it connected. IMHO this could wait though.

packages/main/src/plugin/kubernetes-context-state.ts
feloy: it should be fixed with latest commit (refactor: use dispose pattern for state)

packages/main/src/plugin/kubernetes-context-state.ts
feloy: I have added a commit with timer related changes

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -0,0 +1,295 @@
...
+    }
+    return current;
+  }
+  reset() {
+    this.value = this.initial + this.getJitter();
+  }
+
+  private getJitter(): number {
+    // This is not security-related code
+    return Math.floor(this.jitter * Math.random()); // NOSONAR
```
benoitf: for disabling sonar rules you can do that from the UI of sonar source

but maybe you could just use
```ts
const uintArray = new Uint8Array(1)
const randomInt = crypto.getRandomValues(uintArray)[0]
const randomFloat = randomInt / Math.pow(2, 8)
```

feloy: > for disabling sonar rules you can do that from the UI of sonar source
> 

OK, I'll check agin the UI. Thanks

> but maybe you could just use
> 
> ```ts
> const uintArray = new Uint8Array(1)
> const randomInt = crypto.getRandomValues(uintArray)[0]
> const randomFloat = randomInt / Math.pow(2, 8)
> ```

As it is really not crypto code, I would prefer to keep it simple

lstocchi (commented):
Not sure if i'm doing something wrong. I merged this PR to https://github.com/containers/podman-desktop/pull/5917. 
I start desktop and log to the sandbox cluster.
Now if i go to the context page i see unreachable to all contexts, but i was expecting to be connected to the sandbox one. To verify i'm connected i also go to the pods page and i see the pod i have there. Then i switch to minikube which is turned off and i don't see any pod as expected.
Am i doing something wrong?
pod_context

lstocchi (approved):
Tested by using the new rebased #5917 . Works fine now. The pods/deploy count gets updated correctly as well. LGTM

image

PR #5915, opened 2024-02-08
chore: use entire height when creating container

cdrage:
chore: use entire height when creating container

### What does this PR do?

Uses the entire height instead of limiting it.

Small change, so only visual / no tests.

### Screenshot / video of UI

Before:
Screenshot 2024-02-08 at 4 28 33 PM

After:
Screenshot 2024-02-08 at 4 30 33 PM

N/A

### How to test this PR?

Go to create a container, you should see the height be the entire
window.

### DIFF
--- a/packages/renderer/src/lib/image/RunImage.svelte
+++ b/packages/renderer/src/lib/image/RunImage.svelte
@@ -654,7 +654,7 @@ async function assertAllPortAreValid(): Promise<void> {
           </div>
           <div>
             <Route path="/basic" breadcrumb="Basic" navigationHint="tab">
-              <div class="h-96 overflow-y-auto pr-4">
+              <div class="overflow-y-auto pr-4">
                 <label for="modalContainerName" class="block mb-2 text-sm font-medium text-gray-400 dark:text-gray-400"
                   >Container name:</label>
                 <input


benoitf:
I think it was for the initial size so you see the 'start button' without having to scroll/resize

on your screenshot I think you don't have the 'default size' of the window

cdrage:
> I think it was for the initial size so you see the 'start button' without having to scroll/resize
> 
> on your screenshot I think you don't have the 'default size' of the window

Ah, I was wondering why 96.

The problem is that when I'm resizing my window (especially on a mac with a high resolution screen), I always have to scroll.

benoitf:
Maybe should find something where you see the button by default but if you increase the height it's growing

jeffmaury:
For me the user experience is worse than before. If I have Podman Desktop maximized then I was able to access the Start Container button without the need to scroll whch is not the case anymore

cdrage:
Thanks all for the feedback, I'm going to ask @mairin / UI team for what they think / what we could do to improve this and revisit this later

mairin:
We should probably have a z-index 9999999999999^9999999 (😉) fixed position bar at the bottom of form screens where the primary action button goes and the content scrolls under.

cdrage:
> We should probably have a z-index 9999999999999^9999999 (😉) fixed position bar at the bottom of form screens where the primary action button goes and the content scrolls under.

You read my mind! I opened up https://github.com/containers/podman-desktop/issues/5926 at the same time you commented.

Issue #5675, opened 2024-01-24
Deployed pods not visible after restarting k8 clusters


cbr7:
### Bug description

If the user has a container deployed into a k8 cluster (like kind of minikube) and then turns off the clusters and afterwards starts it back up, the deployed pods will no longer be visible in the Pods page until after the user forces a k8 context change in the application.

### Operating system

macOS

### Installation Method

Installer from website/GitHub releases

### Version

1.7.0

### Steps to reproduce

1. Have kind and minikube k8 clusters available.
2. Deploy a container to kind cluster.
3. Pod should be visible in the Pods page.
4. Turn off the kind cluster either by turning off the control plane container from Containers page or from Settings -> Resources.
5. Turn the kind cluster back on.
6. Go to Pods page.
7. Notice that the deployed pods are no longer visible.
8. Switch k8 context to minikube
9. Switch back to kind cluster
10. Notice pods are again being displayed in the Pods page.

deboer-tim:
This sounds the same as the second point in #5311. Good to track it directly here, and likely indicates it's not related to the new informers.

Issue #5311, opened 2023-12-19
Informer stores improvements


deboer-tim:
### Is your enhancement related to a problem? Please describe

I've been testing with the ingress informer and draft PR for other informers. When they're connected they work excellent - changes are fast and responsive, feels way better than polling/watches. However, there are three things that need improvements:

- Because the informer stores don't use the normal stores/registry they don't show up on the Troubleshooting page. It would be nice if we could change them to naturally show up there when debugging, otherwise we need to add a new section.
- When I start with a running cluster things are fine, but if I start with (e.g.) the current context pointing to a Kind cluster that's stopped and then start it, the informer doesn't automatically sync. If I switch to another page and back it's usually fine though. We could use an event for internal/known clusters but not in general, so we likely need polling or some other mechanism to do an initial kick.
- If the cluster is stopped/disconnected it's really nice that the store isn't lost/reset, but we need some kind of disconnection indication in the UI.

### Describe the solution you'd like

See description.

deboer-tim:
For the three issues listed in the description, I am only going to resolve the second here:
1. **Stores not showing up in troubleshooting** - IMHO we are likely to switch to the new/context informers which do show up, so this issue should be deferred and reevaluated later.
2. **Informers not connecting** - I investigated this today and it is not what I thought: if we attempt to start the informers before we can connect to the current context, getCurrentNamespace() is undefined and we skip creating the informer and never try again. I am working on a PR to change all of the createXInformer() functions to connect _where the context will be_ to avoid this problem, which is what the context informers do.
3. **Connection status** - This is now covered separately by issue #5401.

deboer-tim:
I have a fix that works for the initial problem, but while building it I found other connection-related issues with the initial informers, and noticed several things that our newer informers handle (e.g. reconnection, backoff, jitter) that we'd also want to fix. Instead of doing this work twice, our focus has shifted to using the new informers everywhere and I'm closing this in favour of #6259.

PR #5917, opened 2024-02-09
Kubecontext live information (frontend)

feloy:
### What does this PR do?

This PR displays the kubecontext information (reachable, number of pods/deployments)

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
@@ -102,18 +102,18 @@ async function handleDeleteContext(contextName: string) {
                 {#if $kubernetesContextsState.get(context.name)?.reachable}
                   <div class="flex flex-row pt-2">
                     <div class="w-3 h-3 rounded-full bg-green-500"></div>
-                    <div class="ml-1 font-bold text-[9px] text-green-500">REACHABLE</div>
+                    <div class="ml-1 font-bold text-[9px] text-green-500" aria-label="context-reachable">REACHABLE</div>
                   </div>
                   <div class="flex flex-row gap-4 mt-4">
                     <div class="text-center">
                       <div class="font-bold text-[9px] text-gray-800">PODS</div>
-                      <div class="text-[16px] text-white">
+                      <div class="text-[16px] text-white" aria-label="context-pods-count">
                         {$kubernetesContextsState.get(context.name)?.podsCount}
                       </div>
                     </div>
                     <div class="text-center">
                       <div class="font-bold text-[9px] text-gray-800">DEPLOYMENTS</div>
-                      <div class="text-[16px] text-white">
+                      <div class="text-[16px] text-white" aria-label="context-deployments-count">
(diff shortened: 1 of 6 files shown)


packages/renderer/src/stores/kubernetes-contexts-state.ts
```diff
@@ -0,0 +1,27 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
deboer-tim: 2024

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -95,32 +96,66 @@ async function handleDeleteContext(contextName: string) {
...
+                  <div class="flex flex-row pt-2">
+                    <div class="w-3 h-3 rounded-full bg-gray-900"></div>
+                    <div class="ml-1 font-bold text-[9px] text-gray-900">UNREACHABLE</div>
+                  </div>
+                {/if}
+              </div>
+            {/if}
+            <div class="grow">
+              <div class="text-xs bg-charcoal-800 p-2 rounded-lg mt-1 grid grid-cols-6">
+                <span class="my-auto font-bold col-span-1 text-right">CLUSTER</span>
```
deboer-tim: I still have the problem that if the screen is too narrow, NAMESPACE doesn't fit in its column and overwrites the value, or the url is too long and hangs outside of the box. Can you add a 'overflow-hidden text-ellipsis' to at least the spans for CLUSTER, NAMESPACE, and url... or maybe just drop it in all of them in case anyone is using a large font? Doing this fixes any ugliness for me and is cropping individual fields at a reasonable point.

deboer-tim (commented):
Tested. No more flashing and removing the replica-set solved the main UI issue. Added two minor comments (and #5912 needs to merge) but otherwise I'm happy with this. Maybe just add a test that the new pod/deployment counts are in the UI?

Also confirmed that these informers are better at picking up new/restarted clusters than our other informer pattern. Nicely done 👍🏼. I'm pretty sure that's the issue we need to fix in #5311 (second issue) and #5675 too - which we could apply the same thing as here, or (as per my essay in #5912) move everything to these informers.

feloy:
> Tested. No more flashing and removing the replica-set solved the main UI issue. Added two minor comments (and #5912 needs to merge) but otherwise I'm happy with this. Maybe just add a test that the new pod/deployment counts are in the UI?
> 
> Also confirmed that these informers are better at picking up new/restarted clusters than our other informer pattern. Nicely done 👍🏼. I'm pretty sure that's the issue we need to fix in #5311 (second issue) and #5675 too - which we could apply the same thing as here, or (as per my essay in #5912) move everything to these informers.

I have made the changes. Thanks for the review

deboer-tim (approved):
Confirmed my comments were addressed. Still needs #5912 to merge but looks good otherwise. 👍🏼

dgolovin (approved):
LGTM and WFM

lstocchi (approved):

PR #5931, opened 2024-02-09
chore: extract color palette to its own file

benoitf:
### What does this PR do?
make the palette color being more embeddable elsewhere or reusable
it is like a no-op change

sort colors alphabetically

related to:
- https://github.com/containers/podman-desktop/issues/5878
- https://github.com/containers/podman-desktop/issues/4907

### How to test this PR?

Podman Desktop or website colors should be the same as before


deboer-tim (approved):
I'm curious where you're going with this, but since it is a safe and simple change I'm happy to approve.

cdrage (approved):

PR #5933, opened 2024-02-09
chore: use input in create volume

deboer-tim:
### What does this PR do?

Use the Input component when creating a volume.

Bonus fix: button should be disabled until the volume has a name (if I don't pick a name, UI hung).

### Screenshot / video of UI

Screenshot 2024-02-09 at 2 49 14 PM

Fixes another minor part of #3234.

### How to test this PR?

Create a volume.


benoitf (changes requested):
you can create a volume without a name ( I mean it's expected that you could)

deboer-tim:
ok, I will remove that part/not change it... but that definitely hung PD for me when I tried. Will try again.

benoitf:
ok maybe there is a bug but 'podman create volume' is working (without a name)

deboer-tim:
Ok, that part of the commit is removed and it is working for me now (auto-generates a name). Sorry, no idea why it was failing yesterday.

benoitf (approved):

PR #5935, opened 2024-02-09
chore: use input when creating pod from containers

deboer-tim:
### What does this PR do?

Use the Input component when creating a pod from containers.

### Screenshot / video of UI

Screenshot 2024-02-09 at 2 52 26 PM

Fixes another minor part of #3234.

### How to test this PR?

Select a container and click the Create Pod button.


benoitf (approved):

packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
```diff
@@ -14,6 +14,7 @@ import Fa from 'svelte-fa';
 import SolidPodIcon from '../images/SolidPodIcon.svelte';
 import Button from '../ui/Button.svelte';
 import type { PodCreatePortOptions } from '../../../../main/src/plugin/dockerode/libpod-dockerode';
+import Input from '../ui/Input.svelte';
```
benoitf: probably need to use `/@/` format for the import rather than `../..`

deboer-tim: Updated.

PR #5939, opened 2024-02-09
chore: use input for registries

deboer-tim:
### What does this PR do?

Use the Input component when editing registries. Switching to use Input and PasswordInput I had to clean up some styling and 'use:' code.

Found two minor things needed for the components:
- Input needs to have overflow-hidden to work in narrow spaces.
- Password was missing a placeholder.

While testing I found *lots* of UI issues that are unrelated to this change. None of the columns/sections are aligned, have different spacing/styling/padding or none between user/password/button columns, sometimes controls touching each other, etc. I have another PR ready to go, but am keeping this one limited to just the input component change to keep things separate.

### Screenshot / video of UI

Screenshot 2024-02-09 at 5 14 03 PM

Another small part of #3234.

### How to test this PR?

Edit all three kinds of registries.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
@@ -7,6 +7,8 @@ import DropdownMenuItem from '../ui/DropDownMenuItem.svelte';
 import { faPlusCircle, faTrash, faUser, faUserPen } from '@fortawesome/free-solid-svg-icons';
 import SettingsPage from './SettingsPage.svelte';
 import Button from '../ui/Button.svelte';
+import Input from '/@/lib/ui/Input.svelte';
+import PasswordInput from '/@/lib/ui/PasswordInput.svelte';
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
```diff
@@ -7,6 +7,8 @@ import DropdownMenuItem from '../ui/DropDownMenuItem.svelte';
 import { faPlusCircle, faTrash, faUser, faUserPen } from '@fortawesome/free-solid-svg-icons';
 import SettingsPage from './SettingsPage.svelte';
 import Button from '../ui/Button.svelte';
+import Input from '/@/lib/ui/Input.svelte';
+import PasswordInput from '/@/lib/ui/PasswordInput.svelte';
```
benoitf, line 11: Thanks for the @ imports👍

benoitf (approved):
I reviewed but forgot the submit review button

PR #5943, opened 2024-02-11
chore: use input for proxy settings

deboer-tim:
### What does this PR do?

Use the Input component when configuring a proxy. This is the first control to require disabled inputs, so that support is added here - from the design it is the same as readonly except user text is the same color as placeholder text.

### Screenshot / video of UI

Screenshot 2024-02-11 at 12 08 19 AM
Screenshot 2024-02-11 at 12 08 29 AM

### What issues does this PR fix or reference?

Fixes another minor part of #3234.

### How to test this PR?

Configure a proxy, disable/enable the slider.

PR #5944, opened 2024-02-11
chore: cleanup dark: prefix colors

benoitf:
### What does this PR do?
it is not necessary to have the `dark:` prefix colors as there is another
definition without `dark:` prefix and we won't be using tailwind
theming but color aliasing so we don't need these `dark:` colors


deboer-tim (approved):
May need a rebase/merge due to #5943.

benoitf:
it looks like #5943 has conflict on Input field.

This PR should not conflict to #5943  looks like you're not changing the lines with dark: style

Issue #5855, opened 2024-02-05
Investigate support for building quadlet files from pods.


sallyom:
### Is your enhancement related to a problem? Please describe

There is a lot of boilerplate configuration when creating quadlet files from pods, investigate whether desktop can generate quadlet files from container & pod definitions.
- list of images required by service
- quadlet.kube referencing a YAML
- quadlet.container with ports, mounts

### Describe the solution you'd like

I'd like desktop to offer a suggested quadlet for pod or containers.

cdrage:
I've been using podlet to generate my files: https://github.com/k9withabone/podlet

You can generate a quadlet from a running container or even from `podman run -p 8080:8080 etc` commands using https://github.com/k9withabone/podlet

I'd be awesome to have this as an extension for Podman Desktop since it's a singular binary.

I only suggest this as whenever I've used the systemd generator tool built into podman there is always the large DEPRECATED notice.

PR #5949, opened 2024-02-11
[poc] Dialog form for extensions

feloy:
### What does this PR do?

Proposal for a Form Dialog extension

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1513,6 +1513,8 @@ declare module '@podman-desktop/api' {
      */
     export function showErrorMessage(message: string, ...items: string[]): Promise<string | undefined>;
 
+    export function showFormDialog(title: string): Promise<void>;
+
     /**
      * Show progress in Podman Desktop. Progress is shown while running the given callback
      * and while the promise it returned isn't resolved nor rejected. The location at which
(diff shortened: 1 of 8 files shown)


benoitf (changes requested):
I don't think we would need yet another custom dialog

We have informationMessage/WarningMessage/ErrorMessage, quickpick and inputbox

cdrage:
https://github.com/containers/podman-desktop/assets/6422176/e47954c3-ae50-4b50-b15e-2d1f0b941b6c

Awesome work! I believe this would be very beneficial for any extensions that requires to ask the user multiple questions that would exceed what quickpick could do. 

Ex. simple "checkbox", or being able to see what the previous option was / go "back" and change it.

If someone were to create a https://github.com/kubernetes/kompose extension, it would be nice to ask via a form if I would like to use Deployments, as well as checkbox, and whatever else I'd like to add. 

Another point of discussion was a "quadlet" generator tool #5855 and it would beneficial having a form-style input for generating something as complicated as a systemd file.

cdrage:
My feedback:
* Integration of checkboxes?
* Follow "form" UI style-guide as per UI team mockups
* Clickable icons (perhaps PR in the future?) instead of "select" if I wanted to for example use a Linux icon for Linux, etc.

benoitf:
I would close this PR as we don't need another way

feloy:
> I would close this PR as we don't need another way


Closing this PR, as not a priority

PR #5958, opened 2024-02-12
chore: include a color registry

benoitf:
### What does this PR do?
This is the place where to define colors for light/dark theme
register color definition of color names provided by UX team

### How to test this PR?

unit tests provided

### DIFF
--- a/packages/main/src/plugin/api/color-info.ts
+++ b/packages/main/src/plugin/api/color-info.ts
@@ -26,4 +26,5 @@ export interface ColorDefinition {
 export interface ColorInfo {
   id: string;
   value: Color;
+  cssVar: string;
 }
--- a/packages/main/src/plugin/api/color-info.ts
+++ b/packages/main/src/plugin/api/color-info.ts
@@ -1,0 +1,29 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


benoitf:
@cdrage so there is always a light and dark color, but you can add custom theme and the theme has a parent (either dark or light) so if you don't provide a value for a color, it will default to the dark/light color

deboer-tim:
When I saw #5931 and then RawThemeContribution here I thought themes were going to be delivered through separate json files or contributions, but both light/dark are declared in color-registry.ts.

I know the following example is likely too simplistic, but it seems like it would be easier to manage the themes if they were defined in files like 'dark.json'/'light.json' or contributions?
```
{ "GlobalNav": {
  "Bg":"charcoal-800",
  "Icon": "gray-600",
  ...
```

benoitf:
I choose typescript registration over configuration files as we might use some variables to define colors and because I have type safety

This is the definition of the colors so no separate files as for example you may miss colors ( like you add only in light.json or only in dark.json or you do typos and then the color is missing)

But for sure an extension will be able to contribute a set of colors that will either define or inherits from it's parent theme in case it is missing

So definition is in the code, contribution is with .json files

benoitf:
I've added the other PRs depending on this PR so you can see how it works

at the end we have colors registered as CSS variables for Podman Desktop and for the webviews (so they'll be able to reuse the same colors)

deboer-tim (approved):

lstocchi (approved):
LGTM

cdrage (approved):

benoitf:
updated to use kebab-case color identifier (before registering `TitlebarBg` and now registering `titlebar-bg`) (still using camelCase for JSON contribution)

when listing colors, we specify the name of the CSS variable as well so no need to compute it on client side

css variable name is using the pattern `--pd-<name-of-the-color>` for example `--pd-titlebar-bg`

benoitf:
FYI @deboer-tim @cdrage and @lstocchi as you reviewed/approved before this change

deboer-tim:
Approval still ✅ after changes.

cdrage (approved):

PR #5962, opened 2024-02-12
feat: publish colors to the style of the app

benoitf:
### What does this PR do?
it's publishing the colors in CSS with variables
so components can use the definition of these colors

register each color defined under the name `--pd-<name-of-the-color>`

depends on:
- [x] https://github.com/containers/podman-desktop/pull/5958
- [x] https://github.com/containers/podman-desktop/pull/5960

### Screenshot / video of UI
image

### How to test this PR?

unit tests provided

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -50,6 +50,7 @@ import IngressesRoutesList from './lib/ingresses-routes/IngressesRoutesList.svel
 import Webview from './lib/webview/Webview.svelte';
 import IngressDetails from './lib/ingresses-routes/IngressDetails.svelte';
 import RouteDetails from './lib/ingresses-routes/RouteDetails.svelte';
+import ColorsStyle from './lib/style/ColorsStyle.svelte';
 
 router.mode.hash();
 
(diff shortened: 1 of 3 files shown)


deboer-tim (commented):
FontAwesome uses `--fa-` prefix, Tailwind uses `--tw-`. Could we just use `--pd-`?

The mix of both - and camelcase (i.e. --a-b-CandD) is a little jarring to me, and most others seem to use only -. If this is the main way people will interact I'm wondering if we should switch, or maybe even auto-convert when we expose as vars?

benoitf:
yes I was thinking the same after I started the implementation. (short name rather than podman-desktop)
I'll change

About color names I think they need to have the same name with or without the prefix

so if we want `--pd-globalNavBg` rather than `--pd-GlobalNavBg` we'll need to change the color definition

benoitf:
and we might expose the color palette colors as well but it's surely a separate PR

benoitf:
updated the PR with new shorter color scheme `--pd-color-name`

benoitf:
@deboer-tim are we good now ?

deboer-tim (approved):
Yup, thanks for the changes.

PR #5963, opened 2024-02-12
feat: provides css colors to webviews

benoitf:
### What does this PR do?
Register colors for webviews
so components can use the definition of these colors
register each color defined under the name `--podman-desktop-<name-of-the-color>`

depends on:
- [x] https://github.com/containers/podman-desktop/pull/5958
- [x] https://github.com/containers/podman-desktop/pull/5960

### How to test this PR?

unit tests provided


deboer-tim (approved):
Code & tests 👍🏼. This feels like something we need a webview or sample to start using to really know if we need tweaks, but merging is the right step to get there.

Issue #3233, opened 2023-07-14
UI Tweak: Registries page


mairin:
### Is your enhancement related to a problem? Please describe

I think the visual styling on the Registries page is a little out of step with the rest of the app. It would be nice to update it for consistency / to better match styles established elsewhere in the app. 

Here is a screenshot of how it looks in main today:

![Screenshot from 2023-07-14 15-01-06](https://github.com/containers/podman-desktop/assets/799683/7e8fc391-5c84-4cba-b2ef-c018e65b2b9a)

Two main issues stand out to me:

- The table styling (both the header styling and the use of lines instead of space to delinate rows particularly)
- The placement of the "Add Registry" action button, which on pretty much every other screen is placed in the upper right. I think it's easily missed on the bottom here.

### Describe the solution you'd like

I would suggest something that looks like this:

![Screenshot from 2023-07-14 15-01-06__@](https://github.com/containers/podman-desktop/assets/799683/c85ce3e2-2c32-4449-a3b2-c6955d792c47)

This moves the action button to the upper left (vertically centered along the same line as the page headline "Registries") and it also changes the visual styling of the table to be more consistent with other tables (like the pods/containers/images tables.) 

### Additional context

Note I changed the "Configure" button on suggested registries in the list to be the secondary "outline" form of the button, beacuse the "add registry" button should be the primary button on the page imo.

deboer-tim:
#3245 moved the New button to the header, and #5965 would clean up several alignment issues. Long term I think we need this issue still open to move this page to a Table, to solve these issues and for consistency.

PR #5965, opened 2024-02-12
chore: registry settings layout cleanup

deboer-tim:
### What does this PR do?

The registries page was three columns (location, user, then password & all actions) which caused odd scaling, esp. with the password column. The three types of registries (configured, suggested, and user added) also had different padding and margins for columns, which meant that editing a password would sometimes have touching controls or columns too wide to fit.

This just cleans it up into four consistent columns (location, user, password, and actions), with fixed widths (location 2x, all others 1x), and removes mix old or unused divs or flex classes.

### Screenshot / video of UI

Before:

https://github.com/containers/podman-desktop/assets/19958075/7c7c5a2a-5e6a-49d5-8253-1f62092af730

After:

https://github.com/containers/podman-desktop/assets/19958075/b1726cd4-49af-473a-8192-b1135558edde

N/A

### How to test this PR?

Edit/add registries.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
@@ -236,9 +236,10 @@ function removeExistingRegistry(registry: containerDesktopAPI.Registry) {
     <!-- Registries table start -->
     <div class="w-full border-t border-b border-gray-900" role="table" aria-label="Registries">
       <div class="flex w-full space-x-2" role="rowgroup" aria-label="header">
-        <div class="flex-1 text-left py-4 pl-5 text-sm font-bold w-auto" role="columnheader">Registry Location</div>
-        <div class="text-left py-4 text-sm font-bold w-1/4" role="columnheader">Username</div>
-        <div class="text-left py-4 text-sm font-bold w-2/5" role="columnheader">Password</div>
+        <div class="text-left py-4 text-sm font-bold w-2/5 pl-5" role="columnheader">Registry Location</div>
+        <div class="text-left py-4 text-sm font-bold w-1/5" role="columnheader">Username</div>
+        <div class="text-left py-4 text-sm font-bold w-1/5" role="columnheader">Password</div>
+        <div class="text-left py-4 text-sm font-bold w-1/5" role="columnheader"></div>
       </div>
 
       {#each $registriesInfos as registry}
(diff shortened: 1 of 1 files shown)


benoitf:
you fixed an issue after we click on buttons but in the default view, when I go to the page I'm not sure to like that the buttons are not anymore right aligned with the kebab menu and the eye icon being in the middle vs close to the kebab menu before

deboer-tim:
I'm pretty sure the original alignment was just by accident, and when you click on Configure the Login/Cancel don't line up at all. TBH I noticed too and wasn't sure what the fix should be, but I think it should be full alignment - maybe all the buttons right-justified?

benoitf:
I found https://github.com/containers/podman-desktop/issues/3233 and https://github.com/containers/podman-desktop/issues/4294 (but here problem is skipped ;-)

deboer-tim:
I had almost forgotten about #3233. Based on either link it looks like this should eventually be a Table, but based on the current implementation that could be a pain and not sure we should spend too much time atm.

jeffmaury:
I found strange the default layout when I switch to edit mode:

image

Seems to be that there is space for the Cancel button

deboer-tim:
The Cancel button is slightly wider than Login plus these is a small gap between them now. If you expand and contract the window a bit you can find the shift point, but I'm pretty sure the browser is correct :) and keep in mind the old behaviour was no gap and to let the Cancel button disappear under the right side of the window.

I played with the widths to give that column enough that it won't wrap until the window is fairly small, and to me it was handling this better than before - but at some point we're only going to see improvement if we rewrite the page with something like Table.

lstocchi:
I read the comments above. What about adding some flex justify center/end to the single actions cells? 
Something like 
https://drive.google.com/file/d/1FgTbskhnqxL3ucIB9BoXQSdIFT3fiIpR/view?usp=sharing

this is the edited code - there could be some hidden problems, just wanted to try quickly https://gist.github.com/lstocchi/06133861271f6ecbf97d11c83c39eceb

deboer-tim:
Rebased on main and added the flex as @lstocchi suggested, but went with left-justification to match the Configure/Login-Cancel left justification in other rows.

Screenshot 2024-02-25 at 2 05 26 PM

benoitf (approved):
approving as it's better than what we have in main but I'm still not sure about that left alignment

jeffmaury (approved):
LGTM

deboer-tim:
> approving as it's better than what we have in main but I'm still not sure about that left alignment

I'll take a look at changing the alignment in another PR.

benoitf:
@deboer-tim you may wait for https://github.com/containers/podman-desktop/issues/4294 feedback

PR #5975, opened 2024-02-13
feat: open terminal in the running container

### LINKED ISSUE #4402: K8s pods actions

vzhukovs:
### What does this PR do?

The following changes request adds the ability to open a terminal in a container located in a Kubernetes pod.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/1968177/eb5b498e-3e7b-4e85-bb38-e54b516849bf

partially: #4402 

### How to test this PR?

Create the pod in kubernetes(in our case sample app was deployed in OpenShift Local cluster), navigate to the deployed pod and open Terminal tab.

### DIFF
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -36,12 +36,14 @@ import type {
   V1Pod,
   V1PodList,
   V1Service,
+  V1Status,
 } from '@kubernetes/client-node';
 import {
   ApisApi,
   AppsV1Api,
   CoreV1Api,
   CustomObjectsApi,
+  Exec,
   HttpError,
   KubeConfig,
   KubernetesObjectApi,
@@ -65,6 +67,7 @@ import type { FilesystemMonitoring } from './filesystem-monitoring.js';
 import type { KubeContext } from './kubernetes-context.js';
 import type { ContextGeneralState, ResourceName } from './kubernetes-context-state.js';
 import { ContextsManager } from './kubernetes-context-state.js';
+import { BufferedStreamWriter, ResizableTerminalWriter, StringLineReader } from './kubernetes-exec-transmitter.js';
 import { Uri } from './types/uri.js';
 
 interface KubernetesObjectWithKind extends KubernetesObject {
@@ -1239,5 +1242,71 @@ export class KubernetesClient {
 
   public dispose(): void {
     this.contextsState.dispose();
+  }
+
+  async execIntoContainer(
+    podName: string,
+    containerName: string,
+    onStdOut: (data: Buffer) => void,
+    onStdErr: (data: Buffer) => void,
+    onClose: () => void,
+  ): Promise<{ onStdIn: (data: string) => void; onResize: (columns: number, rows: number) => void }> {
+    let telemetryOptions = {};
+    try {
+      const ns = this.getCurrentNamespace();
+      const connected = await this.checkConnection();
+      if (!ns) {
+        throw new Error('no active namespace');
+      }
+      if (!connected) {
+        throw new Error('not active connection');
+      }
+
+      const stdout = new ResizableTerminalWriter(new BufferedStreamWriter(onStdOut));
+      const stderr = new ResizableTerminalWriter(new BufferedStreamWriter(onStdErr));
+      const stdin = new StringLineReader();
+
+      const exec = new Exec(this.kubeConfig);
+      const conn = await exec.exec(
+        ns,
+        podName,
+        containerName,
+        ['/bin/sh', '-c', 'if command -v bash >/dev/null 2>&1; then bash; else sh; fi'],
+        stdout,
+        stderr,
+        stdin,
+        true,
+        (_: V1Status) => {
+          // need to think, maybe it would be better to pass exit code to the client, but on the other hand
+          // if connection is idle for 15 minutes, websocket connection closes automatically and this handler
+          // does not call. also need to separate SIGTERM signal (143) and normally exit signals to be able to
+          // proper reconnect client terminal. at this moment we ignore status and rely on websocket close event
+        },
+      );
+
+      //need to handle websocket idling, which causes the connection close which is not passed to the execution status
+      //approx time for idling before closing socket is 15 minutes. code and reason are always undefined here.
+      conn.on('close', () => {
+        onClose();
+      });
+
+      return {
+        onStdIn: (data: string): void => {
+          stdin.readLine(data);
+        },
+        onResize: (columns: number, rows: number): void => {
+          if (columns <= 0 || rows <= 0 || isNaN(columns) || isNaN(rows) || columns === Infinity || rows === Infinity) {
+            throw new Error('resizing must be done using positive cols and rows');
+          }
+
+          (stdout as ResizableTerminalWriter).resize({ width: columns, height: rows });
+        },
+      };
+    } catch (error) {
+      telemetryOptions = { error: error };
+      throw this.wrapK8sClientError(error);
+    } finally {
+      this.telemetry.track('kubernetesExecIntoContainer', telemetryOptions);
+    }
   }
 }
--- a/packages/main/src/plugin/kubernetes-exec-transmitter.ts
+++ b/packages/main/src/plugin/kubernetes-exec-transmitter.ts
@@ -1,0 +1,116 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { WritableOptions } from 'node:stream';
+import { Readable, Writable } from 'node:stream';
+
+export const DEFAULT_COLUMNS: number = 80;
+export const DEFAULT_ROWS: number = 60;
+
+export interface TerminalSize {
+  height: number;
+  width: number;
+}
+
+export class ExecStreamWriter extends Writable {
+  protected transmitter: Writable;
+
+  constructor(transmitter: Writable, options?: WritableOptions) {
+    super(options);
+    this.transmitter = transmitter;
+  }
+
+  _write(chunk: unknown, encoding: BufferEncoding, callback: (error?: Error | null) => void): void {
+    this.transmitter._write(chunk, encoding, callback);
+  }
+}
+
+export class ResizableTerminalWriter extends ExecStreamWriter {
+  protected columns: number;
+  protected rows: number;
+
+  constructor(
+    transmitter: Writable,
+    terminalSize: TerminalSize = { width: DEFAULT_COLUMNS, height: DEFAULT_ROWS },
+    options?: WritableOptions,
+  ) {
+    super(transmitter, options);
+    this.columns = terminalSize.width;
+    this.rows = terminalSize.height;
+  }
+
+  _write(chunk: unknown, encoding: BufferEncoding, callback: (error?: Error | null) => void): void {
+    super._write(chunk, encoding, callback);
+  }
+
+  resize(terminalSize: TerminalSize): void {
+    this.columns = terminalSize.width;
+    this.rows = terminalSize.height;
+
+    this.emit('resize');
+  }
+
+  getDimension(): TerminalSize {
+    return {
+      width: this.columns,
+      height: this.rows,
+    } as TerminalSize;
+  }
+}
+
+export class BufferedStreamWriter extends Writable {
+  private readonly transmitFn: (data: Buffer) => void;
+
+  constructor(transmitFn: (data: Buffer) => void, options?: WritableOptions) {
+    super(options);
+    this.transmitFn = transmitFn;
+  }
+
+  _write(chunk: Buffer, encoding: BufferEncoding, callback: (error?: Error | null) => void): void {
+    this.transmitFn(chunk);
+    callback();
+  }
+}
+
+export class StringLineReader extends Readable {
+  private dataQueue: string[] = [];
+  private isReading: boolean = false;
+
+  constructor() {
+    super();
+  }
+
+  readLine(data: string): void {
+    this.dataQueue.push(data);
+    if (!this.isReading) {
+      this.read();
+    }
+  }
+
+  _read(): void {
+    this.isReading = true;
+    const data = this.dataQueue.shift();
+
+    if (data) {
+      this.push(data);
+    } else {
+      this.isReading = false;
+      this.push(undefined);
+    }
+  }
+}
--- a/packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
+++ b/packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
@@ -1,0 +1,86 @@
+<script lang="ts">
+import { onMount } from 'svelte';
+
+import EmptyScreen from '/@/lib/ui/EmptyScreen.svelte';
+import NoLogIcon from '/@/lib/ui/NoLogIcon.svelte';
+import { podsInfos } from '/@/stores/pods';
+
+import { terminalService } from './KubernetesTerminalService';
+import type { PodInfoUI } from './PodInfoUI';
+
+let key = 0;
+
+export let pod: PodInfoUI;
+
+$: currentContainerStatus =
+  $podsInfos
+    .find(p => p.Name === pod.name)
+    ?.Containers.reduce((acc, c) => {
+      acc.set(c.Names, c.Status);
+      return acc;
+    }, new Map<string, string>()) || new Map<string, string>();
+
+let currentContainerName = '';
+
+onMount(() => {
+  if (pod.containers.length > 0) {
+    currentContainerName = pod.containers[0].Names;
+    terminalService.ensureTerminalExists(pod.name, currentContainerName);
+  }
+  key++;
+});
+
+function handleSelectionChange(event: Event) {
+  const target = event.target as HTMLSelectElement;
+  currentContainerName = target.value;
+  terminalService.ensureTerminalExists(pod.name, currentContainerName);
+  key++;
+}
+</script>
+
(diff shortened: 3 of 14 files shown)


axel7083:
What is the expected behavior when multiple containers are in the pods ?

cdrage:
> What is the expected behavior when multiple containers are in the pods ?

Was about to ask the same thing. We would have to implement an option / drop down menu to select which container to connect to.

cdrage:
This is absolutely awesome though @vzhukovs ! With edit / summary / logs and all, this was the last piece of the puzzle for Kubernetes support. Amazing work 🎉

cdrage:
Screenshot 2024-02-13 at 8 50 32 AM

Woo! So there IS a dropdown menu. That's awesome. Although for some reason it's not showing the terminals correctly? And I'm not getting any error / showing. @vzhukovs

vzhukovs:
> > What is the expected behavior when multiple containers are in the pods ?
> 
> Was about to ask the same thing. We would have to implement an option / drop down menu to select which container to connect to.

Yep, the logic of choosing the target container is the same as it implemented in OpenShift console. If pod has only one running container, then it will be automatically chosen as destination one, but if the pod has more then one container, then there will be a dropdown menu above the terminal to choose the right one container.

> lthough for some reason it's not showing the terminals correctly? And I'm not getting any error / showing.

@cdrage you mean, that terminals don't switch after selecting the right one, am I right?

UPD: seems, I found the problematic place, taking a look on it

slemeur:
Looking really great @vzhukovs !

packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
```diff
@@ -0,0 +1,85 @@
...
+<div class="flex py-2">
+  <label
+    for="input-standard-{pod.name}"
+    class="block w-auto text-sm font-medium whitespace-nowrap leading-6 text-gray-900 pl-2 pr-2">Connecting to:</label>
+  <div class="w-full">
+    {#if pod.containers.length > 1}
+      <select
+        on:change="{handleSelectionChange}"
+        aria-labelledby="listbox-label"
+        class="border-b block w-48 p-1 bg-zinc-700 border-violet-500 text-white text-sm checked:bg-violet-50"
```
deboer-tim: nit: zinc is one of our historical colors and shouldn't be used.

packages/renderer/src/lib/pod/PodDetails.svelte
```diff
@@ -65,6 +66,9 @@ onMount(() => {
       <Tab title="Logs" url="logs" />
       <Tab title="Inspect" url="inspect" />
       <Tab title="Kube" url="kube" />
+      {#if pod.kind === 'kubernetes'}
+        <Tab title="Terminal" url="k8s/terminal" />
```
deboer-tim: I noticed that when you go to this tab the breadcrumb at the top goes from `Pod > Inspect` to `Pod Details > Terminal`. I'm not entirely sure what's causing this, but likely due to the tab being two folders down from the parent. Verified that using a simpler path (e.g. `k8s-terminal`) in all 3 places fixes it, so that's probably the easiest/most consistent option.

packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
```diff
@@ -0,0 +1,85 @@
...
+
+$: currentContainerStatus =
+  $podsInfos
+    .find(p => p.Name === pod.name)
+    ?.Containers.reduce((acc, c) => {
+      acc.set(c.Names, c.Status);
+      return acc;
+    }, new Map<string, string>()) || new Map<string, string>();
+
+let terminalCache = new Map();
```
deboer-tim: Limited testing, but this seems to never get cleaned up? Having the history and being able to jump back to the last terminal without losing anything is excellent, but I worry this has a reference to large components and could constantly grow over time. Could you confirm that's not the case, or maybe add some simple clearing like an onDispose() that only keeps e.g. ones that have been touched recently or the last 5?

deboer-tim (changes requested):
Hard to review such a large PR, but this is both great work and working well when testing. 👍🏼 Just a couple minor issues I've found so far.

feloy:
terminal-exit

When I "exit" from the terminal, I'm not able to get a new connection. Could it be possible to respawn the connection, either automatically, or after some key pressed?

packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
vzhukovs: updated, reused the same colours from the select element in Run Image page

vzhukovs: Снимок экрана 2024-02-27 в 17 08 17

packages/renderer/src/lib/pod/PodDetails.svelte
vzhukovs: also updated, breadcrumbs are displaying correctly now

packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
vzhukovs: revised a bit this mechanism, included additional checkings for container status update and pod removing, screencast provided below

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1327,4 +1331,69 @@ export class KubernetesClient {
...
+        ns,
+        podName,
+        containerName,
+        ['/bin/sh', '-c', 'if command -v bash >/dev/null 2>&1; then bash; else sh; fi'],
+        stdout,
+        stderr,
+        stdin,
+        true,
+        (_: V1Status) => {
+          // need to think, maybe it would be better to pass exit code to the client, but on the other hand
```
cdrage, line 1280: comment / still need to implement here?

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1327,4 +1331,69 @@ export class KubernetesClient {
...
+    onStdOut: (data: Buffer) => void,
+    onStdErr: (data: Buffer) => void,
+    onClose: () => void,
+  ): Promise<{ onStdIn: (data: string) => void; onResize: (columns: number, rows: number) => void }> {
+    let telemetryOptions = {};
+    try {
+      const ns = this.getCurrentNamespace();
+      const connected = await this.checkConnection();
+      if (!ns || !connected) {
+        throw new Error('not connected or namespace is empty');
```
cdrage: we should have separate errors instead of saying "or"

like..

"no namespace"

and

"not connected"

packages/main/src/plugin/kubernetes-exec-transmitter.ts
```diff
@@ -0,0 +1,119 @@
...
+    terminalSize: TerminalSize = { width: DEFAULT_COLUMNS, height: DEFAULT_ROWS },
+    options?: WritableOptions,
+  ) {
+    super(transmitter, options);
+    this.columns = terminalSize.width;
+    this.rows = terminalSize.height;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  _write(chunk: any, encoding: BufferEncoding, callback: (error?: Error | null) => void): void {
```
cdrage: i believe we should be using `unknown` instead of `any`  in new code, right @benoitf ?

cdrage (changes requested):
Screenshot 2024-03-01 at 3 35 23 PM

When connected, should say "Connected to:" ?

packages/main/src/plugin/kubernetes-exec-transmitter.ts
benoitf: yes avoid any

vzhukovs: fixed

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: splited into two different checks

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: I'd rather say, it simply a comment to improve handling exit signals in future, at this moment everything works fine

vzhukovs:
> Screenshot 2024-03-01 at 3 35 23 PM
> 
> When connected, should say "Connected to:" ?

improved, now when terminal is connected, the label also shows information, that it is connected

vzhukovs:
> terminal-exit
> 
> When I "exit" from the terminal, I'm not able to get a new connection. Could it be possible to respawn the connection, either automatically, or after some key pressed?

@feloy fixed this issue, now reconnecting to the new terminal should work fine

feloy:
When I try to connect to an image with neither `bash` nor `sh`, it is respawning the connection infinitely, without any delay. I don't know if you can detect this specific error, or at least make some temporization before retrying the connection (or canceling the retries after a few tries)

packages/main/src/plugin/kubernetes-exec-transmitter.ts
```diff
@@ -0,0 +1,116 @@
...
+    super(options);
+    this.transmitter = transmitter;
+  }
+
+  _write(chunk: unknown, encoding: BufferEncoding, callback: (error?: Error | null) => void): void {
+    this.transmitter._write(chunk, encoding, callback);
+  }
+}
+
+export class ResizeableTerminalWriter extends ExecStreamWriter {
```
dgolovin:
```suggestion
export class ResizableTerminalWriter extends ExecStreamWriter {
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1379,4 +1383,72 @@ export class KubernetesClient {
...
+      if (!ns) {
+        throw new Error('no active namespace');
+      }
+      if (!connected) {
+        throw new Error('not active connection');
+      }
+
+      const stdout: ExecStreamWriter = new ResizeableTerminalWriter(new BufferedStreamWriter(onStdOut));
+      const stderr: ExecStreamWriter = new ResizeableTerminalWriter(new BufferedStreamWriter(onStdErr));
+      const stdin: StringLineReader = new StringLineReader();
```
```suggestion
      const stdout = new ResizeableTerminalWriter(new BufferedStreamWriter(onStdOut));
      const stderr = new ResizeableTerminalWriter(new BufferedStreamWriter(onStdErr));
      const stdin = new StringLineReader();
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -1379,4 +1383,72 @@ export class KubernetesClient {
...
+          stdin.readLine(data);
+        },
+        onResize: (columns: number, rows: number): void => {
+          if (columns <= 0 || rows <= 0 || isNaN(columns) || isNaN(rows) || columns === Infinity || rows === Infinity) {
+            throw new Error('resizing must be done using positive cols and rows');
+          }
+
+          if (stdout instanceof ResizeableTerminalWriter) {
+            (stdout as ResizeableTerminalWriter).resize({ width: columns, height: rows } as TerminalSize);
+          }
```
```suggestion
          stdout.resize({ width: columns, height: rows });
```

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -69,6 +71,8 @@ import type { KubernetesInformerResourcesType } from './api/kubernetes-informer-
 import type { IncomingMessage } from 'node:http';
 import { ContextsManager } from './kubernetes-context-state.js';
 import type { ContextGeneralState, ResourceName } from './kubernetes-context-state.js';
+import type { ExecStreamWriter, TerminalSize } from './kubernetes-exec-transmitter.js';
```
```suggestion
```

dgolovin (changes requested):

dgolovin (changes requested):
Resize for terminal is not working. I am not sure is if it is PR problem or kubernetes-client issue.

https://github.com/containers/podman-desktop/assets/620330/5822717f-3a59-4478-9633-5ba95804db37

vzhukovs:
> When I try to connect to an image with neither `bash` nor `sh`, it is respawning the connection infinitely, without any delay. I don't know if you can detect this specific error, or at least make some temporization before retrying the connection (or canceling the retries after a few tries)

I suppose, there is should be the same behaviour, when user tries to open the terminal to a podman container, because the init command is the same as here: https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/container-registry.ts#L1568

vzhukovs:
> Resize for terminal is not working. I ma not sure is if it PR problem or kubernetes-client issue.
> 
>  Screen.Recording.2024-03-06.at.7.56.44.PM.mov

Hi @dgolovin, fixed, now resize works fine:

Снимок экрана 2024-03-07 в 12 56 23
Снимок экрана 2024-03-07 в 12 56 36

packages/main/src/plugin/kubernetes-exec-transmitter.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

packages/main/src/plugin/kubernetes-client.ts
vzhukovs: updated

jeffmaury (approved):
LGTM

cdrage (approved):
Tested, amazing work and the code LGTM!

Big 👍 on all the tests

vzhukovs:
@deboer-tim, green light to merge?

deboer-tim (approved):
Just back in the office today - my requests were done, and have only had time for a very quick review but happy to approve based on the others and get this in.

packages/main/src/plugin/kubernetes-exec-transmitter.ts
```diff
@@ -0,0 +1,116 @@
...
+    const data = this.dataQueue.shift();
+
+    if (data) {
+      this.push(data);
+    } else {
+      this.isReading = false;
+      this.push(undefined);
+    }
+  }
+}
```
benoitf, line 116: I'm not sure to understand the Transmitter

for terminal resize what we did was
https://github.com/containers/podman-desktop/pull/3955/files

packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
```diff
@@ -0,0 +1,86 @@
+<script lang="ts">
+import { onMount } from 'svelte';
+
+import { podsInfos } from '/@/stores/pods';
+
+import EmptyScreen from '../ui/EmptyScreen.svelte';
+import NoLogIcon from '../ui/NoLogIcon.svelte';
```
benoitf: should prefer /@/lib usage rather than `../../`

packages/renderer/src/stores/kubernetes-terminal-state-store.ts
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { writable } from 'svelte/store';
+
+export const terminalStates = writable(new Map());
```
benoitf, line 21: there is no test

benoitf (changes requested):
not clear to me why we're using special Transmitter

for other terminal usage it's just fitting with the terminal size so I don't see here why it would be different

packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
vzhukovs: fixed

packages/main/src/plugin/kubernetes-exec-transmitter.ts
vzhukovs: Yes, this approach works, when we have separate method for stdout resize, like dockerode has, but client-node library doesn't have separate method to change the stdout dimensions. When we create the exec connection, we have to pass streams for stdout, stderr and stdin:

https://github.com/kubernetes-client/javascript/blob/3610c1b20bccb08aeca3aa3d37ad0fd5c1fbc425/src/exec.ts#L37-L39

this is the first point why there were created [fictitious streams](https://github.com/containers/podman-desktop/pull/5975/files#diff-8799a75d3c2928c9fe5136fceab66ed0edad4de3a675f9f5f6d3bc3c61f3dd7bR1409-R1411), which consume the handlers for [input](https://github.com/containers/podman-desktop/pull/5975/files#diff-8799a75d3c2928c9fe5136fceab66ed0edad4de3a675f9f5f6d3bc3c61f3dd7bR1397) and [output](https://github.com/containers/podman-desktop/pull/5975/files#diff-8799a75d3c2928c9fe5136fceab66ed0edad4de3a675f9f5f6d3bc3c61f3dd7bR1394-R1395) operations. Then, mandatory condition is to implement this [interface](https://github.com/kubernetes-client/javascript/blob/master/src/terminal-size-queue.ts#L3-L7) in stdout stream which can handle resize event and has dimention values. And after that, by calling [resize](https://github.com/containers/podman-desktop/pull/5975/files#diff-8799a75d3c2928c9fe5136fceab66ed0edad4de3a675f9f5f6d3bc3c61f3dd7bR1446) on stream, we can fire [event](https://github.com/containers/podman-desktop/pull/5975/files#diff-e4eec56465f6b765285da8c5467d1e766c0bf93df11693904e792927b31c5d50R65) which will be [handled](https://github.com/kubernetes-client/javascript/blob/master/src/terminal-size-queue.ts#L28) in client-node library and pass the resize event through websocket to the container. That's why we should have a custom stream. To easier support the streams in future, I wrapped them as decorators.

vzhukovs:
> not clear to me why we're using special Transmitter
> 
> for other terminal usage it's just fitting with the terminal size so I don't see here why it would be different

Described general approach here in comment: https://github.com/containers/podman-desktop/pull/5975#discussion_r1530839772
In general, only fitting xterm addon doesn't apply resize on container side, environment variables `$COLUMNS` and `$LINES` still not modified, we need to implement ResizableStream from client-node library to be able to resize it correctly.

General approach in dockerode and client-node library are completely different to each other.

packages/renderer/src/stores/kubernetes-terminal-state-store.ts
vzhukovs: added test

benoitf (approved):

dgolovin (approved):

PR #5981, opened 2024-02-13
feat: add HealthCheck parameter when creating container

### LINKED ISSUE #5841: Extend Podman Desktop API to allow heathcheck for a container

AI Studio would like to implement healthcheck for containers in order to make sure they are properly started. That would require to specify heathchecks when creating a container

Extend the API to allow healthchechks to be specified when creating a container

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #5841

### What does this PR do?

Add HealthCheck parameter to the createContainer API

### Screenshot / video of UI

N/A

#5841

### How to test this PR?

Unit test provided

### DIFF
--- a/packages/main/src/plugin/api/container-info.ts
+++ b/packages/main/src/plugin/api/container-info.ts
@@ -71,6 +71,39 @@ export interface HostConfig {
   NetworkMode?: string;
 }
 
+export interface HealthConfig {
+  /**
+   * The test to perform. Possible values are:
+   *
+   * - ```[]``` inherit healthcheck from image or parent image
+   * - ```["NONE"]``` disable healthcheck
+   * - ```["CMD", args...]``` exec arguments directly
+   * - ```["CMD-SHELL", command]``` run command with system's default shell
+   */
+  Test?: string[];
+
+  /**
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/api/container-info.ts
```diff
@@ -71,6 +71,39 @@ export interface HostConfig {
   NetworkMode?: string;
 }
 
+export interface HealthConfig {
```
axel7083, line 74: Why not using the same wording as kubernetes ?
They seems to _mostly_ do the same things except for the ms vs seconds.

initialDelaySeconds -> StartPeriod
Interval -> periodSeconds
failureThreshold -> Retries


See https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

benoitf (approved):

packages/main/src/plugin/api/container-info.ts
benoitf: it's because it's the type exposed by the docker spec
https://github.com/moby/docker-image-spec/blob/main/specs-go/v1/image.go#L35-L53

axel7083: Okey ! I didn't know, thanks for letting me know

axel7083 (approved):
LGTM

PR #5988, opened 2024-02-13
chore: input errors, use input component in run image

deboer-tim:
### What does this PR do?

Uses the Input component in the Run Image form. This is the first time that we have inputs with 'errors', so I've added support to Input for an error. When there's an error the purple changes to red, there's a red X on the right, and the error message is listed underneath the input.

There is more work to do in this form (e.g. most buttons should probably be Buttons, still a little custom formatting), but I kept this PR to just replacing all the text/file inputs.

### Screenshot / video of UI

Screenshot 2024-02-13 at 4 02 46 PM

Another part of #3234.

### How to test this PR?

Run an image, tab through the form.


packages/renderer/src/lib/ui/Input.svelte
```diff
@@ -29,45 +30,58 @@ async function onClear() {
...
-      class:hidden="{!value || readonly || disabled}"
-      aria-label="clear"
-      on:click="{onClear}">
-      <Fa icon="{faXmark}" />
-    </button>
+<div class="flex flex-col w-full">
+  <div
+    class="flex flex-row w-full items-center px-1 py-1 group bg-charcoal-500 border-[1px] border-charcoal-500 {$$props.class ||
+      ''}"
+    class:hover:bg-charcoal-900="{!readonly && !disabled}"
```
benoitf, line 37: It looks like we should find another pattern for specifying classes as I think we reached a state where it's unreadable/unmaintainable
due to all these conditions

benoitf (approved):

benoitf (changes requested):
it is missing the unit tests for input

packages/renderer/src/lib/ui/Input.svelte
deboer-tim: Agreed. I was letting it go a bit b/c I didn't know if we still had other requirements, but I think we've got most of the inputs and it's time to simplify.

Any ideas other than extracting a couple variables for the more complicated/merged states? Ok with separate PR?

deboer-tim:
> it is missing the unit tests for input

Doh. Added.

benoitf (approved):

packages/renderer/src/lib/ui/Input.svelte
benoitf: Can be follow-up of course, required changes was for the test

benoitf:
FYI a file is not formatted

Issue #5998, opened 2024-02-14
Improve issue and PR templates to consider testing aspects


odockal:
### Is your enhancement related to a problem? Please describe

When one is creating a feature request, enhancement or epic feature, it would be nice to have also a section that would remind the creator of the issue if the tests are needed and remind the person to create also an issue for the testing part. 

### Describe the solution you'd like

update particular templates (feature_request, enhancement and epic) to consider also acceptance criteria and eventually if issue the test implementation is needed.

### Additional context

it is hard to keep up with all the incoming features and enhancements with respect to quality. We need to get ahead and at least know in advance what is coming in and that authors are aware of qe aspects of the code and also do not forget to provide correct information for functionality to be tested in the form of acceptance criteria.

odockal:
I have created `qe/review` label so our engineering can also put this label onto new enhancements and features so it wont slip of our radar.

PR #5999, opened 2024-02-14
docs: update issue templates to consider testing aspects

odockal:
### What does this PR do?
It updates the issue templates for epic, feature and enhancement with optional section regarding quality aspects of the code. 

#5998 

### How to test this PR?
TBH I do not know how to test this in advance of the merging.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,14 +2,19 @@
...
 repository (or from another issue tracker). -->
 
 ### How to test this PR?
 
 <!-- Please explain steps to reproduce -->
+
+### Do you have unit tests?
+
+<!-- In case of adding new or updating actual functionality
+you need to provide unit tests -->
```
benoitf: to me it's duplicating the previous entry line 13
if there are unit tests I usually add it there

benoitf (commented):
I would probably split issue template and pr template update in two as you might face challenges in the review and maybe one can go faster than the other one

odockal:
> I would probably split issue template and pr template update in two as you might face challenges in the review and maybe one can go faster than the other one

good idea, here it is: https://github.com/containers/podman-desktop/pull/6005

.github/ISSUE_TEMPLATE/epic.yml
```diff
@@ -23,3 +23,10 @@ body:
     attributes:
       label: Additional context
       description: Add any other context or screenshots here.
+
+  - type: textarea
```
deboer-tim: Most epics created lately are QE epics where this wouldn't apply.
The ones development is using are typically only for grouping, e.g. we need to do 4 features in order to support X. Those features are independent and could be delivered over 3 sprints (or never finished) so IMHO it would be better to align QE work directly to the features than to the epic.

odockal: Make sense. thx.

benoitf (commented):
I'm not kind to have a dedicated section b/c if you think of external people that will submit feature requests/enhancement they really don't know the impact of their request

odockal:
@benoitf That is why it is optional (is it, right?), it can stay empty, or user can simply write that he is not sure, it is then information for us to provide that feedback regarding whether tests will be affected or needed. Or what kind of tests is needed.

I see the main benefit for this sections mainly when the rate of new features and changes incoming into repo is hard to follow by QEs, and so this way anyone adding new stuff will also create tests issue (could be individual depending on the case) and so we will move ahead instead of catching up with what developers are planning to work on.

benoitf:
@odockal I think it's more related to how you sort / triage issues in a QE workflow rather than adding that inside the issue template.
Nobody will add something at the moment of creating the issue

deboer-tim:
Yes, that was my hesitation too. The issue template is for opening new issues, when the person will have no idea what the design, implementation, or impact is. Worse, they might be external and hesitate to open when they don't know this info.

This doesn't come until later when we decide to do the work and figure out what/how we will do it. I agree we need to make sure we assess test impact (and docs, etc), but I don't think that's the issue template.

odockal:
>  Nobody will add something at the moment of creating the issue
> Yes, that was my hesitation too. The issue template is for opening new issues, when the person will have no idea what the design, implementation, or impact is. 

I agree with you both, that this is not known, could be empty as well. 

> Worse, they might be external and hesitate to open when they don't know this info.

The description might be also improved. (something like, `add Unknown if you do not know the impact`)

> This doesn't come until later when we decide to do the work and figure out what/how we will do it. I agree we need to make sure we assess test impact (and docs, etc), but I don't think that's the issue template.

This is not bulletproof. I agree. how we make sure we have this covered? 

> I think it's more related to how you sort / triage issues in a QE workflow rather than adding that inside the issue template.

I can and will sort, triage and plan issues I know about that are ahead of us better if I know about them from you, not that I chase PRs and issues and other places we have to track the work and create them myself. Does not make sense and is overhead for us.

benoitf:
> The description might be also improved. (something like, add Unknown if you do not know the impact)
The issue is that all people will leave empty so why add something optional it is always blank at the end. Let's not add it.


> I can and will sort, triage and plan issues I know about that are ahead of us better if I know about them from you, not that I chase PRs and issues and other places we have to track the work and create them myself. Does not make sense and is overhead for us.

either this is a new label to apply on an issue (so it's coming after the issue has been created) or it's a github project board where you can move issues from columns like

'To triage' | 'Required QE' |  'Not Required' | 'Done'

odockal:
> either this is a new label to apply on an issue (so it's coming after the issue has been created) or it's a github project board where you can move issues from columns like
'To triage' | 'Required QE' |  'Not Required' | 'Done'

To me label sounds better since it is tied to the issue and is visible on the issue itself. Board seems less viable solution to me since it has only visibility in the board (less on issue itself).

odockal:
I have created a label `qe/review`. Thus closing this PR. I am still not really sure this will be used.

PR #6003, opened 2024-02-14
docs: drop the word virtual from lima

afbjorklund:
The linux machine can be both physical and virtual.

Each is referred to as a "Lima instance" (not VM).


afbjorklund:
* https://github.com/lima-vm/lima/discussions/2029

benoitf (approved):

PR #6004, opened 2024-02-14
chore: some styling cleanup on carousel

mairin:
### What does this PR do?

super-minor tweaks on carousel... changing to the chevron left / right icons instead of the lessthan / greaterthan, and tweaking color values from zinc (which we don't use anymore) to charcoal values.

### Screenshot / video of UI

Yeh yeh, hard to tell the difference, but check out the arrows, see?

#### BEFORE
image

#### AFTER
image

none, i just like to tweak, sorry

### How to test this PR?

Screen should look like the after and not the before above :)


benoitf (approved):
looks better 👍

deboer-tim (approved):
LGTM

PR #6016, opened 2024-02-15
feat: input component in play kube

deboer-tim:
### What does this PR do?

Use the Input component when playing a Kube YAML.

When I went to test this I was so confused how to select a file I had to look at the code, and realized you're supposed to click on the readonly text input.

This wasn't a great pattern before, but it's less obvious now that we have disabled styling on Input, so I switched this to the more common pattern of having a Browse button to open the file dialog.

### Screenshot / video of UI

Screenshot 2024-02-15 at 7 55 25 AM

Another part of #3234.

### How to test this PR?

Unit tests updated. Go to container or pods page and Play Kubernetes yaml.

### DIFF
--- a/packages/renderer/src/lib/kube/KubePlayYAML.spec.ts
+++ b/packages/renderer/src/lib/kube/KubePlayYAML.spec.ts
@@ -124,7 +124,10 @@ test('error: When pressing the Play button, expect us to show the errors to the 
   // Simulate selecting a file
   const fileInput = screen.getByRole('textbox', { name: 'Kubernetes YAML file' });
   expect(fileInput).toBeInTheDocument();
-  await userEvent.click(fileInput);
+
+  const browseButton = screen.getByRole('button', { name: 'Browse ...' });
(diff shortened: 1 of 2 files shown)


benoitf (commented):
shouldn't we bring a `<FileInput>` component ?

because I think we won't get consistency, so better to have a component when there is a need to select files/folders

b/c I know that for example for e2e test they're expecting some aria-labels etc

WDYT ?

deboer-tim:
Yeah, you're right. It's a different composite vs the other *Inputs, but we use it in enough places we should have a standard component. Would you prefer:

1. Include FileInput as part of this PR.
2. Merge this PR, then extract.
3. PR for FileInput, then rebase this?

benoitf:
I would go for 2 or 3.

deboer-tim:
I'd prefer 2 then, let's get this merged (approval please) and I'll put up a separate PR for FileInput.

benoitf (approved):

PR #6030, opened 2024-02-15
chore: use checkbox component in deploy to kube

deboer-tim:
### What does this PR do?

We are using standard checkboxes in the Deploy to Kubernetes form instead of our styled Checkbox component. This just switches them and updates the test to be more consistent.

Wrapper <div>s added because Checkbox (a composite) doesn't work as well w/o a layout, but this also fixes the problem where text was wrapping below the checkbox when it shouldn't.

### Screenshot / video of UI

Before:

Screenshot 2024-02-15 at 1 02 09 PM

After:

Screenshot 2024-02-15 at 1 03 08 PM

N/A

### How to test this PR?

Unit tests updated, just check UI.

### DIFF
--- a/packages/renderer/src/lib/pod/DeployPodToKube.spec.ts
+++ b/packages/renderer/src/lib/pod/DeployPodToKube.spec.ts
@@ -417,7 +417,7 @@ test('When deploying a pod, restricted security context is added', async () => {
   expect(createButton).toBeEnabled();
 
   // Click restricted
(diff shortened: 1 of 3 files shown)


benoitf (approved):

cdrage (approved):

benoitf (commented):
loks like `name` property is unsupported on Checkbox component

PR #6031, opened 2024-02-15
chore: fix link cursor

deboer-tim:
### What does this PR do?

I can't believe it, but we're missing the correct cursor when you hover over a Link. Fixed, test updated.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Hover over a link and note the cursor change.


benoitf (approved):
I think it's arguable, most native apps don't change the cursor

but for sure we're not consistent on the cursor

cdrage (approved):

PR #6034, opened 2024-02-15
chore: add stale bot

benoitf:
### What does this PR do?

I enabled `debug-only` flag  + a low `operations-per-run` for now with very old stale range to limit the impact on first run

so it should just do traces for now :-)

related https://github.com/containers/podman-desktop/issues/6032

it won't fix as it will just log the actions (debug)

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -47,7 +47,7 @@ jobs:
             This issue has been automatically marked as stale because it has not had
             activity in the last year. It will be closed in 30 days if no further activity occurs. Please
             feel free to leave a comment if you believe the issue is still relevant.
-            Thank you for your contributions!'
+            Thank you for your contributions!
         close-issue-message: >
             This issue has been automatically closed because it has not had any further
             activity in the last 30 days. Thank you for your contributions!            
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yaml
```diff
@@ -0,0 +1,66 @@
...
+        operations-per-run: 10
+        days-before-issue-stale: 500
+        days-before-issue-close: 30
+        stale-issue-label: lifecycle/stale
+        exempt-issue-labels: lifecycle/frozen
+        stale-issue-message: >
+            This issue has been automatically marked as stale because it has not had
+            activity in the last year. It will be closed in 30 days if no further activity occurs. Please
+            feel free to leave a comment if you believe the issue is still relevant.
+            Thank you for your contributions!'
```
deboer-tim: Unnecessary '

deboer-tim (approved):
LGTM. Stale # of days doesn't match message, but we're debugging now anyway.

.github/workflows/stale.yaml
benoitf:
```suggestion
            Thank you for your contributions!
```

PR #6046, opened 2024-02-16
feat: add Red Hat Container Registry to default registries

vrothberg:
### What does this PR do?

The Red Hat Container Registry (`registry.redhat.io`) is widely used in the ecosystem.  To make it easier for users to access and explore the registry, add it to the list of default registries.  Since the registries of other companies (e.g., Docker, Google, GitHub) are listed as well, I do not perceive this change as forcing Red Hat products onto users but as an UX improvement.

The main motivating driver behind this change was the immediate need of the subscription/sso extension: showing the registry without the logo is undesireable.  However, while re-reading throught the conversation I felt that we may just add the registry by default and make it easier for all users.

[1] https://github.com/redhat-developer/podman-desktop-redhat-account-ext/issues/38

### Screenshot / video of UI

Default list:
Screenshot 2024-02-16 at 13 40 21

After using the subscription/SSO extension:
Screenshot 2024-02-16 at 13 40 12

https://github.com/redhat-developer/podman-desktop-redhat-account-ext/issues/38

### How to test this PR?

Go to the "Registries" menu in the settings.


benoitf:
nobody commented yet but I'll comment

I think this should be kept in a separate Red Hat registries extensions. (as you might want other Red Hat registries like registry.access.redhat.com)

I also think that Google registry should be removed from the extension and be part of a Google extension (if someone want to contribute to) 

Only keep Quay, DockerHub and GitHub

deboer-tim:
+1 to @benoitf comments. Adding a suggested registry is minor (it just gives an icon and saves typing a url), it is exactly the kind of thing that should be provided by extensions. Having a few of the more common/open ones built-in is good but we don't want to propagate that all of them should be contributed directly here.

vrothberg:
I understand the technical arguments but I also think that all of these registries are very commonly used.  It seems like a too simple kind of function to me for having to install an extension.

I guess where I am going is that I (as a user) would expect/prefer this to be my default/vanilla experience rather than having to install too many extensions.

benoitf:
It's not only technical argument. All people want to be there by 'default'. And it's true for all the part of the project. 

The main idea is to reduce the default scope and have contribution from external extensions. 
But in the other way, finding, installing registries etc should be smoother.

Adding custom extension that has the 'registries' category could be promoted in the registries screen (so if you're interested in google registries it'll bring the google stuff, the same for aws, the same for Red Hat, etc.)

vrothberg:
> It's not only technical argument. All people want to be there by 'default'. And it's true for all the part of the project.

I sympathize a lot with the problem; I felt it all too well.  Yet, I don't think that this registry is not just any.

> The main idea is to reduce the default scope and have contribution from external extensions. But in the other way, finding, installing registries etc should be smoother.

How can we balance between the two?  I feel that this registry should be part of the default experience.

I am not married to this PR but we need to improve the UI for the current project.  Time is pressing as well, so we need to progress.

jeffmaury:
Could be done by extensions that requires this registry ?

vrothberg:
> Could be done by extensions that requires this registry ?

Possibly.  I have a hard time understanding why writing (and maintaining) a new extension is preferable over this PR. It puts the Red Hat registry on the same level as the one of Docker and Google and has zero maintenance cost.

jeffmaury:
> > Could be done by extensions that requires this registry ?
> 
> Possibly. I have a hard time understanding why writing (and maintaining) a new extension is preferable over this PR. It puts the Red Hat registry on the same level as the one of Docker and Google and has zero maintenance cost.

That would not require another extension

vrothberg:
> > Possibly. I have a hard time understanding why writing (and maintaining) a new extension is preferable over this PR. It puts the Red Hat registry on the same level as the one of Docker and Google and has zero maintenance cost.
> 
> That would not require another extension

Can you elaborate on that?

I think there are two decision:
1) Should registry.redhat.io be displayed by default?
2) If not, how.

I believe that 1) is an important intentional decision.

jeffmaury:
> > > Possibly. I have a hard time understanding why writing (and maintaining) a new extension is preferable over this PR. It puts the Red Hat registry on the same level as the one of Docker and Google and has zero maintenance cost.
> > 
> > 
> > That would not require another extension
> 
> Can you elaborate on that?
> 
> I think there are two decision:
> 
>     1. Should registry.redhat.io be displayed by default?
> 
>     2. If not, how.
> 
> 
> I believe that 1) is an important intentional decision.

If extension A needs this registry in the workflow it implements, it will register it. If another one needs it, then it should also register it as well.

vrothberg:
@benoitf @jeffmaury shall I close and somebody else takes over?

benoitf:
yes, let's close this PR

PR #6048, opened 2024-02-16
feat: adding secret store

axel7083:
### What does this PR do?

This PR is adding a secret store, which will add a `secrets.json` file next to the `settings.json` file in a similar format but with encrypted _key/value_.

> This PR is not adding the api support, nor usage on the added classes (trying to keep things clear)

Some references links 
- [safe-storage in Electron](https://www.electronjs.org/docs/latest/api/safe-storage)
- [storage interface in vscode](https://github.com/microsoft/vscode/blob/05e5d4bb073221eb49c4f161d55ea4f48ff88958/src/vs/base/parts/storage/common/storage.ts#L77)
- [vscode encryptionMainService](https://github.com/microsoft/vscode/blob/main/src/vs/platform/encryption/electron-main/encryptionMainService.ts)

Part of https://github.com/containers/podman-desktop/issues/6044

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/storage/encryption/encryptionMainService.ts
+++ b/packages/main/src/plugin/storage/encryption/encryptionMainService.ts
@@ -1,0 +1,97 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


benoitf (changes requested):
an overall comment, it looks like a rapid copy/paste

but it should be reduced to the thing that we need

the names are not following the existing patterns
there are tons of unexpected interfaces

axel7083:
> an overall comment, it looks like a rapid copy/paste

I agree that the `EncryptionMainService` is mainly a copy past, as the implementation is already made. 

> but it should be reduced to the thing that we need

In order to reduce and scope I took some inspiration on the `ConfigurationRegistry`

I admit that the `getters` for specific type are a lot, but I was liking them, even if they are not used in the ConfigurationRegistry, but if we want to keep a closer interface than it has, we can only support strings

> the names are not following the existing patterns

Yes I just rename the SecretStorage to SecretRegistry to follow our pattern

axel7083:
Closing as requires more information on the implementation to proceed

PR #6060, opened 2024-02-18
feat: consistent close button

### LINKED ISSUE #2352: Nit: Should we use xmark icon?

Follow up/remaining issue from issue 2241: should we use the xmark for close buttons vs the current times?

Original comment:
- icon on the top right looks a big larger than expected one (seems it's all over Podman Desktop but I think xmark (or circle-xmark on hovering as we're using a lot now circle to provide feedback) icon is better suited than times icon) https://fontawesome.com/icons/xmark?f=classic&s=solid https://fontawesome.com/icons/circle-xmark?f=classic&s=regular vs the current one https://fontawesome.com/v4/icon/times

We'll just make a decision and change or not, ;-)

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

We have 9 close buttons that I could find in Podman Desktop. They were at least 3 different colors, 6 of them changed color when you hover, and one hover changed the hover background too. Except for the one, the lack of/inconsistent hover indication was standing out to me compared with our Links and Buttons.

This provides a common CloseButton component in a medium-dark gray. On hover, it goes white and has the same bg as a Link (but slightly larger since the icon is small). The button accepts either an href or an on:click action to support our different uses.

I've made a call here that all close buttons look the same, instead of leaving those on dialogs/modals white. IMHO this consistency is better, allows the hover indication to be more clear, and we don't need to highlight the Close button in any of these places.

Although we might normally divide up the component + usage in two PRs, the amount of change here wasn't big and I thought it's important to see it in all of the contexts, so I've included both and can split up if requested.

### Screenshot / video of UI

Before:

https://github.com/containers/podman-desktop/assets/19958075/1371cdea-c40d-47f6-8d1f-f195604caeae

After:

https://github.com/containers/podman-desktop/assets/19958075/2904dcb4-5e92-4f0e-b3d8-a7c17591c13b

Fixes #2352.

### How to test this PR?

Unit tests updated. Try modal dialogs, message boxes, and form/detail pages.

### DIFF
--- a/packages/renderer/src/lib/ui/FormPage.svelte
+++ b/packages/renderer/src/lib/ui/FormPage.svelte
@@ -3,6 +3,7 @@ import { lastPage, currentPage } from '../../stores/breadcrumb';
 import { router } from 'tinro';
 import Link from './Link.svelte';
 import LinearProgress from './LinearProgress.svelte';
+import CloseButton from '/@/lib/ui/CloseButton.svelte';
 
 export let title: string;
 export let showBreadcrumb = true;
@@ -31,9 +32,7 @@ function handleKeydown(e: KeyboardEvent) {
             >{$lastPage.name}</Link>
           <div class="mx-2">&gt;</div>
           <div class="grow font-extralight" aria-label="name">{$currentPage.name}</div>
-          <a href="{$lastPage.path}" title="Close" class="justify-self-end text-gray-900">
-            <i class="fas fa-times" aria-hidden="true"></i>
-          </a>
+          <CloseButton href="{$lastPage.path}" class="justify-self-end" />
         </div>
       {/if}
       <div class="flex flex-row items-center pt-1">
@@ -50,9 +49,7 @@ function handleKeydown(e: KeyboardEvent) {
             </div>
           {/if}
           {#if !showBreadcrumb}
-            <a href="{$lastPage.path}" title="Close" class="text-gray-900">
-              <i class="fas fa-times" aria-hidden="true"></i>
-            </a>
+            <CloseButton href="{$lastPage.path}" />
           {/if}
         </div>
       </div>
--- a/packages/renderer/src/lib/dialogs/MessageBox.svelte
+++ b/packages/renderer/src/lib/dialogs/MessageBox.svelte
@@ -7,6 +7,7 @@ import type { MessageBoxOptions } from './messagebox-input';
 import Button from '../ui/Button.svelte';
 import type { ButtonType } from '../ui/Button';
 import { tabWithinParent } from './dialog-utils';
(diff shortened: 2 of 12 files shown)


axel7083:
On certain place we are using the `faXmark` icon, it is also one we want to replace ?

https://github.com/containers/podman-desktop/blob/e64ee5ec39885f0e4f11e6614a31c33cf68d2f36/packages/renderer/src/lib/dashboard/NotificationCardItem.svelte#L35

https://github.com/containers/podman-desktop/assets/42176370/0a360621-ddc5-41c0-87e1-407091f77a8a

https://github.com/containers/podman-desktop/blob/a2c31dd17641b433e64ae6d1acf35cae91b40f47/packages/renderer/src/lib/ui/AuditMessageBox.svelte#L22

https://github.com/containers/podman-desktop/blob/ab0bd5062b0f3c45dd7058d7dabf5cbf4edde5ef/packages/renderer/src/lib/dialogs/CustomPick.svelte#L241

https://github.com/containers/podman-desktop/blob/5baebead51e3baf25b261c1c76ad814d5c125823/packages/renderer/src/lib/image/RunImage.svelte#L790

image

https://github.com/containers/podman-desktop/blob/1c3e61842413d6b45690cd6ce2f652e624fc33d2/packages/renderer/src/lib/ui/Input.svelte#L82

https://github.com/containers/podman-desktop/blob/a2c31dd17641b433e64ae6d1acf35cae91b40f47/packages/renderer/src/lib/window-control-buttons/LinuxControlButton.svelte#L22

deboer-tim:
Most of these are clear buttons, not close dialog/window. I don't think we necessarily have to use the same icon for both (and definitely wasn't trying to unify here).

@mairin do you have a preference for faTimes vs faXmark for either close or clear buttons? Would also be nice to get your +1 on the color difference (thread paragraph in description).

packages/renderer/src/lib/ui/FormPage.svelte
```diff
@@ -50,9 +49,7 @@ function handleKeydown(e: KeyboardEvent) {
             </div>
           {/if}
           {#if !showBreadcrumb}
-            <a href="{$lastPage.path}" title="Close" class="text-gray-900">
-              <i class="fas fa-times" aria-hidden="true"></i>
-            </a>
+            <CloseButton internalRef="{$lastPage.path}" />
```
lstocchi: internalRef?

deboer-tim: Good catch - fixed and rebased.

axel7083 (approved):
LGTM

lstocchi (approved):
LGTM

PR #6068, opened 2024-02-19
feat: improve Podman-Desktop update alert

### LINKED ISSUE #5754: UX: Improvements to new version alerts

We would like to improve the user experience of receiving alerts about new versions of Podman Desktop becoming available.

Here is what we discussed for this feature at the 24 Jan UX check in call:

*Potential places to notify*
- Menu check for updates for macOS users
- Dialog boxes 
- Move from status bar to landing page (like other notifications ‘Podman needs to be set up’

*How to notify:*
- If user is just starting up Podman Desktop from fresh, have a pop up dialog
- If user already has Podman Desktop open and is using:
  - Notification area pop up, click for more info 
- In both cases: dashboard box and pop up dialog gives options:

```
Update available now
Update now | Update later | Update never
```
- Add a preference to settings > preferences to undo the "never" in case you picked it accidentally or have a change of heart

_No response_

### Additional context

_No response_

### LINKED ISSUE #5755: Improvements to new version alerts

We would like to improve the user experience of receiving alerts about new versions of Podman Desktop becoming available.

See design issue here:
https://github.com/containers/podman-desktop/issues/5754

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

According to https://github.com/containers/podman-desktop/issues/5754.

When an update is available
- [x] If user is just starting up Podman Desktop from fresh, have a pop up dialog
- [x] if an update occurs a popup is spawned

A new preference section called `Update` has a property `reminer` with two value `startup` or `never`.

#### starting `podman-desktop`

image

#### Clicking on status bar item when update available (status bar item)

image

Fixes https://github.com/containers/podman-desktop/issues/5754

### How to test this PR?

**Testing manually**

(1) change the version to `1.6.5` in the `package.json` at the root
(2) run `yarn compile:current`
(3) go to `dist` folder and start the portable executable

(Thx to @benoitf for this method)

### DIFF
--- a/packages/main/src/plugin/updater.ts
+++ b/packages/main/src/plugin/updater.ts
@@ -81,7 +81,7 @@ export class Updater {
       UPDATER_UPDATE_AVAILABLE_ICON,
       true,
       'update',
-      undefined,
+      ['status-bar-entry'],
       true,
     );
   }
@@ -108,7 +108,7 @@ export class Updater {
     });
 
     // Update will create a standard "autoUpdater" dialog / update process
-    this.commandRegistry.registerCommand('update', async () => {
+    this.commandRegistry.registerCommand('update', async (context?: 'startup' | 'status-bar-entry') => {
       if (this.#updateAlreadyDownloaded) {
         const result = await this.messageBox.showMessageBox({
           type: 'info',
@@ -136,11 +136,19 @@ export class Updater {
       // Get the version of the update
       const updateVersion = this.#nextVersion ?? '';
 
+      let buttons: string[];
+      if (context === 'startup') {
+        buttons = ['Update now', 'Remind me later', 'Do not show again'];
+      } else {
+        buttons = ['Update now', 'Cancel'];
+      }
+
       const result = await this.messageBox.showMessageBox({
         type: 'info',
         title: 'Update Available now',
         message: `A new version ${updateVersion} of Podman Desktop is available. Do you want to update your current version ${this.#currentVersion}?`,
-        buttons: ['Update now', 'Update later', 'Update never'],
+        buttons: buttons,
+        cancelId: 1,
       });
       if (result.response === 2) {
         this.updateConfigurationValue('never');
@@ -176,7 +184,7 @@ export class Updater {
 
     this.updateAvailableEntry();
     if (this.getConfigurationValue() === 'startup') {
-      this.commandRegistry.executeCommand('update').catch((err: unknown) => {
+      this.commandRegistry.executeCommand('update', 'startup').catch((err: unknown) => {
         console.error('Something went wrong while executing update command', err);
       });
     }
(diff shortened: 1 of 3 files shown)


benoitf (commented):
there is a refactoring and a new addition in the same PR so it's hard to track what is new, what is refactoring

I think having refactored first (extract to a separate class first) and then adding feature would have helped the review

axel7083:
> there is a refactoring and a new addition in the same PR so it's hard to track what is new, what is refactoring
> 
> I think having refactored first (extract to a separate class first) and then adding feature would have helped the review

I made a didicated PR for the refactor in https://github.com/containers/podman-desktop/pull/6159 - keeping this in draft for now

packages/main/src/plugin/updater.ts
```diff
@@ -166,6 +169,56 @@ export class Updater {
...
+   */
+  private registerConfiguration(): void {
+    this.configurationRegistry.registerConfigurations([
+      {
+        id: 'preferences.update',
+        title: 'Update',
+        type: 'object',
+        properties: {
+          ['preferences.update.reminder']: {
+            description: 'Update reminder',
```
benoitf: I think we need to be more verbose in the description
just a suggestion, can be any other wording
`Configure whether you receive update reminders when starting Podman Desktop`

packages/main/src/plugin/updater.ts
```diff
@@ -166,6 +169,56 @@ export class Updater {
...
+      },
+    ]);
+  }
+
+  /**
+   * Retrieves the value of the update reminder configuration.
+   * @returns The value of the update reminder configuration.
+   */
+  private getConfigurationValue(): 'startup' | 'never' {
+    const value = this.configurationRegistry.getConfiguration('preferences').get('update.reminder');
```
benoitf: as get is a generic type you might try `get<'startup' | 'never'>('update.reminder')` to avoid to handle an invalid value

benoitf (commented):
when starting Podman Desktop I have the message

image


while if I click on the version icon I have

image

I like the fact to have the version number of the new version to be displayed
could we have the same sentence ?


about the buttons, I think I see too many 'Update' choice
I'm not sure about the 'Update Later' and 'Update never' (I think the never is looking odd as I can still click on the version to update)

should it be around 'Skip' for 'Update Later'  and 'notifications' (or any other wording)

About the notify later, should it still prompt on every startup or should it be in case of a new version is released.

axel7083:
@benoitf I took some liberty with the original mockup https://github.com/containers/podman-desktop/issues/5754 

# starting `podman-desktop`

image

# Clicking on `v1.6.5`  (status bar item)

image

benoitf (approved):
I find it way better now 🎉 

it's not related to this PR but maybe we should change our wording when there is a new version as the last version that we see at the end of the sentence is the current version

so reading it quickly I have the feeling that it's asking to update to the version number that we have

If others share the same concern, we could do follow-up, if I'm the only one then we can just ignore :-)

benoitf:
before merging, you'll need to update the body of the PR with accurate screenshot/video

PR #6101, opened 2024-02-22
Make informers used for kubernetes contexts send resources data

### LINKED ISSUE #6100: informers used for kubernetes contexts should send information about resources, not only counts

For now, the ContextState is giving `podsCount` and `deploymentsCount` information.

 It should:

- send objects data, instead of count
- send a more generic structure, to be able to add resources to the ContextState easily:

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

- use index signature for kubernetes context states, to be able to add more resources easily
- send complete objects data, not only count
- add a new store, giving the state of the current context (updated as soon as the current context is changed) (also done as part of #6102)
- add new stores to get filtered deployments in the current context (the code will need to be more generic to create stores for different resources)
- use of these new stores in Deployments page

Fixes a bug where resources are still visible in the Deployments list when the context is not reachable:
When an informer is disconnected then reconnected, is does not get new 'add' events for the resources it is watching (bug or feature?), so we cannot reset the resources in our cache, as it won't be populated again after reconnection.
Instead, the resources are not sent anymore to the frontend when the context is not reachable (but not cleared from the cache). Another possibility would be to send the resources in any case, so the frontend could display the resources as 'latest seen', or filter itself, based on the `reachable` state. @deboer-tim WDYT?

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/9973512/5c88838b-73a0-46c1-a23c-2d7cd2ef0c97

Fixes #6100 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -16,6 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import { AppsV1Api, CoreV1Api, KubeConfig, makeInformer } from '@kubernetes/client-node';
 import type {
   Informer,
   KubernetesObject,
@@ -24,7 +25,6 @@ import type {
   V1Deployment,
   V1Pod,
 } from '@kubernetes/client-node';
-import { AppsV1Api, CoreV1Api, KubeConfig, makeInformer } from '@kubernetes/client-node';
 import type { KubeContext } from './kubernetes-context.js';
 import type { ApiSenderType } from './api.js';
 
@@ -103,8 +103,22 @@ class ContextsStates {
     return this.informers.keys();
   }
 
+  noResource() {
+    return {
+      pods: [],
+      deployments: [],
+    };
+  }
+
   getPublished(): Map<string, ContextState> {
-    return this.published;
+    // We don't want to reset the resources when the informer is disconnected,
+    // because the new informer in the same resource won't send 'add' events again after being reconnected (due to some cache in the kubernetes library?)
+    // So, here, we need to send no resources if not reachable, but keep the resources untouched.
+    const result = new Map<string, ContextState>();
+    this.published.forEach((val, key) =>
+      result.set(key, val.reachable ? val : { ...val, resources: this.noResource() }),
+    );
+    return result;
   }
 
   safeSetState(name: string, update: (previous: ContextState) => void) {
@@ -112,10 +126,7 @@ class ContextsStates {
       this.published.set(name, {
         error: undefined,
         reachable: false,
-        resources: {
-          pods: [],
-          deployments: [],
-        },
+        resources: this.noResource(),
       });
     }
     const val = this.published.get(name);
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -45,7 +45,7 @@ export interface ContextState {
 type ResourceName = 'pods' | 'deployments';
 
 export type ContextStateResources = {
-  [resourceName in ResourceName]: number;
+  [resourceName in ResourceName]: KubernetesObject[];
 };
 
 interface CreateInformerOptions<T> {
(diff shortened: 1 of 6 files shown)


axel7083:
Good idea to split in resources ! You can probably rebase as https://github.com/containers/podman-desktop/pull/6099 has been merged!

deboer-tim:
At some point I think we'll want the actual list usable in the UI, e.g. Settings > Kubernetes > deployment counts and Deployments page coming from same data. What do you think about resources containing arrays and preferences can do something like `.resources.pods.length`?

feloy:
> At some point I think we'll want the actual list usable in the UI, e.g. Settings > Kubernetes > deployment counts and Deployments page coming from same data. What do you think about resources containing arrays and preferences can do something like `.resources.pods.length`?

Yes, I have sent a second commit to send complete objects as you say

deboer-tim:
I've spent a while today trying to get the 'old' informers to reconnect correctly, and although I'm getting close I'm finding that I'm now just trying to replicate problems already solved in this informer (e.g. reconnect timeout, backoff).

I then saw @axel's PR with the derived informer, tried the same thing and got a working Deployments page, and then saw it's exactly what you've done here. I know I'm being fickle, but I think I should pause the fix I was creating and we try to get this path working over the next few days.

I think this PR should be split to keep things small and focussed:
- changing pods & deployments to arrays (IMHO could be merged immediately)
- adding derived stores
with future PRs to add ingress/routes/services using the same pattern.

The main risk of this approach would be the number of informers if the user has many contexts, but maybe we can only start some informers for current context? I also ran into one unrelated nit: the backoff grows fairly rapidly so it took a while to auto-connect after I started a cluster. Likewise I think we could special-case the current context and back off a little slower. I will open some issues.

feloy:
> I've spent a while today trying to get the 'old' informers to reconnect correctly, and although I'm getting close I'm finding that I'm now just trying to replicate problems already solved in this informer (e.g. reconnect timeout, backoff).
> 
> I then saw @axel's PR with the derived informer, tried the same thing and got a working Deployments page, and then saw it's exactly what you've done here. I know I'm being fickle, but I think I should pause the fix I was creating and we try to get this path working over the next few days.
> 
> I think this PR should be split to keep things small and focussed:
> 
> * changing pods & deployments to arrays (IMHO could be merged immediately)
> * adding derived stores
>   with future PRs to add ingress/routes/services using the same pattern.
> 

Yes, I created this PR with several commits to be able to test the result easily, but I have isolated commits in it, that I will cherry-pick into dedicated PRs.

> The main risk of this approach would be the number of informers if the user has many contexts, but maybe we can only start some informers for current context? 


> I also ran into one unrelated nit: the backoff grows fairly rapidly so it took a while to auto-connect after I started a cluster. Likewise I think we could special-case the current context and back off a little slower. I will open some issues.

Yes, I agree that this is annoying. Another possibility I'm thinking of would be to provide a 'reconnect' button (perhaps in the State added by Axel in #6102), to retry a connection immediately (and reset the backoff?). Also, a little spinner the time of the connection attempt would be helpful for the user to know if it is still waiting for the next retry or not.

deboer-tim:
Tested, working very well. 🚀

feloy:
I have created #6116 with the first 2 commits (changing pods & deployments to arrays)

deboer-tim:
#6116 is integrated, so if this was rebased the only bits left would be the deployment store and update to the list. FWIW I stole those changes for an experiment, adding services, routes, ingress in #6118. It all works 👍🏼. See the PR description for more info.

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -193,8 +211,12 @@ export class ContextsManager {
     return this.createInformer<V1Pod>(kc, context, path, listFn, {
       timer: this.podTimer,
       backoff: new Backoff(1000, 60_000, 300),
-      onAdd: _obj => this.setStateAndDispatch(context.name, state => state.podsCount++),
-      onDelete: _obj => this.setStateAndDispatch(context.name, state => state.podsCount--),
+      onAdd: obj => this.setStateAndDispatch(context.name, state => state.resources.pods.push(obj)),
```
feloy, line 214: We will also need to listen on `update` events, go get updated values of resources

feloy:
replaced by #6131

PR #6112, opened 2024-02-22
fix(navigation): adding pod details default tab

### LINKED ISSUE #6071: bug(extension-api): navigate to pod does not redirect to default tab
### Bug description

When we use the navigation.navigateToPod we are redirect to `/pods/:kind/:name/:engine/` however this page is blank because we should be redirect to `.../logs` or else just like the ContainerDetails component is doing

https://github.com/containers/podman-desktop/blob/008cc0f2da72ee093e09ca31b7e79b5c317488df/packages/renderer/src/lib/container/ContainerDetails.svelte#L53-L57

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

use navigation api for pods

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

This PR is adding a default route to the Pod Details page. When no path is provided, it will redirect to `/logs` in a similar manner than the ContainerDetails is doing (see issue for details)

Fixes https://github.com/containers/podman-desktop/issues/6071

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/container/ContainerList.svelte
+++ b/packages/renderer/src/lib/container/ContainerList.svelte
@@ -40,8 +40,6 @@ import StateChange from '../ui/StateChange.svelte';
 import SolidPodIcon from '../images/SolidPodIcon.svelte';
 import ProviderInfo from '../ui/ProviderInfo.svelte';
 import { findMatchInLeaves } from '../../stores/search-util';
-import { handleNavigation } from '/@/navigation';
-import { NavigationPage } from '../../../../main/src/plugin/navigation/navigation-page';
 
 const containerUtils = new ContainerUtils();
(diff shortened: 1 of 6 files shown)


benoitf (approved):

deboer-tim (approved):
Any chance we could move the encoding to handleNavigation() so that clients don't need to worry about it?

Looks like the same problem exists for volumes and images, could you open another issue?

axel7083:
> Any chance we could move the encoding to handleNavigation() so that clients don't need to worry about it?
> 
> Looks like the same problem exists for volumes and images, could you open another issue?

I opened https://github.com/containers/podman-desktop/issues/6154

PR #6116, opened 2024-02-22
Add complete object from contexts informers

feloy:
### What does this PR do?

- use index signature for kubernetes context states, to be able to add more resources easily
- send complete objects data, not only count

Part of #6100 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
@@ -108,13 +108,13 @@ async function handleDeleteContext(contextName: string) {
                     <div class="text-center">
                       <div class="font-bold text-[9px] text-gray-800">PODS</div>
                       <div class="text-[16px] text-white" aria-label="context-pods-count">
-                        {$kubernetesContextsState.get(context.name)?.podsCount}
+                        {$kubernetesContextsState.get(context.name)?.resources.pods}
                       </div>
                     </div>
                     <div class="text-center">
                       <div class="font-bold text-[9px] text-gray-800">DEPLOYMENTS</div>
                       <div class="text-[16px] text-white" aria-label="context-deployments-count">
-                        {$kubernetesContextsState.get(context.name)?.deploymentsCount}
+                        {$kubernetesContextsState.get(context.name)?.resources.deployments}
                       </div>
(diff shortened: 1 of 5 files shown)


deboer-tim (changes requested):
Looks like you need to fix up KubernetesCurrentContextConnectionBadge.spec.ts now too, but LGTM otherwise.

deboer-tim (approved):

deboer-tim (approved):
LGTM

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -108,13 +108,13 @@ async function handleDeleteContext(contextName: string) {
...
                       <div class="text-[16px] text-white" aria-label="context-pods-count">
-                        {$kubernetesContextsState.get(context.name)?.podsCount}
+                        {$kubernetesContextsState.get(context.name)?.resources.pods.length}
                       </div>
                     </div>
                     <div class="text-center">
                       <div class="font-bold text-[9px] text-gray-800">DEPLOYMENTS</div>
                       <div class="text-[16px] text-white" aria-label="context-deployments-count">
-                        {$kubernetesContextsState.get(context.name)?.deploymentsCount}
+                        {$kubernetesContextsState.get(context.name)?.resources.deployments.length}
```
dgolovin, line 117: The change should work, but is it looks like we are storing array of objects on client side and then just use length of it? In real cluster there might be hundreds if not thousands of objects and that's might be a problem. I have the same issue with tekton tools at some point when it turned out there are thousands builds were recorded and I was trying to load all of them in UI process >8) It should store Partial<T> in it with info that we might use in UI. Definitely not all of it.

feloy:
> LGTM

Thanks. Sorry, I misclicked on the button to ask your review again

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: Yes you are right. Only the info necessary for displaying the list (in the Deployment/... page) should be passed into the state. Thanks for the advice, I'll make the change

cdrage (approved):
Tested, works, LGTM!

Issue #6114, opened 2024-02-22
Reduce informer backoff for current context


deboer-tim:
### Is your enhancement related to a problem? Please describe

The Kubernetes informer backoff works well, but it drops off fairly fast. This is completely fine for most contexts, but the current context is a little special: it's the one I'm most likely to be interested in, and (e.g.) if I start a new kind or minikube cluster I expect it to connect fairly fast.

### Describe the solution you'd like

Add some preference/weight to the current context, so that it backs off not as fast or not as far as all other contexts.

deboer-tim:
Still valid.

feloy:
Fixed by #9392, #9411, #9443

Issue #6094, opened 2024-02-22
Lazy-loading Kubernetes informers


deboer-tim:
### Is your enhancement related to a problem? Please describe

With the 'old' informers we are creating individual informers on the current context based on whichever pages you navigate to: deployments, ingress, routes, services. This means a max of 4 informers, which stay connected until you switch the context.

With the new/context informers, we create a pod and deployment informer for every context. We've talked about using these informers for services/ingress/routes too, but with the current code this would mean 5 informers x N contexts, which could mean a large number. Most users will not want us to constantly pull all this information from every context, and it could cause performance issues, especially if we support more objects in the future.

### Describe the solution you'd like

The current behaviour should still be an option, but the default should be much more lazy-loading - e.g. only starting the 2-3 necessary informers when you go to Settings > Kubernetes and stopping them after, or only starting each of the others as you go to those pages, and stopping them after

We will likely want a delay on stopping, so that if you switch the current context and switch back, or reopen one of the pages within a short period, we are still connected. Likewise we may want to pre-start some informers in some cases to avoid/reduce connection delays.

deboer-tim:
Still valid.

PR #6118, opened 2024-02-22
draft: full kube informers integrated

### LINKED ISSUE #5311: Informer stores improvements

### LINKED ISSUE #6113: New informer stores for services, routes, ingress

Following the pattern from #6101, we should add support for routes, ingress, and services.

The newer informer pattern has better support for connecting and reconnecting when there is a failure, plus backoff/jitter. Plus, there's the future potential for multi-context or switching back and forth between contexts w/o having to reconnect.

These new objects are not used by the Settings > Kubernetes page, so we need some way to not start all informers for all contexts. Starting these only for the current context would be fine.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

After seeing the filtered stores in #6101 I thought I'd see how hard it would be to do the same for ingress/routes/services. This is a quick pass at:
- Adding informers for the three missing types to main/kubernetes-context-state.ts. This is mostly copy/paste from kubernetes-client.ts, only missing one issue on line 331).
- Adding filtered stores to renderer/kubernetes-contexts-state.ts.
- Switching all the pages over to use these stores.

These changes would clearly have to be broken up into multiple PRs and tests updated, but we're not affecting existing UI until #3. After this, the other stores could be removed, as well as informers in kubernetes-client.ts.

Compared to the existing informers, this would fix the existing issues in #5311 with initial connection/reconnection, and would add backoff, but there are two issues we'd have to resolve:
- #6114 - reduce backoff for the current context, it's a bit slow.
- #6094 - we're probably creating too many informers; only loading these new ones for the current context would be enough for now.

### Screenshot / video of UI

N/A

Fixes #5311. Fixes #6113.

### How to test this PR?

Draft, but works if anyone wants to try.

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -23,15 +23,30 @@ import type {
   ObjectCache,
   V1Deployment,
   V1Pod,
+  V1Service,
+  V1Ingress,
+  KubernetesListObject,
 } from '@kubernetes/client-node';
-import { AppsV1Api, CoreV1Api, KubeConfig, makeInformer } from '@kubernetes/client-node';
+import {
+  AppsV1Api,
+  CoreV1Api,
+  NetworkingV1Api,
+  CustomObjectsApi,
+  KubeConfig,
+  makeInformer,
+} from '@kubernetes/client-node';
 import type { KubeContext } from './kubernetes-context.js';
 import type { ApiSenderType } from './api.js';
+import type { V1Route } from './api/openshift-types.js';
+import type { IncomingMessage } from 'node:http';
 
 // ContextInternalState stores informers for a kube context
 interface ContextInternalState {
   podInformer?: Informer<V1Pod> & ObjectCache<V1Pod>;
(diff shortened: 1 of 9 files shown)


deboer-tim:
Everything here has now been proposed in a non-draft PR, so closing this.

PR #6122, opened 2024-02-23
docs: fix instructions to edit registries.conf

eidottermihi:
### What does this PR do?

Editing /etc/containers/registries.conf requires root. `vim` is also not installed in podman default machine (only `vi`)


benoitf (approved):

benoitf:
thanks @eidottermihi

PR #6125, opened 2024-02-25
chore: improve doc nav styling

deboer-tim:
### What does this PR do?

Makes the main doc navigation elements (sections + root pages) bold, and reduces the size of child elements to make the main elements stand out better and child elements wrap less.

### Screenshot / video of UI

Before:

Screenshot 2024-02-24 at 9 59 08 AM

After:

Screenshot 2024-02-24 at 9 57 04 AM

Fixes styling part of #5338.

### How to test this PR?

`yarn website:dev`

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -149,6 +149,19 @@ li.footer__item a svg {
   display: none;
 }
 
+/* main nav items bold */
+.theme-doc-sidebar-item-link-level-1 {
+  font-weight: bold;
+}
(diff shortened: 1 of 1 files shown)


benoitf:
I think that to reach the goal of the issue, probably we should add the custom components/doc/sidebar/item.tsx component

Then, from there we could also use the tailwind classes

cdrage (approved):
LGTM as this really improves the styling right now.

I do agree though that in the future we should do a separate css file / do something to fix it.

Right now IMHO we should get this in now, as even with my eyes it's difficult to read / differentiate between the menus.

PR #6131, opened 2024-02-26
feat: send contexts status on different channels

### LINKED ISSUE #6123: Proposal: optimized informers

Optimize the listening of different resources on different Kubernetes contexts, using informers
#5311 #6094 #6100 #6113 #6114 

## Current status 

On the backend side, the `ContextsManager` (in `packages/main/src/plugin/kubernetes-context-state.ts`) starts informers for every managed resources (pods, deployemnts for the moment) on every Kubernetes contexts (in kubeconfig file) and caches the list of resources locally. Once connected, the manager listens `add`, `delete` and `update` events to update the cached resources accordingly.

In case of connection error, retries are done using a backoff pattern on each informer (for the moment using the same pattern for each informer).

On every  change, the complete set of cached resources (all resources of all kinds for all contexts) are sent to the frontend.

On the frontend side, a single store asks for the content cache when it starts, and subscribes for receiving updates every time the cache is updated.

Several `derived` stores on the first store give information about specific information on the current context.

## Proposed evolution

The `ContextsManager` continues to collect all the resources for all the contexts.

Several channels replace the current single channel:

1. status and number of a selected list of resources (pods, deployments) for all contexts (including the current one):

```
Map {
  'context1' => { reachable: true, resources: { pods: 2, deployments: 1 } },
  'context2' => { reachable: false, error: 'err', resources: { pods: 0, deployments: 0 } },
}
```

2. Status and number of a selected list of resources (pods, deployments) for current context:

```
{ reachable: false, error: 'err', resources: { pods: 0, deployments: 0 } }
{ reachable: true, resources: { pods: 2, deployments: 1 } }
```

3. Pods in the current context:

```
[ { /* pod1 */}, { /* pod2 */ } ] 
```

4. Deploymenets in the current context:

```
[ { /* deployment1 */}, { /* deployment2 */ } ] 
```

5. ..n Other resources in the current context

```
[ { /* resourceX1 */}, { /* resourceX2 */ } ] 
```

On the frontend side, a store exists for each channel. Svelte stores are created when first used, and discarded when the last user disconnects. With the same pattern, the stores can register to the ContextsManager, to indicate they need updates on the resources data:
- informers for resources other than pods and deployments are not started initially before the first registering, and are started when the store first registers (lazy-loading),
- when the store unregisters, informers can be stopped (after a few time).

## TODO

- [x] #6171
- [x] #6155
- [x] #6167
- [x] #6153
- [x] #6202
- [ ] ~~Start Deployment informer when pod informer is connected~~
- [x] #6198
- [ ] ~~#6232~~

### LINKED ISSUE #6171: Send contexts status on different channels
See #6123 (Proposed evolution)

feloy:
### What does this PR do?

Use 4 different channels to send information to the frontend:

- status and number of a selected list of resources (pods, deployments) for all contexts (including the current one)
- Status and number of a selected list of resources (pods, deployments) for current context
- Pods in the current context
- Deploymenets in the current context

New state is sent or not into channels depending on the event (for example, the "Pods in current context" is not sent when a new Deployment is added to a context)

More unit tests 

Fixes #6171

### How to test this PR?

Enable Kubernetes on Podman Desktop, start/stop deployments, start/stop clusters, and check data is always valid on Deployments List page and in Kubernetes Contexts page

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.spec.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.spec.ts
@@ -141,7 +141,7 @@ afterEach(() => {
 test('should send info of resources in all reachable contexts and nothing in non reachable', async () => {
   const client = new ContextsManager(apiSender);
   const kubeConfig = new kubeclient.KubeConfig();
-  kubeConfig.loadFromOptions({
+  const config = {
     clusters: [
       {
         name: 'cluster1',
@@ -185,7 +185,8 @@ test('should send info of resources in all reachable contexts and nothing in non
       },
     ],
     currentContext: 'context2-1',
-  });
+  };
+  kubeConfig.loadFromOptions(config);
   await client.update(kubeConfig);
   let expectedMap = new Map<string, ContextGeneralState>();
   expectedMap.set('context1', {
@@ -232,44 +233,38 @@ test('should send info of resources in all reachable contexts and nothing in non
   });
   expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-pods-update', [{}]);
   expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-deployments-update', [{}, {}, {}, {}]);
+
+  // switching to unreachable context
+  kubeConfig.loadFromOptions({
+    clusters: config.clusters,
+    users: config.users,
+    contexts: config.contexts,
+    currentContext: 'context1',
+  });
+
+  vi.clearAllMocks();
+  await client.update(kubeConfig);
+
+  await new Promise(resolve => setTimeout(resolve, 1200));
+  expect(apiSenderSendMock).toHaveBeenCalledTimes(4);
+  expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-general-state-update', expectedMap);
+  expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-general-state-update', {
+    reachable: false,
+    error: 'Error: connection error',
+    resources: {
+      pods: 0,
+      deployments: 0,
+    },
+  });
+  // no pods/deployemnt are sent, as the context is not reachable
+  expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-pods-update', []);
+  expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-deployments-update', []);
+
   // => removing contexts, should remving clusters from sent info
   kubeConfig.loadFromOptions({
-    clusters: [
-      {
-        name: 'cluster1',
-        server: 'server1',
-      },
-      {
-        name: 'cluster2',
-        server: 'server2',
-      },
-    ],
-    users: [
-      {
-        name: 'user1',
-      },
-      {
-        name: 'user2',
-      },
-    ],
-    contexts: [
-      {
-        name: 'context1',
-        cluster: 'cluster1',
-        user: 'user1',
-      },
-      {
-        name: 'context2',
-        cluster: 'cluster2',
-        user: 'user2',
-      },
-      {
-        name: 'context2-1',
-        cluster: 'cluster2',
-        user: 'user2',
-        namespace: 'ns1',
-      },
-    ],
+    clusters: config.clusters,
+    users: config.users,
+    contexts: config.contexts.filter(ctx => ctx.name !== 'context2-2'),
     currentContext: 'context2-1',
   });
 
--- a/packages/main/src/plugin/kubernetes-context-state.spec.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.spec.ts
@@ -22,6 +22,7 @@ import { ContextsManager } from './kubernetes-context-state.js';
 import type { ApiSenderType } from './api.js';
 import * as kubeclient from '@kubernetes/client-node';
 import type { ErrorCallback, KubernetesObject, ObjectCallback } from '@kubernetes/client-node';
+import { makeInformer } from '@kubernetes/client-node';
 
 interface InformerEvent {
   delayMs: number;
@@ -71,7 +72,7 @@ export class FakeInformer {
       });
     }
   }
-  stop(): void {}
+  async stop(): Promise<void> {}
   on(
     verb: 'change' | 'add' | 'update' | 'delete' | 'error' | 'connect',
     cb: ErrorCallback | ObjectCallback<KubernetesObject>,
@@ -91,8 +92,13 @@ export class FakeInformer {
   ): void {
     this.offCb.set(verb, cb);
   }
-  get(): void {}
-  list(): void {}
+  get(): KubernetesObject {
+    return {};
+  }
+
+  list(): KubernetesObject[] {
+    return [];
+  }
 }
 
 // fakeMakeInformer describes how many resources are in the different namespaces and if cluster is reachable
@@ -100,7 +106,7 @@ function fakeMakeInformer(
   kubeconfig: kubeclient.KubeConfig,
   path: string,
   _listPromiseFn: kubeclient.ListPromise<kubeclient.KubernetesObject>,
-): FakeInformer {
+): kubeclient.Informer<kubeclient.KubernetesObject> & kubeclient.ObjectCache<kubeclient.KubernetesObject> {
   let connectResult: Error | undefined;
   switch (kubeconfig.currentContext) {
     case 'context1':
@@ -133,17 +139,11 @@ const apiSender: ApiSenderType = {
   receive: vi.fn(),
 };
 
-const mocks = vi.hoisted(() => {
-  return {
-    makeInformerMock: vi.fn(),
-  };
-});
-
 vi.mock('@kubernetes/client-node', async importOriginal => {
   const actual = await importOriginal<typeof kubeclient>();
   return {
     ...actual,
-    makeInformer: mocks.makeInformerMock,
+    makeInformer: vi.fn(),
   };
 });
 
@@ -171,7 +171,7 @@ afterEach(() => {
 });
 
 test('should send info of resources in all reachable contexts and nothing in non reachable', async () => {
-  mocks.makeInformerMock.mockImplementation(fakeMakeInformer);
+  vi.mocked(makeInformer).mockImplementation(fakeMakeInformer);
   const client = new ContextsManager(apiSender);
   const kubeConfig = new kubeclient.KubeConfig();
   const config = {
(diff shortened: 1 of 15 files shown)


deboer-tim (commented):
A really quick, jet-lagged review:
- +1 to how this breaks up sending data to the renderer, makes much more sense. Likewise, makes a better pattern for adding other objects like Services, etc.
- I think we'll still have a problem with creating informers for all objects for every cluster... but maybe that's a deferrable problem for now, or we can solve separately.
- IMHO we didn't need to rename ContextState to make it clear it's general, but I'm fine with it.
- I don't love special-casing current context in the API... but it is expedient and all that's necessary for now, so I'm also good with that.

feloy:
> A really quick, jet-lagged review:
> 
> * +1 to how this breaks up sending data to the renderer, makes much more sense. Likewise, makes a better pattern for adding other objects like Services, etc.
> * I think we'll still have a problem with creating informers for all objects for every cluster... but maybe that's a deferrable problem for now, or we can solve separately.

I'll work on the next PR to add informers for others resources (services, ingresses, etc)
 by starting them lazily (the first time they are queried), and stopping them when not needed anymore.

> * IMHO we didn't need to rename ContextState to make it clear it's general, but I'm fine with it.
> * I don't love special-casing current context in the API... but it is expedient and all that's necessary for now, so I'm also good with that.

feloy:
@benoitf @dgolovin @axel7083 @deboer-tim this PR is ready for review (related to Kubernetes for next release)

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -42,29 +42,58 @@ interface ContextInternalState {
...
+type ResourcesDispatchOptions = {
+  [resourceName in ResourceName]?: boolean;
+};
+
+interface DispatchOptions {
+  // do we send general context for all contexts?
+  contextsGeneralState: boolean;
+  // do we send general context for current context?
+  currentContextGeneralState: boolean;
+  // do we end resources data for each resource kind? default false for all resources
```
deboer-tim: nit: send

deboer-tim (approved):
This is a good step in the right direction, LGTM. All the potential pitfalls I was able to come up with were covered, and worked well in testing.

packages/main/src/plugin/kubernetes-context-state.spec.ts
```diff
@@ -109,39 +133,48 @@ const apiSender: ApiSenderType = {
   receive: vi.fn(),
 };
 
+const mocks = vi.hoisted(() => {
+  return {
+    makeInformerMock: vi.fn(),
+  };
+});
```
benoitf: hello, why do we need this hoisted section ?

packages/main/src/plugin/kubernetes-context-state.spec.ts
```diff
@@ -109,39 +133,48 @@ const apiSender: ApiSenderType = {
...
+
 vi.mock('@kubernetes/client-node', async importOriginal => {
   const actual = await importOriginal<typeof kubeclient>();
   return {
     ...actual,
-    makeInformer: fakeMakeInformer,
+    makeInformer: mocks.makeInformerMock,
```
benoitf: why do we need this and not only

```
makeInformer: vi.fn()
```

packages/main/src/plugin/kubernetes-context-state.spec.ts
```diff
@@ -109,39 +133,48 @@ const apiSender: ApiSenderType = {
...
+  vi.useFakeTimers();
 });
 
 afterEach(() => {
   console.debug = originalConsoleDebug;
+  vi.resetAllMocks();
 });
 
 test('should send info of resources in all reachable contexts and nothing in non reachable', async () => {
+  mocks.makeInformerMock.mockImplementation(fakeMakeInformer);
```
benoitf: vi.mocked(....makeInformer).mockImplementation(fakeMakeInformer);

so you'll have the typechecking

benoitf (commented):
how do we track the context state stores ?

in Podman Desktop, when you go with the light bulb, you can monitor how many objects are in the stores, retrigger the update

it would be nice if there would have update/storage on how many objects are being stored, how many times it has been updated, etc.

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -319,26 +423,84 @@ export class ContextsManager {
...
-    // Debounce: send only the latest value if several values are sent in a short period
-    clearTimeout(this.timeoutId);
-    this.timeoutId = setTimeout(() => {
-      this.apiSender.send(`kubernetes-contexts-state-update`, this.states.getPublished());
-    }, connectTimeout);
+  private setStateAndDispatch(
+    name: string,
+    options: ResourcesDispatchOptions,
+    update: (previous: ContextState) => void,
+  ) {
```
benoitf: it's missing return type

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -319,26 +423,84 @@ export class ContextsManager {
...
+    this.dispatch({
+      contextsGeneralState: true,
+      currentContextGeneralState: true,
+      resources: options,
+    });
   }
 
-  public getContextsState(): Map<string, ContextState> {
-    return this.states.getPublished();
+  private dispatch(options: DispatchOptions) {
```
benoitf: missing return type

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -319,26 +423,84 @@ export class ContextsManager {
...
+    if (options.resources?.deployments) {
+      this.dispatchCurrentContextResource('deployments');
+    }
   }
 
-  private setStateAndDispatch(name: string, update: (previous: ContextState) => void) {
-    this.states.safeSetState(name, update);
-    this.dispatchContextsState();
+  private generalStateTimeoutId: NodeJS.Timeout | undefined;
+  private dispatchContextsGeneralState() {
```
benoitf: missing return type

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -319,26 +423,84 @@ export class ContextsManager {
...
+    return this.states.getContextsGeneralState();
   }
 
   private resetConnectionAttempts<T extends KubernetesObject>(options: CreateInformerOptions<T>) {
     options.backoff.reset();
     clearTimeout(options.timer);
   }
+
+  private currentStateTimeoutId: NodeJS.Timeout | undefined;
+  private dispatchCurrentContextGeneralState() {
```
benoitf: missing return type

benoitf: is there a reason to have all methods being private and not protected for example ?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -319,26 +423,84 @@ export class ContextsManager {
...
+      );
+    }, connectTimeout);
+  }
+
+  public getCurrentContextGeneralState(): ContextGeneralState {
+    return this.states.getCurrentContextGeneralState(this.kubeConfig.currentContext);
+  }
+
+  private resourceTimeoutId = new Map<ResourceName, NodeJS.Timeout>();
+  private dispatchCurrentContextResource(resourceName: ResourceName) {
```
benoitf: missing return type

packages/renderer/src/lib/ui/KubernetesCurrentContextConnectionBadge.svelte
```diff
@@ -1,14 +1,14 @@
 <script lang="ts">
 import { kubernetesCurrentContextState } from '/@/stores/kubernetes-contexts-state';
-import type { ContextState } from '../../../../main/src/plugin/kubernetes-context-state';
+import type { ContextGeneralState } from '../../../../main/src/plugin/kubernetes-context-state';
 import Tooltip from '/@/lib/ui/Tooltip.svelte';
 
-function getText(state: ContextState | undefined): string {
+function getText(state: ContextGeneralState | undefined): string {
```
benoitf: use `state?: ContextGeneralState`

packages/renderer/src/lib/ui/KubernetesCurrentContextConnectionBadge.svelte
```diff
@@ -1,14 +1,14 @@
...
   if (state?.reachable) return 'Connected';
   return 'Cluster not reachable';
 }
 
-function getClassColor(state: ContextState | undefined): string {
+function getClassColor(state: ContextGeneralState | undefined): string {
```
benoitf: use `state?: ContextGeneralState`

packages/main/src/plugin/kubernetes-context-state.ts
feloy: I have the habit to declare as private by default. protected would be the default for you?

benoitf: for testing purpose yes else how do you test your private methods ?

feloy: ok good point. The tests I have written for this class are testing it as a black box, only calling the entrypoint `update`, but I get the point

feloy:
> how do we track the context state stores ?
> 
> in Podman Desktop, when you go with the light bulb, you can monitor how many objects are in the stores, retrigger the update
> 
> it would be nice if there would have update/storage on how many objects are being stored, how many times it has been updated, etc.

Yes, for the moment the stores are using the vanillas Svelte `readable`, but I'll check how to upgrade them to the `EventStore` or similar

benoitf:
FYI my question about the monitoring was a question and a general remark, not blocking the PR

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -113,13 +142,56 @@ class ContextsStates {
...
+      result.set(key, {
+        ...val,
+        resources: {
+          pods: val.reachable ? val.resources.pods.length : 0,
+          deployments: val.reachable ? val.resources.deployments.length : 0,
+        },
+      });
+    });
+    return result;
+  }
```
benoitf, line 157: not a required change, just a FYI
we could have used the map() method to transform the object instead of initializing an object, iterating on a first object, getting value from another one, and then returning an object

benoitf (approved):

packages/main/src/plugin/kubernetes-context-state.ts
feloy: I'm not used to do this with `Map`s objects, only with arrays. I'll check how to do this, thanks

PR #6136, opened 2024-02-26
chore: switch from ts-node to tsx

benoitf:
### What does this PR do?
some packages are being switched to esm 
example: https://github.com/containers/podman-desktop/pull/4776

And ts-node does not handle it properly and require also some configuration while tsx just works out of the box

so get rid of all "ts-node" elements and just use 'tsx'

registries had a ts-node configuration but mostly a copy/paste error

https://github.com/containers/podman-desktop/pull/4776

### How to test this PR?

check that podman files are being downloaded and use 'install:contour' in kind extension and see that it is being downloaded as well

### DIFF
--- a/extensions/kind/package.json
+++ b/extensions/kind/package.json
@@ -95,7 +95,7 @@
     ]
   },
   "scripts": {
-    "install:contour": "npx ts-node ./scripts/download.ts",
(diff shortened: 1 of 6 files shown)


jeffmaury (approved):
LGTM

PR #6137, opened 2024-02-26
feat: show kube current badge only when Kubernetes experimental is enabled

### LINKED ISSUE #6130: Hide KubernetesCurrentContextStatus when Kubernetes experimental feature is disabled

We should not show the `KubernetesCurrentContextConnectionBadge` in pods when kubernetes experimental support is not enable

Hiding it 

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Hide KubernetesCurrentContextStatusBadge from PodList page when Kubernetes experimental feature is disabled
> Why only the PodList ? Because the DeploymentList, ServiceList etc. are not accessible when the feature is disabled.
### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/42176370/728f4912-de0d-4ed1-a3cc-26696265addf

### How to test this PR?

Fixes https://github.com/containers/podman-desktop/issues/6130

**manually**
(1) Go to PodList page with or without the experimental mode
(2) turn it on/off in the preferences
(3) Return to PodList

### DIFF
--- a/packages/renderer/src/lib/pod/PodsList.spec.ts
+++ b/packages/renderer/src/lib/pod/PodsList.spec.ts
@@ -20,7 +20,7 @@
 
 import '@testing-library/jest-dom/vitest';
 import { beforeAll, test, expect, vi } from 'vitest';
-import { fireEvent, render, screen } from '@testing-library/svelte';
+import { fireEvent, render, screen, waitFor } from '@testing-library/svelte';
 import PodsList from '/@/lib/pod/PodsList.svelte';
 import type { ProviderInfo } from '../../../../main/src/plugin/api/provider-info';
 import { get } from 'svelte/store';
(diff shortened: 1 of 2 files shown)


benoitf:
LGTM but wondering if we should not do the opposite, ie: remove the experimental flag

experimental was there to hide the stuff in 1.7 but it should be displayed in 1.8 so probably time to display it by default

axel7083:
> LGTM but wondering if we should not do the opposite, ie: remove the experimental flag
> 
> experimental was there to hide the stuff in 1.7 but it should be displayed in 1.8 so probably time to display it by default

Good question, we might want to talk about that in the next meeting - I will wait for it to be discuss before merging this (if we keep the flag)

axel7083:
Closing as not necessary anymore see https://github.com/containers/podman-desktop/issues/6143

PR #6159, opened 2024-02-27
refactor: moving update logic to a separate file

### LINKED ISSUE #5755: Improvements to new version alerts

axel7083:
### What does this PR do?

Moving the logic of the updater system (Status Bar Item, error handling etc.) from the index to a dedicated file.

Related to https://github.com/containers/podman-desktop/issues/5754

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/updater.spec.ts
+++ b/packages/main/src/plugin/updater.spec.ts
@@ -68,12 +68,12 @@ beforeEach(() => {
   // Simulate PROD env
   vi.stubEnv('PROD', 'true');
 
-  vi.spyOn(app, 'getVersion').mockReturnValue('@debug');
+  vi.mocked(app.getVersion).mockReturnValue('@debug');
   // eslint-disable-next-line no-null/no-null
-  vi.spyOn(autoUpdater, 'checkForUpdates').mockResolvedValue(null);
+  vi.mocked(autoUpdater.checkForUpdates).mockResolvedValue(null);
 
-  vi.spyOn(commandRegistry, 'executeCommand').mockResolvedValue(undefined);
-  vi.spyOn(util, 'isLinux').mockReturnValue(false);
+  vi.mocked(commandRegistry.executeCommand).mockResolvedValue(undefined);
+  vi.mocked(util.isLinux).mockReturnValue(false);
 });
 
 test('expect env PROD to be truthy', () => {
@@ -88,9 +88,8 @@ test('expect init to provide a disposable', () => {
 });
 
 test('expect init to register commands', () => {
-  const registerCommandMock = vi.spyOn(commandRegistry, 'registerCommand');
   new Updater(messageBoxMock, statusBarRegistryMock, commandRegistry).init();
-  expect(registerCommandMock).toHaveBeenCalled();
+  expect(commandRegistry.registerCommand).toHaveBeenCalled();
 });
 
 test('expect update available entry to be displayed when expected', () => {
--- a/packages/main/src/plugin/updater.spec.ts
+++ b/packages/main/src/plugin/updater.spec.ts
@@ -101,7 +101,7 @@ test('expect update available entry to be displayed when expected', () => {
       expect(text).toBe('v@debug');
       expect(tooltip).toBe('Update available');
       expect(iconClass).toBe(UPDATER_UPDATE_AVAILABLE_ICON);
-      expect(enabled).toBe(true);
+      expect(enabled).toBeTruthy();
       expect(command).toBe('update');
       expect(highlight).toBe(true);
     },
--- a/packages/main/src/plugin/updater.spec.ts
+++ b/packages/main/src/plugin/updater.spec.ts
@@ -103,7 +103,7 @@ test('expect update available entry to be displayed when expected', () => {
       expect(iconClass).toBe(UPDATER_UPDATE_AVAILABLE_ICON);
       expect(enabled).toBeTruthy();
       expect(command).toBe('update');
-      expect(highlight).toBe(true);
+      expect(highlight).toBeTruthy();
     },
   );
 
--- a/packages/main/src/plugin/updater.spec.ts
+++ b/packages/main/src/plugin/updater.spec.ts
@@ -23,57 +23,42 @@ import type { CommandRegistry } from '/@/plugin/command-registry.js';
 import { Disposable } from '/@/plugin/types/disposable.js';
 import { UPDATER_UPDATE_AVAILABLE_ICON } from '/@/plugin/index.js';
 import { Updater } from '/@/plugin/updater.js';
-
-const mocks = vi.hoisted(() => ({
-  // MessageBoxMock
-  showMessageBoxMock: vi.fn(),
-  // StatusBarRegistry
-  setEntryMock: vi.fn(),
-  // CommandRegistry
-  registerCommandMock: vi.fn(),
-  executeCommandMock: vi.fn(),
-  // Electron App
-  getVersionMock: vi.fn(),
-  // Electron Update
-  downloadUpdateMock: vi.fn(),
-  quitAndInstallMock: vi.fn(),
-  checkForUpdatesMock: vi.fn(),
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/updater.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  // Electron App
+  getVersionMock: vi.fn(),
+  // Electron Update
+  downloadUpdateMock: vi.fn(),
+  quitAndInstallMock: vi.fn(),
+  checkForUpdatesMock: vi.fn(),
+  onMock: vi.fn(),
+  // utils
+  isLinuxMock: vi.fn(),
+}));
```
benoitf: hello, I'm not sure to understand why we need this block ?

if you do
```ts
vi.mock('electron-updater', () => ({
  autoUpdater: {
    downloadUpdate: vi.fn()
})
```

benoitf: and then `vi.mocked(...)` to access the mocked method

axel7083: I have a preference for this method, of declaring all the mocks in a single places, and accessing them through `mocks.[name]`

We see often some `vi.spy`, of `vi.mock` but when we want to set default implementation in the beforeEach combined with a `vi.resetAllMocks();` it is easier to access those mock isntance through a single place, rather than calling several methods.

There are in the code, also some tests declaring `MockInstance` objects, in multiple places, in AI-Studio we were in favour of using this single mocks objects to keep them in a single places. IMHO I feel like it improve readability rather than using the spy method on objects

benoitf: they'll still be declared at a single place

currently you're losing type by doing that

benoitf (dismissed):
please review the mocking strategy to have something with typesafety

axel7083:
> please review the mocking strategy to have something with typesafety

@benoitf done

packages/main/src/plugin/updater.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  // Simulate PROD env
+  vi.stubEnv('PROD', 'true');
+
+  vi.spyOn(app, 'getVersion').mockReturnValue('@debug');
+  // eslint-disable-next-line no-null/no-null
+  vi.spyOn(autoUpdater, 'checkForUpdates').mockResolvedValue(null);
+
+  vi.spyOn(commandRegistry, 'executeCommand').mockResolvedValue(undefined);
+  vi.spyOn(util, 'isLinux').mockReturnValue(false);
```
benoitf: note: as it's mocked you can avoid the spy and do vi.mocked(app.getVersion).mockReturnValue('@debug');

packages/main/src/plugin/updater.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+test('expect env PROD to be truthy', () => {
+  expect(import.meta.env.PROD).toBeTruthy();
+});
+
+test('expect init to provide a disposable', () => {
+  const updater = new Updater(messageBoxMock, statusBarRegistryMock, commandRegistry);
+  const disposable: unknown = updater.init();
+  expect(disposable).toBeDefined();
+  expect(disposable instanceof Disposable).toBe(true);
```
benoitf:
```suggestion
  expect(disposable instanceof Disposable).toBeTruthy();
```

packages/main/src/plugin/updater.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+  const setEntryMock = vi.spyOn(statusBarRegistryMock, 'setEntry');
+  setEntryMock.mockImplementation(
+    (entryId, _alignLeft, _priority, text, tooltip, iconClass, enabled, command, _commandArgs, highlight) => {
+      expect(entryId).toBe('version');
+      expect(text).toBe('v@debug');
+      expect(tooltip).toBe('Update available');
+      expect(iconClass).toBe(UPDATER_UPDATE_AVAILABLE_ICON);
+      expect(enabled).toBe(true);
+      expect(command).toBe('update');
+      expect(highlight).toBe(true);
```
benoitf:
```suggestion
      expect(highlight).toBeTruthy();
```

packages/main/src/plugin/updater.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+test('expect update available entry to be displayed when expected', () => {
+  const setEntryMock = vi.spyOn(statusBarRegistryMock, 'setEntry');
+  setEntryMock.mockImplementation(
+    (entryId, _alignLeft, _priority, text, tooltip, iconClass, enabled, command, _commandArgs, highlight) => {
+      expect(entryId).toBe('version');
+      expect(text).toBe('v@debug');
+      expect(tooltip).toBe('Update available');
+      expect(iconClass).toBe(UPDATER_UPDATE_AVAILABLE_ICON);
+      expect(enabled).toBe(true);
```
benoitf:
```suggestion
      expect(enabled).toBeTruthy();
```

benoitf (commented):
thanks for the update

I think you don't need the spy as they're already mocked

I'll try the PR

packages/main/src/plugin/updater.spec.ts
axel7083: done

benoitf:
I tried but unfortunately there is an error

https://github.com/containers/podman-desktop/assets/436777/aa41dcb8-623b-4870-9191-70f268e22def

> How I tested ?
I replaced package.json version with 1.6.5 and did `yarn compile:current` and launched the dmg (and it's telling me there is an update from 1.6.5 to 1.7.1)
I click on proceed but then I have a javascript error


Tried with the main branch I got a real error message (saying that my volume was read-only)

axel7083:
@benoitf thanks for pointing out this issue, the tests I've added were not covering properly the error scenarios. The `error` listener was not properly bind to the current class, leading to the javascript error you encountered. 

I fixed the issue and added a regression test.

benoitf:
thanks @axel7083 

let me check again

benoitf (approved):
did the error as expected and successful update when not using read-only volume 🎉 
Screenshot 2024-02-28 at 10 49 48

packages/main/src/plugin/updater.ts
```diff
@@ -0,0 +1,265 @@
...
+    this.#currentVersion = `v${app.getVersion()}`;
+    this.#updateInProgress = false;
+    this.#updateAlreadyDownloaded = false;
+    this.#updateCheckResult = undefined;
+  }
+
+  /**
+   * Sets the default version entry in the status bar.
+   */
+  private defaultVersionEntry() {
```
benoitf: it looks like we're missing return type

packages/main/src/plugin/updater.ts
```diff
@@ -0,0 +1,265 @@
...
+      true,
+      'version',
+      undefined,
+    );
+  }
+
+  /**
+   * Sets the update available entry in the status bar.
+   */
+  private updateAvailableEntry() {
```
benoitf: it looks like we're missing return type

packages/main/src/plugin/updater.ts
```diff
@@ -0,0 +1,265 @@
...
+      'update',
+      undefined,
+      true,
+    );
+  }
+
+  /**
+   * Registers commands related to version and update.
+   */
+  private registerCommands() {
```
benoitf: it looks like we're missing return type

axel7083: Why is the linter not saying something ?

benoitf: probably something is missing 🤔

benoitf: https://github.com/containers/podman-desktop/pull/6183

PR #6161, opened 2024-02-27
fix: don't check default machine on Linux

afbjorklund:
### What does this PR do?

Avoids setting a default system connection on Linux.

Chances are, user wants to decide this themselves...

`podman-remote version`

`podman-remote-static --connection podman-machine-default version`

### How to test this PR?

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -54,7 +54,7 @@ let autoMachineStarted = false;
 let autoMachineName;
 
 // System default notifier
-let defaultMachineNotify = true;
+let defaultMachineNotify = !isLinux();
 let defaultMachineMonitor = true;
 
 // current status of machines
--- a/extensions/podman/src/podman-install.spec.ts
+++ b/extensions/podman/src/podman-install.spec.ts
@@ -70,6 +70,7 @@ vi.mock('./util', async () => {
     runCliCommand: vi.fn(),
     appHomeDir: vi.fn().mockReturnValue(''),
     normalizeWSLOutput: vi.fn().mockImplementation((s: string) => s),
+    isLinux: vi.fn(),
   };
 });
 


cdrage:
Thank you for the quick fix! I've tested this on Linux and it does not show up anymore / no endless looping, so LGTM my side.

Will need 1 more tester though, @lstocchi / @benoitf ?

cdrage (approved):

benoitf (approved):

afbjorklund:
The endless looping is still there though, this just hides the extra dialog on startup.

The workaround was to use the "Ignore" button instead of the default "Yes" button

cdrage:
> The endless looping is still there though, this just hides the extra dialog on startup.
> 
> The workaround was to use the "Ignore" button instead of the default "Yes" button

I tested on Fedora 39 with `podman` cli installed only with no machine, it didn't appear again when using `yarn watch`.

Do you test via flatpak?

afbjorklund:
> Do you test via flatpak?

Releases, yes (flatpak)

Development is usually `yarn watch` like you say. Will delete the machine and try again (with `main`)...

cdrage:
Okay testing via flatpak now on Fedora 39, you are right, it does appear again.

Testing via `yarn watch` the setup somehow dissapears.

I believe going forward we should have some e2e tests implemented eventually @odockal

cdrage:
Okay testing via flatpak now on Fedora 39, you are right, it does appear again.

Testing via `yarn watch` the setup somehow disappears. Unsure if this is particular to my setup, or because it's on `yarn watch`.

I believe going forward we should have some e2e tests implemented eventually @odockal

afbjorklund:
Anyway, please continue discussion in the other issue (or some other venue) - this one seems to be closed now.

PR #5800, opened 2024-02-01
chore(test): implementing podman machine creation through onboarding workflow e2e test

ScrewTSW:
### What does this PR do?
* Updates DashboardPage page object adding a locator for notifications
* Implements Podman Desktop Machine e2e test using Onboarding Workflow

### Screenshot / video of UI
N/A

### What issues does this PR fix or reference?
#4670 

### How to test this PR? (Windows only!!!)
* Podman must be installed and configured (windows virtualization, WSL2 installed), there can be no podman machine present
* append `"test:e2e:machineonboarding": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/src/podman-machine-onboarding.spec.ts --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",` into `package.json`
* run `yarn run test:e2e:build`
* run `yarn run test:e2e:machineonboarding`

### Additional context
Example test runs:
```
$ TEST_PODMAN_MACHINE=true MACHINE_CLEANUP=true yarn run test:e2e:machineonboarding:run

...

 ✓ tests/src/podman-machine-onboarding.spec.ts (13) 82640ms
   ❯ [ beforeAll ]
   ✓ Podman Machine verification (13) 66291ms
     ✓ Podman Machine onboarding workflow (7) 2598ms
       ✓ Setup Podman push notification is present
       ✓ Onboarding navigation (3) 791ms
         ✓ Open Podman Machine Onboarding through Setup Notification
         ✓ Return to Dashboard
         ✓ Re-Open Podman Machine Onboarding through Settings Resources page 394ms
       ✓ Verify Podman Autostart is enabled and proceed to next page 424ms
       ✓ Expect no machine created message and proceed to next page
       ✓ Verify default podman machine settings 1003ms
     ✓ Podman Machine creation and operations (5) 51658ms
       ✓ Create a default Podman machine 26505ms
       ✓ Podman machine operations (4) 25153ms
         ✓ Open podman machine details 1173ms
         ✓ Podman machine operations - STOP 6024ms
         ✓ Podman machine operations - START 5991ms
         ✓ Podman machine operations - RESTART 11965ms
     ✓ Clean Up Podman Machine 12035ms

 Test Files  1 passed (1)
      Tests  13 passed (13)
   Start at  10:01:06
   Duration  85.04s (transform 206ms, setup 18ms, collect 1.08s, tests 82.64s, environment 511ms, prepare 126ms)

[...]

PR #6164, opened 2024-02-27
chore(tests): pod machine created from dashboard

### LINKED ISSUE #4669: Implement E2E test case covering initialization and start up of a Podman Machine via dashboard page
Implement a test case to cover Podman Machine Creation using playwright via Dashboard page - Initialize and start - which spins up machine with default values.

cbr7:
Signed-off-by: Vladimir Lazar <vlazar@redhat.com>

### What does this PR do?
Create e2e test for  podman machine onboarding from the dashboard

https://github.com/containers/podman-desktop/issues/4669

### How to test this PR?
run `yarn test:e2e:pmachine`


package.json
```diff
@@ -55,6 +55,8 @@
     "test:e2e:bootc": "npm run test:e2e:copy && npm run test:e2e:build && npm run test:e2e:bootc:run",
     "test:e2e:bootc:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/src/bootc-extension.spec.ts --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",
     "test:e2e:copy": "cp ../podman-desktop-extension-bootc/tests/src/bootc-extension.spec.ts tests/src/",
+    "test:e2e:pmachine": "npm run test:e2e:build && npm run test:e2e:pmachine:run",
+    "test:e2e:pmachine:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/src/podman-machine-dashboard-onboarding.spec.ts --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",
```
benoitf: not clear to me why we need again new entries there ?

benoitf: if you need some selector I think it might be better to have a generic target where you can provide the specific path to the file

If we need a script for each file I think it's not scaling

tests/src/podman-machine-dashboard-onboarding.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+beforeEach<RunnerTestContext>(async ctx => {
+  ctx.pdRunner = pdRunner;
+});
+
+afterAll(async () => {
+  await pdRunner.close();
+});
+
+describe(`Podman machine onboarding from DP Dashboard`, async () => {
+  test('Create Podman machine from PD Dashboard', async () => {
```
benoitf: typo on DP ?

anyway I don't think we need to use PD or Podman Desktop as anyway we're testing Podman Desktop

cbr7: fixed, removed the naming

package.json
cbr7: The issue is that during normal smoke testing the podman machines are created before the test suite runs, so I assumed that this test would run separately. Maybe @odockal can give some insight.

odockal: @benoitf I agree, this solution is not scaling. 
@cbr7 you are also right, it is hard to include test that messes(tests) with underlying env. and influences other tests. 
In other tests (podman onboarding where PR is opened) @ScrewTSW added a env. vars. to operate the tests, I think this is also something we could make use of. 
I am working on a solution to propagate env. vars, from GHA workflows inputs into testing images on gha.

cbr7: @odockal which PR is this one that you mentioned so I can take a look and use the same env var.

cbr7: Nevermind found the PR.

odockal:
@cbr7 Please rebase as we got another PR in #5800.

cbr7:
@odockal done.

odockal (approved):

ScrewTSW (approved):
LGTM

Issue #6153, opened 2024-02-27
Better support of connection detection


feloy:
When starting an informer:
- a `connect` event is received (with no error) when the connection is initiated
- if the connection succeeds, no event is received to signal this success
- if the connection fails, an `error` event is received
- after connection success, if resources are in the context, `add` event are received

To detect a reachable context, we need to wait after receiving the `connect` signal:
- either some (arbitrary) timeout, after which, if no `error` event is received, the connection is considered successful
-  or `add` events, in which case we can sonsider the connection successful.

PR #6166, opened 2024-02-27
fix: use a timer per context

### LINKED ISSUE #6155: Each context must have its own re-connection timer
On `ContextsManager` (`kubernetes-context-state.ts`), a single `podTimer` is used for all contexts. Each context must have its own timer.

feloy:

Fixes #6155 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -259,13 +259,13 @@ export class ContextsManager {
     informer.on('add', (obj: T) => {
       options.onAdd?.(obj);
       options.onReachable?.(true);
-      options.backoff.reset();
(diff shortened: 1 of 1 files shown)


benoitf (commented):
is that the timers are cancelled at some point ?

feloy:
> is that the timers are cancelled at some point ?

No, good point. As part of #6153, I'll try to have a more concrete Connected state, so I could stop the timers at this moment. For now, it's quite hacky to detect that we are connected and we can stop the connection retries (it is done for the backoff reinitialization on `add`/`delete` events, but it isn't execute if there is no resources on the context)

For now, I'll do it when a reset the backoff

Thanks

benoitf (approved):

deboer-tim (approved):

Issue #5757, opened 2024-01-30
Initial UI component library


jeffmaury:
### Epic domain

Podman Desktop should expose UI components as a library to be used by extension developers.

- [x] #5758
- [x] #6896
- [x] #6897
- [x] #6898
- [x] #6899
- [x] ~~#6900~~
- [x] #6901
- [x] #6902
- [x] #6903
- [x] #6904
- [x] #6905
- [x] #6906
- [x] #6907
- [x] #6908
- [x] #6909
- [x] #6910
- [x] #6918
- [x] #6919
- [x] #6920
- [x] #6921
- [x] #6922
- [x] #6923
- [x] #7690

deboer-tim:
Maybe not required for first pass, but I noticed two more components that could be reused in extensions today: #7964 and #7965.

PR #6172, opened 2024-02-27
feat: introduce a svelte ui package

benoitf:
### What does this PR do?
It brings a new yarn workspace with components that could be reused/shared with other projects.
It should include most of the UI components used to display and for the UI layout

Then we should publish the ui-svelte package to npmjs so other components could reuse components defined there

The renderer package has a dependency to this new module

for now, there is no component being defined but stay tuned 🎉

### Screenshot / video of UI

N/A

#5757 

### How to test this PR?

should work as of today 🤞 


packages/ui/index.html
axel7083, line 1: What is the purpose of this HTML file in the UI packages ? Does svelte requires this ?

axel7083 (approved):

packages/ui/index.html
benoitf: I think it was vite, but as I did several attempts before issuing the PR I'll check if we really need it now that I switched to svelte-package

benoitf: removed html file

I also updated the scripts to match the reality

axel7083: Okey perfect!

jeffmaury (approved):
LGTM

PR #6192, opened 2024-02-28
feat: better reachable detection from Kubernetes informers

### LINKED ISSUE #6153: Better support of connection detection

feloy:

Fixes #6153

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state-constants.ts
+++ b/packages/main/src/plugin/kubernetes-context-state-constants.ts
@@ -16,7 +16,14 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+// set this value as the (reasonably) maximum time a connection can get to be established
+// or the reachable status of an unreachable context will flash as reachable during the connection attempt
 export const connectTimeout = 1000;
+// initial delay between two connection attempts
 export const backoffInitialValue = 1000;
+// maximum delay between two connection attermpts
 export const backoffLimit = 60_000;
+// jitter to add to the delay between two connection attempts
 export const backoffJitter = 300;
+// the time to wait for any update on the data to dispatch before to send it
+export const dispatchTimeout = 100;
--- a/packages/main/src/plugin/kubernetes-context-state-constants.ts
+++ b/packages/main/src/plugin/kubernetes-context-state-constants.ts
@@ -21,7 +21,7 @@
 export const connectTimeout = 1000;
 // initial delay between two connection attempts
 export const backoffInitialValue = 1000;
-// maximum delay between two connection attermpts
+// maximum delay between two connection attempts
 export const backoffLimit = 60_000;
 // jitter to add to the delay between two connection attempts
 export const backoffJitter = 300;
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -34,6 +34,7 @@ import {
   backoffJitter,
   backoffLimit,
   connectTimeout,
+  dispatchTimeout,
 } from './kubernetes-context-state-constants.js';
 import type { IncomingMessage } from 'node:http';
 
@@ -52,6 +53,7 @@ interface ContextState {
 }
 
 // All resources managed by podman desktop
+// TODO: add new resources here when adding new informers
 export type ResourceName = 'pods' | 'deployments';
 
 // A selection of resources, to indicate the 'general' status of a context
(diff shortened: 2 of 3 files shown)


packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -400,14 +427,44 @@ export class ContextsManager {
...
+    options.connectionDelay = setTimeout(() => {
+      this.doSetReachable(options, reachable);
+    }, connectTimeout);
+  }
+
+  private setReachableNow<T extends KubernetesObject>(options: CreateInformerOptions<T>, reachable: boolean): void {
+    clearTimeout(options.connectionDelay);
+    options.connectionDelay = undefined;
+    this.doSetReachable(options, reachable);
+  }
```
benoitf: should it be a common method ?
also, would it be important if now would just call with a timeout of 0

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -400,14 +427,44 @@ export class ContextsManager {
...
+    options.onReachable?.(reachable);
+    if (reachable) {
+      this.resetConnectionAttempts(options);
+    }
+  }
+
+  private resetConnectionAttempts<T extends KubernetesObject>(options: CreateInformerOptions<T>): void {
+    options.backoff.reset();
+    clearTimeout(options.timer);
+    options.timer = undefined;
```
benoitf, line 466: would it be better using disposable ?

```
options.dispose();
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -442,67 +499,48 @@ export class ContextsManager {
...
+      if (options.currentContextGeneralState) {
+        this.dispatchCurrentContextGeneralState();
+      }
+      if (options.resources?.pods) {
+        this.dispatchCurrentContextResource('pods');
+      }
+      if (options.resources?.deployments) {
+        this.dispatchCurrentContextResource('deployments');
+      }
+    }, dispatchTimeout);
```
benoitf: it looks like a lot of copy/paste for almost the same call
probably need to refactor to make less calls

benoitf (commented):
there are still a lot of TODO in the code so I don't know if it's ready for a review

feloy:
> there are still a lot of TODO in the code so I don't know if it's ready for a review

There are two TODOs, not for this PR, but for future PRs that will add new informers

packages/main/src/plugin/kubernetes-context-state.ts
feloy: I changed to always use a timeout, either > 0 or 0

packages/main/src/plugin/kubernetes-context-state.ts
feloy: As there are several timers in options, it would not work here. But I note that these timers need to be disposed when podman desktop exits, I created an issue to track this (#6198)

packages/main/src/plugin/kubernetes-context-state.ts
feloy: I refactored the last two ones. This will be useful when adding more informers

packages/main/src/plugin/kubernetes-context-state.ts
benoitf: thanks

packages/main/src/plugin/kubernetes-context-state.ts
benoitf: thanks

packages/main/src/plugin/kubernetes-context-state-constants.ts
```diff
@@ -16,7 +16,14 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+// set this value as the (reasonably) maximum time a connection can get to be established
+// or the reachable status of an unreachable context will flash as reachable during the connection attempt
 export const connectTimeout = 1000;
+// initial delay between two connection attempts
 export const backoffInitialValue = 1000;
+// maximum delay between two connection attermpts
```
benoitf:
```suggestion
// maximum delay between two connection attempts
```

benoitf (approved):

deboer-tim (approved):
Please remove the TODOs or change into a general 'this is where you add new informers' comment.

PR #6196, opened 2024-02-28
chore: move input ui component to the ui package

benoitf:
### What does this PR do?
move input ui component to the ui package

depends on:
- [x] https://github.com/containers/podman-desktop/pull/6194
- [x] https://github.com/containers/podman-desktop/pull/6195
- [x] https://github.com/containers/podman-desktop/pull/6197
- [x] https://github.com/containers/podman-desktop/pull/6201

### Screenshot / video of UI

no UI change

#5757 

### How to test this PR?

Check that the input components are looking as before


axel7083:
I would not be in favour of copy pasting components in the ui package, like I would prefer either creating new components, understanding the needs, the wording used etc. 

We could for example base some of the input components on other components library, to see how they make them reusable etc.

For example, the input component does not propagate classes properly

https://github.com/containers/podman-desktop/blob/5cd944824c15572cc9b5becbe48d87b06c575a20/packages/ui/src/lib/input/Input.svelte#L36-L38

- why are they taking all width ?
- why the class props is not to the parent container and the first child ?

benoitf:
@axel7083 I guess these comments should have been made on https://github.com/containers/podman-desktop/pull/5904

benoitf:
If we don't move quickly components, there are forks in place so to me it's an higher problem because fixes are not reported to a single place.

If there are required changes to the component, I don't see any problem, the component can evolve.

Also to me Input component looks like a good candidate as anyway you always need input fields and it needs to be consistent across all the UI.

axel7083:
> If we don't move quickly components, there are forks in place so to me it's an higher problem because fixes are not reported to a single place.
> 
> If there are required changes to the component, I don't see any problem, the component can evolve.
> 
> Also to me Input component looks like a good candidate as anyway you always need input fields and it needs to be consistent across all the UI.

Using the components `as it is` from the renderer package, and copying to the ui, makes them instantly very hard to change. I understand that the priority is probably not on rebuilding components, but I feel like this is just going to move the problem from one place to another.

axel7083 (approved):

deboer-tim (approved):
@axel7083 I understand your concern, but @benoitf is right that we have copies and need to move fairly quick to create a shared library. This is not a final API/1.0 library though, we should be clear about that and continue to evolve.

cdrage (approved):

PR #6209, opened 2024-02-29
Start other informers lazily

### LINKED ISSUE #6202: Start Services/Ingress/Route/... informers lazily
For the moment, the `ContextsManager` starts informers for pods and deployments (for each kubernetes context).

It should also start other informers for other resources managed by Podman Desktop (services, ingresses, routes). 

These informers should not be started immediately, but only on demand (for example based on a subscription system), and only for the current context.
  
See also #6113 and #6094 

- [x] start service informer when service store asks the first time the list of services (#6209)
- [x] stop the service informer on previous context when the current context changes (#6209)
- [x] #6254
- [ ] ~~stop service informer after some time the service store terminates (and unregisters from the ContextsManager)~~
- [x] same for ingresses
- [x] same for routes

feloy:
### What does this PR do?

- start service informer when service store asks the first time the list of services
- stop the service informer on previous context when the current context changes

Fixes partially #6202 

### How to test this PR?

Switch to the services list page and check that the services are visible
Debug logs indicate when the informer is started or not
Check by changing context 

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.spec.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.spec.ts
@@ -257,7 +257,6 @@ test('should send info of resources in all reachable contexts and nothing in non
   } as ContextGeneralState);
   vi.advanceTimersToNextTimer();
   vi.advanceTimersToNextTimer();
-  expect(apiSenderSendMock).toHaveBeenCalledTimes(4);
   expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-general-state-update', expectedMap);
   expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-general-state-update', {
     reachable: true,
@@ -282,7 +281,6 @@ test('should send info of resources in all reachable contexts and nothing in non
   await client.update(kubeConfig);
 
   vi.advanceTimersToNextTimer();
-  expect(apiSenderSendMock).toHaveBeenCalledTimes(4);
   expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-general-state-update', expectedMap);
   expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-general-state-update', {
     reachable: false,
@@ -333,7 +331,6 @@ test('should send info of resources in all reachable contexts and nothing in non
   } as ContextGeneralState);
 
   vi.advanceTimersToNextTimer();
-  expect(apiSenderSendMock).toHaveBeenCalledTimes(4);
   expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-general-state-update', expectedMap);
   expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-general-state-update', {
     reachable: true,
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -147,16 +149,36 @@ class ContextsStates {
...
   }
 
+  setResourceInformer(
+    contextName: string,
+    resourceName: ResourceName,
+    informer: Informer<KubernetesObject> & ObjectCache<KubernetesObject>,
+  ): void {
+    const informers = this.informers.get(contextName);
+    if (!informers) {
+      throw new Error('informers for context not found');
```
benoitf: missing contextName ?

packages/main/src/plugin/kubernetes-context-state.spec.ts
```diff
@@ -117,20 +123,20 @@ function fakeMakeInformer(
...
+      return new FakeInformer(kubeconfig.currentContext, path, 4, connectResult, [], []);
     case '/apis/apps/v1/namespaces/ns2/deployments':
-      return new FakeInformer(5, connectResult, [], []);
+      return new FakeInformer(kubeconfig.currentContext, path, 5, connectResult, [], []);
     case '/apis/apps/v1/namespaces/default/deployments':
-      return new FakeInformer(6, connectResult, [], []);
+      return new FakeInformer(kubeconfig.currentContext, path, 6, connectResult, [], []);
   }
-  return new FakeInformer(0, connectResult, [], []);
+  return new FakeInformer(kubeconfig.currentContext, path, 0, connectResult, [], []);
```
benoitf: looks like the only difference is the number so probably could do

```ts
case '/apis/apps/v1/namespaces/ns1/deployments':
   return buildFakeInformer(DEPLOYMENTS_NS1)
case '/apis/apps/v1/namespaces/ns2/deployments':
      return buildFakeInformer(DEPLOYMENTS_NS2)
```
and replace 4 by a constant like DEPLOYMENTS with value of 4

buildFakeInformer(id) -->
```
return new FakeInformer(kubeconfig.currentContext, path, id, connectResult, [], [])
```

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -147,16 +157,36 @@ class ContextsStates {
...
+  hasContext(name: string): boolean {
     return this.informers.has(name);
   }
 
+  hasInformer(context: string, resourceName: ResourceName): boolean {
+    const informers = this.informers.get(context);
+    if (!informers) {
+      return false;
+    }
+    return !!informers[resourceName];
```
benoitf: could it be a oneliner ?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -208,6 +238,11 @@ class ContextsStates {
     return [];
   }
 
+  isReachable(contextName: string): boolean {
+    const state = this.state.get(contextName);
```
benoitf: do we need this intermediate variable ?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -227,11 +264,30 @@ class ContextsStates {
...
 
   async dispose(name: string): Promise<void> {
-    await this.informers.get(name)?.podInformer?.stop();
-    await this.informers.get(name)?.deploymentInformer?.stop();
+    const informers = this.informers.get(name);
+    if (informers) {
+      for (const res in Object.keys(informers)) {
+        const resname = res as ResourceName;
+        await informers[resname]?.stop();
+      }
```
benoitf: wondering why we iterate on the keys and then grab values and not on the entries to avoid the `informers[resName]?`
or also why we do not simply iterate on the values ?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -227,11 +264,30 @@ class ContextsStates {
...
   }
+
+  async disposeSecondaryInformers(contextName: string): Promise<void> {
+    const informers = this.informers.get(contextName);
+    if (informers) {
+      for (const res of Object.keys(informers)) {
+        if (isSecondaryResourceName(res)) {
+          console.debug(`stopping informer for ${res} in ${contextName}`);
+          const resname = res as ResourceName;
+          await informers[resname]?.stop();
```
benoitf: iterate on entries ?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -297,11 +364,29 @@ export class ContextsManager {
...
+      throw new Error('context not found');
+    }
+    const ns = context.namespace ?? 'default';
+    let informer: Informer<KubernetesObject> & ObjectCache<KubernetesObject>;
+    switch (resourceName) {
+      case 'services':
+        informer = this.createServiceInformer(this.kubeConfig, ns, context);
+        break;
+      default:
+        console.debug(`trying to start informer for resource ${resourceName} which is not supported`);
```
benoitf: need the contextName displayed in the log I would say

Also informer looks like too technical/internal if end user look the message IMHO he would expect what was the intent

`Unable to monitor/watch/whatever resource ${resourceName} for context ${contextName} as it is not supported`

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -381,6 +470,45 @@ export class ContextsManager {
...
+        this.setStateAndDispatch(
+          context.name,
+          false,
+          { services: true },
+          state =>
+            (state.resources.services = state.resources.services.filter(d => d.metadata?.uid !== obj.metadata?.uid)),
+        );
+      },
+    });
+  }
```
benoitf, line 502: IMHO this method is almost the same than createDeploymentInformer or createPodInformer except like the path and the listFn

so probably we only need one method to rule them all ?

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -547,6 +676,19 @@ export class ContextsManager {
   }
 
   public getCurrentContextResources(resourceName: ResourceName): KubernetesObject[] {
-    return this.states.getCurrentContextResources(this.kubeConfig.currentContext, resourceName);
+    if (!this.currentContext) {
+      return [];
+    }
+    if (this.states.hasInformer(this.currentContext, resourceName)) {
+      console.debug(`informer for ${resourceName} in ${this.currentContext} running`);
```
benoitf: as said previously I think that informer is for the programmer but as a end-user I would expect more what it does like

`Already watching ${resourceName} of ${context}`

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -547,6 +676,19 @@ export class ContextsManager {
...
+      return this.states.getCurrentContextResources(this.kubeConfig.currentContext, resourceName);
+    }
+    if (!this.states.isReachable(this.currentContext)) {
+      console.debug(`context ${this.currentContext} is not reachable. Not trying to start an informer`);
```
benoitf: Skipping watch of context as it is not reachable

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -547,6 +676,19 @@ export class ContextsManager {
...
+      return [];
+    }
+    console.debug(`informer for ${resourceName} in ${this.currentContext} not found. Starting it...`);
```
benoitf: Start watching resource...

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -52,12 +53,20 @@ interface ContextState {
   resources: ContextStateResources;
 }
 
-// All resources managed by podman desktop
+// A selection of resources, to indicate the 'general' status of a context
+const selectedResources = ['pods', 'deployments'] as const;
```
benoitf, line 55: I think this is the first time someone protect its array in Podman Desktop codebase 🎉

packages/main/src/plugin/kubernetes-context-state.ts
benoitf: For example why ContextInternalState is not a Map ?

packages/renderer/src/stores/kubernetes-contexts-state.ts
```diff
@@ -57,3 +57,21 @@ export const kubernetesCurrentContextDeploymentsFiltered = derived(
...
+});
+
+export const serviceSearchPattern = writable('');
+
+// The services in the current context, filtered with `serviceSearchPattern`
+export const kubernetesCurrentContextServicesFiltered = derived(
+  [serviceSearchPattern, kubernetesCurrentContextServices],
+  ([$searchPattern, $services]) =>
+    $services.filter(service => findMatchInLeaves(service, $searchPattern.toLowerCase())),
+);
```
benoitf, line 77: I think that here as well all the stores are copy/paste so probably need to have a common method

```
const magic = magicMethod('...');
const kubernetesCurrentContextServices = magic.store;
const kubernetesCurrentContextServicesFiltered = magic.filteredStore;
```

benoitf (approved):
most of the comments can be addressed as follow-up

I would mainly target the log first if possible

packages/main/src/plugin/kubernetes-context-state.ts
feloy: I tried to have only one method, but the differences are in the details, and I'm afraid the result would be unreadable. I would prefer to keep them as is

feloy:
> most of the comments can be addressed as follow-up
> 
> I would mainly target the log first if possible

I have made changes, expect for having common methods. I would prefer consider doing this later, once the features are stabilized

Thanks for your reviews

deboer-tim (approved):
I agree with @benoitf's comments that there may be another pattern or two we can extract here (esp. in the store). Also seems like we should also be able to replace setStateAndDispatch()'s sendGeneral + {} parameters with sending the resource name and using isSecondaryResourceName()?

But above all, really nice work on secondary informers and responding to comments here. We can continue to improve, but this is a big step & great base to build on.

PR #6212, opened 2024-02-29
feat(extension-api): adding stats container api

### LINKED ISSUE #6211: feat(extension-api): adding support for statsContainer

Extension cannot access container stats

A `statsContainer` method, to get a stream of stats.

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Adding support for `statsContainer` in the extension-api. Allowing extensions to get streams stats for a running container.

See for documentation preview [netlify.app](https://65e0a3985eac521501f9b295--podman-desktop-pr.netlify.app/api/namespaces/containerEngine/functions/statsContainer)

Fixes https://github.com/containers/podman-desktop/issues/6211

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2655,19 +2655,19 @@ declare module '@podman-desktop/api' {
     title: string;
   }
 
-  interface PidsStats {
+  export interface PidsStats {
     current?: number;
     limit?: number;
   }
 
-  interface BlkioStatEntry {
+  export interface BlkioStatEntry {
     major: number;
     minor: number;
     op: string;
     value: number;
   }
 
-  interface BlkioStats {
+  export interface BlkioStats {
     io_service_bytes_recursive: BlkioStatEntry[];
     io_serviced_recursive: BlkioStatEntry[];
     io_queue_recursive: BlkioStatEntry[];
@@ -2678,21 +2678,21 @@ declare module '@podman-desktop/api' {
     sectors_recursive: BlkioStatEntry[];
   }
 
-  interface StorageStats {
+  export interface StorageStats {
     read_count_normalized?: number;
     read_size_bytes?: number;
     write_count_normalized?: number;
     write_size_bytes?: number;
   }
 
-  interface StorageStats {
+  export interface StorageStats {
     read_count_normalized?: number;
     read_size_bytes?: number;
     write_count_normalized?: number;
     write_size_bytes?: number;
   }
 
-  interface NetworkStats {
+  export interface NetworkStats {
     [name: string]: {
       rx_bytes: number;
       rx_dropped: number;
@@ -2707,7 +2707,7 @@ declare module '@podman-desktop/api' {
     };
   }
 
-  interface MemoryStats {
+  export interface MemoryStats {
     // Linux Memory Stats
     stats: {
       total_pgmajfault: number;
@@ -2751,20 +2751,20 @@ declare module '@podman-desktop/api' {
     privateworkingset?: number;
   }
 
-  interface CPUUsage {
+  export interface CPUUsage {
     percpu_usage: number[];
     usage_in_usermode: number;
     total_usage: number;
     usage_in_kernelmode: number;
   }
 
-  interface ThrottlingData {
+  export interface ThrottlingData {
     periods: number;
     throttled_periods: number;
     throttled_time: number;
   }
 
-  interface CPUStats {
+  export interface CPUStats {
     cpu_usage: CPUUsage;
     system_cpu_usage: number;
     online_cpus: number;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2456,17 +2456,6 @@ declare module '@podman-desktop/api' {
     }>;
   }
 
-  export interface StatsContainerOptions {
-    /**
-     * The engine where the container is deployed
-     */
-    engineId: string;
-    /**
-     * The identifier of the container
-     */
-    id: string;
-  }
-
   export interface ContainerCreateOptions {
     /**
      * Assign the specified name to the container. Must match the regular expression`/?[a-zA-Z0-9][a-zA-Z0-9_.-]+`. If not speficied, the platform assigns a unique name to the container
@@ -3078,7 +3067,8 @@ declare module '@podman-desktop/api' {
     /**
      * Get the streamed stats of a running container.
      *
-     * @param options the options to use {@link StatsContainerOptions}
+     * @param engineId the id of the engine managing the container, obtained from the result of {@link containerEngine.listContainers}
+     * @param id the id or name of the container on this engine, obtained from the result of {@link containerEngine.listContainers} or as the result of {@link containerEngine.createContainer}
      * @param callback the function called when container stats info are emitted.
      *
      * @return A Promise resolving a {@link Disposable} that unregister the callback when called.
@@ -3086,10 +3076,7 @@ declare module '@podman-desktop/api' {
      * @example
      * Here is a usage example
      * ```ts
-     * const disposable = await statsContainer({
-     *  engineId: 'engineId',
-     *  id: 'containerId',
-     * }, (stats: ContainerStatsInfo): void => {
+     * const disposable = await statsContainer('engineId', 'containerId', (stats: ContainerStatsInfo): void => {
      *  console.log('CPU Usage', stats.cpu_stats.cpu_usage.total_usage);
      * });
      *
@@ -3098,7 +3085,8 @@ declare module '@podman-desktop/api' {
      * ```
      */
     export function statsContainer(
-      options: StatsContainerOptions,
+      engineId: string,
+      id: string,
       callback: (stats: ContainerStatsInfo) => void,
     ): Promise<Disposable>;
 
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
     title: string;
   }
 
+  interface PidsStats {
+    current?: number;
+    limit?: number;
+  }
+
+  interface BlkioStatEntry {
+    major: number;
+    minor: number;
+    op: string;
+    value: number;
+  }
+
+  interface BlkioStats {
+    io_service_bytes_recursive: BlkioStatEntry[];
+    io_serviced_recursive: BlkioStatEntry[];
+    io_queue_recursive: BlkioStatEntry[];
+    io_service_time_recursive: BlkioStatEntry[];
+    io_wait_time_recursive: BlkioStatEntry[];
+    io_merged_recursive: BlkioStatEntry[];
+    io_time_recursive: BlkioStatEntry[];
+    sectors_recursive: BlkioStatEntry[];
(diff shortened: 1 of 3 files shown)


lstocchi (approved):
LGTM

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2900,6 +3031,32 @@ declare module '@podman-desktop/api' {
...
+     *
+     * // When no longer needed
+     * disposable.dispose();
+     * ```
+     */
+    export function statsContainer(
+      engineId: string,
+      id: string,
+      callback: (stats: ContainerStatsInfo) => void,
+    ): Promise<Disposable>;
```
benoitf, line 3084: for new API I think it would be nice to use options field so we'll have more flexibility

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1041,6 +1041,18 @@ export class ExtensionLoader {
...
+        containerId: string,
+        callback: (stats: containerDesktopAPI.ContainerStatsInfo) => void,
+      ): Promise<Disposable> {
+        const identifier = await containerProviderRegistry.getContainerStats(engineId, containerId, callback);
+        return Disposable.create(() => {
+          containerProviderRegistry.stopContainerStats(identifier).catch((err: unknown): void => {
+            console.error('Something went wrong while trying to stop container stats', err);
+          });
+        });
+      },
```
benoitf, line 1054: I think it might be the first call doing that, but I'm wondering if we should not include by default this disposable as part of the extensionContext subscriptions so even if developer forget to add it, it'll be auto-removed on extension deactivation

benoitf: (not required to merge)

packages/extension-api/src/extension-api.d.ts
benoitf: (required to merge)

axel7083: I am open to the suggestion and I am willing to change it if you really want to. However, in the current api, this would feel weird as most of the apis related to a specific container are using the `engineId: string, id: string`.

- `inspectContainer(engineId: string, id: string)`
- `startContainer(engineId: string, id: string)`
- `logsContainer(engineId: string, id: string, callback: (name: string, data: string) => void)`
- `stopContainer(engineId: string, id: string)`
- `deleteContainer(engineId: string, id: string)`
- `saveImage(engineId: string, id: string, filename: string)`
etc.

Making the statsContainer looking like `statsContainer(options: {engineId: string, id: string}, callback: (...) => void)` would in my opinion make it really out of place, as it would not match the other methods. 

I am in favour of using options when optional, multiple arguments can be used. But here the statsContainer would, in the same maner than logsContainer, or stopContainer probably _never?_ have something different.

I would prefer keeping the api as logic, and easy for the developer to understand, preferably by keeping the same  pattern.

benoitf: but we should add new methods for the other calls with options and deprecate engineId/id so I would avoid to introduce something to deprecate it

It might be easier for usecase to provide the providerConnection rather than the engineId

axel7083: I don't understand how you would mix the providerConnection argument with the `engineId`, `containerId`.

You would want something like the following ?
```
statsContainer(options: {engineId?: string, provider?: ContainerProviderConnection,  id: string}, callback: (...) => void)
```

In terms of arguments, offering either the engineId, or the provider feel like I could provide none, and I would still get the result, do we want (later) to support that ?

benoitf: when you asked to build an image you asked to be able to omit the provider so it could take the first one in the list
so I don't see why here provider could not be optional as well

axel7083: I updated with your proposition

axel7083: @benoitf tell me if I can merge

benoitf:
@axel7083 I'm not yet convinced (including my proposal)

I see two issues :
1. Options is not optional there so is it a valid pattern to call something an option if it is mandatory ?
2. engineId is required it means we can't add provider being optional as well

Then, should the id be a first parameter and then engineId should be an optional field in the option

or we have a source/provider field being required being of type string or ProviderConnection ( the else could be done later)

Or we have like other API engine options with engineId field and then provider options with provider being required

axel7083:
> @axel7083 I'm not yet convinced (including my proposal)
> 
> I see two issues :
> 
> 1. Options is not optional there so is it a valid pattern to call something an option if it is mandatory ?

No, literally an option feel like it is optional, does not make sense.

> 2. engineId is required it means we can't add provider being optional as well

Yes. This is what I tried to explain, we could, when adding the provider property make the engineId optional, but what do we do when no engineId nor provider is provided ? We raise an error ? bad pattern in my opinion.

> Then, should the id be a first parameter and then engineId should be an optional field in the option
> 
> or we have a source/provider field being required being of type string or ProviderConnection ( the else could be done later)
> 
> Or we have like other API engine options with engineId field and then provider options with provider being required

I am open to the discussio, _but_ currently all the apis are using engineId when we _have_  the resource. Maybe for 1.8 let's keep the same, and have a deeper conversation on the matter, and how we want to improve the relation for extensions with engines and providers ?

benoitf:
ok for the method you added first

axel7083:
> ok for the method you added first

last commit reverted ✅

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+  }
+
+  interface CPUStats {
+    cpu_usage: CPUUsage;
+    system_cpu_usage: number;
+    online_cpus: number;
+    throttling_data: ThrottlingData;
+  }
+
+  export interface ContainerStatsInfo {
```
benoitf, line 2767: I see that we export some interfaces and some we don't

I noticed that we have a lot of non-exported interface in the d.ts

not for this PR but should we export all interfaces by default ?

benoitf (approved):
thanks for your time handling all the requests 👍

packages/extension-api/src/extension-api.d.ts
axel7083: I think that yes we should, does not make really much sense not to export them

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
     title: string;
   }
 
+  interface PidsStats {
```
axel7083:
```suggestion
  export interface PidsStats {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+    current?: number;
+    limit?: number;
+  }
+
+  interface BlkioStatEntry {
```
axel7083:
```suggestion
  export interface BlkioStatEntry {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+    major: number;
+    minor: number;
+    op: string;
+    value: number;
+  }
+
+  interface BlkioStats {
```
axel7083:
```suggestion
  export interface BlkioStats {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+    io_serviced_recursive: BlkioStatEntry[];
+    io_queue_recursive: BlkioStatEntry[];
+    io_service_time_recursive: BlkioStatEntry[];
+    io_wait_time_recursive: BlkioStatEntry[];
+    io_merged_recursive: BlkioStatEntry[];
+    io_time_recursive: BlkioStatEntry[];
+    sectors_recursive: BlkioStatEntry[];
+  }
+
+  interface StorageStats {
```
axel7083:
```suggestion
  export interface StorageStats {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+    read_count_normalized?: number;
+    read_size_bytes?: number;
+    write_count_normalized?: number;
+    write_size_bytes?: number;
+  }
+
+  interface StorageStats {
```
axel7083:
```suggestion
  export interface StorageStats {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+    read_count_normalized?: number;
+    read_size_bytes?: number;
+    write_count_normalized?: number;
+    write_size_bytes?: number;
+  }
+
+  interface NetworkStats {
```
axel7083:
```suggestion
  export interface NetworkStats {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+      tx_bytes: number;
+      tx_dropped: number;
+      tx_errors: number;
+      tx_packets: number;
+      endpoint_id?: string; // not used on linux
+      instance_id?: string; // not used on linux
+    };
+  }
+
+  interface MemoryStats {
```
axel7083:
```suggestion
  export interface MemoryStats {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+    failcnt: number;
+    limit: number;
+
+    // Windows Memory Stats
+    commitbytes?: number;
+    commitpeakbytes?: number;
+    privateworkingset?: number;
+  }
+
+  interface CPUUsage {
```
axel7083:
```suggestion
  export interface CPUUsage {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+    percpu_usage: number[];
+    usage_in_usermode: number;
+    total_usage: number;
+    usage_in_kernelmode: number;
+  }
+
+  interface ThrottlingData {
```
axel7083:
```suggestion
  export interface ThrottlingData {
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -2655,6 +2655,137 @@ declare module '@podman-desktop/api' {
...
+    periods: number;
+    throttled_periods: number;
+    throttled_time: number;
+  }
+
+  interface CPUStats {
```
axel7083:
```suggestion
  export interface CPUStats {
```

packages/extension-api/src/extension-api.d.ts
axel7083: I added all the exports ! Thanks for noticing

feloy (approved):
tested from an extension with the provided example, works great, thanks

PR #6230, opened 2024-03-01
feat(extension-api): allow to get ImageInspectInfo from the ContainerProviderConnection

### LINKED ISSUE #6229: feat(extension-api): allow to get ImageInspectInfo using the ContainerProviderConnection

When dealing with the extension api, we often faces some troubles when getting images details for a specific provider.

Adding a signature for the `getImageInspect` taking ContainerProviderConnection as argument.

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Allowing extensions to get an image inspect specifically for a ContainerProviderConnection. This allows the extension, to let them choose the connection provider they want to use.

#### Reminder

A provider is registered by an extension, it can then register **several** engines. E.g. on windows, a provider can register multiple Podman machine, each will have its own `engineId`.

#### Details 

With the current extension-api, it is difficult to keep a link between `ProviderContainerConnection` and `engine id(s)`. When we have multiple providers, we might want to create a container a on specific one, with the extensions we are able to get them `provider.getContainerConnections()`, but we are not able to get an image inspect for a specific one.

Let's take the `getImageInspect` method signature: `getImageInspect(engineId: string, id: string)`.

We requires to have the `engineId` and the `id` of the image, however, the only way to get those informations are through the `listImages()` but we cannot filter on a specific **provider**. 

This PR is adding the `getImageInspect(ContainerProviderConnection, imageName: string)` to get the `ImageInspecInfo` for an image in a specific ContainerProviderConnection.

#### Avoid conflicts

We were using `pullImage(ContainerProviderConnection, imageName: string)` allowing to pull an image in a specific provider, **but this does not garantee** that when we will `listImages()` we will not get already pulled images from another provider.

Fixes https://github.com/containers/podman-desktop/issues/6229

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -3031,11 +3031,26 @@ declare module '@podman-desktop/api' {
      *
      * @param containerProviderConnection the connection to the local engine
      * @param imageName the image name
+     * @return A promise resolving to {@link ImageInspectInfo} or undefined if the imageName is not found locally.
+     *
+     * @example
+     *
+     * Here is an example
+     * ```ts
+     * // Get the first started provider
+     * const provider = provider.getContainerConnections().find(connection => connection.connection.status() === 'started');
+     * // Get the image inspect
+     * const imageInspectInfo = await containerEngine.findImageInspect(provider, 'hello-world:latest');
+     * // If undefined, we pull the image
+     * if(imageInspectInfo === undefined) {
+     *    await containerEngine.pullImage(provider, 'hello-world:latest', () => {});
+     * }
+     * ```
      */
-    export function getImageInspect(
+    export function findImageInspect(
       containerProviderConnection: ContainerProviderConnection,
       imageName: string,
-    ): Promise<ImageInspectInfo>;
+    ): Promise<ImageInspectInfo | undefined>;
 
     export function info(engineId: string): Promise<ContainerEngineInfo>;
     export const onEvent: Event<ContainerJSONEvent>;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -3026,6 +3026,17 @@ declare module '@podman-desktop/api' {
      */
     export function getImageInspect(engineId: string, id: string): Promise<ImageInspectInfo>;
 
+    /**
+     * Return low-level information about an image using a {@link ContainerProviderConnection}
+     *
+     * @param containerProviderConnection the connection to the local engine
+     * @param imageName the image name
+     */
+    export function getImageInspect(
+      containerProviderConnection: ContainerProviderConnection,
+      imageName: string,
+    ): Promise<ImageInspectInfo>;
+
     export function info(engineId: string): Promise<ContainerEngineInfo>;
     export const onEvent: Event<ContainerJSONEvent>;
 
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -544,25 +544,26 @@ export class ContainerProviderRegistry {
     return flatttenedContainers;
   }
 
+  private async getImageInfo(provider: InternalContainerProvider): Promise<ImageInfo[]> {
+    try {
+      if (!provider.api) {
+        return [];
+      }
+      const images = await provider.api.listImages({ all: false });
+      return images.map(image => {
+        const imageInfo: ImageInfo = { ...image, engineName: provider.name, engineId: provider.id };
+        return imageInfo;
+      });
+    } catch (error) {
+      console.log('error in engine', provider.name, error);
+      return [];
+    }
+  }
+
   async listImages(): Promise<ImageInfo[]> {
-    let telemetryOptions = {};
+    const telemetryOptions = {};
     const images = await Promise.all(
-      Array.from(this.internalProviders.values()).map(async provider => {
-        try {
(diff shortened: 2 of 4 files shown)


packages/main/src/plugin/container-registry.ts
```diff
@@ -541,25 +541,26 @@ export class ContainerProviderRegistry {
...
-            return [];
-          }
-          const images = await provider.api.listImages({ all: false });
-          return images.map(image => {
-            const imageInfo: ImageInfo = { ...image, engineName: provider.name, engineId: provider.id };
-            return imageInfo;
-          });
-        } catch (error) {
-          console.log('error in engine', provider.name, error);
-          telemetryOptions = { error: error };
```
lstocchi, line 562: we're losing this error reporting

packages/extension-api/src/extension-api.d.ts
```diff
@@ -3019,6 +3019,17 @@ declare module '@podman-desktop/api' {
...
+    /**
+     * Return low-level information about an image using a {@link ContainerProviderConnection}
+     *
+     * @param containerProviderConnection the connection to the local engine
+     * @param imageName the image name
+     */
+    export function getImageInspect(
+      containerProviderConnection: ContainerProviderConnection,
+      imageName: string,
+    ): Promise<ImageInspectInfo>;
```
benoitf: I think it should return undefined if the image does not exist

Error is if we're not able to grab image or something is invalid but else IMHO it should return undefined if there is no match

axel7083: That a question I asked myself too, but we are making a call similar to a resource request here, we are using the `get` wording, and not `find`. 

So I feel like we should not make the return undefined a valid return, but throw the error not find. The `getImageInspect(engineId, imageId)` is throwing an error when we use _unknown_ imageId, and I have a preference for keeping this behavior.

> I can update the docs, giving examples, and details for typedoc

benoitf: for getImageInspect(engineId, imageId) usually it should always return a valid image, it's only if you give an invalid imageID that you have an error so you're making an error from the client side

but for the parameter being the image name I would say it's expected to have undefined if there is no match
error would in case we're not able to perform a matching operation

axel7083: I would not see much difference to consider the image name as an identifier at the level than the imageId. 

IMO, when dealing with promises, I always prefer having a catch block, than an if != undefined when dealing with such methods. 

I would be willing to change if other people are in favor (or simply no one care, and you really have this preference) of having undefined added to the return signature. 

> I would also prefer having it merge before 1.8 👀 

@lstocchi @feloy an opinion/preference ?

packages/main/src/plugin/container-registry.ts
axel7083: Yes! Nice catch

packages/extension-api/src/extension-api.d.ts
benoitf: on my side it's a requirement to have `Promise<ImageInspectInfo | undefined>`

axel7083: > on my side it's a requirement to have `Promise<ImageInspectInfo | undefined>`

Ok. I cannot set the return type of the same method name to something different (`Promise<ImageInspectInfo>` and `Promise<ImageInspectInfo | undefined>`, I am getting typescript errors.

To solve this issue, I updated and make a method `findImageInspect`, at the end, I think this is more appropriate.

benoitf (changes requested):
please use getImageInspect and not find (also find usually return arrays)

I don't see an issue on the return type as you have one defined in the extension API and one inside the internal API

implementation can have different method name; no need to use a single method


also the extensionLoader code should not have any logic, just delegate to the container registry code

axel7083:
_Monday, new day, let's try to clarify what you expect, as I am a bit lost 🏝️._

> please use getImageInspect and not find (also find usually return arrays)

I cannot have two method with the name same and the different return type. The following are raising Typecheck errors.

```ts
getImageInspect: async (
        engineId: string,
        id: string,
): Promise<ImageInspectInfo>
getImageInspect: async (
        connection: containerDesktopAPI.ContainerProviderConnection,
       imageName: string,
): Promise<ImageInspectInfo | undefined>
```

See https://github.com/containers/podman-desktop/commit/28cafdc5474e44b46e6d48f9add68f7781994474#commitcomment-139263327 for implementation tentative.

> 
> I don't see an issue on the return type as you have one defined in the extension API and one inside the internal API
> 
> implementation can have different method name; no need to use a single method

Okey, I though `find` verb would have been a good alternative (_find: get the first item matching the condition (imageName: string) and return it, otherwise return undefined._), but I do not understand the issue with this ? 

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#return_value
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/find

**Question**: Does the possibility of having multiple corresponding ImageInspectInfo for an imageName is the problem ? Should we not instead make a `getImages(connection, imageName): Promise<ImageInfo>` method instead ?

> also the extensionLoader code should not have any logic, just delegate to the container registry code

No problem, I will move the logic inside the container registry when I will understand the expected implementation details!

axel7083:
Closing in favor of an alternative solution https://github.com/containers/podman-desktop/issues/6256

PR #6239, opened 2024-03-02
[poc] Add layout components

feloy:
### What does this PR do?

- Add a HorizontalCard component, and use it for  PreferencesCliTool and PreferencesResourcesRendering components
- Add a Card component, and use it in different places
- Add a CardTitle component, and use it in Card and HorizontalCard
- HorizontalCard, Card and CardTitle support dark/light mode (except icon)

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -341,9 +341,24 @@ export class ColorRegistry {
       light: colorPalette.gray[300],
     });
 
+    this.registerColor(`${invCt}card-bg-highlighted`, {
+      dark: colorPalette.charcoal[700],
+      light: colorPalette.gray[200],
+    });
+
+    this.registerColor(`${invCt}card-divide`, {
+      dark: colorPalette.gray[900],
+      light: colorPalette.charcoal[50],
+    });
+
     this.registerColor(`${invCt}card-header-text`, {
       dark: colorPalette.white,
       light: colorPalette.charcoal[900],
+    });
+
+    this.registerColor(`${invCt}card-subtitle-text`, {
+      dark: colorPalette.gray[600],
+      light: colorPalette.charcoal[100],
     });
 
     this.registerColor(`${invCt}card-text`, {
--- a/packages/renderer/src/lib/dashboard/DashboardPage.svelte
+++ b/packages/renderer/src/lib/dashboard/DashboardPage.svelte
@@ -97,10 +97,14 @@ function getInitializationContext(id: string): InitializationContext {
             <ProviderStopped provider="{providerStopped}" />
           {/each}
         {/if}
-        <LearningCenter />
       </div>
-      <div class="px-2 space-y-5 h-full">
-        <Card title="Featured Extensions">
+      <div class="px-2 mt-5 h-full">
+        <Card title="Learning center">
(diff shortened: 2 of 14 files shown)


packages/main/src/plugin/color-registry.ts
```diff
@@ -341,11 +341,26 @@ export class ColorRegistry {
...
+    this.registerColor(`${invCt}card-bg-highlighted`, {
+      dark: colorPalette.charcoal[700],
+      light: colorPalette.gray[200],
+    });
+
+    this.registerColor(`${invCt}card-divide`, {
+      dark: colorPalette.gray[900],
+      light: colorPalette.charcoal[50],
+    });
+
```
benoitf, line 353: Look nice to introduce new components

but you can't introduce new components/colors "like that"

the colors are defined in the Podman Desktop design system

And there is not yet these colors or these components

if you need to extract values, it should be defined before the implementation

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -96,8 +97,18 @@ function getInitializationContext(id: string): InitializationContext {
...
+        <Card title="Learning center">
+          <svelte:fragment slot="content">
+            <LearningCenter />
+          </svelte:fragment>
+        </Card>
+        <Card title="Featured extensions">
+          <svelte:fragment slot="content">
+            <FeaturedExtensions />
+          </svelte:fragment>
+        </Card>
```
benoitf, line 111: I would be really cautious there

For example Learning center does not match the expected design
see https://github.com/containers/podman-desktop/issues/6221

packages/renderer/src/lib/preferences/PreferencesExtensionList.svelte
```diff
@@ -77,51 +78,54 @@ async function updateExtension(extension: ExtensionInfo, ociUri: string) {
...
-        <div class="container w-full flex-col">
+    <Card title="Featured Extensions">
+      <svelte:fragment slot="content">
+        <FeaturedExtensions />
+      </svelte:fragment>
+    </Card>
+
+    <Card title="Install a new extension from OCI Image" label="OCI image installation box">
+      <svelte:fragment slot="content">
+        <div class="flex flex-col w-full">
```
benoitf, line 89: this screen is being changed as part of https://github.com/containers/podman-desktop/issues/2170
so I'm not sure there should be changes for now in that page

benoitf (commented):
introducing component is looking a good enhancement but it should be done by UX team before

It should define the colors, how it looks, etc.

benoitf:
What is the rationale of components and the slots

for example are we not missing CardHeader, CardBody components ?

axel7083:
We might want take some inspiration from existing library and implementation 
- https://www.shadcn-svelte.com/docs/components/card
- https://github.com/huntabyte/shadcn-svelte/tree/main/apps/www/src/lib/registry/default/ui/card

Some remarks
- The content of the card (horitontal, vertical), should not be choosed by a card component, it should be the responsability of the component definiting the content, so if I want to make it a grid, a flex, of anything else I am not bounded by the component choices.
- Use its own folder in the ui library for `Card` (see example in github repo linked above)

PR #6240, opened 2024-03-02
chore: button hidden and icon-only spacing

deboer-tim:
### What does this PR do?

Yes, two things shouldn't be in the same PR, but they are both very minor:
- The Button does not have a hidden state, which will be required for the Run Image form.
- When a Button has no text, it is noticeably smaller vertically than other buttons. This adds some minor padding to bring the button to within 0.5px.

### Screenshot / video of UI

Before:

Screenshot 2024-03-02 at 7 42 36 AM

After:

Screenshot 2024-03-02 at 7 42 10 AM

N/A

### How to test this PR?

Use checkboxes on one of the list pages to show the delete button.

### DIFF
--- a/packages/renderer/src/lib/ui/Button.spec.ts
+++ b/packages/renderer/src/lib/ui/Button.spec.ts
@@ -133,3 +133,13 @@ test('Check selected tab button styling', async () => {
   expect(button).toHaveClass('text-white');
   expect(button).toHaveClass('border-purple-500');
 });
+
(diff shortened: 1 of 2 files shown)


benoitf (commented):
it's missing tests (for the new hidden property)

deboer-tim:
Added test for hidden property.

benoitf (approved):

Issue #6199, opened 2024-02-29
Consistent select controls


deboer-tim:
### Is your enhancement related to a problem? Please describe

There are 14 <select> controls, which have copies of their own styling with some slight variations.

### Describe the solution you'd like

Create a Select component with styling similar to the Input component, and replace all current usage with it.

deboer-tim:
https://github.com/containers/podman-desktop/assets/19958075/57fb4d25-319c-4829-894d-59ca4664feeb

ekidneyrh:
I've done mock-ups for the light-mode versions for these components as part of https://github.com/containers/podman-desktop/issues/6929:
[+/- Button](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#--button) 
[Dropdown theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#dropdown)

PR #6241, opened 2024-03-02
chore: dropdown component

deboer-tim:
### What does this PR do?

Creates a new Dropdown component that is just a styled select control, so that we don't have this styling/slight variations in several places, and one less thing to fix for light-mode. Styling uses the same variable naming pattern as an Input.

I used the new component in three places to start so that there's something to test: deploying a pod to Kube, creating a pod from containers, and enum items in preferences. There are 11 other instances of `<select>` that will need to be switched in future PRs.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/19958075/2e265bd5-08ac-42e8-b305-7899f57bf80a

First step of #6199.

### How to test this PR?

Unit test included; go to the places listed in the UI to test styling/behaviour.

### DIFF
--- a/packages/renderer/src/lib/ui/Dropdown.svelte
+++ b/packages/renderer/src/lib/ui/Dropdown.svelte
@@ -1,0 +1,31 @@
+<script lang="ts">
+export let id: string | undefined = undefined;
+export let name: string | undefined = undefined;
+export let value: unknown;
+export let readonly: boolean = false;
+export let disabled: boolean = false;
+export let required: boolean = false;
+export let error: boolean = false;
+
+let enabled: boolean = true;
+$: enabled = !readonly && !disabled;
+</script>
+
+<select
+  class="w-full p-1 outline-none bg-[var(--pd-dropdown-bg)] border-[1px] border-transparent text-sm"
+  class:border-b-[var(--pd-dropdown-stroke)]="{enabled && !error}"
+  class:border-b-[var(--pd-dropdown-stroke-error)]="{enabled && error}"
+  class:border-b-[var(--pd-dropdown-stroke-readonly)]="{!enabled}"
+  class:text-[color:var(--pd-dropdown-focused-text)]="{enabled}"
+  class:text-[color:var(--pd-dropdown-disabled-text)]="{!enabled}"
+  class:hover:border-b-[var(--pd-dropdown-hover-stroke)]="{enabled && !error}"
+  class:hover:border-b-[var(--pd-dropdown-stroke-error)]="{enabled && error}"
+  class:hover:bg-[var(--pd-dropdown-hover-bg)]="{enabled}"
+  id="{id}"
+  name="{name}"
+  required="{required}"
+  bind:value="{value}"
+  aria-label="{$$props['aria-label']}"
+  aria-invalid="{$$props['aria-invalid']}">
+  <slot />
+</select>
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -230,6 +230,7 @@ export class ColorRegistry {
     this.initInvertContent();
     this.initCardContent();
     this.initInputBox();
+    this.initDropdown();
   }
 
   protected initGlobalNav(): void {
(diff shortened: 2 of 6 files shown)


packages/renderer/src/lib/ui/Select.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
```
benoitf: hello, I was wondering if we should not call the component `Dropdown` as I think most of UI libraries are using that word (and it seems UX people are using that term as well)

packages/renderer/src/lib/ui/Select.svelte
```diff
@@ -0,0 +1,32 @@
...
+<select
+  class="w-full p-1 outline-none bg-transparent border-[1px] border-transparent text-sm"
+  class:border-b-[var(--pd-input-field-stroke)]="{enabled && !error}"
+  class:border-b-[var(--pd-input-field-stroke-error)]="{enabled && error}"
+  class:border-b-[var(--pd-input-field-stroke-readonly)]="{!enabled}"
+  class:text-[color:var(--pd-input-field-focused-text)]="{enabled}"
+  class:text-[color:var(--pd-input-field-disabled-text)]="{!enabled}"
+  class:hover:border-[var(--pd-input-field-stroke)]="{enabled && !error}"
+  class:hover:border-[var(--pd-input-field-stroke-error)]="{enabled && error}"
+  class:hover:bg-[var(--pd-input-field-hover-bg)]="{enabled}"
```
benoitf: I would not grab the colors of the input component. I think it'll make us problems later

Of course the final color might be the same but I think people should be able to override specific colors

packages/renderer/src/lib/ui/Select.svelte
benoitf: Also component colors need to be defined in the Podman Desktop design system made by  @ekidneyrh

deboer-tim: Sure, I can rename.

packages/renderer/src/lib/ui/Select.svelte
deboer-tim: After thinking about it, yes, you're right. I am worried that if there are (e.g.) 17 different variables to change to make purple into blue, nobody will do it. I'll follow up via the UX meeting.

packages/renderer/src/lib/ui/Select.svelte
deboer-tim: Renamed. Added proposed component colors for the design system following the identical pattern to Input, and will review with @ekidneyrh and update if necessary before moving out of draft.

packages/renderer/src/lib/ui/Select.svelte
deboer-tim: Switching to using dropdown variables - identical pattern to Input, but having it's own colors.

deboer-tim:
Rebased, renamed to Dropdown, updated to current Input styling, updated PR description, and added proposed dropdown color variables for UX review.

ekidneyrh:
I'm using Fedora Linux and the dropdown options are the same colour as the bg making it illegible. 
[Screencast from 2024-03-21 12-10-16.webm](https://github.com/containers/podman-desktop/assets/87311656/619fade5-23d2-490a-97b3-4d70d6b9b4c2)

packages/renderer/src/lib/ui/Dropdown.svelte
```diff
@@ -0,0 +1,31 @@
...
+export let readonly: boolean = false;
+export let disabled: boolean = false;
+export let required: boolean = false;
+export let error: boolean = false;
+
+let enabled: boolean = true;
+$: enabled = !readonly && !disabled;
+</script>
+
+<select
```
axel7083, line 14: IMO We should not use `<select>` as it is a _bad_[^1] html element, with very hard customization, limiting strongly its usage :(


[^1]: This is opinionated, based on experience on many projects where select was always annoying, and the result was always using a custom component, ([react-select](https://react-select.com/home), [svelte-select](https://github.com/rob-balfre/svelte-select), [vue-select](https://github.com/sagalbot/vue-select) etc.) all having for the react 4M download/week.

deboer-tim:
Yeah, that's why this got stuck in draft. I'll close it for now until we have a plan for a new component.

jeffmaury:
@deboer-tim should it be closed ?

PR #6242, opened 2024-03-02
chore: switch ingress & routes to new informer

deboer-tim:
### What does this PR do?

Provides ingress and routes informers using the same pattern as the deployment and services informers, then switches the corresponding page to use these too.

The pattern is identical to existing code, with the exception of the context-state tests where I used describe.each() to iterate over all secondary informers instead of maintaining separate tests for each one. backoffInitialValue was raised to 10000 as it was hitting the limit with multiple informers.

### Screenshot / video of UI

N/A

First part of #6259.

### How to test this PR?

Unit tests updated. Confirm no regression in route/ingress page.

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.spec.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.spec.ts
@@ -17,8 +17,8 @@
  ***********************************************************************/
 
 import { beforeEach, afterEach, expect, test, vi, describe } from 'vitest';
-import { type ContextGeneralState, ContextsStates } from './kubernetes-context-state.js';
-import { ContextsManager } from './kubernetes-context-state.js';
+import type { ResourceName, ContextGeneralState } from './kubernetes-context-state.js';
+import { ContextsManager, ContextsStates } from './kubernetes-context-state.js';
 import type { ApiSenderType } from './api.js';
 import * as kubeclient from '@kubernetes/client-node';
 import type { ErrorCallback, KubernetesObject, ObjectCallback } from '@kubernetes/client-node';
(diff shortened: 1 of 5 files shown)


deboer-tim:
Rebased on latest, test fixed. Ready for review.

feloy:
When switching to a context on a vanilla Kubernetes cluster, with the Route resource not registered, the informer is never able to connect. I think we would need to check if the returned error is about not found resource, or something similar (also, in the same kind, it's possible that we get some permissions error on some clusters for some resources)

```
Trying to watch routes on the kubernetes context named "kind-kind-cluster" but got a connection refused, retrying the connection in 42s. HttpError: HTTP request failed)
```

feloy (approved):
tested with an app deployed on Developer sandbox with deployment+service+route
LGTM

deboer-tim:
> When switching to a context on a vanilla Kubernetes cluster, with the Route resource not registered, the informer is never able to connect. I think we would need to check if the returned error is about not found resource, or something similar (also, in the same kind, it's possible that we get some permissions error on some clusters for some resources)
> 
> ```
> Trying to watch routes on the kubernetes context named "kind-kind-cluster" but got a connection refused, retrying the connection in 42s. HttpError: HTTP request failed)
> ```

Yes. I checked the current behaviour/informer just to compare, and we get the same failure every time you visit the Ingress/Routes page with a generic message: `kubernetesInformerWritable.ts:23 Error when starting the informer - Error: error when setting the informer`. Because retry after initial failure is broken we never try again (e.g. if kind was stopped you need to switch contexts to retry) so this is a one-time error. But if we fixed that, this error would appear every 5s...

Although this PR is the thing that's going to trigger more logging and we need to handle ASAP, I've opened #6263 to track and would prefer to handle it separately, as not switching over results in less log messages, but they are cryptic and due to worse behaviour for users.

cdrage (approved):
LGTM!

Tested against ingress / routes and it is **QUICK**. It deletes very quickly / gets removed and re-made which is great vs the older implementation.

LGTM!

PR #6250, opened 2024-03-04
fix: clear local state for services when informer is stopped

feloy:

### What does this PR do?

Clears the local state when the service informer is stopped, because resources will be received again with the next informer.

Details: https://github.com/containers/podman-desktop/issues/6202#issuecomment-1976172159

### How to test this PR?


deboer-tim (approved):
Code looks good and addresses the issue. Tested against multiple contexts and confirmed the problem is gone. 👍🏼

cdrage (approved):
LGTM!

I tested:
- switching contexts
- switching clusters
- disconnecting internet / reconnecting
- closing / re-opening PD and checking services again.

PR #6251, opened 2024-03-04
chore: fix website

benoitf:
### What does this PR do?
https://github.com/containers/podman-desktop/pull/5880 fixes some issues but it has broken the website build

issue is that we have errors being reported but it's not making the build failing....

it's in the typedoc plug-in

need investigation but in the meantime, restore the website build

### How to test this PR?


benoitf:
merging as podman-desktop.io is not working

PR #6261, opened 2024-03-04
fix: services store register, to restart service informer when switching context

### LINKED ISSUE #6254: subscribe for service/... resource, so informer is started directly when switching context and someone is subscribed
### Bug description

When on the Service page, is the context is switched from this page (using the selector on statusbar for example), the informer is not restarted (before we go to another page and come back to the Service page)

### Operating system

mac os

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Add a registry to the ContextsManager to know if some clients have registered to the Services informers, so we can restart the informer automatically when the context is switched by the user. 

Fixes #6254

### How to test this PR?

go to the Services List page, switch the context from the statusbar (without leaving the Service List page), then create a service using kubectl. The Service List should be updated.

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -374,13 +374,13 @@ export class ContextsManager {
         const contextName = ctx.name;
         await this.states.disposeSecondaryInformers(contextName);
       }
-    }
 
-    // Restart informers for secondary resources if watchers are subscribing for this resource
-    for (const resourceName of secondaryResources) {
-      if (this.secondaryWatchers.hasSubscribers(resourceName)) {
-        console.debug(`start watching ${resourceName} in context ${this.currentContext}`);
-        this.startResourceInformer(this.currentContext, resourceName);
+      // Restart informers for secondary resources if watchers are subscribing for this resource
+      for (const resourceName of secondaryResources) {
+        if (this.secondaryWatchers.hasSubscribers(resourceName)) {
+          console.debug(`start watching ${resourceName} in context ${this.currentContext}`);
+          this.startResourceInformer(this.currentContext, resourceName);
+        }
       }
     }
 
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -302,12 +302,40 @@ export class ContextsStates {
   }
 }
 
+class SecondaryResourceWatchersRegistry {
+  // Map resourceName to number of watchers
+  private watchingSecondaryResource = new Map<string, number>();
+
+  subscribe(resourceName: string): void {
+    let count = this.watchingSecondaryResource.get(resourceName);
+    if (count === undefined) {
+      this.watchingSecondaryResource.set(resourceName, 0);
+      count = 0;
+    }
+    this.watchingSecondaryResource.set(resourceName, count + 1);
+  }
+
+  unsubscribe(resourceName: string): void {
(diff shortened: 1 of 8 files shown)


deboer-tim (changes requested):
I confirmed this fixes the original problem, but when testing it is either causing a regression or exposing another one: if I keep selecting the _same_ current context, the resources get shown multiple times. i.e.:

Old behaviour: if I'm on the Services page and select a different context using the status bar the page is empty until I navigate away and come back; if I re-select the context I'm already on nothing changes.

New behaviour: if I change the context the new services appear correctly; if I re-select the current context then 2 services become 4, 6, 8, etc.

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -332,6 +360,14 @@ export class ContextsManager {
       }
     }
 
+    // Restart informers for secondary resources if watchers are subscribing for this resource
+    for (const resourceName of secondaryResources) {
+      if (this.secondaryWatchers.hasSubscribers(resourceName)) {
```
deboer-tim: Needs a check like in getCurrentContextResources():
`if (this.secondaryWatchers.hasSubscribers(resourceName) && !this.states.hasInformer(this.currentContext, resourceName)) {`
This fixes the regression I saw while testing.

feloy: good catch, pushing a new commit with this fix, thanks

feloy: no, moving into this block instead:
```
    if (contextChanged) {

```

feloy:
I still have problems with my tests, they pass in isolation, but fail when ran with the other ones. Checking...

deboer-tim (approved):
Worked flawlessly in testing, switching back and forth between contexts and making changes.

I really don't like that calling a function like getCurrentContextResources() triggers a background informer, and you have to know to call unregisterGetCurrentContextResources() to unregister it - but the informer was there before and this gives the missing hook to stop it so to me it is most important to get this fix in; I can open an issue or we can discuss later if (e.g.) the functions should be named more like add/remove lIstener.

Needs a rebase since Ingress/Route informers have been merged. kubernetes-contexts-state.ts will need more calls to unregister and mocking like ServiceList.spec.ts might be needed elsewhere or changed.

feloy:
> Worked flawlessly in testing, switching back and forth between contexts and making changes.
> 
> I really don't like that calling a function like getCurrentContextResources() triggers a background informer, and you have to know to call unregisterGetCurrentContextResources() to unregister it - but the informer was there before and this gives the missing hook to stop it so to me it is most important to get this fix in; I can open an issue or we can discuss later if (e.g.) the functions should be named more like add/remove lIstener.
>

Sure, I was also not happy with this name, and the fact that the caller needs to know that it needs to call unregister after this one. After some perspective, I think that registerGetCurrentContextResources would be better. I changed to rename it like this.
 
> Needs a rebase since Ingress/Route informers have been merged. kubernetes-contexts-state.ts will need more calls to unregister and mocking like ServiceList.spec.ts might be needed elsewhere or changed.

Done, thanks

deboer-tim (approved):
Thanks for function name change and rebasing. Still working flawlessly in testing, so re-approving.

cdrage (approved):
Was able to:

1. Switch contexts
2. Create a new service
3. It instantly appears in the list.

LGTM code-wise as well as through practical testing.

PR #6262, opened 2024-03-04
chore: fix linux setup loop

### LINKED ISSUE #6151: Podman Desktop endless setup loop on Linux without Podman Machine
### Bug description

With the addition of the optional podman machine detection, is now being prompted as mandatory.

After the setup has been completed, there is a short pause and then the process starts over again.

#### 1
pd1
#### 2
pd2
#### 3
pd3
#### 4
pd4

The only way to make it go away is to run the `machine init`, but then you also end up with a VM.

It should be possible to run Podman Desktop with the (possibly old) system version of Podman...

pd5
pd6

~There is another bug where it suggests to change the default system connection, to the machine.~ (#6161)

Offering to restart the already running `podman.socket` from the app is another issue, #2903. Or is it machine?

If you enable "Autostart" (it is on by default), it will start both a new daemon and also start the machine at launch.

### Operating system

Ubuntu 20.04

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

cdrage:
chore: fix linux setup loop

### What does this PR do?

When using podman machine on Linux, it is OPTIONAL as Podman by default
is installed natively.

We were incorrectly showing the notification in an endless loop thinking
that Podman was setup (podman machine was missing).

This PR, in updateMachines function:
* Does not notify if podman machine errors out when retrieving the json
  (rare, but it happens)
* If zero podman machines are found, do not show the update
* Do not update the provider status to "installed" (would override the
  "ready" status before), since Linux supports podman natively.

### Screenshot / video of UI

Screenshot from 2024-02-07 17-53-09

(no more notification)

Closes https://github.com/containers/podman-desktop/issues/6151

### How to test this PR?

1. Open up Podman Desktop on Linux
2. You will no longer encounter the endless loop issue on Linux

### DIFF
--- a/extensions/podman/src/extension.spec.ts
+++ b/extensions/podman/src/extension.spec.ts
@@ -1088,3 +1088,29 @@ test('checkDisguisedPodmanSocket: runs updateWarnings when called not on Linux',
   await checkDisguisedPodmanSocket(provider);
   expect(updateWarningsMock).toBeCalled();
 });
+
+test('Even with getJSONMachineList erroring, do not show setup notification on Linux', async () => {
+  vi.mocked(isLinux).mockReturnValue(true);
+  vi.spyOn(extensionApi.process, 'exec').mockRejectedValue({
+    name: 'name',
+    message: 'description',
+    stderr: 'error',
+  });
+  await expect(extension.updateMachines(provider)).rejects.toThrow('description');
+  expect(extensionApi.window.showNotification).not.toBeCalled();
+});
+
+test('If machine list is empty, do not show setup notification on Linux', async () => {
+  vi.mocked(isLinux).mockReturnValue(true);
+  const spyExecPromise = vi.spyOn(extensionApi.process, 'exec');
+  spyExecPromise.mockResolvedValue({ stdout: '[]' } as extensionApi.RunResult);
+  await extension.updateMachines(provider);
+  expect(extensionApi.window.showNotification).not.toBeCalled();
+});
+
+test('if there are no machines, make sure checkDefaultMachine is not being ran inside updateMachines', async () => {
+  const spyCheckDefaultMachine = vi.spyOn(extension, 'checkDefaultMachine');
+  vi.spyOn(extensionApi.process, 'exec').mockResolvedValue({ stdout: '[]' } as extensionApi.RunResult);
+  await extension.updateMachines(provider);
+  expect(spyCheckDefaultMachine).not.toBeCalled();
+});
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -125,7 +125,9 @@ export async function updateMachines(provider: extensionApi.Provider): Promise<v
   try {
     machineListOutput = await getJSONMachineList();
   } catch (error) {
-    if (shouldNotifySetup) {
+    // Only on macOS and Windows should we show the setup notification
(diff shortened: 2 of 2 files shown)


cdrage:
@afbjorklund @odockal are you able to test to see if this fixes the issue?

afbjorklund:
It works for me, I no longer get the notification loop. But can still run the other engines, using the secret CLI methods.

pd-dashboard-linux

That is, if they are installed and started they will magically appear as providers. When uninstalled or stopped, disappear.

1. `podman-remote-static machine init --now`
2. `limactl start template://podman`
3. `apt install docker.io`

cdrage:
Gone ahead and added another test that covered the `checkDefaultMachine` change as I noticed a bug for that too.

Ready for review 👍

feloy (approved):
LGTM, resolves the problem on my Linux laptop

jeffmaury (approved):
LGTM

extensions/podman/src/extension.spec.ts
```diff
@@ -1088,3 +1088,33 @@ test('checkDisguisedPodmanSocket: runs updateWarnings when called not on Linux',
...
   expect(updateWarningsMock).toBeCalled();
 });
+
+test('Even with getJSONMachineList erroring, do not show setup notification on Linux', async () => {
+  vi.mocked(isLinux).mockReturnValue(true);
+  vi.spyOn(extensionApi.process, 'exec').mockRejectedValue({
+    name: 'name',
+    message: 'description',
+    stderr: 'error',
+  });
```
benoitf, line 1098: I'm wondering if we need the spyOn as it's mocked probably vi.mocked(extensionApi.process.exec) would do the same trick ?

extensions/podman/src/extension.spec.ts
```diff
@@ -1088,3 +1088,33 @@ test('checkDisguisedPodmanSocket: runs updateWarnings when called not on Linux',
...
+  await extension.updateMachines(provider);
+  expect(extensionApi.window.showNotification).not.toBeCalled();
+});
+
+test('if there are no machines, make sure checkDefaultMachine is not being ran inside updateMachines', async () => {
+  const spyCheckDefaultMachine = vi.spyOn(extension, 'checkDefaultMachine');
+  vi.spyOn(extensionApi.process, 'exec').mockImplementation(() => {
+    return Promise.resolve({ stdout: '[]' } as extensionApi.RunResult);
+  });
```
benoitf: `mockResolvedValue({stdout: []})` instead of a mockImplementation that return a Promise.resolve

benoitf (approved):
nits in tests

extensions/podman/src/extension.spec.ts
cdrage: unfortunately do need it, or else you get this error:

```
AssertionError: expected [Function] to throw error including 'description' but got 'wsl bootstrap script failed: exit sta…'

- Expected
+ Received

- description
+ wsl bootstrap script failed: exit status 0xffffffff
```

When not mocking it / using process.exec trick.

extensions/podman/src/extension.spec.ts
cdrage: updated this test and the other test with your suggestion

benoitf (approved):

Issue #6259, opened 2024-03-04
Move all Kubernetes use to new informers


deboer-tim:
### Is your enhancement related to a problem? Please describe

Our initial informers in kubernetes-client.ts served us well, allowing us to build out all of our Kubernetes UI. However, there are a few bugs with reconnections (e.g. #5311) and we have new set of informers that work across multiple contexts, support reconnection and backoff, and allow us to do more in the future like managing resource use or lazy loading/reconnecting to contexts.

### Describe the solution you'd like

We should:
- Finish the set of supported resources by adding support for ingress & routes. #6242
- Move the ingress & routes page to use the new informer. #6242
- Move the navigation counters to the new informers. #6265
- Move details pages to the new informers. #6264

I've verified that after these changes the initial informers are no longer used and could be removed.

deboer-tim:
This is complete after the 3 linked PRs.

PR #6264, opened 2024-03-04
chore: switch details pages to new informers

deboer-tim:
### What does this PR do?

Switches our 4 Kubernetes details pages over to the newer informer stores. The changes to the pages themselves are very simple, but the old stores were (likely incorrectly) writable and the new stores aren't, so mocking was added to the tests to allow us to update the stores directly.

### Screenshot / video of UI

N/A

Fixes another part of #6259.

### How to test this PR?

Go to details for each resource type, confirm no regression.


feloy (approved):
LGTM, tested by switching contexts, making contexts unreachable / reachable again

PR #6265, opened 2024-03-04
chore: use new informers in main nav

deboer-tim:
### What does this PR do?

Uses the newer informers in the main navigation so that the count in the tooltips match what is on the pages.

Fixes a part of #6259.

### How to test this PR?

Confirm tooltip counts in the main navigator match the number of items on each page.

### DIFF
--- a/packages/renderer/src/AppNavigation.spec.ts
+++ b/packages/renderer/src/AppNavigation.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023,2024 Red Hat, Inc.
+ * Copyright (C) 2023-2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


packages/renderer/src/AppNavigation.spec.ts
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023,2024 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2023-2024 Red Hat, Inc.
```
?

benoitf (approved):

packages/renderer/src/AppNavigation.spec.ts
deboer-tim: Sorry, probably a misnomer from when I was originally told how to do it. Fixed.

feloy (approved):

Issue #6220, opened 2024-03-01
Fix run containers screen after new input component


benoitf:
### Is your enhancement related to a problem? Please describe

Before we had:
![image](https://github.com/containers/podman-desktop/assets/436777/9dd5cce1-ad0b-4153-9c77-5fb56efa5495)

Now we have:
![image](https://github.com/containers/podman-desktop/assets/436777/85c6eb07-69fd-4fab-8450-3d23f0b50c8b)

### Describe the solution you'd like

Buttons should be fixed and maybe removing purple underline by default

PR #6266, opened 2024-03-04
feat: update run image screen buttons

deboer-tim:
### What does this PR do?

Switches all the buttons in the Run Images form to the Button
component. The normal button type is way too bold and stands out, and
the secondary button type has an outline and also stands out a lot,
so I've gone with 'link' since that matches the original look the most.

The Input component was using w-full where it should have been using
grow, which was causing it to overlap with buttons. Fixed and updated
test.

### Screenshot / video of UI

Before:

Screenshot 2024-03-04 at 2 38 48 PM

After:

Screenshot 2024-03-04 at 2 38 14 PM

Fixes one part of [#6220](https://github.com/deboer-tim/desktop/issues/6220).

### How to test this PR?

Open the Run

### DIFF
--- a/packages/renderer/src/lib/image/RunImage.svelte
+++ b/packages/renderer/src/lib/image/RunImage.svelte
@@ -5,8 +5,7 @@ import type { ContainerCreateOptions, HostConfig } from '../../../../main/src/pl
 import type { ImageInspectInfo } from '../../../../main/src/plugin/api/image-inspect-info';
 import FormPage from '../ui/FormPage.svelte';
 import type { ImageInfoUI } from './ImageInfoUI';
-import { faFolderOpen, faMinusCircle, faPlay, faPlusCircle, faXmark } from '@fortawesome/free-solid-svg-icons';
-import Fa from 'svelte-fa';
+import { faFolderOpen, faMinusCircle, faPlay, faPlusCircle } from '@fortawesome/free-solid-svg-icons';
 import { router } from 'tinro';
 import Route from '../../Route.svelte';
 import type { NetworkInspectInfo } from '../../../../main/src/plugin/api/network-info';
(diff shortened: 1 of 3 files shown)


deboer-tim:
Was originally a draft on top of #6240, that has merged so moving to review now.

axel7083:
In my opinion it is too much purple for the buttons. They look like they are in an active state or hovering.

deboer-tim:
As per discussion with UX, their recommendation is to merge this (with normal review) and put in another PR to revert some of the input styling.

axel7083 (approved):
Tested locally, and I do not see visual regression where the Input is used, so LGTM

benoitf (approved):
I would have prefered a commit for replacing w-full by grow in Input and another one for updating the RunImage component

I think we're missing the removal of the purple line by default

Also I think after using for a long time that the change of background/borders everywhere should only be applied on focus, not on hover
else moving the cursor across the RunImage page there are too many highlights

(not blocking the merge of this PR)

deboer-tim:
> I would have prefered a commit for replacing w-full by grow in Input and another one for updating the RunImage component

Sorry, happened organically and agree that would be better in the future.

> I think we're missing the removal of the purple line by default

Correct, this was just updating the buttons to use the component (which did require the minor w-full fix to Input, which was causing problems but not as visible before) and why this was not a 'fixes' for #6220. Recommendation was for separate PR to do something about the input styling.

> Also I think after using for a long time that the change of background/borders everywhere should only be applied on focus, not on hover else moving the cursor across the RunImage page there are too many highlights

That's a good idea, or at least softening the effect a lot. This is clearly a case where the original design just doesn't work in practice and we're trying multiple things to undo/improve but no design/consensus yet.

PR #6280, opened 2024-03-05
fix: debounce kubernetes listeners channels individually

### LINKED ISSUE #6277: Some kubernetes contexts changes are not dispatched
### Bug description

The ContextsManager is using some timer in `dispatch` to limit sending data too often. But because we pass options to `dispatch`, to optionally dispatch selected data, we can miss sending data if a more restrictive option overrides other options.

For example:
- time t: send all
- time t + 10ms: send services

At time t, sending all data will be delayed, and will be canceled by the new call at t+10ms (due to the short delay), but only services data will be sent, and other data will be lost.
 

### Operating system

all

### Installation Method

None

### Version

1.7.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Fixes the ContextsManager to debounce data sent individually

Fix #6277

### How to test this PR?

Switch contexts, and check that all resources (pods, deployments, services, ingresses/routes) are updated every time

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -772,48 +772,59 @@ export class ContextsManager {
     });
   }
 
-  dispatchTimer: NodeJS.Timeout | undefined;
   private dispatch(options: DispatchOptions): void {
-    clearTimeout(this.dispatchTimer);
-    this.dispatchTimer = setTimeout(() => {
-      if (options.contextsGeneralState) {
-        this.dispatchContextsGeneralState();
+    if (options.contextsGeneralState) {
+      this.dispatchContextsGeneralState();
+    }
+    if (options.currentContextGeneralState) {
+      this.dispatchCurrentContextGeneralState();
+    }
+    Object.keys(options.resources).forEach(res => {
+      const resname = res as ResourceName;
+      if (options.resources[resname]) {
+        this.dispatchCurrentContextResource(resname);
       }
-      if (options.currentContextGeneralState) {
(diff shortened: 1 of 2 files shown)


deboer-tim (approved):
LGTM, no problems testing.

deboer-tim:
MacOS check failed, restarted.

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -772,48 +772,59 @@ export class ContextsManager {
...
-      this.states.getContextResources(this.kubeConfig.currentContext, resourceName),
+    clearTimeout(this.dispatchCurrentContextResourceTimers.get(resourceName));
+    this.dispatchCurrentContextResourceTimers.set(
+      resourceName,
+      setTimeout(() => {
+        this.apiSender.send(
+          `kubernetes-current-context-${resourceName}-update`,
+          this.states.getContextResources(this.kubeConfig.currentContext, resourceName),
+        );
+      }, dispatchTimeout),
```
benoitf: this method looks like a copy /paste for each dispatch, should it be unified  with a method doing clear and dispatch ?

benoitf (commented):
Hello, I'm not sure to get the latest refactoring change

why debounce has not a parameter like eventName, value, timer

debounce('kubernetes-contexts-general-state-update', this.states.getContextResources(this.kubeConfig.currentContext),
this.dispatchContextsGeneralStateTimer
?

benoitf (approved):
thanks

PR #6281, opened 2024-03-06
feat: enhance learning center to match the mockup design

### LINKED ISSUE #6221: make learning center matching the mockup design

The mockup design was:
image

but current implementation is:
image

- Add the missing collapse (chevron)
- make it matching background (the title should be inside the card)
- make the text smaller

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

- Add the missing collapse (chevron)
- make it matching background (the title should be inside the card)
- make the text smaller

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/9973512/28285e0d-6be3-456e-ae45-3d0e3532ec74

Fixes #6221 


benoitf:
great job

I think the chevron in 'non-expanded' mode is not the expected one

expanded = chevron-down 
non-expanded = chevron-right (and not chevron-up)

I would not introduce expandable UI widget as part of the UI change as it may not be the name we would or if it's really the behavior, etc.
We could say it's a Card etc

feloy:
> great job
> 
> I think the chevron in 'non-expanded' mode is not the expected one
> 
> expanded = chevron-down non-expanded = chevron-right (and not chevron-up)
> 

I'll change

> I would not introduce expandable UI widget as part of the UI change as it may not be the name we would or if it's really the behavior, etc. We could say it's a Card etc

You mean to integrate the Expandable logic directly into LearningCenter?

packages/renderer/src/lib/ui/animations.ts
```diff
@@ -0,0 +1,24 @@
+import { cubicOut } from 'svelte/easing';
```
jeffmaury, line 1: Missing header

packages/renderer/src/lib/Expandable.svelte
```diff
@@ -0,0 +1,26 @@
+<script lang="ts">
```
jeffmaury: Wondering if this shouldn't be in the UI lib

jeffmaury: Missing unit test on the expanded state

slemeur:
Looking great.
Could we add a hover state on the arrows to move left/right?

benoitf:
> You mean to integrate the Expandable logic directly into LearningCenter?

yes

packages/renderer/src/lib/ui/animations.ts
benoitf: FYI this move should be a small PR to move out the animations
so as it's a pure move, small change, small review, fast merge

packages/renderer/src/lib/ui/animations.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function fadeSlide(
```
benoitf, line 3: moving this to a separate file will require specific test for the function as well

jeffmaury (commented):
Why the new component has been removed ? That was a good idea. What's the rationale ?

benoitf:
@jeffmaury was it a good name, could it been applied to other parts 
Should be part of a Card ? etc

benoitf (approved):

feloy:
> Why the new component has been removed ? That was a good idea. What's the rationale ?

@jeffmaury -> https://github.com/containers/podman-desktop/pull/6281#issuecomment-1980367020

feloy:
> Looking great. Could we add a hover state on the arrows to move left/right?

@slemeur -> https://github.com/containers/podman-desktop/pull/6284

slemeur:
thanks @feloy

jeffmaury (approved):
LGTM

PR #6282, opened 2024-03-06
fix: clear ContextsManager timers when app exits

### LINKED ISSUE #6198: Timers are causing uncaught exceptions when exiting
The `ContextsManager` (`packages/main/src/plugin/kubernetes-context-state.ts`) allocates timers during its lifecycle, but these timers are not disposed when Podman Desktop exits, and this can lead to such error messages when exiting from Podman Desktop:

timer-error

feloy:

### What does this PR do?

Clear ContextsManager timers when app exits.

Fixes #6198

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -1407,6 +1407,7 @@ export class ExtensionLoader {
     await Promise.all(
       Array.from(this.activatedExtensions.keys()).map(extensionId => this.deactivateExtension(extensionId)),
     );
+    this.kubernetesClient.dispose();
   }
 
   async startExtension(extensionId: string): Promise<void> {
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -1375,4 +1375,8 @@ export class KubernetesClient {
   public unregisterGetCurrentContextResources(resourceName: ResourceName): KubernetesObject[] {
     return this.contextsState.unregisterGetCurrentContextResources(resourceName);
   }
+
+  public dispose(): void {
+    this.contextsState.dispose();
+  }
 }
--- a/packages/main/src/plugin/kubernetes-context-state.spec.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.spec.ts
@@ -188,707 +188,225 @@ afterEach(() => {
   vi.resetAllMocks();
 });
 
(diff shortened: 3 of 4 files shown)


packages/main/src/plugin/index.ts
```diff
@@ -181,7 +181,7 @@ export class PluginSystem {
   private uiReady = false;
 
   // true if the application is quitting
-  private isQuitting = false;
+  public isQuitting = false;
```
benoitf: I think that it needs to stay private

if you need the value, you can add an accessor but we can't let anyone being able to override the value

benoitf:
I think it's a workaround correct ? as the root cause is that timers should be canceled

benoitf:
I don't see how it's fixing the root problem/issue (It's just hiding the problem)

feloy:
Yes, it's a workaround, as I didn't find the way to stop the timers when the app was exiting. But working on this side, it helped me realize that the kubernetes-client can be informed to stop the timers when this event is triggered: 
```
app.on('before-quit', () => {
  this.isQuitting = true;
});
```

I'll make the changes

benoitf:
please do not add dependency to app events from anything in the plugin/** part

all objects created during the start of the plug-in context should receive teardown by the same calls that created them

feloy:
> please do not add dependency to app events from anything in the plugin/** part
> 
> all objects created during the start of the plug-in context should receive teardown by the same calls that created them

I tried to apply this. Please tell me if it's ok

packages/main/src/plugin/index.ts
```diff
@@ -459,6 +459,10 @@ export class PluginSystem {
 
     const messageBox = new MessageBox(apiSender);
 
+    app.on('before-quit', () => {
```
benoitf: we've already this event handled in this class and we call extension loader stop all extensions

For now, let's move the kubernetesClient dispose to extension loader.stopAllExtensions

We'll add later a stop all that should call stop all extensions and dispose on the kubernetesClient

feloy:
I still need to add some tests on the dispose() method

benoitf:
thanks @feloy , works for me

deboer-tim (approved):
I'm generally not able to reproduce the original problem, but testing on this PR caused no issues and code LGTM.

benoitf:
created https://github.com/containers/podman-desktop/issues/6288 as follow-up issue

benoitf (approved):

PR #6286, opened 2024-03-06
chore: enable prefering nullish operator

cdrage:
chore: enable prefering nullish operator

### What does this PR do?

Bad habit of using || when it should be ??

This enables erroring out for `yarn lint:fix` where it should be ?? instead of
||

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

N/A

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -80,6 +80,7 @@
     "@typescript-eslint/no-floating-promises": "error",
     "@typescript-eslint/no-misused-promises": "error",
     "@typescript-eslint/prefer-optional-chain": "error",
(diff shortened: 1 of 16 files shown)


cdrage:
Draft just to see how much we have to change to get this to work..

benoitf:
273

benoitf:
@cdrage can you try to have a separate PR just introducing `"strictNullChecks": true,` (it's also missing the part in `website` folder)

cdrage:
Will re-open adding strict and then open this one again with the fixes 👍

PR #6287, opened 2024-03-06
Add listInfos method to API

### LINKED ISSUE #6097: Get podman machine info from extension

From an extension, we would like to have memory/cpu information about podman machines in use.

There is a method in the api `containerEngine.info(engineId)`, which provides this information. 

To get the `engineId`, I understand that we nned to execute a `list*` command and, if there are elements returned by any of these functions, to get the engineId from this result (for example, an image returned by `listImages` conatains the engineId on which it is present.

It would be nice to have a more direct way to get the `engineId`, for example whan we have a `ProviderContainerConnection` value (obtained from `provider.getContainerConnections`).

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

Add a listInfos to the API, for the extension to get cpu/memory/... information about different container providers.

TODO:
- [x] doc new method
- [x] unit tests

Fixes #6097 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -2324,6 +2324,9 @@ export class ContainerProviderRegistry {
     if (provider.libpodApi) {
       const podmanInfo = await provider.libpodApi.podmanInfo();
       return {
+        engineId: provider.id,
+        engineName: provider.name,
+        engineType: provider.connection.type,
         cpus: podmanInfo.host.cpus,
         cpuIdle: podmanInfo.host.cpuUtilization.idlePercent,
         memory: podmanInfo.host.memTotal,
@@ -2334,6 +2337,9 @@ export class ContainerProviderRegistry {
     } else {
       const dockerInfo = await provider.api.info();
       return {
+        engineId: provider.id,
+        engineName: provider.name,
+        engineType: provider.connection.type,
         cpus: dockerInfo.NCPU,
         memory: dockerInfo.MemTotal,
       };
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -2358,7 +2358,7 @@ export class ContainerProviderRegistry {
         try {
           return await this.info(provider.id);
         } catch (error) {
-          console.log('error getting info for engine', provider.name, error);
+          console.error('error getting info for engine', provider.name, error);
           return undefined;
         }
       }),
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -2346,6 +2346,26 @@ export class ContainerProviderRegistry {
     }
   }
 
+  async listInfos(options?: containerDesktopAPI.ListInfosOptions): Promise<containerDesktopAPI.ContainerEngineInfo[]> {
+    let providers: InternalContainerProvider[];
+    if (options?.provider === undefined) {
+      providers = Array.from(this.internalProviders.values());
+    } else {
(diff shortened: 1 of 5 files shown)


benoitf (commented):
LGTM

packages/main/src/plugin/container-registry.ts
```diff
@@ -2334,12 +2337,35 @@ export class ContainerProviderRegistry {
...
+        engineType: provider.connection.type,
         cpus: dockerInfo.NCPU,
         memory: dockerInfo.MemTotal,
       };
     }
   }
 
+  async listInfos(options?: containerDesktopAPI.ListInfosOptions): Promise<containerDesktopAPI.ContainerEngineInfo[]> {
+    let providers: InternalContainerProvider[];
+    if (options?.provider === undefined) {
```
```suggestion
    if (!options?.provider) {
```

packages/main/src/plugin/container-registry.ts
```diff
@@ -2334,12 +2337,35 @@ export class ContainerProviderRegistry {
...
+      Array.from(providers).map(async provider => {
+        try {
+          return await this.info(provider.id);
+        } catch (error) {
+          console.log('error getting info for engine', provider.name, error);
+          return [];
+        }
+      }),
+    );
+    return infos.flat();
```
benoitf: I'm not sure here about the usage of array and then flat
vs filtering out undefined
return Array.from(providers).map(provider => return info(provider).filter(p => !p)

feloy: I cannot get the type checker detect that `filter(info => info)` does not return undefined values. The result is always of type `(containerDesktopAPI.ContainerEngineInfo | undefined)[]`

```
    const infos = await Promise.all(
      Array.from(providers).map(async provider => {
        try {
          return await this.info(provider.id);
        } catch (error) {
          console.log('error getting info for engine', provider.name, error);
          return undefined;
        }
      }).filter(async info => info),
    );
    return infos;
```

Or do I misunderstand what you propose?

benoitf: the trick is to use the cast after the filter call as you know it's filtering out undefined

edit: sorry it's not with as it's with `is` keyword

benoitf: `filter((info): info is ContainerEngineInfo => !!info)`

benoitf (approved):

packages/main/src/plugin/container-registry.ts
```diff
@@ -2334,12 +2337,35 @@ export class ContainerProviderRegistry {
...
+      providers = Array.from(this.internalProviders.values());
+    } else {
+      providers = [this.getMatchingContainerProvider(options?.provider)];
+    }
+    const infos = await Promise.all(
+      Array.from(providers).map(async provider => {
+        try {
+          return await this.info(provider.id);
+        } catch (error) {
+          console.log('error getting info for engine', provider.name, error);
```
axel7083:
```suggestion
          console.error('error getting info for engine', provider.name, error);
```

axel7083 (approved):
LGTM!

feloy:
I'm trying to call the method from an extension (one provided with Podman Desktop, kube-context), but I'm getting this error:

```
16:45:13 [main] Activating extension podman-desktop.kube-context failed error:TypeError: extensionApi__namespace.containerEngine.listInfos is not a function
```

Do I need to rebuild / copy the .d.ts file somewhere?

benoitf:
@feloy `yarn build:preload:types`

benoitf:
hmm no it should even work without that command

feloy:
> hmm no it should even work without that command

It seems it fixed the problem. In any case, it works now. Thanks

PR #6290, opened 2024-03-06
chore: reduce impact of Input styling

### LINKED ISSUE #6220: Fix run containers screen after new input component

deboer-tim:
### What does this PR do?

Changes three things in the Input styling to reduce the changes from 1.7. This is really a judgement call, but to me these were the 3 things that reduced the impact but still kept some of the nicer aspects/improvements in the input design from #3234:

- Changes the bottom border/focus outline (stroke color) from purple to charcoal. (one suggestion was to remove it, but that's both a bigger change and IMHO this still provides some nice minor styling)
- Added a new input-field-bg variable so that you can customize the regular background of the input, and set it to charcoal-800.
- The mouse-over effect was identical to focus (rounded, full border, dark background) and was distracting if you moved the mouse over multiple Inputs. This removes the background change by making the hover effect transparent. This required moving transparent color to our new color palette, and adding a not() so that the hover affect doesn't effect focussed inputs.

### Screenshot / video of UI

1.7:

Screenshot 2024-03-06 at 12 13 52 PM

After:

https://github.com/containers/podman-desktop/assets/19958075/ace68f2c-6cd0-4605-855a-91c65756b4d3

Fixes #6220.

### How to test this PR?

Look at all inputs: search bars, preferences, etc., but especially the Run Image form.

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -371,18 +371,21 @@ export class ColorRegistry {
   protected initInputBox(): void {
     const sNav = 'input-field-';
 
+    this.registerColor(`${sNav}bg`, {
+      dark: colorPalette.charcoal[800],
+      light: colorPalette.charcoal[800],
+    });
     this.registerColor(`${sNav}focused-bg`, {
       dark: colorPalette.charcoal[900],
       light: colorPalette.gray[100],
     });
-
     this.registerColor(`${sNav}disabled-bg`, {
       dark: colorPalette.charcoal[900],
       light: colorPalette.charcoal[900],
     });
     this.registerColor(`${sNav}hover-bg`, {
-      dark: colorPalette.charcoal[900],
-      light: colorPalette.gray[100],
+      dark: colorPalette.transparent,
+      light: colorPalette.transparent,
     });
     this.registerColor(`${sNav}focused-text`, {
       dark: colorPalette.white,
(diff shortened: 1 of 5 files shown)


cdrage (approved):
Big 👍 IMHO regarding improvement, not having the backgrounds makes it a lot easier to read.

This is a great in-between / compromise between the two suggestions.

LGTM UI side.

Implementation wise I'm like 80% LGTM, I'm unsure if the `bg-[var(--pd-input-field-bg)]` is okay as I hadn't seen that type of implementation before.

slemeur:
Nice LGTM

benoitf:
AFAIK we're missing the highlighted color now.
Clicking in the field should bring the purple color

benoitf (changes requested):
adding new colors in color-registry needs to be reviewed by UX team first

deboer-tim:
Updated in response to today's feedback, video in description updated:
- Bring back purple for hover and focus. Turns out we had an unused hover-stroke variable (!) I could use.
- Remove rounding from hover, so it only sets a (purple) bottom border.
- Remove the background again - it's a cleaner look, bg only changes on focus (not hover).

benoitf (approved):

ekidneyrh (approved):

PR #6314, opened 2024-03-07
nit: no right margin in menus

deboer-tim:
### What does this PR do?

Our drop-down menus have a right margin, which means there is always a small unused/white-space to the right.

The margin was added in Nov 2022 when the hover behaviour was added and I can't find any reason for it. (maybe the main pages at the time needed the actions moved from the right edge of the window?) I've been through everywhere we use this component today, and removing the margin had the same minor layout improvement in all cases.

### Screenshot / video of UI

Before:

Screenshot 2024-03-07 at 10 33 33 AM

After:

Screenshot 2024-03-07 at 10 34 00 AM

N/A

### How to test this PR?

Hover over right-most action in any of the Images/Pods/Containers pages to see the padding is now consistent top/right/bottom.


benoitf (approved):

cdrage (approved):
Checked through a bunch of the tabs (including k8s).

LGTM.

PR #6318, opened 2024-03-07
feat: remember window position and size of Podman Desktop window

### LINKED ISSUE #6115: Remember window state and position from last start

Reported via feedback form: every time you start the main window reverts to being open and in the default position.

When you exit Podman Desktop we should save the last window state (position/size, or closed) and reopen in the same state.

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Save and restore the window position where it was before closing Podman Desktop

fixes https://github.com/containers/podman-desktop/issues/6115

### How to test this PR?

start podman desktop, move and resize the window. Then exit and restart
Check that it's at the same position than before

you can disable it through the settings/preferences (window)

### DIFF
--- a/packages/main/src/system/window/window-init.ts
+++ b/packages/main/src/system/window/window-init.ts
@@ -1,0 +1,75 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { IConfigurationNode, IConfigurationRegistry } from '/@/plugin/configuration-registry.js';
+import { WindowSettings } from './window-settings.js';
+
+/**
+ * Save/Restore the window size and position.
+ */
+export class WindowInit {
+  readonly #configurationRegistry: IConfigurationRegistry;
+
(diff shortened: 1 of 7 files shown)


deboer-tim (commented):
Tested & worked, with multiple monitors.

A couple times the window appeared in the default position and then jumped to the correct position. Is there anything we can do to avoid this/set position earlier?

I don't love creating a new Window preferences category as it feels like we have a lot of single preferences in their own category. There isn't anything obvious to group this with though either. Maybe we should start a General category?

Would also be nice if we remembered if the window was closed on last run (i.e. system tray only), but ok if that's a separate issue. (I think we have one in the backlog, and maybe we should still show the loading screen or something so that you know it actually launched)

packages/main/src/system/window/window-init.ts
```diff
@@ -0,0 +1,75 @@
...
+  constructor(configurationRegistry: IConfigurationRegistry) {
+    this.#configurationRegistry = configurationRegistry;
+  }
+
+  init(): void {
+    const heightKey = `${WindowSettings.SectionName}.${WindowSettings.Height}`;
+    const widthKey = `${WindowSettings.SectionName}.${WindowSettings.Width}`;
+    const xPositionKey = `${WindowSettings.SectionName}.${WindowSettings.XPosition}`;
+    const yPositionKey = `${WindowSettings.SectionName}.${WindowSettings.YPosition}`;
+    const restorePosition = `${WindowSettings.SectionName}.${WindowSettings.RestorePosition}`;
```
jeffmaury, line 37: Seems to me it should end by Key to be aligned with previous lines

cdrage:
Tested and works with macOS, but will need to test on Linux and Windows too.

On M1, it's hard to take a video of it, but I get the same thing where it "jumps" to the separate section of the screen.

EDIT: Also need to check plugging in two monitors, putting PD on one monitor, unplugging it and then seeing what happens with PD. Usually some windows will go "back" to the main screen after.

jeffmaury (commented):
I think the split between window init and window handler are making the code less readable and the first one creates conf that is used by the second so a merge would be good.

I have also a concern with storing the position as I have the following use case:
- Working with 2 screens Podman Desktop starts on screen 1 (laptop) and moved to screen 2 (monitor)
- Going back home (laptop only) restarting Podman Desktop but no window being visible

PR #6320, opened 2024-03-07
Initialize the context store

### LINKED ISSUE #4605: Optional options hidden in podman machine creation page on Windows and Mac after force reload
### Bug description

Opening this issue to keep track of a possible bug which does not happen frequently.
When testing the new 1.5.2, I tried to create a new podman machine and all options (rootful, usernetworking) were hidden.
After a restart it started working fine.
The same happen to @odockal once.

I was not able to replicate it again. If someone has more info please post it here. Thanks!!

### Operating system

Windows 10/ 11 and Mac OS 14

### Installation Method

### Version

next (development version)

### Steps to reproduce

1. Install Podman (either using onboarding, or dashboard buttons)
2. Go To Settings -> Resources
3. Verify: Open Create new ... (podman machine) form/page got all options available
4. Go back to Resources and call Force Reload (CTRL + R on windows)
5. Verify the Create new page again
6. Assert: All Buttons are present
Actual result: These optional button are not visible.

### Relevant log output

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

When the user reloads the window, all the context values received before are not in the store anymore.

The store needs to get the current values of the context when some client subscribes to it again.

- Initialize the `context` store with the current values in the store
- test the `context` store
- mock the `context` store when testing components

Fixes #4605

### How to test this PR?

See Steps to reproduce in #4605

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -2112,6 +2112,10 @@ export class PluginSystem {
         });
       },
     );
+    this.ipcHandle(
+      'context:collectAllValues',
+      async (): Promise<Record<string, unknown>> => context.collectAllValues(),
+    );
 
     const dockerDesktopInstallation = new DockerDesktopInstallation(
       apiSender,
--- a/packages/preload/src/index.ts
+++ b/packages/preload/src/index.ts
@@ -1979,6 +1979,10 @@ export function initExposure(): void {
   contextBridge.exposeInMainWorld('listGuides', async (): Promise<Guide[]> => {
     return ipcInvoke('learning-center:listGuides');
   });
+
+  contextBridge.exposeInMainWorld('contextCollectAllValues', async (): Promise<Record<string, unknown>> => {
+    return ipcInvoke('context:collectAllValues');
+  });
 }
 
 // expose methods
--- a/packages/renderer/src/Loader.spec.ts
+++ b/packages/renderer/src/Loader.spec.ts
@@ -22,6 +22,9 @@ import { beforeAll, expect, test, vi } from 'vitest';
 import Loader from './Loader.svelte';
 import { render } from '@testing-library/svelte';
 import { router } from 'tinro';
+import * as contextStore from '/@/stores/context';
+import { writable } from 'svelte/store';
+import { ContextUI } from './lib/context/context';
 
 // first, patch window object
 const callbacks = new Map<string, any>();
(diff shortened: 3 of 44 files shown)


benoitf (commented):
I think that there is something going wrong if we need to patch all spec.ts files

feloy:
> I think that there is something going wrong if we need to patch all spec.ts files

Yes, that's also my opinion. One thing I find strange is that the ListItemButtonIcon depends on the context store. Because of this, all the components using this one are dependent on the store. And it should be a generic component, not depending on this context.

I would like to move the context logic out of this component and place it into the parent(s) for which the logic is dedicated, but I don't know exactly which parents are concerned.

benoitf:
In the description of the PR, there is not much explanation on the root cause.

Explanation is that in the issue reported, people do a 'window reload' operation. Contexts are only updated when their values are updated on the extension's side.

so it means that when we refresh the client side, context is empty as we didn't collect any values

The store is invalid, the store should listen on `extensions-already-started` and refresh its content.

Then, IMHO the proper fix is to only listen on `extensions-already-started`, call the collectAllValues() and update the values

no need to update all spec.ts files or to introduce getCurrentContext etc

feloy:
> In the description of the PR, there is not much explanation on the root cause.
> 
> Explanation is that in the issue reported, people do a 'window reload' operation. Contexts are only updated when their values are updated on the extension's side.
> 
> so it means that when we refresh the client side, context is empty as we didn't collect any values
> 
> The store is invalid, the store should listen on `extensions-already-started` and refresh its content.
> 
> Then, IMHO the proper fix is to only listen on `extensions-already-started`, call the collectAllValues() and update the values
> 
> no need to update all spec.ts files or to introduce getCurrentContext etc

As in #6343?

It seems to me that the problem remains the same, as, during the tests, the call to `window.contextCollectAllValues` fails, because the function does not exist, and is caught:

``` 
npx vitest run PodsList.spec.ts
[...]
error getting current context TypeError: window.contextCollectAllValues is not a function
```  

If we want to get rid of this catch, we need to patch the 45 spec files to declare the window.contextCollectAllValues function (or mock the store, as I have done here).


My reasoning in this PR is that, when the user reloads the window, the components are subscribing again to the `context` store. In the current PR, I have added a second parameter to the `writable` call, a function which is called for the first suscriber, and which reloads the context by calling collectAllValues (through `getCurrentContext`). It seems to me that, at the end, the same calls are made, in both versions.

benoitf:
yes as in https://github.com/containers/podman-desktop/pull/6343 

I'll comment in this PR

feloy:
Closing in favor of #6343

PR #6329, opened 2024-03-08
chore: allow to install dd extensions from the common extension installer

### LINKED ISSUE #6226: common installation method for Podman Desktop extensions or DD extensions

For now there are two endpoints/methods. As we're analyzing OCI images, we can check metadata and then delegate to DD part if we see DD metadata or continue with Podman Desktop installation

step 1: Add checks for DD extensions in the extension installer
it'll allow to have a common installation method

other steps that could be done later: merging contributionManager and extensionLoader but let's iterate

_No response_

### Additional context

part of https://github.com/containers/podman-desktop/issues/6083

benoitf:
### What does this PR do?
allow to call the install of DD extensions from the common installer workflow

it'll allow to remove the specific DD workflow later

fixes https://github.com/containers/podman-desktop/issues/6226

### How to test this PR?

Go to extension installer page (not DD extension page) and try to install trivy image `aquasec/trivy-docker-extension:latest`
you should see a new item in the navbar at the end

### DIFF
--- a/packages/main/src/plugin/install/extension-installer.ts
+++ b/packages/main/src/plugin/install/extension-installer.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,15 +28,24 @@ import type { ApiSenderType } from '../api.js';
 import type { ImageRegistry } from '../image-registry.js';
 import type { ExtensionsCatalog } from '../extensions-catalog/extensions-catalog.js';
 import type { Telemetry } from '../telemetry/telemetry.js';
+import type { Directories } from '/@/plugin/directories.js';
+import { DockerDesktopContribution, DockerDesktopInstaller } from '../docker-extension/docker-desktop-installer.js';
+import type { ContributionManager } from '../contribution-manager.js';
 
 export class ExtensionInstaller {
+  #dockerDesktopInstaller: DockerDesktopInstaller;
+
   constructor(
     private apiSender: ApiSenderType,
     private extensionLoader: ExtensionLoader,
     private imageRegistry: ImageRegistry,
     private extensionCatalog: ExtensionsCatalog,
     private telemetry: Telemetry,
-  ) {}
+    private directories: Directories,
+    private contributionManager: ContributionManager,
+  ) {
+    this.#dockerDesktopInstaller = new DockerDesktopInstaller(contributionManager);
+  }
 
   async extractExtensionFiles(
     tmpFolderPath: string,
@@ -79,9 +88,6 @@ export class ExtensionInstaller {
         return cp(sourceFile, path.join(finalFolderPath, 'host', destFile), { recursive: true });
       }),
     );
-
-    // delete the tmp folder
-    fs.rmSync(tmpFolderPath, { recursive: true });
   }
 
   async unpackTarFile(tarFilePath: string, destFolder: string): Promise<void> {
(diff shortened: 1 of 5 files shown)


axel7083:
## Not related to this PR 

_While testing it_

We could really improve the user feedback on this process

https://github.com/containers/podman-desktop/assets/42176370/9a88efc8-c34a-44c0-839a-7e551a1bd169

Moreover when testing the extension you gave with a hello world image it load indefinelty 

image

packages/main/src/plugin/install/extension-installer.ts
```diff
@@ -120,16 +126,28 @@ export class ExtensionInstaller {
       return;
     }
 
-    const titleLabel = imageConfigLabels['org.opencontainers.image.title'];
+    const titleLabel = imageConfigLabels['org.opencontainers.image.title'] as string;
```
axel7083:
```suggestion
    const titleLabel = imageConfigLabels['org.opencontainers.image.title'] as string | undefined;
```

axel7083 (approved):
LGTM and installation works fine

packages/main/src/plugin/install/extension-installer.ts
benoitf: I guess you meant a `||` ?

a couple of lines after there is test on the availability so empty string or undefined string would act the same way

axel7083: > I guess you meant a `||` ?

No I mean this should not be cast as a string when it could be undefined ? the test after ensure it not undefined, therefore this can be either a string or undefined ?

After the check you mention it is a string, as it cannot be undefined, but before this it can be

packages/main/src/plugin/install/extension-installer.ts
```diff
@@ -120,16 +126,28 @@ export class ExtensionInstaller {
...
-    const titleLabel = imageConfigLabels['org.opencontainers.image.title'];
+    const titleLabel = imageConfigLabels['org.opencontainers.image.title'] as string | undefined;
     const descriptionLabel = imageConfigLabels['org.opencontainers.image.description'];
     const vendorLabel = imageConfigLabels['org.opencontainers.image.vendor'];
     const apiVersion = imageConfigLabels['io.podman-desktop.api.version'];
+    const apiDDVersion = imageConfigLabels['com.docker.desktop.extension.api.version'];
 
-    if (!titleLabel || !descriptionLabel || !vendorLabel || !apiVersion) {
+    if (!titleLabel || !descriptionLabel || !vendorLabel || (!apiVersion && !apiDDVersion)) {
       sendError(`Image ${imageName} is not a Podman Desktop Extension`);
```
jeffmaury, line 136: Message should be updated

packages/main/src/plugin/install/extension-installer.ts
```diff
@@ -145,37 +163,49 @@ export class ExtensionInstaller {
     const tmpFolderPath = path.join(os.tmpdir(), `/tmp/${imageNameWithoutSpecialChars}-tmp`);
 
     // final folder
-    const finalFolderPath = path.join(this.extensionLoader.getPluginsDirectory(), imageNameWithoutSpecialChars);
+    const finalFolderPath = path.join(unpackedFolder, imageNameWithoutSpecialChars);
 
     // grab all extensions
-    const extensions = await this.extensionLoader.listExtensions();
+    if (isPDExtension) {
```
jeffmaury, line 169: Looks like we don"t check for Docker Desktop extensions ?

packages/main/src/plugin/install/extension-installer.ts
benoitf: well it seems we wanted to abstract that everything is a Podman Desktop extension  (don't matter if it's podman or docker) so if message is updated it will bring back the issue no ?

packages/main/src/plugin/install/extension-installer.ts
benoitf: no because it'll require some fields that we don't have yet

benoitf: (and today it's overriding the extension, you can already install an existing Docker Desktop extension, it's reinstalling it)

jeffmaury (approved):
LGTM

PR #6340, opened 2024-03-10
chore: add basic colors for cards

benoitf:
### What does this PR do?

There was only the inverted cards colors

https://github.com/containers/podman-desktop/issues/4907

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/color-registry.spec.ts
+++ b/packages/main/src/plugin/color-registry.spec.ts
@@ -52,6 +52,10 @@ class TestColorRegistry extends ColorRegistry {
   initTitlebar(): void {
     super.initTitlebar();
   }
+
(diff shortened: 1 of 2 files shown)


deboer-tim (approved):
Code and test LGTM.

I am worried that over time the tests checking RGB values will make it hard to make changes though - e.g. if we want to change one variable color we need to update in 2 places and look up the palette value; if we want to tweak the palette or decide 'selection purple should be one palette index darker' it would need a huge PR. Hoping over time we could find a different way to do this, or test vs palette entry.

benoitf:
@deboer-tim I've added a fixup in the PR

deboer-tim:
> @deboer-tim I've added a fixup in the PR

Thank you 👍🏼.

PR #6343, opened 2024-03-10
fix: context store gets current context on extensions-already-started event

### LINKED ISSUE #4605: Optional options hidden in podman machine creation page on Windows and Mac after force reload

feloy:

### What does this PR do?

The current `context` store is initialized with an empty context and receives updates when the context changes.

But if the user reloads the window, the store is reset, but never receives again the previous values.
 
This PR gets the current values of the context to set the new value of the context store when the event `extensions-already-started` is received, event happening when the window is reloaded.

Fixes #4605 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/stores/context.spec.ts
+++ b/packages/renderer/src/stores/context.spec.ts
@@ -1,0 +1,88 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
+import { setup } from './context';
+import { ContextUI } from '../lib/context/context';
+import { get } from 'svelte/store';
+
+const contextCollectAllValues = vi.fn();
+const receiveMock = vi.fn();
+const addEventListenerMock = vi.fn();
+
+beforeAll(() => {
+  (window as any).contextCollectAllValues = contextCollectAllValues;
+  (window as any).events = { receive: receiveMock };
+  (window as any).window.addEventListener = addEventListenerMock;
+});
+
+beforeEach(() => {
+  vi.resetAllMocks();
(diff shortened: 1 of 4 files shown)


benoitf:
tested, it works fine, now need tests 🎉

packages/renderer/src/stores/context.ts
```diff
@@ -20,6 +20,13 @@ import type { Writable } from 'svelte/store';
 import { writable } from 'svelte/store';
 import { ContextUI } from '../lib/context/context';
 
+async function getCurrentContext(): Promise<ContextUI> {
+  return Object.entries(await window.contextCollectAllValues()).reduce((result, [key, value]) => {
+    result.setValue(key, value);
+    return result;
+  }, new ContextUI());
+}
+
```
benoitf: maybe this method content can be moved to the addEventListener method

Also, add a check like we do that only if window.contextCollectAllValues exists

so no need to change all spec.ts files, only the one that would want to override the context

benoitf: or in Loader.svelte we could also send an event to the backend in case of client refresh, so backend could send again new stuff

in that case, backend will send context-value-updated calls for contexts

feloy:
> tested, it works fine, now need tests 🎉

As the code in `context.ts` is global, I need to create mocks in a `vi.hoisted`, so they are mocked before the import of the `context.ts`, right?

benoitf:
you can also do like for other stores

like a class with a setup method, or just a setup function()
in the store it calls setup function

but then in your test you can do in beforeEach the hooks you want and then you manually call the setup method that will redo the init but using your hooks

packages/renderer/src/stores/context.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
+import { setup } from './context';
+import { ContextUI } from '../lib/context/context';
+import { get } from 'svelte/store';
+
+const contextCollectAllValues = vi.fn();
```
benoitf, line 24: not: this one is not suffixed by Mock like the 2 following

benoitf (approved):

jeffmaury (approved):
LGTM. Tested on Win11Pro

PR #6354, opened 2024-03-11
fix: try to listen on all ipv4 interfaces AND 0.0.0.0 before to declare a port free

### LINKED ISSUE #5996: Detection of port already in use fails
### Bug description

Steps to reproduce:

- On Podman Desktop, go to the Images page
- choose an image (for example nginx), and select the icon > (Create a container from image)
- look at port mapping, it should be 9000
- Start Container
- Container should be running

- choose the same image, and select again > (Create a container from image)
- look at the port mapping, it should be 9001

On Mac, the port displayed is still 9000.

Debugging the code, I have seen that https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/util/port.ts#L64C19-L64C20 tries to listen on 127.0.0.1:port, and on my system the detection fails. If I change to 0.0.0.0:port, the detection works correctly.

### Operating system

mac m3 (Sonoma 14.3 (23D56))

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

If an application binds to a port using the INADDR_ANY address (0.0.0.0) and setting the `SO_REUSEADDR` flag, another application can bind to the port on a local address without failure.

> Some folks don't like SO_REUSEADDR because it has a security stigma attached to it. On some operating systems it allows the same port to be used with a different address on the same machine by different processes at the same time. This is a problem because most servers bind to the port, but they don't bind to a specific address, instead they use INADDR_ANY (this is why things show up in netstat output as *.8080). So if the server is bound to *.8080, another malicious user on the local machine can bind to local-machine.8080, which will intercept all of your connections since it is more specific. (https://stackoverflow.com/questions/60217071/node-js-server-listens-to-the-same-port-on-0-0-0-0-and-localhost-without-error)

This PR duplicates the tests where a call to `listen` is done, to listen either in `127.0.0.1` or in `0.0.0.0`.

On Mac, the tests when listening on `0.0.0.0` fail without this patch.

The proposed patch is to try and listen on all local addresses AND 0.0.0.0 before to declare that a port is free.

Fixes #5996 

### How to test this PR?

See Steps to reproduce in #5996 

### DIFF
--- a/packages/main/src/plugin/util/port.ts
+++ b/packages/main/src/plugin/util/port.ts
@@ -55,12 +55,16 @@ export async function getFreePortRange(rangeSize: number): Promise<string> {
   return `${startPort}-${port - 1}`;
 }
 
-export function isFreePort(port: number): Promise<boolean> {
+function isFreeAddressPort(address: string, port: number): Promise<boolean> {
   const server = net.createServer();
   return new Promise((resolve, reject) =>
     server
       .on('error', (error: NodeJS.ErrnoException) => (error.code === 'EADDRINUSE' ? resolve(false) : reject(error)))
       .on('listening', () => server.close(() => resolve(true)))
-      .listen(port, '127.0.0.1'),
+      .listen(port, address),
   );
+}
+
+export async function isFreePort(port: number): Promise<boolean> {
+  return (await isFreeAddressPort('127.0.0.1', port)) && (await isFreeAddressPort('0.0.0.0', port));
 }
--- a/packages/main/src/plugin/util/port.ts
+++ b/packages/main/src/plugin/util/port.ts
@@ -17,6 +17,7 @@
  ***********************************************************************/
 
 import * as net from 'node:net';
+import { type NetworkInterfaceInfoIPv4, networkInterfaces } from 'node:os';
 
 /**
  * Find a free port starting from the given port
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/util/port.ts
```diff
@@ -55,12 +55,16 @@ export async function getFreePortRange(rangeSize: number): Promise<string> {
...
     server
       .on('error', (error: NodeJS.ErrnoException) => (error.code === 'EADDRINUSE' ? resolve(false) : reject(error)))
       .on('listening', () => server.close(() => resolve(true)))
-      .listen(port, '127.0.0.1'),
+      .listen(port, address),
   );
 }
+
+export async function isFreePort(port: number): Promise<boolean> {
+  return (await isFreeAddressPort('127.0.0.1', port)) && (await isFreeAddressPort('0.0.0.0', port));
```
benoitf: I'm wondering, should it try on all interfaces and not only on 127.0.0.1 and 0.0.0.0

let's imagine I'm connected to a router/wifi network, I'll have an ip and I can bind something on that port

and it'll tell me the port is free there no ?

benoitf: maybe something like

const networkInterfaces = [...];
const checkInterfaces = await Promise.all(networkInterfaces.map(interface => isFreeAddressPort(interface, port));
return checkInterfaces.every(element => element === true);

feloy: This makes the tests fail on CI. It seems that listen fails on all ports of some interface in the CI's machine. Looking at it...

benoitf (approved):

benoitf:
@feloy probably need to update PR title and description (as you're checking on all ipv4 network interfaces now)

benoitf:
(also curious why only ipv4 and not ipv6 ?)

feloy:
> (also curious why only ipv4 and not ipv6 ?)

I tried locally without filtering ipv6 addresses, and there are some error trying to bind on some specific ipv6 address. I don't feel comfortable enough with ipv6 addresses exceptions to add a list of ipv6 addresses to remove from the test. And as we are targeting desktops, I feel that mostly ipv4 addresses are used on them.

benoitf:
PTAL @dgolovin and @cdrage

cdrage (approved):
Tested this on Mac M1 and didn't have any issues. Great change and I from when I tested I didn't encounter any of the previous port / address issues. 

LGTM.

PR #6400, opened 2024-03-13
chore: fetch dependencies before publishing

benoitf:
### What does this PR do?
For now we're missing cli tools when publishing as dependencies are not fetched

https://github.com/containers/podman-desktop/actions/runs/8256131999

### How to test this PR?


cdrage (approved):
LGTM. Let's see if this builds correctly 👍

cdrage:
Pushed, saw that it is built, but no changes to: https://www.npmjs.com/package/@podman-desktop/ui-svelte?activeTab=versions yet (probably takes a day?)

PR #6401, opened 2024-03-13
chore: store combining PD and DD extensions

### LINKED ISSUE #6341: Add a store combining Docker Desktop extensions and Podman Desktop extensions

as part of revamping Podman Desktop extensions and Docker desktop extensions, there should be only one store aggregating all extensions

combine all extensions sources into one to make one UI widget

add `type` with `dd` or `pd` to know the source

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
In UI we need to display PD and DD in a common way, use this derived store that is aggregating both sources

it adds a `type` attribute being either `pd` or `dd`

fixes https://github.com/containers/podman-desktop/issues/6341

### How to test this PR?

### DIFF
--- a/packages/renderer/src/stores/all-installed-extensions.ts
+++ b/packages/renderer/src/stores/all-installed-extensions.ts
@@ -23,6 +23,7 @@ import { contributions } from './contribs';
 import type { ExtensionInfo } from '../../../main/src/plugin/api/extension-info';
 
 export interface CombinedExtensionInfoUI extends ExtensionInfo {
+  // type is either 'pd' for Podman Desktop or 'dd' for 'Docker Desktop'
   type: 'dd' | 'pd';
 }
 
--- a/packages/renderer/src/stores/all-installed-extensions.ts
+++ b/packages/renderer/src/stores/all-installed-extensions.ts
@@ -1,0 +1,60 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


packages/renderer/src/stores/all-installed-extensions.ts
```diff
@@ -0,0 +1,60 @@
...
+ ***********************************************************************/
+
+// Combine PD and DD extensions being installed
+import { derived, type Readable } from 'svelte/store';
+import { extensionInfos } from './extensions';
+import { contributions } from './contribs';
+import type { ExtensionInfo } from '../../../main/src/plugin/api/extension-info';
+
+export interface CombinedExtensionInfoUI extends ExtensionInfo {
+  type: 'dd' | 'pd';
```
axel7083, line 27: Why using short names ? Feel a bit confusing to read

benoitf: to avoid any trademark issue

axel7083: I see, even adding a comment is risky ?

axel7083 (approved):

packages/renderer/src/stores/all-installed-extensions.ts
benoitf: I'll add the comment, it's OK

benoitf: @axel7083 comment added, does it work for you ?

axel7083: Yes perfect!! Thanks for it

Issue #6412, opened 2024-03-13
Move E2E test codebase under new sub folder to improve the readability


odockal:
### Is your enhancement related to a problem? Please describe

We could host more tests-related files under `tests` folder. And at the same time we want to make playwright based testing sources an npm package, so we should better align the folder structure.

### Describe the solution you'd like

Move `tests/*` files under `tests/playwright`. Move `spec` files under new `specs` folder. Update `package.json` and `vitest.config` files.

### Additional context

It will be required to also update places that expects these files on former locations. Ie. in podman-desktop-e2e repo, and ie. test or artifcats collector in GHA workflows.

PR #6413, opened 2024-03-13
chore(tests): move e2e test base into playwright subfolder

odockal:
### What does this PR do?
Moving e2e test base.
#6412 

### How to test this PR?
`yarn test:e2e`
Examine the `tests/output` for the test results and the output console. All tests should be passing. 
On all systems - windows, mac, linux. I have tried on Fedora 39 locally and smoke on GHA is also passing. 

- N/A Tests are covering the bug fix or the new feature


benoitf:
I went through sonarcloud to accept the 5 issues blocking the merge, 3 were TODO and 2 were for yaml used in tests so there is no need to have all cpus/etc data

tests/.gitignore
```diff
@@ -1 +1 @@
-output
+output
```
benoitf: not sure if we need this change it's the same content

tests/playwright/.gitignore
```diff
@@ -0,0 +1,3 @@
+dist
+node_modules
+output
```
benoitf: AFAIK it's excluded in upper gitignore files so we don't need it

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -55,15 +55,25 @@ export class PodmanDesktopRunner {
...
+      });
       this._app = await electron.launch({
         ...this._options,
       });
       // setup state
       this._page = await this.getElectronApp().firstWindow();
+      const exe = this.getElectronApp().evaluate(async ({ app }) => {
+        return app.getPath('exe');
+      });
+      console.log(`The Executable Electron app. file: ${exe}`);
```
benoitf: it looks like it's not only a move there

tests/playwright/src/setupFiles/extended-hooks.ts
```diff
@@ -0,0 +1,48 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

tests/playwright/src/setupFiles/extended-hooks.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import * as fs from 'fs';
```
benoitf:
```suggestion
import * as fs from 'node:fs';
```

tests/playwright/src/setupFiles/extended-hooks.ts
```diff
@@ -0,0 +1,48 @@
...
+import * as path from 'path';
```
benoitf:
```suggestion
import * as path from 'node:path';
```

tests/playwright/src/testContext/runner-test-context.ts
```diff
@@ -0,0 +1,24 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

tests/playwright/src/utility/cleanup.ts
```diff
@@ -25,8 +25,8 @@ import { existsSync, rmSync } from 'node:fs';
 export async function removeFolderIfExists(path: string): Promise<void> {
   console.log(`Cleaning up folder: ${path}`);
 
-  if (existsSync(path)) {
+  if (fs.existsSync(path)) {
     console.log(`Folder found, removing...`);
-    rmSync(path, { recursive: true, force: true, maxRetries: 5 });
+    fs.rmSync(path, { recursive: true, force: true, maxRetries: 5 });
```
benoitf: not sure it's better to import using the wildcard there ??

tests/playwright/src/testContext/runner-test-context.ts
benoitf: or 2023-2024

tests/playwright/src/setupFiles/extended-hooks.ts
benoitf: or 2023-2024

benoitf (commented):
overall LGTM, just few comments

nice job 👍

tests/.gitignore
odockal: I do not even understand why is this change here. In any case, we have now one .gitignore and that is in `tests/` folder. For now it will exclude `node_modules` from `playwright` and from what I got locally also from future's argos structure. Otherwise is ignores `output` folder recursively in actual directory and in any subfolder, same for `dist`.  We could add more specific rules later in shared or divided them in particular subfolder, right?

tests/playwright/src/runner/podman-desktop-runner.ts
odockal: Done

tests/playwright/src/utility/cleanup.ts
odockal: I have added specific imports.

odockal:
> I went through sonarcloud to accept the 5 issues blocking the merge, 3 were TODO and 2 were for yaml used in tests so there is no need to have all cpus/etc data

I tried to make the sonar cloud happy with pod specification, not sure wherether it will be for good.

tests/playwright/src/utility/cleanup.ts
benoitf: thanks

tests/.gitignore
benoitf: but `node_modules` and `dist` are excluded at the root level
https://github.com/containers/podman-desktop/blob/main/.gitignore#L1

https://github.com/containers/podman-desktop/blob/main/.gitignore#L3

so not sure why you need to exclude them again ?

benoitf:
it seems it does not pass anymore the smoke tests :-/

odockal:
@benoitf It looks like app is not started.

Update: Nah, my bad. should be fixed now.

cbr7 (approved):

benoitf (commented):
if you could remove dist and node_modules from test/.gitignore then we can merge

tests/.gitignore
odockal: ah, perfect then.

benoitf (approved):
thanks @odockal 👍

PR #6422, opened 2024-03-14
feat: add a safe storage registry

benoitf:
### What does this PR do?

introduce a registry storing key/values in encrypted and then encoded in base64 values

### Screenshot / video of UI

N/A

part 1 of https://github.com/containers/podman-desktop/issues/6044

### How to test this PR?

unit tests added

### DIFF
--- a/packages/main/src/plugin/safe-storage/safe-storage-registry.ts
+++ b/packages/main/src/plugin/safe-storage/safe-storage-registry.ts
@@ -39,14 +39,14 @@ export class SafeStorageRegistry {
     this.#directories = directories;
   }
 
-  protected getSafeStoragePath(): string {
+  protected getSafeStorageDataPath(): string {
     // create directory if it does not exist
     return path.resolve(this.#directories.getSafeStorageDirectory(), 'data.json');
   }
 
   // initialize the safe storage
   public async init(): Promise<void> {
-    const safeStoragePath = this.getSafeStoragePath();
+    const safeStoragePath = this.getSafeStorageDataPath();
 
     const parentDirectory = path.dirname(safeStoragePath);
     if (!existsSync(parentDirectory)) {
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/safe-storage/safe-storage-registry.ts
```diff
@@ -0,0 +1,144 @@
...
+export class SafeStorageRegistry {
+  readonly #directories: Directories;
+
+  #extensionStorage: SafeStorage | undefined;
+
+  constructor(directories: Directories) {
+    this.#directories = directories;
+  }
+
+  protected getSafeStoragePath(): string {
```
jeffmaury: Name is confusing something like getSafeStorageConfigPath

jeffmaury (approved):
LGTM. Wondering if having a specific separate storage is necessary, could'nt be part of the configuration API ?

benoitf:
the problem is that if you store that in the configuration registry, you'll be able to see the entries in the UI (in encrypted format)

packages/main/src/plugin/safe-storage/safe-storage-registry.ts
benoitf: ok but it's not the configuration path it's where the data are stored

benoitf: `getSafeStorageDataPath` ?

benoitf: fixed

benoitf:
rebased

benoitf:
> Base64 is solely an encoding mechanism, not a form of encryption. I can't find the follow up issues/explanation, is this temporary ?

@axel7083 so it's encoded in base64 just to be sure that we can easily read/write files / make it compliant with utf-8 etc

but before the encoding it's encrypted

axel7083:
> > Base64 is solely an encoding mechanism, not a form of encryption. I can't find the follow up issues/explanation, is this temporary ?
> 
> @axel7083 so it's encoded in base64 just to be sure that we can easily read/write files / make it compliant with utf-8 etc
> 
> but before the encoding it's encrypted

I deleted my comment after reading properly the code, yes reading your pull request changes made me realize the description was not what I thought

axel7083 (approved):

benoitf:
I've updated the description to make it clearer

PR #6424, opened 2024-03-14
fix: use expected component in the tests

benoitf:
### What does this PR do?
when working on a change in ExtensionStatus I noticed that some unrelated tests were breaking
it turns out that the test file is not testing the expected component
it is testing ExtensionStatus instead of ProviderStatus
fixing the test to match the component that we need to test

### Screenshot / video of UI

N/A

https://github.com/containers/podman-desktop/pull/6410

### How to test this PR?

unit test provided


odockal:
@benoitf Nice catch!

odockal (approved):

PR #6435, opened 2024-03-15
chore: add a rule to check for copyright headers

### LINKED ISSUE #6372: add header check linter rule

some files are missing the copyright headers

We should have a lint rule to enforce that so no need to have people having to manually review that

add a header estlint rule

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
add rule to enforce to have the license in .ts/.js files

fixes https://github.com/containers/podman-desktop/issues/6372

### How to test this PR?

PR check should be ✅ 


jeffmaury (approved):
LGTM.

Is there a way to check for the full header instead of just the license ID ?

benoitf:
@jeffmaury it seems it should be possible as this rule allow regexp but I noticed that if the regexp becomes too complex it's really really slowing down the check so it was like an intermediate step

Issue #6355, opened 2024-03-11
export and import containers


benoitf:
### Is your enhancement related to a problem? Please describe

Add ability to export the content of a container or a group of containers and restore them

like for images https://github.com/containers/podman-desktop/issues/478

### Describe the solution you'd like

I should be able to export containers and import them later

### Additional context

Need to have the methods exported/available for extension API as well

lstocchi:
@benoitf how should it work exactly? Do we add a button to the containers table, user click on export, we save the container filesystem into `.local\share\containers\podman-desktop\extensions-storage`, then we add an import action to the images page, user select the archive file and an image with the filesystem is created? I see that the import action also accepts some args. Do we want to support them in this first implementation?

benoitf:
I would add an export action for each of the lines

then a button (like the delete when you select multiple rows)

and probably an 'import' button to import again the files

about the location, I would not pickup `.local\share\containers\podman-desktop\extensions-storage` as it's dedicated for extensions.

Maybe it would be easier to prompt the user to select the location (maybe as step 1 we always ask, later we could have a properties to select folders, remember last folder etc)

For now let's try without the args and see how it behave

benoitf:
also I think that backup/restore of images is more important than the backup/restore of containers but I may be wrong

lstocchi:
@benoitf charlie said that he preferred to work on images and we switched issues. He will work on images

benoitf:
ok thanks for the notice 👍

lstocchi:
Closing as an initial implementation of export/import has been merged.

PR #6453, opened 2024-03-18
feat: add generic action to task

lstocchi:
### What does this PR do?

It adds a new property to a task to implement a custom behavior like show an 'open folder' action.
This will be used with the export container action. When the export is executing/has been executed the user could open the taskmanager, click on the open folder link to open the directory where the container content will be saved

### Screenshot / video of UI

N/A

it is part of #6355 

### How to test this PR?

1. run test

### DIFF
--- a/packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
+++ b/packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
@@ -55,8 +55,8 @@ function gotoTask(taskUI: StatefulTaskUI) {
   taskUI?.gotoTask?.();
 }
 
-function openFolder(taskUI: StatefulTaskUI) {
-  taskUI?.openFolder?.();
+function doExecuteAction(taskUI: StatefulTaskUI) {
+  taskUI?.action?.execute();
 }
 </script>
 
@@ -118,13 +118,13 @@ function openFolder(taskUI: StatefulTaskUI) {
     {#if isStatefulTask(taskUI) && taskUI.status !== 'failure'}
       <div class="flex flex-row w-full">
         <div class="flex flex-1 flex-col w-full items-end text-purple-500 text-xs">
-          {#if taskUI.openFolder}
+          {#if taskUI.action}
             <button
               class="text-purple-500 cursor-pointer"
               on:click="{() => {
-                if (isStatefulTask(taskUI)) openFolder(taskUI);
+                if (isStatefulTask(taskUI)) doExecuteAction(taskUI);
               }}"
-              aria-label="open folder">Open folder &gt;</button>
+              aria-label="action button">{taskUI.action.name}</button>
           {/if}
         </div>
       </div>
--- a/packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
+++ b/packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
@@ -124,7 +124,7 @@ function openFolder(taskUI: StatefulTaskUI) {
               on:click="{() => {
                 if (isStatefulTask(taskUI)) openFolder(taskUI);
               }}"
-              aria-label="open folder">Open folder ></button>
+              aria-label="open folder">Open folder &gt;</button>
           {/if}
         </div>
       </div>
--- a/packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
+++ b/packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
@@ -54,6 +54,10 @@ function gotoTask(taskUI: StatefulTaskUI) {
   // and open the task
   taskUI?.gotoTask?.();
 }
+
+function openFolder(taskUI: StatefulTaskUI) {
+  taskUI?.openFolder?.();
+}
 </script>
 
 <!-- Display a task item-->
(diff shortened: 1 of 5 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -1500,7 +1500,7 @@ declare module '@podman-desktop/api' {
     /**
      * The resource the dialog shows when opened.
      */
-    defaultUri?: Uri;
+    defaultUri?: Uri | string;
```
jeffmaury: Can't we keep it as Uri only and assume that it is a file uri ?

lstocchi: What do you mean exactly? Bc i would have to call the openDialog from the client. So i would have something like `window.openDialog(defaultUri: mypath)`. I could create a fake Uri myself with 
```
scheme: file,
path: mypath,
fspath: mypath, 
....
```
but it's a waste imo. In the openDialog it's going to take the fsPath in any case

jeffmaury: I meant the API is already designed to open a file or folder so why do you have to update the API ?

lstocchi: bc there was no way to create a URI item from the frontend and bc the defaultUri can be basically be a string i was extending it. I'm going to create another PR to expose the createUri function and revert it

lstocchi: opened https://github.com/containers/podman-desktop/pull/6454 . I'll use that when calling the openDialog

packages/renderer/src/lib/task-manager/task-manager.ts
```diff
@@ -40,6 +40,7 @@ export class TaskManager {
         status: task.status,
         hasGotoTask: false,
         age: `${humanizeDuration(new Date().getTime() - task.started, { round: true, largest: 1 })} ago`,
+        openFolder: task.openFolder,
```
benoitf: maybe a test in task-manager.spec.ts ?

packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
```diff
@@ -111,6 +115,21 @@ function gotoTask(taskUI: StatefulTaskUI) {
...
+    {#if isStatefulTask(taskUI) && taskUI.status !== 'failure'}
+      <div class="flex flex-row w-full">
+        <div class="flex flex-1 flex-col w-full items-end text-purple-500 text-xs">
+          {#if taskUI.openFolder}
+            <button
+              class="text-purple-500 cursor-pointer"
+              on:click="{() => {
+                if (isStatefulTask(taskUI)) openFolder(taskUI);
+              }}"
+              aria-label="open folder">Open folder ></button>
```
benoitf:
```suggestion
              aria-label="open folder">Open folder ></button>
```

is that the `>` after Open Folder is wanted ?

lstocchi: I did that on purpose to follow the open task but it can be removed.
open_folder

benoitf: maybe use `&gt;` so we know it's an html entity ?

packages/renderer/src/lib/task-manager/task-manager.ts
lstocchi: Done

packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
lstocchi: done

benoitf (approved):

deboer-tim:
I'm finding it hard to follow the comments and changes afterwards, but if this went from a 'folder URI' to a function -which could do anything - why keep the property named 'folder'? Wouldn't it make sense to allow things like 'open file' or 'open logs' as well? Maybe even 'open browser'? Those seem just as likely for other tasks, and would only require a label as well.

lstocchi:
> I'm finding it hard to follow the comments and changes afterwards, but if this went from a 'folder URI' to a function -which could do anything - why keep the property named 'folder'? Wouldn't it make sense to allow things like 'open file' or 'open logs' as well? Maybe even 'open browser'? Those seem just as likely for other tasks, and would only require a label as well.

It was always a function, the thing discussed was how to pass the info ( as a string or as a Uri) to the backend. 
BTW yes, it could be used for anything. But the same for the goToTask.
So you would suggest to have something generic like` action: { title, function }`, so that the title is the text displayed and function is the one executed when clicking?

deboer-tim:
> > I'm finding it hard to follow the comments and changes afterwards, but if this went from a 'folder URI' to a function -which could do anything - why keep the property named 'folder'? Wouldn't it make sense to allow things like 'open file' or 'open logs' as well? Maybe even 'open browser'? Those seem just as likely for other tasks, and would only require a label as well.
> 
> It was always a function, the thing discussed was how to pass the info ( as a string or as a Uri) to the backend. BTW yes, it could be used for anything. But the same for the goToTask. So you would suggest to have something generic like` action: { title, function }`, so that the title is the text displayed and function is the one executed when clicking?

Yes, I think that's what I'm suggesting. Maybe it is even worth replacing goToTask (not necessarily in this PR...) or allowing more than one. We should keep some consistency in the labels, but this gives any task a simple and generic way to link to relevant things, whether it's a status page, logs, a folder, help, etc.

lstocchi:
@deboer-tim updated 👍

benoitf (approved):

deboer-tim (approved):

deboer-tim:
Thanks for the change. 👍🏼

jeffmaury (approved):
LGTM

PR #6454, opened 2024-03-18
chore: make saveDialog function returns Uri

lstocchi:
### What does this PR do?

it makes the saveDialog function returns the Uri instead of a string so that it can be re-used if we need to call the openDialog function from a task -> https://github.com/containers/podman-desktop/pull/6453

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

N/A

### DIFF
--- a/packages/main/src/plugin/dialog-registry.ts
+++ b/packages/main/src/plugin/dialog-registry.ts
@@ -20,6 +20,7 @@ import type { OpenDialogOptions, SaveDialogOptions } from '@podman-desktop/api';
 import type { BrowserWindow } from 'electron';
 import { dialog } from 'electron';
 
+import { Uri } from './types/uri.js';
 import type { Deferred } from './util/deferred.js';
 
 /**
@@ -88,7 +89,7 @@ export class DialogRegistry {
     }
   }
 
-  async saveDialog(options?: SaveDialogOptions, dialogId?: string): Promise<string | undefined> {
+  async saveDialog(options?: SaveDialogOptions, dialogId?: string): Promise<Uri | undefined> {
     if (!this.#browserWindow) {
       throw new Error('Browser window is not available');
     }
@@ -115,11 +116,12 @@ export class DialogRegistry {
     if (response.filePath && !response.canceled) {
       filePath = response.filePath;
     }
+    const fileUri = filePath ? Uri.file(filePath) : undefined;
     // send the response to the renderer part if dialogId is provided
     if (dialogId) {
-      this.#browserWindow.webContents.send('dialog:open-save-dialog-response', dialogId, filePath);
+      this.#browserWindow.webContents.send('dialog:open-save-dialog-response', dialogId, fileUri);
     } else {
-      return filePath;
+      return fileUri;
     }
   }
 }
--- a/packages/main/src/plugin/dialog-registry.ts
+++ b/packages/main/src/plugin/dialog-registry.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import type { OpenDialogOptions, SaveDialogOptions } from '@podman-desktop/api';
+import type { OpenDialogOptions, SaveDialogOptions, Uri as APIUri } from '@podman-desktop/api';
 import type { BrowserWindow } from 'electron';
 import { dialog } from 'electron';
 
@@ -89,7 +89,7 @@ export class DialogRegistry {
     }
   }
 
-  async saveDialog(options?: SaveDialogOptions, dialogId?: string): Promise<Uri | undefined> {
+  async saveDialog(options?: SaveDialogOptions, dialogId?: string): Promise<APIUri | undefined> {
     if (!this.#browserWindow) {
       throw new Error('Browser window is not available');
     }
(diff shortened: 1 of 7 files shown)


benoitf (commented):
I'm not sure to follow when the frontend need to build a file URI

the file URI are usually brought by the backend

lstocchi:
> I'm not sure to follow when the frontend need to build a file URI
> 
> the file URI are usually brought by the backend

My idea was to add an open folder button to the task, so let's say you run the export container, then do something else, then want to retrieve it, maybe you don't remember which folder you selected and you can open it with a click from the task result.
 
open_folder

benoitf:
I understand the feature to open a folder but what I'm not yet understanding is what frontend need to get an Uri from a file.

I would think that backend has the file information and returns the Uri back to the frontend (for example when doing the export)

lstocchi:
> I understand the feature to open a folder but what I'm not yet understanding is what frontend need to get an Uri from a file.
> 
> I would think that backend has the file information and returns the Uri back to the frontend (for example when doing the export)

In my mind the open folder button was active even when the action was progressing. So if the export takes time, you can already open the output folder and see if the file is being created. 
But if you prefer i can make the exportContainer returns the Uri and show it only when it's done.

benoitf:
> In my mind the open folder button was active even when the action was progressing. So if the export takes time, you can already open the output folder and see if the file is being created.
But if you prefer i can make the exportContainer returns the Uri and show it only when it's done.


when you call saveDialog to ask the user where to export containers, at that time we could return the Uri instead of the file
(I think we're doing the change file -> uri in extension-loader but we could do that in the global way)

so when you call showOpenFolder you have the Uri and then you can ask to open a folder from that Uri ? (no need to have it in exportContainer)

note: maybe we'll need something to revive the object as I'm not sure it can be transferred from backend to frontend

lstocchi:
> when you call saveDialog to ask the user where to export containers, at that time we could return the Uri instead of the file (I think we're doing the change file -> uri in extension-loader but we could do that in the global way)
> 
> so when you call showOpenFolder you have the Uri and then you can ask to open a folder from that Uri ? (no need to have it in exportContainer)

Ah i was using the openDialog tbh, maybe the saveDialog could be better and save me some code in the exportContainers.
So, basically, saveDialog should just return the Uri instead of string or should i create a new function? I see that it is already used in the troubleshootingGatherLogs.svelte

benoitf:
>Ah i was using the openDialog tbh, maybe the saveDialog could be better and save me some code in the exportContainers.

Yes I think saveDialog is better as you'll have something that fit better for selecting a folder/file where we want to save data.

> So, basically, saveDialog should just return the Uri instead of string or should i create a new function? I see that it is already used in the troubleshootingGatherLogs.svelte


just the Uri and fix `troubleshootingGatherLogs.svelte`

but before, check how serialization is working as maybe it doesn't work well.
for extensions it's working fine as we're in backend/backend world but here I don't know how it'll work

benoitf:
you can also start by changing the type in the exported function in index.ts file, not the function itself (like we do in extension-loader, etc)

lstocchi:
@benoitf updated so that it changes the returning type of saveDialog

packages/renderer/src/lib/troubleshooting/TroubleshootingGatherLogs.svelte
```diff
@@ -4,14 +4,17 @@ import Fa from 'svelte-fa';
...
+
 let logs: string[] = [];
 
 // Save files as a zip file (we first ask the user for the dialog, and then save the files to the filepath)
 async function saveLogsAsZip() {
   const defaultUri = await window.troubleshootingGenerateLogFileUri('podman-desktop', 'zip');
   const filePath = await window.saveDialog({ title: 'Save Logs as .zip', defaultUri });
   if (filePath) {
-    logs = await window.troubleshootingSaveLogs(filePath);
+    const filePathUri = Uri.revive(filePath);
```
benoitf, line 16: Is it possible to have revive being done in the exposure of saveDialog so we don't need that after each call ? Maybe it's not possible, just asking

benoitf: edited my comment

lstocchi: I tried to put the Uri class in the preload without any success. It still takes the private props as keys.
But if i create a revive function that just returns an object like 
```
  static revive(serialized: APIUri): APIUri {
    if (serialized instanceof Uri) {
      return serialized;
    }
    const serializedProps: Map<string, string> = Object.entries(serialized)
      .map(([key, value]) => [key.startsWith('_') ? key.substring(1) : key, value])
      .reduce((map, [key, value]) => {
        map.set(key, value);
        return map;
      }, new Map());
    return {
      scheme: serializedProps.get('scheme') ?? '',
      authority: serializedProps.get('authority') ?? '',
      path: serializedProps.get('path') ?? '',
      query: serializedProps.get('query') ?? '',
      fragment: serializedProps.get('fragment') ?? '',
    } as APIUri;
  }
```
it works fine. Do i do this?

benoitf: no I think it won't fit well, keep it as you have

benoitf (approved):

packages/main/src/plugin/dialog-registry.ts
```diff
@@ -88,7 +89,7 @@ export class DialogRegistry {
     }
   }
 
-  async saveDialog(options?: SaveDialogOptions, dialogId?: string): Promise<string | undefined> {
+  async saveDialog(options?: SaveDialogOptions, dialogId?: string): Promise<Uri | undefined> {
```
benoitf: should return podmanDesktopApi.Uri ?

Issue #5505, opened 2024-01-11
Show running tasks in status bar


deboer-tim:
### Is your enhancement related to a problem? Please describe

If I start a task and switch to another view, there's nothing in the UI that shows me whether the task is still running, or when it has finished.

### Describe the solution you'd like

There have been some previous designs, but (e.g.) putting a purple dot when there is a change to tasks is not enough. This is one of the previous design that I prefer:

The notification bell should move to be the left-most icon on the status bar, so that the space directly to its left can be reserved for running tasks. (not sure where to move the version string)

If there is a task running, the latest running task name should be shown, along with a small progress bar and/or percentage. If you hover, the task name + notification bell is a single button that opens up the task manager (same as today, just wider selection area).

Here's a rough mockup of what it might look like (but without the notification bell move):

https://github.com/containers/podman-desktop/assets/19958075/507a31c9-63ad-4289-a04e-b9cfcfa87140

deboer-tim:
Still an issue.

PR #6468, opened 2024-03-20
feat: add export container

lstocchi:
### What does this PR do?

It adds the export container action

### Screenshot / video of UI

export_container

it is part of #6355 

### How to test this PR?

1. try to export a container and see it is actually saved in the filesystem

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -19,7 +19,6 @@
 import * as crypto from 'node:crypto';
 import { EventEmitter } from 'node:events';
 import * as fs from 'node:fs';
-import path from 'node:path';
 import { type Stream, Writable } from 'node:stream';
 import { pipeline } from 'node:stream/promises';
 
@@ -2405,23 +2404,10 @@ export class ContainerProviderRegistry {
       throw new Error('no running provider for the matching container');
     }
 
-    // generate the name of the exported file
-    // if there are other entries with the same name, let's append (number of occurences)
-    let containerFile = options.name;
-    const regexFile = new RegExp(`${containerFile}\\s+\\(+\\d+\\)$`);
-    const entries = await fs.promises.readdir(options.outputDirectory, { withFileTypes: true });
-    const similarContainers = entries
-      .filter(entry => entry.isFile())
-      .filter(file => file.name === containerFile || regexFile.test(file.name));
-
-    if (similarContainers.length > 0) {
-      containerFile += ` (${similarContainers.length})`;
-    }
-
     // retrieve the container and export it by copying the content to the final destination
     const containerObject = engine.api.getContainer(options.id);
     const exportResult = await containerObject.export();
-    const fileWriteStream = fs.createWriteStream(path.join(options.outputDirectory, containerFile), {
+    const fileWriteStream = fs.createWriteStream(options.outputTarget, {
       flags: 'w',
     });
 
--- a/packages/renderer/src/lib/container/ContainerActions.svelte
+++ b/packages/renderer/src/lib/container/ContainerActions.svelte
@@ -211,6 +211,7 @@ if (dropdownMenu) {
     icon="{faArrowsRotate}" />
   <ListItemButtonIcon
     title="Export Container"
+    tooltip="Exports container's filesystem contents as a tar archive and saves it on the local machine"
     onClick="{() => exportContainer()}"
     menu="{dropdownMenu}"
     detailed="{detailed}"
--- a/packages/renderer/src/lib/container/ContainerActions.svelte
+++ b/packages/renderer/src/lib/container/ContainerActions.svelte
@@ -2,6 +2,7 @@
 import {
   faAlignLeft,
   faArrowsRotate,
+  faDownload,
   faExternalLinkSquareAlt,
   faFileCode,
   faPlay,
@@ -14,6 +15,7 @@ import { createEventDispatcher, onMount } from 'svelte';
 import { router } from 'tinro';
 
 import ContributionActions from '/@/lib/actions/ContributionActions.svelte';
+import { exportContainerInfo } from '/@/stores/export-container-store';
 
 import type { Menu } from '../../../../main/src/plugin/menu-registry';
 import { MenuContext } from '../../../../main/src/plugin/menu-registry';
@@ -104,6 +106,11 @@ async function deleteContainer(): Promise<void> {
   } finally {
     inProgress(false);
   }
+}
+
+async function exportContainer() {
+  exportContainerInfo.set(container);
+  router.goto('/containers/export');
 }
 
 function openTerminalContainer(): void {
@@ -202,6 +209,12 @@ if (dropdownMenu) {
     menu="{dropdownMenu}"
     detailed="{detailed}"
     icon="{faArrowsRotate}" />
+  <ListItemButtonIcon
+    title="Export Container"
+    onClick="{() => exportContainer()}"
+    menu="{dropdownMenu}"
+    detailed="{detailed}"
+    icon="{faDownload}" />
   <ContributionActions
     args="{[container]}"
     contextPrefix="containerItem"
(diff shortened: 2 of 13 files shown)


packages/renderer/src/lib/container/ContainerActions.svelte
```diff
@@ -202,6 +209,12 @@ if (dropdownMenu) {
     menu="{dropdownMenu}"
     detailed="{detailed}"
     icon="{faArrowsRotate}" />
+  <ListItemButtonIcon
+    title="Export Container"
```
benoitf, line 213: the full length is 'Exports container's filesystem contents as a tar archive and saves it on the local machine'

I don't know if we can add a custom tooltip there ?

packages/renderer/src/stores/export-container-store.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import type { Writable } from 'svelte/store';
+import { writable } from 'svelte/store';
+
+import type { ContainerInfoUI } from '../lib/container/ContainerInfoUI';
+
+/**
+ * Defines the store used to define the container to export.
+ */
+export const exportContainerInfo: Writable<ContainerInfoUI> = writable();
```
benoitf, line 27: I know it's a dummy one but maybe add a test there ?

benoitf (commented):
the export on more than one container will come in follow-up PRs ?

packages/main/src/plugin/container-registry.ts
```diff
@@ -2405,23 +2404,10 @@ export class ContainerProviderRegistry {
...
-    const regexFile = new RegExp(`${containerFile}\\s+\\(+\\d+\\)$`);
-    const entries = await fs.promises.readdir(options.outputDirectory, { withFileTypes: true });
-    const similarContainers = entries
-      .filter(entry => entry.isFile())
-      .filter(file => file.name === containerFile || regexFile.test(file.name));
-
-    if (similarContainers.length > 0) {
-      containerFile += ` (${similarContainers.length})`;
-    }
-
```
benoitf, line 2420: if the file already exists we overwrite the content now ?

lstocchi: Yes.
In my first implementation i was using the openDialog to make the user select the output folder + there was a textbox where to insert the name of the output file (then i was combining the two and if the file already existed i was adding a `(1)` suffix). 
Now i switched to a saveDialog. So if you try to use the name of an already existing file, it already shows you a warning "do you want to replace the existing file?" (or something like that), so the user is already informed. This is why i removed this part.

benoitf: got it

benoitf (approved):

lstocchi:
> the export on more than one container will come in follow-up PRs ?

I missed this comment.
Yes, i'm not actually sure how this should work. At first i thought about asking the user to select the folder and download any single container file by using its own name (and rename it if it was already in the filesystem - like the code removed above). But now that i used the saveDialog i thought that the user can create a sort of archive file or folder and everything will be downloaded in it with the original container file. Any preference?
BTW i was working on the import action so atleast we have a working export-import on a single file.

cdrage:
May have to follow up with UX team, but I believe if it's non-form, it should be a modal instead similar to push images. 

Anything more complicated than 3-4 actions I believe was suppose to be a form, but anything very simple is a modal. 

We could use the same template for push images for this.

lstocchi:
> May have to follow up with UX team, but I believe if it's non-form, it should be a modal instead similar to push images.
> 
> Anything more complicated than 3-4 actions I believe was suppose to be a form, but anything very simple is a modal.
> 
> We could use the same template for push images for this.

Are you talking about the multiple export or this single one? I didn't use a modal here bc there is the saveDialog which opens another dialog so it looked to me too annoying to have two opened up

benoitf:
> But now that i used the saveDialog i thought that the user can create a sort of archive file or folder and everything will be downloaded in it with the original container file. Any preference?
BTW i was working on the import action so atleast we have a working export-import on a single file.

I thought the command was allowing multiple containers (like podman save where you can provide multiple images)

on a preference, I'll go to something simple like if I select a folder it'll export each selected container to that folder 


and then on the import if we select multiple files it'll import multiple containers

lstocchi:
> I thought the command was allowing multiple containers (like podman save where you can provide multiple images)

No, it looks to me that it only accept one item so we need to call it for every single selected item

> on a preference, I'll go to something simple like if I select a folder it'll export each selected container to that folder
> 
> and then on the import if we select multiple files it'll import multiple containers

Ok

packages/renderer/src/lib/container/ContainerExport.svelte
```diff
@@ -0,0 +1,120 @@
...
+            <Input
+              class="grow mr-2"
+              name="{container.id}"
+              readonly
+              value="{outputTarget}"
+              id="input-export-container-name"
+              aria-invalid="{invalidFolder}" />
+            <Button
+              on:click="{() => selectFolderPath()}"
+              icon="{faFolderOpen}"
```
deboer-tim: We don't use icons on Browse buttons anywhere else. There's a new design where we wouldn't even have the Button, but in the meantime I think we should be consistent and not include the icon.

packages/renderer/src/lib/container/ContainerExport.svelte
```diff
@@ -0,0 +1,120 @@
...
+
+{#if container}
+  <FormPage title="Export container {container.name}">
+    <svelte:fragment slot="icon">
+      <i class="fas fa-download fa-2x" aria-hidden="true"></i>
+    </svelte:fragment>
+
+    <div slot="content" class="flex flex-col min-w-full h-fit p-5">
+      <div class="bg-charcoal-600 px-6 py-4 space-y-2 lg:px-8 sm:pb-6 xl:pb-8">
+        <div>
```
deboer-tim, line 95: When I test there is a double-underline: one in the Input component, and one all the way underneath the Input and Button. This second one should be removed.

packages/renderer/src/lib/container/ContainerExport.svelte
```diff
@@ -0,0 +1,120 @@
...
+  container = $exportContainerInfo;
+  if (!container) {
+    // go back to containers list
+    router.goto('/containers/');
+  }
+});
+
+async function selectFolderPath() {
+  const result = await window.saveDialog({
+    title: 'Select the directory where to export the container content',
```
deboer-tim, line 37: By default the output folder is 'Unititled'. Could we change this to the name of the image (or at least be unique)?

packages/renderer/src/lib/container/ContainerExport.svelte
```diff
@@ -0,0 +1,120 @@
...
+              aria-invalid="{invalidFolder}" />
+            <Button
+              on:click="{() => selectFolderPath()}"
+              icon="{faFolderOpen}"
+              title="Open dialog to select the output file"
+              aria-label="Select output file">Browse ...</Button>
+          </div>
+          <div class="pt-5 border-zinc-600 border-t-2"></div>
+          <Button
+            on:click="{() => exportContainer()}"
```
deboer-tim, line 111: When I click on the button, nothing happens. I can click on it multiple times, and then suddenly the form disappears. I don't know if anything worked until I go to the tasks view and see multiple export jobs doing the same thing. Button should be inProgress once you click and there should be some indication that a background task has been started before the form disappears.

deboer-tim (changes requested):
Function worked great 👍🏼 but noted a few usability/UI issues.

lstocchi:
> Function worked great 👍🏼 but noted a few usability/UI issues.

I should have fixed everything. Please give it another try!! Thanks!

lstocchi:
@benoitf clicked the wrong button, i wanted to request @deboer-tim review. But you can give it another look if you want 😆

benoitf (approved):

deboer-tim (approved):
Thanks for making the changes 👍🏼. Working perfectly for me, and the only usability issue is that the form disappears and user doesn't know if it's done / running in the background / what happened, but that is not export-specific and we already have issues open for that (e.g. #5505).

PR #6476, opened 2024-03-20
feat: add experimental flag to install podman v5

### LINKED ISSUE #6375: should allow to install podman v5 using an experimental flag

To make the migration from v4 to v5, it should be possible to have podman v5 embedded so we can do the tests in term of release, fetching artifacts, etc.

add an experimental flag to enable Podman v5 installer

if enabled it should allow to update to v5, do all the save/restore flow, etc.

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
if toggling the flag in preferences, it'll register the 'update to v5.0.0'

depends on this PR
- [x] https://github.com/containers/podman-desktop/pull/6475
for better result

### Screenshot / video of UI

image

image

fixes https://github.com/containers/podman-desktop/issues/6375

### How to test this PR?

you can run Podman Desktop and toggle the flag then you should see "Update to 5.0.0" flag
if you don't see it immediately, refresh the UI of the client or apply https://github.com/containers/podman-desktop/pull/6475 before

### DIFF
--- a/extensions/podman/package.json
+++ b/extensions/podman/package.json
@@ -39,6 +39,11 @@
           "default": "",
           "description": "Custom path to Podman binary (Default is blank)"
         },
+        "podman.experimental.install.v5": {
+          "type": "boolean",
+          "default": false,
+          "description": "Enable podman v5 installation support (experimental)"
+        },
         "podman.machine.cpus": {
           "type": "number",
           "format": "cpu",
--- a/extensions/podman/src/extension.spec.ts
+++ b/extensions/podman/src/extension.spec.ts
@@ -23,10 +23,10 @@ import type { Configuration, ContainerEngineInfo, ContainerProviderConnection } 
 import * as extensionApi from '@podman-desktop/api';
 import { Disposable } from '@podman-desktop/api';
 import type { Mock } from 'vitest';
-import { afterEach, beforeEach, expect, test, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import { DarwinSocketCompatibility } from './compatibility-mode';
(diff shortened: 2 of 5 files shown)


lstocchi (approved):
LGTM
I have some other issues like i cannot create a podman machine 😓 but it installed podman 5 correctly

deboer-tim (approved):
It worked for me, and I was able to install Podman 5 (and create/start a machine).

Only nit: "V5" isn't a great name for the property in the UI. Something like 'experimental' would work better, but then wouldn't be a v5 specific property so may not be a better choice.

benoitf:
ExperimentalV5 would be better as property name displayed in the UI ?

vzhukovs:
Maybe `V5` -> `Experimental Support` and get rid of `(experimental)` in property description? Wdyt?

benoitf:
@vzhukovs it displays only the last 'word' so it'll display either `V5` or `Experimental` if we call it foo.experimental.v5 or foo.experimental.v5

deboer-tim:
This is why I had an issue open a while ago to be able to supply a 'display name' 😆. V5 or ExperimentalV5 are likely the best we can do and I'm fine with either.

vzhukovs:
> @vzhukovs it displays only the last 'word' so it'll display either `V5` or `Experimental` if we call it foo.experimental.v5 or foo.experimental.v5

Then, V5 sounds better

cdrage:
> > @vzhukovs it displays only the last 'word' so it'll display either `V5` or `Experimental` if we call it foo.experimental.v5 or foo.experimental.v5
> 
> Then, V5 sounds better

👍 on just V5

benoitf:
ok so it seems the consensus is that I keep as it is

I'll merge tomorrow unless someone will request changes

Issue #6479, opened 2024-03-20
e2e-tests main workflow is failing on GHA CI with test failures


odockal:
### Bug description

I can see some CI failures on e2e tests main workflow: https://github.com/containers/podman-desktop/actions/runs/8360043287/job/22884780014.

It seems like we missed some update in the UI.

### Operating system

Ubuntu 22.04

### Installation Method

None

### Version

next (development version)

### Relevant log output

```shell
Error: Timed out 5000ms waiting for expect(locator).toHaveText(expected)

Locator: getByLabel('Connection Status Label')
Expected string: "ENABLED"
Received string: "ACTIVE"
Call log:
  - locator._expect with timeout 5000ms
  - waiting for getByLabel('Connection Status Label')
  -   locator resolved to <span aria-label="Connection Status Label" class="my-a…>ACTIVE</span>
  -   unexpected value "ACTIVE"
  -   locator resolved to <span aria-label="Connection Status Label" class="my-a…>ACTIVE</span>
  -   unexpected value "ACTIVE"
```

odockal:
At the same time, we need to make sure to run the full suite during PR in a non-blocking way.

odockal:
#6494 as a CI follow up.

PR #6491, opened 2024-03-21
fix: update e2e tests to fix CI failures

odockal:
### What does this PR do?
Fixes e2e-tests.
#6479 

### How to test this PR?

### DIFF
--- a/tests/playwright/src/model/pages/dashboard-page.ts
+++ b/tests/playwright/src/model/pages/dashboard-page.ts
@@ -30,18 +30,15 @@ export class DashboardPage extends BasePage {
   // dev sandbox
   readonly devSandboxProvider: Locator;
   readonly devSandboxBox: Locator;
-  readonly devSandboxActualState: Locator;
   readonly devSandboxStatusLabel: Locator;
   // openshift local
   readonly openshiftLocalProvider: Locator;
   readonly openshiftLocalBox: Locator;
-  readonly openshiftLocalActualState: Locator;
   readonly openshiftLocalStatusLabel: Locator;
   readonly transitioningState: Locator;
 
   // podman/machine
   readonly podmanProvider: Locator;
-  readonly podmanActualState: Locator;
   readonly podmanStatusLabel: Locator;
   readonly podmanInitilizeAndStartButton: Locator;
 
@@ -62,21 +59,18 @@ export class DashboardPage extends BasePage {
     // Dev Sandbox locators
     this.devSandboxProvider = page.getByRole('region', { name: 'Developer Sandbox Provider' });
     this.devSandboxBox = this.featuredExtensions.getByLabel('Developer Sandbox');
-    this.devSandboxActualState = this.devSandboxProvider.getByLabel(this.ACTUAL_STATE);
-    this.devSandboxStatusLabel = this.devSandboxActualState.getByLabel(this.CONNECTION_STATUS_LABEL);
+    this.devSandboxStatusLabel = this.devSandboxProvider.getByLabel(this.CONNECTION_STATUS_LABEL);
 
     // OpenShift Local locators
     this.openshiftLocalProvider = page.getByRole('region', { name: 'OpenShift Local Provider' });
     this.openshiftLocalBox = this.featuredExtensions.getByLabel('OpenShift Local');
-    this.openshiftLocalActualState = this.openshiftLocalProvider.getByLabel(this.ACTUAL_STATE);
     this.openshiftLocalStatusLabel = this.openshiftLocalProvider.getByLabel(this.CONNECTION_STATUS_LABEL);
 
     // Podman/Machine Provider locators
     this.podmanProvider = page.getByRole('region', { name: 'Podman Provider' });
     this.podmanInitilizeAndStartButton = this.podmanProvider.getByRole('button', { name: 'Initialize and start ' });
     this.transitioningState = this.podmanProvider.getByLabel('Transitioning State');
-    this.podmanActualState = this.podmanProvider.getByLabel(this.ACTUAL_STATE);
-    this.podmanStatusLabel = this.podmanActualState.getByLabel(this.CONNECTION_STATUS_LABEL);
+    this.podmanStatusLabel = this.podmanProvider.getByLabel(this.CONNECTION_STATUS_LABEL);
   }
 
   public getPodmanStatusLocator(): Locator {
--- a/tests/playwright/src/model/pages/dashboard-page.ts
+++ b/tests/playwright/src/model/pages/dashboard-page.ts
@@ -27,15 +27,27 @@ export class DashboardPage extends BasePage {
   readonly heading: Locator;
   readonly notificationsBox: Locator;
   readonly featuredExtensions: Locator;
+  // dev sandbox
   readonly devSandboxProvider: Locator;
   readonly devSandboxBox: Locator;
-  readonly devSandboxEnabledStatus: Locator;
+  readonly devSandboxActualState: Locator;
+  readonly devSandboxStatusLabel: Locator;
+  // openshift local
   readonly openshiftLocalProvider: Locator;
   readonly openshiftLocalBox: Locator;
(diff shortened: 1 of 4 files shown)


tests/playwright/src/model/pages/dashboard-page.ts
```diff
@@ -45,18 +57,26 @@ export class DashboardPage extends BasePage {
...
-    this.transitioningState = page.getByLabel('Transitioning State');
-    this.podmanMachineConnectionStatus = page.getByLabel('Connection Status Label');
+    this.openshiftLocalActualState = this.openshiftLocalProvider.getByLabel(this.ACTUAL_STATE);
+    this.openshiftLocalStatusLabel = this.openshiftLocalProvider.getByLabel(this.CONNECTION_STATUS_LABEL);
+
+    // Podman/Machine Provider locators
+    this.podmanProvider = page.getByRole('region', { name: 'Podman Provider' });
+    this.podmanInitilizeAndStartButton = this.podmanProvider.getByRole('button', { name: 'Initialize and start ' });
+    this.transitioningState = this.podmanProvider.getByLabel('Transitioning State');
+    this.podmanActualState = this.podmanProvider.getByLabel(this.ACTUAL_STATE);
```
cbr7: I'm not sure I understand whatt he goal of this field is, why do we need it? It seems to me the connection status label from below would work just as well from podmanProvider field, I'm not sure I understand the need to abstract so many middle layers.

tests/playwright/src/specs/extension-installation.spec.ts
```diff
@@ -30,6 +30,11 @@ import { WelcomePage } from '../model/pages/welcome-page';
 import { PodmanDesktopRunner } from '../runner/podman-desktop-runner';
 import type { RunnerTestContext } from '../testContext/runner-test-context';
 
+const DISABLED = 'DISABLED';
```
cbr7, line 33: why not create an enum with all these states?

tests/playwright/src/specs/podman-machine-dashboard-onboarding.spec.ts
```diff
@@ -62,9 +62,10 @@ describe.skipIf(os.platform() === 'linux')(async () => {
       test('Create Podman machine from Dashboard', async () => {
         const navigationBar = new NavigationBar(page);
         const dashboardPage = await navigationBar.openDashboard();
-        await playExpect(dashboardPage.initilizeAndStartButton).toBeEnabled();
-        await dashboardPage.initilizeAndStartButton.click();
-        await playExpect(dashboardPage.podmanMachineConnectionStatus).toHaveText('RUNNING', { timeout: 300000 });
+        await playExpect(dashboardPage.podmanProvider).toBeVisible();
```
cbr7: Why is this assertion needed? The initialize and start button is a component of that component, if we're waiting for it to be enabled it stands to reason that the higher level component exists also. IS there some race condition this asset solves that I'm not seeing?

tests/playwright/src/model/pages/dashboard-page.ts
odockal: There could be more providers with the same button available in the Dashboard (openshift local). So it is better to hierarchically group them. You have provider region (big tiles on dashboard) for podman, dev sandbox, etc., then provider have these elements like buttons and statuses. It is much more readable and logically tied together.

tests/playwright/src/specs/extension-installation.spec.ts
odockal: Agree, that would be for a more work, like refactoring.

tests/playwright/src/model/pages/dashboard-page.ts
cbr7: But there can't be more fields with the "podman Provider" label, as such any field that is extracted from podmanProvider would be unique, as such abstracting the Actual State and then the connection label is just adding more abstractions for no benefit.

My point is that "Actual state" field has no function and that extracting connection label from podmanProvider is sufficient to never be a duplicate.

tests/playwright/src/specs/podman-machine-dashboard-onboarding.spec.ts
odockal: not really. I like to start asserting from parent elements.
Ie.: Is a dialog opened? No? it does not make sense to press "Next" button and I know exactly that wizard did not appear. 
In this case the error (if by some coincidence it would happen) would just say Timeout Waiting for locator to get visible:
```
Error: Timed out 5000ms waiting for expect(locator).toBeVisible()

Locator: getByRole('navigation', { name: 'PreferencesNavigation' }).getByRole('link', { name: 'Red Hat OpenShift Local' })
```
And it is not really precise as of which of locators was faulty here. Since it chains it together. Does it make sense?

cbr7: No, I'm not following, what does that code snippet have to do with anything?

The point is that it just adds more lines of code to the test that don't have real practical purpose but on the other hand increase potential failure points. I would agree with you that checking for the parent element makes sense if we would be interacting with it in some way, but we don't, we only interact with the button in it as such it makes no real sense to check the parent element visibility.

If we were to check the visibility of every parent element it would mean we'd have to make ten asserts until we reach the button. It's just code bloat.

tests/playwright/src/model/pages/dashboard-page.ts
cbr7: I would also argue against the idea that it's more readable, to me it seems like we're defining way to many fields that serve no practical purpose not only duplicating a lot of code but actively making it more confusing to understand to somebody that is not familiar with it.

For example how is somebody on first look supposed to know the difference between "Actual state" and "Status label" fields? There's no way, because they are practically identical from an user perspective.

benoitf (approved):
no blocker on my side, up to @cbr7 to have the final word :-)

tests/playwright/src/model/pages/dashboard-page.ts
odockal: It will not work if you install any other extension adding a "Provider" box into a dashboard, since there will be two same elements defined by one locator. So we need to level down.

cbr7: How can it not? the parent element is "Podman Provider" adding another podman provider is not possible, there is only one box for podman as such it's always going to be unique. I really don't see how your hypothetical is possible.

tests/playwright/src/specs/podman-machine-dashboard-onboarding.spec.ts
odockal: OpenShift Local has the same button and the same functionality. In case that you have both or just openshift local, you would be clicking on totally different element.
since I agree with not checking the parent a locator involving getByRole('region', 'Podman Provider') must be there so we are sure are clicking right button.

odockal: the snippet is showing an error: chained locator if, for example, the Podman extension is disabled would occur when you do not see right now which part has failed. Snippet is from different scenario.

cbr7: But the initialize and start button you are waiting for is defined from the podmanProvider section, it's literally in the PR change that you made, as such even if Openshift Local was available and had the hypothetical button available PW would still not pick it up because has the different parent. 

image

The button locator simply cannot be confused with any other button on the page regardless of if it has the same name or not.

As such validating on the parent element is not necessary.

odockal: Yes. As I said, with the adding a provider distinction we could omit it.

tests/playwright/src/model/pages/dashboard-page.ts
odockal: gosh, I though you are pointing at something else. Sorry, now I got it. You are actually pointing at the next line having one necessary layer. Sure thing, let's remove it.

cbr7 (approved):

PR #6492, opened 2024-03-21
feat: import containers images

lstocchi:
### What does this PR do?

It adds the import container action.

### Screenshot / video of UI

import_containers

it is part of #6355 

### How to test this PR?

1. images page -> click import -> try to import container image

### DIFF
--- a/packages/renderer/src/lib/image/ImportContainersImages.svelte
+++ b/packages/renderer/src/lib/image/ImportContainersImages.svelte
@@ -1,15 +1,19 @@
 <script lang="ts">
+/* eslint-disable import/no-duplicates */
+// https://github.com/import-js/eslint-plugin-import/issues/1479
 import { faMinusCircle, faPlay, faPlusCircle } from '@fortawesome/free-solid-svg-icons';
 import { Input } from '@podman-desktop/ui-svelte';
-import { get, onMount } from 'svelte';
+import { onMount } from 'svelte';
+import { get } from 'svelte/store';
 import { router } from 'tinro';
 
+import Button from '/@/lib/ui//Button.svelte';
+import ErrorMessage from '/@/lib/ui//ErrorMessage.svelte';
+import FormPage from '/@/lib/ui//FormPage.svelte';
 import { providerInfos } from '/@/stores/providers';
+import { createTask } from '/@/stores/tasks';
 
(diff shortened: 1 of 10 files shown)


lstocchi:
Not sure about the taskmanager. Should i create one task per image or just a single generic import task?

benoitf:
@lstocchi maybe one generic

packages/renderer/src/lib/image/ImportContainersImages.svelte
```diff
@@ -0,0 +1,166 @@
...
+import { Input } from '@podman-desktop/ui-svelte';
+import { get, onMount } from 'svelte';
+import { router } from 'tinro';
+
+import { providerInfos } from '/@/stores/providers';
+
+import type { ProviderContainerConnectionInfo, ProviderInfo } from '../../../../main/src/plugin/api/provider-info';
+import Button from '../ui/Button.svelte';
+import ErrorMessage from '../ui/ErrorMessage.svelte';
+import FormPage from '../ui/FormPage.svelte';
```
benoitf: prefer `/@/lib/` imports

packages/renderer/src/lib/image/ImportContainersImages.svelte
```diff
@@ -0,0 +1,166 @@
...
+    .filter(providerContainerConnection => providerContainerConnection.status === 'started');
+
+  const selectedProviderConnection = providerConnections.length > 0 ? providerConnections[0] : undefined;
+  selectedProvider = !selectedProvider && selectedProviderConnection ? selectedProviderConnection : selectedProvider;
+
+  const platform = await window.getOsPlatform();
+  if (platform === 'win32') {
+    separator = '\\';
+  }
+});
```
benoitf: I'm wondering if we need to handle path on the client side

shouldn't it be part of the job of the backend ?

also by receiving Uri it should be always `/`

lstocchi: the `openDialog` still returns a string. I just updated the saveDialog.
The separator is just needed to calculate an initial importing name, which could also be not completely correct. It is just to generate something to start with for the user but they can update it.
E.g. I just take the content starting from the last slash to the last dot. So let's say the path is path/file.tar.gz, the resulting importing image name would be `file.tar` while it would better be `file`.
The alternative is to create a function generateFileName which just extract the filename from a path, but i wonder if it's needed as it's just to show something

benoitf (approved):

PR #6499, opened 2024-03-22
fix: support more fa icons

### LINKED ISSUE #6498: bug(Button): component does not show icon for certain fa icons
### Bug description

Icons such as `faGithub` has a prefix `fab`. However the check in the Button component does not support it

https://github.com/containers/podman-desktop/blob/48c0d3905f6acf60a1f4fb659610787f262a6c77/packages/renderer/src/lib/ui/Button.svelte#L25

### Operating system

windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

```svelte
    <Button icon="{faGithub}">Github</Button>
```

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Adding the prefix of unsupported fa icons to properly display them

Fixes https://github.com/containers/podman-desktop/issues/6498

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/Button.svelte
+++ b/packages/renderer/src/lib/ui/Button.svelte
@@ -4,6 +4,7 @@ import Fa from 'svelte-fa';
 
 import type { ButtonType } from './Button';
 import Spinner from './Spinner.svelte';
+import { isIconDefinition } from '/src/lib/ui/IconUtils';
 
 export let title: string | undefined = undefined;
 export let inProgress = false;
@@ -22,7 +23,7 @@ export let padding: string =
 let iconType: string | undefined = undefined;
 
 onMount(() => {
-  if (['fas', 'fab'].includes(icon?.prefix)) {
+  if (isIconDefinition(icon)) {
     iconType = 'fa';
   } else {
     iconType = 'unknown';
--- a/packages/renderer/src/lib/ui/Button.svelte
+++ b/packages/renderer/src/lib/ui/Button.svelte
@@ -22,7 +22,7 @@ export let padding: string =
 let iconType: string | undefined = undefined;
 
 onMount(() => {
-  if (icon?.prefix === 'fas') {
+  if (['fas', 'fab'].includes(icon?.prefix)) {
     iconType = 'fa';
   } else {
     iconType = 'unknown';
--- a/packages/renderer/src/lib/ui/Button.svelte
+++ b/packages/renderer/src/lib/ui/Button.svelte
@@ -2,9 +2,10 @@
 import { onMount } from 'svelte';
 import Fa from 'svelte-fa';
 
+import { isFontAwesomeIcon } from '/@/lib/ui/icon-utils';
+
 import type { ButtonType } from './Button';
 import Spinner from './Spinner.svelte';
-import { isFontAwesomeIcon } from '/@/lib/ui/icon-utils';
 
 export let title: string | undefined = undefined;
 export let inProgress = false;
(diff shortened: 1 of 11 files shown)


packages/renderer/src/lib/ui/Button.svelte
```diff
@@ -21,7 +21,7 @@ export let padding: string =
 let iconType: string | undefined = undefined;
 
 onMount(() => {
-  if (icon?.prefix === 'fas') {
+  if (['fas', 'fab'].includes(icon?.prefix)) {
```
benoitf: I think we have some code with 'fa' as prefix so maybe would be simpler

if you want to cover all icons then you'll be missing regular (`far` and maybe more)
https://github.com/FortAwesome/Font-Awesome/blob/6.x/js-packages/%40fortawesome/free-regular-svg-icons/faAddressCard.js#L3

axel7083: So what you recommend is doing like `icon?.prefix.startsWith('fa')` ?

benoitf: yes

benoitf: or maybe here we should have an unified utility class that really do checks all the available font awesome options

b/c I think each component is doing its own check

deboer-tim: Looks like we're doing this in at least 4 places, agreed a util function would be useful.

axel7083: I created a  `IconUtils` file to handle the logic, and changed the 4 places where the same logic was applied

packages/renderer/src/lib/ui/IconUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { IconDefinition } from '@fortawesome/fontawesome-common-types';
+
+export const isIconDefinition = (icon: unknown): icon is IconDefinition => {
```
benoitf: if you could rename to include FontAwesome maybe

isFontAwesomeIconDefinition

packages/renderer/src/lib/ui/IconUtils.ts
```diff
@@ -0,0 +1,29 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/renderer/src/lib/ui/IconUtils.ts
```diff
@@ -0,0 +1,29 @@
+/**********************************************************************
```
benoitf: The file should be all lowercase for utils

packages/renderer/src/lib/ui/IconUtils.spec.ts
```diff
@@ -0,0 +1,30 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

benoitf (approved):
looks like we're missing the update in ListItemButtonIcon.svelte

benoitf:
nice enhancement 👍

PR #6512, opened 2024-03-24
fix: ensure machine list is up-to-date when running onboarding

### LINKED ISSUE #6511: podman desktop is prompting to create a machine while there is one
### Bug description

note: it can only be reproduced with a 5s delay

when starting the onboarding, we clear all onboarding properties. Then until the `podman machine list` loop occured, during 5s we might have the property missing then it'll ask to create the podman machine

### Operating system

macOS 14.4

### Installation Method

None

### Version

1.8.0

### Steps to reproduce

need to be fast but during 0-5s we might have invalid values

I think that we should split onboarding properties with the enablement properties as well

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Ensure that we have up-to-date information about the number of podman machines.
also what happens is that when restarting an onboarding workflow, all values are removed, so it might say that there is no machine in the workflow and display the wizard to create a machine while we may have one.

Here we enforce that a command is executed before checking the result which should be the 'recommended workflow' when checking values.

### Screenshot / video of UI

N/A

fixes https://github.com/containers/podman-desktop/issues/6511

### How to test this PR?

really hard to reproduce manually at it's only within a 5s delay window after restarting onboarding workflow

### DIFF
--- a/extensions/podman/package.json
+++ b/extensions/podman/package.json
@@ -26,6 +26,10 @@
         "title": "Podman: Check system requirements to install podman"
       },
       {
+        "command": "podman.onboarding.checkPodmanMachineExistsCommand",
+        "title": "Podman: Check if a podman machine exists"
+      },
+      {
         "command": "podman.onboarding.installPodman",
         "title": "Podman: Install podman"
       }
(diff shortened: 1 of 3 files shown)


deboer-tim (approved):
LGTM

lstocchi (approved):
LGTM

Issue #478, opened 2022-09-16
Load & Save Image


trymeouteh:
### Is your feature request related to a problem? Please describe

The ability to load an image tar file into podman and be able to export an image from Podman Desktop and save it as a image tar file.

### Describe the solution you'd like

A Load Image button to load images into Podman Desktop next to the pull image button under the Images tab.

A Save Image button to save images from Podman Desktop next to the pull image button under the Images tab.

elervik:
Ill stab at this.

benoitf:
@elervik I've assigned the issue

elervik:
I have a working load & save for podman, is there demand to extend supporting export and import also?

Initially I wanted to give the ability to select engine (ie docker), but the dockerode API load and save endpoints are misbehaving, load is even 404.. Maybe its just me? :) Is multi-engine a must have for this initial feature extension, or should I commit what is and improve later?

benoitf:
Hello, multi-engine is a pre-requisite but it doesn't prevent to create the PR

elervik:
I dont have time to contribute to this issue at the moment, I removed my assignment in case someone else looks at it before I can get down to business again.

benoitf:
there should have a way to do that also from a 3rd party extension

PR #6520, opened 2024-03-25
feat: add saveImages logic

lstocchi:
### What does this PR do?

it adds the logic to save the images. It will be used by the UI in following PRs

### Screenshot / video of UI

N/A

it is part of #478 

### How to test this PR?

run tests

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -19,6 +19,7 @@
 import * as crypto from 'node:crypto';
 import { EventEmitter } from 'node:events';
 import * as fs from 'node:fs';
+import path from 'node:path';
 import { type Stream, Writable } from 'node:stream';
 import { pipeline } from 'node:stream/promises';
 
@@ -38,6 +39,7 @@ import type {
   ContainerExportOptions,
   ContainerInfo,
   ContainerPortInfo,
+  ImagesSaveOptions,
   NetworkCreateOptions,
   NetworkCreateResult,
   SimpleContainerInfo,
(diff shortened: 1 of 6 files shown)


packages/main/src/plugin/container-registry.ts
```diff
@@ -2426,4 +2428,81 @@ export class ContainerProviderRegistry {
...
+    let errors = '';
+
+    for (const imageGroup of mapImages.entries()) {
+      const engine = this.internalProviders.get(imageGroup[0]);
+      if (!engine?.api) {
+        errors += `Unable to save images ${imageGroup[1].join(',')}. Error: No running provider for the matching images\n`;
+        continue;
+      }
+      if ('getImages' in engine.api) {
+        engine.api.getImages = function (
```
benoitf: here it looks like we'll override the function for each call to saveImages 

maybe move it to enhancePrototypeWithLibPod so it's only done once ?

lstocchi: i extracted it to its own function. I can also move it to the contructor. Wouldn't it be better instead of updating enhancePrototypeWithLibPod as it's not related to libPod?

lstocchi: ah nevermind, not possible as i don't have a ref of the various api there

lstocchi: if i move it to libpod wouldn't it be restricted to podman providers? Don't we want it for docker as well?

benoitf: so as we're touching the prototype of Dockerode in fact it's modifying for all dockerode stuff (podman API or Docker API)

In fact Podman API is Dockerode + extra methods but it's in the same object

benoitf:
else you can move your enhancement call after the init call
```
    internalProvider.api = new Dockerode({ socketPath: containerProviderConnection.endpoint.socketPath });
```

packages/main/src/plugin/container-registry.ts
lstocchi: Done 👍

packages/main/src/plugin/container-registry.ts
```diff
@@ -2426,4 +2428,45 @@ export class ContainerProviderRegistry {
...
+        });
+
+        try {
+          let targetPath = options.outputTarget;
+          if (isMultiProvider) {
+            targetPath = path.join(options.outputTarget, `${imageGroup[0]}-images.tar`);
+          }
+          await pipeline(imagesStream, fs.createWriteStream(targetPath));
+        } catch (e) {
+          console.log(String(e));
```
```suggestion
```

benoitf (approved):

PR #6521, opened 2024-03-25
chore(tests): add optional param to constructor

cbr7:
### What does this PR do?
Add optional param to Desktop Runner constructor

https://github.com/containers/podman-desktop/issues/6519

### DIFF
--- a/tests/playwright/src/runner/podman-desktop-runner.ts
+++ b/tests/playwright/src/runner/podman-desktop-runner.ts
@@ -38,14 +38,16 @@ export class PodmanDesktopRunner {
   private readonly _customFolder;
   private readonly _testOutput: string;
   private _videoAndTraceName: string | undefined;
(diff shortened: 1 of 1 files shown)


odockal (approved):

odockal:
LGTM but we need to refactor this to accept custom settings in the runner as there could be more preferences set, which is not viable to have it done through constructor params but rather settings file itself.

benoitf (approved):
+1 to @odockal you should probably accept a JSON like object else it won't scale and to avoid to change all calls for each new parameter

cbr7:
@odockal @benoitf created https://github.com/containers/podman-desktop/issues/6524 as a follow up issue to crreate a more general solution

odockal:
@benoitf @cbr7 Yes. That is totally doable. I am also thinking about getting an option to use existing settings.json and merging it with a default/passed configuration.

PR #6530, opened 2024-03-25
feat: allow to save images

lstocchi:
### What does this PR do?

this allows to save images.
there are 3 possible scenarios.
1) only one item selected (or user clicks on the action on the right menu) -> it saves an archive containing one item, the user can customize the archive name
2) multiple items selected belonging to the same provider -> it saves one archive, the user can customize the archive name
3) multiple items belonging to different providers -> it saves N archives where N is the number of different providers, the user only select the final directory, then archives with name {providerName}-images-{randomstring}.tar are created.

### Screenshot / video of UI

1) save single item

save_single_item

2) save multiple items of the same provider

save_multiple_item

3) save multiple items from multiple providers

save_multiple_item_multiple_provs

it is part of #478 

### How to test this PR?

1) select images and save them, the load action will be available soon

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -143,7 +143,7 @@ window.events?.receive('navigate', (navigationRequest: unknown) => {
         <Route path="/images/import" breadcrumb="Import Containers">
           <ImportContainersImages />
         </Route>
-        <Route path="/images/save/*" breadcrumb="Save Images">
+        <Route path="/images/save" breadcrumb="Save Images">
           <SaveImages />
         </Route>
         <Route path="/pods" breadcrumb="Pods" navigationHint="root">
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -30,6 +30,7 @@ import ImagesList from './lib/image/ImagesList.svelte';
 import ImportContainersImages from './lib/image/ImportContainersImages.svelte';
 import PullImage from './lib/image/PullImage.svelte';
 import RunImage from './lib/image/RunImage.svelte';
+import SaveImages from './lib/image/SaveImages.svelte';
 import IngressDetails from './lib/ingresses-routes/IngressDetails.svelte';
 import IngressesRoutesList from './lib/ingresses-routes/IngressesRoutesList.svelte';
 import RouteDetails from './lib/ingresses-routes/RouteDetails.svelte';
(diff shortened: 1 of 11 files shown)


packages/renderer/src/lib/image/SaveImages.svelte
```diff
@@ -0,0 +1,187 @@
+<script lang="ts">
+/* eslint-disable import/no-duplicates */
+// https://github.com/import-js/eslint-plugin-import/issues/1479
```
benoitf: do we need these two ? b/c I don't see duplicates

benoitf (approved):

packages/renderer/src/lib/image/SaveImages.svelte
```diff
@@ -0,0 +1,185 @@
...
+  if (singleItemMode) {
+    let lastSlashPos = imagesToSave[0].name.lastIndexOf('/') + 1;
+    let lastColon: number | undefined = imagesToSave[0].name.lastIndexOf(':');
+    if (lastColon === -1 || lastColon < lastSlashPos) {
+      lastColon = undefined;
+    }
+    targetFile = `${imagesToSave[0].name.substring(lastSlashPos, lastColon)}.tar`;
+  }
+  const result = await window.saveDialog({
+    title: 'Select the directory where to export the container content',
```
deboer-tim: Nit: remove 'where'.

packages/renderer/src/lib/image/SaveImages.svelte
```diff
@@ -0,0 +1,185 @@
...
+    return;
+  }
+  const uriAPI = Uri.revive(result);
+  outputPath = uriAPI.fsPath;
+  invalidOutputPath = false;
+}
+
+async function selectOutputDirectoryPath() {
+  const result = await window.openDialog({
+    title: `Select the directory where to save the ${singleItemMode ? 'image' : 'images'}`,
```
deboer-tim: Nit: remove 'where'.

packages/renderer/src/App.svelte
```diff
@@ -142,6 +143,9 @@ window.events?.receive('navigate', (navigationRequest: unknown) => {
         <Route path="/images/import" breadcrumb="Import Containers">
           <ImportContainersImages />
         </Route>
+        <Route path="/images/save/*" breadcrumb="Save Images">
```
deboer-tim: Doesn't look like it needs the /* here.

deboer-tim (approved):
LGTM

benoitf (approved):

PR #6539, opened 2024-03-26
feat: ask to wipe all data when migrating from podman v4 to v5

### LINKED ISSUE #6522: In case of podman v4 -> v5 update, ask the user to wipe out all data

as suggested in https://blog.podman.io/2024/03/migration-of-podman-4-to-podman-5-machines/ as some data are not backward compliant, it might be a good thing to start from scratch after the migration

Prompt to wipe all data before upgrading from v4 to v5

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Also bring a confirmation dialog
it is still an optional step

fixes https://github.com/containers/podman-desktop/issues/6522

### How to test this PR?

1. install v4.9.3
2. enable the experimental v5 upgrade flag
3. click on 'update to 5.0.0

then you should be prompted to wipe all data and confirm it

also unit tests are provided

### DIFF
--- a/extensions/podman/src/podman-install.ts
+++ b/extensions/podman/src/podman-install.ts
@@ -36,6 +36,8 @@ import {
   USER_MODE_NETWORKING_SUPPORTED_KEY,
 } from './extension';
 import { MacCPUCheck, MacMemoryCheck, MacPodmanInstallCheck, MacVersionCheck } from './macos-checks';
+import { PodmanCleanupMacOS } from './podman-cleanup-macos';
+import { PodmanCleanupWindows } from './podman-cleanup-windows';
 import type { InstalledPodman } from './podman-cli';
 import { getPodmanCli, getPodmanInstallation } from './podman-cli';
 import * as podman4JSON from './podman4.json';
@@ -127,7 +129,7 @@ interface Installer {
   requireUpdate(installedVersion: string): boolean;
   update(): Promise<boolean>;
 }
-interface UpdateCheck {
+export interface UpdateCheck {
   hasUpdate: boolean;
   installedVersion: string;
   bundledVersion: string;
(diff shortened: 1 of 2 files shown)


extensions/podman/src/podman-install.ts
```diff
@@ -240,6 +249,59 @@ export class PodmanInstall {
...
+  ): Promise<boolean> {
+    // if (v4 --> v5)
+    if (
+      installedPodman.version.startsWith('4.') &&
+      updateInfo.bundledVersion.startsWith('5.') &&
+      this.providerCleanup
+    ) {
+      // prompt if user wants to wipe all data
+      const answer = await extensionApi.window.showInformationMessage(
+        `You are updating from Podman ${installedPodman.version} to ${updateInfo.bundledVersion}. It is recommended to delete all data (including containers, volumes, networks, podman machines, etc) for this update. DATA WILL BE DESTROYED. Do you want to proceed ?`,
```
lstocchi, line 266: DATA WILL BE DELETED PERMANENTLY

lstocchi: I would wait for the suggestion of a native speaker as i'm not sure destroyed is the verb to use, even if it makes sense to me. Maybe `DATA WILL BE DELETED PERMANENTLY` is more formal and polite?

extensions/podman/src/podman-install.ts
```diff
@@ -249,6 +311,17 @@ export class PodmanInstall {
         await extensionApi.window.showWarningMessage('Podman update has been canceled.', 'OK');
         return;
       }
+
+      // podman v4 -> v5 migration: ask to wipe all data before doing the update
+      const wipeAllDataCompleted = await this.wipeAllDataBeforeUpdatingToV5(installedPodman, updateInfo);
+      if (!wipeAllDataCompleted) {
+        await extensionApi.window.showWarningMessage(
+          'Podman update has been canceled. You can do some backups before restarting the update',
```
lstocchi: Would it be preferred to use a "it is recommended" instead of "you can do"? Something like `It is recommended to perform backups before resuming the update procedure.`

lstocchi (approved):
LGTM, tested on win and works fine.
Just a doubt about the messages from a non-native speaker

axel7083 (approved):
Tested locally I properly have the dialogs. LGTM

extensions/podman/src/podman-install.ts
benoitf: fixed

extensions/podman/src/podman-install.ts
benoitf: fixed

PR #6546, opened 2024-03-27
feat: support for webview reveal

### LINKED ISSUE #5925: Ensure the reveal option works on webviews

For now webviews are created and we expect the user click on the left navbar before displaying it.

but webview could be created by an extension as part of a given workflow and the extension might want to display the webview after its creation (instead of waiting for the user clicking on the webview in the navigation area)

Check we can programmatically reveal a webview

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

When you call a Webview panel's reveal() function today, it fires a webview-panel-reveal event that nothing listens to, so the panel is never opened. There are a few possible ways we could solve this:
- Webview panel could call the navigation manager directly.
- Webview panel could fire a navigation event itself.
- Renderer can listen to the reveal event and navigate.

The first requires the navigation manager to be passed into the Webview, which involves a decent amount of change and a somewhat circular dependency because the navigation manager already depends on the webview registry. The third is really simple and works, but this is essentially a second, custom webview navigation event.

So this is the middle option: just trigger a webview navigation event instead. Although it's not necessary the code uses NavigationPage and casts to NavigationRequest to ensure type safety, and being caught if there are ever changes to the navigation code.

The optional preserveFocus is (still) not respected, since webview navigation doesn't support an option like this; we could add that separately, whether or not we expose it via the navigation API.

### Screenshot / video of UI

N/A

Fixes #5925.

### How to test this PR?

Test updated; you need a webview to call panel.reveal() on itself to test in dev.

### DIFF
--- a/packages/main/src/plugin/webview/webview-panel-impl.ts
+++ b/packages/main/src/plugin/webview/webview-panel-impl.ts
@@ -27,6 +27,8 @@ import type {
 import type { ApiSenderType } from '/@/plugin/api.js';
 import { Emitter } from '/@/plugin/events/emitter.js';
 
+import { NavigationPage } from '../navigation/navigation-page.js';
+import type { NavigationRequest } from '../navigation/navigation-request.js';
 import type { WebviewImpl } from './webview-impl.js';
 import type { WebviewRegistry } from './webview-registry.js';
 
@@ -146,11 +148,16 @@ export class WebviewPanelImpl implements WebviewPanel {
     }
   }
 
-  reveal(preserveFocus?: boolean): void {
+  reveal(_preserveFocus?: boolean): void {
     this.assertNotDisposed();
 
     // notify the renderer to reveal the webview
-    this.#apiSender.send('webview-panel-reveal', { id: this.#internalId, preserveFocus });
+    this.#apiSender.send('navigate', {
+      page: NavigationPage.WEBVIEW,
+      parameters: {
+        id: this.#internalId,
+      },
+    } as NavigationRequest);
   }
 
   dispose(): void {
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/webview/webview-panel-impl.ts
```diff
@@ -146,11 +148,16 @@ export class WebviewPanelImpl implements WebviewPanel {
...
     this.assertNotDisposed();
 
     // notify the renderer to reveal the webview
-    this.#apiSender.send('webview-panel-reveal', { id: this.#internalId, preserveFocus });
+    this.#apiSender.send('navigate', {
+      page: NavigationPage.WEBVIEW,
+      parameters: {
+        id: this.#internalId,
+      },
+    } as NavigationRequest);
```
benoitf: not sure why we're adding `as NavigationRequest` ?

benoitf: I would opt for an assignment like

```ts
const navigationRequest: NavigationRequest =
{
  page: NavigationPage.WEBVIEW,
  parameters: {
    id: this.#internalId,
  },
};

this.#apiSender.send('navigate', navigationRequest);
```

axel7083 (approved):
Work like a charm! LGTM

https://github.com/containers/podman-desktop/assets/42176370/c7a17d17-355c-4436-adc5-9e20afa1115e

packages/main/src/plugin/webview/webview-panel-impl.ts
deboer-tim: Refactored as suggested, agree it's better than `as NavigationRequest`.

benoitf (approved):

PR #6555, opened 2024-03-28
Type clean-up on Dashboard

### LINKED ISSUE #6547: Consistent fonts on Dashboard screen

At the moment, the text on the dashboard screen is not following a consistent set of rules. The visual hierarchy needs to be stronger. 

This relates to https://github.com/containers/podman-desktop/issues/6405 

Go through the page and set the font to the appropriate size, setting them up to be edited later.

_No response_

### Additional context

From this we can go through the other pages and make them follow this format also.

ekidneyrh:
### What does this PR do?

Sized the text to improve the visual hierarchy of information on the Dashboard page. Also gets things in order to define type sizes in the tailwind config file in the future. 

### Screenshot / video of UI
This is what it looks like now:
Screenshot from 2024-03-28 11-14-41

And this is with the changes:
image

Would close https://github.com/containers/podman-desktop/issues/6547

Changes are only slight until changes are made to the type styles. This is done to avoid disruption to the ui of the rest of the app.

### DIFF
--- a/packages/renderer/src/lib/carousel/CarouselTest.svelte
+++ b/packages/renderer/src/lib/carousel/CarouselTest.svelte
@@ -3,7 +3,7 @@ import Carousel from './Carousel.svelte';
 </script>
 
 <div>
-  <p class="text-lg first-letter:uppercase font-bold pb-5">Learning center:</p>
+  <p class="text-lg first-letter:uppercase pb-5">Learning center:</p>
   <div class="bg-charcoal-800 p-4 rounded-lg">
     <Carousel cards="{['card 1', 'card 2', 'card 3']}" cardWidth="{340}" let:card>
       <p>{card}</p>
--- a/packages/renderer/src/lib/dashboard/NotificationsBox.svelte
+++ b/packages/renderer/src/lib/dashboard/NotificationsBox.svelte
@@ -27,7 +27,7 @@ onDestroy(() => {
 {#if notifications.length > 0}
   <div class="bg-charcoal-800 m-5 px-5 py-4 rounded-lg">
     <div class="flex flex-col items-center justify-content space-y-3" role="region" aria-label="Notifications Box">
-      <span class="text-md text-white mb-1">Notifications</span>
+      <span class="text-base text-white mb-1">Notifications</span>
       {#each notifications as notification}
         <NotificationCardItem notification="{notification}" />
       {/each}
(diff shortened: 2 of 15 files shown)


slemeur:
This is very good work @ekidneyrh !

deboer-tim:
These font sizes appear to be the same as the tailwind default so I'm not concerned about that, but:
- We use text-xs in ~10 places, I think this would revert those to the default size since they wouldn't be found? These need to be fixed or text-xs added.
- The defaults have line-heights, and no fontWeights. Changing the weight globally likely doesn't have a big impact, but changing the height could have one. When we removed PatternFly it changed the line height and caused _lots_ of minor regressions, so we should check lots of screens here.
- Tests would have to be fixed.

ekidneyrh:
> We use text-xs in ~10 places, I think this would revert those to the default size since they wouldn't be found? These need to be fixed or text-xs added.

I can amend the config file and add in the xs size 

> The defaults have line-heights, and no fontWeights. Changing the weight globally likely doesn't have a big impact, but changing the height could have one. When we removed PatternFly it changed the line height and caused _lots_ of minor regressions, so we should check lots of screens here.
> 
Do you mean that the line heights have been changed here, or are you pointing out that they can be defined also? 

I was focusing just on the Dashboard for the time being, but I know this has effects throughout. What way do you think would be best to do this? My original plan was to go screen by screen, but if it has major impacts throughout, then could we do everything in one PR?

benoitf:
>, then could we do everything in one PR?

I would opt for small PRs

(like doing it screen by screen or by component)

like : only buttons, only dashboard, etc

it's easier to see regressions/revert when it's done step by step

deboer-tim:
> Do you mean that the line heights have been changed here, or are you pointing out that they can be defined also?

It looks like the Tailwind defaults have line heights, so not defining them in our own would presumably remove them. It might not be the case or might not be an issue, it's just that this is a 'global' change.

ekidneyrh:
Do you think it would be better if I define the type sizes last? That way it doesn't mess up the other screens in the meantime.

benoitf (commented):
AFAIK `text-bt` does not exist in tailwind but I see several usage in this PR

ekidneyrh:
I have text-bt used for button text, which will be defined in the tailwind config file once the type on the rest of the pages are cleaned-up like this.  It seems to just be defaulting to 16px atm.

PR #6565, opened 2024-03-28
feat: detect podman v4 qemu machines after update and delete them

### LINKED ISSUE #6559: Detect previous v4 qemu machines and add prompt to delete them

After people upgraded to Podman v5, they might have old v4 machines that should be removed as unsupported

Detect that and enhance the onboarding podman workflow to handle that

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
It is part of the onboarding workflow

Podman Desktop detects that problem and register a notification
Then you can run the onboarding workflow and fix the issue

fixes https://github.com/containers/podman-desktop/issues/6559

### How to test this PR?

Install podman v4 and create machines

then install podman v5 using the installer (and not using Podman Desktop to do the install)

Start Podman Desktop. It should add a notification about Podman onboarding and you'll be able to remove qemu machines from there

note: It's for macOS only

### DIFF
--- a/extensions/podman/package.json
+++ b/extensions/podman/package.json
@@ -306,7 +306,7 @@
             ],
             [
               {
-                "value": "Interupt the process by clicking on `skip` or `skip the entire setup`"
+                "value": "Interrupt the process by clicking on `skip` or `skip the entire setup`"
               }
             ]
           ],
--- a/extensions/podman/package.json
+++ b/extensions/podman/package.json
@@ -32,6 +32,14 @@
       {
         "command": "podman.onboarding.installPodman",
         "title": "Podman: Install podman"
+      },
+      {
+        "command": "podman.onboarding.checkUnsupportedPodmanMachine",
+        "title": "Podman: Check if an old unsupported podman machine exists"
+      },
+      {
+        "command": "podman.onboarding.removeUnsupportedMachines",
+        "title": "Podman: Remove unsupported podman machines"
       }
     ],
     "configuration": {
@@ -260,7 +268,7 @@
         {
           "id": "installationSuccessView",
           "title": "Podman settings",
-          "when": "!onboardingContext:podmanIsNotInstalled",
+          "when": "!onboardingContext:podmanIsNotInstalled && !podman.needPodmanMachineCleanup",
           "content": [
             [
               {
@@ -268,6 +276,41 @@
               }
             ]
           ]
+        },
+        {
+          "id": "checkUnsupportedPodmanMachineCommand",
+          "title": "Checking if an old unsupported Podman machine exists",
+          "when": "!isLinux",
+          "command": "podman.onboarding.checkUnsupportedPodmanMachine",
+          "completionEvents": [
+            "onCommand:podman.onboarding.checkUnsupportedPodmanMachine"
+          ]
+        },
+        {
+          "id": "welcomeFixUnsupportedPodmanMachineView",
+          "title": "Unsupported podman machines detected",
+          "description": "Podman Desktop will remove the unsupported machines",
+          "completionEvents": [
+            "onboardingContext:unsupportedMachineRemoved == ok"
+          ],
+          "content": [
+            [
+              {
+                "value": "Machines created with podman v4 are no longer supported by podman v5"
+              }
+            ],
+            [
+              {
+                "value": ":button[Remove unsupported podman machines]{command=podman.onboarding.removeUnsupportedMachines}"
+              }
+            ],
+            [
+              {
+                "value": "Interupt the process by clicking on `skip` or `skip the entire setup`"
+              }
+            ]
+          ],
+          "when": "onboardingContext:unsupportedPodmanMachine && !isLinux"
         },
         {
           "id": "checkPodmanMachineExistsCommand",
@@ -309,7 +352,7 @@
           ]
         }
       ],
-      "enablement": "(isLinux && onboardingContext:podmanIsNotInstalled) || (!isLinux && !onboardingContext:podmanMachineExists)"
+      "enablement": "(isLinux && onboardingContext:podmanIsNotInstalled) || (!isLinux && !onboardingContext:podmanMachineExists) || podman.needPodmanMachineCleanup"
     }
   },
   "scripts": {
--- a/extensions/podman/package.json
+++ b/extensions/podman/package.json
@@ -288,7 +288,7 @@
         },
         {
           "id": "welcomeFixUnsupportedPodmanMachineView",
-          "title": "Unsupported podman machines detected",
+          "title": "Unsupported Podman machine detected",
           "description": "Podman Desktop will remove the unsupported machines",
           "completionEvents": [
             "onboardingContext:unsupportedMachineRemoved == ok"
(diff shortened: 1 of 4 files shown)


extensions/podman/src/extension.ts
```diff
@@ -884,6 +903,58 @@ export function registerOnboardingMachineExistsCommand(): extensionApi.Disposabl
...
+
+      // remove folders if exists
+      const foldersToRemove = [];
+      if (fs.existsSync(qemuSharePath)) {
+        foldersToRemove.push(qemuSharePath);
+      }
+      if (fs.existsSync(qemuConfigPath)) {
+        foldersToRemove.push(qemuConfigPath);
+      }
+
```
deboer-tim, line 944: A little redundant and maybe paranoid, but maybe an "if (foldersToRemove.isEmpty) return" here? e.g. maybe there is a delay and user has deleted them since the check.

extensions/podman/package.json
```diff
@@ -269,6 +277,41 @@
...
+          "title": "Checking if an old unsupported Podman machine exists",
+          "when": "!isLinux",
+          "command": "podman.onboarding.checkUnsupportedPodmanMachine",
+          "completionEvents": [
+            "onCommand:podman.onboarding.checkUnsupportedPodmanMachine"
+          ]
+        },
+        {
+          "id": "welcomeFixUnsupportedPodmanMachineView",
+          "title": "Unsupported podman machines detected",
```
deboer-tim: Capitalization of Podman is a bit inconsistent in lines 38/42/282/291/299. I think referring to the command should be lowercase like `podman machine`, but when referring to a machine in normal writing Podman is a name that should be capitalized, e.g. "unsupported Podman machine".

extensions/podman/src/extension.ts
```diff
@@ -884,6 +903,58 @@ export function registerOnboardingMachineExistsCommand(): extensionApi.Disposabl
...
+      // prompt the user to confirm
+      const result = await extensionApi.window.showWarningMessage(
+        'Confirm removal of the unsupported qemu podman machines',
```
deboer-tim: We don't say that these are qemu prior to this, so I think we should just leave it out, otherwise it could be confusing or distracting to some users.

Even though they're moving to 5.0 and won't have access to these machines, they could still have work/images in them. Maybe we should be more explicit too? e.g. "Removing old unsupported Podman machines will delete all of their data. Confirm approval?"

deboer-tim (approved):
I asked for some minor changes, but marking as approved since functionally this looks good and I may not be online to approve later.

extensions/podman/src/extension.ts
benoitf: so the check here is done during the onboarding flow, I mean, it's called if user click on a button so I think there is no time to click and run a command elsewhere

extensions/podman/package.json
benoitf: will do changes

extensions/podman/src/extension.ts
benoitf: will change 👍

extensions/podman/src/extension.ts
deboer-tim: ok 👍🏼

extensions/podman/src/extension.ts
```diff
@@ -550,6 +550,20 @@ async function monitorMachines(provider: extensionApi.Provider): Promise<void> {
   }
 }
 
+async function shouldNotifyQemuMachinesWithV5(installedPodman: InstalledPodman): Promise<boolean> {
```
lstocchi: Why async?

extensions/podman/src/extension.ts
```diff
@@ -884,6 +903,58 @@ export function registerOnboardingMachineExistsCommand(): extensionApi.Disposabl
...
+      );
+      if (result === 'No') {
+        return;
+      }
+
+      for (const folder of foldersToRemove) {
+        try {
+          await fs.promises.rm(folder, { recursive: true, retryDelay: 1000, maxRetries: 3 });
+        } catch (error) {
+          console.error('Error removing folder', folder, error);
```
lstocchi, line 960: Should we show an error if the removal fails? I guess the user would see the onboarding again next time if the removal fails as desktop detects qemu and the user will be puzzled

extensions/podman/package.json
```diff
@@ -269,6 +277,41 @@
             ]
           ]
         },
+        {
+          "id": "checkUnsupportedPodmanMachineCommand",
+          "title": "Checking if an old unsupported Podman machine exists",
+          "when": "!isLinux",
```
lstocchi, line 283: Should this be isMac?

lstocchi (commented):
Code LGTM, i'll try it on my mac

extensions/podman/src/extension.ts
benoitf: I think I started it differently but then did not revert to non-async, I will do the changes

extensions/podman/package.json
benoitf: so I could but there is another issue where in this checkUnspportedPodmanMachineCommand where we will check v4 --> v5 machines like applehv --> applehv or on Windows so I'll just enhance the method

extensions/podman/src/extension.ts
benoitf: good point, I'll notify the user

extensions/podman/package.json
```diff
@@ -269,6 +277,41 @@
...
+              }
+            ],
+            [
+              {
+                "value": ":button[Remove unsupported podman machines]{command=podman.onboarding.removeUnsupportedMachines}"
+              }
+            ],
+            [
+              {
+                "value": "Interupt the process by clicking on `skip` or `skip the entire setup`"
```
lstocchi: Interrupt

lstocchi (approved):
Works fine for me 🚀 I just noticed a typo

extensions/podman/package.json
benoitf: fixed

extensions/podman/src/extension.ts
benoitf: fixed

extensions/podman/src/extension.ts
benoitf: fixed

extensions/podman/src/util.ts
```diff
@@ -38,6 +38,11 @@ export function appHomeDir(): string {
   return xdgDataDirectory + '/podman';
 }
 
+const configDirectory = '.config/containers';
+export function appConfigDir(): string {
+  return `${configDirectory}/podman`;
```
axel7083, line 43: You only seems to be calling this method inside isMac condition, but this is a valid directory for other platform, maybe use path.join ?

benoitf: I mimic previous method

so here it returns a fragment of a directory and not a full path

It means that you need to use with path methods to resolve it (like path.resolve)

And we can use slashes in `path.resolve` method

benoitf: so it's valid even for Windows (like the previous method `appHomeDir` is

axel7083: Thanks for the explanation !

benoitf:
disabling auto-merge as @axel7083 didn't approve and had concerns

axel7083 (approved):
LGTM code wise (cannot test without a mac)

PR #6570, opened 2024-03-29
feat: detect podman v4 machines not compliant with the new format of v5

### LINKED ISSUE #6566: detect podman v4 machines not compliant with the new format of podman v5

some errors are displayed when using `podman machine list` about unmarshalling, etc

allow to clean these previous machines

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Detect invalid applehv v4 on Podman v5 or wsl v4 machines on v5 and ask the user to remove these machines.

fixes https://github.com/containers/podman-desktop/issues/6566

### How to test this PR?

keep Podman Desktop stopped
on macOS:
- install v4.9 edit `~/.config/containers/containers.conf` and add
```toml
[machine]
  provider = "applehv"
```
then create a podman machine using `podman machine init`

then download and install v5.0.0 installer from https://github.com/containers/podman/releases/tag/v5.0.0
(here you can create a new v5 machine like `podman machine init v5`

and start Podman Desktop. 

it should prompt you to delete v4 machines (so only `podman-machine-default` and not `v5`

on Windows:
assuming you have a fresh env (no v4 or v5 machines)
keep Podman Desktop stopped
install v4.9 and then create a podman machine using `podman machine init`
then download and install v5.0.0 installer from https://github.com/containers/podman/releases/tag/v5.0.0
(here you can create a new v5 machine like `podman machine init v5`
and start Podman Desktop. 

it should prompt you to delete v4 machines (so only `podman-machine-default` and not `v5`

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -973,6 +973,7 @@ export function registerOnboardingUnsupportedPodmanMachineCommand(): extensionAp
 export function registerOnboardingRemoveUnsupportedMachinesCommand(): extensionApi.Disposable {
   return extensionApi.commands.registerCommand('podman.onboarding.removeUnsupportedMachines', async () => {
     const fileAndFoldersToRemove = [];
+
     const installedPodman = await getPodmanInstallation();
 
     if (extensionApi.env.isMac && installedPodman?.version.startsWith('5.')) {
@@ -1002,7 +1003,13 @@ export function registerOnboardingRemoveUnsupportedMachinesCommand(): extensionA
     }
 
     // check if unmarshalling errors
-    const machineListOutput = await getJSONMachineList();
+    let machineListError = '';
+    try {
+      const machineListOutput = await getJSONMachineList();
+      machineListError = machineListOutput.stderr;
+    } catch (error) {
+      machineListError = error.stderr;
+    }
 
     let machineFolderToCheck: string | undefined;
     // check invalid config files only with v5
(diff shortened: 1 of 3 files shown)


extensions/podman/src/extension.ts
```diff
@@ -150,6 +164,17 @@ export async function updateMachines(provider: extensionApi.Provider): Promise<v
   if (installedPodman) {
     shouldCleanMachine = shouldNotifyQemuMachinesWithV5(installedPodman);
   }
+  // check if the machine needs to be cleaned for v4 --> v5 format
+  if (!shouldCleanMachine) {
+    shouldCleanMachine = isIncompatibleMachineOutput(machineListOutput.stderr);
```
lstocchi, line 176: On Windows we never reach this point bc the `getJSONMachineList` function above fails and it throws

benoitf: ok so on Windows there is an exit code

benoitf:
@lstocchi ok so I tried to handle now the valid or invalid exit code when we do 'podman machine list'

extensions/podman/src/extension.ts
benoitf: fixed

lstocchi (approved):
Now it works fine.
The only issue is that if you have a running machine, the wsl instance does not get deleted and the user see a big red warning, which is not reassuring. After closing the warning and going forward everything works fine though.
If i switch back to v4 i can see all machines are deleted

```
Error removing WSL machine test1 Command execution failed with exit code 4294967295: Command execution failed with exit code 4294967295
```

benoitf:
the problem on windows is that as `podman machine list` is completely broken we don't have the state of the machines, etc.
I'm not sure we can even stop the machine.

If we do the update with Podman Desktop the stop is performed before the upgrade.

I'll try to see in a follow-up PR

lstocchi:
> the problem on windows is that as `podman machine list` is completely broken we don't have the state of the machines, etc. I'm not sure we can even stop the machine.
> 
> If we do the update with Podman Desktop the stop is performed before the upgrade.
> 
> I'll try to see in a follow-up PR

maybe we should try to terminate/shutdown the wsl instances. 
BTW I agree, I approved as it can be fixed/seen in another PR 👍

PR #6576, opened 2024-04-01
feat: always on buddy widget to manage identities

### LINKED ISSUE #6297: Introduce an always on widget to manage identities

To sign in to 3rd party providers, you need to go to Settings/Authentication page

Signed-in should be visible on the Podman Desktop window.

Add a buddy icon on top of the Settings icon

The buddy icon, should allow by clicking on it to have drop-down like options

dropdown should include :
-  Sign-in for registered providers
- Display identity if already logged in and option to sign-out 
- Link for "Manage authentication" that will redirect the user to Authentication providers page

status bar or title bar

for titlebar it doesn't match well
for statusbar, need to revisit first all items being displayed as if each account has its own title/identity being displayed there it won't scale as the space in status bar is limited

### Additional context

_No response_

dgolovin:
### What does this PR do?

Adds buddy always on widget to manage identities

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/620330/45ab96b6-1313-485b-9eb0-fb730a45c7ca

Fix #6297

### How to test this PR?

Install auth extension and use steps from demo above.

### DIFF
--- a/packages/main/src/plugin/authentication.ts
+++ b/packages/main/src/plugin/authentication.ts
@@ -27,6 +27,9 @@ import type {
   Event,
   ProviderImages,
 } from '@podman-desktop/api';
+import type { MenuItemConstructorOptions } from 'electron';
+import { BrowserWindow, Menu } from 'electron';
+import type { MenuItem } from 'electron/main';
 
 import type { ApiSenderType } from './api.js';
 import { Emitter } from './events/emitter.js';
@@ -63,7 +66,7 @@ export interface AllowedExtension {
 }
 
 export interface SessionRequest {
-  [scopes: string]: string[]; // maps sting with scopes to provider ids
+  [scopes: string]: string[]; // maps string with scopes to extension id's
 }
 
 export interface SessionRequestInfo {
@@ -75,6 +78,64 @@ export interface SessionRequestInfo {
 }
 
 export class AuthenticationImpl {
+  async showAccountsMenu(x: number, y: number): Promise<void> {
+    const template: (MenuItemConstructorOptions | MenuItem)[] = [
+      {
+        label: 'Manage authentication',
+        click: (): void => {
+          this.apiSender.send('navigate', {
+            page: 'authentication-providers',
+          });
+        },
+      },
+      {
+        type: 'separator',
+      } as MenuItem,
+    ];
+
+    this.getSessionRequests().forEach(request => {
+      const provider = this._authenticationProviders.get(request.providerId);
+      if (!provider) return;
+      const providerLabel = provider.label ? provider.label : provider.id;
+      template.push({
+        label: `Sign in with ${providerLabel} to use ${request.extensionLabel}`,
+        click: () => {
+          void this.executeSessionRequest(request.id);
+        },
+      });
+    });
+
+    const providers = Array.from(this._authenticationProviders.values());
+
+    const menuItemPromises = providers.map(meta =>
+      meta.provider.getSessions().then(sessions => {
+        sessions.forEach(session => {
+          template.push({
+            label: `${session.account.label} (${meta.label})`,
+            submenu: [
+              {
+                label: 'Sign out',
+                click: (): void => {
+                  void this.signOut(meta.id, session.id);
+                },
+              },
+            ],
+          });
+        });
+      }),
+    );
+
+    await Promise.all(menuItemPromises);
+
+    const menu = Menu.buildFromTemplate(template);
+    const zf = BrowserWindow.getFocusedWindow()?.webContents.getZoomFactor();
+    const zoomFactor = zf ? zf : 1;
+    menu.popup({
+      x: Math.round(x * zoomFactor),
+      y: Math.round(y * zoomFactor),
+    });
+  }
+
   private _authenticationProviders: Map<string, ProviderWithMetadata> = new Map<string, ProviderWithMetadata>();
   // map of scopes to extension ids
   private _signInRequests: Map<string, SessionRequest> = new Map();
@@ -258,7 +319,11 @@ export class AuthenticationImpl {
       throw new Error(`Requested authentication provider '${data.providerId}' is not installed.`);
     }
 
-    await provider.createSession(data.scopes);
+    const session = await provider.createSession(data.scopes);
+    if (session) {
+      this._signInRequestsData.delete(id);
+      this._signInRequests.delete(data.providerId);
+    }
   }
 
   async removeSession(providerId: string, sessionId: string): Promise<void> {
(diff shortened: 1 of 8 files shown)


benoitf:
Hello, I'm wondering why you're using native menu

When you asked I replied to not use them there
https://github.com/containers/podman-desktop/issues/5979#issuecomment-1999341435


AFAIK Podman Desktop has a min-height/min-width than VS Code don't have
so probably in our case, elements can fit

benoitf (commented):
AFAIK it looks like we're not displaying the icon of the auth provider being signed-in, should we not display it ?

dgolovin:
> AFAIK it looks like we're not displaying the icon of the auth provider being signed-in, should we not display it ?
@benoitf Where I should show it? There is going to be status line item to show that for Red Hat SSO (for example), but that is provider's extension contribution.

benoitf:
let's postpone the icon it's not urgent, was just wondering

dgolovin:
> When you asked I replied to not use them there
https://github.com/containers/podman-desktop/issues/5979#issuecomment-1999341435

@benoitf I missed that one ;(

Do you mean using this kind of menu instead?

image

Does it support submenus?

benoitf:
I don't know if it supports submenu...

benoitf:
but yes we're using these menus on left click in the UI

dgolovin:
@benoitf 

image

I can use only one level, because there is only one submenu item for now. 

So I can show podman-desktop 'native' menu that would trigger 'Sign out'. It would have a label

`` Sign out from email@address.com (Red Hat SSO) ``

Would that work?

benoitf:
but yes we're using these menus on left click in the UI

dgolovin:
Electron's menu Items in recording shown on left click as well.

dgolovin:
Should I change UI to one level DropDownMenu as I suggested above?

benoitf:
I don't know if we can 'fake' submenus by having like the name of the account and the next line being the action

Like we do in tray icon


----
myaccount@foo.com
Sign in
-----
Foo@another.com
Sign out
---


Or just have the action
'signout from foo@another.com'
'signin into foo.com'


we can also merge with menus and iterate

dgolovin:
It seems DropDownMenu has multiple issues that won't make it easy to use here.

it designed to be used in specific cases and always aligned left from the item. It expands down if there is enough space or up without checking if there is enough space.

image

The icon size and selection color on click is not adjustable.

When I try to use it for buddy icon implementation I got this

image

It always shows a button that would need to be suppressed for buddy icon implementation.

benoitf:
Let's defer the usage for the first ireration

dgolovin:
> Let's defer the usage for the first ireration

Usage of the buddy?

benoitf:
@dgolovin no the usage of the podman desktop component and keep native menu the time that we improve the component

deboer-tim:
IMHO it’s pretty confusing that one of the nav menu items opens a menu instead of navigating. Do we have a design for this? I wonder if the task manager pattern might work better or we can at least do something to make it look different.

dgolovin:
>IMHO it’s pretty confusing that one of the nav menu items opens a menu instead of navigating. Do we have a design for this? I wonder if the task manager pattern might work better or we can at least do something to make it look different.

I am open to other suggestions. My understanding was that buddy icon was discussed at UX meeting, even though I implemented it with native menus instead of DropDownMenu.

I would suggest to go with this one and then rework UI later.

benoitf:
yes it was discussed at a UX meeting including several other options like entry in the statusbar, titlebar, etc.
Then we created the issue to implement it.

Can be revisited but for sure there were some discussions

packages/main/src/plugin/authentication.ts
```diff
@@ -75,6 +78,64 @@ export interface SessionRequestInfo {
...
+          });
+        },
+      },
+      {
+        type: 'separator',
+      } as MenuItem,
+    ];
+
+    this.getSessionRequests().forEach(request => {
+      const provider = this._authenticationProviders.get(request.providerId);
```
benoitf: I  think the whole method will make it hard to move the code to the renderer part because here the backend side is doing all the stuff (UI calls and "backend side")

Could we change to have a separate object/class that will do only UI stuff (I suppose this class should not have any BrowserWindow as it should do only authentication stuff, no menu display)

then, when moving the menu to the renderer part this class/file should be untouched

packages/main/src/plugin/authentication.ts
```diff
@@ -75,6 +78,64 @@ export interface SessionRequestInfo {
...
+    const menuItemPromises = providers.map(meta =>
+      meta.provider.getSessions().then(sessions => {
+        sessions.forEach(session => {
+          template.push({
+            label: `${session.account.label} (${meta.label})`,
+            submenu: [
+              {
+                label: 'Sign out',
+                click: (): void => {
+                  void this.signOut(meta.id, session.id);
```
benoitf: should not have the void prefix there

benoitf: I think there should not have any direct calls using this as it's making harder to move the code later to the renderer part.
It should delegate calls to the IPC

packages/main/src/plugin/authentication.ts
dgolovin: We already have all calls required for menu generation on UI side in preload:
1. Sign in
2. Sign out
3. Getting ProvidersInfo with sessions info
4. Getting authentication requests list
5 Navigation on UI side is just router.goto call

This method should be simply thrown out when DropDownMenu is ready to be used.

packages/main/src/plugin/authentication.ts
dgolovin: void is for eslint to stop complaining. it ether this or ```this.signOut(meta.is, session.id).catch((err) => console.log(err))```

benoitf: yes please use the catch, we should not bypass/fake the linter.

packages/main/src/plugin/authentication.ts
benoitf: so why not using a separate class that will do this rather than mixing client side and server side in authentication.ts ?

packages/main/src/plugin/authentication.ts
dgolovin: Updated

packages/main/src/plugin/authentication.ts
dgolovin: I moved out Menu and BrowserWindow related code to authentication-menu.ts
Tests are updates as well.

benoitf (approved):
tested and works fine on my end

dgolovin:
Opened follow up issue #6660.

PR #6602, opened 2024-04-02
fix: make all rows being selectable for image list

### LINKED ISSUE #6592: I can't select some images for bulk operations to save them

Before bulk operation on images was for deletion, so it was not possible to select images that were being used by containers

But now as we can save the content of the images, I don't see why we could not select all the images

Add ability to select all images.

The hard part is what should we do in case we want to delete all images:
- disable the button as soon as we pickup some images being used
- skip images being used when clicking on delete
- do not select/check images by default when clicking on select all but manual click is still possible

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Until now we were restricting the selection to images without having a running container
but then we can't apply other bulk operations like 'save'

Fixes it allows to select all images to save/export them to a tarball

fixes https://github.com/containers/podman-desktop/issues/6592

### How to test this PR?

See that now all lines can be selected

### DIFF
--- a/packages/renderer/src/lib/image/ImagesList.svelte
+++ b/packages/renderer/src/lib/image/ImagesList.svelte
@@ -286,8 +286,7 @@ const columns: Column<ImageInfoUI>[] = [
 ];
 
 const row = new Row<ImageInfoUI>({
-  selectable: image => image.status === 'UNUSED',
-  disabledText: 'Image is used by a container',
+  selectable: () => true,
 });
 </script>
 


deboer-tim (commented):
This fixes the issue for exporting images, but it also means that you can easily bulk delete images. If an image was in use before you had to very intentionally stop the container and delete it first, now you can just delete the image and both will (eventually) disappear. Although this can be handy and quicker, it seems like maybe we should add a warning like "Image(s) are currently in use, are you sure you want to delete them?"

benoitf:
in the issue I gave some possible solutions.

I don't see any good solution for now

we could also disable trash icon if I we select at least one image with a a running container
or select all don't select by default the images with a running container (but manually you can then click on each row)
etc.

Issue #6414, opened 2024-03-13
Show all of Command in Summary


avlec:
### Is your enhancement related to a problem? Please describe

When running a container with more than one argument it shows only a single one of the arguments passed to podman run.
e.g.,  `podman run -it busybox sh -c "sh"` looks like this in the UI. For programs that use a combination of ENTRYPOINT and CMD to supplement arguments this current display does not suffice.

![image](https://github.com/containers/podman-desktop/assets/17734191/e5be7a88-7855-4a28-bd89-e367f4c60350)

### Describe the solution you'd like

Show all the arguments passed to run. From the above example this would be `["sh", "-c" "sh"]` exact quirks of how this is displayed is up for grabs. Perhaps even adding on the entry point so what's going on is more easily understandable

PR #6604, opened 2024-04-03
fix(container-detail): display the entire command instead of the first piece only

GLEF1X:

### What does this PR do?

This pull request modifies the UI to display all arguments passed to a container's run command rather than showing just the first piece for improved clarity on the processes being run within the container.

#### Before
CleanShot 2024-04-02 at 22 56 23@2x

#### After
CleanShot 2024-04-02 at 22 55 17@2x

#6414 - Show all of Command in Summary 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/api/container-info.ts
+++ b/packages/main/src/plugin/api/container-info.ts
@@ -42,7 +42,7 @@ export interface ContainerInfo {
   Names: string[];
   Image: string;
   ImageID: string;
-  Command?: string;
+  Command?: readonly string[];
   Created: number;
   Ports: ContainerPortInfo[];
   Labels: { [label: string]: string };
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -826,7 +826,7 @@ describe('listContainers', () => {
     expect(container.StartedAt).toBe('2023-08-10T13:37:44.000Z');
     expect(container.pod).toBeUndefined();
     expect(container.Id).toBe('31a4b282691420be2611817f203765402d8da7e13cd530f80a6ddd1bb4aa63b4');
-    expect(container.Command).toBe('httpd-foreground');
+    expect(container.Command).toStrictEqual(['httpd-foreground']);
     expect(container.Names).toStrictEqual(['/admiring_wing']);
     expect(container.Image).toBe('docker.io/library/httpd:latest');
     expect(container.ImageID).toBe('sha256:911d72fc5020723f0c003a134a8d2f062b4aea884474a11d1db7dcd28ce61d6a');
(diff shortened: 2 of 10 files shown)


packages/main/src/plugin/api/container-info.ts
```diff
@@ -42,7 +42,7 @@ export interface ContainerInfo {
   Names: string[];
   Image: string;
   ImageID: string;
-  Command?: string;
+  Command?: readonly string[];
```
axel7083, line 45: From my understanding, this interface is inspired from dockerode definiton

https://github.com/codershiba/DefinitelyTyped/blob/2639c808780fbe2c5ac588c14c7344fd8fdd1c1c/types/dockerode/index.d.ts#L351-L356

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -1031,7 +1031,7 @@ describe('listContainers', () => {
     expect(container.StartedAt).toBe('2023-08-10T13:37:44.000Z');
     expect(container.pod).toBeUndefined();
     expect(container.Id).toBe('31a4b282691420be2611817f203765402d8da7e13cd530f80a6ddd1bb4aa63b4');
-    expect(container.Command).toBe(undefined);
+    expect(container.Command).toBe(null);
```
axel7083: I am not sure to understand why null would be used here instead of undefined

packages/main/src/plugin/container-registry.ts
```diff
@@ -476,7 +475,7 @@ export class ContainerProviderRegistry {
                 Created: moment(podmanContainer.Created).unix(),
                 State: podmanContainer.State,
                 StartedAt,
-                Command: podmanContainer.Command?.length > 0 ? podmanContainer.Command[0] : undefined,
+                Command: podmanContainer.Command,
```
axel7083: if the null is coming from here we should try to avoid it, and use a nullish coalescing operator
```suggestion
                Command: podmanContainer.Command ?? undefined,
```

packages/main/src/plugin/container-registry.spec.ts
GLEF1X: AFAIK, `libpodApi` returns `null` if the command is empty, whereas `dockerode` returns `undefined`

packages/main/src/plugin/api/container-info.ts
GLEF1X: AFAIK, you're totally right. However, I changed it to `readonly string[]` because `listContainers` dynamically selects among various providers, leading to the use of workarounds such as `podmanContainer.Command?.length > 0 ? podmanContainer.Command[0]: undefined`. In my opinion, `ContainerInfo` should use an array to represent a command even though `dockerode` returns a string because it enables us to not lose data(in this case other parts of the command).

axel7083: Yes I looked at the podman repository and they are indeed sending an array compared to dockerode definition.

https://github.com/containers/podman/blob/17f36a369035b4d5c27f416af9cb961ede807eed/pkg/domain/entities/types/container_ps.go#L14-L15

I would be in favour to get something closer to what podman provide, the changes you are providing only touch the internal api, so it would not have much side effect, however the `ContainerInfo` is also provided to the `extension-api`. 

https://github.com/containers/podman-desktop/blob/689b3f80eb609b3dc3fece61373f83bf0027bd1b/packages/extension-api/src/extension-api.d.ts#L1970

However for the `extension-api` we have something different, with an interface called `SimpleContainerInfo` which is some extending of dockerode.

https://github.com/containers/podman-desktop/blob/6c5b7848206fe65cda848be4a4d8185e19cd183b/packages/main/src/plugin/api/container-info.ts#L53

And I am wondering if we change/fix it at one place we would not have to propagate the changes also elsewhere... Waiting for @benoitf opinion on the matter

benoitf:
hello I'm reviewing all PRs being opened

@GLEF1X could you rebase your PR ? thanks

benoitf:
closing

PR #6606, opened 2024-04-03
chore(api): creating api package

### LINKED ISSUE #6300: Create `api` package for sharing code between `main` and `renderer`

With electron, there are like 3 main packages, preload, main and renderer

but for example renderer can call main API to grab like a list of containers, it returns objects defined in main package using imports like '../../../main'

The problem is that you may import code that has some dependency to some Node.js classes/interfaces and it's failing

To have a better scope dependency, we should import the definition from main and renderer from another package

move `src/plugin/api` to its own 'api' package

so, `api` will be for internal api, `extension-api` for the extension API

and both main and renderer package will be able to import from `api`

a new alias should be added like `/@api` to ease the import from both main and renderer packages

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Moving directly all `packages/main/src/plugin/api` in one PR would make it hard to review due to the numbers of import that would be impacted. This first PR create the `packages/api` folder, and set the required configuration in the packages that will use the `/@api/*`.

⚠️ Packages that will not use the api are: `packages/ui` and `packages/extension-api`.

Related to https://github.com/containers/podman-desktop/issues/6300

### How to test this PR?

### DIFF
--- a/packages/api/tsconfig.json
+++ b/packages/api/tsconfig.json
@@ -4,7 +4,6 @@
     "target": "esnext",
     "sourceMap": false,
     "moduleResolution": "Node",
-    "skipLibCheck": true,
     "strict": true,
     "allowSyntheticDefaultImports": true,
     "isolatedModules": true,
--- a/packages/api/tsconfig.json
+++ b/packages/api/tsconfig.json
@@ -16,5 +16,5 @@
       "/@/*": ["./src/*"]
     }
   },
-  "include": ["src/**/*.ts", "exposedInDockerExtension.d.ts", "../../types/**/*.d.ts", "../api/src/**/*.ts"]
+  "include": ["src/**/*.ts", "../../types/**/*.d.ts"]
 }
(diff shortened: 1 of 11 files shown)


benoitf:
hi @axel7083 , AFAIK technical debts issues are for post 1.10 release

axel7083:
> hi @axel7083 , AFAIK technical debts issues are for post 1.10 release

Okey, I was not aware as I had the issue assigned for this sprint

benoitf:
@axel7083 I think we moved it from the previous sprint to the next one but it should be moved to the next one and finally we should be able to handle lot of this kind of issues 🎉

benoitf:
need to have 1.10 promoted release before being able to merge (to have easy backports to the 1.10 branch)

benoitf (approved):

packages/api/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "module": "esnext",
+    "target": "esnext",
+    "sourceMap": false,
+    "moduleResolution": "Node",
+    "skipLibCheck": true,
```
benoitf: maybe use false here if we can (or have the stronger checks)


it probably misses the checks for TypeScript as well (and optional unit tests that should be run as part of the root package.json file)

axel7083: Removed! We will see :)

Issue #6627, opened 2024-04-03
Allow user to navigate to container's image from the container detail page


GLEF1X:
### Is your enhancement related to a problem? Please describe

The enhancement allows for a more streamlined user experience by enabling direct navigation from a container's detail page to its associated image information. This improves usability by reducing the steps needed for users to access detailed image data, thereby enhancing the overall efficiency of the Podman desktop client.

### Describe the solution you'd like

![CleanShot 2024-04-03 at 11 37 49@2x](https://github.com/containers/podman-desktop/assets/71976818/651c3958-d9c4-4b0d-b5a5-9378fca6e225)

Make the container detail subtitle a link that would redirect to the image detail summary page

PR #6628, opened 2024-04-03
feat(container-detail): make image name a link that redirects to image detail page

GLEF1X:

### What does this PR do?

This PR enables users to be able to navigate to the image quicker and more efficiently from the container detail page. 

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/71976818/d3b8a86e-59bf-44f3-a18b-0b912d71a97f

#6627 

### How to test this PR?


packages/renderer/src/lib/ui/DetailsPage.svelte
```diff
@@ -44,7 +45,11 @@ function handleKeydown(e: KeyboardEvent) {
             <div class="text-violet-400 ml-2 leading-normal" class:hidden="{!titleDetail}">{titleDetail}</div>
           </div>
           <div>
-            <span class="text-sm leading-none text-gray-900" class:hidden="{!subtitle}">{subtitle}</span>
+            {#if subtitleHref}
+              <Link internalRef="{subtitleHref}">{subtitle}</Link>
+            {:else}
+              <span class="text-sm leading-none text-gray-900" class:hidden="{!subtitle}">{subtitle}</span>
+            {/if}
```
benoitf: hello, AFAIK this is not required

I mean, subtitle is also a slot so if you want to customize it you can pass the subtitle slot rather than the subtitle property

packages/renderer/src/lib/container/ContainerDetails.svelte
```diff
@@ -68,7 +73,11 @@ onMount(() => {
 </script>
 
 {#if container}
-  <DetailsPage title="{container.name}" subtitle="{container.shortImage}" bind:this="{detailsPage}">
+  <DetailsPage
+    title="{container.name}"
+    subtitle="{container.shortImage}"
+    subtitleHref="{containerImageHref}"
```
benoitf: need to use svelte:fragment slot="subtitle" probably

packages/renderer/src/lib/container/ContainerDetailsSummary.spec.ts
```diff
@@ -0,0 +1,51 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/renderer/src/lib/container/ContainerDetailsSummary.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      shortId: 'myContainer',
+      command: command,
+      state: 'RUNNING',
+      hasPublicPort: false,
+    } as unknown as ContainerInfoUI;
+
+    render(ContainerDetailsSummary, { container });
+    expect(screen.getByTestId('container-details-command-tr').classList.contains('hidden')).to.be.true;
+  });
+});
```
benoitf: it does not seem to the tests for this PR

packages/renderer/src/lib/container/ContainerDetails.svelte
```diff
@@ -58,6 +60,9 @@ onMount(() => {
             router.goto(`${currentRouterPath}logs`);
           }
         }
+
+        const base64RepoTag = containerUtils.encodeContainerImage(container.image);
+        containerImageHref = `/images/${inspect.Image}/${container.engineId}/${base64RepoTag}/summary`;
```
benoitf: if we need that information, it should be handle on the 'main side'/server side as a field of the container object

packages/renderer/src/lib/container/container-utils.ts
```diff
@@ -330,4 +330,10 @@ export class ContainerUtils {
       .filter(part => !part.startsWith('is:'))
       .join(' ');
   }
+
+  encodeContainerImage(imageName: string): string {
+    const encoder = new TextEncoder();
+    const uint8Array = encoder.encode(imageName);
+    return window.btoa(String.fromCharCode(...uint8Array));
+  }
```
benoitf: should not be required on the frontend side

packages/renderer/src/lib/ui/DetailsPage.svelte
GLEF1X: CleanShot 2024-04-03 at 16 26 33@2x

Hi Florent, that's a totally valid point. However, implementing this would result in the image name being duplicated, since it would appear both as a standalone name and as a link. I could potentially rename the link to "go to image", but I don't think that would contribute to creating good UI. I feel like it's better to do conditional link/subtitle

packages/renderer/src/lib/container/ContainerDetailsSummary.spec.ts
GLEF1X: Yeah, sorry about that, I'll get rid of it

packages/renderer/src/lib/container/ContainerDetails.svelte
GLEF1X: Okay, thank you for your feedback!

packages/main/src/plugin/container-registry.ts
```diff
@@ -435,7 +435,6 @@ export class ContainerProviderRegistry {
           }
 
           // if we have a libpod API, grab containers using Podman API
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
GLEF1X, line 438: It's not needed here as type is defined for the `containers` variable. I believe this is legacy

packages/renderer/src/lib/ui/DetailsPage.svelte
benoitf: so it won't be duplicated b/c you need to remove the subtitle property that you're giving to DetailsPage

something like:

```
 <DetailsPage title="{container.name}" bind:this="{detailsPage}">
    <StatusIcon slot="icon" icon="{ContainerIcon}" size="{24}" status="{container.state}" />
    <svelte:fragment slot="subtitle">
      <Link internalRef="{container.<your-new-href-property>}">{container.shortImage}</Link>
    </svelte:fragment>
```

image

GLEF1X: Gotcha, I did what you suggested, and it worked. Thank you for the feedback!

packages/renderer/src/lib/container/ContainerDetails.svelte
```diff
@@ -44,12 +44,13 @@ onMount(() => {
   // loading container info
   return containersInfos.subscribe(containers => {
     const matchingContainer = containers.find(c => c.Id === containerID);
+
```
benoitf:
```suggestion
```

packages/renderer/src/lib/container/ContainerDetails.svelte
```diff
@@ -44,12 +44,13 @@ onMount(() => {
...
     if (matchingContainer) {
       container = containerUtils.getContainerInfoUI(matchingContainer);
-
       // look if tty is supported by this container
       window.getContainerInspect(container.engineId, container.id).then(inspect => {
         displayTty = (inspect.Config.Tty || false) && (inspect.Config.OpenStdin || false);
+
```
benoitf:
```suggestion
```

benoitf (approved):

PR #6630, opened 2024-04-03
feat: add createManifest API

### LINKED ISSUE #6624: add API to call createManifest endpoint

enhance libpod Dockerode with createManifest endpoint

https://docs.podman.io/en/latest/_static/api.html#tag/manifests/operation/ManifestCreateLibpod

this method will allow to create a manifest

_No response_

### Additional context

_No response_

cdrage:
feat: add createManifest API

### What does this PR do?

Adds the createManifest API modelled by https://docs.podman.io/en/latest/_static/api.html#tag/manifests/operation/ManifestCreateLibpod

You'll be able to pass in createManifestOptions with the name as well as
an array of image names and be able to create a manifest.

### Screenshot / video of UI

N/A

Closes https://github.com/containers/podman-desktop/issues/6624

### How to test this PR?

Either through tests or you can also do the following in a svelte file:

```ts
    const options: ManifestCreateOptions = {
      name: containerImageName,
      images: [
        containerImageName],

    };
    await window.createManifest(options);
```

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -285,6 +285,16 @@ declare module '@podman-desktop/api' {
     provider?: ContainerProviderConnection;
   }
 
+  export interface ManifestCreateOptions {
+    images: string[];
+    name: string;
+    amend?: boolean;
+    all?: boolean;
+
+    // Provider to use for the manifest creation, if not, we will try to select the first one available (similar to podCreate)
+    provider?: ContainerProviderConnection;
+  }
+
   export interface KubernetesProviderConnectionEndpoint {
     apiURL: string;
   }
@@ -3367,6 +3377,9 @@ declare module '@podman-desktop/api' {
     export function listPods(): Promise<PodInfo[]>;
     export function stopPod(engineId: string, podId: string): Promise<void>;
     export function removePod(engineId: string, podId: string): Promise<void>;
+
+    // Manifest related methods
+    export function createManifest(options: ManifestCreateOptions): Promise<{ Id: string }>;
   }
 
   /**
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -3379,7 +3379,7 @@ declare module '@podman-desktop/api' {
     export function removePod(engineId: string, podId: string): Promise<void>;
 
     // Manifest related methods
-    export function createManifest(options: ManifestCreateOptions): Promise<{ Id: string }>;
+    export function createManifest(options: ManifestCreateOptions): Promise<{ engineId: string; Id: string }>;
   }
 
   /**
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -1225,7 +1225,7 @@ export class ContainerProviderRegistry {
     }
   }
 
-  async createManifest(manifestOptions: ManifestCreateOptions): Promise<{ Id: string }> {
+  async createManifest(manifestOptions: ManifestCreateOptions): Promise<{ engineId: string; Id: string }> {
     let telemetryOptions = {};
     try {
       let internalContainerProvider: InternalContainerProvider;
@@ -1243,7 +1243,7 @@ export class ContainerProviderRegistry {
       }
 
       const result = await internalContainerProvider.libpodApi.podmanCreateManifest(manifestOptions);
-      return { Id: result.Id };
+      return { engineId: internalContainerProvider.id, Id: result.Id };
     } catch (error) {
       telemetryOptions = { error: error };
       throw error;
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -53,6 +53,7 @@ import type { ContainerStatsInfo } from './api/container-stats-info.js';
 import type { HistoryInfo } from './api/history-info.js';
 import type { BuildImageOptions, ImageInfo, ListImagesOptions } from './api/image-info.js';
 import type { ImageInspectInfo } from './api/image-inspect-info.js';
+import type { ManifestCreateOptions } from './api/manifest-info.js';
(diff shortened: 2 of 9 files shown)


cdrage:
Could you do a preliminary review @benoitf ?

I have not written tests yet, but want to make sure this is a correct API implementation.

FYI it returns Promise<void> as it's modeled after the actual call: https://docs.podman.io/en/latest/_static/api.html#tag/manifests/operation/ManifestCreateLibpod similar to our other API calls.

Otherwise, this works and I can confirm it's working via:

```ts
    const options: ManifestCreateOptions = {
      name: containerImageName,
      images: [
        containerImageName],

    };
    await window.createManifest(options);
```

packages/main/src/plugin/container-registry.ts
```diff
@@ -1224,6 +1224,29 @@ export class ContainerProviderRegistry {
     }
   }
 
+  async createManifest(manifestOptions: containerDesktopAPI.ManifestCreateOptions): Promise<void> {
```
benoitf: it should not be ManifestCreateOptions from the desktopAPI there but from Manifest-info.ts

packages/main/src/plugin/container-registry.ts
```diff
@@ -1224,6 +1224,29 @@ export class ContainerProviderRegistry {
...
+      if (manifestOptions.provider) {
+        // grab connection
+        internalContainerProvider = this.getMatchingContainerProvider(manifestOptions.provider);
+      } else {
+        // Get the first running podman connection
+        internalContainerProvider = this.getFirstRunningPodmanContainerProvider();
+      }
+      if (!internalContainerProvider?.libpodApi) {
+        throw new Error('No podman provider with a running engine');
+      }
```
benoitf: if it's finding a provider but the provider is not supporting the podman API (like docker) I should not get the error message that no podman provider is running.


we need to differentiate: not finding a provider or not having the libpodAPI on the matching provider

packages/main/src/plugin/dockerode/libpod-dockerode.ts
```diff
@@ -777,5 +779,29 @@ export class LibpodDockerode {
         });
       });
     };
+
+    // add createManifest
+    prototypeOfDockerode.createManifest = function (manifestOptions: ManifestCreateOptions): Promise<unknown> {
```
benoitf:
```suggestion
    prototypeOfDockerode.podmanCreateManifest = function (manifestOptions: ManifestCreateOptions): Promise<unknown> {
```

benoitf: need to prefix with podman

packages/main/src/plugin/index.ts
```diff
@@ -750,6 +750,15 @@ export class PluginSystem {
         return containerProviderRegistry.removePod(engine, podId);
       },
     );
+
+    // manifest
+    this.ipcHandle(
+      'container-provider-registry:createManifest',
+      async (_listener, manifestOptions: containerDesktopAPI.ManifestCreateOptions): Promise<void> => {
```
benoitf: not the extension api there

packages/preload/src/index.ts
```diff
@@ -282,6 +282,11 @@ export function initExposure(): void {
     return ipcInvoke('container-provider-registry:restartPod', engine, podId);
   });
 
+  // Manifest
+  contextBridge.exposeInMainWorld('createManifest', async (createOptions: containerDesktopAPI.ManifestCreateOptions): Promise<void> => {
```
benoitf: not the extension api, should be from manifest info

cdrage:
Updates:
- ~FYI getting some odd issues with the API endpoint, so hold off on testing via window.createManifest~ clarified with https://github.com/containers/podman/issues/22255

TODO:
- [x] Add undocumented callback...
```ts
▶ curl --unix-socket /var/folders/cn/pm5mgrcs5h339nphypxtvd080000gn/T/podman/podman-machine-default-api.sock -X POST \
     "http://d/v4.2.0/libpod/manifests/test?images=quay.io/bootc-extension/httpd"

{"Id":"6a105d8fe5a76e15c4687f44deb923acf62608d56ede79b74f5a9251de5e2155"}
```

Returns ID which is undocumented in the API.
- [x] Add tests

cdrage:
Opened up issue https://github.com/containers/podman/issues/22255 for podman upstream for clarity on the API.

cdrage:
Tests have been added and return value corrected, ready for review!

cdrage:
Tests have been added and return value corrected, ready for review!

packages/extension-api/src/extension-api.d.ts
```diff
@@ -285,6 +285,14 @@ declare module '@podman-desktop/api' {
     provider?: ContainerProviderConnection;
   }
 
+  export interface ManifestCreateOptions {
+    images: string[];
+    name: string;
+
```
benoitf: we're missing some properties

`all` and `amend`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -3367,6 +3377,9 @@ declare module '@podman-desktop/api' {
     export function listPods(): Promise<PodInfo[]>;
     export function stopPod(engineId: string, podId: string): Promise<void>;
     export function removePod(engineId: string, podId: string): Promise<void>;
+
+    // Manifest related methods
+    export function createManifest(options: ManifestCreateOptions): Promise<{ Id: string }>;
```
benoitf: it might help if we return also the engineId

benoitf (approved):
LGTM (return the engineId in addition to the Id, because you may need the engineId when you may want to inspect it later)

If you could use fixup commits rather than always pushing a single commit during the review, it would help to know the diff.

Now, each time I need to review the whole changes and try to remember what was the previous state

cdrage:
@benoitf I'll make sure it is separate commits now for any review updates.

Updated this PR with returning engineId

benoitf (approved):

deboer-tim (approved):

PR #6657, opened 2024-04-04
fix(featured-extension): make limit configurable

### LINKED ISSUE #6656: feat(feature-extension): the number of feature extension returned should be configurable

axel7083:
### What does this PR do?

This PR is required to make https://github.com/containers/podman-desktop/issues/6215 possible. As we need to have a way, to get all featured extension, and not only the first 6.

This PR is moving out the hardcoded value from the core of the method, to a `limit` argument with a default value. This makes is compatible with any existing usage, while keeping existing behaviour, but allowing new caller to get all by changing this value to `-1`

Fixes https://github.com/containers/podman-desktop/issues/6656

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/featured/featured.spec.ts
+++ b/packages/main/src/plugin/featured/featured.spec.ts
@@ -121,7 +121,7 @@ test('getFeaturedExtensions should check installable extensions', async () => {
   expect(crcExtension?.fetchLink).toBe(ociLink);
 });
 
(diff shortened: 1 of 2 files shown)


benoitf:
I'm not sure how this PR is related to to https://github.com/containers/podman-desktop/issues/6656

also the issue is created a couple of minutes before the PR so we don't have time to comment before a PR is issued

benoitf (approved):

PR #6681, opened 2024-04-05
feat: adding internal logic to manage recommended extensions based on featured extensions

axel7083:
### What does this PR do?

This PR is the first for adding the recommended extension banner in the dashboard, this PR focus on the internal aspect, without affecting the UI.

To ease review, the UI components, using those features will be provided in a follow up PR.

### Screenshot / video of UI

Here is a schema of how the properties of `ExtensionBanner` interface will be displayed in the banner.

image

Since https://github.com/containers/podman-desktop/pull/6673 `FeaturedExtension.svelte` is a reusable component. We can use it to place it inside the banner.

Partially cover https://github.com/containers/podman-desktop/issues/6215

### How to test this PR?


packages/renderer/src/stores/extensionBanners.ts
```diff
@@ -0,0 +1,54 @@
+/**********************************************************************
+ * Copyright (C) 2022-2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/preload/src/index.ts
```diff
@@ -1260,6 +1261,11 @@ export function initExposure(): void {
   contextBridge.exposeInMainWorld('getFeaturedExtensions', async (): Promise<FeaturedExtension[]> => {
     return ipcInvoke('featured:getFeaturedExtensions');
   });
+
+  contextBridge.exposeInMainWorld('getExtensionBanners', async (): Promise<ExtensionBanner[]> => {
+    return ipcInvoke('featured:getExtensionBanners');
```
benoitf: should it be

```suggestion
    return ipcInvoke('recommended:getExtensionBanners');
```

packages/main/src/plugin/index.ts
```diff
@@ -1518,6 +1520,10 @@ export class PluginSystem {
       return featured.getFeaturedExtensions();
     });
 
+    this.ipcHandle('featured:getExtensionBanners', async (): Promise<ExtensionBanner[]> => {
```
```suggestion
    this.ipcHandle('recommended:getExtensionBanners', async (): Promise<ExtensionBanner[]> => {
```

packages/main/src/plugin/recommendations/recommendations-registry.spec.ts
```diff
@@ -53,3 +79,103 @@ test('should register a configuration', async () => {
...
+      categories: [],
+      displayName: '',
+      fetchable: false,
+      icon: '',
+      installed: false,
+    };
+    vi.mocked(featuredMock.getFeaturedExtensions).mockResolvedValue([featured]);
+
+    const extensions = await recommendationsRegistry.getExtensionBanners();
+    expect(extensions.length).toBe(1);
```
benoitf, line 141: looks like tests are failing

benoitf (commented):
forgot to hit submit review last week

packages/main/src/plugin/recommendations/recommendations-registry.spec.ts
axel7083: Yeah I moved the recommendations.json file to the root, and forgot to change the mock import because my IDE is not smart enough !

packages/main/src/plugin/recommendations/recommendations-registry.spec.ts
```diff
@@ -16,23 +16,49 @@
...
 
 import { RecommendationsRegistry } from './recommendations-registry.js';
 
 let recommendationsRegistry: RecommendationsRegistry;
 
-const registerConfigurationsMock = vi.fn();
+const mocks = vi.hoisted(() => ({
+  registerConfigurationsMock: vi.fn(),
+  getRecommendationIgnored: vi.fn(),
+}));
```
benoitf: why do we need to use that pattern

we already has `const registerConfigurationsMock = vi.fn();`

benoitf: AFAIK we don't need these hoisted

axel7083: I removed them !

packages/main/src/plugin/recommendations/recommendations-registry.ts
```diff
@@ -16,12 +16,56 @@
...
+  constructor(
+    private configurationRegistry: ConfigurationRegistry,
+    private featured: Featured,
+  ) {}
+
+  isRecommendationDisabled(): boolean {
+    return this.configurationRegistry
+      .getConfiguration(RecommendationsSettings.SectionName)
+      .get<boolean>(RecommendationsSettings.IgnoreRecommendations, false);
+  }
```
benoitf: I think it works better for readability when the boolean is true if enabled and false if disabled

`isRecommandationsEnabled()`

benoitf (approved):

packages/main/src/plugin/recommendations/recommendations-registry.ts
axel7083: Why not, I added two tests to ensure expected behaviour with the renamed method.

benoitf: It's not about the tests,

it's that when something is disabled, then you have what we could called double negation like

'not (disabled)' meaning 'not not enabled' (meaning enabled)

PR #6684, opened 2024-04-05
chore: extract downloadAndSha to a separate class to reused it

benoitf:
### What does this PR do?
for https://github.com/containers/podman-desktop/issues/6360 I need to validate some github artifacts, and then having a dedicated class will help

related to for https://github.com/containers/podman-desktop/issues/6360 

### How to test this PR?

build should be 🟢 

### DIFF
--- a/extensions/podman/scripts/podman-download.spec.ts
+++ b/extensions/podman/scripts/podman-download.spec.ts
@@ -18,10 +18,17 @@
 
 import { afterEach } from 'node:test';
 import { beforeEach, describe, expect, test, vi } from 'vitest';
-import { PodmanDownload, PodmanDownloadFcosImage, PodmanDownloadFedoraImage, ShaCheck } from './podman-download';
+import {
+  DownloadAndCheck,
+  PodmanDownload,
+  PodmanDownloadFcosImage,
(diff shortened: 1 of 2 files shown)


extensions/podman/scripts/podman-download.spec.ts
```diff
@@ -63,6 +66,10 @@ class TestPodmanDownload extends PodmanDownload {
   public getShaCheck(): ShaCheck {
     return super.getShaCheck();
   }
+
+  public getDownloadAndCheck(): DownloadAndCheck {
+    return super.getDownloadAndCheck();
```
dgolovin, line 71: Just curious, why make method from base class protected, then override it in ancestor and make it public? What is the idea here?

dgolovin (approved):
LGTM

extensions/podman/scripts/podman-download.spec.ts
benoitf: to call the method directly from a unit test (but not being able to call it directly when we instantiate the class in the code base)

benoitf: (or to replace the method during test)

PR #6692, opened 2024-04-06
feat(webview-telemetry): make telemetry for webviews more detailed

### LINKED ISSUE #6686: Telemetry for navigating to webviews

When you navigate to any extension that provides a webview, it only sends a generic telemetry event:
`"name": "Home/Webview"`
regardless of which extension's Webview you navigated to.

We're not able to tell which webview you're using. It would be good if the event included the name of the webview or extension in the event. e.g. for DD Extensions we send:
`  "name": "Home/Extension/Disk usage"`

_No response_

### Additional context

_No response_

GLEF1X:

### What does this PR do?

This PR makes more detailed telemetry tracks for page navigation to the extensions that provide web views.

Fixes #6686 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/webview/Webview.svelte
+++ b/packages/renderer/src/lib/webview/Webview.svelte
@@ -80,7 +80,7 @@ onDestroy(() => {
 </script>
 
 {#if preloadPath && webViewPort && webviewInfo}
-  <Route path="/*">
+  <Route path="/*" breadcrumb="{webviewInfo.name}">
     <webview
       bind:this="{webviewElement}"
       aria-label="Webview {webviewInfo?.name}"


benoitf (commented):
Hello @GLEF1X 

thanks for looking at the issue

I think it might be much simpler if for Webview we add the breadcrumb instead of adding a dependency from the routes to the webview (this dependency should not exists)

I think you can add an extra breadcrumb attribute in the Route of the webview
something like https://github.com/containers/podman-desktop/blob/892b7e71175c8d0aa3209847cacfb3d4fbe95dd2/packages/renderer/src/lib/docker-extension/DockerExtension.svelte#L42

but there:
https://github.com/containers/podman-desktop/blob/892b7e71175c8d0aa3209847cacfb3d4fbe95dd2/packages/renderer/src/lib/webview/Webview.svelte#L83

using for example `webviewInfo.name`

GLEF1X:
@benoitf Thank you for reviewing. Your solution is definitely more concise and makes much more sense.

benoitf (approved):

Issue #6663, opened 2024-04-05
Full UI reload triggered periodically by `RangeError: Maximum call stack size exceeded`


dgolovin:
### Bug description

Monitoring resources with infinite recursive calls eventually triggers the RangeError exception, which triggers reload for current page and probably main thread restart (unconfirmed).

To find out suspects see steps to replicate below \\/ (first suspect is in extensions/podman/src/extension.ts).

### Operating system

macOS M1 Sonoma 14.3

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

It manifests itself when podman desktop is running long enough and multiple extensions, like podman, container-registry and others alike are active. The reason is the pattern used to monitor certain resources:

```
function monitorFunction() {
   // do monitoring steps
  await new Promise(resolve => setTimeout(resolve, timeoutInMilliseconds);
  monitorFunction();
}
``` 

There is recursive monitorFunction call and it fills up stack if it runs long enough.

If all the patterns like above found and `timeoutInMilliseconds` removed the problem manifests itself much faster by reloading current page. You can set up the break point on mount() function for any svelte component on current page and break point is hit periodically which indicates page reload.

There is also another pattern that should be fine to use, because it is not making recursive call.

```
function monitorFunction(params) {
   // do monitoring steps
  setTimeout(() => monitorFunction(params), timeoutInMilliseconds);
}
```

or

```
function monitorFunction() {
   // do monitoring steps
  setTimeout(monitorFunction, timeoutInMilliseconds);
}
``` 

### Relevant log output

```shell
RangeError: Maximum call stack size exceeded
    at monitorProvider (/Users/eskimo/Sources/podman-desktop/extensions/podman/dist/extension.cjs:5176:1)
    at monitorProvider (/Users/eskimo/Sources/podman-desktop/extensions/podman/dist/extension.cjs:5173:3)
    at monitorProvider (/Users/eskimo/Sources/podman-desktop/extensions/podman/dist/extension.cjs:5173:3)
    at monitorProvider (/Users/eskimo/Sources/podman-desktop/extensions/podman/dist/extension.cjs:5173:3)
    at monitorProvider (/Users/eskimo/Sources/podman-desktop/extensions/podman/dist/extension.cjs:5173:3)
    at monitorProvider (/Users/eskimo/Sources/podman-desktop/extensions/podman/dist/extension.cjs:5173:3)
[...]

PR #6693, opened 2024-04-06
fix(extensions): make sure periodical full ui reload doesn't happen

GLEF1X:

### What does this PR do?

Attempts to resolve problems stated in #6663(full UI reload because the nodejs stack is overflowing)

#6663

### How to test this PR?

### DIFF
--- a/extensions/docker/src/extension.ts
+++ b/extensions/docker/src/extension.ts
@@ -30,11 +30,6 @@ let providerState: extensionApi.ProviderConnectionStatus = 'stopped';
 let containerProviderConnection: extensionApi.ContainerProviderConnection;
 let containerProviderConnectionDisposable: extensionApi.Disposable;
 
-async function timeout(time: number): Promise<void> {
-  return new Promise<void>(resolve => {
-    setTimeout(resolve, time);
-  });
-}
 async function isDockerDaemonAlive(socketPath: string): Promise<boolean> {
   const pingUrl = {
     path: '/_ping',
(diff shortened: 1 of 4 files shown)


extensions/docker/src/extension.ts
```diff
@@ -168,14 +167,7 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
   // monitor daemon
-  monitorDaemon(extensionContext).catch((err: unknown) => {
-    console.error('Error while monitoring docker daemon', err);
-    if (err instanceof Error) {
-      extensionApi.env.createTelemetryLogger().logError(err);
-    } else {
-      extensionApi.env.createTelemetryLogger().logError(err.toString());
-    }
-  });
+  void monitorDaemon(extensionContext);
```
benoitf, line 170: hello, here it looks you're bypassing the linter and do not catch error on promises, etc.

extensions/docker/src/extension.ts
```diff
@@ -89,24 +84,28 @@ async function isDisguisedPodman(socketPath: string): Promise<boolean> {
...
-    await timeout(5000);
-    monitorDaemon(extensionContext).catch((err: unknown) => {
-      console.error('Error while monitoring docker daemon', err);
-      if (err instanceof Error) {
-        extensionApi.env.createTelemetryLogger().logError(err);
-      } else {
-        extensionApi.env.createTelemetryLogger().logError(err.toString());
+function monitorDaemon(extensionContext: extensionApi.ExtensionContext): void {
+  const loop = (): NodeJS.Timeout =>
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
benoitf, line 89: we should not ignore the misused-promises in the project

extensions/podman/src/monitor.ts
```diff
@@ -0,0 +1,67 @@
...
+  ) {
+    this.stopLoopPredicate = options.stopMonitorPredicate;
+    this.intervalMs = options.intervalMs ?? DEFAULT_MONITOR_INTERVAL_MS;
+    this.name = options.name;
+  }
+
+  // Start monitoring by calling passed in constructor parameters function until stopLoopPredicate returns false
+  start(): void {
+    const loop = (): NodeJS.Timeout =>
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
benoitf, line 46: should not ignore

benoitf (commented):
thanks for the PR

FYI this PR being more on technical debt issue will be looked/discussed/merged after 1.10 release

extensions/podman/src/monitor.ts
dgolovin: Should be reused among extensions, because it is common pattern throughout:
* run callback on constant intervals, 
* do not fail on error in callback just report it or allow to configure error handling caback,
* provide predicate function for exit from monitoring loop

Monitoring loop can be implemented as (see [timers/promises example](https://nodejs.org/api/timers.html#timerspromisessetintervaldelay-value-options)

``` typescript
import { setInterval } from 'node:timers/promises';

for await (const _interval of setInterval(interval)) {
  try {
    await callback(extensionContext);
  } catch (error) {
    errorhandler?(error)
  }
  if (stopLoop)
    break;
}
```

dgolovin: or use npm package like 'set-interval-async' that would handle some other usecases

``` typescript
// Dynamic strategy.
import { setIntervalAsync, clearIntervalAsync } from 'set-interval-async/dynamic';

const timer = setIntervalAsync(async () => {
  // ...
  if (shouldStop) {
    await clearIntervalAsync(timer);
  }
  try {
     await callback()
  } catch(err) {
     handleError(err);
  }
}, interval);

```

extensions/docker/src/extension.ts
GLEF1X: `monitorDaemon` doesn't return a promise

benoitf:
closing as no response

PR #6707, opened 2024-04-08
chore: add isManifest function to determine a manifest

### LINKED ISSUE #6690: Add functionality to guess if an image is a manifest

As per https://github.com/containers/podman-desktop/issues/6676 discussion, we need to safely determine if an image is a manifest in order to implement it correctly.

We will "guess" if an image is a manifest until it is supported upstream by podman by:

- Determining if it is a small size (usually only a few kb)
- Labels are null
- Containers are 0 (unverified, but should check)
- No history

_No response_

### Additional context

_No response_

cdrage:
chore: add isManifest function to determine a manifest

### What does this PR do?

Adds a guessIsManifest function to try and determine if an image is a
manifest or not:

- By checking the repotags and digests
- Checking labels does not exist, as labels are not added when creating
  a manifest
- VirtualSize is under 1MB as a manifest list is just text that points
  to other images

We are unable to check the history (manifest commonly has no history) as
Dockerode API does not support getting the history of the image.

### Screenshot / video of UI

N/A

Closes  https://github.com/containers/podman-desktop/issues/6690

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/util/manifest.spec.ts
+++ b/packages/main/src/plugin/util/manifest.spec.ts
@@ -1,0 +1,120 @@
+/**********************************************************************
+ *
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { describe, expect, test } from 'vitest';
+
+import type { ImageInfo } from '../api/image-info.js';
+import { guessIsManifest } from './manifest.js';
+
+describe('guessIsManifest function', () => {
+  test('correctly identifies a manifest image', () => {
+    const manifestImage: ImageInfo = {
+      Id: 'manifestImage',
+      Labels: {},
+      engineId: 'engine1',
+      engineName: 'podman',
+      ParentId: '',
+      RepoTags: ['manifestTag'],
+      RepoDigests: ['manifestDigest'],
+      Created: 0,
+      Size: 0,
+      VirtualSize: 500000, // 500KB, less than the 1MB threshold
+      SharedSize: 0,
+      Containers: 0,
+    };
+
+    expect(guessIsManifest(manifestImage)).toBe(true);
+  });
+
+  test('returns false if VirtualSize is over 1MB', () => {
+    const largeImage: ImageInfo = {
+      Id: 'largeImage',
+      Labels: {},
+      engineId: 'engine2',
+      engineName: 'podman',
+      ParentId: '',
+      RepoTags: ['largeImageTag'],
+      RepoDigests: ['largeImageDigest'],
+      Created: 0,
+      Size: 0,
+      VirtualSize: 2000000, // 2MB
+      SharedSize: 0,
+      Containers: 0,
+    };
+
+    expect(guessIsManifest(largeImage)).toBe(false);
+  });
+
+  test('returns false if Labels is not empty', () => {
+    const labeledImage: ImageInfo = {
+      Id: 'labeledImage',
+      Labels: { key: 'value' },
+      engineId: 'engine3',
+      engineName: 'podman',
+      ParentId: '',
+      RepoTags: ['labeledImageTag'],
+      RepoDigests: ['labeledImageDigest'],
+      Created: 0,
+      Size: 0,
+      VirtualSize: 500000, // 500KB
+      SharedSize: 0,
+      Containers: 0,
+    };
+
+    expect(guessIsManifest(labeledImage)).toBe(false);
+  });
+
+  test('returns false if RepoTags is undefined or empty', () => {
+    const noTagImage: ImageInfo = {
+      Id: 'noTagImage',
+      Labels: {},
+      engineId: 'engine4',
+      engineName: 'podman',
+      ParentId: '',
+      RepoTags: [],
+      RepoDigests: ['noTagImageDigest'],
+      Created: 0,
+      Size: 0,
+      VirtualSize: 500000, // 500KB
+      SharedSize: 0,
+      Containers: 0,
+    };
+
+    expect(guessIsManifest(noTagImage)).toBe(false);
+  });
+
+  test('returns false if RepoDigests is undefined or empty', () => {
+    const noDigestImage: ImageInfo = {
+      Id: 'noDigestImage',
+      Labels: {},
+      engineId: 'engine5',
+      engineName: 'podman',
+      ParentId: '',
+      RepoTags: ['noDigestImageTag'],
+      RepoDigests: [],
+      Created: 0,
+      Size: 0,
+      VirtualSize: 500000, // 500KB
+      SharedSize: 0,
+      Containers: 0,
+    };
+
+    expect(guessIsManifest(noDigestImage)).toBe(false);
+  });
+});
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/util/manifest.ts
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ImageInfo } from '../api/image-info.js';
+
+const GUESSED_MANIFEST_SIZE = 1000000; // 1MB
+
+// Function to safely "guess" if the image is actually a manifest
```
benoitf, line 24: do you think a manifest could reach that 1MB limit ? looks like it's much more less than that

packages/main/src/plugin/util/manifest.ts
```diff
@@ -0,0 +1,47 @@
...
+// all of these conditions must be met to (safely) determine if the image is a manifest
+//
+// We will check this within ImageInfo
+export function guessIsManifest(image: ImageInfo): boolean {
+  return Boolean(
+    image.RepoTags &&
+      image.RepoDigests &&
+      image.RepoTags.length > 0 &&
+      image.RepoDigests.length > 0 &&
+      (!image.Labels || Object.keys(image.Labels).length === 0) &&
```
benoitf, line 47: Labels field seem to be `null` (not undefined or with empty size)

packages/main/src/plugin/util/manifest.ts
```diff
@@ -0,0 +1,47 @@
...
+      image.VirtualSize < GUESSED_MANIFEST_SIZE,
+  );
```
benoitf, line 51: it looks like manifest don't have the History field as well

packages/main/src/plugin/util/manifest.ts
cdrage: Probably not, maybe 50KB max?

packages/main/src/plugin/util/manifest.ts
cdrage: With our tests as well as through ImageInfo returning, it is sometimes just `{}` from the api

packages/main/src/plugin/util/manifest.ts
cdrage: Yes :( but we cannot get History as Dockerode API / our current implementation, does not return it. I think we could safely say it's a manifest with the very small image size / no labels.

benoitf: can you check with `docker.io/hello` image ?

benoitf: I mean, to be sure we don't flag hello image as a manifest

benoitf: >Yes :( but we cannot get History as Dockerode API / our current implementation, does not return it. I think we could safely say it's a manifest with the very small image size / no labels.

but we're calling libpod listImages now
so probably we can add History as an optional field

benoitf: if not, we should also check that engineType is podman

cdrage:
Switching to draft for two things:

- [x] Check that the engine is podman
- [x] Check history is blank

You are right @benoitf with `docker.io/hello-world` it would be considered a manifest due to the small size:

```
  {
    "Id": "ee301c921b8aadc002973b2e0c3da17d701dcd994b606769a7e6eaa100b81d44",
    "ParentId": "",
    "RepoTags": [
      "docker.io/library/hello-world:latest"
    ],
    "RepoDigests": [
      "docker.io/library/hello-world@sha256:2d4e459f4ecb5329407ae3e47cbc107a2fbace221354ca75960af4c047b3cb13",
      "docker.io/library/hello-world@sha256:53641cd209a4fecfc68e21a99871ce8c6920b2e7502df0a20671c6fccc73a7c6"
    ],
    "Created": 1683046167,
    "Size": 23301,
    "SharedSize": 0,
    "VirtualSize": 23301,
    "Labels": null,
    "Containers": 0,
    "Names": [
      "docker.io/library/hello-world:latest"
    ],
    "Digest": "sha256:53641cd209a4fecfc68e21a99871ce8c6920b2e7502df0a20671c6fccc73a7c6",
    "History": [
      "docker.io/library/hello-world:latest"
    ]
  }
```

So we have to make sure history is blank / does not exist as well.

cdrage:
Updated and ready for review.

Requires #6713 before merging.

- Switched to 50KB max for the manifest size
- Updated tests
- Check for null / empty history
- Check engine name matches to 'podman'

packages/main/src/plugin/util/manifest.ts
benoitf: can you add a test with like 'real data' coming from the docker hello world for example

cdrage:
TODO:
- [x] Add test for the following output and making sure it's "not a manifest":

```json
  {
    "Id": "ee301c921b8aadc002973b2e0c3da17d701dcd994b606769a7e6eaa100b81d44",
    "ParentId": "",
    "RepoTags": [
      "docker.io/library/hello-world:latest"
    ],
    "RepoDigests": [
      "docker.io/library/hello-world@sha256:2d4e459f4ecb5329407ae3e47cbc107a2fbace221354ca75960af4c047b3cb13",
      "docker.io/library/hello-world@sha256:53641cd209a4fecfc68e21a99871ce8c6920b2e7502df0a20671c6fccc73a7c6"
    ],
    "Created": 1683046167,
    "Size": 23301,
    "SharedSize": 0,
    "VirtualSize": 23301,
    "Labels": null,
    "Containers": 0,
    "Names": [
      "docker.io/library/hello-world:latest"
    ],
    "Digest": "sha256:53641cd209a4fecfc68e21a99871ce8c6920b2e7502df0a20671c6fccc73a7c6",
    "History": [
      "docker.io/library/hello-world:latest"
    ]
  }
  ```

**Comment by cdrage** [comment]
Ready for another review, test has been added 💯

File: packages/main/src/plugin/util/manifest.spec.ts [hunk 4]
```diff
@@ -0,0 +1,144 @@
+/**********************************************************************
+ *
```
benoitf: it looks like this line is not there on other files

packages/main/src/plugin/util/manifest.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      Size: 0,
+      VirtualSize: 500000, // 500KB
+      SharedSize: 0,
+      Containers: 0,
+    };
+
+    expect(guessIsManifest(noDigestImage)).toBe(false);
+  });
+
+  test('return false for a typical helloworld image example, that may be small enough to trigger guessIsManifest as true', () => {
```
benoitf, line 120: thanks

packages/main/src/plugin/util/manifest.ts
```diff
@@ -0,0 +1,52 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ImageInfo } from '../api/image-info.js';
+
+// 50 KB
```
benoitf: looks like images with tons of arches like docker.io has only 5kb manifest so size should be good

just a small comment, usually the comment will be sooner or later out of sync with the value

so maybe, use value instead of 51200 and as comment 50 KB

something like
```
const KB = 1000;
const GUESSED_MANIFEST_SIZE = 50 * KB;
```

benoitf (approved):

packages/main/src/plugin/util/manifest.spec.ts
cdrage: good catch!

packages/main/src/plugin/util/manifest.ts
cdrage: good call! i've updated it!

benoitf:
still approving

packages/main/src/plugin/util/manifest.ts
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ImageInfo } from '../api/image-info.js';
+
+const KB = 1024;
+const GUESSED_MANIFEST_SIZE = 50 * KB; // 50 KB
```
benoitf:
```suggestion
const GUESSED_MANIFEST_SIZE = 50 * KB;
```

PR #6717, opened 2024-04-08
chore: add isManifest to podmanListImages

### LINKED ISSUE #6691: Add isManifest boolean to ImageInfo / list images API.

Depends on #6690

For the image list API, ImageInfo struct, we should be returning isManifest = true/false in order to determine if an image is actually a manifest.

_No response_

### Additional context

_No response_

cdrage:
chore: add isManifest to podmanListImages

### What does this PR do?

* Adds the ability to see what is and isn't a manifest based upon
  guessIsManifest function for podmanListImages, allowing you to use the
  internal function call to determine what's a manifest or not
* Added tests to cover the functionality

### Screenshot / video of UI

N/A

Closes https://github.com/containers/podman-desktop/issues/6691

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -75,6 +75,7 @@ import { Emitter } from './events/emitter.js';
 import type { ImageRegistry } from './image-registry.js';
 import type { Telemetry } from './telemetry/telemetry.js';
 import { Disposable } from './types/disposable.js';
+import { guessIsManifest } from './util/manifest.js';
 
 const tar: { pack: (dir: string) => NodeJS.ReadableStream } = require('tar-fs');
 
@@ -626,6 +627,10 @@ export class ContainerProviderRegistry {
           ...image,
           engineName: provider.name,
           engineId: provider.id,
+          // Using guessIsManifest, determine if the image is a manifest and set isManifest accordingly
+          // NOTE: This is a workaround until we have a better way to determine if an image is a manifest
+          // and may result in false positives until issue: https://github.com/containers/podman/issues/22184 is resolved
+          isManifest: guessIsManifest(image),
         }));
       }),
     );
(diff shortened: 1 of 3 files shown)


cdrage:
Requires https://github.com/containers/podman-desktop/pull/6707 merged before merging this one

packages/main/src/plugin/util/manifest.spec.ts
```diff
@@ -0,0 +1,144 @@
+/**********************************************************************
+ *
```
benoitf:
```suggestion
```

packages/main/src/plugin/container-registry.ts
```diff
@@ -621,6 +622,13 @@ export class ContainerProviderRegistry {
           return fetchedImages;
         }
 
+        // Using guessIsManifest, determine if the image is a manifest and set isManifest accordingly
+        // NOTE: This is a workaround until we have a better way to determine if an image is a manifest
+        // and may result in false positives until issue: https://github.com/containers/podman/issues/22184 is resolved
+        fetchedImages.forEach(image => {
+          image.isManifest = guessIsManifest(image);
+        });
+
```
benoitf: maybe get rid of this forEach and move the lines we have after this block as we iterate again on all the images

add only isManifest field by doing something like

```ts
        // Transform fetched images to include engine name and ID
        return fetchedImages.map(image => ({
          ...image,
          engineName: provider.name,
          engineId: provider.id,
        // Using guessIsManifest, determine if the image is a manifest and set isManifest accordingly
        // NOTE: This is a workaround until we have a better way to determine if an image is a manifest
        // and may result in false positives until issue: https://github.com/containers/podman/issues/22184 is resolved
         isManifest: guessIsManifest(image);
        }));

```

cdrage: You are completely right, it should be in here, 🤦

cdrage:
Had to redo the commits / messed up rebasing, so now it's back to 1 commit again, ready for review

benoitf (approved):

Issue #6727, opened 2024-04-09
Implement a playwright page object model to accomodate authentication extensions UI assets


odockal:
### Is your enhancement related to a problem? Please describe

We have missing some POM for e2e tests that would cover working with authentication providers.

### Describe the solution you'd like

Implement missing and update actual functionality of Pom to accommodate work with authentication providers objects.

PR #6728, opened 2024-04-09
chore(tests): add playwright pom covering authentication providers

odockal:
### What does this PR do?
It covers adding and updating some POM for e2e tests to cover authentication providers and extensions.
#6727 

### How to test this PR?
Runnnig `yarn test:e2e` should still works without any failure.

- N/A Tests are covering the bug fix or the new feature

### DIFF
--- a/tests/playwright/src/model/pages/authentication-page.ts
+++ b/tests/playwright/src/model/pages/authentication-page.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/tests/playwright/src/model/pages/authentication-page.ts
+++ b/tests/playwright/src/model/pages/authentication-page.ts
@@ -1,0 +1,42 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


tests/playwright/src/model/pages/authentication-page.ts
```diff
@@ -0,0 +1,42 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

tests/playwright/src/model/workbench/navigation.ts
```diff
@@ -70,8 +70,11 @@ export class NavigationBar {
   }
 
   async openSettings(): Promise<SettingsBar> {
-    await this.settingsLink.waitFor({ state: 'visible', timeout: 3000 });
-    await this.settingsLink.click({ timeout: 5000 });
-    return new SettingsBar(this.page);
+    const settingsBar = new SettingsBar(this.page);
+    if (!(await settingsBar.settingsNavBar.isVisible())) {
+      await this.settingsLink.waitFor({ state: 'visible', timeout: 3000 });
```
cbr7: We shouldn't use .waitFor() in such cases, if something is needed to be awaited for visibility it's recommended to use `await except(locator).toBeVisible()` however in this case the since the item is clickable it might be better to use .toBeEnabled().

In fact I would even remove the if condition completely and have the `await expect(locator).toBeEnabled()` in the higher scope.

odockal: Problem is the behavior of the link. If it is visible, do not click it, since it will disapper. So the visibility check is fine. We do not want to click on the element itself, but if it is visible, we do not want to click on other one. 
Is `waitFor` used in `locator.isVisible()` method?

cbr7: No, waitFor is not called as part of isVisible() method.

Even if you want to keep the if condition, I would still change the waitFor to `await expect(locator).tobeVisible()` as far as I can tell it fits your use case perfectly.

odockal: Ah, the waitFor that was already part of code. got it. Will change.

cbr7 (approved):

benoitf (approved):

PR #6730, opened 2024-04-09
feat(UI): adding optional title property to featured extension

### LINKED ISSUE #6729: adding title properties to FeaturedExtension
As mention in https://github.com/containers/podman-desktop/pull/6708#issuecomment-2045124331 we need to have a text, or title inside the feature extension component.

To archive this behaivour, and make it a minimum reusable and flexible, we could introduce a `title` property to the `<FeaturedExtension/>` component, that would be used for display

axel7083:
### What does this PR do?

Required for https://github.com/containers/podman-desktop/issues/6215

### Screenshot / video of UI

(E.g. defining the title property to `EXTENSION`).

image

Fixes https://github.com/containers/podman-desktop/issues/6729

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/featured/FeaturedExtension.svelte
+++ b/packages/renderer/src/lib/featured/FeaturedExtension.svelte
@@ -7,6 +7,7 @@ import FeaturedExtensionDownload from './FeaturedExtensionDownload.svelte';
 
 export let featuredExtension: FeaturedExtension;
 export let variant: 'primary' | 'secondary' = 'primary';
+export let title: string | undefined = undefined;
 </script>
 
 <div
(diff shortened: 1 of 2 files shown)


benoitf:
I have the feeling that the `EXTENSION` text is not aligned with the icon below

as in the mockup design
image

benoitf (approved):
just need confirmation of @ekidneyrh for alignment/size of the `EXTENSION` text
code and tests are 👍

axel7083:
> just need confirmation of @ekidneyrh for alignment/size of the `EXTENSION` text code and tests are 👍

I am not sure to understand what you meant about the alignement. It is aligned on the left

image

benoitf:
yes looks like I'm wrong for the alignment 👍 

for the text it looks bigger than the mockup

axel7083:
> yes looks like I'm wrong for the alignment 👍
> 
> for the text it looks bigger than the mockup

I used `text-xs` if I need to make it smaller I need to set custom value.

ekidneyrh:
Would this just be for the banner that the extension is labelled?

packages/renderer/src/lib/featured/FeaturedExtension.svelte
```diff
@@ -7,6 +7,7 @@ import FeaturedExtensionDownload from './FeaturedExtensionDownload.svelte';
 
 export let featuredExtension: FeaturedExtension;
 export let variant: 'primary' | 'secondary' = 'primary';
+export let title: string | undefined = undefined;
```
benoitf: AFAIK the title should be provided by this component, not by the client

so maybe displayTitle: boolean = false

axel7083: Fixed, I changed to an hardcoded text hidden by default

axel7083:
> Would this just be for the banner that the extension is labelled?

This would be for the extension banner yes

benoitf (approved):
thx

Issue #6774, opened 2024-04-12
Global navigation: Handle too many items


benoitf:
### Is your enhancement related to a problem? Please describe

the navbar contains too many items for the default min-height that we have

if there are too many items to be displayed, buddy icon and settings icon should always be there but we should hide the previous icons with a `...` and with either our custom dropdown (like Slack for example) or a menu (like VS Code). I would opt-in for dropdown but it looks like there was a problem when wanting to use it for the buddy dropdown

To mitigate the issue we could reduce the vertical gap in each icon of the navbar from py-3 to py-2

### Describe the solution you'd like

quick & dirty: reduce py-3 to py-2, reduce icon size
proper handle: hide top items and display ... / more options while bottom icons should always be visible (never hidden)

deboer-tim:
Since it's an IDE I typically use a near full-screen window for VS Code, and it has 8 nav items without any extensions.
I typically use Podman Desktop at 60% of that size, and we'll now have 11 in the default configuration. Without changes the first installed extension would be visible, the second would be hidden.

We should absolutely do the ..., py-2, etc., and that will squeeze a couple more in and make it work better for now, but IMHO our core problem is that we have too many items in the nav and keep adding more. It looks cluttered, it's time to remove a few options or look at other options.

benoitf:
commented there https://github.com/containers/podman-desktop/pull/6777#pullrequestreview-2006945794 that we could also reduce the width of the navbar

ekidneyrh:
Just adding my thoughts here ahead of the UX call today. 

Since we have modelled the nav bar from VS code, we can have a look and see what they are doing: 

![Screenshot from 2024-06-20 12-03-33](https://github.com/containers/podman-desktop/assets/87311656/d58d8b40-24b2-4b2b-a6f9-e8684de9dad3)

You can right-click an empty space on the nav bar and then select what icons you want pinned in the nav bar. If right-clicked over an icon, an additional option to ‘unpin’ will appear.

![Screenshot from 2024-06-20 13-57-50](https://github.com/containers/podman-desktop/assets/87311656/f321c1f9-6ecf-4911-8afa-e74a94e3fa3c)


Benefit of us doing the same: developers using podman will notice the similarity in VS Code and will know to right-click to do this. 

Downsides would be users that are not familiar with this layout might not know how. We can work around this by also having the option to pin/unpin to the nav bar through the extensions tab. 
	
[...]

PR #6777, opened 2024-04-13
feat: reduce height of navbar items

benoitf:
### What does this PR do?
before having a solution for #6774 with a design, reduce height of each item from navbar

### Screenshot / video of UI

on the left, current state
on the right, PR state

image

related to https://github.com/containers/podman-desktop/issues/6774

### How to test this PR?

### DIFF
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -62,6 +62,8 @@ let ingressesRoutesCount = '';
 const imageUtils = new ImageUtils();
 export let exitSettingsCallback: () => void;
 
+const iconSize = '22';
+
 onMount(async () => {
   const commandRegistry = new CommandRegistry();
   commandRegistry.init();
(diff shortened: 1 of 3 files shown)


deboer-tim:
I tried playing with this PR. IMHO dropping to py-2.5 looks fine and even going to py-2 looks ok in the sections, but going to 1.5 here looks squished and kind of ugly to me.

It helps a bit if we make the entire bar a little narrower at the same time so that the selection isn't so rectangular... then maybe shrink the icons a little too, but those are both bigger changes that we might not want to make for the long term.

FYI to @ekidneyrh

benoitf:
I agree I was finding it was ok with 2 but went further with 1.5 to get feedback :-)

deboer-tim:
😆

axel7083:
I am wondering if we should not reduce a little bit the size of the icons, as I have the feeling there is too much things in a small space with the new spacing...

| size 24 | Size 22 | Size 20 |
| --- | --- | --- |
| image | image | image |

jeffmaury:
Icon resizing would improve

benoitf:
changed from py-3 to py-2 and reduced the icon size from 24 to 22

deboer-tim (approved):
I think maybe we should shrink the width of the bar too, but that can be a separate discussion.

Issue #6529, opened 2024-03-25
Manifest feature: List manifests on image list screen


cdrage:
### Is your feature request related to a problem? Please describe

When you create a manifest on the CLI:

```
podman manifest create test123
```

It appears on `podman image ls`:

```
▶ podman image ls
REPOSITORY                                TAG                IMAGE ID      CREATED       SIZE
localhost/test123                         latest             8b9101797ef4  2 days ago    1.08 kB
localhost/foobar                          latest             409f1927cb03  3 days ago    112 B
quay.io/podai/test123                     latest             e721af43292a  3 days ago    1.64 GB
```

However, it does not appear within the Podman Desktop UI.

### Describe the solution you'd like

Match the podman CLI behaviour by showing the manifest on the image list.

PR #6792, opened 2024-04-15
feat: add multi-arch building / manifest creation

### LINKED ISSUE #6626: in buildImage, spawn multiple builds if multi-arch

today we delegate to the build endpoint but in case of a comma separated list of arch , the call is failing

If multiple arch 'comma separated' specified, need to create a manifest using
- [x] https://github.com/containers/podman-desktop/issues/6624

then spin up one call to the build endpoint for each arch

once we have the result,

add each image to the manifest calling
- [x] https://github.com/containers/podman-desktop/issues/6625

_No response_

### Additional context

_No response_

cdrage:
feat: add multi-arch building / manifest creation

### What does this PR do?

* When doing a multi-arch / platform selection on the build image page,
  it will now create a singular manifest
* Tells the user that they are creating a manifest
* Tests cover new feature

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/6422176/73759545-62cb-4441-8cd5-5f547d74f44b

Closes https://github.com/containers/podman-desktop/issues/6626

### How to test this PR?

1. Select multiple platforms when building a test image (any image will
   do)
2. Do `podman image ls` afterwards and you should see your image (it's a
   manifest)
3. Examine the confirm it is a manifest by doing `podman manifest
   inspect <image-name>

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -1086,7 +1086,6 @@ export class PluginSystem {
           cancellationTokenRegistry,
           cancellableTokenId,
         );
-
         return containerProviderRegistry.buildImage(
           containerBuildContextDirectory,
           (eventName: string, data: string) => {
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -38,8 +38,19 @@ let providers: ProviderInfo[] = [];
 let providerConnections: ProviderContainerConnectionInfo[] = [];
 let selectedProvider: ProviderContainerConnectionInfo | undefined = undefined;
 let logsTerminal: Terminal;
+let buildIDs = [];
 
+$: platforms = containerBuildPlatform ? containerBuildPlatform.split(',') : [];
 $: hasInvalidFields = !containerFilePath || !containerBuildContextDirectory;
+
+interface BuildOutputItem {
+  stream?: string;
+  aux?: {
+    ID: string;
+  };
+}
+
+type BuildOutput = BuildOutputItem[];
 
 function getTerminalCallback(): BuildImageCallback {
   return {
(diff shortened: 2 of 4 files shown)


cdrage:
Note: Keep in mind that this does not appear in the PD UI image list yet until we have manifests implemented / being shown #6529 

This PR is required for user-flow related tasks related to other another extension.

cdrage:
Figuring out strange failure in windows tests, otherwise this code is ready for review 🎉 

@axel7083 @benoitf @dgolovin

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -215,6 +314,9 @@ async function abortBuild() {
 
         <div hidden="{buildImageInfo?.buildRunning}">
           <label for="containerBuildPlatform" class="block mb-2 text-sm font-bold text-gray-400">Platform</label>
+          {#if platforms.length > 1}
+            <p class="text-sm text-gray-600 mb-2">Multiple platforms selected, a manifest will be created</p>
```
axel7083, line 318: I feel like a new user would not know what is a manifest here. Maybe we could add a `fa-circle-info` with a tooltip, giving a very short explanation on what is manifest

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -88,6 +114,79 @@ async function buildContainerImage(): Promise<void> {
...
+  // Collection of build IDs, this is needed for being able to create the manifest
+  // as we need to know either the IDs or the names of the images that were built
+  buildIDs = [];
+
+  if (containerFilePath && selectedProvider) {
+    // Extract the relative path from the containerFilePath and containerBuildContextDirectory
+    const relativeContainerfilePath = containerFilePath.substring(containerBuildContextDirectory.length + 1);
+
+    // We'll iterate over each platform and build the image
+    for (const platform of platforms) {
```
axel7083, line 131: I am wondering if we should not have this logic in the main, rather than the frontend ? it is a process with a lot of steps, and feel like it should not be placed inside svelte component

cdrage: yeah, you are right, this should be instead in the main code. i'm going to mark this as draft and move the code there instead.

cdrage: @axel7083 changing my opinion, sorry.

After going through the code, I think it's good that keep this in the svelte file / user facing due to our extensive use of streaming data back for buildImage and having multiple builds / creating manifest after would create a large amount of refactoring if the code was more in main rather than user facing.

Because of my limited bandwidth and the 13 other issues i've opened / fixed to get this feature added, I won't be able to implement a more thorough implementation in time for next week's release, so we may have to do this PR temporary (to fix the non-functioning building) until we can do a larger refactor IMHO.

This PR can be a temporary fix until we have a more permanent one / have a function to do multiple builds + create manifest.

I've opened up issue: https://github.com/containers/podman-desktop/issues/6815

cdrage:
See: https://github.com/containers/podman-desktop/issues/6815 I have created this issue to track implementing this more in the "main" code of PD.

Due to time constraints, this would have to be user-facing for now to fix the currently broken feature.

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
axel7083: I understand the bandwidth issue, and the PR LGTM so I will approve it

axel7083 (approved):
Woking fine on Windows. LGTM 🚀

cdrage:
Update: Currently fixing tests

packages/main/src/plugin/index.ts
```diff
@@ -1087,6 +1087,16 @@ export class PluginSystem {
           cancellableTokenId,
         );
 
+        console.log(
```
axel7083: to remove ?

cdrage: haha yes, good catch, that was for debugging

benoitf (approved):
I agree that it should be done on the main side, each client shouldn't need to handle that

need to work on the created issue soon

Issue #2558, opened 2023-05-19
Windows podman machine creation does not respect number of CPUs choosen


odockal:
### Bug description

On windows, I am creating new podman machine and choose ie. 4 CPUs. When the machine is created there is 8 CPUs.
Same result can be observed when I check the command line `podman mahcine ls`.

When I try to create a machine on cli, the result is the same, so I would sum this up to be upstream issue. I have not found this issue here, so I will fill it in anyway.

 

### Operating system

Windows 10

### Version

1.0.0

jeffmaury:
I think it's related to WSL. Can you try from the CLI

odockal:
The same. manually initiating a machine with random number of cpu and starting it ends up with maximum number of cpus dedicated for the machine. So it is upstream issue.

cbr7:
Issue still present in Podman Desktop 1.8 with podman 4.9.3, the podman machine is not created with the indicated values and the values differ when checked in Podman Desktop and podman CLI.

![image (1)](https://github.com/containers/podman-desktop/assets/106525396/5bd0463b-870b-4dd8-96e1-8872435b200a)

The podman machine from the image was supposed to be created with 2 CPUs, 5 GB of RAM and 100 GB disk space.

odockal:
I was not able to reproduce with podman v5 (using cli) and hyperv, which seems to be the difference.

odockal:
Seems like upstream issue (https://github.com/containers/podman/issues/19443) was closed.

odockal:
After reading some man pages for podman-machine-set, I think that podman desktop should be able to notify uset about the fact, that adjusting any of CPU/DiskSize/Memory wont work in podman desktop Create New Machine. 
We could either disable the buttons, or print out a warning in the UI and provide an explanation or a link.

jeffmaury:
CPU, memory and disk size are not relevant for WSL, they should not be available for the user

PR #6810, opened 2024-04-16
feat: recognize WSL and HyperV

lstocchi:
### What does this PR do?

It adds a couple of functions to detect if the user is on WSL or HyperV. This will be in handy in future if someone wants to show some properties/settings based on the virtualization used on Windows. This allows to use these properties on the when clauses or from external extensions
E.g - do not show the cpu slider if the user is on wsl ->` "when": "!isWSL"`

### Screenshot / video of UI

N/A

N/A -> needed for #2558 

### How to test this PR?

run tests

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -3715,6 +3715,16 @@ declare module '@podman-desktop/api' {
      * The system clipboard.
      */
     export const clipboard: Clipboard;
+
+    /**
+     * Indicates if the user is using WSL to run podman machines
+     */
+    export function isWSL(): Promise<boolean>;
+
+    /**
+     * Indicates if the user is using HyperV to run podman machines
+     */
+    export function isHyperV(): Promise<boolean>;
   }
 
   /**
--- a/packages/main/src/util.ts
+++ b/packages/main/src/util.ts
@@ -18,12 +18,42 @@
 
 import * as fs from 'node:fs';
 import * as os from 'node:os';
+import path from 'node:path';
 
 import { BrowserWindow } from 'electron';
 
-const windows = os.platform() === 'win32';
 export function isWindows(): boolean {
-  return windows;
+  return os.platform() === 'win32';
+}
+export async function isWSL(): Promise<boolean> {
+  const hyperv = await isHyperV();
+  return isWindows() && !hyperv;
+}
+export async function isHyperV(): Promise<boolean> {
+  if (!isWindows()) {
+    return false;
+  }
+  // check if the env variable is set with hyperv
+  if (process.env.CONTAINERS_MACHINE_PROVIDER === 'hyperv') {
+    return true;
+  }
+
+  // as a final step we check if the containers.conf file set the provider to hyperv
+  const providerRegex = /provider\s*=\s*"hyperv"/;
+  return isValueInContainersConfig(providerRegex);
+}
+export async function isValueInContainersConfig(regex: RegExp): Promise<boolean> {
+  const containersConfPath = getContainersConfPath();
+  try {
+    const containerConf = await fs.promises.readFile(containersConfPath, 'utf-8');
+    return regex.test(containerConf);
+  } catch (e) {
+    console.log(String(e));
+  }
+  return false;
+}
+export function getContainersConfPath(): string {
+  return path.resolve(os.homedir(), '.config/containers/containers.conf');
 }
 const mac = os.platform() === 'darwin';
 export function isMac(): boolean {
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -32,7 +32,7 @@ import type { NavigationManager } from '/@/plugin/navigation/navigation-manager.
 import type { WebviewRegistry } from '/@/plugin/webview/webview-registry.js';
 
 import { securityRestrictionCurrentHandler } from '../security-restrictions-handler.js';
-import { getBase64Image, isLinux, isMac, isWindows } from '../util.js';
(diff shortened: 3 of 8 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -3715,6 +3715,16 @@ declare module '@podman-desktop/api' {
...
+
+    /**
+     * Indicates if the user is using WSL to run podman machines
+     */
+    export function isWSL(): Promise<boolean>;
+
+    /**
+     * Indicates if the user is using HyperV to run podman machines
+     */
+    export function isHyperV(): Promise<boolean>;
```
axel7083, line 3727: I am wondering as WSL and HyperV are podman providers, we should not have a more generic method like

```ts
    /**
     * @return the hypervisor used (E.g. qemu, applehv, hyperv, wsl)
     */
     export function getPodmanMachineHypervisor(machineId?: string): Promise<string>
```

lstocchi: I thought the same at first but then i thought that starting from podman 5 we only support applehv on mac so isMac is good enough for macOs and this generic method would only return windows stuff.

axel7083 (commented):
I think there is two element that are getting mixte up here. The configuration for future machine, and the existing machine(s) themself.

If I want to create a Podman Machine, I should look at the `CONTAINERS_MACHINE_PROVIDER` as it used when the `--provider` argument is not provided when created a podman machine.

However, I might want to know, if my existing podman machine(s) is using WSL  hypervisor or HyperV, or AppleHv etc. Which would require to look at their config file.

lstocchi:
> I think there is two element that are getting mixte up here. The configuration for future machine, and the existing machine(s) themself.
> 
> If I want to create a Podman Machine, I should look at the `CONTAINERS_MACHINE_PROVIDER` as it used when the `--provider` argument is not provided when created a podman machine.
> 
> However, I might want to know, if my existing podman machine(s) is using WSL hypervisor or HyperV, or AppleHv etc. Which would require to look at their config file.

I don't get it as this PR do not check the config of any existing machine.

This is an extract about creating a hyperV machine, so checking the env is not enough.
"Podman machine now supports HyperV as a provider on Windows. This option can be set via the CONTAINERS_MACHINE_PROVIDER environment variable, or via containers.conf."

Here i'm just checking the current provider so it will be used for future machines and for listing current ones. If you are using hyperV you don't see any WSL one, so i think we can still use the property in any case, even for existing machine.

packages/extension-api/src/extension-api.d.ts
axel7083: Even tho on MacOS we have in v5 a single value, in v4 it is not the case, and in the future, we might have others (we never know).

For example, Podman still uses a switch cases, even if they have only one option https://github.com/containers/podman/blob/9697a5f08be544631c3b1b9889b906237e133c35/pkg/machine/provider/platform.go#L31-L35 they might add another later. For the GPU they were talking about creating a custom provider.

axel7083:
> I don't get it as this PR do not check the config of any existing machine.

That's maybe a confusion on my side, I though I could have multiple machine, like one on WSL, and one on HyperV. Might be related to https://github.com/containers/podman-desktop/issues/279

> Here i'm just checking the current provider so it will be used for future machines and for listing current ones. If you are using hyperV you don't see any WSL one, so i think we can still use the property in any case, even for existing machine.

Okey that's more clear thanks for the explanation

packages/extension-api/src/extension-api.d.ts
benoitf, line 3727: this can't be provided like this in env

env is generic, you can have like no podman or no podman machine running at all

benoitf (changes requested):

packages/extension-api/src/extension-api.d.ts
lstocchi: So should this be just part of the podman extension?

benoitf: yes it should 'leak' into the extension-api.

we could later expose an API provided only by Podman extension (and other extension calling that API) if some specifics are required to be used by other extensions.

this kind of settings could be provided as extra addon into connection object.

also AFAIK while it's still not currently well supported, podman could allow different machine type at the same type, so it can't be a general flag, it's attached to a machine

packages/main/src/util.ts
```diff
@@ -18,12 +18,42 @@
...
+    return false;
+  }
+  // check if the env variable is set with hyperv
+  if (process.env.CONTAINERS_MACHINE_PROVIDER === 'hyperv') {
+    return true;
+  }
+
+  // as a final step we check if the containers.conf file set the provider to hyperv
+  const providerRegex = /provider\s*=\s*"hyperv"/;
+  return isValueInContainersConfig(providerRegex);
```
benoitf, line 43: podman desktop core should not read directly some files about machines, it should be the podman extension.

Then, about the flag, podman extension should reuse `podman machine list --format=json` command that is periodically run, here there is `VMType` information about the podman machine type being used.

lstocchi: This is not related to a specific podman machine. It is the general containers config file, and if you set the provider, it will be used as an indication by podman to create a new machine.

If i have to hide/show some properties based on the fact that the user is gonna create a WSL or HyperV instance, i don't have to know which VMtypes are the old machines.

lstocchi: BTW i'm gonna move this with the rest of the code to podman extension. Just to clarify

benoitf: for creation of podman machines, toggle like WSL / HyperV or default should be managed by podman extension with the properties on the ConnectionFactory and with enablement/when clause

lstocchi:
Closing as i'm going to merge this on the other PR

PR #6812, opened 2024-04-16
feat: add inspectManifest API endpoint

### LINKED ISSUE #6793: Add inspectManifest API endpoint

We need to add the inspectManifest API endpoint in order to get more information of the manifest including the digest ID.

Add inspectManifest to the podman desktop API

### Additional context

Required for https://github.com/containers/podman-desktop-extension-bootc/issues/324

cdrage:
feat: add inspectManifest API endpoint

### What does this PR do?

* Adds the inspectManifest API endpoint so we can retrieve information
  regarding a manifest from the podman libpodApi endpoint

### Screenshot / video of UI

N/A, it's an API call.

Closes https://github.com/containers/podman-desktop/issues/6793

### How to test this PR?

Tests cover, otherwise you could also put the following (after creating
ANY manifest, in a svelte file:)

```ts
  // List all the images
  const images = await window.listImages();

  // Get all the ones that have isManifest set to true
  const manifestImages = images.filter(image => image.isManifest);

  // For each manifestImages use inspectManifest to get the information (passing in engineId as well)
  const imageInfoPromises = manifestImages.map(image => window.inspectManifest(image.engineId, image.Id));

  // Wait for all the promises to resolve
  const imageInfos = await Promise.all(imageInfoPromises);

  // Consoel log each one
  imageInfos.forEach(imageInfo => {
    console.log(imageInfo);
  });
```

### DIFF
--- a/packages/main/src/plugin/api/manifest-info.ts
+++ b/packages/main/src/plugin/api/manifest-info.ts
@@ -27,3 +27,24 @@ export interface ManifestCreateOptions {
   // Provider to use for the manifest creation, if not, we will try to select the first one available (similar to podCreate)
   provider?: ContainerProviderConnection;
 }
+
+export interface ManifestInspectInfo {
+  engineId: string;
+  engineName: string;
+  manifests: {
+    digest: string;
+    mediaType: string;
+    platform: {
+      architecture: string;
+      features: string[];
+      os: string;
+      osFeatures: string[];
+      osVersion: string;
+      variant: string;
+    };
+    size: number;
+    urls: string[];
+  }[];
+  mediaType: string;
+  schemaVersion: number;
+}
--- a/packages/main/src/plugin/api/manifest-info.ts
+++ b/packages/main/src/plugin/api/manifest-info.ts
@@ -36,14 +36,12 @@ export interface ManifestInspectInfo {
     mediaType: string;
     platform: {
       architecture: string;
-      features: string[];
+      features?: string[];
       os: string;
-      osFeatures: string[];
-      osVersion: string;
-      variant: string;
+      variant?: string;
     };
     size: number;
-    urls: string[];
+    urls?: string[];
   }[];
   mediaType: string;
   schemaVersion: number;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -302,14 +302,12 @@ declare module '@podman-desktop/api' {
       mediaType: string;
       platform: {
         architecture: string;
-        features: string[];
+        features?: string[];
         os: string;
-        osFeatures: string[];
-        osVersion: string;
-        variant: string;
+        variant?: string;
       };
       size: number;
-      urls: string[];
+      urls?: string[];
     }[];
     mediaType: string;
     schemaVersion: number;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -294,6 +294,26 @@ declare module '@podman-desktop/api' {
     // Provider to use for the manifest creation, if not, we will try to select the first one available (similar to podCreate)
     provider?: ContainerProviderConnection;
   }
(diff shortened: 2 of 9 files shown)


lstocchi (approved):
Codewise LGTM

packages/main/src/plugin/api/manifest-info.ts
```diff
@@ -27,3 +27,24 @@ export interface ManifestCreateOptions {
...
+  manifests: {
+    digest: string;
+    mediaType: string;
+    platform: {
+      architecture: string;
+      features: string[];
+      os: string;
+      osFeatures: string[];
+      osVersion: string;
+      variant: string;
```
benoitf: are these params mandatory ? AFAIK some of them are optional no ? like OSFeatures

benoitf: It looks it's not as the other Types returned with the API where it is starting by uppercase (like Architecture, OS, Manifests, etc)

cdrage: that's correct, unlike the other API calls throughout the podman API, manifests are all lowercase, see: https://docs.podman.io/en/latest/_static/api.html#tag/manifests/operation/ManifestInspectLibpod

vs something else like inspect pod: https://docs.podman.io/en/latest/_static/api.html#tag/pods/operation/PodInspectLibpod

cdrage: it is just returning information, so they are not mandatory, but no need to set them to optional as it is just there so the API can pick it up.

benoitf: I tried with `docker.io/httpd` image

```
curl --unix-socket /var/run/docker.sock  "http://libpod/v3.0.0/libpod/manifests/docker.io%2Fhttpd/json" | jq .
```

the fields are not there so it should be listed as optional.
                                                                                                    
```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 2095,
      "digest": "sha256:6834aeaa0023c674c53a8891ce77a363579ae9a9679222f9634f95aa72c5781b",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 841,
      "digest": "sha256:d8081c32306039fe31141c53642d73c59678f70678e85f7506c01d15aa5a95e3",
      "platform": {
        "architecture": "unknown",
        "os": "unknown"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 2097,
      "digest": "sha256:623ea480a45a037db310321eb5f9b933df664701f5be02feb2164b302b2e293d",
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v5"
      }
    },
```

benoitf: on this link it's listed as:

"os.features": ["string"],
"os.version": "string",

but here you introduced `osFeatures` and `osVersion`

cdrage:
@benoitf Thanks for the review, I'm going to put this back on draft as I have missed a few things

benoitf:
@cdrage also maybe podman api doc is incorrect and it's not `os.features` but `osFeatures`

cdrage:
Ready for review again @benoitf @lstocchi @feloy 

For @benoitf : I have removed the osFeatures and osVersion as they are rarely used as per: https://github.com/containers/podman/blob/01d0f88a100ca91cd7a60c691dcbba2ab29903dd/docs/source/markdown/podman-manifest-annotate.1.md.in#L38 as well as the fact that they have an uncommon json output (`os.features` instead of `osFeatures`) which doesn't match the rest of the API, so it's safe to say they are not used very often.

If we get a feature request, we should add them in.

Updated the PR & tests.

benoitf (approved):

PR #6826, opened 2024-04-18
chore(test): add aria labels for rows

### LINKED ISSUE #5639: Add ARIA functionality to image/containers/pods table rows

It's very difficult to handle image/container/pod table rows in their respective pages because there aren't any rows specific ARIA locators available.

Add ARIA functionality at the table tow level to allow for easy individual row selections.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Add aria-labels for images and container page rows

https://github.com/containers/podman-desktop/issues/5639

### How to test this PR?
Check image and container rows to have aria-label property with the corresponding values.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerList.svelte
+++ b/packages/renderer/src/lib/container/ContainerList.svelte
@@ -583,7 +583,7 @@ function setStoppedFilter() {
           <!-- Display each container of this group -->
           {#if containerGroup.expanded}
             {#each containerGroup.containers as container, index}
-              <tr class="group h-12 bg-charcoal-800 hover:bg-zinc-700">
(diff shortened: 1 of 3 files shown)


ScrewTSW (approved):
LGTM

odockal (commented):
LGTM, but I would like to have someone from developer to take a look at tests.

deboer-tim:
FWIW the reason I didn't do this before is that name isn't unique, and there's no guarantee that any new item we decide to put in a table will have a name (although most of them appear to have one now). If the tests don't have situations where there are (e.g.) multiple images with the same name, then it'll be fine for now.

odockal:
No problem, I just could see possible ambiguity during testing if there are same names used in labels, but that could be workarounded with adding another role/name check. The unit test relying on multiple tooltips seemed awkward to me, that is all.

Issue #6740, opened 2024-04-10
Opening a webview within podman desktop from the app built in development mode shows empty screen


odockal:
### Bug description

When trying to open a webview from a podman desktop taht is run from e2e tests (podman desktop is built in development mode) I got only a black or empty screen.

### Operating system

Fedora 39

### Installation Method

Other

### Version

next (development version)

### Additional context

When opening the webview from a flatpak installed podman desktop, it opens up, but I see an error in the console:
```
Uncaught Error: The WebView must be attached to the DOM and the dom-ready event emitted before this method can be called.
    at EventEmitter.emit (VM4 sandbox_bundle:2:34768)
    at Object.send (<anonymous>:2:328)
    at IpcRenderer.<anonymous> (<anonymous>:2:720)
    at IpcRenderer.emit (VM4 sandbox_bundle:2:34768)
    at Object.onMessage (VM4 sandbox_bundle:2:50712)
```

cbr7:
![webview_error](https://github.com/containers/podman-desktop/assets/106525396/833e578d-b554-453c-9b59-aa889b0c972b)

I'm not sure if it's related or not, but I've added a screenshot of a warning from the webview console, where the same file `VM4_sandbox_bundle` is referenced as in the error from the main PD console.

Is it possible that some kind of security policy is preventing the dom-ready event from being emitted?

odockal:
@cbr7 TBH I have not seen the dom-ready not being omitted error when running the developer mode at all. Or it is that error is actually consumed somewhere?
I tried to debug the application and found a difference between running in dev mode and the one built locally, but I do not see exact point of failure.

I suspect that there might not be a problem on podman-desktop side, might be in the extension and how it takes over meta variable. Or maybe the problem lays somewhere around http server that should take care of dom elements recreation on podman desktop side. Server is running, I just do not know if it receives the request and if responds anything. I am just judging as there is missing iframe dom element in the dom structure once the button is clicked in podman-desktop devleopment mode.

odockal:
Reproducer on my branch: https://github.com/odockal/podman-desktop/tree/webview-repro
```
git clone git@github.com:odockal/podman-desktop.git
# git remote add odockal git@github.com:odockal/podman-desktop.git
git fetch odockal
git checkout odockal/webview-repro

yarn
# test:e2e build podman-desktop locally in dev mode
# once you have it built, you can re-run e2e tests faster by using test:e2e:run
yarn test:e2e
# yarn test:e2e:run
[...]

PR #6842, opened 2024-04-22
chore: make sure to run webview-preload init when running e2e tests

odockal:
### What does this PR do?
Let's webview-preload init method to be called when running e2e playwright tests
#6740

### How to test this PR?


axel7083 (approved):
LGTM

benoitf (changes requested):
the VITEST was for unit test but we should not include another variable for PLAYWRIGHT

axel7083:
> the VITEST was for unit test but we should not include another variable for PLAYWRIGHT

Since the `VITEST` is defined when running with playwright, the `init` is not call when running e2e tests[^1]. I understand the need of the if condition to allow proper unit testing, however we would need some ways to have the init method called... 

IMO we would either need an ENV stating it is a unit tests or a smoke e2e to be able to achieve the expected behaviour

[^1]: https://github.com/containers/podman-desktop/issues/6740#issuecomment-2066936618

cbr7 (approved):

odockal:
> the VITEST was for unit test but we should not include another variable for PLAYWRIGHT

I guess the VITEST env. var is there by default when using vitest test runner. It does not change the fact the e2e tests are different from unit ones. I welcome any other way how to make sure we do run e2e with webview-preload being correctly loaded.

other options atm: unset VITEST, set VITEST to 'e2e' and use this value to be checked in `webview-preload/index.ts`. But I am not sure about consequences of adjusting VITEST env. var.

xbabalov (approved):

packages/preload-webview/src/index.ts
```diff
@@ -35,7 +35,7 @@ export const init = (): void => {
   webviewPreload.init().catch((error: unknown) => console.error('Error while initializing the exposure', error));
 };
 
-// do not call init methd in case of testing
-if (!process.env.VITEST) {
+// do not call init methd in case of testing, but call when running E2E tests, see #6740
+if (!process.env.VITEST || process.env.PLAYWRIGHT_E2E) {
```
benoitf:
```suggestion
// do not call init method in case of testing
if (import.meta.env.MODE !== 'test') {
```

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -247,9 +247,12 @@ export class PodmanDesktopRunner {
   private setupPodmanDesktopCustomFolder(): object {
     const env: { [key: string]: string } = process.env as { [key: string]: string };
     const dir = join(this._customFolder);
-    console.log(`podman desktop custom config will be written to: ${dir}`);
+    console.log(`Podman Desktop custom config will be written to: ${dir}`);
     env.PODMAN_DESKTOP_HOME_DIR = dir;
 
+    console.log('Setting PLAYWRIGHT_E2E env. variable - used in dev. mode');
+    env.PLAYWRIGHT_E2E = 'true';
+
```
benoitf:
```suggestion
```

packages/preload-webview/src/index.ts
benoitf: use import MODE instead of VITEST

as for e2e we replace the mode when compiling to not be test

benoitf:
we're already avoiding to use test mode when building the files for e2e
https://github.com/containers/podman-desktop/blob/main/package.json#L50

so relies on TEST mode is working. (as variable is resolved at build time)

odockal:
> we're already avoiding to use test mode when building the files for e2e https://github.com/containers/podman-desktop/blob/main/package.json#L50
> 
> so relies on TEST mode is working. (as variable is resolved at build time)

Sounds good!

benoitf (approved):

odockal:
@benoitf Works nice. Can we merge?

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -247,7 +247,7 @@ export class PodmanDesktopRunner {
   private setupPodmanDesktopCustomFolder(): object {
     const env: { [key: string]: string } = process.env as { [key: string]: string };
     const dir = join(this._customFolder);
-    console.log(`podman desktop custom config will be written to: ${dir}`);
+    console.log(`Podman Desktop custom config will be written to: ${dir}`);
```
benoitf: the two changes in this file are unrelated to this PR so I would make it part of another commit/PR

odockal: ok.

benoitf (changes requested):
it looks the changes have been omitted

benoitf (approved):

odockal:
@benoitf Now it should be fine.

ScrewTSW (commented):
Verified latest changes, AI Lab extension is displayed correctly

LGTM

ScrewTSW (commented):
Verified latest changes, AI Lab extension is displayed correctly

LGTM

PR #6850, opened 2024-04-22
feat: use new page for extension details using podman-desktop:extension link

benoitf:
### What does this PR do?
`podman-desktop:extension/` should use the new page
update redirect + remove the old page handling the extension details

### Screenshot / video of UI

before:
image

after:
image

### How to test this PR?

on macOS for example use in a terminal `open podman-desktop:extension/redhat.openshift-local`


axel7083 (approved):
Might not really be directly related to this PR but when I put a wrong value I got an empty blank page

image

Otherwise working fine on windows 👍

benoitf:
good catch, I will follow-up with another PR to enhance the extension details page

deboer-tim (changes requested):
If you launch with an invalid extension id (e.g. a typo) then it doesn't work if you try again unless you go to another page within PD first. This can still be a separate issue, but it's a bit uglier than just not working once and I thought this PR was failing until I figured that out.

Nit: the breadcrumb defaults to the last page you visited (e.g. `Containers > Extension Details`), should be `Extensions` to match what we do elsewhere.

benoitf:
this PR is not about the extension details. Here it is just a switch

https://github.com/containers/podman-desktop/pull/6860 is the PR on which this PR should be rebased as there is a required for changes

benoitf:
rebasing

deboer-tim (approved):
> this PR is not about the extension details. Here it is just a switch
> 
> #6860 is the PR on which this PR should be rebased as there is a required for changes

Yes, my concern was that this PR was fine as-is but without #6860 it could trigger an ugly regression for people who can't type / copy-paste incorrectly. I've tested after rebasing and it is all working good now. 👍🏼

For the nit I was thinking maybe Loader.svelte should set the parent page, e.g.:
```
router.goto(...);
lastPage.set({ name: 'Extensions', path: '/extensions' });
```
but I'll leave that up to you.

PR #6859, opened 2024-04-22
chore: reduce navbar width

deboer-tim:
### What does this PR do?

Now that we've reduced the height of the main navbar items and the icons themselves, the bar itself and item selection area look very wide to me. Making these rectangles a little narrower/more square would match our previous design better and be more visually appealing.

This just reduces the width by 6px, which IMHO is just enough to counteract the effect of the changes.

### Screenshot / video of UI

Before:
Screenshot 2024-04-22 at 3 32 52 PM

After:
Screenshot 2024-04-22 at 3 31 51 PM

Related to #6774, from comment in #6777.

### How to test this PR?

Just look at left navbar.

### DIFF
--- a/tailwind.config.cjs
+++ b/tailwind.config.cjs
@@ -38,11 +38,11 @@ module.exports = {
         width: 'width',
       },
       width: {
-        'leftnavbar': '54px',
+        'leftnavbar': '48px',
         'leftsidebar': '225px',
       },
       minWidth: {
-        'leftnavbar': '54px',
+        'leftnavbar': '48px',
         'leftsidebar': '225px',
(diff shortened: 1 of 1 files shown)


benoitf:
I would not employ terms like `abnormally` 

More something: I prefer squares over rectangles or 'I dislike'

benoitf (approved):

cdrage (approved):

deboer-tim:
> I would not employ terms like `abnormally`
> 
> More something: I prefer squares over rectangles or 'I dislike'

Good point, I have tried to edit the description.

PR #6862, opened 2024-04-22
chore: extensions should be tagged with the Podman Desktop version

### LINKED ISSUE #6391: when releasing next or releases, extensions should have the versions of the product, not stay at 0.0.1 all the time

When displaying extensions, we see that version of embedded/default extensions is 0.0.1 which may confuse people

use the same version than the product (Podman Desktop)

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
avoid to always have versions of the extensions being v0.0.1 since day 1
it should match the version of Podman Desktop for the built-in extensions

fixes https://github.com/containers/podman-desktop/issues/6391

### How to test this PR?


deboer-tim (approved):
+1 to this. Code change LGTM.

My only concern is breaking the build, but Florent is better at this than me and easy enough to fix or revert quickly. :)

PR #6864, opened 2024-04-22
chore: align DD extension actions

deboer-tim:
### What does this PR do?

The action buttons for DD don't align with other extensions. This just adds a dummy Stop button that is never enabled so that the buttons visually match the other extensions.

There are other ways to solve this (changing the row layout, or adding a spacer that matches the width of the action buttons) but IMHO this keeps DD extensions as visually similar and consistent as possible to other extensions. From a user perspective I don't think this is a negative b/c it's clear 'this can't be stopped' vs potentially wondering 'why is there no stop button' before.

### Screenshot / video of UI

Before:
Screenshot 2024-04-22 at 4 25 30 PM

After:
Screenshot 2024-04-22 at 4 18 54 PM

One part of #6855.

### How to test this PR?

Look at extension page with a mix of DD and Podman Desktop extensions and confirm they look similar.

### DIFF
--- a/packages/renderer/src/lib/extensions/InstalledExtensionCardLeftLifecycleStop.svelte
+++ b/packages/renderer/src/lib/extensions/InstalledExtensionCardLeftLifecycleStop.svelte
@@ -16,11 +16,11 @@ async function stopExtension(): Promise<void> {
 }
 </script>
 
-{#if extension.state === 'started' && extension.type !== 'dd'}
+{#if extension.state === 'started'}
   <LoadingIconButton
     clickAction="{() => stopExtension()}"
     action="stop"
     icon="{faStop}"
-    state="{{ status: extension.state, inProgress }}"
+    state="{{ status: extension.type === 'dd' ? 'no-stop' : extension.state, inProgress }}"
     leftPosition="" />
 {/if}
--- a/packages/renderer/src/lib/extensions/InstalledExtensionCardLeftLifecycleStop.svelte
+++ b/packages/renderer/src/lib/extensions/InstalledExtensionCardLeftLifecycleStop.svelte
@@ -21,6 +21,6 @@ async function stopExtension(): Promise<void> {
     clickAction="{() => stopExtension()}"
     action="stop"
     icon="{faStop}"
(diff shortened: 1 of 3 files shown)


lstocchi (approved):
LGTM

packages/renderer/src/lib/extensions/InstalledExtensionCardLeftLifecycleStop.svelte
```diff
@@ -16,11 +16,11 @@ async function stopExtension(): Promise<void> {
...
 </script>
 
-{#if extension.state === 'started' && extension.type !== 'dd'}
+{#if extension.state === 'started'}
   <LoadingIconButton
     clickAction="{() => stopExtension()}"
     action="stop"
     icon="{faStop}"
-    state="{{ status: extension.state, inProgress }}"
+    state="{{ status: extension.type === 'dd' ? 'no-stop' : extension.state, inProgress }}"
```
benoitf: what means the state 'no-stop' ?

benoitf: should it be `unknown` ?

deboer-tim: I could have used any state other than started (to make the button disabled), but I thought a string like this would appear more intentional for someone looking at the code in the future. Maybe 'stop-unsupported' instead?

benoitf: it's just that inside the LoadingIconButton there is no reference to these states `no-stop` or `stop-unsupported`. So I would stick to some states known by the widget to avoid unexpected behavior (like if someone refactor the component, maybe stop-unsupported will not behave as expected)

`unsupported` might be a new state then rather than `unknown` as we know it's not supported so it should be disabled all the time

deboer-tim: I keep commenting/commit just when you respond, so let me try again. :) I've added an unsupported status in the latest commit and use that.

benoitf (approved):

PR #6878, opened 2024-04-23
fix: hide cpu, memory, size sliders on WSL

### LINKED ISSUE #2558: Windows podman machine creation does not respect number of CPUs choosen

lstocchi:
### What does this PR do?

It hides the cpu, memory, size sliders on WSL as they are meaningless.

### Screenshot / video of UI

image

it resolves #2558 

### How to test this PR?

1. try to create a new machine on WSL, you should not see the cpu, memory, size sliders

### DIFF
--- a/extensions/podman/package.json
+++ b/extensions/podman/package.json
@@ -113,7 +113,8 @@
           "minimum": 1,
           "maximum": "HOST_TOTAL_CPU",
           "scope": "ContainerProviderConnectionFactory",
-          "description": "CPU(s)"
+          "description": "CPU(s)",
+          "when": "!isWindows || !podman.isWSL"
         },
         "podman.factory.machine.memory": {
           "type": "number",
@@ -123,7 +124,8 @@
           "maximum": "HOST_TOTAL_MEMORY",
           "scope": "ContainerProviderConnectionFactory",
           "step": 500000000,
-          "description": "Memory"
+          "description": "Memory",
+          "when": "!isWindows || !podman.isWSL"
         },
         "podman.factory.machine.diskSize": {
           "type": "number",
@@ -133,7 +135,8 @@
           "maximum": "HOST_TOTAL_DISKSIZE",
           "step": 500000000,
           "scope": "ContainerProviderConnectionFactory",
-          "description": "Disk size"
+          "description": "Disk size",
+          "when": "!isWindows || !podman.isWSL"
         },
         "podman.factory.machine.image-path": {
           "type": "string",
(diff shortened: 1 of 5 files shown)


axel7083 (approved):
Working fine on my computer (windows) LGTM

extensions/podman/src/podman-configuration.ts
```diff
@@ -203,6 +203,16 @@ export class PodmanConfiguration {
     }
   }
 
+  async isValueInContainersConfig(regex: RegExp): Promise<boolean> {
```
benoitf: maybe use `match` like `matchContainersConfig` or `matchRegexpInContainersConfig`

extensions/podman/src/podman-configuration.ts
```diff
@@ -203,6 +203,16 @@ export class PodmanConfiguration {
...
+    try {
+      const containerConf = await this.readContainersConfigFile();
+      return regex.test(containerConf);
+    } catch (e) {
+      console.log(String(e));
```
benoitf: I think it's missing a bit of context in the log message

benoitf: also is it log or debug or warn or error ?

extensions/podman/package.json
```diff
@@ -113,7 +113,8 @@
           "minimum": 1,
           "maximum": "HOST_TOTAL_CPU",
           "scope": "ContainerProviderConnectionFactory",
-          "description": "CPU(s)"
+          "description": "CPU(s)",
+          "when": "!isWindows || !podman.isWSL"
```
benoitf: do we need to replicate in 3 different places the condition or we introduce direct flags like
```
"when": "podmanMachineCpuSupported"
"when": "podmanMachineDiskSupported"
"when": "podmanMachineMemorySupported"
```

lstocchi: I don't get what you're saying. It is already replicated in three places. Do you prefer direct flags?

benoitf: I mean, it was to avoid such complex conditions as we handle both the package.json and the condition in our case.

so I don't know if we need to put all the logic inside the package.json file or just make it working in the code (as it's more tested)

benoitf (approved):
I think it's a matter of taste if enablement is using 'internal' properties or do the logic.

I would be in favor of a single custom property as we can have unit tests that we don't have well in the package.json enablement but it's not a required for changes

lstocchi:
@benoitf updated with custom properties

benoitf (approved):
thx

PR #6888, opened 2024-04-23
chore(tests): get e2e tests to work with bootc in extension

cbr7:
### What does this PR do?
Gets bootc e2e tests working with webview


cdrage:
@cbr7 Shouldn't this be in https://github.com/containers/podman-desktop-extension-bootc ?

cbr7:
@cdrage no, these changes are part of the test framework, which ties in with the actual tests that are in that repo

Issue #6894, opened 2024-04-24
Update playwright POM and extensions tests after redesigned Extensions pages


odockal:
### Is your enhancement related to a problem? Please describe

After https://github.com/containers/podman-desktop/pull/6856, we need to update the POM for extensions used elsewhere.

### Describe the solution you'd like

Update index.ts of tests-playwright, redo dependencies, etc.

PR #6931, opened 2024-04-24
chore(tests): update extensions tests, pom and related functionality

odockal:
### What does this PR do?
update extensions tests and pom to align with new Extensions pages.
#6894 

### How to test this PR?
`yarn test:e2e:extension`


tests/playwright/src/model/pages/bootc-extension-page.ts
```diff
@@ -18,10 +18,10 @@
 
 import type { Page } from '@playwright/test';
 
-import { ExtensionPage } from './extension-page';
+import { ExtensionDetailsPage } from './extension-details-page';
```
benoitf: why do we have this class in this repository ? AFAIK it should be in its own repository

ScrewTSW (approved):
LGTM

tests/playwright/src/model/pages/extension-card-page.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  public async openExtensionDetails(): Promise<ExtensionDetailsPage> {
+    await playExpect(this.card).toBeVisible();
+    await this.card.scrollIntoViewIfNeeded();
+    await playExpect(this.detailsLink).toBeVisible();
+    await this.detailsLink.click();
+    return new ExtensionDetailsPage(this.page, this.extensionName);
+  }
+
+  async disableExtension(): Promise<this> {
```
cbr7, line 70: Aren't these methods already exposed from extension-details-page.ts?

odockal:
@cbr7 @ScrewTSW Changes done here are reflected in sso repo as wip pr. For you to see how it is gonna change at your extensions.

tests/playwright/src/model/pages/bootc-extension-page.ts
odockal: This is here for some time already. 
Was also an option to have them in their particular repos, altough, if you want to test multiple external resources, you would need to share POM. Ideally on once place and this repo seems like logical place.

benoitf: it should not be there at all. Podman Desktop has no knowledge on 3rd party extension

you may have different model page for each version of the extension so it can't be there

benoitf: > if you want to test multiple external resources, you would need to share POM

I don't get it there ? Nothing related to external extension should be there, only the things concerning the pages provided by Podman Desktop

tests/playwright/src/model/pages/extension-card-page.ts
odockal: Extensions Details page and Extensions/ExtensionsCard Page are two distinguished pages with the same locators.

tests/playwright/src/model/pages/bootc-extension-page.ts
odockal: ok, removing.

benoitf (approved):

odockal:
@benoitf I have noticed failing podman-extension tests. Fixed that by the latest commit.

benoitf:
FYI there is a typo in the last commit

`const exntesionLabelName = 'podman';`

`exntesion` vs `extension`

odockal:
> FYI there is a typo in the last commit
> 
> `const exntesionLabelName = 'podman';``
> 
> `exntesion` vs `extension`

Uf, nice, glad it is variable name.

tests/playwright/src/model/pages/bootc-extension-page.ts
cbr7: Well this is a change from the previous agreement as I understood it, previously it was discussed that extension tests should be in the extension repos, but that the framework part could be here so that we can use it in other tests also.

For example if we move the framework part in their respective extension repos it will be much harder to create an e2e test that could test other features that might also imply some functionality from an extension, like for example if you wanted an e2e test that would use functionality from AI Lab and bootc in one test.

tests/playwright/src/model/pages/extension-card-page.ts
cbr7: I mean yes, but if the functionality is identical it suggests that some kind of abstract class that both page object inherit should be used, there is no reason to have two different POM models implement literally identical functionalities.

PR #6933, opened 2024-04-24
fix: createVolume is returning unexpected object

### LINKED ISSUE #6932: createVolume is throwing errors in the UI
### Bug description

The volume is being created but an error is reported in the console about marshalling issue

### Operating system

macOS

### Installation Method

None

### Version

1.9.0

### Steps to reproduce

go to createVolume and create a volume

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
it turns out that Dockerode signature is saying that createVolume returns a `VolumeCreateResponse` object but in fact it returns a `Volume` object then it means that we have an object with modem field, etc and only the name of the volume, not all attributes of what is sent by the REST API

make the call returning the raw data from the REST API VolumeCreateResponse and not a dumb Volume object.

fixes https://github.com/containers/podman-desktop/issues/6932

### How to test this PR?

test case of the issue or unit tests

### DIFF
--- a/packages/main/src/plugin/dockerode/libpod-dockerode.spec.ts
+++ b/packages/main/src/plugin/dockerode/libpod-dockerode.spec.ts
@@ -259,3 +259,17 @@ test('Check using libpod/manifests/{name}/json endpoint', async () => {
   expect(manifest.manifests[0].size).toBe(0);
   expect(manifest.manifests[0].urls).toEqual([]);
 });
+
+test('Check create volume', async () => {
+  nock('http://localhost')
+    .post(`/volumes/create?Name=foo`)
+    .reply(200, { Name: 'foo', Driver: 'local', Scope: 'local' });
+
+  const api = new Dockerode({ protocol: 'http', host: 'localhost' });
+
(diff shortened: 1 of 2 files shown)


deboer-tim (approved):

cdrage (approved):
https://github.com/containers/podman-desktop/assets/6422176/9cb5c5ea-8374-416d-88cb-64fcd94a14d2

Was able to create the volume totally fine, no changes / breaking on the podman desktop side. Code LGTM

jeffmaury (approved):
LGTM but did not understand the trick :)

benoitf:
dockerode, when calling the `/volume/create` REST API, grab the data returned from the API and only keep the name and create a wrapper, Volume around this object and from there you can call inspect(), for example

but we're interested in the raw result when calling the `create` endpoint, not an object that contains modem field and name field.

so we don't wrap the object into a Volume object as it's done there
https://github.com/apocas/dockerode/blob/master/lib/docker.js#L1060

we just resolve with the data

PR #6934, opened 2024-04-24
fix: stop informer not more valid and handle similar contexts

### LINKED ISSUE #6830: Cluster unreachable even though i can connect
### Bug description

I'm working on a demo using kind and it happened multiple times that desktop shows that the cluster is unreachable even though i can connect to it. If i stop and restart desktop it works again.

One thing to replicate is when you create a new kind cluster. After the creation it tells it is not reachable.

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR fixes some weird behavior with the informers.
So far, we were just using the context name to detect if a context was deleted, updated, added but it may happen that a context only differs for some internal value (e.g the cluster server url). E.g create podman machine, create kind cluster, delete podman machine, the kind cluster is deleted but the kubeconfig remains dirty, so if you create a new kind cluster, the kubeconfig is almost the same expect the server url and the informer are not updated

Another thing is that we restart an informer if it fails at starting. This way, it may happen that you end up having multiple informers even though you only have one cluster. This PR adds a check to prevent a context that do not belong to the kubeconfig to be restarted again.

E.g. As before you start with a dirty kubeconfig. 
Podman desktop starts an informer which will fail as the cluster is not reachable and it will try to restart it again and again after a timeout 

https://github.com/containers/podman-desktop/blob/654b227832b93b27631507ca7cb9e09e603cf3b3/packages/main/src/plugin/kubernetes-context-state.ts#L790
 
In the meantime you create a new kind cluster.
Desktop will try to stop the old informer by calling the kubeclient library stop method (without any success as there is no active informer, we restart it on desktop side), and create a new informer pointing to the new cluster. 
You end up having a working informer and a failing one that keep trying at starting infinitely.

### Screenshot / video of UI

N/A

it resolves #6830 

### How to test this PR?

1. create a podman machine 
2. create a kind cluster
3. delete the podman machine
4. your kubeconfig should still have the entry regarding the kind cluster
5. recreate a podmna machine + kind cluster having the same name as before
6. if you go to the deployment/service/ingress it should be connected to the newly created cluster

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.spec.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.spec.ts
@@ -19,9 +19,10 @@
 import type { ErrorCallback, KubernetesObject, ObjectCallback } from '@kubernetes/client-node';
 import * as kubeclient from '@kubernetes/client-node';
 import { KubeConfig, makeInformer } from '@kubernetes/client-node';
-import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { afterEach, beforeAll, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import type { ApiSenderType } from './api.js';
+import type { KubeContext } from './kubernetes-context.js';
 import type { ContextGeneralState, ResourceName } from './kubernetes-context-state.js';
 import { ContextsManager, ContextsStates } from './kubernetes-context-state.js';
 
@@ -305,8 +306,10 @@ describe('update', async () => {
     });
 
     apiSenderSendMock.mockReset();
+    console.log('DA QUI');
     await client.update(kubeConfig);
 
+    vi.advanceTimersToNextTimer();
     vi.advanceTimersToNextTimer();
     expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-general-state-update', expectedMap);
     expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-current-context-general-state-update', {
@@ -330,6 +333,7 @@ describe('update', async () => {
     });
 
     apiSenderSendMock.mockReset();
+    console.log('DA QUI 2');
     await client.update(kubeConfig);
     expectedMap = new Map<string, ContextGeneralState>();
     expectedMap.set('context1', {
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/kubernetes-context-state.spec.ts
```diff
@@ -305,8 +306,10 @@ describe('update', async () => {
     });
 
     apiSenderSendMock.mockReset();
+    console.log('DA QUI');
```
axel7083:
```suggestion
```

packages/main/src/plugin/kubernetes-context-state.spec.ts
```diff
@@ -330,6 +333,7 @@ describe('update', async () => {
     });
 
     apiSenderSendMock.mockReset();
+    console.log('DA QUI 2');
```
axel7083:
```suggestion
```

packages/main/src/plugin/kubernetes-context-state.spec.ts
lstocchi: 😓 Thanks!!

axel7083 (approved):
LGTM, Just tested it following the steps you provided and works like a charm!

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -359,15 +359,79 @@ export class ContextsManager {
...
+  isContextChanged(kubeconfig: KubeConfig): boolean {
+    // if the current context name in the kubeconfig is different from the latest context we saved -> true
+    if (kubeconfig.currentContext !== this.currentContext?.name) {
+      return true;
+    }
+
+    // if the current context name in the kubeconfig is undefined ->  true
+    if (!kubeconfig.currentContext) {
+      return true;
+    }
```
feloy: What happens if there was already no currentContext in kubeConfig before?
It seems to me that we will detect a change every time

lstocchi: You're right, removed. Thanks!!

feloy (approved):

PR #6938, opened 2024-04-24
chore: increase timeout values

benoitf:
### What does this PR do?
try to increase timeout values to avoid that the catalog is not retrieved on slow networks

(we should try to pull at regular interval if it doesn't succeed for proper fix)

related to https://github.com/containers/podman-desktop/issues/5349

### How to test this PR?


deboer-tim (approved):

cdrage (approved):
LGTM. Tested and saw no difference, tried it on a 5G network and was the same.

PR #6940, opened 2024-04-25
fix: add back the error if an extension fails

### LINKED ISSUE #6917: See extensions error logs when status failed

With the new extension page we lost the possibility to see the error raised by an extension on activate.

image

A way to see the logs

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
With the revamp of extensions, it was no longer possible to see the details in case of an extension failure

Add a new tab in case of failure within the details of the extension to see the error

### Screenshot / video of UI

image

fixes https://github.com/containers/podman-desktop/issues/6917

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/extensions/ExtensionDetails.svelte
+++ b/packages/renderer/src/lib/extensions/ExtensionDetails.svelte
@@ -7,11 +7,13 @@ import { combinedInstalledExtensions } from '/@/stores/all-installed-extensions'
 import { catalogExtensionInfos } from '/@/stores/catalog-extensions';
 
 import FeaturedExtensionDownload from '../featured/FeaturedExtensionDownload.svelte';
+import Button from '../ui/Button.svelte';
 import DetailsPage from '../ui/DetailsPage.svelte';
 import EmptyScreen from '../ui/EmptyScreen.svelte';
 import ExtensionStatus from '../ui/ExtensionStatus.svelte';
 import type { ExtensionDetailsUI } from './extension-details-ui';
 import ExtensionBadge from './ExtensionBadge.svelte';
+import ExtensionDetailsError from './ExtensionDetailsError.svelte';
 import ExtensionDetailsReadme from './ExtensionDetailsReadme.svelte';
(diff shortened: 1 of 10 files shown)


axel7083:
I want to ask an open question, should we not have some kind of generic `logs` for the extension and redirect the stdout to this instead of just having the errors ?

image

We seems to already have some kind of insight about who is producing some logs, might be interesting to have them also in this place.

This would have several advantages, we could ask the user to specifically provide the logs from this tab when opening an issue. And it would be also easier for them to understand the errors/logs as they would be specific to the extension.

(In the current timeframe, I am happy with the error tab)

packages/renderer/src/lib/extensions/ExtensionDetails.svelte
```diff
@@ -63,10 +66,32 @@ $: extension = derived(
...
+          }}"
+          selected="{screen === 'ERROR'}">Error</Button>
+      {/if}
+    </svelte:fragment>
+
     <svelte:fragment slot="content">
       <div class="flex w-full h-full overflow-y-auto p-4 flex-col lg:flex-row">
-        <ExtensionDetailsSummaryCard extensionDetails="{$extension}" />
-        <ExtensionDetailsReadme readme="{$extension.readme}" />
+        {#if screen === 'README'}
```
axel7083, line 89: Why not using navigiation url here ? instead of storing the state in a variable

benoitf: because url is the same

benoitf:
> want to ask an open question, should we not have some kind of generic logs for the extension and redirect the stdout to this instead of just having the errors ?

I think it would be an extra tab for the logs.
In the log you won't see the activation error (unless the application is logging the error)

you can file an enhancement issue

axel7083 (commented):
image

How do I restart the extension ? I do not see start button

benoitf:
@axel7083 should be available now

jeffmaury (approved):
LGTM

PR #6943, opened 2024-04-25
chore(tests): refactor and create bootc page

cbr7:
### What does this PR do?
Adds bootc extension POM pages framework support.


benoitf (commented):
bootc page can't be in this repository

cbr7:
@benoitf @odockal  there are good reasons why this should be here, I'm making this draft PR as a concept to discuss this. The request to move the framework part into the extension repos is a major architectural request that needs to be debated, because it has severe consequences for long term automation, especially in regards to cross functional tests that might require functionalities from one or multiple extensions at the same time.

odockal:
@benoitf We have had external extensions POM definitions here to be part of tests-playwright npm package already, see: #6728, #5797. Not having them here right now changes a concept we started with. Not like it cannot be changed, but we should stick to what we started with.

benoitf:
if there is a cross test, then it's not in this repository, and I assume the test will import packages from each extension providing the model page

benoitf:
we said since day 0 that this repository should only contain code about what is managed by this repository.

Page model are depending on the lifecycle of the extension, not Podman Desktop so it can't be there.

If I rewrite the extension UI, I should not have to update or wait a new release of Podman Desktop to bring the updated model page.

cbr7:
>we said since day 0 that this repository should only contain code about what is managed by this repository.

In that case I don't understand why the test framework cannot be managed outside this repo, with only the tests and the relevant POM models being developed in here. 

The issue currently is that there is a very high chance of ending up with circular dependencies when testing cross functions, not to mention that the extensions aren't as isolated as assumed. For example... navigation to a bootc webview page is done from Images and Image details pages via the "Build Disk Image" button. In this case the button exists as part of PD, however it navigates out of PD and into the webview, but actually having a method in the Images/Image Details pages from PD would not be possible without a reference a hypothetical bootc npm package depedency, and as such making it circular.

benoitf:
I don't see where it could have circular dependencies

Extensions depend on Podman Desktop (not the opposite)

and if you need something that is cross extensions, then, you have code depending on extensions and podman desktop so there is no circular dependency as well

but if you add some specific extension code in Podman Desktop then here you're bringing circular dependency (the extension depends on Podman Desktop and Podman Desktop depends on the extension)

cbr7:
@benoitf @odockal I'll move the bootc pom page into the other repo, refactored the code here to given that

ScrewTSW (approved):
LGTM

PR #6947, opened 2024-04-25
feat(RecommendationsRegistry): uses deterministic method to choose which banner is displayed

### LINKED ISSUE #6941: Extension banners displayed should be deterministic for a period of time

Since the introduction of the new Extension Banner on the Podman Desktop Dashboard, we only had one. So we were not facing issues of flickering.

However, when having multiple, it can randomly changed based on system events. 

We need to enforce a deterministic system to keep the same banner for a set of time. 

@benoitf recommended to have the same for one hour.

#### Pseudo code

The idea is to take the the current hours, and use id modulo the number of banner we have to get it as a starting index.
```
const limit = 5;
const banners: ExtensionBanner[] = getExtensionBanners(limit ) // get all extensions banner

const startingIndex = new Date().getHours() % banners.length;

const output: ExtensionBanner[] = []
for(int i = 0; i < limit ; i ++) {
     output.push(banners[(startingIndex + i) % banners.length])
}
```

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

In todays implementation we are using a random system to determine which banner will be displayed.

https://github.com/containers/podman-desktop/blob/662a07f22197ce70b83b3f8c18695088e206b010/packages/main/src/plugin/recommendations/recommendations-registry.ts#L74

This has a serious drawback, each call a different one could be returned, leading to ui flickering, since the extension banner store is subscribed to specific events (required) it get re-render a little bit too often

https://github.com/containers/podman-desktop/blob/74923cc173ebefbbb852f1d90d02bc83d14fce69/packages/renderer/src/stores/extensionBanners.ts#L24-L32

To avoid this flickering, we need to have the same result for a period of time, here we choose by the hours.

Fixes https://github.com/containers/podman-desktop/issues/6941

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/recommendations/recommendations-registry.spec.ts
+++ b/packages/main/src/plugin/recommendations/recommendations-registry.spec.ts
@@ -223,4 +223,83 @@ describe('getExtensionBanners', () => {
 
     expect(featuredMock.getFeaturedExtensions).toHaveBeenCalled();
   });
+
+  test('same time should return same arrays', async () => {
+    getRecommendationIgnored.mockReturnValue(false);
+    vi.mocked(featuredMock.getFeaturedExtensions).mockResolvedValue(
+      Array.from({ length: 10 }, (_, i) => ({
+        id: `dummy.id-${i}`,
+        builtin: false,
+        description: '',
+        categories: [],
+        displayName: '',
+        fetchable: false,
+        icon: '',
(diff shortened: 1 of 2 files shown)


benoitf (approved):

jeffmaury (approved):
LGTM

PR #6957, opened 2024-04-25
chore: handle starting as mode where stop should be visible

benoitf:
### What does this PR do?
extension is going from stopped to starting to started and we should display stop in starting state as the start button will disappear as soon as stopped condition is gone, so during the starting phase there is nothing which might cause some blinking

### How to test this PR?


cdrage (approved):

jeffmaury (approved):
LGTM

PR #6967, opened 2024-04-25
fix: boolean values should be true and not 'on' using forms

### LINKED ISSUE #2061: Boolean configuration property is not reported as boolean
### Bug description

When a provider is defining a configuration property as boolean, the create resource wizard is using checkbox is the UI but the value given to the create handler is a string 'on' or nothing. The value should be a boolean

### Operating system

Win11 Pro

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
In HTML, when using submit, only checked checkboxes are submitted.
so for example it means that if we have the checkbox rootful, if the user don't check, it's like a missing parameter.

As we know which properties we're expecting, in case of missing parameter, we set to false.

And another issue, when checkbox is checked, in the form we have 'on' as value, not true. So we need to replace by a boolean value so the factories receiving the properties are able to properly handle true/false instead of 'on' 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value

fixes https://github.com/containers/podman-desktop/issues/2061

### How to test this PR?
you could print what an extension receive as params when checking/unchecking some checkboxes

### DIFF
--- a/extensions/kind/src/create-cluster.ts
+++ b/extensions/kind/src/create-cluster.ts
@@ -138,7 +138,7 @@ export async function createCluster(
   let ingressController = false;
 
   if (params['kind.cluster.creation.ingress']) {
-    ingressController = params['kind.cluster.creation.ingress'] === 'on';
+    ingressController = params['kind.cluster.creation.ingress'];
   }
 
   // grab custom kind node image if defined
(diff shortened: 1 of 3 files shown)


jeffmaury (approved):
LGTM

PR #6970, opened 2024-04-25
chore: disable universal build for airgap mode on macOS

benoitf:
### What does this PR do?
macOS universal artifact is too big for GitHub, so need to do one per arch

https://github.com/containers/podman-desktop/issues/6885

### How to test this PR?

you can use `AIRGAP_DOWNLOAD=true yarn compile:current` command on macOS


deboer-tim (approved):

jeffmaury (approved):
LGTM but I noticed the next-universal.zip is also missing

PR #7022, opened 2024-04-30
fix: if installed, display the installed version

### LINKED ISSUE #6989: version mismatched in extension catalog
### Bug description

When installing an extension with a version greater than the one available in the catalog we get a weird behaivour where the catalog page show `installed` with a version which is the one the real one.

The `installed` tab show the real version, and the extension details also. (Not sure what is the expected behaviour here)

### Operating system

Windows 11

### Installation Method

Other

### Version

next (development version)

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
in case of an extension being installed, display the version as being the one installed, not the one that is in the catalog

fixes https://github.com/containers/podman-desktop/issues/6989

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/extensions/extensions-utils.spec.ts
+++ b/packages/renderer/src/lib/extensions/extensions-utils.spec.ts
@@ -186,6 +186,7 @@ const installedExtensions: CombinedExtensionInfoUI[] = [
   },
   {
     id: 'idYInstalled',
+    version: '2.0.0Y',
   },
 ] as unknown[] as CombinedExtensionInfoUI[];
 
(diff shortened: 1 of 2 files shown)


axel7083 (approved):
Tested and works fine LGTM

deboer-tim (changes requested):
It seems like this PR fixes the local version - but still assumes that there is only 1 version of the extension both locally and in the catalog.

I turned off auto-update and hacked the version of my local minikube extension to be old. It now says that's the version in the catalog, and I can't upgrade b/c it (the newer one in the catalog) is already installed.

benoitf:
AFAIK we lost manual update in the new design

about the changes, the other way around I was thinking was to display two versions in the card like catalog version and installed version.

Anyway I don't see any proper fix with the current design using one field with 2 different values being possible

benoitf:
will need a UI design

benoitf:
opened https://github.com/containers/podman-desktop/pull/7025 instead

PR #7024, opened 2024-04-30
fix: use podman machine inspect to know if machine is rootful

### LINKED ISSUE #7000: Switching default podman connection via Podman Desktop always uses rootless connection
### Bug description

Creating/having a rootless machine makes all previously existing or new rootful machines implicitly rootless when they are used as default connections.

I wrote one way to reproduce but the outcome is the same no matter if the rootless machine was created first, last, or new rootful machine was created and connected after the rootless one. Any rootful machine default connection is always rootless unless manually swapped in CLI. 

It can be fixed by deleting all of the machines. After deletion, starting a new rootful machine and setting its connection to default results in a rootful connection. Deleting only the rootless machine does not help.

### Operating system

Windows 10

### Installation Method

### Version

1.10.0

### Steps to reproduce

1. Install Podman Desktop and Podman
2. Create a rootful machine and set its connection to default
3. Assert the connection is rootful in CLI (podman system connection list)
4. Stop the rootful machine and create a new rootless machine, set its connection to default
5. Assert the connection is rootless in CLI
6. Stop the rootless machine, restart the rootful machine and set its connection to default
7. Assert the connection is rootful in CLI
    Actual result: the connection is rootless

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
in the past, the flag was not exposed so we had to parse some configuration files, but now we should rely on the `podman machine inspect` command to know if it's rootful or rootless

fixes https://github.com/containers/podman-desktop/issues/7000

### How to test this PR?

use test case of the linked issue

### DIFF
--- a/extensions/podman/src/extension.spec.ts
+++ b/extensions/podman/src/extension.spec.ts
@@ -762,7 +762,7 @@ test('test checkDefaultMachine - if user wants to change machine, check that it 
     },
   ];
 
-  // return as inspect result a rootful machine
+  // return as inspect result a rootless machine
   const inspectCall = vi.spyOn(extensionApi.process, 'exec').mockResolvedValueOnce({
     stdout: JSON.stringify(fakeInspectJSON),
   } as extensionApi.RunResult);
@@ -799,7 +799,7 @@ test('test checkDefaultMachine - if user wants to change machine, check that it 
     },
   ];
 
-  // return as inspect result a rootful machine
+  // return as inspect result a rootless machine
   const inspectCall = vi.spyOn(extensionApi.process, 'exec').mockResolvedValueOnce({
     stdout: JSON.stringify(fakeInspectJSON),
   } as extensionApi.RunResult);
--- a/extensions/podman/src/extension.spec.ts
+++ b/extensions/podman/src/extension.spec.ts
@@ -708,12 +708,21 @@ test('test checkDefaultMachine - if there is no machine marked as default, take 
 });
 
 test('test checkDefaultMachine - if user wants to change default machine, check if it is rootful and update connection', async () => {
-  const spyExecPromise = vi.spyOn(extensionApi.process, 'exec').mockImplementation(
-    () =>
-      new Promise<extensionApi.RunResult>(resolve => {
-        resolve({ stdout: JSON.stringify(fakeMachineInfoJSON) } as extensionApi.RunResult);
-      }),
-  );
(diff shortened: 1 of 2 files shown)


extensions/podman/src/extension.spec.ts
```diff
@@ -732,30 +741,31 @@ test('test checkDefaultMachine - if user wants to change default machine, check
...
+  } as extensionApi.RunResult);
+
+  const fakeInspectJSON = [
+    {
+      Name: 'podman-machine-default',
+      Rootful: false,
+    },
+  ];
+
+  // return as inspect result a rootful machine
```
feloy:
```suggestion
  // return as inspect result a rootless machine
```

extensions/podman/src/extension.spec.ts
```diff
@@ -774,81 +784,39 @@ test('test checkDefaultMachine - if user wants to change machine, check that it
...
+  ];
 
-test('if user wants to change machine, check that it only change the connection once if it fails at checking rootful because file is empty', async () => {
-  const spyExecPromise = vi.spyOn(extensionApi.process, 'exec').mockImplementation(
-    () =>
-      new Promise<extensionApi.RunResult>(resolve => {
-        resolve({ stdout: JSON.stringify(fakeMachineInfoJSON) } as extensionApi.RunResult);
-      }),
-  );
+  // return as inspect result a rootful machine
```
feloy:
```suggestion
  // return as inspect result a rootless machine
```

extensions/podman/src/extension.ts
```diff
@@ -378,45 +378,35 @@ export async function checkDefaultMachine(machines: MachineJSON[]): Promise<void
...
+      let machineIsRootful = false;
+      // grab result of the command 'podman machine inspect <machine-name> and check if attribute is Rootful
       try {
+        const { stdout: machineInspectJson } = await extensionApi.process.exec(getPodmanCli(), [
+          'machine',
+          'inspect',
+          runningMachine.Name,
+        ]);
+        const machinesInspect = JSON.parse(machineInspectJson);
+        // find the machine machine in the array
```
feloy:
```suggestion
        // find the machine name in the array
```

feloy (approved):
Fixes the problem on mac M3 with podman 5.0.2

odockal:
LGTM. Tested with 5.0.2 and 4.9.4.

odockal (commented):
Approving

PR #7025, opened 2024-04-30
chore: display additional field if installed version is different

### LINKED ISSUE #6989: version mismatched in extension catalog

benoitf:
### What does this PR do?
Display installed version if different than the catalog version
for example if I have a different version than the one in the catalog I should be able to see if while seeing the catalog because if I click on the details I will the version of the installed extension

### Screenshot / video of UI

image

fixes https://github.com/containers/podman-desktop/issues/6989

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/extensions/CatalogExtension.spec.ts
+++ b/packages/renderer/src/lib/extensions/CatalogExtension.spec.ts
@@ -112,3 +112,28 @@ test('Expect to see featured and fetch button', async () => {
     expect.any(Function),
   );
 });
+
+test('Expect to have version of installed one', async () => {
+  const catalogExtensionUI: CatalogExtensionInfoUI = {
+    id: 'myId',
+    displayName: 'This is the display name',
+    isFeatured: false,
+    fetchable: false,
+    fetchLink: '',
+    fetchVersion: '1.0.0',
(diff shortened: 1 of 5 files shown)


deboer-tim (approved):
Confirmed that I can now see the difference between an old installed extension and a newer one in the catalog.

The catalog details page still lists the local version, and if auto-update is disabled I have to delete the local copy to be able to update to the catalog version, but +1 since this is an improvement and now correct. We can track remaining issues for 1.11+ separately.

feloy:
Works for me with a newer installed version than the catalog version.

```
v0.1.3 (installed: v0.1.4)
```

Is it expected to have a different color for the installed version (gray-500) compared to the catalog version (gray-700)?

benoitf:
@feloy no, I will revert

benoitf:
@feloy fixed

feloy (approved):

PR #7042, opened 2024-05-02
chore: bump @types/dockerode version to 3.3.29

axel7083:
### What does this PR do?

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69385 changed the return type of the `pull` method argument `callback` from `callback: Callback<any>` to `callback: Callback<NodeJS.ReadableStream>`.

The problem are from the `docker-modem` library that was resolving an old version of its types `@types/docker-modem:3.0.2` when only the version `@types/docker-modem:3.0.6` (latest) was compatible with the new change. 

### How to test this PR?

Pipeline should be ✅

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,7 +196,6 @@
   "resolutions": {
     "trim": "0.0.3",
     "ssh2": "1.11.0",
-    "@types/node": "^20",
-    "@types/docker-modem": "3.0.6"
+    "@types/node": "^20"
   }
 }
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "@rollup/plugin-node-resolve": "^15.2.3",
     "@rollup/plugin-typescript": "^11.1.6",
     "@types/adm-zip": "^0.5.5",
-    "@types/dockerode": "^3.3.28",
+    "@types/dockerode": "^3.3.29",
     "@types/express": "^4.17.21",
     "@types/getos": "^3.0.4",
     "@types/js-yaml": "^4.0.9",
@@ -196,6 +196,7 @@
   "resolutions": {
     "trim": "0.0.3",
     "ssh2": "1.11.0",
-    "@types/node": "^20"
+    "@types/node": "^20",
+    "@types/docker-modem": "3.0.6"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -196,6 +196,7 @@
   "resolutions": {
     "trim": "0.0.3",
     "ssh2": "1.11.0",
-    "@types/node": "^20"
+    "@types/node": "^20",
+    "@types/docker-modem": "3.0.6"
```
benoitf: we don't need this resolution, you can just have the yarn.lock change

yarn.lock
```diff
@@ -3857,18 +3857,18 @@
   dependencies:
     "@types/ms" "*"
 
-"@types/docker-modem@*":
-  version "3.0.2"
-  resolved "https://registry.npmjs.org/@types/docker-modem/-/docker-modem-3.0.2.tgz"
-  integrity sha512-qC7prjoEYR2QEe6SmCVfB1x3rfcQtUr1n4x89+3e0wSTMQ/KYCyf+/RAA9n2tllkkNc6//JMUZePdFRiGIWfaQ==
+"@types/docker-modem@*", "@types/docker-modem@3.0.6":
```
benoitf:
```suggestion
"@types/docker-modem@*":
```

package.json
axel7083:
```suggestion
    "@types/node": "^20"
```

packages/main/src/plugin/container-registry.ts
```diff
@@ -1132,7 +1132,7 @@ export class ContainerProviderRegistry {
     try {
       const authconfig = this.imageRegistry.getAuthconfigForImage(imageName);
       const matchingEngine = this.getMatchingEngineFromConnection(providerContainerConnectionInfo);
-      const pullStream = await matchingEngine.pull(imageName, {
+      const pullStream: NodeJS.ReadableStream = await matchingEngine.pull(imageName, {
```
benoitf: can't we omit NodeJS.ReadableStream there ?

yarn.lock
```diff
@@ -2114,7 +2114,7 @@
     "@docusaurus/theme-search-algolia" "3.2.1"
     "@docusaurus/types" "3.2.1"
 
-"@docusaurus/react-loadable@5.5.2":
+"@docusaurus/react-loadable@5.5.2", "react-loadable@npm:@docusaurus/react-loadable@5.5.2":
```
benoitf: do you know why docusaurus imports are being changed ?

packages/main/src/plugin/container-registry.ts
axel7083: yes

benoitf (approved):
Thx

PR #7053, opened 2024-05-02
fix: increasing default timeout for extension activation

axel7083:
### What does this PR do?

Going from `10s` to `20s`

### How to test this PR?


packages/main/src/plugin/extension-loader.ts
```diff
@@ -259,7 +259,7 @@ export class ExtensionLoader {
         [ExtensionLoaderSettings.SectionName + '.' + ExtensionLoaderSettings.MaxActivationTime]: {
           description: 'Maximum activation time for an extension, in seconds.',
           type: 'number',
-          default: 10,
+          default: 20,
```
benoitf: maybe introduce a constant for 20 so we use it at both line 262 and 1377

benoitf (approved):

deboer-tim (approved):

PR #7076, opened 2024-05-03
chore: update website for macOS airgap binaries

benoitf:
### What does this PR do?
the airgap installer is no longer 'universal' as it was too big (reach 2GB limit max of GitHub artifacts)

### How to test this PR?

### DIFF
--- a/website/src/pages/downloads/macos.tsx
+++ b/website/src/pages/downloads/macos.tsx
@@ -67,9 +67,6 @@ async function grabfilenameforMac(
 
   const airgapsetupArm64 = macosArm64AirgapSetupAssets?.[0]?.browser_download_url;
 
-  console.log('airgapsetupX64', airgapsetupX64);
-  console.log('airgapsetupArm64', airgapsetupArm64);
-
   const data = {
     version: jsonContent.name,
     universal: unifiedMacLink.browser_download_url,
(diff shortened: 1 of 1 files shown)


website/src/pages/downloads/macos.tsx
```diff
@@ -48,13 +52,32 @@ async function grabfilenameforMac(
...
+  const airgapsetupX64 = macosX64AirgapSetupAssets?.[0]?.browser_download_url;
+
+  const macosArm64AirgapSetupAssets = assets.filter(
+    asset => (asset.name as string).endsWith('-arm64.dmg') && asset.name.includes('airgap'),
+  );
+
+  const airgapsetupArm64 = macosArm64AirgapSetupAssets?.[0]?.browser_download_url;
+
+  console.log('airgapsetupX64', airgapsetupX64);
+  console.log('airgapsetupArm64', airgapsetupArm64);
```
feloy: to be removed?

feloy (approved):
tested, by fetching the release info for v1.10.1 instead of current `latest`, works fine, I can download the 4 different DMG files

website/src/pages/downloads/macos.tsx
benoitf: fixed, thanks

deboer-tim (approved):
Verified against v1.10.2.

PR #7077, opened 2024-05-03
chore: correct port validation

### LINKED ISSUE #6390: Port validation in the Run Image Page is not properly validating the port value until port number has 4 digits
### Bug description

port_already_in_use

### Operating system

Fedora 39

### Installation Method

### Version

1.8.0

### Steps to reproduce

1. Start container with a port exposed (httpd-24)
2. Try to create/start another container with exposed port
3. Assert: default port is different from first app.
4. Try to change the port number and see validation message
5. Assert: Port number is evaluated and the message fits
Actual result: port number under < 1000 shows unfitting message (`port X is already in use`)

### Relevant log output

_No response_

### Additional context

Given the message here, I am assuming that port number < 1000 is not a problem. Also, not a number character should probably trigger different message, although the message is not incorrect then.

Screenshot_20240312_165525

vzhukovs:
### What does this PR do?

The following changes request provides better port validation process for Run Images dialog.

#6390

### How to test this PR?

Try to create two different containers with same exposed port to check that new port is proposed in Run Image dialog and validation of input works proper.

### DIFF
--- a/packages/main/src/plugin/util/port.ts
+++ b/packages/main/src/plugin/util/port.ts
@@ -19,11 +19,6 @@
 import * as net from 'node:net';
 import { type NetworkInterfaceInfoIPv4, networkInterfaces } from 'node:os';
 
-export interface PortStatus {
-  available: boolean;
-  message?: string;
-}
-
 /**
  * Find a free port starting from the given port
  */
@@ -33,8 +28,10 @@ export async function getFreePort(port = 0): Promise<number> {
   }
   let isFree = false;
   while (!isFree) {
-    isFree = (await isFreePort(port)).available;
-    if (!isFree) {
+    try {
+      await isFreePort(port);
+      isFree = true;
+    } catch (e) {
       port++;
     }
   }
@@ -50,10 +47,10 @@ export async function getFreePortRange(rangeSize: number): Promise<string> {
   let startPort = port;
 
   do {
-    const portStatus = await isFreePort(port);
-    if (portStatus.available) {
+    try {
+      await isFreePort(port);
       ++port;
-    } else {
+    } catch (e) {
       ++port;
       startPort = port;
     }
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -183,6 +183,7 @@
     "getos": "^3.2.1",
     "got": "^14.2.1",
     "hpagent": "^1.2.0",
+    "is-elevated": "^4.0.0",
```
axel7083: I am not a big fan of this library as it has no support, not been updated since Aug 11, 2021, it fetched two other library the `is-admin` and `is-root` which have the same problem.

The `is-admin` is running the following on windows `fsutil dirty query C:` and the `is-root` is simply checking the user uid..

I don't know what is our policy for adding packages cc @benoitf

benoitf: I would not proceed with this library

benoitf: I mean, the library is OK from maintenance POV as they're managed by sindresorhus so there is support

but I'm not sure why we would need that library, I mean, in which case people run Podman Desktop with admin privileges ?

I would not allow ports < 1024 in all cases (Windows, macOS or Linux)

also Podman Desktop could be run in admin mode but the gvproxy in user mode so it would fail anyway

vzhukovs: >I would not allow ports < 1024 in all cases (Windows, macOS or Linux)

This simplifies the whole checks, will going to update the PR by removing this check.

I just rely on that scenario, that all ports, that are below 1024 by default requires administrative permissions to expose.

benoitf: @vzhukovs yes but I think it's simpler/easier/more robust to not allow that anyway using Podman Desktop

thanks for updating the PR

vzhukovs: @benoitf new changes are in place

benoitf (commented):
I'm wondering why we need to have an object being returned that provides the value and an optional error message vs the method returning the value or throw an error ?

I mean, why we can't rely on catching an error vs returning an object ?

vzhukovs:
> I'm wondering why we need to have an object being returned that provides the value and an optional error message vs the method returning the value or throw an error ?
> 
> I mean, why we can't rely on catching an error vs returning an object ?

Because, we will have situation, when the method will return `true` when port is available and `false` when some check was not passed and not caused by any exception, but here we have to know what was the reason for that. Also, then we can catch some system exception (like `EACCESS` or `EADDRINUSE` or some else) and we will have the situation, when this scenario will be the same as return `false` but on the client side we have to catch this exception. 

It's just to simplify the processing the final checking result. Because when the port is not available, there should be a reason why. WDYT?

deboer-tim:
I'd say the same thing as @benoitf: there doesn't seem to be a need for a new type to return both a boolean and a message: either the port is free and we need no message or it's not and we need a message. We don't need to expose any underlying errors directly - we can catch them and `throw new Error('Port must be...')` to give our own more useful messages where the code currently does.

vzhukovs:
@deboer-tim @axel7083 new changes are in place

packages/main/src/plugin/util/port.ts
```diff
@@ -60,20 +63,33 @@ function isFreeAddressPort(address: string, port: number): Promise<boolean> {
...
 
 export async function isFreePort(port: number): Promise<boolean> {
-  const intfs = getIPv4Interfaces();
-  // Test this special interface separately, or it will interfere with other tests done in parallel
-  if (!(await isFreeAddressPort('0.0.0.0', port))) {
-    return false;
+  if (isNaN(port) || port > 65535) {
+    throw new Error('The port must have an integer value within the range from 1025 to 65535.');
+  } else if (port < 1024) {
+    throw new Error('The port must be greater than 1024.');
```
axel7083, line 84: Nice having custom message 👍

axel7083 (approved):
LGTM

deboer-tim (approved):
Approved.

Side note: I wonder if we should use NumberInput here (in another PR), would block non-numerical entry and provide +/- for small changes.

vzhukovs:
> Approved.
> 
> Side note: I wonder if we should use NumberInput here (in another PR), would block non-numerical entry and provide +/- for small changes.

Submitted an issue for this proposal: https://github.com/containers/podman-desktop/issues/7198

PR #7082, opened 2024-05-03
fix: provider card UI on dashboard

### LINKED ISSUE #6973: Dashboard podman row is misaligned to the right
### Bug description

It seems that the Podman section on the dashboard is misaligned to the right, making all the buttons not be centered.

Screenshot 2024-04-26 at 10 51 36
Screenshot 2024-04-26 at 10 55 29

### Operating system

macOS, win

### Installation Method

### Version

1.10.0

### Steps to reproduce

1. Install PD 1.10
2. Go to the Dashboard.
3. Notice section is not aligned to the center.

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR updates the provider cards UI to solve the misalignment issue  https://github.com/containers/podman-desktop/issues/6973

### Screenshot / video of UI

I've recorded a video where i cover almost all states.
Install podman, initialize and start, then delete the machine and initialize it + manual start. And i shrink/enlarge the window to show the different elements position.

The initialize step takes a lot, so skip it. I did not cut the video 😆 
https://drive.google.com/file/d/1YYZz5QzkBsiIm5xeRcX_PBs_tbE-zjuA/view?usp=drive_link

it resolves #6973 

### How to test this PR?

1. change state of your provider and see if it looks nice in the dashboard

### DIFF
--- a/packages/renderer/src/lib/dashboard/ProviderCard.svelte
+++ b/packages/renderer/src/lib/dashboard/ProviderCard.svelte
@@ -26,11 +26,15 @@ export let provider: ProviderInfo;
         <div class="flex flex-row" aria-label="Actual State">
           <ProviderStatus status="{provider.status}" />
         </div>
+        <div class="my-3 empty:my-0 w-full">
+          <slot name="update" />
+        </div>
       </div>
     </div>
     <div class="flex items-center flex-row space-x-10 mt-5 w-full lg:mt-0 lg:w-3/4 flex-nowrap">
       <slot name="content" />
     </div>
   </div>
+
   <ProviderLinks provider="{provider}" />
 </div>
--- a/packages/renderer/src/lib/dashboard/ProviderCard.svelte
+++ b/packages/renderer/src/lib/dashboard/ProviderCard.svelte
@@ -11,8 +11,8 @@ export let provider: ProviderInfo;
   class="flex bg-charcoal-800 rounded-md p-5 gap-3 flex-col flex-nowrap"
   role="region"
   aria-label="{provider.name} Provider">
-  <div class="lg:flex lg:flex-row gap-x-4">
-    <div class="grid grid-cols-[3rem_1fr] w-min gap-2">
+  <div class="flex flex-col lg:flex-row gap-x-4">
+    <div class="grid grid-cols-[3rem_1fr] w-1/4 gap-2">
       <IconImage image="{provider?.images?.icon}" class="mx-0 max-h-12" alt="{provider.name}"></IconImage>
       <div class="flex flex-col gap-0 text-gray-400 text-lg whitespace-nowrap" aria-label="context-name">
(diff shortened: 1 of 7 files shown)


deboer-tim:
Scaling generally works better with this, but it causes overlap for OpenShift Local in some states, e.g.:
Screenshot 2024-05-03 at 1 01 41 PM

lstocchi:
> Scaling generally works better with this, but it causes overlap for OpenShift Local in some states

Ah good catch. I think it would be the same on podman if i have an old version that requires an update. Gonna give this a look. Thanks!!

lstocchi:
@ekidneyrh @deboer-tim My idea is to move the update button to prevent those kind of problems.
WDYT about these 2 options? Or do you have any other idea?
1. Adding it just below the name
image
image


2. Create a new row just above the link with the button centered
image
image


For the small screen maybe we can think about moving the title to the center

axel7083:
> @ekidneyrh @deboer-tim My idea is to move the update button to prevent those kind of problems. WDYT about these 2 options? Or do you have any other idea?

Could you include same screenshots with different podman-desktop size to see how the design is responding to different windows sizes ?

lstocchi:
> > @ekidneyrh @deboer-tim My idea is to move the update button to prevent those kind of problems. WDYT about these 2 options? Or do you have any other idea?
> 
> Could you include same screenshots with different podman-desktop size to see how the design is responding to different windows sizes ?

Added. I didn't make any other modification. But, maybe, we should move the title to the center

axel7083:
> > > @ekidneyrh @deboer-tim My idea is to move the update button to prevent those kind of problems. WDYT about these 2 options? Or do you have any other idea?
> > 
> > 
> > Could you include same screenshots with different podman-desktop size to see how the design is responding to different windows sizes ?
> 
> Added. I didn't make any other modification. But, maybe, we should move the title to the center

Thanks!

deboer-tim:
> @ekidneyrh @deboer-tim My idea is to move the update button to prevent those kind of problems. WDYT about these 2 options? Or do you have any other idea?

Fundamentally I think the problem is that we have four columns (name, status description, install button, update button) which is just too many. Long term I'd prefer getting rid of one :) , but short term I agree that essentially making it wrap at the equivalent of 2-3 columns makes the most sense.

Out of these options I prefer 1 since it keeps the update button closer to the version, but maybe removing a bit of the top/bottom padding around the button to keep things a little closer together.

lstocchi:
Updated by moving the update button to the left and reducing the padding/margin.
Full-width
image

Shrinked
image

deboer-tim (approved):
I was still able to get the text to overlap - but I had to try really hard now. I think we're going to have diminishing returns until a bigger overhaul of what's on these tiles so let's merge and come back to this then. Well done Luca.

benoitf (approved):

PR #7085, opened 2024-05-03
feat: extensions should stay disabled after restart

### LINKED ISSUE #268: Extensions automatically reenable
### Bug description

If I disable an extension, it is automatically re-enabled the next time I start. These preferences should be saved across restarts.

### Operating system

macOS 12.4

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Disabled extensions should not start when you restart Podman Desktop.

Adds a new 'extensions.disabled' configuration property that stores an array with the ids of disabled extensions. If an extension is disabled it is read from disk but never activated.

To be able to only store/clear disabled extensions we need to know when the user is intentionally enabling or disabling an extension (vs the extension's normal start/stop with Podman Desktop). startExtension() already exists but stopExtension() redirected to deactivateExtension(), so I created a new stopExtension().

stopExtension() disables the extension, and both start and removeExtensionPerUserRequest (for uninstall/reinstall) make sure the extension is no longer disabled.

### Screenshot / video of UI

N/A

Fixes #268.

### How to test this PR?

Disable a couple extensions and restart, make sure they're disabled.

Disable, uninstall, then reinstall and make sure extension is running.

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -284,9 +284,9 @@ export class ExtensionLoader {
 
   getDisabledExtensionIds(): string[] {
     return (
-      (this.configurationRegistry
+      this.configurationRegistry
         .getConfiguration(ExtensionLoaderSettings.SectionName)
-        .get(ExtensionLoaderSettings.Disabled) as string[]) || []
+        .get<string[]>(ExtensionLoaderSettings.Disabled, [])
     );
   }
 
@@ -715,7 +715,7 @@ export class ExtensionLoader {
 
         await this.activateExtension(extension, runtime);
       } else {
-        console.log('Not activating disabled extension (' + extension.id + ')');
+        console.log(`Extension (${extension.id}) not activated because it is disabled`);
       }
     } catch (err) {
       this.extensionState.set(extension.id, 'failed');
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -266,7 +266,49 @@ export class ExtensionLoader {
       },
     };
 
-    this.configurationRegistry.registerConfigurations([maxActivationTimeConfiguration]);
+    const disabledExtensionConfiguration: IConfigurationNode = {
+      id: 'preferences.extensions',
+      title: 'Extensions',
+      type: 'object',
+      properties: {
+        [ExtensionLoaderSettings.SectionName + '.' + ExtensionLoaderSettings.Disabled]: {
+          description: 'Disabled extensions',
+          type: 'array',
+          hidden: true,
+        },
+      },
+    };
+
+    this.configurationRegistry.registerConfigurations([maxActivationTimeConfiguration, disabledExtensionConfiguration]);
+  }
+
+  getDisabledExtensionIds(): string[] {
+    return (
+      (this.configurationRegistry
+        .getConfiguration(ExtensionLoaderSettings.SectionName)
+        .get(ExtensionLoaderSettings.Disabled) as string[]) || []
+    );
+  }
+
+  setDisabledExtensionIds(disabledExtensionIds: string[]): void {
+    this.configurationRegistry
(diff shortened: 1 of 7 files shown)


axel7083 (approved):
LGTM, working nicely on my computer, I can finally keep docker extension stopped!

benoitf:
hi, it's not clear to me what will happen in case of install/reinstall if for example extension has been manually removed.
I'm not sure it will goes into the ensureExtensionIsEnabled mode ?

deboer-tim:
> hi, it's not clear to me what will happen in case of install/reinstall if for example extension has been manually removed. I'm not sure it will goes into the ensureExtensionIsEnabled mode ?

If the user manually removes through our UI this calls removeExtensionPerUserRequest(), and line 1546 ensures that the extension is no longer disabled so they can reinstall and it'll work. If they manually find and remove the extension from the filesystem then we don't clean up and fresh install would be stopped until they start it.

benoitf:
so, should we also reset the flag when we do an installation ?

deboer-tim:
> so, should we also reset the flag when we do an installation ?

Yes, that would cover 'external uninstall then reinstall' as well.

Rebased to pick up changes (extension start telemetry) and added a line to extension-installer.ts to cover the install case.

packages/main/src/plugin/extension-loader.ts
```diff
@@ -266,7 +266,49 @@ export class ExtensionLoader {
...
+    this.configurationRegistry.registerConfigurations([maxActivationTimeConfiguration, disabledExtensionConfiguration]);
+  }
+
+  getDisabledExtensionIds(): string[] {
+    return (
+      (this.configurationRegistry
+        .getConfiguration(ExtensionLoaderSettings.SectionName)
+        .get(ExtensionLoaderSettings.Disabled) as string[]) || []
+    );
+  }
```
benoitf: AFAIK seems like we could use the cast on the method as it's relying on generics and use the parameter for default value if missing the value

`get<string[]>(extensionLoaderSettings.Disabled, [])`


```suggestion
  getDisabledExtensionIds(): string[] {
    return (
      (this.configurationRegistry
        .getConfiguration(ExtensionLoaderSettings.SectionName)
        .get<string[]>(ExtensionLoaderSettings.Disabled, [])
    );
  }
```

packages/main/src/plugin/extension-loader.ts
```diff
@@ -664,14 +706,17 @@ export class ExtensionLoader {
...
-      const runtime = this.loadRuntime(extension);
-      const afterLoadingRuntime = performance.now();
+        telemetryOptions['loadingRuntimeDuration'] = afterLoadingRuntime - beforeLoadingRuntime;
 
-      telemetryOptions['loadingRuntimeDuration'] = afterLoadingRuntime - beforeLoadingRuntime;
-
-      await this.activateExtension(extension, runtime);
+        await this.activateExtension(extension, runtime);
+      } else {
+        console.log('Not activating disabled extension (' + extension.id + ')');
```
```suggestion
        console.log(`Not activating disabled extension ('${extension.id})`);
```

also should it be more with the id first like
`${extensionId} is not being activated as it is disabled`

packages/main/src/plugin/extension-loader.ts
deboer-tim: Fixed in new commit.

packages/main/src/plugin/extension-loader.ts
deboer-tim: Fixed. I took a look and all the loader messages are like:
  `blah extension (qualified.name) blah`
and all messages from the extension are like:
  `[name] blah`
so I went with:
  `Extension (${extension.id}) not activated because it is disabled`
to be most consistent with the current loader messages.

(in the future maybe we should change these to be a bit more similar. We still need to clear between what's from the loader and what's from the extension, but would be nice to easily scan for everything related to extension X)

benoitf (approved):

PR #7110, opened 2024-05-06
fix: flaky test port availability 

### LINKED ISSUE #7109: flaky test with ports
### Bug description

image

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

Pipeline actions

### Relevant log output

```shell
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/plugin/util/port.spec.ts > return first empty port, no port is used
AssertionError: expected 20001 to be 20000 // Object.is equality

- Expected
+ Received

- 20000
+ 20001

 ❯ src/plugin/util/port.spec.ts:86:20
     84|   const freePort = await port.getFreePort(20000);
     85| 

     86|   expect(freePort).toBe(20000);
       |                    ^
     87|   expect(await port.isFreePort(freePort)).toBe(true);
     88| });

Error: AssertionError: expected 20001 to be 20000 // Object.is equality

- Expected
+ Received

- 20000
+ 20001

 ❯ src/plugin/util/port.spec.ts:86:20
```

### Additional context

The problem come certainly from the tests running in parallel, other tests are also using the `20000`

https://github.com/containers/podman-desktop/blob/35e4265ba6054d7dad85d194d2d1a6e4251e01d2/packages/main/src/plugin/util/port.spec.ts#L114

axel7083:
### What does this PR do?

The test was checking for port availability, but as the port was fixed and hardcoded, sometime it was interfering with other tests using the same port.

This PR change the port and add a random value between `0-100` to ensure the port will not conflict with other tests.

Fixes https://github.com/containers/podman-desktop/issues/7109

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/util/port.spec.ts
+++ b/packages/main/src/plugin/util/port.spec.ts
@@ -81,9 +81,10 @@ test.each(hosts)(
 );
 
 test('return first empty port, no port is used', async () => {
-  const freePort = await port.getFreePort(20000);
+  const start = 21000 + Math.floor(Math.random() * 100);
+  const freePort = await port.getFreePort(start);
 
-  expect(freePort).toBe(20000);
+  expect(freePort).toBe(start);
   expect(await port.isFreePort(freePort)).toBe(true);
 });
 


deboer-tim (approved):
This is definitely much better, so I am approving... it feels odd though as this doesn't eliminate the problem, it just makes it far, far less likely to occur.

Since there could always be other processes running and using whatever port we pick there are lots of other things we could try (e.g. if it fails, wait 2s and try again with another random port), but may not even be worth the effort if this is very rare, so take my +1.

axel7083:
> This is definitely much better, so I am approving... it feels odd though as this doesn't eliminate the problem, it just makes it far, far less likely to occur.
> 
> Since there could always be other processes running and using whatever port we pick there are lots of other things we could try (e.g. if it fails, wait 2s and try again with another random port), but may not even be worth the effort if this is very rare, so take my +1.

I agree that this does not fix the problem, but I also rather have a very unlikely possibilities than seeing too often the pipeline fail! ahah

lstocchi (approved):

Issue #6929, opened 2024-04-24
Next batch of colors in light mode palette


deboer-tim:
### Is your enhancement related to a problem? Please describe

We need the next batch of colors for light mode.

### Describe the solution you'd like

We will meet and figure out how we want to work on this, but we need to define the next batch of colors for light mode.

The coming components/pages in rough order:

- [x] Select component → [+/- Button](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#--button) 
- [x] NavPage (main pages, mostly background colors) → [Nav page theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#nav-page)
- [x] Form pages → [Form page theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#form-page)
- [x] Details page → [Details page theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#details-pages)
- [x] Tables → [Table / Nav page theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#nav-page)
- [x] Dropdown → [Dropdown theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#dropdown)
- [x] Toggle → [Toggle theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#toggle)
- [x] Preferences tiles → [Preference page / inverted content theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#inverted-main-content)
- [x] Tooltips → [Tooltip theming](https://github.com/containers/podman-desktop/wiki/Design-System-Theming#tooltip)

ekidneyrh:
I'm unsure what you mean exactly by NavPage, but is this what you're referring to? I refer to it as the main content theming and then the invert content theming for when you are in the settings.

Content Theming:

[Content Theming.pdf](https://github.com/containers/podman-desktop/files/15198682/Content.Theming.pdf)

![Image](https://github.com/containers/podman-desktop/assets/87311656/271eb453-3274-4a02-b3cd-bed00e7f3345)

Invert Content Theming:

[Invert Content Theming.pdf](https://github.com/containers/podman-desktop/files/15198700/Invert.Content.Theming.pdf)

![Image](https://github.com/containers/podman-desktop/assets/87311656/b11e5b48-84ce-4a36-90f1-f1281b37c7b6)

deboer-tim:
[...]

PR #7119, opened 2024-05-07
feat: add content theming color registry

feloy:
### What does this PR do?

Define content theming colors registry, with colors defined in https://github.com/containers/podman-desktop/issues/6929#issuecomment-2092726542

Part of #6929 

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -227,6 +227,7 @@ export class ColorRegistry {
     this.initGlobalNav();
     this.initSecondaryNav();
     this.initTitlebar();
+    this.initContent();
     this.initInvertContent();
     this.initCardContent();
     this.initInputBox();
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/color-registry.ts
```diff
@@ -411,6 +412,64 @@ export class ColorRegistry {
...
+
+    this.registerColor(`${ct}card-icon`, {
+      dark: colorPalette.gray[400],
+      light: colorPalette.purple[900],
+    });
+
+    this.registerColor(`${ct}divider`, {
+      dark: colorPalette.charcoal[400],
+      light: colorPalette.gray[700],
+    });
```
feloy, line 470: This one is not defined in the PDF referenced in the PR description, but is discussed here: https://github.com/containers/podman-desktop/issues/6929#issuecomment-2097884396

The name `divider` is a proposal, open to discussion

ekidneyrh (approved):

feloy:
@ekidneyrh `content-header-icon` in light mode is defined as `gray-300` in the PDF, but icons would not be visible with background being also `gray-300`. In the light-mode illustration, the icon seems to be some purple instead. What do you think?

ekidneyrh:
> @ekidneyrh `content-header-icon` in light mode is defined as `gray-300` in the PDF, but icons would not be visible with background being also `gray-300`. In the light-mode illustration, the icon seems to be some purple instead. What do you think?

It should be purple-700, apologies for that.

deboer-tim (approved):

benoitf (approved):

PR #7125, opened 2024-05-07
chore(Modal): use tailwindcss classes instead of css

axel7083:
### What does this PR do?

Migrating `<style>` content of the Modal to tailwindcss classes before https://github.com/containers/podman-desktop/issues/6906

Related to https://github.com/containers/podman-desktop/issues/6906

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/dialogs/Modal.svelte
+++ b/packages/renderer/src/lib/dialogs/Modal.svelte
@@ -34,7 +34,7 @@ if (previously_focused) {
 <button class="fixed top-0 left-0 w-full h-full bg-[#00000099]" on:click="{close}"></button>
 
 <div
-  class="absolute top-1/2 left-1/2 z-50 rounded translate-x-[-50%] translate-y-[-50%] overflow-auto w-[calc(200vw - 4em)] max-w-[42em] max-h-[calc(100vh - 4em)]"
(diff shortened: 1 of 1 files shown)


deboer-tim (changes requested):
If you're fixing this, let's go all the way! Dialogs are supposed to fade the entire background like MessageBoxes do, and we shouldn't use hardcoded bg-[] color if we can avoid it.

Stealing from MessageBox I think you can change `bg-[#00000099]` to `bg-black bg-opacity-60 bg-blend-multiply z-50`.

Then:
Screenshot 2024-05-07 at 8 34 51 AM
becomes:
Screenshot 2024-05-07 at 8 42 19 AM

axel7083:
> If you're fixing this, let's go all the way! Dialogs are supposed to fade the entire background like MessageBoxes do, and we shouldn't use hardcoded bg-[] color if we can avoid it.
> 
> Stealing from MessageBox I think you can change `bg-[#00000099]` to `bg-black bg-opacity-60 bg-blend-multiply z-50`.

Maybe `MessageBox;svelte` should use the `Modal.svelte` component, what do you think @deboer-tim 

The `Modal` component has the following usage

image

We could use the classes from the MessageBox

deboer-tim:
> Maybe `MessageBox;svelte` should use the `Modal.svelte` component, what do you think @deboer-tim 

The original use/implementation was different enough that I don't think it made sense, but you're probably right now and would clean up some duplication - just not in this/one PR. e.g. I would still merge this, then we could clean up anything remaining in Modal/make it reusable, then migrate MessageBox.

deboer-tim (approved):

lstocchi (approved):

benoitf (approved):

PR #7135, opened 2024-05-08
feat: allow to select a folder using the fileComponent

lstocchi:
### What does this PR do?

It extends the fileComponent to support selecting a folder instead of a file.
It can be used by a setting, like https://github.com/containers/podman-desktop-extension-ai-lab/pull/1069

### Screenshot / video of UI

select_folder_setting

it is needed by https://github.com/containers/podman-desktop-extension-ai-lab/pull/1069

### How to test this PR?

1. run with https://github.com/containers/podman-desktop-extension-ai-lab/pull/1069
2 run tests

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.spec.ts
@@ -166,6 +166,25 @@ test('Expect an input button with Browse as placeholder when record is type stri
   expect(input.textContent).toBe('Browse ...');
 });
 
+test('Expect an input button with Browse as placeholder when record is type string and format folder', async () => {
+  const record: IConfigurationPropertyRecordedSchema = {
+    title: 'record',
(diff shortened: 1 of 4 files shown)


deboer-tim:
Before Mo left there was a design for a FileInput that I was also brushing off this week:

Screenshot 2024-05-08 at 8 12 45 AM

It supports file/folder selection and is also used by FileItem, but has at least one UX issue. I think it is compatible with this PR, but maybe let's have a quick chat with @ekidneyrh / UX to just make sure we agree on direction?

benoitf (approved):
changing the UI should be part of another PR anyway, here it's just handling another condition

deboer-tim:
> changing the UI should be part of another PR anyway, here it's just handling another condition

Yeah, I just don't love adding more function and tests on something that is supposed to change, making it harder to make the change w/o breaking things. I've tried porting the tests though and it isn't too bad, so I will approve.

deboer-tim (approved):

Issue #7142, opened 2024-05-08
Unable to change configuration (phantom resets)


deboer-tim:
### Bug description

For a while now I've noticed that every once in a while when I change a configuration value... it changes back. At first I thought maybe I'd missed the button or mis remembered, but eventually I figured out it was indeed happening.

Most of the time it's rare, but today I hit it enough that it was very annoying and it took several attempts to change some preferences. Font size is 10, I change it to 9, it goes back to 10. I try again.. it goes back to 10.

### Operating system

macOS 14.4

### Installation Method

None

### Version

next (development version)

### Additional context

I spent way too long trying to figure this out (especially since I didn't actually find a fix) but I narrowed down what's causing it: the afterUpdate() trigger in PreferencesRenderingItem is happening all the time - sometimes not for a couple seconds, but about once a second, and indefinitely. It resets the entire configuration record. If you make a change and it's saved between two of these calls, everything is fine. If it happens just when you clicked, your change is overwritten and lost.

odockal:
@deboer-tim I am also sometimes seeing preference value going back to default, ie. this is the most common on Mac when I switch Exit on Close. I am always confused since I feel like I haven't clicked the mouse properly so I need to repeat. Sometimes even twice.

cdrage:
Re-opening this as unfortunatley this change causes an issue where (most) of the settings no longer show up :( 

![image](https://github.com/user-attachments/assets/01f3db7a-e108-40ce-8d73-1347a7ea93fc)

PR #7143, opened 2024-05-08
chore: number input

deboer-tim:
### What does this PR do?

The current number editor in Settings > Preferences works, but it has a few drawbacks:
 - doesn't match current design, and would need updating to light mode.
 - periodically has some odd 'resets' to a previous number. (this isn't just the input - figured out later and opened #7142)
 - has two error modes - underline for below minimum, icon to the left for invalid value.
 - after invalid entry (e.g. below minimum), incrementing does a string add vs numerical.
 - not easily reusable for editing numbers in other places.

This PR creates a NumberInput component that is just an Input with +/- buttons. Key input is restricted to numbers, and increment/decrement are used to disable +/- appropriately. If the user manually enters a number out of bounds, the Input's error is used to display a short warning.

Input required some minor changes to support this:
 - allow inputClass in order to center numbers in the textbox.
 - passthrough of key events, in order to block non-numerical entry.
 - boolean to block showing the error in a separate span. Although this is the design for inputs, it felt really out of place in preferences.
 - reduce padding slightly - it didn't matter in large Inputs, but number inputs tend to be smaller and extraneous padding was obvious.

New component is used in Settings > Preferences and max retry count for running an image.

### Screenshot / video of UI

Before:

https://github.com/containers/podman-desktop/assets/19958075/697badfb-9871-4e18-87c6-bc92edf2ebe6

After:

https://github.com/containers/podman-desktop/assets/19958075/b9755ced-6367-4a2f-a114-2e2e381914be

Another part of #3234.

### How to test this PR?

Tests added. NumberItem test had to be changed since only numerical inputs work now.

Try changing some preferences. Be aware of #7142 (separate issue) if the value resets.

### DIFF
--- a/packages/renderer/src/lib/image/RunImage.svelte
+++ b/packages/renderer/src/lib/image/RunImage.svelte
@@ -17,6 +17,7 @@ import type { ContainerInfoUI } from '../container/ContainerInfoUI';
 import { splitSpacesHandlingDoubleQuotes } from '../string/string';
 import ErrorMessage from '../ui/ErrorMessage.svelte';
 import FormPage from '../ui/FormPage.svelte';
+import NumberInput from '../ui/NumberInput.svelte';
 import Tab from '../ui/Tab.svelte';
 import type { ImageInfoUI } from './ImageInfoUI';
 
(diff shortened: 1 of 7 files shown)


lstocchi (approved):
Played a bit with it and looks good. Once it didn't save my value correctly but i was not able to replicate it again. I was typing random stuff so it is not a normal use case.

LGTM, nice improvement 🚀

benoitf (approved):

PR #7149, opened 2024-05-10
Light mode for NavPage & FormPage

### LINKED ISSUE #6889: Light mode for NavPage & FormPage

We need to support light mode on these two pages, since they're used extensively.

Define missing bg/tile/divider colors in the light mode palette and use them in these pages (bg color likely one page up).

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

Adds support to light mode on NavPage & FormPage components

#### NavPage

light-navpage

#### FormPage

light-formpage

Fixes #6889

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -99,8 +99,8 @@ window.events?.receive('navigate', (navigationRequest: unknown) => {
 
       <div
         class="flex flex-col w-full h-full overflow-hidden"
-        class:bg-charcoal-700="{!meta.url.startsWith('/preferences')}"
-        class:bg-charcoal-800="{meta.url.startsWith('/preferences')}">
+        class:bg-[var(--pd-content-bg)]="{!meta.url.startsWith('/preferences')}"
+        class:bg-[var(--pd-invert-content-bg)]="{meta.url.startsWith('/preferences')}">
         <TaskManager />
         <SendFeedback />
         <ToastHandler />
--- a/packages/renderer/src/lib/container/ContainerExport.svelte
+++ b/packages/renderer/src/lib/container/ContainerExport.svelte
@@ -89,7 +89,7 @@ async function exportContainer() {
       <i class="fas fa-download fa-2x" aria-hidden="true"></i>
     </svelte:fragment>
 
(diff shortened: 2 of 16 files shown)


axel7083 (approved):
Nice!

ekidneyrh (approved):
Nice! I'm still working on what colours to work for the missing components on here, but I'll update asap on the tracking ticket :)

deboer-tim:
In the current design, forms have an intentional color change between the header area and the content. These color values remove that. @ekidneyrh is this an intentional design change? If so, do we need to keep a separate -bg color? The area between the title and start of form also looks big (since it's effectively double-padded), should we reduce this?

Screenshot 2024-05-10 at 8 05 35 AM

ekidneyrh:
Nav Page:
[Nav Page.pdf](https://github.com/containers/podman-desktop/files/15307995/Nav.Page.pdf)
NavPage

Form Page:
Form Page
[Content Theming.pdf](https://github.com/containers/podman-desktop/files/15308032/Content.Theming.pdf)


Yah I'm in favour of getting rid of of the extra padding and use the same background for the header and the rest of the form. I don't think there's a reason to have two different colours.

Is there any values im missing? I know there is some overlap with the image and volumes pages. I dont think there's a need to have a bg colour for the table-body-label-text on the nav pages.

feloy:
> Yah I'm in favour of getting rid of of the extra padding and use the same background for the header and the rest of the form. I don't think there's a reason to have two different colours.
> 

I have added a commit to remove the top padding of the forms using the FormPage component

feloy:
> Nav Page: [Nav Page.pdf](https://github.com/containers/podman-desktop/files/15307995/Nav.Page.pdf) NavPage
> 

The `table-body-label-text` defined in https://github.com/containers/podman-desktop/pull/7151 was intended for the text `Kubernetes`/`Podman` in the `Environment` column (inside a label). You are reusing this name for the text of the header (which was defined by `table-header-text` in #7151).  I'm ok to use `table-body-label-text` for the header text, but in this case it will be missing the color for the text in the Environment column.

deboer-tim (approved):

PR #7150, opened 2024-05-10
feat: add tip slot to customize tip content

lstocchi:
### What does this PR do?

This PR adds a tip slot so that one can customize the tip content.
In AI Lab we can use it to show the content multiline instead of a single line

### Screenshot / video of UI

An example is the output i want to have in AI Lab.
tip_custom

N/A

### How to test this PR?

run tests

### DIFF
--- a/packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
@@ -27,7 +27,7 @@ function getConditionColour(type: string): string {
 <div class="flex flex-row gap-1">
   {#each object.conditions as condition}
     <Tooltip bottom>
-      <svelte:fragment slot="item">
+      <svelte:fragment slot="content">
         <div class="flex flex-row bg-charcoal-500 items-center p-1 rounded-md text-xs text-gray-500">
           <div class="w-2 h-2 {getConditionColour(condition.type)} rounded-full mr-1"></div>
           {condition.type}
--- a/packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
@@ -26,11 +26,18 @@ function getConditionColour(type: string): string {
 
 <div class="flex flex-row gap-1">
   {#each object.conditions as condition}
-    <Tooltip tip="{condition.message}" bottom>
-      <div class="flex flex-row bg-charcoal-500 items-center p-1 rounded-md text-xs text-gray-500">
-        <div class="w-2 h-2 {getConditionColour(condition.type)} rounded-full mr-1"></div>
-        {condition.type}
-      </div>
+    <Tooltip bottom>
+      <svelte:fragment slot="item">
+        <div class="flex flex-row bg-charcoal-500 items-center p-1 rounded-md text-xs text-gray-500">
+          <div class="w-2 h-2 {getConditionColour(condition.type)} rounded-full mr-1"></div>
+          {condition.type}
+        </div>
+      </svelte:fragment>
+      <svelte:fragment slot="tip">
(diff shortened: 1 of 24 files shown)


axel7083 (commented):
I am wondering if we should not use a <slot/> so we could put whatever we want without being limited to text

deboer-tim:
I assume you meant 'slot'. If so, +1. We may want more interesting tooltips in the future (e.g. include image or link) so slot would give more flexibility.

lstocchi:
Not sure to get it. Do you want something like?

```
<div class="relative inline-block">
  <span class="group tooltip-slot {$$props.class}">
    <slot />
  </span>
  <div
    class="whitespace-nowrap absolute tooltip opacity-0 inline-block transition-opacity duration-150 ease-in-out pointer-events-none z-[10]"
    class:left="{left}"
    class:right="{right}"
    class:bottom="{bottom}"
    class:top="{top}"
    class:top-left="{topLeft}"
    class:top-right="{topRight}"
    class:bottom-left="{bottomLeft}"
    class:bottom-right="{bottomRight}">
    <slot name="tip" />
  </div>
</div>
```

benoitf (approved):

lstocchi:
@benoitf @axel7083 updated

benoitf (approved):

packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
```diff
@@ -26,11 +26,20 @@ function getConditionColour(type: string): string {
 
 <div class="flex flex-row gap-1">
   {#each object.conditions as condition}
-    <Tooltip tip="{condition.message}" bottom>
-      <div class="flex flex-row bg-charcoal-500 items-center p-1 rounded-md text-xs text-gray-500">
-        <div class="w-2 h-2 {getConditionColour(condition.type)} rounded-full mr-1"></div>
-        {condition.type}
-      </div>
+    <Tooltip bottom>
+      <svelte:fragment slot="item">
```
axel7083: could we use something like `content` instead of `item` ?

packages/renderer/src/lib/kube/KubeEditYAML.svelte
```diff
@@ -79,21 +79,35 @@ async function applyToCluster() {
...
   role="group"
   aria-label="Edit Buttons">
-  <Tooltip tip="Apply the changes to the cluster, similar to 'kubectl apply'" topLeft="{true}">
-    <Button
-      type="primary"
-      aria-label="Apply changes to cluster"
-      on:click="{applyToCluster}"
-      disabled="{!changesDetected}"
-      inProgress="{inProgress}">Apply changes to cluster</Button>
+  <Tooltip topLeft="{true}">
```
axel7083:
```suggestion
  <Tooltip topLeft>
```

packages/renderer/src/lib/kube/KubeEditYAML.svelte
```diff
@@ -79,21 +79,35 @@ async function applyToCluster() {
...
+    </svelte:fragment>
   </Tooltip>
-  <Tooltip tip="Revert the changes to the original content" topLeft="{true}">
-    <Button
-      type="secondary"
-      aria-label="Revert changes"
-      class="mr-2 opacity-100"
-      on:click="{revertChanges}"
-      disabled="{!changesDetected}">Revert changes</Button>
+  <Tooltip topLeft="{true}">
```
axel7083:
```suggestion
  <Tooltip topLeft>
```

packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
lstocchi: done

axel7083 (approved):
LGTM, thanks for taking the time to make the change! 🚀

PR #7163, opened 2024-05-10
chore: remove test containerfile

cdrage:
chore: remove test containerfile

### What does this PR do?

Removes an old bootc container file that is no longer needed / should
not be referenced (uses an old FROM, as well as very insecure RUN
command)

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

N/A

### DIFF
--- a/tests/playwright/resources/bootable-containerfile
+++ b/tests/playwright/resources/bootable-containerfile
@@ -1,4 +1,0 @@
-FROM quay.io/centos-bootc/fedora-bootc:eln
-
-# Change the root password
-RUN echo "root:supersecret" | chpasswd


deboer-tim (approved):

cbr7 (changes requested):
Why are you removing all three files? Only the bootable-containerfile should be removed, the other two are in use in tests.

cdrage:
> Why are you removing all three files? Only the bootable-containerfile should be removed, the other two are in use in tests.

Woops, I removed the directory instead of just the single file by mistake. Updated

cbr7 (approved):

benoitf (approved):

Issue #4914, opened 2023-11-21
Light Mode Support


deboer-tim:
### Epic domain

We have had requests for light mode for some time for obvious reasons: astigmatism, system default, or just users who prefer it better. Opening this epic to track the overall work.

### Additional context

Initial request: #576 (maybe can be the 'final' issue we resolve to expose the mode).

Technical implementation:

- [x] #4887 - initial PR to add hidden light mode setting
- [x] #4907 - discussion on mapping colors
- [ ] #4912 - draft to try things out
- [ ] #5116 - remove hardcoded references to dark icons/add light support
- [x] #5878 - Create abstract color references to better enable theming

Sprint 256:
- [x] #6929 - Define colors in palette
- [x] #6889 - Main nav/form page components
- [x] #6890 - Images and volumes tables
- [x] #6924

Sprint 257:
- [x] #7168
- [x] #6925
- [x] #6927
- [x] #7215
- [x] #7183
- [x] #7184
- [x] #7213
- [x] #7362 

Sprint 258:
- [x] #6926
- [x] #7185
- [x] #7214
- [x] #7348 
- [x] #7236
- [x] #7212
- [x] #7346
- [x] #7345
- [x] #7574
- [x] #7575
- [x] #7592
- [x] #7579
- [x] #7718
- [x] #7708
- [x] #7727 

Sprint 259:
- [x] #7547
- [x] #7548
- [x] #7544
- [x] #7608
- [x] #7728 
- [x] #7724
- [x] #6199
- [x] #7936

Sprint 260:
- [x] #7343
- [x] #7344
- [x] #7998
- [x] #7546 
- [x] #7545 
- [x] #8001
- [x] #8077
- [x] #8111
- [x] #8140
- [x] #8202
- [x] #8217
- [x] #8234
- [x] #8201
- [x] #8203

Sprint 261
- [x] #8380
- [x] #8397
- [x] #8393
- [x] #8447
- [x] #8440

Sprint 262
- [x] #8734
- [x] #8411
- [x] #8477
- [x] #8587

benoitf:
@deboer-tim : maybe we change #576 to be the epic as most of the people are following this issue ?

deboer-tim:
I almost did that, then thought most of those users just want to know when it is 'planned' and 'available' and don't care as much about the overall status or steps to get there. If you think it's better we can move to #576 and close this one.

benoitf:
no I'm fine, you're right

[...]

PR #7165, opened 2024-05-10
chore: checkbox light mode

deboer-tim:
### What does this PR do?

Add entries to the color registry for checkbox, with support for light mode. Light mode colors are the same for now since I couldn't find a precedent, except for unchecked.

Took this opportunity to add focus (hover) states since they were missing and didn't match other UI elements. For unchecked the outline becomes purple, for all other states it's one step lighter for dark mode and darker for light mode.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/19958075/e61f5f1d-b86e-4142-973c-1c698fbfce47

Part of #4914.

### How to test this PR?

View checkboxes on main pages; tests added.

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -231,6 +231,7 @@ export class ColorRegistry {
     this.initInvertContent();
     this.initCardContent();
     this.initInputBox();
+    this.initCheckbox();
   }
 
   protected initGlobalNav(): void {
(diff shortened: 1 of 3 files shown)


benoitf (approved):

ekidneyrh:
Spent some time on this this morning. I was having a look at the contrast and was thinking maybe it would be a good idea to simplify the colours and have just two main colours - one for hover and one for the default. By using less colours the hover stands out some more as we can lighten it two steps without messing the contrast. Not all the colours previous could be lightened without the contrast going below an AA rating. 

I also changed the disable colour to match the +/- button here: https://github.com/containers/podman-desktop/issues/6929#issuecomment-2098164688

[Checkbox.pdf](https://github.com/containers/podman-desktop/files/15292766/Checkbox.pdf)
Input-checkbox

WDYT?

deboer-tim:
I really like the simplification (especially getting rid of dusty purple) and have only looked at dark mode so far, but in context I find the colors too bright and somewhat distracting. The purple are lighter than the purples we use in most places, and disabled looks too close to a stop button.

I'm not a designer, but I tried playing with these pulling colors from elsewhere, and he's what I came up with.

Original:

Screenshot 2024-05-13 at 8 21 01 AM

Proposed:

Screenshot 2024-05-13 at 8 14 26 AM

Proposed with darker colors:
purple 500/600 to match Buttons (could be one step brighter IMHO), unchecked gray-400 (same as start/stop/delete buttons), and charcoal-300 (just matching what was there before).

Screenshot 2024-05-13 at 8 28 20 AM

ekidneyrh:
image
Just checking the contrast, I was trying to use AAA rating colours but AA can suffice. I think purple 400 and 500 would work better as purple-600 is too low, and I would stick with the gray-700 for the disabled tho just bc the charcoal-300 is a fail. 

I think the disabled checkbox in context with the other checkboxes on the page would be a signifier to the user what it is. If it's the only one on the page, the hover with the X would let them know it's not selectable. If it's an issue you're concerned with, we could change that box to a different disabled one, such as:
image

We could also just swap the unchecked focused and unfocused colours could make it less distracting. 

Checkbox1

deboer-tim:
Updated with these colors. My preference would still be a darker disabled color (I think a non-AAA would actually be a _good_ thing here, because we don't want disabled to stand out as much as the ones you can select), but if you can approve lets merge and we can respond later if there is feedback.

ekidneyrh (approved):

PR #7170, opened 2024-05-10
chore: expose appearance preference

### LINKED ISSUE #7362: Expose experimental appearance setting

The appearance setting to switch between light and dark mode (or platform default) is hidden, which makes it harder to test light mode and few people are aware of it enough to try it out.

Expose the setting as experimental for now, with dark still the default.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

We have four PRs in review for light mode, and hopefully only a few more until it still has issues but might be ready for experimental support/early feedback. It takes longer to test because the preference is hidden.

Creating this draft PR to make it easy to apply and test, and so that it's ready when the time comes.

### Screenshot / video of UI

Screenshot 2024-05-10 at 4 59 37 PM

Fixes #7362.

### How to test this PR?

Just confirm the setting is visible, and appearance changes when you change it.

### DIFF
--- a/packages/main/src/plugin/appearance-init.ts
+++ b/packages/main/src/plugin/appearance-init.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023, 2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


deboer-tim:
Moving out of draft, I think we're far enough along that we should get more people looking at light mode, and will be easier to test with a setting.

axel7083 (approved):

cdrage:
> Moving out of draft, I think we're far enough along that we should get more people looking at light mode, and will be easier to test with a setting.

agreed and it will help, LGTM my side

cdrage (approved):

Issue #7184, opened 2024-05-13
Light mode for modal dialogs


deboer-tim:
### Is your enhancement related to a problem? Please describe

Modal dialogs do not support light mode.

### Describe the solution you'd like

Extract colors to the registry and add light mode colors. The component structure is being cleaned up, but we possibly have several of these that need to be moved. e.g. Modal, MessageBox, making sure things like the Feedback form and Container > Create are fixed.

deboer-tim:
If you haven't started I can do the changes to Modal as part of #7187, not sure yet if that'll be all that's required or there are other cases.

feloy:
> If you haven't started I can do the changes to Modal as part of #7187, not sure yet if that'll be all that's required or there are other cases.

OK, let's make the changes in relation with #7187, I'll check if there are other places to make changes. Thanks

deboer-tim:
@ekidneyrh the core modal dialog only uses three colors today. Here are the current colors and what I've called them:
- pd-modal-fade - black (uses to fade the rest of the screen at 60% opacity and a blend-multiply)
- pd-modal-bg - charcoal-800
- pd-modal-border - charcoal-500

Let me know if you can think of a better name than 'fade', but I'm assuming the light mode equivalent for the first is safely 'white'. Could you provide light mode color values for the other two?

ekidneyrh:
> @ekidneyrh the core modal dialog only uses three colors today. Here are the current colors and what I've called them:
> 
>     * pd-modal-fade - black (uses to fade the rest of the screen at 60% opacity and a blend-multiply)
> 
>     * pd-modal-bg - charcoal-800
> 
>     * pd-modal-border - charcoal-500
> 
> 
> Let me know if you can think of a better name than 'fade', but I'm assuming the light mode equivalent for the first is safely 'white'. Could you provide light mode color values for the other two?

Yah white for pd-modal-fade looks good. For pd-model-bg I like the gray-300 used in ur PR or we could go lighter with gray-050, and for the pd-modal-boarder I would go darker and use gray-500. I'll post a screen shot of what im thinking

ekidneyrh:
![modal-light](https://github.com/containers/podman-desktop/assets/87311656/26784367-6805-4b17-aa91-64f6ed54a130)

```
pd-modal-fade: white
pd-modal-bg: gray-050
pd-modal-boarder: gray-500
pd-modal-header: purple-500
pd-modal-dropdown-header: charcoal-200
[...]

PR #7196, opened 2024-05-13
chore: table column ellipsis

### LINKED ISSUE #7195: Tables: use ellipsis for narrow columns

The core Table component does not do a good job when columns are too narrow: when column headers are too wide they drift into other headers, and cells get cutoff at the column boundary.

Column headers, SimpleColumn, and DurationColumn should use ellipsis when they're too wide. When headers are too wide the column sorting should still be visible.

There are *lots* of custom column renderers that need to do the same, this should just cover the core/component level for now.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds ellipsis to the standard table components when columns are too narrow: column headers, simple columns, and duration columns. For the header a div is added so that the sort indicators are always visible, and DurationColumn can simplify by just being a SimpleColumn.

This does not cover any custom columns, of which there are many to fix in future PRs.

### Screenshot / video of UI

Before:

Screenshot 2024-05-13 at 3 55 20 PM

Used PR #7184 to make it easier to get this all in one image and screenshot is somewhat arbitrary, but note Environment header moving off to the left and Age header getting pushed to the right, months and MB cutoff.

After:

Screenshot 2024-05-13 at 3 54 42 PM

Fixes #7195.

### How to test this PR?

Images view with #7184 used for image above, but there are SimpleColumns elsewhere. If you don't have any especially long content at the moment use ctrl-+ to make the font bigger.

### DIFF
--- a/packages/ui/src/lib/table/DurationColumn.svelte
+++ b/packages/ui/src/lib/table/DurationColumn.svelte
@@ -3,6 +3,8 @@ import humanizeDuration from 'humanize-duration';
 import moment from 'moment';
 import { onDestroy, onMount } from 'svelte';
 
+import SimpleColumn from './SimpleColumn.svelte';
+
 export let object: Date | undefined;
 let duration: string = '';
 let refreshTimeouts: number[] = [];
@@ -61,6 +63,4 @@ onDestroy(() => {
 });
 </script>
 
-<div class="text-sm text-gray-700">
-  {duration}
-</div>
+<SimpleColumn object="{duration}" />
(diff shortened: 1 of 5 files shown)


lstocchi (approved):
I wonder if we should show like a tooltip with full text in such cases so if i see a cut title/value, i just have to hover the cell. Atm there is no way to manually increase the size of a column so if i'm using a small screen and i cannot actually increase the full window i have no chance to read the cut value. 

BTW the ellipsis works fine 👍

axel7083 (approved):

deboer-tim:
Rebased. @lstocchi good idea, but I will take it in follow-up PR b/c Tooltip is having issues with z-order when I tried it quickly.

PR #7206, opened 2024-05-14
chore: removing duplicates definition of KubernetesGeneratorSelector

axel7083:
### What does this PR do?

We are using a duplicate definition of KubernetesGeneratorSelector, when the one from the `@podman-desktop/api` is the same.

Would be great and would simplify https://github.com/containers/podman-desktop/issues/6300

### How to test this PR?

- Pipeline should be ✅

### DIFF
--- a/packages/main/src/plugin/api/KubernetesGeneratorInfo.ts
+++ b/packages/main/src/plugin/api/KubernetesGeneratorInfo.ts
@@ -15,7 +15,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
-import type { KubernetesGeneratorSelector } from '../kube-generator-registry.js';
+import type { KubernetesGeneratorSelector } from '@podman-desktop/api';
 
 export interface KubernetesGeneratorInfo {
   id: string;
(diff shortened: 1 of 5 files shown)


benoitf:
I'm not sure we should use external API internally

axel7083:
> I'm not sure we should use external API internally

We are already at many, many places:
https://github.com/containers/podman-desktop/blob/cfeb27546799febf3383b236e5602bbcee204970/packages/api/src/cli-tool-info.ts#L19

https://github.com/containers/podman-desktop/blob/cfeb27546799febf3383b236e5602bbcee204970/packages/api/src/image-info.ts#L19

https://github.com/containers/podman-desktop/blob/cfeb27546799febf3383b236e5602bbcee204970/packages/api/src/manifest-info.ts#L19

https://github.com/containers/podman-desktop/blob/cfeb27546799febf3383b236e5602bbcee204970/packages/api/src/provider-info.ts#L19-L28

benoitf:
I'm not sure that because someone did that, that it's the way to go

axel7083:
Closing as will keep as it is

PR #7210, opened 2024-05-14
chore: move Table/SimpleColumn to ui component

### LINKED ISSUE #6903: Implement Table/SimpleColumn UI component

benoitf:
### What does this PR do?
Moving the component as ui component

note:
- introduce the .eslintrc.json (to see window objects)
- replace setTimeout/clearTimeout by window.setTimeout/clearTimeout methods else it thinks it's NodeJS objects and we don't have NodeJS imports in ui library
- remove some any and replace them by unknown
- use Table prefix so renamed Row to TableRow and Column to TableColumn (and TableDurationColumn etc)
- need to import the types in tsconfig.json for humanize-duration library
- had to change the onClick events as with the rule  `@typescript-eslint/explicit-function-return-type` it complains about missing return type and we can't add return type in the Svelte `{}` code, for example `(event): void => something`

fixes https://github.com/containers/podman-desktop/issues/6903

### How to test this PR?


axel7083:
> * introduce the .eslintrc.json (to see window objects)

Did the same in https://github.com/containers/podman-desktop/pull/7207

packages/renderer/src/lib/deployments/DeploymentsList.svelte
```diff
@@ -1,6 +1,13 @@
 <script lang="ts">
 import { faTrash } from '@fortawesome/free-solid-svg-icons';
-import { Button } from '@podman-desktop/ui-svelte';
+import {
+  Button,
+  Table,
+  TableColumn,
+  TableDurationColumn,
+  TableRow,
+  TableSimpleColumn,
```
axel7083, line 9: Not related to this PR, but I think it might be nicer to reduce the number of imports with https://github.com/containers/podman-desktop/issues/7189

packages/ui/package.json
```diff
@@ -36,6 +36,18 @@
       "types": "./dist/button/Button.svelte.d.ts",
       "svelte": "./dist/button/Button.svelte"
     },
+    "./Table": {
```
axel7083, line 39: Not exposing Row ?

axel7083 (approved):

packages/ui/src/lib/table/Table.svelte
```diff
@@ -121,6 +124,10 @@ function setGridColumns() {
     (element as HTMLElement).style.setProperty('grid-template-columns', wid);
   }
 }
+
+function toggleFromEvent(e: CustomEvent<boolean>): void {
```
deboer-tim: I think it would be fine to just change the parameter to toggleAll()?

packages/ui/src/lib/table/Table.svelte
```diff
@@ -150,7 +157,7 @@ function setGridColumns() {
               ? 'justify-self-center'
               : 'justify-self-start'} self-center select-none"
           class:cursor-pointer="{column.info.comparator}"
-          on:click="{() => sort(column)}"
+          on:click="{sort.bind(undefined, column)}"
```
deboer-tim, line 156: I don't love this... but don't have any idea on what else to do - not worth holding up for.

deboer-tim (approved):
LGTM

packages/ui/src/lib/table/Table.svelte
benoitf: yes :-/ either we disable the lint rule or we come with another idea

packages/ui/src/lib/table/Table.svelte
benoitf: replaced toggleAll method to accept customEvent argument

packages/ui/tsconfig.json
```diff
@@ -19,5 +19,5 @@
     "allowJs": true,
     "checkJs": true
   },
-  "include": ["src/**/*.d.ts", "src/**/*.ts", "src/**/*.svelte"]
+  "include": ["src/**/*.d.ts", "src/**/*.ts", "src/**/*.svelte", "../../types/**/*.d.ts"]
```
axel7083, line 22: Why do you need the types folder ?

benoitf: it's related to this part of the description 
`
need to import the types in tsconfig.json for humanize-duration library
`

axel7083: Okey np

PR #7216, opened 2024-05-14
chore: link light mode

### LINKED ISSUE #7215: Light mode for Link

The Link component has hard-coded colors and doesn't support light mode.

Extract current colors to the registry and add light mode colors.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds light mode support to Link component, using color defined in design for #7149.

### Screenshot / video of UI

No change on dark, looks like design on light - looks awful since I couldn't find a page with a Link that supports light mode, will look much better after #7149 has merged :) :

https://github.com/containers/podman-desktop/assets/19958075/d600456f-8d11-4424-ba1a-eac408847755

Fixes #7215.

### How to test this PR?

Manually add preference: `"preferences.appearance": "light"`

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -237,6 +237,7 @@ export class ColorRegistry {
     this.initDetails();
     this.initTab();
     this.initModal();
+    this.initLink();
   }
 
   protected initGlobalNav(): void {
@@ -791,6 +792,20 @@ export class ColorRegistry {
     this.registerColor(`${modal}border`, {
       dark: colorPalette.charcoal[500],
       light: colorPalette.gray[200],
+    });
+  }
+
+  // links
+  protected initLink(): void {
+    const link = 'link';
+
+    this.registerColor(`${link}`, {
+      dark: colorPalette.purple[400],
+      light: colorPalette.purple[700],
+    });
+    this.registerColor(`${link}-hover-bg`, {
+      dark: colorPalette.white + '2',
+      light: colorPalette.black + '2',
     });
   }
 }
--- a/packages/ui/src/lib/link/Link.spec.ts
+++ b/packages/ui/src/lib/link/Link.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023,2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,9 +32,8 @@ test('Check link styling', async () => {
   // check for one element of the styling
   const link = screen.getByRole('link');
   expect(link).toBeInTheDocument();
-  expect(link).toHaveClass('text-purple-400');
-  expect(link).toHaveClass('hover:bg-white');
-  expect(link).toHaveClass('hover:bg-opacity-10');
+  expect(link).toHaveClass('text-[var(--pd-link)]');
+  expect(link).toHaveClass('hover:bg-[var(--pd-link-hover-bg)]');
   expect(link).toHaveClass('cursor-pointer');
 });
 
--- a/packages/ui/src/lib/link/Link.svelte
+++ b/packages/ui/src/lib/link/Link.svelte
@@ -29,7 +29,7 @@ function click(): void {
 <!-- svelte-ignore a11y-no-redundant-roles -->
 <!-- svelte-ignore a11y-interactive-supports-focus -->
 <a
-  class="text-purple-400 hover:bg-white hover:bg-opacity-10 transition-all rounded-[4px] p-0.5 no-underline cursor-pointer {$$props.class ||
+  class="text-[var(--pd-link)] hover:bg-[var(--pd-link-hover-bg)] transition-all rounded-[4px] p-0.5 no-underline cursor-pointer {$$props.class ||
     ''}"
   on:click="{click}"
   role="link"


deboer-tim:
@ekidneyrh I noticed in the design for #7149 that the color for the breadcrumb at the top left was defined, but this is a standalone Link component that is used everywhere we have links - dashboard, forms, deploy to Kube, etc. To support a hover effect on all background colors it currently uses 10% translucent white, so I've defined two colors: `link` and `link-bg`. The colors should be fine since the main ones come from #7149, but let me know if you prefer different names, e.g. maybe should be link-hover-bg?

benoitf (approved):

packages/main/src/plugin/color-registry.ts
```diff
@@ -543,4 +544,18 @@ export class ColorRegistry {
...
+
+  // links
+  protected initLink(): void {
+    const sNav = 'link';
+
+    this.registerColor(`${sNav}`, {
+      dark: colorPalette.purple[400],
+      light: colorPalette.purple[700],
+    });
+    this.registerColor(`${sNav}-bg`, {
```
feloy: I think `hover` should appear in the name

```suggestion
    this.registerColor(`${sNav}-hover-bg`, {
```

packages/main/src/plugin/color-registry.ts
```diff
@@ -543,4 +544,18 @@ export class ColorRegistry {
       light: colorPalette.gray[500],
     });
   }
+
+  // links
+  protected initLink(): void {
+    const sNav = 'link';
```
feloy: I would use a different variable name, for example `link` as this one seems to be a copy-paste propagated from `initSecondaryNav`

ekidneyrh:
> @ekidneyrh I noticed in the design for #7149 that the color for the breadcrumb at the top left was defined, but this is a standalone Link component that is used everywhere we have links - dashboard, forms, deploy to Kube, etc. To support a hover effect on all background colors it currently uses 10% translucent white, so I've defined two colors: `link` and `link-bg`. The colors should be fine since the main ones come from #7149, but let me know if you prefer different names, e.g. maybe should be link-hover-bg?

That sounds good to me! I'm not too picky on the names.

ekidneyrh (approved):

feloy:
> To support a hover effect on all background colors it currently uses 10% translucent white, so I've defined two colors: `link` and `link-bg`

It would be nice to use this opacity technique in other places for hover, and also for disabled components and text emphasis (some examples here: https://m2.material.io/design/color/text-legibility.html). That would help reduce the number of colors in the palettes

deboer-tim:
> > To support a hover effect on all background colors it currently uses 10% translucent white, so I've defined two colors: `link` and `link-bg`
> 
> It would be nice to use this opacity technique in other places for hover, and also for disabled components and text emphasis (some examples here: https://m2.material.io/design/color/text-legibility.html). That would help reduce the number of colors in the palettes

Agreed. Ideally I'd like to use RGBA values but I can't find any way of cleanly specifying them in the registry or how we use them/Tailwind. I checked and the following does work since colors are just strings, but it feels a bit ugly. Am I overreacting, or can you think of a better way to do this?
```
this.registerColor(`${link}-hover-bg-opacity`, {
    dark: '10',
    light: '10',
});
```

feloy:
> > > To support a hover effect on all background colors it currently uses 10% translucent white, so I've defined two colors: `link` and `link-bg`
> > 
> > 
> > It would be nice to use this opacity technique in other places for hover, and also for disabled components and text emphasis (some examples here: https://m2.material.io/design/color/text-legibility.html). That would help reduce the number of colors in the palettes
> 
> Agreed. Ideally I'd like to use RGBA values but I can't find any way of cleanly specifying them in the registry or how we use them/Tailwind. I checked and the following does work since colors are just strings, but it feels a bit ugly. Am I overreacting, or can you think of a better way to do this?
> 
> ```
> this.registerColor(`${link}-hover-bg-opacity`, {
>     dark: '10',
>     light: '10',
> });
> ```

It has to be checked for the difference cases, but it seems to me the opacity would always be the same for dark and light modes (and even the same if one changes the theme colors). I would simply put a `class="opacity-10"` in the template.

This way, we could use, for example:

```
class="text-[var(--pd-link)] opacity-75 hover:opacity-95 disabled:opacity-20"
```

benoitf:
and for some components, it relies on accent-color
https://tailwindcss.com/docs/accent-color

deboer-tim:
I realized the background hover opacity (bg-opacity-10) wasn't working correctly, and the delay looking for a solution required a rebase to pick up Link move to UI package and changes to the color registry.

The only path I've been able to find to get the background hover opacity is by defining this one color as a Tailwind color using the `rgb(from var(...))` syntax, which then allows us to use the new Tailwind opacity modifier (`/10`) instead.

https://tailwindcss.com/docs/upgrade-guide#new-opacity-modifier-syntax

@benoitf or @feloy would appreciate a quick comment to confirm this path is acceptable or if you can think of others. IMHO this path isn't too bad b/c we'll only have to define a couple colors this way.

benoitf:
I would not add the color in tailwind as it won't be shared properly with other projects

deboer-tim:
Any other ideas to set a hover bg opacity in Link? I haven't found any other path that allows you to use Tailwind's `/10` syntax, and the only other working path is putting `dark: colorPalette.white + '2'` directly in the color registry.

feloy:
Because the `bg-opacity-*` notation seems deprecated (https://github.com/tailwindlabs/tailwindcss/discussions/13703#discussioncomment-9479593), and because the `/n` notation does not work for now with the use of css var, and to not define a color in the tailwind config which is not accessible to extensions, it seems that the #rgba 
 (or #rrggbbaa) notation is the best choice for now

```
    this.registerColor(`${link}-hover-bg`, {
      dark: colorPalette.white + '2',
      light: colorPalette.black + '2',
    });
```

deboer-tim:
> ... it seems that the #rgba (or #rrggbbaa) notation is the best choice for now

Yeah, it does seem to be the simpler option atm, and would allow another theme to change the opacity as necessary. I have repushed the branch with this change.

The closest value to 10% would actually be about hex 1A, but since black/white are three-hex values I tried both `1` (11) and `2` (22). Although it's slightly higher, I went with `2` because it is closer and `1` was not enough contrast for smaller fonts.

benoitf (approved):

feloy (approved):

PR #7221, opened 2024-05-15
chore(tests): add arch selection for build image

### LINKED ISSUE #7220: Add architecture type selection when building an image on images page

Currently it's not possible to select the arch type when building an image on images page in the e2e test framework.

Add arch selection to the build method.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Adds arch selection for build image functionality for e2e tests in PD.

https://github.com/containers/podman-desktop/issues/7220


tests/playwright/src/model/core/states.ts
```diff
@@ -42,3 +42,9 @@ export enum PodState {
   Paused = 'PAUSED',
   Unknown = 'UNKNOWN',
 }
+
+export enum ArchitectureType {
```
odockal: I would prefer to have this enum in other file. `states.ts` was intended for states of objects. I am thinking about `platforms` or something else perhaps?

tests/playwright/src/model/pages/build-image-page.ts
```diff
@@ -43,9 +46,16 @@ export class BuildImagePage extends BasePage {
     this.containerFilePathButton = page.getByRole('button', { name: 'Browse...' }).first();
     this.arm64Button = page.getByLabel('linux/arm64');
     this.amd64Button = page.getByLabel('linux/amd64');
+    this.arm64checkbox = page.getByLabel('ARM® aarch64 systems');
```
odockal: Can you please also add another locator that would hold the div/container where all these buttons / checkboxes are? 
there is this container: 
```
<div class="flex flex-col" role="region" aria-label="Build Platform Options">
```
that contains all the archs that could be selected.

another question, why it is that we need to add checkbox when there are already existing buttons (`arm64button`) that if you click on them, the checkbox is being toggled?

tests/playwright/src/model/pages/build-image-page.ts
```diff
@@ -58,10 +68,32 @@ export class BuildImagePage extends BasePage {
       await this.imageNameInput.pressSequentially(imageName, { delay: 50 });
     }
 
+    if (archType !== ArchitectureType.Default) {
+      await this.uncheckedAllCheckboxes();
+      if (archType === ArchitectureType.ARM64) {
```
odockal: How about the rest of the architectures available? We would tackled them later? Maybe the use of a switch would be better then.

odockal (changes requested):

tests/playwright/src/model/pages/build-image-page.ts
cbr7: My goal is to provide the functionality for the archs we're using right now and for which we have use cases right now, we can add more types and refactor to a switch at that point.

tests/playwright/src/model/core/states.ts
cbr7: sure

tests/playwright/src/model/pages/build-image-page.ts
cbr7: I'm not sure I understand, why do we need to grab the parent element here? What is the point of that since we can't do anything with it anyway?

As for your second question... yes the checkboxes are being toggled when the buttons are being clicked, so even though to change toggle states you need to click the buttons, to check the current state of the checkbox you need to validate the checked/unchecked status of the checkbox, not the button.

tests/playwright/src/model/pages/build-image-page.ts
cbr7: refactored to use switch keyword

tests/playwright/src/model/pages/build-image-page.ts
odockal: > I'm not sure I understand, why do we need to grab the parent element here? What is the point of that since we can't do anything with it anyway?

To avoid ambiguity. In future the page can be changed so that we make sure to search for the element in the right place. Also, reading: `page.getByLabel()` in the code is pretty vague. You have to search through whole page. Whilst with ie. `platform.getByLabel` where `platform` is the locator that represents a div/container or region in this case to hold all the elements tied to a platform selection or adding new one, etc. So it is true that you cannot do anything with it (except checking ie. visibility), but it helps navigation. 

> As for your second question... yes the checkboxes are being toggled when the buttons are being clicked, so even though to change toggle states you need to click the buttons, to check the current state of the checkbox you need to validate the checked/unchecked status of the checkbox, not the button.

True! good point.

cbr7: Ok, I extracted the platform region in a field and using it when getting the platform buttons, but to be honest to me it doesn't make much sense given that each button/checkbox is specifically named, I don't really think it helps with debugging at all, to me it seems the same only that we're using an intermediate element instead of going to the element needed.

benoitf (approved):

odockal (approved):

PR #7246, opened 2024-05-17
chore: edge and light mode for modal dialog

### LINKED ISSUE #7187: Outline of modal dialogs doesn't stand out

See #7180 for an example of switching from our older/more hardcoded dialog styling to the correct design. I'll admit there isn't much difference, but to me the old style had just a little bit of an edge on the dialog so you could see there is a panel there, but the correct design blends into the background more - black on black. To me it looks like the menu is now just floating instead of being on a quick pick panel.

I've noticed this on regular dialogs, but with the size it is still clear that you're looking at a dialog. IMHO in both cases it would look better and stand out from the background a little more if there was a tiny border, even in dark grey.

FYI to @ekidneyrh 

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds light mode colors for Modal dialog along with a defined 1px border so that it is not washed out.

Initial light mode color values are taken from other mappings in the registry. It is not possible to use bg-opacity-60 with colors from the registry due to Tailwind CSS post processing order. I looked at several other options (defining colors as r g b instead of hex, trying several css functions, but the simplest is just to use direct styling instead.

### Screenshot / video of UI

Dark:

Screenshot 2024-05-17 at 11 47 34 AM

Light:

Screenshot 2024-05-17 at 11 46 51 AM

Fixes #7187.
First minor part of #7184.

### How to test this PR?

Confirm no regressions and new border in dark mode, check colors (of Modal, not any contents) in light mode.

### DIFF
--- a/packages/ui/src/lib/modal/Modal.svelte
+++ b/packages/ui/src/lib/modal/Modal.svelte
@@ -35,12 +35,12 @@ if (previously_focused) {
 <div class:items-center="{!top}" class="absolute w-full h-full flex justify-center">
   <button
     aria-label="close"
-    class="fixed top-0 left-0 w-full h-full bg-black bg-opacity-60 bg-blend-multiply z-40"
+    class="fixed top-0 left-0 w-full h-full bg-[var(--pd-modal-fade)] bg-blend-multiply opacity-60 z-40 cursor-default"
     on:click="{close}"></button>
 
   <div
     class:translate-y-[-20%]="{!top}"
-    class="bg-charcoal-800 z-50 rounded-xl overflow-auto w-[calc(200vw-4em)] h-fit max-w-[42em] max-h-[calc(100vh-4em)]"
+    class="bg-[var(--pd-modal-bg)] z-50 rounded-xl overflow-auto w-[calc(200vw-4em)] h-fit max-w-[42em] max-h-[calc(100vh-4em)] border-[1px] border-[var(--pd-modal-border)]"
     role="dialog"
     aria-label="{name}"
     aria-modal="true"
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -236,6 +236,7 @@ export class ColorRegistry {
     this.initTable();
     this.initDetails();
     this.initTab();
+    this.initModal();
   }
 
   protected initGlobalNav(): void {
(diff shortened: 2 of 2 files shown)


cdrage (changes requested):
Screenshot 2024-05-22 at 9 47 23 AM

I'm finding it very difficult to read / as though the modal has some form of opacity? I saw the comment:  "It is not possible to use bg-opacity-60 with colors from the registry due to Tailwind CSS post processing order"

I'm assuming it'll be fixed in a follow up PR?

deboer-tim:
> I'm finding it very difficult to read / as though the modal has some form of opacity? I saw the comment: "It is not possible to use bg-opacity-60 with colors from the registry due to Tailwind CSS post processing order"
> 
> I'm assuming it'll be fixed in a follow up PR?

Yes. That's the MessageBox text that's light, because it doesn't have support for light mode yet. This PR is only covering Modal: the dialog outline and background color, and opacity of the rest of the screen. Individual dialogs like MessageBox and others will be covered under #7184.

packages/ui/src/lib/modal/Modal.svelte
```diff
@@ -35,12 +35,13 @@ if (previously_focused) {
 <div class:items-center="{!top}" class="absolute w-full h-full flex justify-center">
   <button
     aria-label="close"
-    class="fixed top-0 left-0 w-full h-full bg-black bg-opacity-60 bg-blend-multiply z-40"
+    class="fixed top-0 left-0 w-full h-full bg-[var(--pd-modal-fade)] bg-blend-multiply z-40"
+    style="opacity: 0.6"
```
feloy: It seems to work by using class `opacity-60` (instead of `bg-opacity-60`)

```suggestion
    class="fixed top-0 left-0 w-full h-full bg-[var(--pd-modal-fade)] bg-blend-multiply z-40 opacity-60"
```

deboer-tim: So focused on replicating what was there I failed to see the obvious. 🤦🏼‍♂️ Fixed.

deboer-tim:
Rebased to fix merge conflict from other light mode changes. Also fixed a nit that was annoying me while retesting: the background should not have a button cursor.

@cdrage and @feloy please re-review. Again, this is just the modal background, edge, and rest of the screen, not any particular dialog contents.

feloy (approved):

cdrage (approved):

PR #7249, opened 2024-05-17
Light mode for DetailsPage

### LINKED ISSUE #6927: Light mode for DetailsPage

We need to support light mode on DetailsPage.

Define missing bg/tile/divider colors in the light mode palette and use them on the component.

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

I have reused `--pd-table-body-text-sub-secondary` for the image id in the header of the image details page. Either we move this variable out of the `table` theming to place in a more generic one, or we create a new entry.

### Screenshot / video of UI

details-dark

details-light

Fixes #6927 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -234,6 +234,7 @@ export class ColorRegistry {
     this.initCheckbox();
     this.initToggle();
     this.initTable();
+    this.initDetails();
   }
 
   protected initGlobalNav(): void {
@@ -430,6 +431,11 @@ export class ColorRegistry {
     this.registerColor(`${ct}header`, {
       dark: colorPalette.white,
       light: colorPalette.charcoal[900],
+    });
+
+    this.registerColor(`${ct}sub-header`, {
+      dark: colorPalette.gray[900],
+      light: colorPalette.purple[900],
     });
 
     this.registerColor(`${ct}header-icon`, {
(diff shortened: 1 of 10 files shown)


deboer-tim (approved):
Every change reveals more that needs to be done :), but this one looks good. Thanks, approved.

ekidneyrh:
I actually have a mock-up made up for this. I tidied it up so it's readable:

[Details page.pdf](https://github.com/containers/podman-desktop/files/15376031/Details.page.pdf)
Details page

For the content-header-icon, you can ignore the colour value there as it follows its own colour scheme depending on whether it's running etc. 

I know there's those mini buttons there too, but I think they can be defined separately. I was unsure about the code part under the kube header - I'm not sure if there is a default styling to that or if it follows a particular rule?

feloy:
> I actually have a mock-up made up for this. I tidied it up so it's readable:
> 
> [Details page.pdf](https://github.com/containers/podman-desktop/files/15376031/Details.page.pdf) Details page
> 
> For the content-header-icon, you can ignore the colour value there as it follows its own colour scheme depending on whether it's running etc.
> 
> I know there's those mini buttons there too, but I think they can be defined separately. I was unsure about the code part under the kube header - I'm not sure if there is a default styling to that or if it follows a particular rule?

A few comments:

- The tabs are also visible in the list pages. Even if the components are different (Tabs in details page vs `Button type="tab"` in list page), I think we want to have only one set of colors for them. Do we want to keep the `details-tab-text/details-tab-highlight` colors defined with the `details-` prefix, or do we want to move it to a specific prefix, for example `tab-text/tab-highlight`?
- in the mockup, the text color  of the selected tab is the same as the others. Currently, the selected tab has a `white` text color, making it highlighted. Do we want to remove this highlight of the text (and only highlight the border)? If not, I have added a new color `tab-text-highlight`,
- I have added a color (`details-tab-hover`), for the color of the bottom border when a tab is hovered.

ekidneyrh:
> The tabs are also visible in the list pages. Even if the components are different (Tabs in details page vs `Button type="tab"` in list page), I think we want to have only one set of colors for them. Do we want to keep the `details-tab-text/details-tab-highlight` colors defined with the `details-` prefix, or do we want to move it to a specific prefix, for example `tab-text/tab-highlight`? 

You're right, we should just have the one set of colours for the tabs/highlight. Since it's reused, I'd go with dropping the details for it. 


> in the mockup, the text color  of the selected tab is the same as the others. Currently, the selected tab has a `white` text color, making it highlighted. Do we want to remove this highlight of the text (and only highlight the border)? If not, I have added a new color `tab-text-highlight`,
> 

Good spot, I'd have that as charcoal-300 in the light mode for the tab-text-highlight.

> I have added a color (`details-tab-hover`), for the color of the bottom border when a tab is hovered.

Would this be similar to the first point that it would be reused, so maybe not have the details- prefix?
I'd suggest having these values be dark mode purple-400, light mode purple-500, same as the input hover :)

feloy:
> > The tabs are also visible in the list pages. Even if the components are different (Tabs in details page vs `Button type="tab"` in list page), I think we want to have only one set of colors for them. Do we want to keep the `details-tab-text/details-tab-highlight` colors defined with the `details-` prefix, or do we want to move it to a specific prefix, for example `tab-text/tab-highlight`?
> 
> You're right, we should just have the one set of colours for the tabs/highlight. Since it's reused, I'd go with dropping the details for it.
> 
> > in the mockup, the text color  of the selected tab is the same as the others. Currently, the selected tab has a `white` text color, making it highlighted. Do we want to remove this highlight of the text (and only highlight the border)? If not, I have added a new color `tab-text-highlight`,
> 
> Good spot, I'd have that as charcoal-300 in the light mode for the tab-text-highlight.
> 
> > I have added a color (`details-tab-hover`), for the color of the bottom border when a tab is hovered.
> 
> Would this be similar to the first point that it would be reused, so maybe not have the details- prefix? I'd suggest having these values be dark mode purple-400, light mode purple-500, same as the input hover :)

Thanks @ekidneyrh , I have made the changes

benoitf (approved):

PR #7271, opened 2024-05-21
feat: adding highlight status bar task manager

### LINKED ISSUE #4243: feat: adding badge to status bar item such as tasks on specific events

axel7083:
### What does this PR do?

When a task is created, we set the status bar highlight to true, adding this little bubble as shown in the screenshots.

This bubble is removed when the task manager is open. 

This PR also change the order priority of the help, to -1 to ensure it is always the one on the most right.

### Screenshot / video of UI

| Before | After |
| --- | --- |
| image | image |
| N/A | image |

Following https://github.com/containers/podman-desktop/pull/5773 

Fixes https://github.com/containers/podman-desktop/issues/4243

### How to test this PR?

Create a notification, you can do so by pushing to kubernetes cluster an image for example.

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -509,7 +509,7 @@ export class PluginSystem {
       }
     });
 
-    statusBarRegistry.setEntry('help', false, 0, undefined, 'Help', 'fa fa-question-circle', true, 'help', undefined);
+    statusBarRegistry.setEntry('help', false, -1, undefined, 'Help', 'fa fa-question-circle', true, 'help', undefined);
 
     statusBarRegistry.setEntry(
       'troubleshooting',
--- a/packages/main/src/plugin/task-manager.spec.ts
+++ b/packages/main/src/plugin/task-manager.spec.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { expect, test, vi } from 'vitest';
+import { beforeEach, expect, test, vi } from 'vitest';
 
 import type { CommandRegistry } from '/@/plugin/command-registry.js';
 import type { StatusBarRegistry } from '/@/plugin/statusbar/statusbar-registry.js';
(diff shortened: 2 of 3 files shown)


packages/main/src/plugin/index.ts
```diff
@@ -509,7 +509,7 @@ export class PluginSystem {
       }
     });
 
-    statusBarRegistry.setEntry('help', false, 0, undefined, 'Help', 'fa fa-question-circle', true, 'help', undefined);
+    statusBarRegistry.setEntry('help', false, -1, undefined, 'Help', 'fa fa-question-circle', true, 'help', undefined);
```
benoitf, line 512: any reason why changing from 0 to -1 ?

benoitf: I mean, I see the comment 
> This PR also change the order priority of the help, to -1 to ensure it is always the one on the most right.

but should it be like that ?

benoitf (approved):

packages/main/src/plugin/index.ts
axel7083: That's the easy fix I found to avoid this

image

The task manager has some kind of triangle position, and with today implementation order, we register the task first or last, making it the most of the right

deboer-tim: 😂 The task manager popup has no relationship to where its status bar item is, and the order of the other items is 'random'. Still trying to decide if this is brilliant or should be done 'correctly' to avoid future breakage.

axel7083: > Still trying to decide if this is brilliant or should be done 'correctly' to avoid future breakage.

I am fine with it :)

cdrage (approved):
Tested, I see the change, code and practical testing worked well, LGTM!

PR #7278, opened 2024-05-22
refactor: move EmptyScreen/FilteredEmptyScreen to ui package

### LINKED ISSUE #6919: Implement EmptyScreen/FilteredEmptyScreen UI component

jeffmaury:
Fixes #6919

### What does this PR do?

Move EmptyScreen/FilteredEmptyScreen to ui library

### Screenshot / video of UI

N/A

#6919 

### How to test this PR?


packages/ui/src/lib/screen/EmptyScreen.svelte
```diff
@@ -56,7 +58,7 @@ let copyTextDivElement: HTMLDivElement;
         <div class="font-mono text-gray-400" bind:this="{copyTextDivElement}" data-testid="copyTextDivElement">
           {commandline}
         </div>
-        <button title="Copy To Clipboard" class="ml-5" on:click="{() => copyRunInstructionToClipboard()}"
+        <button title="Copy To Clipboard" class="ml-5" on:click="{copyRunInstructionToClipboard}"
```
axel7083: Could we make it use the `Button` component ?

axel7083 (approved):

deboer-tim (approved):
As per @axel7083 we should use Button now and needs to be rebased on other changes, but otherwise LGTM.

packages/ui/src/lib/screen/EmptyScreen.svelte
```diff
@@ -21,10 +24,11 @@ onMount(() => {
   processed = true;
 });
 
-function copyRunInstructionToClipboard() {
+function copyRunInstructionToClipboard(): void {
   const text = copyTextDivElement?.textContent;
   if (text) {
-    window.clipboardWriteText(text);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (window as any).clipboardWriteText(text);
```
benoitf: I think this part is problematic as it's a custom API of Podman Desktop exposed to window object, but it won't work if you're like in a webview

jeffmaury: Should be solved now

packages/ui/src/lib/screen/EmptyScreen.svelte
```diff
@@ -14,17 +17,19 @@ export let hidden = false;
...
   }
   processed = true;
 });
 
-function copyRunInstructionToClipboard() {
+function copyRunInstructionToClipboard(): void {
   const text = copyTextDivElement?.textContent;
   if (text) {
-    window.clipboardWriteText(text);
+    dispatch('click');
```
benoitf: I would not name the function copyToClipboard as it's the caller that decide what to do (could be clipboard copy but could be something different)

benoitf: also, shouldn't it dispatch the text of the element ?

either we need to remove the text variable or we should dispatch the value

benoitf (approved):

Issue #7276, opened 2024-05-22
Improve error reporting in containers e2e test


odockal:
### Is your enhancement related to a problem? Please describe

We have this example of the failed test case in `containers-smoke.spec.ts`:
```
 ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 4 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  tests/playwright/src/specs/container-smoke.spec.ts > Verification of container creation workflow > Start a container 'alpine-container'
Error: expect(received).toBeTruthy()

Received: false

Call Log:
- Timeout 10000ms exceeded while waiting on the predicate
 ❯ pollMatcher node_modules/playwright/lib/matchers/expect.js:255:11
 ❯ tests/playwright/src/specs/container-smoke.spec.ts:105:5
    103|     await pdRunner.screenshot('containers-run-image.png');
    104|     const containers = await runImage.startContainer(containerToRun);
    105|     await playExpect
       |     ^
    106|       .poll(async () => await containers.containerExists(containerToRu…
    107|       .toBeTruthy();
```

We know that the container creation was not successful. although, there is an error message showed after "Start container" button was pressed which has more information as of why the action failed.

### Describe the solution you'd like

Handle the error message and show it in the assertion if test case fails so it is immediately visible in the test log.

odockal:
Fixed by #7283

PR #7283, opened 2024-05-22
chore(tests): improve container start up error handling

odockal:
### What does this PR do?
Improves reporting of the error during container start up in e2e tests.
#7276 

### How to test this PR?
If container start up fails in the e2e tests, the error would be thrown. 

Quite a lot to set a container to fail, you can add some custom code to mess up container start up ie. in `run-image-page.ts` file, add this into line 85:
```
      await this.activateTab('Security');
      const secPlaceholder = this.page.getByPlaceholder('Enter a security option');
      await secPlaceholder.fill('notreallyasecurityoption');
```

Then adjust `package.json` to run only container-smoke test, to make is faster, run 
`yarn test:e2e`. 
Test should fail with assumed error shown in the assertion output.


tests/playwright/src/model/pages/run-image-page.ts
```diff
@@ -81,7 +93,15 @@ export class RunImagePage extends BasePage {
       throw Error(`Start Button not enabled: ${errMessage}`);
     }
     await this.startContainerButton.click();
-    return new ContainersPage(this.page);
+    // After clicking on the button there seems to be for possible outcomes
+    // 1. Opening particular container's details page with tty tab opened
+    // 2. Opening Containers page with new container on it
+    // 3. staying on the run image page with an error
+    // 4. Starting a container without entrpoint or command creates a container, but it stays on Run Image Page without error
+    if ((await this.errorAlert.count()) > 0 && (await this.name.isVisible())) {
```
benoitf: just for readability I think it will be easier with intermediate variables

benoitf (approved):

tests/playwright/src/model/pages/run-image-page.ts
odockal: Yes. I have changed the code a bit, added a wait for page to switch (or not) just to make sure to get right error in the line.

tests/playwright/src/model/pages/run-image-page.ts
```diff
@@ -18,8 +18,13 @@
 
 import type { Locator, Page } from '@playwright/test';
 
+import { waitWhile } from '../../utility/wait';
 import { BasePage } from './base-page';
-import { ContainersPage } from './containers-page';
+
+export interface ContainerInteractiveParams {
```
cbr7: This shouldn't be here, it should be moved in one of the other files where interfaces are defined.

tests/playwright/src/model/pages/run-image-page.ts
```diff
@@ -81,7 +94,25 @@ export class RunImagePage extends BasePage {
       throw Error(`Start Button not enabled: ${errMessage}`);
     }
     await this.startContainerButton.click();
-    return new ContainersPage(this.page);
+    // After clicking on the button there seems to be four possible outcomes
+    // 1. Opening particular container's details page with tty tab opened
+    // 2. Opening Containers page with new container on it
+    // 3. staying on the run image page with an error
+    // 4. Starting a container without entrypoint or command creates a container, but it stays on Run Image Page without error
+    await waitWhile(
```
cbr7, line 112: Why not replace this waitWhile() with `await playExpect(this.name).toBeVisible()` ? I'm not clear what happens if the field is not visible in your current implementation before the timer runs out, since the error is not propagated, it seems to me that the test would continue even though the field wouldn't be visible.

tests/playwright/src/model/pages/run-image-page.ts
```diff
@@ -57,21 +62,29 @@ export class RunImagePage extends BasePage {
...
       await textbox.fill(customName);
     }
 
-    if (!interactive) {
+    if (optionalParams?.attachTerminal !== undefined) {
       // disable the checkbox in advanced tab
       await this.activateTab('Advanced');
       const checkbox = this.page.getByRole('checkbox', { name: 'Attach a pseudo terminal' });
-      await checkbox.uncheck();
+      optionalParams.attachTerminal ? await checkbox.check() : await checkbox.uncheck();
```
cbr7, line 85: Checking and unchecking should be waited and asserted on to ensure that the system waits until the checkbox is actually marked.

Please refactor this into a if-else clause where the checking/unchecking is asserted with `await playExpect(checkbox).toBeChecked()/not.toBeChecked()`

tests/playwright/src/model/pages/run-image-page.ts
```diff
@@ -57,21 +62,29 @@ export class RunImagePage extends BasePage {
...
+    }
+
+    if (optionalParams?.interactive !== undefined) {
+      // disable the checkbox in advanced tab
+      await this.activateTab('Advanced');
+      const checkbox = this.page.getByRole('checkbox', { name: 'Interactive: Keep STDIN' });
+      optionalParams.interactive ? await checkbox.check() : await checkbox.uncheck();
```
cbr7, line 93: Same as above.

tests/playwright/src/model/pages/run-image-page.ts
```diff
@@ -81,7 +94,25 @@ export class RunImagePage extends BasePage {
       throw Error(`Start Button not enabled: ${errMessage}`);
     }
     await this.startContainerButton.click();
-    return new ContainersPage(this.page);
```
cbr7, line 84: I don't really agree with removing this, it's completely pattern breaking to have a method that does part of its functionality screen navigation to return void.

tests/playwright/src/specs/container-smoke.spec.ts
```diff
@@ -101,7 +103,9 @@ describe('Verification of container creation workflow', async () => {
     const imageDetails = await images.openImageDetails(imageToPull);
     const runImage = await imageDetails.openRunImage();
     await pdRunner.screenshot('containers-run-image.png');
-    const containers = await runImage.startContainer(containerToRun);
+    await runImage.startContainer(containerToRun, containerStartParams);
+    const containers = new ContainersPage(page);
+    await playExpect(containers.header).toBeVisible();
```
cbr7, line 107: If the startContainer() returns void, why would this be valid? what is handling the navigation to ContainersPage in order to make this assertion consistent?

tests/playwright/src/specs/container-smoke.spec.ts
```diff
@@ -173,7 +177,8 @@ describe('Verification of container creation workflow', async () => {
 
     for (const container of containerList) {
       const images = await navigationBar.openImages();
-      const containersPage = await images.startContainerWithImage(imageToPull, container);
+      await images.startContainerWithImage(imageToPull, container, containerStartParams);
+      const containersPage = new ContainersPage(page);
```
cbr7: As above, this is fundamentally pattern breaking and introduces flakyness.

tests/playwright/src/specs/pod-smoke.spec.ts
```diff
@@ -116,7 +119,9 @@ describe.skipIf(process.env.GITHUB_ACTIONS && process.env.RUNNER_OS === 'Linux')
       let runImage = await imageDetails.openRunImage();
       await pdRunner.screenshot('pods-run-backend-image.png');
       await runImage.setCustomPortMapping('6379:6379');
-      let containers = await runImage.startContainer(backendContainer, false);
+      await runImage.startContainer(backendContainer, containerStartParams);
+      let containers = new ContainersPage(page);
+      await playExpect(containers.header).toBeVisible();
```
cbr7, line 122: Same as above.

tests/playwright/src/specs/pod-smoke.spec.ts
```diff
@@ -136,7 +141,9 @@ describe.skipIf(process.env.GITHUB_ACTIONS && process.env.RUNNER_OS === 'Linux')
       }
       await pdRunner.screenshot('pods-run-frontend-image.png');
       await pdRunner.screenshot('pods-run-frontend-image.png');
-      containers = await runImage.startContainer(frontendContainer, false);
+      await runImage.startContainer(frontendContainer, containerStartParams);
+      containers = new ContainersPage(page);
+      await playExpect(containers.header).toBeVisible();
```
cbr7, line 143: Same as above.

tests/playwright/src/specs/pod-smoke.spec.ts
```diff
@@ -348,7 +355,8 @@ describe.skipIf(process.env.GITHUB_ACTIONS && process.env.RUNNER_OS === 'Linux')
         const runImagePage = await imageDetailsPage.openRunImage();
         await playExpect(runImagePage.heading).toContainText(backendImage);
         await runImagePage.setCustomPortMapping(`${portsList[i]}:${portsList[i]}`);
-        const containersPage = await runImagePage.startContainer(containerNames[i]);
+        await runImagePage.startContainer(containerNames[i], containerStartParams);
+        const containersPage = new ContainersPage(page);
```
cbr7: Same as above.

tests/playwright/src/model/pages/images-page.ts
```diff
@@ -66,10 +66,14 @@ export class ImagesPage extends MainPage {
     return await editImagePage.renameImage(newname);
   }
 
-  async startContainerWithImage(image: string, containerName: string): Promise<ContainersPage> {
+  async startContainerWithImage(
```
cbr7, line 70: This should not return void, it should return the object that it navigates to after it performs whatever function it has.

cbr7 (changes requested):
This new implementation is likely to lead to inconsistencies and I don't understand why the navigation was removed and now some methods that shouldn't are returning void.

tests/playwright/src/model/pages/run-image-page.ts
odockal: Moved to core folder.

tests/playwright/src/model/pages/run-image-page.ts
odockal: Should be fine now.

tests/playwright/src/model/pages/run-image-page.ts
odockal: Should be fine now. assert added.

cbr7 (approved):
LGTM

PR #7288, opened 2024-05-23
fix: remove tinro from Tab and ui

### LINKED ISSUE #7235: remove tinro from ui component library

Tab component brought the tinro component into the ui library

but the UI library should be agnostic on the routing system used by the framework.

Need to refactor the tab component and remove tinro from package.json

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

It removes the tinro dependency from the ui package

### Screenshot / video of UI

N/A

it resolves #7235 

### How to test this PR?

1. start desktop and see the tabs work as before

### DIFF
--- a/packages/ui/src/lib/tab/Tab.svelte
+++ b/packages/ui/src/lib/tab/Tab.svelte
@@ -1,21 +1,20 @@
 <script lang="ts">
-import { router } from 'tinro';
-
+export let routerPath: string;
 export let url: string;
 export let title: string;
 
-let baseURL = $router.path.substring(0, $router.path.lastIndexOf('/'));
+let baseURL = routerPath.substring(0, routerPath.lastIndexOf('/'));
 </script>
 
 <div
   class="pb-1 border-b-[3px] whitespace-nowrap hover:cursor-pointer"
-  class:border-[var(--pd-tab-highlight)]="{$router.path === baseURL + '/' + url}"
-  class:border-transparent="{$router.path !== baseURL + '/' + url}"
-  class:hover:border-[var(--pd-tab-hover)]="{$router.path !== baseURL + '/' + url}">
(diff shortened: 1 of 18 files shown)


packages/ui/src/lib/tab/Tab.svelte
```diff
@@ -1,21 +1,20 @@
 <script lang="ts">
-import { router } from 'tinro';
-
+export let routerPath: string;
 export let url: string;
```
feloy: I think we should extract the route logic from the component, and have only a `selected` parameter for the tab.

lstocchi: Ok, updating

lstocchi:
Updated the PR to use a selected prop on the Tab component. Not sure if there is a better way to do it (i'm open to suggestions) but i'm not a big fan of this change. I kept the commits separated so you can see the difference. 
Let me know

feloy:
>> I think we should extract the route logic from the component, and have only a selected parameter for the tab.

> Updated the PR to use a selected prop on the Tab component. Not sure if there is a better way to do it (i'm open to suggestions) but i'm not a big fan of this change. I kept the commits separated so you can see the difference. Let me know

Personally, I think that this component should be splitted in two parts:
- the part about router and url should be part of Podman Desktop, as this logic is very specific to the app, but is not a UI concern
- the graphical part should be part of the ui package. And it seems that the `Button type="tab"` could be reused for this part

@deboer-tim @benoitf WDYT?

benoitf:
I'm unsure about the pros/cons about using `Button type="tab"` or `Tab`

like should me merge Tab into button or Button with type tab should not be a button ?

else, to me PR is OK as at least we're getting rid of tinro which is the biggest concern for now

benoitf (approved):

feloy (approved):

Issue #7052, opened 2024-05-02
Rootful podman machine gets stuck using rootless connection after another rootless machine is deleted


cbr7:
### Bug description

It seems that in some situations a rootful podman machine can get stuck using a rootless connection if the user deletes another podman machine which is rootless and is set to default at the same time.

Additionally it seems that this scenario also makes the "This machine is not set to default" prompt to be displayed at all until such time that the user manually shuts down Podman Desktop and reopens it.

### Operating system

Windows 11

### Installation Method

Installer from website/GitHub releases

### Version

1.10.0

### Steps to reproduce

1. Have two podman WSL machines available on the system, one rootful and one rootless, the rootless machine should be set to default and both machines should be running at the same time.
2. Go to Settings -> Resources menu.
3. Stop the rootless podman machine.
4. A popup will be triggered during the stop action which will notify the user that the rootful machine is not set to default, select the `Ignore` option on the popup.
5. After the rootless podman machine finishes shutting down, delete it.
6. At this point the bug should have reproduced, going into CLI running the command `podman system connection list` will display the rootful podman machines's connections, but will show that the rootless connection is the one set to default, not the rootful one as it should be.
7. Additionally if the user makes anymore podman machines from PD, either rootful or rootless the "machine is not set as default" prompt will no longer be displayed until the user completely closes PD and reopens it.

benoitf:
when you say 'get stuck' it's all about the CLI ?
if you're facing podman CLI issues, it's not a Podman Desktop issue but an upstream podman issue

cbr7:
@benoitf most of it is CLI yes, but the fact that the popup in PD no longer gets triggered is a PD issue.

benoitf:
but Podman Desktop should not be a tool to workaround podman CLI bugs

maybe it's another one like https://github.com/containers/podman/issues/16483 in the past

cbr7:
@benoitf @odockal created https://github.com/containers/podman/issues/22577 in the podman repo

cbr7:
I wouldn't describe PD's functionality of displaying a popup as a workaround, especially since it's broken until the app is restarted. For example even if the user deletes all the podman machines and creates new ones, and as such everything is perfectly ok with the podman connections, the PD popup that notifies the user about default podman machines will still not be displayed. A full app restart will be required.

cbr7:
Was also able to reproduce this on macOS in PD 1.10.2

ScrewTSW:
Still present in 1.10.3

Issue #6668, opened 2024-04-05
Podman Machine created in Podman Desktop encounters an error using Hyperv Provider


odockal:
### Bug description

When I try to create a new machine in Podman Desktop on windows using Hyperv Provider I am seeing an error, but eventually machine appears in the Resources and can be started/used normally. Eventually it cannot be stopped then and seems to be somehow broken.

![hyperv_machine_create_error_2024-04-05 11-26](https://github.com/containers/podman-desktop/assets/19164299/b0f44842-15b1-448c-8a8f-e9875de32699)

The error:
```
Error: Command execution failed with exit code 125 Command execution failed with exit code 125 Error: failed to add resource: Job failed with error code: 32768 ('podman-machine-default' failed to add resources. (Virtual machine ID 9F272F29-6DD9-4B25-A041-60296B7EE028) 'podman-machine-default' encountered an unexpected error: The specified path is invalid. (0x800700A1). (Virtual machine ID 9F272F29-6DD9-4B25-A041-60296B7EE028))
```

### Operating system

windows 10

### Installation Method

Installer from website/GitHub releases

### Version

1.9.0

### Steps to reproduce

1. Install podman Desktop
2. Install Podman v4.9.4
3. Setup env. var. `CONTAINERS_MACHINE_PROVIDER="hyperv"`
4. Start up podman desktop
5. Create new machine with default values
6. Assert: Machine is created
Actual result: Machine creation throws an error: `failed to add resource`

odockal:
Quite interesting findings done by @cbr7 suggesting that we might need to name the machine other than default `podman-machine-default`. This could be point at the problem with a clean up between using wsl machine and hyperv (assuming keeping the same name for the machine when switching).

lstocchi:
@odockal i think this is fixed in the new podman 5 and can be closed.

Just tested with podman 5 and even if i switch to HyperV i can see the connections created with WSL previously.
This guarantees that you cannot create a machine with the same name on WSL and then hyperv. 
So now, if you try to create a default machine twice you see 
![image](https://github.com/containers/podman-desktop/assets/49404737/a3fb978c-ccfa-47be-9f2b-50642635546f)

odockal:
@lstocchi I am gonna look at hyperv with podman 5 today. So will report back. Do we want to focus on fixing anything 4.9.x related?

lstocchi:
@benoitf ^

odockal:
[...]

PR #7309, opened 2024-05-27
fix: remove limit when checking default machine

lstocchi:
### What does this PR do?

This PR removes the check of the number of machine when checking if the machine has to be marked as default.
It may happen that you only have one machine but multiple connections (maybe old ones or, on windows, that you have connections from wsl and hyperv but only machines of wsl/hyperv are displayed), and the check to inform the user that the running machine is not the default one is skipped. At this point you would see a machine which is not the default one and many actions may fail, like creating a kind cluster.

This could be a potential fix for https://github.com/containers/podman-desktop/issues/6668 or https://github.com/containers/podman-desktop/issues/7052 

### Screenshot / video of UI

N/A

hard to reproduce but maybe related to #6668 and #7052

### How to test this PR?

1. Have some old connections and 0 machine. 
2. Create a new machine and see that desktop tells you to update your default machine.

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -337,10 +337,7 @@ export async function updateMachines(provider: extensionApi.Provider): Promise<v
 
     // Finally, we check to see if the machine that is running is set by default or not on the CLI
     // this will create a dialog that will ask the user if they wish to set the running machine as default.
-    // this should only run if we have multiple machines
-    if (machines.length > 1) {
-      await checkDefaultMachine(machines);
-    }
+    await checkDefaultMachine(machines);
   }
 }
 
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -333,12 +333,10 @@ export async function updateMachines(provider: extensionApi.Provider): Promise<v
         // needs to start a machine
         provider.updateStatus('configured');
       }
-    }
 
-    // Finally, we check to see if the machine that is running is set by default or not on the CLI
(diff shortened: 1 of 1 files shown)


jeffmaury (approved):
LGTM

extensions/podman/src/extension.ts
```diff
@@ -337,8 +337,8 @@ export async function updateMachines(provider: extensionApi.Provider): Promise<v
 
     // Finally, we check to see if the machine that is running is set by default or not on the CLI
     // this will create a dialog that will ask the user if they wish to set the running machine as default.
-    // this should only run if we have multiple machines
-    if (machines.length > 1) {
+    // this should only run if we at least one machine
+    if (machines.length > 0) {
```
axel7083: Do we really need to keep the check ? let's just say we always check for default ?

lstocchi: I think i can move it inside the else above.
The check needs to be somewhere as inside the `checkDefaultMachine` there is this code
```
let defaultMachine = machines.find(machine => machine.Default);
if (!defaultMachine) {
```
so we should update the code in any case

axel7083: I see, as you want then! thx for the explanation

lstocchi: Moved above and removed the check as it is already done above 👍

axel7083: You said `moved` above, but I do not see the change in the `Files changed` ?

axel7083 (approved):
LGTM !

PR #7313, opened 2024-05-27
feat: table rows with children

### LINKED ISSUE #7209: Table component support for parent/children

To support issue #6529, the Table component in the Images page will need to support a table row being expandable to show its children.

Add support for parent/child relationships & multiple types to the Table component.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds the ability for rows within a table to have children, very similar to how container groups exist in the Containers view today.

Optional children function added to RowInformation to provide the necessary API for consumers. Expand/collapse control matches what is in the Containers view today.

In order to allow each row to be hover-able separately, a new parent div creates the outer box for each parent/children group, and when children exist these rows are added after the parent.

Since the UI package can't use stores directly like the Containers view does, a new expanded array property has been exposed to allow consumers to store/reload the list of expanded children. Since UI objects are recreated each time, the list requires all expandable objects to have an id property, and the list of ids is what is saved. This could be a configurable property in the future, but everything we want to expand today has an id.

Svelte sort animation conflicted with this change, so it has been removed. If this PR is approved I'll open up an issue to track finding another way to add it back.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/19958075/d4618755-d573-4d9d-928e-f83350437a16

Fixes #7209.

### How to test this PR?

Since this is component support without the corresponding use in Podman Desktop yet, the easiest way to test is to add code like the following to one of the existing tables like ImagesList.svelte:309 (adds an image as a child of itself):

```
  children: image => {
    if (image.shortId === 'some id') {
      return [image];
    }
    return [];
  },

```

### DIFF
--- a/packages/ui/src/lib/table/Table.svelte
+++ b/packages/ui/src/lib/table/Table.svelte
@@ -7,8 +7,9 @@
 <script lang="ts">
 /* eslint-disable import/no-duplicates */
 // https://github.com/import-js/eslint-plugin-import/issues/1479
+import { faChevronDown, faChevronRight } from '@fortawesome/free-solid-svg-icons';
 import { afterUpdate, onMount, tick } from 'svelte';
-import { flip } from 'svelte/animate';
+import Fa from 'svelte-fa';
 
 import Checkbox from '../checkbox/Checkbox.svelte';
 /* eslint-enable import/no-duplicates */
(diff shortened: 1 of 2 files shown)


axel7083 (approved):
Tested and work like a charm, amazing job!

> ℹ️ It is line 300 not 309 with your changes :)

benoitf (approved):

cdrage (approved):

PR #7315, opened 2024-05-27
chore: remove duplicate/old modal styling

deboer-tim:
### What does this PR do?

When trying to test #7246 I noticed that there are several uses of Modal that override the background color of Modal or try to add their own shadow or other styling as per very old prior designs. This just removes this extraneous styling so that these dialogs use what is in Modal (including light mode) and remaining work can focus on what actually needs to change within the dialogs.

Should be more consistent or no change to existing modals, just clears out unnecessary/duplicate code. IMHO SendFeedback is the only issue here since two custom fields were being set to the (now) background color, not sure if I should temporarily set them to another color or skip this for now.

Will need #7303 to see the rename image modal.

### Screenshot / video of UI

N/A

Part of #7184.

### How to test this PR?

Open up these modal dialogs to check for visual regression.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerList.svelte
+++ b/packages/renderer/src/lib/container/ContainerList.svelte
@@ -690,7 +690,7 @@ function setStoppedFilter() {
     }}">
     <div
       role="presentation"
-      class="inline-block w-full overflow-hidden text-left transition-all transform bg-charcoal-600 z-50 rounded-xl shadow-xl shadow-charcoal-900"
+      class="inline-block w-full overflow-hidden text-left transition-all"
       on:keydown="{keydownChoice}">
       <div class="flex items-center justify-between bg-black px-5 py-4 border-b-2 border-violet-700">
         <h1 class="text-xl font-bold">Create a new container</h1>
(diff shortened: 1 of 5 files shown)


feloy (approved):

feloy (commented):
I can also see the same pattern in PushImageModal, PreferenceProviderInstallationModal and TroubleShootingPageStoreDetails

deboer-tim:
> I can also see the same pattern in PushImageModal, PreferenceProviderInstallationModal and TroubleShootingPageStoreDetails

Yes, my method was ad hoc finding incorrect dialogs. :) Want me to fix those here, do another PR, or your later PR would cover?

feloy:
> > I can also see the same pattern in PushImageModal, PreferenceProviderInstallationModal and TroubleShootingPageStoreDetails
> 
> Yes, my method was ad hoc finding incorrect dialogs. :) Want me to fix those here, do another PR, or your later PR would cover?

My PR #7324 covers these 3 components. I have also added a new commit to cover the changes of this PR.

deboer-tim:
Closing since this is covered by #7324, I've added my approval there.

PR #7319, opened 2024-05-27
chore: fix name spacing in kubernetes dashboards

### LINKED ISSUE #7320: Kubernetes name column spacing is different on services vs deployment and ingress pages.

On services it's nice and close together:

Screenshot 2024-05-27 at 7 49 28 PM

But on ingress & deployments it is spaced and makes looking at the other column information more difficult (has more information "squished")

All 3 to have `1fr` formatting instead of `2fr`

_No response_

### Additional context

_No response_

cdrage:
chore: fix name spacing in kubernetes dashboards

### What does this PR do?

The dashboard for ingress with a lot of information is "squished" and
hard to read.

This PR changes the default width of ingress/route table as well as
deployments to "1fr" which matches the current
Services list formatting.

### Screenshot / video of UI

New changes show the name column a lot more closer to the next column:

Screenshot 2024-05-27 at 7 53 08 PM

Closes https://github.com/containers/podman-desktop/issues/7320

### How to test this PR?

Visual change, no test really required TBH, check `yarn watch` and view
the tables for changes.

### DIFF
--- a/packages/renderer/src/lib/deployments/DeploymentsList.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentsList.svelte
@@ -79,7 +79,6 @@ let statusColumn = new TableColumn<DeploymentUI>('Status', {
 });
 
 let nameColumn = new TableColumn<DeploymentUI>('Name', {
-  width: '2fr',
   renderer: DeploymentColumnName,
   comparator: (a, b) => a.name.localeCompare(b.name),
 });
--- a/packages/renderer/src/lib/ingresses-routes/IngressesRoutesList.svelte
+++ b/packages/renderer/src/lib/ingresses-routes/IngressesRoutesList.svelte
@@ -107,7 +107,6 @@ let statusColumn = new TableColumn<IngressUI>('Status', {
 });
 
 let nameColumn = new TableColumn<IngressUI | RouteUI>('Name', {
-  width: '2fr',
   renderer: IngressRouteColumnName,
   comparator: (a, b) => a.name.localeCompare(b.name),
 });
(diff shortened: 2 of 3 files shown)


deboer-tim (commented):
1fr is the default, so you could just remove it in this case.

My 2 cents: I agree names don't need to be 2x the other columns in these cases, but the name is the 'primary' column, having every column the same width is visually boring :) , and there are always some columns that tend to be wider or narrower than normal. I'd rather see a little more nuance like (e.g.) all name columns at 1.3 and/or reduce/expand a couple of the more non-typical columns too.

cdrage:
> 1fr is the default, so you could just remove it in this case.
> 
> My 2 cents: I agree names don't need to be 2x the other columns in these cases, but the name is the 'primary' column, having every column the same width is visually boring :) , and there are always some columns that tend to be wider or narrower than normal. I'd rather see a little more nuance like (e.g.) all name columns at 1.3 and/or reduce/expand a couple of the more non-typical columns too.

I agree :( unfortunately with the 2fr on ingress, I'm unable to view the entire URL length.

I have updated the PR removing the "width" section since it defaults to 1fr like you mentioned

deboer-tim (approved):
> unfortunately with the 2fr on ingress, I'm unable to view the entire URL length.

I'd argue that is a sign that url column should have >1fr, and would encourage you to do another PR ;-), but since I agree the names shouldn't need double width I'm ok with this PR.

cdrage:
Created a follow up PR @deboer-tim https://github.com/containers/podman-desktop/pull/7338

PR #7331, opened 2024-05-28
chore: add test watching renderer and main

cdrage:
chore: add test watching renderer and main

### What does this PR do?

When you run `yarn test:watch` it also runs e2e tests..

I would like to JUST run renderer or main if I am working only in that
section.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Run the commands and see that it only watches the renderer directory,
etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,8 @@
     "test:ui": "vitest run -r packages/ui --passWithNoTests --coverage",
     "test:tools": "vitest run tools --passWithNoTests --coverage",
     "test:watch": "vitest watch",
+    "test:watch:renderer": "npm run build:ui && vitest -c packages/renderer/vite.config.js watch packages/renderer --passWithNoTests",
+    "test:watch:main": "vitest watch -r packages/main --passWithNoTests",
     "watch": "node scripts/watch.mjs",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -71,6 +71,8 @@
     "test:ui": "vitest run -r packages/ui --passWithNoTests --coverage",
     "test:tools": "vitest run tools --passWithNoTests --coverage",
     "test:watch": "vitest watch",
+    "test-watch:renderer": "npm run build:ui && vitest -c packages/renderer/vite.config.js watch packages/renderer --passWithNoTests",
```
benoitf: not sure why in watch mode we run build:ui first

cdrage: see slack with errors I got, if podman-desktop:ui hasn't been updated in package.json vs what we have in main, we have to build it before running tests, or else it fails.

jeffmaury: in order to follow convention should be rename test:watch:renderer

jeffmaury (commented):
See comment

cdrage:
> See comment



> See comment



> See comment

updated!

PR #7332, opened 2024-05-28
chore: activate each extension sequentially

benoitf:
### What does this PR do?
Activate each extension one after the other

@axel7083 seems to had better result using sequential calls

it targets https://github.com/containers/podman-desktop/issues/7030

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -471,7 +471,10 @@ export class ExtensionLoader {
     const sortedExtensions = this.sortExtensionsByDependencies(analyzedExtensions);
 
     // now, load all extensions
-    await Promise.all(sortedExtensions.map(extension => this.loadExtension(extension)));
+
+    for (const extension of sortedExtensions) {
+      await this.loadExtension(extension);
+    }
   }
 
   // do we have circular dependencies?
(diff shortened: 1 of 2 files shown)


axel7083 (approved):

packages/main/src/plugin/extension-loader.ts
```diff
@@ -471,7 +471,10 @@ export class ExtensionLoader {
     const sortedExtensions = this.sortExtensionsByDependencies(analyzedExtensions);
 
     // now, load all extensions
-    await Promise.all(sortedExtensions.map(extension => this.loadExtension(extension)));
+
+    for (const extension of sortedExtensions) {
+      await this.loadExtension(extension);
```
cdrage, line 476: With all of this now being sequential, I'm assuming it will still timeout correctly (example.. if it got "stuck" loading say.. podman provider, it'll timeout and continue the rest?)

cdrage (changes requested):

packages/main/src/plugin/extension-loader.ts
benoitf: yes, it's done when calling activate method
https://github.com/containers/podman-desktop/blob/bf7c4bfb542d0adc13be8b80374f38034b058855/packages/main/src/plugin/extension-loader.ts#L1450

cdrage: Awesome 👍

cdrage (approved):

dgolovin (approved):
Tested on Windows and it indeed works much faster.

PR #7349, opened 2024-05-29
fix: container start error is not reported

### LINKED ISSUE #6981: Switch to a Containers page after creating a container without entrypoint or command
### Bug description

When I have an image that I want to run, and there is no name picked up, nor entry point or command, the outcome of pressing the Start the container button visually does nothing. Since the page is not switched, I have no idea that I am actually creating containers (without them being actually started). I should be led to the Container's details page so I know that I am actually doing something.

creatingContainerPageSwitch

### Operating system

Windows 10

### Installation Method

None

### Version

1.10.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #6981

### What does this PR do?

Report error if container start errored

### Screenshot / video of UI

start-container-error

#6981 

### How to test this PR?

1. Create a dummy image (FROM scratch)
2. Start this image

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -1878,11 +1878,9 @@ export class ContainerProviderRegistry {
       }
 
       const engine = this.internalProviders.get(engineId);
-      if (engine) {
+      if (engine && (options.start === true || options.start === undefined)) {
+        await container.start();
         await this.attachToContainer(engine, container, options.Tty, options.OpenStdin);
-        if (options.start === true || options.start === undefined) {
-          await container.start();
-        }
       }
       return { id: container.id };
     } catch (error) {
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -2533,6 +2533,47 @@ describe('createContainer', () => {
     expect(createContainerMock).toHaveBeenCalled();
     expect(startMock).not.toHaveBeenCalled();
   });
+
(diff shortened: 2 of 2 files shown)


packages/main/src/plugin/container-registry.ts
```diff
@@ -1878,11 +1878,9 @@ export class ContainerProviderRegistry {
       }
 
       const engine = this.internalProviders.get(engineId);
-      if (engine) {
+      if (engine && (options.start === true || options.start === undefined)) {
+        await container.start();
```
benoitf, line 1882: AFAIK we need to attach before starting a container

benoitf: else you may not receive the output

jeffmaury: The reason for the switch is the following:

- [Docker documentation](https://docs.docker.com/reference/cli/docker/container/attach/) mentions attach is for running containers
- podman attach (CLI) rejects if container is not running
- current behavoir in Podman Desktop is attach API call never returns

benoitf: you're correct, I checked the cli and order is the following

1.POST /v1.41/containers/create HTTP/1.1\r
2. POST /v1.41/containers/1784f98a7d5f2fa80950cc55a7ab066a6acf841066403ed61b3b04cf3b47a262/attach?stderr=1&stdin=1&stdout=1&stream=1 HTTP/1.1\r

benoitf (approved):

PR #7353, opened 2024-05-29
refactor(ListItemButtonIcon): remove confirm property

axel7083:
### What does this PR do?

The `ListItemButtonIcon` uses internal api (window.showMessageBox) for a confirmation dialog, this PR move this logic to parent components.

Required for https://github.com/containers/podman-desktop/issues/6898

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/dialogs/MessageBoxUtils.ts
+++ b/packages/renderer/src/lib/dialogs/MessageBoxUtils.ts
@@ -1,0 +1,31 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export function withConfirmation(func: () => unknown, action: string): void {
+  window
+    .showMessageBox({
+      title: 'Confirmation',
+      message: 'Are you sure you want to ' + action + '?',
+      buttons: ['Yes', 'Cancel'],
+    })
+    .then(result => {
+      if (result && result.response === 0) {
+        func();
+      }
+    });
+}
--- a/packages/renderer/src/lib/compose/ComposeActions.svelte
+++ b/packages/renderer/src/lib/compose/ComposeActions.svelte
@@ -5,7 +5,7 @@ import { createEventDispatcher, onMount } from 'svelte';
 import { router } from 'tinro';
 
 import ContributionActions from '/@/lib/actions/ContributionActions.svelte';
-import { withConfirmation } from '/@/lib/dialogs/MessageBoxUtils';
+import { withConfirmation } from '/@/lib/dialogs/messagebox-utils';
 
 import type { Menu } from '../../../../main/src/plugin/menu-registry';
 import { MenuContext } from '../../../../main/src/plugin/menu-registry';
--- a/packages/renderer/src/lib/compose/ComposeActions.svelte
+++ b/packages/renderer/src/lib/compose/ComposeActions.svelte
@@ -136,7 +136,7 @@ if (dropdownMenu) {
 
 <ListItemButtonIcon
   title="Delete Compose"
-  onClick="{() => withConfirmation(deleteCompose, 'Delete Compose')}"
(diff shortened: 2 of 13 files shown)


packages/renderer/src/lib/dialogs/MessageBoxUtils.ts
```diff
@@ -0,0 +1,31 @@
+/**********************************************************************
```
benoitf: AFAIK file should be all lower case, it's not a svelte component

packages/renderer/src/lib/dialogs/MessageBoxUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+      title: 'Confirmation',
+      message: 'Are you sure you want to ' + action + '?',
+      buttons: ['Yes', 'Cancel'],
+    })
+    .then(result => {
+      if (result && result.response === 0) {
+        func();
+      }
+    });
+}
```
benoitf: I expect to have something being really asynchronous (return a promise) and resolving when user clicked

axel7083: Hummm, but it should go inside the `on:click` therefore cannot return a promise ? I am not sure to follow what you expect

axel7083: cc @benoitf

benoitf: I could call this method from any place as it's exported so I expect the method returns when it's done

```
async withConfirmation(...): Promise<void> {
   const result = await window.showMessageBox({});
   if (result ...) {
    ....
   }
....
}

```

then it's up to you to decide to call it without expecting a response in on:click methods
(but I'm not sure exported method should be without proper async calls)


Also there is something that will fail in case of error as it's not handled in the content of this method. (no catch method)


or if you really want to not wait, probably there should be two, one with proper async, and another being suffixed to say we don't expect a result

axel7083: Another solution could be to have withConfirmation(callback: (err?: unknown) => void).

If you want to have it with an async you can do the following

```
import { promisify } from 'node:util'

const asyncConfirmation = promisify(withConfirmation)
```

> or if you really want to not wait, probably there should be two, one with proper async, and another being suffixed to say we don't expect a result

as I would avoid exporting two, I would prefer stay with the basic callback mechanism what do you think ?

jeffmaury: Why do you want to introduce callback when the code base is using async ?

axel7083: > Why do you want to introduce callback when the code base is using async ?

Because you want to be called back, after the confirmation dialog, on success.

jeffmaury: then return a Promise<boolean>

axel7083: This function would be placed in `on:click="{here}"` it cannot return a promise, otherwise you would have a typecheck error. 

The goal here is to have some kind of machisme calling back a given method after a confirmation dialog, if we use a promise then we would have

  `on:click="{getConfirmation().then(myCallback).catch(err => console.error(err))}"` instead of `on:click={withConfirmation(myCallback)}` as it is used in 10 differents files, i would have a preference for the second option, as it is shorter :/

packages/renderer/src/lib/dialogs/MessageBoxUtils.ts
axel7083: renamed ✅

packages/renderer/src/lib/dialogs/MessageBoxUtils.ts
axel7083: I added documentation to clarify usage

axel7083:
cc @deboer-tim @jeffmaury

deboer-tim (changes requested):
This is a good pattern, and actually more readable to me in each of the actions.

Using `title.toLowerCase()` was the cheapest way to get the generic lowercase 'delete x' string from the titles instead of adding another property or making confirm a string. This PR currently 'regresses' it to upper case like 'Are you sure you want to Delete Pod?'. Could you please change all the strings back to lowercase, i.e. `'delete pod'`? (If you feel that terms like Service needs to be capitalized I'm ok with that, but delete definitely shouldn't be)

This is also a good opportunity to make them even more readable and friendly. Ideally they should be 'Are you sure you want to delete this pod?', or `delete pod '${pod.name}'` would be awesome.

FWIW when this was originally added we discussed the option of having a 'do not ask me again' option in the future, with a way to undo. It looks like we could still do that by (e.g.) passing in a preferences key as an optional third arg.

axel7083:
> This is also a good opportunity to make them even more readable and friendly. Ideally they should be 'Are you sure you want to delete this pod?', or `delete pod '${pod.name}'` would be awesome.

@deboer-tim suggestion applied ✅

deboer-tim (approved):
I'm wondering if we should surround the name with `'` so that it's obvious what part is the name (e.g. esp. if they call their pod 'pod' we'd say `delete pod 'pod'` vs `delete pod pod`). But please take my +1, really nice to include the usability improvement with the change. 👍🏼

PR #7379, opened 2024-05-30
feat: port forwarding backend part

### LINKED ISSUE #7078: Ability to configure port forwarding in kubernetes environment

Due to analysing previous feedbacks. There was a feature request that is missing in Podman Desktop. The feature called Port Forwarding. It should represent `kubectl port-forward` cli command.

Need to create the backend implementation, which will serve port forwards and UI dedicated page to control the port forwarding state.

Main things that need to be implemented:

- [ ] ability to store forward configurations (add/list/remove) https://github.com/containers/podman-desktop/pull/7379
- [ ] ability to start and stop saved port forward https://github.com/containers/podman-desktop/pull/7379
- [ ] dedicated page on the UI that shows the list of forwards
- [ ] dedicated form on the UI that allows to register new port forward based on the input type of service (pod/service/deployment)

_No response_

### Additional context

_No response_

### LINKED ISSUE #9272: Backend implementation of Kubernetes port forwarding

Port forwarding is a request feature from Kubernetes users. This issue is about implementing the logic on the backend side

Implement port forwarding configuration and related management verbs: create, list, delete, start, stop

_No response_

### Additional context

_No response_

vzhukovs:
### What does this PR do?

One of the several PR, that is decided to merge separately.

The purpose of this pull request is to provide the server side code, that responsible for enabling port forwarding in a Kubernetes environment. The main part of code can be divided into two logical components: 1) managing user configurations (creation, updating, listing, deletion); 2) starting the forwarding process for a specific configuration.

The configuration management is based on binding to the currently used context. In other words, if a user calls a method to obtain a service `KubernetesPortForwardServiceProvider.getService`, Kubernetes configuration must be provided as an argument to this service provider. If the service has already been created previously, it will be returned from the local cache to prevent duplication of the service itself.

All user configurations are stored by default in the application's configuration directory in the file `port-forwarding.json`. This file is essentially a dictionary where the keys are the context names, and the values are the user configurations.

The general principle of using this service is as follows:

1. In Kubernetes client should be created four methods that interact with the frontend side.
2. When frontend calls the particular method in Kubernetes client, first thing that does is that Kubernetes config passes to the port forward service provider and retrieves the particular service for the current context.
3. Calls the particular method from the port forward service.

### Screenshot / video of UI

No screenshots.

Fixes #9272  

### How to test this PR?

To be able to test the PR locally, this branch should be cloned locally and preliminary steps need to be performed:

1. Created minikube cluster
2. Applied the following deployment:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:latest
        ports:
        - name: http
          containerPort: 80
        volumeMounts:
        - name: demo-page
          mountPath: /usr/share/nginx/html
      volumes:
      - name: demo-page
        configMap:
          name: demo-page-config

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: demo-page-config
data:
  index.html: |
    <!DOCTYPE html>
    <html>
    <head>
        <title>Welcome to nginx!</title>
        <style>
            body {
                width: 35em;
                margin: 0 auto;
                font-family: Tahoma, Verdana, Arial, sans-serif;
            }
        </style>
    </head>
    <body>
    <h1>Welcome to nginx!</h1>
    <p>If you see this page, the nginx web server is successfully installed and
    working. Further configuration is required.</p>

    <p>For online documentation and support please refer to
    <a href="http://nginx.org/">nginx.org</a>.<br/>
    Commercial support is available at
    <a href="http://nginx.com/">nginx.com</a>.</p>

    <p><em>Thank you for using nginx.</em></p>
    </body>
    </html>
```

3. Applied the following service:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  selector:
    app: nginx
  ports:
  - name: http
    protocol: TCP
    port: 80
    targetPort: http
  type: LoadBalancer
```

4. Created test file in `packages/main/src/plugin`, e.g. `kubernetes-port-forward.spec.ts` with the following content:

```typescript
import * as k8s from '@kubernetes/client-node';
import { test } from 'vitest';

import type { ForwardConfig, PortMapping } from '/@/plugin/kubernetes-port-forward-model.js';
import { WorkloadKind } from '/@/plugin/kubernetes-port-forward-model.js';
import { KubernetesPortForwardServiceProvider } from '/@/plugin/kubernetes-port-forward-service.js';

test(
  'Port Forward test',
  async () => {
    const kc = new k8s.KubeConfig();
    kc.loadFromDefault();

    const provider = new KubernetesPortForwardServiceProvider();
    const portForwardService = provider.getService(kc);

    const config = {
      name: 'nginx-service',
      namespace: 'default',
      kind: WorkloadKind.SERVICE,
      forwards: [
        {
          localPort: 8892,
          remotePort: 80,
        } as PortMapping,
      ],
    } as ForwardConfig;

    const disposable = await portForwardService.startForward(config);
    await pauseForFiveMinutes();
    disposable.dispose();
  },
  5 * 60 * 1000,
);

function pauseForFiveMinutes(): Promise<void> {
  return new Promise(resolve => {
    setTimeout(resolve, 5 * 60 * 1000);
  });
}
```

5. Run the following test, it will run for a five minutes.
6. Check that URL `http://localhost:8892` provides the content of page.
7. Stop the test.
8. Check that `http://localhost:8892` no more reachable.


vzhukovs:
Some unit tests need to be updated to be run on Windows. Will provide updates for this ASAP.

benoitf:
closing as no activity

axel7083 (commented):
Great feature, it works nicely following the steps you provide to test

## Files location

We should move all the Kubernetes logic to `packages/main/src/plugin/kubernetes`, and maybe avoid the prefix `kubernetes-port-forward` ?

## Interoperability 

One of my concern is that the code added do not use the `KubernetesClient`, it read the kubeconfig, and use it to create a client, and use function like `readNamespacedPod`.

We seems to repeating logic, which may be due to the time when the PR has been opened. Maybe @feloy can give a review and see how we can make them communicates.

## Deployment port forwarding 

I do not think we should let the user port forward a deployment, it do not make much sense, as we should only allow pods and services.

OpenLens is not allowing user to port foward highter resources (deployment, statefulset) and I think we should do the same, only pods and services

## Port fowarding persistence

From my understanding of the code, we are saving the port foward in `port-forwards.json`, meaning we are restoring them on startup of podman-desktop ? This feature seems very complex, and might need to have a dedicated PR on its own, we probably can just have in memory port fowarding as a first step, which would probably simplify this PR.

## Global comment

Amazing documentation on functions, and classes. :tada: kudos :tada:

packages/main/src/plugin/kubernetes-port-forward-model.ts
```diff
@@ -0,0 +1,80 @@
+/**********************************************************************
```
axel7083: Interfaces that will be shared between the main and renderer package should be moved to `packages/api/src`

packages/main/src/plugin/kubernetes-port-forward-service.ts
```diff
@@ -0,0 +1,132 @@
...
+import type { IDisposable } from '/@/plugin/types/disposable.js';
+import { isFreePort } from '/@/plugin/util/port.js';
+
+/**
+ * Service provider for Kubernetes port forwarding.
+ * Entrypoint for the third party services to operate with port forwards.
+ * @see KubernetesPortForwardServiceProvider.getService
+ */
+export class KubernetesPortForwardServiceProvider {
+  protected _serviceMap = new Map<string, KubernetesPortForwardService>();
```
axel7083: do not need an underscore

packages/main/src/plugin/kubernetes/kubernetes-port-forward-storage.ts
```diff
@@ -0,0 +1,389 @@
...
+  protected getStoragePathInternal(): string {
+    return path.resolve(this.directories.getConfigurationDirectory(), 'port-forwards.json');
+  }
+}
+
+/**
+ * Class representing file-based configuration storage.
+ * @see ForwardConfigStorage
+ */
+export class FileBasedConfigStorage implements ForwardConfigStorage {
```
axel7083, line 160: to remove ? since there is MemoryBasedStorage ?

packages/main/src/plugin/kubernetes/kubernetes-port-forward-storage.ts
```diff
@@ -0,0 +1,389 @@
...
+   * @throws If the storage is not initialized.
+   */
+  getStoragePath(): string;
+}
+
+/**
+ * Class representing preference folder based storage.
+ * @see FileBasedStorage
+ */
+export class PreferenceFolderBasedStorage implements FileBasedStorage {
```
axel7083, line 89: to remove also?

packages/main/src/plugin/kubernetes/kubernetes-port-forward-storage.ts
```diff
@@ -0,0 +1,389 @@
...
+   * @returns A list of forward configurations.
+   * @see UserForwardConfig
+   */
+  listForwards(): Promise<UserForwardConfig[]>;
+}
+
+/**
+ * Interface for file-based storage.
+ */
+export interface FileBasedStorage {
```
axel7083, line 64: to remove ?

packages/main/src/plugin/kubernetes/kubernetes-port-forward-storage.ts
jeffmaury: It kept it so that we have most of the code when we want to switch but could be removed or add a note

packages/main/src/plugin/kubernetes/kubernetes-port-forward-connection.ts
```diff
@@ -0,0 +1,403 @@
...
+  /**
+   * Retrieves a pod by name and namespace.
+   * @param name - The name of the pod.
+   * @param namespace - The namespace of the pod.
+   * @returns The pod.
+   * @throws If the pod cannot be retrieved.
+   */
+  protected async getPod(name: string, namespace: string): Promise<V1Pod> {
+    return this.kubeClient.readNamespacedPod(name, namespace);
+  }
```
axel7083, line 224: should use `KubernetesClient#readNamespacedPod`

packages/main/src/plugin/kubernetes/kubernetes-port-forward-connection.ts
```diff
@@ -0,0 +1,403 @@
...
+  /**
+   * Retrieves a deployment by name and namespace.
+   * @param name - The name of the deployment.
+   * @param namespace - The namespace of the deployment.
+   * @returns The deployment.
+   * @throws If the deployment cannot be retrieved.
+   */
+  protected async getDeployment(name: string, namespace: string): Promise<V1Deployment> {
+    return this.kubeClient.readNamespacedDeployment(name, namespace);
+  }
```
axel7083, line 235: Is there really a point to port forward deployment ?

jeffmaury: This is implememented by kubectl so let's keep the code and decide at UI level

packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Gets the configuration key for the given Kubernetes configuration.
+   * The key is used for caching services and also for storing information about created forwards linked to each service.
+   *
+   * @param kubeConfig - The Kubernetes configuration.
+   * @returns The configuration key.
+   */
+  protected getKubeConfigKey(kubeConfig: KubeConfig): string {
+    return kubeConfig.currentContext;
+  }
```
axel7083, line 58: This is misleading, kubeconfigkey do not mean anything ? If we need the currentContext, we should use current context wording ?

packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * Service for managing Kubernetes port forwarding.
+ * @see KubernetesPortForwardServiceProvider.getService
+ */
+export class KubernetesPortForwardService {
```
axel7083: I would recommend to make `KubernetesPortForwardService` extend IDisposal, if we need to cleanup everything, the service should keep track of the disposals, and have `KubernetesPortForwardService#dispose` method to cleanup all port foward easily 

Currently the disposal is only provided through the `KubernetesPortForwardService#startForward` but nothing is saving them to cleanup resources.

packages/main/src/plugin/kubernetes/kubernetes-port-forward-connection.ts
axel7083: Oh I was not aware, thanks for the info

axel7083: Oh I was not aware, thanks for the info

packages/main/src/plugin/kubernetes/kubernetes-port-forward-model.ts
```diff
@@ -0,0 +1,80 @@
+/**********************************************************************
```
axel7083: Some of the models here will be used by the frontend, this file need to be moved to `packages/api/src/`

axel7083 (commented):
We have a lack of events; I am not  sure which issue is responsible of providing the communication between the frontend and the backend, but currently with this PR, the backend do not create any instance nor produce events that the frontend can consume.

Inside `packages/main/src/plugin/index.ts` we would need something like the following

```ts
    const kubernetesPortForwardService: KubernetesPortForwardService = new KubernetesPortForwardServiceProvider().getService(kubernetesClient);
```

Should this be a follow up PR ?

and a series of events, for example, the deployment subscribe to the following events `kubernetes-current-context-deployments-update` and have the following ipc method `window.kubernetesRegisterGetCurrentContextResources('deployments')` that is used to update the store.

packages/main/src/plugin/kubernetes/kubernetes-port-forward-storage.ts
axel7083: 👍

jeffmaury:
> We have a lack of events; I am not sure which issue is responsible of providing the communication between the frontend and the backend, but currently with this PR, the backend do not create any instance nor produce events that the frontend can consume.
> 
> Inside `packages/main/src/plugin/index.ts` we would need something like the following
> 
> ```ts
>     const kubernetesPortForwardService: KubernetesPortForwardService = new KubernetesPortForwardServiceProvider().getService(kubernetesClient);
> ```
> 
> Should this be a follow up PR ?
> 
> and a series of events, for example, the deployment subscribe to the following events `kubernetes-current-context-deployments-update` and have the following ipc method `window.kubernetesRegisterGetCurrentContextResources('deployments')` that is used to update the store.

I was thinking to embed the service inside KubernetesClient so that port forwarding is cleaned up upcon context change WDYU

axel7083:
> I was thinking to embed the service inside KubernetesClient so that port forwarding is cleaned up upcon context change WDYU

I like this idea, this kinda makes it all through the client 👍 ! We still need to have the KubernetesPortForwardService firing events, that the frontend can consume.

axel7083 (approved):
As per discussion, a follow-up PR will be necessary to provide the events and api the frontend can consume.

LGTM

package.json
```diff
@@ -162,6 +162,7 @@
     "lint-staged": "^15.2.10",
     "minimist": "^1.2.8",
     "nock": "14.0.0-beta.7",
+    "node-fetch": "^3.3.2",
```
benoitf, line 165: why this library is being included ?

vzhukovs: This library is used to check the availability of exposed port before showing the result to the user. There is no reason to show the user result if exposed port is not reachable.

benoitf: fetch is part of Node.js so we don't need a 3rd party library

PR #7383, opened 2024-05-30
chore: move CloseButton to the ui library

benoitf:
### What does this PR do?
also cut the dependency to tinro and parent element need to handle the dispatch

need it for https://github.com/containers/podman-desktop/issues/6920

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/button/CloseButton.svelte
+++ b/packages/ui/src/lib/button/CloseButton.svelte
@@ -2,18 +2,11 @@
 import { faTimes } from '@fortawesome/free-solid-svg-icons';
 import { createEventDispatcher } from 'svelte';
 import Fa from 'svelte-fa';
-import { router } from 'tinro';
-
-export let href: string | undefined = undefined;
 
 const dispatch = createEventDispatcher<{ click: undefined }>();
 
-function click() {
-  if (href) {
-    router.goto(href);
-  } else {
-    dispatch('click');
-  }
+function click(): void {
+  dispatch('click');
 }
 </script>
 
@@ -21,7 +14,7 @@ function click() {
   type="button"
   class="text-gray-800 hover:bg-white hover:bg-opacity-10 transition-all rounded-[4px] p-1 no-underline cursor-pointer {$$props.class ||
     ''}"
-  on:click="{() => click()}"
+  on:click="{click}"
   title="Close"
   aria-label="Close">
   <Fa icon="{faTimes}" />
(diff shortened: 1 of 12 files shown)


packages/ui/src/lib/button/CloseButton.svelte
```diff
@@ -2,26 +2,21 @@
 import { faTimes } from '@fortawesome/free-solid-svg-icons';
 import { createEventDispatcher } from 'svelte';
 import Fa from 'svelte-fa';
-import { router } from 'tinro';
 
-export let href: string | undefined = undefined;
+export let href: string | undefined = 'undefined';
```
lstocchi: I wonder why we are keeping the href here. Before there was a different behavior based on the value href had. If it was undefined a click was dispatched, if it was defined it worked like a link. 
In the new way we always trigger the click. Can we remove this href?

benoitf: ah it's a leftover from a test, it should not be 'undefined'

benoitf: so now I see it's a suggestion

the idea is that if you expect an href you'll get a string, else you receive undefined

benoitf: you always get a click event, then it's either containing a value or undefined

benoitf: I kept the href so when we get the click event we receive the value, the won't need to keep it on the parent component as an attribute

benoitf: but we could replace `href` by `clickDetail` or `detail`

lstocchi: it looks to me too complex tbh, we could just create the close button component with a callback that gets called on the click.
If you want to keep the current logic, yes maybe renaming would reduce the confusion imo

benoitf: > it looks to me too complex tbh, we could just create the close button component with a callback that gets called on the click.

it's what it does, it call the optional on:click method

but I can remove the field

waiting for other reviews to decide

lstocchi: I was referring to a simplification like the following (so we remove the href, dispatch and click() from the CloseButton) 
```
export let onClose: () => {};

on:click="{onClose}"
```

then from the caller
```
<CloseButton
   onClose=() => { .... }
```

benoitf:
so I used the same logic than the <Link component, we don't provide anymore the href field and we have `on:click` event (as any button)

packages/ui/src/lib/button/CloseButton.svelte
benoitf: I changed the implementation to match Link implementation

lstocchi (approved):
LGTM

deboer-tim (approved):

PR #7384, opened 2024-05-30
feat: allow extensions to expose their own API

### LINKED ISSUE #5990: Provide API to call podman CLI

`podman machine ssh` command has no matching Rest API end point. If an extension needs to execute it or similar command the only way is to find podman CLI in user environment and run it using process API. 

podman extension has implementation already that should be accessible through podman desktop API.

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
allow extensions to expose their own API
and other 3rd party extensions can use these endpoints/API

it also offers a way to list all Podman Desktop extensions

fixes https://github.com/containers/podman-desktop/issues/5990

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -171,6 +171,82 @@ declare module '@podman-desktop/api' {
   }
 
   /**
+   * Represents an extension.
+   *
+   * To get an instance of an `Extension` use {@link extensions.getExtension getExtension}.
+   */
+  export interface Extension<T> {
+    /**
+     * The canonical extension identifier in the form of: `publisher.name`.
+     */
+    readonly id: string;
+
+    /**
+     * The uri of the directory containing the extension.
+     */
+    readonly extensionUri: Uri;
+
+    /**
+     * The absolute file path of the directory containing this extension. Shorthand
+     * notation for {@link Extension.extensionUri Extension.extensionUri.fsPath} (independent of the uri scheme).
+     */
+    readonly extensionPath: string;
+
+    /**
+     * The parsed contents of the extension's package.json.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    readonly packageJSON: any;
+
+    /**
+     * The public API exported by this extension (return value of `activate`).
+     * It is an invalid action to access this field before this extension has been activated.
+     */
+    readonly exports: T;
+  }
+
+  /**
+   * Namespace for dealing with installed extensions. Extensions are represented
+   * by an {@link Extension}-interface which enables reflection on them.
+   *
+   * Extension writers can provide APIs to other extensions by returning their API public
+   * surface from the `activate`-call.
+   *
+   * When depending on the API of another extension add an `extensionDependencies`-entry
+   * to `package.json`, and use the {@link extensions.getExtension getExtension}-function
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/extension-loader.ts
```diff
@@ -1457,16 +1509,20 @@ export class ExtensionLoader {
...
         deactivateFunction,
         extensionContext,
+        exports,
       };
       this.activatedExtensions.set(extension.id, activatedExtension);
       this.extensionState.set(extension.id, 'started');
       this.apiSender.send('extension-started');
     } catch (err) {
       console.log(`Activating extension ${extension.id} failed error:${err}`);
+      this._onDidChange.fire();
```
axel7083: Why fire an event on error ? Nothing has change as it was not completed ?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -170,6 +170,82 @@ declare module '@podman-desktop/api' {
...
+   * to `package.json`, and use the {@link extensions.getExtension getExtension}-function
+   * and the {@link Extension.exports exports}-property, like below:
+   *
+   * ```typescript
+   * const podmanExtension = extensions.getExtension('podman-desktop.podman');
+   * const podmanExtensionAPI = podmanExtension.exports;
+   *
+   * podmanExtensionAPI....
+   * ```
+   */
```
axel7083, line 225: Great having everything documented !!

packages/main/src/plugin/extension-loader.ts
benoitf: fixed

packages/extension-api/src/extension-api.d.ts
```diff
@@ -170,6 +170,82 @@ declare module '@podman-desktop/api' {
...
+     * The absolute file path of the directory containing this extension. Shorthand
+     * notation for {@link Extension.extensionUri Extension.extensionUri.fsPath} (independent of the uri scheme).
+     */
+    readonly extensionPath: string;
+
+    /**
+     * The parsed contents of the extension's package.json.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    readonly packageJSON: any;
```
axel7083, line 199: could we expose the version as a property of Extension ? I would rather do `extension.version` with a nice typing than `extension.packageJSON.version?`

benoitf: I think it would be a separate enhancement not related to this one

Also I don't think an extension should check versions of other extensions

axel7083: I think they totally should ? Let's say AI Lab version 1.1 expose "GetInferenceServers()" but not version 1.0... I might want to handle that case, and preferably not by doing (extension.GetInferenceServers === undefined)

I don't really know how we want to manage typing, by I imagine the ai lab should publish @types package ? providing the methods types it is exporting ?

feloy (approved):
Tested by returning an API from an extension and calling it from another extension. Works as expected.

I'm not sure though how to share the typing of the API published by the extension.

benoitf:
to share API, extension need to publish a package to npmjs and then other extensions can depend on it

axel7083 (approved):
I don't want to block this any further, but I really think the versioning is an import issue. And we might want to find a nice way to handle it to avoid incompatibility issues in the future

Issue #6374, opened 2024-03-11
add rule to prefer the `??` nullish coalescing operator over the OR `||` operator


benoitf:
### Is your enhancement related to a problem? Please describe

Add a rule to enforce that

### Describe the solution you'd like

include a linter rule

PR #7385, opened 2024-05-30
chore: add strict to Compose, Lima, Kubectl, KubeContext and Registries extensions

lstocchi:
### What does this PR do?

This is one of the PRs to complete https://github.com/containers/podman-desktop/issues/6374
To enable the `prefer-nullish-coalescing` rule we have to enable the `strictNullChecks`. 
This PR just covers  Compose, Lima, Kubectl and KubeContext extensions. 

### Screenshot / video of UI

N/A

part of #6374 

### How to test this PR?

N/A

### DIFF
--- a/extensions/compose/src/compose-wrapper-generator.ts
+++ b/extensions/compose/src/compose-wrapper-generator.ts
@@ -19,7 +19,8 @@
 import { promises } from 'node:fs';
 
 import type * as extensionApi from '@podman-desktop/api';
-import { render } from 'mustache';
+// @ts-expect-error ignore type error https://github.com/janl/mustache.js/issues/797
+import mustache from 'mustache';
 
 import type { OS } from './os';
 import batMustacheTemplate from './templates/podman-compose.bat.mustache?raw';
@@ -44,7 +45,7 @@ export class ComposeWrapperGenerator {
     }
 
     // render the template
-    return render(template, { socketPath, binFolder: this.binFolder });
+    return mustache.render(template, { socketPath, binFolder: this.binFolder });
   }
 
   async generate(connection: extensionApi.ProviderContainerConnection, path: string): Promise<void> {
--- a/extensions/compose/src/compose-wrapper-generator.ts
+++ b/extensions/compose/src/compose-wrapper-generator.ts
@@ -19,7 +19,7 @@
 import { promises } from 'node:fs';
 
 import type * as extensionApi from '@podman-desktop/api';
-import mustache from 'mustache';
+import { render } from 'mustache';
 
 import type { OS } from './os';
 import batMustacheTemplate from './templates/podman-compose.bat.mustache?raw';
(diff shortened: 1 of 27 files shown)


benoitf (approved):

cdrage (approved):

extensions/compose/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "strictNullChecks": true,
```
jeffmaury: strict would be prefered as it enable other checks as well

lstocchi: updated

axel7083 (approved):
Nice ! I dream one of one having strict: true everywhere ahah !

extensions/kubectl-cli/types/shell-path.d.ts
```diff
@@ -0,0 +1,22 @@
+/**********************************************************************
```
axel7083, line 1: maybe it should in the `types` folder at the root ?

extensions/registries/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "strictNullChecks": true,
```
axel7083: why this one it not `strict` ? too much changes ?

extensions/kubectl-cli/types/shell-path.d.ts
benoitf: extensions need to be 'independent'

extensions/compose/src/compose-wrapper-generator.ts
```diff
@@ -44,7 +44,7 @@ export class ComposeWrapperGenerator {
     }
 
     // render the template
-    return mustache.render(template, { socketPath, binFolder: this.binFolder });
+    return render(template, { socketPath, binFolder: this.binFolder });
```
benoitf: looks like it's failing using that syntax

extensions/registries/tsconfig.json
lstocchi: Ops, i forgot it. Updated, thanks!! 👍

jeffmaury (approved):
Mustache requires an update, otherwise LGTM

extensions/compose/src/compose-wrapper-generator.ts
lstocchi: maybe we should downgrade mustache https://github.com/janl/mustache.js/issues/797 👀

lstocchi: I haven't found a nice way to make it work. As it works at runtime and it seems there is an error in the typings i just added a command to ignore the error. If you have any suggestion to make it another way let me know

jeffmaury: Is the change mandatory ?

lstocchi: i guess so, it complains that there are no mustache types if i leave the code as it was... but then if i add the types they are missing the default export

benoitf (approved):
would be interesting to have a better solution for mustache but can be a follow up

PR #7391, opened 2024-05-30
chore: add icons for service k8s type

cdrage:
chore: add icons for service k8s type

### What does this PR do?

Adds small icons rather than dots to better represent the type.

### Screenshot / video of UI

Screenshot 2024-05-30 at 3 15 11 PM

Part of https://github.com/containers/podman-desktop/issues/7386

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/service/ServiceColumnType.spec.ts
+++ b/packages/renderer/src/lib/service/ServiceColumnType.spec.ts
@@ -43,7 +43,7 @@ test('Expect basic column styling', async () => {
 
   const dot = text.parentElement?.children[0].children[0];
   expect(dot).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


deboer-tim (approved):
I don't love the pattern with getTypeAttributes() getting called twice to get one of two properties each time, but since it is already done in DeploymentColumnConditions that's ok. Feels like it's time to create a column-badge-colored-icon component soon.

cdrage:
> I don't love the pattern with getTypeAttributes() getting called twice to get one of two properties each time, but since it is already done in DeploymentColumnConditions that's ok. Feels like it's time to create a column-badge-colored-icon component soon.

Yeah :( I'm hoping that once we get the UI parts finalized we can start picking out some things we can refactor, including the "badge" portion where we pass in the name, icon, etc.

PR #7392, opened 2024-05-30
chore: add icon to url for k8s ingress / routes

### LINKED ISSUE #7387: Make URL on Kubernetes ingress section into a "badge"

Every page on Podman Desktop we use a badge to get the user's attention, either the type, provider, etc:

Screenshot 2024-05-30 at 2 45 20 PM

We should have the same for ingress:
Screenshot 2024-05-30 at 2 45 06 PM

Same "badge" style as other sections.

_No response_

### Additional context

_No response_

cdrage:
chore: add icon to url for k8s ingress / routes

### What does this PR do?

Adds little icon to the URL to make it similar to other aspects of
podman desktop, where we have the "badge" for the most important
information on the page (type, provider, etc.)

### Screenshot / video of UI

Screenshot 2024-05-30 at 4 04 52 PM

Closes https://github.com/containers/podman-desktop/issues/7387

### How to test this PR?

View ingress / route page

### DIFF
--- a/packages/renderer/src/lib/ingresses-routes/IngressRouteColumnHostPath.spec.ts
+++ b/packages/renderer/src/lib/ingresses-routes/IngressRouteColumnHostPath.spec.ts
@@ -63,7 +63,7 @@ test('Expect simple column styling with single host/path ingress', async () => {
   const hostPath = ingressRouteUtils.getHostPaths(ingressUI)[0];
   const link = screen.getByLabelText(hostPath.label);
   expect(link).toBeInTheDocument();
-  expect(link.textContent).toBe(hostPath.label);
+  expect(link.textContent).toContain(hostPath.label);
 });
 
 test('Expect simple column styling with multiple paths ingress', async () => {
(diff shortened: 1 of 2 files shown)


lstocchi (approved):
LGTM

deboer-tim:
This does make it more visually appealing but I have mixed feelings about this - we have buttons, links, and now badges could also be clickable (it's a badge with a link, but that's what it looks like). We're also not very consistent on what column might be badges or when to use the link icon prior to this either.

@ekidneyrh could you provide some guidance here on if this is the right approach or what else to do?

axel7083 (approved):

ekidneyrh:
I don't think there's much reason to draw the users' attention to the link the same way we do with the badges. I also agree with Tim that the badges aren't buttons, and shouldn't act as such. This should remain styled as a link for the time being, IMO.

I'll open a ticket to set a clear category system and rules for badges, and another for links.

cdrage:
> I don't think there's much reason to draw the users' attention to the link the same way we do with the badges. I also agree with Tim that the badges aren't buttons, and shouldn't act as such. This should remain styled as a link for the time being, IMO.
> 
> I'll open a ticket to set a clear category system and rules for badges, and another for links.

No problem! I'll close this for now. Thank you for your insight

PR #7405, opened 2024-05-31
chore(tests): add env var to configure tests

### LINKED ISSUE #7404: It should be possible to run e2e tests without cleaning up test folder beforehand

Currently when a test run is initialized some test folders are being cleaned up, however this setting is not configurable and sometimes the cleanup of those folders is not wanted.

Add an environment variable that when set will bypass the initial folder cleanup.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Checks for an environment variable before starting a test run to check if the test session is for debug purposes to not cleanup some test folders.

https://github.com/containers/podman-desktop/issues/7404

### DIFF
--- a/tests/playwright/src/globalSetup/global-setup.ts
+++ b/tests/playwright/src/globalSetup/global-setup.ts
@@ -26,7 +26,7 @@ export async function setup(): Promise<void> {
     // remove all previous testing output files
     // Junit reporter output file is created before we can clean up output folders
     // It is not possible to remove junit output file because it is opened by the process already, at least on windows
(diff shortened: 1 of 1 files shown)


benoitf (commented):
should it be named like `IGNORE_REMOVE_FOLDER` or something like that ?

cbr7:
@benoitf fair, I change the name of the env var

benoitf (approved):
I know I suggested the name but maybe should be `SKIP_REMOVE_FOLDER` but anyway it's not a required change

PR #7412, opened 2024-05-31
chore: k8s deployments to say 'Progressed', sort columns

### LINKED ISSUE #7410: Kubernetes deployment should say progressed not progressing

For Kubernetes the status says: "Successfully progressed":

Screenshot 2024-05-31 at 8 45 09 AM

But in the label we say "Progressing". Which is really confusing.

We should check the Reason and if the reason is "NewReplicaSetAvailable" then it should say **progressed** instead.

 

Say progressed instead of progressing.

_No response_

### Additional context

See documentation here with regards when a deployment has successfully progressed:

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment

### LINKED ISSUE #7411: Sort Kubernetes deployment conditions before showing them

Screenshot 2024-05-31 at 8 50 04 AM

We "flip-flop" between the ordering.

Within k8s this is common as sometimes in the  array the Progressing / Available condition could be before the other condition.

Sort before displaying.

_No response_

### Additional context

_No response_

cdrage:
chore: k8s deployments to say 'Progressed', sort columns

### What does this PR do?

* Sorts the columns by condition (available should be first, then
progressing, etc.)
* Say progressed if it has progressed..
* Updates tests so that they are individual cases against deploymentUI
  rather than using a single one.

### Screenshot / video of UI

Screenshot 2024-05-31 at 9 42 10 AM

Closes https://github.com/containers/podman-desktop/issues/7410
Closes https://github.com/containers/podman-desktop/issues/7411

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
@@ -3,25 +3,31 @@ import { faCheckCircle, faExclamationTriangle, faQuestionCircle, faSync } from '
 import { Tooltip } from '@podman-desktop/ui-svelte';
 import Fa from 'svelte-fa';
 
-import type { DeploymentUI } from './DeploymentUI';
+import type { DeploymentCondition, DeploymentUI } from './DeploymentUI';
 
 export let object: DeploymentUI;
 
 // Determine both the icon and color based on the deployment condition
-function getConditionAttributes(type: string) {
-  switch (type) {
+function getConditionAttributes(condition: DeploymentCondition) {
+  let name = condition.type;
+  switch (condition.type) {
     case 'Available':
       // faCheckCircle: Indicates a successful state, typically used to denote availability and operational readiness
-      return { color: 'text-green-600', icon: faCheckCircle };
+      return { color: 'text-green-600', icon: faCheckCircle, name };
     case 'Progressing':
       // faSync: Often used to represent ongoing processes or operations, fitting for a "Progressing" state
-      return { color: 'text-sky-400', icon: faSync };
+      // If reason has NewReplicaSetAvailable then it's progressed
+      if (condition.reason === 'NewReplicaSetAvailable') {
+        name = 'Progressed';
+      }
+
+      return { color: 'text-sky-400', icon: faSync, name };
     case 'ReplicaFailure':
       // faExclamationTriangle: Alerts and warnings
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
```diff
@@ -3,25 +3,35 @@ import { faCheckCircle, faExclamationTriangle, faQuestionCircle, faSync } from '
...
 import Fa from 'svelte-fa';
 
-import type { DeploymentUI } from './DeploymentUI';
+import type { DeploymentCondition, DeploymentUI } from './DeploymentUI';
 
 export let object: DeploymentUI;
 
+// Make we sort the conditions by type in the object itself before doing anything else.
+// we do not do this with onMount since it will "flicker" the UI / not be performant
+object.conditions.sort((a, b) => a.type.localeCompare(b.type));
```
deboer-tim: Shouldn't this be done in deployment-utils.ts?

deboer-tim (commented):
Kicked failed test, looks like a flake.

One comment, plus date on both .ts files should be updated to add 2024.

cdrage:
You're right, it should of been in the utils class. Updated the PR to move it to utils as well as the copyright @deboer-tim ready for another review 👍

deboer-tim (commented):
The copyright should be `2003,2004` or `2003-2004`, since it was created or updated in both years. Also need to fix the lint error.

cdrage:
> The copyright should be `2003,2004` or `2003-2004`, since it was created or updated in both years. Also need to fix the lint error.

Updated 👍

axel7083 (approved):

packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
```diff
@@ -3,25 +3,31 @@ import { faCheckCircle, faExclamationTriangle, faQuestionCircle, faSync } from '
...
 
-import type { DeploymentUI } from './DeploymentUI';
+import type { DeploymentCondition, DeploymentUI } from './DeploymentUI';
 
 export let object: DeploymentUI;
 
 // Determine both the icon and color based on the deployment condition
-function getConditionAttributes(type: string) {
-  switch (type) {
+function getConditionAttributes(condition: DeploymentCondition) {
```
feloy, line 11: In this function, we are considering that the condition is always "True", but the status of the condition could be "False", which would give another meaning

Issue #4966, opened 2023-11-23
Navigation path API


deboer-tim:
### Is your enhancement related to a problem? Please describe

Our code is littered with references to specific routes like `/volumes/:name/:engineId/*`, and they are referenced in multiple places - where the page is defined in the nav and in every button or component that needs to link to the page. This is:
- Making it harder to find references to a page.
- Making the code brittle and hard to refactor/move things.
- Leading to bugs due to different or missing encodings.

### Describe the solution you'd like

See first commit of PR #4371 as an option. We should have a common component(s) and pattern for all of the main pages you can navigate too. Instead of code saying:
  "go to /volumes/{encode name}/{encode engine}"
it should be something like:
  "go to VolumeRoute.get(volume)" or "go to NavRoutes.getVolumeURL(volume)"

deboer-tim:
Still valid.

deboer-tim:
To clarify, navigation API now exists but we should use it.

PR #7424, opened 2024-06-03
feat: container page table component

### LINKED ISSUE #4843: Move Containers to table component

deboer-tim:
### What does this PR do?

Switch the Containers page to use the Table component. This is a final piece
of technical debt adopting the Table component, and gives a few advantages:
- Table expands/collapses to fit the display width (used to extend off the
  right edge with a scrollbar).
- Sorting of name column to start (wanted to focus on core aspects and get
  this in before sorting other columns).
- Consistent table UI with other pages.
- Basic support for light mode.

The containers table included ContainerGroupInfoUI parents of three types
(pods, compose groups, and standalone containers), all with ContainerInfoUI
children. To avoid UI issues the parent ContainerGroupInfoUI was skipped
when drawing standalone containers (i.e. it only draws the child container
object in this one case).

Since the Table component doesn't support this unique behaviour, the least
invasive/simplest option was to create a derived list where standalone
containers are mapped to actual containers.

Found an issue where Svelte was updating constantly; turns out this was
an existing bug in the timeout when no containers are running.

Ignore changes to Table - reviewed separately under [#7587](https://github.com/deboer-tim/desktop/issues/7587).

### Screenshot / video of UI

Before:

Screenshot 2024-06-11 at 11 29 19 AM

After:

Screenshot 2024-06-11 at 11 49 38 AM

Fixes #4843.

### How to test this PR?

Try Containers page for regular containers, pods, and compose.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerColumnActions.svelte
+++ b/packages/renderer/src/lib/container/ContainerColumnActions.svelte
@@ -1,52 +1,24 @@
 <script lang="ts">
-import { ErrorMessage } from '@podman-desktop/ui-svelte';
-
-import ComposeActions from '../compose/ComposeActions.svelte';
-import PodActions from '../pod/PodActions.svelte';
-import ContainerActions from './ContainerActions.svelte';
+import { ContainerUtils } from './container-utils';
+import ContainerColumnActionsCompose from './ContainerColumnActionsCompose.svelte';
+import ContainerColumnActionsContainer from './ContainerColumnActionsContainer.svelte';
+import ContainerColumnActionsPod from './ContainerColumnActionsPod.svelte';
 import type { ContainerGroupInfoUI, ContainerInfoUI } from './ContainerInfoUI';
 import { ContainerGroupInfoTypeUI } from './ContainerInfoUI';
 
 export let object: ContainerInfoUI | ContainerGroupInfoUI;
+
+const containerUtils = new ContainerUtils();
 </script>
 
-{#if object.type === ContainerGroupInfoTypeUI.POD && object.engineId && object.id && object.shortId && object.status && object.engineName && object.humanCreationDate && object.created}
-  <PodActions
-    pod="{{
-      id: object.id,
-      shortId: object.shortId,
-      status: object.status,
-      name: object.name,
-      engineId: object.engineId,
-      engineName: object.engineName,
-      age: object.humanCreationDate,
-      created: object.created,
-      selected: false,
-      containers: object.containers.map(container => ({
-        Id: container.id,
-        Names: container.name,
-        Status: container.state,
-      })),
-      kind: 'podman',
-    }}"
-    dropdownMenu="{true}"
-    on:update />
-{/if}
-{#if object.type === ContainerGroupInfoTypeUI.COMPOSE && object.status && object.engineId && object.engineType}
-  <ComposeActions
-    compose="{{
-      status: object.status,
-      name: object.name,
-      engineId: object.engineId,
-      engineType: object.engineType,
-      containers: object.containers,
-    }}"
-    dropdownMenu="{true}"
-    on:update />
-{/if}
-{#if object.state}
-  {#if object.actionError}
-    <ErrorMessage error="{object.actionError}" icon />
+{#if containerUtils.isContainerGroupInfoUI(object)}
+  {#if object.type === ContainerGroupInfoTypeUI.POD}
+    <ContainerColumnActionsPod object="{object}" />
   {/if}
-  <ContainerActions container="{object}" dropdownMenu="{true}" on:update />
+  {#if object.type === ContainerGroupInfoTypeUI.COMPOSE}
+    <ContainerColumnActionsCompose object="{object}" />
+  {/if}
+{/if}
+{#if containerUtils.isContainerInfoUI(object)}
+  <ContainerColumnActionsContainer object="{object}" />
 {/if}
--- a/packages/renderer/src/lib/container/ContainerColumnActions.svelte
+++ b/packages/renderer/src/lib/container/ContainerColumnActions.svelte
@@ -13,12 +13,10 @@ const containerUtils = new ContainerUtils();
 
 {#if containerUtils.isContainerGroupInfoUI(object)}
   {#if object.type === ContainerGroupInfoTypeUI.POD}
-    <ContainerColumnActionsPod object="{object}" />
+    <ContainerColumnActionsPod object="{object}" on:update />
+  {:else if object.type === ContainerGroupInfoTypeUI.COMPOSE}
+    <ContainerColumnActionsCompose object="{object}" on:update />
   {/if}
-  {#if object.type === ContainerGroupInfoTypeUI.COMPOSE}
-    <ContainerColumnActionsCompose object="{object}" />
-  {/if}
-{/if}
-{#if containerUtils.isContainerInfoUI(object)}
-  <ContainerColumnActionsContainer object="{object}" />
+{:else if containerUtils.isContainerInfoUI(object)}
+  <ContainerColumnActionsContainer object="{object}" on:update />
 {/if}
--- a/packages/renderer/src/lib/container/ContainerColumnActions.svelte
+++ b/packages/renderer/src/lib/container/ContainerColumnActions.svelte
@@ -1,0 +1,52 @@
+<script lang="ts">
+import { ErrorMessage } from '@podman-desktop/ui-svelte';
+
+import ComposeActions from '../compose/ComposeActions.svelte';
+import PodActions from '../pod/PodActions.svelte';
+import ContainerActions from './ContainerActions.svelte';
+import type { ContainerGroupInfoUI, ContainerInfoUI } from './ContainerInfoUI';
+import { ContainerGroupInfoTypeUI } from './ContainerInfoUI';
+
+export let object: ContainerInfoUI | ContainerGroupInfoUI;
+</script>
+
+{#if object.type === ContainerGroupInfoTypeUI.POD && object.engineId && object.id && object.shortId && object.status && object.engineName && object.humanCreationDate && object.created}
+  <PodActions
+    pod="{{
+      id: object.id,
+      shortId: object.shortId,
+      status: object.status,
+      name: object.name,
+      engineId: object.engineId,
+      engineName: object.engineName,
+      age: object.humanCreationDate,
+      created: object.created,
+      selected: false,
+      containers: object.containers.map(container => ({
+        Id: container.id,
+        Names: container.name,
+        Status: container.state,
+      })),
+      kind: 'podman',
+    }}"
+    dropdownMenu="{true}"
+    on:update />
+{/if}
+{#if object.type === ContainerGroupInfoTypeUI.COMPOSE && object.status && object.engineId && object.engineType}
+  <ComposeActions
+    compose="{{
+      status: object.status,
+      name: object.name,
+      engineId: object.engineId,
+      engineType: object.engineType,
+      containers: object.containers,
+    }}"
+    dropdownMenu="{true}"
+    on:update />
+{/if}
+{#if object.state}
+  {#if object.actionError}
+    <ErrorMessage error="{object.actionError}" icon />
+  {/if}
+  <ContainerActions container="{object}" dropdownMenu="{true}" on:update />
+{/if}
--- a/packages/renderer/src/lib/container/ContainerColumnActionsContainer.svelte
+++ b/packages/renderer/src/lib/container/ContainerColumnActionsContainer.svelte
@@ -1,0 +1,13 @@
+<script lang="ts">
+import { ErrorMessage } from '@podman-desktop/ui-svelte';
+
+import ContainerActions from './ContainerActions.svelte';
+import type { ContainerInfoUI } from './ContainerInfoUI';
+
+export let object: ContainerInfoUI;
+</script>
+
+{#if object.actionError}
+  <ErrorMessage error="{object.actionError}" icon />
+{/if}
+<ContainerActions container="{object}" dropdownMenu="{true}" on:update />
--- a/packages/renderer/src/lib/container/ContainerColumnAge.svelte
+++ b/packages/renderer/src/lib/container/ContainerColumnAge.svelte
@@ -1,11 +1,14 @@
 <script lang="ts">
 import StateChange from '../ui/StateChange.svelte';
+import { ContainerUtils } from './container-utils';
 import type { ContainerGroupInfoUI, ContainerInfoUI } from './ContainerInfoUI';
 
 export let object: ContainerInfoUI | ContainerGroupInfoUI;
+
+const containerUtils = new ContainerUtils();
 </script>
 
-{#if 'state' in object}
+{#if containerUtils.isContainerInfoUI(object)}
   <div class="text-sm text-[var(--pd-table-body-text)]">
     <StateChange state="{object.state}">{object.uptime}</StateChange>
   </div>
--- a/packages/renderer/src/lib/container/ContainerColumnAge.svelte
+++ b/packages/renderer/src/lib/container/ContainerColumnAge.svelte
@@ -6,7 +6,7 @@ export let object: ContainerInfoUI | ContainerGroupInfoUI;
 </script>
 
 {#if 'state' in object}
-  <div class="text-sm text-gray-700">
+  <div class="text-sm text-[var(--pd-table-body-text)]">
     <StateChange state="{object.state}">{object.uptime}</StateChange>
   </div>
 {/if}
--- a/packages/renderer/src/lib/container/ContainerColumnAge.svelte
+++ b/packages/renderer/src/lib/container/ContainerColumnAge.svelte
@@ -1,0 +1,12 @@
+<script lang="ts">
+import StateChange from '../ui/StateChange.svelte';
+import type { ContainerGroupInfoUI, ContainerInfoUI } from './ContainerInfoUI';
+
+export let object: ContainerInfoUI | ContainerGroupInfoUI;
+</script>
+
+{#if 'state' in object}
+  <div class="text-sm text-gray-700">
+    <StateChange state="{object.state}">{object.uptime}</StateChange>
+  </div>
+{/if}
--- a/packages/renderer/src/lib/container/ContainerColumnImage.svelte
+++ b/packages/renderer/src/lib/container/ContainerColumnImage.svelte
@@ -5,7 +5,7 @@ export let object: ContainerInfoUI | ContainerGroupInfoUI;
 </script>
 
 {#if 'image' in object}
-  <div class="text-sm text-gray-700 overflow-hidden text-ellipsis" title="{object.image}">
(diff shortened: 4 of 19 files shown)


packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -463,204 +485,15 @@ function setStoppedFilter() {
...
-      </tbody>
-    </table>
+    <Table
+      kind="container"
+      bind:this="{table}"
+      bind:selectedItemsNumber="{selectedItemsNumber}"
+      data="{containerGroups}"
+      columns="{columns}"
+      row="{row}"
+      on:update="{() => (containerGroups = [...containerGroups])}">
```
cdrage, line 474: should we add `defaultSortColumn="Name"` ?

cdrage (dismissed):

packages/renderer/src/lib/container/ContainerList.svelte
deboer-tim: Done in latest update.

deboer-tim:
Rebased and updated, ready for full review!

deboer-tim:
Rebased on latest #7587 to avoid build failures.

cdrage:
Noticed some things feature wise before I dive deeper into the code:


**Light mode not working:**
Screenshot 2024-06-11 at 2 48 54 PM

**Compose isn't working, but pods are:**

https://github.com/containers/podman-desktop/assets/6422176/1ef95794-fd31-42b3-828d-61cfb09b0d84

To replicate:

```sh
git clone https://github.com/cdrage/composefiles
cd guestbook-go
podman compose up -d
```

And then try it in the interface, the containers do not open or close.

**Missing operators on Status, Environment, Image and Age:**
Screenshot 2024-06-11 at 2 50 57 PM

packages/renderer/src/lib/container/ContainerColumnAge.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
```
axel7083, line 1: Does this column could be replaced by `DurationColumn` from the ui package ?

packages/renderer/src/lib/container/ContainerColumnActions.svelte
```diff
@@ -0,0 +1,52 @@
...
+import ComposeActions from '../compose/ComposeActions.svelte';
+import PodActions from '../pod/PodActions.svelte';
+import ContainerActions from './ContainerActions.svelte';
+import type { ContainerGroupInfoUI, ContainerInfoUI } from './ContainerInfoUI';
+import { ContainerGroupInfoTypeUI } from './ContainerInfoUI';
+
+export let object: ContainerInfoUI | ContainerGroupInfoUI;
+</script>
+
+{#if object.type === ContainerGroupInfoTypeUI.POD && object.engineId && object.id && object.shortId && object.status && object.engineName && object.humanCreationDate && object.created}
```
axel7083: I can see that we often repeat the check `type === ContainerGroupInfoTypeUI.POD`, might be interesting to have an utility method

isPod(object: unknown): object is ContainerGroupInfoUI or something similar ? (not in this PR, just a general feeling)

packages/renderer/src/lib/container/ContainerColumnStatus.svelte
```diff
@@ -0,0 +1,16 @@
...
+export let object: ContainerInfoUI | ContainerGroupInfoUI;
+</script>
+
+{#if 'type' in object && object.type === ContainerGroupInfoTypeUI.POD}
+  <StatusIcon icon="{PodIcon}" status="{object.status}" />
+{:else if 'type' in object && object.type === ContainerGroupInfoTypeUI.COMPOSE}
+  <StatusIcon icon="{PodIcon}" status="{object.status}" />
+{:else if 'state' in object}
+  <StatusIcon icon="{object.icon}" status="{object.state}" />
+{/if}
```
axel7083:
```suggestion
{#if 'type' in object && typeof object.type === 'string'}
  <StatusIcon icon="{PodIcon}" status="{object.status}" />
{:else if 'state' in object && typeof object.type === 'string'}
  <StatusIcon icon="{object.icon}" status="{object.state}" />
{/if}
```

line 11 and 13 are the same, might be nice to simplify this

packages/renderer/src/lib/container/ContainerColumnName.svelte
```diff
@@ -0,0 +1,63 @@
...
+    router.goto(`/compose/details/${encodeURI(containerGroup.name)}/${encodeURI(containerGroup.engineId)}/logs`);
+  }
+}
+
+function openContainerDetails(container: ContainerInfoUI) {
+  router.goto(`/containers/${container.id}/`);
+}
+</script>
+
+{#if 'type' in object && (object.type === ContainerGroupInfoTypeUI.POD || object.type === ContainerGroupInfoTypeUI.COMPOSE)}
```
axel7083: `'type' in object && (object.type === ContainerGroupInfoTypeUI.POD || object.type === ContainerGroupInfoTypeUI.COMPOSE)` could be replaced by an utility method isContainerGroup() or something similar, would improve readiliby

packages/renderer/src/lib/container/ContainerColumnName.svelte
```diff
@@ -0,0 +1,63 @@
...
+  }
+  return result;
+}
+
+function openGroupDetails(containerGroup: ContainerGroupInfoUI) {
+  if (!containerGroup.engineId) {
+    return;
+  }
+  if (containerGroup.type === ContainerGroupInfoTypeUI.POD) {
+    router.goto(`/pods/podman/${encodeURI(containerGroup.name)}/${encodeURIComponent(containerGroup.engineId)}/logs`);
```
axel7083: I am wondering if we should not uses the `handleNavigation` from 

https://github.com/containers/podman-desktop/blob/54cdb5cd43cae0162f533221a1f4e0bd74dce66a/packages/renderer/src/navigation.ts#L32

to avoid hardcoding the urls in the code in multiple places

> Not a blocker

packages/renderer/src/lib/container/ContainerColumnName.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+function displayContainersCount(containerGroup: ContainerGroupInfoUI) {
+  let result = containerGroup.allContainersCount + ' container' + (containerGroup.allContainersCount > 1 ? 's' : '');
+  if (containerGroup.containers.length !== containerGroup.allContainersCount) {
+    result += ` (${containerGroup.allContainersCount - containerGroup.containers.length} filtered)`;
+  }
+  return result;
+}
+
+function openGroupDetails(containerGroup: ContainerGroupInfoUI) {
```
axel7083: Why is the linter not complaining about the return type not provided ?

packages/renderer/src/lib/container/ContainerColumnImage.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+import type { ContainerGroupInfoUI, ContainerInfoUI } from './ContainerInfoUI';
+
+export let object: ContainerInfoUI | ContainerGroupInfoUI;
+</script>
+
+{#if 'image' in object}
```
axel7083:
```suggestion
{#if 'image' in object && typeof object.image === 'string'}
```

packages/renderer/src/lib/container/ContainerColumnAge.svelte
```diff
@@ -0,0 +1,12 @@
...
+import StateChange from '../ui/StateChange.svelte';
+import type { ContainerGroupInfoUI, ContainerInfoUI } from './ContainerInfoUI';
+
+export let object: ContainerInfoUI | ContainerGroupInfoUI;
+</script>
+
+{#if 'state' in object}
```
axel7083:
```suggestion
{#if 'state' in object && typeof object.state === 'string'} 
```

axel7083: I am wondering if typescript should not complain here as state would be unknown type before checking its type

axel7083 (approved):
Amazing job, tested locally and loved it! 😄 🚀

deboer-tim:
Rebased to pick up Table fix and other changes in main. Will respond to the rest of the comments inline, but ready for another review.

> Noticed some things feature wise before I dive deeper into the code:
>
> Light mode not working:

I started this well before we were fixing columns :( and was going to do that next. Simple enough though, added in new commit.

> **Compose isn't working, but pods are:**

Rebased on fix from #7587.

> **Missing operators on Status, Environment, Image and Age:**

I assume you meant the sorting? This was intentional, I was trying to keep this PR focused on the main port and small(er) to get it in, then we can add new capability.

packages/renderer/src/lib/container/ContainerColumnAge.svelte
deboer-tim: It'd still need a wrapper b/c this column shows text like 'Starting' too, but otherwise yes. There is no date exposed in ContainerInfoUI though so I didn't want to include here (and I think pods have the same issue).

packages/renderer/src/lib/container/ContainerColumnActions.svelte
deboer-tim: Yeah, I think we do this (and compose) several places in the code so might be nice to do this.

packages/renderer/src/lib/container/ContainerColumnStatus.svelte
deboer-tim: I simplified in new commit, but just to `#if 'type' in object`. I assume you meant `typeof object.state` in the second case?  There are only two types here (ContainerInfoUI | ContainerGroupInfoUI) and type/state are each only on one, so I didn't really think checking the type of each property was necessary.

packages/renderer/src/lib/container/ContainerColumnName.svelte
deboer-tim: I forgot that I simplified and avoided this ever being called with a standalone container parent, so I was able to remove `object.type ===` in the new commit. But yes, a utility method might be useful in future.

packages/renderer/src/lib/container/ContainerColumnName.svelte
deboer-tim: Yes, we should do this. I opened #4966 a while ago, I think we should target a separate PR (or PRs) to clean this up in other places as well.

packages/renderer/src/lib/container/ContainerColumnName.svelte
deboer-tim: Good catch, I didn't notice while doing copy/paste. Maybe we don't check .svelte in renderer? I fixed all that I could find in the new commit.

packages/renderer/src/lib/container/ContainerColumnImage.svelte
deboer-tim: Ditto to previous comment, there are only two types and image only exists in one of them, so what's the added benefit of checking the property type?

packages/renderer/src/lib/container/ContainerColumnAge.svelte
deboer-tim: Only ContainerInfoUI has state and it's a string there though?

packages/renderer/src/lib/container/ContainerColumnImage.svelte
axel7083: type safety, never trust pattern

packages/renderer/src/lib/container/ContainerColumnAge.svelte
axel7083: what if I am sending you a { state: true }

packages/renderer/src/lib/container/ContainerColumnStatus.svelte
axel7083: Yeah having it everywhere would be redundant

packages/renderer/src/lib/container/ContainerColumnAge.svelte
axel7083: you could use rendermapping to convert it to a date object

benoitf (commented):
we're missing functions returning `is ContainerInfoUI` and `is ContainerGroupInfoUI`
as you're checking for properties like `type` or `state` in svelte objects while we shouldn't.

```
it should be const containerInfo = isContainerInfo(object)
if (containerInfo) {
  // here you can use containerInfo.state as there is the cast
}
```

and probably have renderer dedicated to containers and groups

so the final renderer should only have object being `ContainerInfoUI` or `ContainerGroupInfoUI` not the union type

packages/renderer/src/lib/container/ContainerColumnAge.svelte
deboer-tim: Not directly, since the column also uses the status. (I agree with direction, was saying more that it should be a future PR)

packages/renderer/src/lib/container/ContainerColumnStatus.svelte
deboer-tim: Added isContainerGroupInfoUI/isContainerInfoUI helper utils, and they use this logic.

packages/renderer/src/lib/container/ContainerColumnImage.svelte
deboer-tim: As per other comment, I've added isContainerGroupInfoUI/isContainerInfoUI helper utils which do this.

packages/renderer/src/lib/container/ContainerColumnAge.svelte
deboer-tim: As per other comment, I've added isContainerGroupInfoUI/isContainerInfoUI helper utils which do this.

packages/renderer/src/lib/container/ContainerColumnActions.svelte
```diff
@@ -0,0 +1,24 @@
...
+export let object: ContainerInfoUI | ContainerGroupInfoUI;
+
+const containerUtils = new ContainerUtils();
+</script>
+
+{#if containerUtils.isContainerGroupInfoUI(object)}
+  {#if object.type === ContainerGroupInfoTypeUI.POD}
+    <ContainerColumnActionsPod object="{object}" />
+  {/if}
+  {#if object.type === ContainerGroupInfoTypeUI.COMPOSE}
```
axel7083:
```suggestion
  {:else if object.type === ContainerGroupInfoTypeUI.COMPOSE}
```

An object type cannot be POD and COMPOSE at the same time

packages/renderer/src/lib/container/ContainerColumnActionsContainer.svelte
```diff
@@ -0,0 +1,13 @@
...
+import ContainerActions from './ContainerActions.svelte';
+import type { ContainerInfoUI } from './ContainerInfoUI';
+
+export let object: ContainerInfoUI;
+</script>
+
+{#if object.actionError}
+  <ErrorMessage error="{object.actionError}" icon />
+{/if}
+<ContainerActions container="{object}" dropdownMenu="{true}" on:update />
```
axel7083, line 13:
```suggestion
<ContainerActions container="{object}" dropdownMenu="{true}" />
```

packages/renderer/src/lib/container/ContainerColumnAge.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
+import StateChange from '../ui/StateChange.svelte';
+import { ContainerUtils } from './container-utils';
+import type { ContainerGroupInfoUI, ContainerInfoUI } from './ContainerInfoUI';
+
+export let object: ContainerInfoUI | ContainerGroupInfoUI;
+
+const containerUtils = new ContainerUtils();
```
axel7083, line 8: I would argue that we could have static method instead of having a dedicated class for this, I know @benoitf have a preference for it, as it _ease_ the testing _in some cases_ but here I feel like it adds complexity when reading it, as we might think ContainerUtils has internal logic, as it is a class

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -1,11 +1,10 @@
 <script lang="ts">
-import { faChevronDown, faChevronRight, faPlusCircle, faTrash } from '@fortawesome/free-solid-svg-icons';
-import { Button, Checkbox, ErrorMessage, FilteredEmptyScreen, Modal, NavPage } from '@podman-desktop/ui-svelte';
+import { faPlusCircle, faTrash } from '@fortawesome/free-solid-svg-icons';
+import { Button, FilteredEmptyScreen, Modal, NavPage, Table, TableColumn, TableRow } from '@podman-desktop/ui-svelte';
```
axel7083, line 3: <3 so many UI components

axel7083 (approved):

deboer-tim:
> we're missing functions returning `is ContainerInfoUI` and `is ContainerGroupInfoUI` as you're checking for properties like `type` or `state` in svelte objects while we shouldn't.
> 
> ```
> it should be const containerInfo = isContainerInfo(object)
> if (containerInfo) {
>   // here you can use containerInfo.state as there is the cast
> }
> ```

Ok, added these in new commit to avoid doing this in Svelte. There are still a couple cases where .type is checked for pod vs compose and we could do util for these too, but let me know what you think.

> and probably have renderer dedicated to containers and groups
> 
> so the final renderer should only have object being `ContainerInfoUI` or `ContainerGroupInfoUI` not the union type

I struggled with this because in the 'worst case' it means four files for every column: initial switcher column, plus components for pod, compose, and container. I did this for actions since there is different logic for each one, but not for age because there's just one if and div. Again, let me know if you think it should go further.

packages/renderer/src/lib/container/ContainerColumnActions.svelte
deboer-tim: Fixed.

packages/renderer/src/lib/container/ContainerColumnActionsContainer.svelte
deboer-tim: Fixed.

packages/renderer/src/lib/container/ContainerList.svelte
deboer-tim: Yup :(

packages/renderer/src/lib/container/ContainerColumnAge.svelte
deboer-tim: Yeah, I added here b/c well, everything else related is here. Could do static functions in the same file? Easy switch if someone prefers.

axel7083: yeah I would rather have static method where ContainerUtils is defined ?

axel7083: I just review the `containers-utils` file and it should probably be kept like you did for consistency, so forget my comment

benoitf (approved):
works for me now

PR #7430, opened 2024-06-03
chore: add tooltip to show load balancer IP for service

### LINKED ISSUE #7426: Show external IP when using load balancer for Kubernetes

Screenshot 2024-06-03 at 9 20 23 AM

I deployed a load balancer, but from the table list I cannot see what IP address it is.

On other GUI applications, this is usually in separate column.

Does not make sense to make another column as most services do not use load balancers (clusterIP, nodeport, etc) as load balancers are often "expensive" on cloud services.

We could add this in brackets to the Load Balancer badge instead with the IP address.

_No response_

### Additional context

_No response_

cdrage:
chore: add tooltip to show load balancer IP for service

### What does this PR do?

Adds a tooltip with the IP so you can easily see what the load balancer
IP is.

As a follow up PR, it'd be nice to copy this to clipboard.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/6422176/022c3478-5a0b-4b26-a608-a0c4e19760ef

Closes https://github.com/containers/podman-desktop/issues/7426

### How to test this PR?

Create a load balancer and hover over the badge.

### DIFF
--- a/packages/renderer/src/lib/service/service-utils.ts
+++ b/packages/renderer/src/lib/service/service-utils.ts
@@ -30,6 +30,7 @@ export class ServiceUtils {
       selected: false,
       type: service.spec?.type ?? '',
       clusterIP: service.spec?.clusterIP ?? '',
+      loadBalancerIPs: service.status?.loadBalancer?.ingress?.map(ingress => ingress.ip).join(', '),
       ports: (service.spec?.ports ?? []).map(port => port.port + '/' + port.protocol).join(', '),
     };
   }
--- a/packages/renderer/src/lib/service/ServiceColumnType.spec.ts
+++ b/packages/renderer/src/lib/service/ServiceColumnType.spec.ts
@@ -84,3 +84,20 @@ test('Expect column styling NodePort', async () => {
   expect(dot).toBeInTheDocument();
   expect(dot).toHaveClass('text-fuschia-600');
 });
+
+test('Expect tooltip to say loadBalancerIPs if it exists in service object', async () => {
+  service.type = 'LoadBalancer';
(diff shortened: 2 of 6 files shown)


axel7083 (approved):

packages/renderer/src/lib/service/service-utils.ts
```diff
@@ -30,6 +30,7 @@ export class ServiceUtils {
       selected: false,
       type: service.spec?.type ?? '',
       clusterIP: service.spec?.clusterIP ?? '',
+      loadBalancerIP: service.status?.loadBalancer?.ingress?.map(ingress => ingress.ip).join(', '),
```
jeffmaury: As it represents several values, worth renaming to loadBalancerIPs ?

cdrage:
@jeffmaury you are right, it should be loadBalancerIPs, updated!

feloy:
I would find reasonable to have this information in the name column, below the name, the same way we are displaying port information for containers, IDs for images, etc.

I'm not very comfortable having it as a tooltip on the type information, as it needs to be discovered by the user that it is in this place, and also because we cannot copy the text of the value.

cdrage:
> I would find reasonable to have this information in the name column, below the name, the same way we are displaying port information for containers, IDs for images, etc.
> 
> I'm not very comfortable having it as a tooltip on the type information, as it needs to be discovered by the user that it is in this place, and also because we cannot copy the text of the value.

You are right, I'm going to close this PR and re-open with the information below the name. It's a more reasonable location instead of as a tooltip as it's a load balancer and pretty important information.

I've also opened up issue #7474 as your comment made me think of how we should be putting more information below the name.

Issue #4970, opened 2023-11-23
Reduce overall number of telemetry events


deboer-tim:
### Is your enhancement related to a problem? Please describe

We are capturing a lot of telemetry. It is not much per user and has helped us improve, but the total amount is growing as the overall number of users grow.

### Describe the solution you'd like

It's time to be more efficient, and take a look at some of the most common events to see if we can merge or reduce the overall number of events, or implement #4710.

deboer-tim:
Reopening since I still plan to fix.

PR #7432, opened 2024-06-03
feat: move telemetry settings to json

deboer-tim:
### What does this PR do?

Move filtering of telemetry events to a json file so that it is easy to configure the filtering of events without changing code.

Prior to this, 'list' events were hardcoded to be filtered to once per day per instance of Podman Desktop. Now, this setting is done via a JSON config file, and there are three options that we could use for other events:
- disabled (true/false) - totally disable an event
- ratio (0.0-1.0) - only send x% of this event
- frequency - only supports 'dailyPerInstance' to start (supports list events), but could support things like 'daily' or 'weekly' in the future.

### Screenshot / video of UI

N/A

Part of #4970.

### How to test this PR?

Automated tests updated for new options.

### DIFF
--- a/packages/main/src/plugin/telemetry/telemetry-api.ts
+++ b/packages/main/src/plugin/telemetry/telemetry-api.ts
@@ -1,0 +1,28 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface TelemetryRule {
+  // start of telemetry event id
+  event: string;
+  // ratio to send events, e.g. 0.5
+  ratio?: number;
+  // disable this event entirely
+  disabled?: boolean;
+  // limit sending to 'dailyPerInstance'
+  frequency?: string;
+}
--- a/packages/main/src/plugin/telemetry/telemetry-api.ts
+++ b/packages/main/src/plugin/telemetry/telemetry-api.ts
@@ -16,13 +16,21 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+/**
+ * Frequency of collection:
+ *  dailyPerInstance - log the first time the event happens after startup.
+ *
+ * Enum to allow for future expansion, e.g. 'daily' or 'weekly'.
+ */
+export type Frequency = 'dailyPerInstance';
+
 export interface TelemetryRule {
-  // start of telemetry event id
+  // regex matching telemetry event
   event: string;
   // ratio to send events, e.g. 0.5
   ratio?: number;
   // disable this event entirely
   disabled?: boolean;
-  // limit sending to 'dailyPerInstance'
-  frequency?: string;
+  // control the frequency of sending events
+  frequency?: Frequency;
 }
--- a/packages/main/src/plugin/telemetry/telemetry.ts
+++ b/packages/main/src/plugin/telemetry/telemetry.ts
@@ -63,6 +63,7 @@ export class Telemetry {
   private static readonly SEGMENT_KEY = 'Mhl7GXADk5M1vG6r9FXztbCqWRQY8XPy';
 
   private cachedTelemetrySettings: TelemetryRule[] | undefined;
+  private regexp: Map<string, RegExp> = new Map();
 
   private identity: Identity;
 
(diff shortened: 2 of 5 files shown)


packages/main/src/plugin/telemetry/telemetry-api.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface TelemetryRule {
+  // start of telemetry event id
+  event: string;
+  // ratio to send events, e.g. 0.5
+  ratio?: number;
+  // disable this event entirely
+  disabled?: boolean;
+  // limit sending to 'dailyPerInstance'
+  frequency?: string;
```
jeffmaury: can't we have an enum here for better readability

packages/main/src/plugin/telemetry/telemetry-api.ts
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface TelemetryRule {
+  // start of telemetry event id
+  event: string;
```
jeffmaury, line 29: If having a regexp too expensive ?

packages/main/src/plugin/telemetry/telemetry.ts
```diff
@@ -275,25 +292,45 @@ export class Telemetry {
...
+      if (eventName.startsWith(entry.event)) {
+        if (entry.disabled) {
+          // telemetry is entirely disabled for this event
+          dropIt = true;
+        }
+        if (entry.ratio && entry.ratio < 1 && Math.random() > entry.ratio) {
+          // if a ratio is specified, we randomly drop
+          dropIt = true;
+        }
+        if (entry?.frequency === 'dailyPerInstance') {
```
jeffmaury: Why do we use ? here ?

packages/main/src/plugin/telemetry/telemetry.ts
```diff
@@ -275,25 +292,45 @@ export class Telemetry {
...
+              this.lastTimeEvents.set(eventName, now);
+            } else {
+              dropIt = true;
+            }
+          }
+        }
+      }
+    });
+
+    console.log('telem: ' + eventName + ' ' + dropIt + ' ' + telem.length);
```
jeffmaury: I think we should remove this

packages/main/src/plugin/telemetry/telemetry-api.ts
deboer-tim: Fixed in new commit.

packages/main/src/plugin/telemetry/telemetry-api.ts
deboer-tim: It shouldn't be, I was just trying to keep it simple. Switched to regex in new commit.

packages/main/src/plugin/telemetry/telemetry.ts
deboer-tim: Was from a previous revision, removed in new commit.

packages/main/src/plugin/telemetry/telemetry.ts
deboer-tim: Yes, debug statement I meant to remove. Gone now.

deboer-tim:
Rebased and responded to PR review comments.

deboer-tim:
Rebased since it had been almost a week. Could use a review please.

packages/main/src/plugin/telemetry/telemetry-api.ts
```diff
@@ -0,0 +1,36 @@
+/**********************************************************************
```
axel7083, line 1: Should this file be in the `packages/api` ?

axel7083 (approved):
LGTM Codewise

packages/main/src/plugin/telemetry/telemetry-api.ts
deboer-tim: recommendations.json is at the root, and like in that case the next PR here will be to try to download a fresher file at runtime so I kept it in the same place. I don't have a strong opinion if anyone thinks it should move though.

axel7083: I'm okey with it

packages/main/src/plugin/telemetry/telemetry-api.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Frequency of collection:
+ *  dailyPerInstance - log the first time the event happens after startup.
+ *
+ * Enum to allow for future expansion, e.g. 'daily' or 'weekly'.
+ */
+export type Frequency = 'dailyPerInstance';
+
+export interface TelemetryRule {
+  // regex matching telemetry event
+  event: string;
```
jeffmaury, line 29: If it is a regexp why it's not defined as a regexp ?

packages/main/src/plugin/telemetry/telemetry.ts
```diff
@@ -275,25 +292,45 @@ export class Telemetry {
...
-      }
+    const telem = this.getTelemetrySettings();
+    if (!telem || !eventName) {
       return true;
     }
-    return false;
+
+    let dropIt = false;
+    telem.forEach(entry => {
+      const regex = RegExp(entry.event);
```
jeffmaury: The regexp should be created at load time for performance reasons

packages/main/src/plugin/telemetry/telemetry-api.ts
deboer-tim: The interface matches what's in the json file to make for easy parsing, and needs to stay as a string since that's what you define there. I've created a map in latest commit to create and cache the RegExp at load time.

packages/main/src/plugin/telemetry/telemetry.ts
deboer-tim: Done in new commit.

jeffmaury (approved):
LGTM although conversion from string to RegExp could be done using reviver parameter in JSON.parse

deboer-tim:
> LGTM although conversion from string to RegExp could be done using reviver parameter in JSON.parse

Thanks, I'll pull this in a future PR.

PR #7436, opened 2024-06-03
chore: tooltip simplification

### LINKED ISSUE #7435: Restore default slot and simple tip for Tooltip

When I went to implement light mode on Tooltip (#6926) I saw that issue #7150 changed it so there are 2 slots (content and tip) and the bg and text colors are replicated on every single instance of Tooltip. We created the component to avoid duplicate code, so we should try to reduce this.

Do a few things to make it easier to consume Tooltip:
- Move 'content' back to be the default slot, since content must always be there and is naturally the visible child.
- Make the outer box consistent - i.e. you are welcome to customize the _content_ of tooltips, but the component creates the frame and bg color.
- In additional to the 'tip' slot, allow 'tip' to be a property, so in the simple cases you can still do:
```
<Tooltip right tip="some tip">
  <something/>
</Tooltip>
```
and when you want to totally customize the tip:
```
<Tooltip right>
  <something/>
  <svelte:fragment slot="tip">tip content goes here</svelte:fragment>
</Tooltip>
```
I know the 'tip can either be property or slot' is a bit different (could use different names?), but I think this handles the simple/normal case the easiest while still allowing for customization.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Changes:
- Move 'content' back to be the default slot, since content must always be there and is naturally the visible child.
- Make the outer box consistent - i.e. you are welcome to customize the _content_ of tooltips, but the component creates the frame and bg color.
- In additional to the 'tip' slot, allow 'tip' to be a property, so in the simple cases you can still do:
```
<Tooltip right tip="some tip">
  <something/>
</Tooltip>
```
and when you want to totally customize the tip content:
```
<Tooltip right>
  <something/>
  <svelte:fragment slot="tip">tip content goes here</svelte:fragment>
</Tooltip>
```

If someone accidentally supplies both, the simple tip is used. Having two if-statements is minor duplication, but it is more obvious and avoids extra spaces in the html.

I know this undoes a bunch of change and 'tip can either be property or slot' is a bit different (could use different names?), but I think this handles the simple/normal case as easy as before and has a single 'frame', while still allowing for enough customization.

### Screenshot / video of UI

N/A

Fixes #7435.

### How to test this PR?

Test various tooltips. PR tests updated.

### DIFF
--- a/packages/ui/src/lib/tooltip/Tooltip.svelte
+++ b/packages/ui/src/lib/tooltip/Tooltip.svelte
@@ -49,7 +49,8 @@
 }
 </style>
 
-<script>
+<script lang="ts">
+export let tip: string | undefined = undefined;
 export let top = false;
 export let topLeft = false;
 export let topRight = false;
@@ -62,7 +63,7 @@ export let left = false;
 
 <div class="relative inline-block">
   <span class="group tooltip-slot {$$props.class}">
-    <slot name="content" />
+    <slot />
   </span>
   <div
     class="whitespace-nowrap absolute tooltip opacity-0 inline-block transition-opacity duration-150 ease-in-out pointer-events-none z-[10]"
@@ -74,6 +75,15 @@ export let left = false;
     class:top-right="{topRight}"
     class:bottom-left="{bottomLeft}"
     class:bottom-right="{bottomRight}">
-    <slot name="tip" />
+    {#if tip}
+      <div class="inline-block py-2 px-4 rounded-md bg-charcoal-800 text-xs text-white" aria-label="tooltip">
+        {tip}
+      </div>
+    {/if}
+    {#if $$slots.tip && !tip}
+      <div class="inline-block rounded-md bg-charcoal-800 text-xs text-white" aria-label="tooltip">
+        <slot name="tip" />
+      </div>
+    {/if}
   </div>
 </div>
--- a/packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
@@ -34,23 +34,14 @@ function getConditionAttributes(condition: DeploymentCondition) {
 
 <div class="flex flex-row gap-1">
   {#each object.conditions as condition}
-    <Tooltip bottom>
-      <svelte:fragment slot="content">
-        <div class="flex flex-row bg-charcoal-500 items-center p-1 rounded-md text-xs text-gray-500">
-          <Fa
-            size="1x"
-            icon="{getConditionAttributes(condition).icon}"
-            class="{getConditionAttributes(condition).color} mr-1" />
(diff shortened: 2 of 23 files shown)


packages/ui/src/lib/tooltip/Tooltip.svelte
```diff
@@ -74,6 +75,15 @@ export let left = false;
...
     class:bottom-left="{bottomLeft}"
     class:bottom-right="{bottomRight}">
-    <slot name="tip" />
+    {#if tip}
+      <div class="inline-block py-2 px-4 rounded-md bg-charcoal-800 text-xs text-white" aria-label="tooltip">
+        {tip}
+      </div>
+    {/if}
+    {#if $$slots.tip && !tip}
+      <div class="inline-block py-2 px-4 rounded-md bg-charcoal-800 text-xs text-white" aria-label="tooltip">
```
lstocchi: Should we remove this wrapper from the slot? I expect that if one wants to use the slot, he's going to also add different classes and this could cause some interference

cdrage (approved):
IMHO this is much easier to interpret and write svelte-code for the tooltip, IMO this looks good and a great improvement.

Tests are passing, I do not see any visual change throughout the code which is good, LGTM.

Tested it a bit with "tipSlot" and it worked fine.

deboer-tim:
Rebased to fix conflicts due to #7412.

packages/ui/src/lib/tooltip/Tooltip.svelte
deboer-tim: I saw that's how it was before, but I thought it was odd for client to have to know/be required to use something like inline-block and aria-label, plus to allow completely custom popup with no styling from the library. Do you see cases where you think we should allow completely custom (e.g. square) popups?

When I rebased I removed the padding, and I think this is the sweet spot - the client doesn't need to know about inline/aria, is forced to paint within a rounded window (our tooltip style), and picks up our default styling for bg/text, but these can be overridden easily.

deboer-tim: @lstocchi please take another look

lstocchi: sorry @deboer-tim , i read your comment yesterday and then forgot to reply 🤦.
The only thing i could think of was a time where i wanted to add a text-wrap rule in ai-lab, but eventually we changed approach so we're not using it anymore and i don't remember if there was some other changes other than that.
So ok for me. We can change it in future if we face some issues 👍

lstocchi (approved):

packages/ui/src/lib/tooltip/Tooltip.svelte
deboer-tim: Thx. 👍🏼 FWIW I couldn't find any existing code/examples in AI Lab, but I did confirm you can do text wrap, change layout, or add icons using a div in the tip here.

PR #7446, opened 2024-06-04
chore(tests): Create rootless machine e2e verification

xbabalov:
### What does this PR do?
Implements  test verification of creation of a rootless machine in Podman Desktop.

https://github.com/containers/podman-desktop/issues/4665

### How to test this PR?
npm run test:e2e


tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,78 @@
...
+    super(page);
+    this.heading = this.page.getByRole('heading', { name: 'Create Podman Machine' });
+    this.podmanMachineConfiguration = this.page.getByRole('form', { name: 'Properties Information' });
+
+    this.machineNameBox = this.podmanMachineConfiguration.getByRole('textbox', { name: 'Name' });
+    this.imagePathBox = this.page.getByRole('textbox', { name: 'Image Path (Optional) ' });
+    this.browseImagesButton = this.page.getByRole('button', { name: 'button-Image Path (Optional)' });
+    this.rootPriviledgesCheckbox = this.podmanMachineConfiguration
+      .getByRole('checkbox', { name: 'Machine with root privileges' })
+      .locator('..');
```
odockal: You cannot toggle the checkbox itself, right? That is why you are going for a parent element?

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.userModeNetworkingCheckbox = this.page.getByRole('checkbox', {
+      name: 'User mode networking (traffic relayed by a user process). See [documentation](https://docs.podman.io/en/latest/markdown/podman-machine-init.1.html#user-mode-networking).',
```
odockal: anything in the parenthesis can be omitted since it is gonna work with just first part of the string. This would work if `exact` is not set to true.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,78 @@
...
+  readonly podmanMachineConfiguration: Locator;
+  readonly imagePathBox: Locator;
+  readonly browseImagesButton: Locator;
+  readonly rootPriviledgesCheckbox: Locator;
+  readonly userModeNetworkingCheckbox: Locator;
+  readonly startNowCheckbox: Locator;
+  readonly closeButton: Locator;
+  readonly createMachineButton: Locator;
+
+  constructor(page: Page) {
```
odockal, line 31: On general it is fine, just I would like you and Tibor to look at `podman-onboarding-page.ts` which has the same parts as this new page object model. With this new page object, the onboarding could make use of it.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,78 @@
...
+      .locator('..');
+    this.userModeNetworkingCheckbox = this.page.getByRole('checkbox', {
+      name: 'User mode networking (traffic relayed by a user process). See [documentation](https://docs.podman.io/en/latest/markdown/podman-machine-init.1.html#user-mode-networking).',
+    });
+    this.startNowCheckbox = this.page.getByRole('checkbox', { name: 'Start the machine now' });
+    this.closeButton = this.page.getByRole('button', { name: 'Close' });
+    this.createMachineButton = this.page.getByRole('button', { name: 'Create' });
+  }
+
+  async createMachine(machineName: string, isRootless: boolean): Promise<ResourcesPage> {
```
odockal: I would make `isRootless = false` default value, since it is the default.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,78 @@
...
+    await playExpect(this.podmanMachineConfiguration).toBeVisible();
+    await this.machineNameBox.fill(machineName);
+
+    if (!isRootless) {
+      await playExpect(this.rootPriviledgesCheckbox).toBeVisible();
+      const upperElement = this.rootPriviledgesCheckbox.locator('..');
+      const clickableCheckbox = upperElement.getByText('Enable');
+      await clickableCheckbox.click();
```
odockal: maybe also add `playExpect(checkbox).toBeEnabled` ? To check it was enabled?

tests/playwright/src/model/pages/create-machine-page.ts
odockal: and change the name to rootful, and make it true by default.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.startNowCheckbox = this.page.getByRole('checkbox', { name: 'Start the machine now' });
+    this.closeButton = this.page.getByRole('button', { name: 'Close' });
+    this.createMachineButton = this.page.getByRole('button', { name: 'Create' });
+  }
+
+  async createMachine(machineName: string, isRootless: boolean): Promise<ResourcesPage> {
+    await playExpect(this.podmanMachineConfiguration).toBeVisible();
+    await this.machineNameBox.fill(machineName);
+
+    if (!isRootless) {
```
odockal: I would also always make sure that the boolean switch is set/unset by the value you pass into function? so, if would be removed, and there would be a logic that will set rootful/rootless by the value passed. Unless it already is set to a correct value

tests/playwright/src/model/pages/create-machine-page.ts
odockal: @ScrewTSW ^^

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,78 @@
...
+  readonly createMachineButton: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.heading = this.page.getByRole('heading', { name: 'Create Podman Machine' });
+    this.podmanMachineConfiguration = this.page.getByRole('form', { name: 'Properties Information' });
+
+    this.machineNameBox = this.podmanMachineConfiguration.getByRole('textbox', { name: 'Name' });
+    this.imagePathBox = this.page.getByRole('textbox', { name: 'Image Path (Optional) ' });
+    this.browseImagesButton = this.page.getByRole('button', { name: 'button-Image Path (Optional)' });
```
odockal: Align the locators with what is in the above mentioned onboarding page.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,78 @@
...
+      const clickableCheckbox = upperElement.getByText('Enable');
+      await clickableCheckbox.click();
+    }
+
+    await this.createMachineButton.click();
+
+    const successfulCreationMessage = this.page.getByText('Successful operation');
+    const goBackToResourcesButton = this.page.getByRole('button', { name: 'Go back to resources' });
+
+    await playExpect(successfulCreationMessage).toBeVisible({ timeout: 100000 });
```
odockal: I would say that we need to improve the handling of what might happen once you click on Create button. 
we might got into what is here in case of success. What if there is an error? We should fail the test and throw that error that might appear in the form, right? Maybe also open log?

tests/playwright/src/model/pages/create-machine-page.ts
odockal: We could also incorporate other params like userModeNetworking and imagePath? Also start the Machine? So we have covered all three sliders.

As of cpu/memory/disk size. On WSL, user cannot really change these, but it might change with hyperv. Just to note that. no need to change it now.

tests/playwright/src/model/pages/create-machine-page.ts
odockal: Also, if there will be existing machine, this new one gonna get a dialogs regarding switching of the default connection. So it is going to be important to consider that aspect as well.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  page = await pdRunner.start();
+  pdRunner.setVideoAndTraceName('podman-rootless-machine-e2e');
+
+  const welcomePage = new WelcomePage(page);
+  await welcomePage.handleWelcomePage(true);
+  navBar = new NavigationBar(page);
+});
+
+afterAll(async () => {
+  await deletePodmanMachine(page, PODMAN_MACHINE_NAME);
```
odockal: I am truly curious how this test is going to work with a test env. in mind. Like we have usually podman machine already present. This means that the connection could be switched. But at the test end, we need to make sure switch it back again. 
This test would need some parameters to operate the workflow in a right way. and to fix the test among the rest.

odockal (changes requested):

jeffmaury (commented):
checkboxes are not handled in a uniform manner, any reason ?

tests/playwright/src/model/pages/create-machine-page.ts
xbabalov: Yes, I have to click on the label next to it instead.

tests/playwright/src/model/pages/create-machine-page.ts
xbabalov: I thought about extending the method to the other parameters as well but then I would have to add the functionality as well and not sure if it's within this issue for now as it's always extendable in future if needed.

xbabalov:
> checkboxes are not handled in a uniform manner, any reason ?

Yes, I wanted to access the parent element but forgot I did it in the definition already, I'll change it for consistency.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,116 @@
...
+  readonly startNowCheckbox: Locator;
+  readonly closeButton: Locator;
+  readonly createMachineButton: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.heading = this.page.getByRole('heading', { name: 'Create Podman Machine' });
+    this.podmanMachineConfiguration = this.page.getByRole('form', { name: 'Properties Information' });
+    this.podmanMachineName = this.podmanMachineConfiguration.getByRole('textbox', { name: 'Name' });
+    this.imagePathBox = this.page.getByRole('textbox', { name: 'Image Path (Optional) ' });
```
odockal: Please use the right locator to search for locator in, ie: `this.podmanMachineConfiguration`.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    await playExpect(successfulCreationMessage).toBeVisible({ timeout: 100000 });
+    await playExpect(goBackToResourcesButton).toBeVisible();
+    await goBackToResourcesButton.click();
+
+    return new ResourcesPage(this.page);
+  }
+
+  async isEnabled(checkbox: Locator): Promise<boolean> {
+    await playExpect(checkbox).toBeVisible();
```
odockal: Can we test only whether checkbox is enabled by using 'toBeEnabled' assert method? It should work since we found the element. having to find its parent locator is necessary only is we want to toggle the checkbox I think.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+beforeEach<RunnerTestContext>(async ctx => {
+  ctx.pdRunner = pdRunner;
+});
+
+describe.skipIf(os.platform() === 'linux')('Rootless Podman machine Verification', async () => {
+  test('Create a rootless machine', async () => {
+    await navBar.openSettings();
+    const resourcesPage = new ResourcesPage(page);
+
+    const createMachineButton = resourcesPage.podmanResources.getByRole('button', {
```
odockal: If there will be more that one Resource provider it will fail on ambiguity. 
update the logic so it finds the podman provider. Look under `model/pages/` and search for `resource-page.ts` and `resources-podman-connections-page.ts`. There will be new part besides `this.providerConnections = this.podmanResources.getByRole('region', { name: 'Provider Connections' });` : `Provider Setup`.

odockal (changes requested):

tests/playwright/src/model/pages/create-machine-page.ts
xbabalov: I've tried the default method, but it always results as enabled even when the checkbox is disabled, so I implemented a custom one. Tried it on the checkbox element itself.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
xbabalov: The constructor for that page takes a machine name as a parameter so I thought it would be better to use after the machine is created.

tests/playwright/src/model/pages/create-machine-page.ts
odockal: Ok!

tests/playwright/src/specs/podman-machine-rootless.spec.ts
odockal: nah, it should not have, since you can have more than one machine. Ok then!

odockal:
@xbabalov Overal PR is good. How about running it on windows workflow?

tests/playwright/src/utility/operations.ts
```diff
@@ -181,27 +183,57 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
...
     if ((await podmanResourceCard.resourceElementConnectionStatus.innerText()) === ResourceElementState.Running) {
       await podmanResourceCard.performConnectionAction(ResourceElementActions.Stop);
       await playExpect(podmanResourceCard.resourceElementConnectionStatus).toHaveText(ResourceElementState.Off, {
-        timeout: 30_000,
+        timeout: 300_000,
       });
     }
     await podmanResourceCard.performConnectionAction(ResourceElementActions.Delete);
     await playExpect(podmanResourceCard.resourceElement).toBeHidden({ timeout: 30_000 });
+    await page.waitForTimeout(30000);
```
ScrewTSW: would be nice to have consistent number scheme :D I personally prefer the underscore format, but we shouldn't have a mix of them.

ScrewTSW (approved):
LGTM, just one general comment

benoitf:
moving to draft, looks like not ready

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.closeButton = this.page.getByRole('button', { name: 'Close' });
+    this.createMachineButton = this.page.getByRole('button', { name: 'Create' });
+  }
+
+  async createMachine(
+    machineName: string,
+    { isRootful = true, enableUserNet = false, startNow = true, setAsDefault = true },
+  ): Promise<ResourcesPage> {
+    await playExpect(this.podmanMachineConfiguration).toBeVisible();
+    await this.podmanMachineName.fill(machineName);
```
cbr7:
```suggestion
    await this.podmanMachineName.clear();
    await this.podmanMachineName.fill(machineName);
```

Usually this field has a default value I think, the textbox should be cleared before entering the desired name.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    await this.ensureCheckboxState(isRootful, this.rootPriviledgesCheckbox);
+    await this.ensureCheckboxState(enableUserNet, this.userModeNetworkingCheckbox);
+    await this.ensureCheckboxState(startNow, this.startNowCheckbox);
+
+    await this.createMachineButton.click();
+    await this.page.waitForTimeout(60_000);
```
cbr7: These kind of hard waits should be replaced with something that the framework can await on, it's not clear that creating the machine will take 60 seconds, and if it's shorter then it will wait more than needed and if it's longer it won't wait long enough.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+    { isRootful = true, enableUserNet = false, startNow = true, setAsDefault = true },
+  ): Promise<ResourcesPage> {
+    await playExpect(this.podmanMachineConfiguration).toBeVisible();
+    await this.podmanMachineName.fill(machineName);
+
+    await this.ensureCheckboxState(isRootful, this.rootPriviledgesCheckbox);
+    await this.ensureCheckboxState(enableUserNet, this.userModeNetworkingCheckbox);
+    await this.ensureCheckboxState(startNow, this.startNowCheckbox);
+
+    await this.createMachineButton.click();
```
cbr7:
```suggestion
    await playExpect(this.createMachineButton).toBeEnabled();
    await this.createMachineButton.click();
```

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+    await this.page.waitForTimeout(60_000);
+
+    const successfulCreationMessage = this.page.getByText('Successful operation');
+    const goBackToResourcesButton = this.page.getByRole('button', { name: 'Go back to resources' });
+
+    await delay(5000);
```
cbr7: Same as above, hard waits should be avoided if at all possible.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (desiredState !== (await this.isEnabled(checkbox))) {
+      await this.switchCheckbox(checkbox);
+    }
+  }
+
+  async isEnabled(checkbox: Locator): Promise<boolean> {
+    await playExpect(checkbox).toBeVisible();
+    const upperElement = checkbox.locator('..').locator('..');
+    const clickableCheckbox = upperElement.getByText('Enabled');
+    return await clickableCheckbox.isVisible();
```
cbr7: I'm not sure I understand what is happening here, why are we looking for some text that says Enabled when the checkbox exposes the method isChecked() which should return current state. Does isChecked() not work for some reason?

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  async switchCheckbox(checkbox: Locator): Promise<void> {
+    await playExpect(checkbox).toBeVisible();
+    const upperElement = checkbox.locator('..').locator('..');
+
+    const wasEnabled = await this.isEnabled(checkbox);
```
cbr7: Same as above.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+    let checkText;
+    if (wasEnabled) {
+      checkText = 'Enabled';
+    } else {
+      checkText = 'Disabled';
+    }
+
+    const clickableCheckbox = upperElement.getByText(checkText);
+    await clickableCheckbox.click();
```
cbr7: After performing the switching the new state should be asserted upon to ensure that the value of the checkbox actually changed.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+    await this.ensureCheckboxState(startNow, this.startNowCheckbox);
+
+    await this.createMachineButton.click();
+    await this.page.waitForTimeout(60_000);
+
+    const successfulCreationMessage = this.page.getByText('Successful operation');
+    const goBackToResourcesButton = this.page.getByRole('button', { name: 'Go back to resources' });
+
+    await delay(5000);
+    await this.handleConnectionDialog(machineName, setAsDefault);
```
cbr7: What if there is no previous podman machine in the system, as such the new machine is automatically set as default, wouldn't this call throw an error since the popup would not be displayed?

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
...
+    const dialogMessage = connectionDialog.getByText(
+      new RegExp(
+        `Podman Machine '${machineName}' is running but not the default machine .+ Do you want to set it as default?`,
+      ),
+    );
+    if ((await connectionDialog.isVisible()) && (await dialogMessage.isVisible())) {
+      let handleButtonName = 'Yes';
+      if (!setAsDefault) {
+        handleButtonName = 'Ignore';
+      }
```
cbr7:
```suggestion
    const dialogMessage = connectionDialog.getByLabel('Dialog Message');
    await playExpect(dialogMessage).toContainText('Do you want to set it as default');
   
    let handleButtonName = 'Yes';
    if (!setAsDefault) {
        handleButtonName = 'Ignore';
     }
```

tests/playwright/src/specs/podman-machine-rootless.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  await welcomePage.handleWelcomePage(true);
+});
+
+test.afterAll(async ({ pdRunner }) => {
+  await pdRunner.close();
+});
+
+test.describe('Rootless Podman machine Verification', () => {
+  test.skip(os.platform() === 'linux', 'runs only on windows and mac');
```
cbr7: This test.skip() should be outside the describe, because if it's outside it will be validated before the beforeAll, if it's inside the describe the beforeAll will still execute.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    test.setTimeout(200000);
+
+    await navigationBar.openSettings();
+    const podmanResources = new ResourceConnectionCardPage(page, 'podman');
+
+    await podmanResources.createButton.click();
+
+    const createMachinePage = new CreateMachinePage(page);
+    await createMachinePage.createMachine(PODMAN_MACHINE_NAME, { isRootful: false, setAsDefault: false });
+    await delay(5000);
```
cbr7: Same as above, hard waits should be avoided, please await on some element to ensure stability.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    await createMachinePage.handleConnectionDialog(PODMAN_MACHINE_NAME, false);
```
cbr7: isn't the `createMachine()` method supposed to handle the popup dialog on line 83 of the `create-machine-page.ts`? Why is it also handled in the test?

tests/playwright/src/utility/operations.ts
```diff
@@ -57,7 +57,7 @@ export async function deleteContainer(page: Page, name: string): Promise<void> {
     // wait for container to disappear
     try {
       console.log('Waiting for container to get deleted ...');
-      await playExpect.poll(async () => await containers.getContainerRowByName(name), { timeout: 30000 }).toBeFalsy();
+      await playExpect.poll(async () => await containers.getContainerRowByName(name), { timeout: 10_000 }).toBeFalsy();
```
cbr7: Why lower the timeout here by 20 seconds?

tests/playwright/src/utility/operations.ts
```diff
@@ -140,7 +140,7 @@ export async function deletePod(page: Page, name: string): Promise<void> {
         async () => {
           return !!(await pods.getPodRowByName(name));
         },
-        { timeout: 20000 },
+        { timeout: 20_000 },
```
cbr7: are all these number format changes needed?

tests/playwright/src/utility/operations.ts
```diff
@@ -205,16 +205,30 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
     if ((await podmanResourceCard.resourceElementConnectionStatus.innerText()) === ResourceElementState.Running) {
       await podmanResourceCard.performConnectionAction(ResourceElementActions.Stop);
       await playExpect(podmanResourceCard.resourceElementConnectionStatus).toHaveText(ResourceElementState.Off, {
-        timeout: 30_000,
+        timeout: 100_000,
       });
     }
     await podmanResourceCard.performConnectionAction(ResourceElementActions.Delete);
     await playExpect(podmanResourceCard.resourceElement).toBeHidden({ timeout: 30_000 });
+    await delay(10000);
```
cbr7: No hard waits please.

tests/playwright/src/utility/operations.ts
```diff
@@ -205,16 +205,30 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
...
+    await handleResetDefaultConnectionDialog(page);
   } else {
     console.log(`Podman machine [${machineVisibleName}] not present, skipping deletion.`);
   }
 }
 
+export async function handleResetDefaultConnectionDialog(page: Page): Promise<void> {
+  const connectionDialog = page.getByRole('dialog', { name: 'Podman' });
+  if (await connectionDialog.isVisible()) {
```
cbr7: What happens if isVisible() returns false because it's checked before the popup is displayed? Shouldn't the visibility be awaited here?

cbr7 (changes requested):
Left some comments above.

tests/playwright/src/model/pages/create-machine-page.ts
odockal: I think that filling it in works as expected, otherwise test would fail. Regardless, clearing it before is good practice.

tests/playwright/src/model/pages/create-machine-page.ts
xbabalov: Yes, it doesn't work, it always returns as checked no matter what the actual status is. I have to check by the label next to it.

tests/playwright/src/utility/operations.ts
xbabalov: Not really, I just wanted all numbers to be have consistent format in the same file.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
xbabalov: When testing locally sometimes the message didn't pop up until back in the resources so I call the method twice in case.

tests/playwright/src/model/pages/create-machine-page.ts
cbr7: hey Tamara, I was able to get the rootfyk checkbox to be validated in a recent test I just merged to main, the short of it is that the checkbox can be validated for state, as such checking for Enabled/Disabled text is not required, however to trigger the changing of the state we need to click the parent element, please see https://github.com/containers/podman-desktop/blob/main/tests/playwright/src/specs/podman-machine-tests.spec.ts#L96 and following lines for a model.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,137 @@
+/**********************************************************************
```
cbr7, line 1: Recently as part of creating a new podman machine test I've noticed that the `podman-onboarding-page.ts` contains exactly the same content as this new page would, the issue is that the onboarding page is simply misnamed, initially it was created a POM class to be used in the onboarding workflow for podman machine tests, but what it actually models is that podman resource creation machine, just like this one, it would make sense to rename the old page and refactor it instead of creating a new page with the same content but with a new name.

In my test I actually repurposed that page and used it for creating a podman machine.

odockal (changes requested):

tests/playwright/src/model/pages/create-machine-page.ts
xbabalov: They are not the same page, it just contains the same properties information form (name, checkboxes). I could add the properties form into workbench and then have it repurposed in both pages, but I'd like to keep the pages separate.

cbr7: The pages are the same, the issue is that the page is reused in the app in different contexts, but the creation page is identical, we should not have multiple POM pages that address the same app page, yes there are some differences due to how the onboarding workflow works, but one can think of those differences as extra pages also. Maintaining two POMs for the same page is not scalable and defeats the purpose of the POM model.

benoitf:
hoping we'll be able to merge this PR 🤞 it's there since 3 months+

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,75 @@
...
+      new RegExp(
+        `Podman Machine '${machineName}' is running but not the default machine .+ Do you want to set it as default?`,
+      ),
+    );
+
+    const handleButtonName = setAsDefault ? 'Yes' : 'Ignore';
+    const handleButton = connectionDialog.getByRole('button', { name: handleButtonName });
+    await handleButton.click();
+  }
+}
```
cbr7, line 75: There's already a handleDialog method in the `operations.ts` file, do we need a custom one here and not use the general one from there?

cbr7:
Looks good overall.

cbr7 (approved):

tests/playwright/src/model/pages/create-machine-page.ts
odockal: I think here we got three options?

cbr7: I'm not sure that's relevant, from what I remember you can send the string by which it will select the button you want, so it doesn't matter if there are two, three or ten buttons on the dialog.

odockal (approved):

tests/playwright/src/model/pages/create-machine-page.ts
odockal: @cbr7 true, let's wait for @xbabalov :)

odockal:
LGTM. Just please rebase, resolve conflicts and maybe dialog handling, but I guess we are good to go.

Issue #6494, opened 2024-03-21
Trigger whole e2e-test-main workflow on PR check in a non-blocking way


odockal:
### Is your enhancement related to a problem? Please describe

Sometimes e2e-tests-main workflow runs are failing due to changes or issues, but it is hard to keep attention on them since PR is merged and smoke was passing, so no one notices - unless specifically being watching these which is hard. 

### Describe the solution you'd like

Make sure that we run full tests suite on PR check as well as the smoke e2e run but also make the run non-blocking. 

### Describe alternatives you've considered

We could trigger existing e2e-tests-main workflow, passing parameters using `workflow_call` event, but we would need to consider the security aspect of triggering remote branches from non approved sources.

PR #7458, opened 2024-06-04
chore(ci): include parallel all e2e tests on PR check

odockal:
### What does this PR do?
Runs all e2e tests on PR check. This workflow should be non-blocking.
#6494 

### How to test this PR?
On your own fork only.

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -281,6 +281,7 @@ jobs:
   e2e-tests:
     name: e2e tests
     runs-on: ubuntu-22.04
+    needs: smoke-e2e-tests
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -277,3 +277,52 @@ jobs:
         with:
           name: smoke-e2e-tests
           path: ./tests/output/
+
+  e2e-tests:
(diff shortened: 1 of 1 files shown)


odockal:
I am thinking about adding `needs` keyword into `e2e-tests` to depend on smoke job since it would run only if smoke has passed.

cbr7:
I'm not seeing the step being run in the pr-check on here.

cbr7 (approved):

odockal:
@benoitf @cbr7 after discussion I am reverting the need for smoke-e2e for e2e tests. This way it will be faster as it can run in parallel and it would also show us possible flakiness of the tests if one passes and other not.

benoitf:
how long does it take ?
also what's the value in running the smoke tests if we run all the tests ?

benoitf:
I see 5mn for smoke and 7mn for all the tests so should we drop smoke tests ?

odockal:
@benoitf Initial idea was to know if there is something broken besides the core functionality. 

In this particular case, e2e tests differ in a way that they also test extensions installation (dev sandbox and openshift local) and some basic registry handling (where we do not have secrets available so no real adding of the registries). 

Now, what is core functionality that needs to be always passing was denoted as smoke tests. The rest is like all-tests. If we want to distinguish or not, is good question.

jeffmaury:
@odockal this PR is almost 2 monthes old. What is the status ?

benoitf:
closing as no activity

PR #7463, opened 2024-06-04
Fix files watched in a non existing directory 

### LINKED ISSUE #6406: Kube context is not loaded in podman desktop until restart
### Bug description

I have no `~/.kube/config` file, thus not kube context available. I install DevSandbox extension in Podman Desktop and create a dev-sandbox resource (aka logging in and getting oc login command). At this moment I have my `~/.kube.config` filled with information about dev sandbox cluster and technically can connect to it. Podman Desktop has not information about the context until I restart it, then it works.

### Operating system

Windows 10, 11

### Installation Method

### Version

1.8.0

### Steps to reproduce

0. Make sure to remove any kubeconfig file beforehand. `rm -rf ~/.kube`
1. Start up podman desktop
2. Crate and start podman machine
3. Download Kind/Minikube or install dev sandbox
4. Create a kind cluster (or any other that touches ~/.kube/config and add new kube context)
5. Assert: New kube context is available in the Podman Desktop
Actual result: No kube context added. Only after restart or reenblement of the kube-context extension

### Relevant log output

_No response_

### Additional context

Kind and minikube could have the same problem, I believe.

feloy:

### What does this PR do?

When a file is watched and its parent directory does not exist, then when the directory is created later, the file creation cannot be detected (except on mac).

Example:
``` 
rm -rf /home/user/.kube
watch /home/user/.kube/config
mkdir /home/user/.kube
echo '...' > /home/user/.kube/config // the creation of the file is not detected
```

This PR changes the Watch behaviour, to watch only a file if the parent directory exists. If the parent directory does not exist, a watch in the parent directory is triggered instead, and the initial watch is triggered when the parent is created (recursively).

Fixes #6406 

### How to test this PR?

- Completely terminate Podman Desktop
- `cd $HOME`
- `mv .kube .kube-save`
- Start Podman Desktop
- Kubernetes icons should not be displayed
- `mkdir .kube`
- `cp .kube-save/config .kube/`
- Kubernetes icons should appear

### DIFF
--- a/packages/main/src/plugin/filesystem-monitoring.spec.ts
+++ b/packages/main/src/plugin/filesystem-monitoring.spec.ts
@@ -112,3 +112,38 @@ test('should send event onDidCreate when a directory is created into a watched d
   expect(changeListener).not.toHaveBeenCalled();
   expect(unlinkListener).not.toHaveBeenCalled();
 });
+
+test('should send event onDidCreate when a file is created inside a non-existent directory', async () => {
+  const watchedFile = path.join(rootdir, 'dir/file.txt');
+  watcher = new FileSystemWatcherImpl(watchedFile);
+
+  const readyListener = vi.fn();
+  watcher.onReady(readyListener);
+
+  const createListener = vi.fn();
+  watcher.onDidCreate(createListener);
+  const changeListener = vi.fn();
+  watcher.onDidChange(changeListener);
+  const unlinkListener = vi.fn();
+  watcher.onDidDelete(unlinkListener);
+
+  expect(createListener).not.toHaveBeenCalled();
+  expect(changeListener).not.toHaveBeenCalled();
+  expect(unlinkListener).not.toHaveBeenCalled();
+
+  await vi.waitFor(async () => {
+    expect(readyListener).toHaveBeenCalled();
+  });
+
+  const dir = path.dirname(watchedFile);
+  await promises.mkdir(dir);
+
+  const h = await promises.open(watchedFile, 'a');
+  await h.close();
+
+  await vi.waitFor(async () => {
+    expect(createListener).toHaveBeenCalledWith(Uri.file(watchedFile));
+  });
+  expect(changeListener).not.toHaveBeenCalled();
+  expect(unlinkListener).not.toHaveBeenCalled();
+});
--- a/packages/main/src/plugin/filesystem-monitoring.ts
+++ b/packages/main/src/plugin/filesystem-monitoring.ts
@@ -36,6 +36,10 @@ export class FileSystemWatcherImpl implements containerDesktopAPI.FileSystemWatc
     } else if (parent !== path) {
       // we stop the recursion at /
       const parentWatcher = new FileSystemWatcherImpl(parent);
+      const dispoReady = parentWatcher.onReady(() => {
+        this._onReady.fire();
+        dispoReady.dispose();
+      });
       const disposable = parentWatcher.onDidCreate((uri: containerDesktopAPI.Uri) => {
         if (uri.path === parent) {
           this.doWatch(path);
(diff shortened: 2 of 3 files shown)


packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -188,10 +188,10 @@ export class KubernetesClient {
     const kubernetesConfiguration = this.configurationRegistry.getConfiguration('kubernetes');
     const userKubeconfigPath = kubernetesConfiguration.get<string>('Kubeconfig');
     if (userKubeconfigPath) {
+      this.kubeconfigPath = userKubeconfigPath;
```
feloy, line 192: Moving this here, or the `kubeconfigPath` value (used in `refresh`). could be wrong when a user-defined value is used and the files does not exist initially.

packages/main/src/plugin/filesystem-monitoring.ts
```diff
@@ -34,6 +53,10 @@ export class FileSystemWatcherImpl implements containerDesktopAPI.FileSystemWatc
       const uri: containerDesktopAPI.Uri = Uri.file(addedPath);
       this._onDidCreate.fire(uri);
     });
+    watcher.on('addDir', (addedPath: string) => {
+      const uri: containerDesktopAPI.Uri = Uri.file(addedPath);
+      this._onDidCreate.fire(uri);
+    });
```
feloy: I don't know why the `addDir` event was not listened. If it is on purpose, I will need to look at another way.

packages/main/src/plugin/filesystem-monitoring.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+import { existsSync } from 'node:fs';
+
+import { FSWatcher, watch } from 'chokidar';
+import { beforeEach, expect, test, vi } from 'vitest';
+
+import { FileSystemWatcherImpl } from './filesystem-monitoring.js';
+import { Uri } from './types/uri.js';
+
+beforeEach(() => {
+  vi.mock('chokidar');
```
axel7083: maybe we should not mock chokidar to make some real test on it ?

feloy: For which reason? We generally try to avoid testing dependencies on unit tests

axel7083: To ensure it is working, or have a real regression test

benoitf: I think I'm 50-50

I like to have a unit test with the real implementation, so when we have dependabot, we check if it'll break or not the update of the dependency with the unit tests

But we still to mock the dependency for most of the tests

feloy: I would be ok to create integration tests (using real chokidar/fs) in addition to the unit tests I have created in this PR, to cover the library/fs related behaviour.

Because `vi.mock('...')` is hoisted to the top of the file, it is not possible to mix tests mocking and not mocking a given library in a same spec file. Is it ok to create some `filesystem-monitoring.integration.spec.ts` in the same directory, or do want to place these tests somewhere else?

axel7083: I am okey having another file for some kind integration, it would improve how we deal with external library if we are able to make this pattern for other package

jeffmaury: I would add that if we can write a unit test without mock chokidar in a simple way (let's say simpler than with mocking instructions) we should choose that

feloy:
I'll work on another PR first, as I have encountered a bug when running unit tests with native chokidar (#7501)

packages/main/src/plugin/filesystem-monitoring.spec.ts
feloy: the tests are now using the original chokidar

jeffmaury (approved):
LGTM

cdrage (approved):
Works and LGTM!

https://github.com/containers/podman-desktop/assets/6422176/79e87cf0-a135-4015-abeb-8812736c6e50

PR #7466, opened 2024-06-05
fix: Modal component is shown too high and gets out of viewport

dgolovin:
### What does this PR do?

Removes absolute position for Modal component because it is already has w-full and h-full

### Screenshot / video of UI

That will change this behavior 

image

to a normal one

image

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/modal/Modal.svelte
+++ b/packages/ui/src/lib/modal/Modal.svelte
@@ -32,7 +32,7 @@ if (previously_focused) {
 
 <svelte:window on:keydown="{handle_keydown}" />
 
-<div class:items-center="{!top}" class="fixed top-0 left-0 right-0 bottom-0 w-full h-full flex justify-center z-50">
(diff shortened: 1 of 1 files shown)


axel7083:
cc @deboer-tim

lstocchi (commented):
It introduces some issue in some cases.

Like when you click on the store name (troubleshooting -> stores -> click on a store name)
image

or if you try to push an image (images -> push image)
image

deboer-tim:
Took a quick look. If anything I think we need to keep inset-0 on both the outer div and button (replacing top/left/w-*). I think we need to go back to compare against previous release to see how adopting Modal affected this - there's clearly a bigger issue here where we'll need to use z-ordering instead of overflow-hidden, but might be easier in the meantime to use a workaround like class:mt-[40px]="{top}"?

dgolovin:
Closing this PR, because it is just a workaround for me to test other issue and not a real fix.

PR #7472, opened 2024-06-05
chore: move FormPage to its own component

### LINKED ISSUE #6920: Implement FormPage UI component
depends on
- [x] https://github.com/containers/podman-desktop/issues/6901
- [x] https://github.com/containers/podman-desktop/issues/6921

benoitf:
### What does this PR do?
remove tinro dependency to include it in the UI library
keep a tinro FormPage in renderer part that is just delegating the calls

fixes https://github.com/containers/podman-desktop/issues/6920

### How to test this PR?


lstocchi (approved):
Tested and looks good!!

axel7083 (approved):
LGTM

benoitf:
the problem I have is that my PR is failing on CI :-( but on my side it's working so I'm still investigating

packages/ui/src/lib/layouts/FormPage.svelte
```diff
@@ -0,0 +1,82 @@
+<script lang="ts">
+import { CloseButton, LinearProgress, Link } from '@podman-desktop/ui-svelte';
```
axel7083: Maybe we should not import itself as a package ?

axel7083:
> the problem I have is that my PR is failing on CI :-( but on my side it's working so I'm still investigating

I just tried locally, and tests are passing too... but I think it may come from the import itself package

packages/ui/src/lib/layouts/FormPage.svelte
benoitf: what is strange is why it's only reproducing on the CI, I would hope that developers have the same failure

axel7083: maybe try to delete the export/dist folder, and you _should_ be able to reproduce ?

deboer-tim: Or delete node_modules? i.e. you've already done a yarn install and have a local cache of what it will build here.

PR #7478, opened 2024-06-05
chore: fix table sorting

### LINKED ISSUE #7475: Table name sorting not working on load the page

https://github.com/containers/podman-desktop/assets/6422176/a1a42fa4-3b6b-4382-be54-92d267540f0a

When loading kubernetes pages such as deployments / services / etc it is not sorted on load.

Sort correctly on viewing

_No response_

### Additional context

_No response_

cdrage:
chore: fix table sorting

### What does this PR do?

When sorting, we are sorting the columns on the onMount, however, we are
not waiting for data (tick()) as well as not sorting it afterwards
(sortImpl()).

This PR fixes it by sorting the table on launch.

Fixes https://github.com/containers/podman-desktop/issues/7475

### How to test this PR?

View the pages and see they are ordered now.

### DIFF
--- a/packages/ui/src/lib/table/Table.spec.ts
+++ b/packages/ui/src/lib/table/Table.spec.ts
@@ -265,3 +265,24 @@ test('Expect update callback', async () => {
 
   expect(callback).toHaveBeenCalled();
 });
+
+test('Expect table to be sorted by Id on load', async () => {
+  render(TestTable, {});
+
+  // Wait for the table to load and fetch all rows
+  const rows = await screen.findAllByRole('row');
+  expect(rows).toBeDefined();
+  expect(rows.length).toBe(4);
+
+  const headers = await screen.findAllByRole('columnheader');
+  expect(headers).toBeDefined();
+  expect(headers.length).toBe(6);
+  expect(headers[2].textContent).toContain('Id');
+
+  // Check that Id column is sorted in ascending order
+  // since Id is by 1,2,3 in TestTable, just using a for loop and converting
+  // to string is enough to check the order
+  for (let i = 1; i < 4; i++) {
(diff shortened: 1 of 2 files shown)


packages/ui/src/lib/table/Table.spec.ts
```diff
@@ -265,3 +270,29 @@ test('Expect update callback', async () => {
...
+  expect(rows).toBeDefined();
+  expect(rows.length).toBe(4);
+
+  const headers = await screen.findAllByRole('columnheader');
+  expect(headers).toBeDefined();
+  expect(headers.length).toBe(6);
+  expect(headers[2].textContent).toContain('Id');
+
+  // onMount will wait for tick() to complete, so we need to wait for the sort
+  while (!headers[2].innerHTML.includes('fa-sort-up')) {
```
jeffmaury: I remember there is a better way to wait for an event but can't find the name but @dgolovin may help

packages/ui/src/lib/table/Table.svelte
```diff
@@ -97,10 +97,12 @@ function sortImpl(): void {
 }
 
 onMount(async () => {
+  await tick(); // Ensure all initializations are complete.
```
deboer-tim: Maybe tick() should be moved immediately before sortImpl()? i.e. if no sorting then onMount() doesn't wait, if sorted then we mark the sort column immediately (no need for lines 106-110 in test) then tick and do the actual sort. Lower part of test would need to be rewritten.

cdrage:
@jeffmaury @deboer-tim thanks for the reviews!

* Updated the tick so it is right before the sorting. 
* Removed the unneeded while loop in the tests.

Ready for another review!

deboer-tim (approved):

PR #7497, opened 2024-06-06
chore(github): adding storybook build workflow

axel7083:
### What does this PR do?

To avoid breaking storybook build when updating dependencies (ref https://github.com/containers/podman-desktop/pull/7486#pullrequestreview-2101408827), adding a workflow for building storybook.

### How to test this PR?

- Pipeline should be ✅

### DIFF
--- a/.github/workflows/pr-check-storybook.yaml
+++ b/.github/workflows/pr-check-storybook.yaml
@@ -1,0 +1,53 @@
+#
+# Copyright (C) 2024 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Build storybook on pull request
+
+on:
+  pull_request:
+    paths:
+      - 'package.json'
+      - 'yarn.lock'
+      - 'storybook/**'
+
+jobs:
+  storybook-build:
+    name: build storybook
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-check-storybook.yaml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Execute yarn
+        run: yarn --frozen-lockfile
+
+      - name: Run website
```
benoitf: looks like a copy/paste issue

about the workflow I don't know how much it'll take, to know if we could not add it as part of the linters/checkers tests

.github/workflows/pr-check-storybook.yaml
```diff
@@ -0,0 +1,53 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+name: Build storybook on pull request
+
+on:
+  pull_request:
+    paths:
+      - 'package.json'
+      - 'yarn.lock'
+      - 'storybook/**'
```
benoitf, line 25: I would say that you want to check storybook if you modify the components as well
(let say I remove Button component or move it elsewhere I would hope that PR check fails)

axel7083: Oh yeah smart

axel7083:
@benoitf I changed the `storybook:build` to build the ui package before building storybook, as it depends on it. I also added as suggested the ui package to the trigger path.

.github/workflows/pr-check-storybook.yaml
benoitf: maybe just add an entry https://github.com/containers/podman-desktop/blob/main/.github/workflows/pr-check.yaml#L176-L221 if it's taking only few seconds

benoitf (approved):
~~we can add the~~ (wrong text :-)

.github/workflows/pr-check-storybook.yaml
axel7083: We need to build the ui package first, the linter would not have the packages build I guess ?

.github/workflows/pr-check-storybook.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - 'packages/ui/src/**'
+
+jobs:
+  storybook-build:
+    name: build storybook
+    runs-on: ubuntu-22.04
```
benoitf:
```suggestion
    runs-on: ubuntu-24.04
```

PR #7503, opened 2024-06-06
fix: close watcher when FileSystemWatcher is disposed

### LINKED ISSUE #7501: Unit testing FileWatcher fails on mac, as watcher is not closed
### Bug description

When writing unit tests invloving `FileSystemWatcherImpl`, unit tests fail with the error:

```
Assertion failed: (napi_create_external(env, instance, fse_instance_destroy, NULL, &result) == napi_ok), function FSEStart, file ../src/fsevents.c, line 217.
```

It seems related to the fact that the watcher is not closed when the `FileSystemWatcherImpl` instances are disposed

### Operating system

mac os

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Fixes #7501 

### How to test this PR?

Unit tests should pass.

Specifically, the tests in packages/main/src/plugin/filesystem-monitoring.spec.ts. 

If there is only one test in packages/main/src/plugin/filesystem-monitoring.spec.ts, the issue #7501 is not reproducible, but adding a second test (by duplicating the first one for example), makes `yarn test:unit` fail with the error:

```
Assertion failed: (napi_create_external(env, instance, fse_instance_destroy, NULL, &result) == napi_ok), function FSEStart, file ../src/fsevents.c, line 217.
```

### DIFF
--- a/packages/main/src/plugin/filesystem-monitoring.ts
+++ b/packages/main/src/plugin/filesystem-monitoring.ts
@@ -32,6 +32,10 @@ export class FileSystemWatcherImpl implements containerDesktopAPI.FileSystemWatc
       persistent: true,
     });
 
+    this.watcher.on('ready', () => {
+      this._onReady.fire();
+    });
+
     this.watcher.on('add', (addedPath: string) => {
       const uri: containerDesktopAPI.Uri = Uri.file(addedPath);
       this._onDidCreate.fire(uri);
@@ -47,14 +51,16 @@ export class FileSystemWatcherImpl implements containerDesktopAPI.FileSystemWatc
       this._onDidDelete.fire(uri);
     });
 
-    this._disposable = Disposable.from(this._onDidCreate, this._onDidChange, this._onDidDelete);
+    this._disposable = Disposable.from(this._onReady, this._onDidCreate, this._onDidChange, this._onDidDelete);
   }
 
   private _disposable: containerDesktopAPI.Disposable;
+  private readonly _onReady = new Emitter<void>();
   private readonly _onDidChange = new Emitter<containerDesktopAPI.Uri>();
   private readonly _onDidCreate = new Emitter<containerDesktopAPI.Uri>();
   private readonly _onDidDelete = new Emitter<containerDesktopAPI.Uri>();
 
+  readonly onReady: containerDesktopAPI.Event<void> = this._onReady.event;
   readonly onDidChange: containerDesktopAPI.Event<containerDesktopAPI.Uri> = this._onDidChange.event;
   readonly onDidCreate: containerDesktopAPI.Event<containerDesktopAPI.Uri> = this._onDidCreate.event;
   readonly onDidDelete: containerDesktopAPI.Event<containerDesktopAPI.Uri> = this._onDidDelete.event;
--- a/packages/main/src/plugin/filesystem-monitoring.ts
+++ b/packages/main/src/plugin/filesystem-monitoring.ts
@@ -24,23 +24,25 @@ import { Disposable } from './types/disposable.js';
 import { Uri } from './types/uri.js';
 
 export class FileSystemWatcherImpl implements containerDesktopAPI.FileSystemWatcher {
+  watcher: chokidar.FSWatcher;
+
   constructor(path: string) {
     // needs to call chokidar
-    const watcher = chokidar.watch(path, {
+    this.watcher = chokidar.watch(path, {
       persistent: true,
     });
 
-    watcher.on('add', (addedPath: string) => {
+    this.watcher.on('add', (addedPath: string) => {
       const uri: containerDesktopAPI.Uri = Uri.file(addedPath);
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/filesystem-monitoring.ts
```diff
@@ -58,6 +60,11 @@ export class FileSystemWatcherImpl implements containerDesktopAPI.FileSystemWatc
   readonly onDidDelete: containerDesktopAPI.Event<containerDesktopAPI.Uri> = this._onDidDelete.event;
 
   dispose(): void {
+    if (this.watcher) {
+      this.watcher.close().catch((err: unknown) => {
```
feloy, line 70: I'm not sure if we want to do this, or if we can change the signature of the method to `async dispose(): Promise<void>`. I think that it won't respect the contract of the `Disposable` class anymore if we do this.

benoitf: we can't change the signature

benoitf (approved):

cdrage (approved):

packages/main/src/plugin/filesystem-monitoring.ts
```diff
@@ -24,27 +24,40 @@ import { Disposable } from './types/disposable.js';
...
   constructor(path: string) {
     // needs to call chokidar
-    const watcher = chokidar.watch(path, {
+    this.watcher = chokidar.watch(path, {
       persistent: true,
     });
 
-    watcher.on('add', (addedPath: string) => {
+    this.watcher.on('add', (addedPath: string) => {
+      console.log('add event');
```
benoitf: Need to be removed

packages/main/src/plugin/filesystem-monitoring.ts
```diff
@@ -24,27 +24,40 @@ import { Disposable } from './types/disposable.js';
...
+      const uri: containerDesktopAPI.Uri = Uri.file(addedPath);
+      this._onDidCreate.fire(uri);
+    });
+
+    this.watcher.on('addDir', (addedPath: string) => {
+      console.log('addDir event');
```
benoitf: to be removed

packages/main/src/plugin/filesystem-monitoring.ts
```diff
@@ -24,27 +24,40 @@ import { Disposable } from './types/disposable.js';
...
       const uri: containerDesktopAPI.Uri = Uri.file(addedPath);
       this._onDidCreate.fire(uri);
     });
 
-    watcher.on('change', (addedPath: string) => {
+    this.watcher.on('change', (addedPath: string) => {
+      console.log('change event');
```
benoitf: remove

packages/main/src/plugin/filesystem-monitoring.ts
```diff
@@ -24,27 +24,40 @@ import { Disposable } from './types/disposable.js';
...
       const uri: containerDesktopAPI.Uri = Uri.file(addedPath);
       this._onDidChange.fire(uri);
     });
 
-    watcher.on('unlink', (addedPath: string) => {
+    this.watcher.on('unlink', (addedPath: string) => {
+      console.log('unlink  event');
```
benoitf: Remove

packages/main/src/plugin/filesystem-monitoring.ts
```diff
@@ -24,27 +24,40 @@ import { Disposable } from './types/disposable.js';
...
       const uri: containerDesktopAPI.Uri = Uri.file(addedPath);
       this._onDidDelete.fire(uri);
     });
 
+    console.log('max listeners', this.watcher.getMaxListeners());
```
benoitf: Remove

feloy:
Tests constantly fail on CI. May be related to https://github.com/paulmillr/chokidar/issues/790

feloy:
> Tests constantly fail on CI. May be related to [paulmillr/chokidar#790](https://github.com/paulmillr/chokidar/issues/790)

The latest commit added the support of the `ready` event, which fixes the problem on Windows/Linux.

But for some reason that I'm still investigating, this is not compatible with the use of the `fsevents` (by default on Mac), as no events are sent after the `ready` event (which would be expected if `persistent` was false, but is set to true).

feloy:
> > Tests constantly fail on CI. May be related to [paulmillr/chokidar#790](https://github.com/paulmillr/chokidar/issues/790)
> 
> The latest commit added the support of the `ready` event, which fixes the problem on Windows/Linux.
> 
> But for some reason that I'm still investigating, this is not compatible with the use of the `fsevents` (by default on Mac), as no events are sent after the `ready` event (which would be expected if `persistent` was false, but is set to true).

Debugging in the chokidar sources, the problem occurs when the watched file is a symlink (see #7512). Changing the test to resolve the path, to avoid the problem

benoitf (approved):

PR #7511, opened 2024-06-07
chore: support different parent/child types in tables

deboer-tim:
### What does this PR do?

A previous PR added support for having child items in tables, but the type of the parent and child objects is the same.

For the Containers page, the parents are always ContainerGroupInfoUI and the children are always ContainerInfoUI. This change just adds an optional second type (identically to how we did rendering types in Column) so that you can declare the different types naturally as:
  `ContainerGroupInfoUI, ContainerInfoUI`
instead of having to do:
  `ContainerGroupInfoUI | ContainerInfoUI`
and deal with typing issues in the client code.

### Screenshot / video of UI

N/A

Part of #4843.

### How to test this PR?

Just existing tests; doesn't change any functionality.

### DIFF
--- a/packages/ui/src/lib/table/table.ts
+++ b/packages/ui/src/lib/table/table.ts
@@ -96,7 +96,7 @@ export class Column<Type, RenderType = Type> {
 /**
  * Options to be used when creating a Row.
  */
-export interface RowInformation<Type> {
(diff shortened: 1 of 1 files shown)


feloy (approved):

deboer-tim:
Minor typing change, merging.

Issue #7515, opened 2024-06-07
Update and make Compose e2e tests stable on windows and allow their execution on Linux


odockal:
### Is your enhancement related to a problem? Please describe

Compose e2e tests are failing on windows and are not run on linux.

### Describe the solution you'd like

Allow running the tests on linux, do not let them run on GHA due to #6175. fix windows failures. Improve the tests workflow so on linux they can be run partially.

odockal:
Tests are now passing on windows workflows.

PR #7517, opened 2024-06-07
chore(tests): update and stabilize compose e2e tests

odockal:
### What does this PR do?
Improves and stabilize compose-e2e tests.
#7515

### How to test this PR?
On linux set, `export COMPOSE_PARTIAL_INSTALL=true` before running the tests so it passes. `docker-compose` must not be installed on the system
```
yarn test:e2e
```


axel7083 (approved):

odockal:
I tried to run the tests on windows and it looks good: https://github.com/containers/podman-desktop-e2e/actions/runs/9418416211:

Screenshot_20240607_172700

cbr7 (approved):

benoitf (changes requested):
we should not have properties for github actions in the tests

we have settings/toggle that you can add inside the GHA but it's not something that should be detected

odockal:
> we should not have properties for github actions in the tests
> 
> we have settings/toggle that you can add inside the GHA but it's not something that should be detected

Is the latest change enough or we should proceed further and set other env. vars. for the particular tests?

rostalan (approved):
LGTM

benoitf (approved):

PR #7526, opened 2024-06-10
chore: add different colours / outputs for deployment conditions

### LINKED ISSUE #7518: Conditions in deployment do not update if there is an issue with deployment.

Screenshot 2024-06-07 at 10 17 30 AM

So we have two "conditions" for Deployment:
* Available (if it has minimum availability)
* Progressing (if replicas are up) 

However.. we are not updating this correctly and it always seems that it's 100% working.

* [ ] Colours (gray / amber / green) for if it is not yet working / fully progressed / available
* [ ] Update wording similar to Progressed / Progressing for "Available" (not available, available, etc.)

_No response_

### Additional context

_No response_

cdrage:
chore: add different colours / outputs for deployment conditions

### What does this PR do?

* Adds different colours and outputs to conditions for deployments.
* Uses all available "reasons" according to kubernetes API
* Refactors to a condition map

### Screenshot / video of UI

Screenshot 2024-06-09 at 10 44 28 PM

Closes https://github.com/containers/podman-desktop/issues/7518

### How to test this PR?

Modify a deployment / change it so that it becomes unavailable (easiest
way is to change to a non-existing image name).

See the changes on the deployment list.

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -1166,5 +1166,11 @@ export class ColorRegistry {
       dark: colorPalette.gray[500],
       light: colorPalette.gray[800],
     });
+    // Scaled / updated, use blue as it's a 'neutral' color
+    // to indicate that it's informative but not a problem
+    this.registerColor(`${status}updated`, {
+      dark: colorPalette.sky[500],
+      light: colorPalette.sky[500],
+    });
   }
 }
--- a/packages/renderer/src/lib/deployments/DeploymentColumnConditions.spec.ts
+++ b/packages/renderer/src/lib/deployments/DeploymentColumnConditions.spec.ts
@@ -50,55 +50,6 @@ test('Expect column styling available', async () => {
   expect(svg).toHaveClass('text-[var(--pd-status-running)]');
 });
 
-/* Add more tests for
-   // Condition map for easier lookup
-  const conditionMap: { [key: string]: { name: string; color: string; icon?: IconDefinition } } = {
-    'Available:MinimumReplicasAvailable': {
-      name: 'Available',
-      color: 'text-[var(--pd-status-running)]',
-      icon: faCheckCircle,
-    },
-    'Available:MinimumReplicasUnavailable': {
-      name: 'Unavailable',
-      color: 'text-[var(--pd-status-degraded)]',
-      icon: faTimesCircle,
-    },
-    'Progressing:ReplicaSetUpdated': {
-      name: 'Updated',
-      color: 'text-[var(--pd-status-updated)]',
-    },
-    'Progressing:NewReplicaSetCreated': {
-      name: 'New Replica Set',
-      color: 'text-[var(--pd-status-updated)]',
-    },
-    'Progressing:NewReplicaSetAvailable': {
-      name: 'Progressed',
-      color: 'text-[var(--pd-status-running)]',
-      icon: faSync,
-    },
-    'Progressing:ReplicaSetScaledUp': {
-      name: 'Scaled Up',
-      color: 'text-[var(--pd-status-updated)]',
-      icon: faArrowUp,
-    },
-    'Progressing:ReplicaSetScaledDown': {
-      name: 'Scaled Down',
-      color: 'text-[var(--pd-status-updated)]',
-      icon: faArrowDown,
-    },
-    'Progressing:ProgressDeadlineExceeded': {
-      name: 'Deadline Exceeded',
-      color: 'text-[var(--pd-status-dead)]',
-      icon: faTimesCircle,
-    },
-    ReplicaFailure: {
-      name: 'Replica Failure',
-      color: 'text-[var(--pd-status-dead)]',
-      icon: faExclamationTriangle,
-    },
-  };
-  */
-
 test('Expect column styling unavailable', async () => {
   const deployment = createDeploymentUI([
     { type: 'Available', message: 'Running fine', reason: 'MinimumReplicasUnavailable' },
--- a/packages/renderer/src/lib/deployments/DeploymentColumnConditions.spec.ts
+++ b/packages/renderer/src/lib/deployments/DeploymentColumnConditions.spec.ts
@@ -37,51 +37,176 @@ function createDeploymentUI(conditions: DeploymentCondition[]) {
 }
 
 test('Expect column styling available', async () => {
-  const deployment = createDeploymentUI([{ type: 'Available', message: 'Running fine' }]);
+  const deployment = createDeploymentUI([
+    { type: 'Available', message: 'Running fine', reason: 'MinimumReplicasAvailable' },
+  ]);
   render(DeploymentColumnConditions, { object: deployment });
 
   const text = screen.getByText('Available');
   expect(text).toBeInTheDocument();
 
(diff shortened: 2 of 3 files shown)


packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
```diff
@@ -9,26 +18,64 @@ export let object: DeploymentUI;
...
 
-      return { color: 'text-sky-400', icon: faSync, name };
-    case 'ReplicaFailure':
-      // faExclamationTriangle: Alerts and warnings
-      return { color: 'text-amber-600', icon: faExclamationTriangle, name };
-    default:
-      // faQuestionCircle: Uncertain / unknown
-      return { color: 'text-gray-900', icon: faQuestionCircle, name };
-  }
+  // Condition map for easier lookup
```
benoitf, line 27: need to check how these colors would behave in light mode.

And I guess that we should no longer use tailwind color but rather use CSS variable property

cdrage: You are right, I should be using https://github.com/containers/podman-desktop/blob/b6d146ab712018c5e834f4f40d3ff2a974b7aa94/tailwind.config.cjs#L49 I'll update the PR.

benoitf: no, you should use the color-registry.ts now (tailwind should only contains the palette colors)

cdrage:
Screenshot 2024-06-12 at 2 23 35 PM
Screenshot 2024-06-12 at 2 24 13 PM


I have done some additional changes @deboer-tim and @benoitf 

I've switched to using the status colours added. I've also added an "update" colour which is common in k8s GUI which is blue, meant for informational purposes and not indicating an issue with the deployment.

I've also changed the colour of the "progressing" to green if it has progressed instead of blue, so it's easier to figure out if it has actually progressed or not. It was odd that it was "ready" and available, but still blue.

Ready for another review 👍

packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
```diff
@@ -9,33 +18,74 @@ export let object: DeploymentUI;
...
 }
 </script>
 
 <div class="flex flex-row gap-1">
   {#each object.conditions as condition}
     <Label tip="{condition.message}" name="{getConditionAttributes(condition).name}">
-      <Fa size="1x" icon="{getConditionAttributes(condition).icon}" class="{getConditionAttributes(condition).color}" />
+      <Fa
+        size="1x"
+        icon="{getConditionAttributes(condition).icon ?? faQuestionCircle}"
```
deboer-tim: Don't need defaults above and ?? here, can remove this.

packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
```diff
@@ -9,33 +18,74 @@ export let object: DeploymentUI;
...
-      return { color: 'text-amber-600', icon: faExclamationTriangle, name };
-    default:
-      // faQuestionCircle: Uncertain / unknown
-      return { color: 'text-gray-900', icon: faQuestionCircle, name };
-  }
+  // Condition map for easier lookup
+  const conditionMap: { [key: string]: { name: string; color: string; icon?: IconDefinition } } = {
+    'Available:MinimumReplicasAvailable': {
+      name: 'Available',
+      color: 'text-[var(--pd-status-running)]',
```
deboer-tim, line 30: Maybe easier to read/less text if this just returns the color variable and outer code adds the rest?

packages/renderer/src/lib/deployments/DeploymentColumnConditions.spec.ts
```diff
@@ -37,51 +37,176 @@ function createDeploymentUI(conditions: DeploymentCondition[]) {
...
 
-  const dot = text.parentElement?.children[0];
-  expect(dot).toBeInTheDocument();
-  expect(dot).toHaveClass('text-green-600');
+  const svg = text.parentElement?.querySelector('svg');
+  expect(svg).toBeInTheDocument();
+  expect(svg).toHaveClass('text-[var(--pd-status-running)]');
+});
+
+/* Add more tests for
```
deboer-tim: I don't love this much copypasta b/c someone adding or changing an element in the future might not realize it's here. Not a blocker, but I would remove.

cdrage: woops! those were in-progress comments I was making

packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
cdrage: Screenshot 2024-06-13 at 11 16 28 AM

TypeScript complains unfortunately since we cannot pass in undefined.

However, I've updated the code to use a default icon in a const instead.

cdrage:
@deboer-tim updated 🚀

packages/main/src/plugin/color-registry.ts
```diff
@@ -1149,5 +1149,11 @@ export class ColorRegistry {
       dark: colorPalette.gray[500],
       light: colorPalette.gray[800],
     });
+    // Scaled / updated, use blue as it's a 'neutral' color
+    // to indicate that it's informative but not a problem
+    this.registerColor(`${status}updated`, {
```
feloy, line 1171: Can we use a more generic term for the color name, ie `status-neutral`, so we can use it for other statuses than `updated`?

packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
```diff
@@ -1,41 +1,94 @@
...
-      return { color: 'text-green-600', icon: faCheckCircle, name };
-    case 'Progressing':
-      // faSync: Often used to represent ongoing processes or operations, fitting for a "Progressing" state
-      // If reason has NewReplicaSetAvailable then it's progressed
-      if (condition.reason === 'NewReplicaSetAvailable') {
-        name = 'Progressed';
-      }
+  const defaults = {
+    name: condition.type,
+    color: 'text-gray-500',
```
feloy: Can we use a color from the color-registry here?

packages/renderer/src/lib/deployments/DeploymentColumnConditions.svelte
deboer-tim: I meant why not remove the '?' from the map in line 29 `{ name: string; color: string; icon?: IconDefinition }` instead? It always returns an icon already. Then you can remove the ??, return faQuestionCircle from line 25, and delete line 18.

deboer-tim (approved):
I would still prefer this simplification:
- Remove `?` from line 29.
- Remove `?? defaultUnknownIcon` from line 90.
- Replace `defaultUnknownIcon` with `faQuestionCircle` on line 25.
- Remove line 18.

But approving either way. Thanks for using status colors 👍🏼.

cdrage:
@deboer-tim I apologize, I thought I covered it in the review, but I didn't. Thank you, I've added a new commit called 'update' with the changes. Looks a lot more cleaner 👍

deboer-tim (approved):
👍🏼

PR #7535, opened 2024-06-10
chore: moving color definitions to dedicated file

axel7083:
### What does this PR do?

We currently add all colors definition to the `color-registry.ts` file. This PR moves them to a dedicated **dependency-free** file.

The last part is important, as for https://github.com/containers/podman-desktop/issues/7439 we need to create a simple CLI tool, that will generate a stylesheet based on those colors, however the ColorRegistry class has too much dependency and `tsx` cannot execute it, as it uses internal imports etc.

Related to https://github.com/containers/podman-desktop/issues/7439

### How to test this PR?

To ensure we did not lost any color while moving them, I run in debugger mode Podman-Desktop, and break on 

https://github.com/containers/podman-desktop/blob/2fa6e77963a7b93de09010117f9b6e04c6e1bbf3/packages/main/src/plugin/color-registry.ts#L219

With my PR checkout I copy the result of `this.listColors('dark')` and `this.listColors('light')` to a dedicated file. Then repeat the operation on the main branch.

Finally I compare using a diff the results from my checkout and the main to ensure no colors has been lost.


benoitf:
I'm not sure to follow the 'to a dedicated dependency-free file.' as we're importing files in this file anyway.

Check the `/` path imports of color-registry, I guess that using relative imports should make it work as well (so no need for extraction) (replace `/@/plugin/foo.js` by `./foo.js`

deboer-tim:
What happens if we create another theme, e.g. 'light', 'dark', and then 'tweed' or 'apricot'? Maybe in another extension, built-in or not. Do we just say we only show the first two in Storybook? Otherwise I think we'd always need something in the build that would extract these at PD 'runtime'.

axel7083:
> I'm not sure to follow the 'to a dedicated dependency-free file.' as we're importing files in this file anyway.

I see two reasons why we should split the colors defintion from the ColorRegistry

## Separation of concern

The colors are constant, and make the color-registry harder to read as it makes it very big for nothing.

## Maintainability for the CLI script

The color-registry would force use to mock the ApiSenderType and ConfigurationRegistry to be able to use it, while we do not really need all of its feature, **it does not generate the css*, it only stores the color, so we can avoid some pain by moving the colors to a file which is importing the minimum `tailwind-color-palette.json`, all other imports being types, we will not face any issue by running tsx on it nor having to mock anything now or in the future.

>  I guess that using relative imports should make it work as well

If at some point anyone uses a non-relative import in any of dependencies of the ColorRegistry it would break, this would add unnecessary maintenance and caution to have, while we are fully using aliases, that would be weird to avoid in some placing using it, and other continue using it.

axel7083:
> What happens if we create another theme, e.g. 'light', 'dark', and then 'tweed' or 'apricot'? Maybe in another extension, built-in or not. Do we just say we only show the first two in Storybook? Otherwise I think we'd always need something in the build that would extract these at PD 'runtime'.

I honestly do not like the way we produce the color in the backend when it is a frontend matter.

benoitf:
the problem is that people are modifying a lot this file during this sprint so I would defer the change on the file. It will avoid conflicting files.
Also probably refactor the file before moving the content to a separate file (as it's the worst case as someone could merge another PR while the file has been moved, etc)

> If at some point anyone uses a non-relative import in any of dependencies of the ColorRegistry it would break, 

it would be the same for the color file as well, you would just move it one step ahead.
but as the CLI should be called anyway during a PR check then it would break the build so we can't merge the PR

> I honestly do not like the way we produce the color in the backend when it is a frontend matter.

I think I already answered the same exact question. Css is loaded on the frontend but generated at the backend level.

axel7083:
> the problem is that people are modifying a lot this file during this sprint so I would defer the change on the file. It will avoid conflicting files.

Closing then

benoitf:
I like the idea of moving it to a separate file anyway, but probably after the 1.11 release

Issue #7536, opened 2024-06-10
Add Extension-Installation.spec.ts missing video recording name, fix test issues in Windows Pipeline


ScrewTSW:
### Is your enhancement related to a problem? Please describe

* We are currently missing a named video recording from extensions-installation.spec.ts
* Test is unstable, times out at extension installation

### Describe the solution you'd like

* Add pdRunner.setVideoAndTraceName() call
* Fix test instability

### Additional context

![brave_G9AJR7KpKK](https://github.com/containers/podman-desktop/assets/16451875/2cc6d942-3a10-48cb-9dd3-84e518d0b26f)
![brave_Og9KXnqYG2](https://github.com/containers/podman-desktop/assets/16451875/f9776168-4643-4356-b81a-05cf74d94d3a)
![brave_t6kAGniWxG](https://github.com/containers/podman-desktop/assets/16451875/bad9431c-f3d3-418d-a5a5-119d00ae655d)
![brave_gNDhhVB5Fr](https://github.com/containers/podman-desktop/assets/16451875/67dee8a8-77c1-4245-a136-58f60d1a0735)

PR #7537, opened 2024-06-10
chore(test): added missing recording tag and trace

ScrewTSW:
### What does this PR do?
* Adds missing trace and filename to video recording

#7536  

### How to test this PR?

### DIFF
--- a/tests/playwright/src/specs/extension-installation.spec.ts
+++ b/tests/playwright/src/specs/extension-installation.spec.ts
@@ -84,6 +84,7 @@ describe.each([
 
   test('Initialize extension type', async () => {
     initializeLocators(extensionType);
+    pdRunner.setVideoAndTraceName(`${extensionName}-installation-e2e`);
     extensionCard = new ExtensionCardPage(page, extensionLabel, extensionName);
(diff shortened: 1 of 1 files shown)


tests/playwright/src/specs/extension-installation.spec.ts
```diff
@@ -56,6 +56,7 @@ let navBar: NavigationBar;
 const _startup = async function (): Promise<void> {
   pdRunner = new PodmanDesktopRunner();
   page = await pdRunner.start();
+  pdRunner.setVideoAndTraceName('extension-installation-e2e');
```
odockal: This is parameterized test. You need to use naming by a parameter value. Otherwise it will get overriden or fail to save the file.

odockal (changes requested):

tests/playwright/src/specs/extension-installation.spec.ts
ScrewTSW: fixed

cbr7 (approved):

odockal (approved):

PR #7540, opened 2024-06-10
feat: add rosetta support

### LINKED ISSUE #7367: Add Rosetta settings for Apple Silicon Mac

Rosetta support has been added in [Podman v5.1.0](https://github.com/containers/podman/releases/tag/v5.1.0).
Rosetta is enabled by default on Apple Silicon Mac Podman machines.
To disable this Rosetta and instead enable the traditional qemu-user-static-x86_64, add `rosetta=false` to containers.conf.

This RFE allows you to enable/disable the rosetta feature via GUI.

Screenshot 2024-05-30 at 9 39 50

https://github.com/containers/common/blob/ed2a5b62597ce1c35d939945c787658a3679e2f8/docs/containers.conf.5.md?plain=1#L957-L960

_No response_

### Additional context

Rosetta configuration for Docker Desktop
docker_rosetta

cdrage:
feat: add rosetta support

### What does this PR do?

* Adds rosetta support to the configuration page of Podman.
* When disabling it will add the rosetta=false key
* When enabling, it will remove the rosetta key.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/6422176/e01e2d54-3f68-462d-8f38-0e5b29eb2c68

Closes https://github.com/containers/podman-desktop/issues/7367

### How to test this PR?

1. Be on Mac
2. Disable the setting (set to false)
3. See it edited containers.conf and added rosetta=false
4. Switch to enable
5. See it removed the rosetta key.


extensions/podman/src/podman-configuration.ts
```diff
@@ -93,6 +95,81 @@ export class PodmanConfiguration {
...
+        if (e.affectsConfiguration(`podman.${configurationRosetta}`)) {
+          await this.handleRosettaSetting();
+        }
+      });
+    }
+  }
+
+  async handleRosettaSetting(): Promise<void> {
+    const useRosetta =
+      extensionApi.configuration.getConfiguration('podman').get<boolean>(configurationRosetta) ?? false;
```
feloy: Isn't the default value `true`? (in package.json)

extensions/podman/package.json
```diff
@@ -170,6 +170,12 @@
           "default": false,
           "description": "Docker compatibility mode is a feature that allows you to use Podman as a drop-in replacement for Docker-compatible CLI tools. Podman will automatically emulate a Docker API socket connection (typically) to /var/run/docker.sock. NOTE: This requires administrative privileges.",
           "when": "isMac"
+        },
+        "podman.setting.rosetta": {
+          "type": "boolean",
+          "default": true,
+          "description": "Use macOS Rosetta to run x86_64 Linux binaries, disabling will use qemu. NOTE: Recreating the Podman Machine VM is required for the changes to take effect.",
```
deboer-tim: 'NOTE: Recreating the Podman Machine VM is required for the changes to take effect.' is a bit odd if we don't know if you have a machine, or whether you used the setting. Maybe 'NOTE: This setting is only applied when creating machines.' or 'NOTE: This setting is only applied to new machines.'?

cdrage:
@feloy @deboer-tim updated!

slemeur:
How do we handle the case podman version is < 5.1?

cdrage:
> How do we handle the case podman version is < 5.1?

We don't have a good way of detecting the version within configuration and not displaying the option.

The option however is ignored if running < 5.1

So we can just say in the description "Only works on Podman >5.1.0" if that works with you.

deboer-tim (approved):
Code LGTM and worked in testing.

Need a reply from @slemeur if current wording is good.

extensions/podman/src/podman-configuration.ts
```diff
@@ -93,6 +95,82 @@ export class PodmanConfiguration {
...
+    const containersConfContent = {
+      containers: toml.Section({}),
+      engine: toml.Section({
+        env: [] as string[],
+      }),
+      machine: toml.Section({}),
+      network: toml.Section({}),
+      secrets: toml.Section({}),
+      configmaps: toml.Section({}),
+    };
```
feloy, line 127: Why are you creating an intermediary structure for the config? 

Isn't it possible to work directly on the result of `toml.parse(containersConfigFile)`, so even if the config file evolves and there are some new sections, it will still work?

cdrage: unable too as the information passed back by `toml.parse` is any and typescript complains.

I tried implementing it similar to your suggestion, but wasn't able to find a good solution. The code below (proxy update) by @benoitf was the inspiration I took this from.

this initializing + making sure that the information gets filled by checking the saved information would ensure that it all gets passed through.

feloy (approved):

slemeur:
Works for me

PR #7541, opened 2024-06-10
feat: generate storybook css dynamically

### LINKED ISSUE #7439: generate and store the css for the light and dark mode

when we're using Podman Desktop, the backend generates styles and we expose colors on the fly by injecting the content/ result of the CSS in the main page or inside the webview

As we have other places where we could reuse the styles, it will be nice if the color-registry could have a command to generate the two css so for example we could import and reuse those css as part of the stories used by storybook

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Generate the css dynamically using a typescript file located at `scripts/generate-stylesheet.ts`.

Fixes https://github.com/containers/podman-desktop/issues/7439

### How to test this PR?


scripts/generate-stylesheet.ts
```diff
@@ -0,0 +1,48 @@
...
+  receive: () => ({ dispose: () => {} }),
+};
+
+const configurationRegistryMock: ConfigurationRegistry = {
+  onDidChangeConfiguration: () => {},
+} as unknown as ConfigurationRegistry;
+
+// Create color registry
+const registry = new ColorRegistry(apiSenderTypeMock, configurationRegistryMock);
+registry.init();
```
benoitf: can we have another constructor in ColorRegistry ?

benoitf (commented):
I'm unsure about the theme argument (like why it can't accept multiple values at the same time like both light and dark)

and probably we can generate all the files in the output folder (output being a folder and not a file) ?

so the CLI will generate all the css files for all the supported themes of Podman Desktop (can be light, dark, and if tomorrow there are themes with higher contrasts, etc it will be generated automatically as well)

scripts/generate-stylesheet.ts
```diff
@@ -0,0 +1,48 @@
...
+function getArgument(key: string): string {
+  const args = process.argv.slice(2);
+
+  const index = args.indexOf(key);
+  if (index === -1 || args.length <= index + 1) throw new Error(`argument ${key} not defined.`);
+  return args[index + 1];
+}
+
+const output = getArgument('--output');
+const theme = getArgument('--theme');
```
benoitf: you might use the commander or yargs library that is already in the transitive dependencies of Podman Desktop to ease the parsing of flags

axel7083:
> I'm unsure about the theme argument (like why it can't accept multiple values at the same time like both light and dark)
> and probably we can generate all the files in the output folder (output being a folder and not a file) ?

No here I am using the output argument as a file that will be written 

> so the CLI will generate all the css files for all the supported themes of Podman Desktop (can be light, dark, and if tomorrow there are themes with higher contrasts, etc it will be generated automatically as well)

Okey will change it

scripts/generate-stylesheet.ts
```diff
@@ -0,0 +1,48 @@
+import { ColorRegistry } from '../packages/main/src/plugin/color-registry';
```
benoitf, line 20: missing copyright headers and the header to execute the file directly
```
#!/usr/bin/env tsx
```

scripts/generate-stylesheet.ts
axel7083: `TS2392: Multiple constructor implementations are not allowed.` apparently no

benoitf: it's possible but you need to handle all of them with the same block

benoitf: anyway it's a low comment, as the goal is to move it to a separate file soon

scripts/generate-stylesheet.ts
axel7083: Applied

scripts/generate-stylesheet.ts
axel7083: I will be using minimist also in transitive dependency

scripts/generate-stylesheet.ts
axel7083: Ok

benoitf (approved):
missing a unit test for generate-stylesheet

axel7083:
> missing a unit test for generate-stylesheet

@benoitf  `scripts` is not a package, therefore vitest cannot find the `*.spec.ts` file in it.

benoitf:
@axel7083 either we include scripts or you move the generator to the main package

axel7083:
> @axel7083 either we include scripts or you move the generator to the main package

Tests has been _painfully_ added ✅

PR #7542, opened 2024-06-10
Light mode for extensions catalog

### LINKED ISSUE #7236: Light mode for extension catalog

We need to support light mode in the extensions/catalog - extension names, details page/readme, etc.

Define missingcolors in the light mode palette and use them on the components.

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Light mode for Extensions catalog

Fixes #7236 

### How to test this PR?

Navigate in Extensions catalog, installed, and details pages

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -446,6 +446,11 @@ export class ColorRegistry {
       light: colorPalette.charcoal[900],
     });
 
+    this.registerColor(`${ct}text`, {
+      dark: colorPalette.gray[700],
+      light: colorPalette.gray[900],
+    });
+
     this.registerColor(`${ct}sub-header`, {
       dark: colorPalette.gray[900],
       light: colorPalette.purple[900],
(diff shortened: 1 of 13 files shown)


packages/main/src/plugin/color-registry.ts
```diff
@@ -840,6 +857,10 @@ export class ColorRegistry {
       dark: colorPalette.white + '2',
       light: colorPalette.black + '2',
     });
+    this.registerColor(`${link}-light`, {
+      dark: colorPalette.gray[200],
+      light: colorPalette.charcoal[900],
+    });
```
benoitf: not sure to understand the meaning of link-light ?

deboer-tim: This looks like the "More details" button. IMHO it would be better to make the entire card clickable (so this link isn't special/doesn't need its own color) or use Link here vs defining another custom color.

deboer-tim (commented):
The button to install (FeaturedExtensionDownload) is still in dark mode, and (before this change) has different colors on the catalog page vs extension details page. Even though it is a custom button I think we should reuse the secondary button colors vs defining new ones.

packages/main/src/plugin/color-registry.ts
feloy: OK, I changed to use a `Button type="link"`, does it seem ok?

feloy:
> The button to install (FeaturedExtensionDownload) is still in dark mode, and (before this change) has different colors on the catalog page vs extension details page. Even though it is a custom button I think we should reuse the secondary button colors vs defining new ones.

I added the FeaturedExtensionDownload, by reusing the style of secondary button

benoitf (approved):

lstocchi (approved):
Maybe it's just personal taste but I would add a shadow or use a different color to make the separation between the left bar and the body of the page more clear. 
Overall LGTM
image

PR #7554, opened 2024-06-10
fix(typing): improve ColumnInformation renderer property typing

axel7083:
### What does this PR do?

Improving the renderer property from `any` to more defined. This enforce that column have the `object` property matching the expected type.

This change force use to correct the existing typing which were used for columns arrays, which were incorrect: I removed it as not needed.

### How to test this PR?

- Pipeline should be ✅


deboer-tim (approved):
It's almost unfortunately that the type is 'removed' from the columns, but typing in table.ts is awesome and far outweighs this. Thanks!

PR #7563, opened 2024-06-10
chore: adds updated details page to images and manifests

### LINKED ISSUE #7270: Manifests "more in depth" details page.

Elaboration on issue: https://github.com/containers/podman-desktop/issues/7255

We should take the information that's shown in `podman manifest inspect` and incorporate it into the details page.

This will involve adding a function such as "inspect manifest" when loading summary.

Show more robust details for example:

```json
podman manifest inspect test123
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 11045,
            "digest": "sha256:56de9526080be867bb85a5e254cdede7a97b14a22ed03324803edbf5dfd60f52",
            "platform": {
                "architecture": "arm64",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 11046,
            "digest": "sha256:e8dd33efa3b425e0f2db22244c2c3fc54b7eadb090b78fe08bf9b7fdf8891129",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        }
    ]
}
```

_No response_

### Additional context

_No response_

### LINKED ISSUE #7561: Images: Summary UI update

See: https://github.com/containers/podman-desktop/issues/7558

Update the summary page to match rest of the UI.

Update for image

_No response_

### Additional context

_No response_

cdrage:
chore: adds updated details page to images and manifests

### What does this PR do?

* Changes the UI for images to the new format that is on other summary
  pages
* Adds details for manifest (using inspect)

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/6422176/c4cfc162-7dac-4f60-92ad-2aa8c7eaedc3

Closes https://github.com/containers/podman-desktop/issues/7561
Closes https://github.com/containers/podman-desktop/issues/7270

### How to test this PR?

View the summary pages for both an image and a manifest

### DIFF
--- a/packages/renderer/src/lib/kube/details/KubeServiceArtifact.svelte
+++ b/packages/renderer/src/lib/kube/details/KubeServiceArtifact.svelte
@@ -1,8 +1,8 @@
 <script lang="ts">
 import type { V1ServiceSpec } from '@kubernetes/client-node';
 
-import Cell from './ui/Cell.svelte';
-import Title from './ui/Title.svelte';
+import Cell from '/@/lib/details/DetailsCell.svelte';
+import Title from '/@/lib/details/DetailsTitle.svelte';
 
 export let artifact: V1ServiceSpec | undefined;
 </script>
--- a/packages/renderer/src/lib/details/DetailsCell.svelte
+++ b/packages/renderer/src/lib/details/DetailsCell.svelte
@@ -3,6 +3,6 @@ export let style: string = '';
 export let onClick: () => void = () => {};
 </script>
 
-<td class="pt-1 pl-3 text-gray-300 {style}" on:click="{onClick}">
+<td class="pt-1 pl-3 {style}" on:click="{onClick}">
   <slot />
 </td>
--- a/packages/renderer/src/lib/details/DetailsSubtitle.svelte
+++ b/packages/renderer/src/lib/details/DetailsSubtitle.svelte
@@ -1,0 +1,7 @@
+<script lang="ts">
+export let style: string = '';
+</script>
+
+<td class="pl-2 text-md font-semibold text-[var(--pd-table-body-text-sub-secondary)] {style}">
+  <slot />
+</td>
--- a/packages/renderer/src/lib/details/DetailsTable.svelte
+++ b/packages/renderer/src/lib/details/DetailsTable.svelte
@@ -1,0 +1,7 @@
+<div class="flex px-5 py-4 flex-col items-start h-full overflow-auto text-[var(--pd-table-body-text)]">
+  <table class="w-full">
+    <tbody>
+      <slot />
+    </tbody>
+  </table>
+</div>
--- a/packages/renderer/src/lib/details/DetailsTitle.svelte
+++ b/packages/renderer/src/lib/details/DetailsTitle.svelte
@@ -1,0 +1,3 @@
+<td class="pt-1 text-lg font-semibold text-[var(--pd-table-body-text-sub-secondary)]">
+  <slot />
+</td>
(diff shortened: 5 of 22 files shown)


deboer-tim (dismissed):
Code LGTM and worked well in testing. However, the page looks ok in light mode before this change and is almost entirely washed out afterward. Might be related to removing `text-[var(--pd-details-body-text)]`?

IMHO we've passed the point where unrelated changes shouldn't regress light mode, and likely at the point where we should fix code we touch for unrelated reasons. Can you please fix the 'regression' here, and then would you mind follow-up PR to define the colors for Title/Subtitle/Cell in the color registry and switch to using variables?

cdrage:
> Code LGTM and worked well in testing. However, the page looks ok in light mode before this change and is almost entirely washed out afterward. Might be related to removing `text-[var(--pd-details-body-text)]`?
> 
> IMHO we've passed the point where unrelated changes shouldn't regress light mode, and likely at the point where we should fix code we touch for unrelated reasons. Can you please fix the 'regression' here, and then would you mind follow-up PR to define the colors for Title/Subtitle/Cell in the color registry and switch to using variables?

Thank you! Learned a lot on how to actually implement the colors properly.

I've updated the PR to also update Subtitle, Cell, Title and Table svelte files for details.

I have used the "pd-table" section of color-registry.ts so it will match / be consistent similar to our dashboard tables. This does change the dark mode a little bit so that it's now more gray / rather than white, which I believe is okay since it's part of the contrast table that @ekidneyrh was using.

Screenshots are below, let me know if this is good and will follow up with a PR for https://github.com/containers/podman-desktop/issues/7575 pages as this PR will need a follow up PR as it breaks summary pages for k8s (unviewable in light mode).

Screenshot 2024-06-11 at 9 44 48 AM
Screenshot 2024-06-11 at 9 45 07 AM

lstocchi (approved):
Tested. LGTM

packages/renderer/src/lib/details/Subtitle.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
```
benoitf: name should be DetailsSubtitle (to not let think it's a Subtitle on its own)

packages/renderer/src/lib/details/Table.svelte
```diff
@@ -0,0 +1,7 @@
+<div class="flex px-5 py-4 flex-col items-start h-full overflow-auto text-[var(--pd-table-body-text)]">
```
benoitf: rename it to DetailsTable

packages/renderer/src/lib/details/Title.svelte
```diff
@@ -0,0 +1,3 @@
+<td class="pt-1 text-lg font-semibold text-[var(--pd-table-body-text-sub-secondary)]">
```
benoitf: rename it to DetailsTitle

packages/renderer/src/lib/kube/details/KubeServiceArtifact.svelte
```diff
@@ -1,8 +1,8 @@
 <script lang="ts">
 import type { V1ServiceSpec } from '@kubernetes/client-node';
 
-import Cell from './ui/Cell.svelte';
-import Title from './ui/Title.svelte';
+import Cell from '../../details/Cell.svelte';
```
benoitf: you have `/@/` imports to avoid to use relative imports

cdrage:
@benoitf updated with the suggested changes

benoitf (approved):
thanks

Issue #7360, opened 2024-05-29
Fix failing dev sandbox extension installation test


odockal:
### Is your enhancement related to a problem? Please describe

It seems like instability is present on windows workflow regarding Dev Sandbox e2e tests. Ref: https://github.com/containers/podman-desktop-e2e/actions/runs/9281067966

### Describe the solution you'd like

Increase the timeout for the extension to get installed in PD.

PR #7565, opened 2024-06-10
chore(tests): redo and stabilize extensions e2e tests

odockal:
### What does this PR do?
Adjusts the extension-installation tests to use extensions catalog which should improve the stability of the tests

#7360

### How to test this PR?
`yarn test:e2e:extensions`

### DIFF
--- a/tests/playwright/src/model/pages/extension-catalog-card-page.ts
+++ b/tests/playwright/src/model/pages/extension-catalog-card-page.ts
@@ -48,15 +48,16 @@ export class ExtensionCatalogCardPage extends BasePage {
   public async isInstalled(): Promise<boolean> {
     await this.parent.scrollIntoViewIfNeeded();
     const downloadButton = this.parent.getByRole('button', { name: 'Install' });
-    return (await this.alreadyInstalledText.isVisible()) && !(await downloadButton.isVisible()) ? true : false;
+    return (await this.alreadyInstalledText.count()) > 0 && (await downloadButton.count()) === 0 ? true : false;
   }
 
   public async install(timeout: number): Promise<void> {
-    if (!(await this.isInstalled())) {
-      await playExpect(this.downloadButton).toBeVisible();
-      await this.downloadButton.click();
-      await playExpect(this.alreadyInstalledText).toBeVisible({ timeout: timeout });
+    if (await this.isInstalled()) {
+      console.log(`Extension ${this.extensionName} is already installed`);
+      return;
     }
-    console.log(`Extension ${this.extensionName} is already installed`);
+    await playExpect(this.downloadButton).toBeEnabled();
+    await this.downloadButton.click();
+    await playExpect(this.alreadyInstalledText).toBeVisible({ timeout: timeout });
   }
 }
--- a/tests/playwright/src/model/pages/extension-catalog-card-page.ts
+++ b/tests/playwright/src/model/pages/extension-catalog-card-page.ts
@@ -1,0 +1,62 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Locator, Page } from '@playwright/test';
+import { expect as playExpect } from '@playwright/test';
+
+import { BasePage } from './base-page';
+import { ExtensionDetailsPage } from './extension-details-page';
+
+export class ExtensionCatalogCardPage extends BasePage {
+  readonly parent: Locator;
+  readonly extensionName: string;
+  readonly detailsButton: Locator;
+  readonly downloadButton: Locator;
+  readonly alreadyInstalledText: Locator;
+
+  constructor(page: Page, extensionName: string) {
+    super(page);
+    this.extensionName = extensionName;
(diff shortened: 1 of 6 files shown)


tests/playwright/src/model/pages/extension-catalog-card-page.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.extensionName = extensionName;
+    this.parent = this.page.getByRole('group', { name: this.extensionName });
+    this.detailsButton = this.parent.getByRole('button', { name: 'More details' });
+    this.downloadButton = this.parent.getByRole('button', { name: 'Install' });
+    this.alreadyInstalledText = this.parent.getByText('Already installed', { exact: true });
+  }
+
+  public async openDetails(): Promise<ExtensionDetailsPage> {
+    await this.parent.scrollIntoViewIfNeeded();
+    await playExpect(this.detailsButton).toBeVisible();
```
cbr7, line 43: toBeEnabled() should be used here or whenever following the assert there will be interaction as it is here with the click() command

tests/playwright/src/model/pages/extension-catalog-card-page.ts
```diff
@@ -0,0 +1,62 @@
...
+    await this.detailsButton.click();
+    return new ExtensionDetailsPage(this.page, this.extensionName);
+  }
+
+  public async isInstalled(): Promise<boolean> {
+    await this.parent.scrollIntoViewIfNeeded();
+    const downloadButton = this.parent.getByRole('button', { name: 'Install' });
+    return (await this.alreadyInstalledText.isVisible()) && !(await downloadButton.isVisible()) ? true : false;
```
cbr7: await locator.count() > 0 should be used to check existence of an element on the page since count() will not ever throw exceptions and also please do not use ternary operators in such complex and long commands, it makes it very difficult to debug and understand what is happening, especially since you also use negative conditions in it.

tests/playwright/src/model/pages/extension-catalog-card-page.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public async install(timeout: number): Promise<void> {
+    if (!(await this.isInstalled())) {
```
cbr7: This can be simplified 
`` if (await this.isInstalled()){
console.log(`Extension ${this.extensionName} is already installed`);
return
}``

followed by the rest of the commands outside brakets, this way we avoid using negative conditionals

tests/playwright/src/model/pages/extension-catalog-card-page.ts
```diff
@@ -0,0 +1,62 @@
...
+      await playExpect(this.downloadButton).toBeVisible();
```
cbr7: toBeEnabled() should be used here.

tests/playwright/src/specs/extension-installation.spec.ts
```diff
@@ -73,60 +69,59 @@ beforeEach<RunnerTestContext>(async ctx => {
...
+    await extensionCatalog.install(90000);
 
-  test('Install extension through Extensions page', async () => {
-    const extensionsPage = new ExtensionsPage(page);
-
-    await extensionsPage.installExtensionFromOCIImage(imageLink);
-    const extensionDetailsPage = await extensionsPage.openExtensionDetails(extensionName, extensionLabel);
+    await extensionsPage.openInstalledTab();
+    await playExpect
+      .poll(async () => await extensionsPage.extensionIsInstalled(extensionLabel), { timeout: 5000 })
```
cbr7: Default timeout for waits is 5 seconds, so there is no reason to add it explicitly except if you want to set a different value.

tests/playwright/src/specs/extension-installation.spec.ts
```diff
@@ -73,60 +69,59 @@ beforeEach<RunnerTestContext>(async ctx => {
...
+    const extensionCatalog = new ExtensionCatalogCardPage(page, extensionType);
+    await playExpect(extensionCatalog.parent).toBeVisible();
 
-    // requires to be rewritten, as extension are not shuffled anymore, but are part of extensions catalog
-    test.runIf(extensionBoxVisible)('Check Settings extension component for installation availability', async () => {
-      const installButton = extensionCard.card.getByRole('button', { name: installButtonLabel });
-      await playExpect(installButton).toBeVisible();
-    });
-  });
+    await playExpect.poll(async () => await extensionCatalog.isInstalled(), { timeout: 5000 }).toBeFalsy();
```
cbr7: Default timeout for waits is 5 seconds, so there is no reason to add it explicitly except if you want to set a different value.

cbr7 (changes requested):
In general .isVisible() should never be used as a call to check existence, count() > 0 is the recommended pattern

tests/playwright/src/model/pages/extension-catalog-card-page.ts
odockal: Usually I would agree, except here it is very strange usage of the button where it serves as a link (visually). 
Screenshot_20240611_081001

tests/playwright/src/specs/extension-installation.spec.ts
odockal: did not find in javadoc. good.

tests/playwright/src/model/pages/extension-catalog-card-page.ts
odockal: done.

tests/playwright/src/specs/extension-installation.spec.ts
odockal: nice.

tests/playwright/src/model/pages/extension-catalog-card-page.ts
odockal: done.

tests/playwright/src/model/pages/extension-catalog-card-page.ts
odockal: improved. now it looks better, although two conditions without a negative usage seems something that should not be hard to read in use with ternary operator.

tests/playwright/src/model/pages/extension-catalog-card-page.ts
cbr7: weird, ok

cbr7 (approved):

PR #7552, opened 2024-06-10
feat: adds images under manifest

cdrage:
feat: adds images under manifest

### What does this PR do?

* Adds "children" images under the main manifest

### Screenshot / video of UI

Screenshot 2024-06-10 at 11 43 55 AM

https://github.com/containers/podman-desktop/assets/6422176/0077cb19-7cf2-4c0f-bce8-ee020b243907

### What issues does this PR fix or reference?

Closes https://github.com/containers/podman-desktop/issues/7181

### How to test this PR?

- [X] Tests are covering the bug fix or the new feature

1. Create a manifest
2. No longer see the <none> images and it will instead be under the
   manifest image.

Signed-off-by: Charlie Drage <charlie@charliedrage.com>

PR #7587, opened 2024-06-11
fix: table component issues with children

### LINKED ISSUE #7585: Table component issues with children
### Bug description

Testing the Table component with child objects (especially the complexities in the Containers page), I've found a few more issues to fix:
- The default should be expanded children, not collapsed. This means switching over all the logic from 'expanded' to 'collapsed'.
- The selectedItemsNumber and selectedAllCheckboxes were wrong because they weren't taking children into account.
- Child object selection div has incorrect rounded corners when there is more than one child.

### Operating system

macOS 14.5

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Fixes the following things for tables with children:
- Changes the default from all collapsed to all expanded. Easier to see everything and is the default in the Containers view. This meant switching over all the logic from 'expanded' to 'collapsed'. Downstream users would now bind to the list of collapsed items instead of expanded ones.
- Fixed selectedItemsNumber and selectedAllCheckboxes, which weren't taking children into account.
- Fixed child object selection div rounded corners when there is more than one child.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/19958075/03962f5e-3d07-4f4b-8529-5240921178a6

Fixes #7585.

### How to test this PR?

I'll update #7424 to include this in a moment or you can merge with #7552, but otherwise hard to replicate atm.

### DIFF
--- a/packages/ui/src/lib/table/Table.svelte
+++ b/packages/ui/src/lib/table/Table.svelte
@@ -20,7 +20,6 @@ export let kind: string;
 export let columns: Column<any>[];
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export let row: Row<any>;
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export let data: { selected?: boolean; name?: string }[];
 export let defaultSortColumn: string | undefined = undefined;
 export let collapsed: string[] = [];
--- a/packages/ui/src/lib/table/Table.svelte
+++ b/packages/ui/src/lib/table/Table.svelte
@@ -21,7 +21,7 @@ export let columns: Column<any>[];
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export let row: Row<any>;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let data: { selected?: boolean; name?: string; id?: string }[];
+export let data: { selected?: boolean; name?: string }[];
 export let defaultSortColumn: string | undefined = undefined;
 export let collapsed: string[] = [];
 
@@ -161,18 +161,18 @@ function objectChecked(object: { selected?: boolean }): void {
   }
 }
 
-function toggleChildren(id: string | undefined): void {
-  if (!id) {
+function toggleChildren(name: string | undefined): void {
+  if (!name) {
     return;
   }
 
-  if (collapsed.includes(id)) {
-    const index = collapsed.indexOf(id, 0);
+  if (collapsed.includes(name)) {
+    const index = collapsed.indexOf(name, 0);
     if (index > -1) {
       collapsed.splice(index, 1);
     }
   } else {
-    collapsed.push(id);
+    collapsed.push(name);
   }
   // trigger Svelte update
   collapsed = collapsed;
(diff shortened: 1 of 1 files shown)


packages/ui/src/lib/table/Table.svelte
```diff
@@ -25,17 +25,27 @@ export let data: { selected?: boolean; name?: string; id?: any }[];
 export let defaultSortColumn: string | undefined = undefined;
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let expanded: any[] = [];
+export let collapsed: any[] = [];
```
axel7083: Do we really need the any here ?

deboer-tim: I assume you mean 'does it really need to be 'any' or should it be a string'?

It was 'any' because originally I was storing the row object itself, which turned out to be a bad idea b/c the UI objects are ephemeral, then I switched it to requiring that any object you want to collapse must have an id. I think it's reasonable enough and fits our cases that the id is a string, so I've switched it in another commit.

axel7083: Okey good!

packages/ui/src/lib/table/Table.svelte
```diff
@@ -21,21 +21,31 @@ export let columns: Column<any>[];
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export let row: Row<any>;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
axel7083: No needed anymore ?

axel7083 (approved):

packages/ui/src/lib/table/Table.svelte
deboer-tim: Good catch! Pushed another commit to remove it.

deboer-tim:
After Charlie noticed a problem in the Containers page porting I went back, re-evaluated and replaced the last commit. I had been requiring an 'id' property to exist for things that are expandable/collapsable, but we have been using the name property for this and some objects don't have an id. Switched to using name for everything, which is simpler and removes one requirement.

cdrage:
> After Charlie noticed a problem in the Containers page porting I went back, re-evaluated and replaced the last commit. I had been requiring an 'id' property to exist for things that are expandable/collapsable, but we have been using the name property for this and some objects don't have an id. Switched to using name for everything, which is simpler and removes one requirement.

There are unfortunately cases that there are images with the same name such as the `<none>` tag :(

This would most likely break the manifest children images functionality since it will no longer be unique:

Screenshot 2024-06-12 at 12 21 24 PM

deboer-tim:
The case posted should work, because the name is only being used to remember which items are collapsed (i.e. `localhost/foobar (manifest)` in this case), not the children.

FWIW I think there _are_ going to be cases where we will face collisions in the future and need to use a different key, but I think it will break other things already (aria-labels, the key we used to use for svelte sort animation/uniqueness, etc), so I was deferring a requirement to have an id for now.

cdrage (changes requested):
https://github.com/containers/podman-desktop/assets/6422176/09e8c2e8-22c2-4974-9b67-e68362f47359

I rebased this against my manifest PR with the multiple children images and unfortunately actions do not work?

I'm unsure if it's because of change to name instead of ID, but I'm unable to delete anything that's a children image.

deboer-tim:
> I rebased this against my manifest PR with the multiple children images and unfortunately actions do not work?
> 
> I'm unsure if it's because of change to name instead of ID, but I'm unable to delete anything that's a children image.

I can't find anything in this change that would affect child actions, and I'm able to delete and do other actions against #7552 with or without this change... Is there an error message for you, or maybe an issue but unrelated to this PR?

cdrage:
> > I rebased this against my manifest PR with the multiple children images and unfortunately actions do not work?
> > I'm unsure if it's because of change to name instead of ID, but I'm unable to delete anything that's a children image.
> 
> I can't find anything in this change that would affect child actions, and I'm able to delete and do other actions against #7552 with or without this change... Is there an error message for you, or maybe an issue but unrelated to this PR?

Nope no error! I will LGTM this PR as the code looks okay and it's most likely to do with my implementation.

Then I'd be able to rebase and try and see what's causing it.

cdrage (approved):

PR #7612, opened 2024-06-13
chore: automatically cleanup resources created during the activate method

### LINKED ISSUE #6288: Objects assigned to the ExtensionLoader should be able to clean their resources on exit

Today, on exit, only extensions are properly stopped, (deactivated) else resources are destroyed as part of the deletion of the process.

Objects could dispose some optional resources/timers/etc.

Make the objects that should manage it implementing Disposable interface and then, dispose() all resource when shutting down Podman Desktop

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
if developer forgot to unregister it, it'll done automatically

fixes https://github.com/containers/podman-desktop/issues/6288

### How to test this PR?

you can check with the minikube extension (and first remove the minikube binary on the system)

without the fix: status bar item is not removed when stopping the extension
with the fix: status bar item is removed when stopping the extension

(check that https://github.com/containers/podman-desktop-extension-minikube/pull/49 is not applied/included in your minikube extension first)

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -75,6 +75,7 @@ import {
 import type { StatusBarRegistry } from './statusbar/statusbar-registry.js';
 import type { Telemetry } from './telemetry/telemetry.js';
 import type { TrayMenuRegistry } from './tray-menu-registry.js';
+import type { IDisposable } from './types/disposable.js';
 import { Disposable } from './types/disposable.js';
 import { TelemetryTrustedValue } from './types/telemetry.js';
 import { Uri } from './types/uri.js';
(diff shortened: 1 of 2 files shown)


lstocchi (approved):
LGTM

packages/main/src/plugin/extension-loader.ts
```diff
@@ -787,7 +787,9 @@ export class ExtensionLoader {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         thisArg?: any,
       ): containerDesktopAPI.Disposable {
-        return commandRegistry.registerCommand(command, callback, thisArg);
+        const registration = commandRegistry.registerCommand(command, callback, thisArg);
```
axel7083, line 790: We could have a utility method 

```js
function registerDisposable(disposable, disposables): void {
    disposables.push(disposable)
    return disposable;
}
```

and do

`return registerDisposable(commandRegistry.registerCommand(...),  disposables)`

so we get one line everywhere instead of three

axel7083 (approved):
See comment, but LGTM, great improvement

PR #7627, opened 2024-06-13
fix: inform user when default connection is not in line with machine native option

### LINKED ISSUE #7052: Rootful podman machine gets stuck using rootless connection after another rootless machine is deleted

lstocchi:
### What does this PR do?

This PR adds a check from the desktop side to see if someone is running a machine using the wrong connection. Like the running machine is rootful but the default connection is rootless.

There is a known bug on podman https://github.com/containers/podman/issues/22993 that it happens when you delete a machine but it also may happen if the user mess up using the cli. 
This way we prompt a message to inform the user that they should update the connection

### Screenshot / video of UI

update_connection

It resolves https://github.com/containers/podman-desktop/issues/7052 and maybe https://github.com/containers/podman-desktop/issues/6455

### How to test this PR?

1. Create a rootful machine and start it
2. Create another rootful/rootless (it's the same) machine and start it
3. At this point, on Podman Desktop, you will be asked if you want to update the default connection -> yes
4. Stop the latest rootful/rootless machine -> on Podman Desktop you will be asked if you want to update the default connection -> yes -> the rootful connection of the first machine is set to default
5. delete the latest rootful/rootless machine.
6. if you check again you will see the default connection of the first machine has been updated to the rootless one
7. a message is prompted to update the connection

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -56,6 +56,7 @@ let autoMachineName: string | undefined;
 
 // System default notifier
 let defaultMachineNotify = !isLinux();
+let defaultConnectionNotify = !isLinux();
 let defaultMachineMonitor = true;
 
 // current status of machines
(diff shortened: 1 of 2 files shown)


extensions/podman/src/extension.ts
```diff
@@ -367,6 +368,39 @@ export async function checkDefaultMachine(machines: MachineJSON[]): Promise<void
     }
   }
 
+  // check if connection is in sync with machine. If the default connection is rootless but the machine is rootful ask the user to update the connection
+  if (defaultConnectionNotify && !!runningMachine?.Default) {
+    const defaultConnection = await getDefaultConnection();
+    const isRootful = await isRootfulMachine(runningMachine.Name);
+    if (!defaultConnection?.Name.endsWith(ROOTFUL_SUFFIX) && isRootful) {
```
jeffmaury, line 375: I don't get it you are comparing the default connection with a non default machine ?

lstocchi: In the if above i check that the running machine is the default one

jeffmaury (approved):
LGTM

lstocchi:
Rebasing as i saw one gh workflow didn't trigger

lstocchi:
@benoitf are you good with this change?

benoitf (approved):
@lstocchi yes

PR #7637, opened 2024-06-13
chore: update Podman pods summary page UI

### LINKED ISSUE #7559: Pods: Summary UI update

See: https://github.com/containers/podman-desktop/issues/7558

Update the summary page to match rest of the UI.

Implemented for Pods

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?

Changes the UI of the summary page of Podman pods to match other summary pages

### Screenshot / video of UI

Screenshot from 2024-06-13 10-10-03

Closes https://github.com/containers/podman-desktop/issues/7559

### How to test this PR?

View the summary page of a Podman pod

### DIFF
--- a/packages/renderer/src/lib/pod/PodmanPodDetailsSummary.spec.ts
+++ b/packages/renderer/src/lib/pod/PodmanPodDetailsSummary.spec.ts
@@ -1,0 +1,63 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+
+import { render, screen } from '@testing-library/svelte';
+import { expect, test } from 'vitest';
+
+import type { PodInfoUI } from './PodInfoUI';
+import PodmanPodDetailsSummary from './PodmanPodDetailsSummary.svelte';
+
+const fakePod: PodInfoUI = {
+  id: 'fakePodId',
+  shortId: 'FPI',
+  name: 'pod1',
+  engineId: 'fakeEngineId',
+  engineName: 'fakeEngineName',
+  status: 'RUNNING',
+  age: '3 days',
+  created: '2021-01-01T00:00:00Z',
+  selected: false,
+  containers: [
+    {
+      Id: 'fakeCId1',
+      Names: 'fakeContainer1',
+      Status: 'running',
+    },
+    {
+      Id: 'fakeCId2',
+      Names: 'fakeContainer2',
+      Status: 'running',
+    },
+  ],
+  kind: 'podman',
+};
+
+// Test render PodmanPodDetailsSummary with the PodInfoUi object
+test('PodmanPodDetailsSummary renders with PodInfoUi object', async () => {
+  // Render
+  render(PodmanPodDetailsSummary, { pod: fakePod });
+
+  // Check that the rendered text is correct
+  expect(screen.getByText('pod1')).toBeInTheDocument();
+  expect(screen.getByText('3 days')).toBeInTheDocument();
+  expect(screen.getByText('fakeContainer1')).toBeInTheDocument();
+  expect(screen.getByText('fakeContainer2')).toBeInTheDocument();
+  expect(screen.getAllByText('running')[0]).toBeInTheDocument();
+});
--- a/packages/renderer/src/lib/pod/PodmanPodDetailsSummary.spec.ts
+++ b/packages/renderer/src/lib/pod/PodmanPodDetailsSummary.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -49,8 +49,8 @@ const fakePod: PodInfoUI = {
   kind: 'podman',
 };
 
-// Test render PodmanPodDetailsSummary with the PodInfoUi object
-test('PodmanPodDetailsSummary renders with PodInfoUi object', async () => {
+// Test render PodmanPodDetailsSummary with the PodInfoUI object
+test('PodmanPodDetailsSummary renders with PodInfoUI object', async () => {
   // Render
   render(PodmanPodDetailsSummary, { pod: fakePod });
 
--- a/packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
+++ b/packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
@@ -8,67 +8,68 @@ import DetailsTable from '../details/DetailsTable.svelte';
 import DetailsTitle from '../details/DetailsTitle.svelte';
 import type { PodInfoUI } from './PodInfoUI';
 
-export let pod: PodInfoUI;
-let creationTime: Date = new Date(pod.created);
+export let pod: PodInfoUI | undefined;
+let creationTime: Date;
+if (pod) {
+  creationTime = new Date(pod.created);
+}
 </script>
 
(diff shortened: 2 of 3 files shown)


packages/renderer/src/lib/pod/PodmanPodDetailsSummary.spec.ts
```diff
@@ -0,0 +1,63 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
deboer-tim: Correct year :)

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+      </tr>
+      {#if pod}
+        <tr>
+          <DetailsTitle>Pod Status</DetailsTitle>
+        </tr>
+        <tr>
+          <DetailsCell>Status</DetailsCell>
+          <DetailsCell>{pod.status.toLowerCase()}</DetailsCell>
+        </tr>
+        {#if pod.containers.some(containerStatus => containerStatus.Status)}
```
deboer-tim: If the pod has no containers, wouldn't we still want to show the Containers title? (maybe wth a helpful 'no containers'?)

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+      </tr>
+      <tr>
+        <DetailsCell>Created</DetailsCell>
+        <DetailsCell>{creationTime}</DetailsCell>
+      </tr>
+      <tr>
+        <DetailsCell>Age</DetailsCell>
+        <DetailsCell>{pod.age}</DetailsCell>
+      </tr>
+      {#if pod}
```
axel7083: You seems to already be in a {#if pod} (from line 16), you can probably remove this one

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.spec.ts
cdrage: 2024

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+        <tr>
+          <DetailsCell>Status</DetailsCell>
+          <DetailsCell>{pod.status.toLowerCase()}</DetailsCell>
+        </tr>
+        {#if pod.containers.some(containerStatus => containerStatus.Status)}
+          <tr>
+            <DetailsTitle>Containers</DetailsTitle>
+          </tr>
+          {#each pod.containers as containerStatus}
+            {#if containerStatus.Status}
```
deboer-tim: Seems like an odd check - doesn't the status always exist? (and if not, wouldn't we still want to show the container exists/its name?

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+              </tr>
+              <tr>
+                <DetailsCell>Status</DetailsCell>
+                <DetailsCell>{containerStatus.Status}</DetailsCell>
+              </tr>
+            {/if}
+          {/each}
+        {/if}
+      {/if}
+    </DetailsTable>
```
cdrage: Should be at end of file, see similar to: https://github.com/containers/podman-desktop/blob/c43ae821cf989e544ecf51ef2223d007fea57a1b/packages/renderer/src/lib/image/ImageDetailsSummary.svelte#L30

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+  {:else}
+    <p class="text-purple-500 font-medium">Loading ...</p>
+  {/if}
+</div>
```
cdrage: remove div

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+import { Link } from '@podman-desktop/ui-svelte';
+import { router } from 'tinro';
+
+import DetailsCell from '../details/DetailsCell.svelte';
+import DetailsSubtitle from '../details/DetailsSubtitle.svelte';
+import DetailsTable from '../details/DetailsTable.svelte';
+import DetailsTitle from '../details/DetailsTitle.svelte';
+import type { PodInfoUI } from './PodInfoUI';
+
+export let pod: PodInfoUI;
```
axel7083: The typing might be wrong here, as you are checking bellow `{#if pod}` we may prefer to clarify it and express that this component handle the case where pod is undefined.

```suggestion
export let pod: PodInfoUI | undefined;
```

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+        </tr>
+        <tr>
+          <DetailsCell>Status</DetailsCell>
+          <DetailsCell>{pod.status.toLowerCase()}</DetailsCell>
+        </tr>
+        {#if pod.containers.some(containerStatus => containerStatus.Status)}
+          <tr>
+            <DetailsTitle>Containers</DetailsTitle>
+          </tr>
+          {#each pod.containers as containerStatus}
```
deboer-tim: Can we rename `containerStatus` to maybe `container` or `containerUI`? Had to take a second look below b/c I wasn't sure it was referring to the container.

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+import DetailsSubtitle from '../details/DetailsSubtitle.svelte';
+import DetailsTable from '../details/DetailsTable.svelte';
+import DetailsTitle from '../details/DetailsTitle.svelte';
+import type { PodInfoUI } from './PodInfoUI';
+
+export let pod: PodInfoUI;
+let creationTime: Date = new Date(pod.created);
+</script>
+
+<div class="flex px-5 py-4 flex-col items-start h-full overflow-auto">
```
cdrage: remove div

axel7083 (commented):
Great improvement for the Pod summary page !

Here is a side by side comparison

| before | After |
| --- | --- |
| image  | image |

I added some comments to your code !

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -0,0 +1,74 @@
...
+  {#if pod}
```
cdrage, line 19: if statement within <DetailsTable> similar to https://github.com/containers/podman-desktop/pull/7610/files#diff-0b58c8cb00282f3510a707967a7f5350c86037a07bd74d857b869604ad06622aR22

cdrage (changes requested):

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
SoniaSandler: I moved the title outside the if, and changed the if statement to check pod.containers.length > 0

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
SoniaSandler: I double checked, and the status of a container always exists (according to the `PodInfoContainerUI` type), so I removed the if

jeffmaury (commented):
Seems to me elements are missing for containers:

- ports
- image
- env variables

SoniaSandler:
@jeffmaury iirc, Podman pods use the `PodInfoUI` interface, so that is where I got the summary page info from. In the `PodInfoUI`, there is an array of `PodInfoContainerUI` for the containers of the pod, and this interface only includes name, id, and status. However, the name of the container in this summary page is also a link to the container log page (I can change it to the container summary page if it makes more sense), but if needed, I can try and find a way to include more info about the containers in this summary page

jeffmaury:
> @jeffmaury iirc, Podman pods use the `PodInfoUI` interface, so that is where I got the summary page info from. In the `PodInfoUI`, there is an array of `PodInfoContainerUI` for the containers of the pod, and this interface only includes name, id, and status. However, the name of the container in this summary page is also a link to the container log page (I can change it to the container summary page if it makes more sense), but if needed, I can try and find a way to include more info about the containers in this summary page

Ok I got it. I would arg that the image name shoud be presented here, the other can be access through a link to the container

SoniaSandler:
@jeffmaury I'm also working on the container summary page right now, and it will also include the image name

slemeur:
Nice job @SoniaSandler

cdrage (approved):

deboer-tim (approved):
I agree with @jeffmaury that having the container image here would be really useful to tell the containers apart, but since that's not a regression I'm approving. Nice work @SoniaSandler.

Issue #7445, opened 2024-06-04
Checkbox element alignment in deploy to Kubernetes Page is missing padding


odockal:
### Bug description

There seems to be missing a padding with checkbox elements in Deploy to Kubernetes Page.

![Screenshot_20240604_085753](https://github.com/containers/podman-desktop/assets/19164299/ad793786-dbd2-4898-b74a-b85dbc7a6c59)

### Operating system

Fedora 40

### Installation Method

Installer from website/GitHub releases

### Version

1.10.2

PR #7644, opened 2024-06-13
chore: improve checkbox component

### LINKED ISSUE #7609: Odd behaviour using Checkbox component

I tried to use the Checkbox component in the BootC extension (https://github.com/containers/podman-desktop-extension-bootc/issues/537) and found some annoyances:
- The checkbox overflows its own area, so if you put something directly beside it they overlap.
- It's a common pattern to have text beside the checkbox, but you need to create that flow layout for each line. (and unless you know to add some padding, the text is partially hidden behind the checkbox)
- Also noticed that the native checkbox is causing odd issues on mouseover.

https://github.com/containers/podman-desktop/assets/19958075/b3f6c61c-915f-4b08-a9c7-c1761676e20e

Fix these to make the Checkbox easier to reuse.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Fixes to make Checkbox behave better and be easier to use as a component:

- Increase padding from 1 to 2 so that the checkbox fits within its bounds.
- Adds an extra div with flex-row and a slot so that you can easily place text beside the Checkbox. When the slot is filled we add a little margin to the right and top (so that it aligns better with text in the slot). (We should maybe also set the default text size and color here? Leaving that for now)
- Hover issues are due to the native checkbox. We can't remove or hide it completely - but we can move it to be a 1px box at the top left and match the pointer.

These changes are applied to run image and deploy to pod forms to show the simplified way to add text without the wrapper divs.

### Screenshot / video of UI

Before:

Screenshot 2024-06-13 at 3 37 58 PM

Screenshot 2024-06-13 at 3 41 33 PM

After:

Screenshot 2024-06-13 at 3 39 58 PM

Screenshot 2024-06-13 at 3 40 31 PM

Fixes #7609.

### How to test this PR?

Check out the two forms, and no regression to checkboxes in other places like the tables.

### DIFF
--- a/packages/renderer/src/lib/image/RunImage.svelte
+++ b/packages/renderer/src/lib/image/RunImage.svelte
@@ -787,13 +787,11 @@ async function assertAllPortAreValid(): Promise<void> {
               <div class="h-96 overflow-y-auto pr-4">
                 <!-- Use tty -->
                 <label for="containerTty" class="block mb-2 text-sm font-medium text-gray-400">Use TTY:</label>
-                <div class="flex flex-row justify-start items-center align-middle w-full text-gray-700 text-sm">
-                  <Checkbox bind:checked="{useTty}" class="mx-2" title="Attach a pseudo terminal" />
-                  Attach a pseudo terminal
-                </div>
-                <div class="flex flex-row justify-start items-center align-middle w-full text-gray-700 text-sm">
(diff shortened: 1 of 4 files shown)


axel7083 (approved):

cdrage (approved):
Looks good! Thank you for the update.

odockal:
Also fixes #7445. Thanks @deboer-tim !

PR #7645, opened 2024-06-13
fix: handle hidden property for Button ui component

### LINKED ISSUE #7252: Sections disappear when pressing "- / remove" in Create Container
### Bug description

https://github.com/containers/podman-desktop/assets/6422176/0ca64113-83e5-4909-b96f-5e1d90420323

When pressing any of the minuses / - / remove buttons, the section disappears gone forever until you refresh the page. 

### Operating system

mac m1

### Installation Method

None

### Version

1.10.2

### Steps to reproduce

Press any of the "-" buttons on create container page.

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
while looking at the bug https://github.com/containers/podman-desktop/issues/7252 it turns out that the minus and plus icons should not be visible at the same time as they're using the hidden property to hide/show it
but hidden property was not handled by the Button so this is why it was displayed

Fixing by adding the property hidden in the Button component

fixes https://github.com/containers/podman-desktop/issues/7252

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/button/Button.spec.ts
+++ b/packages/ui/src/lib/button/Button.spec.ts
@@ -191,3 +191,9 @@ test('Button no progress no icon do not have spinner', async () => {
   const svg = screen.queryByRole('img');
   expect(svg).toBeNull();
 });
+
+test('Button hidden should be hidden', async () => {
+  render(Button, { hidden: true });
+  const button = screen.queryByRole('button');
+  expect(button).not.toBeInTheDocument();
+});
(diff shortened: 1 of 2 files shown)


deboer-tim (approved):
Doh, this was removed because it was 'unused' (#7092). I'm good with adding the capability back.

jeffmaury (approved):
LGTM

benoitf:
cc @axel7083

axel7083:
> cc @axel7083

I am okey with adding it back, even tho I would have a preference for parent component to use class:hidden={true} but I do not have that much a strong opinion on it, so go for it !

axel7083 (approved):

benoitf:
> I am okey with adding it back, even tho I would have a preference for parent component to use class:hidden={true} but I do not have that much a strong opinion on it, so go for it !

but hidden is something native on widgets `<button hidden>` is working so I don't see why it should be a class while it's an attribute on HTML components

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden#description

PR #7657, opened 2024-06-14
Check only current kubecontext when there are more than 10 contexts in config

### LINKED ISSUE #7409: Extensions Catalog fails to render when local kube.config is excessively large
### Bug description

My kube.config file had around 2000 config endpoints in it (I had forgotten to clean it). Podman Desktop starts up OK but then will not render any entries in the extensions catalog. Also the Troubleshoot/Log panel took 10 mins+ to render the log (which was full of non-responses from the kube.config endpoints)

### Operating system

MacOS 14.5

### Installation Method

### Version

1.10.2

### Steps to reproduce

Generate a 'config' file in ~/.kube with 1000+ entries, start Podman Desktop, Click on Extensions, choose Catalog

### Relevant log output

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

Check only current kubecontext when there are more than 10 contexts in config

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/9973512/262ff89a-a4d7-45ae-9cb2-247b80d8cb67

Fixes #7409 

### How to test this PR?

Add contexts in your kubeconfig and check that only the current one displays ccontext information in the Kubernetes Contexts page

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -60,6 +60,10 @@ import {
   dispatchTimeout,
 } from './kubernetes-context-state-constants.js';
 
+// If the number of contexts in the kubeconfig file is greater than this number,
+// only the connectivity to the current context will be checked
+const MAX_NON_CURRENT_CONTEXTS_TO_CHECK = 10;
+
 // ContextInternalState stores informers for a kube context
 type ContextInternalState = Map<ResourceName, Informer<KubernetesObject> & ObjectCache<KubernetesObject>>;
 
@@ -432,7 +436,7 @@ export class ContextsManager {
   }
 
   async update(kubeconfig: KubeConfig): Promise<void> {
-    const checkOnlyCurrentContext = kubeconfig.contexts.length > 10;
+    const checkOnlyCurrentContext = kubeconfig.contexts.length > MAX_NON_CURRENT_CONTEXTS_TO_CHECK;
 
     this.kubeConfig = kubeconfig;
     let contextChanged = false;
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -431,9 +431,9 @@ export class ContextsManager {
     );
   }
 
-  // update is the reconcile function, it gets as input the last known kube config
-  // and starts/stops informers for different kube contexts, depending on these inputs
   async update(kubeconfig: KubeConfig): Promise<void> {
+    const checkOnlyCurrentContext = kubeconfig.contexts.length > 10;
+
     this.kubeConfig = kubeconfig;
     let contextChanged = false;
     if (this.isContextChanged(kubeconfig)) {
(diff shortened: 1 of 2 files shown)


deboer-tim (approved):
I think we could get much more clever in the future, e.g. not disconnect immediately or keep the last 10 with the minimal informer, but this fixes the current problem and behaved exactly as described in testing.

IMHO the UI is a bit odd though, with the appearing/disappearing sections on the left. Especially if you're coming from the <10 behaviour it's not clear if we are trying to connect or not. I'd prefer if the left section was always there but empty counts + different status. What do you think? We can open separate issue to track.

feloy:
> I think we could get much more clever in the future, e.g. not disconnect immediately or keep the last 10 with the minimal informer, but this fixes the current problem and behaved exactly as described in testing.
> 
> IMHO the UI is a bit odd though, with the appearing/disappearing sections on the left. Especially if you're coming from the <10 behaviour it's not clear if we are trying to connect or not. I'd prefer if the left section was always there but empty counts + different status. What do you think? We can open separate issue to track.

Yes, I agree that the section not visible on the left is disappointing.

I've some code ready for displaying the check status: the backend sends to the frontend when it is trying to connect for each context, and a spinner is visible the time the informer is trying to connect (visible ~ 0.5s if connection fails quickly). I would like to improve this code for the backend to declare that it does not do anything, and in this case the frontend would display a 'check' button in the left section, for the user to trigger manually a connection attempt. WDYT?

deboer-tim:
> > I think we could get much more clever in the future, e.g. not disconnect immediately or keep the last 10 with the minimal informer, but this fixes the current problem and behaved exactly as described in testing.
> > IMHO the UI is a bit odd though, with the appearing/disappearing sections on the left. Especially if you're coming from the <10 behaviour it's not clear if we are trying to connect or not. I'd prefer if the left section was always there but empty counts + different status. What do you think? We can open separate issue to track.
> 
> Yes, I agree that the section not visible on the left is disappointing.
> 
> I've some code ready for displaying the check status: the backend sends to the frontend when it is trying to connect for each context, and a spinner is visible the time the informer is trying to connect (visible ~ 0.5s if connection fails quickly). I would like to improve this code for the backend to declare that it does not do anything, and in this case the frontend would display a 'check' button in the left section, for the user to trigger manually a connection attempt. WDYT?

+1 and thanks for opening #7661. I will review the PR, but that sounds good - we just need a spinner each time it tries to connect.

packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -403,9 +403,9 @@ export class ContextsManager {
     );
   }
 
-  // update is the reconcile function, it gets as input the last known kube config
-  // and starts/stops informers for different kube contexts, depending on these inputs
   async update(kubeconfig: KubeConfig): Promise<void> {
+    const checkOnlyCurrentContext = kubeconfig.contexts.length > 10;
```
cdrage: IMHO the `10` should be a const at the top  of this file so we can change this in the future if it's an issue?

cdrage (approved):
LGTM and tested! Works really well, one tiny fix / suggestion, but it's something small.

cdrage (approved):

PR #7658, opened 2024-06-14
feat(context): create when svelte component

### LINKED ISSUE #6898: Implement ListItemButtonIcon component

axel7083:
### What does this PR do?

Creating a `<When>` svelte component taking a `when` expression and contextUI object. It will be responsible of providing the result of the expression evaluation based on context provided and global context.

Example usage

```svelte
const context = new ContextUI();
context.setValue('fruits', ['banana', 'pineapple']);
...
<When contextUI={context} expression="banana in fruits" let:result>
     {#if result} .... {/if}
</When>
```

It not directly possible to test slot[^1], therefore I created a wrapper around the component to be able to read the result, and test it properly

[^1]: https://github.com/testing-library/svelte-testing-library/issues/48

Related to https://github.com/containers/podman-desktop/issues/6898

### How to test this PR?

**Manually**

Change the kind push to kubernetes to `true` 

https://github.com/containers/podman-desktop/blob/e41ffd97589ecfa070653a58ebe4503d2f573f5f/extensions/kind/package.json#L74

And you should get the following result in the image page

image

### DIFF
--- a/packages/renderer/src/lib/actions/ContributionActions.svelte
+++ b/packages/renderer/src/lib/actions/ContributionActions.svelte
@@ -8,6 +8,7 @@ import { removeNonSerializableProperties } from '/@/lib/actions/ActionUtils';
 import type { ContextUI } from '/@/lib/context/context';
 import { ContextKeyExpr } from '/@/lib/context/contextKey';
 import { transformObjectToContext } from '/@/lib/context/ContextUtils';
+import When from '/@/lib/context/When.svelte';
 import { context as storeContext } from '/@/stores/context';
 
 import type { Menu } from '../../../../main/src/plugin/menu-registry';
@@ -98,12 +99,13 @@ async function executeContribution(menu: Menu): Promise<void> {
 </script>
 
 {#each filteredContributions as menu}
-  <ListItemButtonIcon
-    title="{menu.title}"
-    onClick="{() => executeContribution(menu)}"
-    menu="{dropdownMenu}"
-    icon="{getIcon(menu)}"
-    detailed="{detailed}"
-    disabledWhen="{menu.disabled}"
-    contextUI="{globalContext}" />
+  <When expression="{menu.disabled}" contextUI="{globalContext}" let:enabled>
+    <ListItemButtonIcon
+      title="{menu.title}"
+      onClick="{() => executeContribution(menu)}"
+      menu="{dropdownMenu}"
+      icon="{getIcon(menu)}"
+      detailed="{detailed}"
+      enabled="{!enabled}" />
+  </When>
 {/each}
--- a/packages/renderer/src/lib/actions/ContributionActions.svelte
+++ b/packages/renderer/src/lib/actions/ContributionActions.svelte
@@ -99,13 +99,13 @@ async function executeContribution(menu: Menu): Promise<void> {
 </script>
 
 {#each filteredContributions as menu}
-  <When expression="{menu.disabled}" contextUI="{globalContext}" let:enabled>
+  <When expression="{menu.disabled}" contextUI="{globalContext}" let:result>
     <ListItemButtonIcon
       title="{menu.title}"
       onClick="{() => executeContribution(menu)}"
       menu="{dropdownMenu}"
       icon="{getIcon(menu)}"
       detailed="{detailed}"
-      enabled="{!enabled}" />
+      enabled="{!result}" />
   </When>
 {/each}
--- a/packages/renderer/src/lib/context/When.spec.ts
+++ b/packages/renderer/src/lib/context/When.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/renderer/src/lib/context/When.spec.ts
+++ b/packages/renderer/src/lib/context/When.spec.ts
@@ -17,7 +17,7 @@
  ***********************************************************************/
 
 import { render, screen } from '@testing-library/svelte';
-import { beforeEach,expect, test, vi } from 'vitest';
+import { beforeEach, expect, test, vi } from 'vitest';
 
 import { ContextUI } from '/@/lib/context/context';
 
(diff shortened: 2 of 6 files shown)


packages/renderer/src/lib/context/When.spec.ts
```diff
@@ -0,0 +1,89 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
axel7083:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/renderer/src/lib/actions/ContributionActions.svelte
```diff
@@ -98,12 +99,13 @@ async function executeContribution(menu: Menu): Promise<void> {
...
 {#each filteredContributions as menu}
-  <ListItemButtonIcon
-    title="{menu.title}"
-    onClick="{() => executeContribution(menu)}"
-    menu="{dropdownMenu}"
-    icon="{getIcon(menu)}"
-    detailed="{detailed}"
-    disabledWhen="{menu.disabled}"
-    contextUI="{globalContext}" />
+  <When expression="{menu.disabled}" contextUI="{globalContext}" let:enabled>
```
feloy: Isn't the expression the opposite? `!menu.disabled` 
This way, you could use `enabled="{enabled}" in the `ListItemButtonIcon` which is meaningful

feloy (changes requested):

packages/renderer/src/lib/actions/ContributionActions.svelte
axel7083: You cannot use `!menu.disabled` because it is a string. I am negating the result after

https://github.com/containers/podman-desktop/blob/41afb829ec3875a3d68c73614a8a3317315f2c7a/packages/renderer/src/lib/actions/ContributionActions.svelte#L109

Here the way we store the expression is `disabled`, which is a negation, I would rather output from the When component `enabled` (evaluation result of the expression) than its negation, which IMO would not make sense.

So at the end I am getting the result of the expression through the enabled properly, which I agree could probably be replaced by a better wording, `value` ? `result` ? but the When component does not care what you do with the value, it take an expression and evaluate it, and provide the result through a value.

But I am not outputting the negation of the result of the evaluation of the expression. This is not the responsibility of the When component here. Otherwise we rename the component NotWhen

feloy: ok, I didn't get that `menu.disabled` was the formula, and not a boolean. But my concern remains, that we have to use `enabled="{!enabled}"` at the end, which is quite disturbing. 

Could it be possible to let to the user of the When component the choice of the name for the `let:` attribute, so the user would choose the more appriopriate one - here, `let:disabled` for example

axel7083: > Could it be possible to let to the user of the When component the choice of the name for the let: attribute

Trying to do that is probably not something I want to do, as it would definitely be some ugly js code to achieve this specific mechanism.

Moreover it would be very difficult to ensure type safety.

Therefore, I would be against that proposal. But I think it might be interesting to rename `let:enabled` to something different, to remove some potential confusion as you raised.

feloy: yeah, if Svelte is not providing some simple mechanism, it would be better to have a generic name instead

axel7083: I renamed to `result`

feloy (approved):

jeffmaury (commented):
Looks strange to me to have a component just for the sake of testing, defeats a little bit the purpose

axel7083:
> Looks strange to me to have a component just for the sake of testing, defeats a little bit the purpose

@jeffmaury that's the recommended way for slot testing in svelte;

We are doing it for the table

https://github.com/containers/podman-desktop/blob/b6e040711fb312737a5bb8d16b5f9693bcb67489/packages/ui/src/lib/table/TestTable.svelte#L1

And also the Form

https://github.com/containers/podman-desktop/blob/f25dc1befb7a87a5ae688d60c58f4c70819c93ad/packages/ui/src/lib/layouts/FormPageSpec.svelte#L1

You can find this method recommended in their documentation

> Slots are more difficult to test as there is no programmatic interface for working with them either inside or outside of a Svelte component. The simplest way is to create a test specific component that utilises a dynamic component to and passes in a default slot to that component which can be asserted against.[^1]


[^1]: https://www.sveltesociety.dev/recipes/testing-and-debugging/unit-testing-svelte-component

axel7083:
cc @benoitf @deboer-tim

benoitf:
I don't see the value of this When component vs a direct usage to a if clause ?

packages/renderer/src/lib/actions/ContributionActions.svelte
```diff
@@ -98,12 +99,13 @@ async function executeContribution(menu: Menu): Promise<void> {
...
-  <ListItemButtonIcon
-    title="{menu.title}"
-    onClick="{() => executeContribution(menu)}"
-    menu="{dropdownMenu}"
-    icon="{getIcon(menu)}"
-    detailed="{detailed}"
-    disabledWhen="{menu.disabled}"
-    contextUI="{globalContext}" />
+  <When expression="{menu.disabled}" contextUI="{globalContext}" let:result>
+    <ListItemButtonIcon
```
benoitf, line 103: I have the feeling that we should not use a UI component for setting a value

I would expect
```
{@const isEnabled = ..... }
<ListItemButtonIcon
  enabled={isEnabled}
```

axel7083:
> I don't see the value of this When component vs a direct usage to a if clause ?

This component manage the subscribing and unsubscribing to the global context, which takes at least 50 lines in each component. Yes we could have a if everywhere, but we would also have to add _a lot_ of repeating code in each component using it.

This component aims to ease the usage of contexts.

packages/renderer/src/lib/actions/ContributionActions.svelte
axel7083: https://github.com/containers/podman-desktop/pull/7658#issuecomment-2188297551

benoitf (changes requested):
I still believe it is an anti-pattern being used.

axel7083:
Closing as the solution has been rejected

> Tinro is doing the same, I don't agree that it is an anti pattern.

PR #7659, opened 2024-06-14
fix: support multiple scopes on create/edit configurations

lstocchi:
### What does this PR do?

This PR fixes a couple of issues related to the configuration values I discovered while working on the crc extension.

The first is that we just checked that the property scope was equal to a value `property.scope === propertyScope` but scope can also be an array. In that case we should check if the array contains that scope.

The second is that the inner map we use to retrieve a value (used for memory/cpus/disksize) was only filled when working with a connection. (e.g. podman-machine -> cpus 2, memory 3, disk 10) . The problem is that on crc the cpus,memory and disksize are not bound to any connection bc crc does not create new machines. There is just one VM per time. So the values associated to these properties are generic.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.spec.ts
@@ -399,3 +399,66 @@ test(`Expect create with unchecked and checked checkboxes`, async () => {
     undefined,
   );
 });
+
+test(`Expect create with unchecked and checked checkboxes having multipl scopes`, async () => {
+  let providedKeyLogger: ((key: symbol, eventName: LoggerEventName, args: string[]) => void) | undefined;
+  const taskId = 4;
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.spec.ts
```diff
@@ -399,3 +399,66 @@ test(`Expect create with unchecked and checked checkboxes`, async () => {
     undefined,
   );
 });
+
+test(`Expect create with unchecked and checked checkboxes having multipl scopes`, async () => {
```
jeffmaury:
```suggestion
test(`Expect create with unchecked and checked checkboxes having multiple scopes`, async () => {
```

jeffmaury (approved):
LGTM

axel7083 (approved):

PR #7677, opened 2024-06-17
fix: Delete the removal of sha256: prefix in imageHref

### LINKED ISSUE #6801: bug(ImageDetails): navigation to container image crashes Podman Desktop
### Bug description

https://github.com/containers/podman-desktop/assets/42176370/997bcbc4-f28c-45d5-a91c-2c3329496e3d

After that the application became completely crashed and user cannot perform any action except restarting Podman Desktop.

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

image

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
Deletes the removal of `sha256:` prefix in the `imageHref` property of the `ContainerInfoUI` interface 
(the `sha256:` prefix stays in `imageHref`)
Fixes https://github.com/containers/podman-desktop/issues/6801

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/container/container-utils.ts
+++ b/packages/renderer/src/lib/container/container-utils.ts
@@ -139,15 +139,14 @@ export class ContainerUtils {
   }
 
   getImageHref(containerInfo: ContainerInfo): string {
-    if (containerInfo.ImageBase64RepoTag) {
-      const imageTag = Buffer.from(containerInfo.ImageBase64RepoTag, 'base64').toString();
-      if (imageTag.startsWith('sha256:')) {
-        return `/images/${containerInfo.ImageID}/${containerInfo.engineId}`;
-      } else {
-        return `/images/${containerInfo.ImageID}/${containerInfo.engineId}/${containerInfo.ImageBase64RepoTag}/summary`;
-      }
+    const repoTag = containerInfo.ImageBase64RepoTag
+      ? Buffer.from(containerInfo.ImageBase64RepoTag, 'base64').toString()
+      : '';
+    const shortUrl = `/images/${containerInfo.ImageID}/${containerInfo.engineId}`;
+    if (repoTag.startsWith('sha256:') || repoTag === '') {
+      return shortUrl;
     } else {
-      return `/images/${containerInfo.ImageID}/${containerInfo.engineId}`;
+      return `${shortUrl}/${containerInfo.ImageBase64RepoTag}/summary`;
     }
   }
 
--- a/packages/renderer/src/lib/container/container-utils.ts
+++ b/packages/renderer/src/lib/container/container-utils.ts
@@ -165,7 +165,7 @@ export class ContainerUtils {
       labels: containerInfo.Labels,
       icon: this.iconClass(containerInfo, context, viewContributions) ?? ContainerIcon,
       imageBase64RepoTag: containerInfo.ImageBase64RepoTag,
-      imageHref: `/images/${containerInfo.ImageID.startsWith('sha256:') ? containerInfo.ImageID.slice(7) : containerInfo.ImageID}/${containerInfo.engineId}/${containerInfo.ImageBase64RepoTag}/summary`,
+      imageHref: `/images/${containerInfo.ImageID}/${containerInfo.engineId}/${containerInfo.ImageBase64RepoTag}/summary`,
     };
   }
 
--- a/packages/renderer/src/lib/container/container-utils.ts
+++ b/packages/renderer/src/lib/container/container-utils.ts
@@ -18,6 +18,8 @@
 
 import type { Port } from '@podman-desktop/api';
 import { ContainerIcon } from '@podman-desktop/ui-svelte/icons';
+// eslint-disable-next-line unicorn/prefer-node-protocol
+import { Buffer } from 'buffer';
 import { filesize } from 'filesize';
 import humanizeDuration from 'humanize-duration';
 import moment from 'moment';
--- a/packages/renderer/src/lib/container/container-utils.ts
+++ b/packages/renderer/src/lib/container/container-utils.ts
@@ -136,6 +136,19 @@ export class ContainerUtils {
     return containerInfo.engineName;
   }
 
+  getImageHref(containerInfo: ContainerInfo): string {
+    if (containerInfo.ImageBase64RepoTag) {
+      const imageTag = Buffer.from(containerInfo.ImageBase64RepoTag, 'base64').toString();
+      if (imageTag.startsWith('sha256:')) {
+        return `/images/${containerInfo.ImageID}/${containerInfo.engineId}`;
+      } else {
+        return `/images/${containerInfo.ImageID}/${containerInfo.engineId}/${containerInfo.ImageBase64RepoTag}/summary`;
+      }
+    } else {
+      return `/images/${containerInfo.ImageID}/${containerInfo.engineId}`;
+    }
+  }
+
   getContainerInfoUI(
     containerInfo: ContainerInfo,
     context?: ContextUI,
@@ -165,7 +178,7 @@ export class ContainerUtils {
       labels: containerInfo.Labels,
       icon: this.iconClass(containerInfo, context, viewContributions) ?? ContainerIcon,
       imageBase64RepoTag: containerInfo.ImageBase64RepoTag,
-      imageHref: `/images/${containerInfo.ImageID}/${containerInfo.engineId}/${containerInfo.ImageBase64RepoTag}/summary`,
+      imageHref: this.getImageHref(containerInfo),
     };
   }
 
--- a/packages/renderer/src/lib/container/container-utils.spec.ts
+++ b/packages/renderer/src/lib/container/container-utils.spec.ts
@@ -320,7 +320,7 @@ test('check parsing of container info without labels', async () => {
   containerUtils.adaptContextOnContainer(context, containerInfo);
 });
 
-test('should expect imageHref not to have sha256: prefix', async () => {
+test('should expect imageHref to use exact image id', async () => {
   const containerInfo = {
     Id: 'container1',
     Image: 'docker.io/kindest/node:foobar',
@@ -331,21 +331,7 @@ test('should expect imageHref not to have sha256: prefix', async () => {
     ImageBase64RepoTag: 'dummy-base-64',
   } as unknown as ContainerInfo;
   const containerUI = containerUtils.getContainerInfoUI(containerInfo);
-  expect(containerUI.imageHref).toBe('/images/dummy-sha256/dummy-engine-id/dummy-base-64/summary');
-});
-
-test('should expect imageHref to use exact image id if no sha256: prefix', async () => {
-  const containerInfo = {
-    Id: 'container1',
-    Image: 'docker.io/kindest/node:foobar',
-    Names: ['container1'],
-    State: 'STOPPED',
-    ImageID: 'dummy-sha256',
-    engineId: 'dummy-engine-id',
-    ImageBase64RepoTag: 'dummy-base-64',
-  } as unknown as ContainerInfo;
-  const containerUI = containerUtils.getContainerInfoUI(containerInfo);
-  expect(containerUI.imageHref).toBe('/images/dummy-sha256/dummy-engine-id/dummy-base-64/summary');
+  expect(containerUI.imageHref).toBe('/images/sha256:dummy-sha256/dummy-engine-id/dummy-base-64/summary');
 });
 
 test('should be able to identify container groups', async () => {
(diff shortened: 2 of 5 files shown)


deboer-tim:
Maybe this would be a good time to start switching to the navigation API (#5556)?

SoniaSandler:
@deboer-tim I'll take a look at the navigation API

axel7083 (changes requested):
Globally this does not fix in certain scenario.

https://github.com/containers/podman-desktop/assets/42176370/af01ca2c-b3c7-417c-ab9f-34731db0d961

### Specific scenario

I am not really sure why, but in some cases it does not work, let's break the example down.

In my container list, I have a pod, and three container in it. One of the container, as visible in the video is called `llamacpp-server-podified`.

The link is the following reading from the `Inspect.image`: `sha256:462f6b668ede74b92d4823a309eda3033809dee5d3bd536f78308cd5afd7aaaf`

Let's go in the image list, and search this sha256

We do get the right image which is being used by the container

image

Now let's look at the url.

`http://localhost:5173/#/images/sha256:462f6b668ede74b92d4823a309eda3033809dee5d3bd536f78308cd5afd7aaaf/podman.Podman%20Machine/cXVheS5pby9haS1sYWIvbGxhbWFjcHBwX3B5dGhvbjpsYXRlc3Q=/summary`

Now let's compare with the blank page url.

`http://localhost:5173/#/images/sha256:462f6b668ede74b92d4823a309eda3033809dee5d3bd536f78308cd5afd7aaaf/podman.Podman%20Machine/c2hhMjU2OjQ2MmY2YjY2OGVkZTc0YjkyZDQ4MjNhMzA5ZWRhMzAzMzgwOWRlZTVkM2JkNTM2Zjc4MzA4Y2Q1YWZkN2FhYWY=/summary`

Let's diff

```diff
- http://localhost:5173/#/images/sha256:462f6b668ede74b92d4823a309eda3033809dee5d3bd536f78308cd5afd7aaaf/podman.Podman%20Machine/cXVheS5pby9haS1sYWIvbGxhbWFjcHBwX3B5dGhvbjpsYXRlc3Q=/summary
+ http://localhost:5173/#/images/sha256:462f6b668ede74b92d4823a309eda3033809dee5d3bd536f78308cd5afd7aaaf/podman.Podman%20Machine/c2hhMjU2OjQ2MmY2YjY2OGVkZTc0YjkyZDQ4MjNhMzA5ZWRhMzAzMzgwOWRlZTVkM2JkNTM2Zjc4MzA4Y2Q1YWZkN2FhYWY=/summary
```

The last part of the url id different, which is the `ImageBase64RepoTag`, weird... 

Let's decode it:
```
$: atob("cXVheS5pby9haS1sYWIvbGxhbWFjcHBwX3B5dGhvbjpsYXRlc3Q=")
'quay.io/ai-lab/llamacppp_python:latest'
```

```
$: atob("c2hhMjU2OjQ2MmY2YjY2OGVkZTc0YjkyZDQ4MjNhMzA5ZWRhMzAzMzgwOWRlZTVkM2JkNTM2Zjc4MzA4Y2Q1YWZkN2FhYWY=")
'sha256:462f6b668ede74b92d4823a309eda3033809dee5d3bd536f78308cd5afd7aaaf'
```

Okey okey, so the `ImageBase64RepoTag` is one case is the one of the sha256, and the other of the image tag. This is a problem.

cc @deboer-tim @benoitf the consequence is the following: in `image-utils.ts` we are trying to find an image based on the base64Tag provided

https://github.com/containers/podman-desktop/blob/61eefc784b4c191bb14242dc3ecb9cd905ca346c/packages/renderer/src/lib/image/image-utils.ts#L244-L245

But we construct the image href with a base64 value from the container info

https://github.com/containers/podman-desktop/blob/cb188d7d5b71bf295c7f8d02c1b5b5234d9eae00/packages/renderer/src/lib/container/container-utils.ts#L161

which seems to not exists in the images available.

## Solution proposal

We are generating a list of imageInfo for the ImageDetails 

https://github.com/containers/podman-desktop/blob/6b615e4cf1941451834b531a126a0640231e0bd8/packages/renderer/src/lib/image/ImageDetails.svelte#L75

Those image are generated by iterating over the `RepoTags: string[]` property like shown bellow

https://github.com/containers/podman-desktop/blob/61eefc784b4c191bb14242dc3ecb9cd905ca346c/packages/renderer/src/lib/image/image-utils.ts#L204

As we understood above, one condition that is not generated is the cases where the `base64RepoTag` is the sha256 itself. So we could add this possiblity and do the following

```diff
- return imageInfo.RepoTags.map(repoTag => {
+ return [...imageInfo.RepoTags, imageInfo.Id].map(repoTag => {
```

This solutions works on my computer **in addition** of  the change made by @SoniaSandler.

Moreover, to cover all cases, we might also generate one more imageInfo for the case where there are not RepoTags

https://github.com/containers/podman-desktop/blob/61eefc784b4c191bb14242dc3ecb9cd905ca346c/packages/renderer/src/lib/image/image-utils.ts#L181

deboer-tim:
Excellent digging @axel7083. I think your solution is good, but I also can't help feeling that we need to understand/define what the full unique identifier(s) are for an image, and this code should only exist with the navigation helper and either image details or utils. i.e. places like container utils should have no idea how to build the url for an image or what the unique properties are. Then we should also repeat for all other types. :)

SoniaSandler:
@deboer-tim @axel7083 I've added Axel's solution to the PR. However, I ran into some issues with AI-lab, so I wasn't able to get an image that might have the same problem of two different tags and check that the new addition does solve this problem completely. The images that I have work with the original change (just adding back the `sha256`), so please let me know if that works or if additional change is required. If this change works, it fixes this bug for now (but it also add duplicate images to `ImageList` as seen in the screenshot below), but I'm not sure if this problem goes deeper than just this link issue since my images don't have it.
Also, using the navigation manager might also have the same problem since `navigateToImage` requires also an image tag, and here we have two different image tags from `ContainerInfo` and `ImageInfo`
Screenshot from 2024-07-09 19-34-21

axel7083:
> it fixes this bug for now (but it also add duplicate images to `ImageList` as seen in the screenshot below)

This is a pretty big issue sadly, I did not notice this side effect when I made the change, this is probably not something we would want. I am not sure that would be acceptable.

Maybe one possible long term solution would be to stop using a mixte of many elements for the image url. As you mention the `navigateToImage` requires also some `image tag`.

I am not sure of the solution, it should probably be discussed with @benoitf when he will be back from PTO in a daily.

benoitf:
The basic idea of the links is that when you do click on a specific image, it means you may have an associated tag and  you don't want to see other images sharing the same sha256 id.

And we don't want to list extra images with only the ids. (Here we want to have the same behavior than Docker Desktop, one line per tag even and not one line per image id or one extra line just with the image id)

From the provided analysis, it seems that for some case we don't have tags but only the ids.

We should not then add the id of the image to the tag as it'll cause extra lines in the image lists.

I would suggest to have different links, one for tags and one for id or a single link but with a prefix, either giving the id or the name of the tag.

so either a link to the id of the image like `#/images/sha256:462f6b66/engineid` (and in that case we could display all the tags of the image) or #/images/sha256:462f6b66/engineid/<base64 of the tag>`

or use `#/images/sha256:462f6b66/engineid/<base64 of either tag:tagName or sha256:id>` 

when decoding the base64, if it starts by tag we do a matching per tag, if sha wedo the matching on the image id.

SoniaSandler:
> so either a link to the id of the image like `#/images/sha256:462f6b66/engineid` (and in that case we could display all the tags of the image) or #/images/sha256:462f6b66/engineid/`

@benoitf If I understood correctly, for id of image, the link will lead to an `ImageList` page with all the image with the same `imageId` we are looking for, right?

benoitf:
@SoniaSandler yes, unless we can be sure that we never have a case with only the id and we always have the tag of the image ?

SoniaSandler:
@benoitf Axel got a container where the link to the image has the id as the tag, so I don't think we can be 100% sure it won't happen again, but right now there is no page rendering for `images/imageId/engineId`, so I can add it in a separate PR if we want to show a list of images with only a specific `imageId`

SoniaSandler:
@benoitf @axel7083 right now this PR is finished from my side. However, I can't verify that Axel's edge case works since I don't have images/containers with similar problems. Working links of images with Id instead of a tag should lead to pages with the route `images/imageId/engineId`.
However, since we currently don't have rendering for this route, in this PR, they will lead to an empty page. I opened a issue for the creation of page rendering for `images/imageId/engineId`- https://github.com/containers/podman-desktop/issues/8216

axel7083:
> @benoitf @axel7083 right now this PR is finished from my side. However, I can't verify that Axel's edge case works since I don't have images/containers with similar problems. Working links of images with Id instead of a tag should lead to pages with the route `images/imageId/engineId`. However, since we currently don't have rendering for this route, in this PR, they will lead to an empty page. I opened a issue for the creation of page rendering for `images/imageId/engineId`- #8216

On my side it seems to be fixing the problem

https://github.com/user-attachments/assets/99a86733-b25f-4379-b942-ebeac8164a79

packages/renderer/src/lib/container/container-utils.ts
```diff
@@ -18,6 +18,8 @@
 
 import type { Port } from '@podman-desktop/api';
 import { ContainerIcon } from '@podman-desktop/ui-svelte/icons';
+// eslint-disable-next-line unicorn/prefer-node-protocol
+import { Buffer } from 'buffer';
```
axel7083, line 22:
```suggestion
import { Buffer } from 'node:buffer';
```

We have a preference for `node:*` import, that's why they was a eslint warning

packages/renderer/src/lib/container/container-utils.ts
```diff
@@ -136,6 +138,19 @@ export class ContainerUtils {
...
+      if (imageTag.startsWith('sha256:')) {
+        return `/images/${containerInfo.ImageID}/${containerInfo.engineId}`;
+      } else {
+        return `/images/${containerInfo.ImageID}/${containerInfo.engineId}/${containerInfo.ImageBase64RepoTag}/summary`;
+      }
+    } else {
+      return `/images/${containerInfo.ImageID}/${containerInfo.engineId}`;
+    }
+  }
+
```
benoitf: it seems we are not testing the case without sha256 ?

packages/renderer/src/lib/container/container-utils.ts
SoniaSandler: @axel7083 I get an error and a white screen for PD if I use `node:buffer` instead of `buffer`
Screenshot from 2024-07-25 08-47-35

benoitf: ah yes it's because you're on the frontend side

benoitf: you need to keep `buffer`

packages/renderer/src/lib/container/container-utils.ts
SoniaSandler: @benoitf I'm not sure I understand what you mean. Do you mean if there is no tag, but the `imageId` doesn't start with `sha256` -> add `containerInfo.ImageBase64RepoTag === containerInfo.ImageId` to the `if` statement in line 144?

packages/renderer/src/lib/container/container-utils.ts
axel7083: Okey I see the issue. But why are we using Buffer to decode a base64 instead of `btoa` @benoitf  any idea ?

benoitf (approved):
make sure the lines of getImageHref method are all tested

also maybe name a constant for `/images/${containerInfo.ImageID}/${containerInfo.engineId}`;
as we copy/paste 3 times the same value

but I'm not sure we need all these if else if else

could be a single condition
```ts
const imageTag = containerInfo.ImageBase64RepoTag ? Buffer.from(...) : ''
if (imageTag.startsWith('sha256')
 ....
else
...
```

so no need to have the constant

packages/renderer/src/lib/container/container-utils.ts
benoitf: sorry I wasn't clear enough
I mean to have tests in the spec.ts file covering all the branches of your new method

packages/renderer/src/lib/container/container-utils.ts
benoitf: I think it's because they're flagged as deprecated

packages/renderer/src/lib/container/container-utils.ts
SoniaSandler: @benoitf Line 323 (in the `speac.ts` file) tests the url with an image tag, and the test in line 337 checks the url with an image id instead of a tag, but it looks like I'll have to add another one for the case of `if (containerInfo.ImageBase64RepoTag)`

packages/renderer/src/lib/container/container-utils.ts
axel7083: https://stackoverflow.com/a/70851350

As mention in the issue, only `btoa` on node has been deprecated not the one on DOM

benoitf (approved):

SoniaSandler:
@axel7083 do you approve these changes? merging is blocked due to request changes from June 18

axel7083 (approved):
LGTM ! Sorry I missed the request review

Issue #7680, opened 2024-06-17
ConfigMaps & Secrets Icon for Kubernetes dashboard


cdrage:
### Is your enhancement related to a problem? Please describe

We will need an icon that can represent both ConfigMaps and Secrets.

This is what Kubernetes uses to represent them:

![Screenshot 2024-06-17 at 4 26 37 PM](https://github.com/containers/podman-desktop/assets/6422176/11df62e4-578b-40f1-8378-53fb18ab62ed)

### Describe the solution you'd like

Perhaps an icon that looks similar with the lines?

### Additional context

Needed for #configmaps

ekidneyrh:
Here are two ideas, one is the document page we have throughout atm but with a lock in the middle.

SVG: 
![configsecret_idea2](https://github.com/containers/podman-desktop/assets/87311656/63d2627c-b85f-44bd-b35e-25ca8b66af7f)

Screenshot:
![image](https://github.com/containers/podman-desktop/assets/87311656/e07db22e-aa69-4729-8ed9-12d6d81bcec2)

The other is similar to the one shared above on the ticket - using lines to represent a document. 

SVG:
![configsecret_idea1](https://github.com/containers/podman-desktop/assets/87311656/66226e6a-e02c-4df7-b7be-4024b47dd75e)

Screenshot:
![image](https://github.com/containers/podman-desktop/assets/87311656/b6b2e7be-c64d-45e4-b0ed-30285fe447b1)

Both combine the config data (document) with the lock icon (secret/security). 

LMKWYT :)

deboer-tim:
If we went with the first I think I'd drop the document and just have the lock, but I think either way it gets the 'secrets' but isn't as obvious for 'config maps'.

For the second, I don't know how you'd make a solid version :) but again it conveys the secrets to me more than config maps. In the lower left of the Kube icon at the top the lines are broken up like key/value pairs or properties - maybe it would work if we broke up the lines a bit like that?

cdrage:
Same thoughts as @deboer-tim ^^ I like how the Kube icon does the whole broken up lines as well to convey that it's a key/value, so that plus the lock would be great.

I do find the lock looks a bit like a suitcase though? / I see no keyhole?

feloy:
Secret in Kubernetes is not about protecting the values, but about encoding the values, so you can use binaries values for secrets.

For this (and for trying to not propagate the belief that Kubernetes authors tried to protect values by encoding them), I would not use a lock in the icon

[...]

PR #7640, opened 2024-06-13
feat: add volumes (PVC) to kubernetes

cdrage:
feat: add volumes (PVC) to kubernetes

### What does this PR do?

* Adds PVC to Kubernetes
* Lists in the dashboard, available to edit, delete, etc.

### Screenshot / video of UI

Screenshot 2024-06-13 at 9 18 00 AM

https://github.com/containers/podman-desktop/assets/6422176/a438eda6-b41e-46ab-85b2-92f35ffc025e

### What issues does this PR fix or reference?

Closes https://github.com/containers/podman-desktop/issues/7341

### How to test this PR?

- [X] Tests are covering the bug fix or the new feature

Deploy a Kubernetes artifact with a PVC and view / check / play around
with the new dashboard for PVC.

Signed-off-by: Charlie Drage <charlie@charliedrage.com>

PR #7681, opened 2024-06-17
feat: add configmaps & secrets to kubernetes

### LINKED ISSUE #7190: Kubernetes Epic: Add rest of essentials: Volumes, Nodes, ConfigMap, Secrets.
### Epic domain

Kubernetes already has: Deployments, Services and Ingresses & Routes.

However, there are a few other "essential" parts that I suggest to be added to get to 90% functionality of a Kubernetes GUI.

- [x] Nodes (show node information / ex. k8s1, k8s2, k8snode, etc. was requested in chat) https://github.com/containers/podman-desktop/issues/7339
- [x] Persistent Volume Claims https://github.com/containers/podman-desktop/issues/7341
- [x] ConfigMap & Secrets (a lot of similar columns and would be good to have it on the same page) https://github.com/containers/podman-desktop/issues/7342

IMHO the reason why we include PVC and not Volumes is because Volumes is more of an administrative task (here is nfs storage, here is S3 storage, etc.) for your cluster. While PVC is "requesting" a volume.

These are the most "popular" parts of Kubernetes I believe should be added to cover all "essential" areas.

### Additional context

_No response_

### LINKED ISSUE #7342: Add "Storage" (ConfigMap & Secrets) section for Kubernetes page

Part of https://github.com/containers/podman-desktop/issues/7190

We should add a "Storage" section that will consistent of a combination of ConfigMap and Secrets.

This is commonly used in order to configure Deployments such as setting the environment variable & passing in certificates.

Added to Kubernetes page 

_No response_

### Additional context

_No response_

cdrage:
feat: add configmaps & secrets to kubernetes

### What does this PR do?

* Adds ConfigMaps and Secrets to Kubernetes
* Ability to delete / view / edit both configmap and secrets

### Screenshot / video of UI

Screenshot 2024-06-17 at 4 24 23 PM

https://github.com/containers/podman-desktop/assets/6422176/41108199-7e2b-4cc2-81f6-2b267b1c887a

Closes https://github.com/containers/podman-desktop/issues/7342
Closes https://github.com/containers/podman-desktop/issues/7190 (its the last part of it)

### How to test this PR?

1. Create a ConfigMap and Secrets on your k8s cluster.
2. Delete / Edit / View them

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -1191,6 +1191,17 @@ export class ColorRegistry {
       dark: colorPalette.gray[100],
       light: colorPalette.gray[400],
     });
+    // If it's considered "good" (like a successful operation), use green
+    this.registerColor(`${status}good`, {
+      dark: colorPalette.green[400],
+      light: colorPalette.green[400],
+    });
+    // If it's "secure" (like a successful TLS connection), use yellow
+    // as often keys are yellow in the real world / "caution" / padlock colour uses on other GUIs
+    this.registerColor(`${status}secure`, {
+      dark: colorPalette.amber[600],
+      light: colorPalette.amber[600],
+    });
     // Connections / login
     this.registerColor(`${status}connected`, {
       dark: colorPalette.green[600],
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -1191,17 +1191,6 @@ export class ColorRegistry {
       dark: colorPalette.gray[100],
       light: colorPalette.gray[400],
     });
-    // If it's considered "good" (like a successful operation), use green
-    this.registerColor(`${status}good`, {
-      dark: colorPalette.green[400],
-      light: colorPalette.green[400],
-    });
-    // If it's "secure" (like a successful TLS connection), use yellow
-    // as often keys are yellow in the real world / "caution" / padlock colour uses on other GUIs
-    this.registerColor(`${status}secure`, {
-      dark: colorPalette.amber[600],
-      light: colorPalette.amber[600],
-    });
     // Connections / login
     this.registerColor(`${status}connected`, {
       dark: colorPalette.green[600],
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -13,6 +13,7 @@ import type { ImageInfo } from '/@api/image-info';
 import { CommandRegistry } from './lib/CommandRegistry';
 import NewContentOnDashboardBadge from './lib/dashboard/NewContentOnDashboardBadge.svelte';
 import { ImageUtils } from './lib/image/image-utils';
+import ConfigMapSecretIcon from './lib/images/ConfigMapSecretIcon.svelte';
 import DashboardIcon from './lib/images/DashboardIcon.svelte';
 import DeploymentIcon from './lib/images/DeploymentIcon.svelte';
 import ExtensionIcon from './lib/images/ExtensionIcon.svelte';
@@ -34,11 +35,13 @@ import { contributions } from './stores/contribs';
 import { imagesInfos } from './stores/images';
 import { kubernetesContexts } from './stores/kubernetes-contexts';
 import {
+  kubernetesCurrentContextConfigMapsFiltered,
   kubernetesCurrentContextDeployments,
   kubernetesCurrentContextIngresses,
   kubernetesCurrentContextNodes,
   kubernetesCurrentContextPersistentVolumeClaims,
   kubernetesCurrentContextRoutes,
+  kubernetesCurrentContextSecretsFiltered,
   kubernetesCurrentContextServices,
 } from './stores/kubernetes-contexts-state';
 import { podsInfos } from './stores/pods';
@@ -55,12 +58,17 @@ let persistentVolumeClaimsSubscribe: Unsubscriber;
 let servicesSubscribe: Unsubscriber;
 let ingressesSubscribe: Unsubscriber;
 let routesSubscribe: Unsubscriber;
+let configmapsSubscribe: Unsubscriber;
+let secretsSubscribe: Unsubscriber;
 let combinedInstalledExtensionsSubscribe: Unsubscriber;
 
 let podCount = '';
 let containerCount = '';
 let imageCount = '';
 let volumeCount = '';
+let configmapsCount = 0;
+let secretsCount = 0;
+let configmapSecretsCount = '';
 let persistentVolumeClaimsCount = '';
 let contextCount = 0;
 let deploymentCount = '';
@@ -145,6 +153,15 @@ onMount(async () => {
     routesCount = value.length;
     updateIngressesRoutesCount(ingressesCount + routesCount);
   });
+
+  configmapsSubscribe = kubernetesCurrentContextConfigMapsFiltered.subscribe(value => {
+    configmapsCount = value.length;
+    updateConfigMapSecretsCount(configmapsCount + secretsCount);
+  });
+  secretsSubscribe = kubernetesCurrentContextSecretsFiltered.subscribe(value => {
+    secretsCount = value.length;
+    updateConfigMapSecretsCount(configmapsCount + secretsCount);
+  });
   contextsSubscribe = kubernetesContexts.subscribe(value => {
     contextCount = value.length;
   });
@@ -185,8 +202,16 @@ onDestroy(() => {
   if (servicesSubscribe) {
     servicesSubscribe();
   }
+  if (configmapsSubscribe) {
+    configmapsSubscribe();
+  }
+  if (secretsSubscribe) {
+    secretsSubscribe();
+  }
   ingressesSubscribe?.();
   routesSubscribe?.();
+  configmapsSubscribe?.();
+  secretsSubscribe?.();
   combinedInstalledExtensionsSubscribe?.();
 });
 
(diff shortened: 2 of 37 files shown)


cdrage:
FYI the icon is the "rocket" temporary until #7680 has a PR.

Apologizes for the large PR! Tried splitting it up a bit, but a lot of the tests are intertwined.

Most of this code is very similar to #7640 however.

packages/renderer/src/AppNavigation.svelte
```diff
@@ -34,11 +35,13 @@ import { contributions } from './stores/contribs';
 import { imagesInfos } from './stores/images';
 import { kubernetesContexts } from './stores/kubernetes-contexts';
 import {
+  kubernetesCurrentContextConfigMapsFiltered,
```
deboer-tim: Should be the unfiltered, not the filtered store here.

packages/renderer/src/AppNavigation.svelte
```diff
@@ -34,11 +35,13 @@ import { contributions } from './stores/contribs';
...
   kubernetesCurrentContextDeployments,
   kubernetesCurrentContextIngresses,
   kubernetesCurrentContextNodes,
   kubernetesCurrentContextPersistentVolumeClaims,
   kubernetesCurrentContextRoutes,
+  kubernetesCurrentContextSecretsFiltered,
```
deboer-tim: Should be the unfiltered, not the filtered store here.

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretUI.ts
```diff
@@ -0,0 +1,27 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
deboer-tim: 2024

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretList.svelte
```diff
@@ -0,0 +1,189 @@
...
+  if (selectedConfigMapsSecrets.length > 0) {
+    bulkDeleteInProgress = true;
+    await Promise.all(
+      selectedConfigMapsSecrets.map(async configmapSecret => {
+        const isSecret = configmapSecretUtils.isSecret(configmapSecret);
+        try {
+          if (isSecret) {
+            await window.kubernetesDeleteSecret(configmapSecret.name);
+          } else {
+            await window.kubernetesDeleteConfigMap((configmapSecret as ConfigMapSecretUI).name);
```
deboer-tim: I think it would be worth the readability/type checking if you created a configmapSecretUtils.isConfigMap() and just turned this into a `configmapSecretUtils.isSecret() {...} else if configmapSecretUtils.isConfigMap() {...}`

packages/renderer/src/lib/ui/Label.svelte
```diff
@@ -3,12 +3,13 @@ import { Tooltip } from '@podman-desktop/ui-svelte';
 
 export let name = '';
 export let tip = '';
+export let capitalize = true;
```
deboer-tim: I went with 'capitalize on' because it was on the first instance I made a component out of, but most of the files that have used it since haven't needed this. As a component I wonder if it's better if capitalize is off by default, i.e. an optional feature instead.

packages/main/src/plugin/color-registry.ts
```diff
@@ -1157,6 +1157,17 @@ export class ColorRegistry {
       dark: colorPalette.gray[100],
       light: colorPalette.gray[400],
     });
+    // If it's considered "good" (like a successful operation), use green
+    this.registerColor(`${status}good`, {
```
deboer-tim: To me there's a disconnect between the name 'good' and the actual use 'config map exists'. I think we should either just use an existing color (e.g. 'updated' from #7526) or name it something more generic like (object) 'exists'.

packages/main/src/plugin/color-registry.ts
```diff
@@ -1157,6 +1157,17 @@ export class ColorRegistry {
...
+      dark: colorPalette.green[400],
+      light: colorPalette.green[400],
+    });
+    // If it's "secure" (like a successful TLS connection), use yellow
+    // as often keys are yellow in the real world / "caution" / padlock colour uses on other GUIs
+    this.registerColor(`${status}secure`, {
```
deboer-tim: Amber is more of a warning color than a yellow lock. I think @ekidneyrh should weigh in on what color 'secure' is, whether we need another color in the palette or should just use 'status-connected' (green 600).

If we do go the route of a 'secure' color different from connected, I'd wonder if Log In or other places should use that instead of -connected.

deboer-tim (commented):
Code worked fine for me in testing and added a few comments, but this PR is so massive that I haven't gone through it all. Parts of this really should be broken out so that we can review and merge independently.

cdrage:
@deboer-tim I've updated the PR with a new commit with the suggested changes:

- Added isConfigMap / made the if statements more intuitive
- Use green for the "running" / working configmap and secret.
- Updated the incorrect filtered areas


Still unsure what to do with Label? I guess another PR / issue, since removing the capitalization will affect other parts of PD.

packages/renderer/src/lib/configmaps-secrets/configmap-secret-utils.ts
```diff
@@ -0,0 +1,56 @@
...
+export class ConfigMapSecretUtils {
+  // If it is a secret, then it will have a type property,
+  // if not it's configmap
+  isSecret(storage: V1ConfigMap | V1Secret): storage is V1Secret {
+    return 'type' in storage && storage.type !== 'ConfigMap';
+  }
+
+  // If it is a configMap, then it will **not** have a type property
+  isConfigMap(storage: V1ConfigMap | V1Secret): storage is V1ConfigMap {
+    return !('type' in storage);
```
deboer-tim: It would probably be better to have a 'positive' check for something unique about config map vs 'it's not a secret' 😆. Is there nothing that's only in configmap?

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretList.svelte
deboer-tim: I think you can remove the `as ConfigMapSecretUI` now?

packages/renderer/src/lib/configmaps-secrets/configmap-secret-utils.ts
```diff
@@ -0,0 +1,56 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
deboer-tim: Should be 2024

packages/renderer/src/lib/configmaps-secrets/configmap-secret-utils.ts
cdrage: actually yes! I'll update it to also check for 'binaryData' too.

feloy:
configmaps-list

The values in the Type column for secrets are the ones displayed in `kubectl get secret`, but I would prefer to have the ones used when using `kubectl create secret <type>` (docker-registry, generic or tls). WDYT?

For specific secrets (docker, tls), I wonder if we need to display the keys, as they are standard (and not really useful to the user).

`kubectl get cm / secret` display the number of data, without the names of the keys. I wonder if it is relevant to display the key names in the list, as it can easily be unreadable

feloy:
configmap-details

I would prefer to have the display of the data moved up one level:

```
Details

Type      Opaque
Immutable  No


Data

key1       value1
key2       value2
key3       value3
key4       value4
key5       value5
```

cdrage:
Thank you for the suggestions @feloy I agree. Having keys on there is not ideal. Other GUI's do it, but because of our spacin and simplicity it's nice to say the length / how many rather than every name of the key..

I've updated the UI with the "Summary" changes as well as the list changes. I agree it looks a lot better.

Screenshot 2024-06-19 at 2 10 40 PM
Screenshot 2024-06-19 at 2 10 54 PM

packages/renderer/src/lib/kube/details/KubeSecretArtifact.svelte
```diff
@@ -21,12 +23,13 @@ export let artifact: V1Secret | undefined;
...
-          <div>{key}: {value}</div>
-        {/each}
-      </Cell>
+      <Subtitle>Data</Subtitle>
+    </tr>
+    <tr>
+      {#each Object.entries(artifact.data) as [key, value]}
+        <Cell>{key}</Cell>
+        <Cell>{value}</Cell>
+      {/each}
```
feloy: All the keys appear in the same line

cdrage:
Screenshot 2024-06-19 at 11 50 34 AM
@feloy updated

cdrage:
Screenshot 2024-06-19 at 4 36 14 PM
Updated so now they are on newlines @feloy

packages/renderer/src/lib/ui/Label.svelte
deboer-tim: This was delivered via #7788, can remove it here.

cdrage:
Rebased against the label changes and ready for review @feloy @deboer-tim 

I did not make any changes other than the Label conflict.

deboer-tim:
> Rebased against the label changes and ready for review @feloy @deboer-tim
> 
> I did not make any changes other than the Label conflict.

Build failing. TBH this is still a massive PR and I would still feel more comfortable if you threw the main/preload/store in another PR first, as I think those could be merged easy and would cut down the size/focus of this a lot.

cdrage:
> > Rebased against the label changes and ready for review @feloy @deboer-tim
> > I did not make any changes other than the Label conflict.
> 
> Build failing. TBH this is still a massive PR and I would still feel more comfortable if you threw the main/preload/store in another PR first, as I think those could be merged easy and would cut down the size/focus of this a lot.

No problem! Going to close this and split it into 1 for API / backend, other for UI.

PR #7684, opened 2024-06-18
feat: show node and namespace information for k8s pods

### LINKED ISSUE #7646: Remove the "id" for Kubernetes pod and replace with node name?

The ID is really odd in the Kubernets pod listing because it doesn't relate to anything other than UUID which is rarely used...

IMHO. We should replace it instead with the node that the pod is located on, as well as the namespace beside it?

Instead of "67990026" have the node it is located on instead as well as the namespace beside it.

Screenshot 2024-06-13 at 4 23 49 PM

Similar to `kubectl get pods -o wide`:

```sh
NAME                                   READY   STATUS    RESTARTS   AGE     IP           NODE   NOMINATED NODE   READINESS GATES
neko-67ff554889-c4l8m                  1/1     Running   0          11m     10.42.1.35   k8s2   <none>           <none>
git-7f6dc4f666-pqhbr                   1/1     Running   0          11m     10.42.5.38   k8s4   <none>           <none>
adguard-79b5ff976d-25wfz               1/1     Running   0          11m     10.42.5.39   k8s4   <none>           <none>
```

_No response_

### Additional context

_No response_

cdrage:
feat: show node and namespace information for k8s pods

### What does this PR do?

* Shows node and namespace information for Kubernetes pods on the Pods
  page

### Screenshot / video of UI

Screenshot 2024-06-17 at 10 39 00 PM

Screenshot 2024-06-17 at 10 36 56 PM

Closes https://github.com/containers/podman-desktop/issues/7646

### How to test this PR?

View Kubernetes pods and see the node and namespace information.

### DIFF
--- a/packages/main/src/plugin/api/pod-info.ts
+++ b/packages/main/src/plugin/api/pod-info.ts
@@ -29,6 +29,8 @@ export interface PodInfo extends LibPodPodInfo {
   engineId: string;
   engineName: string;
   kind: 'kubernetes' | 'podman';
+  // Optional information only used by Kubernetes
+  node?: string;
 }
 
 export interface PodInspectInfo extends LibPodPodInspectInfo {
--- a/packages/main/src/plugin/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes-client.ts
@@ -112,6 +112,7 @@ function toPodInfo(pod: V1Pod, contextName?: string): PodInfo {
     engineId: contextName ?? 'kubernetes',
     engineName: 'Kubernetes',
     kind: 'kubernetes',
+    node: pod.spec?.nodeName,
   };
 }
 
--- a/packages/renderer/src/lib/pod/PodColumnName.spec.ts
+++ b/packages/renderer/src/lib/pod/PodColumnName.spec.ts
@@ -38,6 +38,8 @@ const pod: PodInfoUI = {
   selected: false,
   containers: [],
   kind: 'podman',
+  node: 'node1',
+  namespace: 'default',
 };
 
 test('Expect simple column styling', async () => {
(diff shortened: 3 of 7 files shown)


packages/main/src/plugin/api/pod-info.ts
```diff
@@ -29,6 +29,9 @@ export interface PodInfo extends LibPodPodInfo {
   engineId: string;
   engineName: string;
   kind: 'kubernetes' | 'podman';
+  // Used for Kubernetes
+  node?: string;
+  namespace?: string;
```
axel7083: Instead of adding a lot of optional fields, I would rather have a new interface

```ts
interface KubernetesPodInfo extend PodInfo {
    node: string;
    namespace: string;
}
```
And an utility method

```ts
export function isKubernetesPod(pod: PodInfo): pod is KubernetesPodInfo  {
    return pod.kind === 'kubernetes';
}
```

cdrage: So in the end it looks like only one is being added for `node` which I think would be okay.

Namespace is already being used by PodInfo: https://github.com/containers/podman-desktop/blob/c1ac1291a66c09caf559edccda1350ef137f65bf/packages/main/src/plugin/dockerode/libpod-dockerode.ts#L31

I've adjusted the comment as `node` would be kubernetes only.

cdrage:
@axel7083 Updated, I adjusted the comment, since it's only one additional UI entry, I think it'd be okay? Namespace looks to be used between pods + kubernetes anyways.

axel7083 (approved):
I am okey with it 👍  LGTM

packages/main/src/plugin/api/pod-info.ts
axel7083: If namespace is also in podman, should it be displayed ?

packages/main/src/plugin/kubernetes-client.ts
```diff
@@ -112,6 +112,8 @@ function toPodInfo(pod: V1Pod, contextName?: string): PodInfo {
     engineId: contextName ?? 'kubernetes',
     engineName: 'Kubernetes',
     kind: 'kubernetes',
+    node: pod.spec?.nodeName,
+    namespace: pod.metadata?.namespace,
```
lstocchi: As you said above there is already the Namespace field, so why not reusing it? It has the same content. -> https://github.com/containers/podman-desktop/pull/7684/files#diff-8799a75d3c2928c9fe5136fceab66ed0edad4de3a675f9f5f6d3bc3c61f3dd7bL109

lstocchi: Maybe you didn't push your commit

packages/main/src/plugin/api/pod-info.ts
cdrage: good question, it looks like it's blank? but somehow it's in PodInfo. Not entirely sure why.

cdrage:
@lstocchi @axel7083 Updated again, apologizes as for some reason my updated commit did not go through. Ready for another review  👍

lstocchi (approved):
LGTM

axel7083 (approved):

PR #7710, opened 2024-06-18
chore: PVC icon

deboer-tim:
### What does this PR do?

Adds a new PVC icon as per design.

### Screenshot / video of UI

Before:

Screenshot 2024-06-18 at 9 05 03 AM

After:

Screenshot 2024-06-18 at 9 04 05 AM

### What issues does this PR fix or reference?

Fixes #7641.

### How to test this PR?

Check main nav, use PVC page.

Issue #7641, opened 2024-06-13
Icon for Persistent Volume Claim for Kubernetes


cdrage:
### Is your enhancement related to a problem? Please describe

We should have an icon that's different than the normal "volumes" icon for PVC. Right now in the PR it is re-using the volume icon: https://github.com/containers/podman-desktop/pull/7640

Taken from google, I was thinking something like this?

![image](https://github.com/containers/podman-desktop/assets/6422176/917bbfa5-631b-45cc-b1ad-89c39bf4a282)

It makes sense for it to be a "dashed" volume since it's "claiming".

### Describe the solution you'd like

Icon for PVC

ekidneyrh:
SVGs:
![pvc](https://github.com/containers/podman-desktop/assets/87311656/82d01d5f-edeb-458f-b70d-840b3956ed56)
![pvc-filled](https://github.com/containers/podman-desktop/assets/87311656/84075628-57fc-40d5-b34e-cc22659e6888)

PNGs:
![pvc-filled](https://github.com/containers/podman-desktop/assets/87311656/02a80fcb-7a8f-4d16-b387-0b410f549568)
![pvc](https://github.com/containers/podman-desktop/assets/87311656/2e9dcd29-cca0-4304-85ab-b7238c88d7e9)

Here's a screenshot from inkscape for better visibility :) 

![image](https://github.com/containers/podman-desktop/assets/87311656/39065cf6-2eaf-4807-8f0a-77f2e852f662)

WDYT?

cdrage:
[...]

PR #7711, opened 2024-06-18
chore: improved solid volume icon

deboer-tim:
### What does this PR do?

Updated solid volume icon. IMHO this version stands out (pops) better and is a little clearer on the solid background, plus it will match with the PVC icon (PR #7710) that really requires the full outline.

### Screenshot / video of UI

Before:

Screenshot 2024-06-18 at 7 59 15 AM

After:

Screenshot 2024-06-18 at 7 58 35 AM

Raised as part of #7641, matches PVC icon in #7710.

### How to test this PR?

Open volumes page.


axel7083 (approved):
Should we also change the one in the nav bar ? Otherwise look good 

image

deboer-tim:
> Should we also change the one in the nav bar ? Otherwise look good

The outline (non-solid) version of the new PVC icon is just a dashed version of the volume icon, so they're still consistent. IMHO adding a second outline would look odd compared to everything else, but that's up to @ekidneyrh.

lstocchi (approved):
LGTM

deboer-tim:
Rebased to avoid build issues.

PR #7723, opened 2024-06-18
fix: table properly support undefined value result from rendermapping

### LINKED ISSUE #7722: bug(Table): rendermapping result undefined provide all object to renderer
### Bug description

https://github.com/containers/podman-desktop/blob/d722e153b22df3e56c0fab7263d5c59123bbbf0f/packages/ui/src/lib/table/Table.svelte#L273

As visible above, there is a clear problem in a specific case, when the result of rendermapping is undefined we provide the full object.

This problem occured, as the `TableColumnDuration` support undefined, but when returning `undefined` through rendermapping, the renderer was getting the full object, instead of the value I returned.

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Fixing the following line

https://github.com/containers/podman-desktop/blob/d722e153b22df3e56c0fab7263d5c59123bbbf0f/packages/ui/src/lib/table/Table.svelte#L273

with 

https://github.com/containers/podman-desktop/blob/2c6b12385c8e9d3af2138c70227891687c4b6b36/packages/ui/src/lib/table/Table.svelte#L273

### Screenshot / video of UI

N/A

Fixes https://github.com/containers/podman-desktop/issues/7722

### How to test this PR?


packages/ui/src/lib/table/Table.svelte
```diff
@@ -270,7 +270,7 @@ function toggleChildren(name: string | undefined): void {
               {#if column.info.renderer}
                 <svelte:component
                   this="{column.info.renderer}"
-                  object="{column.info.renderMapping?.(object) ?? object}"
+                  object="{column.info.renderMapping ? column.info.renderMapping(object) : object}"
```
deboer-tim, line 273: Please apply the identical fix to line 310 (child rows).

deboer-tim (approved):
Thanks for fixing 👍🏼, just need to apply the same fix to child rows.

packages/ui/src/lib/table/Table.svelte
axel7083: I applied the fix to the children too! Thanks for catch

cdrage (approved):

cdrage:
Hi sorry @axel7083 pressed the "update" button to try and fix the rebasing issue, but I believe this failure is part of the PR if you are able to update it.

benoitf (approved):

axel7083:
> Hi sorry @axel7083 pressed the "update" button to try and fix the rebasing issue, but I believe this failure is part of the PR if you are able to update it.

Yeah I did a bad copy past... fixed it !

Issue #7728, opened 2024-06-18
Light mode for status icon


deboer-tim:
### Is your enhancement related to a problem? Please describe

Status icon is picking up the default text color for 'started' and likely other modes.

### Describe the solution you'd like

Define any missing colors if necessary and set an explicit color.

Issue #7727, opened 2024-06-18
Light mode for status bar


deboer-tim:
### Is your enhancement related to a problem? Please describe

The status bar page doesn't support light mode and is just picking up the default text color.

### Describe the solution you'd like

Define any missing colors in the light mode palette and get the status bar fully supporting light mode.

Issue #7708, opened 2024-06-18
Extension error message in Error tab is not readable in light mode


cbr7:
### Bug description

The error messages in the `Error` tab in the Extension details menu are not readable with light mode enabled.

![image (5)](https://github.com/containers/podman-desktop/assets/106525396/f326987d-7ade-4428-9335-45646b81fa04)

### Operating system

macOS

### Installation Method

Installer from website/GitHub releases

### Version

next (development version)

### Steps to reproduce

1. Have latest PD release.
2. Switch to light mode.
3. Install an extension that fails to startup (e.g openshift local)
4. Go to the extension details menu for it and click the `Error` tab to view the error message.
5. Notice that the error message is not readable in light mode.

PR #7729, opened 2024-06-18
chore: use color var in index.html

### LINKED ISSUE #7724: Light mode default text color (Replace text-white in index.html with color variable)

I noticed that some of the light mode issues we're facing are due to text-white being set in index.html, regardless of which light/dark mode is in use.

Although this is helping to find cases where the variables are missing for light mode, we should set a default text color that's appropriate to the appearance mode. Otherwise, this is just tilting things toward dark mode and making missing variables have much worse default behaviour.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

We should not hardcode the default text colour to white, especially in light mode. This is the (temporary) cause of many issues like #7708 where we hadn't yet set a light mode color so you get white on light.

We don't have a 'default text color' and none of the existing palette entries fit, so I talked to Emma and decided to create a new variable.

Unfortunately, even though this is a 'correct' change, setting it right now has the reverse problem: anything that is still missing light mode support will see the default text color change from white to dark and have problems on dark backgrounds. A quick test showed the following:

- some forms (#7214)
- #7727
- #7728

So we should not move this out of draft until most of these are fixed, and definitely not immediately before a release. In the meantime, it's handy to set this to something really obvious (e.g. amber) to see what still needs light mode support.

### Screenshot / video of UI

N/A

Fixes #7724.

### How to test this PR?

TODO

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -228,6 +228,7 @@ export class ColorRegistry {
   }
 
   protected initColors(): void {
+    this.initDefault();
     this.initGlobalNav();
     this.initSecondaryNav();
     this.initTitlebar();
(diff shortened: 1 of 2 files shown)


deboer-tim:
Replaced by #8061.

PR #7733, opened 2024-06-19
fix(UI): formpage missing content shadow

### LINKED ISSUE #7549: Form page shadow doesn't extend past header
### Bug description

The shadow that should extend down the full left column is only in the header for form pages, including Help, Troubleshooting, Build, etc.

Including screen cap from light mode as it is much more obvious:
Screenshot 2024-06-10 at 10 15 20 AM

### Operating system

macOS 14.5

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

As suggested by @deboer-tim in https://github.com/containers/podman-desktop/issues/7549#issuecomment-2176511633 since we are using a single color for the header and the content, removing the border and moving up the bg-color to avoid hidden the shadow

### Screenshot / video of UI

| Before | After |
| --- | --- |
|  image | image |

Fixes https://github.com/containers/podman-desktop/issues/7549

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/layouts/FormPage.svelte
+++ b/packages/ui/src/lib/layouts/FormPage.svelte
@@ -34,7 +34,7 @@ function handleKeydown(e: KeyboardEvent): void {
 
 <svelte:window on:keydown="{handleKeydown}" />
 
-<div class="flex flex-col w-full h-full shadow-pageheader">
+<div class="flex flex-col w-full h-full shadow-pageheader bg-[var(--pd-content-bg)]">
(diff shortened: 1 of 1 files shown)


lstocchi (approved):
Much better 🚀 LGTM!!

benoitf (approved):

deboer-tim (commented):
It took me one minute too long to review 😛 but here it is anyway: +1. Worth noting that the Troubleshooting page looks 'worse' (no line below the tabs) but I think that's a sign it should actually be a DetailsPage.

PR #7773, opened 2024-06-21
feat: enable light mode for buildImageFromContainerfile page

lstocchi:
### What does this PR do?

This PR enables light mode in the build image from container file page.
It is built over https://github.com/containers/podman-desktop/pull/7768

### Screenshot / video of UI

buildimagefromcont

it is part of #7214 

### How to test this PR?

go to images -> click on build button (top right)

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -464,7 +464,7 @@ export class ColorRegistry {
 
     this.registerColor(`${ct}card-header-text`, {
       dark: colorPalette.gray[100],
-      light: colorPalette.charcoal[900],
+      light: colorPalette.purple[900],
     });
 
     this.registerColor(`${ct}card-bg`, {
@@ -495,6 +495,11 @@ export class ColorRegistry {
     this.registerColor(`${ct}card-inset-bg`, {
       dark: colorPalette.charcoal[900],
       light: colorPalette.dustypurple[200],
+    });
+
+    this.registerColor(`${ct}card-hover-inset-bg`, {
+      dark: colorPalette.charcoal[700],
+      light: colorPalette.dustypurple[300],
     });
 
     this.registerColor(`${ct}bg`, {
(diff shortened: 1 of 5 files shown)


deboer-tim (commented):
Code LGTM and worked well in testing. My only concern is if we should really have --pd-formpage-card*, i.e. whether these should be generic cards (variable names without formpage) or there are existing color variables we should use. @ekidneyrh any preference?

jeffmaury (approved):
LGTM. Same remarks about reusing variables

lstocchi:
Not sure how to change the variables name. By looking at other cases i see that the cards used for carousel have `content-card-carousel-*`, then we have generic content card `content-card-*`, then card related to `invert-content-` have prefix `invert-content-card-*`. So based on their scope we have different prefixes. As these are related to formpage i thought that using formpage prefix made sense. 
So which are your suggestions? 
By looking at the code, they are actually buttons `<button></button>` styled like cards. Is `content-button-card-*` good enough ?? Anything else?

@deboer-tim @jeffmaury

deboer-tim:
> Not sure how to change the variables name. By looking at other cases i see that the cards used for carousel have `content-card-carousel-*`, then we have generic content card `content-card-*`, then card related to `invert-content-` have prefix `invert-content-card-*`. So based on their scope we have different prefixes. As these are related to formpage i thought that using formpage prefix made sense. So which are your suggestions? By looking at the code, they are actually buttons `<button></button>` styled like cards. Is `content-button-card-*` good enough ?? Anything else?

Our design has most items that always look the same no matter where they are (Buttons, Inputs, Checkboxes, etc), and then some things like content cards that are different depending on context like Dashboard vs Settings. If the design/colors for the platform cards don't match tiles like on the dashboard or elsewhere then I'm fine if they are new vars just for forms, but card-input-bg is definitely 'wrong' because inputs shouldn't change based on where they are. Can you pick up some of the input-field colors instead or switch to use Input instead of defining a new color?

lstocchi:
@deboer-tim updated. TBH i completely missed the form page design made by Emma https://github.com/containers/podman-desktop/wiki/Design-System-Theming#form-page when i started working on the form pages. I was copying what was done in other pages. By following her mockup the job is much easier and i've been removing all formpage styles. 
I will create another PR to remove it from previous updates. Let me know if this looks good now. I edited the gif in the first message

deboer-tim (approved):
Thanks for the changes, working well. 👍🏼

PR #7775, opened 2024-06-21
chore: bump vale action to latest

axel7083:
### What does this PR do?

Vale seems to fail all the time see https://github.com/containers/podman-desktop/actions/workflows/pr-check-vale.yaml

> Might be related to the ubuntu upgrade ?

This aims to upgrade to their latest version to check if that fix the issue :/

### How to test this PR?


benoitf (approved):
either this or we downgrade Ubuntu but yes I would avoid to downgrade

deboer-tim (approved):

PR #7725, opened 2024-06-18
feat: uses TableColumnDuration for container list table

axel7083:
### What does this PR do?

Migrate the custom refresh mechanism to compute uptime, to uses the column duration

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

Fixes https://github.com/containers/podman-desktop/issues/7716

⚠️ there is a bug when the container is not running, rebase after https://github.com/containers/podman-desktop/pull/7723 will fix it.

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature

PR #7787, opened 2024-06-23
chore: use duration column for pod age

### LINKED ISSUE #2341: Pod age disappears
### Bug description

When I create a new pod (used podify) and then go to the Pods page, the age disappears after a few seconds. When I navigate away and back they are there again... and then disappear again.

https://user-images.githubusercontent.com/19958075/236006497-0300f675-b6e9-4383-86d9-a823938980be.mov

After a minute or so the pod age is fine and no longer disappears.

### Operating system

macOS 13.3 M1

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

### LINKED ISSUE #4972: Refactor the refreshing of age in tables

deboer-tim:
### What does this PR do?

Same change as #7725, but for pods. Use the table duration column instead of custom pod age. This cleans up the source and (by removing the 'if running' there is no more age disappearing.

There's a separate issue that user might prefer 'time since started', but that is unrelated.

### Screenshot / video of UI

N/A

Fixes #4972.
Fixes #2341.

### How to test this PR?

Check the pods page for regressions creating/starting/stopping pods.


benoitf (approved):

axel7083 (approved):
Love to see such changes in morning!

Issue #7801, opened 2024-06-24
Display the files present in an image


feloy:
### Is your enhancement related to a problem? Please describe

As a user, I would like to know the images present in an image, the files contained on each filesystem layer of the image, the sizes of the different layers, etc

### Describe the solution you'd like

Provide an API for extensions to register as an Image files provider, and adds a `File` tab in the details page of an Image, to display the files in the different layers (in a way similar to [dive](https://github.com/wagoodman/dive)).

PR #7802, opened 2024-06-24
add API for Image Files providers

feloy:

### What does this PR do?

Adds an API for extensions to registrer as an Image Files provider.

Inspiration from [spec](https://github.com/opencontainers/image-spec/blob/main/spec.md) and [Node implementation of Image builder](https://github.com/google/nodejs-container-image-builder).

Part of #7801

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -687,6 +687,110 @@ declare module '@podman-desktop/api' {
     connection: ContainerProviderConnection;
   }
 
+  /**
+   * the description of a file in an image filesystem layer
+   */
+  export interface ImageFile {
+    path: string;
+    type: 'directory' | 'symlink' | 'file';
+    // File mode encoded on 4 octal digits
+    mode: number;
+    uid: number;
+    gid: number;
+    size: number;
+    ctime: Date;
+    atime: Date;
+    mtime: Date;
+  }
+
+  /**
+   * the description of a symlink in an image filesystem layer
+   */
+  export interface ImageFileSymlink extends ImageFile {
+    type: 'symlink';
+    // path of the target file
+    linkPath: string;
+  }
+
+  /**
+   * a filesystem layer of an image as defined by [spec](https://github.com/opencontainers/image-spec/blob/main/spec.md)
+   */
+  export interface ImageFilesystemLayer {
+    /**
+     * unique id of the layer
+     */
+    id: string;
+    /**
+     * the command which created the layer
+     */
+    createdBy?: string;
+    /**
+     * files indicate the paths of the files added or modified in the layer
+     */
+    files?: ImageFile[];
+    /**
+     * whiteouts indicate the paths of the files to be deleted from previous layers.
+     */
+    whiteouts?: string[];
+    /**
+     * opaque whiteouts indicate the directories in which the content has to be completely deleted from previous layers.
+     */
+    opaqueWhiteouts?: string[];
+  }
+
+  /**
+   * the complete list of filesystem layers
+   */
+  export interface ImageFilesystemLayers {
+    layers: ImageFilesystemLayer[];
+  }
+
+  /**
+   * Interface to be implemented by image files providers
+   */
+  export interface ImageFilesCallbacks {
+    /**
+     *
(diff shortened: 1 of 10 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -4290,6 +4290,81 @@ declare module '@podman-desktop/api' {
...
 
+  /**
+   * the description of a file in an image filesystem layer
+   */
+  export interface ImageFile {
+    type: 'directory' | 'symlink' | 'file';
+    // File mode encoded on 4 octal digits
+    mode: number;
+    // if type is symlink, linkPath is the path of the target file
+    linkPath?: string;
```
benoitf: should we have a ImageFileSymlink interface extending ImageFile and providing the linkPath attribute ?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4290,6 +4290,81 @@ declare module '@podman-desktop/api' {
...
+   * An Image Files provider returns the list of filesystem layers in an image
+   */
+  export namespace imageFiles {
+    /**
+     * Register the extension as an Image Files provider.
+     *
+     * As an image files provider, a provider needs to implement a specific interface, so the core
+     * application can call the provider with specific tasks when necessary.
+     *
+     * @param imageFilesProvider an object implementing the `ImageFilesProvider` interface
```
benoitf: maybe we add `@beta` until it's somehow finalized ?

benoitf (commented):
it's still unclear to me:

- if the namespace is correct (like shouldn't be part of the `provider` namespace)
- the metadata argument as we already provide the object (like why it's not part of the object given)
- if the pattern is `register` where we provide everything or `create` (returning an object with helper methods)

feloy:
> it's still unclear to me:
> 
> * if the namespace is correct (like shouldn't be part of the `provider` namespace)
> * the metadata argument as we already provide the object (like why it's not part of the object given)
> * if the pattern is `register` where we provide everything or `create` (returning an object with helper methods)

I wanted to keep the same pattern as for the image checker API

benoitf:
@feloy but why it can't go to the `provider` namespace vs having to declare a new `imageFiles` namespace ?

feloy:
> @feloy but why it can't go to the `provider` namespace vs having to declare a new `imageFiles` namespace ?

I mean that if we move the function to the `provider` namespace, there will be a specific namespace for the image checker providers `imageChecker`, but for the image files providers, the function will be in the more generic `provider`. Also, if we change from `register` pattern to `create`, there will be 2 different patterns to register image files vs checkers providers.

I would prefer to be consistent between both providers. Or do you think we should, eventually, move the function from the `imageChecker` namespace to the `providers` one?

benoitf:
probably imageChecker should be moved as well

benoitf:
> there will be 2 different patterns to register image files vs checkers providers.

we already have 2 patterns for contributions, as they're bringing different kind of value.

`create` pattern and `register` pattern

if we have common 'helper functions' in all the implementations, probably it's more following the 'create' pattern

`createImageFileProvider(...) : ImageFileProvider`
vs
`registerImageFileProvider(...): Disposable`

feloy:
> > there will be 2 different patterns to register image files vs checkers providers.
> 
> we already have 2 patterns for contributions, as they're bringing different kind of value.
> 
> `create` pattern and `register` pattern
> 
> if we have common 'helper functions' in all the implementations, probably it's more following the 'create' pattern
> 
> `createImageFileProvider(...) : ImageFileProvider` vs `registerImageFileProvider(...): Disposable`

OK, I'll have a try using the `create` pattern.

If I understand correctly, the `ImageFileProvider` would provide a `registerFiles`?

If so, would it be preferable to have an `ImageProvider`, providing `registerFiles`, `registerCheck`, etc?

feloy:
I changed the pattern, for the API to provide helper functions to the extension, to help populate the response to `getFilesystemLayers`
cc @benoitf

benoitf (approved):
thanks for the big changes @feloy 

I think now it's more straightforward

cdrage (approved):
This looks amazing and I have no issues either with the code after going through it, amazing work!

LGTM.

Issue #6789, opened 2024-04-15
Kubernetes Context Page E2E Test implementation


rostalan:
implement Kubernetes Context page E2E tests

rostalan:
Closed via: https://github.com/containers/podman-desktop/pull/7843

PR #7843, opened 2024-06-27
chore(tests): kubernetes contexts e2e tests

rostalan:
### What does this PR do?

Creates new page models for "Kubernetes Contexts" and for "Preferences".
Implemented new test covering the switching and deletion contextts.

Screencast of this PR
### Screenshot / video of UI

[Peek 2024-06-27 20-03.webm](https://github.com/containers/podman-desktop/assets/75624842/efc5bd01-c276-4e3d-8a2c-97ca86a71a22)

Implements [#6789](https://github.com/containers/podman-desktop/issues/6789).

### How to test this PR?

Run `yarn test:e2e`


tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  async contextReachable(name: string): Promise<boolean> {
+    const row = await this.getContextRowByName(name);
+    if (row === undefined) {
+      throw Error(`Context: '${name}' does not exist`);
+    }
+    const contextReachable = row.getByLabel('Context Reachable');
+
+    return contextReachable.isVisible();
```
cbr7: Please use `await contextReachable.count() > 0` for checking if something exists on the page and also the method is not awaited.

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (contextRow === undefined) {
+      throw Error(`Context: '${name}' does not exist`);
+    }
+    const switchButton = contextRow.getByLabel('Set as Current Context');
+    await switchButton.click();
+  }
+
+  async handleConfirmationDialog(): Promise<void> {
+    const confirmationDialog = this.page.getByLabel('Delete Context').filter({ hasText: 'Continue' });
+    if (await confirmationDialog.isVisible()) {
```
cbr7: Please use count() > 0 to check for existence.

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,92 @@
...
+      await confirmationDialog.getByRole('button', { name: 'Yes' }).click();
```
cbr7: Please asset the button is enabled before trying to click it.

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  async deleteContext(name: string): Promise<void> {
+    const contextRow = await this.getContextRowByName(name);
+    if (contextRow === undefined) {
+      throw Error(`Context: '${name}' does not exist`);
+    }
+    await contextRow.getByLabel('Delete Context').click();
```
cbr7: Please check that the button is enabled before clicking.

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,92 @@
...
+    return contextReachable.isVisible();
+  }
+
+  async setDefaultContext(name: string): Promise<void> {
+    const contextRow = await this.getContextRowByName(name);
+    if (contextRow === undefined) {
+      throw Error(`Context: '${name}' does not exist`);
+    }
+    const switchButton = contextRow.getByLabel('Set as Current Context');
+    await switchButton.click();
```
cbr7: Please check that the button is enabled before clicking using `await playExpect(locator).toBeEnabled()`

tests/playwright/resources/test-kube-config
```diff
@@ -0,0 +1,45 @@
+apiVersion: v1
```
odockal, line 17: Please add copyright.

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Locator, Page } from '@playwright/test';
+import { expect as playExpect } from '@playwright/test';
+
+import { SettingsPage } from './settings-page';
+
+export class KubeContextPage extends SettingsPage {
```
odockal: Add any new page/class from the playwright tests into index.ts. Otherwise the class wont be available when `@podman-desktop/tests-playwright` npm package is used outside of the podman-desktop repo.

tests/playwright/src/model/pages/preferences-page.ts
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Locator, Page } from '@playwright/test';
+
+import { SettingsPage } from './settings-page';
+
+export class PreferencesPage extends SettingsPage {
```
odockal, line 24: Also add into index.ts.

tests/playwright/src/model/pages/preferences-page.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  async selectKubeFile(pathToKube: string): Promise<PreferencesPage> {
+    if (!pathToKube) {
+      throw Error(`Path to Kube config file is incorrect or not provided!`);
+    }
+    // TODO: remove this after the https://github.com/containers/podman-desktop/issues/7689 is implemented
+    await this.kubePathInput.evaluate(node => node.removeAttribute('readonly'));
+    await this.kubePathInput.evaluate(node => node.removeAttribute('disabled'));
+    await this.kubePathInput.selectText();
+    await this.kubePathInput.fill(pathToKube);
```
odockal, line 44: Add an assert that text/value is in the locator. we do not click or anything to apply it. so at least that would do.

tests/playwright/src/model/pages/preferences-page.ts
```diff
@@ -0,0 +1,48 @@
...
+    return new PreferencesPage(this.page);
```
odockal: Why do we need to return the page object?

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
+/**********************************************************************
+ * Copyright (C) 2023-2024 Red Hat, Inc.
```
odockal: Just 2024.

tests/playwright/src/model/pages/preferences-page.ts
```diff
@@ -0,0 +1,48 @@
+/**********************************************************************
+ * Copyright (C) 2023-2024 Red Hat, Inc.
```
odockal: just 2024

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+let navBar: NavigationBar;
+
+const testContexts = ['context-1', 'context-2', 'context-3'];
+
+beforeAll(async () => {
+  pdRunner = new PodmanDesktopRunner();
+  page = await pdRunner.start();
+
+  // check that kubectl binary is installed
+  const child = require('node:child_process');
```
odockal: please extract and make more general (if possible) into utilities. could be very useful and used elsewhere.

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  if (child.spawnSync('which', ['kubectl']).status !== 0) {
```
odockal: will it work on windows?

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    throw new Error('kubectl is not installed');
```
odockal: do we need it for kube-context tests?

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    preferencesPage = await preferencesPage.selectKubeFile(kubeConfigPathDst);
+    await playExpect(preferencesPage.heading).toBeVisible();
+  }, 12000);
+
+  test('Check if kubeconfig is loaded correctly', async () => {
+    const settingsBar = await navBar.openSettings();
+    const kubePage = await settingsBar.openTabPage(KubeContextPage);
+    await playExpect(kubePage.heading).toBeVisible();
+
+    playExpect(await kubePage.pageIsEmpty()).toBeFalsy();
```
odockal: better would be usage of playExpect.poll. It runs the inside callback multiple times. Look how it works in other specs.

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    await pdRunner.screenshot('kubeconfig-loaded.png');
```
odockal: remove once we know it works.

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    for (const context of testContexts) {
+      const row = await kubePage.getContextRowByName(context);
+      if (row === undefined) {
+        throw Error(`Context: '${context}' not found`);
+      }
+      playExpect(await row.getByLabel('Context Name', { exact: true }).innerText()).toBe(context);
```
odockal: better to use toHaveText or similar. Same for the rest.

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+      playExpect(await row.getByLabel('Context Cluster').innerText()).toBe(context + '-cluster');
+      playExpect(await row.getByLabel('Context Server').innerText()).toBe(context + '-server');
+      playExpect(await row.getByLabel('Context User').innerText()).toBe(context + '-user');
+
+      if (context === 'context-1') {
+        // check if the context is default
+        playExpect(await kubePage.contextDefault(context)).toBeTruthy();
+        await playExpect(row.getByLabel('Set as Current Context')).not.toBeVisible();
+
+        playExpect(await row.getByLabel('Context Namespace').innerText()).toBe(context + '-namespace');
```
odockal: same as above

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    const settingsBar = await navBar.openSettings();
+    const kubePage = await settingsBar.openTabPage(KubeContextPage);
+    await playExpect(kubePage.heading).toBeVisible();
+
+    await kubePage.setDefaultContext('context-2');
+    // check that switch worked - current context banner should be visible
+    playExpect(await kubePage.contextDefault('context-1')).toBeFalsy();
+    playExpect(await kubePage.contextDefault('context-2')).toBeTruthy();
+    playExpect(await kubePage.contextDefault('context-3')).toBeFalsy();
+    await pdRunner.screenshot('kube-context-switch.png');
```
odockal: also, remove later.

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  }, 12000);
+
+  test('Delete all contexts, check that the page is empty', async () => {
+    const settingsBar = await navBar.openSettings();
+    const kubePage = await settingsBar.openTabPage(KubeContextPage);
+    await playExpect(kubePage.heading).toBeVisible();
+
+    for (const context of testContexts) {
+      await kubePage.deleteContext(context);
```
odockal: I would add an assert that the right one we wanted to delete is not present anymore.

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    await kubePage.setDefaultContext('context-2');
+    // check that switch worked - current context banner should be visible
+    playExpect(await kubePage.contextDefault('context-1')).toBeFalsy();
+    playExpect(await kubePage.contextDefault('context-2')).toBeTruthy();
+    playExpect(await kubePage.contextDefault('context-3')).toBeFalsy();
+    await pdRunner.screenshot('kube-context-switch.png');
+  }, 12000);
+
+  test('Delete all contexts, check that the page is empty', async () => {
```
odockal: I would keep just `Can delete all contexts from Kubernetes Context Page`

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+beforeEach<RunnerTestContext>(async ctx => {
+  ctx.pdRunner = pdRunner;
+});
+
+afterAll(async () => {
+  await pdRunner.close();
+}, 90000);
+
+describe('Verification of kube context management', async () => {
+  test('Load kubeconfig', async () => {
```
odockal: Load custom kubeconfig in Preferences

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    await settingsBar.expandPreferencesTab();
+    let preferencesPage = await settingsBar.openTabPage(PreferencesPage);
+    await playExpect(preferencesPage.heading).toBeVisible();
+
+    const kubeConfigPathDst = path.resolve(__dirname, '..', '..', 'resources', 'kube-config');
+    preferencesPage = await preferencesPage.selectKubeFile(kubeConfigPathDst);
+    await playExpect(preferencesPage.heading).toBeVisible();
+  }, 12000);
+
+  test('Check if kubeconfig is loaded correctly', async () => {
```
odockal: `Can load kube contexts in Kubernetes page`

odockal (changes requested):

tests/playwright/src/specs/kube-contexts.spec.ts
rostalan: should I implement it even though I won't use it in this suite? after reviewing the test cases it is clear that kubectl is not needed at all for any of the tests in currently this suite.

tests/playwright/resources/test-kube-config
rostalan: is it really necessary? it is not present in the other config files in resources, so either it is missing there as well or it is not needed...

ScrewTSW (approved):
LGTM

tests/playwright/resources/test-kube-config
odockal: Yes, the others are missing it ;)

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  async getContextRowByName(name: string): Promise<Locator | undefined> {
+    if (await this.pageIsEmpty()) {
+      return undefined;
+    }
+    return this.contextTable.getByLabel(name, { exact: true });
+  }
+
+  async contextDefault(name: string): Promise<boolean> {
```
odockal, line 48: `setDefaultContext` would be proper name.

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,98 @@
...
+    const contextRow = await this.getContextRowByName(name);
+    if (contextRow === undefined) {
+      throw Error(`Context: '${name}' does not exist`);
+    }
+    const switchButton = contextRow.getByLabel('Set as Current Context');
+    await playExpect(switchButton).toBeEnabled();
+    await switchButton.click();
+  }
+
+  async handleConfirmationDialog(): Promise<void> {
```
odockal: there is `handleConfigrmationDialog` function in the utilities. Please use that one instead of this custom handler. It should totally work.

odockal (changes requested):

odockal:
Quite good. I am interested how it runs on CI. On e2e tests here in this repo, and on windows in e2e. Please test it there.

benoitf:
moving to draft, looks like not ready

rostalan:
Test results: https://github.com/containers/podman-desktop/actions/runs/10812946629
https://github.com/containers/podman-desktop-e2e/actions/runs/10812967881

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+});
+
+test.afterAll(async ({ runner }) => {
+  test.setTimeout(120000);
+  await runner.close();
+});
+
+test.describe.serial('Verification of kube context management', () => {
+  test('Load custom kubeconfig in Preferences', async ({ navigationBar }) => {
+    test.setTimeout(12000);
```
odockal: Timeout is in ms, so this timeout is 12 sec. Default is 60 sec, you sure you want to make it shorter?

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+      }
+      await playExpect(row.getByLabel('Context Name', { exact: true })).toHaveText(context);
+      await playExpect(row.getByLabel('Context Cluster')).toHaveText(context + '-cluster');
+      await playExpect(row.getByLabel('Context Server')).toHaveText(context + '-server');
+      await playExpect(row.getByLabel('Context User')).toHaveText(context + '-user');
+
+      if (context === 'context-1') {
+        // check if the context is default
+        playExpect(await kubePage.contextDefault(context)).toBeTruthy();
+        await playExpect(row.getByLabel('Set as Current Context')).not.toBeVisible();
```
odockal: please place these locators into kube-context page. The page can hold all the possible locators, not matter if they exist/are visible, etc. assertion for their visibility happens in the test. So place them in the page.

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+        await playExpect(row.getByLabel('Context Namespace')).toHaveText(context + '-namespace');
+      } else {
+        await playExpect(row.getByLabel('Set as Current Context')).toBeVisible();
+      }
+    }
+  });
+
+  test('Can switch default context', async ({ navigationBar }) => {
+    test.setTimeout(12000);
```
odockal: same as above

tests/playwright/src/specs/kube-contexts.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    await kubePage.setDefaultContext('context-2');
+    // check that switch worked - current context banner should be visible
+    playExpect(await kubePage.contextDefault('context-1')).toBeFalsy();
+    playExpect(await kubePage.contextDefault('context-2')).toBeTruthy();
+    playExpect(await kubePage.contextDefault('context-3')).toBeFalsy();
+  });
+
+  test('Can delete all contexts from Kubernetes Contexts page', async ({ navigationBar }) => {
+    test.setTimeout(12000);
```
odockal: again.

odockal (changes requested):

odockal:
@rostalan  And then the Kubernetes menu was redesigned: https://github.com/containers/podman-desktop/pull/8833. Please consider the change...

Update: Kubernetes page in settings is not affected, yet, please verify.

tests/playwright/src/specs/kube-contexts.spec.ts
odockal: discussed. it is ok since we do not know where exactly the button will appear.

rostalan:
New tests: https://github.com/containers/podman-desktop-e2e/actions/runs/10990688461 and 
https://github.com/containers/podman-desktop/actions/runs/10990679332

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,92 @@
+/**********************************************************************
```
cbr7: This file needs to be added to the index.ts.

cbr7 (changes requested):
File missing from index.ts

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,92 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Locator, Page } from '@playwright/test';
+import { expect as playExpect } from '@playwright/test';
+
+import { handleConfirmationDialog } from '../../utility/operations';
+import { SettingsPage } from './settings-page';
+
+export class KubeContextPage extends SettingsPage {
```
cbr7: Why does this page inherit from SettingsPage? That doesnt seem correct.

tests/playwright/src/model/pages/kube-context-page.ts
```diff
@@ -0,0 +1,92 @@
...
+    return this.contextTable.getByLabel(name, { exact: true });
+  }
+
+  async isContextDefault(name: string): Promise<boolean> {
+    const row = await this.getContextRowByName(name);
+    if (row === undefined) {
+      throw Error(`Context: '${name}' does not exist`);
+    }
+    const defaultContextBanner = row.getByLabel('Current Context');
+    const bannerText = await defaultContextBanner.textContent();
```
cbr7:
```suggestion
    const bannerText = await row.getByLabel('Current Context').textContent();
```

tests/playwright/src/model/pages/kube-context-page.ts
rostalan: Unlike other Kubernetes pages this one is located inside settings and all the test file pages nested there use SettingsPage.

tests/playwright/src/model/pages/preferences-page.ts
```diff
@@ -0,0 +1,46 @@
...
+    );
+  }
+
+  async selectKubeFile(pathToKube: string): Promise<void> {
+    if (!pathToKube) {
+      throw Error(`Path to Kube config file is incorrect or not provided!`);
+    }
+    await this.kubePathInput.selectText();
+    playExpect(await this.kubePathInput.inputValue()).toBeDefined();
+    await this.kubePathInput.fill(pathToKube);
```
cbr7: I'm not clear on the goal here, what are the first two lines supposed to do, especially the first line? Also in general it's good practice to perform a clear() call before you fill() any textbox.

tests/playwright/src/specs/kubernetes-context-smoke.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+  test('Load custom kubeconfig in Preferences', async ({ navigationBar }) => {
+    // open preferences page
+    const settingsBar = await navigationBar.openSettings();
+    await settingsBar.expandPreferencesTab();
+    const preferencesPage = await settingsBar.openTabPage(PreferencesPage);
+    await playExpect(preferencesPage.heading).toBeVisible();
+
+    const kubeConfigPathDst = path.resolve(__dirname, '..', '..', 'resources', 'kube-config');
+    await preferencesPage.selectKubeFile(kubeConfigPathDst);
+    await playExpect(preferencesPage.heading).toBeVisible();
```
cbr7: Why is this check needed? Aren't you already in preferencesPage as validated on line 54? There doesn't seem to be any navigation between line 54 and 58 where it would make this check required.

tests/playwright/src/specs/kubernetes-context-smoke.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    const settingsBar = await navigationBar.openSettings();
+    const kubePage = await settingsBar.openTabPage(KubeContextPage);
+    await playExpect(kubePage.heading).toBeVisible();
+
+    await playExpect.poll(async () => await kubePage.pageIsEmpty(), { timeout: 10000 }).toBeFalsy();
+    for (const context of testContexts) {
+      const row = await kubePage.getContextRowByName(context);
+      if (row === undefined) {
+        throw Error(`Context: '${context}' not found`);
+      }
```
cbr7: `row` here can't ever be undefined, since the only way it can be undefined is if the page is empty, however you are validating that the page is not empty on line 66 as such `getContextRowByName(context)` will always return a `Locator` object here which will satisfy the contition, note that that does not mean that the context is actually available, only that the Locator is.

```suggestion
      const row = await kubePage.getContextRowByName(context);
      await playExpect(row).toBeVisible();
```

tests/playwright/src/specs/kubernetes-context-smoke.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+      await playExpect(row.getByLabel('Context Server')).toHaveText(context + '-server');
+      await playExpect(row.getByLabel('Context User')).toHaveText(context + '-user');
+
+      if (context === 'context-1') {
+        // check if the context is default
+        await playExpect.poll(async () => await kubePage.isContextDefault(context), { timeout: 10000 }).toBeTruthy();
+        await playExpect(row.getByLabel('Set as Current Context')).not.toBeVisible();
+        await playExpect(row.getByLabel('Context Namespace')).toHaveText(context + '-namespace');
+      } else {
+        await playExpect(row.getByLabel('Set as Current Context')).toBeVisible();
```
cbr7: All of these properties should be exposed from the relevant page object model via methods that accept the context name as a parameter because we might want to check these fields in other tests too not to mention and doing it like this is not scalable. 

For example for `Context Cluster` it should be something like `await kubePage.getContextCluster(context)` which returns a string type.

This way we abstract the looking for the row in question and also the getByLabel() method in the POM and if/when the locators will change we will only have to change them in the POM, not individually in every test that might use them.

tests/playwright/src/specs/kubernetes-context-smoke.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    await playExpect(kubePage.heading).toBeVisible();
+
+    for (const context of testContexts) {
+      // confirmation only pops up if the context is default
+      if (await kubePage.isContextDefault(context)) {
+        await kubePage.deleteContext(context);
+      } else {
+        await kubePage.deleteContext(context, false);
+      }
+      await playExpect(kubePage.heading).toBeVisible();
```
cbr7: Same as above, there doesn't seem to be any navigation between line 103 and this, why are we checking it again here?

tests/playwright/src/specs/kubernetes-context-smoke.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    for (const context of testContexts) {
+      await playExpect
+        .poll(async () => await kubePage.getContextRowByName(context), { timeout: 10000 })
+        .toBeUndefined();
```
cbr7: This approach is vulnerable to a race condition, since getContextRowByName() only returns undefined when the page is empty, however deletion is an async process as such it's possible that deleting the last row will not have been completed before the validation on the first context is done in this check, which would cause it to throw.

```suggestion
      await playExpect(kubePage.getContextRowByName(context)).not.toBeVisible();
```

tests/playwright/src/specs/kubernetes-context-smoke.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // check that the page is empty
+    await playExpect(kubePage.heading).toBeVisible();
```
cbr7: Same as above, no navigation has been performed.

tests/playwright/src/specs/kubernetes-context-smoke.spec.ts
rostalan: True, seems to be leftover from when I wanted to do this by selecting the path in file explorer.

tests/playwright/src/specs/kubernetes-context-smoke.spec.ts
rostalan: The `getContextRowByName(context)` was set to return either Locator or Undefined.  Changed to match as it makes sense to only return Locator.

tests/playwright/src/model/pages/kubernetes-context-page.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  async getContextNamespace(row: Locator): Promise<Locator> {
+    return row.getByLabel('Context Namespace');
+  }
+
+  async getSetCurrentContextButton(row: Locator): Promise<Locator> {
+    return row.getByLabel('Set as Current Context');
+  }
+}
```
cbr7: There's no reason these methods should need to take a locator param nor for that matter should they return a locator param, I understand that it's easier to do this in order to `await playExpect(locator).toHaveText(text)` but the naming of the methods suggests they return the actual value of that field, in this case it should return a string type.

Also it would be even better if you only provided the actual name/context as a param, not a locator.

Additionally given the current implementation why are they even returning a Promise<Locator>? there is nothing async in these, which makes it all the more confusing.

rostalan: I tried to address this, but am not sure if in the correct way. I wasn`t able to find a way to remove the Promises from the returns of these functions as the async construction requires it.

tests/playwright/src/model/pages/kubernetes-context-page.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  async getContextCluster(context: string): Promise<string | null> {
+    const row = await this.getContextRowByName(context);
+    return row.getByLabel('Context Cluster').textContent();
+  }
+
+  async getContextServer(context: string): Promise<string | null> {
+    const row = await this.getContextRowByName(context);
+    return row.getByLabel('Context Server').textContent();
+  }
```
cbr7:
```suggestion
  async getContextServer(context: string): Promise<string> {
    return await this.getContextRowByName(context).getByLabel('Context Server').innerText();
  }
```

using innerText you can avoid the null, since innerText only returns a string, it returns an empty string if the locator has no text.

you can make all the methods the same way.

cbr7 (approved):

cbr7:
Please perform a test run on windows platform in the azure environments to make sure the tests are ok on this branch before merging into main branch.

odockal:
@rostalan @cbr7 I think this is the run: https://github.com/containers/podman-desktop-e2e/actions/runs/11001937127. Linux smoke also passed with new tests. From my POV, everything is fine. @rostalan I would rebase the PR so we have up to date branch with upstream/main, and once the builds are green, we gonna merge. Which would be great.

odockal (approved):
LGTM

PR #7844, opened 2024-06-27
display image files

### LINKED ISSUE #7801: Display the files present in an image

feloy:

### What does this PR do?

Display the files in an image.

Current limitations:

- works only if a single provider is installed
- doc needs to be written to explain the sizes displayed in the layers cards and what does the checkbox `Show layer only`
- design to be discussed

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/9973512/01fe9386-abd4-495b-9447-7c595f53fa65

Fixes #7801 

### How to test this PR?

Start Podman Desktop with the fake extension https://github.com/feloy/podman-desktop-extension-image-files-fake, for example:

```
yarn watch --extension-folder ../podman-desktop-extension-image-files-fake/
```

Go to the details page of an image, and select the Files tab.

Navigate through the layers and see the files in this layer.

A checkbox `Show layer only` gives the choice to:
- (off) display the state of the filesystem at this layer (considering all the layers up to this one)
- (on) display only the files added/deleted/modified in this layer
 
### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -477,6 +477,11 @@ export class ColorRegistry {
       light: colorPalette.purple[200],
     });
 
+    this.registerColor(`${ct}card-selected-bg`, {
+      dark: colorPalette.charcoal[400],
+      light: colorPalette.purple[100],
+    });
+
     this.registerColor(`${ct}card-text`, {
       dark: colorPalette.gray[400],
       light: colorPalette.purple[900],
--- a/packages/renderer/src/lib/image/FilesystemLayerView.svelte
+++ b/packages/renderer/src/lib/image/FilesystemLayerView.svelte
@@ -1,0 +1,87 @@
+<script context="module" lang="ts">
+const expansionState = new Map<string, boolean>();
+</script>
+
+<script lang="ts">
+import type { ImageFile, ImageFileSymlink } from '@podman-desktop/api';
+
+import type { FilesystemNode } from './filesystem-tree';
+import { ImageUtils } from './image-utils';
+import { isExec, modeString } from './imageDetailsFiles';
+
+export let tree: FilesystemNode<ImageFile>;
+export let margin = 0;
+export let root = true;
+export let layerMode = false;
+
+$: label = tree.name;
+$: children = tree?.children;
+$: file = tree?.data;
+$: colorClass = getColor(tree);
+function getColor(tree: FilesystemNode<ImageFile>) {
+  if (tree.hidden) {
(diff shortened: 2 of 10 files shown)


benoitf:
just wanted to confirm if we proceed with the UI here or if it goes to a self-contained extension providing its own webview ?

deboer-tim:
Personally I think we should provide the ability for extensions to extend details pages with extra tabs using webviews, and things like this, image checker, etc could be provided as extensions unless/until there's a reason to integrate, like if it made sense having multiple providers integrated on the same UI screen. Even developing as a built-in extension would ensure clean arch/API/separation vs continually adding UI to Podman Desktop for one extension to use.

That said, holding this up for a core feature would be a huge impediment so I am fine moving forward as-is. 👍🏼

jeffmaury:
> Personally I think we should provide the ability for extensions to extend details pages with extra tabs using webviews, and things like this, image checker, etc could be provided as extensions unless/until there's a reason to integrate, like if it made sense having multiple providers integrated on the same UI screen. Even developing as a built-in extension would ensure clean arch/API/separation vs continually adding UI to Podman Desktop for one extension to use.
> 
> That said, holding this up for a core feature would be a huge impediment so I am fine moving forward as-is. 👍🏼

I think it depends on the nature of the extension: for the image checker, you may have several implementations but I think it does not make much sense that each extension provide its own UI. In this case, I think there will be a single extension so the 2 options are opened

deboer-tim:
> I think it depends on the nature of the extension: for the image checker, you may have several implementations but I think it does not make much sense that each extension provide its own UI. In this case, I think there will be a single extension so the 2 options are opened

Yeah, totally agreed, but even in this case there is no difference until there are 2 providers. IMHO the 'ideal progression' in cases like this would be starting as a standalone extension until it's clear there are multiple providers and one UI makes sense, and even then it could probably be done via shared extension now. In practice (and even 2 months ago) that's not always feasible, but we need to avoid everything needing to be done in renderer. (and again, not for this PR :) )

benoitf (approved):
works for me as first iteration

deboer-tim (approved):
Ok merging, but please open up a followup issue to replace hardcoded colors. Especially ones like gray-900, gray-700, but should do layer colors too.

PR #7850, opened 2024-06-28
fix: throw an error when override for a module requires extension path

### LINKED ISSUE #7833: extensions using npm package with dependency on `http` leads to crash
### Bug description

image

error is raised here https://github.com/containers/podman-desktop/blob/7d997640c368bb62037bed8ed337c732d78b3484/packages/main/src/plugin/module-loader.ts#L90

### Operating system

Windows 11

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

Add `@kubernetes/client-node` npm package to any extensions, import it and use it (avoid tree shaking).

### Relevant log output

_No response_

### Additional context

cc @benoitf

dgolovin:
### What does this PR do?

There are two types of module override:
* function that get extension metadata as a parameter and returns a module (usually just returns `api` property of an extension object)
* module name to overridden module object mapping

PR changes module loader logic. If override is a function then module loading should be requested from extension folder or exception to be thrown. If override is a mapping, then parent.path is used as a key for chache.

### Screenshot / video of UI

n/a

Fix #7833

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/module-loader.ts
+++ b/packages/main/src/plugin/module-loader.ts
@@ -64,31 +64,28 @@ export class ModuleLoader {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       this._module._load = function load(request: string, parent: any): any {
         const override = overrides.get(request);
-        if (!override) {
-          // eslint-disable-next-line prefer-rest-params,prefer-spread
-          return internalLoad.apply(this, arguments);
-        }
-
         if (override) {
-          const ext = Array.from(analyzedExtensions.values()).find(extension =>
-            path.normalize(parent.filename).startsWith(path.normalize(extension.path)),
-          );
-          if (ext) {
-            if (override instanceof Function) {
+          if (override instanceof Function) {
+            const ext = Array.from(analyzedExtensions.values()).find(extension =>
+              path.normalize(parent.filename).startsWith(path.normalize(extension.path)),
+            );
+            if (ext) {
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/module-loader.ts
```diff
@@ -64,31 +64,28 @@ export class ModuleLoader {
...
-            let cache = extModuleCache.get(ext.path);
-            if (!cache) {
-              extModuleCache.set(ext.path, (cache = {}));
-            }
-            if (!cache[request]) {
-              // eslint-disable-next-line @typescript-eslint/no-explicit-any
-              cache[request] = <any>{ ...override };
-            }
-            return cache[request];
+            throw Error(`Cannot find extension for ${parent.path}`);
```
benoitf, line 75: how do you handle the fact that in dev mode a library can be inside `root-folder-extension/node_modules` while extension backend is inside `root-dolder-extension/packages/backend`

(like adding `[kubernetes/client-node](https://www.npmjs.com/package/@kubernetes/client-node)` in ai-lab backend)

dgolovin: The override for http and https does not need search for extension, because it is direct mapping 
module name => overridden module object, so it is just using parent.path to cache it and then return the overridden module from cache.

axel7083 (approved):
Great job, it fixes the problem raised ! Thanks for the reactivity

Issue #7860, opened 2024-06-29
Drop Playwright raw traces folder from the stored test artifacts on GitHub Actions to reduce size of the archive


odockal:
### Is your enhancement related to a problem? Please describe

We are reaching 500 MB of the archive size. It takes time to download, extract, etc.

### Describe the solution you'd like

Easy was to reduce the amount is not to upload raw traces since we already have traces for all tests in archives so this is a redundancy.

PR #7863, opened 2024-06-30
chore(ci): exclude playwright raw traces from archived artifacts

odockal:
### What does this PR do?
Prevents uploading of raw playwright traces, we have traces in archives already.
#7860

### How to test this PR?

The test artifacts now should be smaller (previous size around 180 MB, now should be less) and there should not be `**/tests/output/traces/raw` folder or its content.

### DIFF
--- a/.github/workflows/e2e-main.yaml
+++ b/.github/workflows/e2e-main.yaml
@@ -101,3 +101,4 @@ jobs:
           path: |
             ./tests/output/
             ./tests/**/output/junit*.xml
+            !./tests/**/traces/raw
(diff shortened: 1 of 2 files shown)


benoitf (approved):
should we just avoid to generate them on GHA based on a flag ?

odockal:
> should we just avoid to generate them on GHA based on a flag ?

In cases where the test fails, it is super beneficial for debugging. Although, we could probably, based on the results of the test (passed) remove the traces. I will create an issue to tackle that.

odockal:
@benoitf Here we go: https://github.com/containers/podman-desktop/issues/7864

PR #7867, opened 2024-07-01
feat: allow to handle update of extensions only for specific range of Podman Desktop

### LINKED ISSUE #7663: Podman Desktop should check if the version is compliant before trying to update

today, Podman Desktop updates extensions to their latest published version.

But if an extension requires an higher version, Podman Desktop should not update to this version

1. enhance catalog with metadata (checking metadata contained in the extension metadata)
1. check field `podmanDesktopVersion` if present.

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
for now, when there is a new version of an extension it is automatically updating the version
here we're bringing another level of control where we can flag extensions requiring a given minimal version of Podman Desktop to do the update. (updating metadata of the catalog)

fixes https://github.com/containers/podman-desktop/issues/7663

### How to test this PR?


packages/main/src/plugin/extensions-updater/extensions-updater.ts
```diff
@@ -137,8 +139,26 @@ export class ExtensionsUpdater {
...
+          } else {
+            // if no version is specified, keep the version
+            return true;
+          }
+        });
+
+        const filteredPreviewVersions = availableVersions.filter(version => version.preview === false);
         // take latest version
-        const latestAvailableVersion = availableExtension.versions.filter(version => version.preview === false)?.[0];
+        const latestAvailableVersion = filteredPreviewVersions?.[0];
```
axel7083, line 161: I am wondering if we should not display the latest version but a warning it cannot be updated, this would be a nice incentive to nudge use to update their podman desktop version

benoitf: users already have nudge if they're not using the latest Podman Desktop so it looks like a dup ?

axel7083: I was thinking if they uses the `Stop remind me` option, then we have nothing telling them to update nicely. But even if they do have it, showing only compatible latest seem counter productive, as we probably do not want to promote older version

axel7083: Moreover, user could be misleaded, they would think they use the latest version, when they are using the latest version they are compatible with

benoitf: why would they be misleaded. the version that they're using is being displayed

Also it needs to be thought in a different way:
if we release a new AI Lab version, it will be updated on all Podman Desktop instances and it won't work if there are some API changes
You want to have automatic updates only if it will work.

axel7083: Oh I don't want automatic update ! Sorry for the confusion, i forget we automatically update it, i thought we were simply displaying the latest version, and update on user action

benoitf: here it's the code for the automatic update

jeffmaury (approved):
LGTM. Seems this PR only deals with automatic updates. Is manual update handled in a separate issue/PR

benoitf:
@jeffmaury there is no manual update supported right now so yes it would come separately

PR #7874, opened 2024-07-01
chore(tests): delete traces when not needed

### LINKED ISSUE #7864: Drop playwright traces in case when e2e test passes

Playwright traces and additional screenshots/videos of the e2e tests take a lot of space. 

We can introduce simple logic that could drop the traces (maybe also videos) in cases where test is passing.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Adds logic to delete traces and videos for passed e2e tests.

https://github.com/containers/podman-desktop/issues/7864

### DIFF
--- a/tests/playwright/src/runner/podman-desktop-runner.ts
+++ b/tests/playwright/src/runner/podman-desktop-runner.ts
@@ -40,7 +40,7 @@ export class PodmanDesktopRunner {
   private _videoAndTraceName: string | undefined;
   private _autoUpdate: boolean;
   private _autoCheckUpdate: boolean;
-  private _testStatus: boolean;
+  private _testPassed: boolean;
 
   constructor({
     profile = '',
@@ -55,7 +55,7 @@ export class PodmanDesktopRunner {
     this._videoAndTraceName = undefined;
     this._autoUpdate = autoUpdate;
     this._autoCheckUpdate = autoCheckUpdate;
-    this._testStatus = true;
+    this._testPassed = true;
 
     // Options setting always needs to be last action in constructor in order to apply settings correctly
     this._options = this.defaultOptions();
@@ -214,13 +214,17 @@ export class PodmanDesktopRunner {
       rmSync(rawTracesPath, { recursive: true, force: true, maxRetries: 5 });
     }
 
-    if (this._testStatus && !process.env.KEEP_TRACES_ON_PASS && this._videoAndTraceName) {
+    if (!this._testPassed || !this._videoAndTraceName) return;
+
+    if (!process.env.KEEP_TRACES_ON_PASS) {
       const tracesPath = join(this._testOutput, 'traces', `${this._videoAndTraceName}_trace.zip`);
       if (existsSync(tracesPath)) {
         console.log(`Removing traces folder: ${tracesPath}`);
         rmSync(tracesPath, { recursive: true, force: true, maxRetries: 5 });
       }
+    }
 
+    if (!process.env.KEEP_VIDEOS_ON_PASS) {
       const videoPath = join(this._testOutput, 'videos', `${this._videoAndTraceName}.webm`);
       if (existsSync(videoPath)) {
         console.log(`Removing video folder: ${videoPath}`);
@@ -315,12 +319,12 @@ export class PodmanDesktopRunner {
     return this._testOutput;
   }
 
-  public getTestStatus(): boolean {
-    return this._testStatus;
+  public getTestPassed(): boolean {
+    return this._testPassed;
   }
 
-  public setTestStatus(value: boolean): void {
-    this._testStatus = value;
+  public setTestPassed(value: boolean): void {
+    this._testPassed = value;
   }
 
   public get options(): object {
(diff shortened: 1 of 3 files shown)


tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -200,6 +203,30 @@ export class PodmanDesktopRunner {
...
+      const tracesPath = join(this._testOutput, 'traces', `${this._videoAndTraceName}_trace.zip`);
+      if (existsSync(tracesPath)) {
+        console.log(`Removing traces folder: ${tracesPath}`);
+        rmSync(tracesPath, { recursive: true, force: true, maxRetries: 5 });
+      }
+
+      const videoPath = join(this._testOutput, 'videos', `${this._videoAndTraceName}.webm`);
+      if (existsSync(videoPath)) {
+        console.log(`Removing video folder: ${videoPath}`);
+        rmSync(videoPath, { recursive: true, force: true, maxRetries: 5 });
```
ScrewTSW, line 231: In my opinion it's still useful to have the video recording even if the test passes, to be able to rule out false negatives and general debugging

ScrewTSW (commented):
Generally I'm not opposed, but I'd like my comment addressed

tests/playwright/src/runner/podman-desktop-runner.ts
benoitf: a `KEEP_VIDEOS_ON_PASS` env var ?

cbr7: I mean I can change it to use a different env var for video if you want but I thought using two different ones seems like a lot and @odockal wanted to have the videos removed by default also in order to have the final trace files as small as possible.

@odockal what do you think?

odockal: I would not mind I guess. Also, from my point of view I do not recall opening videos or even downloading artifacts from passed tests. On the other hand, when you develop new test, it might be useful. So let's have more fined grained options. It is not really causing any harm. but I would not keep it as default.

odockal: Example for the smoke run: Videos are 20 MB, traces (just zips) are 90 MB. so it is roughly 1:4,5 rate videos x traces. I will leave the decision on you, guys. If @ScrewTSW wants it since he is using it, I would not mind.

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -175,6 +177,7 @@ export class PodmanDesktopRunner {
     const video = this.getPage().video();
     if (video) {
       await video.saveAs(path);
+      await video.delete();
```
odockal, line 180: This will delete the `hash.webm` file and keep named file?

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -200,6 +203,30 @@ export class PodmanDesktopRunner {
...
+
+  async removeTracesOnFinished(): Promise<void> {
+    const rawTracesPath = join(this._testOutput, 'traces', 'raw');
+
+    if (existsSync(rawTracesPath)) {
+      console.log(`Removing raw traces folder: ${rawTracesPath}`);
+      rmSync(rawTracesPath, { recursive: true, force: true, maxRetries: 5 });
+    }
+
+    if (this._testStatus && !process.env.KEEP_TRACES_ON_PASS && this._videoAndTraceName) {
```
odockal: I would put second and third condition into subconditions like this:
```
if (testPassed) {
  if (KEEP_TRACES)
  if (KEEP_VIDEO)
}
```
or something like this.

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -54,6 +55,7 @@ export class PodmanDesktopRunner {
     this._videoAndTraceName = undefined;
     this._autoUpdate = autoUpdate;
     this._autoCheckUpdate = autoCheckUpdate;
+    this._testStatus = true;
```
odockal: maybe better naming? Like testFailed, testPassed ?

odockal (changes requested):

tests/playwright/src/runner/podman-desktop-runner.ts
cbr7: yes

odockal: Awesome!

tests/playwright/src/runner/podman-desktop-runner.ts
cbr7: Added a new env var named KEEP_VIDEOS_ON_PASS to enable video saving separate from traces.

tests/playwright/src/runner/podman-desktop-runner.ts
cbr7: done.

tests/playwright/src/runner/podman-desktop-runner.ts
cbr7: changed named to `_testPassed` for the field.

odockal (approved):
Nice. Thanks!

Issue #971, opened 2022-12-05
Ability to browse images on remote repositories


benoitf:
### Is your enhancement related to a problem? Please describe

Today, to start a container, we need to pull an image.

But then we need to know the exact name of this image.

### Describe the solution you'd like

I would like to be able to browse images and select one to pull or create a container from there.

### Additional context

https://github.com/containers/podman-desktop/issues/759

feloy:
As discussed in https://github.com/containers/podman-desktop/pull/7875, the requirements are more important, as we also want to browse the images of a registry, independently of any container engine

feloy:
Searching and listing images in an image registry is done using the `/v1/search` and `/v2/_catalog` of the Registry API.

Go implementation (for reference): https://github.com/containers/image/blob/main/docker/docker_client.go#L355

Node.js implementations:
- https://github.com/TritonDataCenter/node-docker-registry-client
-

benoitf:
I think we've some code in `image-registry.ts` (to handle token used to communicate with the API) (so probably it should go to this location)

PR #7875, opened 2024-07-01
API provides ability to browse images on remote repositories

feloy:

### What does this PR do?

Adds a `searchImages` function to the API

Spec: https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImageSearchLibpod

TODO: use Docker API instead? (https://docs.docker.com/engine/api/v1.46/#tag/Image/operation/ImageSearch)

Part of #971 

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -3417,6 +3417,71 @@ declare module '@podman-desktop/api' {
   }
 
   /**
+   * Options when searching images in registries
+   */
+  export interface SearchImagesOptions {
+    /**
+     * Term to search, in the form `[<registry>/]searchTerm`. For example, `quay.io/http` to search images with `http` search term in `quay.io` registry
+     */
+    term: string;
+    /**
+     * Maximum number of results to return. Defaults to 25
+     */
+    limit?: number;
+    /**
+     * If true, the `Tag` field of the result will be filled. Defaults to false
+     */
+    listTags?: boolean;
+    /**
+     * If true, require HTTPS and verify signatures when contacting registries. Defaults to true
+     */
+    tlsVerify?: boolean;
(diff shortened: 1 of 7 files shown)


benoitf:
I'm wondering why we do have the Search API usage of podman ?

AFAIK I want to be able to browse and search images without a container engine running.

benoitf:
also why it's exposed in the extension API. I'm not even sure 3rd party extensions might want to search something ?

feloy:
> AFAIK I want to be able to browse and search images without a container engine running.

In #971, the finality is to pull the image, operation for which we need a container engine. Or do we want to browse / search files in a more generic way?

benoitf:
> In https://github.com/containers/podman-desktop/issues/971, the finality is to pull the image, operation for which we need a container engine. Or do we want to browse / search files in a more generic way?

not sure about browsing files but for sure it's to browse images

IMHO we want to browse images like we're checking the auth of registries, it means between Podman Desktop and the registries, not using a 3rd party engine.

And you want to see the number of pull of images, restrict some registries

feloy:
> > In #971, the finality is to pull the image, operation for which we need a container engine. Or do we want to browse / search files in a more generic way?
> 
> not sure about browsing files but for sure it's to browse images
> 

My bad, I wanted to say images

> IMHO we want to browse images like we're checking the auth of registries, it means between Podman Desktop and the registries, not using a 3rd party engine.
> 

OK, that changes the user story. I'll check for a way to do this without any container engine
 
> And you want to see the number of pull of images, restrict some registries

feloy:
closing as we don't want to use container engine

benoitf:
I think at some point also, we might want also to suggest ubi images if you search for httpd as for example httpd can't run on OpenShift (due to the non-root)

PR #7878, opened 2024-07-01
feat: add option to require user confirmation for certain actions

### LINKED ISSUE #7832: Add settings for confirmation

For some usecases, there is a need to confirm like
https://github.com/containers/podman-desktop/issues/7591

but user should be able to disable the confirmation and just having clicking proceeding with the action

This issue is to introduce a flag for 'user confirmation' that will be used for example by https://github.com/containers/podman-desktop/issues/7591

_No response_

### Additional context

_No response_

SoniaSandler:
Adds a new setting that allows users to require user confirmation before bulk actions

###  Screenshot / video of UI

Screenshot from 2024-07-03 17-58-16

###  What issues does this PR fix or reference?

Closes https://github.com/containers/podman-desktop/issues/7832

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/confirmation-init.spec.ts
+++ b/packages/main/src/plugin/confirmation-init.spec.ts
@@ -1,0 +1,50 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { expect, test, vi } from 'vitest';
+
+import type { ConfigurationRegistry } from './configuration-registry.js';
+import { ConfirmationInit } from './confirmation-init.js';
+
+let confirmationInit: ConfirmationInit;
+
+const registerConfigurationsMock = vi.fn();
+
+const configurationRegistryMock = {
+  registerConfigurations: registerConfigurationsMock,
+} as unknown as ConfigurationRegistry;
+
+test('should register a configuration', async () => {
+  confirmationInit = new ConfirmationInit(configurationRegistryMock);
+
+  confirmationInit.init();
+
+  expect(configurationRegistryMock.registerConfigurations).toBeCalled();
+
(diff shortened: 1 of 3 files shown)


lstocchi (approved):
LGTM

deboer-tim:
IMHO this is too broad and should at least be scoped to deletion. As a user I don't know what 'certain actions' are and what this setting would do, and if we had certain actions that aren't deletion (things that require root/admin access?) I'd probably want it to be a separate setting.

benoitf:
>  aren't deletion (things that require root/admin access?) 

in that case you're always prompted as you need to either set the admin password or the fingerprint

benoitf (commented):
it is missing a unit test

benoitf (commented):
replace "certain confirmation" by either just 'confirmation' or we use 'bulk modifications'

deboer-tim:
> replace "certain confirmation" by either just 'confirmation' or we use 'bulk modifications'

That's exactly why it should be renamed, I thought 'certain actions' was implying single deletions as well. ;-)

I'd also recommend using an id like 'preferences.confirmation.bulk', since we've had others requested and if we do them then things like 'preferences.confirmation.delete' or 'preferences.confirmation.prune' would be similar.

SoniaSandler:
@deboer-tim I've seen that in other settings, the  id of the entire configuration node is something like `preferences.userConfirmation`, and then the ids of the properties are more like `userConfirmation.bulk`. is this okay for this setting?

deboer-tim:
Node is ok, it's the property that's saved to file which I would change. With the current PR the settings file would have:
`"preferences.userConfirmation": false`
which makes it harder to add peer preferences.

SoniaSandler:
@benoitf I've added a unit test for this setting (used `plugin/open-devtools-init.spec.ts` as a template). Since the file I added is just the setting itself, without any usage or actions yet, I didn't know what else I can test except for the fields' values. If you have any other suggestions as to what I can check or change, please lmk.

benoitf (approved):
thanks @SoniaSandler for the unit test

from english POV, should it be `modifications` or `changes` ?

deboer-tim:
> from english POV, should it be `modifications` or `changes` ?

From a natural language point of view both read fine, but if you google what the actual difference is you get variations of: `Both refer to altering something. However, 'modify' refers to slight changes for improvements, whereas, 'change' indicates transforming the nature of something to something completely new.`

Since we're mostly talking about delete here, 'change' is more correct. IMHO 'operations' or 'actions' would also be fine, and maybe 'actions' is slightly better? Or since we're adding this for delete we could just limit it to 'delete' for now.

FWIW, I'd just take modifications out of the preference `userConfirmation.bulkModifications` (i.e. just end in `.bulk`, or `.bulk.delete` if we limit the scope) vs changing it to match the text, since we never know if we'd change the text again in the future.

SoniaSandler:
@benoitf @deboer-tim I've changed the property id to `.bulk`, so now the title of the setting is just `Bulk`
Is this okay, or should it be something more specific/descriptive for the users?

Screenshot from 2024-07-05 09-06-03

benoitf:
I'm fine with it

PR #7893, opened 2024-07-02
chore: process podman system and user configuration

### LINKED ISSUE #7027: Look for containers conf files in the same folders as `podman` does

Podman Desktop looks for the containers conf file to find the current machine provider. It uses that to decide which set of machine config features to expose to the user. 

The problem is that containers conf can be in folders that are currently ignored by Podman Desktop.

Podman Desktop should look for containers conf file in different folders [following the same logic as `podman`](https://github.com/containers/common/blob/main/pkg/config/new.go).

### Additional context

As an example [after this windows installer update](https://github.com/containers/podman/pull/22506), `podman` installation will create a configuration in `%APPDATA%/containers/containers.conf.d/99-podman-machine-provider.conf` (and later in `%PROGRAMDATA%/containers/containers.conf.d/99-podman-machine-provider.conf`, see [issue](https://github.com/containers/podman/issues/22411)). If Podman Desktop ignores the file it will assume that the provider is `wsl` although podman machine is configured to use `hyperv`.

vzhukovs:
### What does this PR do?

The purpose of this changes request is to provide the complex mechanism to process Podman system and user based configuration. According to the latest changes in Podman [new.go](https://github.com/containers/common/blob/main/pkg/config/new.go) there is a need to implement the same logic in Podman Desktop to process these configuration files.

### Screenshot / video of UI

No screenshots. Changes are invisible from the user perspective.

#7027 

### How to test this PR?

Podman Desktop should take into account Podman system-based and user-based configuration. So, no visual check on the UI. The only thing is to check, that the current PR works the same as https://github.com/containers/podman-desktop/pull/7881 and https://github.com/containers/podman-desktop/pull/7540 intended.

### DIFF
--- a/extensions/podman/packages/extension/src/podman-configuration.ts
+++ b/extensions/podman/packages/extension/src/podman-configuration.ts
@@ -16,27 +16,295 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import type { PathLike } from 'node:fs';
 import * as fs from 'node:fs';
+import { promises as fsPromises } from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
+import { resolve } from 'node:path';
 
 import type { ProxySettings } from '@podman-desktop/api';
 import * as extensionApi from '@podman-desktop/api';
 import { Mutex } from 'async-mutex';
 import * as toml from 'smol-toml';
 
-const configurationRosetta = 'setting.rosetta';
+export function merge(source: Record<string, unknown>, target: Record<string, unknown>): Record<string, unknown> {
+  const result: Record<string, unknown> = { ...source };
+
+  for (const key in target) {
+    if (Object.prototype.hasOwnProperty.call(target, key)) {
+      if (typeof target[key] === 'object' && !Array.isArray(target[key]) && target[key] !== undefined) {
+        if (typeof result[key] === 'object' && !Array.isArray(result[key]) && result[key] !== undefined) {
+          result[key] = merge(result[key] as Record<string, unknown>, target[key] as Record<string, unknown>);
+        } else {
+          result[key] = { ...(target[key] as Record<string, unknown>) };
+        }
+      } else {
+        result[key] = target[key];
+      }
+    }
+  }
+
+  return result;
+}
 
 /**
- * Manages access to the containers.conf configuration file used to configure Podman
+ * Provides the ability to operate with system-wide and user based Podman properties file.
+ * There is a particular implementation for each supported os. Not intended to be directly created.
+ * @see {PodmanConfigurationProvider}
+ * @see {UnixPodmanConfiguration}
+ * @see {WindowsPodmanConfiguration}
  */
-export class PodmanConfiguration {
-  private mutex: Mutex = new Mutex();
-  async init(): Promise<void> {
-    let httpProxy = undefined;
-    let httpsProxy = undefined;
-    let noProxy = undefined;
+export abstract class PodmanConfiguration {
+  protected configPath = path.join('containers', 'containers.conf');
+  protected userOverrideContainersConfig = path.join('.config', this.configPath);
 
+  protected userConfigPath(): PathLike {
+    const xdgRuntimeDir: PathLike | undefined = process.env.XDG_RUNTIME_DIR;
+    if (xdgRuntimeDir !== undefined) {
+      const xdgConfigPath = path.join(xdgRuntimeDir, this.configPath);
+      if (this.isValid(xdgConfigPath)) {
+        return xdgConfigPath;
+      }
+    }
+
+    return path.join(os.homedir(), this.userOverrideContainersConfig);
+  }
+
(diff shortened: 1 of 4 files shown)


extensions/podman/src/extension.ts
```diff
@@ -1603,7 +1611,7 @@ export async function calcPodmanMachineSetting(podmanConfiguration: PodmanConfig
 
   if (isWindows()) {
     const isPodmanHyperv_Env = process.env.CONTAINERS_MACHINE_PROVIDER === 'hyperv';
-    const isPodmanHyperv_Config = await podmanConfiguration.matchRegexpInContainersConfig(/provider\s*=\s*"hyperv"/);
+    const isPodmanHyperv_Config = await new HyperVConfiguration(podmanConfiguration).isHyperVEnabled();
```
lstocchi: Just tried on Windows and i saw that when calling the `isHyperVEnabled()` the configuration file which is used to verify if hyperv is enabled is `.config/containers/containers.conf`. However if i check my filesystem i see that i don't have any file in that location but i have one in `C:\ProgramData\containers\containers.conf.d\...`. Is it missing something?

benoitf: PTAL @vzhukovs

vzhukovs: @lstocchi @benoitf the logic of processing configuration files is the following and completely repeats the logic that is implemented in podman:

0) creating empty stack of file list
1) we are looking to env variable `CONTAINERS_CONF` and if exists, push to the stack
2) look for the default containers config file, which is empty in windows, so this step skips
3) look for overriden container config file, on windows it is located in `%PROGRAMDATA%\containers\containers.conf`, push to the stack tail
3.1) look for the directory `%PROGRAMDATA%\containers\containers.conf.d` and find any `*.conf` files in it, push them to the stack tail
4) look for user config in `%APPDATA%\containers\containers.conf`, push to the stack tail
4.1) look for the directory `%APPDATA%\containers\containers.conf.d`, do the same as in 3.1
5) create an empty config object and iterating from the stack head to the stack tail by merging every configuration with created config object. merging - means, that if in user configuration there is a parameter, e.g. hyperv = true, it will override system one. 

can you provide more details, which config files do you have in system?

vzhukovs:
@jeffmaury the PR is ready to be reviewed again

extensions/podman/packages/extension/src/podman-configuration.ts
```diff
@@ -16,56 +16,349 @@
...
+  defaultContainersConfig(): PathLike {
+    return path.join('/usr/share', this.configPath);
+  }
+
+  overrideContainersConfigPath(): PathLike {
+    return path.join('/usr/local/etc', this.configPath);
+  }
+}
+
+export class LinuxPodmanConfiguration extends PodmanConfiguration {
```
jeffmaury: This class is duplicated from DarwinPodmanConfiguration they should be merged

jeffmaury (changes requested):

extensions/podman/packages/extension/src/podman-configuration.ts
vzhukovs: Merged

odockal:
I have added `qe/testing-required` label since I think we need to make sure not to introduce regression. I have run windows e2e tests with the branch:
* hyperv provider: https://github.com/podman-desktop/e2e/actions/runs/12749549135 (1 failure on win 10 - investigating...)
* wsl: https://github.com/podman-desktop/e2e/actions/runs/12749557340

Would be great to have also Mac OS tested since I think mac is also impacted, @cbr7 PTAL, please?

cbr7:
I've ran all the e2e tests on mac for both applehv and libkrun podman machines and the tests ran fine, seems good to me.

cbr7 (approved):

odockal (approved):
For the rosetta enablement re-verification and the preferences value change -> quite hard to achieve to uninstall rosetta to verify first part and for the second I think we have sufficient unit tests. 
For changes in Proxy configuration process in podman extension, unfortunately we have not yet automated the proxy setup, so I cannot tell on that side.

Regression tests on win:
> hyperv provider: https://github.com/podman-desktop/e2e/actions/runs/12749549135 (1 failure on win 10 - investigating...)

The failure seems just like a timeout. so no regression that automation would point to.

LGTM.

@vzhukovs great work!

benoitf (commented):
existing requested for changes

vzhukovs:
@jeffmaury can we proceed with merge?

benoitf:
PTAL @jeffmaury

jiridostal:
@vzhukovs can you please have a look at the conflicts so it can be merged?

vzhukovs:
@jiridostal this PR needs to be rewritten from scratch, because there were a few breaking changes in this functionality, so will take a look on it asap, maybe there will be a separate PR and this will be closed.

cdrage:
I'm going to mark this as draft so people do not get pinged to review the PR until conflicts have been updated.

benoitf:
closing as not updated

PR #7900, opened 2024-07-02
chore: migrate StatusIcon to ui library

### LINKED ISSUE #7899: Migrate `StatusIcon` to the UI library

AI-Lab still has its own implementation of the StatusIcon component, and should uses the one from podman-desktop as it is just a copy 

Avoid https://github.com/containers/podman-desktop-extension-ai-lab/issues/1271

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Moving the `StatusIcon` ui component to ui package as used by extensions with webview.

Fixes https://github.com/containers/podman-desktop/issues/7899
Required for https://github.com/containers/podman-desktop-extension-ai-lab/issues/1271

### How to test this PR?

- Pipeline should be ✅


benoitf (commented):
I thought you brought icons to `/icons` namespace

benoitf (commented):
the naming maybe is causing issues

because we do have icons that are just svg file and this one being a component but they do have all the same name pattern

mynameIcon

axel7083:
> I thought you brought icons to `/icons` namespace

It is not an icon, but a component with some logic, therefore I don't think it should go into the /icons namespace

> the naming maybe is causing issues
>because we do have icons that are just svg file and this one being a component but they do have all the same name pattern
> mynameIcon

Locally I added the export in the dist folder instead of packages/ui/src/index.ts...

benoitf (approved):
I still don't know if we should not pickup another name to not let think it's an 'icon' but anyway it's a move and the component is already named like that

feloy (approved):

PR #7905, opened 2024-07-02
chore: light mode for feedback form

### LINKED ISSUE #7544: Light mode for feedback form

The feedback form has some hardcoded colors for the text box and email address.

Would be better to have components that set the correct colors, but for now it should be enough to use existing colors from the color registry to correct this.

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?

Removes hard-coded colors from the feedback form and uses the color registry

### Screenshot / video of UI

Screenshot from 2024-07-02 16-06-19

Screenshot from 2024-07-03 11-02-02

Closes https://github.com/containers/podman-desktop/issues/7544

### How to test this PR?

Open the feedback form in dark and light modes

### DIFF
--- a/packages/renderer/src/lib/feedback/SendFeedback.svelte
+++ b/packages/renderer/src/lib/feedback/SendFeedback.svelte
@@ -71,25 +71,33 @@ async function sendFeedback(): Promise<void> {
           <button aria-label="very-sad-smiley" on:click="{() => selectSmiley(1)}">
             <Fa
               size="1.5x"
-              class="cursor-pointer {smileyRating === 1 ? 'text-violet-400' : 'text-gray-900'}"
+              class="cursor-pointer {smileyRating === 1
+                ? 'text-[var(--pd-button-primary-bg)]'
+                : 'text-[var(--pd-button-disabled-text)]'}"
               icon="{faFrown}" />
           </button>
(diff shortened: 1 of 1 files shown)


lstocchi (approved):
LGTM

packages/renderer/src/lib/feedback/SendFeedback.svelte
```diff
@@ -71,25 +71,33 @@ async function sendFeedback(): Promise<void> {
           <button aria-label="very-sad-smiley" on:click="{() => selectSmiley(1)}">
             <Fa
               size="1.5x"
-              class="cursor-pointer {smileyRating === 1 ? 'text-violet-400' : 'text-gray-900'}"
+              class="cursor-pointer {smileyRating === 1
+                ? 'text-[var(--pd-invert-content-button-active)]'
```
deboer-tim: I think using some of the --pd-button colors would be better here, but since this is better and wasn't the focus I'm ok with it.

deboer-tim (approved):

packages/renderer/src/lib/feedback/SendFeedback.svelte
SoniaSandler: @deboer-tim I updated the colors to use `--pd-button-primary-bg` and `--pd-button-disabled-text`

PR #7906, opened 2024-07-03
fix: restart informers for not current contexts after changes

### LINKED ISSUE #7707: Changing details of a non-current Kube context does not restart informers with new details
### Bug description

The issues #6934 and #7688 cover the case when the current context is modified (cluster, user, namespace). This should also be supported for non-current contexts.
 

### Operating system

mac

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

- Have two Kubernetes clusters running
- Start Podman Desktop
- select cluster 1 as current
- in cluster 2, create 2 namespaces ns1 and ns2, create 1 deployment in ns1 and 2 deployments in ns2, and set ns1 as namespace of context of cluster 2
- edit ~/.kube/config file with an editor, to change the namespace of the context of cluster 2 to ns2
- the number of deployments should be changed to 2

### Relevant log output

_No response_

### Additional context

_No response_

dgolovin:
### What does this PR do?

Iterates over contexts left after getting rid of deleted and added contexts to find changed ones and restart informers.

### Screenshot / video of UI

https://github.com/containers/podman-desktop/assets/620330/0af4d00e-14f4-49a2-bf0e-83f23086e3f3

Fix #7707

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes-context-state.ts
+++ b/packages/main/src/plugin/kubernetes-context-state.ts
@@ -19,12 +19,14 @@
 import type { IncomingMessage } from 'node:http';
 
 import type {
+  Cluster,
   Context,
   Informer,
   KubernetesListObject,
   KubernetesObject,
   ListPromise,
   ObjectCache,
+  User,
   V1Deployment,
   V1DeploymentList,
   V1Ingress,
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/kubernetes-context-state.ts
```diff
@@ -435,9 +435,32 @@ export class ContextsManager {
     );
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  compareObjects(newu: any, oldu: any): boolean {
+    if (!newu || !oldu) return false;
```
feloy: If could return true if both are 'zero' value

feloy (approved):
Works great, thanks, just a little comment

jeffmaury (approved):
LGTM

packages/main/src/plugin/kubernetes-context-state.ts
dgolovin: updated the compareObject mehtod and added tests for specific update use-cases

PR #7912, opened 2024-07-03
chore(StatusIcon): adding accepting values for status

### LINKED ISSUE #7909: StatusIcon should expose the status it supports

@deboer-tim suggested in https://github.com/containers/podman-desktop/issues/7899#issuecomment-2203427898 that StatusIcon component which has been migrated to the UI package in https://github.com/containers/podman-desktop/pull/7900 should not use `''` as type

https://github.com/containers/podman-desktop/blob/ae7e9c93e868a07adabe7e659e9503e3be159a47/packages/ui/src/lib/statusIcon/StatusIcon.svelte#L7

Use something like

export let status: 'VALUE1' | 'VALUE2' | undefined

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Add the accepted value, **and `string`** because most of our interface are using string for status.

https://github.com/containers/podman-desktop/blob/bf1acf37c99661ea8b7b8563894e37078fbe21df/packages/renderer/src/lib/compose/ComposeInfoUI.ts#L25

https://github.com/containers/podman-desktop/blob/ea3267976bfe065c155d876832b43b8371df1ea1/packages/renderer/src/lib/ingresses-routes/IngressUI.ts#L24

https://github.com/containers/podman-desktop/blob/a0ca7600815ce7b57dd1d3a4f8a5881d31d14e28/packages/renderer/src/lib/pvc/PVCUI.ts#L22

etc.

Fixes https://github.com/containers/podman-desktop/issues/7909

### How to test this PR?

- Pipeline should be ✅

### DIFF
--- a/packages/ui/src/lib/statusIcon/StatusIcon.svelte
+++ b/packages/ui/src/lib/statusIcon/StatusIcon.svelte
@@ -4,7 +4,7 @@ import Spinner from '../progress/Spinner.svelte';
 
 // status: one of RUNNING, STARTING, USED, CREATED, DELETING, or DEGRADED
 // any other status will result in a standard outlined box
-export let status = '';
+export let status: 'RUNNING' | 'STARTING' | 'USED' | 'DEGRADED' | 'DELETING' | 'CREATED' | string = 'UNKNOWN';
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export let icon: any = undefined;
 export let size = 20;


deboer-tim (approved):
That is an odd enum, but thank you, I think this helps an adopter to understand what status values are going to get special treatment.

feloy (approved):

benoitf (approved):

PR #7927, opened 2024-07-03
fix(navigation): remove svelte store for container export logic

### LINKED ISSUE #7550: Export container form embedded within details page
### Bug description

The export container form works fine when you launch it from the Containers page, but if you go to a container's details and then click on it, the form appears on the bottom of the page instead of full size.

Screenshot 2024-06-10 at 10 19 14 AM

### Operating system

macOS 14.5

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

### LINKED ISSUE #7925: Export container should not use dedicated store

Currently we have an router path `/containers/export` which requires a store 

https://github.com/containers/podman-desktop/blob/8419cb69b7608045f2f4a96808a436335054a4e1/packages/renderer/src/stores/export-container-store.ts#L27

When the ContainerAction wants to export a container it fill the store with the ContainerInfo, then redirect to the page.

https://github.com/containers/podman-desktop/blob/423d44cc7857dc3c1b225c4a17ce37623dda9923/packages/renderer/src/lib/container/ContainerActions.svelte#L118-L119

This seems to be a very weird and heavy pattern for nothing.

Proposal: change `/containers/export` to `/container/<container-id>/export` so the component can fetch the container id, and the corresponding `ContainerInfoUI` it requires.

We then can delete this store and avoid having all managed in the export component with the router parameter

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

The current export page was using a dedicated svelte store to store the current container we want to export, this is not necessary as we can simply provide the container as a url parameter, and the component can get it from existing store.

### Screenshot / video of UI

image

Fixes https://github.com/containers/podman-desktop/issues/7925
Fixes https://github.com/containers/podman-desktop/issues/7550

### How to test this PR?

Go to containers page, select one and try to export


deboer-tim (changes requested):
+1 on changes overall, code LGTM.

First the good: you can add 'Fixes #7550' since the route changes avoid the overlapping urls and I can throw out my fix for that. ;-)

The bad: this appears to break breadcrumb navigation. I haven't looked at why yet, but if I go to the Containers list and click on a container the breadcrumb is 'Container Details > Container Details' and I can't go back.

axel7083:
> First the good: you can add 'Fixes #7550' since the route changes avoid the overlapping urls and I can throw out my fix for that. ;-)

Nice, love fixing issue without knowing they exists x)

> The bad: this appears to break breadcrumb navigation. I haven't looked at why yet, but if I go to the Containers list and click on a container the breadcrumb is 'Container Details > Container Details' and I can't go back.

Okey I fixed by moving down the breadcrumb, I tried and it is working as I would expect. It is just different depending if you open from containers details, you will get a breadcrumb containers details, if you open from the containers list, you will get containers breadcrumb.

deboer-tim (approved):
Thanks, confirmed the fix (as well as intended behaviour) works well. 👍🏼

PR #7930, opened 2024-07-03
Search images in registry (backend)

feloy:

### What does this PR do?

Add function to search images in registries

TODO
- [x] add unit tests

### Screenshot / video of UI

(backend only)

Part of #971 

### How to test this PR?

### DIFF
--- a/packages/api/src/image-registry.ts
+++ b/packages/api/src/image-registry.ts
@@ -1,0 +1,24 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface ImageSearchResult {
+  name: string;
+  description: string;
+  star_count: number;
+  is_official: boolean;
+}
--- a/packages/api/src/image-registry.ts
+++ b/packages/api/src/image-registry.ts
@@ -15,6 +15,11 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
+export interface ImageSearchOptions {
+  registry?: string;
+  query: string;
+  limit?: number;
+}
 
 export interface ImageSearchResult {
   name: string;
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/image-registry.spec.ts
```diff
@@ -826,3 +826,48 @@ test('getToken without registry auth', async () => {
...
+    .reply(200, {
+      results: [
+        {
+          name: 'image1',
+          description: 'desc',
+        },
+      ],
+    });
+  const result = await imageRegistry.searchImages('https://quay.io', 'http', 10);
+  console.log('result', result);
```
benoitf:
```suggestion

```

packages/main/src/plugin/image-registry.ts
```diff
@@ -872,6 +874,17 @@ export class ImageRegistry {
       throw Error('Unable to validate provided credentials.');
     }
   }
+
+  async searchImages(registryName: string, term: string, limit: number): Promise<ImageSearchResult[]> {
```
benoitf: can we omit the registryName and assume default value is dockerhub ?

Also maybe we should have a single variable as parameter like searchOptions

benoitf: searchOptions?: {
registry?: string
query (rather than term)
limit (also should be optional and we default to like 20 or 30)

benoitf (approved):

deboer-tim (approved):

packages/main/src/plugin/index.ts
```diff
@@ -1495,6 +1496,13 @@ export class PluginSystem {
       },
     );
 
+    this.ipcHandle(
+      'image-registry:searchImages',
+      async (_listener, registryName: string, term: string, limit: number): Promise<ImageSearchResult[]> => {
+        return imageRegistry.searchImages(registryName, term, limit);
```
benoitf: I think this file is requiring a change (probably the file has not been committed with the change ? to use the options parameter rather than several parameters)

feloy: Yes right, I just pushed a new commit for this. Can you please validate the change?

benoitf (approved):

jeffmaury (approved):
LGTM

Only question is using got for http request respect proxy settings ?

feloy:
> LGTM
> 
> Only question is using got for http request respect proxy settings ?

good catch, it didn't. This should be fixed now with latest commit. 

In parallel, I'm fixing https://github.com/containers/podman-desktop/issues/7940, I'll add unit tests when this is fixed.

packages/api/src/image-registry.ts
```diff
@@ -0,0 +1,29 @@
...
+  registry?: string;
+  query: string;
+  limit?: number;
+}
+
+export interface ImageSearchResult {
+  name: string;
+  description: string;
+  star_count: number;
+  is_official: boolean;
```
axel7083, line 28: Why are you using underscore separation ? would `isOfficial` and `starCount` be consistent with _most_ of the code base ?

feloy: To match with what is returned by the API call, so `return JSON.parse(resultJSON.body).results` works without any mapping

axel7083: okey 👍

feloy:
I got some headache to find a solution to test the proxy usage using nock, that should be ok now :)

PR #7933, opened 2024-07-03
chore(navigation): adding type safety handleNavigation

### LINKED ISSUE #7922: Navigation API should be type safe

Currently the entry 

https://github.com/containers/podman-desktop/blob/4fc2eaca1bc4ea0a919e49c39c1ac3d25bb5ef9f/packages/renderer/src/navigation.ts#L32

is not really type safe, as we do not check for parameters provided.

using typescript feature we could enhance by defining the parameters definition

```
// Define the type mapping for parameters
type NavigationParameters = {
  [NavigationPage.CONTAINERS]?: {};
  [NavigationPage.CONTAINER]?: { id: string };
  [NavigationPage.IMAGE]?: { id: string, engineId: string, tag: string };
  ...
};

// Update the handleNavigation function
export const handleNavigation = <T extends NavigationPage>(
  page: T,
  parameters?: NavigationParameters[T]
) => {
...
}
```

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

The `handleNavigation` was having a problem, the arguments are way too permissive, and nothing enforce the right parameters to be given. An error could be easily made, were we would be sending the wrong arguments, and therefore breaking the navigation.

Thankfully typescript is the right language to solve that. A few solution were possible,

- overloading the function by adding a new line, therefore any caller would be type-safe, but not the function body
- adding a dedicated method for each navigation, making a lot of code redundant

I decided to choose another one, which is creating a custom interface holding the parameters type, and using typescript Distributive Conditional Types[^1] to infer the argument based on the switch in the `handleNavigation` while having the safety on the caller.

[^1]: https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types

Fixes https://github.com/containers/podman-desktop/issues/7922

### How to test this PR?

### DIFF
--- a/packages/api/src/navigation-request.ts
+++ b/packages/api/src/navigation-request.ts
@@ -18,7 +18,37 @@
 
 import type { NavigationPage } from './navigation-page.js';
 
-export interface NavigationRequest {
-  page: NavigationPage;
-  parameters?: { [key: string]: string };
+// Define the type mapping for parameters
+export interface NavigationParameters {
+  [NavigationPage.CONTAINERS]: never;
+  [NavigationPage.CONTAINER]: { id: string };
+  [NavigationPage.CONTAINERS_EXPORT]: { id: string };
+  [NavigationPage.CONTAINER_LOGS]: { id: string };
+  [NavigationPage.CONTAINER_INSPECT]: { id: string };
+  [NavigationPage.CONTAINER_TERMINAL]: { id: string };
+  [NavigationPage.CONTAINER_KUBE]: { id: string };
+  [NavigationPage.DEPLOY_TO_KUBE]: { id: string; engineId: string };
+  [NavigationPage.IMAGES]: never;
+  [NavigationPage.IMAGE]: { id: string; engineId: string; tag: string };
+  [NavigationPage.PODS]: never;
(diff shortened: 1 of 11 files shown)


benoitf (commented):
I'm wondering why it has been changed to a single object vs first the name of the page and then the parameters for the page ?

axel7083:
> I'm wondering why it has been changed to a single object vs first the name of the page and then the parameters for the page ?

Because typescript does not allow inferencing arguments based on others :( it has to be a single object otherwise we cannot do something simple, and need to overload the function for each argument...

```
// navigation.d.ts
handleNavigation(page: NavigationPage.CONTAINERS): void;
handleNavigation(page: NavigationPage.CONTAINER, parameters: {id: string});
etc..
```

But with this approach, you do not have the right inference type inside the function, as you need to define a generic function

```
handleNavigation(page: NavigationPage, parameters?: Record<string, string>)
```

and do the checking/casting manually.

With a single object, we can infer properly as we have a differentiator (here page).


Moreover, it unify with the way we were doing in the main package.

jeffmaury (approved):
LGTM

Issue #7938, opened 2024-07-03
Use FileInput consistently


deboer-tim:
### Is your enhancement related to a problem? Please describe

We have a FileInput component, but haven't used it in many places yet.

### Describe the solution you'd like

Replace existing Input + Browse button instances with the FileInput:
- Build Image
- Play Kube
- Run Image

PR #7939, opened 2024-07-03
chore: use fileinput in build image form

deboer-tim:
### What does this PR do?

Just some tech debt, use the FileInput component in the build image form.

### Screenshot / video of UI

Before:

Screenshot 2024-07-03 at 4 22 27 PM

After:

Screenshot 2024-07-03 at 4 22 40 PM

Part of #7938.

### How to test this PR?

Build an image.

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.spec.ts
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2024 Red Hat, Inc.
  *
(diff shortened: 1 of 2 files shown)


benoitf:
It looks like that the e2e tests need to be updated

feloy:
On Mac, I cannot select files named Containerfile or Dockerfile:

build-image-openfile

benoitf (approved):

deboer-tim:
> It looks like that the e2e tests need to be updated

I had unintentionally renamed the directory placeholder,  the change was unnecessary so I removed it.

> On Mac, I cannot select files named Containerfile or Dockerfile:

Unfortunately the Electron file filter doesn't allow different capitalization or regex, so I've removed the filter as well.

feloy (approved):

PR #7942, opened 2024-07-04
fix(compose-extension): pre-existing system-wide installation 

### LINKED ISSUE #5250: Prior compose install not showing correctly in CLI Tools
### Bug description

The same problem as #5216 appears to exist for compose as well.

I have docker-compose installed at `/usr/local/bin/docker-compose` and it works with `podman compose`, but we are unable to detect it/the version in CLI Tools:

Screenshot 2023-12-13 at 11 55 59 AM

### Operating system

macOS 14.1

### Installation Method

None

### Version

1.6.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Improve the `docker-compose` detection, by searching existing system wide install first.

### Screenshot / video of UI

**Trying to update a version in a path unknown to podman-desktop**

> Expected error raised 

https://github.com/containers/podman-desktop/assets/42176370/b8f55352-22bf-4c6e-a97d-1eb0790d82d8

**Trying to update a version in a path known by podman-desktop**

> Expected everything go smoothly 

https://github.com/containers/podman-desktop/assets/42176370/b69f22c2-47e4-490a-9543-637d88a47c80

Fixes https://github.com/containers/podman-desktop/issues/5250

### How to test this PR?

- [x] Added `extension.spec.ts` for extension activation / post activation

### DIFF
--- a/extensions/compose/src/detect.spec.ts
+++ b/extensions/compose/src/detect.spec.ts
@@ -151,7 +151,7 @@ describe('Check getDockerComposePath uses proper tooling by platform', () => {
     );
   });
 
-  test('linux should uses which', async () => {
+  test('linux should use which', async () => {
     (extensionApi.env.isLinux as boolean) = true;
     (extensionApi.env.isWindows as boolean) = false;
     (extensionApi.env.isMac as boolean) = false;
@@ -160,7 +160,7 @@ describe('Check getDockerComposePath uses proper tooling by platform', () => {
     expect(extensionApi.process.exec).toHaveBeenCalledWith('which', ['docker-compose']);
   });
 
-  test('mac should uses which', async () => {
+  test('mac should use which', async () => {
     (extensionApi.env.isMac as boolean) = true;
     (extensionApi.env.isWindows as boolean) = false;
     (extensionApi.env.isLinux as boolean) = false;
@@ -169,7 +169,7 @@ describe('Check getDockerComposePath uses proper tooling by platform', () => {
     expect(extensionApi.process.exec).toHaveBeenCalledWith('which', ['docker-compose']);
   });
 
-  test('windows should uses which', async () => {
+  test('windows should use which', async () => {
     (extensionApi.env.isWindows as boolean) = true;
     (extensionApi.env.isMac as boolean) = false;
     (extensionApi.env.isLinux as boolean) = false;
--- a/extensions/compose/src/detect.spec.ts
+++ b/extensions/compose/src/detect.spec.ts
@@ -50,6 +50,11 @@ vi.mock('@podman-desktop/api', async () => {
     process: {
       exec: vi.fn(),
     },
+    env: {
+      isLinux: false,
+      isWindows: false,
+      isMac: false,
+    },
   };
 });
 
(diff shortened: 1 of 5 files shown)


extensions/compose/src/extension.ts
```diff
@@ -216,56 +214,68 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
+    // if local version exists
+    if (extensionExecutable.length !== 0) {
+      binaryInfo = await detect.getDockerComposeBinaryInfo(executable, detect.getExtensionStorageBin());
+    }
   }
 
-  // Register the CLI tool so it appears in the preferences page. We will detect which version is being ran by
-  // checking the local storage folder for the binary. If it exists, we will run `--version` and parse the information.
+  // if no binary detected let's just stop here
+  if (!binaryInfo) return;
```
deboer-tim, line 252: If compose is not installed, this removes it from the Settings entirely, i.e. no trace that Podman Desktop even knows what compose is. If it is not installed it should still be in the Settings page, show that it's not installed, and ideally install it.

extensions/compose/src/detect.spec.ts
```diff
@@ -136,6 +141,72 @@ describe('Check storage path', async () => {
...
+  beforeEach(() => {
+    vi.mocked(extensionApi.process.exec).mockImplementation(
+      () =>
+        new Promise<extensionApi.RunResult>(resolve => {
+          resolve({ exitCode: 0, stdout: 'hello-world' } as extensionApi.RunError);
+        }),
+    );
+  });
+
+  test('linux should uses which', async () => {
```
deboer-tim: nit: should be 'should _use_ which' here and the other two cases.

deboer-tim (changes requested):

extensions/compose/src/extension.ts
```diff
@@ -216,56 +214,68 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
-    console.error(`Error getting compose version: ${e}`);
+async function postActivate(composeDownload: ComposeDownload, detect: Detect): Promise<void> {
+  // build executable name for current platform
+  const executable = os.isWindows() ? composeCliName + '.exe' : composeCliName;
+
+  // binary info
+  let binaryInfo: { version: string; path: string; updatable?: boolean } | undefined = undefined;
+
+  // let's check for system-wide
+  const installedSystemWide = await detect.checkSystemWideDockerCompose();
```
dgolovin, line 239: nit: `Detect.checkSystemWideDockerCompose()` and `detect.getDockerComposeBinaryInfo()` run exactly the same command. There should be just one call of `detect.getDockerComposeBinaryInfo()` used in try/catch block to have the same information about compose system wide availability, but one compose execution less in case it is installed.

extensions/compose/src/extension.ts
axel7083: https://github.com/containers/podman-desktop/assets/42176370/6e0df8b7-2f8d-4f73-bfdd-85005f30648f

We do not have a status for not-installed in the CLI tool, we need to provide some required arguments, 

https://github.com/containers/podman-desktop/blob/0fb809c40babc26327022fbfe3ce2bb5b9d7d6d2/packages/extension-api/src/extension-api.d.ts#L4269-L4284

You need to use the setup workflow to have it install

extensions/compose/src/detect.spec.ts
axel7083:
```suggestion
  test('linux should use which', async () => {
```

extensions/compose/src/detect.spec.ts
```diff
@@ -136,6 +141,72 @@ describe('Check storage path', async () => {
...
+    (extensionApi.env.isLinux as boolean) = true;
+    (extensionApi.env.isWindows as boolean) = false;
+    (extensionApi.env.isMac as boolean) = false;
+
+    await detect.getDockerComposePath('docker-compose');
+    expect(extensionApi.process.exec).toHaveBeenCalledWith('which', ['docker-compose']);
+  });
+
+  test('mac should uses which', async () => {
```
axel7083:
```suggestion
  test('mac should use which', async () => {
```

extensions/compose/src/detect.spec.ts
```diff
@@ -136,6 +141,72 @@ describe('Check storage path', async () => {
...
+    (extensionApi.env.isMac as boolean) = true;
+    (extensionApi.env.isWindows as boolean) = false;
+    (extensionApi.env.isLinux as boolean) = false;
+
+    await detect.getDockerComposePath('docker-compose');
+    expect(extensionApi.process.exec).toHaveBeenCalledWith('which', ['docker-compose']);
+  });
+
+  test('windows should uses which', async () => {
```
axel7083:
```suggestion
  test('windows should use which', async () => {
```

extensions/compose/src/extension.ts
deboer-tim: Personally, I don't think CLI tools should be under both Resources and CLI Tools, and that it's poor that you have to go Resources to install something you can only see the details of in CLI Tools... but that's an issue for another day.

My concern here was that before this change I can see compose listed under CLI Tools even if I don't have it installed yet, but after this PR it's not listed at all. We might need a separate issue to show 'not installed' state better, but IMHO it should remain visible. Otherwise, I would get the impression that it is only visible in Resources.

axel7083: > I don't think CLI tools should be under both Resources and CLI Tools

This is not a problem related to the issue here. Basically the way the extension api is made, the extension should register _an existing and valid_ version, and it accepts an update callback.

The current design does not support registering an installation, and here it is in the Resources because it is part of the onBoarding workflow of the Compose extension.

> My concern here was that before this change I can see compose listed under CLI Tools even if I don't have it installed yet, but after this PR it's not listed at all.

Even if it was listed, you could not install it, because it was not suppose to be here, leading to user being perplex has to understand if it was or not installed. 

> We might need a separate issue to show 'not installed' state better

I agree, that we should have a setup/install option here, it would simplify a lot.

> but IMHO it should remain visible

Not in this PR.

deboer-tim: As discussed on the standup call I don't like this 'regression' of compose not appearing in CLI Tools anymore if you don't have it installed, but I'm ok with this as a temporary state with #8003 open to track longer term behaviour. Thanks. 👍🏼

deboer-tim (approved):

PR #7951, opened 2024-07-04
Use the real path when watching files

### LINKED ISSUE #7512: watching a file fails if its path is not the real one
### Bug description

When using `FileSystemWatcher` from the API, if the call watches a file, using a non-real path (using a symbolic link of the file or one of its parent dir), the watch fails.

### Operating system

mac

### Installation Method

### Version

1.10.2

### Steps to reproduce

On Mac:

```
cd $HOME
mv .kube .kube-save
mkdir /tmp/.kube
ln -s /tmp/.kube .
```

- Start Podman Desktop
- Create a Kind cluster (or anything else to create a ~/.kube/config file)
- Podman Desktop should detect the kubeconfig (by displaying Kubernetes icons at least), but it doesn't 

### Relevant log output

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

Use the real path when watching files

Fixes #7512 

### How to test this PR?

See Steps to reproduce in #7512 

### DIFF
--- a/packages/main/src/plugin/filesystem-monitoring.spec.ts
+++ b/packages/main/src/plugin/filesystem-monitoring.spec.ts
@@ -18,7 +18,7 @@
 
 import { promises } from 'node:fs';
 import os from 'node:os';
-import path from 'node:path';
+import path, { join } from 'node:path';
 
 import { afterEach, beforeEach, expect, test, vi } from 'vitest';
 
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM

Just wondering if it would be more readable to see f.realpathSync(path) but maybe it's not allowed on file pathes

benoitf (approved):
looks like we're fixing symlink issues

I'm seeing @jeffmaury issue there 
https://github.com/containers/podman-desktop/pull/7926

I don't know if we could use/switch to an asynchronous mode (with await) without doing it in the constructor

Issue #7465, opened 2024-06-04
Different colour header on top of modal for push images


cdrage:
### Bug description

![Screenshot 2024-06-04 at 2 01 46 PM](https://github.com/containers/podman-desktop/assets/6422176/57f0f6d6-35c8-493e-b10f-2cacc031665c)

If you go to push an image, there is some odd colouring happening at the top?

### Operating system

mac m1

### Installation Method

None

### Version

1.10.2

### Steps to reproduce

Push any image

axel7083:
All modals bellow are using the `--pd-modal-header-bg` variable which is black or gray depending on the theme

https://github.com/containers/podman-desktop/blob/88907772c08058ac7bb91a79e9afec367f43ca75/packages/main/src/plugin/color-registry.ts#L808

| Modal | Screenshot |
| --- | --- |
| `Install Extension from OCI image` | ![image](https://github.com/containers/podman-desktop/assets/42176370/df2f83b1-a995-49d9-a4ff-d36a4dc23d6a) |
| `Create a new container` | ![image](https://github.com/containers/podman-desktop/assets/42176370/bea19f41-a6dd-4a8e-912c-99a664c1c4d2) |
| `Share your feedback` | ![image](https://github.com/containers/podman-desktop/assets/42176370/63eddd86-b8be-4141-af08-7ff8aeb5dc8b) |
| `Push image` | ![image](https://github.com/containers/podman-desktop/assets/42176370/6354fa80-e050-4902-9e93-7c3335b747b9) |
| `Edit Image` | ![image](https://github.com/containers/podman-desktop/assets/42176370/0aa7466a-fb22-48cc-bcd6-baba94fdccad) |
| `PreferencesProviderInstallationModal.svelte` | N/A |
| `TroubleshootingPageStoreDetails.svelte` | ![image](https://github.com/containers/podman-desktop/assets/42176370/488068fc-9366-45cc-9ffd-48272030a18c) |

⚠️ The modal in TroubleshootingPageStoreDetails.svelte is very buggy

Some of the modal have a purple line between the header and the body. We probably need to define some guideline to choose a consistent style for the modals, as for the majority we are not using some specific color for the header.

ekidneyrh:
Is there a way we could get rid of the black header and the purple line? It's an older iteration of the modal dialog. I dont think there's a need to have them there anymore. The spacing between the header and the body text would have to be adjusted too probably. 

They should all match this one: 

[...]

PR #7967, opened 2024-07-05
feat: dialog component

deboer-tim:
### What does this PR do?

We have a lot of inconsistencies in our dialogs - they all use Modal now, but as shown in #7465 the header, content, and button bar are often styled differently. As proposed this creates a Dialog component (based on what was in MessageBox, which is the approved design) that has a fixed header (title bar and close button), a padded content area, and a right-aligned bottom button bar with a space on the left for validation.

I've used the new component for MessageBox, SendFeedback, and RenameImageModal (Edit Image) so that there is something to review; if this goes in we can then reuse it for other dialogs.

I did _not_ attempt to change the styling within the content area or buttons of these dialogs except where it was required to adopt the Dialog.

### Screenshot / video of UI

Before:

Screenshot 2024-07-04 at 11 41 25 AM

Screenshot 2024-07-04 at 11 41 57 AM

Screenshot 2024-07-04 at 11 42 30 AM

After:

Screenshot 2024-07-05 at 10 56 44 AM

Screenshot 2024-07-05 at 10 56 01 AM

Screenshot 2024-07-05 at 10 56 29 AM

First part of #7465.

### How to test this PR?

Open these three dialogs and check dialog styling.

### DIFF
--- a/packages/renderer/src/lib/dialogs/Dialog.svelte
+++ b/packages/renderer/src/lib/dialogs/Dialog.svelte
@@ -19,12 +19,14 @@ export let onclose: () => void = () => {
     <CloseButton on:click="{() => onclose()}" />
   </div>
 
-  <div class="max-h-96 overflow-auto text-[var(--pd-modal-text)] px-10 py-4">
+  <div class="max-h-80 overflow-auto text-[var(--pd-modal-text)] px-10 py-4">
     <slot name="content" />
   </div>
 
-  <div class="px-5 py-5 mt-2 flex flex-row w-full justify-end space-x-5">
-    <slot name="validation" />
+  <div class="px-5 py-5 mt-2 flex flex-row w-full space-x-5">
+    <div class="grow">
+      <slot name="validation" />
+    </div>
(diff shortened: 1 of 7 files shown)


packages/renderer/src/lib/dialogs/Dialog.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
```
axel7083, line 1: Do we want to expose it in the ui library ?

axel7083 (approved):

packages/renderer/src/lib/dialogs/Dialog.svelte
deboer-tim: I think we should, I just didn't want to start there since I was aware of any imminent extension need. i.e. get it in, use in a few more places and make sure it's ok, then export.

lstocchi (approved):
LGTM

ekidneyrh (approved):
Looks great! thanks Tim

PR #7975, opened 2024-07-08
fix: podman machine switch may lose API connection

### LINKED ISSUE #6455: Kind cluster not visible with Podman 5 rc7
### Bug description

I do not see kind cluster in podman destkop. No error was shown in the console log regarding kind cluster creation. I am using podman 5 (rc7) with hyperv provider. The Machine is rootfull with user mode networking enabled.
podman5-usermode-kind-cluster-not-visible

### Operating system

Windows 10

### Installation Method

### Version

1.8.0

### Steps to reproduce

Install podman desktop and podman 5 rc 7 from installer.
1. Create new podman machine - rootful and with usermode networking enabled
2. Create a kind cluster
3. Kind cluster created
4. assert: Podman Desktop sees the kind related resources (images, containers, etc.)
actual result: Podman desktop does not show the cluster resources nor container objects.

### Relevant log output

```shell
VM37:2 main ↪️ [podman]  Can't get machine usermode resource usage error Error: no running provider for the matching container
console.<computed>	@	VM37:2
(anonymous)	@	VM37:2
emit	@	VM36 sandbox_bundle:2
onMessage	@	VM36 sandbox_bundle:2
```

### Additional context

I am observing that podman cli works nice, If I output `podman images` I can see `kindest` image present.

### LINKED ISSUE #6982: Podman Machine sometime is not able to pull image via Podman Desktop on Windows with WSL with Error: no running provider for the matching container
### Bug description

I have a rootful machine or rootless one, individually both works fine. When I got into scenario, where I have one machine stopped with connection being used by this machine and creates new machine and starts it, I am asked by podman desktop if I want to switch the defautl connection, which I accepts and all seems fine. Although when I try to pull an image I got: `Error: no running provider for the matching container` in the UI.

Although If I try to pull an image from cli, everything works fine.
It is same issue as #6455, but this time the reproducer differs.

### Operating system

Windows 10

### Installation Method

### Version

1.10.0

### Steps to reproduce

1. Create and start a machine using podman 5.0.2.
2. Stop the machine
3. Create new machine different than the other
4. Start new machine
5. Accept switching of the default connection
6. Assert: Machine started, `podman system connection ls` show right connection being set to default
7. Pull an image using podman desktop
8. Assert: Can pull image
Actual result: image is not pulled, Error appers.

`podman pull alpine` works fine.

### Relevant log output

_No response_

### Additional context

Referenced issue #6455. Very likely the problem is the same, although symptoms differ.

### Workaround

Restarting the machine sometimes helps

axel7083:
### What does this PR do?

The problem was coming from a racing issue (most probable). 

Let's break it down. We have a `setupConnectionAPI` which is responsible for updating the `api` and `libpodApi` of the `InternalContainerProvider`.

https://github.com/containers/podman-desktop/blob/6b600670b6342613c47c9d5cda27fc9ce486aebd/packages/main/src/plugin/container-registry.ts#L258

However, this method has a special check:

https://github.com/containers/podman-desktop/blob/6b600670b6342613c47c9d5cda27fc9ce486aebd/packages/main/src/plugin/container-registry.ts#L263-L266

If the `ContainerProviderConnection` has a `stopped` status it will not create the `Dockerode` object.

This method is called from several places, the most important is the one in the `setInterval`

https://github.com/containers/podman-desktop/blob/6b600670b6342613c47c9d5cda27fc9ce486aebd/packages/main/src/plugin/container-registry.ts#L359-L363

This code is executed every 2 seconds, the problem: We might have the `containerProviderConnection` still with a `stopped` status when `setupConnectionAPI` is called. And after it is called, the `previousStatus = newStatus` and will never be executed again!

This is tricky, as we could fix this in different way. The best I found was to have some interest on the update provided by the `providerRegistry`.

We are adding a listener on their `onBeforeDidUpdateContainerConnection` event. However, on this event, since it BEFORE, the status _may be_ `stopped`, therefore the call to setupConnectionAPI is also useless.

The fix was to add a specific listener to the `onAfterDidUpdateContainerConnection` when the status would be `started` to call the `setupConnectionAPI`.

### Screenshot / video of UI

N/A

Fixes https://github.com/containers/podman-desktop/issues/6982 and https://github.com/containers/podman-desktop/issues/6455

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -336,7 +336,7 @@ export class ContainerProviderRegistry {
       if (
         event.providerId === provider.id && // ensure provider id is matching
         event.connection.name === containerProviderConnection.name && // ensure connection is matching
-        event.status === 'started' && // when status is tarted
+        event.status === 'started' && // when status is started
         !internalProvider.api &&
         !internalProvider.libpodApi // api & libpodApi are undefined we need to setup
       ) {
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -318,18 +318,30 @@ export class ContainerProviderRegistry {
     let previousStatus = containerProviderConnection.status();
 
     providerRegistry.onBeforeDidUpdateContainerConnection(event => {
-      if (event.providerId === provider.id && event.connection.name === containerProviderConnection.name) {
-        const newStatus = event.status;
-        if (newStatus === 'stopped') {
-          internalProvider.api = undefined;
-          internalProvider.libpodApi = undefined;
-          this.apiSender.send('provider-change', {});
-        }
-        if (newStatus === 'started') {
-          this.setupConnectionAPI(internalProvider, containerProviderConnection);
-        }
+      if (
+        event.providerId === provider.id &&
+        event.connection.name === containerProviderConnection.name &&
+        event.status === 'stopped'
+      ) {
+        internalProvider.api = undefined;
+        internalProvider.libpodApi = undefined;
+        this.apiSender.send('provider-change', {});
       }
       previousStatus = event.status;
+    });
+
+    // we need to call setupConnectionAPI after the update, as it is checking that the connection
+    // is started.
+    providerRegistry.onAfterDidUpdateContainerConnection(event => {
+      if (
+        event.providerId === provider.id && // ensure provider id is matching
+        event.connection.name === containerProviderConnection.name && // ensure connection is matching
+        event.status === 'started' && // when status is tarted
+        !internalProvider.api &&
+        !internalProvider.libpodApi // api & libpodApi are undefined we need to setup
+      ) {
+        this.setupConnectionAPI(internalProvider, containerProviderConnection);
+      }
     });
 
     if (containerProviderConnection.status() === 'started') {
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -330,20 +330,6 @@ export class ContainerProviderRegistry {
       previousStatus = event.status;
     });
 
-    // we need to call setupConnectionAPI after the update, as it is checking that the connection
-    // is started.
-    providerRegistry.onAfterDidUpdateContainerConnection(event => {
-      if (
-        event.providerId === provider.id && // ensure provider id is matching
-        event.connection.name === containerProviderConnection.name && // ensure connection is matching
-        event.status === 'started' && // when status is started
-        !internalProvider.api &&
-        !internalProvider.libpodApi // api & libpodApi are undefined we need to setup
-      ) {
-        this.setupConnectionAPI(internalProvider, containerProviderConnection);
-      }
-    });
-
     if (containerProviderConnection.status() === 'started') {
       this.setupConnectionAPI(internalProvider, containerProviderConnection);
     }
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/container-registry.ts
```diff
@@ -317,20 +317,32 @@ export class ContainerProviderRegistry {
...
       previousStatus = event.status;
     });
 
+    // we need to call setupConnectionAPI after the update, as it is checking that the connection
+    // is started.
+    providerRegistry.onAfterDidUpdateContainerConnection(event => {
+      if (
+        event.providerId === provider.id && // ensure provider id is matching
+        event.connection.name === containerProviderConnection.name && // ensure connection is matching
+        event.status === 'started' && // when status is tarted
```
dgolovin:
```suggestion
        event.status === 'started' && // when status is started
```

dgolovin (approved):
LGTM with a nit

lstocchi:
😱 cool, this should also solve https://github.com/containers/podman-desktop/issues/6455 . @odockal could you try to replicate the issue you faced using this PR, whenever you have time?

odockal:
@lstocchi @axel7083 I managed to get into problem outlined in #6982 once. 10 attempts to switch between machines (I got two, rootless and rootful) to reproduce the problem. Then another 20 attempts and I cannot get into it. From my perspective there is an improvement. Which is awesome. And 21st attempt (roughly) ended up with a problem again.

Anyway, only thing I can debug from PD side, is the dev console messages mapped to the actions.
Successful switch, no matter if from rootful to rootless or vice versa looks like this:
stop-start-successful
Erroneous switch messages look like this:
stop-start-error

So there is one more warning message at the end that could help to identify the code, that was not finished/triggered again.
```
main ↪️ [podman]  Can't get machine rootless resource usage error Error: no running provider for the matching container
console.<computed> @ VM5:2
(anonymous) @ VM5:2
emit @ VM4 sandbox_bundle:2
onMessage @ VM4 sandbox_bundle:2
```

axel7083:
>  And 21st attempt (roughly) ended up with a problem again.

I think there is a clear race condition, and I am really not sure of the best choice to do here. If this change is able to reduce enough the risk of it happening, it is still better than nothing :/ 

No sure what to do here @odockal

axel7083:
> And 21st attempt (roughly) ended up with a problem again.

@odockal okey, so what I changed now, is instead of trying to be event reactive, I am uses the interval update mechanism we are already using, to setup the api if it is not defined.

Could you retry ?

axel7083:
cc @odockal

odockal:
@axel7083 On it.

odockal:
@axel7083 I have not see the problem again at all.

Side issue maybe: Once, I clicked on Ignore (when switching the connection), and since I cannot get the dialog back. Is that ok?

axel7083:
> Side issue maybe: Once, I clicked on Ignore (when switching the connection), and since I cannot get the dialog back. Is that ok?

I think it should yes

lstocchi:
Yes, the ignore should prevent the dialog from reappearing.
Kudos @axel7083 for fixing such a tricky issue 🚀

lstocchi (approved):
LGTM codewise, i was not able to replicate the problem even before so i could not test it but Ondrej verified it

axel7083:
Thanks @lstocchi and @odockal for the testing and feedback.

Two approval: enabling auto-merge :heavy_check_mark:

odockal:
@axel7083 Thanks for the patch and @lstocchi for feedback! During the release testing we will take the patch into account and see if we can close some other issue - possibly related.

Issue #5976, opened 2024-02-13
Kind Resource QE Coverage


odockal:
### Epic domain

Kind feature was implemented, we need to make sure, to have QE coverage plan for this feature and taken into account.

Consists of multiple steps

* [x] QE doc: document Kind feature, prepare examples and steps to verify its functionality
* [x] https://github.com/containers/podman-desktop/issues/8445
* [x] #7977
* [x] https://github.com/containers/podman-desktop/pull/7976
* [x] https://github.com/containers/podman-desktop/pull/8151
* [x] #8053
* [x] Test automation on Linux (due to limitations of the infrastructure we do not have linux tests now)
* [x] https://github.com/containers/podman-desktop/issues/8169
* [x] https://github.com/containers/podman-desktop/issues/8944

### Additional context

Kind is a lightweight kubernetes cluster for local development. It by default requires rootfull podman machine (windows/Mac), on linux is bit different, try to explore, one option is to run it locally and something else is how it will behave on GHA ubuntu runner.

Kind E2E test implementation can be broken down to multiple steps.
1. Install kind via system tray
2. Handle system wide installation requirement (system native privilege confirmation)
3. Kind Built-in extension in Podman Desktop (explore extensions and how it behaves)
4. Kind connection in Resources Page
5. Creating of a Kind cluster (with ingress controller)
6. Verification of the created cluster via podman desktop and basic operations -> restart, stop, start, delete
7. Check for resources added with kind cluster (images, containers, volumes, etc)
8. Switching to newly created kube-context
9. Deploying a pod/container for a Kubernetes cluster via Deploy To Kubernetes option
10. verification of the deployed app.

Resources: 
* https://podman-desktop.io/docs/kind
* https://kind.sigs.k8s.io/docs/user/rootless/ (in case of running on linux rootless)

PR #7976, opened 2024-07-08
test: add installation component for new e2e test of Kind feature

amisskii:

### What does this PR do?
This PR introduces the installation component of a new end-to-end test for the Kind feature, which was recently added to the Podman Desktop Application.

#5976 #7977 

### How to test this PR?

Run the tests with yarn test:e2e:run

### DIFF
--- a/tests/playwright/src/model/pages/resources-page.ts
+++ b/tests/playwright/src/model/pages/resources-page.ts
@@ -25,6 +25,7 @@ export class ResourcesPage extends SettingsPage {
   readonly featuredProviderResources: Locator;
   readonly podmanResources: Locator;
   readonly composeResources: Locator;
+  readonly kindResources: Locator;
 
   constructor(page: Page) {
     super(page, 'Resources');
@@ -32,5 +33,6 @@ export class ResourcesPage extends SettingsPage {
     this.featuredProviderResources = this.content.getByRole('region', { name: 'Featured Provider Resources' });
     this.podmanResources = this.featuredProviderResources.getByRole('region', { name: 'podman', exact: true });
     this.composeResources = this.featuredProviderResources.getByRole('region', { name: 'Compose', exact: true });
+    this.kindResources = this.featuredProviderResources.getByRole('region', { name: 'kind', exact: true });
   }
 }
--- a/tests/playwright/src/model/workbench/status-bar.ts
+++ b/tests/playwright/src/model/workbench/status-bar.ts
@@ -1,0 +1,40 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Locator, Page } from '@playwright/test';
+
+import { handleConfirmationDialog } from '../../utility/operations';
+import { BasePage } from '../pages/base-page';
+
+export class StatusBar extends BasePage {
+  readonly kindInstallationButton: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.kindInstallationButton = this.page.getByTitle(
+      'Kind not found on your system, click to download and install it',
+    );
+  }
+
+  public async installKindCLI(): Promise<void> {
+    await this.kindInstallationButton.click();
+    await handleConfirmationDialog(this.page, 'Kind');
+    await handleConfirmationDialog(this.page, 'Kind');
+    await handleConfirmationDialog(this.page, 'Kind', true, 'OK');
+  }
+}
--- a/tests/playwright/src/specs/kind.spec.ts
+++ b/tests/playwright/src/specs/kind.spec.ts
@@ -1,0 +1,87 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { type Page } from '@playwright/test';
+import { expect as playExpect } from '@playwright/test';
+import { afterAll, beforeAll, beforeEach, describe, test } from 'vitest';
+
+import { ExtensionsPage } from '../model/pages/extensions-page';
+import { ResourcesPage } from '../model/pages/resources-page';
+import { WelcomePage } from '../model/pages/welcome-page';
+import { NavigationBar } from '../model/workbench/navigation';
+import { StatusBar } from '../model/workbench/status-bar';
+import { PodmanDesktopRunner } from '../runner/podman-desktop-runner';
+import type { RunnerTestContext } from '../testContext/runner-test-context';
+import { waitForPodmanMachineStartup } from '../utility/wait';
+
+let pdRunner: PodmanDesktopRunner;
+let page: Page;
+let navigationBar: NavigationBar;
+let resourcesPage: ResourcesPage;
+let statusBar: StatusBar;
+const extensionLabel: string = 'podman-desktop.kind';
+const skipKindInstallation = process.env.SKIP_KIND_INSTALL ? process.env.SKIP_KIND_INSTALL : false;
+
+beforeAll(async () => {
+  pdRunner = new PodmanDesktopRunner();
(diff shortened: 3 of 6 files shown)


tests/playwright/src/model/pages/resources-page.ts
```diff
@@ -25,12 +25,18 @@ export class ResourcesPage extends SettingsPage {
...
   constructor(page: Page) {
     super(page, 'Resources');
     this.heading = this.header.getByRole('heading', { name: 'Title' }).and(this.header.getByText('Resources'));
     this.featuredProviderResources = this.content.getByRole('region', { name: 'Featured Provider Resources' });
     this.podmanResources = this.featuredProviderResources.getByRole('region', { name: 'podman', exact: true });
     this.composeResources = this.featuredProviderResources.getByRole('region', { name: 'Compose', exact: true });
+    this.resource = this.page.getByRole('region', { name: 'Resources' });
+  }
+
+  public ResourceExists(label: string): Locator {
```
cbr7: methods should not be named starting with a capital letter and also this naming implies the methods should return a boolean value with value true with the resource Exists.

tests/playwright/src/model/workbench/status-bar.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.dialogOkButton = this.kindDialogWindow.getByRole('button', { name: 'OK' });
+    this.dialogCloseButton = this.kindDialogWindow.getByRole('button', { name: 'Close' });
+  }
+
+  public async installKindCLI(): Promise<void> {
+    await this.kindInstallationButton.click({ timeout: 5000 });
+    await expect(this.kindDialogWindow).toBeVisible();
+    await this.dialogYesButton.click({ timeout: 5000 });
+    await this.dialogYesButton.click({ timeout: 5000 });
+    await this.dialogCloseButton.click({ timeout: 5000 });
```
cbr7: Default timeout is 5 seconds, adding it manually does nothing.

tests/playwright/src/model/workbench/status-bar.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public getKindInstallationButton(): Locator {
+    return this.kindInstallationButton;
```
cbr7: What is the point of this method? We can directly access the locator field that it returns if it's specifically needed.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+});
+
+afterAll(async () => {
+  await pdRunner.close();
+});
+
+describe('Kind End-to-End Tests', async () => {
+  test.skipIf(isCI && isLinux)('Install Kind CLI if Not Present', async () => {
+    await navigationBar.openSettings();
+    if ((await resourcePage.ResourceExists('kind').isVisible({ timeout: 5000 })) === false && !isLinux) {
```
cbr7: isVisible() methods from PW does not benefit from timeout, it will always return instantly the value as it's currently in the system, it has no implicit waiter baked in. Also I'm not at all sure I understand why you are checking equality agaisnt false && !isLinux given that the test is skipped if it is linux.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+      await statusBar.installKindCLI();
```
cbr7: There's no waiter or validation for the installation from what I see,

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  });
+  test('Verify Kind CLI Installation', async () => {
+    await navigationBar.openSettings();
+    await playExpect(statusBar.getKindInstallationButton()).not.toBeVisible({ timeout: 5000 });
+    await playExpect(resourcePage.ResourceExists('kind')).toBeVisible({ timeout: 5000 });
```
cbr7: Default timeout is 5000 ms, so there is no point to specifying it as such.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  });
+  test('Explore and manage Kind extension lifecycle', async () => {
+    const extensionsPage = new ExtensionsPage(page);
+    await navigationBar.openExtensions();
+    await playExpect.poll(async () => await extensionsPage.extensionIsInstalled(extensionLabel)).toBeTruthy();
+    await playExpect(extensionsPage.getExtension(extensionLabel).getByText('ACTIVE')).toBeVisible();
```
cbr7: There is a method in the POM that returns the status of an extension, you do not need to manually grab the status field in the test, in general handling the web elements themselves inside the actual tests should be avoided as much as possible since it defeats the purpose of using POM

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    await extensionsPage.getExtension(extensionLabel).getByRole('button', { name: 'Stop' }).click();
```
cbr7: This should be encapsulated and exposed as a method from POM, not handled directly in the test.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    await playExpect(extensionsPage.getExtension(extensionLabel).getByText('DISABLED')).toBeVisible();
```
cbr7: Same as above.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    await navigationBar.openSettings();
+    await playExpect(resourcePage.ResourceExists('kind')).not.toBeVisible({ timeout: 5000 });
+    await navigationBar.openExtensions();
+    await extensionsPage.getExtension(extensionLabel).getByRole('button', { name: 'Start' }).click();
```
cbr7: Same as above for the Stop button,.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  });
+  test('Explore and manage Kind extension lifecycle', async () => {
+    const extensionsPage = new ExtensionsPage(page);
+    await navigationBar.openExtensions();
+    await playExpect.poll(async () => await extensionsPage.extensionIsInstalled(extensionLabel)).toBeTruthy();
+    await playExpect(extensionsPage.getExtension(extensionLabel).getByText('ACTIVE')).toBeVisible();
+    await extensionsPage.getExtension(extensionLabel).getByRole('button', { name: 'Stop' }).click();
+    await playExpect(extensionsPage.getExtension(extensionLabel).getByText('DISABLED')).toBeVisible();
+    await navigationBar.openSettings();
+    await playExpect(resourcePage.ResourceExists('kind')).not.toBeVisible({ timeout: 5000 });
```
cbr7: Please see previous comment about ResourceExists() and how it should be refactored, it should return true or false

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    await navigationBar.openExtensions();
+    await extensionsPage.getExtension(extensionLabel).getByRole('button', { name: 'Start' }).click();
+    await navigationBar.openSettings();
+    await playExpect(resourcePage.ResourceExists('kind')).toBeVisible({ timeout: 5000 });
```
cbr7: Same as above.

cbr7 (changes requested):
Left some comments.

package.json
```diff
@@ -50,7 +50,7 @@
     "test:unit": "npm run test:main && npm run test:preload && npm run test:preload-docker-extension && npm run test:preload-webview && npm run test:ui && npm run test:renderer && npm run test:scripts:stylesheet && npm run test:tools && npm run test:extensions",
     "test:e2e": "npm run test:e2e:build && npm run test:e2e:run",
     "test:e2e:build": "cross-env NODE_ENV=development MODE=development DEBUG=pw:browser npm run build",
-    "test:e2e:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/playwright/src/specs/ --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",
+    "test:e2e:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/playwright/src/specs/kind.spec.ts --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",
```
cbr7: Please revert this change, it will make it so only this test file would be executed.

cbr7 (changes requested):

cbr7 (approved):

tests/playwright/src/utility/operations.ts
```diff
@@ -155,6 +155,13 @@ export async function handleConfirmationDialog(
   await button.click();
 }
 
+export async function closeConfirmationDialog(page: Page, dialogTitle = 'Confirmation'): Promise<void> {
```
odockal: I think that you can improve `handleConfimationDialog` function so it take another two parameters (type string, first `confirmationButton` with default value "Yes" and `cancelButton` with default value `Cancel`) which would then work well for your kind installation dialogs.

tests/playwright/src/model/workbench/status-bar.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Locator, Page } from '@playwright/test';
+
+import { closeConfirmationDialog, handleConfirmationDialog } from '../../utility/operations';
+import { BasePage } from '../pages/base-page';
+
+export class StatusBar extends BasePage {
```
odockal, line 24: You need to add also this class to the `index.ts` so it will be part of npm `@podman-desktop/tests-playwright` package.

odockal (changes requested):
Do not forget to all add env. var into `e2e-test-main.yaml` e2e test job.

odockal:
Workflow run: https://github.com/containers/podman-desktop/actions/runs/9889805221

odockal (approved):

odockal (approved):

PR #7978, opened 2024-07-08
fix: prevent submit event propagation when clicking on number input buttons

### LINKED ISSUE #7660: Changing port while creating Kind cluster triggers creation
### Bug description

I have a Kind cluster already running, so when I wanted to create a second one I gave it a new name and then clicked the + button to use a non-conflicting port. This triggered it to try to create a Kind cluster and error out, because the second port still conflicted. When I clicked to change the second port the screen switched to confirm creation.

### Operating system

macOS 14.5

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This fixes https://github.com/containers/podman-desktop/issues/7660 . It just adds a preventDefault to prevent the default submit behavior of a button that submit the form when changing the port

### Screenshot / video of UI

N/A

it resolves #7660 

### How to test this PR?

1. go to the create kind cluster page
2. click on the +/- buttons to change the ports. 
3. the ports value should change without any other behavior.

### DIFF
--- a/packages/renderer/src/lib/ui/NumberInput.svelte
+++ b/packages/renderer/src/lib/ui/NumberInput.svelte
@@ -41,8 +41,8 @@ function onDecrement(e: MouseEvent) {
 }
 
 function onIncrement(e: MouseEvent) {
-  value = Number(value) + 1;
   e.preventDefault();
+  value = Number(value) + 1;
 }
 </script>
 
--- a/packages/renderer/src/lib/ui/NumberInput.svelte
+++ b/packages/renderer/src/lib/ui/NumberInput.svelte
@@ -35,12 +35,14 @@ function onKeyPress(event: any) {
   }
 }
 
-function onDecrement() {
+function onDecrement(e: MouseEvent) {
   value = Number(value) - 1;
+  e.preventDefault();
 }
 
-function onIncrement() {
+function onIncrement(e: MouseEvent) {
   value = Number(value) + 1;
+  e.preventDefault();
 }
 </script>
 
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/ui/NumberInput.svelte
```diff
@@ -35,12 +35,14 @@ function onKeyPress(event: any) {
   }
 }
 
-function onDecrement() {
+function onDecrement(e: MouseEvent) {
   value = Number(value) - 1;
+  e.preventDefault();
```
axel7083:
```suggestion
  e.preventDefault();
  value = Number(value) - 1;
```

packages/renderer/src/lib/ui/NumberInput.svelte
```diff
@@ -35,12 +35,14 @@ function onKeyPress(event: any) {
...
 }
 
-function onIncrement() {
+function onIncrement(e: MouseEvent) {
   value = Number(value) + 1;
+  e.preventDefault();
```
axel7083:
```suggestion
  e.preventDefault();
  value = Number(value) + 1;
```

axel7083 (approved):
LGMT, I think the convention is to prevent default at the top of the function

deboer-tim (approved):

lstocchi:
@axel7083 out of curiosity, where did you check what is the convention? Looks to me that everyone put it where they prefer. In MDN they put it at the end in one example, in the middle in another https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault. 
BTW merging your stuff as i don't have a preference.

axel7083:
> @axel7083 out of curiosity, where did you check what is the convention? Looks to me that everyone put it where they prefer. In MDN they put it at the end in one example, in the middle in another https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault. BTW merging your stuff as i don't have a preference.

Good question, can't recall actually :/

PR #7995, opened 2024-07-09
Font sizes and UI consistency

deboer-tim:
### What does this PR do?

Stevan and I spent some time yesterday walking through the UI and trying to
address a few related issues:
- The base (default) font size is too big.
- As a result just about everything has a text-sm or -xs, or even a hardcoded
  font size instead of fixing the core problem.
- Many pages have 3, 4, or even 5 font sizes, and then several variations of
  font weight from light to bold.
- We've been through at least 4 design iterations, and it shows. Need to
  clean up and make more of the UI consistent, and

Unfortunately, the only way to address 1+2 is to reset the default font size,
which means *everything* will need a minor font change. So, here we go! The
core philosophy of this change is:
- Regular controls like Buttons, Inputs, and text descriptions should all be
  at the base (default) font size.
- The default size should be reduced from 16px to 12px. (but this would at
  least be configurable now)
- Most other text (see below) should only be -lg or -sm, with only a couple
  variations of weight.

This first commit changes the font sizes in Tailwind config, applies the new
default in index.html, and fixes the core components. Next commits will each
cover a different area:

Nav/form/settings pages
- Titles in text-lg, font-bold
- Breadcrumbs in text-sm but keep close button text-lg so it's clickable
- Align things between settings nav and settings pages

Tables:
- Header should be text-sm and font-semibold
- Column font regular size, except secondary lines should be text-sm and not
  highlighted
- Labels text-sm

Forms:
- Field labels in font-bold
- Everything justified
- Later: bottom button bar right justified like in dialogs

### Screenshot / video of UI

Screenshot 2024-07-09 at 8 05 18 AM

Screenshot 2024-07-09 at 8 05 28 AM

Screenshot 2024-07-09 at 8 05 45 AM

N/A

### How to test this PR?

Just walk through the UI.


axel7083 (commented):
I feel like some icons got too small to even be noticed, or properly understood

Especially the one in the status bar

image

The profile icon seems to be not at the right size compared to others

image

Also we are losing some centering

| Before | After |
| --- | --- |
| image | image |

jeffmaury (commented):
I found some elements are too small:

- status bar
- tooltip on the navigation bar
- extension page

ekidneyrh:
Some of the elements appear tiny. I know it's a WIP, but I wouldn't have anything below 10px. I think the base text being 12px is fine, but if we are to stick to that, I'd recommend only having one text size below that.

The 10px (sm in this case) should only be used for labels / connection statuses and similar elements. I wouldn't have important information in that size like the description on the extensions as an example:

image

I'm aware we don't like the larger text but I don't want this to become an accessibility concern. We have an option for users to make the text smaller / larger in the settings so I would highlight that for users who prefer smaller text.

deboer-tim:
I'm working my way through extensions pages, those are definitely too small atm as they hadn't been fixed. Will increase tooltips at the same time, and look at status bar.

I'm recommending we bump text-sm up from 10 to 10.5px, that makes it still noticeably smaller for me but more readable. And yes, text-xs should be extremely rare.

deboer-tim:
Pushed another commit:
- Fixed extensions/catalog pages.
- Increased text-sm by 0.5 to 10.5px (a little more readable and helps status bar).
- Improve status bar - refactored so selection is full height, everything is centered correctly.
- Fix tooltip font size.

ekidneyrh:
Looking a lot better :)

deboer-tim:
Rebased to fix conflict and raise minimum font sizes. As discussed on standup/UX call I'm putting this up for review, happy to still fix issues but due to the massive size the goal is to get this to merge (or rejected) asap, and we'll continue work in this area through more targeted PRs.

ekidneyrh (approved):
LGTM! As said in your above comment, if there's anything missed, it can be addressed in another PR.

lstocchi (approved):
LGTM, i agree with Emma. Let's merge this huge beast and tackle issues in following PRs otherwise you will waste a lot of time rebasing.

cdrage (approved):
Tested, looks great and really appealing to the eye / more consistency. LGTM!

PR #7999, opened 2024-07-09
fix: resize terminal so it is entirely visible

### LINKED ISSUE #7791: Unable to see bottom of terminal / input when going to Kubernetes tab with data
### Bug description

https://github.com/containers/podman-desktop/assets/6422176/5e367ab9-8061-466b-89d0-80fb588f2120

When going to the terminal tab when there has been data outputted to the bottom, I'm unable to view / enter information / it's covered.

### Operating system

mac m1

### Installation Method

None

### Version

1.11.0

### Steps to reproduce

1. Press enter a bunch of times to create space on the terminal screen
2. Go to the tab.

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

it updates the style applied to the terminal div parents so that the terminal and its content is completely visible

### Screenshot / video of UI

terminal

it resolves #7791 

### How to test this PR?

1. go to the details page of a pod -> terminal tab -> ress enter a bunch of times to create space on the terminal screen -> you should always be able to see the last row

### DIFF
--- a/packages/renderer/src/lib/pod/KubernetesTerminal.svelte
+++ b/packages/renderer/src/lib/pod/KubernetesTerminal.svelte
@@ -153,4 +153,4 @@ function saveTerminalState(podName: string, containerName: string, state: State)
 }
 </script>
 
-<div class="h-full" bind:this="{terminalXtermDiv}"></div>
+<div class="h-full w-full" bind:this="{terminalXtermDiv}"></div>
--- a/packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
+++ b/packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
@@ -38,46 +38,50 @@ function handleSelectionChange(event: Event) {
 }
 </script>
 
-<div class="flex py-2">
-  <label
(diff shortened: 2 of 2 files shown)


axel7083 (approved):
Works nicely for Terminal&Tty 🚀 great !

deboer-tim (approved):
Approving since the behaviour is much better than before and solves the core issue, but I noticed two things while testing:
- The terminal doesn't always reach the bottom of the window, so (especially in light mode) you're left with a strip of white at the bottom. This does happen to the other tabs (Logs seemed slightly better behaved but same thing; Inspect/Kube have a single line) so maybe we need a separate issue - even just changing the bg to match the terminal would cover it up fine.
- If I resize the window I can reproduce the original problem, I think we're missing some code to handle resizing as well.

Screenshot 2024-07-09 at 10 16 23 AM

lstocchi:
@deboer-tim found that adding a min height fixes the problem when resizing.
terminal-resize

i'm going to open an issue for the first point -> https://github.com/containers/podman-desktop/issues/8006

PR #8020, opened 2024-07-10
feat: add confirmation when bulk-delete if required based on preferences

### LINKED ISSUE #7591: Add confirmation when bulk-deleting

Screenshot 2024-06-11 at 4 20 38 PM

Although we can multi-select, we should really have a confirmation button when bulk-deleting since we are dealing with important concepts such as storage, etc.

Add confirmation for bulk deletion.

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?

Requires user confirmation when bulk-deleting if this option is on in preferences

### Screenshot / video of UI

Screenshot from 2024-07-10 10-27-16

Closes https://github.com/containers/podman-desktop/issues/7591

### How to test this PR?

Try bulk-delete in any of the list pages

### DIFF
--- a/packages/renderer/src/lib/container/ContainerList.svelte
+++ b/packages/renderer/src/lib/container/ContainerList.svelte
@@ -373,7 +373,11 @@ $: containersAndGroups = containerGroups.map(group =>
     {#if selectedItemsNumber > 0}
       <div class="inline-flex space-x-2">
         <Button
-          on:click="{() => withBulkConfirmation(deleteSelectedContainers, `delete ${selectedItemsNumber} containers`)}"
+          on:click="{() =>
+            withBulkConfirmation(
+              deleteSelectedContainers,
+              `delete ${selectedItemsNumber} container${selectedItemsNumber > 1 ? 's' : ''}`,
+            )}"
           aria-label="Delete selected containers and pods"
           title="Delete {selectedItemsNumber} selected items"
           bind:inProgress="{bulkDeleteInProgress}"
--- a/packages/renderer/src/lib/container/ContainerList.svelte
+++ b/packages/renderer/src/lib/container/ContainerList.svelte
@@ -26,6 +26,7 @@ import { podsInfos } from '../../stores/pods';
 import { providerInfos } from '../../stores/providers';
 import { findMatchInLeaves } from '../../stores/search-util';
 import { viewsContributions } from '../../stores/views';
+import { withBulkConfirmation } from '../actions/BulkActions';
 import type { ContextUI } from '../context/context';
 import Dialog from '../dialogs/Dialog.svelte';
 import type { EngineInfoUI } from '../engine/EngineInfoUI';
@@ -372,7 +373,7 @@ $: containersAndGroups = containerGroups.map(group =>
     {#if selectedItemsNumber > 0}
       <div class="inline-flex space-x-2">
         <Button
-          on:click="{() => deleteSelectedContainers()}"
+          on:click="{() => withBulkConfirmation(deleteSelectedContainers, `delete ${selectedItemsNumber} containers`)}"
           aria-label="Delete selected containers and pods"
           title="Delete {selectedItemsNumber} selected items"
           bind:inProgress="{bulkDeleteInProgress}"
--- a/packages/renderer/src/lib/deployments/DeploymentsList.spec.ts
+++ b/packages/renderer/src/lib/deployments/DeploymentsList.spec.ts
@@ -33,13 +33,9 @@ import { kubernetesCurrentContextDeployments } from '/@/stores/kubernetes-contex
 import DeploymentsList from './DeploymentsList.svelte';
 
 const kubernetesRegisterGetCurrentContextResourcesMock = vi.fn();
-const getConfigurationValueMock = vi.fn();
-const deleteDeploymentMock = vi.fn();
 
 beforeAll(() => {
   (window as any).kubernetesRegisterGetCurrentContextResources = kubernetesRegisterGetCurrentContextResourcesMock;
-  (window as any).kubernetesDeleteDeployment = deleteDeploymentMock;
-  (window as any).getConfigurationValue = getConfigurationValueMock.mockResolvedValue(false);
 });
 
 beforeEach(() => {
@@ -48,8 +44,10 @@ beforeEach(() => {
   (window as any).kubernetesGetContextsGeneralState = () => Promise.resolve(new Map());
   (window as any).kubernetesGetCurrentContextGeneralState = () => Promise.resolve({});
   (window as any).window.kubernetesUnregisterGetCurrentContextResources = () => Promise.resolve(undefined);
-  (window as any).kubernetesDeleteDeployment = deleteDeploymentMock;
-  (window as any).getConfigurationValue = getConfigurationValueMock.mockResolvedValue(false);
+  (window as any).getConfigurationValue = vi.fn();
+  vi.mocked(window.getConfigurationValue).mockResolvedValue(false);
+  (window as any).kubernetesDeleteDeployment = vi.fn();
+  vi.mocked(window.kubernetesDeleteDeployment);
 });
 
 async function waitRender(customProperties: object): Promise<void> {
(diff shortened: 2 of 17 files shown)


packages/renderer/src/lib/deployments/DeploymentsList.svelte
```diff
@@ -126,7 +127,14 @@ const row = new TableRow<DeploymentUI>({ selectable: _deployment => true });
...
       <Button
-        on:click="{() => deleteSelectedDeployments()}"
+        on:click="{() =>
+          window
+            .getConfigurationValue('userConfirmation.bulk')
+            .then(value =>
+              value
+                ? withConfirmation(deleteSelectedDeployments, `delete ${selectedItemsNumber} deployments`)
+                : deleteSelectedDeployments(),
+            )}"
```
axel7083: Maybe having an utility function somewhere similar to withConfirmation

Something looking like

```ts
export async function withBulkConfirmation(callback: () => void, text: string): Promise<void> {
  const confirm: boolean = await window.getConfigurationValue('userConfirmation.bulk');
  if(confirm) return withConfirmation(callback, text);
  else callback();
}
```

packages/renderer/src/lib/pvc/PVCList.spec.ts
```diff
@@ -16,11 +16,16 @@
...
+
 import '@testing-library/jest-dom/vitest';
 
 import type { V1PersistentVolumeClaim } from '@kubernetes/client-node';
-import { render, screen } from '@testing-library/svelte';
+import { fireEvent, render, screen } from '@testing-library/svelte';
+/* eslint-disable import/no-duplicates */
+import { tick } from 'svelte';
 import { get } from 'svelte/store';
+/* eslint-enable import/no-duplicates */
```
axel7083, line 28: Particular reason why this is needed ?

packages/renderer/src/lib/volume/VolumesList.spec.ts
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2024 Red Hat, Inc.
```
axel7083, line 2: Nice having the headers updated! 👍

packages/renderer/src/lib/deployments/DeploymentsList.spec.ts
```diff
@@ -154,3 +160,45 @@ test('Expect filter empty screen', async () => {
...
+      replicas: 2,
+      selector: {},
+      template: {},
+    },
+  };
+  kubernetesRegisterGetCurrentContextResourcesMock.mockResolvedValue([deployment]);
+
+  while (get(kubernetesCurrentContextDeployments).length === 0) {
+    await new Promise(resolve => setTimeout(resolve, 500));
+  }
```
axel7083: Let's avoid timeouts, and do something more `vitest` friendly.

I would recommend looking at https://vitest.dev/api/vi.html#vi-waitfor

I know this is some code that is already inside those test, but I believe this should be avoided 

```suggestion
  await vi.waitFor(() => get(kubernetesCurrentContextDeployments).length > 0);
```

packages/renderer/src/lib/deployments/DeploymentsList.spec.ts
```diff
@@ -33,9 +33,13 @@ import { kubernetesCurrentContextDeployments } from '/@/stores/kubernetes-contex
 import DeploymentsList from './DeploymentsList.svelte';
 
 const kubernetesRegisterGetCurrentContextResourcesMock = vi.fn();
+const getConfigurationValueMock = vi.fn();
```
axel7083: I know this is the way it is already done in this file, but this is not recommended has we lose type safety.

I will provide an example with the existing kubernetesRegisterGetCurrentContextResources.

```ts
beforeAll(() => {
  (window as any).kubernetesRegisterGetCurrentContextResources = vi.fn();
});

// when you need to mock kubernetesRegisterGetCurrentContextResources  you do the following

vi.mocked(window.kubernetesRegisterGetCurrentContextResources).mockResolvedValue([]);
```

To give some explanation, `vi.fn()` is kinda `(args: any[]): any` function type. This is a bit problematic if we change later by something different the signature, as typecheck will not be able to catch that in our test, making it harder to understand and debug where the problem may come from.

axel7083 (commented):
Left some comments

packages/renderer/src/lib/deployments/DeploymentsList.svelte
axel7083: ℹ️ On github don't hesite to respond / `Resolve` / 👍  the comments once they are addressed!

packages/renderer/src/lib/ingresses-routes/IngressesRoutesList.svelte
```diff
@@ -170,7 +171,8 @@ const row = new TableRow<IngressUI | RouteUI>({ selectable: _ingressRoute => tru
   <svelte:fragment slot="bottom-additional-actions">
     {#if selectedItemsNumber > 0}
       <Button
-        on:click="{() => deleteSelectedIngressesRoutes()}"
+        on:click="{() =>
+          withBulkConfirmation(deleteSelectedIngressesRoutes, `delete ${selectedItemsNumber} Ingresses Routes`)}"
```
deboer-tim: Ingresses and routes are two separate kinds of things, we're just showing them in one table because they are similar and you either have one or the other, depending on whether you're running OpenShift or not.

So, you either need to get fancy and use code to figure out what you're deleting, or change it to something like `ingresses/routes` or `ingresses or routes`.

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -372,7 +373,7 @@ $: containersAndGroups = containerGroups.map(group =>
     {#if selectedItemsNumber > 0}
       <div class="inline-flex space-x-2">
         <Button
-          on:click="{() => deleteSelectedContainers()}"
+          on:click="{() => withBulkConfirmation(deleteSelectedContainers, `delete ${selectedItemsNumber} containers`)}"
```
deboer-tim: Software with incorrect text like '1 containers' is a pet peeve of mine. :) Could you please either use an if statement or a `{ selectedItendsNumber > 1 ? 's':''}` block so that it is '1 container', '2 containers'? (or reword to avoid it somehow)

deboer-tim (commented):
Just a couple minor comments, looks good otherwise and nice improvement!

packages/renderer/src/lib/pvc/PVCList.spec.ts
SoniaSandler: Since I added the import of tick, the linter puts them together in one import to be `import {get, tick} from 'svelte'` which is incorrect

packages/renderer/src/lib/deployments/DeploymentsList.spec.ts
SoniaSandler: I changed it only in some of my tests, because in a few of them, this proposed change caused some problems with the rendering of components (not enough time?) leading to errors like `Unable to find an accessible element with the role...`

SoniaSandler:
@axel7083 I had to update my tests  because i realized that they were wrong (used `mockResolvedValueOnce` and `toHaveBeenCalledOnce` instead of without the `Once`). However, the fix lead to the tests returning `expected "spy" to be called at least once` to checks like `expect(window.removeVolume).toHaveBeenCalled();`, `expect(window.kubernetesDeletePersistentVolumeClaim).toHaveBeenCalled();`, etc. in all of my tests. Those lines are supposed to be under the `expect(window.showMessageBox).toHaveBeenCalledTimes(2);` lines when returning 0 (yes) to the `showMessageBox` value. For now, I have left the second `showMessageBox` and click to see if there is a way for the tests to recognize the `window.delete/remove` calls, but if it isn't possible, I'll remove the duplications.

axel7083 (approved):
Great job ! Thanks for the changes 🚀

deboer-tim (approved):
LGTM, thanks for the changes as well!

PR #8045, opened 2024-07-12
feat: provide podman ext API to call podman CLI

### LINKED ISSUE #5990: Provide API to call podman CLI

dgolovin:
### What does this PR do?

* adds package with Podman Extension API declaration file
* changes podman extension activate method return to PodmanExtensionApi instance
* adds next and latest package publishing

### Screenshot / video of UI

n/a

Fix #5990.

### How to test this PR?


packages/podman-extension-api/package.json
```diff
@@ -0,0 +1,23 @@
+{
```
axel7083: I wanted to discuss that with @benoitf, but I feel like it would be easier if the types of the extensions would be in the extension folder itself.

This is not really possible today, as to publish a npm package the `name` property of the `package.json` must be equal to the name of the package. 

However, IMO creating a new package just for the type for each extensions we want to expose an api would be too much... We might want to discuss how we could do this nicely.

I also feel like it is very distrurbing to have 

- `packages/api`
- `packages/extension-api`
- `packages/podman-extension-api`

where it feel messy

packages/podman-extension-api/package.json
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  "scripts": {
+    "prepare": "",
+    "clean": "rimraf lib *.tsbuildinfo",
+    "build": "",
+    "watch": "",
+    "publish:next": "yarn publish  --registry=https://registry.npmjs.org/ --no-git-tag-version --new-version 0.0.1-\"$(date +%s)\""
+  },
+  "dependencies": {
+    "@podman-desktop/api": "0.0.1"
```
axel7083: Why would we need to include "@podman-desktop/api" ?

packages/podman-extension-api/src/podman-extension-api.d.ts
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { RunOptions, RunResult } from '@podman-desktop/api';
+
+export interface PodmanExtensionApi {
+  exec(args?: string[], options?: RunOptions): Promise<RunResult>;
```
axel7083: Not sure the `RunOptions` of the @podman-desktop/api are a good match here ?

https://github.com/containers/podman-desktop/blob/4f9e876667fe5ab48d26bd9d2d1ab728e3a89dd5/packages/extension-api/src/extension-api.d.ts#L4033

The isAdmin bother me

packages/podman-extension-api/package.json
dgolovin: See imports in d.ts file.

packages/podman-extension-api/src/podman-extension-api.d.ts
dgolovin: > Not sure the RunOptions of the @podman-desktop/api are a good match here ?

It is a good first iteration

> The isAdmin bother me

Some commands if you run them require admin privileges. it does not mean you getting them right away, it just force PD to request user to login.

packages/podman-extension-api/package.json
dgolovin: That is 'internal implementation' we can start with this location and then move packages around. That won't affect the API itself.

benoitf:
https://github.com/containers/podman-desktop/pull/8247 by @lstocchi  introduced a new helper method in utils to execute podman commands

here it won't handle the env variable

dgolovin:
Thanks for the heads-up! Will use https://github.com/containers/podman-desktop/blob/75e0b95f62b5f0b44f49c25b459e5f07d1f33b31/extensions/podman/src/util.ts#L138 instead.

dgolovin:
@benoitf updated pr to use utli.execPodman function.

extensions/podman/package.json
```diff
@@ -379,6 +379,7 @@
   "dependencies": {
     "@ltd/j-toml": "^1.38.0",
     "@podman-desktop/api": "^0.0.1",
+    "@podman-desktop/podman-extension-api": "^0.0.1",
```
axel7083, line 387: I am wondering if we could improve the meaning/clarity of the packae by adding a `/`

Here is my idea

`@podman-desktop/extension-api/podman`

what do you think @benoitf ?

This would probably be easier for developers, as we could say all `@podman-desktop/extension-api/*` would be extension specific.

benoitf: AFAIK npmjs packages are only `@org-name/package-name`

benoitf: https://docs.npmjs.com/cli/v10/using-npm/scope#publishing-scoped-packages

axel7083: Oh that's so said, thanks for the info

dgolovin:
Re-based on top of main to fix conflicts.

axel7083 (approved):
LGTM

benoitf (dismissed):
if we introduce packages in a folder, then it means we're inside a multi-root mode,

then the root package.json should only contain some common definition and then the existing source code of the application should go inside a subfolder.

here we're bypassing the classic layout by having extension packages being defined at the top level and not at the extension level.

layout should be
extensions/podman/packages/backend (or extension)
extensions/podman/packages/api

so the packages/modules should be side by side.

dgolovin:
> layout should be
> extensions/podman/packages/backend (or extension)
> extensions/podman/packages/api
>
> so the packages/modules should be side by side.

@benoitf updated layout and related files, tested local build on windows to ensure the podman extension is loading properly at runtime. Yarn watch script is also updated. I have not tested the installer for disconnected env.

axel7083 (dismissed):
Tested locally and works fine with `yarn watch`

extensions/podman/packages/api/package.json
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  "scripts": {
+    "prepare": "",
+    "clean": "rimraf lib *.tsbuildinfo",
+    "build": "",
+    "watch": "",
+    "publish:next": "yarn publish  --registry=https://registry.npmjs.org/ --no-git-tag-version --new-version 0.0.1-\"$(date +%s)\""
+  },
+  "dependencies": {
+    "@podman-desktop/api": "0.0.1"
```
benoitf: this dependency looks odd, because if I have a separate extension project, I will probably depends on a specific version of Podman Desktop

like depends on `"@podman-desktop/api": "1.12.0"` and there it'll bring another dependency to a published 0.0.1 which is not expected

benoitf: I guess we're missing in release and next-build an override of the version to match the version we're publising

benoitf: PTAL @dgolovin

packages/main/src/plugin/extension-loader.ts
```diff
@@ -580,7 +580,11 @@ export class ExtensionLoader {
     return entries
       .filter(entry => entry.isDirectory())
       .filter(directory => directory.name !== 'node_modules')
-      .map(directory => path.join(folderPath, directory.name));
+      .map(directory =>
+        fs.existsSync(path.join(folderPath, directory.name, 'package.json'))
+          ? path.join(folderPath, directory.name)
+          : path.join(folderPath, directory.name, 'packages', 'extension'),
+      );
```
benoitf, line 587: We may have extensions having a `package.json` file at the root of the extensions/folder ? like a monorepository ?

in that case it may fail here ?

extensions/podman/packages/api/package.json
dgolovin:
```suggestion
    "@podman-desktop/api": "^1.12.0"
```
Would that work?

benoitf: we'll miss the update on every release

we need to have the version being set when publishing

dgolovin: Oh, of course. I was looking at the last line only (>.<). Thanks for pointing that out.

packages/main/src/plugin/extension-loader.ts
dgolovin: I can give priority to packages/extension folder in case or monorepo.

extensions/podman/packages/api/package.json
dgolovin: While dependency to @podman-desktop/api should be definitely updated to ^1.12.0, I am still not sure under what version api package should be published. 
Does it make sense to increment it every build even there are no changes in the api package?
Should it be just the same as @podman-desktop/api?

benoitf: > Does it make sense to increment it every build even there are no changes in the api package?

we never know what will happen.

but it could also means that we would have inconsistent graph


like: myExtension depending on 'podman-api' ^1.10 and myExtension depending on 'podman-ext" : 1.10 it would also get a transitive dependency through podman-api to a different "podman-ext" version


so probably yes a safe path is to use the same version


> While dependency to @podman-desktop/api should be definitely updated to ^1.12.0, I am still not sure under what version api package should be published.
> Does it make sense to increment it every build even there are no changes in the api package?

I think it's ok as the lifecycle of the podman extension is the same than podman desktop for now.

so I'm not I would understand to use podman-api v1.3 and podman-desktop-api v1.12

dgolovin: Dep to @podman-desktop/api is updated to `^1.12.0` in package.json.  I wonder if I can update it to latest podman-desktop release version during release running 

`yarn add @podman-desktop/api@${{ needs.tag.outputs.desktopVersion }}`

in release.yaml at release time

.github/workflows/release.yaml
```diff
@@ -280,6 +280,12 @@ jobs:
           sed -i  "s#version\":\ \"\(.*\)\",#version\":\ \"${{ needs.tag.outputs.desktopVersion }}\",#g" packages/ui/package.json
           cd packages/ui && yarn build && yarn publish --tag latest --no-git-tag-version --new-version "${{ needs.tag.outputs.desktopVersion }}" --access public
 
+      - name: Publish Podman Extension API to npmjs
+        run: |
+          echo "Using version ${{ needs.tag.outputs.desktopVersion }}"
+          sed -i  "s#version\":\ \"\(.*\)\",#version\":\ \"${{ needs.tag.outputs.desktopVersion }}\",#g" extensions/podman/packages/api/package.json
+          cd extensions/podman/packages/api && yarn publish --tag latest --no-git-tag-version --new-version "${{ needs.tag.outputs.desktopVersion }}" --access public
```
dgolovin, line 287: To have dependency to @podman-desktop/api and @podman-extension/api 
```suggestion
          cd extensions/podman/packages/api && yarn add @podman-desktop/api@${{ needs.tag.outputs.desktopVersion }} && yarn publish --tag latest --no-git-tag-version --new-version "${{ needs.tag.outputs.desktopVersion }}" --access public
```

dgolovin: The publishing is already using `needs.tag.outputs.desktopVersion` as version.

benoitf: next-build.yaml don't but I think it's the same issue in this job ?

benoitf:
@dgolovin would it be possible to split this PR into multiple chunks

1. move `extensions/podman/src` to `extensions/podman/packages/extension`

here I think it's a simple move, extension-loader having to take care of that.

2. add the new `extensions/podman/packages/api` package

I see no issue as well there

3. add the publication of api to npmjs

(here is where it seems we're still discussing)

dgolovin:
> @dgolovin would it be possible to split this PR into multiple chunks
> 
>     1. move `extensions/podman/src` to `extensions/podman/packages/extension`
> 
> 
> here I think it's a simple move, extension-loader having to take care of that.
> 

Created PR #8791.

dgolovin:
> 
>     2. add the new `extensions/podman/packages/api` package
> 
> 
> I see no issue as well there

PR created https://github.com/containers/podman-desktop/pull/8891.

benoitf:
sets draft mode to the PR as it depends on others

dgolovin:
> 5\. add the publication of api to npmjs
> 
> 
> (here is where it seems we're still discussing)

Opened PR for publishing https://github.com/containers/podman-desktop/pull/9061.

Closing this PR.

PR #8076, opened 2024-07-15
chore: use fileinput in run image

### LINKED ISSUE #7938: Use FileInput consistently

deboer-tim:
### What does this PR do?

Uses the FileInput component in the Run Image form, for both the volume mounts and environment controls.

### Screenshot / video of UI

Before:

Screenshot 2024-07-15 at 10 04 33 AM

After:

Screenshot 2024-07-15 at 10 03 13 AM

Fixes #7938.

### How to test this PR?

Run an image and make sure both fields work.

### DIFF
--- a/packages/renderer/src/lib/image/RunImage.svelte
+++ b/packages/renderer/src/lib/image/RunImage.svelte
@@ -601,12 +601,12 @@ async function assertAllPortAreValid(): Promise<void> {
   invalidPorts = invalidHostPorts.length > 0 || invalidContainerPortMapping.length > 0;
 }
 
-let volumeDialogOptions: OpenDialogOptions = {
+const volumeDialogOptions: OpenDialogOptions = {
   title: 'Select a directory to mount in the container',
   selectors: ['openDirectory'],
 };
 
-let envDialogOptions: OpenDialogOptions = {
+const envDialogOptions: OpenDialogOptions = {
   title: 'Select environment file',
   selectors: ['openFile'],
 };
@@ -767,7 +767,6 @@ let envDialogOptions: OpenDialogOptions = {
                     id="filePath.{index}"
                     placeholder="Environment file containing KEY=VALUE items"
                     bind:value="{environmentFile}"
-                    readonly
                     options="{envDialogOptions}"
                     aria-label="environmentFile.{index}" />
                   <Button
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/image/RunImage.svelte
```diff
@@ -773,18 +763,13 @@ async function assertAllPortAreValid(): Promise<void> {
...
-                    <Button
-                      on:click="{() => selectEnvironmentFile(index)}"
-                      id="filePath.{index}"
-                      aria-label="button-select-env-file-{index}">Browse ...</Button>
-                  </div>
+                  <FileInput
+                    id="filePath.{index}"
+                    placeholder="Environment file containing KEY=VALUE items"
+                    bind:value="{environmentFile}"
+                    readonly
```
jeffmaury: Any reason why this is readonly ?

packages/renderer/src/lib/image/RunImage.svelte
```diff
@@ -620,6 +600,16 @@ async function assertAllPortAreValid(): Promise<void> {
   const invalidContainerPortMapping = containerPortMapping?.filter(port => port.error) ?? [];
   invalidPorts = invalidHostPorts.length > 0 || invalidContainerPortMapping.length > 0;
 }
+
+let volumeDialogOptions: OpenDialogOptions = {
```
jeffmaury: Why not const ?

packages/renderer/src/lib/image/RunImage.svelte
deboer-tim: Solely because that's what it was doing before. I don't see any reason to keep that though, so I've removed it in new commit.

packages/renderer/src/lib/image/RunImage.svelte
deboer-tim: Yes, good catch. Updated in new commit.

packages/renderer/src/lib/image/RunImage.svelte
feloy: My removing the readonly property, the path is settable in two ways: either by opening the system file browser, or by typing (or copy-pasting) the path directly in the field. It is set in some places, but not everywhere. What would be the general desired setting? (cc @ekidneyrh)

deboer-tim: I'm going to go ahead and say we should almost always allow copy/paste or editing because that's what I prefer :) - but that also greatly increases the odds that the file is invalid so we have to make sure each form has good validation.

feloy (approved):

cdrage (approved):
Tested and LGTM! 

Went over the code looks good.

PR #8087, opened 2024-07-15
chore: reduce unused telemetry events

### LINKED ISSUE #4970: Reduce overall number of telemetry events

deboer-tim:
### What does this PR do?

As we've hit telemetry thresholds we disabled several events that were output extremely often but we either weren't using or had questionable value vs the load. This included the following events:

- shellInContainer
- containerInspect
- loadExtension
- kubernetesReadNamespacedPod

With continued volume on each of these we will never re-enable as-is, and we should not continue sending them in (using up some minor bandwidth) if they will never be used.

Instead of removing them entirely, this PR renames each event (so that we can track/filter it separately from the current event) and only sends it when there is a failure. These should be _far_ less frequent and we would still get information on where we have failures.

Since the same pattern as kubernetesReadNamespacedPod is used for every Kubernetes object I applied it to all of them. We already track pages to know when someone looks at each kind of object so we don't need to track the underlying API call, and this keeps the code consistent.

### Screenshot / video of UI

N/A

Fixes #4970.

### How to test this PR?

Code check, or testing and checking dev telemetry.

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -1697,7 +1697,6 @@ export class ContainerProviderRegistry {
     onError: (error: string) => void,
     onEnd: () => void,
   ): Promise<{ write: (param: string) => void; resize: (w: number, h: number) => void }> {
-    let telemetryOptions = {};
     try {
       const exec = await this.getMatchingContainer(engineId, id).exec({
         AttachStdin: true,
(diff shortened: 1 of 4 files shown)


axel7083 (approved):
LGTM

jeffmaury (approved):
LGTM

Issue #7757, opened 2024-06-20
Page layout


feloy:
### Is your enhancement related to a problem? Please describe

We have several page layouts (NavPage, DetailsPage, FormPage, SettingsPage). All are very similar in their design, with a few specificities for each:

- a search bar
- a breadcrumb (and a close button)
- an icon in front of the title
- a specific set of colors (background, text, card background, card text, ...)

### Describe the solution you'd like

We could consider having only one page layout, with all these specificities as parameters.
That would help factorise the design of all the pages, and have more consistent pages.

For specific colors, we could consider passing these specific colors in the context, so the colors used in the page layout and inside the fragments would use the local colors instead of colors from the registry.

Schematically:
```
<Page let:bg="pd-content-bg" let:text="pd-content-text">
  <div class="bg-${bg} text-${text}">Some text using colors from the context</div>
</Page>
```

ekidneyrh:
@feloy  I briefly looked around the app and all the pages follow the same structure just +/- some elements

- Breadcrumb
- Icon + Header + buttons
- Subtitle (can be text or link)
- Search bar
- Tabs
- Content (table, list)

I probably shouldn't have assigned this to myself, as I can't do more in terms of the code. Would it be beneficial to have me make a diagram for this? Maybe include the colours on it too? WDYT

deboer-tim:
I know there are some differences between the main nav page and form page header layout, it would probably help to understand how the uber page should be laid out in terms of flex box, e.g. if there are cases where we have no breadcrumb but an X how the header changes. The Container Details page is one of the more complicated ones where I'm not sure the layout is great at different sizes.

But yes, otherwise I think we should assign for implementation.

feloy:
> @feloy I briefly looked around the app and all the pages follow the same structure just +/- some elements
> 
> * Breadcrumb
> * Icon + Header + buttons
> * Subtitle (can be text or link)
> * Search bar
> * Tabs
> * Content (table, list)
> 
> I probably shouldn't have assigned this to myself, as I can't do more in terms of the code. Would it be beneficial to have me make a diagram for this? Maybe include the colours on it too? WDYT

@ekidneyrh, yes some mockups would be helpful IMO, representing the page with the different options set or not (the page with or without breadcrumb, the page with or without icon, etc). And also which colors would be different based on the context (content / inverted content / details / ...)

ekidneyrh:
[...]

PR #8093, opened 2024-07-15
chore: reusable base page

### LINKED ISSUE #7739: Troubleshooting should be a DetailsPage

Now that the FormPage has a single background color and no horizontal line, the Troubleshooting page looks odd because the tabs are floating.

Screenshot 2024-06-19 at 8 41 41 AM

Looks like this page should be a DetailsPage, which would also allow removing some of the 'duplicate' titles.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Provides a single base layout page that has enough options to support the current FormPage and DetailsPages.

This is enough that both Help and Troubleshooting pages can use this new layout instead of FormPage. This means they need to handle the close directly (maybe it's worth another page in the middle to handle this for both?), but it means we can remove the showBreadcrumb option from FormPage to start to clean it up.

### Screenshot / video of UI

N/A

First part of #7757. (NavPage, SettingsPage, or merging in any other pages is not attempted here)

Fixes #7739.

### How to test this PR?

Check several form pages, details pages, troubleshooting, and help for regressions/change, should be none.


packages/ui/src/lib/layouts/BasePage.svelte
```diff
@@ -0,0 +1,105 @@
...
+export let subtitle: string | undefined = undefined;
+export let breadcrumbLeftPart: string | undefined = undefined;
+export let breadcrumbRightPart: string | undefined = undefined;
+export let breadcrumbTitle: string | undefined = '';
+export let hasClose: boolean = true;
+export let inProgress: boolean = false;
+
+let showBreadcrumb = breadcrumbLeftPart ?? breadcrumbRightPart;
+
+const dispatchClose = createEventDispatcher<{ close: undefined }>();
```
axel7083: We are not suppose to use dispatcher with svelte 5

packages/ui/src/lib/layouts/BasePage.svelte
```diff
@@ -0,0 +1,105 @@
...
+import { LinearProgress } from '..';
+import CloseButton from '../button/CloseButton.svelte';
+import Link from '../link/Link.svelte';
+
+export let title: string;
+export let titleDetail: string | undefined = undefined;
+export let subtitle: string | undefined = undefined;
+export let breadcrumbLeftPart: string | undefined = undefined;
+export let breadcrumbRightPart: string | undefined = undefined;
+export let breadcrumbTitle: string | undefined = '';
```
axel7083: Not related to this PR but I know that `left | right` breadcrumb has been used in other component, but I really don't like it, as it limit to only two when we could need more ?

deboer-tim:
Rebased and made a couple fixes/simplifications.

packages/ui/src/lib/layouts/BasePage.svelte
deboer-tim: This PR is merging the dispatcher from two components into a single parent, IMHO that would make reviewing harder and should be a separate PR.

packages/ui/src/lib/layouts/BasePage.svelte
deboer-tim: I agree, but this PR is just merging two copies of that breadcrumb code into one parent, not trying to improve it at this point.

axel7083 (approved):

benoitf:
I would rename `BasePage` to just `Page`

deboer-tim:
> I would rename `BasePage` to just `Page`

Yeah, that's better, renamed.

benoitf (approved):

cdrage (approved):

PR #8104, opened 2024-07-16
chore(test): add some debug console logs

cbr7:
### What does this PR do?
Adds some console logs for debugging image building e2e tests.

https://github.com/containers/podman-desktop-extension-bootc/issues/629

### DIFF
--- a/tests/playwright/src/model/pages/build-image-page.ts
+++ b/tests/playwright/src/model/pages/build-image-page.ts
@@ -58,6 +58,10 @@ export class BuildImagePage extends BasePage {
     contextDirectory: string,
     archType = ArchitectureType.Default,
   ): Promise<ImagesPage> {
+    console.log(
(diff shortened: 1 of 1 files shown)


tests/playwright/src/model/pages/build-image-page.ts
```diff
@@ -58,6 +58,10 @@ export class BuildImagePage extends BasePage {
     contextDirectory: string,
     archType = ArchitectureType.Default,
   ): Promise<ImagesPage> {
+    console.log(
```
axel7083, line 61: Can't we use console.debug ?

cbr7: I would rather not, all the other logging in the test framework is done with console.log, changing it might cause it to be missed by accident.

axel7083 (approved):

PR #8108, opened 2024-07-16
feat: add manifest api for pushing

### LINKED ISSUE #8057: Manifest: Add "push manifest" API

Part of https://github.com/containers/podman-desktop/issues/7041

Add push manifest button

_No response_

### Additional context

_No response_

cdrage:
feat: add manifest api for pushing

### What does this PR do?

Took a while and found a bug regarding trying to push, but here is the
API for manifest pushing.

* Adds the API needed to push a manifest image
* Needs credentials, so does base64 encoding of credentials to API calls
* Requires ?all=true to work correctly.

There are other "optional" paramters (see:
https://docs.podman.io/en/latest/_static/api.html#tag/manifests/operation/ManifestPushLibpod), but we do not need to add them as
they are a not needed at the moment.

### Screenshot / video of UI

N/A, API.

Closes https://github.com/containers/podman-desktop/issues/8057

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -440,7 +440,18 @@ declare module '@podman-desktop/api' {
 
   // Matches required API parameters of Podman options
   export interface ManifestPushOptions {
+    /**
+     * The destination of the manifest list to push.
+     * this can be the same as 'name' or a different one.
+     * @example
+     * "quay.io/myrepo/myotherdestination"
+     */
     destination: string;
+    /**
+     * The name of the manifest list to push
+     * @example
+     * "quay.io/myrepo/mymanifest"
+     */
     name: string;
     // Provider to use for the manifest pushing, if not, we will default to the first one available
     provider?: ContainerProviderConnection;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -437,6 +437,15 @@ declare module '@podman-desktop/api' {
     // Provider to use for the manifest creation, if not, we will try to select the first one available (similar to podCreate)
     provider?: ContainerProviderConnection;
   }
+
+  // Matches required API parameters of Podman options
+  export interface ManifestPushOptions {
+    destination: string;
+    name: string;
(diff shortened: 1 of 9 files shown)


axel7083 (approved):
LGTM ! 🚀 nice unit tests

packages/extension-api/src/extension-api.d.ts
```diff
@@ -437,6 +437,15 @@ declare module '@podman-desktop/api' {
     // Provider to use for the manifest creation, if not, we will try to select the first one available (similar to podCreate)
     provider?: ContainerProviderConnection;
   }
+
+  // Matches required API parameters of Podman options
+  export interface ManifestPushOptions {
+    destination: string;
+    name: string;
```
feloy: Could you please add some comments (with /** to be exported to doc) on these properties, it is not obvious to me what they represent.

packages/main/src/plugin/dockerode/libpod-dockerode.ts
```diff
@@ -842,6 +843,49 @@ export class LibpodDockerode {
...
+          400: 'bad parameter in request',
+          404: 'no such manifest',
+          500: 'server error',
+        },
+        headers: headers,
+        // We require all=true to always be present in the URL in order for the manifest to be pushed correctly.
+        // If you do not provide it, it will return a "uknown manifest blob" error as it's trying to push a manifest blob with no images.
+        options: {
+          all: 'true',
+        },
```
feloy, line 877: You are forcing all=true here, is it necessary to pass this option as a parameter in manifestOptions?

cdrage: Yup, should always enforce all=true or else no point doing push Manifest, so that's why I did not add it to manifestOptions.

I have it as "manifestOptions" to follow the same formatting / style as the other functions in this file.

packages/extension-api/src/extension-api.d.ts
cdrage: you're right! they represent the podman api same names. i'll add more information.

feloy (approved):
LGTM

PR #8128, opened 2024-07-17
test: delete from volume page and prune volume e2e tests

danivilla9:
### What does this PR do?
This PR extends the volume-smoke.spec.ts test suite to include the test cases of deleting a volume from the Volumes page and checking that the correct number of volumes are initialized when starting a container from the **quay.io/centos-bootc/bootc-image-builder** image on Podman Desktop.

https://github.com/containers/podman-desktop/issues/8040

### How to test this PR?
Run the volume tests with yarn test:e2e:smoke

### DIFF
--- a/tests/playwright/src/model/pages/main-page.ts
+++ b/tests/playwright/src/model/pages/main-page.ts
@@ -18,6 +18,7 @@
 
 import type { Locator, Page } from '@playwright/test';
 
+import { waitUntil } from '../../utility/wait';
 import { BasePage } from './base-page';
 
 /**
@@ -70,6 +71,10 @@ export abstract class MainPage extends BasePage {
     return this.content.getByRole('table');
   }
 
+  async rowsAreVisible(): Promise<boolean> {
+    return await this.page.getByRole('row').first().isVisible();
+  }
+
   async getRowFromTableByName(name: string): Promise<Locator | undefined> {
     if (await this.pageIsEmpty()) {
       return undefined;
@@ -93,5 +98,33 @@ export abstract class MainPage extends BasePage {
       console.log(`Exception caught on ${this.title} page with message: ${err}`);
     }
     return undefined;
+  }
+
+  async getRowsFromTableByStatus(status: string): Promise<Array<Locator>> {
+    await waitUntil(async () => await this.rowsAreVisible(), { sendError: false });
+
+    const table = this.content.getByRole('table');
+    if (await table.isVisible()) {
+      const rows = await table.getByRole('row').all();
+      const filteredRows = [];
+      for (let rowNum = 1; rowNum < rows.length; rowNum++) {
+        //skip header
+        const statusCount = await rows[rowNum].getByRole('cell').nth(2).getByTitle(status, { exact: true }).count();
+        if (statusCount > 0) filteredRows.push(rows[rowNum]);
+      }
+      return filteredRows;
+    }
+    return [];
+  }
+
+  async countRowsFromTable(): Promise<number> {
+    await waitUntil(async () => await this.rowsAreVisible(), { sendError: false });
+
+    const table = this.content.getByRole('table');
+    if (await table.isVisible()) {
+      const rows = await table.getByRole('row').all();
+      return rows.length > 1 ? rows.length - 1 : 0;
+    }
+    return 0;
   }
 }
--- a/tests/playwright/src/model/pages/main-page.ts
+++ b/tests/playwright/src/model/pages/main-page.ts
@@ -100,7 +100,7 @@ export abstract class MainPage extends BasePage {
     return undefined;
   }
 
-  async getRowsFromTableByStatus(status: string): Promise<Array<Locator>> {
+  async getRowsFromTableByStatus(status: string): Promise<Locator[]> {
     await waitUntil(async () => await this.rowsAreVisible(), { sendError: false });
 
     const table = this.content.getByRole('table');
--- a/tests/playwright/src/model/pages/main-page.ts
+++ b/tests/playwright/src/model/pages/main-page.ts
@@ -104,27 +104,20 @@ export abstract class MainPage extends BasePage {
     await waitUntil(async () => await this.rowsAreVisible(), { sendError: false });
 
     const table = this.content.getByRole('table');
-    if (await table.isVisible()) {
-      const rows = await table.getByRole('row').all();
-      const filteredRows = [];
-      for (let rowNum = 1; rowNum < rows.length; rowNum++) {
-        //skip header
-        const statusCount = await rows[rowNum].getByRole('cell').nth(2).getByTitle(status, { exact: true }).count();
-        if (statusCount > 0) filteredRows.push(rows[rowNum]);
-      }
-      return filteredRows;
+    const rows = await table.getByRole('row').all();
+    const filteredRows = [];
+    for (let rowNum = 1; rowNum < rows.length; rowNum++) {
+      //skip header
+      const statusCount = await rows[rowNum].getByRole('cell').nth(2).getByTitle(status, { exact: true }).count();
+      if (statusCount > 0) filteredRows.push(rows[rowNum]);
     }
-    return [];
+    return filteredRows;
   }
 
   async countRowsFromTable(): Promise<number> {
     await waitUntil(async () => await this.rowsAreVisible(), { sendError: false });
-
     const table = this.content.getByRole('table');
-    if (await table.isVisible()) {
-      const rows = await table.getByRole('row').all();
-      return rows.length > 1 ? rows.length - 1 : 0;
-    }
-    return 0;
+    const rows = await table.getByRole('row').all();
+    return rows.length > 1 ? rows.length - 1 : 0;
   }
 }
--- a/tests/playwright/src/model/pages/volumes-page.ts
+++ b/tests/playwright/src/model/pages/volumes-page.ts
@@ -101,19 +101,6 @@ export class VolumesPage extends MainPage {
     return true;
   }
 
-  //returns true if the number of volumes with status "UNUSED" is 0 (prune is finished)
-  async waitForPruneVolumes(): Promise<boolean> {
-    let unusedVolumes;
-    await waitWhile(
-      async () => {
-        unusedVolumes = await this.getRowsFromTableByStatus(VolumeState.Unused);
-        return unusedVolumes.length !== 0;
-      },
-      { sendError: false },
-    );
-    return true;
-  }
-
   async pruneVolumes(): Promise<VolumesPage> {
     await playExpect(this.pruneVolumesButton).toBeEnabled();
     await this.pruneVolumesButton.click();
--- a/tests/playwright/src/model/pages/volumes-page.ts
+++ b/tests/playwright/src/model/pages/volumes-page.ts
@@ -87,8 +87,7 @@ export class VolumesPage extends MainPage {
   }
 
   async countUsedVolumesFromTable(): Promise<number> {
-    const usedVolumes = await this.getRowsFromTableByStatus(VolumeState.Used);
-    return usedVolumes.length;
+    return (await this.getRowsFromTableByStatus(VolumeState.Used)).length;
   }
 
   async waitForVolumeExists(name: string): Promise<boolean> {
--- a/tests/playwright/src/model/pages/volumes-page.ts
+++ b/tests/playwright/src/model/pages/volumes-page.ts
@@ -19,7 +19,9 @@
 import type { Locator, Page } from '@playwright/test';
 import { expect as playExpect } from '@playwright/test';
 
+import { handleConfirmationDialog } from '../../utility/operations';
 import { waitUntil, waitWhile } from '../../utility/wait';
+import { VolumeState } from '../core/states';
 import { CreateVolumePage } from './create-volume-page';
 import { MainPage } from './main-page';
 import { VolumeDetailsPage } from './volume-details-page';
@@ -58,6 +60,19 @@ export class VolumesPage extends MainPage {
     return new VolumeDetailsPage(this.page, volumeName);
   }
 
+  async deleteVolume(volumeName: string): Promise<VolumesPage> {
+    const volumeRow = await this.getVolumeRowByName(volumeName);
+    if (volumeRow === undefined) {
+      throw Error(`Volume: ${volumeName} does not exist`);
+    }
+    const containerRowDeleteButton = volumeRow.getByRole('button', { name: 'Delete Volume' });
+    await playExpect(containerRowDeleteButton).toBeEnabled();
+    await containerRowDeleteButton.click();
+    await handleConfirmationDialog(this.page);
+
+    return this;
+  }
+
   async getVolumeRowByName(name: string): Promise<Locator | undefined> {
     return this.getRowFromTableByName(name);
   }
@@ -65,6 +80,15 @@ export class VolumesPage extends MainPage {
   protected async volumeExists(name: string): Promise<boolean> {
     const result = await this.getVolumeRowByName(name);
     return result !== undefined;
+  }
+
+  async countVolumesFromTable(): Promise<number> {
+    return this.countRowsFromTable();
+  }
+
+  async countUsedVolumesFromTable(): Promise<number> {
+    const usedVolumes = await this.getRowsFromTableByStatus(VolumeState.Used);
+    return usedVolumes.length;
   }
 
   async waitForVolumeExists(name: string): Promise<boolean> {
@@ -75,5 +99,25 @@ export class VolumesPage extends MainPage {
   async waitForVolumeDelete(name: string): Promise<boolean> {
     await waitWhile(async () => await this.volumeExists(name));
     return true;
+  }
+
+  //returns true if the number of volumes with status "UNUSED" is 0 (prune is finished)
+  async waitForPruneVolumes(): Promise<boolean> {
+    let unusedVolumes;
+    await waitWhile(
+      async () => {
+        unusedVolumes = await this.getRowsFromTableByStatus(VolumeState.Unused);
+        return unusedVolumes.length !== 0;
+      },
+      { sendError: false },
+    );
+    return true;
+  }
+
+  async pruneVolumes(): Promise<VolumesPage> {
+    await playExpect(this.pruneVolumesButton).toBeEnabled();
+    await this.pruneVolumesButton.click();
+    await handleConfirmationDialog(this.page, 'Prune');
+    return this;
   }
 }
--- a/tests/playwright/src/specs/volume-smoke.spec.ts
+++ b/tests/playwright/src/specs/volume-smoke.spec.ts
@@ -87,11 +87,10 @@ describe('Volume workflow verification', async () => {
   test('Delete volume from Volumes page', async () => {
     let volumesPage = await navBar.openVolumes();
     await playExpect(volumesPage.heading).toBeVisible();
-    let volumeRow = await volumesPage.getVolumeRowByName(volumeName);
+    const volumeRow = await volumesPage.getVolumeRowByName(volumeName);
     playExpect(volumeRow).not.toBeUndefined();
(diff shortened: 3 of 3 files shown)


tests/playwright/src/model/pages/main-page.ts
```diff
@@ -94,4 +94,20 @@ export abstract class MainPage extends BasePage {
...
+    let numberOfRows = 0;
+    const table = await this.getTable();
+    const rows = await table.getByRole('row').all();
+    for (let i = rows.length - 1; i >= 0; i--) {
+      const nameCell = await rows[i].getByRole('cell').nth(3).count();
+      if (nameCell) {
+        numberOfRows++;
+      }
+    }
+    return numberOfRows;
```
cbr7: I'm not sure I understand, doesn't the all() method get all the current rows? shouldn't the value of that represent the count? Why do you need to go through each row afterwards?

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -58,6 +59,20 @@ export class VolumesPage extends MainPage {
...
   }
 
+  async deleteVolume(volumeName: string): Promise<VolumesPage> {
+    const volumeRow = await this.getVolumeRowByName(volumeName);
+    if (volumeRow === undefined) {
+      throw Error(`Volume: ${volumeName} does not exist`);
+    }
+    const containerRowDeleteButton = volumeRow.getByRole('button', { name: 'Delete Volume' });
+    await playExpect(containerRowDeleteButton).toBeVisible();
+    await playExpect(containerRowDeleteButton).toBeEnabled();
```
cbr7: There isn't a point in using toBeVisible() before toBeEnabled(), since toBeEnabled() checks for visibility implicitly.

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -76,4 +91,10 @@ export class VolumesPage extends MainPage {
     await waitWhile(async () => await this.volumeExists(name));
     return true;
   }
+
+  async pruneVolumes(): Promise<VolumesPage> {
+    await this.pruneVolumesButton.click();
```
cbr7, line 105: validate that the button is enabled before clicking it.

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -62,16 +68,84 @@ describe('Volume workflow verification', async () => {
...
+    containers = await navigationBar.openContainers();
+    const containersPage = await containers.deleteContainer(containerToRun);
+    await playExpect(containersPage.heading).toBeVisible();
+    await playExpect
+      .poll(async () => await containersPage.containerExists(containerToRun), { timeout: 15000 })
+      .toBeFalsy();
+
+    //prune unused volumes
+    volumesPage = await navigationBar.openVolumes();
+    await volumesPage.pruneVolumes();
```
cbr7: Shouldnt there be some validation here to check that the prune worked?

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -62,16 +68,84 @@ describe('Volume workflow verification', async () => {
...
+    //pull image from quay.io/centos-bootc/bootc-image-builder
+    const navigationBar = new NavigationBar(page);
+    let images = await navigationBar.openImages();
+    const pullImagePage = await images.openPullImage();
+    images = await pullImagePage.pullImage(imageToPull, imageTag);
+    await playExpect.poll(async () => await images.waitForImageExists(imageToPull)).toBeTruthy();
+
+    //count the number of existing volumes
+    let volumesPage = await navigationBar.openVolumes();
+    const previousVolumes = await volumesPage.countRowsFromTable();
```
cbr7: If you do this step before pulling the image you save up on navigating to and from the Image page to count the volumes.

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -62,16 +68,84 @@ describe('Volume workflow verification', async () => {
...
+    //start a container from the image (generates 4 new volumes)
+    images = await navigationBar.openImages();
+    const imageDetails = await images.openImageDetails(imageToPull);
+    const runImage = await imageDetails.openRunImage();
+    let containers = await runImage.startContainer(containerToRun, containerStartParams);
+    await playExpect(containers.header).toBeVisible();
+    await playExpect
+      .poll(async () => await containers.containerExists(containerToRun), { timeout: 10000 })
+      .toBeTruthy();
+    await containers.startContainer(containerToRun);
```
cbr7, line 151: Shouldnt you validate that he container started running before switching to volumes page?

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -62,16 +68,84 @@ describe('Volume workflow verification', async () => {
...
+
+    //check that four volumes are created (in addition to the existing ones)
+    volumesPage = await navigationBar.openVolumes();
+    await playExpect(volumesPage.heading).toBeVisible();
+    const newVolumes = await volumesPage.countRowsFromTable();
+    playExpect(newVolumes - previousVolumes === 4).toBeTruthy();
```
cbr7:
```suggestion
    playExpect(newVolumes - previousVolumes).toBe(4);
```

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -62,16 +68,84 @@ describe('Volume workflow verification', async () => {
...
+
+    //delete the container
+    containers = await navigationBar.openContainers();
+    const containersPage = await containers.deleteContainer(containerToRun);
```
cbr7, line 167: You should stop the container first, before deleting it.

cbr7 (changes requested):
Left some comments

tests/playwright/src/specs/volume-smoke.spec.ts
danivilla9: I'm not sure how the check that the prune worked. I can compare to the previous number of volumes, but it's possible that some of the previous volumes were inactive and got pruned so it's not 100% accurate. Should I prune the volumes before starting the test?

cbr7: That's certainly one possibility, another would be to check that all the rows that are still there are for volumes that are still in use.

Choose whichever version you think is best.

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -94,4 +94,15 @@ export abstract class MainPage extends BasePage {
...
   }
+
+  async countRowsFromTable(): Promise<number> {
+    try {
+      await this.page.waitForSelector('[role="row"]', { state: 'visible', timeout: 1000 });
+      const table = await this.getTable();
+      const rows = await table.getByRole('row').all();
+      return rows.length - 1; //counts the header of the table as a row, that's why one is substracted
+    } catch (error) {
+      return 0; //there are 0 visible rows
```
cbr7: I don't really understand this implementation, why do you need to wait for Selector? Especially since this requires a ugly solution in the try catch. Doesn't the all() method return 0 if there is nothing there? I assume in this case it should return 1 since there is the header, so why can't you just return the count of all()?

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -76,4 +94,26 @@ export class VolumesPage extends MainPage {
...
+    try {
+      await waitWhile(async () => {
+        currentVolumes = await this.countVolumesFromTable();
+        return previousVolumes === currentVolumes;
+      });
+    } catch (error) {
+      return true;
+    }
+    return true;
+  }
```
cbr7: I think I know what you want to do with this, but I don't see how this implementation works.

1. You are calling this method after the prune command was already issues, so by the time it executes some/all volumes might already be pruned, so there is an inate race condition involved.
2. You are awaiting while previousVolume === currentVolume, but what happens if you have many volumes and when you issue the prune command the first ones will disaapear(being deleted) but not yet all, it's possible that previousVolumes and currentVolumes might no longer be equal but that the prune command isn't finished either.
3. I also don't understand why you need to use try catch for waitWhile, the waitWhie method accepts param to not throw error at all.
4. What if no volumes can be pruned since they are currently in use?
5. Method always return `true` or am I not seeing the case where it responds with `false`?

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -55,23 +62,100 @@ describe('Volume workflow verification', async () => {
...
 
+  test('Delete volume from Volumes page', async () => {
+    let volumesPage = await navBar.openVolumes();
+    await playExpect(volumesPage.heading).toBeVisible();
+    let volumeRow = await volumesPage.getVolumeRowByName(volumeName);
+    playExpect(volumeRow).not.toBeUndefined();
+    volumesPage = await volumesPage.deleteVolume(volumeName);
+    await playExpect.poll(async () => await volumesPage.waitForVolumeDelete(volumeName)).toBeTruthy();
+    volumeRow = await volumesPage.getVolumeRowByName(volumeName);
+    playExpect(volumeRow).toBeUndefined();
```
cbr7: I'm not sure I understand why this is needed, aren't you already awaiting for the volume to be deleted on line 76? Why are you checking for it again on 78?

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -55,23 +62,100 @@ describe('Volume workflow verification', async () => {
...
+    await playExpect(volumesPage.heading).toBeVisible();
+    let volumeRow = await volumesPage.getVolumeRowByName(volumeName);
     playExpect(volumeRow).not.toBeUndefined();
 
     const volumeDetails = await volumesPage.openVolumeDetails(volumeName);
     volumesPage = await volumeDetails.deleteVolume();
 
     await playExpect.poll(async () => await volumesPage.waitForVolumeDelete(volumeName)).toBeTruthy();
+    volumeRow = await volumesPage.getVolumeRowByName(volumeName);
+    playExpect(volumeRow).toBeUndefined();
```
cbr7: Same as above.

tests/playwright/src/model/pages/main-page.ts
danivilla9: I tried a similar implementation to getRowFromTableByName(), using only all(), but I kept getting the wrong number of rows because the page takes a second or so to load the table that contains the volumes (specially if the volumes are created manually, not sure why), so if you navigate to the volumes page or prune the volumes and don't give it a little bit of time, you get inconsistent results. There has to be a better way to do it but I'm not sure how.
[not waiting for row.webm](https://github.com/user-attachments/assets/0a6225db-2c66-40a1-9a69-04c3719fcc15)
[waiting for row.webm](https://github.com/user-attachments/assets/0b52808b-588b-40c6-8014-84b676e44b04)

cbr7: Create a boolen method that returns the value of `await this.page.getByRole('row').first().isVisible()` then use that method as a callback in the waitUntil() method while setting the `sendError` param of waitUntil to `false`, this way you can wait for the header row to be visible without throwing an exception and needing the try catch to handle it.

tests/playwright/src/model/pages/volumes-page.ts
danivilla9: I think I came up with a solution to some of the problems you pointed out, but my guess is that I'm going to have to change the approach I take to this.
```
async waitForPruneVolumes(previousVolumes: number): Promise<boolean> {
    let currentVolumes;
    await waitWhile(async () => {
      currentVolumes = await this.countVolumesFromTable();
      return previousVolumes === currentVolumes;
    }, {sendError: false});
    return true;
  }
```
1. If I pass the previous number of volumes as a parameter (it is calculated before starting the prune always) the race condition should be avoided, as the method only counts the new existing volumes.
2. From my testing, after pruning the volumes the user interface does not refresh until it is done removing them, so I did not consider this possibility. Is this something common? I've only tried with ~20 or so custom volumes.

4 and 5. As long as a volume exists, the prune volumes button is enabled. In my opinion, this should not be the case, but rather it should be enabled only when there is at least one volume that can be deleted. But given this, and my approach to "prune first just in case", we can't know whether one or more volumes are going to get deleted or not after pruning. That's why the method always returns true, either it finishes pruning and the number of volumes changes (which ends the loop) or finishes pruning without deleting anything, which after certain time we consider it has finished pruning and we stop waiting for it. 

The more I think about this the less I like it, so I'm going to check the status of the rows one by one, as it does not make sense to try to prune volumes if we know it's not going to do anything, and this way there is no awkward "wait for nothing to happen" in some cases.

tests/playwright/src/specs/volume-smoke.spec.ts
danivilla9: The previous test (`'Delete volume from Volumes page'`) does that same check, so I figured out that on the new method for the same operation it would be adequate to check it as well.

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -94,4 +99,33 @@ export abstract class MainPage extends BasePage {
     }
     return undefined;
   }
+
+  async getRowsFromTableByStatus(status: string): Promise<Array<Locator>> {
+    await waitUntil(async () => await this.rowsAreVisible(), { sendError: false });
+
+    const table = this.content.getByRole('table');
+    if (table) {
```
cbr7: If condition does not have a purpose here, `table` will always be defined since it's a locator, I think you might be confusing the Locator object with the underlining web element, which might or might not exist.

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -94,4 +99,33 @@ export abstract class MainPage extends BasePage {
...
+      const rows = await table.getByRole('row').all();
+      rows.shift(); //first row is the headers row
```
cbr7: You should check that rows.length is greater than 1 before shifting, since you need a minimum of length 2 in order to know there is an actual volume there, since first row is the header.

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -94,4 +99,33 @@ export abstract class MainPage extends BasePage {
...
+      const filteredRows = [];
+      for (const row of rows) {
```
cbr7: Are you sure that for here does not iterate over the header row also? It seems to me that a classic for with a standard iterator would be better, since you can initialize the iterator from 1, instead of 0.

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -94,4 +99,33 @@ export abstract class MainPage extends BasePage {
...
+      return filteredRows;
+    }
+    return [];
+  }
+
+  async countRowsFromTable(): Promise<number> {
+    await waitUntil(async () => await this.rowsAreVisible(), { sendError: false });
+
+    const table = this.content.getByRole('table');
+    if (table) {
```
cbr7: Same as above, if will always be true here.

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -94,4 +99,33 @@ export abstract class MainPage extends BasePage {
...
+      const rows = await table.getByRole('row').all();
+      rows.shift(); //first row is the headers row
+      return rows.length;
```
cbr7: Don't really understand why all this is needed, why not check that length if greater than 1 and return length - 1?

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -76,4 +100,24 @@ export class VolumesPage extends MainPage {
...
+  //returns true if the number of volumes with status "UNUSED" is 0 (prune is finished)
+  async waitForPruneVolumes(): Promise<boolean> {
+    let unusedVolumes;
+    await waitWhile(
+      async () => {
+        unusedVolumes = await this.getRowsFromTableByStatus(VolumeState.Unused);
+        return unusedVolumes.length !== 0;
+      },
+      { sendError: false },
+    );
```
cbr7: What happens if the timeout on the waitWhile expires before unusedVolume count is zero? Since you are not throwing an error, since you are setting sendError to false that means the method would return true even though it unused volume count would not be zero.

Given that you created the getRowsFromTableByStatus method, woudln't it be simpler to just call that inside the test after the prune command is called and assert on it, for example something like `await playExpect(volumesPage.getRowsFromTableByStatus(VolumeState.Unused).toBe(0)` ?

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -94,4 +99,32 @@ export abstract class MainPage extends BasePage {
...
+      return filteredRows;
+    }
+    return [];
+  }
+
+  async countRowsFromTable(): Promise<number> {
+    await waitUntil(async () => await this.rowsAreVisible(), { sendError: false });
+
+    const table = this.content.getByRole('table');
+    if (await table.isVisible()) {
```
cbr7: I'm not sure what this if statement's purpose is, if you remove it and the final `return 0` also the code seems to be functionally the same.

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -67,6 +82,15 @@ export class VolumesPage extends MainPage {
     return result !== undefined;
   }
 
+  async countVolumesFromTable(): Promise<number> {
+    return this.countRowsFromTable();
+  }
+
+  async countUsedVolumesFromTable(): Promise<number> {
+    const usedVolumes = await this.getRowsFromTableByStatus(VolumeState.Used);
+    return usedVolumes.length;
```
cbr7:
```suggestion
    return await this.getRowsFromTableByStatus(VolumeState.Used).length;
```

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -77,16 +83,102 @@ describe('Volume workflow verification', async () => {
...
   });
+
+  test('Delete volume from Volumes page', async () => {
+    let volumesPage = await navBar.openVolumes();
+    await playExpect(volumesPage.heading).toBeVisible();
+    let volumeRow = await volumesPage.getVolumeRowByName(volumeName);
+    playExpect(volumeRow).not.toBeUndefined();
+    volumesPage = await volumesPage.deleteVolume(volumeName);
+    await playExpect.poll(async () => await volumesPage.waitForVolumeDelete(volumeName)).toBeTruthy();
+    volumeRow = await volumesPage.getVolumeRowByName(volumeName);
```
cbr7: I'm not sure I understand what this line does, why are you getting the volumeRow after you already deleted the volume, shouldnt it just be undefined here?

tests/playwright/src/specs/volume-smoke.spec.ts
```diff
@@ -77,16 +83,102 @@ describe('Volume workflow verification', async () => {
...
+    volumesPage = await navBar.openVolumes();
+    await playExpect(volumesPage.heading).toBeVisible();
+    let volumeRow = await volumesPage.getVolumeRowByName(volumeName);
     playExpect(volumeRow).not.toBeUndefined();
 
     const volumeDetails = await volumesPage.openVolumeDetails(volumeName);
     volumesPage = await volumeDetails.deleteVolume();
 
     await playExpect.poll(async () => await volumesPage.waitForVolumeDelete(volumeName)).toBeTruthy();
+    volumeRow = await volumesPage.getVolumeRowByName(volumeName);
```
cbr7: Same here, shouldnt this just be undefined? Since you already deleted the volume.

cbr7 (changes requested):
Just some minor things to clarify, but it looks good for the most part.

tests/playwright/src/specs/volume-smoke.spec.ts
danivilla9: I think before there was a statement after those lines that checked that the volume did not exist. I deleted them a while ago as you suggested, but I seemed to forget about deleting these ones as well.

cbr7 (approved):

PR #8144, opened 2024-07-18
fix(Tasks): task lifecycle managed in main

### LINKED ISSUE #7396: Merge all tasks management in main package

Currently, tasks can be created or updated both at the API/main level and within the renderer (see tasks store). This means that things in the main package (e.g. the status bar) can't tell if there are tasks / state doesn't match up with what's shown in the UI. e.g. if the status bar changed when there was a running task, it would only work for certain tasks.

All tasks should be created/managed from main, and renderer only handles the UI. Any function related to creating/managing tasks or updating status should either be removed or delegate to API in main.

_No response_

### Additional context

_No response_

axel7083:
### 🔎  What does this PR do?

Makes all the Tasks being managed in the main.

### 📑  Notable change

**new ipcs**

- `tasks:clear-all`  -> `window.clearTasks(): void`
- `tasks:clear`  -> `window.clearTask(taskId: string): void`
- `tasks:execute` -> `window.executeTask(taskId: string): void`

**Interfaces change**

Creating `TaskInfo` and `NotificationTaskInfo` in the `api` package.

> :information_source: Those objects are **readonly**, the renderer **cannot** modify any fields.

Creating `Task` and `NotificationTask` in the `extension-api`

> :information_source:  Those objects are **writable**, and will be automatically notify the frontend if a field are edited. The logic to achieve this behaviour has been made through `TaskImpl` and `NotificationTaskImpl` classes.

> ⚠️ For now the Tasks and NotificationTask object cannot be obtain by the extensions, as this will be done in https://github.com/containers/podman-desktop/issues/7709

**TaskManager change**

- No longer need to call `taskManager.updateTask(task)` as the Task intercept all setter inside `TaskImpl` class, it is automatically managed and propagated

> ℹ️ This avoid several potential problem where the task are not updated properly.

### 🔨 Future work

This PR remove all the tasks created on the renderer side **WITHOUT** replacing them by equivalent on the `main`. This choice is to ease the testing of this PR.

The list of tasks to created has been listed here https://github.com/containers/podman-desktop/issues/7396#issuecomment-2236420289

### 🖼️  Screenshot / video of UI

https://github.com/user-attachments/assets/14b06e90-4a5b-493a-9d71-159ab55ca829

### 🐛  What issues does this PR fix or reference?

Fixes https://github.com/containers/podman-desktop/issues/7396
Prior work for https://github.com/containers/podman-desktop/issues/7709

Requires
- https://github.com/containers/podman-desktop/pull/8292

### How to test this PR?

<details>
  <summary><h4>🪲 Debug commands palette</h4></summary>
  
You can add a series of debug commands to the palette to generate fake tasks.

Press `F1` to open Command Palette

```ts
// packages/main/src/plugin/index.ts
    commandRegistry.registerCommand('debug-task-generate-info-notification', () => {
      taskManager.createNotificationTask({
        type: 'info',
        body: 'Hello world',
      });
    });

    commandRegistry.registerCommand('debug-task-generate-error-task', () => {
      const task = taskManager.createTask({
        title: 'Dummy error task'
      });
      task.error = 'Something went wrong on purpose :)';
      task.state = 'error';
    });

    commandRegistry.registerCommand('debug-task-generate-indeterminate-task', () => {
      taskManager.createTask({
        title: 'Dummy indeterminate task'
      });
    });

    commandRegistry.registerCommand('debug-task-generate-50-loading-task', () => {
      const task = taskManager.createTask({
        title: 'Dummy loading 50 task'
      });
      task.progress = 50;
    });

    commandRegistry.registerCommand('debug-task-generate-success-task', () => {
      const task = taskManager.createTask({
        title: 'Dummy indeterminate task'
      });
      task.state = 'success';
    });

    commandRegistry.registerCommand('debug-task-generate-navigation-task', () => {
      const task = taskManager.createTask({
        title: 'Dummy navigable task'
      });
      task.state = 'success';
      task.action = {
        name: 'Go to containers >',
        execute: (): void => {
          navigationManager.navigateToContainers();
        },
      };
    });

    commandRegistry.registerCommand('debug-task-generate-actionable-task', () => {
      const task = taskManager.createTask({
        title: 'Dummy actionable task'
      });
      task.action = {
        name: 'explode',
        execute: (): void => {
          messageBox
            .showMessageBox({
              type: 'info',
              title: 'Task action',
              message: 'Hello from task action in the main package!',
              buttons: ['Set task error', 'Clear task'],
            })
            .then(result => {
              console.log('showMessageBox', result);
              switch (result.response) {
                case 0:
                  task.state = 'error';
                  break;
                case 1:
                  task.dispose();
                  break;
              }
            })
            .catch((err: unknown) => {
              console.error(err);
            });
        },
      };
    });

    commandRegistry.registerCommandPalette(
      {
        command: 'debug-task-generate-info-notification',
        title: '[debug-task] generate-info-notification',
        icon: undefined,
      },
      {
        command: 'debug-task-generate-error-task',
        title: '[debug-task] generate-error-task',
        icon: undefined,
      },
      {
        command: 'debug-task-generate-indeterminate-task',
        title: '[debug-task] generate-indeterminate-task',
        icon: undefined,
      },
      {
        command: 'debug-task-generate-50-loading-task',
        title: '[debug-task] generate-50-loading-task',
        icon: undefined,
      },
      {
        command: 'debug-task-generate-success-task',
        title: '[debug-task] generate-success-task',
        icon: undefined,
      },
      {
        command: 'debug-task-generate-navigation-task',
        title: '[debug-task] generate-navigation-task',
        icon: undefined,
      },
      {
        command: 'debug-task-generate-actionable-task',
        title: '[debug-task] generate-actionable-task',
        icon: undefined,
      },
    );
```

</details>

- [x] Unit tests has been provided


packages/extension-api/src/extension-api.d.ts
```diff
@@ -1429,6 +1429,15 @@ declare module '@podman-desktop/api' {
 
   type NotificationType = 'info' | 'warn' | 'error';
 
+  export interface NotificationTask extends Task {
```
feloy: Why do we want to expose these interfaces (this one and also the other added ones) to the extensions API? If I place myself as a user of the extension, I really don't understand what to do with these interfaces, as they are referenced in no function of the API

axel7083: Because they will be needed for https://github.com/containers/podman-desktop/issues/7709

When using `showNotification` you will get a NotificationTask instead of just a Disposable 

https://github.com/containers/podman-desktop/blob/86c7ee8b7dd09225a0b59dcfcf535021664bc48f/packages/extension-api/src/extension-api.d.ts#L1962

We will also expose a new `createTask` method similar to `showNotification` to let the extension do more stuff with the notifications/tasks

benoitf: I don't see yet why Tasks should leak into the extension API

it also still need to return a Disposable for showNotification (you can't change as it'll break API)

benoitf (dismissed):
I see some changes that can't be done

- Tasks leaking into API (we have notifications here, not tasks)
- Tasks creation that have purely removed from renderer part (it says "to ease the PR") but it can't be merged using that approach.

packages/extension-api/src/extension-api.d.ts
axel7083: NotificationTask extend Task which extend Disposable, it would not break the API

axel7083:
> * Tasks leaking into API (we have notifications here, not tasks)

I can place it in the `packages/api` while waiting for more discussion on https://github.com/containers/podman-desktop/issues/7709

> * Tasks creation that have purely removed from renderer part (it says "to ease the PR") but it can't be merged using that approach.

@benoitf my proposal was to make a follow-up PR which would replace those by creating them in the main. I will include them inside this PR; the list are available here https://github.com/containers/podman-desktop/issues/7396#issuecomment-2236420289

Here is a video demo with the latest commit

https://github.com/user-attachments/assets/aa827a6a-6f30-4949-adf5-a02c391570d4

packages/extension-api/src/extension-api.d.ts
axel7083: I moved the Task and NotificationTask interface to the main.

packages/main/src/plugin/index.ts
```diff
@@ -1128,57 +1141,138 @@ export class PluginSystem {
...
+              abortController,
+              buildargs,
+            },
+          )
+          .then(result => {
+            task.state = 'success';
+            return result;
+          })
+          .catch((err: unknown) => {
+            task.error = `Something went wrong while trying to build image: ${String(err)}`;
```
lstocchi, line 1208: Should we also set `task.state = 'error'` ? If true, this comments is related to all similar modifications in this file

packages/main/src/plugin/tasks/task-manager.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  protected isNotificationTask(task: Task): task is NotificationTask {
+    return 'title' in task;
+  }
+
+  public getTask(taskId: string): Task {
+    const task = this.tasks.get(taskId);
+    if (!task) {
+      console.log(Array.from(this.tasks.entries()));
```
lstocchi: to be removed

lstocchi (commented):
Is it just me or the progressing tasks are not displayed? E.g if i stop a podman machine i don't see anything in the task manager

packages/main/src/plugin/index.ts
axel7083: No need, if you set the property `error` it automatically set the sate to `error`

https://github.com/containers/podman-desktop/blob/3863310317f38f420a1981a8572ab1ada884e70f/packages/main/src/plugin/tasks/task-impl.ts#L69-L75

packages/main/src/plugin/tasks/task-manager.ts
axel7083: oh yes, nice catch :+1:

axel7083:
> Is it just me or the progressing tasks are not displayed? E.g if i stop a podman machine i don't see anything in the task manager

You were right, some tasks were missing. I updated the list of frontend tasks in https://github.com/containers/podman-desktop/issues/7396#issuecomment-2236420289

I think this PR is too big. I will put into draft, and see how to split it.

axel7083:
rebased after https://github.com/containers/podman-desktop/pull/8292

lstocchi (commented):
Works great now. The only thing that does not convince me is the name of the category 'completed'. We also lists notifications there and it is a bit weird to see 'podman needs to be set up' under completed. Seems like it was already done. Maybe updating the name of the section? Or it was like this even before? 🤔 

image

axel7083:
> Maybe updating the name of the section? Or it was like this even before? 🤔

Before it was `notification`, I could replace completed by notification, I could revert, but this could be in a later PR where we discuss it in a UX call.

axel7083:
rebased :heavy_check_mark:

packages/api/src/taskInfo.ts
```diff
@@ -16,27 +16,23 @@
...
+export type TaskState = 'loading' | 'success' | 'error';
 
-export interface Task {
+export interface NotificationTaskInfo extends TaskInfo {
+  body: string;
+  markdownActions?: string;
+  // overwrite some fields type
+  state: 'success';
+  progress: undefined;
+  error: undefined;
```
benoitf: here I don't see why we set undefined, AFAIK it's optional in the parent interface so it's already undefined

benoitf (commented):
this PR is really huge,

maybe reduce some scope, for example some stuff is also being moved to a new subfolder `tasks`
so maybe, move to tasks folder the existing objects prior to this PR

packages/api/src/taskInfo.ts
axel7083: We want to limit the value possible, as NotificationTaskInfo extend TaskInfo, not having this could let someone set an error or a progress value, by enforcing undefined we prevent a value to be set

benoitf: ok

benoitf: in this case should it be using `Omit` ? to really don't see these 2 fields

axel7083: Oh yeah, good idea

benoitf (commented):
reviewing the PR, I'm wondering why we don't have anymore a state and a status but only a state ?

axel7083:
> reviewing the PR, I'm wondering why we don't have anymore a state and a status but only a state ?

I removed it as from my understanding it was adding complexity and was kinda overlapping itself

TaskState = 'running' | 'completed';
TaskStatus = 'in-progress' | 'success' | 'failure';

You can deduce the state by looking at the status

_state = if status is success or failure then completed else running_

benoitf:
the issue I see is that we will probably enhance status with `canceled` or `created` status probably but then you'll need to update all the conditions and it won't work nicely.

meaning that you can have a task planned but not being running, being executed or being completed.

I would avoid to change this during a refactoring that was is meant to only move stuff from renderer side to the main side

axel7083:
> the issue I see is that we will probably enhance status with `canceled` or `created` status probably but then you'll need to update all the conditions and it won't work nicely.
> 
> meaning that you can have a task planned but not being running, being executed or being completed.
> 
> I would avoid to change this during a refactoring that was is meant to only move stuff from renderer side to the main side

Following the feedback I restored the `TaskState` and `TaskStatus`.

Setting the status will set the state;

- status = failure => state = completed
- status = success => state = completed
- status = in-progress => state = running

axel7083:
Rebased after https://github.com/containers/podman-desktop/pull/8534

benoitf:
it looks like we have some invalid code (unit test failing)

overall LGTM, lot of work 👍

axel7083:
> it looks like we have some invalid code (unit test failing)

Yeah, some change introduced in other PRs appear after rebase. Fixed now :heavy_check_mark:

benoitf (approved):
no changes in the external/extension-api so 👍

lstocchi (approved):
LGTM

PR #8146, opened 2024-07-18
chore: fix multiple connection warnings

### LINKED ISSUE #7582: A lot of warnings on load for PreferencesConnectionRendering..
### Bug description

Screenshot 2024-06-11 at 10 04 19 AM

Unsure exactly what's happening, this is happening as well on main.

When loading PD there is a lot of warnings regarding auditing the PreferencesConnectionRendering files.

### Operating system

mac M1

### Installation Method

None

### Version

1.10.2

### Steps to reproduce

`yarn watch` on main branch and open up PD / see console logs.

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
Removes the console warning `The provider does not support audit for connection parameters` that pops up multiple times

Closes https://github.com/containers/podman-desktop/issues/7582

### How to test this PR?

Go to one of the extensions `create new` page in the preferences page and check that the warning mentioned above doesn't pop up

### DIFF
--- a/packages/main/src/plugin/provider-registry.ts
+++ b/packages/main/src/plugin/provider-registry.ts
@@ -873,9 +873,10 @@ export class ProviderRegistry {
   ): Promise<AuditResult> {
     // grab the correct provider
     const provider = this.getMatchingProvider(internalProviderId);
+    const emptyAuditResult: AuditResult = { records: [] };
 
     if (!provider.connectionAuditor) {
-      throw new Error('The provider does not support audit for connection parameters');
+      return emptyAuditResult;
     }
 
     return provider.connectionAuditor.auditItems(params);


packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
```diff
@@ -184,11 +184,7 @@ onMount(async () => {
     }
   }
   if (!connectionInfo) {
-    try {
```
SoniaSandler: I wasn't sure if I should also remove the try catch block in line 215 since it is a bit different. Is there a special case there that I might be missing?

jeffmaury: No you should keep the try / catch as the optional audit method from a provider may throw an exception so we should be protected

jeffmaury (changes requested):

jeffmaury (approved):
LGTM

PR #8147, opened 2024-07-19
chore: support-multiline-messages-in-message-dialogs

### LINKED ISSUE #7685: Support multiline messages in message dialogs

Sometimes it is better to show message using several paragraphs. Below an example of what sign out confirmation request text could look like:

image

it could be multiple extension using the same authentication session. Where you see 'OpenShift Toolkit' in our case could be list of extensions:

OpenShif Local,
OpenShift Developer Sandbox

Current implementation ignores `\n` escape and shows everything in one line.

MessageBox component should split the message on `\n` and show every part as separate p or div element with padding.

_No response_

### Additional context

Related to #6632.

gastoner:
### What does this PR do?
Supports multiline messages in message dialogs

### Screenshot / video of UI
Before:
image

After:
image
image

Closes #7685 

### How to test this PR?

You have to be signed to Red Hat SSO extension
Try to sign-off

- Tests are covering the bug fix or the new feature


ekidneyrh:
There seems so be some extra space added between the last line and the buttons 

image

Would we be able to get rid of that?

feloy:
As the change is in the dialog, which is used in many places, we have to take care of negative side effects on other dialogs. 

For example, in light mode, I can see the text is now clearer:

before:
dialog-before

after:
dialog-after

ekidneyrh:
@feloy @gastoner I had a look at the colours on this and both light and dark mode is showing gray-700 currently when it should be gray-600 for dark mode and light mode charcoal-300. 

When I look at the main branch the colours are appearing as they are meant to be as they are already defined in the colour config I believe.

feloy:
pre tag does not work, as long lines are not cut (see screenshot)

pre-newlines

Probably one of `whitespace-` tailwind class (https://tailwindcss.com/docs/whitespace) would do what we want

feloy (approved):
LGTM

ekidneyrh (approved):

Issue #5966, opened 2024-02-12
Update the Close Details pages locators na the main pages and add a redirection tests


odockal:
### Is your enhancement related to a problem? Please describe

For example at the moment we have defined locators in the details page (Container's one) like this:
`this.closeLink = page.getByRole('link', { name: 'Close Details' });`
but the DOM of the page changed and looks like this, for breadcrumbs navigation:
```
<a class="text-purple-400 hover:bg-white hover:bg-opacity-10 transition-all rounded-[4px] p-0.5 no-underline text-sm" role="link" aria-label="back">Containers</a>
```
or like this for cross element:
```
<a href="/containers" title="Close" class="justify-self-end text-gray-900"><i class="fas fa-times" aria-hidden="true"></i></a>
```

It probably changed in the time, but there were no tests for navigation and so it did not fail anywhere.

### Describe the solution you'd like

Update the locators in the containers, images, etc. pages. Consider improving the ARIA labels and roles used for the close elements. Add a short tests into existing test files to make sure the navigation is not broken between details and the main pages.

PR #8151, opened 2024-07-19
test: add e2e test scenario for Kind extension

### LINKED ISSUE #8053: Deploy a pod to a Kubernetes (kind cluster)
Implement an End-to-End (E2E) test to verify the deployment of a pod/container in a Kubernetes cluster using the "Deploy to Kubernetes" option.
 
Steps to Implement the E2E Test:

- Creating of a Kind cluster (with ingress controller).
- Verification of the created cluster via podman desktop and basic operations -> restart, stop, start, delete.
- Check for resources added with kind cluster (images, containers, volumes, etc).
- Switching to newly created kube-context.
- Deploying a pod/container for a Kubernetes cluster via Deploy To Kubernetes option.
- Verification of the deployed app.

amisskii:

### What does this PR do?

This PR introduces an end-to-end test scenario for the Kind extension. 
The scenario involves: 
1. **Kind Cluster Creation**
   - Sets up a Kind cluster for testing purposes.

2. **Resource Validation**
   - Verifies that resources have been correctly added to the Kind cluster.

3. **Kubeconfig Context Validation**
   - Ensures that the kubeconfig context points to the newly created Kind cluster.

4. **Cluster Operations**

#5976 #8053 
https://github.com/containers/podman-desktop/pull/8152

### How to test this PR?

Run the tests with yarn test:e2e:run

### DIFF
--- a/tests/playwright/src/model/core/types.ts
+++ b/tests/playwright/src/model/core/types.ts
@@ -23,9 +23,9 @@ export interface ContainerInteractiveParams {
 
 export interface KindClusterOptions {
   clusterName?: string;
-  changeProvideType?: boolean;
+  providerType?: string;
   httpPort?: string;
   httpsPort?: string;
-  disableIngressController?: boolean;
+  useIngressController?: boolean;
   containerImage?: string;
 }
--- a/tests/playwright/src/model/core/types.ts
+++ b/tests/playwright/src/model/core/types.ts
@@ -20,3 +20,12 @@ export interface ContainerInteractiveParams {
   interactive?: boolean;
   attachTerminal?: boolean;
 }
+
+export interface KindClusterOptions {
+  clusterName?: string;
+  changeProvideType?: boolean;
+  httpPort?: string;
+  httpsPort?: string;
+  disableIngressController?: boolean;
+  containerImage?: string;
+}
--- a/tests/playwright/src/model/pages/create-kind-cluster-page.ts
+++ b/tests/playwright/src/model/pages/create-kind-cluster-page.ts
@@ -79,25 +79,28 @@ export class CreateKindClusterPage extends BasePage {
       } else {
         throw new Error(`${providerType} doesn't exist`);
       }
-
-      if (httpPort) {
-        await this.fillTextbox(this.httpPort, httpPort);
-      }
-      if (httpsPort) {
-        await this.fillTextbox(this.httpsPort, httpsPort);
-      }
-
-      if (!useIngressController) {
-        await playExpect(this.controllerCheckbox).toBeEnabled();
-        await this.controllerCheckbox.uncheck();
-        await playExpect(this.controllerCheckbox).not.toBeChecked();
-      }
-
-      if (containerImage) {
-        await this.fillTextbox(this.containerImage, containerImage);
-      }
-      await this.createCluster(timeout);
     }
+
+    if (httpPort) {
+      await this.fillTextbox(this.httpPort, httpPort);
+    }
+    if (httpsPort) {
+      await this.fillTextbox(this.httpsPort, httpsPort);
+    }
+
+    await playExpect(this.controllerCheckbox).toBeEnabled();
+    if (!useIngressController) {
+      await this.controllerCheckbox.uncheck();
+      await playExpect(this.controllerCheckbox).not.toBeChecked();
+    } else {
+      await this.controllerCheckbox.check();
+      await playExpect(this.controllerCheckbox).toBeChecked();
+    }
+
+    if (containerImage) {
+      await this.fillTextbox(this.containerImage, containerImage);
+    }
+    await this.createCluster(timeout);
   }
 
   private async createCluster(timeout: number = 120000): Promise<void> {
--- a/tests/playwright/src/model/pages/create-kind-cluster-page.ts
+++ b/tests/playwright/src/model/pages/create-kind-cluster-page.ts
@@ -18,7 +18,7 @@
 
 import { expect as playExpect, type Locator, type Page } from '@playwright/test';
 
-import { CheckboxOperations } from '../core/operations';
+import type { KindClusterOptions } from '../core/types';
 import { BasePage } from './base-page';
 
 export class CreateKindClusterPage extends BasePage {
@@ -29,6 +29,10 @@ export class CreateKindClusterPage extends BasePage {
   readonly goBackButton: Locator;
   readonly checkboxHolder: Locator;
   readonly logsButton: Locator;
+  readonly providerType: Locator;
+  readonly httpPort: Locator;
+  readonly httpsPort: Locator;
+  readonly containerImage: Locator;
 
   constructor(page: Page) {
     super(page);
@@ -40,42 +44,63 @@ export class CreateKindClusterPage extends BasePage {
     // Locator for the parent element of the ingress controller checkbox, used to change its value
     this.checkboxHolder = this.controllerCheckbox.locator('..');
     this.clusterCreationButton = this.clusterPropertiesInformation.getByRole('button', { name: 'Create', exact: true });
+    this.logsButton = this.clusterPropertiesInformation.getByRole('button', { name: 'Show Logs' });
+    this.providerType = this.clusterPropertiesInformation.getByRole('combobox', { name: 'Provider Type' });
+    this.httpPort = this.clusterPropertiesInformation.getByLabel('HTTP Port');
+    this.httpsPort = this.clusterPropertiesInformation.getByLabel('HTTPS Port');
+    this.containerImage = this.clusterPropertiesInformation.getByPlaceholder('Leave empty for using latest.');
     this.goBackButton = this.page.getByRole('button', { name: 'Go back to resources' });
-    this.logsButton = this.page.getByRole('button', { name: 'Show Logs' });
   }
 
-  public async createCluster(clusterName: string): Promise<void> {
-    await playExpect(this.clusterNameField).toBeVisible();
-    if ((await this.clusterNameField.textContent()) !== clusterName) {
-      await this.clusterNameField.fill('');
-      await this.clusterNameField.fill(clusterName);
+  public async createClusterDefault(clusterName: string): Promise<void> {
+    await this.fillTextbox(this.clusterNameField, clusterName);
+    await playExpect(this.providerType).toHaveValue('podman');
+    await playExpect(this.httpPort).toHaveValue('9090');
+    await playExpect(this.httpsPort).toHaveValue('9443');
+    await playExpect(this.checkboxHolder).toBeChecked();
+    await playExpect(this.containerImage).toBeEmpty();
+    await this.createCluster();
+  }
+
+  public async createClusterParametrized({
+    clusterName = 'kind-cluster',
+    changeProvideType = false,
+    httpPort = '9090',
+    httpsPort = '9443',
+    disableIngressController = false,
+    containerImage = '',
+  }: KindClusterOptions = {}): Promise<void> {
+    await this.fillTextbox(this.clusterNameField, clusterName);
+
+    if (changeProvideType) {
+      await playExpect(this.providerType).toBeVisible();
+      await this.providerType.selectOption({ value: 'docker' });
+      await playExpect(this.providerType).toHaveValue('docker');
     }
+
+    await this.fillTextbox(this.httpPort, httpPort);
+    await this.fillTextbox(this.httpsPort, httpsPort);
+
+    if (disableIngressController) {
+      await playExpect(this.checkboxHolder).toBeEnabled();
+      await this.checkboxHolder.uncheck();
+      await playExpect(this.checkboxHolder).not.toBeChecked();
+    }
+
+    await this.fillTextbox(this.containerImage, containerImage);
+    await this.createCluster();
+  }
+
+  private async createCluster(): Promise<void> {
     await playExpect(this.clusterCreationButton).toBeVisible();
     await this.clusterCreationButton.click();
     await playExpect(this.goBackButton).toBeVisible({ timeout: 120000 });
     await this.goBackButton.click();
   }
 
-  public async manageCheckboxState(operation: CheckboxOperations): Promise<void> {
-    await playExpect(this.controllerCheckbox).toBeVisible();
-
-    const isChecked = await this.checkboxHolder.isChecked();
-    if (
-      (operation === CheckboxOperations.Check && isChecked) ||
-      (operation === CheckboxOperations.Uncheck && !isChecked)
-    ) {
-      return;
-    }
-
-    switch (operation) {
-      case CheckboxOperations.Check:
-        await this.checkboxHolder.check();
-        await playExpect(this.controllerCheckbox).toBeChecked();
-        break;
-      case CheckboxOperations.Uncheck:
-        await this.checkboxHolder.uncheck();
-        await playExpect(this.controllerCheckbox).not.toBeChecked();
-        break;
-    }
+  private async fillTextbox(textbox: Locator, text: string): Promise<void> {
+    await playExpect(textbox).toBeVisible();
+    await textbox.fill(text);
+    await playExpect(textbox).toHaveValue(text);
   }
 }
--- a/tests/playwright/src/model/pages/create-kind-cluster-page.ts
+++ b/tests/playwright/src/model/pages/create-kind-cluster-page.ts
@@ -28,7 +28,7 @@ export class CreateKindClusterPage extends BasePage {
   readonly clusterCreationButton: Locator;
   readonly goBackButton: Locator;
   readonly logsButton: Locator;
-  readonly providerType: Locator;
+  readonly providerTypeCombobox: Locator;
   readonly httpPort: Locator;
   readonly httpsPort: Locator;
   readonly containerImage: Locator;
@@ -45,7 +45,7 @@ export class CreateKindClusterPage extends BasePage {
       .locator('..');
     this.clusterCreationButton = this.clusterPropertiesInformation.getByRole('button', { name: 'Create', exact: true });
     this.logsButton = this.clusterPropertiesInformation.getByRole('button', { name: 'Show Logs' });
-    this.providerType = this.clusterPropertiesInformation.getByRole('combobox', { name: 'Provider Type' });
+    this.providerTypeCombobox = this.clusterPropertiesInformation.getByRole('combobox', { name: 'Provider Type' });
     this.httpPort = this.clusterPropertiesInformation.getByLabel('HTTP Port');
     this.httpsPort = this.clusterPropertiesInformation.getByLabel('HTTPS Port');
     this.containerImage = this.clusterPropertiesInformation.getByPlaceholder('Leave empty for using latest.');
@@ -54,7 +54,7 @@ export class CreateKindClusterPage extends BasePage {
 
   public async createClusterDefault(clusterName: string): Promise<void> {
     await this.fillTextbox(this.clusterNameField, clusterName);
-    await playExpect(this.providerType).toHaveValue('podman');
+    await playExpect(this.providerTypeCombobox).toHaveValue('podman');
     await playExpect(this.httpPort).toHaveValue('9090');
     await playExpect(this.httpsPort).toHaveValue('9443');
     await playExpect(this.controllerCheckbox).toBeChecked();
@@ -63,32 +63,45 @@ export class CreateKindClusterPage extends BasePage {
   }
 
   public async createClusterParametrized({
-    clusterName = 'kind-cluster',
-    changeProvideType = false,
-    httpPort = '9090',
-    httpsPort = '9443',
-    disableIngressController = false,
-    containerImage = '',
+    clusterName,
+    providerType,
+    httpPort,
+    httpsPort,
+    useIngressController,
+    containerImage,
   }: KindClusterOptions = {}): Promise<void> {
-    await this.fillTextbox(this.clusterNameField, clusterName);
-
-    if (changeProvideType) {
-      await playExpect(this.providerType).toBeVisible();
-      await this.providerType.selectOption({ value: 'docker' });
-      await playExpect(this.providerType).toHaveValue('docker');
+    if (clusterName) {
+      await this.fillTextbox(this.clusterNameField, clusterName);
     }
 
-    await this.fillTextbox(this.httpPort, httpPort);
-    await this.fillTextbox(this.httpsPort, httpsPort);
+    if (providerType) {
+      await playExpect(this.providerTypeCombobox).toBeVisible();
+      const providerTypeOptions = await this.providerTypeCombobox.locator('option').allInnerTexts();
+      if (providerTypeOptions.includes(providerType)) {
+        await this.providerTypeCombobox.selectOption({ value: providerType });
+        await playExpect(this.providerTypeCombobox).toHaveValue(providerType);
+      } else {
+        throw new Error(`${providerType} doesn't exist`);
+      }
 
-    if (disableIngressController) {
-      await playExpect(this.controllerCheckbox).toBeEnabled();
-      await this.controllerCheckbox.uncheck();
-      await playExpect(this.controllerCheckbox).not.toBeChecked();
+      if (httpPort) {
+        await this.fillTextbox(this.httpPort, httpPort);
+      }
+      if (httpsPort) {
+        await this.fillTextbox(this.httpsPort, httpsPort);
+      }
+
+      if (!useIngressController) {
+        await playExpect(this.controllerCheckbox).toBeEnabled();
+        await this.controllerCheckbox.uncheck();
+        await playExpect(this.controllerCheckbox).not.toBeChecked();
+      }
+
+      if (containerImage) {
+        await this.fillTextbox(this.containerImage, containerImage);
+      }
+      await this.createCluster();
     }
-
-    await this.fillTextbox(this.containerImage, containerImage);
-    await this.createCluster();
   }
 
   private async createCluster(): Promise<void> {
--- a/tests/playwright/src/model/pages/create-kind-cluster-page.ts
+++ b/tests/playwright/src/model/pages/create-kind-cluster-page.ts
@@ -58,6 +58,15 @@ export class CreateKindClusterPage extends BasePage {
 
   public async manageCheckboxState(operation: CheckboxOperations): Promise<void> {
     await playExpect(this.controllerCheckbox).toBeVisible();
+
+    const isChecked = await this.checkboxHolder.isChecked();
+    if (
+      (operation === CheckboxOperations.Check && isChecked) ||
+      (operation === CheckboxOperations.Uncheck && !isChecked)
+    ) {
+      return;
+    }
+
     switch (operation) {
       case CheckboxOperations.Check:
         await this.checkboxHolder.check();
--- a/tests/playwright/src/model/pages/create-kind-cluster-page.ts
+++ b/tests/playwright/src/model/pages/create-kind-cluster-page.ts
@@ -1,0 +1,72 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { expect as playExpect, type Locator, type Page } from '@playwright/test';
+
+import { CheckboxOperations } from '../core/operations';
+import { BasePage } from './base-page';
+
+export class CreateKindClusterPage extends BasePage {
+  readonly clusterPropertiesInformation: Locator;
+  readonly clusterNameField: Locator;
+  readonly controllerCheckbox: Locator;
+  readonly clusterCreationButton: Locator;
+  readonly goBackButton: Locator;
+  readonly checkboxHolder: Locator;
+  readonly logsButton: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.clusterPropertiesInformation = this.page.getByRole('form', { name: 'Properties Information' });
+    this.clusterNameField = this.clusterPropertiesInformation.getByRole('textbox', { name: 'Name', exact: true });
+    this.controllerCheckbox = this.clusterPropertiesInformation.getByRole('checkbox', {
+      name: 'Setup an ingress controller',
+    });
+    // Locator for the parent element of the ingress controller checkbox, used to change its value
+    this.checkboxHolder = this.controllerCheckbox.locator('..');
+    this.clusterCreationButton = this.clusterPropertiesInformation.getByRole('button', { name: 'Create', exact: true });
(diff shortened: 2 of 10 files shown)


tests/playwright/src/model/pages/pods-details-page.ts
```diff
@@ -61,7 +61,7 @@ export class PodDetailsPage extends BasePage {
   }
 
   async getState(): Promise<string> {
-    const currentState = await this.page.getByRole('status').getAttribute('title');
+    const currentState = await this.page.getByRole('status').nth(0).getAttribute('title');
```
amisskii: Maybe we should change the HTML code of the pod details page by defining new attributes for the parent area where the pod status and buttons are located, instead of using the nth() function?

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -54,34 +62,102 @@ beforeEach<RunnerTestContext>(async ctx => {
...
+      await kindExtension.disableExtension();
+      await navigationBar.openSettings();
+      await playExpect(resourcesPage.kindResources).not.toBeVisible();
+      await navigationBar.openExtensions();
+      await kindExtension.enableExtension();
+      await navigationBar.openSettings();
+      await playExpect(resourcesPage.kindResources).toBeVisible();
+    });
   });
+  describe.skipIf(process.env.GITHUB_ACTIONS && process.env.RUNNER_OS === 'Linux')(
```
amisskii, line 102: Create_Kind_cluster_failure
There is an error with creating a Kind cluster using the Podman Desktop app on a Linux virtual machine.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -54,34 +62,102 @@ beforeEach<RunnerTestContext>(async ctx => {
...
+
+        const volumesPage = await navigationBar.openVolumes();
+        await playExpect.poll(async () => await volumesPage.kindClusterVolumeExists(), { timeout: 10000 }).toBeTruthy();
+      });
+      test('Validate the right Kubernetes context is chosen', async () => {
+        await statusBar.validateKubernetesContext(kubernetesKontex);
+      });
+      test('Create a Pod to be deployed to the Kind cluster', async () => {
+        const podsPage = await navigationBar.openPods();
+        await createPod(podsPage, podAppName);
```
amisskii: The https://github.com/containers/podman-desktop/pull/8152 must be merged first in order to use this function.

tests/playwright/src/model/pages/deploy-to-kubernetes-page.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { BasePage } from '../pages/base-page';
+
+export class DeployToKubernetesPage extends BasePage {
+  readonly deployButton: Locator;
+  readonly doneButton: Locator;
+  readonly podNameField: Locator;
+  constructor(page: Page) {
+    super(page);
+    this.podNameField = this.page.locator('#podName');
```
amisskii: We need to add attribute for this locator.

tests/playwright/src/model/pages/pods-details-page.ts
cbr7: I'm not sure I understand why you needed to add the nth(0) param, is there another cell in that row with the role of `status`? The methods that are calling this element seem to be working fine.

tests/playwright/src/model/pages/pods-page.ts
```diff
@@ -80,4 +85,11 @@ export class PodsPage extends MainPage {
     }
     return row.getByRole('button', { name: 'kebab menu', exact: true });
   }
+
+  public async openDeployToKubernetePage(name: string): Promise<DeployToKubernetesPage> {
+    const podActionMenu = await this.getPodActionsMenu(name);
+    await podActionMenu.click();
+    await this.deployToKubernetesButton.click();
```
cbr7: Should await for button to be enabled before clicking it.

tests/playwright/src/model/pages/resources-page.ts
```diff
@@ -34,5 +40,26 @@ export class ResourcesPage extends SettingsPage {
...
+
+  public async kindClusterExists(label = 'kind-cluster'): Promise<boolean> {
+    return (await this.kindProviderConnections.getByRole('region', { name: label }).count()) > 0;
+  }
+
+  public async operateWithKindCluster(operation: string): Promise<void> {
+    await playExpect(
+      this.kindProviderConnections.getByRole('button', { name: operation, exact: true }),
+    ).not.toBeDisabled({ timeout: 7000 });
+    await this.kindProviderConnections.getByRole('button', { name: operation, exact: true }).click();
```
cbr7: If you are going to implement such a general method the param for it should be a value from an enum, not just a plain string, otherwise users won't know what possible values there are.

tests/playwright/src/model/pages/pods-details-page.ts
amisskii: I encountered an error while testing on a Windows machine. There was another element with the role of status.

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -36,6 +36,12 @@ export class VolumesPage extends MainPage {
     this.collectUsageDataButton = this.additionalActions.getByRole('button', { name: 'Gather volume sizes' });
   }
 
+  async kindClusterVolumeExists(): Promise<boolean> {
+    const volumesTable = await this.getTable();
+    const kindClusterVolume = volumesTable.getByRole('row').first();
+    return (await kindClusterVolume.count()) > 0;
+  }
+
```
cbr7: Can you clarify on what this does? How is it different from just getting the first row in the volumes page?

tests/playwright/src/model/pages/pods-details-page.ts
cbr7: Which element? There shouldnt be another element I don't think, perhaps you've encountered a bug in the app.

cbr7 (changes requested):
The smoke test execution for this branch just exceeded 50 mins, I'm not sure on what is happened, but normal runtime is like 6-7 minutes.

https://github.com/containers/podman-desktop/actions/runs/10007792370/job/27663212571?pr=8151

tests/playwright/src/model/pages/volumes-page.ts
amisskii: All available methods require at least one parameter, which is the volume name. Since a volume created with the cluster has a different name each time, I need a method that can check for the existence of any volume. I also attempted to extract volume names from the container's inspect data but I didn't get it.

cbr7: ok, but how do you know the first volume is the volume for the kind cluster? What if the kind cluster volume is 2nd or 5th in the list?

tests/playwright/src/model/pages/pods-details-page.ts
amisskii: Screenshot from 2024-07-19 15-51-03

cbr7: Which button is this "Tasks"?

amisskii: I suppose that it is a bug, there isn't any 'Tasks' button on the test screenshots.

tests/playwright/src/model/pages/create-cluster-page.ts
```diff
@@ -0,0 +1,63 @@
...
+    this.clusterHttpPort = this.clusterPropertiesInformation.getByLabel('HTTP Port', { exact: true });
+    this.clusterHttpsPort = this.clusterPropertiesInformation.getByLabel('HTTPS Port', { exact: true });
+    this.clusterIngressControllerCheckbox = this.clusterPropertiesInformation.getByRole('checkbox');
+    this.clusterCreationButton = this.clusterPropertiesInformation.getByRole('button', { name: 'Create', exact: true });
+    this.goBackButton = this.page.getByRole('button', { name: 'Go back to resources' });
+    this.logsButton = this.page.getByRole('button', { name: 'Show Logs' });
+  }
+
+  public async checkClusterProperties(): Promise<void> {
+    await playExpect(this.clusterPropertiesInformation).toBeVisible();
```
odockal: @amisskii  I am not sure if we need this kind of check for every element. We gonna check for some once we ie. expect the page to appear after create New button was pressed, then once we gonna start filling some info into fields, etc. whyt @cbr7 ?

tests/playwright/src/model/pages/deploy-to-kubernetes-page.ts
odockal: I would say we need to add more than just one things here. Whole page could make use of some ARIA functionality revision.

tests/playwright/src/model/pages/pods-details-page.ts
odockal: Details page needs to get some ARIA functionality - #5966

tests/playwright/src/model/pages/resources-page.ts
```diff
@@ -26,6 +29,10 @@ export class ResourcesPage extends SettingsPage {
   readonly podmanResources: Locator;
   readonly composeResources: Locator;
   readonly kindResources: Locator;
+  readonly kindProviderSetup: Locator;
```
odockal: I would say this does not belong here. Every "Provider connection card" on the resources page is generated dynamically. We should also approach it this way, thus, let's create an abstract/general class to cover shared functionality and behavior of the every resource connection card, and then we can add specific provider card  implementations if that differs. Something similar as in Extensions.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -54,34 +68,107 @@ beforeEach<RunnerTestContext>(async ctx => {
...
+
+        const volumesPage = await navigationBar.openVolumes();
+        await playExpect
+          .poll(async () => await volumesPage.kindClusterVolumeExists(kindContainerName), { timeout: 10000 })
+          .toBeTruthy();
+      });
+      test('Validate the right Kubernetes context is chosen', async () => {
+        await statusBar.validateKubernetesContext(kubernetesKontex);
+      });
+      test('Create a Pod to be deployed to the Kind cluster', async () => {
```
odockal: I would stop here, this could be added in another PR, otherwise this PR will grow too much.

odockal (changes requested):

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,48 @@
...
+  readonly clusterIngressControllerCheckbox: Locator;
+  readonly clusterCreationButton: Locator;
+  readonly goBackButton: Locator;
+  readonly logsButton: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.clusterPropertiesInformation = this.page.getByRole('form', { name: 'Properties Information' });
+    this.clusterNameField = this.clusterPropertiesInformation.getByLabel('Name');
+    this.clusterIngressControllerCheckbox = this.clusterPropertiesInformation.getByRole('checkbox');
```
odockal: please add accessible name into getByrole. And, please add a locator for a parent element, that would allow you to actually change the checkbox value. Known issue. This is a workaround

tests/playwright/src/model/workbench/status-bar.ts
```diff
@@ -37,4 +39,10 @@ export class StatusBar extends BasePage {
     await handleConfirmationDialog(this.page, 'Kind');
     await handleConfirmationDialog(this.page, 'Kind', true, 'OK');
   }
+
+  public async validateKubernetesContext(context: string): Promise<void> {
+    await playExpect(this.kubernetesContextButton).toBeVisible();
+    const currentContex = await this.kubernetesContextButton.textContent();
```
odockal: it might be enough to just:
`await playExpect(kubernetesContext).toHaveText('context')`.

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -36,6 +36,30 @@ export class VolumesPage extends MainPage {
     this.collectUsageDataButton = this.additionalActions.getByRole('button', { name: 'Gather volume sizes' });
   }
 
+  public async getKindClusterVolumeName(containerName: string): Promise<string | undefined> {
```
odockal: please rename the function as it can be generally used.

tests/playwright/src/model/pages/resources-kind-page.ts
```diff
@@ -0,0 +1,55 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { expect as playExpect, type Locator, type Page } from '@playwright/test';
+
+import type { KindClusterOperations } from '../core/kind-cluster-operations';
+import { CreateKindClusterPage } from './create-kind-cluster-page';
+import { ResourcesPage } from './resources-page';
+
+export class ResourcesKindCard extends ResourcesPage {
```
odockal: you should not inherit from resourcesPage since you cannot really use the locator from there in here. Just use BasePage. example: `src/model/pages/extension-card-page`.

odockal: And make it please universal, since any other connection type can be used, add parameter for constructor.

odockal: Maybe, the best would be if you create a follow up issue, where you will refactor the Resource Card Pages. Ie.:
Abstract ResourceCardPage and two classes that inherits from it, ResourceConnectionCardPage (podman, kind, etc.) and CliResourceCardPage (kubectl, compose).

amisskii: https://github.com/containers/podman-desktop/pull/8310

package.json
```diff
@@ -50,7 +50,7 @@
     "test:unit": "npm run test:main && npm run test:preload && npm run test:preload-docker-extension && npm run test:preload-webview && npm run test:ui && npm run test:renderer && npm run test:scripts:stylesheet && npm run test:tools && npm run test:extensions",
     "test:e2e": "npm run test:e2e:build && npm run test:e2e:run",
     "test:e2e:build": "cross-env NODE_ENV=development MODE=development DEBUG=pw:browser npm run build",
-    "test:e2e:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/playwright/src/specs/ --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",
+    "test:e2e:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/playwright/src/specs/kind.spec.ts --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",
```
odockal: do not forget to remove the spec from filter.

amisskii:
https://github.com/containers/podman-desktop-e2e/actions/runs/10384270373

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,72 @@
...
+    await playExpect(this.controllerCheckbox).toBeVisible();
+    switch (operation) {
+      case CheckboxOperations.Check:
+        await this.checkboxHolder.check();
+        await playExpect(this.controllerCheckbox).toBeChecked();
+        break;
+      case CheckboxOperations.Uncheck:
+        await this.checkboxHolder.uncheck();
+        await playExpect(this.controllerCheckbox).not.toBeChecked();
+        break;
```
cbr7: You should check that the desired end state of the checkbox is not also the initial state of it, because otherwise checking and unchecking will not work as you want them. If you want to check the box you should validate that the box isnt iniitally checked before performing the check action, if it is checked initially you can just return out of the method as completed.

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -75,14 +75,18 @@ export abstract class MainPage extends BasePage {
     return await this.page.getByRole('row').first().isVisible();
   }
 
+  async getAllTableRows(): Promise<Array<Locator>> {
```
cbr7: Why is this a promise when it getting a list of Locators? It can be simplified to return Locator[]

tests/playwright/src/model/pages/main-page.ts
```diff
@@ -75,14 +75,18 @@ export abstract class MainPage extends BasePage {
...
+    const table = await this.getTable();
+    return table.getByRole('row').all();
```
cbr7:
```suggestion
    return await this.getTable().getByRole('row').all();
```

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -38,6 +38,30 @@ export class VolumesPage extends MainPage {
...
+            }
+          }
+          await volumeDetails.backLink.click();
+        }
+      }
+    } catch (err) {
+      console.log(`Exception caught on ${this.title} page with message: ${err}`);
+    }
+  }
+
```
cbr7: I'm honestly not sure what this method is doing, but it strikes me as a bad idea to do all those actions and navigations inside a for loop like that, can you please explain the flow of this method and why it should be like this? As far as I understand you are performing page navigation, multiple times in fact, inside a for loop.

tests/playwright/src/model/pages/main-page.ts
amisskii: The return type of an async method must be the global Promise<T> type

tests/playwright/src/model/pages/volumes-page.ts
amisskii: Of course! We need to find a volume that was added to the Kind cluster. However, the information indicating that the volume belongs to the Kind cluster is only in the summary tab of the volume details page. Therefore, I will go through all the volumes in the volumes table and check if the summary tab contains information about this volume being used by the Kind container.

cbr7: In that case that should be logic you implement inside the test, not inside a method, since this validation is test specific. Such logic is very brittle and should only be used inside the well known conditions of a test, it should not be part of exposed functionality from a POM.

tests/playwright/src/model/pages/main-page.ts
odockal: where did you see that?

tests/playwright/src/model/pages/volumes-page.ts
odockal: You are trying to find an association of the container's volume in a volume Page. And to map it, you need to open volume's details summary page, If I am not mistaken. Unless we got that info in Volume Page rows directly, only way.

tests/playwright/src/model/pages/create-kind-cluster-page.ts
amisskii: The first property of the check() method is: _Ensure that element is a checkbox or a radio input. If not, this method throws. If the element is already checked, this method returns immediately_. Therefore, I believe there is no need to assert whether the element is checked before calling this method.

amisskii: Same for uncheck()

cbr7: That is not what I said, I said that you should check the checkbox is not already checked/unchecked before check/unchecking yourself.

tests/playwright/src/model/pages/volumes-page.ts
amisskii: Yes, I’m trying to find it on the details summary page and then just return the volume name to assert if it is running (used). Regarding implementing this method directly in the test, I think it could be useful in other tests as well. So, for me, the POM object is a good place for that functionality.

cbr7: It really isn't because you can't control how the method will be used in the future and it will require heavy refactoring to be made non brittle, additionally the functionality is not even standard for POM objects, since the method is performing repeated navigations outside through different POM pages.

> I think it could be useful in other tests as well.

That is exactly what I am concerned about, the fact that people might be tempted to use this method in other tests without understanding what it involves and accounting for the different navigations that it performs.

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    switch (operation) {
+      case CheckboxOperations.Check:
+        await this.checkboxHolder.check();
+        await playExpect(this.controllerCheckbox).toBeChecked();
+        break;
+      case CheckboxOperations.Uncheck:
+        await this.checkboxHolder.uncheck();
+        await playExpect(this.controllerCheckbox).not.toBeChecked();
+        break;
```
cbr7, line 78:
```suggestion
    switch (operation) {
      case CheckboxOperations.Check:
        await playExpect(this.controllerCheckbox).not.toBeChecked();
        await this.checkboxHolder.check();
        await playExpect(this.controllerCheckbox).toBeChecked();
        break;
      case CheckboxOperations.Uncheck:
        await playExpect(this.controllerCheckbox).toBeChecked();
        await this.checkboxHolder.uncheck();
        await playExpect(this.controllerCheckbox).not.toBeChecked();
        break;
```

tests/playwright/src/model/pages/volumes-page.ts
odockal: Let's move it then outside of pom. into ie. utilities/operations.

tests/playwright/src/model/pages/create-kind-cluster-page.ts
odockal: I would divide the create kind cluster function into two. One the will create and verifu default values in the form, and custom one, which takes multiple parameters and set them into the form.

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,106 @@
...
+    httpPort = '9090',
+    httpsPort = '9443',
+    disableIngressController = false,
+    containerImage = '',
+  }: KindClusterOptions = {}): Promise<void> {
+    await this.fillTextbox(this.clusterNameField, clusterName);
+
+    if (changeProvideType) {
+      await playExpect(this.providerType).toBeVisible();
+      await this.providerType.selectOption({ value: 'docker' });
```
odockal: Given the nature of the element -> select/drop down. We always can have multiple option defined by either index or better string. So the string is the right type here. You can choose to add a check whether such option exists and throw if it does not, but select passed string value.

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,106 @@
...
+    await playExpect(this.providerType).toHaveValue('podman');
+    await playExpect(this.httpPort).toHaveValue('9090');
+    await playExpect(this.httpsPort).toHaveValue('9443');
+    await playExpect(this.controllerCheckbox).toBeChecked();
+    await playExpect(this.containerImage).toBeEmpty();
+    await this.createCluster();
+  }
+
+  public async createClusterParametrized({
+    clusterName = 'kind-cluster',
```
odockal: you could probably add an undefined as or type to each property, which would mean to go with what is default. Counts for every property here.

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,106 @@
...
+  }: KindClusterOptions = {}): Promise<void> {
+    await this.fillTextbox(this.clusterNameField, clusterName);
+
+    if (changeProvideType) {
+      await playExpect(this.providerType).toBeVisible();
+      await this.providerType.selectOption({ value: 'docker' });
+      await playExpect(this.providerType).toHaveValue('docker');
+    }
+
+    await this.fillTextbox(this.httpPort, httpPort);
```
odockal: It would change a little here:
```
if (options.httpPort) {
   await this.fillTextbox(this.httpPort, httpPort);
}
```

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,106 @@
...
+    await this.fillTextbox(this.httpsPort, httpsPort);
```
odockal: same as above

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,106 @@
...
+    await playExpect(this.containerImage).toBeEmpty();
+    await this.createCluster();
+  }
+
+  public async createClusterParametrized({
+    clusterName = 'kind-cluster',
+    changeProvideType = false,
+    httpPort = '9090',
+    httpsPort = '9443',
+    disableIngressController = false,
```
odockal: I would go for directly named property: `userIngressController`. Default is `true`.

tests/playwright/src/model/core/types.ts
```diff
@@ -20,3 +20,12 @@ export interface ContainerInteractiveParams {
   interactive?: boolean;
   attachTerminal?: boolean;
 }
+
+export interface KindClusterOptions {
+  clusterName?: string;
+  changeProvideType?: boolean;
```
odockal: consider `string | undefined` types.

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    await this.fillTextbox(this.containerImage, containerImage);
+    await this.createCluster();
+  }
+
+  private async createCluster(): Promise<void> {
+    await playExpect(this.clusterCreationButton).toBeVisible();
+    await this.clusterCreationButton.click();
+    await playExpect(this.goBackButton).toBeVisible({ timeout: 120000 });
```
odockal: timeout could be parametrized.

tests/playwright/src/model/pages/main-page.ts
odockal: why not: `Promise<Locator[]>` ?

tests/playwright/src/model/workbench/status-bar.ts
```diff
@@ -16,19 +16,21 @@
...
 export class StatusBar extends BasePage {
   readonly kindInstallationButton: Locator;
+  readonly kubernetesContext: Locator;
 
   constructor(page: Page) {
     super(page);
     this.kindInstallationButton = this.page.getByTitle(
       'Kind not found on your system, click to download and install it',
     );
+    this.kubernetesContext = this.page.getByTitle('Current Kubernetes contex');
```
odockal: There is a typo: `context`

tests/playwright/src/utility/operations.ts
```diff
@@ -215,3 +217,23 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
 export function checkForFailedTest(result: TaskResult, runner: PodmanDesktopRunner): void {
   if (result.errors && result.errors.length > 0) runner.setTestPassed(false);
 }
+
+export async function getVolumeNameForContainer(page: Page, containerName: string): Promise<string | undefined> {
+  const volumePage = new VolumesPage(page);
```
odockal: You must open the Volumes Page first. Look at other functions here.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -64,33 +70,89 @@ afterAll(async () => {
...
+    });
   });
+  describe.skipIf(process.env.GITHUB_ACTIONS && process.env.RUNNER_OS === 'Linux')(
+    'Kind cluster creation',
+    async () => {
+      test('Create a Kind cluster', async () => {
+        await navigationBar.openSettings();
+        await playExpect.poll(async () => resourcesPage.resourceCardIsVisible(RESOURCE_NAME)).toBeTruthy();
+        await playExpect(kindResourceCard.markdownContent).toBeVisible();
+        const createButton = kindResourceCard.createButton;
```
odockal: this assignment is not neccessary.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -64,33 +70,89 @@ afterAll(async () => {
...
+      });
+      test('Validate the right Kubernetes context is chosen', async () => {
+        await statusBar.validateKubernetesContext(KUBERNETES_CONTEXT);
+      });
+      test('Verify Kind cluster basic operations', async () => {
+        await navigationBar.openSettings();
+        await playExpect(kindResourceCard.resourceElementConnectionStatus).toHaveText(ResourceElementState.Running);
+        await kindResourceCard.performConnectionAction(ResourceElementActions.Stop);
+        await playExpect(kindResourceCard.resourceElementConnectionStatus).toHaveText(ResourceElementState.Off, {
+          timeout: 25000,
```
odockal: I think that this timeouts might not be enough on CI.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -64,33 +70,89 @@ afterAll(async () => {
...
+        });
+        await kindResourceCard.performConnectionAction(ResourceElementActions.Stop);
+        await playExpect(kindResourceCard.resourceElementConnectionStatus).toHaveText(ResourceElementState.Off, {
+          timeout: 25000,
+        });
+        await kindResourceCard.performConnectionAction(ResourceElementActions.Delete);
+        await playExpect(kindResourceCard.markdownContent).toBeVisible({ timeout: 25000 });
+      });
+    },
```
odockal, line 173: Also add delete test.

odockal (changes requested):

tests/playwright/src/model/pages/resource-connection-card-page.ts
```diff
@@ -39,7 +39,7 @@ export class ResourceConnectionCardPage extends ResourceCardPage {
 
   public async performConnectionAction(operation: ResourceElementActions): Promise<void> {
     const button = this.resourceElementConnectionActions.getByRole('button', { name: operation, exact: true });
-    await playExpect(button).toBeEnabled({ timeout: 10000 });
+    await playExpect(button).toBeEnabled({ timeout: 25000 });
```
ScrewTSW: Could also maybe be parametrized

ScrewTSW (approved):
LGTM, just one comment from me

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,115 @@
...
+    await this.clusterCreationButton.click();
+    await playExpect(this.goBackButton).toBeVisible({ timeout: timeout });
+    await this.goBackButton.click();
+  }
+
+  private async fillTextbox(textbox: Locator, text: string): Promise<void> {
+    await playExpect(textbox).toBeVisible();
+    await textbox.fill(text);
+    await playExpect(textbox).toHaveValue(text);
+  }
```
amisskii: Last week, I was looking for an issue that I could work on and noticed that we have at least two issues related to the Kubernetes cluster. Perhaps we should move the functions for creating a Kind cluster and checking if the Kind CLI is installed into the operations? Or would it be better to do this in a separate PR? WDYT @cbr7  @odockal?

cbr7: different issues need different PRs, we should always make PRs as small as possible for better granularity when it comes to changes

odockal: Yes, I agree, if there is a functionality that will be shared eventually, do it as part of another PR.

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      if (httpPort) {
+        await this.fillTextbox(this.httpPort, httpPort);
+      }
+      if (httpsPort) {
+        await this.fillTextbox(this.httpsPort, httpsPort);
+      }
+
+      if (!useIngressController) {
```
odockal: I would add else branch since if the default setup will change, passing the `true` value wont change anything. so please also add else part.

odockal (approved):

cbr7 (approved):

PR #8152, opened 2024-07-19
chore: extract pod creation functionality into separate function

amisskii:

### What does this PR do?

This PR extracts the pod creation functionality into a separate function. 
This refactoring aims to modularize the code, making the pod creation logic reusable across different scenarios.

This PR is related to the https://github.com/containers/podman-desktop/pull/8151 that addresses the implementation of end-to-end tests for the Kind extension.

### DIFF
--- a/tests/playwright/src/utility/operations.ts
+++ b/tests/playwright/src/utility/operations.ts
@@ -16,10 +16,13 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import path from 'node:path';
+
 import type { Page } from '@playwright/test';
 import { expect as playExpect } from '@playwright/test';
 import type { TaskResult } from 'vitest';
 
+import type { PodsPage } from '../model/pages/pods-page';
 import { RegistriesPage } from '../model/pages/registries-page';
 import { ResourcesPage } from '../model/pages/resources-page';
 import { ResourcesPodmanConnections } from '../model/pages/resources-podman-connections-page';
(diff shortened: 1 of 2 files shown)


tests/playwright/src/utility/operations.ts
```diff
@@ -144,6 +147,17 @@ export async function deletePod(page: Page, name: string): Promise<void> {
...
+  await playExpect(podsPage.heading).toBeVisible();
+
+  const playYamlPage = await podsPage.openPlayKubeYaml();
+  await playExpect(playYamlPage.heading).toBeVisible();
+
+  const yamlFilePath = path.resolve(__dirname, '..', '..', 'resources', `${podAppName}.yaml`);
+  podsPage = await playYamlPage.playYaml(yamlFilePath);
+  await playExpect(podsPage.heading).toBeVisible();
+}
+
```
cbr7, line 160: Playing yaml file does not necessarily mean that a pod is created, that is just the consequence of the test that this runs, but it does not make sense for the method to be named as such, since playing a yaml could just as easily not create a pod.

Also it's not a good idea to abstract the path to the yaml file this way at all, basically you are saying that only a yaml file in the resources folder can ever be used, what if we need to create a test with a yaml file in a different location?

Also I don't really understand why this method even needs to exist, the playYaml() method already exists, refactoring the entire test into one bigger method just creates needless abstractisation and makes it much harder to actually use.

If you really want to encapsulate the entire functionality, this should be a method exposed from the PodsPage POM classbut which receives the entire path to the yaml file, not just the name and also should be renamed.

cbr7 (changes requested):

package.json
```diff
@@ -50,7 +50,7 @@
     "test:unit": "npm run test:main && npm run test:preload && npm run test:preload-docker-extension && npm run test:preload-webview && npm run test:ui && npm run test:renderer && npm run test:scripts:stylesheet && npm run test:tools && npm run test:extensions",
     "test:e2e": "npm run test:e2e:build && npm run test:e2e:run",
     "test:e2e:build": "cross-env NODE_ENV=development MODE=development DEBUG=pw:browser npm run build",
-    "test:e2e:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/playwright/src/specs/ --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",
+    "test:e2e:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- vitest run tests/playwright/src/specs/yaml-smoke.spec.ts --pool=threads --poolOptions.threads.singleThread --poolOptions.threads.isolate --no-file-parallelism",
```
cbr7: This change needs to be reverted.

PR #8171, opened 2024-07-22
fix: remove set up podman notification if podman is set 

### LINKED ISSUE #7969: Podman Desktop on Windows: "Podman needs to be set up" notication
### Bug description

Fresh install (v1.11.1). This happens with earlier versions too.

Every boot, this notification alwas appears: "Podman needs to be set up"

image

Click on notification does nothing (it dissapear, but after some seconds apears again).

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

Fresh install

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

There have been some complains about the notification that asks the user to set up podman being visible even after podman has actually been set up https://github.com/containers/podman-desktop/issues/7979 and https://github.com/containers/podman-desktop/issues/7969. 

I was not able to find an exact replicator so this PR is just a proposal by looking at the code.
What changes:
1) The Podman extension disposes its own notification if podman is correctly set up
2) The podman core is updated so that when a notification is disposed, also its related task in the taskManager is removed.

This way, we should be safe that, in any moment, when podman is correctly set up, the user will not see the "set up" notification anywhere. 

### Screenshot / video of UI

N/A

it resolves #7969 and 7979

### How to test this PR?

1. have podman installed without any machine
2. verify the set up notification is there
3. create a new machine
4. you should see no notifications
5. remove the machine
6. you should see the notification again

### DIFF
--- a/packages/main/src/plugin/notification-registry.ts
+++ b/packages/main/src/plugin/notification-registry.ts
@@ -71,7 +71,7 @@ export class NotificationRegistry {
     // return disposable object
     return Disposable.create(() => {
       disposeShowNotification.dispose();
-      this.removeNotificationById(this.notificationId);
+      this.removeNotificationById(notification.id);
       this.taskManager.deleteTask(notificationTask);
     });
   }
--- a/packages/main/src/plugin/notification-registry.ts
+++ b/packages/main/src/plugin/notification-registry.ts
@@ -57,7 +57,7 @@ export class NotificationRegistry {
     // send event
     this.apiSender.send('notifications-updated');
     // create task
-    this.taskManager.createNotificationTask({
+    const notificationTask = this.taskManager.createNotificationTask({
       title: notification.title,
       body: notification.body,
       markdownActions: notification.markdownActions,
@@ -72,6 +72,7 @@ export class NotificationRegistry {
     return Disposable.create(() => {
       disposeShowNotification.dispose();
       this.removeNotificationById(this.notificationId);
+      this.taskManager.deleteTask(notificationTask);
     });
   }
 
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -140,6 +140,13 @@ export function isIncompatibleMachineOutput(output: string | undefined): boolean
   }
 }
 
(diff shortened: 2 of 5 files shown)


axel7083:
Fixes https://github.com/containers/podman-desktop/issues/8172 (I wanted to open the issue in the morning, and you provided a fix before I submit it!!)

packages/main/src/plugin/notification-registry.ts
```diff
@@ -72,6 +72,7 @@ export class NotificationRegistry {
     return Disposable.create(() => {
       disposeShowNotification.dispose();
       this.removeNotificationById(this.notificationId);
+      this.taskManager.deleteTask(notificationTask);
```
axel7083, line 75: <3

axel7083 (approved):
LGTM

jeffmaury (changes requested):
It did not fix the issue and I think it's because of the PATH. When you have not yet installed Podman on your machine, the PATH does not have the Podman folder. After Podman is setup, the notification is still there because the PATH has been updated on system settings but not for the current process. 

I was able to reproduce on Windows
- remove Podman
- remote Podman folder from PATH
- yarn watch
- setup podman from Podman Desktop
- after setup notification is still there

lstocchi:
> It did not fix the issue and I think it's because of the PATH. When you have not yet installed Podman on your machine, the PATH does not have the Podman folder. After Podman is setup, the notification is still there because the PATH has been updated on system settings but not for the current process.
> 
> I was able to reproduce on Windows
> 
> * remove Podman
> * remote Podman folder from PATH
> * yarn watch
> * setup podman from Podman Desktop
> * after setup notification is still there

Found the bug -> https://github.com/containers/podman-desktop/pull/8171/files#diff-3b11a34f78272c187d0fab5a7f206ff7cc49720665361f294ce707d7b51af715R74 . We were using the notificationId which is updated for every new notification so when installing podman for the first time we have
notification 1 - setup podman
notification 2 - podman in successfully installed
then when disposing, the notification 2 was the deleted one. This is why the setup notification was there

jeffmaury (approved):
LGTM

Issue #8160, opened 2024-07-19
Add ARM64 Support To publish-to-winget.yaml


marshallwp:
### Is your enhancement related to a problem? Please describe

The current [publish-to-winget.yaml](https://github.com/containers/podman-desktop/blob/d08425e19c463e6f84d4a84348417e0b0abc5432/.github/workflows/publish-to-winget.yaml#L65-L69) workflow only publishes the x64 installer to the [winget manifest](https://github.com/microsoft/winget-pkgs/blob/affcd3dda1ce028f4daae96ee53af37456146938/manifests/r/RedHat/Podman-Desktop/1.11.1/RedHat.Podman-Desktop.installer.yaml). This means that despite officially supporting Windows on ARM, users cannot install the ARM64 build via winget.

### Describe the solution you'd like

1. Update the winget manifest to list both installers (I'm working on this, see: https://github.com/microsoft/winget-pkgs/pull/163917).
2. Update publish-to-winget.yaml to include URL to both the x64 and ARM64 installers in space-delimited format. e.g., `wingetcreate update <PackageIdentifier> --urls "<InstallerUrl1>|x64" "<InstallerUrl2>|arm64"`

These must be done in this order as `wingetcreate update` requires the number of URLs in an update to match that in the existing manifest.

### Describe alternatives you've considered

I mean I could just update the manifest on winget...but then you wouldn't be able to publish future releases to winget using your existing pipeline.

benoitf:
@BinaryWizard904 will you work on 2. (update publish-to-winget action) ?

marshallwp:
@benoitf, yes, I'll go ahead and take a look at that.

marshallwp:
I've created a pull request that makes the needed changes the publish-to-winget.yaml file.  Additionally, the dependency, https://github.com/microsoft/winget-pkgs/pull/163917 was merged, so the number of URLs being updated will match the number in the manifest.

PR #8173, opened 2024-07-22
Add arm64 installer to publish-to-winget.yaml

marshallwp:
### What does this PR do?
Changes the winget publication workflow to create winget manifests that include _both_ x64 and arm64 builds.  The previous version only published manifests with the x64 build.

PowerShell command alias 'iwr' replaced with the full command name 'Invoke-WebRequest' as per [PowerShell best practices](https://learn.microsoft.com/en-us/powershell/scripting/learn/shell/using-aliases?view=powershell-7.4#dont-use-aliases-in-scripts).

#8160 

### How to test this PR?
Local testing is farily simple:
```powershell
# Install wingetcreate
Invoke-WebRequest https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
# Define testing values
$URL_X64 = 'https://github.com/containers/podman-desktop/releases/download/v1.11.1/podman-desktop-1.11.1-setup-x64.exe|x64'
$URL_ARM64 = 'https://github.com/containers/podman-desktop/releases/download/v1.11.1/podman-desktop-1.11.1-setup-arm64.exe|arm64'
$VERSION = '1.11.1'
# Execute wingetcreate and output result to file instead of publishing.
.\wingetcreate.exe update RedHat.Podman-Desktop -u $URL_X64 $URL_ARM64 -v $VERSION -o "$env:USERPROFILE\winget-manifest"
```

I'm less certain of how to test a github workflow specifically as I've never really worked on one before.

### DIFF
--- a/.github/workflows/publish-to-winget.yaml
+++ b/.github/workflows/publish-to-winget.yaml
@@ -61,9 +61,10 @@ jobs:
     steps:
       - name: Create winget PR
         run: |
-          iwr https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
-          .\wingetcreate.exe update RedHat.Podman-Desktop -u $Env:URL -v $Env:VERSION -t $Env:TOKEN --submit
+          Invoke-WebRequest https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
+          .\wingetcreate.exe update RedHat.Podman-Desktop -u $Env:URL_X64 $Env:URL_ARM64 -v $Env:VERSION -t $Env:TOKEN --submit
         env:
           TOKEN: ${{ secrets.PODMAN_DESKTOP_BOT_TOKEN }}
           VERSION: ${{ needs.version.outputs.desktopVersion }}
-          URL: ${{ format('https://github.com/containers/podman-desktop/releases/download/v{0}/podman-desktop-{0}-setup.exe|x64', needs.version.outputs.desktopVersion) }}
+          URL_X64: ${{ format('https://github.com/containers/podman-desktop/releases/download/v{0}/podman-desktop-{0}-setup.exe|x64', needs.version.outputs.desktopVersion) }}
+          URL_ARM64: ${{ format('https://github.com/containers/podman-desktop/releases/download/v{0}/podman-desktop-{0}-setup-arm64.exe|arm64', needs.version.outputs.desktopVersion) }}
--- a/.github/workflows/publish-to-winget.yaml
+++ b/.github/workflows/publish-to-winget.yaml
@@ -66,5 +66,5 @@ jobs:
         env:
           TOKEN: ${{ secrets.PODMAN_DESKTOP_BOT_TOKEN }}
           VERSION: ${{ needs.version.outputs.desktopVersion }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-winget.yaml
```diff
@@ -61,9 +61,10 @@ jobs:
...
         run: |
-          iwr https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
-          .\wingetcreate.exe update RedHat.Podman-Desktop -u $Env:URL -v $Env:VERSION -t $Env:TOKEN --submit
+          Invoke-WebRequest https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
+          .\wingetcreate.exe update RedHat.Podman-Desktop -u $Env:URL_X64 $Env:URL_ARM64 -v $Env:VERSION -t $Env:TOKEN --submit
         env:
           TOKEN: ${{ secrets.PODMAN_DESKTOP_BOT_TOKEN }}
           VERSION: ${{ needs.version.outputs.desktopVersion }}
-          URL: ${{ format('https://github.com/containers/podman-desktop/releases/download/v{0}/podman-desktop-{0}-setup.exe|x64', needs.version.outputs.desktopVersion) }}
+          URL_X64: ${{ format('https://github.com/containers/podman-desktop/releases/download/v{0}/podman-desktop-{0}-setup.exe|x64', needs.version.outputs.desktopVersion) }}
```
benoitf:
```suggestion
          URL_X64: ${{ format('https://github.com/containers/podman-desktop/releases/download/v{0}/podman-desktop-{0}-setup-x64.exe|x64', needs.version.outputs.desktopVersion) }}
```

should we use the x64 installer rather the one including both ?

.github/workflows/publish-to-winget.yaml
```diff
@@ -61,9 +61,10 @@ jobs:
     steps:
       - name: Create winget PR
         run: |
-          iwr https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
-          .\wingetcreate.exe update RedHat.Podman-Desktop -u $Env:URL -v $Env:VERSION -t $Env:TOKEN --submit
+          Invoke-WebRequest https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
+          .\wingetcreate.exe update RedHat.Podman-Desktop -u $Env:URL_X64 $Env:URL_ARM64 -v $Env:VERSION -t $Env:TOKEN --submit
```
jeffmaury, line 65: aren't we missing -u in from of the second URL ?

benoitf: AFAIK no, it's a list of URLs
https://github.com/microsoft/winget-create/blob/main/doc/update.md#how-does-winget-create-know-which-installers-to-match-when-executing-an-update

.github/workflows/publish-to-winget.yaml
marshallwp: I thought that was the x64-specific installer given it was already marked as x64 specific.  But if its not, changing it to one that is would be ideal. Winget will handle choosing which installer to use based on the computer's architecture.

.github/workflows/publish-to-winget.yaml
marshallwp: Correct.  I don't know why they implemented it like this, but all documentation I could find indicated it was a single, space-delimited list of URLs.  The full parameter name of '-u' is actually '--urls' (plural), so naming is consistent at least.

benoitf (approved):

Issue #8174, opened 2024-07-22
Refactor terminal svelte


cdrage:
### Is your enhancement related to a problem? Please describe

While implementing https://github.com/containers/podman-desktop/issues/7545 we are copying-and-pasting a lot of terminal information everywhere.

We should consolidate this into (possibly) a single importable svelte file for terminal.  

### Describe the solution you'd like

Refactor / technical debt for terminal styling / control / etc.

axel7083:
Duplicate of https://github.com/podman-desktop/podman-desktop/issues/9009 ?

PR #8194, opened 2024-07-23
feat: terminal light and dark mode selection

### LINKED ISSUE #7545: Light mode for terminals

We use terminals in many places, they should all support light mode.

Define any missing colors in the color registry and move all terminals to use them.

_No response_

### Additional context

_No response_

cdrage:
feat: terminal light and dark mode selection

### What does this PR do?

* Adds the ability to select your terminal theme from dark or light

### Screenshot / video of UI

https://github.com/user-attachments/assets/3a90d27f-1b86-4663-b430-1cea26e79ed9

Closes https://github.com/containers/podman-desktop/issues/7545

We will follow up with:
https://github.com/containers/podman-desktop/issues/8190 and
https://github.com/containers/podman-desktop/issues/8174 in the future

### How to test this PR?

1. Go to Terminal preferences and set your theme
2. Go to your Pod logs / Terminal, or anywhere else in PD that has a
   terminal
3. See the new theme.

**NOTE:** If you are going back to the "same" section and it's cached
(ex. terminal), you will have to refresh the screen.

### DIFF
--- a/packages/main/src/plugin/terminal-init.ts
+++ b/packages/main/src/plugin/terminal-init.ts
@@ -47,10 +47,9 @@ export class TerminalInit {
           maximum: 4,
         },
         [TerminalSettings.SectionName + '.' + TerminalSettings.Theme]: {
-          description:
-            'Theme to be used when displaying terminal operations. Requires refreshing terminal screen after configuration change.',
+          description: 'Theme for terminal windows. Requires refreshing terminal screen after configuration change.',
           type: 'string',
-          default: 'Dark',
+          default: 'dark',
           // Uses the list of themes in terminal-theme.ts
           enum: Object.keys(themes),
         },
--- a/packages/main/src/plugin/terminal-init.ts
+++ b/packages/main/src/plugin/terminal-init.ts
@@ -18,6 +18,7 @@
 
 import type { IConfigurationNode, IConfigurationRegistry } from './configuration-registry.js';
 import { TerminalSettings } from './terminal-settings.js';
+import { themes } from './terminal-theme.js';
 
 export class TerminalInit {
   private static DEFAULT_LINE_HEIGHT = 1;
@@ -44,6 +45,14 @@ export class TerminalInit {
           default: TerminalInit.DEFAULT_LINE_HEIGHT,
           minimum: 1,
           maximum: 4,
+        },
+        [TerminalSettings.SectionName + '.' + TerminalSettings.Theme]: {
+          description:
+            'Theme to be used when displaying terminal operations. Requires refreshing terminal screen after configuration change.',
+          type: 'string',
+          default: 'Dark',
+          // Uses the list of themes in terminal-theme.ts
+          enum: Object.keys(themes),
         },
       },
     };
--- a/packages/main/src/plugin/terminal-theme.ts
+++ b/packages/main/src/plugin/terminal-theme.ts
@@ -1,0 +1,207 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import type { ITheme } from 'xterm';
+
+/**
+ * Below is a curated list of terminal themes that are available for selection / use within Podman Desktop.
+ * The 'colors' are sourced from multiple repos and cross-checked for accuracy.
+ * Each theme will have references to the original repo.
+ *
+ * Notes:
+ * - Light themes will have a darker cursor colour as well as a darker selectionForeground colour to improve visibility
+ */
+
+// Common 'visibility' elements for light theme requirements
+const LightThemeVisibilityChanges = {
+  cursor: '#000000',
+  selectionBackground: '#f0f0f0',
+  selectionForeground: '#000000',
+};
+
+// From: https://github.com/ysk2014/xterm-theme/blob/master/src/iterm/Solarized_Dark.js
+const SolarizedDark: ITheme = {
+  foreground: '#708284',
+  background: '#001e27',
+  cursor: '#708284',
+
+  black: '#002831',
+  brightBlack: '#001e27',
+
+  red: '#d11c24',
+  brightRed: '#bd3613',
+
+  green: '#738a05',
+  brightGreen: '#475b62',
+
+  yellow: '#a57706',
+  brightYellow: '#536870',
+
+  blue: '#2176c7',
+  brightBlue: '#708284',
+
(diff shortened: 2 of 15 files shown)


cdrage:
Please note that this is the initial implementation and we will have to follow up with refactoring + another minor feature in the future: https://github.com/containers/podman-desktop/issues/8190 and https://github.com/containers/podman-desktop/issues/8174

packages/main/src/plugin/terminal-init.ts
```diff
@@ -45,6 +46,14 @@ export class TerminalInit {
           minimum: 1,
           maximum: 4,
         },
+        [TerminalSettings.SectionName + '.' + TerminalSettings.Theme]: {
+          description:
+            'Theme to be used when displaying terminal operations. Requires refreshing terminal screen after configuration change.',
```
deboer-tim: "Theme to be used when displaying terminal operations." just seems a bit verbose, maybe just
"Theme for terminal windows."?

packages/main/src/plugin/terminal-theme.ts
```diff
@@ -0,0 +1,207 @@
...
+
+// Same as DefaultDark, but with different background, foreground and cursor colors
+const DefaultLight: ITheme = {
+  ...DefaultDark,
+  ...LightThemeVisibilityChanges,
+  foreground: '#000000',
+  background: '#ffffff',
+};
+
+export const themes: { [key: string]: ITheme } = {
```
deboer-tim, line 72: We used lowercase for appearance but they should probably match. Some of these are names so likely works better capitalized, maybe we should change appearance?

packages/renderer/src/lib/compose/ComposeDetailsLogs.svelte
```diff
@@ -119,14 +119,15 @@ async function refreshTerminal() {
   const lineHeight = await window.getConfigurationValue<number>(
     TerminalSettings.SectionName + '.' + TerminalSettings.LineHeight,
   );
+  const terminalTheme = await window.getConfigurationValue<string>(
```
deboer-tim, line 122: Instead of having every file look up the configuration value and pass it to getTerminalTheme, why not have terminal-theme.ts look up the configuration value instead?

packages/main/src/plugin/terminal-theme.ts
```diff
@@ -0,0 +1,207 @@
...
+
+// Common 'visibility' elements for light theme requirements
+const LightThemeVisibilityChanges = {
+  cursor: '#000000',
+  selectionBackground: '#f0f0f0',
+  selectionForeground: '#000000',
+};
+
+// From: https://github.com/ysk2014/xterm-theme/blob/master/src/iterm/Solarized_Dark.js
+const SolarizedDark: ITheme = {
```
deboer-tim: Still on the fence about whether these should be defined and extendable like the appearance themes are (or whether we track that separately).

deboer-tim (commented):
I really like the change and choice of a few themes, it's really slick.

The fact that the change is only picked up by some windows and not others is really confusing to the user though, especially since it's one of the only preferences like that - it's easy to miss the description and wonder why it's not working correctly. Have you checked whether it's possible to force theme updates, or maybe we should just throw out our terminal caches when this setting changes?

benoitf (changes requested):
IMHO it looks confusing to have terminal theming option.

I would expect to select a global appearance theme, not having something specific for terminals. 

For example it means I could have a light terminal with a dark theme or the opposite.

To me, I would expect that the colors of the terminal would be provided by the definition of our global colors. And if you want to customize, you would install an extension providing the theme. Here it is limiting the extensibility by hardcoding available themes in Podman Desktop. While it should be contributed by 3rd party extensions

cdrage:
> IMHO it looks confusing to have terminal theming option.
> 
> 
> 
> I would expect to select a global appearance theme, not having something specific for terminals. 
> 
> 
> 
> For example it means I could have a light terminal with a dark theme or the opposite.
> 
> 
> 
> To me, I would expect that the colors of the terminal would be provided by the definition of our global colors. And if you want to customize, you would install an extension providing the theme. Here it is limiting the extensibility by hardcoding available themes in Podman Desktop. While it should be contributed by 3rd party extensions

How about just dark and light option and no extra themes? So we could at least implement it for light mode / have the option. 

The reason for selective preference is so you can have a dark terminal despite selecting light theme. As users like me may want to have PD in light mode but terminals dark for visibility.

packages/main/src/plugin/terminal-theme.ts
cdrage: Changed! to match the other sections, I've changed it to lowercase now.

packages/renderer/src/lib/compose/ComposeDetailsLogs.svelte
cdrage: I initially did this, but found it did not "sync" when updating. So I added this to match the other parts of the terminal.

I opened up #8174 for when we refactor Terminal

cdrage:
@benoitf @deboer-tim

I've updated the PR with a new commit that:

- Only adds dark and light. I agree with @benoitf after thinking about it. If we added some "common" themes like Zenburn, Monokai, it may introduce debates / discussions on what colours are the "right ones", etc. Since there is so many different combinations / how everyone views the colours differently. So may as well just stick to the default dark / light.
- Switches to lowercase light / dark in description

@deboer-tim regarding the "automatic" clear cache / updating terminal. I could not find an elegant solution without beginning to refactor every terminal page, so instead let's update it in #8174 ?

benoitf:
> How about just dark and light option and no extra themes? So we could at least implement it for light mode / have the option.

>The reason for selective preference is so you can have a dark terminal despite selecting light theme. As users like me may > want to have PD in light mode but terminals dark for visibility.

I don't think there should be a separated theming option just for Terminal.
A theme include all options including the terminal rendering.

so, if you select a light theme you'll get a light mode terminal and if you select a dark theme you'll get a dark mode terminal.

if you want to have different colors, then you'll provide a custom theme that would override colors of the terminal extending/overriding a given theme.

deboer-tim:
We had a brief chat on the standup today and can continue tomorrow on UX, but dropping my thoughts here in the meantime. I agree with Florent that the best way to deliver this for now (which might be easier as a clean PR/branch) is just  adding variables to the color registry, one set of light/dark options, and using those with no separate preference. It works, follows our extensibility pattern, gives a consistent theme with one setting, and will be an easy review/merge.

Personally, yes, I think it would be nice to allow more options for the terminal without requiring every theme to be extended with every terminal option, but that discussion is separate and going to take longer to conclude one way or the other, so we should track it separately.

I'm still hesitant to deliver without the ability to update terminal colors live and not knowing when #8174 might drop (might be nice to confirm we have a technical solution or start by fixing just one view like Logs), but yes, that is a separate issue too.

cdrage:
> We had a brief chat on the standup today and can continue tomorrow on UX, but dropping my thoughts here in the meantime. I agree with Florent that the best way to deliver this for now (which might be easier as a clean PR/branch) is just adding variables to the color registry, one set of light/dark options, and using those with no separate preference. It works, follows our extensibility pattern, gives a consistent theme with one setting, and will be an easy review/merge.
> 
> Personally, yes, I think it would be nice to allow more options for the terminal without requiring every theme to be extended with every terminal option, but that discussion is separate and going to take longer to conclude one way or the other, so we should track it separately.
> 
> I'm still hesitant to deliver without the ability to update terminal colors live and not knowing when #8174 might drop (might be nice to confirm we have a technical solution or start by fixing just one view like Logs), but yes, that is a separate issue too.

I agree and thank you for talking about it! I'll close this for now and tackle it in a different direction (more integrated, use color registry, use current pattern), as well as see if we can do a live update to the colours.

PR #8199, opened 2024-07-23
chore: revert afterUpdate in items

### LINKED ISSUE #8082: Experimental setting for light / dark mode appears everywhere?
### Bug description

Unsure if it's just me, but the setting appears on each settings screen?

https://github.com/user-attachments/assets/7d0493a4-f3ba-4261-aaa4-af3f0e24dd90

### Operating system

mac M1

### Installation Method

None

### Version

1.11.1

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

cdrage:
chore: revert afterUpdate in items lsit

### What does this PR do?

* Reverts back to using afterUpdate due to the settings no longer
  appearing for most sections

We should look for a different solution for https://github.com/containers/podman-desktop/issues/7142

### Screenshot / video of UI

Before:

image

After (shows the correct setting now):

Screenshot 2024-07-23 at 11 11 41 AM

Closes https://github.com/containers/podman-desktop/issues/8082

### How to test this PR?

Go to setting (example, kubernetes context) and see that the setting is
back now.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 import { getInitialValue } from '/@/lib/preferences/Util';
+import { afterUpdate } from 'svelte';
 
 import type { IConfigurationPropertyRecordedSchema } from '../../../../main/src/plugin/configuration-registry';
 import Markdown from '../markdown/Markdown.svelte';
@@ -14,6 +15,10 @@ let recordUI: {
   markdownDescription?: string;
   original: IConfigurationPropertyRecordedSchema;
 };
+
+afterUpdate(() => {
+  update();
+});
 
 // add space from camel case and upper case on the first letter
 function startCase(str: string): string {


cdrage:
Re-adding this, I do not experience the phantom issue however? Was there a change somewhere else in the code (was trying to use git blame, but couldn't find anything).

cc @lstocchi @deboer-tim

deboer-tim:
FWIW I haven't seen any other changes that would affect it - but it was sporadic before, sometimes I couldn't reproduce and other times it would happen every single time I change a setting.

lstocchi:
I didn't notice at the time this problem, btw by a quick test, just by updating the `update` function signature to accept a IConfigurationPropertyRecordedSchema and use the `record` variable  inside the reactive call seems to fix the problem 

```
function update(record: IConfigurationPropertyRecordedSchema) {
....
}

$: {
  update(record);
}
```

without re-adding the afterUpdate. Maybe i'm missing some other reason why it was there though

cdrage:
> I didn't notice at the time this problem, btw by a quick test, just by updating the `update` function signature to accept a IConfigurationPropertyRecordedSchema and use the `record` variable inside the reactive call seems to fix the problem
> 
> ```
> function update(record: IConfigurationPropertyRecordedSchema) {
> ....
> }
> 
> $: {
>   update(record);
> }
> ```
> 
> without re-adding the afterUpdate. Maybe i'm missing some other reason why it was there though

I'm going to close this PR so we can update some changes, is it possible you can open up a PR with the fix?

PR #8200, opened 2024-07-23
fix: fix tooltip error message overflow

### LINKED ISSUE #7057: Tooltip for pod deletion error is never fully visible due to alignment
### Bug description

When a 'Delete' command fails for a pod a red icon with exclamation point appears next to it to indicate error state. The tool-tip for the icon shows the actual error. Due to alignment of the tool-tip it is not always possible to see the error.

image

### Operating system

macOS M1 Sonoma 14.3

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. Deploy a pod to kubernetes cluster
2. Delete it outside of Podman Desktop using dashboard or CLI tool
3. Try to delete it from Pods page in Podman Desktop
4. Hover mouse pointer over error icon to in pod's row to see the error message

### Relevant log output

_No response_

### Additional context

_No response_

### LINKED ISSUE #8153: Failed container start error message cut off
### Bug description

Opening issue on behalf of user feedback report:
The error icon appearing when starting a container fails shows a message on hover over that extends beyond the right side of the window and is therefore not readable.

### Operating system

macOS 14.5

### Installation Method

None

### Version

1.11.1

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
Fixes the error tooltip overflow to make the entire error visible inside the page boundaries

### Screenshot / video of UI
Screenshot from 2024-07-22 21-33-56

Screenshot from 2024-07-22 21-33-45
Screenshot from 2024-07-22 21-33-34

Closes https://github.com/containers/podman-desktop/issues/8153
Closes https://github.com/containers/podman-desktop/issues/7057

### How to test this PR?

Check that the error messages in the container list and pod list pages are visible inside the page boundaries

### DIFF
--- a/packages/ui/src/lib/alert/ErrorMessage.svelte
+++ b/packages/ui/src/lib/alert/ErrorMessage.svelte
@@ -7,13 +7,13 @@ import Tooltip from '../tooltip/Tooltip.svelte';
 
 export let error: string;
 export let icon = false;
-export let actionErrorInfo = false;
+export let wrapMessage = false;
 let customClassWidth = '';
 let left = false;
 let top = false;
 
 onMount(() => {
-  if (actionErrorInfo) {
+  if (wrapMessage) {
     customClassWidth = 'w-max max-w-[650px] overflow-hidden text-ellipsis text-wrap';
     left = true;
   } else {
--- a/packages/ui/src/lib/alert/ErrorMessage.svelte
+++ b/packages/ui/src/lib/alert/ErrorMessage.svelte
@@ -6,13 +6,26 @@ import Tooltip from '../tooltip/Tooltip.svelte';
 
 export let error: string;
 export let icon = false;
+export let actionErrorInfo = false;
 </script>
 
 {#if icon}
   {#if error !== undefined && error !== ''}
-    <Tooltip top tip={error}>
-      <Fa size="1.1x" class="cursor-pointer text-[var(--pd-state-error)] {$$props.class}" icon={faExclamationCircle} />
-    </Tooltip>
+    {#if actionErrorInfo}
+      <Tooltip left tip={error} tipWidth="w-max max-w-[650px] overflow-hidden text-wrap">
+        <Fa
+          size="1.1x"
+          class="cursor-pointer text-[var(--pd-state-error)] {$$props.class}"
+          icon={faExclamationCircle} />
+      </Tooltip>
+    {:else}
+      <Tooltip top tip={error}>
+        <Fa
+          size="1.1x"
+          class="cursor-pointer text-[var(--pd-state-error)] {$$props.class}"
+          icon={faExclamationCircle} />
+      </Tooltip>
+    {/if}
   {/if}
 {:else}
   <div
--- a/packages/ui/src/lib/alert/ErrorMessage.svelte
+++ b/packages/ui/src/lib/alert/ErrorMessage.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
+import { onMount } from 'svelte';
 import Fa from 'svelte-fa';
 
 import Tooltip from '../tooltip/Tooltip.svelte';
@@ -7,25 +8,25 @@ import Tooltip from '../tooltip/Tooltip.svelte';
 export let error: string;
 export let icon = false;
 export let actionErrorInfo = false;
+let customClassWidth = '';
+let left = false;
+let top = false;
+
+onMount(() => {
+  if (actionErrorInfo) {
+    customClassWidth = 'w-max max-w-[650px] overflow-hidden text-ellipsis text-wrap';
+    left = true;
+  } else {
+    top = true;
+  }
+});
 </script>
 
 {#if icon}
   {#if error !== undefined && error !== ''}
-    {#if actionErrorInfo}
-      <Tooltip left tip={error} tipWidth="w-max max-w-[650px] overflow-hidden text-wrap">
-        <Fa
-          size="1.1x"
-          class="cursor-pointer text-[var(--pd-state-error)] {$$props.class}"
-          icon={faExclamationCircle} />
-      </Tooltip>
-    {:else}
-      <Tooltip top tip={error}>
-        <Fa
-          size="1.1x"
-          class="cursor-pointer text-[var(--pd-state-error)] {$$props.class}"
-          icon={faExclamationCircle} />
-      </Tooltip>
-    {/if}
+    <Tooltip left={left} top={top} tip={error} class={customClassWidth}>
+      <Fa size="1.1x" class="cursor-pointer text-[var(--pd-state-error)] {$$props.class}" icon={faExclamationCircle} />
+    </Tooltip>
   {/if}
 {:else}
   <div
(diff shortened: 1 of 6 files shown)


packages/ui/src/lib/alert/ErrorMessage.svelte
```diff
@@ -6,13 +6,26 @@ import Tooltip from '../tooltip/Tooltip.svelte';
...
+          class="cursor-pointer text-[var(--pd-state-error)] {$$props.class}"
+          icon={faExclamationCircle} />
+      </Tooltip>
+    {:else}
+      <Tooltip top tip={error}>
+        <Fa
+          size="1.1x"
+          class="cursor-pointer text-[var(--pd-state-error)] {$$props.class}"
+          icon={faExclamationCircle} />
+      </Tooltip>
```
benoitf: it looks like we're duplicating all the code just to add tipWidth attribute.
maybe add a conditional attribute


```
in onMount:

let tipWidth = undefined;
if (actionErrorInfo) {
  tipWidth="w-max...."
}
```

and then `<Tooltip ..... tipWidth={tipWidth} ...`

benoitf: and for left and top (it should be in the onMount as well like top=false, left=false, etc)

packages/ui/src/lib/tooltip/Tooltip.svelte
```diff
@@ -51,6 +51,7 @@
 
 <script lang="ts">
 export let tip: string | undefined = undefined;
+export let tipWidth = '';
```
benoitf: I would not call it tipWidth here because you're just passing extra class properties

so use `{$$props.class}` and in the parent component, just use class="your extra attributes"

packages/ui/src/lib/alert/ErrorMessage.svelte
```diff
@@ -1,16 +1,30 @@
...
 import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
+import { onMount } from 'svelte';
 import Fa from 'svelte-fa';
 
 import Tooltip from '../tooltip/Tooltip.svelte';
 
 export let error: string;
 export let icon = false;
+export let actionErrorInfo = false;
+let tipWidth: string | undefined = undefined;
```
SoniaSandler: @benoitf can `tipWidth` being `undefined` cause problems when we use it in line 27 here or line 80 in `Tooltip.svelte`? I tried playing with it a bit and it doesn't look like it's problematic, but I just wanted to make sure. Is it better to leave it as undefined or maybe change it to start as an empty string?

benoitf: I think that now as you're using recommended `class={}` you should use as default an empty string and rename tipWidth to like class or customClass or something to say you provide custom class elements

packages/ui/src/lib/alert/ErrorMessage.svelte
```diff
@@ -1,16 +1,30 @@
 <script lang="ts">
 import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
+import { onMount } from 'svelte';
 import Fa from 'svelte-fa';
 
 import Tooltip from '../tooltip/Tooltip.svelte';
 
 export let error: string;
 export let icon = false;
+export let actionErrorInfo = false;
```
benoitf: I'm wondering if it's a property name that can be understood by callers. The name makes me feel it could be a string to provide while it's expecting a boolean.

maybe should it contain something like what it really does  like `showFullMessage` 

also, should it be the default behavior ?

SoniaSandler: @benoitf `ErrorMessage` is used in many places, so I didn't want to change the original behavior. Those that I changed are the ones we know that have problematic placement with the error message. As we encounter other problematic places similar to the ones fixed here, we can fix them by using that boolean value, and if we notice that most places that use `ErrorMessage` need that fix, we can make it the default behavior. (I can also try and go through all `ErrorMessage` instances and check if this fix could be applied to their placements)
Regarding the property name, I can change it to `wrapErrorInfo`  or `wrapMessage` since the main change is that the text is wrapped across multiple lines instead of showing in one long line

packages/ui/src/lib/tooltip/Tooltip.svelte
```diff
@@ -77,7 +77,7 @@ export let left = false;
     class:bottom-right={bottomRight}>
     {#if tip}
       <div
-        class="inline-block py-2 px-4 rounded-md bg-[var(--pd-tooltip-bg)] text-[var(--pd-tooltip-text)] border-[1px] border-[var(--pd-tooltip-border)]"
+        class="inline-block py-2 px-4 rounded-md bg-[var(--pd-tooltip-bg)] text-[var(--pd-tooltip-text)] border-[1px] border-[var(--pd-tooltip-border)] {$$props.class}"
```
SoniaSandler, line 80: @benoitf since `Tooltip.svelte` is also used by components other than `ErrorMessage`, could `$${props.class}` be a problem if it is undefined in other components?

benoitf: no it's 👍

packages/ui/src/lib/alert/ErrorMessage.svelte
benoitf: `wrapMessage` looks more intuitive for a boolean property

SoniaSandler: leave it as a non-default behavior for now?

benoitf: yes

packages/ui/src/lib/alert/ErrorMessage.svelte
```diff
@@ -1,16 +1,30 @@
...
 export let icon = false;
+export let wrapMessage = false;
+let customClassWidth = '';
+let left = false;
+let top = false;
+
+onMount(() => {
+  if (wrapMessage) {
+    customClassWidth = 'w-max max-w-[650px] overflow-hidden text-ellipsis text-wrap';
+    left = true;
```
feloy, line 18: Placing the tooltip on the left is associated with the fact we want to wrap the message. I'm not sure it will scale, as we will, at some point, want to wrap the message for elements on the left of the window (at error messages can be very long, even when started on the left of the window)

SoniaSandler: @feloy I can change the boolean property name to be more specific, for example `wrapRightMessage` (and maybe in the future also add `wrapLeftMessage`), to indicate the page placement of the message we want to wrap and then change the tooltip placement based on that.

benoitf (approved):

packages/ui/src/lib/alert/ErrorMessage.svelte
SoniaSandler: @benoitf should I change the boolean property name to also indicate position on page so that we'll be able to add the same wrapping property for other positions in the future? right now it is not critical, but I do agree that it could be helpful for future changes/updates

feloy: let's keep as is it for now, we can change later when we find other problematic tooltips.

feloy (approved):

PR #8204, opened 2024-07-23
feat: apply multiple kube yamls

### LINKED ISSUE #8198: Ability to apply multiple Kubernetes YAML files

Today you can click the Apply button on any Kubernetes page, but can only select a single YAML file.

It would be nice to be able to select multiple files.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds the ability to apply multiple Kubernetes YAML files at once.

While doing this I couldn't help but improve the dialog message. Instead of:
  "Successfully applied 2 resource(s)"
you now either get:
  "Successfully applied 1 <kind>."
or a more detailed message like:
  "Successfully applied 4 resources (2 <kind1>, 1 <kind2>)."

### Screenshot / video of UI

https://github.com/user-attachments/assets/dcf205f7-5030-4794-991c-a38cd5527e76

Fixes #8198.

### How to test this PR?

Try a variety of files, resources.

### DIFF
--- a/packages/main/src/plugin/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes-client.spec.ts
@@ -1115,7 +1115,7 @@ test('Expect apply with empty yaml should throw error', async () => {
     expectedError = err;
   }
   expect(expectedError).to.be.a('Error');
-  expect((expectedError as Error).message).equal('No valid Kubernetes resources found in file');
+  expect((expectedError as Error).message).equal('No valid Kubernetes resources found');
 });
 
 test('Expect apply should create if object does not exist', async () => {
@@ -1164,6 +1164,23 @@ test('Expect apply should patch with specific field manager', async () => {
 
   await client.applyResourcesFromFile('default', 'some-file.yaml');
   expect(patchMock).toHaveBeenCalledWith(expect.any(Object), undefined, undefined, 'podman-desktop');
+});
+
+test('Expect apply should work with multiple files', async () => {
+  const client = createTestClient('default');
+  const manifests = { kind: test, metadata: { annotations: test } } as unknown as KubernetesObject;
+  const createdObjs = [{ kind: 'created' }, { kind: 'created' }];
+  let count = 0;
+  vi.spyOn(client, 'loadManifestsFromFile').mockResolvedValue([manifests]);
+  makeApiClientMock.mockReturnValue({
+    create: vi.fn().mockReturnValue({ body: createdObjs[count++] }),
+  });
+
+  const objects = await client.applyResourcesFromFile('default', ['some-file.yaml', 'another-file.yaml']);
+
+  expect(objects).toHaveLength(2);
+  expect(objects[0]).toEqual(createdObjs[0]);
+  expect(objects[1]).toEqual(createdObjs[1]);
 });
 
 test('If Kubernetes returns a http error, output the http body message error.', async () => {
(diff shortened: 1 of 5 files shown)


axel7083 (commented):
Work nicely when having multiple objects in a single yaml file.

image


However, the error handling is an issue for me.

image

Given the following yaml where the secret is invalid, I do not get an error.

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: dotfile-secret
datadd:
  .ssssecret-file: dmFsdWUtMg0KDQo=
---
apiVersion: v1
kind: Pod
metadata:
  name: secret-dotfiles-pod
spec:
  volumes:
    - name: secret-volume
      secret:
        secretName: dotfile-secret
  containers:
    - name: dotfile-test-container
      image: registry.k8s.io/busybox
      command:
        - ls
        - "-l"
        - "/etc/secret-volume"
      volumeMounts:
        - name: secret-volume
          readOnly: true
          mountPath: "/etc/secret-volume"
```

When I select multiple file however the behavior is not the same (let's say I select `invalid-secret.yaml` and `valid-volume.yaml`): 

image

Depending on the order, the volume may be applied, the secret not. I only get the feedback for the error, even tho the volume might be applied.

I am not sure of what could be done to improve the feedback here.

benoitf (commented):
I think it should be handle by the remote side rather than the client side 
(one call with 3files vs 3 calls with one file)

The changes on the resources being created should handle plural and singular

like Deployments with an s if there are multiple entries or with (s) like Deployment(s)

line 66  it'll display 1 resources rather than 1 resource or 1 resource(s)

deboer-tim:
> Depending on the order, the volume may be applied, the secret not. I only get the feedback for the error, even tho the volume might be applied.

I will look at what we can do to improve error handling & feedback, but partial applying is 'expected' even when a single file contains multiple resources - kube apply walks through in order, and if you get an error then everything before that is applied and everything after isn't.

> I think it should be handle by the remote side rather than the client side (one call with 3files vs 3 calls with one file)

Sure, I can move the core support to main.

> The changes on the resources being created should handle plural and singular
> 
> like Deployments with an s if there are multiple entries or with (s) like Deployment(s)

We have support/pages for several kinds of objects now, but the yaml really could contain anything, i.e. the extendable list of Kubernetes kinds like NetworkPolicy, GatewayClass, EndpointSlice, (Tekton) Task, ClusterRole, ServiceAccount ... Many of these are like NetworkPolicies or GatewayClasses where you can't just add an '(s)', so I was trying to avoid it. Would you expect a hard-coded list to pluralize just the ones we know about/are most likely, you'd rather not see the list of kinds at all, or can we defer this?

> line 66 it'll display 1 resources rather than 1 resource or 1 resource(s)

If there is only one it is handled on line 48 now, it says 'Successfully applied 1 <kind>'.

deboer-tim:
Rebased so it is current and added a commit to move multi-file support to main and not in renderer.

@axel7083 I have tested and confirmed that the behaviour applying multiple files is identical to single file with multiple resources - i.e. it isn't changed by this PR. I think we can improve the usability with multiple resources that include a failure, but should be a separate issue.

@benoitf Not sure if my previous explanation was sufficient but function was moved to main package.

axel7083 (approved):
LGTM

cdrage (approved):
Tested against a YAML with multiple resources, as well as selecting multiple YAML files.

Code and implementation LGTM.

lstocchi (approved):
LGTM

PR #8210, opened 2024-07-24
Display info when layers of an image are being loaded

### LINKED ISSUE #8158: display info when Image layers are long to be displayed

For large images, getting the content of the image layers can take a few seconds. Some information message or feedback on the screen would be necessary to help the user understand what is happening

- first: Display an information message explaining that the operation can take a few seconds
- improve API for the extension to be able to send information about the advancement of the operation

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

When the user opens the File tab of an image, the layers of the image are loaded, to be displayed. A message is displayed the time the layers are loaded and displayed.

### Screenshot / video of UI

https://github.com/user-attachments/assets/26a53632-555d-42b2-8f7a-c7acdc46d124

### How to test this PR?

Fix #8158 

### DIFF
--- a/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
+++ b/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
@@ -18,6 +18,7 @@ let filesProvidersUnsubscribe: Unsubscriber;
 let filesProvider: ImageFilesInfo;
 let selectedLayer: ImageFilesystemLayerUI;
 let showLayerOnly: boolean;
+let loading: boolean;
 
 function onSelectedLayer(event: CustomEvent<ImageFilesystemLayerUI>) {
   selectedLayer = event.detail;
@@ -27,9 +28,15 @@ onMount(async () => {
   filesProvidersUnsubscribe = imageFilesProviders.subscribe(providers => {
     if (providers.length === 1 && imageInfo) {
       filesProvider = providers[0];
-      window.imageGetFilesystemLayers(filesProvider.id, imageInfo).then(layers => {
-        imageLayers = layers;
-      });
+      loading = true;
+      window
+        .imageGetFilesystemLayers(filesProvider.id, imageInfo)
+        .then(layers => {
+          imageLayers = layers;
+        })
+        .finally(() => {
+          loading = false;
+        });
     }
   });
 });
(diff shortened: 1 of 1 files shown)


axel7083 (commented):
I really like this feature, but I am wondering since we could have very large image if the process should not be triggered by the user instead of automatically when opening the page ? what do you think @feloy ?

feloy:
> I really like this feature, but I am wondering since we could have very large image if the process should not be triggered by the user instead of automatically when opening the page ? what do you think @feloy ?

With some experience, the user will click on the File button only if he is ready to wait for displaying the files. But this can be disturbing for new comers, I agree. I would be tempted to add a preference for the user to choose to start the process automatically or not (with a non-automatic default value). WDYT @axel7083 ?

axel7083:
> > I really like this feature, but I am wondering since we could have very large image if the process should not be triggered by the user instead of automatically when opening the page ? what do you think @feloy ?
> 
> With some experience, the user will click on the File button only if he is ready to wait for displaying the files. But this can be disturbing for new comers, I agree. I would be tempted to add a preference for the user to choose to start the process automatically or not (with a non-automatic default value). WDYT @axel7083 ?

I like the preference idea!

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -27,9 +28,15 @@ onMount(async () => {
   filesProvidersUnsubscribe = imageFilesProviders.subscribe(providers => {
     if (providers.length === 1 && imageInfo) {
       filesProvider = providers[0];
-      window.imageGetFilesystemLayers(filesProvider.id, imageInfo).then(layers => {
-        imageLayers = layers;
-      });
+      loading = true;
+      window
+        .imageGetFilesystemLayers(filesProvider.id, imageInfo)
+        .then(layers => {
```
axel7083, line 34: should we handle potential errors ?

axel7083 (approved):
LGTM, I am wondering for the error handling, but as this is not the topic of the PR it is not blocking

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
feloy: Yes, this is also my wondering, I'll create a specific issue for this

benoitf (approved):

PR #8232, opened 2024-07-25
make saveImage cancellable

### LINKED ISSUE #8214: Make the api.containerEngine.saveImage cancellable

Calling saveImage can be a long-running operation, when the image to save is large. The user may want to cancel the operation, and it would be easier to finalise the operations when the user cancels the operation, if the saving of the file on disk is aborted correctly.

Add an optional cancellation token to the `saveImage` method:

```
export function saveImage(engineId: string, id: string, filename: string, token?: CancellationToken): Promise<void>;
```

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

make the call to `api.containerEngine.saveImage` cancellable

Fixes #8214 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -19,6 +19,7 @@
 import * as crypto from 'node:crypto';
 import { EventEmitter } from 'node:events';
 import * as fs from 'node:fs';
+import { rm } from 'node:fs/promises';
 import path from 'node:path';
 import { Writable } from 'node:stream';
 import { pipeline } from 'node:stream/promises';
@@ -2221,7 +2222,12 @@ export class ContainerProviderRegistry {
     }
   }
 
-  async saveImage(engineId: string, id: string, filename: string): Promise<void> {
+  async saveImage(
+    engineId: string,
+    id: string,
+    filename: string,
+    token?: containerDesktopAPI.CancellationToken,
+  ): Promise<void> {
     let telemetryOptions = {};
     try {
       // need to find the container engine of the container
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/container-registry.ts
```diff
@@ -2235,8 +2241,27 @@ export class ContainerProviderRegistry {
...
-        const imageStream = await imageObject.get();
-        return pipeline(imageStream, fs.createWriteStream(filename));
+        // make the download of image cancellable
+        const getImageObjectPromise = imageObject.get();
+        const cancelPromise = new Promise<NodeJS.ReadableStream>((_, reject) => {
+          token?.onCancellationRequested(() => {
+            reject(new Error('saveImage operation canceled'));
+          });
+        });
+        const imageStream = await Promise.race<NodeJS.ReadableStream>([getImageObjectPromise, cancelPromise]);
```
axel7083, line 2251: This does not cancel the `image.get` call, it give the illusion but does not cancel the request.

As visible in the dockerode implementation of the image get, they do not support cancelling this request.

https://github.com/apocas/dockerode/blob/49bd8c4c231efe9bb432b28d51dbae3e50ddd010/lib/image.js#L125-L135

You can compare to other method which support to have an `AbortSignal`

https://github.com/apocas/dockerode/blob/49bd8c4c231efe9bb432b28d51dbae3e50ddd010/lib/image.js#L160-L172

If this call is not cancellable, it might be because it is a very small operation, which execute almost instantly, therefore would not need this racing logic.

If the call is something which can be heavy, it would be better to open a pull request on their side to add the signal option support.

Then on https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/dockerode adding the type for the signal argument.

feloy: Yes, I agree that the call to `image.get` itself is not canceled. 

But my intention here is to at least 'detach' from this call, so the function can continue, while the call to `image.get` continues on its side (this call can be long, the time to download the image in memory, generally the longest part in my tests for large images). This call having side effects only in memory, I don't feel that there is any incidence (except some optimization).

The most important part for me is the second part, where the call to `pipeline` is effectively aborted, which is important because we want to stop the writing on disk before to delete the file.

benoitf (approved):
ok for me

axel7083 (approved):

Issue #8234, opened 2024-07-25
Light mode for provider info


deboer-tim:
### Is your enhancement related to a problem? Please describe

When looking at #6925 I noticed that ProviderInfo (provider status on dashboard and Settings > Resources) still has hard-coded colors.

### Describe the solution you'd like

Use color variables.

PR #8236, opened 2024-07-25
chore: improved status colors

### LINKED ISSUE #6925: Light mode for Dashboard

The Dashboard doesn't support light mode.

Define any missing colors in the light mode palette and get the main components of the Dashboard working in light mode.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Issue #6925 was reopened to tweak the provider status and card carousel colors. The first thing I found was that providers weren't using the correct colors (variables or values) to begin with, so that is covered in issue #8234/PR #8235.

Since we are now using consistent/one set of color variables, changing these status colors for providers implies changing these *everywhere*, including the status icons for containers, pods, deployments, etc. That said, I thought the previous colors were a little bright and I prefer these corrected ones in both light and dark mode.

### Screenshot / video of UI

N/A, minor change.

Fixes #6925.

### How to test this PR?

Look at the guide carousel in the Dashboard, and status icons on every page, in both light and dark mode.

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -552,7 +552,7 @@ export class ColorRegistry {
 
     this.registerColor(`${ct}card-carousel-card-text`, {
       dark: colorPalette.gray[400],
-      light: colorPalette.purple[900],
+      light: colorPalette.charcoal[500],
     });
 
     this.registerColor(`${ct}card-carousel-nav`, {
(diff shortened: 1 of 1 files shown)


benoitf (approved):

packages/main/src/plugin/color-registry.ts
```diff
@@ -1139,31 +1139,31 @@ export class ColorRegistry {
...
+      light: colorPalette.red[700],
     });
     this.registerColor(`${status}waiting`, {
       dark: colorPalette.amber[600],
       light: colorPalette.amber[600],
     });
     // Podman only
     this.registerColor(`${status}starting`, {
-      dark: colorPalette.green[600],
+      dark: colorPalette.green[500],
```
cdrage, line 1156: Maybe I missed it, but starting and running is now the same colour? Should it be slightly different?

cdrage: Either way I'm 👍, as the colour difference wasn't much anyways, just wanted to confirm that this is intended.

cdrage (approved):

packages/main/src/plugin/color-registry.ts
deboer-tim: Yup, it is intentional: https://github.com/containers/podman-desktop/issues/6925#issuecomment-2188592434. It's the one part I'm not totally convinced about, but as you say it is a minor difference.

PR #8237, opened 2024-07-25
feat: light and dark mode for terminals

### LINKED ISSUE #7545: Light mode for terminals

cdrage:
feat: light and dark mode for terminals

### What does this PR do?

* Switches to light / dark mode based on appearance settings for
  terminal
* Updates Kubernetes terminal theme on load even if using saved state
* Updated all terminals throughout PD

### Screenshot / video of UI

https://github.com/user-attachments/assets/492eb4dd-adca-4b66-9087-2c9fcc0bdb68

Closes https://github.com/containers/podman-desktop/issues/7545

### How to test this PR?

1. Go to appearance settings
2. Switch to light mode
3. Check any page with a terminal (container or pods terminal)
4. See that it is in light mode

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -254,6 +254,7 @@ export class ColorRegistry {
     this.initOnboarding();
     this.initStates();
     this.initFiles();
+    this.initTerminal();
   }
 
   protected initDefaults(): void {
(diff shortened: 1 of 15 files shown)


cdrage:
@deboer-tim @benoitf This is an update from the other PR which implements it the color-registry.ts way.

I've made it as simple as possible (see the function in `terminal-theme.ts` which makes it easy to implement any other themes in the future (we just add it to color-registry and that's it!).

Let me know if this is a good way to implement these changes.

FYI. The reasoning for https://github.com/containers/podman-desktop/pull/8237/files#diff-e67540f322a92657454ed7f3e2fbfeb3008024f4a2f4dbef9b0915b880bc9b1fR49 is that it makes it easy to get the CSS variables on load from the root document / CSS styling for instant loading from the renderer. If there is a better way to do it, let me know, but I found this the most straight-forward way to add themes (and more themes in the future), if one were to edit the (for example the `ansiGreen` within another extension).

benoitf (approved):
thanks @cdrage it looks more in sync with our theming solution.

I guess the remaining item is the TODO for being able to change/refresh the colors of the terminal when the theme is being changed on the Operating system or through the configuration



I'm wondering if we need the code about extracting the CSS variables.

Assuming we could maybe just take all the variables starting by `--pd-terminal-`
What would be the issue if we have eventually more values ?
it could reduce the size and reduce the steps for the usage. Only need to add a value in the color-registry to make it available in the terminal theme vs adding the definition but adding the property name

cdrage:
> thanks @cdrage it looks more in sync with our theming solution.
> 
> 
> 
> I guess the remaining item is the TODO for being able to change/refresh the colors of the terminal when the theme is being changed on the Operating system or through the configuration
> 
> 
> 
> 
> 
> 
> 
> I'm wondering if we need the code about extracting the CSS variables.
> 
> 
> 
> Assuming we could maybe just take all the variables starting by `--pd-terminal-`
> 
> What would be the issue if we have eventually more values ?
> 
> it could reduce the size and reduce the steps for the usage. Only need to add a value in the color-registry to make it available in the terminal theme vs adding the definition but adding the property name

I fixed it! With this solution the terminal auto updates when you change the theme and you no longer need to restart / refresh. 

I'm unsure on your second point what you mean. If we add more variables if we wanted to customize xterm more, we could. For example, overriding the default "high contrast" colours. But that'd be a different theme in the future I guess!

benoitf:
> I'm unsure on your second point what you mean. If we add more variables if we wanted to customize xterm more, we could. For example, overriding the default "high contrast" colours. But that'd be a different theme in the future I guess!


my second point is:
instead of having 

```
// List of properties which are part of ITheme that we will use
type ThemeProperties =
  | 'foreground'
  | 'background'
  | 'cursor'
  | 'selectionBackground'
  | 'selectionForeground'
  | 'black'
  | 'red'
  | 'green'
  | 'yellow'
  | 'blue'
  | 'magenta'
  | 'cyan'
  | 'white'
  | 'brightBlack'
  | 'brightRed'
  | 'brightGreen'
  | 'brightYellow'
  | 'brightBlue'
  | 'brightMagenta'
  | 'brightCyan'
  | 'brightWhite';
```

you just use the property names collected by listing all css colors starting from `--pd-terminal-` and keeping the last part as property name

if `--pd-terminal-background` is found, then we will add `background` property in the ITheme

so no need of ThemeProperties, we based everything from `computedStyle` and properties starting by `--pd-terminal-`

cdrage:
> > I'm unsure on your second point what you mean. If we add more variables if we wanted to customize xterm more, we could. For example, overriding the default "high contrast" colours. But that'd be a different theme in the future I guess!
> 
> my second point is: instead of having
> 
> ```
> // List of properties which are part of ITheme that we will use
> type ThemeProperties =
>   | 'foreground'
>   | 'background'
>   | 'cursor'
>   | 'selectionBackground'
>   | 'selectionForeground'
>   | 'black'
>   | 'red'
>   | 'green'
>   | 'yellow'
>   | 'blue'
>   | 'magenta'
>   | 'cyan'
>   | 'white'
>   | 'brightBlack'
>   | 'brightRed'
>   | 'brightGreen'
>   | 'brightYellow'
>   | 'brightBlue'
>   | 'brightMagenta'
>   | 'brightCyan'
>   | 'brightWhite';
> ```
> 
> you just use the property names collected by listing all css colors starting from `--pd-terminal-` and keeping the last part as property name
> 
> if `--pd-terminal-background` is found, then we will add `background` property in the ITheme
> 
> so no need of ThemeProperties, we based everything from `computedStyle` and properties starting by `--pd-terminal-`

Ah I agree, have it more dynamically and cross-check with ITheme. I'll add logic too so that it will add it to ITheme if it's an available option, etc.

cdrage:
Ready for another review @benoitf I spent a few hours trying to see if we could dynamically retrieve the properties from the ITheme object, but was unable to. Either I was able to do it and it would conflict with the built-in 'extendedAnsi' string[] array key, or retrieving the properties would appear blank.

Either way, I refactored it so we only have one `KNOWN_THEME_PROPERTIES` value, which covers what's in ITheme (with the exception of extendedAnsi), and makes it a safer for retrieving.

benoitf (approved):
re-approving

deboer-tim (approved):

lstocchi (approved):
tested on win, works fine 🚀

PR #8247, opened 2024-07-26
feat: allow to use libkrun and applehv machines starting from 5.2.0

### LINKED ISSUE #8084: Allow to create libkrun machines on MacOS if Podman 5.2+ is installed
Implementation proposal:

- when listing machines, list both applehv and libkrun machines
- when creating a machine use CONTAINERS_MACHINE_PROVIDER to switch between providers
- the machine provider type select box should be displayed only on MacOS and if Podman 5.2+ is detected

lstocchi:
### What does this PR do?

This PR allows to manage both applehv and libkrun machines at the same time starting from 5.2.0 .
Even though a provider is defined in the `containers.conf` you should be able to work with both applehv and libkrun machines (so create, start, stop, remove, edit ... actions).

The only problem is that libkrun only supports max 8 cpus. If you create a machine with 8+ cpus, you are able to create it but it will always fail to start. So, before creating a machine, I reduce the cpu value to 8 if it's larger. It would be good to edit the maximum value at the UI level but I didn't find an easy way so far. Any idea is appreciated.

If Podman is less than 5.1.0 or you are not on a Mac it should work as before.

In draft as I still have to try it on Windows and write tests but it's ready if you want to start playing with it

### Screenshot / video of UI

https://github.com/user-attachments/assets/fd13ca3c-a075-40a2-86fc-e03b325b966f

it resolves #8084 

### How to test this PR?

1. edit the value here with the Podman version you are using https://github.com/containers/podman-desktop/compare/main...lstocchi:i8084?expand=1#diff-c885162dd627b7cd812c83544b19b9b166179fec79ddbc13273fede7c2566791R1792
2. you should see and being able to use both applehv and libkrun machines. When using the cli you will only have visibility on the current provider set.

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -43,10 +43,13 @@ import {
   appHomeDir,
   execPodman,
   getAssetsFolder,
+  getProviderByLabel,
+  getProviderLabel,
   isLinux,
   isMac,
   isWindows,
   LoggerDelegator,
+  VMTYPE,
 } from './util';
 import { getDisguisedPodmanInformation, getSocketPath, isDisguisedPodman } from './warnings';
 import { WslHelper } from './wsl-helper';
@@ -271,7 +274,7 @@ export async function updateMachines(
         machineInfo?.memory !== undefined && machineInfo?.memoryUsed !== undefined && machineInfo?.memoryUsed > 0
           ? (machineInfo?.memoryUsed * 100) / machineInfo?.memory
           : 0,
-      vmType: machine.VMType,
+      vmType: getProviderLabel(machine.VMType),
     });
 
     if (!podmanMachinesStatuses.has(machine.Name)) {
@@ -796,7 +799,7 @@ export async function registerProviderFor(
     endpoint: {
       socketPath,
     },
-    vmType: machineInfo.vmType,
+    vmType: getProviderLabel(machineInfo.vmType),
   };
 
   // Since Podman 4.5, machines are using the same path for all sockets of machines
@@ -1850,14 +1853,14 @@ export async function createMachine(
 
   let provider: string | undefined;
   if (params['podman.factory.machine.provider']) {
-    provider = params['podman.factory.machine.provider'];
+    provider = getProviderByLabel(params['podman.factory.machine.provider']);
   }
 
   // cpus
   if (params['podman.factory.machine.cpus']) {
     let cpusValue = params['podman.factory.machine.cpus'];
     // libkrun has an issue that prevent to start a machine that has been created with more than 8 cpus, so we limit it here
-    if (provider === 'libkrun' && parseInt(cpusValue) > 8) {
+    if (provider === VMTYPE.LIBKRUN && parseInt(cpusValue) > 8) {
       cpusValue = '8';
     }
     parameters.push('--cpus');
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -99,6 +99,7 @@ export type MachineJSON = {
   Running: boolean;
   Starting: boolean;
   Default: boolean;
+  VMType: string;
   UserModeNetworking?: boolean;
 };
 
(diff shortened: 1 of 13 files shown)


extensions/podman/src/extension.ts
```diff
@@ -297,13 +305,16 @@ export async function updateMachines(
...
-            machineName,
-          ]);
+          const podmanMachineInfo = podmanMachinesInfo.get(machineName);
+          const { stdout: socket } = await extensionApi.process.exec(
+            getPodmanCli(),
+            ['machine', 'inspect', '--format', '{{.ConnectionInfo.PodmanSocket.Path}}', machineName],
+            {
+              env: {
+                CONTAINERS_MACHINE_PROVIDER: podmanMachineInfo?.vmType ?? '',
+              },
```
benoitf: this code is repeating too many times that I think we may miss the right call.

I guess we would need a custom exec function for podman extension adding automatically that env variable

lstocchi: Done

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -530,8 +531,12 @@ function hasAnyConfiguration(provider: ProviderInfo) {
...
                 updateConnectionStatus={updateContainerStatus}
                 addConnectionToRestartingQueue={addConnectionToRestartingQueue} />
-              <div class="mt-1.5 text-gray-900 text-[9px]" aria-label="Connection Version">
-                <div>{provider.name} {provider.version ? `v${provider.version}` : ''}</div>
+              <div class="mt-1.5 text-gray-900 text-[9px] flex justify-between">
+                <div aria-label="Connection Version">
+                  {provider.name}
+                  {provider.version ? `v${provider.version}` : ''}
+                </div>
+                <div aria-label="Connection Type">{container.vmType ? capitalize(container.vmType) : ''}</div>
```
jeffmaury, line 539: Only capitalize the first letter that looks strange to me ?

jeffmaury:
Can't make it to work: I have 5.2.0-rc2 installed on my Mac-M2, I can see the VMtype but can't choose the VM type while create a Podman machine

jeffmaury:
> Can't make it to work: I have 5.2.0-rc2 installed on my Mac-M2, I can see the VMtype but can't choose the VM type while create a Podman machine

My bad, I did not follow the trick for the version check

jeffmaury (approved):
LGTM. Tested on Mac Pro M2.

Should'nt we have the VM type after the name of the machine ?

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
lstocchi: Would you prefer to have the name lowercase?

extensions/podman/src/podman-install.ts
```diff
@@ -209,9 +215,9 @@ export class PodmanInstall {
     const machinesRunning: MachineJSON[] = [];
     try {
       const machineListOutput = await getJSONMachineList();
-      const machines = JSON.parse(machineListOutput.stdout) as MachineJSON[];
+      console.log(JSON.stringify(machineListOutput));
```
benoitf: to be removed I suppose ?

lstocchi:
> LGTM. Tested on Mac Pro M2.
> 
> Should'nt we have the VM type after the name of the machine ?

There was no mockup so i put it there but i can move it. 
@ekidneyrh any preference where it should be located?

extensions/podman/src/podman-install.ts
lstocchi: Thanks, missed it. Done 👍

jeffmaury:
> @ekidneyrh any preference where it should be located?

After the name looks good to me as it is also related to the custom image path so they will be close one to the other

benoitf:
I'm wondering if we should not add a markdownDescription to the field of selection for vmType

Also I'm not sure if for an end user, appleHV or libkrun is meaningful. I know it's the value of the provider of podman, but do we need to expose that first or something like `default (Apple HyperVisor)` and `GPU passthrough (LibKrun)` or something else

benoitf:
also, should it be the first option if it's not clear for users ? or it's a more advanced parameter (bottom option)

lstocchi:
@benoitf @jeffmaury updated by following Florent suggestion. You can see the new labels in the video in the first post.

@jeffmaury i kept the name in the same position as you said, let me know what you think.

benoitf (approved):

jeffmaury (approved):
LGTM

PR #8251, opened 2024-07-26
chore: improvements to light mode

deboer-tim:
### What does this PR do?

A small group of us met to try to improve some issues with light mode. We didn't get through the whole UI, but made the following changes:
- Added a new lighter gray (-25) b/c we needed another shade between -50 and white.
- Tweaked color values, generally making everything as light as possible while still having some contrast.
- Replaced light/dusty purples in main and Settings navigation with light gray.
- Fix color of Settings selected nav item text (was hard-coded color and incorrect class).
- Reduced font weight of settings navigation and preferences items.

We also discussed changing the box shadows to a single line - will cover this in a separate PR.

### Screenshot / video of UI

Before:

Screenshot 2024-07-26 at 3 06 38 PM

Screenshot 2024-07-26 at 3 06 53 PM

After:

Screenshot 2024-07-26 at 2 57 01 PM

Screenshot 2024-07-26 at 2 56 36 PM

### What issues does this PR fix or reference?

First part of #8250.

### How to test this PR?

Check regular nav pages like Containers, Images, and the Setting pages.

Issue #8250, opened 2024-07-26
Improved light mode


deboer-tim:
### Is your enhancement related to a problem? Please describe

A group of us met to try to improve the look of light mode.

### Describe the solution you'd like

I'm going to leave this somewhat open in case we meet again, but in the first session we identified several areas to improve:
- Generally lighten backgrounds/cards to make the mode lighter.
- Reduced use of purple, more similar to how we use it in dark mode.
- Remove or reduce the box shadows as they aren't working well in light mode.

PR #8252, opened 2024-07-26
chore: remove color: from color vars

deboer-tim:
### What does this PR do?

While working on #8251 I was confused by a bunch of places where we use '-[color:var(' instead of just '-[var('. I'm not sure where it came from, but for consistency we should just keep the simpler format.

### Screenshot / video of UI

N/A

Indirectly part of #8250.

### How to test this PR?

Just confirm tests pass, no obvious regressions.


benoitf (changes requested):
It's to remove ambiguities
https://tailwindcss.com/docs/adding-custom-styles#resolving-ambiguities

It should be kept

deboer-tim:
I'm a little confused as Tailwind was figuring out the type of these values fine when I removed color: and we're not using it in the majority of cases, but ok, I'll withdraw this.

PR #8262, opened 2024-07-29
feat: basic light mode for monaco editors

### LINKED ISSUE #8201: Light mode for MonacoEditor

We use MonacoEditor in many places, complete oversight on my part that they aren't terminals (#7545) and they need to support light mode as well.

Do something similar to #7545 to add light mode (and maybe reuse same themes).

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

The Monaco editor currently uses the vs-dark theme with a custom hard-coded background color. This updates it to use the Monaco built-in light or dark theme based on the current appearance, and uses our theme's terminal background color for consistency.

We may want to support full customization of Monaco in the future, but there are lots of possible colors (variables, constants, comments, etc.) so it is not trivial. This just does the basic step to keep what we have today with a consistent look in light mode.

### Screenshot / video of UI

Before:

Screenshot 2024-07-29 at 10 49 07 AM

After:

Screenshot 2024-07-29 at 10 43 11 AM

Fixes #8201.

### How to test this PR?

Open Inspect and Kube tabs in light and dark mode.

### DIFF
--- a/packages/renderer/src/lib/editor/MonacoEditor.svelte
+++ b/packages/renderer/src/lib/editor/MonacoEditor.svelte
@@ -5,7 +5,7 @@ import jsonWorker from 'monaco-editor/esm/vs/language/json/json.worker?worker';
 import { createEventDispatcher, onDestroy, onMount } from 'svelte';
 
 import { EditorSettings } from '../../../../main/src/plugin/editor-settings';
-import { getPanelDetailColor } from '../color/color';
+import { AppearanceUtil } from '../appearance/appearance-util';
 
 let divEl: HTMLDivElement;
 let editor: monaco.editor.IStandaloneCodeEditor;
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/editor/MonacoEditor.svelte
```diff
@@ -35,12 +35,31 @@ onMount(async () => {
...
+  // Monaco only supports 6 char rgb values, so convert if we have 3 char format
+  if (bgColor?.length < 6) {
+    bgColor = bgColor
+      .split('')
+      .map(c => {
+        return c === '#' ? c : c + c;
+      })
+      .join('');
+  }
+
```
benoitf, line 55: should it be a function of appearanceUtil, to be able to get the value of a css property ?

benoitf (approved):

axel7083 (approved):
Looks great !

image

packages/renderer/src/lib/editor/MonacoEditor.svelte
deboer-tim: Yup, I think it will come up again. Since I have approvals here I'll do this via followup PR.

deboer-tim: #8264

PR #8269, opened 2024-07-30
chore: use dropdownmenu component for always on buddy widget popup menu

### LINKED ISSUE #6660: Use DropDownMenu component for always on `Buddy widget` popup menu

Always on 'Buddy' widget on top of settings cog is using native electron popup menu to represent authentication requests and active sessions. 

image

Everywhere else `DropDownMenu` component is used instead. 

image

The problem is DropDownMenu always creates a Button widget and always shown on to the left of that button and cannot be used at the moment.

image

Fix `DropDownMenu` behavior and migrate `Buddy widget` to use it.

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Use DropDownMenu component for always on Buddy widget popup menu

### Screenshot / video of UI
Before:

https://github.com/user-attachments/assets/2da92972-d5fd-4a27-baf8-c4777331d9c1

After:

https://github.com/user-attachments/assets/c3bf1ae2-e5ce-430e-badf-17798e461bd1

Closes #6660 

### How to test this PR?
Try to open account in the NavBar

- Tests are covering the bug fix or the new feature


gastoner:
There are a frew things that needs to be fixed:
1. The NavBar and the DropdownMenu has their own button 
2. The onhover thing on nav bar is blocking the DropDownMenuItems => solved by changing `z-10` to `z-20` in DropDownMenuItems 
3. Once clicked the Account NavBar / DropdownMenu => screen will go to unknown page? => solved by `onClick={() => {}}`
Any ideas how to fix this 1?

deboer-tim:
> There are a frew things that needs to be fixed:
> 
> 1. The NavBar and the DropdownMenu has their own button
> 2. The onhover thing on nav bar is blocking the DropDownMenuItems => solved by changing `z-10` to `z-20` in DropDownMenuItems
> 3. Once clicked the Account NavBar / DropdownMenu => screen will go to unknown page? => solved by `onClick={() => {}}`
>    Any ideas how to fix this 1?

I haven't looked in any detail, but I'd guess you're going to have to make the NavItem natively support dropdown/selection instead of putting a dropdown inside of it. That would solve 1 & 3 (nav item wouldn't trigger a url in this case, it would trigger the drop down).

For 2, changing the z order isn't enough/not the right fix because you'd have a tooltip partially hidden under a menu. But again, the same solution above would fix this too.

dgolovin:
There are issues that prevents to use DropDownMenu component for the Buddy NavItem. I Would suggest to fix them with different PRs to make it easier to review. 

This pr should be straight forward: 
- expose 'getAccountsMenuInfo'
- in AuthActions iterate over array of menuInfo items to build DropDownMenu

dgolovin:
There should be the issues opened for DropDownMenu:
1. It should let to `attach` DropDownMenu to existing element so there would be no second button
2. The placement of DropDownMenu should be fixed, right now it is always shown on the left of element it attached to.
3. (optional) Extreme use cases, like what if there are so many items the DropDownMenu does not fit into the window.

gastoner:
> This pr should be straight forward:
>     expose 'getAccountsMenuInfo'

There is no need to expose the `getAccountsMenuInfo` since there is already a variable containing the array of providers (the same thing is done somewhere else in the code)

>    in AuthActions iterate over array of menuInfo items to build DropDownMenu

This is fixed in current commit

gastoner:
> There should be the issues opened for DropDownMenu:
>    1. It should let to `attach` DropDownMenu to existing element so there would be no second button
>    2. The placement of DropDownMenu should be fixed, right now it is always shown on the left of element it attached to.
>    3. (optional) Extreme use cases, like what if there are so many items the DropDownMenu does not fit into the window.

1. I've removed the `DropdownMenu` element (but left the `DropDownMenuItems`) which removed the redundant button
2. So create a new issue/PR for changes in `DropDownMenuItems.svelte` (left/right alignment)
3. For this would be required to change some top/bottom paddings in `DropDownMenuItems.svelte`

packages/ui/src/lib/dropdownMenu/DropDownMenuItems.svelte
```diff
@@ -17,13 +21,20 @@ onMount(() => {
   } else {
     dropDownElement.style.top = '20px';
   }
+
+  if (window.innerWidth - clientX - NAV_BAR_WIDTH < dropDownWidth) {
```
axel7083: could you add some comments in the code to explain the purpose of the following changes ?

If I understand correctly it is to choose if the dropdown should be placed on the right or left of the user click position ?

axel7083 (commented):
I would recommend to ease the review to move the changes of `DropDownMenuItems.svelte` and `DropdownMenu` in a separate PR with a dedicated test

packages/renderer/src/lib/authentication/AuthActions.svelte
```diff
@@ -0,0 +1,66 @@
+<script lang="ts">
+import { faKey, faSignIn, faSignOut } from '@fortawesome/free-solid-svg-icons';
+
+import DropDownMenuItem from '../../../../ui/dist/dropdownMenu/DropDownMenuItem.svelte';
```
axel7083:
```suggestion
import { DropdownMenu } from '@podman-desktop/ui-svelte';
```

You cannot import manually from the `ui` package, you need to use `@podman-desktop/ui-svelte`

To uses the DropDownMenuItem you should import DropdownMenu, and uses like `DropdownMenu.Item`

You can find example of usage here

https://github.com/containers/podman-desktop/blob/516922e2ef043c208d55e9921edda79f7798aa20/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte#L346-L351

axel7083:
cc @benoitf you are assign to https://github.com/containers/podman-desktop/issues/7758, is there a risk of conflict/overlap here?

From https://github.com/containers/podman-desktop/issues/7758 
_Users should be able to right-click on the navigation bar to see a list of all navigation items in a checkbox list, and easily add or remove the visible items._

This may have some similarity with this PR goals, at least for the changes in the `NavItem`

benoitf:
@axel7083 I see no overlap as right click is/will not be handled by the svelte dropdown menu (it's based on electron right clicks menu)

only left click should/are using svelte components.

packages/ui/src/lib/dropdownMenu/DropDownMenuItems.svelte
dgolovin: IMO dropdown menu should have no knowledge about navbar. It is independent UI component that can be used anywhere.

dgolovin (changes requested):
Please rebase on top of latest origin/main and remove merge artifacts.

packages/ui/src/lib/dropdownMenu/DropDownMenuItems.svelte
gastoner: @dgolovin the `NAV_BAR_WIDTH` has been removed

packages/renderer/src/lib/authentication/AuthActions.svelte
```diff
@@ -0,0 +1,66 @@
+<script lang="ts">
+import { faKey, faSignIn, faSignOut } from '@fortawesome/free-solid-svg-icons';
+import { DropdownMenu } from '@podman-desktop/ui-svelte';
+
+import DropDownMenuItems from '../../../../ui/dist/dropdownMenu/DropDownMenuItems.svelte';
```
axel7083: You should not import from this package, please use `@podman-desktop/ui-svelte`

dgolovin (changes requested):
image

The text for sign out item should include Authentication Provider label, because the same email can be used with different providers and it would not be possible to distinguish github account from Red Hat SSO.

`Sign out of Red Hat SSO (account@gmail.com)`

Would look better.

packages/renderer/src/lib/authentication/AuthActions.svelte
```diff
@@ -0,0 +1,66 @@
...
+</script>
+
+<svelte:window on:keyup={handleEscape} on:click={onWindowClick} />
+
+{#if showMenu}
+  <DropDownMenuItems clientY={clientY} clientX={clientX}>
+    <DropdownMenu.Item
+      title="Manage authentication"
+      icon={faKey}
+      onClick={() => (window.location.href = '#/preferences/authentication-providers')} />
```
dgolovin:
```suggestion
      onClick={() => handleNavigation({
        page: NavigationPage.AUTHENTICATION
      })}
```

packages/renderer/src/lib/authentication/AuthActions.svelte
```diff
@@ -0,0 +1,66 @@
+<script lang="ts">
+import { faKey, faSignIn, faSignOut } from '@fortawesome/free-solid-svg-icons';
+import { DropdownMenu } from '@podman-desktop/ui-svelte';
+
+import DropDownMenuItems from '../../../../ui/dist/dropdownMenu/DropDownMenuItems.svelte';
+import { authenticationProviders } from '../../stores/authenticationProviders';
+
```
dgolovin, line 9:
```suggestion
import { handleNavigation } from '/@/navigation';
import { NavigationPage } from '/@api/navigation-page';

```

dgolovin (changes requested):

dgolovin (dismissed):
Sign in request is now show as 

image

but should be 

image

gastoner:
> Sign in request is now show as
> 
> image
> 
> but should be
> 
> image

I've forgot to add it also to sign-in :D

gastoner:
I have discussed this with @axel7083 and we have come to a point where we need to decide which options would be best, either this ^^ or [this one](https://github.com/gastoner/podman-desktop/pull/2) which uses the `DropdownMenu` and is neater, but have a few drawbacks: 
1. The `Account NavItem` is slightly taller than it should be
2. The `DropDownMenuItems` will be only opened when is clicked on the `AccountImage`

What do you think? More complex (This PR ^^) or more neater solution ([this one](https://github.com/gastoner/podman-desktop/pull/2)) Thanks 🙏 @benoitf @deboer-tim @jeffmaury

dgolovin:
> I have discussed this with @axel7083 and we have come to a point where we need to decide which options would be best, either this ^^ or [this one](https://github.com/gastoner/podman-desktop/pull/2) which uses the `DropdownMenu` and is neater, but have a few drawbacks:
> 
>     1. The `Account NavItem` is slightly taller than it should be
> 
>     2. The `DropDownMenuItems` will be only opened when is clicked on the `AccountImage`
> 
> 
> What do you think? More complex (This PR ^^) or more neater solution ([this one](https://github.com/gastoner/podman-desktop/pull/2)) Thanks 🙏

`Buddy` icon spacing is the same in both branches. It is slightly bigger because of two 'Tooltip' component nested one in another. One comes from NavItem another one wrapping AuthActions. The second PR has different behavior on DropdownMenu cancellation using keybord. That is the only difference I see.

On the other hand it seems that Dropdown Menu on buddy icon just duplicates Authentication settings page features in a form of dropdown menu. We might consider just move Authentication out of preferences and show it when user clicks on buddy icon.

packages/renderer/src/lib/authentication/AuthActions.svelte
```diff
@@ -0,0 +1,69 @@
...
+  <DropdownMenu.Items clientY={clientY} clientX={clientX}>
+    <DropdownMenu.Item
+      title="Manage authentication"
+      icon={faKey}
+      onClick={() => handleNavigation({ page: NavigationPage.AUTHENTICATION })} />
+
+    {#each $authenticationProviders as provider}
+      {console.error(provider)}
+      {@const sessionRequests = provider.sessionRequests ?? []}
+      {#if provider?.accounts?.length > 0}
```
feloy, line 52: I'm not comfortable with this logic. It seems to me that if you are connected with one account, you will get a Sign out for all accounts.

Were you able to test with several auth providers?

packages/renderer/src/lib/authentication/AuthActions.svelte
```diff
@@ -0,0 +1,69 @@
...
+        {#each provider.accounts as account}
+          <DropdownMenu.Item
+            title="Sign out of {provider.displayName} ({account.label})"
+            onClick={() => window.requestAuthenticationProviderSignOut(provider.id, account.id)}
+            icon={faSignOut} />
+        {/each}
+      {/if}
+
+      {#each sessionRequests as request}
```
feloy, line 60: I'm not comfortable with this logic, I cannot see easily the exclusion between Sign In and Sign out.
Could it be possible to write it with an if/else instead?

gastoner: I've used the same logic as in packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.svelte file. But sure I can simplify it

feloy (approved):
Works as expected
LGTM

dgolovin (approved):
Tested with multiple extensions adding requests for authentication.
LGTM

PR #8249, opened 2024-07-26
chore: add page rendering for images/imageId/engineId

SoniaSandler:
### What does this PR do?
Adds image list page rendering for the `images/imageId/engineId` path necessary for the no tag case of https://github.com/containers/podman-desktop/pull/7677 

### Screenshot / video of UI

[Screencast from 2024-07-26 11-25-42.webm](https://github.com/user-attachments/assets/ef6623f7-4560-4147-b54b-e1cf1268fc6e)

^ this is just an example where images with tags were used, but the `images/imageId/engineId` path will be used only for images with id instead of a tag

### What issues does this PR fix or reference?
Closes https://github.com/containers/podman-desktop/issues/8216

### How to test this PR?

- [X] Tests are covering the bug fix or the new feature

PR #8270, opened 2024-07-30
fix: move route with params after route with fixed value

### LINKED ISSUE #8267: Can't start image (Start Container button not accessible)
### Bug description

If Podman Desktop is not maximized I cannot access the Start Container button while starting an image

### Operating system

Win11Pro

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. Go to the image page
2. Start an image

### Relevant log output

_No response_

### Additional context

_No response_

### LINKED ISSUE #8268: Strange default filter when starting image
### Bug description

I tried to start fedora:40 and the image filter in the start container page is set to `run`

run

### Operating system

Win11Pro

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

#8249 introduced a new route `/images/:id/:engineId`, before an existing route `/images/run/*`, which has for effect to catch `/images/run/basic` in the first route instead of the second.

This PR moves the `/images/run/*` route before all other routes with parameters.

Fixes #8267 
Fixes #8268

### How to test this PR?

### DIFF
--- a/packages/renderer/src/App.spec.ts
+++ b/packages/renderer/src/App.spec.ts
@@ -1,0 +1,78 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { render } from '@testing-library/svelte';
+import { tick } from 'svelte';
+import { router } from 'tinro';
+import { beforeEach, expect, test, vi } from 'vitest';
+
+import App from './App.svelte';
+
+const mocks = vi.hoisted(() => ({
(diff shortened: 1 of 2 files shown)


axel7083 (approved):
Thanks ! LGTM

jeffmaury (approved):
LGTM

benoitf (approved):

feloy:
I have added some new unit tests for App.svelte. Feel free to review again @benoitf @axel7083 @jeffmaury

benoitf (approved):

packages/renderer/src/App.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+  expect(mocks.RunImage).toHaveBeenCalled();
+});
+
+test('test /images/:id/:engineId route', async () => {
+  render(App);
+  expect(mocks.ImagesList).not.toHaveBeenCalled();
+  expect(mocks.DashboardPage).toHaveBeenCalled();
+  router.goto('/images/an-image/an-engine');
+  await tick();
+  expect(mocks.ImagesList).toHaveBeenCalled();
```
axel7083, line 77: is await vi.waitFor more reliable than the tick? I never know which one to use

feloy: tick is what id used in other tests inside `waitRender`. If I understand correctly, it is the time for Svelte to render the component again. waitFor seems more for unpredictable duration (with some callback reading disk, ...)

benoitf: I agree with @feloy 
I would say tick is the more accurate as you expect the screen to be updated by svelte, while for waitFor you wait for a condition

while at the end it may be the same result

axel7083: Okey thanks noted!

PR #8302, opened 2024-07-31
chore: use isManifestList for manifest guessing

### LINKED ISSUE #7543: Use isManifestList from podman API

`IsManifestList` was added to the API: https://github.com/containers/podman/pull/22266

Instead of using the "guessIsManifest" functionality, we should **first** check to see if IsManifestList is true, so we can avoid guessing.

If IsManifestList does not exist in the API call, we use the guess functionality.

Prioritize IsManifestList over guessing.

_No response_

### Additional context

_No response_

cdrage:
chore: is isManifestList for manifest guessing

### What does this PR do?

* Use isManifestList instead of guessing.
* Guessing has proven to be very accurate, but this instead will use the
  built-in API from podman that was added in 5.0.2 and above: https://github.com/containers/podman/pull/22266

### Screenshot / video of UI

N/A, should appear as normal for manifests

Closes https://github.com/containers/podman-desktop/issues/7543

### How to test this PR?

1. Create a manifest
2. View image list like normal and see no changes

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -640,11 +640,16 @@ export class ContainerProviderRegistry {
 
         return Promise.all(
           Array.from(fetchedImages).map(async image => {
+            // If image.isManifestList is NOT undefined (5.0.2+), we can use it to determine if the image is a manifest
+            // rather than guessing
+            const isManifest = image.isManifestList ?? guessIsManifest(image, provider.connection.type);
+
+            // Return the base image with the engineName and engineId as well as our isManifest parameter.
             const baseImage = {
               ...image,
               engineName: provider.name,
               engineId: provider.id,
-              isManifest: guessIsManifest(image, provider.connection.type),
+              isManifest,
               Id: image.Digest ? `sha256:${image.Id}` : image.Id,
               Digest: image.Digest || `sha256:${image.Id}`,
             };
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -4706,6 +4706,18 @@ test('manifest is listed as true with podmanListImages correctly', async () => {
     Containers: 0,
   };
 
+  // Purposely set isManifestList to false
+  const manifestImageWithIsManifestListFalse = {
+    ...manifestImage,
+    isManifestList: false,
+  };
+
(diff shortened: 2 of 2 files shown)


packages/api/src/image-info.ts
```diff
@@ -27,6 +27,11 @@ export interface ImageInfo extends Dockerode.ImageInfo {
   History?: string[];
   Digest: string;
   isManifest?: boolean;
+  // Do not use isManifestList. It was added in >5.0.2 of podman and is here for compatibility reasons,
+  // use isManifest instead in the API.
+  // We add isManifestList to be able to retrieve it from the API podmanListImages function call as it returns <unknown>
+  // note that isManifestList is purposely not added to the extension-api.d.ts file as this is only for internal use.
+  isManifestList?: boolean;
```
benoitf: why are we not removing this property instead of saying 'do not use' ?

cdrage: Sorry, I'm confused.

We have to add isManifestList so we can get use the value when calling libpod.listImages, as it's `<unknown>` return and we rely on ImageInfo to retrieve all information.

We also have to keep `isManifest` since extensions (bootc) and other parts of Podman Desktop use it. 

That's why isManifestList is in image-info.ts, but not in the extensions api.

cdrage: NVM! You are right.

We do not need it as we use:

[let fetchedImages: any[] = [];](https://github.com/cdrage/podman-desktop/blob/6c9c331bcf11e8e91df3c150ebc18944ac2e89dd/packages/main/src/plugin/container-registry.ts#L626-L627)

In the podmanListImages call!

I'll make the change now.

packages/main/src/plugin/container-registry.ts
```diff
@@ -640,11 +640,16 @@ export class ContainerProviderRegistry {
 
         return Promise.all(
           Array.from(fetchedImages).map(async image => {
+            // If image.isManifestList is NOT undefined (5.0.2+), we can use it to determine if the image is a manifest
+            // rather than guessing
+            const isManifest = image.isManifestList ?? guessIsManifest(image, provider.connection.type);
```
axel7083, line 645: Nice handling previous version !

axel7083 (approved):

lstocchi (approved):

Issue #8281, opened 2024-07-30
Indicator of Podman Remote in UI


cdrage:
### Is your enhancement related to a problem? Please describe

We should indicate if a connection is Podman remote, and what the name of the connection is (ex. dev-server, mycoolserver, etc.).

### Describe the solution you'd like

UI indicator for podman remote connection

afbjorklund:
Will it still use the local Podman client version, instead of the remote Podman server version?

* https://github.com/containers/podman-desktop/issues/4215

afbjorklund:
Probably needs some more context, on where the connection is coming from (type vs name)

* https://github.com/containers/podman-desktop/issues/5609#issuecomment-1937837023

afbjorklund:
The lima sockets will show up as "local", since it is doing the ssh tunneling to the "remote"

But ideally they would show the name of the instance used, similar to the ssh server here.
(currently no information is exported, on where the connection is _actually_ coming from)

even if a podman system connection is created (it is optional), it is `unix:` and not `ssh:`

---

Default message looks like:

```
INFO[0011] Message from the instance "podman":          
To run `podman` on the host (assumes podman-remote is installed), run the following commands:
------
podman system connection add lima-podman "unix:///Users/afb/.lima/podman/sock/podman.sock"
podman system connection default lima-podman
podman run quay.io/podman/hello
------
```

The details about the connection is in a ssh config file, but those are not supported yet?

```
ssh -F `limactl ls --format='{{.SSHConfigFile}}' podman` lima-podman
```

If you do create a connection, then the port number needs to be updated on each restart.
You are also missing out on some connection optimizations, like the AES and the ssh.sock

So it would be preferred if the instance name ("podman") could be provided somehow...

```shell
limactl ls --format='ssh://{{.SSHAddress}}:{{.SSHLocalPort}}' "podman"
# if not adding the default ssh key, then it needs to use the lima key
CONTAINER_SSHKEY=$(limactl ls --format='{{.IdentityFile}}' "podman")
```

afbjorklund:
* https://github.com/containers/podman/issues/21113

afbjorklund:
The feature probably should be provided for Docker connections too, but as a separate issue...

The `docker context` has the same unix/ssh separation as the `podman system connection` does?

afbjorklund:
With the current implementation of showing the name instead of the provider, it _will_ show the instance name

* https://github.com/containers/podman-desktop/issues/5207

[...]

PR #8303, opened 2024-07-31
feat: improved environment column (provider info)

deboer-tim:
### What does this PR do?

Currently, the environment columns all show a ProviderInfo that only says 'Podman', 'Docker', or 'Kubernetes'. When you have multiple Podman instances or switch Kubernetes contexts it would be more useful to provide some more information.

This PR makes changes to both environments:
- For Kubernetes contexts, use the name of the provider (e.g. Kind or OpenShift Local) if we can find it based on the current context endpoint (since we only support the current context).
- For container providers, use the name of the provider, e.g. 'Podman'. If the provider has more than one connection (e.g. I have two Podman machines), use the name of the connection (machine)

### Screenshot / video of UI

Before:

Screenshot 2024-07-31 at 10 27 18 AM

After:

Screenshot 2024-07-31 at 10 26 58 AM

Part of #8281.

### How to test this PR?

Try in a simple setup with just one machine/Kubernetes context, as well as with multiple machines, switching context, etc.

Would especially like some feedback from people who have tried the Podman remote support.

### DIFF
--- a/packages/renderer/src/lib/ui/ProviderInfo.spec.ts
+++ b/packages/renderer/src/lib/ui/ProviderInfo.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


axel7083 (commented):
**Before**

image

**after**

image

Love it ! Just have an overflow issue, we might want to add `text-overflow: ellipsis;` and a tooltip?

benoitf:
I don't think it'll handle all cases and how it can handle long names properly

probably the information should be on the hover/tooltip rather than the field itself

axel7083:
> probably the information should be on the hover/tooltip rather than the field itself

I really like having this information available directly without hovering. I think an overflow nicely set handle it well

image

benoitf (commented):
I think the way on retrieving and extrapolating data is not something that should happen on the frontend side.

AFAIK, reconciling data is a server side thing.

afbjorklund:
This doesn't show any difference, between providers that are "local" and providers that are "remote"

i.e. it relies on naming the machines or the instances in some sensible way, to be able to tell them apart

benoitf:
moving to draft

PR #8308, opened 2024-08-01
feat: allow to open dropdownmenu on right side of a menu button

### LINKED ISSUE #8307: Allow to open DropdownMenu on right side of a menu button

Related to #6660 , the DropdownMenu currently doesn't support opening on the right side of a button, only on the left side.
Now:

image

When trying to switch from the `Buddy Widget` to DropdownMenu:

image

Tweek left/right alignment of `DropdownMenuItems` to open on the right side, if the dropdown would open outside of a window

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Allows to open dropdownmenu on right side of a menu button

Closes #8307 
Related to PR #8269 

### How to test this PR?

- [x]  Tests are covering the bug fix or the new feature

### DIFF
--- a/packages/ui/src/lib/dropdownMenu/DropDownMenuItems.svelte
+++ b/packages/ui/src/lib/dropdownMenu/DropDownMenuItems.svelte
@@ -2,9 +2,12 @@
 import { onMount } from 'svelte';
 
 let dropDownHeight: number;
+let dropDownWidth: number;
 let dropDownElement: HTMLElement;
+let sideAlign: string;
 
 export let clientY: number;
+export let clientX: number;
 
 const STATUS_BAR_HEIGHT = 24;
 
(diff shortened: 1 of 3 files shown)


packages/ui/src/lib/dropdownMenu/DropDownMenuItems.svelte
```diff
@@ -17,13 +21,22 @@ onMount(() => {
   } else {
     dropDownElement.style.top = '20px';
   }
+
+  // When initializing the widget, set the placement on left or right
+  // depending on the clientX position (cursor position) and the width of the dropdown menu to display
+  if (window.innerWidth - clientX - NAV_BAR_WIDTH < dropDownWidth) {
```
axel7083: Why do we need the NAV_BAR_WIDTH ?

gastoner: Probably we don't :D

axel7083 (commented):
Okey I though some test would be required to validate the new behaviour, but testing this was surely going to be a bit complicated.

So give it a try to help, the most annoying part is to mock the `clientHeight` and `clientWidth` that are binded values.

The best way of doing this is to directly set the value of the `HTMLElement.prototype`.

```
Object.defineProperty(HTMLElement.prototype, 'clientHeight', { value: 100 });
Object.defineProperty(HTMLElement.prototype, 'clientWidth', { value: 50 });
```

The DropDownMenuItems also requires a mock of the ResizeObserver, which can be copy past from 

https://github.com/containers/podman-desktop/blob/520ea06f058ff97ec90a8cef14a1ca8723092208/packages/ui/src/lib/dropdownMenu/DropdownMenu.spec.ts#L26

Finally, the `window.innerWidth` and `window.innerHeight` could be mock in each test with the following

```
  (window as any).innerHeight = 100;
  (window as any).innerWidth = 100;
```

With all of that you should be able to write some tests, ensuring the position of the dropdown based on screen size, dropdown size !

axel7083 (approved):
Great tests, thanks for adding them. LGTM :rocket:

lstocchi (approved):
LGTM

Issue #8309, opened 2024-08-01
Refactor resource card pages


amisskii:
### Is your enhancement related to a problem? Please describe

The current implementation of Resource Card Pages uses separate classes for each resource type (e.g., Podman, Kind). As a result, introducing a new resource page type requires creating an additional class, leading to code duplication.

### Describe the solution you'd like

To address the issues of code duplication, a refactoring of the Resource Card Pages is necessary:

1. **Introduce an abstract base class (`ResourceCardPage`):**  
   Create a base class that encapsulates common attributes and methods shared by all Resource Cards.

2. **Introduce the subclasses:**
   - **`ResourceConnectionCardPage`:**  
     A subclass for handling resource management tools such as Podman, Kind and Docker.
   - **`ResourceCliCardPage`:**  
     A subclass for managing resources with command-line tools like Kubectl and Docker Compose.

PR #8310, opened 2024-08-01
refactor: refactor resource card pages

amisskii:

### What does this PR do?

1. **Introduce an abstract base class (`ResourceCardPage`):**  
   Create a base class that encapsulates common attributes and methods shared by all Resource Cards.

2. **Introduce the subclasses:**
   - **`ResourceConnectionCardPage`:**  
     A subclass for handling resource management tools such as Podman, Kind and Docker.
   - **`ResourceCliCardPage`:**  
     A subclass for managing resources with command-line tools like Kubectl and Docker Compose.

#8309 

### How to test this PR?
1. Run the [workflow](https://github.com/containers/podman-desktop-e2e/actions/workflows/desktop-e2e-test-job-windows.yaml) on a Windows VM.
2. Use `amisskii` as the fork parameter for the Podman Desktop repository.
3. Use `resources-card` as the branch parameter for the Podman Desktop repository.


tests/playwright/src/model/core/operations.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export enum ResourceElementActions {
```
odockal, line 19: You need to put it on the `index.ts`

tests/playwright/src/model/pages/resource-card-page.ts
```diff
@@ -0,0 +1,40 @@
...
+  readonly providerConnections: Locator;
+
+  constructor(page: Page, resourceName: string) {
+    super(page);
+    this.resourceCard = this.featuredProviderResources.getByRole('region', { name: resourceName, exact: true });
+    this.providerSetup = this.resourceCard.getByRole('region', { name: 'Provider Setup', exact: true });
+    this.providerConnections = this.resourceCard.getByRole('region', { name: 'Provider Connections', exact: true });
+  }
+
+  public async clickOnProviderSetupButton(buttonLabel: string): Promise<void> {
```
odockal: Is this functionality shared across all the card?
Kind, minikube, podman, dev sandbox, openshift local, compose and kubectl?

odockal: possibly, if just one card does not offer it, you could override it and throw an error in the child class.

cbr7 (changes requested):
All new files need to be added to index.ts in order to be included in the npm package

tests/playwright/src/model/pages/resource-card-page.ts
cbr7: I kind of dislike this implementation because it's not obvious that you need to provide the aria-label to the method in order for it to work, this needs to be made clearer.

tests/playwright/src/model/pages/resource-cli-card-page.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { type Page } from '@playwright/test';
+
+import { ResourceCardPage } from './resource-card-page';
+
+export class ResourceCliCardPage extends ResourceCardPage {
+  constructor(page: Page, resourceName: string) {
+    super(page, resourceName);
+  }
+}
```
cbr7, line 27: What is the purpose of this class? It's an empty shell and it's not even used anymore as far as I can tell.

amisskii:
There are two more spec files that need to be refactored (compose-onboarding-smoke.spec.ts, kind.spec.ts)

amisskii:
https://github.com/containers/podman-desktop-e2e/actions/runs/10197385286
https://github.com/containers/podman-desktop-e2e/actions/runs/10210867194

tests/playwright/src/model/pages/resource-cli-card-page.ts
amisskii: I'm not sure what to do with this class. On one hand, I use it in the Compose onboarding spec. On the other hand, this class doesn’t contain any additional attributes or functions that the ResourceCardPage class already has. So, perhaps we should use the ResourceCardPage class for CLI card pages, but I'm not certain.. @odockal @cbr7  WDYT?

cbr7: I don't see anything wrong in using the parent page object.

tests/playwright/src/model/pages/resource-card-page.ts
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { type Locator, type Page } from '@playwright/test';
+
+import { ResourcesPage } from './resources-page';
+
+export class ResourceCardPage extends ResourcesPage {
```
odockal: Do not inherit from `ResourcePage`. Card class should not have access to the whole parent content as we are looking for specialization in this case. If you look at how `ExtensionsPage` and ExtensionsCardPage are implemented, you can use the same class members and similar logic/behavior.

tests/playwright/src/model/pages/resource-card-page.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(page: Page, resourceName: string) {
+    super(page);
+    this.resourceCard = this.featuredProviderResources.getByRole('region', { name: resourceName, exact: true });
+    this.providerSetup = this.resourceCard.getByRole('region', { name: 'Provider Setup', exact: true });
+    this.providerConnections = this.resourceCard.getByRole('region', { name: 'Provider Connections', exact: true });
+    this.markdownContent = this.providerConnections.getByLabel('markdown-content');
+  }
+
+  public async resourceCardIsVisible(): Promise<boolean> {
```
odockal: This method should be part of `ResourcesPage`, when writing a test, we will go from the top level to the bottom level.

tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
```diff
@@ -66,14 +69,13 @@ describe.skipIf(isCI && isLinux)('Compose onboarding workflow verification', asy
   });
 
   test('Compose onboarding button Setup is available', async () => {
-    await navBar.openSettings();
-    const settingsBar = new SettingsBar(page);
-    const resourcesPage = await settingsBar.openTabPage(ResourcesPage);
+    settingsBar = await navBar.openSettings();
+    await settingsBar.openTabPage(ResourcesPage);
+    composeResourceCardPage = new ResourceCardPage(page, 'Compose');
```
odockal: If you will move the `resourceCardIsVisible` then the check will happen before you actually instantiate the Card Page class. which is more logical.

tests/playwright/src/model/pages/resources-page.ts
```diff
@@ -23,16 +23,10 @@ import { SettingsPage } from './settings-page';
...
   readonly featuredProviderResources: Locator;
-  readonly podmanResources: Locator;
-  readonly composeResources: Locator;
-  readonly kindResources: Locator;
 
   constructor(page: Page) {
     super(page, 'Resources');
     this.heading = this.header.getByRole('heading', { name: 'Title' }).and(this.header.getByText('Resources'));
     this.featuredProviderResources = this.content.getByRole('region', { name: 'Featured Provider Resources' });
-    this.podmanResources = this.featuredProviderResources.getByRole('region', { name: 'podman', exact: true });
```
odockal, line 34: You can add functions for checking if provider connection/cli is actually visible/available and also a function that return particular page object. Similarly as in the `ExtensionsPage`.

odockal (changes requested):

tests/playwright/src/model/pages/resource-cli-card-page.ts
odockal: We have different set of behavior for these resource objects. 
One is, let's call it, connection provider and other is CLI provider. Podman resource itself is hybrid of these two.
First one offers Create New button only. Second has Setup button only. Podman has Setup button, create button and then "gear" icon button. 

Which lead me to think, that the best would be to have one parent class (ResouceCardPage) and then specialized children classes for each provider. wdyt?

So it could be broken down like this: 
```
abstract ResourceCardPage
|--> PodmanResourceCardPage (create new, setup, gear buttons, etc., offers actions for each machine)
|--> ComposeResourceCardPage (setup, no actions)
|--> KubectlResourceCardPage (setup, no actions)
|--> DevSandboxResourceCardPage (create new)
|--> OpenShiftResourceCardPage (create new, actions available)
|--> KindResourceCardPage (create new, offers actions for each cluster)
|--> MinikubeResourceCardPage (create new, offers actions for each cluster)
```

Which offers a space for various design approaches. There seems to be multiple features shared across these cards... Ie. what Anton first used seems also ok -> division between CLI (offers setup button and no other actions) and then other that all offers Create New button. So I think at least keeping this approach seems fine for future:

```
abstract ResourceCardPage
|--> ResourceConnectionCardPage (create new) -> with possible inheritance shared in specialized child classes, ie. Podman, Kind, OpenShift, etc.
|--> CLIResourceCardPage (setup, no actions) -> Kubectl and compose.
```

ScrewTSW (approved):
Nice work. If you test it and it works, I'd approve the merge

tests/playwright/src/model/pages/resource-card-page.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(page: Page, resourceName: string) {
+    super(page);
+    this.parent = this.page.getByRole('region', { name: 'content' });
+    this.card = this.parent.getByRole('region', { name: resourceName, exact: true });
+    this.providerSetup = this.card.getByRole('region', { name: 'Provider Setup', exact: true });
+    this.providerConnections = this.card.getByRole('region', { name: 'Provider Connections', exact: true });
+    this.markdownContent = this.providerConnections.getByLabel('markdown-content');
+  }
+
+  public async getSetupButton(buttonLabel: string): Promise<Locator> {
```
amisskii: @cbr7 I think we can use this function instead of defining it in spec files as` provideSetup.getByRole('button')`. The `buttonLabel` parameter corresponds to the button's visible title, which is the same as its `aria-label` in the application. This makes it intuitive to use the button title as seen in the UI. Additionally, this function is useful because it provides an assertion to check if the button exists. WDYT?

tests/playwright/src/model/pages/resources-page.ts
amisskii: I attempted to implement this function, but I find it more complicated than initializing an object in a specific scenario. We need a generic function that can return two types of objects (cli or content).  The function definition should look like this:

```
async getResourceCardPage<T>(
  resourceName: string,
  type: new (page: any, resourceName: string, resourceElementVisibleName?: string) => T,
  resourceElementVisibleName?: string
): Promise<T> 
```

cbr7: Why not implement an interface and return the interface?

tests/playwright/src/model/pages/resource-card-page.ts
cbr7: Hi Anton, there are some issues with this kind of implementation:

1. A method that returns a Locator object does not even need to be a Promise type in reality, since Locator is a an artificial construct of the Playwright framework and it's created synchronously, however you can certainly return a Promise<Locator> for pattern reasons.
2. Getting a locator can be done at any time, even if the actual web element does not exist in the present moment, as such having a method that returns a Locator to asset the object's existence goes against the intended pattern. Asserting existence of an web element should only really be done before that element is actually used.
3. Catching an exception thrown from an assert kind of defeats the purpose, why even use the assert if you are just going to catch it in a try-catch block?

tests/playwright/src/model/pages/resources-page.ts
amisskii: Unfortunately, I am not sure how to implement such a solution. @cbr7, could you please provide a more detailed description?

cbr7: What I mean is that the method does not need to be generic at all, all it needs is to return an interface type which is implemented in all the POM classes that the the method could refer too, for example:

`async getResourceCardPage(resourceName: string, resourceElementVisibleName?: string): Promise<nameOfInterface>`

And all the method can return different POM classes, potentially from a switch statement, as long as all the POM classes that it can return implement nameOFInterface

tests/playwright/src/model/pages/resource-connection-card-page.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.resourceElement = this.providerConnections.getByRole('region', { name: resourceElementVisibleName });
+    this.resourceElementDetailsButton = this.resourceElement.getByRole('button', { name: 'details' });
+    this.resourceElementConnectionStatus = this.resourceElement.getByLabel('Connection Status Label');
+    this.resourceElementConnectionActions = this.resourceElement.getByRole('group', { name: 'Connection Actions' });
+  }
+
+  public async performConnectionAction(operation: ResourceElementActions): Promise<void> {
+    const button = this.resourceElementConnectionActions.getByRole('button', { name: operation, exact: true });
+    await playExpect(button).toBeVisible();
+    await playExpect(button).not.toBeDisabled({ timeout: 10000 });
```
cbr7:
```suggestion
    await playExpect(button).toBeEnabled({ timeout: 10000 });
```

tests/playwright/src/model/pages/resources-page.ts
amisskii: I don't think using an interface is a good idea due to the different properties in POM classes. For example, in:

```typescript
export class ResourceConnectionCardPage extends ResourceCardPage implements IResourceCardPage {
  readonly resourceElement: Locator;
  readonly resourceElementDetailsButton: Locator;
  readonly resourceElementConnectionStatus: Locator;
  readonly resourceElementConnectionActions: Locator;
  ...
}

```

I have specific properties that need to be added to the interface. However, if I include them as optional properties like `resourceElementConnectionActions?: Locator`, it can cause issues when I need to use them in specs because they may be undefined. If I don't make them optional, I would have to implement them in all POM classes, even when they aren't needed. 

I would like to keep my current solution, but maybe I just didn't understand your intentions. Please let me know what you think.

cbr7: I'm not sure what the issue is, if you need to use them somewhere you need to check that they are defined, this is true for all properties in all classes so I'm not sure how it's any different here.

Don't take this the wrong way, but generic methods are extremely difficult to debug and should not be used unless absolutely required and they aren't required here.

Additionally if you dont want to have fields defined in the interface, that is fine, don't have them defined there, have them only in the POM, nobody says the interface itself must have them, all you need is to cast the resulting object that is returned from the interface, which will be of Type interface into whatever POM class you know it is.

amisskii: I think we have a little misunderstanding. :) I don’t use a generic function; I use https://github.com/containers/podman-desktop/blob/49508fb2696dfff445ccc3680d73c95f50272363/tests/playwright/src/specs/compose-onboarding-smoke.spec.ts#L77. If you think that using a function with an interface would be better, I can try that approach again.

cbr7: I was talking about your post where you pasted:

```
async getResourceCardPage<T>(
  resourceName: string,
  type: new (page: any, resourceName: string, resourceElementVisibleName?: string) => T,
  resourceElementVisibleName?: string
): Promise<T> 
```

Which is why I used the same method name when I gave you an example of using the interface for a return type.

I don't know anything about this new line of code you quoted in the last post.

cbr7 (approved):

odockal (approved):
LGTM. good work.

PR #8322, opened 2024-08-01
feat: allow to show/hide navigation items

### LINKED ISSUE #7758: Ability to show/hide navigation items

As we add more navigation items there are more items that are rarely used, extensions that add 2 items but you only use one of them, or cases where you want to temporarily hide the icon without uninstalling the extension.

Users should be able to right-click on the navigation bar to see a list of all navigation items in a checkbox list, and easily add or remove the visible items.

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Allow to toggle the visibility of left navbar items

### Screenshot / video of UI

https://github.com/user-attachments/assets/fbfadd1a-bfb1-475d-9ff8-1232a63ef0e4

fixes https://github.com/containers/podman-desktop/issues/7758

### How to test this PR?

Right click on an item like Containers, Pods and select `Hide <item>`
or right click on any place on the navbar and then check or uncheck the checkbox for a given item

### DIFF
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -47,15 +47,18 @@ function clickSettings(b: boolean) {
   {#each $navigationRegistry as navigationRegistryItem}
     <!-- This is a section -->
     {#if navigationRegistryItem.type === 'section' && navigationRegistryItem.enabled}
-      <NavSection tooltip={navigationRegistryItem.name}>
-        <svelte:component this={navigationRegistryItem.icon.iconComponent} size={iconSize} slot="icon" />
+      {@const allItemsHidden = (navigationRegistryItem.items ?? []).every(item => item.hidden)}
+      {#if !allItemsHidden}
+        <NavSection tooltip={navigationRegistryItem.name}>
+          <svelte:component this={navigationRegistryItem.icon.iconComponent} size={iconSize} slot="icon" />
 
-        {#if navigationRegistryItem.items}
-          {#each navigationRegistryItem.items as item}
-            <NavRegistryEntry entry={item} bind:meta={meta} />
-          {/each}
-        {/if}
-      </NavSection>
+          {#if navigationRegistryItem.items}
+            {#each navigationRegistryItem.items as item}
+              <NavRegistryEntry entry={item} bind:meta={meta} />
+            {/each}
+          {/if}
+        </NavSection>
+      {/if}
     {:else if navigationRegistryItem.items && navigationRegistryItem.type === 'group'}
       <!-- This is a group, list all items from the entry -->
       {#each navigationRegistryItem.items as item}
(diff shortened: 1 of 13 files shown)


deboer-tim (changes requested):
Function working great, but I have some nits and ran into related issues:
- The tooltip stays open when the menu appears. This is the same issue seen in #6660 but not sure the solution would be the same, when you open the menu the tooltip should go away.
- I don't think the menu items should be sorted. Although there's a separate Hide option I was looking for the current menu in the list and this made it harder to scan for.
- Items like 'ConfigMaps & Secrets' are missing the ampersand and show up like 'ConfigMaps  Secrets'. It might be interpreted as an accelerator key and need to be escaped.
- When using `yarn watch` the first two menu items are `Copy Link` and `Inspect`. I'm assuming the Inspect would be gone in production build but Copy Link is distracting (breaks the flow and makes me realize this is electron). Is there a quick way to remove it, or maybe we can track via separate issue?

benoitf:
fixed 2 and 3 (label escaping and sorting)

for 1. and 4. it's already there so should be other PRs/issue.

deboer-tim:
> fixed 2 and 3 (label escaping and sorting)

Thanks for fixing!

> for 1. and 4. it's already there so should be other PRs/issue.

Yeah, it's just unfortunate that users probably wouldn't have known we had these issues until we tell them about this feature. I opened #8335 for #1. For #4 I noticed several other native menus so I opened the broader design question to start with in #8336.

deboer-tim (approved):

axel7083 (changes requested):
Weird behaviour with the account and settings, where we should probably not see the hide action?


https://github.com/user-attachments/assets/23c2ae3c-6ee1-43ae-bdd7-246daf056c9b

If am hidding all kubernetes, the group menu should disapear ?

image

benoitf:
@axel7083 should be fixed

axel7083 (commented):
Works nicely

packages/renderer/src/AppNavigation.svelte
```diff
@@ -47,15 +47,18 @@ function clickSettings(b: boolean) {
   {#each $navigationRegistry as navigationRegistryItem}
     <!-- This is a section -->
     {#if navigationRegistryItem.type === 'section' && navigationRegistryItem.enabled}
-      <NavSection tooltip={navigationRegistryItem.name}>
-        <svelte:component this={navigationRegistryItem.icon.iconComponent} size={iconSize} slot="icon" />
+      {@const allItemsHidden = (navigationRegistryItem.items ?? []).every(item => item.hidden)}
+      {#if !allItemsHidden}
```
axel7083, line 51: Why declaring a variable which has a single usage ?

axel7083, line 51:
```suggestion
      {#if !(navigationRegistryItem.items ?? []).every(item => item.hidden)}
```

axel7083 (approved):
LGTM

packages/renderer/src/AppNavigation.svelte
benoitf: it looks a little bit too long to be directly a condition. I find it more readable.

axel7083: I approved so that's not blocking for me

PR #8374, opened 2024-08-07
chore: add try/catch to ssh connection

cdrage:
chore: add try/catch to ssh connection

### What does this PR do?

Right now we have no error out put when we cannot initially connect, for
example, having a bad key being passed in.

This adds a try / catch so we can make sure that we at least put the
error output in the console.

N/A

### How to test this PR?

N/A, adds try/catch.

### DIFF
--- a/extensions/podman/src/podman-remote-connections.ts
+++ b/extensions/podman/src/podman-remote-connections.ts
@@ -169,7 +169,11 @@ export class PodmanRemoteConnections {
       const sshTunnel = this.createTunnel(host, port, username, privateKey, remotePath, localPath);
 
       // connect the tunnel
-      sshTunnel.connect();
+      try {
+        sshTunnel.connect();
+      } catch (error) {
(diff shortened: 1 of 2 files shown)


extensions/podman/src/podman-remote-connections.ts
```diff
@@ -169,7 +169,11 @@ export class PodmanRemoteConnections {
       const sshTunnel = this.createTunnel(host, port, username, privateKey, remotePath, localPath);
 
       // connect the tunnel
-      sshTunnel.connect();
+      try {
+        sshTunnel.connect();
+      } catch (error) {
+        console.error(`Error connecting to Podman SSH Remote: ${error}`);
```
axel7083, line 175:
```suggestion
        console.error(`Error connecting to Podman SSH Remote`, error);
```

I think we may lose some traces if we stringify the error

cdrage:
Sorry, closing this. It was an actual accidental PR push :(

PR #8379, opened 2024-08-07
docs: add release checklist

cdrage:
docs: add release checklist

### What does this PR do?

Adds a checklist of last steps for the release / make sure all steps are
completed.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

N/A

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -301,3 +301,15 @@ The complete list of issues fixed in this release is available [here](https://gi
 
 Get the latest release from the [Downloads](/downloads) section of the website and boost your development journey with Podman Desktop. Additionally, visit the [GitHub repository](https://github.com/containers/podman-desktop) and see how you can help us make Podman Desktop better.
 ```
+
+## Checklist
+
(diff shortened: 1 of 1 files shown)


dgolovin (approved):
LGTM, but would be Issue Template for release better? check list can be added as default content and after using the template a person responsible for release would select finished tasks and it would make it possible to track progress for the release.

gastoner (approved):

benoitf:
I agree with @dgolovin  that it might be better to move the checkboxes in the template of the `kind/release` issue

so it's always there inside the issue we create.

but we can add an item saying to verify that the release issue is really created using that template as sometimes it's created using a blank form

cdrage:
> I agree with @dgolovin that it might be better to move the checkboxes in the template of the `kind/release` issue
> 
> so it's always there inside the issue we create.
> 
> but we can add an item saying to verify that the release issue is really created using that template as sometimes it's created using a blank form

Ill add this as a github template / issue!

Issue #8103, opened 2024-07-16
Implement new Registries page design


ekidneyrh:
### Is your enhancement related to a problem? Please describe

As decided on https://github.com/containers/podman-desktop/issues/4294, we have a new design for the Registries page. 
The main change is that it's now in the table format that we see already across PD.  Also have 'verifed' labels when the details are validated.

On the UX call it was discussed validating more often to see if the credentials were still working and tell the user when they need to be updates, e.g. through the 'disconnected' label. Hopefully this something that can be implemented at some stage but for now we can use the verified label as a placeholder.

### Describe the solution you'd like

Here's the full mock-up to be implemented:
![#4294 Registries screen @emma](https://github.com/user-attachments/assets/720fba2c-2c1a-43fb-a74f-f06ea7219e47)

The modal to edit an existing registry:
![edit-registry](https://github.com/user-attachments/assets/37f0032c-6ba2-47cb-81fc-a7884325a0c3)

For this modal, the url for the registry is read-only.

The modal to add a new registry:
![add-registry png](https://github.com/user-attachments/assets/ad41924b-5ea8-4146-bf66-281cd65a4d37)

PR #8402, opened 2024-08-08
chore: add registry dialog

deboer-tim:
### What does this PR do?

First part of #8103. Switches adding a new registry from a row at the bottom of the table to a dialog, as per the design.

Keeping this relatively simple for the first step. Future PRs could modify the dialog to support editing as well.

### Screenshot / video of UI

Screenshot 2024-08-08 at 1 52 16 PM

First part of #8103.

### How to test this PR?

Add a new registry.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
@@ -172,6 +172,11 @@ async function loginToRegistry(registry: containerDesktopAPI.Registry) {
   registry.source = defaultProviderSourceName;
 
   const newRegistry = registry === newRegistryRequest;
+  if (newRegistry) {
+    registry.serverUrl = registry.serverUrl.trim();
+    registry.username = registry.username.trim();
+    registry.secret = registry.secret.trim();
+  }
 
   // Always check credentials before creating image / updating to see if they pass.
   // if we happen to get a certificate verification issue, as the user if they would like to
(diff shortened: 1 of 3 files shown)


axel7083 (approved):
LGTM, great work ! Tested and works fine

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
```diff
@@ -438,46 +439,51 @@ function removeExistingRegistry(registry: containerDesktopAPI.Registry) {
...
+    </div>
+    <svelte:fragment slot="validation"
+      ><ErrorMessage error={errorResponses.find(o => o.serverUrl === newRegistryRequest.serverUrl)?.error ?? ''}
+      ></ErrorMessage
+      ></svelte:fragment>
+    <svelte:fragment slot="buttons">
+      <Button type="link" on:click={() => (showNewRegistryForm = false)}>Cancel</Button>
+      <Button
+        type="primary"
+        disabled={!newRegistryRequest.serverUrl || !newRegistryRequest.username || !newRegistryRequest.secret}
```
dgolovin: nit:
```suggestion
        disabled={!newRegistryRequest.serverUrl.trim() || !newRegistryRequest.username.trim() || !newRegistryRequest.secret.trim()}
```

dgolovin (approved):
LGTM.
nit: trim call required for disabled  attribute, right now it would let create registry even if I put spaces in every field.

gastoner (approved):
I tried it out, and it works fine!

packages/renderer/src/lib/ui/PasswordInput.svelte
```diff
@@ -30,7 +30,7 @@ async function onShowHide() {
   class={$$props.class || ''}
   id="password-{id}"
   name="password-{id}"
-  placeholder="Password"
+  placeholder="password"
```
benoitf, line 33: looks like unrelated

benoitf (commented):
image


should trim whitespaces

I think I would expect `Add` button rather than `Login` button

title is 'Add a registry' so I expect to `add` it rather than `login`

image

packages/renderer/src/lib/preferences/PreferencesRegistriesEditing.svelte
deboer-tim: Included in new commit (lint reformatted).

deboer-tim:
New commit:
- Renames button to Add. (I had done this initially and then 'fixed' it to match the design, but this is better)
- Trims whitespace when checking if the button should be enabled.
- Automatically trims whitespace when adding (so if you accidentally type or copy/paste ' https//quay.io ' it is fixed).

packages/renderer/src/lib/ui/PasswordInput.svelte
deboer-tim: Password is a shared component that had a capitalized Password placeholder, but it is only used on this page so far and the design has lowercase placeholders for username and password so I changed it here. I don't have a strong opinion (other than the cases should match) but lower looked slightly better to me.

PR #8403, opened 2024-08-08
chore: allow simple column to wrap to 2 lines

### LINKED ISSUE #5442: Kubernetes services page: List of ports is hidden

When a service has a lot of ports, they are hidden, we should have it wrap to the next line / be able to view them:

Screenshot 2024-01-05 at 2 05 40 PM

Word wrapping / way to view it all.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

I had a change to make service port a special column that could wrap to two lines, but then I figured why not all simple columns? If text is too wide to fit into a column this will cause it to wrap to two lines. Longer text will still use ellipsis. (we could go to 3 or 4 lines, but allowing twice as much text seemed like a reasonable balance that doesn't affect the rest of the table layout)

### Screenshot / video of UI

Before:

Screenshot 2024-08-08 at 2 57 41 PM

After:

Screenshot 2024-08-08 at 2 56 46 PM

Fixes #5442.

### How to test this PR?

Having multiple service ports are the most obvious way to trigger this, but on very narrow columns something like "2 months" will be wrapped to: "2
months"
instead of:
"2 mo..."

### DIFF
--- a/packages/ui/src/lib/table/SimpleColumn.spec.ts
+++ b/packages/ui/src/lib/table/SimpleColumn.spec.ts
@@ -31,6 +31,6 @@ test('Expect simple column styling', async () => {
   expect(text).toBeInTheDocument();
   expect(text).toHaveClass('text-[var(--pd-table-body-text)]');
   expect(text).toHaveClass('max-w-full');
-  expect(text).toHaveClass('overflow-hidden');
-  expect(text).toHaveClass('text-ellipsis');
+  expect(text).toHaveClass('text-wrap');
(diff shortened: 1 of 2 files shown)


cdrage (approved):
Big 👍 👍 

This looks so much better / cleaner and makes me happy since some of my services / deployments have a lot of ports open.

LGTM!

Screenshot 2024-08-08 at 3 53 28 PM
Screenshot 2024-08-08 at 3 53 35 PM

lstocchi (approved):

benoitf (approved):

PR #8406, opened 2024-08-08
chore: fix positioning of check circle

### LINKED ISSUE #8405: Check circle has a fixed placement in dialog box, even when scrolling, when pushing an image
### Bug description

In the dialog box of push image, the green check circle stays in a fixed place even when scrolling.

[Screencast from 2024-08-08 16-24-35.webm](https://github.com/user-attachments/assets/e9fc4295-c8ed-4c40-b8b7-42ecae1be106)

### Operating system

Linux Fedora 39

### Installation Method

### Version

next (development version)

### Steps to reproduce

Go to the push image page and while having the scroll bar of the dialog box available (during an image push), scroll and check the placement of the check circle

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
Fixes the positioning of the check circle in push image page to make sure it stays in the image tag area when scrolling on the dialog box

### Screenshot / video of UI

[Screencast from 2024-08-08 16-50-25.webm](https://github.com/user-attachments/assets/43514c2e-9121-4d29-827f-12e226de1084)

Closes https://github.com/containers/podman-desktop/issues/8405

### How to test this PR?

Go to the push image page and while having the scroll bar of the dialog box available (during an image push), scroll and check the placement of the check circle

### DIFF
--- a/packages/renderer/src/lib/dialogs/Dialog.svelte
+++ b/packages/renderer/src/lib/dialogs/Dialog.svelte
@@ -19,7 +19,7 @@ export let onclose: () => void = () => {
     <CloseButton on:click={() => onclose()} />
   </div>
 
-  <div class="max-h-80 overflow-auto text-[var(--pd-modal-text)] px-10 py-4">
+  <div class="relative max-h-80 overflow-auto text-[var(--pd-modal-text)] px-10 py-4">
     <slot name="content" />
   </div>
 
--- a/packages/renderer/src/lib/image/PushImageModal.svelte
+++ b/packages/renderer/src/lib/image/PushImageModal.svelte
@@ -115,7 +115,7 @@ $: window.hasAuthconfigForImage(imageInfoToPush.name).then(result => (isAuthenti
   on:close={() => {
     closeCallback();
   }}>
-  <div slot="content" class="relative flex flex-col text-sm leading-5 space-y-5">
+  <div slot="content" class="flex flex-col text-sm leading-5 space-y-5">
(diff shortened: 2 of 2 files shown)


deboer-tim (commented):
The fix is good, but I'm assuming this fix is required b/c the Dialog has absolute positioning? I'm wondering if this is likely to happen elsewhere, and if so whether the `relative` should be 'higher up' in the Dialog itself.

SoniaSandler:
@deboer-tim should I make the content slot of the Dialog component relative or just the one in the `PushImageModal` file? Some of the places that use the Dialog component don't have scrollbar, so I don't think this problem affects them, but there are a few UI components that use the Dialog component, so I'm not sure how the relative change will affect them

deboer-tim:
> @deboer-tim should I make the content slot of the Dialog component relative or just the one in the `PushImageModal` file?

I'd suggest putting it on the content area in Dialog.svelte:22, which is where the overflow-auto is that's causing the scrolling.

> Some of the places that use the Dialog component don't have scrollbar, so I don't think this problem affects them, but there are a few UI components that use the Dialog component, so I'm not sure how the relative change will affect them

Agreed, not many dialogs will have a scrollbar, and fewer still will also use absolute positioning. But making any absolute relative to the content area is a safe and 'correct' change that would avoid the problem ever happening again.

deboer-tim (approved):

cdrage (approved):

PR #8413, opened 2024-08-09
chore: remove random divs created by Xterm terminal

### LINKED ISSUE #8197: Divs are being added every time a pod or a container log page is visited
### Bug description

Every time I go to a log page of a container or a pod, a weird div is added in the developer tools window. Also, sometimes when moving between detail pages, `context-view` and `monaco-aria-container` divs are also added.
[Screencast from 2024-07-22 22-05-42.webm](https://github.com/user-attachments/assets/76c1a7dd-eab4-413f-b3a0-7ebff32e899b)
Screenshot from 2024-07-22 23-22-09

### Operating system

Linux Fedora 39

### Installation Method

### Version

next (development version)

### Steps to reproduce

Go to any container or pod details page, and click the log tab

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
Disposes of Xterm terminals when leaving their page to remove the extra divs created by them in the DOM

Closes https://github.com/containers/podman-desktop/issues/8197

### How to test this PR?
Go to any container's or pod's log page and check that after leaving that page, the div that is added to the DOM at the end is removed.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
@@ -76,6 +76,9 @@ test('expect being able to attach terminal ', async () => {
   // search a div having aria-live="assertive" attribute
   await waitFor(() => expect(renderObject.container.querySelector('div[aria-live="assertive"]')));
   const terminalLinesLiveRegion = renderObject.container.querySelector('div[aria-live="assertive"]');
+
+  await new Promise(resolve => setTimeout(resolve, 2000));
+
   // check the content
   expect(terminalLinesLiveRegion).toHaveTextContent('hello world');
 
--- a/packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
@@ -79,7 +79,6 @@ test('expect being able to attach terminal ', async () => {
   // search a div having aria-live="assertive" attribute
   await waitFor(() => renderObject.container.querySelector('div[aria-live="assertive"]') !== undefined);
   const terminalLinesLiveRegion = renderObject.container.querySelector('div[aria-live="assertive"]');
-  console.log(terminalLinesLiveRegion);
   // check the content
   expect(terminalLinesLiveRegion).toHaveTextContent('hello world');
 
--- a/packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
@@ -26,6 +26,11 @@ import type { ContainerInfoUI } from './ContainerInfoUI';
 
 const getConfigurationValueMock = vi.fn();
 const attachContainerMock = vi.fn();
+vi.mock('xterm', () => {
+  return {
+    Terminal: vi.fn().mockReturnValue({ loadAddon: vi.fn(), open: vi.fn(), write: vi.fn(), dispose: vi.fn() }),
+  };
+});
 
 beforeAll(() => {
   (window as any).getConfigurationValue = getConfigurationValueMock;
--- a/packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
@@ -68,15 +68,18 @@ test('expect being able to attach terminal ', async () => {
   // wait attachContainerMock is called
   await waitFor(() => expect(attachContainerMock).toHaveBeenCalled());
 
+  await new Promise(resolve => setTimeout(resolve, 1000));
+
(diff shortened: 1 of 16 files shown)


deboer-tim:
I know it's more work short term, but is there any chance to pull some of these Terminal instances into a common component so that in the future we're fixing one place vs 10?

SoniaSandler:
@deboer-tim we already have `TerminalWindow` component in the renderer `ui`, so I can migrate the other files in here to use it. However, there is a problem with the `dispose()` method for `ContainerDetailsTerminal` since we save the terminal and we might reuse it. Disposing it makes it unusable a second time, but not disposing it leave its weird div in the DOM and a new one every time we visit this page

SoniaSandler:
Dispose cannot be used for `ContainerDetailsTerminal` since the terminal is reused and disposing of it makes it unusable. Even when trying to use dispose for it, there are some issues as the terminal is saved to the store.
https://github.com/xtermjs/xterm.js/tree/master/addons/addon-serialize, could help with ContainerDetailsTerminal  by serializing the terminal's content, save the content to the store instead of the terminal, dispose of the terminal, and then reuse the old content in write() . However, this addon is flagged as experimental.

packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
```diff
@@ -67,15 +68,18 @@ test('expect being able to attach terminal ', async () => {
...
 
   // wait 1s
-  await new Promise(resolve => setTimeout(resolve, 1000));
+  await new Promise(resolve => setTimeout(resolve, 2500));
 
   // search a div having aria-live="assertive" attribute
+  await waitFor(() => renderObject.container.querySelector('div[aria-live="assertive"]') !== undefined);
   const terminalLinesLiveRegion = renderObject.container.querySelector('div[aria-live="assertive"]');
-
+  console.log(terminalLinesLiveRegion);
```
benoitf: remove ?

benoitf:
it seems that this PR is about two things, one is refactoring (using TerminalWindow and one about calling dispose)

probably better to have two separate things, one with just refactoring (like should be a noop), no bugfix, just preparing things

and then we'll target dispose stuff ?

benoitf (commented):
looks like I forgot to hit the button when I did a review

packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
SoniaSandler: @benoitf this has been removed in one of the more recent commits

packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
```diff
@@ -67,15 +68,17 @@ test('expect being able to attach terminal ', async () => {
...
   await waitFor(() => expect(attachContainerMock).toHaveBeenCalled());
 
+  await new Promise(resolve => setTimeout(resolve, 1000));
+
   // write some data on the terminal
   onDataCallback(Buffer.from('hello\nworld'));
 
   // wait 1s
-  await new Promise(resolve => setTimeout(resolve, 1000));
+  await new Promise(resolve => setTimeout(resolve, 2500));
```
benoitf: need to change the comment on the line before as it's no longer 1s

also do we still need the await timeout if we're waiting for the line 80

packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
```diff
@@ -67,15 +68,17 @@ test('expect being able to attach terminal ', async () => {
...
 
   // search a div having aria-live="assertive" attribute
+  await waitFor(() => renderObject.container.querySelector('div[aria-live="assertive"]') !== undefined);
```
benoitf: shouldn't we use an expect assertion rather than !== undefined

packages/renderer/src/lib/container/ContainerDetailsTtyTerminal.spec.ts
SoniaSandler: @benoitf I tried removing the await timeout, but the unit test failed in the PR checks, so I just readded it after the the waitFor

benoitf: but then it seems we're expecting a content

so what if we're just waiting to have the right text content ?
```
 await new Promise(resolve => setTimeout(resolve, 2000));

  // check the content
  expect(terminalLinesLiveRegion).toHaveTextContent('hello world');
```

==>

```
await vi.waitFor(() => expect(terminalLinesLiveRegion).toHaveTextContent('hello world'));
```

and as default timeout for waitFor is 1s, 

=> 

```
await vi.waitFor(() => expect(terminalLinesLiveRegion).toHaveTextContent('hello world'), {timeout: 2500});
```

benoitf (approved):

axel7083 (approved):
LGTM 🚀 nice fix

PR #8432, opened 2024-08-12
chore: update electron-updater to v6.3.3

benoitf:
### What does this PR do?
update to v6.3.3
fixes some security fixes reported in previous electron-updater

### How to test this PR?

modify package.json to be an old version (like 1.0.0) and build the binary
then launch the binary and check you're able to update

I've checked on macOS


benoitf:
rebased to fix conflict on the yarn.lock file

PR #8437, opened 2024-08-13
chore: adds missing colors for modal dropdown element

gastoner:
### What does this PR do?
Adds missing colors for modal dropdown element

Reference:
https://github.com/user-attachments/files/16070597/Modal.Dialog.pdf

First of two PRs for 8411 "Light Mode: Input boxes (changing kube-context) still using dark theme when in Light Mode"
Link to second one: https://github.com/containers/podman-desktop/pull/8549

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -1090,6 +1090,8 @@ export class ColorRegistry {
   protected initDropdown(): void {
     const dropdown = 'dropdown-';
     const select = 'select-';
+    const modal = 'modal-';
+    const input = 'input-';
 
     this.registerColor(`${dropdown}bg`, {
       dark: colorPalette.charcoal[600],
@@ -1132,6 +1134,19 @@ export class ColorRegistry {
     this.registerColor(`${dropdown}disabled-item-bg`, {
       dark: colorPalette.charcoal[800],
       light: colorPalette.gray[200],
+    });
+
+    this.registerColor(`${modal}${dropdown}highlight`, {
+      dark: colorPalette.purple[600],
+      light: colorPalette.purple[300],
+    });
+    this.registerColor(`${modal}${dropdown}text`, {
+      dark: colorPalette.white,
+      light: colorPalette.charcoal[900],
+    });
+    this.registerColor(`${input}${select}hover-text`, {
+      dark: colorPalette.gray[900],
+      light: colorPalette.charcoal[200],
     });
   }
 


axel7083 (approved):
LGTM

deboer-tim (changes requested):
We have dropdown (select controls) in many places - preferences, deploying pods to kube, etc.

We really need a consistent component (#6199) and to reuse it in most of these places, and reuse the colors where we can't. I'm good with fixing this as a 'one off', but I think we should define the colors as we would use them in a component - i.e. separately in an initDropdown() vs colors specific to a dropdown in a modal.

gastoner:
@deboer-tim So, do I get it right that I should move it to initDropdown? Or should I use existing colors, which are already there?

deboer-tim:
> @deboer-tim So, do I get it right that I should move it to initDropdown? Or should I use existing colors, which are already there?

Sorry, I didn't realize there was an existing initDropdown() - but yes, you should be able to use colors from there (we shouldn't have different colors for dropdown in modal vs dropdown elsewhere), or edit/add to that function.

ekidneyrh:
I might be looking at the wrong place but I cant see the changes. Would I be able to see a screenshot? Or is just change in the code rather than a visual change?

gastoner:
> I might be looking at the wrong place but I cant see the changes. Would I be able to see a screenshot? Or is just change in the code rather than a visual change?

@ekidneyrh So far it is just change in the code, I would add the visual changes in the next PR. 
Issue ref. https://github.com/containers/podman-desktop/issues/8411
Colors: https://github.com/user-attachments/files/16070597/Modal.Dialog.pdf

gastoner:
@deboer-tim PR updated!

ekidneyrh (approved):
LGTM :) Thanks Evzen!

deboer-tim (commented):
Just back from PTO. Thanks for making the changes, but it still looks to me like the code is implying that dropdowns in modals are going to be different from other dropdowns. Is this actually the case (why?) or could we keep the prefixes consistent with just `${dropdown}`?

gastoner:
> Just back from PTO. Thanks for making the changes, but it still looks to me like the code is implying that dropdowns in modals are going to be different from other dropdowns. Is this actually the case (why?) or could we keep the prefixes consistent with just `${dropdown}`?

@deboer-tim 
Honestly, I don't know why it is called the dropdown in the first place, It seems to me like a totally different component with different behavior (e.g. can be indexed by arrows, and have input to filer-out some contexts) - the component is called `QuickPickInput`. But yes it shows the options in the dropdown.

It is using naming schema and colors from https://github.com/user-attachments/files/16708192/Modal.Dialog.pdf

deboer-tim (approved):

cdrage (approved):

PR #8446, opened 2024-08-13
refactor: add aria-labels for "Deploy to Kubernetes" page

### LINKED ISSUE #8445: Add aria-labels to 'Deploy to Kubernetes' page for improved accessibility

We need to add aria-labels to the "Deploy to Kubernetes" page because many elements lack them, making navigation and element location challenging.

- Add aria-labels to the relevant page elements.

_No response_

### Additional context

https://github.com/containers/podman-desktop/issues/5976

amisskii:

### What does this PR do?

-     Add aria-labels to the relevant page elements.

https://github.com/containers/podman-desktop/issues/8445

### DIFF
--- a/packages/renderer/src/lib/pod/DeployPodToKube.svelte
+++ b/packages/renderer/src/lib/pod/DeployPodToKube.svelte
@@ -461,7 +461,7 @@ function updateKubeResult() {
           name="serviceName"
           id="serviceName"
           class=" cursor-default w-full p-2 outline-none text-sm bg-[var(--pd-select-bg)] rounded-sm text-[var(--pd-content-card-text)]"
-          aria-label="Select a Pod"
+          aria-label="Select a Port"
           required>
           <option value="" disabled selected>Select a port</option>
           {#each containerPortArray as port}
--- a/packages/renderer/src/lib/pod/DeployPodToKube.svelte
+++ b/packages/renderer/src/lib/pod/DeployPodToKube.svelte
@@ -393,7 +393,13 @@ function updateKubeResult() {
       <div class="pt-2 pb-4">
         <label for="contextToUse" class="block mb-1 text-sm font-medium text-[var(--pd-content-card-header-text)]"
           >Pod Name:</label>
-        <Input bind:value={bodyPod.metadata.name} name="podName" id="podName" class="w-full" required />
+        <Input
+          bind:value={bodyPod.metadata.name}
+          name="podName"
+          aria-label="Pod Name"
+          id="podName"
+          class="w-full"
+          required />
       </div>
     {/if}
 
@@ -405,6 +411,7 @@ function updateKubeResult() {
         class="text-[var(--pd-content-card-text)] text-sm ml-1"
         name="useServices"
         id="useServices"
+        title="Use Services"
         required>
         Replace .hostPort exposure on containers by Services. It is the recommended way to expose ports, as a cluster
         policy may prevent to use hostPort.</Checkbox>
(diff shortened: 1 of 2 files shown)


tests/playwright/src/model/pages/deploy-to-kubernetes-page.ts
```diff
@@ -0,0 +1,35 @@
...
+  readonly deployButton: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.content = this.page.getByRole('region', { name: 'Tab Content' });
+    this.podName = this.content.getByRole('textbox', { name: 'Pod Name' });
+    this.kubernetesContext = this.content.getByRole('textbox', { name: 'Kubernetes Context' });
+    this.deployButton = this.content.getByRole('button', { name: 'Deploy' });
+  }
+}
```
cbr7: Maybe I don't understand how it's used, but to me this looks like it's not used anywhere, why are we creating it? If the class is not used it shouldn't exist.

amisskii: You’re right! I’m working on [Issue #8053](https://github.com/containers/podman-desktop/issues/8053), which involves deploying a pod/container to a Kind cluster. I will need this class soon, so I was considering including it in the current refactoring of this page, along with locators that use the recent changes. However, it might be better to include these changes in a PR specifically for the deploying to Kind cluster scenario.

cbr7:
@amisskii Did you run the e2e test on windows with this branch to ensure that there is no regression from the name changes?

cbr7 (approved):

amisskii:
@cbr7 There’s still an issue with the Windows workflows :( I’ll wait until it’s resolved and then test it.

cbr7:
oh :(

odockal:
@amisskii nice work. I am quite surprised that not unit test failed during the change. I assume, that they were missing totally, or not relying on the label as is usual. Would be nice to take a look.

packages/renderer/src/lib/pod/DeployPodToKube.svelte
```diff
@@ -393,7 +393,7 @@ function updateKubeResult() {
       <div class="pt-2 pb-4">
         <label for="contextToUse" class="block mb-1 text-sm font-medium text-[var(--pd-content-card-header-text)]"
           >Pod Name:</label>
-        <Input bind:value={bodyPod.metadata.name} name="podName" id="podName" class="w-full" required />
+        <Input bind:value={bodyPod.metadata.name} aria-label="Pod Name" id="podName" class="w-full" required />
```
odockal: do not remove attribute `name`.

odockal (changes requested):
Between Create and Ingress Checkbox and Use Routes there is a select element `Select a Port` that would also make use of aria-label.
Similarly for Kubernetes Namespace select.
Also for DeployWarning and DeployError elements.

Check how the Link components are using aria-labels.

In order to get these elements visible, you would need valid kube-context for either kind or ocp (devsandbox) -> openshift routes vs. ingress routes and a container to deploy with exposed port (ie. httpd-24).

odockal (approved):

packages/renderer/src/lib/pod/DeployPodToKube.svelte
```diff
@@ -454,6 +461,7 @@ function updateKubeResult() {
           name="serviceName"
           id="serviceName"
           class=" cursor-default w-full p-2 outline-none text-sm bg-[var(--pd-select-bg)] rounded-sm text-[var(--pd-content-card-text)]"
+          aria-label="Select a Pod"
```
odockal: @amisskii typo: `Select a Port` is correct.

odockal (changes requested):
LGTM: just one type to fix.

odockal (approved):

PR #8456, opened 2024-08-13
fix: update webview urls on change

### LINKED ISSUE #8291: Extension navigation broken after item inserted or moved due to install/remove/activate/deactivate an extension
### Bug description

The navigation bar links are not updated correctly when navigation items for extensions move around in navbar due to update after installing/removing or activating/deactivating extension. This leads to navigation to wrong extension webview.

https://github.com/user-attachments/assets/19dcacb7-c4e5-4091-a60d-e5da0a74edac

### Operating system

macOS M1 Sonoma 14

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

Install BootC from catalog page and then install OpenShift Docker Desktop Extension from docker hub using `redhatdeveloper/openshift-dd-ext`. Clicking on OpenShift Item will open BootC extension webview.

### Relevant log output

```shell
n/a
```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

NavItems were never built to be reactive, i.e. have the href or title change after they are created. However, the new navigation registry causes exactly that: when the list of webviews change any existing items are getting updated instead of replaced and so any remaining ones have old uris. This change just makes the uri property reactive to match the others.

### Screenshot / video of UI

N/A

Fixes #8291.

### How to test this PR?

Add/remove/hide/unhide with multiple webviews and confirm you can still open each of them.

### DIFF
--- a/packages/renderer/src/lib/ui/NavItem.svelte
+++ b/packages/renderer/src/lib/ui/NavItem.svelte
@@ -16,6 +16,7 @@ export let onClick: MouseEventHandler<HTMLAnchorElement> | undefined = undefined
 export let counter: number | undefined = undefined;
 
 let inSection: boolean = false;
+let uri: string;
 $: uri = encodeURI(href);
 let selected: boolean;
 $: selected = meta.url === uri || (uri !== '/' && meta.url.startsWith(uri));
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/ui/NavItem.svelte
```diff
@@ -16,7 +16,7 @@ export let onClick: MouseEventHandler<HTMLAnchorElement> | undefined = undefined
 export let counter: number | undefined = undefined;
 
 let inSection: boolean = false;
-let uri = encodeURI(href);
+$: uri = encodeURI(href);
```
axel7083, line 20:
```suggestion
let uri: string;
$: uri = encodeURI(href);
```

benoitf (approved):
and +1 to @axel7083 comments

I don't know if we can enforce the definition of the variables/typing with an eslint rule

axel7083 (approved):

PR #8485, opened 2024-08-15
fix: update native theme directly on config change

### LINKED ISSUE #8477: No light mode when Podman Desktop set to "system" and system is light.
### Bug description

Light mode isn't shown when PD is on "system theme" and system is on light mode.

### Operating system

mac m1

### Version

next (development version)

### Steps to reproduce

1. Set system to light mode
2. Set PD to "system"
3. bootc wont be light mode

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

PR #8289 set the native theme so that extensions render correctly, but once we set the native theme, window.matchMedia('(prefers-color-scheme: dark)') just returns what we set the native theme to - i.e. if system is in light mode and you explicitly set our config value to dark mode and then to system, we will ask the system what it prefers and it still says 'dark', because that's what we told it last. This leads to some parts of the UI updating correctly while other parts don't; restarting fixes the issue.

The fix is to set the native theme directly (and immediately) based on the value of our configuration setting. That way when we call isDarkMode() later on and we're in system mode we will get the correct value from prefers-color-scheme.

Since we don't need nativeTheme exposed to renderer anymore I'm removing it.

FWIW I also tried updating native theme on config change in Appearance's onDidChangeConfigurationCallback. It works fine, but overall this felt simpler and more direct.

### Screenshot / video of UI

N/A

Fixes #8477.

### How to test this PR?

Install at least one Docker Desktop extension and one extension with a webview (e.g. AI Lab or bootc), then:
- Set the appearance to dark mode, verify that Podman Desktop and the extensions change.
- Set the appearance to system, verify that Podman Desktop and the extensions change.
- Set the appearance to light mode, verify that Podman Desktop and the extensions change.
- Set the appearance to system, verify that Podman Desktop and the extensions change.

### DIFF
--- a/packages/main/src/plugin/appearance-init.ts
+++ b/packages/main/src/plugin/appearance-init.ts
@@ -21,6 +21,8 @@ import { nativeTheme } from 'electron';
 import { AppearanceSettings } from './appearance-settings.js';
 import type { IConfigurationNode, IConfigurationRegistry } from './configuration-registry.js';
 
+const APPEARANCE_FULL_KEY = `${AppearanceSettings.SectionName}.${AppearanceSettings.Appearance}`;
+
 export class AppearanceInit {
   constructor(private configurationRegistry: IConfigurationRegistry) {}
 
@@ -30,7 +32,7 @@ export class AppearanceInit {
       title: 'Appearance',
       type: 'object',
       properties: {
-        [AppearanceSettings.SectionName + '.' + AppearanceSettings.Appearance]: {
+        [APPEARANCE_FULL_KEY]: {
           description: 'Select between light or dark mode, or use your system setting.',
           type: 'string',
           enum: ['system', 'dark', 'light'],
@@ -42,7 +44,7 @@ export class AppearanceInit {
     this.configurationRegistry.registerConfigurations([appearanceConfiguration]);
 
     this.configurationRegistry.onDidChangeConfiguration(async e => {
-      if (e.key === AppearanceSettings.SectionName + '.' + AppearanceSettings.Appearance) {
+      if (e.key === APPEARANCE_FULL_KEY) {
         this.updateNativeTheme(e.value);
       }
     });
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/appearance-init.ts
```diff
@@ -38,5 +40,22 @@ export class AppearanceInit {
     };
 
     this.configurationRegistry.registerConfigurations([appearanceConfiguration]);
+
+    this.configurationRegistry.onDidChangeConfiguration(async e => {
+      if (e.key === AppearanceSettings.SectionName + '.' + AppearanceSettings.Appearance) {
```
benoitf:
```suggestion
      if (e.key === `{AppearanceSettings.SectionName}.${AppearanceSettings.Appearance}`) {
```
also maybe compute the key once as a constant instead of inside the if

```ts
const APPEARANCE_FULL_KEY  = `{AppearanceSettings.SectionName}.${AppearanceSettings.Appearance}`
...
...
...

 if (e.key === APPEARANCE_FULL_KEY) {
...
```

benoitf (dismissed):

gastoner:
Looks like when selected "system" I get white PD and dark AI Lab
WSL (Unix):

https://github.com/user-attachments/assets/5c831f4f-754c-462d-976d-4d90917c169d

packages/main/src/plugin/appearance-init.ts
deboer-tim: Forgot to push a commit before PTO - pushed now. Uses a constant for both instances.

axel7083:
Works fine with podman-desktop, but ai lab does not seems to update :/

https://github.com/user-attachments/assets/be7a2ad5-4d90-4777-92e3-434cf9cc5a77

deboer-tim:
Rebased to pick up pnpm / improved typechecking.

axel7083 (commented):
okey on windows it is working fine after rebase

https://github.com/user-attachments/assets/32848a42-30e6-4270-8af2-d6ab38386532

axel7083 (approved):
Tested again on Fedora 40 and on Windows 11 and seems to be working fine ! LGTM 🚀

deboer-tim:
@gastoner can you try again (or someone else, since it looks like the WSL issue is fixed)?

gastoner (approved):

PR #8495, opened 2024-08-16
fix: storybook svelte 5

axel7083:
### What does this PR do?

Thanks to the investigation in https://github.com/storybookjs/addon-svelte-csf/pull/195 we noticed that the problem was related to `svelte-preprocess`.

Removing the `svelte-preprocess` solve the issue, however this package is used to configure the typescript language of the svelte component (lang="ts").

Looking at the documentation of `vitejs`, and particularly `vite-plugin-svelte` they currently have their own preprocess, which is the one used for `.ts` files.

Looking more in depth at the documentation lead to the experimental support of `postcss` in  `vite-plugin-svelte` , which were the reason why we were using `svelte-preprocess`.

### How to test this PR?

- Run `yarn run website:prod` and go to `localhost:3000`


benoitf (approved):
nice investigation

PR #8498, opened 2024-08-16
chore: use Xterm serialization to save container terminal content

### LINKED ISSUE #8475: Terminal divs stay after leaving page but can't use dispose for ContainerDetailsTerminal since the terminal is reused
### Bug description

In https://github.com/containers/podman-desktop/issues/8197, in every page that has Xterm terminal, the divs are staying in the DOM even after leaving the page, and every new visit adds a new div, possibly leading to multipe divs.
https://github.com/containers/podman-desktop/pull/8413 solves the problem for most pages, by adding dispose to the `TerminalWindow` UI component and changing the pages that create a new Xterm teminal to use that component. However, this solution doesn't work for `ContainerDetailsTerminal` since the terminal is saved and reused, and disposing of it makes it unusable a second time.
A possible solution I've found and need to check is using  https://github.com/xtermjs/xterm.js/tree/master/addons/addon-serialize, which might help with ContainerDetailsTerminal  by serializing the terminal's content, save the content to the store instead of the terminal, dispose of the terminal, and then reuse the old content in write(). However, this addon is labeled as experimental. 

### Operating system

Linux Fedora 39

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
Uses Xterm's serialization addon to save terminal's content. By saving the terminal's content, we can safely dispose of the terminal when leaving the `ContainerDetailsTerminal` page and recreate a new one with the old content when visiting again.

Closes https://github.com/containers/podman-desktop/issues/8475

### How to test this PR?

### DIFF
--- a/packages/renderer/package.json
+++ b/packages/renderer/package.json
@@ -25,6 +25,7 @@
     "@types/js-yaml": "^4.0.9",
     "@types/validator": "^13.12.0",
     "@typescript-eslint/eslint-plugin": "7.18.0",
+    "@xterm/addon-serialize": "^0.13.0",
     "autoprefixer": "^10.4.20",
     "eslint-plugin-svelte": "^2.43.0",
     "filesize": "^10.1.4",
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
@@ -67,7 +67,7 @@ test('expect being able to reconnect ', async () => {
   );
 
   // render the component with a terminal
-  const renderObject = render(ContainerDetailsTerminal, { container, screenReaderMode: true });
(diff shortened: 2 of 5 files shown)


benoitf (commented):
it seems there is one case that is not working

run a httpd container and go to the terminal tab

then click on the stop button
and then click on the start button

I'm no longer able to enter data

(but it looks like it's also failing on the current main branch as well)

SoniaSandler:
@benoitf for the current main branch, there is this issue about terminal not working after stopping and starting, https://github.com/containers/podman-desktop/issues/8363. I'll check how I can solve it, and maybe it will also close the mentioned issue

benoitf (approved):

lstocchi (approved):
LGTM

PR #8501, opened 2024-08-19
chore: refactor how to propagate configuration registry

benoitf:
### What does this PR do?
propagate configuration registry as soon as we have it initialized.

when implementing https://github.com/containers/podman-desktop/issues/6115 there was a delay where the window stayed for a while before being moved. It's because as we were receiving the configuration registry at the end of the initialization, the window was already displayed (and then moved)
By receiving it earlier, we don't notice this delay/move

### How to test this PR?


lstocchi (approved):
LGTM

PR #8511, opened 2024-08-19
feat: remember window position and size of Podman Desktop window

### LINKED ISSUE #6115: Remember window state and position from last start

benoitf:
### What does this PR do?
remember window position and size of Podman Desktop window

in case we can't restore, use the initial position (centered on the current screen)

fixes https://github.com/containers/podman-desktop/issues/6115

### How to test this PR?

### DIFF
--- a/packages/main/src/mainWindow.ts
+++ b/packages/main/src/mainWindow.ts
@@ -19,7 +19,7 @@
 import { join } from 'node:path';
 import { URL } from 'node:url';
 
-import type { BrowserWindowConstructorOptions } from 'electron';
+import type { BrowserWindowConstructorOptions, Rectangle } from 'electron';
 import { app, autoUpdater, BrowserWindow, ipcMain, Menu, nativeTheme, screen } from 'electron';
 import contextMenu from 'electron-context-menu';
 import { aboutMenuItem } from 'electron-util/main';
@@ -27,6 +27,7 @@ import { aboutMenuItem } from 'electron-util/main';
 import { NavigationItemsMenuBuilder } from './navigation-items-menu-builder.js';
 import { OpenDevTools } from './open-dev-tools.js';
 import type { ConfigurationRegistry } from './plugin/configuration-registry.js';
+import type { WindowHandler } from './system/window/window-handler.js';
 import { isLinux, isMac, stoppedExtensions } from './util.js';
 
 const openDevTools = new OpenDevTools();
@@ -72,18 +73,20 @@ async function createWindow(): Promise<BrowserWindow> {
   }
 
   const browserWindow = new BrowserWindow(browserWindowConstructorOptions);
+
   const { getCursorScreenPoint, getDisplayNearestPoint } = screen;
   const workArea = getDisplayNearestPoint(getCursorScreenPoint()).workArea;
 
   const x = Math.round(workArea.width / 2 - INITIAL_APP_WIDTH / 2 + workArea.x);
   const y = Math.round(workArea.height / 2 - INITIAL_APP_HEIGHT / 2);
 
-  browserWindow.setBounds({
+  const initialBounds: Rectangle = {
     x: x,
     y: y,
-    width: browserWindowConstructorOptions.width,
-    height: browserWindowConstructorOptions.height,
-  });
+    width: INITIAL_APP_WIDTH,
+    height: INITIAL_APP_HEIGHT,
+  };
+  browserWindow.setBounds(initialBounds);
 
   /**
    * If you install `show: true` then it can cause issues when trying to close the window.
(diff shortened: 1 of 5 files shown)


feloy:
Code LGTM

It works for me on Mac, except for a specific case: I have 2 screens; the position and size are saved on the first screen. When I move the app to the second screen, the size is reset to the default one. When I move the app back to the first screen, the size if set again to the saved one.

packages/main/src/mainWindow.ts
```diff
@@ -72,18 +73,20 @@ async function createWindow(): Promise<BrowserWindow> {
...
   const y = Math.round(workArea.height / 2 - INITIAL_APP_HEIGHT / 2);
 
-  browserWindow.setBounds({
+  const initialBounds: Rectangle = {
     x: x,
     y: y,
-    width: browserWindowConstructorOptions.width,
-    height: browserWindowConstructorOptions.height,
-  });
+    width: INITIAL_APP_WIDTH,
```
jeffmaury, line 86: Not sure to understand this change

packages/main/src/system/window/window-handler.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  #debounceSaveTimeout: NodeJS.Timeout | undefined;
+
+  constructor(configurationRegistry: ConfigurationRegistry, browserWindow: BrowserWindow) {
+    this.#browserWindow = browserWindow;
+    this.#configurationRegistry = configurationRegistry;
+  }
+
+  // try to restore the window position and size
+  restore(initalBounds: Rectangle): void {
```
jeffmaury:
```suggestion
  restore(initialBounds: Rectangle): void {
```

jeffmaury (approved):
A single remark: Podman Desktop was on the second screen when I stopped it, I disconnected the second screen when I restarted it it went back with width and height defaulted maybe we can try to restore them if it fits

packages/main/src/mainWindow.ts
benoitf: it's a shortcut as we were giving INITIAL_APP_WIDTH to browserWindowConstructorOptions but I can withdraw the change as at some point the code was elsewhere but now it went back there so I have access to browserWindowConstructorOptions

benoitf:
> It works for me on Mac, except for a specific case: I have 2 screens; the position and size are saved on the first screen. When I move the app to the second screen, the size is reset to the default one. When I move the app back to the first screen, the size if set again to the saved one.

@feloy it's when you move and exit and restart the app ? or just moving the window is doing the resize ?

feloy:
> > It works for me on Mac, except for a specific case: I have 2 screens; the position and size are saved on the first screen. When I move the app to the second screen, the size is reset to the default one. When I move the app back to the first screen, the size if set again to the saved one.
> 
> @feloy it's when you move and exit and restart the app ? or just moving the window is doing the resize ?

Just moving the window from one screen to another changes its side.

I'm not sure of the initial conditions to be in this state, I was not able to reproduce a second time

benoitf:
@feloy oh that strange because we trigger the restore (and changing the bounds) only at the startup

packages/main/src/mainWindow.ts
benoitf: ok now I recall, it's because browserWindowConstructorOptions.width return a number or undefined and then we don't want to have an undefined value

feloy:
> @feloy oh that strange because we trigger the restore (and changing the bounds) only at the startup

I have added a log when the `browserWindow`'s `resize` and `move` events are triggered, and a log in the `doSavePositionAndSize` function to log the bounds:
```
   browserWindow.on('resize', () => {
+    console.log('==> resize');
     windowHandler?.savePositionAndSize();
   });
 
   browserWindow.on('move', () => {
+    console.log('==> move');
     windowHandler?.savePositionAndSize();
   });
```

```
  protected async doSavePositionAndSize(): Promise<void> {
    const bounds = this.#browserWindow.getBounds();
    const windowPreferences = this.#configurationRegistry.getConfiguration(WindowSettings.SectionName);
+   console.log('==> prefs', windowPreferences.get(WindowSettings.Bounds));
    await windowPreferences?.update(WindowSettings.Bounds, bounds);
  }
``` 

Here are the logs when I move the window from one screen to another (the resize event happens when I drop the window after the drag):

two_screens_restore

benoitf:
@feloy we do save to the configuration file the position when you're moving or changing the position of the window so AFAIK it's a normal behavior
and when we start the app we use the saved data

not sure to see what you're highlighting ?

benoitf:
@jeffmaury new commit in the PR will try to restore the persisted width/height if it fits even if the screen is no longer available

packages/main/src/system/window/window-handler.ts
benoitf: fixed 👍

feloy:
> @feloy we do save to the configuration file the position when you're moving or changing the position of the window so AFAIK it's a normal behavior and when we start the app we use the saved data
> 
> not sure to see what you're highlighting ?

That a `resize` event is triggered when the browser window is dropped in another screen. In the logs below, I also display the bounds received with a `will-resize` event, which are different when I drop the window in screen 1 or screen 2:


will-resize

benoitf:
@feloy PR amended to use setPosition/setSize rather than setBounds

feloy (approved):
LGTM
Now works when moving the window across different screens. 

Also restored with good size when a screen is removed

benoitf:
thanks @feloy for the help

Issue #8003, opened 2024-07-09
CLI Tools should support installing the binary


axel7083:
### Is your feature request related to a problem? Please describe

![image](https://github.com/containers/podman-desktop/assets/42176370/deda1158-0602-401b-abae-ac1b771d9b36)

Currently we still register CLITool using the extension API 

https://github.com/containers/podman-desktop/blob/4f9e876667fe5ab48d26bd9d2d1ab728e3a89dd5/packages/extension-api/src/extension-api.d.ts#L4350

> ℹ️ I first wanna say I do not like the wording `create`, we should be using some kind of `register`, as we do not create anything.

When ~creating~ a CLI tool through the extension API we are getting a `CLITool` object

https://github.com/containers/podman-desktop/blob/4f9e876667fe5ab48d26bd9d2d1ab728e3a89dd5/packages/extension-api/src/extension-api.d.ts#L4319-L4336

On this object we can call `registerUpdate` to register an update, the following will allow a visual indication to the user, that an update is possible

![image](https://github.com/containers/podman-desktop/assets/42176370/2686cdd1-c7d4-4600-85ac-c3ea69cb0045)

Something similar should be possible to trigger the installation of the CLI Tool.

### Describe the solution you'd like

My proposal would be to change the `cli.createCliTool` signature. The CliToolOptions are not flexible enough.

First let's combine the `version` and `path` to a single **optional** property (proposal: binary, open to suggestion)

```diff
export interface CliToolOptions {
  name: string;
  displayName: string;
  markdownDescription: string;
  images: ProviderImages;

- version: string;
- path: string;
+  binary?: {
+   version: string;
+   path: string;
+  }
}
```

To avoid making too much changes to the API, we could add a new method `registerInstall` next to the `registerUpdate` in the `CliTool` interface.

axel7083:
I made a quick branch to demo how it could be done with the change proposal above https://github.com/axel7083/podman-desktop/tree/feature/cli-install-workflow

⚠️ it does not work like it, I changed the code locally of compose and kubectl to use the new api 

https://github.com/containers/podman-desktop/assets/42176370/7ccc27ac-d966-460b-b7da-4643e94daa86

cc @deboer-tim

deboer-tim:
I had a long response typed up, but I'm going to try shorter:
- If we need breaking changes elsewhere then `binary` is good, but I'm wondering if it's easier to just make those two properties optional.
[...]

PR #8513, opened 2024-08-19
feat: allow to upgrade/downgrade cli tool

lstocchi:
### What does this PR do?

This PR is just a part of https://github.com/containers/podman-desktop/issues/8003 
It enhances the update process to allow upgrading/downgrading to a specific version.
If the cli tool has been installed by the user (it is in a path different from the one used by desktop), the update is not allowed.

The PR is splitted in 2 commits: the first is about the core, the second is about the extensions. I can also split them in 2 PRs if you prefer

### Screenshot / video of UI

upgrade_downgrade_cli

it is part of #8003 

### How to test this PR?

1. try to upgrade/downgrade compose or kubectl

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4300,6 +4300,8 @@ declare module '@podman-desktop/api' {
     export function setValue(key: string, value: any, scope?: 'onboarding'): void;
   }
 
+  export type CliToolInstallationSource = 'userInstalled' | 'appInstalled';
+
   /**
    * Options to create new CliTool instance and register it in podman desktop
    */
@@ -4319,6 +4321,13 @@ declare module '@podman-desktop/api' {
      */
     version: string;
     path: string;
+
+    /**
+     * How the cli tool has been installed
+     * - user: it has been installed by the user externally from podman desktop. Its update process is disable.
+     * - app: it has been installed through podman desktop. It can be updated
+     */
+    installationSource?: CliToolInstallationSource;
   }
 
   /**
@@ -4330,10 +4339,16 @@ declare module '@podman-desktop/api' {
     markdownDescription?: string;
     images?: ProviderImages;
     path?: string;
+    installationSource?: CliToolInstallationSource;
   }
 
   export interface CliToolUpdate {
     version: string;
+    doUpdate: (logger: Logger) => Promise<void>;
+  }
+
+  export interface CliToolSelectUpdate {
+    selectVersion: () => Promise<string>;
     doUpdate: (logger: Logger) => Promise<void>;
   }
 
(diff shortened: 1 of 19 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -4319,6 +4321,13 @@ declare module '@podman-desktop/api' {
      */
     version: string;
     path: string;
+
+    /**
+     * How the cli tool has been installed
+     * - user: it has been installed by the user externally from podman desktop. Its update process is disable.
+     * - app: it has been installed through podman desktop. It can be updated
+     */
+    installationSource?: CliToolInstallationSource;
```
benoitf: hello, I'm not sure here to follow.

>
> * How the cli tool has been installed
>     * - user: it has been installed by the user externally from podman desktop. Its update process is disable.
>     * - app: it has been installed through podman desktop. It can be updated

it seems it's a flag provided by the extension, so how does it work ? an extension can tell it has been installed by Podman Desktop while it has been installed by the user

maybe just the wording. Rather than 'podman desktop' it seems it's the 'extension'

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4300,6 +4300,8 @@ declare module '@podman-desktop/api' {
     export function setValue(key: string, value: any, scope?: 'onboarding'): void;
   }
 
+  export type CliToolInstallationSource = 'userInstalled' | 'appInstalled';
```
benoitf: not sure to repeat the word 'installed' as the property name is installation source

is it like something more simple 'extension' or 'external' (as it may not be the user but the system or something else)

```typescript
 export type CliToolInstallationSource = 'extension' | 'external'
```

packages/extension-api/src/extension-api.d.ts
lstocchi: I added it bc i thought that an extension could download the binary in a specific place and podman desktop could not know it. So an extension is responsible to tell if it has been installed by them or not. If not specified, the default value is user/external so it cannot be updated.

lstocchi: if that's sound ok to you, it's just a wording issue, yes

benoitf: yes wording, as it's managed by the extension, not Podman Desktop.

packages/extension-api/src/extension-api.d.ts
lstocchi: updated

packages/extension-api/src/extension-api.d.ts
lstocchi: updated

benoitf (approved):

lstocchi:
Merging so i can progress. There will be other PRs related to the same topic so for any change we can do some followups

PR #8517, opened 2024-08-20
chore: upgrade kubernetes client library

### LINKED ISSUE #8516: update kubernetes library

today, the library is relying on `request` library that is no longer maintained and there are some `Moderate` issues related

Use the newer version, 1.0.0 (being planned) that is using node-fetch

_No response_

### Additional context

there are breaking changes in the library

benoitf:
### What does this PR do?
the current version that we are using is relying on https://www.npmjs.com/package/request that is deprecated
we've also some reports from dependabot with moderate CVE
https://github.com/request/request/issues/3442

the kubernetes client has a new version no longer relying on this package and using node-fetch

but due to the switch and the generator switch, a lot of requests have their parameters being changed so it's a deep change

I think it's easier now as instead of like sometimes 8 parameters, we only provide a complex type with only the fields we want to submit and it's avoiding the current pattern `(param1, undefined, undefined, undefined, param2)` being used.

https://github.com/kubernetes-client/javascript/issues/1828#issuecomment-2265568631

fixes https://github.com/containers/podman-desktop/issues/8516

### How to test this PR?

Test all kubernetes features

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
   },
   "dependencies": {
     "@docker/extension-api-client-types": "0.3.4",
-    "@kubernetes/client-node": "^0.21.0",
+    "@kubernetes/client-node": "^1.0.0-rc6",
     "@segment/analytics-node": "^2.1.2",
     "@types/semver": "^7.5.8",
     "@types/stream-json": "^1.7.7",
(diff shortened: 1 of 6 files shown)


lstocchi (approved):
Code LGTM, i tested it with a kind cluster and everything seems working as before 👍

feloy (approved):
Code LGTM, tested on mac with kind, works as before

benoitf:
ok merging, we should be able to detect troubles before the next release if any

PR #8524, opened 2024-08-20
Allow to set Zero value on numeric preference

### LINKED ISSUE #8523: Zero value not saved in numeric Preference
### Bug description

When an extension defines a numeric preference, for which the zero (0) value is possible, this value is not saved correctly: if the user sets the value as 0, the value is changed to the default value of the preference.

https://github.com/user-attachments/assets/18b7d34d-edcb-4ea6-8850-27764f73b078

### Operating system

mac

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Allow to set Zero value on numeric preference

Fixes #8523 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/preferences/item-formats/NumberItem.spec.ts
+++ b/packages/renderer/src/lib/preferences/item-formats/NumberItem.spec.ts
@@ -20,10 +20,14 @@ import '@testing-library/jest-dom/vitest';
 
 import { render, screen } from '@testing-library/svelte';
 import userEvent from '@testing-library/user-event';
-import { expect, test } from 'vitest';
+import { afterEach, expect, test, vi } from 'vitest';
 
 import type { IConfigurationPropertyRecordedSchema } from '../../../../../main/src/plugin/configuration-registry';
 import NumberItem from './NumberItem.svelte';
+
+afterEach(() => {
+  vi.useRealTimers();
+});
 
 test('Expect tooltip if value input is invalid', async () => {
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/preferences/item-formats/NumberItem.spec.ts
```diff
@@ -126,3 +130,25 @@ test('Expect decrement button only works one if minimum value is reached after o
...
+    maximum: 34,
+  };
+  const onChange = vi.fn();
+  const value = 1;
+  render(NumberItem, { record, value, onChange });
+
+  const input = screen.getByLabelText('decrement');
+  expect(input).toBeInTheDocument();
+  await userEvent.click(input);
+  vi.advanceTimersByTime(510);
```
benoitf, line 152: should it use timers or just ticks ?

feloy: I think that we need to rely on timers, as the Svelte component is not refreshed, but only a call to onChange is called after 500ms:

```
    valueUpdateTimeout = setTimeout(() => {
      onChange(recordId, newValue);
      lastValue = newValue;
    }, 500);
```

benoitf: understood

benoitf (approved):

lstocchi (approved):
LGTM

PR #8528, opened 2024-08-21
test: E2E test case for the creation of the podman machine with user mode enabled

danivilla9:
### What does this PR do?
This PR extends the E2E test suite by adding the test case of creating a rootfull/rootless Podman machine with user mode networking enabled

https://github.com/containers/podman-desktop/issues/4666

### How to test this PR?
Run the e2e tests with yarn test:e2e


tests/playwright/src/utility/operations.ts
```diff
@@ -181,27 +183,57 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
...
+      await playExpect(podmanResourceCard.resourceElementConnectionStatus).toHaveText(ResourceElementState.Off, {
+        timeout: 30_000,
+      });
+      console.log('Podman machine stopped via CLI');
+    }
     if ((await podmanResourceCard.resourceElementConnectionStatus.innerText()) === ResourceElementState.Running) {
       await podmanResourceCard.performConnectionAction(ResourceElementActions.Stop);
       await playExpect(podmanResourceCard.resourceElementConnectionStatus).toHaveText(ResourceElementState.Off, {
-        timeout: 30_000,
+        timeout: 300_000,
```
ScrewTSW: 300 seconds for a machine to go offline seems a bit excessive, I'd say if it takes that long, that's a bug and should be reported to devs

ScrewTSW (approved):
LGTM, just a comment on timeout

tests/playwright/src/utility/operations.ts
```diff
@@ -214,6 +214,26 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
   }
 }
 
+export async function handleResetDefaultConnectionDialog(page: Page): Promise<void> {
+  const connectionDialog = page.getByRole('dialog', { name: 'Podman' });
+  const dialogMessage = connectionDialog.getByText(
+    new RegExp(
+      `.* Podman Machine .+ does not match default connection .+ Do you want to update the default connection?`,
```
cbr7: the dialog message has aria-labels, why are you looking for it in this way, pleasae use the coresponding label.

tests/playwright/src/utility/operations.ts
```diff
@@ -214,6 +214,26 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
...
+    ),
+  );
+  if ((await connectionDialog.isVisible()) && (await dialogMessage.isVisible())) {
+    const handleButton = connectionDialog.getByRole('button', { name: 'Yes' });
```
cbr7: Shouldn't checking for the existance/enablement of this button be sufficient? The button only exists when the popup is displayed so there is no point in checking at this time connectionDialog.isVisible() and dialogMessage.isVisible()

tests/playwright/src/utility/operations.ts
```diff
@@ -214,6 +214,26 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
...
+    await handleButton.click();
+    await playExpect(connectionDialog).toBeVisible();
+    const successDialogMessage = connectionDialog.getByText(
+      new RegExp(`Podman Machine .+ is now the default machine on the CLI.`),
+    );
```
cbr7: Please use the aria-labels.

cbr7 (changes requested):
Left some comments.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  await welcomePage.handleWelcomePage(true);
+  navBar = new NavigationBar(page);
+});
+
+test.afterAll(async () => {
+  await deletePodmanMachine(page, MACHINE_VISIBLE_NAME);
+  await pdRunner.close();
+});
+
+test.describe.serial('Rootless Podman machine Verification', () => {
```
xbabalov: Is there a need for this file in testing the user mode machine?

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    const clickableCheckbox = upperElement.getByText(checkText);
+    await clickableCheckbox.click();
+  }
+
+  async handleConnectionDialog(machineName: string, setAsDefault: boolean): Promise<void> {
+    const connectionDialog = this.page.getByRole('dialog', { name: 'Podman' });
+    const dialogMessage = connectionDialog.getByLabel('Dialog Message');
+    if ((await connectionDialog.isVisible()) && (await dialogMessage.isVisible())) {
```
xbabalov: I was checking the regex of dialog message so I could be sure the connection dialog was regarding specifically the newly created machine, but if we don't want to check that we don't need to check the visibility of the dialog message if we know the dialog is already there.

tests/playwright/src/utility/operations.ts
```diff
@@ -202,16 +202,35 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
...
     }
     await podmanResourceCard.performConnectionAction(ResourceElementActions.Delete);
-    await playExpect(podmanResourceCard.resourceElement).toBeHidden({ timeout: 30_000 });
+    await playExpect(podmanResourceCard.resourceElement).toBeHidden({ timeout: 100_000 });
   } else {
     console.log(`Podman machine [${machineVisibleName}] not present, skipping deletion.`);
   }
 }
 
+export async function handleResetDefaultConnectionDialog(page: Page): Promise<void> {
```
xbabalov: This function is defined but never used

tests/playwright/src/specs/podman-machine-rootless.spec.ts
danivilla9: No, it's not needed. I updated it because it prompted an error when testing it on GH Actions. If you have a different version you can substitute it completely.

xbabalov: I'm still working on that file and it's not on containers main yet, just my branch where I'm testing it, so you don't have to update it in your PR.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,134 @@
...
+    this.closeButton = this.page.getByRole('button', { name: 'Close' });
+    this.createMachineButton = this.page.getByRole('button', { name: 'Create' });
+  }
+
+  async createMachine(
+    machineName: string,
+    isRootful: boolean = true,
+    enableUserNet: boolean = false,
+    startNow: boolean = true,
+    setAsDefault: boolean = true,
```
cbr7: Please refactor this to use object descructuring, these are too many optional params not to be descructured.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    return new ResourcesPage(this.page);
+  }
+
+  async isEnabled(checkbox: Locator): Promise<boolean> {
+    await playExpect(checkbox).toBeVisible();
+    const upperElement = checkbox.locator('..').locator('..');
+    const clickableCheckbox = upperElement.getByText('Enabled');
+    return await clickableCheckbox.isVisible();
+  }
```
cbr7: why do you not just check if the checkbox isChecked()?

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,134 @@
...
+    let checkText;
+    if (wasEnabled) {
+      checkText = 'Enabled';
+    } else {
+      checkText = 'Disabled';
+    }
+
+    const clickableCheckbox = upperElement.getByText(checkText);
+    await clickableCheckbox.click();
+  }
```
cbr7, line 127: Same as above, the checkbox should be checkable for the isChecked() state.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,134 @@
...
+  async handleConnectionDialog(setAsDefault: boolean): Promise<void> {
+    const connectionDialog = this.page.getByRole('dialog', { name: 'Podman' });
+    if (await connectionDialog.isVisible()) {
+      let handleButtonName = 'Yes';
+      if (!setAsDefault) {
+        handleButtonName = 'Ignore';
+      }
+      const handleButton = connectionDialog.getByRole('button', { name: handleButtonName });
+      await handleButton.click();
+    }
```
cbr7, line 138: This is not the right way to check this, the method isVisible() always returns immediately, it has no innate waiter, as such it can return false just as easily as true, simply because of system lag, this is very flaky behavior.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    await this.createMachineButton.click();
+
+    const successfulCreationMessage = this.page.getByText('Successful operation');
+    const goBackToResourcesButton = this.page.getByRole('button', { name: 'Go back to resources' });
+
+    await playExpect(successfulCreationMessage).toBeVisible({ timeout: 200_000 });
+    await playExpect(goBackToResourcesButton).toBeVisible();
+    await this.page.waitForTimeout(2_000); //wait for dialog to appear, in the needed cases
+    await this.handleConnectionDialog(setAsDefault);
```
cbr7, line 101: This is not the best approach, you should be waiting until the element that you need is available, I would say that in this case you should use `await playExpect(setAsDefault).toBeVisible()` instead of waitForTimeout(), which could very well introduce flakyness.

By waiting on the specific element it also allows the implementation of handleConnectionDialog() method to be refactored, as pointed out below.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,134 @@
...
+      await this.switchCheckbox(this.rootPriviledgesCheckbox);
+    }
+
+    if (enableUserNet !== (await this.isEnabled(this.userModeNetworkingCheckbox))) {
+      await this.switchCheckbox(this.userModeNetworkingCheckbox);
+    }
+
+    if (startNow !== (await this.isEnabled(this.startNowCheckbox))) {
+      await this.switchCheckbox(this.startNowCheckbox);
+    }
```
cbr7, line 91: These can be refactor in order for the if statements to not be required, there should be some method named `ensureCheckboxState()` that accepts the checkbox locator and the desired state of the checkbox which checks the initial state and switches the state in case it does not match the desired state.

tests/playwright/src/model/pages/create-machine-page.ts
```diff
@@ -0,0 +1,134 @@
+/**********************************************************************
```
cbr7, line 1: File should be added to index.ts file to be exported.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  process.env.KEEP_TRACES_ON_PASS = 'true';
+
+  const welcomePage = new WelcomePage(page);
+  await welcomePage.handleWelcomePage(true);
+  navBar = new NavigationBar(page);
+});
+
+test.afterAll(async () => {
+  await deletePodmanMachine(page, MACHINE_VISIBLE_NAME);
+  await pdRunner.close();
```
cbr7: the close() method call should always be in a finally block, in order to ensure that it's executed even if some exception is thrown in the deletePodmanMachine() method.

tests/playwright/src/specs/podman-machine-rootless.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    await podmanResources.createButton.click();
+
+    const createMachinePage = new CreateMachinePage(page);
+    await createMachinePage.createMachine(PODMAN_MACHINE_NAME, false, false, true, false);
+    await createMachinePage.handleConnectionDialog(PODMAN_MACHINE_NAME, false);
+
+    const machineBox = new ResourceConnectionCardPage(page, 'podman', MACHINE_VISIBLE_NAME);
+    const connectionStatusLabel = await machineBox.resourceElementConnectionStatus.textContent();
+    playExpect(connectionStatusLabel === 'RUNNING').toBeTruthy();
```
cbr7:
```suggestion
    await playExpect(await machineBox.resourceElementConnectionStatus).toContain('RUNNING');
```

tests/playwright/src/specs/podman-machine-user-mode.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+const PODMAN_MACHINE_NAME: string = 'podman-machine-user-mode';
+const MACHINE_VISIBLE_NAME: string = 'Podman Machine user-mode';
+
+test.skip(!isWindows, 'Test should run only on Windows');
+
+test.beforeAll(async () => {
+  pdRunner = new PodmanDesktopRunner();
+  page = await pdRunner.start();
+  pdRunner.setVideoAndTraceName('podman-machine-user-mode');
+  process.env.KEEP_TRACES_ON_PASS = 'true';
```
cbr7: Why are we setting this env var here?

tests/playwright/src/specs/podman-machine-user-mode.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    const podmanResources = new ResourceConnectionCardPage(page, 'podman');
+
+    await podmanResources.createButton.click();
+
+    const createMachinePage = new CreateMachinePage(page);
+    await createMachinePage.createMachine(PODMAN_MACHINE_NAME, false, true, true, false);
+
+    const machineBox = new ResourceConnectionCardPage(page, 'podman', MACHINE_VISIBLE_NAME);
+    const connectionStatusLabel = await machineBox.resourceElementConnectionStatus.textContent();
+    playExpect(connectionStatusLabel === 'RUNNING').toBeTruthy();
```
cbr7, line 53:
```suggestion
await playExpect(await machineBox.resourceElementConnectionStatus).toContain('RUNNING');
```

tests/playwright/src/specs/podman-machine-user-mode.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    const podmanResources = new ResourceConnectionCardPage(page, 'podman');
+
+    await podmanResources.createButton.click();
+
+    const createMachinePage = new CreateMachinePage(page);
+    await createMachinePage.createMachine(PODMAN_MACHINE_NAME, true, true, true, false);
+
+    const machineBox = new ResourceConnectionCardPage(page, 'podman', MACHINE_VISIBLE_NAME);
+    const connectionStatusLabel = await machineBox.resourceElementConnectionStatus.textContent();
+    playExpect(connectionStatusLabel === 'RUNNING').toBeTruthy();
```
cbr7, line 73:
```suggestion
await playExpect(await machineBox.resourceElementConnectionStatus).toContain('RUNNING');
```

tests/playwright/src/utility/operations.ts
```diff
@@ -202,16 +202,20 @@ export async function deletePodmanMachine(page: Page, machineVisibleName: string
...
+    await playExpect(podmanResourceCard.resourceElement).toBeHidden({ timeout: 100_000 });
   } else {
     console.log(`Podman machine [${machineVisibleName}] not present, skipping deletion.`);
   }
 }
 
+export function checkForFailedTest(result: TaskResult, runner: PodmanDesktopRunner): void {
+  if (result.errors && result.errors.length > 0) runner.setTestPassed(false);
+}
+
```
cbr7, line 221: What is this method used for?

tests/playwright/src/utility/operations.ts
```diff
@@ -137,7 +137,7 @@ export async function deletePod(page: Page, name: string): Promise<void> {
         async () => {
           return !!(await pods.getPodRowByName(name));
         },
-        { timeout: 20000 },
+        { timeout: 20_000 },
```
cbr7, line 143: Are all these timeout refactorings required, what is the point of them?

cbr7 (changes requested):
Left some comments in the PR.

odockal:
@danivilla9 @cbr7 We can rebase and finish the test, right?

cbr7:
Test is already done from a different PR, we can close this.

PR #8534, opened 2024-08-21
feat: support cli tool installer

lstocchi:
### What does this PR do?

It enables extensions to register an installer so that a cli tool can be installed from the cli tools page.

The uninstall will be added in a separate PR.
I didn't modified how the kind extension works. If not installed, there is always the item in the statusbar, but we can add an installer as the other extensions in a follow-up

### Screenshot / video of UI

install_cli_tool

it is part of #8003 

### How to test this PR?

1. uninstall compose, kind, kubectl
2. go to the tools page and install them

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4335,7 +4335,7 @@ declare module '@podman-desktop/api' {
    * Options to update CliTool instance
    */
   export interface CliToolUpdateOptions {
-    version?: string;
+    version: string;
     displayName?: string;
     markdownDescription?: string;
     images?: ProviderImages;
@@ -4373,7 +4373,7 @@ declare module '@podman-desktop/api' {
     };
 
     updateVersion(version: CliToolUpdateOptions): void;
-    onDidUpdateVersion: Event<string | undefined>;
+    onDidUpdateVersion: Event<string>;
 
     // register cli update flow
     registerUpdate(update: CliToolUpdate | CliToolSelectUpdate): Disposable;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4318,9 +4318,10 @@ declare module '@podman-desktop/api' {
      * Passing in path will also help to show where the CLI tool is expected to be installed.
      * This is usually the ~/.local/share/containers/podman-desktop/extensions-storage directory.
      * Note: The expected value should not include 'v'.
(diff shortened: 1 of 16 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -4367,10 +4373,13 @@ declare module '@podman-desktop/api' {
     };
 
     updateVersion(version: CliToolUpdateOptions): void;
-    onDidUpdateVersion: Event<string>;
+    onDidUpdateVersion: Event<string | undefined>;
```
benoitf: in which case we don't have the version when it's being updated ?

lstocchi: when we uninstall. At first i wanted to add the uninstall in this PR as well then i changed my mind as I thought it would have been easier to split them. 
I can remove this part and re-add in the uninstall PR, if preferred.

lstocchi: I think it's better to clean it, i'll update the PR

benoitf: maybe yes remove it as it'll be clearer

lstocchi: removed 👍

packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
```diff
@@ -46,8 +53,38 @@ async function update(cliTool: CliToolInfo) {
...
+}
+
+async function install(cliTool: CliToolInfo) {
+  // user has to select the version to install
+  let versionToInstall;
+  try {
+    versionToInstall = await window.selectCliToolVersionToInstall(cliTool.id);
+  } catch (e) {
+    // do nothing
+    console.log(e);
```
benoitf:
```suggestion
    console.error(e);
```

benoitf: error shouldn't be displayed to the end user or thrown ?

like not selecting a version and having an error in the select looks different usecase ?

benoitf (approved):

packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
lstocchi: i took advantage of the error message already in the component to show it

benoitf: thanks

lstocchi:
I'm gonna merge in the next minutes to keep working on the next features (uninstall, list tools) if nobody is looking at it. Let me know

PR #8547, opened 2024-08-22
feat: allow to uninstall cli tool

### LINKED ISSUE #8003: CLI Tools should support installing the binary

lstocchi:
### What does this PR do?

It implements the logic to register the uninstall action and update the compose + kubectl extensions

### Screenshot / video of UI

uninstall_cli_tool

it is part of #8003 

### How to test this PR?

1. uninstall a cli you installed and see it is actually removed.

### DIFF
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -16,6 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import * as fs from 'node:fs';
 import * as path from 'node:path';
 
 import { Octokit } from '@octokit/rest';
@@ -361,6 +362,30 @@ async function postActivate(
       currentVersion = releaseVersionToInstall;
       releaseVersionToInstall = undefined;
       releaseToInstall = undefined;
+    },
+    doUninstall: async _logger => {
+      if (!currentVersion) {
+        throw new Error(`Cannot uninstall ${kubectlCliName}. No version detected.`);
+      }
+
+      // delete the executable stored in the storage folder
+      const storagePath = getStorageKubectlPath(extensionContext);
+      if (fs.existsSync(storagePath)) {
+        await fs.promises.unlink(storagePath);
+      }
+
+      // delete the executable in the system path
+      const systemPath = getSystemBinaryPath(kubectlCliName);
+      if (fs.existsSync(systemPath)) {
+        await fs.promises.unlink(systemPath);
+      }
+
+      // update the version to undefined
+      currentVersion = undefined;
+      kubectlCliTool?.updateVersion({
+        version: undefined,
+        path: undefined,
+      });
     },
   });
 
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -382,10 +382,6 @@ async function postActivate(
 
       // update the version to undefined
       currentVersion = undefined;
-      kubectlCliTool?.updateVersion({
-        version: undefined,
-        path: undefined,
-      });
     },
   });
 
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -370,15 +370,11 @@ async function postActivate(
 
       // delete the executable stored in the storage folder
       const storagePath = getStorageKubectlPath(extensionContext);
-      if (fs.existsSync(storagePath)) {
-        await fs.promises.unlink(storagePath);
-      }
+      await deleteFile(storagePath);
 
       // delete the executable in the system path
       const systemPath = getSystemBinaryPath(kubectlCliName);
-      if (fs.existsSync(systemPath)) {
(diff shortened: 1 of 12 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -4373,7 +4374,7 @@ declare module '@podman-desktop/api' {
     };
 
     updateVersion(version: CliToolUpdateOptions): void;
-    onDidUpdateVersion: Event<string>;
+    onDidUpdateVersion: Event<string | undefined>;
```
benoitf: I'm wondering, should we receive an update event or a uninstall event ?

extensions/kubectl-cli/src/extension.ts
```diff
@@ -362,6 +363,30 @@ async function postActivate(
...
+      if (fs.existsSync(systemPath)) {
+        await fs.promises.unlink(systemPath);
+      }
+
+      // update the version to undefined
+      currentVersion = undefined;
+      kubectlCliTool?.updateVersion({
+        version: undefined,
+        path: undefined,
+      });
```
benoitf: here it's not clear to me, should we have a remove or uninstall method instead ?

lstocchi: by thinking at it maybe it should not be the extension to update the version. When the core calls the uninstall and succeed then it knows that the version is not there anymore and updates it? So no need for a remove/uninstall method?

benoitf: maybe it may look more natural the way you describe

but then there should be event fired by the framework (like onDidUninstall)

lstocchi: sure 👍

packages/extension-api/src/extension-api.d.ts
lstocchi: updated

extensions/kubectl-cli/src/extension.ts
lstocchi: updated

benoitf (approved):
looks cleaner, thanks

extensions/kubectl-cli/src/extension.ts
```diff
@@ -362,6 +363,26 @@ async function postActivate(
...
+      // delete the executable stored in the storage folder
+      const storagePath = getStorageKubectlPath(extensionContext);
+      if (fs.existsSync(storagePath)) {
+        await fs.promises.unlink(storagePath);
+      }
+
+      // delete the executable in the system path
+      const systemPath = getSystemBinaryPath(kubectlCliName);
+      if (fs.existsSync(systemPath)) {
+        await fs.promises.unlink(systemPath);
```
axel7083: We need admin privilege to remove on the system path

image

axel7083 (commented):
Great job! I think we may need to have some confirmation dialog for the un-installation

lstocchi:
@axel7083 added a confirmation dialog and fixed the admin error. Give it a look whenever you can. Thanks 👍

extensions/kubectl-cli/src/extension.ts
lstocchi: Thanks for noticing, i was doing a test with the admin user and i didn't face it

axel7083 (approved):
Tested again, and LGTM, I feel like we may need to indicate the path of the binary that is going to be deleted but this is not blocking for me. Great job !

lstocchi:
> Tested again, and LGTM, I feel like we may need to indicate the path of the binary that is going to be deleted but this is not blocking for me. Great job !

Going to merge this PR as it is . Not sure about the path bc i think it will always be the same (the one in extensions-storage and one in system path)? 🤔  But yes, we can add it in a new PR easily

PR #8552, opened 2024-08-22
feat: allow to retrieve a cli tool

lstocchi:
### What does this PR do?

This is part of #8003 . In the issue it was required to enhance the cli namespace to allow retrieving  cli tool 

### Screenshot / video of UI

N/A

it is part of #8003 

### How to test this PR?

1. run tests

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4405,12 +4405,12 @@ declare module '@podman-desktop/api' {
     export function getCliTool(id: string): CliToolInfo | undefined;
 
     /**
-     * All extensions currently known to the system.
+     * All cli tools currently known to the system.
      */
     export const all: readonly CliToolInfo[];
 
     /**
-     * An event which fires when `cli.all` changes. This can happen when extensions are
+     * An event which fires when `cli.all` changes. This can happen when cli are
      * installed, uninstalled, enabled or disabled.
      */
     export const onDidChange: Event<void>;
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4395,6 +4395,23 @@ declare module '@podman-desktop/api' {
      * @returns CliTool instance
      */
     export function createCliTool(options: CliToolOptions): CliTool;
+
+    /**
+     * given an id, return the corresponding CLI Tool
+     * @param id cli tool
+     */
+    export function getCliTool(id: string): CliTool | undefined;
+
+    /**
+     * All extensions currently known to the system.
+     */
+    export const all: readonly CliTool[];
+
+    /**
+     * An event which fires when `cli.all` changes. This can happen when extensions are
+     * installed, uninstalled, enabled or disabled.
+     */
+    export const onDidChange: Event<void>;
   }
 
   /**
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4360,7 +4360,18 @@ declare module '@podman-desktop/api' {
 
   export type CliToolState = 'registered';
 
-  export interface CliTool extends Disposable {
+  export interface CliTool extends CliToolInfo, Disposable {
+    updateVersion(version: CliToolUpdateOptions): void;
+    onDidUpdateVersion: Event<string>;
+
+    // register cli update flow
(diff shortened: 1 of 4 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -4395,6 +4395,23 @@ declare module '@podman-desktop/api' {
...
+    /**
+     * given an id, return the corresponding CLI Tool
+     * @param id cli tool
+     */
+    export function getCliTool(id: string): CliTool | undefined;
+
+    /**
+     * All extensions currently known to the system.
+     */
+    export const all: readonly CliTool[];
```
benoitf: here I'm not sure. Because an extension B should not be able to receive the object provided by the extension A for security/responsability

you should receive some information (but a subset in readonly)

lstocchi: updated, not sure what kind of subset you meant. I just returned all clitool properties excluding its methods

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4395,6 +4397,23 @@ declare module '@podman-desktop/api' {
      * @returns CliTool instance
      */
     export function createCliTool(options: CliToolOptions): CliTool;
+
+    /**
+     * given an id, return the corresponding CLI Tool
+     * @param id cli tool
+     */
+    export function getCliTool(id: string): CliTool | undefined;
```
benoitf: here it should be CliToolInfo as well

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1320,10 +1320,19 @@ export class ExtensionLoader {
...
-        const cliTool = this.cliToolRegistry.createCliTool(extensionInfo, options);
+        const cliTool = instance.cliToolRegistry.createCliTool(extensionInfo, options);
         disposables.push(cliTool);
         return cliTool;
       },
+      getCliTool: (id: string): containerDesktopAPI.CliTool | undefined => {
+        return instance.cliToolRegistry.getCliTool(id);
+      },
+      get all() {
+        return instance.cliToolRegistry.getCliTools();
```
benoitf, line 1330: the object that we're returning has still all the properties, it should be a subset (new object) as here it's possible to call all the methods

packages/extension-api/src/extension-api.d.ts
lstocchi: right, done

packages/main/src/plugin/cli-tool-registry.ts
```diff
@@ -137,4 +145,40 @@ export class CliToolRegistry {
...
+        id: cliTool.extensionInfo.id,
+        label: cliTool.extensionInfo.label,
+      },
+    };
+  }
+
+  getCliTools(): CliToolInfoApi[] {
+    return Array.from(this.cliTools.values()).map(cliTool => {
+      return {
+        id: cliTool.id,
```
benoitf: maybe a toInfo() like method shared between getCliTool and getCliTools (transforming a CliToo to CliTooInfo ?)

lstocchi: yes, better. done

benoitf (approved):

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4360,7 +4360,18 @@ declare module '@podman-desktop/api' {
...
+    onDidUpdateVersion: Event<string>;
+
+    // register cli update flow
+    registerUpdate(update: CliToolUpdate | CliToolSelectUpdate): Disposable;
+
+    // register cli installer
+    registerInstaller(installer: CliToolInstaller): Disposable;
+  }
+
+  export interface CliToolInfo {
```
axel7083, line 4374: I am wondering if we should not include the version in some way ? If I need at some point to require a specific cli tool, I feel like I would check the version to ensure compatibility with my code

axel7083: I was looking at the code, and it feel odd that when registering a cli tool we need to provide a version (or undefined); but the version field is not in the Cli object.

axel7083: This could be a follow up PR as this is a problem that was already here before

axel7083 (approved):
LGTM

PR #8561, opened 2024-08-23
chore: ensure we don't avoid promises to be awaited using void prefix

benoitf:
### What does this PR do?
Ensure we're not tempted to use void prefix to bypass the warning

### How to test this PR?


jeffmaury (approved):
LGTM

PR #8572, opened 2024-08-25
chore: use stricter typescript rules on the main package

benoitf:
### What does this PR do?
Enable `"@tsconfig/strictest"` typescript rule on the main package to enforce more checks int he source code

but disable for now `exactOptionalPropertyTypes` from this rule

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -453,7 +453,9 @@ export class ContainerProviderRegistry {
               if (podmanContainer.Labels) {
                 // copy all labels
                 for (const label of Object.keys(podmanContainer.Labels)) {
-                  Labels[label] = podmanContainer.Labels[label];
+                  if (podmanContainer.Labels[label]) {
+                    Labels[label] = podmanContainer.Labels[label];
+                  }
                 }
               }
 
@@ -970,7 +972,7 @@ export class ContainerProviderRegistry {
     });
 
     const matchingConnection = matchingContainerProviders[0];
-    if (!matchingConnection[1].api) {
+    if (!matchingConnection?.[1].api) {
       throw new Error('No provider with a running engine');
     }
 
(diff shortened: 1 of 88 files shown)


packages/main/src/plugin/container-registry.ts
```diff
@@ -1070,7 +1072,7 @@ export class ContainerProviderRegistry {
 
   getImageName(inspectInfo: Dockerode.ImageInspectInfo): string {
     const tags = inspectInfo.RepoTags;
-    if (!tags) {
+    if (!tags?.[0]) {
```
axel7083, line 1075:
```suggestion
    if (!tags || tags.length === 0 || !tags[0]) {
```

axel7083 (approved):
Thanks for the works done :rocket: kudos for taking the time to do it.

> I can't really read and check every change due to the size of the PR, but changes LGTM

packages/main/src/plugin/container-registry.ts
benoitf: hi @axel7083 , could you give me why we would need these 3 condition rather than one ?

benoitf: I think when editing at some point I received an error telling me to merge the condition as well

axel7083: image

I just checked, and there is no need, as javascript do not raise error while accessing index that does not exists !

benoitf: yes the optional chaining is checking that for us

packages/main/src/plugin/kubernetes-client-kubeconfig.spec.ts
```diff
@@ -109,7 +113,16 @@ describe('context tests', () => {
...
     client.saveKubeConfig = vi.fn().mockImplementation((_config: KubeConfig) => {});
 
+    if (!originalContexts[0]?.name) {
+      throw new Error('originalContexts[0].name should be defined');
+    }
+
     await client.deleteContext(originalContexts[0].name);
+
+    if (!originalContexts[1]?.name) {
+      throw new Error('originalContexts[0].name should be defined');
```
lstocchi:
```suggestion
      throw new Error('originalContexts[1].name should be defined');
```

lstocchi (approved):
LGTM

PR #8586, opened 2024-08-26
chore: switch to eslint v9.9.1

### LINKED ISSUE #8565: update eslint to v9

eslint v8 will become unsupported in october
https://eslint.org/version-support/

time to switch to eslint v9

there is an unified configuration file now and a migration to follow

switch the multiple configuration files and use the new configuration files

.eslintrc, .eslintrc.json, .eslintrc.yml are gone and it's now using flat configuration files

_No response_

### Additional context

we may have to find other plug-ins as there are breaking changes even in the 'compat' support provided to ease the migration

benoitf:
### What does this PR do?

Switch to the new version of the engine using flat config files
https://eslint.org/docs/latest/use/configure/configuration-files

(so instead of having many files across all the repository, for ignoring, configuration, etc, now there is only one file)

one plug-in is not compliant with the new engine, so I had to remove it, will see if we find an alternative as it's no longer maintained.
https://github.com/Stuk/eslint-plugin-header/issues/57

And for the plug-in import, I had to disable one rule `import/namespace` that is failing
There is work in progress there so hopefully we'll be able to include it again soon
https://github.com/import-js/eslint-plugin-import/issues/2948

For sonarJS I was able to switch to the recommended one (rather than recommended-legacy as of today)

I had to remove the --ext flag in package.json scripts as the flag has been removed in the CLI v9
(need to specify supported files in the config file)

fixes https://github.com/containers/podman-desktop/issues/8565

### How to test this PR?

`lint:check` command should be ✅ 


axel7083 (commented):
`yarn run lint:check` is giving the following warnings

```
/home/axel7083/github/redhat/podman-desktop/website/src/pages/storybook/sidebar.cjs
  1:20   error  Strings must use singlequote  quotes
  1:27   error  Strings must use singlequote  quotes
  1:33   error  Strings must use singlequote  quotes
  1:38   error  Strings must use singlequote  quotes
...
```

I also added ` 'storybook/storybook-static/**/*'` to the list of ignores to faster the process (took 11 minutes without cache)

benoitf:
I amended the PR

axel7083 (approved):
LGTM

PR #8589, opened 2024-08-26
chore: switch to pnpm instead of yarn

### LINKED ISSUE #8566: switch to pnpm as package manager

Today, Podman Desktop is using yarn package manager.

project is using the v1 of yarn.
Now there is a v4 so we could bump to v4 or either switch to another package manager.

A lot of the javascript projects have moved to pnpm https://pnpm.io/workspaces#usage-examples
In 2024 it looks like it's the default package manager.

We do have support for the features we're using https://pnpm.io/feature-comparison

Use pnpm as package manager

it means that we should update the source code and the documentation.

The tricky part is the flatpak builder as flatpak builder out of the box only support npm and yarn v1

the yarn v3 is not supported neither https://github.com/flatpak/flatpak-builder-tools/issues/361 (and now yarn is at v4)

One advantage of pnpm over yarn is the local-store that we can then backup and restore in the flatpak builder environement and make it work

Here is an attempt illustrating that it's even easier to use pnpm than yarn for flathub (copy and restore the pnpm store)
https://github.com/flathub/io.podman_desktop.PodmanDesktop/pull/53

From performance POV there is a benchkmark from yarn website https://yarnpkg.com/features/performances
https://p.datadoghq.eu/sb/d2wdprp9uki7gfks-c562c42f4dfd0ade4885690fa719c818

and one from pnpm: https://pnpm.io/fr/benchmarks

I think it's not easily possible for now to use some yarn or pnpm features like the symlinks and we need the 'legacy structure' for electron
https://github.com/pnpm/pnpm/issues/3415

Github actions have also now an easy way to handle the cache (directly on the node action)
https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

(it supports npm, yarn and pnpm)

and npm tools is now supporting corepack so installing a new package manager is straightforward https://pnpm.io/installation#using-corepack

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Migrate to pnpm rather than using yarn v1

- https://pnpm.io/installation

Open PR so people can experiment the usage of pnpm instead of yarn and see if things are running smoothly or not

Fixes #8566 

### How to test this PR?

use `pnpm <command>` instead of `yarn <command>`


jeffmaury (commented):
Looks like a regression from the perfomance POV as it took me 6mn to get the dependencies !!!!

I think yarn.lock should be removed

benoitf:
> Looks like a regression from the perfomance POV as it took me 6mn to get the dependencies !!!!

it's the time to initialize your global store I guess.

but then each time you clone dependencies it'll fetch only the delta


if you remove `%LOCALAPPDATA%/Yarn/config/global` (and node_modules, how long does it take with yarn ? )

benoitf:
> I think yarn.lock should be removed

it's removed from the PR but I guess then it should be ignored ?

jeffmaury:
> > Looks like a regression from the perfomance POV as it took me 6mn to get the dependencies !!!!
> 
> it's the time to initialize your global store I guess.
> 
> but then each time you clone dependencies it'll fetch only the delta
> 
> if you remove `%LOCALAPPDATA%/Yarn/config/global` (and node_modules, how long does it take with yarn ? )

Once I remove the yarn cache I got the same amount of time to resolve the dependencies

jeffmaury:
> > I think yarn.lock should be removed
> 
> it's removed from the PR but I guess then it should be ignored ?

Removal seems enough for me.

jeffmaury (approved):
LGTM.

Was able to run several tasks

lstocchi (approved):
Works fine on my win machine. I don't see any gain in performances tbh but works fine

jeffmaury:
FYI I add an issue after testing this PR. I was not able to run yarn on Podman AI Lab. When running yarn on Podman AI Lab it modifies the package.json and added a new packageManager field then errored because of that. I add to run corepack disable pnpm to be ok

odockal (approved):
Tried locally on my fedora machine. I have not seen any issues, except the javascript problem with electron dep. I can build, compile, etc. Looks good. We will have to update also other jobs (windows and mac) once this is merged. I will create issues.

PR #8592, opened 2024-08-26
chore: update got to v14.4.2

benoitf:
### What does this PR do?
I don't know why dependabot is not updating this dependency...

### How to test this PR?


benoitf:
ok it is because at some point we were not using Node.js v20 but 18

axel7083 (approved):

PR #8603, opened 2024-08-27
fix: remove state from CliToolInfo

### LINKED ISSUE #8599: [extension-api] CliToolState are not enough

When getting all the `CliToolInfo` we are getting an array of object, but extensions have no way of knowing if they are installed or not.

https://github.com/containers/podman-desktop/issues/8598 could be a solution, but there is the `CliToolState` which is always `registered`. This is not a useful information, as extension may need to know if it installed or not.

https://github.com/containers/podman-desktop/blob/013552134921d5326359cff048f59576bfa699b0/packages/extension-api/src/extension-api.d.ts#L4361

We should add a new value to the `CliToolState`. My proposal would be

```
export type CliToolState = 'registered' | 'installed'
```

Therefore, extensions will be able to determine if the cli tool is installed or not.

_No response_

### Additional context

cc @benoitf

axel7083:
### What does this PR do?

As discussed in https://github.com/containers/podman-desktop/issues/8599 moving the state to CliTool for backward compliance.

### Screenshot / video of UI

N/A

Fixes https://github.com/containers/podman-desktop/issues/8599

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4361,6 +4361,7 @@ declare module '@podman-desktop/api' {
   export type CliToolState = 'registered';
 
   export interface CliTool extends CliToolInfo, Disposable {
+    state: CliToolState;
     updateVersion(version: CliToolUpdateOptions): void;
     onDidUpdateVersion: Event<string>;
 
@@ -4376,7 +4377,6 @@ declare module '@podman-desktop/api' {
     name: string;
     displayName: string;
     markdownDescription: string;
-    state: CliToolState;
     images: ProviderImages;
     version?: string;
     extensionInfo: {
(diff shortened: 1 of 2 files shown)


benoitf:
it needs to be removed there as well

https://github.com/containers/podman-desktop/blob/2a758a388d1f3e8ae239906ad62d1b8caf60e626/packages/main/src/plugin/cli-tool-registry.ts#L165

axel7083:
> it needs to be removed there as well
> 
> https://github.com/containers/podman-desktop/blob/2a758a388d1f3e8ae239906ad62d1b8caf60e626/packages/main/src/plugin/cli-tool-registry.ts#L165

Oh yeah, good catch

lstocchi (approved):

benoitf (approved):

PR #8608, opened 2024-08-27
feat: extension can define task action

### LINKED ISSUE #7709: feat(Task): extensions should be able to define an action
In AI Lab, we are interesting to be able to set the Task action, to be able to open a specific page in AI Lab when something is running

https://github.com/containers/podman-desktop/blob/cfeb27546799febf3383b236e5602bbcee204970/packages/api/src/task.ts#L28-L36

However, tasks cannot define this field in the `main` package, as they are executed in the `renderer`. Trying to set this field lead to Serialization issue, as the task cannot be serialized.

## Solution proposal

In the extension api, precisely ProgressOptions, extensions should be able to define `action`, which should be visible in the TaskManager. 

https://github.com/containers/podman-desktop/blob/1ba569f2af51a3b1c6c6bcd72e3966dfb50a1707/packages/extension-api/src/extension-api.d.ts#L900

image

axel7083:
### What does this PR do?

Allowing extensions to define an action when calling `window.withProgress`. Backward compatible as it is an optional option. 

### Screenshot / video of UI

**Demo example**

https://github.com/user-attachments/assets/eb7ae27e-874d-441f-81a7-8874c65820fe

> This is not implemented, only demo purpose

Fixes https://github.com/containers/podman

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1070,6 +1070,51 @@ declare module '@podman-desktop/api' {
      * button.
      */
     cancellable?: boolean;
+
+    /**
+     * Progress may have an associate action
+     *
+     * @example
+     * Action could be used to redirect the user to a specific page
+     * ```ts
+     * import { window, navigation } from '@podman-desktop/api';
+     *
+     * window.withProgress({
+     *   title: 'Collecting container logs',
+     *   action: {
+     *     name: 'See details',
+     *     execute: () => {
+     *       navigation.navigateToContainerLogs('container-id');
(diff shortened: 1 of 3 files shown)


benoitf (commented):
I think it's not how usually 3rd party actions are declared.

it's in two different steps

extension has commands, and you have a command-id as post-action or action

benoitf:
also withProgress in its design is not meant to be interactive

axel7083:
Okey so we may need to have some discussion to clarify the path to follow

axel7083:
Closing as out of context with the direction taken

PR #8615, opened 2024-08-27
fix: title_bar_should_reflect_selected_light_mode

### LINKED ISSUE #8440: Title bar should reflect selected light mode
### Bug description

The title bar should reflect selected dark/light mode. 

Light mode selected:
image

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

Switch to Light mode

### Relevant log output

_No response_

### Additional context

_No response_

gastoner:
### Screenshot / video of UI
prev:
image

now:

https://github.com/user-attachments/assets/26859737-6353-4744-85cf-d6488b852854

Closes #8440 

### How to test this PR?
Run Windows version of the PD in Light mode

### DIFF
--- a/packages/main/src/plugin/color-registry.spec.ts
+++ b/packages/main/src/plugin/color-registry.spec.ts
@@ -30,6 +30,7 @@ import type { ColorDefinition } from '/@api/color-info.js';
 import type { RawThemeContribution } from '/@api/theme-info.js';
 
 import colorPalette from '../../../../tailwind-color-palette.json';
+import * as util from '../util.js';
 import { ColorRegistry } from './color-registry.js';
 
 class TestColorRegistry extends ColorRegistry {
@@ -182,22 +183,48 @@ test('initColors', async () => {
   expect(spyOnRegisterColor.mock.calls.length).toBeGreaterThan(20);
 });
 
-test('initTitlebar', async () => {
-  // mock the registerColor
-  const spyOnRegisterColor = vi.spyOn(colorRegistry, 'registerColor');
-  spyOnRegisterColor.mockReturnValue(undefined);
+describe('initTitlebar', () => {
+  test('Check on Windows', async () => {
+    // mock the registerColor
+    const spyOnRegisterColor = vi.spyOn(colorRegistry, 'registerColor');
+    spyOnRegisterColor.mockReturnValue(undefined);
 
-  colorRegistry.initTitlebar();
+    //mock the isWindows to force using Windows colors
+    vi.spyOn(util, 'isWindows').mockReturnValue(true);
 
-  expect(spyOnRegisterColor).toHaveBeenCalled();
+    colorRegistry.initTitlebar();
 
-  // at least 3 times
-  expect(spyOnRegisterColor.mock.calls.length).toBeGreaterThanOrEqual(3);
+    expect(spyOnRegisterColor).toHaveBeenCalled();
 
-  // check the first call
-  expect(spyOnRegisterColor.mock.calls[0]?.[0]).toStrictEqual('titlebar-bg');
-  expect(spyOnRegisterColor.mock.calls[0]?.[1].light).toBe('#f9fafb');
-  expect(spyOnRegisterColor.mock.calls[0]?.[1].dark).toBe('#0f0f11');
+    // at least 3 times
+    expect(spyOnRegisterColor.mock.calls.length).toBeGreaterThanOrEqual(3);
+
+    // check the first call
+    expect(spyOnRegisterColor.mock.calls[0]?.[0]).toStrictEqual('titlebar-bg');
+    expect(spyOnRegisterColor.mock.calls[0]?.[1].light).toBe('#f9fafb');
+    expect(spyOnRegisterColor.mock.calls[0]?.[1].dark).toBe('#202020');
+  });
+
+  test('Check on macOS/Linux', async () => {
+    // mock the registerColor
+    const spyOnRegisterColor = vi.spyOn(colorRegistry, 'registerColor');
+    spyOnRegisterColor.mockReturnValue(undefined);
+
+    //mock the isWindows to force using macOS/Linux colors
+    vi.spyOn(util, 'isWindows').mockReturnValue(false);
+
+    colorRegistry.initTitlebar();
+
+    expect(spyOnRegisterColor).toHaveBeenCalled();
+
+    // at least 3 times
+    expect(spyOnRegisterColor.mock.calls.length).toBeGreaterThanOrEqual(3);
+
+    // check the first call
+    expect(spyOnRegisterColor.mock.calls[0]?.[0]).toStrictEqual('titlebar-bg');
+    expect(spyOnRegisterColor.mock.calls[0]?.[1].light).toBe('#f9fafb');
+    expect(spyOnRegisterColor.mock.calls[0]?.[1].dark).toBe('#0f0f11');
+  });
 });
 
 test('initCardContent', async () => {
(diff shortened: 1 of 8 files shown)


packages/renderer/src/lib/ui/TitleBar.svelte
```diff
@@ -15,11 +15,9 @@ onMount(async () => {
 
 <header
   id="navbar"
-  class="{platform === 'win32' ? 'bg-[#202020]' : 'bg-[var(--pd-titlebar-bg)]'} body-font z-[999] relative {platform ===
```
benoitf, line 18: hello, where {platform === 'win32' ? 'bg-[#202020]' } did go ?

packages/renderer/src/lib/window-control-buttons/WindowsControlButton.svelte
```diff
@@ -58,10 +58,10 @@ function executeAction() {
   on:click={() => executeAction()}
   aria-label={name}
   title={titleName}
-  class="h-[32px] w-[45px] cursor-pointer text-white {name === 'Close'
+  class="h-[32px] w-[45px] cursor-pointer {name === 'Close'
     ? 'hover:bg-[#be4425]'
-    : 'hover:bg-[#2d2d2d]'} flex place-items-center justify-center">
+    : 'hover:bg-charcoal-300'} flex place-items-center justify-center">
```
benoitf: why charcoal-300 and not 2d2d2d ?

packages/renderer/src/lib/ui/TitleBar.svelte
gastoner: Hi @benoitf , I have removed it to force Windows to use (`bg-[var(--pd-titlebar-bg)]`) light/dark titlebar as the Mac/Linux

packages/renderer/src/lib/window-control-buttons/WindowsControlButton.svelte
gastoner: @benoitf That's what Linux uses, I've discussed this with @ekidneyrh, and she will create a ticket for herself to change colors to be united and slightly redesign the topbar in the next sprint

packages/renderer/src/lib/images/WindowsMinIcon.svelte
```diff
@@ -1,5 +1,6 @@
 <script lang="ts">
 export let size = 16;
+export let color = '#ffffff';
```
jeffmaury: I'm wondering why this is not from color-registry and hard coded ?

packages/renderer/src/lib/window-control-buttons/WindowsControlButton.svelte
benoitf: AFAIK the colors were picking up the default colors of Windows so the bar was looking the same than the native widget of users.

gastoner: Ok, I'll keep the 2d2d2d then

packages/renderer/src/lib/images/WindowsMinIcon.svelte
gastoner: This is currently hard-coded in the icon, this allows to change the text color. But I can try to change it to use the color-registry ones

packages/renderer/src/lib/window-control-buttons/WindowsControlButton.svelte
benoitf: well I think you should use a variable, but in color-registry maybe we should override the value in case of Windows 

if (isWindows()) {
do not use charcoal but like 2d2d
and handle the light and dark colors
}

packages/renderer/src/lib/ui/TitleBar.svelte
benoitf: yes but probably override the value in case of Windows to match the default colors of Windows

maybe a later improvement, we could also ask the windows registry the colors of the user to match their values

if user has title bars being like 'orange' we should use the same color ? (but it's not related to this PR)

packages/renderer/src/lib/window-control-buttons/WindowsControlButton.svelte
gastoner: Well this is just the Windows file (Linux is has own implementation [here](https://github.com/containers/podman-desktop/blob/main/packages/renderer/src/lib/window-control-buttons/LinuxControlButton.svelte) and Mac is native? since I don't see the implementation in repo).

So I think adding this would be probably unnecessary?

benoitf: no color definition should happen outside of the color-registry

the reason is that if you use a variable, it's possible to override the value by an extension or other mechanism while if it's hardcoded in the svelte file, you need to recompile the binary

gastoner: Ok, so I will add be4425 and 2d2d2d to color-registry and use the variables in here

gastoner: Should I add those colors to `tailwind-color-palette.json` and from there use them in color-registry?

benoitf: I would say no as it's not part of the color palette

gastoner:
@benoitf @jeffmaury Does it look better now?

packages/main/src/plugin/color-registry.ts
```diff
@@ -331,6 +331,15 @@ export class ColorRegistry {
...
       light: colorPalette.purple[900],
     });
+
+    this.registerColor('titlebar-windows-hover-exit-bg', {
+      dark: '#be4425',
+      light: '#be4425',
+    });
+    this.registerColor('titlebar-windows-hover-bg', {
+      dark: '#2d2d2d',
+      light: '#2d2d2d',
```
benoitf, line 342: is it expected that colors for light and dark mode are the same ?

benoitf:
@gastoner I still don't see an override in color-registry for `--pd-titlebar-bg` in case of Windows to set `#202020` color

packages/main/src/plugin/color-registry.ts
gastoner: We could use - hover:charcoal-300 for light mode (Linux has those colors)

gastoner:
> @gastoner I still don't see an override in color-registry for `--pd-titlebar-bg` in case of Windows to set `#202020` color

I've added titlebar-bg for windows to color-refistry and now the windows should use it

packages/main/src/plugin/color-registry.ts
```diff
@@ -322,6 +322,11 @@ export class ColorRegistry {
       light: colorPalette.gray[50],
     });
 
+    this.registerColor('titlebar-windows-bg', {
+      dark: '#202020',
+      light: colorPalette.gray[50],
+    });
```
benoitf: shouldn't we set these values into `titlebar-bg`

but adding a condition like
```
dark : isWindows() ? '#202020' : colorPalette.something
```

so there is only one definition of the color, just that the default value change based on the current Operating System
the svelte component just read the value

packages/main/src/plugin/color-registry.ts
gastoner: Light theme hover on non-exit buttons: gray-400:

https://github.com/user-attachments/assets/9fc5a230-24b5-4881-8cf4-f8cc6c0074b1

packages/main/src/plugin/color-registry.ts
gastoner: @benoitf Like so?

benoitf (approved):
yes thanks

benoitf:
I tested on macOS so probably need someone to validate on Windows and Linux

gastoner:
> I tested on macOS so probably need someone to validate on Windows and Linux

I was developing this on Windows ^^ screenshots/videos.
I can try it on WSL Linux too

benoitf:
> I was developing this on Windows ^^ screenshots/videos.

true but usually someone else need to validate the work. (let say you forgot to push a file, it may work for you but only you)
That said, I'm not saying that we can't trust the authors ;-)

feloy (commented):
I cannot test on Windows, but lgtm on Mac and Linux

packages/main/src/plugin/color-registry.spec.ts
```diff
@@ -433,7 +434,7 @@ describe('registerExtensionThemes', () => {
     expect(colors).toBeDefined();
     titlebarBg = colors.find(c => c.id === 'titlebar-bg');
     expect(titlebarBg).toBeDefined();
-    expect(titlebarBg?.value).toBe('#0f0f11');
+    expect(titlebarBg?.value).toBe(isWindows() ? '#202020' : '#0f0f11');
```
benoitf: here I think the unit test should mock the windows condition

like we should have 2 tests,
one checking the expected value when it'll return Windows and one for the other

Else it means that if I'm on mac/Linux, the test will pass locally but only fails on the CI environment, while I would expect that I'm able to check the non-regression even if I'm on Linux/macOS

gastoner: It looks like it has passed, but I can split it to Windows and others test
Something like: 
`if (isWindows()) {test(windowsTest)} 
else {test(otherTest)}`

benoitf: it should be something like:

```
test('check on Windows', () => {
mock util to return isWindows to true
check value
})

test('check on macOs/Linux', () => {
mock util to return isWindows to false
check value
})
```

benoitf (approved):

packages/main/src/plugin/color-registry.spec.ts
```diff
@@ -30,6 +30,8 @@ import type { ColorDefinition } from '/@api/color-info.js';
 import type { RawThemeContribution } from '/@api/theme-info.js';
 
 import colorPalette from '../../../../tailwind-color-palette.json';
+import { isWindows } from '../util.js';
+import * as util from '../util.js';
```
benoitf: I think there is a linting rule to only have one line when importing from the same location

gastoner: Oh my bad, fixing right away!

gastoner: And I've also missed the test on the bottom

gastoner: @benoitf could you please check?

benoitf (approved):
still approving

lstocchi (approved):
Tested on Windows. Works fine. GJ!!

PR #8617, opened 2024-08-27
test: deploy container to the kind cluster scenario

### LINKED ISSUE #8053: Deploy a pod to a Kubernetes (kind cluster)

amisskii:

### What does this PR do?

- Deploying a pod/container for a Kubernetes cluster via Deploy To Kubernetes option.

- Verification of the deployed app.

https://github.com/containers/podman-desktop/issues/8053


jeffmaury (approved):
LGTM

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -121,21 +128,43 @@ describe('Kind End-to-End Tests', async () => {
...
       });
       test('Validate correct Kubernetes context is selected', async () => {
         await statusBar.validateKubernetesContext(KUBERNETES_CONTEXT);
       });
+      describe('Deploy a container to the Kind cluster', async () => {
+        test('Pull an image and start a container', async () => {
+          const imagesPage = await navigationBar.openImages();
+          const pullImagePage = await imagesPage.openPullImage();
+          await pullImagePage.pullImage(IMAGE_TO_PULL);
+          await imagesPage.waitForImageExists(IMAGE_NAME, 10000);
```
cbr7: Shouldn't this be asserted on?

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -121,21 +128,43 @@ describe('Kind End-to-End Tests', async () => {
...
+          await imagesPage.startContainerWithImage(IMAGE_NAME, CONTAINER_NAME);
```
cbr7: Shouldn't this be validated that the container exists and starts correctly?

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -121,21 +128,43 @@ describe('Kind End-to-End Tests', async () => {
...
+        });
+        test('Deploy the container ', async () => {
+          const containerDetailsPage = new ContainerDetailsPage(page, CONTAINER_NAME);
+          await playExpect(containerDetailsPage.heading).toBeVisible();
+          await playExpect(containerDetailsPage.deployButton).toBeVisible();
```
cbr7: AS far as I know the method startContainersWithImage method returns a ContainersPage object, which indicates that is the page that should be currently navigated on, yet I see you are creating a container details page here and checking for location, is this correct? What ensure the navigation from containers page to container details page?

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -121,21 +128,43 @@ describe('Kind End-to-End Tests', async () => {
...
+          await containerDetailsPage.deployButton.click();
+          const deployToKubernetesPage = new DeployToKubernetesPage(page);
```
cbr7: I would prefer for this to be encapsulated and exposed as a method from the containers details page that handles the button clicking, navigation and returns a DeployToKubernetesPage object.

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -121,21 +128,43 @@ describe('Kind End-to-End Tests', async () => {
...
+          await deployToKubernetesPage.deployPod(CONTAINER_NAME, KUBERNETES_CONTEXT);
+        });
+        test('Verify the deployed application', async () => {
+          const podPage = await navigationBar.openPods();
+          console.log(DEPLOYED_POD_NAME);
+          await playExpect.poll(async () => podPage.deployedPodExists(DEPLOYED_POD_NAME)).toBeTruthy();
+        });
+      });
```
cbr7: I don't really understand why validating the pod creation needs to be its own test, why not check that at the end of the test that actually does the deploying?

cbr7 (changes requested):
Left a couple of comments.

tests/playwright/src/specs/kind.spec.ts
amisskii: You are right; the `startContainerWithImage` method returns a `ContainersPage` object. However, when we start a container with an image manually, we navigate to the container details page. Therefore, I think the `startContainerWithImage` method needs to be refactored to return a `ContainerDetailsPage` instead.

amisskii: WDYT @cbr7 ?

cbr7: @amisskii are you certain this is correct? Because there are many tests that use this method and I would expect to be broken if that were true. But if the flow did indeed change then certainly the method should be refactored to return the correct next page yes.

amisskii: At least that is how it works for me :) I will open a new pr.

cbr7: @amisskii I think there might be a difference on if you start the container with interactive terminal or not, if you have terminal it would lead to the details page but if there is no terminal it will navigate the the containers page.

If the navigation would always lead to the details page there would be some tests in the container smoke test suite that would fail.

cbr7 (approved):

tests/playwright/src/model/pages/container-details-page.ts
```diff
@@ -40,6 +42,7 @@ export class ContainerDetailsPage extends DetailsPage {
     this.stopButton = this.controlActions.getByRole('button').and(this.page.getByLabel('Stop Container'));
     this.deleteButton = this.controlActions.getByRole('button').and(this.page.getByLabel('Delete Container'));
     this.imageLink = this.header.getByRole('link', { name: 'Image Details' });
+    this.deployButton = this.controlActions.getByRole('button').and(this.page.getByLabel('Deploy to Kubernetes'));
```
odockal: `this.getByRole('button', { name: 'Deploy to Kubernetes' })` should works just fine, no need for `.and`

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -106,6 +106,7 @@ export class CreateKindClusterPage extends BasePage {
   private async createCluster(timeout: number = 120000): Promise<void> {
     await playExpect(this.clusterCreationButton).toBeVisible();
     await this.clusterCreationButton.click();
+    await this.logsButton.click();
```
odockal, line 115: maybe also add `scrollIntoView` function which will move the focus no the element.

tests/playwright/src/model/pages/deploy-to-kubernetes-page.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { BasePage } from './base-page';
+
+export class DeployToKubernetesPage extends BasePage {
+  readonly content: Locator;
+  readonly podName: Locator;
+  readonly kubernetesContext: Locator;
+  readonly deployButton: Locator;
+  readonly doneButton: Locator;
+  readonly namespaceCombobox: Locator;
```
odockal, line 29: I am missing ingrees/routes, here. And other options. Select a port, etc.

tests/playwright/src/model/pages/pods-page.ts
```diff
@@ -80,4 +80,15 @@ export class PodsPage extends MainPage {
     }
     return row.getByRole('button', { name: 'kebab menu', exact: true });
   }
+
+  public async deployedPodExists(podName: string): Promise<boolean> {
```
odockal: returns true only if pod is deployed to a cluster? I would add a param, what env. is expected for the pod to be in. I am guessing that also some ambiguity is possible when having pod in podman env. and in kubernetes?

odockal (changes requested):

tests/playwright/src/model/pages/create-kind-cluster-page.ts
amisskii: For some reason it doesn't work
```
TimeoutError: locator.scrollIntoViewIfNeeded: Timeout 30000ms exceeded.
Call log:
  - waiting for getByRole('form', { name: 'Properties Information' }).getByRole('button', { name: 'Show Logs' })
```
But it works fine without scrolling :)

tests/playwright/src/model/pages/pods-page.ts
amisskii: Since a pod name consists of the pod name itself, the control plane where it is deployed, and the namespace, I don't think ambiguity is possible.
`const DEPLOYED_POD_NAME: string = `${CONTAINER_NAME} ${KIND_CONTAINER_NAME} ${NAMESPACE}`;`

amisskii:
https://github.com/containers/podman-desktop-e2e/actions/runs/10678957214

tests/playwright/src/model/pages/pods-page.ts
```diff
@@ -80,4 +80,15 @@ export class PodsPage extends MainPage {
     }
     return row.getByRole('button', { name: 'kebab menu', exact: true });
   }
+
+  public async deployedPodExists(podName: string, environment: string): Promise<boolean> {
```
odockal: default could be Podman for the environment, right?

odockal (changes requested):
Please change a little the PR as discussed. Add locators for parent div containers in create-kind-cluster page. And also create new spec for deploy-to-kubernetes tests.

tests/playwright/src/model/pages/deploy-to-kubernetes-page.ts
```diff
@@ -0,0 +1,69 @@
...
+    const currentNamespace = await this.namespaceCombobox.inputValue();
+    if (currentNamespace !== namespace) {
+      const namespaceOptions = await this.namespaceCombobox.locator('option').allInnerTexts();
+      if (namespaceOptions.includes(namespace)) {
+        await this.namespaceCombobox.selectOption({ value: namespace });
+        await playExpect(this.namespaceCombobox).toHaveValue(namespace);
+      } else {
+        throw new Error(`${namespace} doesn't exist`);
+      }
+    }
```
odockal, line 104: Please also consider adding handling of the checkboxes like Security context, K. Services and Exposing the service/creating ingress.

tests/playwright/src/model/pages/deploy-to-kubernetes-page.ts
```diff
@@ -0,0 +1,119 @@
...
+      if (!namespaceOptions.includes(namespace)) {
+        throw new Error(`${namespace} doesn't exist`);
+      }
+      await this.namespaceCombobox.selectOption({ value: namespace });
+      await playExpect(this.namespaceCombobox).toHaveValue(namespace);
+    }
+
+    await playExpect(this.deployButton).toBeEnabled();
+    await this.deployButton.click();
+    await playExpect(this.doneButton).toBeVisible({ timeout: 30000 });
```
odockal: please make it a parameter with default value. I think that 30 sec. wont be enough.

tests/playwright/src/utility/operations.ts
```diff
@@ -241,3 +244,74 @@ export async function getVolumeNameForContainer(page: Page, containerName: strin
...
+  await kindResourceCard.performConnectionAction(ResourceElementActions.Stop);
+  await playExpect(kindResourceCard.resourceElementConnectionStatus).toHaveText(ResourceElementState.Off, {
+    timeout: 50000,
+  });
+  await kindResourceCard.performConnectionAction(ResourceElementActions.Delete);
+  await playExpect(kindResourceCard.markdownContent).toBeVisible({ timeout: 50000 });
+  const containersPage = await navigationBar.openContainers();
+  await playExpect.poll(async () => containersPage.containerExists(containerName), { timeout: 10000 }).toBeFalsy();
+
+  await page.waitForTimeout(2000);
```
odockal: do we need a hard wait? If yes, try to find some other on which we can safely wait for. ie. toBeVisible or something. You can use waitUntil/While.

odockal (approved):
LGTM, just minor changes please.

PR #8618, opened 2024-08-28
chore: close tooltip when accounts menu open

SoniaSandler:

### What does this PR do?
When the accounts menu opens, the tooltip of the buddy icon disappears and only reappears on hover when the menu is closed

### Screenshot / video of UI

[Screencast from 2024-08-27 19-57-46.webm](https://github.com/user-attachments/assets/746af548-6d9c-4c86-a3e3-ebba13f56798)

Related to https://github.com/containers/podman-desktop/issues/8335

### How to test this PR?

Click on the buddy icon and check that the tooltip doesn't show when the dropdown menu is open

### DIFF
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -23,6 +23,7 @@ let { exitSettingsCallback, meta = $bindable() }: { exitSettingsCallback: () => 
 
 let authActions = $state<AuthActions>();
 let outsideWindow = $state<HTMLDivElement>();
+let showAccountsTooltip = $state(true);
 
 const iconSize = '22';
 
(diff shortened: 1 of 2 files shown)


gastoner (approved):
Works well!

benoitf (commented):
I commented in another PR, but IMHO it's the Tooltip component responsibility to be hidden if some events are coming

(and then we can send events when context menu is being closed or opened)

SoniaSandler:
@benoitf following a recent PR https://github.com/containers/podman-desktop/pull/8269, the accounts menu is not a context menu but a `DropDownMenu`

benoitf:
it's still the same global issue pattern.

Opening or Closing dropdown send events, and Tooltip should react directly (or indirectly to have a common state)


AuthActions should not be involved, only Tooltip (else let say tomorrow you add a new dropdown at another location, you'll need to handle that usecase and it'll lead to errors)

benoitf:
components should be coupled only for what they're exchanging/providing, the remaining part (out of the concern/scope of the direct component) should be loosely coupled

PR #8650, opened 2024-08-29
chore: switch playwright tests to use playwright test runner instead of vitest

### LINKED ISSUE #8630: Switch from vitest test runner to playwright

cbr7:
### What does this PR do?
switches e2e tests to run on playwright test runner instead of vitest

https://github.com/containers/podman-desktop/issues/8630


tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
```diff
@@ -16,11 +16,8 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { afterEach } from 'node:test';
```
benoitf, line 19: interesting here, it means that the afterEach is probably never called today

cbr7: why would that be?

benoitf: vitest analyze only the vitest afterEach

cbr7: oh I see, that is interesting.

cbr7:
Closing PR for now to not spam the cicd infrastructure.

cbr7:
I also ran the full e2e test set [here](https://github.com/containers/podman-desktop/actions/runs/10630075870/job/29468239099)

benoitf:
@cbr7 could you rebase your branch, there was an issue in the website build that is now fixed

benoitf (approved):

odockal:
@cbr7 On the first look it looks good. I am interested in the running the new test runner in windows and mac, have you tried?

cbr7:
@odockal I tried but it didn't work because the windows workflows are still using yarn instead of pnpm so it throws an error.

Unless you mean locally, in which case yes, I ran them on my macbook.

PR #8651, opened 2024-08-29
feat: add default button action for containers empty state

### LINKED ISSUE #8568: Default action for Containers's empty state
There should be two options in the following screen to allow the user to quickly get started:
Screenshot 2024-08-25 at 12 49 43

The first one should be a button:
"Start your first container"

Then in the screen, we display a label "OR" and the line bellow is the actual commande line we already have in the screen

gastoner:
### What does this PR do?
Adds button to Container empty state screen 
=> User can use button/command to do the same action

Other pages (Image + Pods + Volumes) would probably use the same EmptyScreen slot

### Screenshot / video of UI

https://github.com/user-attachments/assets/d0f049ae-c976-4d88-8738-1e446fcae732

Closes #8568 

### How to test this PR?

-Delete all Containers
-Assert that you see the button 

### DIFF
--- a/packages/renderer/src/lib/container/ContainerEmptyScreen.svelte
+++ b/packages/renderer/src/lib/container/ContainerEmptyScreen.svelte
@@ -1,13 +1,22 @@
 <script lang="ts">
-import { EmptyScreen } from '@podman-desktop/ui-svelte';
+import { Button, EmptyScreen } from '@podman-desktop/ui-svelte';
 import { ContainerIcon } from '@podman-desktop/ui-svelte/icons';
+
+import { providerInfos } from '../../stores/providers';
 
 export let runningOnly: boolean;
 export let stoppedOnly: boolean;
+const helloImage: string = 'quay.io/podman/hello:latest';
 
+$: inProgress = false;
 $: title = getTitle(runningOnly, stoppedOnly);
-$: message = getMessage(stoppedOnly);
+$: messageCommandLine = getMessageCommandLine(stoppedOnly);
+$: messageButton = getMessageButton(stoppedOnly);
 $: commandLine = getCommandLine(stoppedOnly);
+$: selectedProviderConnection = $providerInfos
+  .map(provider => provider.containerConnections)
+  .flat()
+  .find(providerContainerConnection => providerContainerConnection.status === 'started');
 
 function getTitle(runningOnly: boolean, stoppedOnly: boolean): string {
   if (runningOnly) {
@@ -18,11 +27,18 @@ function getTitle(runningOnly: boolean, stoppedOnly: boolean): string {
     return 'No containers';
   }
 }
-function getMessage(stoppedOnly: boolean): string {
+function getMessageCommandLine(stoppedOnly: boolean): string {
   if (stoppedOnly) {
     return '';
   } else {
     return 'Run a first container using the following command line:';
+  }
+}
+function getMessageButton(stoppedOnly: boolean): string {
+  if (stoppedOnly) {
+    return '';
+  } else {
+    return 'Run a first container by clicking on this button:';
   }
 }
 function getCommandLine(stoppedOnly: boolean): string {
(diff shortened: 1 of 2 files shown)


feloy:
I think that the user will want some feedback between the time it clicks the button and the time the container is started, to understand that the operation is in progress. For example setting the button in progress mode?

packages/renderer/src/lib/container/ContainerEmptyScreen.svelte
```diff
@@ -18,25 +26,66 @@ function getTitle(runningOnly: boolean, stoppedOnly: boolean): string {
...
   }
 }
+async function runContainer(commandLine: string) {
+  try {
+    inProgress = true;
+    const helloImage = 'quay.io/podman/hello:latest';
+    if (selectedProviderConnection) {
+      await window.pullImage(selectedProviderConnection, helloImage, () => {});
+      const listImages = window.listImages();
+      const image = (await listImages).find(item => item.RepoTags?.includes(helloImage));
```
feloy:
```suggestion
      const listImages = await window.listImages();
      const image = listImages.find(item => item.RepoTags?.includes(helloImage));
```

packages/renderer/src/lib/container/ContainerEmptyScreen.svelte
```diff
@@ -18,25 +26,66 @@ function getTitle(runningOnly: boolean, stoppedOnly: boolean): string {
...
+      if (image) {
+        await window.createAndStartContainer(image.engineId, { Image: helloImage });
+      } else {
+        console.error(`Could not find ${helloImage} in images`);
```
feloy: you could use a dialog box, as the logs won't appear to the _normal_ user

packages/renderer/src/lib/container/ContainerEmptyScreen.svelte
```diff
@@ -18,25 +26,66 @@ function getTitle(runningOnly: boolean, stoppedOnly: boolean): string {
...
+      }
+      window.telemetryTrack('startFirstContainerByButton');
+    }
+  } catch (err) {
+    console.error(`Error while executing ${commandLine}: ${String(err)}`);
```
feloy: same here

feloy:
At the beginning, the EmptyScreen component was very constrained, with only icon/title/message/detail/commandline.

After, has been introduced the default slot (free form) placed at the bottom of the component.

With these changes, we are adding a named slot, to add its (free form) content between the title and the detail.

To add more consistency to the component, I would consider:
- either making an inventory of the usages we do of the component, and find an abstraction to keep it constrained as it was at the beginning
- or making it mostly completely free-form, with only an icon/title as parameters

@slemeur @jeffmaury @ekidneyrh

gastoner:
> At the beginning, the EmptyScreen component was very constrained, with only icon/title/message/detail/commandline.
> 
> After, has been introduced the default slot (free form) placed at the bottom of the component.
> 
> With these changes, we are adding a named slot, to add its (free form) content between the title and the detail.
> 
> To add more consistency to the component, I would consider:
> 
>  either making an inventory of the usages we do of the component, and find an abstraction to keep it constrained as it was at the beginning
> 
>  or making it mostly completely free-form, with only an icon/title as parameters
> 
> 
> @slemeur @jeffmaury @ekidneyrh

Thanks Philippe! Personally I would suggest the second option, make it free form with icon and title as parameters

jeffmaury (approved):
LGTM but I think the slot should be named differently as it's more that a button

jeffmaury:
Maybe a tooltip on the button saying that it's will put quay.io/podman/hello and start it would be good

gastoner:
@jeffmaury e.g. "Pull quay.io/podman/hello:latest image and start container" and slot could be named "upperContent"?

packages/preload/vite.config.js.timestamp-1725005385529-5571b54447365.mjs
```diff
@@ -0,0 +1,84 @@
+// .electron-vendors.cache.json
```
feloy: I suppose this file should be removed

feloy (changes requested):

packages/preload/vite.config.js.timestamp-1725005385529-5571b54447365.mjs
gastoner: @feloy removed!

feloy (approved):
LGTM

benoitf:
looking at another PR linking this one, I see that no unit tests were provided for this new feature while it should

PR #8653, opened 2024-08-29
chore(tests): add podman desktop update e2e test

### LINKED ISSUE #7931: Implement the E2E update path of the Podman Desktop

We are seeing a problem with electron when updating the podman desktop to the latest released version on windows: https://github.com/containers/podman-desktop/issues/7114.

We need to came up with a automated test that will make sure we know if we can actually perform the update or not.

* checkout main upstream branch
* change the version in `package.json` to 1.0.0
* build the app from the codebase
* start up the e2e tests (testing the update path)
* check that we have option to update to the newest podman-desktop version (latest tag on github) - which is actually a version below what we are testing
* update
* check that update was sucessfull
* We might end the test once we get "Do you want to restart?" dialog, since we could probably not be able to let app to get restarted (test would crash)
* Alternative would be to start another instance in the next step and check the version and let GHA/CI to evaluate the result, not the tests

_No response_

### Additional context

_No response_

odockal:
### What does this PR do?
Brings new E2E update test. The test is intended to be run solely in the specific test environment. Windows (now) and Mac will be target env. 

The update e2e test will run after each merge, and in the PR check if it was initialized by dependabot where we saw the most problems.

#7931 

### How to test this PR?
On windows only now. See the details in the linked issue. You need change version of PD in package.json to something like `1.0.0` and then run `yarn test:e2e:update`. Also, if you want to perform actual update during the testing, add env. var: `UPDATE_PODMAN_DESKTOP=true`
In order to be able to run the update in locally built podman desktop, you need to set `ELECTRON_ENABLE_INSPECT` env. var. and build the bits to have correct artifacts: `pnpm compile:current --win nsis`. It will produce executable podman desktop binary, which will offer an update option. In order to execute the e2e tests now with built executable, use `PODMAN_DESKTOP_BINARY="C:\path\to\dist\win-unpacked\Podman Desktop.exe"` when running test target.


odockal:
@ScrewTSW In this pr there is the example of how to build and use the compiled binary with playwright.

xbabalov (approved):

tests/playwright/src/specs/image-smoke.spec.ts
```diff
@@ -51,6 +51,7 @@ test.beforeAll(async () => {
 });
 
 test.afterAll(async () => {
+  // test
```
cbr7: Does this comment need to be here? Maybe it's an artifact from some previous debugging?

cbr7 (approved):

.github/workflows/pr-check-update-e2e.yaml
```diff
@@ -0,0 +1,126 @@
...
+  pull_request:
+    paths:
+      - '**'
+      - 'pnpm-lock.yaml'
+    types: [labeled, synchronize, opened, reopened]
+
+jobs:
+  detect_pnpm_changes:
+    name: Detect pnpm lock file changes
+    runs-on: ubuntu-latest
```
benoitf:
```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/pr-check-update-e2e.yaml
```diff
@@ -0,0 +1,126 @@
...
+            echo "PNPM_LOCK_CHANGED=true" >> $GITHUB_OUTPUT
+          else
+            echo "PNPM_LOCK_CHANGED=false" >> $GITHUB_OUTPUT
+          fi
+
+  run-update-e2e-test:
+    name: win update e2e tests
+    needs: detect_pnpm_changes
+    if: contains(github.event.pull_request.labels.*.name, 'area/update') || needs.detect_pnpm_changes.outputs.pnpm_lock_changed == 'true'
+    runs-on: ${{ matrix.os }}
```
benoitf: why do we have a matrix if title and os are hardcoded to windows ?

benoitf: if matrix is used, name should use matrix.os and not win

odockal: I was a step ahead, will keep it for windows for now.

.github/workflows/pr-check.yaml
```diff
@@ -324,3 +326,98 @@ jobs:
...
+    steps:
+      - name: Evaluate the Windows Update test results
+        run: | 
+          echo "Windows updater result: ${{ needs.run-update-e2e-test.result }}"
+          if [ "${{ needs.run-update-e2e-test.result }}" = "failure" ]; then
+            echo "Windows udpater test failed..."
+            exit 1
+          else
+            echo "Windows updater test succeeded or was skipped..."
+          fi
```
ScrewTSW: missing new line

ScrewTSW (approved):
LGTM

benoitf (approved):

.github/workflows/pr-check.yaml
```diff
@@ -324,3 +326,98 @@ jobs:
...
+        with:
+          name: win-update-e2e-test
+          path: |
+            ./tests/**/output/
+            ./tests/**/output/junit*.xml
+            !./tests/**/traces/raw
+
+  update-e2e-test:
+    name: update E2E test
+    runs-on: ubuntu-22.04
```
benoitf: why not 24 04

tests/playwright/src/specs/installation/update-install.spec.ts
```diff
@@ -0,0 +1,94 @@
+/**********************************************************************
+ * Copyright (C) 2023-2024 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

.github/workflows/pr-check.yaml
odockal: mistake, updated.

tests/playwright/src/specs/installation/update-install.spec.ts
odockal: fixed, thx.

PR #8664, opened 2024-08-30
feat: add telemetry for usage of podman machine with libkrun

gastoner:
### What does this PR do?
Adds telemetry for creating podman machine against libkrun

Closes https://github.com/containers/podman-desktop-internal/issues/313

### DIFF
--- a/extensions/podman/src/extension.ts
+++ b/extensions/podman/src/extension.ts
@@ -1860,6 +1860,8 @@ export async function createMachine(
     provider = getProviderByLabel(params['podman.factory.machine.provider']);
   }
 
+  telemetryRecords.provider = provider;
+
   // cpus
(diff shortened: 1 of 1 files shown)


axel7083 (commented):
Could we have a test ?

benoitf (changes requested):
it needs to be moved to the if block like for other properties

also it could be undefined


I suppose we need to remove these values:
https://github.com/containers/podman-desktop/blob/5ad050d452fb8a2aa708237652c94db7500ba86c/extensions/podman/src/extension.ts#L1805-L1815

and provides the version of vfkit and libkrunkit as well (here it can be follow-up actions)

feloy:
replaced by #8680

PR #8672, opened 2024-08-30
chore: always build the website on PR checks

benoitf:
### What does this PR do?
now, there are some scripts to generate css from color registry if this class is modified it should also triggers the website build but it may have changed in some dependencies, etc.

so might be safer to always try to run the build

ie: this morning all PRs started to fail due to a change in one color-registry dependency

### How to test this PR?


benoitf:
it depends on https://github.com/containers/podman-desktop/pull/8671 before being successful

gastoner (approved):

PR #8680, opened 2024-08-30
feat: send telemetry related to krunkit

feloy:

### What does this PR do?

On Mac, send telemetry related to krunkit when creating a Podman machine

When a machine is created with libkrun, new parameters:

```
event: podman-desktop.podman.podman.machine.init
{
  "krunkitPath": "/opt/podman/bin",
  "krunkitVersion": "0.1.2",
  "provider": "libkrun",
}
```

When a machine (created with libkrun) is started, new parameters:
```
event: podman-desktop.podman.podman.machine.start
{
  "krunkitPath": "/opt/podman/bin",
  "krunkitVersion": "0.1.2",
  "provider": "libkrun",
}
```

When a machine (created with libkrun) is stopped, new event:
```
event: podman-desktop.podman.podman.machine.stop
{
  "krunkitPath": "/opt/podman/bin",
  "krunkitVersion": "0.1.2",
  "provider": "libkrun",
  [...]
}
```

When a machine is created with hyperV, new parameters:

```
event: podman-desktop.podman.podman.machine.init
{
  "provider": "applehv",
}
```

When a machine (created with hyperV) is started, new parameters:
```
event: podman-desktop.podman.podman.machine.start
{
  "provider": "applehv",
}
```

When a machine (created with hyperV) is stopped, new event:
```
event: podman-desktop.podman.podman.machine.stop
{
  "provider": "applehv",
  [...]
}
```

Fixes https://github.com/containers/podman-desktop-internal/issues/313

### How to test this PR?


benoitf (commented):
- we will miss vfkit data like the version, etc (could be a separate issue)

- what is happening if in my containers.conf, I'm using libkrun as the default provider
 should we send that data in telemetry (like user has changed its default provider to be libkrun in its containers.conf file)
b/c if we create libkrunkit VM and containers.conf is vfkit or missing, it means that the CLI won't list the machines, so maybe we should report this kind of data ?

feloy:
> * we will miss vfkit data like the version, etc (could be a separate issue)

> * what is happening if in my containers.conf, I'm using libkrun as the default provider
>   should we send that data in telemetry (like user has changed its default provider to be libkrun in its containers.conf file)
>   b/c if we create libkrunkit VM and containers.conf is vfkit or missing, it means that the CLI won't list the machines, so maybe we should report this kind of data ?

I think the information should be displayed to the user first (for example, by setting the pre-selected provider type as the one in the containers.conf in the machine creation form, and alerting the user that the machine won't be accessible with the cli when it selects the other provider type)

benoitf:
> I think the information should be displayed to the user first (for example, by setting the pre-selected provider type as the one in the containers.conf in the machine creation form, and alerting the user that the machine won't be accessible with the cli when it selects the other provider type)

But I was meaning on the telemetry side, probably we need to report if containers.conf is set to libkrun by default or not, etc

feloy:
> > I think the information should be displayed to the user first (for example, by setting the pre-selected provider type as the one in the containers.conf in the machine creation form, and alerting the user that the machine won't be accessible with the cli when it selects the other provider type)
> 
> But I was meaning on the telemetry side, probably we need to report if containers.conf is set to libkrun by default or not, etc

Yes I understand, but my feeling is that this information would be meaningless if the user is not informed correctly

benoitf:
> Yes I understand, but my feeling is that this information would be meaningless if the user is not informed correctly

yes but it may depend if 90% of users are impacted, 50% or 0.1%

feloy:
> > Yes I understand, but my feeling is that this information would be meaningless if the user is not informed correctly
> 
> yes but it may depend if 90% of users are impacted, 50% or 0.1%

OK, I'll add the information of the machine provider defined (or not) in containers.conf

axel7083 (commented):
On Linux the machine is still using qemu, maybe we should keep it and add a section for linux users ?

We have a section for mac and windows but none for linux

extensions/podman/src/vfkit-helper.ts
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import * as extensionApi from '@podman-desktop/api';
+
+export class VfkitHelper {
```
axel7083: Those helper all looks the same, can this be generify ?

extensions/podman/src/extension.ts
```diff
@@ -1802,16 +1804,26 @@ function sendTelemetryRecords(
...
 
-      // add qemu version
+      // add krunkit version
       try {
-        const qemuVersion = await qemuHelper.getQemuVersion(qemuPath);
-        if (qemuPath) {
-          telemetryRecords.qemuPath = qemuPath;
+        const krunkitVersion = await krunkitHelper.getKrunkitVersion(krunkitPath);
+        if (krunkitPath) {
+          telemetryRecords.krunkitPath = krunkitPath;
```
jeffmaury: I don't understand why it's call krunkit path and not podman path ?

jeffmaury (commented):
Also seems we record information only when creating the machine would be good to record only on start/stop time ?

feloy:
> > > Yes I understand, but my feeling is that this information would be meaningless if the user is not informed correctly
> > 
> > 
> > yes but it may depend if 90% of users are impacted, 50% or 0.1%
> 
> OK, I'll add the information of the machine provider defined (or not) in containers.conf

@benoitf I had to add some indirection when calling createMachine, to add the `podmanConfiguration` parameter, but the `createMachine` being fixed (as part of the API). The pattern used is very common in Golang, but I don't know if there is a preferred one in TS to do the same thing

extensions/podman/src/extension.ts
feloy: It is the same path as podman (as it is the path where all the binaries installed by the podman installer), but it is intended to be used to find the krunkit binary

extensions/podman/src/vfkit-helper.ts
feloy: I think there is a little difference between qemu/krunkit and vfkit, as the first ones are installed by podman, and the other one is installed by the user. And this makes some difference for the helpers. But I would like to have confirmation before to make it generic.

feloy:
> On Linux the machine is still using qemu, maybe we should keep it and add a section for linux users ?
> 
> We have a section for mac and windows but none for linux

I would prefer to focus on Mac for this PR

axel7083:
> I would prefer to focus on Mac for this PR

Yeah but you are removing code that can be used on linux.

slemeur:
I think the scope of the PR should remain on reporting the telemetry for krunkit podman machine.
We want to be able to track when:
- a podman machine is getting created from podman desktop, using krunkit
- a podman machine is started/stopped from podman desktop using krunkit.

For the sake of simplifying things in this PR:
- we shouldn't change what we are reporting for other OS.
- we shouldn't try to catch default containers.conf settings. I think this should be a separate effort, as I'm not sure we are reporting any information with regards to configuration applied in that file (as of today).

feloy:
> I think the scope of the PR should remain on reporting the telemetry for krunkit podman machine. We want to be able to track when:
> 
> * a podman machine is getting created from podman desktop, using krunkit
> * a podman machine is started/stopped from podman desktop using krunkit.
> 

For now, no event is sent to the telemetry when a podman machine is stopped. This will add a new event `podman-desktop.podman.podman.machine.stop`, with the information similar to the one in `start`. Is it ok @slemeur ?

jeffmaury (commented):
LGTM.

But sending the label instead of the raw vmType as the provider seems strange to me as it is subject to change as it is used in the UI and vmType is more stable

feloy:
> LGTM.
> 
> But sending the label instead of the raw vmType as the provider seems strange to me as it is subject to change as it is used in the UI and vmType is more stable

This is because only the label is registered in the machine information on started machines. But #8723 seems to have changed that, I'll try to replace the label with the code.

feloy:
> > LGTM.
> > But sending the label instead of the raw vmType as the provider seems strange to me as it is subject to change as it is used in the UI and vmType is more stable
> 
> This is because only the label is registered in the machine information on started machines. But #8723 seems to have changed that, I'll try to replace the label with the code.

@jeffmaury it should be ok now

extensions/podman/src/qemu-helper.ts
```diff
@@ -1,56 +0,0 @@
-/**********************************************************************
```
axel7083: Could we keep those files even if they are not used ? so they can be used in another PR for Linux ?

jeffmaury (approved):
LGTM

axel7083 (approved):
LGTM Codewise

PR #8685, opened 2024-08-30
fix: number widget should accept float value for number type

### LINKED ISSUE #8663: number item widget is not allowing float values
### Bug description

when specifying the schema value of a configuration we can choose between integer and number

if integer then it should be 1 or 1.0 (only .0 is allowed) and for number it can be anything

but the widget is not letting us enter the `.` key

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

try to define a value with number value and try like `1.5`. You won't see the . being displayed and you can either enter 1 or 15 but . is not being accepted

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
The widget Number should handle both integer and number values and validate the values as it's expected from JSON schema

Integer for example 1 and 1.0 are accepted
number: 1, 3.14 etc are valid

Today we can't use `.` character at all

fixes https://github.com/containers/podman-desktop/issues/8663

### How to test this PR?

set/edit the type of a property to `integer` or `number` and check that you can use the `.` key and only `.0` for integers

### DIFF
--- a/packages/renderer/src/lib/ui/NumberInput.svelte
+++ b/packages/renderer/src/lib/ui/NumberInput.svelte
@@ -31,12 +31,18 @@ function validateNumber() {
 }
 
 function onKeyPress(event: any) {
-  // if type is integer it means we can only accept nu
-  // see https://json-schema.org/understanding-json-schema/reference/numeric
   // Numbers with a zero fractional part are considered integers
+  // see https://json-schema.org/understanding-json-schema/reference/numeric
 
-  // first, add the new character to the extisting value
-  const wantedValue = `${value}${event.key}`;
+  // get cursor position
+  const cursorPosition = event.target.selectionStart;
+
+  console.log('cursorPosition', cursorPosition);
+
+  // add the new character to the cursor position
+  const wantedValue = `${event.target.value.slice(0, cursorPosition)}${event.key}${event.target.value.slice(cursorPosition)}`;
+
+  console.log('wantedValue', wantedValue);
 
   // now, check if type is integer if the value is value without digits or with zero fractional part
   if (type === 'integer' && Number.isInteger(Number(wantedValue))) {
--- a/packages/renderer/src/lib/ui/NumberInput.svelte
+++ b/packages/renderer/src/lib/ui/NumberInput.svelte
@@ -37,12 +37,8 @@ function onKeyPress(event: any) {
   // get cursor position
   const cursorPosition = event.target.selectionStart;
 
-  console.log('cursorPosition', cursorPosition);
-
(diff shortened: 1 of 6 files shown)


gastoner (approved):

gastoner (approved):
Tested on this, but this is probably not related to this PR right? (Windows)
image

benoitf:
@gastoner https://github.com/containers/podman-desktop/issues/8648

packages/renderer/src/lib/ui/NumberInput.svelte
```diff
@@ -30,7 +31,20 @@ function validateNumber() {
 }
 
 function onKeyPress(event: any) {
-  if (isNaN(Number(event.key))) {
+  // if type is integer it means we can only accept nu
```
feloy: >   // if type is integer it means we can only accept nu
Is the sentence complete?

packages/renderer/src/lib/ui/NumberInput.svelte
```diff
@@ -30,7 +31,20 @@ function validateNumber() {
...
+  // see https://json-schema.org/understanding-json-schema/reference/numeric
+  // Numbers with a zero fractional part are considered integers
+
+  // first, add the new character to the extisting value
+  const wantedValue = `${value}${event.key}`;
```
feloy: The problem is that the test is done with the character added at the end, but the user can have placed the cursor anywhere.

For example for an integer input, if the input contains `12` and the cursor is between `1` and `2` and the user types `.`,  the character is accepted (because `12.` is considered as an integer), but the input contains `1.2` which is not an integer.

packages/renderer/src/lib/ui/NumberInput.svelte
benoitf: fixed

packages/renderer/src/lib/ui/NumberInput.svelte
benoitf: fixed

benoitf:
@feloy good catch, should be fixed now (test added also)

packages/renderer/src/lib/ui/NumberInput.svelte
feloy: Nice fix . Don't forget to remove the debug logs

feloy (approved):

packages/renderer/src/lib/ui/NumberInput.svelte
benoitf: fixed

PR #8690, opened 2024-09-02
feat: optional display name container connection

axel7083:
### What does this PR do?

Adding an optional `displayName` when registering connections. Allowing user friendly name.

E.g. podman extension register `Podman Machine` as display name, and keep `podman-machine-default` as name (real name).

ℹ️ The name will always be the fallback if the display name is not available
### Screenshot / video of UI

Places where the `displayName` is used

image

image

Need rebase after https://github.com/containers/podman-desktop/pull/8689

First part of https://github.com/containers/podman-desktop/issues/8484

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
@@ -471,7 +471,7 @@ function hasAnyConfiguration(provider: ProviderInfo) {
                 </Tooltip>
               </div>
               <div class="{container.status !== 'started' ? 'text-gray-900' : ''} font-semibold">
-                {container.name}
+                {container.displayName ?? container.name}
               </div>
               <div class="flex" aria-label="Connection Status">
                 <ConnectionStatus status={container.status} />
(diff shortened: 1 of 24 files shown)


packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -471,7 +471,7 @@ function hasAnyConfiguration(provider: ProviderInfo) {
                 </Tooltip>
               </div>
               <div class="{container.status !== 'started' ? 'text-gray-900' : ''} font-semibold">
-                {container.name}
+                {container.displayName ?? container.name}
```
benoitf: probably make it mandatory on the info object (there will always be a displayName from renderer side)

but then on the server side, if displayName is missing from the definition we set it to the name

so, no need to handle any special case from the UI side, it's always here

axel7083: I made the info object have the displayName mandatory, and keep optional on the api for backward compatibility

benoitf (approved):

PR #8700, opened 2024-09-02
feat(configuration): allow to define the value of step for Input widget 

### LINKED ISSUE #8662: widget to select number is always adding/decreasing the number by 1

there is a step value that we can define

it should increment with the value of the step if defined

handle step value for the increment instead of the hardcoded 1 value

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Allow to handle step value when doing increment/decrement in the number widget

depends on https://github.com/containers/podman-desktop/pull/8698 (will remove commits once PR is merged)

Fixes https://github.com/containers/podman-desktop/issues/8662

### How to test this PR?

Apply `step` configuration in JSON for a number type

### DIFF
--- a/packages/renderer/src/lib/preferences/item-formats/NumberItem.svelte
+++ b/packages/renderer/src/lib/preferences/item-formats/NumberItem.svelte
@@ -44,6 +44,7 @@ function onValidation(newValue: number, validationError?: string) {
     bind:error={error}
     aria-label={record.description}
     minimum={record.minimum}
+    step={record.step}
     type={record.type === 'integer' ? 'integer' : 'number'}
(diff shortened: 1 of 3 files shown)


feloy:
I suppose the step is only for incrementing/decrementing the value with the -/+ buttons, and is not taken into consideration for validating the value. If so, do we want to document it in some way?

benoitf:
I would say it's up to the author of the extension.
It's like if you specify that it's a string but you expect to be a number in your code or the opposite, etc.

feloy (approved):

benoitf:
rebased after dependent PR being merged

axel7083 (approved):

PR #8705, opened 2024-09-02
chore: fix float issue

benoitf:
### What does this PR do?
Due to rounding we may have unexpected values in the InputNumber widget (like 0.4999999999 rather than 0.5)

follow-up of https://github.com/containers/podman-desktop/issues/8662

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/NumberInput.spec.ts
+++ b/packages/renderer/src/lib/ui/NumberInput.spec.ts
@@ -253,3 +253,41 @@ test('Expect decrement works with custom step', async () => {
 
   expect(input).toHaveValue('4.2');
 });
+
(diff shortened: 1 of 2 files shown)


gastoner (commented):
If I get this right, there is an option that after inc/dec, the value can be, e.g., 49.9999 (but it should be 50).
If this is the case, wouldn't be better to use math.round to round the number?

benoitf:
@gastoner AFAIK Math.round is to round to the nearest integer (so it won't be a float anymore)
`Math.round(0.5) will be 0`

benoitf:
rebasing to get new PR checks running

gastoner (approved):

axel7083 (approved):
LGTM

PR #8710, opened 2024-09-02
chore(test): update __dirname usage and adds cfg path

### LINKED ISSUE #8711: Updated __dirname usage in the e2e test framework
### Bug description

After latest changes __dirname returns undefined in the e2e test framework, it needs to be fixed.

### Operating system

macOS

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Update __dianme usage and adds path to pw config file and adds type = module to pw package.json

https://github.com/containers/podman-desktop/issues/8711

### DIFF
--- a/tests/playwright/tsconfig.json
+++ b/tests/playwright/tsconfig.json
@@ -6,7 +6,7 @@
     "moduleResolution": "Node",
     "declaration": true,
     "outDir": "dist",
-    "rootDir": "src",
+    "rootDirs": ["src", "../../playwright.config.ts"],
     "skipLibCheck": true,
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,
--- a/tests/playwright/package.json
+++ b/tests/playwright/package.json
@@ -2,6 +2,7 @@
   "name": "@podman-desktop/tests-playwright",
   "version": "1.9.0-next",
   "description": "Playwright-based testing libraries for Podman Desktop and its extensions",
+  "type": "module",
   "publishConfig": {
     "access": "public"
   },
--- a/tests/playwright/src/specs/image-smoke.spec.ts
+++ b/tests/playwright/src/specs/image-smoke.spec.ts
@@ -17,6 +17,7 @@
  ***********************************************************************/
 
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 
 import type { Page } from '@playwright/test';
 import { expect as playExpect, test } from '@playwright/test';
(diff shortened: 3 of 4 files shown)


tests/playwright/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "moduleResolution": "Node",
     "declaration": true,
     "outDir": "dist",
-    "rootDir": "src",
+    "rootDirs": ["src", "../../playwright.config.ts"],
```
benoitf, line 9: shouldn't playwright config file be part of this package ?

cbr7: It's not really required, because we need a different playwright config in the extension solution anyway so when I created it initially I put it in the same place the vitest config file was located in order not to change too many things at once.

benoitf: yes but now I think it's time to move it ( we have one in argos folder)
And it does not make sense to have something at the root folder to reference files only in this tests/playwright folder

cbr7: @benoitf I just tried moving the config locally into `tests/playwright` and it becomes very complex very quickly. The issue is that in order for the config to be used it needs to be in the same folder from where the playwright test execution command is issued, which means that if the config is in `tests/playwright` we will also have to move all the relevant scripts from the root `package.json` file into the `package.json` in `tests/playwright`, but the build commands for the entire solution are still in the root file, so we'd have to have half of the commands in the root `package.json` and half in the `package.json` in `tests/playwright` even all that I was still not able to get it to run locally right now, since the main entry point into the application is not the same.

While I'm sure there is some configuration that makes it work like this, it's a lot more convoluted then is in scope for this PR.

jeffmaury: Can't the root package.json commands just delegate to this package.json ?

cbr7: @jeffmaury I'm not sure what you mean by delegate in this context, we can certainly run commands from the root `package.json` towards the `tests/playwright` `package.json` using the `npm run --prefix path/to/folder <command>` command, however the `<command>` would still have to be located in the `tests/playwright` `package.json` so it would still involve splitting all the scripts.

But in addition to that the `main` entry point into the app is not the same, when I use the one that is currently present in the `tests/playwright` the PD app doesn't even start up and if I replace and adapt it to the one that is in the root `package.json` I get errors like these:

image

benoitf (approved):
we can defer the move of the location out of the scope of this PR but we should definitely have a follow-up issue

cbr7:
@benoitf I created https://github.com/containers/podman-desktop/issues/8726 to track this change.

PR #8728, opened 2024-09-03
chore: split checks in smaller chunks

### LINKED ISSUE #8660: split PR check job to smaller jobs

For now, we have a PR check job doing linting, typechecking, unit tests, etc
But it may take too many time to get the result (linter's job is taking more and more time)
and if we do have linter that is failing, we don't have the result of the unit tests for example

Split into smaller chunks the PR check job doing the checks like linting, testing, typescript check, etc.

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
- move typecheck/svelte to a dedicated job (Linux only)
- move linter/formatter to a dedicated job (Linux only)
- keep unit tests in a dedicated job (Windows & Linux)

Fixes https://github.com/containers/podman-desktop/issues/8660

### How to test this PR?

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -194,14 +194,10 @@ jobs:
             PR_NUMBER
             PR_SHA
 
-  lint-format-unit:
-    name: linter, formatters and unit tests / ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
+  lint-format:
+    name: linter, formatters
+    runs-on: ubuntu-24.04
     timeout-minutes: 40
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [windows-2022, ubuntu-24.04]
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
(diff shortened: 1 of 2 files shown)


axel7083 (approved):

.github/workflows/pr-check.yaml
```diff
@@ -223,24 +219,65 @@ jobs:
...
       # Check we don't have changes in git
       - name: Check no changes in git
-        if: ${{ matrix.os=='ubuntu-24.04'}}
         run: |
           if ! git diff --exit-code; then
             echo "Found changes in git"
             exit 1
           fi
 
+  tpecheck:
```
odockal: not like it matters a lot, but there is a typo: just missing `y`

benoitf: thanks I think I need to change my glasses !

benoitf:
```suggestion
  typecheck:
```

odockal:
@benoitf np. I am think about if we can make some jobs dependent on others. Ie. using `needs` keyword in the jobs. Like it make sense to run unit tests only if we can compile the codebase. In our case, it might be that running jobs with name Windows, Linux require (`needs`) unit tests jobs first. It does not make sense build artifacts and upload them if we cannot build/run unit tests... Wdyt?

benoitf:
@odockal well it seems developers asked to have all the errors at once and not `fail-fast` as soon as there is something that is not working

so for example you could have something working but some tests failing. So maybe you may still want to be able to grab a binary to try the feature even if the tests are not completed/all passing

or you want to know that both unit tests and typecheck are failing (and not see that as it's failing the linter you don't have the result for the other jobs)

deboer-tim:
+1. Aside from cases where there really is a dependency I'd rather see all the problems and be able to grab artifacts vs multiple passes of commit + fail

odockal:
@benoitf @deboer-tim I totally understand that. I am just curious which job would be essential for all others, as I can imagine that without building/compiling the codebase nothing else has meaning to even try to run, right? But we do not have such job precisely, I think.

benoitf:
it still seems that people want to know where it is failing 

you can have a failure on electron build but it works in development mode

so for example only the build could fail while the others jobs would succeed or all of them

But while before (currently) it was a sequence and we only get the first failure, developers share their interest in knowing all jobs that would fail and not only the first one

odockal:
@benoitf Absolutely understandable! I am sharing more of my preference as QE. 
Seems like two jobs did not get run, maybe rebasing would help.

benoitf:
@odockal the jobs were required but their names has been changed, updating the configuration

PR #8744, opened 2024-09-04
Fixed changing background image based on light/dark theme

### LINKED ISSUE #8526: Poor contrast on Supercharge your apps with Red Hat banner graphic's text on Dashboard
### Bug description

I'm running Podman Desktop 1.12 on my Mac and the graphic at the bottom of the dashboard advertising the Red Hat extension pack uses white text on a light pink background, which is very hard to read. Could something be done next version to ensure there is sufficient contrast to easily read?

Screenshot 2024-08-20 at 12 27 42 PM

### Operating system

M2 Pro MacBook 14.6.1

### Installation Method

### Version

1.12.0

### Steps to reproduce

Open Podman Desktop and look at the bottom of the Dashboard area to see the graphic

### Relevant log output

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Fix changing background image based on light/dark theme

### Screenshot / video of UI

https://github.com/user-attachments/assets/eff29de6-1c78-4eb3-82b8-41de1482ddb4

Close #8526 

### How to test this PR?
Switch between light/dark theme and check if the recommendation banners reflect the change

### DIFF
--- a/packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
+++ b/packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
@@ -1,19 +1,17 @@
 <script lang="ts">
-import { onMount } from 'svelte';
-
 import FeaturedExtension from '/@/lib/featured/FeaturedExtension.svelte';
 
 import type { ExtensionBanner } from '../../../../main/src/plugin/recommendations/recommendations-api';
 
-export let banner: ExtensionBanner;
 // Pass in the theme appearance colour of PD to the banner, we do it here so we don't have to do multiple isDark checks when rendering multiple banners.
-export let isDark: boolean;
+let { banner, isDark }: { banner: ExtensionBanner; isDark: boolean } = $props();
 
-let style: string | undefined = undefined;
-let hasBackground: boolean =
-  !!banner.background && ((!!banner.background.dark && !!banner.background.light) || !!banner.background.gradient);
+let style = $state<string | undefined>(undefined);
+let hasBackground = $state(
+  !!banner.background && ((!!banner.background.dark && !!banner.background.light) || !!banner.background.gradient),
+);
 
-onMount(() => {
+$effect(() => {
   if (banner.background?.gradient) {
     style = `background: linear-gradient(${banner.background.gradient.start}, ${banner.background.gradient.end});`;
   } else if (isDark && banner.background?.dark) {


packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
```diff
@@ -13,7 +13,7 @@ let style: string | undefined = undefined;
 let hasBackground: boolean =
   !!banner.background && ((!!banner.background.dark && !!banner.background.light) || !!banner.background.gradient);
 
-onMount(() => {
+afterUpdate(() => {
```
axel7083: afterUpdate is deprecated in svelte 5 https://svelte-5-preview.vercel.app/docs/deprecations

benoitf: I think the update needs to occur when isDark is updated

gastoner:
Updated file to svelte5

packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
gastoner: Yes I've pushed conversion to svelte5 which could also work (option 1)
Or just add `$: if (isDark) {
  changeStyle();
}` function which will be also called in `onMount()` (option 2)
@benoitf @axel7083

axel7083 (commented):
https://github.com/user-attachments/assets/9fc67499-0043-463d-996c-8b8edff9c5e2

When trying to very quickly change the theme, I am able to have some opposite result

gastoner:
> Screencast.from.2024-09-04.16-34-14.mp4
> 
> When trying to very quickly change the theme, I am able to have some opposite result

Hmm, I got the same result even with the second option that I've proposed. Any ideas how to make this quicker or fix this, because it seems like if you return to the dashboard when is the light theme same, the image wont change

axel7083:
The `ExtensionBanners.svelte` fetch the isDark on mount, and it is not refresh as the component is reused.

https://github.com/containers/podman-desktop/blob/aba56f0e96f92bcecd9db6ff556b2dfd2b6b1341/packages/renderer/src/lib/recommendation/ExtensionBanners.svelte#L11-L15

gastoner:
@axel7083  try it now!

axel7083:
> The `ExtensionBanners.svelte` fetch the isDark on mount, and it is not refresh as the component is reused.
> 
> https://github.com/containers/podman-desktop/blob/aba56f0e96f92bcecd9db6ff556b2dfd2b6b1341/packages/renderer/src/lib/recommendation/ExtensionBanners.svelte#L11-L15

We might need to either, get the `isDark` at every render (possible but not really performance wise) or trying to get an event listener for it :?

gastoner:
> > The `ExtensionBanners.svelte` fetch the isDark on mount, and it is not refresh as the component is reused.
> > https://github.com/containers/podman-desktop/blob/aba56f0e96f92bcecd9db6ff556b2dfd2b6b1341/packages/renderer/src/lib/recommendation/ExtensionBanners.svelte#L11-L15
> 
> We might need to either, get the `isDark` at every render (possible but not really performance wise) or trying to get an event listener for it :?

The listener would be probably good, now it should get the boolean value each time user enters the dashboard

benoitf:
it looks like ApperanceUtil isDark or getTheme should be removed and it should be a state (v5) or store v4 being updated/refreshed when configuration is changing

axel7083:
> it looks like ApperanceUtil isDark or getTheme should be removed and it should be a state (v5) or store v4 being updated/refreshed when configuration is changing

The ApperanceUtil is doing some _hacky_ stuff to determine the theme

https://github.com/containers/podman-desktop/blob/9c946fddb1c614e7c0db0ff4a59c50f33d5e8d06/packages/renderer/src/lib/appearance/appearance-util.ts#L41-L53

Looking at the api of MediaQueryList there is a `addEventListener` and `removeEventListener` which would allow the base system for a quick hack.

But since we are using a configuration value (light / dark / system), we might want to deprecate the AppearanceUtil and have a store as suggested by @benoitf

gastoner:
> it looks like ApperanceUtil isDark or getTheme should be removed and it should be a state (v5) or store v4 being updated/refreshed when configuration is changing

Should I convert it to v5 then? I've done something like this https://github.com/containers/podman-desktop/commit/6db3c75bc2814a84aa31a9bb84db1187396130f0

feloy:
The PR breaks the "system" appearance:
- When you switch to the system appearance, the latest selected one is kept (if you were in dark mode, selecting the system mode displays the dark mode, even if the system mode is light)
- changing the system mode does not change the Podman Desktop mode reactively (it was changing it before)

Before:
https://github.com/user-attachments/assets/66329798-22b8-41ad-84ea-1426c225950c

After:
https://github.com/user-attachments/assets/ceacccf5-de49-449c-8527-0bc316d8a0b4

feloy (changes requested):
see comment above

gastoner:
@feloy This PR needs to be updated after will be #8759 closed, probably should convert it to draft for now

feloy:
> @feloy This PR needs to be updated after will be #8759 closed, probably should convert it to draft for now

sorry, my comment was for #8759

gastoner:
There would be a need to improve system change detection (similar problem in MonacoEditor)

packages/renderer/src/lib/recommendation/ExtensionBanners.svelte
```diff
@@ -1,10 +1,29 @@
...
+onDestroy(() => {
+  if (isDarkUnsubscribe) {
+    isDarkUnsubscribe();
+  }
+});
 </script>
 
 {#each $extensionBannerInfos as banner (banner.extensionId)}
-  <ExtensionBanner banner={banner} isDark={$isDark} />
+  <ExtensionBanner banner={banner} isDark={isDarkTheme} />
```
feloy: As far as I understand svelte reactivity, it seems to me that the changes in this file are not necessary, we are just adding a layer of reactivity.
(and testing without these changes, it works the same)

gastoner: I see, thank you, I thought that I've forgot to use subscribe here

feloy (approved):

benoitf (approved):
nice to see using also the new svelte5 syntax

PR #8745, opened 2024-09-04
feat(test): add pw fixtures

### LINKED ISSUE #8727: Create a components or action package that contains shared functionality used across multiple POM
Create some PW fixtures that can be used in tests:

Exemple: NavigationBar fixture, Podman Runner fixture, etc.

cbr7:
### What does this PR do?
Create several custom pw fixtures to be used in the e2e test framework.

https://github.com/containers/podman-desktop/issues/8727


benoitf (commented):
I think it would help if navigationBar is not renamed to navBar as part of this PR

tests/playwright/src/utility/fixtures.ts
```diff
@@ -0,0 +1,43 @@
+import type { Page } from '@playwright/test';
```
benoitf, line 19: missing copyright headers

tests/playwright/src/utility/fixtures.ts
```diff
@@ -0,0 +1,43 @@
...
+import { PodmanDesktopRunner } from '../runner/podman-desktop-runner';
+
+export type TestFixtures = {
+  pdRunner: PodmanDesktopRunner;
+  navBar: NavigationBar;
+  welcomePage: WelcomePage;
+  page: Page;
+};
+
+export type PDRunnerOptions = {
```
benoitf:
```suggestion
export type PodmanDesktopRunnerOptions = {
```

benoitf (commented):
I think it would help if navigationBar is not renamed to navBar as part of this PR

tests/playwright/src/utility/fixtures.ts
benoitf: (or just get rid of PodmanDesktop prefix as we're anyway inside Podman Desktop)

cbr7: renamed to PodmanDesktopRunnerOptions because you can use them also in the extension test frameworks when you import the test framework.

tests/playwright/src/utility/fixtures.ts
cbr7: added, good spot!

tests/playwright/src/utility/fixtures.ts
benoitf: but usually when you import stuff there is no prefix in the imports

like

```
import {Foo} from 'podman-desktop-library'
```

not
```
import {PodmanDesktopFoo} from 'podman-desktop-library'
```

it's like `test` you import `test` from `playwright` or from `jest` or `vitest` the prefix is in fact the namespace that we import

and if you want to get rid of some ambiguities you change the name when doing the import but it's a consumer choice

cbr7: So you would like it to be just `RunnerOptions`?

cbr7: Renamed to RunnerOptions. :)

benoitf: yes I think PodmanDesktopRunner should be exported also under Runner (so for a while both can be imported) and here RunnerOptions

```ts
import { Runner, RunnerOptions} from '@podman-desktop/tests-playwright';
```

cbr7: Going to create a different issue to have the runner renamed, don't want to have it here, PR is long enough as it is.

cbr7: Created https://github.com/containers/podman-desktop/issues/8749 to track the runner renaming.

benoitf: sure, thanks

benoitf: I think the PR is longer also due to all the `navBar` usage instead of `navigationBar`

cbr7: yeah it's true, we didn't actually have a standard for it, half of the test files were one way and half the other. :/

benoitf: if you prefer you can just issue a PR doing a change to have something uniform between all `navBar` or `navigationBar` as sometimes `navBar` is currently used or `navigationBar`

odockal (approved):

tests/playwright/src/utility/fixtures.ts
```diff
@@ -0,0 +1,61 @@
...
+  autoUpdate: boolean;
+  autoCheckUpdate: boolean;
+};
+
+export const test = base.extend<TestFixtures & RunnerOptions>({
+  profile: ['', { option: true }],
+  customFolder: ['podman-desktop', { option: true }],
+  autoUpdate: [true, { option: true }],
+  autoCheckUpdate: [true, { option: true }],
+  pdRunner: async ({ profile, customFolder, autoUpdate, autoCheckUpdate }, use) => {
```
benoitf, line 45:
```suggestion
  runner: async ({ profile, customFolder, autoUpdate, autoCheckUpdate }, use) => {
```

benoitf: (can be follow-up)

cbr7: This will be part of the renaming of the PD runner class, for now we had a standard to reference the podman desktop runner object as pdRunner in all tests, so for now I'd rather keep it as is, but will rename the fixture to runner with the rest of the refactoring.

benoitf: yes I agree 👍

PR #8747, opened 2024-09-04
feat: allow to set a zoom level

### LINKED ISSUE #1660: Increase font size

The font size of the logs is very small (at least on my Fedora instance).  

I did in the end find in the Settings > Editor|Terminal > Integrated: Font Size how to change the font size but that was not very intuitive to find.  

Also, in the 'Inspect' tab for a container, when I right click I see a command palette where I can change the font size for that particular view.  It's a bit strange that this only applies to the Inspect view. 

 It would be nice if the default font size was larger and/or if I could use ctrl + '+' or '-' to change the font size.   

 It would be nice if the command palette of the Inspect view would apply to the other views as well; and if I could set keyboard shortcuts for eg. changing the font. 

_No response_

### Additional context

_No response_

### LINKED ISSUE #8640: Allow users to change their zoom level in the settings

It is not clear how I can change the font size of the entire UI. We are able to do it by using ctrl +/- however that is not clear. 

A suggestion would to add this as an option in the `Settings > Preferences > Appearance` area. We can maybe call it `zoom level`? 

A label could also be added that mentions "Shortcut: Ctrl +/-"

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
user can define a zoom to get bigger fonts/UI elements inside Podman Desktop

### Screenshot / video of UI

https://github.com/user-attachments/assets/4b0ef3e0-5a36-4219-aff2-e28ada06f8ec

fixes https://github.com/containers/podman-desktop/issues/8640
fixes https://github.com/containers/podman-desktop/issues/1660

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/appearance-settings.ts
+++ b/packages/main/src/plugin/appearance-settings.ts
@@ -22,4 +22,5 @@ export enum AppearanceSettings {
   LightEnumValue = 'light',
   DarkEnumValue = 'dark',
   SystemEnumValue = 'system',
+  ZoomLevel = 'zoomLevel',
 }
--- a/packages/main/src/application-menu-builder.spec.ts
+++ b/packages/main/src/application-menu-builder.spec.ts
@@ -23,6 +23,7 @@ import { aboutMenuItem } from 'electron-util/main';
 import { beforeEach, expect, test, vi } from 'vitest';
 
 import { ApplicationMenuBuilder } from './application-menu-builder.js';
+import type { ZoomLevelHandler } from './plugin/zoom-level-handler.js';
 
 vi.mock('electron', async () => {
   class MyCustomWindow {
@@ -63,18 +64,16 @@ vi.mock('electron-util/main', async () => {
   };
 });
 
+const zoomLevelHandlerMock = {} as unknown as ZoomLevelHandler;
+
 let applicationMenuBuilder: ApplicationMenuBuilder;
 
 beforeEach(() => {
   vi.resetAllMocks();
-  applicationMenuBuilder = new ApplicationMenuBuilder();
+  applicationMenuBuilder = new ApplicationMenuBuilder(zoomLevelHandlerMock);
 });
 
 test('check about menu is added', () => {
-  vi.mocked(Menu.buildFromTemplate).mockImplementation(a => {
-    return a as unknown as Menu;
-  });
-
   // set a menu
   vi.mocked(Menu.getApplicationMenu).mockReturnValue({
     items: [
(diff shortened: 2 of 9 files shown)


packages/main/src/plugin/appearance-settings.ts
```diff
@@ -22,4 +22,6 @@ export enum AppearanceSettings {
   LightEnumValue = 'light',
   DarkEnumValue = 'dark',
   SystemEnumValue = 'system',
+
```
axel7083:
```suggestion
```

gastoner (commented):
I got some unexpected inc/dec steps (only in zoom level)



https://github.com/user-attachments/assets/1cc2672f-1198-496f-8bb7-73d05a88ad52

benoitf:
@gastoner the PR wasn't including https://github.com/containers/podman-desktop/pull/8705

I rebased the PR, so please retry

axel7083:
> I got some unexpected inc/dec steps (only in zoom level)

Facing the same issue 

https://github.com/user-attachments/assets/c8c9700d-db83-419a-9ba8-2490d8d005e3

benoitf:
also anyway it's not related to this PR

axel7083 (approved):
LGTM the only element I would notice is the delay, but I suppose this is expected due to the event architecture.

gastoner (approved):
Same as Axel

dgolovin (dismissed):
This PR mess up with Ctrl +/- zoom in different ways. 

1. If I set it up with Ctrl +/- and restart podman desktop it going to reset to zoom level in settings, not one I had in previous session.
2. If I set zoom with Ctrl +/- and switch to Appearance prefernces it resets is right away to zoom level stored in preferences

This fix should probably write/read window zoom level directly rather then store it in preferences.

Another option listen to window zoom level and update value saved in preferences.

benoitf:
There is no listener / event AFAIK for the shortcuts, only for mouse scroll

dgolovin:
There is https://github.com/electron/electron/blob/main/docs/api/web-contents.md#event-zoom-changed 'zoom-changed' event, but I have not tested it.

benoitf:
Emitted when the user is requesting to change the zoom level using the mouse wheel.

benoitf:
I'll investigate other ways of intercepting

dgolovin:
The 'resize' event is triggered every time I zoom in/out. ResizeObserver can be used on always visible element, nav for example, to detect zoom change. 

```
diff --git a/packages/renderer/src/AppNavigation.svelte b/packages/renderer/src/AppNavigation.svelte
index 0eaada6f88..b8349d81da 100644
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -25,10 +25,13 @@ let authActions = $state<AuthActions>();
 let outsideWindow = $state<HTMLDivElement>();
 
 const iconSize = '22';
-
+const resizeObserver = new ResizeObserver(() => {
+  console.log('zoom detected');
+})
 onMount(async () => {
   commandRegistry.init();
+  resizeObserver.observe(document.getElementById('navigation') as Element);
 });
 
 function clickSettings(b: boolean) {
@@ -41,7 +44,7 @@ function clickSettings(b: boolean) {
 </script>
 
 <svelte:window />
-<nav
+<nav id="navigation"
   class="group w-leftnavbar min-w-leftnavbar flex flex-col hover:overflow-y-none bg-[var(--pd-global-nav-bg)] border-[var(--pd-global-nav-bg-border)] border-r-[1px]"
   aria-label="AppNavigation">
   <NavItem href="/" tooltip="Dashboard" bind:meta={meta}>
~
```

the test above prints 'zoom detected' when window size or zoom changed.

benoitf:
@dgolovin it seems it's not the case on some os ( I found some items on GitHub issues or stack overflow when implementing this PR)

I switched then to https://github.com/containers/podman-desktop/pull/8760

where we'll define our shortcuts ( and then can handle when user is calling it) and it'll also fix issues where zoom in/out shortcuts are not working on Windows or Linux as expected

benoitf:
removing previous reviews as I reworked the entire stuff

new actions to perform `zoom in` , `zoom out` or `zoom reset`

and using shortcuts should persist the value of the configuration property

axel7083 (commented):
https://github.com/user-attachments/assets/a1318f15-4d28-491a-a3de-4f09ef9a25cf

If I zoom on the preference page, I don't see the value being updated directly, otherwise everything seems to be working great

dgolovin (changes requested):
Tested it and noticed couple issues:
1. When I am using key combination to zoom in/out if I do it multiple times back and forth the zoom value is getting values like `1.0000000003` very quickly
2. When zooming with key combination if I get value over 3 and then switch to appearance settings the zoom resets to 0.

dgolovin:
From usability point of view showing zoom level in % would be much better. Firefox shows it in % too 

image

benoitf:
https://github.com/containers/podman-desktop/pull/8793 and https://github.com/containers/podman-desktop/pull/8795 would at least need to be merged

benoitf:
rebased to include two fixes that have been merged

axel7083 (commented):
https://github.com/user-attachments/assets/cd06cfeb-2996-4b4b-ab38-c755229667d6

Everything seems to be working well, except a weird flickering when using `ctrl + num0` (from numpad) to reset the zoom.

As visible in the video, weird behavior: the value seems to be restore to 0 but the preference does not get updated

dgolovin (approved):
LGTM

There are minor glitches that does not really affect user experience.
1. After start change zoom and go to preferences page and try to use +/- button and and the value is not updated on the page, then it suddenly starts to work (hard to replicate)
2. I also could not to set value to 1 using +/i values, then it started to work

benoitf:
ok, no other reviews since a couple of days, 2 approvals and no required changes, will merge

PR #8751, opened 2024-09-04
Update settings.json configuration in podman-desktop-runner.ts

ScrewTSW:
### What does this PR do?
* parametrizes customSettings object
* fixes deleted recordings if status is any other than `failed`

### Screenshot / video of UI
N/A

Tangent to https://github.com/containers/podman-desktop-extension-ai-lab/issues/1504

### How to test this PR?
N/A

### DIFF
--- a/tests/playwright/src/runner/podman-desktop-runner.ts
+++ b/tests/playwright/src/runner/podman-desktop-runner.ts
@@ -40,8 +40,6 @@ export class PodmanDesktopRunner {
   private readonly _customFolder;
   private readonly _testOutput: string;
   private _videoAndTraceName: string | undefined;
-  private _autoUpdate: boolean;
-  private _autoCheckUpdate: boolean;
   private _customSettingsObject: { [key: string]: unknown } = {};
 
   private static _instance: PodmanDesktopRunner | undefined;
@@ -49,16 +47,18 @@ export class PodmanDesktopRunner {
   static async getInstance({
     profile = '',
     customFolder = 'podman-desktop',
-    autoUpdate = true,
-    autoCheckUpdate = true,
+    customSettingsObject = {
+      'preferences.OpenDevTools': 'none',
+      'extensions.autoUpdate': true,
+      'extensions.autoCheckUpdates': true,
+    },
   }: {
     profile?: string;
     customFolder?: string;
-    autoUpdate?: boolean;
-    autoCheckUpdate?: boolean;
+    customSettingsObject?: { [key: string]: unknown };
   } = {}): Promise<PodmanDesktopRunner> {
     if (!PodmanDesktopRunner._instance) {
-      PodmanDesktopRunner._instance = new PodmanDesktopRunner({ profile, customFolder, autoUpdate, autoCheckUpdate });
+      PodmanDesktopRunner._instance = new PodmanDesktopRunner({ profile, customFolder, customSettingsObject });
       await PodmanDesktopRunner._instance.start();
     }
     return PodmanDesktopRunner._instance;
@@ -67,16 +67,14 @@ export class PodmanDesktopRunner {
   private constructor({
     profile = '',
     customFolder = 'podman-desktop',
-    autoUpdate = true,
-    autoCheckUpdate = true,
     customSettingsObject = {
       'preferences.OpenDevTools': 'none',
+      'extensions.autoUpdate': true,
+      'extensions.autoCheckUpdates': true,
     },
   }: {
     profile?: string;
     customFolder?: string;
-    autoUpdate?: boolean;
-    autoCheckUpdate?: boolean;
     customSettingsObject?: { [key: string]: unknown };
   } = {}) {
     this._running = false;
@@ -84,11 +82,7 @@ export class PodmanDesktopRunner {
     this._testOutput = join('tests', 'playwright', 'output', this._profile);
     this._customFolder = join(this._testOutput, customFolder);
     this._videoAndTraceName = undefined;
-    this._autoUpdate = autoUpdate;
-    this._customSettingsObject['extensions.autoUpdate'] = this._autoUpdate;
-    this._autoCheckUpdate = autoCheckUpdate;
-    this._customSettingsObject['extensions.autoCheckUpdates'] = this._autoCheckUpdate;
-    if (customSettingsObject) this._customSettingsObject = customSettingsObject;
+    this._customSettingsObject = customSettingsObject;
 
     // Options setting always needs to be last action in constructor in order to apply settings correctly
     this._options = this.defaultOptions();
--- a/tests/playwright/src/runner/podman-desktop-runner.ts
+++ b/tests/playwright/src/runner/podman-desktop-runner.ts
@@ -42,6 +42,7 @@ export class PodmanDesktopRunner {
   private _videoAndTraceName: string | undefined;
   private _autoUpdate: boolean;
   private _autoCheckUpdate: boolean;
+  private _customSettingsObject: object | undefined;
 
   private static _instance: PodmanDesktopRunner | undefined;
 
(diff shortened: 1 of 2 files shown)


tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -314,11 +352,25 @@ export class PodmanDesktopRunner {
...
       'extensions.autoCheckUpdates': this._autoCheckUpdate,
       'extensions.autoUpdate': this._autoUpdate,
+      'extensions.disabled': [
+        !this._composeEnabled ? 'podman-desktop.compose' : '',
+        !this._dockerEnabled ? 'podman-desktop.docker' : '',
+        !this._kindEnabled ? 'podman-desktop.kind' : '',
+        !this._kubeContextEnabled ? 'podman-desktop.kube-context' : '',
+        !this._kubectlCliEnabled ? 'podman-desktop.kubectl-cli' : '',
+        !this._registriesEnabled ? 'podman-desktop.registries' : '',
+        !this._limaEnabled ? 'podman-desktop.lima' : '',
```
cbr7: What are these and why do we need them?

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -314,11 +352,25 @@ export class PodmanDesktopRunner {
...
+      ],
+      'podman.binary.path': !this._isLinux
+        ? `${process.env.USERPROFILE}\\tools\\podman\\podman-5.2.0\\usr\\bin\\podman.exe`
+        : '/usr/bin/podman',
```
cbr7: Why are we hardcoding some paths here? What if the test is running on mac? What is this at all?

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -233,26 +267,30 @@ export class PodmanDesktopRunner {
...
       console.log(`Caught exception in removing traces: ${err}`);
       return;
     }
 
-    if (!process.env.KEEP_TRACES_ON_PASS) {
-      const tracesPath = join(this._testOutput, 'traces', `${this._videoAndTraceName}_trace.zip`);
-      if (existsSync(tracesPath)) {
-        console.log(`Removing traces folder: ${tracesPath}`);
-        rmSync(tracesPath, { recursive: true, force: true, maxRetries: 5 });
+    if (testStatus && testStatus === 'passed') {
```
cbr7: What was the issue with this check previously? why do we need another nested if statement?

cbr7 (changes requested):
Left some comments above.

benoitf (commented):
I think you reached the limit on the constructor parameters

Probably need to provide a complex object instead

also I think I would expect to just have an array of string of extension id to disable or enable rather than one parameter by extension

benoitf:
instead of doing 3 things I think it should be one PR per thing/ item

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: yeah, will change it to pass an optional config object into the method

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: it gives us an option to disable unnecessary extensions that might interfere with the tests or clog up the logs. Changing it to passing an entire settings object into the method

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: it could pass `timedOut` or `interrupted` status and the traces and video would get deleted

ScrewTSW:
> I think you reached the limit on the constructor parameters
> 
> Probably need to provide a complex object instead
> 
> also I think I would expect to just have an array of string of extension id to disable or enable rather than one parameter by extension

yes, discussed with Ondra, will be changing it to an object parameter

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -233,26 +267,30 @@ export class PodmanDesktopRunner {
       rmSync(rawTracesPath, { recursive: true, force: true, maxRetries: 5 });
     }
 
+    let testStatus: string | undefined;
     try {
-      if (test.info().status === 'failed') return;
+      testStatus = test.info().status;
+      console.log(`Test finished with status:${testStatus}`);
```
cbr7:
```suggestion
      if(test.info().status !== 'passed') return;
```

tests/playwright/src/runner/podman-desktop-runner.ts
cbr7: This additional nested if statement can be removed, the issue can be addressed as per the suggestion from above.

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: fixed

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: fixed

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: fixed

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: fixed

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -256,7 +265,9 @@ export class PodmanDesktopRunner {
     }
 
     try {
-      if (test.info().status === 'failed') return;
+      const testStatus = test.info().status;
+      console.log(`Test finished with status:${testStatus}`);
+      if (testStatus === 'passed' || testStatus === 'skipped') return;
```
cbr7:
```suggestion
      if (testStatus !== 'passed' && testStatus !== 'skipped') return;
```

Shouldn't it be like this? If the test isn't passed and isn't skipped then we can proceed with the trace/video deletion, otherwise we don't delete any video/traces.

The way it's written currently if the test is failed then the doesn't meet the if statement criteria and as such won't trigger the return statement so the it will proceed to delete the traces and video.

tests/playwright/src/runner/podman-desktop-runner.ts
```diff
@@ -336,11 +347,13 @@ export class PodmanDesktopRunner {
       mkdirSync(parentDir, { recursive: true });
     }
 
-    const settingsContent = JSON.stringify({
-      'preferences.OpenDevTools': 'none',
-      'extensions.autoCheckUpdates': this._autoCheckUpdate,
-      'extensions.autoUpdate': this._autoUpdate,
-    });
+    const settingsContent = this._customSettingsObject
+      ? JSON.stringify(this._customSettingsObject)
```
odockal: Include the values we use in settings as default into default object for customSettingsObject. So we will get right of them in constructor.

odockal (changes requested):

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: oh, good catch!

ScrewTSW: fixed

tests/playwright/src/runner/podman-desktop-runner.ts
ScrewTSW: fixed

odockal (approved):

cbr7 (approved):

PR #8753, opened 2024-09-05
chore: add release notes to dashboard - main and preload

### LINKED ISSUE #8521: Make Release Note available from the release version displayed in the status bar

As a user I would like to be able to view the latest release note of Podman Desktop

Display the release notes behind the release version button from the status bar

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
As part of the dashboard revamp, this PR adds the release notes to the dashboard page. 
This PR is focused on the backend side of the release notes retrieval, and will be followed by https://github.com/containers/podman-desktop/pull/9190 (renderer) and https://github.com/containers/podman-desktop/pull/9058 ( `version` button in the status bar)

Related to https://github.com/containers/podman-desktop/issues/8521

### How to test this PR?


SoniaSandler:
@ekidneyrh @slemeur let me know if the release notes banner in the dashboard looks good (I just need to add the icon to the update button and the current version on the bottom right side of the banner). The summary is taken from the release notes on the website, and the max number of lines in the summary that can be visible is 6, anything more than that is hidden.
In the status bar, when clicking on the version, there are a couple of different message boxes the pop up depending on the OS (Linux vs Mac/Windows) and if an update is available. The screenshot I added is for the case of no update is available. If there is an update available (only for Mac or Windows), there are two buttons, `Update now` and `Cancel`, so I added another button `View release notes`between them with a link to the release notes of the new version.
@benoitf @deboer-tim for the release notes banner, I added a JSON file with information about the release notes and the image of the release in the same directory as the `ReleaseNotesBox` in `renderer`, but I'll more than happy to move it to `main` if needed (like the guides in the dashboard page). I'm also not sure if the release notes information and image is something that we'd like to change manually every release/version or if there is a way to do it automatically?
There are still some things that I have to add like update availability events for the visibility of update button and double check the updater (the `app.getVersion()` doesn't match the `podman.io` link version format, so I still need to extract the correct format from `app.getVersion()` and `updateVersion`).

ekidneyrh:
@SoniaSandler Looks good! Could we round the corners on the image also? We could probably optimize the release notes going forward so it doesnt immediately start listing bullet points so it looks nicer in the banner. Maybe cut it to three/four lines so it reduces the height of the banner.

slemeur:
Nice job @SoniaSandler ! 

I have a couple of questions:
- As a user, once I've read the release note, I probably don't need that to be displayed on the dashboard. I think we should add a "close" button, so the user can remove the box. But it must still reappear if there is one new release ;)
- Would it be possible to have a flexible height, probably depending on the summary lenght?
- Would we be able to provide a bit of formatting?

SoniaSandler:
@slemeur I think the close button is a good option. Do we want the release notes to be the ones of the current version (the one that is installed) or the latest release?
It is possible to have flexible height to show all the summary lines, but @ekidneyrh suggested to reduce the number of available lines. Regarding the formatting, right now I'm taking the release notes information from a JSON file, so I think it might be possible.

benoitf (dismissed):
I think there is an egg and chicken issue there
- how to embed release notes in the binary while it's not published until binaries are verified/tested

Shouldn't it fetch the release notes if available remotely and display the markdown content. Here it means that we would need to add content both in the binary (in that json file) and on the website

If the release notes are not available but the release is there, it should display the things from https://github.com/containers/podman-desktop/releases/tag/v1.12.0

Image should be retrieved from the remote side.

In case of airgapped mode, nothing is displayed

deboer-tim:
I always thought it was chicken and egg issue. 😉

As a user I'd prefer just seeing what's normally at the top of our release notes (the image plus what we have in the 'This release includes:' section) with a link to the full blog. That text would be much easier to finish early and include in the binary, I just don't know if there's an easy way to do that technically. i.e. I'm wondering if we can do something like this:
- png + release-1.x.md contains the highlight bullets and has to be finished before cutting the binary. Build process pulls a copy into the binary. Offline support works, and link goes to the full blog if you want to read more.
- The top of the release notes is copied from here or uses an include to avoid duplication.

benoitf:
AFAIK release notes are never ready when cutting the binaries and probably it won't so it should NOT be included inside the binary

SoniaSandler:
@deboer-tim @benoitf I was able to get the same info as in the screenshot from the release notes in `https://podman-desktop.io/`. The new code I have right now in this PR is the general idea of how to get the notes from remote (I still need to add a lot of checks and refine the code), but the issue with this solution is that it heavily relies on the format of the release notes, so even a small change like the placement of `Podman Desktop 1.12 is now available. Click here to download it` in the release notes of `1.12` vs `1.11` will impact what can be seen in the summary in the app.
You can modify line 16 in `ReleaseNotesBox` to check the release notes of different versions, and it will be removed before merging

packages/renderer/src/lib/dashboard/release-notes/ReleaseNotesBox.svelte
```diff
@@ -0,0 +1,79 @@
...
+let imageUrl: string = '';
+let imageAlt: string = '';
+
+async function getImageFromNotes() {
+  currentVersion = '1.12.0';
+  let curVersionSplit = currentVersion.split('.', 2);
+  const urlVersionFormat = curVersionSplit.join('.');
+  let notesURL = `https://podman-desktop.io/blog/podman-desktop-release-${urlVersionFormat}`;
+  const response = await fetch(notesURL);
+  if (!response.ok) {
```
benoitf: nice improvement

here we might need to use the source file rather than the published one (open question)

https://raw.githubusercontent.com/containers/podman-desktop/main/website/blog/2024-08-08-release-1.12.md

(we might need a json file with the mapping between the version and the source file or another pattern to find the file)

but then we have a file with all the data

also `<!--truncate-->` needs to be defined to know when to split the full content

SoniaSandler: @benoitf my first thought was also to use the raw md github file instead of the website, but having the date in the link makes it more challenging since we cannot guess it automatically like the version in the link. Since this file might not even exists when the binary is cut, how can we add it to a mapping that is included in the app before the release? We could add a mapping with some specific date/file name before the release and rely on the notes' author to use the same date/name for the file.

SoniaSandler: I've opened a PR for the mapping, and the new commit I pushed shows how I'm planning on fetching the information from the raw files. Right now, you can see the difference between using the website's notes (`ReleaseNotesBox`) and GitHub's notes (`ReleaseNotesBoxCopy`). `ReleaseNotesBoxCopy` has a placeholder notes link that I'll change once https://github.com/containers/podman-desktop/pull/8790 is merged

packages/preload/src/index.ts
```diff
@@ -1156,6 +1156,30 @@ export function initExposure(): void {
     },
   );
 
+  contextBridge.exposeInMainWorld(
+    'updatePodmanDesktop',
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
feloy: Is this necessary? I don't see any `any`

packages/main/src/plugin/updater.ts
```diff
@@ -62,6 +62,23 @@ export class Updater {
...
+    console.log('called opennotes');
+    if (version === 'current') {
+      version = app.getVersion();
+    } else if (version === 'latest') {
+      version = this.#nextVersion?.substring(1) ?? '';
+    }
+    console.log(version);
+    const urlVersionFormat = version.split('.', 2).join('.');
+    let notesURL = `https://podman-desktop.io/blog/podman-desktop-release-${urlVersionFormat}`;
+    const response = await fetch(notesURL);
```
feloy: I always get an error thrown here, even if I hardcode notesURL with an existing one, for example 
```
    notesURL = `https://podman-desktop.io/blog/podman-desktop-release-1.12`;
```

SoniaSandler: @feloy Is the error thrown when clicking on the `learn more` in the release notes banner or the version button in the status bar? Could you show the error?

feloy: From both banner and popup from status bar.

By changing this code in `openReleaseNotes`:
```
-   let notesURL = `https://podman-desktop.io/blog/podman-desktop-release-${urlVersionFormat}`;
+   let notesURL = `https://podman-desktop.io/blog/podman-desktop-release-1.12`;
+   try {
      const response = await fetch(notesURL);
      if (!response.ok) {
        notesURL = `https://github.com/containers/podman-desktop/releases/tag/v${version}`;
      }
      await shell.openExternal(notesURL);  
+   } catch (err: unknown) {
+     console.log('err', err);
+   }
```

I get this error in the console:
```
main ↪️ err TypeError: fetch failed
    at node:internal/deps/undici/undici:13178:13
    at async Updater.openReleaseNotes (/Users/phmartin/Documents/gits/podman-desktop/packages/main/dist/index.cjs:27519:25)
    at async /Users/phmartin/Documents/gits/podman-desktop/packages/main/dist/index.cjs:27575:9
    at async /Users/phmartin/Documents/gits/podman-desktop/packages/main/dist/index.cjs:127502:9
    at async /Users/phmartin/Documents/gits/podman-desktop/packages/main/dist/index.cjs:126653:26
    at async WebContents.<anonymous> (node:electron/js2c/browser_init:2:86078)Caused by: Error
    at internalConnectMultiple (node:net:1118:18)
    at afterConnectMultiple (node:net:1685:7)
```

SoniaSandler: @feloy I wan't able to reproduce your error, but I did change `fetch` to `https.get` to check if the page exists, so lmk if that works for you now

feloy: @SoniaSandler yes, it works now, thanks

SoniaSandler:
@benoitf @feloy right now, the close button of the banner works with a svelte store, so once the button is clicked, the value of the store is updated to false (don't show release notes banner). 
However, IIUC, using svelte store means that every time the app is closed and reopened, the store is created again and has its initial value (in this case true -> show banner), so every time a user closes PD and reopens the app, the banner will show again even if they closed it. 
Would it be better to save the show-banner value as something like the preferences?
When PD is updated, are the preferences reset or kept from the previous version?

benoitf:
@SoniaSandler yes you'll need to store the value as a preference 

Look at how it's done for telemetry ( you only see the welcome screen once)

There is a hidden flag on properties so it's not displayed in the settings page but you can use the property for this purpose

deboer-tim (changes requested):
Tried this out and noticed a few issues:
- Since it is on 1.13.0-next branch there are no release notes - how do I test? (couldn't find it in a comment above)
- When release notes can't be found the formatting is odd (see below). Although we hope this will never happen in the release, I think we should fix the formatting in case it ever does.
- When I close it, the property that's saved is `"releaseNotesBanner.show": false`. Won't this mean that if I update to 1.13 and close the release notes, I'll never see the release notes for 1.14, 1.15, etc? I think the value should be the current version, so when you update you always see the new release notes on startup.
- Once I closed I couldn't find any way to reopen the section w/o editing the settings.json. Most other apps have a way to reopen them, e.g. VS Code has `Help > Show Release Notes`. Could be a separate issue but I think we should have this.
- Clicking the version in the status bad opens the external URL, IMHO it should just open the widget again (but I don't think that would replace my previous point).

Screenshot 2024-09-23 at 2 33 22 PM

packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
```diff
@@ -0,0 +1,79 @@
...
+let showBanner = true;
+let updateAvilable = false;
+let currentVersion: string;
+let title: string = '';
+let summary: string = '';
+let imageUrl: string = '';
+let imageAlt: string = '';
+
+async function getInfoFromNotes() {
+  let curVersionSplit = currentVersion.split('.', 2);
```
SoniaSandler: @deboer-tim add after line 16 `currentVersion = '1.12.0'`. I've had this line before for testing purposes, but the tests failed with it

SoniaSandler:
@deboer-tim I wasn't sure if the preferences stay the same when updating, so I'll change it to be by version.
About reopening the notes, I thought about adding something visible in the preferences section, but wan't sure if we needed it, so I'll update this too.

slemeur:
@SoniaSandler could you please create two different PRs?
one for the status bar and one for the improvement on the dashboard?

slemeur:
your PR for the status bar, should be referring to: https://github.com/containers/podman-desktop/issues/8521

SoniaSandler:
@feloy @benoitf @deboer-tim could you please take a look at this PR and let me know what other changes needed?
I moved the status bar part to https://github.com/containers/podman-desktop/pull/9058, but it won't work until this PR is merged.
I'm debating where to put another place you could control the visibility of the banner (besides the status bar) -> in preferences with a toggle button, a link in the help page that would take you to the dashboard page where the release notes banner will be visible, or some other place.

deboer-tim (changes requested):
Thanks for fixing the formatting when the release notes can't be found, this is getting better.

When the release notes can't be found I wonder if it's better to link to `https://podman-desktop.io/blog` vs the current `https://github.com/containers/podman-desktop/releases/tag/<version>`?

For the preferences, I meant using the version as a value not the key, e.g. `"releaseNotesBanner":"1.12.0"` not `"releaseNotesBanner.show.1.12.0": false`. Using the value would reopen the release notes whenever the version changes, and avoid adding keys indefinitely.

IMHO the release image should be aligned to the top.

deboer-tim:
> I moved the status bar part to #9058, but it won't work until this PR is merged.
👍🏼

> I'm debating where to put another place you could control the visibility of the banner (besides the status bar) -> in preferences with a toggle button, a link in the help page that would take you to the dashboard page where the release notes banner will be visible, or some other place.
Personally I'd add a menu item like VS Code, or Help page as secondary option. It doesn't really fit well as a preference since it is a one-time 'show this now'. Let's open another issue though, doesn't need to be part of this PR.

SoniaSandler:
@deboer-tim I can add another check for `https://podman-desktop.io/blog` before using the GitHub link, but based on the way it works now -> the release notes banner uses https://podman-desktop.io/release-notes/1.12.json (created from the release notes that are published to the website), and if it is not accessible it would probably mean that either `podman-desktop.io` has an issue or that the release notes themselves are not available yet. 
If you meant `https://podman-desktop.io/blog/podman-desktop-release-<version>`, I think it will also have some problems if the release notes for the banner are not accessible, but if it was just the blog link, I think https://github.com/containers/podman-desktop/releases/tag/<version> provides more relevant information for the user about the current  release.

About the preferences, how could it indicate that the release notes are hidden or shown? changing that value to 0 (or something similar) on close and then back to the version on show? will it automatically update to the new version if the release notes are closed (e.g. value is 0)

benoitf:
> When the release notes can't be found I wonder if it's better to link to https://podman-desktop.io/blog vs the current https://github.com/containers/podman-desktop/releases/tag/<version>?

I'm not sure to follow why I should be redirected to the blogs post (where there won't be the blog post for the release I'm interested) if I clicked to have the details of what is inside 1.13.0 for example. At least on GitHub release page I have the data.

Also it is a temporary redirect as when we announce, move the binary to the release artifact, usually the blog post is there.

deboer-tim:
> @deboer-tim I can add another check for `https://podman-desktop.io/blog` before using the GitHub link, but based on the way it works now -> the release notes banner uses https://podman-desktop.io/release-notes/1.12.json (created from the release notes that are published to the website), and if it is not accessible it would probably mean that either `podman-desktop.io` has an issue or that the release notes themselves are not available yet. If you meant `https://podman-desktop.io/blog/podman-desktop-release-<version>`, I think it will also have some problems if the release notes for the banner are not accessible, but if it was just the blog link, I think https://github.com/containers/podman-desktop/releases/tag/ provides more relevant information for the user about the current release.

Ok, I guess that's fair enough on accessibility. It was inconsistent for me but I think that's because it was picking up the 1.13.0-next in several places.

> About the preferences, how could it indicate that the release notes are hidden or shown? changing that value to 0 (or something similar) on close and then back to the version on show? will it automatically update to the new version if the release notes are closed (e.g. value is 0)

If (preference value != current version), show it. Set it on close as you're doing, but to the current version.

deboer-tim:
> > When the release notes can't be found I wonder if it's better to link to https://podman-desktop.io/blog vs the current https://github.com/containers/podman-desktop/releases/tag/?
> 
> I'm not sure to follow why I should be redirected to the blogs post (where there won't be the blog post for the release I'm interested) if I clicked to have the details of what is inside 1.13.0 for example. At least on GitHub release page I have the data.
> 
> Also it is a temporary redirect as when we announce, move the binary to the release artifact, usually the blog post is there.

As per above, I was confused b/c I had hardcoded the version in renderer but was still going there, was just another version in main. Leaving this is fine. 👍🏼

packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
```diff
@@ -0,0 +1,95 @@
...
+    <div class="flex bg-[var(--pd-content-card-bg)] rounded-md p-5 gap-3 flex-row flex-nowrap h-[230px] items-center">
+      {#if imageUrl}
+        <img src={imageUrl} class="max-h-[190px] w-auto max-w-[20%] object-contain rounded-md" alt={imageAlt} />
+      {/if}
+      <div class="flex flex-col flex-1">
+        <div class="flex flex-row items-center justify-between">
+          <p class="text-[var(--pd-content-card-header-text)] font-bold text-xl ml-2">
+            {title}
+          </p>
+          <CloseButton on:click={() => onclose()} />
```
jeffmaury:
```suggestion
          <CloseButton on:click={onclose} />
```

as we have/will have an eslint rule that enforce return type on functions/lamba declarations and Svelte does not support lamba with defined return type

Also `onclose` shoud be renamed to `onClose`

packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
```diff
@@ -0,0 +1,95 @@
...
+        </div>
+        <div
+          class="text-[var(--pd-content-card-text)] trunace text-ellipsis overflow-hidden whitespace-pre-line line-clamp-6">
+          <Markdown markdown={summary}></Markdown>
+        </div>
+        <div class="flex flex-row justify-end items-center gap-3 mt-2">
+          <Link on:click={() => window.podmanDesktopOpenReleaseNotes('current')}>Learn more</Link>
```
jeffmaury: Same here create an intermediate function

packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
```diff
@@ -0,0 +1,95 @@
...
+          <Button on:click={() => window.updatePodmanDesktop()} hidden={!updateAvilable} icon={faCircleArrowUp}
```
jeffmaury: Same here create an intermediate function

packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
```diff
@@ -0,0 +1,95 @@
...
+      </div>
+    </div>
+  {:else}
+    <div class="flex bg-[var(--pd-content-card-bg)] rounded-md p-5 gap-3 flex-row flex-nowrap h-auto items-center">
+      <div class="flex flex-row items-center justify-between w-full">
+        <p class="text-[var(--pd-content-card-header-text)] font-bold text-lg ml-2 w-full items-center">
+          Release notes are currently unavailable, please check again later or try this
+          <a href={notesURL} class="text-[var(--pd-link)]">link</a>
+        </p>
+        <CloseButton on:click={() => onclose()} />
```
jeffmaury:
```suggestion
        <CloseButton on:click={onclose} />
```

SoniaSandler:
@ekidneyrh this is how the release notes banner is set to change according to the size of the window right now.
Would it be better to have the image aligned to the top or stay in the center (vertically)?

[Screencast from 2024-09-30 09-29-52.webm](https://github.com/user-attachments/assets/b6867bbe-6577-4478-8e2d-ce560859fdcf)

ekidneyrh:
@SoniaSandler I think it would look better if aligned to the top - can we see what that would look like?

SoniaSandler:
@ekidneyrh 

[Screencast from 2024-09-30 10-23-33.webm](https://github.com/user-attachments/assets/34dd351b-85bd-4f50-aa27-50bb3ba005d0)

deboer-tim:
+1 from me, but it also looks like there is padding at the top/bottom/right that could be reduced a bit so that the top of the image and title align (and would also help reduce space overall).

'Podman Desktop v1.12.0' text seems redundant (won't it always be in the title?) and takes up an entire line. Can we remove it, or only show when the release notes can't be loaded?

ekidneyrh:
> @ekidneyrh
> 
>  Screencast.from.2024-09-30.10-23-33.webm

anchored to the top looks way better!

SoniaSandler:
@deboer-tim I think we can remove the `Podman Desktop v...` text since it already shows up in the same page in the status bar
Screenshot from 2024-09-30 11-27-10

SoniaSandler:
@benoitf @feloy @deboer-tim @jeffmaury could you please take a look at this PR?
It's all done from my side, so lmk if additional changes are needed.

packages/main/src/plugin/updater.ts
```diff
@@ -62,6 +64,22 @@ export class Updater {
...
   }
 
+  public async openReleaseNotes(version: string): Promise<void> {
+    if (version === 'current') {
+      version = app.getVersion();
+    } else if (version === 'latest') {
+      version = this.#nextVersion?.substring(1) ?? '';
+    }
+    const urlVersionFormat = version.split('.', 2).join('.');
+    let notesURL = `https://podman-desktop.io/blog/podman-desktop-release-${urlVersionFormat}`;
```
benoitf: I think here we should rely on the field `homepage` of the package.json file

(it is missing now but you can add it there: `https://www.podman-desktop.io`

packages/main/src/plugin/updater.ts
```diff
@@ -62,6 +64,22 @@ export class Updater {
...
+    https.get(notesURL, res => {
+      if (res.statusCode !== 200) {
+        notesURL = `https://github.com/containers/podman-desktop/releases/tag/v${version}`;
```
benoitf: IMHO it should rely on the `repository` field of the package.json file

benoitf: for example the pre-releases are in another repository and that field will be accurate

packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
```diff
@@ -0,0 +1,106 @@
...
+}
+
+function updatePodmanDesktop() {
+  window.updatePodmanDesktop();
+}
+
+async function getInfoFromNotes() {
+  let curVersionSplit = currentVersion.split('.', 2);
+  const urlVersionFormat = curVersionSplit.join('.');
+  notesURL = `https://podman-desktop.io/release-notes/${urlVersionFormat}.json`;
```
benoitf: the frontend component should not have podman-desktop.io,

The link and the fetch should come from the backend part
It'll avoid any CORS issue (and it's not the browser that does the connection, but the backend side)

benoitf (changes requested):
the fetch of the files should be done exclusively on the 'main/backend' side and not from the frontend part

Also we should reuse as much as possible the existing definition from package.json (like the github repository URL and podman-desktop.io instead of copying/pasting the constant everywhere)

to move forward, I think also you should split into 2 PR

one going to the main/src + preload side
one going to the frontend side (as usually this is where it is more biased so at least one part of the PR (backend side can be merged sooner)

jeffmaury (commented):
Two remarks:
- I think we should check if the extra link we offer in case we cannot display the release notes is available
- Click on version in status bar does nothing for me

SoniaSandler:
@jeffmaury the status bar change has been moved to a separate PR

packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
```diff
@@ -0,0 +1,106 @@
...
+        </div>
+      </div>
+    </div>
+  {:else}
+    <div class="flex bg-[var(--pd-content-card-bg)] rounded-md p-5 flex-col flex-nowrap h-auto items-center">
+      <div class="flex flex-row items-center justify-between w-full">
+        <p class="text-[var(--pd-content-card-header-text)] font-bold text-lg w-full items-center">
+          Release notes are currently unavailable, please check again later
+          {#if notesURL}
+            or try this <a href={notesURL} class="text-[var(--pd-link)]">link</a>
```
deboer-tim: Should be another Link.

packages/preload/src/index.ts
```diff
@@ -1157,6 +1157,26 @@ export function initExposure(): void {
...
+  contextBridge.exposeInMainWorld('updatePodmanDesktop', async (): Promise<void> => {
+    return ipcInvoke('app:update');
+  });
+
+  contextBridge.exposeInMainWorld('podmanDesktopUpdateAvailable', async (): Promise<boolean> => {
+    return ipcInvoke('app:update-available');
+  });
+
+  contextBridge.exposeInMainWorld('podmanDesktopOpenReleaseNotes', async (version: string): Promise<void> => {
+    return ipcInvoke('app:open-release-notes', version);
```
deboer-tim: This is only used once (for 'current') by the UI, after having successfully received the podmanDesktopGetReleaseNotes. Why not just return the URL in the notesURL of podmanDesktopGetReleaseNotes (or add an extra property) and then the UI can just follow the link? This would remove some complexity/tests and need to expose this API.

packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
SoniaSandler: @deboer-tim which link should we add if also the GitHub one is not available?
(FYI, the renderer changes have been moved to https://github.com/containers/podman-desktop/pull/9190)

deboer-tim: Just saying to use the component, i.e.:
`<a href={notesURL} class="text-[var(--pd-link)]">link</a>`
should be like:
`<Link on:click={() => window.openExternal(notesURL)}>link</a>`

SoniaSandler: Fixed! (in https://github.com/containers/podman-desktop/pull/9190)

SoniaSandler:
@benoitf I have created a separate PR for the renderer changes and only left the backend here:
main/preload - this PR
renderer - https://github.com/containers/podman-desktop/pull/9190
status bar - https://github.com/containers/podman-desktop/pull/9058
I also used the values from `package.json` for the links. They work, but I'm not sure if I did right.

packages/main/src/plugin/index.ts
```diff
@@ -1321,6 +1332,22 @@ export class PluginSystem {
...
+    });
+
+    this.ipcHandle('app:update-available', async (): Promise<boolean> => {
+      return podmanDesktopUpdater.updateAvailable();
+    });
+
+    this.ipcHandle(
+      'app:get-release-notes',
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      async (): Promise<{ releaseNotesAvailable: boolean; notesURL: string; notes?: any }> => {
```
benoitf: here what is the goal of the optional any notes parameter as it is not used inside the function ?

packages/preload/src/index.ts
```diff
@@ -1157,6 +1157,22 @@ export function initExposure(): void {
...
+    return ipcInvoke('app:update');
+  });
+
+  contextBridge.exposeInMainWorld('podmanDesktopUpdateAvailable', async (): Promise<boolean> => {
+    return ipcInvoke('app:update-available');
+  });
+
+  contextBridge.exposeInMainWorld(
+    'podmanDesktopGetReleaseNotes',
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf: same here why do we have several parameters unused

packages/main/src/plugin/index.ts
SoniaSandler: It's the notes information that is passed to the renderer side. If there are no release notes available, there is nothing to pass in `notes`, but if there are release notes available, they will be passed in the promise. I'll change the any type to the type of the release notes `{
  image: string;
  blog: string;
  title: string;
  summary: string;
}`

packages/preload/src/index.ts
SoniaSandler: they are all used on the renderer side

benoitf (commented):
@SoniaSandler you should add a new type like ReleaseNoteInfo in the API package and then import this type on main and renderer part rather than copying 3 times the definition ( so we're sure we're are handling the same return type everywhere)

packages/preload/src/index.ts
```diff
@@ -1157,6 +1158,25 @@ export function initExposure(): void {
...
+  });
+
+  contextBridge.exposeInMainWorld('podmanDesktopUpdateAvailable', async (): Promise<boolean> => {
+    return ipcInvoke('app:update-available');
+  });
+
+  contextBridge.exposeInMainWorld(
+    'podmanDesktopGetReleaseNotes',
+    async (): Promise<{
+      releaseNotesAvailable: boolean;
```
benoitf: you should have a return type for the whole method, not only for one field

benoitf (approved):
Thanks for all the changes 

Did you check in offline mode ? ( Not connected to internet)

SoniaSandler:
@benoitf offline:
Screenshot from 2024-10-03 14-10-44

benoitf:
as follow up maybe we should need a refresh/ retry button 

also is it working behind a proxy as I'm not sure you passed proxy data and here I don't know if it's applied on all the global configurations ( including custom certificates)
cc @jeffmaury

SoniaSandler:
@benoitf I'm not sure if it work behind a proxy since I'm not using one, but for now, in any case the app cannot retrieve the notes, it will show the banner in the previous screenshot.

deboer-tim (approved):
Another thought for later: maybe instead of `boolean` podmanDesktopUpdateAvailable should return something like `string | undefined` with the version in case the UI needs it?

Thanks for all the updates and improvements @SoniaSandler!

SoniaSandler:
@benoitf I'll merge this PR for now so that I could work on the other PRs for this issue, but if a problem will arise with the fetch and proxy, we can open a separate PR to fix it.

benoitf:
yes

PR #8759, opened 2024-09-05
refactor: turn appearance isDarkMode function to a reactive variable

gastoner:
### What does this PR do?
Refactors appearance-utils isDarkMode function to be reactive => changes it to reactive variable

Prerequisite of https://github.com/containers/podman-desktop/pull/8744

### DIFF
--- a/packages/renderer/src/lib/appearance/Appearance.spec.ts
+++ b/packages/renderer/src/lib/appearance/Appearance.spec.ts
@@ -22,6 +22,8 @@ import '@testing-library/jest-dom/vitest';
 import { render, type RenderResult } from '@testing-library/svelte';
 import { beforeEach, expect, test, vi } from 'vitest';
 
+import { configurationProperties } from '/@/stores/configurationProperties';
+
 import { AppearanceSettings } from '../../../../main/src/plugin/appearance-settings';
 import Appearance from './Appearance.svelte';
 
@@ -71,10 +73,11 @@ test('Expect light mode using system when OS is set to light', async () => {
   });
 
   getConfigurationValueMock.mockResolvedValue(AppearanceSettings.SystemEnumValue);
+  configurationProperties.set([]);
 
   const { baseElement } = await awaitRender();
   // expect to have no (dark) class as OS is using light
-  expect(getRootElementClassesValue(baseElement)).toBe('');
+  await vi.waitFor(() => expect(getRootElementClassesValue(baseElement)).toBe(''));
 });
 
 test('Expect dark mode using system when OS is set to dark', async () => {
@@ -85,14 +88,16 @@ test('Expect dark mode using system when OS is set to dark', async () => {
   });
 
   getConfigurationValueMock.mockResolvedValue(AppearanceSettings.SystemEnumValue);
+  configurationProperties.set([]);
 
   const { baseElement } = await awaitRender();
   // expect to have class being "dark" as OS is using dark
-  expect(getRootElementClassesValue(baseElement)).toBe('dark');
+  await vi.waitFor(() => expect(getRootElementClassesValue(baseElement)).toBe('dark'));
 });
 
 test('Expect light mode using light configuration', async () => {
   getConfigurationValueMock.mockResolvedValue(AppearanceSettings.LightEnumValue);
+  configurationProperties.set([]);
 
   const { baseElement } = await awaitRender();
 
@@ -105,10 +110,11 @@ test('Expect light mode using light configuration', async () => {
 
 test('Expect dark mode using dark configuration', async () => {
   getConfigurationValueMock.mockResolvedValue(AppearanceSettings.DarkEnumValue);
+  configurationProperties.set([]);
 
   const { baseElement } = await awaitRender();
 
-  // expect to have class being "dark" as we should be in light mode
+  // expect to have class being "dark" as we should be in dark mode
   expect(getRootElementClassesValue(baseElement)).toBe('dark');
 
   // expect to have color-scheme: dark
--- a/packages/renderer/src/lib/appearance/Appearance.svelte
+++ b/packages/renderer/src/lib/appearance/Appearance.svelte
@@ -1,18 +1,20 @@
 <script lang="ts">
 import { onDestroy, onMount } from 'svelte';
+import type { Unsubscriber } from 'svelte/store';
 
 import { AppearanceSettings } from '../../../../main/src/plugin/appearance-settings';
+import { isDark } from '../../stores/appearance';
 import { onDidChangeConfiguration } from '../../stores/configurationProperties';
-import { AppearanceUtil } from './appearance-util';
+
+let isDarkUnsubscribe: Unsubscriber;
+let isDarkTheme = false;
 
 const APPEARANCE_CONFIGURATION_KEY = AppearanceSettings.SectionName + '.' + AppearanceSettings.Appearance;
 async function updateAppearance(): Promise<void> {
   const html = document.documentElement;
 
   // toggle the dark class on the html element
-  const appearanceUtil = new AppearanceUtil();
-  let isDark = await appearanceUtil.isDarkMode();
-  if (isDark) {
+  if (isDarkTheme) {
     html.classList.add('dark');
     html.setAttribute('style', 'color-scheme: dark;');
   } else {
@@ -35,6 +37,11 @@ onMount(async () => {
     // notify changes
     window.dispatchEvent(new Event('appearance-changed'));
   });
+
+  isDarkUnsubscribe = isDark.subscribe(value => {
+    isDarkTheme = value;
+    updateAppearance();
+  });
 });
 
 // now, need to do the same for the appearance setting
@@ -43,5 +50,9 @@ onDidChangeConfiguration.addEventListener(APPEARANCE_CONFIGURATION_KEY, onDidCha
 // remove callback when the component is destroyed
 onDestroy(() => {
   onDidChangeConfiguration.removeEventListener(APPEARANCE_CONFIGURATION_KEY, onDidChangeConfigurationCallback);
+
+  if (isDarkUnsubscribe) {
+    isDarkUnsubscribe();
+  }
 });
 </script>
--- a/packages/renderer/src/lib/appearance/IconImage.spec.ts
+++ b/packages/renderer/src/lib/appearance/IconImage.spec.ts
@@ -26,6 +26,15 @@ import { AppearanceSettings } from '../../../../main/src/plugin/appearance-setti
 import IconImage from './IconImage.svelte';
 
 const getConfigurationValueMock = vi.fn();
+const getImageMock = vi.fn();
+
+vi.mock('./appearance-util', () => {
+  return {
+    AppearanceUtil: class {
+      getImage = getImageMock;
+    },
+  };
+});
 
 beforeEach(() => {
   vi.resetAllMocks();
@@ -38,7 +47,7 @@ beforeEach(() => {
 });
 
 test('Expect valid source and alt text with dark mode', async () => {
-  getConfigurationValueMock.mockResolvedValue(AppearanceSettings.DarkEnumValue);
+  getImageMock.mockReturnValue('dark.png');
 
   const image = render(IconImage, { image: { light: 'light.png', dark: 'dark.png' }, alt: 'this is alt text' });
 
@@ -49,12 +58,12 @@ test('Expect valid source and alt text with dark mode', async () => {
   const imageElement = image.getByRole('img');
 
   // expect to have valid source
-  expect(imageElement).toHaveAttribute('src', 'dark.png');
+  await vi.waitFor(async () => expect(imageElement).toHaveAttribute('src', 'dark.png'));
   expect(imageElement).toHaveAttribute('alt', 'this is alt text');
 });
 
 test('Expect valid source and alt text with light mode', async () => {
-  getConfigurationValueMock.mockResolvedValue(AppearanceSettings.LightEnumValue);
+  getImageMock.mockReturnValue('light.png');
 
   const image = render(IconImage, { image: { light: 'light.png', dark: 'dark.png' }, alt: 'this is alt text' });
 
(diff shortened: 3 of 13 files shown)


gastoner:
@axel7083 @benoitf PTAL

packages/renderer/src/lib/appearance/appearance-util.ts
```diff
@@ -17,25 +17,36 @@
...
 import { AppearanceSettings } from '../../../../main/src/plugin/appearance-settings';
+import { configurationProperties } from '../../stores/configurationProperties';
 
 export class AppearanceUtil {
-  async isDarkMode(): Promise<boolean> {
-    // get the configuration of the appearance
-    const appearance = await window.getConfigurationValue<string>(
-      AppearanceSettings.SectionName + '.' + AppearanceSettings.Appearance,
-    );
+  isDark: boolean = false;
```
benoitf: it needs to be reactive (state/ or writable store)

gastoner: The `$state` rune is only available inside `.svelte` and `.svelte.js/ts` files

benoitf: appearance-util.ts can be renamed to appearance.util.svelte.ts (it's on the client side and imported only in svelte file)
you can also use a Writable (or Readable store)

axel7083:
You could take a look at /@/stores/* content in the renderer package, containing store examples

https://github.com/containers/podman-desktop/blob/fba9eead7078174bf7806e89c5f48c60dd8d0c37/packages/renderer/src/stores/cli-tools.ts#L45

gastoner:
@axel7083 like so?

gastoner:
Or I should add separate file at /@/stores/*, but to my understanding the `configurationProperties.subscribe` should do the same thing as in the `/packages/renderer/src/stores/cli-tools.ts` file?

benoitf (commented):
I think it should be moved to the stores path so it's more aligned with the others and we can easily discover it

then, you can also calls and extending current store object, we have helpers, we can see stores in the UI, refresh them manually, see when they've been called, etc

gastoner:
@benoitf I've updated the PR, but I've seen your PR: https://github.com/containers/podman-desktop/pull/8795/files, which seems to me like it does a similar thing. Should I update mine?

benoitf:
hello @gastoner I'm not sure it's doing the same thing but I think yes if my PR is approved, it's more generic and then you'll need to rebase at some point your PR

gastoner:
Otherwise, do the changes look good?

gastoner:
TODO tests and typecheck

packages/renderer/src/stores/appearance.ts
```diff
@@ -0,0 +1,49 @@
...
+import { AppearanceSettings } from '../../../main/src/plugin/appearance-settings';
+import { configurationProperties } from './configurationProperties';
+
+export const isDark: Writable<boolean> = writable(false);
+const appearance: Writable<undefined | string> = writable();
+
+configurationProperties.subscribe(() => {
+  if (window.getConfigurationValue) {
+    window
+      .getConfigurationValue<string>(AppearanceSettings.SectionName + '.' + AppearanceSettings.Appearance)
```
benoitf: use `?.` like `window?.getConfigurationValue`

the key can be a constant initialized once for all

IMHO use await for more readability rather than chaining too many then clause.

(use an async function)

packages/renderer/src/stores/appearance.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (value) {
+          appearance.set(value);
+        }
+      })
+      .then(() => updateIsDark());
+  }
+});
+
+function updateIsDark() {
+  if (get(appearance) === AppearanceSettings.SystemEnumValue) {
```
benoitf: looks odd to retrieve the appearance value using get while when updateIsDark is called we have the value so maybe provide the apperance value to the function ?

benoitf (commented):
it looks like a look of tests are failing

packages/renderer/src/stores/appearance.ts
```diff
@@ -0,0 +1,49 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { type Writable, writable } from 'svelte/store';
+
+import { AppearanceSettings } from '../../../main/src/plugin/appearance-settings';
+import { configurationProperties } from './configurationProperties';
+
+export const isDark: Writable<boolean> = writable(false);
+const appearance: Writable<undefined | string> = writable();
```
benoitf: wondering why it's needed ?

gastoner: Yea its not, removing it now

gastoner:
@benoitf @axel7083 @feloy PTAL

packages/renderer/src/lib/appearance/Appearance.svelte
```diff
@@ -1,18 +1,20 @@
...
+import { isDark } from '../../stores/appearance';
 import { onDidChangeConfiguration } from '../../stores/configurationProperties';
-import { AppearanceUtil } from './appearance-util';
 
 const APPEARANCE_CONFIGURATION_KEY = AppearanceSettings.SectionName + '.' + AppearanceSettings.Appearance;
 async function updateAppearance(): Promise<void> {
   const html = document.documentElement;
 
+  // Wait for isDark to get updated
+  await new Promise(resolve => setTimeout(resolve, 0));
```
feloy: I don't think this should be used here. You should find a way to react on `$isDark` change

gastoner: This will get called when is the appearance changed, but the `isDark` will return previous value, since it doesn't have enough time to change its value. 

With this change if you e.g. change from light->dark the `isDark` will be before line 14 e.g. `false`, and after line `14` it will be `true`. Withou this, the `get(isDark)` would return `false`

packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.spec.ts
```diff
@@ -245,6 +248,8 @@ test('Expects images.icon.dark option to be used when theme is dark', async () =
       sessionRequests: [],
     },
   ];
+  (window as any).getConfigurationValue = vi.fn().mockResolvedValue(AppearanceSettings.DarkEnumValue);
+  configurationProperties.set([]);
```
feloy: Try to mock `getImage` instead, or better the `IconImage` component

packages/renderer/src/lib/image/ImageColumnName.spec.ts
```diff
@@ -106,6 +108,7 @@ test('Expect badge with simple color', async () => {
 
 test('Expect badge with dark color', async () => {
   (window as any).getConfigurationValue = vi.fn().mockResolvedValue(AppearanceSettings.DarkEnumValue);
+  configurationProperties.set([]);
```
feloy: try to mock a nearer dependency, to avoid relying on the implementation of the `isDark` store

packages/renderer/src/lib/editor/MonacoEditor.svelte
```diff
@@ -37,7 +38,6 @@ onMount(async () => {
 
   // check if we're in light or dark mode and get the terminal background color
   const appearanceUtil = new AppearanceUtil();
-  const isDark = await appearanceUtil.isDarkMode();
```
feloy, line 40: you will need to replace isDark with $isDark

packages/renderer/src/lib/appearance/IconImage.spec.ts
```diff
@@ -39,6 +41,7 @@ beforeEach(() => {
 
 test('Expect valid source and alt text with dark mode', async () => {
   getConfigurationValueMock.mockResolvedValue(AppearanceSettings.DarkEnumValue);
+  configurationProperties.set([]);
```
feloy: or you could mock the getImage function

packages/renderer/src/lib/appearance/Appearance.spec.ts
```diff
@@ -85,14 +88,16 @@ test('Expect dark mode using system when OS is set to dark', async () => {
...
 
   const { baseElement } = await awaitRender();
   // expect to have class being "dark" as OS is using dark
-  expect(getRootElementClassesValue(baseElement)).toBe('dark');
+  await vi.waitFor(() => expect(getRootElementClassesValue(baseElement)).toBe('dark'));
 });
 
 test('Expect light mode using light configuration', async () => {
   getConfigurationValueMock.mockResolvedValue(AppearanceSettings.LightEnumValue);
+  configurationProperties.set([]);
```
feloy, line 101: add a setTimeout here, to let time to the store to emit the event before to build the component and call onMount => updateAppearance (all this synchronously)

feloy (changes requested):
The PR breaks the "system" appearance:

When you switch to the system appearance, the latest selected one is kept (if you were in dark mode, selecting the system mode displays the dark mode, even if the system mode is light)
changing the system mode does not change the Podman Desktop mode reactively (it was changing it before)
Before:

https://github.com/user-attachments/assets/e16b1b1f-0888-4ded-b933-ae723668c0fa



After:

https://github.com/user-attachments/assets/4f6167ea-0ee9-4056-96cf-9a4e0471d702

axel7083 (dismissed):
## On your branch

The store do not get updated if the user change the dark mode from the system

image

https://github.com/user-attachments/assets/bd4b8686-3244-411a-b7d5-1bb232c55aed

Probably need to take a look at https://github.com/containers/podman-desktop/pull/8485

## On main branch

https://github.com/user-attachments/assets/22155b28-3846-478f-8dcf-ada2f4faeb39

packages/renderer/src/lib/appearance/Appearance.svelte
```diff
@@ -1,18 +1,17 @@
...
 
 const APPEARANCE_CONFIGURATION_KEY = AppearanceSettings.SectionName + '.' + AppearanceSettings.Appearance;
 async function updateAppearance(): Promise<void> {
   const html = document.documentElement;
 
   // toggle the dark class on the html element
-  const appearanceUtil = new AppearanceUtil();
-  let isDark = await appearanceUtil.isDarkMode();
-  if (isDark) {
+  if (get(isDark)) {
```
benoitf: need to subscribe to is dark and then apply all the code when it's changing ( and not use get)

benoitf (commented):
I also don't get why we have all the setTimeout/0 in tests

Either it's a tick() or a waitFor condition

packages/renderer/src/lib/editor/MonacoEditor.svelte
```diff
@@ -37,12 +39,11 @@ onMount(async () => {
...
   const bgColor = appearanceUtil.getColor('--pd-terminal-background');
 
   // create a theme with the current light or dark mode, but customize the background color
   Monaco.editor.defineTheme('podmanDesktopTheme', {
-    base: isDark ? 'vs-dark' : 'vs',
+    base: get(isDark) ? 'vs-dark' : 'vs',
```
benoitf: need to subscribe else it's only done once

packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.spec.ts
```diff
@@ -227,24 +240,28 @@ test('Expects images.icon option to be used when no themes are present', async (
...
 test('Expects images.icon.dark option to be used when theme is dark', async () => {
+  const darkIcon = './icon-dark.png';
   const providerWithImageIcon = [
     {
       id: 'test',
       displayName: 'Test Authentication Provider',
       accounts: [],
       images: {
         icon: {
-          dark: './icon-dark.png',
```
benoitf: you should extract this kind of refactoring to a separate very small PR that we can merge before this PR

gastoner:
@axel7083 @feloy the PR was not rebased, could you try again please?

packages/renderer/src/lib/appearance/appearance-util.ts
```diff
@@ -64,10 +48,10 @@ export class AppearanceUtil {
       return icon;
     }
 
-    const isDark: boolean = await this.isDarkMode();
-    if (isDark && icon.dark) {
+    const isDarkTheme = get(isDark);
```
benoitf: not for this PR, but we need to create a follow-up issue for the getTheme method
it will need to be transformed to a reactive pattern as well.

as here you're using the get method and not the subscribe it means it's only done once.

packages/renderer/src/lib/editor/MonacoEditor.svelte
```diff
@@ -64,10 +67,17 @@ onMount(async () => {
     // Emit the content change so we can use it in the parent component
     dispatch('contentChange', editor.getValue());
   });
+
+  isDarkUnsubscribe = isDark.subscribe(value => {
+    isDarkTheme = value;
```
feloy: You are updating `isDarkTheme` when `isDark` emits a new value, but I don't think that the new value of `isDarkTheme` is used anywhere (the only place it is used is in `onMount`, which won't be run again.

packages/renderer/src/lib/appearance/appearance-util.ts
feloy: You are exploiting that isDark is a store.
I suppose that we need to transform the `getImage` function with a function to which the caller can subscribe to get new values of the image when `isDark` changes (by subscribing to `isDark` here, and re-computing the image and re-emitting its value)

axel7083 (commented):
The theme system is fixed, but the monaco editor seems to bit problematic

https://github.com/user-attachments/assets/016f902d-529e-46d9-8163-558935672179

packages/renderer/src/lib/appearance/appearance-util.ts
feloy: (same as Florent's comment)

gastoner: Oh I've probably forgot to change this too

Is there a way/need to unsubscribe in `.ts` file? In `.svelte`, I was using onDestroy, but I think that TS is using a garbage collector, so the unsubscribe method is probably not needed

gastoner:
https://github.com/user-attachments/assets/9a74e8f3-9ec7-4abd-9c20-00ba53b597e2

The change in MonacoEditor seems quite slow, but it will eventually reflect the changes. 
Looks like `isDarksubscribe` is taking some time to get the change
@axel7083 @feloy

axel7083 (approved):
LGTM

packages/renderer/src/lib/editor/MonacoEditor.svelte
```diff
@@ -13,15 +13,35 @@ let divEl: HTMLDivElement;
...
+    rules: [{ token: 'custom-color', background: bgColor }],
+    colors: {
+      'editor.background': bgColor,
+    },
+  });
+}
+
 onMount(async () => {
+  isDarkUnsubscribe = isDark.subscribe(value => {
+    updateTheme(value);
```
feloy, line 42: On my system, the terminal theme is not updated when I change the system theme, even if I way a few minutes.

I realize that `isDark` emits events when the configuration changes, but not when the system theme changes

feloy: I suppose this can be part of another PR
Approving these changes

feloy (approved):
Thanks for all the work on this issue
LGTM

PR #8764, opened 2024-09-05
Refactor pd runner and rename

### LINKED ISSUE #8749: Rename and refactor the pdRunner name in the test framework to Runner and export it as such

Currently the pdRunner class is exported as PodmanDesktopRunner, it's been requested [here](https://github.com/containers/podman-desktop/pull/8745#discussion_r1743900236) that this be refactored and exported as just Runner, since it's already part of the podman desktop test framework, so the additional naming does not make sense.

the podman desktop runner class should be renamed to Runner and exported as such.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Refactors pdRunner by renaming and creating a DTO for the configuration.

https://github.com/containers/podman-desktop/issues/8749

### DIFF
--- a/tests/playwright/src/index.ts
+++ b/tests/playwright/src/index.ts
@@ -19,6 +19,7 @@
 // export core modules
 export * from './globalSetup/global-setup';
 export * from './runner/podman-desktop-runner';
+export * from './runner/runner-options';
 export * from './setupFiles/setup-registry';
 export type { RunnerTestContext } from './testContext/runner-test-context';
 export * from './utility/cleanup';
--- a/tests/playwright/src/runner/runner-options.ts
+++ b/tests/playwright/src/runner/runner-options.ts
@@ -54,13 +54,21 @@ export class RunnerOptions {
   public createSettingsJson(): string {
     console.log(`Binary path: ${this._binaryPath}`);
 
-    return (
-      JSON.stringify({
+    if (this._binaryPath) {
+      return JSON.stringify({
         'preferences.OpenDevTools': this._openDevTools,
         'extensions.autoUpdate': this._autoUpdate,
         'extensions.autoCheckUpdates': this._autoCheckUpdates,
         'extensions.disabled': this._extesionsDisabled,
-      }) + (this._binaryPath ? `,'podman.binary.path': ${this._binaryPath}` : '')
-    );
+        'podman.binary.path': this._binaryPath,
+      });
+    }
+
+    return JSON.stringify({
+      'preferences.OpenDevTools': this._openDevTools,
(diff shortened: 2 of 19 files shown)


tests/playwright/src/index.ts
```diff
@@ -19,6 +19,7 @@
 // export core modules
 export * from './globalSetup/global-setup';
 export * from './runner/podman-desktop-runner';
+export * from './runner/runner-options';
```
benoitf, line 22: I don't know if you want to export Runner under the name PodmanDesktopRunner for a couple of weeks the time to have changes in the consumer side

benoitf (approved):

tests/playwright/src/runner/runner-options.ts
```diff
@@ -0,0 +1,58 @@
...
+    this._extesionsDisabled = extesionsDisabled;
+  }
+
+  public createSettingsJson(): string {
+    return JSON.stringify({
+      'preferences.OpenDevTools': this._openDevTools,
+      'extensions.autoUpdate': this._autoUpdate,
+      'extensions.autoCheckUpdates': this._autoCheckUpdates,
+      'extensions.disabled': this._extesionsDisabled,
+    });
```
ScrewTSW, line 72: I also need `podman.binary.path` as a parameter
Could you please add it?

tests/playwright/src/specs/extension-installation.spec.ts
```diff
@@ -50,8 +50,8 @@ let ociImageUrl: string;
 let navigationBar: NavigationBar;
 
 async function _startup(extensionName: string): Promise<void> {
-  pdRunner = await PodmanDesktopRunner.getInstance();
-  page = await pdRunner.start();
+  pdRunner = await Runner.getInstance();
```
ScrewTSW, line 53: Why is it called `pdRunner` here?

ScrewTSW (changes requested):

tests/playwright/src/specs/extension-installation.spec.ts
cbr7: Because that is local to that test, it's not from the fixture, that test suite calls has a very weird flow where the beforeAll/afterAll hooks are called as part of some function, I didn't make any changes to that test file that I didn't absolutely need, which is why it's not using fixtures.

tests/playwright/src/runner/runner-options.ts
cbr7: added

ScrewTSW (approved):
LGTM, great work!

PR #8770, opened 2024-09-06
fix: adds top padding to tab bar in detail page based on slot.detail content

### LINKED ISSUE #8574: cpu/memory widget inside container details is overlapping
### Bug description

Overlap:
image

The cpu / memory widget is cropped at the bottom for the terminal

image

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

go to the terminal details

### Relevant log output

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds top padding to the tab in detail page based on the height of the content of `slot.detail`

### Screenshot / video of UI
prev: 
image
image

after:

https://github.com/user-attachments/assets/baef965a-1481-43e8-a9fb-8ef8e77c5982

Closes #8574 

### How to test this PR?
Run container -> Container Detail 
Assert that cpu/memory usage is not overlapping with the content below

### DIFF
--- a/packages/ui/src/lib/layouts/Page.spec.ts
+++ b/packages/ui/src/lib/layouts/Page.spec.ts
@@ -20,7 +20,7 @@ import '@testing-library/jest-dom/vitest';
 
 import { fireEvent, getByRole, render, screen } from '@testing-library/svelte';
 import userEvent from '@testing-library/user-event';
-import { beforeEach, expect, test, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import Page from './Page.svelte';
 
@@ -146,4 +146,45 @@ test('Expect subtitle is defined and cut', async () => {
 
   // expect class has the clamp
   expect(subtitleElement).toHaveClass('line-clamp-1');
+});
+
+describe('Check for MutationObserver', () => {
+  let mutationObserverObserve: typeof MutationObserver.prototype.observe | undefined;
+  let mutationObserverDisconnect: typeof MutationObserver.prototype.disconnect | undefined;
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mutationObserverObserve = MutationObserver.prototype.observe;
+    mutationObserverDisconnect = MutationObserver.prototype.disconnect;
+
+    MutationObserver.prototype.observe = vi.fn();
+    MutationObserver.prototype.disconnect = vi.fn();
+  });
+
+  afterEach(() => {
+    if (!mutationObserverObserve || !mutationObserverDisconnect) {
+      return;
+    }
+    MutationObserver.prototype.observe = mutationObserverObserve;
+    MutationObserver.prototype.disconnect = mutationObserverDisconnect;
+  });
+
+  test('Expect disconected MutationObserver after onDestroy', async () => {
+    const subtitle = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit';
+    const page = render(Page, {
+      title: '',
+      subtitle,
+    });
+
+    const subtitleElement = screen.getByText(subtitle);
+    expect(subtitleElement).toBeInTheDocument();
+
+    expect(MutationObserver.prototype.observe).toHaveBeenCalledTimes(1);
+    expect(MutationObserver.prototype.disconnect).toHaveBeenCalledTimes(0);
+
+    // run unmount to trigger the onDestroy callback
+    page.unmount();
+
+    expect(MutationObserver.prototype.disconnect).toHaveBeenCalledTimes(1);
+  });
 });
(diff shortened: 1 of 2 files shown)


axel7083 (commented):
I feel like in other pages it add useless white space 

image

image

Maybe the tabs should have some kind of sizing depending on the content

gastoner:
@axel7083 I was trying to add the top-padding only when is `$$slots.detail` present, but it looks like there is always detail even if the detail is invisible, so maybe if the content of the detail is visible or `!= ""`, but I'm not sure if this can be done

jeffmaury (approved):
LGTM

axel7083 (commented):
I like the reactive solution, and as we discussed that's probably the _easiest_ way of doing it without completely changing the layout.

Could we have a test for that ? At least checking that the observer is properly disconnected.

gastoner:
@axel7083 Yes, I agree; I just don't know about any prettier approach how to detect that the content in the slot has changed. But I'll add the test!

packages/ui/src/lib/layouts/Page.spec.ts
```diff
@@ -147,3 +148,29 @@ test('Expect subtitle is defined and cut', async () => {
...
+    subtitle,
+  });
+
+  const subtitleElement = screen.getByText(subtitle);
+  expect(subtitleElement).toBeInTheDocument();
+  expect(mutationObserverMock.mock.results[0].value.observe).toHaveBeenCalledTimes(1);
+  expect(mutationObserverMock.mock.results[0].value.disconnect).toHaveBeenCalledTimes(0);
+  page.unmount();
+  expect(mutationObserverMock.mock.results[0].value.disconnect).toHaveBeenCalledTimes(1);
+  expect(mutationObserverMock.mock.results[0].value.disconnect).not.toHaveBeenCalledTimes(2);
```
axel7083:
```suggestion
test('Expect disconected observer after onDestroy', async () => {
  const subtitle = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit';

  const page = render(Page, {
    title: '',
    subtitle,
  });

  const subtitleElement = screen.getByText(subtitle);
  expect(subtitleElement).toBeInTheDocument();
  expect(MutationObserver.prototype.observe).toHaveBeenCalledTimes(1);
  expect(MutationObserver.prototype.disconnect).toHaveBeenCalledTimes(0);
  // run unmount to trigger the onDestroy callback
  page.unmount();
  expect(MutationObserver.prototype.disconnect).toHaveBeenCalledTimes(1);
});
```

I would avoid changing the global object without restoring it.

You could add at the begining of the test

```ts
let mutationObserverObserve: typeof MutationObserver.prototype.observe | undefined;
let mutationObserverDisconnect: typeof MutationObserver.prototype.disconnect | undefined;

beforeAll(() => {
  mutationObserverObserve = MutationObserver.prototype.observe;
  mutationObserverDisconnect = MutationObserver.prototype.disconnect;

  MutationObserver.prototype.observe = vi.fn();
  MutationObserver.prototype.disconnect = vi.fn();
});

afterAll(() => {
  if(!mutationObserverObserve || !mutationObserverDisconnect) {
    return;
  }
  MutationObserver.prototype.observe = mutationObserverObserve;
  MutationObserver.prototype.disconnect = mutationObserverDisconnect;
});
```

packages/ui/src/lib/layouts/Page.spec.ts
```diff
@@ -147,3 +150,44 @@ test('Expect subtitle is defined and cut', async () => {
   // expect class has the clamp
   expect(subtitleElement).toHaveClass('line-clamp-1');
 });
+
+describe('Check for MutationObserver', async () => {
```
axel7083:
```suggestion
describe('Check for MutationObserver', () => {
```

packages/ui/src/lib/layouts/Page.svelte
```diff
@@ -97,7 +116,10 @@ function handleKeydown(e: KeyboardEvent): void {
   {#if inProgress}
     <LinearProgress />
   {/if}
-  <div class="flex flex-row px-2 border-b border-[var(--pd-content-divider)]" aria-label="Tabs" role="region">
+  <div
+    class="flex flex-row px-2 {`pt-${heightOfDetail > 50 ? 4 : 0}`} border-b border-[var(--pd-content-divider)]"
```
axel7083: This would not scale very well as tailwind classes are generated at compile time

image

(ref https://v2.tailwindcss.com/docs/just-in-time-mode)

So you need to use the style here instead of class

axel7083: and probably better to use height property instead of padding?

gastoner:
@axel7083 PTAL

packages/ui/src/lib/layouts/Page.spec.ts
```diff
@@ -15,12 +15,13 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
axel7083: what for ?

axel7083 (approved):
LGTM

PR #8786, opened 2024-09-06
feat: search image

### LINKED ISSUE #971: Ability to browse images on remote repositories

feloy:
### What does this PR do?

Pull image with typeahead using a custom component for searching an image (inspired by https://github.com/pstanoev/simple-svelte-autocomplete)

TODO:
- [x] display loading...
- [x] close list when clicking outside
- [x] unit tests 

### Screenshot / video of UI

https://github.com/user-attachments/assets/61e05b60-04aa-4770-98d7-7bab08f993b7

Fixes #971 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/Typeahead.spec.ts
+++ b/packages/renderer/src/lib/ui/Typeahead.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/renderer/src/lib/ui/Typeahead.spec.ts
+++ b/packages/renderer/src/lib/ui/Typeahead.spec.ts
@@ -1,0 +1,227 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import '@testing-library/jest-dom/vitest';
+
+import { render, screen, within } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
+import { tick } from 'svelte';
+import { expect, test } from 'vitest';
+
+import Typeahead from './Typeahead.svelte';
+
+window.HTMLElement.prototype.scrollIntoView = function () {};
+
+function assertIsListVisible(visible: boolean): void {
+  if (visible) {
+    const list = screen.getByRole('row');
+    expect(list).toBeDefined();
+  } else {
+    const list = screen.queryByRole('row');
+    expect(list).toBeNull();
+  }
+}
+
+function assertItemSelected(items: HTMLElement[], r: number): void {
+  for (let i = 0; i < items.length; i++) {
+    if (i === r) {
+      expect(items[i]).toHaveClass('bg-[var(--pd-content-card-hover-bg)]');
+    } else {
+      expect(items[i]).not.toHaveClass('bg-[var(--pd-content-card-hover-bg)]');
+    }
+  }
+}
+
+test('a textbox is created', async () => {
+  render(Typeahead);
+
+  const input = screen.getByRole('textbox');
+  expect(input).toBeDefined();
+});
+
+test('placeholder is set', async () => {
+  render(Typeahead, {
+    placeholder: 'a placeholder',
+  });
+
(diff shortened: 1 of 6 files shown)


feloy:
@deboer-tim could you please have a look at the UX of this implementation?
cc @ekidneyrh

deboer-tim:
I love it, to me this is much simpler and easy to use, and natural to developers. Just a couple nits:

1. There doesn't seem to be a way to close the dropdown without selecting something and it blocks the Pull Image button. This is annoying if I think I've typed a valid image and want to pull w/o picking from the list, but worse if I come in and paste a valid image like 'quay.io/strimzi/kafka' since I need that extra click to 'reselect it' before pulling.
a) We should close the dropdown if you click outside of it.
b) Maybe there's something simple we can do when the text is a valid image, e.g. closing the dropdown and putting a checkmark on the right? Or at least not opening it when you paste a valid image?

2. I'm not sure how obvious it is that you need at add ':' to see the valid tags - but maybe that's fine, it's an obvious step and people will figure it out?

feloy:
> I love it, to me this is much simpler and easy to use, and natural to developers. Just a couple nits:
> 
> 1. There doesn't seem to be a way to close the dropdown without selecting something and it blocks the Pull Image button. This is annoying if I think I've typed a valid image and want to pull w/o picking from the list, but worse if I come in and paste a valid image like 'quay.io/strimzi/kafka' since I need that extra click to 'reselect it' before pulling.

In this video, I am pasting `quay.io/soketi/soketi` and pressing enter, without having to select it to the list. Note that when you type a text, the first item of the list is not selected, you need to press Down/PageDown to enter the list. When you are in this state (no item selected), you just need to press Enter to start pulling the image.

https://github.com/user-attachments/assets/7caa489f-6b2e-4d3c-b4ef-b2c4be9f1009


>    a) We should close the dropdown if you click outside of it.

Yes, I'll add this, this is useful. The list can be closed with Esc, or by pressing Up when the first item of the list is highlighted.

>    b) Maybe there's something simple we can do when the text is a valid image, e.g. closing the dropdown and putting a checkmark on the right? Or at least not opening it when you paste a valid image?

I'm not sure, the list could display useful images, for example `quay.io/strimzi/kafka-bridge` when you type `quay.io/strimzi/kafka` which also exists.

> 2. I'm not sure how obvious it is that you need at add ':' to see the valid tags - but maybe that's fine, it's an obvious step and people will figure it out?

cdrage:
Screenshot 2024-09-10 at 8 53 41 AM
Design feedback before going code diving. I don't like how with my private hosted registry I always have the 404 symbol, making me think that my registry is always down :(

I feel as though we should never show an error / issue regarding being able to do an auto-completion? As this is a convenience feature, we shouldn't show 404 errors just for auto-completion in the GUI.

feloy:
> Design feedback before going code diving. I don't like how with my private hosted registry I always have the 404 symbol, making me think that my registry is always down :(
> 
> I feel as though we should never show an error / issue regarding being able to do an auto-completion? As this is a convenience feature, we shouldn't show 404 errors just for auto-completion in the GUI.

The message is stating `Error: searching images` of `Error: getting tags of image`, which should indicate to the user that it's only about autocompletion. Perhaps should we change the icon/color so it is only a warning? info?

If we don't show anything, I'm afraid that users won't understand why sometimes there are autocompletions, sometimes not.

deboer-tim:
@ekidneyrh might know if there's a pattern to follow, but IMHO warning would still look like there's something wrong so I'd lean towards making it info.

benoitf:
I think that on typeahead search input there is a feedback on the text entered when it's recognized and validated (like the data has a background behind it, text color is changed, etc) but nothing if it's not validated/recognized (it's still freeform)

so maybe reverse the notification: nothing if it's not recognized but if recognized it's highlighted

ekidneyrh:
> @ekidneyrh might know if there's a pattern to follow, but IMHO warning would still look like there's something wrong so I'd lean towards making it info.

I like the spinner thats currently there when it's searching for images. When a private image is pulled I wouldn't have anything in the field so it's the same as when pulling an auto completed option. 

When an invalid image is entered and the pull button it pressed an error message pops up and i think that's enough feedback to give to the user that it's invalid. 

I think it looks good!

packages/renderer/src/lib/ui/Typeahead.spec.ts
```diff
@@ -0,0 +1,197 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
deboer-tim: 2024

benoitf (approved):
feature is working for me,

I found potential enhancements, should be follow-up if not part of this PR.

It's mainly how we see completion

if I search for `quay.io/podman`

I see at the top of list some stuff that is not what I typed
image
(for example I see curl while I would expect to see something starting by quay.io/podman 

I see for example some values but they're not at the top (like quay.io/podman/stable)

so I don't know if for example we should also add a filter with a '/' at the end like also searching for `podman/` even if I only typed `podman` (without the trailing /)

when entering `:` for a tag, it seems to be sorted by alphabetical order but I'm wondering if it's convenient, I would like to see the newest by default:
image

I need to scroll down to the bottom to fetch the latest fedora tag


side note:
if I'm not typing `:` I don't see any tags. Perfect. But it means I don't know that I can have completion on the tags (I need to type `:` )
So I'm wondering if we could display the available tags when image is recognized

if I type `quay.io/fedora/fedora`, I could see `:latest`, `:40` as completion items as well (because in that case I have already entered a repository, a owner and an image name so I also have an exact match so it could bring as code completion some image tags)

deboer-tim (approved):

PR #8790, opened 2024-09-06
chore: generate summary of release notes with JSON format to the website

SoniaSandler:

### What does this PR do?
This PR maps the release notes of each version to a separate JSON file that includes a link to the release image, link to the blog post, title, and summary.
The mapped JSON files will be fetched remotely to be used in https://github.com/containers/podman-desktop/pull/8753.
Using this mapping, the link `https://podman-desktop.io/release-notes/<version>.json` will be used to fetch the release notes remotely.
This mapping will be updated every time new release notes are made available.

### How to test this PR?


SoniaSandler:
I wasn't sure about where to place this file, so if there is a better place, please lmk.
Since this file will be fetched from GitHub, its placement in PD doesn't have a significant effect on its use.

benoitf (commented):
I like the fact to have a mapping file

as we need to consume it externally/remotely and it's related to the blog update, I think we would need to move it to the `website/` directory and make it available through http://podman-desktop.io website then

SoniaSandler:
@benoitf any specific place in the `website/` directory?

benoitf:
maybe root folder (on the website) or in the `blogs` directory as it's mainly related to a blog post entry.

or maybe we could generate the file automatically from the blog post article with a plug-in.

so it looks at all blog post files and if there is for a release, it adds the entry to the file

so no need to maintain this file, we just need to follow a convention

SoniaSandler:
@benoitf now the mapping file is populated automatically from the `docusaurus.config.js` file, but I still haven't figured out how to show it on the website since it is a json file (create an md file?). 
However, I'm wondering if this solution is better/worse than using the `podman-desktop.io` notes (or no difference). IIUC, the way it works now includes creating the file when building the website, fetching the mapping from from `podman-desktop.io`, and then the using the mapping to fetch the notes from GitHub. On the other hand, the first solution involves only fetching the notes from `podman-desktop.io`.
I'm open to working on either solution, but I just wanted to check in with you to see if there is a preference or additional comments/suggestions regarding those possible solutions.

benoitf:
do you prefer to generate / add inside the json file the notes directly ?
It has pro/cons but for example fetching all release notes might be lot of data. But maybe not, just a few KB so we could provide in this json directly the data we need to be one exposed m, as long as it is generated from the MD file

SoniaSandler:
@benoitf do you mean instead of mapping to a file name, map the version directly to the summary notes?
If we do continue with the mapping option, it could be a good idea - it will also create a blog entry of the summary of all release notes in one place, but I was wondering about the option of using https://raw.githubusercontent.com/containers/podman-desktop/main/website/blog/2024-08-08-release-1.12.md (need to create mapping -> fetch mapping -> use mapping to fetch the GitHub link (initial idea)) vs https://podman-desktop.io/blog/podman-desktop-release-1.12 (only need to fetch this page)
rn, the other PR has both options

benoitf:
the issue I see with fetching the blog post data is that you're fetching a page that has been rendered, so there is a layout etc and if you break something when publishing the blog post, it's affecting all clients (for example docusaurus could change something, etc)

if the client fetch some data that are within some structure, we can validate all data are correct at PR check time

benoitf:
I think that if you add the JSON file to the static directory it would be reachable

benoitf:
(but need to check if there is some expiration time being set on the files as we want to have newer data when fetching the file)

SoniaSandler:
@benoitf so the fact that the page is rendered is the tricky part?  In this case, I agree that it might be better to use the GH files. For now I'll continue with the mapping version to file names option since the code for parsing the GH notes has already been added to the other PR, and I'll also explore the option of adding the summary notes straight into the JSON file

benoitf:
@SoniaSandler I think that if you're able to generate the file so it's included at the end in the static folder then it might do the trick (as it'll keep the file untouched)


so generate the JSON file including the text and the link to the release notes image
and fetch that file from your widget

SoniaSandler:
@benoitf I updated the mapping by using the original parsing code from the other PR (will be removed from there), and the json file is accessible  from the website.
lmk what you think
Screenshot from 2024-09-10 12-10-55

benoitf:
I like this approach.

Maybe use a shorter name like `releases.json` or `release-notes.json`

remove Url suffix of image

and adds an entry to the release notes blog post

like 
```json
{
'blog': 'https://podman-desktop.io/blog/podman-desktop-release-1.12'
}
```

benoitf:
and the image link should be the one from podman-desktop.io, not from raw.github.com

website/release-notes-parser.ts
```diff
@@ -0,0 +1,53 @@
...
+  summary: string;
+} {
+  const fileContent = fs.readFileSync(filename, { encoding: 'utf-8' });
+  const resultText = fileContent.split('---', 3);
+
+  // get release image url
+  const imagePath = /image: (.+)\n/.exec(resultText[1]);
+  const imageName = imagePath ? imagePath[1] : '';
+  const imageUrl = imageName ? `https://podman-desktop.io${imageName}` : '';
+  const blogUrl = `https://podman-desktop.io/blog/podman-desktop-release-${releaseVersion}`;
```
benoitf: shouldn't we look at the slug field from the markdown file ?

example with https://raw.githubusercontent.com/containers/podman-desktop/main/website/blog/2024-08-08-release-1.12.md

`slug: podman-desktop-release-1.12`

website/release-notes-parser.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // get summary part of release notes
+  const text = resultText[2]
+    .split('<!--truncate-->', 1)[0]
+    .replace("import ReactPlayer from 'react-player'\n", '') //remove react import statement
+    .replace(/!\[.+\)\n/, '') // remove image link
+    .replace(/\[Click here to download it\]\(.+\)!/, '') //remove download new version
+    .replace(/\n(\n)+/g, '\n')
+    .split('\n'); // change all multi-newlines chars to one newline char
```
benoitf: should it be any import not only ReactPlayer ?

if no truncate split, should we consider it as an error and stop/fail the build (as what will happen ?)

SoniaSandler:
@benoitf done
Screenshot from 2024-09-10 12-32-47

benoitf (commented):
as it seems you need to iterate on each blog file (it's not a hook called once)

open question: should we instead of generating one json file containing all the entries, generate one file per release

```
/static/release-notes/1.11.0.json
/static/release-notes/1.12.0.json
```

website/release-notes-parser.ts
SoniaSandler: @benoitf I'll change it to exclude any import. Regarding the stop/fail build without  truncate, I think that could be a good idea since without this information, there will be no way to present the release notes banner. However, I'm not sure how significant is release notes banner because we might decide that we HAVE to show it with each release, so there is no option of building without the correct info/template, but if we decide that it'll be ok not showing them if there is an error, we can continue with the build as intended and add checks in the renderer to see if the info is available, and if not, just not show the banner.

benoitf: but how do you will consider that the data is "small enough" to be displayed for a release ?

like shouldn't it break the build if in the release notes, `<truncate>` is for example after too many lines or missing ?

else it would mean that client will have to display 1000 lines or the whole release notes.

SoniaSandler: @benoitf since we only display the title and bullet points from the notes, I can limit how many lines to add to the json file for L48-49, for example take/use the first n lines from `summary` at L48.

SoniaSandler:
@benoitf separating it into different json files for each release could be a good idea, as this single file will grow with every release, so when we fetch, we only get the info we need (although I'm not sure how significant is difference between the sizes of the separate files vs a single file or how impactful it is when fetching).

I'm not sure I understand
> as it seems you need to iterate on each blog file (it's not a hook called once)

isn't `parseFrontMatter` doing it already?

benoitf:
>> as it seems you need to iterate on each blog file (it's not a hook called once)
>isn't parseFrontMatter doing it already?

From what I see, you open and save the file for each file we're handling

it's not:
initialize the content, then grab all the chunks and at the end, write the file

This is why I was suggesting to write one file per call (you analyze the file, extract the data and write the chunk to a file)

SoniaSandler:
@benoitf like you suggested, I can create a separate json file for each release, and then I can also check if this file already exists, so that we won't rewrite every file during every build (if `/static/release-notes/1.12.0.json` already exists, no need to do `parseNotes` and rewrite it), and only do the `parseNotes` for new `/static/release-notes/<version>` that don't exist and we create them there. Wdyt? or would it be better to rewrite all of them every time if there is the possibility that they're updated (does it even happen? updating old notes)?

benoitf:
I think it's ok to rewrite
It'll be git that will check if the content is the same or not

SoniaSandler:
@benoitf now there is a separate json file for each release. I also removed the `<truncate>` part since the way I do the parsing now has changed a bit from my original idea and there is no need for that anymore (there are also a few existing release notes without `<truncate>` in them, but it still worked). I also limited the number of bullet points in the summary section to 4, but it is a number we can change if needed.

benoitf (commented):
I think the files should not be committed. They'll be included in the build of the website when building the website, so as generated files, we don't include them in the git repository

benoitf (commented):
you need to add unit tests for the parser file and then I think we'll be ready to review/merge

website/release-notes-parser.ts
```diff
@@ -0,0 +1,52 @@
...
+  const folderName = './static/release-notes';
+  const fileContent = fs.readFileSync(filename, { encoding: 'utf-8' });
+  const resultText = fileContent.split('---', 3);
+
+  // get release image url
+  const imagePath = /image: (.+)\n/.exec(resultText[1]);
+  const imageName = imagePath ? imagePath[1] : '';
+  const imageUrl = imageName ? `https://podman-desktop.io${imageName}` : '';
+  const pageName = /slug: (.+)\n/.exec(resultText[1]);
+  const blogName = pageName ? pageName[1] : `podman-desktop-release-${releaseVersion}`;
```
benoitf: I think my comment may have been lost, but here the URL will depend on the slugName of the blog post, here we assume it'll be construct following a given scheme but it may change

shouldn't we look at the slug field from the markdown file ?

example with https://raw.githubusercontent.com/containers/podman-desktop/main/website/blog/2024-08-08-release-1.12.md
extract the `slug` being `slug: podman-desktop-release-1.12`

website/docusaurus.config.js
```diff
@@ -24,6 +25,23 @@ const config = {
...
+
+      if (
+        result.frontMatter.title &&
+        String(result.frontMatter.title).match(/Release/) &&
+        String(result.frontMatter.title).match(/\d+\.\d+/)
+      ) {
+        let versionMatch = String(result.frontMatter.title).match(/(\d+\.\d+)/) ?? [];
+        let version = versionMatch[1] ? versionMatch[1] : '';
+        if (version) {
+          parseNotes(params.filePath, version);
```
benoitf: maybe adapt/change the method name as with a method parse I'm expecting a result while here it's also writing a file.

Also parseNotes should be async (then we await)

website/release-notes-parser.ts
```diff
@@ -0,0 +1,52 @@
...
+  const summaryText = summary.slice(0, 4).join('\n'); // limit the number of bullet points to 4
+  const titleText = text.filter(line => !line.includes('import') && line)[0];
+
+  const jsonInput = { image: imageUrl, blog: blogUrl, title: titleText, summary: summaryText };
+
+  if (!fs.existsSync(folderName)) {
+    fs.mkdirSync(folderName);
+  }
+
+  fs.writeFileSync(`${folderName}/${releaseVersion}.json`, JSON.stringify(jsonInput));
```
benoitf: we should use the promises async versions rather than the sync methods

website/docusaurus.config.js
benoitf: here, this code is not tested at all, so should we move more stuff to the ts file and just call from the .js file a function

```
  parseFrontMatter: async params => {
      myFunction(params);
  }
```

and then it's inside the function that you do all the stuff (and then you have unit tests)

website/release-notes-parser.ts
SoniaSandler: @benoitf  L30 extracts slug, but if slug doesn't exists, it will use the given scheme. If we know for sure that slug is always given, I can remove L31

benoitf (commented):
as there was no test in website folder, there is no script in package.json to launch the test in `website` folder

so calling `pnpm test` is never triggering the tests and then it means that PR check is green even if tests might not pass.

(it should be a separate PR bringing the launch of unit tests in website folder (with `--passWithNoTests` option)

then, trying to run the tests manually, I faced:

```

 ❯ release-notes-parser.spec.ts (3)
   × create release-notes directory when it does not exist
   × Do not create release-nnotes directory when it exist
   × parse provided release notes as expected

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 3 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  release-notes-parser.spec.ts > create release-notes directory when it does not exist
TypeError: params.defaultParseFrontMatter is not a function
 ❯ Module.createNotesFiles release-notes-parser.ts:32:31
     30|   },
     31| ): Promise<ParseFrontMatterResult> {
     32|   const result = await params.defaultParseFrontMatter(params);
       |                               ^
     33|   // eslint-disable-next-line sonarjs/slow-regex
     34|   const versionRegex = /\d+\.\d+/;
 ❯ release-notes-parser.spec.ts:66:9
```

SoniaSandler:
@benoitf I'm currently working on updating the tests for the most recent commit. The ones that are available in this PR were pushed before the refactoring

benoitf:
Ok maybe we keep the PR in draft mode then

SoniaSandler:
@benoitf the tests have been updated and I opened https://github.com/containers/podman-desktop/pull/8850 for the testing scripts

benoitf:
ok you'll need to rebase this PR (to include the commit of https://github.com/containers/podman-desktop/pull/8850)

website/package.json
```diff
@@ -26,8 +26,8 @@
     "@docusaurus/core": "^3.5.2",
     "@docusaurus/plugin-client-redirects": "^3.5.2",
     "@docusaurus/preset-classic": "^3.5.2",
-    "@docusaurus/theme-mermaid": "^3.5.2",
```
benoitf: this change is out of the scope of this PR

(moving it few lines after)

benoitf (approved):

website/package.json
SoniaSandler: @benoitf I changed it back to the original order

benoitf (approved):

cdrage (approved):
Great work on the tests, I tested this locally and it's correctly mapping from the looks of it. Going through the code I have no suggestions.

LGTM

benoitf:
@SoniaSandler if you can update the description of this PR as I think it's not up-to-date

(What does this PR do?)

PR #8791, opened 2024-09-07
feat: move podman extension to packages/extension folder and add dev and runtime support for it

dgolovin:
### What does this PR do?

This PR adds support for extensions with API, when extension folder have multiple packages api and extension itself under extensions/ext-name/packages folder.

podman extension structure changed to new one by moving extensions/podman to extensions/podman/packages/extension folder.

This PR is extracted from #8045 to simplify review.

### How to test this PR?

1. Pull the branch
2. `git status` should not show any folders form `extensions/podman/` as untracked
3. `git -C extensions/podman/ clean -dxf` to remove ignored files left after checkout
1. pnpm typecheck:extensions:podman
4. pnpm lint:check
5. pnpm build:extensions:podman
6. pnpm test:extensions:podman
7. pnpm watch
8. build inastaller and run it to test in production environment or pull the branch https://github.com/dgolovin/podman-desktop/tree/allow-to-run-in-electron with rebase on top of this one. run `pnpm build` && `npx electron .`


benoitf:
doing `pnpm watch` command I see:

```
[vite]: Rollup failed to resolve import "ps-list" from "/Users/benoitf/git/containers/podman-desktop-6/extensions/podman/packages/extension/src/podman-cleanup-macos.ts".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
```

benoitf:
I don't know how to handle the 'intermediate' step but I had previously a working podman desktop

then it's still trying to use my 'old' dist folder
image

maybe need to handle the case where we have both 'old layout' and 'new layout' as it will occur for a lot of developers (not a fresh clone)

benoitf:
step 5: `pnpm yarn watch` (remove yarn)

dgolovin:
> step 5: `pnpm yarn watch` (remove yarn)

Updated.

dgolovin:
It looks like I forgot to commit pnpm-workspace.yaml.
Added.

benoitf:
looks like the lockfile is not correct

dgolovin:
> looks like the lockfile is not correct

@benoitf  updated.

dgolovin:
@benoitf I went through the scenario you pointed out and I did replicate the issues. It turned out pnpm-wokspace.yaml change was lost. After adding new pattern for podman ext, I did not see the problem anymore.

After pulling the branch there are some non git folders left behind in extension/podman folder. They seems not be harmless and can be deleted manually. I can add cleanup code to the build so it would check for node_modules, dist and cdix in previous location and get rid of them.

benoitf:
@dgolovin maybe we could look at .gitignore file if exists and exclude from scanned folders these folders (extension-loader)

so we won't commit again these files (after the move) and we won't try to load extensions from these files

dgolovin:
@benoitf The extensions/name/packages/extension/ takes precedence over old location, so extensions/podman is not scanned in dev or prod.

benoitf (approved):
extensions/podman/assets/ needs to be added to .gitignore as after updating my branch I have these files remaining

dgolovin:
> extensions/podman/assets/ needs to be added to .gitignore as after updating my branch I have these files remaining

Does 'these file remaining' means they are reported as untracked in `git status` or they are still present in filesystem?

benoitf:
untracked in git but present on my filesystem

dgolovin:
The only thing we can fix is to remove them from untracked in `git status`.  Then `git clean -xdf' will do the job.

PR #8799, opened 2024-09-09
feat: add initialFocus to Input

### LINKED ISSUE #8798: Initial focus on Input component

I would like to be able to set the focus on the `Input` when loading a page.

Add an optional `initialFocus` to the `Input` component
 

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

add initialFocus to Input

Fixes #8798 

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/inputs/Input.spec.ts
+++ b/packages/ui/src/lib/inputs/Input.spec.ts
@@ -163,3 +163,21 @@ test('Expect inputClass styling', async () => {
   expect(element).toBeInTheDocument();
   expect(element).toHaveClass(thisClass);
 });
+
+test('Expect not have focus', async () => {
+  render(Input, {
+    initialFocus: false,
+  });
+  const element = screen.getByRole('textbox');
+  expect(element).toBeInTheDocument();
+  expect(element).not.toHaveFocus();
+});
+
+test('Expect to have focus', async () => {
+  render(Input, {
+    initialFocus: true,
+  });
+  const element = screen.getByRole('textbox');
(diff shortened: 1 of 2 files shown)


benoitf:
I'm wondering if it's up to the component to provide that as we could have multiple components on a page and it's not possible that they have all this option

feloy:
> I'm wondering if it's up to the component to provide that as we could have multiple components on a page and it's not possible that they have all this option

I expect that it would be the responsability of the caller to set this option on only one component.

I can see on Podman Desktop sources that we are using a native `input` and not an `Input` everywhere we need to set focus (either at creation time or during the lifecycle). Working on #8786, I wonder if I should use a native `input` or an `Input`. In the first case, I will need to duplicate the `Input` style and behaviour, in the second case, I am missing the focus feature.

gastoner (approved):
LGTM

feloy:
> > I'm wondering if it's up to the component to provide that as we could have multiple components on a page and it's not possible that they have all this option
> 
> I expect that it would be the responsability of the caller to set this option on only one component.
> 
> I can see on Podman Desktop sources that we are using a native `input` and not an `Input` everywhere we need to set focus (either at creation time or during the lifecycle). Working on #8786, I wonder if I should use a native `input` or an `Input`. In the first case, I will need to duplicate the `Input` style and behaviour, in the second case, I am missing the focus feature.

Well, finally I can see there are many other events I would like to handle in the input, which are not handled by `Input`. I'll go with a native `input` for now

PR #8800, opened 2024-09-09
feat(renderer): enhance feedback form with github star link

axel7083:
### What does this PR do?

On the feedback form, user selecting the best emoji will have a text suggesting to give a star to the podman-desktop repository.

### Screenshot / video of UI

image

https://github.com/user-attachments/assets/5c51f258-28e4-41ac-85bc-ca4d8ff074ea

Fixes https://github.com/containers/podman-desktop-internal/issues/314

### How to test this PR?

**Manually**

- start podman-desktop
- open feedback form
- select emoji
- assert suggestion visible on best emoji selected

### DIFF
--- a/packages/renderer/src/lib/feedback/SendFeedback.spec.ts
+++ b/packages/renderer/src/lib/feedback/SendFeedback.spec.ts
@@ -20,7 +20,7 @@ import '@testing-library/jest-dom/vitest';
 
 import { fireEvent, render, screen } from '@testing-library/svelte';
 import userEvent from '@testing-library/user-event';
-import { beforeAll, expect, test } from 'vitest';
+import { beforeAll, expect, test, vi } from 'vitest';
 
 import SendFeedback from './SendFeedback.svelte';
 
@@ -32,6 +32,9 @@ beforeAll(() => {
       func();
     },
   };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (window as any).openExternal = vi.fn();
 });
 
 test('Expect that the button is disabled when loading the page', async () => {
@@ -112,4 +115,25 @@ test('Expect sad smiley warns without feedback', async () => {
 
   // and the indication is gone
   expect(warn).not.toBeInTheDocument();
+});
+
+test('Expect GitHub dialog visible when very-happy-smiley selected', async () => {
+  render(SendFeedback, {});
+
+  // expect to have indication why the button is disabled
+  expect(screen.getByText('Please select an experience smiley')).toBeInTheDocument();
+
+  // click on a smiley
+  const smiley = screen.getByRole('button', { name: 'very-happy-smiley' });
+  await fireEvent.click(smiley);
+
+  // and the GitHub star text is visible
+  expect(screen.getByLabelText('Like Podman-Desktop? Give us a star on GitHub')).toBeInTheDocument();
+
+  const link = screen.getByRole('link', { name: 'GitHub' });
+  await fireEvent.click(link);
+
+  await vi.waitFor(() => {
+    expect(window.openExternal).toHaveBeenCalledWith('https://github.com/containers/podman-desktop');
+  });
 });
--- a/packages/renderer/src/lib/feedback/SendFeedback.spec.ts
+++ b/packages/renderer/src/lib/feedback/SendFeedback.spec.ts
@@ -35,6 +35,7 @@ beforeAll(() => {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   (window as any).openExternal = vi.fn();
+  (window as any).telemetryTrack = vi.fn();
 });
 
 test('Expect that the button is disabled when loading the page', async () => {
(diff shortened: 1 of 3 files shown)


deboer-tim (changes requested):
Podman Desktop shouldn't have a dash in it (or maybe we don't need to repeat the name at all?).

When this appears it changes the height of the dialog, and since the dialog is vertically centred it causes everything to move, which is a little jarring. What do you think about moving this to the right of the smileys? Might need slightly shorter text.

axel7083:
@deboer-tim here are some proposal

#### Moving in the footer 

This would have the advantages of the consistency, as other warning messages are also displayed here. But it get it smaller.

image

https://github.com/user-attachments/assets/0f2eae74-d4e6-4848-878f-c181a26987e2

#### Moving next to the emojis

As you suggested we could move it next to the emoji

image

benoitf:
just 2 cents I would replace `like` just by 💜 and have the whole sentence being clickable

axel7083:
> just 2 cents I would replace `like` just by 💜 and have the whole sentence being clickable

image

I feel like making the full sentence clickable would make it too purple and diminish the contrast between the heart and the star with the text. Here is a side by side

image

image

benoitf:
you may also try with a shorter sentence by removing `on GitHub` and `Podman Desktop`

`💜 Like it? Drop us a ⭐` or `💜 Happy with it? Reward us with a ⭐`  `💜 Love it? ⭐ us`

(could be very bad ideas)

axel7083:
> you may also try with a shorter sentence by removing `on GitHub` and `Podman Desktop`

I am open to new suggestion of sentence, but what I feel odd is the full sentence clickable: too purplish

deboer-tim:
For the text, I'd personally be more likely to click if we kept the 'on Github', but including Podman Desktop is long & repetitive. So I'd prefer one of the options like `💜 it? ⭐️us on Github`. (or we can go with the Canadian version and say `Please ⭐️us on Github` 🙂)

How about making the action part (`⭐️us on Github`) a link and not the reset of the sentence? Might be an ok balance with a bigger link but not too much purple.

Out of the options I still prefer making it appear to the right of the smiley - doesn't change the layout and is nice and close to where you clicked to make it appear.

axel7083:
Following the discussion on the UX call of the 12 of September. We came to the conclusion that the message should at the bottom for consistency, and visible for the third and fourth emoji but be slightly different.

Happy emoji: `💜 Like it? Drop us a ⭐ on GitHub` with GitHub clickable (Link)

Very happy emoji:  `💜 Love it? Drop us a ⭐ on GitHub` with GitHub clickable (Link)

axel7083:
I added a telemetry event `feedback.openGithub` to track the usage/impact.

tailwind.config.cjs
```diff
@@ -74,6 +74,7 @@ module.exports = {
         600: tailwindColors.violet[600],
         700: tailwindColors.violet[700],
       },
+      'github-star-icon': '#e3b341'
```
benoitf: it should not go to the tailwind config

axel7083: Do I hardcode the value ?

packages/renderer/src/lib/feedback/SendFeedback.svelte
```diff
@@ -53,6 +61,12 @@ async function sendFeedback(): Promise<void> {
   // close the modal
   hideModal();
 }
+
+async function openGithub(): Promise<void> {
```
benoitf: Github -> GitHub

```suggestion
async function openGitHub(): Promise<void> {
```

packages/renderer/src/lib/feedback/SendFeedback.svelte
```diff
@@ -53,6 +61,12 @@ async function sendFeedback(): Promise<void> {
...
+  hideModal();
+  await window.telemetryTrack('feedback.openGithub');
```
benoitf:
```suggestion
  await window.telemetryTrack('feedback.openGitHub');
```

tailwind.config.cjs
benoitf: you add it into the color-registry class

(it means that an extension would be able to override the color value and we can have different values for light and dark mode)

benoitf: side question: we can't use one value of the amber palette color ?

axel7083: Hummm I used the color from the GitHub website

benoitf: the color on the GitHub website has different values for the dark and light theme and I'm not sure we want to pick up the same exact color ?

so maybe we could pick-up one of the amber color, I'm sure some are close of the color

axel7083:
image

tailwind.config.cjs
```diff
@@ -56,7 +56,7 @@ module.exports = {
     colors: {
       // import colors from the color palette
       ...colorPalette,
-      
+
```
benoitf:
```suggestion
      
```

benoitf: (shouldn't have this file being changed)

benoitf (approved):

tailwind.config.cjs
axel7083: Yeah the linter was harassing me, since this is the way the file should look after lint-staged I had to revert and commit manually :(

packages/renderer/src/lib/feedback/SendFeedback.svelte
```diff
@@ -125,6 +139,15 @@ async function sendFeedback(): Promise<void> {
         <ErrorMessage class="text-xs" error="Please share contact info or details on how we can improve" />
       {:else if smileyRating === 2 && !hasFeedback}
         <WarningMessage class="text-xs" error="We would really appreciate knowing how we can improve" />
+      {:else}
+        <div class="text-[var(--pd-modal-text)] p-1 flex flex-row items-center text-xs">
+          <Fa size="1.125x" class="cursor-pointer" icon={faQuestionCircle} />
+          <span aria-label="Like Podman Desktop? Give us a star on GitHub" class="flex items-center">
+            <Fa class="px-1 text-purple-500" icon={faHeart} />{smileyRating === 3 ? 'Like' : 'Love'} It ? Give us a <Fa
```
feloy: No space in English 

```suggestion
            <Fa class="px-1 text-purple-500" icon={faHeart} />{smileyRating === 3 ? 'Like' : 'Love'} It? Give us a <Fa
```

packages/renderer/src/lib/feedback/SendFeedback.spec.ts
```diff
@@ -113,3 +117,46 @@ test('Expect sad smiley warns without feedback', async () => {
...
+test('Expect message for very-happy-smiley to use love', async () => {
+  const { getByRole, getByLabelText } = render(SendFeedback, {});
+
+  // click on a smiley
+  const smiley = getByRole('button', { name: 'very-happy-smiley' });
+  await fireEvent.click(smiley);
+
+  // and the GitHub star text is visible
+  const region = getByLabelText('Like Podman Desktop? Give us a star on GitHub');
+  expect(region.textContent).toBe('Love It ? Give us a on GitHub');
```
feloy:
```suggestion
  expect(region.textContent).toBe('Love It? Give us a on GitHub');
```

packages/renderer/src/lib/feedback/SendFeedback.spec.ts
```diff
@@ -113,3 +117,46 @@ test('Expect sad smiley warns without feedback', async () => {
...
+test('Expect message for happy-smiley to use like', async () => {
+  const { getByRole, getByLabelText } = render(SendFeedback, {});
+
+  // click on a smiley
+  const smiley = getByRole('button', { name: 'happy-smiley' });
+  await fireEvent.click(smiley);
+
+  // and the GitHub star text is visible
+  const region = getByLabelText('Like Podman Desktop? Give us a star on GitHub');
+  expect(region.textContent).toBe('Like It ? Give us a on GitHub');
```
feloy:
```suggestion
  expect(region.textContent).toBe('Like It? Give us a on GitHub');
```

packages/renderer/src/lib/feedback/SendFeedback.svelte
```diff
@@ -125,6 +139,15 @@ async function sendFeedback(): Promise<void> {
         <ErrorMessage class="text-xs" error="Please share contact info or details on how we can improve" />
       {:else if smileyRating === 2 && !hasFeedback}
         <WarningMessage class="text-xs" error="We would really appreciate knowing how we can improve" />
+      {:else}
```
feloy: this will also appear with 'Love' if rating is 1 or 2 and feedback is done. I suppose it is not on purpose
```suggestion
      {:else if smileyRating > 2}
```

feloy (approved):

axel7083:
cc @deboer-tim your review is blocking

deboer-tim (approved):

PR #8821, opened 2024-09-10
feat: add preflights check for HyperV

### LINKED ISSUE #6207: Add checks for Hyper-V in Podman extension

as part of the epic https://github.com/containers/podman-desktop/issues/5113 we should have preflight checks to see if Hyper-V is installed or not

check if hyper-v is installed or not

we should check for WSL and Hyper-V and then creating the machine should be a Hyper-V or WSL depending on the supported platform

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #6207

### What does this PR do?

Add a new preflight check for HyperV

### Screenshot / video of UI

N/A

#6207 

### How to test this PR?

Run Podman extension unit test or remove WSL and HyperV and install Podman

### DIFF
--- a/extensions/podman/packages/extension/src/podman-install.ts
+++ b/extensions/podman/packages/extension/src/podman-install.ts
@@ -23,7 +23,7 @@ import { promisify } from 'node:util';
 import * as extensionApi from '@podman-desktop/api';
 import { compare } from 'compare-versions';
 
-import { BaseCheck } from './base-check';
+import { BaseCheck, OrCheck, SequenceCheck } from './base-check';
 import { getDetectionChecks } from './detection-checks';
 import type { MachineJSON } from './extension';
 import {
@@ -457,8 +457,11 @@ export class WinInstaller extends BaseInstaller {
       new WinVersionCheck(),
       new WinMemoryCheck(),
       new VirtualMachinePlatformCheck(),
-      new WSLVersionCheck(),
-      new WSL2Check(this.extensionContext),
+      new OrCheck(
+        'Windows virtualization',
+        new SequenceCheck('WSL platform', [new WSLVersionCheck(), new WSL2Check(this.extensionContext)]),
+        new HyperVCheck(),
+      ),
     ];
   }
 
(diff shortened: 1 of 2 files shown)


benoitf (commented):
maybe introduce SequenceCheck and OrCheck in a separate PR, then use it
(and add tests for these 2 new objects)

also I'm not sure why isUserAdmin is moved to a pure function (looks like it can be in a WindowsCheck class extending BaseCheck

benoitf (approved):
code LGTM but as testing on macOS the code is not used, so probably need a review from someone on Windows

extensions/podman/src/podman-install.ts
```diff
@@ -807,3 +816,61 @@ class WSLVersionCheck extends BaseCheck {
...
+      ]);
+      return result.stdout === 'True';
+    } catch (err: unknown) {
+      return false;
+    }
+  }
+
+  private async isHyperVRunning(): Promise<boolean> {
+    try {
+      const result = await extensionApi.process.exec('powershell.exe', ['@(Get-Service vmms).Status']);
```
feloy: On my Windows machine, I get `Running` as output of this command, not `True`

feloy (commented):
on my machine, I uninstalled HyperV and WSL. 

I started Podman Desktop, the check indicates that Hyoer-V is not running.

I follow the instructions to install it, reboot, and check that it is running (`(Get-Service vmms).Status` is Running)

I restart Podman Desktop, but it still says that HyperV is not running

extensions/podman/packages/extension/src/podman-install.ts
```diff
@@ -807,3 +816,61 @@ class WSLVersionCheck extends BaseCheck {
...
+        },
+      });
+    }
+    return this.createSuccessfulResult();
+  }
+
+  private async isHyperVinstalled(): Promise<boolean> {
+    try {
+      const result = await extensionApi.process.exec('powershell.exe', [
+        '@(Get-Wmiobject Win32_ComputerSystem).HypervisorPresent',
```
feloy: I have uninstalled hyperv (from the Turn Windows features on or off), rebooted, but this command still returns True

jeffmaury: Good catch, updated

feloy:
The preflights LGTM: starting with neither wsl not hyperV installed,  the preflight makes me install Hyper-V, then pass.

But then, when I try to create a machine, I get an error that `wsl` is not found, but I would expect that it is created with Hyper-V (as only Hyper-V is installed on my system)

jeffmaury:
> The preflights LGTM: starting with neither wsl not hyperV installed, the preflight makes me install Hyper-V, then pass.
> 
> But then, when I try to create a machine, I get an error that `wsl` is not found, but I would expect that it is created with Hyper-V (as only Hyper-V is installed on my system)

This PR is only about adding the checks
Machine creation and query is another PR

feloy (approved):

Issue #8825, opened 2024-09-10
Kubernetes Welcome page


deboer-tim:
### Is your enhancement related to a problem? Please describe

As part of revamping the Kubernetes menu items (#8754) it was suggested that instead of defaulting to the Nodes or Deployments page, we could have a Kubernetes welcome page as the default page.

### Describe the solution you'd like

When a user first discovered the Kubernetes extension we want to guide and show them all the features it is capable of. In the AI Lab extension we are using a Welcome/About page to do that. We can reuse that pattern. The content that would make sense of that page is:

As a new user with, no Kubernetes resources, we should explain the main concepts. 
* Explain that Podman Desktop can be set up to connect to a Kubernetes Cluster
* Explain that Podman Desktop can create a local Kubernetes cluster for you. Using Kind, Minikube etc 
* Links to docs & blogposts:
  * https://podman-desktop.io/docs/kubernetes 
  * https://developers.redhat.com/articles/2023/06/09/deploy-and-test-kubernetes-containers-using-podman-desktop# 
  * https://developers.redhat.com/articles/2023/11/06/working-kubernetes-podman-desktop 

As a user that has some Kubernetes resources, we want to show guides where they can dig deeper and learn about how Podman works together with Kubernetes & Pods

**Requirements:**
- Clicking on the `Kubernetes` item will bring you to this Welcome/About page
- As a new user, this should be the first page you see, afterwards we should remember the previous page (#9057)

Currently the AI Lab extension has a similar pattern that could work here: 
![image](https://github.com/user-attachments/assets/1190f64a-4a6b-4711-b58f-4cdd3ab07a6b)

axel7083:
I have some doubt on this one, as I feel like user may just want to see the pods/nodes, and having this default page would make it one more click needed to access the desired page.

Some possibilities to empower the user: we could have a preference `default kubernetes page` with `Welcome` by default, and let the user choose which page would be by default.

deboer-tim:
My intention was that this is the first page in the list and shown the first time you open Kubernetes, but afterward it should remember the last page you were on. I opened up #9057 to track that separately.

slemeur:
@Firewall is working on the requirements for this. 
Moving this to an epic.

cdrage:
> I have some doubt on this one, as I feel like user may just want to see the pods/nodes, and having this default page would make it one more click needed to access the desired page.
> 
[...]

PR #8833, opened 2024-09-11
feat: change Kubernetes menu to submenu

### LINKED ISSUE #7759: Kubernetes Navigation: Replace nav sections by group nav item + individual items

As we grow the set of Kubernetes objects, there are too many to keep in the main nav, even with the current collapsable nav section. However, moving to a single item has the drawback of requiring two clicks to get to anything (feels like a second-class citizen). The current form will also conflict with hiding nav items as we plan in #6774. 

Nav sections should be removed/replaced. Each of the Kubernetes pages should be contributed as separate nav items, and the group should be contributed as a special 'group' nav item. Clicking on a group nav item opens a submenu identical to the look of the Settings or the AI Lab extension today, with a list of all the nav items in the group.

It is important to point out that both the group and each individual nav item are 'regular' nav items. In particular, they could each be added or removed from the main navigation individually as per #7758. Kubernetes would initially appear as just the one group nav item with a submenu, but here are some examples of what the user could do:
- Add Deployments and Services to the main nav, while keeping the Kubernetes group also visible (to get to all of the other items).
- Add Nodes to the main nav and hide the Kubernetes group, because they're only using nodes at the moment.
- Hide everything, because they're not using Kubernetes.

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Change Kubernetes menu to submenu

### Screenshot / video of UI

https://github.com/user-attachments/assets/f2580b81-746b-4456-9ce3-6c16e14034cd

Fixes #7759 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/App.spec.ts
+++ b/packages/renderer/src/App.spec.ts
@@ -22,11 +22,13 @@ import { router } from 'tinro';
 import { beforeEach, expect, test, vi } from 'vitest';
 
 import App from './App.svelte';
+import { navigationRegistry } from './stores/navigation/navigation-registry';
 
 const mocks = vi.hoisted(() => ({
   DashboardPage: vi.fn(),
   RunImage: vi.fn(),
   ImagesList: vi.fn(),
+  SubmenuNavigation: vi.fn(),
 }));
 
 vi.mock('./lib/dashboard/DashboardPage.svelte', () => ({
@@ -52,6 +54,10 @@ vi.mock('./lib/context/ContextKey.svelte', () => ({
 
 vi.mock('./lib/appearance/Appearance.svelte', () => ({
   default: vi.fn(),
+}));
+
+vi.mock('./SubmenuNavigation.svelte', () => ({
+  default: mocks.SubmenuNavigation,
 }));
 
 const dispatchEventMock = vi.fn();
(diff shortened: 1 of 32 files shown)


axel7083 (commented):
I feel like we should have a default view (E.g. Node while we do not have the pods) rather than an empty white page for the default.

Otherwise tested locally and LGTM

benoitf (commented):
FYI SubmenuNavigation.svelte has no unit tests

about the default view, yes usually we're defaulting the first click to a given page

(like if you open the container's details you land on a given tab)

I'm fine to have a follow-up PR to handle this default landing page as it seems it might be the pod view and with this PR there is no pod view (as expected for the issue)

feloy:
> FYI SubmenuNavigation.svelte has no unit tests
> 
Added, thanks

> about the default view, yes usually we're defaulting the first click to a given page
> 
> (like if you open the container's details you land on a given tab)
> 
> I'm fine to have a follow-up PR to handle this default landing page as it seems it might be the pod view and with this PR there is no pod view (as expected for the issue)

Or a welcome page (#8825)

benoitf (approved):

axel7083 (approved):
LGTM :rocket:

PR #8837, opened 2024-09-11
feat(extension-api): allow extension to register custom navigation route

axel7083:
### What does this PR do?

Following a discussion we had with @benoitf about https://github.com/containers/podman-desktop/issues/7709. The idea would be to make the extension register a unique navigation route id.

When they create a notification or a task, they could register the navigation route id with some arguments, so the core could navigate to the appropriate route.

### Screenshot / video of UI

N/A

Required for https://github.com/containers/podman-desktop/issues/7709

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4687,16 +4687,38 @@ declare module '@podman-desktop/api' {
     export function navigateToEditProviderContainerConnection(connection: ProviderContainerConnection): Promise<void>;
 
     /**
-     * Allow to define custom route
-     * @param routeId
-     * @param commandId
+     * Allow to define custom route for an extension.
+     *
+     * @remarks
+     * You commandId used must have been registered through {@link commands.registerCommand}
+     *
+     * @example
+     * ```ts
+     * import { navigation, commands } from '@podman-desktop/api';
+     *
+     * commands.registerCommand('redirect-download-command', (trackingId: string) => {
+     *   // todo: do something with the trackingId
+     * });
+     *
+     * // register the route
+     * navigation.register('download-page', 'redirect-download-command');
+     *
+     * // when needed call the navigate with the route id registered to
+     * // trigger the command
+     * navigation.navigate('download-page', 'dummy-tracking-id');
+     * ```
+     *
+     * @param routeId a unique string value that could be used in {@link navigation.navigate}
+     * @param commandId the command that will be executed on navigate
      */
     export function register(routeId: string, commandId: string): Disposable;
 
     /**
-     * Allow extension to navigate to a custom route
-     * @param routeId
-     * @param args
+     * Allow extension to navigate to a custom route.
+     * The route need to have been registered using {@link navigation.register}
+     *
+     * @param routeId the identifier of the route to use
+     * @param args the arguments to provide to the command linked to the routeId
      */
     export function navigate(routeId: string, ...args: unknown[]): Promise<void>;
   }
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4690,7 +4690,7 @@ declare module '@podman-desktop/api' {
      * Allow to define custom route for an extension.
      *
      * @remarks
-     * You commandId used must have been registered through {@link commands.registerCommand}
+     * The commandId used must have been registered through {@link commands.registerCommand}
      *
      * @example
      * ```ts
@@ -4715,7 +4715,7 @@ declare module '@podman-desktop/api' {
 
     /**
      * Allow extension to navigate to a custom route.
-     * The route need to have been registered using {@link navigation.register}
+     * The route needs to have been registered using {@link navigation.register}
      *
      * @param routeId the identifier of the route to use
      * @param args the arguments to provide to the command linked to the routeId
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4685,6 +4685,20 @@ declare module '@podman-desktop/api' {
      * Navigate to the Edit Provider Container Connection page
      */
     export function navigateToEditProviderContainerConnection(connection: ProviderContainerConnection): Promise<void>;
+
+    /**
+     * Allow to define custom route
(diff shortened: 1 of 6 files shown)


benoitf (dismissed):
👍

benoitf (commented):
I think routeid should be prefixed by the extension ID

so we're sure that there is no collapsing

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1402,6 +1402,19 @@ export class ExtensionLoader {
       ): Promise<void> => {
         await this.navigationManager.navigateToEditProviderContainerConnection(connection);
       },
+      navigate: async (routeId: string, ...args: unknown[]): Promise<void> => {
+        return this.navigationManager.navigateToRoute(routeId, args);
```
benoitf: here

packages/main/src/plugin/extension-loader.ts
```diff
@@ -1402,6 +1402,19 @@ export class ExtensionLoader {
...
+      },
+      register: (routeId: string, commandId: string): Disposable => {
+        const disposable = this.navigationManager.registerRoute({
```
benoitf, line 1427: and here, add the extensionID for register/navigate

packages/main/src/plugin/extension-loader.ts
axel7083: If we put here the prefix, then it mean that extension A cannot use navigation route of extension B

benoitf: yes

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4549,6 +4549,20 @@ declare module '@podman-desktop/api' {
...
+     * @param commandId
+     */
+    export function register(routeId: string, commandId: string): Disposable;
+
+    /**
+     * Allow extension to navigate to a custom route
+     * @param routeId
+     * @param args
+     */
+    export function navigate(routeId: string, ...args: unknown[]): Promise<void>;
```
benoitf: still wondering if we need simple type `(routeId, commandId)` or a complex type `({id, command})`

and the same for the navigate method
```ts
{
id: 
args?:
}
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4549,6 +4549,20 @@ declare module '@podman-desktop/api' {
      * Navigate to the Edit Provider Container Connection page
      */
     export function navigateToEditProviderContainerConnection(connection: ProviderContainerConnection): Promise<void>;
+
+    /**
+     * Allow to define custom route
+     * @param routeId
+     * @param commandId
+     */
+    export function register(routeId: string, commandId: string): Disposable;
```
feloy, line 4714: could you document where the values of the parameters come from? As they are (free-form) string, it is difficult to know from where we need to get these values.

packages/extension-api/src/extension-api.d.ts
axel7083: I would keep it like for consistency with the others functions in the navigation namespace

packages/extension-api/src/extension-api.d.ts
axel7083: Yeah np

benoitf (approved):

packages/main/src/plugin/navigation/navigation-manager.ts
```diff
@@ -19,27 +19,64 @@
...
+  ) {
+    this.#registry = new Map();
+  }
 
   navigateTo<T extends NavigationPage>(navigateRequest: NavigationRequest<T>): void {
     this.apiSender.send('navigate', navigateRequest);
   }
 
+  registerRoute(route: NavigationRoute): Disposable {
+    this.#registry.set(route.routeId, route);
```
benoitf, line 59: shouldn't we check that the route does not exist before registering it ?

packages/main/src/plugin/navigation/navigation-manager.ts
```diff
@@ -19,27 +19,64 @@
...
+
+    return Disposable.create(() => {
+      this.#registry.delete(route.routeId);
+    });
+  }
+
+  hasRoute(routeId: string): boolean {
+    return this.#registry.has(routeId);
+  }
```
benoitf, line 68: when is it called ?

packages/main/src/plugin/navigation/navigation-manager.ts
axel7083: Yes ! Updated.

packages/main/src/plugin/navigation/navigation-manager.ts
axel7083: By the new commit I just made :)

benoitf (approved):

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4685,6 +4685,42 @@ declare module '@podman-desktop/api' {
      * Navigate to the Edit Provider Container Connection page
      */
     export function navigateToEditProviderContainerConnection(connection: ProviderContainerConnection): Promise<void>;
+
+    /**
+     * Allow to define custom route for an extension.
+     *
+     * @remarks
+     * You commandId used must have been registered through {@link commands.registerCommand}
```
feloy:
```suggestion
     * The commandId used must have been registered through {@link commands.registerCommand}
```

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4685,6 +4685,42 @@ declare module '@podman-desktop/api' {
...
+     * ```
+     *
+     * @param routeId a unique string value that could be used in {@link navigation.navigate}
+     * @param commandId the command that will be executed on navigate
+     */
+    export function register(routeId: string, commandId: string): Disposable;
+
+    /**
+     * Allow extension to navigate to a custom route.
+     * The route need to have been registered using {@link navigation.register}
```
feloy:
```suggestion
     * The route needs to have been registered using {@link navigation.register}
```

feloy (approved):
just a couple of typos. 
Thanks for adding the doc
LGTM

PR #8865, opened 2024-09-12
feat: more info about sizes displayed in images layers

### LINKED ISSUE #43: UI about a layer is not obvious
Using fedora:40

I can see 3 sizes but it's not clear what is the difference. Maybe a tooltip can help

feloy:

### What does this PR do?

Display more info about sizes displayed in images layers

### Screenshot / video of UI

layer-siezs-help

Fixes https://github.com/containers/podman-desktop-extension-layers-explorer/issues/43

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/ImageDetailsFilesLayers.ts
+++ b/packages/renderer/src/lib/image/ImageDetailsFilesLayers.ts
@@ -1,0 +1,28 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { ImageUtils } from './image-utils';
+
+export function signedHumanSize(n: number): string {
+  const value = new ImageUtils().getHumanSize(n);
+  if (n < 0) {
+    return value;
+  } else {
+    return '+' + value;
+  }
+}
--- a/packages/renderer/src/lib/image/ImageDetailsFilesLayers.ts
+++ b/packages/renderer/src/lib/image/ImageDetailsFilesLayers.ts
@@ -23,6 +23,6 @@ export function signedHumanSize(n: number): string {
   if (n < 0) {
     return value;
   } else {
-    return '+' + value;
+    return `+${value}`;
   }
 }
--- a/packages/renderer/src/lib/image/ImageDetailsFilesLayers.spec.ts
+++ b/packages/renderer/src/lib/image/ImageDetailsFilesLayers.spec.ts
@@ -1,0 +1,67 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+
+import type { ImageFile } from '@podman-desktop/api';
+import { render, screen, within } from '@testing-library/svelte';
+import { expect, test } from 'vitest';
+
+import type { FilesystemTree } from './filesystem-tree';
+import type { ImageFilesystemLayerUI } from './imageDetailsFiles';
+import { signedHumanSize } from './ImageDetailsFilesLayers';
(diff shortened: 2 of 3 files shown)


axel7083:
I like the improvement! :rocket:

| Before | After |
| --- | --- |
| image | image |

However, I am not sure to completely understand what is `at rest`.

I really like the `+ <size>` it makes me think about the GitHub file diff, and why not introduce some colours ?

feloy:
> I like the improvement! 🚀
> 
> Before	After
> image	image
> However, I am not sure to completely understand what is `at rest`.
> 

I wanted to indicate that it is the size in the image (on the disk), not in a container started from this image. I see the term 'at rest' generally used for data stored on disk, and not in network traffic, but I'm not sure it could be used here. 

I cannot find a concise way to indicate that the first one is the size 'on disk' and the second and third ones are in the filesystem of the container. Perhaps some icon would help?
  @ekidneyrh would you have some advice?

> I really like the `+ <size>` it makes me think about the GitHub file diff, and why not introduce some colours ?

jeffmaury (approved):
LGTM

ekidneyrh:
maybe a disk icon would work to make it clear it means it's the size of the image stored physically?

feloy:
I have updated the labels. WDYT?

sizes-on-disk-fs

(I feel that having only one icon for the first one would not be homogeneous, adding 3 icons would take too much place and completely replacing the text by icons would not help to understand)

benoitf (changes requested):
no unit tests

packages/renderer/src/lib/image/ImageDetailsFilesLayers.ts
```diff
@@ -0,0 +1,28 @@
...
+ ***********************************************************************/
+
+import { ImageUtils } from './image-utils';
+
+export function signedHumanSize(n: number): string {
+  const value = new ImageUtils().getHumanSize(n);
+  if (n < 0) {
+    return value;
+  } else {
+    return '+' + value;
```
benoitf:
```suggestion
    return `+${value}`;
```

benoitf:
thanks for the tests,

Trying the feature since last week, nice improvement but I think the way it's displayed is confusing for the end user as we see 3 values in a row and it's really hard to know what is going on.

Maybe we should extract the size of the layer and put it on a different line as we're displaying layers, it's the kind of the primary information. (How can I reduce the size of my layers)

Then maybe for contribute to FS and total FS, it could be just the delta, like how much it is adding from the previous layer
`delta: +59MB` (and a hover where you can have a tooltip description)

here is my point, why do we have both contribute to FS and total FS while we don't see the total for the layers size ?

because on one hand, we have layers size (for the image size on the repository/to be downloaded) and on the other hand the size once we unpack the layers.
so why 3 numbers, as probably it's either 2 or 4 (either only the chunk size or both chunk and total) 

Then disk I find it confusing myself, as disk can be the space it's using on the filesystem as well

maybe
- an icon with layers like: 182MB (image size : 182MB)
- an icon with folder icon: +12MB (disk space: 182MB)

but also can be at different place on the layer screen, maybe split between image and filesystem

I think then the layer size should be the first information, not the instruction (as if i go to files it's to know the size, if I just want the instructions I have that in the history tab)
```
Layers:
- - - - - - - - - - - - - - - - - - - - - - - - 
Layer icon: 182MB (image icon with tooltip: 182MB)
/instruction being executed in Dockerfile
Filesystem icon with tooltip: +182MB (total: 182MB)
- - - - - - - - - - - - - - - - - - - - - - - - 

Layer Icon: 107.66MB (Image icon with tooltip: 290.13MB)
/instruction being executed in Dockerfile
Filesystem icon with tooltip: +59MB (total: 241.9MB)
```

side note: is that blobs/sha256 is really important that it's displayed on all the layers ? I would drop it ?

of course, can be follow ups.

I would love to see also a pie chart to see what layer is taking the most space and if I hover I see the instructions/details

benoitf (approved):

feloy:
> thanks for the tests,
> 
> Trying the feature since last week, nice improvement but I think the way it's displayed is confusing for the end user as we see 3 values in a row and it's really hard to know what is going on.
> 
> Maybe we should extract the size of the layer and put it on a different line as we're displaying layers, it's the kind of the primary information. (How can I reduce the size of my layers)
> 
> Then maybe for contribute to FS and total FS, it could be just the delta, like how much it is adding from the previous layer `delta: +59MB` (and a hover where you can have a tooltip description)
> 
> here is my point, why do we have both contribute to FS and total FS while we don't see the total for the layers size ?
> 
> because on one hand, we have layers size (for the image size on the repository/to be downloaded) and on the other hand the size once we unpack the layers. so why 3 numbers, as probably it's either 2 or 4 (either only the chunk size or both chunk and total)
> 
> Then disk I find it confusing myself, as disk can be the space it's using on the filesystem as well
> 
> maybe
> 
> * an icon with layers like: 182MB (image size : 182MB)
> * an icon with folder icon: +12MB (disk space: 182MB)
> 
> but also can be at different place on the layer screen, maybe split between image and filesystem
> 
> I think then the layer size should be the first information, not the instruction (as if i go to files it's to know the size, if I just want the instructions I have that in the history tab)
> 
> ```
> Layers:
> - - - - - - - - - - - - - - - - - - - - - - - - 
> Layer icon: 182MB (image icon with tooltip: 182MB)
> /instruction being executed in Dockerfile
> Filesystem icon with tooltip: +182MB (total: 182MB)
> - - - - - - - - - - - - - - - - - - - - - - - - 
> 
> Layer Icon: 107.66MB (Image icon with tooltip: 290.13MB)
> /instruction being executed in Dockerfile
> Filesystem icon with tooltip: +59MB (total: 241.9MB)
> ```
> 
> side note: is that blobs/sha256 is really important that it's displayed on all the layers ? I would drop it ?
> 
> of course, can be follow ups.
> 
> I would love to see also a pie chart to see what layer is taking the most space and if I hover I see the instructions/details

Great feedback, thanks.

I was thinking that the information was probably not the good one if it wasn't easy to explain it clearly... 

After reading your feedback, I tend to think that we could have only the layer size on the Layer information (on the left), and have the filesystem contributed size on the right part (where the filesystem is displayed) in some way (it's actually the total of the sizes of the directories listed, so we could for example add the `/` directory, displaying as its size the total contributed size).

feloy:
Merging as this, I'll create a follow-up PR taking into account Florent's feedback

PR #8874, opened 2024-09-13
website: update website hero and description

slemeur:
### What does this PR do?

Update the website hero and description

[](https://github.com/containers/podman-desktop/issues/8873)

### How to test this PR?

### DIFF
--- a/website/src/pages/index.tsx
+++ b/website/src/pages/index.tsx
@@ -109,8 +109,8 @@ function Hero(): JSX.Element {
               Best Free & Open Source Tool for Containers & Kubernetes - Podman Desktop
             </h1>
             <p className="text-base md:text-lg">
-              Podman Desktop is the best free and open source tool to work with Containers and Kubernetes for developers. 
-              Get an intuitive and user-friendly interface to effortlessly build, manage, and deploy containers and Kubernetes — all from your desktop. 
+              Podman Desktop is the best free and open source tool to work with Containers and Kubernetes for developers.
+              Get an intuitive and user-friendly interface to effortlessly build, manage, and deploy containers and Kubernetes — all from your desktop.
               Join the growing community of developers using Podman Desktop!
             </p>
             <div className="flex-none">
--- a/website/src/pages/index.tsx
+++ b/website/src/pages/index.tsx
@@ -106,11 +106,12 @@ function Hero(): JSX.Element {
         <div className="text-center lg:w-2/3 w-full bg-hero-pattern bg-no-repeat bg-center">
           <div className="bg-white/30 dark:bg-transparent">
             <h1 className="title-font sm:text-4xl text-3xl lg:text-6xl mb-8 font-medium text-gray-900 dark:text-white leading-[1.2]">
-              Containers and Kubernetes for application developers
+              Best Free & Open Source Tool for Containers & Kubernetes - Podman Desktop
             </h1>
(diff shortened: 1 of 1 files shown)


website/src/pages/index.tsx
```diff
@@ -106,11 +106,12 @@ function Hero(): JSX.Element {
         <div className="text-center lg:w-2/3 w-full bg-hero-pattern bg-no-repeat bg-center">
           <div className="bg-white/30 dark:bg-transparent">
             <h1 className="title-font sm:text-4xl text-3xl lg:text-6xl mb-8 font-medium text-gray-900 dark:text-white leading-[1.2]">
-              Containers and Kubernetes for application developers
+              Best Free & Open Source Tool for Containers & Kubernetes - Podman Desktop
```
feloy: I suppose you need to escape these characters
```suggestion
              Best Free &amp; Open Source Tool for Containers &amp; Kubernetes - Podman Desktop
```

website/src/pages/index.tsx
```diff
@@ -106,11 +106,12 @@ function Hero(): JSX.Element {
...
             </h1>
             <p className="text-base md:text-lg">
-              Podman Desktop is an open source graphical tool enabling you to seamlessly work with containers and
-              Kubernetes from your local environment.
+              Podman Desktop is the best free and open source tool to work with Containers and Kubernetes for developers.
+              Get an intuitive and user-friendly interface to effortlessly build, manage, and deploy containers and Kubernetes — all from your desktop.
```
feloy: the `and Kubernetes` is not clear. It applies well with _manage_, but not with _deploy_

axel7083:
What would you think about re-using some styling I made for the AI Lab page

## Current

Screenshot from 2024-09-13 10-23-56

## Gradient color

image

The colors are configurable. I also feel like the `- Podman Desktop` could be removed at the end, at it is already in the name of the website

slemeur:
Go for it ! Good idea with the gradient!

website/src/pages/index.tsx
slemeur: What do you think about:
"Get an intuitive and user-friendly interface to effortlessly bridge the gap between local development and Kubernetes clusters. Build, manage, and deploy with ease from a single tool."

benoitf (commented):
I would not mix styling and text content at the same time

axel7083:
> I would not mix styling and text content at the same time

You don't like the gradient ?

website/src/pages/index.tsx
feloy: What about finishing with:
Build, manage, and deploy everywhere with ease from a single tool

benoitf:
during a review, if you mix things, then you need a higher consensus:
both message and style would need to be approved while maybe all people may agree only on one topic


and on a side note, yes I dislike gradient title. To me it's part of the past.

axel7083:
> during a review, if you mix things, then you need a higher consensus: both message and style would need to be approved while maybe all people may agree only on one topic
> 
> and on a side note, yes I dislike gradient title. To me it's part of the past.

OH yeah sorry I forgot, the styling will not be included in the PR, I said stevan that it could be in a follow up issue where the discussion could occur

feloy (approved):

benoitf:
on my end the rendering is the following:
image

the - is at the end of the sentence and I'm not sure the text should be that long.

PR #8877, opened 2024-09-13
chore(test): expand extension's details page pom and make extensions …

### LINKED ISSUE #8876: Update E2E test Extension's details page POM for handling errors

Extension's details page POM can be updated to offer representation of the error states. Also we might extend the extensions tests to be part of smoke testing suite since the offer an overview of extension's area functionality. 

Add another locator of the page to offer error handling in e2e test (especially extensions). Add smoke tag to `extension-installation.spec.ts` and include test verifying possible extension error states.

_No response_

### Additional context

_No response_

odockal:
…e2e test smoke

### What does this PR do?
Adds some pom object into extension's details page, extends test for error states and includes extension-installation into smoke testing suite.

#8876 

### How to test this PR?
`pnpm test:e2e:smoke` -> should run extensions e2e tests as well.

### DIFF
--- a/tests/playwright/src/model/pages/extension-details-page.ts
+++ b/tests/playwright/src/model/pages/extension-details-page.ts
@@ -23,22 +23,32 @@ import { BasePage } from './base-page';
 import { ExtensionsPage } from './extensions-page';
 
 export class ExtensionDetailsPage extends BasePage {
+  readonly header: Locator;
+  readonly tabs: Locator;
+  readonly tabContent: Locator;
+  readonly buttonGroup: Locator;
   readonly enableButton: Locator;
   readonly disableButton: Locator;
   readonly removeExtensionButton: Locator;
   readonly status: Locator;
   readonly heading: Locator;
+  readonly errorStackTrace: Locator;
 
   constructor(
     page: Page,
     public readonly extensionName: string,
   ) {
     super(page);
-    this.heading = page.getByRole('heading', { name: extensionName });
-    this.enableButton = page.getByRole('button', { name: 'Start' });
-    this.disableButton = page.getByRole('button', { name: 'Stop' });
-    this.removeExtensionButton = page.getByRole('button', { name: 'Delete' });
-    this.status = page.getByLabel('Extension Status Label');
+    this.header = page.getByRole('region', { name: 'Header' });
+    this.tabs = page.getByRole('region', { name: 'Tabs' });
+    this.tabContent = page.getByRole('region', { name: 'Tab Content' });
+    this.heading = this.header.getByRole('heading', { name: extensionName });
+    this.buttonGroup = this.header.getByRole('group', { name: 'Control Actions' });
+    this.enableButton = this.buttonGroup.getByRole('button', { name: 'Start' });
+    this.disableButton = this.buttonGroup.getByRole('button', { name: 'Stop' });
+    this.removeExtensionButton = this.buttonGroup.getByRole('button', { name: 'Delete' });
+    this.status = this.header.getByLabel('Extension Status Label');
+    this.errorStackTrace = this.tabContent.getByRole('group', { name: 'Stack Trace', exact: true });
   }
 
   async disableExtension(): Promise<this> {
@@ -61,5 +71,12 @@ export class ExtensionDetailsPage extends BasePage {
     await this.disableExtension();
     await this.removeExtensionButton.click();
     return new ExtensionsPage(this.page);
+  }
+
+  async activateTab(tabName: string): Promise<this> {
+    const tabItem = this.tabs.getByRole('button', { name: tabName, exact: true });
+    await playExpect(tabItem).toBeVisible();
+    await tabItem.click();
+    return this;
   }
 }
--- a/tests/playwright/src/model/pages/extension-details-page.ts
+++ b/tests/playwright/src/model/pages/extension-details-page.ts
@@ -26,7 +26,6 @@ export class ExtensionDetailsPage extends BasePage {
   readonly header: Locator;
   readonly tabs: Locator;
   readonly tabContent: Locator;
-  readonly buttonGroup: Locator;
   readonly enableButton: Locator;
   readonly disableButton: Locator;
   readonly removeExtensionButton: Locator;
@@ -43,10 +42,9 @@ export class ExtensionDetailsPage extends BasePage {
     this.tabs = page.getByRole('region', { name: 'Tabs' });
     this.tabContent = page.getByRole('region', { name: 'Tab Content' });
     this.heading = this.header.getByRole('heading', { name: extensionName });
-    this.buttonGroup = this.header.getByRole('group', { name: 'Control Actions' });
-    this.enableButton = this.buttonGroup.getByRole('button', { name: 'Start' });
-    this.disableButton = this.buttonGroup.getByRole('button', { name: 'Stop' });
-    this.removeExtensionButton = this.buttonGroup.getByRole('button', { name: 'Delete' });
+    this.enableButton = this.header.getByRole('button', { name: 'Start' });
+    this.disableButton = this.header.getByRole('button', { name: 'Stop' });
+    this.removeExtensionButton = this.header.getByRole('button', { name: 'Delete' });
     this.status = this.header.getByLabel('Extension Status Label');
     this.errorStackTrace = this.tabContent.getByRole('group', { name: 'Stack Trace', exact: true });
   }
--- a/tests/playwright/src/specs/extension-installation-smoke.spec.ts
+++ b/tests/playwright/src/specs/extension-installation-smoke.spec.ts
@@ -27,6 +27,7 @@ import { SettingsBar } from '../model/pages/settings-bar';
 import { WelcomePage } from '../model/pages/welcome-page';
 import { NavigationBar } from '../model/workbench/navigation';
 import { Runner } from '../runner/podman-desktop-runner';
+import { waitUntil } from '../utility/wait';
 
 const DISABLED = 'DISABLED';
 const ACTIVE = 'ACTIVE';
@@ -76,7 +77,7 @@ const extentionTypes = [
 ];
 
 for (const { extensionName, extensionType } of extentionTypes) {
-  test.describe.serial(`Extension installation for ${extensionType}`, () => {
+  test.describe.serial(`Extension installation for ${extensionType} @smoke`, () => {
     test.beforeAll(async () => {
       await _startup(extensionName);
     });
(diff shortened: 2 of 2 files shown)


tests/playwright/src/specs/extension-installation-smoke.spec.ts
```diff
@@ -140,6 +141,26 @@ for (const { extensionName, extensionType } of extentionTypes) {
...
+        await waitUntil(
+          async () => {
+            if ((await errorTab.count()) > 0) {
+              stackTrace = await errorTab.innerText();
+              return true;
+            }
+            return false;
+          },
+          { timeout: 3500, sendError: false },
+        );
```
cbr7: I don't really understand why this is needed, can't we just check for the error tab count when we open the page and that's it? Why is this logic needed? Is there some sort of innate delay that makes the error tabs show up later than the page opening? And if so how do you know 3.5 seconds is the correct amount of time?

tests/playwright/src/model/pages/extension-details-page.ts
```diff
@@ -23,22 +23,32 @@ import { BasePage } from './base-page';
 import { ExtensionsPage } from './extensions-page';
 
 export class ExtensionDetailsPage extends BasePage {
+  readonly header: Locator;
+  readonly tabs: Locator;
+  readonly tabContent: Locator;
+  readonly buttonGroup: Locator;
```
cbr7: I don't really see why these are needed, it looks like needless abstraction and complexity where none if required, what problem do these container objects are meant to solve?

tests/playwright/src/specs/extension-installation-smoke.spec.ts
odockal: This is for the stacktrace message. Check for errorTab in expect happens afterwards. but if there will be error, we wont know what it was, just locator expect assertion error. but no info about actual error. Nothing than this.

tests/playwright/src/model/pages/extension-details-page.ts
odockal: This is the representation of the dom structure on the page.

tests/playwright/src/specs/extension-installation-smoke.spec.ts
cbr7: No, that part I understand, what I don't get is why does it need to be in an waitUntil() method? what is wrong with the following code:

```suggestion
            if ((await errorTab.count()) > 0) {
              stackTrace = await errorTab.innerText();
            }
```

tests/playwright/src/model/pages/extension-details-page.ts
cbr7: I know that, but they are superfluous to the goal, which is to handle the webelement, there is no benefit in getting the buttonGroup before getting the actual buttons for example, it adds an extra abstraction layer that has no benefit but increases complexity.

Just because a container is displayed in the DOM doesn't mean it needs to be present in the POM model, POM models should be limited to only what is useful, it should not have superfluous elements added to it.

tests/playwright/src/specs/extension-installation-smoke.spec.ts
odockal: Count does not wait, it evaluates immediately, right? So it seems safer, we do multiple call of the count this way.

cbr7: Which was exactly my question, why do we need to make multiple calls to count() ? Is there some practical reason this is required? And if so how do you know that 3.5 seconds worth of count() calls is the correct amount? Why not 1 second? Why not 5 seconds?

odockal: why not 3.5? seems like compromise between 1 and 5. Not long, not short. :)
Waits are good in UI testing. There is a lot situations where test might be faster than UI which is being rendered. I am claiming this is good practice. Question is whether we get into state where error tab is not yet present when we reach this code. I will remove the wait and we see if we got empty stackTrace in the message if there will be error.

cbr7: Because 3.5 is too long is a wait isn't actually required and might be too short if it is. But more to the point my impression is that the page is generated with the error tabs when you enter the extension details page so I don't see why any wait would be required **as long as the navigation to the extension page is finished** which I'm guessing could be ensured with `await playExpect(relevant locator).toBeVisible()` where the navigation is being done.

I don't see any navigation in the code snippet that is being changed so I assume the navigation is being done in some other place.

tests/playwright/src/model/pages/extension-details-page.ts
odockal: These containers have their reason, it is aria functionality, why not to include them if we have them in the structure? I understand that it is adding more complexity in this case, but if someone decides to remove the reasonable grouping of the actions on the page, we would know it, since tests will start to fail. Additionally, any changes to the page (like additional actions added for any reason) wont break anything. Also one does not really goes into class and study the particular containers and  their alignment, he uses locators he searches for. 
So I still think this has its value.

cbr7: The containers have their reason from a development perspective, but not from a testing/functional perspective, just because they are there to help the devs frame something doesn't mean that it brings value to testing the end product. The goal of e2e testing is to test functionality, not some abstract concepts.

e2e testing shouldn't care of some container element that frames the buttons changes from X to Y, the tests shouldnt break if no functionality is broken, but that is exactly what you are introducing with this abstractions, because now if the container object changes or gets renamed the test will fail, which is wrong because the functionality is the same and it works.

You are confusing module/unit tests with e2e tests and are trying to squeeze both of them at the same time, this is the wrong approach, these containers should not be in scope for e2e tests.

tests/playwright/src/specs/extension-installation-smoke.spec.ts
odockal: It is done inside `openExtensionDetails` method. So if that is correct, then it is alright. I just do not know how that works inside.

tests/playwright/src/model/pages/extension-details-page.ts
odockal: I am not confusing component tests with e2e test. we need to have robust e2e test that wont break if something new is added. It is done in images, containers pages, etc. So I wont change that concept since it is right. How is it aligned inside the hierarchy is not visible during usage. I am fine with this approach. If we wont agree, we might ask others since we have more people that are affected.

tests/playwright/src/specs/extension-installation-smoke.spec.ts
cbr7: in that case I would say that adding `await playExpect(extensionPage.heading).toBeVisible();` on line 146 of the spec file is good practice, this ensure the navigation is finished.

odockal: in doing.

tests/playwright/src/model/pages/extension-details-page.ts
cbr7: > we need to have robust e2e test that wont break if something new is added

But this approach is exactly the opposite of this goal, having robust tests is good and this manner of abstraction leads to more flakyness because we end up with e2e tests that fail because somebody changed the name of some container that holds some buttons somewhere from X to Y, while the buttons are operating completely fine.

>I am fine with this approach

But why are you fine with it when it adds complexity and flakyness and brings no benefit to testing functionality of the app?

tests/playwright/src/specs/extension-installation-smoke.spec.ts
```diff
@@ -134,10 +134,19 @@ for (const { extensionName, extensionType } of extentionTypes) {
         await playExpect(extensionDetailsPage.status).toBeVisible({ timeout: 15000 });
       });
 
-      test('Extension is active', async () => {
+      test('Extension is active and there are not errors', async () => {
         const extensionsPage = await navigationBar.openExtensions();
         const extensionPage = await extensionsPage.openExtensionDetails(extensionName, extensionLabel, extensionType);
+        await playExpect(extensionPage.header).toBeVisible();
```
cbr7: See this is exactly what I was talking about adding complexity that is not required, this needs to be:
```suggestion
        await playExpect(extensionPage.heading).toBeVisible();
```

Because there are two fields now named `header` and `heading` nobody knows which one to use, header is just the general header, it will return true wherever  the navigation is located, because there is a header everywhere, it's `heading` that validates we are on the correct page.

` this.header = page.getByRole('region', { name: 'Header' });`
` this.heading = this.header.getByRole('heading', { name: extensionName });`

In this case... because you've added the field `header` which is common on the entire app, the check would return true all the time, regardless of the navigation state. Not only is `header` not bringing any value, in this case it actively denies it by introducing confusion and potentially bugs.

tests/playwright/src/model/pages/extension-details-page.ts
odockal: We can take a look at opposite situation, someone adds another view into page (ie. for dependency that extension now can have). With precise usage of container, tests wont be affected. They will still be testing the same functionality. Without it, it is going to fail. 
So I am fine adding complexity, since only complexity is the one during pr review. Making it harder to read, I agree. On the other hand, I disagree with flakiness. It either fails or pass. 
Tests should fail is something is changed, ideally in before the code is merged. That counts for both cases.

odockal: I am finding it much harder to think of not adding more abstraction next time to just avoid long discussion under PR review. For me, we should respect the structure as it is (any extreme is bad), and not to overthink just not to add more properties into poms. @ScrewTSW @amisskii @xbabalov any opinion would be nice! I might be thinking about it wrong.

tests/playwright/src/specs/extension-installation-smoke.spec.ts
odockal: mistake, thanks.

tests/playwright/src/model/pages/extension-details-page.ts
cbr7: >Without it, it is going to fail.

No, that's simply not true, if your test fails from such a thing it means whatever locators you are using are not sufficient and/or wrong, and that means when you fix it you make it better and more robust, on the other hand the way you are doing it will always be vulnerable to container name changes or simply changing the container.

On top of that yes it absolutely adds complexity to code review but it also adds complexity to test debugging and to test writing as can be seen literally in this PR.

odockal: So it is true then, test is gonna fail, and you need to fix it (locators), which is alright. If I do not add it, it might also fail. I do not see to much reasoning between these two. Both situation requires update to the pom. One will fail for not finding the locator, the other for ambiguity. 

It has zero impact on test writing, you do not use directly the parts in between, you are using the end elements.

But it is true that debugging of more elements in the failed locator assertion is harder. Alright then, will remove it.

cbr7 (approved):

PR #8883, opened 2024-09-13
fix: fix lint errors

jeffmaury:
### What does this PR do?

Fix lint errors found in svelte update

### Screenshot / video of UI

NA

N/A

### How to test this PR?

Check workflows or run `pnpm lint:check`

### DIFF
--- a/packages/renderer/src/lib/image/FilesystemLayerView.svelte
+++ b/packages/renderer/src/lib/image/FilesystemLayerView.svelte
@@ -1,8 +1,5 @@
 <script module lang="ts">
 const expansionState = new Map<string, boolean>();
-</script>
-
-<script lang="ts">
 import type { ImageFile, ImageFileSymlink } from '@podman-desktop/api';
 
 import type { FilesystemNode } from './filesystem-tree';
--- a/packages/renderer/src/lib/image/FilesystemLayerView.svelte
+++ b/packages/renderer/src/lib/image/FilesystemLayerView.svelte
@@ -1,4 +1,4 @@
-<script module lang="ts">
+<script lang="ts">
 const expansionState = new Map<string, boolean>();
 import type { ImageFile, ImageFileSymlink } from '@podman-desktop/api';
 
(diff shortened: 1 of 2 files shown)


feloy:
I'm getting this error when runing `pnpm watch`:

```
[plugin:vite-plugin-svelte] packages/renderer/src/lib/image/FilesystemLayerView.svelte:5:0 A component can have a single top-level `<script>` element and/or a single top-level `<script context="module">` element
packages/renderer/src/lib/image/FilesystemLayerView.svelte:5:0
3 |  </script>
 4 |  
 5 |  <script lang="ts">
      ^
 6 |  import type { ImageFile, ImageFileSymlink } from '@podman-desktop/api';
 7 |
```

I understand that this is a change in svelte 5, I ran `pnpm i`, but I still get the error

jeffmaury:
> I'm getting this error when runing `pnpm watch`:
> 
> ```
> [plugin:vite-plugin-svelte] packages/renderer/src/lib/image/FilesystemLayerView.svelte:5:0 A component can have a single top-level `<script>` element and/or a single top-level `<script context="module">` element
> packages/renderer/src/lib/image/FilesystemLayerView.svelte:5:0
> 3 |  </script>
>  4 |  
>  5 |  <script lang="ts">
>       ^
>  6 |  import type { ImageFile, ImageFileSymlink } from '@podman-desktop/api';
>  7 |
> ```
> 
> I understand that this is a change in svelte 5, I ran `pnpm i`, but I still get the error

Should be fixed now

benoitf:
just curious why we need these modules anyway ?

feloy:
> just curious why we need these modules anyway ?

On `FilesystemLayerView`, `<script context="module"` is used to have a common `expansionState` variable across all instances of this component. But I realize that this introduces a problem, because, as it records only the last part of the path expanded, if the user expands`/usr/bin`, then `/usr/local/bin` and all other `bin` directories are expanded.

If I replace with `<script module`, the variable is not common anymore, so this notation does not seem equivalent.

Finally, @jeffmaury, you can completely remove `context="module"` for this component and keep only `<script lang="ts">`

feloy (approved):

PR #8884, opened 2024-09-13
fix: prevent configuration from being deleted on extension deactivate

axel7083:
### What does this PR do?

The `Disposale` returned by the `ConfigurationRegistry::registerConfigurations` is particular, as **it delete completely the configuration from the settings discarding any user change**.

The extension loader was adding it to the subscriptions, and the AutoStart was also adding it so another disposale, leading to the side effect of deleting the settings when the extension is deactivated. 

### Screenshot / video of UI

N/A

Related to https://github.com/containers/podman-desktop/issues/8435

### How to test this PR?

Replicate steps in https://github.com/containers/podman-desktop/issues/8435#issue-2461547983

### DIFF
--- a/packages/main/src/plugin/autostart-engine.spec.ts
+++ b/packages/main/src/plugin/autostart-engine.spec.ts
@@ -111,6 +111,19 @@ test('Check that default value is true if provider autostart setting is not set'
   expect(mockRegisterConfiguration).toBeCalledWith([autoStartConfigurationNode]);
 });
 
+test('Disposing the provider should not delete the configuration', async () => {
+  vi.spyOn(configurationRegistry, 'getConfiguration').mockImplementation(() => {
+    return {
+      get: (_section: string, defaultValue: boolean) => defaultValue,
+    } as Configuration;
+  });
+
+  const disposable = autostartEngine.registerProvider(extensionId, extensionDisplayName, 'internalId');
+  disposable.dispose();
+
+  expect(configurationRegistry.deregisterConfigurations).not.toHaveBeenCalled();
+});
+
 test('Check that runAutostart is called once if only one provider has registered autostart process', async () => {
   vi.spyOn(configurationRegistry, 'getConfiguration').mockImplementation(() => {
     return {
--- a/packages/main/src/plugin/autostart-engine.spec.ts
+++ b/packages/main/src/plugin/autostart-engine.spec.ts
@@ -52,35 +52,6 @@ beforeAll(() => {
   providerRegistry.runAutostart = mockRunAutostart;
 });
 
-test('Check that default value is false if provider autostart setting is false', async () => {
-  vi.spyOn(configurationRegistry, 'getConfiguration').mockImplementation(() => {
-    return {
-      get: (_section: string, _defaultValue: boolean) => false,
-    } as Configuration;
-  });
-
-  const autoStartConfigurationNode: IConfigurationNode = {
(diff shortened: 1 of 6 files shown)


benoitf (dismissed):
I think the documentation being added should be at least be an extra commit or a separate PR as it's not related to the issue being fixed

cbr7:
@axel7083 I'm not sure this PR addresses the concerns in the original issue, I just tried to reproduce the original problem with PD running from this branch and I was able to reproduce it.

I think the issue is that this PR is only addressing some edge case of the original issue, where the configuration is deleted by enabling/disabling the podman extension from PD, however the issue is that the `settings.json` parameter for the `autostart` gets removed from the settings file when the podman machine is started and stopped. For example:

Initial state: podman machine is running.

1. Open PD and go to podman extension and disable autostart podman machine.
2. Go to resources page in Settings and stop currently running podman machine -> consequence: sometimes(not all the time!!!) the autostart param is removed from the `settings.json` when the podman machine is stopped at this step.
3. In the cases where the `autorun` setting is not removed from the `settings.json` after step 2, if the user goes back to the Resources page and starts the podman machine back up in almost all cases the `autostart` param in the `settings.json` will be deleted here.

EDIT: To be clear I think this is somewhat of a systemic issue, not limited to `autorun`, I was able to observe the same removal behaviour upon starting/stopping podman machines for other parameters in the `settings.json`, for example the docker compatibility setting.

gastoner (approved):
Couldn't reproduce the bug => autostart works 
LGTM

axel7083:
> consequence: sometimes(not all the time!!!) the autostart param is removed from the `settings.json` when the podman machine is stopped at this step.

The important line in the configuration-registry is the following

https://github.com/containers/podman-desktop/blob/6b2cc63eec3f97c690753a399de081a8aecd83cb/packages/main/src/plugin/configuration-registry.ts#L343

We do not save the value in the settings if it is the default value. 

## Additional problem

A weird problem is the following, if you deactivate / reactivate the podman extension 2 times, the value will be inverted.

https://github.com/containers/podman-desktop/blob/6b2cc63eec3f97c690753a399de081a8aecd83cb/packages/main/src/plugin/autostart-engine.ts#L41-L42

This value will then be used as default, leading to inconsistency.

1. podman extension start => read `engine.autostart` with `true` as default => get `true`
2. user disable the provider auto start in the preferences
3. stop podman extension
4. start podman extension => read `engine.autostart` with `true` as default but has `false` saved => get `false`
5. set `false` as default => remove it from the settings.json

odockal:
@axel7083 I am suspecting that there might be a problem distinguishing between `podman extension` start, stop and `podman machine` start stop actions.

axel7083:
> @axel7083 I am suspecting that there might be a problem distinguishing between `podman extension` start, stop and `podman machine` start stop actions.

I mean, I understand you are talking about the machine start stop leading to the error, and as I said during the daily, this PR fixes a problem, which might not be the problem linked to the issue but could be potentially related.

deboer-tim (approved):
Tested, worked for me.

axel7083:
So this fix is not really related to https://github.com/containers/podman-desktop/issues/8435, similar but apparently do not fix it. But still fix some problem.

I had to change the e2e of podman extension, as the settings should stay, even after the activation is disabled. cc @containers/podman-desktop-qe-reviewers

cbr7 (approved):

PR #8888, opened 2024-09-13
feat: podman machine should allow image URL and registry in addition to local file path

### LINKED ISSUE #8109: Create podman machine should allow image URL not only local file path

podman machine init allows to specify a bootable image as a local file path or an URL, the implementation in Podman Desktop should allow the same

Image

Allow to specify an URL for this field and update the descriptin

_No response_

### Additional context

- [Property editor implementations](https://vscode.dev/github/dgolovin/podman-desktop/blob/main/packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.svelte#L84) 
- [Connection creation or editing form implementation](https://vscode.dev/github/dgolovin/podman-desktop/blob/main/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte#L35)

dgolovin:
### What does this PR do?

This pr introduce a way to say that an option can have different formats and in this specific fix it allows to switch between different formats for podman desktop --image-path options. It does not have full implementation with validation for URI and URL fields using just StringInput and let `podman machine init` fail if something is not right with option value. 

Specific issues should be opened for URI and URL inputs if this PR is merged.

### Screenshot / video of UI

https://github.com/user-attachments/assets/d70e46aa-ea6e-4794-b964-348f2e980486

Fix #8109.

### How to test this PR?

See screen recording above

### DIFF
--- a/extensions/podman/packages/extension/package.json
+++ b/extensions/podman/packages/extension/package.json
@@ -144,10 +144,10 @@
           "when": "podman.podmanMachineDiskSupported"
         },
         "podman.factory.machine.image-path": {
+          "markdownDescription": "Image path (optional) in a form of fully qualified registry, path, or URL to a VM image. Registry target must be in the form of _docker://registry/repo/image:version_.",
           "type": "string",
-          "format": "file",
-          "scope": "ContainerProviderConnectionFactory",
-          "description": "Image Path (Optional)"
+          "format": "file,url,fullyQualifiedRegistry",
+          "scope": "ContainerProviderConnectionFactory"
         },
         "podman.factory.machine.rootful": {
           "type": "boolean",
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
@@ -496,11 +496,6 @@ function getConnectionResourceConfigurationValue(
                   {:else if configurationKey.markdownDescription && configurationKey.type !== 'markdown'}
                     <Markdown markdown={configurationKey.markdownDescription} />
                   {/if}
-                  {#if configurationKey.format?.includes(',')}
-                    <!-- <div>
-                      very complicated element with radio buttons
-                    </div> -->
-                  {/if}
                   {#if configurationKey.format === 'memory' || configurationKey.format === 'diskSize' || configurationKey.format === 'cpu'}
                     <div class="text-gray-600">
                       <EditableConnectionResourceItem
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
@@ -496,6 +496,11 @@ function getConnectionResourceConfigurationValue(
                   {:else if configurationKey.markdownDescription && configurationKey.type !== 'markdown'}
                     <Markdown markdown={configurationKey.markdownDescription} />
                   {/if}
+                  {#if configurationKey.format?.includes(',')}
+                    <!-- <div>
+                      very complicated element with radio buttons
+                    </div> -->
+                  {/if}
                   {#if configurationKey.format === 'memory' || configurationKey.format === 'diskSize' || configurationKey.format === 'cpu'}
                     <div class="text-gray-600">
                       <EditableConnectionResourceItem
--- a/packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { getInitialValue, startCase } from '/@/lib/preferences/Util';
+import { getInitialValue } from '/@/lib/preferences/Util';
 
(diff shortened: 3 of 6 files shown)


packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
```diff
@@ -496,6 +496,11 @@ function getConnectionResourceConfigurationValue(
                   {:else if configurationKey.markdownDescription && configurationKey.type !== 'markdown'}
                     <Markdown markdown={configurationKey.markdownDescription} />
                   {/if}
+                  {#if configurationKey.format?.includes(',')}
+                    <!-- <div>
+                      very complicated element with radio buttons
+                    </div> -->
+                  {/if}
```
dgolovin:
```suggestion
```

packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { getInitialValue } from '/@/lib/preferences/Util';
+import { getInitialValue, startCase } from '/@/lib/preferences/Util';
```
dgolovin, line 2:
```suggestion
import { getInitialValue } from '/@/lib/preferences/Util';
```

jeffmaury (approved):
LGTM

packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.svelte
```diff
@@ -169,6 +171,37 @@ async function onChange(recordId: string, value: boolean | string | number): Pro
...
+            <FileItem
+              record={record}
+              value={prevSelectedFormat === selectedFormat ? (recordValue ?? '') : ''}
+              onChange={onChange} />
+          {:else}
+            <StringItem
+              record={record}
+              value={prevSelectedFormat === selectedFormat ? (recordValue ?? '') : ''}
+              onChange={onChange} />
+          {/if}
```
feloy, line 202: This will support only file and string fields. It would be nice to have some recursive component, so we don't need to repeat the logic here

packages/renderer/src/lib/preferences/Util.ts
```diff
@@ -179,3 +179,9 @@ export function calcHalfCpuCores(osCpu: string): number {
   const hCores = Math.floor(cores / 2);
   return hCores === 0 ? 1 : hCores;
 }
+
+export function startCase(str: string): string {
+  return str.replace(/([A-Z])/g, ' $1').replace(/^./, str => {
+    return str.toUpperCase();
+  });
+}
```
feloy, line 187: could you please add some unit test, to illustrate what the function is doing?

extensions/podman/packages/extension/package.json
```diff
@@ -144,10 +144,10 @@
           "when": "podman.podmanMachineDiskSupported"
         },
         "podman.factory.machine.image-path": {
+          "markdownDescription": "Image path (optional) in a form of fully qualified registry, path, or URL to a VM image. Registry target must be in the form of _docker://registry/repo/image:version_.",
           "type": "string",
-          "format": "file",
-          "scope": "ContainerProviderConnectionFactory",
-          "description": "Image Path (Optional)"
+          "format": "file,url,fullyQualifiedRegistry",
```
benoitf, line 149: AFAIK format has been designed to be a single entry not a comma separated list.

benoitf (changes requested):
request for changes as format is a single entry

extensions/podman/packages/extension/package.json
dgolovin: `--image-path` option accepts three different formats and this change is reflecting that. Other approaches would increase complexity and require more changes in codebase.

packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.svelte
dgolovin: I was thinking about recursion too, but did not want to increase number of changes in this PR.

dgolovin: There could be UriInput and UriInput components too, but I am using the fact that not known formats are rendered as TextInputs. Validation is done on extension side for now, but could be implemented for specific UI components.

extensions/podman/packages/extension/package.json
benoitf: it's definitely more complex to introduce more format options

I could use several exclusive formatter for a given type that could be worst

Why in the UI we can't have a field to specify a file path and another field to specify an URI

Then if I select both options, as we can validate parameters it will say that it is incorrect and if I select a file I have the nice browse and select window

Under the hood it can call the same parameter to give image-path

dgolovin:
It was nice 8888 number for PR. I am closing it to open new one, because it is going to be completely different.

benoitf:
you can still push a complete rewrite to this branch and keep this lucky PR number

or you wait for the next 9999

dgolovin:
it was not lucky after all ;)

extensions/podman/packages/extension/package.json
dgolovin: > Then if I select both options, as we can validate parameters it will say that it is incorrect and if I select a file I have the nice browse and select window

BTW, you still have nice browse and select window in this PR at 2:05.

PR #8891, opened 2024-09-16
feat: add podman extension api package and use it in podman extension

dgolovin:
### What does this PR do?

1. introduces podman extension API package
2. adds it to workspace
3. uses it in podman extension

Related to #5990.

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/api/package.json
+++ b/extensions/podman/packages/api/package.json
@@ -15,7 +15,7 @@
     "clean": "rimraf lib *.tsbuildinfo",
     "build": "",
     "watch": "",
-    "publish:next": "yarn publish  --registry=https://registry.npmjs.org/ --no-git-tag-version --new-version 0.0.1-\"$(date +%s)\""
+    "publish:next": "pnpm publish  --registry=https://registry.npmjs.org/ --no-git-tag-version --new-version 0.0.1-\"$(date +%s)\""
   },
   "dependencies": {
     "@podman-desktop/api": "workspace:*"
--- a/extensions/podman/packages/api/package.json
+++ b/extensions/podman/packages/api/package.json
@@ -18,6 +18,6 @@
     "publish:next": "yarn publish  --registry=https://registry.npmjs.org/ --no-git-tag-version --new-version 0.0.1-\"$(date +%s)\""
   },
   "dependencies": {
-    "@podman-desktop/api": "^1.12.0"
+    "@podman-desktop/api": "workspace:*"
   }
 }
--- a/extensions/podman/packages/api/package.json
+++ b/extensions/podman/packages/api/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "@podman-desktop/podman-extension-api",
+  "version": "0.0.1",
+  "description": "API for Podman Desktop Podman extension",
+  "publishConfig": {
+    "access": "public"
+  },
+  "license": "Apache-2.0",
+  "types": "./src/podman-extension-api.d.ts",
+  "files": [
+    "src"
+  ],
+  "scripts": {
+    "prepare": "",
+    "clean": "rimraf lib *.tsbuildinfo",
+    "build": "",
+    "watch": "",
+    "publish:next": "yarn publish  --registry=https://registry.npmjs.org/ --no-git-tag-version --new-version 0.0.1-\"$(date +%s)\""
+  },
+  "dependencies": {
+    "@podman-desktop/api": "^1.12.0"
+  }
+}
--- a/extensions/podman/packages/api/src/podman-extension-api.d.ts
+++ b/extensions/podman/packages/api/src/podman-extension-api.d.ts
@@ -17,6 +17,7 @@
  ***********************************************************************/
 
 import type { ProviderContainerConnection, RunOptions, RunResult } from '@podman-desktop/api';
+
 export interface PodmanRunOptions extends RunOptions {
   connection?: ProviderContainerConnection;
 }
--- a/extensions/podman/packages/api/src/podman-extension-api.d.ts
+++ b/extensions/podman/packages/api/src/podman-extension-api.d.ts
@@ -17,7 +17,10 @@
  ***********************************************************************/
 
 import type { ProviderContainerConnection, RunOptions, RunResult } from '@podman-desktop/api';
+export interface PodmanRunOptions extends RunOptions {
+  connection?: ProviderContainerConnection;
+}
 
 export interface PodmanExtensionApi {
-  exec(args: string[], connection?: ProviderContainerConnection, options?: RunOptions): Promise<RunResult>;
+  exec(args: string[], options?: PodmanRunOptions): Promise<RunResult>;
 }
--- a/extensions/podman/packages/api/src/podman-extension-api.d.ts
+++ b/extensions/podman/packages/api/src/podman-extension-api.d.ts
@@ -16,8 +16,8 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import type { RunOptions, RunResult } from '@podman-desktop/api';
+import type { ProviderContainerConnection, RunOptions, RunResult } from '@podman-desktop/api';
 
 export interface PodmanExtensionApi {
-  exec(args: string[], options?: RunOptions): Promise<RunResult>;
+  exec(args: string[], connection?: ProviderContainerConnection, options?: RunOptions): Promise<RunResult>;
 }
--- a/extensions/podman/packages/api/src/podman-extension-api.d.ts
+++ b/extensions/podman/packages/api/src/podman-extension-api.d.ts
@@ -1,0 +1,23 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { RunOptions, RunResult } from '@podman-desktop/api';
+
+export interface PodmanExtensionApi {
+  exec(args: string[], options?: RunOptions): Promise<RunResult>;
+}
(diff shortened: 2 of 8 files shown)


gastoner (approved):

benoitf (dismissed):
missing unit tests

extensions/podman/packages/api/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "types": "./src/podman-extension-api.d.ts",
+  "files": [
+    "src"
+  ],
+  "scripts": {
+    "prepare": "",
+    "clean": "rimraf lib *.tsbuildinfo",
+    "build": "",
+    "watch": "",
+    "publish:next": "yarn publish  --registry=https://registry.npmjs.org/ --no-git-tag-version --new-version 0.0.1-\"$(date +%s)\""
```
benoitf: FYI yarn is no longer used

benoitf: looks an old copy

extensions/podman/packages/api/vite.config.js
```diff
@@ -0,0 +1,46 @@
...
+const PACKAGE_ROOT = __dirname;
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  mode: process.env.MODE,
+  root: PACKAGE_ROOT,
+  resolve: {
+    alias: {
+      '/@/': join(PACKAGE_ROOT, 'src') + '/',
+    },
```
benoitf: I think it can be removed as we only have a declaration file

extensions/podman/packages/api/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
```
benoitf, line 3: could you please include   "extends": "@tsconfig/strictest/tsconfig.json",

so for new module, we ensure we have good strict rules (like in packages/main)

extensions/podman/packages/api/src/podman-extension-api.d.ts
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { RunOptions, RunResult } from '@podman-desktop/api';
+
+export interface PodmanExtensionApi {
+  exec(args: string[], options?: RunOptions): Promise<RunResult>;
+}
```
benoitf: here I don't see how consumers will be able to use that API in case of multiple providers (applehv, libkrun)

For example I see an hardcoded 'podman' value in the implementation:

```ts
return execPodman(args, 'podman', options);
```

but if I look at execPodman method

```ts
export function execPodman(
  args: string[],
  containersProvider?: string,
  options?: extensionApi.RunOptions,
): Promise<extensionApi.RunResult> {
````

containersProvider is an option to specify like is it Apple Hypervisor, is it libkrun, etc.

**Review by benoitf — COMMENTED** [review]
I think it might help if we're able to remove the code from other extensions using dedicated podman code

Here it seems we're assuming other extensions are using the CLI when calling 'podman stuff' but for example Ai-Lab is using the API, not the CLI to a lot of code.

https://github.com/containers/podman-desktop-extension-ai-lab/blob/0045a65681889ada6763bf05a12ad93d0bbe14b0/packages/backend/src/managers/podmanConnection.ts

so the API should provide a way to get all the podman connections as well directly


could be interesting that to have PR in other repositories to see if the API will fit, before trying to provide an API that extensions won't consume/solve the problem.

File: extensions/podman/packages/api/src/podman-extension-api.d.ts (hunk 4 above)
Comment (axel7083) [reply extensions/podman/packages/api/src/podman-extension-api.d.ts]: I think we should pass the object `ProviderContainerConnection` or `ContainerProviderConnection` instead of a string

If we pass `ProviderContainerConnection` podman extension could add a security checking that the providerId corresponds and we are not sending one registered by another extensions.

however;

`ContainerProviderConnection` would be easier as extension get this object pretty easily through, and is something used in the `BuildOptions`, `VolumeCreateOptions`, `pullImage`, `createNetwork` etc....

File: extensions/podman/packages/api/package.json (hunk 1 above)
Comment (dgolovin) [reply extensions/podman/packages/api/package.json]: updated to 'pnpm'

File: extensions/podman/packages/api/vite.config.js (hunk 2 above)
Comment (dgolovin) [reply extensions/podman/packages/api/vite.config.js]: removed.

File: extensions/podman/packages/api/tsconfig.json (hunk 3 above)
Comment (dgolovin) [reply extensions/podman/packages/api/tsconfig.json:3]: extends attribute added.

File: extensions/podman/packages/api/src/podman-extension-api.d.ts (hunk 4 above)
Comment (dgolovin) [reply extensions/podman/packages/api/src/podman-extension-api.d.ts]: I added additional optional parameter `connection: ProviderContainerConnection`, so clients can get instance using `provider.getContainerConnections` and use it to call exec.

**Comment by dgolovin** [comment]
> so the API should provide a way to get all the podman connections as well directly

You are suggesting to have getConnections() method that would be called when needed, instead of having all of those listeners added to track changes in connections related to podman provider, aren't you?

That should work, but should we add that as different pull request, after merging API with exec and implementing publishing?

I can definitely add that in this PR, but it would be out of the issue scope IMO.

**Comment by benoitf** [comment]
I just want to be sure that @axel7083 and others on AI lab extension will be able to use the extension endpoint.
You're saying it's out of the scope but it seems the scope is to be able to call this extension rather doing some private fork/own code

If there is an exec endpoint that is useless for other extensions then they won't use it and still keep their own code

**Comment by benoitf** [comment]
that said, it needs to be something in the middle where we might not have all the endpoints but we need to be sure that what we're adding is working smoothly for the other extensions

File: extensions/podman/packages/api/src/podman-extension-api.d.ts [hunk 5]
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ProviderContainerConnection, RunOptions, RunResult } from '@podman-desktop/api';
+
+export interface PodmanExtensionApi {
+  exec(args: string[], connection?: ProviderContainerConnection, options?: RunOptions): Promise<RunResult>;
+}
```
axel7083:
```suggestion
export interface PodmanRunOptions extends RunOptions {
  connection?: ProviderContainerConnection;
}

export interface PodmanExtensionApi {
  exec(args: string[], options?: PodmanRunOptions): Promise<RunResult>;
}
```

I must admit I would have a preference for having everything in the options (personal preference). cc @benoitf

axel7083:
> I just want to be sure that @axel7083 and others on AI lab extension will be able to use the extension endpoint.

To make some tests in AI Lab I added the following functions

```ts
import type { ExtensionContext, ProviderContainerConnection, RunOptions, RunResult } from '@podman-desktop/api';
import { extensions, provider } from '@podman-desktop/api';

async function podmanExtensionExperiment(): Promise<void> {
  const podman = extensions.getExtension('podman-desktop.podman');
  if(!podman) throw new Error('cannot found podman extension api');

  const podmanApi: {
    exec(args: string[], connection?: extensionApi.ProviderContainerConnection, options?: RunOptions): Promise<RunResult>;
  } = podman.exports;

  const machines = await podmanApi.exec(['machine', 'list', '--format=json']);
  console.log('podman machine list', JSON.parse(machines.stdout));

  const connections: ProviderContainerConnection[] = provider.getContainerConnections();
  for (const connection of connections) {
    await podmanApi.exec(['run','hello-world'], connection);
  }
}
```

I have two connections: `podman-machine-default` (qemu machine) and `Podman` (podman native).

After running the function above in AI Lab, I cannot see the hello-world on the qemu machine.

```bash
$: podman container ls -a
d67b196eac60  quay.io/podman/hello:latest                                                                                 /usr/local/bin/po...  3 minutes ago  Exited (0) 3 minutes ago                                                                          elastic_ganguly
1682a7a50ae1  quay.io/podman/hello:latest                                                                                 /usr/local/bin/po...  3 minutes ago  Exited (0) 3 minutes ago                                                                          pensive_saha
$: podman --connection=podman-machine-default container ls -s
# nothing
```

We seems to have both hello-world container running on the podman default

## Throobleshooting

For some weird reason the following command

```
$: CONTAINERS_MACHINE_PROVIDER=podmanm-machine-default podman container ls -a
```

return the container list of the podman native connection. The only way to get the podman from the qemu is to use the following

```
$: podman --connection=podmanm-machine-default container ls
```

extensions/podman/packages/api/src/podman-extension-api.d.ts
benoitf: I agree, a single option parameter is better, even more when it's optional

dgolovin:
> that said, it needs to be something in the middle where we might not have all the endpoints but we need to be sure that what we're adding is working smoothly for the other extensions

Looking through podmanConnection.ts and it looks like it can be just moved over to podman extension and used as an API prototype. Whatever extension needs to work with podman directly would be able to use it directly from podman extension avoiding dealing with provider API.

benoitf:
we can probably iterate there,

move as suggested all the options to a single option parameter

```
export interface PodmanRunOptions extends RunOptions {
  connection?: ProviderContainerConnection;
}
export interface PodmanExtensionApi {
  exec(args: string[], options?: PodmanRunOptions): Promise<RunResult>;
}
```

then we will be able to start publishing the artifact

then we can add the other parts.

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -1244,7 +1245,15 @@ export function registerOnboardingRemoveUnsupportedMachinesCommand(): extensionA
   });
 }
 
-export async function activate(extensionContext: extensionApi.ExtensionContext): Promise<void> {
+async function exec(
+  args: string[],
+  connection?: extensionApi.ProviderContainerConnection,
+  options?: extensionApi.RunOptions,
+): Promise<extensionApi.RunResult> {
+  return execPodman(args, connection?.connection.vmTypeDisplayName, options);
```
benoitf: AFAIK it's not the display name to use but the vmType

axel7083: If I have two wsl vmType, I need to use the connection name to be able to determine which one is supposed to be targeted

dgolovin: Both can be used, because if known display name is passed, it would be converted to vmType, otherwise if the passed value cannot be converted, it is used as as vmType (see [execPodman](https://github.com/containers/podman-desktop/blob/87c5065fefc26a7ff207e1f180b01a8fe36be32a/extensions/podman/packages/extension/src/util.ts#L148)).

axel7083: The `CONTAINERS_MACHINE_PROVIDER` allow us to target a provider (`qemu`, `wsl` etc.) but not a specific machine in the provider.

But as Florent said, let's iterate after this PR

axel7083 (approved):
LGTM

benoitf (approved):
ok for step 1

we have a short window until the API goes public with 1.13.0 release
so we can still change the API until that if we see it's not working in the extensions until that time

PR #8892, opened 2024-09-16
fix: use docker.io registry credentials for index.docker.io images

### LINKED ISSUE #8293: Unable to install an extension from a Docker private repository
### Bug description

I'm trying to install an extension from an image stored in a private Docker repository, but I get an error `Error while analyzing image: Error: Unable to get access`

### Operating system

mac

### Installation Method

### Version

next (development version)

### Steps to reproduce

I tried with the image names:
- `feloy/test:latest`
- `docker.io/feloy/test:latest`

### Relevant log output

_No response_

### Additional context

- I can install the same image from a private Quay.io repository

- Login/pwd have been added for registry `docker.io`.

- It works if I also add login/pwd for registry `index.docker.io`

feloy:
### What does this PR do?

Use docker.io registry credentials for index.docker.io images

Fixes #8293 

### How to test this PR?

Try to install an extension from a private Docker repository (you should have added credentials for the Docker registry)

### DIFF
--- a/packages/main/src/plugin/image-registry.ts
+++ b/packages/main/src/plugin/image-registry.ts
@@ -114,7 +114,10 @@ export class ImageRegistry {
   }
 
   getAuthconfigForServer(registryServer: string): Dockerode.AuthConfig | undefined {
-    const matchingUrl = registryServer;
+    let matchingUrl = registryServer;
+    if (matchingUrl === 'index.docker.io') {
+      matchingUrl = 'docker.io';
+    }
     // grab authentication data for this server
     const matchingRegistry = this.getRegistries().find(
       registry => registry.serverUrl.toLowerCase() === matchingUrl.toLowerCase(),
--- a/packages/main/src/plugin/image-registry.spec.ts
+++ b/packages/main/src/plugin/image-registry.spec.ts
@@ -867,6 +867,21 @@ test('getAuthconfigForServer returns the expected authconfig', async () => {
   expect(config?.serveraddress).toBe('my-podman-desktop-fake-registry.io');
 });
 
+test('getAuthconfigForServer returns docker.io authconfig when server is index.docker.io', async () => {
+  imageRegistry.registerRegistry({
+    serverUrl: 'docker.io',
+    username: 'foo',
+    secret: 'my-secret',
+    source: 'podman-desktop',
+  });
+  const config = imageRegistry.getAuthconfigForServer('index.docker.io');
+
+  expect(config).toBeDefined();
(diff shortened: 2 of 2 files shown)


packages/main/src/plugin/image-registry.ts
```diff
@@ -921,20 +924,24 @@ export class ImageRegistry {
...
-      options.registry = 'https://index.docker.io';
-    }
-    if (options.registry === 'docker.io') {
-      options.registry = 'index.docker.io';
-    }
-    if (!options.registry.startsWith('http')) {
-      options.registry = 'https://' + options.registry;
+    try {
+      if (!options.registry) {
+        options.registry = 'https://index.docker.io';
```
axel7083: (Future PR) Should we have this in a configuration value, like `default registry` ?

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -0,0 +1,233 @@
+<script lang="ts">
```
axel7083: Does this PR is suppose to include the Typeahead component ? Giving the title I thought it would be backend only

benoitf (commented):
for separation of concern, it seems this PR is achieving way more stuff than expected

I'm expecting only the 

```ts
 let matchingUrl = registryServer;
    if (matchingUrl === 'index.docker.io') {
      matchingUrl = 'docker.io';
    }
```
change + the unit tests

but here it's also changing the return method like throwing error instead of empty result, etc, new UI components, etc that should be in separate PRs

packages/renderer/src/lib/ui/Typeahead.svelte
feloy: I created the branch based on the Typeahead's branch, to be able to test the typeahead of private images with the fix. I'll rebase it to main.

Only the latest commit is part of this PR

feloy:
> for separation of concern, it seems this PR is achieving way more stuff than expected
> 
> I'm expecting only the
> 
> ```ts
>  let matchingUrl = registryServer;
>     if (matchingUrl === 'index.docker.io') {
>       matchingUrl = 'docker.io';
>     }
> ```
> 
> change + the unit tests
> 
> but here it's also changing the return method like throwing error instead of empty result, etc, new UI components, etc that should be in separate PRs

https://github.com/containers/podman-desktop/pull/8892#discussion_r1760945546

axel7083 (approved):
LGTM Codewise :rocket:

benoitf (approved):

PR #8902, opened 2024-09-16
feat: allowing extension to provide routeId to progress task

### LINKED ISSUE #7709: feat(Task): extensions should be able to define an action

axel7083:
### What does this PR do?

Allow extensions to provide a `navigation` object when using `window.withProgress`.

### Screenshot / video of UI

https://github.com/user-attachments/assets/40fa7bee-adc3-4c27-987c-ee351012d26b

Fixes https://github.com/containers/podman-desktop/issues/7709

### How to test this PR?

**Testing manually**

Inside the activate of an extension add the following

```ts
  extensionApi.commands.registerCommand('podman.dummy-command-id', (...args) => {
    console.log('[podman] command triggered ', args);
    return extensionApi.navigation.navigateToContainers();
  });

  extensionApi.navigation.register('dummy-route-id', 'podman.dummy-command-id');

  extensionApi.window.withProgress<string>({
    details: {
      routeId: 'dummy-route-id',
      routeArfs: ['tracking-id'],
    },
    title: 'Task example'
  }, async (): Promise<string> => {
    return new Promise<string>((resolve, reject) => {
      setTimeout(() => {
        resolve('patate');
      }, 10 * 1000);
    });
  }).catch((err: unknown) => {
    console.error(err);
  });
```

and add the following to the extension package.json

```
{
....
  "contributes": {
    "commands": [
      {
        "command": "podman.dummy-command-id",
        "title": "Podman: Dummy command (navigation related)"
      },
....
```

1. Then activate / deactivate the given extension
2. See the TaskManager

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1205,9 +1205,9 @@ declare module '@podman-desktop/api' {
      *     {
      *       location: ProgressLocation.TASK_WIDGET,
      *       title: 'My task',
-     *       navigation: {
+     *       details: {
      *         routeId: 'dummy-route-id',
-     *         arguments: ['hello', 'world'],
+     *         routeArgs: ['hello', 'world'],
      *       }
      *     },
      *     async () => {
@@ -1216,15 +1216,15 @@ declare module '@podman-desktop/api' {
      *   );
      * ```
      */
-    navigation?: {
+    details?: {
       /**
        * The routeId used in {@link navigation.register}
        */
       routeId: string;
       /**
-       * The arguments to provide the navigation route
+       * The arguments to provide the route
        */
-      arguments: string[];
+      routeArgs: string[];
     };
   }
 
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -1193,6 +1193,39 @@ declare module '@podman-desktop/api' {
      * button.
      */
     cancellable?: boolean;
+
+    /**
+     * You may specify a navigation object, making the task having a
+     * navigate action that the user can trigger.
+     * @example
+     * ```ts
+     * import { window, type ProgressLocation } from '@podman-desktop/api';
+     *
+     * await window.withProgress<string>(
+     *     {
+     *       location: ProgressLocation.TASK_WIDGET,
+     *       title: 'My task',
+     *       navigation: {
+     *         routeId: 'dummy-route-id',
+     *         arguments: ['hello', 'world'],
+     *       }
+     *     },
+     *     async () => {
+     *       return 'dummy result';
+     *     },
+     *   );
+     * ```
+     */
+    navigation?: {
+      /**
+       * The routeId used in {@link navigation.register}
+       */
+      routeId: string;
+      /**
+       * The arguments to provide the navigation route
+       */
+      arguments: string[];
+    };
   }
 
   /**
--- a/packages/main/src/plugin/tasks/progress-impl.spec.ts
+++ b/packages/main/src/plugin/tasks/progress-impl.spec.ts
@@ -21,6 +21,7 @@
 import type { Event } from '@podman-desktop/api';
 import { beforeEach, expect, test, vi } from 'vitest';
 
+import type { NavigationManager } from '/@/plugin/navigation/navigation-manager.js';
 import type { Task, TaskAction, TaskUpdateEvent } from '/@/plugin/tasks/tasks.js';
 import type { TaskState, TaskStatus } from '/@api/taskInfo.js';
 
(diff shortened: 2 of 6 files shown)


gastoner (approved):
LGTM, Works nicely!

packages/main/src/plugin/tasks/progress-impl.ts
```diff
@@ -78,14 +83,34 @@ export class ProgressImpl {
...
+  protected getTaskAction(options: extensionApi.ProgressOptions): TaskAction | undefined {
+    if (!options.navigation) return undefined;
+
+    if (!this.navigationManager.hasRoute(options.navigation.routeId)) {
+      console.warn(`cannot created task action for unknown routeId ${options.navigation.routeId}`);
+      return undefined;
+    }
+
+    return {
+      name: 'Goto task >',
```
feloy:
```suggestion
      name: 'Go to task >',
```

packages/main/src/plugin/tasks/progress-impl.spec.ts
```diff
@@ -127,3 +133,43 @@ test('Should update the task name', async () => {
...
+    async () => {
+      return 'dummy result';
+    },
+  );
+
+  await vi.waitFor(() => {
+    expect(taskAction).toBeDefined();
+  });
+
+  expect(taskAction?.name).toBe('Goto task >');
```
feloy:
```suggestion
  expect(taskAction?.name).toBe('Go to task >');
```

packages/main/src/plugin/tasks/progress-impl.ts
```diff
@@ -78,14 +83,34 @@ export class ProgressImpl {
...
+      execute: (): unknown => {
+        if (!options.navigation) return;
+        return this.navigationManager.navigateToRoute(options.navigation.routeId, ...options.navigation.arguments);
```
feloy: Could it be possible (and ok) to close the Task window at this point?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -1193,6 +1193,39 @@ declare module '@podman-desktop/api' {
...
+     *         arguments: ['hello', 'world'],
+     *       }
+     *     },
+     *     async () => {
+     *       return 'dummy result';
+     *     },
+     *   );
+     * ```
+     */
+    navigation?: {
```
benoitf: I was thinking that navigation might not be intuitive, like why should it navigate somewhere as I'm executing a task or when is it navigating.

The main goal is to provide details of the task being executed, so maybe use `details` (navigation is how it's implemented)

```ts
details?: {
   routeId: string
   routeArgs: string[]
}
```

packages/main/src/plugin/tasks/progress-impl.ts
benoitf: I would just use `Details` or `More details`

axel7083: Stevan added a comment, and have a preference for `view` it is okay for you ?

packages/extension-api/src/extension-api.d.ts
axel7083: Why not !

packages/main/src/plugin/tasks/progress-impl.ts
axel7083: We currently do not have a way to close the TaskManager from the backend, so might be a follow-up issue

packages/main/src/plugin/tasks/progress-impl.ts
```diff
@@ -78,14 +83,34 @@ export class ProgressImpl {
...
+  protected getTaskAction(options: extensionApi.ProgressOptions): TaskAction | undefined {
+    if (!options.details) return undefined;
+
+    if (!this.navigationManager.hasRoute(options.details.routeId)) {
+      console.warn(`cannot created task action for unknown routeId ${options.details.routeId}`);
+      return undefined;
+    }
+
+    return {
+      name: 'View',
```
benoitf, line 95: I still believe `View details` or `Inspect` is more accurate but as it's not part of the spec (.d.ts file), it can be  easily changed over time (or not) later.

can merge and discuss it during UX discussions

benoitf (approved):

PR #8905, opened 2024-09-17
chore(tests): added runner context to fixtures and registry tests

xbabalov:
Signed-off-by: xbabalov <t.babalova.17@gmail.com>

### What does this PR do?
Adds new context in between non-depending registry tests to prevent isolation failures.

https://github.com/containers/podman-desktop/issues/6972

### How to test this PR?
npm run test:e2e

### DIFF
--- a/tests/playwright/src/specs/registry-image.spec.ts
+++ b/tests/playwright/src/specs/registry-image.spec.ts
@@ -52,7 +52,8 @@ test.afterAll(async ({ runner, page }) => {
 });
 
 test.describe.serial('Pulling image from authenticated registry workflow verification', () => {
-  test('Cannot pull image from unauthenticated registry', async ({ page, navigationBar }) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  test('Cannot pull image from unauthenticated registry', async ({ context, page, navigationBar }) => {
     const imagesPage = await navigationBar.openImages();
 
     const fullImageTitle = imageUrl.concat(':' + imageTag);
--- a/tests/playwright/src/specs/registry.spec.ts
+++ b/tests/playwright/src/specs/registry.spec.ts
@@ -39,7 +39,8 @@ test.afterAll(async ({ runner }) => {
 });
 
 test.describe.serial('Registries handling verification', () => {
-  test('Check Registries page components and presence of default registries', async ({ navigationBar }) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  test('Check Registries page components and presence of default registries', async ({ context, navigationBar }) => {
     const settingsBar = await navigationBar.openSettings();
     const registryPage = await settingsBar.openTabPage(RegistriesPage);
 
@@ -53,7 +54,8 @@ test.describe.serial('Registries handling verification', () => {
   });
 
   test.describe.serial('Registry addition workflow verification', () => {
-    test('Cannot add invalid registry', async ({ page, navigationBar }) => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    test('Cannot add invalid registry', async ({ context, page, navigationBar }) => {
       await navigationBar.openDashboard();
       const settingsBar = await navigationBar.openSettings();
       const registryPage = await settingsBar.openTabPage(RegistriesPage);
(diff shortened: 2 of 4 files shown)


tests/playwright/src/specs/registry-image.spec.ts
```diff
@@ -52,7 +52,8 @@ test.afterAll(async ({ runner, page }) => {
 });
 
 test.describe.serial('Pulling image from authenticated registry workflow verification', () => {
-  test('Cannot pull image from unauthenticated registry', async ({ page, navigationBar }) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  test('Cannot pull image from unauthenticated registry', async ({ context, page, navigationBar }) => {
```
cbr7, line 56: context fixture is not being used in this test, why are you instancing it?

tests/playwright/src/specs/registry.spec.ts
```diff
@@ -39,7 +39,8 @@ test.afterAll(async ({ runner }) => {
 });
 
 test.describe.serial('Registries handling verification', () => {
-  test('Check Registries page components and presence of default registries', async ({ navigationBar }) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  test('Check Registries page components and presence of default registries', async ({ context, navigationBar }) => {
```
cbr7, line 43: context fixure is not being used, why are you instancing it?

tests/playwright/src/specs/registry.spec.ts
```diff
@@ -53,7 +54,8 @@ test.describe.serial('Registries handling verification', () => {
   });
 
   test.describe.serial('Registry addition workflow verification', () => {
-    test('Cannot add invalid registry', async ({ page, navigationBar }) => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    test('Cannot add invalid registry', async ({ context, page, navigationBar }) => {
```
cbr7, line 58: same as above

tests/playwright/src/utility/fixtures.ts
```diff
@@ -30,21 +30,29 @@ export type TestFixtures = {
...
   },
-  page: async ({ runner }, use) => {
-    await use(runner.getPage());
+  context: async ({ runner }, use) => {
+    newContext = true;
+    await use(runner.getNewContext());
+  },
+  page: async ({ runner, context }, use) => {
+    const page = newContext ? context.pages()[0] : runner.getPage();
+    await use(page);
```
cbr7, line 55: in this situation `newContext` can't ever be `false` it will always be `true` because as part of page fixtures it will call `context` fixture since it's a dependency of creating `page` fixture.

tests/playwright/src/utility/fixtures.ts
```diff
@@ -30,21 +30,29 @@ export type TestFixtures = {
...
   runner: async ({ runnerOptions }, use) => {
     const runner = await Runner.getInstance({ runnerOptions });
     await use(runner);
   },
-  page: async ({ runner }, use) => {
-    await use(runner.getPage());
+  context: async ({ runner }, use) => {
+    newContext = true;
+    await use(runner.getNewContext());
+  },
```
cbr7, line 52: why does this need to be a fixture, I'm not sure where you are actually using it, but it does not need to be a fixture, you can just can use the `runner` fixture and call the `context()` or call the method you created in the `podman-desktop-runner.ts` method exposed by the PW API to use whatever BroserContext is active.

tests/playwright/src/utility/fixtures.ts
```diff
@@ -30,21 +30,29 @@ export type TestFixtures = {
...
   page: Page;
+  context: BrowserContext;
   statusBar: StatusBar;
 };
 
 export type FixtureOptions = {
   runnerOptions: RunnerOptions;
 };
 
+let newContext = false;
```
cbr7, line 41: This is anti-pattern and doesn't work, since `newContext` will always be true when checked by `page` fixture, because `context` fixture will be called before the page fixture due to the dependency chain defined in this PR.

cbr7 (changes requested):
I'm not clear on what this PR is trying to achieve, the `context` fixture isn't really needed, access to the `BrowserContext` entity can be obtained from the `runner` fixture, but also see that the fixture even when defined and instance isn't used in any test, so I don't understand what purpose it serves at all.

PR #8906, opened 2024-09-17
fix: update  dependabot svelte-5.0.0-next.246 update PR

jeffmaury:
### What does this PR do?

Fix build errors

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Check workflows


benoitf:
it's not the only fix based on other errors so can we merge in main this one or it would fail as it also require a new svelte package ?

PR #8918, opened 2024-09-17
chore(test): fix tests on windows

### LINKED ISSUE #8919: Several automated e2e tests are failing on window platform
### Bug description

There are several e2e tests that are failing on windows which require fixing.

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Fixes several broken e2e tests on windows.

https://github.com/containers/podman-desktop/issues/8919

### DIFF
--- a/tests/playwright/src/specs/podman-machine-onboarding.spec.ts
+++ b/tests/playwright/src/specs/podman-machine-onboarding.spec.ts
@@ -30,6 +30,7 @@ import { expect as playExpect, test } from '../utility/fixtures';
 import { deletePodmanMachine } from '../utility/operations';
 
 const PODMAN_MACHINE_STARTUP_TIMEOUT: number = 360_000;
+const PODMAN_FULL_STARTUP_TIMEOUT = PODMAN_MACHINE_STARTUP_TIMEOUT + 30000;
 const PODMAN_MACHINE_NAME: string = 'podman-machine-default';
 const RESOURCE_NAME: string = 'podman';
 
@@ -128,7 +129,7 @@ test.describe.serial('Podman Machine verification', () => {
     test.skip(process.env.TEST_PODMAN_MACHINE !== 'true');
 
     test('Create a default Podman machine', async () => {
-      test.setTimeout(PODMAN_MACHINE_STARTUP_TIMEOUT + 30000);
+      test.setTimeout(PODMAN_FULL_STARTUP_TIMEOUT);
       await podmanOnboardingPage.podmanMachineCreateButton.click();
       await playExpect(podmanOnboardingPage.podmanMachineShowLogsButton).toBeVisible();
       await podmanOnboardingPage.podmanMachineShowLogsButton.click();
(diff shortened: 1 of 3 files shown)


tests/playwright/src/specs/podman-machine-onboarding.spec.ts
```diff
@@ -128,6 +128,7 @@ test.describe.serial('Podman Machine verification', () => {
     test.skip(process.env.TEST_PODMAN_MACHINE !== 'true');
 
     test('Create a default Podman machine', async () => {
+      test.setTimeout(PODMAN_MACHINE_STARTUP_TIMEOUT + 30000);
```
benoitf: shouldn't the timeout be a constant itself ? like why do we add 30s to an existing constant ?

I would expect something to be always like `test.setTimeout(PODMAN_MACHINE_SOMETHING_TIMEOUT);` (and then there is a timeout for the creation of a machine) else if we increase/decrease timeout per test, I'm not sure we need to declare these constants anyway

tests/playwright/src/specs/podman-machine-tests.spec.ts
```diff
@@ -165,14 +166,19 @@ test.describe.serial(`Podman machine switching validation `, () => {
 
   test('Clean up rootless podman machine', async ({ page }) => {
     await deletePodmanMachine(page, ROOTLESS_PODMAN_MACHINE_VISIBLE);
+
+    if (isWindows) {
+      await handlePopupDialog(page, 'Yes');
```
odockal: We have `handleConfirmationDialog` in utilities package, I think it should serve well for this purpose.

tests/playwright/src/specs/podman-machine-onboarding.spec.ts
cbr7: The constant is used to define the wait time for the podman machine creation, but the test itself does more than just create the machine so it need to be longer than the podman machine creation wait time.

Currently the podman machine creation time is however long the constant is initialized at but the test itself timesout at 60 seconds because that is the default value for playwright.

odockal (dismissed):

tests/playwright/src/specs/podman-machine-onboarding.spec.ts
benoitf: I would expect to have like a `const PODMAN_FULL_STARTUP_TIMEOUT = PODMAN_MACHINE_STARTUP_TIMEOUT + 30000` (but where constants are defined, not in the test)

cbr7: agreed, changed to this form now.

tests/playwright/src/specs/podman-machine-tests.spec.ts
cbr7: done

benoitf (approved):

ScrewTSW (approved):
LGTM

cdrage (approved):

cbr7:
e2e test runs with these fixes [here](https://github.com/containers/podman-desktop-e2e/actions/runs/10904872975)

PR #8924, opened 2024-09-17
fix: change Containerfile relative path to build context

SoniaSandler:
### What does this PR do?
This PR fixes the Containerfile's relative path when building an image.

https://github.com/containers/podman-desktop/issues/8460

### How to test this PR?
- Open build image page
- Choose containerfile to use for the build
- Change the build context to something different from the default
- Check that the image is successfully built


SoniaSandler:
It looks like there is a built in function in the path module of node (path.relative) to get the relative path, but I was not able to import that module to `BuildImageFromContainerFile`. IIUC, it is not possible to import `node:path` on the renderer side.

PR #8931, opened 2024-09-17
fix: update app label before displaying pod yaml

### LINKED ISSUE #8412: Changing Pod name while deploying a container to Kubernetes does not update the YAML preview correctly
### Bug description

When trying to deploy a container to Kubernetes, if you change the Pod Name, the preview of the YAML file does not update correctly the 'app' field (the text stays one keystroke behind what it's supposed to). Notably, if you start the deployment, the text is correct in the file.

[Screencast from 2024-08-09 12-26-30.webm](https://github.com/user-attachments/assets/ae407e38-ff04-4d8a-ad7e-857cf2144f85)

### Operating system

Windows 10 Pro

### Installation Method

None

### Version

1.12.0

### Steps to reproduce

1. Choose a container, and click on 'Deploy to Kubernetes'
2. Change the value from the 'Pod Name' field

### Relevant log output

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

It just reorders the reactive clauses so that the `updateKubeResult` is called after the app label is updated

### Screenshot / video of UI

update_yaml

it fixes #8412 

### How to test this PR?

1. update the pod name on the deploy pod page and check that the name and app labels in the generated yaml are the same

### DIFF
--- a/packages/renderer/src/lib/pod/DeployPodToKube.spec.ts
+++ b/packages/renderer/src/lib/pod/DeployPodToKube.spec.ts
@@ -23,6 +23,7 @@
 import '@testing-library/jest-dom/vitest';
 
 import { fireEvent, render, screen, waitFor } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
 import * as jsYaml from 'js-yaml';
 import { tick } from 'svelte';
 import { router } from 'tinro';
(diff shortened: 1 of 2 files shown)


axel7083 (commented):
Could we have a unit test ?

lstocchi:
> Could we have a unit test?

Done. Actually there was already a test for this use case `'When modifying the pod name, metadata.apps.label should also have been changed'` but it was a copy/paste from another one and it was not related at all with the pod name. Fixed.

axel7083 (approved):
Great, thanks :rocket:

benoitf (approved):

Issue #8939, opened 2024-09-18
Increase timeout value on pull-image-page.ts for test stability


danivilla9:
### Is your enhancement related to a problem? Please describe

During recent E2E test executions on CI, the volume-smoke.spec.ts case is running out of time while waiting for the download of an image (quay.io/centos-bootc/bootc-image-builder) on the pull-image-page-ts here:
> 48 |     const doneButton = this.page.getByRole('button', { name: 'Done' });
> 49 |     await playExpect(doneButton).toBeEnabled({ timeout: timeout }); //out of time

### Describe the solution you'd like

Increase the timeout from 60 seconds to 120 seconds in order to reduce the possibility that external causes that could delay the download affect the results of the test.

PR #8940, opened 2024-09-18
chore: increase timeout on pull-image-page

danivilla9:
### What does this PR do?
Increase the timeout on pull-image-page to increase test stability
#8939 

### How to test this PR?
Run `pnpm test:e2e:smoke`

### DIFF
--- a/tests/playwright/src/model/pages/pull-image-page.ts
+++ b/tests/playwright/src/model/pages/pull-image-page.ts
@@ -39,7 +39,7 @@ export class PullImagePage extends BasePage {
     this.imageNameInput = page.getByLabel('Image to Pull');
   }
 
-  async pullImage(imageName: string, tag = '', timeout = 60000): Promise<ImagesPage> {
+  async pullImage(imageName: string, tag = '', timeout = 120000): Promise<ImagesPage> {
     const fullImageName = `${imageName}${tag.length === 0 ? '' : ':' + tag}`;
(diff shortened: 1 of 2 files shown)


axel7083 (approved):

tests/playwright/src/model/pages/pull-image-page.ts
```diff
@@ -39,7 +39,7 @@ export class PullImagePage extends BasePage {
     this.imageNameInput = page.getByLabel('Image to Pull');
   }
 
-  async pullImage(imageName: string, tag = '', timeout = 60000): Promise<ImagesPage> {
+  async pullImage(imageName: string, tag = '', timeout = 120000): Promise<ImagesPage> {
```
odockal: I would change rather the place you call the `pullImage` from than changing it in everywhere. Look here: https://github.com/containers/podman-desktop/blob/main/tests/playwright/src/specs/volume-smoke.spec.ts#L122. And make sure to have enought timeout set for the test function in the spec.

odockal (changes requested):

tests/playwright/src/model/pages/pull-image-page.ts
danivilla9: That makes complete sense. I've put it in the correct place and added 60 seconds to the timeout of the case, as that's what I added to the pullImage part.

cbr7 (approved):

odockal (approved):

PR #8956, opened 2024-09-18
feat(extension-api): adding the id property to the WebviewPanel object

axel7083:
### What does this PR do?

To use the `navigation.navigateToWebview` we need the `id` of the webview, but it is not on the WebviewPanel object...
 The only way to get this object is to list the webviews and search by title `windows.listWebviews().find(panel => panel.title === "<my-title>"`.

N/A

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/webview/webview-panel-impl.ts
+++ b/packages/main/src/plugin/webview/webview-panel-impl.ts
@@ -76,6 +76,10 @@ export class WebviewPanelImpl implements WebviewPanel {
     this.#viewType = panelDetails.viewType;
   }
 
+  get id(): string {
+    return this.internalId;
+  }
+
   get internalId(): string {
     return this.#internalId;
   }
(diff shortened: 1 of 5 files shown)


benoitf (changes requested):
edited:
we have internalId so it means we want to keep it secret/hidden, if we need exposure, need to rename

benoitf:
to comment to myself, if `id` needs to be a public thing, then the variable name should not be 'internalId' that we expose at the end.

packages/main/src/plugin/webview/webview-panel-impl.ts
```diff
@@ -76,6 +76,10 @@ export class WebviewPanelImpl implements WebviewPanel {
     this.#viewType = panelDetails.viewType;
   }
 
+  get id(): string {
```
benoitf, line 79: if id needs exposure then probably to rename `#internalId` to `#id`

and just rename existing
```
get internalId(): string {
```

to 
```
get id(): string {
  this.#id
}
```

axel7083: Good idea

benoitf (approved):

benoitf:
on hold, maybe switch to the usage of `webview.reveal()` method

axel7083:
> on hold, maybe switch to the usage of `webview.reveal()` method

Just tested, the method is working well, closing

Issue #8775, opened 2024-09-06
simplify kubernetes-context-state.ts 


axel7083:
### Is your enhancement related to a problem? Please describe

The [kubernetes-context-state.ts](https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts) file is very hard to apprehend due to its complexity and numerous objects definition and implementation in it.

## Types definitions

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L71

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L88

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L102-L104

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L110-L112

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L122-L124

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L149-L151

## Interfaces

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L74

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L80

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L115

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L126

[...]

PR #8960, opened 2024-09-18
refactor: move ContextsStates to its own file + more unit tests

feloy:

### What does this PR do?

refactor: move ContextsStates to its own file + more unit tests

Part of #8775 

### How to test this PR?


packages/preload/src/index.ts
```diff
@@ -94,12 +94,12 @@ import type {
 } from '../../main/src/plugin/dockerode/libpod-dockerode';
 import type { CatalogExtension } from '../../main/src/plugin/extensions-catalog/extensions-catalog-api';
 import type { FeaturedExtension } from '../../main/src/plugin/featured/featured-api';
+import type { ContextGeneralState, ResourceName } from '../../main/src/plugin/kubernetes/contexts-states';
```
axel7083: Since you created the file, can we move it to the `/@/api` ? Or at least their types to avoid those main imports in non-main package

jeffmaury (approved):
LGTM

benoitf (approved):

packages/preload/src/index.ts
feloy: done

packages/api/src/kubernetes-contexts-states.ts
```diff
@@ -0,0 +1,58 @@
...
+  'services',
+  'ingresses',
+  'routes',
+  'configmaps',
+  'secrets',
+  'nodes',
+  'persistentvolumeclaims',
+] as const;
+
+export function isSecondaryResourceName(value: string): value is SecondaryResourceName {
```
axel7083: I don't think we want to have functions exposed in the api packages, moreover this function is not used by any package other than the main ?

feloy: I also hesitate, I placed it here because it is very close to the `secondaryResourceName` type defined below. I'll move it back to the main package

axel7083 (approved):
LGTM, thanks for the time spent on this 🚀

PR #8961, opened 2024-09-18
fix: update containerState in effect

SoniaSandler:

### What does this PR do?
This PR fixes the reactivity of `ContainerDetailsTerminal` so that `containerState` is updated every time `$effect` runs.
This fixes the problem of `containerState` not always updating, causing the terminal to not work when starting the container from the terminal tab or stopping and starting again.

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
@@ -31,6 +31,7 @@ let lastState = $state('');
 let containerState = $state(container);
 
 $effect(() => {
+  containerState = container;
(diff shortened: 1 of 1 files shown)


benoitf (approved):
If possible, add a test that is reproducing the issue

SoniaSandler:
@benoitf , I'm not sure if it is possible since the tests passed also without this change (I think because of `rerender` with the container state we want), but when using PD manually, the container status can change when already in this page without the need to update/rerender the page. So unless there is a way to update `container.state` (STARTING/RUNNING/EXITTED) dynamically without rerendering the page, I don't think this is something that could be reproduced in unit tests

deboer-tim (approved):

PR #8963, opened 2024-09-19
feat: add default actions for Pods empty state

### LINKED ISSUE #8570: Default actions for Pods empty state
There should be two options in the following screen to allow the user to quickly get started:

Screenshot 2024-08-25 at 13 05 15

The first one should be a button:
"Start your first pod"

Ideally, we would also like to provide sample pods (we might be able to pull those from https://github.com/redhat-developer/podman-desktop-demo

Then in the screen, we display a label "OR" and the line bellow is the actual commande line we already have in the screen

dgolovin:
### What does this PR do?

Adds button 'Start my first pod' to Pods page empty screen

### Screenshot / video of UI

https://github.com/user-attachments/assets/88cdfa1d-4894-45dd-8e4f-cfe7f20e5c0a

Fix #8570.

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/pod/PodEmptyScreen.spec.ts
+++ b/packages/renderer/src/lib/pod/PodEmptyScreen.spec.ts
@@ -1,0 +1,127 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/pod/PodEmptyScreen.spec.ts
```diff
@@ -0,0 +1,50 @@
+/**********************************************************************
+ * Copyright (C) 2023-2024 Red Hat, Inc.
```
feloy:
```suggestion
 * Copyright (C)2024 Red Hat, Inc.
```

packages/renderer/src/lib/pod/PodEmptyScreen.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+
+import { afterEach } from 'node:test';
```
feloy:
```suggestion
import { afterEach } from 'vitest';
```

feloy (changes requested):

feloy (approved):
LGTM

I don't find very demonstrative to start a pod with only the infra container. I think it would be better to start a pod with some payload

dgolovin:
> LGTM
> 
> I don't find very demonstrative to start a pod with only the infra container. I think it would be better to start a pod with some payload

Sounds good to me, because without payload even terminal cannot function. Will look into it.

benoitf:
setting draft mode as there will be changes

feloy (approved):
LGTM
Thanks for adding the hello container

PR #8971, opened 2024-09-19
chore: changed type and casting in input and numberimput component

### LINKED ISSUE #8743: Input widget provides value type being string or number or undefined but the type is only string
### Bug description

HTMLInputElement is defining value as string

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/value

but there Input ui component is defining as https://github.com/containers/podman-desktop/blob/main/packages/ui/src/lib/inputs/Input.svelte#L9

but in fact it's not really a number so there are some cast to Number 
https://github.com/containers/podman-desktop/blob/main/packages/renderer/src/lib/ui/NumberInput.svelte#L26

but normally we shouldn't have to cast if it was really a number

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

the typescript type is incorrect in Input.svelte and NumberInput.svelte so at the end we're doing some conversion

the new sveltecheck is throwing errors as it detects that we can't map numbers to string, etc

### Relevant log output

_No response_

### Additional context

https://github.com/containers/podman-desktop/pull/8721

gastoner:
### What does this PR do?
Changes types and casting between string and number

Required for #8721 
Closes #8743 
### How to test this PR?
`pnpm typecheck`
And try increase/decrease numbers in NumberInput component (e.g. preferences->zoom)

### DIFF
--- a/packages/renderer/src/lib/ui/NumberInput.svelte
+++ b/packages/renderer/src/lib/ui/NumberInput.svelte
@@ -17,13 +17,14 @@ export let onValidation = (_value: number, _error?: string) => {};
 
 let minimumEnabled: boolean;
 let maximumEnabled: boolean;
+$: valueAsString = String(value);
 
-$: if (value !== undefined || disabled) {
+$: if (valueAsString !== undefined || disabled) {
   validateNumber();
 }
 
 function validateNumber() {
-  const numberToValidate = Number(value);
+  const numberToValidate = Number(valueAsString);
   if (maximum !== undefined && numberToValidate > maximum) {
     error = `The value cannot be greater than ${maximum}`;
   } else if (minimum !== undefined && numberToValidate < minimum) {
@@ -34,6 +35,8 @@ function validateNumber() {
   minimumEnabled = !disabled && (minimum === undefined || minimum < numberToValidate);
   maximumEnabled = !disabled && (maximum === undefined || maximum > numberToValidate);
 
+  // update converted value
+  value = numberToValidate;
   // send the callback
   onValidation(numberToValidate, error);
 }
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/ui/NumberInput.svelte
```diff
@@ -17,13 +17,14 @@ export let onValidation = (_value: number, _error?: string) => {};
 
 let minimumEnabled: boolean;
 let maximumEnabled: boolean;
+$: valueAsString = String(value);
 
-$: if (value !== undefined || disabled) {
+$: if (valueAsString !== undefined || disabled) {
```
lstocchi, line 22: Maybe check if value is undefined when you convert it to String? I think that String(undefined) = "undefined"

benoitf:
side note: it looks like the check for semantic commits is ✅ as commits are using semantics but the title of this PR is not.
might need to tune the configuration file to flag this PR as 🟥 b/c title should also follow semantic convention.

packages/renderer/src/lib/ui/NumberInput.svelte
```diff
@@ -17,13 +17,14 @@ export let onValidation = (_value: number, _error?: string) => {};
 
 let minimumEnabled: boolean;
 let maximumEnabled: boolean;
+$: valueAsString = String(value);
```
benoitf:
```suggestion
$: valueAsString = value !== undefined ? String(value) : undefined;
```

same suggestion than @lstocchi

benoitf (approved):
approving modulo suggestion

gastoner:
@lstocchi @benoitf Updated!

benoitf (approved):

lstocchi (approved):
LGTM

Issue #8571, opened 2024-08-25
Default actions for Kubernetes screens and their empty state


slemeur:
As a user, I might not have a kubernetes environment configured and up. In this context, the user will see the screen empty, without noticing that Podman Desktop can be used to setup a kubernetes environment locally.

<img width="1162" alt="Screenshot 2024-08-25 at 13 05 08" src="https://github.com/user-attachments/assets/e5e9407c-d207-458a-b41b-41eeb9e12c89">

In this issue, we would like to highlight the available extensions to configure a kubernetes environment.

We should split the empty screen into two sections:
1- Run Kubernetes Locally
--> Actions for Minikube, Kind and OpenShift Local extensions

2- Or use free remote Kubernetes environments
--> Developer Sandbox extension

For each of the extensions:
- Title, description
- Button to jump to the install page
- Button to open the documentation

feloy:
> In this context, the user will see the screen empty, without noticing that Podman Desktop can be used to setup a kubernetes environment locally.

Please note that when there is no Kubernetes context, the Kubernetes-related buttons are hidden (including the Deployment button), so there is no possibility for the user to land into this page in this case

feloy:
When there is no current context:
- [x] display an "empty page" with the list of uninstalled extensions of the Kubernetes category: #8989
- [x] filter Kubernetes extensions to display only the ones we want
- [x] display links to start resources for installed Kubernetes extensions
- [x] hide Kubernetes submenu (#9087)
- [x] Telemetry (#9106)
- [ ] design to have more focus on first section (creating resources)

benoitf:
closing as done, enhancements are related to https://github.com/containers/podman-desktop/issues/9142

PR #8973, opened 2024-09-19
fix: do not display badge when no current context

feloy:

### What does this PR do?

On Kubernetes pages, do not display the connection badge when there is no current context.

(other PRs will follow to also hide other unnecessary buttons when there is no context)

Part of #8571 

### How to test this PR?

Start Podman Desktop with no current Kubernetes context, and check that the badge is not displayed in Kubernetes pages.

Define a current context and check that the badge is visible.


feloy:
closing as should not be necessary

PR #8985, opened 2024-09-19
feat: handle both hyperv and wsl machines

### LINKED ISSUE #8862: Manage WSL and HyperV machines
If WSL is enabled collect and manage Podman machine using CONTAINERS_MACHINE_PROVIDER=wsl
If HyperV is enabled collect and manage Podman machine using CONTAINERS_MACHINE_PROVIDER=hyperv

lstocchi:
### What does this PR do?

This PR enhances Desktop to show both WSL and HyperV machines if both are enabled. 
I added a missing check for hyperv - desktop has to be run with elevated rights

### Screenshot / video of UI

wsl_hyperv

it resolves #8862 

### How to test this PR?

1. create a hyper-v and wsl machines using the cli, then you should see both on desktop and being able to interact with them

### DIFF
--- a/extensions/podman/packages/extension/src/extension.spec.ts
+++ b/extensions/podman/packages/extension/src/extension.spec.ts
@@ -40,6 +40,7 @@ import { PodmanConfiguration } from './podman-configuration';
 import type { UpdateCheck } from './podman-install';
 import { PodmanInstall } from './podman-install';
 import { getAssetsFolder, isLinux, isMac, isWindows, LIBKRUN_LABEL, LoggerDelegator, VMTYPE } from './util';
+import * as util from './util';
 
 const config: Configuration = {
   get: () => {
@@ -270,21 +271,6 @@ vi.mock('./podman-info-helper', async () => {
       return {
         updateWithPodmanInfoRecords: vi.fn().mockImplementation(() => {
           return Promise.resolve();
-        }),
-      };
-    }),
-  };
-});
-vi.mock('./wsl-helper', async () => {
-  return {
-    WslHelper: vi.fn().mockImplementation(() => {
-      return {
-        getWSLVersionData: vi.fn().mockImplementation(() => {
-          return Promise.resolve({
-            wslVersion: '1.2.3',
-            kernelVersion: '1.2.3',
-            windowsVersion: '1.2.3',
-          });
         }),
       };
     }),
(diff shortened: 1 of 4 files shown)


jeffmaury (commented):
For me we must handle WSL and HyperV separately so that if WSL is here we collect WSL machines if HyperV is here we collect HyperV machines. Seems your code requires both to be installed

lstocchi:
> For me we must handle WSL and HyperV separately so that if WSL is here we collect WSL machines if HyperV is here we collect HyperV machines. Seems your code requires both to be installed

I just check if both are installed bc only in that case it behaves differently from the normal behavior. If only one between WSL or HyperV is enabled, it works fine without any change.

jeffmaury:
> > For me we must handle WSL and HyperV separately so that if WSL is here we collect WSL machines if HyperV is here we collect HyperV machines. Seems your code requires both to be installed
> 
> I just check if both are installed bc only in that case it behaves differently from the normal behavior. If only one between WSL or HyperV is enabled, it works fine without any change.

Not sure to get it does it rely on having the provider configured in containers.conf ? Or did I miss something ?

lstocchi:
> > I just check if both are installed bc only in that case it behaves differently from the normal behavior. If only one between WSL or HyperV is enabled, it works fine without any change.
> 
> Not sure to get it does it rely on having the provider configured in containers.conf ? Or did I miss something ?

This issue is just related to handle both wsl and hyperv. So i assume that hyperv (when the only provider) works fine. And if you want to work with hyperv you have to either set the env variable or configure the containers.conf as mentioned in the doc. And in that case, it will end up in the else clause below the else if i just added so nothing has to be done.
https://github.com/containers/podman-desktop/blob/main/website/docs/installation/windows-install/index.md

However it seems to me that what you want is allowing ppl to use hyperv without setting the env variable or updating the containers.conf file. We see that hyperv is enabled and we handled everything.

So this issue is more "we check by ourselves if the user can work with wsl and hyper and we enabled the one (possibly both) that can run. No matter the configuration/env variable set by the user", correct?

Update: ok after re-reading the issue it's like that. Now i get why it was written that way 😆

lstocchi:
@jeffmaury updated

jeffmaury (commented):
I have one WSL (started) machine and one HyperV (stopped)

On a normal shell I see only the WSL one => good
On an elevated shell, I see only the WSL one => not good

lstocchi:
> I have one WSL (started) machine and one HyperV (stopped)
> 
> On a normal shell I see only the WSL one => good On an elevated shell, I see only the WSL one => not good

Works fine on my side. I use the terminal for my test. 
Maybe the hyperv check fails somewhere. Could you try running the commands we use on the elevated shell?

for elevated check
```
powershell.exe (New-Object Security.Principal.WindowsPrincipal([Security.Principal.WindowsIdentity]::GetCurrent())).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
```

for hyperv installed
```
powershell.exe', ['Get-Service vmms']
```

for hyperv running
```
'powershell.exe', ['@(Get-Service vmms).Status']
```

jeffmaury:
> > I have one WSL (started) machine and one HyperV (stopped)
> > On a normal shell I see only the WSL one => good On an elevated shell, I see only the WSL one => not good
> 
> Works fine on my side. I use the terminal for my test. Maybe the hyperv check fails somewhere. Could you try running the commands we use on the elevated shell?
> 
> for elevated check
> 
> ```
> powershell.exe (New-Object Security.Principal.WindowsPrincipal([Security.Principal.WindowsIdentity]::GetCurrent())).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
> ```
True
> 
> for hyperv installed
> 
> ```
> powershell.exe', ['Get-Service vmms']
Status   Name               DisplayName
------   ----               -----------
Running  vmms               Gestion d’ordinateurs virtuels Hyper-V
> ```
> 
> for hyperv running
> 
> ```
> 'powershell.exe', ['@(Get-Service vmms).Status']
Running

> ```

jeffmaury (approved):
LGTM

feloy (commented):
LGTM

I can see both WSL and HyperV machines on Podman Desktop, and pull images, run containers with both providers

I first created both machines with the same name and there was a conflict on the display, only one was displayed.

If I create both machines with different names, they are both displayed

feloy (approved):

lstocchi:
> I can see both WSL and HyperV machines on Podman Desktop, and pull images, run containers with both providers
> 
> I first created both machines with the same name and there was a conflict on the display, only one was displayed.
> 
> If I create both machines with different names, they are both displayed

Thanks for the feedback. Gonna merge it as this is just the first part and i think the conflict should be handled by podman itself with the system connection name but we can fix it in a later PR if there is some misbehavior

PR #8986, opened 2024-09-19
chore: update style of xterm terminal

### LINKED ISSUE #8079: Add padding to all terminal screens.

Screenshot 2024-07-15 at 10 47 59 AM

I've noticed throughout all our terminal outputs it's a bit hard to read as the letters go right to the edge.

This wasn't very noticeable on dark mode, but with light mode it's very noticiable.

Padding to all terminal screens.

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
This PR adds padding to all terminal instances

### Screenshot / video of UI
before:
Screenshot from 2024-09-19 15-17-49
Screenshot from 2024-09-19 13-48-29

after:
Screenshot from 2024-09-19 17-08-26

Screenshot from 2024-09-19 17-08-13

Closes https://github.com/containers/podman-desktop/issues/8079

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 import '@xterm/xterm/css/xterm.css';
-import '../ui/xterm.css';
 
 import { EmptyScreen } from '@podman-desktop/ui-svelte';
 import { FitAddon } from '@xterm/addon-fit';
@@ -160,7 +159,11 @@ onDestroy(() => {
 });
 </script>
 
-<div class="h-full" bind:this={terminalXtermDiv} class:hidden={container.state !== 'RUNNING'}></div>
+<div
+  class="h-full p-[5px] pr-0 bg-[var(--pd-terminal-background)]"
+  bind:this={terminalXtermDiv}
+  class:hidden={container.state !== 'RUNNING'}>
+</div>
 
 <EmptyScreen
   hidden={container.state === 'RUNNING'}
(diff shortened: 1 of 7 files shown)


benoitf (commented):
hello, it is not possible to include the padding in the terminal component ?

AFAIK we should not have to include css file each time we're including a component

SoniaSandler:
@benoitf I went through the `xterm.d.ts` file and couldn't find anything for padding, and while searching in the xterm repo, I only found the option of setting the padding in the CSS file.
However, I think I found a workaround by setting the background of the terminal's div to be the same as the background of the terminal and then setting the padding for the div, so I'll update this PR.

benoitf (commented):
side note not for this commit: should we remove also import `import '@xterm/xterm/css/xterm.css';` from the various place

as it's imported by TerminalWindow

jeffmaury (approved):
LGTM. Shoudn't we have the same padding left and right ?

gastoner (approved):
LGTM

SoniaSandler:
@jeffmaury on the right, xterm terminals have the scrollbar, and since this padding of for the div that the terminal is inside and not the terminal itself, having also padding on the right side will push the scrollbar a bit to the left with empty space to its right. I'm not sure it is something we would like to have

SoniaSandler:
@benoitf there are a few instances of xterm terminals that don't use `TerminalWindow`, so to remove the other `import '@xterm/xterm/css/xterm.css';` except for the one in `TerminalWindow`, we need to move all the remaining terminals to use `TerminalWindow`. I'll open an issue for that.

benoitf:
Ok thanks

benoitf (approved):

PR #8989, opened 2024-09-20
feat: display Kubernetes empty page when no current context

feloy:

### What does this PR do?

Display Kubernetes empty page when no current context

Part of #8571, see https://github.com/containers/podman-desktop/issues/8571#issuecomment-2362916912 for next steps

### Screenshot / video of UI

https://github.com/user-attachments/assets/0caafe75-e3d5-478a-997a-8a9ec435dd41

Part of #8571

There is a glitch on the Download button after the download is finished and the extension is removed from the list (the button of the next extension is displayed as Downloading for a second). Visible in the video around 0:16. The fix will be done as part of another PR

### How to test this PR?

### DIFF
--- a/packages/renderer/src/App.spec.ts
+++ b/packages/renderer/src/App.spec.ts
@@ -24,6 +24,7 @@ import { beforeEach, expect, test, vi } from 'vitest';
 
 import * as kubeContextStore from '/@/stores/kubernetes-contexts-state';
 import type { ContextGeneralState } from '/@api/kubernetes-contexts-states';
+import { NO_CURRENT_CONTEXT_ERROR } from '/@api/kubernetes-contexts-states';
 
 import App from './App.svelte';
 import { navigationRegistry } from './stores/navigation/navigation-registry';
@@ -170,7 +171,7 @@ test('do not display kubernetes empty screen if current context', async () => {
 test('displays kubernetes empty screen if no current context', async () => {
   vi.mocked(kubeContextStore).kubernetesCurrentContextState = readable({
     reachable: false,
-    error: 'no current context',
+    error: NO_CURRENT_CONTEXT_ERROR,
     resources: { pods: 0, deployments: 0 },
   } as ContextGeneralState);
 
--- a/packages/renderer/src/App.spec.ts
+++ b/packages/renderer/src/App.spec.ts
@@ -18,8 +18,12 @@
 
 import { render, waitFor } from '@testing-library/svelte';
 import { tick } from 'svelte';
+import { readable } from 'svelte/store';
 import { router } from 'tinro';
 import { beforeEach, expect, test, vi } from 'vitest';
+
+import * as kubeContextStore from '/@/stores/kubernetes-contexts-state';
+import type { ContextGeneralState } from '/@api/kubernetes-contexts-states';
 
 import App from './App.svelte';
 import { navigationRegistry } from './stores/navigation/navigation-registry';
@@ -29,6 +33,8 @@ const mocks = vi.hoisted(() => ({
   RunImage: vi.fn(),
   ImagesList: vi.fn(),
   SubmenuNavigation: vi.fn(),
+  KubernetesEmptyPage: vi.fn(),
+  DeploymentsList: vi.fn(),
 }));
 
 vi.mock('./lib/dashboard/DashboardPage.svelte', () => ({
@@ -60,6 +66,18 @@ vi.mock('./SubmenuNavigation.svelte', () => ({
   default: mocks.SubmenuNavigation,
 }));
 
+vi.mock('./lib/kube/KubernetesEmptyPage.svelte', () => ({
+  default: mocks.KubernetesEmptyPage,
+}));
+
+vi.mock('./lib/deployments/DeploymentsList.svelte', () => ({
+  default: mocks.DeploymentsList,
+}));
+
+vi.mock('/@/stores/kubernetes-contexts-state', async () => {
+  return {};
+});
+
 const dispatchEventMock = vi.fn();
 const messages = new Map<string, (args: any) => void>();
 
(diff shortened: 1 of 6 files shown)


gastoner (approved):
LGTM, tested on Fedora

benoitf:
I like the iterations but IMHO it should be applied in some of reverse order.

1. redirect people to the existing resources being installed (if no cluster then I should have a redirect to kind extension first)
no need to install something else.

1. enhance the embedded catalog to filter resources (keywords)

1. enhance catalog of extensions with keywords for extensions that can bring a kubernetes cluster (locally or remotely)

1. then include the catalog to this page (this PR) where people will see the page

IMHO the filtering is a pre-requisites before showing the link to the available extensions



----

On the page showed by the PR, I would expect maybe a note just on the top of the catalog explaining why it's displayed.

kind of sentence : `Don't have a kubernetes cluster, install one these extensions to help you to bring one:`

jeffmaury (commented):
Not sure if it is to be reviewed as it is a draft

feloy:
> Not sure if it is to be reviewed as it is a draft

No, there is still work to do on this PR

feloy:
> Not sure if it is to be reviewed as it is a draft

The features are now ready for review ~(still in Draft mode as I need to write some more unit tests)~

packages/renderer/src/App.spec.ts
```diff
@@ -18,9 +18,13 @@
 
 import { render, waitFor } from '@testing-library/svelte';
 import { tick } from 'svelte';
+import { readable } from 'svelte/store';
 import { router } from 'tinro';
 import { beforeEach, expect, test, vi } from 'vitest';
 
+import * as kubeContextStore from '/@/stores/kubernetes-contexts-state';
+import type { ContextGeneralState } from '/@api/kubernetes-contexts-states';
```
dgolovin, line 26:
```suggestion
import type { NO_CURRENT_CONTEXT_ERROR, ContextGeneralState } from '/@api/kubernetes-contexts-states';
```

packages/renderer/src/App.spec.ts
```diff
@@ -135,3 +158,25 @@ test('opens submenu when a `submenu` menu is opened', async () => {
...
+  router.goto('/kubernetes/deployments');
+  await tick();
+  expect(mocks.KubernetesEmptyPage).not.toHaveBeenCalled();
+  expect(mocks.DeploymentsList).toHaveBeenCalled();
+});
+
+test('displays kubernetes empty screen if no current context', async () => {
+  vi.mocked(kubeContextStore).kubernetesCurrentContextState = readable({
+    reachable: false,
+    error: 'no current context',
```
dgolovin:
```suggestion
    error: NO_CURRENT_CONTEXT_ERROR,
```

packages/renderer/src/lib/kube/KubernetesEmptyPage.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+// mock the router
+vi.mock('tinro', () => {
+  return {
+    router: {
+      goto: vi.fn(),
+    },
+  };
+});
+
+test('expect to call EmbeddableCatealogExtensionList for Kubernetes providers local and remote', () => {
```
dgolovin:
```suggestion
test('expect to call EmbeddableCatalogExtensionList for Kubernetes providers local and remote', () => {
```

dgolovin (approved):
LGTM + nits

axel7083 (commented):
First I want to say: Love it!

A small issue I encountered 

Screenshot from 2024-09-25 09-58-18

I do not see minikube option, even tho I have it installed and enabled

image

I tried enabling/disabling it, same result.

feloy:
> First I want to say: Love it!
> 

Thanks :) 


> A small issue I encountered
> 
> 
> I do not see minikube option, even tho I have it installed and enabled
> 
> 
> I tried enabling/disabling it, same result.


The logic should be the same as the "Create new..." button in the Resources page, for Minikube. Is the button visible there?

I have seen the same behaviour with another provider, where the button was not visible because a resource was already declared. Perhaps you already have a minikube cluster running?

benoitf (approved):
as a feedback, I think my eyes are more seeing the installation of other extensions than seeing the ability to already being able to create a cluster.

I don't know if we could balance more the upper part or add some divider between the top and bottom screen, or making more appealing the links to create a cluster

but it could be follow-up as it's a personal view.

axel7083:
> The logic should be the same as the "Create new..." button in the Resources page, for Minikube. Is the button visible there?

Oh yeah, the minikube executable is not installed, therefore cannot create a cluster. That's a bit problematic, as when the user click for the first time to install the extension it would not allow him to directly create the cluster, since it would need to download the CLI first

packages/renderer/src/App.svelte
```diff
@@ -224,76 +227,82 @@ window.events?.receive('navigate', (navigationRequest: unknown) => {
...
-        <Route
-          path="/kubernetes/ingressesRoutes/route/:name/:namespace/*"
-          breadcrumb="Route Details"
-          let:meta
-          navigationHint="details">
-          <RouteDetails name={decodeURI(meta.params.name)} namespace={decodeURI(meta.params.namespace)} />
-        </Route>
+        {#if $kubernetesCurrentContextState.error === NO_CURRENT_CONTEXT_ERROR}
+          <Route path="/kubernetes/*" breadcrumb="Kubernetes" navigationHint="root">
+            <KubernetesEmptyPage />
```
axel7083, line 232: Would it be possible not to have the navigation menu (Nodes, config etc.) when we are on the empty page ? could be a follow-up PR (not sure if it is already plan)

feloy: Yes, it should be the next step: https://github.com/containers/podman-desktop/issues/8571#issuecomment-2362916912

axel7083: Perfect ! Thanks :+1:

jeffmaury (approved):
LGTM. 

In line with @benoitf comments, having more focus on Create would be nice

feloy:
> as a feedback, I think my eyes are more seeing the installation of other extensions than seeing the ability to already being able to create a cluster.
> 
> I don't know if we could balance more the upper part or add some divider between the top and bottom screen, or making more appealing the links to create a cluster
> 
> but it could be follow-up as it's a personal view.


@ekidneyrh would you have some advice to make the first part more visible?

kubernetes-empty-page

PR #8995, opened 2024-09-20
chore: rework variables to avoid potential undefined values

benoitf:
### What does this PR do?
avoid to have `wantedOses.keys().next().value` that can be undefined

in newer typescript version, it fixed the behavior of getting the first value of an iterator that can return undefined

but then map.get(undefined) is not possible and then leading to an error

prevent that by ensuring that we have valid values before continuing

related to https://github.com/containers/podman-desktop/pull/8814

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/image-registry.ts
+++ b/packages/main/src/plugin/image-registry.ts
@@ -718,7 +718,7 @@ export class ImageRegistry {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  getBestManifest(manifests: any[], wantedArch: string, wantedOs: string): any {
+  getBestManifest(manifests: any[], wantedArch: string, wantedOs: string): any | undefined {
     // eslint-disable-next-line etc/no-commented-out-code
     // manifestsMap [os] [arch] = manifest
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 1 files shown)


gastoner (approved):

packages/main/src/plugin/image-registry.ts
```diff
@@ -718,7 +718,7 @@ export class ImageRegistry {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  getBestManifest(manifests: any[], wantedArch: string, wantedOs: string): any {
+  getBestManifest(manifests: any[], wantedArch: string, wantedOs: string): any | undefined {
```
axel7083, line 721: Should an error be throw instead of returning `undefined` ?

jeffmaury (approved):
LGTM

packages/main/src/plugin/image-registry.ts
benoitf: it's already returning undefined today as a normal case, I just added it into the signature to make it clearer

axel7083 (approved):

packages/main/src/plugin/image-registry.ts
benoitf: https://github.com/containers/podman-desktop/blob/e070a2a678fae3a22a4754e3d24ee66c862c6715/packages/main/src/plugin/image-registry.ts#L707-L717

axel7083: I know convention naming can be discussed, but shouldn't we name `findBestManifest` then?

benoitf: I agree on the naming, but it'll be a follow-up PR

axel7083: np

PR #8997, opened 2024-09-20
chore(ci): adding concurrency option to pr-check

### LINKED ISSUE #8811: Using the concurrency option in github workflows

As per the documentation _By default, GitHub Actions allows multiple jobs within the same workflow, multiple workflow runs within the same repository, and multiple workflow runs across a repository owner's account to run concurrently. This means that multiple workflow runs, jobs, or steps can run at the same time._[^1]

This mean, if I commit first, then re-commit, the check for the first commit are running before the new one.

[^1]: https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#overview

I would expect the workflows running would be cancel when my new commit appear.

We could add something like the following to workflows running in the PR check, to only run for the last commit and cancel any existing workflows when a new one is triggered. 

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

_No response_

### Additional context

Why ? 🍃 🌎 🔥

axel7083:
### What does this PR do?

Adding the `concurrency` option to pr-check, making Github cancel previous github actions for a given branch if new one are started (E.g. new commits, rebase etc.)

### Screenshot / video of UI

N/A

Fixes https://github.com/containers/podman-desktop/issues/8811

### How to test this PR?

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -17,9 +17,13 @@
 
 name: pr-check
 
-on: 
+on:
   pull_request:
     types: [labeled, synchronize, opened, reopened]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   windows:
@@ -337,7 +341,7 @@ jobs:
         with:
           fetch-depth: 2
 
-      - name: Evaluate changes in files   
+      - name: Evaluate changes in files
         id: pnpm_changed
         run: |
           git fetch origin ${{ github.event.pull_request.base.ref }}
@@ -411,7 +415,7 @@ jobs:
       always()
     steps:
       - name: Evaluate the Windows Update test results
-        run: | 
+        run: |
           echo "Windows updater result: ${{ needs.run-update-e2e-test.result }}"
           if [ "${{ needs.run-update-e2e-test.result }}" = "failure" ]; then
             echo "Windows udpater test failed..."
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -22,7 +22,7 @@ on:
     types: [labeled, synchronize, opened, reopened]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
 jobs:


.github/workflows/pr-check.yaml
```diff
@@ -411,7 +415,7 @@ jobs:
       always()
     steps:
       - name: Evaluate the Windows Update test results
-        run: | 
+        run: |
           echo "Windows updater result: ${{ needs.run-update-e2e-test.result }}"
           if [ "${{ needs.run-update-e2e-test.result }}" = "failure" ]; then
             echo "Windows udpater test failed..."
```
axel7083, line 421: the linter did the changes

.github/workflows/pr-check.yaml
```diff
@@ -17,10 +17,14 @@
 
 name: pr-check
 
-on: 
+on:
   pull_request:
     types: [labeled, synchronize, opened, reopened]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
benoitf: looks like sometimes the pattern is 
```
${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

axel7083: Where did you find this ? Because I was reading https://github.com/opencv/ci-gha-workflow/pull/78 and it does not seems to work.

The [github documentation](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency) recommend `${{ github.workflow }}-${{ github.ref }}`

axel7083: _For example, you can use the concurrency keyword immediately after where trigger conditions are defined to limit the concurrency of entire workflow runs for a specific branch_ ref https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-using-concurrency-and-the-default-behavior

benoitf: https://github.com/pytorch/pytorch/blob/main/.github/workflows/pull.yml#L17
https://github.com/jaegertracing/jaeger/blob/main/.github/workflows/fossa.yml#L11

axel7083: What would be the difference compare to the value given in the github documentation ?

benoitf: you can change the branch of the PR I guess ?

axel7083:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
```

benoitf (approved):

PR #8998, opened 2024-09-20
fix(docusaurus): storybook empty page

### LINKED ISSUE #8840: Storybook empty page is not properly centered
### Bug description

image

It uses all the height available (which is determined by the sidebar, which is probably wrong

### Operating system

Fedora 40

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Removing the `align-item` as it was making the empty text vertically center, but when the navigation was containing too many items, it was making invisible as too bar bellow.

### Screenshot / video of UI

Before

image

After

image

Fixes https://github.com/containers/podman-desktop/issues/8840

### How to test this PR?

Uses the branch from https://github.com/containers/podman-desktop/pull/8836 and apply fix

### DIFF
--- a/website/src/pages/storybook/empty-page.tsx
+++ b/website/src/pages/storybook/empty-page.tsx
@@ -21,7 +21,7 @@ import React from 'react';
 
 export default function EmptyPage(): JSX.Element {
   return (
-    <div className="justify-center grow flex w-full">
+    <div className="justify-center items-center grow flex w-full">
       <span>No component selected</span>
     </div>
   );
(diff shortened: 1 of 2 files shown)


benoitf (commented):
do we have an issue between the left and the right

The sidebar on the left should scroll with an independent way of the right part, no ?

benoitf:
I think we might have a better outcome adding `h-screen` to the sidebar

https://github.com/containers/podman-desktop/blob/4b84cff01e735c8e469c53f7de3571b0c09ef1b1/website/src/pages/storybook/index.tsx#L100

at this line


https://github.com/user-attachments/assets/7b0b4049-4251-492f-8dbb-a58d1b59799b

axel7083:
> I think we might have a better outcome adding `h-screen` to the sidebar

Suggestion applied :+1:

benoitf (approved):

PR #8999, opened 2024-09-20
feat: filter embeddable extensions by keywords

feloy:

### What does this PR do?

Filter embeddable extensions by keywords

Part of #8571 (https://github.com/containers/podman-desktop/issues/8571#issuecomment-2362916912)

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extensions-catalog/extensions-catalog-api.ts
+++ b/packages/main/src/plugin/extensions-catalog/extensions-catalog-api.ts
@@ -36,6 +36,7 @@ export interface CatalogExtension {
   extensionName: string;
   displayName: string;
   categories: string[];
+  keywords: string[];
   unlisted: boolean;
   versions: CatalogExtensionVersion[];
 }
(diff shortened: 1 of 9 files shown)


benoitf (approved):
needs to update another spec file

packages/main/src/plugin/extensions-catalog/extensions-catalog-api.ts
```diff
@@ -36,6 +36,7 @@ export interface CatalogExtension {
   extensionName: string;
   displayName: string;
   categories: string[];
+  keywords: string[];
```
feloy, line 39: I'm not sure if we want to have these keywords mandatory, or optional?

feloy:
> needs to update another spec file

Yes, but I wonder if we want to add keywords everywhere, or make them optional. I can see they are set for all extensions in the catalog, do we want them to be required?

packages/main/src/plugin/extensions-catalog/extensions-catalog-api.ts
benoitf: it's stored in the remote catalog so yes

but I would say that if omitted in the online catalog, default to []

so it's always here but it will return an empty array if omitted online

benoitf (approved):
reapproving

PR #9013, opened 2024-09-23
feat: add options to customize [Embeddable]CatalogExtensionList

feloy:

### What does this PR do?

Add options to customize CatalogExtensionList and EmbeddableCatalogExtensionList

Will be helpful for #8989

Part of #8571 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/extensions/CatalogExtensionList.svelte
+++ b/packages/renderer/src/lib/extensions/CatalogExtensionList.svelte
@@ -7,7 +7,7 @@ import CatalogExtension from './CatalogExtension.svelte';
 
 export let catalogExtensions: CatalogExtensionInfoUI[];
 export let title: string = 'Available extensions';
-export let hideIfEmpty: boolean = false;
+export let hideEmpty: boolean = false;
 
 async function fetchCatalog() {
   try {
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/extensions/CatalogExtensionList.svelte
```diff
@@ -6,6 +6,8 @@ import type { CatalogExtensionInfoUI } from './catalog-extension-info-ui';
 import CatalogExtension from './CatalogExtension.svelte';
 
 export let catalogExtensions: CatalogExtensionInfoUI[];
+export let title: string = 'Available extensions';
+export let hideIfEmpty: boolean = false;
```
axel7083: `hideEmpty` is probably enough, I am not a big fan of having conditions in the naming,

other ideas would be `showEmptyScreen`, `displayEmptyScreen` or having a slot, and being empty by default

feloy: changed

axel7083 (approved):
LGTM

benoitf (commented):
I think it might be easier with the `show` property rather than `hide`

`hideEmpty` --> `showEmpty`

so we have `if showEmpty` condition rather than a double not : `if !hideEmpty`

benoitf (commented):
also at first I think it was something else so maybe add also the Screen to the property name

so we know it's about a screen that is displayed

benoitf (approved):

PR #9036, opened 2024-09-24
chore(test): fixup isCI was type (string|bool)

ScrewTSW:
### What does this PR do?
* typefix of platform.ts -> isCI from `string|boolean` to `boolean`

### Screenshot / video of UI
N/A

N/A

### How to test this PR?
N/A

### DIFF
--- a/tests/playwright/src/utility/platform.ts
+++ b/tests/playwright/src/utility/platform.ts
@@ -22,4 +22,4 @@ export const isLinux = os.platform() === 'linux';
 export const isMac = os.platform() === 'darwin';
 export const isWindows = os.platform() === 'win32';
 
-export const isCI = process.env.CI ? process.env.CI : false;
+export const isCI = process.env.CI ? process.env.CI === 'true' : false;


amisskii (approved):

tests/playwright/src/utility/platform.ts
```diff
@@ -22,4 +22,4 @@ export const isLinux = os.platform() === 'linux';
 export const isMac = os.platform() === 'darwin';
 export const isWindows = os.platform() === 'win32';
 
-export const isCI = process.env.CI ? process.env.CI : false;
+export const isCI = process.env.CI ? process.env.CI === 'true' : false;
```
cbr7, line 25:
```suggestion
export const isCI = process.env.CI ? !!process.env.CI : false;
```

Wouldn't this be better?

ScrewTSW: this is my first time seeing double exclamation point :D so maybe?

cbr7: double exclamation returns a boolean representation of whatever it's added to it, it's basically a symbolic way to valuate if something is "Truthy"

ScrewTSW: in that case it would set to `true` if any value is passed to `process.env.CI` even if it's value is `'false'`, wouldn't it?

cbr7: no, because false is falsy.

ScrewTSW: Screenshot_20240924_100136
Screenshot_20240924_100512
Screenshot_20240924_100612

odockal (approved):
Approving as it is.

PR #9037, opened 2024-09-24
fix: add missing aria-label attributes

dgolovin:
### What does this PR do?

add missing aria-label attributes to avoid linter errors after migration to next version of svelte see (#9033 PR from dependabot)

Related to #9033.

### How to test this PR?


benoitf (approved):
so the reason is that a button without a name should have an aria-label

PR #9075, opened 2024-09-25
fix: call fileItem onChange only when there is actually a change

lstocchi:
### What does this PR do?

This PR updates how the fileItem onChange function is called. Before there was a reactive clause which was triggered even when the component got loaded, which was wrong because there was no need to save the value (it was given by the parent). Now it behaves more naturally, when the on:change gets triggered, the onChange gets called. This prevents from having a first useless call which caused some misbehavior when working on the create new page with hyperv (following https://github.com/containers/podman-desktop/pull/8985)

### Screenshot / video of UI

From a user perspective nothing changes
N/A

it is part of the hyperv/wsl work

### How to test this PR?

1. run tests

### DIFF
--- a/packages/renderer/src/lib/ui/FileInput.svelte
+++ b/packages/renderer/src/lib/ui/FileInput.svelte
@@ -19,6 +19,11 @@ async function openDialog() {
     onChange(value);
   }
 }
+
+function onInput(event: Event): void {
+  const inputEvent = event as Event & { target: HTMLInputElement };
+  onChange(inputEvent.target.value);
+}
 </script>
 
 <div class="flex flex-row grow space-x-1.5">
@@ -27,7 +32,7 @@ async function openDialog() {
     name={name}
     class={$$props.class || ''}
     bind:value={value}
-    on:input
+    on:input={onInput}
     on:keypress
     placeholder={placeholder}
     readonly={readonly}
--- a/packages/renderer/src/lib/ui/FileInput.svelte
+++ b/packages/renderer/src/lib/ui/FileInput.svelte
@@ -2,7 +2,6 @@
 import { faFolderOpen } from '@fortawesome/free-solid-svg-icons';
 import type { OpenDialogOptions } from '@podman-desktop/api';
 import { Button, Input } from '@podman-desktop/ui-svelte';
-import { createEventDispatcher } from 'svelte';
 
 export let placeholder: string | undefined = undefined;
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/ui/FileInput.svelte
```diff
@@ -11,10 +12,13 @@ export let options: OpenDialogOptions;
 export let readonly: boolean = false;
 export let required: boolean = false;
 
+const dispatch = createEventDispatcher<{ change: string }>();
```
benoitf: as event dispatcher are deprecated in svelte5, is it possible to use an optional callback method rather than introducing an event dispatcher ?

benoitf: https://svelte-5-preview.vercel.app/docs/event-handlers

lstocchi: i didn't know... updated 👍

benoitf (approved):

jeffmaury (approved):
LGTM.

Just a remark there is a lint rule that is forcing every function and lambda to declare the return type thus you can have lambda in Svelte so it is better to have a function but it seems the rule is not yet enabled.

Also seems typecheck are failing but that seems related to @gastoner PR

lstocchi:
> Just a remark there is a lint rule that is forcing every function and lambda to declare the return type thus you can have lambda in Svelte so it is better to have a function but it seems the rule is not yet enabled.

Extracted the lambda 👍

jeffmaury (approved):
LGTM again

lstocchi:
There is actually a problem bc the fileinput could be readonly and editable. 
This new way only supports the readonly (the file has to be picked from the file system). However there are other scenario, like the kubeconfig file input which can be edited manually.

However if you add an empty kubeconfig there are errors on the kubernetes client and kube-context extensions as they do not have any check against an empty kubeconfig path entry. So trying to fix everything

lstocchi:
@benoitf @jeffmaury sorry but there was an issue and i needed to add another handler for the input. Be aware that there is a bug on the kubeconfig setting so if you want to try the change on that field, do not empty it or your desktop will stop responding -> https://github.com/containers/podman-desktop/issues/9101

benoitf (approved):

jeffmaury (approved):
LGTM

PR #9087, opened 2024-09-25
feat: hide Kubernetes menu when kubernetes empty page is displayed

### LINKED ISSUE #8571: Default actions for Kubernetes screens and their empty state

feloy:

### What does this PR do?

Hide Kubernetes menu when kubernetes empty page is displayed

### Screenshot / video of UI

https://github.com/user-attachments/assets/8d130a9a-05e9-4b1e-a2b9-0d6512ec3ac8

Part of #8571 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/stores/navigation/navigation-registry-kubernetes.svelte.spec.ts
+++ b/packages/renderer/src/stores/navigation/navigation-registry-kubernetes.svelte.spec.ts
@@ -17,20 +17,24 @@
  ***********************************************************************/
 
 import type { KubernetesObject } from '@kubernetes/client-node';
+import { readable } from 'svelte/store';
 import { beforeEach, expect, test, vi } from 'vitest';
 
 import KubeIcon from '/@/lib/images/KubeIcon.svelte';
+import { type ContextGeneralState, NO_CURRENT_CONTEXT_ERROR } from '/@api/kubernetes-contexts-states';
 
+import * as kubeContextStore from '../kubernetes-contexts-state';
 import { createNavigationKubernetesGroup } from './navigation-registry-kubernetes.svelte';
+
+vi.mock('../kubernetes-contexts-state', async () => {
+  return {};
+});
 
 beforeEach(() => {
   vi.resetAllMocks();
-  (window as any).kubernetesRegisterGetCurrentContextResources = kubernetesRegisterGetCurrentContextResourcesMock;
 });
 
-const kubernetesRegisterGetCurrentContextResourcesMock = vi.fn();
-
-test('createNavigationImageEntry', async () => {
+test('createNavigationImageEntry with current context', async () => {
   const nodes: KubernetesObject[] = [
     {
       apiVersion: 'v1',
@@ -47,7 +51,15 @@ test('createNavigationImageEntry', async () => {
       },
     },
   ];
-  kubernetesRegisterGetCurrentContextResourcesMock.mockResolvedValue(nodes);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextState = readable<ContextGeneralState>({} as ContextGeneralState);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextNodes = readable<KubernetesObject[]>(nodes);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextDeployments = readable<KubernetesObject[]>([]);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextServices = readable<KubernetesObject[]>([]);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextIngresses = readable<KubernetesObject[]>([]);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextRoutes = readable<KubernetesObject[]>([]);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextConfigMaps = readable<KubernetesObject[]>([]);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextSecrets = readable<KubernetesObject[]>([]);
+  vi.mocked(kubeContextStore).kubernetesCurrentContextPersistentVolumeClaims = readable<KubernetesObject[]>([]);
 
   const entry = createNavigationKubernetesGroup();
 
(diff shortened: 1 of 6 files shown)


packages/renderer/src/stores/navigation/navigation-registry-kubernetes.svelte.ts
```diff
@@ -28,6 +32,8 @@ import type { NavigationRegistryEntry } from './navigation-registry';
 
 let kubernetesNavigationGroupItems: NavigationRegistryEntry[] = $state([]);
 
+export let unsubscribe: Unsubscriber;
```
axel7083: never read ? only written

feloy: called on unit tests (afterEach), to be able to call the function several times without conflicts

axel7083: This feel weird :thinking:

feloy: Yes. The pattern used in the files in `packages/renderer/src/stores/navigation` does not expect to unsubscribe all the subscriptions (and because there is always 1 test for each file there, the problem was not raised before).

If we want to use a cleaner pattern (using a class extending Disposable for example), this will need to refactor all the files in this directory

axel7083: Thanks for the explanation. This could be a follow up PR, could you open the issue ?

packages/renderer/src/stores/navigation/navigation-registry-kubernetes.svelte.spec.ts
```diff
@@ -17,20 +17,28 @@
...
 import KubeIcon from '/@/lib/images/KubeIcon.svelte';
+import { NO_CURRENT_CONTEXT_ERROR } from '/@api/kubernetes-contexts-states';
 
-import { createNavigationKubernetesGroup } from './navigation-registry-kubernetes.svelte';
+import { createNavigationKubernetesGroup, unsubscribe } from './navigation-registry-kubernetes.svelte';
+
+const kubernetesRegisterGetCurrentContextResourcesMock = vi.fn();
+const kubernetesGetCurrentContextGeneralStateMock = vi.fn();
+
+(window as any).kubernetesRegisterGetCurrentContextResources = kubernetesRegisterGetCurrentContextResourcesMock;
```
axel7083: why not using `vi.mocked(window.kubernetesRegisterGetCurrentContextResourcesMock)` and inside the beforeEach do `(window as any).kubernetesRegisterGetCurrentContextResources = vi.fn()`

axel7083 (commented):
Tested locally and working great, the page update itself when a new cluster is added in the background.

packages/renderer/src/stores/navigation/navigation-registry-kubernetes.svelte.ts
feloy: -> https://github.com/containers/podman-desktop/issues/9089

packages/renderer/src/stores/navigation/navigation-registry-kubernetes.svelte.spec.ts
feloy: to keep the same pattern used into this file (I just moved this line from inside the `beforeEach`.

packages/renderer/src/stores/navigation/navigation-registry-kubernetes.svelte.ts
benoitf: not sure I follow everything but we can't add unsubscribe in the onDestroy() method instead of calling it in afterEach in the test (and having to export it) ?

benoitf: I'm afraid we're in the blurry part as we are in .ts file which are fragment of code, etc.

benoitf (changes requested):
a workaround for tests should not bring stuff in the renderer/production files

benoitf (commented):
from my understanding createNavigationKubernetesGroup() is using a state with the 
items to be displayed
```
let kubernetesNavigationGroupItems: NavigationRegistryEntry[] = $state([]);
```

this is the list you want to filter

so we need a derived pattern  `$derived` and get mainGroupEntry will return the derived rune.

and you subscribe to kubernetesCurrentContextState for updating the state

I don't know why then we're not using derived, state/effect and why we would need to have disposable methods. We only need reactivity/derived store from what I see

feloy:
I changed the implementation of the tests, by mocking the stores, so there is no more conflicts on stores between tests

benoitf (commented):
I still don't understand why we're using

```
kubernetesCurrentContextState.subscribe(value => {
    kubernetesNavigationGroupItems = value.error === NO_CURRENT_CONTEXT_ERROR ? [] : newItems;
  });
```

and not the $derived() pattern which looks designed for it as kubernetesNavigationGroupItems is a state

benoitf (commented):
the other problem that I see is that subscribing to a model is leaking data as we only had states in these files.

it's like we're trying to put svelte4 stuff in a svelte5 runes only.

close to the kubernetesCurrentContextState, I would add a state object being the contextStateError state

then here, you would just read that state in a $derived

feloy:
@benoitf  I'm not sure to follow

I'm trying to follow the same logic as in the other similar files in the same directory, for example `navigation-registry-k8s-deployments.svelte.ts`.

Is what I'm doing different?
```
let count = $state(0);

export function createNavigationKubernetesDeploymentsEntry(): NavigationRegistryEntry {
  kubernetesCurrentContextDeployments.subscribe(nodes => {
    count = nodes.length;
  });
  const registry: NavigationRegistryEntry = {
    name: 'Deployments',
    icon: { iconComponent: DeploymentIcon },
    link: '/kubernetes/deployments',
    tooltip: 'Deployments',
    type: 'entry',
    get counter() {
      return count;
    },
  };
  return registry;
}
```

benoitf:
@feloy: you're using `kubernetesCurrentContextState.subscribe` which is a store subscriber pattern (v4 approach)

$state/$props/$derived are the runes approach of svelte5

and in the .ts file we can't use the reactive $storeName (only limited to .svelte file)

so you subscribe to the store, but then, there is a leak as we never unsubscribe.

so in the current code, we should never try to use svelte4 pattern ( `store.subscribe`)

and just import a state object being updated when we receive some events

benoitf (commented):
https://svelte.dev/blog/runes#beyond-components

benoitf (approved):
thanks @feloy

feloy:
@axel7083 after discussion with Florent, I have updated this PR. Feel free to have another look. Thanks :)

axel7083 (approved):
LGTM

PR #9090, opened 2024-09-25
chore: reduce size of kubernetes navigation

### LINKED ISSUE #8771: UX design: Resize the podman-desktop navigation menu (preferences)

Following some discussing on the UX call, and https://github.com/containers/podman-desktop/pull/8754, we noticed that the navigation menu components are taking a lot of spaces. 

> [!NOTE]
> The navigation menu takes ~1/5 of the podman-desktop screen. (on a 1050px podman-desktop window, the menu takes 225 px => ~1/5) 

With the generalization of this sidebar in Kubernetes we are facing some space issue

image

a concern from @cdrage was the readability of the Kubernetes objects with the navigation menu taking 1/5 of the space 

While discussing this subject we took the example of OpenLens, we noticed they are using a way smaller navigation menu, and is **resizable**

https://github.com/user-attachments/assets/aacf27b5-0bc8-4649-a6a2-f4deae1edb01

_No response_

### Additional context

_No response_

cdrage:
chore: reduce size of kubernetes navigation

### What does this PR do?

* Reduces the size of the navigation menu to 'mini' version that's passed
into SubmenuNavigation.
* Adds "mini" version which is just a reduced width version of the
  submenu

Font improvements in the future (if needed) for the "mini" version which
may come in handy with other submenu components in the future that need
that whitespace.

### Screenshot / video of UI

Before:
Screenshot 2024-09-25 at 10 12 31 AM

After:
Screenshot 2024-09-25 at 10 02 10 AM

Closes https://github.com/containers/podman-desktop/issues/8771

### How to test this PR?

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -117,7 +117,12 @@ window.events?.receive('navigate', (navigationRequest: unknown) => {
       {/if}
       {#each $navigationRegistry.filter(item => item.type === 'submenu') as navigationRegistryItem}
         {#if meta.url.startsWith(navigationRegistryItem.link)}
-          <SubmenuNavigation meta={meta} title={navigationRegistryItem.tooltip} items={navigationRegistryItem.items} />
+          <!-- If the meta url starts with "kubernetes" we will pass mini=true to the submenu navigation to reduce the size of it for less whitespace -->
+          <SubmenuNavigation
+            meta={meta}
+            title={navigationRegistryItem.tooltip}
+            items={navigationRegistryItem.items}
+            mini={meta.url.startsWith('/kubernetes')} />
         {/if}
       {/each}
 
--- a/packages/renderer/src/SubmenuNavigation.spec.ts
+++ b/packages/renderer/src/SubmenuNavigation.spec.ts
@@ -16,6 +16,8 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import '@testing-library/jest-dom';
+
 import { SettingsNavItem } from '@podman-desktop/ui-svelte';
 import { render, screen } from '@testing-library/svelte';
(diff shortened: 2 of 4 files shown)


benoitf (changes requested):
This issue was not planned for this sprint. (It is not planned until 1.14)
UX team needs to provide first the proper design, it's still under exploration.

cdrage:
> This issue was not planned for this sprint. (It is not planned until 1.14) UX team needs to provide first the proper design, it's still under exploration.

No problem, I'll close for now, was exploring what we should do. 

Issue https://github.com/containers/podman-desktop/issues/8771 was in backlog for this current sprint and not marked for 1.14 so thought to bring it up.

Should we mark it 1.14 and switch to next sprint?

benoitf:
@cdrage anyway thanks for the exploration but yes AFAIK it was not planned, it's not yet 1.14 as the goals are not yet fully fixed

Also it'll need to work with AI Lab that is using the same layout.

cdrage:
> @cdrage anyway thanks for the exploration but yes AFAIK it was not planned, it's not yet 1.14 as the goals are not yet fully fixed
> 
> Also it'll need to work with AI Lab that is using the same layout.

Yup I agree, they will be using the same layout with the submenu, so it'll be good to have that as a "mini" boolean option.

Either way, we'll coordinate with @ekidneyrh

ekidneyrh:
@cdrage I dont like that the categories spill over onto two lines instead of one. A reduction on the font size, a step down of what it is atm, would work and then have the padding be equal on the right and left of the categories - it'll matter most on the longest category. This will make it not look as squashed and be more breathable. 

We could alternatively have an icon assigned to each category and have an option to only show the icons in the mini menu if the user wishes for more room. This could be something we could bring to AI Lab too if the menu concern is also there @axel7083 

Or a collapsible menu that the user can hide?

PR #9096, opened 2024-09-26
feat: add editor to provide uri or fully qualified registry for image to use for VM

### LINKED ISSUE #8109: Create podman machine should allow image URL not only local file path

dgolovin:
### What does this PR do?

PR introduces new field to provide URL or fully qualified registry for VM image when creating podman VM.
1. When both fields filled, the local file takes priority, URL/Registry ignored.
2. Validation is on podman, if incorrect url or registry entered podman machine creation fails and error is shown on top of the page

### Screenshot / video of UI

Generic placeholder shown for image reference

image

Properties audit reports error when path and URL/image reference filled at the same time. `Create` button is disabled when errors are reported.

image

Fixes #8109.

### How to test this PR?

1. Pull the PR
2. pnpm install && pnpm watch
3. Open Setting/Resources page
4. Push 'Create new...' for Podman
5. Use URL to VM from any of podman releases on github or `docker://quay.io/podman/machine-os:5.2` or any other available on quay version


benoitf:
I think we should maybe drop the requirement on `docker://` prefix in the UI because I think users won't really provide it at first, meaning that we can reduce the text description

benoitf:
> Validation is on podman, if incorrect url or registry entered podman machine creation fails and error is shown on top of the page

we can validate the field on the extension side, so I would report an error if user enter both fields (image path and image URL). so user need to remove one option

And for the URL, I would expect also some validation in the UI directly as it's avoiding to see error later (here could be follow-up PR)

gastoner:
What about change the text to "Image URL or fully qualified registry (Optional)" and move the `docker://registry/repo/image:version` to the inputbox? like a template similar to: 
image
When starting to type, only written text will be visible
Same input box just with template (instead of "Search Container..." it would be `docker://registry/repo/image:version` )

dgolovin:
> What about change the text to "Image URL or fully qualified registry (Optional)" and move the `docker://registry/repo/image:version` to the inputbox? like a template similar to: image When starting to type, only written text will be visible Same input box just with template (instead of "Search Container..." it would be `docker://registry/repo/image:version` )

There should be some text  in addition to example of an input, because otherwise it would look like default value. We can skip mention of URL in placeholder and just say how to use image reference. 

image

@gastoner @benoitf  WDYT?

benoitf:
I'm fine with this proposal

dgolovin:
Description for Image URI updated and placeholder set to 'Use ...'.
Check added for for the case when both fields are filled. I used just throw new Error from create connection method, because auditItems approach does not disable `Create` button and `podman init` executed anyway (see #9118).

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -1972,6 +1976,12 @@ export async function createMachine(
     parameters.push('--image-path');
     parameters.push(params['podman.factory.machine.image-path']);
     telemetryRecords.imagePath = 'custom';
+  } else if (params['podman.factory.machine.image-uri']) {
+    parameters.push('--image-path');
+    parameters.push(params['podman.factory.machine.image-uri']);
```
dgolovin:
```suggestion
    parameters.push(`docker://${params['podman.factory.machine.image-uri']}`);
```

extensions/podman/packages/extension/src/extension.spec.ts
```diff
@@ -350,6 +350,137 @@ test('verify create command called with correct values', async () => {
...
+      'machine',
+      'init',
+      '--cpus',
+      '2',
+      '--memory',
+      '999',
+      '--disk-size',
+      '232',
+      '--image-path',
+      'registry/repo/image:version',
```
dgolovin:
```suggestion
      'docker://registry/repo/image:version',
```

extensions/podman/packages/extension/package.json
```diff
@@ -149,6 +149,12 @@
           "scope": "ContainerProviderConnectionFactory",
           "description": "Image Path (Optional)"
         },
+        "podman.factory.machine.image-uri": {
+          "type": "string",
+          "scope": "ContainerProviderConnectionFactory",
+          "description": "Image URL or image reference (Optional)",
+          "placeholder": "Use 'quay.io/podman/machine-os:version' for image reference"
```
benoitf: it seems that for Windows it'll be `https://quay.io/repository/podman/machine-os-wsl` so I don't know if we need like two properties, one being activated on windows and the others on macOS

I don't know also if the placeholder link is working for Linux users

dgolovin: I did not mean to point to specific image. I can change placepolder to `Use 'registry/org/image:version' for image reference` to avoid confusion.

The explanation how to use the URI field is not something we can fit in UI. We can consider add page in docs on podman-desktop.io, change description to markupDescription and have a link to documentation there.

dgolovin: It seems there is only tag 5.3 for https://quay.io/repository/podman/machine-os-wsl?tab=tags, but current release is v5.2.3.

dgolovin: Should we just let it to be provided by user and put `Use 'registry/org/image:version' for image reference` as placeholder for upcoming release?

benoitf: yes let's keep it very generic

dgolovin:
Added audit implementation, to fail fast without calling podman cli.

dgolovin:
Tests should be updated to reflect usage of properties auditor instead of verifying that exception is thrown when bot path and uri for image are present.

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -2010,6 +2018,16 @@ export async function createMachine(
...
     telemetryRecords.imagePath = 'custom';
+  } else if (params['podman.factory.machine.image-uri']) {
+    const imageUri = params['podman.factory.machine.image-uri'].trim();
+    parameters.push('--image-path');
+    if (imageUri.startsWith('https://') || imageUri.startsWith('http://')) {
+      parameters.push(imageUri);
+      telemetryRecords.imagePath = 'custom-url';
+    } else {
+      parameters.push(`docker://${imageUri}`);
+      telemetryRecords.imagePath = 'custom-registry';
```
benoitf, line 2029: should we send the hash of the URL/images ?

benoitf (approved):

gastoner:
Probably not related to this PR but I get `no valid disk artifact found` when trying to specify any image

extensions/podman/packages/extension/src/extension.ts
dgolovin: I opened #9178 to discuss what we can send in telemetry when custom image is used.

dgolovin:
> Probably not related to this PR but I get `no valid disk artifact found` when trying to specify any image

I saw this when I was using random images to test UI. It comes form podman when image does not contain podman VM.

PR #9116, opened 2024-09-27
chore: add warning for users about short image names

### LINKED ISSUE #8817: Pulling short names should be consistent between Podman CLI and Pull Image form
### Bug description

If I pull a short name from the command line, I get:
```
% podman pull fedora
Resolved "fedora" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull registry.fedoraproject.org/fedora:latest...
```
but if I go to Images > Pull Image and type in `fedora`, I get `docker.io/library/fedora`.

I'm not sure where it is lost in translation, but we should pull the same image in these cases.

### Operating system

macOS 14.5

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
Since Podman Desktop uses the Dockerode API (Docker based) to pull images, along with Podman's Docker compatibility, when pulling images with their short names, the images are pulled from Docker Hub instead of following the `shortnames.conf` file as some users might expect based on Podman CLI behavior. 
This PR add a visible warning to users to let them know that shortname images will be pulled from Docker Hub

### Screenshot / video of UI
[Screencast from 2024-10-01 16-52-08.webm](https://github.com/user-attachments/assets/4b34e44a-ce2a-419d-a4cc-cce637cd4770)

Closes https://github.com/containers/podman-desktop/issues/8817

### How to test this PR?

Input a shortname for an image to pull, assert that if you check the Podman FQN checkbox, it pulls the image based on `shortname.conf`, and without checking the checkbox, it pulls the image from docker hub

### DIFF
--- a/packages/renderer/src/lib/image/PullImage.svelte
+++ b/packages/renderer/src/lib/image/PullImage.svelte
@@ -121,8 +121,10 @@ async function pullImage() {
 
   pullInProgress = true;
   try {
-    if (usePodmanFQN && podmanFQN) {
-      await window.pullImage(selectedProviderConnection, podmanFQN.trim(), callback);
+    if (podmanFQN) {
+      usePodmanFQN
+        ? await window.pullImage(selectedProviderConnection, podmanFQN.trim(), callback)
+        : await window.pullImage(selectedProviderConnection, `docker.io/${imageToPull.trim()}`, callback);
     } else {
       await window.pullImage(selectedProviderConnection, imageToPull.trim(), callback);
     }
--- a/packages/renderer/src/lib/image/PullImage.svelte
+++ b/packages/renderer/src/lib/image/PullImage.svelte
@@ -39,23 +39,24 @@ const lineNumberPerId = new Map<string, number>();
 let lineIndex = 0;
 
 async function resolveShortname(): Promise<void> {
-  if (selectedProviderConnection && selectedProviderConnection.type === 'podman') {
-    if (imageToPull && !imageToPull.includes('/')) {
-      shortnameImages = (await window.resolveShortnameImage(selectedProviderConnection, imageToPull)) ?? [];
-      // not a shortname
-    } else {
-      podmanFQN = '';
-      shortnameImages = [];
-      usePodmanFQN = false;
-    }
-    // checks if there is no FQN that is from dokcer hub
-    if (!shortnameImages.find(name => name.includes('docker.io'))) {
-      podmanFQN = shortnameImages[0];
-    } else {
-      podmanFQN = '';
-      shortnameImages = [];
-      usePodmanFQN = false;
-    }
+  if (!selectedProviderConnection || selectedProviderConnection.type !== 'podman') {
+    return;
+  }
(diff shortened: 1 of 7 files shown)


SoniaSandler:
@benoitf @deboer-tim these are some of the possible options I have to let the users know about the expected behavior for this scenario. With those warnings, we could add the suggested link to more information about `Docker compatibility` (not sure which specific document).
 lmk what you think about this (which option is better) and if there is something to add, modify, or if you have any other ideas.

benoitf:
on my end, I prefer something like option 2 as it's not interrupting my workflow.

Then for 99.99% of the images being pulled, the resolution is the same (will use docker.io prefix). So displaying a warning all the time to cover this 0.01% might be too much.

Using the podman API, we know how to resolve an image name into a FQN (there is an API for that)

so maybe, if the engine being used is podman, we could have a small checkbox (use shortname resolution, can be unchecked by default)
or we could display the warning in case the shortname resolution is different than the 'dockerhub' resolution.

so if I enter 'httpd', no warning as it'll pull docker.io/httpd but if I enter 'fedora' I see the warning telling me that it would resolve to docker.io/fedora but I can enforce the usage of shortname resolution to use `registry.fedoraproject.org/fedora` instead

the only issue that I see is how to avoid to have the layout moving


in short: maybe a conditional warning rather than an always on warning (b/c why should I see the warning if I enter a FQN like quay.io/podman/hello)

deboer-tim:
Same thoughts here: it should not interrupt workflow, ideally you only see any new UI when we know you're using a short name, and it doesn't cause the UI to shift.

I'm pretty sure the design has form buttons like Pull Image right-justified. Maybe this is an opportunity to 'fix' this page, which would open up space on the left for any new warning or checkbox to appear?

benoitf (approved):
wondering if the name of the image in case it's using podman resolver could be displayed

image

As we have this information maybe we could show it.

`Image will be resolved to index.docker.io while you might expect registry.fedoraproject.org/fedora`

or with the hint into the checkbox

Using the podman resolver will resolve the image to registry.fedoraproject.org/fedora

can be follow-up PR

packages/renderer/src/lib/image/PullImage.svelte
```diff
@@ -34,6 +38,27 @@ let selectedProviderConnection: ProviderContainerConnectionInfo | undefined;
 const lineNumberPerId = new Map<string, number>();
 let lineIndex = 0;
 
+async function resolveShortname(): Promise<void> {
+  if (selectedProviderConnection && selectedProviderConnection.type === 'podman') {
```
axel7083: Instead of having lots of if inside if statement, you could do return early pattern, IMHO it ease the reading

```ts
if (!selectedProviderConnection || selectedProviderConnection.type !== 'podman')
    return

// your logic
```

axel7083 (approved):
Tested locally and works fine, LGTM 🚀

benoitf:
to be merged post 1.13

benoitf:
1.13 is out so it can be merged

PR #9119, opened 2024-09-27
fix: disable `Create` button if properties audit returns errors

### LINKED ISSUE #9118: If `auditItems` function returns errror 'Create' button stays enabled on create resource page
### Bug description

An extension can provide auditItems function when setting up container connection factory. auditItems function can return records reportin problems with new container connection properties. The `Create` button meantime stays enabled and user can proceed with creating container connection.

Would we good to diable `Create` button in case when connection properties audit returns audit records with type `error`.

### Operating system

macOS M1 Sonoma 14.5

### Installation Method

None

### Version

1.11.1

### Steps to reproduce

If Kind cluster with the same name is created there should be an error on top of the new kind cluster form, but 'Create' button should still be available.

### Relevant log output

_No response_

### Additional context

Relates to #9096.

dgolovin:
### What does this PR do?

PR disables 'Create' button in new resource form if audit returned record with type 'error'.

### Screenshot / video of UI

https://github.com/user-attachments/assets/c82dcf6b-f317-4499-99c5-51459472137e

Fix #9118

### How to test this PR?

Try to create kind cluster with the same name

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
@@ -216,7 +216,7 @@ async function handleValidComponent() {
 
   try {
     const auditResult = await window.auditConnectionParameters(providerInfo.internalId, data as AuditRequestItems);
-    isValid = auditResult?.records?.filter(record => record.type === 'error')?.length === 0;
+    isValid = auditResult.records.filter(record => record.type === 'error').length === 0;
     connectionAuditResult = auditResult;
   } catch (err: unknown) {
     if (err instanceof Error) {
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
@@ -223,6 +223,8 @@ async function handleValidComponent() {
       console.warn(String(err));
     }
   }
+  const auditErrors = connectionAuditResult?.records?.filter(record => record.type === 'error');
+  isValid = auditErrors !== undefined && auditErrors.length === 0;
 }
 
 function internalSetConfigurationValue(id: string, modified: boolean, value: string | boolean | number) {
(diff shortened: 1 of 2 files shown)


dgolovin:
Test have not been inmplemented yet.

packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
```diff
@@ -223,6 +223,8 @@ async function handleValidComponent() {
       console.warn(String(err));
     }
   }
+  const auditErrors = connectionAuditResult?.records?.filter(record => record.type === 'error');
+  isValid = auditErrors !== undefined && auditErrors.length === 0;
```
benoitf: would it work using only `audit errors?.length === 0`

benoitf (approved):

feloy:
I'm probably doing things wrong, I cannot find a way to test this.

For example, when I create a podman machine or a Kind cluster, I always get as result of `window.auditConnectionParameters` an empty array of `records`

dgolovin:
> I'm probably doing things wrong, I cannot find a way to test this.
> 
> For example, when I create a podman machine or a Kind cluster, I always get as result of `window.auditConnectionParameters` an empty array of `records`

Actually, you partially right. There is no way to trigger 'error' on create Kind cluster page. You can trigger 'info' and 'warning' but, that would not disable Create/Update button.

There is no easy way to verify it now, because there is no audit implementation in main branch that would report error.

I'll update PR #9096 to use audit, so you can pull this PR, then pull #9096 with `--rebase` option and test it.

feloy (approved):
LGTM based on unit tests

axel7083:
> Try to create kind cluster with the same name

Here are my existing kind clusters

image

But the `Create` button seems to still be active

image

After clicking on it I am getting the error (expected)

image

dgolovin:
> > Try to create kind cluster with the same name
>
> But the `Create` button seems to still be active 
> 
> After clicking on it I am getting the error (expected)
> 

@axel7083 Sorry, it turned out there is no check for same name in Kind extension. I thought I saw it when I was looking for examples of audit implementations, but after @feloy comment above I checked again ad realized there is no audit for it. I opened follow up for it #9152.

PR #9122, opened 2024-09-27
refactor: move dispatch functions to ContextsStates class

feloy:

### What does this PR do?

refactor: move dispatch functions to ContextsStates class

Part of #8775 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-states.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-states.ts
@@ -277,7 +277,7 @@ export class ContextsStates {
     }
   }
 
-  async disposeContext(name: string): Promise<void> {
+  async deleteContextState(name: string): Promise<void> {
     this.state.delete(name);
   }
 
--- a/packages/main/src/plugin/kubernetes/contexts-states.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-states.ts
@@ -93,6 +93,26 @@ export function isSecondaryResourceName(value: string): value is SecondaryResour
   return secondaryResources.includes(value as SecondaryResourceName);
 }
 
+/**
+ * ContextsStates is responsible for sending through apiSender or retrieve the state of the different Kubernetes contexts.
+ *
+ * The caller can use:
+ * - `setStateAndDispatch` to update the state of a specific context and dispatch it,
+ * - `dispatch` to dispatch the current state of a specific context.
+ *
+ * Different events are sent, depending on the options passed to these methods:
+ * - `kubernetes-contexts-checking-state-update`: A map containing a CheckingState value for each Kubernetes context
+ * - `kubernetes-contexts-general-state-update`: A map containing a ContextGeneralState (checking state, connectivity error, is the context reachable, number of primary resources - pods and deployments)
+ *   for each Kubernetes context
+ * - `kubernetes-current-context-general-state-update`: the ContextGeneralState for the current context only
+ * - `kubernetes-current-context-${resname}-update` (resname in pods, deployments, services, routes, etc): the details of the Kubernetes resources of a specific kind for the current context only
+ *
+ * Methods are available to get these information synchronously:
+ * - `getContextsCheckingState`: data in event `kubernetes-contexts-checking-state-update`
+ * - `getContextsGeneralState`: data in event `kubernetes-contexts-general-state-update`
+ * - `getCurrentContextGeneralState`: data in event `kubernetes-current-context-general-state-update`
+ * - `getContextResources`: data in event `kubernetes-current-context-${resname}-update`
+ */
 export class ContextsStates {
   private state = new Map<string, ContextState>();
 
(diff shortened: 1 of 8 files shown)


packages/main/src/plugin/kubernetes/kubernetes-context-state.ts
```diff
@@ -52,21 +52,20 @@ import {
 } from '@kubernetes/client-node';
```
axel7083, line 52: I noticed we have this file named `kubernetes-context-state` and another in the same folder `context-state`, is this expected ?

packages/main/src/plugin/kubernetes/contexts-states.ts
```diff
@@ -135,7 +201,83 @@ export class ContextsStates {
...
+  dispose(): void {
+    this.disposed = true;
+    clearTimeout(this.dispatchContextsGeneralStateTimer);
+    clearTimeout(this.dispatchCurrentContextGeneralStateTimer);
+    for (const timer of this.dispatchCurrentContextResourceTimers.values()) {
+      clearTimeout(timer);
+    }
+  }
+
+  async disposeContext(name: string): Promise<void> {
```
axel7083: `disposeContext` is not called when `dispose()` is called ?

packages/main/src/plugin/kubernetes/contexts-states.ts
```diff
@@ -135,7 +201,83 @@ export class ContextsStates {
     update(val);
   }
 
-  async dispose(name: string): Promise<void> {
+  dispatch(options: DispatchOptions): void {
```
axel7083, line 225: Could we have a few line of documentation above, TBH I am not sure to understand what this is doing :'(

packages/main/src/plugin/kubernetes/fake-informer.ts
```diff
@@ -0,0 +1,103 @@
+/**********************************************************************
```
axel7083: Should this class use `.spec` ? Can we import stuff from one `.spec` file to another `.spec` ?

If we can, we should probably do something like `test-informer.spec.ts`

If we cannot I think we should use `informer-test.ts` rather than `fake`

We use the terminology `Test*` for classes often 

https://github.com/containers/podman-desktop/blob/66734cee5ce7677d84e2565e3da34231a201b96f/packages/renderer/src/lib/pod/KubernetesTerminalService.spec.ts#L28

https://github.com/containers/podman-desktop/blob/72e287a0f3992ebacfdc4893f73a5facd342d8ab/packages/main/src/plugin/color-registry.spec.ts#L36

https://github.com/containers/podman-desktop/blob/fba9eead7078174bf7806e89c5f48c60dd8d0c37/packages/main/src/plugin/directories.spec.ts#L26

packages/main/src/plugin/kubernetes/fake-informer.ts
```diff
@@ -0,0 +1,103 @@
...
+
+interface InformerErrorEvent {
+  delayMs: number;
+  verb: string;
+  error: string;
+}
+
+export const informerStopMock = vi.fn();
+
+export class FakeInformer {
```
axel7083: Some comment above as this is a dedicated file

packages/main/src/plugin/kubernetes/kubernetes-context-state.ts
feloy: I would like to work on renaming files/classes globally in the latest PR (https://github.com/containers/podman-desktop/issues/8775#issuecomment-2358182299)

packages/main/src/plugin/kubernetes/contexts-states.ts
feloy: `disposeContext` is probably badly named, as it is not disposing system resources, as `dispose` does, but only _deletes_ an entry from the map.

I'm pushing a new commit to rename disposeContext and ContextInformers.dispose with a better name IMO

packages/main/src/plugin/kubernetes/fake-informer.ts
feloy: > Should this class use `.spec` ? Can we import stuff from one `.spec` file to another `.spec` ?
> 

Sonarjs is not happy with: 
```
  0:1  error  Add some tests to this file or delete it  sonarjs/no-empty-test-file
```

feloy:
@axel7083 I have made some changes, and added more documentation

packages/main/src/plugin/kubernetes/fake-informer.ts
axel7083: I would still rather have a basic test, than having this without the `.spec` format :/

packages/main/src/plugin/kubernetes/kubernetes-context-state.ts
axel7083: 👍

axel7083 (approved):
LGTM

benoitf (approved):

PR #9125, opened 2024-09-27
chore: specify devices when creating a container

### LINKED ISSUE #5989: Add --device option when creating a container

Podman Desktop does not have the option to add a device like --device /dev/fuse when running the container from the UI.
My requirement does not allow me to use a terminal.

 Is it possible to do so inside the Podman Desktop? Or adding this option possibly in the "Advanced" tab?

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Specify devices when creating a container, equivalent of the `--device` option

A typeahead could help the user, it could be part of a follow-up PR

Fixes #5989

### How to test this PR?

Go to the Container creation form, in Advanced tab, and add devices (should exist in the podman machine), and check in the Inspect tab of the created container that the `HostConfig.Devices` info is present.

### DIFF
--- a/packages/api/src/container-info.ts
+++ b/packages/api/src/container-info.ts
@@ -89,7 +89,7 @@ interface MountSettings {
 
 type MountConfig = MountSettings[];
 
-export interface Device {
+export interface DeviceMapping {
   CgroupPermissions: string;
   PathInContainer: string;
   PathOnHost: string;
@@ -111,7 +111,7 @@ export interface HostConfig {
   ExtraHosts?: string[];
   NetworkMode?: string;
   Mounts?: MountConfig;
-  Devices?: Device[];
+  Devices?: DeviceMapping[];
 }
 
 export interface HealthConfig {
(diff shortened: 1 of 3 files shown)


deboer-tim (approved):
Does what it says on the tin. 

Nit: I noticed that the checkbox text was slightly bigger than others, seems everything else on this form is set to text-sm, please do that here too.

feloy:
> Does what it says on the tin.
> 
> Nit: I noticed that the checkbox text was slightly bigger than others, seems everything else on this form is set to text-sm, please do that here too.

Good catch, thanks. I've also added some space between checkboxes, to better distinguish which label is attached to which checkbox

axel7083:
> A typeahead could help the user, it could be part of a follow-up PR

We don't have a way to get easily and quickly the list of devices ? The only way I think about is something like `podman machine ssh ls /dev` and I do not really like it. (related to https://github.com/containers/podman/issues/24042)

packages/api/src/container-info.ts
```diff
@@ -89,6 +89,12 @@ interface MountSettings {
 
 type MountConfig = MountSettings[];
 
+export interface DeviceMapping {
```
axel7083, line 92: Need rebase

axel7083 (approved):
LGTM Tested with device `/dev/dxg` (gpu on windows)

PR #9141, opened 2024-09-30
chore(test): add some additonal e2e tests

### LINKED ISSUE #9081: Increase e2e test coverage for image search functionality

Currently the image search functionality is tested as part of a smoke test, we should have a separate test suite that covers more scenarios for image search.

Create a test suite with more test cases for more scenarios.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Adds additional e2e tests for image search functionality

https://github.com/containers/podman-desktop/issues/9081

### DIFF
--- a/tests/playwright/src/model/pages/pull-image-page.ts
+++ b/tests/playwright/src/model/pages/pull-image-page.ts
@@ -58,7 +58,12 @@ export class PullImagePage extends BasePage {
     });
   }
 
-  async getAllSearchResultsFor(imageName: string, searchForVersion: boolean, imageTag = ''): Promise<string[]> {
+  async getAllSearchResultsFor(
+    imageName: string,
+    searchForVersion: boolean,
+    imageTag = '',
+    resultsExpected = true,
+  ): Promise<string[]> {
     return await test.step(`Get all search results for ${imageName}:${imageTag}`, async () => {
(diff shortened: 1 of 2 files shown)


odockal:
I think that we should cover this scenarios, for inspiration:
1. (Offering and selection) Start typing the image name, let's consider `httpd`, If I type `htt`, I should see results already (check for >0). Then I finish `httpd` and I should see results, that for sure contains `docker.io/httpd`. If I select the image from the list, it should be filled in the input.
2. (Tag offering and selection) If I start typing image name, ie. again `httpd`, it should provide a list, I select `docker.io/httpd` which appears in the input. I then type `:` which now enableds the tag selection for particular image. List of tags is offered, and I can select on the the tags: ie: `docker.io/httpd:2-alpine`. 
3. Considering user registry added via Registries page. If such registry is chosen, it should cound with it in the image search and offers results on the top over ie. generic docker.io.
4. I think that other than `docker.io` registry results are not working, or working correctly, If I type `ghcr.io/` I got only images with such name under `docker.io` instead of using the registry itself. Only If I type `ghcr.io/linuxcontainers/alpine` I can then use the tag to see the offering. Not sure if intended behavior.

odockal (approved):

PR #9144, opened 2024-09-30
fix: add podman version check to hyperv enablement

lstocchi:
### What does this PR do?

After talking with @jeffmaury there was a missing piece in the hyperv check - the podman version should be greater than 5.2.0.
This PR just enhances the hyperv part to check if podman has a valid version

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

1. run tests or try to use an older version of podman to see hyperv is not enabled

### DIFF
--- a/extensions/podman/packages/extension/src/podman-install.ts
+++ b/extensions/podman/packages/extension/src/podman-install.ts
@@ -821,12 +821,12 @@ export class WSLVersionCheck extends BaseCheck {
 
 export class HyperVCheck extends WindowsCheck {
   title = 'Hyper-V installed';
-  PODMAN_MINIMUM_VERSION_FOR_HYPERV = '5.2.0';
+  static readonly PODMAN_MINIMUM_VERSION_FOR_HYPERV = '5.2.0';
 
   async execute(): Promise<extensionApi.CheckResult> {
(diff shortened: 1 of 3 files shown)


extensions/podman/packages/extension/src/podman-install.ts
```diff
@@ -821,8 +821,14 @@ export class WSLVersionCheck extends BaseCheck {
 
 export class HyperVCheck extends WindowsCheck {
   title = 'Hyper-V installed';
+  PODMAN_MINIMUM_VERSION_FOR_HYPERV = '5.2.0';
```
jeffmaury: Maybe add const ?

jeffmaury (approved):
LGTM

extensions/podman/packages/extension/src/podman-install.ts
benoitf: static readonly

benoitf (approved):
code ok but checked on macOS so partial check

extensions/podman/packages/extension/src/podman-install.ts
lstocchi: done 👍

PR #9146, opened 2024-09-30
feat: add Preference to ask before fetching image layers

### LINKED ISSUE #8212: Add a preference for starting the loading of image layers automatically

When a user enters the File tab of an Image (when the Layers explorer extension is installed), the process of loading the layers is started automatically. For very large images, this can be annoying for users discovering the tab.

In the other hand, users experienced with this tab will want to start the process ASAP, without having to click again.

Add a preference to start the process of loading the layers automatically or not (by default non automatically)

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

add Preference to ask before fetching image layers

### Screenshot / video of UI

https://github.com/user-attachments/assets/7f6ae46c-e10a-4245-a664-aa5cbe3bcd10

Fixes #8212 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
+++ b/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
@@ -25,61 +25,55 @@ let filesProvider: ImageFilesInfo | undefined = undefined;
 let cancellableTokenId: number = 0;
 let askFetchLayers: boolean = true;
 
-function onSelectedLayer(event: CustomEvent<ImageFilesystemLayerUI>) {
+function onSelectedLayer(event: CustomEvent<ImageFilesystemLayerUI>): void {
   selectedLayer = event.detail;
 }
 
-function fetchImageLayers(provider: ImageFilesInfo, img: ImageInfo) {
-  loading = true;
-  window.getCancellableTokenSource().then(token => {
-    cancellableTokenId = token;
-    window
-      .imageGetFilesystemLayers(provider.id, img, cancellableTokenId)
-      .then(layers => {
-        imageLayers = layers;
-      })
-      .catch((err: unknown) => {
-        error = String(err);
-      })
-      .finally(() => {
-        loading = false;
-      });
-  });
+async function fetchImageLayers(provider: ImageFilesInfo, img: ImageInfo): Promise<void> {
+  try {
+    loading = true;
+    cancellableTokenId = await window.getCancellableTokenSource();
+    const layers = await window.imageGetFilesystemLayers(provider.id, img, cancellableTokenId);
+    imageLayers = layers;
+  } catch (err: unknown) {
+    error = String(err);
+  } finally {
+    loading = false;
+  }
+}
+
+async function onFetchLayers(): Promise<void> {
+  showFetchButton = false;
+  if (filesProvider !== undefined && imageInfo !== undefined) {
+    await fetchImageLayers(filesProvider, imageInfo);
+  }
 }
 
 onMount(async () => {
-  window
-    .getConfigurationValue<boolean>('userConfirmation.fetchImageFiles')
-    .then(value => {
-      if (value !== undefined) {
-        askFetchLayers = value;
-      }
-    })
-    .finally(() => {
-      filesProvidersUnsubscribe = imageFilesProviders.subscribe(providers => {
-        if (providers.length === 1 && imageInfo) {
-          filesProvider = providers[0];
-          if (askFetchLayers) {
-            showFetchButton = true;
-          } else {
-            fetchImageLayers(filesProvider, imageInfo);
-          }
+  try {
+    const value = await window.getConfigurationValue<boolean>('userConfirmation.fetchImageFiles');
+    if (value !== undefined) {
+      askFetchLayers = value;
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -18,48 +18,76 @@ let selectedLayer = $state<ImageFilesystemLayerUI>();
...
+      });
+    });
 });
 
 onDestroy(() => {
   window.cancelToken(cancellableTokenId);
   filesProvidersUnsubscribe?.();
 });
+
+function onFetchLayers() {
```
axel7083: return type

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -18,48 +18,76 @@ let selectedLayer = $state<ImageFilesystemLayerUI>();
...
+        error = String(err);
+      })
+      .finally(() => {
+        loading = false;
       });
-    }
   });
+}
+
+onMount(async () => {
```
axel7083, line 51: no need to be async ?

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -18,48 +18,76 @@ let selectedLayer = $state<ImageFilesystemLayerUI>();
...
-          .then(layers => {
-            imageLayers = layers;
-          })
-          .catch((err: unknown) => {
-            error = String(err);
-          })
-          .finally(() => {
-            loading = false;
-          });
+function fetchImageLayers(provider: ImageFilesInfo, img: ImageInfo) {
```
axel7083: return type

benoitf: why this method is not an async ? (also what is the return type)

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -18,48 +18,76 @@ let selectedLayer = $state<ImageFilesystemLayerUI>();
...
+      .imageGetFilesystemLayers(provider.id, img, cancellableTokenId)
+      .then(layers => {
+        imageLayers = layers;
+      })
+      .catch((err: unknown) => {
+        error = String(err);
+      })
+      .finally(() => {
+        loading = false;
       });
```
benoitf: wondering why we're using all the then/then/then/ call vs the more readable pattern
try {
  await 1
  await 2
} finally {
  
}

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
feloy: effectively, it could be async. I'll change

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -18,48 +18,76 @@ let selectedLayer = $state<ImageFilesystemLayerUI>();
...
-    }
   });
+}
+
+onMount(async () => {
+  window
+    .getConfigurationValue<boolean>('userConfirmation.fetchImageFiles')
+    .then(value => {
```
benoitf: why are we not using await ?

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
feloy: yep, I'll change

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
feloy: same, I'll change

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
benoitf: I think it's better to use async/await (so I'll replace the promise.then pattern to use await promise)

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -18,36 +18,56 @@ let selectedLayer = $state<ImageFilesystemLayerUI>();
...
+function onSelectedLayer(event: CustomEvent<ImageFilesystemLayerUI>): void {
   selectedLayer = event.detail;
 }
 
+async function fetchImageLayers(provider: ImageFilesInfo, img: ImageInfo): Promise<void> {
+  try {
+    loading = true;
+    cancellableTokenId = await window.getCancellableTokenSource();
+    const layers = await window.imageGetFilesystemLayers(provider.id, img, cancellableTokenId);
+    imageLayers = layers;
```
benoitf: just wondering if we need to use this intermediate variable

benoitf (approved):

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
feloy: No, removing it

jeffmaury (approved):
LGTM

PR #9148, opened 2024-09-30
refactor: ImagesDetailsComponent to Svelte5

feloy:
### What does this PR do?

- Use `$props()` and `$state()`
- Remove unused slot

Part of #8212 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
+++ b/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
@@ -13,11 +13,11 @@ import ImageDetailsFilesLayers from './ImageDetailsFilesLayers.svelte';
 
 let { imageInfo }: { imageInfo: ImageInfo | undefined } = $props();
 
-let imageLayers: ImageFilesystemLayers | undefined = $state();
-let selectedLayer: ImageFilesystemLayerUI | undefined = $state();
-let loading: boolean = $state(false);
-let error: string = $state('');
-let showLayerOnly: boolean = $state(false);
+let imageLayers = $state<ImageFilesystemLayers>();
+let selectedLayer = $state<ImageFilesystemLayerUI>();
+let loading = $state<boolean>(false);
+let error = $state<string>('');
+let showLayerOnly = $state<boolean>(false);
 
 let filesProvidersUnsubscribe: Unsubscriber;
 let filesProvider: ImageFilesInfo;
--- a/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
+++ b/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
@@ -11,16 +11,17 @@ import FilesystemLayerView from './FilesystemLayerView.svelte';
 import { type ImageFilesystemLayerUI, toImageFilesystemLayerUIs } from './imageDetailsFiles';
 import ImageDetailsFilesLayers from './ImageDetailsFilesLayers.svelte';
 
-export let imageInfo: ImageInfo | undefined;
+let { imageInfo }: { imageInfo: ImageInfo | undefined } = $props();
 
-let imageLayers: ImageFilesystemLayers | undefined;
+let imageLayers: ImageFilesystemLayers | undefined = $state();
+let selectedLayer: ImageFilesystemLayerUI | undefined = $state();
+let loading: boolean = $state(false);
+let error: string = $state('');
+let showLayerOnly: boolean = $state(false);
+
 let filesProvidersUnsubscribe: Unsubscriber;
 let filesProvider: ImageFilesInfo;
-let selectedLayer: ImageFilesystemLayerUI;
-let showLayerOnly: boolean;
-let loading: boolean;
 let cancellableTokenId: number = 0;
-let error: string = '';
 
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -11,16 +11,17 @@ import FilesystemLayerView from './FilesystemLayerView.svelte';
 import { type ImageFilesystemLayerUI, toImageFilesystemLayerUIs } from './imageDetailsFiles';
 import ImageDetailsFilesLayers from './ImageDetailsFilesLayers.svelte';
 
-export let imageInfo: ImageInfo | undefined;
+let { imageInfo }: { imageInfo: ImageInfo | undefined } = $props();
+
+let imageLayers: ImageFilesystemLayers | undefined = $state();
+let selectedLayer: ImageFilesystemLayerUI | undefined = $state();
```
axel7083:
```suggestion
let imageLayers: ImageFilesystemLayers | undefined = $state(undefined);
let selectedLayer: ImageFilesystemLayerUI | undefined = $state(undefined);
```

Are we sure svelte is using `undefined` and not null?

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -11,16 +11,17 @@ import FilesystemLayerView from './FilesystemLayerView.svelte';
 import { type ImageFilesystemLayerUI, toImageFilesystemLayerUIs } from './imageDetailsFiles';
 import ImageDetailsFilesLayers from './ImageDetailsFilesLayers.svelte';
 
-export let imageInfo: ImageInfo | undefined;
+let { imageInfo }: { imageInfo: ImageInfo | undefined } = $props();
+
+let imageLayers = $state<ImageFilesystemLayers>();
```
axel7083, line 16: What is the type of imageLayers now ?

feloy: It seems to be a supported notation (and the one used in other components of podman desktop). Here, `$state<ImageFilesystemLayers>()` returns a value of type `ImageFilesystemLayers | undefined`

axel7083 (approved):

benoitf (approved):

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -65,9 +66,6 @@ onDestroy(() => {
   </div>
 {:else if imageLayers}
   <div class="flex flex-col w-full h-full p-8 pr-0 text-[var(--pd-content-text)] bg-[var(--pd-content-bg)]">
-    <div class="pr-4">
-      <slot name="header-info" />
-    </div>
```
benoitf, line 70: how is it replaced now ?

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
jeffmaury, line 16: $state is used to declare reactive data. Is this required here ?

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -65,9 +66,6 @@ onDestroy(() => {
   </div>
 {:else if imageLayers}
   <div class="flex flex-col w-full h-full p-8 pr-0 text-[var(--pd-content-text)] bg-[var(--pd-content-bg)]">
-    <div class="pr-4">
```
jeffmaury, line 68: Same concern here

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
feloy: I removed this part as it is never used: this component is inspired from the `ProviderResultPage` component, where this slot is used, but is never used in `ImageDetailsFiles`. 

Also, moving to svelte5, the checks refuses to use this svelte4 notation, I had to change or remove it.

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
feloy: It needs to be a state, as it is modified in line 40 (`imageLayers = layers`) and read in line 67 (`{:else if imageLayers}`)

jeffmaury: Yes but it is modified in onMount so do we really need reactive here ?

feloy: It is on a `subscribe` (line 31) initiated in onMount, it can happen at any time

jeffmaury (approved):
LGTM

simonrey1:
@feloy for file `packages/renderer/src/lib/image/ImageDetailsFiles.svelte` I think you also need to change:

`imageLayers = await window.imageGetFilesystemLayers(provider.id, img, cancellableTokenId);`

to 

`imageLayers = await window.imageGetFilesystemLayers(provider.id, $state.snapshot(img), cancellableTokenId);`

PR #9157, opened 2024-10-01
feat: dropdown component

deboer-tim:
### What does this PR do?

As per the last time this item was discussed, I spent a while trying to use svelte-select and a few other off-the-shelf select components to work, but ultimately they all had the same problem as we had with pulling images: even after tying to style them, I couldn't get any of them to fit into our current styling well enough that I was comfortable submitting a PR.

Two other minor things to note:
- Most components used some form of item[] instead of child option elements. This might be useful in the future (and could be added here) but would have required more migration up front.
- We don't have any immediate plans to expose this component in the UI library, but using a third-party library would have excluded that possibility.

So, here it is instead: I used the code from the Typeahead component, mixing in bits from Tooltip and other places to create a custom styled Dropdown component. I tried to keep required migration to a minimum, but some tests needed to be updated and I switched onInput to svelte 5 style.

Updated the preferences page and Run Image form to use the component so that there is something to test. There are ~13 other places where we use select which would be migrated in follow-up PRs.

### Screenshot / video of UI

https://github.com/user-attachments/assets/a908e81b-4ac9-406f-b0bc-ac5b75b32702

Core of issue #6199.

### How to test this PR?

Go to Settings > Preferences and Run Image form.

### DIFF
--- a/packages/ui/src/lib/dropdown/Dropdown.svelte
+++ b/packages/ui/src/lib/dropdown/Dropdown.svelte
@@ -1,29 +1,48 @@
 <script lang="ts">
 import { faCaretDown, faCheck } from '@fortawesome/free-solid-svg-icons';
+import type { Snippet } from 'svelte';
 import Fa from 'svelte-fa';
 
-export let id: string | undefined = undefined;
-export let name: string | undefined = undefined;
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let value: any = '';
-export let disabled: boolean = false;
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let onInput: (val: any) => void = () => {};
-
 interface Option {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  value: any;
+  value: unknown;
   label: string;
 }
 
-let opened: boolean = false;
-let selectLabel: string = '';
-let options: Option[] = [];
-let highlightIndex: number = -1;
-let pageStep: number = 10;
+let {
+  id,
+  name,
+  value = $bindable(),
+  disabled,
+  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+  onChange = () => {},
+  options = [],
+  class: className = '',
+  ariaInvalid = false,
+  ariaLabel = '',
+  children = undefined,
+}: {
+  id?: string;
+  name?: string;
+  value?: unknown;
+  disabled?: boolean;
+  onChange?: (val: unknown) => void;
+  options?: Option[];
+  class?: string;
+  ariaInvalid?: boolean | 'grammar' | 'spelling';
+  ariaLabel?: string;
+  children?: Snippet;
+} = $props();
+
+let opened: boolean = $state(false);
+let selectLabel: string = $state('');
+let highlightIndex: number = $state(-1);
 let comp: HTMLElement;
 
-$: selectLabel = options.find(o => o.value === value)?.label ?? value;
+const pageStep: number = 10;
+
+$effect(() => {
+  selectLabel = options.find(o => o.value === value)?.label ?? (typeof value === 'string' ? (value as string) : '');
+});
 
 function onKeyDown(e: KeyboardEvent): void {
   switch (e.key) {
@@ -105,9 +124,8 @@ function onEnter(i: number): void {
   highlightIndex = i;
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function onSelect(e: Event, newValue: any): void {
-  onInput(newValue);
+function onSelect(e: Event, newValue: unknown): void {
+  onChange(newValue);
   value = newValue;
   close();
   e.preventDefault();
@@ -132,7 +150,9 @@ function close(): void {
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function buildOptions(node: HTMLSelectElement): any {
-  options = [...node.options].map(o => ({ value: o.value, label: o.textContent ?? '' }));
+  if (options?.length === 0) {
+    options = [...node.options].map(o => ({ value: o.value, label: o.textContent ?? '' }));
+  }
   return {
     destroy(): void {
       options = [];
@@ -150,8 +170,7 @@ function onWindowClick(e: Event): void {
 <svelte:window on:click={onWindowClick} />
 
 <div
-  class="flex flex-row grow items-center px-1 py-1 group bg-[var(--pd-input-field-bg)] border-[1px] border-transparent min-w-24 relative {$$props.class ||
-    ''}"
+  class="flex flex-row grow items-center px-1 py-1 group bg-[var(--pd-input-field-bg)] border-[1px] border-transparent min-w-24 relative {className}"
   class:not(focus-within):hover:bg-[var(--pd-input-field-hover-bg)]={!disabled}
   class:focus-within:bg-[var(--pd-input-field-focused-bg)]={!disabled}
   class:focus-within:rounded-md={!disabled}
@@ -159,8 +178,8 @@ function onWindowClick(e: Event): void {
   class:border-b-[var(--pd-input-field-stroke)]={!disabled}
   class:hover:border-b-[var(--pd-input-field-hover-stroke)]={!disabled}
   class:border-b-[var(--pd-input-field-stroke-readonly)]={disabled}
-  aria-invalid={$$props['aria-invalid']}
-  aria-label={$$props['aria-label']}
+  aria-label={ariaLabel}
+  aria-invalid={ariaInvalid}
   bind:this={comp}>
   <button
     class="flex flex-row w-full outline-0 bg-[var(--pd-input-field-bg)] placeholder:text-[color:var(--pd-input-field-placeholder-text)] items-center text-start"
@@ -172,8 +191,8 @@ function onWindowClick(e: Event): void {
     disabled={disabled}
     id={id}
     name={name}
-    on:click={toggleOpen}
-    on:keydown={onKeyDown}>
+    onclick={toggleOpen}
+    onkeydown={onKeyDown}>
     <span class="grow">{selectLabel}</span>
     <div
       class:text-[var(--pd-input-field-stroke)]={!disabled}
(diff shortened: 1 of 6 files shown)


axel7083:
> We don't have any immediate plans to expose this component in the UI library

Actually I would have though we should do it the other way, first add it the ui library and create a Storybook page for it, so we can experiment with it.

> but using a third-party library would have excluded that possibility.

We are exposing the following dependencies through the @podman-desktop/ui-svelte library

https://github.com/containers/podman-desktop/blob/8e268d8df7f75422ace9cd396d4c4045de1344f0/packages/ui/package.json#L129-L135

---

image

I don't think the popup should be over the form itself.

If we look at some libraries we can clearly see they put is bellow.

Screenshot from 2024-10-02 09-46-11

packages/renderer/src/lib/ui/Dropdown.svelte
```diff
@@ -0,0 +1,201 @@
...
+    class:group-hover-placeholder:text-[color:var(--pd-input-field-placeholder-text)]={!disabled}
+    disabled={disabled}
+    id={id}
+    name={name}
+    on:click={open}
+    on:keydown={onKeyDown}>
+    <span class="grow">{selectLabel}</span>
+    <div
+      class:text-[var(--pd-input-field-stroke)]={!disabled}
+      class:text-[var(--pd-input-field-disabled-text)]={!!disabled}
```
gastoner: Shouldn't be here `!` instead of `!!` ?

gastoner:
Also, it is probably a different issue; on the Proxy page, we are not using the same dropdown (which we probably should?). 
image

And +1 to Axels comment with popup over form

deboer-tim:
> > We don't have any immediate plans to expose this component in the UI library
> 
> Actually I would have though we should do it the other way, first add it the ui library and create a Storybook page for it, so we can experiment with it.
> 
> > but using a third-party library would have excluded that possibility.
> 
> We are exposing the following dependencies through the @podman-desktop/ui-svelte library

I thought we took a decision that we would keep that list short, i.e. we provide basic controls to match PD's style, but if an extension wants something more complex like a chart we would not provide it; extension should use and style third-party library directly.

I get what you're saying about directly exposing in the UI library, but I would prefer to be more cautious, i.e. we add a component, make sure it works for all our use cases, go through at least one QE cycle, confirm there's an extension that wants/will use it, then expose. I would assume it's safe enough in this case (?) but didn't want to presume or overload a PR.

> I don't think the popup should be over the form itself.
> 
> If we look at some libraries we can clearly see they put is bellow.

I was trying to match the native one, which is apparently the only one that goes over top. I've moved the popup down, and since you can now click the control again that will now close it.

packages/renderer/src/lib/ui/Dropdown.svelte
deboer-tim: Yes, good catch. Typo fixed.

deboer-tim:
> Also, it is probably a different issue; on the Proxy page, we are not using the same dropdown (which we probably should?).

Correct, there are >15 HTML select components and only 4 (preferences enums + 3 in Run Image form) are fixed here to have something to test. Another PR or two will be required to switch all the others and mark #6199 complete.

> And +1 to Axels comment with popup over form

👍🏼 PR updated.

jeffmaury (approved):
LGTM but I don't get why teh component did not land in UI library

gastoner (approved):

packages/renderer/src/lib/ui/Dropdown.svelte
```diff
@@ -0,0 +1,209 @@
...
+      class:text-[var(--pd-input-field-stroke)]={!disabled}
+      class:text-[var(--pd-input-field-disabled-text)]={!disabled}
+      class:group-hover:text-[var(--pd-input-field-hover-stroke)]={!disabled}>
+      <Fa icon={faCaretDown} />
+    </div>
+  </button>
+
+  {#if opened}
+    <div
+      class="absolute top-full right-0 z-10 w-full max-h-80 rounded bg-[var(--pd-dropdown-bg)] border-[var(--pd-input-field-hover-stroke)] border-[1px] overflow-y-auto whitespace-nowrap">
```
axel7083: Here we are using `rounder`, and above we are using `rounded-md`

https://github.com/containers/podman-desktop/blob/70d2e3604d71ab5b867f682e62af733856afafc7/packages/renderer/src/lib/ui/Dropdown.svelte#L156

image

And some component above are using `rounder-sm`

https://github.com/containers/podman-desktop/blob/70d2e3604d71ab5b867f682e62af733856afafc7/packages/renderer/src/lib/image/RunImage.svelte#L853-L854

Should we make them match ?

benoitf (dismissed):
IMHO the PR should only focus on bringing the component and not bringing the component AND changing the usage of other parts using it.

At some point, it was hard to demonstrate the component without experiencing it. But now, we have a way to showcase a component, this is why we made storybook stories. You can do trial and live change to the component and be focused only on the component.

Also to me the end goal is to have in the UI library so if it's first in renderer and we change the code to use it there, then we move it to UI, it would make additional PR to replace internal to external that we can prevent doing it only once.

We could flag the component as being in 'beta' state

deboer-tim:
Several people commented on going straight to UI component, so I moved it there. Added some typing due to higher checks in ui package, added a basic story, and removed the changes in renderer. Rebased to clear up multiple changes to renderer + removal, and get on latest.

Fixed inconsistent rounded-md.

Not sure how to mark as 'beta' or whether it's worth it.

axel7083:
> Not sure how to mark as 'beta' or whether it's worth it.

Dropdown can be tested here https://66fea7a62443dcad09ce6f63--podman-desktop-pr.netlify.app/storybook?id=dropdown--docs

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -0,0 +1,210 @@
...
+export let id: string | undefined = undefined;
+export let name: string | undefined = undefined;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export let value: any = '';
+export let disabled: boolean = false;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export let onInput: (val: any) => void = () => {};
+
+interface Option {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf: I don't know why we have so many 'any' there
it looks like we're having more than 10 // eslint-disable-next-line @typescript-eslint/no-explicit-any


should we have a generics for the type of the field we want to handle ?
or is it just `unknown`

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -0,0 +1,210 @@
...
+import { faCaretDown, faCheck } from '@fortawesome/free-solid-svg-icons';
+import Fa from 'svelte-fa';
+
+export let id: string | undefined = undefined;
+export let name: string | undefined = undefined;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export let value: any = '';
+export let disabled: boolean = false;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export let onInput: (val: any) => void = () => {};
```
benoitf: it's not clear to me what the `onInput` will do, is it more accurate to use `onChange` ?

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -0,0 +1,210 @@
+<script lang="ts">
+import { faCaretDown, faCheck } from '@fortawesome/free-solid-svg-icons';
+import Fa from 'svelte-fa';
+
+export let id: string | undefined = undefined;
+export let name: string | undefined = undefined;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export let value: any = '';
```
benoitf: IMHO you should use $props for new components as reactivity is better handled (it can say for example if we're missing some $state on some fields, etc)

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -0,0 +1,210 @@
...
+}
+
+let opened: boolean = false;
+let selectLabel: string = '';
+let options: Option[] = [];
+let highlightIndex: number = -1;
+let pageStep: number = 10;
+let comp: HTMLElement;
+
+$: selectLabel = options.find(o => o.value === value)?.label ?? value;
```
benoitf: ==> $effect

benoitf: and options should be a state then

storybook/src/stories/Dropdown.stories.svelte
```diff
@@ -0,0 +1,62 @@
...
+      description: 'Id of the dropdown',
+      defaultValue: undefined,
+    },
+    name: {
+      control: 'text',
+      description: 'Name of the dropdown',
+      defaultValue: undefined,
+    },
+  },
+});
```
benoitf, line 31: I think it's not telling the callback like
https://github.com/containers/podman-desktop/blob/main/storybook/src/stories/button/Button.stories.svelte#L17-L19

storybook/src/stories/Dropdown.stories.svelte
```diff
@@ -0,0 +1,62 @@
...
+      defaultValue: false,
+    },
+    id: {
+      control: 'text',
+      description: 'Id of the dropdown',
+      defaultValue: undefined,
+    },
+    name: {
+      control: 'text',
+      description: 'Name of the dropdown',
```
benoitf: where the name is used, is it for just an aria-label or something else

packages/renderer/src/lib/ui/Dropdown.svelte
deboer-tim: Updated both to rounded-md, which is what design uses.

packages/ui/src/lib/dropdown/Dropdown.svelte
deboer-tim: Most should be unknown, updated.

packages/ui/src/lib/dropdown/Dropdown.svelte
deboer-tim: I had used onInput since that's what select had, but switched to more consistent onChange instead.

packages/ui/src/lib/dropdown/Dropdown.svelte
deboer-tim: Updates to props. As per comment, this triggered requiring everything to Svelte 5.

packages/ui/src/lib/dropdown/Dropdown.svelte
deboer-tim: Fixed.

storybook/src/stories/Dropdown.stories.svelte
deboer-tim: Added onChange function to Storybook.

storybook/src/stories/Dropdown.stories.svelte
deboer-tim: Just the control name, used by some playwright tests. Looks like we aren't adding it to Storybook in most other cases so I removed it.

deboer-tim:
Rebased on main and made changes based on review comments. Adding some Svelte 5 kept triggering more changes required so it is now fully updated to Svelte 5.
- any fixed to unknown and eslint disables removed
- use Svelte 5 props, bindable, effect, state
- use Svelte 5 render/snippets
- use Svelte 5 events / functions
- updated Storybook. Removed id/name (inconsistent, but we seem to omit simple/aria props for most components) and added options.

Ready for another review.

PR #9166, opened 2024-10-02
fix: handle userModeNetwork visibility even when only one provider is available

lstocchi:
### What does this PR do?

https://github.com/containers/podman-desktop/pull/9105 introduced a dropdownlist to select wsl or hyperv (when both are available) to specify the proider to use when creating a new machine. 
Based on the choice the user-mode-networking option is visible or not (wsl=visible, hyperv=hidden).

However, there was a use case not handled. When only one provider is enabled, the user-mode-network is always hidden. But if the provider is wsl it should be visible.

### Screenshot / video of UI

N/A

N/A, raised on slack

### How to test this PR?

1. run tests or have only one provider enabled

### DIFF
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -98,7 +98,7 @@ const podmanBinaryHelper = new PodmanBinaryLocationHelper();
 const podmanInfoHelper = new PodmanInfoHelper();
 
 let createWSLMachineOptionSelected = false;
-let wslAndHypervEnabled = false;
+let wslAndHypervEnabledContextValue = false;
 let wslEnabled = false;
 
 let shouldNotifySetup = true;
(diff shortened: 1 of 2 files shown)


lstocchi:
cc @cbr7 @odockal

cbr7:
Checked the PR, user mode networking slider is available again for wsl.

cbr7 (approved):

jeffmaury (approved):
LGTM

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -1281,7 +1282,8 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
     extensionApi.context.setValue(START_NOW_MACHINE_INIT_SUPPORTED_KEY, isStartNowAtMachineInitSupported(version));
     extensionApi.context.setValue(USER_MODE_NETWORKING_SUPPORTED_KEY, isUserModeNetworkingSupported(version));
     extensionApi.context.setValue(PODMAN_PROVIDER_LIBKRUN_SUPPORTED_KEY, isLibkrunSupported(version));
-    const wslHypervEnabled = (await isWSLEnabled()) && (await isHyperVEnabled());
+    wslEnabled = await isWSLEnabled();
+    const wslHypervEnabled = wslEnabled && (await isHyperVEnabled());
```
odockal: Not really a problem, since the constant is defined in the local scope, but it might lead to a problem in future. We have the constant named `wslHypervEnabled` with local scope and we have also file scope variable `wslAndHypervEnabled` defined on the line 101.

lstocchi: The file scope variable is used to keep state of the context value `WSL_HYPERV_ENABLED_KEY` so to verify if we have to actually update the context or it has already updated. 
The local scope variable is just a temporary variable used to store the result of `wslEnabled && (await isHyperVEnabled())` and pass it to `updateWSLHyperVEnabledValue` . So they are two different things. I cannot replace one with the other.

So, I'm not sure to understand the concern. Do you want me to update the name of the variables or remove the local scope and call directly `updateWSLHyperVEnabledValue(wslEnabled && (await isHyperVEnabled()))` or something else?

odockal: Just the name, nothing else.

odockal: I have tested it in multiple combinations, wsl and hyperv, standalone hyperv, standalone wsl and in standard and elevated mode. Now all good!

lstocchi: @odockal variable name updated 👍

odockal (approved):
LGTM

PR #9172, opened 2024-10-02
fix: skip podman check on hypervCheck.execute when called as installation preflight

### LINKED ISSUE #9158: Cannot install podman even with Hyperv Windows feature enabled
### Bug description

On the windows, there is now a requirement check for hyperv virtualization that allows to install podman only if podman is 5.2.0 or greater. But I do not have any podman installed and so I am stuck.

### Operating system

windows

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. Have no previous podman installed on the system
2. Do not have Virtual Machine platform feature enabled
3. Have HyperV feature enabled
4. Install podman desktop
5. Click on install podman in dashboard or via onboarding
6. assert: Podman installation is allowed -> I can install podman when I have hyperv feature enabled
Actual result: The check does not allow me to continue even though I have hyperv enabled. 

Update: I cannot instal podman at all, even though I have hyperv disabled and VMP enabled. Check passes via VMP but got stuck on the: "Hyper-V is only supported with podman version >= 5.2.0"

### Relevant log output

_No response_

### Additional context

I should be able to install podman with one of the virtualization providers enabled, either Virtual Machine Platform or hyperv.

lstocchi:
### What does this PR do?

This PR updates the hypervCheck to skip the podman version check when hypervCheck is called as an installation preflight.

If you are installing podman, the version is unknown, and so the hyperv check will always fail. With this PR we skip that check as not needed in such as situation

### Screenshot / video of UI

N/A

it resolves #9158 

### How to test this PR?

1. try to install podman from desktop

### DIFF
--- a/extensions/podman/packages/extension/src/podman-install.spec.ts
+++ b/extensions/podman/packages/extension/src/podman-install.spec.ts
@@ -633,6 +633,21 @@ describe('HyperV', () => {
     expect(result.description).equal('Hyper-V is only supported with podman version >= 5.2.0.');
   });
 
+  test('expect HyperV preflight skips podman version check if installationPreflightMode is true', async () => {
+    const execSpy = vi.spyOn(extensionApi.process, 'exec').mockImplementation(() => {
+      throw new Error();
+    });
+
+    const hyperVCheck = new HyperVCheck(true);
+    const result = await hyperVCheck.execute();
+    // exec should only be called once inside the isUserAdmin function, if so the podman check has been skipped as expected
+    expect(execSpy).toBeCalledTimes(1);
+    expect(execSpy).toBeCalledWith('powershell.exe', [
+      '$null -ne (whoami /groups /fo csv | ConvertFrom-Csv | Where-Object {$_.SID -eq "S-1-5-32-544"})',
+    ]);
+    expect(result.successful).toBeFalsy();
+  });
+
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM

benoitf (approved):
code ok but using macOS I was not able to check

odockal:
I can now install podman without podman install. But I see a follow up issues here, I need VMP in order to install podman. If I have only hyperv, I cannot install podman since it says: `Virtual Machine Platform should be enabled to be able to run Podman` -> could be a follow up issue?

jeffmaury:
> Virtual Machine Platform

#9177 created

lstocchi:
> I can now install podman without podman install. But I see a follow up issues here, I need VMP in order to install podman. If I have only hyperv, I cannot install podman since it says: `Virtual Machine Platform should be enabled to be able to run Podman` -> could be a follow up issue?

By looking quickly at the code I think it is different from this. 
The virtual machine platform check should only be executed with the wsl check but it is outside the orCheck. 
So, yes, different PR
https://github.com/containers/podman-desktop/blob/e82beb554f23139adb49c648e12e1702a0eccb99/extensions/podman/packages/extension/src/podman-install.ts#L456-L464

odockal:
@lstocchi please rebase and let's merge :)

odockal (approved):

Issue #9025, opened 2024-09-23
widget to display status of system docker socket(npipe)


benoitf:
### Is your enhancement related to a problem? Please describe

in the docker compatibility page, displays the status of the system connect
- is it responding
- if yes, what is the engine behind

### Describe the solution you'd like

see at mockup sections https://github.com/containers/podman-desktop/issues/8338

PR #9179, opened 2024-10-02
chore: add a way to get status information for mac/linux/windows docker system socket

benoitf:
### What does this PR do?
This is the backend part of https://github.com/containers/podman-desktop/issues/9025

query /var/run/docker.sock or npipe and provide info on the connection to the frontend
correlating for macOS the provider serving this connection

related to #9025 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/docker/docker-compatibility.spec.ts
+++ b/packages/main/src/plugin/docker/docker-compatibility.spec.ts
@@ -45,7 +45,7 @@ const providerRegistry: ProviderRegistry = {
 export class TestDockerCompatibility extends DockerCompatibility {
   public override getTypeFromServerInfo(
     info: { OperatingSystem?: string },
-    podmanInfo: unknown,
+    podmanInfo?: unknown,
   ): DockerSocketServerInfoType {
     return super.getTypeFromServerInfo(info, podmanInfo);
   }
@@ -108,10 +108,7 @@ describe('getTypeFromServerInfo', async () => {
     const serverInfo = {
       OperatingSystem: 'Docker Desktop',
     };
-
-    const podmanInfo = {};
-
-    expect(dockerCompatibility.getTypeFromServerInfo(serverInfo, podmanInfo)).toBe('docker');
+    expect(dockerCompatibility.getTypeFromServerInfo(serverInfo)).toBe('docker');
   });
 
   test('check podman OS without podman info', async () => {
@@ -131,7 +128,9 @@ describe('getTypeFromServerInfo', async () => {
       OperatingSystem: 'foo',
     };
 
-    const podmanInfo = {};
+    const podmanInfo = {
+      ServerVersion: '1.0.0',
+    };
 
     expect(dockerCompatibility.getTypeFromServerInfo(serverInfo, podmanInfo)).toBe('podman');
   });
--- a/packages/main/src/plugin/docker/docker-compatibility.spec.ts
+++ b/packages/main/src/plugin/docker/docker-compatibility.spec.ts
@@ -18,14 +18,66 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { beforeAll, expect, test, vi } from 'vitest';
+import type { Stats } from 'node:fs';
+import { promises } from 'node:fs';
 
+import type { ProviderContainerConnection } from '@podman-desktop/api';
+import type { Mock } from 'vitest';
+import { beforeAll, describe, expect, test, vi } from 'vitest';
+
+import type { DockerSocketServerInfoType } from '/@api/docker-compatibility-info.js';
+import type { ProviderInfo } from '/@api/provider-info.js';
+
+import * as util from '../../util.js';
 import type { ApiSenderType } from '../api.js';
 import { ConfigurationRegistry } from '../configuration-registry.js';
 import type { Directories } from '../directories.js';
+import type { ProviderRegistry } from '../provider-registry.js';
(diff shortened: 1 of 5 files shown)


gastoner (approved):
There were a few `as unknown as something` is this correct?

packages/main/src/plugin/docker/docker-compatibility.ts
```diff
@@ -46,4 +60,106 @@ export class DockerCompatibility {
...
+      const resolved = await promises.readlink(path);
+      return this.resolveLinkIfAny(resolved);
+    }
+
+    return path;
+  }
+
+  async getSystemDockerSocketMappingStatus(): Promise<DockerSocketMappingStatusInfo> {
+    // first, try to connect to the docker socket if on Linux or macOS
+    // if the connection is successful, return @the status as running
```
feloy: Is this comment still accurate?

packages/main/src/plugin/docker/docker-compatibility.ts
```diff
@@ -46,4 +60,106 @@ export class DockerCompatibility {
...
+        // is UNIX_SOCKET_PATH a symbolic link ?
+        const socketPath = await this.resolveLinkIfAny(DockerCompatibility.UNIX_SOCKET_PATH);
+
+        // do we have a matching provider for the socket path ?
+        const currentConnections = this.#providerRegistry.getContainerConnections();
+
+        // search if we have a connection with the same socket path
+        const foundConnection = currentConnections.find(c => c.connection.endpoint.socketPath === socketPath);
+
+        // provider is
```
feloy: is it complete?

feloy (approved):

packages/main/src/plugin/docker/docker-compatibility.ts
benoitf: I think I moved to another task 😂

I will check what I wanted to say

packages/main/src/plugin/docker/docker-compatibility.ts
benoitf: accurate but I'll change the wording as it looks like it's unclear 👍

packages/main/src/plugin/docker/docker-compatibility.ts
benoitf: fixed

packages/main/src/plugin/docker/docker-compatibility.ts
benoitf: fixed, please review

benoitf:
@feloy typedoc amended

packages/main/src/plugin/docker/docker-compatibility.spec.ts
```diff
@@ -44,8 +96,204 @@ test('should register a configuration', async () => {
...
+describe('getTypeFromServerInfo', async () => {
+  test('Docker Desktop', async () => {
+    const dockerCompatibility = new TestDockerCompatibility(configurationRegistry, providerRegistry);
+    const serverInfo = {
+      OperatingSystem: 'Docker Desktop',
+    };
+
+    const podmanInfo = {};
+
+    expect(dockerCompatibility.getTypeFromServerInfo(serverInfo, podmanInfo)).toBe('docker');
```
axel7083:
```suggestion
    expect(dockerCompatibility.getTypeFromServerInfo(serverInfo, {})).toBe('docker');
```

No eslint complain ?

packages/main/src/plugin/docker/docker-compatibility.ts
```diff
@@ -46,4 +60,112 @@ export class DockerCompatibility {
 
     this.#configurationRegistry.registerConfigurations([dockerCompatibilityConfiguration]);
   }
+
+  protected getTypeFromServerInfo(info: { OperatingSystem?: string }, podmanInfo: unknown): DockerSocketServerInfoType {
```
axel7083: We cannot have a better typing for `podmanInfo` ?

benoitf: we don't care on the structure, we only want to know if there is data in it

packages/main/src/plugin/docker/docker-compatibility.spec.ts
benoitf: why should it complain, it's an empty object.

eslint complain about empty interface or empty function

packages/main/src/plugin/docker/docker-compatibility.ts
axel7083: then make it a boolean an check beforehand ? having unknown feel weird. Even with the requirement on it, I feel like having `podmanInfo: Info | undefined` would be clearer than unknown

benoitf: but it means that I would have to declare a new interface `Info` to declare what is inside the data where I don't need to know what is inside

benoitf: I will change parameter to be `podmanInfo?: unknown` so it's clear it's optional

benoitf: amended

packages/main/src/plugin/docker/docker-compatibility.spec.ts
benoitf: removed the object and made the podmanInfo parameter optional

axel7083 (approved):
Code LGTM

PR #9181, opened 2024-10-03
chore(test): test framework refactoring

### LINKED ISSUE #9170: Several e2e tests are failing in cicd and need to be fixed
### Bug description

Some e2e tests are failing in cicd and need fixing.

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Some fixes for e2e tests and some code refactoring to use test.step functionality.

https://github.com/containers/podman-desktop/issues/9170

### DIFF
--- a/tests/playwright/src/model/pages/settings-bar.ts
+++ b/tests/playwright/src/model/pages/settings-bar.ts
@@ -46,7 +46,8 @@ export class SettingsBar {
   public async openTabPage<T extends SettingsPage>(type: new (page: Page) => T): Promise<T> {
     return await test.step(`Open ${type.name} tab from Settings`, async () => {
       const desiredPage = new type(this.page);
-      await (await desiredPage.getTab()).click();
+      const tab = await desiredPage.getTab();
+      await tab.click();
       return desiredPage;
     });
   }
--- a/tests/playwright/src/model/pages/settings-bar.ts
+++ b/tests/playwright/src/model/pages/settings-bar.ts
@@ -44,7 +44,7 @@ export class SettingsBar {
   }
 
   public async openTabPage<T extends SettingsPage>(type: new (page: Page) => T): Promise<T> {
-    return await test.step(`Open ${type.name} tab from Settings`, async () => {
+    return test.step(`Open ${type.name} tab from Settings`, async () => {
       const desiredPage = new type(this.page);
       const tab = await desiredPage.getTab();
       await tab.click();
--- a/tests/playwright/src/model/pages/settings-bar.ts
+++ b/tests/playwright/src/model/pages/settings-bar.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import type { Locator, Page } from '@playwright/test';
+import { type Locator, type Page, test } from '@playwright/test';
 
 import type { SettingsPage } from './settings-page';
 
@@ -44,9 +44,11 @@ export class SettingsBar {
   }
 
   public async openTabPage<T extends SettingsPage>(type: new (page: Page) => T): Promise<T> {
-    const desiredPage = new type(this.page);
-    await (await desiredPage.getTab()).click();
-    return desiredPage;
+    return await test.step(`Open ${type.name} tab from Settings`, async () => {
+      const desiredPage = new type(this.page);
+      await (await desiredPage.getTab()).click();
+      return desiredPage;
+    });
   }
 
   public getSettingsNavBarTabLocator(name: string): Locator {
--- a/tests/playwright/src/model/pages/volumes-page.ts
+++ b/tests/playwright/src/model/pages/volumes-page.ts
@@ -39,7 +39,7 @@ export class VolumesPage extends MainPage {
   }
 
   async openCreateVolumePage(volumeName: string): Promise<CreateVolumePage> {
-    return await test.step('Open Create Volume Page', async () => {
+    return test.step('Open Create Volume Page', async () => {
       const row = await this.getVolumeRowByName(volumeName);
       if (row !== undefined) {
         throw Error('Volume is already created');
@@ -52,7 +52,7 @@ export class VolumesPage extends MainPage {
   }
 
   async openVolumeDetails(volumeName: string): Promise<VolumeDetailsPage> {
-    return await test.step('Open Volume Details Page', async () => {
+    return test.step('Open Volume Details Page', async () => {
       const volumeRow = await this.getVolumeRowByName(volumeName);
       if (volumeRow === undefined) {
         throw Error(`Volume: ${volumeName} does not exist`);
@@ -65,7 +65,7 @@ export class VolumesPage extends MainPage {
   }
 
   async deleteVolume(volumeName: string): Promise<VolumesPage> {
-    return await test.step('Delete Volume', async () => {
+    return test.step('Delete Volume', async () => {
       const volumeRow = await this.getVolumeRowByName(volumeName);
       if (volumeRow === undefined) {
         throw Error(`Volume: ${volumeName} does not exist`);
@@ -84,7 +84,7 @@ export class VolumesPage extends MainPage {
   }
 
   protected async volumeExists(name: string): Promise<boolean> {
-    return await test.step(`Check if volume ${name} exists`, async () => {
+    return test.step(`Check if volume ${name} exists`, async () => {
       const result = await this.getVolumeRowByName(name);
       return result !== undefined;
     });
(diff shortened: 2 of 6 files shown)


tests/playwright/src/model/pages/settings-bar.ts
```diff
@@ -44,9 +44,11 @@ export class SettingsBar {
   }
 
   public async openTabPage<T extends SettingsPage>(type: new (page: Page) => T): Promise<T> {
-    const desiredPage = new type(this.page);
-    await (await desiredPage.getTab()).click();
-    return desiredPage;
+    return await test.step(`Open ${type.name} tab from Settings`, async () => {
```
benoitf: wondering on the pattern, if it's ok to return the promise, not the await of the promise (as the signature of the method is to return a promise)

tests/playwright/src/model/pages/settings-bar.ts
```diff
@@ -44,9 +44,11 @@ export class SettingsBar {
...
+      const desiredPage = new type(this.page);
+      await (await desiredPage.getTab()).click();
```
benoitf: I think it may be better to cut the line in two like avoiding 2 await in a row
```
const tab = await desiredPage.getTab();
await tab.click();
```
```suggestion
      const tab = await desiredPage.getTab();
      await tab.click();
```

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -39,47 +39,55 @@ export class VolumesPage extends MainPage {
...
   async openCreateVolumePage(volumeName: string): Promise<CreateVolumePage> {
-    const row = await this.getVolumeRowByName(volumeName);
-    if (row !== undefined) {
-      throw Error('Volume is already created');
-    }
-
-    await playExpect(this.createVolumeButton).toBeEnabled();
-    await this.createVolumeButton.click();
-    return new CreateVolumePage(this.page);
+    return await test.step('Open Create Volume Page', async () => {
```
benoitf: same remark, usually we return the promise, not the await of a promise

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -39,47 +39,55 @@ export class VolumesPage extends MainPage {
...
   async openVolumeDetails(volumeName: string): Promise<VolumeDetailsPage> {
-    const volumeRow = await this.getVolumeRowByName(volumeName);
-    if (volumeRow === undefined) {
-      throw Error(`Volume: ${volumeName} does not exist`);
-    }
-    const containerRowName = volumeRow.getByRole('cell').nth(3);
-    await containerRowName.click();
-
-    return new VolumeDetailsPage(this.page, volumeName);
+    return await test.step('Open Volume Details Page', async () => {
```
benoitf: same remark

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -39,47 +39,55 @@ export class VolumesPage extends MainPage {
...
-    if (volumeRow === undefined) {
-      throw Error(`Volume: ${volumeName} does not exist`);
-    }
-    const containerRowDeleteButton = volumeRow.getByRole('button', { name: 'Delete Volume' });
-    await playExpect(containerRowDeleteButton).toBeEnabled();
-    await containerRowDeleteButton.click();
-    await handleConfirmationDialog(this.page);
-
-    return this;
+    return await test.step('Delete Volume', async () => {
```
benoitf: same remark

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -90,26 +98,32 @@ export class VolumesPage extends MainPage {
...
   }
 
-  async waitForVolumeExists(name: string): Promise<boolean> {
-    if (!name) {
-      throw Error('Volume name is not provided');
-    }
-    await waitUntil(async () => await this.volumeExists(name));
-    return true;
+  async waitForVolumeExists(name: string, timeout = 30_000): Promise<boolean> {
+    return await test.step(`Wait for volume ${name} to exist`, async () => {
```
benoitf: same remark, why the await

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -90,26 +98,32 @@ export class VolumesPage extends MainPage {
...
   }
 
-  async waitForVolumeDelete(name: string): Promise<boolean> {
-    if (!name) {
-      throw Error('Volume name is not provided');
-    }
-    await waitWhile(async () => await this.volumeExists(name));
-    return true;
+  async waitForVolumeDelete(name: string, timeout = 30_000): Promise<boolean> {
+    return await test.step(`Wait for volume ${name} to be deleted`, async () => {
```
benoitf: same remark

tests/playwright/src/model/pages/volumes-page.ts
```diff
@@ -90,26 +98,32 @@ export class VolumesPage extends MainPage {
...
+      return true;
+    });
   }
 
   async pruneVolumes(): Promise<VolumesPage> {
-    await playExpect(this.pruneVolumesButton).toBeEnabled();
-    await this.pruneVolumesButton.click();
-    await handleConfirmationDialog(this.page, 'Prune');
-    return this;
+    return await test.step('Prune Volumes', async () => {
```
benoitf: same remark

ScrewTSW (commented):
Agreed with @benoitf

tests/playwright/src/model/pages/settings-bar.ts
cbr7: I'm not really sure what you mean. The await there is for test.step, test.step is an async call which needs to be awaited.

This is the recommended usage as per the documentation. https://playwright.dev/docs/api/class-test#test-step

tests/playwright/src/model/pages/volumes-page.ts
cbr7: Same as above, the await is for test.step, the base functionality has not changed.

tests/playwright/src/model/pages/settings-bar.ts
cbr7: ok, I changed it and split it on two lines even though that was how it was initially.

tests/playwright/src/model/pages/settings-bar.ts
benoitf: I'm not speaking about playwright, I'm speaking about TypeScript

if you have a method returning a promise, you return a promise, not the await of the promise.
as anyway the caller of the method will await on it.

```ts
async myMethod: Promise<unknown> {
  return step(); // that returns the promise
}
```


or, if you just need to wait, then the signature of the method is Promise<void>

and then you can just do:

```ts
async myMethod: Promise<void> {
  await step();
}
```

cbr7: ok I understand what you mean, but this is the current practice we've used in the test framework and I'd rather not have two different approaches at the same time without some sort of understanding.

We can certainly refactor it to the way you suggest, however I don't think that it's in scope for this PR.

Such a change would not be limited to just test.step implementations, it would require refactoring for every async call in the test framework.

benoitf: > We can certainly refactor it to the way you suggest, however I don't think that it's in scope for this PR.

but the code before this PR was not using any 'return await', it's part of this PR that is introducing the `return await` so it's in the scope of this PR

I'm asking how playwright behave if you just return the promise without awaiting it

using
```
return test.step(`Open ${type.name} tab from Settings`, async () => {
```

instead of
```
return await test.step(`Open ${type.name} tab from Settings`, async () => {
```

cbr7: Both those forms will work because in the end the call is awaited in the actual test, there won't be a functional difference.

image

benoitf: so why are we not using less code ?

cbr7: Because as I mentioned this was the approach that was taken in the test framework, the same argument can be made for most(all?) async calls in there so I didn't want to implement a new approach just because it would be slightly less code. As I said we can certainly refactor it to this way, functionally there won't be a difference but I think that requires consensus on it, not just changing it suddenly in an unrelated PR.

EDIT: @odockal how do you feel about changing the framework and removing the somewhat superfluous await calls inside the methods that are promises in the test framework.

jeffmaury (commented):
Was not convinced initially by remarks about extra await but found the rule quite logical: do not await the last expression of an async method

cbr7:
@benoitf @jeffmaury As requested I've removed the final awaits in async methods

cbr7:
Test run here: https://github.com/containers/podman-desktop-e2e/actions/runs/11164800808

odockal (approved):

benoitf (approved):
thanks

PR #9183, opened 2024-10-03
feat: bring docker compatibility page if flag is enabled

benoitf:
### What does this PR do?
bring docker compatibility page if flag is enabled
note: page will be empty (no widgets in it for now) and be displayed only if experimental flag is enabled

part of https://github.com/containers/podman-desktop/issues/9025

### How to test this PR?

### DIFF
--- a/packages/renderer/src/PreferencesNavigation.spec.ts
+++ b/packages/renderer/src/PreferencesNavigation.spec.ts
@@ -19,25 +19,35 @@
 import '@testing-library/jest-dom/vitest';
 
 import { render, screen } from '@testing-library/svelte';
-import { beforeAll, expect, test } from 'vitest';
+import { tick } from 'svelte';
+import type { TinroRouteMeta } from 'tinro';
+import { beforeEach, expect, test, vi } from 'vitest';
 
 import PreferencesNavigation from './PreferencesNavigation.svelte';
 
 // fake the window.events object
-beforeAll(() => {
-  (window.events as unknown) = {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    receive: (_channel: string, func: any) => {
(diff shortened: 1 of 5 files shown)


packages/renderer/src/PreferencesNavigation.spec.ts
```diff
@@ -53,3 +63,42 @@ test('Test rendering of the preferences navigation bar and its items', () => {
...
+  // expect getConfigurationValue to be called
+  expect(window.getConfigurationValue).toBeCalledWith('dockerCompatibility.enabled');
+
+  const dockerCompatLink = screen.getByRole('link', { name: 'Docker Compatibility' });
+  expect(dockerCompatLink).toBeVisible();
+});
+
+test('Test rendering of the compatibility docker page is hidden if disabled', async () => {
+  // mock window.getConfigurationValue
+  vi.mocked(window.getConfigurationValue<boolean>).mockReset();
```
gastoner, line 86: Is this necessary here, since there is called `vi.resetAllMocks();` on top of the file

gastoner (approved):

packages/renderer/src/PreferencesNavigation.spec.ts
benoitf: just after the mocks are reset, we defined one result for a mock
https://github.com/containers/podman-desktop/blob/21d10c36f86dd04e6a3da7479c09865191f63c8f/packages/renderer/src/PreferencesNavigation.spec.ts#L43

so to be sure someone will not initialize the object with different patterns, like with an implementation, etc in the beforeEach call, by doing this I'm 100% sure it's reset and I can override the result

axel7083 (approved):
LGTM

PR #9186, opened 2024-10-03
feat(UI): task indicator 

### LINKED ISSUE #7400: Visibility of running tasks

#7399 will add a toast notification when a task is created, but we expect that either the user can close the toast or it will close automatically after a few seconds. This means that the user can't easily tell when it completes or if there is still a task running without opening the task manager.

We need some visible indication when there is a task running. Some options that have been raised:
- animation of the current bell
- adding the number of running tasks to the bell
- status bar entry with the name of the currently running task (or some flavour of '2 tasks running')
- progress bar or spinner directly on the status bar

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Waiting for feedback (see video section), first proposal to resolve https://github.com/containers/podman-desktop/issues/7400

### Screenshot / video of UI

https://github.com/user-attachments/assets/186729bc-35bf-4f18-9805-f914ded2aa3c

image

image

image

### How to test this PR?

- [x] Tests has been added

Add the following to the `settings.json` **before** started pd to test

```json
{
  "tasks.StatusBar": true
}
```

### Additionals

- [x] Adding experimental flags to enable or not (should be hidden)
- [x] Limit the maximum width of the title (using text-ellipsis)
- [x] Hover to see the title

### DIFF
--- a/packages/main/src/plugin/tasks/task-manager.ts
+++ b/packages/main/src/plugin/tasks/task-manager.ts
@@ -18,11 +18,13 @@
 
 import type { NotificationOptions } from '@podman-desktop/api';
 
+import type { ConfigurationRegistry } from '/@/plugin/configuration-registry.js';
 import { NotificationImpl } from '/@/plugin/tasks/notification-impl.js';
 import type { NotificationTask } from '/@/plugin/tasks/notifications.js';
 import { TaskImpl } from '/@/plugin/tasks/task-impl.js';
 import type { Task, TaskAction, TaskUpdateEvent } from '/@/plugin/tasks/tasks.js';
 import type { NotificationTaskInfo, TaskInfo } from '/@api/taskInfo.js';
+import { ExperimentalTasksSettings } from '/@api/tasks-preferences.js';
 
 import type { ApiSenderType } from '../api.js';
 import type { CommandRegistry } from '../command-registry.js';
@@ -37,6 +39,7 @@ export class TaskManager {
     private apiSender: ApiSenderType,
     private statusBarRegistry: StatusBarRegistry,
     private commandRegistry: CommandRegistry,
+    private configurationRegistry: ConfigurationRegistry,
   ) {}
 
   public init(): void {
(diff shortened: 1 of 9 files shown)


deboer-tim:
+1. For my preferences just two minor comments:
- The progress bar is a little wide, I'd go ~20-25% smaller.
- A poor task might have a really long name, set a max-w and text-ellipsis for the title. IMHO even `Stop podman-machine-default` should be shortened.

benoitf:
if truncated, hover should display the name of the task

packages/main/src/plugin/tasks/task-manager.ts
```diff
@@ -47,6 +50,21 @@ export class TaskManager {
       this.apiSender.send('toggle-task-manager', '');
       this.setStatusBarEntry(false);
     });
+
+    this.configurationRegistry.registerConfigurations([
+      {
+        id: 'preferences.experimental.tasks',
+        title: 'Tasks preference',
```
deboer-tim: Remove 'preference', it's redundant and none of our other preferences have it.

Actually, I am wondering if we should call the section Experimental? Have a single place to group any experimental preferences that we expect will disappear (become default behaviour or be rejected) or change ids in the future.

packages/main/src/plugin/tasks/task-manager.ts
```diff
@@ -47,6 +50,21 @@ export class TaskManager {
...
+        type: 'object',
+        properties: {
+          [`${ExperimentalTasksSettings.SectionName}.${ExperimentalTasksSettings.StatusBar}`]: {
+            description: 'Enable the running tasks to be displayed in the status bar',
```
deboer-tim: A bit long. Maybe just 'Show running tasks in the status bar'?

deboer-tim (commented):
Confirmed suggestions from UX call were implemented. Test failure needs to be fixed. Worked great in testing except that a task with progress (bootc build) still had an indeterminate progress bar. Showed correctly in the task manager so likely reactivity issue in TaskIndicator?

But otherwise, you know I love this. :) It's a huge improvement especially for long-running tasks and can't wait to get it in.

packages/main/src/plugin/tasks/task-manager.ts
axel7083:
```suggestion
        title: 'Tasks',
```

packages/main/src/plugin/tasks/task-manager.ts
axel7083:
```suggestion
            description: 'Show running tasks in the status bar',
```

packages/main/src/plugin/tasks/task-manager.ts
axel7083: > Actually, I am wondering if we should call the section Experimental? Have a single place to group any experimental preferences that we expect will disappear (become default behaviour or be rejected) or change ids in the future.

I like this idea, I will open an issue to discuss

axel7083: https://github.com/containers/podman-desktop/issues/9236

benoitf:
to be merged after 1.13 tag is done

axel7083:
1.13 tag is available :rocket: :tada:

benoitf (approved):
let's go then 🚀  and grab feedback

PR #9187, opened 2024-10-03
fix: set a podman machine as running only when not starting

### LINKED ISSUE #8984: PD displays incorrect podman machine status and causes race condition in some scenarios
### Bug description

It seems that when PD is used to create a podman machine via onboarding workflows, either using the `Set up` button or the `Initialize and start` button on the dashboard, spinning up the podman machine takes much longer, but after some time the status of the podman machine is displayed in PD as `RUNNING`, however in the podman CLI the podman machine status is still in `Currently starting`, if in this state the user closes the PD app and then reopens it, waits until podman machine is displayed as `RUNNING` again in the app and tries to turn off the podman machine from Settings -> Resources the podman machine will permanently freeze.

https://github.com/user-attachments/assets/df849562-a916-4a91-b5c0-f62186fac499

### Operating system

macOS

### Installation Method

### Version

1.12.0

### Steps to reproduce

1. Have no podman machine on the system.
2. Open PD application.
3. Make sure the "Exit on Close" option in settings is enabled.
4. Using the `Set up` or `Initialize and start` buttons create a podman machine with autostart enabled.
5. As soon as the podman machine is shown with status `RUNNING` on the Dashboard, close the PD app.
6. Reopen PD app and wait for podman machine to reach `RUNNING` state on the dashboard.
7. Go into Settings -> Resources and stop the podman machine.
8. Notice that the podman machine will change states from `RUNNING` to `STARTING` and will freeze.

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

The podman machine has 2 fields to indicate its state: `Running` and `Starting`.

The global state of the machine can have 3 values:
- `!Running`
- `Running` && `Starting`
- `Running` && `!Starting`

Previously, the second state was considered as running, but at this point, the `podman machine init` is not finished, it will terminate only when the third state is reached. To help the user know when the `podman machine init` is really finished, this PR changes this second state as `Starting` and not `Running`

Fixes #8984 

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -250,9 +250,9 @@ export async function updateMachines(
     const running = machine?.Running === true;
     let status: extensionApi.ProviderConnectionStatus = running ? 'started' : 'stopped';
 
-    // update the status to starting if the machine is starting but not yet running
+    // update the status to starting if the machine is running but still starting
     const starting = machine?.Starting === true;
-    if (!running && starting) {
+    if (starting) {
       status = 'starting';
     }
 
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -379,7 +379,7 @@ export async function updateMachines(
         provider.updateStatus('installed');
       }
     } else {
-      const atLeastOneMachineRunning = machines.some(machine => machine.Running);
+      const atLeastOneMachineRunning = machines.some(machine => machine.Running && !machine.Starting);
       const atLeastOneMachineStarting = machines.some(machine => machine.Starting);
       // if a machine is running it's started else it is ready
       if (atLeastOneMachineRunning) {
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -76,7 +76,7 @@ let defaultConnectionNotify = !isLinux();
 let defaultMachineMonitor = true;
 
 // current status of machines
-const podmanMachinesStatuses = new Map<string, extensionApi.ProviderConnectionStatus>();
+export const podmanMachinesStatuses = new Map<string, extensionApi.ProviderConnectionStatus>();
 let podmanProviderStatus: extensionApi.ProviderConnectionStatus = 'started';
 const podmanMachinesInfo = new Map<string, MachineInfo>();
 const currentConnections = new Map<string, extensionApi.Disposable>();
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -379,6 +379,12 @@ export async function updateMachines(
         provider.updateStatus('installed');
       }
     } else {
+      /*
+       * The machine can have 3 states, based on `Starting` and `Running` fields:
+       * - !Running && !Starting -> configured
+       * -  Running &&  Starting -> starting
+       * -  Running && !Starting -> ready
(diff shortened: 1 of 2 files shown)


benoitf (approved):
if you could add a comment saying why we're also checking Starting

benoitf (commented):
(I mean based on what you have in the PR's description --> comment inside the code)

cbr7:
@feloy The correct state is now shown on the dashboard, however should the same not be done for the state in Settings -> Resources page? With this change you can get that `Starting` state is shown on the dashboard and `Running` state is shown in Settings -> Resources card

feloy:
> @feloy The correct state is now shown on the dashboard, however should the same not be done for the state in Settings -> Resources page? With this change you can get that `Starting` state is shown on the dashboard and `Running` state is shown in Settings -> Resources card

@cbr7  Good catch, thanks. I was expecting to be fixed in this page as well, but didn't verify. 

It should be fixed now.

benoitf (approved):

benoitf (commented):
are we missing unit test ?

cbr7:
@feloy I just tried it again and now the status is indeed the same, but I'm encountering a weird behavior now, if I create a podman machine using the `Set up` button on the dashboard, after the podman machine is created the app will have navigated to the Settings -> Resources page, instead of the Dashboard where it was before, I did not notice this before, do any of your changes related to screen navigation from dashboard to settings page? Have you noticed the same behavior?

feloy:
> @feloy I just tried it again and now the status is indeed the same, but I'm encountering a weird behavior now, if I create a podman machine using the `Set up` button on the dashboard, after the podman machine is created the app will have navigated to the Settings -> Resources page, instead of the Dashboard where it was before, I did not notice this before, do any of your changes related to screen navigation from dashboard to settings page? Have you noticed the same behavior?

Right, I can see the difference too. I will search what can be the cause.

feloy:
> > @feloy I just tried it again and now the status is indeed the same, but I'm encountering a weird behavior now, if I create a podman machine using the `Set up` button on the dashboard, after the podman machine is created the app will have navigated to the Settings -> Resources page, instead of the Dashboard where it was before, I did not notice this before, do any of your changes related to screen navigation from dashboard to settings page? Have you noticed the same behavior?
> 
> Right, I can see the difference too. I will search what can be the cause.

Finally, this is a bug, but not related to this change. I've created an issue https://github.com/containers/podman-desktop/issues/9206 (reproducible in main branch)

cbr7 (approved):

feloy:
setting as draft, the time to write some unit tests

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -269,7 +269,9 @@ export async function updateMachines(
     if (previousStatus !== status) {
       // notify status change
       listeners.forEach(listener => listener(machine.Name, status));
+      console.log('setting ', machine.Name, status);
```
jeffmaury: Should be removed, no ?

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -269,7 +269,9 @@ export async function updateMachines(
...
       podmanMachinesStatuses.set(machine.Name, status);
+      console.log(podmanMachinesStatuses);
```
jeffmaury: Same

extensions/podman/packages/extension/src/extension.ts
feloy: yes, thanks

Issue #9026, opened 2024-09-23
add a widget in docker compatibility page to display/switch the docker cli context


benoitf:
### Is your enhancement related to a problem? Please describe

adds a way to display data on the current context and switch the docker cli context

### Describe the solution you'd like

implement the design provided by https://github.com/containers/podman-desktop/issues/8338

### Describe alternatives you've considered

I don't know if it should also be added to the status bar like we have a way to change the kubernetes context ?

PR #9197, opened 2024-10-03
chore: add methods to read or set docker contexts

benoitf:
### What does this PR do?
add methods to list and set the docker contexts
there is always a default context even when there is no files (like .docker/config.json or any .docker/context/** files) on the filesystem, it's what we're calling the system socket path.

unit tests provided

related to #9026

### How to test this PR?

### DIFF
--- a/packages/api/src/docker-compatibility-info.ts
+++ b/packages/api/src/docker-compatibility-info.ts
@@ -42,3 +42,16 @@ export interface DockerSocketMappingStatusInfo {
     architecture: string;
   };
 }
+
+export interface DockerContextInfo {
+  name: string;
+  isCurrentContext: boolean;
+  metadata: {
+    description: string;
+  };
+  endpoints: {
+    docker: {
+      host: string;
+    };
+  };
+}
--- a/packages/api/src/docker-compatibility-info.ts
+++ b/packages/api/src/docker-compatibility-info.ts
@@ -43,6 +43,8 @@ export interface DockerSocketMappingStatusInfo {
   };
 }
 
+// handle the context information of the docker contexts
+// https://docs.docker.com/engine/manage-resources/contexts/
 export interface DockerContextInfo {
   name: string;
   isCurrentContext: boolean;
--- a/packages/main/src/plugin/docker/docker-context-handler.spec.ts
+++ b/packages/main/src/plugin/docker/docker-context-handler.spec.ts
@@ -1,0 +1,268 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import * as fs from 'node:fs';
+import { homedir } from 'node:os';
+import { join } from 'node:path';
+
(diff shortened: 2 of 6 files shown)


packages/main/src/plugin/docker/docker-context-handler.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  test('should return context if docker config exist', async () => {
+    vi.spyOn(fs, 'existsSync').mockReturnValue(true);
+    vi.spyOn(fs.promises, 'readFile').mockResolvedValue(JSON.stringify({ currentContext: 'test-context' }));
+
+    const currentContext = await dockerContextHandler.getCurrentContext();
+    expect(currentContext).toBe('test-context');
+  });
+
+  test('should return undefined if docker config exist but fails to parse it', async () => {
```
deboer-tim: Says 'undefined' but returns 'default', I think test title just needs to be updated.

packages/main/src/plugin/docker/docker-context-handler.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+    ]);
+    vi.spyOn(fs.promises, 'readFile').mockResolvedValue('not a json');
+
+    await expect(() => dockerContextHandler.switchContext('foo')).rejects.toThrowError(
+      'Error parsing docker config file',
+    );
+  });
+});
+
+describe('listContexts', () => {
```
deboer-tim, line 146: Missing test for list or getContexts if folder does not exist.

packages/preload/src/index.ts
```diff
@@ -2260,6 +2260,14 @@ export function initExposure(): void {
       return ipcInvoke('docker-compatibility:getSystemDockerSocketMappingStatus');
     },
   );
+
+  contextBridge.exposeInMainWorld('getDockerContexts', async (): Promise<DockerContextInfo[]> => {
+    return ipcInvoke('docker-compatibility:listDockerContexts');
+  });
+
+  contextBridge.exposeInMainWorld('switchDockerContext', async (contextName: string): Promise<DockerContextInfo[]> => {
+    return ipcInvoke('docker-compatibility:switchDockerContext', contextName);
```
deboer-tim, line 2269: Minor: I think I'd prefer setDockerContext. I know there are typically going to be 2, but switch sounds like toggling between two states, where set sounds like picking from the list.

deboer-tim (approved):
Minor comments but approving.

packages/preload/src/index.ts
benoitf: there is always the default state, there is never an unknown state

packages/main/src/plugin/docker/docker-context-handler.spec.ts
benoitf: docker-context-handler.ts           |     100 |      100 |     100 |     100 |

100% coverage in this file now

jeffmaury (commented):
Context are retrieved through the internal implementation dependant Docker specific file. Why not use docker context ls --format json instead ? Might be broken by Docker undocumented change, no ?

benoitf:
>Context are retrieved through the internal implementation dependant Docker specific file. Why not use docker context ls --format json instead ? Might be broken by Docker undocumented change, no ?

To not have a direct dependency on the docker CLI from the Podman Desktop core.
Also docker is really conservative on changes,

packages/api/src/docker-compatibility-info.ts
```diff
@@ -42,3 +42,16 @@ export interface DockerSocketMappingStatusInfo {
     architecture: string;
   };
 }
+
+export interface DockerContextInfo {
```
axel7083, line 48: Could you add some comments on what this object is, and where the values came from? I must admit I feel a bit lost which the docker contexts as I've never used them

benoitf: I'll add a link to https://docs.docker.com/engine/manage-resources/contexts/

axel7083: Perfect thanks

benoitf: ✅

axel7083 (approved):
LGTM Codewise

jeffmaury (approved):
LGTM

PR #9198, opened 2024-10-03
feat: add ability to switch the docker CLI context

### LINKED ISSUE #9026: add a widget in docker compatibility page to display/switch the docker cli context

benoitf:
### What does this PR do?
add ability to switch the docker CLI context

- [x] : Depends on https://github.com/containers/podman-desktop/pull/9197

### Screenshot / video of UI

image

fixes https://github.com/containers/podman-desktop/issues/9026

### How to test this PR?

1. Add `"dockerCompatibility.enabled": true` in `~/.local/share/containers/podman-desktop/configuration/settings.json` file
1. go to the settings / Docker compatibility page
1. see the dropdown to select the docker contexts

### DIFF
--- a/packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.spec.ts
+++ b/packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.spec.ts
@@ -1,0 +1,138 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+
+import { render, screen } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
+import { tick } from 'svelte';
+import { beforeEach, expect, type Mock, test, vi } from 'vitest';
+
+import type { DockerContextInfo } from '/@api/docker-compatibility-info';
+
+import PreferencesDockerCompatibilityDockerContext from '././PreferencesDockerCompatibilityDockerContext.svelte';
+
+const contextInfos = [
+  {
+    name: 'default',
+    isCurrentContext: false,
+    metadata: { description: 'default context' },
+    endpoints: {
+      docker: {
+        host: 'unix:///var/run/docker.sock',
+      },
+    },
+  },
+  {
+    name: 'podman',
+    isCurrentContext: true,
+    metadata: { description: 'podman context' },
+    endpoints: {
+      docker: {
+        host: 'unix:///var/podman.sock',
+      },
+    },
+  },
+];
+
+const getDockerContextsMock: Mock<() => Promise<DockerContextInfo[]>> = vi.fn();
+const switchDockerContextMock: Mock<(contextName: string) => Promise<void>> = vi.fn();
+
+beforeEach(() => {
+  vi.resetAllMocks();
+
+  Object.defineProperty(global, 'window', {
+    value: {
+      navigator: {
+        clipboard: {
+          writeText: vi.fn(),
+        },
+      },
+      getDockerContexts: getDockerContextsMock,
+      switchDockerContext: switchDockerContextMock,
+    },
+    writable: true,
+  });
+});
+
+test('only default context', async () => {
+  getDockerContextsMock.mockResolvedValue([
+    {
+      name: 'default',
+      isCurrentContext: true,
+      metadata: { description: 'default context' },
+      endpoints: {
+        docker: {
+          host: 'unix:///var/run/docker.sock',
+        },
+      },
+    },
+  ]);
+
+  render(PreferencesDockerCompatibilityDockerContext);
+
+  // wait for the promise to resolve
+  await vi.waitFor(() => expect(getDockerContextsMock).toBeCalled());
+  await tick();
+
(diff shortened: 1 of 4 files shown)


benoitf:
rebased as the PR required for this work has been merged

benoitf:
unit tests added

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.svelte
```diff
@@ -0,0 +1,68 @@
...
+
+        <div class="mx-2">
+          <RefreshButton label="Refresh the context" onclick={refreshDockerContext} />
+        </div>
+      </div>
+      {#if dockerContexts.length === 0}
+        <div class="mt-2">No Docker context found</div>
+      {:else}
+        <div class="mt-2">
+          Display and select between your different Docker-compatible socket contexts.
```
deboer-tim: 'Display and select between' is a little odd and verbose to me. Maybe just 'Select the active Docker-compatible socket context'?

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.svelte
```diff
@@ -0,0 +1,68 @@
...
+        Docker Context
+
+        <div class="mx-2">
+          <RefreshButton label="Refresh the context" onclick={refreshDockerContext} />
+        </div>
+      </div>
+      {#if dockerContexts.length === 0}
+        <div class="mt-2">No Docker context found</div>
+      {:else}
+        <div class="mt-2">
```
deboer-tim: Should have text-sm to match Preferences sections. Ditto for 'No context found'.

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.svelte
```diff
@@ -0,0 +1,68 @@
...
+          Display and select between your different Docker-compatible socket contexts.
+          <select
```
deboer-tim: The dropdown looks ugly, but that's a separate issue. :)

It is touching the text though. Needs either some minor vertical padding between them, or switch to the same row layout as Preferences where the dropdown is on the right.

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityRendering.svelte
```diff
@@ -1,10 +1,17 @@
...
 </script>
 
 <SettingsPage title="Docker compatibility">
   <div slot="subtitle" class="text-sm font-thin mt-2">
     Podman Desktop provides compatibility with Docker, allowing you to use your existing Docker commands, images and
     workflows.
   </div>
+  <div class="flex flex-col space-y-4">
+    <div class="container" role="list">
+      Docker CLI Context
```
deboer-tim: To match Preferences should have 'text-lg font-medium'.

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+
+test('only default context', async () => {
+  getDockerContextsMock.mockResolvedValue([
+    {
+      name: 'default',
+      isCurrentContext: true,
+      metadata: { description: 'default context' },
+      endpoints: {
+        docker: {
+          host: 'unix:///var/run/docker.sock',
```
cdrage: should these be const's since it's been used multiple times?

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.svelte
```diff
@@ -0,0 +1,68 @@
...
+  // get current selected context
+  const currentSelectedContext = dockerContexts.find(context => context.isCurrentContext);
+  if (!selectedContext || (currentSelectedContext && selectedContext?.name === currentSelectedContext.name)) {
+    // do nothing if we're not changing the context
+    return;
+  }
+
+  // set the selected context
+  window.switchDockerContext(selectedContext.name).catch((error: unknown) => {
+    console.error(`Failing to switch docket context to ${selectedContext?.name}`, error);
```
cdrage:
```suggestion
    console.error(`Failing switching docker context to ${selectedContext?.name}`, error);
```

cdrage (changes requested):
Was able to switch the contexts fine without having Docker installed (emulated it), did not test it against Docker Desktop however since I do not have a license.

But the code LGTM.

No concerns regarding UX/UI, but I see @deboer-tim reviewed.

Two minor issues regarding code but nothing major.

jeffmaury (approved):
LGTM.

Simple remark: in the case there is a single context, shouldn't we skip the select ?

deboer-tim:
> Simple remark: in the case there is a single context, shouldn't we skip the select ?

I thought the same at first, but I think it's ok to have a select with one option - reinforces that there's only one now and this is where you'd change it.

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.spec.ts
benoitf: applied constant

benoitf:
@cdrage your comments have been fixed

cdrage (approved):

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityDockerContext.svelte
benoitf: I took the text of the Mockup 
image


I changed the text but I think we should try to raise these concerns before during the review of the mockup

benoitf:
@deboer-tim I think your comments have been fixed

deboer-tim (approved):

benoitf:
@jeffmaury yes I think it's showing that it can contain many items

but I'll defer the choice to @ekidneyrh as I extrapolate what I found in the mockup but there was not the case with only one value.

Also we never have the case with "no context" as default context is always there

ekidneyrh:
Having the drop down with only one context is fine - it's a common pattern I believe. It allows the user to see that when / if they add more it'll show up there.

PR #9205, opened 2024-10-04
fix: enable 'Update' for kubectl CLI

### LINKED ISSUE #8838: better distinction between `upgrade` and `change version` in CLI Tools

Given the current UI we can't know if a new version is available. This is a bit problematic IMO. 

image

On this picture for example, kubectl a new version is available, but compose I am already using the latest.

I would like to have a clear, "A new update is available" indicator, and a separate, non-intrusive option to change the current version.

_No response_

### Additional context

_No response_

dgolovin:
### What does this PR do?

PR updates kubectl CLI lifecycle so it shows 'Update' command when current installed version is older than latest available. 

### Screenshot / video of UI

Before: there is always 'Upgrade/Downgrade' visible, but never 'Update', even after startup if there is update, it would not be visible on CLI page.

https://github.com/user-attachments/assets/0ac1c7c7-4db6-4e49-99ed-574d032842fb

After: if there is newer version available, 'Update' link shown.

https://github.com/user-attachments/assets/e4523e7b-a50f-4a07-a09c-2a0fc4f8daa5

Part 1 out of 3 to fix #8838.
Two more fixes for compose and kind should be implemented to close the issue.

### How to test this PR?

Pull PR and try to install/update/upgrade/uninstall kubectl-cli tool.

PR adds:
1. Downloading latest meta for CLI in post activation

Known limitation (probably require new issues to be opened):
1. Install does not install latest kubectl by default, but ask to select specific verison (probably should be additional command `Install Specific version ...`)
2. If latest version is installed, the you will see `Upgrade/Downgrade` (if there is `Install Specific Version ...` command that would not needed
3. If not latest version is installed, then you see 'Update' link that install latest available release and `Upgrade/Downgrade` is not available ( if there is `Install Specific Version ...` always available it would let install any version if latest is not an option)

Known issues (will open follow up as well):
1. version list is not sorted  (at least for kubecli #9221, but should be checked if sorted for other CLIs)
2. version list contains 'Kubernetes' and sorted incorrectly #9221
3. cancel selection of version to install should not show an error, because it it is what user requested #9222
4. `Path` tooltip is empty after installation #9220.

### DIFF
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -382,7 +382,7 @@ async function postActivate(
       releaseVersionToInstall = selected.tag.slice(1);
       return releaseVersionToInstall;
     },
-    doInstall: async () => {
+    doInstall: async (_logger) => {
       if (currentVersion) {
         throw new Error(`Cannot install ${kubectlCliName}. Version ${currentVersion} is already installed.`);
       }
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -340,9 +340,15 @@ async function postActivate(
   // check if there is a new version to be installed and register the updater
   let releaseToUpdateTo: KubectlGithubReleaseArtifactMetadata | undefined;
   let releaseVersionToUpdateTo: string | undefined;
-  const latestAsset = await kubectlDownload.getLatestVersionAsset();
+  let latestAsset: KubectlGithubReleaseArtifactMetadata | undefined;
+  try {
+    latestAsset = await kubectlDownload.getLatestVersionAsset();
+  } catch (error: unknown) {
+    console.error('Error when downloading kubectl CLI latest release information.', String(error));
+  }
+
   const update = {
-    version: latestAsset.tag.slice(1) !== kubectlCliTool.version ? latestAsset.tag.slice(1) : undefined,
+    version: latestAsset?.tag.slice(1) !== kubectlCliTool.version ? latestAsset?.tag.slice(1) : undefined,
     selectVersion: async (): Promise<string> => {
       const selected = await kubectlDownload.promptUserForVersion(currentVersion);
       releaseToUpdateTo = selected;
@@ -369,7 +375,7 @@ async function postActivate(
       if (releaseToUpdateTo === latestAsset) {
         delete update.version;
       } else {
-        update.version = latestAsset.tag.slice(1);
+        update.version = latestAsset?.tag.slice(1);
       }
       releaseVersionToUpdateTo = undefined;
       releaseToUpdateTo = undefined;
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -382,7 +382,7 @@ async function postActivate(
       releaseVersionToInstall = selected.tag.slice(1);
       return releaseVersionToInstall;
     },
-    doInstall: async (_logger) => {
+    doInstall: async _logger => {
       if (currentVersion) {
         throw new Error(`Cannot install ${kubectlCliName}. Version ${currentVersion} is already installed.`);
       }
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -226,11 +226,9 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
   // Push the CLI tool as well (but it will do it postActivation so it does not block the activate() function)
   // Post activation
-  setTimeout(() => {
-    postActivate(extensionContext, kubectlDownload).catch((error: unknown) => {
-      console.error('Error activating extension', error);
-    });
-  }, 0);
+  postActivate(extensionContext, kubectlDownload).catch((error: unknown) => {
+    console.error('Error activating extension', error);
+  });
 }
 
 interface CliFinder {
@@ -391,17 +389,23 @@ async function postActivate(
   // check if there is a new version to be installed and register the updater
   let releaseToUpdateTo: KubectlGithubReleaseArtifactMetadata | undefined;
   let releaseVersionToUpdateTo: string | undefined;
-
-  kubectlCliToolUpdaterDisposable = kubectlCliTool.registerUpdate({
-    selectVersion: async () => {
+  const latestAsset = await kubectlDownload.getLatestVersionAsset();
+  const update = {
+    version: latestAsset.tag.slice(1) !== kubectlCliTool.version ? latestAsset.tag.slice(1) : undefined,
+    selectVersion: async (): Promise<string> => {
       const selected = await kubectlDownload.promptUserForVersion(currentVersion);
       releaseToUpdateTo = selected;
       releaseVersionToUpdateTo = selected.tag.slice(1);
       return releaseVersionToUpdateTo;
     },
-    doUpdate: async _logger => {
+    doUpdate: async (): Promise<void> => {
       if (!releaseToUpdateTo || !releaseVersionToUpdateTo) {
-        throw new Error(`Cannot update ${path}. No release selected.`);
+        if (latestAsset) {
+          releaseToUpdateTo = latestAsset;
+          releaseVersionToUpdateTo = latestAsset.tag.slice(1);
+        } else {
+          throw new Error(`Cannot update ${path}. No release selected.`);
+        }
       }
       // download, install system wide and update cli version
       const binaryPath = await kubectlDownload.download(releaseToUpdateTo);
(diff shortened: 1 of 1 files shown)


extensions/kubectl-cli/src/extension.ts
```diff
@@ -226,11 +226,9 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
   // Push the CLI tool as well (but it will do it postActivation so it does not block the activate() function)
   // Post activation
-  setTimeout(() => {
-    postActivate(extensionContext, kubectlDownload).catch((error: unknown) => {
-      console.error('Error activating extension', error);
-    });
-  }, 0);
+  postActivate(extensionContext, kubectlDownload).catch((error: unknown) => {
+    console.error('Error activating extension', error);
+  });
```
axel7083: Why removing the setTimeout ?

It was used as per the comment above to avoid blocking the activate function, reason to remove ?

benoitf: I would say that as there was no await, it was not blocking anyway the activate function
but yes it seems not related to the scope of this PR

axel7083: I have a preference for the promise without await than the setTimout anyway

benoitf:
the title and the description of the PR is not clear to me

> PR updates kubectl CLI lifecycle

what does it mean ? which part of the lifecycle is changed, what for ?

also it's not clear what is being refactored and what is being added

deboer-tim:
Ditto, it's not clear what behaviour this is changing (which would help to know that this is unrelated/should not fix the known issues).

extensions/kubectl-cli/src/extension.ts
```diff
@@ -337,15 +337,52 @@ async function postActivate(
...
+  };
   kubectlCliToolUpdaterDisposable = kubectlCliTool.registerInstaller({
     selectVersion: async () => {
       const selected = await kubectlDownload.promptUserForVersion(currentVersion);
       releaseToInstall = selected;
       releaseVersionToInstall = selected.tag.slice(1);
       return releaseVersionToInstall;
     },
-    doInstall: async _logger => {
+    doInstall: async () => {
```
dgolovin:
```suggestion
    doInstall: async (_logger) => {
```

extensions/kubectl-cli/src/extension.ts
```diff
@@ -337,15 +337,52 @@ async function postActivate(
...
+  };
   kubectlCliToolUpdaterDisposable = kubectlCliTool.registerInstaller({
     selectVersion: async () => {
       const selected = await kubectlDownload.promptUserForVersion(currentVersion);
       releaseToInstall = selected;
       releaseVersionToInstall = selected.tag.slice(1);
       return releaseVersionToInstall;
     },
-    doInstall: async _logger => {
+    doInstall: async (_logger) => {
```
dgolovin:
```suggestion
    doInstall: async _logger => {
```

extensions/kubectl-cli/src/extension.ts
```diff
@@ -355,11 +392,18 @@ async function postActivate(
       // download, install system wide and update cli version
       const binaryPath = await kubectlDownload.download(releaseToInstall);
       await installBinaryToSystem(binaryPath, 'kubectl');
+      const destinationPath = getSystemBinaryPath('kubectl');
```
dgolovin:
```suggestion
```

extensions/kubectl-cli/src/extension.ts
```diff
@@ -355,11 +392,18 @@ async function postActivate(
...
       kubectlCliTool?.updateVersion({
         version: releaseVersionToInstall,
         installationSource: 'extension',
+        path: destinationPath,
```
dgolovin:
```suggestion
```

extensions/kubectl-cli/src/extension.ts
```diff
@@ -355,11 +392,18 @@ async function postActivate(
...
       });
       currentVersion = releaseVersionToInstall;
+      if (releaseToInstall === latestAsset) {
+        delete update.version;
+      } else {
+        update.version = latestAsset.tag.slice(1);
+      }
```
dgolovin:
```suggestion
```

dgolovin:
Removed change for different issue #9220.

extensions/kubectl-cli/src/extension.ts
```diff
@@ -337,7 +337,44 @@ async function postActivate(
   let releaseToInstall: KubectlGithubReleaseArtifactMetadata | undefined;
   let releaseVersionToInstall: string | undefined;
   let currentVersion = kubectl.version;
-
+  // check if there is a new version to be installed and register the updater
+  let releaseToUpdateTo: KubectlGithubReleaseArtifactMetadata | undefined;
+  let releaseVersionToUpdateTo: string | undefined;
+  const latestAsset = await kubectlDownload.getLatestVersionAsset();
```
dgolovin: Loading for latest available version is added.

extensions/kubectl-cli/src/extension.ts
```diff
@@ -337,7 +337,44 @@ async function postActivate(
...
+  const update = {
+    version: latestAsset.tag.slice(1) !== kubectlCliTool.version ? latestAsset.tag.slice(1) : undefined,
```
dgolovin: To show 'Update' for CLI tool if current version is older than latest available the 'version' property should be set.

extensions/kubectl-cli/src/extension.ts
```diff
@@ -337,7 +337,44 @@ async function postActivate(
...
+    selectVersion: async (): Promise<string> => {
+      const selected = await kubectlDownload.promptUserForVersion(currentVersion);
+      releaseToUpdateTo = selected;
+      releaseVersionToUpdateTo = selected.tag.slice(1);
+      return releaseVersionToUpdateTo;
+    },
+    doUpdate: async (): Promise<void> => {
+      if (!releaseToUpdateTo || !releaseVersionToUpdateTo) {
```
dgolovin, line 359: In case of update, the version being installed is defined in 'version' attribute above and 'selectVerison' function call is going to be skipped on UI side, hence `releaseToUpdateTo` and `releaseVersionToUpdateTo` are empty and that the means the update is requested.

extensions/kubectl-cli/src/extension.ts
```diff
@@ -337,7 +337,44 @@ async function postActivate(
...
+      }
+      // download, install system wide and update cli version
+      const binaryPath = await kubectlDownload.download(releaseToUpdateTo);
+      await installBinaryToSystem(binaryPath, 'kubectl');
+      kubectlCliTool?.updateVersion({
+        version: releaseVersionToUpdateTo,
+        installationSource: 'extension',
+      });
+      currentVersion = releaseVersionToUpdateTo;
+      if (releaseToUpdateTo === latestAsset) {
```
dgolovin, line 375: If latest release installed `version` property is removed to switch card to show `Upgrade/Downgrade` command.

extensions/kubectl-cli/src/extension.ts
```diff
@@ -337,7 +337,44 @@ async function postActivate(
...
+        delete update.version;
+      } else {
+        update.version = latestAsset.tag.slice(1);
```
dgolovin: Or if `Upgrade/Downgrade` is performed, set latest version to `version` property so CLI card show `Update` command.

extensions/kubectl-cli/src/extension.ts
```diff
@@ -337,7 +337,44 @@ async function postActivate(
   let releaseToInstall: KubectlGithubReleaseArtifactMetadata | undefined;
   let releaseVersionToInstall: string | undefined;
   let currentVersion = kubectl.version;
-
+  // check if there is a new version to be installed and register the updater
+  let releaseToUpdateTo: KubectlGithubReleaseArtifactMetadata | undefined;
+  let releaseVersionToUpdateTo: string | undefined;
+  const latestAsset = await kubectlDownload.getLatestVersionAsset();
+  const update = {
```
dgolovin, line 350: Have to move to separate object to be able to update version property, to control card showing ether `Update` or `Upgrade/Downgrade` commands.

dgolovin:
I removed not related code, opened issues where appropriate and added some comments.

deboer-tim (approved):
Worked in testing. Thanks for clarifying what is being fixed here, and making sure we have followup issues for improving unrelated usability issues.

benoitf (approved):

extensions/kubectl-cli/src/extension.ts
dgolovin: I think there is a problem here. getLatestVersioAsset call can throw exception for various reasons:
1. github API is down
2. network is down
3. github API rate limit exceeded

In this implementation that would lead to cli tool update is not registered at all. To get out of it restart would be required.

benoitf (approved):

extensions/kubectl-cli/src/extension.ts
dgolovin: I've added try catch and checks for not initializes `latestAsset`. It prevents from failure to set updater and installer in case of  problems listed above and let at least use `Install`, `Upgrade/Downgrade` without need to restart podman desktop.

deboer-tim (approved):

PR #9219, opened 2024-10-04
fix: remove timeout for async

dgolovin:
### What does this PR do?

Removes timeout for async call, which is not required, because async call returns promise without blocking extension activation.
Used Error to String type coercion to convert error to string for the message

Extracted from #9205.

### How to test this PR?

### DIFF
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -226,11 +226,9 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
   // Push the CLI tool as well (but it will do it postActivation so it does not block the activate() function)
   // Post activation
-  setTimeout(() => {
-    postActivate(extensionContext, kubectlDownload).catch((error: unknown) => {
-      console.error('Error activating extension', error);
-    });
-  }, 0);
+  postActivate(extensionContext, kubectlDownload).catch((error: unknown) => {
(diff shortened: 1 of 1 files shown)


benoitf (approved):

extensions/kubectl-cli/src/extension.ts
```diff
@@ -226,11 +226,9 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
   // Push the CLI tool as well (but it will do it postActivation so it does not block the activate() function)
   // Post activation
-  setTimeout(() => {
-    postActivate(extensionContext, kubectlDownload).catch((error: unknown) => {
-      console.error('Error activating extension', error);
-    });
-  }, 0);
+  postActivate(extensionContext, kubectlDownload).catch((error: unknown) => {
+    console.error('Error activating extension', String(error));
```
jeffmaury: Seems to be that you are losing the stack trace, is it intentional

dgolovin: It is actually produces the same output for dev tools console.

dgolovin:
```suggestion
    console.error('Error activating extension', error);
```

deboer-tim (approved):

jeffmaury (approved):
LGTM

PR #9225, opened 2024-10-06
chore: do not display any release notes widget if there is nothing to show

### LINKED ISSUE #9224: Do not display release notes if they're unreachable

In air gapped mode or without internet connection we see the widget
Image

but as there is no possible action (like a refresh button) or something else, we see this message which has not a great value.

I'm wondering if in this case, the widget should not be hidden.

Unreachable = no display of the release notes instead of displaying a widget with an error from where we can't recover

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
If we don't have internet connection, it was saying "check later" but there is no way to trigger the refresh, etc.

so remove the display of the widget in this case.

fixes https://github.com/containers/podman-desktop/issues/9224

### How to test this PR?

Launch Podman Desktop without an internet connection.


deboer-tim (approved):
If I started when I temporarily didn't have wifi, I wouldn't know I missed this or have any way to refresh; once again makes me feel like a pattern where we lock down the major release notes bullets earlier and include _in_ the release would be better. But, barring that this is an improvement and no concerns with the code.

PR #9238, opened 2024-10-07
docs(website): added a tutorial for interacting with a database server

shipsing:
Signed-off-by: Shipra Singh <shipsing@redhat.com>

### What does this PR do?
Add a tutorial to interact with a database server from within the Podman Desktop UI
### Screenshot / video of UI

PR preview: https://6708e0a5eeef3afe65732022--podman-desktop-pr.netlify.app/tutorial/interacting-with-a-database-server
https://github.com/containers/podman-desktop/issues/8871

### How to test this PR?


website/tutorial/interacting-with-a-database-server.md
```diff
@@ -0,0 +1,71 @@
...
+
+This tutorial focuses on creating a MySQL server instance by pulling a MySQL server image from the _quay.io_ registry.
+
+## Before you begin
+
+Make sure you have:
+
+- [Installed Podman Desktop application](/docs/installation).
+- [A running Podman machine](/docs/podman/creating-a-podman-machine).
+- [Configured the _quay.io_ registry](/docs/containers/registries) to pull a MySQL Server image, such as `quay.io/fedora/mysql-80`.
```
jeffmaury: You don't need to configure a registry if you only pull images from

website/tutorial/interacting-with-a-database-server.md
```diff
@@ -0,0 +1,71 @@
...
+description: Interacting with databases from the Podman Desktop application
+keywords: [podman desktop, podman, databases]
+tags: [podman-desktop, interacting-with-databases]
+---
+
+# Interacting with a database server
+
+This tutorial covers the following end-to-end tasks required to interact with a database server from within the Podman Desktop UI:
+
+- Pulling a SQL server image
```
jeffmaury: SQL Server is different from MySQL

jeffmaury (commented):
I'm wondering whether MySQL is a good choice when everybody now speak about Postgresql

Also I think there is a missing piece on how to access the database from an application (Java, NodeJS,...)

shipsing:
Thank you, @jeffmaury, for your suggestions. If both community and Red Hat-based users prefer to use PodstgreSQL in place of MySQl, then we can definitely replace the tutorial example.

website/tutorial/interacting-with-a-database-server.md
```diff
@@ -0,0 +1,71 @@
...
+This tutorial focuses on creating a MySQL server instance by pulling a MySQL server image from the _quay.io_ registry.
+
+## Before you begin
+
+Make sure you have:
+
+- [Installed Podman Desktop application](/docs/installation).
+- [A running Podman machine](/docs/podman/creating-a-podman-machine).
+- [Configured the _quay.io_ registry](/docs/containers/registries) to pull a MySQL Server image, such as `quay.io/fedora/mysql-80`.
+- A developer role.
```
gastoner, line 31: What this means? Like in quay.io?

gastoner (commented):
I would probably add the left nav bar, I think the user could make use of this and navigate through the PD using the "images" and "container" icons. 

Like so: 
image

I think you need to rebase/ download newest version, the cpu/memory usage indicator should now now overlap with the content below (like in picture)

website/tutorial/interacting-with-a-database-server.md
shipsing: @jeffmaury, Without configuring the registry, we cannot pull any images. I got access denied error when i tried to do so. the message states registry requires authentication.

shipsing: removed this prerequisite

website/tutorial/interacting-with-a-database-server.md
shipsing: I thought that a user requires to configure this registry before pulling an image --- as a prerequisite. But, this is not the case. So, i removed the bullet point.

shipsing:
Thank you @gastoner for the suggestion. Most of the times, a user automatically navigates to a specific component page as part of the procedural flow. Whenever any manual intervention is required, we can add a step (like click Containers in the left navigation pane) in the procedure to complete the flow. So, no need to capture the entire left nav in the screenshot. 
It is a good practice to capture only the relevant part in the screenshot to keep the reader focused on what is happening with each step. Also, the components you see in the left nav might vary based on your Podman Desktop configuration (whether you have set up any extension or a kube cluster). This might confuse readers. We also need to think from maintenance perspective. If any major change happens in the left nav, we might need to update all the screenshots.

shipsing:
@jeffmaury @gastoner, Could you please verify the PR and confirm? Thanks

website/tutorial/interacting-with-a-database-server.md
gastoner: I was thinking the "developer role" (I should specify better), I would imagine that the developer role has something to do with the quay.io, but since it is now removed, I think that this should be removed too

shipsing: I see the point of concern. I kept the fragmented part in the beginning. @gastoner, the complete prerequisite reads as `Make sure you have a developer role.`  this is basically clarifying for whom (audience) this content is for or who is the audience for this content? Should i add Database Administrators as well. Something like developer or DBA role??

shipsing:
@jeffmaury, Could you please verify the PR and confirm? Thanks

website/tutorial/interacting-with-a-database-server.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Creating a PostgreSQL server instance
+
+1. Click **Images** in the left navigation pane.
+1. Click the **Run Image** icon corresponding to the PostgreSQL server image you want to run.
+   ![running an image](img/run-image-icon.png)
+1. Configure the basic details, such as container name, port mapping, and environment variables for the PostgreSQL server instance.
+1. Click **Start Container**.
+   ![starting a container](img/starting-a-postgresql-container.png)
+1. View the successful operation notification in the _Tty_ tab of the _Container Details_ page.
```
jeffmaury, line 54: I'm wondering why the Tty tab is used and not the Logs one ?

jeffmaury (approved):
LGTM.

I'm missing the integration into external applications (NodeJS, Java, Python) but could be a follow up PR

website/tutorial/interacting-with-a-database-server.md
shipsing: Because when you perform the actual procedure, this tty tab automatically opens after you click `Start Container` (the previous step). The tty tab shows that the database server has started successfully.

shipsing:
@jeffmaury, I will create a follow-up PR. Merging that integration piece in this tutorial might make the content a bit complex. I might set up a quick call with you to understand what would be the procedural flow so that i can better plan the content? Thanks a lot.

PR #9239, opened 2024-10-07
chore: add jest-dom types

benoitf:
### What does this PR do?
we can track the problem with newer vitest dependencies

assertions like toBeInTheDocument, etc should be declared as types

### How to test this PR?


axel7083 (approved):
LGTM

PR #9274, opened 2024-10-08
fix: reset extra assets array in beforePack call to avoid including podman vm image for x64 and arm64 in arm64.zip

### LINKED ISSUE #9276: Mac OS Airgapped arm64 installer bundles also x64 artefacts
### Bug description

There is a problem with electron-builder config, which bundles the artifacts cumulatively, and this arm64 build not contains x64 artifacts as well. Result is airgapped dmg and zip files over 2 GB that cannot be uploaded to a github releases.

### Operating system

MacOS

### Installation Method

### Version

1.13.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

dgolovin:
This fix avoid including x64 and arm64 podman vm images to amd64 build

### What does this PR do?

PR removes const array and init extra assets array with new one in every beforfPack call.

### How to test this PR?

### DIFF
--- a/.electron-builder.config.cjs
+++ b/.electron-builder.config.cjs
@@ -79,14 +79,16 @@ const config = {
   buildDependenciesFromSource: false,
   npmRebuild: false,
   beforePack: async context => {
-    context.packager.config.extraResources = ['packages/main/src/assets/**'];
+    const DEFAULT_ASSETS = ['packages/main/src/assets/**'];
+    context.packager.config.extraResources = DEFAULT_ASSETS;
+
     // universal build, add both pkg files
     // this is hack to avoid issue https://github.com/electron/universal/issues/36
     if (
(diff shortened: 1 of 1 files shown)


odockal:
Fixes #9276

odockal:
Now I can produce these installers on Mac OS using `export AIRGAP_DOWNLOAD=true;pnpm compile:next`:
Screenshot_20241008_222930

Works for me.

odockal (approved):

benoitf (approved):
I think it would request additional comments in the code to say to not use a constant as the array will be reused at two different places with different values, etc

so that we don't do that again

or if you move the constant to beforePack maybe it's solving the issue (as the issue is not within the beforePack call but just that we're calling multiple times beforePack

dgolovin:
I moved constant inside beforePack.

PR #9279, opened 2024-10-08
 fix: reset extra assets array in beforePack call to avoid including podman vm image for x64 and arm64 in arm64.zip #9274 

dgolovin:
This fix avoid including x64 and arm64 podman vm images to amd64 build

### What does this PR do?

PR moves DEFAULT_ASSETS const inside beforPack callback to avoid including both x64 and arm64 podman vm images into arm64 podman desktop airgap zip.

Original PR for main branch #9274.

### How to test this PR?

### DIFF
--- a/.electron-builder.config.cjs
+++ b/.electron-builder.config.cjs
@@ -65,8 +65,6 @@ async function addElectronFuses(context) {
   });
 }
 
-const DEFAULT_ASSETS = ['packages/main/src/assets/**'];
-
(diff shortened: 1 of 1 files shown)


SoniaSandler (approved):

deboer-tim (approved):
Would be nice to reference the issue and PR on main too.

benoitf (approved):

PR #9306, opened 2024-10-09
fix: Done button in DeployPodToKube

SoniaSandler:

### What does this PR do?
Fixes the Done button in the DeployPodToKube page so that it would go back to the previous page

https://github.com/containers/podman-desktop/issues/9295

### How to test this PR?
Go to the DeployPodToKube page, finish the pod deployment process, click the Done button, and check that it goes back to the previous page

### DIFF
--- a/packages/renderer/src/lib/pod/DeployPodToKube.spec.ts
+++ b/packages/renderer/src/lib/pod/DeployPodToKube.spec.ts
@@ -29,6 +29,8 @@ import { tick } from 'svelte';
 import { router } from 'tinro';
 import { beforeEach, expect, test, vi } from 'vitest';
 
+import { lastPage } from '/@/stores/breadcrumb';
+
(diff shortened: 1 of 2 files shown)


benoitf (approved):
if you can add a unit test

deboer-tim (approved):
LGTM, same fix I was about to do. Was tempted to say it should always go to Pods page, but that's a separate issue.

Tested launching from Pods page or Pod Details, worked from both. We should really have a unit test here too, but since a unit test wouldn't have avoided this regression I'm approving anyway.

deboer-tim:
Got to this and #9307 out of order, agree with Florent to merge the other and take more time to write unit test for this (even if unit test might not have caught this, there is value).

SoniaSandler:
@benoitf @deboer-tim added tests

benoitf (approved):

benoitf:
thanks

PR #9325, opened 2024-10-10
chore: reduce width and font size subnavigation

### LINKED ISSUE #9299: Implementation of smaller subnavigation menu

See: https://github.com/containers/podman-desktop/issues/8771

Implementation of the finalization of UX/UI  with regards to smaller navigation menu for Kubernetes / AI lab / etc.

_No response_

### Additional context

_No response_

cdrage:
chore: reduce width and font size subnavigation

### What does this PR do?

Reduces the width and font size of the subnavigation for Kubernetes and
settings menus (and other components using the UI).

It does the following:
- Reduces from large to medium font
- Reduces the width of navigation menu
- Decreases spacing between the section name (Kubernetes / Preferences)
  and menu items to match similarly to how it was before with large
  font.

### Screenshot / video of UI

Before:

Screenshot 2024-10-10 at 12 21 50 PM
Screenshot 2024-10-10 at 12 21 54 PM

After:
Screenshot 2024-10-10 at 12 20 49 PM
Screenshot 2024-10-10 at 12 20 53 PM

Closes https://github.com/containers/podman-desktop/issues/9299

### How to test this PR?

N/A, minor UI change

### DIFF
--- a/tailwind.config.cjs
+++ b/tailwind.config.cjs
@@ -33,11 +33,11 @@ module.exports = {
       },
       width: {
         leftnavbar: '48px',
-        leftsidebar: '225px',
+        leftsidebar: '170px',
       },
       minWidth: {
         leftnavbar: '48px',
-        leftsidebar: '225px',
+        leftsidebar: '170px',
       },
     },
     fontSize: {
--- a/packages/renderer/src/PreferencesNavigation.svelte
+++ b/packages/renderer/src/PreferencesNavigation.svelte
@@ -56,9 +56,9 @@ onMount(async () => {
   class="z-1 w-leftsidebar min-w-leftsidebar flex-col justify-between flex transition-all duration-500 ease-in-out bg-[var(--pd-secondary-nav-bg)] border-[var(--pd-global-nav-bg-border)] border-r-[1px]"
   aria-label="PreferencesNavigation">
   <div class="flex items-center">
-    <div class="pt-4 px-3 mb-10">
+    <div class="pt-4 px-3 mb-5">
       <p
-        class="text-2xl font-semibold text-[color:var(--pd-secondary-nav-header-text)] border-l-[4px] border-transparent">
+        class="text-xl font-semibold text-[color:var(--pd-secondary-nav-header-text)] border-l-[4px] border-transparent">
         Settings
       </p>
     </div>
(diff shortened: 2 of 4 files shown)


cdrage:
cc @ekidneyrh @benoitf @deboer-tim

tailwind.config.cjs
```diff
@@ -33,11 +33,11 @@ module.exports = {
       },
       width: {
         leftnavbar: '48px',
-        leftsidebar: '225px',
+        leftsidebar: '170px',
       },
       minWidth: {
         leftnavbar: '48px',
-        leftsidebar: '225px',
+        leftsidebar: '170px',
```
benoitf, line 40: not directly related to this PR, but it makes me feeling that maybe the dimension of the sidebar should not be part of the tailwind configuration but of the component or the core (like the color-registry)

where the value are then being injected into the CSS

because if we change the value there, probably that AI lab would have to change it in their fork of the tailwind config while we might think that the width is set by Podman Desktop

cdrage: @benoitf should AI lab plan to change it in their extensions too / implement the navigation bar? i'm assuming it's a separate svelte file for them?

would it be suitable to include the pixel width in the color-registry file so it's easily retrievable like everything else? unsure if this should be a separate PR or not

benoitf: (it's just thoughts there, that I wanted to share but it won't hold the PR)

benoitf: but yes it's defined there:
https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/frontend/tailwind.config.cjs#L51-L57

but still having the item being defined there https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/frontend/src/lib/Navigation.svelte#L5

so probably enhancement need to be done on the component's side

cdrage: Ahh that makes sense. Let's open up an issue regarding this so we can help standardize this. I know we've been discussing webviews as well as subnavigation, so I'm unsure what the result of that discussion may impact this either.

ekidneyrh:
Looks great! Just one nit - can we center the arrow on the preferences menu item so its inline with the text

image

cdrage:
> Looks great! Just one nit - can we center the arrow on the preferences menu item so its inline with the text
> 
> image

Done!

Screenshot 2024-10-11 at 8 53 46 AM

Added `top-0.5` and it centered (2px change, which corresponds to the change in font).

Keeping the `absolute` value in there (not using relative) so that the transitions work fine still.

axel7083:
In AI Lab here is what it would look likes

| Before | After |
| --- | --- |
| image | image |

gastoner (approved):
LGTM

SoniaSandler (approved):
LGTM

PR #9336, opened 2024-10-11
chore: switch from prettier to biome for formatting the files

### LINKED ISSUE #8970: formatter: investigate if we can switch from prettier to biome

in Podman Desktop we have a lot of files, would be nice if we can format and lint the files faster.

investigate if we can use biomejs for the formatter.

Can also check if we could move some eslint rules to biomejs

oxc formatter

maybe biome for the formatting and oxc for the linting or a mix of the two.

### Additional context

_No response_

benoitf:
### What does this PR do?
use a faster formatter and handling more files by default out of the box
it can format the entire repository in few ms (without any cache)

for review, there are 2 commits, so you might look at each one separately
- one replacing prettier by biome
- one applying then the formatter (and then there are few formatting changes from prettier)
   mostly on JSON files or with some playwright tests. I can't find any rules to make the exact formatting

fixes https://github.com/containers/podman-desktop/issues/8970

### How to test this PR?


gastoner (approved):
Tested with `pnpm format:check/fix` commands on Linux

jeffmaury (commented):
I am getting an error on fprmat:check just after checkout of this PR.

storybook\.storybook\themes.css

benoitf:
@jeffmaury and if you run format:fix what is the change ?

benoitf:
@jeffmaury should be good now, there was some missing folders to exclude (the files being generated like the storybook css and the .ui-sveltekit directory for ui package)

axel7083 (approved):
`Checked 1600 files in 266ms. No fixes applied.`

Omg, I'm.... speecheless, can we replace eslint ?

> biome is made in rust
> image
> Rust <3 let's rewrite pd in rust

jeffmaury:
> @jeffmaury and if you run format:fix what is the change ?

Seems related to my local clone. Using a brand new clone works well.

FYI: format:fix reports a single file being updated but does not get listed by git status

jeffmaury (approved):
LGTM

benoitf:
@axel7083 the idea is that we could switch some eslint rules to biome (because it's not only a formatter but also a linter)

and yes rust tooling is very fast compare to go/javascript

one problem is that there is no plug-in for the rust tools.

benoitf:
also in the issue I mentioned that there is also oxlint

https://github.com/oxc-project/oxc?tab=readme-ov-file#%EF%B8%8F-performance

but for now I don't know what would be the best combo to replace some of our eslint rules
need to check the eslint rules consuming most of the time

PR #9366, opened 2024-10-14
fix: call onUpdate callback if configuration is updated programmatically

### LINKED ISSUE #9194: Configuration change initiated by extension API does not send update event
### Bug description

If extension uses `configuration.update` to modify configuration, no event is sent to the listeners registered through the `onDidUpdateConfiguration` method

### Operating system

Win11Pro

### Installation Method

None

### Version

1.12.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
Fix the fact that the emitter was not called

fixes https://github.com/containers/podman-desktop/issues/9194

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/configuration-registry.spec.ts
+++ b/packages/main/src/plugin/configuration-registry.spec.ts
@@ -276,17 +276,13 @@ test('addConfigurationEnum with a previous default value', async () => {
 
 describe('should be notified when a configuration is updated', async () => {
   test('expect correct properties', async () => {
-    let called = false;
-    let updatedProperties: string[] = [];
-    configurationRegistry.onDidUpdateConfiguration(e => {
-      called = true;
-      updatedProperties = e.properties;
-    });
+    const listener = vi.fn();
+    configurationRegistry.onDidUpdateConfiguration(listener);
     const config = configurationRegistry.getConfiguration('my.fake.property', 'myValue');
     await config.update('myKey', 'myValue');
 
-    expect(called).toBeTruthy();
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/configuration-registry.spec.ts
```diff
@@ -273,3 +273,20 @@ test('addConfigurationEnum with a previous default value', async () => {
...
 });
+
+describe('should be notified when a configuration is updated', async () => {
+  test('expect correct properties', async () => {
+    let called = false;
+    let updatedProperties: string[] = [];
+    configurationRegistry.onDidUpdateConfiguration(e => {
+      called = true;
+      updatedProperties = e.properties;
+    });
```
axel7083:
```suggestion
    const subscriberMock = vi.fn();
    configurationRegistry.onDidUpdateConfiguration(subscriberMock);
    
    ...
    
    expect(subscriberMock).toHaveBeenCalledWith(...)
```

Would something like this be simpler than having to create boolean and property object to check it has been called ?

benoitf: I updated the code, it was using a copy/paste for the previous test for onDidChangeConfiguration

axel7083 (approved):
LGTM

packages/main/src/plugin/configuration-registry.spec.ts
axel7083: :+1:

jeffmaury (approved):
LGTM

PR #9371, opened 2024-10-15
fix: release notes banner update button

### LINKED ISSUE #9352: Dashboard panel tells to update when there is no update required
### Bug description

Image

Image

Image

### Operating system

Windows 11

### Installation Method

### Version

1.13.2

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

When clicking the version below, it works as expected
Image

### LINKED ISSUE #9365: podman wants to update to same version
### Bug description

Image

### Operating system

macos silicon

### Installation Method

None

### Version

1.13.0

### Steps to reproduce

got notice about update availalbe - ran it and now podman desktop shows up asking to update v.1.13.2 to v.1.13.2 ...and when clicking update it reports error and to look in dev console.

under thebell icon I have:
`Unable to download v1.13.2 update: Error: Please check update first`

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
Fixes the release notes banner update button bug where the update button still shows up even after updating to the latest release.

Closes https://github.com/containers/podman-desktop/issues/9365
Closes https://github.com/containers/podman-desktop/issues/9352

### How to test this PR?

### DIFF
--- a/packages/renderer/src/stores/update-store.ts
+++ b/packages/renderer/src/stores/update-store.ts
@@ -27,9 +27,9 @@ try {
   console.log('Cannot check for update');
 }
 
-export const updateAvailable = await setup();
+export const updateAvailable = setup();
 
-export async function setup(): Promise<Writable<boolean>> {
+export function setup(): Writable<boolean> {
   const store = writable(podmanDesktopUpdateAvailable);
 
   window.events?.receive('app-update-available', isUpdateAvailable => {
--- a/packages/renderer/src/stores/update-store.ts
+++ b/packages/renderer/src/stores/update-store.ts
@@ -16,29 +16,41 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import type { Writable } from 'svelte/store';
 import { writable } from 'svelte/store';
 
-let podmanDesktopUpdateAvailable = false;
+import { EventStore } from './event-store';
 
-try {
-  podmanDesktopUpdateAvailable = await window.podmanDesktopUpdateAvailable();
-} catch (e) {
-  console.log('Cannot check for update');
+export const updateAvailable = writable(false);
+
+const windowEvents = ['app-update-available'];
+
+const windowListeners = ['extensions-already-started'];
+
+export async function checkForUpdate(eventName: string): Promise<boolean> {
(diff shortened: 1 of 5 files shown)


packages/renderer/src/stores/update-store.ts
```diff
@@ -0,0 +1,44 @@
...
+  window.events?.receive('app-update-available', isUpdateAvailable => {
+    if (isUpdateAvailable) {
+      updateAvailable.set(true);
+    } else {
+      updateAvailable.set(false);
+    }
+  });
+
+  return store;
+}
```
benoitf: if we go with a store, it should be like the other one and use the `new EventStore<....>` pattern (where you can add events that you're interested in)

you might look at existing stores importing `EventStore`

packages/renderer/src/stores/update-store.ts
```diff
@@ -0,0 +1,56 @@
...
+  } else if ('app-update-available' === eventName) {
+    return true;
+  }
+  return false;
+}
+
+const isUpdateAvailable = (...args: unknown[]): Promise<boolean> => {
+  const eventArg = args.length > 0 ? args[0] : false;
+  if (typeof eventArg === 'boolean') {
+    return Promise.resolve(eventArg);
```
benoitf: here we can't use async keyword on the function ?
so we don't need to use Promise resolve but just returning the value ?

SoniaSandler: @benoitf the type of the updater function in `event-store` is `private updater: (...args: unknown[]) => Promise<T>`, so if I don't do the promise resolve is gives me a type error

benoitf: but if you set this function as being 'async' ?

gastoner (approved):

benoitf (approved):

SoniaSandler:
@benoitf just double checking, are the ` onUpdateNotAvailable` and ` onUpdateAvailable` events happen on each start of the app or also only every 12 hours?
In a any case, the ` return this.#updateCheckResult !== undefined && this.#currentVersion !== this.#nextVersion;` check should solve this issue

Issue #9156, opened 2024-10-01
Kubernetes cannot recover after losing the connection


axel7083:
### Bug description

While working on https://github.com/containers/podman-desktop/issues/8341 I noticed a problem when podman desktop lose the connection it is **never able** to recover (on branch main 01/10/2024).

https://github.com/user-attachments/assets/577789f3-296e-4031-a3a1-2f63c15902f3

### Operating system

windows 11

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. create kind cluster
2. assert node visible in kubernetes page
3. stop kind container
4. assert connection lost
5. start kind container 
6. assert nothing is hapening

The only way to reconnect is to close podman desktop and restart it.

### Relevant log output

```shell
17:36:19 [main] Trying to watch pods on the kubernetes context named "kind-kind-cluster" but got a connection refused, retrying the connection in 18s. Error: Forbidden)

17:37:22 [main] Kube deployment watch ended AbortError: The user aborted a request.
17:37:28 [main] Trying to watch routes on the kubernetes context named "kind-kind-cluster" but got a connection refused, retrying the connection in 84s. Error: HTTP-Code: 404
Message: Unknown API Status Code!
Body: "404 page not found\n"
Headers: {"audit-id":"51677f0f-4cab-48ab-9825-6ee0a846f25c","cache-control":"no-cache, private","connection":"close","content-length":"19","content-type":"text/plain; charset=utf-8","date":"Tue, 01 Oct 2024 15:37:28 GMT","x-content-type-options":"nosniff","x-kubernetes-pf-flowschema-uid":"1cc99633-0cfc-4d6b-8aa5-142664664a08","x-kubernetes-pf-prioritylevel-uid":"996d21c0-4bbf-42df-b078-33bcf6bfab1c"})

17:38:52 [main] Trying to watch routes on the kubernetes context named "kind-kind-cluster" but got a connection refused, retrying the connection in 84s. Error: HTTP-Code: 404
Message: Unknown API Status Code!
Body: "404 page not found\n"
[...]

PR #9380, opened 2024-10-15
fix: do not add again an existing Kubernetes resource

feloy:

### What does this PR do?

Adds a debug log when a Kubernetes resource is added again to the state, and prevent from having the same resource displayed several times in the frontend.

This should not happen, as this should be handled by the informer, but it some circumstances it actually happens. This will help debug why it happens.

Part of #9156 

### How to test this PR?

Start, stop clusters, change current context, etc.

When navigating to the Kubernetes resources pages, the resources should not be duplicated.

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-manager.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager.ts
@@ -392,8 +392,8 @@ export class ContextsManager {
           currentContext: this.kubeConfig.currentContext,
           resources: { pods: true },
           update: state => {
-            if (state.resources.pods.filter(o => o.metadata?.uid !== obj.metadata?.uid).length) {
-              console.debug(`==> resource ${obj.metadata?.name} already added`);
+            if (state.resources.pods.some(o => o.metadata?.uid !== obj.metadata?.uid)) {
+              console.debug(`pod ${obj.metadata?.name} already added in context ${this.kubeConfig.currentContext}`);
             }
             state.resources.pods = state.resources.pods.filter(o => o.metadata?.uid !== obj.metadata?.uid);
             state.resources.pods.push(obj);
--- a/packages/main/src/plugin/kubernetes/contexts-manager.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager.ts
@@ -391,7 +391,13 @@ export class ContextsManager {
           sendGeneral: true,
           currentContext: this.kubeConfig.currentContext,
           resources: { pods: true },
-          update: state => state.resources.pods.push(obj),
+          update: state => {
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -386,7 +386,13 @@ export class ContextsManager {
           sendGeneral: true,
           currentContext: this.kubeConfig.currentContext,
           resources: { pods: true },
-          update: state => state.resources.pods.push(obj),
+          update: state => {
```
axel7083, line 394: update is maybe juste updating the content of a pod ? like status or deleteTimestamps or idk which property. We should replace existing one with the new one provided.

Maybe state.resources.pods should be a Map<string, Pod> instead of an array, so we can quickly and efficiently update (replace, delete) a pod by its id.

feloy: Here, we are in the `onAdd` callback, when the informer sends an `ADD` event. In the case the resource is updated, this is handled below, in the `onUpdate` callback.

The map will be optimised for replacing objects, but as we want to send an array to the frontend, the extraction of the array from the map every time we want to send it won't be optimised anymore. I would prefer to keep this optimisation.

packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -386,7 +386,13 @@ export class ContextsManager {
...
+            if (state.resources.pods.filter(o => o.metadata?.uid !== obj.metadata?.uid).length) {
+              console.debug(`==> resource ${obj.metadata?.name} already added`);
```
benoitf: I would for consistency, not use any prefix like ==>
Also if it happens, the trace should contain more contextual information like what is the kubeConfig.currentContext, etc

also about the `filter(predicate).length`, should it just use the `some(predicate)` operation

packages/main/src/plugin/kubernetes/test-informer.ts
```diff
@@ -57,7 +57,7 @@ export class TestInformer {
     }
     if (this.connectResponse === undefined) {
       for (let i = 0; i < this.resourcesCount; i++) {
-        this.onCb.get('add')?.({});
+        this.onCb.get('add')?.({ metadata: { uid: `${i}` } });
```
benoitf:
```suggestion
        this.onCb.get('add')?.({ metadata: { uid: String(i) } });
```

benoitf: or i.toString()

benoitf (approved):

gastoner (approved):

PR #9382, opened 2024-10-15
feat: allow to ssh to podman virtual machine - api.d.ts changes

### LINKED ISSUE #889: Provide ability to easily get a shell in a machine

It happens sometimes that I need to check the status of my podman-machine and instruct some commands into it.

I'd like the ability to get a shell into the podman-machine so I can more easily interact with it and eventually configure/fix it.

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Provide the ability to ssh to podman machine - api.d.ts changes
1/4 PR's
-Needs to be merged first

### Screenshot / video of UI

Part of #889 

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -790,24 +790,63 @@ declare module '@podman-desktop/api' {
   }
 
   /**
-   * the callbacks of ssh shell in provider connection
+   * Callback for openning shell session
    */
   export interface ProviderConnectionShellAccess {
+    /**
+     * Opens new session using ProviderConnectionShellAccessImpl class
+     * @example
+     * const providerConnectionShellAccess = new ProviderConnectionShellAccessImpl(machineInfo);
+     * const session = providerConnectionShellAccess.open()
+     */
+    open(): ProviderConnectionShellAccessSession;
+  }
+
+  /**
+   * Callbacks for interaction with shell session
+   */
+  export interface ProviderConnectionShellAccessSession {
+    /**
+     * Receiving data event
+     * @example
+     * session.onData(data => {...
+     */
     onData: Event<ProviderConnectionShellAccessData>;
+
+    /**
+     * Error event
+     * @example
+     * session.onError(error => {...
+     */
     onError: Event<ProviderConnectionShellAccessError>;
+
+    /**
+     * End event
+     * @example
+     * session.onEnd(onEnd);
+     */
     onEnd: Event<void>;
 
-    // function used for writting to shell (e.g. `ls`)
-    write(data: string): void;
+    /**
+     * Sends data
+     * @example
+     * session.write(data)
+     */
+    write(data: string | Uint8Array): void;
 
-    // function that starts the ssh connection to providers connection shell
-    startConnection(): void;
+    /**
+     * Notifies server that terminal window has been resized
+     * @example
+     * session.resize(dimensions)
+     */
+    resize(dimensions: ProviderConnectionShellDimensions): void;
 
-    // function that stops running ssh connection
-    stopConnection(): void;
-
-    // function that sets window size of shell
-    setWindow(dimensions: ProviderConnectionShellDimensions): void;
+    /**
+     * Closes opened session and removes all listeners
+     * @example
+     * session.close()
+     */
+    close(): void;
   }
 
   export interface ProviderConnectionShellDimensions {
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -393,6 +393,7 @@ declare module '@podman-desktop/api' {
     displayName?: string;
     type: 'docker' | 'podman';
     endpoint: ContainerProviderConnectionEndpoint;
+    shellAccess?: ProviderConnectionShellAccess;
     lifecycle?: ProviderConnectionLifecycle;
     status(): ProviderConnectionStatus;
     vmType?: string;
@@ -786,6 +787,40 @@ declare module '@podman-desktop/api' {
   export interface ProviderContainerConnection {
     providerId: string;
     connection: ContainerProviderConnection;
+  }
+
+  /**
+   * the callbacks of ssh shell in provider connection
+   */
+  export interface ProviderConnectionShellAccess {
+    onData: Event<ProviderConnectionShellAccessData>;
+    onError: Event<ProviderConnectionShellAccessError>;
+    onEnd: Event<void>;
+
+    // function used for writting to shell (e.g. `ls`)
+    write(data: string): void;
+
+    // function that starts the ssh connection to providers connection shell
+    startConnection(): void;
+
+    // function that stops running ssh connection
+    stopConnection(): void;
+
+    // function that sets window size of shell
+    setWindow(dimensions: ProviderConnectionShellDimensions): void;
+  }
+
+  export interface ProviderConnectionShellDimensions {
+    rows: number;
+    cols: number;
+  }
+
+  export interface ProviderConnectionShellAccessError {
+    error: string;
+  }
+
+  export interface ProviderConnectionShellAccessData {
+    data: string;
   }
 
   /**
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -796,8 +796,39 @@ declare module '@podman-desktop/api' {
     /**
      * Opens new session using ProviderConnectionShellAccessImpl class
      * @example
-     * const providerConnectionShellAccess = new ProviderConnectionShellAccessImpl(machineInfo);
-     * const session = providerConnectionShellAccess.open()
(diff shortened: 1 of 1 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,29 @@ declare module '@podman-desktop/api' {
...
+    onData: Event<ProviderConnectionShellAccessData>;
+    onError: Event<ProviderConnectionShellAccessError>;
+    onEnd: Event<void>;
+    write(data: string): void;
+    startConnection(): void;
+    stopConnection(): void;
+    setWindow(dimensions: ShellDimensions): void;
+  }
+
+  export interface ShellDimensions {
```
benoitf: should be prefixed for now as well
```suggestion
  export interface ProviderConnectionShellDimensions {
```

benoitf (commented):
you should add some sample/typedoc 

also I don't understand starConnection and stopConnection

`setWindow` should be maybe `resize`

gastoner:
> you should add some sample/typedoc
> 
> also I don't understand starConnection and stopConnection
> 
> `setWindow` should be maybe `resize`

The start/stop is used in stop ssh connection at first if there is any/ or to start the connection when are all callbacks set 

The `resize` is used in container terminal and the ssh2 lib uses `setWindow`, so I wanted to stick with ssh2 lib theme

benoitf (commented):
I think the API might be hard to understand if you still use ssh2 method as setWindow means nothing here, so please use resize as we're resizing

Then I'm not sure the on callback should be there if you follow that approach of having also "start and stop"

because the on callback and write are useless if you're not connected as they're tied to a connection/session

it means, from API POV, the events and method to interact with a session should only be part of the result of the start()

And from what I see, it's more to connect/disconnect than start and stop

or open and close

so maybe it's connect() method that gives you an object and on this object you have the on('data', 'error', 'end') events and the write method and the disconnect and the resize



also please remove any mention of 'ssh connection' as it's up to the implementation to see how it connects. API is not relying on ssh

benoitf (commented):
I'm not sure that write is with string parameter, maybe it could also be Uint8Array

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,38 @@ declare module '@podman-desktop/api' {
...
+  }
+
+  /**
+   * Callbacks for interaction with shell session
+   */
+  export interface ProviderConnectionShellAccessSession {
+    onData: Event<ProviderConnectionShellAccessData>;
+    onError: Event<ProviderConnectionShellAccessError>;
+    onEnd: Event<void>;
+    write(data: string): void;
```
```suggestion
    write(data: string | Uint8Array): void;
```

benoitf (commented):
you might add samples with `@example`

axel7083:
How can I test as you are doing in the video, which branches should I combined with this PR ?

gastoner:
@axel7083 all changes are in https://github.com/gastoner/podman-desktop/tree/provide_ability_to_easily_get_a_shell_in_a_machine_test branch 

Otherwise there is a provide_ability_to_easily_get_a_shell_in_a_machine + "a"/"b"/"c" branches (4 total)

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,79 @@ declare module '@podman-desktop/api' {
...
   }
 
+  /**
+   * Callback for openning shell session
+   */
+  export interface ProviderConnectionShellAccess {
+    /**
+     * Opens new session using ProviderConnectionShellAccessImpl class
+     * @example
+     * const providerConnectionShellAccess = new ProviderConnectionShellAccessImpl(machineInfo);
```
axel7083: Example need to be in triple quote, and preferably with the language associate

https://github.com/containers/podman-desktop/blob/6c8c434fa61f175288c33e9b0ae043119166783a/packages/extension-api/src/extension-api.d.ts#L704-L705

gastoner: @axel7083  Needs to be inside the ```typescript valid typescript code? thinking about this => `session.onData(data => {...`

axel7083: I am not sure to understand what you mean

axel7083: thanks

axel7083 (approved):
LGTM

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,93 @@ declare module '@podman-desktop/api' {
...
+  export interface ProviderConnectionShellAccess {
+    /**
+     * Opens new session using ProviderConnectionShellAccessImpl class
+     * @example
+     * ```typescript
+     * const providerConnectionShellAccess = new ProviderConnectionShellAccessImpl(machineInfo);
+     * const session = providerConnectionShellAccess.open()
+     * ```
+     */
+    open(): ProviderConnectionShellAccessSession;
```
benoitf, line 833: not sure I get the example.

If I look at the example part, it's the code I'll write in my extension so my extension will never call the open method, it will be Podman Desktop

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,93 @@ declare module '@podman-desktop/api' {
...
+     * session.onData(data => {...
+     * ```
+     */
+    onData: Event<ProviderConnectionShellAccessData>;
+
+    /**
+     * Error event
+     * @example
+     * ```typescript
+     * session.onError(error => {...
```
benoitf: I think you could add session.OnError(console.error) rather than incomplete `=> { ...`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,93 @@ declare module '@podman-desktop/api' {
...
+     * session.write(data)
+     * ```
+     */
+    write(data: string | Uint8Array): void;
+
+    /**
+     * Notifies server that terminal window has been resized
+     * @example
+     * ```typescript
+     * session.resize(dimensions)
```
benoitf: here I don't get as well the sample

it's not you that call resize, in the extension you implement the resize method

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,93 @@ declare module '@podman-desktop/api' {
...
+     * ```
+     */
+    onEnd: Event<void>;
+
+    /**
+     * Sends data
+     * @example
+     * ```typescript
+     * session.write(data)
+     * ```
```
benoitf: don't get it as well

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,93 @@ declare module '@podman-desktop/api' {
...
+     */
+    resize(dimensions: ProviderConnectionShellDimensions): void;
+
+    /**
+     * Closes opened session and removes all listeners
+     * @example
+     * ```typescript
+     * session.close()
+     * ```
```
benoitf: example needs to be from the extension's POV

benoitf (changes requested):

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,131 @@ declare module '@podman-desktop/api' {
...
+     * open(): ProviderConnectionShellAccessSession {
+     *  // Create stream 
+     *  #connected = true;
+     *  return {
+     *    onData: onData,
+     *    onError: onError,
+     *    onEnd: onEnd,
+     *    write: write.bind(this),
+     *    resize: resize.bind(this),
+     *    close: close,
```
benoitf: when left and right part is the same, you can just use the name

return { onData}

but still, if I look at this example, how will it help me ? it seems here I won't do anything with this example

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,131 @@ declare module '@podman-desktop/api' {
...
+    /**
+     * Receiving data event
+     * @example
+     * ```typescript
+     * onDataEmit = new EventEmitter<ProviderConnectionShellAccessData>();
+     * onData: Event<ProviderConnectionShellAccessData> = this.onDataEmit.event;
+     * 
+     * // later in code when "data" are emmited
+     * onDataEmit.fire({ data: data });
+     * ```
```
benoitf: I don't understand again, you're describing the code of Podman Desktop here

an extension will not fire anything to onData, just consume it

we need to use a code that extension will use/follow

so here, we don't describe what Podman Desktop core is doing, just how I can do things from my extension

benoitf (changes requested):
if the examples mean nothing, then we should not provide any example

the thing, when writing examples is

how will it help me in my extension if I need to implement that

here it's not helping me but misleading me as I see often how it's used inside Podman Desktop while I expect to see the code I want to write in my extensions

benoitf (approved):
thanks for the changes, now the example is more helpful and in the context of an author of the extension

PR #9383, opened 2024-10-15
feat: allow to ssh to podman virtual machine - podman changes

### LINKED ISSUE #889: Provide ability to easily get a shell in a machine

gastoner:
### What does this PR do?
Provide the ability to ssh to podman machine - api.d.ts changes
3/4 PR's
-Needs to be rebased
### Screenshot / video of UI

Part of #889 

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -814,6 +814,7 @@ export async function registerProviderFor(
   }
 
   const providerConnectionShellAccess = new ProviderConnectionShellAccessImpl(machineInfo);
+  storedExtensionContext?.subscriptions.push(providerConnectionShellAccess);
   const containerProviderConnection: extensionApi.ContainerProviderConnection = {
     name: machineInfo.name,
     displayName: prettyMachineName(machineInfo.name),
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -39,6 +39,7 @@ import { getPodmanCli, getPodmanInstallation } from './podman-cli';
 import { PodmanConfiguration } from './podman-configuration';
 import { PodmanInfoHelper } from './podman-info-helper';
 import { HyperVCheck, PodmanInstall, WSL2Check, WSLVersionCheck } from './podman-install';
+import { PodmanMachineStream } from './podman-machine-stream';
 import { PodmanRemoteConnections } from './podman-remote-connections';
 import { QemuHelper } from './qemu-helper';
 import { RegistrySetup } from './registry-setup';
@@ -812,11 +813,14 @@ export async function registerProviderFor(
     };
   }
 
+  const shellAccessProvider = new PodmanMachineStream(machineInfo);
+
   const containerProviderConnection: extensionApi.ContainerProviderConnection = {
     name: machineInfo.name,
     displayName: prettyMachineName(machineInfo.name),
     type: 'podman',
     status: () => podmanMachinesStatuses.get(machineInfo.name) ?? 'unknown',
+    shellAccess: shellAccessProvider,
     lifecycle,
     endpoint: {
       socketPath,
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -39,7 +39,7 @@ import { getPodmanCli, getPodmanInstallation } from './podman-cli';
 import { PodmanConfiguration } from './podman-configuration';
 import { PodmanInfoHelper } from './podman-info-helper';
 import { HyperVCheck, PodmanInstall, WSL2Check, WSLVersionCheck } from './podman-install';
-import { PodmanMachineStream } from './podman-machine-stream';
+import { ProviderConnectionShellAccessImpl } from './podman-machine-stream';
 import { PodmanRemoteConnections } from './podman-remote-connections';
 import { QemuHelper } from './qemu-helper';
 import { RegistrySetup } from './registry-setup';
@@ -813,14 +813,13 @@ export async function registerProviderFor(
     };
   }
 
-  const shellAccessProvider = new PodmanMachineStream(machineInfo);
-
+  const providerConnectionShellAccess = new ProviderConnectionShellAccessImpl(machineInfo);
   const containerProviderConnection: extensionApi.ContainerProviderConnection = {
     name: machineInfo.name,
     displayName: prettyMachineName(machineInfo.name),
     type: 'podman',
     status: () => podmanMachinesStatuses.get(machineInfo.name) ?? 'unknown',
-    shellAccess: shellAccessProvider,
+    shellAccess: providerConnectionShellAccess,
     lifecycle,
     endpoint: {
       socketPath,
--- a/extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
+++ b/extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
@@ -1,133 +1,172 @@
-/**********************************************************************
- * Copyright (C) 2024 Red Hat, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- *
- * SPDX-License-Identifier: Apache-2.0
- ***********************************************************************/
-
-/* eslint-disable @typescript-eslint/no-explicit-any */
-import * as fs from 'node:fs';
-import { rm } from 'node:fs/promises';
-import { type AddressInfo, createConnection, createServer } from 'node:net';
-import { tmpdir } from 'node:os';
-import { join } from 'node:path';
-
-import { Server } from 'ssh2';
-import { generatePrivateKey } from 'sshpk';
-import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
+import { Client } from 'ssh2';
+import { beforeEach, describe, expect, test,vi } from 'vitest';
 
 import type { MachineInfo } from './extension';
-import { PodmanMachineStream } from './podman-machine-stream';
+import { ProviderConnectionShellAccessImpl } from './podman-machine-stream';
 
-// mock the API
+const onMock = vi.fn();
+const connectMock = vi.fn();
+const emitMock = vi.fn();
+const writeStreamMock = vi.fn();
+
 vi.mock('@podman-desktop/api', async () => {
   return {
-    process: {
-      exec: vi.fn(),
-    },
     EventEmitter: vi.fn(),
   };
 });
 
-vi.mock('node:fs');
+const streamMock = new (require('node:events').EventEmitter)();
 
-beforeEach(() => {
-  vi.resetAllMocks();
+streamMock.write = writeStreamMock;
+
+// Mock ssh2 Client
+vi.mock('ssh2', () => {
+  return {
+    Client: vi.fn().mockImplementation(() => ({
+      on: onMock,
+      connect: connectMock,
+      shell: vi.fn(callback => {
+        callback(null, streamMock);
+      }),
+      emit: emitMock,
+    })),
+  };
 });
 
-let dummyKey: string;
-beforeAll(async () => {
-  // generate on the fly a dummy key
-  dummyKey = generatePrivateKey('ed25519').toString('ssh');
-});
-
-test('should be able to connect', async () => {
-  vi.spyOn(fs, 'readFileSync').mockReturnValue(dummyKey);
-  let sshPort = 0;
-  let connected = false;
-  let authenticated = false;
-
-  // create ssh server
-  const sshServer = new Server(
-    {
-      hostKeys: [dummyKey],
-    },
-    client => {
-      connected = true;
-
-      client
-        .on('authentication', ctx => {
-          ctx.accept();
-        })
-        .on('ready', () => {
-          authenticated = true;
-        });
-    },
-  ).listen(0, '127.0.0.1', () => {
-    const address: AddressInfo = sshServer.address() as AddressInfo;
-    sshPort = address?.port;
-  });
-
-  // wait that the server is listening
-  await vi.waitFor(() => expect(sshPort).toBeGreaterThan(0));
-
-  // create a npipe/socket server
-  // on windows it's an npipe, on macOS a socket file
-  let socketOrNpipePathLocal: string;
-  let socketOrNpipePathRemote: string;
-  if (process.platform === 'win32') {
-    socketOrNpipePathLocal = '\\\\.\\pipe\\test-local';
-    socketOrNpipePathRemote = '\\\\.\\pipe\\test-remote';
-  } else {
-    socketOrNpipePathLocal = join(tmpdir(), 'test-local.sock');
-    socketOrNpipePathRemote = join(tmpdir(), 'test-remote.sock');
+class TestProviderConnectionShellAccessImpl extends ProviderConnectionShellAccessImpl {
+  isConnected(): boolean {
+    return super.isConnected();
   }
+}
 
-  // delete file if exists
-  await rm(socketOrNpipePathLocal, { force: true });
-  await rm(socketOrNpipePathRemote, { force: true });
+describe('Test SSH Client', () => {
+  let client: Client;
 
-  let listenReady = false;
-
-  // start a remote server
-  const npipeServer = createServer(_socket => {}).listen(socketOrNpipePathRemote, () => {
-    listenReady = true;
+  beforeEach(() => {
+    client = new Client();
   });
 
-  await vi.waitFor(() => expect(listenReady).toBeTruthy());
+  test('should register the ready event', () => {
+    const onReady = vi.fn();
 
-  const machineInfo: MachineInfo = {
-    port: sshPort,
-    remoteUsername: 'user',
-    identityPath: '/some/path',
-  } as unknown as MachineInfo;
+    // Adds callback for 'ready'
+    client.on('ready', onReady);
 
-  const podmanMachineStream = new PodmanMachineStream(machineInfo);
-
-  podmanMachineStream.startConnection();
-
-  // wait authenticated and connected
-  await vi.waitFor(() => expect(connected && authenticated).toBeTruthy());
-
-  let connectedToLocal = false;
-  // send a request to the tunnel using the socket path
-  const client = createConnection({ path: socketOrNpipePathLocal }, () => {
-    connectedToLocal = true;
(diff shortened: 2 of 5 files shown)


slemeur:
Nice!!

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,172 @@
+import { Client } from 'ssh2';
```
benoitf, line 19: missing copyright headers

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+const emitMock = vi.fn();
+const writeStreamMock = vi.fn();
+
+vi.mock('@podman-desktop/api', async () => {
+  return {
+    EventEmitter: vi.fn(),
+  };
+});
+
+const streamMock = new (require('node:events').EventEmitter)();
```
benoitf: why do we need to use 'require' and not import ?

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+      if (eventName === 'ready') {
+        onReady(...args);
+      }
+    });
+
+    // Emit event 'ready'
+    client.emit('ready');
+
+    // Wait for the shell callback to be invoked
+    await new Promise(resolve => setTimeout(resolve, 150));
```
benoitf: need to use then waitFor callback.toHaveBeenCalled

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,172 @@
...
```
benoitf: need to use waitFor for a condition

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (this.#stream) {
+      this.#stream.write(data);
+    }
+  }
+
+  resize(dimensions: ProviderConnectionShellDimensions): void {
+    if (this.#stream) {
+      // rows and cols override width and height when rows and cols are non-zero.
+      this.#stream.setWindow(dimensions.rows, dimensions.cols, 0, 0);
+    }
```
benoitf:
```suggestion
    // rows and cols override width and height when rows and cols are non-zero.
    this.#stream?.setWindow(dimensions.rows, dimensions.cols, 0, 0);
```

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,145 @@
...
+  onErrorEmit = new EventEmitter<ProviderConnectionShellAccessError>();
+  onError: Event<ProviderConnectionShellAccessError> = this.onErrorEmit.event;
+
+  onEndEmit = new EventEmitter<void>();
+  onEnd: Event<void> = this.onEndEmit.event;
+
+  write(data: string): void {
+    if (this.#stream) {
+      this.#stream.write(data);
+    }
```
benoitf:
```suggestion
    this.#stream?.write(data);
```

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,145 @@
...
+      // rows and cols override width and height when rows and cols are non-zero.
+      this.#stream.setWindow(dimensions.rows, dimensions.cols, 0, 0);
+    }
+  }
+
+  disposeListeners(): void {
+    this.onDataEmit.dispose();
+    this.onErrorEmit.dispose();
+    this.onEndEmit.dispose();
+  }
```
benoitf: why it's not only called dispose() (as we have an interface IDisposable)

packages/extension-api/src/extension-api.d.ts
```diff
@@ -788,6 +789,79 @@ declare module '@podman-desktop/api' {
     connection: ContainerProviderConnection;
   }
 
+  /**
+   * Callback for openning shell session
+   */
+  export class ProviderConnectionShellAccess {
+    /**
```
benoitf: changes should be removed there (need to rebase as it has been merged in api)

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+const onMock = vi.fn();
+const onStreamMock = vi.fn();
+const streamMock = {
+  on: onStreamMock,
+  write: vi.fn(),
+};
+
+afterEach(() => {
+  vi.clearAllMocks();
+});
```
benoitf: FYI I think pattern we have in most of the tests is to reset on beforeEach

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+
+// Mock ssh2 Client
+vi.mock('ssh2', () => {
+  return {
+    Client: vi.fn().mockImplementation(() => ({
+      on: onMock,
+      connect: vi.fn(),
+      shell: vi.fn(callback => {
+        callback(null, streamMock);
```
benoitf: can we use undefined rather than null ?

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+    })),
+  };
+});
+
+describe('Test SSH Client', () => {
+  let client: Client;
+  let providerConnectionShellAccess: TestProviderConnectionShellAccessImpl;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
```
benoitf: it looks strange to have again a clear there

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+    expect(providerConnectionShellAccess.onErrorEmit.fire).toHaveBeenCalledWith({ error: 'Error message' });
+  });
+});
+
+describe('Test SSH Stream', () => {
+  let providerConnectionShellAccess: TestProviderConnectionShellAccessImpl;
+  let client: Client;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
```
benoitf: same remark

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+    onStreamMock.mockImplementation((eventName, fn) => {
+      if (eventName === 'close') {
+        fn();
+      }
+      return streamMock;
+    });
+
+    providerConnectionShellAccess.open();
+    const connected = providerConnectionShellAccess.isConnected();
+    vi.waitFor(() => expect(connected).toBeFalsy());
```
benoitf: missing the await

axel7083:
```suggestion
    await vi.waitFor(() => expect(connected).toBeFalsy());
```

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+  test('already connected', async () => {
+    const disposeMock = vi.spyOn(providerConnectionShellAccess, 'dispose');
+
+    // stream.on needs to return ClientChannel (streamMock)
+    onStreamMock.mockReturnValue(streamMock);
+
+    Object.defineProperty(providerConnectionShellAccess, '#connected', { value: true });
+
+    const connected = providerConnectionShellAccess.isConnected();
+    vi.waitFor(() => expect(connected).toBeTruthy());
```
axel7083:
```suggestion
    await vi.waitFor(() => expect(connected).toBeTruthy());
```

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+      onEnd: providerConnectionShellAccess.onEnd,
+      write: expect.any(Function),
+      resize: expect.any(Function),
+      close: expect.any(Function),
+    });
+
+    const connected1 = providerConnectionShellAccess.isConnected();
+    expect(connected1).toBeTruthy();
+
+    vi.waitFor(() => expect(disposeMock).toHaveBeenCalled());
```
axel7083:
```suggestion
    await vi.waitFor(() => expect(disposeMock).toHaveBeenCalled());
```

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.#port = podmanMachine.port;
+    this.#username = podmanMachine.remoteUsername;
+    this.#privateKey = podmanMachine.identityPath;
+  }
+
+  protected isConnected(): boolean {
+    return this.#connected;
+  }
+
+  onDataEmit = new EventEmitter<ProviderConnectionShellAccessData>();
```
axel7083, line 55: might need to mock `EventEmitter` inside extension.spec.ts

gastoner:
Can I use `Object.defineProperty(providerConnectionShellAccess, '#connected', { value: true });` to set `#connected` property to true?

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+
+const onMock = vi.fn();
+const onStreamMock = vi.fn();
+const streamMock = {
+  on: onStreamMock,
+  write: vi.fn(),
+};
+
+beforeEach(() => {
+  vi.clearAllMocks();
```
benoitf, line 37:
```suggestion
  vi.resetAllMocks();
```

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.onErrorEmit.dispose();
+    this.onEndEmit.dispose();
+  }
+
+  close(): void {
+    this.#connected = false;
+    this.#client?.end();
+    this.#client?.destroy();
+    this.dispose();
+  }
```
benoitf, line 84: maybe merge the two methods into close (or at least dispose should not be public)

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  open(): ProviderConnectionShellAccessSession {
+    // Avoid multiple connections
+    if (this.#connected) {
+      this.dispose();
```
axel7083: I would say this probably deserve a console.warn, as we are closing an existing connection

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,141 @@
...
+  ProviderConnectionShellAccessSession,
+  ProviderConnectionShellDimensions,
+} from '@podman-desktop/api';
+import { EventEmitter } from '@podman-desktop/api';
+import type { ClientChannel } from 'ssh2';
+import { Client } from 'ssh2';
+
+import type { MachineInfo } from './extension';
+
+export class ProviderConnectionShellAccessImpl implements ProviderConnectionShellAccess {
```
axel7083:
```suggestion
export class ProviderConnectionShellAccessImpl implements ProviderConnectionShellAccess extends Disposable {
```

Makes ProviderConnectionShellAccessImpl  extends Disposable, and add it to the subscriptions, so it is cleaned up on extension disactivate

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -812,11 +813,13 @@ export async function registerProviderFor(
     };
   }
 
+  const providerConnectionShellAccess = new ProviderConnectionShellAccessImpl(machineInfo);
```
axel7083, line 816: here `providerConnectionShellAccess` should be (extends) a disposable and push to subscriptions

extensions/podman/packages/extension/src/podman-machine-stream.spec.ts
gastoner: For seme reason the resetAllMocks does not work (all of mine tests fails), but it works with clear, any ideas?

benoitf: probably due to
```
 EventEmitter: vi.fn().mockImplementation(() => ({
      fire: vi.fn(),
      dispose: vi.fn(),
    })),
```

benoitf: (all the mock implementations you're doing in mocking the modules)

extensions/podman/packages/extension/src/podman-machine-stream.ts
gastoner: The dispose is just removing listers, the close function will close the connection
But sure, I can add warning e.g. "Removing listeners"

extensions/podman/packages/extension/src/podman-machine-stream.ts
gastoner: I would have to change the api.d.ts file for extending the `ProviderConnectionShellAccess`, I can add the `extends Disposable` before `implements` but then I get additional errors like "Class extends value #<Object> is not a constructor or null

axel7083: did you import Disposable class from the api ?

axel7083: Use implements would work **

see the example in ai-lab

https://github.com/containers/podman-desktop-extension-ai-lab/blob/8e9ce351a673be43e54a21c2cfb35555861de50d/packages/backend/src/managers/inference/inferenceManager.ts#L37

We import the type from podman-desktop api

https://github.com/containers/podman-desktop-extension-ai-lab/blob/8e9ce351a673be43e54a21c2cfb35555861de50d/packages/backend/src/managers/inference/inferenceManager.ts#L20

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,147 @@
...
+  ProviderConnectionShellAccessSession,
+  ProviderConnectionShellDimensions,
+} from '@podman-desktop/api';
+import { Disposable, EventEmitter } from '@podman-desktop/api';
+import type { ClientChannel } from 'ssh2';
+import { Client } from 'ssh2';
+
+import type { MachineInfo } from './extension';
+
+export class ProviderConnectionShellAccessImpl extends Disposable implements ProviderConnectionShellAccess {
```
axel7083: yes sorry it is implements not extends 🤯

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,147 @@
...
+    this.#client?.end();
+    this.#client?.destroy();
+    this.disposeListeners();
+  }
+
+  open(): ProviderConnectionShellAccessSession {
+    // Avoid multiple connections
+    if (this.#connected) {
+      console.warn('Removing listeners');
+      this.disposeListeners();
```
benoitf: please remove the console warn

about the if condition, could you please tell in which case does it happen . I think this is the part where I'm not sure to follow and why there is a specific disposeListeners method etc

So it makes me thing that there is something to dig as I have the feeling here that if I request, as Podman Desktop, to open two connections, I will only be connected to a single ssh session 


IMHO, each open should return a new independent session. so we should not have the 'if connected'

axel7083: I suggested the warn as it seems an edge case that should not happen

gastoner: The open is called every time when you click on a "terminal," and this function will be called https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/provider-registry.ts#L1284, which calls "open" (opens a new session and adds listeners for it).

But when you go to a different tab and go back to the "terminal" +1 listener is added 
This results in duplicated data in the terminal; that's why there is the `if`, which checks if a given providerConnection already exists session, and if yes, removes all listeners (because they are added in lines L1284+ anyways). 

Another edge case is typing "exit" which also some problems executing 
There should be all the changes / "working solution" but without suggestions from reviews https://github.com/gastoner/podman-desktop/tree/provide_ability_to_easily_get_a_shell_in_a_machine_test

benoitf: I expect to have a new session each time I call open

so there should not be any isConnected stuff

how the UI will reuse a session or create a new one should not be up to the extension

The API is saying that calling open returns a new session, so I expect to have a dedicated session/terminal on each open call, not to share with others my session

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.onDataEmit.dispose();
+    this.onErrorEmit.dispose();
+    this.onEndEmit.dispose();
+  }
+
+  close(): void {
+    this.#connected = false;
+    this.#client?.end();
+    this.#client?.destroy();
+    this.disposeListeners();
```
benoitf: if there are two clients connected (that have called open), if the first one call close() it would close everything for both of them

this is why I think it's simpler/easier to not have the isConnected flag, and just always open when calling `open()` and closing when calling `close()`

gastoner: I would say that each connection has its session because each one of them will store the `ProviderConnectionShellAccessImpl` instance in its shellAccess. But the session of the given connection is opened only at the first call of "open". I would have to export the "close" function to the frontend to call it during onDestroy. This should result (+ some changes in this PR and another PR for exporting close function to backed) in making a new session on going to the "terminal tab"

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.#privateKey = podmanMachine.identityPath;
+  }
+
+  dispose(): void {
+    this.close();
+  }
+
+  protected isConnected(): boolean {
+    return this.#connected;
+  }
```
benoitf:
```suggestion
```
todo: remove it as we're not using it

extensions/podman/packages/extension/src/podman-machine-stream.ts
```diff
@@ -0,0 +1,140 @@
...
+import type { MachineInfo } from './extension';
+
+export class ProviderConnectionShellAccessImpl implements ProviderConnectionShellAccess, Disposable {
+  #host: string;
+  #port: number;
+  #username: string;
+  #privateKey: string;
+  #client: Client | undefined;
+  #stream: ClientChannel | undefined;
+  #connected: boolean = false;
```
benoitf:
```suggestion
```

todo: remove it as we're not using it

benoitf (approved):

axel7083 (approved):
LGTM

PR #9388, opened 2024-10-15
fix: start secondary informers when context is reachable

feloy:

### What does this PR do?

Start secondary informers when context is reachable and if these resources have been subscribed.

This makes possible to see the nodes, services, etc when PD starts with the cluster stopped, and the cluster is started afterward.

Part of #9156

### How to test this PR?

- Create a Kind cluster, and stop it
- Start Podman Desktop
- check that the context is unreachable
- start the Kind cluster
- wait for the context to be reachable
- check that the node, services, configmap, etc are listed in the Kubernetes section

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
@@ -1498,6 +1498,169 @@ describe('update', async () => {
     );
   });
 
+  test('current context becoming reachable should start service informer on it if watchers have subscribed', async () => {
+    vi.useFakeTimers();
+    const makeInformerMock = vi.mocked(makeInformer);
+    makeInformerMock.mockImplementation(
+      (
+        kubeconfig: kubeclient.KubeConfig,
+        path: string,
+        _listPromiseFn: kubeclient.ListPromise<kubeclient.KubernetesObject>,
+      ) => {
+        return new TestInformer(
+          kubeconfig.currentContext,
+          path,
+          0,
+          undefined,
+          [
+            {
+              delayMs: 100,
+              verb: 'add',
+              object: { metadata: { uid: '123' } },
+            },
+          ],
+          [
+            {
+              delayMs: 0,
+              error: 'an error',
+              verb: 'error',
+            },
+          ],
+        );
+      },
+    );
+    client = new TestContextsManager(apiSender);
+    const kubeConfig = new kubeclient.KubeConfig();
+    const config = {
+      clusters: [
+        {
+          name: 'cluster1',
+          server: 'server1',
+        },
+      ],
+      users: [
+        {
+          name: 'user1',
+        },
+      ],
+      contexts: [
+        {
+          name: 'context1',
+          cluster: 'cluster1',
+          user: 'user1',
+          namespace: 'ns1',
+        },
+        {
+          name: 'context2',
+          cluster: 'cluster1',
+          user: 'user1',
+          namespace: 'ns2',
+        },
+      ],
+      currentContext: 'context1',
+    };
+    kubeConfig.loadFromOptions(config);
+    await client.update(kubeConfig);
+    vi.advanceTimersToNextTimer();
+    vi.advanceTimersToNextTimer();
+
+    makeInformerMock.mockClear();
+
+    client.registerGetCurrentContextResources('services');
+
+    // service informer is not started
+    expect(makeInformerMock).not.toHaveBeenCalled();
+
+    // wait for context to become reachable (receiving an add event)
+    vi.advanceTimersByTime(110);
+
+    // service informer is now started
+    expect(makeInformerMock).toHaveBeenCalled();
+    expect(makeInformerMock).toHaveBeenCalledWith(
+      expect.any(KubeConfig),
+      '/api/v1/namespaces/ns1/services',
+      expect.anything(),
+    );
+  });
+
+  test('current context becoming reachable should not start service informer on it if no watchers have subscribed', async () => {
+    vi.useFakeTimers();
+    const makeInformerMock = vi.mocked(makeInformer);
+    makeInformerMock.mockImplementation(
+      (
+        kubeconfig: kubeclient.KubeConfig,
+        path: string,
+        _listPromiseFn: kubeclient.ListPromise<kubeclient.KubernetesObject>,
+      ) => {
+        return new TestInformer(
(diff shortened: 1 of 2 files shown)


axel7083 (commented):
Do not fix the issue I reported in https://github.com/containers/podman-desktop/issues/9156

If I follow the steps you are mentioning, the kind cluster is able to recover (reconnect); but if I do the following (which are the same steps I followed in the issue I reported)

- assert have a kind cluster **running** `podman ps --filter label=io.x-k8s.kind.cluster`
- Start Podman Desktop
- assert clusters resources visible (nodes, etc.)
- Stop kind cluster (E.g. through the cli: `podman stop <container-id>`
- assert connection error in Kubernetes page
- Wait 30s
- Start kind cluster (again through cli)
- Wait 120 seconds
- assert error in logs and cluster not connected, never


https://github.com/user-attachments/assets/c8fef978-a33f-4c3c-aed0-67e0afe8d15d

feloy:
> Do not fix the issue I reported in #9156
> 
> If I follow the steps you are mentioning, the kind cluster is able to recover (reconnect); but if I do the following (which are the same steps I followed in the issue I reported)
> 
> * assert have a kind cluster **running** `podman ps --filter label=io.x-k8s.kind.cluster`
> * Start Podman Desktop
> * assert clusters resources visible (nodes, etc.)
> * Stop kind cluster (E.g. through the cli: `podman stop <container-id>`
> * assert connection error in Kubernetes page
> * Wait 30s
> * Start kind cluster (again through cli)
> * Wait 120 seconds
> * assert error in logs and cluster not connected, never
> 
>  kind-connection-lost.mp4

Yes, for the error you reported specifically, this should be fixed with the patch https://github.com/kubernetes-client/javascript/pull/1934

axel7083:
> Yes, for the error you reported specifically

:+1:

axel7083 (approved):

packages/main/src/plugin/kubernetes/contexts-informers-registry.ts
```diff
@@ -36,6 +36,9 @@ export class ContextsInformersRegistry {
   }
 
   setInformers(name: string, informers: ContextInternalState | undefined): void {
+    if (this.informers.get(name)) {
+      console.debug(`primary informers in context ${name} already set`);
```
jeffmaury: Are we gonna keep this trace ? Shouldn't we return in that case ?

packages/main/src/plugin/kubernetes/contexts-informers-registry.ts
```diff
@@ -36,6 +36,9 @@ export class ContextsInformersRegistry {
   }
 
   setInformers(name: string, informers: ContextInternalState | undefined): void {
+    if (this.informers.get(name)) {
```
benoitf: curious why we're not using the method `hasContext` or `hasInformer`
(same remark for line 53)

packages/main/src/plugin/kubernetes/contexts-informers-registry.ts
feloy: Yes, I would like to keep the trace, and not return here.

The purpose is to check that the informer was correctly deleted from the map before we set it again here, to avoid any memory leak. In any case, we want to set the new one.

packages/main/src/plugin/kubernetes/contexts-informers-registry.ts
feloy: We are doing a `this.informers.set(name, value)` below, I find more understandable to do a `this.informers.get(name)` directly (to check that we are not overriding a value)

benoitf: but then why using `map.get` and not `map.has`

packages/main/src/plugin/kubernetes/contexts-informers-registry.ts
benoitf: if there is something in the map, shouldn't we do something about it instead of just logging a trace

I don't see how the trace will be useful 

Either there is a bug and it needs to be fixed or there are tests

feloy: Yes, I completely get your point and I also had the same thought. 

About fixing the bug, I still have work to do related to this issue and others (in other PRs), these traces will help me detect if there are bugs when I work on these PRs.

About tests, it is not straightforward to cover all the possibilities, as the Kubernetes API and the Kubernetes library have behaviours we can hardly reproduce completely with unit tests. I would prefer to have this trace (in addition to existing tests) to detect unknown cases or regressions.

benoitf: I still believe this kind of traces is not the way to go

why ? because Podman Desktop is not only a blackbox that we run in a batch, then collect traces and then diagnose them.

We had issues in stores, etc, in the past, so how we did monitor the right usage, by reporting things directly in a way that from UI you can inspect/interact and monitor the status.

so if you want to check "memory leak" then, some data needs to be exposed so we can query it from a page and display the current data, reset the values, etc

benoitf: an example:

I'm flooded by traces about informers backoff when I open the console, while probably if I want to check the state I just want to go to a page where I see informers object, the last time where they were refreshed, I could ask for a refresh, I could ask to delete it, etc.

feloy: Effectively, I was also thinking about such a page, but I don't know exactly how/when to introduce it

feloy:
setting as draft for now, to look at potential bugs and remove the traces

benoitf (approved):
I'm fine to merge it like this as an intermediate step, just do not want to add more and more traces for long term, being able to visualize/ monitor them inside the UI would be more helpful

Also here in case of duplicates we could add the number of times we hit the duplicates in addition to only logging the trace

feloy:
> I'm fine to merge it like this as an intermediate step, just do not want to add more and more traces for long term, being able to visualize/ monitor them inside the UI would be more helpful
> 
> Also here in case of duplicates we could add the number of times we hit the duplicates in addition to only logging the trace

I was also concerned about adding these traces, I prefer to remove them and find another method.

For now, I have checked for this PR that they are never displayed for the use cases I can think about. I'll keep them on hand for the next PRs I'll work on.

Thanks for your feedback on this

PR #9392, opened 2024-10-15
feat: set slower backoff for current context

feloy:

### What does this PR do?

Set a slower backoff for current context (multiplier 1.2 instead of 2.0, and limit 10s instead of 60s)

Part of #6114 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/backoff.spec.ts
+++ b/packages/main/src/plugin/kubernetes/backoff.spec.ts
@@ -21,7 +21,7 @@ import { expect, test } from 'vitest';
 import { Backoff } from './backoff.js';
 
 test('backoff increment and limit without jitter', () => {
-  const backoff = new Backoff(100, 2, 500, 0);
+  const backoff = new Backoff({ value: 100, multiplier: 2, max: 500 });
   expect(backoff.get()).toEqual(100);
   expect(backoff.get()).toEqual(200);
   expect(backoff.get()).toEqual(400);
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/kubernetes/backoff.spec.ts
```diff
@@ -21,7 +21,7 @@ import { expect, test } from 'vitest';
 import { Backoff } from './backoff.js';
 
 test('backoff increment and limit without jitter', () => {
-  const backoff = new Backoff(100, 500, 0);
+  const backoff = new Backoff(100, 2, 500, 0);
```
benoitf: if possible, move to an object as parameter with default value

b/c reading new Backoff(100, 2, 500, 0) it's hard to know what is doing what

so either move 100 to BACKOFF_ VALUE, 2 to BACKOFF_MULTIPLIER

or introduce `new Backoff({value: 100, multiplier: 2, max: 500}) (and then jitter has an optional/default value of 0 anyway)

benoitf (approved):
thanks

packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
```diff
@@ -38,6 +38,44 @@ const DEPLOYMENTS_DEFAULT = 6;
...
+      }
+    } catch {
+      return {
+        message: () => `${map} is not a map`,
+        pass: false,
+      };
+    }
+  },
+});
+
```
benoitf: here I don't see the value of augmenting expect rather than just having a method getState() method returning true or false

feloy: We want to test that the checking state has been set to true during the period, but we cannot test that it is currently true, as it can have been changed to false during the period.

benoitf (approved):
approving again and thanks for handling reviews

gastoner (approved):
Codewise LGTM

PR #9402, opened 2024-10-16
fix: add implementation for compose update

### LINKED ISSUE #8838: better distinction between `upgrade` and `change version` in CLI Tools

dgolovin:
### What does this PR do?

Implements update support for compose CLI

Fix #8838.

### How to test this PR?

### DIFF
--- a/extensions/compose/src/cli-run.ts
+++ b/extensions/compose/src/cli-run.ts
@@ -44,7 +44,7 @@ export function getSystemBinaryPath(binaryName: string): string {
 }
 
 // Takes a binary path (e.g. /tmp/docker-compose) and installs it to the system. Renames it based on binaryName
-export async function installBinaryToSystem(binaryPath: string, binaryName: string): Promise<void> {
+export async function installBinaryToSystem(binaryPath: string, binaryName: string): Promise<string | undefined> {
   const system = process.platform;
 
   // Before copying the file, make sure it's executable (chmod +x) for Linux and Mac
@@ -91,6 +91,7 @@ export async function installBinaryToSystem(binaryPath: string, binaryName: stri
     // Use admin prileges / ask for password for copying to /usr/local/bin
     await extensionApi.process.exec(command, args, { isAdmin: true });
     console.log(`Successfully installed '${binaryName}' binary.`);
+    return destinationPath;
   } catch (error) {
     console.error(`Failed to install '${binaryName}' binary: ${error}`);
     throw error;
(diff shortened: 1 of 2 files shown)


benoitf (approved):

feloy (approved):

extensions/compose/src/cli-run.ts
```diff
@@ -44,7 +44,7 @@ export function getSystemBinaryPath(binaryName: string): string {
 }
 
 // Takes a binary path (e.g. /tmp/docker-compose) and installs it to the system. Renames it based on binaryName
-export async function installBinaryToSystem(binaryPath: string, binaryName: string): Promise<void> {
+export async function installBinaryToSystem(binaryPath: string, binaryName: string): Promise<string | undefined> {
```
axel7083, line 47: should not be Promise<string | undefined> the signature should always return a string, if the function was not able to install system-wide (user cancel, or refuse to give admin privilege) it should throw an error

extensions/compose/src/extension.ts
```diff
@@ -284,9 +284,73 @@ async function registerCLITool(composeDownload: ComposeDownload, detect: Detect)
...
+        throw new Error(
+          `Cannot update ${binaryInfo?.path} version ${binaryVersion} to ${releaseVersionToUpdateTo} as it was not installed by podman-desktop`,
+        );
+      }
+
+      // download, install system wide and update cli version
+      await composeDownload.download(releaseToUpdateTo);
+      // get the binary in the extension folder
+      const storagePath = await detect.getStoragePath();
+      binaryPath = await installBinaryToSystem(storagePath, composeCliName);
```
axel7083, line 333: try {
 // try to install system wide
} catch {
// log the error
 } finally {
// update the composeCliTool with either the path from the extension folder or the system wide if install success
}

extensions/compose/src/cli-run.ts
dgolovin: @axel7083 I'll open follow up on this, because I don't want add too much new functionality to the fix.

extensions/compose/src/extension.ts
dgolovin: Same as previous comment.

PR #9403, opened 2024-10-16
fix: show kubectl CLI version quick picks in descending order without `Kubernetes`

### LINKED ISSUE #9221: kubectl CLI `Upgrade/Downgrade` shows unsorted versions quick pick list with `Kubernetes` prefix
### Bug description

The list should contain only versions and should be sorted starting with latest released version

Image

### Operating system

macOS Sequoia

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. Open CLI settings
2. Select `Install` or `Upgrade/Downgrade` for kubectl CLI 

### Relevant log output

_No response_

### Additional context

_No response_

dgolovin:
### What does this PR do?

Removes `Kubernetes` from version quick pick and sorts version to start with latest one. 

### Screenshot / video of UI

Before:
image

After:
image

Fix #9221.

### How to test this PR?

### DIFF
--- a/extensions/kubectl-cli/src/download.spec.ts
+++ b/extensions/kubectl-cli/src/download.spec.ts
@@ -122,7 +122,7 @@ test('pick the 4th option option in the quickpickmenu and expect it to return th
   const kubectlDownload = new KubectlDownload(extensionContext, kubectlGitHubReleasesMock, os);
   const result = await kubectlDownload.promptUserForVersion();
   expect(result).toBeDefined();
-  expect(result).toEqual(releases[3]); // "4th" option was picked
+  expect(result).toEqual(releases[5]); // "6th" option was picked
 });
 
 test('test download of kubectl passes and that mkdir and executable mocks are called', async () => {
--- a/extensions/kubectl-cli/src/download.spec.ts
+++ b/extensions/kubectl-cli/src/download.spec.ts
@@ -109,7 +109,7 @@ test('expect getLatestVersionAsset to return the latest release from a list of r
   expect(result.tag).toEqual('v1.29.0-rc.1');
 });
 
-test('pick the 4th option option in the quickpickmenu and expect it to return the github release information', async () => {
+test('pick the 6th option option in the quickpickmenu and expect it to return the github release information', async () => {
   grabLatestsReleasesMetadataMock.mockImplementation(() => {
     return releases;
   });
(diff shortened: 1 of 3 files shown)


extensions/kubectl-cli/src/download.spec.ts
```diff
@@ -122,7 +122,7 @@ test('pick the 4th option option in the quickpickmenu and expect it to return th
   const kubectlDownload = new KubectlDownload(extensionContext, kubectlGitHubReleasesMock, os);
   const result = await kubectlDownload.promptUserForVersion();
   expect(result).toBeDefined();
-  expect(result).toEqual(releases[3]); // "4th" option was picked
+  expect(result).toEqual(releases[5]); // "6th" option was picked
```
gastoner, line 125: Update name of the test as well?

extensions/kubectl-cli/src/download.ts
```diff
@@ -47,6 +47,7 @@ export class KubectlDownload {
   async promptUserForVersion(currentKubectlTag?: string): Promise<KubectlGithubReleaseArtifactMetadata> {
     // Get the latest releases
     let lastReleasesMetadata = await this.kubectlGitHubReleases.grabLatestsReleasesMetadata();
+    lastReleasesMetadata.sort((a, b) => rcompare(a.tag, b.tag));
```
gastoner, line 50: Different PR?

gastoner (approved):

extensions/kubectl-cli/src/download.ts
axel7083, line 50: avoid using `.sort` please have a preference for `.toSort`

benoitf: why should we not use sort ? here we want to sort the current object, not get a copy

jeffmaury (approved):
Code LGTM but PR title is misleading

extensions/kubectl-cli/src/download.ts
axel7083: IMHO this is to be avoid has it change the object itself,  which in this case is fine, but could have unwanted side effect in other scenarios

axel7083: This is not blocking I do not have a strong opinion

benoitf: I think the wrong pattern is

```
const sorted = myArray.sort()
```

but this is is monitored by https://sonarsource.github.io/rspec/#/rspec/S4043/javascript

axel7083: Oh yes I see, thanks

dgolovin:
Updated name to match the actual changes

extensions/kubectl-cli/src/download.ts
dgolovin: Updated PR name to include both changes.

PR #9411, opened 2024-10-16
feat: make Kubernetes informers cancellable

feloy:

### What does this PR do?

For #6114 , we will need to restart the informers every time we change the current context, to change their backoff configuration.

Working on this, I realized that the informers were not correctly stopped when not necessary anymore. 

This PR makes the informers cancellable, so we can make them stop their infinite loop of retries.

Part of #6114

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-manager.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager.ts
@@ -60,7 +60,7 @@ import type { ApiSenderType } from '../api.js';
 import { Backoff } from './backoff.js';
 import { backoffInitialValue, backoffJitter, backoffLimit, connectTimeout } from './contexts-constants.js';
 import { ContextsInformersRegistry } from './contexts-informers-registry.js';
-import type { ContextInternalState } from './contexts-states-registry.js';
+import type { CancellableInformer, ContextInternalState } from './contexts-states-registry.js';
 import { ContextsStatesRegistry, dispatchAllResources, isSecondaryResourceName } from './contexts-states-registry.js';
 import { ResourceWatchersRegistry } from './resource-watchers-registry.js';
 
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -917,6 +927,9 @@ export class ContextsManager {
         return;
       }
       options.timer = setTimeout(() => {
+        if (options.cancel) {
```
axel7083: From my understanding this do not cancel the setTimeout, it just set a `cancel` property, and wait for it to be triggered, then the first line are to check if it has been cancelled.

packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -364,7 +366,7 @@ export class ContextsManager {
     };
   }
 
-  private createPodInformer(kc: KubeConfig, namespace: string, context: KubeContext): Informer<V1Pod> {
+  private createPodInformer(kc: KubeConfig, namespace: string, context: KubeContext): CancellableInformer {
```
axel7083, line 367: Why are we losing the generic type, is it important ?

packages/main/src/plugin/kubernetes/contexts-manager.ts
feloy: Yes, right, this is what is does. But you are right cancelling the timer directly would be simpler. I'll change

packages/main/src/plugin/kubernetes/contexts-manager.ts
feloy: I'll check again, does it does not seem to be useful

feloy: I don't think so:

The `V1Pod` is necessary when calling `this.createInformer<V1Pod>`, to have he `onAdd` and other options have the matching type. But the result of `createPodInformer` is stored in an `Informer<KubernetesObject>`; the specific type is never used on this side.

axel7083: Okey thanks

packages/main/src/plugin/kubernetes/contexts-manager.ts
axel7083: Okey thanks

axel7083 (approved):
LGTM Codewise

benoitf (approved):

PR #9412, opened 2024-10-16
fix(ui): layers explorer margin issue

### LINKED ISSUE #9410: bug(layer explorer): tree issue
### Bug description

Today I was using the file explorer to find why the instructlab image is so large, and I found a visual glitch with the tree explorer when we reach some depth.

https://github.com/user-attachments/assets/bb256506-29f1-4c70-8072-26c95821389a

### Operating system

fedora 40

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. reach any depth above 4

### Relevant log output

_No response_

### Additional context

The problem is coming from the following line

https://github.com/containers/podman-desktop/blob/8f80bf47418f99ba118d8c865da5a13594de21a1/packages/renderer/src/lib/image/FilesystemLayerView.svelte#L69

and

https://github.com/containers/podman-desktop/blob/8f80bf47418f99ba118d8c865da5a13594de21a1/packages/renderer/src/lib/image/FilesystemLayerView.svelte#L80

Using `ml-${margin}` will not work, because tailwindcss do not create value when they are not available at compile time, therefore for `margin === 10`, it do not create the corresponding class. Leading to css not applied

axel7083:
### What does this PR do?

Avoid using runtime classes of tailwindcss, since tailwindcss do not generate classes if do not exists at compile time[^1]

[^1]: https://stackoverflow.com/questions/69687530/dynamically-build-classnames-in-tailwindcss

### Screenshot / video of UI

| Before | After |
| --- | --- |
| image | image |

Fixes https://github.com/containers/podman-desktop/issues/9410

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/FilesystemLayerView.svelte
+++ b/packages/renderer/src/lib/image/FilesystemLayerView.svelte
@@ -77,7 +77,7 @@ function getLink(file: ImageFile | undefined): string {
       {/if}
     {:else}
       <div class={`${colorClass}`}>
-        <span style="margin-left: {margin}rem" class={`pl-4`}></span>
+        <span style="margin-left: {margin}rem" class="pl-4"></span>
         {label}<span class="text-[var(--pd-content-text)] opacity-70">{getLink(tree?.data)}</span>
       </div>
     {/if}
--- a/packages/renderer/src/lib/image/FilesystemLayerView.svelte
+++ b/packages/renderer/src/lib/image/FilesystemLayerView.svelte
@@ -58,7 +58,7 @@ function getLink(file: ImageFile | undefined): string {
   {#if root}
     {#if children}
       {#each children as [_, child]}
-        <svelte:self root={false} margin={margin + 2} tree={child} layerMode={layerMode} />
+        <svelte:self root={false} margin={margin + 0.5} tree={child} layerMode={layerMode} />
       {/each}
     {/if}
   {:else}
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/image/FilesystemLayerView.svelte
```diff
@@ -58,26 +58,26 @@ function getLink(file: ImageFile | undefined): string {
...
       {#if expanded && children}
         {#each children as [_, child]}
-          <svelte:self root={false} margin={margin + 2} tree={child} layerMode={layerMode} />
+          <svelte:self root={false} margin={margin + 0.5} tree={child} layerMode={layerMode} />
         {/each}
       {/if}
     {:else}
       <div class={`${colorClass}`}>
-        <span class={`pl-4 ml-${margin}`}></span>
+        <span style="margin-left: {margin}rem" class={`pl-4`}></span>
```
feloy:
```suggestion
        <span style="margin-left: {margin}rem" class="pl-4"></span>
```

feloy (approved):
thanks for the fix

benoitf (commented):
wondering why we're kind using a flat display to represent a tree

Like why not using ul/li or parent/child ?

I mean, why there is a always a bigger shift from the left
I would expect that we only add a child that has always the same shift but as we're imbricating children, it does the tree /shift

feloy:
> wondering why we're kind using a flat display to represent a tree
> 
> Like why not using ul/li or parent/child ?
> 
> I mean, why there is a always a bigger shift from the left I would expect that we only add a child that has always the same shift but as we're imbricating children, it does the tree /shift

I remember trying this solution, but also encountering some other problem (sorry, cannot remember which one)

axel7083:
> Like why not using ul/li or parent/child ?

my proposal is to merge this as it fixes easily the issue, and we can open a technical issue on this one, is that okay with you @benoitf ?

benoitf:
sure I'm not requesting changes

PR #9427, opened 2024-10-17
chore: replace colors by using registry colors in service, task-manager, troubleshooting, and window-control-buttons

SoniaSandler:

### What does this PR do?
*Needs to be merged after https://github.com/containers/podman-desktop/pull/9424
Updates any bg- and text- hardcoded colors to use the color registry

Changes to the color registry:
The services have assigned colors in PD, so they also have been added to the registry based on `ServiceColumnType.svelte`:
- ClusterIP - sky[500]
- LoadBalancer - purple[500]
- NodePort - fuschia[600]
- unknown - gray

I have also added `title-hover-bg` so that the circle will be visible when hovering the buttons in the titlebar after changing it to use the color registry

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -258,6 +258,7 @@ export class ColorRegistry {
     this.initTerminal();
     this.initProgressBar();
     this.initBadge();
+    this.initProviders();
   }
 
   protected initDefaults(): void {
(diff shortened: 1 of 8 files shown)


SoniaSandler:
@deboer-tim this PR includes new colors in the registry for services, providers, and the titlebar hover bg.
Based on https://github.com/containers/podman-desktop/pull/9420#issuecomment-2423121485, do we want to keep the provider colors hardcoded? and as for the service colors, should I just move them to the badge section?

SoniaSandler:
@deboer-tim I've updated the service colors to have more general names and be in the badge section. lmk if I should keep the providers' colors hardcoded or if having them in the colors registry is ok.

deboer-tim:
I still don't love having three provider colors (podman, docker, Kubernetes) and putting these in the registry; if we did then I think we should at least use the correct colors ourselves vs the palette. This might be a good conversation for the UX call though, maybe I am a minority or someone has a better idea.

IMHO this PR is still too big though, I would recommend to keep splitting out the obvious goodness & cleanup into separate PRs.

benoitf (approved):

feloy (approved):

SoniaSandler:
rebased

Issue #9379, opened 2024-10-15
Proxy Settings stucked in "System" mode


rv2931:
### Bug description

I have installed tje 1.13.2 podman desktop version with 5.1.1 podman version (not the latest) and I use my machine sometimes with VPN&Proxy sometimes without
I have to switch of Proxy States. With new versino the Proxy Settings pannel has changed. As previous version of Desktop (1.12) seems to detect proxy configuration from OS I tried to switch to "System" Proxy settings but now it is impossible de switch back to "Manual" or "Disabled".
I got to settings, proxy settings, switch to manual for example, graphically the menu shows "manual" but no widget to enter proxy settings appears and if close settings and come back, "system" is still selected

### Operating system

Windows 11

### Installation Method

Installer from website/GitHub releases

### Version

1.13.2

### Steps to reproduce

- Go to settings
- Go to "Proxy"
- Set Proxy setting to "System"
- Close "Settings"
- Maybe you have to restart padman
- Go to settings
- Go to "Proxy"
- Set Proxy setting to "Manual"
- You can see that nothing change in the panel, no widget added to enter proxy settings
- Close settings pannel
- Go to settings
- Go to "Proxy"
- The settings is still "System" instead of manual

### Additional context

![Image](https://github.com/user-attachments/assets/0077223f-5497-4bf0-956b-02df44e41aca)

FrCornaire:
Hello
After some digging, found the culprit

The TLDR is a value not being updated ,hence the app looking "stuck"

A quick fix would be to update your podman-desktop settings at 

`%userprofile%\.local\share\containers\podman-desktop\configuration\settings.json`

and add the value proxy.enabled: 

```json
{
... //Your current settings

"proxy.enabled" : 1
}
```

Made a PR at #9442  for a fix

PR #9442, opened 2024-10-17
fix: proxy settings state not updating on change

FrCornaire:
Signed-off-by: Francis CORNAIRE <francis.cornaire@axa.fr>

### What does this PR do?

Fix Proxy Settings state not being updated when the UI state changed

### Screenshot / video of UI

Current behavior
theProblem

The fix
theFix

#9379

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
@@ -66,14 +66,26 @@ function validate(event: any) {
       <select
         class="p-2 outline-none text-sm bg-charcoal-600 rounded-sm text-gray-700 placeholder-gray-700"
         id="toggle-proxy"
-        bind:value={proxyState}>
+        bind:value={proxyState}
+        on:change={async () => {
+          await window.setProxyState(proxyState);
+
+          if (proxyState === ProxyState.PROXY_MANUAL && !proxySettings) {
+            proxySettings = {
+              httpProxy: '',
+              httpsProxy: '',
+              noProxy: '',
+            };
+          }
+        }}
+        >
         <option value={ProxyState.PROXY_SYSTEM}>System</option>
         <option value={ProxyState.PROXY_MANUAL}>Manual</option>
         <option value={ProxyState.PROXY_DISABLED}>Disabled</option>
       </select>
     </label>
 
-    {#if proxySettings}
+    {#if proxySettings && proxyState === ProxyState.PROXY_MANUAL}
       <div class="space-y-2">
         <label
           for="httpProxy"


axel7083:
Hi @FrCornaire thanks for the pull request and the fix!

With the fix you are providing, the inputs are often visible, often not. This is because of the following line

https://github.com/containers/podman-desktop/blob/109741f9571b67989a5226d4d6201e98874bbef0/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte#L88

where we only display the proxy inputs if the config exists, but as you are updating it, we may need to add a condition:

```diff
- {#if proxySettings} 
+ {#if proxySettings && proxyState === ProxyState.PROXY_MANUAL} 
``` 

This would make the proxy inputs only visible if we selected the `PROXY_MANUAL` manual state. Do you think you can make the change ?

FrCornaire:
> Hi @FrCornaire thanks for the pull request and the fix!
> 
> With the fix you are providing, the inputs are often visible, often not. This is because of the following line
> 
> https://github.com/containers/podman-desktop/blob/109741f9571b67989a5226d4d6201e98874bbef0/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte#L88
> 
> where we only display the proxy inputs if the config exists, but as you are updating it, we may need to add a condition:
> 
> ```diff
> - {#if proxySettings} 
> + {#if proxySettings && proxyState === ProxyState.PROXY_MANUAL} 
> ```
> 
> This would make the proxy inputs only visible if we selected the `PROXY_MANUAL` manual state. Do you think you can make the change ?

Sure!

jeffmaury:
> Hi @FrCornaire thanks for the pull request and the fix!
> 
> With the fix you are providing, the inputs are often visible, often not. This is because of the following line
> 
> https://github.com/containers/podman-desktop/blob/109741f9571b67989a5226d4d6201e98874bbef0/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte#L88
> 
> where we only display the proxy inputs if the config exists, but as you are updating it, we may need to add a condition:
> 
> ```diff
> - {#if proxySettings} 
> + {#if proxySettings && proxyState === ProxyState.PROXY_MANUAL} 
> ```
> 
> This would make the proxy inputs only visible if we selected the `PROXY_MANUAL` manual state. Do you think you can make the change ?

We need to change the test as if system is in place (default) proxySettings may be undefined in case nothing is configured on system level. That would be the only required change

jeffmaury (changes requested):
Thank you for the collaboration but this breaks the existing logic has everythink is collected locally and configuration is changed only when user clicks the Update button

FrCornaire:
> Thank you for the collaboration but this breaks the existing logic has everythink is collected locally and configuration is changed only when user clicks the Update button

Well my first approach was to save every time the state changed but it might be a bit too much .

slemeur:
Thanks for the PR, very nice @FrCornaire !

benoitf:
outdated

PR #9443, opened 2024-10-17
refactor: simplify context states update

feloy:

### What does this PR do?

Working on #6114, informers need to be restarted in more circumstances, to be able to change their backoff configuration.

After a few iterations working on these informers, the logic to detect which informers need to be started/stopped is more and more complex, and there are still missing conditions.

With this PR, I propose to simplify the logic, with:

- when the kubeconfig file changes:
  - all the informers (primary and secondary) for all contexts are stopped
  - primary informers for the current context are started
  - if there are less than 11 contexts, primary contexts for non-current contexts are started

- when the current context becomes reachable:
  - if the frontend subscribed to the secondary informers, the secondary informers for this current context are started (1)

- when the current context becomes unreachable: 
  - the secondary informers for this current context are stopped
 
- when the frontend subscribes to the secondary informers:
  - if the current context is reachable, the secondary informers for the current context are started (2)

> Note that (1) and (2) have the same result, it depends on which one happens first.
> 
> - Primary contexts are used to display the number of pods / deployments in each context in the Settings > Kubernetes page
> - Secondary informers are used to display the resources information in the Kubernetes pages (Nodes, Services, ...)

Part of #6114 

### How to test this PR?


benoitf:
unit tests and type check is failing maybe a draft pr ?

packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -965,7 +882,11 @@ export class ContextsManager {
       return [];
     }
     console.debug(`start watching ${resourceName} in context ${this.currentContext}`);
-    this.startResourceInformer(this.currentContext.name, resourceName);
+    // start secondary informers only if current context is reachable
+    const currentContextState = this.states.getCurrentContextGeneralState(this.currentContext.name);
+    if (currentContextState.reachable) {
```
deboer-tim: I'm having trouble following the async states - how do we know the reachable state of the current context will be valid / updated when this function is called, and isn't that basically the same as the call to .isReachable() a few lines up?

packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -138,97 +136,38 @@ export class ContextsManager {
...
+      await this.informers.deleteContextInformers(name);
+      await this.states.deleteContextState(name);
+      // secondaries
+      await this.informers.disposeSecondaryInformers(name);
+      await this.states.disposeSecondaryStates(name);
     }
 
-    // Add informers for new contexts (only current context if we are checking only it)
-    const added: string[] = [];
+    // Add primary informers for contexts (only current context if we are checking only it)
```
deboer-tim: Double use of 'only' is confusing to read here

packages/main/src/plugin/kubernetes/contexts-manager.ts
feloy: yes, you are right this is the same, I'll change to simplify

deboer-tim (approved):

Issue #7399, opened 2024-05-31
New task toast notification


deboer-tim:
### Is your enhancement related to a problem? Please describe

When a task is created the status bar task icon adds a notification dot, but if you haven't clicked on it the second created task doesn't do anything. It's not very visible to the user that 'something is now running in the background'.

### Describe the solution you'd like

When a task is created there should be a toast popup at the lower right. We'll need design input on what it looks like and to decide if it stays up until the user closes it/task finishes or just for a few seconds.

axel7083:
Might need to wait for https://github.com/containers/podman-desktop/pull/8144 to be merged

PR #9446, opened 2024-10-17
chore: remove timer to refresh tasks in frontend side

benoitf:
### What does this PR do?
there is a timer refreshing every second the tasks but AFAIK it's useless as now tasks are managed on the backend side

part of #7399

### How to test this PR?

play with tasks being created


axel7083 (approved):
Oh yes, thanks for the cleanup/catch

deboer-tim (approved):

PR #9451, opened 2024-10-17
chore: remember last child page visited for kubernetes and preferences pages

### LINKED ISSUE #9057: Child navigation should return to previous page

When you click on Kubernetes and then Deployments, then go to another page and back to Kubernetes, it forgets which page you were on and goes back to the default (which is no page currently).

Navigation should remember the last 'child' page you were on. e.g. if you go from Kubernetes > Deployments to Images, and then click on Kubernetes again, the Deployments page should still be selected.

The same should probably be true for Settings? I often exit Settings > Preferences to check something and it's a lot of clicks to get back.

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
This PR adds stores to keep track of the last sections visited in submenu pages, so that when leaving and coming back, the user will come back to the last section they visited in the same submenu (can check with Kubernetes for now)

### Screenshot / video of UI

[Screencast from 2024-10-21 00-38-16.webm](https://github.com/user-attachments/assets/906d4488-8b33-41e3-a2c1-95a991d2a9bd)

Closes https://github.com/containers/podman-desktop/issues/9057

### How to test this PR?
Enter the Kubernetes (nodes) page, go to one of the other pages in Kubernetes (e.g. Deployments), exit to some other page in PD that is not in Kubernetes, and then go back to Kubernetes. You should be back in your last visited section page (e.g. Deployments)

### DIFF
--- a/packages/renderer/src/SubmenuNavigation.svelte
+++ b/packages/renderer/src/SubmenuNavigation.svelte
@@ -22,6 +22,7 @@ export let meta: TinroRouteMeta;
   </div>
   <div class="h-full overflow-hidden hover:overflow-y-auto" style="margin-bottom:auto">
     {#each items ?? [] as item}
+      {console.log(item.name + ' in submenunav')}
       <SettingsNavItem title={item.tooltip} href={item.link} selected={meta.url.startsWith(item.link)}
       ></SettingsNavItem>
     {/each}
--- a/packages/renderer/src/SubmenuNavigation.svelte
+++ b/packages/renderer/src/SubmenuNavigation.svelte
@@ -12,8 +12,7 @@ export let meta: TinroRouteMeta;
 
 let pages = $lastSubmenuPages;
 if (!pages[title]) {
-  let defaultLink = link === '/kubernetes' ? '/kubernetes/nodes' : link;
-  pages[title] = defaultLink;
+  pages[title] = link;
   lastSubmenuPages.set(pages);
 }
 </script>
--- a/packages/renderer/src/SubmenuNavigation.svelte
+++ b/packages/renderer/src/SubmenuNavigation.svelte
@@ -31,7 +31,7 @@ if (!pages[title]) {
   </div>
   <div class="h-full overflow-hidden hover:overflow-y-auto" style="margin-bottom:auto">
     {#each items ?? [] as item}
-      <SettingsNavItem title={item.tooltip} href={item.link} selected={meta.url.startsWith(item.link)} onClick={() => {pages[title] = item.link;}}
+      <SettingsNavItem title={item.tooltip} href={item.link} selected={meta.url.startsWith(item.link)} onClick={() => pages[title] = item.link}
       ></SettingsNavItem>
     {/each}
   </div>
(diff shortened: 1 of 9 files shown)


SoniaSandler:
@deboer-tim I wasn't sure how specific we want the last visited page to be, so for now it is the very last one (could be summary/details page in kubernetes, or `create new` pages in preferences), but if needed, I can change it to just keep track of the last general section visited (Nodes, Deployments, Services, Preferences, CLI Tools, etc.)

gastoner (approved):
LGTM, works well!

deboer-tim (changes requested):
This change seems very hard-coded, i.e. every time we add or change a submenu we also need to update App, Route, breadcrumb, and associated tests. I haven't looked at the code, but is there something we could do that is more generic / based on NavigationRegistryEntry submenus? Settings likely isn't a submenu today but could be (and that part could be a separate PR)

I was hoping for something along the lines of:
- enter submenu for first time -> pick first page
- navigate within submenu -> store path for next entry

Failing tests need to be fixed.

SoniaSandler:
@deboer-tim since settings is not yet a submenu, should I keep the changes for it or remove them and just update the submenu?

deboer-tim:
> @deboer-tim since settings is not yet a submenu, should I keep the changes for it or remove them and just update the submenu?

I'd suggest checking if we can do something simpler/less hard-coded for submenus first. If so then yes I'd break it up, otherwise we can re-evaluate.

SoniaSandler:
@deboer-tim I've changes the PR to work for submenus and removed the changes for the preferences section, but now the last page visited is less specific. For example, when coming back to the nodes page, it will go to the node list page, even if the last visited Kubernetes page was a specific node, but I can modify it to be more specific if needed

cdrage:
Silly question, but with this PR, unsure if this should be done within this PR or as a follow-up, but what about remembering sections in extensions that use SubmenuNavigation imported? (ex bootc extension)

SoniaSandler:
@cdrage I looked in the bootc repo (same for ai-lab), and it looks like it only uses `SettingsNavItem` from this PR. I think that for extensions we'll have to update their navigation files manually to also remember last child page.
TBH, I'm not sure if there is a way for me to generalize it more so that it would also automatically include extensions' submenus. 
For now, this PR should work for all submenus created/rendered with `SubmenuNavigation.svelte`, but I can explore the topic of the extension submenus more if needed.

packages/renderer/src/SubmenuNavigation.svelte
```diff
@@ -2,11 +2,20 @@
...
 export let items: NavigationRegistryEntry[] | undefined;
+export let link: string;
 export let meta: TinroRouteMeta;
+
+let pages = $lastSubmenuPages;
+if (!pages[title]) {
+  let defaultLink = link === '/kubernetes' ? '/kubernetes/nodes' : link;
+  pages[title] = defaultLink;
+  lastSubmenuPages.set(pages);
+}
```
benoitf: AFAIK we can't have anything here related to kubernetes

benoitf (approved):

deboer-tim (approved):
Thank you for the changes, this commit is much better and works well in testing.

cdrage (approved):
Works well for me! LGTM and great work.

PR #9452, opened 2024-10-18
chore: update container engine selects to Dropdown component

### LINKED ISSUE #8283: Selection in dark mode UI has no background
### Bug description

Light mode:

Screenshot from 2024-07-30 09-19-21

Dark mode:
Screenshot from 2024-07-30 09-19-47

The selection background disappears / has issues showing correctly.

### Operating system

mac M1

### Installation Method

None

### Version

1.11.1

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Updates all the container engine select controls [*] to the Dropdown component, plus another select that was in KubePlayYaml. An odd set, but the code in each of these is almost identical.

I kept this to just the component swap as much as possible, but almost all of these pages had existing formatting inconsistencies, with the container engine standing out the most. In each case I tried to make the minimum formatting change to allow the container engine option to fit in or fix existing formatting to be more consistent. There are still issues here, but doing any less looked odd.

[*] I left out CreateVolume because I'm still deciding how to do the tests, will PR later.

Noticed an issue while using Dropdown: when objects are passed via html parameters like 'option value={obj}', Svelte puts the stringified value in .value and keeps the real value in .__value. Although it hasn't changed in several years, use of this property is not Svelte API (https://github.com/sveltejs/svelte/issues/2734), which means we can't use option child elements when using non-string values. I just switched to the options parameter instead, which is what most third-party components use.

### Screenshot / video of UI

Screenshot 2024-10-17 at 12 20 56 PM
Screenshot 2024-10-17 at 12 22 13 PM
Screenshot 2024-10-17 at 12 22 39 PM
Screenshot 2024-10-17 at 12 23 27 PM
Screenshot 2024-10-17 at 12 24 36 PM
Screenshot 2024-10-17 at 12 26 01 PM

Fixes #8283.

### How to test this PR?

Make sure you have two container engines, then go to each of these pages.

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -3,7 +3,7 @@
 // https://github.com/import-js/eslint-plugin-import/issues/1479
 import { faCube, faMinusCircle, faPlusCircle } from '@fortawesome/free-solid-svg-icons';
 import type { OpenDialogOptions } from '@podman-desktop/api';
-import { Button, Input } from '@podman-desktop/ui-svelte';
+import { Button, Dropdown, Input } from '@podman-desktop/ui-svelte';
 import type { Terminal } from '@xterm/xterm';
 import { onDestroy, onMount } from 'svelte';
 import { get } from 'svelte/store';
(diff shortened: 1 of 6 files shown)


feloy (approved):

cdrage (approved):
LGTM. Tests pass (tests already written would of covered obviously the dropdown).

No changes suggested regarding code-wise, simple swapin for the Dropdown.

Pulled the PR and went through kubeplayyaml + pod create and all looked good!

PR #9481, opened 2024-10-21
chore: add experimental property for toasts

benoitf:
### What does this PR do?
add property in backend for toasts

part of https://github.com/containers/podman-desktop/issues/7399

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/tasks/task-manager.ts
+++ b/packages/main/src/plugin/tasks/task-manager.ts
@@ -62,6 +62,11 @@ export class TaskManager {
             type: 'boolean',
             default: false,
           },
+          [`${ExperimentalTasksSettings.SectionName}.${ExperimentalTasksSettings.Toast}`]: {
+            description: 'Display a notification toast when task is created',
+            type: 'boolean',
+            default: false,
+          },
         },
       },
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/tasks/task-manager.ts
```diff
@@ -62,6 +62,11 @@ export class TaskManager {
             type: 'boolean',
             default: false,
           },
+          [`${ExperimentalTasksSettings.SectionName}.${ExperimentalTasksSettings.Toast}`]: {
+            description: 'Display a notification toast when task is created',
+            type: 'boolean',
+            default: false,
+          },
```
axel7083, line 69: not hidden ?

benoitf: it seems that 'description: 'Show running tasks in the status bar',' is not hidden as well

so either both of them need to be hidden or all of them need to be visible

axel7083: Oh yeah I thought it was hidden, weird

axel7083 (approved):
LGTM

Issue #8341, opened 2024-08-02
Kubernetes pages should have a refresh button when cluster connection is not established 


axel7083:
### Is your feature request related to a problem? Please describe

After seeing https://github.com/containers/podman-desktop/pull/8340 I directly saw the need for kubernetes page to have similar feature.

As we often show an empty screen when the cluster is `not reachable`

![image](https://github.com/user-attachments/assets/3f32ad38-23be-46fb-b210-6648653c8287)

But when you start manually the cluster, you often have to wait some time , or even force the context switch to trigger a refresh/connection to the cluster

### Describe the solution you'd like

Same as https://github.com/containers/podman-desktop/pull/8340 a button to refresh, retry the connection to the kubernetes cluster (current context)

deboer-tim:
+1. However, unlike #8340 where there is only a single attempt, this is polling and likely just not frequent enough / happening when you know the Kubernetes context state or connection to it has changed. Once you actually connect to a context it should be live.

So, I'd prefer this to be implemented more like a 'will try again in 30s' with a 'try now' button that disappears when it is connected.

axel7083:
I am assigning it myself for next sprint :)

axel7083:
![image](https://github.com/user-attachments/assets/7625f697-ee67-4ff6-8853-ef7974488be3)

deboer-tim:
I like the simplicity, but it's not super obvious that I can click on it to refresh. (we have lots of Labels elsewhere that look almost identical but you can't click) Maybe if the refresh icon looked more like one of the column actions? Colored and on a little circle?

axel7083:
I moved this issue to `Postponed` because the work required is quick a lot, and the way the informer system is designed does not allow easily manual intervention.

In the `kubernetes-context-state.ts` there is a `restartInformer` method with the following signature

https://github.com/containers/podman-desktop/blob/4c147ed43308839c03ac76d648fd03213f3ac8dc/packages/main/src/plugin/kubernetes/kubernetes-context-state.ts#L1094-L1098

However the `options: CreateInformerOptions<T>` are never saved anywhere, they are always passed down through function scope and function argument callback. Meaning, for an external caller this is never available.

Moreover this is a problem, as each options is specific to resource type `pod`, `deployment` etc.

---

Moreover, the reconnection timeout id is saved inside this option object iself

[...]

PR #9487, opened 2024-10-21
feat: refresh the state of a specific context by restarting its informers

feloy:

### What does this PR do?

Adds a method to `window` to restart informers for a specific context, from the frontend.

(frontend part to be done in another PR)

Part of #8341 

### How to test this PR?

#9491 can be used to test from the frontend

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -2519,6 +2519,10 @@ export class PluginSystem {
       },
     );
 
+    this.ipcHandle('kubernetes-client:refreshContextState', async (_listener, context: string): Promise<void> => {
+      return kubernetesClient.refreshContextState(context);
+    });
+
     this.ipcHandle('feedback:send', async (_listener, feedbackProperties: unknown): Promise<void> => {
       return telemetry.sendFeedback(feedbackProperties);
     });
--- a/packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
@@ -2930,4 +2930,37 @@ describe('refreshContextState', () => {
     await client.refreshContextState('context1');
     expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-checking-state-update', expectedMap);
   });
+
+  test('should throw an error if the context does not exist', async () => {
+    vi.mocked(makeInformer).mockImplementation(fakeMakeInformer);
+    const client = new TestContextsManager(apiSender);
+    const kubeConfig = new kubeclient.KubeConfig();
+    const config = {
+      clusters: [
+        {
+          name: 'cluster1',
+          server: 'server1',
+        },
+      ],
+      users: [
+        {
+          name: 'user1',
+        },
+      ],
+      contexts: [
+        {
+          name: `context1`,
+          cluster: 'cluster1',
+          user: 'user1',
+        },
+      ],
+      currentContext: 'context1',
+    };
+
+    kubeConfig.loadFromOptions(config);
+    await client.update(kubeConfig);
+    await expect(async () => await client.refreshContextState('unknown-context')).rejects.toThrowError(
+      'context unknown-context not found',
+    );
+  });
 });
--- a/packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
@@ -2884,3 +2884,50 @@ describe('isContextChanged', () => {
     expect(routeInformer).toHaveBeenCalledOnce();
   });
 });
+
+describe('refreshContextState', () => {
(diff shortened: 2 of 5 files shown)


packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
```diff
@@ -2884,3 +2884,50 @@ describe('isContextChanged', () => {
...
+    vi.advanceTimersToNextTimer();
+    expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-checking-state-update', expectedMap);
+    apiSenderSendMock.mockClear();
+    vi.advanceTimersByTime(9000);
+    expect(apiSenderSendMock).not.toHaveBeenCalled();
+    vi.advanceTimersByTime(1100);
+    expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-checking-state-update', expectedMap);
+    apiSenderSendMock.mockClear();
+    await client.refreshContextState('context1');
+    expect(apiSenderSendMock).toHaveBeenCalledWith('kubernetes-contexts-checking-state-update', expectedMap);
```
benoitf, line 2931: looks like it's not checking when the context we ask to refresh is not there

image

but then, I'm wondering, should it be a no-op or an error to ask to refresh something that is not existing ?

packages/main/src/plugin/index.ts
```diff
@@ -2519,6 +2519,10 @@ export class PluginSystem {
       },
     );
 
+    this.ipcHandle('kubernetes-client:refreshContextState', async (_listener, context: string): Promise<void> => {
```
axel7083, line 2522: I am bit confused with the wording, the PR title says `restart` and here we have `refresh` is there a semantic reason   ?

feloy: I would say that one is what is seen from the caller, and the other is what is implemented: the state is refreshed by restarting the informers

packages/main/src/plugin/kubernetes/contexts-manager.spec.ts
feloy: Yes, that's a good point, I'm not sure either. I have chosen no-op here, as IMO there is no chance that the refresh is asked if the context does not exist (but it will depend on the frontend implementation).

And if really the context has been deleted in the meantime, it seems to me that it would not be an error, because the user just doesn't care anymore about it.

packages/main/src/plugin/index.ts
axel7083: what the user should see ? a button with restart or refresh ?

feloy: I think "refresh", as the user doesn't care how it is implemented, by restarting informers, or reinitializing them, etc

feloy: (I have renamed the PR to clarify)

benoitf (approved):
would throw an error in case the context is not found to have feedback from user's POV

if I execute an order I would like to know if the order has been executed or not (and not just ignored)

packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -976,6 +976,25 @@ export class ContextsManager {
     return [];
   }
 
+  public async refreshContextState(contextName: string): Promise<void> {
```
axel7083, line 987: please add some documentation above, this seems to be the entrypoint of the logic

packages/main/src/plugin/kubernetes/kubernetes-client.ts
```diff
@@ -1726,4 +1726,8 @@ export class KubernetesClient {
     // pod.metadata?.labels && 'pod-template-hash' in pod.metadata.labels
     return podMetadata.ownerReferences?.find((ref: V1OwnerReference) => ref.controller === true);
   }
+
+  public async refreshContextState(context: string): Promise<void> {
```
axel7083, line 1741: or here ?

axel7083 (approved):
LGTM, just some documentation above the functions added would be a +

benoitf (approved):
re-approving

feloy:
> LGTM, just some documentation above the functions added would be a +

I have added doc on both methods, with a different level of details depending on the method

packages/main/src/plugin/kubernetes/kubernetes-client.ts
```diff
@@ -1726,4 +1726,19 @@ export class KubernetesClient {
     // pod.metadata?.labels && 'pod-template-hash' in pod.metadata.labels
     return podMetadata.ownerReferences?.find((ref: V1OwnerReference) => ref.controller === true);
   }
+
+  /**
+   * Ask forr getting the state of the context as soon as possible.
```
axel7083:
```suggestion
   * Ask for getting the state of the context as soon as possible.
```

axel7083 (approved):
LGTM, thanks for the documentation!

PR #9488, opened 2024-10-21
feat: add experimental flag to display toasts when we have notifications

### LINKED ISSUE #7399: New task toast notification

benoitf:
### What does this PR do?
Allow to display a toast when tasks are created

### Screenshot / video of UI

https://github.com/user-attachments/assets/b1a8ceed-b975-4252-86d0-f295bcb0b2ad

fixes https://github.com/containers/podman-desktop/issues/7399

### How to test this PR?

Enable the flag in preferences screen then open long taks with progress (like bootc build image) or stuff from AI Lab

### DIFF
--- a/packages/renderer/src/lib/toast/ToastCustomUi.svelte
+++ b/packages/renderer/src/lib/toast/ToastCustomUi.svelte
@@ -1,12 +1,18 @@
 <script lang="ts">
-import { faCheckCircle, faCircleExclamation, faSpinner } from '@fortawesome/free-solid-svg-icons';
-import { CloseButton, Link } from '@podman-desktop/ui-svelte';
+import { faCheckCircle, faCircleExclamation } from '@fortawesome/free-solid-svg-icons';
+import { CloseButton, Link, Spinner } from '@podman-desktop/ui-svelte';
 import { toast } from '@zerodevx/svelte-toast';
 import Fa from 'svelte-fa';
 
 import type { TaskInfo } from '/@api/taskInfo';
 
-let { toastId, taskInfo, onpop = () => {} }: { toastId: number; taskInfo: TaskInfo; onpop?: () => void } = $props();
+interface Props {
+  toastId: number;
+  taskInfo: TaskInfo;
+  onpop?: () => void;
+}
+
+let { toastId, taskInfo, onpop = () => {} }: Props = $props();
 
 const closeAction = (): void => {
   toast.pop(toastId);
@@ -19,7 +25,8 @@ const executeAction = (): void => {
 </script>
 
 <div
-  class="flex min-h-10 max-h-30 flex-col p-2 border-[var(--pd-button-tab-border-selected)] border rounded-md bg-[var(--pd-modal-bg)]"
+  class="flex min-h-10 cursor-default max-h-30 max-w-[var(--toastWidth)] flex-col p-2 border-[var(--pd-button-tab-border-selected)] border rounded-md bg-[var(--pd-modal-bg)]"
+  title={taskInfo.name}
 >
   <div class="mb-1 flex flex-row items-center">
     <div
@@ -28,7 +35,7 @@ const executeAction = (): void => {
       aria-label={taskInfo.status}
     >
       {#if taskInfo.status === 'in-progress'}
-        <Fa class="animate-spin" icon={faSpinner} />
+        <Spinner size="1em"/>
       {:else if taskInfo.status === 'success'}
         <Fa icon={faCheckCircle} />
       {:else if taskInfo.status === 'failure'}
@@ -36,7 +43,7 @@ const executeAction = (): void => {
       {/if}
     </div>
 
-    <div class="font-bold text-ellipsis line-clamp-1 max-w-32">
+    <div class="font-bold text-ellipsis line-clamp-1 overflow-hidden">
       {taskInfo.name}
     </div>
 
@@ -48,13 +55,15 @@ const executeAction = (): void => {
       <CloseButton on:click={closeAction} />
     </div>
   </div>
-  <div class="flex flex-row items-center italic">
+  <div class="flex flex-row items-center italic line-clamp-4">
     {#if taskInfo.error}
       <p class="flex-1 text-sm line-clamp-4 text-[var(--pd-state-error)]">
         {taskInfo.error}
       </p>
     {:else}
+    <p class="flex-1 text-sm text-ellipsis overflow-hidden">
       {taskInfo.name}
+    </p>
     {/if}
   </div>
   {#if taskInfo.action}
--- a/packages/renderer/src/lib/toast/ToastCustomUi.svelte
+++ b/packages/renderer/src/lib/toast/ToastCustomUi.svelte
@@ -1,4 +1,3 @@
-
 <script lang="ts">
 import { faCheckCircle, faCircleExclamation, faSpinner } from '@fortawesome/free-solid-svg-icons';
 import { CloseButton, Link } from '@podman-desktop/ui-svelte';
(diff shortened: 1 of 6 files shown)


packages/renderer/src/lib/toast/ToastTaskNotifications.svelte
```diff
@@ -0,0 +1,111 @@
...
+  // grab the tasks that are not in the currentTasks
+  const newTasks = tasks.filter(task => !currentTasks.find(currentTask => currentTask.id === task.id));
+
+  // get the existing one that we should update
+  const toUpdateTasks = tasks.filter(task => currentTasks.find(currentTask => currentTask.id === task.id));
+
+  // get deleted tasks
+  const toDeleteTasks = currentTasks.filter(currentTask => !tasks.find(task => task.id === currentTask.id));
+
+  console;
```
feloy: to remove?

packages/renderer/src/lib/toast/ToastTaskNotifications.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+          taskInfo: IN_PROGRESS_TASK,
+        }),
+        sendIdTo: 'toastId',
+        src: expect.anything(),
+      },
+    }),
+  );
+});
+
+test('Check no toast is being if disabled', async () => {
```
feloy:
```suggestion
test('Check no toast is being created if disabled', async () => {
```

packages/renderer/src/lib/toast/ToastTaskNotifications.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+  // make it disabled
+  vi.mocked(window.getConfigurationValue).mockResolvedValue(false);
+
+  tasksInfo.set([]);
```
feloy: shouldn't we have a task here?

packages/renderer/src/lib/toast/ToastTaskNotifications.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  render(ToastTaskNotifications, {});
+
+  // check we have called toast library to create a toast
```
feloy:
```suggestion
  // check we have not called toast library to create a toast
```

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -0,0 +1,64 @@
+
+<script lang="ts">
```
feloy, line 1: It seems to have a problem with the indentation of this file

benoitf:
pr updated

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -0,0 +1,65 @@
...
+import { faCheckCircle, faCircleExclamation, faSpinner } from '@fortawesome/free-solid-svg-icons';
+import { CloseButton, Link } from '@podman-desktop/ui-svelte';
+import { toast } from '@zerodevx/svelte-toast';
+import Fa from 'svelte-fa';
+
+import type { TaskInfo } from '/@api/taskInfo';
+
+let { toastId, taskInfo, onpop = () => {} }: { toastId: number; taskInfo: TaskInfo; onpop?: () => void } = $props();
+
+const clicked = (): void => {
```
feloy: this name is not really meaningful. Could we name it `close` instead?

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -0,0 +1,65 @@
...
+    {#if taskInfo.error}
+      <p class="flex-1 text-sm line-clamp-4 text-[var(--pd-state-error)]">
+        {taskInfo.error}
+      </p>
+    {:else}
+      {taskInfo.name}
+    {/if}
+  </div>
+  <div class="text-right text-xs text-[var(--pd-content-text)]">
+    {#if taskInfo.action}
```
feloy: I suppose we can move the `if` one level up, to avoid an empty `div` if there is no action

benoitf: it was to avoid content scrolling, but I've moved one level up

benoitf:
pr updated

axel7083 (commented):
Should we open a https://github.com/containers/podman-desktop/discussions for the experimental tasks feature ?

I feel like having a dedicated forum to discuss the design would be nice

axel7083 (approved):
Since https://github.com/containers/podman-desktop/issues/9501 is open, IMO we can merge this PR and iterate through discussion

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -0,0 +1,65 @@
+<script lang="ts">
+import { faCheckCircle, faCircleExclamation, faSpinner } from '@fortawesome/free-solid-svg-icons';
+import { CloseButton, Link } from '@podman-desktop/ui-svelte';
+import { toast } from '@zerodevx/svelte-toast';
+import Fa from 'svelte-fa';
+
+import type { TaskInfo } from '/@api/taskInfo';
+
+let { toastId, taskInfo, onpop = () => {} }: { toastId: number; taskInfo: TaskInfo; onpop?: () => void } = $props();
```
axel7083:
```suggestion
interface Props {
    toastId: number;
    taskInfo: TaskInfo;
    onpop?: () => void;
}

let { toastId, taskInfo, onpop = () => {} }: Props = $props();
```

I feel like it is easier to read and understand the components props

deboer-tim (changes requested):
The right edge of the toast is missing for me:
task

Toasts 'conflict' with the task manager, i.e. if you have a toast open you cannot see parts of the task manager. Maybe this is somewhat intentional but it feels unplanned, like if a form had a submit button you can't get to.

Personally I find the toasts redundant with the status bar, and feels fairly redundant with the task manager too - why do we show tasks in both and have two popups for the same thing? - but since this is experimental / for feedback I would not hold for that reason.

UI/usability nits:
- The toasts never go away, even after the task is finished (maybe I was supposed to click on the X, but I couldn't...).
- The spinner is 'wrong', doesn't use the normal one.
- The title ellipsis is too short, cropping the title unnecessarily (and shorter than what we have on the status bar).

benoitf:
> Should we open a https://github.com/containers/podman-desktop/discussions for the experimental tasks feature ?
> 
> I feel like having a dedicated forum to discuss the design would be nice

UX issues have a new template since few weeks so I think the overall idea was to get it in the main tracker

It's just that here the topic was discussed in some calls but while I was assigned I discovered there was no formal ticket, so it's like using the cart before the horse

benoitf:
@deboer-tim you would need to provide a reproducer as trying with `registry.gitlab.com/fedora/bootc/base-images/fedora-bootc-minimal` (or other images) I always have the border
image


about the z-index, there are pros/cons. Because there is no stack ranking like last click win for setting the order on the z-axis. It's always the same.
so if the z-index of the toast is lower, then the task manager will be displayed on top of the notifications

deboer-tim:
> @deboer-tim you would need to provide a reproducer as trying with `registry.gitlab.com/fedora/bootc/base-images/fedora-bootc-minimal` (or other images) I always have the border

Figured it out - pull an image like registry.gitlab.com/fedora/bootc/examples/httpd and try to bootc build it, or just start any task that has a long name with no ' ' or '-', i.e. no obvious place to insert a LF. Just needs an ellipsis, wider toast, or better splitting.

> about the z-index, there are pros/cons. Because there is no stack ranking like last click win for setting the order on the z-axis. It's always the same. so if the z-index of the toast is lower, then the task manager will be displayed on top of the notifications

Agreed. IMHO there is functional overlap too, so less an issue for this as experimental PR and more 'if users like this, we probably need to revisit how this works with task manager before removing experimental'.

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -0,0 +1,65 @@
...
+      {#if taskInfo.status === 'in-progress'}
+        <Fa class="animate-spin" icon={faSpinner} />
+      {:else if taskInfo.status === 'success'}
+        <Fa icon={faCheckCircle} />
+      {:else if taskInfo.status === 'failure'}
+        <Fa icon={faCircleExclamation} class="text-[var(--pd-state-error)]" />
+      {/if}
+    </div>
+
+    <div class="font-bold text-ellipsis line-clamp-1 max-w-32">
```
deboer-tim: Task name is using ellipsis at just past half-width, even when there is no percentage. Remove `max-w-32` here. Might need a grow, or conversely also a max-w on the whole card.

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -0,0 +1,65 @@
...
+
+    {#if taskInfo.progress && taskInfo.status === 'in-progress'}
+      <span class="ml-1">{taskInfo.progress}%</span>
+    {/if}
+
+    <div class="flex flex-grow flex-col items-end">
+      <CloseButton on:click={closeAction} />
+    </div>
+  </div>
+  <div class="flex flex-row items-center italic">
```
deboer-tim: Maybe text-sm? I noticed the font is relatively big.

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -0,0 +1,65 @@
...
+  class="flex min-h-10 max-h-30 flex-col p-2 border-[var(--pd-button-tab-border-selected)] border rounded-md bg-[var(--pd-modal-bg)]"
+>
+  <div class="mb-1 flex flex-row items-center">
+    <div
+      class="mr-1 text-purple-500"
+      role="status"
+      aria-label={taskInfo.status}
+    >
+      {#if taskInfo.status === 'in-progress'}
+        <Fa class="animate-spin" icon={faSpinner} />
```
deboer-tim: Should use Spinner.

deboer-tim (commented):
Noticed another minor issue: if there are completed tasks that haven't been removed from the task manager and I enable this preference, then all tasks appear when I start a new one. (Less likely, but likewise if I close all toasts, disable the preference for a bit, then turn it on and create one new task, all toasts reappear)

benoitf:
pr updated

feloy (approved):

deboer-tim (approved):
Tested, confirmed the blocking usability issues are fixed, thanks!

slemeur:
The PR can be merged in the current state. This will be enabled with an experimental flag.
We'll continue to iterate and phase potential improvements.

PR #9491, opened 2024-10-21
feat: refresh button in Kubernetes empty pages

### LINKED ISSUE #8341: Kubernetes pages should have a refresh button when cluster connection is not established

feloy:

### What does this PR do?

Frontend part of #8341 

Adds a `Refresh` button to the Kubernetes page when the context is not reachable.

The button is disabled and in progress when the context connectivity is being checked.

### Screenshot / video of UI

https://github.com/user-attachments/assets/7559d508-d055-44ec-82f4-aa76726170e1

Fixes #8341 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
@@ -2,9 +2,13 @@
 import { EmptyScreen } from '@podman-desktop/ui-svelte';
 
 import ConfigMapSecretIcon from '../images/ConfigMapSecretIcon.svelte';
+import KubernetesCheckConnection from '../ui/KubernetesCheckConnection.svelte';
 </script>
 
 <EmptyScreen
   icon={ConfigMapSecretIcon}
   title="No configmaps or secrets"
-  message="Try switching to a different context or namespace" />
+  message="Try switching to a different context or namespace">
+  <KubernetesCheckConnection />
+</EmptyScreen>
+
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
@@ -1,14 +1,10 @@
 <script lang="ts">
-import { EmptyScreen } from '@podman-desktop/ui-svelte';
-
 import ConfigMapSecretIcon from '../images/ConfigMapSecretIcon.svelte';
-import KubernetesCheckConnection from '../ui/KubernetesCheckConnection.svelte';
+import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
 </script>
 
-<EmptyScreen
+<KubernetesEmptyScreen
   icon={ConfigMapSecretIcon}
   title="No configmaps or secrets"
-  message="Try switching to a different context or namespace">
-  <KubernetesCheckConnection />
-</EmptyScreen>
+  message="Try switching to a different context or namespace" />
 
--- a/packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
+++ b/packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
@@ -1,7 +1,7 @@
 <script lang='ts'>
 import { EmptyScreen } from '@podman-desktop/ui-svelte';
 
-import KubernetesCheckConnection from '../ui/KubernetesCheckConnection.svelte';
+import KubernetesCheckConnection from '/@/lib/ui/KubernetesCheckConnection.svelte';
 
 export let icon: any;
 </script>
--- a/packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
+++ b/packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
@@ -1,0 +1,11 @@
+<script lang='ts'>
+import { EmptyScreen } from '@podman-desktop/ui-svelte';
+
+import KubernetesCheckConnection from '../ui/KubernetesCheckConnection.svelte';
+
+export let icon: any;
+</script>
+
+<EmptyScreen icon={icon} {...$$props}>
+  <KubernetesCheckConnection />
+</EmptyScreen>
--- a/packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
+++ b/packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
@@ -1,11 +1,12 @@
 <script lang='ts'>
 import { EmptyScreen } from '@podman-desktop/ui-svelte';
+import type { ComponentProps } from 'svelte';
 
 import KubernetesCheckConnection from '/@/lib/ui/KubernetesCheckConnection.svelte';
 
-export let icon: any;
+let { icon, ...restProps }: ComponentProps<EmptyScreen> = $props();
 </script>
 
-<EmptyScreen icon={icon} {...$$props}>
+<EmptyScreen icon={icon} {...restProps}>
   <KubernetesCheckConnection />
 </EmptyScreen>
--- a/packages/renderer/src/lib/node/NodeEmptyScreen.svelte
+++ b/packages/renderer/src/lib/node/NodeEmptyScreen.svelte
@@ -1,11 +1,9 @@
 <script lang="ts">
-import { EmptyScreen } from '@podman-desktop/ui-svelte';
-
 import { kubernetesCurrentContextState } from '/@/stores/kubernetes-contexts-state';
 import type { ContextGeneralState } from '/@api/kubernetes-contexts-states';
 
 import NodeIcon from '../images/NodeIcon.svelte';
-import KubernetesCheckConnection from '../ui/KubernetesCheckConnection.svelte';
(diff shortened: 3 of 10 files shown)


packages/renderer/src/lib/node/NodeEmptyScreen.svelte
```diff
@@ -17,6 +18,15 @@ function getText(state: ContextGeneralState | undefined): string {
...
+function checkContext(): void {
+  if (currentContextName) {
+    window.kubernetesRefreshContextState(currentContextName);
+  }
+}
 </script>
 
-<EmptyScreen icon={NodeIcon} title="No nodes" message={text} />
+<EmptyScreen icon={NodeIcon} title="No nodes" message={text}>
+  <button onclick={checkContext}>Check connection now</button>
```
benoitf: I think we don't need the now as the button implies that it would be when we click on it

method being refresh then maybe just `Refresh`

benoitf: also do we need to see the button if there is no context ? as we only do the refresh in case of a context

feloy: This is really to test the backend part, I'm not sure at all if it will be the implementation we want to do in the frontend

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
```diff
@@ -2,9 +2,13 @@
...
 import ConfigMapSecretIcon from '../images/ConfigMapSecretIcon.svelte';
+import KubernetesCheckConnection from '../ui/KubernetesCheckConnection.svelte';
 </script>
 
 <EmptyScreen
   icon={ConfigMapSecretIcon}
   title="No configmaps or secrets"
-  message="Try switching to a different context or namespace" />
+  message="Try switching to a different context or namespace">
+  <KubernetesCheckConnection />
```
benoitf: suggestion: I'm wondering if we should introduce a `<KubernetesEmptyScreen>` extending EmptyScreen but then adding the refresh option,

so if I add a new page, it'll be automatically updated and we won't forget to add the extra < KubernetesCheckConnection> and duplicate it everywhere

axel7083:
This should be a draft while there is no backend ?

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
```diff
@@ -0,0 +1,27 @@
...
+
+import {
+  kubernetesContextsCheckingStateDelayed,
+  kubernetesCurrentContextState,
+} from '/@/stores/kubernetes-contexts-state';
+
+import { kubernetesContexts } from '../../stores/kubernetes-contexts';
+
+let currentContextName: string | undefined;
+$: currentContextName = $kubernetesContexts.find(c => c.currentContext)?.name;
```
benoitf: suggestion: use new svelte5 runes

```suggestion
let currentContextName = $derived($kubernetesContexts.find(c => c.currentContext)?.name)
```

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
```diff
@@ -0,0 +1,27 @@
...
+
+function refresh(): void {
```
benoitf, line 14: issue: AFAIK kubernetesRefreshContextState is an async method so there is a missing await or a missing catch

slemeur:
Let's move forward with the proposal. This greatly improve the experience for the end-user.   
If there is an automatic attempt to reconnect to the cluster, then, the connectivity indicator will display activity with (#9535) and the button can't be clicked while that is happening. This is completely fine.

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
feloy: Good idea, will try in a follow-up PR

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
feloy: Will change to svelte5 in a follow-up PR

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
benoitf: IMHO  you'll modify the same amount of files doing directly the change in this PR

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
benoitf: you only have to click on the button :-)

feloy: But it would be a mix between svelte 4 and 5, no?

I would also like to change the `$:`s by `$state`s

But that's not a big deal. I'll make the changes here

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
feloy: done

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
feloy: done

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang='ts'>
+import { EmptyScreen } from '@podman-desktop/ui-svelte';
+
+import KubernetesCheckConnection from '../ui/KubernetesCheckConnection.svelte';
```
benoitf: suggestion(non-blocking): use the `/@/lib`

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
```diff
@@ -0,0 +1,11 @@
...
+
+export let icon: any;
+</script>
+
+<EmptyScreen icon={icon} {...$$props}>
+  <KubernetesCheckConnection />
+</EmptyScreen>
```
benoitf: todo: please add a test for this file

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
benoitf: thanks

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
benoitf: thanks

benoitf (approved):
only missing a unit test for the new component

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang='ts'>
+import { EmptyScreen } from '@podman-desktop/ui-svelte';
+
+import KubernetesCheckConnection from '/@/lib/ui/KubernetesCheckConnection.svelte';
+
+export let icon: any;
+</script>
+
+<EmptyScreen icon={icon} {...$$props}>
```
axel7083: I am not sure svelte is able to infer properly the props of the `KubernetesEmptyScreen` as an inherit component of EmptyScreen here ? since we just pass down the props, (might be wrong)

In svelte 5 the following is possible

```svelte
<script lang='ts'>
import { EmptyScreen } from '@podman-desktop/ui-svelte';
import KubernetesCheckConnection from '/@/lib/ui/KubernetesCheckConnection.svelte';

let { icon, ...restProps }: ComponentProps<EmptyScreen> = $props();
</script>

<EmptyScreen icon={icon} {...restProps}>...</EmptyScreen>
```

feloy: The title and message properties are correctly displayed (and the unit tests also assert they are passed to `EmptyScreen`). Or am I misunderstanding your point?

benoitf: `$$props` is a legacy/v4 property

benoitf: explanation: we're running in legacy mode, so it works but if we disable legacy mode, it won't

axel7083 (approved):

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
feloy: ok, I'll change to Svelte 5, thanks

axel7083 (approved):

PR #9504, opened 2024-10-21
refactor: context checking state as derived store

feloy:

### What does this PR do?

In the Settings > Kubernetes page, a spinner is displayed when the connectivity is being checked,to inform the user when the checks happen. A delay of 2 seconds is used, so the spinner is displayed for at least 2 seconds, to make it visible for the user. 

The state used to display the spinner with this delay was computed in the component. This PR moves this computation in a store, so the store can be reused in other pages (in the Kubernetes pages not to name them).

### Screenshot / video of UI

https://github.com/user-attachments/assets/5aaa142f-868b-4c21-85fd-e6b4633dd97a

Part of #8341

### How to test this PR?


benoitf (approved):
it's a move but this block of code is really hard to understand
also there is even a setTimeout inside it, lot of almost duplicated lines

deboer-tim (approved):
Ditto to @benoitf's comments - easy enough to approve code-wise, but maybe something to improve later.

'CheckingStateDelayed' in the name sounds to me like the user won't see it right away. Buffered isn't quite right either. Maybe like 'ExtendedCheckingState'?

PR #9510, opened 2024-10-21
chore: use dropdown component in Kube terminal

deboer-tim:
### What does this PR do?

Switches the select in the Kube terminal to a Dropdown.

The select control automatically selects the first value if none is provided. Dropdown wasn't doing that so I added the appropriate line, hoping it is small enough to justify including it here.

Remove redundant aria-labelledby (label already points to control).

### Screenshot / video of UI

Screenshot 2024-10-21 at 3 36 06 PM

Part of #6199.

### How to test this PR?

Deploy a pod to Kubernetes with more than one container, go to pod's Terminal tab.

### DIFF
--- a/packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
+++ b/packages/renderer/src/lib/pod/KubernetesTerminalBrowser.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { EmptyScreen } from '@podman-desktop/ui-svelte';
+import { Dropdown, EmptyScreen } from '@podman-desktop/ui-svelte';
 import { onMount } from 'svelte';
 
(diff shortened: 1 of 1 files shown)


benoitf (changes requested):
the PR is not only using the dropdown in Kube terminal but changing the behavior of the dropdown


--> needs to be in a separate PR

deboer-tim:
> the PR is not only using the dropdown in Kube terminal but changing the behavior of the dropdown
> 
> --> needs to be in a separate PR

👍🏼. Broke out #9537 and will rebase this on top.

benoitf (approved):

deboer-tim:
Rebased on main and down to only the one change, ready for re-review.

jeffmaury (approved):
LGTM

PR #9523, opened 2024-10-22
fix: work on related context, not current context

feloy:

### What does this PR do?

#9443 introduced a problem, the secondary informers were deleted in the current context at the time of stopping them, not on the context the informers were created for. In case the user switches from a non reachable context to a reachable one, the informers for the now accessible current context were deleted.

Part od #6114 

### How to test this PR?

Have 2 contexts, one reachable and the other non reachable.
Switch from 1 context to the other, and check that the resources (nodes, ...) for the reachable one remain visible.


axel7083 (commented):
~Nevermind~

axel7083 (approved):
LGTM tested as suggested with two contexts and works fine

benoitf (approved):

PR #9524, opened 2024-10-22
fix: proxy settings page stuck

### LINKED ISSUE #9379: Proxy Settings stucked in "System" mode

jeffmaury:
Fixes #9379

### What does this PR do?

Update the Proxies preference page

### Screenshot / video of UI

N/A

#9379 

### How to test this PR?

Go to the proxies preference page and change to either manual or disabled

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
@@ -73,7 +73,7 @@ function validate(event: any) {
       </select>
     </label>
 
-    {#if proxySettings}
+    {#if proxyState !== ProxyState.PROXY_SYSTEM}
       <div class="space-y-2">
         <label
           for="httpProxy"
--- a/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 import { faPen } from '@fortawesome/free-solid-svg-icons';
-import type { ProxySettings } from '@podman-desktop/api';
 import { Button, ErrorMessage, Input } from '@podman-desktop/ui-svelte';
 import { onMount } from 'svelte';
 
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -73,7 +73,7 @@ function validate(event: any) {
       </select>
     </label>
 
-    {#if proxySettings}
+    {#if proxyState !== ProxyState.PROXY_SYSTEM}
```
axel7083: if I have selected `manual` then switch back to system, I will not see the update button, therefore I will not be able to update the settings ?

axel7083 (changes requested):
image

The page crash on my machine

image

Because this is not the real type

axel7083 (changes requested):
image

The page crash on my machine

image

Because this is not the real type
https://github.com/containers/podman-desktop/blob/cd668fe99f28d75c4029f5479e581db212c0ed8f/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte#L12C5-L12C18

it is `let proxySettings: ProxySettings | undefined;` and we are reading the value of undefined for the inputs

axel7083 (approved):
If we are in a hurry to release this is an acceptable solution 👍 , and we could do follow up PR to improve the design.

We have a dedicated issue for testing https://github.com/containers/podman-desktop/issues/9525

## UX remark

image

The `Update` button is always active, but should be disabled after we save: if we have nothing to update , the button should be disabled, this is missleading to always see it visible.

I still believe that disabled and system should hide the inputs and automatically save through the select on:change property. Avoiding the user to have to click on update after  setting to system or disabled.

jeffmaury:
> If we are in a hurry to release this is an acceptable solution 👍 , and we could do follow up PR to improve the design.
> 
> We have a dedicated issue for testing #9525
> ## UX remark
> 
> image
> 
> The `Update` button is always active, but should be disabled after we save: if we have nothing to update , the button should be disabled, this is missleading to always see it visible.
> 
> I still believe that disabled and system should hide the inputs and automatically save through the select on:change property. Avoiding the user to have to click on update after setting to system or disabled.

That is to be discussed and out of scope of this PR.

gastoner (approved):
LGTM

PR #9528, opened 2024-10-22
feat: add a way to clean logs in container's log page

### LINKED ISSUE #8311: add ability to clean the logs

Reported by feedback form, user wants the ability to be able to clean the logs (of a container for example)

It was to be able to debug more easily

add a way to clean the logs

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
add a way to clear logs in container's logs

### Screenshot / video of UI

https://github.com/user-attachments/assets/182ecb60-3667-449b-929b-6dea500dd95c

fixes https://github.com/containers/podman-desktop/issues/8311

### How to test this PR?

start a container, go to the logs and then click on the clear button

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsLogsClear.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsLogsClear.svelte
@@ -1,0 +1,20 @@
+<script lang="ts">
+import { faEraser } from '@fortawesome/free-solid-svg-icons';
+import type { Terminal } from '@xterm/xterm';
+import Fa from 'svelte-fa';
+
+const { terminal }: { terminal: Terminal } = $props();
+
+function clear(): void {
+  terminal.clear();
+}
+</script>
+
+<div class="absolute top-0 right-0 px-1 z-50 m-1 opacity-50 space-x-1">
+  <button title="Clear logs" onclick={clear}>
+    <Fa
+      class="cursor-pointer rounded-full bg-[var(--pd-button-disabled)] min-h-8 w-8 p-1.5"
+      icon={faEraser}
+    />
+  </button>
+</div>
--- a/packages/renderer/src/lib/container/ContainerDetailsLogsClear.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsLogsClear.svelte
@@ -10,10 +10,10 @@ function clear(): void {
 }
 </script>
 
-<div class="absolute top-0 right-0 px-1 z-50 m-1 opacity-50 space-x-1">
-  <button title="Clear logs" onclick={clear}>
+<div class="absolute top-0 right-2 px-1 z-50 m-1 opacity-50 space-x-1">
+  <button title="Clear logs" onclick={clear} class="">
     <Fa
-      class="cursor-pointer rounded-full bg-[var(--pd-button-disabled)] min-h-8 w-8 p-1.5"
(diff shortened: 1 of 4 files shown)


danivilla9:
I just tried it on Fedora 40, and looks like it works to me. The 'Clear logs' button is overlapped by the scrollbar, unlike on mac, though:
Screenshot from 2024-10-22 16-53-06
And when changing to a different tab inside of container details and back to Logs the 'deleted' text appears again. I suppose this is expected, right?

benoitf:
before implementing I tried several approach and I found out that switching back and getting the logs again was allowing people to get back the log.
Originally I explored the idea of setting a timestamp after which we're asking logs, but then, there was not an easy way to reset this one. So I had the feeling the approach chosen in this PR was less intrusive.

about the overlap of the scrollbar, I can try to move it more to the left

packages/renderer/src/lib/container/ContainerDetailsLogsClear.svelte
```diff
@@ -0,0 +1,20 @@
...
+import Fa from 'svelte-fa';
+
+const { terminal }: { terminal: Terminal } = $props();
+
+function clear(): void {
+  terminal.clear();
+}
+</script>
+
+<div class="absolute top-0 right-0 px-1 z-50 m-1 opacity-50 space-x-1">
```
deboer-tim: The first Logs I went to I wouldn't have known the button was there if I wasn't looking. Maybe a little more opacity?
Screenshot 2024-10-22 at 3 51 42 PM

Likewise, it is over top of the mem tooltip. z-order should be lowered so that tooltips remain on top.
Screenshot 2024-10-22 at 3 53 41 PM

packages/renderer/src/lib/container/ContainerDetailsLogsClear.svelte
```diff
@@ -0,0 +1,20 @@
...
+  <button title="Clear logs" onclick={clear}>
+    <Fa
+      class="cursor-pointer rounded-full bg-[var(--pd-button-disabled)] min-h-8 w-8 p-1.5"
```
deboer-tim: Missing a hover effect.

The icon isn't as bright as the actions on top, but if I remove the opacity it is too bright. I think it's missing a text- and regular opacity would help with the previous comment.

deboer-tim (commented):
Mac scrollbars are narrower so that bit wasn't an issue for me, but I noticed some other minor UI issues as noted.

The 'reappearing logs' were also a surprise to me, and rules out uses like 'clear logs, go do X, come back to see what logs were generated'. Instead of a timestamp would 'remove top X lines' be any simpler, or is it clearing it the issue?

jeffmaury (approved):
LGTM

The icon overloaps on the scrollbar on MacOS can me move it a little bit on the left ?

axel7083 (approved):
Feature is working as described

axel7083:
Need rebase (conflict)

benoitf:
putting to draft as some fixes have been requested

benoitf:
icon moved to the left, added hover state

for the tooltip z order issue, I'll open another PR for the tooltip as in fact if I put z-order below tooltip then we can't click on the button as it's behind the terminal lines.

so it's the tooltip that need to be at an higher z-axis

axel7083 (approved):
re-approving, tested and works as expected :+1: 

image

PR #9533, opened 2024-10-22
chore(test): implement aria-label for dropdown button and update UI tests

### LINKED ISSUE #9532: Implement aria-label for dropdown button and update UI tests

With the recent UI changes from a combo box to a button for the provider type menu, our e2e tests that use the Kind cluster are broken.

- Update the locator to use the new button.
- Assign the proper aria-label to this button.
- Modify our UI tests to verify the accessibility of the button using its aria-label. https://github.com/containers/podman-desktop/pull/9520#discussion_r1810367941

_No response_

### Additional context

Related to: https://github.com/containers/podman-desktop/issues/9522

amisskii:

### What does this PR do?

implements aria-label for dropdown button and updates UI tests

https://github.com/containers/podman-desktop/issues/9532

### How to test this PR?

pnpm test:ui

### DIFF
--- a/packages/ui/src/lib/dropdown/Dropdown.spec.ts
+++ b/packages/ui/src/lib/dropdown/Dropdown.spec.ts
@@ -27,7 +27,7 @@ import DropdownTest from './DropdownTest.svelte';
 test('a button is created', async () => {
   render(Dropdown, { ariaLabel: 'Dropdown' });
 
-  const input = screen.getByRole('button', { name: 'Dropdown Button' });
+  const input = screen.getByRole('button', { name: 'Dropdown' });
   expect(input).toBeInTheDocument();
 });
 
@@ -37,7 +37,7 @@ test('initial value is visible', async () => {
     ariaLabel: 'Dropdown',
   });
 
-  const input = screen.getByRole('button', { name: 'Dropdown Button' });
+  const input = screen.getByRole('button', { name: 'Dropdown' });
   expect(input).toBeInTheDocument();
   expect(input).toHaveTextContent('a value');
 });
@@ -45,7 +45,7 @@ test('initial value is visible', async () => {
 test('opening dropdown does not submit forms (prevents default button action)', async () => {
   render(Dropdown, { ariaLabel: 'Dropdown' });
 
-  const input = screen.getByRole('button', { name: 'Dropdown Button' });
+  const input = screen.getByRole('button', { name: 'Dropdown' });
   expect(input).toBeInTheDocument();
 
   const event = createEvent.click(input);
@@ -59,7 +59,7 @@ test('disabling changes state and styling', async () => {
     ariaLabel: 'Dropdown',
   });
 
-  const input = screen.getByRole('button', { name: 'Dropdown Button' });
+  const input = screen.getByRole('button', { name: 'Dropdown' });
   expect(input).toBeInTheDocument();
   expect(input).toBeDisabled();
   expect(input).toHaveClass('text-[color:var(--pd-input-field-disabled-text)]');
@@ -69,13 +69,13 @@ test('disabling changes state and styling', async () => {
 test('initial focus is not set by default', async () => {
   render(Dropdown, { ariaLabel: 'Dropdown' });
 
-  const input = screen.getByRole('button', { name: 'Dropdown Button' });
+  const input = screen.getByRole('button', { name: 'Dropdown' });
   expect(input).not.toHaveFocus();
 });
 
 test('should be able to navigate with keys', async () => {
   render(DropdownTest);
-  const input = screen.getByRole('button', { name: 'Dropdown Button' });
+  const input = screen.getByRole('button', { name: 'Dropdown' });
   expect(input).toBeInTheDocument();
   expect(input).toHaveTextContent('initial value');
   input.focus();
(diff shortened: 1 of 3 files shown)


packages/ui/src/lib/dropdown/Dropdown.spec.ts
```diff
@@ -25,56 +25,57 @@ import Dropdown from './Dropdown.svelte';
...
 
-  const input = screen.getByRole('button');
+  const input = screen.getByRole('button', { name: 'Dropdown' });
   expect(input).toBeInTheDocument();
 });
 
 test('initial value is visible', async () => {
   render(Dropdown, {
     value: 'a value',
+    ariaLabel: 'Dropdown',
```
benoitf, line 37: @deboer-tim wondering why the ariaLabel is not using the name as default aria-label but use ''

https://github.com/containers/podman-desktop/blob/main/packages/ui/src/lib/dropdown/Dropdown.svelte#L21

packages/ui/src/lib/dropdown/Dropdown.spec.ts
```diff
@@ -25,56 +25,57 @@ import Dropdown from './Dropdown.svelte';
...
   expect(input).toBeInTheDocument();
   expect(input).toHaveTextContent('a value');
 });
 
 test('opening dropdown does not submit forms (prevents default button action)', async () => {
-  render(Dropdown);
+  render(Dropdown, { ariaLabel: 'Dropdown' });
 
-  const input = screen.getByRole('button');
+  const input = screen.getByRole('button', { name: 'Dropdown' });
```
benoitf, line 48: I don't understand why we need to replace all the getByRole in all the tests ?

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -192,6 +191,7 @@ function onWindowClick(e: Event): void {
     disabled={disabled}
     id={id}
     name={name}
+    aria-label={ariaLabel}
```
benoitf, line 194: why the aria-label attribute is moved ?

amisskii: I don't think we need an aria-label for the div, only for the button for testing purposes. The aria-label on the div seems unnecessary.

packages/ui/src/lib/dropdown/Dropdown.spec.ts
amisskii: I thought we wanted to check the accessibility of the button with the aria-label in unit tests.

packages/ui/src/lib/dropdown/Dropdown.spec.ts
deboer-tim: I couldn't find a way to make it work with Svelte 5 props, kept getting errors whenever it included a - or tried to do something like className on :19. Docs and examples for Svelte 5 still sparse so couldn't confirm if there should be a way to do it or not.

packages/ui/src/lib/dropdown/Dropdown.svelte
deboer-tim: Do you have some test code this is trying to fix or is it just an aria lookup checking the text content? There is one case in CreateVolume using userEvent.selectOptions(), but this change (alone) would not fix it.

amisskii: My second proposal is to add an aria-label to both the div and the button elements. Button will have `ariaLabel={'${aria-label} Button'}`

odockal: Does the button's div have any function other than a container? I am wondering why the container div of a button should have aria-label at all. It makes sense for some elements with role, ie. this button. Unless there is some use case.

benoitf:
After reading the comments, it seems that we're not focusing on the right pattern like where to put a name for a button.

IMHO the expectation is to have this component following the combobox semantic while it is not

I should not search for a button but for a combobox using aria-label

so the role 'combobox' should be added (and all the others like aria-expanded, etc)

then for the option, it's the same for the options, I don't see any `option` role

@deboer-tim why this component has no combobox/option aria-labels ? AFAIK it should be a drop-in replacement from tests framework from Select/Options to Dropdown (tests should not be changed)

deboer-tim:
> After reading the comments, it seems that we're not focusing on the right pattern like where to put a name for a button.
> 
> IMHO the expectation is to have this component following the combobox semantic while it is not
> 
> I should not search for a button but for a combobox using aria-label
> 
> so the role 'combobox' should be added (and all the others like aria-expanded, etc)

Yes, there needs to be a `role="combobox"` on the same element as `aria-label`. This will allow code like `screen.getByRole('combobox', { name: 'x' })` to work, but not methods `userEvent.selectOptions()`, which requires at least a couple other aria-labels.

> then for the option, it's the same for the options, I don't see any `option` role
> 
> @deboer-tim why this component has no combobox/option aria-labels ? AFAIK it should be a drop-in replacement from tests framework from Select/Options to Dropdown (tests should not be changed)

I had a commit that added all the labels as per an example on MDN, but vitest still failed. I went down the rabbit hole of starting to look at vitest source to see if it was actually hardcoded to select, and thought I should look at how UI libraries do it instead. I tried I couple that I had looked at prior to building this - and they literally had no aria support.

The first 10 instances I switched didn't use aria-labels at all and both unit and e2e tests passed, so I assumed it was a problem solvable later...

Short term, yes, let's do this and I'll be trying again for full aria. Long term, we should probably make sure that all locators are part of PR check?

PR #9535, opened 2024-10-22
feat: display a spinner when connectivity is being checked in Kubernetes pages 

feloy:
### What does this PR do?

Display a spinner when connectivity is being checked in Kubernetes pages.

Similarly to what is done in the page Settings > Kubernetes

### Screenshot / video of UI

https://github.com/user-attachments/assets/45ae9401-3716-4464-a240-80e5a90615df

Part of #8341 

### How to test this PR?

Have a non reachable context, and go to the different Kubernetes pages, to check that the spinner is displayed regularly

### DIFF
--- a/packages/renderer/src/lib/ui/KubernetesCurrentContextConnectionBadge.svelte
+++ b/packages/renderer/src/lib/ui/KubernetesCurrentContextConnectionBadge.svelte
@@ -21,7 +21,7 @@ function getClassColor(state?: ContextGeneralState): string {
 }
 
 $: text = getText($kubernetesCurrentContextState);
-$: currentContextName = $kubernetesContexts.find(c => c.currentContext)?.name;
+$: currentContextName = $kubernetesContexts?.find(c => c.currentContext)?.name;
 </script>
 
 {#if $kubernetesCurrentContextState}
--- a/packages/renderer/src/lib/ui/KubernetesCurrentContextConnectionBadge.svelte
+++ b/packages/renderer/src/lib/ui/KubernetesCurrentContextConnectionBadge.svelte
@@ -1,5 +1,11 @@
 <script lang="ts">
-import { kubernetesCurrentContextState } from '/@/stores/kubernetes-contexts-state';
+import { Spinner } from '@podman-desktop/ui-svelte';
+
+import { kubernetesContexts } from '/@/stores/kubernetes-contexts';
+import {
+  kubernetesContextsCheckingStateDelayed,
+  kubernetesCurrentContextState,
+} from '/@/stores/kubernetes-contexts-state';
 import type { ContextGeneralState } from '/@api/kubernetes-contexts-states';
 
 import Label from './Label.svelte';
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/ui/KubernetesCurrentContextConnectionBadge.svelte
```diff
@@ -15,9 +21,15 @@ function getClassColor(state?: ContextGeneralState): string {
 }
 
 $: text = getText($kubernetesCurrentContextState);
+$: currentContextName = $kubernetesContexts?.find(c => c.currentContext)?.name;
```
axel7083, line 25:
```suggestion
let currentContextName: string | undefined;
$: currentContextName = $kubernetesContexts?.find(c => c.currentContext)?.name;
```

packages/renderer/src/lib/ui/KubernetesCurrentContextConnectionBadge.svelte
```diff
@@ -15,9 +21,15 @@ function getClassColor(state?: ContextGeneralState): string {
...
 {#if $kubernetesCurrentContextState}
-  <Label role="status" name={text} tip={$kubernetesCurrentContextState.error}
-    ><div class="w-2 h-2 {getClassColor($kubernetesCurrentContextState)} rounded-full mx-1"></div></Label>
+  <div class="flex items-center flex-row">
+    {#if !!currentContextName && $kubernetesContextsCheckingStateDelayed?.get(currentContextName)}
+      <div class="mr-1"><Spinner size="12px"></Spinner></div>
+    {/if}
+    <Label role="status" name={text} tip={$kubernetesCurrentContextState.error}
+      ><div class="w-2 h-2 {getClassColor($kubernetesCurrentContextState)} rounded-full mx-1"></div></Label>  
+  </div>
```
axel7083, line 35: We should not display both the spinner and the status, should be either loading, either status, what do you think ?

cdrage: I was thinking the same thing, also why not just edit status button to be in-progress similar to our button UI?

feloy: I suppose there are many ways to do it. For now, I tried to make something very similar to what is in the Settings > Kubernetes page.

feloy: @ekidneyrh could you please have a look at this design?

deboer-tim: Maybe this should be on the empty screen where the focus is instead? We've also talked about a Try Again / Refresh Now option in the future, which would fit better there.

That said, it is better to merge this sprint (it's useful now; might get feedback) vs ideal UX, so if there is no immediate agreement on better & easily implementable I would approve as-is and we can continue discussion.

feloy: Right, I started to work on this refresh button in https://github.com/containers/podman-desktop/pull/9491

I wanted to have this indicator that the check is occurring, before to add a refresh button, which could help the user understand better what happens

benoitf (approved):
I see lot of improvements ideas in comments that can be follow-up PRs but is it better than today, yes, so here is my approval

slemeur:
This is a nice improvement already, showing that there is activity going on to check the connection with the kubernetes cluster. 
Let's move forward, the comments and enhancements are valuable but can be filled as other issues to address in the future.

slemeur:
LGTM

PR #9537, opened 2024-10-22
chore: default initial selection in dropdown

deboer-tim:
### What does this PR do?

The select control automatically selects the first value if none is provided. Dropdown wasn't doing that so I added the appropriate line.

### Screenshot / video of UI

N/A

Part of #6199. First part of #9510.

### How to test this PR?

PR checks.

### DIFF
--- a/packages/ui/src/lib/dropdown/Dropdown.svelte
+++ b/packages/ui/src/lib/dropdown/Dropdown.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 import { faCaretDown, faCheck } from '@fortawesome/free-solid-svg-icons';
-import type { Snippet } from 'svelte';
+import { onMount, type Snippet } from 'svelte';
 import Fa from 'svelte-fa';
 
 interface Option {
@@ -39,6 +39,12 @@ let highlightIndex: number = $state(-1);
 let comp: HTMLElement;
 
 const pageStep: number = 10;
+
+onMount(() => {
+  if (!value && options?.length > 0) {
+    value = options[0].value;
+  }
+});
 
 $effect(() => {
   selectLabel = options.find(o => o.value === value)?.label ?? (typeof value === 'string' ? (value as string) : '');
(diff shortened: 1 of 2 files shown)


axel7083 (approved):
LGTM

benoitf (approved):

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -40,6 +40,12 @@ let comp: HTMLElement;
 
 const pageStep: number = 10;
 
+onMount(() => {
+  if (!value && options?.length > 0) {
+    value = options[0].value;
+  }
```
benoitf, line 46: suggestion: wondering if it's in onMount or in effect

if you update the options it should refresh the initial value I suppose ?

benoitf: issue: it seems that if one option is removed, value will be invalid

so wondering if we don't need an 'initialValue' and a flag to check if the user has modified the value to either change it back to the initialValue when options are changing and if the user has not pickup anything

benoitf: issue: it looks like it's not reactive

deboer-tim: FWIW this was not meant to be reactive: we had ~15 instances of select, afaik only 1 was relying on this behaviour and not binding the value, and none are changing the options after creation. I can rework this to be reactive though.

benoitf: I'm not able to create a kind cluster on my end without the initial value, so probably this PR will fix the issue

I've approved, reactivity can be follow-up.

The problem here is that we're building a UI component in a library shared outside, so we never know all usecase that people will use. Just wanted to bring awareness that as option is a parameter, doing it only at the onMount event and not when data are changing may lead to an issue

deboer-tim: Ok, opened #9565 to follow up and self-assigned. Merging this, please shout if it doesn't solve the kind issue.

PR #9550, opened 2024-10-23
feat: provide ability to easily get a shell in a machine - backend changes - additional changes

gastoner:
### What does this PR do?
Provide the ability to ssh to podman machine - backedn changes

-Needs to be rebased
### Screenshot / video of UI

Part of #889 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -1158,6 +1158,7 @@ export class PluginSystem {
       {
         write: (param: string) => void;
         resize: (dimensions: containerDesktopAPI.ProviderConnectionShellDimensions) => void;
+        close: () => void;
       }
     >();
     this.ipcHandle(
@@ -1210,6 +1211,16 @@ export class PluginSystem {
         const callback = providerRegistryShellInProviderConnectionSendCallback.get(onDataId);
         if (callback) {
           callback.resize(dimensions);
+        }
+      },
+    );
+
+    this.ipcHandle(
+      'provider-registry:shellInProviderConnectionClose',
+      async (_listener, onDataId: number): Promise<void> => {
+        const callback = providerRegistryShellInProviderConnectionSendCallback.get(onDataId);
+        if (callback) {
+          callback.close();
         }
       },
     );
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -1219,9 +1219,7 @@ export class PluginSystem {
       'provider-registry:shellInProviderConnectionClose',
       async (_listener, onDataId: number): Promise<void> => {
         const callback = providerRegistryShellInProviderConnectionSendCallback.get(onDataId);
-        if (callback) {
-          callback.close();
-        }
+        callback?.close();
       },
     );
 
--- a/packages/main/src/plugin/provider-registry.spec.ts
+++ b/packages/main/src/plugin/provider-registry.spec.ts
@@ -1167,17 +1167,25 @@ describe('shellInProviderConnection', () => {
     const closeMock = vi.fn();
     const openMock = vi.fn();
 
-    const disposeFuncOnDataMock = vi.fn();
-    const disposeFuncOnErrorMock = vi.fn();
-    const disposeFuncOnEndMock = vi.fn();
+    const disposeOnDataMock = vi.fn();
+    const disposeOnErrorMock = vi.fn();
+    const disposeOnEndMock = vi.fn();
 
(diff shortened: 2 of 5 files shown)


axel7083 (commented):
I am not sure to understand what is the change goal in this PR :( 

> I am guessing some kind of close action, but what for ? why ?

gastoner:
> I am not sure to understand what is the change goal in this PR :(

The goal is to expose close function to frontend using `window.shellInProviderConnectionClose`

packages/main/src/plugin/provider-registry.ts
```diff
@@ -1274,21 +1274,36 @@ export class ProviderRegistry {
     onData: (data: string) => void,
     onError: (error: string) => void,
     onEnd: () => void,
-  ): Promise<{ write: (param: string) => void; resize: (dimensions: ProviderConnectionShellDimensions) => void }> {
+  ): Promise<{
+    write: (param: string) => void;
+    resize: (dimensions: ProviderConnectionShellDimensions) => void;
+    close: () => void;
+  }> {
```
axel7083, line 1281: With the number of object in it, this could be a dedicated interface

packages/main/src/plugin/provider-registry.ts
```diff
@@ -1298,6 +1313,12 @@ export class ProviderRegistry {
         resize: (dimension: ProviderConnectionShellDimensions): void => {
           connection?.resize(dimension);
         },
+        close: (): void => {
+          disposables.forEach(dispose => {
+            dispose.dispose();
+          });
+          connection?.close();
```
axel7083, line 1320: might be nice to have an unit test to connection is closed when we call close

packages/main/src/plugin/provider-registry.ts
```diff
@@ -1298,6 +1313,12 @@ export class ProviderRegistry {
         resize: (dimension: ProviderConnectionShellDimensions): void => {
           connection?.resize(dimension);
         },
+        close: (): void => {
+          disposables.forEach(dispose => {
+            dispose.dispose();
+          });
```
benoitf, line 1319: suggestion: linter should tell you to use for of loop

packages/main/src/plugin/index.ts
```diff
@@ -1214,6 +1215,16 @@ export class PluginSystem {
       },
     );
 
+    this.ipcHandle(
+      'provider-registry:shellInProviderConnectionClose',
+      async (_listener, onDataId: number): Promise<void> => {
+        const callback = providerRegistryShellInProviderConnectionSendCallback.get(onDataId);
+        if (callback) {
+          callback.close();
+        }
```
```suggestion
        callback?.close();
```
suggestion

benoitf (commented):
todo: add unit tests for the disposables and the close in provider registry

adding new behaviour = need to provide the tests going into these branches

gastoner:
@benoitf @axel7083 updated, I just can't make the mocks working as I want (I want to check that the dispose is called for every listener)

packages/main/src/plugin/provider-registry.spec.ts
```diff
@@ -1140,3 +1142,94 @@ describe('startProvider', () => {
...
+    // Check that all callbacks are set and session is created
+    expect(openMock).toBeCalled();
+    expect(onDataMock).toBeCalled();
+    expect(onErrorMock).toBeCalled();
+    expect(onEndMock).toBeCalled();
+
+    // Check that close is called and listeners are disposed
+    shellInProviderConnection.close();
+
+    expect(disposeFuncOnDataMock).toBeCalled();
```
axel7083: maybe promise issue, try with 
```ts
await vi.waitiFor(() => {
     expect(disposeFuncOnDataMock).toHaveBeenCalled();
});
```

gastoner: Still failing, can you check the mocks please? If they are ok ? 
E.g. `connection?.onData` have those params: (listener: (e: ProviderConnectionShellAccessData) => any, thisArgs?: any, disposables?: Disposable[]) => Disposable  and I'm adding the disposable to the `disposables`

benoitf (approved):
I'm fine to have the tests being worked on in a follow-up PR

benoitf:
suggestion: move out the test that is not working in another PR (so we don't forget), and remove the non-working test from this one

then we can iterate in parallel

packages/main/src/plugin/provider-registry.spec.ts
axel7083: Really sorry I was not able to find the time to look at it, as suggested by Florent tag me in a follow-up PR so we can iterate

axel7083 (approved):
LGTM

PR #9553, opened 2024-10-23
feat: collect Deployments events and dispatch them

### LINKED ISSUE #6210: Display Events attached to Kubernetes resource

When exploring a pod on Kubernetes (or any other resource), we don't have access to the events attached to the resource (the ones visible in the output of tht `kubectl describe pod` command) 

- [x] Deployments: #9553 #9657 
- [x] Pods: #9664 #9672
- [x] Nodes: #9681 #9691
- [x] Services: #9692 #9733 
- [ ] Pods events on Deployment page? 
- [ ] ...

A section in the resources details page ~~tab in the resource details, to display the events attached to the resource~~

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

collect Deployments events and dispatch them

### Screenshot / video of UI

No screenshot, backend only

Part of #6210 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-manager.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager.ts
@@ -18,6 +18,8 @@
 
 import type {
   Context,
+  CoreV1Event,
+  CoreV1EventList,
   Informer,
   KubernetesListObject,
   KubernetesObject,
@@ -254,6 +256,9 @@ export class ContextsManager {
         break;
       case 'secrets':
         informer = this.createSecretInformer(this.kubeConfig, ns, context);
+        break;
+      case 'events':
+        informer = this.createEventInformer(this.kubeConfig, ns, context);
         break;
       default:
         console.debug(`unable to watch ${resourceName} in context ${contextName}, as this resource is not supported`);
@@ -719,6 +724,59 @@ export class ContextsManager {
             (state.resources.routes = state.resources.routes.filter(
               d => d.metadata?.uid !== (obj.metadata as V1ObjectMeta)?.uid,
             )),
+        });
+      },
+    });
+  }
+
+  public createEventInformer(kc: KubeConfig, namespace: string, context: KubeContext): CancellableInformer {
+    const k8sApi = kc.makeApiClient(CoreV1Api);
+    const listFn = (): Promise<CoreV1EventList> => k8sApi.listNamespacedEvent({ namespace });
+    const path = `/api/v1/namespaces/${namespace}/events`;
+    let timer: NodeJS.Timeout | undefined;
+    let connectionDelay: NodeJS.Timeout | undefined;
+    this.setConnectionTimers('events', timer, connectionDelay);
+    return this.createInformer<CoreV1Event>(kc, context, path, listFn, {
+      resource: 'events',
+      timer: timer,
+      backoff: this.getBackoffForContext(context.name),
+      connectionDelay: connectionDelay,
+      onAdd: obj => {
+        this.states.setStateAndDispatch(context.name, {
+          currentContext: this.kubeConfig.currentContext,
+          resources: { events: true },
+          update: state => {
+            if (obj.involvedObject.kind === 'Deployment') {
+              state.resources.events.push(obj);
+            }
+          },
+        });
+      },
+      onUpdate: obj => {
+        this.states.setStateAndDispatch(context.name, {
+          currentContext: this.kubeConfig.currentContext,
+          resources: { events: true },
+          update: state => {
+            if (obj.involvedObject.kind === 'Deployment') {
+              state.resources.events = state.resources.events.filter(
+                o => o.metadata?.uid !== (obj.metadata as V1ObjectMeta)?.uid,
+              );
+              state.resources.events.push(obj);
+            }
+          },
+        });
+      },
+      onDelete: obj => {
+        this.states.setStateAndDispatch(context.name, {
+          currentContext: this.kubeConfig.currentContext,
+          resources: { events: true },
+          update: state => {
+            if (obj.involvedObject.kind === 'Deployment') {
+              state.resources.events = state.resources.events.filter(
+                d => d.metadata?.uid !== (obj.metadata as V1ObjectMeta)?.uid,
+              );
+            }
+          },
         });
       },
     });
--- a/packages/api/src/kubernetes-contexts-states.ts
+++ b/packages/api/src/kubernetes-contexts-states.ts
@@ -36,6 +36,7 @@ export const secondaryResources = [
   'secrets',
   'nodes',
   'persistentvolumeclaims',
+  'events',
 ] as const;
 
 export type SecondaryResourceName = (typeof secondaryResources)[number];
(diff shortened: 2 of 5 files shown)


benoitf (commented):
suggestion: as a general recommendation split the backend/preload part from the renderer part

packages/main/src/plugin/index.ts
```diff
@@ -2530,7 +2531,7 @@ export class PluginSystem {
 
     this.ipcHandle(
       'kubernetes-client:registerGetCurrentContextResources',
-      async (_listener, resourceName: ResourceName): Promise<KubernetesObject[]> => {
+      async (_listener, resourceName: ResourceName): Promise<KubernetesObject[] | CoreV1Event[]> => {
```
axel7083: Maybe have a dedicated ipc as events are really not the same as KubernetesObject

packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -870,7 +928,7 @@ export class ContextsManager {
     return this.states.getCurrentContextGeneralState(this.kubeConfig.currentContext);
   }
 
-  public registerGetCurrentContextResources(resourceName: ResourceName): KubernetesObject[] {
+  public registerGetCurrentContextResources(resourceName: ResourceName): KubernetesObject[] | CoreV1Event[] {
```
axel7083: same remark, maybe we should keep a difference between the KubernetesObject and the events

feloy: ok, I'll try to make a `registerGetCurrentContextResources<T>` on the backend side, and have 2 separate ipc calls (I don't think it is possible to hava a generic through IPC?)

axel7083: I do not know what is the right approach, as I am not familiar enough with the code, but do we really need to collect all events ? and store them all ?

Shouldn't we do something like logs, and only fetch them for a specific resources, when we need them ? I don't know, but since we do not display all kubernetes resources, if we have events we don't use, that's a bit sad to store them all

Some ideas from the top of my head:

We either have a dedicated ipc `registerGetCurrentContextEvents` and keeps the `registerGetCurrentContextResources` for KubernetesObject only, or we could do something like `getKubernetesObjectEvents(): Promise<CoreV1Event[]>` when we open the details page we subscribe, or do as the logs, but that would mean a dedicated informers, which is overkill I agree, so I don't know what the best solution here

feloy: The advantage of the current method is that the events list is refreshed as soon as new events happen, it can be very helpful when troubleshooting a deployment.

There are some filtering to store only Deployments related events:
```
 onAdd: obj => {
        this.states.setStateAndDispatch(context.name, {
          resources: { events: true },
          update: state => {
            if (obj.involvedObject.kind === 'Deployment') {
              state.resources.events.push(obj);
            }
          },
        });
      },
```

axel7083: We will want to extend this to other resources I guess ? That's a good solution to have this filtering

I agree with the advantages, but still think it should be a ipc on its own, as this would make more complicated to have two object as result each time :/

Firewall:
Thanks for sending over the link. This looks awesome!

A couple of suggestions/questions (based of the video): 
1. The padding after the `Events` title is a bit bigger than the rest, feels very out of place. 
2. Is there a limit on the amount of events shown? I worry this can get long quickly 
3. Should we reverse the order? Most recent events show on top? Again, if you have a large amount of events you would be scrolling a long time otherwise

feloy:
> Thanks for sending over the link. This looks awesome!
> 
> A couple of suggestions/questions (based of the video):
> 
> 1. The padding after the `Events` title is a bit bigger than the rest, feels very out of place.

Yes, I'll try to fix it.

> 2. Is there a limit on the amount of events shown? I worry this can get long quickly

There is no limit. But Kubernetes server removes the events after a few time (and will be deleted in Podman Desktop due to the synchronization). These are the same events which are visible at the end of the output of `kubectl describe deployment my-deploy`; the list is generally short IMO.

> 3. Should we reverse the order? Most recent events show on top? Again, if you have a large amount of events you would be scrolling a long time otherwise

I have to check, to display them in the same order as in `kubectl describe` I suppose


Thanks for your feedback

packages/main/src/plugin/kubernetes/contexts-manager.ts
feloy: @axel7083  I have updated the commit with changes, with the events handled separately from the other resources, and with a specific IPC for events. 

I'll have to double-check and test everything, but this is a first step to have the big picture.

Do you think it is ok this way?

benoitf (changes requested):
blocking: break down the PR in 2
1. (main/preload) 
2. renderer

feloy:
setting as draft for now for a few days, I'll continue by splitting it in several PRs, if the proposal is ok for everyone

packages/renderer/src/lib/events/EventUI.ts
```diff
@@ -0,0 +1,25 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
axel7083:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/renderer/src/lib/events/EventUI.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface EventUI {
```
axel7083: Suggestion (not blocking): I saw that you have been adding the uid to kube resources, might also be interesting to have it for the EventUi ? What do you think?

packages/main/src/plugin/index.ts
feloy: I realized that CoreV1Event is actually a superset of KubernetesObject, this is finally not necessary

packages/main/src/plugin/kubernetes/contexts-manager.ts
feloy: Finally, I changed back to consider events as a normal secondary resource, as the CoreV1Event is a superset of KubernetesObject

feloy:
> blocking: break down the PR in 2
> 
> 1. (main/preload)
> 2. renderer

I removed the frontend part, to keep main ~and preload~ only on this PR

axel7083:
> I removed the frontend part, to keep main and preload only on this PR

I do not see any new IPC for the preload ?

feloy:
> > I removed the frontend part, to keep main and preload only on this PR
> 
> I do not see any new IPC for the preload ?

Right, my comment was erroneous, the events are managed by the same IPCs as the other secondary resources.

axel7083 (approved):

benoitf (approved):

PR #9575, opened 2024-10-24
fix: dispose existing port forwards on delete

### LINKED ISSUE #9561: `KubernetesClient#deletePortForward` do not dispose the server
### Bug description

I was working on https://github.com/containers/podman-desktop/issues/9273 and https://github.com/containers/podman-desktop/issues/9275 and I found a problem with the way we _delete_ port foward added in https://github.com/containers/podman-desktop/pull/7379

When we delete a port forward we get the `KubernetesPortForwardService`

https://github.com/containers/podman-desktop/blob/2600ef0449b96d916cf282437d1320fbb5e1bbf1/packages/main/src/plugin/kubernetes/kubernetes-client.ts#L1772-L1774

The service itself ask the configuration to delete it

https://github.com/containers/podman-desktop/blob/2600ef0449b96d916cf282437d1320fbb5e1bbf1/packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts#L105-L108

And the configuration remove it from its configuration: **but do not dispose the server!!**

https://github.com/containers/podman-desktop/blob/2600ef0449b96d916cf282437d1320fbb5e1bbf1/packages/main/src/plugin/kubernetes/kubernetes-port-forward-storage.ts#L324-L332

## Why ?

When we start the port forward, we do not store the disposable, meaning, we are never able to remove it.

https://github.com/containers/podman-desktop/blob/2600ef0449b96d916cf282437d1320fbb5e1bbf1/packages/main/src/plugin/kubernetes/kubernetes-client.ts#L1768

The only places where the disposables are stored are in the `KubernetesPortForwardService`, but they are an unusable Array<IDisposable>

https://github.com/containers/podman-desktop/blob/ec3ee08dbe4ff0343f5cb12f70756d286c4f1e8a/packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts#L68

### Operating system

Fedora 40

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. start a port foward using code from https://github.com/containers/podman-desktop/pull/7379
2. call delete
3. assert server still running

### Relevant log output

_No response_

### Additional context

We would need to have somewhere, probably the `KubernetesPortForwardService` a Map linking a resources to the disposable, or something similar

jeffmaury:
Fixes #9561

### What does this PR do?

Dispose an existing port forwarding on delete

### Screenshot / video of UI

N/A

#9561 

### How to test this PR?

Run `pnpm test:main`

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
@@ -65,7 +65,7 @@ export class KubernetesPortForwardServiceProvider {
  * @see KubernetesPortForwardServiceProvider.getService
  */
 export class KubernetesPortForwardService implements IDisposable {
-  #disposables: IDisposable[] = [];
+  #disposables: Map<string, IDisposable> = new Map();
 
   /**
    * Creates an instance of KubernetesPortForwardService.
@@ -83,6 +83,10 @@ export class KubernetesPortForwardService implements IDisposable {
 
   dispose(): void {
     this.#disposables.forEach(disposable => disposable.dispose());
+  }
+
+  private getPortForwardKey(config: ForwardConfig): string {
+    return `${config.namespace}/${config.name}/${config.kind}`;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
```diff
@@ -79,6 +79,10 @@ export class KubernetesPortForwardService implements IDisposable {
     this.#disposables.forEach(disposable => disposable.dispose());
```
axel7083, line 79: not sure this works anymore, as the `disposables` is now a Map instead of an array, maybe we need something like `Array.from(this.#disposables.values()).forEach(disposable => disposable.dispose());`

axel7083 (approved):
LGTM

gastoner (approved):

PR #9594, opened 2024-10-24
fix: use `utf16le` ecoding in powershell call for Virtual Machine Platform detection

### LINKED ISSUE #7414: Podman Desktop unable to detect the Virtual Machine Platform already enabled
### Bug description

After uninstalling the older version, I am trying to set up Podman Desktop and Podman on my Windows Laptop. I keep getting the error below on the check requirements screen. 

`Virtual Machine Platform should be enabled to be able to run Podman.
`

### Operating system

Windows 11

### Installation Method

### Version

1.10.2

### Steps to reproduce

https://github.com/containers/podman-desktop/discussions/7413

### Relevant log output

```shell
https://github.com/containers/podman-desktop/discussions/7413
```

### Additional context

https://github.com/containers/podman-desktop/discussions/7413

dgolovin:
### What does this PR do?

Uses default encoding for powershell execution similar to https://github.com/containers/podman-desktop/pull/4918 to find string in powershell output.

Fix #7414.

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/podman-install.spec.ts
+++ b/extensions/podman/packages/extension/src/podman-install.spec.ts
@@ -298,6 +298,9 @@ test('expect winVirtualMachine preflight check return successful result if the v
 
   const winVirtualMachinePlatformCheck = new VirtualMachinePlatformCheck();
   const result = await winVirtualMachinePlatformCheck.execute();
+  expect(extensionApi.process.exec).toBeCalledWith(expect.anything(), expect.arrayContaining([expect.anything()]), {
+    encoding: 'utf16le',
+  });
   expect(result.successful).toBeTruthy();
 });
 
--- a/extensions/podman/packages/extension/src/podman-install.ts
+++ b/extensions/podman/packages/extension/src/podman-install.ts
@@ -647,9 +647,13 @@ export class VirtualMachinePlatformCheck extends BaseCheck {
   async execute(): Promise<extensionApi.CheckResult> {
     try {
       // set CurrentUICulture to force output in english
-      const { stdout: res } = await extensionApi.process.exec('powershell.exe', [
-        '(Get-WmiObject -Query "Select * from Win32_OptionalFeature where InstallState = \'1\'").Name | select-string VirtualMachinePlatform',
-      ]);
+      const { stdout: res } = await extensionApi.process.exec(
+        'powershell.exe',
+        [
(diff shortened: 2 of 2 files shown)


benoitf (approved):
we already had this issue in the past and fixed using 'utf16le' in podman

this PR works for me but I think we need a follow-up having a powerShell helper in podman extension where we always pass the encoding to not forget it

suggestion: we had a wslHelper, could we have a powershellHelper

todo: there is a missing unit test before merging

dgolovin:
Bad news, it does not work as expected. I am not sure about #4918 ether, because I am not sure it was tested on real OS and the test is not actually testing the case.

With powershell it is even worse, because we are running powershell and ask it to execute a script. Will try to replicate the problem first.

dgolovin:
Added command to set output encoding to match encoding we set on stdout.

Added test to verify encoding option is passed to the process.exec.

benoitf (approved):
approving for the code but needs to be checked on windows

cc @containers/podman-desktop-qe-reviewers

gastoner (approved):
Tried to create a machine on Windows ✅ , not sure what else should I try out

dgolovin:
> approving for the code but needs to be checked on windows
> 
> cc @containers/podman-desktop-qe-reviewers

@gastoner Could you try to go through onboarding for podman? That is when the code is going to be executed to check for VirtualMachinePlatform is installed.

gastoner:
@dgolovin 

https://github.com/user-attachments/assets/1247cb96-c25d-4f00-86ca-87c3c09f37f5

Should be OK (this step should contain the check for VirtualMachinePlatform  right?)

odockal:
> @dgolovin
> Screen.Recording.2024-11-13.151619.mp4
> 
> Should be OK (this step should contain the check for VirtualMachinePlatform right?)

Nope, it is when you do not have podman installed, then you can either go through onboarding ( on fresh start up of podman desktop or via Settings -> Resources -> Setup Podman to run the requirements check for podman installation) or via dashboard (Install Podman).

odockal (changes requested):
I cannot get through the WSL2 version check:
wsl-requirements-check

odockal:
But it might not be related to this PR at all. 1.14.1 behaves the same. I am doing a clean up of the PC.

odockal (approved):
Finally, I managed to actually get `wsl --install --no-distribution` applied. Hence everything works fine. 

Problem might have been that there was a hanging windows update. I updated and all works normally.

PR #9595, opened 2024-10-24
fix: next version check

### LINKED ISSUE #9521: Podman Desktop x.xx Released updater wrongly appearing
### UX Description

After installing Podman Desktop via Homebrew on macOS I've noticed that an update prompt appears wrongly when no other version actually appears to be visible:
image

Most recent Release available:
image

What pressing "Update" button shows:
image

If I try the update, which I shouldn't (need to do), it yields:
image

Obviously this should not be happening.

### Request type

UX analysis/suggestions for improvement

### Primary Contact

_No response_

### Deadline for request

_No response_

SoniaSandler:

### What does this PR do?
This PR changes the nextVersion value in updater to be set only when its is greater than the current version. It also updates the function `updateAvailable()` to return boolean based on if nextVersion is set or undefined.

Closes https://github.com/containers/podman-desktop/issues/9521

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/updater.spec.ts
+++ b/packages/main/src/plugin/updater.spec.ts
@@ -782,3 +782,26 @@ test('update is not available when next version is less than the current version
 
   expect(updater.updateAvailable()).toBeFalsy();
 });
+
+test('update is not available when next version empty', async () => {
+  vi.mocked(autoUpdater.checkForUpdates).mockResolvedValue({
+    updateInfo: {
+      version: '',
+    },
+  } as unknown as UpdateCheckResult);
+
+  vi.mocked(app.getVersion).mockReturnValue('1.5.1');
+  const updater = new Updater(
+    messageBoxMock,
+    configurationRegistryMock,
+    statusBarRegistryMock,
+    commandRegistryMock,
+    taskManagerMock,
+    apiSenderMock,
+  );
+  updater.init();
+
+  await vi.waitFor(() => expect(autoUpdater.checkForUpdates).toBeCalled());
+
+  expect(updater.updateAvailable()).toBeFalsy();
+});
--- a/packages/main/src/plugin/updater.spec.ts
+++ b/packages/main/src/plugin/updater.spec.ts
@@ -373,11 +373,9 @@ describe('expect update command to depends on context', async () => {
 
     vi.mocked(autoUpdater.checkForUpdates).mockResolvedValue({
       updateInfo: {
-        version: '1.2.0',
+        version: '@debug-next',
       },
     } as unknown as UpdateCheckResult);
-
-    vi.mocked(app.getVersion).mockReturnValue('1.1.0');
 
     let mListener: UpdateCommandListener | undefined;
     vi.mocked(commandRegistryMock.registerCommand).mockImplementation(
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/updater.spec.ts
```diff
@@ -373,10 +373,12 @@ describe('expect update command to depends on context', async () => {
 
     vi.mocked(autoUpdater.checkForUpdates).mockResolvedValue({
       updateInfo: {
-        version: '@debug-next',
```
axel7083: question: What is happening when we have non-semvers compatible as version ? Such as @debug-next what is the outcome ? False I guess ?

> might not be relevant, but curious if the string is malformed

packages/main/src/plugin/updater.ts
```diff
@@ -403,15 +404,18 @@ export class Updater {
       .checkForUpdates()
       .then(result => {
         this.#updateCheckResult = result ?? undefined;
-        this.#nextVersion = this.getFormattedVersion(this.#updateCheckResult?.updateInfo);
+        this.#nextVersion =
+          compare(app.getVersion(), this.#updateCheckResult?.updateInfo.version ?? '') === -1
+            ? this.getFormattedVersion(this.#updateCheckResult?.updateInfo)
+            : undefined;
```
axel7083: Suggestion (not blocking): A bit compact, might be a bit hard to understand, might be nice to use if else format to ease the reading

axel7083 (approved):
LGTM :rocket:

axel7083:
No issue linked ?

gastoner (approved):
LGTM

benoitf (approved):
todo before merging: I would keep a text with a non semantic version (like the current dummy version in one test) to see if it's not breaking in that case.

question: what should we do in case it's non semantic, do we upgrade because we found something anyway or we say it's not an update ? I would opt for 'upgrade'

SoniaSandler:
@benoitf I updated the PR so that if one of the versions is not a valid semver, it will still update the `nextVersion` value.
In case the `this.#updateCheckResult?.updateInfo.version` is not defined or an empty string it will still update `nextVersion` (since it is not a valid semver) but `!!nextVersion` will be false.

benoitf:
> this.#updateCheckResult?.updateInfo.version is not defined or an empty string it will still update  but !!nextVersion will be false.

~~why `#nextVersion` would be false ? `#nextVersion` should be `updaterInfo.version` so even if the version name is `foobar` it should set the version to `foobar` and then updateAvailable() would return true~~

yes it's fine, undefined or empty means no version is there to be updated

benoitf (approved):

packages/main/src/plugin/updater.spec.ts
```diff
@@ -712,3 +712,117 @@ test('get release notes in dev mode', async () => {
...
+
+  await vi.waitFor(() => expect(autoUpdater.checkForUpdates).toBeCalled());
+
+  expect(updater.updateAvailable()).toBeFalsy();
+});
+
+test('update version is not full semver', async () => {
+  vi.mocked(autoUpdater.checkForUpdates).mockResolvedValue({
+    updateInfo: {
+      version: '1.5',
```
benoitf, line 810: question: you didn't choose like a 'non-version' pattern like 'foo' or 'anything', is semver failing in the same way than a partial version like `1.5`

SoniaSandler: 1.5 is not a valid semver based on the valid check, but I can add another test

packages/main/src/plugin/updater.spec.ts
SoniaSandler: @axel7083 following Florent's comment and based on the most recent commit, if either one of the versions (current or next) is not a valid semver, it will still update `nextVersion` value

axel7083 (approved):

PR #9596, opened 2024-10-24
fix: check if newly installed extensions dependencies are already installed

### LINKED ISSUE #9069: Cannot install an extension with dependency on another
### Bug description

I tried to install OpenShift Local extension and I got the error message:

`Error: Missing dependencies for this extension: redhat.redhat-authentication`

But the extension is installed

### Operating system

Win11Pro

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
This PR fixes the dependency check for extensions that were installed after starting PD.

Closes https://github.com/containers/podman-desktop/issues/9069

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -430,7 +430,6 @@ export class ExtensionLoader {
     }
 
     // now we have all extensions, we can load them
-    console.log(analyzedExtensions.length);
     await this.loadExtensions(analyzedExtensions, true);
   }
 
@@ -501,13 +500,9 @@ export class ExtensionLoader {
       extension.manifest?.extensionDependencies?.forEach(async (dependency: string) => {
         if (!existingExtensions.find(existingExtension => existingExtension.id === dependency)) {
           missingDependencies.push(dependency);
-          console.log('extension: ' + extension + ' dependency: ' + dependency + ' missing');
         }
-        console.log(missingDependencies.length);
       });
       extension.missingDependencies = missingDependencies;
-
-      console.log(`extension length: ` + extension.missingDependencies.length);
     });
   }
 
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/extension-loader.ts
```diff
@@ -493,21 +493,24 @@ export class ExtensionLoader {
...
+        if (!existingExtensions.find(existingExtension => existingExtension.id === dependency)) {
           missingDependencies.push(dependency);
         }
       });
       extension.missingDependencies = missingDependencies;
     });
   }
 
-  async loadExtensions(analyzedExtensions: AnalyzedExtension[]): Promise<void> {
+  async loadExtensions(analyzedExtensions: AnalyzedExtension[], onStart: boolean): Promise<void> {
```
benoitf: issue(blocking): no parameter needs to be added there

all consumers ask to load extensions, no 'onStart' (or another name paramerer)

packages/main/src/plugin/extension-loader.ts
```diff
@@ -494,10 +494,11 @@ export class ExtensionLoader {
   // check if all dependencies are available
   // if not, set the missingDependencies property
   searchForMissingDependencies(analyzedExtensions: AnalyzedExtension[]): void {
+    const existingExtensions = [...new Set([...this.analyzedExtensions.values(), ...analyzedExtensions])];
```
```suggestion
    const existingExtensions = Array.from(new Set([...this.analyzedExtensions.values(), ...analyzedExtensions]));
```

benoitf: suggestion: use the Array.from syntax

benoitf (approved):
thanks for the changes

packages/main/src/plugin/extension-loader.ts
SoniaSandler: done

axel7083 (approved):

PR #9602, opened 2024-10-25
chore: properties can be part of a group

benoitf:
### What does this PR do?
for some case, we need to be able to say two properties are part of a given group. Add the optional property so it's possible to group properties

usecase: group compose & podman socket helper property to a common block in the UI

part of https://github.com/containers/podman-desktop/issues/9242

### How to test this PR?

unit test provided


axel7083 (approved):
LGTM :+1:

gastoner (approved):

Issue #9242, opened 2024-10-07
continue the implementation of Docker Compatibility page


benoitf:
### Is your enhancement related to a problem? Please describe

There is an experimental flag to bring this new section

It needs to be completed with the compose part and the podman part (from the mockup it's the sections matching the vertical podman bar in purple)

### Describe the solution you'd like

continue the implementation

PR #9612, opened 2024-10-25
chore: prefer method of the extension over system to get OS

benoitf:
### What does this PR do?
change the if condition (easier to mock)

related to #9242 

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -1378,7 +1378,7 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
   // Compatibility mode status bar item
   // only available for macOS
-  if (isMac()) {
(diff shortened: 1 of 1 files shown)


deboer-tim (commented):
question: The same function still has a call to isMac(), why one and not the other? This file already has a mix of both too, I'm worried it will be confusing for future maintainer to understand why the mix.

benoitf:
> question: The same function still has a call to isMac(), why one and not the other? 

sure, it does not go in some sections being tested

The mix is there for legacy reason, it's taking time to rollout the usage of some new API.
some repositories have created issues for that https://github.com/containers/podman-desktop-extension-bootc/issues/881

The reason here is to get an easier mocking

benoitf:
filed https://github.com/containers/podman-desktop/issues/9617

deboer-tim (approved):

PR #9619, opened 2024-10-25
chore: register methods allowing to call enable/disable of the compat mode

benoitf:
### What does this PR do?

There are 2 commits but one is part of another PR: (not yet reviewed)
- [x] depends on https://github.com/containers/podman-desktop/pull/9611
need to rebase before merging

It allows to call with methods the enable or disable methods to enable or disable the compatibility mode on macOS

related to #9242 

### How to test this PR?

unit test provided

### DIFF
--- a/extensions/podman/packages/extension/src/extension.spec.ts
+++ b/extensions/podman/packages/extension/src/extension.spec.ts
@@ -2588,6 +2588,87 @@ async function testAudit(path: string, uri: string, condition: typeof expect | t
   expect(auditRecords.records).toEqual(condition.arrayContaining([expect.objectContaining({ type: 'error' })]));
 }
 
+test('activate on mac register commands for setting compatibility moide ', async () => {
+  vi.mocked(isMac).mockReturnValue(true);
+  vi.mocked(isWindows).mockReturnValue(false);
+  vi.mocked(isLinux).mockReturnValue(false);
+  vi.spyOn(PodmanInstall.prototype, 'checkForUpdate').mockResolvedValue({
+    hasUpdate: false,
+  } as unknown as UpdateCheck);
+  const contextMock = {
+    subscriptions: [],
+    secrets: {
+      delete: vi.fn(),
+      get: vi.fn(),
+      onDidChange: vi.fn(),
+      store: vi.fn(),
+    },
+  } as unknown as extensionApi.ExtensionContext;
+
+  // mock getSocketCompatibility
+  const disableMock = vi.fn();
+  const enableMock = vi.fn();
+  const isEnabledMock = vi.fn().mockReturnValue(false);
+  const mock = vi.spyOn(compatibilityModeLib, 'getSocketCompatibility');
+  mock.mockReturnValue({
+    isEnabled: isEnabledMock,
+    enable: enableMock,
+    disable: disableMock,
+    details: '',
+    tooltipText: (): string => {
+      return '';
+    },
+  });
+
+  const api = await extension.activate(contextMock);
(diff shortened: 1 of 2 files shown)


benoitf:
rebased

extensions/podman/packages/extension/src/extension.spec.ts
```diff
@@ -2588,6 +2588,87 @@ async function testAudit(path: string, uri: string, condition: typeof expect | t
   expect(auditRecords.records).toEqual(condition.arrayContaining([expect.objectContaining({ type: 'error' })]));
 }
 
+test('activate on mac register commands for setting compatibility moide ', async () => {
+  vi.mocked(isMac).mockReturnValue(true);
+  vi.mocked(isWindows).mockReturnValue(false);
+  vi.mocked(isLinux).mockReturnValue(false);
+  vi.spyOn(PodmanInstall.prototype, 'checkForUpdate').mockResolvedValue({
+    hasUpdate: false,
+  } as unknown as UpdateCheck);
```
axel7083, line 2597: question: could this have side effect on the other tests ?

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -1390,8 +1391,23 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
       }
     });
 
+    // register two commands to enable and disable compatibility mode
+    extensionContext.subscriptions.push(
+      extensionApi.commands.registerCommand('podman.disableCompatibilityMode', async () => {
```
axel7083, line 1396: question: should we had the command to the package.json ? other podman commands are listed in it.

axel7083 (approved):
LGTM code-wise. I do not see the commands added on windows (command palette) as the changes only target macOS. 

remark (not blocking): I am able to see the `podman.socketCompatibilityMode` command in the command palette even tho, it has not been registered and do nothing when I use it🤔 

image

extensions/podman/packages/extension/src/extension.spec.ts
benoitf: not more than the other https://github.com/containers/podman-desktop/blob/a230b9d8c2c3102df3df01955f4ac5134a860ca9/extensions/podman/packages/extension/src/extension.spec.ts#L2532

but anyway there is
https://github.com/containers/podman-desktop/blob/a230b9d8c2c3102df3df01955f4ac5134a860ca9/extensions/podman/packages/extension/src/extension.spec.ts#L176

so having all mocks being reset it should not impact any other test

extensions/podman/packages/extension/src/extension.ts
benoitf: we don't need as we don't provide any shorcuts/nice display name etc, being mostly internal

benoitf:
> remark (not blocking): I am able to see the podman.socketCompatibilityMode command in the command palette even tho, it has not been registered and do nothing when I use it🤔

it's registered there
https://github.com/containers/podman-desktop/blob/c8e8e4381afb489a6af2a1e5c09b05914dbd55df/extensions/podman/packages/extension/src/extension.ts#L1410

My PR is adding `podman.enableCompatibilityMode` and `podman.disableCompatibilityMode` (not `podman.socketCompatibilityMode`)


This part will be removed when all the settings will be done using the docker compliance page (but I can't already remove the code as the page is only at experimental stage)

cdrage (approved):
Pulled the PR and tested and was able to still enable / disable it fine. 

Going through the code it LGTM (more tests than actual code which is awesome!).

LGTM

PR #9627, opened 2024-10-26
fix: proxy settings do not take effect when switched

### LINKED ISSUE #9625: Proxy settings do not take effect when switched to system or disabled
### Bug description

Proxy settings do not take effect when switched to system or disabled. After restart, only the first setting is valid.

https://github.com/user-attachments/assets/2d74256e-15ec-4ee5-a41d-e8262f4492eb

### Operating system

Windows 11

### Installation Method

### Version

1.13.3

### Steps to reproduce

1. Set system proxy setting to `127.0.0.1:7890`
2. Select `System` in Podman Desktop -> Setting -> Proxy
3. Restart podman desktop
4. Push `Update` in Podman Desktop -> Setting -> Proxy. This will set env `http_proxy=http:127.0.0.1:7890` in the config file.
5. Select Proxy configuration to `Manual`. Change Web Proxy (HTTP) to `http:127.0.0.1:7891` and push `Update`. This will set env `http_proxy=http:127.0.0.1:7891` in the config file.
6. Select Proxy configuration to `System` and push `Update`. This did not change the config file.
7. Select Proxy configuration to `Disabled` and push `Update`. This did not change the config file.

### Relevant log output

_No response_

### Additional context

_No response_

Indekkusu545:
### What does this PR do?

Fixes proxy settings do not take effect when switched to system or disabled

Fixes #9625 

### How to test this PR?

- [ x] Tests are covering the bug fix or the new feature

### DIFF
--- a/packages/main/src/plugin/proxy.ts
+++ b/packages/main/src/plugin/proxy.ts
@@ -154,6 +154,8 @@ export class Proxy {
       newProxy.httpsProxy = ensureURL(newProxy.httpsProxy);
     } else if (this.proxyState === ProxyState.PROXY_SYSTEM) {
       newProxy = await getProxySettingsFromSystem(this);
+    } else if (this.proxyState === ProxyState.PROXY_DISABLED) {
+      newProxy = {} as ProxySettings;
     }
 
     // update configuration
--- a/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
@@ -25,9 +25,7 @@ onMount(async () => {
 
 async function updateProxySettings() {
   await window.setProxyState(proxyState);
-  if (proxyState !== ProxyState.PROXY_SYSTEM) {
-    await window.updateProxySettings({ httpProxy, httpsProxy, noProxy });
-  }
+  await window.updateProxySettings({ httpProxy, httpsProxy, noProxy });
 
   // loop over all providers and container connections to see if there are any running engines
   const providerInfos = await window.getProviderInfos();


packages/renderer/src/lib/preferences/PreferencesProxiesRendering.svelte
```diff
@@ -25,9 +25,7 @@ onMount(async () => {
 
 async function updateProxySettings() {
   await window.setProxyState(proxyState);
-  if (proxyState !== ProxyState.PROXY_SYSTEM) {
```
Indekkusu545, line 28: Note that this condition was added by @jeffmaury a while ago to fix #9379.

This condition causes the proxy configuration not to be updated when the proxy configuration is switched to System. When I deleted this condition, the test did not find any errors.

I am not sure why this condition was added, maybe further discussion is needed.

axel7083 (approved):
Thanks @Indekkusu545 for the fix, the change LGTM, waiting for @jeffmaury confirmation

packages/main/src/plugin/proxy.ts
```diff
@@ -154,6 +154,8 @@ export class Proxy {
       newProxy.httpsProxy = ensureURL(newProxy.httpsProxy);
     } else if (this.proxyState === ProxyState.PROXY_SYSTEM) {
       newProxy = await getProxySettingsFromSystem(this);
+    } else if (this.proxyState === ProxyState.PROXY_DISABLED) {
+      newProxy = {} as ProxySettings;
```
jeffmaury, line 158: Should be undefined as we are preventing use of empty proxy setting

jeffmaury (changes requested):

benoitf:
outdated

PR #9642, opened 2024-10-28
feat(ui): adding kubernetesCurrentContextPortForwards store

axel7083:
### What does this PR do?

We need a store for the port forward config, to be able to list them (https://github.com/containers/podman-desktop/issues/9275) and properly show the actions such as delete, open in browser (https://github.com/containers/podman-desktop/issues/9273)

### Screenshot / video of UI

N/A

### How to test this PR?

### DIFF
--- a/packages/renderer/src/stores/kubernetes-contexts-state.ts
+++ b/packages/renderer/src/stores/kubernetes-contexts-state.ts
@@ -20,6 +20,7 @@ import type { KubernetesObject } from '@kubernetes/client-node';
 import { derived, type Readable, readable, writable } from 'svelte/store';
 
 import type { CheckingState, ContextGeneralState } from '/@api/kubernetes-contexts-states';
+import type { UserForwardConfig } from '/@api/kubernetes-port-forward-model';
 
 import { findMatchInLeaves } from './search-util';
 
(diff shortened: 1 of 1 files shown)


packages/renderer/src/stores/kubernetes-contexts-state.ts
```diff
@@ -246,3 +247,15 @@ export const kubernetesCurrentContextRoutesFiltered = derived(
...
 );
+
+// Port Forwarding
+
+export const kubernetesCurrentContextPortForwards = readable<UserForwardConfig[]>([], set => {
+  window.getKubernetesPortForwards().then(value => {
+    set(value);
+  });
+  window.events?.receive('kubernetes-port-forwards-update', (value: unknown) => {
+    set((value ?? []) as UserForwardConfig[]);
```
deboer-tim: question: based on where the event is fired it looks like it is always a value and we don't need the `?? []` ? We don't do it in the only other place we have this pattern in kubernetesContextsState.

packages/renderer/src/stores/kubernetes-contexts-state.ts
```diff
@@ -246,3 +247,15 @@ export const kubernetesCurrentContextRoutesFiltered = derived(
...
+  });
+  return () => {};
```
deboer-tim: question: do we need this line? (again, we don't do it in the other case)

deboer-tim (approved):
LGTM, added some comments but ok to merge.

packages/renderer/src/stores/kubernetes-contexts-state.ts
axel7083: Yes, you are right, it was the case at some point, but was fixed, removed!

packages/renderer/src/stores/kubernetes-contexts-state.ts
axel7083: Yes, removing !

cdrage (approved):

PR #9654, opened 2024-10-29
fix: use process close event to resolve promise returned form process.exec instead of exit event

dgolovin:
Docs for nodejs process on('exit') says stdout/stderr can still be open even after process exit and that can lead to loss of output text

### What does this PR do?

Potentially solves #7414, because output delay happen on windows pretty often

related to #7414.

### How to test this PR?


jeffmaury (approved):
LGTM

PR #9657, opened 2024-10-29
feat: display Deployments events

feloy:

### What does this PR do?

Display the Events related to a Deployment in the Summary page of the Deployment

### Screenshot / video of UI

deploy-events

Part of #6210 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/events/EventUI.ts
+++ b/packages/renderer/src/lib/events/EventUI.ts
@@ -16,13 +16,8 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import type { V1ObjectMeta } from '@kubernetes/client-node';
-
 export interface EventUI {
   count?: number;
-  apiVersion?: string;
-  kind?: string;
-  metadata?: V1ObjectMeta;
   type?: string;
   reason?: string;
   firstTimestamp?: Date;
--- a/packages/renderer/src/lib/events/EventUI.ts
+++ b/packages/renderer/src/lib/events/EventUI.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,20 +16,17 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-export interface DeploymentUI {
-  uid: string;
-  name: string;
-  status: string;
-  namespace: string;
-  replicas: number;
-  ready: number;
-  created?: Date;
-  selected: boolean;
-  conditions: DeploymentCondition[];
-}
+import type { V1ObjectMeta } from '@kubernetes/client-node';
 
-export interface DeploymentCondition {
-  type: string;
+export interface EventUI {
+  count?: number;
+  apiVersion?: string;
+  kind?: string;
+  metadata?: V1ObjectMeta;
+  type?: string;
   reason?: string;
+  firstTimestamp?: Date;
+  lastTimestamp?: Date;
+  reportingComponent?: string;
   message?: string;
 }
--- a/packages/renderer/src/lib/kube/details/EventsTable.spec.ts
+++ b/packages/renderer/src/lib/kube/details/EventsTable.spec.ts
@@ -1,0 +1,83 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+
+import { render, screen, within } from '@testing-library/svelte';
+import { afterEach, beforeEach, expect, test, vi } from 'vitest';
+
+import type { EventUI } from '../../events/EventUI';
+import EventsTable from './EventsTable.svelte';
+
+beforeEach(() => {
+  vi.useFakeTimers();
+});
+
+afterEach(() => {
+  vi.useRealTimers();
+});
+
+test('expect the events are displayed', async () => {
+  const events: EventUI[] = [
+    {
+      count: 1,
+      type: 'Normal',
+      reason: 'FirstReason',
+      reportingComponent: 'a-controller',
+      message: 'a first message',
+      firstTimestamp: new Date(),
+      lastTimestamp: new Date(),
+    },
+    {
+      count: 2,
+      type: 'Warning',
+      reason: 'SecondReason',
+      reportingComponent: 'another-controller',
+      message: 'a second message',
+      firstTimestamp: new Date(),
+    },
+  ];
+
+  vi.advanceTimersByTime(5_000);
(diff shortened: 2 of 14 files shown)


feloy:
@Firewall replying to https://github.com/containers/podman-desktop/pull/9553#issuecomment-2432543298

1. the issue should be fixed
2. Events are grouped together (with a "5x ..." for example), the list should be always short
3. I forced the order to be the oldest firsts, as in `kubectl describe`

packages/renderer/src/lib/events/EventUI.ts
```diff
@@ -0,0 +1,32 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { V1ObjectMeta } from '@kubernetes/client-node';
+
+export interface EventUI {
+  count?: number;
+  apiVersion?: string;
+  kind?: string;
+  metadata?: V1ObjectMeta;
```
axel7083: Some field feel unnecessary to me, why do  we need metadata property ? why don't we have `namespace` and `name` just like DeploymentUI ?

packages/renderer/src/lib/kube/details/EventsTable.svelte
```diff
@@ -0,0 +1,39 @@
...
+    <table class="w-full ml-2.5">
+      <tbody>
+        <tr>
+          <th align="left">Type</th><th align="left">Reason</th><th align="left">Age</th><th align="left">From</th><th align="left">Message</th>
+        </tr>
+        {#each events as event}
+          <tr>
+            <td>{event.type}</td>
+            <td>{event.reason}</td>
+            <td>{getAgeAndCount(event)}</td>
```
axel7083, line 37: We should be using `TableDurationColumn` here, as the age will not be refresh otherwise

packages/renderer/src/lib/events/EventUI.ts
feloy: right, I first added them to make the structure a superset of KubernetesObject, but it is not necessary. I'm removing them

packages/renderer/src/lib/kube/details/EventsTable.svelte
feloy: The `TableDurationColumn` component is specific to the `ui-svelte`'s `Table` component, but here a simple HTML `table` is used. Also, this component is able to display a simple `Date` only, here we want to display something more elaborate (`19 seconds (2x over 25 seconds)` for example)

I agree that refreshing the date would be nice, but I would prefer to do it in a follow-up PR, as IMO it in non blocking (the events themselves are refreshed, so the user will still see the new age when the events are modified)

axel7083: yes np thanks for investigating

packages/renderer/src/lib/events/EventUI.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+//import type { V1ObjectMeta } from '@kubernetes/client-node';
```
axel7083: to remove ?

packages/renderer/src/lib/events/EventUI.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface EventUI {
+  count?: number;
```
axel7083, line 20: we should probably have the uid ?

packages/renderer/src/lib/kube/details/EventsTable.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
+import humanizeDuration from 'humanize-duration';
+import moment from 'moment';
+
+import type { EventUI } from '../../events/EventUI';
+
+export let events: EventUI[];
+
+$: events.sort((ev1, ev2) => ((ev1.lastTimestamp ?? Date.now()) < (ev2.lastTimestamp ?? Date.now()) ? -1 : 1));
```
axel7083: suggestion: using svelte 5 ?

packages/renderer/src/lib/events/EventUI.ts
feloy: I don't think so. UID has been added to the Deployment because they can be referenced by Events using this UID. But here, an event won't be referenced by another event.

axel7083: okey 👍

packages/renderer/src/lib/kube/details/EventsTable.spec.ts
```diff
@@ -80,4 +80,28 @@ test('expect the events are displayed', async () => {
...
+    count: 1,
+    type: 'Normal',
+    reason: 'ThirdReason',
+    reportingComponent: 'a-controller',
+    message: 'a third message',
+    firstTimestamp: new Date(),
+    lastTimestamp: new Date(),
+  });
+  vi.advanceTimersByTime(10_000);
+  await rendered.rerender({ events: events });
```
axel7083, line 95: why re-render ? wouldn't it better to make two different tests ?

axel7083: The test name is `expect the events are displayed` so I am not sure to understand the need to re-render it

feloy: The purpose is to test that the events are refreshed when the entry is modified (I mean that the `$effect` works as expected)

feloy: (renaming the test to illustrate that the display must be reactive)

packages/renderer/src/lib/kube/details/EventsTable.svelte
```diff
@@ -0,0 +1,46 @@
...
+
+interface Props {
+  events: EventUI[];
+}
+
+let { events }: Props = $props();
+
+$effect(() => {
+  events.sort((ev1, ev2) => ((ev1.lastTimestamp ?? Date.now()) < (ev2.lastTimestamp ?? Date.now()) ? -1 : 1));
+});
```
axel7083: Maybe we should not sort, but create a dedicated state derived from the props

```ts
let sortedEvents: EventUI[] = $derived(events.toSorted((ev1, ev2) => ((ev1.lastTimestamp ?? Date.now()) < (ev2.lastTimestamp ?? Date.now()) ? -1 : 1)));
```

So we don't have to modify the props

packages/renderer/src/lib/kube/details/EventsTable.spec.ts
axel7083: Thanks

axel7083 (approved):
LGTM 🚀 , thanks for the changes, great work 👍

cdrage (approved):
Going through the code I have no issues with this, testing this works wonderfully:

Screenshot 2024-10-29 at 11 49 55 AM

cdrage (changes requested):
Actually, the spacing does not match "Conditions" for Node status which uses another similar table.

Should the spacing either change for Conditions, or should this change for Events?

EDIT: Maybe we should open an issue for Conditions to decrease spacing?

Screenshot 2024-10-29 at 11 51 02 AM

feloy:
> Actually, the spacing does not match "Conditions" for Node status which uses another similar table.
> 
> Should the spacing either change for Conditions, or should this change for Events?
> 
> EDIT: Maybe we should open an issue for Conditions to decrease spacing?
> 
> Screenshot 2024-10-29 at 11 51 02 AM

It should be fixed by rebasing, conditions have been fixed with https://github.com/containers/podman-desktop/pull/9658

cdrage (approved):

cdrage:
> > Actually, the spacing does not match "Conditions" for Node status which uses another similar table.
> > Should the spacing either change for Conditions, or should this change for Events?
> > EDIT: Maybe we should open an issue for Conditions to decrease spacing?
> > Screenshot 2024-10-29 at 11 51 02 AM
> 
> It should be fixed by rebasing, conditions have been fixed with #9658

Thanks! I see it's fixed now with the rebase.

PR #9664, opened 2024-10-29
feat: collect Pods events

feloy:

### What does this PR do?

Collect Pods events (backend only)

### Screenshot / video of UI

No screenshot, backend only

Part of #6210

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-manager.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager.ts
@@ -76,6 +76,8 @@ import { ResourceWatchersRegistry } from './resource-watchers-registry.js';
 // only the connectivity to the current context will be checked
 const MAX_NON_CURRENT_CONTEXTS_TO_CHECK = 10;
 
+const kindsWithEvents = ['Deployment', 'Pod'];
+
 interface CreateInformerOptions<T> {
   // resource name, for logging
   resource: string;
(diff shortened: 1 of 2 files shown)


gastoner (approved):
LGTM codewise

packages/main/src/plugin/kubernetes/contexts-manager.ts
```diff
@@ -757,7 +759,7 @@ export class ContextsManager {
           currentContext: this.kubeConfig.currentContext,
           resources: { events: true },
           update: state => {
-            if (obj.involvedObject.kind === 'Deployment') {
+            if (!!obj.involvedObject.kind && kindsWithEvents.includes(obj.involvedObject.kind)) {
```
benoitf, line 762:
```ts
if (kindsWithEvents.includes(obj.involvedObject.kind))
```
 is enough, do not need the `!!obj.involvedObject.kind`

feloy: The linter gives me this error:

```
Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.ts(2345)
```

benoitf: thanks @feloy maybe it's because the array is of type string[] and not Array<string | undefined>[]

PR #9672, opened 2024-10-30
feat: display pods events

feloy:

### What does this PR do?

Display pods events

### Screenshot / video of UI

pods-events

Part of #6210 

### How to test this PR?

Start a deployment with an unknown image, events should appear in the related Pod summary.

### DIFF
--- a/packages/renderer/src/lib/kube/KubePodDetailsSummary.svelte
+++ b/packages/renderer/src/lib/kube/KubePodDetailsSummary.svelte
@@ -1,13 +1,18 @@
 <script lang="ts">
 import type { V1Pod } from '@kubernetes/client-node';
 
+import Cell from '/@/lib/details/DetailsCell.svelte';
 import Table from '/@/lib/details/DetailsTable.svelte';
+import Title from '/@/lib/details/DetailsTitle.svelte';
 
+import type { EventUI } from '../events/EventUI';
+import EventsTable from './details/EventsTable.svelte';
 import KubeObjectMetaArtifact from './details/KubeObjectMetaArtifact.svelte';
 import KubePodSpecArtifact from './details/KubePodSpecArtifact.svelte';
 import KubePodStatusArtifact from './details/KubePodStatusArtifact.svelte';
 
 export let pod: V1Pod | undefined;
+export let events: EventUI[];
 </script>
 
 <Table>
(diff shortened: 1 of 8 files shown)


packages/renderer/src/lib/kube/KubePodDetailsSummary.svelte
```diff
@@ -1,21 +1,34 @@
...
     <KubePodSpecArtifact artifact={pod.spec} />
-  {:else}
+    <tr>
+      <Title>Events</Title>
+    </tr>
+    {#if events?.length}
+      <EventsTable events={events} />
+    {:else}
+      <tr><Cell>No events</Cell></tr>
+    {/if}
```
axel7083: Can we make a re-usage component, such as `KubeEventArtifact`, that we can share between Pods, Deployments, services etc.. (all kube object) ?

packages/renderer/src/lib/pod/PodDetailsSummary.svelte
```diff
@@ -18,6 +22,7 @@ onMount(async () => {
       const getKubePod = await window.kubernetesReadNamespacedPod(pod.name, ns);
       if (getKubePod) {
         kubePod = getKubePod;
+        events = $kubernetesCurrentContextEvents.filter(ev => ev.involvedObject.uid === getKubePod.metadata?.uid);
```
axel7083: Is this reactive ? I feel like this is not a derived state, and will only be called on the mount ?

feloy: changed to svelte5 and made it reactive

packages/renderer/src/lib/kube/KubePodDetailsSummary.svelte
feloy: added a reusable KubeEventArtifact component

packages/renderer/src/lib/pod/PodDetailsSummary.svelte
```diff
@@ -2,13 +2,20 @@
...
 
-export let pod: PodInfoUI;
-let kubePod: V1Pod | undefined;
-let getKubePodError: string;
+interface Props {
+  pod: PodInfoUI;
+}
+
+let { pod }: Props = $props();
+let kubePod: V1Pod | undefined = $state();
```
benoitf, line 16: question: is it a better solution to use/follow $state() to initialize a value to undefined ? Rather than $state(undefined)

benoitf (approved):

PR #9678, opened 2024-10-30
chore(test): remove an assertion that blocks the check for an existing kind cluster

### LINKED ISSUE #9677: Resource cleanup failure in test scenario

In the deploy-to-kubernetes.spec.ts test scenario, we have an afterAll hook that deletes all resources used during the test, including the Kind Cluster. The problem arises when some resources are not deleted properly, blocking the deletion of the Kind Cluster. This results in failures in other test scenarios due to the cluster still being present.

For example, see the workflow run: https://github.com/containers/podman-desktop-e2e/actions/runs/11585875336

An assertion that blocks the check for an existing kind cluster should be removed

_No response_

### Additional context

_No response_

amisskii:

### What does this PR do?

Ensures that the Kind Cluster will always be deleted in the deploy-to-kubernetes test scenario. Move the assertion of cluster presence.

https://github.com/containers/podman-desktop/issues/9677

### How to test this PR?

### DIFF
--- a/tests/playwright/src/utility/operations.ts
+++ b/tests/playwright/src/utility/operations.ts
@@ -332,7 +332,6 @@ export async function createKindCluster(
     const resourcesPage = await settingsPage.openTabPage(ResourcesPage);
     await playExpect(resourcesPage.heading).toBeVisible({ timeout: 10_000 });
     await playExpect.poll(async () => resourcesPage.resourceCardIsVisible('kind')).toBeTruthy();
-    await playExpect(kindResourceCard.markdownContent).toBeVisible();
     await playExpect(kindResourceCard.createButton).toBeVisible();
 
     if (await kindResourceCard.doesResourceElementExist()) {


cbr7 (changes requested):
I don't understand what behavior either of these changes is supposed to help with, why would deleting the kind cluster first be better than container and image?

I also don't understand why moving the validation for the createButton lower does anything?

amisskii:
> I don't understand what behavior either of these changes is supposed to help with, why would deleting the kind cluster first be better than container and image?
> 
> I also don't understand why moving the validation for the createButton lower does anything?

The first thing we should do, I think, is to move the deleteKindCluster method to ensure that the kind cluster is deleted as a priority. As mentioned in the issue, when a resource deletion fails, like for a container or image, it blocks the deletion of a kind cluster, leading to failures in other test scenarios where kind is used. We could also consider fixing the deleteContainer method, which failed in the linked workflow run, though it's unclear what caused this failure

My intention is to move the markdownContent assertion, not the creation button assertion. This is because, if at least one kind cluster exists, the markdownContent is not visible, causing the tests to stop at this assertion, and the condition that skips cluster creation never occurs. Ideally, after validating that a kind card exists, it should check whether there is a cluster, then verify if the markdownContent is visible, and proceed from there.

cbr7:
But by moving the deletekind cluster first you are increasing the chance of errors in other tests because that can fail and then the containers/images will not be deleted and this will impact test suites that are outside the k8s tests.

I don't even see the issue, the createKindCluster method has a check for if the kind cluster already exists, so in reality it shouldn't be a problem if the kind cluster already exists and is shared between test sets.

Your current approach isn't fixing the issue, it at most leaves it the same or worse moving it into a different test suite.

For the markdown check why not just remove it? It's superfluous anyway, it can be removed completely and nothing would change.

amisskii:
> But by moving the deletekind cluster first you are increasing the chance of errors in other tests because that can fail and then the containers/images will not be deleted and this will impact test suites that are outside the k8s tests.
> 
> I don't even see the issue, the createKindCluster method has a check for if the kind cluster already exists, so in reality it shouldn't be a problem if the kind cluster already exists and is shared between test sets.
> 
> Your current approach isn't fixing the issue, it at most leaves it the same or worse moving it into a different test suite.
> 
> For the markdown check why not just remove it? It's superfluous anyway, it can be removed completely and nothing would change.

The check for an existing cluster was blocked by the markdownContent assertion, I removed it.

cbr7 (approved):

PR #9680, opened 2024-10-30
refactor(kube-port): making kube port generic

### LINKED ISSUE #9578: Kubernetes V1Container's ports should show their name if defined

When defining a container we usually set a name, like `web`, `api` etc. Currently in the UI, we do not show this information. We could prefix it with the name

Instead of doing `8080/TCP` we could do `web: 8080/TCP`.

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

@cdrage noted in https://github.com/containers/podman-desktop/pull/9643/ that the `KubeContainerPort` component would need to change to be able to be reused for the Services, the Deployments and the Nodes.

This PR is a refactor of this component to make it generic so it can be used for service (follow-up PR)

### Notable change

- Renaming`KubeContainerPort` to `KubePort`
- Creating a `KubePortInfo` interface, as V1ServicePort, V1ContainerPort are not the same.

### Screenshot / video of UI

N/A

Required for https://github.com/containers/podman-desktop/issues/9273 
Fixes https://github.com/containers/podman-desktop/issues/9578

### How to test this PR?


packages/renderer/src/lib/kube/details/KubeContainerArtifact.svelte
```diff
@@ -2,7 +2,9 @@
 import type { V1Container } from '@kubernetes/client-node';
 
 import Cell from '/@/lib/details/DetailsCell.svelte';
-import KubeContainerPorts from '/@/lib/kube/details/KubeContainerPorts.svelte';
+import { WorkloadKind } from '/@api/kubernetes-port-forward-model';
+
+import KubePorts from './KubePortsList.svelte';
```
feloy: Could we use the same name for the imported resource and the original one, `KubePortsList`?

packages/renderer/src/lib/kube/details/KubePort.svelte
```diff
@@ -1,44 +1,46 @@
 <script lang="ts">
 import { faSquareUpRight, faTrash } from '@fortawesome/free-solid-svg-icons';
-import type { V1ContainerPort } from '@kubernetes/client-node';
 import { Button, ErrorMessage } from '@podman-desktop/ui-svelte';
 
-import { type PortMapping, type UserForwardConfig, WorkloadKind } from '/@api/kubernetes-port-forward-model';
+import type { type PortMapping, type UserForwardConfig, WorkloadKind } from '/@api/kubernetes-port-forward-model';
```
feloy: `type` inside `{ }` are not necessary I suppose

packages/renderer/src/lib/kube/details/KubePortsList.spec.ts
```diff
@@ -22,8 +22,10 @@ import { render } from '@testing-library/svelte';
 import { readable } from 'svelte/store';
 import { beforeEach, expect, test, vi } from 'vitest';
 
-import KubeContainerPorts from '/@/lib/kube/details/KubeContainerPorts.svelte';
 import * as kubeContextStore from '/@/stores/kubernetes-contexts-state';
+import { WorkloadKind } from '/@api/kubernetes-port-forward-model';
+
+import KubePorts from './KubePortsList.svelte';
```
feloy: Same as above

packages/renderer/src/lib/kube/details/KubePortsList.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+import Cell from '/@/lib/details/DetailsCell.svelte';
+import { kubernetesCurrentContextPortForwards } from '/@/stores/kubernetes-contexts-state';
+import type { type UserForwardConfig, WorkloadKind } from '/@api/kubernetes-port-forward-model';
```
feloy: `type` inside `{}` are not necessary?

packages/renderer/src/lib/kube/details/KubePortsList.spec.ts
axel7083: Yes, nice catch, I renamed it twice, sadly the IDE is not smart enough to rename the import :( Updated !

packages/renderer/src/lib/kube/details/KubePortsList.svelte
axel7083: Yes, prettier got crazy, this does not even transpile, fixed

packages/renderer/src/lib/kube/details/KubeContainerArtifact.svelte
axel7083: Done, nice catch

feloy:
@axel7083 do you have an example on how I can have ports listed by the `KubePortsList` component in a Kubernetes pod? For example, I started a pod with the `httpd` image, but I cannot see any port in the pods details

pods-ports

Edit: finally, I got a port displayed by starting a pod with `kubectl run httpd --image httpd --port 80`

axel7083:
> @axel7083 do you have an example on how I can have ports listed by the `KubePortsList` component in a Kubernetes pod? For example, I started a pod with the `httpd` image, but I cannot see any port in the pods details

@feloy have you declared the ports ? Here is an example of an nginx pod

```
apiVersion: v1
kind: Pod
metadata:
  name: static-web
spec:
  containers:
    - image: nginx
      name: web
      ports:
        - containerPort: 80
          name: web
          protocol: TCP
        - containerPort: 8080
          name: hello
          protocol: TCP
```

Just retried it, and here is what the summary page look like.

image

feloy:
> > @axel7083 do you have an example on how I can have ports listed by the `KubePortsList` component in a Kubernetes pod? For example, I started a pod with the `httpd` image, but I cannot see any port in the pods details
> 
> @feloy have you declared the ports ? Here is an example of an nginx pod
> 
> ```
> apiVersion: v1
> kind: Pod
> metadata:
>   name: static-web
> spec:
>   containers:
>     - image: nginx
>       name: web
>       ports:
>         - containerPort: 80
>           name: web
>           protocol: TCP
>         - containerPort: 8080
>           name: hello
>           protocol: TCP
> ```
> 
> Just retried it, and here is what the summary page look like.
> 
> image

OK, got it, thanks.

Is there another place where this component is used and I need to check, or only in this place (the Kubernetes Pod details)?

feloy (approved):
works great for me. Tested with `kubectl run httpd --image httpd --port 80` and looking at the pod's details page, the port is displayed as expected

axel7083:
> Is there another place where this component is used and I need to check, or only in this place (the Kubernetes Pod details)?

No ! The next place would be https://github.com/containers/podman-desktop/pull/9682 :)

PR #9684, opened 2024-10-30
fix(ui): prevent non-tcp port to be forwarded

### LINKED ISSUE #9676: prevent port forwarding for UDP ports as it's not supported by K8S

As raised in https://github.com/containers/podman-desktop/pull/9643#issuecomment-2442365873 we do not support UDP port forwarding, we should prevent user to use it

Either disable the button with a tooltip, or hide the button

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Preventing (frontend) to forward a port non-TCP.

### Screenshot / video of UI

image

Fixes https://github.com/containers/podman-desktop/issues/9676

### How to test this PR?

1. apply a Kubernetes pod with a non-TCP port
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: static-web
  labels:
    role: myrole
spec:
  containers:
    - image: nginx
      name: web
      ports:
        - containerPort: 80
          name: web
          protocol: TCP
        - containerPort: 8080
          name: complex
          protocol: UDP
```
2. Go to pod details
3. assert TCP port have forward button
4. assert UDP port do not have forward button

### DIFF
--- a/packages/renderer/src/lib/kube/details/KubePort.svelte
+++ b/packages/renderer/src/lib/kube/details/KubePort.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
-import { faSquareUpRight, faTrash } from '@fortawesome/free-solid-svg-icons';
-import { Button, ErrorMessage } from '@podman-desktop/ui-svelte';
+import { faQuestionCircle, faSquareUpRight, faTrash } from '@fortawesome/free-solid-svg-icons';
+import { Button, ErrorMessage, Tooltip } from '@podman-desktop/ui-svelte';
+import Fa from 'svelte-fa';
 
 import type { PortMapping, UserForwardConfig, WorkloadKind } from '/@api/kubernetes-port-forward-model';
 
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/kube/details/KubePort.svelte
```diff
@@ -75,18 +76,39 @@ async function removePortForward(): Promise<void> {
...
+      class="px-1 py-0.5"
+      padding="0">
       Remove
     </Button>
-  {:else}
-    <Button title="Forward port {port.value}" disabled={loading} on:click={onForwardRequest.bind(undefined, port)} class="px-1 py-0.5" padding="0">
+  {:else if (port.protocol ?? 'TCP') === 'TCP'}
+    <Button
+      title="Forward port {port.value}"
+      disabled={loading || (!!port.protocol && port.protocol !== 'TCP')}
```
feloy: Can the condition `!!port.protocol && port.protocol !== 'TCP'` be false? (we are in a `if !protocol || protocol === 'TCP'` which is the exact opposite)

axel7083: Oh yeah, it should not be here, sorry at first I though about disabling the button, then I thought it would be better to just remove it

axel7083:
```suggestion
      disabled={loading}
```

cdrage (approved):

benoitf (approved):
non-blocking: just wondering if we need the question mark/tooltip saying it can't be forwarded for UDP b/c I'm not sure it'll more helpful

axel7083:
> non-blocking: just wondering if we need the question mark/tooltip saying it can't be forwarded for UDP b/c I'm not sure it'll more helpful

I don't know TBH, but I do not want to see an issue in the future `forward button is missing on certain port` xD

PR #9687, opened 2024-10-30
feat: create tasks for long running operations when using Podman Desktop API

### LINKED ISSUE #7398: API for (possibly silent) tasks for background jobs

Performing long running actions through our UI usually generates a user visible task, but if another part of the code triggers it or an extension calls an API to do the same thing, there is no task.

Using our API to trigger background jobs like building or pulling images, creating machines, etc. should create tasks. Code in the main package can do this directly; extensions can use API like withProgess. For instance, triggering a build from an extension UI should end up creating a user-visible task just like going through the renderer's build page. We may need to move some existing tasks from the renderer to main.

The API caller should have options like triggering a 'silent' (or maybe background, or child) tasks. For instance, an extension that needs to create a manifest by creating two build tasks and a third task to create the actual manifest might want all three to be visible, might only want a single parent/composite task to be visible to the user, or hide it completely in other cases.

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
This PR creates `Task`s for long running operations called by extensions and adds a new API function that allows extensions create their own `Task`s.

Closes https://github.com/containers/podman-desktop/issues/7398

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2107,6 +2107,31 @@ declare module '@podman-desktop/api' {
     readonly webviewPanel: WebviewPanel;
   }
 
+  export type TaskState = 'running' | 'completed';
+  export type TaskStatus = 'in-progress' | 'success' | 'failure';
+
+  export interface TaskAction {
+    name: string;
+    execute: (task: Task) => void;
+  }
+
+  export interface TaskUpdateEvent {
+    action: 'update' | 'delete';
+    task: Task;
+  }
+
+  export interface Task extends Disposable {
+    readonly id: string;
+    name: string;
+    readonly started: number;
+    state: TaskState;
+    status: TaskStatus;
+    error?: string;
+    progress?: number;
+    action?: TaskAction;
+    readonly onUpdate: Event<TaskUpdateEvent>;
+  }
+
   export namespace window {
     /**
      * Show an information message. Optionally provide an array of items which will be presented as
@@ -2340,6 +2365,8 @@ declare module '@podman-desktop/api' {
     export function createWebviewPanel(viewType: string, title: string, options?: WebviewOptions): WebviewPanel;
 
     export function listWebviews(): Promise<WebviewInfo[]>;
+
+    export function createTask(options?: { title?: string; action?: TaskAction }): Task;
   }
 
   export namespace kubernetes {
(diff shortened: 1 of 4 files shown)


benoitf:
blocker: no API change should be done

IMHO the issue description is misleading with how I think it should be achieved and I disagree with the issue and the suggested pattern 


> For instance, triggering a build from an extension UI should end up creating a user-visible task just like going through the renderer's build page. We may need to move some existing tasks from the renderer to main.

The extension can control notifications using withProgress and that's all
There should not be any silent, create task, etc whatever
Triggering a build from an extension Ui should NOT create a visible task.
It's up to the extension to decide if it's internally or not.


AFAIK I don't want that extension decide if an internal behavior/calling something from the extension is silent or not

If calling from an extension = everything is silent, it's up to the extension to decide what is being created/ displayed

If triggered within from Podman Desktop, then default task can be displayed

deboer-tim:
For cases where an extension just wants to trigger a default task like pulling an image, what would you suggest? Different level of API? It is more work for each extension to replicate the tasks, and things like task name, progress, UI or goto link aren't exposed, and copying them will get out of sync.

benoitf:
If an extension pull an image, I may not care. Also probably I asked a different task so I want to get notification about what I asked for.

If I want to start a http server, I don't want to see a notification telling me it's pulling httpd image and that now this task is ended and I could open this image and run it because this is not the workflow I asked for. It's up to the extension to create the long task and manage the progress for what I've asked for.

benoitf (changes requested):
Extension API calls like pull image, buildImage, etc. should not create any tasks

SoniaSandler:
Should I just leave the createTask API (without creating tasks in other API methods), or remove it too?

deboer-tim:
@benoitf I totally agree for composite tasks or extension-specific ones, but the kind of use I'm asking about is literally just pulling an image. e.g. Bootc extension has links to pull a sample image that should behave identically to the user copy-pasting a name into the pull image form, ideally with the same task name, progress, and link. I'm not sure if you have another option but I don't think bootc should duplicate the UI for this.

benoitf:
>  I'm not sure if you have another option but I don't think bootc should duplicate the UI for this.

Here, podman desktop should register the `pullImage` command and then extensions wanted to 'mimic' how the internal/official command works, could call it.

deboer-tim:
> > I'm not sure if you have another option but I don't think bootc should duplicate the UI for this.
> 
> Here, podman desktop should register the `pullImage` command and then extensions wanted to 'mimic' how the internal/official command works, could call it.

I'm good with that - by default we only do API which doesn't expose any visible task; for the few cases where an extension might need to trigger a task we also expose a command.

I'll add a comment to #7398, suggest we close this PR.

PR #9688, opened 2024-10-30
feat: adding kubernetes navigation path

axel7083:
### What does this PR do?

Inside the `Port Forwarding` kubernetes page, we have a table of each port forwarding, in the requirement (https://github.com/containers/podman-desktop/pull/9645) we seems that we want to be able to forward on Deployment, Service and Pods.

We need to be able to redirect to those resource, as _it is nice to be able to jump to a resource that we are forwarding_[^1], therefore we need navigation ! And instead of using `router.goto` I would rather use the navigation system, but.. Kubernetes route are not defined yet.

[^1]: Quote from UX call 

### Screenshot / video of UI

N/A

Related to https://github.com/containers/podman-desktop/pull/9645

### How to test this PR?


benoitf (approved):
note: if it stays internally why not but it should not leak to the extension API as it'll make harder any attempt to separate kubernetes code from the core part

deboer-tim (approved):

cdrage (approved):

axel7083:
> note: if it stays internally why not but it should not leak to the extension API as it'll make harder any attempt to separate kubernetes code from the core part 
> 
> 

I agree with you!

PR #9689, opened 2024-10-31
chore: replace colors by using registry colors in ui

SoniaSandler:

### What does this PR do?
Updates any bg- and text- hardcoded colors to use the color registry
No changes to the color registry:

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/ExtensionStatus.svelte
+++ b/packages/renderer/src/lib/ui/ExtensionStatus.svelte
@@ -12,55 +12,55 @@ const statusesStyle = new Map<string, ConnectionStatusStyle>([
   [
     'started',
     {
-      bgColor: 'bg-green-500',
-      txtColor: 'text-green-500',
+      bgColor: 'bg-[var(--pd-status-running)]',
+      txtColor: 'text-[var(--pd-status-running)]',
       label: 'ACTIVE',
     },
   ],
   [
     'starting',
     {
-      bgColor: 'bg-green-500',
-      txtColor: 'text-green-500',
+      bgColor: 'bg-[var(--pd-status-starting)]',
+      txtColor: 'text-[var(--pd-status-starting)]',
       label: 'ACTIVATING',
     },
   ],
   [
     'stopped',
     {
-      bgColor: 'bg-gray-900',
-      txtColor: 'text-gray-900',
+      bgColor: 'bg-[var(--pd-status-stopped)]',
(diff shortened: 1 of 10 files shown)


gastoner (approved):
Codewise LGTM

packages/renderer/src/lib/ui/ExtensionStatus.svelte
```diff
@@ -12,55 +12,55 @@ const statusesStyle = new Map<string, ConnectionStatusStyle>([
...
+      txtColor: 'text-[var(--pd-status-terminated)]',
       label: 'FAILED',
     },
   ],
   [
     'downloadable',
     {
-      bgColor: 'bg-purple-600',
-      txtColor: 'text-gray-700',
+      bgColor: 'bg-[var(--pd-notification-dot)]',
```
deboer-tim, line 55: This one is a bit odd... but I don't have a better suggestion atm.

packages/renderer/src/lib/ui/PasswordInput.svelte
```diff
@@ -38,7 +38,7 @@ async function onShowHide() {
   bind:element={element}>
   <svelte:fragment slot="right">
     <button
-      class="px-1 cursor-pointer text-gray-700 group-hover:text-gray-900 group-focus-within:text-gray-900"
+      class="px-1 cursor-pointer text-[var(--pd-content-text)] group-hover:text-[var(--pd-content-sub-header)] group-focus-within:text-[var(--pd-content-sub-header)]"
```
deboer-tim: PasswordInput extends Input, so these should come from `--pd-input-field-`*. Otherwise someone could customize just inputs to be blue and this button wouldn't match. The values from NumberInput would be a good starting point.

deboer-tim (changes requested):
PasswordInput just needs to use input colors, everything else looks good.

packages/renderer/src/lib/ui/ExtensionStatus.svelte
SoniaSandler: I agree that this color choice is not really straight forward, but that's the only color that kind of made sense with the usage context and the original hardcoded color (notifies the user that they can download the extension?)

packages/renderer/src/lib/ui/PasswordInput.svelte
SoniaSandler: fixed!

deboer-tim (approved):

PR #9691, opened 2024-10-31
feat: display nodes events

feloy:

### What does this PR do?

Display nodes events

### Screenshot / video of UI

nodes-events

Part of #6210 

### How to test this PR?

Create a new Kind cluster, and check the details of the node

### DIFF
--- a/packages/renderer/src/lib/node/NodeDetails.svelte
+++ b/packages/renderer/src/lib/node/NodeDetails.svelte
@@ -9,7 +9,6 @@ import { kubernetesCurrentContextEvents, kubernetesCurrentContextNodes } from '/
 
 import Route from '../../Route.svelte';
 import MonacoEditor from '../editor/MonacoEditor.svelte';
-import type { EventUI } from '../events/EventUI';
 import NodeIcon from '../images/NodeIcon.svelte';
 import KubeEditYAML from '../kube/KubeEditYAML.svelte';
 import DetailsPage from '../ui/DetailsPage.svelte';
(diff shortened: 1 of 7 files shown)


axel7083 (commented):
Should the same solution to make it reactive be used than you did in https://github.com/containers/podman-desktop/pull/9672 ? I feel like we have the same issue for reactivity

feloy:
> Should the same solution to make it reactive be used than you did in #9672 ? I feel like we have the same issue for reactivity

When testing, I can see that the display is updated when events change, but I don't clearly understand how.

I'll change to make it clearly reactive as in #9672

cdrage (approved):
LGTM! Tested and was able to show the warnings from my cluster (purposely took a node down).

packages/renderer/src/lib/node/NodeDetails.svelte
```diff
@@ -18,14 +17,17 @@ import { NodeUtils } from './node-utils';
...
-let detailsPage: DetailsPage;
-let kubeNode: V1Node | undefined;
-let kubeError: string;
+let node: NodeUI | undefined = $state(undefined);
+let detailsPage: DetailsPage | undefined = $state(undefined);
+let kubeNode: V1Node | undefined = $state(undefined);
+let kubeError: string | undefined = $state(undefined);
 
-let events: EventUI[] = [];
+let events = $derived($kubernetesCurrentContextEvents.filter(ev => ev.involvedObject.uid === node?.uid));
```
axel7083:
```suggestion
let events: EventUI[] = $derived($kubernetesCurrentContextEvents.filter(ev => ev.involvedObject.uid === node?.uid));
```

suggestion (non-blocking): since we have the types above, might as well adding it too

axel7083 (approved):
LGTM tested it and work great

PR #9695, opened 2024-10-31
chore(test): sync beforeAll timeout for Kind cluster creation

amisskii:

### What does this PR do?
This PR synchronizes the beforeAll hook timeout with the Kind cluster creation timeout.


cbr7:
@amisskii If this PR is to address the timeout in the e2e test from last night, this change will not help, the issue isn't that the timeout expired while waiting, it's that the screen was frozen and it didn't change when it should have. You could set the timeout to 500000 seconds and the test would still time out as it did last night.

amisskii:
> @amisskii If this PR is to address the timeout in the e2e test from last night, this change will not help, the issue isn't that the timeout expired while waiting, it's that the screen was frozen and it didn't change when it should have. You could set the timeout to 500000 seconds and the test would still time out as it did last night.

I see, thank you. However, it’s strange to have a shorter timeout in the beforeAll hook than the cluster creation timeout. I would decrease the cluster creation timeout to 250_000 then; it seems to work stable

cbr7:
@amisskii I don't really mind the timeout increase, my point is that it won't address what failed in the test last night.

cbr7 (approved):

Issue #9288, opened 2024-10-09
Multiple errors like `TypeError: window.function-name is not a function` in `pnpm test:renderer`


dgolovin:
### Bug description

There are multiple errors like one below when running `pnpm test:renderer` which indicates missing mocks for window funcitons.

```
Failed to update image checker providers TypeError: window.getImageCheckerProviders is not a function
    at EventStore.getImageCheckerProvidersInfo [as updater] (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/image-checker-providers.ts:42:17)
    at EventStore.performUpdate (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:142:35)
    at doUpdate (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:204:20)
    at update (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:209:15)
Failed to update views TypeError: window.listViewsContributions is not a function
    at EventStore.listViewsContributions [as updater] (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/views.ts:43:17)
    at EventStore.performUpdate (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:142:35)
    at doUpdate (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:204:20)
    at update (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:209:15)
Failed to update image files providers TypeError: window.getImageFilesProviders is not a function
    at EventStore.getImageFilesProvidersInfo [as updater] (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/image-files-providers.ts:42:17)
    at EventStore.performUpdate (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:142:35)
    at doUpdate (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:204:20)
    at update (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/event-store.ts:209:15)
Failed to update config TypeError: window.getConfigurationProperties is not a function
    at EventStore.fetchConfigurationProperties [as updater] (/Users/eskimo/Sources/podman-desktop/packages/renderer/src/stores/configurationProperties.ts:47:85)
[...]

PR #9700, opened 2024-10-31
fix: handle promise result

### LINKED ISSUE #9699: Restore the no-floating-promise ESLint rule

See https://github.com/containers/podman-desktop/blob/2db0de05182038207f103700ed74326d83792941/eslint.config.mjs#L305

Update code once rule is enabled

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #9699

### What does this PR do?

Update code to handle new no floating promises rule

### Screenshot / video of UI

N/A

#9699

### How to test this PR?

Run unit tests

### DIFF
--- a/packages/renderer/src/lib/pod/PodEmptyScreen.svelte
+++ b/packages/renderer/src/lib/pod/PodEmptyScreen.svelte
@@ -10,9 +10,8 @@ const helloImage = 'quay.io/podman/hello:latest';
 const commandLine = `podman run -dt --pod new:${myFirstPod} ${helloImage}`;
 
 let inProgress = false;
-const provideInfosCopy = $providerInfos;
 
-$: selectedProviderConnection = provideInfosCopy
+$: selectedProviderConnection = $providerInfos
   .map(provider => provider.containerConnections)
   .flat()
   .find(providerContainerConnection => providerContainerConnection.status === 'started');
@@ -30,7 +29,7 @@ async function startPod(): Promise<void> {
       } else {
         await window.showMessageBox({
           title: `Error when running a pod`,
-          message: `Could not find '${helloImage}' in images`,
+          message: `Could not find '${helloImage}'' in images`,
         });
       }
     } catch (error) {
--- a/packages/renderer/src/lib/pod/PodEmptyScreen.svelte
+++ b/packages/renderer/src/lib/pod/PodEmptyScreen.svelte
@@ -10,8 +10,9 @@ const helloImage = 'quay.io/podman/hello:latest';
 const commandLine = `podman run -dt --pod new:${myFirstPod} ${helloImage}`;
 
 let inProgress = false;
+const provideInfosCopy = $providerInfos;
 
-$: selectedProviderConnection = $providerInfos
+$: selectedProviderConnection = provideInfosCopy
   .map(provider => provider.containerConnections)
   .flat()
   .find(providerContainerConnection => providerContainerConnection.status === 'started');
@@ -29,7 +30,7 @@ async function startPod(): Promise<void> {
       } else {
         await window.showMessageBox({
           title: `Error when running a pod`,
-          message: `Could not find '${helloImage}'' in images`,
+          message: `Could not find '${helloImage}' in images`,
         });
       }
     } catch (error) {
--- a/packages/renderer/src/lib/pvc/PVCDetails.svelte
+++ b/packages/renderer/src/lib/pvc/PVCDetails.svelte
@@ -33,12 +33,8 @@ onMount(() => {
   return kubernetesCurrentContextPersistentVolumeClaims.subscribe(pvcs => {
     const matchingPVC = pvcs.find(pvc => pvc.metadata?.name === name && pvc.metadata?.namespace === namespace);
     if (matchingPVC) {
-      try {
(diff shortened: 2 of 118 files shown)


axel7083 (commented):
Amazing job, some tests seems to be failing (expected with the number of change!)

dgolovin:
I fixed some of it. Strange errors in learning-center which I implemented a while ago. Will take a look at it later.

dgolovin:
rebased on top of origin/main.

dgolovin:
One test remains in question. It seems that manual testing reveals the same problem with restarting the terminal for container and it should be investigated as follow up.

dgolovin (approved):
Follow up issue is needed for terminal restart issue.

benoitf:
I would say the test needs to be fixed rather than being skipped

dgolovin:
> I would say the test needs to be fixed rather than being skipped

I agree, but test failure probably related to #9771. My hope was to fix it quick to avoid conflict with #9288 fix.

packages/renderer/src/lib/pod/PodEmptyScreen.svelte
```diff
@@ -10,8 +10,9 @@ const helloImage = 'quay.io/podman/hello:latest';
 const commandLine = `podman run -dt --pod new:${myFirstPod} ${helloImage}`;
 
 let inProgress = false;
+const provideInfosCopy = $providerInfos;
 
-$: selectedProviderConnection = $providerInfos
+$: selectedProviderConnection = provideInfosCopy
```
benoitf: here I'm not sure to follow why there is this change regarding promise ?

packages/renderer/src/lib/pvc/PVCDetails.svelte
```diff
@@ -33,12 +33,8 @@ onMount(() => {
...
     const matchingPVC = pvcs.find(pvc => pvc.metadata?.name === name && pvc.metadata?.namespace === namespace);
     if (matchingPVC) {
-      try {
-        pvc = pvcUtils.getPVCUI(matchingPVC);
-        loadDetails();
-      } catch (err) {
-        console.error(err);
-      }
+      pvc = pvcUtils.getPVCUI(matchingPVC);
+      loadDetails().catch((err: unknown) => console.error(`Error getting PVC details ${pvc.name}`, err));
```
benoitf, line 37: could we have used here `async( pvcs)` and add the await before loadDetails()

packages/renderer/src/lib/window-control-buttons/ControlButtons.spec.ts
```diff
@@ -46,7 +46,7 @@ describe.each([{ platform: 'linux' }, { platform: 'win32' }])('Platform($platfor
   test('Expect minimize is called', async () => {
     render(ControlButtons, { platform });
 
-    const minimizeMock = vi.fn();
+    const minimizeMock = vi.fn().mockResolvedValue(undefined);
```
benoitf: not sure to follow why we have this change ?

packages/renderer/src/lib/window-control-buttons/ControlButtons.svelte
```diff
@@ -4,15 +4,16 @@ import WindowControlButton from './ControlButton.svelte';
...
 }
 
 function maximize(): void {
-  window.windowMaximize();
+  window.windowMaximize().catch((err: unknown) => console.error('Error maximizing window', err));
 }
 
 function close(): void {
-  window.windowClose();
+  window.windowClose().catch((err: unknown) => console.error('Error closing window', err));
```
benoitf: functions can't be async ?

packages/renderer/src/lib/window-control-buttons/ControlButtons.spec.ts
benoitf: usually our functions are async and then error will be thrown to the svelte button

so probably better to add async tot the method + await rather catch

packages/renderer/src/stores/tasks.ts
```diff
@@ -30,7 +30,7 @@ export const tasksInfo: Writable<TaskInfo[]> = writable([]);
 
 // remove element from the store
 export function removeTask(taskId: string): void {
-  window.clearTask(taskId);
+  window.clearTask(taskId).catch((err: unknown) => console.error(`Error removing task ${taskId}`, err));
```
benoitf: prefer async on the function + await rather than promise.catch

benoitf (approved):

packages/renderer/src/lib/pod/PodEmptyScreen.svelte
jeffmaury: Seems unrelated reverted by 7b3e7853b0dfb5515bfbe50164695c227632a782

packages/renderer/src/lib/pvc/PVCDetails.svelte
jeffmaury: Tried it but it it is not supported by the subscribe method signature

packages/renderer/src/lib/window-control-buttons/ControlButtons.svelte
jeffmaury: Done

PR #9708, opened 2024-10-31
fix: create new connection form dropdown FormData

### LINKED ISSUE #9706: Cannot create libkrun podman machine - selected option for a dropdown not respected when submitting the form
### Bug description

On Mac OS when I try to create podman machine using libkrun, it actually creates apple HV.

https://github.com/user-attachments/assets/5a603964-2baa-445f-b9b9-07c2733f94fc

### Operating system

Mac OS 14

### Installation Method

### Version

1.14.0

### Steps to reproduce

Option 1.
1. Install podman desktop on Mac OS
2. Install Podman (5.2.5)
3. Create Podman Machine with LibKRun provider
4. Assert: Machine exists on Resources page, there is Libkrun value shown
Actual result: Machine is created but it is AppleHV VM Type. One can check via cli as well.

Option 2.
0. DO NOT HAVE DOCKER INSTALLED
1. Install podman Desktop, podman, minikube
2. Create and start podman machine (any)
3. Create Minikube cluster with default values (driver == `podman`)
4. Assert: Cluster is created
Actual result: `Exiting due to PROVIDER_DOCKER_NOT_FOUND: The 'docker' provider was not found: exec: "docker": executable file not found in $PATH, install Docker ...`

### Relevant log output

_No response_

### Additional context

I think that it is related to any dropdown in the podman Desktop, not only Podman Machine. 

Possible related PR, issues: https://github.com/containers/podman-desktop/pull/9537, https://github.com/containers/podman-desktop/pull/9458

SoniaSandler:

### What does this PR do?
This PR fixes the `Create new ...` form with the dropdown values by adding a name to the hidden `select` component in `Dropdown` so that `FormData`  would get this field and its value

Fixes https://github.com/containers/podman-desktop/issues/9706

### How to test this PR?
Option 1
1. Install podman desktop on Mac OS
2. Install Podman (5.2.5)
3. Create Podman Machine with LibKRun provider
4. Assert: Machine exists on Resources page, there is Libkrun value shown

Option 2.
0. DO NOT HAVE DOCKER INSTALLED
1. Install podman Desktop, podman, minikube
2. Create and start podman machine (any)
3. Create Minikube cluster with default values (driver == podman)
4. Assert: Cluster is created

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
@@ -350,6 +350,12 @@ async function handleOnSubmit(e: any) {
     if (key.id && !formData.has(key.id) && key.type === 'boolean') {
       data[key.id] = false;
     }
+    if (key.id && !formData.has(key.id)) {
+      const value = configurationValues.get(key.id)?.value;
+      if (value !== undefined) {
+        data[key.id] = value;
+      }
+    }
   });
 
   for (let field of formData) {
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
@@ -350,12 +350,6 @@ async function handleOnSubmit(e: any) {
     if (key.id && !formData.has(key.id) && key.type === 'boolean') {
       data[key.id] = false;
     }
-    if (key.id && !formData.has(key.id)) {
-      const value = configurationValues.get(key.id)?.value;
-      if (value !== undefined) {
-        data[key.id] = value;
-      }
-    }
   });
 
   for (let field of formData) {
--- a/packages/ui/src/lib/dropdown/Dropdown.svelte
+++ b/packages/ui/src/lib/dropdown/Dropdown.svelte
@@ -230,7 +230,7 @@ function onWindowClick(e: Event): void {
     </div>
   {/if}
 
-  <select use:buildOptions class="hidden" bind:value={value}>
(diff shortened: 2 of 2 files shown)


deboer-tim:
question: I haven't had a chance to test yet. I suspect this fixes the issue, but the actual regression is that Dropdown fires events differently than select did? Not sure if it should be fixed there instead.

SoniaSandler:
@deboer-tim TBH, I'm not sure. The `Dropdown` work as intended everywhere (so far in places where I've used it) like the appearance preferences. It also updates the values in the same file I've changed in this PR, but I think the problem might be with https://github.com/containers/podman-desktop/blob/c632348acfe03a686e583926cf2d57688ed6ec1e/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte#L342 
and since I don't know how `FormData` works, I'm not sure what might cause it.
Could it be that `FormData` needs some native components (even if they're nested within our UI components) like `select` and doesn't work with `Dropdown`?

benoitf:
I agree with @deboer-tim that it's not fixing the root cause

If someone is using the UI drop-down component in a form as a replacement from select it should work out of the box in the forms. They should not have to change their code

SoniaSandler:
@benoitf @deboer-tim found the problem with `Dropdown`. The hidden `select` that is bound to the value of the `Dropdown` didn't have a name, and for `FormData`  to get an input field with its value, the input field has to have a name attribute.

benoitf (approved):

odockal (approved):
Tested on Mac OS - works. It seems that the selected value of the dropdown was respected when submitting the form via Create Button click. I was able to create podman machines with both applehv and libkrun provider, and also Kind and Minikube Create Resource forms respected the value selected in the dropdown.

Issue #9502, opened 2024-10-21
pods page take forever to update


axel7083:
### Bug description

When I create a pod from the containers page, it may takes 30s up to a minute to be displayed in the pods page, but in the containers page it is instantaneous 

https://github.com/user-attachments/assets/69c06388-6fe1-41d1-812e-b92c3b48cf9e

Deleting them from the containers page is also instantaneous but takes forever on the pods page, even ghost resources at some point

https://github.com/user-attachments/assets/ac82f8ab-4b63-4a22-9a66-5bee6678deda

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. create pod from nginx container `podman run -d nginx`
2. assert pods not listed directly on the pods page

### Relevant log output

Not much in the logs

### Additional context

- I restarted podman-desktop: same result
- I restarted my podman machine: same result

deboer-tim:
I haven't hit this issue, but I did see some lagginess last week while comparing to other kube objects, and I have a feeling this might be because pods page (kube client) was never updated to use the informers.

cdrage:
https://github.com/user-attachments/assets/6ad452be-d03d-42c9-aab6-c0eeade34fa6

I'm experiencing this issue as well with a general lagginess when trying to delete / restart / etc.

odockal:
@deboer-tim Since it seems that you have a patch ready (might be incomplete?), I have assigned you to the issue and added it to the actual sprint. Feel free to update anything that does not seems right.

deboer-tim:
The text I added was 'possibly needed to fix #9502' and unfortunately Github views that as 'fixes'. I've changed the text in the PR, and am unassigning myself as I am not directly intending to fix this issue.

There is a short term fix possible, but IMHO the better solution is to completely separate Podman pods from Kubernetes pods, using the same informer code paths as we do for all other Kube resources instead of having special cases/code paths for pods.

shipsing:
I noticed this behavior:
After deploying a pod to a Kubernetes cluster, the Pods component page does not automatically refresh and you would not be able to see the newly created pod. 
I was able to see the newly created pod after using this manual approach:

- Switch to another Kube context and then switch back to the original context. After doing this, you can view the Pods component page in an up-to-date state.

dgolovin:
Probably the same problem described in this issue #10356.

deboer-tim:
[...]

Issue #8818, opened 2024-09-10
Move Kubernetes Pods to separate section / create new section


cdrage:
### Is your enhancement related to a problem? Please describe

Pods are not located in the Kubernetes section and requires us to click "away" from it to the Pods section.

### Describe the solution you'd like

"Kubernetes Pods" could be better placed in the Kubernetes section.

Issue #9169, opened 2024-10-02
Kubernetes mechanisms 


axel7083:
### Is your enhancement related to a problem? Please describe

While working on https://github.com/containers/podman-desktop/issues/8341 I had to dive into the Kubernetes code in Podman Desktop. Code I haven't really much touch since I arrived.

When I first started reading it, I had to ask for clarification due to its complexity, and I opened https://github.com/containers/podman-desktop/issues/8775 that @feloy worked on.

---

I will in this issue try to cover the Kubernetes mechanisms

The mains classes related to Kubernetes features are the following

https://github.com/containers/podman-desktop/blob/d12f5423df41f098e9c2a65e0f67755aaf1ad031/packages/main/src/plugin/kubernetes/kubernetes-client.ts#L149

https://github.com/containers/podman-desktop/blob/d12f5423df41f098e9c2a65e0f67755aaf1ad031/packages/main/src/plugin/kubernetes/contexts-manager.ts#L96

There is a particular interactions between `KubernetesClient` and the `ContextsManager`

Let's dig a bit by exploring the `KubernetesClient#init` function.

```mermaid
sequenceDiagram
    System->>KubernetesClient: init()
    KubernetesClient-->>KubernetesClient: registerConfigurations()
    Note right of KubernetesClient: watch the kubeconfig
    KubernetesClient-->>KubernetesClient: setupWatcher() 
    critical setupKubeWatcher()
        Note left of KubernetesClient: Pod and Deployment watchers
        KubernetesClient-->>KubernetesClient: fetchAPIGroups()
        KubernetesClient->>KubernetesClient: watch /pods
        KubernetesClient->>KubernetesClient: watch /deployments
    end
    Note right of KubernetesClient: Refresh is called every time the kubeconfig is updated
    critical refresh()
        KubernetesClient-->>KubernetesClient: load new KubeConfig
        KubernetesClient-->>ContextsManager: update kubeconfig
        loop For each Context
            Note left of ContextsManager: Pod and Deployment
            ContextsManager-->>ContextsManager: createKubeContextInformers() 
            Note left of ContextsManager: SecondaryResources
            ContextsManager-->>ContextsManager: startResourceInformer() 
        end
    end
```

## Informers vs Watchers

We have two concept here, which have distinct logic over two different classes

### The watchers (Pod and Deployment)

There are managed by the `KubernetesClient`.

[...]

PR #9723, opened 2024-11-01
chore: kubernetes pods store

deboer-tim:
### What does this PR do?

We have informers for all the Kube objects, except for pods where we have the informer in main but don't expose the store in renderer.

### Screenshot / video of UI

N/A

For consistency/completeness, a very small part of #8818, and possibly needed for #9502.

### How to test this PR?

PR checks enough for now.

### DIFF
--- a/packages/renderer/src/stores/kubernetes-contexts-state.ts
+++ b/packages/renderer/src/stores/kubernetes-contexts-state.ts
@@ -166,6 +166,26 @@ export const kubernetesCurrentContextNodesFiltered = derived(
   ([$searchPattern, $nodes]) => $nodes.filter(node => findMatchInLeaves(node, $searchPattern.toLowerCase())),
 );
 
+// Pods
+
+export const kubernetesCurrentContextPods = readable<KubernetesObject[]>([], set => {
+  window.kubernetesRegisterGetCurrentContextResources('pods').then(value => set(value));
+  window.events?.receive('kubernetes-current-context-pods-update', (value: unknown) => {
+    set(value as KubernetesObject[]);
+  });
+  return () => {
+    window.kubernetesUnregisterGetCurrentContextResources('pods');
+  };
+});
+
+export const podSearchPattern = writable('');
+
+// The pods in the current context, filtered with `podSearchPattern`
+export const kubernetesCurrentContextPodsFiltered = derived(
+  [podSearchPattern, kubernetesCurrentContextPods],
+  ([$searchPattern, $pods]) => $pods.filter(pod => findMatchInLeaves(pod, $searchPattern.toLowerCase())),
+);
+
 // PersistentVolumeClaims
 
 export const kubernetesCurrentContextPersistentVolumeClaims = readable<KubernetesObject[]>([], set => {
--- a/packages/renderer/src/stores/kubernetes-contexts-state.ts
+++ b/packages/renderer/src/stores/kubernetes-contexts-state.ts
@@ -169,12 +169,17 @@ export const kubernetesCurrentContextNodesFiltered = derived(
 // Pods
 
 export const kubernetesCurrentContextPods = readable<KubernetesObject[]>([], set => {
-  window.kubernetesRegisterGetCurrentContextResources('pods').then(value => set(value));
+  window
+    .kubernetesRegisterGetCurrentContextResources('pods')
+    .then(value => set(value))
+    .catch((err: unknown) => console.log('Error registering Kubernetes pods', err));
   window.events?.receive('kubernetes-current-context-pods-update', (value: unknown) => {
     set(value as KubernetesObject[]);
   });
   return () => {
-    window.kubernetesUnregisterGetCurrentContextResources('pods');
+    window
+      .kubernetesUnregisterGetCurrentContextResources('pods')
+      .catch((err: unknown) => console.log('Error unregistering Kubernetes pods', err));
   };
 });
 
--- a/packages/renderer/src/stores/kubernetes-contexts-state.spec.ts
+++ b/packages/renderer/src/stores/kubernetes-contexts-state.spec.ts
@@ -36,7 +36,9 @@ import {
 const kubernetesRegisterGetCurrentContextResourcesMock: Mock<
   (resourceName: ResourceName) => Promise<KubernetesObject[]>
 > = vi.fn();
-const kubernetesUnregisterGetCurrentContextResourcesMock: Mock<(resourceName: ResourceName) => Promise<[]>> = vi.fn();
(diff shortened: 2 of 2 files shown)


cdrage (approved):
LGTM as it just exposes it, I'm assuming no need for tests as there is no test file for this / just exposes.

axel7083 (changes requested):
What would be the following steps ? We have some discussion started on Kubernetes in https://github.com/containers/podman-desktop/issues/9169 but I don't see any issue requested the change in its requirements.

Would adding those lines duplicate the numbers of watchers for pods ?

We have the `primary` informers (pods, services), if we need a store here, where does the data came from ? Which informers/watcher ? What system would it replace ?

I can summarise my concern: does adding this store have side effects ? Or it is consuming an existing already created watcher (cc @feloy )

:information: I'm putting a request change as we need to make sure that we are not multiplying the number of connection here (might be wrong, and I'm sorry if that's the case, but I'm having some doubts)

deboer-tim:
> What would be the following steps ? We have some discussion started on Kubernetes in #9169 but I don't see any issue requested the change in its requirements.

Currently, we have the informers described in #9169 for deployments, pods, nodes, services, ingress, etc., and these are all exposed to the renderer... except pods, where prior to this PR only the count from the informer is exposed, and used in Settings > Kubernetes.

Then for pods we have a completely different path from before the informers, with podman/docker socket plus Kubernetes watcher, used by the Pods page.

Sorry you asked yet? :) My intention over time is that everything Kube-related should *only* have the single informer path - with whatever improvements we come up with, which would apply to everything equally/at once. Kube pods should not also have a different path than everything else Kube-related.

> Would adding those lines duplicate the numbers of watchers for pods ?

Until something uses it, no. And then after, still unlikely because pods are a primary informer for the current context anyway. And, we could eventually drop the pod watcher so we're not watching it both ways.

> We have the `primary` informers (pods, services), if we need a store here, where does the data came from ? Which informers/watcher ? What system would it replace ?

We already have the primary informer for pods, this just makes it available for renderer to use too.

> I can summarise my concern: does adding this store have side effects ? Or it is consuming an existing already created watcher (cc @feloy )

Yeah, @feloy is the expert here and can confirm. But AFAIK until something uses this it does nothing, and even when something does it's just pulling data we're already collecting. But more importantly to me / even if it wasn't, it's more important that this is one (small) step closer to everything in Kube working identically / with one code path.

> :information: I'm putting a request change as we need to make sure that we are not multiplying the number of connection here (might be wrong, and I'm sorry if that's the case, but I'm having some doubts)

Yeah, np, that is a valid concern. Let's make sure it's all explained and valid.

axel7083:
> Yeah, np, that is a valid concern. Let's make sure it's all explained and valid.

Thanks for the detailed explanation, context and plans !

axel7083 (approved):
LGTM :+1:

feloy (approved):
Tested by displaying the number of pods in the dashboard, works as expected

benoitf (commented):
I don't see unit tests ?

deboer-tim:
Rebased and tests covering pod store added, as well as services/deployments/nodes, @benoitf please take a quick look.

packages/renderer/src/stores/kubernetes-contexts-state.ts
```diff
@@ -166,6 +166,31 @@ export const kubernetesCurrentContextNodesFiltered = derived(
...
 
+// Pods
+
+export const kubernetesCurrentContextPods = readable<KubernetesObject[]>([], set => {
+  window
+    .kubernetesRegisterGetCurrentContextResources('pods')
+    .then(value => set(value))
+    .catch((err: unknown) => console.log('Error registering Kubernetes pods', err));
+  window.events?.receive('kubernetes-current-context-pods-update', (value: unknown) => {
+    set(value as KubernetesObject[]);
```
benoitf, line 177: issue: need a check before the cast ?
(if the event is not the one expected)

packages/renderer/src/stores/kubernetes-contexts-state.ts
```diff
@@ -166,6 +166,31 @@ export const kubernetesCurrentContextNodesFiltered = derived(
   ([$searchPattern, $nodes]) => $nodes.filter(node => findMatchInLeaves(node, $searchPattern.toLowerCase())),
 );
 
+// Pods
+
+export const kubernetesCurrentContextPods = readable<KubernetesObject[]>([], set => {
+  window
+    .kubernetesRegisterGetCurrentContextResources('pods')
+    .then(value => set(value))
+    .catch((err: unknown) => console.log('Error registering Kubernetes pods', err));
```
benoitf, line 175: pure question: do we expect it's only a `.log` or it's an `.error`

packages/renderer/src/stores/kubernetes-contexts-state.ts
```diff
@@ -166,6 +166,31 @@ export const kubernetesCurrentContextNodesFiltered = derived(
...
+  window.events?.receive('kubernetes-current-context-pods-update', (value: unknown) => {
+    set(value as KubernetesObject[]);
+  });
+  return () => {
+    window
+      .kubernetesUnregisterGetCurrentContextResources('pods')
+      .catch((err: unknown) => console.log('Error unregistering Kubernetes pods', err));
```
benoitf, line 182: pure question: do we expect it's only a `.log` or it's an `.error`

packages/renderer/src/stores/kubernetes-contexts-state.ts
deboer-tim: This pattern was merged 5 days ago in #9700 for all the Kubernetes stores. I agree it probably is unexpected enough to be an `.error`, but separate PR.

packages/renderer/src/stores/kubernetes-contexts-state.ts
deboer-tim: Is there anything we can do to enforce event-specific type safety / allow typed values?

In general I agree we should, or have more defensive code, but this same pattern is used 8 other places within this file, as well as several other events, and many commands that have arguments. I can open a new issue to cover this, but IMHO for this PR it is best to follow the existing pattern.

packages/renderer/src/stores/kubernetes-contexts-state.ts
deboer-tim: Ditto, same pattern from #9700.

benoitf (approved):
questions are follow ups

packages/renderer/src/stores/kubernetes-contexts-state.ts
deboer-tim: #9866 opened as follow-up.

Issue #8923, opened 2024-09-17
UX: Move the provider elements to the Statusbar


Firewall:
In order to have more space on the dashboard we want to move some of the information that takes a lot of space to the statusbar. In this step we want to put the current providers (aka resources) (podman, kubectl, kind, developer sandbox and compose) to be in the statusbar instead of the dashboard. 

More info in the [Epic](https://github.com/containers/podman-desktop/issues/8922)

**Requirements:** 
- We create a pattern that can be re-used for every provider.
  - For each provider we should have indicator for status. Possible options for status are:
      - 🟢 Healthy (All good)
      - 🔴 Unhealthy (some sort of error)
      - 🆙  Update available (There is an update for this provider available)
      - ⚠  Not setup (the onboarding for the provider is not completed, something still needs to be set up)
      - ⏹ Stopped (this provider is currently stopped)
  - Clicking (or right clicking maybe?) the icon should take the user to the resources/provider page where they can configure them
  - Depending on the plugin, clicking on the icon should also have an effect. eg. Clicking on the Kubernetes context icon lets the user switch to another kubernetes context.
- Current list of providers that should be in the StatusBar: (Podman machine, Openshift local, Kubectl, kind, developer sandbox, compose)
- idea: Make the statusbar slightly taller so that the icons are a bit bigger and thus are more clear to understand.
Masha already an exploration on that I think ![Image](https://github.com/user-attachments/assets/83b18fb0-92dd-44cb-9736-fcf1148d0384)

- idea: Currently there is [another task to clean up the right side of the statusbar](https://github.com/containers/podman-desktop/issues/9077), but it is possible that we include those changes here

ekidneyrh:
@Firewall I just started to look at this. There was a few things I was wondering.

The list you have above for providers: podman, kubectl, kind, developer sandbox and compose - is this the full list? 

Is the order of things that would be displayed on the left side of the bar be: provider, context, extension, cluster?  would all of these have status's or just the provider? I know atm some of the other items have errors. 

I agree with making it taller and I also think we could potentially make the bar in general be a bit bigger to allow for bigger icons so they are more noticeable.

Firewall:
> The list you have above for providers: podman, kubectl, kind, developer sandbox and compose - is this the full list?

I think there are some more that can be provided by extensions, like MiniKube and Lima. That's why I will emphazise that we need a pattern as there might be more in the future that want to appear there. 

[...]

PR #9724, opened 2024-11-01
chore: add ProviderWidget component for StatusBar

### LINKED ISSUE #9555: Generic status widget for a provider

As part of the [Move providers to the statusbar epic](https://github.com/containers/podman-desktop/issues/8922), we want to create a consistent widget for the statusbar for each provider. This will help us create similar behaviors between all our providers. 

- A widget that can be reused for each of the providers
- A widget can show the status of provider (matching [the design](https://github.com/containers/podman-desktop/issues/8923)). Start/Starting/Stopped/Stopping/Failed
- Clicking on the widget takes you to the resources page
- A widget can be pinned and unpinned from the dropdown menu
- By default we will only have the `Podman` container provider pinned

In some cases a provider can have multiple instances (eg. you can have multiple machines). In those cases we will keep the same behavior as on the dashboard currently and show the status of the most recent one.

SoniaSandler:

### What does this PR do?
This PR introduces a new ui component `ProviderWidget` that will be used as a part of the Dashboard revamp to show providers in the status bar

### Screenshot / video of UI
(used the ProvidorInfo store for this example)
Screenshot from 2024-11-21 10-27-32

Closes https://github.com/containers/podman-desktop/issues/9555

### How to test this PR?


SoniaSandler:
@deboer-tim @benoitf @Firewall This PR currently uses `ProviderInfo`, the same information that is used in the Dashboard page, to show providers such as Podman and Developer Sandbox in the status bar. 
However, in the design issue, I also see that the Kubernetes context and things like AI lab will be shown in the status bar. 
How will their information be passed to the status bar? For example, the context uses `StatusBarEntry` to get info from main.
For this widget, will there be multiple types of information? or a new type/registry will be created that encompasses all of the needed information for all the entries in this change?
Also, since the design issue is still in progress/review, are we going to include the provider names?

SoniaSandler:
~~Still need to add tests for `ProviderWidget`~~

benoitf (dismissed):
we can't hardcode anything on the frontend side

The providers are provided by the extensions. So we can't have any OpenShift, developer sandbox icons, etc in Podman Desktop core. This is why we have extensions

So, no ProviderIcons. It's an anti-pattern.

benoitf:
moving to draft as it could be a prototype but it can't be merged using this design pattern.

SoniaSandler:
@benoitf I see what you mean. So in order to have the white provider icons in the status bar, we'll have to add each one of them to their respective extension?

benoitf:
@SoniaSandler 
yes, the idea is that a provider expose its metadata, like we do have the providers that register some icons and fonts already (kind font) for example in their contribution

Then, if I enable a 3rd party extension contributing a provider, Podman Desktop has the list of providers and can use this metadata.

If Podman Desktop is aware of what is brought by the extension, it's not something expected.


that said a provider is already providing icons when registering a provider, so it would not be convenient and backward compliant if a provider should register yet another icon.

so maybe convert on registration the icon to two colors. (or another pattern)

SoniaSandler:
@benoitf 
grayscale option for now:
Screenshot from 2024-11-21 10-27-32

vs regular icons:
Screenshot from 2024-11-21 10-27-41

benoitf:
thanks, we can iterate on the specific topic of the color as follow-up later

SoniaSandler:
I think I'll leave the regular icons for now since the grayscale (especially for OpenShift) looks a bit weird

benoitf:
I think you can keep it, AFAIK it's not a default extension and it'll pin for now the podman provider where the rendering is ok

SoniaSandler:
> I think you can keep it, AFAIK it's not a default extension and it'll pin for now the podman provider where the rendering is ok

keep the grayscale or the color icons?

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
```diff
@@ -0,0 +1,68 @@
+<script lang="ts">
+import { router } from 'tinro';
+
+import type { ProviderInfo } from '/@api/provider-info';
+
+import IconImage from '../appearance/IconImage.svelte';
+
+let { entry, command = () => router.goto('/preferences/resources') }: { entry: ProviderInfo; command?: () => void } =
```
axel7083: Please define an interface Prop above

```ts
interface Props {
 entry: ProviderInfo;
 command?: () => void
}

let { entry, command }: Props = $props();
```

easier to read

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
```diff
@@ -0,0 +1,68 @@
...
+      statusesStyle.get(entry.containerConnections[0].status) ??
+      'bg-[var(--pd-status-not-running)] text-[var(--pd-status-not-running)]';
+  } else if (entry.kubernetesConnections.length > 0) {
+    statusStyle =
+      statusesStyle.get(entry.kubernetesConnections[0].status) ??
+      'bg-[var(--pd-status-not-running)] text-[var(--pd-status-not-running)]';
+  } else {
+    statusStyle = statusesStyle.get(entry.status) ?? 'bg-gray-900 text-[var(--pd-content-text)]';
+  }
+});
```
axel7083: This could be a derived state since only `statusStyle` is modified

Something like the following
```ts
let statusStyle = $devided.by(() => {
  if (entry.containerConnections.length > 0) {
    return
      statusesStyle.get(entry.containerConnections[0].status) ??
      'bg-[var(--pd-status-not-running)] text-[var(--pd-status-not-running)]';
  } else if (entry.kubernetesConnections.length > 0) {
    return
      statusesStyle.get(entry.kubernetesConnections[0].status) ??
      'bg-[var(--pd-status-not-running)] text-[var(--pd-status-not-running)]';
  } else {
    return statusesStyle.get(entry.status) ?? 'bg-gray-900 text-[var(--pd-content-text)]';
  }
})
```

benoitf:
> keep the grayscale or the color icons?

grayscale

packages/renderer/src/lib/statusbar/ProviderWidget.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+test('Expect title to include Kubernetes provider connections', () => {
+  providerMock.kubernetesConnections = [
+    { name: 'connection 1' } as unknown as ProviderKubernetesConnectionInfo,
+    { name: 'connection 2' } as unknown as ProviderKubernetesConnectionInfo,
+  ];
+  render(ProviderWidget, { entry: providerMock });
+
+  expect(screen.getByTitle('connection 1, connection 2')).toBeInTheDocument();
+});
```
axel7083, line 98: great testing !

benoitf:
before being able to merge, we need an experimental flag (should be a separate pull request)

then by default, even if it is merged we won't see it until we opt-it

SoniaSandler:
@benoitf I opened https://github.com/podman-desktop/podman-desktop/pull/10061.
This PR only adds the provider widget component, which isn't used anywhere yet, so it is not visible yet.

SoniaSandler:
https://github.com/podman-desktop/podman-desktop/pull/10061 has been merged (experimental setting for statusbar providers)

axel7083:
I am not sure to understand how to test this PR

> #10061 has been merged (experimental setting for statusbar providers)

What is the link with this PR ?

SoniaSandler:
> before being able to merge, we need an experimental flag (should be a separate pull request)
> 
> then by default, even if it is merged we won't see it until we opt-it

@axel7083 I'm not entirely sure but it was requested ^
This PR doesn't add any visual or functional changes, but just introduces the UI component that will be used for providers in the status bar.
You can test it by adding 
```
{#each $providerInfos as entry}
    <ProviderWidget entry={entry}/>
{/each}
```
in `StatusBar.svelte` line 70. (sorry for not adding this earlier)

The status bar might change its height, but it can be easily fixed with the right styling properties (no overflow).
It is out of the scope for this PR, and can be fixed once this component will be used.

axel7083:
> You can test it by adding ...

Thanks that was what I was looking for.

## Implementation result

Yersterday @Firewall asked me to review so here it is :woman_dancing: 

So here is the screenshots of all providers in the status bar (this PR)

image

My first feedback: I don't understand the icons, half moon for compose ? circle in circle for kind ?

Let's decompose. 

## The design From https://github.com/podman-desktop/podman-desktop/issues/8923#issuecomment-2485475105

The issue has been closed without an engineering side view so here are my comments/feedback, and I am sorry @SoniaSandler to comment here

image

1. I don't understand why we have an icon for some providers, and not others ? This is not clear in the design cc @ekidneyrh do you have some clarification ? (Here podman has a circle icon but not kind ?) Does it depend on the status of the provider ? Maybe we should have a clear table with a matching, and all the providers listed to see ?)
2. ai lab is not a provider, it just add text to the status bar

Some weird scenarios: the provider without a status. It is always with an unknown status

compsoe 
https://github.com/podman-desktop/podman-desktop/blob/cdee0b7ab1b158a4b4d44a6d5bb90554490eceb1/extensions/compose/src/extension.ts#L218

Same for kind

https://github.com/podman-desktop/podman-desktop/blob/9d80c320c1ef4ace724dfa0c1a89abcd0c417299/extensions/kind/src/extension.ts#L258

How would we handle that ? Sonia added a default icon `half moon` when unknown status, which is not something clear for the final implementation (IMO, cc @jeffmaury share my opinion)

In the design I am also missing, how do I change Kubernetes context ?

Firewall:
@axel7083 and me discussed this a bit but I would be happy to discuss it with you when you are back to @SoniaSandler. 

I also learned some things along the way, so sorry if I make some mistakes. 

> I don't understand why we have an icon for some providers, and not others ? This is not clear in the design cc @Emma Kidney do you have some clarification ? (Here podman has a circle icon but not kind ?) Does it depend on the status of the provider ? Maybe we should have a clear table with a matching, and all the providers listed to see ?)

`kind` doesn't doesn't have a status in the current implementation. I have [created a task](https://github.com/podman-desktop/podman-desktop/issues/10141) to improve that. 

However, probably in the mean time, an extension where the `status = unknown` we should probably not show them on the statusbar. Not only does the "moon" icon look out of place there isn't much benefit to having these providers there. They are not adding new information for users and just add clutter.

> ai lab is not a provider, it just add text to the status bar

Correct, this task is only for providers. On top of that for AI Lab [we are going to make some changes to remove it form the statusbar ](https://github.com/containers/podman-desktop-extension-ai-lab/issues/1978)

> In the design I am also missing, how do I change Kubernetes context ?

In the image you shared, the `kind-cluster` is the Kubernetes context selector.  No changes in the behavior there. 
image

SoniaSandler:
@Firewall I think that we should keep the unknown status icon here to at least have a fallback for that if we ever need it, but once we render/add the different providers to the status bar using this component, we can filter them so that only providers with not unknown status will be shown there

Firewall:
@SoniaSandler That makes sense. However, can we change it to a more discriptive icon? Maybe a `?`. The moon icon is really confusing if a user ever sees that, even acidentally.

SoniaSandler:
@Firewall @axel7083 updated moon icon to question mark
Screenshot from 2024-12-13 00-30-41

axel7083 (approved):

Firewall:
I will let others review the code but that looks awesome, thanks!

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
```diff
@@ -0,0 +1,73 @@
...
+  ['stopped', 'fa-regular fa-circle-dot'],
+  ['stopping', 'animate-spin border border-solid border-[var(--pd-action-button-spinner)] border-t-transparent'],
+  ['error', 'fa-regular fa-circle-xmark'],
+  ['not-installed', 'fa fa-exclamation-triangle'],
+  ['update available', 'fa-regular fa-circle-up'],
+  ['unknown', 'fa-regular fa-circle-question'],
+]);
+</script>
+  
+<button
```
jeffmaury: Any reason not to use Button component ?

jeffmaury (approved):
LGTM

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
SoniaSandler: no specific reason, but I can update it to the `Button` component

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
```diff
@@ -0,0 +1,73 @@
...
+
+function tooltipText(entry: ProviderInfo): string {
+  let tooltip = '';
+  if (entry.containerConnections.length > 0) {
+    tooltip = entry.containerConnections.map(c => c.name).join(', ');
+  } else if (entry.kubernetesConnections.length > 0) {
+    tooltip = entry.kubernetesConnections.map(c => c.name).join(', ');
+  }
+  return tooltip;
+}
```
benoitf: question: why do we have a function there and we call then tooltipText(entry) at the bottom ?

issue: It looks like we do have a single entry for the whole object, so why do we need to give it as a parameter ?

it shoul be simplified by

`let tooltipText = $derived(....)`

and later `title={tooltipText}`

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
```diff
@@ -0,0 +1,73 @@
...
+  ['ready', 'fa-regular fa-circle-check'],
+  ['started', 'fa-regular fa-circle-check'],
+  ['starting', 'animate-spin border border-solid border-[var(--pd-action-button-spinner)] border-t-transparent'],
+  ['stopped', 'fa-regular fa-circle-dot'],
+  ['stopping', 'animate-spin border border-solid border-[var(--pd-action-button-spinner)] border-t-transparent'],
+  ['error', 'fa-regular fa-circle-xmark'],
+  ['not-installed', 'fa fa-exclamation-triangle'],
+  ['update available', 'fa-regular fa-circle-up'],
+  ['unknown', 'fa-regular fa-circle-question'],
+]);
```
benoitf: issue: this kind of pattern is not really working with tailwind (as we parse some constants, it's close to the limit)

pattern is `class:border-t-transparent={the condition to apply this class}`

I would expect also a dedicated ProviderWidgetStatus component handling only statuses

benoitf (dismissed):

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
benoitf: you may look at StatusIcon component to give you an idea

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
SoniaSandler: updated to use `Button`

axel7083 (approved):
LGTM

jeffmaury (approved):
LGTM.

Maybe extend ProviderWidgetStatus.spec.ts to add a test to the click event

SoniaSandler:
> Maybe extend ProviderWidgetStatus.spec.ts to add a test to the click event

@jeffmaury I'm not sure I understand since there is nothing to click on in the status. The ProviderWidget already includes test for the click action

jeffmaury:
> > Maybe extend ProviderWidgetStatus.spec.ts to add a test to the click event
> 
> @jeffmaury I'm not sure I understand since there is nothing to click on in the status. The ProviderWidget already includes test for the click action

You're right. So rephrasing: why do you need to mock tinro in ProviderWidgetStatus.spec.ts

SoniaSandler:
@jeffmaury I think I forgot to remove it when I copied the status tests :sweat_smile: . Thanks for catching it, I'll remove the mock

PR #9757, opened 2024-11-05
feat: display a warning when there is no latest tag for image to pull

### LINKED ISSUE #8927: pulling/searching images without latest tag

now, it's possible to get typeahead completion on images and on the tags

I would like to be notified if the image has not a latest tag (it does not exists a lot but it's still exist so I should get a warning if image is recognized/found and we see the tag does not exist so I should pick up a tag)

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Display a warning when no latest tag for image to pull.

### Screenshot / video of UI

https://github.com/user-attachments/assets/01abd98b-f992-4d03-b1d2-9630ab210b6f

Fixes #8927 

### How to test this PR?

Try to pull an image with or without "latest" tag (for example quay.io/phmartin/* images mostly don't have it)


packages/renderer/src/lib/image/PullImage.svelte
```diff
@@ -204,6 +205,29 @@ async function searchImages(value: string): Promise<string[]> {
...
+  }
+  try {
+    let image = imageToPull;
+    if (image.startsWith(DOCKER_PREFIX + '/')) {
+      image = image.slice(DOCKER_PREFIX.length + 1);
+    }
+    const tags = await window.listImageTagsInRegistry({ image });
+    const latestFound = tags.includes('latest');
+    if (!latestFound) {
+      latestTagMessage = '"latest" tag not found. You can search a tag by typing ":"';
```
benoitf: question: wondering about replacing typing word by appending or adding

benoitf (approved):
image

works as expected 👍

packages/renderer/src/lib/image/PullImage.svelte
```diff
@@ -204,6 +205,29 @@ async function searchImages(value: string): Promise<string[]> {
...
+let latestTagMessage: string | undefined = undefined;
+async function searchLatestTag(): Promise<void> {
+  if (imageNameIsInvalid || !imageToPull || imageToPull.includes(':')) {
+    latestTagMessage = undefined;
+    return;
+  }
+  try {
+    let image = imageToPull;
+    if (image.startsWith(DOCKER_PREFIX + '/')) {
+      image = image.slice(DOCKER_PREFIX.length + 1);
```
benoitf: question: should it be another constant ?

so no need to add a new / each time or to add +1 as well

packages/renderer/src/lib/image/PullImage.svelte
feloy: Appending seems more accurate to me. I'll change

packages/renderer/src/lib/image/PullImage.svelte
feloy: I'm not sure to find a constant name more explicit than these formulas

benoitf: top of the file near const DOCKER_PREFIX = 'docker.io';
```ts
const DOCKER_PREFIX = 'docker.io';
const DOCKER_PREFIX_WITH_SLASH = DOCKER_PREFIX + '/'
```

and then here

```ts
if (image.startsWith(DOCKER_PREFIX_WITH_SLASH)) {
  image = image.slice(DOCKER_PREFIX_WITH_SLASH.length);
}
```

dgolovin (approved):
Resolved conflicts and checked locally.
LGTM.

PR #9762, opened 2024-11-05
chore: add option to view nav icon title

### LINKED ISSUE #9662: Provide default option for showing icon + title in navigation

As per epic #9191, we've had requests from new users to show the title for each navigation item.

- Add a configuration setting along with support for showing either just 'nav icon' (current) or 'nav icon + title' (new) in the main navigation.
- When the title is shown the tooltip should only show the counter.
- The option should be on by default (for new users).
- Need to confirm with design whether the toggle should be in the navigation popup, preferences, or both.

_No response_

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
This PR adds a preferences option to show titles along with the icons in the navigation menu. This option by default is on, and the user can change it in the settings.
Possible follow ups - increase the width of the nav menu and add this option to the navigation popup (if needed)

### Screenshot / video of UI
Screenshot from 2024-11-05 13-33-11

Screenshot from 2024-11-04 23-20-14

Closes https://github.com/containers/podman-desktop/issues/9662

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/appearance-init.ts
+++ b/packages/main/src/plugin/appearance-init.ts
@@ -50,8 +50,8 @@ export class AppearanceInit {
         [`${AppearanceSettings.SectionName}.${AppearanceSettings.NavigationAppearance}`]: {
           description: 'Select icon and title or just icon for navigation icons',
           type: 'string',
-          enum: ['icon + title', 'icon'],
-          default: import.meta.env.DEV ? 'icon + title' : 'icon',
+          enum: [AppearanceSettings.IconAndTitle, AppearanceSettings.Icon],
+          default: import.meta.env.DEV ? AppearanceSettings.IconAndTitle : AppearanceSettings.Icon,
         },
       },
     };
--- a/packages/main/src/plugin/appearance-init.ts
+++ b/packages/main/src/plugin/appearance-init.ts
@@ -47,6 +47,12 @@ export class AppearanceInit {
           default: 0,
           step: 0.1,
         },
+        [`${AppearanceSettings.SectionName}.${AppearanceSettings.NavigationAppearance}`]: {
+          description: 'Select icon and title or just icon for navigation icons',
+          type: 'string',
+          enum: ['icon and title', 'icon'],
+          default: 'icon and title',
+        },
       },
     };
 
--- a/packages/main/src/plugin/appearance-init.ts
+++ b/packages/main/src/plugin/appearance-init.ts
@@ -50,8 +50,8 @@ export class AppearanceInit {
         [`${AppearanceSettings.SectionName}.${AppearanceSettings.NavigationAppearance}`]: {
           description: 'Select icon and title or just icon for navigation icons',
           type: 'string',
-          enum: ['icon and title', 'icon'],
-          default: 'icon and title',
+          enum: ['icon + title', 'icon'],
+          default: import.meta.env.DEV ? 'icon + title' : 'icon',
         },
       },
     };
--- a/packages/main/src/plugin/appearance-settings.ts
+++ b/packages/main/src/plugin/appearance-settings.ts
@@ -23,4 +23,5 @@ export enum AppearanceSettings {
   DarkEnumValue = 'dark',
   SystemEnumValue = 'system',
   ZoomLevel = 'zoomLevel',
+  NavigationAppearance = 'navigationAppearance',
 }
--- a/packages/main/src/plugin/appearance-settings.ts
+++ b/packages/main/src/plugin/appearance-settings.ts
@@ -23,5 +23,5 @@ export enum AppearanceSettings {
   DarkEnumValue = 'dark',
   SystemEnumValue = 'system',
   ZoomLevel = 'zoomLevel',
-  NavigationAppearance = 'navigationAppearance',
+  NavigationAppearance = 'navigationBarLayout',
 }
--- a/packages/main/src/plugin/appearance-settings.ts
+++ b/packages/main/src/plugin/appearance-settings.ts
@@ -24,4 +24,6 @@ export enum AppearanceSettings {
   SystemEnumValue = 'system',
   ZoomLevel = 'zoomLevel',
   NavigationAppearance = 'navigationBarLayout',
+  Icon = 'icon',
+  IconAndTitle = 'icon + title',
 }
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -73,7 +73,7 @@ function onDidChangeConfigurationCallback(e: Event): void {
           <NewContentOnDashboardBadge />
         </div>
         {#if iconWithTitle}
-          <div class="text-xs text-center ml-[2px]" title="Dashboard">
+          <div class="text-xs text-center ml-[2px]" aria-label="Dashboard title">
             Dashboard
           </div>
         {/if}
@@ -99,7 +99,7 @@ function onDidChangeConfigurationCallback(e: Event): void {
         <div class="flex flex-col items-center w-full h-full">
           <AccountIcon size={iconSize} />
           {#if iconWithTitle}
-            <div class="text-xs text-center ml-[2px]" title="Accounts">
+            <div class="text-xs text-center ml-[2px]" aria-label="Accounts title">
               Accounts
             </div>
           {/if}
@@ -112,7 +112,7 @@ function onDidChangeConfigurationCallback(e: Event): void {
   <NavItem href="/preferences" tooltip="Settings" bind:meta={meta} onClick={handleClick}>
     <SettingsIcon size={iconSize} />
     {#if iconWithTitle}
-      <div class="text-xs text-center ml-[2px]" title="Settings">
+      <div class="text-xs text-center ml-[2px]" aria-label="Settings title">
         Settings
       </div>
     {/if}
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -31,16 +31,12 @@ const NAV_BAR_LAYOUT = `${AppearanceSettings.SectionName}.${AppearanceSettings.N
 
 onDidChangeConfiguration.addEventListener(NAV_BAR_LAYOUT, onDidChangeConfigurationCallback);
 
-let minNavbarWidth = $state('min-w-leftnavbar');
-
-$effect(() => {
-  minNavbarWidth = iconWithTitle ? 'min-w-fit' : 'min-w-leftnavbar';
-});
+let minNavbarWidth = $derived(iconWithTitle ? 'min-w-fit' : 'min-w-leftnavbar');
 
 onMount(async () => {
   const commandRegistry = new CommandRegistry();
   commandRegistry.init();
-  iconWithTitle = (await window.getConfigurationValue(NAV_BAR_LAYOUT)) === 'icon + title';
+  iconWithTitle = (await window.getConfigurationValue(NAV_BAR_LAYOUT)) === AppearanceSettings.IconAndTitle;
 });
 
 onDestroy(() => {
@@ -59,7 +55,7 @@ function onDidChangeConfigurationCallback(e: Event): void {
   if ('detail' in e) {
     const detail = e.detail as { key: string; value: string };
     if (NAV_BAR_LAYOUT === detail?.key) {
-      iconWithTitle = detail.value === 'icon + title';
+      iconWithTitle = detail.value === AppearanceSettings.IconAndTitle;
     }
   }
 }
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -67,14 +67,16 @@ function onDidChangeConfigurationCallback(e: Event): void {
   <NavItem href="/" tooltip="Dashboard" bind:meta={meta} bind:iconWithTitle={iconWithTitle}>
     <div class="relative w-full">
       <div class="flex flex-col items-center w-full h-full">
-        <DashboardIcon size={iconSize} />
+        <div class="flex items-center w-fit h-full relative">
+          <DashboardIcon size={iconSize} />
+          <NewContentOnDashboardBadge />
+        </div>
         {#if iconWithTitle}
           <div class="text-xs text-center">
             Dashboard
           </div>
         {/if}
       </div>
-      <NewContentOnDashboardBadge />
     </div>
   </NavItem>
   {#each $navigationRegistry as navigationRegistryItem}
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -63,7 +63,7 @@ function onDidChangeConfigurationCallback(e: Event): void {
 <nav
   class="group w-fit min-w-leftnavbar max-w-[65px] flex flex-col hover:overflow-y-none bg-[var(--pd-global-nav-bg)] border-[var(--pd-global-nav-bg-border)] border-r-[1px]"
   aria-label="AppNavigation">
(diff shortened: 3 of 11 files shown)


packages/main/src/plugin/appearance-init.ts
```diff
@@ -47,6 +47,12 @@ export class AppearanceInit {
           default: 0,
           step: 0.1,
         },
+        [`${AppearanceSettings.SectionName}.${AppearanceSettings.NavigationAppearance}`]: {
+          description: 'Select icon and title or just icon for navigation icons',
+          type: 'string',
+          enum: ['icon and title', 'icon'],
```
benoitf: question: wondering if we do 'icon and title' or just 'icon+title'

packages/main/src/plugin/appearance-settings.ts
```diff
@@ -23,4 +23,5 @@ export enum AppearanceSettings {
   DarkEnumValue = 'dark',
   SystemEnumValue = 'system',
   ZoomLevel = 'zoomLevel',
+  NavigationAppearance = 'navigationAppearance',
```
benoitf: suggestion of the value: navigationBarLayout ?

packages/renderer/src/AppNavigation.svelte
```diff
@@ -37,16 +46,30 @@ function handleClick(): void {
     handleNavigation({ page: NavigationPage.RESOURCES });
   }
 }
+
+function onDidChangeConfigurationCallback(e: Event): void {
+  if ('detail' in e) {
+    const detail = e.detail as { key: string; value: string };
+    if ('preferences.navigationAppearance' === detail?.key) {
```
benoitf: need to use constants from appearance-settings ?

packages/renderer/src/AppNavigation.svelte
```diff
@@ -37,16 +46,30 @@ function handleClick(): void {
...
+      iconWithTitle = detail.value === 'icon and title';
```
benoitf: issue: need a constant imported from the appearance-setting file

packages/renderer/src/lib/ui/NavItem.svelte
```diff
@@ -23,7 +24,7 @@ $: selected = meta.url === uri || (uri !== '/' && meta.url.startsWith(uri));
 
 const navItems: Writable<number> = getContext('nav-items');
 
-$: tooltipText = counter ? `${tooltip} (${counter})` : tooltip;
+$: tooltipText = counter ? (iconWithTitle ? `${counter}` : `${tooltip} (${counter})`) : tooltip;
```
benoitf: maybe we need to split into two lines as here it's starting to be complex to read

packages/renderer/src/lib/ui/NavRegistryEntry.svelte
```diff
@@ -8,11 +8,15 @@ import type { NavigationRegistryEntry } from '/@/stores/navigation/navigation-re
 
 import NavItem from './NavItem.svelte';
 
-let { entry, meta = $bindable() }: { entry: NavigationRegistryEntry; meta: TinroRouteMeta } = $props();
+let {
+  entry,
+  meta = $bindable(),
+  iconWithTitle = $bindable(),
+}: { entry: NavigationRegistryEntry; meta: TinroRouteMeta; iconWithTitle: boolean } = $props();
```
benoitf: suggestion: introduce Props interface few lines before

benoitf:
I agree that default should be with text for newcomers, but maybe, first we collect feedback having it as default = icon

and we change default to icon+text for development mode `if (import.meta.env.DEV) {` --> 'icon+text'

once we're ok with the new rendering with text we change the default from icon to icon and text

SoniaSandler:
@benoitf (option from the issue) should we add a way to change this setting in the navigation popup (is this the context menu)?

benoitf:
> @benoitf (option from the issue) should we add a way to change this setting in the navigation popup (is this the context menu)?

it is the context menu (right click on the left navbar).

On my end, I would not have it for now, can be a follow-up. Also from the issue it says it needs to be checked with the design (it's quite easy to add the option)

That said I'm not the owner of the epic https://github.com/containers/podman-desktop/issues/9191 so I can only give some hints.

benoitf:
issue(blocking): on my end, the text is not fitting
image

Dashboard: D is not complete (I don't see entirely the left D letter and it's inside the left marker of the selected item)

same for Kubernetes, Extensions


question (non-blocking) And I'm not sure to understand the meaning of the tooltip saying '3'.
image
@deboer-tim asked for that in the issue but as a user I would think the tooltip is somehow broken.

packages/renderer/src/AppNavigation.svelte
```diff
@@ -37,28 +48,45 @@ function handleClick(): void {
...
   }
 }
+
+function onDidChangeConfigurationCallback(e: Event): void {
+  if ('detail' in e) {
+    const detail = e.detail as { key: string; value: string };
+    if (NAV_BAR_LAYOUT === detail?.key) {
+      iconWithTitle = detail.value === 'icon + title';
+    }
+    console.log(detail.value);
```
feloy: can be removed

packages/renderer/src/AppNavigation.svelte
```diff
@@ -37,28 +48,45 @@ function handleClick(): void {
...
+  }
+}
 </script>
 
 <svelte:window />
 <nav
   class="group w-leftnavbar min-w-leftnavbar flex flex-col hover:overflow-y-none bg-[var(--pd-global-nav-bg)] border-[var(--pd-global-nav-bg-border)] border-r-[1px]"
   aria-label="AppNavigation">
-  <NavItem href="/" tooltip="Dashboard" bind:meta={meta}>
+  <NavItem href="/" tooltip="Dashboard" bind:meta={meta} bind:iconWithTitle={iconWithTitle}>
```
feloy: Is `bind:` necessary here? NavItem is not changing its value, a `iconWithTitle={iconWithTitle}` would be enough I think

packages/renderer/src/lib/webview/Webviews.svelte
```diff
@@ -8,14 +8,21 @@ import NavItem from '/@/lib/ui/NavItem.svelte';
 import { webviews } from '/@/stores/webviews';
 
 export let meta: TinroRouteMeta;
+export let iconWithTitle = true;
 </script>
 
 {#each $webviews as webview (webview.id)}
-  <NavItem href="/webviews/{webview.id}" bind:meta={meta} tooltip={webview.name}>
+{console.log(webview.name)}
```
feloy: can be removed

axel7083:
> and we change default to icon+text for development mode `if (import.meta.env.DEV) {` --> 'icon+text'

+1 for https://github.com/containers/podman-desktop/issues/9236

axel7083:
> Dashboard: D is not complete (I don't see entirely the left D letter and it's inside the left marker of the selected item)

Here is a suggestion, we may use text-[8px] instead of text-xs (10px). @ekidneyrh would it be acceptable to have text 8px instead of 10px ? Here is the comparison

 | 10px (text-xs) | 8px (custom) |
| --- | --- |
| image | image |

axel7083 (commented):
Love the work done 👍 and output !

ekidneyrh:
> > Dashboard: D is not complete (I don't see entirely the left D letter and it's inside the left marker of the selected item)
> 
> Here is a suggestion, we may use text-[8px] instead of text-xs (10px). @ekidneyrh would it be acceptable to have text 8px instead of 10px ? Here is the comparison
> 
> 10px (text-xs)	8px (custom)

I fear that 8px is too small for readability. We have this [discussion](https://github.com/containers/podman-desktop/pull/7995#issuecomment-2220444862) before about not going below 10px.

Could we add a small bit more padding on the nav bar when text is visible to allow the labels more room?

SoniaSandler:
> issue(blocking): on my end, the text is not fitting image

@benoitf I added this problem as a possible follow-up because I wasn't sure if the width of the nav bar should be modified in this PR or a separate one as it uses the set width from `tailwind.config.cjs`
And as for the tooltip, I could change it to be as it was before `containers (3)`, and when Tim comes back, we could open a follow up to adjust it if needed

benoitf:
IMHO if we need to change the width for that case, it should be a pre-req PR and not a follow-up as if we merge we should always have something at least readable.

benoitf:
@ekidneyrh suggested
> Could we add a small bit more padding on the nav bar when text is visible to allow the labels more room?

but I guess padding will have an issue if the width stay the same ? either we reduce the text size (seems a no) or we increase the width or we reduce the text of the title

SoniaSandler:
@benoitf @ekidneyrh  since only icons "need" less width, would it be better to have the width change for only icons vs icons+titles or just have the same (slightly bigger) width for both cases?

gastoner (commented):
When is selected title+icon and you select "kubernetes" or "settings" the navigation width is glitching. 


https://github.com/user-attachments/assets/e8479e2e-ccc9-4292-813e-921fb8d6450a


When I started PD I've had selected "icon"option (the second in the dropddown) - I would imagine that the default one should be the first one? 

Probably related to changing the "width" when is kubernetes/settings selected
image

axel7083 (approved):
LGTM 👍 

image

Follow-up PR, i'm wondering if we could add the option to show or hide the titles with a right click on the status bar

Something like the following in JetBrain IDE

image

We have existing logic for showing/hidding specific elements

gastoner (approved):

SoniaSandler:
> On my end, I would not have it for now, can be a follow-up. Also from the issue it says it needs to be checked with the design (it's quite easy to add the option)

@benoitf I've created a follow up issue for adding this option to the navbar context menu (https://github.com/podman-desktop/podman-desktop/issues/9885).
> question (non-blocking) And I'm not sure to understand the meaning of the tooltip saying '3'

Regarding the icon's tooltip, since it was requested both in the issue and the discussion in the epic to only have the counter in the tooltip when the title is show, is the current tooltip format okay?
This is something that can be easily changed, so we can try it out for now with only the counter and see if it works for us or if we'd prefer to have a different tooltip format when the title is visible

packages/main/src/plugin/appearance-init.ts
```diff
@@ -47,6 +47,12 @@ export class AppearanceInit {
           default: 0,
           step: 0.1,
         },
+        [`${AppearanceSettings.SectionName}.${AppearanceSettings.NavigationAppearance}`]: {
+          description: 'Select icon and title or just icon for navigation icons',
+          type: 'string',
+          enum: ['icon + title', 'icon'],
+          default: import.meta.env.DEV ? 'icon + title' : 'icon',
```
benoitf: nitpick: use constants for 'icon' and 'icon + title', so in case we change the enum, it'll automatically change the default value

packages/renderer/src/AppNavigation.svelte
```diff
@@ -16,18 +17,34 @@ import SettingsIcon from './lib/images/SettingsIcon.svelte';
...
+let minNavbarWidth = $state('min-w-leftnavbar');
+
+$effect(() => {
+  minNavbarWidth = iconWithTitle ? 'min-w-fit' : 'min-w-leftnavbar';
+});
 
 onMount(async () => {
   const commandRegistry = new CommandRegistry();
   commandRegistry.init();
+  iconWithTitle = (await window.getConfigurationValue(NAV_BAR_LAYOUT)) === 'icon + title';
```
benoitf: nitpick: reuse the constant there

packages/renderer/src/AppNavigation.svelte
```diff
@@ -37,28 +54,44 @@ function handleClick(): void {
     handleNavigation({ page: NavigationPage.RESOURCES });
   }
 }
+
+function onDidChangeConfigurationCallback(e: Event): void {
+  if ('detail' in e) {
+    const detail = e.detail as { key: string; value: string };
+    if (NAV_BAR_LAYOUT === detail?.key) {
+      iconWithTitle = detail.value === 'icon + title';
```
benoitf: nitpick: use the constant for 'icon + title'

packages/renderer/src/AppNavigation.svelte
```diff
@@ -16,18 +17,34 @@ import SettingsIcon from './lib/images/SettingsIcon.svelte';
...
 const iconSize = '22';
+const NAV_BAR_LAYOUT = `${AppearanceSettings.SectionName}.${AppearanceSettings.NavigationAppearance}`;
+
+onDidChangeConfiguration.addEventListener(NAV_BAR_LAYOUT, onDidChangeConfigurationCallback);
+
+let minNavbarWidth = $state('min-w-leftnavbar');
+
+$effect(() => {
+  minNavbarWidth = iconWithTitle ? 'min-w-fit' : 'min-w-leftnavbar';
+});
```
benoitf: nitpick: I would have just specified minNavbarWidth as a $derived state

```ts
minNavbarWidth = $derived(iconWithTitle ? 'min-w-fit' : 'min-w-leftnavbar');
```

packages/renderer/src/lib/webview/Webviews.svelte
```diff
@@ -8,14 +8,20 @@ import NavItem from '/@/lib/ui/NavItem.svelte';
...
 {#each $webviews as webview (webview.id)}
-  <NavItem href="/webviews/{webview.id}" bind:meta={meta} tooltip={webview.name}>
+  <NavItem href="/webviews/{webview.id}" bind:meta={meta} tooltip={webview.name} iconWithTitle={iconWithTitle}>
     {#if !webview.icon}
       <Fa icon={faPuzzlePiece} size="1.5x" />
     {:else}
       <IconImage image={webview?.icon} alt={webview.name} class="max-h-6 max-w-[24px]" />
     {/if}
+    {#if iconWithTitle}
+      <div class="text-xs text-center" id="iconTitle">
```
benoitf: issue: remove the id

if for tests, add an aria-label and role

packages/renderer/src/lib/ui/NavRegistryEntry.svelte
```diff
@@ -23,5 +29,10 @@ let { entry, meta = $bindable() }: { entry: NavigationRegistryEntry; meta: Tinro
     {:else if entry.icon.iconImage && typeof entry.icon.iconImage === 'string'}
       <img src={entry.icon.iconImage} width="22" height="22" alt={entry.name} />
     {/if}
+    {#if iconWithTitle && entry.icon}
+      <div class="text-xs text-center max-w-[60px] ml-[2px]" title="Icon title">
```
benoitf: question: what means 'Icon title' ? I mean either we need to be specific with the entry.name or no title

packages/renderer/src/lib/webview/Webviews.spec.ts
```diff
@@ -60,7 +60,7 @@ test('check we have list of webviews in navigation bar', async () => {
 
   webviews.set(webviewTestList);
   const fakeMeta = { url: '/webviews' } as unknown as TinroRouteMeta;
-  render(Webviews, { meta: fakeMeta });
+  render(Webviews, { meta: fakeMeta, iconWithTitle: false });
 
   const links = screen.getAllByRole('link');
```
benoitf, line 65: issue: it looks like there is no test checking iconWithTitle behavior (that the name is displayed)

packages/renderer/src/AppNavigation.svelte
```diff
@@ -2,11 +2,12 @@
 
 <script lang="ts">
 import { Tooltip } from '@podman-desktop/ui-svelte';
```
benoitf, line 4: issue: I see no new test in AppNavigation.spec.ts for checking the new behaviour

benoitf (commented):
> Regarding the icon's tooltip, since it was requested both in the issue and the discussion in the epic to only have the counter in the tooltip when the title is show, is the current tooltip format okay?

yes it was a non blocking question

benoitf (approved):
thx for the changes

PR #9782, opened 2024-11-06
feat: add error visual indicator in typeahead component

### LINKED ISSUE #8926: pulling/searching image: notify if image is not recognized

following https://github.com/containers/podman-desktop/issues/971 if I copy/paste an image name in the field I have not the typeahead indication

I would like to have a visual indication that the image is recognized (or not recognized)
like changing the color if not found or changing the color if found

any suggestion that could help the user to have a small indicator

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds error classes and adds new property

Part of #8926 
1/2 PR's

### DIFF
--- a/packages/renderer/src/lib/ui/Typeahead.spec.ts
+++ b/packages/renderer/src/lib/ui/Typeahead.spec.ts
@@ -196,3 +196,38 @@ test('should navigate in list with keys', async () => {
   // copies the item in the input
   screen.getByDisplayValue('term03');
 });
+
+test('should show error border', async () => {
+  const searchFunction = async () => {
+    const result: string[] = ['term0', 'term1', 'term2', 'unknown'];
+    return result;
+  };
+  render(Typeahead, {
+    initialFocus: true,
+    searchFunction,
+    delay: 10,
+  });
+  const input = screen.getByRole('textbox');
+  await userEvent.type(input, 'term');
+
+  await new Promise(resolve => setTimeout(resolve, 11));
+
+  const list = screen.getByRole('row');
+  const items = within(list).getAllByRole('button');
+  expect(items.length).toBe(4);
+
+  // No item is selected first
+  assertItemSelected(items, -1);
+
+  const cellOutsideInput = screen.getByRole('cell');
+  expect(cellOutsideInput).toHaveClass('border-b-[var(--pd-input-field-stroke-error)]');
+  expect(cellOutsideInput).toHaveClass('focus-within:border-[var(--pd-input-field-stroke-error)]');
+  expect(cellOutsideInput).not.toHaveClass('hover:border-b-[var(--pd-input-field-hover-stroke)]');
+
+  // to finish the fisrt term => term0
+  await userEvent.type(input, '0');
+  await new Promise(resolve => setTimeout(resolve, 11));
+  expect(cellOutsideInput).not.toHaveClass('border-b-[var(--pd-input-field-stroke-error)]');
+  expect(cellOutsideInput).not.toHaveClass('focus-within:border-[var(--pd-input-field-stroke-error)]');
+  expect(cellOutsideInput).toHaveClass('hover:border-b-[var(--pd-input-field-hover-stroke)]');
+});
--- a/packages/renderer/src/lib/ui/Typeahead.spec.ts
+++ b/packages/renderer/src/lib/ui/Typeahead.spec.ts
@@ -20,7 +20,7 @@ import '@testing-library/jest-dom/vitest';
 import { render, screen, within } from '@testing-library/svelte';
 import userEvent from '@testing-library/user-event';
 import { tick } from 'svelte';
-import { expect, test } from 'vitest';
+import { beforeEach, expect, test, vi } from 'vitest';
 
 import Typeahead from './Typeahead.svelte';
 
@@ -46,6 +46,11 @@ function assertItemSelected(items: HTMLElement[], r: number): void {
     }
   }
 }
+
+beforeEach(() => {
+  vi.resetAllMocks();
+  vi.restoreAllMocks();
+});
 
 test('a textbox is created', async () => {
   render(Typeahead);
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -11,6 +11,7 @@ export let disabled: boolean = false;
 export let initialFocus: boolean = false;
 export let id: string | undefined = undefined;
 export let name: string | undefined = undefined;
+export let isShortName: boolean = false;
```
axel7083: I do not think the TypeaHead component should be responsible of validating the input.

To avoid adding more complexity to the component; the parent should be responsible to define the state (error, okey) of the component. Here is an example https://flowbite-svelte.com/docs/forms/input-field#Validation

benoitf: I confirm here that Typeahead is agnostic as how it's used (so short name is not something valid for the component)

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -11,6 +11,7 @@ export let disabled: boolean = false;
 export let initialFocus: boolean = false;
 export let id: string | undefined = undefined;
 export let name: string | undefined = undefined;
+export let isRecognized: boolean = true;
```
axel7083: `isRecognized` feel very odd to me, if we want to show red, let's call it error: boolean

axel7083:
The PR title is misleading, could you rename it ? We are not changing any notification related stuff, probably something like `feat: add error visual indicator pull image input`

axel7083:
We have some inconsistencies with the shortnames

image

image

By typing and removing sometime I am getting red or purple for `hello-world`

axel7083 (approved):
LGTM

feloy:
There are some circumstances where, by design, the images are not searchable, for example for private images.

In this case (for me, with the image `quay.io/phmartin/private` which is private, and with credentials for quay.io/phmaritn added to the Podman Desktop registries), the image is not searchable (by design), but it is still possible to pull it. The red line will be a false negative in this case.

benoitf (commented):
first part about introducing a new field for TypeAhead should be another PR (and I think it would be quickly merged)

then, for the second part, it can't rely on the result of the search query. We need to check querying the registry for a specific image name.

third: we could have enhancement saying that the image is not matching the current architecture (so not an error but a warning)

gastoner:
@axel7083 can you please re-check? I've split this PR into two smaller ones

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -181,17 +182,19 @@ function onWindowClick(e: Event): void {
 </script>
 
 <svelte:window on:click={onWindowClick} />
-
 <div
+  role='cell'
```
axel7083: Not sure about the `cell` role, see documentation

> The cell value of the ARIA role attribute identifies an element as being a cell in a tabular container that does not contain column or row header information. To be supported, the cell must be nested in an element with the role of row.[^1]

[^1]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/cell_role

packages/renderer/src/lib/ui/Typeahead.spec.ts
```diff
@@ -196,3 +200,29 @@ test('should navigate in list with keys', async () => {
...
+  render(Typeahead, {
+    initialFocus: true,
+    error: false,
+    delay: 10,
+  });
+
+  const cellOutsideInput = screen.getByRole('cell');
+  expect(cellOutsideInput).not.toHaveClass('border-b-[var(--pd-input-field-stroke-error)]');
+  expect(cellOutsideInput).not.toHaveClass('focus-within:border-[var(--pd-input-field-stroke-error)]');
+  expect(cellOutsideInput).toHaveClass('hover:border-b-[var(--pd-input-field-hover-stroke)]');
```
benoitf: you can check the parent element of the textbox

```suggestion
  const cellOutsideInput = screen.getByRole('textbox');
  const parentInput = cellOutsideInput.parentElement;
  expect(parentInput).not.toHaveClass('border-b-[var(--pd-input-field-stroke-error)]');
  expect(parentInput).not.toHaveClass('focus-within:border-[var(--pd-input-field-stroke-error)]');
  expect(parentInput).toHaveClass('hover:border-b-[var(--pd-input-field-hover-stroke)]');
```

packages/renderer/src/lib/ui/Typeahead.spec.ts
```diff
@@ -196,3 +200,29 @@ test('should navigate in list with keys', async () => {
...
+  render(Typeahead, {
+    initialFocus: true,
+    error: true,
+    delay: 10,
+  });
+
+  const cellOutsideInput = screen.getByRole('cell');
+  expect(cellOutsideInput).toHaveClass('border-b-[var(--pd-input-field-stroke-error)]');
+  expect(cellOutsideInput).toHaveClass('focus-within:border-[var(--pd-input-field-stroke-error)]');
+  expect(cellOutsideInput).not.toHaveClass('hover:border-b-[var(--pd-input-field-hover-stroke)]');
```
benoitf:
```suggestion
  const cellOutsideInput = screen.getByRole('textbox');
  const parentInput = cellOutsideInput.parentElement;
  expect(parentInput).toHaveClass('border-b-[var(--pd-input-field-stroke-error)]');
  expect(parentInput).toHaveClass('focus-within:border-[var(--pd-input-field-stroke-error)]');
  expect(parentInput).not.toHaveClass('hover:border-b-[var(--pd-input-field-hover-stroke)]');
```

packages/renderer/src/lib/ui/Typeahead.svelte
benoitf:
```suggestion
```

use the parentElement property of the `textbox`

benoitf (approved):

axel7083 (approved):
LGTM

cdrage (approved):

PR #9783, opened 2024-11-06
chore: adding id property to forward config

axel7083:
### What does this PR do?

First step of https://github.com/containers/podman-desktop/issues/9637, the end goal is to avoid passing the full object between the frontend and the backend to trigger action on it.

Currently it is complex to manipulate individual port forwarding, https://github.com/containers/podman-desktop/pull/9592 was made as a temporary solution. 

This PR adds a unique identifier (unused) to the `ForwardConfig` interface. This will allow in a follow-up PR to simplify the IPC  `kubernetes-client:deletePortForward` and the `KubernetesPortForwardService#deleteForward` implementation.

> ℹ️ I choose `uuid` because I find it easy to work with, but I know we have some places in podman desktop where we use a counter that we increment (we also use uuid for webview management). I am open to change if there is concern about uuids

### Screenshot / video of UI

N/A

Part of https://github.com/containers/podman-desktop/issues/9637

### How to test this PR?

### DIFF
--- a/packages/api/src/kubernetes-port-forward-model.ts
+++ b/packages/api/src/kubernetes-port-forward-model.ts
@@ -75,7 +75,7 @@ export interface ForwardConfig {
   /**
    * Identifier of the forward config
    */
-  uuid: string;
+  id: string;
   /**
    * The name of the resource.
    */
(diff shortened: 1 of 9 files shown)


benoitf (approved):
question: should it be just `id` (I would say how it's computed/generated is internal)
(it could be just an incremental counter)

axel7083:
> question: should it be just `id` (I would say how it's computed/generated is internal) (it could be just an incremental counter)

It could 👍 , I do not have any opinion on that matter.

PR #9787, opened 2024-11-06
refactor(kubernetes): simplify the forward config interface

axel7083:
### What does this PR do?

This PR simplify the `ForwardConfig` interface. The change update the `forwards` property as followed

```diff
- forwards: PortMapping[],
+ forward: PortMapping,
```

The idea is that a ForwardConfig should only manage a single port mapping, as it is very difficult to manipulate (create, delete) multiple port forwarding with a single object.

Direct consequence is simplifying the `kubernetes-client:deletePortForward` IPC and the subsequence code implementation that was introduced in https://github.com/containers/podman-desktop/pull/9592

### Screenshot / video of UI

N/A

- Following https://github.com/containers/podman-desktop/pull/9783
- Part of https://github.com/containers/podman-desktop/issues/9637

All steps can be found in https://github.com/containers/podman-desktop/issues/9637#issuecomment-2459684176

### How to test this PR?

You can check that there is no regression by creating a pod and a service, starting multiple port forward **for a single k8s resource**

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: static-web
  namespace: default
  labels:
    role: myrole
spec:
  containers:
    - name: web
      image: nginx
      ports:
        - name: web
          containerPort: 80
          protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: default
spec:
  ports:
    - name: demo
      protocol: TCP
      port: 8080
      targetPort: 80
    - name: dummy
      protocol: TCP
      port: 8081
      targetPort: 80
  selector:
    role: myrole
```

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
@@ -27,12 +27,7 @@ import { ForwardConfigRequirements } from '/@/plugin/kubernetes/kubernetes-port-
 import type { IDisposable } from '/@/plugin/types/disposable.js';
 import { Disposable } from '/@/plugin/types/disposable.js';
 import { isFreePort } from '/@/plugin/util/port.js';
-import type {
-  ForwardConfig,
-  ForwardOptions,
-  PortMapping,
-  UserForwardConfig,
-} from '/@api/kubernetes-port-forward-model.js';
+import type { ForwardConfig, ForwardOptions, UserForwardConfig } from '/@api/kubernetes-port-forward-model.js';
 
 /**
  * Service provider for Kubernetes port forwarding.
@@ -58,7 +53,7 @@ export class KubernetesPortForwardServiceProvider {
 
   /**
    * Gets the configuration key for the given Kubernetes configuration.
-   * The key is used for caching services and also for storing information about created forwards linked to each service.
+   * The key is used for caching services and also for storing information about created forward linked to each service.
    *
    * @param kubeConfig - The Kubernetes configuration.
    * @returns The configuration key.
@@ -86,7 +81,7 @@ export class KubernetesPortForwardService implements IDisposable {
     private forwardingConnectionService: PortForwardConnectionService,
     private apiSender: ApiSenderType,
   ) {
-    this.apiSender.send('kubernetes-port-forwards-update', []);
+    this.apiSender.send('kubernetes-port-forward-update', []);
   }
 
   dispose(): void {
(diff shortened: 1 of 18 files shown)


packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
```diff
@@ -95,47 +90,28 @@ export class KubernetesPortForwardService implements IDisposable {
...
+   * Return the identifier of the ForwardConfig
    * @param config
-   * @param mapping
    * @private
    */
-  private getPortForwardKey(config: ForwardConfig, mapping: PortMapping): string {
-    return `${config.namespace}/${config.name}/${config.kind}/${mapping.remotePort}`;
+  private getPortForwardKey(config: ForwardConfig): string {
+    return config.id;
   }
```
benoitf: question: do we need this private method anymore if we just need to use `.id` on the object ?

packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
```diff
@@ -144,36 +120,17 @@ export class KubernetesPortForwardService implements IDisposable {
...
-
-    if (newConfig.forwards.length === 0) {
-      await this.configManagementService.deleteForward(config);
-    } else {
-      await this.configManagementService.updateForward(config, newConfig);
-    }
+  async deleteForward(config: UserForwardConfig): Promise<void> {
+    const key = this.getPortForwardKey(config);
+
+    const disposable = this.#disposables.get(key);
```
benoitf: suggestion: remove key and getPortForwardKey method

```suggestion
    const disposable = this.#disposables.get(config.id);
```

packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
```diff
@@ -194,17 +151,15 @@ export class KubernetesPortForwardService implements IDisposable {
...
-        this.#disposables.set(key, disposable);
-      }
-    }
-
-    return Disposable.from(...disposables);
+    const key = this.getPortForwardKey(config);
+    if (this.#disposables.has(key)) throw new Error('forward already started');
+
+    const disposable = await this.forwardingConnectionService.startForward(config);
+    this.#disposables.set(key, disposable);
```
benoitf: same remark there, use `config.id`

benoitf (approved):

packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
axel7083: I wanted to _reduce_ the number of change, but yes this method will disappear

axel7083: I will update the PR and remove the `getPortForwardKey`

packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
axel7083: 👍

packages/main/src/plugin/kubernetes/kubernetes-port-forward-service.ts
axel7083: 👍

feloy (approved):
tested with several forwards on the same service, works as expected

PR #9788, opened 2024-11-06
fix: remove nullable detailsPage binding

### LINKED ISSUE #9717: Deleting Kind resource from Resource Page breaks Podman Desktop navigation, hard kill and restart required
### Bug description

User is unable to navigate Podman Desktop unless it's restarted

### Operating system

Windows 10 22H2 WSL

### Installation Method

### Version

1.14.0

### Steps to reproduce

* Create a Kind cluster
* After creation go back to Resources page
* Open Kind Cluster resource details
* Stop the cluster
* Delete the cluster
* Attempt to navigate anywhere

### Relevant log output

N/A

### Additional context

https://github.com/user-attachments/assets/292e75f6-3522-4173-91fa-79d28f1f530a

axel7083:
### What does this PR do?

Removing the nullable detailsPage binding, and using the handleNavigation to redirect to the resource page when the connection( either container or Kubernetes ) is removed.

The previous behaviour was trying to reach the last page through the close function of the detailsPage implementation, but two issue
1. the binding can be null, therefore no redirect + error
2. if we switch from summary to logs, the last page is summary ? (not sure, but safer to just redirect to resources and avoid any issue)

### Screenshot / video of UI

https://github.com/user-attachments/assets/1ffa3a3e-b9d9-4fe7-8419-4257b6b14e95

Fixes https://github.com/containers/podman-desktop/issues/9717

### How to test this PR?

You can reproduce steps in https://github.com/containers/podman-desktop/issues/9717 to ensure the problem is fixed, or the video above


odockal (approved):
LGTM
Tried on Linux with Kind and it works well now. No broken app after deleting the kind cluster from the details page. I also tested minikube.

benoitf:
> if we switch from summary to logs, the last page is summary ? (not sure, but safer to just redirect to resources and avoid any issue)

AFAIK here, @deboer-tim introduced this thing and I think that if you stay on the same page (logs or summay being part of the same "page" then lastPage is not changed)

I would be more interested to know why the bindable is null rather than changing the whole thing as AFAIK detailsPage.close() won't work (you're just avoiding to call it)

axel7083:
> I would be more interested to know why the bindable is null rather than changing the whole thing as AFAIK detailsPage.close() won't work

I explained in details why the binding is null here https://github.com/podman-desktop/podman-desktop/issues/9717#issuecomment-2459812614

>  (you're just avoiding to call it)

Moreover, I would argue, the parent component should usually avoid to try to call its children, if we want to keep the lastPage we could simply use the router.goto in the parent component, instead of calling a function in the child, but going back to the resources page ensure we do not go back to a page that could have inconstency (let's say I am on container details page, am redirected to engine details, delete it, lastPage is a container that do not exist anymore, could be problematic, I agree we should handle those cases, but we can reduce risks by going to the resources page.)

deboer-tim:
FWIW the renderer DetailsPage closes using the breadcrumb because we have several cases where you can reach the same page from multiple places in the UI, e.g. you can get to Container Details from the Containers page or Pod Details > Summary, and Image Details from Container Details > Summary. In those cases we want to return to the previous page, not the 'natural parent'. navigationHints in App.svelte provides metadata so that switching between tabs is ignored / doesn't change what the last page is.

In both of the cases listed above it's going back to a list instead of the details page you came from, but that looks like an unrelated regression...

If we know that this page will never be accessible from anywhere other than Resources then the fix is fine (just hard-coding the previous page), otherwise I'd say we could move the if statement here down a line (within the DetailsPage) to avoid the 'blank' page that triggers this.

https://github.com/containers/podman-desktop/blob/9e8aad114dde4ad02cb012cd144a4b2848fae247/packages/renderer/src/lib/preferences/PreferencesContainerConnectionRendering.svelte#L139-L140

axel7083:
I am okey with either hardcoding the router.goto('/lastpage') or keeping this has it is

axel7083:
After trying to move down the `  {#if connectionInfo}` I am getting severals errors that need way more works that the fix I was initially provided.

`<svelte:fragment slot="*">` cannot be conditionally placed or not, the following

```svelte
<DetailsPage title={connectionInfo?.name ?? '<unknown>'} bind:this={detailsPage}>
  {#if connectionInfo}
  <svelte:fragment slot="subtitle">
    <div class="flex flex-row">
      <ConnectionStatus status={connectionInfo.status} />
      <ConnectionErrorInfoButton status={connectionStatus} />
    </div>
  </svelte:fragment>
``` 

is invalid because `Svelte: <svelte:fragment> must be the direct child of a component`

If I want to move down the `{#if connectionInfo}` I need to put inside each <svelte:fragment....
Morever I have to provide a default title for the DetailsPage since title need to be a string, and connectionInfo?.name can be undefined.

odockal:
@benoitf @deboer-tim @axel7083 This issue is one on the list for 1.14.2 bug fix release. Is it possible to review it so we can backport it, please?

gastoner (approved):
LGTM, tested on Windows 🚀

deboer-tim:
Thank you for checking the back link option Axel, unfortunate that's not an option. Just dropping a note that I'm fine with merging and there are two +1s.

gastoner:
@mergifyio backport 1.14.x

PR #9791, opened 2024-11-06
fix: status bar progress bar percentage display

### LINKED ISSUE #9418: Tasks on status bar should show progress

The new task progress on the status bar is awesome (thanks @axel7083!) but it doesn't show progress for long running tasks.

https://github.com/user-attachments/assets/24aa19a7-349d-46d7-87d7-a797488d8d6a

Tasks that have incremental progress (e.g. 7%, 10%, 25%...) show up correctly in the task manager, but on the status bar they are indeterminate/animated. The incremental progress should show up correctly in the status bar.

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Fix the progress status bar not showing the progress value. There is also a missalignement when we show the percentage, that can be fixed by providing an align class to the `ProgressBar.svelte`

### Screenshot / video of UI

image

Fixes https://github.com/containers/podman-desktop/issues/9418

### How to test this PR?

#### Testing manually

You may add the following in an extension activate function

```ts
  extensionApi.window.withProgress(
    { location: extensionApi.ProgressLocation.TASK_WIDGET, title: `Fake loading` },
    async (progress) => {
      progress.report({ increment: 20 });

      return new Promise((resolve, reject) => {
        setTimeout(() => {
          resolve();
        }, 60_000);
      });
    },
  )
```

### DIFF
--- a/packages/renderer/src/lib/task-manager/ProgressBar.svelte
+++ b/packages/renderer/src/lib/task-manager/ProgressBar.svelte
@@ -31,8 +31,8 @@
 <script lang="ts">
 interface Props {
   progress?: number;
-  width: string;
-  height: string;
+  width?: string;
+  height?: string;
   class?: string;
 }
 
--- a/packages/renderer/src/lib/task-manager/ProgressBar.svelte
+++ b/packages/renderer/src/lib/task-manager/ProgressBar.svelte
@@ -35,7 +35,6 @@ interface Props extends HTMLAttributes<HTMLElement> {
   progress?: number;
   width?: string;
   height?: string;
-  class?: string;
 }
 
 let { progress, width = 'w-36', height = 'h-4', class: className, ...restProps }: Props = $props();
--- a/packages/renderer/src/lib/task-manager/ProgressBar.svelte
+++ b/packages/renderer/src/lib/task-manager/ProgressBar.svelte
@@ -29,17 +29,19 @@
 </style>
 
 <script lang="ts">
-interface Props {
+import type { HTMLAttributes } from 'svelte/elements';
+
+interface Props extends HTMLAttributes<HTMLElement> {
   progress?: number;
   width?: string;
   height?: string;
   class?: string;
 }
 
-let { class: className, progress, width = 'w-36', height = 'h-4' }: Props = $props();
+let { progress, width = 'w-36', height = 'h-4', class: className, ...restProps }: Props = $props();
 </script>
 
-<div class="flex flex-row {className}">
+<div class="flex flex-row {className}" {...restProps} >
   <div class="{width} {height} rounded-full bg-[var(--pd-progressBar-bg)] progress-bar overflow-hidden">
     {#if progress !== undefined}
       <div
--- a/packages/renderer/src/lib/task-manager/ProgressBar.svelte
+++ b/packages/renderer/src/lib/task-manager/ProgressBar.svelte
@@ -29,12 +29,17 @@
 </style>
 
 <script lang="ts">
-export let progress: number | undefined;
-export let width: string = 'w-36';
-export let height: string = 'h-4';
+interface Props {
+  progress?: number;
+  width: string;
+  height: string;
+  class?: string;
+}
+
+let { class: className, progress, width = 'w-36', height = 'h-4' }: Props = $props();
 </script>
 
-<div class="flex flex-row">
+<div class="flex flex-row {className}">
   <div class="{width} {height} rounded-full bg-[var(--pd-progressBar-bg)] progress-bar overflow-hidden">
     {#if progress !== undefined}
       <div
--- a/packages/renderer/src/lib/statusbar/TaskIndicator.spec.ts
+++ b/packages/renderer/src/lib/statusbar/TaskIndicator.spec.ts
@@ -98,3 +98,20 @@ test('multiple tasks running should display them', async () => {
   expect(status).toBeDefined();
   expect(status.textContent).toBe('2 tasks running');
 });
+
+test('task with defined progress value should display it', async () => {
+  tasksInfo.set([
+    {
+      name: 'Dummy Task',
+      state: 'running',
+      status: 'in-progress',
+      started: 0,
+      id: 'dummy-task',
+      progress: 50,
+    },
+  ]);
+
+  const { getByText } = render(TaskIndicator);
+  const span = getByText('50%');
+  expect(span).toBeDefined();
+});
--- a/packages/renderer/src/lib/statusbar/TaskIndicator.svelte
+++ b/packages/renderer/src/lib/statusbar/TaskIndicator.svelte
@@ -13,7 +13,7 @@ let title: string | undefined = $derived.by(() => {
 });
 
 let progress: number | undefined = $derived.by(() => {
-  if (runningTasks.length !== 0) return undefined;
+  if (runningTasks.length === 0) return undefined;
   return runningTasks[0].progress ?? 0;
 });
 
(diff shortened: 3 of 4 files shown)


packages/renderer/src/lib/task-manager/ProgressBar.svelte
```diff
@@ -29,12 +29,17 @@
...
 
 <script lang="ts">
-export let progress: number | undefined;
-export let width: string = 'w-36';
-export let height: string = 'h-4';
+interface Props {
+  progress?: number;
+  width?: string;
+  height?: string;
+  class?: string;
```
benoitf: question: usually we let any other props being passed, we don't have a field for class

axel7083: then how do get the `class` prop in svelte 5 :thinking:  ? I took the solution from https://www.reddit.com/r/sveltejs/comments/1dodc08/svelte_5_class_prop/?rdt=40307

benoitf: maybe something like spread operator
```
let { progress, width = 'w-36', height = 'h-4', ...restProps }: Props = $props();
```

and then `{props.class}`

I suspect maybe it'll be missing the type so either
```
let { progress, width = 'w-36', height = 'h-4', ...restProps }: Props & {class? : string} = $props();
```

or allowing more attributes that we can set on divs:
```
let { progress, width = 'w-36', height = 'h-4', ...restProps }: Props & Record<string, unknown> = $props();
```

and using `{...restProps}` but in your class as you want to concatenate the value of class

`<div class="flex flex-row {restProps.class}">`

axel7083: Why would we want to allow untyped element to be added ? A component is suppose to have a determined number of props ? I am not really seeing the purpose to add `unknown` props ?

If we need to have more props that we want to propagate we should have them defined, and properly typed ?

benoitf: you can use & HTMLAttributes and {...restProps} meaning all attributes you can set on an html element can be propagated

axel7083: This is a bit complicated, as HTMLAttributes is a generic type. See definition bellow (from 'svelte/elements')

```ts
export interface HTMLAttributes<T extends EventTarget> extends AriaAttributes, DOMAttributes<T> {
```

We could try to work around by using the `HTMLElement` object, since `HTMLElement extend Node` and `Node extend EventTarget` and do something like

```ts
import type { HTMLAttributes } from 'svelte/elements';

interface Props extends HTMLAttributes<HTMLElement> {
     // my-custom component props
}

let { trackingId, ...restProps }: Props = $props();
```

restProps now can contains all HTML basic attributes such as `aria-*`, `role`, `title` etc.

This is the most typed and html friendly I can do, what do you think ?

benoitf: go for it :-)

benoitf (approved):

packages/renderer/src/lib/task-manager/ProgressBar.svelte
```diff
@@ -29,12 +29,19 @@
...
-export let progress: number | undefined;
-export let width: string = 'w-36';
-export let height: string = 'h-4';
+import type { HTMLAttributes } from 'svelte/elements';
+
+interface Props extends HTMLAttributes<HTMLElement> {
+  progress?: number;
+  width?: string;
+  height?: string;
+  class?: string;
```
benoitf:
```suggestion
```

packages/renderer/src/lib/task-manager/ProgressBar.svelte
```diff
@@ -29,12 +29,19 @@
...
+}
+
+let { progress, width = 'w-36', height = 'h-4', class: className, ...restProps }: Props = $props();
```
benoitf, line 40:
```suggestion
let { progress, width = 'w-36', height = 'h-4', ...restProps }: Props = $props();
```

packages/renderer/src/lib/task-manager/ProgressBar.svelte
axel7083: Yes, just did remove it :) !

packages/renderer/src/lib/task-manager/ProgressBar.svelte
```diff
@@ -29,12 +29,19 @@
...
 </script>
 
-<div class="flex flex-row">
+<div class="flex flex-row {className}" {...restProps} >
```
benoitf, line 43:
```suggestion
<div  {...restProps} class="flex flex-row {restProps.class}">
```

axel7083: This would duplicate the class attribute, leading to one being overwritten ?

benoitf: I'm wondering if it should not be {...restProps} first as I don't know it'll handle two `class` attribute (last win or first win ?)

axel7083: first win, but here I do not want to overwrite the first one, I want to combine them

benoitf: AFAIK it's last win, so in this order `flex flex-row` will disappear

benoitf: > This would duplicate the class attribute, leading to one being overwritten ?

I think it'll keep only the last one

axel7083: https://github.com/user-attachments/assets/2a8f6a6e-0d3e-4901-be84-bf69e94ba4ed

packages/renderer/src/lib/task-manager/ProgressBar.svelte
axel7083: I would keep class outside `restProps` so we do not overwrite the existing class attribute by spreading to the parent div

packages/renderer/src/lib/task-manager/ProgressBar.svelte
benoitf: `<div  {...restProps} class="flex flex-row {restProps.class}">` I don't know about your 'first' but svelte will keep only the class field `class="flex flex-row {restProps.class}"` even if defined in restProps in this order. So to me it's last definition = winner.

In this order `<div  class="flex flex-row {restProps.class}"  {...restProps}>` it will only keep class with the value from `restProps`

axel7083: IMO this synthax is problematic, as a reader might not understand what is happening when trying to figure out how the component works.

Since this seems to rely on under the hood stuff that decide which one should be applied, I have a preference for the one where the class is not included in the restProps, this is explicit and avoid having a misunderstanding of how it operate :/

benoitf: I didn't use devtools console, I modified the source code and it's last win

axel7083: Yes I understand your point, and this is valid, but this is not something trivial, so I am not sure that this would be easy to understand a mistake could easily be made with this synthax, because this is not something we would document on each component

gastoner:
On WSL Linux:
image
- the example works, just the popup notification does not reflect dark/light theme (not related to this PR)
What do you think about adding some timeout (e.g., 5/10s), after which would the popup disappear?

axel7083:
> What do you think about adding some timeout (e.g., 5/10s), after which would the popup disappear?

This is not related to this PR, but you may open an issue to track that idea

gastoner (approved):
LGTM

PR #9792, opened 2024-11-06
feat(api): enhance the doUpdate signature

### LINKED ISSUE #9339: CliToolUpdate & CliToolInstaller are not using the result of `selectVersion`

While working on https://github.com/containers/podman-desktop-extension-minikube/issues/149 I found some weird behaviour.

Here is the interface for the `CliToolInstaller` from the extension api.

https://github.com/containers/podman-desktop/blob/bca0cdcbf8a2380186d8e830615f7406252993e6/packages/extension-api/src/extension-api.d.ts#L4536-L4540

Here is the workflow we currently have 

1. extension register `CliToolInstaller`
2. User go to CLI Page and click on install
3. Core call `selectVersion` (do nothing with the result ??)
4. Core call `doInstall`

The `doInstall` should take as argument the version got from `selectVersion`, currently this is not the case, and this is how extension are doing:

```ts
  let releaseToInstall: KindGithubReleaseArtifactMetadata | undefined;

 let releaseToInstall: KindGithubReleaseArtifactMetadata | undefined;
  let releaseVersionToInstall: string | undefined;
  kindCli.registerInstaller({
    selectVersion: async () => {
      const selected = await installer.promptUserForVersion();
      releaseToInstall = selected;
      return releaseVersionToInstall; // /!\ useless 
    },
    doInstall: async _logger => {
      if (!releaseToInstall) {
        throw new Error(`Cannot install ${KIND_CLI_NAME}. No release selected.`);
      }

     // install releaseToInstall

      releaseToInstall = undefined;
    },
    doUninstall: async _logger => {
      ...
    },
  });
```

The extension declare a scoped variable that is shared between the `selectVersion` and `doInstall` function. It increase a lot the complexity.

## What happen to the return value of `selectVersion`

It get lost

1. The frontend request SelectVersion

https://github.com/containers/podman-desktop/blob/9e77a7f6d3be727f25781aafcf069e4aea18e106/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte#L45

2. The CliToolRegistry ask the extension the value, and pass it down to the frontend

https://github.com/containers/podman-desktop/blob/31f6a23a40eb8dd48d33fcb5ed520d489bc60ecf/packages/main/src/plugin/cli-tool-registry.ts#L98

3. The frontend send it back to the core

https://github.com/containers/podman-desktop/blob/9e77a7f6d3be727f25781aafcf069e4aea18e106/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte#L58

4. The core use it to create a task `Updating kind to X.Y.Z`

https://github.com/containers/podman-desktop/blob/e312a4dcdbe4aaf7dac89964934ffbc97df3dd8f/packages/main/src/plugin/index.ts#L1405

And that's where it stop.

Solution proposal would be backward compatible

The `CliToolInstaller` should be changed to 

```diff
  export interface CliToolInstaller {
    selectVersion: () => Promise<string>;
-   doInstall: (logger: Logger) => Promise<void>;
+   doInstall: (logger: Logger, version: string) => Promise<void>;
    doUninstall: (logger: Logger) => Promise<void>;
  }
```

And that's all :) 

abvousily this should also to `CliToolUpdate`

```diff
  export interface CliToolSelectUpdate {
    selectVersion: () => Promise<string>;
-   doUpdate: (logger: Logger) => Promise<void>;
+   doUpdate: (logger: Logger,  version: string) => Promise<void>;
  }
```

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

This PR change the signature of the `doUpdate` method registered through `registerUpdate`. This change is backward compatible, as it add an additional argument (the version).

When the user select a version the extension has to keep stored which result the user selected, which might not be ideal, and would be easier to have the version provided by the api once selected.

### Screenshot / video of UI

N/A

Fixes https://github.com/containers/podman-desktop/issues/9339

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4600,7 +4600,7 @@ declare module '@podman-desktop/api' {
 
   export interface CliToolSelectUpdate {
     selectVersion: () => Promise<string>;
-    doUpdate: (logger: Logger) => Promise<void>;
+    doUpdate: (logger: Logger, version: string) => Promise<void>;
   }
 
   export interface CliToolInstaller {
--- a/packages/main/src/plugin/cli-tool-registry.spec.ts
+++ b/packages/main/src/plugin/cli-tool-registry.spec.ts
@@ -228,17 +228,42 @@ suite('cli module', () => {
       const newCliTool = cliToolRegistry.createCliTool(extensionInfo, options);
       // register the updater and call it
       const disposeUpdater = cliToolRegistry.registerUpdate(newCliTool as CliToolImpl, updater);
-      await cliToolRegistry.updateCliTool(newCliTool.id, {} as unknown as Logger);
 
-      expect(updateMock).toBeCalled();
+      const loggerMock: Logger = {} as unknown as Logger;
+      await cliToolRegistry.updateCliTool(newCliTool.id, '1.1.2', loggerMock);
+
+      expect(updateMock).toHaveBeenCalledWith(loggerMock);
       updateMock.mockReset();
 
       // dispose the updater and check it is not called again
       disposeUpdater.dispose();
       expect(apiSender.send).toBeCalledWith('cli-tool-change', 'ext-publisher.ext-name.tool-name');
-      await cliToolRegistry.updateCliTool(newCliTool.id, {} as unknown as Logger);
+      await cliToolRegistry.updateCliTool(newCliTool.id, '1.1.2', {} as unknown as Logger);
 
       expect(updateMock).not.toBeCalled();
+    });
+
+    test('expect CliToolSelectUpdate updater to receive version selected', async () => {
+      const updater: CliToolSelectUpdate = {
+        doUpdate: vi.fn(),
+        selectVersion: vi.fn(),
+      };
+      const options: CliToolOptions = {
+        name: 'tool-name',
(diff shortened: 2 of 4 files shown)


benoitf (changes requested):
> This change is backward compatible, as it add an additional argument (the version).

The change is not backward compatible. If you update some code to this new API, the code will not compile

axel7083:
> > This change is backward compatible, as it add an additional argument (the version).
> 
> The change is not backward compatible. If you update some code to this new API, the code will not compile

Oh my bad, I though typescript ignore new arguments ~ closing

axel7083:
Hummm, i closed a bit fast, why would it not compile ?

https://github.com/containers/podman-desktop/blob/cdee0b7ab1b158a4b4d44a6d5bb90554490eceb1/extensions/compose/src/extension.ts#L315 

we seems to be already ignoring the logger, what would be the difference with the version ?

The only place where it would be an issue would be the tests, which are calling the method, which I admit would be potential problem

benoitf:
well, backward means that code compile as before (even tests)

using `version?: string` is the only backward compliant way.

axel7083:
> using `version?: string` is the only backward compliant way.

This is not ideal, I will think about it :/

PR #9793, opened 2024-11-06
fix: terminal is not restarted if container if not running + state management

### LINKED ISSUE #9771: UI get unresponsive if Container stopped/started while terminal tab is open
### Bug description

UI gets unresponsive after running container gets stopped and then started while terminal tab is open.

### Operating system

macOS Sequoia 15.0.1

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

https://github.com/user-attachments/assets/939ee109-da09-4f0b-a9a2-44f1d3f8ff04

### Relevant log output

```
localhost/:1 Uncaught (in promise) Error: (HTTP code 500) server error - can only create exec sessions on running containers: container state improper 
Promise.then		
receiveEndCallback	@	ContainerDetailsTerminal.svelte:60
(anonymous)	@	VM5:391
emit	@	VM4 sandbox_bundle:2
onMessage	@	VM4 sandbox_bundle:2
```

```
Last ten effects were:  (10) [ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ]0: () => {…}length: 0name: "$effect"arguments: (...)caller: (...)[[FunctionLocation]]: ContainerDetailsTerminal.svelte:33[[Prototype]]: ƒ ()[[Scopes]]: Scopes[3]1: () => {…}2: () => {…}3: () => {…}4: () => {…}5: () => {…}6: () => {…}7: () => {…}8: () => {…}9: () => {…}length: 0name: "$effect"arguments: (...)caller: (...)[[FunctionLocation]]: ContainerDetailsTerminal.svelte:33[[Prototype]]: ƒ ()[[Scopes]]: Scopes[3]length: 10[[Prototype]]: Array(0)
```

```
Uncaught Svelte error: effect_update_depth_exceeded
Maximum update depth exceeded. This can happen when a reactive block or effect repeatedly sets a new value. Svelte limits the number of nested updates to prevent infinite loops
```

### Additional context

_No response_

feloy:

### What does this PR do?

- the terminal is not restarted if the container is not running
- Svelte state is reduced to only the container state, not the entire container information, or  the $effect was respawned in an infinite loop when a container restarts.

### Screenshot / video of UI

https://github.com/user-attachments/assets/9fe9bb5d-709a-4ce4-a9b5-03607700e762

Fixes #9771 

### How to test this PR?

- Start a container, go to the terminal page, check that the terminal works as expected
- Stop the container
- restart the container
- the terminal should be working again
- the app should work 
- no error should appear in the console

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
@@ -28,11 +28,11 @@ let sendCallbackId: number | undefined;
 let terminalContent: string = '';
 let serializeAddon: SerializeAddon;
 let lastState = $state('');
-let containerState = $state(container);
+let containerState = $state(container.state);
 
 $effect(() => {
-  containerState = container;
-  if (lastState === 'STARTING' && containerState.state === 'RUNNING') {
+  containerState = container.state;
+  if (lastState === 'STARTING' && containerState === 'RUNNING') {
     restartTerminal();
   }
   lastState = container.state;
@@ -54,16 +54,20 @@ function receiveDataCallback(data: Buffer) {
 }
 
 function receiveEndCallback() {
-  // need to reopen a new terminal
-  window
-    .shellInContainer(container.engineId, container.id, receiveDataCallback, () => {}, receiveEndCallback)
-    .then(id => {
-      sendCallbackId = id;
-
-      shellTerminal?.onData(data => {
-        window.shellInContainerSend(id, data);
+  // need to reopen a new terminal if container is running
+  if (containerState === 'RUNNING') {
+    window
+      .shellInContainer(container.engineId, container.id, receiveDataCallback, () => {}, receiveEndCallback)
+      .then(id => {
+        sendCallbackId = id;
+        shellTerminal?.onData(data => {
+          window.shellInContainerSend(id, data);
+        });
+      })
+      .catch((err: unknown) => {
+        console.error('error starting shell', err);
       });
-    });
+  }
 }
 
 // call exec command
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
```diff
@@ -28,11 +28,11 @@ let sendCallbackId: number | undefined;
 let terminalContent: string = '';
 let serializeAddon: SerializeAddon;
 let lastState = $state('');
-let containerState = $state(container);
+let containerState = $state(container.state);
 
 $effect(() => {
-  containerState = container;
-  if (lastState === 'STARTING' && containerState.state === 'RUNNING') {
+  containerState = container.state;
```
axel7083: can't we use `let containerState = $derived(container.state);` instead of updating in the effect ?

packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
```diff
@@ -54,16 +54,20 @@ function receiveDataCallback(data: Buffer) {
...
-  // need to reopen a new terminal
-  window
-    .shellInContainer(container.engineId, container.id, receiveDataCallback, () => {}, receiveEndCallback)
-    .then(id => {
-      sendCallbackId = id;
-
-      shellTerminal?.onData(data => {
-        window.shellInContainerSend(id, data);
+  // need to reopen a new terminal if container is running
+  if (containerState === 'RUNNING') {
```
axel7083, line 57: suggestion (non-blocking): early return is easier to read

```ts
if (containerState !== 'RUNNING') return;
```

feloy:
I have found this issue https://github.com/containers/podman-desktop/issues/9794
the problem seems very similar (using restart button instead of stop + start), but the root cause seems different, as this PR does not fix the problem, I'll create another PR to fix that other issue #9794

SoniaSandler (approved):
LGTM, tested the changes and it works

axel7083 (approved):

PR #9796, opened 2024-11-06
fix: restart container on terminal page

### LINKED ISSUE #9794: terminal unresponsive when restarting a container
### Bug description

When on a terminal page for a container, if I press the "restart container", the terminal reappears, but is unresponsive.

I need to go to another page and come back to the terminal to make it work again.

https://github.com/user-attachments/assets/6097c86d-78ec-46dd-9ca2-2e4100a77c56

The problem seems very similar to #9771 (except using restart button instead of stop + start), but the root cause seems different, as the PR #9793 does not fix this problem

### Operating system

mac

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Restart the terminal when the container is in RUNNING state with a previous RESTARTING state (before, only the STARTING state was taken into consideration).

Fixes #9794 

### How to test this PR?

- Start a container and go to its terminal page
- Restart the container (with the "Restart container" button, not stop+start)
- check that the terminal is visible again and that you can type character and get response.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
@@ -31,7 +31,8 @@ let lastState = $state('');
 let containerState = $derived(container.state);
 
 $effect(() => {
-  if (lastState === 'STARTING' && containerState === 'RUNNING') {
+  if ((lastState === 'STARTING' || lastState === 'RESTARTING') && containerState === 'RUNNING') {
+    console.log('==> restart');
     restartTerminal();
   }
   lastState = container.state;
@@ -55,6 +56,7 @@ function receiveDataCallback(data: Buffer) {
 function receiveEndCallback() {
   // need to reopen a new terminal if container is running
   if (containerState === 'RUNNING') {
+    console.log('==> shellInContainer 2');
     window
       .shellInContainer(container.engineId, container.id, receiveDataCallback, () => {}, receiveEndCallback)
       .then(id => {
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
```diff
@@ -31,7 +31,8 @@ let lastState = $state('');
 let containerState = $derived(container.state);
 
 $effect(() => {
-  if (lastState === 'STARTING' && containerState === 'RUNNING') {
+  if ((lastState === 'STARTING' || lastState === 'RESTARTING') && containerState === 'RUNNING') {
+    console.log('==> restart');
```
benoitf: remove

packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
```diff
@@ -55,6 +56,7 @@ function receiveDataCallback(data: Buffer) {
 function receiveEndCallback() {
   // need to reopen a new terminal if container is running
   if (containerState === 'RUNNING') {
+    console.log('==> shellInContainer 2');
```
benoitf: Remove

packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
```diff
@@ -76,6 +78,7 @@ async function executeShellIntoContainer() {
   }
 
   // grab logs of the container
+  console.log('==> shellInContainer 1');
```
benoitf: Remove

benoitf (approved):

SoniaSandler (approved):

benoitf:
to be backported to 1.14.x as well

Issue #9585, opened 2024-10-24
Enable gathering system information and enabled extensions when reporting bugs


slemeur:
### Is your enhancement related to a problem? Please describe

To improve bug reporting and troubleshooting, let's explore enabling the automatic collection of system information and relevant logs when users encounter issues. This would facilitate the resolution process.

### Describe the solution you'd like

Add checkboxes to:
- gather information from the system 
- gather logs 
- collect enabled extensions

Similarly to the following mockup:

![Image](https://github.com/user-attachments/assets/8df65d21-bde2-4035-b95c-62de2bcb1cc4)

axel7083:
Depends on the followings
-  https://github.com/podman-desktop/podman-desktop/pull/9954
- https://github.com/podman-desktop/podman-desktop/pull/10000
- https://github.com/podman-desktop/podman-desktop/pull/9975

axel7083:
Let's split that

- [ ] collect system info
  - [x] https://github.com/podman-desktop/podman-desktop/pull/10098 (backend)
  - [x] https://github.com/podman-desktop/podman-desktop/pull/10116 (frontend)
- [ ] collect logs (truncate ?) hard to know

axel7083:
There is a hard limit (2028) to the size of an URL, we cannot provide the logs through the query parameters, because this is way too big (we could try to truncate, but I am not sure this is the solution). We could try to collect the errors ? But this is a pretty complicated task as we would need to parse etc.

slemeur:
As the ability to gather all the logs is technically not possible without using a 3party service to store them - we will consider that this issue will be solved once those tickets are done:

- [x] https://github.com/podman-desktop/podman-desktop/pull/10276
- [x] https://github.com/podman-desktop/podman-desktop/pull/10098
- [x] https://github.com/podman-desktop/podman-desktop/pull/10116

axel7083:
Thanks @slemeur I rename the issue

axel7083:
Closing as all sub-issues completed

PR #9833, opened 2024-11-10
feat: add a bug report feedback form

### LINKED ISSUE #9584: Differentiate the feedback form for bug reports

Bugs reported in the feedback form are not always actionable. We would like to be able to follow-up and use facilitate tracking those issues (as well as making them public)

Follow the mockup:
Image

(Automatic gathering of the system information will be done in another issue)

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
In this PR, a new feedback form template had been added, the developers and GitHub issue (new form added in this PR) feedback forms were refactored to use the new template, and `SendFeedback.svelte`, was updated to include the new feedback components of `DevelopersFeedback` and `GitHubIssueFeedback`

### Screenshot / video of UI
[Screencast from 2024-11-12 12-36-45.webm](https://github.com/user-attachments/assets/ea330ab9-1abc-4cf3-946b-1abe262502db)

Closes https://github.com/podman-desktop/podman-desktop/issues/9584

### How to test this PR?


slemeur:
That's looking really promising @SoniaSandler ! 
How does that behave when the user doesn't have a github account or is not authenticated? 

(I'm also guessing, the management of the system info + logs will come along later on)

SoniaSandler:
@slemeur It looks like you have to have a GitHub account to open an issue, so if you are not connected to GitHub in your browser, the link from PD will direct you to the GitHub sign in page. From what I understood in the issue, this bug template will only help create a GitHub issue and nothing more, so without a GitHub account, there is not much this template can do.

packages/renderer/src/lib/feedback/SendFeedback.svelte
```diff
@@ -78,6 +98,20 @@ async function openGitHub(): Promise<void> {
...
+  displayModal = false;
+  await window.openExternal('https://github.com/containers/podman-desktop/issues');
+  displayModal = true;
+}
+
+async function sendToGitHub(): Promise<void> {
+  displayModal = false;
+  let linkTitle = issueTitle.split(' ').join('+');
+  let linkDescription = issueDescription.split(' ').join('+');
+  const bugLink = `https://github.com/containers/podman-desktop/issues/new?template=bug_report.yml&title=${linkTitle}&bug-description=${linkDescription}`;
```
axel7083: Might be easier to read to use something like the `URLSearchParams` see https://stackoverflow.com/a/49701878

axel7083 (commented):
That's very cool ! I really like the progress on this one 🚀 !

I have some remarks:

### Template completeness 

You are providing the `bug-description` and `title` to the bug template, if the user select the following options:

image

we might want to include the `os`, `install` (version), and add the list of extensions to the `additional-context` maybe.

### Logic handling

_(open to discussion, throwing ideas to centralize the code related to feedback, ease testing and integration)_

IMHO the logic of opening the url should be in the main package, the reasons behind this idea, is that collecting the system information, list of extensions, maybe logs etc. would be easier in the main. it would also probably be easier to test.

We could have a `feedback.ts` with a FeedBack class, with some generics method

```ts
function reportBug(options: { 
  title: string, 
  description: string, 
  includeSystemInformation: boolean, 
  includeExtensionsList: boolean
 }): Promise<void> { ... }
```

Today, the feedback entry is defined in `packages/main/src/plugin/index.ts`, which could be included in the feedback.ts init method,

https://github.com/podman-desktop/podman-desktop/blob/0f5fbe84bc40c04759824e16923dfcb9a4b09628/packages/main/src/plugin/index.ts#L563-L564

Same for the commands related feedback

https://github.com/podman-desktop/podman-desktop/blob/0f5fbe84bc40c04759824e16923dfcb9a4b09628/packages/main/src/plugin/index.ts#L586-L588

This can be a separate issues

benoitf (commented):
issue: to have specific tests and reduce the scope, you should introduce new .svelte components moving the content to this sub-components rather than a lot of if content else anotherContent

One handling developers, one handling github issue submission

b/c issueTitle, issueDescription is global in this PR while it's only useful for github part.


another issue: split the work for collecting system information and collecting extension information from this PR


like:
 - one PR doing only refactoring of the current code by introducing sub components (no new change)
 - one PR: add new collect options
 - one PR adding GitHub submission

benoitf:
> we might want to include the os, install (version), 

this is already tracked on each telemetry event

SoniaSandler:
@benoitf the current collect info buttons don't do anything, they're just there as part of the design. There is already a separate issue to add the info collection code https://github.com/podman-desktop/podman-desktop/issues/9585
@axel7083 those are some good ideas, thanks! I'll probably use them in a separate PR as per Florent's suggestion

axel7083:
> the current collect info buttons don't do anything,

If they do nothing, better to not have them.

packages/renderer/src/lib/feedback/DevelopersFeedback.svelte
```diff
@@ -0,0 +1,169 @@
...
+
+import type { FeedbackCategory, FeedbackProperties } from '/@api/feedback';
+
+import Dialog from '../dialogs/Dialog.svelte';
+import WarningMessage from '../ui/WarningMessage.svelte';
+
+export let displayModal = false;
+
+const DEFAULT_CATEGORY: FeedbackCategory = 'developers';
+export let category: FeedbackCategory = DEFAULT_CATEGORY;
```
axel7083: Svelte 5 ?

```ts
interface Props {
     displayModal?: boolean;
     category?: FeedbackCategory ;
}

let { displayModal = false, category = DEFAULT_CATEGORY }: Props = $props();
```

packages/renderer/src/lib/feedback/DevelopersFeedback.svelte
```diff
@@ -0,0 +1,169 @@
...
+]);
+
+// feedback of the user
+let smileyRating = 0;
+let tellUsWhyFeedback = '';
+let contactInformation = '';
+
+$: hasFeedback =
+  (tellUsWhyFeedback && tellUsWhyFeedback.trim().length > 4) ||
+  (contactInformation && contactInformation.trim().length > 4);
```
axel7083: if svelte 5, this could be a $derived state

packages/renderer/src/lib/feedback/SendFeedback.svelte
```diff
@@ -1,173 +1,24 @@
...
-    <svelte:fragment slot="buttons">
-      <Button disabled={smileyRating === 0 || (smileyRating === 1 && !hasFeedback)} on:click={() => sendFeedback()}
-        >Send feedback</Button>
-    </svelte:fragment>
-  </Dialog>
+  {#if category === 'developers'}
+    <DevelopersFeedback bind:displayModal={displayModal} bind:category={category} />
+  {:else}
+    <GitHubIssueFeedback bind:displayModal={displayModal} bind:category={category} />
+  {/if}
```
axel7083: IMHO, the Dropdown should be in the `SendFeedback`, and the GitHubIssueFeedback, and DevelopersFeedback  should only be forms that submit to the parent component.

suggestion: Here we are adding a lot of complexity by binding values to children. I would create a folder `packages/renderer/src/lib/feedback/forms` and add `GitHubFeedbackForm`, `DeveloperFeedbackForm`.

idea (not necessary): I would also try to make a generic interface for the properties of those components, so we can new one easily

packages/renderer/src/lib/feedback/GitHubIssueFeedback.svelte
```diff
@@ -0,0 +1,91 @@
...
+  displayModal = true;
+}
+</script>
+
+<Dialog title="Share your feedback" on:close={() => hideModal()}>
+  <svelte:fragment slot="content">
+    <label for="category" class="block mb-2 text-sm font-medium text-[var(--pd-modal-text)]">Category</label>
+    <Dropdown name="category" bind:value={category}
+    options={Array.from(FEEDBACK_CATEGORIES).map(e => ({ value: e[0], label: e[1] }))}>
+    </Dropdown>
```
axel7083: issue: see https://github.com/podman-desktop/podman-desktop/pull/9833/files#r1838514274, we might want to avoid duplicating the dropdown, and have it inside the parent component

packages/renderer/src/lib/feedback/SendFeedback.svelte
SoniaSandler: since Modal uses slots, is there a way I can connect buttons from the developer feedback or github feedback to the Modal in SendFeedback?

benoitf (commented):
hello, is that PR is in final form ? or a work in progress before splitting the work into small chunks ?

b/c here it looks like it contains both refactoring and new stuff so it's hard to see quickly what is new, what is just being moved or refactoring ?

SoniaSandler:
@benoitf for now I'm done with this PR until additional reviews. The info collection buttons are not included here and the most recent commits focused on separating the different feedback templates into smaller components.
As for the GitHub submission, I can remove it from this PR, and add in a new one

benoitf:
> As for the GitHub submission, I can remove it from this PR, and add in a new one

I think it would help if you split with smaller chunks / other PRs

First one being just about refactoring ( nothing related to GitHub issue submission)
Same functionality/ feature, it should behave as before but will ease the addition of new changes

And then another one that will bring this part

SoniaSandler:
removed GitHub submission from this PR, so for now, the Preview on GitHub button opens the issues page

packages/renderer/src/lib/feedback/SendFeedback.svelte
```diff
@@ -43,131 +24,29 @@ function hideModal(): void {
...
-  hideModal();
-  await window.telemetryTrack('feedback.openGitHub');
-  await window.openExternal('https://github.com/containers/podman-desktop');
-}
 </script>
 
 {#if displayModal}
-  <Dialog title="Share your feedback" on:close={() => hideModal()}>
-    <svelte:fragment slot="content">
+  <Modal name="Share your feedback" on:close={hideModal}>
```
benoitf, line 32: question : why the switch from dialog to modal ?

SoniaSandler: Because the Dialog template uses slots for each part (content, validation, and buttons), I wasn't able to find a way to both include the dropdown in `SendFeedback` and to "separate" `DevelopersFeedback` or `GitHubIssueFeedback` between the slots of this Dialog (is it even possible?).
 `DevelopersFeedback` and `GitHubIssueFeedback` have different buttons with different functionalities and disabled situations, and different validations, so they cannot share those slots.
 The new `FeedbackForm.svelte` component uses the same slots introduced in Dialog, and  `DevelopersFeedback` and `GitHubIssueFeedback` use it, while the current Modal is almost the same as the Dialog excluding the slots.

packages/renderer/src/lib/feedback/SendFeedback.svelte
axel7083: Yes, this might be a little complex and use the svelte 5 reactivity. I wrote the following quickly, to illustrate (do not take as it is)

Let's have a `FeedBackModal`, I like the idea of using our Modal, but we need to make sure this do not introduce any visual regression (maybe it is smaller in width or height?)

```svelte
<script lang="ts">
import type { FeedbackFormProps, FormContent } from './feedback-form-props';
import type { FeedbackCategory } from '/@api/feedback';
import GitHubForm from '/@/lib/feedback/GitHubForm.svelte';

// let's have our FeedbackCategory in a state
let category: FeedbackCategory = $state<FeedbackCategory>('developers');
// The content of our form should also be in a state and handled by the parent
let form: FormContent = $state({
  title: '',
  description: '',
});

// We can create a FormComponent state, which 
let FormComponent: Component<FeedbackFormProps> = $derived.by(() => {
  switch (category) {
    case 'developers':
      return DeveloperForm;
    case 'feature':
      return GitHubForm; // might need something dedicated ?
    case 'bug':
      return GitHubForm; // might need something dedicated ? 
  }
});

// function that would handle form change in child component
async function onFormChange(value: FormContent): Promise<void> {
  form = value;
}

// function that would hande dropdown change in this component
function onDropdownChange(nCategory: FeedbackCategory): void {
  // apply new category
  category = nCategory;
  // reset forms when switching category?
  form = {
    title: '',
    description: '',
  };
}
</script>

{/* todo: should handle the hidden stuff */}
<Modal {...}>
  {/* Dropdown */}
  <FormComponent category={category} onchange={onChange} />
  {/* Buttons */}
</Modal>
```

Here are some types I wrote for my example (dedicated file so several component can import and share it)

```ts
import type { FeedbackCategory } from '/@api/feedback';

export interface FormContent {
  // mandatory properties
  title: string;
  description: string;
}

// This should be the props of our form components, it should be shared
export interface FeedbackFormProps {
  category: FeedbackCategory;
  value: FormContent;
  onchange: (value: FormContent) => Promise<void>;
}
```

Let's see how our GitHubForm component would look like, the idea here, would to only have the forms field, the child component will update the value of the form using the onchange props provided

```svelte
<script lang="ts">
import type { FeedbackFormProps } from '/@/lib/feedback/feedback-form-props';
import { Input } from '@podman-desktop/ui-svelte';

let { category, onchange, value }: FeedbackFormProps = $props();

function onTitleChange(e: Event) {
  onchange({
    ...value,
    title: e.event.target,
  });
}

function onDescriptionChange(e: Event) {
  onchange({
    ...value,
    description: e.event.target,
  });
}
</script>

{/* todo: handle ui nicely with labels, styling etc */}
<Input value={value.title} title="title" on:change={onTitleChange}></Input>

{/* todo: handle ui nicely with labels, styling etc */}
<Input value={value.description} title="description" on:change={onDescriptionChange}></Input>
```

SoniaSandler: thanks for showing this, I see what you mean. However, one thing that it still doesn't solve is the fact that the buttons are still in the main code, which means that if we'll add more categories, we might add more button options to the main code, and our `if else` blocks will grow as we add the buttons. 
I planned on moving all the category based components and logic to separate files for each category so that the main code will only include the dropdown and the change between the categories.
I did update this PR yesterday, and added a new template for the forms (includes input, validation, and buttons), and the DevelopersFeedback and GitHubIssueFeedback now use it.

benoitf (approved):

packages/renderer/src/lib/feedback/SendFeedback.svelte
axel7083: Yes, you are right, better to have the buttons on the form, but then who is submitting :thinking: the parent of the child... might be better to have a `submit` function on the children, and a generic logic on the parent

axel7083 (commented):
We may need to disable the button on the GitHub component when no title is provided?

image

SoniaSandler:
for the description too?

axel7083:
> for the description too?

Maybe why not !

packages/renderer/src/lib/feedback/feedbackForms/DevelopersFeedback.spec.ts
```diff
@@ -0,0 +1,174 @@
+/**********************************************************************
+ * Copyright (C) 2023-2024 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/renderer/src/lib/feedback/SendFeedback.spec.ts
```diff
@@ -24,170 +24,47 @@ import { beforeAll, expect, test, vi } from 'vitest';
 
 import SendFeedback from './SendFeedback.svelte';
 
-// fake the window.events object
+const mocks = vi.hoisted(() => ({
+  gitHubIssueMock: vi.fn(),
+  developersFeedbackMock: vi.fn(),
+}));
```
benoitf: issue: please do not use vi.hoisted and mocks

vi.mock and vi.mocked is usually enough

packages/renderer/src/lib/feedback/feedbackForms/DevelopersFeedback.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+import userEvent from '@testing-library/user-event';
+import { beforeAll, expect, test, vi } from 'vitest';
+
+import DevelopersFeedback from './DevelopersFeedback.svelte';
+
+beforeAll(() => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (window as any).openExternal = vi.fn();
+  (window as any).telemetryTrack = vi.fn();
+  (window as any).sendFeedback = vi.fn();
```
benoitf: nitpick: can use `Object.defineProperty(global, 'window',`  pattern so no need to use explicit-any

packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.svelte
```diff
@@ -0,0 +1,57 @@
...
+  } else {
+    issueValidaionError = 'Please enter bug title';
+  }
+} else {
+  if (!issueDescription) {
+    issueValidaionError = 'Please enter bug description';
+  }
+}
+
+export let hideModal = () => {};
```
benoitf: issue: it looks strange to have a method named hideModal that do nothing

either it's a hideModal that do something or if we don't know what it's doing, it's a callback like: beforeOpen or afterOpen, etc and the hide thing is not known by this component

SoniaSandler: this was meant to be the `hideModel` method that is passed from `SendFeedback.svelte`, so that this component could call it when needed.

benoitf: yes but this component is not aware of any modal. Let say you import this component from another page. You might not have any modal page.

so you're interested in having events from this specific component, and it's the parent component that do the mapping between an event or a callback method and the action

so for the parent component, it's to hide a modal, but for the child it's just a beforeOpenGithubIssue callback

packages/renderer/src/lib/feedback/FeedbackForm.svelte
```diff
@@ -0,0 +1,12 @@
+<div>
```
benoitf, line 1: issue: this component has no unit tests

packages/renderer/src/lib/feedback/feedbackForms/DevelopersFeedback.spec.ts
SoniaSandler: for some reason, using the `Object.defineProperty(global, 'window'` pattern caused the tests to fail (rendered components from previous tests "leaked" into the tests after them), so I used `Object.defineProperty(window, '<method name>'` for each one

axel7083 (approved):

packages/renderer/src/lib/feedback/FeedbackForm.spec.ts
```diff
@@ -0,0 +1,31 @@
+/**********************************************************************
+ * Copyright (C) 2023-2024 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/renderer/src/lib/feedback/feedbackForms/FeedbackFormTest.svelte
```diff
@@ -0,0 +1,12 @@
...
+</script>
+  
+  <FeedbackForm>
+    <p slot="content" aria-label="content">Content</p>
+  
+    <p slot="validation" aria-label="validation">Validation</p>
+  
+    <p slot="buttons" aria-label="buttons">Buttons</p>
+  </FeedbackForm>
+
```
benoitf: issue: missing new line/EOF

packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.svelte
```diff
@@ -0,0 +1,57 @@
...
+  if (!issueDescription) {
+    issueValidaionError = 'Please enter bug description';
+  }
+}
+
+export let onCloseForm = () => {};
+
+async function openGitHubIssues(): Promise<void> {
+  onCloseForm();
+  await window.openExternal('https://github.com/containers/podman-desktop/issues');
```
benoitf:
```suggestion
  await window.openExternal('https://github.com/podman-desktop/podman-desktop/issues');
```

packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.svelte
```diff
@@ -0,0 +1,57 @@
...
+}
+</script>
+
+<FeedbackForm>
+  <svelte:fragment slot="content">
+    <div class="text-sm text-[var(--pd-state-warning)]">You can search existing issues on <Link aria-label="GitHub issues" onclick={openGitHubIssues}>github.com/containers/podman-desktop/issues</Link></div>
```
benoitf:
```suggestion
    <div class="text-sm text-[var(--pd-state-warning)]">You can search existing issues on <Link aria-label="GitHub issues" onclick={openGitHubIssues}>github.com/podman-desktop/podman-desktop/issues</Link></div>
```

packages/renderer/src/lib/feedback/feedbackForms/DevelopersFeedback.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+  vi.resetAllMocks();
+  Object.defineProperty(window, 'openExternal', {
+    value: openExternalMock,
+  });
+  Object.defineProperty(window, 'telemetryTrack', {
+    value: telemetryTrackMock,
+  });
+  Object.defineProperty(window, 'sendFeedback', {
+    value: sendFeedbackMock,
+  });
```
benoitf: what is the issue with the single defineProperty ?

AFAIK you should put it in beforeAll block and not beforeEach to make it work properly (to avoid issue about trying to redefine something)

benoitf (approved):
only few nitpicks

packages/renderer/src/lib/feedback/feedbackForms/DevelopersFeedback.spec.ts
SoniaSandler: I'm not sure why, but for some reason, using the Object.defineProperty(global, 'window' pattern (in a beforeAll) caused the tests to fail (rendered components from previous tests "leaked" into the tests after them), so I used Object.defineProperty(window, '<method name>' for each one

benoitf: ok, it's not blocking the PR anyway

PR #9867, opened 2024-11-13
feat: implement dropdown menu for status bar help button

### LINKED ISSUE #9407: Move Help page content to a Dropdown

Part of [Epic](https://github.com/containers/podman-desktop/issues/9077), there are a lot useful links hidden away in the help page. We want to give quicker acess to those resources by moving them into a context menu instead. 

Keep only 1 icon and move all the options into a menu. As it is a menu it also allows us to add a text label which will help our users understand what the icons mean. 

Check the [Epic](https://github.com/containers/podman-desktop/issues/9077) to see what the content of the help menu should be.

dgolovin:
### What does this PR do?

Adds dropdown menu for status bar help icon

### Screenshot / video of UI

image

Fix #9407.

### How to test this PR?


dgolovin:
Tests have not implemented yet.

Firewall:
Thanks for the PR @dgolovin, looking great so far!

A couple of questions:
1. Will we remove the `troubleshooting` and `share your feedback` buttons from the statusbar in another PR?
2. I think it would be nice if somehow showed some hierarchy between the headers and the actual links. Maybe an indent? Maybe make the title not have an icon? Maybe a divider? I don't know what our options are with the current compoennt

dgolovin:
@Firewall Updated the PR and screenshot.

benoitf:
@Firewall the issue was not telling to remove the other icons from the status bar

Firewall:
Thanks @dgolovin, this is a lot clearer in my opinion. 

Seeing the content this way... Do you think we should move the `Create Github Discussion` and `Share your feedback` to under the `Report a Bug` instead? The way it is now they look like sub-items of `Learning` and that is not the intention.

benoitf:
IMHO I would separate the removal of other icons from this PR. Scope is too big.
First: add things/rework only help and get feedback, second: remove things, not do too many things at once.

The title of the issue or the PR is to move help page to a dropdown, not removing bulb and share feedback icons, so as a reviewer, it's not doing what is said, so I'm requesting changes.

About the content, it looks like the original intent was to reduce the number of clicks while here I would do extra clicks.(if bulb and report feedback direct action icons are moved)

But my point is that to click on useful links, I would have to move the cursor far way from the first click it looks like it'll do the opposite of what it was describing (looks like import links are on the top while clicking on a menu from the bottom I expect the opposite)

Firewall:
Ok, I created a sperate ticket for removing them: https://github.com/podman-desktop/podman-desktop/issues/9900

@benoitf I'll respond to your comments on that issue too.

Firewall:
> looks like import links are on the top while clicking on a menu from the bottom I expect the opposite)

Now that I saw the new submenu in action. I agree. @dgolovin can we try to reverse the list and see if that helps gets the most important actions to stand out more?

The list also feels very "tall" for not that many options. Maybe we can reduce the height of each item a bit?

axel7083:
suggestion (non-blocking): I don't think we need the padding on the left for the dropdown items since we have a clear separation, IMO it feel a bit odd to have a padding

image

axel7083:
> The list also feels very "tall" for not that many options. Maybe we can reduce the height of each item a bit?

@Firewall trying to reduce the size might lead to inconsistency with other dropdowns, maybe we should see with @ekidneyrh on the UX call today ?

ekidneyrh:
> > The list also feels very "tall" for not that many options. Maybe we can reduce the height of each item a bit?
> 
> @Firewall trying to reduce the size might lead to inconsistency with other dropdowns, maybe we should see with @ekidneyrh on the UX call today ?

I agree with the left padding feeling odd given the divider already in place. 

I'm not sure what can be done about the height spacing though given that that is the pattern of the drop downs? Unless we were to change them for everything. I personally dont think it being high much of an issue tho. Some menus are just long.

We could divide up the menu with an icon for each section instead of having it as a long menu?

Firewall:
@dgolovin I brought this as a topic to the UX call and we got a bunch of feedback from the team.

1. The options for `Matrix` and `Libera` can be removed, they are not used.
2. For Discord and Slack we we use the icons for the respective services? It was also suggested we can remove the words `Join` and `channel`. So it would just be `:discord-icon: #podman-dekstop`and`:slack-icon: #general`
3. The learning links feel out of place and are duplicate of the content on the learning center on the dashboard. As a result we can remove them.
4. There was a feeling of having too many options to choose the same thing. `Report a bug`, `Create Github Discussion` and `Share your feedback`. It was suggested we delete the `Create Github discussion` option for now as this will be an option in the `Share your feedback` dialog soon. 
5. It was also voted for now to not remove the icons in the statusbar so I will close the other ticket as `wont-do`

Let me know if you have any questions.

dgolovin:
@Firewall After removing everything you asked it turned to image below. 
Please note:
1. I had to remove Learning, because it would have only feedback item
2. Removed indent for Community, as @axel7083 suggested
3. Cannot use icons for discord and matrix due to lack of support custom icons for DropdownMenuItem, so I just left names for discord and matrix.

image

benoitf (commented):
once ppl are ok, before being able to merge, unit tests need to be added (there are none for now)

benoitf:
> Cannot use icons for discord and matrix due to lack of support custom icons for DropdownMenuItem, so I just left names for discord and matrix.

AFAIK we can use custom icons using fonts but as we can use font awesome icons easily, discord icon is part of 
https://www.npmjs.com/package/@fortawesome/free-brands-svg-icons
https://fontawesome.com/icons/discord?f=brands&s=solid

dgolovin:
> > Cannot use icons for discord and matrix due to lack of support custom icons for DropdownMenuItem, so I just left names for discord and matrix.
> 
> AFAIK we can use custom icons using fonts but as we can use font awesome icons easily, discord icon is part of https://www.npmjs.com/package/@fortawesome/free-brands-svg-icons https://fontawesome.com/icons/discord?f=brands&s=solid

@benoitf Would follow up issue work here?

benoitf:
yes for icon as follow-up, not for unit tests as follow-up

axel7083:
Comment from @Firewall (https://github.com/podman-desktop/podman-desktop/pull/9867#issuecomment-2476450314)

> The options for Matrix and Libera can be removed, they are not used.

Tested the lasted version of the PR, I am getting the following
image

Discord and Matrix ? I thought we would keep the slack and discord ?

dgolovin:
Good catch @axel7083, will fix that. Thank!

dgolovin:
@Firewall Final version

image

packages/renderer/src/lib/help/HelpActions.svelte
```diff
@@ -0,0 +1,88 @@
...
+    onClick();
+  };
+}
+
+function createOpenExternal(link: string) {
+  return createOnClick(() => {
+    window.openExternal(link).catch(console.error);
+  });
+}
+</script>ß
```
axel7083:
```suggestion
</script>
```

Weird letter ?

image

Firewall:
@dgolovin Looks great! So close :D but I think for Discord the channel is also called `#podman-desktop` and not `#general`

packages/renderer/src/lib/help/HelpActions.svelte
```diff
@@ -0,0 +1,88 @@
...
+        icon={faComment}
+        onClick={createOnClick(()=> { window.executeCommand('feedback').catch(console.error); })}
+      />
+      <DropdownMenu.Item
+        title="Community"
+        icon={faListUl}
+        enabled={false}
+      />
+      <DropdownMenu.Item
+        title="#general"
```
dgolovin:
```suggestion
        title="#podman-desktop"
```

dgolovin:
> @dgolovin Looks great! So close :D but I think for Discord the channel is also called `#podman-desktop` and not `#general`
Fixed.

axel7083 (approved):
LGTM, great improvement !

Firewall (approved):
Awesome perfect now, thanks!

dgolovin:
I added the tests to verify if all the items are showed up in menu.

shipsing:
Nice improvement in GUI . Thanks @dgolovin for simplification.

packages/renderer/src/lib/help/HelpActions.svelte
```diff
@@ -0,0 +1,53 @@
...
+  if (key === 'Escape') {
+    showMenu = false;
+  }
+}
+
+function onWindowClick(e: any): void {
+  showMenu = outsideWindow.contains(e.target);
+}
+
+function createOnClick(action?: ItemAction): () => void | undefined {
```
benoitf: question: wondering why the method is not async ?

packages/renderer/src/lib/help/HelpActions.svelte
```diff
@@ -0,0 +1,53 @@
...
+        title={item.title}
+        tooltip={item.tooltip}
+        icon={item.icon}
+        enabled={item.enabled}
+        onClick={createOnClick(item.action)}
+      />
+      {/each}
+    </HelpMenu>
+  {/if}
+</div>
```
benoitf: EOF/newline

packages/renderer/src/lib/help/HelpMenu.svelte
```diff
@@ -0,0 +1,33 @@
...
+  bind:offsetHeight={dropDownHeight}
+  bind:offsetWidth={dropDownWidth}
+  bind:this={dropDownElement}
+  class="absolute">
+  <div
+    title="Help Menu Items"
+    class="z-10 m-1 rounded-md shadow-lg bg-[var(--pd-dropdown-bg)] ring-2 ring-[var(--pd-dropdown-ring)] hover:ring-[var(--pd-dropdown-hover-ring)] divide-y divide-[var(--pd-dropdown-divider)] focus:outline-none">
+    <slot />
+  </div>
+</div>
```
benoitf: newline

packages/renderer/src/lib/help/HelpActions.svelte
dgolovin: Do you mean we should have something like this, so svelte would handle exceptions for us?
```
function createOnClick(action?: ItemAction): () => Promise<unknown> {
  return async () => {
    toggleMenu();
    if (action?.kind === ActionKind.LINK) {
      return window.openExternal(action.parameter);
    } else if (action?.kind === ActionKind.COMMAND) {
      return window.executeCommand(action.parameter);
    }
  };
}
```

packages/renderer/src/lib/help/HelpActions.svelte
dgolovin:
```suggestion
</div>

```

packages/renderer/src/lib/help/HelpMenu.svelte
dgolovin:
```suggestion
</div>

```

dgolovin:
Re-based on top of origin/main.

packages/renderer/src/lib/help/HelpItems.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export const Items: readonly ItemInfo[] = [
+  {
+    title: 'Getting Started',
+    icon: faExternalLinkAlt,
+    enabled: true,
+    action: {
+      kind: ActionKind.LINK,
+      parameter: 'https://podman-desktop.io/docs/intro',
```
benoitf: https://github.com/podman-desktop/podman-desktop/blob/main/package.json#L5

if we can reuse the link from the package.json so we only change at a single location

packages/renderer/src/lib/help/HelpItems.ts
```diff
@@ -0,0 +1,109 @@
...
+      parameter: 'troubleshooting',
+    },
+  },
+  {
+    title: 'Report a Bug',
+    icon: faExclamationTriangle,
+    enabled: true,
+    action: {
+      kind: ActionKind.LINK,
+      parameter: 'https://github.com/podman-desktop/podman-desktop/issues/new/choose',
```
benoitf: https://github.com/podman-desktop/podman-desktop/blob/main/package.json#L4 ? for the repository GitHub link

benoitf (commented):
some nitpicks to fix

I'm worried that we're adding links on the frontend side. Frontend should just be a display thing.
Tutorials are managed on the backend side, here it should be the same

It's hard to review all the changes (let say when we changed the organization location for example)

I would then move the links to something more editable and reuse as much as possible the existing links that we have

for example the website root location is part of the package.json so as we're doing for the release notes it should reuse that field

packages/renderer/src/lib/help/HelpActions.svelte
benoitf: it's not possible to have directly the method being async and use await when calling the window methods ?



```
function async createOnClick(): Promise<unknown> {
  toggleMenu();
  if () {
     await window....
```

benoitf: the main point is to use await when calling the window. methods, not having to catch all the calls.

and yes svelte will display the error anyway in the console if the call is failing

dgolovin: Updated function to async and called it as `onClick{()  => onClick()}`.
The same pattern seems to be used everywhere.

packages/renderer/src/lib/help/HelpItems.ts
dgolovin: It cannot be reused right now without additional changes on core side, because packages/* included into electron app assembly without package.json files

image

if we include package.json into assembly as main/package.json, then it can be referenced at runtime/devtime with relative path.

dgolovin: It is actually right there as a sibling of packages folder.

dgolovin: It looks like vite build just bundles package.json into cjs file. Tested on macOS. Works for me.

benoitf (approved):
Thx for the changes

PR #9874, opened 2024-11-13
fix(k8s-port-forward): delete config if start fails

### LINKED ISSUE #9640: bug(k8s-port-forward): host not reachable still create the port forward
### Bug description

If we create a kubernetes resource and specify a port which has no listener, the port forward will fail, but still create a mapping.

### Operating system

Fedora 40

### Installation Method

None

### Version

1.13.3

### Steps to reproduce

1. create a pod with ports without any listener in it
2. Try to forward the fake port

### Relevant log output

Image

### Additional context

_No response_

axel7083:
### What does this PR do?

Before we were creating the port forward even if an error was raised in the middle, this is because we are creating it in two phases. (creation, then starting).

If the start phase fail we should delete the port foward, as we are not able to show a failing state, but we handle the error, so we should only display the error

### Screenshot / video of UI

https://github.com/user-attachments/assets/bfbd6013-e5b0-454e-8f9c-8c8c30233c26

Fixes https://github.com/podman-desktop/podman-desktop/issues/9640

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
@@ -46,10 +46,10 @@ import type { FileSystemWatcher } from '@podman-desktop/api';
 import { beforeAll, beforeEach, describe, expect, type Mock, test, vi } from 'vitest';
 
 import type {
-  type KubernetesPortForwardService,
+  KubernetesPortForwardService,
   KubernetesPortForwardServiceProvider,
 } from '/@/plugin/kubernetes/kubernetes-port-forward-service.js';
-import type { ForwardConfig, type ForwardOptions, WorkloadKind } from '/@api/kubernetes-port-forward-model.js';
+import { type ForwardConfig, type ForwardOptions, WorkloadKind } from '/@api/kubernetes-port-forward-model.js';
 import type { V1Route } from '/@api/openshift-types.js';
 
 import type { ApiSenderType } from '../api.js';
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
```diff
@@ -265,6 +270,17 @@ const apiSender: ApiSenderType = {
...
+const mocks = vi.hoisted(() => ({
+  getServiceMock: vi.fn(),
+}));
+
+vi.mock('/@/plugin/kubernetes/kubernetes-port-forward-service', () => ({
+  KubernetesPortForwardService: vi.fn(),
+  KubernetesPortForwardServiceProvider: vi.fn().mockReturnValue({
+    getService: mocks.getServiceMock,
+  } as unknown as KubernetesPortForwardServiceProvider),
+}));
```
axel7083, line 282: I don't like having to use `vi.hoisted` but I do not have any choice here...

The `KubernetesPortForwardServiceProvider` is not initialized in the constructor of the KubeClient but statically, when the file is imported

https://github.com/podman-desktop/podman-desktop/blob/36d8254e73baafcc79e9a3640e92772e8aaae89a/packages/main/src/plugin/kubernetes/kubernetes-client.ts#L182

This mean the `KubernetesPortForwardServiceProvider` constructor is called before the beforeEach and before the tests are ran... so we cannot mock in other places...

benoitf:
there are errors in PR checks (before looking at the hoisted issue)

benoitf:
provided https://github.com/axel7083/podman-desktop/pull/2 to remove hoisted

gastoner (approved):
LGTM 🚀 


https://github.com/user-attachments/assets/381a5f17-2380-4a36-8dab-da63f9a5ac3b

benoitf:
my comment and PR have been ignored

axel7083:
> my comment and PR have been ignored

I'm very sorry I will revert tomorrow. I don't know why but I've not seen the comments when I clicked on merged.

PR #9876, opened 2024-11-13
fix(ui): better handle errors on `KubePort` component

### LINKED ISSUE #9686: port forwarding error should be removed when the port forward is deleted
### Bug description

As noted by @feloy in https://github.com/containers/podman-desktop/pull/9682#issuecomment-2447549286 the error is staying even after the user has deleting the port forwarding.

We should remove/reset the error state after the deletion.

### Operating system

Fedora 40

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. forward a port that does not exist
2. assert error visible
3. remove the port forwarding
4. assert error still visible

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Improve the error handling for the port forward actions (display delete errors if any, cleanup previous error when retrying)

### Screenshot / video of UI

N/A

Fixes https://github.com/podman-desktop/podman-desktop/issues/9686

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/kube/details/KubePort.svelte
+++ b/packages/renderer/src/lib/kube/details/KubePort.svelte
@@ -59,11 +59,13 @@ async function removePortForward(): Promise<void> {
   if (!mapping) return;
   if (!forwardConfig) return;
   loading = true;
+  error = undefined;
 
   try {
     await window.deleteKubernetesPortForward(forwardConfig);
   } catch (err: unknown) {
     console.error(err);
+    error = String(err);
   } finally {
     loading = false;
   }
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/kube/details/KubePort.svelte
```diff
@@ -59,11 +59,13 @@ async function removePortForward(): Promise<void> {
   if (!mapping) return;
   if (!forwardConfig) return;
   loading = true;
+  error = undefined;
 
   try {
     await window.deleteKubernetesPortForward(forwardConfig);
   } catch (err: unknown) {
     console.error(err);
```
feloy, line 67: Do we want to keep to error on the console, as it is now displayed in the UI?

feloy (approved):
Tested, works as expected.

packages/renderer/src/lib/kube/details/KubePort.svelte
axel7083: I would argue yes, because traces, when we display to the UI, we might lose information because of stringify, I am not sure what to follow here, we don't have much guidelines on this point

benoitf (approved):

PR #9910, opened 2024-11-14
fix: a containerfile outside context can be used when building image

### LINKED ISSUE #8460: Building an image with different context folder selected is broken
### Bug description

I was trying to build an image using containerfile mentioned in README of the sso extension.

I got: `.../build/oolbox.containerfile: no such file or directory` during the build. It seems that building the image is missing a letter. I assume that the problem is on podman-desktop side.

Screenshot_20240809_173829

### Operating system

Mac OS 14 M2

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. Open Build an image page
2. choose a containerfile to build
3. Change the context folder location to something else
4. Build an image
5. Assert: Image is successfully built
Actual result: There is an error during the build: `.../build/oolbox.containerfile: no such file or directory`

### Relevant log output

_No response_

### Additional context

Issue was opened on sso side before: https://github.com/redhat-developer/podman-desktop-redhat-account-ext/issues/233

feloy:
### What does this PR do?

A containerfile outside context can be used when building image

Part of the PR originates from @SoniaSandler 's PR https://github.com/podman-desktop/podman-desktop/pull/8924

Fixes #8460

### How to test this PR?

Build an image with a Containerfile outside of the context directory

Tested on Windows, Linux (with Docker), on Mac


packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -1530,6 +1554,80 @@ describe('buildImage', () => {
...
+    vi.spyOn(dockerAPI.modem, 'followProgress').mockImplementation((_s, f, _p) => {
+      return f(null, []);
+    });
+
+    vi.mocked(fs.existsSync).mockImplementation(path => {
+      return String(path).endsWith('Containerfile.0') || String(path).endsWith('Containerfile.1');
+    });
+    vi.mocked(dockerAPI.buildImage).mockReset();
+
+    contextDir = await mkdtemp(path.join(os.tmpdir(), 'pd-tests-'));
```
benoitf: I don't see why we need to create a directory. why are we not mocking all the fs operations ?
it will avoid to have to delete some folders at the end

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -410,6 +418,17 @@ vi.mock('node:stream/promises', async () => {
...
+vi.mock(import('node:fs/promises'), async importOriginal => {
+  const actual = await importOriginal();
+  return {
+    ...actual,
+    readFile: vi.fn(),
+    rm: vi.fn(),
+  };
+});
+
+let contextDir: string | undefined;
```
benoitf: issue: would be a lot easier to mock all fs operations
new tests could be complicated to write if half of the calls are real and the other not.

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -431,6 +450,11 @@ beforeEach(() => {
 
 afterEach(() => {
   vi.useRealTimers();
+  if (contextDir) {
+    // only remove directory, as the contextDir is never populated during tests
+    fsActual.rmdirSync(contextDir);
+    contextDir = undefined;
+  }
```
benoitf: to be removed with mocking everything

packages/main/src/plugin/container-registry.ts
```diff
@@ -2426,11 +2427,30 @@ export class ContainerProviderRegistry {
...
 
+      let tarStream: NodeJS.ReadableStream;
+      if (options?.containerFile?.startsWith('../')) {
+        // If containerfile is outside the context, we add it to the tar archive, without overriding an existing one
+        const containerFileContent = await readFile(path.join(containerBuildContextDirectory, options.containerFile));
+        let i: number = 0;
+        while (fs.existsSync(path.join(containerBuildContextDirectory, `Containerfile.${i}`))) {
+          i++;
+        }
+        options.containerFile = `./Containerfile.${i}`;
```
benoitf, line 2442: how does it work with docker backend ? or with Dockerfiles ?

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -133,13 +133,34 @@ async function buildContainerImage(): Promise<void> {
...
+  // add .. for the remaining length of the context path (if the context path didn't fully match part of the file path)
+  // then add the remaining path of the file
+  if (index < contextPath.length) {
+    let back = Array(contextPath.length - index).fill('..');
+    relativePath = `${back.join('/')}/${filePath.slice(index).join('/')}`;
+  } else {
+    relativePath = filePath.slice(index).join('/');
+  }
+  return relativePath;
+}
```
benoitf: it looks barely understandable, should it call something from the backend ? instead of trying to handle paths on the client side ?

SoniaSandler: There is a method `path.relative(from, to)` in the `path` module, which can't be imported on the renderer side, so if this path handling will be moved to the backend, then maybe that method could be used there.

benoitf: yes, might be more convenient to delegate something on the backend to handle this use-case

packages/main/src/plugin/container-registry.spec.ts
feloy: right, it got me some time to understand that `tar-fs` module uses `lstat.call(...)` to call `lstat`, and because of this the mocks are not working. Fortunately `tar.pack` accepts an optional `fs`, to pass the mocked fs.

I changed to use this optional `fs`

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
feloy: I'm wondering if that would be interesting to change the signature of the `window.buildImage` function, to accept an absolute path instead of the relative path it accepts currently

benoitf: we can change everything on `window.buildImage` as it's internal

benoitf: but what is sure, it's that we should avoid any path handling on the frontend side while we have all API on backend side

feloy: Yes, I agree with this. 
But having the frontend compute the relative path seems weird to me with this change, as the containerfile is not expected to be inside the context anymore.

It seems more logical to pass the two paths as absolute, and let the backend compute the relative path (as it is now an implementation detail)

feloy: But in the other hand, this will introduce a difference between the api's `buildImage` which will still accept a relative path (and which we cannot change) and the internal `buildImage` which would accept an absolute path. That can be confusing for developers of the core.

feloy: finally, I added a `window.pathRelative` function, to keep the same signatures for both `buildImage` functions (internal and API)

packages/main/src/plugin/container-registry.ts
feloy: Tested on Linux with Docker backend and an original file named Dockerfile, it works

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -410,6 +415,15 @@ vi.mock('node:stream/promises', async () => {
...
 
+vi.mock(import('node:fs/promises'), async importOriginal => {
+  const actual = await importOriginal();
+  return {
+    ...actual,
+    readFile: vi.fn(),
+    rm: vi.fn(),
+  };
+});
+
```
benoitf: question: hello, I'm wondering why we're mocking half of the method

spying only readFile and rm at the beginning of the unit test is not enough ?

feloy: I'm not sure what would be the best solution. I pushed a new commit. Is that what you are thinking about?

benoitf: yes there were several choice, you took one, I'm fine

benoitf (approved):
thanks for the changes, much more readable now 🚀 
I approved, code is looking good fine to me now

I added QE as reviewers so they can review it before being able to merge

odockal (approved):
LGTM. I tried on Linux and two images can be built from a containerfile in different context dir.

SoniaSandler (approved):
Tested on Linux and it works, thanks for the fix!

PR #9926, opened 2024-11-16
fix(extensions/kind): adjust contour download script for new octokit version

Blaimi:
the current version of octokit requires to use ESM imports


benoitf (approved):
thanks for the fix

PR #9928, opened 2024-11-16
feat(extensions/kind): enable the selection between different ingress-controllers

### LINKED ISSUE #2142: Kind: We should use nginx-ingress instead of Contour

Nginx-ingress is more popular as well as used in the official Kubernetes documentation: https://kubernetes.io/docs/concepts/services-networking/ingress/

It's production ready and developers are familiar with it compared to Contour.

Nginx-ingress would be more stable, better backwards compatibility and more familiarity with common nginx annotations that developers add such as: `nginx.ingress.kubernetes.io/rewrite-target: /`

This would make our ingress controller more compatible.

Nginx-ingress replacing contour

_No response_

### Additional context

_No response_

Blaimi:
Closes: #2142


Blaimi:
nginx successfully tested to get a 404 on http://localhost:9090 on fedora 41 on a fresh cluster. nothing else tested.

benoitf:
outdated

Blaimi:
@benoitf I rebased this. Since there are still activities in #2142, it might be useful to reopen this PR.

Blaimi:
[Kong](https://docs.konghq.com/kubernetes-ingress-controller/latest/install/helm/?install=oss) and Traefik could also be an option. Traefik is probably a bit tricky, because they only provide a installation-method via helm. k3s is installing it via its own helm-operator. Another solution could be to `helm template` in the download-step with some prepared values but this would require `helm` to be available during this step.

benoitf:
@Blaimi I can't reopen the PR as to be able to reopen the PR, it needs to be in the same state when closed

PR #9940, opened 2024-11-18
chore: switch to msw instead of nock

### LINKED ISSUE #9232: see if nock can still be used or switch only to mswjs

we have failures in the update of the nock library, 

They moved to mswjs interceptors in nock library so there is little less value in using nock

Thus maybe we can drop nock and just use msw library

avoid to have an unmerged PR from dependabot

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
there were some issues since new beta versions
also nock switched to msw interceptors so better to just use msw

only spec files are being updated

fixes https://github.com/podman-desktop/podman-desktop/issues/9232

### How to test this PR?


axel7083 (approved):
Thanks for the refactoring, this is great :+1:

gastoner (approved):
Codewise LGTM

PR #9944, opened 2024-11-18
chore: increase z-index order for tooltips

benoitf:
### What does this PR do?
A lot of elements are using z-50 order, so pick an higher value

related to https://github.com/podman-desktop/podman-desktop/pull/9528#issuecomment-2482838257

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/tooltip/Tooltip.spec.ts
+++ b/packages/ui/src/lib/tooltip/Tooltip.spec.ts
@@ -39,3 +39,11 @@ test('tooltip is not empty string when tooltipHidden value false', async () => {
   await tick();
   expect(screen.queryByText('test 1')).toBeInTheDocument();
 });
+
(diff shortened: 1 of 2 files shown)


axel7083 (approved):
I don't see immediate issue.

suggestion (non-blocking, but would be nice IMO): Might be future proof to make the value configurable with a default at 60. Maybe we will need in specific cases (Modal, dropdown etc.) to have other values, so might as well have it configurable

benoitf:
assuming tailwind only provide range to z-50 (without hardcoding) I took a value higher than the allowed list from tailwind

Modal is using `z-50`, DropDown `z-10`

axel7083:
> assuming tailwind only provide range to z-50 (without hardcoding) I took a value higher than the allowed list from tailwind

Okey

gastoner (approved):
LGTM

PR #9946, opened 2024-11-18
feat: port-forward on deployments

### LINKED ISSUE #9661: Support port forwarding for deployments

Port forwarding support is only supported for pods

Extend to deployments

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

Port-forward on deployments (will port forward on the first pod managed by the deployment)

### Screenshot / video of UI

https://github.com/user-attachments/assets/b8db3ac9-9925-4aaa-9dea-637498fd8c27

Fixes #9661 

### How to test this PR?

Start a deployment with one or several pods, go to the Deployment Summary and create the port forward.

If the deployment has several replicas, check that the port forward is done in a specific pod.

### DIFF
--- a/packages/renderer/src/lib/deployments/DeploymentDetailsSummary.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentDetailsSummary.svelte
@@ -25,7 +25,7 @@ basic information -->
   {#if deployment}
     <KubeObjectMetaArtifact artifact={deployment.metadata} />
     <KubeDeploymentStatusArtifact artifact={deployment.status} />
-    <KubeDeploymentArtifact artifact={deployment.spec} />
+    <KubeDeploymentArtifact deploymentName={deployment.metadata?.name} namespace={deployment.metadata?.namespace} artifact={deployment.spec} />
     <KubeEventsArtifact events={events} />
   {:else}
     <p class="text-[var(--pd-state-info)] font-medium">Loading ...</p>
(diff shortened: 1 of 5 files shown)


axel7083 (commented):
If I forward a port from the deployment details page, I do not see the port forward inside the pod summary, which is IMO a problem... :/

**deployment details**
image

**pod details**
image

I don't know if we should remove the `WorkloadKind.DEPLOYMENT` because it does not make any sense, since we forward the managed pod, or we should check for type WorkloadKind.DEPLOYMENT in the pod summary :/

feloy:
> If I forward a port from the deployment details page, I do not see the port forward inside the pod summary, which is IMO a problem... :/
> 
> **deployment details** image
> 
> **pod details** image
> 
> I don't know if we should remove the `WorkloadKind.DEPLOYMENT` because it does not make any sense, since we forward the managed pod, or we should check for type WorkloadKind.DEPLOYMENT in the pod summary :/

I don't understand why it is a problem, compared to how the port forward is currently done for Services.

When a port forward for a service is created, if I understand correctly, we create a port forward on the first pod matching the selector of the service. But the port forward is not displayed in the Pod summary.

axel7083:
> if I understand correctly, we create a port forward on the first pod matching the selector of the service.

Hummm I thought service where something on its own, but after reading the documentation, you are right

benoitf (approved):
I thought I approved but it was just pending on my side

axel7083 (approved):

Issue #9952, opened 2024-11-18
Dropdown child options not treated as state


deboer-tim:
### Is your enhancement related to a problem? Please describe

See #9917 for background. When you have a dropdown with child options, Svelte does not treat the options like state since they are neither set via the options property nor $state variables. In the rare case where you can transition between screens that each have one of these dropdowns, Svelte reuses the same component instance and updates all set properties and state but not the options, causing the wrong options to be shown.

### Describe the solution you'd like

I've tried a few simple fixes and they didn't work b/c Svelte was optimizing out the reactivity, and I found one path that works using $effect but the code is triggered way too often. Some options are:

- Follow up with Svelte, it's possible this is a bug / oversight.
- Remove support for child options.
- Find another fix.

PR #9953, opened 2024-11-18
fix: use option property to ensure options is treated as state

### LINKED ISSUE #9917: Preferences -> Update and Light Mode Preferences Dropdown values are switching
### Bug description

Update Preference string dropdown button offering in the preferences offers only options from Light mode.

Image

Screencast:
https://github.com/user-attachments/assets/4e060e25-6e3b-4e96-bb1b-24ebba2d16fb

### Operating system

Windows 10

### Installation Method

None

### Version

1.14.0

### Steps to reproduce

1. Open Preferences -> appearance
2. Set the value to System or other (requires some playing)
3. Open Preferences -> Update Section
4. Check values on dropdown
5. Assert: startup/never
Actual: System, dark, light

It also works vice versa, if value is changed in update, the dropdown values changed in appearance.

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

See #9917 for background. Svelte reuses component instances whenever possible. The problem here is that the Dropdown options are being set via child option elements, created via buildOptions. When you switch pages all of the explicitly set properties and state are applied, but since options is not set as a property, nor state, it is not applied.

The simplest fix in this case is using the options parameter directly, which we do in other cases and causes Svelte to correctly treat it as state.

This is only avoiding the problem though, so I've opened #9952 for future follow up. I've checked and we have no other child options where this would currently happen, but the real fix is possibly a Svelte fix, removing child option support, or finding a fix directly in Dropdown.

### Screenshot / video of UI

N/A

Fixes #9917.

### How to test this PR?

See #9917.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesRenderingItemFormat.spec.ts
@@ -247,7 +247,6 @@ test('Expect a select when record is type string and has enum values', async () 
   const input = screen.getByLabelText('record-description');
   expect(input).toBeInTheDocument();
   expect(input.children[0]).toHaveAttribute('name', 'record');
-  expect(input).toHaveTextContent('first');
 });
 
 test('Expect enum to have the givenValue selected', async () => {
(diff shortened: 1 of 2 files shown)


axel7083:
Some tests are failing

axel7083 (approved):
LGTM

cdrage (approved):
LGTM. Tested and it's updating correctly.

PR #9964, opened 2024-11-19
chore: exclude one rule not working with eslint v9.15.0

benoitf:
### What does this PR do?
sonarjs plugin is embedding its own version of the typescript-eslint parser and not relying on a dependency.
so sometimes upgrades are failing when the parser differs from the runtime objects.

### How to test this PR?

linter check should be ✅ 


jeffmaury (approved):
LGTM

feloy (approved):
tested with eslint v9.15.0, works fine

PR #9981, opened 2024-11-19
chore: k8s - show pending PVCs as stopped icon

### LINKED ISSUE #9322: Applying a PVC that is purposely "pending" does not show
### Bug description

Apply a PVC:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: task-pv-claim
spec:
  storageClassName: manual
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 3Gi
```

See that it shows in the CLI:

```sh
▶ k get pvc
NAME                        STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
task-pv-claim               Pending                                                                        manual         <unset>                 3m37s
```

It then does not show within the PD k8s UI

### Operating system

mac m1

### Installation Method

None

### Version

1.11.1

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

cdrage:
chore: k8s - show pending PVCs as stopped icon

### What does this PR do?

* Pending PVCs show up as "green" / "active", when instead it should show
as "Stopped" as they are not yet in the Bound phase.
* Based upon the similar interface to Deployments

### Screenshot / video of UI

Screenshot 2024-11-19 at 9 56 40 AM

Closes https://github.com/podman-desktop/podman-desktop/issues/9322

### How to test this PR?

Deploy a PVC that is purposely pending:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: task-pv-claim
spec:
  storageClassName: manual
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 3Gi
```

See that it's status is "Stopped" in the UI.

### DIFF
--- a/packages/renderer/src/lib/pvc/pvc-utils.ts
+++ b/packages/renderer/src/lib/pvc/pvc-utils.ts
@@ -29,13 +29,14 @@ export class PVCUtils {
     // PVC's only have "Pending", "Bound" and "Lost" as phases.
     // However, we can also determine if it is in the middle of deleting by
     // checking if the phase is "Terminating" through checking if the object has deletionTimestamp.
+    //
+    // Note: If it is "Pending" it will be considered as "STOPPED" as it is not yet bound, this
+    // is similar to how we handle Deployment statuses.
     let status = 'STOPPED';
     if (pvc.metadata?.deletionTimestamp) {
       status = 'TERMINATING';
     } else if (pvc.status?.phase === 'Bound') {
       status = 'RUNNING';
-    } else if (pvc.status?.phase === 'Pending') {
-      status = 'STARTING';
     } else if (pvc.status?.phase === 'Lost') {
       status = 'DEGRADED';
     }
--- a/packages/renderer/src/lib/pvc/pvc-utils.ts
+++ b/packages/renderer/src/lib/pvc/pvc-utils.ts
@@ -30,15 +30,19 @@ export class PVCUtils {
     // However, we can also determine if it is in the middle of deleting by
     // checking if the phase is "Terminating" through checking if the object has deletionTimestamp.
     //
-    // Note: If it is "Pending" it will be considered as "STOPPED" as it is not yet bound, this
-    // is similar to how we handle Deployment statuses.
-    let status = 'STOPPED';
+    let status: string;
     if (pvc.metadata?.deletionTimestamp) {
       status = 'TERMINATING';
     } else if (pvc.status?.phase === 'Bound') {
       status = 'RUNNING';
     } else if (pvc.status?.phase === 'Lost') {
       status = 'DEGRADED';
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/pvc/PVCList.svelte
```diff
@@ -40,6 +40,7 @@ const pvcUtils = new PVCUtils();
 onMount(() => {
   return kubernetesCurrentContextPersistentVolumeClaimsFiltered.subscribe(value => {
     pvcs = value.map(pvc => pvcUtils.getPVCUI(pvc));
+    console.log('pvcs', pvcs);
```
axel7083:
```suggestion
```

cdrage: Woops!

packages/renderer/src/lib/pvc/pvc-utils.ts
```diff
@@ -29,13 +29,14 @@ export class PVCUtils {
     // PVC's only have "Pending", "Bound" and "Lost" as phases.
     // However, we can also determine if it is in the middle of deleting by
     // checking if the phase is "Terminating" through checking if the object has deletionTimestamp.
+    //
+    // Note: If it is "Pending" it will be considered as "STOPPED" as it is not yet bound, this
+    // is similar to how we handle Deployment statuses.
```
benoitf: I would prefer I think to have the else if entry, just assigning to status = 'STOPPED' in the else condition.

benoitf: or is it possible to have PENDING status ?

cdrage: it is by default set to `let status = 'STOPPED';` (see line 35), so didn't need to add an if statement.

We could add Pending, but  need to make changes to UI component as there is no "PENDING" only starting.

benoitf: yes it's implicit.

```ts
let status = 'STOPPED'
if (foo = 'bar') { // only if we don't want stopped
 then
}
```

I would prefer explicit way with if / elseif/elseif and else at the end
or using case with every type possible, so if someone introduce a new state, we need to update the function with the mapping being required
 ```ts
let status;
if (foo === '...') {
   status = ...
} else if
} else if

} else {
  // here the default
  status = 'STOPPED'
}
 ```

benoitf (approved):
suggestion (non blocking)

cdrage:
@benoitf updated with your suggestion

benoitf (approved):
thanks @cdrage

PR #9984, opened 2024-11-19
chore: api change, add platform option to pullImage

### LINKED ISSUE #9387: Be able to specify architecture when pulling an image.

When pulling an image that has multiple architectures (AMD64, ARM64, etc.), we cannot specify which architecture to search for / pull.

For example, we cannot pull an image with the API that's the equivalent of:

```
podman pull cdrage/helloworld --arch arm64
```

Modify:

```ts
    /**
     * Pull an image from a registry
     *
     * @param containerProviderConnection the connection to the local engine to use for pulling the image
     * @param imageName the name of the image to pull
     * @param callback the function called when new logs are emitted during the pull operation
     */
    export function pullImage(
      containerProviderConnection: ContainerProviderConnection,
      imageName: string,
      callback: (event: PullEvent) => void,
    ): Promise<void>;
```

To be able to pull a certain architecture as well. 

We have the "Arch" parameter we can add to the API: https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImagePullLibpod

_No response_

### Additional context

_No response_

cdrage:
chore: api change, add platform option to pullImage

### What does this PR do?

* Adds option to use `platform` for pulling images
* Incorporates to pullImage API where you simply specify platform:
  `linux/arm64`, etc.

### Screenshot / video of UI

N/A

Closes https://github.com/podman-desktop/podman-desktop/issues/9387

### How to test this PR?

1. Change this line:

   https://github.com/cdrage/podman-desktop/blob/fa8cf9f4ac636829483b4c23393dd5d6ee4f89ed/packages/renderer/src/lib/image/PullImage.svelte#L132-L133
   to: `await window.pullImage(selectedProviderConnection,
   imageToPull.trim(), callback, 'linux/arm64');`
2. Pull an image that has a different architecture than your host
   (example, arm64). Good example is `https://quay.io/podman/hello`
3. See that the image is pulled successfully.
4. Inspect the image within PD and see it is arm64.

### DIFF
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -1220,12 +1220,9 @@ test('pull unknown image fails with error 403', async () => {
   );
 });
 
-test('pulling an image with platform=linux/arm64 will add the platform to the pull options', async () => {
-  const getMatchingEngineFromConnectionSpy = vi.spyOn(containerRegistry, 'getMatchingEngineFromConnection');
-  //vi.spyOn(containerRegistry, 'pullImage').mockReturnValue(Promise.resolve());
-
+test('pulling an image with platform linux/arm64 will add platform to pull options', async () => {
+  // Mock the pulling and dockerode
   const pullMock = vi.fn();
-
   const fakeDockerode = {
     pull: pullMock,
     modem: {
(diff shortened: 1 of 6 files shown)


packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -1220,6 +1220,37 @@ test('pull unknown image fails with error 403', async () => {
...
+  getMatchingEngineFromConnectionSpy.mockReturnValue(fakeDockerode);
+
+  const containerConnectionInfo = {} as ProviderContainerConnectionInfo;
+
+  pullMock.mockResolvedValue({ statusCode: 201 });
+
+  const callback = vi.fn();
+
+  // Ignore no floating promises warning as we just care about the pullMock call, not the promise
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
benoitf: issue: we should not ignore this error

could we just add await before the containerRegistry.pullImage call ?

benoitf (approved):
could you please not ignore the linter rule in the test before merging

axel7083 (approved):

packages/main/src/plugin/container-registry.spec.ts
cdrage: I thought this too, but it get's "stuck" when doing await, as if it's waiting for something to complete, despite me having the call back to resolve. 

I'll try another way of doing the test

packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -1220,6 +1220,39 @@ test('pull unknown image fails with error 403', async () => {
...
+  const containerConnectionInfo = {} as ProviderContainerConnectionInfo;
+
+  pullMock.mockResolvedValue({ statusCode: 201 });
+
+  const callback = vi.fn();
+
+  // Expect pull image to pass
+  // we use .catch to avoid an unhandled promise rejection, we don't care about the result, we just
+  // want to check that the call was made with the correct platform
+  containerRegistry.pullImage(containerConnectionInfo, 'unknown-image', callback, 'linux/arm64').catch(() => {});
```
benoitf: it's still odd to me why the mocking is failing ? It shouldn't throw any exception

Just calling await and that's all

cdrage: Can't :( we time out no matter what mocking I do, so suggested we just do a `catch` with it.

cdrage:
@benoitf Spent all day figuring it out before realizing that `followProgress` is mocked incorrectly in most of our code.

Instead of:
```
    modem: {
      followProgress: vi.fn(),
    },
```

It should be:
```
  vi.spyOn(fakeDockerode.modem, 'followProgress').mockImplementation((_s, f, _p) => {
    return f(null, []);
  });
 ```
 
 So it officially does not timeout (was getting 5 second timeout issues).
 
 Tests have now been updated and is good to go!

benoitf (approved):

benoitf:
@cdrage glad you found out the reason !

PR #10003, opened 2024-11-20
chore(deps-dev): bump vitest to v2.1.5

benoitf:
### What does this PR do?
update to v2.1.5

### How to test this PR?


axel7083 (approved):

axel7083:
Hello dependabo... euhh Florent !

gastoner (approved):

benoitf:
I don't know why dependabot doesn't want to update this dependency, maybe it's not working well...

benoitf:
so, the root cause is that afterAll is not found in a setupFiles

trying 2.2.0-beta2 it works, so I would assume it's a bug only in 2.1.5, closing

cbr7 (approved):

benoitf:
ok so fixing transitive dependencies, it should work now

Issue #9818, opened 2024-11-07
Checking connectivity of Kubernetes contexts


feloy:
### Epic domain

Today, connectivity of Kubernetes clusters is checked directly by running informers on "primary" informers, which are informers on Pods and Deployments resources.

Informers are too complex structures and are not designed for checking connectivity (they are primarily designed for running in-cluster, where there is very little connectivity problem between nodes and API).

We are also starting informers on resources we qre not sure the user has access to (for example users don't have access to Node information on Openshift clusters, or on Routes on non-Openshift clusters).

We will instead:
- use the health check endpoint of clusters to check connectivity (only a few times, then stop)
- use a request on `SelfSubjectRulesReview` resource to check which resources the user has access to
- start informers on resources the user has access to only

We need to be sure that all this work, including the load of the Kubeconfig file (which can be heavy if relying on third-party tools to renegociate auth tokens, etc) is done asynchronously and does not block the startup of the UI

- [x] https://github.com/podman-desktop/podman-desktop/issues/9938
- [x] https://github.com/podman-desktop/podman-desktop/issues/8394
- [x] https://github.com/podman-desktop/podman-desktop/issues/7629
- [x] https://github.com/podman-desktop/podman-desktop/issues/9778
- [x] https://github.com/podman-desktop/podman-desktop/issues/9879

PR #10018, opened 2024-11-20
feat: introduce an hidden option kubernetes.statesExperimental

feloy:

### What does this PR do?

Introduce an hidden option kubernetes.statesExperimental

As the epic #9818 will introduce many changes in the way the states of the contexts are managed, I propose to introduce an hidden preference `kubernetes.statesExperimental` (false by default) 
to work on this epic without removing the current way the states are managed.

Part of #9818

### How to test this PR?

To test the new version, set this preference in `~/.local/share/containers/podman-desktop/configuration/settings.json`:

```
```

For now, the new version does nothing. The state of the context is always `UNKNOWN` in Preferences > Kubernetes, and there are no Kubernetes resources in Nodes / Deployments / ...

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
@@ -20,7 +20,7 @@ import type { KubeConfig, KubernetesObject } from '@kubernetes/client-node';
 
 import type { ContextGeneralState, ResourceName } from '/@api/kubernetes-contexts-states.js';
 
-export class ContextsManagerV2 {
+export class ContextsManagerExperimental {
   async update(_kubeconfig: KubeConfig): Promise<void> {}
 
   getContextsGeneralState(): Map<string, ContextGeneralState> {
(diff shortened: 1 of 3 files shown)


benoitf (commented):
I would use the experimental pattern rather than v2

as there are other issues where we would like to centralize all "experimental' stuff

axel7083 (commented):
It would be helpful to have an elaborated plan of how this will replace the existing ContextsManager.

This look like a full rewrite of the ContextsManager, which is a 1000 lines file. Maybe we should take this opportunity to create some diagrams to discuss the abstract of how components will interact, depends on each other etc.

You listed some steps in the epic https://github.com/podman-desktop/podman-desktop/issues/9818

> We will instead:
> - use the health check endpoint of clusters to check connectivity (only a few times, then stop)
> - use a request on SelfSubjectRulesReview resource to check which resources the user has access to
> - start informers on resources the user has access to only

IMHO this is not explicit enough, and this does not help the review nor the understanding of what you are planning here.

For example a problem in the ContextManager is that it only handle informers for current context, and switching between contexts is an heavy operation, that we may want to avoid in the future, since this may be a full rewrite, it might be nice to pause and ask ourself how we want to design this for our later requirements.

benoitf (approved):

axel7083 (approved):

PR #10032, opened 2024-11-20
refactor: replace router goto containers by handle navigation

### LINKED ISSUE #8294: chore(renderer): components should uses the navigation api

Currently we have both the navigation api 

https://github.com/podman-desktop/podman-desktop/blob/c0b1b75fad8429e8603338ec3be504f6ec3c5e4a/packages/renderer/src/navigation.ts#L37

which makes it easier to centralise the  path definition. However we still have a lot of places using manually the `router.goto` for path which are already defined in the navigation api.

We should start makes them uses it.

image

:warning: There are still a lot of path which are not supported by the navigation api.

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Refactor the `router.goto('/containers/*` to use the handleNavigation

Part of https://github.com/podman-desktop/podman-desktop/issues/8294

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
+++ b/packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
@@ -5,6 +5,8 @@ import { onDestroy, onMount } from 'svelte';
 import type { Unsubscriber } from 'svelte/store';
 import { router } from 'tinro';
 
+import { handleNavigation } from '/@/navigation';
+import { NavigationPage } from '/@api/navigation-page';
 import type { ProviderContainerConnectionInfo, ProviderInfo } from '/@api/provider-info';
 
 import type { PodCreatePortOptions } from '../../../../main/src/plugin/dockerode/libpod-dockerode';
@@ -291,7 +293,7 @@ function getWarningText(): string {
 
       <div class="w-full grid justify-items-end mt-5">
         <div>
-          <Button type="link" on:click={() => router.goto('/containers')}>Close</Button>
+          <Button type="link" on:click={handleNavigation.bind(undefined, { page: NavigationPage.CONTAINERS })}>Close</Button>
           <Button
             icon={SolidPodIcon}
             bind:disabled={createInProgress}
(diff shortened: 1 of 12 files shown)


benoitf (approved):
nitpick: the notation method.bind(undefined, params) is far from being user-friendly
so I would prefer a new function navigateToLogs and just use `on:click={navigateToLogs}`

axel7083:
> method.bind(undefined, params)

I really like this notation xD

packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
```diff
@@ -291,7 +293,7 @@ function getWarningText(): string {
 
       <div class="w-full grid justify-items-end mt-5">
         <div>
-          <Button type="link" on:click={() => router.goto('/containers')}>Close</Button>
+          <Button type="link" on:click={handleNavigation.bind(undefined, { page: NavigationPage.CONTAINERS })}>Close</Button>
```
jeffmaury: Not sure to get why bind is required here

axel7083: To avoid 

```ts
on:click={() => handleNavigation({ page: NavigationPage.CONTAINERS })}
```

I find it nicer to have a bind

jeffmaury:
> > method.bind(undefined, params)
> 
> I really like this notation xD

Seems you're alone 😄 as I did not catch the logic as well

axel7083:
> Seems you're alone 😄 as I did not catch the logic as well

I removed it.

ℹ️ I like the `Function#bind` method because it create a new method without enclosure, you can also create partial function with bind, IMO this is a nice tool

packages/renderer/src/lib/pod/PodmanPodDetailsSummary.svelte
```diff
@@ -50,7 +52,7 @@ if (pod) {
       {#each pod.containers as container}
         <tr>
           <DetailsSubtitle>
-            <Link on:click={() => router.goto(`/containers/${container.Id}/logs`)}>
+            <Link on:click={handleNavigation.bind(undefined, {page: NavigationPage.CONTAINER_LOGS,parameters: {id: container.Id }})}>
```
benoitf: looks like it's not fully removed

axel7083: now it should be good!

benoitf (approved):

PR #10055, opened 2024-11-21
chore: fix spaces in experimental section website

Firewall:
Signed-off-by: Matt Demyttenaere <mdemytte@redhat.com>

### What does this PR do?
Adds spaces around the links in the experimental section

### Screenshot / video of UI
After:
image

Fixing this: 
Before
image


benoitf (approved):
not sure it's 100% the way to go but looks like this hack is already used in other part of the website so here is my approval

Firewall:
> like this hack is already used in other part of the website

Yes I copied it from a bit higher in the same file

cdrage (approved):

Issue #10054, opened 2024-11-21
Npm target test:e2e does not work on windows using default system Shell (cmd or powershell)


odockal:
### Bug description

Running E2E tests on windows using npm target `test:e2e` is limited by an os configuration, this command in particular: `npx playwright test tests/playwright/src/specs/ --grep-invert '@update-install|@podman-remote'"`. On linux and mac all works ok, but on windows the reg exp expression is processed with cmd or powershell and simply tries to run podman-remote as a command and fails on error. 

Related issue: https://github.com/microsoft/playwright/issues/9682

### Operating system

Windows

### Installation Method

None

### Version

next (development version)

### Relevant log output

```shell

```

### Additional context

I tried multiple ways to do it using either grep or grep-invert. Without success. I am deciding to use the easiest way, adding `@all` tag into tests we want to be run in all tests suite.

PR #10057, opened 2024-11-21
chore(test): add @all tag for all tests that are run as part of test:e2e

odockal:
### What does this PR do?
Add @all tag to all e2e tests that should be run on `test:e2e`. And to fix problem with processing grep reg. exp on various platforms by different shells.
#10054 

### How to test this PR?

`pnpm test:e2e` should run all expected tests on all platforms. no failures.

### DIFF
--- a/tests/playwright/src/specs/deploy-to-kubernetes.spec.ts
+++ b/tests/playwright/src/specs/deploy-to-kubernetes.spec.ts
@@ -76,7 +76,7 @@ test.skip(
   'Tests suite should not run on Linux platform',
 );
 
-test.describe.serial('Deploy a container to the Kind cluster', { tag: ['@smoke', '@all'] }, () => {
+test.describe.serial('Deploy a container to the Kind cluster', { tag: ['@k8s_e2e', '@all'] }, () => {
   test('Pull an image and start a container', async ({ navigationBar }) => {
     const imagesPage = await navigationBar.openImages();
     const pullImagePage = await imagesPage.openPullImage();
--- a/tests/playwright/src/specs/deploy-to-kubernetes.spec.ts
+++ b/tests/playwright/src/specs/deploy-to-kubernetes.spec.ts
@@ -76,7 +76,7 @@ test.skip(
   'Tests suite should not run on Linux platform',
 );
 
-test.describe.serial('Deploy a container to the Kind cluster', { tag: '@k8s_e2e' }, () => {
+test.describe.serial('Deploy a container to the Kind cluster', { tag: ['@smoke', '@all'] }, () => {
   test('Pull an image and start a container', async ({ navigationBar }) => {
     const imagesPage = await navigationBar.openImages();
     const pullImagePage = await imagesPage.openPullImage();
--- a/tests/playwright/src/specs/kind.spec.ts
+++ b/tests/playwright/src/specs/kind.spec.ts
@@ -57,7 +57,7 @@ test.afterAll(async ({ runner, page }) => {
   }
 });
 
-test.describe.serial('Kind End-to-End Tests', { tag: '@k8s_e2e' }, () => {
+test.describe.serial('Kind End-to-End Tests', { tag: ['@smoke', '@all'] }, () => {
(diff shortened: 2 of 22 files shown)


tests/playwright/src/specs/deploy-to-kubernetes.spec.ts
```diff
@@ -76,7 +76,7 @@ test.skip(
   'Tests suite should not run on Linux platform',
 );
 
-test.describe.serial('Deploy a container to the Kind cluster', { tag: '@k8s_e2e' }, () => {
+test.describe.serial('Deploy a container to the Kind cluster', { tag: ['@smoke', '@all'] }, () => {
```
cbr7: Why did you remove the k8s_e2e tag and added smoke?

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -57,7 +57,7 @@ test.afterAll(async ({ runner, page }) => {
   }
 });
 
-test.describe.serial('Kind End-to-End Tests', { tag: '@k8s_e2e' }, () => {
+test.describe.serial('Kind End-to-End Tests', { tag: ['@smoke', '@all'] }, () => {
```
cbr7: Same as above.

tests/playwright/src/specs/kubernetes-edit-yaml.spec.ts
```diff
@@ -79,7 +79,7 @@ test.skip(
   'Tests suite should not run on Linux platform',
 );
 
-test.describe.serial('Kubernetes Edit YAML Feature E2E Test', { tag: '@k8s_e2e' }, () => {
+test.describe.serial('Kubernetes Edit YAML Feature E2E Test', { tag: ['@smoke', '@all'] }, () => {
```
cbr7: same as above.

tests/playwright/src/specs/kubernetes.spec.ts
```diff
@@ -87,7 +87,7 @@ test.skip(
   'Tests suite should not run on Linux platform',
 );
 
-test.describe('Kubernetes resources End-to-End test', { tag: '@k8s_e2e' }, () => {
+test.describe('Kubernetes resources End-to-End test', { tag: ['@smoke', '@all'] }, () => {
```
cbr7: same as above.

cbr7 (changes requested):
I don't really understand why we need an @all tag at all, if you want to run all the tests you can just select the folder where all the test files are located, that will run all the tests. Adding a tag to all test files means that you don't need to add it to any test file.

odockal:
@cbr7 Because:
> Running E2E tests on windows using npm target test:e2e is limited by an os configuration, this command in particular: `npx playwright test tests/playwright/src/specs/ --grep-invert '@update-install|@podman-remote'"`. On linux and mac all works ok, but on windows the reg exp expression is processed with cmd or powershell and simply tries to run podman-remote as a command and fails on error.

I would love to keep it as it is, but I simply haven't found a way to make it work universally on all platforms. I can try a couple more reg exps and escaping if it fill all, but nothing worked for me me so far.

cbr7:
@odockal Wouldn't it be just simpler to move the update and podman-remote e2e tests into another folder? There is no particular reason they need to be under `/specs/`, you can move them to `tests/playwright/src/special_tests/<custom folder for each test suite here>` and that will solve all problems permanently, since you will call always whichever folder you want to run, without any params at all.

odockal:
> @odockal Wouldn't it be just simpler to move the update and podman-remote e2e tests into another folder? There is no particular reason they need to be under `/specs/`, you can move them to `tests/playwright/src/special_tests/<custom folder for each test suite here>` and that will solve all problems permanently, since you will call always whichever folder you want to run, without any params at all.

that might work!

odockal:
Closing as better and simpler solution proposed.

odockal:
New PR opened to have clear review: https://github.com/podman-desktop/podman-desktop/pull/10060

Issue #9351, opened 2024-10-12
Podman Desktop does not see all docker contexts


tony-sol:
### Bug description

In preferences, on "Docker Compatibility" page, there is only one default docker context to select - `default (unix:///var/run/docker.sock)`

![Image](https://github.com/user-attachments/assets/157c9811-7c9e-4dc9-b280-513d5f4c4383)

! version 1.13.2 (not present in version selector bellow)

### Operating system

macos Sequoia 15.0.1

### Installation Method

Brew (macOS)

### Version

next (development version)

### Steps to reproduce

1. Create several docker contexts, e.g. using lima or orbstack
2. Update to latest version available via self update (it's 1.13.2 in my case)
3. In Podman Desktop go to preferences, enable `Experimental (Docker Compatibility)` (Docker Compatibility in `Extension: Podman` must be off)
4. See Docker Context list (`Select the active Docker CLI context:`) has only single default context

tony-sol:
Bug is probably here https://github.com/containers/podman-desktop/blob/main/packages/main/src/plugin/docker/docker-context-handler.ts#L86

I guess it should either respect `DOCKER_CONFIG` env variable or call `docker` binary itself, to correctly follow through configuration to the actual contexts

benoitf:
Thanks for the report

tony-sol:
Also, find another related issue (i guess it's) connected: system socket exists, but not recognized

![Image](https://github.com/user-attachments/assets/250ca4f5-0f70-4f45-8e63-d494a81efdd8)

benoitf:
@tony-sol what is the output of

```bash
curl --unix-socket /var/run/docker.sock "http://localhost/_ping"
```

tony-sol:
Sorry, looks like it was my mistake, after reboot lima vm it fetched a bit of info from `docker system info`

![Image](https://github.com/user-attachments/assets/77786166-1f63-4f0e-93fb-87d1215a6d62)

UI seems to be odd in my scenario and i guess i found why: my docker.sock is provied by lima virtual machine, so [this](https://github.com/podman-desktop/podman-desktop/blob/d686f0441c82e93b69f8175f3c6d8f1f98b44122/packages/main/src/plugin/docker/docker-compatibility.ts#L71) is not designed to be used with anything other than docker desktop or podman

And sadly, i don't really know how to improve this, here is a bit context which may be useful to improve compatibility feature (i really, really like this, so i hope it become ever more cool)
```
[...]

PR #10059, opened 2024-11-21
fix: fetch all docker contexts

tony-sol:
### What does this PR do?

Fixes Docker Compatibility settings, by improving fetching docker context respecting `DOCKER_HOST` and `DOCKER_CONTEXT` env variables

### Screenshot / video of UI

https://github.com/user-attachments/assets/97e565f8-e10a-495b-aac6-142344c367da

https://github.com/podman-desktop/podman-desktop/issues/9351

### How to test this PR?

> [!IMPORTANT]
> all "export" steps means, that env variable will be exported in *rc, *env or *profile files
> (e.g.: .bashrc, .bash_profile, .zshrc, .zshenv, .zprofile, .profile, etc.)
> and may require Podman Desktop restart for reload shell state

- create multiple docker contexts via `docker context create`
- verify that all contexts displayed in UI
- export `DOCKER_CONFIG` with new any location, e.g. `~/some/new/dir/`
- move `~/.docker` directory to `DOCKER_CONFIG`, e.g. `~/some/new/dir/{contexts/, config.json}`
- verify that all contexts still displayed in UI
- export `DOCKER_HOST` with host different from your current context host
- verify that all contexts still displayed in UI and current context changed to defined via `DOCKER_HOST`
- export `DOCKER_CONTEXT` with other context name
- verify that all contexts still displayed in UI and current context changed to defined via `DOCKER_CONTEXT`

### DIFF
--- a/packages/main/src/plugin/docker/docker-context-handler.ts
+++ b/packages/main/src/plugin/docker/docker-context-handler.ts
@@ -36,19 +36,43 @@ export interface DockerContextParsingInfo extends Omit<DockerContextInfo, 'isCur
  * Handle the `docker context`, allowing to list them and switch between them.
  */
 export class DockerContextHandler {
-  protected getDockerConfigPath(): string {
-    return join(homedir(), '.docker', 'config.json');
+  protected getDockerConfig(): string {
+    // DOCKER_CONFIG is defines path to docker client configuration files
+    // Docker uses ~/.docker if DOCKER_CONFIG is not set
+    // see: https://docs.docker.com/reference/cli/docker/ for more details
+    const dockerConfig = process.env['DOCKER_CONFIG'];
+    if (dockerConfig) {
+      return dockerConfig;
+    }
+    return join(homedir(), '.docker');
   }
 
-  protected async getCurrentContext(): Promise<string> {
+  protected getDockerContext(): string | undefined {
+    return process.env['DOCKER_CONTEXT'];
+  }
+
+  protected getDockerHost(): string | undefined {
+    return process.env['DOCKER_HOST'];
+  }
+
+  protected getDockerConfigFile(): string {
+    return join(this.getDockerConfig(), 'config.json');
+  }
+
+  protected async getCurrentContextFromConfiguration(): Promise<string> {
+    const dockerContext = this.getDockerContext();
+    if (dockerContext) {
+      return dockerContext;
+    }
+
     let currentContext: string = 'default';
 
     // if $HOME/.docker/config.json exists, read it and get the current context
-    const dockerConfigExists = existsSync(this.getDockerConfigPath());
+    const dockerConfigExists = existsSync(this.getDockerConfigFile());
     if (dockerConfigExists) {
       // read the file and get the current context
       // if there is a current context, add it to the list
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/docker/docker-context-handler.ts
```diff
@@ -36,19 +36,43 @@ export interface DockerContextParsingInfo extends Omit<DockerContextInfo, 'isCur
  * Handle the `docker context`, allowing to list them and switch between them.
  */
 export class DockerContextHandler {
-  protected getDockerConfigPath(): string {
-    return join(homedir(), '.docker', 'config.json');
+  protected getDockerConfig(): string {
+    // DOCKER_CONFIG is defines path to docker client configuration files
+    // Docker uses ~/.docker if DOCKER_CONFIG is not set
+    // see: https://docs.docker.com/reference/cli/docker/ for more details
+    const dockerConfig = process.env['DOCKER_CONFIG'];
```
benoitf, line 43: the problem is that we can't read the env variable.

on macOS, the applications are started without having access to the env variables (that you will define in your login script)

tony-sol: right, it completely slipped my mind :c

and i guess i understood, why it works in test suite - if i run ui via `pnpm watch` to test, it inherits all env from my shell

sooo, i think we have two options here: either call `docker` via `processApi.exec()` or add another option into settings with `DOCKER_CONFIG`, `DOCKER_HOST` and `DOCKER_CONTEXT`

tony-sol: i mean, if second: besides reading envs, fetch values from settings (settings would have priority over envs)

tony-sol: ...but first is easier to implement

benoitf: I guess that if you call docker cli it will not have the environment variable as well so the CLI won't find it neither

tony-sol: yeah, it slipped either

so, i'll make a few options in Settings->Preferences->Docker Compatibility, right after "Enabled" checkbox

good?

benoitf: I think we will need to move the logic of this part out of podman desktop core and put it in the docker extension and here we will be able to put the preferences

Need to think on how it can be integrated well through the new docker compatibility section

benoitf:
setting to draft, as, after discussing it, we may need to move some code to another location/refactor before that.

packages/main/src/plugin/docker/docker-context-handler.ts
tony-sol: @benoitf may we use same workaround as [here](https://github.com/podman-desktop/podman-desktop/pull/10086) to make this work?

benoitf: yes but the issue is that it needs to be part of the docker extension, not of Podman Desktop core

benoitf:
replaced by https://github.com/podman-desktop/podman-desktop/pull/10555

PR #10062, opened 2024-11-21
fix: proxy settings do not take effect when switched to system or disabled

### LINKED ISSUE #9625: Proxy settings do not take effect when switched to system or disabled

jeffmaury:
Fixes #9625

### What does this PR do?

Ensure events are correctly sent when switching proxy state

### Screenshot / video of UI

N/A

#9625 

### How to test this PR?

Update proxy settings

### DIFF
--- a/packages/main/src/plugin/proxy.ts
+++ b/packages/main/src/plugin/proxy.ts
@@ -156,18 +156,19 @@ export class Proxy {
       newProxy = await getProxySettingsFromSystem(this);
     }
 
-    // update configuration
-    if (newProxy) {
-      // update
-      this.proxySettings = newProxy;
+    // update
+    this.proxySettings = newProxy;
 
+    if (newProxy) {
       // notify
       this._onDidUpdateProxy.fire(newProxy);
-      const proxyConfiguration = this.configurationRegistry.getConfiguration('proxy');
-      await proxyConfiguration.update('http', newProxy.httpProxy);
-      await proxyConfiguration.update('https', newProxy.httpsProxy);
-      await proxyConfiguration.update('no', newProxy.noProxy);
     }
+
+    // update configuration
+    const proxyConfiguration = this.configurationRegistry.getConfiguration('proxy');
+    await proxyConfiguration.update('http', newProxy?.httpProxy);
+    await proxyConfiguration.update('https', newProxy?.httpsProxy);
+    await proxyConfiguration.update('no', newProxy?.noProxy);
   }
 
   get proxy(): ProxySettings | undefined {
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/proxy.ts
```diff
@@ -178,15 +179,16 @@ export class Proxy {
...
 
     // update configuration
     const proxyConfiguration = this.configurationRegistry.getConfiguration('proxy');
-    proxyConfiguration.update('enabled', state).catch((err: unknown) => {
-      console.error('Error updating proxy state', err);
-    });
+    await proxyConfiguration.update('enabled', state);
 
+    if (state === ProxyState.PROXY_SYSTEM) {
+      await this.setProxy({} as ProxySettings);
```
benoitf: I think setProxy should accept undefined as a parameter and here we call setProxy(undefined)

benoitf (approved):
issue(non-blocking): my POV is that we should not assign new values to the parameters of a function (immutable)
probably another rule to enforce in the linter

code is OK

nitpick: the title of the PR needs to be fixed (it ends with 'dis...')

benoitf:
the rule is https://eslint.org/docs/latest/rules/no-param-reassign

benoitf (approved):
thx for the changes

need to fix the title of the PR before merging

benoitf:
@jeffmaury do we need to backport the fix to 1.14.x branch ?

jeffmaury:
I would not qualifier as a blocker as it's a specific use case but would be a nice plus

benoitf:
@Mergifyio backport 1.14.x

PR #10064, opened 2024-11-22
fix: do console logging only if app is not quitting

SoniaSandler:

### What does this PR do?
This PR possibly fixes the javascript epipe error that sometimes pops up when closing PD with `ctrl + c` after `pnpm watch` by logging in the console only if the app is not quitting.

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -303,11 +303,11 @@ export class PluginSystem {
         .then(extName => {
           if (extName) args.unshift(extName);
 
-          // still display as before by invoking original method
-          originalConsoleMethod(...args);
-
           // but also send the content remotely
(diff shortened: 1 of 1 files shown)


benoitf (commented):
it means we no longer have logs when the app is quitting

benoitf (approved):

SoniaSandler:
Oh, I see. I'm not sure what kind of logs can happen on quitting except for the stopped extensions, but if that's the case, maybe we should open it for discussion or I can try and look for another solution

benoitf:
I'm fine merging

jeffmaury (approved):
I'm wondering if the whole code block should be protected because I remember having the an error on web content sender but seems to fix the issue

Issue #9092, opened 2024-09-25
Task created by extension are not cancellable 


axel7083:
### Bug description

We have in the API, a `cancellable` option 

https://github.com/containers/podman-desktop/blob/208b2a07fc172213fc5b3ee143e463a24fbfe597/packages/extension-api/src/extension-api.d.ts#L1195

However, since the refactor, we may have lost the capabilities of cancelling a task. (did we have before ? I don't remember)

### Operating system

Fedora 40

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

In an extension

```ts
import { window, ProgressLocation } from '@podman-desktop/api';

await window.withProgress({
  location: ProgressLocation.TASK_WIDGET,
  title: `Long running task`,
  cancellable: true,
}, (progress, token) => {
  token.onCancellationRequested(() => {
    console.log('request cancel');
  });
  return // todo long running task
});
````

Even after defining `cancellable` the task do not appear to be cancellable in the TaskManager

![image](https://github.com/user-attachments/assets/c54bdac5-68dc-4f80-b6f4-33876c857d5a)

### Additional context

Potential fix

### Quick fixes

We currently have an action that extension can define, the `ProgressImpl` could create a `Cancel` action when the `withProgress` is called with `cancellable: true`.

However this would overwrite any route that the extension would have defined.

### Better fix

We could add a `CancellationTokenImpl` that we provide to the withProgress task to the `CancellationTokenRegistry` and add the cancellation token id to the task, we would add a `cancellableTokenId` or `cancelId`  or any new field, allowing the frontend to call `window.cancelToken` to cancel a given task.

This would not change any api, and would require a small change in the `TaskManagerItem.svelte` to support an additional `cancel` optional button next to the action.

**Comment by axel7083** [comment]
cc @benoitf an opinion?

**Comment by benoitf** [comment]
I don't know why there is rush on discussion for this task but what I can say for sure is that it shouldn't use quickfix provided solution

it seems also the doc is stating:
```
     * To monitor if the operation has been cancelled by the user, use the provided {@linkcode CancellationToken}.
     * Note that currently only `ProgressLocation.Notification` is supporting to show a cancel button to cancel the
     * long-running operation.
```

while I only see
```
  export enum ProgressLocation {
    APP_ICON
    TASK_WIDGET
  }
```

PR #10065, opened 2024-11-22
chore: introduce cancellable field in the tasks

benoitf:
### What does this PR do?
if cancellable attribute is provided, it registers the token in the registry so frontend could ask
to cancel this specific token

if a task is cancellable, it contains a cancellationTokenSourceId attribute
this is the attribute that the can use to cancel a task.

This PR is not providing any frontend code to cancel the task

### How to test this PR?

### DIFF
--- a/packages/api/src/taskInfo.ts
+++ b/packages/api/src/taskInfo.ts
@@ -35,4 +35,6 @@ export interface TaskInfo {
   error?: string;
   progress?: number;
   action?: string;
+  cancellable: boolean;
+  cancellationTokenSourceId?: number;
 }
--- a/packages/main/src/plugin/tasks/task-manager.ts
+++ b/packages/main/src/plugin/tasks/task-manager.ts
@@ -113,11 +113,23 @@ export class TaskManager {
     this.setStatusBarEntry(true);
   }
 
-  public createTask(options?: { title?: string; action?: TaskAction }): Task {
+  public createTask(options?: {
+    title?: string;
+    action?: TaskAction;
+    cancellable?: boolean;
+    cancellationTokenSourceId?: number;
+  }): Task {
     this.taskId++;
 
     const task = new TaskImpl(`task-${this.taskId}`, options?.title ? options.title : `Task ${this.taskId}`);
     task.action = options?.action;
+    task.cancellable = options?.cancellable ?? false;
+    if (task.cancellable && !options?.cancellationTokenSourceId) {
+      throw new Error('cancellable task requires a cancellationTokenSourceId');
+    }
+    if (options?.cancellationTokenSourceId) {
+      task.cancellationTokenSourceId = options?.cancellationTokenSourceId;
+    }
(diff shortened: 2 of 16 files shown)


packages/api/src/taskInfo.ts
```diff
@@ -35,4 +35,6 @@ export interface TaskInfo {
   error?: string;
   progress?: number;
   action?: string;
+  cancellable: boolean;
+  cancellationTokenSourceId?: number;
```
axel7083, line 39: we could just have the `cancellationTokenSourceId` ? if it is define, the task is cancellable ?

benoitf: I started and thought about using only this pattern but it was not so clear when consuming

packages/main/src/plugin/tasks/task-manager.ts
```diff
@@ -113,11 +113,23 @@ export class TaskManager {
...
+    cancellationTokenSourceId?: number;
+  }): Task {
     this.taskId++;
 
     const task = new TaskImpl(`task-${this.taskId}`, options?.title ? options.title : `Task ${this.taskId}`);
     task.action = options?.action;
+    task.cancellable = options?.cancellable ?? false;
+    if (task.cancellable && !options?.cancellationTokenSourceId) {
+      throw new Error('cancellable task requires a cancellationTokenSourceId');
+    }
```
feloy, line 129: Is the other way possible, to have a `cancellationTokenSourceId` but not `cancellable`?

benoitf: no

feloy: in this case, IMO you would have to add the check, or not ask for `cancellable` from the consumer and have `cancellable = !! cancellationTokenSourceId`

benoitf: updated

feloy (approved):

axel7083 (approved):
LGTM, works nicely 👍 

Sorry for the review delay, I wanted to test it fully !

To test it, I added the following to the activate function of the podman extension

```ts
  extensionApi.window.withProgress({
    title: `Task Example Cancellable`,
    cancellable: true,
  }, (progress, token) => {
    return new Promise((resolve, reject) => {
      setTimeout(resolve, 20_000);
      token.onCancellationRequested(reject);
    });
  });
```

Then inside the `TaskManagerItem.svelte` I added the following (testing purpose)

```svelte
{#if task.cancellable && task.cancellationTokenSourceId}
      <button on:click={() => window.cancelToken(task.cancellationTokenSourceId ?? 0)}>Cancel</button>
{/if}
```

and I can see clearly that the cancel action works as expected 

https://github.com/user-attachments/assets/619bafd6-cd4d-4ddf-a8b9-e9ff4cf0bd58

Great work, this is nice ! 🚀

Issue #10052, opened 2024-11-21
vscode focus border is a weird blue


axel7083:
### Bug description

We have some blue lines at top and bottom of the editor

![Image](https://github.com/user-attachments/assets/444610fa-8cfc-4583-a9cb-dafa6ac4abff)

This is controled by the `--vscode-focusBorder #007fd4` , I think we should set this to `transparent`

This is the first time I noticed it, but they seems to have some issue open on their side https://github.com/microsoft/monaco-editor/issues/4455 ?

### Operating system

Windows 11

### Installation Method

Installer from website/GitHub releases

### Version

next (development version)

### Steps to reproduce

1. goto any resource and check for a monaco editor

### Relevant log output

```shell

```

benoitf:
I would say this issue is not urgent, we have other bugs with an higher priority in the backlog

axel7083:
>I would say this issue is not urgent, we have other bugs with an higher priority in the backlog

I encounter that while working on https://github.com/podman-desktop/podman-desktop/issues/8845, that's why, this is very ugly :(

PR #10073, opened 2024-11-22
feat(k8s): adding create resource page

### LINKED ISSUE #8845: `Apply yaml` should be more flexible

image

I usually have the yaml I want to apply in my clipboard, clip on the `Apply Yaml` and sadly it opens a file picker.

I think we should allow providing either an url, a plain text input or a file picker to make the `Apply yaml` action more interesting to use

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

In every Kubernetes page add a new `Create Resource` page.

> ℹ️ Might be better (follow-up PR) to have something like `KubernetesAction`, because the `KubeApplyYAMLButton` is also on every page, and this is a lot of repeated code to add the create button 🤔 

### Notable logic

If you deploy only one kubernetes resource (one pod, one service etc.) after clicking on `Create resource` you will be redirected to the corresponding podman-desktop page of the resource.

If you deploy multiple resources, you will be redirected to `$lastPage`.

### Known issue

The blue focus on the monaco editor see https://github.com/podman-desktop/podman-desktop/issues/10052 for the fix

### Screenshot / video of UI

image

**Error handling**

image

Fixes https://github.com/podman-desktop/podman-desktop/issues/8845
 
### How to test this PR?

**Manually**

1. checkout and pnpm watch
2. assert have a kubernetes cluster running
3. go to any kubernetes page (except pods)

4. click on `Create Resource`

image

5. Copy paste any Kubernetes YAML (example bellow)
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    image: docker.io/library/nginx:1.14.2
    ports:
    - containerPort: 80
```

6. click on `Create Resource`

image

### DIFF
--- a/packages/renderer/src/lib/kube/KubeActions.svelte
+++ b/packages/renderer/src/lib/kube/KubeActions.svelte
@@ -1,5 +1,15 @@
 <script lang="ts">
+import { Button } from '@podman-desktop/ui-svelte';
+
 import KubeApplyYamlButton from '/@/lib/kube/KubeApplyYAMLButton.svelte';
+import { handleNavigation } from '/@/navigation';
+import { NavigationPage } from '/@api/navigation-page';
+
+function navigateToKubernetesResourceCreate(): void {
+  return handleNavigation({ page: NavigationPage.KUBERNETES_RESOURCE_CREATE });
+}
 </script>
 
 <KubeApplyYamlButton />
+<Button title="Create Kubernetes Resource" on:click={navigateToKubernetesResourceCreate}>Create Resource</Button>
+
--- a/packages/renderer/src/lib/kube/KubeActions.svelte
+++ b/packages/renderer/src/lib/kube/KubeActions.svelte
@@ -1,15 +1,8 @@
 <script lang="ts">
-import { Button } from '@podman-desktop/ui-svelte';
-
 import KubeApplyYamlButton from '/@/lib/kube/KubeApplyYAMLButton.svelte';
-import { handleNavigation } from '/@/navigation';
-import { NavigationPage } from '/@api/navigation-page';
-
-function navigateToKubernetesResourceCreate(): void {
-  return handleNavigation({ page: NavigationPage.KUBERNETES_RESOURCE_CREATE });
-}
+import KubeCreateResourceButton from '/@/lib/kube/KubeCreateResourceButton.svelte';
 </script>
 
 <KubeApplyYamlButton />
-<Button title="Create Kubernetes Resource" on:click={navigateToKubernetesResourceCreate}>Create Resource</Button>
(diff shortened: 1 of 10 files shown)


packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretList.svelte
```diff
@@ -151,6 +153,7 @@ const row = new TableRow<ConfigMapSecretUI>({ selectable: _configmapSecret => tr
 <NavPage bind:searchTerm={searchTerm} title="configmaps & secrets">
   <svelte:fragment slot="additional-actions">
     <KubeApplyYamlButton />
+    <Button on:click={handleNavigation.bind(undefined, { page: NavigationPage.KUBERNETES_RESOURCE_CREATE })}>Create Resource</Button>
```
benoitf: issue: blocking: can we have a pre-req PR that would refactor the code and introduce a component like `<KubeAdditionalActions>` or introduce a `KubernetesNavPage` with the additional actions included there or any solution like that to avoid to copy/paste the exact same lines across 6-7 pages when we need to add a button in all the kube pages ?

axel7083: Yeah, sure, will do that!

packages/renderer/src/lib/kube/KubeActions.svelte
```diff
@@ -1,5 +1,15 @@
...
+import { handleNavigation } from '/@/navigation';
+import { NavigationPage } from '/@api/navigation-page';
+
+function navigateToKubernetesResourceCreate(): void {
+  return handleNavigation({ page: NavigationPage.KUBERNETES_RESOURCE_CREATE });
+}
 </script>
 
 <KubeApplyYamlButton />
+<Button title="Create Kubernetes Resource" on:click={navigateToKubernetesResourceCreate}>Create Resource</Button>
```
benoitf: enhancement: I would expect a `<KubeCreateResourceButton>` so we can reuse it somewhere else as well (like it's done for ApplyYaml)

axel7083: Done ✅

benoitf (commented):
it works as expected.

It does not prevent to merge but I think there is something to improve on the UX

I first click on the top right corner "Create Resource" and it opens the page to create my resource
and I need to click on the exact same location to proceed but I have the feeling that if I do that I will just enter again into the same page


usually the action button is below the text, and the redirect are at the top right

for example if I edit the yaml file, the actions button are at the top bottom

image


I would expect to have the same kind of buttons when creating the resource ?

benoitf:
outdated

PR #10074, opened 2024-11-22
chore: patch kubernetes-client with health class

feloy:
Signed-off-by: Philippe Martin <phmartin@redhat.com>

### What does this PR do?

Patch the `@kubernetesclient-node` import with the PR https://github.com/kubernetes-client/javascript/pull/2035

### What issues does this PR fix or reference?

Part of #9938

Issue #9938, opened 2024-11-18
Use health check to check Kubernetes cluster connectivity


feloy:
### Is your enhancement related to a problem? Please describe

See https://github.com/podman-desktop/podman-desktop/issues/9818

### Describe the solution you'd like

Use the healthcheck endpoint(s) to check cluster connectivity

### Additional context

There is currently no function to check  the health of a cluster in the Kubernetes library, but seems easy to implement, and there is the opportunity to provide a patch for the library: https://github.com/kubernetes-client/javascript/issues/693

### TODO

- [x] backend part: #10076 
- [ ] dispatch to frontend: https://github.com/podman-desktop/podman-desktop/pull/10349
- [ ] frontend part: https://github.com/podman-desktop/podman-desktop/pull/10349

feloy:
added to the sprint 266 as it is a pre-requisite of #7629

feloy:
Proposed a PR on kubernetes library to handle health checks: https://github.com/kubernetes-client/javascript/pull/2035

PR #10076, opened 2024-11-22
feat: check connectivity with health check

feloy:
### What does this PR do?

check connectivity with health check, in the context of experimental kubernetes states,
and saves the state in a `ContextsConnectivityRegistry`.

The next step is to dispatch the state to the frontend.

(The first commit of the PR is a copy of the one in https://github.com/podman-desktop/podman-desktop/pull/10074)

### Screenshot / video of UI

backend only

Part of #9938 

### How to test this PR?

```
```

You can have some cluster behind a VPN to provoke some timeout and check by editing the kubeconfig file several times quickly (by changing the current context for example), that previous checks are aborted and only a response from the last (un-aborted) check is received.

Some logs can be added to the `ContextsConnectivityRegistry` to display the state each time it is changed


packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,66 @@
...
+export interface CheckOptions {
+  // timeout in ms
+  timeout?: number;
+}
+
+type ReadinessCallback = (ready: boolean) => void;
+
+// HealthChecker checks the readiness of a Kubernetes context
+// by requesting the readiness endpoint of its server
+export class HealthChecker {
```
axel7083: We should make it extend Disposable, so we can recycle all of that easily (abort)

packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,66 @@
...
+  #kubeConfig: KubeConfig;
```
axel7083: do we really need to keep a ref to the kubeconfig here ? Can't we just create the Health object and call `readyz` when we need ?

packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,66 @@
...
+  #abortController: AbortController;
+  #onReadinessCallback: ReadinessCallback = (_ready: boolean) => {};
```
axel7083: EventEmitter would be nice, might need more than one subscriber in the future

packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  public abort(): void {
+    this.#abortController.abort();
+  }
+
+  // onReadiness sets the function to call when the checkReadiness has received a response
+  // If called several times, the previous callback is overwritten
+  public onReadiness(callback: ReadinessCallback): void {
+    this.#onReadinessCallback = callback;
+  }
```
axel7083: not needed, we could use an EventEmitter to manage that for us

packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,66 @@
...
+    const health = new Health(this.#kubeConfig);
+    try {
+      const result = await health.readyz({ signal: this.#abortController.signal, timeout: opts?.timeout });
+      this.#onReadinessCallback(result);
+    } catch (err: unknown) {
+      if (err instanceof AbortError) {
+        // do nothing
+      } else {
+        this.#onReadinessCallback(false);
+      }
```
axel7083:
```suggestion
      if (!(err instanceof AbortError)) {
        this.#onReadinessCallback(false);
      }
```

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -16,12 +16,57 @@
...
+      return;
+    }
+
+    // Abort previous health checkers and remove them from registry
+    for (const [contextName, healthChecker] of this.#healthCheckers.entries()) {
+      healthChecker.abort();
+      this.#healthCheckers.delete(contextName);
+    }
+
+    this.#kubeConfigCheck = JSON.stringify(kubeconfig);
```
axel7083: stringify is an heavy operation, don't do it twice, you are already doing this line 42

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -48,4 +93,16 @@ export class ContextsManagerExperimental {
   dispose(): void {}
```
axel7083: abort all the health checks here

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -48,4 +93,16 @@ export class ContextsManagerExperimental {
...
 
   async refreshContextState(_contextName: string): Promise<void> {}
+
+  // getIsolatedKubeConfig returns a KubeConfig with only kubeContext, set as current one
+  private getIsolatedKubeConfig(kubeconfig: KubeConfig, kubeContext: Context): KubeConfig {
```
axel7083: I would rather avoid doing copy of copy of copy of kubeconfig, why the HealthCheck class cannot juste take another option such as `context: string` ?

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -16,12 +16,57 @@
...
+import { HealthChecker } from './health-checker.js';
+
+const HEALTH_CHECK_TIMEOUT_MS = 5_000;
+
 export class ContextsManagerExperimental {
-  async update(_kubeconfig: KubeConfig): Promise<void> {}
+  // kubeConfigCheck is used by `update` to check if the received kubeconfig has changed
+  // its value must not be used except for detecting this change
+  // Methods of this class should receive a kubeconfig from `update`
+  #kubeConfigCheck = JSON.stringify(new KubeConfig());
```
axel7083: I would really have a preference to avoid doing that and have a `compare(a: KubeConfig, b: KubeConfig)` method.

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
axel7083: After reading the code of the informers and https://github.com/kubernetes-client/javascript/pull/2035 we cannot specify a specific cluster, this is not nice IMO, but nothing we can't do atm

We seems to already doing that in the informers

https://github.com/podman-desktop/podman-desktop/blob/b1ff0c7b25c21330e07096a14ac091cf5f11ce64/packages/main/src/plugin/kubernetes/contexts-manager.ts#L205-L208

packages/main/src/plugin/kubernetes/health-checker.ts
feloy: yes, good point, I'll change

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
feloy: I'm not sure to understand what you propose.
But one thing important to know is that for the check, we don't only need the cluster information. We also need the user information (as the health check endpoint may require authorization)

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
feloy: This would need a deep comparison, and some fields are of type any (`User.exec?: any;` for example). I'm not sure it would be easy to find/write a robust-enough and efficient `compare` implementation.

I think that it could be a follow-up optimization

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
axel7083: Ye forgot what I said

axel7083 (approved):

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
benoitf: AFAIK we have events monitoring the kubeconfig file.

kubeconfig file modified = we receive a event, so I'm wondering why we would need a comparison there ?

packages/main/src/plugin/kubernetes/contexts-connectivity-registry.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  // setReachable saves in the registry if the context `contextName` is reachable
+  setReachable(contextName: string, reachable: boolean): void {
+    const connectivity = this.getConnectivity(contextName);
+    connectivity.reachable = reachable;
+    this.#connectivities.set(contextName, connectivity);
+  }
+
+  protected getConnectivity(contextName: string): ContextConnectivity {
+    return this.#connectivities.get(contextName) ?? { checking: false, reachable: false };
```
benoitf: I would expect to receive undefined if it's not there

packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,61 @@
...
+  onReadiness: Event<boolean> = this.#onReadinessEmit.event;
+
+  // builds an HealthChecker which will check the cluster of the current context of the given kubeConfig
+  constructor(kubeConfig: KubeConfig) {
+    this.#abortController = new AbortController();
+    this.#health = new Health(kubeConfig);
+  }
+
+  // start checking the readiness
+  public async checkReadiness(opts?: CheckOptions): Promise<void> {
```
benoitf: I have an issue on the wording.
Using check method, I would expect to have a boolean returned.

I would expect just a start() method ?

packages/main/src/plugin/kubernetes/contexts-connectivity-registry.ts
```diff
@@ -0,0 +1,51 @@
...
+    connectivity.checking = checking;
+    this.#connectivities.set(contextName, connectivity);
+  }
+
+  // setReachable saves in the registry if the context `contextName` is reachable
+  setReachable(contextName: string, reachable: boolean): void {
+    const connectivity = this.getConnectivity(contextName);
+    connectivity.reachable = reachable;
+    this.#connectivities.set(contextName, connectivity);
+  }
```
benoitf: reading it twice time, I had a hard to figure it out the pattern like why it's not atomic. I can set for example checking to true but maybe nobody is really checking or I could have values not consistent between checking and reachable.

IMHO we're mixing internal and external data.

so, on my end, I would expect to have something with a clean state about which part is doing what.

If I have an object that checks the state for the contexts, I expect to have a map with the context name being checked.

So each object handling something maintains the state of its own data.

Then, exporting the data for a client is something about reconciling data but there is no need to have an external registry. 

Here, I don't know who is really interacting with this registry.

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -16,12 +16,54 @@
...
+      this.#healthCheckers.set(contextName, healthChecker);
+      healthChecker.onReadiness((ready: boolean) => {
+        this.#contextsConnectivityRegistry.setChecking(contextName, false);
+        this.#contextsConnectivityRegistry.setReachable(contextName, ready);
+      });
+      this.#contextsConnectivityRegistry.setChecking(contextName, true);
+      healthChecker
+        .checkReadiness({ timeout: HEALTH_CHECK_TIMEOUT_MS })
+        .catch((err: unknown) => console.error(`error checking readiness for context ${contextName}: ${err}`));
+    }
```
benoitf: IMHO it's very complex to read.

I don't see why there is a checking thing in a separate 3rd party registry

AFAIK, if there is a health checker object attached, it means that this context is being monitored.

benoitf: and as we have one health checker by context, the health checker can just store the value of the the context
rather than here, having an event to update another map


we already have a map of health checkers
if context is in the map = we monitor the context and if it's reachable we have that information that could be inside the health checker object

benoitf (commented):
IMHO I don't see the need of the registry, and the event in the health checker as it can just keep the value


ExperimentalChecker:
 - map of healthzChecker
   for each checker there is inside the value if it's reachable or not

many disposables can be removed, and half of the code

packages/main/src/plugin/kubernetes/contexts-connectivity-registry.ts
axel7083: or throw an error

feloy:
> IMHO I don't see the need of the registry, and the event in the health checker as it can just keep the value
> 
> ExperimentalChecker:
> 
> * map of healthzChecker
>   for each checker there is inside the value if it's reachable or not
> 
> many disposables can be removed, and half of the code

I would like to validate that this pattern is valid with the idea that we want to send to the frontend all these states `Map<contextName: string, state: { checking, reachable }>` every time one of the state changes. In the current pattern, the registry is used to save this data ready to be sent, and as the object sending the data every time a `set` is called.

With this new pattern, we would also need some signal from the checkers to know when the state of a checker changes and to collect all the data from the checkers to send it.

feloy:
I agree that the `onReadiness(); checkReadiness();` is confusing, and that setting `checking` and `reachable` not atomically is confusing. 

IMHO, as the checkers have a short life (they are killed as soon as the kubeconfig is modified again), it seems not reasonable to keep the state of the check inside them, as they will be deleted / recreated every time the `update` is called.

Here is a new proposal by simplfying the `onReadiness / checkReadiness`  part and having an atomic  `setState` (prototype, tests need to be finalized)

The registry is very simple, but I expect to add code to send the data to the frontend in this registry.

@benoitf WDYT?

benoitf:
I think my issue is why having a separate registry. State should be closer to the object being checked.

> IMHO, as the checkers have a short life (they are killed as soon as the kubeconfig is modified again), it seems not reasonable to keep the state of the check inside them, as they will be deleted / recreated every time the update is called.


On my side I see no issue having the state part of the object responsible to check the state of a context.
Also why they would be recreated/deleted if their data is the same as before ?
I can update the configuration of another context
They're the object closest to the state of the cluster. (the 'isolated kubeconfig part') If this part is not changing no need to delete/recreate them.


If you don't want to keep it as part of the HealthCheck object, make another layer between the two, but not a side instance
ContextChecker instance would use a HealthCheck instance.

Also the HealthCheck class is dedicated to a specific context as it's using a kubeconfig object which is a context. So I think it's missing in the name that it's for a specific Context. HealthCheckContext.

> The registry is very simple, but I expect to add code to send the data to the frontend in this registry.

Yes it was my worries when I read the code. To me we should separate the inner state of the data vs what we want to expose publicly having the state the closest possible of the object maintaining it.

packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,55 @@
...
+  #abortController: AbortController;
+
+  // builds an HealthChecker which will check the cluster of the current context of the given kubeConfig
+  constructor(kubeConfig: KubeConfig) {
+    this.#abortController = new AbortController();
+    this.#health = new Health(kubeConfig);
+  }
+
+  // start checking the readiness
+  public async start(opts?: CheckOptions): Promise<boolean> {
```
benoitf: wondering if options are not part of the constructor

packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { KubeConfig } from '@kubernetes/client-node';
+import { AbortError, Health } from '@kubernetes/client-node';
+import type { Disposable } from '@podman-desktop/api';
+
+export interface CheckOptions {
```
benoitf: needs to add somehow the prefix 'HealthCheck' like HealthCheckOptions

packages/main/src/plugin/kubernetes/health-checker.ts
benoitf: now you changed it to return a boolean flag so the previous name was OK

start = if there is nothing being returned as before you changed the content

as it returns something, the previous checks is ok

packages/main/src/plugin/kubernetes/health-checker.ts
```diff
@@ -0,0 +1,55 @@
...
+  // timeout in ms
+  timeout?: number;
+}
+
+// HealthChecker checks the readiness of a Kubernetes context
+// by requesting the readiness endpoint of its server
+export class HealthChecker implements Disposable {
```
benoitf: KubeConfigHealthCheck ? (as it's taking a KubeConfig object)

feloy:
> I think my issue is why having a separate registry. State should be closer to the object being checked.
> 
> > IMHO, as the checkers have a short life (they are killed as soon as the kubeconfig is modified again), it seems not reasonable to keep the state of the check inside them, as they will be deleted / recreated every time the update is called.
> 
> On my side I see no issue having the state part of the object responsible to check the state of a context. Also why they would be recreated/deleted if their data is the same as before ? I can update the configuration of another context They're the object closest to the state of the cluster. (the 'isolated kubeconfig part') If this part is not changing no need to delete/recreate them.
> 

OK, I was checking if the whole kubeconfig is changed or not, but we can do the check at the `isolated kube config` level, and creating a health checker only if its specific context changes. This will make the checkers long-running, and make your pattern possible

I'll make the changes.

Thanks for your feedback

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -16,12 +16,54 @@
...
+
+    this.#kubeConfigCheck = kubeConfigToCheck;
+
+    for (const kubeContext of kubeconfig.getContexts()) {
+      const contextName = kubeContext.name;
+      const isolatedConfig = this.getIsolatedKubeConfig(kubeconfig, kubeContext);
+      const healthChecker = new HealthChecker(isolatedConfig);
+      this.#healthCheckers.set(contextName, healthChecker);
+
+      this.#contextsConnectivityRegistry.setState(contextName, true, false);
```
benoitf: IMHO we should not separate the code from the state.

ContextChecker --> define the state for a context (can be merged with HealthChecker)

so instead of having some code here like

```
       const healthChecker = new HealthChecker(isolatedConfig);
      this.#healthCheckers.set(contextName, healthChecker);
      this.#contextsConnectivityRegistry.setState(contextName, true, false);
```

only one object:
```
const contextChecker = new ContextWhateverName(contextName, isolatedConfig);
this.#contextCheckers.set(contextName, contextChecker);
```

Inside the contextChecker object, we know if there is a pending operation (HealthCheck object being instantiated)
last time the reachable has been found, if it's reachable


then in this class, we only need to destroy/recreate the checker object if the config for this context has been changed. 

need an event in the context checker to be notified when a context state is being updated

from that, we can compute what we would like to send to the frontend, can be chunks, full data, whatever, we don't really care as we will just build the data based on all the objects in the tree

packages/main/src/plugin/kubernetes/contexts-states-dispatcher.ts
```diff
@@ -0,0 +1,34 @@
...
+export class ContextsStatesDispatcher {
+  constructor(private manager: ContextsManagerExperimental) {}
+
+  init(): void {
+    this.manager.onContextHealthStateChange((_state: ContextHealthState) => this.update());
+    this.manager.onContextDelete((_state: DispatcherEvent) => this.update());
+  }
+
+  update(): void {
+    console.log('current health check states', this.manager.getHealthCheckersStates());
```
feloy, line 32: Not sure if we want to merge with this, but it can be useful for testing/reviewing, and I can remove it when LGTM'd

benoitf: ok for me to keep this trace

benoitf (approved):
thanks for all the improvements

it looks like easier to understand, and each part is doing and managing its own data, so in case of failure, easy to know which part is failing and it we want to enhance parts, contract being clear, we can easily swap one component by another

axel7083 (approved):
LGTM

PR #10088, opened 2024-11-25
chore: set status to canceled if a task has been canceled

benoitf:
### What does this PR do?
if a user ask to cancel a task, today it'll end in a task having the success status. With this commit, it will make the status being canceled which is more accurate.

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/tasks/progress-impl.ts
+++ b/packages/main/src/plugin/tasks/progress-impl.ts
@@ -135,11 +135,6 @@ export class ProgressImpl {
       action: this.getTaskAction(options),
     });
 
-    let cancelRequest = false;
-    cancellationToken.onCancellationRequested(() => {
-      cancelRequest = true;
-    });
-
     return task(
       {
         report: value => {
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/tasks/progress-impl.ts
```diff
@@ -135,6 +135,11 @@ export class ProgressImpl {
       action: this.getTaskAction(options),
     });
 
+    let cancelRequest = false;
+    cancellationToken.onCancellationRequested(() => {
+      cancelRequest = true;
```
axel7083: cancellationToken has a property `isCancellationRequested` instead of declaring a variable `cancelRequest`

packages/main/src/plugin/tasks/progress-impl.ts
```diff
@@ -150,7 +155,11 @@ export class ProgressImpl {
     )
       .then(value => {
         // Middleware to capture the success of the task
-        t.status = 'success';
+        if (cancelRequest) {
```
axel7083: question: shouldn't we throw an error since it is cancelled ? :/

benoitf: I would tend to think it's not an error. If I cancel something, I expect at the end it's canceled, not an error.
Error = unexpected thing while here I really ask for that.

packages/main/src/plugin/tasks/progress-impl.ts
benoitf: updated

axel7083 (approved):

PR #10089, opened 2024-11-25
chore: display canceled task with a different color

benoitf:
### What does this PR do?
change the color of canceled task

depends on 
- [x] : https://github.com/podman-desktop/podman-desktop/pull/10088

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
+++ b/packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
@@ -57,7 +57,7 @@ async function doExecuteAction(task: TaskInfo) {
 <div class="flex flew-row w-full py-2">
   <!-- first column is the icon-->
   <div class="flex w-3 {iconColor} justify-center">
-    <div class="align-top" role="img" aria-label="Icon status of task {task.id}">
+    <div class="align-top" role="img" aria-label="{task.status} icon of task {task.name}">
     <Fa size="0.875x" icon={icon} />
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/task-manager/TaskManagerItem.svelte
```diff
@@ -52,7 +57,9 @@ async function doExecuteAction(task: TaskInfo) {
 <div class="flex flew-row w-full py-2">
   <!-- first column is the icon-->
   <div class="flex w-3 {iconColor} justify-center">
+    <div class="align-top" role="img" aria-label="Icon status of task {task.id}">
```
axel7083: does not make much sense to have the task id in the aria label? ids are internal, task name would make more sense about accessibility

benoitf: task name then ?

benoitf: I was wondering if I should display the status as part of the label because it won't help much to know there is an icon

axel7083: Maybe yes, or just `task icon status`, I am also not sure we need the `role=img` as the <Fa> component already has it (from my memory, need to be fact checked)

benoitf: updated

axel7083 (approved):

PR #10098, opened 2024-11-25
feat(feedback): adding support to get sys info

axel7083:
### What does this PR do?

Backend part of https://github.com/podman-desktop/podman-desktop/issues/9585.

The goal is to have a `include my system information`

image

That will prefill the GitHub bug issue with the system info.

(not included in this PR, only backend)

image

### Sidenote

Getting a short and usable system information through the node api is very annoying actually, depending on the platform we do not get the same information.

I added documentation above each function to precise how, and why we use certain node:os function for each platform.

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/feedback-handler.spec.ts
+++ b/packages/main/src/plugin/feedback-handler.spec.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { arch, release, version } from 'node:os';
+import { release } from 'node:os';
 
 import { shell } from 'electron';
 import { beforeEach, describe, expect, test, vi } from 'vitest';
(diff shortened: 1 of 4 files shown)


benoitf (commented):
instead of having the pattern:
```
getDarwinInfo()
   if (!isMac) { error}
   return

getLinuxInfo()
   if (!isInfo) { error}
   return

```

Use the following pattern

``` 
 
 if (isLinux()) {
   this.grabInfo = new LinuxGrabInfo();
 } else if (isDarwin()) {
  this.grabInfo = new DarwinGrabInfo();
  
  
  
 getOsInfo(): string {
    return this.grabInfo.getInfo()
}

axel7083:
> instead of having the pattern:
> ...
> Use the following pattern
> ...

Yes sure! Updated

benoitf:
hmm it's not really what I asked, everything here is in the same file, you can still internally call the other protected methods which is not expected.

you need to instantiate once the specific object/class, then, all callers can call the method, and they'll never be able to call another method that they should not be able to call

Also splitting across multiple chunks: you import only what you need, you have test only for a subpart, etc.

axel7083:
Oh yeah, this is not what I understood sorry, will fix

benoitf (approved):
thanks, it's much more robust 👍

PR #10100, opened 2024-11-25
feat: allow to cancel a task that is cancellable

### LINKED ISSUE #9092: Task created by extension are not cancellable

benoitf:
### What does this PR do?
Get a cancel icon in the task manager to be able to cancel a task flagged as cancellable.

### Screenshot / video of UI

https://github.com/user-attachments/assets/30d55ecb-2bf2-4947-a45d-10056cc7b670

fixes https://github.com/podman-desktop/podman-desktop/issues/9092

### How to test this PR?

1. Install the extension from `quay.io/fbenoit/long-task-example:v1.0``
2. press `F1` key to bring the command palette
3. select `Dummy Long task` (type dummy it will highlight the task)
4. click enter
5. Now open the task manager, see the task progressing and click on the cancel button to cancel it

### DIFF
--- a/packages/renderer/src/lib/task-manager/LegacyTaskManagerItem.spec.ts
+++ b/packages/renderer/src/lib/task-manager/LegacyTaskManagerItem.spec.ts
@@ -19,7 +19,7 @@
 import '@testing-library/jest-dom/vitest';
 
 import { fireEvent, render, screen } from '@testing-library/svelte';
-import { expect, test, vi } from 'vitest';
+import { beforeAll, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import type { TaskInfo } from '/@api/taskInfo';
 
@@ -45,6 +45,16 @@ const IN_PROGRESS_TASK_2: TaskInfo = {
   cancellable: false,
 };
 
+const CANCELLABLE_TASK: TaskInfo = {
+  id: '1',
+  name: 'Cancellable Task 1',
+  state: 'running',
+  status: 'in-progress',
+  started,
+  cancellable: true,
+  cancellationTokenSourceId: 1,
+};
+
 const CANCELED_TASK: TaskInfo = {
   id: '1',
   name: 'Canceled Task 1',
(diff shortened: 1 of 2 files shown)


axel7083:
Amazing job! Very nice to see the frontend come all together!

Do you think we could include the `cancel` action on the status bar ? (Here is an example of what I mean)

image

benoitf:
I wanted to ask UX people where to put the cancel icon on every implementation (panel, status, toast)

benoitf:
I mean, now it's very easy to handle it, just need to put an icon somewhere :-)

axel7083:
> I wanted to ask UX people where to put the cancel icon on every implementation (panel, status, toast)

You want to wait the next UX call and put this PR on hold?

slemeur:
We'll need other tasks to add the cancel action in the status bar, or in the notifications. For now, I think this is a great start. We will have the rework of the task panel coming up shortly where the cancel action will be easier to trigger.

benoitf:
Everything can be follow-up, it's where we put the cursor.

axel7083 (approved):
I am okey with merging this and add the cancel action on other component (etc. status bar) LGTM

benoitf:
Discussing the subject, it has been decided to not add the ability to cancel a task from the toast, but yes from the status bar

here is the follow-up for the status bar: https://github.com/podman-desktop/podman-desktop/pull/10100

odockal:
@benoitf I would add an issue for e2e test to cover the feature, or do you think component tests are enough?

benoitf:
unit tests are ok but you can still have a e2e workflow

I would wait for the new task manager for that

benoitf:
pr rebased as there was a conflict

PR #10116, opened 2024-11-26
feat(feedback): adding frontend checkbox to include system info

axel7083:
### What does this PR do?

This PR is the follow up of https://github.com/podman-desktop/podman-desktop/pull/10098 (which added the backend support) to collect the system-info of the host machine.

### Screenshot / video of UI

image

**demo**
https://github.com/user-attachments/assets/1de482ab-1b0b-4d68-8ac6-2ec57f48a526

> ℹ️ design not updated

Part of https://github.com/podman-desktop/podman-desktop/issues/9585

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.spec.ts
+++ b/packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.spec.ts
@@ -21,7 +21,7 @@ import '@testing-library/jest-dom/vitest';
 import { render, type RenderResult } from '@testing-library/svelte';
 import userEvent from '@testing-library/user-event';
 import { type Component, type ComponentProps } from 'svelte';
-import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
+import { beforeAll, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import type { FeedbackCategory } from '/@api/feedback';
 
@@ -57,6 +57,7 @@ function renderGitHubIssueFeedback(props: ComponentProps<typeof GitHubIssueFeedb
   title: HTMLInputElement;
   description: HTMLTextAreaElement;
   preview: HTMLButtonElement;
+  includeSystemInfo?: HTMLElement;
 } & RenderResult<Component<ComponentProps<typeof GitHubIssueFeedback>>> {
   const { getByRole, queryByTitle, ...restResult } = render(GitHubIssueFeedback, props);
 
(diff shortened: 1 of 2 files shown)


benoitf:
issue(blocking): checkbox is far away from the text

image

non blocking, follow-up: related to the other PR (follow-up), I guess we should use macOS rather than Darwin when doing the report. It was not obvious when reviewing the other PR but I think it'll be more meaningful to have macOS in the name
placeholder being:
image


image


issue:(follow-up or part of this PR)
Having the fields being completed by the link, I see one issue is that for example it doesn't set the version
image

I'm using .next version but then clicking on the link it selected 1.14.1 so I may open bugs for a wrong version

axel7083:
> issue(blocking): checkbox is far away from the text

Fixed ✅ 

image

> non blocking, follow-up: related to the other PR (follow-up), I guess we should use macOS rather than Darwin when doing the report.

I created https://github.com/podman-desktop/podman-desktop/issues/10118 and assigned myself 

> issue:(follow-up or part of this PR)
> Having the fields being completed by the link, I see one issue is that for example it doesn't set the version

If it was working I would do as follow up, but this is not possible:

The following url:

https://github.com/podman-desktop/podman-desktop/issues/new?template=bug_report.yml&title=hello&version=1.13.2 

Do not allow us to set the version. I opened an issue on the GitHub report, the selected value is always the latest...
We could fix that by changing the dropdown to a text input..

benoitf (approved):
if you could update the screenshot before merging

benoitf:
it seems the vertical align of the checkbox and the text is not the same as in other part of the UI, maybe needs to be checked before the merge

packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.spec.ts
```diff
@@ -109,7 +115,6 @@ test('Expect Preview on GitHub button to be disabled if there is no title or des
     expect(preview).toBeEnabled();
   });
 });
-
```
gastoner: Formated or missing new line?

axel7083:
> it seems the vertical align of the checkbox and the text is not the same as in other part of the UI, maybe needs to be checked before the merge

This seems to originate inside the `Checkbox` component

https://github.com/podman-desktop/podman-desktop/blob/516922e2ef043c208d55e9921edda79f7798aa20/packages/ui/src/lib/checkbox/Checkbox.svelte#L29

image

Not something that can be addressed in this PR.

gastoner (approved):
LGTM, tested on Windows

packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.spec.ts
axel7083: yes! fixed that, nice catch 👍

Issue #7629, opened 2024-06-13
Use equivalent of `kubectl auth can-i` on Kube contexts to check connectivity instead of informers


feloy:
### Is your enhancement related to a problem? Please describe

The current implementation for checking the connectivity to Kubernetes contexts uses informers. This solution has some drawbacks:
- the informer is not able to detect that the cluster is unreachable; Podman Desktop needs to define a timeout after which the context is considered unreachable, which complicates and makes the result less accurate
- Podman Desktop doesn't know if the user has the credentials to watch the resources, but we are always watching all resources. The user having limited credentials may result in wrong information (for example for nodes, which are generally not readable in shared clusters, for example OpenShift Sandbox)

### Describe the solution you'd like

I propose to query the [`SelfSubjectRulesReview`](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-rules-review-v1/) resource in the cluster to check the connectivity to the cluster, before to start informers.
The advantages are:
- the request is using a POST HTTP operation, with which Podman Desktop can detect connectivity error (still using some timeout)
- the `SelfSubjectRulesReview` are expected to be accessible for all users
- Podman Desktop can use the information to start informers only on the resources the user can watch. Ultimately, we could also check the `list` permission, to have a degraded implementation using `list` operation and not `watch` (informers), if the user has `list` but not `watch` permissions.

### TODO

- [x] backend part: #10133  
- [x] dispatch to frontend: #10383
- [ ] frontend part: #10655

feloy:
cc @deboer-tim @cdrage

cdrage:
👍 Would really help improvement. I'm assuming this works fine with OpenShift too?

feloy:
> 👍 Would really help improvement. I'm assuming this works fine with OpenShift too?

Yes, I have made a quick PoC, and we have access to the information on OpenShift Sandbox

jeffmaury:
kubernetes client used groups to deal with resource manipulations but noticed we were processing Routes even on non OpenShift clusters so how do this relates ? Does it includes non standard CRDs ?

feloy:
> kubernetes client used groups to deal with resource manipulations but noticed we were processing Routes even on non OpenShift clusters so how do this relates ? Does it includes non standard CRDs ?

Yes, the request includes CRDs, so we could be able to know if routes extension exists or not, and we could behave accordingly

feloy:
[...]

Issue #8348, opened 2024-08-05
Provide the ability to select kubernetes namespaces


slemeur:
### Is your enhancement related to a problem? Please describe

As a developer who is using Podman Desktop and the kubernetes features, I'm usually working with multiple namespaces. Unfortunately, with Podman Desktop, I can't select the namespace I want to visualize.

### Describe the solution you'd like

It would be nice to allow users to select the namespace to visualize the objects deployed within their Kubernetes environment.

One solution would be to allow a dropdown here:
<img width="958" alt="Screenshot 2024-08-05 at 15 07 07" src="https://github.com/user-attachments/assets/d4b72853-f496-4ea5-adca-ecc04d8f92df">

Also, it would be important to display which namespace is currently selected in the kubernetes selector + in the screen allowing to interact with kubernetes environments.

### Additional context

Linking to: https://github.com/containers/podman-desktop/issues/7191

afbjorklund:
You can change it from the command-line, I guess that was "good enough" for the feature

* https://github.com/containers/podman-desktop/issues/4990

shipsing:
![image](https://github.com/user-attachments/assets/6e48f4d3-10c3-4642-814d-5b95488d354a)

We have this field `Kubernetes Namespace` in the `Deploy to Kubernetes` page. I changed the namespace from `default` to  `kube-public`. I think user has the ability to change namespaces from the UI. We have mentioned the same in  [our doc](https://podman-desktop.io/docs/kubernetes/deploying-a-pod-to-kubernetes) on step no-4.
Also, when i change the kube context from `kind-cluster` to `dev-sandbox`, i do not see this `Kubernetes Namespace` field. So, seems like some sort of discrepancy.

Are we now going to provide this edit namespace ability from the Settings > Kubernetes section as well??

deboer-tim:
@shipsing this screen is only changing the target for one deployment, and if you pick a namespace that you don't have a context for there's no easy way to view it from within Podman Desktop today.

Yes, the request above is specifically suggesting to allow editing it in the Kubernetes section.

Bjego:
Yes this would be super helpfull - have a look into the vscode kubernetes plugin there you can select the namespace in the statusbar next to the cluster - maybe this could be a valid approach here as well

evanshortiss:
Would love to see this feature happen. I ran this issue with viewing my Pods in Kind this week.

Firewall:
All sub-issues closed. Coming with 1.18! 🚀

PR #10133, opened 2024-11-27
feat: check permissions on resources

feloy:

### What does this PR do?

The kubernetes manager (experimental) checks the permissions to watch resources, to be able to start informers only on permitted resources.

The permission check starts as soon as the health check is done on each kube context.

For now, only the "pods" and "deployments" resources are checked, on all kube contexts.

> Note that for non standard resources (like Openshift routes), the permission check will return true, even if the resource does not exist in the API. A call to the discovery API will be necessary, 

The check is done by using the underlying resource [SelfSubjectAccessReview](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-access-review-v1/). This resource supports to do a request for all resources/groups or for a specific resource/group.

We use this capability to first do a single request for `*/*` (all resources in all groups). This is optimal for `kind` (and probably other local) clusters, as the permission in done for all.

In case of denial for this `*/*`, a series of more granular requests are done, one for each resource.

> Note that we may add a step, to request for all resources in each group: `*/apps`, `*/core`, etc, before to ask for each specific resource.

### Screenshot / video of UI

Backend only

Part of #7629 

### How to test this PR?

```
```

### DIFF
--- a/packages/main/src/plugin/kubernetes/context-permissions-checker.spec.ts
+++ b/packages/main/src/plugin/kubernetes/context-permissions-checker.spec.ts
@@ -1,0 +1,298 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import util from 'node:util';
+
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+
+import type { ContextResourcePermission } from './context-permissions-checker.js';
+import { ContextPermissionsChecker } from './context-permissions-checker.js';
+import type { KubeConfigSingleContext } from './kubeconfig-single-context.js';
+
+describe('ContextPermissionsChecker is built with a non recursive request', async () => {
+  let permissionsChecker: ContextPermissionsChecker;
+  const createSelfSubjectAccessReviewMock = vi.fn();
+  const onPermissionResultCB = vi.fn();
+
+  let kubeConfig: KubeConfigSingleContext;
+
+  const attrs = {
+    namespace: 'ns1',
+    group: '*',
+    resource: '*',
+    verb: 'watch',
+  };
+  const resources = ['resource1', 'resource2'];
+
+  beforeEach(async () => {
+    vi.clearAllMocks();
+    kubeConfig = {
+      getKubeConfig: vi.fn().mockReturnValue({
+        makeApiClient: vi.fn().mockReturnValue({
+          createSelfSubjectAccessReview: createSelfSubjectAccessReviewMock,
+        }),
+      }),
+    } as unknown as KubeConfigSingleContext;
+  });
+
+  describe('permission is allowed', async () => {
+    beforeEach(async () => {
+      createSelfSubjectAccessReviewMock.mockResolvedValue({
+        status: {
+          allowed: true,
+          reason: 'a reason',
+        },
+      });
+      permissionsChecker = new ContextPermissionsChecker(kubeConfig, {
+        attrs,
+        resources,
+      });
+      permissionsChecker.onPermissionResult(onPermissionResultCB);
+      await permissionsChecker.start();
+    });
+
+    test('onPermissionResult is fired with permitted true', async () => {
+      expect(onPermissionResultCB).toHaveBeenCalledWith({
+        attrs,
+        resources,
+        permitted: true,
+        reason: 'a reason',
+      });
+    });
+
+    test('getPermissions returns permitted true', async () => {
+      const permissions = permissionsChecker.getPermissions();
+      const expected = new Map<string, ContextResourcePermission>();
+      expected.set('resource1', {
+        attrs,
+        permitted: true,
+        reason: 'a reason',
+      });
+      expected.set('resource2', {
+        attrs,
+        permitted: true,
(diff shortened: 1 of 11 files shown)


packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -61,29 +68,77 @@ export class ContextsManagerExperimental {
...
 
   private onDelete(state: DispatcherEvent): void {
-    const previous = this.#healthCheckers.get(state.contextName);
-    previous?.dispose();
+    const previousHealthChecker = this.#healthCheckers.get(state.contextName);
+    previousHealthChecker?.dispose();
     this.#healthCheckers.delete(state.contextName);
+    const previousPermissionsChecker = this.#permissionsCheckers.get(state.contextName);
+    previousPermissionsChecker?.dispose();
+    this.#permissionsCheckers.delete(state.contextName);
```
axel7083, line 131: why using the wording previous, you are deleting stuff, they are not yet deleted you are deleting them, why previous?

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -93,6 +148,15 @@ export class ContextsManagerExperimental {
     return result;
   }
 
+  /* getPermissions returns the current permissions */
+  getPermissions(): Map</* contextName */ string, Map</* resource */ string, ContextResourcePermission>> {
```
axel7083, line 152: Please create a interface for the result something with worded accessors, `Map<string, Map<string, ContextResourcePermission>>` is not friendly;

Ideally when dealing with kubernetes cluster I want objects, for example if I can list pods I want to do something like

```ts
// checking if I can access pods in a given namespace
KubernetesClient#getClusterConnection(context)#getPermission({ resource: "pods", namespace: "default" }): ContextResourcePermission
// checking all the permissions
KubernetesClient#getClusterConnection(context)#listPermissions(): ContextResourcePermission[]
```

packages/main/src/plugin/kubernetes/context-permissions-checker.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+          };
+        } else if (util.isDeepStrictEqual(param.body.spec.resourceAttributes, attrsResource1)) {
+          return {
+            status: {
+              allowed: true,
+              reason: 'a reason 1',
+            },
+          };
+        }
+        console.log('==> ', param.body.spec.resourceAttributes);
```
benoitf: remove this trace ?

benoitf (approved):

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
axel7083: We don't need to Map resource to V1ResourceAttributes since it contains the resource name

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
feloy: good point, I'll remove the prefix

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
feloy: I would prefer not focus too much on this part for now, it will most probably change when we will dispatch this data to the frontend, in a follow-up PR.

The current structure is to have a clear idea of the data in the `console.log` called in the state-dispatcher

gastoner (approved):
Codewise LGTM

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -61,29 +68,77 @@ export class ContextsManagerExperimental {
...
+
+      const namespace = state.kubeConfig.getNamespace();
+      const newPermissionChecker = new ContextPermissionsChecker(state.kubeConfig, {
+        attrs: {
+          namespace,
+          group: '*',
+          resource: '*',
+          verb: 'watch',
+        },
+        resources: ['pods', 'deployments'],
```
axel7083, line 91: Why are we only checking the permissions for pods and deployments ?

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
```diff
@@ -61,29 +68,77 @@ export class ContextsManagerExperimental {
...
+    const newHealthChecker = new ContextHealthChecker(event.config);
+    this.#healthCheckers.set(event.contextName, newHealthChecker);
+    newHealthChecker.onStateChange(this.onStateChange.bind(this));
+
+    newHealthChecker.onReachable(async (state: ContextHealthState) => {
+      // register and start permissions checker
+      const previousPermissionsChecker = this.#permissionsCheckers.get(state.contextName);
+      previousPermissionsChecker?.dispose();
+
+      const namespace = state.kubeConfig.getNamespace();
```
axel7083, line 83: This is not scalable, when we will address https://github.com/podman-desktop/podman-desktop/issues/8348 we cannot just update the kubeconfig, delete all informers, and restart everything with the health check etc.

I am not sure how this should be addressed, but if we can avoid rewriting everything in 2 months would be nice.

The system should from top permissions to bottom

1. Can I list nodes ?
2. Can I list namespace ? Yes ? which one ?
3. Can I list everything on namespace X ? No ? then list one by one each resources ? can we simplify the number of calls ?

Moreover the health check should be for a cluster specific, not a namespace, here we are mixing everything

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
feloy: One step at a time. For now, these are the resources needed to provide information on contexts

packages/main/src/plugin/kubernetes/contexts-manager-experimental.ts
feloy: Informers are started on a specific namespace (for namespaced resources), if the user changes the namespace, I don't think there is a way to keep the informers.

Health checks cannot be executed on cluster, as authentication (part of the user) is necessary. Because of this, we need to work at the level of the contexts.

I agree that we can optimize by running the health checks on merging the contexts ignoring their namespace, but I don't think the current architecture is blocking making this optimization.

axel7083 (approved):

PR #10137, opened 2024-11-27
refactor(ui): creating kube actions component

axel7083:
### What does this PR do?

https://github.com/podman-desktop/podman-desktop/pull/10073 will add a new actions to kube, which will be displayed in each page, we can simplify by creating a shared `KubeActions` component, to ease adding/removing actions in the future.

### Screenshot / video of UI

N/A

Related to https://github.com/podman-desktop/podman-desktop/issues/8845

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/kube/KubeActions.svelte
+++ b/packages/renderer/src/lib/kube/KubeActions.svelte
@@ -1,0 +1,7 @@
+<script lang="ts">
+import KubeApplyYamlButton from '../kube/KubeApplyYAMLButton.svelte';
+</script>
+
+<div class="flex flex-row gap-x-2">
+  <KubeApplyYamlButton />
+</div>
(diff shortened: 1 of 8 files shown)


packages/renderer/src/lib/kube/KubeActions.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+import KubeApplyYamlButton from '../kube/KubeApplyYAMLButton.svelte';
```
benoitf: suggestion (non-blocking): here you could use the `/@/` import scheme

benoitf (approved):
thanks for the refactoring, it'll be easier to introduce new stuff without multiple copy/paste, forgetting one item, etc

packages/renderer/src/lib/kube/KubeActions.svelte
```diff
@@ -0,0 +1,7 @@
...
+</script>
+
+<div class="flex flex-row gap-x-2">
```
deboer-tim: The additional-action section already defines button spacing, we shouldn't create a subcomponent that applies its own.

axel7083: You are right, I was not aware, fixed :+1:

packages/renderer/src/lib/kube/KubeActions.svelte
axel7083: Fixed :+1:

Issue #9406, opened 2024-10-16
got unhandled error traces from chokidar v4


benoitf:
### Bug description

These are the traces I have in Podman Desktop

```
main ↪️ (node:34279) UnhandledPromiseRejectionWarning: Error: UNKNOWN: unknown error, watch '/private/var/run/.sim_diagnosticd_socket' at FSWatcher.<computed> (node:internal/fs/watchers:247:19) at watch (node:fs:2491:36) at createFsWatchInstance (/Applications/Podman Desktop.app/Contents/Resources/app.asar/node_modules/chokidar/handler.js:130:31) at setFsWatchListener (/Applications/Podman Desktop.app/Contents/Resources/app.asar/node_modules/chokidar/handler.js:175:19) at NodeFsHandler._watchWithNodeFs (/Applications/Podman Desktop.app/Contents/Resources/app.asar/node_modules/chokidar/handler.js:329:22) at NodeFsHandler._handleFile (/Applications/Podman Desktop.app/Contents/Resources/app.asar/node_modules/chokidar/handler.js:395:29) at NodeFsHandler._addToNodeFs (/Applications/Podman Desktop.app/Contents/Resources/app.asar/node_modules/chokidar/handler.js:619:31) (Use `Podman Desktop --trace-warnings ...` to show where the warning was created)
main ↪️ (node:34279) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
[...]

PR #10149, opened 2024-11-28
chore: enable unit tests on macOS for PR check job

benoitf:
### What does this PR do?
It looks like there is an issue with chokidar but only on macOS
if we have a unit tests for that, it won't be failing the PR check as we're not running tests on macOS

https://github.com/podman-desktop/podman-desktop/issues/9406#issuecomment-2505369469

### How to test this PR?

unit tests on macOS should be ✅ 


jeffmaury (approved):
LGTM

PR #10165, opened 2024-11-28
fix: cache pod exec websockets

### LINKED ISSUE #9467: Review kubernetesExecIntoContainer telemetry

The kubernetesExecIntoContainer event is triggered a lot, but it tends to be very sporadic and looking at the unique users it's relatively low compared to similar events.

Investigate a little deeper. I noticed that the event is triggered every time you go to the Terminal tab on a pod, I think the volume is primarily caused by users switching in and out of the tab. We could trigger it only the first time you enter (or at least first time for each new pod), only send it once per instance, etc.

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #9467

### What does this PR do?

Caches the websocket connection when a terminal is created for a pod/container.

### Screenshot / video of UI

N/A

#9467 

### How to test this PR?

1. Launch a Kubernetes
2. Switch from Kube to Terminal tabs several times
3. Verify the sh/bash is only launched once

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
@@ -1473,6 +1473,40 @@ test('Test should exec into container ', async () => {
   execResp.onResize(1, 1);
 });
 
+test('Test should exec into container only once', async () => {
+  const client = createTestClient('default');
+  makeApiClientMock.mockReturnValue({
+    getCode: () => Promise.resolve({ body: { gitVersion: 'v1.20.0' } }),
+  });
+
+  const onStdOutFn = vi.fn();
+
+  const onStdErrFn = vi.fn();
+
+  const onCloseFn = vi.fn();
+
+  execMock.mockImplementation(
+    (
+      _namespace: string,
+      _podName: string,
+      _containerName: string,
+      _command: string | string[],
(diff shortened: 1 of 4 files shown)


axel7083 (commented):
I have some concern about keeping the websocket opened forever, the user do not have any control on the lifecycle, if I open the terminal, then go to another page, the connection is maintained until timeout.

I can confirm that this PR works as intended, but I am not sure this is reasonable to keep the connections forever in background without the user knowing

jeffmaury:
> I have some concern about keeping the websocket opened forever, the user do not have any control on the lifecycle, if I open the terminal, then go to another page, the connection is maintained until timeout.
> 
> I can confirm that this PR works as intended, but I am not sure this is reasonable to keep the connections forever in background without the user knowing

Yes but it is required if you want to maintain history and consistency in the terminal. Maybe a follow up PR as this one as about reducing resources usage (telemetry and shells). WDYT ?

axel7083:
> Yes but it is required if you want to maintain history and consistency in the terminal. Maybe a follow up PR as this one as about reducing resources usage (telemetry and shells). WDYT ?

Yes, sure! oOuld you create an issue to keep track of this ?

axel7083 (approved):
LGTM, tested and works fine on fedora 41

PR #10166, opened 2024-11-29
feat: improve libpod API support

mhdawson:
### What does this PR do?

This PR improves use cases where LibPod is used by Podman desktop by:
1. adding support for some additinal configuration options that were not previously translated from the input options
2. forcing the use of LibPod if an nvidia gpu is requested on linux.  Requesting the device "nvidia.com/gpu=all" only works through the LibPod API, not the docker compatibility API. Therefore, force the use of LibPod over the compatiliby API if this device is requested.

These changes were needed for me to get GPU acceleration working on linux for [podman-desktop-extension-ai-lab](https://github.com/containers/podman-desktop-extension-ai-lab), however, I think they are just general improvements that are useful on their own in cases where containers are started and LibPod would have already been used as the API.

### Screenshot / video of UI

N/A

### How to test this PR?

The tests were updated to cover the additions in terms of the addional options translated into the podman options.

For live test, I tested this PR along with this PR to podman-desktop-extension-ai-lab  - https://github.com/containers/podman-desktop-extension-ai-lab/pull/2180, by creating a model service and validating that the service reported GPU acceleration, and used the GPU when submitted requests to the service.  I also tested with the chatbot Node.js recipe I was working on that it got GPU acceleration as well.

### DIFF
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -3947,6 +3947,13 @@ describe('createContainerLibPod', () => {
       pod: 'pod',
       name: 'name',
       HostConfig: {
+        Devices: [
+          {
+            PathOnHost: 'nvidia.com/gpu=all',
+            PathInContainer: '',
+            CgroupPermissions: '',
+          },
+        ],
         Mounts: [
           {
             Target: 'destination',
@@ -3959,7 +3966,7 @@ describe('createContainerLibPod', () => {
           },
         ],
         NetworkMode: 'mode',
-        SecurityOpt: ['default'],
+        SecurityOpt: ['default', 'label=disable'],
         PortBindings: {
           '8080': [
             {
@@ -3994,7 +4001,8 @@ describe('createContainerLibPod', () => {
     const expectedOptions: PodmanContainerCreateOptions = {
       name: options.name,
       command: options.Cmd,
-      entrypoint: options.Entrypoint,
+      devices: [{ path: 'nvidia.com/gpu=all' }],
+      entrypoint: [options.Entrypoint as string],
       env: {
         key: 'value',
       },
@@ -4015,6 +4023,7 @@ describe('createContainerLibPod', () => {
         nsmode: 'mode',
       },
       seccomp_policy: 'default',
+      selinux_opts: ['disable'],
       portmappings: [
         {
           container_port: 8080,
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -70,6 +70,7 @@ import type {
   LibPod,
   PlayKubeInfo,
   PodInfo as LibpodPodInfo,
+  PodmanDevice,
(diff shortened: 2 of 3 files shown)


mhdawson:
I should have also mentioned that the change to the entrypoint was need to match the spec in https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerCreateLibpod and it did not work as a single string. My assumption is that nothing within podman desktop had used the entrypoint passing in a string in the  options (the compat options allow either a string or array of strings)).

benoitf (commented):
hello @mhdawson I think there are too many different issues trying to be solved at once in this PR

Thanks for the contribution but could you please target a specific issue in subsequent PR

like: one for entrypoint, another for selinux options

then, forcing the usage of another API can have side effect 

I'm also wondering why it works only if using the libpod API and not the compat API ? is it an issue from podman side ?

mhdawson:
@benoitf this is the issue I used to ask about "libpod API and not the compat API" in podman - https://github.com/containers/podman/issues/24687.  From that understanding is that it only works through the LibPod API and I don't see an interest on the podman side to change that.

In terms of "then, forcing the usage of another API can have side effect".  I agree but that will only happen in cases where the nvidia device is requested, and if there was any existing user that was falling into that case it would not be working as they would not get the device they requested.

Do you want one PR for each additionally supported option in the podMan options or would 2 PRs, one with the additionally supported options and one for forcing LibPod make sense.

I think it would either be 4 for:

- entrypoint change
- adding devices to podman options
- adding selinux_opts
- forcing LibPod

Or 2:
- one with additionsl to options
- one for forcing LibPod

Let me know what you perfer.

I had run the typecheck locally and I thought it passed, but obviously not from the CI run. Will need to take another look.

benoitf:
I would go with the 4 PRs

we can then discuss very specific points in each one, and some might be merged quickly

mhdawson:
Pushed change to address the typecheck failures.

Will now work on splitting out the different changes into 4 PRs

mhdawson:
@benoitf first PR - https://github.com/podman-desktop/podman-desktop/pull/10178

mhdawson:
@benoitf second PR  - https://github.com/podman-desktop/podman-desktop/pull/10179

mhdawson:
third PR - https://github.com/podman-desktop/podman-desktop/pull/10180

mhdawson:
@benoitf I think that the fourth PR needs the third one which adds the initial device support (https://github.com/podman-desktop/podman-desktop/pull/10180) so I'll wait to create that one until the third PR hopefully lands

mhdawson:
@benoitf here is the last PR of the four - https://github.com/podman-desktop/podman-desktop/pull/10251

mhdawson:
Ok, all four PRs covering what this had have landed. @benoitf many thanks in your helps and suggestions, closing.

benoitf:
@mhdawson thanks for the split across multiple chunks and rebasing each of them 👍

PR #10172, opened 2024-11-29
fix: enable the '@typescript-eslint/no-explicit-any' rule

### LINKED ISSUE #10019: enable the '@typescript-eslint/no-explicit-any' rule in renderer package for the linter

today the linter is being disabling the rule on the renderer side

need to fix the code so we can enable by default the rule

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #10019

### What does this PR do?

Enabling the '@typescript-eslint/no-explicit-any' rule

### Screenshot / video of UI

N/A

#10019 

### How to test this PR?

Check unit tests

### DIFF
--- a/packages/renderer/src/lib/appearance/Appearance.spec.ts
+++ b/packages/renderer/src/lib/appearance/Appearance.spec.ts
@@ -21,7 +21,7 @@ import '@testing-library/jest-dom/vitest';
 
 import { render, type RenderResult } from '@testing-library/svelte';
 import type { Component, ComponentProps } from 'svelte';
-import { beforeEach, expect, test, vi } from 'vitest';
+import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
 import { configurationProperties } from '/@/stores/configurationProperties';
 
@@ -30,16 +30,21 @@ import Appearance from './Appearance.svelte';
 
 const getConfigurationValueMock = vi.fn();
 const addEventListenerMock = vi.fn();
+const matchMediaMock = vi.fn();
+
+beforeAll(() => {
+  Object.defineProperty(window, 'getConfigurationValue', { value: getConfigurationValueMock });
+  Object.defineProperty(window, 'matchMedia', { value: matchMediaMock });
+  Object.defineProperty(window, 'setNativeTheme', { value: vi.fn() });
+});
 
 beforeEach(() => {
   vi.resetAllMocks();
-  (window as any).getConfigurationValue = getConfigurationValueMock;
-  (window as any).matchMedia = vi.fn().mockReturnValue({
+  matchMediaMock.mockReturnValue({
     matches: false,
     addEventListener: addEventListenerMock,
     removeEventListener: vi.fn(),
   });
-  (window as any).setNativeTheme = vi.fn();
 });
 
 function getRootElement(container: HTMLElement): HTMLElement {
@@ -67,7 +72,7 @@ async function awaitRender(): Promise<RenderResult<Component<ComponentProps<Appe
 }
 
 test('Expect light mode using system when OS is set to light', async () => {
-  (window as any).matchMedia = vi.fn().mockReturnValue({
+  matchMediaMock.mockReturnValue({
     matches: false,
     addEventListener: vi.fn(),
     removeEventListener: vi.fn(),
@@ -82,7 +87,7 @@ test('Expect light mode using system when OS is set to light', async () => {
 });
 
 test('Expect dark mode using system when OS is set to dark', async () => {
-  (window as any).matchMedia = vi.fn().mockReturnValue({
+  matchMediaMock.mockReturnValue({
     matches: true,
     addEventListener: vi.fn(),
     removeEventListener: vi.fn(),
--- a/packages/renderer/src/lib/authentication/AuthActions.svelte
+++ b/packages/renderer/src/lib/authentication/AuthActions.svelte
@@ -25,8 +25,8 @@ function handleEscape({ key }: KeyboardEvent): void {
   }
 }
 
-function onWindowClick(e: MouseEvent): void {
-  showMenu = outsideWindow.contains(e.target as Node);
+function onWindowClick(e: Event): void {
+  showMenu = e.target instanceof Node && outsideWindow.contains(e.target);
 }
 
 export function onButtonClick(e: MouseEvent): void {
--- a/packages/renderer/src/lib/authentication/AuthActions.svelte
+++ b/packages/renderer/src/lib/authentication/AuthActions.svelte
@@ -25,8 +25,8 @@ function handleEscape({ key }: KeyboardEvent): void {
   }
 }
 
-function onWindowClick(e: any): void {
-  showMenu = outsideWindow.contains(e.target);
+function onWindowClick(e: MouseEvent): void {
+  showMenu = outsideWindow.contains(e.target as Node);
 }
 
 export function onButtonClick(e: MouseEvent): void {
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
@@ -74,7 +74,7 @@ test('expect being able to reconnect ', async () => {
     ) => {
       onDataCallback = onData;
       // return a callback id
-      return sendCallbackId;
+      return Promise.resolve(sendCallbackId);
     },
   );
 
@@ -148,7 +148,7 @@ test('terminal active/ restarts connection after stopping and starting a contain
         onEnd();
       }, 500);
       // return a callback id
-      return sendCallbackId;
+      return Promise.resolve(sendCallbackId);
     },
   );
 
@@ -216,7 +216,7 @@ test('terminal active/ restarts connection after restarting a container', async 
         onEnd();
       }, 500);
       // return a callback id
-      return sendCallbackId;
+      return Promise.resolve(sendCallbackId);
     },
   );
 
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
@@ -21,7 +21,7 @@ import '@testing-library/jest-dom/vitest';
 import { render, screen, waitFor } from '@testing-library/svelte';
 import { tick } from 'svelte';
 import { get } from 'svelte/store';
-import { beforeEach, expect, test, vi } from 'vitest';
+import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
 import { containerTerminals } from '/@/stores/container-terminal-store';
 
@@ -39,13 +39,19 @@ vi.mock('xterm', () => {
   };
 });
 
+const matchMediaMock = vi.fn();
+
+beforeAll(() => {
+  Object.defineProperty(window, 'getConfigurationValue', { value: getConfigurationValueMock });
+  Object.defineProperty(window, 'shellInContainer', { value: shellInContainerMock });
+  Object.defineProperty(window, 'shellInContainerResize', { value: shellInContainerResizeMock });
+
+  Object.defineProperty(window, 'matchMedia', { value: matchMediaMock });
+});
+
 beforeEach(() => {
   vi.resetAllMocks();
-  (window as any).getConfigurationValue = getConfigurationValueMock;
-  (window as any).shellInContainer = shellInContainerMock;
-  (window as any).shellInContainerResize = shellInContainerResizeMock;
-
-  (window as any).matchMedia = vi.fn().mockReturnValue({
+  matchMediaMock.mockReturnValue({
     addListener: vi.fn(),
     removeListener: vi.fn(),
   });
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
@@ -39,22 +39,20 @@ vi.mock('xterm', () => {
   };
 });
 
-const matchMediaMock = vi.fn();
-
 beforeAll(() => {
   Object.defineProperty(window, 'getConfigurationValue', { value: getConfigurationValueMock });
   Object.defineProperty(window, 'shellInContainer', { value: shellInContainerMock });
   Object.defineProperty(window, 'shellInContainerResize', { value: shellInContainerResizeMock });
 
-  Object.defineProperty(window, 'matchMedia', { value: matchMediaMock });
+  Object.defineProperty(window, 'matchMedia', { value: vi.fn() });
 });
 
 beforeEach(() => {
   vi.resetAllMocks();
-  matchMediaMock.mockReturnValue({
+  vi.mocked(window.matchMedia).mockReturnValue({
     addListener: vi.fn(),
     removeListener: vi.fn(),
-  });
+  } as unknown as MediaQueryList);
 
   // reset terminals
   containerTerminals.set([]);
--- a/packages/renderer/src/lib/image/ImageDetails.spec.ts
+++ b/packages/renderer/src/lib/image/ImageDetails.spec.ts
@@ -21,7 +21,7 @@ import '@testing-library/jest-dom/vitest';
 import { fireEvent, render, screen, waitFor } from '@testing-library/svelte';
 import { get } from 'svelte/store';
 import { router } from 'tinro';
-import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { afterEach, beforeAll, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import { lastPage } from '/@/stores/breadcrumb';
 import { containersInfos } from '/@/stores/containers';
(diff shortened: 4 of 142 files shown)


packages/renderer/src/lib/validation/FieldValidation.ts
```diff
@@ -41,13 +41,13 @@ export function createFieldValidator(
...
 
-  function action(_node: any, binding: any): UpdateAction {
+  function action(_node: unknown, binding: unknown): UpdateAction {
     function validate(value: string, dirty: boolean): void {
       const result = validator(value, dirty);
       writableObject.set(result);
     }
 
-    validate(binding, false);
+    validate(binding as string, false);
```
benoitf: we can't do this kind of casting

String(binding) yes but not `binding as string` (let say you get a number or an object as the unknown object)

jeffmaury: Not sure to get the difference as the old code defined binding as any but did not perform cast but will update code

benoitf: the right approach would be to update validate method to accept an unknown object (and not a string)
then the validator could check the value like: is it a string ? yes, then I do this kind of check on top of it

here by doing String() it'll ensure we give/call the toString() operation on the object while doing `object as string` we might have a runtime issue

packages/renderer/src/lib/authentication/AuthActions.svelte
```diff
@@ -25,8 +25,8 @@ function handleEscape({ key }: KeyboardEvent): void {
   }
 }
 
-function onWindowClick(e: any): void {
-  showMenu = outsideWindow.contains(e.target);
+function onWindowClick(e: MouseEvent): void {
+  showMenu = outsideWindow.contains(e.target as Node);
```
benoitf: question: here I think we have a problem as we should not assume it's a Node

packages/renderer/src/lib/image/ImageDetails.spec.ts
```diff
@@ -68,19 +68,22 @@ delete myNoneNameImage.RepoTags;
...
+  Object.defineProperty(window, 'showMessageBox', { value: showMessageBoxMock });
+  Object.defineProperty(window, 'listImages', { value: listImagesMock });
+  Object.defineProperty(window, 'listContainers', { value: vi.fn() });
+  Object.defineProperty(window, 'deleteImage', { value: deleteImageMock });
+  Object.defineProperty(window, 'hasAuthconfigForImage', { value: hasAuthMock });
+  Object.defineProperty(window, 'getImageCheckerProviders', { value: vi.fn().mockResolvedValue([]) });
+  Object.defineProperty(window, 'listViewsContributions', { value: vi.fn().mockResolvedValue([]) });
+  Object.defineProperty(window, 'getImageFilesProviders', { value: vi.fn().mockResolvedValue([]) });
+  Object.defineProperty(window, 'getConfigurationProperties', { value: vi.fn().mockResolvedValue({}) });
+  Object.defineProperty(window, 'getContributedMenus', { value: getContributedMenusMock });
```
benoitf, line 81: question, why the multiple calls to `Object.defineProperty(window)` vs a single `Object.defineProperty(global, 'window', {`

packages/renderer/src/lib/image/RenameImageModal.svelte
```diff
@@ -27,8 +27,8 @@ function disableSave(name: string, tag: string): boolean {
 }
 
 let imageNameErrorMessage = '';
-function validateImageName(event: any): void {
-  let inputName = event.target.value;
+function validateImageName(event: Event): void {
+  let inputName = (event.target as unknown as Input).value;
```
benoitf: issue: we cannot assume this cast

packages/renderer/src/lib/image/RenameImageModal.svelte
```diff
@@ -37,8 +37,8 @@ function validateImageName(event: any): void {
 }
 
 let imageTagErrorMessage = '';
-function validateImageTag(event: any): void {
-  let inputName = event.target.value;
+function validateImageTag(event: Event): void {
+  let inputName = (event.target as unknown as Input).value;
```
benoitf: issue: we cannot assume this cast

packages/renderer/src/lib/image/RunImage.svelte
```diff
@@ -572,8 +572,8 @@ function deleteDevice(index: number) {
 }
 
 // called when user change the container's name
-function checkContainerName(event: any) {
-  const containerValue = event.target.value;
+function checkContainerName(event: Event) {
+  const containerValue = (event.target as unknown as Input).value;
```
benoitf: issue: we cannot assume this cast

packages/renderer/src/lib/kube/KubePlayYAML.svelte
```diff
@@ -37,6 +37,7 @@ let currentNamespace: string | undefined;
 let allNamespaces: V1NamespaceList;
 
 let playKubeResultRaw: string;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf, line 40: issue: it should not be any but contains the optional field that we may have

packages/renderer/src/lib/kube/KubePlayYAML.svelte
```diff
@@ -62,6 +63,7 @@ const kubeFileDialogOptions: OpenDialogOptions = {
   ],
 };
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf: issue: type should be added (not any)

packages/renderer/src/lib/pod/DeployPodToKube.svelte
```diff
@@ -35,6 +35,7 @@ let deployUsingServices = true;
 let deployUsingRoutes = true;
 let deployUsingRestrictedSecurityContext = false;
 let createdPod: V1Pod | undefined = undefined;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf: issue: the type should not be any

benoitf: unknown, then checking if some fields are here and then the optional Types being in the type
if (field in myObject) {
 ....
}

packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
```diff
@@ -40,6 +40,7 @@ export let providerInfo: ProviderInfo;
 export let propertyScope: string;
 export let callback: (
   param: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf, line 43: could it be unknown type ?

packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
```diff
@@ -347,9 +349,9 @@ function updateStore() {
   });
 }
 
-async function handleOnSubmit(e: any) {
+async function handleOnSubmit(e: SubmitEvent) {
   errorMessage = undefined;
-  const formData = new FormData(e.target);
+  const formData = new FormData(e.target as HTMLFormElement);
```
benoitf: issue: there should not be any direct cast

packages/renderer/src/lib/preferences/PreferencesContainerConnectionEdit.svelte
```diff
@@ -44,6 +44,7 @@ onDestroy(() => {
 
 async function editConnection(
   internalProviderId: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf, line 47: unknown ?

packages/renderer/src/lib/ui/Steps.svelte
```diff
@@ -7,6 +7,7 @@
 <script lang="ts">
 import { Steps } from 'svelte-steps';
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf, line 10: issue: unkown or some real types ?

packages/renderer/src/lib/webview/Webview.svelte
```diff
@@ -37,6 +37,7 @@ let webviewElement: HTMLElement | undefined;
 const postMessageToWebview = (webviewEvent: unknown) => {
   const webviewEventTyped = webviewEvent as { id: string; message: unknown };
   if (id === webviewEventTyped.id) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf: probably use a valid type of webviewElement ? (or check if method send exists so we do not need to use any)

packages/renderer/src/stores/navigation/navigation-registry.ts
```diff
@@ -34,6 +34,7 @@ export interface NavigationRegistryEntry {
   name: string;
   icon: {
     iconImage?: string | { readonly light: string; readonly dark: string };
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     iconComponent?: any;
```
benoitf: IconDefinition or Component or string ?

packages/renderer/src/lib/image/ImageDetails.spec.ts
jeffmaury: Because I translated all (window as any).xxx = to Object.property. We could merge but not sure it is better for readability

packages/renderer/src/lib/image/RenameImageModal.svelte
jeffmaury: Why ? Unless I miss something Input is where the event is coming from

benoitf: it's an assumption, it's not the case from the types

```
if (something instanceOf ...) {
 // I can do the cast
}
```

```
if ('property' in object) {
 // I can use 'property' from the object
}
```

packages/renderer/src/stores/navigation/navigation-registry.ts
axel7083: See example here

https://github.com/podman-desktop/podman-desktop/blob/42f27c4dc761f6bbea645a73b4d4038fc761dc5a/packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte#L15

packages/renderer/src/lib/markdown/Markdown.spec.ts
```diff
@@ -30,7 +30,7 @@ async function waitRender(customProperties: object): Promise<void> {
 }
 
 beforeAll(() => {
-  (window as any).executeCommand = vi.fn();
+  Object.defineProperty(window, 'executeCommand', { value: vi.fn(), writable: true });
```
gastoner: is the writable required here?

jeffmaury: Fixed by 4f443b7

benoitf:
AFAIK you could remove all Object.defineProperty methods for window objects that only set a vi.fn() method as a result of https://github.com/podman-desktop/podman-desktop/pull/10426

packages/renderer/src/stores/navigation/navigation-registry.ts
jeffmaury: Fixed by 43d0444

packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
jeffmaury: Spans IPC methods so delayed to a follow up PR

packages/renderer/src/lib/preferences/PreferencesContainerConnectionEdit.svelte
jeffmaury: Follow up PR as it spans IPC methods

packages/renderer/src/lib/ui/Steps.svelte
jeffmaury: Tried to define types but got strange error in Svelte check, delaying

packages/renderer/src/lib/authentication/AuthActions.svelte
jeffmaury: Added check

packages/renderer/src/lib/image/RenameImageModal.svelte
jeffmaury: Added check

packages/renderer/src/lib/image/RenameImageModal.svelte
jeffmaury: Added check

packages/renderer/src/lib/image/RunImage.svelte
jeffmaury: Added check

packages/renderer/src/lib/webview/Webview.svelte
jeffmaury: Added check

axel7083 (approved):

packages/renderer/src/lib/appearance/Appearance.spec.ts
```diff
@@ -30,16 +30,21 @@ import Appearance from './Appearance.svelte';
 
 const getConfigurationValueMock = vi.fn();
 const addEventListenerMock = vi.fn();
+const matchMediaMock = vi.fn();
+
+beforeAll(() => {
+  Object.defineProperty(window, 'getConfigurationValue', { value: getConfigurationValueMock });
+  Object.defineProperty(window, 'matchMedia', { value: matchMediaMock });
```
benoitf, line 37: non blocking: this method is now globally exposed so I would say it can be removed

packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
```diff
@@ -39,13 +39,19 @@ vi.mock('xterm', () => {
   };
 });
 
+const matchMediaMock = vi.fn();
```
benoitf: why do we need to define matchMediaMock ?

packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
```diff
@@ -39,13 +39,19 @@ vi.mock('xterm', () => {
...
+});
+
 beforeEach(() => {
   vi.resetAllMocks();
-  (window as any).getConfigurationValue = getConfigurationValueMock;
-  (window as any).shellInContainer = shellInContainerMock;
-  (window as any).shellInContainerResize = shellInContainerResizeMock;
-
-  (window as any).matchMedia = vi.fn().mockReturnValue({
+  matchMediaMock.mockReturnValue({
```
benoitf: why not vi.mocked(window.matchMedia).mockReturnValue ?

packages/renderer/src/lib/kube/KubePlayYAML.svelte
benoitf: I guess it'll be a follow-up ? as the issue is still present

packages/renderer/src/lib/kube/KubePlayYAML.svelte
```diff
@@ -91,12 +92,14 @@ async function playKubeFile(): Promise<void> {
         if (playKubeResultJSON.Pods.length > 0) {
           // Filter out the pods that have container errors, but check to see that container errors exists first
           const containerErrors = playKubeResultJSON.Pods.filter(
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf, line 95: will be a follow-up ?

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -26,6 +26,7 @@ export function createListener(
     value?: unknown,
   ) => void,
 ) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf, line 29: follow-up ?

packages/renderer/src/lib/markdown/micromark-warnings-directive.ts
```diff
@@ -67,6 +67,7 @@ export function decode(value: string) {
 /**
  * @type {import('micromark-extension-directive').Handle}
  */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benoitf: d is not Directive type ?

packages/renderer/src/lib/pod/DeployPodToKube.svelte
```diff
@@ -161,16 +164,21 @@ async function deployToKube() {
...
 
   // if we deploy using services, we need to get rid of .hostPort and generate kubernetes services object
   if (deployUsingServices) {
     // collect all ports
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     bodyPod.spec?.containers?.forEach((container: any) => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       container?.ports?.forEach((port: any) => {
-        let portName = `${bodyPod.metadata.name}-${port.hostPort}`;
+        let portName = `${bodyPod.metadata?.name}-${port.hostPort}`;
```
benoitf: issue: should we skip if metadata.name is undefined rather than using ? everywhere below

for example I'm not sure `${bodyPod.metadata?.name}-${port.hostPort}` producing 'undefined-something' is expected ?

benoitf (approved):
all the new excluded any can be follow-up

I'm just worried on the optional metadata?.name . I think the element should be skipped if metadata is not defined rather than having a potential undefined value

packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
jeffmaury: Fixed by e7f3ec1

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
jeffmaury: Fixed by aef1308

packages/renderer/src/lib/pod/DeployPodToKube.svelte
jeffmaury: Fixed by 8901f95

PR #10174, opened 2024-11-29
feat: informers for Kubernetes experimental 

feloy:

### What does this PR do?

Start informers for Kubernetes experimental.

By accessing the underlying cache of the informer, the get/list methods of the cache can be exposed to the world. If the cluster is offline, the cache is still accessible an exposing the latest known resources.

The informer trigger an event every time the cache is updated. For now, only a generic CacheUpdated event, but can be upgraded to [Add/Update/Delete]Resource events if necessary.

The informers trigger an `offline` event when they are disconnected, and expose a `reconnect` method, so they can be restarted (by the user, after some time, etc - not implemented in this PR)

Part of #9938

### How to test this PR?

```
```

### DIFF
--- a/packages/main/src/plugin/kubernetes/resource-factory-handler.ts
+++ b/packages/main/src/plugin/kubernetes/resource-factory-handler.ts
@@ -34,8 +34,8 @@ export class ResourceFactoryHandler {
 
   getPermissionsRequests(namespace: string): ContextPermissionsRequest[] {
     return [
-      ...this.getNamespacedOrNotPermissionsRequests(this.#resourceFactories, true, namespace),
-      ...this.getNamespacedOrNotPermissionsRequests(this.#resourceFactories, false),
+      ...this.getNamespacedOrNotPermissionsRequests(this.#resourceFactories, namespace),
+      ...this.getNamespacedOrNotPermissionsRequests(this.#resourceFactories),
     ];
   }
 
@@ -45,15 +45,9 @@ export class ResourceFactoryHandler {
 
   private getNamespacedOrNotPermissionsRequests(
     factories: ResourceFactory[],
-    isNamespaced: boolean,
     namespace?: string,
   ): ContextPermissionsRequest[] {
-    if (!isNamespaced && !!namespace) {
-      throw new Error('request for non-namespaced must not define a namespace');
-    }
-    if (isNamespaced && !namespace) {
-      throw new Error('request for namespaced must define a namespace');
-    }
+    const isNamespaced = !!namespace;
     const filteredResourceFactories = factories
       .filter(isResourceFactoryWithPermissions)
       .filter(f => f.permissions.isNamespaced === isNamespaced);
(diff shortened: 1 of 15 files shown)


packages/main/src/plugin/kubernetes/resource-factory-handler.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (this.getResourceFactoryByResourceName(factory.resource)) {
+      throw new Error(`a factory for resource ${factory.resource} has already been added`);
+    }
+    this.#resourceFactories.push(factory);
+  }
+
+  getPermissionsRequests(namespace: string): ContextPermissionsRequest[] {
+    return [
+      ...this.getNamespacedOrNotPermissionsRequests(this.#resourceFactories, true, namespace),
+      ...this.getNamespacedOrNotPermissionsRequests(this.#resourceFactories, false),
```
benoitf: question: I'm wondering, why we need to provide the isNamespaced and the namespace

I mean, could we do if namespace is provided, it means, isNamespaced is true, if not provided, isNamespaced is false

packages/main/src/plugin/kubernetes/resource-factory-handler.ts
```diff
@@ -0,0 +1,99 @@
...
+    factories: ResourceFactory[],
+    isNamespaced: boolean,
+    namespace?: string,
+  ): ContextPermissionsRequest[] {
+    if (!isNamespaced && !!namespace) {
+      throw new Error('request for non-namespaced must not define a namespace');
+    }
+    if (isNamespaced && !namespace) {
+      throw new Error('request for namespaced must define a namespace');
+    }
```
benoitf: nitpick: if we merge isNamespaced and namespace we could not need the usage of these tests ?

benoitf (approved):
nitpick: could increase some coverage of the new parts
image

packages/main/src/plugin/kubernetes/resource-informer.ts
```diff
@@ -0,0 +1,127 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface CacheUpdatedEvent extends BaseEvent {}
+
+export interface OfflineEvent extends BaseEvent {
+  offline: boolean;
+  reason?: string;
+}
+
+export class ResourceInformer<T extends KubernetesObject> implements Disposable {
```
feloy, line 40: add a namespace as parameter?

axel7083: yes!

feloy: In fact, the namespace is already part of the parameters, as part of the `path`.

For the reason we have discussed (being able to list a resource for several namespaces of a same cluster), I think that it would be better to inject these namespaces as part of the kubeconfig we pass to the `ContextManagerExperimental.update`, by duplicating the context and changing its namespace (either by modifying the user's kubeconfig file, or creating one in-memory for this), so we can do the entire process, including checking the permissions.

axel7083 (approved):

packages/main/src/plugin/kubernetes/resource-informer.ts
benoitf: maybe there was a previous discussion not shared before the 'add a namespace as parameter' but yes, I agree with the last comment, I think the scope is to have the namespace part of the configuration object that we're receiving, as it's for one context and a given namespace.
New namespace to monitor = new sub config to watch but using the same pattern/ logic.

PR #10179, opened 2024-11-29
fix: improve libpod API translation - extension

mhdawson:
### What does this PR do?

fixes translation of entrypoint in options passed when the libPod API is used. From the spec - https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerCreateLibpod is must be an array and it did not work as a single string in my testing of a use case that needed to use it.

My assumption is that nothing within podman desktop had used the entrypoint passing in a string in the options whe the libPod API is used (the docker compat API options allow either a string or array of strings)).

Fixed the translation to ensure it is always an array when passed to the libPod API.

My assumption is that nothing within podman desktop had used the entrypoint passing in a string in the options whe the libPod API is used (the docker compat API options allow either a string or array of strings)).

See, https://github.com/podman-desktop/podman-desktop/pull/10166 for more context as to one of the use cases where this is needed. 

### Screenshot / video of UI

N/A

### How to test this PR?

Also tested as part of running [podman-desktop-extension-ai-lab](https://github.com/containers/podman-desktop-extension-ai-lab) with linux GPU acceleration as explained in https://github.com/podman-desktop/podman-desktop/pull/10166

### DIFF
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -4062,7 +4062,7 @@ describe('createContainerLibPod', () => {
       name: options.name,
       command: options.Cmd,
       devices: [{ path: 'nvidia.com/gpu=all' }],
-      entrypoint: options.Entrypoint,
+      entrypoint: [options.Entrypoint as string],
       env: {
         key: 'value',
       },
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/container-registry.spec.ts
```diff
@@ -3994,7 +3994,7 @@ describe('createContainerLibPod', () => {
     const expectedOptions: PodmanContainerCreateOptions = {
       name: options.name,
       command: options.Cmd,
-      entrypoint: options.Entrypoint,
+      entrypoint: [options.Entrypoint as string],
```
benoitf, line 4065: issue: we're missing some unit tests to check the multiple cases (single entry, no entry or array)

packages/main/src/plugin/container-registry.ts
```diff
@@ -2072,7 +2072,7 @@ export class ContainerProviderRegistry {
     const podmanOptions: PodmanContainerCreateOptions = {
       name: options.name,
       command: options.Cmd,
-      entrypoint: options.Entrypoint,
+      entrypoint: !options.Entrypoint || Array.isArray(options.Entrypoint) ? options.Entrypoint : [options.Entrypoint],
```
benoitf, line 2089: issue: all the new paths should be covered by unit tests

packages/main/src/plugin/container-registry.spec.ts
mhdawson: Thanks for taking a look, I can add some additonal tests. Will try to get to that tomorrow.

mhdawson: @benoitf added additional tests and rebased.

benoitf (approved):
nit: was just wondering why we're applying mockClear inside the test multiple times rather than having 3 different unit tests (where the mocks are automatically reset)

like:
```
 describe('entrypoints) {
  ...mocking...
  beforeEach
   test(...test1...)
   test(...test1...)
   test(...test1...)
}
```

mhdawson:
@benoitf on the question of using multiple tests. I looked at duplicating the tests but that would have added a lot more lines of code which were duplicates.  There might be a another way to structure the tests that would move some of that into before/after setup but that would need more time for me to understand how the test frame works, and I think would still required a good amount of duplicate code.

benoitf:
@mhdawson if you can fix the conflict thanks

mhdawson:
@benoitf rebased

benoitf (approved):

Issue #9924, opened 2024-11-15
Implementation of tasks panel phase 1


slemeur:
### Is your feature request related to a problem? Please describe

Implementation task for https://github.com/podman-desktop/podman-desktop/issues/8628#issuecomment-2470938372 

### Describe the solution you'd like

Implementation task for https://github.com/podman-desktop/podman-desktop/issues/8628#issuecomment-2470938372

PR #10194, opened 2024-12-02
chore: refactor type so we can have an iterator

benoitf:
### What does this PR do?
allow to be able to iterate on values used by a type

#9924 

### How to test this PR?

### DIFF
--- a/packages/api/src/taskInfo.ts
+++ b/packages/api/src/taskInfo.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2024 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


packages/api/src/taskInfo.ts
```diff
@@ -17,7 +17,8 @@
  ***********************************************************************/
 
 export type TaskState = 'running' | 'completed';
-export type TaskStatus = 'in-progress' | 'success' | 'failure' | 'canceled';
+export const TASK_STATUSES = ['in-progress', 'success', 'failure', 'canceled'] as const;
```
axel7083, line 20: Why not sing an enum ? This synthax seems a bit complicated, why do we need to iterate on the statuses anyway?

benoitf: enum = it'll break everything (all the code using TaskStatus)

> why do we need to iterate on the statuses anyway?

to display tabs to filter status. There is a need of one tab per status (so iterating)

benoitf: we also need the iterator to be able to remove/check for filters

https://github.com/podman-desktop/podman-desktop/pull/10195

axel7083 (approved):

Issue #10154, opened 2024-11-28
Help menu doesn't close when clicking icon again


Firewall:
### Bug description

Opening the menu and then clicking the help icon again doesn't close the menu. 

### Operating system

MacOS

### Installation Method

Installer from website/GitHub releases

### Version

next (development version)

### Steps to reproduce

1. Click icon to open help menu
2. Expected: Click icon again to close menu
3. Actual: Menu doesn't close

PR #10206, opened 2024-12-03
feat: add new task manager

### LINKED ISSUE #9924: Implementation of tasks panel phase 1

benoitf:
### What does this PR do?

Implements new task manager using more space, table design and filtering options

### Screenshot / video of UI

https://github.com/user-attachments/assets/4afd70af-5c8f-4c42-a060-da36b436d772

fixes https://github.com/podman-desktop/podman-desktop/issues/9924

### How to test this PR?

unit tests provided.

You can install the extension `quay.io/fbenoit/multiple-tasks-example:v1.0` and then bring the command palette (F1 key) and run `Multiple Tasks` command


packages/renderer/src/lib/task-manager/TaskManager.svelte
```diff
@@ -0,0 +1,72 @@
...
+  showTaskManager = false;
+}
+
+// update the search pattern store when the variable is updated
+$effect(() => {
+  searchPattern.set(searchTerm);
+});
+
+// task or tasks depending on the number of selected items
+const taskWordPlural = $derived(selectedItemsNumber > 1 ? 'tasks' : 'task');
```
axel7083, line 39: Nice touch 👍

packages/renderer/src/lib/task-manager/TaskManagerActions.spec.ts
```diff
@@ -0,0 +1,60 @@
+/**********************************************************************
```
axel7083: Could we add some depth to the `lib/task-manager` by adding some subfolders ?

- `lib/task-manager/actions`
- `lib/task-manager/table`
etc.

That's a lot of files, and we can probably ease the navigation by creating some subfolders

packages/renderer/src/lib/task-manager/TaskManagerEmptyFilteredScreen.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+import { FilteredEmptyScreen } from '@podman-desktop/ui-svelte';
+
+import TaskIcon from '/@/lib/images/TaskIcon.svelte';
+
+let { searchTerm = $bindable(), onReset = () => {} }: { searchTerm: string; onReset?: (term: string) => void } =
```
axel7083: suggestion (non-blocking): might create a Props above when we have more that one?

axel7083: nice usage of svelte5 btw

packages/renderer/src/lib/task-manager/TaskManagerTabs.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+import { Button } from '@podman-desktop/ui-svelte';
+
+import { IS_TASK_STATUSES } from '/@/stores/tasks';
+import { TASK_STATUSES } from '/@api/taskInfo';
+
+let {
+  searchTerm = $bindable(),
+  onUpdate = (): void => {},
+}: { searchTerm: string; onUpdate?: (searchTerm: string) => void } = $props();
```
axel7083: suggestion (not-blocking): props interface above ?

packages/renderer/src/lib/task-manager/TaskManagerWindowEvents.svelte
```diff
@@ -0,0 +1,54 @@
+<script lang="ts">
+let {
+  showTaskManager = $bindable(false),
+  outsideWindow = $bindable(),
+  onUpdate = (): void => {},
+}: {
+  showTaskManager: boolean;
+  outsideWindow: HTMLDivElement | undefined;
+  onUpdate?: (val: boolean) => void;
+} = $props();
```
axel7083: suggestion (not-blocking): props interface ?

axel7083 (commented):
I don't have much to say about the code expect that it is excellent: all components are tested properly, each one has its own behaviour defined and covered by the tests.

The tests are very clear to understand, simple while still covering properly the different cases.

After testing it, by installing the extensions `quay.io/fbenoit/multiple-tasks-example:v1.0` you provided, (kudo for doing this 👍 , it really ease the testing)

I just notice one problem when I open/close the TaskManager, I'm restarted podman-desktop twice, cleaned the node_modules... 

https://github.com/user-attachments/assets/3d35e709-29f4-45a0-a61a-f6fc989a3c08

I tried to checkout main, and I am not getting the issue :/

⚠️ I also noticed that clicking on the bell open the task manager, but clicking again do not close it. (not blocking IMO, because the help button has the same problem)

ℹ️ I really like that clicking outside the task manager close it 👍 !

benoitf:
looks like if I remove the usage of the Table component, it's not reproducible

so looks like an interference between two instance of the Table on the same "page"

deboer-tim (commented):
Took a quick look, unfortunately (unrelated) the first task I tried it on was build, which has no progress, no cancellation, and no output if you reopen. :(

As Alex said everything looks in order here, just a few comments:
- In every other table the name column is clickable so I expected that to work here too, I think we should add it for consistency.
- Clicking on the task manager icon a second time doesn't do anything but often makes the window flash. Maybe it should close? (if not, shouldn't flash)
- Design: currently this page takes 80-100% of the screen, which is very awkward to me: you can see lots of tasks but nothing underneath, so why is it a popup? IMHO it could be half the height, or should just be a full screen/normal page.
- Design: you can already filter and sort by progress (status), and users typically only need to see the last few tasks anyway. Personally I'd rather see an extra task and remove the tabs...
- Nit: 'Search Task manager' should be 'Search tasks'.

benoitf:
> Took a quick look, unfortunately (unrelated) the first task I tried it on was build, which has no progress, no cancellation, and no output if you reopen. :(

This is why I'm providing a test extension that is creating multiple tasks (see How to test this PR? section)


> As Alex said everything looks in order here, just a few comments:
>In every other table the name column is clickable so I expected that to work here too, I think we should add it for consistency.

will look at it

> Clicking on the task manager icon a second time doesn't do anything but often makes the window flash. Maybe it should close? (if not, shouldn't flash)

will look at it

>Design: currently this page takes 80-100% of the screen, which is very awkward to me: you can see lots of tasks but nothing underneath, so why is it a popup? IMHO it could be half the height, or should just be a full screen/normal page.
Design: you can already filter and sort by progress (status), and users typically only need to see the last few tasks anyway. Personally I'd rather see an extra task and remove the tabs...

it's implementing the design that it is like that.

> Nit: 'Search Task manager' should be 'Search tasks'.

the component says that if title is "Something" it'll then display a "Search Something..."
here the design is using the name "Task Manager" so the component is displaying "Search Task Manager..."

slemeur:
@deboer-tim :
- Design: currently this page takes 80-100% of the screen, which is very awkward to me: you can see lots of tasks but nothing underneath, so why is it a popup? IMHO it could be half the height, or should just be a full screen/normal page.

Until we have a tabUI mechanism, this is the solution we decided to move forward with. As the the adjustments of the size, I'd rather try to collect user feedbacks. 
Eventually, in a future iteration, we could make the height of the popover adjusted based on the number of items the table will be displaying. IE: if there is only one task in the table, the popover could be reduced in height, but increased if more items.

- Design: you can already filter and sort by progress (status), and users typically only need to see the last few tasks anyway. Personally I'd rather see an extra task and remove the tabs...

The filters are providing a consistent pattern with the other tables we have in general in the UI of the tool. Also, think about this panel as a panel in an IDE displaying building errors and warnings. 
There are cases where an error will arise and clicking on it will open the task.
Again, that is something we can revisit if we are getting a lot of feedback on, otherwise, will follow the UX recommendation and approved design.

slemeur:
Tested and really like it! this is neat work, especially with the testing extension that allows to check all the cases! 

One feedback: when I click on the "bell" icon, I would like the second click to close the window. This way it will be consistent with the current behavior and https://github.com/podman-desktop/podman-desktop/issues/10154

deboer-tim:
> the component says that if title is "Something" it'll then display a "Search Something..." here the design is using the name "Task Manager" so the component is displaying "Search Task Manager..."

Yeah, so maybe just `<NavPage title="task manager"` to fix the capitalization issue?

@slemeur Yes, understood, just giving my feedback that I don't like these choices and find them inconsistent.

benoitf:
> In every other table the name column is clickable so I expected that to work here too, I think we should add it for consistency.

@deboer-tim are you speaking about the column to be able to sort it, which is already the case on my end?

benoitf:
@slemeur @ekidneyrh is it ok to rename the title to be `Tasks` instead of `Task Manager` ?

to know if I need to enhance the component to be able to override the search bar title

ekidneyrh:
> @slemeur @ekidneyrh is it ok to rename the title to be `Tasks` instead of `Task Manager` ?
> 
> to know if I need to enhance the component to be able to override the search bar title

ok by me

slemeur:
> is it ok to rename the title to be Tasks instead of Task Manager ?

It's fine for me as well

benoitf:
PR rebased to include the fix of the Table component

also added support to be able to click on the task icon to close if open
rename task manager to tasks

benoitf:
(I'll introduce Props interface and split to subfolders)

benoitf:
updated

packages/renderer/src/lib/task-manager/TaskManagerTabs.svelte
benoitf: fixed

packages/renderer/src/lib/task-manager/TaskManagerEmptyFilteredScreen.svelte
benoitf: fixed

packages/renderer/src/lib/task-manager/TaskManagerWindowEvents.svelte
benoitf: fixed in the latest commit

packages/renderer/src/lib/task-manager/TaskManagerActions.spec.ts
benoitf: fixed in the latest commit

slemeur:
LGTM

deboer-tim:
> > In every other table the name column is clickable so I expected that to work here too, I think we should add it for consistency.
> 
> @deboer-tim are you speaking about the column to be able to sort it, which is already the case on my end?

Sorry, I meant the names themselves. i.e. in the Pods column I can click on the name pod-a or pod-b to go to each pod's respective details. I think it would be handy and consistent to do the same thing here, i.e. clicking on 'Build x' or 'Pull y' would go to the task (trigger the Go To action).

axel7083 (approved):
LGTM, tested again, I can confirm the issue with the multiple tables on the same page has been fixed.

Great job :+1:

benoitf:
> Sorry, I meant the names themselves. i.e. in the Pods column I can click on the name pod-a or pod-b to go to each pod's respective details. I think it would be handy and consistent to do the same thing here, i.e. clicking on 'Build x' or 'Pull y' would go to the task (trigger the Go To action).

the action is optional so you wouldn't be able to click on each name as well.

feloy (approved):
Tested on Mac, works great

deboer-tim:
> > Sorry, I meant the names themselves. i.e. in the Pods column I can click on the name pod-a or pod-b to go to each pod's respective details. I think it would be handy and consistent to do the same thing here, i.e. clicking on 'Build x' or 'Pull y' would go to the task (trigger the Go To action).
> 
> the action is optional so you wouldn't be able to click on each name as well.

Yes, I know. Personally I still think it's worth it for the ones that can (because that's what I instinctively tried) and obvious enough when you can't.

benoitf:
> Yes, I know. Personally I still think it's worth it for the ones that can (because that's what I instinctively tried) and obvious enough when you can't.

It would not be consistent neither. It's like if you would like to have the click on a container's name in container's list opening the terminal tab.

also this is under experimental to collect more feedback

PR #10211, opened 2024-12-03
fix: table component should be scoped when searching items

### LINKED ISSUE #10210: table component is not working with multiple instances
### Bug description

implementing the task manager, Axel found while testing a layout issue

investigating, I think the flaw is related to how the afterUpdate method is implemented

https://github.com/podman-desktop/podman-desktop/blob/main/packages/ui/src/lib/table/Table.svelte#L148-151

it will update all the matching css parts of the document

### Operating system

macOS

### Installation Method

None

### Version

1.14.1

### Steps to reproduce

you might look at https://github.com/podman-desktop/podman-desktop/pull/10206

### Relevant log output

```shell

```

### Additional context

_No response_

benoitf:
### What does this PR do?
Table componet was searching and updating items on the whole document/page
It should be scoped and only search/update in its children

fixes https://github.com/podman-desktop/podman-desktop/issues/10210

### How to test this PR?

you might try this PR on top of the task manager PR but I also added a unit test

### DIFF
--- a/packages/ui/src/lib/table/Table.spec.ts
+++ b/packages/ui/src/lib/table/Table.spec.ts
@@ -380,3 +380,20 @@ test('Expect duration cell to be empty when undefined', async () => {
     expect(cells[6].textContent?.trim()).toBe(expected[i - 1]);
   }
 });
+
+test('Expect table is scoped for css manipulation', async () => {
+  render(TestTable, {});
+
+  // Wait for the table to update
+  await tick();
+
+  const table = await screen.findByRole('table');
+  expect(table).toBeDefined();
+
+  // ok so now look at our dummy component
+  const dummyComponent = await screen.findByRole('group', { name: 'dummy component' });
+  expect(dummyComponent).toBeDefined();
+
+  // and there should be no style applied to this group as it's not part of the table
+  expect(dummyComponent.style.gridTemplateColumns).toBe('');
+});
--- a/packages/ui/src/lib/table/Table.spec.ts
+++ b/packages/ui/src/lib/table/Table.spec.ts
@@ -389,6 +389,12 @@ test('Expect table is scoped for css manipulation', async () => {
 
   const table = await screen.findByRole('table');
   expect(table).toBeDefined();
+  // get the elements having the role "row" inside the table html element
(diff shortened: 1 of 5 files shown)


packages/ui/src/lib/table/Table.spec.ts
```diff
@@ -380,3 +380,20 @@ test('Expect duration cell to be empty when undefined', async () => {
     expect(cells[6].textContent?.trim()).toBe(expected[i - 1]);
   }
 });
+
+test('Expect table is scoped for css manipulation', async () => {
```
dgolovin, line 384: nit: the reason for this change is to keep fix table rendering in the case when multiple tables are visible. The test proves table rendering does not conflict with other UI elements. There should be tests for two tables rendered in the same page.

packages/ui/src/lib/table/Table.spec.ts
```diff
@@ -380,3 +380,20 @@ test('Expect duration cell to be empty when undefined', async () => {
...
+  render(TestTable, {});
+
+  // Wait for the table to update
+  await tick();
+
+  const table = await screen.findByRole('table');
+  expect(table).toBeDefined();
```
dgolovin, line 391: nit: to be sure that changes does not affect table rendering there should be check for `table.style.gridTemplateColumns` not an empty string.

dgolovin (approved):
LGTM with nits.

packages/ui/src/lib/table/Table.spec.ts
axel7083: I share the same feedback, having a MultipleTableTest.svelte component or something similar

packages/ui/src/lib/table/Table.spec.ts
benoitf: test updated

packages/ui/src/lib/table/Table.spec.ts
benoitf: test updated

benoitf:
pr updated

axel7083 (approved):
Thanks for the change and the fix

dgolovin (approved):
LGTM!
Thank you for update.

Issue #9536, opened 2024-10-22
Improve titles of some tasks


Firewall:
### Is your enhancement related to a problem? Please describe

With the tasks now showing on the statusbar we are exposing the titles of tasks more often. A few of these can be improved.

A couple of examples are:
1. `Create Container provider` when you are creating a new Podman Machine.  Should `Creating Podman Machine` 
3. `Stop podman-machine-default` -> `Stopping podman-machine-default`
4. `Start podman-libkrun` -> `Starting podman-machine-default`
5. `Create Kubernetes provider` (no difference between creating `kind` cluster or `minikube` cluster). Ideally this would be `Creating Kind provider` `Creating MiniKube provider`
6. When building an image the name is just `Build`. Can we make it `Building image <image-name>`?
We are also missing some:
1. When you pull an image, there is no message

### Describe the solution you'd like

Update the texts

slemeur:
On this issue, we would like to see if we can extract all the different messages we are sending as notification. So we can then review them.

SoniaSandler:
Those are the ones that I found:

- `Build ${imageName}`
- `Export container`
- `Import container`
- `Save images`
- `Load images`
- `Update ${tool.name} to v${version}`
- `Install ${tool.name} to v${versionToInstall}`
- `Uninstall ${tool.name} v${tool.version}`
- `Start ${providerConnectionInfo.name}`
- `Stop ${providerConnectionInfo.name}`
- `Create Container provider`
- `Delete ${providerConnectionInfo.name}`
- `Create Kubernetes provider`
- `Downloading ${updateVersion} update`

Firewall:
Thank you @SoniaSandler that helps a lot!

Then I think this is what we should change them too. 
- `Build ${imageName}` -> `Building ${imageName}`
- `Export container` -> `Exporting ${containerName}`
- `Import container` -> `Importing ${containerName}`
- `Save images` -> `Saving image` <- Why is the `images` in plural? Are we saving more than 1 image?
- `Load images` -> `Loading image` <- Why is the `images` in plural? Are we saving more than 1 image?
- `Update ${tool.name} to v${version}` -> `Updating ${tool.name} to v${version}`
- `Install ${tool.name} to v${versionToInstall}` -> `Installing ${tool.name} to v${versionToInstall}`
- `Uninstall ${tool.name} v${tool.version}` -> `Uninstalling ${tool.name} v${tool.version}`
[...]

PR #10212, opened 2024-12-03
chore: add a way to get provider info from internal provider id

SoniaSandler:

### What does this PR do?
This PR add a function to get a provider's name using its internal provider id. This will be used in the PR of https://github.com/podman-desktop/podman-desktop/issues/9536 to improve some of the task titles.

Related to https://github.com/podman-desktop/podman-desktop/issues/9536

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/provider-registry.ts
+++ b/packages/main/src/plugin/provider-registry.ts
@@ -1325,4 +1325,8 @@ export class ProviderRegistry {
       throw error;
     }
   }
+
+  getProviderName(internalProviderId: string): string | undefined {
+    return this.providers.get(internalProviderId)?.extensionDisplayName;
+  }
 }
--- a/packages/main/src/plugin/provider-registry.ts
+++ b/packages/main/src/plugin/provider-registry.ts
@@ -169,7 +169,7 @@ export class ProviderRegistry {
           // If the status does not match the current one, we will send a listener event and update the status
           if (status !== this.providerStatuses.get(providerKey)) {
             provider.updateStatus(status);
-            this.listeners.forEach(listener => listener('provider:update-status', this.getProviderInfo(provider)));
+            this.listeners.forEach(listener => listener('provider:update-status', this.toProviderInfo(provider)));
             this.providerStatuses.set(providerKey, status);
           }
         }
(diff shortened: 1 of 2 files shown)


benoitf (commented):
hello, I'm not sure to follow the need of that ?

We have providerInfos containing the values of all providers including the name ?

SoniaSandler:
@benoitf for tasks like `Create Container provider` and `Create Kubernetes provider` only the `internalProviderId` is provided, but the improved task titles should include the Provider's name. 
In the `provider-registry` there is `getMatchingProvider` that takes internal id and returns the provider, but it is a protected method. There is also `getProviderInfos`, but it return all the providers, and I think that iterating through them all just to get one name every time a task is created is not the best solution.
The method that is added here just returns the name of the provider based on its internal id without too much work.
If there is an already existing method that could be useful for this purpose, lmk (or if `getMatchingProvider` could become public instead of protected).

benoitf:
I would go for getProviderInfo then ( keep the protected method that is returning the implementation object which should not leak outside of the class)

benoitf:
introducing getProviderInfo method returning all the fields / info object, not having one method per field

benoitf (commented):
The current method getProviderInfo should be renamed toProviderInfo() ( and keep being protected)

packages/main/src/plugin/provider-registry.ts
```diff
@@ -1325,4 +1325,13 @@ export class ProviderRegistry {
...
   }
+
+  getProviderInfo(internalProviderId: string): ProviderInfo | undefined {
+    try {
+      const provider = this.getMatchingProvider(internalProviderId);
+      return this.toProviderInfo(provider);
+    } catch (e) {
+      return undefined;
+    }
+  }
```
benoitf: issue: I'm not sure we should use getMatchingProvider because it's throwing an error that we need to catch here
`this.providers.get(internalId)` is returning directly the provider implementation


```suggestion
    const provider = this.providers.get(id);
    if (provider) {
      return this.toProviderInfo(provider);
    }
    return undefined;
```

benoitf (approved):

PR #10235, opened 2024-12-04
chore: remove ignore warning no longer raised

benoitf:

### What does this PR do?
Fix being applied to the dependabot branch

Cannot apply to main as with the existing version of svelte, if we remove the comment there is an error and with the new version of svelte, if we keep the comment, there is an error

related to https://github.com/podman-desktop/podman-desktop/pull/10234

### How to test this PR?

PR check should be ✅ 


benoitf:
oops I click on the auto-merge but it seems there is no check on custom branches...

PR #10239, opened 2024-12-05
chore: show only one error enoent message untill podman machine reconnects

### LINKED ISSUE #8681: Podman machine connection is unstable
### Bug description

This sometimes happens with development and production builds (1.13.0-next)

We see a flickering "No Podman machine is running" warning in AI Lab
Console is flooded with event errors, ENOENT \\.\pipe\podman-machine-default

### Operating system

Windows 10 22h2 Enterprise, Windows 11 23h2 Enterprise

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

* podman machine init --rootful
* start Podman Desktop
* install AI Lab extension
* Try to deploy an app
* Try to pull an image

### Relevant log output

_No response_

### Additional context

Screenshot_20240830_161932
Screenshot_20240830_161708
Screenshot_20240830_161915
[podman-machine-init.log](https://github.com/user-attachments/files/16818482/podman-machine-init.log)
[output.log](https://github.com/user-attachments/files/16818484/output.log)
[ai-lab-e2e.webm](https://github.com/user-attachments/assets/1a89a0b1-3f11-4880-af8f-bbc64e2fb055)

gastoner:
### What does this PR do?
Reduces number of ENOENT error messages in console

Closes #8681 

### How to test this PR?
On Windows, try to delete win_something_proxy process
OR
1. Delete all machines
2. Create machine1 using PD (default options)
3. Create machine2 using PD (default options)
4. Wait until both machines are started
5. Machine1 should be the default
6. Stop machine2 and immediately stop machine1

In both cases, you should see only limited number of "Assert you see the error is Error: connect ENOENT \\.\pipe\machine_name at PipeConnectWrap.afterConnect [as oncomplete] (node:net:1611:16)" messages in console.

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -373,6 +373,12 @@ export class ContainerProviderRegistry {
     });
   }
 
+  notifyConsole(message: string): void {
+    if (this.notify) {
+      console.log(message);
+    }
+  }
+
   // do not use inspect information
   async listSimpleContainers(abortController?: AbortController): Promise<SimpleContainerInfo[]> {
     let telemetryOptions = {};
@@ -397,9 +403,7 @@ export class ContainerProviderRegistry {
             }),
           );
         } catch (error) {
-          if (this.notify) {
-            console.log('error in engine', provider.name, error);
-          }
+          this.notifyConsole(`error in engine ${provider.name} ${error}`);
           telemetryOptions = { error: error };
           return [];
         }
@@ -558,9 +562,7 @@ export class ContainerProviderRegistry {
             }),
           );
         } catch (error) {
-          if (this.notify) {
-            console.log('error in engine', provider.name, error);
-          }
+          this.notifyConsole(`error in engine ${provider.name} ${error}`);
           telemetryOptions = { error: error };
           return [];
         }
@@ -599,9 +601,7 @@ export class ContainerProviderRegistry {
             return imageInfo;
           });
         } catch (error) {
-          if (this.notify) {
-            console.log('error in engine', provider.name, error);
-          }
+          this.notifyConsole(`error in engine ${provider.name} ${error}`);
           telemetryOptions = { error: error };
           return [];
         }
@@ -724,9 +724,7 @@ export class ContainerProviderRegistry {
             return podInfo;
           });
         } catch (error) {
-          if (this.notify) {
-            console.log('error in engine', provider.name, error);
-          }
+          this.notifyConsole(`error in engine ${provider.name} ${error}`);
           telemetryOptions = { error: error };
           return [];
         }
@@ -849,9 +847,7 @@ export class ContainerProviderRegistry {
           });
           return { Volumes: volumeInfos, Warnings: volumeListInfo.Warnings, engineName, engineId };
         } catch (error) {
-          if (this.notify) {
-            console.log('error in engine', provider.name, error);
-          }
+          this.notifyConsole(`error in engine ${provider.name} ${error}`);
           telemetryOptions = { error: error };
           return [];
         }
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -129,7 +129,7 @@ export class ContainerProviderRegistry {
 
   protected containerProviders: Map<string, containerDesktopAPI.ContainerProviderConnection> = new Map();
   protected internalProviders: Map<string, InternalContainerProvider> = new Map();
-  protected activeTimeouts: Map<string, NodeJS.Timeout | undefined> = new Map();
+  protected notify: boolean = true;
 
   // map of streams per container id
   protected streamsPerContainerId: Map<string, NodeJS.ReadWriteStream> = new Map();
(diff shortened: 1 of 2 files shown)


jeffmaury (commented):
Not sure it fixes the issue. When connection to machine is lost, it will prevent the lot of messages being printed but the real issue is that we should not try to reconnect like crazy

gastoner:
I think that the reconnect is happening every 5sec, but the callback `errorHandler` https://github.com/podman-desktop/podman-desktop/blob/main/packages/main/src/plugin/container-registry.ts#L288 is called every time some error event is called. So I think there are accumulating error messages. 

I've tried to add to the timeout some variable and each time some error is detected, I'll check if the timeout is running, if so I'll just skip the error. Otherwise start timeout to reconnect in 5s

jeffmaury (commented):
Seems strange to me as if you have 2 podman machine that fails it seems to me that as the timer is global only 1 of the 2 will reconnect

gastoner:
Yes I have update for this PR => each provider have his own timer, I would update this PR if the concept makes sense. Should I do it?

gastoner:
@jeffmaury updated!

jeffmaury:
Code looks better but when I run a test I see the error message at a higher rate than every 5s

gastoner:
Yes because the error is emitted by the `dockerode.getEvents` function which emits this error like every second.

packages/main/src/plugin/container-registry.ts
```diff
@@ -290,12 +291,22 @@ export class ContainerProviderRegistry {
       internalProvider.api = undefined;
       internalProvider.libpodApi = undefined;
 
+      // if is the timeout running dont start other for events
+      // connect ENOENT \.\pipe\machine_name at PipeConnectWrap.afterConnect [as oncomplete]
+      if (this.activeTimeouts.get(containerProviderConnection.endpoint.socketPath)) {
```
axel7083:
```suggestion
      if (this.activeTimeouts.has(containerProviderConnection.endpoint.socketPath)) {
```

would there be any difference of checking `get` vs `has` here ? The NodeJS.Timeout value will not change depending if the timeout has been executed or not.

image

packages/main/src/plugin/container-registry.ts
```diff
@@ -129,6 +129,7 @@ export class ContainerProviderRegistry {
 
   protected containerProviders: Map<string, containerDesktopAPI.ContainerProviderConnection> = new Map();
   protected internalProviders: Map<string, InternalContainerProvider> = new Map();
+  protected activeTimeouts: Map<string, NodeJS.Timeout | undefined> = new Map();
```
axel7083: what is the point of saving the NodeJS.Timeout ? We could just have a boolean Map since we only check for existence. See my comment here https://github.com/podman-desktop/podman-desktop/pull/10239/files#r1895614426

Would need to rename to `initializedConnectionAPI: Map<string, boolean>`

packages/main/src/plugin/container-registry.ts
```diff
@@ -290,12 +291,22 @@ export class ContainerProviderRegistry {
...
+      this.activeTimeouts.set(
+        containerProviderConnection.endpoint.socketPath,
+        setTimeout(() => {
+          this.activeTimeouts.set(containerProviderConnection.endpoint.socketPath, undefined);
+          this.setupConnectionAPI(internalProvider, containerProviderConnection);
+        }, this.retryDelayEvents),
+      );
     };
 
     this.handleEvents(internalProvider.api, errorHandler);
```
axel7083: We duplicate the events everytime we call this function

https://github.com/podman-desktop/podman-desktop/blob/333f183edbebe12b50232db8dda6768d3b20eed8/packages/main/src/plugin/container-registry.ts#L148

Here `handleEvents` should return a Disposable that we should dispose before recalling it ?

packages/main/src/plugin/container-registry.ts
```diff
@@ -290,12 +291,24 @@ export class ContainerProviderRegistry {
...
-        this.setupConnectionAPI(internalProvider, containerProviderConnection);
-      }, this.retryDelayEvents);
+      this.initializedConnectionAPI.set(
+        containerProviderConnection.endpoint.socketPath,
+        setTimeout(() => {
+          this.initializedConnectionAPI.set(containerProviderConnection.endpoint.socketPath, false);
+          this.setupConnectionAPI(internalProvider, containerProviderConnection);
+        }, this.retryDelayEvents)
+          ? true
+          : false,
```
axel7083: This is non-sense ? `!!setTimeout(() => {}, 2000)` is always true, why checking the result of setTimeout ?

gastoner:
@axel7083 @jeffmaury @benoitf can you please check?

benoitf (changes requested):
issue: it's not clear to me the problem being solved.

Is it for the traces ? because I see comments about stopping showing duplicated error messages but the logic is to stop reconnecting.
If it's just for traces then it should only have the if block around the console.warn item

about the logic, if the problem is debounce, backoff, etc or other things it seems we will not retry to connect

packages/main/src/plugin/container-registry.ts
```diff
@@ -550,7 +558,9 @@ export class ContainerProviderRegistry {
             }),
           );
         } catch (error) {
-          console.log('error in engine', provider.name, error);
+          if (this.notify) {
+            console.log('error in engine', provider.name, error);
+          }
```
benoitf: nit: replace all if console.log by a method notifyConsole('message') that will do the if

benoitf (approved):

gastoner:
@benoitf updated!

benoitf (approved):

ScrewTSW (approved):
LGTM

PR #10241, opened 2024-12-05
fix: visibility of docker compatibility settings

### LINKED ISSUE #9559: Docker Compatibility page is show when we enable the setting but does not disappear when we disable it
### Bug description

When on the settings page enabling the setting will display the entry in the navbar but disabling it does not remove it

### Operating system

MacOS

### Installation Method

None

### Version

1.13.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Fixes visibility of docker compatibility settings item when enabling/disabling

### Screenshot / video of UI

https://github.com/user-attachments/assets/56a1fe87-fefb-478e-b11c-1e9fe5552172

Closes #9559 

### How to test this PR?
Try to enable/disable docker compatibility

### DIFF
--- a/packages/renderer/src/PreferencesNavigation.spec.ts
+++ b/packages/renderer/src/PreferencesNavigation.spec.ts
@@ -24,6 +24,7 @@ import type { TinroRouteMeta } from 'tinro';
 import { beforeEach, expect, test, vi } from 'vitest';
 
 import PreferencesNavigation from './PreferencesNavigation.svelte';
+import { configurationProperties } from './stores/configurationProperties';
 
 // fake the window.events object
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


benoitf (approved):

benoitf:
issue: it's missing a unit test

jeffmaury (commented):
LGTM

Having a unit test would be welcomed

gastoner:
@benoitf @jeffmaury added tests, but I'm not sure why one test is failing (if I remove `await` before vi.waitFor it passes 😕) any ideas?

packages/renderer/src/PreferencesNavigation.spec.ts
```diff
@@ -102,3 +103,92 @@ test('Test rendering of the compatibility docker page is hidden if disabled', as
   const dockerCompatLink = screen.queryByRole('link', { name: 'Docker Compatibility' });
   expect(dockerCompatLink).toBeNull();
 });
+
+test('Test rendering of the compatibility docker page does change if config changes from enabled to disabled', async () => {
+  // mock window.getConfigurationValue
+  vi.mocked(window.getConfigurationValue<boolean>).mockClear();
+  vi.mocked(window.getConfigurationValue<boolean>).mockResolvedValueOnce(true);
+  vi.mocked(window.getConfigurationValue<boolean>).mockResolvedValueOnce(false);
```
jeffmaury:
```suggestion
  vi.mocked(window.getConfigurationValue<boolean>).mockResolvedValue(false);
```

jeffmaury:
> @benoitf @jeffmaury added tests, but I'm not sure why one test is failing (if I remove `await` before vi.waitFor it passes 😕) any ideas?

I think the test is not well designed as you expect an event to be sent to the view but you are only changing the mock behavior

> @benoitf @jeffmaury added tests, but I'm not sure why one test is failing (if I remove `await` before vi.waitFor it passes 😕) any ideas?

See my proposition

gastoner:
Yes I'm mocking return value of `getConfigurationValue` which should give me value each time configuration changes, thanks to `subscribe` method in the code. And I'm "triggering" change in configuration calling `configurationProperties.set([]);`. So the test should simulate changing the visibility of DC page from preferences. Or am I wrong?

jeffmaury:
> Yes I'm mocking return value of `getConfigurationValue` which should give me value each time configuration changes, thanks to `subscribe` method in the code. And I'm "triggering" change in configuration calling `configurationProperties.set([]);`. So the test should simulate changing the visibility of DC page from preferences. Or am I wrong?

My proposition fixes the issue, it changes from second call return false to second call and following return false so I would guest there are several calls not only 2 but don't know why

gastoner:
Oh my bad :D Yes there are multiple calls for getting the config, but I don't know why. I think during a startup there are multiple triggers of the subscriber function

jeffmaury (approved):
LGTM

benoitf (approved):

PR #10251, opened 2024-12-06
feat: force use of libPod if nvidia device requested

mhdawson:
### What does this PR do?

Requesting an nvidia device through the container toolkit is only supported through the libPod API not the compatibility API. Recognize the request and force libPod in a similar manner to what's done when the container is associated with a pod (it seems like a similar situation in that association with a pod is only available through libPod and not the compatibility API) .

This is the issue I used to ask about how to request an nvidia device available through the container toolkit in podman - https://github.com/containers/podman/issues/24687. From that, my understanding is that it only works through the libPod API and I don't see an interest on the podman side to change that.

See, https://github.com/podman-desktop/podman-desktop/pull/10166 for more context as to one of the use cases where this is needed.

### Screenshot / video of UI

N/A

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -1941,7 +1941,18 @@ export class ContainerProviderRegistry {
     let telemetryOptions = {};
     try {
       let container: Dockerode.Container;
-      if (options.pod) {
+      let forceLibPod = false;
+
+      // the device option requesting an nvidia gpu on linux only works
+      // if the LibPod API is used. Check if such a device is requested
+      // and if so force the use of LibPod
+      for (const device of options.HostConfig?.Devices ?? []) {
+        if (device.PathOnHost === 'nvidia.com/gpu=all') {
+          forceLibPod = true;
+          break;
+        }
+      }
+      if (options.pod ?? forceLibPod) {
         container = await this.createContainerLibPod(engineId, options);
       } else {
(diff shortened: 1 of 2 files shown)


mhdawson:
Rebased

packages/main/src/plugin/container-registry.ts
```diff
@@ -1941,7 +1941,20 @@ export class ContainerProviderRegistry {
...
     try {
       let container: Dockerode.Container;
-      if (options.pod) {
+      let forceLibPod = false;
+
+      // the device option requesting an nvidia gpu on linux only works
+      // if the LibPod API is used. Check if such a device is requested
+      // and if so force the use of LibPod
+      if (options.HostConfig?.Devices) {
+        for (const device of options.HostConfig?.Devices ?? []) {
```
benoitf: hello, I'm not sure why we're checking if there is Devices and then on the next line, we also assume it can be empty and we default to []

looks like the if condition is useless

mhdawson: Good point, will see if I can remove it

mhdawson: @benoitf removed the extra line and also rebased.

mhdawson:
@benoitf I'm off on PTO starting Thursday. If you need me to rebase before them just let me know.

benoitf (approved):

axel7083 (approved):
LGTM thanks for the change

Issue #8482, opened 2024-08-15
Execute image push operation as a background process


agiertli:
### Is your enhancement related to a problem? Please describe

When I initiate "image push" operation, the pop up window is displayed.
This pop up window does not displays any progress of the operation (if you execute push using cli, you see _some_ progress).

Moreover, while the push operation is in the progress, the rest of the podman desktop is not accessible - one needs to wait for push to finish, which is cumbersome if you are pushing large images.

### Describe the solution you'd like

- [x] Execute the push operation as background operation
- [x] Allow rest of the application be accessible while the push is ongoing
- [ ] Display some sort of progress information in the main page (from where the push was initiated) with the ability to drill down to the push operation logs

### Additional context

Push Dialog implementation ([link](https://github.com/containers/podman-desktop/blob/main/packages/renderer/src/lib/image/PushImageModal.svelte))
Example of using Tasks to save state of operation ([link](https://github.com/containers/podman-desktop/blob/87c5065fefc26a7ff207e1f180b01a8fe36be32a/packages/main/src/plugin/index.ts#L2106))
Also look at [PreferencesProviderRendering.svelte](https://github.com/containers/podman-desktop/blob/main/packages/renderer/src/lib/preferences/PreferencesProviderRendering.svelte)

agiertli:
<img width="1344" alt="issue" src="https://github.com/user-attachments/assets/cb5c94d5-32a7-4b95-a1c0-ebea2fb03d9e">
Attaching the problematic pop up window - no progress information displayed, window blocks the rest of the application, making it inaccessible while the push is going on,

slemeur:
Thanks for the report @agiertli

dgolovin:
@benoitf @deboer-tim @jeffmaury When one selects 'Push Image' action for image it opens modal dialog for push. I can start background process and create task in Tasks View and close dialog, but there would be no way to add link for the task to return back to modal push dialog to see the log. 

Would implementing Push Image similar to Pull Image solve the issue? If there is page for Push image command then it would be possible to open it from task view and show logs for the push command.

deboer-tim:
I think we technically _could_ reopen the dialog, but I agree it would be pretty odd. This is the only case I can think of where we have a dialog that starts a potentially long-running background task, so IMHO switching to a form is probably the simplest and most consistent way to fix this.

rujutashinde:
[...]

PR #10266, opened 2024-12-06
fix: push modal dialog error processing

### LINKED ISSUE #6893: 'Push Image' command from Images page has no exception handling
### Bug description

When 'Push Image' command invoked from image drop down menu fails it does not report error to user. It just keeps going forever.

https://github.com/containers/podman-desktop/assets/620330/0b400f84-6399-40a1-8c8f-eeedc23449e5

The code has no exception handling for window.pushImage call 

https://github.com/containers/podman-desktop/blob/7401b2e3472fc594c783eae805a64ccaf5fd3e80/packages/renderer/src/lib/image/PushImageModal.svelte#L80

### Operating system

macOS M1 Sonoma 14.3

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

dgolovin:
### What does this PR do?

This PR fixes error processing for imagePush call on main package side and generated required messages that push dialog on renderer side is expecting in case of error.

Fixes #6893 to allow work on #8482, so task would not stay forever in progrees task manager in case of error

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/index.spec.ts
+++ b/packages/main/src/plugin/index.spec.ts
@@ -20,10 +20,11 @@
 import { EventEmitter } from 'node:events';
 import { tmpdir } from 'node:os';
 
-import type { BrowserWindow, WebContents } from 'electron';
-import { clipboard, shell } from 'electron';
+import type { WebContents } from 'electron';
+import { app, BrowserWindow, clipboard, ipcMain, shell } from 'electron';
 import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
+import { Updater } from '/@/plugin/updater.js';
 import type { NotificationCardOptions } from '/@api/notification.js';
 
 import { securityRestrictionCurrentHandler } from '../security-restrictions-handler.js';
@@ -31,11 +32,14 @@ import type { TrayMenu } from '../tray-menu.js';
 import type { ApiSenderType } from './api.js';
 import { CancellationTokenRegistry } from './cancellation-token-registry.js';
 import type { ConfigurationRegistry } from './configuration-registry.js';
+import { ContainerProviderRegistry } from './container-registry.js';
 import type { Directories } from './directories.js';
 import { Emitter } from './events/emitter.js';
+import { ExtensionLoader } from './extension-loader.js';
 import { PluginSystem } from './index.js';
 import type { MessageBox } from './message-box.js';
 import { Deferred } from './util/deferred.js';
+import { HttpServer } from './webview/webview-registry.js';
 
 let pluginSystem: TestPluginSystem;
 
@@ -67,9 +71,18 @@ beforeAll(() => {
       },
       app: {
         on: vi.fn(),
+        getVersion: vi.fn(),
       },
       clipboard: {
         writeText: vi.fn(),
+      },
+      ipcMain: {
+        handle: vi.fn(),
+        emit: vi.fn().mockReturnValue(true),
+        on: vi.fn(),
+      },
+      BrowserWindow: {
+        getAllWindows: vi.fn(),
       },
     };
   });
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/image/PushImageModal.spec.ts
```diff
@@ -148,38 +148,143 @@ async function waitRender(customProperties: object): Promise<void> {
...
+    });
+
+    if (step === 'DialogOpened') return;
+
+    const pushButton = screen.getByRole('button', { name: 'Push image' });
+    await fireEvent.click(pushButton);
+
+    if (step === 'PushPressed') return;
+
+    callback!('first-message');
```
benoitf: issue: please, never use `!` at all, use `?` (and in all subsequent calls)

packages/renderer/src/lib/image/PushImageModal.svelte
```diff
@@ -108,7 +108,7 @@ $: window
         </p>{/if}
     </div>
 
-    <div class="h-[185px]" hidden={initTerminal === false}>
+    <div class="h-[185px]" hidden={initTerminal === false} role="term">
```
benoitf: question: why the role term would be on this one and not inside the TerminalWindow object which seems to be more legit to be the terminal ?

packages/main/src/plugin/index.ts
```diff
@@ -1097,9 +1097,15 @@ export class PluginSystem {
...
-        return containerProviderRegistry.pushImage(engine, imageId, (name: string, data: string) => {
-          this.getWebContentsSender().send('container-provider-registry:pushImage-onData', callbackId, name, data);
-        });
+        const msgName = 'container-provider-registry:pushImage-onData';
+        return containerProviderRegistry
+          .pushImage(engine, imageId, (name: string, data: string) => {
+            this.getWebContentsSender().send(msgName, callbackId, name, data);
+          })
+          .catch((error: unknown) => {
+            this.getWebContentsSender().send(msgName, callbackId, 'data', JSON.stringify({ error: String(error) }));
```
benoitf: question: why the error would be on the data channel and not on the error channel ?
usually we're sending the data on a specific onError handler

`JSON.stringify({ error: String(error) })` looks like a little bit overkill to me vs providing the `String(error)` into the error callback

packages/renderer/src/lib/image/PushImageModal.svelte
dgolovin: moved to terminal component

packages/main/src/plugin/index.ts
dgolovin: fixed renderer and main side

benoitf (approved):

packages/main/src/plugin/index.spec.ts
```diff
@@ -284,3 +314,39 @@ test('configurationRegistry propagated', async () => {
...
+  let registeredCallback: (name: string, data: string) => void | undefined;
+  vi.spyOn(ContainerProviderRegistry.prototype, 'pushImage').mockImplementation(
+    (_engine, _imageId, callback: (name: string, data: string) => void) => {
+      registeredCallback = callback;
+      return Promise.resolve();
+    },
+  );
+  await handle(undefined, 'podman', 'registry.com/repo/image:latest', 1);
+  expect(registeredCallback!).not.equal(undefined);
+  registeredCallback!('data', 'push image output');
```
benoitf: issue: please, do not use !

packages/main/src/plugin/index.ts
```diff
@@ -1101,9 +1101,15 @@ export class PluginSystem {
...
       async (_listener, engine: string, imageId: string, callbackId: number): Promise<void> => {
-        return containerProviderRegistry.pushImage(engine, imageId, (name: string, data: string) => {
-          this.getWebContentsSender().send('container-provider-registry:pushImage-onData', callbackId, name, data);
-        });
+        const msgName = 'container-provider-registry:pushImage-onData';
+        return containerProviderRegistry
+          .pushImage(engine, imageId, (name: string, data: string) => {
+            this.getWebContentsSender().send(msgName, callbackId, name, data);
+          })
+          .catch((error: unknown) => {
```
jeffmaury, line 1109: I don't get why the error is not transfered through the promise and needs this special processing

packages/main/src/plugin/index.spec.ts
dgolovin: I missed that one :(

packages/main/src/plugin/index.ts
dgolovin: I was not going to improve implementation, I just wanted to make current implementation to work.
The renderer part does not not need full error object, just a text of the error. It is much simpler to send it in form of string 'error' message.

jeffmaury: Yes but why don't throwing this string instead of implementing a complex exchange mechanism ?

dgolovin: Opened #10572 for simpler error processing here.

jeffmaury (approved):
LGTM

PR #10267, opened 2024-12-07
fix: create task when running push image command

### LINKED ISSUE #8482: Execute image push operation as a background process

dgolovin:
### What does this PR do?

Creates Task that appears in Task Manager UI. It does not show real progress but it finishes successfully or with error, so if you closed Push Image dialog you know how it ended.

### Screenshot / video of UI

https://github.com/user-attachments/assets/4a41a59b-a12b-4b7b-ba98-9a014bdceaad

Fixes #8482.

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/index.spec.ts
+++ b/packages/main/src/plugin/index.spec.ts
@@ -38,6 +38,8 @@ import { Emitter } from './events/emitter.js';
 import { ExtensionLoader } from './extension-loader.js';
 import { PluginSystem } from './index.js';
 import type { MessageBox } from './message-box.js';
+import { TaskManager } from './tasks/task-manager.js';
+import { Disposable } from './types/disposable.js';
 import { Deferred } from './util/deferred.js';
 import { HttpServer } from './webview/webview-registry.js';
 
@@ -91,8 +93,10 @@ beforeAll(() => {
 });
 
 const handlers = new Map<string, any>();
+
 beforeEach(() => {
   vi.clearAllMocks();
+  handlers.clear();
   vi.mocked(ipcMain.handle).mockImplementation((channel: string, listener: any) => {
     handlers.set(channel, listener);
   });
@@ -105,7 +109,7 @@ beforeEach(() => {
     ];
   });
   vi.mocked(app.getVersion).mockReturnValue('100.0.0');
-  vi.spyOn(Updater.prototype, 'init').mockReturnValue({ dispose: vi.fn() } as any);
+  vi.spyOn(Updater.prototype, 'init').mockReturnValue(new Disposable(vi.fn()));
   vi.spyOn(ExtensionLoader.prototype, 'readDevelopmentFolders').mockResolvedValue([]);
   // to avoid port conflict when tests are running on windows host
   vi.spyOn(HttpServer.prototype, 'start').mockImplementation(vi.fn());
@@ -318,7 +322,8 @@ test('configurationRegistry propagated', async () => {
 const pushImageHandlerId = 'container-provider-registry:pushImage';
 const pushImageHandlerOnDataEvent = `${pushImageHandlerId}-onData`;
 
-test('push image command sends onData message with callbackId, event name and event data ', async () => {
+test('push image command sends onData message with callbackId, event name and data, mark task as success on end event', async () => {
+  const createTaskSpy = vi.spyOn(TaskManager.prototype, 'createTask');
   await pluginSystem.initExtensions(new Emitter<ConfigurationRegistry>());
   const handle = handlers.get(pushImageHandlerId);
   expect(handle).not.equal(undefined);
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/index.spec.ts
```diff
@@ -67,18 +72,44 @@ beforeAll(() => {
...
+      BrowserWindow: {
+        getAllWindows: vi.fn(),
+      },
     };
   });
   const trayMenuMock = {} as unknown as TrayMenu;
   pluginSystem = new TestPluginSystem(trayMenuMock, mainWindowDeferred);
 });
 
+const handlers = new Map<string, any>();
```
feloy, line 95: I think you will need to reset its value in `beforeEach`

axel7083 (commented):
Screenshot From 2024-12-09 16-05-26

Would it possible to improve the error ? IMO the [object Object] is a bit problematic, we should either have something with more info or omit it

dgolovin:
> Would it possible to improve the error ? IMO the [object Object] is a bit problematic, we should either have something with more info or omit it

This is a problem ether on dockerode modem side (see https://github.com/apocas/docker-modem/issues/171) or quay.io.

packages/main/src/plugin/index.spec.ts
dgolovin, line 95:
```suggestion
const handlers = new Map<string, any>();

```

packages/main/src/plugin/index.spec.ts
```diff
@@ -67,18 +72,44 @@ beforeAll(() => {
...
 beforeEach(() => {
   vi.clearAllMocks();
+  vi.mocked(ipcMain.handle).mockImplementation((channel: string, listener: any) => {
```
dgolovin, line 100:
```suggestion
  handlers.clear();
  vi.mocked(ipcMain.handle).mockImplementation((channel: string, listener: any) => {
```

packages/main/src/plugin/index.spec.ts
dgolovin: @feloy good catch. Thanks!

axel7083 (approved):

dgolovin:
This PR does not implement all the feats from the issue, it creates task and reports error in Task Manager, but does not provide the way to reopen push command to see the log. I'll create follow up issue for it.

dgolovin:
@axel7083 @deboer-tim Does it make sense to have a button 'Run in Background`? Right now one can hide dialog by clicking anywhere outside of it and it is not obvious.

deboer-tim:
> @axel7083 @deboer-tim Does it make sense to have a button 'Run in Background`? Right now one can hide dialog by clicking anywhere outside of it and it is not obvious.

As I commented in https://github.com/podman-desktop/podman-desktop/issues/8482#issuecomment-2413877408 this dialog is pretty unique and I'd rather see it (e.g.) changed to a form vs spending time trying to fix individual usability issues. That's longer term though, so if it's minor to change the button I'd support it, but should be separate PR.

packages/main/src/plugin/index.spec.ts
```diff
@@ -67,18 +72,46 @@ beforeAll(() => {
...
+  vi.mocked(BrowserWindow.getAllWindows).mockImplementation(() => {
+    return [
+      {
+        isDestroyed: () => false,
+        webContents,
+      } as unknown as BrowserWindow,
+    ];
+  });
+  vi.mocked(app.getVersion).mockReturnValue('100.0.0');
+  vi.spyOn(Updater.prototype, 'init').mockReturnValue({ dispose: vi.fn() } as any);
```
jeffmaury: Is any required here ?

dgolovin: Turned out there is no need to mock. I just returned an instance of original class.

gastoner (approved):
LGTM, tested on Windows
image

PR #10284, opened 2024-12-09
fix(refactor): refactor Powershell 5 interactions in a separate file

### LINKED ISSUE #10281: Refactor code to isolate Powershell interactions

In order to better support Powershell (5 vs 7), we first need to isolate Powershell interactions into a separate file

Refactor the code and isolate every Powershell call into a single file

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #10281

### What does this PR do?

refactor code so that Powershell 5 interface is isolated in a separate file

### Screenshot / video of UI

N/A

#10281

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/podman-install.ts
+++ b/extensions/podman/packages/extension/src/podman-install.ts
@@ -43,6 +43,7 @@ import { PodmanCleanupWindows } from './podman-cleanup-windows';
 import type { InstalledPodman } from './podman-cli';
 import { getPodmanCli, getPodmanInstallation } from './podman-cli';
 import * as podman5JSON from './podman5.json';
+import type { PowerShellClient } from './powershell';
 import { getPowerShellClient } from './powershell';
 import { getAssetsFolder, normalizeWSLOutput } from './util';
 import { WslHelper } from './wsl-helper';
@@ -666,9 +667,17 @@ export class VirtualMachinePlatformCheck extends BaseCheck {
 }
 
 abstract class WindowsCheck extends BaseCheck {
+  #powerShellClient?: Promise<PowerShellClient> = undefined;
+  protected async getPowerShellClient(): Promise<PowerShellClient> {
+    if (!this.#powerShellClient) {
+      this.#powerShellClient = getPowerShellClient();
+    }
+    return this.#powerShellClient;
+  }
+
   async isUserAdmin(): Promise<boolean> {
     try {
-      return await (await getPowerShellClient()).isUserAdmin();
+      return (await this.getPowerShellClient()).isUserAdmin();
     } catch (err: unknown) {
       return false;
     }
@@ -893,7 +902,7 @@ export class HyperVCheck extends WindowsCheck {
 
   private async isPodmanDesktopElevated(): Promise<boolean> {
     try {
-      return await (await getPowerShellClient()).isRunningElevated();
+      return (await this.getPowerShellClient()).isRunningElevated();
     } catch (err: unknown) {
       return false;
     }
@@ -901,7 +910,7 @@ export class HyperVCheck extends WindowsCheck {
 
   private async isHyperVinstalled(): Promise<boolean> {
     try {
-      return await (await getPowerShellClient()).isHyperVInstalled();
+      return (await this.getPowerShellClient()).isHyperVInstalled();
     } catch (err: unknown) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


extensions/podman/packages/extension/src/podman-install.ts
```diff
@@ -674,10 +668,7 @@ export class VirtualMachinePlatformCheck extends BaseCheck {
 abstract class WindowsCheck extends BaseCheck {
   async isUserAdmin(): Promise<boolean> {
     try {
-      const { stdout: res } = await extensionApi.process.exec('powershell.exe', [
-        '$null -ne (whoami /groups /fo csv | ConvertFrom-Csv | Where-Object {$_.SID -eq "S-1-5-32-544"})',
-      ]);
-      return res.trim() === 'True';
+      return await (await getPowerShellClient()).isUserAdmin();
```
benoitf: I think we reached the maximum of await instructions
we can remove the first one

benoitf (changes requested):
I think the powershellClient should be a field of the checker, not a function that we call all the time

```
return this.powershellClient.IsUserAdmin()
```

vs
```
await (await getPowerShellClient()).isUserAdmin();
```


then, I don't understand why getPowerShellClient is returning a promise and is being async as it's just returning a new instance of PowerShell5Client.

but let say we need to change how it's built later (from sync to async, if it's a field of the class it's only one await to add rather than changing all the calls)


If a function returns a Promise, we can return a promise
no need to do: return await promise;

extensions/podman/packages/extension/src/podman-install.ts
jeffmaury: Fixed

extensions/podman/packages/extension/src/podman-install.ts
```diff
@@ -672,12 +667,17 @@ export class VirtualMachinePlatformCheck extends BaseCheck {
...
+    return this.#powerShellClient;
+  }
+
   async isUserAdmin(): Promise<boolean> {
     try {
-      const { stdout: res } = await extensionApi.process.exec('powershell.exe', [
-        '$null -ne (whoami /groups /fo csv | ConvertFrom-Csv | Where-Object {$_.SID -eq "S-1-5-32-544"})',
-      ]);
-      return res.trim() === 'True';
+      return (await this.getPowerShellClient()).isUserAdmin();
```
dgolovin: Try/catch block is not needed around client calls, because client methods already handle the exceptions.

dgolovin (changes requested):

extensions/podman/packages/extension/src/podman-install.ts
jeffmaury: getPowerShellClient may return an exception in the future when we start implementing PowerShell 7 support

dgolovin (approved):

benoitf (commented):
it looks like we still have a double await and that the client needs to be aware of the powershellclient


I would expect to have in podman-install only:

await isVirtualMachineAvailable();



then in an exported function isVirtualMachineAvailable

```ts
await powerShell = getPowerShellClient();
return powerShell.isVirtualMachineAvailable()
```

and getPowerShellClient() can be globally lazy

jeffmaury:
> it looks like we still have a double await and that the client needs to be aware of the powershellclient
> 
> I would expect to have in podman-install only:
> 
> await isVirtualMachineAvailable();
> 
> then in an exported function isVirtualMachineAvailable
> 
> ```ts
> await powerShell = getPowerShellClient();
> return powerShell.isVirtualMachineAvailable()
> ```
> 
> and getPowerShellClient() can be globally lazy

Fixed by 3478208

extensions/podman/packages/extension/src/powershell.ts
```diff
@@ -0,0 +1,92 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import * as extensionApi from '@podman-desktop/api';
+
+export async function getPowerShellClient(): Promise<PowerShellClient> {
+  return new PowerShell5Client();
+}
```
benoitf, line 22: later improvement: cache the value

benoitf (approved):
approving on the logic but wasn't able to try the function as I'm on macOS

extensions/podman/packages/extension/src/powershell.ts
jeffmaury: Yes I planned it when implementing PowerShell 7 support

inntran:
When I saw powershell.exe under podman desktop, it scared me a bit until I found this.

PR #10287, opened 2024-12-09
chore: add telemetry events to GitHubIssue feedback forms

SoniaSandler:

### What does this PR do?
This PR adds telemetry events to track when a user opens a bug or a feature feedback form, and when they click on the `Preview on GitHub` button. (as request in https://github.com/podman-desktop/podman-desktop/issues/9584#issuecomment-2466104205)

related to https://github.com/podman-desktop/podman-desktop/issues/9584

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.svelte
+++ b/packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.svelte
@@ -60,13 +60,22 @@ async function previewOnGitHub(): Promise<void> {
     includeSystemInfo: $state.snapshot(includeSystemInfo),
     includeExtensionInfo: $state.snapshot(includeExtensionInfo),
   };
-  try {
-    await window.previewOnGitHub(issueProperties);
-    onCloseForm(false);
-    await window.telemetryTrack(`feedback.FormSubmitted`, { feedbackCategory: category });
-  } catch (error: unknown) {
-    console.error('There was a problem with preview on GitHub', error);
-  }
+  let telemetryEventProperties: { [property: string]: unknown } = { feedbackCategory: category };
+
+  window
+    .previewOnGitHub(issueProperties)
+    .then(() => {
+      onCloseForm(false);
+    })
+    .catch((error: unknown) => {
+      telemetryEventProperties['error'] = error;
+      console.error('There was a problem with preview on GitHub', error);
+    })
+    .finally(() => {
+      window
+        .telemetryTrack(`feedback.FormSubmitted`, telemetryEventProperties)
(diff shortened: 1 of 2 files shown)


benoitf (commented):
two questions: 

1. Do we need a different event for each category or the category is an attribute
b/c If I wand to know how many people are opening the form, I need to sum up all different categoryFormOpened ?

2.  I'm wondering why the telemetry is inside GithubFeedback
Like shouldn't it be part of the SendFeedback page ?
here, if I click on the toggle to change the category, it'll send multiple times the FormOpened event but I'm just switching the category.

SoniaSandler:
> 1. Do we need a different event for each category or the category is an attribute
>    b/c If I wand to know how many people are opening the form, I need to sum up all different categoryFormOpened ?

@benoitf The original request was for the bug form, so I wasn't sure if we would like to track the separate categories or just the feedback form usage.
> 2. I'm wondering why the telemetry is inside GithubFeedback
>    Like shouldn't it be part of the SendFeedback page ?
>    here, if I click on the toggle to change the category, it'll send multiple times the FormOpened event but I'm just switching the category.

This was also a part of the request, to see if users finish the flow or abandon the form before opening github, and since the `Preview on GitHub` button is in `GitHubFeeback`, I added it there.

@slemeur lmk what do you think about this and if there is something you'd like to adjust or add.

packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.svelte
```diff
@@ -41,6 +42,10 @@ let existingIssuesLink = $state(
 
 $effect(() => contentChange(Boolean(issueTitle || issueDescription)));
 
+onMount(async () => {
+  await window.telemetryTrack(`feedback.${category}FormOpened`);
```
jeffmaury: I would recommend category to be a property of the event and event named feedback.formOpened

@slemeur WDYT ?

slemeur: Yeah that would make sense.

SoniaSandler: updated!

jeffmaury (commented):
If would prefer to have the submitted event before calling previewOnGitHub before if something wrong happens in previewOnGitHub  we'll miss that event

Or even better call the telemetry in the finally block and if there is an error add it to the event properties so that we know when errors are happening but I'm not sure we will make heavy use of it :)

benoitf (commented):
looks like @jeffmaury comments have not been addressed yet

slemeur (approved):

jeffmaury (approved):
LGTM

PR #10309, opened 2024-12-11
feat: added option to show svg files in addition to fa icons

gastoner:
### What does this PR do?
Adds option to show svg icon 

Required for https://github.com/containers/podman-desktop-extension-ai-lab/pull/2246

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/settingsNavItem/IconTest.svelte
+++ b/packages/ui/src/lib/settingsNavItem/IconTest.svelte
@@ -1,0 +1,5 @@
+<div role="img">
+    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+        <path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2z" />
+    </svg>
+</div>
--- a/packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
+++ b/packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
@@ -1,6 +1,9 @@
 <script lang="ts">
 import type { IconDefinition } from '@fortawesome/free-regular-svg-icons';
+import type { Component } from 'svelte';
 import Fa from 'svelte-fa';
+
+import { isFontAwesomeIcon } from '../utils/icon-utils';
 
 interface Props {
   title: string;
@@ -9,7 +12,7 @@ interface Props {
   expanded?: boolean;
   child?: boolean;
   selected?: boolean;
-  icon?: IconDefinition;
+  icon?: IconDefinition | Component;
   onClick?: () => void;
 }
 
@@ -66,7 +69,12 @@ function click(): void {
     class:hover:border-[var(--pd-secondary-nav-text-hover-bg)]={!selected}>
     <span class="group-hover:block flex flex-row items-center" class:capitalize={!child}>
       {#if icon}
-        <Fa class="mr-4" icon={icon} />
+        {#if isFontAwesomeIcon(icon)}
+          <Fa class="mr-4" {icon} />
+        {:else}
+          {@const Icon = icon}
+          <Icon size="14" class="mr-4"/>
+        {/if}
       {/if}
       {title}
     </span>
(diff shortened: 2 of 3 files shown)


packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
```diff
@@ -9,7 +10,7 @@ interface Props {
   expanded?: boolean;
   child?: boolean;
   selected?: boolean;
-  icon?: IconDefinition;
+  icon?: IconDefinition | SvelteComponent;
```
axel7083: I think `SvelteComponent` is a legacy type and `Component` should be used ?

packages/ui/src/lib/settingsNavItem/Icon.svelte
```diff
@@ -0,0 +1,5 @@
+<div role="img">
```
axel7083: Rename this file to `IconTest.svelte` since it is only used in a spec file

packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
```diff
@@ -46,6 +47,10 @@ function click(): void {
   expanded = !expanded;
   onClick();
 }
+
+function isIconDefinition(icon: IconDefinition | SvelteComponent): icon is IconDefinition {
```
axel7083: This is something we are using in multiple place (from the top of my memory) might be nice to have a `icon-utils.ts` file somewhere

gastoner:
@axel7083 I'm not sure about rendering the image in some size, as it is not. Should it be, hopefully okay?

packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
```diff
@@ -66,7 +69,12 @@ function click(): void {
     class:hover:border-[var(--pd-secondary-nav-text-hover-bg)]={!selected}>
     <span class="group-hover:block flex flex-row items-center" class:capitalize={!child}>
       {#if icon}
-        <Fa class="mr-4" icon={icon} />
+        {#if isFontAwesomeIcon(icon)}
+          <Fa class="mr-4" {icon} />
+        {:else}
+          <!-- svelte-ignore svelte_component_deprecated -->
```
axel7083: Hummm, can't we avoid that ?

packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
```diff
@@ -66,7 +69,12 @@ function click(): void {
...
+          <svelte:component this={icon} size={'14'} class="mr-4"/>
```
axel7083: `<icon size=.../>` might work ?

packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
gastoner: I was able to make it work only with the <svelte:component> and because it is deprecated, it needs to have this 
If I try something like <icon size.. class..></icon> I think that the Component is not rendered (assuming that the test is not passing)

packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
benoitf: (I think you might need to edit your comment as the `<` and `>` are stripped from the rendering

axel7083: hihi yes

packages/ui/src/lib/settingsNavItem/IconTest.svelte
```diff
@@ -0,0 +1,5 @@
+<div role="img">
+    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+        <path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2z" />
+    </svg>
+</div>
```
benoitf: missing EOF/EOL

packages/ui/src/lib/settingsNavItem/SettingsNavItem.svelte
```diff
@@ -1,15 +1,18 @@
 <script lang="ts">
 import type { IconDefinition } from '@fortawesome/free-regular-svg-icons';
+import type { Component } from 'svelte';
 import Fa from 'svelte-fa';
 
+import { isFontAwesomeIcon } from '../utils/icon-utils';
```
benoitf, line 6: nit: use `/@/` aliases imports if possible rather than relative imports

benoitf (approved):

axel7083 (approved):
LGTM

PR #10310, opened 2024-12-11
chore: create a command for pullImage and add a task to it

### LINKED ISSUE #9989: Create/export a command pullImage so 3rd party extensions could call it

Podman Desktop provides some functionalities and some behavior.

Some extension, in particular bootc extension, want to provides in its UI a button to pull an image but behind this button, it should behave exactly if I would have called the pullImage command from the UI of Podman Desktop.

It means, for example the task should be created and I should see the task in the task manager, etc.

Provides and register the pullImage providing the pull of a given image

_No response_

### Additional context

While there is an API to pull an image, here the extension want to replicate the behavior of pulling an image from Podman Desktop.

So for example the extension endpoint will not create a task (it's up to the caller of the endpoint), etc.

more discussion around the topic can be found there: https://github.com/podman-desktop/podman-desktop/issues/7398

SoniaSandler:

### What does this PR do?
This PR adds a `pullImage` command so that extensions could use it to have the same pullImage behavior as Podman Desktop UI, in addition to adding a task to track the pullImage status.

### Screenshot / video of UI
Screenshot from 2024-12-10 22-02-21

### How to test this PR?
Closes https://github.com/podman-desktop/podman-desktop/issues/9989

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -700,7 +700,7 @@ declare module '@podman-desktop/api' {
   /**
    * The commands module provides functions to register and execute commands
    * Existing command available for extensions to use:
-   * - `pullImage`: uses Podman Desktop's UI pull image behavior. This command will create a visible task to show the progress of the pulImage action.
+   * - `pullImage`: uses Podman Desktop's UI pull image behavior. This command will create a visible task to show the progress of the pulImage action with the option to include a task action.
    *
    * @example
    * ```typescript
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -146,16 +146,14 @@ declare module '@podman-desktop/api' {
    * }
    * ```
    */
-  export interface Event<T> {
-    /**
-     * @param listener The listener function will be called when the event happens.
-     * @param thisArgs The `this`-argument which will be used when calling the event listener.
-     * @param disposables An array to which the resulting {@link Disposable} will be added.
-     * @return A disposable which unsubscribes the event listener.
-     */
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    (listener: (e: T) => any, thisArgs?: any, disposables?: Disposable[]): Disposable;
-  }
+  /**
+   * @param listener The listener function will be called when the event happens.
+   * @param thisArgs The `this`-argument which will be used when calling the event listener.
+   * @param disposables An array to which the resulting {@link Disposable} will be added.
+   * @return A disposable which unsubscribes the event listener.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  export type Event<T> = (listener: (e: T) => any, thisArgs?: any, disposables?: Disposable[]) => Disposable;
 
   /**
    * A class to create and manage an {@link Event} for clients to subscribe to.
@@ -701,6 +699,8 @@ declare module '@podman-desktop/api' {
 
   /**
    * The commands module provides functions to register and execute commands
+   * Existing command available for extensions to use:
+   * - `pullImage`: uses Podman Desktop's UI pull image behavior. This command will create a visible task to show the progress of the pulImage action.
    *
    * @example
    * ```typescript
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -703,6 +703,7 @@ declare module '@podman-desktop/api' {
    * The commands module provides functions to register and execute commands
    * Existing command available for extensions to use:
    * - `pullImage`: uses Podman Desktop's UI pull image behavior. This command will create a visible task to show the progress of the pulImage action with the option to include a task action.
+   * It uses the same parameters as the original pullImage function, in addition to having `taskActionName: string` and `taskActionCallback: () => void` as parameters to create a task action (optional).
    *
    * @example
    * ```typescript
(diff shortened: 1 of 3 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -146,16 +146,14 @@ declare module '@podman-desktop/api' {
    * }
    * ```
    */
-  export interface Event<T> {
```
SoniaSandler: updated automatically when committing

packages/main/src/plugin/index.ts
```diff
@@ -1075,6 +1076,40 @@ export class PluginSystem {
...
     );
+
+    commandRegistry.registerCommand(
+      'pullImage',
+      async (
+        providerContainerConnectionInfo: ProviderContainerConnectionInfo,
+        imageName: string,
+        callbackId: number,
+        platform?: string,
+        taskAction?: TaskAction,
```
SoniaSandler: Do we want to give extensions the option to define a task action? and if so, we'll either have to expose TaskAction in the API or I can break this parameter into optional taskActionName and taskActionExecute parameters, and only if both are provided, we can create a task action inside the command

axel7083: We do not expose the `TaskAction` in the API, so might be better to use a callback ? just a function.

Moreover, since this might be called from an extension, we should have a safeguard, checking all arguments, to be sure they are defined (for those which are required), and their type.

Should the command be hidden ? Because currently I think it would be visible in the command palette, but it would probably create an error ?

axel7083: Idially we might want the argument to be _similar_ to https://podman-desktop.io/docs/containers/images/pulling-an-image for consistency

packages/extension-api/src/extension-api.d.ts
jeffmaury: Should be removed why is this happening ?

jeffmaury (commented):
Is there a way to add unit test(s) ?

packages/main/src/plugin/index.ts
jeffmaury: Seems to me this is the implementation here not what is exposed to extensions ?

packages/extension-api/src/extension-api.d.ts
SoniaSandler: @jeffmaury it looks like this has been changed in https://github.com/podman-desktop/podman-desktop/blob/badc68f73b16d6df79542794072a63f3dad68418/packages/main/src/plugin/events/emitter.ts#L36 , and now that I modified this file, it automatically updated other changes?

packages/main/src/plugin/index.ts
SoniaSandler: @jeffmaury extensions have different behavior for pulling images, but this PR adds a command that will allow extensions (if they choose) to use the same pull image behavior as the Podman Desktop UI. The current default implementation of pull image for extensions was not meant to be changed in this PR.

packages/extension-api/src/extension-api.d.ts
jeffmaury: I think this is just of reformat so should be removed from your PR

packages/main/src/plugin/index.ts
SoniaSandler: > Should the command be hidden ? Because currently I think it would be visible in the command palette, but it would probably create an error ?

@axel7083 I checked, and this command is not visible. It looks like none, or at least most of the commands that are created in PD core are not visible in the command palette

packages/main/src/plugin/index.ts
```diff
@@ -1083,6 +1084,59 @@ export class PluginSystem {
         return containerProviderRegistry.resolveShortnameImage(providerContainerConnectionInfo, shortName);
       },
     );
+
+    commandRegistry.registerCommand(
+      'pullImage',
+      async (
+        providerContainerConnectionInfo: ProviderContainerConnectionInfo,
+        imageName: string,
+        callbackId: number,
```
benoitf: issue: how an extension would know which callbackId to set there ?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -701,6 +701,9 @@ declare module '@podman-desktop/api' {
 
   /**
    * The commands module provides functions to register and execute commands
+   * Existing command available for extensions to use:
+   * - `pullImage`: uses Podman Desktop's UI pull image behavior. This command will create a visible task to show the progress of the pulImage action with the option to include a task action.
```
gastoner: `pullImage`?

axel7083 (approved):
I am not a big fan of adding commands to overcome some limitation we have with the tasks system, but let's go with it for now

LGTM

packages/main/src/plugin/index.ts
```diff
@@ -1100,6 +1101,52 @@ export class PluginSystem {
...
+      ) => {
+        if (!providerContainerConnectionInfo || !imageName || !callback || typeof callback !== 'function') {
+          return;
+        }
+
+        let taskAction: TaskAction | undefined;
+
+        if (
+          taskActionName &&
+          typeof taskActionName === 'string' &&
```
gastoner: isn't this redundant? `typeof taskActionName === 'string'` should cover the `taskActionName` too, or am I wrong?
same thing for the statement below

gastoner (approved):
Windows: 
image

packages/main/src/plugin/index.ts
SoniaSandler: I think you might be right about the function check, but for `taskActionName`, for an empty string, `typeof taskActionName === 'string'` will return true but `taskActionName` will return false

benoitf (approved):
non blocking remark: could be a later improvement but I would expect that calling the command without an image would prompt me which image I would want

no provider = take the first one or prompt me one to select if multiple

benoitf (approved):
same remark as the last review

axel7083 (approved):
LGTM

PR #10311, opened 2024-12-11
fix: test errors due to missing window.matchMedia function

dgolovin:
### What does this PR do?

Fixes unit test errors in renderer tests.

### How to test this PR?


benoitf (changes requested):
we should not use hoisted pattern

dgolovin:
> we should not use hoisted pattern

@benoitf That the only pattern worked.

Please review another approach.

PR #10318, opened 2024-12-12
fix: test errors after migration to svelte@5.11.2

dgolovin:
### What does this PR do?

Fixes test errors due to missing window.matchMedia method.

### How to test this PR?


axel7083 (approved):

packages/renderer/vite.tests.setup.js
```diff
@@ -17,3 +17,5 @@
  ***********************************************************************/
 
 import 'vitest-canvas-mock';
+
+global.window.matchMedia = () => {};
```
benoitf, line 21: question: should it be `vi.fn()` rather than `() => {}`; as we're in the mocking parts

benoitf (approved):

packages/renderer/vite.tests.setup.js
dgolovin: It is never mocked in tests directly, so it does not matter IMO.

dgolovin: Merged this commit to new dependabot pr for svelte@5.12.0

PR #10322, opened 2024-12-12
feat: additional config value to hide recommended banners 

### LINKED ISSUE #9415: Make it possible to hide the banner promoting extensions

The banner for promoting extensions is always displayed, until the user has installed the extension. But that can be frustrating for the user, as there is no way to hide the banner, if that user is not interested by the "promoted" extension.

- Provide a way to hide the banner
- Make the choice persistent each time podman desktop is restarted 

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds option to hide promoted extensions banners
PR 1/2 - needs to be merged first - follow up - https://github.com/podman-desktop/podman-desktop/pull/10414 
These are just backend changes

Required for #9415 

### How to test this PR?
Run Tests

### DIFF
--- a/packages/main/src/plugin/recommendations/recommendations-registry.ts
+++ b/packages/main/src/plugin/recommendations/recommendations-registry.ts
@@ -21,10 +21,10 @@ import type { ExtensionLoader } from '/@/plugin/extension-loader.js';
 import type { ExtensionsCatalog } from '/@/plugin/extensions-catalog/extensions-catalog.js';
 import type { Featured } from '/@/plugin/featured/featured.js';
 import type { FeaturedExtension } from '/@/plugin/featured/featured-api.js';
-import type {
-  ExtensionBanner,
-  RecommendedRegistry,
-  RecommendedRegistryExtensionDetails,
+import {
+  type ExtensionBanner,
+  type RecommendedRegistry,
+  type RecommendedRegistryExtensionDetails,
 } from '/@/plugin/recommendations/recommendations-api.js';
 
 import { default as recommendations } from '../../../../../recommendations.json';
@@ -38,6 +38,14 @@ export class RecommendationsRegistry {
     private extensionsCatalog: ExtensionsCatalog,
   ) {}
 
+  isBannerRecommendationEnabled(): boolean {
+    const bannersIgnored = !this.configurationRegistry
+      .getConfiguration(RecommendationsSettings.SectionName)
+      .get<boolean>(RecommendationsSettings.IgnoreBannerRecommendations, false);
+    const recommendationsIgnored = this.isRecommendationEnabled();
+    return bannersIgnored && recommendationsIgnored;
+  }
+
   isRecommendationEnabled(): boolean {
     return !this.configurationRegistry
       .getConfiguration(RecommendationsSettings.SectionName)
@@ -49,6 +57,7 @@ export class RecommendationsRegistry {
     if (!this.isRecommendationEnabled()) {
       return [];
     }
+
     const installedExtensions = await this.extensionLoader.listExtensions();
 
     const fetchableExtensions = await this.extensionsCatalog.getFetchableExtensions();
@@ -86,7 +95,7 @@ export class RecommendationsRegistry {
    */
   async getExtensionBanners(limit = 1): Promise<ExtensionBanner[]> {
     // Do not recommend any extension when user selected the ignore preference
-    if (!this.isRecommendationEnabled()) return [];
+    if (!this.isBannerRecommendationEnabled()) return [];
 
     const featuredExtensions: Record<string, FeaturedExtension> = Object.fromEntries(
       (await this.featured.getFeaturedExtensions()).map(featured => [featured.id, featured]),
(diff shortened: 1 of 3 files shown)


slemeur:
This is looking great, thanks @gastoner ! 

I think it would be better if the settings was not as granular, and instead of being a setting per banner, being a setting for all the extension recommendation banners. 

Also - remember adding the telemetry on the following actions:
- clicking on "close" button on the recommendation banner
- clicking on the buttons that are in the confirmation popup

packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
```diff
@@ -46,6 +66,9 @@ $effect(() => {
 
   <!-- feature extension actions -->
   <div class="flex flex-col">
+    <div class="flex flex-row justify-end">
+      <CloseButton on:click={onClose} />
```
axel7083: image

The close button is not visible in light mode

Might be interesting to have a color property on the CloseButton, currently there is not way to avoid the `text-gray-800`.

https://github.com/podman-desktop/podman-desktop/blob/516922e2ef043c208d55e9921edda79f7798aa20/packages/ui/src/lib/button/CloseButton.svelte#L15

A nice way would be to use component scoped variable (here is an example https://svelte.dev/playground/bb42c17d585b4236824e1d5e48fe92b5?version=3.38.0)

we could add `--text-color`, `--text-hover` etc.?

gastoner: I've created a PR for setting color properties of Close button

packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
```diff
@@ -22,6 +24,24 @@ $effect(() => {
     style = undefined;
   }
 });
+
+async function onClose(): Promise<void> {
+  await window.telemetryTrack('hideRecomendationExtension.popupOpened', {
```
jeffmaury:
```suggestion
  await window.telemetryTrack('hideRecommendationExtension.popupOpened', {
```

packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
```diff
@@ -22,6 +24,24 @@ $effect(() => {
...
+    title: 'Hide extension recommendation banners',
+    message: `Do you want to hide extension recommendation banners?`,
+    type: 'warning',
+    buttons: [`Don't hide`, 'Hide'],
+  });
+
+  if (result && result.response === 1) {
+    await window.updateConfigurationValue(`extensions.ignoreBannerRecommendations`, true, 'DEFAULT');
+
+    await window.telemetryTrack('hideRecomendationExtension.confirmationClicked');
```
jeffmaury:
```suggestion
    await window.telemetryTrack('hideRecommendationExtension.confirmationClicked');
```

benoitf (changes requested):
correct me if I'm wrong but it looks like that now it's only checking the fine grained pattern while it needs to be 'global' + fine grained

ignoreRecommendations = true, ignore everything
ignoreBannerRecommendations = ignore only banner

so either I can ignore the whole thing or a single parameter

benoitf (changes requested):
the title of the PR is misleading

it says 'add an option to hide' but it's renaming one option and also adding one for registries
(so it would break all existing settings being saved)

And the PR to the svelte file can be a separate PR, as it'll just set the value of an existing property.

packages/main/src/plugin/recommendations/recommendations-registry.ts
```diff
@@ -38,6 +38,12 @@ export class RecommendationsRegistry {
     private extensionsCatalog: ExtensionsCatalog,
   ) {}
 
+  isBannerRecommendationEnabled(): boolean {
+    return !this.configurationRegistry
+      .getConfiguration(RecommendationsSettings.SectionName)
+      .get<boolean>(RecommendationsSettings.IgnoreBannerRecommendations, false);
+  }
+
```
benoitf: issue: hello, on my end, I'm expecting that this method would be in charge of  checking both the specific flag and the global flag rather than having to do the check inside the getExtensionBanners method

gastoner: Updated

packages/main/src/plugin/recommendations/recommendations-registry.ts
```diff
@@ -145,6 +154,12 @@ export class RecommendationsRegistry {
           default: false,
           hidden: false,
         },
+        [RecommendationsSettings.SectionName + '.' + RecommendationsSettings.IgnoreBannerRecommendations]: {
```
benoitf: nit: I think we have some rules encouraging to use string literals rather than concatenation

benoitf (approved):

packages/main/src/plugin/recommendations/recommendations-settings.ts
```diff
@@ -19,4 +19,5 @@
 export enum RecommendationsSettings {
   SectionName = 'extensions',
   IgnoreRecommendations = 'ignoreRecommendations',
+  IgnoreBannerRecommendations = 'ignoreBannerRecommendations',
```
axel7083, line 22: sidenote, not related to this PR, but we should try avoid repetition in enum, the enum is named `RecommendationsSettings` we can probably simplify `IgnoreRecommendations` to `IgnoreAll`, and `IgnoreBannerRecommendations` to `IgnoreBanner`.

(Let's keep this for now, as we don't want to change the `IgnoreRecommendations` in this PR)

axel7083 (approved):
LGTM codewise

packages/main/src/plugin/recommendations/recommendations-settings.ts
benoitf: as long as the value is not changed yes we can make variable shorter 👍 (and of course in other PRs)

packages/main/src/plugin/recommendations/recommendations-registry.ts
gastoner: Updated

benoitf (approved):
still approving 👍

PR #10339, opened 2024-12-12
fix: toggle of help menu

### LINKED ISSUE #10154: Help menu doesn't close when clicking icon again

cdrage:
fix: toggle of help menu

### What does this PR do?

Fixes toggling the help menu

### Screenshot / video of UI

https://github.com/user-attachments/assets/87babb2c-ffa5-44ac-9c2a-ccbdd258f4a4

Closes https://github.com/podman-desktop/podman-desktop/issues/10154

### How to test this PR?

Toggle the help menu button, it should close now

### DIFF
--- a/packages/renderer/src/lib/help/HelpActions.svelte
+++ b/packages/renderer/src/lib/help/HelpActions.svelte
@@ -19,9 +19,9 @@ function handleEscape({ key }: KeyboardEvent): void {
 }
 
 function onWindowClick(e: any): void {
-  // We listen to the click event on anything BUT the question circle icon which is
-  // where the help menu is. This is to ensure the menu is closed if the user clicks the circle again.
-  if (!e.srcElement.classList.contains('fa-question-circle')) {
+  const target = e.target as HTMLElement;
+  // Listen to anything **but** the button that has "data-task-button" attribute with a value of "Help"
+  if (target && target.getAttribute('data-task-button') !== 'Help') {
     showMenu = outsideWindow.contains(e.target);
   }
 }
--- a/packages/renderer/src/lib/help/HelpActions.svelte
+++ b/packages/renderer/src/lib/help/HelpActions.svelte
@@ -19,7 +19,11 @@ function handleEscape({ key }: KeyboardEvent): void {
 }
 
 function onWindowClick(e: any): void {
-  showMenu = outsideWindow.contains(e.target);
+  // We listen to the click event on anything BUT the question circle icon which is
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/help/HelpActions.svelte
```diff
@@ -19,7 +19,11 @@ function handleEscape({ key }: KeyboardEvent): void {
 }
 
 function onWindowClick(e: any): void {
-  showMenu = outsideWindow.contains(e.target);
+  // We listen to the click event on anything BUT the question circle icon which is
+  // where the help menu is. This is to ensure the menu is closed if the user clicks the circle again.
+  if (!e.srcElement.classList.contains('fa-question-circle')) {
+    showMenu = outsideWindow.contains(e.target);
+  }
```
benoitf: issue: the unit test is not covering this part

unit test should create a element in the dom, add the fa-question-circle and click on this element (that should trigger the onWindowClick, etc.

ideally the fa-question-circle should come from the definition of the status bar entry being added for the help menu

b/c let's say tomorrow we change the status bar icon for the help menu from a question-circle to another icon, this unit test will continue to work but the app won't work (as the icon would have changed)

cdrage: > issue: the unit test is not covering this part
> 
> unit test should create a element in the dom, add the fa-question-circle and click on this element (that should trigger the onWindowClick, etc.
> 
> ideally the fa-question-circle should come from the definition of the status bar entry being added for the help menu
> 
> b/c let's say tomorrow we change the status bar icon for the help menu from a question-circle to another icon, this unit test will continue to work but the app won't work (as the icon would have changed)

You are 100% right, this test doesn't cover it at all and that's my mistake. Sorry!

I'll re-write the test and have it inject the fa-question-circle so you can click that and it'll toggle correctly.

gastoner (approved):
Tested on Windows

cdrage:
@benoitf Did some refactoring and added more tests (some were missing) to enhance helpactions. Found out why I was having difficulty implementing it and it's because of how we mock `window.events?.receive('toggle-help-menu', toggleMenu);` within the tests. Fixed it via manually mocking each test instead of generic beforeEach()

benoitf (approved):

PR #10345, opened 2024-12-13
feat: start monitoring manually on non current contexts

### LINKED ISSUE #10323: Monitoring all Kubernetes clusters raises too much authentication
### Bug description

Podman Desktop currently tries to monitor all Kubernetes contexts (start informers for Pods/Deployments in all contexts) when Podman Desktop starts. This is problematic for many users, as the first step of starting informers is to authenticate to the cluster, and for clusters from cloud providers, a token needs to be re-negociated regularly. Depending on the cloud provider, this may open browser tabs to ask to authenticate with cloud provider credentials, etc.

A proposal is to have a 'Connect/Disconnect' button for each context in the Settings > Kubernetes page, so the user can start the monitoring for each context independently, and manage the authentication at this moment.

cc @slemeur 

### Operating system

all

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

feloy:

### What does this PR do?

Start monitoring manually on non current contexts

Send telemetry when:
- monitoring is started for a non-current context (from Settings > Kubernetes page)
- monitoring is retsarted from the current context (from any Kubernetes resource page) 

Note that, when changing the current context, all monitoring started manually are stopped, and not restarted. This may be covered in follow-up PR, depending on the usage seen in telemetry

### Screenshot / video of UI

kubecontext-connect

Fixes #10323 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/kube/KubeContextUI.ts
+++ b/packages/renderer/src/lib/kube/KubeContextUI.ts
@@ -36,21 +36,17 @@ export function setKubeUIContextError(contexts: Context[], contextName: string, 
     if (ctx.name === contextName) {
       return { ...ctx, error: String(error) };
     } else {
-      return ctx;
+      return { ...ctx };
     }
   });
 }
 
-export function clearKubeUIContextErrors(contexts: Context[]): Context[] {
-  return contexts.map(ctx => ({ ...ctx, error: undefined }));
-}
-
-export function clearKubeUIContextError(contexts: Context[], contextName: string): Context[] {
+export function clearKubeUIContextErrors(contexts: Context[], contextName?: string): Context[] {
   return contexts.map(ctx => {
-    if (ctx.name === contextName) {
+    if (!contextName || ctx.name === contextName) {
       return { ...ctx, error: undefined };
     } else {
-      return ctx;
+      return { ...ctx };
     }
   });
 }
--- a/packages/renderer/src/lib/kube/KubeContextUI.ts
+++ b/packages/renderer/src/lib/kube/KubeContextUI.ts
@@ -44,3 +44,13 @@ export function setKubeUIContextError(contexts: Context[], contextName: string, 
 export function clearKubeUIContextErrors(contexts: Context[]): Context[] {
   return contexts.map(ctx => ({ ...ctx, error: undefined }));
 }
+
+export function clearKubeUIContextError(contexts: Context[], contextName: string): Context[] {
+  return contexts.map(ctx => {
+    if (ctx.name === contextName) {
+      return { ...ctx, error: undefined };
+    } else {
+      return ctx;
+    }
+  });
+}
--- a/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
@@ -111,10 +111,8 @@ function isContextBeingChecked(contextName: string, experimental: boolean): bool
   return !!$kubernetesContextsCheckingStateDelayed?.get(contextName);
 }
 
-async function startMonitoring(contextName: string, contextNumber: number): Promise<void> {
-  await window.telemetryTrack('kubernetes.monitoring.start.non-current', {
-    contextNumber,
-  });
+async function startMonitoring(contextName: string): Promise<void> {
+  await window.telemetryTrack('kubernetes.monitoring.start.non-current');
   $kubernetesContexts = clearKubeUIContextErrors($kubernetesContexts, contextName);
   window.kubernetesRefreshContextState(contextName).catch((e: unknown) => {
     if (e instanceof Error) {
@@ -141,7 +139,7 @@ async function startMonitoring(contextName: string, contextNumber: number): Prom
         Go to Resources
       </Button>
     </EmptyScreen>
-    {#each kubernetesContextsWithStates as context, iContext}
+    {#each kubernetesContextsWithStates as context}
       <!-- If current context, use lighter background -->
       <div
         role="row"
@@ -227,7 +225,7 @@ async function startMonitoring(contextName: string, contextNumber: number): Prom
                     {/if}                    
                   </div>
                   {#if !$kubernetesContextsState.get(context.name)}
-                    <div><Button on:click={() => startMonitoring(context.name, iContext)}>Start monitoring</Button></div>
+                    <div><Button on:click={() => startMonitoring(context.name)}>Start monitoring</Button></div>
                   {/if}
                 </div>
               {/if}
--- a/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
@@ -9,7 +9,7 @@ import { kubernetesContextsCheckingStateDelayed, kubernetesContextsState } from 
 import type { KubeContext } from '/@api/kubernetes-context';
 
 import { kubernetesContexts } from '../../stores/kubernetes-contexts';
-import { clearKubeUIContextErrors, setKubeUIContextError } from '../kube/KubeContextUI';
+import { clearKubeUIContextError, clearKubeUIContextErrors, setKubeUIContextError } from '../kube/KubeContextUI';
 import EngineIcon from '../ui/EngineIcon.svelte';
 import ListItemButtonIcon from '../ui/ListItemButtonIcon.svelte';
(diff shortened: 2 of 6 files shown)


packages/renderer/src/lib/kube/KubeContextUI.ts
```diff
@@ -44,3 +44,13 @@ export function setKubeUIContextError(contexts: Context[], contextName: string,
...
 }
+
+export function clearKubeUIContextError(contexts: Context[], contextName: string): Context[] {
+  return contexts.map(ctx => {
+    if (ctx.name === contextName) {
+      return { ...ctx, error: undefined };
+    } else {
+      return ctx;
+    }
+  });
```
benoitf: question: I'm wondering, as we don't really return a copy of the parameters why there is a return of the array.
Either it's always returning a copy of objects or if it's just updating the objects, should be void

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -61,6 +61,18 @@ async function handleDeleteContext(contextName: string) {
     }
   }
 }
+
+async function startMonitoring(contextName: string, contextNumber: number): Promise<void> {
+  await window.telemetryTrack('kubernetes.monitoring.start.non-current', {
+    contextNumber,
+  });
```
benoitf: question: not sure the goal of providing the context number ?

packages/renderer/src/lib/kube/KubeContextUI.ts
```diff
@@ -44,3 +44,13 @@ export function setKubeUIContextError(contexts: Context[], contextName: string,
 export function clearKubeUIContextErrors(contexts: Context[]): Context[] {
   return contexts.map(ctx => ({ ...ctx, error: undefined }));
 }
+
+export function clearKubeUIContextError(contexts: Context[], contextName: string): Context[] {
```
benoitf: question: wondering if it can't be merged with the previous function

ie: If I provide a contextName parameter then it'll reset only for a given context

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
```diff
@@ -11,9 +11,10 @@ import {
 let currentContextName = $derived($kubernetesContexts.find(c => c.currentContext)?.name);
 let error = $state('');
 
-function refresh(): void {
+async function refresh(): Promise<void> {
   error = '';
   if (currentContextName) {
+    await window.telemetryTrack('kubernetes.monitoring.start.current');
```
benoitf: question: wondering here like if we could send if there is a current error on the current state, what was the last time the check occurred, etc

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: The goal is to indicate if the user repeats the operation on a specific context, without giving its real name. But the index won't be fixed when deleting contexts, perhaps a hash would be better

packages/renderer/src/lib/kube/KubeContextUI.ts
feloy: ok, I changed to always return a clone

packages/renderer/src/lib/kube/KubeContextUI.ts
feloy: ok, I merged both functions

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
feloy: @slemeur @Firewall what would be the information of interest, for this telemetry and for the other (when the user asks to start the monitoring of a non-current context)?

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
benoitf: but I mean, does this information matters

like maybe it's more interesting to know the number of existing contexts as well I don't know. I mean, I'm not sure what we could extract knowing the index of a context

packages/renderer/src/lib/kube/KubeContextUI.ts
benoitf: ok thx

gastoner (commented):
https://github.com/user-attachments/assets/cb2c5ac3-8611-41b7-96ef-9ce1168e5188

Is expected to "restart monitoring" when switching default context?

feloy:
> Screen.Recording.2024-12-13.155204.mp4 
> Is expected to "restart monitoring" when switching default context?

Yes, this is expected for now:

> Note that, when changing the current context, all monitoring started manually are stopped, and not restarted. This may be covered in follow-up PR, depending on the usage seen in telemetry

gastoner (approved):

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
Firewall: What @benoitf mentions would definitely be useful. Maybe this is not a property on this event, but knowning the total amount of connections that the user has would also be useful.

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: I'm removing the index for now, and keeping only the `kubernetes.monitoring.start.non-current` code, as the main purpose is to know if users are using this button. If they do, we will be able to add more information

packages/renderer/src/lib/ui/KubernetesCheckConnection.svelte
feloy: I'm removing this telemetry, as it is not really related to the PR

axel7083:
I have a shared opinion, I understand the _need_ but from a user POV, it does not really make sense, "Monitoring" mean nothing to a user ?

I feel like a user would only expect "connect | disconnect"... 

cc @slemeur

feloy:
> I have a shared opinion, I understand the _need_ but from a user POV, it does not really make sense, "Monitoring" mean nothing to a user ?
> 
> I feel like a user would only expect "connect | disconnect"...
> 
> cc @slemeur

I'm trying to not use the `connect` | `disconnet` terms, but other terms like `reachable` | `unreachable`, as Kubernetes is not really a connected environment from a developer point of view: the developer is briefly connecting to the cluster only when running kubectl or similar commands. 

IMHO, the fact that we are using watchers (informers) which are, them, connected, is implementation-specific and should not be visible for the user. cc @slemeur @Firewall WDYT?

Firewall:
@feloy I also expected `connect | disconnect` to be honest or otherwise a `refresh` could also make sense. We only need a text for when the user is not monitoring, right? When we are successfully monitoring we show the number of pods and deployments.

feloy:
> @feloy I also expected `connect | disconnect` to be honest or otherwise a `refresh` could also make sense. We only need a text for when the user is not monitoring, right? When we are successfully monitoring we show the number of pods and deployments.

Yes, the button will appear only when the monitoring is failing.

OK, I'll change to connect / disconnect

feloy:
@benoitf @axel7083 @Firewall I have made changes in this PR, it is ready for review

axel7083 (approved):
Tested LGTM

> Note that, when changing the current context, all monitoring started manually are stopped, and not restarted. This may be covered in follow-up PR, depending on the usage seen in telemetry

I would expect to have a disconnect button in follow-up PR too !

feloy:
> Tested LGTM
> 
> > Note that, when changing the current context, all monitoring started manually are stopped, and not restarted. This may be covered in follow-up PR, depending on the usage seen in telemetry
> 
> I would expect to have a disconnect button in follow-up PR too !

@slemeur @Firewall WDYT?

benoitf:
hello @feloy this PR has broken the CI
old PR should be rebased before getting merged
it doesn't pass the any check validation

benoitf:
FYI the error is being reported in https://github.com/podman-desktop/podman-desktop/actions/runs/12766231704/job/35582148291?pr=10520

feloy:
> FYI the error is being reported in https://github.com/podman-desktop/podman-desktop/actions/runs/12766231704/job/35582148291?pr=10520

OK, sorry for the issue. I'm working on a fix

edit: https://github.com/podman-desktop/podman-desktop/pull/10635

Issue #10226, opened 2024-12-04
Settings: List of experimental features


slemeur:
### Is your feature request related to a problem? Please describe

This issue will be the implementation of the list of experimental features corresponding to the mockup in issue #10223 

### Describe the solution you'd like

This issue will be the implementation of the list of experimental features corresponding to the mockup in issue #10223

axel7083:
Hummm let's try to break down this issue

![Image](https://github.com/user-attachments/assets/539b489e-4731-4c3c-b8b4-de3d92fda09a)

Given the mock-ups in https://github.com/podman-desktop/podman-desktop/issues/10223 we have a dedicated section in the preferences, this is a bit odd as we are moving out of the preferences, the experimental preferences but let's go with it!

Here is the code we are using for the Task Manager _experimental_ features.[^1]

[^1]: https://github.com/podman-desktop/podman-desktop/blob/a5c44196c94905e56991cb6c03e35f7c97084f92/packages/main/src/plugin/tasks/task-manager.ts#L61

```ts
    this.configurationRegistry.registerConfigurations([
      {
        id: 'preferences.experimental.tasks',
        title: 'Tasks',
        type: 'object',
        properties: {
          [`${ExperimentalTasksSettings.SectionName}.${ExperimentalTasksSettings.StatusBar}`]: {
            description: 'Show running tasks in the status bar',
            type: 'boolean',
            default: false,
          },
        },
      },
    ]);
```

We have several possibilities, we could add a `scope: 'Experimental'` here is today the list of scopes available

https://github.com/podman-desktop/podman-desktop/blob/106f199487abba30767d4d01f36191725bee28c4/packages/main/src/plugin/configuration-registry.ts#L78-L85

> I must admit I am not sure how the scopes are used, but could be a nice filtering, let's explore that

PR #10346, opened 2024-12-13
refactor(ui): preferences navigation component

axel7083:
### What does this PR do?

While working on https://github.com/podman-desktop/podman-desktop/issues/10226 I studied the preferences component because I need to create a new one, but found some serious issues.

This is a refactor / cleaning / fixing the `PreferencesNavigation.svelte` file. Here is a break down of the problem with it

**big problem with the Map value type**

Here it is declared as `Map<string, { id: string; title: string }>` as shown bellow

https://github.com/podman-desktop/podman-desktop/blob/e9eb09390bed6c7f96fef2eebfdf66a416839b62/packages/renderer/src/PreferencesNavigation.svelte#L13

but later, it is cast to any, to be able to push an array inside

https://github.com/podman-desktop/podman-desktop/blob/e9eb09390bed6c7f96fef2eebfdf66a416839b62/packages/renderer/src/PreferencesNavigation.svelte#L39

**Object.entries(Map) lose the type**

The following lose the type, and prevent typescript to do the job, as both `configSection` and `configItems` are cast to any

https://github.com/podman-desktop/podman-desktop/blob/e9eb09390bed6c7f96fef2eebfdf66a416839b62/packages/renderer/src/PreferencesNavigation.svelte#L80

Moreover we don't need to use `Object.entries` on a map, as there is `Map#entries` function available, which keep the types.

Moreover we access the map item using `[]` when we should be using `Map#get` and `Map#set`.

**unnecessary any everywhere**

https://github.com/podman-desktop/podman-desktop/blob/e9eb09390bed6c7f96fef2eebfdf66a416839b62/packages/renderer/src/PreferencesNavigation.svelte#L19-L21

https://github.com/podman-desktop/podman-desktop/blob/e9eb09390bed6c7f96fef2eebfdf66a416839b62/packages/renderer/src/PreferencesNavigation.svelte#L30

https://github.com/podman-desktop/podman-desktop/blob/e9eb09390bed6c7f96fef2eebfdf66a416839b62/packages/renderer/src/PreferencesNavigation.svelte#L37

Migrate to svelte 5 because it ease the refactor.

### Screenshot / video of UI

N/A

Preamble to https://github.com/podman-desktop/podman-desktop/issues/10226

### How to test this PR?

There are a tons of tests in `PreferencesNavigation.spec.ts` which ensure the behaviour is the same.

### DIFF
--- a/packages/renderer/src/PreferencesNavigation.svelte
+++ b/packages/renderer/src/PreferencesNavigation.svelte
@@ -40,7 +40,7 @@ function updateDockerCompatibility(): void {
 }
 
 function sortItems(items: NavItem[]): NavItem[] {
-  return items.toSorted((a: { title: string }, b: { title: string }) => a.title.localeCompare(b.title));
+  return items.toSorted((a, b) => a.title.localeCompare(b.title));
 }
 
 onMount(() => {
--- a/packages/renderer/src/PreferencesNavigation.svelte
+++ b/packages/renderer/src/PreferencesNavigation.svelte
@@ -8,52 +8,63 @@ import { ExperimentalSettings } from '/@api/docker-compatibility-info';
 
 import { configurationProperties } from './stores/configurationProperties';
 
-export let meta: TinroRouteMeta;
-
-let configProperties: Map<string, { id: string; title: string }>;
-
-$: configProperties = new Map();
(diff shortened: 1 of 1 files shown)


gastoner (approved):
Tested on Windows

PR #10349, opened 2024-12-13
feat: experimental health frontend

### LINKED ISSUE #9938: Use health check to check Kubernetes cluster connectivity

feloy:
### What does this PR do?

Display the health of kubernetes contexts in Settings > Kubernetes

Fixes #9938 

### How to test this PR?

Activate the experimental mode in `~/.local/share/containers/podman-desktop/configuration/settings.json`:

```
```

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
@@ -6,6 +6,7 @@ import { router } from 'tinro';
 
 import { kubernetesContextsHealths } from '/@/stores/kubernetes-context-health';
 import { kubernetesContextsCheckingStateDelayed, kubernetesContextsState } from '/@/stores/kubernetes-contexts-state';
+import type { KubeContext } from '/@api/kubernetes-context';
 
 import { kubernetesContexts } from '../../stores/kubernetes-contexts';
 import { clearKubeUIContextErrors, setKubeUIContextError } from '../kube/KubeContextUI';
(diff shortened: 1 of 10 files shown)


axel7083:
> Display the health of kubernetes contexts in Settings > Kubernetes

I guess you are talking about the following ?

image

axel7083 (approved):
LGTM

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -125,7 +158,7 @@ async function handleDeleteContext(contextName: string) {
         <div class="grow flex-column divide-gray-900 text-[var(--pd-invert-content-card-text)]">
           <div class="flex flex-row">
             <div class="flex-none w-36">
-              {#if $kubernetesContextsState.get(context.name)?.reachable}
+              {#if isContextReachable(context.name, experimentalStates)}
```
benoitf: I would say, that from UI side (html part), we should not call or do any computation at the display time, so no call to a map or a function, the data providing already all the conditions for the rendering.

so IMHO it should be part of the data we're receiving, so only

```svelte
{#if someObjectToDisplay.reachable}
   ....
{/if}
```svelte

and not utility function called at various places in the .svelte/html part

File: packages/renderer/src/stores/kubernetes-context-health.ts [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+
+export const kubernetesContextsHealths = readable<ContextHealth[]>([], set => {
+  window.events?.receive('kubernetes-contexts-healths', (value: unknown) => {
+    set(value as ContextHealth[]);
+  });
+  window
+    .kubernetesGetContextsHealths()
+    .then(value => set(value))
+    .catch((err: unknown) => console.log('Error getting contexts healths', err));
+});
```
benoitf: question: why this store is not using the new EventStore pattern so we can monitor and track the calls ?

feloy: I considered using it, but I hardly measure the cost of using it.

I'll look more deeply at it

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: I moved the logic out of the html. Is it what you expect, or do we want to create a store for this?

packages/renderer/src/stores/kubernetes-context-health.ts
feloy: ok, I replaced with an EventStore

benoitf (approved):
thanks, the enhancement in the svelte part is much more readable (on my end)

thanks for enhancing the store to use the EventStore pattern

Issue #9821, opened 2024-11-07
Code Guidelines


feloy:
### Is your feature request related to a problem? Please describe

Difficult to know which method to use for mocking in tests, how to use Svelte v5

### Describe the solution you'd like

Code guidelines for mocking in tests, svelte v5

In CONTRIBUTING.md

axel7083:
Axel notes, do not take too seriously, I will be updating this as I encountered problems/logic with code

## Vitest

### Try { } catch

We should probably never use `try catch` in `*.spec.ts` file. As we aim for deterministic tests, vitest provide us tools, and method to expect for errors.

```ts
test('expect error to be properly catched and verified', async () => {
  await expect(async () => {
    // replace Promise.reject by the async method you want to call that will
    // throw an error
    await Promise.reject(new Error('Dummy Error'));
  }).rejects.toThrowError('Dummy Error');
});
```

### Avoid using `screen` global

(Personal preference), when we want to `getByRole` an HTML element when testing we use the `screen` from `@testing-library/svelte` but here is an altertive I like

```diff
- import { render, screen } from '@testing-library/svelte';
+ import { render } from '@testing-library/svelte';

test('render something', async () => {
-  render(Link);
+  const { getByRole } = render(Link);
-  const element = screen.getByRole('link');
+  const element = getByRole('link');
  expect(link).toBeInTheDocument();
});
``` 

I do not like using the screen, because it is a global object with complex mechanism, by taking the result of the render function, we have access to all the function that are available on the `screen.*`.

## Svelte 5

### Declaring  props

It is easier and nice to read to have a multi line interface named Props.

```diff 
- let { varA, varB }: { varA: string, varB: boolean } = $props();
+ interface Props {
+    varA: string;
+   varB: boolean;
+ }
+ let { varA, varB }: Props  = $props();
```

> This is recommended inside the official documentation [typescript#Typing-$props](https://svelte.dev/docs/svelte/typescript#Typing-$props)

### Advanced HTML element genericity 

We may need to catch some of the HTML default attribute element, please do not add them to the Props interface: we can do better! Let's say I need to allow parent component to provide any html attribute.

```svelte
<script lang="ts">
import type { HTMLAttributes } from 'svelte/elements';
interface Props extends HTMLAttributes<HTMLElement> {
  varA?: number;
}

let { varA, ...restProps }: Props = $props();
</script>

<div {...restProps}>{varA}</div>
```

benoitf:
[...]

PR #10370, opened 2024-12-16
chore(code-guidelines): adding try catch example

axel7083:
### What does this PR do?

Another example for the code guidelines 

Related to https://github.com/podman-desktop/podman-desktop/issues/9821

### How to test this PR?

### DIFF
--- a/CODE-GUIDELINES.md
+++ b/CODE-GUIDELINES.md
@@ -37,3 +37,26 @@ test('...', () => {
   vi.mocked(window.windowMethod).mockResolvedValue('a string');
 });
 ```
+
+### Do not use `try {} catch` in tests
+
+The following code has a problem, if the code in the try block do not raise an error, we do not check the error message.
+
+```ts
+// bad
+test('expect error to be properly catched and verified', async () => {
+  try {
+    await Promise.reject(new Error('Dummy Error'));
+  } catch(err: unknown) {
+    expect(err.message).toBe('Dummy Error')
+  }
+});
+```
+
+Here is a version where we are sure the block rejected an error.
+
+```ts
+  await expect(async () => {
+    await Promise.reject(new Error('Dummy Error'));
+  }).rejects.toThrowError('Dummy Error');
+```
--- a/CODE-GUIDELINES.md
+++ b/CODE-GUIDELINES.md
@@ -40,23 +40,32 @@ test('...', () => {
 
 ### Do not use `try {} catch` in tests
 
-The following code has a problem, if the code in the try block do not raise an error, we do not check the error message.
+Let's have an async function such as
+
+```ts
+export function foo(): Promise<void> {
+  throw new Error('bar');
+}
+```
+
+Let's say we want to check that under certain condition the function throw an error (here always), we _could_ write the following code
 
 ```ts
 // bad
 test('expect error to be properly catched and verified', async () => {
   try {
-    await Promise.reject(new Error('Dummy Error'));
+    await foo();
   } catch(err: unknown) {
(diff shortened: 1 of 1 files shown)


CODE-GUIDELINES.md
```diff
@@ -37,3 +37,26 @@ test('...', () => {
...
+});
+```
+
+Here is a version where we are sure the block rejected an error.
+
+```ts
+  await expect(async () => {
+    await Promise.reject(new Error('Dummy Error'));
+  }).rejects.toThrowError('Dummy Error');
+```
```
benoitf: I would not use the Promise.reject inside the code directly but saying we're calling a method that is throwing an error

more like https://vitest.dev/api/expect#rejects

async function erroneousMethod(): Promise<void> {
 ...
  throw new Error('Dummy Error');
 }

 try {
    await erroneousMethod();
  } catch(err: unknown) {
    expect(err.message).toBe('Dummy Error')
  }

axel7083: Thanks, applied

CODE-GUIDELINES.md
```diff
@@ -37,3 +37,35 @@ test('...', () => {
...
+
+But it has a problem, if the code in the try block do not raise an error, we do not check the error message.
+Here is a version where we are sure the block rejected an error.
+
+```ts
+  await expect(async () => {
+    await foo();
+  }).rejects.toThrowError('Dummy Error');
+```
```
benoitf, line 71: from https://vitest.dev/api/expect#rejects it looks like it can be simplified 
(no async () => and you can use toThrow rather than toThrowError

```suggestion
  await expect(foo()).rejects.toThrow('Dummy Error');
```
```

File: CODE-GUIDELINES.md [hunk 3]
```diff
@@ -37,3 +37,35 @@ test('...', () => {
...
+});
+```
+
+But it has a problem, if the code in the try block do not raise an error, we do not check the error message.
+Here is a version where we are sure the block rejected an error.
+
+```ts
+  await expect(async () => {
+    await foo();
+  }).rejects.toThrowError('Dummy Error');
```
benoitf, line 70: from https://vitest.dev/api/expect#rejects it looks like it can be simplified 
(no async () => and you can use toThrow rather than toThrowError

```suggestion
  await expect(foo()).rejects.toThrow('Dummy Error');
```
```

File: CODE-GUIDELINES.md [hunk 4]
```diff
@@ -37,3 +37,35 @@ test('...', () => {
...
+```
+
+Let's say we want to check that under certain condition the function throw an error (here always), we _could_ write the following code
+
+```ts
+// bad
+test('expect error to be properly catched and verified', async () => {
+  try {
+    await foo();
+  } catch(err: unknown) {
```
gastoner, line 58: Can be probably ignored since we want to make it simple, 
but can't we use instead of `undefined` something like `Error`? Won't the linter yell at you that the undefined does not have a message attribute? (Thinking of creating an interface Error, or something like that). 

`err && typeof err === 'object' && 'message' in err && typeof err.message === 'string'`

gastoner (approved):

CODE-GUIDELINES.md
benoitf: maybe we just do `expect(String(err)).contains('bar')` or something else because I think the idea is to say that we should not use expect in a catch clause

It might be overkill to use all the syntax check to verify if err has a field message just for this purpose. I agree it's not compiling on the current code but maybe we can have a in-the-middle solution where we can still have something short and with a valid typechecking

Issue #10223, opened 2024-12-04
UX: Provide a list of experimental features in the Settings


slemeur:
### UX Description

In order to simplify the discoverability and management of the experimental features, we would like to introduce a list in the settings.

That list should be at the primary navigation level in the settings.
For the end-user the list must allow the following:
- Understand what the experimental feature enable and the use case it unlocks
- See the state of the experimental feature, if it's enable or disable
- Have an action enabling to share feedback

### Request type

Application mockups/designs

### Primary Contact

@slemeur

slemeur:
Some mockups have already been provided by @MariaLeonova.

![Image](https://github.com/user-attachments/assets/928adefe-a3f8-4193-a2cf-2b164a0ddcdf)

We need to further refine the mockup.

**Navigation:**
- The list of experimental features should be at the top level otherwise people are going to miss them. That could probably go under settings

**List of experimental features:**
I think the option 1 is probably the best one, but would like to further iterate on it. 
- the "Share feedback" is too easy to miss
- the icon for "thumbs up" or "thumbs down" are not very clear.

Should sharing a feedback be a 2 steps flow? 
- First the user share a "lazy feedback"
![Image](https://github.com/user-attachments/assets/acc2fffe-9101-40cf-ae83-0dca04f07844)
- Second we bring a "share your thoughts and join the discussion" popup, where the can tell a bit more.

MariaLeonova:
>The list of experimental features should be at the top level otherwise people are going to miss them. That could probably go under settings

Certainly can be! Should it come last on the list? The Setting items don't seem to be arranged in alphabetical order.

>Should sharing a feedback be a 2 steps flow?

My thoughts were: they click on the smiley / thumb up (down) /  share feedback link and the pop up comes up. In my experience, once I've selected some feedback, I get annoyed buy the suggestion to add more comments. It would be more honest to bring the whole pop up right away.

>the "Share feedback" is too easy to miss

I was following some of the existing patterns, such as: 
<img src="https://github.com/user-attachments/assets/45b49eb9-94d4-4f70-b07a-6ab5d7c7b7b4" width="500">
  
<img src="https://github.com/user-attachments/assets/7573fad0-1f29-4c5b-ad6a-3df3b4d7ae44" width="500">
    
<img src="https://github.com/user-attachments/assets/e218ff47-feec-462c-a917-83f6de986e10" width="500">

If those are not visible enough, the usual ways to **make UI elements stand out** are: 
- color
- size
- shape
[...]

PR #10371, opened 2024-12-16
feat: introducing experimental section

axel7083:
### What does this PR do?

First step of https://github.com/podman-desktop/podman-desktop/issues/10226.

This PR add a new `Experimental` section in the settings. I moved the docker compatibility and tasks to the experimental (as they are?).

❓ I hardcoded the the flask icon on the frontend, not sure if we want to add the full logic of adding an icon to the configuration.. 

### Screenshot / video of UI

image

image

Related to https://github.com/podman-desktop/podman-desktop/issues/10226

### How to test this PR?

### DIFF
--- a/packages/api/src/configuration/constants.ts
+++ b/packages/api/src/configuration/constants.ts
@@ -18,3 +18,8 @@
 
 export const CONFIGURATION_DEFAULT_SCOPE = 'DEFAULT';
 export const CONFIGURATION_ONBOARDING_SCOPE = 'Onboarding';
+
+export enum CONFIGURATION_SECTION {
+  PREFERENCES = 'preferences',
+  EXPERIMENTAL = 'experimental',
+}
--- a/packages/main/src/plugin/docker/docker-compatibility.spec.ts
+++ b/packages/main/src/plugin/docker/docker-compatibility.spec.ts
@@ -25,6 +25,7 @@ import type { ProviderContainerConnection } from '@podman-desktop/api';
 import type { Mock } from 'vitest';
 import { beforeAll, describe, expect, test, vi } from 'vitest';
 
+import { CONFIGURATION_SECTION } from '/@api/configuration/constants.js';
 import type { DockerSocketServerInfoType } from '/@api/docker-compatibility-info.js';
 import type { ProviderInfo } from '/@api/provider-info.js';
 
@@ -92,8 +93,8 @@ test('should register a configuration', async () => {
 
   expect(configurationRegistry.registerConfigurations).toBeCalled();
   const configurationNode = vi.mocked(configurationRegistry.registerConfigurations).mock.calls[0]?.[0][0];
-  expect(configurationNode?.id).toBe('preferences.experimental.dockerCompatibility');
-  expect(configurationNode?.title).toBe('Experimental (Docker Compatibility)');
+  expect(configurationNode?.id).toBe(`${CONFIGURATION_SECTION.EXPERIMENTAL}.dockerCompatibility`);
+  expect(configurationNode?.title).toBe('Docker Compatibility');
   expect(configurationNode?.properties).toBeDefined();
   expect(Object.keys(configurationNode?.properties ?? {}).length).toBe(1);
   expect(configurationNode?.properties?.[TestDockerCompatibility.ENABLED_FULL_KEY]).toBeDefined();
--- a/packages/main/src/plugin/docker/docker-compatibility.ts
+++ b/packages/main/src/plugin/docker/docker-compatibility.ts
@@ -21,6 +21,7 @@ import { promises } from 'node:fs';
 import Dockerode from 'dockerode';
 
 import { isMac, isWindows } from '/@/util.js';
+import { CONFIGURATION_SECTION } from '/@api/configuration/constants.js';
 import type {
   DockerContextInfo,
   DockerSocketMappingStatusInfo,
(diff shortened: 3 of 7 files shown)


gastoner (approved):
LGTM, tested on Windows. 
Just the icon seems to me quite lonely there - in future add other icons/ or try to update settingsNavComponent to chose where show icon => text is now shifted due to the icon

axel7083:
> Just the icon seems to me quite lonely there - in future add other icons/ or try to update settingsNavComponent to chose where show icon => text is now shifted due to the icon

In  the mock-up the icon is on the right, this is not something supported currently, I could change that by adding a `icon-position: 'left' | 'right' ` property 🤔 

mock-up link https://github.com/podman-desktop/podman-desktop/issues/10223#issuecomment-2522593149

slemeur:
I think we are fine getting it on the left side as well - it drives a bit more attention onto the section as well.

slemeur (approved):
Looking great to me

benoitf (changes requested):
- there should not be any hardcoded stuff
- some changes are refactoring and should not be part of this PR

some ids are changed from 'preferences.experimental.tasks' to 'experimental.tasks'
it looks like a breaking change.

why do we also need to change preferences.experimental to experimental.

and IMHO we should have to change ids at all



On the mockup the icon was on the right. Here it looks very odd to have only one item with an icon (so I would even moved out the icon from the first PR)

1. refactoring
2. experimental section added
3. icon for experimental

axel7083:
> * there should not be any hardcoded stuff
> * some changes are refactoring and should not be part of this PR

The way the navigation are build is _complicated_ we cannot create an experimental section, **it is generated by using `parentId`**.

Here is what the frontend has access

https://github.com/podman-desktop/podman-desktop/blob/0698e4d2fd1f3fcb64bfc8d66a96003545b6ee89/packages/renderer/src/stores/configurationProperties.ts#L34

We have a store of `Array<IConfigurationPropertyRecordedSchema>`, let's look at what it is 

https://github.com/podman-desktop/podman-desktop/blob/e9eb09390bed6c7f96fef2eebfdf66a416839b62/packages/main/src/plugin/configuration-registry.ts#L51-L55

Each configuration property has a parentId, which is a string, and only exist has a string. Here is the code transforming the `Array<IConfigurationPropertyRecordedSchema>` into a `Map<string, IConfigurationPropertyRecordedSchema>`.

https://github.com/podman-desktop/podman-desktop/blob/138ed5fd30fb1ff38b079d049183dd1ff44c9940/packages/renderer/src/PreferencesNavigation.svelte#L53-L68

When we look at the navigation created here is the _interesting_ part

https://github.com/podman-desktop/podman-desktop/blob/138ed5fd30fb1ff38b079d049183dd1ff44c9940/packages/renderer/src/PreferencesNavigation.svelte#L92-L94

Today, (not in this PR) there is a single `parentId`, which is `preferences`, so we only see the preferences navigation.

> some ids are changed from 'preferences.experimental.tasks' to 'experimental.tasks'

The parentId yes, to create a new section, not to change the property id, the parent id are not used in the `settings.json`, 

This can be visible by enabling the docker compatiblity, on main or on this branch we still get the same change

```json
{
  "dockerCompatibility.enabled": true
}
```

We do not use the parentId, it seems it is only used to group together the settings under preferences.

> and IMHO we should have to change ids at all

It depend on what we want to do here, changing the **parentId** allow to use the existing code to create a new section.

If we want to keep the id, we need to add a new property `experimental?: boolean`, create a new section, change the implementation of `PreferencesNavigation.svelte` to filter out the properties with this flag, add some kind of new section, of find a way to provide from the backend to the frontend the info about the parent node, because currently we only _guess_ the parentId, by parsing it and splitting using `.`.

> On the mockup the icon was on the right. Here it looks very odd to have only one item with an icon (so I would even moved out the icon from the first PR)

I asked here https://github.com/podman-desktop/podman-desktop/pull/10371#issuecomment-2545709783 and got a response here https://github.com/podman-desktop/podman-desktop/pull/10371#issuecomment-2545805907

I don't mind doing the change, I don't have any opinion on that.

To comment the steps

> 1. refactoring

Depends on how we want to achieve

> 2. experimental section added

This is not something doable with how the `preferences` section is added currently

> 3. icon for experimental

If we want to define the icon from  the backend side, what do we use ? icon.woff2 ? 

_ℹ️ side note, personal opinion, the configuration system and navigation settings are extremely complicated, and this is frustrating to work with. For example the Kubernetes navigation use a registry and a custom navigation system (ref https://github.com/podman-desktop/podman-desktop/pull/8833) where the settings don't, there is a lot of duplicate code due to this change, and this makes very long to understand how the code works together. If we want to migrate the settings to work the same way the Kubernetes navigation work this issue should be moved away at least to the next sprint because this cannot be done all together in a single sprint._

benoitf:
hello,

1. regarding the experimental setting, it needs to be a mode (like hidden) so yes we need to bring a new attribute to declare a configuration setting being experimental.

Because we want to able to filter the settings, not touching any ids and removing the flag is just removing the attribute. 

2. Once we added the attribute to all the settings that we want, (in the future it means that we would avoid to have to name it with experimental in the setting's name) then it's how to have a category to display/filter them. Also AFAIK we still want to be able to see the settings in their own categories.

3. adding the icon is for me an incremental step (it's the friction point in the discussion so I would move it out of the initial enhancement)

axel7083:
supersede by https://github.com/podman-desktop/podman-desktop/issues/10226

PR #10374, opened 2024-12-16
feat: added optional properties to close button

gastoner:
### What does this PR do?
Adds properties to close button to enable setting color of background and text

Related to #10322

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/color-registry.ts
+++ b/packages/main/src/plugin/color-registry.ts
@@ -1000,6 +1000,10 @@ export class ColorRegistry {
       dark: colorPalette.white,
       light: colorPalette.black,
     });
+    this.registerColor(`${button}close-hover-bg`, {
+      dark: colorPalette.white + '2',
+      light: colorPalette.black + '2',
+    });
     this.registerColor(`${button}link-text`, {
       dark: colorPalette.purple[400],
       light: colorPalette.purple[700],
--- a/packages/ui/src/lib/button/CloseButton.svelte
+++ b/packages/ui/src/lib/button/CloseButton.svelte
@@ -12,7 +12,7 @@ function click(): void {
 
 <button
   type="button"
-  class="text-gray-800 hover:bg-white hover:bg-opacity-10 transition-all rounded-[4px] p-1 no-underline cursor-pointer {$$props.class ||
+  class="hover:bg-[var(--pd-button-close-hover-bg)] hover:bg-opacity-10 transition-all rounded-[4px] p-1 no-underline cursor-pointer {$$props.class ||
     ''}"
   on:click={click}
   title="Close"
--- a/packages/ui/src/lib/button/CloseButton.spec.ts
+++ b/packages/ui/src/lib/button/CloseButton.spec.ts
@@ -29,8 +29,7 @@ test('Check button styling', async () => {
   // check for one element of the styling
   const button = screen.getByRole('button');
   expect(button).toBeInTheDocument();
-  expect(button).toHaveClass('text-gray-800');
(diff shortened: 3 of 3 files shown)


gastoner:
cc @axel7083 
I wanted to migrate it to svelte5 but the `props.class` property is problematic

packages/ui/src/lib/button/CloseButton.svelte
```diff
@@ -8,11 +8,14 @@ const dispatch = createEventDispatcher<{ click: undefined }>();
 function click(): void {
   dispatch('click');
 }
+
+export let textColor = 'text-gray-800';
+export let hoverBgColor = 'bg-white';
```
axel7083: See https://github.com/podman-desktop/podman-desktop/issues/9821#issuecomment-2471192647 `Advanced HTML element genericity` maybe it will help

benoitf (changes requested):
issue: this approach is not working for tailwindcss as it won't interpret the value correctly (let say you provide a value not used somewhere, it will have no effect)

axel7083:
What I would recommend, is to remove those variables

https://github.com/podman-desktop/podman-desktop/blob/516922e2ef043c208d55e9921edda79f7798aa20/packages/ui/src/lib/button/CloseButton.svelte#L15

text-gray-800 hover:bg-white are hardcoded, and we should use something `--pd-??` 

Then, I would argue, if we need to specific a custom value we could do something like

`class="text-[color:var(--color, --pd-close-??)]"`

deboer-tim:
This doesn't sound like the right approach to me - e.g. we wouldn't expose the purple color of buttons so that they can be red in other contexts. Some suggestions:
- The X could default to the current/parent text color (would need to check other usage), or have a mode (something like default vs card) instead of exporting the color directly.
- Hover could use transparency (similar to link) or other methods so that it works on more backgrounds.

gastoner:
I would prefer just removing the text-gray-800 to make it reflect the parent color 
Add new color `--pd-button-cancel-hover-bg` which will be same as the Links hover:
image
image
cc @deboer-tim @axel7083 @benoitf

axel7083:
> Add new color `--pd-button-cancel-hover-bg` which will be same as the Links hover:

The button is named `close` so might be better to have something like `--pd-button-close-hover-bg` ?

gastoner:
> > Add new color `--pd-button-cancel-hover-bg` which will be same as the Links hover:
> 
> The button is named `close` so might be better to have something like `--pd-button-close-hover-bg` ?

Yes ofc, "the other c-word" :D close is correct and not cancel whoops

benoitf (commented):
it's missing unit test

benoitf (approved):

SoniaSandler (approved):

packages/main/src/plugin/color-registry.ts
```diff
@@ -1000,6 +1000,10 @@ export class ColorRegistry {
       dark: colorPalette.white,
       light: colorPalette.black,
     });
+    this.registerColor(`${button}close-hover-bg`, {
+      dark: colorPalette.white + '2',
+      light: colorPalette.black + '2',
```
axel7083, line 1005: +2 ?

deboer-tim: Makes it translucent (3 char rgb + an 'a' value). Not great, but we already had to use the same in Link.

gastoner: It same as the hover `Link` component as proposed by Tim

axel7083 (approved):

PR #10383, opened 2024-12-16
feat: dispatch permissions to frontend in experimental contexts mode

### LINKED ISSUE #7629: Use equivalent of `kubectl auth can-i` on Kube contexts to check connectivity instead of informers

feloy:
### What does this PR do?

- Dispatch permissions to frontend in experimental contexts mode
- the latest commit refactors existing code to use `KubernetesResourceName` instead of `string` 

### Screenshot / video of UI

No UI changes in this PR

UI will come after https://github.com/podman-desktop/podman-desktop/issues/10385

Fixes #7629

### How to test this PR?


packages/api/src/kubernetes-resources.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export const kubernetesResources = [
```
benoitf: issue: looks odd to have a constant with camelCase pattern. Should be all UPPER_CASE

packages/main/src/plugin/kubernetes/contexts-states-dispatcher.ts
```diff
@@ -56,6 +57,21 @@ export class ContextsStatesDispatcher {
...
+      for (const [resourceName, contextResourcePermission] of permissions) {
+        value.push({
+          contextName,
+          resourceName,
+          permitted: contextResourcePermission.permitted,
+          reason: contextResourcePermission.reason,
+        });
+      }
+    }
+    return value;
```
benoitf: question: should it use the .map pattern rather two for loop ?

packages/main/src/plugin/kubernetes/kubernetes-client.ts
```diff
@@ -1852,4 +1853,11 @@ export class KubernetesClient {
     }
     return this.contextsStatesDispatcher?.getContextsHealths();
   }
+
+  public getContextsPermissions(): ContextPermission[] {
+    if (!this.contextsStatesDispatcher) {
+      throw new Error('contextsStatesDispatcher is undefined. This should not happen in Kubernetes experimental');
+    }
+    return this.contextsStatesDispatcher?.getContextsPermissions();
```
benoitf: TypeScript is not able to infer that it's always here ?
```suggestion
    return this.contextsStatesDispatcher.getContextsPermissions();
```

feloy:
PR set in draft, as I'm not sure if it is appropriate to have a strict type KubernetesResourceName for resources names, instead of a generic `string` type, as this could block us from managing CRDs. 

I'll make a quick PoC to check how CRDs could be handled, and if we may use or not a strict type for resources names.

=> #10418

With this PoC, we cannot have a strict type for resource names

benoitf (commented):
looks like I forgot to hit send when I read the PR...

packages/api/src/kubernetes-resources.ts
feloy: I finally removed this type

packages/api/src/kubernetes-contexts-permissions.ts
```diff
@@ -0,0 +1,32 @@
...
+  resourceName: string;
+  // permitted if allowed and not denied
+  // > When multiple authorization modules are configured, each is checked in sequence.
+  // > If any authorizer approves or denies a request, that decision is immediately returned
+  // > and no other authorizer is consulted. If all modules have no opinion on the request,
+  // > then the request is denied. An overall deny verdict means that the API server rejects
+  // > the request and responds with an HTTP 403 (Forbidden) status.
+  // (source: https://kubernetes.io/docs/reference/access-authn-authz/authorization/)
+  permitted: boolean;
+  reason?: string;
```
benoitf, line 34: nit: document this optional attribute ?

benoitf (approved):

packages/renderer/src/stores/kubernetes-context-permission-experimental.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  value: {
+    kubernetesGetContextsPermissions: vi.fn(),
+    getConfigurationValue: vi.fn(),
+    addEventListener: eventEmitter.receive,
+    events: {
+      receive: eventEmitter.receive,
+    },
+  },
+  writable: true,
+});
```
benoitf: question: should it be in the beforeAll method ?

benoitf (approved):

packages/renderer/src/stores/kubernetes-context-permission-experimental.spec.ts
feloy: yes, that seems more reasonable. Moving it

packages/renderer/src/stores/kubernetes-context-permission-experimental.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  Object.defineProperty(global, 'window', {
+    value: {
+      kubernetesGetContextsPermissions: vi.fn(),
+      getConfigurationValue: vi.fn(),
+      addEventListener: eventEmitter.receive,
+      events: {
+        receive: eventEmitter.receive,
+      },
+    },
+    writable: true,
```
jeffmaury: Is that required ?

packages/renderer/src/stores/kubernetes-context-permission-non-experimental.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  Object.defineProperty(global, 'window', {
+    value: {
+      kubernetesGetContextsPermissions: vi.fn(),
+      getConfigurationValue: vi.fn(),
+      addEventListener: eventEmitter.receive,
+      events: {
+        receive: eventEmitter.receive,
+      },
+    },
+    writable: true,
```
jeffmaury: Same here ?

packages/renderer/src/stores/kubernetes-context-permission-experimental.spec.ts
feloy: removed

jeffmaury (approved):
LGTM

PR #10414, opened 2024-12-18
feat(ui): add close button to hide banners

### LINKED ISSUE #9415: Make it possible to hide the banner promoting extensions

gastoner:
### What does this PR do?
Adds UI Close button to hide promoted extensions banners
PR 2/2 - first PR https://github.com/podman-desktop/podman-desktop/pull/10322 (Needs to be merged first)
Frontend changes

### Screenshot / video of UI

https://github.com/user-attachments/assets/4b0a3b3d-673b-40ac-97f9-dd35fa950be4

Closes #9415 

### How to test this PR?
Try to show/ hide banners on Dashboard/ in the preferences

### DIFF
--- a/packages/renderer/src/lib/recommendation/ExtensionBanner.spec.ts
+++ b/packages/renderer/src/lib/recommendation/ExtensionBanner.spec.ts
@@ -16,9 +16,9 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { render, screen } from '@testing-library/svelte';
+import { render, screen, waitFor } from '@testing-library/svelte';
 import { tick } from 'svelte';
-import { beforeAll, beforeEach, describe, expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 
 import ExtensionBanner from '/@/lib/recommendation/ExtensionBanner.svelte';
 
@@ -59,14 +59,6 @@ const imageBackground: IExtensionBanner = {
     dark: 'data:image/png;base64-image-dark',
   },
 };
-
-const showMessageBoxMock = vi.fn();
-
-beforeAll(() => {
-  Object.defineProperty(window, 'showMessageBox', {
-    value: showMessageBoxMock,
-  });
-});
 
 beforeEach(() => {
   vi.resetAllMocks();
@@ -148,7 +140,9 @@ describe('backgrounds', () => {
   });
 });
 
-test('opening messageBox', async () => {
+test('opening messageBox and hiding banner', async () => {
+  vi.mocked(window.showMessageBox).mockResolvedValue({ response: 1 });
+
   render(ExtensionBanner, {
     banner: gradientBackground,
     isDark: true,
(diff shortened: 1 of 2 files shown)


dgolovin (changes requested):
Component test is missing.

packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
```diff
@@ -22,6 +24,24 @@ $effect(() => {
     style = undefined;
   }
 });
+
+async function onClose(): Promise<void> {
+  await window.telemetryTrack('hideRecommendationExtension.popupOpened', {
```
benoitf: I think we should merge the telemetry events in a try/finally block

like: one event and what was the result rather than 2 events

It's easier to filter using one event, like give me all 'displayCloseRecommendationsPanel' and group by the status like  hide vs the don't hide vs the escape etc

gastoner: Is sending just `result.response` ok? it can be 0,1 (clicking on hide/dont hide) or undefined (by clinking on x or outside the dialog). Or should I rather send some text like hideBanner and keepBanner?

benoitf: it might be safer to send data that is not a direct value for later iterations

while it works with result.response, let say we change some order, move buttons, I don't know change the dialog etc, it'll hard to monitor on long term the data

so I would vote for a specific text

also from analysis the data, how do you know easily what to filter if you have 0 and 1 is 0 closed or 1 closed

```
choice = 'hide' | 'keep'
await window.telemetryTrack('recommendations.banner.dismissDialog', {choice})
```

```
  result = await window.showMessageBox({
      title: 'Hide extension recommendation banners',
      message: `Do you want to hide extension recommendation banners?`,
      type: 'warning',
      buttons: [`Don't hide`, 'Hide'],
    });
```

Maybe say 'No, keep them' rather than `Don't hide` (no strong opinion there and it can always be changed later)
and 'Yes, hide' than 'Hide'

benoitf (approved):

benoitf (commented):
the telemetry part should be in the finally block, not in the try block

and there is no need to have the catch clause

benoitf (commented):
and in the test, it's missing the check to see the value sent to the telemetry

and it should check what happens if user click on yes and if user click on no

packages/renderer/src/lib/recommendation/ExtensionBanner.spec.ts
```diff
@@ -60,8 +60,24 @@ const imageBackground: IExtensionBanner = {
...
+    value: telemetryTrackMock,
+  });
+  Object.defineProperty(window, 'updateConfigurationValue', {
+    value: vi.fn(),
+  });
+});
+
 beforeEach(() => {
   vi.resetAllMocks();
+  telemetryTrackMock.mockImplementation(() => Promise.resolve());
```
benoitf: if you look at recommended pattern, it's always vi.mocked(window.telemetryTrack)

that said I don't know why we would need an empty implementation ?

packages/renderer/src/lib/recommendation/ExtensionBanner.spec.ts
```diff
@@ -60,8 +60,24 @@ const imageBackground: IExtensionBanner = {
   },
 };
 
+const showMessageBoxMock = vi.fn();
+const telemetryTrackMock = vi.fn();
```
benoitf:
```suggestion
```

just use vi.fn() inside the beforeAll
that said I think that with my changes this is now useless

packages/renderer/src/lib/recommendation/ExtensionBanner.spec.ts
```diff
@@ -60,8 +60,24 @@ const imageBackground: IExtensionBanner = {
...
+  Object.defineProperty(window, 'showMessageBox', {
+    value: showMessageBoxMock,
+  });
+  Object.defineProperty(window, 'telemetryTrack', {
+    value: telemetryTrackMock,
+  });
+  Object.defineProperty(window, 'updateConfigurationValue', {
+    value: vi.fn(),
+  });
+});
```
benoitf: following https://github.com/podman-desktop/podman-desktop/pull/10426 you can just forget these defitions

gastoner:
@benoitf those current changes to tests are amazing!

benoitf (approved):
yes I don't know why we didn't do it earlier ! ;-)

packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
```diff
@@ -22,6 +25,25 @@ $effect(() => {
...
 });
+
+async function onClose(): Promise<void> {
+  let result: MessageBoxReturnValue = { response: -1 };
+  try {
+    result = await window.showMessageBox({
+      title: 'Hide extension recommendation banners',
+      message: `Do you want to hide extension recommendation banners?`,
+      type: 'warning',
+      buttons: [`No, keep them`, 'Yes, hide'],
```
axel7083, line 36: image

Should we invert the buttons ? Look odds to me

axel7083 (approved):
LGTM

packages/renderer/src/lib/recommendation/ExtensionBanner.svelte
gastoner: @axel7083 I was thinking that ideally we want to keep them. But I can switch those

gastoner:
@dgolovin can you check again please?

dgolovin (approved):

PR #10415, opened 2024-12-18
fix: kube health store is inactive in non experimental mode

### LINKED ISSUE #9938: Use health check to check Kubernetes cluster connectivity

feloy:

### What does this PR do?

The PR #10349 introduced a store to be used in Experimental states mode only, but the store was fetching the health values in all cases (experimental mode or not).

Because of this, a log 'Failed to update contexts healths Error: contextsStatesDispatcher is undefined. This should not happen in Kubernetes experimental' was displayed in the console.

This PR makes the Kubernetes health store inactive in non experimental mode, and the log is not displayed anymore.

Fixes #9938 

### How to test this PR?

Define or not the experimental mode, and check that the health is coorectly displayed in experimental mode, and no logs are displayed when the experimental mode is not set:

````
"kubernetes.statesExperimental": true,
````

### DIFF
--- a/packages/renderer/src/stores/kubernetes-context-health-non-experimental.spec.ts
+++ b/packages/renderer/src/stores/kubernetes-context-health-non-experimental.spec.ts
@@ -17,7 +17,7 @@
  ***********************************************************************/
 
 import { get } from 'svelte/store';
-import { expect, test, vi } from 'vitest';
+import { beforeAll, expect, test, vi } from 'vitest';
 
 import { kubernetesContextsHealths, kubernetesContextsHealthsStore } from './kubernetes-context-health';
 
@@ -32,15 +32,17 @@ const eventEmitter = {
   },
 };
 
-Object.defineProperty(global, 'window', {
-  value: {
-    kubernetesGetContextsHealths: vi.fn(),
-    getConfigurationValue: vi.fn(),
-    addEventListener: eventEmitter.receive,
-    events: {
-      receive: eventEmitter.receive,
+beforeAll(() => {
+  Object.defineProperty(global, 'window', {
+    value: {
+      kubernetesGetContextsHealths: vi.fn(),
+      getConfigurationValue: vi.fn(),
+      addEventListener: eventEmitter.receive,
+      events: {
+        receive: eventEmitter.receive,
+      },
     },
-  },
+  });
 });
 
 test('kubernetesContextsHealths not in experimental states mode', async () => {
--- a/packages/renderer/src/stores/kubernetes-context-health-non-experimental.spec.ts
+++ b/packages/renderer/src/stores/kubernetes-context-health-non-experimental.spec.ts
@@ -1,0 +1,76 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { get } from 'svelte/store';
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/renderer/src/stores/kubernetes-context-health.ts [hunk 1]
```diff
@@ -25,8 +25,18 @@ import { EventStore } from './event-store';
...
+
 export async function checkForUpdate(eventName: string): Promise<boolean> {
+  // check for update only in experimental states mode
+  if (experimentalStates === undefined) {
+    experimentalStates = (await window.getConfigurationValue<boolean>('kubernetes.statesExperimental')) ?? false;
+  }
+  if (experimentalStates === false) {
+    return false;
+  }
+
```
benoitf, line 39: question: hello, is it ok to check only once the configuration ? if for example I'm turning on the experimental flag later it won't be handled ?

should it monitor the value of this property ?

feloy: The live change is not supported elsewhere in the code, the user needs to restart the app (as the config is hidden, it does not seem necessary to me, and it would complexify the `kubernetes-client` logic).

packages/renderer/src/stores/kubernetes-context-health-non-experimental.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // send 'extensions-already-started' event
+  const callbackExtensionsStarted = callbacks.get('extensions-already-started');
+  expect(callbackExtensionsStarted).toBeDefined();
+  await callbackExtensionsStarted();
+
+  // values are never fetched
+  await new Promise(resolve => setTimeout(resolve, 500));
+  const currentValue = get(kubernetesContextsHealths);
+  expect(currentValue).toEqual([]);
```
benoitf, line 75: nit: should we use a waitFor get equal [] ?

benoitf (approved):

packages/renderer/src/stores/kubernetes-context-health-non-experimental.spec.ts
feloy: I don't think, as we want to test that the value *remains* `[]` for all this time. By using `waitFor`, it would return true immediately as the initial value is `[]`, but it would not check if the value changes after a few time.

benoitf: ok I read too fast, I thought initial values were not empty and we were waiting that it's returning to be empty.

should it check that we never call the method to fetch remote context then ?

feloy: Yes, this check is done the line below

SoniaSandler (approved):
LGTM codewise

packages/renderer/src/stores/kubernetes-context-health-non-experimental.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import { kubernetesContextsHealths, kubernetesContextsHealthsStore } from './kubernetes-context-health';
+
+// We need to have separate tests files to run different tests, as there are global variables in the store file, which cannot be reset between tests
+
+// This file can be removed as soon as the experimental states mode is adopted as non experimental
+
+const callbacks = new Map<string, any>();
+const eventEmitter = {
+  receive: (message: string, callback: any) => {
```
jeffmaury: I think the usage of any can be prevented

packages/renderer/src/stores/kubernetes-context-health-non-experimental.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+Object.defineProperty(global, 'window', {
+  value: {
+    kubernetesGetContextsHealths: vi.fn(),
+    getConfigurationValue: vi.fn(),
+    addEventListener: eventEmitter.receive,
+    events: {
+      receive: eventEmitter.receive,
+    },
+  },
+  writable: true,
```
jeffmaury: Is this required ?

jeffmaury (approved):
LGTM

Issue #10405, opened 2024-12-17
flaky test in `ContainerDetailsTerminal.spec.ts:236:55` on macos


axel7083:
### Bug description

I see often the [unit tests / macos-15](https://github.com/podman-desktop/podman-desktop/actions/runs/12375466605/job/34540333141?pr=10403#logs) being flaky, the problematic test seems to be the following

https://github.com/podman-desktop/podman-desktop/blob/455637a329078d773244291d944aa9d566174e2e/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts#L193

### Operating system

MacOS

### Installation Method

Other

### Version

next (development version)

### Relevant log output

```shell
Error: Error: expect(element).toHaveTextContent()

Expected element to have text content:
  hello world
Received:


Ignored nodes: comments, script, style
<html>
  <head />
  <body>
    <div>
      <div
        class="h-full p-[5px] pr-0 bg-[var(--pd-terminal-background)]"
      >
        <div
          class="terminal xterm xterm-dom-renderer-owner-4"
          dir="ltr"
        >
          <div
            class="xterm-accessibility"
          >
            <div
              class="xterm-accessibility-tree"
              role="list"
            >
              <div
                aria-posinset="1"
                aria-setsize="24"
                role="listitem"
                tabindex="-1"
              />
              <div
                aria-posinset="2"
                aria-setsize="24"
                role="listitem"
                tabindex="-1"
              />
```

axel7083:
Re-opening as I see more and more tests failing

axel7083:
```
src/lib/container/ContainerDetailsTerminal.spec.ts > expect being able to reconnect

Error: expect(element).toHaveTextContent()

Expected element to have text content:
  hello world
Received:

 ❯ src/lib/container/ContainerDetailsTerminal.spec.ts:126:36
```

feloy:
I suspect that the initial intent was to mock the `xterm` library during the test, but the path of the library has changed from `xterm` to `@xterm/xterm` and the import is still mocking `xterm`

axel7083:
I unassigned myself, if you want to give it a try !

But the fact this is only flaky on macos, is weird

dgolovin:
I can replicate that removing `retry:3` from svelte configuration. They appear not only in ContainerDetailsTerminal.spec.ts

PR #10419, opened 2024-12-18
refactor: container details terminal tests

axel7083:
### What does this PR do?

While working on https://github.com/podman-desktop/podman-desktop/issues/10405 I updated the `ContainerDetailsTerminal.spec.ts` to try to find out the origin of the bug.

Here are a list of changes

1. Uses the vi.mocked(windows.<function>) instead of having global mock 
2. Simplify how we check for content in the Terminal
Before this PR, we were querying some specific div with the following code

```ts
  // wait 1s
  await waitFor(() => renderObject.container.querySelector('div[aria-live="assertive"]'));

  // search a div having aria-live="assertive" attribute
  const terminalLinesLiveRegion = renderObject.container.querySelector('div[aria-live="assertive"]');

  // check the content
  await waitFor(() => expect(terminalLinesLiveRegion).toHaveTextContent('hello world'));
```

we do not need to do that as we can use the renderObject and use the `renderObject.getByText` function to do the same.

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
@@ -18,39 +18,49 @@
 
 import '@testing-library/jest-dom/vitest';
 
-import { render, screen, waitFor } from '@testing-library/svelte';
+import { render, waitFor } from '@testing-library/svelte';
 import { tick } from 'svelte';
 import { get } from 'svelte/store';
-import { beforeEach, expect, test, vi } from 'vitest';
+import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
 import { containerTerminals } from '/@/stores/container-terminal-store';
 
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
```diff
@@ -18,39 +18,49 @@
...
 import type { ContainerInfoUI } from './ContainerInfoUI';
 
-const getConfigurationValueMock = vi.fn();
-const shellInContainerMock = vi.fn();
-const shellInContainerResizeMock = vi.fn();
 vi.mock('xterm', () => {
   return {
-    Terminal: vi
-      .fn()
-      .mockReturnValue({ loadAddon: vi.fn(), open: vi.fn(), write: vi.fn(), dispose: vi.fn(), onData: vi.fn() }),
```
axel7083, line 38: no needed because we were doing `vi.resetAllMocks();` so this code was not used

axel7083:
Merge conflict too annoying

PR #10424, opened 2024-12-18
chore: handle more cases where methods can be mocked

benoitf:
### What does this PR do?
handle more cases where methods can be mocked
(unit test case only)

in some cases, the method is there but not mocking a promise, so handle this case

### How to test this PR?


gastoner (approved):
LGTM

PR #10426, opened 2024-12-18
chore: generate all the mocks of exposed methods during the setup of vitest

### LINKED ISSUE #10420: Generate all the mocks automatically from the declaration file in the tests

today, when, we write tests, each time we need to use a method from the window object, we're adding a mock for it in the beforeAll

these mocks should be generated for us automatically

generate the mock automatically in the setup of the vitest environment

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
avoid to have to think about it in all unit tests. Still need to provide the returned value if needed but no need to initialize all vi.fn()

fixes https://github.com/podman-desktop/podman-desktop/issues/10420

### How to test this PR?

### DIFF
--- a/packages/renderer/vite.tests.setup.js
+++ b/packages/renderer/vite.tests.setup.js
@@ -16,6 +16,61 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import path from 'node:path';
+import { readFileSync } from 'node:fs';
 import 'vitest-canvas-mock';
+import typescript from 'typescript';
+import { expect } from 'vitest';
 
 global.window.matchMedia = () => {};
+
+// read the given path and extract the method names from the Window interface
+function extractWindowMethods(filePath) {
+  // Read the content of the .d.ts file
+  const fileContent = readFileSync(filePath, 'utf-8');
+
+  // Create a TypeScript SourceFile
+  const sourceFile = typescript.createSourceFile(filePath, fileContent, typescript.ScriptTarget.Latest, true);
+
+  const methodNames = [];
+
+  // Visit each node in the AST
+  const visit = node => {
+    // Look for the Window interface
+    if (
+      typescript.isInterfaceDeclaration(node) &&
+      node.name.text === 'Window' // Target the "Window" interface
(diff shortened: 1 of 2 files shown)


dgolovin (commented):
Looks like it needs "build:preload:types" to be executed before tests

```
"test:renderer": "npm run build:preload:types && npm run build:ui && vitest -c packages/renderer/vite.config.js run packages/renderer --passWithNoTests --coverage"
```

benoitf:
@dgolovin yes I was wondering if it could be smarter ( like to not do it everytime but it looks like the simplest option)

packages/renderer/vite.tests.setup.js
```diff
@@ -16,6 +16,61 @@
...
+          const name = member.name.text;
+          methodNames.push(name);
+        }
+      }
+    }
+
+    typescript.forEachChild(node, visit);
+  };
+
+  visit(sourceFile);
```
dgolovin, line 55: nit: the sake of using the same visitor pattern I would call the visitor first time using
```suggestion
  typescript.forEachChild(sourceFile, visit);
```

dgolovin (approved):

feloy (approved):

axel7083 (approved):
LGTM

packages/renderer/vite.tests.setup.js
benoitf: I think it might be less readable, 
- usually we are calling the visitor on the tree (first node) and it's going recursive from there
- but with your suggestion, we iterate on all the children first and do the same thing than the last part of the visitor, so I'm not even sure we're optimizing a lot.

What other people thoughts here ? I will do whatever is the most upvoted

cc @feloy @axel7083 the other reviewers

feloy: I'm fine with the current implementation, with the first call using the `visit` function, and the call to `typescript.forEachChild` hidden into the implementation (the recursivity being an implementation detail)

axel7083: I am okey with both

PR #10436, opened 2024-12-19
chore: update @kubernetes/client-node to v1.0.0

feloy:
### What does this PR do?

Update @kubernetes/client-node to v1.0.0 and removes previous patches.

Do not mark the library as external anymore as the format has been chaged to ESM: https://github.com/kubernetes-client/javascript/pull/2062

### How to test this PR?


package.json
```diff
@@ -197,6 +197,7 @@
     "hpagent": "^1.2.0",
     "minimist": "^1.2.8",
     "moment": "^2.30.1",
+    "node-fetch": "^3.3.2",
```
benoitf: issue: fetch is now part of nodejs it should not be added with a 3rd party library

feloy: I added it as I got an error when commit'ing the changes. I'll check again

feloy: I finally removed the use of the `AbortError` type, to use the same check as in the library:

https://github.com/kubernetes-client/javascript/blob/release-1.x/src/health.ts#L52

```
if (err instanceof Error && err.name === 'AbortError') {
```

packages/main/src/plugin/kubernetes/context-health-checker.ts
```diff
@@ -111,4 +111,8 @@ export class ContextHealthChecker implements Disposable {
   public getState(): ContextHealthState {
     return this.#currentState;
   }
+
+  private isAbortError(err: unknown): boolean {
```
axel7083, line 115:
```suggestion
  private isAbortError(err: unknown): err is AbortError {
```

feloy: But if we do this, we need again the AbortError type. I wanted to get rid of it

axel7083: :/ okey

axel7083 (approved):
LGTM

packages/main/vite.config.js
```diff
@@ -51,7 +51,6 @@ const config = {
       external: [
         'electron',
         'chokidar',
-        '@kubernetes/client-node',
```
benoitf: highly recommended: I would remove this part from this pull request as it's not directly related to 1.0.0-RC7 to 1.0.0

benoitf (approved):

packages/main/vite.config.js
feloy: I think it is related, as the library has been changed to ESM between the 2 versions.

benoitf: but doing it there and now can have side consequences in production mode.

so it's always preferable to not mix two things at a single time

feloy: OK, I've created https://github.com/podman-desktop/podman-desktop/pull/10439

jeffmaury: So this change should be removed ?

feloy: yes, it will be removed after rebase

PR #10437, opened 2024-12-19
feat(backend): adding experimental property to configuration schema

axel7083:
### What does this PR do?

First step of the experimental list, adding the experimental property on the configuration schema.

> :question: I am wondering I should be moving the `IConfigurationPropertySchema`, `IConfigurationPropertyRecordedSchema`, `IConfigurationChangeEvent` and `IConfigurationPropertySchemaType` to the `packages/api` :thinking: 

Related to https://github.com/podman-desktop/podman-desktop/issues/10226

### How to test this PR?

### DIFF
--- a/packages/api/src/configuration/models.ts
+++ b/packages/api/src/configuration/models.ts
@@ -1,0 +1,22 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


packages/api/src/configuration/models.ts
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface IExperimentalConfiguration {
+  // href to the discussion
+  discussion: string;
```
benoitf:
```suggestion
  githubDiscussionLink?: string;
```

question: I'm concerned about it being mandatory first. Maybe start as optional and over time we could enforce it not being optional

suggest: possibly a field `enabled` to have all the configuration ready but not enabled

benoitf (commented):
just a reminder about your suggestion that if you move files, you need to move files before or after changing some properties in it

benoitf (approved):

PR #10446, opened 2024-12-19
chore: migrate ProviderResultPage (and ImageDetailsCheck) components to svelte 5

benoitf:
### What does this PR do?

migrate component to svelte 5

- change header-info to headerInfo (as it's not possible to have header-info in a Snippet/slot....)
- move the content of the method to the $derived methods so it can check the values being used

### How to test this PR?

install for example openshift checker extension

### DIFF
--- a/packages/renderer/src/lib/image/ImageDetailsCheck.svelte
+++ b/packages/renderer/src/lib/image/ImageDetailsCheck.svelte
@@ -15,14 +15,18 @@ import ProviderResultPage from '../ui/ProviderResultPage.svelte';
 const orderStatus = ['failed', 'success'];
 const orderSeverity = ['critical', 'high', 'medium', 'low', undefined];
 
-export let imageInfo: ImageInfo | undefined;
+interface Props {
+  imageInfo?: ImageInfo;
+}
 
-let providers: ProviderUI[];
-let results: CheckUI[] = [];
-let cancellableTokenId: number = 0;
+const { imageInfo }: Props = $props();
 
-let remainingProviders: number;
-let aborted = false;
+let providers: ProviderUI[] = $state([]);
+let results: CheckUI[] = $state([]);
+let cancellableTokenId: number = $state(0);
+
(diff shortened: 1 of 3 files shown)


feloy:
I can see that the provider's `SlideToggle` on `ProviderResultPage` is not updated correctly anymore:

before:
convert-before


after:
convert-after

benoitf:
thanks @feloy for the catch. I upgraded the ImageCheck to runes mode as well. It looks like the update on providers array was not reflected properly in the ProviderPage component

feloy (approved):
LGTM, works correctly

packages/renderer/src/lib/image/ImageDetailsCheck.svelte
```diff
@@ -127,7 +131,9 @@ async function handleAbort() {
 </script>
 
 <ProviderResultPage providers={providers} results={results}>
-  <div class="flex flex-row" slot="header-info">
+  <!-- eslint-disable-next-line sonarjs/no-unused-vars -->
```
axel7083, line 134: what is considered unused vars ? snippet ?

axel7083 (approved):
LGTM

packages/renderer/src/lib/image/ImageDetailsCheck.svelte
benoitf: yes it considers that the function headerInfo() `{#snippet headerInfo()}` is unused

benoitf: I don't really know how to proceed. I saw that some other IDEs have bugs opened around the Snippet / unused variables

axel7083: maybe we can open an issue on sonarjs / eslint side ?

axel7083: I am okay with merging it, I understand the issue

Issue #5609, opened 2024-01-19
Show the Lima provider with a lime green color, instead of the current boring gray color


afbjorklund:
### Is your enhancement related to a problem? Please describe

Currently it is shown in grey (`bg-gray-900`).

* `#111827`

### Describe the solution you'd like

It would be nicer to show in e.g. lime green

* `#32cd32`

### Describe alternatives you've considered

It could also be colored by the connection type, instead of by the provider type?

So that a docker Lima would colored in blue, and a podman Lima would be purple.

### Additional context

```typescript
// Each provider has a colour associated to it within tailwind, this is a map of those colours.
// bg-purple-600 = podman
// bg-sky-300 = docker
// bg-sky-600 = kubernetes
// bg-gray-900 = unknown
function getProviderColour(providerName: string): string {
  switch (providerName?.toLowerCase()) {
    case ContainerGroupInfoTypeUI.PODMAN:
      return 'bg-purple-600';
    case ContainerGroupInfoTypeUI.DOCKER:
      return 'bg-sky-400';
    case PodGroupInfoTypeUI.KUBERNETES:
      return 'bg-sky-600';
    default:
      return 'bg-gray-900';
  }
}
```

----

![tailwind-malachite](https://github.com/containers/podman-desktop/assets/10364051/8ca3b946-2b8f-451a-975a-3d2b42e71b40)

```ts
'malachite': {
    '50': '#f2fcf1',
    '100': '#defade',
    '200': '#c0f3bf',
    '300': '#8de88d',
    '400': '#54d454',
    '500': '#32cd32',
    '600': '#209920',
    '700': '#1c791d',
    '800': '#1b601c',
    '900': '#184f1a',
    '950': '#082b09',
},
```

Side note: lima switched from "lime" to "lime green", due to some usabiity issues when running in light mode...

![image](https://github.com/containers/podman-desktop/assets/10364051/88d4b7d9-edd1-4d2d-90f7-720c1d3c5ea2)

deboer-tim:
Maybe provider could (optionally) provide a color? I would prefer a simple mechanism to more hardcoding.

afbjorklund:
> It could also be colored by the connection type, instead of by the provider type?

Actually it was the **images** that are colored by provider, the containers are colored by the connection

container environment: "Podman"

![podman-desktop-container-env](https://github.com/containers/podman-desktop/assets/10364051/0179016e-91b0-4aa3-93df-fb7a4bb15e20)

[...]

PR #10447, opened 2024-12-19
feat: add server type for Lima instances

afbjorklund:
### What does this PR do?

Adds detection of "server type" for lima docker socket

Previously it was listed as "Server: unknown", if linked

### Screenshot / video of UI

pd-lima
pd-light

https://github.com/podman-desktop/podman-desktop/issues/9351#issuecomment-2488721180

### How to test this PR?

You need to create the `/var/run/docker.sock`  symlink.

---

The "provider info" is **only** used for the server type, the connection type is still the actual engine type...

So the server type is "lima", but the connection type is "docker" or "podman" or "kubernetes" (containerd)

It is shown in the right color for the containers, but it is still broken for the images (showing as "lima")

* https://github.com/podman-desktop/podman-desktop/issues/5609#issuecomment-1900956942

pd-container

pd-image

They are both supposed to show the engine type (connection type), and not the provider type

So the wrong type is being used, but since it is all string and they're all the same for the others...

### DIFF
--- a/packages/main/src/plugin/docker/docker-compatibility.ts
+++ b/packages/main/src/plugin/docker/docker-compatibility.ts
@@ -63,12 +63,28 @@ export class DockerCompatibility {
     this.#configurationRegistry.registerConfigurations([dockerCompatibilityConfiguration]);
   }
 
+  protected getNameFromServerInfo(info: { Name?: string; OperatingSystem?: string }, podmanInfo?: unknown): string {
+    if (info.OperatingSystem === 'Docker Desktop') {
+      return 'docker';
+    } else if (info.Name?.startsWith('lima')) {
+      // lima instance host names all start with "lima"
+      return 'lima';
+    } else if (info.OperatingSystem === 'podman' || podmanInfo) {
+      // if podman info is available, then it is podman
+      return 'podman';
+    }
+    return 'unknown';
+  }
+
   protected getTypeFromServerInfo(
-    info: { OperatingSystem?: string },
+    info: { Name?: string; OperatingSystem?: string },
     podmanInfo?: unknown,
   ): DockerSocketServerInfoType {
     if (info.OperatingSystem === 'Docker Desktop') {
       return 'docker';
+    } else if (info.Name?.startsWith('lima')) {
+      // lima instance host names all start with "lima"
+      return podmanInfo ? 'podman' : 'docker';
     } else if (info.OperatingSystem === 'podman' || podmanInfo) {
       // if podman info is available, then it is podman
       return 'podman';
@@ -122,6 +138,7 @@ export class DockerCompatibility {
       const status: DockerSocketMappingStatusInfo = {
         status: 'running',
         serverInfo: {
+          name: this.getNameFromServerInfo(compatInfo, podmanInfo),
           type: this.getTypeFromServerInfo(compatInfo, podmanInfo),
           serverVersion,
           operatingSystem: compatInfo.OperatingSystem,
--- a/packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilitySocketMappingStatus.svelte
+++ b/packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilitySocketMappingStatus.svelte
@@ -53,7 +53,7 @@ onMount(async () => {
           <RefreshButton label="Refresh the status" onclick={refreshSocketMappingStatus} />
         </div>
         {#if dockerSocketMappingStatusInfo?.status === 'running' && dockerSocketMappingStatusInfo?.serverInfo}
-          <Label name="{dockerSocketMappingStatusInfo.serverInfo.type} is listening">
+          <Label name="{dockerSocketMappingStatusInfo.serverInfo.name} is listening">
             <ProviderInfoCircle type={engineType} />
           </Label>
         {:else if dockerSocketMappingStatusInfo?.status === 'unreachable'}
@@ -100,7 +100,7 @@ onMount(async () => {
       aria-label="Server information">
       <div class="grid grid-cols-2 gap-x-8 gap-y-2">
         <div>Server:</div>
-        <div>{dockerSocketMappingStatusInfo.serverInfo.type}</div>
+        <div>{dockerSocketMappingStatusInfo.serverInfo.name}</div>
 
         <div>Version:</div>
         <div>{dockerSocketMappingStatusInfo.serverInfo.serverVersion}</div>
--- a/packages/api/src/docker-compatibility-info.ts
+++ b/packages/api/src/docker-compatibility-info.ts
@@ -35,6 +35,7 @@ export interface DockerSocketMappingStatusInfo {
     displayName: string;
   };
   serverInfo?: {
+    name: string;
     type: DockerSocketServerInfoType;
     serverVersion: string;
     operatingSystem: string;
--- a/packages/main/src/plugin/docker/docker-compatibility.spec.ts
+++ b/packages/main/src/plugin/docker/docker-compatibility.spec.ts
@@ -43,8 +43,14 @@ const providerRegistry: ProviderRegistry = {
 } as unknown as ProviderRegistry;
 
 export class TestDockerCompatibility extends DockerCompatibility {
+  public override getNameFromServerInfo(
+    info: { Name?: string; OperatingSystem?: string },
+    podmanInfo?: unknown,
+  ): string {
+    return super.getNameFromServerInfo(info, podmanInfo);
+  }
   public override getTypeFromServerInfo(
-    info: { OperatingSystem?: string },
+    info: { Name?: string; OperatingSystem?: string },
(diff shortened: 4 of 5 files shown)


packages/main/src/plugin/docker/docker-compatibility.ts
```diff
@@ -69,14 +69,17 @@ export class DockerCompatibility {
...
-    info: { OperatingSystem?: string },
+    info: { Name?: string; OperatingSystem?: string },
     podmanInfo?: unknown,
   ): DockerSocketServerInfoType {
     if (info.OperatingSystem === 'Docker Desktop') {
       return 'docker';
     } else if (info.OperatingSystem === 'podman' || podmanInfo) {
       // if podman info is available, then it is podman
       return 'podman';
+    } else if (info.Name?.startsWith('lima')) {
```
tony-sol, line 85: what exactly `name` is this?
i mean, isn't this a user-defined name? If so - we can't rely on it, cause it may be anything

afbjorklund: It is the hostname. The user can set it to "lima-foo" or "lima-bar", normally it is "lima-docker" [-rootful]

afbjorklund:
Forgot the change from "lime" to "lime green" ([Malachite](https://en.wikipedia.org/wiki/Malachite)), due to usability issues on light background...

* `#CCFF00` _electric lime_
* `#BFFF00` _lime_
* `#32CD32` _lime green_
* `#0BDA51` _malachite_

https://simple.wikipedia.org/wiki/Green

Lima

* `#A8CF35`

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilitySocketMappingStatus.svelte
```diff
@@ -16,7 +16,8 @@ let isLinux = $state(false);
 
 let dockerSocketMappingStatusInfo: DockerSocketMappingStatusInfo | undefined = $state(undefined);
 
-let engineType: 'kubernetes' | 'podman' | 'docker' | undefined = $state(undefined);
+let engineType: 'kubernetes' | 'podman' | 'docker' | 'lima' | undefined = $state(undefined);
```
axel7083: question (non-blocking): is the type here `DockerSocketServerInfoType` ? or something different, because I see `undefined`, and not `unknown` ?

If the type derive from DockerSocketServerInfoType like `DockerSocketServerInfoType | undefined` would be nice to use it

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilitySocketMappingStatus.svelte
```diff
@@ -16,7 +16,8 @@ let isLinux = $state(false);
...
+// 'lima' can provide any engine of 'kubernetes' or 'podman' or 'docker', either one or several
```
axel7083: suggestion (non-blocking): I think this comment would be better in the `packages/api/src/docker-compatibility-info.ts` ?

packages/renderer/src/lib/ui/ProviderInfoCircle.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
 import { providerColors } from './ProviderInfoCircle';
 
-let { type }: { type: 'kubernetes' | 'podman' | 'docker' | undefined } = $props();
+let { type }: { type: 'kubernetes' | 'podman' | 'docker' | 'lima' | undefined } = $props();
```
axel7083: question (non-blocking): same as https://github.com/podman-desktop/podman-desktop/pull/10447/files#r1893707560

tailwind-color-palette.json
```diff
@@ -110,6 +110,18 @@
     "800": "#1a4a72",
     "900": "#1b3f5f"
   },
+  "malachite": {
```
axel7083: cc @deboer-tim does adding color here is okey ?

axel7083 (approved):
Amazing contribution @afbjorklund :rocket: ! Very impressed. Code LGTM, I cannot test on windows, so will need someone from our reviewers that can test it

> I left a few comments/suggestion

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilitySocketMappingStatus.svelte
afbjorklund: The name "engineType" is not ideal, since it actually refers to the serverType.

i.e. the string is just converted from "Podman Machine OS" to "Podman Engine"

axel7083: Humm okay thanks for the details

packages/renderer/src/lib/ui/ProviderInfoCircle.svelte
afbjorklund: In this case it is the serverType as input, and that string is thrown into providerType

Which is different from ProviderInfo, because there connectionType is used as input

tailwind-color-palette.json
afbjorklund: Could use green, but that is used as "enabled"

See related discussion in https://github.com/podman-desktop/podman-desktop/issues/5609#issuecomment-2262847693

packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilitySocketMappingStatus.svelte
afbjorklund: Note that "kubernetes" is unused

benoitf (changes requested):
issue: we can't add new colors to the tailwind-color-palette

tailwind-color-palette.json
deboer-tim: Sorry, I completely missed the notification from Axel in Dec above. As per Florent's comment today, we have a recommended palette from design and are not supposed to add additional colors (plus we use variables to allow light/dark customization, etc).

I would support allowing each provider to supply a color outside of this list (and even have old commits that do it), but every time we've tried to show this in context it has been thrown out because it is too hard to avoid the user confusion of 'green provider = success, red provider = something is wrong'. So, we are not planning to allow provider colors unless/until we have a UX breakthrough for other places in the UI as well.

Net: I'd suggest just picking an existing color/variable that's close enough for now.

afbjorklund: Let's pick something that already exists then. That is how it works for the limited consoles (256) as well :-)

https://colorkit.co/color/a9cf38/ says "lime-600" so that should probably be fine, better than the current grey.

afbjorklund: > So, we are not planning to allow provider colors

Maybe should color it in the provider **type** instead?

```typescript
export const providerColors: Record<string, string> = {
  podman: 'bg-purple-600',
  docker: 'bg-sky-400',
  kubernetes: 'bg-sky-600',
  unknown: 'bg-gray-900',
};
```

Instead of trying to add another provider to the mix.

```
          <Label name="{dockerSocketMappingStatusInfo.serverInfo.name} is listening">
            <ProviderInfoCircle type={engineType} />
          </Label>
```

afbjorklund:
We can use the name for the string, and use type only for the color: (in this case, it is a `docker.sock`)

docker-sock-lima

And similar when running template://podman instead of template://docker, that is: a `podman.sock`

podman-sock-lima

Since "docker" doesn't mean docker type but Docker Desktop*, and "podman" means Podman Machine

It is trying to show **who** provides the OS and the VM, not what type of connection (docker/podman) it is...

----

\* added a small fix, to also recognize Docker Engine:

* #10456

Podman Desktop should _already_ recognize Podman Engine.

(except on Linux, where `/run/podman/podman.sock` is unreachable)

afbjorklund:
It is mostly hypothetical, anyway. Since the lima templates recommend to use configuration instead of override:

```yaml
message: |
  To run `docker` on the host (assumes docker-cli is installed), run the following commands:
  ------
  docker context create lima-{{.Name}} --docker "host=unix://{{.Dir}}/sock/docker.sock"
  docker context use lima-{{.Name}}
  docker run hello-world
  ------
```

```yaml
message: |
  To run `podman` on the host (assumes podman-remote is installed), run the following commands:
  ------
  podman system connection add lima-{{.Name}} "unix://{{.Dir}}/sock/podman.sock"
  podman system connection default lima-{{.Name}}
  podman{{if eq .HostOS "linux"}} --remote{{end}} run quay.io/podman/hello
  ------
```

So the user has to manually replace the `/var/run/docker.sock` with a symlink, for this information to be in effect.

afbjorklund:
Not so happy with the result, since it looks like it is the Lima version (while it is the Docker or Podman version showing)

The OS and Arch seems to be fully un-standardized, but that's a bug with the underlying compat API

PR #10456, opened 2024-12-20
fix: add server type for default Docker Engine

afbjorklund:
### What does this PR do?

Show Docker as "docker" instead of "unknown"

### Screenshot / video of UI

Before
Image

After
pd-engine

https://github.com/podman-desktop/podman-desktop/issues/9351#issuecomment-2555794075

### How to test this PR?

Make sure that Docker Engine is running, on Linux

### DIFF
--- a/packages/main/src/plugin/docker/docker-compatibility.ts
+++ b/packages/main/src/plugin/docker/docker-compatibility.ts
@@ -20,7 +20,7 @@ import { promises } from 'node:fs';
 
 import Dockerode from 'dockerode';
 
-import { isMac, isWindows } from '/@/util.js';
+import { getHostname, isLinux, isMac, isWindows } from '/@/util.js';
 import type { DockerSocketMappingStatusInfo, DockerSocketServerInfoType } from '/@api/docker-compatibility-info.js';
 import { ExperimentalSettings } from '/@api/docker-compatibility-info.js';
 
@@ -64,7 +64,7 @@ export class DockerCompatibility {
   }
 
   protected getTypeFromServerInfo(
-    info: { OperatingSystem?: string },
+    info: { Name?: string; OperatingSystem?: string },
     podmanInfo?: unknown,
   ): DockerSocketServerInfoType {
     if (info.OperatingSystem === 'Docker Desktop') {
@@ -72,6 +72,9 @@ export class DockerCompatibility {
     } else if (info.OperatingSystem === 'podman' || podmanInfo) {
       // if podman info is available, then it is podman
       return 'podman';
+    } else if (isLinux() && info.Name === getHostname()) {
+      // Docker Engine
+      return 'docker';
     }
     return 'unknown';
   }
--- a/packages/main/src/plugin/docker/docker-compatibility.spec.ts
+++ b/packages/main/src/plugin/docker/docker-compatibility.spec.ts
@@ -44,7 +44,7 @@ const providerRegistry: ProviderRegistry = {
 
 export class TestDockerCompatibility extends DockerCompatibility {
   public override getTypeFromServerInfo(
-    info: { OperatingSystem?: string },
+    info: { Name?: string; OperatingSystem?: string },
     podmanInfo?: unknown,
   ): DockerSocketServerInfoType {
     return super.getTypeFromServerInfo(info, podmanInfo);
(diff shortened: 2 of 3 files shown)


packages/main/src/plugin/docker/docker-compatibility.ts
```diff
@@ -69,14 +69,17 @@ export class DockerCompatibility {
...
     podmanInfo?: unknown,
   ): DockerSocketServerInfoType {
     if (info.OperatingSystem === 'Docker Desktop') {
       return 'docker';
     } else if (info.OperatingSystem === 'podman' || podmanInfo) {
       // if podman info is available, then it is podman
       return 'podman';
+    } else if (isLinux() && info.Name === getHostname()) {
+      // Docker Engine
+      return 'docker';
```
benoitf, line 77: question: hello, I don't see how this part can be specific to docker and not the others. IMHO relying only on the hostname is too weak.

afbjorklund: It just assumes that if there is a local docker.sock that is not podman, then it is provided by docker

(The hostname check is just to verify that it is not being tunneled from a VM or a remote server)

afbjorklund: But you are not wrong, for instance nerdctld would show up as "docker" too...

One could look at Platform (if available), and look for "Docker Engine" there?

```json
  "Server": {
    "Platform": {
      "Name": "Docker Engine - Community"
    },
```

It could be empty (if you build Moby yourself), or missing (like in older versions)

https://github.com/moby/moby/commit/9152e63290e4a4e586b811cce39082efc649b912

docker.io: `"Name": ""`

afbjorklund: `!info.Platform?.Name || info.Platform?.Name === '' || info.Platform?.Name?.startsWith('Docker Engine'`

But it does seem paranoid, I don't _anyone_ (but Podman) that is overwriting the Docker `/var/run/docker.sock`

afbjorklund: Normally when using a different engine, you would config the docker client but leave the docker server alone.

Something like DOCKER_HOST: https://github.com/afbjorklund/nerdctld?tab=readme-ov-file#running-daemon

afbjorklund:
The documentation for this feature (Docker Compatibility) is currently missing for Linux:

https://podman-desktop.io/docs/migrating-from-docker/managing-docker-compatibility

It doesn't actually **do** anything (but show the system socket info), the rest is just config:

https://podman-desktop.io/docs/migrating-from-docker/using-the-docker_host-environment-variable

pd-compat

Just showing the system daemon: https://packages.ubuntu.com/focal-updates/docker.io

```console
$ docker context ls
NAME        DESCRIPTION                               DOCKER ENDPOINT                            ERROR
default *   Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                
```

afbjorklund:
Here is the documentation for podman-docker (replacing Docker Engine):

* https://github.com/podman-desktop/podman-desktop/pull/10481

```console
$ docker context ls
NAME       DESCRIPTION                               DOCKER ENDPOINT                            ERROR
default    Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                
podman *                                             unix:///run/user/1000/podman/podman.sock   
```

axel7083 (commented):
I can see that the `Name` property you are using is from the

https://github.com/podman-desktop/podman-desktop/blob/92c70ba5806f91751fff4183b22475db67b0f550/packages/main/src/plugin/docker/docker-compatibility.ts#L111

Is there any other information in this that could be used instead of the Name?

I share the opinion of @benoitf in https://github.com/podman-desktop/podman-desktop/pull/10456/files#r1893993722 let's say we have another engine which is not docker nor podman, we would put `docker`🤔 

Something like `!podman && info?.Platform?.Name?.startsWith('Docker')` seems also less problematic, but as you mention podman overwrite this but checking `Not podman` first would maybe be enough

afbjorklund:
You can leave it as "unknown" for the system (distro) docker, and only use "docker" for the vendor docker from docker.com

afbjorklund:
There is no Platform in `info`, it is only available in `version`. And even in there, there are no guarantees.

```console
$ docker version --format json | jq .Server.Platform
{
  "Name": ""
}
```

Issue #10240, opened 2024-12-05
svelte5 migration : Phase 1


benoitf:
### Epic domain

update components raising warnings in the code and use runes mode inside

- [x]  (packages/​renderer/​src/​lib/​actions/​ContributionActions.svelte:26:0) https://github.com/podman-desktop/podman-desktop/pull/10491
- [x]  (packages/​renderer/​src/​lib/​extensions/​ExtensionDetailsReadme.svelte:12:0) https://github.com/podman-desktop/podman-desktop/pull/10492
- [x]  (packages/​renderer/​src/​lib/​pod/​PodsList.svelte:35:0) https://github.com/podman-desktop/podman-desktop/pull/10459
- [x]  (packages/​renderer/​src/​lib/​preferences/​PreferencesRendering.svelte:27:0) https://github.com/podman-desktop/podman-desktop/pull/10496
- [x]  (packages/​renderer/​src/​lib/​preferences/​PreferencesRenderingItem.svelte:43:0) https://github.com/podman-desktop/podman-desktop/pull/10494
- [x]  (packages/​renderer/​src/​lib/​preferences/​PreferencesRenderingItemFormat.svelte:35:0) https://github.com/podman-desktop/podman-desktop/pull/10495
- [x]  (packages/​renderer/​src/​lib/​ui/​AuditMessageBox.svelte:12:0) https://github.com/podman-desktop/podman-desktop/pull/10493
- [x]  (packages/​renderer/​src/​lib/​ui/​ListItemButtonIcon.svelte:36:0) https://github.com/podman-desktop/podman-desktop/pull/10490
- [x]  (packages/​renderer/​src/​lib/​ui/​LoadingIconButton.svelte:45:0) https://github.com/podman-desktop/podman-desktop/pull/10489
- [x]  (packages/​renderer/​src/​lib/​ui/​ProviderResultPage.svelte:40:0) https://github.com/podman-desktop/podman-desktop/pull/10446

benoitf:
done

PR #10459, opened 2024-12-20
chore: move PodsList to svelte5 runes mode

benoitf:
### What does this PR do?
Use runes of svelte5

part of https://github.com/podman-desktop/podman-desktop/issues/10240

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/pod/PodsList.svelte
+++ b/packages/renderer/src/lib/pod/PodsList.svelte
@@ -31,23 +31,32 @@ import PodColumnStatus from './PodColumnStatus.svelte';
 import PodEmptyScreen from './PodEmptyScreen.svelte';
 import type { PodInfoUI } from './PodInfoUI';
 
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/pod/PodsList.svelte
```diff
@@ -31,23 +31,32 @@ import PodColumnStatus from './PodColumnStatus.svelte';
...
-$: searchPattern.set(searchTerm);
+interface Props {
+  searchTerm?: string;
+}
+
+let { searchTerm = $bindable('') }: Props = $props();
+
+$effect(() => {
+  searchPattern.set(searchTerm);
+});
```
axel7083, line 42: remark (non-blocking): this pattern is weird, not blocking as was already the case in the non-rune mode, but this is looking odd

axel7083 (approved):
LGTM

feloy (approved):

PR #10461, opened 2024-12-20
fix: add hidden input component in Dropdown for form submissions

### LINKED ISSUE #10435: dropdown value is not sent from the form when creating a new provider
### Bug description

When creating a new provider (tried creating a new Podman machine and Kind cluster), the values of the dropdown options are not being taken from the form and sent to the backend.
This bug is a possible regression due to https://github.com/podman-desktop/podman-desktop/pull/9953

It looks like this bug only affects Dropdown values when used in form submission with `new FormData()` to collect all values as `new FormData()` only adds basic input components (e.g. select) and their values.
Most other places use the value of the custom Dropdown component and not the hidden select

### Operating system

macOS Sequoia 15.1.1, M1

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. Go to Settings -> resources
2. Click on `Create new...` for Podman
3. Pick the non-default option for `Provider Type` (GPU enabled LibKrun)

Expected: LibKrun machine
Actual result: default (Apple HyperVisor) machine

### Relevant log output

```shell

```

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
This PR adds a hidden input component to ensure that the value of the Dropdown is always passed in form submissions, regardless if children elements or property are used.

From a quick check, it looks like the only place that uses the value of the hidden `select` in `Dropdown` is `new FormData()` in `PreferencesConnectionCreationOrEditRendering.svelte` as it relies on basic input components to gather the form data. It also seems like the value of `select` cannot be set to a value it doesn't contain, so if there are no children to render in the hidden `select`, its value cannot be updated to the value of the `Dropdown`. 

Fixes https://github.com/podman-desktop/podman-desktop/issues/10435

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/dropdown/Dropdown.svelte
+++ b/packages/ui/src/lib/dropdown/Dropdown.svelte
@@ -231,6 +231,12 @@ function onWindowClick(e: Event): void {
   {/if}
 
   <select use:buildOptions class="hidden" name={name} bind:value={value}>
-    {@render children?.()}
+    {#if !children || children.length === 0}
+      {#each options as option}
+        <option value={option.value}></option>
+      {/each}
+    {:else}
+      {@render children?.()}
+    {/if}
   </select>
 </div>
--- a/packages/ui/src/lib/dropdown/Dropdown.svelte
+++ b/packages/ui/src/lib/dropdown/Dropdown.svelte
@@ -230,13 +230,9 @@ function onWindowClick(e: Event): void {
     </div>
   {/if}
 
-  <select use:buildOptions class="hidden" name={name} bind:value={value}>
-    {#if !children || children.length === 0}
-      {#each options as option}
-        <option value={option.value}></option>
-      {/each}
-    {:else}
-      {@render children?.()}
-    {/if}
+  <input name={name} value={value} type="hidden" aria-label="hidden input"/>
+
+  <select use:buildOptions class="hidden" bind:value={value}>
+    {@render children?.()}
   </select>
 </div>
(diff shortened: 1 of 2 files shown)


packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -231,6 +231,12 @@ function onWindowClick(e: Event): void {
   {/if}
 
   <select use:buildOptions class="hidden" name={name} bind:value={value}>
-    {@render children?.()}
+    {#if !children || children.length === 0}
+      {#each options as option}
+        <option value={option.value}></option>
+      {/each}
```
SoniaSandler: I could also do only `<option value={value}></option>` instead of the entire `each` block, with `{value}` being the dropdown's value (tried and it worked for the Kind and Minikube creation forms), but I'm not sure what implications it might have with only one value that is dynamic and maybe the `bind:value={value}` part

jeffmaury (approved):
LGTM

SoniaSandler:
Following a disscusion with @deboer-tim, I've used a similar approach to `svelte-select` as seen in https://github.com/rob-balfre/svelte-select/blob/4e6aa96afd52a384beb2af4092ee49424e73867f/src/lib/Select.svelte#L822. This way ensures that this component will work with any form submission and it will not rely on future changes to the hidden select component

jeffmaury (approved):
LGTM

deboer-tim (approved):

PR #10463, opened 2024-12-22
chore: make getProxySettingsFromSystem() work on FreeBSD

arrowd:
### What does this PR do?

This change adds the `isUnixLike()` function which returns true for Linux and FreeBSD OSes. It then used to fix `getProxySettingsFromSystem()` for the FreeBSD case.

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/proxy-system.spec.ts
+++ b/packages/main/src/plugin/proxy-system.spec.ts
@@ -47,6 +47,7 @@ function setupPlatform(platform: Platform): void {
   vi.spyOn(util, 'isWindows').mockReturnValue(platform === Platform.WINDOWS);
   vi.spyOn(util, 'isMac').mockReturnValue(platform === Platform.MACOS);
   vi.spyOn(util, 'isLinux').mockReturnValue(platform === Platform.LINUX);
+  vi.spyOn(util, 'isUnixLike').mockReturnValue(platform === Platform.LINUX);
 }
 
 describe('Windows platform tests', () => {
--- a/packages/main/src/plugin/proxy-system.ts
+++ b/packages/main/src/plugin/proxy-system.ts
@@ -20,7 +20,7 @@ import { promisify } from 'node:util';
 import { type ProxySettings } from '@podman-desktop/api';
 import WinReg from 'winreg';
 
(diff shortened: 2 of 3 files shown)


jeffmaury (changes requested):
Core cannot call code from an extension. The isLinux method to be called is located in util.ts

arrowd:
Ok, will do. Should I also remove the code I added to `os.ts`?

arrowd:
Can we get this in?

benoitf (dismissed):
@arrowd are you running Podman Desktop on FreeBSD ?

note: the change in extensions/compose should be removed

benoitf:
FYI you'll need to edit the commit to be compliant with https://www.conventionalcommits.org/en/v1.0.0/

benoitf:
semantic PR check is failing for now

arrowd:
> @arrowd are you running Podman Desktop on FreeBSD ?

Yes, at least I'm trying to do so. I had some success with previous version - the app started, but there was nothing useful that can be done there, because our Podman implementation is missing `podman machine *` subcommands family.

Now this version doesn't start due to that proxy engine error, whcih I'm hoping to fix.

arrowd:
It seems I still fail to please the "Semantic PR" check. Do I have to create issues for both of my commits?

benoitf:
you could edit your history and only keep one

but they need to follow the convention like

`chore: make getProxySettingsFromSystem() work on FreeBSD` and the same for the PR's title

note that other checks are also failing

benoitf (commented):
will need a ✅ on all PR checks to be able to merge

benoitf:
@arrowd also if you can apply the official template in the PR as tools are searching for the correct format
https://raw.githubusercontent.com/podman-desktop/podman-desktop/refs/heads/main/.github/PULL_REQUEST_TEMPLATE.md

arrowd:
Thanks for guiding me through the submission!

benoitf (approved):

jeffmaury (approved):
LGTM

arrowd:
So, can we get this in?

arrowd:
Great, thank you. Can I expect this change to be present in the upcoming release (1.16.2 or whatever it is) ?

PR #10481, opened 2024-12-29
docs: add Linux documentation for Docker Compatibility

afbjorklund:
### How to test this PR?


afbjorklund:
The system socket is not reachable. This is normal behaviour, with podman (it requires using `sudo`).

`permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/version": dial unix /var/run/docker.sock: connect: permission denied`

`/var/run/docker.sock -> /run/podman/podman.sock`

`srw-rw---- 1 root root`

benoitf (approved):

PR #10516, opened 2025-01-03
chore: allow to set values for dropdown in DockerCompatibility

### LINKED ISSUE #10514: allow extensions to provide and customize drop down items for the docker compatibility page

today, it's for an extension to provide markdown description containing buttons.

But in case the property is an enum, the page should display a dropdown and then the extension should be able to specify the values of the dropdown dynamically as we don't know the list when we define the values in package.json file.

- registering an enum property for the scope DockerCompatibility should then display the values in a drop down component
- extension should be able to set the values using the context.setValue with 'DockerCompatibility' scope
- extension can track configuration changes subscribing the event on configuration change (using the 'DockerCompatibility' scope)

related to https://github.com/podman-desktop/podman-desktop/issues/9351

### Additional context

_No response_

benoitf:
### What does this PR do?
extensions can register enum values for a DockerCompatibility property using the `context.setValue` and scope 'DockerCompatibility'
It'll register the dropdown. Each time the user select a value in the dropdown, a configuration change event is thrown that extensions can intercept

fixes https://github.com/podman-desktop/podman-desktop/issues/10514

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityContributions.svelte
+++ b/packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityContributions.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 /* eslint-disable import/no-duplicates */
+import { Dropdown } from '@podman-desktop/ui-svelte';
 import { onDestroy, onMount } from 'svelte';
 import { get, type Unsubscriber, type Writable } from 'svelte/store';
 
@@ -50,8 +51,17 @@ onDestroy(() => {
   }
 });
 
+interface EnumItem {
+  label: string;
+  value: string;
+  selected: boolean;
+}
+
 interface PropertyWithDisplayName extends IConfigurationPropertyRecordedSchema {
+  id: string;
   displayName: string;
+  selectedValue?: string;
+  enumItems?: EnumItem[];
 }
 
 interface GroupItem {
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/preferences/docker-compat/PreferencesDockerCompatibilityContributions.svelte
```diff
@@ -115,12 +125,35 @@ function extractGroupItems(): void {
...
+
+        // do we have a dynamic enum ?
+        if (property.enum) {
+          // get the enum from the context using the following key
+          // ${groupName}.DockerCompatibility.${property.id}
+
+          // get the enum from the context
+          const enumKey = `${name}.DockerCompatibility.${property.id}`;
+          const enumItems = globalContext.getValue<EnumItem[]>(enumKey);
+          if (enumItems) {
```
axel7083: As this may be set by an extension, should we check the type instead of assuming ? or it is okay to assert ?

benoitf: yes I thought the same before issuing the PR (to be more robust)
it'll be an addendum

benoitf: handled by a new commit

axel7083 (approved):
LGTM, thanks for the change

feloy (approved):

jeffmaury (approved):
LGTM

PR #10526, opened 2025-01-06
chore: move docker cli context select implem from core to docker extension

### LINKED ISSUE #10517: move the implementation of selecting a docker cli context from Podman Desktop core to the docker extension

There is an open issue #9351 where it's not possible to list some contexts using some custom DOCKER_CONFIG properties.
Then we should provide a way to monitor a different DOCKER_CONFIG path with a custom property.
This property should not be part of the 'Podman Desktop core'

With the extensibility of the Docker Compatibility page, it's now possible to have extensions contributing to this page.
Then, the part managing the cli context should go to a 3rd party extension (docker extension)

move all the logic to handle the CLI context from Podman Desktop core to the docker extension.

then it'll be possible to handle and solve #9351 

_No response_

### Additional context

_No response_

benoitf:
### What does this PR do?
docker cli context parsing was done in the core part. It should have been done inside the docker extension.
Move then all the current logic from the core to the extension.
It'll allow to fix new issues as for example requiring a custom DOCKER_CONFIG etc. These settings are only meaningful from the docker extension, not from the core.

### Screenshot / video of UI

before:
image

after:
image

fixes https://github.com/podman-desktop/podman-desktop/issues/10517

### How to test this PR?

Try to list and toggle a docker context


extensions/docker/package.json
```diff
@@ -25,5 +25,27 @@
...
+        "category": "Docker",
+        "enablement": "false"
+      }
+    ],
+    "configuration": {
+      "title": "Docker",
+      "properties": {
+        "docker.cli.context": {
+          "type": "string",
+          "enum": [""],
```
axel7083, line 43: Why do we have an empty string ? Does the extension change this value ?

axel7083: Are we forced to defined an empty string here ? Since this is set through the context

benoitf: it's to tell that the property is an enum

benoitf: could be an empty array but I think I had error reported but could be a false positive error feedback with empty

axel7083 (approved):
LGTM, disabling the docker extension do remove the card.

Sidenote: the docker cli context feel a bit weird since I don't have docker ?

image

> I don't have docker on my machine, so I cannot test it fully.

benoitf:
here, maybe we could register the entry only if the `docker` CLI is being found on the system.

But today it was always displayed as well (no matter if the CLI is installed or not), so I kept it using the same behavior.

feloy (approved):

PR #10527, opened 2025-01-06
fix(push-manifest): terminal overflow

axel7083:
### What does this PR do?

The `PushManifestModal.svelte` is heavily inspired from `PushImageModal`. In the PushImageModal the following line ensure the terminal properly fit the area

https://github.com/podman-desktop/podman-desktop/blob/40c3807845576a8e31bb649e208d10fd1e60bc9f/packages/renderer/src/lib/image/PushImageModal.svelte#L36

so this PR simply add this missing line to the PushManifestModal. Moreover, the PushImageTerminal had some classes to define the sizes and height, this PR copy the values that were missing to ensure consistency 

https://github.com/podman-desktop/podman-desktop/blob/40c3807845576a8e31bb649e208d10fd1e60bc9f/packages/renderer/src/lib/image/PushImageModal.svelte#L111-L113

### Screenshot / video of UI

**Before**
image

**After**
image

### What issues does this PR fix or reference?

Fixes https://github.com/podman-desktop/podman-desktop/issues/10306

### How to test this PR?

- [x] Tests are covering the bug fix or the new feature

Issue #9726, opened 2024-11-03
Feature request: search in Pods


savbran:
### Is your enhancement related to a problem? Please describe

Hi! It could be great if the pod log viewer could have the possibility to search text in it. Do you think it's possible to have it one day! Thanks for the attention.

### Describe the solution you'd like

The possibility to press CTRL + F and see the text find field on the log tab.

### Describe alternatives you've considered

none

### Additional context

My request is related to Pods section, Log tab showing log for the pod

slemeur:
Thanks for creating the feature request issue @savbran

axel7083:
We are using `@xterm/xterm` for our logging component, we have a svelte component for all terminals `TerminalWindow`[^1]

Since we rely heavily on `@xterm/*` we should look at what they offer for searching, thanksfully they have a search addon [@xterm/addon-search](https://www.npmjs.com/package/@xterm/addon-search). I have not looked much into it, but this could be a way to integrate a search bar.

### Several issues still exists

#### Style consistency between Monaco editor and Terminal

Today, you can CTRL+F in the inspect tab to search

![Image](https://github.com/user-attachments/assets/07a2d739-fc36-415f-ba51-ee4cfddae68c)

The search addon is not a ui component, but an api that we can use, we could have to handle the ui component for the search bar. We have ongoing work with the logs, to provide the ability to clean logs (see https://github.com/podman-desktop/podman-desktop/issues/8311)


[^1]: packages/renderer/src/lib/ui/TerminalWindow.svelte

axel7083:
Since this issue was in the sprint I assigned myself because I like it :) 

Here is a POC using the [@xterm/addon-search](https://www.npmjs.com/package/@xterm/addon-search) library

https://github.com/user-attachments/assets/5cff88b7-46fa-4dac-b1da-74af6f703aad

PR #10531, opened 2025-01-06
test(ui): adding unit tests for TerminalWindows

axel7083:
### What does this PR do?

While working on https://github.com/podman-desktop/podman-desktop/pull/10527 I noticed we do not have much tests for the `TerminalWindows` component.

Since we are using it in a lot of places, adding unit tests. 

### Screenshot / video of UI

N/A

Required for https://github.com/podman-desktop/podman-desktop/issues/9726

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/TerminalWindows.spec.ts
+++ b/packages/renderer/src/lib/ui/TerminalWindows.spec.ts
@@ -22,7 +22,7 @@ import { render } from '@testing-library/svelte';
 import { FitAddon } from '@xterm/addon-fit';
 import { Terminal } from '@xterm/xterm';
 import { writable } from 'svelte/store';
-import { beforeEach, expect, test, vi } from 'vitest';
+import { afterEach, beforeEach, expect, test, vi } from 'vitest';
 
 import TerminalWindow from '/@/lib/ui/TerminalWindow.svelte';
 
@@ -54,10 +54,10 @@ beforeEach(() => {
 
   vi.mocked(Terminal).mockReturnValue(TerminalMock);
   vi.mocked(FitAddon).mockReturnValue(FitAddonMock);
+});
 
-  Object.defineProperty(window, 'addEventListener', {
-    value: vi.fn(),
-  });
+afterEach(() => {
+  vi.restoreAllMocks();
 });
 
 test('expect terminal constructor to have been called on mount', async () => {
@@ -139,6 +139,9 @@ test('addon fit should be loaded on mount', async () => {
 });
 
 test('matchMedia resize listener should trigger fit addon', async () => {
+  // spy the event listener
+  vi.spyOn(window, 'addEventListener');
+
   render(TerminalWindow, {
     terminal: writable() as unknown as Terminal,
   });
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/ui/TerminalWindows.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+  onResize: vi.fn(),
+  dispose: vi.fn(),
+  loadAddon: vi.fn(),
+  open: vi.fn(),
+  write: vi.fn(),
+} as unknown as Terminal;
+
+const FitAddonMock: FitAddon = {
+  fit: vi.fn(),
+} as unknown as FitAddon;
```
benoitf: question: I'm wondering why TerminalMock definition is not part of the vi.mock('@xterm/xterm') definition and FitAddonMock in @xterm/addon-fit

and the pattern
```
   vi.mocked(Terminal).mockReturnValue(TerminalMock);
   vi.mocked(FitAddon).mockReturnValue(FitAddonMock);
```

seems to not fit in an existing test pattern.

axel7083: If I do the following 

```ts
vi.mock('@xterm/addon-fit', () => ({
  FitAddon: vi.fn().mockReturnedValue({
      fit: vi.fn(),
  }),
}));
```

The mockReturnedValue is reset if I have `vi.resetAllmock()` in the `beforeEach`. So I have to define everytime I reset it 🤔 

How do you mock the constructor then and access the object created ? 

Defining `TerminalMock` and `FitAddonMock` allow me quickly access the object, and check if some functions on it has been called.

packages/renderer/src/lib/ui/TerminalWindows.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+beforeEach(() => {
+  vi.resetAllMocks();
+
+  vi.mocked(Terminal).mockReturnValue(TerminalMock);
+  vi.mocked(FitAddon).mockReturnValue(FitAddonMock);
+
+  Object.defineProperty(window, 'addEventListener', {
```
jeffmaury: I thought this was not required anymore

axel7083: AddEventListener is something in the default windows api, not something we mock through the IPC (my understanding is that we only mock the IPC we defined through electron)

jeffmaury: We already process matchMedia so maybe we should include it as well (see https://github.com/podman-desktop/podman-desktop/pull/10426/files).
@benoitf WDYT ?

benoitf: yes I think it's good to move all common methods to the global vite.tests.setup file

axel7083: We seems to have a lot of tests which uses the logic of the event system, so we cannot mock it globally :/

benoitf: in that case, as the addEventListener already exists, we should "spy" and not replace the definition of the method by a mock

axel7083: I removed the define property and replaced it with a vi#spy

packages/renderer/src/lib/ui/TerminalWindows.spec.ts
axel7083: Following our discussion, found a nice way to simplify all the tests: everyone happy!

benoitf (approved):
thanks for the changes

PR #10532, opened 2025-01-06
chore(test): handle cli tool confirmation dialog

amisskii:

### What does this PR do?
Implements a confirmation dialog for installing CLI tools, applicable, for example, to the Minikube CLI installation.

### How to test this PR?

### DIFF
--- a/tests/playwright/src/model/pages/cli-tools-page.ts
+++ b/tests/playwright/src/model/pages/cli-tools-page.ts
@@ -19,6 +19,8 @@
 import test, { expect as playExpect } from '@playwright/test';
 import type { Locator, Page } from 'playwright';
 
+import { handleConfirmationDialog } from '/@/utility/operations';
+
 import { SettingsPage } from './settings-page';
 
 export class CLIToolsPage extends SettingsPage {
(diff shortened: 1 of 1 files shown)


axel7083 (approved):

tests/playwright/src/model/pages/cli-tools-page.ts
```diff
@@ -87,6 +89,12 @@ export class CLIToolsPage extends SettingsPage {
 
       await playExpect(this.getVersionSelectionButton(version)).toBeEnabled();
       await this.getVersionSelectionButton(version).click();
+
+      const confirmationDialog = this.page.getByRole('dialog', { name: toolName });
+      if ((await confirmationDialog.count()) > 0) {
```
cbr7: This condition is vulnerable to a race condition, count() method has no implicit waiter associated, so if the dialog isn't available at the moment of checking, but instead appears a fraction of a second after, the popup will no be handled correctly.

cbr7 (changes requested):
Unhandled possible race condition.

cbr7 (approved):

PR #10541, opened 2025-01-07
docs: code guideline to use mocks vs spies

feloy:

### What does this PR do?

add a code guideline to choose between using mocks vs spies, and how to reset/restore them.

### Screenshot / video of UI

N/A

Part of #9821

### How to test this PR?

### DIFF
--- a/CODE-GUIDELINES.md
+++ b/CODE-GUIDELINES.md
@@ -40,7 +40,15 @@ test('...', () => {
 
 ### Mock complete modules, spy on parts of module for specific tests
 
-To reduce as much as possible the coverage of the test to the module you are testing, mock completely the imported modules, with `vi.mock('/path/to/module)`, and define mock implementation for each test with `vi.mocked(function).mock...()`.
+When testing a module, you have to decide for each imported module if you mock the entire module or if you spy on specific functions of the module
+for specific tests and keep the real implementation for the other functions.
+
+System modules (`node:fs`, etc) are most generally mocked, so you are sure that unit tests are executed in isolation of the system. For internal modules,
+it's up to you to decide if you want to mock them or not, depending on the coverage you want for the unit tests.
+
+#### Mock a complete module
+
+Mock completely an imported module with `vi.mock('/path/to/module)`, and define mock implementation for each test with `vi.mocked(function).mock...()`.
 
 Use `vi.resetAllMocks()` in the top-level `beforeEach` to reset all mocks to a no-op function returning `undefined` before to start each test.
 
@@ -84,23 +92,31 @@ test('file existence is not defined', () => {
 });
 ```
 
-When you want to mock only one or a small number of functions of a module (for example a function of the module you are testing) for a particular test, you can use `vi.spyOn(module, 'function')` to mock only `function` and keep the original implementation for the rest of the module.
+#### Spy on a function for a specifc test
+
+When you want to mock only one or a small number of functions of a module (for example a function of the module you are testing, or a function of an helper module from which you want to use real implementation for some functions) for a particular test, you can use `vi.spyOn(module, 'function')` to mock only `function` and keep the original implementation for the rest of the module.
 
 To be sure that the spied function is restored to its original implementation for the other tests, use `vi.mocked(function).mockRestore()` in the `afterEach` of the tests declaring this spy.
 
 ```ts
+// helpers.ts
+export function f1(): boolean {
+  return true;
+}
+
 // mymodule.ts
-class MyModuleToTest {
-  f1(): boolean {
-    return true;
-  }
+import { f1 } from './helpers.js';
+
+export class MyModuleToTest {
   f2(): boolean {
-    return this.f1();
+    return f1();
   }
 }
 
 // mymodule.spec.ts
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { MyModuleToTest } from './mymodule.js';
+import * as helpers from './helpers.js';
 
 let myModuleToTest: MyModuleToTest;
 
@@ -110,17 +126,17 @@ beforeEach(() => {
 
 describe('f1 returns false', () => {
   beforeEach(() => {
-    vi.spyOn(myModuleToTest, 'f1').mockReturnValue(false);
+    vi.spyOn(helpers, 'f1').mockReturnValue(false);
   });
 
   afterEach(() => {
     // restore f1 to its original implementation
-    vi.mocked(myModuleToTest.f1).mockRestore();
+    vi.mocked(helpers.f1).mockRestore();
   });
 
   test('f2 returns false', () => {
     expect(myModuleToTest.f2()).toBeFalsy();
-    expect(myModuleToTest.f1).toHaveBeenCalledOnce();
+    expect(helpers.f1).toHaveBeenCalledOnce();
   });
 });
 
@@ -128,6 +144,6 @@ test('f2 returns true', () => {
   // use the original implementation of f1
   expect(myModuleToTest.f2()).toBeTruthy();
   // this won't work, as f1 is not spied for this test
-  // expect(myModuleToTest.f1).toHaveBeenCalledOnce();
+  // expect(helpers.f1).toHaveBeenCalledOnce();
 });
 ```
--- a/CODE-GUIDELINES.md
+++ b/CODE-GUIDELINES.md
@@ -86,7 +86,7 @@ test('file existence is not defined', () => {
 
 When you want to mock only one or a small number of functions of a module (for example a function of the module you are testing) for a particular test, you can use `vi.spyOn(module, 'function')` to mock only `function` and keep the original implementation for the rest of the module.
 
-To be sure that the spied function is restored to its original implementation for the other tests, use `vi.restoreAllMocks()` in the `afterEach` of the tests declaring this spy.
+To be sure that the spied function is restored to its original implementation for the other tests, use `vi.mocked(function).mockRestore()` in the `afterEach` of the tests declaring this spy.
 
 ```ts
 // mymodule.ts
@@ -115,7 +115,7 @@ describe('f1 returns false', () => {
 
   afterEach(() => {
     // restore f1 to its original implementation
-    vi.restoreAllMocks();
+    vi.mocked(myModuleToTest.f1).mockRestore();
   });
 
   test('f2 returns false', () => {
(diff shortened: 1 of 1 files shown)


jeffmaury (approved):
LGTM

benoitf (commented):
I'm not sure to follow why we would need to do mockRestore

as myFuncSpy is a new instance at each test run (we recreate the object on each test run)

feloy:
> I'm not sure to follow why we would need to do mockRestore
> 
> as myFuncSpy is a new instance at each test run (we recreate the object on each test run)

Some test outside of (and after) the `describe('spy on myfunc')` may want to use the original `myfunc`, but as I understand, without `mockRestore`, the function would still be mocked.

benoitf:
I tried

```ts
import type { MockInstance} from 'vitest';
import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';

describe('spy on myfunc', () => {
  let myfuncSpy: MockInstance<() => void>;

  beforeEach(() => {
    myfuncSpy = vi.spyOn(console, 'log');
  });

  afterEach(() => {
    // myfuncSpy.mockRestore();
  });

  test('use console.log', () => {
    console.log('hello world');
    expect(myfuncSpy).toHaveBeenCalled();
  });

  test('another one', () => {
    console.warn('warn hello world');
    expect(myfuncSpy).not.toHaveBeenCalled();
  });


  test('third one', () => {
    // not calling so not been called
    expect(myfuncSpy).not.toHaveBeenCalled();
  });



});

```


so no mockRestore and I don't see any issues

feloy:
> I tried
> 
> ```ts
> import type { MockInstance} from 'vitest';
> import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
> 
> describe('spy on myfunc', () => {
>   let myfuncSpy: MockInstance<() => void>;
> 
>   beforeEach(() => {
>     myfuncSpy = vi.spyOn(console, 'log');
>   });
> 
>   afterEach(() => {
>     // myfuncSpy.mockRestore();
>   });
> 
>   test('use console.log', () => {
>     console.log('hello world');
>     expect(myfuncSpy).toHaveBeenCalled();
>   });
> 
>   test('another one', () => {
>     console.warn('warn hello world');
>     expect(myfuncSpy).not.toHaveBeenCalled();
>   });
> 
> 
>   test('third one', () => {
>     // not calling so not been called
>     expect(myfuncSpy).not.toHaveBeenCalled();
>   });
> 
> 
> 
> });
> ```
> 
> so no mockRestore and I don't see any issues

Right, it is necessary only when the spy is also mocked, like in:

```ts

describe('spy on myfunc', () => {
  let myfuncSpy: MockInstance<() => void>;

  beforeEach(() => {
    myfuncSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
  });

  afterEach(() => {
    myfuncSpy.mockRestore();
  });

  test('use console.log', () => {
    console.log('hello world');
    expect(myfuncSpy).toHaveBeenCalled();
  });

  test('another one', () => {
    console.warn('warn hello world');
    expect(myfuncSpy).not.toHaveBeenCalled();
  });


  test('third one', () => {
    // not calling so not been called
    expect(myfuncSpy).not.toHaveBeenCalled();
  });
});

test('real console', () => {
  console.log('==> should be displayed on console');
});
```

In this example, the `real console` does not display anything if `mockRestore` is not called.

I'll change to add an implementation to the spy.

benoitf:
but we recommend to use resetAllMocks to have clean state on every test, so still I don't see why we would need mockRestore ?

it'll work as expected:

```
describe('spy on myfunc', () => {
  let myfuncSpy: MockInstance<() => void>;

  beforeEach(() => {
    vi.resetAllMocks();
    myfuncSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
  });

  test('use console.log', () => {
    console.log('hello world');
    expect(myfuncSpy).toHaveBeenCalled();
  });

  test('another one', () => {
    console.warn('warn hello world');
    expect(myfuncSpy).not.toHaveBeenCalled();
  });


  test('third one', () => {
    // not calling so not been called
    expect(myfuncSpy).not.toHaveBeenCalled();
  });
});

test('real console', () => {
  console.log('==> should be displayed on console');
});
```

feloy:
> but we recommend to use resetAllMocks to have clean state on every test, so still I don't see why we would need mockRestore ?
> 
> it'll work as expected:
> 
> ```
> describe('spy on myfunc', () => {
>   let myfuncSpy: MockInstance<() => void>;
> 
>   beforeEach(() => {
>     vi.resetAllMocks();
>     myfuncSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
>   });
> 
>   test('use console.log', () => {
>     console.log('hello world');
>     expect(myfuncSpy).toHaveBeenCalled();
>   });
> 
>   test('another one', () => {
>     console.warn('warn hello world');
>     expect(myfuncSpy).not.toHaveBeenCalled();
>   });
> 
> 
>   test('third one', () => {
>     // not calling so not been called
>     expect(myfuncSpy).not.toHaveBeenCalled();
>   });
> });
> 
> test('real console', () => {
>   console.log('==> should be displayed on console');
> });
> ```

No, I cannot see it working, it does not display anything in the console for the `real console` test.

This needs 2 changes:

- use `restoreAllMocks` instead of `resetAllMocks` (1)
- call it from `afterEach`

```ts
describe('spy on myfunc', () => {
  let myfuncSpy: MockInstance<() => void>;

  beforeEach(() => {
    myfuncSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
  });

  afterEach(() => {
    vi.restoreAllMocks();
  });

  test('use console.log', () => {
    console.log('hello world');
    expect(myfuncSpy).toHaveBeenCalled();
  });

  test('another one', () => {
    console.warn('warn hello world');
    expect(myfuncSpy).not.toHaveBeenCalled();
  });


  test('third one', () => {
    // not calling so not been called
    expect(myfuncSpy).not.toHaveBeenCalled();
  });
});

test('real console', () => {
  console.log('==> should be displayed on console');
});
```

(1) https://vitest.dev/api/vi#vi-restoreallmocks
```
vi.resetAllMocks
Will call [.mockReset()](https://vitest.dev/api/mock#mockreset) on all spies. This will clear mock history and reset its implementation to an empty function (will return undefined).

vi.restoreAllMocks
Will call [.mockRestore()](https://vitest.dev/api/mock#mockrestore) on all spies. This will clear mock history and reset its implementation to the original one.
```

benoitf:
I think originally we were doing

```
  vi.resetAllMocks();
  vi.restoreAllMocks(); 
```

~~but @dgolovin said we should not do restore as it was part of reset~~

so now it seems we should have stick to the original rule of doing both ? 🙄

edited: 

seems he complained about

```
vi.resetAllMocks();
vi.clearAllMocks();
```

feloy:
> I think originally we were doing
> 
> ```
>   vi.resetAllMocks();
>   vi.restoreAllMocks(); 
> ```
> 
> ~but @dgolovin said we should not do restore as it was part of reset~
> 
> so now it seems we should have stick to the original rule of doing both ? 🙄
> 
> edited:
> 
> seems he complained about
> 
> ```
> vi.resetAllMocks();
> vi.clearAllMocks();
> ```

I find the response of this question very instructive: https://stackoverflow.com/questions/58151010/difference-between-resetallmocks-resetmodules-resetmoduleregistry-restoreallm

tl;dr:
```
clearAllMocks ⊂ resetAllMocks 
clearAllMocks ⊂ restoreAllMocks
```

clear: reset only counters
reset: set mocks to noop function (useful for vi.fn().mock...())
restore: set mocks to original function (useful for vi.spyOn().mock...())

Because of the difference between `resetAllMocks` and `restoreAllMocks`, it is dangerous to call one of them (or both) in a `beforeEach` if the tests are mocking both `spies` and `vi.fn()`, as they are targeting each of them.

I will elaborate a little bit more this code guideline, to explain to `restore` spies individually, and to `reset` mocks made on `fi.fn()` individually

benoitf:
so should we do both `resetAllMocks` and `restoreAllMocks` in all beforeEach methods to be sure that all tests are independent and are not interfering each other

feloy:
> so should we do both `resetAllMocks` and `restoreAllMocks` in all beforeEach methods to be sure that all tests are independent and are not interfering each other

I suppose the latest one will take precedence: if you call `reset` after `restore`, all `spies` and `vi.fn()` mocks will be eventually reset, but we may want to restore the spies.

This makes me understand that there are several effects in the PR https://github.com/containers/podman-desktop-extension-ai-lab/pull/2326 around replacing `reset` by `restore`:

In this PR, if you restore individually all mocked spies, but still call `resetAllMocks` in the top-level `beforeEach`, this will fail, as the original functions will first be restored individually with individual `myspy.mockRestore()` , but will be reset to `noop` when calling `resetAllMocks` in the `beforeEach`. The thing to do would be to restore all spies individually, then only call `clearAllMocks` in `beforeEach`

CODE-GUIDELINES.md
```diff
@@ -37,3 +37,97 @@ test('...', () => {
   vi.mocked(window.windowMethod).mockResolvedValue('a string');
 });
 ```
+
+### Mock complete modules, spy on parts of module for specific tests
```
jeffmaury, line 41: Does it apply also to internal modules ?

feloy:
I realize that my second example (for spies) is not really realistic, as the spy is created in an object re-created for each test, which makes restoring the spy not necessary. I'll change the example

CODE-GUIDELINES.md
feloy: I would not like to enforce this point, it would be to the developer to choose the limit of what is mocked/unit tested. I'll develop a little bit more to clarify this point

feloy:
> I realize that my second example (for spies) is not really realistic, as the spy is created in an object re-created for each test, which makes restoring the spy not necessary. I'll change the example

I've updated the example, and added some explanations concerning the choice between mock vs spyOn

feloy:
@SoniaSandler @benoitf @axel7083 @dgolovin PR ready for review

benoitf (commented):
for the first block/part I see no issue

but for the second one, should we not like have

restoreAllMocks() and resetAllMocks() in all beforeEach instead of not forgetting to do an afterEach for each method we use spyOn

my point it that it'll be easily forgotten and I don't see how to enforce the rule of always have an afterEach for each spyOn

benoitf (approved):

PR #10545, opened 2025-01-07
docs: code guideline to run mocks inside beforeEach

feloy:

### What does this PR do?

Adds a code guideline to run mocks inside beforeEach

### Screenshot / video of UI

N/A

Part of #9821

### How to test this PR?

N/A

### DIFF
--- a/CODE-GUIDELINES.md
+++ b/CODE-GUIDELINES.md
@@ -40,8 +40,10 @@ test('...', () => {
 
 ### Mock before tests, not during tests
 
-If you need to mock functions for some tests, execute the mock from the `beforeEach` callback inside a `describe`,
-so the environment is clearly defined in the `describe` and the `test` is only executing the tests:
+When you mock functions for some tests, it is recommended to define the mocks from the `beforeEach` callback inside a `describe`,
+so the environment created by the mocks is clearly defined in the `describe`, and the `test` is only executing the tests:
+
+You also need to reset (or restore) the mocks from the `afterEach` callback, so these mocks are not used for other tests.
 
 Do not write:
 
@@ -64,11 +66,20 @@ test('second behaviour', () => {
 Instead, write:
 
 ```ts
+import { existsSync, PathLike } from 'node:fs';
+import { afterEach, beforeEach, describe, expect, MockedFunction, test, vi } from 'vitest';
+
 vi.mock('node:fs');
 
 describe('the file exists', () => {
+  let existsSyncMock: MockedFunction<(path: PathLike) => boolean>;
+
   beforeEach(() => {
-    vi.mocked(fs.existsSync).mockReturnValue(true);
+    existsSyncMock = vi.mocked(existsSync).mockReturnValue(true);
+  });
+
+  afterEach(() => {
+    existsSyncMock.mockReset();
   });
 
   test('first behaviour', () => {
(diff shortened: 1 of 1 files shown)


CODE-GUIDELINES.md
```diff
@@ -37,3 +37,59 @@ test('...', () => {
...
+Instead, write:
+
+```ts
+import { existsSync, PathLike } from 'node:fs';
+import { afterEach, beforeEach, describe, expect, MockedFunction, test, vi } from 'vitest';
+
+vi.mock('node:fs');
+
+describe('the file exists', () => {
+  let existsSyncMock: MockedFunction<(path: PathLike) => boolean>;
```
axel7083, line 75: We should not try to declare the mock, we should access them by doing `vi.mocked()` or directly `expect(existsSync)`

CODE-GUIDELINES.md
```diff
@@ -37,3 +37,59 @@ test('...', () => {
   vi.mocked(window.windowMethod).mockResolvedValue('a string');
 });
 ```
+
+### Mock before tests, not during tests
+
+When you mock functions for some tests, it is recommended to define the mocks from the `beforeEach` callback inside a `describe`,
+so the environment created by the mocks is clearly defined in the `describe`, and the `test` is only executing the tests:
+
+You also need to reset (or restore) the mocks from the `afterEach` callback, so these mocks are not used for other tests.
```
axel7083, line 46: I don't understand why we would need to do that, when we almost never had to. Having vi.resetAllMock() should be enough in the beforeEach.

CODE-GUIDELINES.md
```diff
@@ -37,3 +37,59 @@ test('...', () => {
   vi.mocked(window.windowMethod).mockResolvedValue('a string');
 });
 ```
+
+### Mock before tests, not during tests
+
+When you mock functions for some tests, it is recommended to define the mocks from the `beforeEach` callback inside a `describe`,
```
axel7083, line 43: It really depends on the test, if all the tests share the same mocked value, yes, if not, it is easier for a reader to see the specific mocked value inside the test

CODE-GUIDELINES.md
feloy: see discussion https://github.com/podman-desktop/podman-desktop/pull/10541

CODE-GUIDELINES.md
```diff
@@ -37,3 +37,59 @@ test('...', () => {
...
+  vi.mocked(fs.existsSync).mockReturnValue(true);
+  codeToTest();
+  expect(...);
+});
+
+test('second behaviour', () => {
+  vi.mocked(fs.existsSync).mockReturnValue(true);
+  otherCodeToTest();
+  expect(...);
+});
```
axel7083, line 63: I would argue that the following is better

```ts
import { existsSync } from 'node:fs';

vi.mock('node:fs', () => ({
  existsSync: vi.fn(),
}));

beforeEach(() => {
  vi.resetAllMocks();
  vi.mocked(existsSync).mockReturnValue(true);
});

test('check something', () => {
  // ...call method using existsSync
   
  expect(existsSync).toHaveBeenCalledWith('/path/');
})
````

feloy: >   vi.resetAllMocks();

This will be problematic if you have defined spies, see discussion on https://github.com/podman-desktop/podman-desktop/pull/10541

feloy:
closing for now

PR #10557, opened 2025-01-08
chore: change order of welcome page for podman

### LINKED ISSUE #10244: Change order of onboarding steps

Currently we have Podman as the last option during the onboarding: 
Image

I suggest we reverse the order, so `Podman, Kubernetes, Compose`. This is to make it clear that `Podman` is the most important extension to be installed

Reverse the order, so `Podman, Kubernetes, Compose`

_No response_

### Additional context

_No response_

cdrage:
chore: change order of welcome page for podman

### What does this PR do?

* Adds a function that will re-sort the providers and make sure that
  Podman is first on the Welcome Page

Closes https://github.com/podman-desktop/podman-desktop/issues/10244

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/welcome/WelcomePage.svelte
+++ b/packages/renderer/src/lib/welcome/WelcomePage.svelte
@@ -31,6 +31,8 @@ interface OnboardingInfoWithSelected extends OnboardingInfo {
 
 let onboardingProviders: OnboardingInfoWithSelected[] = [];
 
+const podmanProviderName = 'podman';
+
 onMount(async () => {
   const ver = await welcomeUtils.getVersion();
   if (!ver) {
@@ -46,13 +48,16 @@ onMount(async () => {
   podmanDesktopVersion = await window.getPodmanDesktopVersion();
 
   onboardingsUnsubscribe = onboardingList.subscribe(value => {
-    // Add "selected" property to each provider and add to onboardingEnabledProviders
-    onboardingProviders = value.map(provider => {
+    // Add "selected" property to each provider
+    const sortedProviders = value.map(provider => {
       return {
         ...provider,
         selected: true,
       };
     });
+
+    // Sort providers by podman name so it appears first
+    onboardingProviders = sortByPodmanName(sortedProviders);
   });
 });
 
@@ -70,6 +75,12 @@ async function closeWelcome() {
   if (showTelemetry) {
     await welcomeUtils.setTelemetry(telemetry);
   }
+}
+
+// Function to sort providers by "Podman" name, make sure that a provider with the name "podman" is always first
+function sortByPodmanName(providers: OnboardingInfoWithSelected[]): OnboardingInfoWithSelected[] {
+  // We slice to avoid mutating the original array
+  return providers.slice().sort((a, b) => (a.name === podmanProviderName ? -1 : b.name === podmanProviderName ? 1 : 0));
 }
 
 // Function to toggle provider selection
--- a/packages/renderer/src/lib/welcome/WelcomePage.spec.ts
+++ b/packages/renderer/src/lib/welcome/WelcomePage.spec.ts
@@ -177,3 +177,73 @@ test('Expect welcome screen to show three checked onboarding providers', async (
   expect(checkbox3).toBeInTheDocument();
   expect(checkbox3).toBeChecked();
 });
+
+test('Make sure the provider with name podman appears first even if its 2nd in the list', async () => {
+  onboardingList.set([
+    {
+      extension: 'id',
+      title: 'onboarding',
+      name: 'foobar1',
+      displayName: 'FooBar1',
+      icon: 'data:image/png;base64,foobar1',
+      steps: [
+        {
+          id: 'step',
+          title: 'step',
+          state: 'failed',
+          completionEvents: [],
+        },
+      ],
+      enablement: 'true',
+    },
+    {
+      extension: 'id',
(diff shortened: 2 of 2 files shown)


packages/renderer/src/lib/welcome/WelcomePage.svelte
```diff
@@ -72,6 +77,12 @@ async function closeWelcome() {
   }
 }
 
+// Function to sort providers by "Podman" name, make sure that a provider with the name "podman" is always first
+function sortByPodmanName(providers: OnboardingInfoWithSelected[]): OnboardingInfoWithSelected[] {
+  // We slice to avoid mutating the original array
+  return providers.slice().sort((a, b) => (a.name === podmanProviderName ? -1 : b.name === podmanProviderName ? 1 : 0));
```
axel7083, line 83: From the official documentation of [Array.prototype.splice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)

> The splice() method of [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) instances changes the contents of an array by removing or replacing existing elements and/or adding new elements [in place](https://en.wikipedia.org/wiki/In-place_algorithm).

> To create a new array with a segment removed and/or replaced without mutating the original array, use [toSpliced()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSpliced). To access part of an array without modifying it, see [slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice).

cdrage: "To access part of an array without modifying it, see [slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)."

= That's why I use slice? When I would do providers.sort() it gave me a linting error (no mutating arrays)

benoitf (changes requested):
the core should not have any dependency to an extension

packages/renderer/src/lib/welcome/WelcomePage.svelte
```diff
@@ -46,13 +48,16 @@ onMount(async () => {
   podmanDesktopVersion = await window.getPodmanDesktopVersion();
 
   onboardingsUnsubscribe = onboardingList.subscribe(value => {
-    // Add "selected" property to each provider and add to onboardingEnabledProviders
-    onboardingProviders = value.map(provider => {
+    // Add "selected" property to each provider
+    const sortedProviders = value.map(provider => {
```
axel7083, line 52: Maybe we should not define a custom function sortByPodmanName, and instead create a function that can be used directly with .sort.

```ts
value.map(...).sort(sortByPodmanName)

function sortByPodmanName(a: OnboardingInfoWithSelected, b: OnboardingInfoWithSelected): number {
  // handle logic of prioritising podman
  // do not forget to properly sort the others so we have a deterministic sorting
  // Example using a.name.localeCompare(b.name) if neither a or b is podman
 }
```

packages/renderer/src/lib/welcome/WelcomePage.svelte
axel7083: There is the [Array.prototype.toSorted()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) if there is a need to avoid mutating the original array.

axel7083: I would still prefer https://github.com/podman-desktop/podman-desktop/pull/10557/files#r1907257842

axel7083: > "To access part of an array without modifying it, see [slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)."

Sorry I am confusing `slice` and `splice`.

> = That's why I use slice? When I would do providers.sort() it gave me a linting error (no mutating arrays)

Yes, better to do https://github.com/podman-desktop/podman-desktop/pull/10557#discussion_r1907257842

cdrage:
> the core should not have any dependency to an extension

I agree.

What would you suggest how we should change this?

The issue assigned is requesting Podman be first, but how would we do that without changing the core?

benoitf:
>> the core should not have any dependency to an extension

> I agree.
> What would you suggest how we should change this?
> The issue assigned is requesting Podman be first, but how would we do that without changing the core?


it's to sort from a feature being provided, not from its name

for example we leverage first the onboarding of extensions that provides a container engine before the extensions providing just a cli tool or kubernetes cluster, etc.

cdrage:
> > > the core should not have any dependency to an extension
> 
> > I agree.
> > What would you suggest how we should change this?
> > The issue assigned is requesting Podman be first, but how would we do that without changing the core?
> 
> it's to sort from a feature being provided, not from its name
> 
> for example we leverage first the onboarding of extensions that provides a container engine before the extensions providing just a cli tool or kubernetes cluster, etc.

Ah, so instead of by name (podman), let's sort and have it so that container engine's are first before CLI tools? Check each provider if it's going to provide an engine vs non-container-engine, etc.

benoitf:
or for now just provider engine first and the others

but yes the idea to filter by a category/feature not hardcode a name

cdrage:
Closing and will re-open with different PR

PR #10562, opened 2025-01-08
chore: change order of welcome page to prioritize container engines

### LINKED ISSUE #10244: Change order of onboarding steps

cdrage:
chore: change order of welcome page to prioritize container engines

### What does this PR do?

Re-orders the welcome page to prioritize extensions with container
engines over CLI extensions:
* Retrieves information from providerInfos store
* Gets the container engine information
* Sorts by container engine

Closes https://github.com/podman-desktop/podman-desktop/issues/10244

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/welcome/WelcomePage.spec.ts
+++ b/packages/renderer/src/lib/welcome/WelcomePage.spec.ts
@@ -28,6 +28,8 @@ import { get } from 'svelte/store';
 import { beforeAll, expect, test, vi } from 'vitest';
 
 import { onboardingList } from '/@/stores/onboarding';
+import { providerInfos } from '/@/stores/providers';
+import type { ProviderInfo } from '/@api/provider-info';
 
 import WelcomePage from './WelcomePage.svelte';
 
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/welcome/WelcomePage.svelte
```diff
@@ -21,15 +22,36 @@ let telemetry = true;
...
+
+// Using providerInfos as well as the information we have from onboarding,
+// we will by default auto-select as well as add containerEngine to the list as true/false
+// so we can make sure that extensions with container engines are listed first
+$: onboardingProviders = $onboardingList
+  .map(provider => {
+    const hasContainerConnection = providersWithContainerConnections.some(
+      connectionProvider => connectionProvider.extensionId === provider.extension,
+    );
+    return {
```
axel7083, line 43: On linux fedora it is not sorted :/

image


I decided to add
```ts
console.log(`provider ${provider.extension} hasContainerConnection ? ${hasContainerConnection}`)
``` 
and here is the results

image

packages/renderer/src/lib/welcome/WelcomePage.spec.ts
```diff
@@ -177,3 +179,92 @@ test('Expect welcome screen to show three checked onboarding providers', async (
...
+
+  // wait until the onboarding list is populated
+  while (get(onboardingList).length === 0) {
+    await new Promise(resolve => setTimeout(resolve, 100));
+  }
+
+  await waitRender({ showWelcome: true });
+
+  // wait until aria-label 'providerList' is populated
+  while (screen.queryAllByLabelText('providerList').length === 0) {
```
axel7083, line 262:
```suggestion
  await vi.waitFor(() => {
    expect(screen.queryAllByLabelText('providerList').length).toBeGreaterThan(0);
  });
```

We should avoid using while loop

packages/renderer/src/lib/welcome/WelcomePage.spec.ts
```diff
@@ -177,3 +179,92 @@ test('Expect welcome screen to show three checked onboarding providers', async (
...
+    await new Promise(resolve => setTimeout(resolve, 100));
+  }
```
axel7083:
```suggestion
  await vi.waitFor(() => {
    expect(screen.queryAllByLabelText('providerList').length).toBeGreaterThan(0);
  });
```

We should avoid using while loop

packages/renderer/src/lib/welcome/WelcomePage.svelte
cdrage: > On linux fedora it is not sorted :/
> 
> image
> 
> I decided to add
> 
> ```ts
> console.log(`provider ${provider.extension} hasContainerConnection ? ${hasContainerConnection}`)
> ```
> 
> and here is the results
> 
> image

Thank you, on Windows & MacOS it was showing as true / sorting.

cdrage: @benoitf Suggestions what to use?

Really odd hasContainerConnection is showing as false for Linux. Makes sense I guess since it's native.

I have the following from Podman for example:
Screenshot 2025-01-08 at 11 41 07 AM

gastoner (approved):
LGTM on Windows
image

packages/renderer/src/lib/welcome/WelcomePage.svelte
benoitf: @cdrage filter on containerConnections length > 0 and not containerConnectionCreation

because we want to check if the provider has registered some socket API, not if the provider is able to create a 'vm/machine'

cdrage:
@axel7083 mind doing another review? Updated to use containerConnections but unable to get my Linux machine running right now.

Just tested on mac and windows and it works :)

axel7083 (approved):
LGTM nicely sorted !

image

PR #10568, opened 2025-01-08
feat: make build task cancellable

### LINKED ISSUE #10080: No way to cancel build, start another one
### Bug description

I started a build using the GPU test here [1] and it appeared to hang - well over four minutes with no feedback in the UI. It was actually caused by #10077, but I didn't know at the time and this issue is for the usability during a hung or extended build, not the specific cause.

- If you leave the page and come back the console output has disappeared.
- No progress, additional console output, or anything to confirm if it's actually still running.
- Cancel button doesn't work.
- Closing and reopening the form brings you back to the same build instead of allowing you to start a new one.

[1] https://podman-desktop.io/docs/podman/gpu#verification

Image

### Operating system

macOS 15.1

### Installation Method

None

### Version

1.14.1

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

axel7083:
### What does this PR do?

Reported in https://github.com/podman-desktop/podman-desktop/issues/10080, the build task is not cancellable from the task manager. 

Since a build is cancellable from the page, we uses a cancellation token, we can just check for cancellation token, and if it exists, make the task cancellable.

### Screenshot / video of UI

image

image

Fixes https://github.com/podman-desktop/podman-desktop/issues/10080 (maybe ?)

### How to test this PR?

> Tests were a bit hard to do :cry: 

You can try by creating a Containerfile such as

```Dockerfile
FROM docker.io/library/alpine:latest

RUN sleep 150000
```

And use it in the build page.

### DIFF
--- a/packages/main/src/plugin/index.spec.ts
+++ b/packages/main/src/plugin/index.spec.ts
@@ -20,10 +20,11 @@
 import { EventEmitter } from 'node:events';
 import { tmpdir } from 'node:os';
 
-import type { WebContents } from 'electron';
+import type { IpcMainInvokeEvent, WebContents } from 'electron';
 import { app, BrowserWindow, clipboard, ipcMain, shell } from 'electron';
-import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
+import { afterEach, beforeEach, expect, test, vi } from 'vitest';
 
+import type { Task } from '/@/plugin/tasks/tasks.js';
 import { Updater } from '/@/plugin/updater.js';
 import type { NotificationCardOptions } from '/@api/notification.js';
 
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/index.spec.ts
```diff
@@ -59,26 +78,68 @@ webContents.send = vi.fn();
...
+  isDestroyed: vi.fn(),
+  webContents: {
+    on: vi.fn(),
+  },
+} as unknown as BrowserWindow;
+
 beforeEach(() => {
   vi.clearAllMocks();
+
+  vi.mocked(BrowserWindow.getAllWindows).mockReturnValue([MAIN_WINDOW_MOCK]);
```
benoitf: issue: I think this pattern is not expected. I have never seen it either in official vitest documentation or in our tests.

it's really hard to then understand what is being done, as the mock declared in vi.mock is not used at all.

axel7083: The mock in vi.mock is needed otherwise when accessing `BrowserWindow.getAllWindows` we would get a `not a mocked object`.

I feel like there is nothing new here ? How do you mock for all the tests the return value of a function ? Here I want `getAllWindows` to return the MAIN_WINDOW_MOCK for all the tests. If I do the following

```
vi.mock('electron', () => {
  return {
    BrowserWindow: {
      getAllWindows: vi.fn().mockReturnValue([MAIN_WINDOW_MOCK]),
    },
});
```

The `vi.clearAllMocks()` will remove the mock return value for my tests. So I need to mock the return value in the beforeEach.

benoitf: there is this one and these :
```
    vi.mocked(TaskManager).mockReturnValue(TASK_MANAGER_MOCK);
    vi.mocked(ContainerProviderRegistry).mockReturnValue(CONTAINER_PROVIDER_REGISTRY);
    vi.mocked(CONTAINER_PROVIDER_REGISTRY.buildImage).mockResolvedValue(undefined);
```

which is also related to https://github.com/containers/podman-desktop-extension-ai-lab/pull/2345

it looks like you're pushing a new pattern which is not documented or widely used.

axel7083: Removed all non-standard way of testing (masking as resolved as I changed all the tests)

axel7083:
Too much conflict in `packages/main/src/plugin/index.spec.ts`

Issue #9978, opened 2024-11-19
Implement design for the create container from image workflow


benoitf:
### Is your enhancement related to a problem? Please describe

Following the design of https://github.com/podman-desktop/podman-desktop/issues/9977, need to implement it

### Describe the solution you'd like

Implement the mockup to run a container from an image (without having first the image being pulled)

PR #10573, opened 2025-01-08
chore: update typeahead to support headings

SoniaSandler:

### What does this PR do?
This PR updates the UI `typeahead` component to support multiple search functions and heading for the different search functions results
Will be used in https://github.com/podman-desktop/podman-desktop/issues/9978

### Screenshot / video of UI
Screenshot from 2025-01-08 13-27-32

Related to https://github.com/podman-desktop/podman-desktop/issues/9978

### How to test this PR?
1. Go to `PullImage.svelte`
2. Optional: add additional values to `searchResult` with different groups/headings
3. Make sure that the headings (if you provided them) are added in the right places in the rendered UI, and the values are correctly sorted by their groups


SoniaSandler:
In case there are no results between two or more headings, there are a couple of options for the design - displaying all headings or just the ones that have results. 
Which option looks better?:
Screenshot from 2025-01-08 13-55-11
Screenshot from 2025-01-08 13-55-21

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -66,6 +69,7 @@ function onKeyDown(e: KeyboardEvent): void {
       onEnterKey(e);
       break;
   }
+  onChange(value);
```
gastoner: Is this move required?

gastoner (commented):
https://github.com/user-attachments/assets/47c9d122-6d05-4d8f-bfc2-204b6ec2cbd5


Is this correct? 
I dont se the disabled button regardless of what I type for some reason :( (Or if is the heading button just to show that there are multiple "search results"? )

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -12,8 +12,10 @@ export let initialFocus: boolean = false;
 export let id: string | undefined = undefined;
 export let name: string | undefined = undefined;
 export let error: boolean = false;
+// the headings will be shown in the provided order: 1st heading - 1st function result, 2nd heading - 2nd function result, ...
+export let headings: string[] = [];
 
-export let searchFunction: SearchFunction = async (_s: string) => [];
+export let searchFunctions: SearchFunction[] = [async (_s: string) => []];
```
feloy: As a user, I'm not very confident when I have to fill several correlated arrays. There is always a risk there are not aligned.

I would prefer to have a single array of a structure `{ heading, searchFunction }`

packages/renderer/src/lib/ui/Typeahead.svelte
SoniaSandler: Since the value changes only inside the case, having the `onChange` before them will apply it to the "old value". Moving it after the cases (after the value has been updated to the new one) ensures that the `onChange` uses the new value.
Before this, the `onChange` would always be one step behind in terms of the values it uses (unless that was intentional?)

SoniaSandler:
> Is this correct? I dont se the disabled button regardless of what I type for some reason :( (Or if is the heading button just to show that there are multiple "search results"? )

@gastoner no changes have been made to `PullImage` typeahead values, so you shouldn't see any visual changes, but I've added how to test this PR in the description now if you'd like check the visual changes

packages/renderer/src/lib/ui/Typeahead.svelte
axel7083: I am wondering if it would not be better if we have a single searchFunction, and it would return `Array<{ value: string, group?: string }>`

axel7083: Moreover, I am wondering, if the `Typeahead` should not have a searchFunctions, and instead have a `onChange()` property, and the parent component should be responsible of doing the dirty work.

Then change items property from `Array<string>` to `Array<{value: string, group?: string }>`.

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -141,35 +145,55 @@ function makeVisible(): void {
...
-          return 1;
+  for (const [index, searchFunction] of searchFunctions.entries()) {
+    searchFunction(value)
+      .then(result => {
+        // if the component has been disabled in the meantime
+        if (disabled) {
+          return;
+        }
+        searchResults[index] = result.toSorted((a: string, b: string) => {
+          const dockerIoValue = `docker.io/${value}`;
```
axel7083: +1 for https://github.com/podman-desktop/podman-desktop/pull/10573/files#r1908966973 as we should not have pullImage specific code in the Typeahead component

axel7083 (commented):
My general feedback is that the Typeahead component should not be that much complex, it should have a simple, and easy interface.

If we need complex behavior the parent should be responsible.

My opinion is to remove the search from the typeahead, change the props to something like

```ts
interface Props {
    value: string;
    items: Array<{ id: string, label: string, group?: string}>;
    onInputChange(input: string): void;
    onSelect(itemId: string): void;
    loading?: boolean;
    disable?: boolean;
    ...
}
```

So we can have a flexible component, make the search logic outside the component; and simplify it, so if we it to reuse it, we can. 


> Edit: here is an example of a component that have similar behavior https://svelte-select-examples.vercel.app/examples/props/group-header-selectable their props are interesting to look at as inspiration.

SoniaSandler:
@axel7083 I've updated the PR according to some of your suggestions (probably not all of them), generalized `Typeahead` and migrated it to svelte5. lmk what do you think.

packages/renderer/src/lib/image/PullImage.svelte
```diff
@@ -244,6 +244,22 @@ function checkIfTagExist(image: string, tags: string[]): void {
 
   isValidName = tags.some(t => t === tag);
 }
+
+async function searchFunction(value: string): Promise<{ values: string[]; group?: string; sorted?: boolean }[]> {
+  const result = (await searchImages(value)).toSorted((a: string, b: string) => {
+    const dockerIoValue = `docker.io/${value}`;
+    const aStartsWithValue = a.startsWith(value) || a.startsWith(dockerIoValue);
+    const bStartsWithValue = b.startsWith(value) || b.startsWith(dockerIoValue);
+    if ((aStartsWithValue && bStartsWithValue) || (!aStartsWithValue && !bStartsWithValue)) {
```
gastoner: `aStartsWithValue === bStartsWithValue` ?

gastoner (approved):
LGTM

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -1,33 +1,55 @@
...
+  delay = 200,
+  disabled = false,
+  initialFocus = false,
+  id,
+  name,
+  error = false,
+  className,
+  onInputChange = async (_s: string) => [],
+  onChange = function (_s: string) {},
+  onEnter = function () {},
```
axel7083: Probably better to keep them undefined, and use conditional calling.

E.g.

```ts
onChange: (() => void) | undefined;

...
// Conditional calling
onChange?.();
```

axel7083: This can be a future PR, don't feel forced to include it, as this was already the case before, I don't mind merging it.

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -1,33 +1,55 @@
...
+  placeholder?: string;
+  required?: boolean;
+  delay?: number;
+  disabled?: boolean;
+  initialFocus?: boolean;
+  id?: string;
+  name?: string;
+  error?: boolean;
+  className?: string;
+  onInputChange?: (s: string) => Promise<GroupItem[]>;
```
axel7083: It may feel easy to return from the onInputChange, but it is easier to split this in two.

```suggestion
  onInputChange?: (s: string) => void;
  values: GroupItem[]
```

So if we need to define initial value or change it based on external events, we can just change the values props without having to wait for an InputEvent.

The parent component should handle the logic of correlating the input and the values, this is a bit more work for it, as it need to have a `values = $state<GroupItem[]>(..)` but this is more svelte like

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -142,30 +164,44 @@ function makeVisible(): void {
...
-        const dockerIoValue = `docker.io/${value}`;
-        const aStartsWithValue = a.startsWith(value) || a.startsWith(dockerIoValue);
-        const bStartsWithValue = b.startsWith(value) || b.startsWith(dockerIoValue);
-        if ((aStartsWithValue && bStartsWithValue) || (!aStartsWithValue && !bStartsWithValue)) {
-          return a.localeCompare(b);
-        } else if (aStartsWithValue && !bStartsWithValue) {
-          return -1;
-        } else {
-          return 1;
+      for (let { values, group, sorted } of result) {
```
axel7083: Following https://github.com/podman-desktop/podman-desktop/pull/10573/files#r1912902298 you should be able to use a derived state for the `groupedValues` (or better naming to be found) instead of having logic in the processInput

packages/renderer/src/lib/ui/Typeahead.ts
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface GroupItem {
+  values: string[];
+  group?: string;
+  sorted?: boolean;
```
axel7083: Why do you need to expose sorted ?

axel7083: if this is something internal to Typeahead, this should not be exposed to the parent, and should be internal.

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -1,51 +1,101 @@
...
+  placeholder,
+  required = false,
+  delay = 200,
+  disabled = false,
+  initialFocus = false,
+  id,
+  name,
+  error = false,
+  className,
+  groupValues = [],
```
axel7083: I feel like for the parent, using `items` is easier to understand the usage ?

packages/renderer/src/lib/ui/Typeahead.ts
SoniaSandler: @axel7083 my intention with this sorted property is to check if we need to do the default sort in the Typeahead or not. For example, in the `pullImage` page, there is a custom sort with the docker prefix, so in this case we don't want typeahead to do its own sorting on the values coming from pullImage.
For my other issue, I'm not sorting the local images in the parent component, so Typeahead can do its default sorting, but for the registry images, I'm using the custom sorting with the docker prefix (taken from `pullImage`).
Another option we could do is to remove also the default sorting from `Typeahead` and leave this to the parent components, so if the values coming from there are not sorted, they won't be sorted in `Typeahead`.

benoitf (commented):
question: is sorting should be part of the component ?

making only displaying the content being provided would make it easier

why should it sort data ?

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -10,8 +12,7 @@ interface Props {
   id?: string;
   name?: string;
   error?: boolean;
-  resultItems?: string[];
-  sort?: boolean;
+  resultItems?: GroupItem[];
```
axel7083: we should not expose a `resultItems` this is confusing, 

Sorry for the late feedback, but GroupItem should be the other way around

We should expose something like the following

```ts
interface Item {
     label: string;
     value: string;
     group?: string;
}
```

The parent should not be responsible of doing the grouping

packages/renderer/src/lib/ui/Typeahead.spec.ts
```diff
@@ -273,3 +271,53 @@ test('should show error border', async () => {
   expect(parentInput).toHaveClass('focus-within:border-[var(--pd-input-field-stroke-error)]');
   expect(parentInput).not.toHaveClass('hover:border-b-[var(--pd-input-field-hover-stroke)]');
 });
+
+test('should include heading based on given order and searchFunctions order', async () => {
+  let searchResult: GroupItem[] = [];
+  const searchFunction = async (s: string): Promise<void> => {
+    await new Promise(resolve => setTimeout(resolve, 100));
```
axel7083: Please use `vi.waitFor`, we should avoid using setTimeout in tests

packages/renderer/src/lib/ui/Typeahead.ts
axel7083: The sorted property should be remove as it should not be exposed, instead the Typeahead method should expose a `compare: (a: Item, b: Item): number` optional prop that the parent component can provide.

By default the Typeahead should do a basic string sorting, that can be overwritten by the parent through the compare method.

SoniaSandler:
@benoitf this was some sort of enhancement to the existing Typeahead component that already include sorting in it for `PullImage`. Not wanting to change it too much from its original design, I moved the specific sorting to the `PullImage` component and adjusted it to be more generic in `Typeahead`. I have no problem removing it if it would be better to just display the given values the way they were passed to the component.

packages/renderer/src/lib/ui/Typeahead.svelte
SoniaSandler: @axel7083 I'm not sure I understand why it would be better to do the grouping in `Typeahead`?
By having separate items, with each having one value and group, we would need to go through all of them in the parent component to assign them a group and then go through all of them again in `Typeahead` to group them.
The way it is now, we assign one group only once and we don't need to iterate through all the items to group them in `Typeahead`.

benoitf:
for now the component is not part of the UI library so I'm fine with any change

I would opt for something being simple, component doing only the rendering logic. An don my side I would prefer any logic/complex task being deferred to the component using it

IMHO only once used by many locations we could start to move a common logic

packages/renderer/src/lib/ui/Typeahead.svelte
axel7083: It is easier to have an array of item with properties than an array of GroupItem with many items per group.  This make it more similar to the Dropdown, which it should have a very similar interface.

> The way it is now, we assign one group only once and we don't need to iterate through all the items to group them in Typeahead.

The group should be optional, and by forcing the parent to do the grouping we force it to do the grouping.

axel7083 (approved):
Amazing work! LGTM 🚀 

> maybe the title of this PR need to change to remove the search functions part

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -28,37 +30,47 @@ let {
...
+      if (headings[currentItems.length]) {
+        headings[currentItems.length].push(group);
+      } else {
+        headings[currentItems.length] = [group];
+      }
+    }
+    currentItems = currentItems.concat(groupedItems[group]);
+  }
+  items = currentItems;
+  itemHeadings = headings;
```
benoitf: IHMO it can be two derived state

groupItems = one derived, currentItems another one, iteamHeadings another one

benoitf (commented):
overall LGTM but improvements

note: I would have split the enhancement in two

1st: move the boolean sort to a compare function
2nd: introduce headings

b/c here it says that it's introducing headings but it's changing also an existing behavior so it's hard to track which change is related to what.

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -67,6 +79,28 @@ function onItemSelected(s: string): void {
   onChange?.(s);
 }
 
+function groupItems(itemsList: TypeaheadItem[]): { [group: string]: string[] } {
```
benoitf: should be turned into a $derived rather than having an effect function calling a groupItems function

packages/renderer/src/lib/ui/Typeahead.svelte
SoniaSandler: @benoitf I'll update groupedItems to a derived state, but for items and itemHeadings, they kind of go together here.
If I'll do derived state for each one of them, it will require two iterations over groupedItems, which might not be a siginificant change, but would it be better or it doesn't matter?

benoitf: on my end I'm fine with two iterations because then it's easy to read. Usually I prefer readability, easiness over potential optimization.

I would optimize on specific cases and where the time is being spent in Podman Desktop.

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -40,26 +42,66 @@ let input: HTMLInputElement | undefined = $state();
...
+    return [];
+  }
+  let currentItems: string[] = [];
+  for (const group in groupedItems) {
+    currentItems = currentItems.concat(groupedItems[group]);
+  }
+  return currentItems;
+});
+
+function groupItems(itemsList: TypeaheadItem[]): { [group: string]: string[] } {
```
gastoner: Nit: create an interface groupedItems for `{ [group: string]: string[] }` ? And use it here and in `updateHeadings` function?

gastoner (approved):
LGTM, tested on Windows

SoniaSandler:
@axel7083 @benoitf could you please take another looks at this PR following some new changes?

packages/renderer/src/lib/ui/Typeahead.svelte
axel7083: +1

packages/renderer/src/lib/ui/Typeahead.svelte
```diff
@@ -40,26 +42,66 @@ let input: HTMLInputElement | undefined = $state();
...
-  ),
-);
 let opened: boolean = $state(false);
 let highlightIndex: number = $state(-1);
 let pageStep: number = $state(10);
 let userValue: string = $state('');
 let loading: boolean = $state(false);
 
+let groupedItems: { [group: string]: string[] } = $derived(groupItems(resultItems));
+let itemHeadings: { [index: number]: string[] } = $derived(updateHeadings(groupedItems));
```
benoitf: as these two derived states are the only callers of the function, I would inline these function in the derived block

$derived.by(() => inline the function 

idem for updateHeadings

packages/renderer/src/lib/ui/Typeahead.svelte
axel7083: or  a `type GroupedItem = { [group: string]: string[] } `

axel7083 (approved):
LGTM :+1: Great work

packages/renderer/src/lib/ui/Typeahead.svelte
SoniaSandler: @gastoner @axel7083 updated in `Typeahead.ts` and used in `Typeahead.svelte`

dgolovin: or `ReturnType<typeof groupItems>`

dgolovin (approved):

benoitf (approved):
thanks

PR #10553, opened 2025-01-08
fix: update developer sandbox extension name in e2e tests

dgolovin:
### What does this PR do?

Update sandbox extension name to 'Developer Sandbox' because it was renamed in catalog.

### What issues does this PR fix or reference?

Fixes e2e tests in pr-check workflow.

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature

PR #10576, opened 2025-01-09
chore(tests): adjust sandbox name in e2e tests after catalog update

odockal:
### What does this PR do?
Includes specific condition in e2e tests to accept different name of the extension in catalog and in its details page after installation.

Adjustment to https://github.com/podman-desktop/podman-desktop/pull/10553

### How to test this PR?

### DIFF
--- a/tests/playwright/src/specs/extension-installation-smoke.spec.ts
+++ b/tests/playwright/src/specs/extension-installation-smoke.spec.ts
@@ -34,7 +34,7 @@ const RUNNING = 'RUNNING';
 const NOT_INSTALLED = 'NOT-INSTALLED';
 const DOWNLOADABLE = 'DOWNLOADABLE';
 const OPENSHIFT_LOCAL = 'Red Hat Openshift Local';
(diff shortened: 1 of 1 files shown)


cbr7 (approved):

odockal:
@cbr7 @dgolovin @jeffmaury This PR is to show where is the problem. Right now test need to use new catalog name in some case, and old name when checking the extension heading which still has former name. I think that correct approach would be to release new extension version + update catalog + update e2e tests. At the same time, this should not be done too often. Extension e2e tests in podman-desktop are for verification that we can install extension from catalog and through OCI image, not tailored for specific extension, but to check that functionality, while complex e2e tests should cover this in their extension's repositories. 

Note: I am truly curious, how is it possible that we saw these tests failing (without update to e2e tests to accept new Sandbox name) when the catalog change was merged into `gh-pages` branch. Any idea? Is there a place where this branch might be used in the codebase?

jeffmaury (approved):
LGTM

PR #10612, opened 2025-01-10
feat: adding search controls logs window

axel7083:
### What does this PR do?

Using the [@xterm/addon-search](https://www.npmjs.com/package/@xterm/addon-search) to allow searching in the terminal. This PR add a very basic `TerminalSearchControls` component, that will be placed above the terminal when needed.

For this PR I only enabled the search for the container logs as a demo. We can enable it in every logs after this PR.

### Screenshot / video of UI

image

https://github.com/user-attachments/assets/19d79a07-eb53-46c4-acc6-c372679862d6

Part of https://github.com/podman-desktop/podman-desktop/issues/9726 (I decided to add the logs to the containers instead of pods for this first PR)

### How to test this PR?

1. go to container details
2. go to logs tab 
3. assert search bar visible
4. `ctrl+f` to auto focus in
5. assert arrow up / down working

### DIFF
--- a/packages/renderer/src/lib/ui/TerminalSearchControls.svelte
+++ b/packages/renderer/src/lib/ui/TerminalSearchControls.svelte
@@ -12,7 +12,7 @@ interface Props {
 
 let { terminal }: Props = $props();
 
-let searchAddon: SearchAddon;
+let searchAddon: SearchAddon | undefined;
 let searchTerm: string = $state('');
 
 let input: HTMLInputElement | undefined = $state();
@@ -33,13 +33,13 @@ function onKeyPressed(event: KeyboardEvent): void {
 }
 
 function onSearchNext(incremental = false): void {
-  searchAddon.findNext(searchTerm, {
+  searchAddon?.findNext(searchTerm, {
     incremental: incremental,
   });
 }
 
 function onSearchPrevious(incremental = false): void {
-  searchAddon.findPrevious(searchTerm, {
+  searchAddon?.findPrevious(searchTerm, {
     incremental: incremental,
   });
 }
@@ -50,7 +50,6 @@ function onSearch(event: Event): void {
 }
 
 function onKeyUp(e: KeyboardEvent): void {
-  console.log('onKeyUp', e);
   if (e.ctrlKey && e.key === 'f') {
     input?.focus();
   }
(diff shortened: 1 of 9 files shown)


packages/renderer/src/lib/ui/TerminalSearchControls.svelte
```diff
@@ -0,0 +1,84 @@
...
+
+let input: HTMLInputElement | undefined = $state();
+
+onMount(() => {
+  searchAddon = new SearchAddon();
+  searchAddon.activate(terminal);
+});
+
+onDestroy(() => {
+  searchAddon?.dispose();
```
gastoner, line 26: Is the `?` required here? since the `searchAddon` can have only `SearchAddon` type?

gastoner (approved):
LGTM, tested on Windows

packages/renderer/src/lib/ui/TerminalSearchControls.svelte
axel7083: I would say that you are right, but I don't know if we call at the same time the destructor and the mount if there is no risk of race condition.. :thinking: 

I put the `?` because we already faced in the past some scenarios where it was undefined at some render cycle.

Maybe changing the type definition to `SearchAddon | undefined` is better, but then I would have to add a check in the html part

packages/renderer/src/lib/ui/TerminalSearchControls.svelte
```diff
@@ -0,0 +1,84 @@
...
+import { onDestroy, onMount } from 'svelte';
+import Fa from 'svelte-fa';
+
+interface Props {
+  terminal: Terminal;
+}
+
+let { terminal }: Props = $props();
+
+let searchAddon: SearchAddon;
```
benoitf:
```suggestion
let searchAddon: SearchAddon | undefined;
```

benoitf: better handle

benoitf (approved):

packages/renderer/src/lib/ui/TerminalSearchControls.svelte
axel7083: suggestion applied ✅

PR #10613, opened 2025-01-10
fix: remove disablement of no explicit any rule in compose extension

### LINKED ISSUE #10594: Remove disablement of no-explicit any from compose extension

Impacted files:

- src/detect.spec.ts
- src/download.spec.ts
- src/handler.spec.ts

Remove use of any type

_No response_

### Additional context

_No response_

dgolovin:
### What does this PR do?

Removes disablement of no explicit any rule

Fixes #10594

### How to test this PR?

### DIFF
--- a/extensions/compose/src/detect.spec.ts
+++ b/extensions/compose/src/detect.spec.ts
@@ -16,8 +16,6 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-/* eslint-disable @typescript-eslint/no-explicit-any */
-
 import { EventEmitter } from 'node:events';
 import * as fs from 'node:fs';
 import * as http from 'node:http';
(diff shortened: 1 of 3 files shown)


extensions/compose/src/detect.spec.ts
```diff
@@ -228,19 +226,19 @@ describe('Check docker socket', async () => {
       };
     });
 
-    const spyGet = vi.spyOn(http, 'get') as unknown as MockInstance;
+    const spyGet = vi.spyOn(http, 'get') as MockInstance;
```
feloy: By doing this, we are losing the type of the spy.

Instead, I would do:

(it seems we are forced to cast the signature of `http.get` to the one we want to use, as `vi.mocked` is not able to determine which signature to use, and fallbacks to the wrong one)
```
// this one may be moved to the top of the file, as it is hoisted anyway (https://vitest.dev/api/vi#vi-mock).
vi.mock('node:http');

const clientRequestEmitter = new EventEmitter();
const myRequest = clientRequestEmitter as http.ClientRequest;

vi.mocked(
  http.get as (
    options: http.RequestOptions | string | URL,
    callback: (res: http.IncomingMessage) => void,
  ) => http.ClientRequest,
).mockImplementation((_options, callback) => {
[...]
```

benoitf (approved):

cdrage (approved):

extensions/compose/src/detect.spec.ts
dgolovin: Fixed by using similar approach.

PR #10621, opened 2025-01-13
feat: component Dropdown with multiple choices

### LINKED ISSUE #10620: Make Dropdown ui component multi-selectable

For https://github.com/containers/podman-desktop-extension-ai-lab/issues/1627, we would need a Dropdown with multi-select.

Provide a `multiple` option to the `Dropdown` ui component 

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

- Add a `multiple` option to the `Dropdown` ui component.
- Update the storybook to feature the *multiple* version of the Dropdown component
 
Fixes #10620

### How to test this PR?

Test on Storybook: https://6784f39c949e4310890be866--podman-desktop-pr.netlify.app/storybook?id=dropdown--multiple

- [ ] Tests are covering the bug fix or the new feature (TODO)

### DIFF
--- a/packages/ui/src/lib/dropdown/Dropdown.svelte
+++ b/packages/ui/src/lib/dropdown/Dropdown.svelte
@@ -3,6 +3,8 @@ import { faCaretDown, faCheck } from '@fortawesome/free-solid-svg-icons';
 import { onMount, type Snippet } from 'svelte';
 import Fa from 'svelte-fa';
 
+import Checkbox from '../checkbox/Checkbox.svelte';
+
 interface Option {
   value: unknown;
   label: string;
@@ -20,6 +22,8 @@ let {
   ariaInvalid = false,
   ariaLabel = '',
   children = undefined,
+  multiple = false,
+  label = '',
 }: {
   id?: string;
   name?: string;
@@ -31,23 +35,40 @@ let {
   ariaInvalid?: boolean | 'grammar' | 'spelling';
   ariaLabel?: string;
   children?: Snippet;
+  multiple?: boolean;
+  label?: string;
 } = $props();
 
 let opened: boolean = $state(false);
 let selectLabel: string = $state('');
+let selectLabelPlus: string | undefined = $state(undefined);
 let highlightIndex: number = $state(-1);
 let comp: HTMLElement;
 
 const pageStep: number = 10;
 
 onMount(() => {
-  if (!value && options?.length > 0) {
+  if (!multiple && !value && options?.length > 0) {
     value = options[0].value;
   }
 });
 
 $effect(() => {
-  selectLabel = options.find(o => o.value === value)?.label ?? (typeof value === 'string' ? (value as string) : '');
+  if (multiple && Array.isArray(value)) {
+    selectLabelPlus = undefined;
+    if (!value.length) {
+      selectLabel = label;
+    }
+    const firstValueLabel = options.find(o => Array.isArray(value) && o.value === value[0])?.label ?? '';
+    if (value.length >= 1) {
+      selectLabel = firstValueLabel;
+    }
+    if (value.length > 1) {
+      selectLabelPlus = `+${value.length - 1}`;
+    }
+  } else {
+    selectLabel = options.find(o => o.value === value)?.label ?? (typeof value === 'string' ? (value as string) : '');
+  }
 });
 
 function onKeyDown(e: KeyboardEvent): void {
@@ -131,15 +152,28 @@ function onEnter(i: number): void {
 }
 
 function onSelect(e: Event, newValue: unknown): void {
-  onChange(newValue);
-  value = newValue;
-  close();
-  e.preventDefault();
+  if (multiple) {
+    if (Array.isArray(value)) {
+      if (value.includes(newValue)) {
+        value = value.filter(v => v !== newValue);
+      } else {
+        value = [...value, newValue];
+      }
+    }
+    onChange(value);
+  } else {
+    onChange(newValue);
+    value = newValue;
+    close();
+    e.preventDefault();
+  }
 }
 
 function toggleOpen(e: Event): void {
   if (opened) {
-    close();
+    if (!multiple) {
+      close();
+    }
   } else {
     open();
   }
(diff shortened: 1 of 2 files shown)


packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -213,19 +252,28 @@ function onWindowClick(e: Event): void {
...
-          class:text-[var(--pd-dropdown-item-hover-text)]={highlightIndex === i}>
-          <div class="min-w-4 max-w-4">
-            {#if option.value === value}<Fa icon={faCheck} />{/if}
-          </div>
-          <div class="grow">{option.label}</div>
-        </button>
+        {#if multiple}
+          
+        <Checkbox
+          class={'px-2 py-1' + (highlightIndex === i ? ' bg-[var(--pd-dropdown-item-hover-bg)] text-[var(--pd-dropdown-item-hover-text)]' : '')}
```
gastoner, line 258: Maybe use `px-2 py-1 {highlightIndex === i ? 'bg-[var(--pd-dropdown-item-hover-bg)] text-[var(--pd-dropdown-item-hover-text)]' : ''}` instead of string concatenation?

gastoner (commented):
Codewise LGTM

gastoner:
Is the dark them in the storybook effect the switching the color of the text? e.g. when having dark background the text is black, but works with light theme in browser

When clinking on dropdown again it won't close it 


https://github.com/user-attachments/assets/26e15831-ca6a-44ae-9980-ba88c778db44

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -31,23 +35,40 @@ let {
...
+let selectLabelPlus: string | undefined = $state(undefined);
 let highlightIndex: number = $state(-1);
 let comp: HTMLElement;
 
 const pageStep: number = 10;
 
 onMount(() => {
-  if (!value && options?.length > 0) {
+  if (!multiple && !value && options?.length > 0) {
     value = options[0].value;
```
axel7083, line 52: This is a real problem that we cannot have undefined as a valid option, like nothing selected

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -31,23 +35,40 @@ let {
   ariaInvalid?: boolean | 'grammar' | 'spelling';
   ariaLabel?: string;
   children?: Snippet;
+  multiple?: boolean;
+  label?: string;
 } = $props();
```
axel7083, line 40: We need to improve the typing of the onChange method.

https://github.com/podman-desktop/podman-desktop/blob/b087da26a4123323860e5031e198cee8d63be0ed/packages/ui/src/lib/dropdown/Dropdown.svelte#L32

The typing above is not friendly at all, and should be addressed before making it more complex :/

It should either provide an array, which would be of length one if multiple is not selected

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -31,23 +35,40 @@ let {
...
+    const firstValueLabel = options.find(o => Array.isArray(value) && o.value === value[0])?.label ?? '';
+    if (value.length >= 1) {
+      selectLabel = firstValueLabel;
+    }
+    if (value.length > 1) {
+      selectLabelPlus = `+${value.length - 1}`;
+    }
+  } else {
+    selectLabel = options.find(o => o.value === value)?.label ?? (typeof value === 'string' ? (value as string) : '');
+  }
```
axel7083, line 71: This is too complex, and hard to understand

axel7083: image

this is not clear to the user, I would expect all values to be displayed with separated coma

axel7083 (commented):
Looking at some public library, having the checkbox does not seems to be the _traditional_ method

https://svelte-select-examples.vercel.app/examples/props/multiple

image

https://react-select.com/home

image

https://vue-multiselect.js.org/

image

feloy:
> Looking at some public library, having the checkbox does not seems to be the _traditional_ method
> 
> https://svelte-select-examples.vercel.app/examples/props/multiple
> 
> image
> 
> https://react-select.com/home
> 
> image
> 
> https://vue-multiselect.js.org/
> 
> image

My PR is based on the mockup https://github.com/containers/podman-desktop-extension-ai-lab/issues/1627#issuecomment-2321429148.

@MariaLeonova WDYT?

MariaLeonova:
@axel7083 @feloy I was taking inspiration from [the Patternfly library](https://www.patternfly.org/patterns/filters/#mixed-select-filter-group), and I believe that if in the future users were to select a few options at once, then checkboxes will definitely be a way to go. They could also be supported by dismissible labels for each selected row.

Axel, is there any reason why you would be opposed to the checkboxes? They make it really clear what options have already be selected, and as a user I really appreciate that.

axel7083:
> Axel, is there any reason why you would be opposed to the checkboxes? They make it really clear what options have already be selected, and as a user I really appreciate that.

@MariaLeonova I am not opposed, what I did is: I search on google the most known select library for Svelte, React and Vue and all of them use the same design pattern so that's what I commented.

The checkbox pattern is okay, I can see it for example on Slack

image

But let's look at the Podman-Desktop search system in containers for example

image

We tried to use a similar synthax than GitHub and when we add more filters we had them in the search bar

image

We have a lot of different way of doing the search in Podman-Desktop, and the Dropdown multiselect was not used before, and I am questioning why we don't use the same search system than GitHub since we seems to have some inspiration, and now don't.

MariaLeonova:
> We have a lot of different way of doing the search in Podman-Desktop, and the Dropdown multiselect was not used before, and I am questioning why we don't use the same search system than GitHub since we seems to have some inspiration, and now don't.

Quite frankly, from my non-developer POV dropdown multi-select is a much more accessible way to search for things, as you can see all the options at once and select them, which is in support of the usability heuristics.

I believe we already had this discussion in another ticket and decided we could do both, i.e. have the information show up in this way in search (tool:x, framework:y, etc) and have dropdowns at the same time.

Ultimately, it's up to the team, I only make recommendations :)

axel7083:
> Quite frankly, from my non-developer POV dropdown multi-select is a much more accessible way to search for things, as you can see all the options at once and select them, which is in support of the usability heuristics.

Let's go with checkbox them, as I said I don't have a strong opinion, I just to raise that we should try to target the consistency

packages/ui/src/lib/dropdown/Dropdown.svelte
feloy: @MariaLeonova could you please give me some recommendation on how to display the selected items, when many items are selected?

feloy:
closing as non necessary for now

PR #10623, opened 2025-01-13
fix: handle property the watch mode for extension in development mode

### LINKED ISSUE #10619: watching and reloading extensions is broken
### Bug description

as part of https://github.com/podman-desktop/podman-desktop/issues/8616 to be able to develop an extension using a production binary, the watch mode of reloading an extension should work out of the box

today, it works only once and partially. (upon reload it's removing the watcher as we call deactivate on the extension)
and as it's watching all the files, if I modify a .ts file for example, it triggers a reload but the output js file is removed until generated again and then it's failing

### Operating system

macOS

### Installation Method

None

### Version

1.15.0

### Steps to reproduce

try to modify an extension while in watch mode

actual: extension is not reloaded, broken state

expected: extension is reloaded

### Relevant log output

```shell

```

### Additional context

_No response_

benoitf:
### What does this PR do?
current implementation is broken:
- remove watcher on the first reload (called in deactivate)
- trigger reload even if changes are in the source repository

Move the detection of changes/callback to a separate class handling only the reload
handle properly debounce (do not refresh 10 times if we call the reload method several times in few seconds)
ignore the changes from the source directory

⚠️ introduce the [get-tsconfig](https://www.npmjs.com/package/get-tsconfig) library handling the includes section of the tsconfig.json with matchers. It is being downloaded over 10M each week

fixes https://github.com/podman-desktop/podman-desktop/issues/10619

### How to test this PR?

Use an extension (like ai lab, boot-c) in development mode
Having the extension in watch mode (and Podman Desktop)
change a file in the frontend or backend and wait to see a refresh upon the new build of the extension

### DIFF
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -1725,7 +1725,7 @@ export class ExtensionLoader {
       Array.from(this.activatedExtensions.keys()).map(extensionId => this.deactivateExtension(extensionId)),
     );
     this.kubernetesClient.dispose();
-    this.extensionWatcher.stop();
+    this.extensionWatcher.dispose();
   }
 
   async startExtension(extensionId: string): Promise<void> {
--- a/packages/main/src/plugin/extension-loader.ts
+++ b/packages/main/src/plugin/extension-loader.ts
@@ -53,6 +53,7 @@ import type { Directories } from './directories.js';
 import type { Event } from './events/emitter.js';
 import { Emitter } from './events/emitter.js';
 import { DEFAULT_TIMEOUT, ExtensionLoaderSettings } from './extension-loader-settings.js';
+import type { ExtensionWatcher } from './extension-watcher.js';
 import type { FilesystemMonitoring } from './filesystem-monitoring.js';
 import type { IconRegistry } from './icon-registry.js';
 import type { ImageCheckerImpl } from './image-checker.js';
@@ -141,7 +142,6 @@ export class ExtensionLoader {
 
   protected activatedExtensions = new Map<string, ActivatedExtension>();
   protected analyzedExtensions = new Map<string, AnalyzedExtension>();
-  private watcherExtensions = new Map<string, containerDesktopAPI.FileSystemWatcher>();
   private reloadInProgressExtensions = new Map<string, boolean>();
   protected extensionState = new Map<string, string>();
   protected extensionStateErrors = new Map<string, unknown>();
@@ -194,6 +194,7 @@ export class ExtensionLoader {
     private dialogRegistry: DialogRegistry,
     private safeStorageRegistry: SafeStorageRegistry,
     private certificates: Certificates,
+    private extensionWatcher: ExtensionWatcher,
   ) {
     this.pluginsDirectory = directories.getPluginsDirectory();
     this.pluginsScanDirectory = directories.getPluginsScanDirectory();
@@ -431,6 +432,13 @@ export class ExtensionLoader {
 
     // now we have all extensions, we can load them
     await this.loadExtensions(analyzedExtensions);
+
+    // handle the reload extensions callback
+    this.extensionWatcher.onNeedToReloadExtension(extension => {
+      this.reloadExtension(extension, false).catch((error: unknown) => {
+        console.error('error while reloading extension', error);
+      });
+    });
   }
 
   async analyzeExtension(extensionPath: string, removable: boolean): Promise<AnalyzedExtension> {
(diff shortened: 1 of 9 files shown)


axel7083 (commented):
image

suggestion (non-blocking): could we move the extension logic to a dedicated folder ?

axel7083 (commented):
image

question (non-blocking): If I am on the webview and it refreshes, can we be redirected to the new webview ?

packages/main/src/plugin/extension-loader.ts
```diff
@@ -673,8 +680,18 @@ export class ExtensionLoader {
...
     } finally {
-      this.reloadInProgressExtensions.set(extension.id, false);
+      this.reloadInProgressExtensions.delete(extension.id);
     }
+
+    const notification = this.notificationRegistry.addNotification({
+      extensionId: extension.id,
+      type: 'info',
+      title: `Extension ${extension.manifest.displayName} has been updated`,
+    });
```
axel7083, line 690: image

small issue: I don't want to have a notification every time I change a file :( (I know I can just disable the extension to avoid having it)

packages/main/src/plugin/extension-watcher.ts
```diff
@@ -0,0 +1,127 @@
...
+import type { FileMatcher } from 'get-tsconfig';
+
+import type { Event } from './events/emitter.js';
+import { Emitter } from './events/emitter.js';
+import type { ActivatedExtension, AnalyzedExtension } from './extension-loader.js';
+import { ExtensionTypeScriptConfigParser } from './extension-tsconfig-parser.js';
+import type { FilesystemMonitoring } from './filesystem-monitoring.js';
+
+// In charge of watching the extension and reloading it when it changes
+export class ExtensionWatcher {
```
axel7083: question (non-blocking): shouldn't we use `Disposable` pattern instead of `stop` ?

benoitf:
> question (non-blocking): If I am on the webview and it refreshes, can we be redirected to the new webview ?

I noticed that as well but the webview id is changing. need to follow-up where the smart thing would be (keep the previous id or the webview move to the new id)

gastoner (commented):
The AI lab does not want to load on my machine after detecting some update

axel7083:
> The AI lab does not want to load on my machine after detecting some update

The update must be either in  the media or dist folder from my understanding, so if you update the frontend you need to build it, or do `cd packages/frontend && npm run watch`

benoitf:
@gastoner as pointed out by @axel7083 did you run the watch mode on all the parts of the ai lab extension (frontend need to rebuild the package to have Podman Desktop detecting the changes)

benoitf:
rebased and fixed the windows issue in tests

packages/main/src/plugin/extension-loader.ts
benoitf: I think default is for newcomers. I develop my first extension, I would like to see when it's being refreshed ?

packages/main/src/plugin/extension-watcher.ts
benoitf: why not, updated to use Disposable 👍

packages/main/src/plugin/extension-loader.ts
axel7083: I don't know, I feel like this is a bit too much, and in comparison to other tool/framework you don't get notifications.

- When podman-desktop refresh the ui, it does not send a notification :/
- When my spring boot application refresh it does not send a notification
- When my python flask server refresh itself, it does not send a notification.

I am not used to see a notification, and I would expect to be able to disable them at least

benoitf: in the first case, you see traces in the console and if you change on the backend it refreshes.

The last two sample you're referring, it looks like a biased sample or a wrong comparison. You're comparing a Desktop app and a CLI stuff. spring-boot or python app are server-side/cli launch and there is no Desktop/GUI integration but you see something like a trace. If I save my quarkus app I'll see in how many ms the app has been restarted.

axel7083 (approved):
LGTM :rocket: , we discussed the notification and this is something that can be rework later.

Thanks for the disposable change

> Maybe the following comment was lost https://github.com/podman-desktop/podman-desktop/pull/10623#pullrequestreview-2546891112 but that's still something that may be nice to have

benoitf:
> Maybe the following comment was lost https://github.com/podman-desktop/podman-desktop/pull/10623#pullrequestreview-2546891112 but that's still something that may be nice to have

would be a follow-up to move also existing files
I avoid to introduce and move files at the same time

gastoner (approved):
LGTM, but I was able to make it work only with `pnpm build` command `pnpm watch` worked for me like 1 out of 10 times (I was changing svelte file) even If I had pnpm watch in root, frontend and backend, it only turned the extension off

benoitf:
I'll try to give a try on Windows.

I'll merge as today it's broken, so at least it's fixing the problem on Linux and macOS

PR #10534, opened 2025-01-06
feat: init experimental for preferences items

axel7083:
### What does this PR do?

This PR change the `PreferencesRenderingItem.svelte` component to support redirecting the experimental GitHub discussion link we introduced in https://github.com/podman-desktop/podman-desktop/pull/10437.

> The goal would be to have a dedicated sections in the settings, next to the preferences, where we would group the experimental feature. We will duplicate the items, experimental preferences will always be visible on the preferences, the experimental section will be some kind of filtering.

> ⁉️ I added a flask icon on the title of the item to differentiate them from the others, I feel in the preferences we need to have some kind of indication, otherwise the user may wondering _Why some feature have a link and not others?_ @slemeur  made the remark that we may want to have `experimental` as a tooltip, or as text next to the flask, open to feedback. We can iterate and see.

### Screenshot / video of UI

image

image

### What issues does this PR fix or reference?

- Following https://github.com/podman-desktop/podman-desktop/pull/10437
- Part of https://github.com/podman-desktop/podman-desktop/issues/10223

### How to test this PR?

- [x] Tests are covering the bug fix or the new feature

PR #10647, opened 2025-01-14
feat: support children render for badge component

axel7083:
### What does this PR do?

For https://github.com/podman-desktop/podman-desktop/pull/10534 we need to have the badge with an icon. See UX bellow

image

To allow adding the icon, two options

- adding an `icon: ?` field
- using default slot

I decided to use the default slot for flexibility.

### Screenshot / video of UI

N/A

Required for https://github.com/podman-desktop/podman-desktop/pull/10534

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/Badge.spec.ts
+++ b/packages/renderer/src/lib/ui/Badge.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2024 Red Hat, Inc.
+ * Copyright (C) 2024-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


deboer-tim:
What is the difference between Label and Badge? (@ekidneyrh?) It seems like they are slowly converging instead of having distinct differences.

benoitf (commented):
could you split in two commits: change to svelte5 and then introduce a new thing (the children)

because here it's doing both changes at the same time

gastoner (approved):
Codewise LGTM

axel7083:
> What is the difference between Label and Badge?

Need to work on that

axel7083:
Closing as I can replace the badge with the label component.

PR #10658, opened 2025-01-15
chore(test): e2e test for autodetect of compose up

### LINKED ISSUE #4823: Automate detection of Compose created containers
We would like to cover Podman Desktop ability to detect a compose created objects with an e2e test.

Example: outside of the Podman Desktop. On CLI, we have docker-compose binary installed on the system (binary is on the PATH). We checkout some compose.yaml file, ie. awesome-compose or the one we have in our testing repo: https://github.com/odockal/podman-desktop-qe/tree/main/examples/compose, and run `podman compose up -d` on the cli. 
If Podman Desktop is opened, it will detect the containers/image and label container as one created by compose.

Screenshot_20240610_161112

cbr7:
### What does this PR do?
Adds e2e test case to test autodetection functionality of PD when using podman compose up in CLI

https://github.com/podman-desktop/podman-desktop/issues/4823

### DIFF
--- a/tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
+++ b/tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
@@ -54,6 +54,7 @@ test.beforeAll(async ({ runner, welcomePage }) => {
 });
 
 test.afterAll(async ({ page, runner }) => {
+  test.setTimeout(180_000);
   try {
     await deleteContainer(page, backendContainerName);
     await deleteContainer(page, frontendContainerName);
--- a/tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
+++ b/tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
@@ -16,9 +16,6 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import path from 'node:path';
-import { fileURLToPath } from 'node:url';
-
 import type { Page } from '@playwright/test';
 
 import { CLIToolsPage } from '../model/pages/cli-tools-page';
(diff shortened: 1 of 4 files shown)


cbr7:
Test run from windows cicd on azure here: https://github.com/podman-desktop/e2e/actions/runs/12785451322

tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
```diff
@@ -140,7 +158,30 @@ test.describe.serial('Compose onboarding workflow verification', { tag: '@smoke'
...
+  test('Check Podman Desktop autorefresh when using podman compose up', async ({ navigationBar }) => {
+    test.setTimeout(300_000);
+
+    const composeFilePath = path.resolve(__dirname, '..', '..', 'resources', `compose.yaml`);
+    await runComposeUpFromCLI(composeFilePath);
+
+    const containersPage = await navigationBar.openContainers();
+    await playExpect(containersPage.heading).toBeVisible();
+
+    await playExpect
```
odockal: Looks good, just part of the test should be a check for compose tagged container (kinda parent container containing these two). 
Screenshot_20250115_125224
This way user knows it is compose created container.

benoitf (approved):

odockal (changes requested):

tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
```diff
@@ -140,7 +158,30 @@ test.describe.serial('Compose onboarding workflow verification', { tag: '@smoke'
...
     const composeRow = cliToolsPage.toolsTable.getByLabel(RESOURCE_NAME);
     const composeVersionInfo = composeRow.getByLabel('cli-version');
-    await playExpect(composeVersionInfo).toHaveText('docker-compose ' + composeVersion);
+    await playExpect(composeVersionInfo).toContainText('docker-compose ' + composeVersion);
+  });
+
+  test('Check Podman Desktop autorefresh when using podman compose up', async ({ navigationBar }) => {
+    test.setTimeout(300_000);
+
+    const composeFilePath = path.resolve(__dirname, '..', '..', 'resources', `compose.yaml`);
```
odockal: I think we could divide the test into two. first would be verification of `podman compose up` - we could say we are veryfing that installed binary actually works if command execution ends up without error. And then we check the compose containers and that they are marked as `compose`.

cbr7: Why? This does not make much sense to me, it's literally adding an extra test for no reason, if the CLI command fails the test will end on the spot since the CLI execution will throw an exception with the error message from the CLI.

This is why it's an e2e test, it can test more than one thing at a time and there is no confusion since the error messages will be different depending on what fails.

tests/playwright/src/specs/compose-onboarding-smoke.spec.ts
cbr7: done

odockal (approved):
LGTM. Good work.

Issue #8616, opened 2024-08-27
Add a way to develop extension without the need of using Podman Desktop in development mode


benoitf:
### Is your enhancement related to a problem? Please describe

for now, when we develop an extension we also need to run Podman Desktop in development mode.

While it has its advantages when you'll need to tune/update/change some internals of Podman Desktop like UI library of the API, people wanting to write an extension should be able to just start using a binary version of Podman Desktop.

### Describe the solution you'd like

While a CLI command like `yarn --extension-folder ../path` is nice when you're using the terminal, there should have something possible when using the binary as well. (No need to use the CLI) 

- Have a development mode to be able for example to get the developer console of a webview

PR #10659, opened 2025-01-15
feat: introduce a flag to be able to see 'show DevTools' even using production binaries

benoitf:
### What does this PR do?
Introduce a configuration flag, allowing to turn on development mode
using this flag it's then possible to show the debug menu in the UI like 'show Devtools' while right clicking on the webview icon

related to #8616 

### How to test this PR?

unit tests added

manual tests:
Launch in dev mode, turn off the property and see that right click on a webview icon is not showing the view Devtools menu
turn again the property and see it's working.
Build a production binary, then check you can see the menu (depending on the selected config value)

### DIFF
--- a/packages/main/src/development-mode-tracker.spec.ts
+++ b/packages/main/src/development-mode-tracker.spec.ts
@@ -1,0 +1,160 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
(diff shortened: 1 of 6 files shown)


axel7083 (approved):
Tested and LGTM, the dev tools do appear and disappear by changing the configuration.

PR #10701, opened 2025-01-16
fix: respect appearance theme during render GitHub icon

### LINKED ISSUE #10007: Github registry icon doesn't update when switching to light mode
### Bug description

The Github registry icon doesn't reflect the appearance mode change:
Image

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

vzhukovs:
### What does this PR do?

According to the issue, GitHub icon should respect the theme change and be the light on the dark theme and dark on the dark on the light theme.

### Screenshot / video of UI

Снимок экрана 2025-01-16 в 17 32 44
Снимок экрана 2025-01-16 в 17 32 33

#10007 

### How to test this PR?

Switch between themes and check that GitHub icon also changes own appearance.


extensions/registries/src/extension.ts
```diff
@@ -19,7 +19,7 @@ import * as extensionApi from '@podman-desktop/api';
 
 import dockerIoImage from './images/docker.io.png';
 import gcrIoImage from './images/gcr.io.png';
-import ghcrIoImage from './images/ghcr.io.png';
+import ghcrIoImage from './images/ghcr.io.svg';
```
axel7083, line 22: might need to create a Svelte component with the svg content in it.

Example https://github.com/podman-desktop/podman-desktop/blob/516922e2ef043c208d55e9921edda79f7798aa20/packages/ui/src/lib/icons/ContainerIcon.svelte

axel7083 (commented):
I am wondering if we should not provide more flexibility for the icon property of `RegistrySuggestedProvider`

https://github.com/podman-desktop/podman-desktop/blob/be499999e49627d74cad14efe98c9dbe3fed1111/packages/extension-api/src/extension-api.d.ts#L1085

For example, we could allow multiple value (light, dark) like we do for other component

https://github.com/podman-desktop/podman-desktop/blob/be499999e49627d74cad14efe98c9dbe3fed1111/packages/extension-api/src/extension-api.d.ts#L1632

vzhukovs:
> I am wondering if we should not provide more flexibility for the icon property of `RegistrySuggestedProvider`
> 
> https://github.com/podman-desktop/podman-desktop/blob/be499999e49627d74cad14efe98c9dbe3fed1111/packages/extension-api/src/extension-api.d.ts#L1085
> 
> For example, we could allow multiple value (light, dark) like we do for other component
> 
> https://github.com/podman-desktop/podman-desktop/blob/be499999e49627d74cad14efe98c9dbe3fed1111/packages/extension-api/src/extension-api.d.ts#L1632

Thanks for pointing this. Will take a look if we can do this, so will provide an update to the PR soon.

benoitf (commented):
waiting changes as suggested by Axel

extensions/registries/src/extension.ts
```diff
@@ -58,16 +53,11 @@ const defaultRegistries: extensionApi.RegistrySuggestedProvider[] = [
...
   },
   {
     name: 'Google Container Registry',
     url: 'gcr.io',
     icon: gcrIoImage,
   },
 ];
-
-// Remove data:image/png;base64,
-export function stripImagePrefix(imageContent: string): string {
```
axel7083, line 71: some tests seems to be failing since this has been removed?

vzhukovs: right, thanks, method was removed, but the following test wasn't, fixed in the following commit

cdrage (approved):
Tested and LGTM. The GitHub icon is displayed properly now.

axel7083 (approved):
Codewise LGTM

PR #10705, opened 2025-01-16
chore: enable the @typescript-eslint/explicit-function-return-type for renderer part

### LINKED ISSUE #10020: enable the @typescript-eslint/explicit-function-return-type for renderer part

this rule is being disabled but when people contribute new code we're asking to add the return type

it should be the job of the linter.

Fix all missing return type in the code so we can enable it

fix all @typescript-eslint/explicit-function-return-type issues

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

Enable the @typescript-eslint/explicit-function-return-type for renderer part

Fixes #10020 

### How to test this PR?

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -305,7 +305,6 @@ export default [
     rules: {
       '@typescript-eslint/no-empty-function': 'off',
       'no-undef': 'off',
-      '@typescript-eslint/explicit-function-return-type': 'off',
       'no-self-assign': 'off',
       'sonarjs/no-empty-function': 'off',
       'sonarjs/sonar-prefer-regexp-exec': 'off',
--- a/packages/renderer/src/App.spec.ts
+++ b/packages/renderer/src/App.spec.ts
@@ -150,7 +150,7 @@ test('opens submenu when a `submenu` menu is opened', async () => {
       link: '/tosubmenu',
       tooltip: 'With submenu',
       type: 'submenu',
-      get counter() {
+      get counter(): number {
         return 0;
       },
       items: [{} as NavigationRegistryEntry],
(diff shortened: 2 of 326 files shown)


benoitf (commented):
looks ok
with the new svelte version and checker it might be good but with previous one we were not able to define a return type inside the svelte component

I think this is why we tried to say to always have a function and call the function from the svelte block rather than having the return type in the html/svelte block of the svelte file

benoitf:
you could split the PR in several chunks (just the last one will contain the removal of the rule)

feloy:
> you could split the PR in several chunks (just the last one will contain the removal of the rule)

ok, I'll first be sure that all necessary changes are done and tests pass, then I'll split the PR into several ones

benoitf:
I mean, if you have it at the end you don't need to split

the idea was to do small chunks still passing the actual checks, so it's small PRs, it won't stay days, and it won't produce merge conflicts with other changes

feloy:
> I mean, if you have it at the end you don't need to split
> 
> the idea was to do small chunks still passing the actual checks, so it's small PRs, it won't stay days, and it won't produce merge conflicts with other changes

It also depends on how reviews go. I would not like to give too much work on only 2 reviewers; splitting the PR would distribute the work across people

If it can help, you can review per commit (I tried to classify files types in different commits), and I can move commits to other PRs as soon as they are validated

benoitf (approved):

benoitf:
I would go ahead and merge this PR if there is no other review in the 4 upcoming hours

axel7083 (approved):
This PR make my Browser lag :(

PR #10712, opened 2025-01-17
chore: add a property to store folders used as extension folders

benoitf:
### What does this PR do?
add a hidden property to store folders used as extension folders
no UI for now it's just adding backend way on adding/editing/removing and the store keeping the values

### Screenshot / video of UI

N/A

### How to test this PR?

unit tests added

### DIFF
--- a/packages/main/src/plugin/extension/extension-development-folders.ts
+++ b/packages/main/src/plugin/extension/extension-development-folders.ts
@@ -1,0 +1,150 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ConfigurationRegistry, IConfigurationNode } from '/@/plugin/configuration-registry.js';
+import { Emitter } from '/@/plugin/events/emitter.js';
+import type { ExtensionDevelopmentFolderInfo } from '/@api/extension-development-folders-info.js';
+import { ExtensionDevelopmentFolderInfoSettings } from '/@api/extension-development-folders-info.js';
+
+import type { ApiSenderType } from '../api.js';
+import type { ExtensionLoader } from './extension-loader.js';
+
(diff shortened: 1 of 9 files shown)


packages/main/src/plugin/extension/extension-development-folders.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  async removeDevelopmentFolder(path: string): Promise<void> {
+    this.#developmentFolders = this.#developmentFolders.filter(folder => folder !== path);
+
+    // persist the changes
+    await this.saveToConfiguration();
+
+    // dispatch
+    this.#onDidUpdateDevelopmentFolders.fire(this.getDevelopmentFolders());
```
feloy, line 144: should the extension be unloaded here, as it has been loaded in `addDevelopmentFolder`?

benoitf: the UI prevent to stop a running extension.

so the only thing that could be added is a check to throw an error if there is a matching extension running

packages/main/src/plugin/extension/extension-development-folders.ts
```diff
@@ -0,0 +1,150 @@
...
+      .getConfiguration(ExtensionDevelopmentFolderInfoSettings.SectionName)
+      .update(ExtensionDevelopmentFolderInfoSettings.DevelopmentExtensionsFolders, this.#developmentFolders);
+
+    // send an event to refresh the value
+    this.#apiSender.send('extensions-development-folders-changed');
+  }
+
+  async addDevelopmentFolder(path: string): Promise<void> {
+    // check we have extension loader
+    if (!this.#extensionLoader) {
```
axel7083, line 111: No way to avoid the circular dependencies ?

packages/renderer/src/stores/extensionDevelopmentFolders.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+beforeEach(() => {
+  vi.restoreAllMocks();
+  vi.resetAllMocks();
+  vi.mocked(window.events).receive.mockImplementation((channel, args) => {
+    return eventEmitter(channel, args);
+  });
+
+  const spyAddEventListener = vi.spyOn(window, 'addEventListener');
+  spyAddEventListener.mockImplementation((event, callback) => {
```
axel7083:
```suggestion
  vi.spyOn(window, 'addEventListener').mockImplementation((event, callback) => {
```

suggestion: inline?

packages/main/src/plugin/extension/extension-development-folders.ts
benoitf: here, it'll be a follow up to have the analyzer of an extension not being part of extension loader

I don't want to cause a big bang yet

axel7083: Okey np, good to know that this is planned 👍

axel7083 (approved):
LGTM

feloy (approved):

packages/renderer/src/stores/extensionDevelopmentFolders.spec.ts
benoitf: fixed

PR #10717, opened 2025-01-20
feat: update create container workflow

### LINKED ISSUE #9978: Implement design for the create container from image workflow

SoniaSandler:
### What does this PR do?
This PR updates the create container from existing image workflow. When choosing this option, the user will be taken to a separate page similar to the `Pull Image` page, where they'll have the option to either select a local image or pull one from a registry to use. After selecting the image, the user will be moved to the build container page.
Most of the search image in remote registry and pull image code is taken from `PullImage.svelte`

### Screenshot / video of UI

[Screencast from 2025-02-24 15-23-15.webm](https://github.com/user-attachments/assets/fa870987-266d-4e53-8e20-aceb0bc2fb40)

Closes https://github.com/podman-desktop/podman-desktop/issues/9978

### How to test this PR?

### DIFF
--- a/packages/api/src/navigation-page.ts
+++ b/packages/api/src/navigation-page.ts
@@ -21,7 +21,7 @@ export enum NavigationPage {
   CONTAINERS = 'containers',
   CONTAINER_EXPORT = 'container-export',
   CONTAINER = 'container',
-  CONTAINER_CREATE_EXISITNG_IMAGE = 'create-container-existing-image',
+  EXISITNG_IMAGE_CREATE_CONTAINER = 'existing-image-create-container',
   CONTAINER_LOGS = 'container-logs',
   CONTAINER_INSPECT = 'container-inspect',
   CONTAINER_TERMINAL = 'container-terminal',
--- a/packages/api/src/navigation-page.ts
+++ b/packages/api/src/navigation-page.ts
@@ -21,6 +21,7 @@ export enum NavigationPage {
   CONTAINERS = 'containers',
   CONTAINER_EXPORT = 'container-export',
   CONTAINER = 'container',
+  CONTAINER_CREATE_EXISITNG_IMAGE = 'create-container-existing-image',
   CONTAINER_LOGS = 'container-logs',
   CONTAINER_INSPECT = 'container-inspect',
   CONTAINER_TERMINAL = 'container-terminal',
--- a/packages/api/src/navigation-page.ts
+++ b/packages/api/src/navigation-page.ts
@@ -21,7 +21,7 @@ export enum NavigationPage {
   CONTAINERS = 'containers',
   CONTAINER_EXPORT = 'container-export',
   CONTAINER = 'container',
-  EXISITNG_IMAGE_CREATE_CONTAINER = 'existing-image-create-container',
+  EXISTING_IMAGE_CREATE_CONTAINER = 'existing-image-create-container',
   CONTAINER_LOGS = 'container-logs',
   CONTAINER_INSPECT = 'container-inspect',
   CONTAINER_TERMINAL = 'container-terminal',
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -156,7 +156,7 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
           <ContainerList searchTerm={meta.query.filter || ''} />
         </Route>
         <Route path="/containers/:id/*" let:meta firstmatch>
-          <Route path="/create-container-existing-image" breadcrumb="Select image" >
+          <Route path="/existing-image-create-container" breadcrumb="Select image" >
             <CreateContainerFromExisitingImage />
           </Route>
           <Route path="/export" breadcrumb="Export Container">
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -19,7 +19,7 @@ import SecretDetails from './lib/configmaps-secrets/SecretDetails.svelte';
 import ContainerDetails from './lib/container/ContainerDetails.svelte';
 import ContainerExport from './lib/container/ContainerExport.svelte';
 import ContainerList from './lib/container/ContainerList.svelte';
-import CreateContainerFromExisitingImage from './lib/container/CreateContainerFromExistingImage.svelte';
+import CreateContainerFromExistingImage from './lib/container/CreateContainerFromExistingImage.svelte';
 import ContextKey from './lib/context/ContextKey.svelte';
 import CronJobDetails from './lib/cronjob/CronJobDetails.svelte';
 import CronJobList from './lib/cronjob/CronJobList.svelte';
@@ -156,9 +156,6 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
           <ContainerList searchTerm={meta.query.filter || ''} />
         </Route>
         <Route path="/containers/:id/*" let:meta firstmatch>
-          <Route path="/existing-image-create-container" breadcrumb="Select image" >
-            <CreateContainerFromExisitingImage />
-          </Route>
           <Route path="/export" breadcrumb="Export Container">
             <ContainerExport containerID={meta.params.id} />
           </Route>
@@ -175,6 +172,9 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
         </Route>
         <Route path="/images" breadcrumb="Images" navigationHint="root">
           <ImagesList />
+        </Route>
+        <Route path="/images/existing-image-create-container" breadcrumb="Select image" >
+          <CreateContainerFromExistingImage />
         </Route>
         <Route path="/images/:id/:engineId" breadcrumb="Images" let:meta navigationHint="root">
           <ImagesList searchTerm={meta.params.id} imageEngineId={meta.params.engineId} />
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -19,7 +19,7 @@ import SecretDetails from './lib/configmaps-secrets/SecretDetails.svelte';
 import ContainerDetails from './lib/container/ContainerDetails.svelte';
 import ContainerExport from './lib/container/ContainerExport.svelte';
 import ContainerList from './lib/container/ContainerList.svelte';
-import CreateContainerFromExisitingImage from './lib/container/CreateContainerFromExisitingImage.svelte';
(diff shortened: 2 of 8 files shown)


packages/renderer/src/navigation.ts
```diff
@@ -49,6 +49,9 @@ export const handleNavigation = (request: InferredNavigationRequest<NavigationPa
     case NavigationPage.CONTAINER:
       router.goto(`/containers/${request.parameters.id}/`);
       break;
+    case NavigationPage.EXISITNG_IMAGE_CREATE_CONTAINER:
+      router.goto(`/containers/0/existing-image-create-container`);
```
SoniaSandler: Since it is a part of the containers page, I think like the path should be related to containers, but not having the `/0/` as part of it will cause it to clash with the `"/containers/:id/*"` path in `App.svelte`.
We could do a completely separate path from containers (just `/existing-image-create-container`?).
I'm open to any other suggestions or ideas.

packages/api/src/navigation-page.ts
```diff
@@ -21,6 +21,7 @@ export enum NavigationPage {
   CONTAINERS = 'containers',
   CONTAINER_EXPORT = 'container-export',
   CONTAINER = 'container',
+  EXISITNG_IMAGE_CREATE_CONTAINER = 'existing-image-create-container',
```
benoitf:
```suggestion
  EXISTING_IMAGE_CREATE_CONTAINER = 'existing-image-create-container',
```
issue: typo

packages/renderer/src/lib/container/CreateContainerFromExistingImage.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+    { name: 'image12', description: '', star_count: 3, is_official: true },
+  ]);
+  render(CreateContainerFromExistingImage);
+  await tick();
+
+  const inputBox = screen.getByPlaceholderText('Select an exisiting image');
+  expect(inputBox).toBeInTheDocument();
+  await userEvent.type(inputBox, 'image12');
+
+  await tick();
```
benoitf: question: wondering if we need all these tick and not only the waitFor ?

benoitf (commented):
I'll need to check live the PR

benoitf (dismissed):
nice enhancement

I have some minor (non blocking comments):

1. 
The dropdown could be in an expanded state or have a visual indicator on the right that it's a dropdown

Current screen:
image

I need to enter something to see the existing images in the dropdown
image

2. non blocking: more a question to @ekidneyrh 

image

if I select an image then I have a button saying 'Select image' but if I click on the button I'll get the run image prompt
so should it say instead "Run image..."

3. non blocking (question to @ekidneyrh )

If I enter an image that I don't have, I have the button being "Pull Image".
image

Great, it means that I don't have it and I'll need to pull the image

but then I have this screen
image

At first I wanted to run a container from the image. So should I have a button named "Pull image and run..." where it pulls the image and I'm then redirected to the run screen prompt
image

My point is that I would avoid one click and I get a message saying "Image is being pulled, redirect to the creation of the container once the image is downloaded"


4. blocking issue

Depending on how I click or select the image name, at some point I get a 'Select Image' button but I can't click on it. I can only click on the cancel button

Here is an attached recording

https://github.com/user-attachments/assets/47a4eb15-d42e-43ce-9a42-68324e700879

packages/renderer/src/App.svelte
```diff
@@ -17,6 +17,7 @@ import SecretDetails from './lib/configmaps-secrets/SecretDetails.svelte';
 import ContainerDetails from './lib/container/ContainerDetails.svelte';
 import ContainerExport from './lib/container/ContainerExport.svelte';
 import ContainerList from './lib/container/ContainerList.svelte';
+import CreateContainerFromExisitingImage from './lib/container/CreateContainerFromExistingImage.svelte';
```
benoitf: filename has a typo on existing

packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
```diff
@@ -0,0 +1,421 @@
...
+  </svelte:fragment>
+  <svelte:fragment slot="actions">
+    <Button on:click={gotoManageRegistries} icon={faCog}>Manage registries</Button>
+  </svelte:fragment>
+  <div slot="content" class="space-y-2 flex flex-col">
+    <div class="flex flex-col">
+      <Typeahead
+        id="imageName"
+        name="imageName"
+        placeholder="Select an exisiting image"
```
benoitf:
```suggestion
        placeholder="Select an existing image"
```
typo

benoitf: cc @ekidneyrh 

(non blocking)

I'm not a native english speaker but I found 'existing' wording might be confusing. As it's existing in the world being locally or remote one.

But when reading that, I think existing is from 'my local storage', an image that I have already downloaded. Because 'select' let me think I can only select, not enter text to find something I don't have yet

suggestions: 

- Enter the image to use
- Choose or enter an image to run
- Select or enter an image to use

packages/renderer/src/navigation.ts
```diff
@@ -49,6 +49,9 @@ export const handleNavigation = (request: InferredNavigationRequest<NavigationPa
     case NavigationPage.CONTAINER:
       router.goto(`/containers/${request.parameters.id}/`);
       break;
+    case NavigationPage.EXISITNG_IMAGE_CREATE_CONTAINER:
```
benoitf: typo in the constant

packages/renderer/src/navigation.ts
benoitf: I would move the page to /images/ but maybe there is the same issue ?

SoniaSandler:
> 4. blocking issue
> 
> Depending on how I click or select the image name, at some point I get a 'Select Image' button but I can't click on it. I can only click on the cancel button

@benoitf the problem was with `docker.io/<some image name>` images since they show up in the local list as `docker.io/library/<some image name>`. It is fixed now along with the other reviews (except for those that are waiting for UX input and the [open dropdown](https://github.com/podman-desktop/podman-desktop/pull/11319)).

ekidneyrh:
> 2. non blocking: more a question to @ekidneyrh

> if I select an image then I have a button saying 'Select image' but if I click on the button I'll get the run image prompt so should it say instead "Run image..."
> 
Yeah this works as we have the heading telling the user to select an image.

> 3. non blocking (question to @ekidneyrh )
> 
> If I enter an image that I don't have, I have the button being "Pull Image". 
> Great, it means that I don't have it and I'll need to pull the image
> but then I have this screen 
> At first I wanted to run a container from the image. So should I have a button named "Pull image and run..." where it pulls the image and I'm then redirected to the run screen prompt 
> 
> My point is that I would avoid one click and I get a message saying "Image is being pulled, redirect to the creation of the container once the image is downloaded"
> 

This also works! So a message appears and then redirects the user? Where would it appear? In the task manager or on the screen?

packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
ekidneyrh: I think the last option you have works best with 'run' instead of 'use'. 'Select or enter an image to run'

Would we have to change the button text on the 'Create' dialog as that used the wording also?

benoitf:
> This also works! So a message appears and then redirects the user? Where would it appear? In the task manager or on the screen?

my initial though was on the screen so I see an action being in progress and if I wait few seconds (depending on how fast I can download the images) I will be able to run the image
(as I can get/see the progress download)

benoitf:
> @benoitf the problem was with docker.io/<some image name> images since they show up in the local list as docker.io/library/<some image name>. It is fixed now along with the other reviews (except for those that are waiting for UX input and the https://github.com/podman-desktop/podman-desktop/pull/11319).

ah thanks for catching the source of the failure

benoitf (approved):

ekidneyrh:
> > This also works! So a message appears and then redirects the user? Where would it appear? In the task manager or on the screen?
> 
> my initial though was on the screen so I see an action being in progress and if I wait few seconds (depending on how fast I can download the images) I will be able to run the image (as I can get/see the progress download)

Will there be sufficient time for the user to see the message? Could the image be pulled quickly? 

We could leave that part out and just have "Pull image and run" and redirection to the run screen prompt?

packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
```diff
@@ -0,0 +1,429 @@
...
+import Fa from 'svelte-fa';
+import { router } from 'tinro';
+
+import { lastPage } from '/@/stores/breadcrumb';
+import { runImageInfo } from '/@/stores/run-image-store';
+import type { ImageSearchOptions } from '/@api/image-registry';
+import type { ProviderContainerConnectionInfo } from '/@api/provider-info';
+import type { PullEvent } from '/@api/pull-event';
+
+import { providerInfos } from '../../stores/providers';
```
benoitf: nit: check all imports use /@ here some have some don't

SoniaSandler:
@benoitf @ekidneyrh I've updated the placeholder text and button text.
For the pull image and run suggestion, this is how it will look like if we decide to redirect to the run screen right after the pull is done:

[Screencast from 2025-02-24 14-50-33.webm](https://github.com/user-attachments/assets/6db5d659-b15f-4e1b-9cc4-73cf9a408442)

SoniaSandler:
@benoitf the pull image and run workflow has been also updated as requested and all the imports are now using `@`.
For the open dropdown on focus I created https://github.com/podman-desktop/podman-desktop/pull/11319

benoitf:
@SoniaSandler on my end I'm fine to merge and do follow-ups

SoniaSandler:
@benoitf so no need for another approval?

benoitf:
well technically as it's a feat, we need another approval, you might ping North America committers

I was speaking for myself

SoniaSandler:
rebased following the merge of https://github.com/podman-desktop/podman-desktop/pull/11319

benoitf:
@dgolovin @gastoner please review the PR

gastoner (approved):
LGTM on Windows

PR #10728, opened 2025-01-20
refactor: migrate TerminalWindow to svelte5

axel7083:
### What does this PR do?

While trying to migrate `PullImage` component to svelte5 I face a typecheck issues with the TerminalWindows with the `bind:terminal`. I fixed it by migrating the TerminalWindows to svelte 5.

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

Following https://github.com/podman-desktop/podman-desktop/pull/10727
Required for https://github.com/podman-desktop/podman-desktop/issues/10726

### How to test this PR?

- [x] Tests are covering the bug fix or the new feature

PR #10727, opened 2025-01-20
refactor: migrate typeahead to svelte5

axel7083:
### What does this PR do?

Migrate the Typeahead component to ease refactoring.

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

Step 1 of https://github.com/podman-desktop/podman-desktop/issues/10726

### How to test this PR?

- [x] Tests are covering the bug fix or the new feature

Issue #10726, opened 2025-01-20
refactor(Typeahead): make it more generic


axel7083:
### Is your enhancement related to a problem? Please describe

As started in https://github.com/podman-desktop/podman-desktop/pull/10573 this is more work than expected, and should be split into smaller chunk

Required for
- https://github.com/podman-desktop/podman-desktop/issues/9978

The design is the one bellow

![Image](https://github.com/user-attachments/assets/8c7a069b-e15f-42d5-89b6-35d800a13b70)

### Describe the solution you'd like

Let's split some steps to reach the goal

1. [x] #10729
2. [x] refactor(Typeahead): remove the search function from the Typeahead component (PR: https://github.com/podman-desktop/podman-desktop/pull/10737)
3. [x] feat(Typeahead): support grouping items (PR: https://github.com/podman-desktop/podman-desktop/pull/10573)

axel7083:
Tech dept, still important

axel7083:
not sure what to implement actually

PR #10730, opened 2025-01-20
refactor: migrate pull image svelte 5

axel7083:
### What does this PR do?

Migrate the PullImage component which uses the Typeahead component. Will be easier to continue working on https://github.com/podman-desktop/podman-desktop/issues/10726 by having them both to svelte5

Following https://github.com/podman-desktop/podman-desktop/pull/10727

Need rebase after the following
- https://github.com/podman-desktop/podman-desktop/pull/10728
- https://github.com/podman-desktop/podman-desktop/pull/10733

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/PullImage.svelte
+++ b/packages/renderer/src/lib/image/PullImage.svelte
@@ -20,23 +20,34 @@ import RecommendedRegistry from './RecommendedRegistry.svelte';
 const DOCKER_PREFIX = 'docker.io';
 const DOCKER_PREFIX_WITH_SLASH = DOCKER_PREFIX + '/';
 
-let logsPull: Terminal;
-let pullError = '';
-let pullInProgress = false;
-let pullFinished = false;
-let shortnameImages: string[] = [];
-let podmanFQN = '';
-let usePodmanFQN = false;
-let isValidName = true;
+interface Props {
+  imageToPull?: string;
+}
 
-export let imageToPull: string | undefined = undefined;
+let { imageToPull }: Props = $props();
 
-$: providerConnections = $providerInfos
-  .map(provider => provider.containerConnections)
-  .flat()
-  .filter(providerContainerConnection => providerContainerConnection.status === 'started');
+let logsPull: Terminal | undefined = $state();
+let pullError = $state('');
+let pullInProgress = $state(false);
+let pullFinished = $state(false);
+let shortnameImages: string[] = $state([]);
+let podmanFQN = $state('');
+let usePodmanFQN = $state(false);
+let isValidName = $state(true);
 
-let selectedProviderConnection: ProviderContainerConnectionInfo | undefined;
+let latestTagMessage: string | undefined = $state();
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/image/PullImage.svelte
```diff
@@ -20,23 +20,34 @@ import RecommendedRegistry from './RecommendedRegistry.svelte';
...
-  .flat()
-  .filter(providerContainerConnection => providerContainerConnection.status === 'started');
+let latestTagMessage: string | undefined = $state();
+let imageNameInvalid: string | undefined = $state();
+let imageNameIsInvalid = $state(imageToPull === undefined || imageToPull.trim() === '');
 
-let selectedProviderConnection: ProviderContainerConnectionInfo | undefined;
+// all started connection
+let providerConnections: ProviderContainerConnectionInfo[] = $derived(
+  $providerInfos
```
benoitf, line 44: AFAIK I had issues with the `$` usage of stores within state/derived not being reactive.
Did you test like you go the pullImage screen, and you start the podman machine later. is it updating the providerConnections as expected ?

packages/renderer/src/lib/image/PullImage.svelte
```diff
@@ -46,7 +57,9 @@ async function resolveShortname(): Promise<void> {
     return;
   }
   if (imageToPull && !imageToPull.includes('/')) {
-    shortnameImages = (await window.resolveShortnameImage(selectedProviderConnection, imageToPull)) ?? [];
+    shortnameImages =
+      (await window.resolveShortnameImage($state.snapshot(selectedProviderConnection), $state.snapshot(imageToPull))) ??
```
benoitf, line 61: question: new pattern to follow ?

benoitf: or when exposing the method, the proxy should handle these Proxy objects and unwrap them using state snapshot ?

axel7083: If we don't use `$state.snapshot` we get the proxy which cannot be serialized, we could maybe change the proxy logic, but I don't know how


We already use it some places

https://github.com/podman-desktop/podman-desktop/blob/4cd79ac55b66db0229f75e5b99304cfc236630ae/packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.svelte#L52

packages/renderer/src/lib/image/PullImage.svelte
axel7083: Indeed, seems some issues are possible, moving to draft, will try to fix it

axel7083: The problem was coming from the Dropdown, and I applied the fix I found in https://github.com/podman-desktop/podman-desktop/issues/10734#issuecomment-2602654447

axel7083:
Closing as require our `Dropdown` component to support Svelte5.

PR #10731, opened 2025-01-20
feat: allow to use production binary to develop extensions

### LINKED ISSUE #8616: Add a way to develop extension without the need of using Podman Desktop in development mode

benoitf:
### What does this PR do?
Allow to add/track folder having extensions after turning on the development mode flag

### Screenshot / video of UI

image

fixes https://github.com/podman-desktop/podman-desktop/issues/8616

### How to test this PR?

Unit tests added

Manual test:

1.  start Podman Desktop in watch mode (but without an --extension-folder flag on the CLI) or production mode 
1. enable the 'development mode' in the settings/preferences for extensions
1. go to the puzzle icon/extension in main navbar on the left
1. click on the add... button in the list and select the backend/extension folder of an extension (you can select a dummy folder but it'll report an error asking you to select a valid extension folder)
1. then you should see the extension being loaded.

you can even restart Podman Desktop, the extension should be still there

### DIFF
--- a/packages/renderer/src/lib/extensions/ExtensionList.svelte
+++ b/packages/renderer/src/lib/extensions/ExtensionList.svelte
@@ -11,6 +11,7 @@ import { featuredExtensionInfos } from '/@/stores/featuredExtensions';
 
 import type { CatalogExtensionInfoUI } from './catalog-extension-info-ui';
 import CatalogExtensionList from './CatalogExtensionList.svelte';
+import DevelopmentExtensionList from './dev-mode/DevelopmentExtensionList.svelte';
 import { ExtensionsUtils } from './extensions-utils';
 import InstallManuallyExtensionModal from './InstallManuallyExtensionModal.svelte';
 
@@ -66,7 +67,7 @@ function closeModal(): void {
   installManualImageModal = false;
 }
 
-let screen: 'installed' | 'catalog' = 'installed';
+let screen: 'installed' | 'catalog' | 'development' = 'installed';
 
 let installManualImageModal: boolean = false;
 </script>
@@ -108,7 +109,13 @@ let installManualImageModal: boolean = false;
         screen = 'catalog';
       }}
       selected={screen === 'catalog'}>Catalog</Button>
-  {/snippet}
+      <Button
+      type="tab"
+      on:click={(): void => {
+        screen = 'development';
+      }}
+      selected={screen === 'development'}>Local Extensions</Button>
+ {/snippet}
 
   {#snippet content()}
   <div class="flex min-w-full h-full">
@@ -121,7 +128,7 @@ let installManualImageModal: boolean = false;
           on:resetFilter={(): string => (searchTerm = '')} />
       {/if}
       <InstalledExtensionList extensionInfos={$filteredInstalledExtensions} />
-    {:else}
+    {:else if screen === 'catalog'}
       {#if searchTerm && $filteredCatalogExtensions.length === 0}
         <FilteredEmptyScreen
           icon={ExtensionIcon}
@@ -130,6 +137,8 @@ let installManualImageModal: boolean = false;
           on:resetFilter={(): string => (searchTerm = '')} />
       {/if}
       <CatalogExtensionList showEmptyScreen={!searchTerm} catalogExtensions={$filteredCatalogExtensions} />
+    {:else if screen === 'development'}
+      <DevelopmentExtensionList />
     {/if}
   </div>
   {/snippet}
--- a/packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionEmptyScreen.svelte
+++ b/packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionEmptyScreen.svelte
@@ -1,0 +1,17 @@
+<script lang="ts">
+import { faCog } from '@fortawesome/free-solid-svg-icons';
+import { Button, EmptyScreen } from '@podman-desktop/ui-svelte';
+
+async function openExtensionDocumentation(): Promise<void> {
+  await window.openExternal('https://podman-desktop.io/docs/extensions/developing');
+}
+</script>
+
+<EmptyScreen
+  icon={faCog}
+  title="Local extensions (developer mode)"
+  message="Enable Preferences > Extensions > Development Mode to test local extensions">
+  <div class="flex gap-2 justify-center">
+    <Button type="link" on:click={openExtensionDocumentation}>How to write your first extension</Button>
+  </div>
+</EmptyScreen>
--- a/packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
+++ b/packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
@@ -84,7 +84,7 @@ async function addLocalFolderExtension(): Promise<void> {
       <!-- List all -->
       <div class="bg-[var(--pd-content-bg)] rounded flex flex-col p-2">
         <div class="flex flex-row w-full">
-          <div class="text font-medium first-letter:uppercase pb-2">Local extension folders being tracked:</div>
+          <div class="text font-medium first-letter:uppercase pb-2">Local extension folders being tracked</div>
           <div class="flex-grow flex flex-col items-end">
             <Button on:click={addLocalFolderExtension}>Add a local folder extension...</Button>
           </div>
--- a/packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
+++ b/packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
@@ -1,0 +1,106 @@
+<script lang="ts">
+import { Button } from '@podman-desktop/ui-svelte';
+import { onDestroy, onMount } from 'svelte';
+import type { Unsubscriber } from 'svelte/store';
+
+import DevelopmentExtensionListTable from '/@/lib/extensions/dev-mode/table/ListTable.svelte';
+import { extensionDevelopmentFolders } from '/@/stores/extensionDevelopmentFolders';
+import { extensionInfos } from '/@/stores/extensions';
+import type { ExtensionDevelopmentFolderInfo } from '/@api/extension-development-folders-info';
+import type { ExtensionInfo } from '/@api/extension-info';
+import { ExtensionLoaderSettings } from '/@api/extension-loader-settings';
+
+import type { SelectableExtensionDevelopmentFolderInfoUI } from './development-folder-info-ui';
+import DevelopmentExtensionEmptyScreen from './DevelopmentExtensionEmptyScreen.svelte';
+
+let isDevelopmentModeEnabled = $state(false);
+
+const unsubscribers: Unsubscriber[] = [];
+
+let currentExtFolders: ExtensionDevelopmentFolderInfo[] = $state([]);
+let currentExtensionInfos: ExtensionInfo[] = $state([]);
+
+const selectableExtensionDevelopmentFolders: SelectableExtensionDevelopmentFolderInfoUI[] = $derived(
+  currentExtFolders.map(folder => {
+    // do we have a matching extension
+    const matchingExtension = currentExtensionInfos.find(ext => ext.path === folder.path);
+    let extension = undefined;
+    if (matchingExtension) {
+      extension = { name: matchingExtension.name, state: matchingExtension.state, id: matchingExtension.id };
+    }
+
+    return { ...folder, selected: false, name: folder.path, extension };
+  }),
+);
+
+onMount(async () => {
+  //
+  // Check if development mode is enabled
+  isDevelopmentModeEnabled =
+    (await window.getConfigurationValue(
+      `${ExtensionLoaderSettings.SectionName}.${ExtensionLoaderSettings.DevelopmentMode}`,
+    )) ?? false;
+
+  // subscribe to extension changes
+  unsubscribers.push(
+    extensionInfos.subscribe(value => {
+      currentExtensionInfos = value;
+    }),
+  );
+  unsubscribers.push(
+    extensionDevelopmentFolders.subscribe(value => {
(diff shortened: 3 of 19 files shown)


packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionTableActionsColumn.spec.ts
```diff
@@ -0,0 +1,61 @@
+/**********************************************************************
```
axel7083: question: Can we simplify the naming ? I can't even see the full name, maybe we can create subfolders ? 

I think we can simplify the filename if they parent name is explicit, example `dev-mode/extension/EmptyPage.svelte` would be explicit enough ? what do you think?

packages/renderer/src/lib/extensions/dev-mode/SelectableExtensionDevelopmentFolderActionUntrack.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+const extensionFolderWithExtensionStarted: SelectableExtensionDevelopmentFolderInfoUI = {
+  path: 'foo',
+  extension: {
+    id: 'extensionid',
+    name: 'My Extension',
+    state: 'started',
+  },
+} as unknown as SelectableExtensionDevelopmentFolderInfoUI;
+
+test('Expect untrack action being displayed if there is an extension', async () => {
```
feloy:
```suggestion
test('Expect untrack action being displayed if there is a stopped extension', async () => {
```

packages/renderer/src/lib/extensions/dev-mode/SelectableExtensionDevelopmentFolderActionUntrack.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  expect(untrackButton).toBeInTheDocument();
+
+  // click on the button
+  await fireEvent.click(untrackButton);
+
+  // expect the window.untrackExtensionFolder to be called
+  expect(window.untrackExtensionFolder).toHaveBeenCalledWith(extensionFolderWithExtensionFound.path);
+});
+
+test('Expect untrack action being hidden if there is an extension', async () => {
```
feloy:
```suggestion
test('Expect untrack action being hidden if there is a started extension', async () => {
```

feloy (approved):

packages/renderer/src/lib/extensions/ExtensionList.svelte
```diff
@@ -104,6 +105,12 @@ let installManualImageModal: boolean = false;
         screen = 'catalog';
       }}
       selected={screen === 'catalog'}>Catalog</Button>
+      <Button
+      type="tab"
+      on:click={(): void => {
+        screen = 'development';
+      }}
+      selected={screen === 'development'}>Local extensions</Button>
```
deboer-tim: `Extensions` should be capitalized.

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionEmptyScreen.svelte
```diff
@@ -0,0 +1,17 @@
...
+import { Button, EmptyScreen } from '@podman-desktop/ui-svelte';
+
+async function openExtensionDocumentation(): Promise<void> {
+  await window.openExternal('https://podman-desktop.io/docs/extensions/developing');
+}
+</script>
+
+<EmptyScreen
+  icon={faCog}
+  title="Usage of local extensions (developer mode)"
```
deboer-tim: Suggestion: just 'Local extensions (developer mode)'

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionEmptyScreen.svelte
```diff
@@ -0,0 +1,17 @@
...
+  message="Please enable development mode feature in the preferences">
```
deboer-tim: We don't normally say please and doesn't hurt to be more explicit. Suggest something like 'Enable Preferences > Extensions > Development Mode to test local extensions'.

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionEmptyScreen.svelte
```diff
@@ -0,0 +1,17 @@
...
+  <div class="flex gap-2 justify-center">
+    <Button type="link" on:click={openExtensionDocumentation}>Documentation: How to write your first extension</Button>
```
deboer-tim: Suggestion - we don't normally label links that go to docs and it is obvious enough, remove the word Documentation.

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
```diff
@@ -0,0 +1,106 @@
...
+  for (const unsubscriber of unsubscribers) {
+    unsubscriber();
+  }
+});
+
+async function addLocalFolderExtension(): Promise<void> {
+  // call the openDialog
+  const result = await window.openDialog({
+    selectors: ['openDirectory'],
+    openLabel: 'Select this folder as extension folder',
```
deboer-tim: Too long for a button, just 'Select'.

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
```diff
@@ -0,0 +1,106 @@
...
+}
+</script>
+
+{#if isDevelopmentModeEnabled}
+  <div class="bg-[var(--pd-content-card-bg)] m-2 w-full space-y-2 p-4 rounded-lg">
+    <div class="flex flex-col">
+      <!-- List all -->
+      <div class="bg-[var(--pd-content-bg)] rounded flex flex-col p-2">
+        <div class="flex flex-row w-full">
+          <div class="text font-medium first-letter:uppercase pb-2">Current local extension folders being tracked:</div>
```
deboer-tim: Suggestion: remove 'Current', it is obvious/redundant.

Suggestion: maybe 'monitored' or 'watched' is better? (both here and in the dialog title)

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionListTable.svelte
```diff
@@ -0,0 +1,77 @@
...
+
+const nameColumn = new TableColumn<SelectableExtensionDevelopmentFolderInfoUI, string>('Name', {
+  width: '3fr',
+  renderer: TableSimpleColumn,
+  renderMapping: (extensionFolder): string => extensionFolder.name,
+  comparator: (a, b): number => a.name.localeCompare(b.name),
+});
+
+const extensionNameColumn = new TableColumn<SelectableExtensionDevelopmentFolderInfoUI, string>('Extension', {
+  width: '2fr',
```
deboer-tim: Suggestion: typically folder names are very long, extension/state just one word. Maybe drop these two to 1fr and/or shrink the action column width? Would have avoided wrapping for me.

deboer-tim (changes requested):
I'm totally willing to believe I was using it wrong, but I could not get this to work. My first attempt was in watch mode with the bootc extension that was (inadvertently) already installed. The page look like it worked but then I didn't see any change in the extension, and when I removed it the extension still showed up under Installed, but was listed as a stopped, built-in extension, that I couldn't start nor remove. I had to restart Podman Desktop to clear it.

I tried removing bootc and ai-lab extensions, restarting to make sure things were clear, and trying again. I couldn't get either to appear. Sometimes the state would be blank after adding, sometimes 'starting'. Often even after removing them I'd go to Installed and see them still listed, in the same state as above: stopped, built-in.

Usually there is no feedback when I try to start, but once I had bootc fail to start, and another time I would get `Extension (redhat.bootc) not activated because it is disabled` on both extensions.

I tried loading the root folder of bootc just to see what happens - it looked the same in the UI even though there's not really anything there; not sure if we can do anything to avoid this/warn the user.

Added some UI suggestions inline. I noticed there was a push while I was typing this out, so I'll try again.

deboer-tim:
Good news: I updated to the latest, it worked! 🎉

Bad news: I found three issues and now it doesn't work at all again:
- When removing via the Local Extensions tab, things aren't totally cleaned up - the extension is still visible under Installed.
- You have to stop to be able to remove the extension, and that state isn't cleared when the extension is removed. As a result, the second time you add the extension it is still stopped and has to be manually started. IMHO we should clear this state when you remove an extension so that adding always has the same behaviour.
- After removing an extension from this page (and it still listed under Installed, as stopped) I thought I would restart to see if that cleared it. The extension doesn't appear under installed anymore, but if I try to add the folder again the extension is listed as 'Unknown' and the logs say 'Extension (redhat.ai-lab) not activated because it is disabled'. So disabled state is persisted (confirmed by looking at settings.json).

Good news: It's probably a single cleanup issue and explains what went wrong the first time.

dgolovin (commented):
Why use different UI if the same UI from Installed can be used? Than would let dev to see how it would look like with name and description at production.

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionEmptyScreen.svelte
benoitf: fixed

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionEmptyScreen.svelte
benoitf: fixed

packages/renderer/src/lib/extensions/ExtensionList.svelte
benoitf: fixed

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionEmptyScreen.svelte
benoitf: fixed

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
benoitf: fixed

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
benoitf: fixed

benoitf:
@deboer-tim updated

deboer-tim:
I don't know if I'm missing something 🤷🏼‍♂️, but after updating I can't get it to work:
- If I pick a folder like .../packages/backend it doesn't seem to pick it up correctly / doesn't start.
- If I pick the root folder then the extension starts, but content not loaded correctly (e.g. no webview or other contributions).
- If I stop/delete then I can't re-add an extension until I restart PD.

benoitf:
hello, issues should have been fixed

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
```diff
@@ -0,0 +1,106 @@
...
+}
+</script>
+
+{#if isDevelopmentModeEnabled}
+  <div class="bg-[var(--pd-content-card-bg)] m-2 w-full space-y-2 p-4 rounded-lg">
+    <div class="flex flex-col">
+      <!-- List all -->
+      <div class="bg-[var(--pd-content-bg)] rounded flex flex-col p-2">
+        <div class="flex flex-row w-full">
+          <div class="text font-medium first-letter:uppercase pb-2">Local extension folders being tracked:</div>
```
deboer-tim: Nit: remove ':', isn't used on other pages.

packages/renderer/src/lib/extensions/dev-mode/table/ListTable.svelte
```diff
@@ -0,0 +1,77 @@
...
+const extensionNameColumn = new TableColumn<SelectableExtensionDevelopmentFolderInfoUI, string>('Extension', {
+  width: '2fr',
+  renderer: TableSimpleColumn,
+  renderMapping: (extensionFolder): string => extensionFolder.extension?.name ?? 'Unknown',
+  comparator: (a, b): number => a.name.localeCompare(b.name),
+});
+const extensionStateColumn = new TableColumn<SelectableExtensionDevelopmentFolderInfoUI, string>('State', {
+  width: '2fr',
+  renderer: TableSimpleColumn,
+  renderMapping: (extensionFolder): string => extensionFolder.extension?.state ?? '',
```
deboer-tim, line 32: Extensions use 'active/disabled' in other places, here it uses 'started/stopped'. Should really be consistent, but follow-up PR fine.

deboer-tim (approved):
Changes all worked perfectly. it correctly rejected several incorrect folders and picked up the correct one. Starting/stopped worked consistently now with no obvious leaks/incomplete shutdown.

We still need to document that you are still responsible for 'pnpm watch' or 'build' (and would even be nice to automate this/yarn in future issue) but this is a huge improvement. Nicely done!

packages/renderer/src/lib/extensions/dev-mode/DevelopmentExtensionList.svelte
benoitf: fixed

PR #10742, opened 2025-01-21
refactor: extract analyzeExtension to a dedicated class

benoitf:
### What does this PR do?
move out analyze extension from extension-loader to a separate class

related to https://github.com/podman-desktop/podman-desktop/pull/10712#discussion_r1920355635

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extension/extension-analyzer.spec.ts
+++ b/packages/main/src/plugin/extension/extension-analyzer.spec.ts
@@ -36,7 +36,7 @@ beforeEach(() => {
   extensionAnalyzer = new ExtensionAnalyzer();
 });
 
-describe('analyze extension and main', async () => {
+describe('analyze extension and main', () => {
   test('check for extension with main entry', async () => {
     // mock fs.existsSync
     const fsExistsSyncMock = vi.spyOn(fs, 'existsSync');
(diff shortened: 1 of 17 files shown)


packages/main/src/plugin/index.ts
```diff
@@ -711,9 +718,9 @@ export class PluginSystem {
       certificates,
       extensionWatcher,
       extensionDevelopmentFolders,
+      extensionAnalyzer,
     );
     await this.extensionLoader.init();
-    extensionDevelopmentFolders.setExtensionLoader(this.extensionLoader);
```
axel7083, line 716: Nice, thanks for removing the circular dependency

packages/main/src/plugin/extension/extension-analyzer.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+
+vi.mock('node:fs');
+vi.mock('node:fs/promises');
+
+beforeEach(() => {
+  vi.resetAllMocks();
+  extensionAnalyzer = new ExtensionAnalyzer();
+});
+
+describe('analyze extension and main', async () => {
```
axel7083:
```suggestion
describe('analyze extension and main', () => {
```

describe don't need to be async?

axel7083 (approved):
Tested and LGTM, thanks for the refactoring 🚀

packages/main/src/plugin/extension/extension-analyzer.spec.ts
benoitf: removed, but it was a move of the existing content from extension-loader.spec.ts file

gastoner (approved):

Issue #10600, opened 2025-01-10
Remove disablement of no-explicit any from api/main packages


jeffmaury:
### Is your enhancement related to a problem? Please describe

Impacted files:

- ~~packages/api/src/container-info.ts~~
- ~~packages/api/src/onboarding.ts~~
- ~~packages/main/src/application-menu-builder.spec.ts~~
- ~~packages/main/src/electron-util.spec.ts~~
- ~~packages/main/src/mainWindow.ts~~
- ~~packages/main/src/plugin/appearance-init.spec.ts~~
- ~~packages/main/src/plugin/cancellation-token-registry.spec.ts~~
- ~~packages/main/src/plugin/cancellation-token.spec.ts~~
- ~~packages/main/src/plugin/cancellation-token.ts~~
- ~~packages/main/src/plugin/color-registry.spec.ts~~
- ~~packages/main/src/plugin/command-registry.ts~~
- ~~packages/main/src/plugin/configuration-impl.ts~~
- ~~packages/main/src/plugin/configuration-registry.ts~~
- ~~packages/main/src/plugin/container-registry.spec.ts~~
- ~~packages/main/src/plugin/container-registry.ts~~
- ~~packages/main/src/plugin/context/context.ts~~
- ~~packages/main/src/plugin/contribution-manager.spec.ts~~
- ~~packages/main/src/plugin/contribution-manager.ts~~
- ~~packages/main/src/plugin/dialog-registry.spec.ts~~
- ~~packages/main/src/plugin/docker/docker-compatibility.spec.ts~~
- ~~packages/main/src/plugin/docker-extension/docker-desktop-installation.spec.ts~~
- ~~packages/main/src/plugin/docker-extension/docker-desktop-installation.ts~~
- ~~packages/main/src/plugin/docker-extension/docker-plugin-adapter.ts~~
- ~~packages/main/src/plugin/dockerode/libpod-dockerode.spec.ts~~
- ~~packages/main/src/plugin/dockerode/libpod-dockerode.ts~~

### Describe the solution you'd like

Remove use of any type

PR #10749, opened 2025-01-21
chore: remove no-explicit-any

benoitf:
### What does this PR do?
remove no-explicit-any

related to https://github.com/podman-desktop/podman-desktop/issues/10600

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/cancellation-token.ts
+++ b/packages/main/src/plugin/cancellation-token.ts
@@ -21,8 +21,7 @@ import type * as extensionApi from '@podman-desktop/api';
 import { Emitter } from './events/emitter.js';
 import type { IDisposable } from './types/disposable.js';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const shortcutEvent: extensionApi.Event<any> = Object.freeze(function (callback, context?): IDisposable {
+const shortcutEvent: extensionApi.Event<unknown> = Object.freeze((callback, context?): IDisposable => {
   const handle = setTimeout(callback.bind(context), 0);
   return {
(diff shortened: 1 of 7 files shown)


packages/main/src/plugin/cancellation-token.ts
```diff
@@ -33,8 +32,7 @@ const shortcutEvent: extensionApi.Event<any> = Object.freeze(function (callback,
 
 export class CancellationTokenImpl implements extensionApi.CancellationToken {
   private _isCancellationRequested: boolean;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  private emitter: Emitter<any> | undefined;
+  private emitter: Emitter<unknown> | undefined;
```
axel7083, line 35: We only emit `undefined`

https://github.com/podman-desktop/podman-desktop/blob/0698e4d2fd1f3fcb64bfc8d66a96003545b6ee89/packages/main/src/plugin/cancellation-token.ts#L47

Should we use `<void>` instead ? so we can simplify 

If we use void, we may change the extension-api too ?

https://github.com/podman-desktop/podman-desktop/blob/be499999e49627d74cad14efe98c9dbe3fed1111/packages/extension-api/src/extension-api.d.ts#L1315

gastoner (approved):
Codewise LGTM

packages/main/src/plugin/cancellation-token.ts
benoitf: we won't be able to change anything in the extension-api.d.ts

axel7083 (approved):

Issue #10632, opened 2025-01-14
dispatch kube resources to frontend in Kubernetes experimental mode


feloy:
### Is your feature request related to a problem? Please describe

On Kubernetes Experimental mode, we need to display the Kubernetes resources collected by the informers

### Describe the solution you'd like

Collect the kubernetes resources from the experimental kubernetes manager, and diplay them in the Kubernetes pages

feloy:
reopening, the frontend should be able to ask resources for specific contexts only

PR #10760, opened 2025-01-22
refactor: deployments list uses svelte 5

feloy:

### What does this PR do?

Refactor deployments list component to use Svelte 5

This will be needed for #10632 

Other PRs will come for other resources

### Screenshot / video of UI

No changes

Part of #10632

### How to test this PR?

Go to the page Kubernetes > Deployments, do different actions

### DIFF
--- a/packages/renderer/src/lib/deployments/DeploymentsList.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentsList.svelte
@@ -10,7 +10,6 @@ import {
   TableRow,
 } from '@podman-desktop/ui-svelte';
 import moment from 'moment';
-import { onMount } from 'svelte';
 
 import KubeActions from '/@/lib/kube/KubeActions.svelte';
 import KubernetesCurrentContextConnectionBadge from '/@/lib/ui/KubernetesCurrentContextConnectionBadge.svelte';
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/deployments/DeploymentsList.svelte
```diff
@@ -30,21 +29,23 @@ import DeploymentColumnStatus from './DeploymentColumnStatus.svelte';
...
-  });
+$effect(() => {
+  deploymentSearchPattern.set(searchTerm);
 });
 
+const deploymentUtils = new DeploymentUtils();
+const deployments = $derived(
+  $kubernetesCurrentContextDeploymentsFiltered.map(deployment => deploymentUtils.getDeploymentUI(deployment)),
+);
+
```
benoitf, line 46: AFAIK derived objects from $store pattern are not refreshed when the store is updated

feloy: It seems that it works, as the deployments are refreshed when the search term is modified. Also, this outputs logs when the search term is modified:

```
const deployments = $derived.by(() => {
    console.log('==> derive');
    return $kubernetesCurrentContextDeploymentsFiltered.map(deployment => deploymentUtils.getDeploymentUI(deployment));
  }
);
```

benoitf: I'm saying that if kubernetesCurrentContextDeploymentsFiltered is updated, deployments won't be refreshed

feloy: Yes, I understand. When searchTerm is modified, kubernetesCurrentContextDeploymentsFiltered is updated, and I can see that the deployments list is modified.

Another example where `$derived($store)` seems to work correctly: https://svelte.dev/playground/a686354d4c9f4ce0bdb8ba63672ff775?version=5.19.1

Or am I missing something?

axel7083: > AFAIK derived objects from $store pattern are not refreshed when the store is updated

On my side it seems it is also refreshed on other example

benoitf: good to know, it must have been fixed in some new updates of svelte v5

please add a test where kubernetesCurrentContextDeploymentsFiltered is being updated while the component is already rendered initially to ensure we'll have no regression on that part

feloy: done, I had to refactor the unit tests, to mock the stores directly, to be able to add this test

benoitf (approved):

axel7083 (approved):
LGTM, thanks for the refactor

PR #10768, opened 2025-01-22
feat: adding version to container provider connection

### LINKED ISSUE #4215: Podman Version reported on a podman machine under Resources is wrong or confussing
### Bug description

There is a podman version 4.7.0 mentioned under my "Podman-2" machine on Podman Desktop (Resources) which is not correct as I have a podman 4.6.2 installed inside the VM. Or It is confusing if that is a client version of a podman.

Screenshot_20231005_104308

### Operating system

macOS 13 (Intel)

### Installation Method

Brew (macOS)

### Version

next (development version)

### Steps to reproduce

1. Install any podman available through the installer or manually, podman client should be 4.7.0
2. Use a custom image for a VM to be spawned (image-path), ie.: https://builds.coreos.fedoraproject.org/browser?stream=stable&arch=x86_64 (podman package is in version 4.6.2)
3. Create a podman machine using this image
4. verify that Podman reported is the one used in the image (`podman machine ssh, podman version or podman version on cli`)
Actual result: Podman version suggest 4.7.0, should be 4.6.2 if that is a server side version.

### Relevant log output

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Podman machines may have a different version that the CLI, and several machines may have different version too.

Solution proposal for https://github.com/podman-desktop/podman-desktop/issues/4215 

See https://github.com/podman-desktop/podman-desktop/issues/4215#issuecomment-2607178592 for more details

### Screenshot / video of UI

image

Fixes https://github.com/podman-desktop/podman-desktop/issues/4215

### How to test this PR?

**Testing manually**

1. assert have a old podman machine ( machine < 5.4 )
2. install podman 5.4 (cli)
3. go to podman desktop
4. assert the version on the card is the expected version (should match `podman machine ssh "podman --version"`)

### DIFF
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -76,6 +76,8 @@ let defaultMachineMonitor = true;
 
 // current status of machines
 export const podmanMachinesStatuses = new Map<string, extensionApi.ProviderConnectionStatus>();
+
+const podmanServerVersions = new Map<string, string>();
 let podmanProviderStatus: extensionApi.ProviderConnectionStatus = 'started';
 const podmanMachinesInfo = new Map<string, MachineInfo>();
 const currentConnections = new Map<string, extensionApi.Disposable>();
@@ -277,6 +279,13 @@ async function doUpdateMachines(
 
     const previousStatus = podmanMachinesStatuses.get(machine.Name);
     if (previousStatus !== status) {
+      if (status === 'started') {
+        await updatePodmanServer({
+          name: machine.Name,
+          vmType: machine.VMType,
+        });
+      }
+
       // notify status change
       listeners.forEach(listener => listener(machine.Name, status));
       podmanMachinesStatuses.set(machine.Name, status);
@@ -768,6 +777,28 @@ function prettyMachineName(machineName: string): string {
   return name;
 }
 
+async function updatePodmanServer(machineInfo: { name: string; vmType?: string }, force = false): Promise<void> {
+  // do not check if we already have
+  if (podmanServerVersions.has(machineInfo.name) && !force) return;
+
+  let serverVersion: string | undefined;
+  try {
+    const { stdout } = await execPodman(
+      ['--connection', machineInfo.name, 'version', '--format=json'],
+      machineInfo.vmType,
+    );
+    const parsed = JSON.parse(stdout);
+    serverVersion = parsed['Server']['Version'];
+  } catch (err: unknown) {
+    console.error(`Something went wrong while trying to get machine ${machineInfo.name} version`, err);
+  }
+
+  if (!serverVersion) return;
+
+  // update the version
+  podmanServerVersions.set(machineInfo.name, serverVersion);
+}
+
 export async function registerProviderFor(
   provider: extensionApi.Provider,
   podmanConfiguration: PodmanConfiguration,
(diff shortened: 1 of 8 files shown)


extensions/podman/packages/extension/src/extension.ts
```diff
@@ -761,6 +770,33 @@ function prettyMachineName(machineName: string): string {
   return name;
 }
 
+/**
+ * This function will check for the version of a podman connection
```
feloy:
```suggestion
 * This function will check for the version of a podman connection and populate podmanServerVersions with it
```

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -1892,8 +1929,15 @@ export async function getJSONMachineListByProvider(containerMachineProvider?: st
...
-  stopLoop = true;
   console.log('stopping podman extension');
+  // cleanup
+  stopLoop = true;
+  listeners.clear();
+  podmanServerVersions.clear();
+  podmanMachinesInfo.clear();
+  currentConnections.clear();
+  containerProviderConnections.clear();
+
```
feloy: Is this related to this issue (except podmanServerVersions.clear())?

axel7083: missing a rebase 👍 should be good now

feloy (approved):

packages/extension-api/src/extension-api.d.ts
```diff
@@ -401,6 +401,10 @@ declare module '@podman-desktop/api' {
      * the vmTypeDisplayName property cannot be set if vmType is undefined
      */
     vmTypeDisplayName?: string;
+    /**
+     * the remote may be different from the client version
+     */
+    version?(): string | undefined;
```
benoitf, line 407: Question: I don't get why it is a method ?

why not like vmType ? version?: string

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -551,7 +551,7 @@ function hasAnyConfiguration(provider: ProviderInfo): boolean {
               <div class="mt-1.5 text-[var(--pd-content-sub-header)] text-[9px] flex justify-between">
                 <div aria-label="Connection Version">
                   {provider.name}
-                  {provider.version ? `v${provider.version}` : ''}
+                  {container.version ?? ''}
```
benoitf, line 554: issue: I expect to see both versions

benoitf (changes requested):
question in the api.d.ts

also I think the PR is too big

AFAIK, there is separate work:
- provide a new optional version attribute
- implement the part in podman extension
- change UI components to display this new attribute

so at least 3 PRs

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
axel7083: But this wouldn't this be inaccurate to fallback on the provider version?

packages/extension-api/src/extension-api.d.ts
axel7083: So the problem is the same as the `status`

https://github.com/podman-desktop/podman-desktop/blob/d091c29966f7f94ea3681145fdacdb9ebe64a66f/packages/extension-api/src/extension-api.d.ts#L398

The version cannot be fetch if the machine is not running, so I used the same logic as the status, to be able to get it from the main when the extension trigger an update event

jeffmaury: Can't we have it as a property that will get lazy initialized when machine is started so that we can get the info even if the machine is stopped ?

axel7083:
Was not able to go further with https://github.com/podman-desktop/podman-desktop/pull/11399, need discussion on the issue

PR #10781, opened 2025-01-22
docs: adding remark to containerEngine#listInfos

axel7083:
### What does this PR do?

I noticed that the `containerEngine#listInfos` can only work on running connections.  Looking at the implementation this make sense, but as an external user this could be misleading, I was getting `no engine matching this container` error throwed when calling it before understanding the issue.

So adding a small remark to the doc!

### Screenshot / video of UI

See https://67911d0c0155cd18ad095771--podman-desktop-pr.netlify.app/api/namespaces/containerEngine/functions/listInfos

image

N/A

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -3905,7 +3905,7 @@ declare module '@podman-desktop/api' {
      * @param options optional options for listing information
      * @returns A promise resolving to an array of engine information.
      *
-     * @remarks only return the {@link ContainerEngineInfo} of **running** connection
+     * @remarks only returns the {@link ContainerEngineInfo} of **running** connections
      * @example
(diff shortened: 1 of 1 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -3905,6 +3905,7 @@ declare module '@podman-desktop/api' {
      * @param options optional options for listing information
      * @returns A promise resolving to an array of engine information.
      *
+     * @remarks only return the {@link ContainerEngineInfo} of **running** connection
```
SoniaSandler:
```suggestion
     * @remarks only returns the {@link ContainerEngineInfo} of **running** connections
```

benoitf (approved):
nit: needs to fix English suggested by @SoniaSandler

PR #10791, opened 2025-01-23
docs: added goatcounter for download buttons

gastoner:
### What does this PR do?
Adds goatcounter (telemetry) for download button clicks on website

Closes [#Gather metrics](https://github.com/containers/podman-desktop-internal/issues/432)

### How to test this PR?


benoitf (dismissed):
goat counter is already included so I don't see why it needs to be imported again

Import on '..' is also odd

gastoner:
- [x] TODO: Update paths/events properties

benoitf (commented):
to avoid to import functions in a lot of place

I would suggest to introduce a new Link directive like `<TelemetryLink` or `TrackLink` etc

And the usage of goat counter is only inside this directive

website/src/components/TelemetryLink.tsx
```diff
@@ -0,0 +1,29 @@
+import Link from '@docusaurus/Link';
+
+type TelemetryLinkProps = {
+  to: string;
```
axel7083, line 5: please use `href` as it is a _standard_

website/src/components/TelemetryLink.tsx
```diff
@@ -0,0 +1,29 @@
+import Link from '@docusaurus/Link';
```
axel7083, line 1: missing copyright?

gastoner: I don't see it in other tsx files, but I can add it

gastoner:
@benoitf @axel7083 PR updated

axel7083 (approved):

benoitf (commented):
question: why a new `<span>` element is being added ?

gastoner:
> question: why a new `<span>` element is being added ?

Because there must be only one child element: 
`This JSX tag's 'children' prop expects a single child of type 'Element | undefined', but multiple children were provided.`

benoitf:
@gastoner but Link is a tsx/jsx element from Docusaurus and it works
https://github.com/facebook/docusaurus/blob/d025403f73cf3bb5fd6130f50111c3a5e0aa7812/packages/docusaurus/src/client/exports/Link.tsx


AFAIK it's you that is saying that you want a single child there: https://github.com/podman-desktop/podman-desktop/blob/293b727ab645fb5ce6398226e52e9961494794ea/website/src/components/TelemetryLink.tsx#L7

benoitf:
@axel7083 why did you ask to use href

Link is using the `to` directive
https://docusaurus.io/docs/docusaurus-core#to-string

website/src/components/TelemetryLink.tsx
benoitf: AFAIK `to` is the standard there https://docusaurus.io/docs/docusaurus-core#to-string

axel7083:
> @axel7083 why did you ask to use href
> 
> Link is using the `to` directive https://docusaurus.io/docs/docusaurus-core#to-string

My bad, I didn't know they were using `to`.

gastoner:
> @axel7083 why did you ask to use href
> 
> Link is using the `to` directive https://docusaurus.io/docs/docusaurus-core#to-string

Switching back to `to` then

benoitf (commented):
nit: could you move all the eventTitle attributes before the to directive so the diff will be reduced ? (between className and to)

then I need to try it out live to be sure it's sending the attribute before approving

benoitf:
issue1: trying the website in dev mode I got

```
Cannot read properties of undefined (reading 'count')
TypeError: Cannot read properties of undefined (reading 'count')
    at sendGoatCounterEvent (webpack-internal:///./src/components/TelemetryLink.tsx:7:55)
```

It's related to the fact that the plugin is not injecting stuff in dev mode
https://github.com/Addono/docusaurus-plugin-goatcounter
```
When building your website, make sure that NODE_ENV=production, as the plugin will avoid injecting the Goatcounter tag in development environments.
```

it means that TelemetryLink should not try to send data in dev mode (can log the stuff or do not emit it)


issue2: it looks the data are missing like the path ? you're saying it's added automatically but if I look at the console I see
image
meaning I see 'no event name'



issue3: event name `landing-download-button-clicked`

I guess I would expect the path should be LandingPage and the event could be shorter like download-button ? I mean the clicked is an action. Or it's the opposite we send a 'downloads' path with a name "from-landing-page"

so I could compare from where come people looking at "downloads-from" path


looking at event being received I'm figuring out how to quickly find them
image

gastoner:
> issue1: trying the website in dev mode I got
> 
> ```
> Cannot read properties of undefined (reading 'count')
> TypeError: Cannot read properties of undefined (reading 'count')
>     at sendGoatCounterEvent (webpack-internal:///./src/components/TelemetryLink.tsx:7:55)
> ```
> 
> It's related to the fact that the plugin is not injecting stuff in dev mode https://github.com/Addono/docusaurus-plugin-goatcounter
> 
> ```
> When building your website, make sure that NODE_ENV=production, as the plugin will avoid injecting the Goatcounter tag in development environments.
> ```
> 
> it means that TelemetryLink should not try to send data in dev mode (can log the stuff or do not emit it)
> 
> issue2: it looks the data are missing like the path ? you're saying it's added automatically but if I look at the console I see image meaning I see 'no event name'
> 
> issue3: event name `landing-download-button-clicked`
> 
> I guess I would expect the path should be LandingPage and the event could be shorter like download-button ? I mean the clicked is an action. Or it's the opposite we send a 'downloads' path with a name "from-landing-page"
> 
> so I could compare from where come people looking at "downloads-from" path
> 
> looking at event being received I'm figuring out how to quickly find them image

1. I'll add some check to not send data when in dev mode
2. I'm afraid that I can't do anythink about this in this PR
at first iteration I was hardcoding the path (`/` and `/download`) but the goatcounter for some reason removes the `/` => which results in (no-event-name) and (download). After removing the path/title entirely from code (goatcounter should use path/title based on page which should be deafult). Problem is that the goatcounter is aggregating this somehow based on event-name and title/path 

This means that it should be `/` as path and `landing-download-button-clicked`, and other should be `/download` + download-windows/mac/linux

I think we need to just remove the history of those events to fix this 

3. This was chosen from a discussion with @slemeur described in the issue

benoitf:
from the documentation the path of an event can't have '/' as first character

https://www.goatcounter.com/help/events#sending-events-from-javascript-374

gastoner:
So in the case of a landing page, I would add a parameter "landing"? and the rest would be "download/..."

benoitf:
> So in the case of a landing page, I would add a parameter "landing"? and the rest would be "download/..."

Here I would say how you see things. It seems you're seing it as landing event saying to go to the download page while I see a download page event with a from event saying 'hero-button'

now, on the logic.

I think I would be interested to know if people are more clicking on the hero button or the navigation bar button (so I would like to track also the people clicking on the 'Downloads' in the navigation bar and what is the name page when they click on the button

gastoner:
> > So in the case of a landing page, I would add a parameter "landing"? and the rest would be "download/..."
> 
> Here I would say how you see things. It seems you're seing it as landing event saying to go to the download page while I see a download page event with a from event saying 'hero-button'
> 
> now, on the logic.
> 
> I think I would be interested to know if people are more clicking on the hero button or the navigation bar button (so I would like to track also the people clicking on the 'Downloads' in the navigation bar and what is the name page when they click on the button

I agree with monitoring also the nav bar, I've just followed the instruction in the issue from @slemeur but I can add it for sure

slemeur:
We can add a tracking event for the "download" in the navigation bar. 

About the naming of the event, if we can add the path, that will indeed be ideal to easily identify the event in goatcounter. 
In that case, I'd propose the following.

There will be 2 events for the landing page:
- `hero-download` : to track each time a user is clicking on the hero "Download" button on the landing page 
- `navigation-download` : to track each time a user is clicking on the "Downloads" link in the navigation bar

Then 3 events for the download page:
- `download-mac` : all clicks on the download links for mac binaries
- `download-windows`: all clicks on the download links for windows
- `download-linux`: all clicks on the download links for linux
(here, we will not differentiate between the different binaries available, as that can be tracked elsewhere via the github releases numbers)

gastoner:
@benoitf @slemeur updated!

website/docusaurus.config.js
```diff
@@ -420,7 +420,15 @@ const config = {
             label: 'Documentation',
           },
           { to: '/features', label: 'Features', position: 'left' },
-          { to: '/downloads', label: 'Downloads', position: 'left' },
+          {
+            type: 'custom-telemetryLink',
+            position: 'left',
+            children: 'Downloads',
+            className: 'navbar__item navbar__link',
```
benoitf: question: why do we need to add children and className which are not added to the previous items ?

gastoner: If I don't specify this explicitly, the outcome will be 
image
-> Without a class and any text
Also this item is a custom item

benoitf: my point is that here, I should not add any other customization in docusaurus config file

if I need to add a specific className, then it's the overridden MDX component that should add this className before sending it to the TelemetryLink


docusaurus.config.js: only the data about the events (no className, no children)

customization of the navbaritem: here it does all the logic handling the existing param and if missing, add any class/attributes about navbar item

TelemetryLink: do only the telemetry stuff

slemeur (approved):

cdrage (changes requested):
When testing this on Netlify I'm unable to click on Downloads. So it prevents us from ever previewing that page...

It also means that we may run into this issue too in production when it's deployed merged in?

I also saw errors with my adblock enabled (unable to load count.js).


Screenshot 2025-02-10 at 8 53 31 AM




https://github.com/user-attachments/assets/81851088-0ba2-4117-b899-b21987c8f1c5

Screenshot 2025-02-10 at 8 55 51 AM

gastoner:
> When testing this on Netlify I'm unable to click on Downloads. So it prevents us from ever previewing that page...
> 
> It also means that we may run into this issue too in production when it's deployed merged in?
> 
> I also saw errors with my adblock enabled (unable to load count.js).
> 
> Screenshot 2025-02-10 at 8 53 31 AM
> Screen.Recording.2025-02-10.at.8.53.41.AM.mov
> 
> Screenshot 2025-02-10 at 8 55 51 AM

Hmm that's weird, on my machine it works: 

https://github.com/user-attachments/assets/efcd26fc-9135-40bc-bee9-1dd7de0abd6b

cdrage:
> > When testing this on Netlify I'm unable to click on Downloads. So it prevents us from ever previewing that page...
> > It also means that we may run into this issue too in production when it's deployed merged in?
> > I also saw errors with my adblock enabled (unable to load count.js).
> > Screenshot 2025-02-10 at 8 53 31 AM
> > Screen.Recording.2025-02-10.at.8.53.41.AM.mov
> > Screenshot 2025-02-10 at 8 55 51 AM
> 
> Hmm that's weird, on my machine it works:
> 
>  Screen.Recording.2025-02-10.161005.mp4

Did you try on Netlify?

I'm using Chrome & Ublock adblocker.

Enable and disabled, still unable to get to the page + get the javascript error.

gastoner:
> > > When testing this on Netlify I'm unable to click on Downloads. So it prevents us from ever previewing that page...
> > > It also means that we may run into this issue too in production when it's deployed merged in?
> > > I also saw errors with my adblock enabled (unable to load count.js).
> > > Screenshot 2025-02-10 at 8 53 31 AM
> > > Screen.Recording.2025-02-10.at.8.53.41.AM.mov
> > > Screenshot 2025-02-10 at 8 55 51 AM
> > 
> > 
> > Hmm that's weird, on my machine it works:
> > Screen.Recording.2025-02-10.161005.mp4
> 
> Did you try on Netlify?
> 
> I'm using Chrome & Ublock adblocker.
> 
> Enable and disabled, still unable to get to the page + get the javascript error.

Yes, I'm using netlify, and I'm using firefox + ublock orogin
I've tried something and it shows on my side too now 
But without any error message (both on Firefox/chrome )

gastoner:
@cdrage Hopefully, it should be ok now, can you please check again?

slemeur:
Works on my side as well, testing with Netlify

benoitf (changes requested):
if I read https://www.goatcounter.com/help/events#sending-events-from-javascript-374

it says:
```
The path doubles as the event name. This cannot have / as the first character.
```

but in the code we send as path the eventPath attribute

and all attributes are like

eventPath: '/download',
eventPath="/landing"

they start with `/` as first character

cdrage (approved):
Works well for me now + I'm able to access it.

benoitf (approved):
checked on netlify and goatcounter

PR #10773, opened 2025-01-22
chore(test): run Kubernetes e2e tests with Docker provider on GitHub Actions Linux runner

amisskii:
Signed-off-by: Anton Misskii <amisskii@redhat.com>

### What does this PR do?

Fixes an issue with Kubernetes E2E tests on the GitHub Actions Linux runner and refactors the cluster parameterized creation method.

### What issues does this PR fix or reference?

https://github.com/podman-desktop/podman-desktop/issues/10716

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature

PR #10797, opened 2025-01-23
chore: add Dropdown options aria-label

amisskii:

### What does this PR do?

Adds an aria-label for a dropdown options div.. Related PRs: https://github.com/podman-desktop/podman-desktop/pull/10773, https://github.com/podman-desktop/extension-minikube/pull/355

### DIFF
--- a/packages/ui/src/lib/dropdown/Dropdown.svelte
+++ b/packages/ui/src/lib/dropdown/Dropdown.svelte
@@ -210,7 +210,7 @@ function onWindowClick(e: Event): void {
 
   {#if opened}
     <div
-      class="absolute top-full right-0 z-10 w-full max-h-80 rounded-md bg-[var(--pd-dropdown-bg)] border-[var(--pd-input-field-hover-stroke)] border-[1px] overflow-y-auto whitespace-nowrap">
+      class="absolute top-full right-0 z-10 w-full max-h-80 rounded-md bg-[var(--pd-dropdown-bg)] border-[var(--pd-input-field-hover-stroke)] border-[1px] overflow-y-auto whitespace-nowrap" aria-label={ariaLabel + ' Options'}>
       {#each options as option, i}
         <button
           onkeydown={onKeyDown}


benoitf (commented):
hello, not sure why we would need that ?

benoitf:
each option being a button you can search for a button having the option name being specified

axel7083:
> each option being a button you can search for a button having the option name being specified

@benoitf  I was not able to do that I think name only work for aria-label, when I added test for PullImage I had to do the following (not really nice IMO)

https://github.com/podman-desktop/podman-desktop/blob/4601f9fa0ad305eb5549fa1cb5fc1938666319d6/packages/renderer/src/lib/image/PullImage.spec.ts#L454-L459

packages/ui/src/lib/dropdown/Dropdown.svelte
```diff
@@ -210,7 +210,7 @@ function onWindowClick(e: Event): void {
 
   {#if opened}
     <div
-      class="absolute top-full right-0 z-10 w-full max-h-80 rounded-md bg-[var(--pd-dropdown-bg)] border-[var(--pd-input-field-hover-stroke)] border-[1px] overflow-y-auto whitespace-nowrap">
+      class="absolute top-full right-0 z-10 w-full max-h-80 rounded-md bg-[var(--pd-dropdown-bg)] border-[var(--pd-input-field-hover-stroke)] border-[1px] overflow-y-auto whitespace-nowrap" aria-label={ariaLabel + ' Options'}>
```
axel7083, line 213: Maybe add the aria-label to the button?

benoitf: we don't have like in unit test getByRole('button', {name: 'button name'}) ?

benoitf: it seems it's there https://playwright.dev/docs/locators#locate-by-role

`await page.getByRole('button', { name: /submit/i }).click();`

axel7083: I just re-checked with the following

```ts
--- Example.svelte
<button class="foo">bar</button>
--- Example.spec.ts
import { render } from '@testing-library/svelte';
import { expect, test } from 'vitest';
import Example from '/@/Example.svelte';

test('example', () => {
  const { getByRole } = render(Example);
  const btn = getByRole('button', { name: 'bar'});
  expect(btn).toBeDefined();
});
```

and you are right, it works

amisskii:
Hello @benoitf, I use this in our e2e tests: https://github.com/podman-desktop/podman-desktop/blob/d412e5864b7ccb6ac5664ca545c8f90c21ee896b/tests/playwright/src/model/pages/create-kind-cluster-page.ts#L69-L78. We’ve parameterized the method for creating a kind cluster, but if we use podman as the provider parameter (which is a default provider value) for this method, the tests fail. This happens because two buttons with the same name appear on the page—one opens dropdown options, and the other is inside the options div. We could check if the provider already has the desired value and skip opening the dropdown, but dropdown elements appear in many places. I think it’s better to add an aria-label to distinguish the button inside the dropdown options.

odockal (approved):
LGTM. The solution is not optimal. Although, this way we can at least identify and distinguish between button that opens dropdown and is also a selection, and the buttons (options) in the dropdown itself.

I think there could be much more improvements done, in order to avoid adding labels, like using particular roles in divs or by using html elements that have specific roles by default (ie. select + options). Instead we have divs + buttons.

benoitf:
AFAIK first you select the parent, and then you can search from the parent the child
so you search first the drop-down name and then you select the value

issue: it still feels hacky to me, so I'm requested changes

benoitf (changes requested):
issue: looks like an anti-pattern

PR #10036, opened 2024-11-20
fix: use correct header for play kube operation

benoitf:
### What does this PR do?
there is a change in the latest releases of podman that is checking the headers provided when calling the play/kube operation.
https://github.com/containers/podman/commit/1dd90dbe20c280f8000919cea86c4cf48191a851#diff-5b8eed2974716af4dc1f528414ca9b3127e1b696b722a18b01d7d87dff178644R58

dockerode and docker-modem are assuming that if there is a file as parameter, it needs to be a application/tar content-type while now on the backend, you can submit yaml file or tar file.
For backward compatibility, we use yaml but then we need to provide the yaml content type

### What issues does this PR fix or reference?

fixes https://github.com/podman-desktop/podman-desktop/issues/10010

### How to test this PR?

use test case of the issue (or unit test)

- [x] Tests are covering the bug fix or the new feature

PR #10801, opened 2025-01-23
feat: kube play build support

### LINKED ISSUE #10091: feat: kube play build support

Following https://github.com/podman-desktop/podman-desktop/pull/10036, we support YAML content for the kube play. A new tar format has been added.

Reading the [podman 5.3 changlog](https://github.com/containers/podman/releases/tag/v5.3.0), we can notice the following

> The Play API for Kubernetes YAML now supports application/x-tar compressed context directories (https://github.com/containers/podman/pull/24015).

After looking at https://github.com/containers/podman/pull/24015, the API support to build the containers needed in the Kubernetes resources, this would be incredibly useful for AI Lab, as this could replace our custom format format. 

Since this is only an API, we should support parsing the YAML, and creating a tarball from the contexts required by the pods containers images, and send it to the API.

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Adding build support for kube play. Learn more in https://docs.podman.io/en/latest/markdown/podman-kube-play.1.html#build

### Screenshot / video of UI

image

Fixes https://github.com/podman-desktop/podman-desktop/issues/10091

### How to test this PR?

**Testing manually**

**Requirements**
- **Podman 5.3 and** above

1. Create tmp dir
2. Create play.yaml file
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: demo-build-remote
spec:
  containers:
    - name: container
      image: foobar
      ports:
        - name: http
          containerPort: 80
```
3. Create foobar/Containerfile
```
FROM docker.io/library/nginx:alpine

LABEL "quadlet"="demo-kube"
```
4. Go to **Pods > Play Kubernetes YAML** page
image
5. Select your `play.yaml` file you created
6. Check `build` checkbox
7. Click Play
8. assert pod running
image
9. open container details
image
10. assert image is `localhost/foobar:latest`
11. Open corresponding image
12. assert label is visible
image

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -31,6 +31,7 @@ import type { ContainerAttachOptions, ImageBuildOptions } from 'dockerode';
 import Dockerode from 'dockerode';
 import { loadAll } from 'js-yaml';
 import moment from 'moment';
+import { coerce, gtr } from 'semver';
 import StreamValues from 'stream-json/streamers/StreamValues.js';
 import type { Headers, Pack, PackOptions } from 'tar-fs';
 
@@ -2435,6 +2436,21 @@ export class ContainerProviderRegistry {
     }
   }
 
+  protected async isTarPlayBuildSupported(internalProvider: InternalContainerProvider): Promise<boolean> {
+    if (!internalProvider.api || !internalProvider.libpodApi) {
+      throw new Error('No provider with a running engine');
+    }
+
+    const version = await internalProvider.api.version();
+    console.log('[container-registry] isTarPlayBuildSupported', version);
+
+    // let's nicely format it
+    const coerced = coerce(version.Version);
+    if (!coerced) throw new Error(`api version cannot be coerced ${version.Version}`);
+
+    return gtr(coerced.version, '5.3');
+  }
+
   async playKube(
     kubernetesYamlFilePath: string,
     selectedProvider: ProviderContainerConnectionInfo,
@@ -2453,6 +2469,13 @@ export class ContainerProviderRegistry {
       if (!options?.build) {
         return provider.libpodApi.playKube(kubernetesYamlFilePath);
       }
+
+      // ensure build support is true, otherwise let's throw a nice user friendly error
+      const buildSupported: boolean = await this.isTarPlayBuildSupported(provider);
+      if (!buildSupported)
+        throw new Error(
+          `kube play build is not supported on ${provider.connection.name}: Podman 5.3 and above support this feature`,
+        );
 
       const content = await readFile(kubernetesYamlFilePath, 'utf8');
       const resources = loadAll(content);
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -2471,7 +2471,7 @@ export class ContainerProviderRegistry {
       const buildSupported: boolean = await this.isTarPlayBuildSupported(provider);
       if (!buildSupported)
         throw new Error(
-          `kube play build is not supported on ${provider.connection.name}: Podman 5.3 and above support this feature`,
+          `kube play build is not supported on ${provider.connection.name}: Podman 5.3.0 and above supports this feature`,
         );
 
       const kubePlay = KubePlayContext.fromFile(kubernetesYamlFilePath);
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -2440,7 +2440,6 @@ export class ContainerProviderRegistry {
     }
 
     const version = await internalProvider.api.version();
-    console.log('[container-registry] isTarPlayBuildSupported', version);
 
     // let's nicely format it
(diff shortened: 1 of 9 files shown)


gastoner (approved):
LGTM

packages/main/src/plugin/container-registry.ts
```diff
@@ -2432,24 +2434,59 @@ export class ContainerProviderRegistry {
...
 
+  protected async isTarPlayBuildSupported(internalProvider: InternalContainerProvider): Promise<boolean> {
+    if (!internalProvider.api || !internalProvider.libpodApi) {
+      throw new Error('No provider with a running engine');
+    }
+
+    const version = await internalProvider.api.version();
+
+    // let's nicely format it
+    const coerced = coerce(version.Version);
```
cdrage, line 2445: This section I'm a bit confused on haha. Maybe a comment be added. Do we "fake" that it's version 5.3.0 or something? I feel like 5.3.0 should be a constant somewhere instead?

packages/main/src/plugin/container-registry.ts
```diff
@@ -2432,24 +2434,59 @@ export class ContainerProviderRegistry {
...
+      // if we don't build, we use the file directory
+      if (!options?.build) {
+        return provider.libpodApi.playKube(kubernetesYamlFilePath);
+      }
+
+      // ensure build support is true, otherwise let's throw a nice user friendly error
+      const buildSupported: boolean = await this.isTarPlayBuildSupported(provider);
+      if (!buildSupported)
+        throw new Error(
+          `kube play build is not supported on ${provider.connection.name}: Podman 5.3 and above support this feature`,
```
cdrage:
```suggestion
          `kube play build is not supported on ${provider.connection.name}: Podman 5.3.0 and above supports this feature`,
```

packages/main/src/plugin/dockerode/libpod-dockerode.ts
```diff
@@ -805,6 +808,10 @@ export class LibpodDockerode {
         options: {},
       };
 
+      // if we don't build - we should send Content-Type application/yaml
```
cdrage, line 844: I could be remembering this incorrectly, but wasn't there a bug / PR pushed to correctly use `application/tar` or was that a different API call that was being updated upstream side?

cdrage (changes requested):
Sorry for the late review!

Few comments / confusions, but otherwise, amazing implementation.

Followed all the steps for using the new feature and it worked wonderfully :)

packages/main/src/plugin/container-registry.ts
axel7083: Calling `api#version` gave the following

```
{
  "Platform": {
    "Name": "linux/amd64/fedora-40"
  },
  "Components": [...],
  "Version": "5.3.1",
  "ApiVersion": "1.41",
  "MinAPIVersion": "1.24",
  "GitCommit": "",
  "GoVersion": "go1.22.7",
  "Os": "linux",
  "Arch": "amd64",
  "KernelVersion": "5.15.167.4-microsoft-standard-WSL2",
  "BuildTime": "2024-11-21T01:00:00+01:00"
}
```

So I just check for the `Version` field to be greater than 5.3.0, we don't risk to overlap with docker or other container engine as we filter the engine type to `podman`.

I can move the 5.3.0, but probably better to keep it inside the `isTarPlayBuildSupported` as the `container-registry` file is 2800 lines long, and adding things on top will probably be missed, here it is clear.

packages/main/src/plugin/dockerode/libpod-dockerode.ts
axel7083: When adding the build support in the api (https://github.com/containers/podman/pull/24015) they changed the accepted content type.

- when sending json/yaml you need to use `application/yaml` or `application/json`.
- when sending tarball you need to use `application/x-tar`.

By default, dockerode uses `application/tar` but this is not supported anymore for kube play (see list of accepted content type in https://github.com/containers/podman/pull/24015/files#diff-de26209aed278aeaa5803ab09bcb66127ec5346f34cfa4e0b21fead87adc6afeR43-R53)

axel7083: @benoitf made the fixes when they updated their API to use `application/x-tar` in https://github.com/podman-desktop/podman-desktop/pull/10036

packages/main/src/plugin/container-registry.ts
cdrage: ACK!

packages/main/src/plugin/dockerode/libpod-dockerode.ts
cdrage: Ah that makes sense, so much for clarifying. This LGTM.

cdrage (approved):
LGTM. Appreciate the changes and the explanations!

cdrage:
@axel7083 feel free to merge if you're comfortable with the code, since we are post 1.17.0

EDIT: Tested again and it worked fine still with regards to the build.

PR #10809, opened 2025-01-23
fix: use section titles and re-capitalize entry ids

### LINKED ISSUE #10808: Missing title spaces and entry capitalization in experimental features
### Bug description

While testing I found two issues with the fix for #10796:
- It uses the id for the section title instead of its title, which is missing spaces and potentially has other difference. So instead of e.g. 'Exit On Close' you'd get 'ExitOnClose'.
- Entry ids are no longer capitalized, so 'Enabled' is now 'enabled'.

Image

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Reverts the key back to what it was before, but immediately uses startCase(), then prefixes the record title if necessary.

### Screenshot / video of UI

Before:

Screenshot 2025-01-23 at 1 43 08 PM

After:

Screenshot 2025-01-23 at 1 20 09 PM

Fixes #10808.

### How to test this PR?

Open Settings > Experimental.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesRenderingItem.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesRenderingItem.spec.ts
@@ -92,7 +92,7 @@ test('props title full should use full record id', async () => {
   });
 
   await vi.waitFor(() => {
-    const element = getByText('Hello world foo Bar');
+    const element = getByText('Hello Foo Bar');
     expect(element).toBeDefined();
     expect(element).toHaveClass('font-semibold');
   });
--- a/packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesRenderingItem.svelte
@@ -30,19 +30,11 @@ const recordUI = $derived.by(() => {
   const id = record.id;
 
   // split id
-  const split: string[] = id?.split('.') ?? [''];
-
-  // take string after the last dot
-  let key: string;
-  switch (title) {
-    case 'full':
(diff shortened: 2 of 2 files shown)


deboer-tim:
Forgot to comment: I don't love the repeated 'Experimental (x)' here, but IMHO that's separate or follow-on issue. Using the id here is inconsistent with the Preferences and is missing spaces (looks like typo in this case), the title is the only 'external' string to use.

benoitf:
issue: it's not matching what has been decided in the design https://github.com/podman-desktop/podman-desktop/issues/10223

deboer-tim:
I want to be clear I was not trying to reopen or change the design - I just saw a couple bugs while testing and was trying to fix them!

Fixing the capitalization is easy enough, but using more segments of the id is causing issues because they have not been used as labels in the past and this would put a new requirement on how you structure them.

From my standpoint, we already have section titles so I defaulted to using those and keeping titles consistent between Experimental and Preferences. My thought was a second PR to remove the 'Experimental ()' from these two because they also show up differently than the other two experimental ones under Preferences. No new config id requirements, nice and consistent for users, and AFAIK that's consistent with the design (but I might have missed something).

I think the other option would be to close this and replace it with a PR to fix only the capitalization, and then another PR to change the status bar id to avoid the typo. I'm ok either way, and it's not a critical issue.

benoitf (changes requested):
issue: not compliant with the provided design

benoitf:
outdated

PR #10818, opened 2025-01-24
fix: switched to sequential writing to containers.conf file

### LINKED ISSUE #10709: proxy configuration via podman-desktop gui settings breaks containers.conf
### Bug description

I changed my Podman-Desktop proxy settings via GUI > Settings > Proxy Settings multiple times (manual to disabled back and forth) and when disabling proxy configuration again, the "~/.config/containers/containers.conf" file got malformed.

```
$ podman pull quay.io/podman/hello
Failed to obtain podman configuration: reading system config "/Users/nicowde/.config/containers/containers.conf": decode configuration /Users/nicowde/.config/containers/containers.conf: toml: line 16 (last key "configmaps"): expected a top-level item to end with a newline, comment, or EOF, but got '0' instead
``` 

### Operating system

macOS Sequoia 15.1.1 (24B91)

### Installation Method

Brew (macOS)

### Version

1.15.0

### Steps to reproduce

- GUI > Settings > Proxy Settings
- for Proxy configuration select"Manual" ->  insert variables -> click "Update" button
- e.g. podman pull quay.io/podman/hello -> works
- for Proxy configuration select"Disabled" ->  click "Update" button
- e.g. podman pull quay.io/podman/hello (again) -> errors

### Relevant log output

malformed containers.conf:

```
[containers]


[engine]
env = []

[machine]


[network]


[secrets]


[configmaps]80", "http_proxy=http://redacted:1234", "no_proxy=localhost,.local,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,127.0.0.1,.redact.ed,.redact2.ed,.re.dact.ed,.de.tcad.re,kubernetes.docker.internal" ]

[machine]


[network]


[secrets]


[configmaps]
``` 

clean / expected containers.conf (after hitting 'Update' via GUI > Settings > Proxy Settings)

```
[containers]


[engine]
env = [ "http_proxy=http://redacted:1234", "https_proxy=http://redacted:1234", "no_proxy=localhost,.local,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,127.0.0.1,.redact.ed,.redact2.ed,.re.dact.ed,.de.tcad.re,kubernetes.docker.internal" ]

[machine]


[network]


[secrets]


[configmaps]
``` 

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds sequential mechanism for writing to containers.conf file to avoid simultaneous write

Closes #10709 

### How to test this PR?
Go to PD, try to restart podman machine and while stopping/starting try to edit proxy settings switching between manual/disabled/system

### DIFF
--- a/extensions/podman/packages/extension/src/podman-configuration.spec.ts
+++ b/extensions/podman/packages/extension/src/podman-configuration.spec.ts
@@ -181,11 +181,11 @@ test('updateProxySettings should be called one at the time', async () => {
   const call1 = podmanConfiguration.doUpdateProxySettings(undefined);
   const call2 = podmanConfiguration.doUpdateProxySettings(proxySettings);
 
-  await Promise.all([call1, call2]);
-
-  // Check that the proxy update was called twice
-  expect(updateProxySettingsMock).toHaveBeenCalledTimes(2);
-
+  await call1;
+  expect(updateProxySettingsMock).toHaveBeenCalledTimes(1);
   expect(updateProxySettingsMock.mock.calls[0][0]).toBe(undefined);
+
+  await call2;
+  expect(updateProxySettingsMock).toHaveBeenCalledTimes(2);
   expect(updateProxySettingsMock.mock.calls[1][0]).toBe(proxySettings);
 });
--- a/extensions/podman/packages/extension/src/podman-configuration.spec.ts
+++ b/extensions/podman/packages/extension/src/podman-configuration.spec.ts
@@ -18,6 +18,7 @@
 
 import * as fs from 'node:fs';
 
+import type { ProxySettings } from '@podman-desktop/api';
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import { PodmanConfiguration } from './podman-configuration';
@@ -164,4 +165,27 @@ describe('isRosettaEnabled', () => {
 
     expect(isEnabled).toBeFalsy();
   });
+});
+
+test('updateProxySettings should be called one at the time', async () => {
+  const proxySettings: ProxySettings = {
+    httpProxy: 'httpProxy',
+    httpsProxy: 'httpsProxy',
+    noProxy: 'noProxy',
+  };
+
+  // Mock updateProxySettings
+  const updateProxySettingsMock = vi.spyOn(podmanConfiguration, 'updateProxySettings');
+
+  // Simultaneously call the function twice
+  const call1 = podmanConfiguration.doUpdateProxySettings(undefined);
+  const call2 = podmanConfiguration.doUpdateProxySettings(proxySettings);
+
+  await Promise.all([call1, call2]);
+
+  // Check that the proxy update was called twice
+  expect(updateProxySettingsMock).toHaveBeenCalledTimes(2);
+
+  expect(updateProxySettingsMock.mock.calls[0][0]).toBe(undefined);
+  expect(updateProxySettingsMock.mock.calls[1][0]).toBe(proxySettings);
 });
--- a/extensions/podman/packages/extension/src/podman-configuration.ts
+++ b/extensions/podman/packages/extension/src/podman-configuration.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2022-2024 Red Hat, Inc.
+ * Copyright (C) 2022-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,9 +22,11 @@ import * as path from 'node:path';
 
 import type { ProxySettings } from '@podman-desktop/api';
 import * as extensionApi from '@podman-desktop/api';
+import { Mutex } from 'async-mutex';
 import * as toml from 'smol-toml';
 
 const configurationRosetta = 'setting.rosetta';
+const mutex = new Mutex();
 
 /**
  * Manages access to the containers.conf configuration file used to configure Podman
@@ -37,7 +39,9 @@ export class PodmanConfiguration {
 
     // we receive an update for the current proxy settings
     extensionApi.proxy.onDidUpdateProxy(async (proxySettings: ProxySettings) => {
-      await this.updateProxySettings(proxySettings);
+      await mutex.runExclusive(async () => {
+        await this.updateProxySettings(proxySettings);
+      });
     });
 
     // in case of proxy being enabled or disabled we need to update the containers.conf file
@@ -45,9 +49,13 @@ export class PodmanConfiguration {
       // eslint-disable-next-line sonarjs/no-selector-parameter
       if (enabled) {
         const updatedProxySettings = extensionApi.proxy.getProxySettings();
-        await this.updateProxySettings(updatedProxySettings);
+        await mutex.runExclusive(async () => {
+          await this.updateProxySettings(updatedProxySettings);
+        });
       } else {
-        await this.updateProxySettings(undefined);
+        await mutex.runExclusive(async () => {
+          await this.updateProxySettings(undefined);
+        });
       }
     });
 
--- a/extensions/podman/packages/extension/src/podman-configuration.ts
+++ b/extensions/podman/packages/extension/src/podman-configuration.ts
@@ -26,12 +26,12 @@ import { Mutex } from 'async-mutex';
 import * as toml from 'smol-toml';
 
 const configurationRosetta = 'setting.rosetta';
-const mutex = new Mutex();
 
 /**
  * Manages access to the containers.conf configuration file used to configure Podman
  */
 export class PodmanConfiguration {
+  private mutex: Mutex = new Mutex();
   async init(): Promise<void> {
     let httpProxy = undefined;
     let httpsProxy = undefined;
@@ -39,9 +39,7 @@ export class PodmanConfiguration {
 
     // we receive an update for the current proxy settings
     extensionApi.proxy.onDidUpdateProxy(async (proxySettings: ProxySettings) => {
-      await mutex.runExclusive(async () => {
-        await this.updateProxySettings(proxySettings);
-      });
+      await this.doUpdateProxySettings(proxySettings);
     });
 
     // in case of proxy being enabled or disabled we need to update the containers.conf file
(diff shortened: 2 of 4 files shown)


benoitf:
note: I don't think the issue is coming from this part

gastoner:
@benoitf can you try this PR please? From my testing, it fixes the issue (tested remotely on Mac)

benoitf:
to me it's not fixing the root problem, the cause. Changing would just hide/mask/reduce the issue

gastoner:
Updated based on @jeffmaury tips

extensions/podman/packages/extension/src/podman-configuration.ts
```diff
@@ -105,6 +107,20 @@ export class PodmanConfiguration {
     }
   }
 
+  async updateProxySettigsSync(proxy: ProxySettings | undefined): Promise<void> {
```
jeffmaury: Looks wroong to me as if there is a promise then the proxy argument is not beging processed.

I was thinking of something:
if (!promise) {
  promise = this.updateProxySettings(proxy);
} else {
  promise.then(() => promise = this.updateProxySettings());
}

jeffmaury (changes requested):

extensions/podman/packages/extension/src/podman-configuration.ts
gastoner: Thank you! Updated

gastoner:
@jeffmaury PR Updated, can you please check again?

jeffmaury (approved):
LGTM

extensions/podman/packages/extension/src/podman-configuration.ts
```diff
@@ -30,24 +30,26 @@ const configurationRosetta = 'setting.rosetta';
  * Manages access to the containers.conf configuration file used to configure Podman
  */
 export class PodmanConfiguration {
+  updatedProxySettingsPromise: Promise<void> | undefined = undefined;
```
axel7083: We should have this private

extensions/podman/packages/extension/src/podman-configuration.ts
```diff
@@ -30,24 +30,26 @@ const configurationRosetta = 'setting.rosetta';
...
+
   async init(): Promise<void> {
     let httpProxy = undefined;
     let httpsProxy = undefined;
     let noProxy = undefined;
 
     // we receive an update for the current proxy settings
     extensionApi.proxy.onDidUpdateProxy(async (proxySettings: ProxySettings) => {
-      await this.updateProxySettings(proxySettings);
+      await this.updateProxySettigsSync(proxySettings);
```
axel7083: The `*Sync` method name is usually reserved for blocking io, see node method `existsSync`  :/ so having a await is a bit weird

extensions/podman/packages/extension/src/podman-configuration.ts
```diff
@@ -105,6 +107,16 @@ export class PodmanConfiguration {
     }
   }
 
+  async updateProxySettigsSync(proxy: ProxySettings | undefined): Promise<void> {
+    if (!this.updatedProxySettingsPromise) {
+      this.updatedProxySettingsPromise = this.updateProxySettings(proxy);
```
axel7083: We are not awaiting for `this.updatedProxySettingsPromise`, this is a problem? like do we want the `updateProxySettigsSync` to finish before the updateProxySettings promise resolved ?

gastoner: The problem is that the `this.updateProxySettings()` function is called in some cases 2 times, which results in corrupted file. 
The purpose of this is to execute this function sequentially using promises/await

axel7083: I feel like this is an issue to have several call, I would argue we should ignore the second one, or throw an error if the resource is locked.

The root cause is probably the fact that it is called twice.

Otherwise if this is not feasable, we should improve the method signature.

```
queueProxySettingsUpdate(proxy: ProxySettings | undefined): void;
```

This is not a sync operation, but a queue, as we chain the promises until resolve.
Please also add a cleaning mechanism, the promise should be undefined if all are resolved

benoitf (commented):
suggestion:

adds https://www.npmjs.com/package/async-mutex

do :

```ts
// update the proxy only once at a time
extensionApi.proxy.onDidUpdateProxy(async (proxySettings: ProxySettings) => {
   await mutex.runExclusive(async () => {
     await this.updateProxySettings(proxySettings);   
   });
});
```

or do the mutex inside the proxySettings call  as you suggested

```ts
const mutex = new Mutex();

updateProxySettings(proxySettings) {
   await mutex.runExclusive(async () => {
     await this.doUpdateProxySettings(proxySettings);   
   });
}

doUpdateProxySettings(...) {

}
```

benoitf (commented):
I would love to see a test ensuring that the method updateProxySetings is not called at the same time 

without the new code = it fails, with the new code it succeeed

Else LGTM

gastoner:
@benoitf Added the lib, but I have some linting errors

> I would love to see a test ensuring that the method updateProxySetings is not called at the same time
> 
> without the new code = it fails, with the new code it succeeed
> 
> Else LGTM

I'll add the unit test, do you know how to fix the linter check? running the command does not work

benoitf:
@gastoner yes you need to add the dependency in the package.json file of the extension of podman, not in the root package.json file

extensions/podman/packages/extension/src/podman-configuration.ts
```diff
@@ -30,24 +31,26 @@ const configurationRosetta = 'setting.rosetta';
  * Manages access to the containers.conf configuration file used to configure Podman
  */
 export class PodmanConfiguration {
+  mutex: Mutex | undefined;
```
benoitf: issue: initialize in the constructor and use private, it should not be accessible outside of this class

extensions/podman/packages/extension/src/podman-configuration.spec.ts
```diff
@@ -165,3 +166,45 @@ describe('isRosettaEnabled', () => {
...
+
+  // Mock updateProxySettings to introduce a delay
+  const updateProxySettingsMock = vi.spyOn(podmanConfiguration, 'updateProxySettings').mockImplementation(async () => {
+    await new Promise(resolve => setTimeout(resolve, 100));
+  });
+
+  if (podmanConfiguration.mutex) {
+    // Spy on acquire and release
+    const acquireMock = vi.spyOn(podmanConfiguration.mutex, 'acquire');
+    const releaseMock = vi.spyOn(podmanConfiguration.mutex, 'release');
```
benoitf: AFAIK it should not be reading the value the private value

also what happens if the value is not defined ? all the test will be skipped without error

extensions/podman/packages/extension/src/podman-configuration.spec.ts
```diff
@@ -165,3 +171,44 @@ describe('isRosettaEnabled', () => {
...
+
+  // Mock updateProxySettings
+  const updateProxySettingsMock = vi.spyOn(podmanConfiguration, 'updateProxySettings');
+
+  // Spy on acquire and release
+  const mutex = podmanConfiguration.getMutex();
+
+  const acquireSpy = vi.spyOn(mutex, 'acquire');
+  const releaseMock = vi.fn();
+  acquireSpy.mockResolvedValue(releaseMock);
```
benoitf: question: what I'm not understanding is why we need to intercept/monitor the mutex ?

benoitf: like for example if I throw 1000 events in parallel and the doUpdateProxySettings increment a value by one
at the end, do I have the counter being 1000 or as there well multiple calls in parallel it didn't increase correctly


Here I have the feeling that we're checking that the mutex library is working, while I would check that the method is really doing stuff sequentially

gastoner: To check that the functions are called in right order, or the locking/unlocking mechanism works as expected

gastoner: But need to test it on Mac once it is built

jeffmaury: I think mutex should not be exposed through test but rather that calls to updateProxySettings be done in the right order and with the expected parameters

jeffmaury (commented):
See https://github.com/podman-desktop/podman-desktop/pull/10818/files#r1949496560

gastoner:
@jeffmaury suggestions applied

jeffmaury (approved):
LGTM

extensions/podman/packages/extension/src/podman-configuration.spec.ts
```diff
@@ -165,3 +166,26 @@ describe('isRosettaEnabled', () => {
...
+  };
+
+  // Mock updateProxySettings
+  const updateProxySettingsMock = vi.spyOn(podmanConfiguration, 'updateProxySettings');
+
+  // Simultaneously call the function twice
+  const call1 = podmanConfiguration.doUpdateProxySettings(undefined);
+  const call2 = podmanConfiguration.doUpdateProxySettings(proxySettings);
+
+  await Promise.all([call1, call2]);
```
feloy: The test does not cover that the calls are done one after the other (if you remove the use of the mutex in `doUpdateProxySettings`, the test still passes).

I would expect this instead:

```
  await call1;
  expect(updateProxySettingsMock).toHaveBeenCalledTimes(1);

  expect(updateProxySettingsMock.mock.calls[0][0]).toBe(undefined);

  await call2;
  expect(updateProxySettingsMock).toHaveBeenCalledTimes(2);
  expect(updateProxySettingsMock.mock.calls[1][0]).toBe(proxySettings);

```

feloy (approved):

extensions/podman/packages/extension/src/podman-configuration.ts
```diff
@@ -105,6 +107,15 @@ export class PodmanConfiguration {
     }
   }
 
+  async doUpdateProxySettings(proxy: undefined | ProxySettings): Promise<void> {
```
benoitf, line 110: issue: to me it looks like the opposite on how methods usually work

internal methods call updateProxySettings and then it's the method updateProxySettings that call doUpdateProxySettings

benoitf: here, all existing calls were replaced from updateProxySettings to doUpdateProxySettings while they should not have been changed

gastoner: Should I create an additional PR to change `updateProxySettings` to `doUpdateProxySettings` and vice versa?

benoitf: It would be awesome 😎

gastoner: Created https://github.com/podman-desktop/podman-desktop/pull/11081

PR #10852, opened 2025-01-27
feat: kube pods page

deboer-tim:
### What does this PR do?

Adds a Kubernetes Pods page, similar to Deployments/Services/etc, but with the same columns and actions as the Kubernetes pods in the current Pods page.

Given the number of files just to add the page, this commit intentionally does not include a details page. That would come in following PRs.

Since we already have a PodsList in src/lib/pod, I've put this in /src/lib/kube/pods, not sure if there is a better location or we should start organizing Kube folders differently.

### Screenshot / video of UI

Screenshot 2025-01-27 at 4 39 47 PM
Screenshot 2025-01-27 at 4 40 18 PM

Part of #8818.

### How to test this PR?

Connect to a Kube cluster and work with Pods; compare to other pages.

### DIFF
--- a/packages/renderer/src/lib/kube/pods/PodActions.svelte
+++ b/packages/renderer/src/lib/kube/pods/PodActions.svelte
@@ -1,0 +1,114 @@
+<script lang="ts">
+import { faArrowsRotate, faExternalLinkSquareAlt, faTrash } from '@fortawesome/free-solid-svg-icons';
+import { DropdownMenu } from '@podman-desktop/ui-svelte';
+import { createEventDispatcher, onMount } from 'svelte';
+
+import { withConfirmation } from '/@/lib/dialogs/messagebox-utils';
+
+import FlatMenu from '../../ui/FlatMenu.svelte';
+import ListItemButtonIcon from '../../ui/ListItemButtonIcon.svelte';
+import type { PodUI } from './PodUI';
+
+export let pod: PodUI;
+export let dropdownMenu = false;
+export let detailed = false;
+
+const dispatch = createEventDispatcher<{ update: PodUI }>();
+export let onUpdate: (update: PodUI) => void = update => {
+  dispatch('update', update);
+};
+
+$: openingKubernetesUrls = new Map();
+
+onMount(async () => {
+  const ns = await window.kubernetesGetCurrentNamespace();
+  if (ns) {
+    const kubepod = await window.kubernetesReadNamespacedPod(pod.name, ns);
+    if (kubepod?.metadata?.labels?.app) {
+      const appName = kubepod.metadata.labels.app;
+      const routes = await window.kubernetesListRoutes();
+      const appRoutes = routes.filter(r => r.metadata.labels && r.metadata.labels['app'] === appName);
+      appRoutes.forEach(route => {
+        openingKubernetesUrls = openingKubernetesUrls.set(
+          route.metadata.name,
+          route.spec.tls ? `https://${route.spec.host}` : `http://${route.spec.host}`,
+        );
+      });
+    }
+  }
+});
+
+async function restartPod(): Promise<void> {
+  pod.status = 'RESTARTING';
+  onUpdate(pod);
+
+  await window.restartKubernetesPod(pod.name);
+}
+
+async function deletePod(): Promise<void> {
+  pod.status = 'DELETING';
+  onUpdate(pod);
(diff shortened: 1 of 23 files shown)


cdrage:
https://github.com/user-attachments/assets/bd98abf8-d15d-4329-b035-9fef8ff315ba

I'm a bit early reviewing, but looks like "Pod Restart" does not work although deletion works.

deboer-tim:
I'm still chasing down one test failure, but code is ready and time to start full review.

@cdrage I'm getting the same failure restarting, but test and log output confirms restartKubernetesPod() is getting called, and identical failure is happening when trying from the current Pods page (i.e. it's a bug, but unrelated to this PR). I'll test on 1.16 and open an issue to track it.

feloy:
Tested, LGTM

packages/renderer/src/lib/kube/pods/PodActions.svelte
```diff
@@ -0,0 +1,114 @@
...
+export let pod: PodUI;
+export let dropdownMenu = false;
+export let detailed = false;
+
+const dispatch = createEventDispatcher<{ update: PodUI }>();
+export let onUpdate: (update: PodUI) => void = update => {
+  dispatch('update', update);
+};
+
+$: openingKubernetesUrls = new Map();
```
axel7083, line 21: svelte5?

packages/renderer/src/stores/navigation/kubernetes/navigation-registry-k8s-pods.svelte.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+import { createNavigationKubernetesDeploymentsEntry } from './navigation-registry-k8s-deployments.svelte';
+
+beforeEach(() => {
+  vi.resetAllMocks();
+  Object.defineProperty(window, 'kubernetesRegisterGetCurrentContextResources', {
+    value: kubernetesRegisterGetCurrentContextResourcesMock,
+  });
+});
+
+const kubernetesRegisterGetCurrentContextResourcesMock = vi.fn();
```
axel7083: We should not define mock like this, we should use `vi.mocked(window.kubernetesRegisterGetCurrentContextResources)` to access it, defining mock like this lose all the typing of the mocked function

packages/renderer/src/lib/kube/pods/PodColumnActions.spec.ts
```diff
@@ -0,0 +1,40 @@
+/**********************************************************************
```
axel7083, line 1: suggestion (non-blocking, non-mandatory): I am wondering if we should not try to use a better folder structures instead of prefixing all component with `PodColumn` let's just create a `lib/kube/pods/column` folder and add `Actions.svelte`, `Containers.svelte` etc.

packages/renderer/src/stores/navigation/kubernetes/navigation-registry-k8s-pods.svelte.spec.ts
deboer-tim: Updated to use vi.mocked in new commit (expect ResizeObserver, which I couldn't figure out).

Side note: I agree with the improved standards, but now we'll have 7 'copy/paste' Kubernetes resource files, and this one that's better but unique. Need to make sure we follow-up and make the others match.

packages/renderer/src/lib/kube/pods/PodActions.svelte
deboer-tim: I updated all the rest, but touching this one meant I had to switch to svelte5 events - and this one is currently propagating an event from the actions, through the svelte4 column on:update, through the svelte4 table on:update, back to the pods list...  and I got nervous. I'll try to do more testing; maybe svelte5 now has enough reactivity that the event can be safely removed.

packages/renderer/src/lib/kube/pods/PodColumnActions.spec.ts
deboer-tim: That's actually a pretty good idea. I don't want to start that here if it's not agreed in general, but if we get a +1 or two I'll start that here and open an issue for the others.

axel7083 (approved):
LGTM, thanks for thee changes 🚀

feloy (approved):
thanks 🚀

PR #10853, opened 2025-01-28
fix: show recommended ext card when podman is ready

### LINKED ISSUE #10332: Hide call-to-actions on dashboard until ready

When a user does not have a the ability to start a container they will not be able to use extensions like AI Lab (yet!). We should hide call to actions on the dashoards until the user has atleast a container provider ready. 

When the user doesnt have a Podman Machine or any other container provider, hide the call to actions on the dashbaord. Currently the only action we have that will benefit from this behavior is AI Lab.

dgolovin:
### What does this PR do?

Straightforward implementation to show/hide recommended extension cards (there is only one for AI Lab now) based on status of podman provider.

### Screenshot / video of UI

https://github.com/user-attachments/assets/6ca21258-a561-4d84-8f50-8f24b5440e34

Fixes #10332. 

### How to test this PR?


benoitf (changes requested):
issue: restricting recommendations only if podman is there is not a solution. It would prevent to display any card just b/c podman is not there

packages/renderer/src/lib/recommendation/ExtensionBanners.spec.ts
```diff
@@ -17,58 +17,94 @@
...
-    receive: (_channel: string, func: () => void): void => {
-      func();
-    },
-  };
-});
+suite('Recommendation banners', () => {
+  // fake the window.events object
+  beforeEach(() => {
+    vi.resetAllMocks();
+    Object.defineProperty(window, 'getExtensionBanners', { value: getExtensionBannersMock });
```
benoitf: can be removed as we have a global mock for every function 'exposed in main world' now

same for all the functions returning a vi.fn()

to override a behavior, use vi.mocked like `vi.mocked(window...).....` no need to define the property now

gastoner (changes requested):
After those changed it does not seem that using the hide button is no longer reactive

https://github.com/user-attachments/assets/6e31cfb7-6be4-4835-9e98-c9fec2f8de4f

gastoner (approved):
LGTM

benoitf (approved):
worked on my end, now it's way better with the when clause 👏

PR #10863, opened 2025-01-28
docs: Add missing -o flag in `Importing saved containers to Podman`

gaetschwartz:
### What does this PR do?

Adds missing `-o` flag.

Reference: https://docs.docker.com/reference/cli/docker/container/export/

### DIFF
--- a/website/docs/migrating-from-docker/importing-saved-containers.md
+++ b/website/docs/migrating-from-docker/importing-saved-containers.md
@@ -30,7 +30,7 @@ Consider importing saved containers to continue using familiar containers.
     <TabItem value="docker" label="Docker">
 
   ```shell-session
-  $ docker export <your_container> > <your_container_archive>.tar
+  $ docker export <your_container> > -o <your_container_archive>.tar
   ```
 
     </TabItem>


axel7083:
Thanks @gaetschwartz for the fix ! 

We uses semantic commits and PR title ! So the title need to be changed to something like `docs: adding missing -o flag` same for the commit.

cdrage (approved):

benoitf (approved):

benoitf:
thanks @gaetschwartz for the fix, you'll need to amend your commit to make it compliant with conventional commits

gaetschwartz:
Should I also rename the commit ? @axel7083

benoitf:
@gaetschwartz yes

`Semantic PR — make sure every commit AND the PR title is semantic`

benoitf:
there is a commit hook if you're running the `git` command locally that is checking it's valid

benoitf:
@gaetschwartz do you need help to edit your commit message ?

gaetschwartz:
@benoitf I was doing in remotely thus didn't get to use the git hook, amended the commit it should be good now :)

benoitf:
@gaetschwartz it's now missing the Signed-off

`git commit --amend -s`

benoitf:
thanks

PR #10877, opened 2025-01-29
fix: do not check connection after listen

### LINKED ISSUE #10476: Kubernetes Port Forwarding always fails with "Host is not reachable"
### Bug description

Installed Podman Desktop via flatpak and tried the port forwarding feature but it never works. It always shows "Error: Host is not reachable: This operation was aborted".

Image

Port forwarding using kubectl or VSCode works fine and without any issues.

### Operating system

Ubuntu 24.04

### Installation Method

Flathub (Linux)

### Version

1.15.0

### Steps to reproduce

1. Start a pod in kubernetes
2. Go to "Pod Details"
3. Click on Summary > "Forward..." 

### Relevant log output

```shell

```

### Additional context

_No response_

feloy:

### What does this PR do?

This PR removes the check after `server.listen` when creating a port forward:

- the check is much more restrictive: it is done with `fetch`, which covers only HTTP connections, and checks for a 200 status result. This makes impossible to port forward non-HTTP apps, and apps not returning 200 (app to debug returning 500 for example)
- should we change with a TCP check? I'm not sure @vzhukovs do you think it is necessary to keep a check here?
 
Fixes #10476

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-port-forward-connection.spec.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-port-forward-connection.spec.ts
@@ -30,7 +30,7 @@ import {
   PortForwardConnectionService,
 } from '/@/plugin/kubernetes/kubernetes-port-forward-connection.js';
 import type { Telemetry } from '/@/plugin/telemetry/telemetry.js';
-import { Disposable, type IDisposable } from '/@/plugin/types/disposable.js';
+import { type IDisposable } from '/@/plugin/types/disposable.js';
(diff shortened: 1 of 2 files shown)


gastoner (approved):
Codewise LGTM

vzhukovs:
I think, we can provide more detailed check in the future, but at this moment, okay to delete it. LGTM.

vzhukovs (approved):

feloy:
@Mergifyio backport 1.16.x

PR #10879, opened 2025-01-29
feat: added option to update to prereleases

### LINKED ISSUE #10208: Podman Desktop should support multiple update channels to allow for automatic installation of unofficial releases

It would be good to be able to have support for multiple release channels (e.g alpha/beta/releases) channels in order to be able to automatically update to edge versions that are not official releases.

Setting should be available initially as an experimental flag in the Settings menu. 

PD should support multiple update channels which would permit the user to automatically update to unofficial releases.

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds hidden option that allows updates to pre-releases

Closes #10208 

### How to test this PR?
add this line to `settings.json` file: 
`"preferences.update.prererelease": true` 
The PD should offer you an update to pre-release (must be available at the moment)

### DIFF
--- a/packages/main/src/plugin/updater.ts
+++ b/packages/main/src/plugin/updater.ts
@@ -298,6 +298,19 @@ export class Updater {
           },
         },
       },
+      {
+        id: 'preferences.update.prerelease',
+        title: 'Allow prerelease',
+        type: 'object',
+        properties: {
+          ['preferences.update.prerelease']: {
+            description: 'Configure whether you want to allow updating to prerelease verions of Podman Desktop',
+            type: 'boolean',
+            hidden: true,
+            default: 'false',
+          },
+        },
+      },
     ]);
   }
 
@@ -322,6 +335,14 @@ export class Updater {
       .catch((err: unknown) => {
         console.error('Something went wrong while trying to update update.reminder preference', err);
       });
+  }
+
+  /**
+   * Retrieves the value of the prerelease update configuration.
+   * @returns The value of the prerelease update configuration.
+   */
+  private getConfigurationPrereleaseValue(): boolean {
+    return this.configurationRegistry.getConfiguration('preferences').get<boolean>('update.prerelease', false);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


gastoner:
@benoitf this can be tested when the 1.16.1 will have pre-release, I've tested this with 1.14 version I think

benoitf:
hello, I'm not sure it was the goal of the issue but I can be mistaken.

Like if I download today a pre-release from https://github.com/podman-desktop/prereleases I expect to be able to update to a new one

I mean, it should work out of the box. I downloaded a binary from the prelease repository so I want to be able to update


Here, it would make opt-in to allow to use 1.16.1 for example before it's being rolled out and in that case I don't think it's a good idea at all to let people upgrade from stable release to a release that is not yet flagged as a release

gastoner:
> hello, I'm not sure it was the goal of the issue but I can be mistaken.
> 
> Like if I download today a pre-release from https://github.com/podman-desktop/prereleases I expect to be able to update to a new one
> 
> I mean, it should work out of the box. I downloaded a binary from the prelease repository so I want to be able to update
> 
> Here, it would make opt-in to allow to use 1.16.1 for example before it's being rolled out and in that case I don't think it's a good idea at all to let people upgrade from stable release to a release that is not yet flagged as a release

Didn't notice that there are binaries :D but yess I agree

The idea was to check if the update system was working to prevent scenarios like the one with the update using 1.12 
Or allow to user if he knows the settings to get the pre-release version ahead of the others

benoitf:
>The idea was to check if the update system was working to prevent scenarios like the one with the update using 1.12
Or allow to user if he knows the settings to get the pre-release version ahead of the others

AFAIK the problem in the update is not in the target version but in the source version

if I'm in x.y.z version I can't update to x.y.foo version but then testing this scenario is too late as x.y.z is already released. This is why we have a test that package the current version as a previous version and then we try to upgrade

based on your comment, let say you see you're in 1.16.0 and you detect you can't update 1.16.1 then it's too late as 1.16.0 is already released

odockal:
I think that crucial part is that we do not know, when we are building the bits in production, whether they will became stable or not since the only concept we are using is Github release tags -> latest and pre-release.

For this to work as you outlined, we would need to introduce channels - beta or alpha in the update process (I think). But this is another level of complexity. 

Having hidden settings that one must set manually via cli in order to be able to use pre-release releases does not seem like something wrong to me, user should be aware what can happen when enabling this.

Issue #10657, opened 2025-01-15
feat: display resources in the frontend in experimental mode


feloy:
### Is your feature request related to a problem? Please describe

The backend provides resources (https://github.com/podman-desktop/podman-desktop/issues/10632).

### Describe the solution you'd like

Display this information in the frontend, using the data provided by the experimental manager instead of the legacy one

### Additional context

- [x] Depends on refactor: https://github.com/podman-desktop/podman-desktop/issues/10907
- [x] Deployments: https://github.com/podman-desktop/podman-desktop/pull/10880
- [x] Configmaps / secrets: https://github.com/podman-desktop/podman-desktop/pull/10928
- [x] Build a `KubernetesObjectList` (see https://github.com/podman-desktop/podman-desktop/pull/10880#pullrequestreview-2585997358) and use it for Deployments and Configmaps/secrets: https://github.com/podman-desktop/podman-desktop/pull/10928
- [x] Pods: https://github.com/podman-desktop/podman-desktop/pull/11103
- [x] Nodes (https://github.com/podman-desktop/podman-desktop/pull/11054)
- [x] Services: https://github.com/podman-desktop/podman-desktop/pull/11020
- [x] Ingresses/Routes: https://github.com/podman-desktop/podman-desktop/pull/11065
- [x] PVCs: https://github.com/podman-desktop/podman-desktop/pull/11021
- [x] CronJobs: https://github.com/podman-desktop/podman-desktop/pull/11100

Display details:
- [x] depends on #10916
- [x] Build a `listenResource` helper: #11196 
- [x] Pods
  - [x] https://github.com/podman-desktop/podman-desktop/pull/11378
  - [x] https://github.com/podman-desktop/podman-desktop/pull/11418
- [x] Deployments: https://github.com/podman-desktop/podman-desktop/pull/11306
- [x] Configmaps: https://github.com/podman-desktop/podman-desktop/pull/11335
- [x] Secrets: https://github.com/podman-desktop/podman-desktop/pull/11336
- [x] Nodes: https://github.com/podman-desktop/podman-desktop/pull/11305
- [x] Services: https://github.com/podman-desktop/podman-desktop/pull/11196
- [x] Ingresses: https://github.com/podman-desktop/podman-desktop/pull/11341
- [x] Routes
  - [x] https://github.com/podman-desktop/podman-desktop/pull/11349
  - [x] https://github.com/podman-desktop/podman-desktop/pull/11377
[...]

PR #10880, opened 2025-01-29
feat: display deployments from experimental informers

feloy:

### What does this PR do?

Display deployments in experimental mode

### Screenshot / video of UI

deployments-experimental

Part of #10657 

### How to test this PR?

Set the Kubernetes experimental mode in `$HOME/.local/share/containers/podman-desktop/configuration/settings.json`:

```
```

Also check that it is still working in non experimental mode

### DIFF
--- a/packages/renderer/src/lib/deployments/DeploymentsList.spec.ts
+++ b/packages/renderer/src/lib/deployments/DeploymentsList.spec.ts
@@ -21,13 +21,16 @@ import '@testing-library/jest-dom/vitest';
 import type { KubernetesObject, V1Deployment } from '@kubernetes/client-node';
 import { fireEvent, render, screen, within } from '@testing-library/svelte';
 import { writable } from 'svelte/store';
-import { beforeEach, expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 
 import * as states from '/@/stores/kubernetes-contexts-state';
+import type { KubernetesContextResources } from '/@api/kubernetes-resources';
 
+import * as resourcesListen from '../kube/resources-listen';
 import DeploymentsList from './DeploymentsList.svelte';
 
 vi.mock('/@/stores/kubernetes-contexts-state');
+vi.mock('../kube/resources-listen');
 
 beforeEach(() => {
   vi.resetAllMocks();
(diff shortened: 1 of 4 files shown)


axel7083 (commented):
I feel like this is adding extra complexity to the DeploymentList component, can't we just overwrite the `kubernetesCurrentContextDeploymentsFiltered` store?

Having many store subscribe too, disposed, filtered etc. is making the component hard to understand when it should be just a list :/

feloy:
> I feel like this is adding extra complexity to the DeploymentList component, can't we just overwrite the `kubernetesCurrentContextDeploymentsFiltered` store?
> 
> Having many store subscribe too, disposed, filtered etc. is making the component hard to understand when it should be just a list :/

Using a store was my first idea, but @benoitf argued that stores lifetime should be the app lifetime, but here we would have had to create a store every time the user enters a Kubernetes page

deboer-tim:
> > I feel like this is adding extra complexity to the DeploymentList component, can't we just overwrite the `kubernetesCurrentContextDeploymentsFiltered` store?
> > Having many store subscribe too, disposed, filtered etc. is making the component hard to understand when it should be just a list :/
> 
> Using a store was my first idea, but @benoitf argued that stores lifetime should be the app lifetime, but here we would have had to create a store every time the user enters a Kubernetes page

I share the same concern as Axel - this is a lot of change to the deployments page which is really just a UI to display a list. It shouldn't be involved in any complexity around how the list is made or where it comes from.

This will be more of an issue if this works great, because then we'd copy/paste all this logic around to services, ingress, routes, pvcs, etc.

So maybe it's not a store, but an interim file that can handle this complexity, or more pushed to resources-listen?

feloy:
@deboer-tim @axel7083 I moved more logic to the resourceListener, please tell me what do you think

benoitf (approved):
non-blocking:
I would hope that all the copy/paste code block should be managed in a different object

DeploymentList using KubernetesObjectList svelte component passing only type="deployments" and a callback to receive raw kubernetesObject array or even better I give the function

to handle
 const deployments = $derived(resources?.map(deployment => deploymentUtils.getDeploymentUI(deployment)) ?? []);
 
 so if I want to add a new List for 'foobar' type I'll just have
 type='foobar' and onKubernetesObject wrapping it
 
 Meaning that tomorrow it's possible to change the whole underlying logic without changing each individual files

feloy:
> non-blocking: I would hope that all the copy/paste code block should be managed in a different object
> 
> DeploymentList using KubernetesObjectList svelte component passing only type="deployments" and a callback to receive raw kubernetesObject array or even better I give the function
> 
> to handle const deployments = $derived(resources?.map(deployment => deploymentUtils.getDeploymentUI(deployment)) ?? []);
> 
> so if I want to add a new List for 'foobar' type I'll just have type='foobar' and onKubernetesObject wrapping it
> 
> Meaning that tomorrow it's possible to change the whole underlying logic without changing each individual files

I prefer to start introduce the experimental mode in some of other lists (at least in the Configmaps / secrets list which introduces a couple of resources to list), to be sure that this  `KubernetesObjectList` will fit all the needs. I will add this component ASAP

benoitf:
I'm fine iterating, just expressing if we're all in the same direction

deboer-tim (approved):
Discussed with @feloy to make sure I had all the context. +1, agree we should do KubernetesObjectList soon (ok to look at a page that combines like configmap/secrets or ingress/route, but IMHO the page can continue to do the merging) and nail the pattern, then roll out to other pages.

Issue #10795, opened 2025-01-23
update to tailwindcss v4


benoitf:
### Is your enhancement related to a problem? Please describe

v4 has been released and we're using v3

need to do some changes to be able to upgrade

### Describe the solution you'd like

upgrade to v4

PR #10883, opened 2025-01-29
chore: use podman desktop colors rather than theme colors

benoitf:
### What does this PR do?
while updating to tailwindcss v4, there are changes on the theming colors, so it would not work anymore.
But then, now that Podman Desktop is exporting some css variables for colors, it should match the one from the components rather than the theme colors

So apply same style for the links that 'our' Link item

### Screenshot / video of UI
before
image

after:
image

related to #10795 
### How to test this PR?

check links in the settings/resources for example and the steps (remove all podman machines and click on 'init and start' button on the podman provider card, see the bgcolor of the step.

### DIFF
--- a/packages/renderer/src/lib/markdown/Markdown.svelte
+++ b/packages/renderer/src/lib/markdown/Markdown.svelte
@@ -35,12 +35,13 @@ UI guidelines -->
   line-height: normal;
 }
 .markdown :global(a) {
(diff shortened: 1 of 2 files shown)


deboer-tim (approved):
Nice cleanup and consistency.  👍🏼

axel7083 (approved):

PR #10888, opened 2025-01-29
chore: postinstall should only install chromium

axel7083:
### What does this PR do?

Question: do we need to have all browser installed when running playwright install?

I noticed while working on https://github.com/podman-desktop/extension-podman-quadlet/pull/236 that we download every browser in existence, which take some time and resources.

Since electron is only using chromium, we may be able to limit the browser download (might be wrong?)

### How to test this PR?

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "storybook:dev": "npm run storybook:css && cd storybook && pnpm run dev",
     "storybook:build": "npm run build:ui && npm run storybook:css && cd storybook && pnpm run build",
     "prepare": "husky install",
-    "postinstall": "playwright install"
+    "postinstall": "playwright install chromium"
   },
   "lint-staged": {
     "*.{js,ts,tsx,svelte}": [


benoitf (commented):
Electron is using chromium 
But these browsers are for playwright / e2e tests

I'm not sure if here chromium is used as default browser on all the platforms

axel7083:
> I'm not sure if here chromium is used as default browser on all the platforms

On the quadlet extension it download all browser, but only use chromium :/ 

cc @podman-desktop/qe-reviewers any idea ?

cbr7:
@axel7083 Do you mean downloading the browsers when installing playwright dependency? I think it's the default behavior to install all of them unless specifically provided the extra arguments that specify which browser to install as per the [documentation](https://playwright.dev/docs/browsers)

axel7083:
> @axel7083 Do you mean downloading the browsers when installing playwright dependency?

My question is: can we limit to chromium ? So we don't download every browser if we only need one?

benoitf:
I would be in favor of limiting and revert if we see a regression

benoitf (approved):

cbr7:
@axel7083 @benoitf according to [this](https://github.com/microsoft/playwright/issues/812) there are specific npm packages for each browser stack, so we'd have to change the dependencies to the chromium one instead of @playwright/test

benoitf:
I think the comment it outdated / old versions of playwright, now the imports are on '@playwright/test'

benoitf:
https://github.com/microsoft/playwright/issues/812#issuecomment-1070933740 (same thread but comment below)

cbr7:
@benoitf no, I saw that but the browser specific packages still exist and are up to date, here's the chromium one https://www.npmjs.com/package/playwright-chromium?activeTab=versions

You can either use the main package and install only the browser that you want as per the documentation or install the browser specific package, both should work.

benoitf:
it's the legacy/old way

now the recommended way is to use `@playwright/test`

benoitf:
these packages are used under the hood and should not be imported directly to the project

gastoner (approved):

odockal:
Let's see how that work on win and mac. Otherwise I think it should be alright.

axel7083:
> Let's see how that work on win and mac. Otherwise I think it should be alright.

@odockal tell me when I can merge (or not)!

odockal:
> > Let's see how that work on win and mac. Otherwise I think it should be alright.
> 
> @odockal tell me when I can merge (or not)!

It works on windows pretty well (sorry for delay). So I believe that on Mac it should be the same, I was only worried about linux and it worked. LGTM.

odockal (approved):

PR #10897, opened 2025-01-30
chore: bump svelte 5.19.5 and weird reactivity fix

axel7083:
### What does this PR do?

Very strange bug... the following button is having some issues when it is supposed to be enabled

https://github.com/podman-desktop/podman-desktop/blob/8321e0b7ad10eb87c6fa24a0833a794eb4bb7f0c/packages/renderer/src/lib/feedback/feedbackForms/GitHubIssueFeedback.svelte#L117

the condition `disabled={!issueTitle || !issueDescription}` is failing, here is the button (when failing)

```svelte
<button
  aria-label="Preview on GitHub"
  class="[...] bg-[var(--pd-button-disabled)] text-[var(--pd-button-disabled-text)]"
  disabled=""
  type="button"
/>
```

The disabled is `""` but might be considered to true if not explicit to false ? very weird... I don't understand why it had this empty string...

I really don't understand, creating a derived state fixed it :/

### How to test this PR?

### DIFF
--- a/packages/renderer/package.json
+++ b/packages/renderer/package.json
@@ -37,7 +37,7 @@
     "monaco-editor": "^0.52.2",
     "postcss": "^8.5.1",
     "postcss-load-config": "^6.0.1",
-    "svelte": "5.19.4",
+    "svelte": "5.19.5",
     "svelte-check": "^4.1.4",
     "svelte-eslint-parser": "^0.43.0",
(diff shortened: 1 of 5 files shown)


benoitf:
I would close the PR as it was this time a regression in svelte
https://github.com/podman-desktop/podman-desktop/pull/10911 update went fine

PR #10901, opened 2025-01-30
chore: set tailwind framework as being the 'winner of the CSS'

benoitf:
### What does this PR do?
instead of trying to revert some styles from docusaurus, tell tailwindcss to behave with important styling so it is always winning the rendering style

will make easier to move to tailwindcss v4 as the layers rendering is changing

note: there are some small UI changes but I think it's better to stick to tailwindcss rules over any docusaurus styling.

### Screenshot / video of UI

Check using netlify or argos-ci

#10795 

### How to test this PR?

Check using netlify or argos-ci

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -179,31 +179,39 @@ a {
 }
 
 /* headings */
-.theme-doc-markdown.markdown h1 {
+
+article .markdown h1 {
   font-size: 1.5rem;
   font-weight: bolder;
 }
 
-.theme-doc-markdown.markdown h2 {
+nav h2,
+article .markdown h2 {
   font-size: 1.2rem;
   font-weight: bolder;
 }
 
+nav h3,
+article .markdown h3 {
+  font-size: 1.1rem;
+  font-weight: bolder;
+}
+
 /* lists */
-.theme-doc-markdown.markdown ul,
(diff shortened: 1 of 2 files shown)


axel7083 (approved):

gastoner (approved):

deboer-tim:
Found two issues while testing locally:
- Text in the 'Experiment with' section on the extensions/ai-lab page is not visible in light or dark mode.
- Section titles in the blogs and documentation aren't styled or too similar to paragraph text now.

Otherwise, this is much better. 👍🏼

deboer-tim:
Bulleted or numbered lists also missing styling.

benoitf:
@deboer-tim another bunch of fixes has been applied

deboer-tim (approved):
AI lab, doc headings, bullet/numbered lists all fixed, thank you!

The headings in blogs still have no/little styling, could just use a bold or +1 font size but ok if this is a follow-up.

Before:

Screenshot 2025-01-30 at 1 54 37 PM

After:

Screenshot 2025-01-30 at 1 54 46 PM

benoitf:
@deboer-tim ok now it should be also fixed

deboer-tim (approved):
Tested, worked for me. Thanks!

Issue #10903, opened 2025-01-30
Update existing Kubernetes pages to pod standard


deboer-tim:
### Is your enhancement related to a problem? Please describe

When doing #10852 we updated to newer standards for svelte 5 and vi.mocking - which is great but makes pods 'unique'. Doing the same throughout will take time, but there is some low hanging fruit I can do while I'm here.

### Describe the solution you'd like

Update Kubernetes navigation stores to new mocking.
Update Kubernetes table columns to svelte 5.

deboer-tim:
Columns all updated to Svelte 5 and navigation tests updated as well. Found one small nit when reviewing everything (#11289) but closing this issue as everything directly related is complete.

PR #10906, opened 2025-01-30
chore: svelte 5 kubernetes columns

deboer-tim:
### What does this PR do?

Update all the Kubernetes columns to svelte 5: $props, $state, and onclick

### Screenshot / video of UI

N/A

Part of #10903.

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnActions.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnActions.svelte
@@ -2,7 +2,10 @@
 import ConfigmapSecretActions from './ConfigMapSecretActions.svelte';
 import type { ConfigMapSecretUI } from './ConfigMapSecretUI';
 
-export let object: ConfigMapSecretUI;
+interface Props {
+  object: ConfigMapSecretUI;
+}
+let { object }: Props = $props();
 </script>
 
 <ConfigmapSecretActions configMapSecret={object} on:update />
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnName.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnName.svelte
@@ -4,7 +4,10 @@ import { router } from 'tinro';
 import { ConfigMapSecretUtils } from './configmap-secret-utils';
 import type { ConfigMapSecretUI } from './ConfigMapSecretUI';
 
-export let object: ConfigMapSecretUI;
+interface Props {
+  object: ConfigMapSecretUI;
+}
+let { object }: Props = $props();
 
 function openDetails(): void {
   const configmapSecretUtils = new ConfigMapSecretUtils();
@@ -22,7 +25,7 @@ function openDetails(): void {
 }
 </script>
 
-<button class="hover:cursor-pointer flex flex-col max-w-full" on:click={openDetails}>
+<button class="hover:cursor-pointer flex flex-col max-w-full" onclick={openDetails}>
   <div class="text-sm text-[var(--pd-table-body-text-highlight)] max-w-full overflow-hidden text-ellipsis">
     {object.name}
   </div>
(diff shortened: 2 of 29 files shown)


axel7083 (approved):
Great! 🚀

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnActions.svelte
```diff
@@ -2,7 +2,11 @@
 import ConfigmapSecretActions from './ConfigMapSecretActions.svelte';
 import type { ConfigMapSecretUI } from './ConfigMapSecretUI';
 
-export let object: ConfigMapSecretUI;
+let {
+  object,
+}: {
+  object: ConfigMapSecretUI;
+} = $props();
```
benoitf: AFAIK we use a Props interface when declaring properties now

benoitf (commented):
non blocking: wondering if before svelte5, it should not be merged into a single object having name and details property as for most of the ojects it's a copy paste so it could just a Default column object with a name, namespace and details property

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnActions.svelte
deboer-tim: We do? We have both in the codebase and at least some of this style weren't from me. 😀 @axel7083 do you have a preference?

deboer-tim:
> non blocking: wondering if before svelte5, it should not be merged into a single object having name and details property as for most of the ojects it's a copy paste so it could just a Default column object with a name, namespace and details property

I agree on trying to reduce boilerplate, but not sure I understand the suggestion. Each page (set of columns) has a different type so broad sharing would have to be something generic like:
```
interface Props<Type> {
  object: Type;
}
```
and each file has new import and:
`let { object }: Props<DeploymentUI> = $props();`
?

benoitf:
The common type was for example the details link and the name and the namespace

so assuming all these UI objects are being assigned these properties

we could do 

MyGenericK8STable.svelte

interface Props {
  name: string;
  details: string;
  namespace: string;
}
...


and we provide DeploymentUI, FooUI, FooBarUI object (that have the right/expected properties)

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnActions.svelte
benoitf: we started without and switched to Props as it's suggested by the svelte documentation

https://svelte.dev/docs/svelte/typescript#The-Component-type

deboer-tim:
Rebased on latest, using the Props syntax.

@benoitf I started down the route you're suggesting. I think we could have expected properties or a parent KubernetesUIObject interface that the KubernetesObjectsList transformer outputs. I don't think it will save much code right now other than being able to share a name column because everything else uses resource-specific properties, but it is a good idea and would give us the path to the details page for/from any object. That said, I would prefer to self-assign a new issue for it.

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnActions.svelte
axel7083: Yes, this is my preference, but I am not blocking it (since it is not in the code guidelines yet)

My 2 cents would be to go further.

```
.
└── table/
    └── columns/
        ├── Foo.svelte
        ├── Bar.svelte
        └── props.ts
```

Instead of redefining the same props again and again, since all the columns for a given table have the same, we should group them in a folder, create a `props.ts` file with the props definition in it, that each component should import and use. :)

axel7083 (approved):
Re-approving

> I forgot this PR, so **don't take this is as a blocking**, but I made a suggestion (3 minutes ago) on how we could improve the props interface duplicate code https://github.com/podman-desktop/podman-desktop/pull/10906#discussion_r1946237346

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretColumnActions.svelte
deboer-tim: Like this?

Screenshot 2025-02-07 at 8 57 30 AM

Yeah, I like it better. I'd rather commit this and do it separately though, so I opened #11035.

Might make sense to do something similar for the details pages in the future too.

axel7083: yeah, follow-up PR! As I said this is more like improvement area!

deboer-tim:
@SoniaSandler @dgolovin could I get a second review? Would like to merge this and get working on the follow-up ideas.

SoniaSandler (approved):
LGTM codewise

Issue #10916, opened 2025-01-31
Resources obtained by informers do not contain kind/apiVersion


feloy:
### Bug description

The Kubernetes API does not set the kind and apiVersion fields for items of a list/watch request. It sets it only on a request for a specific object: https://github.com/kubernetes/kubernetes/issues/3030

In experimental mode, we will want to use the objects in the informers cache (obtained by watch), to display the details of the object. This is problematic at least for the `KubeEditYAML.svelte` component, which will display the YAML without these fields, and which will send back the content of the manifest when updating the resource without kind/apiVersion, which makes the update fail.

### Operating system

all

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

```
$ kubectl get --raw /apis/apps/v1/namespaces/default/deployments |json_pp
{
   "apiVersion" : "apps/v1",
   "kind" : "DeploymentList",
   "metadata" : {
      "resourceVersion" : "365718"
   },
   "items" : [
      {
          --> no apiVersion / kind here <---
         "metadata" : { ... },
         "spec" : { ... }
      }
   ]
}
```

### Relevant log output

```shell

```

xbabalov:
A unit test to cover this should probably be part of the fix as well.

PR #10928, opened 2025-02-01
feat: create a KubernetesObjectsList

feloy:
### What does this PR do?

- create a KubernetesObjectsList
- use it for DeploymentsList
- use it for ConfigMapSecretList
- start informers for configmaps and secrets

These changes are made incrementally in the different commit, to facilitate review.

I can split the PR into several ones if you prefer.

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature (to be continued)

### DIFF
--- a/packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
+++ b/packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
@@ -1,0 +1,145 @@
+<script lang="ts">
+import { faTrash } from '@fortawesome/free-solid-svg-icons';
+import type { KubernetesObject } from '@kubernetes/client-node';
+import type { TableColumn, TableRow } from '@podman-desktop/ui-svelte';
+import { Button, FilteredEmptyScreen, NavPage, Table } from '@podman-desktop/ui-svelte';
+import { onDestroy, onMount, type Snippet } from 'svelte';
+import { type Readable, type Unsubscriber, type Writable } from 'svelte/store';
+
+import type { IDisposable } from '../../../../main/src/plugin/types/disposable';
+import { withBulkConfirmation } from '../actions/BulkActions';
+import KubeActions from '../kube/KubeActions.svelte';
+import { listenResources } from '../kube/resources-listen';
+import KubernetesCurrentContextConnectionBadge from '../ui/KubernetesCurrentContextConnectionBadge.svelte';
+
+export interface Kind {
+  resource: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  transformer: (o: KubernetesObject) => any;
+  delete: (name: string) => Promise<void>;
+  searchPatternStore: Writable<string>;
+  legacyObjectStore: Readable<KubernetesObject[]>;
+}
+
+interface Props {
+  kind: Kind;
+  searchTerm: string;
+  singular: string;
+  plural: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  icon: any;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  columns: TableColumn<any>[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  row: TableRow<any>;
+
+  emptySnippet: Snippet;
+}
+
+let { kind, singular, plural, icon, searchTerm, columns, row, emptySnippet }: Props = $props();
+
+let resources = $state<KubernetesObject[] | undefined>(undefined);
+let resourceListener: IDisposable | undefined;
+let legacyUnsubscriber: Unsubscriber;
+
+const objects = $derived(resources?.map(object => kind.transformer(object)) ?? []);
+
+$effect(() => {
+  kind.searchPatternStore.set(searchTerm);
+});
+
+onMount(async () => {
+  resourceListener = await listenResources(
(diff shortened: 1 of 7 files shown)


benoitf (approved):

packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
```diff
@@ -0,0 +1,159 @@
...
+import KubeActions from '../kube/KubeActions.svelte';
+import { listenResources } from '../kube/resources-listen';
+import KubernetesCurrentContextConnectionBadge from '../ui/KubernetesCurrentContextConnectionBadge.svelte';
+
+export interface Kind {
+  resource: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  transformer: (o: KubernetesObject) => any;
+  delete: (name: string) => Promise<void>;
+  isResource: (o: KubernetesObject) => boolean;
```
deboer-tim, line 20: This part feels a little bit 'internal' to me - you provide one or more sets with a delete function for each set, but then you also need to provide a function to say whether an object came from a set. Can't we just look at each set again to find the object? (so we wouldn't need this property)

packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
```diff
@@ -0,0 +1,159 @@
...
+  searchPatternStore: Writable<string>;
```
deboer-tim: For the current config maps & secrets we only used two search pattern stores because of the underlying implementation - there's really only one filter bar / one search term that we apply to both. Could this be renamed legacySearchPatternStore to be clear the intent is that it would be removed later?

IMHO having the search term / pattern store alongside each object store was one of the things we got 'wrong' in the current implementation. We don't have a shared list of filtered deployments, we have multiple pages that each supporting filtering. i.e. filtering should be managed by KubernetesObjectsList here, not the stores or shared between lists.

deboer-tim (commented):
Sorry, took me longer to review because this wasn't the change I was expecting and I had to get my head around it and the implications - but it was a good thing, I like this direction 👍🏼 and just have 2 suggestions.

Each page might need different actions or other customizations, but that's easy enough to extend later.

packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
feloy: When https://github.com/podman-desktop/podman-desktop/issues/10916 is fixed, we will be able to find internally what is the kind of the object (by looking the kind field and comparing with a kind field passed in this structure), and won't need anymore the `isResource`

packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
feloy: ok, renaming to `legacySearchPatternStore`, and I agree that we can find a better pattern for the search

axel7083 (approved):

packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
deboer-tim: Thinking about it a bit more I don't think #10916 would directly help, because the objects passed to delete are already transformed (like PVCUI, PodUI, etc) and not the underlying Kube object. We could also just keep track internally of which resource array each object came from.

Either way I think we can remove the need for isResource, but it is not blocking. I'm fine if you can open a follow-up issue.

packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
```diff
@@ -0,0 +1,159 @@
...
+        {
+          searchTermStore: kind.legacySearchPatternStore,
+        },
+        (updatedResources: KubernetesObject[]) => {
+          resources[kind.resource] = updatedResources;
+        },
+      ),
+    );
+
+    legacyUnsubscribers.push(kind.legacyObjectStore.subscribe(o => (resources[kind.resource] = o)));
```
deboer-tim: Received an error repeatedly while testing:
`Assignment to `secrets` property (packages/​​renderer/​​src/​​lib/​​objects/​​KubernetesObjectsList.svelte:68:68) will evaluate to the right-hand side, not the value of `secrets` following the assignment. This may result in unexpected behaviour.https://svelte.dev/e/assignment_value_stale`

I think you meant to have { } brackets instead of ( ): `{ resources[kind.resource] = o })`.

deboer-tim (approved):
Note the 2 comments, but otherwise LGTM and thanks for the changes!

packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
feloy: Thanks for the catch. Fixed

feloy: > Thinking about it a bit more I don't think https://github.com/podman-desktop/podman-desktop/issues/10916 would directly help, because the objects passed to delete are already transformed (like PVCUI, PodUI, etc) and not the underlying Kube object. We could also just keep track internally of which resource array each object came from.

> Either way I think we can remove the need for isResource, but it is not blocking. I'm fine if you can open a follow-up issue.

-> #11002

PR #10940, opened 2025-02-03
chore: replace unknown/any by real type

benoitf:
### What does this PR do?
replace unknown/any by real type
for backward compliance, allow HostPort to be either a string or a number

#10600 

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2676,6 +2676,17 @@ declare module '@podman-desktop/api' {
     PathOnHost: string;
   }
 
+  interface HostConfigPortBindings {
+    [port: string]: {
+      HostPort?: number | string;
+      HostIp?: string;
+      HostPortEnd?: number;
+      Port?: number;
+      IP?: string;
+      Proto?: number;
+    }[];
+  }
+
   interface HostConfig {
     AutoRemove?: boolean;
     Binds?: string[];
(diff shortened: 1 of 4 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -2676,6 +2676,17 @@ declare module '@podman-desktop/api' {
     PathOnHost: string;
   }
 
+  interface HostConfigPortBindings {
+    [port: string]: {
+      HostPort?: number | string;
```
axel7083, line 2681: I was looking at the documentation, trying to find out if `HostPort` was mandatory or not (should be?)

We have 3 places where we have a `PortBindings`, and they seems to only be (HostPort - HostIp)

1. [ContainerCreate](https://docs.podman.io/en/latest/_static/api.html#tag/containers-(compat)/operation/ContainerCreate)

image

2. [ContainerList](https://docs.podman.io/en/latest/_static/api.html#tag/containers-(compat)/operation/ContainerList)

image

3. [ContainerInspectLibpod](https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerInspectLibpod)

image

I cannot found much about the other properties ?

benoitf: source was there https://github.com/moby/moby/blob/96ded2a1badcc5cb99d6d2ab7904d6f9e527cc4b/libnetwork/types/types.go#L71-L77

axel7083: I am questioning which one should be taken, the internal code ? is it something really exposed ?

Because this would let the user think it can use it, when the libpod api says no?

Issue #10734, opened 2025-01-20
bug(Dropdown): using state as value break the selection


axel7083:
### Bug description

While working on migrating PullImage to svelte5 I faced a weird behaviour.

![Image](https://github.com/user-attachments/assets/b615e800-4730-4f5c-9ffc-7837744e53aa)

The Dropdown component was not displaying the container engine I selected.

 Let's take a deeper look at how we use the Dropdown component.

https://github.com/podman-desktop/podman-desktop/blob/eaa421aff79ec65f5e739287b6f1179e158d05f6/packages/renderer/src/lib/image/PullImage.svelte#L305-L313

We bind `selectedProviderConnection` which is defined as followed

https://github.com/podman-desktop/podman-desktop/blob/eaa421aff79ec65f5e739287b6f1179e158d05f6/packages/renderer/src/lib/image/PullImage.svelte#L39

In svelte<5 this is not an issue, as the `selectedProviderConnection` is not a proxy object, but if we change the selectedProviderConnection definition as following

```
-  let selectedProviderConnection: ProviderContainerConnectionInfo | undefined; 
+ let selectedProviderConnection: ProviderContainerConnectionInfo | undefined = $state(); 
```

The object is now a **ProxyObject**. Let's take a look at how the Dropdown is finding the label of a given value to display

https://github.com/podman-desktop/podman-desktop/blob/52e04a96ee4cf2ba1dc8e503290559e19cd4a642/packages/ui/src/lib/dropdown/Dropdown.svelte#L50

The `options.find(o => o.value === value)` **is failing** if we try to compare the object with its proxy. 

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

N/A

### Relevant log output

```shell

```

axel7083:
Some solution are:

- Restrict the value to string, let's avoid to complexify the Dropdown component and force the parent to found a unique string for each options.
- Make the Dropdown component check if one of the values are ProxyObject and use `$state.snapshot()` to compare them (not sure it will work)

## Edit

Solution two (using state.snapshot) does not seems to be working, probably doing some kind of deep copy, therefore not be the same object.

axel7083:
Here is an example on how to use the Dropdown in svelte5, we need to avoid the binding, and use a string as value, so the comparison can be done nicely

```diff
<Dropdown
  id="providerChoice"
  name="providerChoice"
+ onChange={(value) => {
[...]

PR #10950, opened 2025-02-03
refactor(dropdown): create container connections dropdown component

axel7083:
### What does this PR do?

Factorising all code using the `Dropdown` component for selecting a container engine.

> To be able to use the `Dropdown` component in svelte5, **we need** to use a primitive for the value, so I am using `ProviderContainerConnection#endpoint#socketPath` from my understanding this is unique to each connection?

### Screenshot / video of UI

N/A

Part of https://github.com/podman-desktop/podman-desktop/issues/10734

### How to test this PR?


packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
```diff
@@ -282,15 +283,12 @@ function navigateToContainers(): void {
           for="providerConnectionName"
           class="block mb-2 font-semibold rounded-sm text-[var(--pd-content-card-header-text)]"
           >Container engine:</label>
-        <Dropdown
+        <ContainerConnectionDropdown
           class="w-full"
           name="providerChoice"
```
gastoner, line 288: Also add the `id`?

packages/renderer/src/lib/forms/ContainerConnectionDropdown.svelte
```diff
@@ -0,0 +1,48 @@
...
+import { Dropdown } from '@podman-desktop/ui-svelte';
+
+import type { ProviderContainerConnectionInfo } from '/@api/provider-info';
+
+interface Props {
+  connections: ProviderContainerConnectionInfo[];
+  value: ProviderContainerConnectionInfo | undefined;
+  name?: string;
+  id?: string;
+  onchange?(value: ProviderContainerConnectionInfo | undefined): void;
```
gastoner, line 11: onChange ? or is it due to keyword things?

gastoner (approved):
LGTM codewise

packages/renderer/src/lib/forms/ContainerConnectionDropdown.svelte
axel7083: hummm, good question, I am not sure we decided yet to use which cases rules :/

packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
axel7083: I would rather act as a drop-in replacement here, if the previous component were not using it I will keep as it is

benoitf (commented):
question: I'm curious why I need to provide the connections to the component

shouldn't it read directly from the stores the existing provider connections ?

If I place the component on my page, I don't need to be worried on how I collect the list, I defered that to this component

benoitf (changes requested):
> To be able to use the Dropdown component in svelte5, we need to use a primitive for the value, so I am using ProviderContainerConnection#endpoint#socketPath from my understanding this is unique to each connection?

I would not rely on that as in the past it was the case but at some point, on macOS they switched to a common path, then changed it again. So it can come again. And when it'll come again it would break everything.

This is why in the past we had to change the equals() from socketPath to the combination of the socketPath and the providerconnection name

axel7083:
> question: I'm curious why I need to provide the connections to the component

You can filter them, the component should not be responsible of fetching them, let's say the parent only want you to select the Podman providers.

axel7083:
> I would not rely on that as in the past it was the case but at some point, on macOS they switched to a common path, then changed it again. So it can come again. And when it'll come again it would break everything.
> 
> This is why in the past we had to change the equals() from socketPath to the combination of the socketPath and the providerconnection name

Thanks @benoitf  for the feedback, I was really not sure, in AI Lab and Quadlet we use a combination of `providerId:connectionName` would this be suitable ?

benoitf:
> You can filter them, the component should not be responsible of fetching them, let's say the parent only want you to select the Podman providers.

then I have an optional filter method on this component

deboer-tim (commented):
I thought the intention was to make Dropdown only handle strings, but this doesn't touch that. Was the intention to consolidate all of these, then there is only 1 place + dropdown to change later?

nit: I don't love adding a new /forms folder, it feels like just 'not more /ui' . :) Seems like we could use a little cleanup and organization of what goes in each folder (but not part of this issue).

axel7083:
> I thought the intention was to make Dropdown only handle strings, but this doesn't touch that. Was the intention to consolidate all of these, then there is only 1 place + dropdown to change later?

Yes, instead of having to change all the place at once, we will only have one

> nit: I don't love adding a new /forms folder, it feels like just 'not more /ui' . :) Seems like we could use a little cleanup and organization of what goes in each folder (but not part of this issue).

Me too, but I don't know where :/

axel7083:
> I would not rely on that as in the past it was the case but at some point, on macOS they switched to a common path, then changed it again. So it can come again. And when it'll come again it would break everything.

@benoitf I am facing another issue, the parent components are providing `Array<ProviderContainerConnectionInfo>` which does not contain the providerId.

With this implementation the component only have access to the following

https://github.com/podman-desktop/podman-desktop/blob/48e6391ead8860d6fcef738d54ac768044f6325f/packages/api/src/provider-info.ts#L32-L42

I am tempted to use `${vmType}:${type}:${name}` as a unique key but this feel a bit hacky 🤔

benoitf:
maybe we should add internally a `key` or `id` on the objects so if you need a unique id we could rely on the id

axel7083:
> maybe we should add internally a `key` or `id` on the objects so if you need a unique id we could rely on the id

Oh great idea!

benoitf (approved):
so, the 'key' has been changed from connection.path to `connection.type:connection.name` so there should be unique values

thanks for the changes

Issue #10598, opened 2025-01-10
Remove disablement of no-explicit any from podman extension


jeffmaury:
### Is your enhancement related to a problem? Please describe

Impacted files:

- [x] packages/extension/src/compatibility-mode.spec.ts (https://github.com/podman-desktop/podman-desktop/pull/10994)
- [x] packages/extension/src/extension.spec.ts (https://github.com/podman-desktop/podman-desktop/pull/10995)
- [x] packages/extension/src/extension.ts (https://github.com/podman-desktop/podman-desktop/pull/11017)
- [x] packages/extension/src/podman-configuration.ts (https://github.com/podman-desktop/podman-desktop/pull/11008)
- [x] packages/extension/src/podman-info-helper.spec.ts (https://github.com/podman-desktop/podman-desktop/pull/11010)
- [x] packages/extension/src/podman-remote-connections.spec.ts (https://github.com/podman-desktop/podman-desktop/pull/11011)
- [x] packages/extension/src/podman-remote-ssh-tunnel.spec.ts (https://github.com/podman-desktop/podman-desktop/pull/11011)
- [x] packages/extension/src/registry-setup.ts (https://github.com/podman-desktop/podman-desktop/pull/11012)
- [ ] packages/extension/src/util.ts (blocked by https://github.com/podman-desktop/podman-desktop/issues/10966)

### Describe the solution you'd like

Remove use of any type

feloy:
In `packages/extension/src/extension.ts`, `any` is used in:

```
export async function createMachine(
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  params: { [key: string]: any },
  logger?: extensionApi.Logger,
  token?: extensionApi.CancellationToken,
): Promise<void> {

```

to implement the interface from `extension.d.ts`:
```
  export interface KubernetesProviderConnectionFactory extends ProviderConnectionFactory {
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    create?(params: { [key: string]: any }, logger?: Logger, token?: CancellationToken): Promise<void>;
  }
```

so I suppose the use of `any` cannot be removed. @benoitf WDYT?

benoitf:
the place where we can't remove it is extension.d.ts but it might be possible to trick somehow the extension.ts file to use unknown

feloy:
Closing, as the remaining is out of scope and covered by #10966

PR #10952, opened 2025-02-03
chore: include electron.d.ts for podman extension

feloy:

### What does this PR do?

Include electron.d.ts for podman extension, so process type comes from electron.d.ts, not node types.

### Screenshot / video of UI

N/A

Part of #10598 

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/util.ts
+++ b/extensions/podman/packages/extension/src/util.ts
@@ -47,12 +47,10 @@ export interface RunOptions {
 }
 
 export function getAssetsFolder(): string {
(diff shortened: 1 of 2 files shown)


benoitf (commented):
issue: this fixes the issue but also highlights an underlying problem regarding extensions. They are not supposed to know that Electron is present. They could very well run in a separate process. In fact, it's like an anti-pattern. If the extension ever needs to know something, it's up to the framework to provide that.

so we need to think on how to reverse the logic so that the extension should never imports any electron stuff

feloy:
Closing this PR, I created an issue https://github.com/podman-desktop/podman-desktop/issues/10966

Issue #10560, opened 2025-01-08
Implementation of UX Cronjobs


nichjones1:
### Is your enhancement related to a problem? Please describe

Implementation of UX ChronJob and Pods design here: https://github.com/podman-desktop/podman-desktop/issues/10559

### Describe the solution you'd like

Implementation of UX ChronJob and Pods design

cdrage:
I believe this issue should just be CronJob?

Pods is already being worked on by @deboer-tim and in the issue https://github.com/podman-desktop/podman-desktop/issues/8818 

There would be a lot of code for both cronjob and pods in a single implementation.

I'm also confused as to why we are adding CronJobs first and not Jobs?

cc @slemeur @Firewall @feloy

cdrage:
From sprint planning + slack discussions, I have:

- Renamed this to CronJobs implementation
- This request came from Discord
- We should implement Jobs if we add CronJobs

I have also linked this to the epic here: https://github.com/podman-desktop/podman-desktop/issues/7192

PR #10955, opened 2025-02-03
chore: cronjob use native functions rather than custom API call

cdrage:
chore: cronjob use native functions rather than custom API call

### What does this PR do?

Slight refactor, cronjobs can be called natively using batchV1Api rather
than using a custom call.

Does the same function as the previous one, but instead uses the native
call.

### Screenshot / video of UI

N/A

Part of https://github.com/podman-desktop/podman-desktop/issues/10560

### How to test this PR?


deboer-tim (approved):
LGTM

gastoner (approved):
Codewise LGTM

Issue #10954, opened 2025-02-03
Simplify exception handling in ipcMain.handle


dgolovin:
### Bug description

The multiple accounts of `Object can’t be cloned` error were reported by QE and the users. The reason for that is how original error is transformed with `encodeIpcError` function code below

```
  encodeIpcError(e: unknown): { name?: string; message: unknown; extra?: Record<string, unknown> } {
    let builtError;
    if (e instanceof Error) {
      builtError = { name: e.name, message: e.message, extra: { ...e } };
    } else {
      builtError = { message: e };
    }
    return builtError;
  }

  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  ipcHandle(channel: string, listener: (event: IpcMainInvokeEvent, ...args: any[]) => Promise<void> | any): any {
    ipcMain.handle(channel, async (...args) => {
      try {
        return { result: await Promise.resolve(listener(...args)) };
      } catch (e) {
        return { error: this.encodeIpcError(e) };
      }
    });
  }
```

The original error gets lost when pcMain.handle trying to pass encoded error above from main to renderer process if there are Functions, Promises, Symbols, WeakMaps, or WeakSets in it or its properties down the hierarchy of extra property. For example, when requesting releases from GitHub.com and request rated limit is exceeded the Error instance would have response and request fields in it that cannot be cloned. The error reported to the user would be `Error when selecting a version: Error: Error invoking remote method 'cli-tool-registry:selectCliToolVersionToInstall': Error: An object could not be cloned.`.

If I change the code above to just 

```
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  ipcHandle(channel: string, listener: (event: IpcMainInvokeEvent, ...args: any[]) => Promise<void> | any): any {
    ipcMain.handle(channel, async (...args) => {
        return { result: await Promise.resolve(listener(...args)) };
    });
  }
```

The error would be 

```
Error when selecting a version: Error: Error invoking remote method 'cli-tool-registry:selectCliToolVersionToInstall': HttpError: API rate limit exceeded for 73.222.44.218. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) - https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting
```

If I exclude encodeIpcError call and pass error directly

```
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
[...]

Issue #10530, opened 2025-01-06
Clicking Compose Setup throws error: An object could not be cloned


subshade:
### Bug description

When click Compose Setup, it fails to download compose and throw an error in dev console.
![Image](https://github.com/user-attachments/assets/a0b4fe14-6383-47ff-9bb4-4a8e360bc6bc)

### Operating system

Windows 10 22H2 19045.5247

### Installation Method

Installer from website/GitHub releases

### Version

1.15.0

### Relevant log output

```shell
Onboarding.svelte:217 Error: Error invoking remote method 'command-registry:executeCommand': Error: An object could not be cloned.
C	@	Onboarding.svelte:217
$	@	Onboarding.svelte:204
await in $		
c	@	Onboarding.svelte:160
M	@	Onboarding.svelte:263
await in M		
N	@	Onboarding.svelte:252
click	@	Onboarding.svelte:491
Br	@	misc.js:39
(anonymous)	@	Button.svelte:76
(anonymous)	@	events.js:66
Hg	@	shared.js:44
a	@	events.js:65
Show less
```

### Additional context

call stack when click compose setup
![Image](https://github.com/user-attachments/assets/8dc099ae-16e6-43e7-add9-27a62eafbc95)
![Image](https://github.com/user-attachments/assets/7062ba61-73c4-4d08-b5c4-38a437704210)
after that, it throws an error
![Image](https://github.com/user-attachments/assets/f487f5e9-8083-43d7-8a33-8ec246e9eed7)

benoitf:
thanks for the report @subshade

dgolovin:
1. The reason for the 'An object could not be cloned` error is #10954.
2. There is also issue in CLI Tools page that ignores errors (network related or `API rate limit exceeded` from github.com) when requesting available versions for the tool.

PR #10964, opened 2025-02-04
fix: update encodeIpcError method to correctly handle Error instance

dgolovin:
### What does this PR do?

Fixes 'An object could not be cloned' in case error instance contain non clonable properties  anywhere deep in the object structure and let electron implementation to pass message property to the renderer side.

Fix for #10954.
Related to #10530.

### How to test this PR?


jeffmaury (commented):
What issue is it supposed to fix ?

dgolovin:
> What issue is it supposed to fix ?

@jeffmaury I updated description.

jeffmaury (commented):
Can we get a unit test ?

benoitf (commented):
looks like the test is still expected

benoitf (approved):

packages/main/src/plugin/index.spec.ts
```diff
@@ -387,3 +387,29 @@ test('Pull image creates a task', async () => {
...
+  const errorInstance = new Error('error');
+  createTaskSpy.mockImplementation(() => {
+    throw errorInstance;
+  });
+
+  const handleReturn = await handle(undefined, '1');
+  expect(handleReturn.error).toEqual(errorInstance);
+});
+
+test('ipcMain.handle returns caught error as objects message property if it isn not instance of error', async () => {
```
benoitf:
```suggestion
test('ipcMain.handle returns caught error as objects message property if it is not instance of error', async () => {
```

jeffmaury (approved):
LGTM

PR #10980, opened 2025-02-04
feat: ask to install kind when creating cluster

### LINKED ISSUE #9318: Users should know that kind cli is required before trying to create a cluster

I was trying to create kind cluster, but I did not have cli installed. In order to find this out, I had to check the error message after I clicked on create Kind cluster.

Improve the user experience in kind workflow. I can image that some information would be available to user, the cli is required before trying to create a kind cluster. also, the Create button should not be enabled. 

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

When the user wants to create a cluster, if kind is not installed, asking to install it.

### Screenshot / video of UI

https://github.com/user-attachments/assets/ad49eaae-e7a2-4f55-a1ca-2e76ea434913

Fixes https://github.com/podman-desktop/podman-desktop/issues/9318

### How to test this PR?

### DIFF
--- a/extensions/kind/src/extension.spec.ts
+++ b/extensions/kind/src/extension.spec.ts
@@ -22,6 +22,7 @@ import type * as extensionApi from '@podman-desktop/api';
 import * as podmanDesktopApi from '@podman-desktop/api';
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
 
+import { createCluster } from './create-cluster';
 import * as extension from './extension';
 import type { KindGithubReleaseArtifactMetadata } from './kind-installer';
 import { KindInstaller } from './kind-installer';
@@ -38,6 +39,9 @@ vi.mock('./image-handler', () => {
     }),
   };
 });
+vi.mock('./create-cluster', () => ({
+  createCluster: vi.fn(),
+}));
 
 vi.mock('./kind-installer');
 
@@ -45,6 +49,7 @@ vi.mock('@podman-desktop/api', () => ({
   window: {
     withProgress: vi.fn(),
     createStatusBarItem: vi.fn(),
+    showInformationMessage: vi.fn(),
   },
   cli: {
     createCliTool: vi.fn(),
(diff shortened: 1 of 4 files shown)


SoniaSandler (commented):
When I try to uninstall Kind after installing it with the new prompt, I get `Unable to uninstall Kind. Error: Error: Cannot uninstall kind. No version detected`. I need to stop and start the extension to be able to uninstall it. Uninstalling it after installing it through the CLI tools works as intended.

axel7083:
> When I try to uninstall Kind after installing it with the new prompt, I get `Unable to uninstall Kind. Error: Error: Cannot uninstall kind. No version detected`. I need to stop and start the extension to be able to uninstall it. Uninstalling it after installing it through the CLI tools works as intended.

Hummm, I will take a look! Nice catch

axel7083:
This will need rebase after the following
- https://github.com/podman-desktop/podman-desktop/pull/11032

extensions/kind/src/extension.ts
```diff
@@ -65,18 +66,50 @@ let installer: KindInstaller;
...
   const disposable = provider.setKubernetesProviderConnectionFactory(
     {
-      create: (params: { [key: string]: unknown }, logger?: Logger, token?: CancellationToken) => {
-        if (kindPath) {
-          return createCluster(params, kindPath, telemetryLogger, logger, token);
+      create: async (params: { [key: string]: unknown }, logger?: Logger, token?: CancellationToken) => {
+        // if kind is not installed, let's ask the user to install it
+        if (!kindPath) {
+          const result = await window.showInformationMessage(
+            'Kind is not installed, do you can to install the latest version?',
```
SoniaSandler:
```suggestion
            'Kind is not installed, do you want to install the latest version?',
```

extensions/kind/src/extension.ts
```diff
@@ -65,18 +66,50 @@ let installer: KindInstaller;
...
+
+  // update cli tool
+  kindCli?.updateVersion({
+    version: removeVersionPrefix(latest.tag),
+    path: cliPath,
+    installationSource: 'extension',
+  });
+
+  // return new path
+  return cliPath;
```
SoniaSandler, line 90: question (non-blocking): is there a difference between returning `cliPath` from this function vs setting `kindPath` (e.g. `kindPath = cliPath`) inside it and returning nothing?

SoniaSandler (approved):
LGTM, tested and it works, both install and uninstall :tada:

extensions/kind/src/extension.spec.ts
```diff
@@ -484,3 +489,68 @@ describe('cli#uninstall', () => {
...
+    // get the create method
+    const { create } = vi.mocked(PROVIDER_MOCK.setKubernetesProviderConnectionFactory).mock.calls[0][0];
+    expect(create).toBeDefined();
+
+    await expect(() => {
+      return create?.({}, undefined, undefined);
+    }).rejects.toThrowError('Unable to create kind cluster. No kind cli detected');
+
+    expect(podmanDesktopApi.window.showInformationMessage).toHaveBeenCalledWith(
+      'Kind is not installed, do you can to install the latest version?',
```
axel7083:
```suggestion
      'Kind is not installed, do you want to install the latest version?',
```

extensions/kind/src/extension.ts
axel7083: Technically no, but I always have a preference to create function without sideeffect, so I prefer the caller to set the global variable than this function

jeffmaury (approved):
LGTM

Issue #10604, opened 2025-01-10
Remove disablement of no-explicit any from renderer package


jeffmaury:
### Is your enhancement related to a problem? Please describe

Impacted files:

- packages/renderer/src/lib/featured/FeaturedExtensionDownload.spec.ts
- packages/renderer/src/lib/featured/FeaturedExtensions.spec.ts
- packages/renderer/src/lib/feedback/SendFeedback.spec.ts
- packages/renderer/src/lib/image/ActionsMenu.spec.ts
- packages/renderer/src/lib/image/BuildImageFromContainerfile.spec.ts
- packages/renderer/src/lib/image/BuildImageFromContainerfileCard.spec.ts
- packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
- packages/renderer/src/lib/image/BuildImageFromContainerfileCards.spec.ts
- packages/renderer/src/lib/image/ImagesList.spec.ts
- packages/renderer/src/lib/image/ImportContainersImages.spec.ts
- packages/renderer/src/lib/image/LoadImages.spec.ts
- packages/renderer/src/lib/image/PullImage.spec.ts
- packages/renderer/src/lib/image/PushImageModal.spec.ts
- packages/renderer/src/lib/image/RecommendedRegistry.spec.ts
- packages/renderer/src/lib/image/RunImage.spec.ts
- packages/renderer/src/lib/image/SaveImages.spec.ts
- packages/renderer/src/lib/ingresses-routes/IngressesRoutesList.spec.ts
- packages/renderer/src/lib/ingresses-routes/IngressRouteEmptyScreen.spec.ts
- packages/renderer/src/lib/kube/KubePlayYAML.svelte
- packages/renderer/src/lib/learning-center/GuideCard.spec.ts
- packages/renderer/src/lib/learning-center/LearningCenter.spec.ts
- packages/renderer/src/lib/markdown/micromark-listener-handler.ts
- packages/renderer/src/lib/node/NodeEmptyScreen.spec.ts
- packages/renderer/src/lib/node/NodesList.spec.ts
- packages/renderer/src/lib/pod/DeployPodToKube.spec.ts

### Describe the solution you'd like

Remove use of any type

SoniaSandler:
All related PRs have been merged. Closing as completed

PR #10981, opened 2025-02-04
chore: remove @typescript-eslint/no-explicit-any usage from featured, feedback, and kube

SoniaSandler:

### What does this PR do?
This PR removes @typescript-eslint/no-explicit-any usage from featured, feedback, and kube in renderer

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/kube/KubePlayYAML.svelte
+++ b/packages/renderer/src/lib/kube/KubePlayYAML.svelte
@@ -37,8 +37,7 @@ let currentNamespace: string | undefined;
 let allNamespaces: V1NamespaceList;
 
 let playKubeResultRaw: string;
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-let playKubeResultJSON: any;
+let playKubeResultJSON: unknown;
 
 let userChoice: 'podman' | 'kubernetes' = 'podman';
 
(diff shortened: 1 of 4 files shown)


gastoner (approved):

packages/renderer/src/lib/kube/KubePlayYAML.svelte
```diff
@@ -334,12 +346,12 @@ function goBackToPodsPage(): void {
         <ErrorMessage class="text-sm" error={runError} />
       {/if}
 
-      {#if playKubeResultJSON}
+      {#if playKubeResultJSON && typeof playKubeResultJSON === 'object'}
```
benoitf: it looks that in the ts code we're already handing 
```
f (playKubeResultJSON && typeof playKubeResultJSON === 'object' && 'Pods' in playKubeResultJSON) {
          if (
```

question: I'm wondering then if we could have a playKubeResult object with proper type like
{Pods: []...}

so in the ts part, we do the checks and assign the value to playKubeResult

then in the svelte part, if the object is defined, we have the proper/expected type and we don't need to use all the checks again

benoitf:
hello, I don't see changes based on my previous feedback so I don't know if the new changes are ok ? or if it's missing.
The idea was to have a proper type so that the svelte/rendering part (not the TypeScript part of the svelte component) can just use the fields with proper typing and do not try to do instanceof multiple times

packages/renderer/src/lib/kube/KubePlayYAML.svelte
```diff
@@ -37,8 +37,8 @@ let currentNamespace: string | undefined;
 let allNamespaces: V1NamespaceList;
 
 let playKubeResultRaw: string;
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-let playKubeResultJSON: any;
+let playKubeResultJSON: unknown;
+let playKubeResult: { Pods?: unknown[] } | undefined = undefined;
```
SoniaSandler, line 41: @benoitf these are the new changes. Is this what you meant? I've also updated the svelte part

benoitf (approved):
yes thanks, somehow I was not seeing the expected data

PR #10986, opened 2025-02-05
chore: remove @typescript-eslint/no-explicit-any usage in renderer

SoniaSandler:

### What does this PR do?
This PR removes @typescript-eslint/no-explicit-any usage from ingresses-routes, learning-center, node, and pod in renderer

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/node/NodeEmptyScreen.spec.ts
+++ b/packages/renderer/src/lib/node/NodeEmptyScreen.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2024 Red Hat, Inc.
+ * Copyright (C) 2024-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/node/NodeEmptyScreen.spec.ts
```diff
@@ -44,12 +37,9 @@ vi.mock('../../stores/kubernetes-contexts-state', () => ({
 beforeEach(() => {
   vi.resetAllMocks();
   mocks.subscribeMock.mockImplementation(listener => {
-    listener(mocks.getCurrentKubeContextState());
```
jeffmaury, line 47: Is this change related to any removal ?

packages/renderer/src/lib/node/NodeEmptyScreen.spec.ts
```diff
@@ -59,14 +49,17 @@ test('Expect deployment empty screen', async () => {
...
 test('If kubernetesCurrentContextState is not empty, expect that the empty screen say you do not have permission to view the nodes', async () => {
-  mocks.getCurrentKubeContextState.mockReturnValue({
-    error: undefined,
-    reachable: true,
-    resources: {
-      pods: 0,
-      deployments: 0,
-    },
-  } as ContextGeneralState);
+  mocks.subscribeMock.mockImplementation(listener => {
```
jeffmaury, line 52: Same here ?

packages/renderer/src/lib/node/NodeEmptyScreen.spec.ts
SoniaSandler: @jeffmaury yes, the removal of lines 52 and 62-69, because without this update, the tests failed. The listener only uses the value returned from the function, so I provided it when needed, like in the test of the second comment

jeffmaury (approved):
LGTM

packages/renderer/src/lib/pod/DeployPodToKube.spec.ts
```diff
@@ -165,7 +164,7 @@ beforeEach(() => {
   listSimpleContainersByLabelMock.mockResolvedValue([simpleContainerInfo]);
 });
 
-async function waitRender(customProperties: any): Promise<void> {
+async function waitRender(customProperties: object): Promise<void> {
```
axel7083: object is no better than any here.

```suggestion
async function waitRender(customProperties:  Partial<ComponentProps<DeployPodToKube>>): Promise<void> {
```

The waitRender here take some properties that could be passed to the render function. So the render function takes `ComponentProps<T>`; but if we use this type, we will have to provide all  the props everytime we call it, so we can use Partial to make all props optional.

benoitf (approved):

axel7083 (approved):

PR #11014, opened 2025-02-06
chore: update vitest peer dependencies to be vite v6

benoitf:
### What does this PR do?
vitest rolled back from vite 6 to vite 5 during v2.1.6 --> v2.1.7
but we're using vite6 so keep the dependency as expected

### How to test this PR?


axel7083:
Do you have the link where they explain why they rolled back?

axel7083 (approved):

PR #11032, opened 2025-02-07
refactor(kind): extract cli logic from activate

axel7083:
### Changes

1. Remove scoped variables (`binaryVersion` and `binaryPath`) were scoped to `activate` and were used like global variables

https://github.com/podman-desktop/podman-desktop/blob/bbcc04a9ce18f9b3a5b273122f7239ed174bdefa/extensions/kind/src/extension.ts#L365-L372

Instead of using the `binaryVersion` we use the `CliTool#version` property, so this is a real global variable
Instead of using `binaryPath` we use the global variable `kindPath`

2. Cleaning `kindPath` & `kindCli` in `deactivate`: tests were a bit leaking between each other, (the scoped variable prevented the leak, but since removing them, tests were failing due to interacting with each other)

Required for https://github.com/podman-desktop/podman-desktop/pull/10980

### How to test this PR?

### DIFF
--- a/extensions/kind/src/extension.spec.ts
+++ b/extensions/kind/src/extension.spec.ts
@@ -20,7 +20,7 @@ import * as fs from 'node:fs';
 
 import type * as extensionApi from '@podman-desktop/api';
 import * as podmanDesktopApi from '@podman-desktop/api';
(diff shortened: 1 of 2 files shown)


gastoner (approved):
Codewise LGTM

PR #11038, opened 2025-02-07
chore: remove @typescript-eslint/no-explicit-any usage in markdown

SoniaSandler:

### What does this PR do?
Remove the explicit use of any in `markdown/micromark-listener-handler`.
Used the suggestion from https://github.com/podman-desktop/podman-desktop/pull/10981#discussion_r1944479907 to reduce the amount of checks

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/markdown/micromark-listener-handler.ts
+++ b/packages/renderer/src/lib/markdown/micromark-listener-handler.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,8 @@
 import { executeButtonCommand } from './component/micromark-button';
 import { executeExpandableToggle } from './component/micromark-expandable-section';
 
+const eventObject: { target?: object } = {};
+
 export function createListener(
   inProgressMarkdownCommandExecutionCallback: (
     command: string,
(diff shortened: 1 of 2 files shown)


benoitf:
still in draft ?

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -19,23 +19,42 @@
 import { executeButtonCommand } from './component/micromark-button';
 import { executeExpandableToggle } from './component/micromark-expandable-section';
 
+const eventObject: { target?: object } = {};
```
benoitf: question: should it be a variable outside of the function ?

benoitf (commented):
added some questions

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
SoniaSandler: it doesn't have to be outside, so I moved it to be inside the function

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -26,16 +26,35 @@ export function createListener(
...
+    let command: string | undefined;
+    let expandable: string | undefined;
+
+    if (
+      eventObject.target &&
+      'dataset' in eventObject.target &&
+      eventObject.target.dataset &&
+      typeof eventObject.target.dataset === 'object'
+    ) {
+      if ('command' in eventObject.target.dataset && typeof eventObject.target.dataset.command === 'string') {
```
benoitf: reduce copy/paste of eventObject.target.dataset with a temp variable

benoitf (commented):
it looks like we never use eventObject directly but always use  eventObject.target

so should we not have a new variable eventObjectTarget and it not defined, we abort early so in all remaining stuff we can assume eventObjectTarget is defined and do not need to check all the time if it's defined or not

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -26,16 +26,33 @@ export function createListener(
     value?: unknown,
   ) => void,
 ) {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return (e: any): void => {
+  return (e: unknown): void => {
+    let eventTarget: object | undefined = undefined;
```
axel7083: `object` ? what kind ?

SoniaSandler: @axel7083 just a general object to reduce the number of repeated type checks as we assert that `e.target` is an object in line 33.

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -26,16 +26,33 @@ export function createListener(
     value?: unknown,
   ) => void,
 ) {
```
axel7083: `createListener` should return `EventListener`

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -26,16 +26,33 @@ export function createListener(
...
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return (e: any): void => {
+  return (e: unknown): void => {
```
axel7083:
```suggestion
  return (e: Event): void => {
```

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
axel7083:
```suggestion
    let eventTarget: EventTarget;
```

You are doing a if else bellow setting the value, so typescript is smart enough to let you type it properly, since you `return` if you can't set a value.

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -26,16 +26,33 @@ export function createListener(
...
+    let eventTarget: object | undefined = undefined;
+
+    if (e && typeof e === 'object' && 'target' in e && e.target && typeof e.target === 'object') {
+      eventTarget = e.target;
```
axel7083:
```suggestion
    if (e.target && typeof e.target === 'object') {
      eventTarget = e.target;
```

axel7083 (commented):
Would be interesting to update the following

https://github.com/podman-desktop/podman-desktop/blob/9f8b322d1c979e7de852c84b6bcb193cba25d0fc/packages/renderer/src/lib/markdown/Markdown.svelte#L75

Since `Markdown` component is using the `createListener` function.

We can simplify to 

```diff
+ const eventListeners: EventListener[] = [];
```

benoitf (approved):

axel7083 (approved):

PR #11042, opened 2025-02-09
chore: fix unit tests for vitest v3

benoitf:
### What does this PR do?
some tests are invalid with vite v3 but passing with vitest v2
cleanup/try to rewrite the tests using vi.mocked, etc to make them pass

https://github.com/podman-desktop/podman-desktop/pull/11023

### How to test this PR?

### DIFF
--- a/extensions/docker/src/docker-context-handler.spec.ts
+++ b/extensions/docker/src/docker-context-handler.spec.ts
@@ -210,11 +210,11 @@ describe('getContexts', () => {
     vi.spyOn(fs.promises, 'readdir').mockResolvedValue([
       {
         isDirectory: () => true,
-        name: 'fcde2b2edba56bf408601fb721fe9b5c338d10ee429ea04fae5511b68fbf8fb9',
+        name: '2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae',
       } as fs.Dirent,
       {
         isDirectory: () => true,
-        name: '2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae',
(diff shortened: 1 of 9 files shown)


extensions/docker/src/docker-context-handler.spec.ts
```diff
@@ -210,11 +210,11 @@ describe('getContexts', () => {
     vi.spyOn(fs.promises, 'readdir').mockResolvedValue([
       {
         isDirectory: () => true,
-        name: 'fcde2b2edba56bf408601fb721fe9b5c338d10ee429ea04fae5511b68fbf8fb9',
+        name: '2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae',
```
jeffmaury, line 213: Not sure to get why this is required ?

jeffmaury (approved):
LGTM

ScrewTSW (approved):
LGTM

extensions/docker/src/docker-context-handler.spec.ts
benoitf: hello @jeffmaury it turns out that the test is fully broken (I don't know how it can work today)

the name 'foo' is encoded using sha256 into `2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae` and 'bar'
into `fcde2b2edba56bf408601fb721fe9b5c338d10ee429ea04fae5511b68fbf8fb9`

and how it's before this change means that foo and bar was the opposite values.

I suppose it's reusing some previous data in vite v2 while in vite3 the cleanup is really done so the order is not good (the test is broken)

PR #11058, opened 2025-02-10
fix: support multiple permissions requests

feloy:

### What does this PR do?

In Kubernetes experimental mode, support multiple permissions requests (see https://github.com/podman-desktop/podman-desktop/pull/11054 for more context)

### Screenshot / video of UI

N/A

Part of #10657 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/context-permissions-checker.ts
+++ b/packages/main/src/plugin/kubernetes/context-permissions-checker.ts
@@ -58,7 +58,7 @@ export interface ContextPermissionResult extends Permission {
 }
 
 export class ContextPermissionsChecker implements Disposable {
-  contextName: string;
+  #contextName: string;
   #kubeconfig: KubeConfigSingleContext;
   #client: AuthorizationV1Api;
   #request: ContextPermissionsRequest;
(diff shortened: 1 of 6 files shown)


packages/main/src/plugin/kubernetes/context-permissions-checker.ts
```diff
@@ -38,45 +38,56 @@ export interface ContextPermissionsRequest {
 }
 
 // ContextResourcePermission is the permission for a specific resource
```
axel7083: need to update the comment

jeffmaury (approved):
LGTM

packages/main/src/plugin/kubernetes/context-permissions-checker.ts
```diff
@@ -37,46 +37,58 @@ export interface ContextPermissionsRequest {
...
+    this.#contextName = contextName;
     this.#request = request;
-    this.#results = new Map<string, ContextResourcePermission>();
+    this.#results = [];
     this.#client = this.#kubeconfig.getKubeConfig().makeApiClient(AuthorizationV1Api);
   }
 
+  public isForContext(contextName: string): boolean {
+    return this.#contextName === contextName;
+  }
```
benoitf: hello, I have some remarks about this method.

For example, why the context of this permission checker is not something we can get from the outside ? like `permissionChecker.contextName` (readonly attribute)

b/c the isForContext seems to be for another usecase

or accept ContextHealthState attribute


permissionChecker.matchContextHealthState(ctx): boolean

feloy: I was not sure what was the best, to use a readonly attribute or such a method. I'll change to an attribute

packages/main/src/plugin/kubernetes/context-permissions-checker.ts
```diff
@@ -58,7 +58,7 @@ export interface ContextPermissionResult extends Permission {
 }
 
 export class ContextPermissionsChecker implements Disposable {
-  #contextName: string;
+  contextName: string;
```
benoitf: this is a problem there as it's a public attribute there

I think it should stay with #contextName (private)


and then add a `public get contextName(): string {` method that will expose the context but only as a read only attribute

benoitf (approved):

Issue #10467, opened 2024-12-23
Kubernetes experimental


feloy:
### Epic domain

An experimental mode for Kubernetes has been integrated, to support the use of Health check and the checking of permissions (and cleaning up the current logic for Kubernetes contexts monitoring).

These steps need to be implemented in this experimental mode:
- #10468
- #10632
- #10385 
- #10655 
- #10656 
- #11183 
- #10657 
- #7629 
- #8394 
- #11108 
- not all contexts are monitored (current only? local only? etc): done in #11211 
- provide a button to start monitoring on any context: done in #11211
- fetch the API resources to check which CRD are present (routes, etc) and dispatch information
- provide a button to refresh monitoring (permissions or installed CRDs may have changed since last time the monitoring started)
- handle gracefully token refresh with executable declared in the kubeconfig file

### Additional context

Two objects are created from KubernetesClient:
- ContextsManagerExperimental (replace ContextsManager)
- ContextsStatesDispatcher

#### ContextsManagerExperimental 

Starts a ContextsDispatcher which receives different signals about Kubernetes contexts, below, and sends signals for starting/stopping monitoring on contexts:
- Update of the Kubeconfig file (done),
- Which contexts the user wants to monitor (todo),
- Additional contexts the user wants to start/stop monitoring, based on existing context with different namespace (todo),
- The user wants to refresh the monitoring of a context (permissions changed, etc) (todo)

For each context to start monitoring, the steps are:
- Check the health of the context’s cluster (only once for now)
- As soon as the cluster is healthy, the context is considered “reachable”
- For each Kubernetes resource provided by the resourceFactoryHandler:
  - Check if the context’s user has permission to watch the resource in the cluster
  - If the permission is granted, an informer is started to cache the list of objects of this kind locally

#### ContextsStatesDispatcher

The ContextsStatesDispatcher receives signals from the ContextsManagerExperimental for new health check results, permission check results, context deletions, when resources are updated, and sends related events for the frontend.

These events do not contain payload and the ContextsStatesDispatcher provides methods to fetch updated information (healths state, permissions state, resources counts, resources lists)

rujutashinde:
closing this epic, the remaining tasks are covered in https://github.com/podman-desktop/podman-desktop/issues/11533

Issue #11002, opened 2025-02-05
KubernetesObjectsList needs a isResource


feloy:
### Is your enhancement related to a problem? Please describe

See https://github.com/podman-desktop/podman-desktop/pull/10928#discussion_r1943237887

### Describe the solution you'd like

Change the way the resource kind is detected in the `KubernetesObjectsList` component, to remove the need for the caller to pass a `isResource`

PR #11065, opened 2025-02-10
feat: display ingresses/routes in Kubernetes experimental mode

feloy:
### What does this PR do?

Display ingresses/routes in Kubernetes experimental mode

Note that the user will get a 404 error on routes endpoint on non-openshift clusters, as an informer will be started even if the route CRD is not defined: we need to implement `fetch the API resources to check which CRD are present (routes, etc) and dispatch information` (https://github.com/podman-desktop/podman-desktop/issues/10467) to remove this error.
 
Part of #10657 

### How to test this PR?

Check in a kind cluster that ingresses are correctly displayed
Check in Sandbox Openshift that routes are correctly displayed

Also check in non-experimental mode


packages/renderer/src/lib/ingresses-routes/IngressesRoutesList.svelte
```diff
@@ -140,46 +90,35 @@ const columns = [
...
-  </div>
-</NavPage>
+
+<KubernetesObjectsList
+  kinds={[
+    {
+      resource: 'ingresses',
+      transformer: ingressRouteUtils.getIngressUI,
+      delete: window.kubernetesDeleteIngress,
+      isResource: (o: KubernetesObject): boolean => ingressRouteUtils.isIngress(o as IngressUI | RouteUI),
```
benoitf, line 100: question: there is no other kind of check before assuming doing the cast ?
I would assume that we have some fields to check on the kubernetesObject to be sure we receive the right kind of element before converting to ingress or like line 106 to a routeUI

deboer-tim: Yes, the cast is a bit ugly. We know it will be one of these types but unfortunate to have to do this. OTOH, I don't think we should need isResource at all (#11002) so I'd say it's ok for now.

packages/renderer/src/lib/ingresses-routes/IngressesRoutesList.svelte
```diff
@@ -140,46 +90,35 @@ const columns = [
...
+      legacySearchPatternStore: ingressSearchPattern,
+      legacyObjectStore: kubernetesCurrentContextIngressesFiltered,
+    },
+    {
+      resource: 'routes',
+      transformer: (o: KubernetesObject): RouteUI => ingressRouteUtils.getRouteUI(o as V1Route),
+      delete: window.kubernetesDeleteRoute,
+      isResource: (): boolean => true,
```
deboer-tim, line 108: I was expecting something like an ingressRouteUtils.isRoute() here (or maybe !isIngress() since isRoute() doesn't exist), but I guess this works b/c if it's not an ingress it must be a route and uses this delete? Having this fallback is a bit ugly too, but again I'm ok with that if we hope to remove the need for isResource entirely.

deboer-tim (approved):

packages/renderer/src/lib/ingresses-routes/IngressesRoutesList.svelte
feloy: Yes, logically, it is the same as `!isIngress`, as isResource is called in ingresses then routes. I agree that this is not maintainable, but I really would like to fix this 
with #11002, and not try to find some intermediary solution.

gastoner (approved):
LGTM on Windows: 
image

Issue #11035, opened 2025-02-07
Kubernetes UI object and column restructuring


deboer-tim:
### Is your enhancement related to a problem? Please describe

Florent and Alex had good ideas for cleaning up and simplifying the Kubernetes UI objects and columns in the review for #10906.

### Describe the solution you'd like

Implement both ideas!

KubernetesObjectsList transformers should require all objects to have name/namespace/details props. Name column can be shared.

Move each set of *Columns into a /columns folder with a shared props.ts, and simplify names:
<img width="302" alt="Image" src="https://github.com/user-attachments/assets/4e155da6-f7ff-4cfe-ac7a-fef6e53e6b83" />

deboer-tim:
All of the work to refactor, reuse, and clean up has been completed for Kubernetes, the focus of this issue. We might want to consider the same work for other areas in the future, but I'm marking this issue as closed.

PR #11094, opened 2025-02-11
chore: common kubernetes ui object

deboer-tim:
### What does this PR do?

First step of #11035 - a common Kubernetes UI object. This just adds some minimal typing which allows the name column to be shared. Namespace is optional since it doesn't exist on nodes.

The name column is used for pods as an example before I do the rest. It's unfortunate requiring a 'resource' property and still requiring PodColumnName to exist, but I haven't come up with anything simpler yet.

The rest of #11035 (other name columns, column folders/naming, props.ts) would come in future PRs.

### Screenshot / video of UI

N/A

Part of #11035.

### How to test this PR?

PR checks.


packages/renderer/src/lib/objects/KubernetesUI.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface KubernetesUI {
```
feloy: I think it would be nice to call it `KubernetesObjectUI` to make the link with the `KubernetesObject` provided by the kubernetes library, which is its Kubernetes counterpart:
```
export interface KubernetesObject {
    apiVersion?: string;
    kind?: string;
    metadata?: V1ObjectMeta;
}
```

Another point, for https://github.com/podman-desktop/podman-desktop/issues/11002, I would like to add the `kind` field to this `Kubernetes[Object]UI` interface, so the kind is propagated from the Kubernetes version to the UI version, and we can guess on the UI version of the object what is its kind.

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretUI.ts
```diff
@@ -16,7 +16,9 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-export interface ConfigMapSecretUI {
+import type { KubernetesUI } from '../objects/KubernetesUI';
+
+export interface ConfigMapSecretUI extends KubernetesUI {
   name: string;
   namespace: string;
```
feloy: I suppose you can remove name and namespace here (and for the other resources)

deboer-tim:
Ok, here's another pass at this. The object was renamed to `KubernetesObjectUI` as suggested, and due to the amount of change I just rebased at the same time.

I added a `status` property to the object because all sub-interfaces had it, and I removed `name` and `status` from each of those interfaces. I did not remove `namespace`, because it is optional on `KubernetesObjectUI` but not on other objects, so keeping it avoids type errors.

I also added a `kind` property. I would prefer to use something like `static readonly` on an abstract class, but without this you can see the impact adding a few properties. We can fully share the name column though, which is nice. I think this will work with all the other objects, except Ingress/Routes which use a unique url. We'll either need to make these urls consistent or find another approach.

packages/renderer/src/lib/objects/KubernetesUI.ts
deboer-tim: Fixed and rebased.

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretUI.ts
deboer-tim: As per the comment below I removed `name` but did not remove `namespace`, because it is optional on `KubernetesObjectUI` but is required for all sub-interfaces except `NodeUI`.

packages/renderer/src/lib/kube/pods/PodUI.ts
```diff
@@ -16,11 +16,11 @@
...
  ***********************************************************************/
 
+import type { KubernetesObjectUI } from '../../objects/KubernetesObjectUI';
 import type { PodInfoContainerUI } from '../../pod/PodInfoUI';
 
-export interface PodUI {
-  name: string;
-  status: string;
+export interface PodUI extends KubernetesObjectUI {
+  kind: 'pods';
```
feloy: If we want to double-check that the `KubernetesObject` being transformed to `KubernetesObjectUI` is of the appropriate kind (when #10916 is fixed), I think we would need to have `Pod` here, which is the value in the `kind` field of Kubernetes pods.

packages/renderer/src/lib/kube/pods/pod-utils.ts
```diff
@@ -50,6 +50,7 @@ export class PodUtils {
 
     return {
       name: pod.metadata?.name ?? '',
+      kind: 'pods',
```
feloy: Here also (`Pod`), and in the future (after #10916), we may double-check that it is the same as `pod.kind`

packages/renderer/src/lib/objects/KubernetesObjectUI.ts
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface KubernetesObjectUI {
+  name: string;
+  namespace?: string;
+  kind?: string;
```
axel7083: Should we take this opportunity to set this property to mandatory, all resources have a kind 

> All resource types have a concrete representation (their object schema) which is called a kind


From https://kubernetes.io/docs/reference/using-api/api-concepts/

packages/renderer/src/lib/kube/pods/PodUI.ts
deboer-tim: I prefer that, but would break the url scheme. Some options are:
- change all Kubernetes urls to use the official kind :)
- have a big case statement in KubernetesColumnName
- change / augment navigation api to have something like:
```
handleNavigation({
    page: NavigationPage.KUBERNETES_RESOURCE,
    parameters: {
      kind: 'pod',
      name: 'pod-name',
      namespace: 'default',
    },
  });
```

I think I like the last option regardless. Thoughts?

feloy: CRD (and native resources) define 2 names: `kind` and `plural` (see https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/custom-resource-definition-v1/#CustomResourceDefinitionSpec).

`kind` is used for serialization, while `plural` is used in urls. I would prefer to keep this logic in PD: using `pods` in URL, while having `Pod` in the kind field of the objects.

It's not clear to me why the value of this `kind` field is related to the url? (I find acceptable to pass, when necessary, these 2 values as parameter (kind and plural)

deboer-tim: > It's not clear to me why the value of this `kind` field is related to the url? (I find acceptable to pass, when necessary, these 2 values as parameter (kind and plural)

One of the hopes of the issue is being able to share the name column - which is easy with `KubernetesObjectUI` except you need to know what url to go to when you click on it. See `KubernetesColumnName` line 13.

Our current URLs don't map nicely to `kind` or `plural`, mostly due to the merged /ingressRoutes and /configmapSecrets. We could skip sharing the column or defer the logic back to each page, but I think I prefer the  option from my prior comment: add an API to `handleNavigation` to map from `kind` to Podman Desktop url. This is likely to be useful in other places too, and would allow us to use `kind` cleanly here.

feloy: ok, got it, thanks. 

If I understand correctly, you could pass into your call to `handleNavigation`, `ingressesRoutes/ingress` from an ingress, and `ingressesRoutes/route` from a route? (with some minor change to use the plural: `ingressesRoutes/ingresses` and `ingressesRoutes/routes`)

I would be more comfortable if you introduce the changes for configmapSecret (or ingressRoutes) in addition to the pods in this PR, to be sure it works for multi-resources urls.

deboer-tim: Essentially yes: do a separate PR first that would add `handleNavigation` via kind + name + namespace including any special rules, then this one can just call it.

deboer-tim:
Moving to draft since I've rebased on PR #11122 and the intended PR for issue #11124. You can ignore all navigation* changes here, but it should give a better idea how it could work and I'm still interested in feedback.

@feloy I've switched to using the correct Kubernetes `kind` values (using the navigation kind for lack of anything I could find better), and with the navigation here it all works and fits together again.

@axel once you make `kind` required there are >35 other files that require trivial changes like in /pod. I'd prefer to use this PR to confirm the pattern and support pods, then use future PRs to replace other name columns and make kind required.

deboer-tim:
Thanks for the early feedback. All the proposed changes here have been evolved, split up, and submitted under PRs #11166, #11220, and #11230, so I'm closing this. 👍🏼

PR #10490, opened 2024-12-30
chore: migrate ListItemButtonIcon to svelte 5 runes

benoitf:
### What does this PR do?

migrate ListItemButtonIcon to svelte 5 runes

### What issues does this PR fix or reference?

related to https://github.com/podman-desktop/podman-desktop/issues/10240

### How to test this PR?

- [x] Tests are covering the bug fix or the new feature

PR #11110, opened 2025-02-12
fix(ListItemButtonIcon): only use inline-flex when hidden is not set

### LINKED ISSUE #11109: container actions are showing stop button when container is exited.
### Bug description

Image

Clicking on it will create an error

Image

### Operating system

Linux, fedora 41

### Installation Method

### Version

next (development version)

### Steps to reproduce

1. Go to containers table
2. assert one container is exited
3. assert this container has the stop button

### Relevant log output

```shell

```

### Additional context

Technically the problem is not the actions

https://github.com/podman-desktop/podman-desktop/blob/bb1c3c0fbddcbabd2d5950d2939f6e920bc0d40f/packages/renderer/src/lib/container/ContainerActions.svelte#L200

Might be related to ListItemButtonIcon

axel7083:
### What does this PR do?

When both class `inline-flex` and `hidden` are set, the `inline-flex` is taking over the hidden. Leading to this property not being used.

image

I suspect the svelte5 migration ( https://github.com/podman-desktop/podman-desktop/pull/10490 ) to have some internal changes with the `class:<name>={condition}` to change the ordering and therefore behaviour.

Fixes https://github.com/podman-desktop/podman-desktop/issues/11109

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/ListItemButtonIcon.spec.ts
+++ b/packages/renderer/src/lib/ui/ListItemButtonIcon.spec.ts
@@ -155,3 +155,35 @@ test('With custom Fa icon', async () => {
   // check it is a svelte-fa class
   expect(svgElement).toHaveClass('svelte-fa');
 });
+
+test('expect button to not have inline-flex when hidden is true', async () => {
+  const title = 'title';
+
+  render(ListItemButtonIcon, {
+    title,
+    icon: faRocket,
+    hidden: true,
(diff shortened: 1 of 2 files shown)


deboer-tim (approved):
Looks more like a Taillwind change than Svelte? Either way, fix is better. LGTM

feloy (approved):
works for me

PR #11119, opened 2025-02-12
chore(extension-api): adding missing properties to Mounts property in `ContainerInspectInfo`

axel7083:
### What does this PR do?

The `Mounts` properties for the `ContainerInspectInfo` is missing the `Type` and  `Driver` property. 

We can see it is already defined in the `ContainerInfo`.

https://github.com/podman-desktop/podman-desktop/blob/f06bacce25592549d76c253063d0c2c5d243b4f3/packages/extension-api/src/extension-api.d.ts#L2552-L2561

And we can get this info from `podman container inspect <id>`.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -2837,8 +2837,10 @@ declare module '@podman-desktop/api' {
     };
     Mounts: Array<{
       Name?: string;
+      Type: string;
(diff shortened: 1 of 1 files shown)


axel7083:
Same issue like last time; dockerode do not have those properties

Issue #8819, opened 2024-09-10
Pods section should not show Kubernetes Pods by default / only show Podman and Docker


cdrage:
### Is your enhancement related to a problem? Please describe

After https://github.com/containers/podman-desktop/issues/8818 is completed, Kubernetes pods will be shown in two places.

### Describe the solution you'd like

By default, the Pods section should filter by non-kubernetes by default.

axel7083:
> By default, the Pods section should filter by non-kubernetes by default.

This mean that there should be a way to display them ? Like a configuration ?

Some ideas on top of my head

## Better search bar

![Image](https://github.com/user-attachments/assets/56bda45a-c106-4198-8e03-2c6a09c363e8)

We could make the search bar similar to the one on GitHub and allow some filtering operation on it.

![Image](https://github.com/user-attachments/assets/d7ef01f4-bca2-4f98-80f7-474865333431)

## Just not showing them

We simply get rid of the kubernetes pods

## Configuration option

We add in the preferences something like `pods.showKubernetes: boolean`, and have it disabled by default. 

Any opinion @cdrage@deboer-tim @benoitf ?

cdrage:
> > By default, the Pods section should filter by non-kubernetes by default.
> 
> This mean that there should be a way to display them ? Like a configuration ?
> 
> Some ideas on top of my head
> 
> ## Better search bar
> ![Image](https://github.com/user-attachments/assets/56bda45a-c106-4198-8e03-2c6a09c363e8)
> 
> We could make the search bar similar to the one on GitHub and allow some filtering operation on it.
> 
> ![Image](https://github.com/user-attachments/assets/d7ef01f4-bca2-4f98-80f7-474865333431)
> 
> ## Just not showing them
> We simply get rid of the kubernetes pods
> 
> ## Configuration option
> We add in the preferences something like `pods.showKubernetes: boolean`, and have it disabled by default.
> 
> Any opinion [@cdrage](https://github.com/cdrage)@deboer-tim [@benoitf](https://github.com/benoitf) ?

Hmmm, unsure if it's good UI...

My  idea is that on load for the "Pods" section, we simply auto-fill on first load the `is:provider:"podman"`

Then for the Kubernetes section load the exact same file and have it auto-fill as: `is:provider:"podman"`

I feel like this is a hidden "feature" and may showcase it a bit better by having it auto-filled?

Makes it much easier to maintain too as it's loading from the same file and no need to always cross-make-changes, etc.

[...]

PR #11122, opened 2025-02-12
feat: navigate to kubernetes pods and cron jobs

### LINKED ISSUE #11121: Navigation to Kubernetes pods and cron jobs

Navigation API has support for most Kubernetes objects, but we didn't update it when adding pods and cron jobs.

Add these two to the API.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

We're missing navigation for the new pods and cron jobs pages, this just adds them as expected.

### Screenshot / video of UI

N/A

Fixes #11121.

### How to test this PR?


axel7083 (commented):
Should we for extensive clarity rename the following

https://github.com/podman-desktop/podman-desktop/blob/60bc877775a146c2831bf5a21b3cd5d25faeb26a/packages/api/src/navigation-page.ts#L33-L33

to `PODMAN_PODS` (I am in favour of doing the rename)

deboer-tim:
> Should we for extensive clarity rename the following ...  to `PODMAN_PODS` (I am in favour of doing the rename)

I'm not sure I'm a fan and this is technically breaking, but since both of you agree I've done that. :)

axel7083:
> this is technically breaking

Why? This is internal and not exposed so only internals stuff need to change

deboer-tim:
> > this is technically breaking
> 
> Why? This is internal and not exposed so only internals stuff need to change

Oh - true. I was thinking this was extension-api, but it's just (internal) api. Anyway, commit making the change is here.

axel7083 (approved):

benoitf (commented):
I would avoid to introduce new constants for each kubernetes entries.

Kubernetes navigation sub-items should not leak into the main navigation entries.

deboer-tim:
> I would avoid to introduce new constants for each kubernetes entries.
> 
> Kubernetes navigation sub-items should not leaked into the main navigation entries.

As Axel said, nothing is in the external/extension API. There are 8 resources exposed in the internal navigation API but 2 are missing and this would add them.

Are you saying we shouldn't make these external (I agree) or not to fix the current set? I was assuming that even if we move these in the future it'd be better to have a correct set. (unless we want to undo #9688)

benoitf:
my first dummy remark would be: if there is nothing calling these methods, maybe they should not be there: Else it's a dead code as it's not externally visible. So until someone need it, we may not need this code.

While it could solve an hypothetical short term thing, it's bringing technical debt. Because it's bringing more stuff to the 'core' part of Podman Desktop, while the routing of Kubernetes sub-entries should be delegated to a command handling that routing.

So my advice is to not add any new method but use the extensibility pattern where you can add a routing scheme and handle on your own. This way, it would help any kubernetes refactoring and reduce the core part of the kubernetes dependencies.

This way it's scalable and moving Kubernetes to its own extension will be solved as it'll keep the same routing thing. (which is the external way)

deboer-tim:
The other resources were added as part of the port forwarding feature so I'm assuming at least one of them is being used? I was adding pods b/c I was working on issue #8819 / PR #11125 and thought a link would be nicer (and odd I could navigate to deployments or services but not pods).

I put #11125 up for review without a link, and if people are ok with that then I don't need this PR. I think as part of #11094 we're still going to want a way to navigate to arbitrary Kube resources by kind, but I can investigate the comments you put in #11124.

To summarize, I think I need a day to look at look at this. If we're not going to merge this PR then I think it's a sign we should open an issue to remove the others in a coming sprint, for cleanup and avoiding future PRs like this.

benoitf (approved):
bringing tech debt but it's consistent with the existing stuff

jeffmaury (approved):
LGTM

Issue #10812, opened 2025-01-23
Improving usability with two Pods pages


deboer-tim:
### UX Description

When we create a Kubernetes Pods page (#8818), there will be pods in two places in the UI. There is a separate discussion about whether we remove Kubernetes pods from the current page (#8819) or not, but _this is not that discussion_. In either case it seems likely that we'll have two pages, with two different sets of pods, and users may not initially understand the difference between the two pages, why we have two, or enter one page and wonder why some pods are 'missing'.

As discussed on the UX call this issue is to explore the options and make some recommendations:
- Should we rename either page, or use different pod icons?
- Should we include a link on one or both pages, e.g. 'Don't see your Podman pods? Click here'?
- Explore what else we can do to help the user understand the difference between the pages or the different behaviour/lifecycle between the  pods.

### Request type

None

### Primary Contact

@deboer-tim @slemeur

ekidneyrh:
Looking at this and thinking that maybe a message how suffice on the regular pod page like @deboer-tim has done on the PR https://github.com/podman-desktop/podman-desktop/pull/11125 but have it at the top of the page under the header rather than at the bottom like so:

![Image](https://github.com/user-attachments/assets/c46dc288-dedf-4fcd-b110-a5c2fb0fefd4)

And then after a release or 2 we can remove that message or put it behind an 'i' icon or a '?' icon. 

I dont think we need to design a new icon on second thought. I was thinking of ways to have the Kubernetes pages themed or styled a bit differently (like how we have the inverted theme for settings) but I think having the Kubernetes side bar is enough to indicate to the user that they are in the Kubernetes section of the application for now? We could start exploring ways to differentiate that whole section as Kubernetes but I believe that would be a problem for another ticket :D

deboer-tim:
Long term I wouldn't want to add a message there, because it makes the entire list start lower down than all other list pages so switching back and forth it is distracting - but that's also why I think it works well short term. Users will notice the message here more.

The nav page component doesn't currently allow us to add anything above the search bar though, and I would prefer to add things where we can (button bar at top right, to the right of the search bar [used for multi-select actions], or tabs area) vs changing a shared component given this should be temporary. Would below the search be ok? (you can also try out #11125)

<img width="688" alt="Image" src="https://github.com/user-attachments/assets/75010ef5-a444-457a-bedf-53f9e7b6f5a1" />

[...]

PR #11125, opened 2025-02-12
chore: remove kubernetes pods from pods page

deboer-tim:
### What does this PR do?

First part of #8819. Removes the data for Kubernetes pods from the Pods page, and
adds a simple message to the top for anyone trying to figure out where their Kube
pods are.

This is the core initial/minimal breaking change for the user, without any of the
internal or UI cleanup.

### Screenshot / video of UI

Screenshot 2025-02-24 at 10 49 56 AM

First part of #8819.

### How to test this PR?

Create pods on Podman and a cluster. Confirm that the Pods page only shows the former and includes the new message.

### DIFF
--- a/packages/renderer/src/lib/pod/PodsList.svelte
+++ b/packages/renderer/src/lib/pod/PodsList.svelte
@@ -250,7 +250,7 @@ const row = new TableRow<PodInfoUI>({ selectable: (_pod): boolean => true });
       defaultSortColumn="Name"
       on:update={(): PodInfoUI[] => (pods = pods)}>
     </Table>
-    
+
     {#if $filtered.length === 0 && providerConnections.length === 0}
       <NoContainerEngineEmptyScreen />
     {:else if $filtered.length === 0}
--- a/packages/renderer/src/lib/pod/PodsList.svelte
+++ b/packages/renderer/src/lib/pod/PodsList.svelte
@@ -3,6 +3,7 @@ import { faTrash } from '@fortawesome/free-solid-svg-icons';
 import {
   Button,
   FilteredEmptyScreen,
+  Link,
   NavPage,
   Table,
   TableColumn,
@@ -116,6 +117,10 @@ async function deleteSelectedPods(): Promise<void> {
   bulkDeleteInProgress = false;
 }
 
+async function openKubePods(): Promise<void> {
+  await window.navigateToRoute('kubernetes', { kind: 'Pod' });
+}
+
 let selectedItemsNumber: number = $state(0);
 let table: Table;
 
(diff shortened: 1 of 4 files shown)


gastoner (approved):
LGTM, tested on windows.
Question: 
1. do we want to keep the text even if is the list full of pods e.g. 
image
it is visible but could be better.
2. What about adding a link to the message to allow to jump quickly to kubernetes->pods?

feloy (approved):

benoitf (changes requested):
I would wait for UX design on how to redirect to kubernetes pods page before trying to provide a solution.

- The text is overlapping on the list of pods.
- lazy people will expect a link
- should we display this link if there is no kubernetes pod ?

benoitf (commented):
also I would just unmerge the k8s pods first before doing any cleanup at the same time

grabAllPods -> do not ask kubernetesListPods

there are just too many steps trying to be achieved at the same time

ekidneyrh:
I like the look of it at the bottom but practically I think it will have to go at the top like where we have descriptions like on the kubernetes dashboard & have the page linked like Florent mentioned. I haven't looked at the ticket in detail yet so this is for the context given above :)

deboer-tim:
Force pushed, replacing the commit to make the following changes:
- Updated the warning to the design proposed in #10812 - except below the search bar due to current limitations of the NavPage.
- Rebased on main to pick up ability to link to the Kubernetes > Pods page.
- Removed all changes to actions/tests except the bare minimum, to address @benoitf's comments.

benoitf (approved):

packages/renderer/src/lib/pod/PodsList.svelte
```diff
@@ -239,7 +250,7 @@ const row = new TableRow<PodInfoUI>({ selectable: (_pod): boolean => true });
       defaultSortColumn="Name"
       on:update={(): PodInfoUI[] => (pods = pods)}>
     </Table>
-
+
```
benoitf: nit: remove

packages/renderer/src/lib/pod/PodsList.svelte
```diff
@@ -193,39 +198,45 @@ const row = new TableRow<PodInfoUI>({ selectable: (_pod): boolean => true });
...
-          .trim()
-          .split(' ')
-          .filter(term => term !== 'is:stopped' && term !== 'is:running')
-          .join(' ')
-          .trim();
-        searchTerm = temp ? `${temp} is:stopped` : 'is:stopped';
-      }}
-      selected={searchTerm.includes('is:stopped')}>Stopped</Button>
+    <div class="flex flex-col gap-3">
+      <div class="self-center text-[var(--pd-table-body-text)]">Looking for pods running on a Kubernetes cluster? We have moved them to the <Link on:click={openKubePods}>Kubernetes &gt; Pods</Link> page.</div>
```
benoitf, line 202: question: should it be in italic. ?

deboer-tim: @ekidneyrh thoughts?

benoitf: (it's non blocking on my side)

packages/renderer/src/lib/pod/PodsList.svelte
deboer-tim: Not sure where it came from - removed.

deboer-tim:
Enabling auto-merge. Two of the three approvals were from before the last change, but given the two things that changed were reducing scope (removed some files) and updating UI to match current design I am assuming they still hold.

At this point it's better to merge and get wider feedback / unblock other changes and if there's feedback or design is updated I'll do another PR.

PR #11156, opened 2025-02-14
fix: handle ApiException error during waitForPod/JobDeletion

### LINKED ISSUE #10841: Restarting of the pod on kubernetes cluster makes it disappear
### Bug description

When I have a pod deployed to a kubernetes cluster (kind), ie. simple nginx and I want to restart a pod, it disapper from podman-desktop and the namespace is gone from cluster. Maybe this is correct behavior, but it is quite confusing.

https://github.com/user-attachments/assets/1fea802b-5448-4883-98d2-7318abdce0b3

### Operating system

Linux, Windows 10

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. Create a kind cluster via PD
2. deploy a container to a cluster
3. Assert: pod is available and running under `default` namespace - use kubectl
4. Restart pod
5. Assert: Pod is restarted
Actual: pod and namespace is gone

### Relevant log output

```shell
# relevant part of log
kubectl get events -A
default              66s         Normal    Pulling                             pod/static-web                                 Pulling image "nginx"
default              65s         Normal    Pulled                              pod/static-web                                 
Successfully pulled image "nginx" in 1.077s (1.077s including waiting). Image size: 195858651 bytes.
default              65s         Normal    Created                             pod/static-web                                 Created container: web
default              65s         Normal    Started                             pod/static-web                                 Started container web
default              53s         Normal    Killing                             pod/static-web                                 Stopping container web
```

### Additional context

_No response_

feloy:

### What does this PR do?

Fix `waitForPodDeletion` and `waitForJobDeletion` to handle error as `ApiException`

The format of the error returned by the kubernetes-client library changed when updated to v1.x.

Fixes #10841 

### How to test this PR?

- Start a standalone pod, then restart it. It should be restarted correctly
- Create a job, then restart a pod created by this job. The job should be re-created correctly

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
@@ -283,12 +283,11 @@ vi.mock('/@/plugin/kubernetes/kubernetes-port-forward-service', async () => {
 
 const execMock = vi.fn();
 beforeAll(() => {
-  vi.mock('@kubernetes/client-node', async () => {
-    // eslint-disable-next-line @typescript-eslint/consistent-type-imports
-    const actual = await vi.importActual<typeof import('@kubernetes/client-node')>('@kubernetes/client-node');
+  vi.mock('@kubernetes/client-node', async importOriginal => {
+    const original = await importOriginal<typeof clientNode>();
     return {
       // we need to use original ApiException
-      ...actual,
+      ...original,
       KubeConfig: vi.fn(),
       CoreV1Api: {},
       AppsV1Api: {},
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
@@ -284,7 +284,11 @@ vi.mock('/@/plugin/kubernetes/kubernetes-port-forward-service', async () => {
 const execMock = vi.fn();
 beforeAll(() => {
   vi.mock('@kubernetes/client-node', async () => {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+    const actual = await vi.importActual<typeof import('@kubernetes/client-node')>('@kubernetes/client-node');
     return {
+      // we need to use original ApiException
+      ...actual,
       KubeConfig: vi.fn(),
       CoreV1Api: {},
       AppsV1Api: {},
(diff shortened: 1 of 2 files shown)


benoitf (approved):

packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
```diff
@@ -284,7 +284,11 @@ vi.mock('/@/plugin/kubernetes/kubernetes-port-forward-service', async () => {
 const execMock = vi.fn();
 beforeAll(() => {
   vi.mock('@kubernetes/client-node', async () => {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+    const actual = await vi.importActual<typeof import('@kubernetes/client-node')>('@kubernetes/client-node');
```
axel7083: In the documentation they have the following (https://vitest.dev/api/vi#vi-mock)

```ts
vi.mock(import('./path/to/module.js'), async (importOriginal) => {
  const mod = await importOriginal() // type is inferred
  return {
    ...mod,
    // replace some exports
    total: vi.fn(),
  }
})
```

Can this be used instead of using typeof import ? In the vitest documentation it is explicit `type is inferred` with the method above, so we would avoid the typeof .... and the eslint disable (if it works?)

feloy: Yes, that seems to be much better. Changing, and adding the rule to be added to the code guideline (https://github.com/podman-desktop/podman-desktop/issues/9821#issuecomment-2659493666)

feloy: finally, I cannot make it work...
In this code:
```
  const mod = await importOriginal() // type is inferred
```

mod is of type `unknown`, and we get the error `Spread types may only be created from object types`

feloy: but this works correctly, without disabling anything:
```
import * as clientNode from '@kubernetes/client-node';

vi.mock('@kubernetes/client-node', async (importOriginal) => {
    const original = await importOriginal<typeof clientNode>();
```

feloy: I have mis-read the piece of code and missed the `import(...)` in the call to `vi.mock`.

Effectively, using this makes the type inferred, but we get many errors that `vi.fn()` are not matching the expected types.

That's an interesting improvement, but I don't think I will be able to use it for this PR

axel7083 (approved):

PR #11159, opened 2025-02-14
fix(run-image-page): checkbox interactive name used

axel7083:
### What does this PR do?

When using the following code in a playwright test

```ts
const containers = await runImage.startContainer('potatoes', {
  interactive: false,
});
```

The test timeout, and the root cause were the name used for the checkbox locator of the interactive option. The name is the `title`of the checkbox component here

https://github.com/podman-desktop/podman-desktop/blob/bb1c3c0fbddcbabd2d5950d2939f6e920bc0d40f/packages/renderer/src/lib/image/RunImage.svelte#L828

Which is `Use interactive` not `Interactive: Keep STDIN`.

This was not an issue before, as we don't seems to have any tests starting a container and disabling the interactive option.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Not sure if I should be writing a test 

cc @podman-desktop/qe-reviewers

### DIFF
--- a/tests/playwright/src/model/pages/run-image-page.ts
+++ b/tests/playwright/src/model/pages/run-image-page.ts
@@ -121,7 +121,7 @@ export class RunImagePage extends BasePage {
         // disable the checkbox in advanced tab
         await this.activateTab('Advanced');
         const checkbox = this.page.getByRole('checkbox', {
-          name: 'Interactive: Keep STDIN',
+          name: 'Use interactive',
         });
         if (optionalParams.interactive) {
           await checkbox.check();


cbr7 (approved):

deboer-tim (approved):

tests/playwright/src/model/pages/run-image-page.ts
```diff
@@ -121,7 +121,7 @@ export class RunImagePage extends BasePage {
         // disable the checkbox in advanced tab
         await this.activateTab('Advanced');
         const checkbox = this.page.getByRole('checkbox', {
-          name: 'Interactive: Keep STDIN',
+          name: 'Use interactive',
```
benoitf, line 124: does it mean we should check the `.toBeVisible()` on this item ?

on vitest getByRole enforce the element exists on the page (and queryByRole doesn't)
is that playwright getByRole is equivalent to the queryByRole ?

axel7083: With playwright, using `getByRole` will return a [Locator](https://playwright.dev/docs/api/class-locator) (not the actual object), it will not throw an error if the Locator does not point to anything;

If you want all buttons for example you do `await page.getByRole('button').all()`, so it is not an equivalent of `queryByRole`, it is like another function with a different behavior :(

benoitf: so probably we're missing an assertion checking that the checkbox exists (even we don't change its value / check the box) so tomorrow if we change the label the test would fail

benoitf: but up to @podman-desktop/qe-reviewers

cbr7: I don't really see what the issue is to be honest, getByRole does not ensure the web element actually exists, nor should it, but when you try to use the element, like calling check() method, there is an implicit status check on the element that PW does in the background, these are documented [here](https://playwright.dev/docs/actionability)

PR #11162, opened 2025-02-14
feat: add apiVersion and kind to kubernetes objects

### LINKED ISSUE #10916: Resources obtained by informers do not contain kind/apiVersion

feloy:

### What does this PR do?

Add apiVersion and kind to kubernetes objects, by passing to the informer constructor a listFn adding the kind and apiVersion, instead of the native listFn provided by Kubernetes.

In the message https://github.com/kubernetes/kubernetes/pull/127361#issuecomment-2353199198, it is explained that only the list operation does not return the kind/apiVersion, other operations do return them.

It is said in the above message that the client is removing these information, but this is not the case with the Javascript client. 

We can then expect that kind/apiVersion will be present in all resources.

Note that the listFn is only called when an informer is started, to get the original list of resources. listFn should not be called after this initial call.

Fixes #10916 

### How to test this PR?

- Add some log into the definition of `typedList` to check that this function is called only once (per resource), after the beginning.

- Add some log to display resources received in the frontend (in KubernetesObjectsList for example)

- Start Podman Desktop with existing resources, then add, update, delete resources. All resources displayed by the frontend should contain kind/apiVersion

### DIFF
--- a/packages/main/src/plugin/kubernetes/configmaps-resource-factory.ts
+++ b/packages/main/src/plugin/kubernetes/configmaps-resource-factory.ts
@@ -54,6 +54,6 @@ export class ConfigmapsResourceFactory extends ResourceFactoryBase implements Re
     const apiClient = kubeconfig.getKubeConfig().makeApiClient(CoreV1Api);
     const listFn = (): Promise<V1ConfigMapList> => apiClient.listNamespacedConfigMap({ namespace });
     const path = `/api/v1/namespaces/${namespace}/configmaps`;
-    return new ResourceInformer<V1ConfigMap>(kubeconfig, path, listFn, 'configmaps');
+    return new ResourceInformer<V1ConfigMap>(kubeconfig, path, listFn, { kind: 'ConfigMap', plural: 'configmaps' });
   }
 }
--- a/packages/main/src/plugin/kubernetes/configmaps-resource-factory.ts
+++ b/packages/main/src/plugin/kubernetes/configmaps-resource-factory.ts
@@ -54,6 +54,6 @@ export class ConfigmapsResourceFactory extends ResourceFactoryBase implements Re
     const apiClient = kubeconfig.getKubeConfig().makeApiClient(CoreV1Api);
     const listFn = (): Promise<V1ConfigMapList> => apiClient.listNamespacedConfigMap({ namespace });
     const path = `/api/v1/namespaces/${namespace}/configmaps`;
-    return new ResourceInformer<V1ConfigMap>(kubeconfig, path, listFn, { kind: 'ConfigMap', plural: 'configmaps' });
+    return new ResourceInformer<V1ConfigMap>({ kubeconfig, path, listFn, kind: 'ConfigMap', plural: 'configmaps' });
   }
 }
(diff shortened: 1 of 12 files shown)


packages/main/src/plugin/kubernetes/configmaps-resource-factory.ts
```diff
@@ -54,6 +54,6 @@ export class ConfigmapsResourceFactory extends ResourceFactoryBase implements Re
     const apiClient = kubeconfig.getKubeConfig().makeApiClient(CoreV1Api);
     const listFn = (): Promise<V1ConfigMapList> => apiClient.listNamespacedConfigMap({ namespace });
     const path = `/api/v1/namespaces/${namespace}/configmaps`;
-    return new ResourceInformer<V1ConfigMap>(kubeconfig, path, listFn, 'configmaps');
+    return new ResourceInformer<V1ConfigMap>(kubeconfig, path, listFn, { kind: 'ConfigMap', plural: 'configmaps' });
```
benoitf: issue: I think we reached the Point of No Return in the parameter list

so it's easy to add/remove stuff and to identify the parameters and the order is not mandatory
new ResourceInformer<V1ConfigMap>({kubeConfig, path, listFn, kind, plural})

benoitf (commented):
too many params --> need to switch to an object parameter

packages/main/src/plugin/kubernetes/configmaps-resource-factory.ts
feloy: Sure, I was wondering what was the level of the Point of No Return 😅
I'll switch to an object

benoitf: I think you won 🎉 🎁

packages/main/src/plugin/kubernetes/resource-informer.ts
```diff
@@ -38,11 +44,20 @@ export interface OfflineEvent extends BaseEvent {
...
+  plural: string;
+}
+
 export class ResourceInformer<T extends KubernetesObject> implements Disposable {
   #kubeConfig: KubeConfigSingleContext;
   #path: string;
   #listFn: ListPromise<T>;
-  #resourceName: string;
+  #pluralName: string;
+  #kindName: string;
```
axel7083, line 64: Could you add a short comment on each of them to precise their prupose, I never know which one is what :crab:

axel7083 (approved):
LGTM, just one comment on the documentation, to add a bit of clarity

benoitf (approved):

PR #11166, opened 2025-02-14
feat: kubernetes resource navigation

### LINKED ISSUE #11124: Navigation API for Kubernetes resources by kind

From discussion here: https://github.com/podman-desktop/podman-desktop/pull/11094#discussion_r1952815779

The navigation API has lots of ways to open the right page for a resource, e.g. `NavigationPage.KUBERNETES_SERVICES` to list services or `NavigationPage.KUBERNETES_DEPLOYMENT` to get to a specific deployment, but all of these are single path. There's no way to generically open the page for a resource of a given kind.

I also noticed that where we have a combined page like ingresses and routes, there is a single KUBERNETES_INGRESSES_ROUTES. This matches our current UI, but from an API standpoint it's not consistent and the client just wants to say 'show the routes' and if the page/url is shared that is an implementation detail. It also restricts us a little as we couldn't decide to split ingresses and routes into different pages without 'breaking'.

Provide a way to navigate to any supported Kubernetes resource by passing the kind to `NavigationPage.KUBERNETES_RESOURCES` and `NavigationPage.KUBERNETES_RESOURCE`.

Split ingress/route and configmap/secret to make them consistent with other resources.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds a navigation route to any Kubernetes resource.

- Adds a new 'kubernetes-navigation' command that accepts a KubernetesNavigationRequest with kind, name, and namespace, along with a 'kubernetes' navigation route.
- The command fires an event to the renderer, and a listener in App.svelte forwards to kubernetesNavigation to route to the correct URL.
- Exposes the navigation manager navigateToRoute() to the renderer. Used in the deployment name column to show it in use.

The only interesting bit in kubernetesNavigation are the special cases where our URLs don't match the normal pattern. We could improve this, but that's a separate issue.

Future PRs would convert other name columns, and we would discuss if we should remove Kubernetes support from navigation.ts.

### Screenshot / video of UI

N/A

Fixes #11124.

### How to test this PR?

Go to Deployments, clicking on a name should still bring you to the details.

### DIFF
--- a/packages/api/src/kubernetes-navigation.ts
+++ b/packages/api/src/kubernetes-navigation.ts
@@ -1,0 +1,23 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export type KubernetesNavigationRequest = {
+  kind: string;
+  name?: string;
+  namespace?: string;
+};
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -674,6 +674,12 @@ export class PluginSystem {
       onboardingRegistry,
     );
 
+    commandRegistry.registerCommand('kubernetes-navigation', args => {
+      apiSender.send('kubernetes-navigation', args);
+    });
+
+    navigationManager.registerRoute({ routeId: 'kubernetes', commandId: 'kubernetes-navigation' });
+
     const extensionAnalyzer = new ExtensionAnalyzer();
 
     const extensionWatcher = new ExtensionWatcher(fileSystemMonitoring);
@@ -2837,6 +2843,13 @@ export class PluginSystem {
       }
       window.close();
     });
+
+    this.ipcHandle(
+      'navigation:navigateToRoute',
+      async (_listener, routeId: string, ...args: unknown[]): Promise<void> => {
+        return navigationManager.navigateToRoute(routeId, ...args);
+      },
+    );
 
     this.ipcHandle('onboardingRegistry:listOnboarding', async (): Promise<OnboardingInfo[]> => {
       return onboardingRegistry.listOnboarding();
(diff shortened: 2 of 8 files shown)


packages/main/src/plugin/index.ts
```diff
@@ -674,6 +674,12 @@ export class PluginSystem {
       onboardingRegistry,
     );
 
+    commandRegistry.registerCommand('kubernetes-navigation', args => {
```
benoitf: question: what is the type of args ?

deboer-tim: I added the type (KubernetesNavigationRequest).

packages/renderer/src/kubernetesNavigation.ts
```diff
@@ -0,0 +1,62 @@
...
+
+function resourceKindToURL(kind: string): string {
+  // handle the special cases in our urls
+  if (kind === 'Ingress' || kind === 'Route') {
+    return 'ingressesRoutes';
+  } else if (kind === 'ConfigMap' || kind === 'Secret') {
+    return 'configmapsSecrets';
+  }
+  // otherwise do the simple conversion
+  return kind.toLowerCase() + 's';
```
feloy, line 61: We will need to take care of this conversion, as it is not universal (see ingress -> ingresses)

packages/api/src/kubernetes-navigation.ts
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export type KubernetesNavigationRequest = {
+  kind: string;
```
feloy, line 20: Don't we want to use `resource` (or `plural`) here, as we are managing URLs, so we don't need to do the singular->plural conversion later.

feloy: I feel that we are facing the same problem as in the Kubernetes API, that we will need sometimes to convert between kind and resource (= plural). In Kubernetes API, this is the GVK <=> GVR conversion problem. In the client-go library (at least), the problem is resolved by using a *scheme*, in which all resources are registered:
- https://github.com/kubernetes/client-go/blob/master/kubernetes/scheme/register.go
- https://github.com/kubernetes/api/blob/master/apps/v1/register.go

This makes possible to convert between go type/Kind/resource. 

I think that the `ResourceFactoryHandler` introduced in the Kubernetes experimental mode will help manage this problem, as it is registering all the resources (with their resource name, we would need to add its kind to make the conversion possible)

packages/renderer/src/kubernetesNavigation.ts
deboer-tim: Yes. We don't follow the convention for ingress anyway (due to ingress/route on same page), but as per your other comment I agree a way to lookup the plural form would be better.

packages/api/src/kubernetes-navigation.ts
deboer-tim: +1 to having the kind/plural etc mapping somewhere, and ResourceFactoryHandler seems like the logical place. We also never type `kind`, I've been waffling about whether we should, even just so that there's a list of the 'supported' types.

Either way, our urls don't match the plural forms for several cases so we are likely to always need some logic here. Today, navigation.ts just has a huge mapping for each url. IMHO kubernetesNavigation.ts at least brings that down to 2 functions with some exceptions, which we could improve once we have mapping or can simplify the exceptions.

axel7083 (approved):
Following our discussion I approve this PR with the requirement of having a follow-up PR to simplify the navigation.

The ingress and routes can have the same fallback, so we can have the same `/kubernetes/routes/*` and `/kubernetes/ingress` to fallback. Goal being to use the kind/plural in the navigation (never know which one to use)

feloy (approved):

deboer-tim:
Thanks. Follow up in #11219.

PR #11184, opened 2025-02-17
chore: fix cve on esbuild

benoitf:
### What does this PR do?
fix cve by updating esbuild component used by vite

https://github.com/advisories/GHSA-67mh-4wv8-2f99

### How to test this PR?

### DIFF
--- a/package.json
+++ b/package.json
@@ -224,7 +224,8 @@
       "@kubernetes/client-node>jsonpath-plus": "^10.0.7",
       "cssstyle>@asamuzakjp/css-color": "^2.8.3-b.2",
       "vitest>vite": "^6.1.0",
-      "tsx>esbuild": "^0.25.0"
+      "tsx>esbuild": "^0.25.0",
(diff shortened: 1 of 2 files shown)


axel7083 (approved):

package.json
```diff
@@ -224,7 +224,8 @@
       "@kubernetes/client-node>jsonpath-plus": "^10.0.7",
       "cssstyle>@asamuzakjp/css-color": "^2.8.3-b.2",
       "vitest>vite": "^6.1.0",
-      "tsx>esbuild": "^0.25.0"
+      "tsx>esbuild": "^0.25.0",
+      "vite>esbuild": "^0.25.0"
```
axel7083, line 228: what happen to this when vite upgrade the dependencies ?

benoitf: it's an higher range so there is no big issue
(also I'm looking at dependencies to check if we can remove the overrides)

axel7083: Okay, thanks

PR #11036, opened 2025-02-07
feat(status-bar): pin-unpin providers

axel7083:
### What does this PR do?

So after following discussion with many many people over a pretty long period of time. This is an iteration of pinning unpinning.

I am not writing any tests because I am waiting for feedback before going further.

cc @deboer-tim @slemeur @Firewall 

#### Some notes

- Users can pin/unpin **PROVIDERS**
- Only the Podman provider is pinned by default.
- At the start of the statusbar there is a pin icon
- User pin are persistent over restart of Podman Desktop

#### Technical limitation

Status Bar Providers (added in https://github.com/podman-desktop/podman-desktop/pull/9724) are not status bar entries. So this is two distinct elements.

With the current code, as noted by Tim we cannot unpin status bar entries (extension registered entries, version, help etc.) this is because we are treating the providers widget as status bar entries

#### Existing issues

The height of the status bar is changing, this is because the Provider Widget component is using too much space: need to be address in another PR. 

### Screenshot / video of UI

https://github.com/user-attachments/assets/737936b7-0f13-453a-975f-237579a4b540

### What issues does this PR fix or reference?

Fixes https://github.com/podman-desktop/podman-desktop/issues/9950 
Also fixes https://github.com/podman-desktop/podman-desktop/issues/10920

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature (TODO)

PR #11194, opened 2025-02-17
chore: update ProviderWidget tooltips

### LINKED ISSUE #10921: Combine and enhance tooltips for provider statusbar

Great job on the statusbar work so far. After I tried it I think we can significantly improve the tooltips that show up on the statusbar. Currently there is 2 tooltips, one for the the status and one for the name of the provider. As a user I found that pretty confusing.

Merge the tooltips into 1. 

Image

SoniaSandler:

### What does this PR do?
This PR updated the tooltips of the ProviderWidget to be one tooltip with the icon, status, and provider names

### Screenshot / video of UI
[Screencast from 2025-02-18 13-33-54.webm](https://github.com/user-attachments/assets/62b32e9c-98d4-4882-a4cd-dc1ddb6cb4c6)

Fixes https://github.com/podman-desktop/podman-desktop/issues/10921

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/statusbar/ProviderWidget.svelte
+++ b/packages/renderer/src/lib/statusbar/ProviderWidget.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { Button } from '@podman-desktop/ui-svelte';
+import { Button, Tooltip } from '@podman-desktop/ui-svelte';
 import { router } from 'tinro';
 
 import type { ProviderInfo } from '/@api/provider-info';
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/statusbar/ProviderWidgetStatus.ts
```diff
@@ -0,0 +1,48 @@
...
+    case 'error':
+      return 'Error';
+    case 'starting':
+      return 'Starting';
+    case 'stopping':
+      return 'Stopping';
+    case 'stopped':
+      return 'Off';
+    case 'unknown':
+      return 'Unknown';
```
axel7083: Should be ignore in favour of the default; case

packages/renderer/src/lib/statusbar/ProviderWidgetStatus.ts
```diff
@@ -0,0 +1,48 @@
...
+ ***********************************************************************/
+
+import type { ProviderConnectionStatus, ProviderStatus } from '@podman-desktop/api';
+
+export function getStatusName(status: ProviderStatus | ProviderConnectionStatus): string {
+  switch (status) {
+    case 'ready':
+      return 'Running';
+    case 'started':
+      return 'Running';
```
axel7083:
```suggestion
    case 'ready':
    case 'started':
      return 'Running';
```

packages/renderer/src/lib/statusbar/ProviderWidget.svelte
```diff
@@ -14,31 +16,63 @@ interface Props {
...
-let tooltipText = $derived.by(() => {
-  let tooltip = '';
+let connectionsStatuses = $derived.by(() => {
+  let statuses: ProviderStatus | ProviderConnectionStatus[];
+  let connectionsStatuses: { status: ProviderStatus | ProviderConnectionStatus; connecions: string }[] = [];
   if (entry.containerConnections.length > 0) {
-    tooltip = entry.containerConnections.map(c => c.name).join(', ');
+    statuses = [...new Set(entry.containerConnections.map(connection => connection.status))];
+    let connections = '';
+    for (const status of statuses) {
```
axel7083: Set is iterable [Set#Symbol.iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Symbol.iterator)

So you can do the following

````ts
const statuses: Set<string> = new Set(entry.containerConnections.map(connection => connection.status)))
for(const status of statuses) {
	....
}
````

axel7083: image

I was curious, and I used `https://jsbench.me/` to compare if spreading the set was faster/slower than using the build in iterator, result: we should use the internal iterator!

axel7083 (commented):
This PR will conflict with https://github.com/podman-desktop/podman-desktop/pull/11036

What I would suggest:

- A props should exist to disable the tooltip

deboer-tim:
I find the grouping by status and changing/different format (single vs multiline) a bit odd. If I have two or more connections, the most common use for the tooltip is going to be quickly checking the status of connection X, but with this implementation the position of X could be different each time I open the tooltip, and depends on the status of Y or Z. I need to scan to find X in one of the lists, then visually backtrack to see what state it is in.

If I have 3 connections, I'd rather the tooltip was always 3 lines high, and each line consistently showed the status of one connection.

axel7083:
> If I have 3 connections, I'd rather the tooltip was always 3 lines high, and each line consistently showed the status of one connection.

I agree, but this is not the design chosen in https://github.com/podman-desktop/podman-desktop/issues/10921 (discussion should be moved their then)

deboer-tim:
> > If I have 3 connections, I'd rather the tooltip was always 3 lines high, and each line consistently showed the status of one connection.
> 
> I agree, but this is not the design chosen in #10921

I noticed there was a design there after I posted - as far as I can tell it doesn't show any grouping, but doesn't say not to either (i.e. doesn't cover this). :(

> (discussion should be moved their then)

Agreed, I already added a comment there.

Firewall:
Yes, it was not clear from my description but I would also have 1 line for each provider. 

Sorting I would do it the same as on the provider page.

SoniaSandler:
@Firewall @deboer-tim @axel7083 I've updated the PR based on the reviews and requests, separated the providers in the tooltip to be one per line, and added a prop to hide the tooltip

jeffmaury (approved):
LGTM

axel7083 (approved):
image

LGTM

Firewall (approved):
Works great now! Thanks

deboer-tim:
Thanks for the changes @SoniaSandler !

Issue #7593, opened 2024-06-11
Do not show managedField for Kubernetes for Kube or Inspect.


cdrage:
### Is your enhancement related to a problem? Please describe

On our GUI, we show managed fields:

![Screenshot 2024-06-11 at 5 44 58 PM](https://github.com/containers/podman-desktop/assets/6422176/98002786-7d44-4ca7-a828-70d495e3215b)

vs CLI (via `kubectl edit`):

![Screenshot 2024-06-11 at 5 45 20 PM](https://github.com/containers/podman-desktop/assets/6422176/37453484-820b-420d-b6a4-8c77fbf397cf)

### Describe the solution you'd like

Managedfields provides no information for us to edit and is meant to be internal use only for Kubernetes.

It should not show when editing.

PR #11196, opened 2025-02-17
feat: display service details on Kubernetes experimental mode

### LINKED ISSUE #11183: feat: collect kube events in experimental mode

Informers for Events resources are started in non experimental mode.

We need to start same informers in experimental mode

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

In Kubernetes experimental mode:

- registers the Event resource
- create a `listenResource` helper for Details pages
- use `listenResource` for Services details

Fixes #11183 
Part of #10657 

### How to test this PR?

- create a service 
- go to Kubernetes > Services > Details
- check that the summary of service is displayed
- check that the service can be updated from the Kube tab
- create and delete events for the service(1) and check they are displayed reactively in the Summary page

(1) you can create events manually with:
```
apiVersion: v1
count: 1
involvedObject:
  apiVersion: v1
  kind: Service
  name: svc1                                <-- replace with name of service
  namespace: default                        <-- replace with namespace of service
  uid: e9cef2cc-2fb8-409c-9c92-ff4f57e247fb <-- replace with uid of service
kind: Event
lastTimestamp: "2025-02-17T15:05:49Z"
message: test message
metadata:
  creationTimestamp: "2025-02-17T08:22:46Z"
  name: svc1-event1
  namespace: default                        <-- replace with namespace of service
reason: BackOff
reportingComponent: kubelet
reportingInstance: kind-cluster-control-plane
source:
  component: kubelet
  host: kind-cluster-control-plane
type: Normal
```

### DIFF
--- a/packages/renderer/src/lib/kube/resource-listen.spec.ts
+++ b/packages/renderer/src/lib/kube/resource-listen.spec.ts
@@ -1,0 +1,341 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { CoreV1Event, KubernetesObject } from '@kubernetes/client-node';
+import { type Writable, writable } from 'svelte/store';
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+
+import type { IDisposable } from '../../../../main/src/plugin/types/disposable';
+import { listenResource } from './resource-listen';
+import * as resourcesListen from './resources-listen';
+
+vi.mock(import('./resources-listen'), async importOriginal => {
+  const original = await importOriginal();
+  return {
+    // we want to access the original isKubernetesExperimentalMode
+    ...original,
+    listenResources: vi.fn(),
+  };
+});
+
+type initListsReturnType = {
+  updateResources: (objects: KubernetesObject[]) => void;
+  updateEvents: (objects: CoreV1Event[]) => void;
+  legacyResourceStore?: Writable<KubernetesObject[]>;
+  legacyEventsStore?: Writable<CoreV1Event[]>;
+};
+
+describe.each<{
+  experimental: boolean;
+  initLists: (resources: KubernetesObject[], events: CoreV1Event[]) => initListsReturnType;
+}>([
+  {
+    experimental: true,
+    initLists: (resources: KubernetesObject[], events: CoreV1Event[]): initListsReturnType => {
+      let resourcesCallback: (resources: KubernetesObject[]) => void;
+      let eventsCallback: (events: CoreV1Event[]) => void;
+      vi.mocked(resourcesListen.listenResources).mockImplementation(
+        async (resourceName, _options, cb): Promise<IDisposable> => {
+          if (resourceName === 'events') {
+            eventsCallback = cb;
+            setTimeout(() => eventsCallback(events));
+            return {
+              dispose: (): void => {},
+            };
+          } else {
+            resourcesCallback = cb;
+            setTimeout(() => resourcesCallback(resources));
+            return {
+              dispose: (): void => {},
+            };
+          }
+        },
+      );
+      return {
+        updateResources: (updatedObjects: KubernetesObject[]): void => {
+          resourcesCallback(updatedObjects);
+        },
+        updateEvents: (updatedObjects: CoreV1Event[]): void => {
+          eventsCallback(updatedObjects);
+        },
+      };
+    },
+  },
+  {
+    experimental: false,
+    initLists: (services: KubernetesObject[], events: CoreV1Event[]): initListsReturnType => {
+      const legacyResourceStore = writable<KubernetesObject[]>(services);
+      const legacyEventsStore = writable<CoreV1Event[]>(events);
+      return {
+        legacyResourceStore,
+        legacyEventsStore,
+        updateResources: (resources: KubernetesObject[]): void => {
+          legacyResourceStore.set(resources);
+        },
+        updateEvents: (events: CoreV1Event[]): void => {
+          legacyEventsStore.set(events);
+        },
+      };
+    },
+  },
+])('experimental is $experimental', ({ experimental, initLists }) => {
+  describe.each<{ listenEvents: boolean }>([
+    {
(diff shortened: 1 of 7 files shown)


packages/renderer/src/lib/kube/resource-listen.ts
```diff
@@ -0,0 +1,132 @@
...
+      eventsListener?.dispose();
+      resourcesUnsubscriber?.();
+      eventsUnsubscriber?.();
+    },
+  };
+}
+
+// removes "verbose" fields from a Kubernetes resource description:
+// - metadata.managedFields
+function nonVerbose(resource: KubernetesObject): KubernetesObject {
```
axel7083, line 121: question: why do we delete the managed fields ?

remark: maybe this filtering is something that should have been done on the `main` side ? To avoid the unnecessary payload size with the data to finally delete it.

packages/renderer/src/lib/kube/resource-listen.ts
```diff
@@ -0,0 +1,132 @@
...
+  onEventsUpdated?: (events: CoreV1Event[]) => void;
+}
+
+/** listenResource listens for a specific resource (resourceName, name, namespace)
+ *  and fires events when:
+ *  - the resource is not found
+ *  - the resource is updated (will send the resource without managedFields)
+ *  - the events attached to the resource are updated, optionally
+ */
+export async function listenResource(options: ListenResourceOptions): Promise<IDisposable | undefined> {
```
axel7083, line 53: suggestion(non-blocking): Based on the usage of `listenResource` I am really wondering if we should not create a class extending the svelte/store, this is very similar, have subscribers etc.

For example something like the following

```ts
export class ResourceStore<T> implements Readable<Array<T>> {
....
}
```

I have an example I made for the quadlet extension to contain the logs see [extension-podman-quadlet#packages/frontend/src/stores/logger-store.ts](https://github.com/podman-desktop/extension-podman-quadlet/blob/f12b86cc8fd08e425bb41a0fe5dd82d0cf2dfbf2/packages/frontend/src/stores/logger-store.ts#L16)

packages/renderer/src/lib/kube/resource-listen.ts
```diff
@@ -0,0 +1,132 @@
...
+      eventsUnsubscriber?.();
+    },
+  };
+}
+
+// removes "verbose" fields from a Kubernetes resource description:
+// - metadata.managedFields
+function nonVerbose(resource: KubernetesObject): KubernetesObject {
+  if (resource.metadata?.managedFields) {
+    const result = structuredClone(resource);
```
axel7083, line 123: Why using `structuredClone`, can't we just mutate the original object ?

packages/renderer/src/lib/kube/resource-listen.ts
feloy: > question: why do we delete the managed fields ?

To have the same behaviour as currently (#7593). In non-experimental mode, it is removed when calling `loadDetails` (in main, see 2nd point)

> remark: maybe this filtering is something that should have been done on the main side ? To avoid the unnecessary payload size with the data to finally delete it.

In non-experimental, it is removed only when calling `loadDetails`, but is still present in other places. This information can be interesting for some developers, and could be displayed in a readable form to help us (debug operators for example). I'm trying to remove it only at this level, to keep it in other places.

packages/renderer/src/lib/kube/resource-listen.spec.ts
```diff
@@ -0,0 +1,341 @@
...
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+
+import type { IDisposable } from '../../../../main/src/plugin/types/disposable';
+import { listenResource } from './resource-listen';
+import * as resourcesListen from './resources-listen';
+
+vi.mock(import('./resources-listen'), async importOriginal => {
+  const original = await importOriginal();
+  return {
+    // we want to access the original isKubernetesExperimentalMode
```
axel7083: question: why? why not just mocking the `isKubernetesExperimentalMode`?

packages/renderer/src/lib/kube/resource-listen.ts
feloy: The caller may want to keep the verbose object

packages/renderer/src/lib/service/ServiceDetails.spec.ts
```diff
@@ -41,102 +42,171 @@ const service: V1Service = {
...
+type initListsReturnType = {
+  updateServices: (objects: KubernetesObject[]) => void;
+  updateEvents: (objects: CoreV1Event[]) => void;
+};
 
-  const routerGotoSpy = vi.spyOn(router, 'goto');
+describe.each<{
+  experimental: boolean;
+  initLists: (services: KubernetesObject[], events: CoreV1Event[]) => initListsReturnType;
+}>([
```
axel7083, line 70: :+1: I like such testing !

packages/renderer/src/lib/kube/resource-listen.spec.ts
feloy: Yes, you are right, I'll change to simplify

packages/renderer/src/lib/kube/resource-listen.ts
axel7083: After reading more in depth, we have the associated events, so maybe an object with several stores ? I don't know, throwing some ideas.

```
export class ResourceListener .... {

get events: Readable<Array<KubernetesObjects>> { .. }

}
```

packages/renderer/src/lib/kube/resource-listen.ts
axel7083: I see

feloy:
@cdrage, I'm adding you as reviewer as you have worked on the managedFields

packages/renderer/src/lib/kube/resource-listen.ts
axel7083: You are the caller here, this method is not exported, so do you want to keep the verbose object? :laughing:

packages/renderer/src/lib/kube/resource-listen.ts
feloy: As I understand, stores ~will~ might be deprecated by svelte, I'm not sure it is reasonable to rely on them here. @benoitf wdyt?

axel7083: Store are not deprecated, I can't find any reliable source stating this

packages/renderer/src/lib/kube/resource-listen.ts
feloy: I've made some back and forth on this: I'm still wondering if `listenResource` should return the non verbose version, or if instead `nonVerbose` should be exported and called by the Details component, to send the non verbose version to the YAML Editor, and keep the verbose for the Summary/Inspect components.
@cdrage do you have an opinion on this?

feloy:
@benoitf @dgolovin @gastoner @cdrage could you please have a look on this PR please? Thanks

packages/renderer/src/lib/kube/resource-listen.ts
benoitf: I would not bet on stores, as states can now be part of ts files as well.

But I would not bet also on Readable and Writable stuff

benoitf (approved):
I see room for improvement and for sharing more data across different pages but it can be iterations

PR #11220, opened 2025-02-18
chore: common kubernetes ui objects

deboer-tim:
### What does this PR do?

Creates a new KubernetesObjectUI and uses this as the parent for all Kubernetes UI objects like DeploymentUI, etc., with an optional kind property (*). This improves typing and in future PRs will allow us to have a shared Kubernetes Name column, map from kind to plural (url), and remove Kind.isResource() from KubernetesObjectsList.

(*) Although we could make more properties required over time only name + status are required for now to keep the change manageable. E.g. enforcing kind would require changes in 35 additional test files. Look at this as the first step; we can enforce columns or evolve from here.

UI objects are using hard-coded kinds like `kind: 'Pod'` for now, because the kind is not set on our resources. As we move to the experimental mode this shouldn't be required and we could use `kind: pod.kind ?? ''`. Likewise, the kind doesn't strictly need to be set on the test objects yet but I did this so they'll be correct later.

### Screenshot / video of UI

N/A

Part of #11035. Replaces 90% of #11094 and a follow up PR will share the Kubernetes Name columns.

### How to test this PR?

Just a refactoring, so code review + existing tests.

### DIFF
--- a/packages/renderer/src/lib/node/node-utils.ts
+++ b/packages/renderer/src/lib/node/node-utils.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023-2024 Red Hat, Inc.
+ * Copyright (C) 2023-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -60,6 +60,7 @@ export class NodeUtils {
     const containerRuntime = node.status?.nodeInfo?.containerRuntimeVersion ?? '';
 
     return {
+      kind: 'Node',
       uid: node.metadata?.uid ?? '',
       name: node.metadata?.name ?? '',
       status,
--- a/packages/renderer/src/lib/objects/KubernetesObjectUI.ts
+++ b/packages/renderer/src/lib/objects/KubernetesObjectUI.ts
@@ -19,7 +19,10 @@
 export interface KubernetesObjectUI {
   kind?: string;
   name: string;
-  namespace?: string;
   status: string;
-  selected?: boolean;
+}
+
+export interface KubernetesNamespacedObjectUI extends KubernetesObjectUI {
+  namespace: string;
+  selected: boolean;
 }
--- a/packages/renderer/src/lib/objects/KubernetesObjectUI.ts
+++ b/packages/renderer/src/lib/objects/KubernetesObjectUI.ts
@@ -16,25 +16,10 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import type { V1CronJob } from '@kubernetes/client-node';
-
-import type { CronJobUI } from './CronJobUI';
-
-export class CronJobUtils {
-  getCronJobUI(cronjob: V1CronJob): CronJobUI {
-    return {
-      kind: 'CronJob',
-      uid: cronjob.metadata?.uid ?? '',
-      name: cronjob.metadata?.name ?? '',
-      status: 'RUNNING',
-      namespace: cronjob.metadata?.namespace ?? '',
-      created: cronjob.metadata?.creationTimestamp,
-      selected: false,
-      schedule: cronjob.spec?.schedule ?? '',
-      suspended: cronjob.spec?.suspend ?? false,
-      lastScheduleTime: cronjob.status?.lastScheduleTime,
-      // Get the number of "active" jobs, we just get the length of the array
-      active: cronjob.status?.active?.length ?? 0,
-    };
-  }
+export interface KubernetesObjectUI {
+  kind?: string;
+  name: string;
+  namespace?: string;
+  status: string;
+  selected?: boolean;
 }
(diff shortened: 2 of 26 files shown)


packages/renderer/src/lib/node/node-utils.ts
```diff
@@ -60,6 +60,7 @@ export class NodeUtils {
     const containerRuntime = node.status?.nodeInfo?.containerRuntimeVersion ?? '';
 
     return {
+      kind: 'Node',
       uid: node.metadata?.uid ?? '',
```
axel7083, line 64: does we properly have on the renderer side an `uid` for each object?

I feel like we really should have the uid object, this is something important in Kubernetes

> Each [object](https://kubernetes.io/docs/concepts/overview/working-with-objects/#kubernetes-objects) in your cluster has a [Name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) that is unique for that type of resource. Every Kubernetes object also has a [UID](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids) that is unique across your whole cluster.

If we do not have each Info/UI object with it, we should probably have an issue to add them all somewhere

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

packages/renderer/src/lib/objects/KubernetesObjectUI.ts
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface KubernetesObjectUI {
+  kind?: string;
+  name: string;
+  namespace?: string;
+  status: string;
+  selected?: boolean;
+}
```
axel7083: I would not include `status` nor `selected` as this is not something shared by all KubernetesObject

Not all kubernetes object have statuses, and `selected` is something specific to the Table component, and is not related to the Kubernetes object.

I would go further and split this in two

```ts
export interface KubernetesObjectUI {
	kind: string; // mandatory?? not sure need to have feloy to confirm (maybe it is plural?)
	name: string; // mandatory
	uid: string, // hopefully we can add it? maybe future PR
}

export interface KubernetesNamespacedObjectUI extend KubernetesObjectUI  {
	namespace: string;
}
```

It may feel verbose, but having this strongely type will avoid annoying `object?.namespace ?? 'unknown' ` everywhere. If we **know** from Kubernetes docs, than an object must have an uid, and a name, shouldn't we type them accordingly?

packages/renderer/src/lib/node/node-utils.ts
deboer-tim: I think I only saw uuid on maybe 2-3 of the current 10 objects. Since we are using it on those (like line 64 above) I'd expect it is there correctly for all the others though.

I agree we should add it, just wasn't the primary focus here / isn't needed for any specific use that I'm aware of atm, so I'd prefer to leave that for future PR.

axel7083: Yeah future PR is okay for me

packages/renderer/src/lib/objects/KubernetesObjectUI.ts
deboer-tim: These are UI objects, so all 10 of them already have status today (basically, what color/icon to use in the status column) and 9/10 had selection - only node didn't because we have no bulk actions there (yet). TBH I was tempted to just add it to node now and remove the ? since it seems likely enough at some point in the future.

Since these objects really are just representations of the Kubernetes resource for displaying in the UI, and these two properties already exist on almost every object, I think we should keep them here.

Conversely node is the only object without a namespace, but I can split up the interface so that we can have clear no-? for all the others. 👍🏼

axel7083: I am not really liking this solution, but I will not hold the PR, leaving to other opinion to decide.

> Since these objects really are just representations of the Kubernetes resource for displaying in the UI, and these two properties already exist on almost every object, I think we should keep them here.

I hear this argument, and make sense 👍 . I like separation of concern, generics, and precisely typed objects, so that's why I am having some issues with an interface with a lot of optional fields to try to make it match all

deboer-tim: Ok, new commit splits the interface, and since node happened to be the only child without the `selected` property, I moved it to the new interface and made it required as well, allowing me to remove `namespace` and `selected` from all the other interfaces.

FWIW I looked quickly at adding `uuid` or enforcing `kind`. Both are totally possible, but next PR (shared name column) will delete >25% of the affected files. So: keep this PR scoped and that can be done separately.

packages/renderer/src/lib/node/node-utils.ts
feloy: I vote for having the `uid` in the UI version of the object (in a future PR).

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids

It will be present for all objects, and is important if we want to distinguish different versions of a resource with a same name (a user can delete a resource and create it again with the same name, in this case only the uid is helpful to distinguish them). It is also used to associate Events to resources, or to reference owner via `ownerReferences`, and probably more

axel7083 (approved):

feloy (approved):

PR #11230, opened 2025-02-19
chore: common kubernetes name column

deboer-tim:
### What does this PR do?

Creates a new sharable Kubernetes name column, using the new navigation route to open the correct details page.

Uses the column on all list pages, and deletes all the old column files and tests.

Identical function, far less code! 🎉

### Screenshot / video of UI

N/A

Part of #11035.

### How to test this PR?

Go to each of the Kubernetes pages, confirm the name column hasn't changed and clicking on it goes to the correct details.


packages/renderer/src/lib/objects/KubernetesColumnName.svelte
```diff
@@ -0,0 +1,32 @@
...
+
+interface Props {
+  object: KubernetesObjectUI;
+}
+
+let { object }: Props = $props();
+
+function isNamespaced(object: KubernetesObjectUI): object is KubernetesNamespacedObjectUI {
+  return 'namespace' in object;
+}
```
axel7083: nice 👍

packages/renderer/src/lib/objects/KubernetesColumnName.svelte
```diff
@@ -0,0 +1,32 @@
...
+
+async function openDetails(): Promise<void> {
+  if (isNamespaced(object)) {
+    await window.navigateToRoute('kubernetes', [{ kind: object.kind, name: object.name, namespace: object.namespace }]);
+  } else {
+    await window.navigateToRoute('kubernetes', [{ kind: object.kind, name: object.name }]);
```
axel7083: I did not notice before, but why do we need to pass an array? it only accept a single argument?

packages/renderer/src/lib/objects/KubernetesColumnName.svelte
feloy: I have a `isCoreV1Event` in PR https://github.com/podman-desktop/podman-desktop/pull/11196/files
It would be nice to have all these "is"-functions in a single place, for example in `packages/renderer/src/lib/objects/is-resource`?

axel7083 (approved):
Love to see more deletion that addition!

Great refactoring 👍 , we clearly simplify a lot of code here. 

Tested with a local kind cluster for pods, node and services, and LGTM 🚀

packages/renderer/src/lib/objects/KubernetesColumnName.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
```
axel7083: the `objects` directory may feel a bit meaningless? maybe something `shared` or `common`?

packages/renderer/src/lib/objects/KubernetesColumnName.svelte
deboer-tim: Good catch (and thanks for helping me figure it out). navigateToRoute was missing a spread, fixed in new commit.

packages/renderer/src/lib/objects/KubernetesColumnName.svelte
benoitf: https://github.com/podman-desktop/podman-desktop/tree/main/packages/renderer/src/lib/kube ?

packages/preload/src/index.ts
```diff
@@ -244,7 +244,7 @@ export function initExposure(): void {
     },
   );
 
-  contextBridge.exposeInMainWorld('navigateToRoute', async (routeId: string, args: unknown[]): Promise<void> => {
+  contextBridge.exposeInMainWorld('navigateToRoute', async (routeId: string, ...args: unknown[]): Promise<void> => {
```
benoitf: issue: this commit looks unrelated to the changes

benoitf (commented):
question: wondering why the kubernetes files are not inside the kube folder ?

packages/renderer/src/lib/objects/KubernetesColumnName.svelte
deboer-tim: I added it here b/c the KubernetesObjectList and KubernetesObjectUI that it will always be used with are in this folder ... but I'm not sure why they weren't in /kube either. I'll move the column, but instead of:
  `/kube/KubernetesColumnName.svelte`
how about:
  `/kube/column/Name.svelte`
to match the proposal that @axel7083 had on column naming?

packages/renderer/src/lib/objects/KubernetesColumnName.svelte
deboer-tim: It was suggested to move the columns under /kube, and we already have a pattern of creating (e.g.) /node/node-utils.ts so I created a kube-utils.ts and moved isNamespaced() there.

packages/renderer/src/lib/objects/KubernetesColumnName.svelte
deboer-tim: I liked it, so new commit moves the column to `kube/column/Name.svelte`.

packages/preload/src/index.ts
deboer-tim: Correct - unrelated fix found by Axel above. But since the fix is only 3 characters and this PR both removes the one old use of it (DeploymentColumnName) and adds the replacement (kube/column/Name) I was hoping I could get away with it 😀.

benoitf: I would issue just a small PR for that as it's a separate fix

also small PRs have a merge approval within minutes usually

deboer-tim: Ok: #11232

deboer-tim:
Rebased on main to pick up #11232.

benoitf (approved):

Issue #10603, opened 2025-01-10
Remove disablement of no-explicit any from renderer package


jeffmaury:
### Is your enhancement related to a problem? Please describe

Impacted files:

- packages/renderer/src/App.svelte
- packages/renderer/src/lib/actions/ActionUtils.ts https://github.com/podman-desktop/podman-desktop/pull/11278
- packages/renderer/src/lib/actions/ContributionActions.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11278
- packages/renderer/src/lib/appearance/appearance-utils.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11282
- packages/renderer/src/lib/carousel/Carousel.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11396
- packages/renderer/src/lib/carousel/Carousel.svelte  https://github.com/podman-desktop/podman-desktop/pull/11396
- packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11396
- packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretList.spec.ts N/A, already done
- packages/renderer/src/lib/container/ContainerExport.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11477
- packages/renderer/src/lib/container/ContainerList.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11477
- packages/renderer/src/lib/container/ContainerListCompose.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11477
- packages/renderer/src/lib/context/contextKey.spec.ts **Exception? This is copy-and-pasted from upstream sources**
- packages/renderer/src/lib/context/contextKey.ts **Exception? This is copy-and-pasted from upstream sources**
- packages/renderer/src/lib/dashboard/ProviderCard.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11478
- packages/renderer/src/lib/dashboard/ProviderConfigured.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11478
- packages/renderer/src/lib/dashboard/ProviderInstalled.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11478
- packages/renderer/src/lib/dashboard/ProviderReady.spec.ts https://github.com/podman-desktop/podman-desktop/pull/11478
[...]

PR #11282, opened 2025-02-20
chore: remove no-explicit-any from lib/appearance

cdrage:
chore: remove no-explicit-any from lib/appearance

### What does this PR do?

Removes no-explicit-any

### Screenshot / video of UI

N/A

Part of https://github.com/podman-desktop/podman-desktop/issues/10603

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/appearance/appearance-utils.spec.ts
+++ b/packages/renderer/src/lib/appearance/appearance-utils.spec.ts
@@ -25,19 +25,19 @@ import { AppearanceUtil } from './appearance-util';
 
 const appearanceUtil: AppearanceUtil = new AppearanceUtil();
 
-const mockedMatches = vi.fn();
+const matchMediaMock = vi.fn();
 
 beforeAll(() => {
-  Object.defineProperty(window, 'matchMedia', { value: vi.fn() });
+  Object.defineProperty(window, 'matchMedia', { value: matchMediaMock });
 });
 
 beforeEach(() => {
   vi.resetAllMocks();
-  vi.mocked(window.matchMedia).mockReturnValue({
-    matches: mockedMatches,
+  matchMediaMock.mockReturnValue({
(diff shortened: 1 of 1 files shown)


feloy (approved):

packages/renderer/src/lib/appearance/appearance-utils.spec.ts
```diff
@@ -27,115 +25,115 @@ import { AppearanceUtil } from './appearance-util';
...
-    getConfigurationValueMock.mockResolvedValue(customTheme);
+    vi.mocked(window.getConfigurationValue).mockResolvedValue(customTheme);
 
     const theme = await appearanceUtil.getTheme();
     expect(theme).toBe(customTheme);
   });
 
   test('should return custom value even if os is dark when matches found', async () => {
-    (window as any).matchMedia = vi.fn().mockReturnValue({
+    vi.mocked(window.matchMedia).mockReturnValue({
```
benoitf: it looks like this instruction should be in the beforeEach block as being copied in all the tests

cdrage: I put it here since we use true/false in each different test.

benoitf: vi.mocked(window.matchMedia).mockReturnValue is almost the same so you could return empty in before each and only override the value in a specific test

benoitf: we return an empty object in many cases

gastoner (approved):

benoitf (approved):

cdrage:
@benoitf Did a bit more organizing, see the latest commit. Had to fix up the tests to completely mock matchMedia, as a partial mock does not work. Let me know if this is okay.

benoitf (approved):

Issue #11277, opened 2025-02-20
display that a cluster is "offline"


feloy:
### Is your enhancement related to a problem? Please describe

In Kubernetes experimental mode, when the informers are able to connect to the cluster, then are disconnected, they are marked as "offline" in the backend.

At this point, the caches still contain the resources as they were when the informers have been disconnected, and they are still displayed in the different pages. This can be useful for the user, to examine the resources in the state they were just before the connection was lost.

### Describe the solution you'd like

To explain to the user this specific state, it would be helpful to display a state 'offline', or 'connection lost'.

- [x] in Kubernetes Contexts page (Settings > Kubernetes): https://github.com/podman-desktop/podman-desktop/pull/11285
- [x] in Kubernetes resources pages (`KubernetesCurrentContextConnectionBadge`): https://github.com/podman-desktop/podman-desktop/pull/11381

PR #11285, opened 2025-02-20
feat: display offline status in Kubernetes Contexts page

feloy:

### What does this PR do?

Display offline status in Kubernetes Contexts page

(see details in #11277)

### Screenshot / video of UI

https://github.com/user-attachments/assets/421e4d62-0f95-4e57-872d-c0d46f2a5d9f

Part of #11277 

### How to test this PR?

Start Podman Desktop with the current context accessible, then stop the cluster

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
@@ -20,6 +20,10 @@ import ListItemButtonIcon from '../ui/ListItemButtonIcon.svelte';
 import SettingsPage from './SettingsPage.svelte';
 
 interface KubeContextWithStates extends KubeContext {
+  // some informers have been disconnected, and their caches are still populated with the last seen resources
+  isOffline: boolean;
+  // the context has been marked as reachable (during health check in experimental mode)
+  // the context will still be marked as reachable even if it is offline
   isReachable: boolean;
   isKnown: boolean;
   isBeingChecked: boolean;
@@ -40,6 +44,7 @@ const kubernetesContextsWithStates: KubeContextWithStates[] = $derived(
     .map(kubeContext => ({
       ...kubeContext,
       isReachable: isContextReachable(kubeContext.name, experimentalStates),
+      isOffline: isContextOffline(kubeContext.name, experimentalStates),
       isKnown: isContextKnown(kubeContext.name, experimentalStates),
       isBeingChecked: isContextBeingChecked(kubeContext.name, experimentalStates),
       podsCount: getResourcesCount(kubeContext.name, 'pods', experimentalStates),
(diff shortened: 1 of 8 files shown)


axel7083 (approved):
question (non-blocking): shouldn't we see a retry / connect button when the connection is offline

otherwise LGTM, I can reproduce the demo video

feloy:
> question (non-blocking): shouldn't we see a retry / connect button when the connection is offline

Yes, this will be part of other PR(s)

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -243,14 +257,25 @@ async function connect(contextName: string): Promise<void> {
         <div class="grow flex-column divide-gray-900 text-[var(--pd-invert-content-card-text)]">
           <div class="flex flex-row">
             <div class="flex-none w-36">
-              {#if context.isReachable}
+              {#if context.isReachable || context.isOffline}
```
deboer-tim, line 260: nit: I think:
```
{#if context.isOffline}
...
{:else if context.isReachable}
...
{/if}
```
would be more readable than the embedded if.

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
```diff
@@ -243,14 +257,25 @@ async function connect(contextName: string): Promise<void> {
...
                 <div class="flex flex-row pt-2">
-                  <div class="w-3 h-3 rounded-full bg-[var(--pd-status-connected)]"></div>
-                  <div
-                    class="ml-1 font-bold text-[9px] text-[var(--pd-status-connected)]"
-                    aria-label="Context Reachable">
-                    REACHABLE
-                  </div>
+                  {#if context.isOffline}
+                    <Tooltip class="flex flex-row" tip="connection to the cluster lost, the resources displayed can be out of date">
```
deboer-tim: suggestion: The message is a bit long (and cluster/displayed are implied), 'may be' is more natural than 'can be', and maybe 'out of sync' is more common?

'connection lost, resources may be out of sync'?

deboer-tim (approved):
LGTM, just a couple suggestions.

Not related to this PR, but it threw me a bit when you changed the context in the video and the old context became 'unknown'. I'm not sure how else we'd show it, but didn't feel like 'we intentionally disconnected' or connected to your action. Maybe in the future there's some graceful disconnect delay or another way we can show that.

packages/renderer/src/lib/preferences/PreferencesKubernetesContextsRendering.svelte
feloy: But there are 30 lines in common, after the embedded if

PR #11288, opened 2025-02-20
chore: kubernetes icon component

deboer-tim:
### What does this PR do?

Every Kubernetes resource kind has an associated icon, so when we know the kind why not look it up vs having to specify the icon everywhere? This creates a utility function in kube-utils to return the icon based on kind, and uses that in the dashboard resource card to remove the property.

Isn't this just moving this function? Sure. But it means we can now look up the icon *anywhere* we know the kind, e.g. anywhere the UI objects exist. The next PR can create a shared status column and remove the individual status columns from all lists. We might be able to do something similar with the empty pages.

### Screenshot / video of UI

N/A

Part of #11035.

### How to test this PR?

Verify Kubernetes dashboard resource card icons are unchanged.

### DIFF
--- a/packages/renderer/src/lib/images/ConfigMapSecretIcon.svelte
+++ b/packages/renderer/src/lib/images/ConfigMapSecretIcon.svelte
@@ -9,6 +9,7 @@ export let size = '40';
   style={$$props.style}
   viewBox="0 0 16.015625 16"
   version="1.1"
+  data-testid="ConfigMap Secret"
   xml:space="preserve"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:xlink="http://www.w3.org/1999/xlink">
--- a/packages/renderer/src/lib/images/CronJobIcon.svelte
+++ b/packages/renderer/src/lib/images/CronJobIcon.svelte
@@ -15,6 +15,7 @@ let style: string = baseStyle + (solid ? 'fill:currentColor' : 'fill:none');
   height={size}
   viewBox="1 1 6.4666665 6.4666666"
   version="1.1"
+  data-testid="CronJob"
   xml:space="preserve"
   xmlns="http://www.w3.org/2000/svg"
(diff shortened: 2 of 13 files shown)


packages/renderer/src/lib/kube/kube-utils.ts
```diff
@@ -16,8 +16,46 @@
...
+    case 'Route':
+      return IngressRouteIcon;
+    case 'ConfigMap':
+      return ConfigMapSecretIcon;
+    case 'Secret':
+      return ConfigMapSecretIcon;
+    case 'PersistentVolumeClaim':
+      return PvcIcon;
+    case 'CronJob':
+      return CronJobIcon as Component;
```
axel7083: Why this one need to be casted ?

deboer-tim: It has a different type compared to the others (`Component<Props, {}, "">`) since it's our first Svelte 5 prop icon. Charlie mentioned it when he added it to the dashboard and just copied here (i.e. not related or changed by this PR). The function could return `Component<any>` instead, don't know if there's a better option.

gastoner (approved):
LGTM on Windows

benoitf (changes requested):
I dislike the exported function pattern. It makes tons of import in a kube-utils file that are really specific to icons

if I look at where it's used

we do
```
let Icon = $derived(getIcon(kind));
<Icon size={40}/></div
```

so why have an helper function to then call a object to render it

vs having a `<KubernetesIcon>` component where I specify the kind property

so, all icons imports are in a single component that can be easily tested and has a proper scope

deboer-tim:
> [...]
> 
> vs having a `<KubernetesIcon>` component where I specify the kind property
> 
> so, all icons imports are in a single component that can be easily tested and has a proper scope

Yes, that does make more sense. Since the scope is different I'll close this and open a new PR so that it's clean.

Issue #10673, opened 2025-01-15
Implementation mirror registries configuration


slemeur:
### Is your feature request related to a problem? Please describe

As described in the issue #10550, this ticket is dedicated to the implementation of the support of configuring mirror registries

### Describe the solution you'd like

See in #10550

Issue #10327, opened 2024-12-12
Podman extension structure enhancement


axel7083:

axel7083:
## Iteration 1 - Folder structure

Without changing any code source (expect imports) going from flat to tree structures. Bellow is a very basic grouping of the files. The goal would be to only have the `extension.ts` at the root.

> For clarity I did not included the `*.spec` files.

```bash
.
└── src/
    ├── helpers/
    │   ├── wsl-helper.ts
    │   ├── qemu-helper.ts
    │   ├── krunkit-helper.ts
    │   ├── podman-info-helper.ts
    │   └── podman-binary-location-helper.ts
    ├── utils/
    │   ├── util.ts
    │   ├── warnings.ts
    │   ├── powershell.ts
    │   ├── podman-configuration.ts
    │   ├── podman-cli.ts 
    │   ├── compatibility-mode.ts
    │   ├── podman-install.ts
    │   └── registry-setup.ts
    │   └── podman-machine-stream.ts
    ├── checks/
    │   ├── base-check.ts
    │   ├── detection-checks.ts
    │   └── macos-checks.ts
    ├── remote/
    │   ├── podman-remote-ssh-tunnel.ts
    │   ├── podman-remote-connections.ts
    ├── cleanup/
    │   ├── podman-cleanup-abstract.ts
    │   ├── podman-cleanup-macos.ts
    │   └── podman-cleanup-windows.ts
    ├── extension.ts 
    └── podman5.json
``` 

## Iteration 2 - Splitting files

Some rules I think would be nice to follow (Open to discussion, let's see which guidelines we want to follow!)
- One class / interface per file
- filename should match the class / interface in it

Some files are very complex (too complex), here is an example `podman-install.ts` has 12 classes defined, 6 functions, some constants and some interfaces. Here is a breakdown from my IDE.

![Image](https://github.com/user-attachments/assets/0cdaa44c-9e45-4cb7-a9d6-20326aef505e)

axel7083:
# Steps 

## Moving files

ℹ this step should limit the code change

- https://github.com/podman-desktop/podman-desktop/pull/11218
- https://github.com/podman-desktop/podman-desktop/pull/11249
- https://github.com/podman-desktop/podman-desktop/pull/11315
- https://github.com/podman-desktop/podman-desktop/pull/11354

## Splitting files

TODO

axel7083:
[...]

PR #11304, opened 2025-02-24
chore: mount the registries.conf inside the podman VM

benoitf:
### What does this PR do?
when creating a podman machine, mount the registries.conf file from the host into the VM allowing to sync the configuration of registry mirroring
note: do this only starting from podman 5.4.0+ cli

related to https://github.com/podman-desktop/podman-desktop/issues/10673

### How to test this PR?

1. check podman version is 5.4.0+
1. create a file into your host: `$HOME/.config/containers/registries.conf`

with the content, for example:
```toml
[[registry]]
location = "docker.io"

[[registry.mirror]]
location = "ghcr.io"
```

1. create a podman machine from Podman Desktop UI

1. connect into the podman machine and check the content of `/etc/containers/registries.conf.d/999-podman-desktop-registries-from-host.conf`

1. using cli or podman desktop, pull the image 'docker.io/httpd'
1. ensure it works

1. try to pull an invalid image like `docker.io/foofoofoofoofoo` but using the CLI (as the compat API is not sending the full error message)

1. ensure you see it checks on the mirror:

```
podman pull docker.io/foofoofoofoofoo
... (Mirrors also failed: [ghcr.io/library/foofoofoofoofoo:latest:  ....
```

### DIFF
--- a/extensions/podman/packages/extension/src/configuration/registry-configuration.ts
+++ b/extensions/podman/packages/extension/src/configuration/registry-configuration.ts
@@ -17,10 +17,9 @@
  ***********************************************************************/
 
 import { mkdir, writeFile } from 'node:fs/promises';
-import { homedir } from 'node:os';
+import { homedir, tmpdir } from 'node:os';
 import { dirname, resolve } from 'node:path';
 
-import type { ExtensionContext } from '@podman-desktop/api';
 import { env } from '@podman-desktop/api';
 import mustache from 'mustache';
 
@@ -36,8 +35,6 @@ export interface RegistryConfiguration {
  * Manages the registry configuration file (inside the Podman VM for macOS/Windows)
  */
 export class RegistryConfigurationImpl implements RegistryConfiguration {
-  constructor(private readonly context: ExtensionContext) {}
-
   // provides the path to the file being on the host
   // $HOME/.config/containers/registries.conf
   getRegistryConfFilePath(): string {
@@ -77,7 +74,7 @@ export class RegistryConfigurationImpl implements RegistryConfiguration {
     });
 
     // write the content to a temp file inside the storage folder of the extension
-    const playbookFile = resolve(this.context.storagePath, 'podman-machine', 'playbook-setup-registry-conf-file.yml');
+    const playbookFile = resolve(tmpdir(), 'podman-desktop', 'podman-machine', 'playbook-setup-registry-conf-file.yml');
     // create the folder if it doesn't exist
     const parentFolder = dirname(playbookFile);
     await mkdir(parentFolder, { recursive: true });
--- a/extensions/podman/packages/extension/src/configuration/registry-configuration.ts
+++ b/extensions/podman/packages/extension/src/configuration/registry-configuration.ts
@@ -1,0 +1,91 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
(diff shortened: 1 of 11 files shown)


extensions/podman/packages/extension/src/configuration/registry-configuration.ts
```diff
@@ -0,0 +1,91 @@
+/**********************************************************************
```
axel7083, line 1: new folder <3 thanks for taking https://github.com/podman-desktop/podman-desktop/issues/10327 into consideration

extensions/podman/packages/extension/src/configuration/registry-configuration.ts
```diff
@@ -0,0 +1,91 @@
...
+  async getPlaybookScriptPath(): Promise<string> {
+    // create the content of the file
+
+    const playbookScriptContent = mustache.render(playbookRegistryConfFileTemplate, {
+      configurationFileInsideVmPath: this.getPathToRegistriesConfInsideVM(),
+      configurationFileName: '999-podman-desktop-registries-from-host.conf',
+    });
+
+    // write the content to a temp file inside the storage folder of the extension
+    const playbookFile = resolve(this.context.storagePath, 'podman-machine', 'playbook-setup-registry-conf-file.yml');
```
axel7083: Shouldn't we use something like `tmpdir` from `node:os` to let the OS do the cleanup on reboot? does this file is overwritten everytime we create a machine?

benoitf: > does this file is overwritten everytime we create a machine?

yes it is recreated each time we ask to create a machine and the lifetime we expect to have it is until the podman machine is fully started

axel7083 (approved):
On Windows (version 5.3.2) I cannot fully test, but I can confirm the following

1. `playbook.yaml` is available in the newly created machine
Here is the content
```yaml
---
- name: Create a symbolic link for registries.conf from host to VM
  hosts: localhost
  become: true
  tasks:
    - name: Create symbolic link with sudo from the host
      command: sudo ln -s /mnt/c/Users/axels/.config/containers/registries.conf /etc/containers/registries.conf.d/999-podman-desktop-registries-from-host.conf
```

If I run manually the command, **and restart the podman machine** (restart is needed)
I am able to see the following
```
$: podman --connection=demo-playbook image pull docker.io/foofoofoofoofoo
Trying to pull docker.io/library/foofoofoofoofoo:latest...
Error: initializing source docker://foofoofoofoofoo:latest: (Mirrors also failed: [ghcr.io/library/foofoofoofoofoo:latest: Requesting bearer token: invalid status code from registry 403 (Forbidden)]): docker.io/library/foofoofoofoofoo:latest: reading manifest latest in docker.io/library/foofoofoofoofoo: requested access to the resource is denied
```

We can see that the mirror is properly tested so take my approval 👍

extensions/podman/packages/extension/src/configuration/registry-configuration.ts
axel7083: Still think that we should use tmpdir for this file instead of the extension's storage

benoitf: moved to a temporary directory

benoitf:
moved generated file to a temporary directory (thus removing all extension context parameters)
rebased

PR #11305, opened 2025-02-24
feat: display node details in experimental mode

feloy:

### What does this PR do?

Display node details in experimental mode

Part of #10657 

### How to test this PR?

- go to Kubernetes > Nodes > Details
- check that the summary of node is displayed
- check that the node can be updated from the Kube tab (add annotation, etc)
- create and delete events for the node(1) and check they are displayed reactively in the Summary page

(1) you can create events manually with:
```
apiVersion: v1
count: 1
involvedObject:
  apiVersion: v1
  kind: Node
  name: node1                                <-- replace with name of node
  uid: e9cef2cc-2fb8-409c-9c92-ff4f57e247fb <-- replace with uid of node
kind: Event
lastTimestamp: "2025-02-17T15:05:49Z"
message: test message
metadata:
  creationTimestamp: "2025-02-17T08:22:46Z"
  name: node1-event1
reason: BackOff
reportingComponent: kubelet
reportingInstance: kind-cluster-control-plane
source:
  component: kubelet
  host: kind-cluster-control-plane
type: Normal
```

### DIFF
--- a/packages/renderer/src/lib/node/NodeDetails.spec.ts
+++ b/packages/renderer/src/lib/node/NodeDetails.spec.ts
@@ -26,7 +26,7 @@ import { beforeEach, describe, expect, test, vi } from 'vitest';
 import * as states from '/@/stores/kubernetes-contexts-state';
 
 import type { IDisposable } from '../../../../main/src/plugin/types/disposable';
-import * as resourcesListen from '../kube/resources-listen';
+import { isKubernetesExperimentalMode, listenResources } from '../kube/resources-listen';
 import NodeDetails from './NodeDetails.svelte';
 import * as nodeDetailsSummary from './NodeDetailsSummary.svelte';
 
@@ -88,23 +88,21 @@ describe.each<{
     initLists: (nodes: KubernetesObject[], events: CoreV1Event[]): initListsReturnType => {
       let nodesCallback: (resources: KubernetesObject[]) => void;
       let eventsCallback: (resources: CoreV1Event[]) => void;
-      vi.mocked(resourcesListen.listenResources).mockImplementation(
-        async (resourceName, _options, cb): Promise<IDisposable> => {
-          if (resourceName === 'nodes') {
-            nodesCallback = cb;
-            setTimeout(() => nodesCallback(nodes));
-            return {
-              dispose: (): void => {},
-            };
-          } else {
-            eventsCallback = cb;
-            setTimeout(() => eventsCallback(events));
-            return {
-              dispose: (): void => {},
-            };
-          }
-        },
-      );
+      vi.mocked(listenResources).mockImplementation(async (resourceName, _options, cb): Promise<IDisposable> => {
+        if (resourceName === 'nodes') {
+          nodesCallback = cb;
+          setTimeout(() => nodesCallback(nodes));
+          return {
+            dispose: (): void => {},
+          };
+        } else {
+          eventsCallback = cb;
+          setTimeout(() => eventsCallback(events));
+          return {
+            dispose: (): void => {},
+          };
+        }
+      });
       return {
         updateNodes: (updatedObjects: KubernetesObject[]): void => {
           nodesCallback(updatedObjects);
@@ -117,7 +115,7 @@ describe.each<{
   },
 ])('is experimental: $experimental', ({ experimental, initLists }) => {
   beforeEach(() => {
-    vi.mocked(resourcesListen.isKubernetesExperimentalMode).mockResolvedValue(experimental);
+    vi.mocked(isKubernetesExperimentalMode).mockResolvedValue(experimental);
   });
 
   test('Confirm renders node details', async () => {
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/node/NodeDetails.spec.ts
```diff
@@ -38,63 +40,136 @@ const node: V1Node = {
...
+  updateEvents: (objects: CoreV1Event[]) => void;
+};
+
+describe.each<{
+  experimental: boolean;
+  initLists: (nodes: KubernetesObject[], events: CoreV1Event[]) => initListsReturnType;
+}>([
+  {
+    experimental: false,
+    initLists: (nodes: KubernetesObject[], events: CoreV1Event[]): initListsReturnType => {
```
feloy, line 71: I will factorize this code which is duplicated in ServiceDetails and DeploymentDetails tests in a specific PR

axel7083 (approved):

packages/renderer/src/lib/node/NodeDetails.spec.ts
```diff
@@ -18,13 +18,15 @@
...
 import { writable } from 'svelte/store';
 import { router } from 'tinro';
-import { beforeAll, expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 
-import * as kubeContextStore from '/@/stores/kubernetes-contexts-state';
+import * as states from '/@/stores/kubernetes-contexts-state';
 
+import type { IDisposable } from '../../../../main/src/plugin/types/disposable';
+import * as resourcesListen from '../kube/resources-listen';
```
benoitf: question: why do we use wildcard imports vs specific imports ?

packages/renderer/src/lib/node/NodeDetails.spec.ts
```diff
@@ -18,13 +18,15 @@
...
 import type { CoreV1Event, KubernetesObject, V1Node } from '@kubernetes/client-node';
-import { render, screen, waitFor } from '@testing-library/svelte';
+import { render, screen } from '@testing-library/svelte';
 import { writable } from 'svelte/store';
 import { router } from 'tinro';
-import { beforeAll, expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 
-import * as kubeContextStore from '/@/stores/kubernetes-contexts-state';
+import * as states from '/@/stores/kubernetes-contexts-state';
```
benoitf, line 26: same question there

packages/renderer/src/lib/node/NodeDetails.svelte
```diff
@@ -1,17 +1,19 @@
...
+import type { CoreV1Event, KubernetesObject, V1Node } from '@kubernetes/client-node';
 import { StatusIcon, Tab } from '@podman-desktop/ui-svelte';
-import { onMount } from 'svelte';
+import { onDestroy, onMount } from 'svelte';
 import { router } from 'tinro';
 import { stringify } from 'yaml';
 
 import { kubernetesCurrentContextEvents, kubernetesCurrentContextNodes } from '/@/stores/kubernetes-contexts-state';
 
+import type { IDisposable } from '../../../../main/src/plugin/types/disposable';
```
benoitf, line 10: Isn't it an issue to get the Disposable type from the server side there ?

benoitf: usually this is used for data sent from the server side to the client side but here it's really used only on client side

packages/renderer/src/lib/node/NodeDetails.spec.ts
feloy: No need for this one, I've changed

packages/renderer/src/lib/node/NodeDetails.spec.ts
feloy: I don't how to write this without wildcard import:

```
      vi.mocked(states).kubernetesCurrentContextNodes = nodesStore;
```

packages/renderer/src/lib/node/NodeDetails.svelte
feloy: Right. But I wonder which one to use in `listenResource` and `listenResources`. The native `Disposable`? The `Disposable` provided by podman-desktop/api?

benoitf: I mean, it should be in a package directly available in case we would like to use such pattern on the client side

for now this type is only a server side type

benoitf (approved):
no blockers but Disposable should not be imported from the server side package

packages/renderer/src/lib/node/NodeDetails.svelte
feloy: I've created the issue https://github.com/podman-desktop/podman-desktop/issues/11311 to work on this

Issue #10655, opened 2025-01-15
feat: represent non accessible Kubernetes resources in the frontend


feloy:
### Is your feature request related to a problem? Please describe

Backend provides the information for each Kubernetes resource kind if the kind is accessible or not by the user. 

### Describe the solution you'd like

Represent this accessibility, with the help of #10385 

### Additional context

- [x] Kubernetes Contexts List: https://github.com/podman-desktop/podman-desktop/pull/10856
- [ ] Kubernetes dashboard: 
  - [x] https://github.com/podman-desktop/podman-desktop/issues/11106

~- [ ]  [ ] Kubernetes Resources (deployment, etc) List: https://github.com/podman-desktop/podman-desktop/issues/11107~ (tracked separately)

#### How to create a context with a specific user and give this user some permissions

Create a new user for a kind cluster: https://medium.com/@lionelvillard/creating-users-in-kind-cluster-6c5ee35db3fe

By default, the user won't have any permissions.

You can give permissions to Pod only with:

```
$ cat <<EOF | kubectl apply -f -
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: user1
rules:
- apiGroups: [""]
  resources: ["pods"]
  verbs: ["get", "watch", "list"]

---

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: user1
subjects:
- kind: User
  name: user1
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: user1
  apiGroup: rbac.authorization.k8s.io
EOF
```

Or to Deployment only:

```
cat <<EOF | kubectl apply -f -
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: user1
rules:
- apiGroups: ["apps"]
  resources: ["deployments"]
  verbs: ["get", "watch", "list"]

---

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: user1
subjects:
- kind: User
  name: user1
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: user1
  apiGroup: rbac.authorization.k8s.io
EOF
```

PR #11307, opened 2025-02-24
feat: added visual indicator for kubernetes permissions

### LINKED ISSUE #11107: feat: represent Kubernetes resources permissions on resources lists

Kubernetes Experimental mode provides user's permissions on each resource (for each context), see `packages/renderer/src/stores/kubernetes-context-permission.ts`.

We can provide visual information to the user, with respect to design: https://github.com/podman-desktop/podman-desktop/issues/10385#issuecomment-2615490295 (third screen)

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds visual indicator for users to differentiate between permissions to kubernetes resources

### Screenshot / video of UI
Mockup (only for deployments): 
Image

Closes #11107 

### How to test this PR?
Create an user with not all permissions: 
https://github.com/podman-desktop/podman-desktop/issues/10655

## Summary by Sourcery

Implements a visual indicator for Kubernetes deployment permissions. It disables KubeActions and updates the empty screen message based on user permissions.

New Features:
- Adds a visual indicator to differentiate between permissions for Kubernetes deployments, disabling KubeActions and displaying a message on the empty screen when the user does not have the required permissions.

Tests:
- Adds a test case to verify that the KubeApplyYAMLButton is disabled when the disabled property is set to true.

### DIFF
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
@@ -6,7 +6,7 @@ import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
 <KubernetesEmptyScreen
   icon={ConfigMapSecretIcon}
   resources={['configmaps', 'secrets']}
-  titleEmpty={['No configmaps', 'No secretes']}
+  titleEmpty={['No configmaps', 'No secrets']}
   titleNotPermitted={['Configmaps not accessible', 'Secrets not accessible']}
   message="Try switching to a different context or namespace" />
 
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
@@ -6,7 +6,7 @@ import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
 <KubernetesEmptyScreen
   icon={ConfigMapSecretIcon}
   resources={['configmaps', 'secrets']}
-  titleEmpty={['No configmaps', 'No secrets']}
-  titleNotPermitted={['Configmaps not accessible', 'Secrets not accessible']}
+  titleEmpty='No configmaps or secrets'
+  titleNotPermitted='Configmaps and secrets not accessible'
   message="Try switching to a different context or namespace" />
 
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
@@ -5,6 +5,8 @@ import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
 
 <KubernetesEmptyScreen
   icon={ConfigMapSecretIcon}
-  title="No configmaps or secrets"
+  resources={['configmaps', 'secrets']}
+  titleEmpty={['No configmaps', 'No secretes']}
+  titleNotPermitted={['Configmaps not accessible', 'Secrets not accessible']}
   message="Try switching to a different context or namespace" />
 
--- a/packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
@@ -14,5 +14,5 @@ onMount(() => {
 });
 </script>
 
-<KubernetesEmptyScreen icon={DeploymentIcon} isPermited={deploymentsPermitted} title={deploymentsPermitted ? 'No deployments' : 'Deployments not accessible'} message="Try switching to a different context or namespace" />
+<KubernetesEmptyScreen icon={DeploymentIcon} isPermitted={deploymentsPermitted} title={deploymentsPermitted ? 'No deployments' : 'Deployments not accessible'} message="Try switching to a different context or namespace" />
 
--- a/packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
@@ -1,7 +1,18 @@
 <script lang="ts">
+import { onMount } from 'svelte';
+
 import DeploymentIcon from '../images/DeploymentIcon.svelte';
 import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
+import { listenResourcePermitted } from '../kube/resource-permission';
+
+let deploymentsPermitted: boolean = $state(true);
+
+onMount(() => {
+  return listenResourcePermitted('deployments', (permitted: boolean) => {
+    deploymentsPermitted = permitted;
+  });
+});
 </script>
 
-<KubernetesEmptyScreen icon={DeploymentIcon} title="No deployments" message="Try switching to a different context or namespace" />
+<KubernetesEmptyScreen icon={DeploymentIcon} isPermited={deploymentsPermitted} title={deploymentsPermitted ? 'No deployments' : 'Deployments not accessible'} message="Try switching to a different context or namespace" />
 
--- a/packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
+++ b/packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
@@ -1,25 +1,7 @@
 <script lang="ts">
-import { onDestroy, onMount } from 'svelte';
-
-import type { IDisposable } from '/@api/disposable';
-
 import DeploymentIcon from '../images/DeploymentIcon.svelte';
 import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
-import { listenResourcePermitted } from '../kube/resource-permission';
-
-let deploymentsPermitted: boolean = $state(true);
-let disposable: IDisposable | undefined;
-
-onMount(async () => {
-  disposable = await listenResourcePermitted('deployments', (permitted: boolean) => {
-    deploymentsPermitted = permitted;
-  });
-});
-
-onDestroy(() => {
-  disposable?.dispose();
-});
 </script>
 
-<KubernetesEmptyScreen icon={DeploymentIcon} title={deploymentsPermitted ? 'No deployments' : 'Deployments not accessible'} message="Try switching to a different context or namespace" />
+<KubernetesEmptyScreen icon={DeploymentIcon} resource='deployments' titleEmpty='No deployments' titleNotPermitted='Deployments not accessible' message="Try switching to a different context or namespace" />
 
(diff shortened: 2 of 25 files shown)


packages/renderer/src/lib/service/ServicesEmptyScreen.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import '@testing-library/jest-dom/vitest';
+
+import { render, screen } from '@testing-library/svelte';
+import { afterAll, expect, test, vi } from 'vitest';
+
+import { getResourcePermitted } from '/@/stores/kubernetes-context-permission';
+
+import ServicesEmptyScreen from './ServiceEmptyScreen.svelte';
```
deboer-tim: Only noticed due to sorting, but this file should be renamed ServiceEmptyScreen to match the component.

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
```diff
@@ -1,10 +1,32 @@
...
+
+import { getResourcePermitted } from '/@/stores/kubernetes-context-permission';
+import { kubernetesContexts } from '/@/stores/kubernetes-contexts';
+
 import ConfigMapSecretIcon from '../images/ConfigMapSecretIcon.svelte';
 import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
+
+let contextName = $derived($kubernetesContexts.find(c => c.currentContext)?.name);
+let experimentalStates: boolean = $state(false);
+let configMapsSecretsPermitted: boolean = $derived(
```
deboer-tim: Configmaps could be accessible and secrets not or vice-versa, but message only changes when neither is? Ditto for ingress/routes.

deboer-tim (commented):
This is a lot of change and most of it is boilerplate / same for every page. I think only the message for nodes is different? It looks like it would be possible to move the check for getResourcePermitted() to KubernetesEmptyPage and maybe the message to avoid so much identical code.

Not related since one issue per PR, but FYI with #11317 the icon could be replaced by a kind property too and KubernetesEmptyPage uses KubernetesIcon.

feloy:
There are several things done in a single PR: at least actions, emptyScreen

Could you please make separate PRs for each change?

gastoner:
PR https://github.com/podman-desktop/podman-desktop/pull/11473 needs to be merged before this
cc @feloy

gastoner:
This PR will be rebased after the first one is merged

packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
```diff
@@ -1,7 +1,18 @@
...
+import { onMount } from 'svelte';
+
 import DeploymentIcon from '../images/DeploymentIcon.svelte';
 import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
+import { listenResourcePermitted } from '../kube/resource-permission';
+
+let deploymentsPermitted: boolean = $state(true);
+
+onMount(() => {
+  return listenResourcePermitted('deployments', (permitted: boolean) => {
```
feloy: Finally, returning the result of the `listen...` does not work as we expected when working on this code, because the value is a Promise and not synchronous:

```
If a function is returned synchronously from onMount, it will be called when the component is unmounted.
```

You will need to save the disposable and call it from `onDestoy`

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
```diff
@@ -4,9 +4,11 @@ import type { ComponentProps } from 'svelte';
...
 
-let { icon, ...restProps }: ComponentProps<EmptyScreen> = $props();
+let { icon, isPermitted, ...restProps }: ComponentProps<EmptyScreen> = $props();
 </script>
 
 <EmptyScreen icon={icon} {...restProps}>
-  <KubernetesCheckConnection />
+  {#if isPermitted}
+    <KubernetesCheckConnection />
+  {/if}
```
feloy: I don't think this is related.

The connection is at the cluster level, and a user may want to restart the connection from this page, even if access to Deployment is not permitted?

feloy (changes requested):

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
gastoner: It is in the mockup (in the desc of this PR), I was raly just implementing the mockup here

feloy: (I understand that this change is visible in the mockup, but is not 100% related to displaying the info that the Deployment is not accessible - we may change our mind in displaying the connect button, which would not affect the other part of the PR, only this one - so IMHO it would be better to have this as a separate PR)

packages/renderer/src/lib/deployments/DeploymentEmptyScreen.svelte
```diff
@@ -1,7 +1,25 @@
...
+
+onMount(async () => {
+  disposable = await listenResourcePermitted('deployments', (permitted: boolean) => {
+    deploymentsPermitted = permitted;
+  });
+});
+
+onDestroy(() => {
+  disposable?.dispose();
+});
```
feloy: I think that we could simplify by putting this logic into `KubernetesEmptyScreen`, and passing the resource name (`deployments`) to the `KubernetesEmptyScreen` component (and probably replacing property `title` with properties `titleEmpty` and `titleNotPermitted`).

feloy:
LGTM
As you have changed the signature of `KubernetesEmptyScreen`, I suppose the simpler would be to apply the changes to all the `NodeEmptyScreen`, etc as part of this PR, at the changes will be very small in these pages

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
```diff
@@ -1,12 +1,29 @@
...
 import type { ComponentProps } from 'svelte';
+import { onDestroy, onMount } from 'svelte';
 
 import KubernetesCheckConnection from '/@/lib/ui/KubernetesCheckConnection.svelte';
+import type { IDisposable } from '/@api/disposable';
 
-let { icon, ...restProps }: ComponentProps<EmptyScreen> = $props();
+import { listenResourcePermitted } from '../kube/resource-permission';
+
+let { icon, resource, titleEmpty, titleNotPermitted, ...restProps }: ComponentProps<EmptyScreen> = $props();
```
feloy: It seems that this definition does not make resource, tileEmpty, ... required. I don't know if there would be another way to write this by enforcing these requirements

gastoner: What about something like 
```
interface Props extends ComponentProps<EmptyScreen> {
  resource: string;
  titleEmpty: string;
  titleNotPermitted: string;
}
```

gastoner:
> LGTM As you have changed the signature of `KubernetesEmptyScreen`, I suppose the simpler would be to apply the changes to all the `NodeEmptyScreen`, etc as part of this PR, at the changes will be very small in these pages

Also I see that e.g. in PodEmptyScreen.svelte we are not using the KubernetesEmptyScreen, but rather EmptyScreen, because we are passing some content as a child. So I would suggest to add another prop and place forward it from KubernetesEmptyScreen to EmptyScreen, so I could use KubernetesEmptyScreen everywhere

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
feloy: Yes, that seems to work, as you now get errors in other components (ServiceEmptyscreen, etc)

feloy:
> > LGTM As you have changed the signature of `KubernetesEmptyScreen`, I suppose the simpler would be to apply the changes to all the `NodeEmptyScreen`, etc as part of this PR, at the changes will be very small in these pages
> 
> Also I see that e.g. in PodEmptyScreen.svelte we are not using the KubernetesEmptyScreen, but rather EmptyScreen, because we are passing some content as a child. So I would suggest to add another prop and place forward it from KubernetesEmptyScreen to EmptyScreen, so I could use KubernetesEmptyScreen everywhere

There are 2 `PodEmptyScreen.svelte` , one for Podman Pods, the other one for Kubernetes pods. I can see that the Kuebrnetes one uses `KubernetesEmptyScreen`

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
feloy: By using this, you will need this fix: https://github.com/podman-desktop/podman-desktop/pull/11616

gastoner:
@feloy updated for all resources

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
```diff
@@ -1,12 +1,57 @@
...
+    } else {
+      resourcesTitle += titleNotPermitted[i];
+    }
+
+    // Adding new line
+    if (i + 1 < lengthOfResources) resourcesTitle += '\n';
+  }
+  return resourcesTitle;
+});
+
```
feloy: I don't think this is what we want.

For the version with no resource, I think that we don't want to change the behaviour, so we should still have 'No ingresses or routes', or nothing if there is at least one ingress or one route.

For permissions, I think that we want to display that we don't have permissions for **all** resources, or nothing. If the user has permission for at least one resource type, nothing should be marked.

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretEmptyScreen.svelte
```diff
@@ -6,7 +6,7 @@ import KubernetesEmptyScreen from '../kube/KubernetesEmptyScreen.svelte';
 <KubernetesEmptyScreen
   icon={ConfigMapSecretIcon}
   resources={['configmaps', 'secrets']}
-  titleEmpty={['No configmaps', 'No secrets']}
-  titleNotPermitted={['Configmaps not accessible', 'Secrets not accessible']}
+  titleEmpty='No configmaps or secrets'
+  titleNotPermitted='Configmaps or secrets not accessible'
```
feloy:
```suggestion
  titleNotPermitted='Configmaps and secrets not accessible'
```

same for others

packages/renderer/src/lib/kube/KubernetesEmptyScreen.svelte
```diff
@@ -1,12 +1,40 @@
...
+let { icon, resources, titleEmpty, titleNotPermitted, ...restProps }: Props = $props();
+
+let resourcePermitted: boolean = $state(true);
+let disposables: IDisposable[] = [];
+
+onMount(async () => {
+  for (let i = 0; i < resources?.length; i++) {
+    const disposable = await listenResourcePermitted(resources[i], (permitted: boolean) => {
+      resourcePermitted = resourcePermitted && permitted;
```
feloy: This will mark the set of resources as not permitted as soon as one is not permitted. But we want to mark the set as not permitted if all the resources are not permitted

Please add some unit tests for all the cases, so we can check on unit tests the result

feloy: (you would need to start from `resourcePermitted = false` and have an `||` here)

gastoner: Updated

gastoner:
@feloy updated

feloy (approved):

jeffmaury (approved):
LGTM

PR #11317, opened 2025-02-24
chore: kubernetes icon component

deboer-tim:
### What does this PR do?

Every Kubernetes resource kind has an associated icon, so when we know the kind why not use a component vs specifying the icon everywhere? This creates a new KubernetesIcon that can be used by the dashboard resource card, empty pages, or status columns in the future.

There is no easy way to test that a rendered svg matches, so I added a test id to every Kubernetes icon with the resource kind(s) that it represents.

### Screenshot / video of UI

N/A

Part of #11035.

### How to test this PR?


packages/renderer/src/lib/kube/KubernetesIcon.svelte
```diff
@@ -0,0 +1,51 @@
...
+import PvcIcon from '../images/PVCIcon.svelte';
+import ServiceIcon from '../images/ServiceIcon.svelte';
+
+interface Props {
+  kind: string;
+  size?: string;
+}
+let { kind, size }: Props = $props();
+
+let Icon = $derived.by(() => {
```
axel7083:
```suggestion
let Icon: Component = $derived.by(() => {
```

here, we have an implicit inferring, probably better to be explicit since we have a lot of options in the switch

deboer-tim: 👍🏼 Fixed.

benoitf (commented):
question: wondering the usage of data-testid vs `title` ?

packages/renderer/src/lib/kube/KubernetesIcon.svelte
```diff
@@ -0,0 +1,51 @@
...
+      return CronJobIcon as Component;
+    case 'Job':
+      return JobIcon as Component;
+    default:
+      return KubeIcon;
+  }
+});
+</script>
+
+<Icon size={size}/>
```
benoitf: looks like there is a missing newline

packages/renderer/src/lib/kube/KubernetesIcon.svelte
```diff
@@ -0,0 +1,51 @@
...
+import NodeIcon from '../images/NodeIcon.svelte';
+import PodIcon from '../images/PodIcon.svelte';
+import PvcIcon from '../images/PVCIcon.svelte';
+import ServiceIcon from '../images/ServiceIcon.svelte';
+
+interface Props {
+  kind: string;
+  size?: string;
+}
+let { kind, size }: Props = $props();
```
benoitf: issue: what is the default size if missing ? undefined

packages/renderer/src/lib/kube/KubernetesIcon.svelte
deboer-tim: Looking forward to biome svelte formatting. :) Fixed.

packages/renderer/src/lib/kube/KubernetesIcon.svelte
deboer-tim: The underlying icons all have a default size, so this is fine - except that several have _different_ defaults. Added a default size (the same as what most icons are using) so that it is consistent.

benoitf (commented):
still the question of datatest-id vs giving title property to the svg

deboer-tim:
> question: wondering the usage of data-testid vs `title` ?

The title property isn't allowed on svg elements. We could use a title element instead, but this felt like a bigger change plus I wasn't sure we should directly use the Kubernetes kind for accessibility. Other attributes like aria-label normally only have 1 value so it wasn't as clear how you would say 'this is icon for Ingress and Route'.

benoitf:
I don't know I would avoid any `data-testid` in the Podman Desktop codebase

To me if there is an information it's useful for everyone, not only for passing a unit test

so either a `<title>` or an aria-label (in any case the img might be non-visible by default so it's not an issue then)

deboer-tim:
> I don't know I would avoid any `data-testid` in the Podman Desktop codebase
> 
> To me if there is an information it's useful for everyone, not only for passing a unit test
> 
> so either a `<title>` or an aria-label (in any case the img might be non-visible by default so it's not an issue then)

Updated, went with aria-label since it is a smaller change and we use it more often.

packages/renderer/src/lib/kube/KubernetesIcon.svelte
```diff
@@ -0,0 +1,51 @@
...
+import ConfigMapSecretIcon from '../images/ConfigMapSecretIcon.svelte';
+import CronJobIcon from '../images/CronJobIcon.svelte';
+import DeploymentIcon from '../images/DeploymentIcon.svelte';
+import IngressRouteIcon from '../images/IngressRouteIcon.svelte';
+import JobIcon from '../images/JobIcon.svelte';
+import KubeIcon from '../images/KubeIcon.svelte';
+import NodeIcon from '../images/NodeIcon.svelte';
+import PodIcon from '../images/PodIcon.svelte';
+import PvcIcon from '../images/PVCIcon.svelte';
+import ServiceIcon from '../images/ServiceIcon.svelte';
```
benoitf: nit: as other PR, prefer the `/@...` import rather than ../ (in case of moving this file to another folder, imports will remain the same)

benoitf (approved):
LGTM

packages/renderer/src/lib/kube/KubernetesIcon.svelte
deboer-tim: Yup, I was just doing copy/paste without cleanup. Fixed.

jeffmaury (approved):
LGTM

PR #11318, opened 2025-02-24
fix: pods should have kubernetes empty screen

### LINKED ISSUE #11231: Kubernetes Pod Empty Screen has podman related actions
### Bug description

inside the Kubernetes pods page, when you don't have any pods, the EmptyScreen is the following

Image

`Start your first pod` will create a podman pod, and the command bellow is to create a podman pod.

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. Go to kubernetes page
2. Open pods page
3. assert no pods running
4. assert wrong empty screen visible

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

The Kubernetes Pods screen shares its 'empty screen' with the regular Pods page, but it should have the same empty screen as all the other Kubernetes pages. This just aligns it with what every other Kube page does.

### Screenshot / video of UI

Screenshot 2025-02-24 at 10 19 42 AM

Fixes #11231.

### How to test this PR?

Go to the Kubernetes > Pods page when there are no pods (e.g. disconnected/stopped context).

### DIFF
--- a/packages/renderer/src/lib/kube/pods/PodEmptyScreen.svelte
+++ b/packages/renderer/src/lib/kube/pods/PodEmptyScreen.svelte
@@ -1,0 +1,9 @@
+<script lang="ts">
+import PodIcon from '../../images/PodIcon.svelte';
+import KubernetesEmptyScreen from '../KubernetesEmptyScreen.svelte';
+</script>
+
+<KubernetesEmptyScreen
+  icon={PodIcon}
+  title="No pods"
+  message="Try switching to a different context or namespace" />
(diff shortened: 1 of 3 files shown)


axel7083 (approved):

packages/renderer/src/lib/kube/pods/PodEmptyScreen.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+import PodIcon from '../../images/PodIcon.svelte';
+import KubernetesEmptyScreen from '../KubernetesEmptyScreen.svelte';
```
benoitf: nit: `/@..` imports

benoitf (approved):

packages/renderer/src/lib/kube/pods/PodEmptyScreen.svelte
deboer-tim: I don't love that this is different from all 7 other Kubernetes empty pages 🙈, but fixed.

benoitf: I have not looked on how to enforce that rule of using prefix scoping rather than relative scope

PR #11327, opened 2025-02-24
chore: port forwarding should open kubernetes pod page

deboer-tim:
### What does this PR do?

The port forwarding feature was relying on the list of pods from the Pods store, (which used the Kubernetes watcher), then opened the regular Pods page.

The watcher could be out of sync/delayed; we should be using the Kubernetes informer and opening the new Kubernetes > Pods page instead.

@feloy I'm switching this from the old store to the new one. I don't know if this was on your list to support experimental as well or we just leave it for now, just making sure you're aware.

### Screenshot / video of UI

N/A

Part of #8819.

### How to test this PR?

Create a port forward on a Kubernetes pod, then go to the Port Forwarding page and ensure it opens the Kubernetes > Pods > Details page when you click on it.


jeffmaury (approved):
LGTM

axel7083 (approved):

PR #11341, opened 2025-02-25
feat: display Ingress details in experimental mode

feloy:

### What does this PR do?

Display Ingress details in experimental mode

Part of #10657 

### How to test this PR?

- create ingress (1)
- go to Kubernetes > Ingresses/Routes > Details
- check that the summary of ingress is displayed
- check that the ingress can be updated from the Kube tab (add annotation, etc)

(1) example of ingress

```
kind: Ingress
metadata:
  name: ingress1
  namespace: default
spec:
  rules:
  - host: localhost
```

### DIFF
--- a/packages/renderer/src/lib/ingresses-routes/IngressDetails.spec.ts
+++ b/packages/renderer/src/lib/ingresses-routes/IngressDetails.spec.ts
@@ -20,18 +20,23 @@ import '@testing-library/jest-dom/vitest';
 
 import type { KubernetesObject, V1Ingress } from '@kubernetes/client-node';
 import { fireEvent, render, screen } from '@testing-library/svelte';
-import { writable } from 'svelte/store';
 import { router } from 'tinro';
-import { beforeAll, expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 
+import { isKubernetesExperimentalMode } from '/@/lib/kube/resources-listen';
+import {
+  initListExperimental,
+  initListsNonExperimental,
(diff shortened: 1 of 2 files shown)


benoitf (approved):

deboer-tim (commented):
The first few of these were merged before I had a chance to ask, but did you check if there's a way to abstract or extract some of the logic roughly similar to what you did for KubernetesObjectsList? I'm worried that this will create 8 more files that we will just need to edit again in a couple months if experimental mode works well. Ideally these files would only contain the display logic, 'getting the data' should be as common as possible.

feloy:
> The first few of these were merged before I had a chance to ask, but did you check if there's a way to abstract or extract some of the logic roughly similar to what you did for KubernetesObjectsList? I'm worried that this will create 8 more files that we will just need to edit again in a couple months if experimental mode works well. Ideally these files would only contain the display logic, 'getting the data' should be as common as possible.



> The first few of these were merged before I had a chance to ask, but did you check if there's a way to abstract or extract some of the logic roughly similar to what you did for KubernetesObjectsList? I'm worried that this will create 8 more files that we will just need to edit again in a couple months if experimental mode works well. Ideally these files would only contain the display logic, 'getting the data' should be as common as possible.

I tried my best to move all the logic inside `listenResource`. There is still some logic around the `loadDetails` which may have been moved inside `listenResource` too. I remember trying to move the logic to `listenResource`, but finally chosen to keep it in the components (but don't remember for which reason)

deboer-tim (approved):

PR #11349, opened 2025-02-25
fix: display of route details 

### LINKED ISSUE #11346: route details are not displayed
### Bug description

When entering the details page for a route, the page is empty

https://github.com/user-attachments/assets/9ef94533-565e-4020-8dc2-43bd76e2e3d8

### Operating system

macOS

### Installation Method

### Version

next (development version)

### Steps to reproduce

see video above

### Relevant log output

```shell

```

### Additional context

The regression has been introduced when updating the kubernetes-client to v1.x and when changing pattern of urls for Kubernetes resources

feloy:
### What does this PR do?

fix: display of route details 

Fixes #11346 

### How to test this PR?

Connect to an openshift cluster with a route defined, and go to the details page for the route

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.ts
@@ -952,6 +952,9 @@ export class KubernetesClient {
       if (route?.metadata?.managedFields) {
         delete route.metadata.managedFields;
       }
+      if (Object.keys(route).length === 0) {
+        return undefined;
+      }
       return route;
     } catch (error) {
       this.telemetry.track('kubernetesReadNamespacedRoute.error', error);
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/kubernetes/kubernetes-client.ts
```diff
@@ -948,7 +948,7 @@ export class KubernetesClient {
         plural: 'routes',
         name,
       });
-      const route = res?.body as V1Route;
+      const route = res as V1Route;
```
benoitf, line 951: question: how do we check this through a unit test ?

feloy: This would depend on what the library returns, meaning that we would need some kind of integration tests, with at least the library

benoitf (approved):

packages/main/src/plugin/kubernetes/kubernetes-client.ts
deboer-tim: Can be tracked through new issue and it's probably an e2e test since it would require a cluster of some kind, but sounds like we should have a test that fails when we pick up a new client that changes the expected output? Otherwise we're waiting until someone find an issue in the UI later.

deboer-tim (approved):

PR #11352, opened 2025-02-25
chore: remove podman pod kubernetes navigation

deboer-tim:
### What does this PR do?

Since you can no longer have Kubernetes pods in the regular Pods page, you can't navigate to any pod kind other than 'podman'. The kind parameter in the navigation can be removed since it can never have another value.

This affects the extension API, but that is what we want as Kubernetes navigation should be done using the route and not part of the extension API. As far as I can tell the main extensions we have today are not using this yet, so no breakage.

The hard-coded 'podman' in the call to containerRegistry.podExist() and in the navigation.ts url is a bit odd, but: one change, one PR. These could be cleaned up in future PRs.

### Screenshot / video of UI

N/A

Part of #8819.

### How to test this PR?

PR checks (just cleanup, existing tests modified).

### DIFF
--- a/packages/api/src/navigation-request.ts
+++ b/packages/api/src/navigation-request.ts
@@ -36,7 +36,7 @@ export interface NavigationParameters {
   [NavigationPage.IMAGE]: { id: string; engineId: string; tag: string };
   [NavigationPage.ONBOARDING]: { extensionId: string };
   [NavigationPage.PODMAN_PODS]: never;
-  [NavigationPage.PODMAN_POD]: { kind: string; name: string; engineId: string };
+  [NavigationPage.PODMAN_POD]: { name: string; engineId: string };
   [NavigationPage.VOLUMES]: never;
   [NavigationPage.VOLUME]: { name: string };
   [NavigationPage.CONTRIBUTION]: { name: string };
--- a/packages/main/src/plugin/extension/extension-loader.spec.ts
+++ b/packages/main/src/plugin/extension/extension-loader.spec.ts
@@ -1445,7 +1445,6 @@ describe('Navigation', async () => {
     expect(sendMock).toBeCalledWith('navigate', {
       page: NavigationPage.PODMAN_POD,
       parameters: {
-        kind: 'valid-kind',
         name: 'valid-name',
         engineId: 'valid-engine',
       },
--- a/packages/main/src/plugin/extension/extension-loader.ts
+++ b/packages/main/src/plugin/extension/extension-loader.ts
@@ -1397,6 +1397,9 @@ export class ExtensionLoader {
         await this.navigationManager.navigateToPods();
       },
       navigateToPod: async (kind: string, name: string, engineId: string): Promise<void> => {
+        if (kind === 'kubernetes') {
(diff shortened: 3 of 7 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -4784,8 +4784,8 @@ declare module '@podman-desktop/api' {
 
     // Navigate to the Pods page
     export function navigateToPods(): Promise<void>;
-    // Navigate to a specific pod referenced by kind, name and engineId
-    export function navigateToPod(kind: string, name: string, engineId: string): Promise<void>;
+    // Navigate to a specific pod referenced by name and engineId
+    export function navigateToPod(name: string, engineId: string): Promise<void>;
```
axel7083: I don't know if we can just break changes like that, I feel like we should maybe introduce a new method, deprecate this one for a version, then remove it ?

AI Lab is using `navigateToPod` so it need to be updated, we might need to have some issue for that?

Here is a suggestion

```suggestion
    /**
     * Navigate to a specific pod referenced by kind, name and engineId
     * @deprecated will be removed in Podman Desktop 1.18 use {@link navigateToPodmanPod} instead
     * @param kind
     * @param name
     * @param engineId
     */
    export function navigateToPod(kind: string, name: string, engineId: string): Promise<void>;
    export function navigateToPodmanPod(name: string, engineId: string): Promise<void>;
```

deboer-tim: I was searching for the wrong thing :-/. I added a new commit that implements your suggestion, allowing AI Lab or other extensions time to migrate. The deprecated support will only work for Podman pods, but that's all AI Lab is using. I will open an issue to track the migration once we confirm this direction.

axel7083: That's good on my side !

benoitf (changes requested):
unfortunately, there is no way we can remove something in extension-api.d.ts (saying it'll be removed)

deboer-tim:
> unfortunately, there is no way we can remove something in extension-api.d.ts (saying it'll be removed)

Sorry, not sure which you mean:
- The function needs to remain indefinitely (i.e. navigateToPod should now redirect to Pods > Details or Kubernetes > Pods > Details depending on what's passed in).
- It's ok to 'break' function for Kubernetes pod, but just leave current function working only for Podman pods.
- We can mark navigateToPod as deprecated to encourage people to use the new / cleaner function, but should never actually remove it.

benoitf:
>Sorry, not sure which you mean:

you added in the .d.ts file `* @deprecated will be removed in Podman Desktop 1.18 use {@link navigateToPodmanPod} instead`

but we can't remove methods like that. it requires backward compliance. so the method should always do something

deboer-tim:
Even after the discussion today (thank you!), I'm having second thoughts about renaming / new extension API. So, I decided to bring it back to basics: one change, one PR.

I've rebased and pushed a new commit that _only_ adds the error if you use the current extension API to try to navigate to a Kubernetes pod, and removes the unused parameter from the internal navigation call. Hopefully this is easier to review, and we can always have a second PR to improve the external API.

benoitf (approved):

Issue #11363, opened 2025-02-25
Dashboard component


deboer-tim:
### Is your enhancement related to a problem? Please describe

The Kubernetes dashboard has an updated design, and we agreed to use this design for other dashboards going forward, including updating the bootc and AI Lab extension dashboards.

When trying to work on https://github.com/podman-desktop/extension-bootc/issues/968 I found there is a lot of copy/paste to implement the same look, since there is no component.

### Describe the solution you'd like

Create a Dashboard component in the UI library (for new dashboards and extensions to use), and port the current Kubernetes dashboard to use it.

PR #11364, opened 2025-02-25
feat: expandable component

deboer-tim:
### What does this PR do?

While working on https://github.com/podman-desktop/extension-bootc/issues/976 and investigating #11363 I noticed that we have at least 3 expandable sections in Podman Desktop (one in the Learning Center and 2 on the Kubernetes Dashboard), and 3 in the bootc extension.

All of these have one minor bug in common (text moves right/left as you expand/collapse). The ones in bootc aren't animated; the ones in Podman Desktop are but use two different animation speeds, and all three of these have glitchy animation (they jump the last bit) because the transition is on the wrong element.

Instead of copying fixes, this creates an expandable component with the basic function of these instances.

Expanded+initialized could be a single property, but since most users will just want a simple boolean expanded I kept them separate. Storybook kept simple to start; I had issues using Svelte 5 and light/dark.

### Screenshot / video of UI

Next PR would use the component on the dashboard. Note the title doesn't shift and the last bit of the animation is now smooth.

Before:

https://github.com/user-attachments/assets/6e4875b8-503c-4afc-941b-d57afc73b17a

After:

https://github.com/user-attachments/assets/8521598a-08b4-4287-bec2-c192eb68bd39

Related to #11363.
Related to https://github.com/podman-desktop/extension-bootc/issues/976.

### How to test this PR?

Test in Storybook to check title/animation, next PR will be in context.


packages/ui/src/lib/button/Expandable.svelte
```diff
@@ -0,0 +1,46 @@
...
+        <Fa class="w-4" icon={faChevronDown} />
+      {:else}
+        <Fa class="w-4" icon={faChevronRight} />
+      {/if}
+      {@render title?.()}
+    </div>
+  </button>
+{#if initialized}
+{#if expanded}
+  <div role="region" class="pt-5" transition:fadeSlide={{ duration: 250 }}>
```
benoitf: question: why do we need a new custom function vs

`import { fade, slide } from "svelte/transition";`
`transition:fade|slide`

deboer-tim: I hadn't touched the transition, I assume it was there originally because Svelte didn't support merged/composite transitions. Using this syntax doesn't work (linting says slide is unused, didn't work at runtime, can't find it in the docs) so I assume that's still the case in Svelte 5.

That said, the other way people get around the limitation is just nesting two transitions, so I did that instead - no more utility animation for same effect. 👍🏼

benoitf: thanks for the explanation

packages/ui/src/lib/button/Expandable.svelte
```diff
@@ -0,0 +1,47 @@
...
+{#if initialized}
+{#if expanded}
+  <div role="region" class="pt-5" transition:slide={{ duration: 250 }}>
+    <div transition:fade={{ duration: 250 }}>
+      {@render children?.()}
+    </div>
+  </div>
+{/if}
+{/if}
+</div>
```
benoitf: newline

benoitf (approved):

packages/ui/src/lib/button/Expandable.svelte
deboer-tim: Fixed. :)

ekidneyrh (approved):
Looks great!

packages/ui/src/lib/button/Expandable.svelte
```diff
@@ -0,0 +1,47 @@
...
+  title = undefined,
+  children = undefined,
+  onclick,
+}: {
+  expanded?: boolean;
+  initialized?: boolean;
+  title?: Snippet;
+  children?: Snippet;
+  onclick?: (expanded: boolean) => void;
+} = $props();
```
axel7083:
```suggestion
interface Props {
  expanded?: boolean;
  initialized?: boolean;
  title?: Snippet;
  children?: Snippet;
  onclick?: (expanded: boolean) => void;
}

let {
  expanded = $bindable(true),
  initialized = $bindable(true),
  title = undefined,
  children = undefined,
  onclick,
}: Props = $props();
```

deboer-tim: Fixed.

axel7083 (approved):
LGTM as per https://67bf2b1e03b53f91d4e2d29d--podman-desktop-pr.netlify.app/storybook?id=expandable--docs

PR #11380, opened 2025-02-26
docs(website): add SELinux info to GPU guide

cyqsimon:
### Screenshot / video of UI

N/A. This is a documentation change.

N/A.

### How to test this PR?

Try the original procedure on Fedora. You will run into "Failed to initialize NVML: Insufficient Permissions" due to SELinux issues.

## Summary by Sourcery

Documentation:
- Add instructions for configuring SELinux on SELinux-enabled operating systems like Fedora to allow containers to access the GPU.

### DIFF
--- a/website/docs/podman/gpu.md
+++ b/website/docs/podman/gpu.md
@@ -191,7 +191,9 @@ Important note that the virtualized GPU (Virtio-GPU Venus (Apple M1 Pro)) only s
 #### Procedure
 
 1. Install the latest NVIDIA GPU Driver for your OS.
+
 2. Follow the instructions on [installing the NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html) in relation to your Linux distribution.
+
 3. Generate the CDI Specification file for Podman:
 
 This file is saved either to /etc/cdi or /var/run/cdi on your Linux distribution and is used for Podman to detect your GPU(s).
@@ -209,6 +211,21 @@ $ nvidia-ctk cdi list
 ```
 
 More information as well as troubleshooting tips can be found [on the official NVIDIA CDI guide](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html).
+
+4. If on a SELinux-enabled OS (e.g. Fedora family), configure SELinux policy:
+
+Check whether SELinux is installed and enabled:
+
+```sh
+$ getenforce
+```
+
+- If `getenforce` is not found, or its output is `Permissive` or `Disabled`, no action is need.
+- If the output is `Enforcing`, configure SELinux to allow containers device access:
+
+```sh
+$ sudo setsebool -P container_use_devices true
+```
 
 #### Verification
 
--- a/website/docs/podman/gpu.md
+++ b/website/docs/podman/gpu.md
@@ -212,7 +212,9 @@ $ nvidia-ctk cdi list
 
 More information as well as troubleshooting tips can be found [on the official NVIDIA CDI guide](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html).
 
-4. If on a SELinux-enabled OS (e.g. Fedora family), configure SELinux policy:
+4. Configure SELinux (if applicable)
+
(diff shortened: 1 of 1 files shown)


benoitf (approved):

benoitf:
thanks @cyqsimon

website/docs/podman/gpu.md
```diff
@@ -210,6 +212,21 @@ $ nvidia-ctk cdi list
 
 More information as well as troubleshooting tips can be found [on the official NVIDIA CDI guide](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html).
 
+4. If on a SELinux-enabled OS (e.g. Fedora family), configure SELinux policy:
```
shipsing:
```suggestion
4. Optional: Configure the SELinux policy on a SELinux-enabled OS, such as OSs from the Fedora family:
```

website/docs/podman/gpu.md
```diff
@@ -210,6 +212,21 @@ $ nvidia-ctk cdi list
...
+
+Check whether SELinux is installed and enabled:
+
+```sh
+$ getenforce
+```
+
+- If `getenforce` is not found, or its output is `Permissive` or `Disabled`, no action is need.
+- If the output is `Enforcing`, configure SELinux to allow containers device access:
```
shipsing:
```suggestion
- If `getenforce` is not found or its output is `Permissive` or `Disabled`, no action is needed.
- If the output is `Enforcing`, configure SELinux to enable access to container devices:
```

website/docs/podman/gpu.md
```diff
@@ -210,6 +212,21 @@ $ nvidia-ctk cdi list
...
+
+```sh
+$ sudo setsebool -P container_use_devices true
+```
```
shipsing, line 230: We can indent this command to align with the previous bullet point to which it applies.

shipsing (commented):
Provided some suggestions. Thanks for your contribution.

website/docs/podman/gpu.md
cyqsimon: I did not want to use the "optional" wording because it can potentially convey the wrong message. The user may see this as "something that will improve the experience but not strictly necessary", but that's not the case. If you don't do this on SELinux-enabled systems, your GPU won't work.

shipsing: Thanks for the clarification. I mentioned Optional because if the user is not on any SELinux-enabled OS, then this step is not necessary for them. But, i agree it might convey a wrong message. What about Conditional, if the user is on a SELinux-enabled OS?

shipsing: 4. Conditional: On a SELinux-enabled OS, such as OSs from the Fedora family, configure the SELinux policy:

website/docs/podman/gpu.md
cyqsimon:
```
configure SELinux to enable access to container devices:
```

This is wrong. Maybe this?

```suggestion
- If `getenforce` is not found or its output is `Permissive` or `Disabled`, no action is needed.
- If the output is `Enforcing`, configure SELinux to enable device access for containers:
```

website/docs/podman/gpu.md
cyqsimon: Yeah sure that'd be nice. But I'm not sure how to do this with the specific renderer we're using.

cyqsimon: Github's renderer for example won't let you do this sort of thing. Unless there's some arcane trick I'm not aware of.

- List item 0
- List item 1
- ```sh
some code here
```

Comment (shipsing) [reply website/docs/podman/gpu.md:230]: No worries, i will handle it once the PR gets merged.

File: website/docs/podman/gpu.md (hunk 1 above)
Comment (cyqsimon) [reply website/docs/podman/gpu.md]: Perhaps it's best to keep the heading simple and talk about the conditional stuff on the first line.

```suggestion
4. Configure SELinux (if applicable)

On SELinux-enabled OSes, such as OSs from the Fedora family, the default policy usually disallows containers to have direct access to devices. We make sure it's allowed.
```

cyqsimon:
Um, can we disregard the semantic requirement for the review commit and just squash merge?

benoitf:
@cyqsimon yes

PR #11383, opened 2025-02-26
fix: ensure that chaining calls to showQuickPick is working

### LINKED ISSUE #11250: issue when calling multiple times showQuickPick/showInputBox
### Bug description

it seems that when we want to chain calls to showQuickPick then showInputBox, etc
we don't see new calls being displayed

### Operating system

macOS

### Installation Method

None

### Version

1.14.1

### Steps to reproduce

call showQuickPick then call showInputBox when receiving the first reponse

### Relevant log output

```shell

```

### Additional context

_No response_

benoitf:
### What does this PR do?
Ensure that we can call showQuickPick/showInputBox as soon as we receive the result of a previous call to showQuickPick

issue was like cleanup was occurring after we sent the result, meaning that we were receiving an order to display items then turning display to true) and then cleanup is occurring, setting display to false so we don't see the quickpick menu

fixes https://github.com/podman-desktop/podman-desktop/issues/11250

### How to test this PR?

unit test is provided

## Summary by Sourcery

Ensures that chaining calls to `showQuickPick` and `showInputBox` work correctly by adjusting the timing of the cleanup process.

Bug Fixes:
- Fixes an issue where subsequent calls to `showQuickPick` or `showInputBox` immediately after a previous `showQuickPick` call would fail due to premature cleanup. The cleanup is now performed after returning the value.

Tests:
- Adds a unit test to verify that two quick picks can be displayed in a row.

### DIFF
--- a/packages/renderer/src/lib/dialogs/QuickPickInput.spec.ts
+++ b/packages/renderer/src/lib/dialogs/QuickPickInput.spec.ts
@@ -400,4 +400,63 @@ describe('QuickPickInput', () => {
     const item1 = getByTitle('Select item1');
     expect(item1).toBeDefined();
   });
+
+  test('Expect to see two quickpick in a row', async () => {
+    // first quickpick being displayed
+    const idRequest = 123;
+    const quickPickOptions: QuickPickOptions = {
+      items: ['item1', 'item2'],
+      canPickMany: false,
+      placeHolder: 'placeHolder',
+      prompt: '',
+      id: idRequest,
+      onSelectCallback: false,
+    };
+
+    const quickPickOptions2: QuickPickOptions = {
+      items: ['foo', 'bar'],
+      canPickMany: false,
+      placeHolder: 'placeHolder',
+      prompt: 'enter a custom value',
+      id: idRequest,
+      onSelectCallback: false,
+    };
+
+    let eventCallback: ((options: QuickPickOptions) => void) | undefined;
+
+    receiveFunctionMock.mockImplementation((message: string, callback: (options: QuickPickOptions) => void) => {
+      if (message === 'showQuickPick:add') {
+        eventCallback = callback;
+      }
+    });
+
+    // when getting the response of the first quickpick, we ask to display the second quickpick
+    sendShowQuickPickValuesMock.mockImplementation(() => {
+      eventCallback?.(quickPickOptions2);
+    });
+
+    // render the first quickpick
+    const { getByTitle } = render(QuickPickInput, {});
+
+    // wait that the event callback is set
+    await vi.waitFor(() => eventCallback !== undefined);
+
+    // ok ask to display a first quickpick
+    expect(eventCallback).toBeDefined();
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/dialogs/QuickPickInput.spec.ts
```diff
@@ -400,4 +400,64 @@ describe('QuickPickInput', () => {
...
+    // wait that the event callback is set
+    await vi.waitFor(() => eventCallback !== undefined);
+
+    // ok ask to display a first quickpick
+    expect(eventCallback).toBeDefined();
+    eventCallback?.(quickPickOptions);
+
+    // check the first quick pick options is displayed
+    await vi.waitFor(() => getByTitle('Select item1'));
+    const item1 = getByTitle('Select item1');
```
axel7083:
```suggestion
    const item1 = await vi.waitFor(() => getByTitle('Select item1'));
```

We can simplify

packages/renderer/src/lib/dialogs/QuickPickInput.spec.ts
```diff
@@ -400,4 +400,63 @@ describe('QuickPickInput', () => {
...
+      onSelectCallback: false,
+    };
+
+    let eventCallback: ((options: QuickPickOptions) => void) | undefined;
+
+    receiveFunctionMock.mockImplementation((message: string, callback: (options: QuickPickOptions) => void) => {
+      if (message === 'showQuickPick:add') {
+        eventCallback = callback;
+      }
+    });
```
axel7083, line 431: I feel like we should avoid mocking implementation to get an argument when we can do something more "vitest-ish" 

For example with the following we can get the callback later in the code without mocking the implementation, and without declaring a possible undefined variable

```ts
const eventCallback = vi.waitFor<((options: QuickPickOptions) => void)>(() => {
  // ensure the function has been called with the expect message
  expect(receiveFunctionMock).toHaveBeenCalledWith('showQuickPick:add', expect.any(Function));
  
  // extract the call 
  const call = vi.mocked(receiveFunctionMock).mock.calls.find(([message]) => message === 'showQuickPick:add');
  // typescript sugar to avoid error
  if(!call) throw new Error('cannot find listener for message showQuickPick:add');
  // return the callback
  return call[1]
});
```

axel7083 (commented):
I am not sure to understand what is the purpose of the cleanup function here

packages/renderer/src/lib/dialogs/QuickPickInput.spec.ts
benoitf: I found several issues in the test file but I want to separate cleanup and providing the fix as it's using the same content as the previous test of the file

axel7083: I am okay if this is the way it was already done in the previous test (was not seeing that with the GitHub review)

axel7083 (approved):
LGTM

Issue #16, opened 2022-03-11
Build Dockerfiles with any name


deboer-tim:
**Is your feature request related to a problem? Please describe.**
Most dockerfiles have the default name of Dockerfile, but some don't. Need to add support for building files like build.Dockerfile or myproject.Dockerfile.

**Describe the solution you'd like**
The current UI and defaulting to simple case is good, it just needs a way to override and select a specific file.

rhatdan:
This tool should default to Containerfile as well. Please don't use the Docker adjective unless absolutely necessary.

PR #11388, opened 2025-02-26
chore(test): skip kind tests that run on rootless mode machines

danivilla9:
### What does this PR do?
Adds an skip condition to the kind tests so that they are not run on rootless machines. Fixes partially [#315 ](https://github.com/podman-desktop/e2e/issues/315).

[Kind E2E tests on windows rootless should be expected to fail, but aren't #315 ](https://github.com/podman-desktop/e2e/issues/315)
Related PR: [chore(e2e): add ROOTFUL_MODE envvar for kind tests #16](https://github.com/odockal/pde2e-runner/pull/16)

### How to test this PR?
Run `pnpm test:e2e:all`

## Summary by Sourcery

Skip Kind E2E tests on rootless machines. This change ensures that Kind E2E tests are only executed on rootful machines by adding a skip condition based on the `ROOTFUL_MODE` environment variable.

### DIFF
--- a/tests/playwright/src/setupFiles/setup-kind.ts
+++ b/tests/playwright/src/setupFiles/setup-kind.ts
@@ -1,0 +1,33 @@
+/**********************************************************************
+ * Copyright (C) 2024-2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { isLinux, isMac, isWindows } from '../utility/platform';
+
+export function setupKind(): boolean[] {
+  const rootfulMode = process.env.ROOTFUL_MODE === 'true'; // undefined or "false" => false
+  const runKindTests = process.env.RUN_KIND_TESTS === 'true';
+  return [rootfulMode, runKindTests];
+}
+
+export function canRunKindTests(): boolean {
+  const [rootfulMode, runKindTests] = setupKind();
+  // to run the kind tests there are 2 options:
+  // 1-the OS is linux or mac + runKindTests == true
+  // 2-the OS is windows, the podman machine is rootful
+  return ((isLinux || isMac) && runKindTests) || (isWindows && rootfulMode);
+}
--- a/tests/playwright/src/setupFiles/setup-kind.ts
+++ b/tests/playwright/src/setupFiles/setup-kind.ts
@@ -26,8 +26,9 @@ export function setupKind(): boolean[] {
 
 export function canRunKindTests(): boolean {
   const [rootfulMode, runKindTests] = setupKind();
-  // to run the kind tests there are 2 options:
-  // 1-the OS is linux or mac + runKindTests == true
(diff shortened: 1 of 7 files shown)


benoitf (approved):

cbr7 (approved):

amisskii (changes requested):
There are other test cases that use the Kind cluster: kubernetes.spec.ts, kubernetes-deployments.spec.ts, kubernetes-port-forwarding.spec.ts, and so on. Please take a look at these test scenarios and add a skip condition for them if needed.

odockal (changes requested):
Right now, we are working with this env. var. on windows workflows. We should stick to this, since we can run these tests on GHA Ubuntu when we use docker engine. I am also curious about behavior on mac.

cdrage (approved):

cdrage (approved):
Tests passed fine for me, so giving approval, but there are some other comments on here for fixing.

amisskii (changes requested):
I think we don’t use a Kind cluster in the` kubernetes-context-smoke` test; instead, we import a prepared custom kubeconfig. Therefore, please remove the skip condition from it

danivilla9:
> I think we don’t use a Kind cluster in the` kubernetes-context-smoke` test; instead, we import a prepared custom kubeconfig. Therefore, please remove the skip condition from it

Removed it - Thanks for notifying it

tests/playwright/src/specs/kind.spec.ts
```diff
@@ -54,7 +55,9 @@ test.beforeAll(async ({ runner, page, welcomePage }) => {
 
 test.afterAll(async ({ runner, page }) => {
   try {
-    await deleteCluster(page, RESOURCE_NAME, KIND_NODE, CLUSTER_NAME);
+    if (!canRunKindTests)
```
odockal: If you place `test.skip` outside the describe part, the afterall should not be called at all. Like here: https://github.com/podman-desktop/podman-desktop/blob/main/tests/playwright/src/specs/compose-onboarding-smoke.spec.ts#L39

you can do it for all specs, since the condition skips all tests.

amisskii (approved):
LGTM. Although the lint check failed, running `pnpm lint:check --fix should` resolve it.

amisskii (approved):
LGTM. Although the lint check failed, running `pnpm lint:check --fix `should resolve it.

tests/playwright/src/setupFiles/setup-kind.ts
```diff
@@ -0,0 +1,33 @@
...
+  const runKindTests = process.env.RUN_KIND_TESTS === 'true';
+  return [rootfulMode, runKindTests];
+}
+
+export function canRunKindTests(): boolean {
+  const [rootfulMode, runKindTests] = setupKind();
+  // to run the kind tests there are 2 options:
+  // 1-the OS is linux or mac + runKindTests == true
+  // 2-the OS is windows, the podman machine is rootful
+  return ((isLinux || isMac) && runKindTests) || (isWindows && rootfulMode);
```
odockal: I would change the evaluation when to run kind tests.

One restriction is that it does not run on windows with rootless, right?
The other one is that we may have them running on linux or mac...
The last one is that if there is a RUN_KIND_TESTS==true, we want to run them.

So from my perspective: 
`return runKindTests || (isMac || isLinux) || (isWindows && rootfulMode)`.
This way, we have env. var which will trigger kind tests in any case (in cases when we want to run kind with preconfigured rootless - which is possible), then we want to run them on mac o on linux, and finally, we will run tests on windows rootfull.

odockal (changes requested):

odockal (approved):
LGTM.

cbr7:
I see that this skips the tests only on windows rootless podman machines, but what should happen if there is a rootless podman machine on a mac?

danivilla9:
> I see that this skips the tests only on windows rootless podman machines, but what should happen if there is a rootless podman machine on a mac?

I've just tried to create a kind cluster and push an image to it on a mac with a rootless podman (5.5) machine and both operations worked, so I guess that's possible now?

PR #11390, opened 2025-02-26
feat: add a command to configure registries of Podman

### LINKED ISSUE #10673: Implementation mirror registries configuration

benoitf:
### What does this PR do?
Add a command to be able to configure podman registries

### Screenshot / video of UI

https://github.com/user-attachments/assets/151b7f6d-0366-4075-b549-3e83d02eed45

fixes https://github.com/podman-desktop/podman-desktop/issues/10673

### How to test this PR?

1. recreate the machine assuming you're on macOS (on Windows it seems you need to manually tweak the mount)
1. bring command palette and select "podman registry setup" command as in the recording

then enter the registry and then the mirror and check the command using the cli inside the VM to get the right answer

```
podman pull docker.io/invalidddd
Trying to pull docker.io/library/invalidddd:latest...
Error: initializing source docker://invalidddd:latest: (Mirrors also failed: [ghcr.io/library/invalidddd:latest: Requesting bearer token: invalid status code from registry 403 (Forbidden)]): docker.io/library/invalidddd:latest: reading manifest latest in docker.io/library/invalidddd: requested access to the resource is denied
```

see it tries a mirror

## Summary by Sourcery

Adds a new command `podman.setupRegistry` to configure registries for Podman. The command allows users to add registry locations and configure mirrors. The configuration is persisted in a TOML file.

New Features:
- Adds a command to configure Podman registries, allowing users to add registry locations and configure mirrors via a quick pick interface.
- The registry configuration is persisted in a TOML file.

### DIFF
--- a/extensions/podman/packages/extension/src/configuration/registry-configuration.ts
+++ b/extensions/podman/packages/extension/src/configuration/registry-configuration.ts
@@ -16,18 +16,52 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { mkdir, writeFile } from 'node:fs/promises';
+import { existsSync } from 'node:fs';
+import { mkdir, readFile, writeFile } from 'node:fs/promises';
 import { homedir, tmpdir } from 'node:os';
 import { dirname, resolve } from 'node:path';
 
-import { env } from '@podman-desktop/api';
+import type { Disposable, QuickPickItem } from '@podman-desktop/api';
+import { commands, env, window } from '@podman-desktop/api';
 import mustache from 'mustache';
+import * as toml from 'smol-toml';
 
+import { getJSONMachineList } from '../extension';
+import { execPodman } from '../util';
 import playbookRegistryConfFileTemplate from './playbook-setup-registry-conf-file.mustache?raw';
 
+interface RegistryEntryQuickPickItem extends QuickPickItem {
+  entry: RegistryConfigurationEntry;
+}
+
+export enum ActionEnum {
+  ADD_REGISTRY_ACTION = 'addRegistry',
+  END_REGISTRY_ACTION = 'endConfiguringRegistry',
+}
+
+interface ActionQuickPickItem extends QuickPickItem {
+  actionName: ActionEnum;
+}
+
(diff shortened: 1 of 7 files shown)


axel7083 (commented):
On native Linux it works as expected.

However, this start to be confusing when we have multiple machine. I think if we have more than one connection available, we should first select which connection we want to edit?

benoitf:
> However, this start to be confusing when we have multiple machine. I think if we have more than one connection available, we should first select which connection we want to edit?

the setting file is global, not per machine
the idea is to synchronize the changes from your host so all machine shared the same mirroring, etc.

extensions/podman/packages/extension/src/configuration/registry-configuration.ts
```diff
@@ -16,26 +16,229 @@
...
 export class RegistryConfigurationImpl implements RegistryConfiguration {
-  // provides the path to the file being on the host
+  async init(): Promise<Disposable[]> {
+    const disposables: Disposable[] = [];
+    disposables.push(this.registerSetupRegistryCommand());
+    return disposables;
+  }
+
+  registerSetupRegistryCommand(): Disposable {
+    return commands.registerCommand('podman.setupRegistry', async () => this.setupRegistryCommandCallback());
```
jeffmaury:
```suggestion
    return commands.registerCommand('podman.setupRegistry', this.setupRegistryCommandCallback);
```

jeffmaury (commented):
This implementation will NOOP if the file is not mounted into one of the machine, is that on purpose or should we just skip the wrongly configured machines ?

benoitf:
it is on purpose, to not let you think you can configure a file if it's not used

extensions/podman/packages/extension/src/configuration/registry-configuration.ts
benoitf: updated

axel7083 (approved):

PR #11399, opened 2025-02-26
chore(extension-api): adding version property to ContainerProviderConnection

axel7083:
### What does this PR do?

As per requested in https://github.com/podman-desktop/podman-desktop/pull/10768 splitting it.

### Notes

Why using a function (getter) for the `version` field instead of a string field ?

When the extension `registerContainerProviderConnection` it provide an object, and those object cannot be changed after. To overcome this limitation, the status field is a function, that the core can call to get the status.

Since the version cannot be determine when the podman machine is not running, we would put `undefined` when registering the container connection, and would have no way to update it when finally the machine would be started.

Therefore, in a similar manner, having a function allow the core to call it to be able to get it when the extension notify an update.

### Screenshot / video of UI

N/A

Part of https://github.com/podman-desktop/podman-desktop/issues/4215

### How to test this PR?

## Summary by Sourcery

Add a `version` property to the `ContainerProviderConnection` interface in the extension API. This property is a function that returns the version of the container provider connection.

Chores:
- Add a `version` property to the `ContainerProviderConnection` interface in the extension API.
- The `version` property is a function that returns the version of the container provider connection, allowing the core to call it to get the version when the extension notifies an update.

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -401,6 +401,10 @@ declare module '@podman-desktop/api' {
      * the vmTypeDisplayName property cannot be set if vmType is undefined
      */
     vmTypeDisplayName?: string;
+    /**
+     * the remote may be different from the client version
+     */
+    version?(): string | undefined;
   }
 
   export interface PodCreatePortOptions {
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -402,9 +402,13 @@ declare module '@podman-desktop/api' {
      */
     vmTypeDisplayName?: string;
     /**
-     * the remote may be different from the client version
+     * The Container Provider Connection may have a specific version
+     * that differ from the client/host.
+     *
+     * This is common when dealing with remote host, such as Podman Machines.
      */
-    version?(): string | undefined;
+    version?: string;
+    onDidUpdateVersion?: Event<string>;
   }
 
   export interface PodCreatePortOptions {


benoitf:
it needs to follow the same path than the provider version field

```
    readonly version: string | undefined;
    updateVersion(version: string): void;
    onDidUpdateVersion: Event<string>;
```

axel7083:
> it needs to follow the same path than the provider version field
> 
> ```
>     readonly version: string | undefined;
>     updateVersion(version: string): void;
>     onDidUpdateVersion: Event<string>;
> ```

Why not, but this seems overkill, for a given machine, the version will never change?

benoitf:
I mean, you don't know, one implementation could dispatch you to another remote machine while connected

axel7083:
> I mean, you don't know, one implementation could dispatch you to another remote machine while connected

Valid argument

axel7083:
@benoitf I was applying the suggestion you provided, but I am having big doubt for the `updateVersion(version: string): void;`

This feel very private, and exposing it in the `ContainerProviderConnection` would means that any extension could update any `ContainerProviderConnection` :thinking: .

I am not sure we should add this method in the `ContainerProviderConnection` object

https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/extension-api/src/extension-api.d.ts#L391-L404

Let's break it down how an extension (such as Podman), register a container connection

They call the following method

https://github.com/podman-desktop/podman-desktop/blob/1846379fa4195f3400e2ded412df94a9b0975c70/packages/extension-api/src/extension-api.d.ts#L669

Which mean they provide the full object, maybe they should only provide the two following

```
readonly version: string | undefined;
onDidUpdateVersion: Event<string>;
```

but this would be a breaking change, as onDidUpdateVersion would be mandatory, and there need to be optional, same as version. :/

axel7083:
Moving to draft as waiting for feedback

axel7083:
Discussion needed, should happen on the linked issue

PR #11420, opened 2025-02-27
feat: add the ability to auto scroll to provider from statusbar

### LINKED ISSUE #11061: Clicking on provider should scroll to correct provider in the page

When I click on a provider in the statusbar it takes me to the top of the provider page. Ths is confusing in situations where I have a provider that is much lower down the list, e.g. `kind`. 

Scroll down to the provider I click on on the Resources page.

bmahabirbu:
### What does this PR do?

Adds the ability to automatically scroll to the provider in the resource page when clicking on the provider in the statusbar

Fixes https://github.com/podman-desktop/podman-desktop/issues/11061

## Summary by Sourcery

New Features:
- Implement auto-scrolling to a provider in the resources page when the provider is clicked in the status bar.

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.spec.ts
@@ -333,6 +333,21 @@ describe.each<{
     expect(button).toHaveTextContent('Connect ...');
   });
 
+  test('Expect to scroll to the focused element if focus prop is provided', async () => {
+    window.HTMLElement.prototype.scrollIntoView = vi.fn();
+    const customProviderInfo: ProviderInfo = {
+      ...providerInfo,
+      id: 'test-provider',
+      name: 'Test Provider',
+    };
+    providerInfos.set([customProviderInfo]);
+    render(PreferencesResourcesRendering, { focus: 'test-provider' });
+    await vi.waitFor(() => {
+      // Check if scrollIntoView was called
+      expect(window.HTMLElement.prototype.scrollIntoView).toHaveBeenCalledWith({ behavior: 'auto', block: 'start' });
+    });
+  });
+
   describe('provider connections', () => {
     test('Expect to have two container connection region', () => {
       providerInfos.set([providerInfo]);
--- a/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
@@ -393,6 +393,13 @@ function hasAnyConfiguration(provider: ProviderInfo): boolean {
   );
 }
 
+export let focus: string | undefined;
+let providerElementMap: Record<string, HTMLElement> = {};
+
+$: if (focus && providerElementMap[focus]) {
+  providerElementMap[focus].scrollIntoView({ behavior: 'auto', block: 'start' });
+}
+
 function handleError(errorMessage: string): void {
   console.error(errorMessage);
 }
@@ -414,6 +421,8 @@ function handleError(errorMessage: string): void {
 
     {#each providers as provider (provider.id)}
       <div
+        id={provider.id}
+        bind:this={providerElementMap[provider.id]}
         class="bg-[var(--pd-invert-content-card-bg)] mb-5 rounded-md p-3 divide-x divide-[var(--pd-content-divider)] flex"
         role="region"
         aria-label={provider.id}>
--- a/packages/renderer/src/lib/preferences/PreferencesPage.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesPage.svelte
@@ -64,8 +64,8 @@ onMount(async () => {
       properties={properties}
       taskId={+meta.params.taskId} />
   </Route>
-  <Route path="/resources" breadcrumb="Resources" navigationHint="root">
-    <PreferencesResourcesRendering />
+  <Route path="/resources" breadcrumb="Resources" navigationHint="root" let:meta>
+    <PreferencesResourcesRendering focus={meta.query.focus}/>
   </Route>
   <Route path="/docker-compatibility" breadcrumb="Docker Compatibility">
     <PreferencesDockerCompatibilityRendering />
--- a/packages/renderer/src/lib/statusbar/ProviderWidget.spec.ts
+++ b/packages/renderer/src/lib/statusbar/ProviderWidget.spec.ts
@@ -42,6 +42,7 @@ vi.mock('tinro', () => {
 
 const providerMock = {
   name: 'provider1',
+  id: 'provider1-id',
   containerConnections: [],
   kubernetesConnections: [],
   status: 'ready' as ProviderStatus,
@@ -71,7 +72,7 @@ test('Provider widget takes user to /preferences/resources on click by default',
   expect(widget).toBeInTheDocument();
 
   await userEvent.click(widget);
-  expect(router.goto).toBeCalledWith('/preferences/resources');
+  expect(router.goto).toBeCalledWith(`/preferences/resources?focus=${providerMock.id}`);
 });
 
 test('Expect the prop command to be used when it is passed with the entry', async () => {
--- a/packages/renderer/src/lib/statusbar/ProviderWidget.svelte
+++ b/packages/renderer/src/lib/statusbar/ProviderWidget.svelte
@@ -18,7 +18,7 @@ interface Props {
 
 let {
   entry,
-  command = (): void => router.goto('/preferences/resources'),
+  command = (): void => router.goto(`/preferences/resources?focus=${entry.id}`),
   disableTooltip = false,
   class: className,
   tooltipTopRight = false,


axel7083 (commented):
Hey @bmahabirbu !

The approach you took was to handle the scrolling from the `ProviderWidget` component, and this is working in practice 👍 , but may cause some issue in the future. 

Let me explain my POV!

## Possible issue(s)

Let's say later we need to have another component redirecting to the Resource Page, and scroll to a given card just like the Provider Widget, we would duplicate the code you currently wrote, thus making it a little bit harder to maintain.

A possible suggestion I would recommend is to to take a top down approach, the _caller_ here should provide some kind of argument, for the _receiver_ (The resource page) to focus on.

The idea is always to have the core logic closer to the responsible component, so this can be tested in a unit manner, and refactor easily in the future if needed!

## Solution proposal

What option do we have? Thankfully we use a router, `router.goto('/preferences/resources')`, an idea (from top of my head), could be to use a query parameter in the URL.

For example, you use `router.goto('/preferences/resources?focus=<provider-id>')`.

Then, the _caller_ delegate the responsibility of doing the animation, scrolling etc. to the receiver, here our resource page.

## Accessing the `query` parameters

We can see that the page preferences is routed with the following

https://github.com/podman-desktop/podman-desktop/blob/f53a26d89dedf6b591ab3e0aa83c3fdfa39013be/packages/renderer/src/App.svelte#L358-L360

If we go inside the component, we can see the following

https://github.com/podman-desktop/podman-desktop/blob/94f348a17c036c558d94aa82e95f3d3eb251b4e4/packages/renderer/src/lib/preferences/PreferencesPage.svelte#L66-L68

which is where the routing system will display our `PreferencesResourcesRendering` component when we are located in `/preferences/resources`.

From here we can use some sugar synthax of our routing library to get the `focus` query parameter.

```
<Route path="/resources" breadcrumb="Resources" navigationHint="root" let:meta>
  <PreferencesResourcesRendering focus={meta.query.focus}/>
</Route>
```

See here, the `<Route>` component can export a meta object, which has a `query` object with all our query parameters (this is cool! it avoid use to parse the query parameters ourself.)

From this point, we can change the Props of `PreferencesResourcesRendering` to add an **optional** `focus` property (typed as `focus?: string`).

Then inside this component do the animation.

## Conclusion

Using a query parameter is one solution I think would fit well here our use case, as this makes it reusable by any _caller_ in the application, and allow to centralise the code in the responsible component.

To be able to acces

deboer-tim:
+1 to trying query param; logic should definitely be in the page, not client.

bmahabirbu:
Thank you so much @axel7083 for the detailed explanation and @deboer-tim!

axel7083 (commented):
Nice improvement !

I was able to test the application, and confirm the expected behaviour described in https://github.com/podman-desktop/podman-desktop/issues/11061.

## Some comments 💬 

Some pr-check are failing. An important one is the `pr-check: linter, formatter`.

When cloning Podman Desktop, depending on the IDE, you should have some git hook running before committing, which would automatically format the code in the indented manner. 

Others check failing are the unit tests, reading at the logs, we can see the following

```diff
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/lib/statusbar/ProviderWidget.spec.ts > Provider widget takes user to /preferences/resources on click by default
AssertionError: expected "spy" to be called with arguments: [ '/preferences/resources' ]

Received: 

  1st spy call:

  [
-   "/preferences/resources",
+   "/preferences/resources?focus=undefined",
  ]
``` 

inside the `src/lib/statusbar/ProviderWidget.spec.ts` something is broken, and it seems to be linked to the change you made to the url; let's look at the test failing.

https://github.com/podman-desktop/podman-desktop/blob/ed5b08e3bb67dea127ded5a20673048f79efe48e/packages/renderer/src/lib/statusbar/ProviderWidget.spec.ts#L55-L63

There is a check ensuring we are redirecting to the right page, since we change the url to include a query parameter.

> we would need to update that, but here focus is undefined 🤔 might be some mocking issue! We should have the provider id in the updated test and not an undefined value.

Moreover, would be interesting to add a test in `packages/renderer/src/lib/preferences/PreferencesResourcesRendering.spec.ts` to ensure that on mount, if focus props is provided, we focus on it.

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.spec.ts
```diff
@@ -140,6 +140,24 @@ test('Expect to see elements regarding foo provider', async () => {
...
+    id: 'test-provider',
+    name: 'Test Provider',
+  };
+  providerInfos.set([customProviderInfo]);
+  render(PreferencesResourcesRendering, { focus: 'test-provider' });
+  await vi.waitFor(() => {
+    // Check if scrollIntoView was called
+    expect(scrollIntoViewMock).toHaveBeenCalledWith({ behavior: 'auto', block: 'start' });
+  });
+  scrollIntoViewMock.mockRestore();
```
axel7083: nit: you can directly set the mock, and use expect on it,

```suggestion
  window.HTMLElement.prototype.scrollIntoView = vi.fn();
  const customProviderInfo: ProviderInfo = {
    ...providerInfo,
    id: 'test-provider',
    name: 'Test Provider',
  };
  providerInfos.set([customProviderInfo]);
  render(PreferencesResourcesRendering, { focus: 'test-provider' });
  await vi.waitFor(() => {
    // Check if scrollIntoView was called
    expect(window.HTMLElement.prototype.scrollIntoView).toHaveBeenCalledWith({ behavior: 'auto', block: 'start' });
  });
```

feloy:
Thanks @bmahabirbu for your contribution.

As a user, I'm not completely comfortable with the current execution: 
- as the list of providers is pretty short, many of them will not be scrolled on top of the page, by lack of providers after them (the bottom of the page being reached before), and it's difficult to find it (more than before IMO).
- the scroll does not work when I'm already on the Providers page and I click on another provider in the statusbar

bmahabirbu:
Hi @feloy thanks for the feedback. Since this PR was for 1.7, im thinking this will be pushed back to 1.8 since theres no blockers and there's still some discussion to be had on the solution.

For your first point im not sure I understand (maybe because my lack of front end exp). My thought process is if the providers in the resource page is all within view then no scrolling will happen. This shouldnt make it any harder than before to find the right provider. But I would love to hear more on how we can address this!
 
For the second point im using `OnMount` from svelte to trigger the scrolling when the page first renders. Im thinking we can switch this out for another function that can trigger the scrolling when on the page. Maybe `tick` from svelte life cycle hooks can also work?

axel7083 (approved):
On my side LGTM :+1:

feloy:
> Hi @feloy thanks for the feedback. Since this PR was for 1.7, im thinking this will be pushed back to 1.8 since theres no blockers and there's still some discussion to be had on the solution.
> 
> For your first point im not sure I understand (maybe because my lack of front end exp). My thought process is if the providers in the resource page is all within view then no scrolling will happen. This shouldnt make it any harder than before to find the right provider. But I would love to hear more on how we can address this!
> 
kind-scrolled-to

In the screenshot above, I just clicked in the Kind provider. The Kind section is placed in the middle of the page. I'm not sure what would be the design to apply here. @ekidneyrh is there a rule defined for that?

> For the second point im using `OnMount` from svelte to trigger the scrolling when the page first renders. Im thinking we can switch this out for another function that can trigger the scrolling when on the page. Maybe `tick` from svelte life cycle hooks can also work?

That works now with you latest change, thanks

bmahabirbu:
@feloy added the highlight to the provider PTAL thank you!

feloy (approved):
thanks for adding the highlight. LGTM

bmahabirbu:
Thanks @feloy! I'll fix those failing tests and we should be good to go

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -370,11 +370,46 @@ function hasAnyConfiguration(provider: ProviderInfo): boolean {
...
+    }
+    100% {
+      outline: 2px solid rgba(255, 0, 0, 0);
+      outline-offset: 2px;
+    }
+  }
+
+  .highlight {
+    animation: fadeOutline 0.5s ease-in-out forwards;
+  }
```
benoitf: hello, not sure about the usage of a custom css, there is in tailwind the animate / fade, etc

bmahabirbu: Thank you Florent for taking a look! I didn't know we had built in animations I'll do the necessary changes 

Have a good weekend!

benoitf: https://svelte.dev/docs/svelte/svelte-transition

deboer-tim (changes requested):
I tried and the scrolling isn't working unless I have the page open. I'm guessing it is trying to scroll before the element has been created.

We don't use a ⭐️ anywhere else, and the space for the new 'column' is changing the layout of the page - the card titles are indented and the vertical line on the tiles is gone. Design might have some ideas but I'd suggest something that doesn't affect layout, e.g. bg color, border, or an overlay. Maybe even fades away once you've seen it.

Tests also need to be fixed.

benoitf:
I reported it one month ago there https://github.com/podman-desktop/podman-desktop/issues/11061#issuecomment-2687576315 but I think having a page with only one provider might be easier/faster to implement anyway

so I would have only Podman provider on a page, if I close the details it will come back to the list of resources

Firewall:
> I tried and the scrolling isn't working unless I have the page open. 

Same thing is happening to me 

> I reported it one month ago there https://github.com/podman-desktop/podman-desktop/issues/11061#issuecomment-2687576315 but I think having a page with only one provider might be easier/faster to implement anyway
>
>so I would have only Podman provider on a page, if I close the details it will come back to the list of resources

@benoitf I don't think this ticket is the right ticket to do that re-organization. I created a ticket for that here (https://github.com/podman-desktop/podman-desktop/issues/12012), I think its more an Epic. Feel free to add other suggestions on that issue

deboer-tim:
Scrolling to the provider is working for me now. 👍🏼

I don't love the animation. The first time (loaded slower?) I missed the 'expansion' and was wondering why Kind didn't have any icon, until it reappeared. I added this to tomorrow's UX call to get some input.

bmahabirbu:
Thanks @deboer-tim for the feedback! I used the built in tailwind ping animation to be more prominent but it does look out of place at times, maybe in the ux meeting I can showoff the other animation of try something different!

bmahabirbu:
@benoitf  and @deboer-tim, I updated the PR to remove animations for now since 1.18 is due to release Monday.  I can make a separate issue and an addition later in case we want animation in the meantime, while we work on the new Epic

benoitf (approved):
works for me

deboer-tim (approved):
Thanks @bmahabirbu! Works for me too.

Issue #9950, opened 2024-11-18
Allow users to Pin and un-Pin providers on the Statusbar


Firewall:
### Is your enhancement related to a problem? Please describe

We added the new statusbar widgets for the providers in this [task](https://github.com/podman-desktop/podman-desktop/issues/9555). However we also want to add the ability for users to Pin and un-Pin providers from the Statusbar like we can see in the [UX Designs](https://github.com/podman-desktop/podman-desktop/issues/8923#issuecomment-2475997628). 

### Describe the solution you'd like

- Users can pin/unpin items.
- Only the Podman provider is pinned by default.
- There is no limit on the amount of items a user can pin.
- At the start of the statusbar there is a pin icon. This pin icon allows you to turn on or off the categories which are `Providers` and `Extensions` as you can see [here](https://github.com/podman-desktop/podman-desktop/issues/8923#issuecomment-2470280968) 
- There is a popup menu that allows users to pin and unpin providers.
- There is a second popup menu that allows uers to pin and unpin extensions.
- Can we make the dropdown look consistent with what we did in [this task](https://github.com/podman-desktop/podman-desktop/pull/9867)?

axel7083:
Requires:
- https://github.com/podman-desktop/podman-desktop/pull/9724

deboer-tim:
Just recording my feedback from the UX call - if provider widgets are on the status bar (there was some discussion about potentially moving them to nav bar) then I really dislike having this feature specific to provider items, or having different behaviour for Podman vs others. We already promote installing Podman first so its widget will be first; it's confusing and unnecessary for Kubernetes or other container engines to behave differently.

IMHO this function could be identical to what we have in the nav bar where you can right-click and show/hide any item. It's simple, consistent, and more generally useful.

Firewall:
@deboer-tim 
> Just recording my feedback from the UX call - if provider widgets are on the status bar (there was some discussion about potentially moving them to nav bar) then I really dislike having this feature specific to provider items, or having different behaviour for Podman vs others. 

The behavior will be the same. `Podman` will just be `Pinned by default` but you can unpin it like any other provider. 

> We already promote installing Podman first so its widget will be first; it's confusing and unnecessary for Kubernetes or other container engines to behave differently.

No difference in behavior. Clicking on the provider will always take the take user to the same place, the resources/provider page

>IMHO this function could be identical to what we have in the nav bar where you can right-click and show/hide any item. It's simple, consistent, and more generally useful.

[...]

PR #11422, opened 2025-02-27
feat(main): adding pin-registry for providers

axel7083:
### What does this PR do?

Split of https://github.com/podman-desktop/podman-desktop/pull/11036, you can see the full PR working on the PR their.

This one only add the changes in the `main` package.

### Screenshot / video of UI

N/A

Part of https://github.com/podman-desktop/podman-desktop/issues/9950

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/statusbar/pin-registry.ts
+++ b/packages/main/src/plugin/statusbar/pin-registry.ts
@@ -103,7 +103,7 @@ export class PinRegistry implements IDisposable {
           properties: {
             [STATUS_BAR_PIN_CONSTANTS.PINNED_CONFIGURATION_PROPERTY]: {
               description: 'pinned items in the status bar',
-              type: ['string'],
+              type: 'array',
               hidden: true,
               default: ['podman'],
             },
--- a/packages/main/src/plugin/statusbar/pin-registry.ts
+++ b/packages/main/src/plugin/statusbar/pin-registry.ts
@@ -1,0 +1,128 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/statusbar/pin-registry.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.#disposables.push(
+      this.configurationRegistry.registerConfigurations([
+        {
+          id: STATUS_BAR_PIN_CONSTANTS.STATUS_BAR_CONFIGURATION,
+          title: 'Status Bar Pin',
+          type: 'object',
+          properties: {
+            [STATUS_BAR_PIN_CONSTANTS.PINNED_CONFIGURATION_PROPERTY]: {
+              description: 'pinned items in the status bar',
+              type: ['string'],
```
jeffmaury: Why is this not an an array (like disabled extensions ?

gastoner (approved):
Codewise LGTM

packages/main/src/plugin/statusbar/pin-registry.ts
axel7083: cc @benoitf from my understand using `type: ['string'],` synthax mean an array of string? and using `type: string` mean just a string?

jeffmaury: See https://github.com/podman-desktop/podman-desktop/blob/8dd7b6de9f8e3292cb1648548689c3bbfbfed18b/packages/main/src/plugin/extension/extension-loader.ts#L296 for a similar definition

axel7083: Thanks! Updated

benoitf (approved):

PR #11444, opened 2025-02-28
fix: display empty screen only when received data

### LINKED ISSUE #11437: Kubernetes screens shouldn't flash empty screen when navigating

I don't know if it's a case of 'dev is slower than prod build' or due to the experimental work/other changes, but in 1.16 I can tab between the Kubernetes resource pages like Pod or Deployment and I see data right away. In current builds, I see the empty screen for a split second first.

When we're connected to a cluster, you shouldn't see the empty screen flash when changing between Kubernetes pages.

_No response_

### Additional context

_No response_

feloy:

### What does this PR do?

On Kubernetes lists, display empty screen only when received data.

Fixes #11437

### How to test this PR?

## Summary by Sourcery

This PR fixes an issue where the empty screen was displayed before data was received. The empty screen is now only displayed after data has been received from the Kubernetes API.

Bug Fixes:
- Fixes an issue where the empty screen was displayed before data was received, causing a flicker.

Tests:
- Updates tests to wait for the 'No resources' message to be displayed before asserting that it is visible.

### DIFF
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretList.spec.ts
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretList.spec.ts
@@ -42,8 +42,10 @@ test('Expect configmap empty screen', async () => {
   vi.mocked(states).kubernetesCurrentContextConfigMapsFiltered = writable<KubernetesObject[]>([]);
   vi.mocked(states).kubernetesCurrentContextSecretsFiltered = writable<KubernetesObject[]>([]);
   render(ConfigMapSecretList);
-  const noNodes = screen.getByRole('heading', { name: 'No configmaps or secrets' });
-  expect(noNodes).toBeInTheDocument();
+  await vi.waitFor(() => {
+    const noNodes = screen.getByRole('heading', { name: 'No configmaps or secrets' });
+    expect(noNodes).toBeInTheDocument();
+  });
 });
 
 test('Expect configmap and secrets list', async () => {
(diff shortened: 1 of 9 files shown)


axel7083 (commented):
Codewise LGTM

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretList.spec.ts
```diff
@@ -42,8 +42,10 @@ test('Expect configmap empty screen', async () => {
   vi.mocked(states).kubernetesCurrentContextConfigMapsFiltered = writable<KubernetesObject[]>([]);
   vi.mocked(states).kubernetesCurrentContextSecretsFiltered = writable<KubernetesObject[]>([]);
   render(ConfigMapSecretList);
-  const noNodes = screen.getByRole('heading', { name: 'No configmaps or secrets' });
-  expect(noNodes).toBeInTheDocument();
+  await vi.waitFor(() => {
+    const noNodes = screen.getByRole('heading', { name: 'No configmaps or secrets' });
+    expect(noNodes).toBeInTheDocument();
```
benoitf, line 47: nit: I think that getByRole is failing if it's not in the document anyway (and I see the same lines being repeated over and over)

benoitf (approved):
not sure we need all the toBeInTheDocument for each getByRole

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretList.spec.ts
feloy: Yes right, I see it everywhere in the tests, so I do the same, but they are effectively not necessary. I'll add an entry to the code guideline and make some cleanup, as part of other issues

benoitf: yes it's not blocking the merge

deboer-tim (approved):
Tested, worked well. (actually made it feel faster too :) )

PR #11471, opened 2025-03-03
fix: add back button cursor pointer

### LINKED ISSUE #11436: Buttons and toggles no longer have pointer cursors
### Bug description

I spent a while trying to figure out why the cursor wasn't changing to a pointer when I hovered over a button I was editing, only to eventually figure out this was a behaviour change when we upgraded to Tailwind v4:
https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor

cc @ekidneyrh @MariaLeonova 

If we're ok with the change, we can just close this. If not, we could add the workaround in the link or add cursor-pointer to a few of our components. This just felt like something we should choose intentionally vs unknown change.

### Operating system

macOS 15.3

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

Hover over any button or toggle in Podman Desktop. The cursor changes to a pointer in 1.16, but doesn't in the current branch.

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Reverts the Tailwind v4 change to remove the default button cursor pointer. (see https://tailwindcss.com/docs/upgrade-guide)

As discussed on the standup, we'll undo this change for the current release to avoid regression (especially on components that have no hover effect), then possibly undo this later and replace with targeted fixes for elements where we have no hover or focus indicator.

### Screenshot / video of UI

N/A (recording isn't capturing cursor correctly)

Fixes #11436.

### How to test this PR?

Hover over the Learning Center title, buttons. Prior to this change there's no change to the cursor (no indication the LC is clickable), with this PR the cursor changes to a pointer like in 1.16.

### DIFF
--- a/packages/renderer/src/app.css
+++ b/packages/renderer/src/app.css
@@ -13,3 +13,10 @@ a {
   --spacing-leftnavbar: 48px;
   --spacing-leftsidebar: 170px;
 }
+
+@layer base {
+  button:not(:disabled),
+  [role='button']:not(:disabled) {
+    cursor: pointer;
+  }
+}


axel7083 (approved):

cdrage (approved):

SoniaSandler:
@Mergifyio backport 1.17.x

Issue #10385, opened 2024-12-16
How to represent that Kubernetes resources are not accessible?


feloy:
### UX Description

With https://github.com/podman-desktop/podman-desktop/issues/7629, the frontend now knows which Kubernetes resources (pods, deployments, nodes, etc) the user can access.

How do we want to represent that some resources are not accessible, in these different pages?

- In this page, Pods and/or Deployments may not be accessible:

![Image](https://github.com/user-attachments/assets/84a35ba4-a6c7-46dc-bcee-f65283e55cd5)

- In the Kubernetes menu, some Kubernetes resources may not be accessible:
- In the Kubernetes Dashboard, some Kubernetes resources may not be accessible:

![Image](https://github.com/user-attachments/assets/db8c1a93-9dd5-42f1-8510-2d0d38d3205d)

The backend sends two information for each Kubernetes resource:
- is the resource accessible (more exactly, does the user have the permission to watch the resource, used to get the list of objects)?
- optionally, a reason why it is accessible / not accessible

For now, the permissions to create/delete/... (all other operations except watch/list) are not known.

### Request type

UX analysis/suggestions for improvement

### Primary Contact

@feloy 

### Deadline for request

as part of the current sprint/next sprint

MariaLeonova:
@feloy there are several ways to mark something inaccessible:
 - grey out a particular element
 - add an icon / badge; potentially accompanied by a small label text (`not accessible`)
 - display an error message next to it

There also needs to be a way to provide more info for the user, in form of a tooltip, for example on what the status is and why. 

@ekidneyrh the dashboard is your design, WDYT? Let's discuss!

feloy:
> [@feloy](https://github.com/feloy) there are several ways to mark something inaccessible:
> 
> * grey out a particular element
> * add an icon / badge; potentially accompanied by a small label text (`not accessible`)
> * display an error message next to it
> 
> There also needs to be a way to provide more info for the user, in form of a tooltip, for example on what the status is and why.
> 

For the first page (the list of Kubernetes contexts), I was thinking about greying the `Pods` and/or `Deployments` text, and displaying a lock (🔒) instead of the number of pods/resources, with a tooltip explaining that the user doesn't have the authorization. That seems aligned with your proposal.


> [@ekidneyrh](https://github.com/ekidneyrh) the dashboard is your design, WDYT? Let's discuss!

ekidneyrh:
[...]

PR #11473, opened 2025-03-03
feat: added visual indicator for kubernetes permissions - apply button

gastoner:
### What does this PR do?
Adds visual indicator for users to differentiate between permissions to kubernetes deployments - just APPLY button in upper right corner

This is just for deployments - apply button

### Screenshot / video of UI
Mockup (only for deployments): 
Image

Required for #11107 
Required for #11307 - needs to be merged first

### How to test this PR?
Create an user with not all permissions: 
https://github.com/podman-desktop/podman-desktop/issues/10655

### DIFF
--- a/packages/renderer/src/lib/kube/KubeActions.svelte
+++ b/packages/renderer/src/lib/kube/KubeActions.svelte
@@ -1,5 +1,11 @@
 <script lang="ts">
 import KubeApplyYamlButton from '/@/lib/kube/KubeApplyYAMLButton.svelte';
+
+interface Props {
+  disabled?: boolean;
+}
+
+let { disabled = false }: Props = $props();
 </script>
 
-<KubeApplyYamlButton />
+<KubeApplyYamlButton {disabled}/>
--- a/packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
+++ b/packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
@@ -8,9 +8,11 @@ import { type Readable, type Unsubscriber, type Writable } from 'svelte/store';
 
 import { listenResources } from '/@/lib/kube/resources-listen';
 import type { IDisposable } from '/@api/disposable.js';
+import type { ResourceName } from '/@api/kubernetes-contexts-states';
 
 import { withBulkConfirmation } from '../actions/BulkActions';
 import KubeActions from '../kube/KubeActions.svelte';
+import { listenResourcePermitted } from '../kube/resource-permission';
 import KubernetesCurrentContextConnectionBadge from '../ui/KubernetesCurrentContextConnectionBadge.svelte';
 import type { KubernetesObjectUI } from './KubernetesObjectUI';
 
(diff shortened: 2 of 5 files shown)


gastoner:
This is half of PR #11307

packages/renderer/src/lib/objects/KubernetesObjectsList.svelte
```diff
@@ -45,6 +47,9 @@ let resources = $state<{ [key: string]: KubernetesObject[] | undefined }>({});
 let resourceListeners: (IDisposable | undefined)[] = [];
 let legacyUnsubscribers: Unsubscriber[] = [];
 
+let kubeActionsDisabled: boolean = $state(false);
+let selectedResource: ResourceName = $state('deployments');
```
deboer-tim, line 51: Should not have deployment-specific code in this file or default to deployments.

packages/renderer/src/lib/kube/KubeActions.svelte
```diff
@@ -1,5 +1,11 @@
 <script lang="ts">
 import KubeApplyYamlButton from '/@/lib/kube/KubeApplyYAMLButton.svelte';
+
+interface Props {
+  disabled?: boolean;
```
deboer-tim, line 5: There could be several actions, and maybe not all of them are disabled when the resource is not accessible. Seems like this should just be an `accessible` or permissions property and below we decide which action(s) it disables.

deboer-tim:
I'm a little confused with disabling the Apply YAML button, because I can't find an issue or design where this is discussed. Specifically, unless _every_ resource is inaccessible, I can always apply yaml since there is no guaranteed correlation between what you're looking at and what file you give it. (i.e. I can be on the Pods page and apply a yaml that has a Deployment)

I'm more _likely_ to be on a page where the resource I'm going to apply is listed, and we'd certainly want to warn the user if applying a resource is going to fail (or did fail) b/c it's not accessible, but I'm not sure we should be disabling Apply.

gastoner:
> I'm a little confused with disabling the Apply YAML button, because I can't find an issue or design where this is discussed. Specifically, unless _every_ resource is inaccessible, I can always apply yaml since there is no guaranteed correlation between what you're looking at and what file you give it. (i.e. I can be on the Pods page and apply a yaml that has a Deployment)
> 
> I'm more _likely_ to be on a page where the resource I'm going to apply is listed, and we'd certainly want to warn the user if applying a resource is going to fail (or did fail) b/c it's not accessible, but I'm not sure we should be disabling Apply.

I'm really just following the mockup in https://github.com/podman-desktop/podman-desktop/issues/10385#issuecomment-2615490295 (third image) from Emma

cdrage (changes requested):
I agree with @deboer-tim  and I believe this should go back into discussion mode.

Having "Apply YAML" grayed out right above the icon that says "Connected" does not make sense at all. If my cluster is accessible and connected, I should be able to Apply YAML.

Perhaps as a user I need to use the Apply YAML is an Admission Controller that would "re-enable" Deployments if it's not accessible at the time.. https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/

Doesn't make sense to gray it out even though deployments isn't accessible.

gastoner:
cc: @feloy @ekidneyrh how should I continue ?

feloy:
> cc: @feloy @ekidneyrh how should I continue ?

Let's discuss these questions during the next UX call

gastoner:
Closing this after feedback from UX call => keeping apply button enabled

PR #11475, opened 2025-03-03
chore: reduce default gap in expandable

deboer-tim:
### What does this PR do?

When I created the Expandable component I used the standard pt-5 that we use everywhere as a gap between the control/title and content. This is fine as a default in Podman Desktop where we tend to use the control with big headers, but isn't a good default for a component - e.g. when I tried it on the bootc build page where it would be used for smaller section expanders.

This just uses a more standard gap-2 instead, and adds a pt to the Kubernetes dashboard to compensate.

### Screenshot / video of UI

N/A

Related to #11364.

### How to test this PR?

Just confirm no visual change to Kubernetes dashboard guide section.


jeffmaury (approved):
LGTM

PR #11492, opened 2025-03-04
chore: added permission listener for kubernetes permissions

gastoner:
### What does this PR do?
Adds kubernetes permission listener

Required for #11107

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/kube/resource-permission.spec.ts
+++ b/packages/renderer/src/lib/kube/resource-permission.spec.ts
@@ -51,35 +51,25 @@ beforeEach(() => {
 });
 
 describe('listenResourcePermitted', () => {
-  test('should enable apply yaml when resource is permitted', async () => {
-    let callbackCalled = false;
-    const callback = async (permitted: boolean): Promise<void> => {
-      await vi.waitFor(() => expect(permitted).toBe(true));
-      callbackCalled = true;
-    };
+  test('resource shoudl be permitted', async () => {
+    const callbackMock = vi.fn();
 
     kubernetesContexts.set([mockContext1, mockContext2]);
-
     kubernetesContextsPermissions.set([{ contextName: 'context-name2', resourceName: 'deployments', permitted: true }]);
 
-    await listenResourcePermitted('deployments', callback);
-    expect(callbackCalled).toBeTruthy();
+    await listenResourcePermitted('deployments', callbackMock);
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/kube/resource-permission.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  vi.mocked(window.getConfigurationValue<boolean>).mockResolvedValue(true);
+});
+
+describe('listenResourcePermitted', () => {
+  test('should enable apply yaml when resource is permitted', async () => {
+    let callbackCalled = false;
+    const callback = async (permitted: boolean): Promise<void> => {
+      await vi.waitFor(() => expect(permitted).toBe(true));
+      callbackCalled = true;
+    };
```
feloy: Instead, you could write somrthing like `expect(callbackMock).toBeCalledWith(true)`

packages/renderer/src/lib/kube/resource-permission.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  currentContext: true,
+};
+
+beforeEach(() => {
+  vi.resetAllMocks();
+  vi.mocked(window.getConfigurationValue<boolean>).mockResolvedValue(true);
+});
+
+describe('listenResourcePermitted', () => {
+  test('should enable apply yaml when resource is permitted', async () => {
```
feloy: the doc should not speak about `apply yaml`, it is just an example of usage

packages/renderer/src/lib/kube/resource-permission.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  clusterInfo: {
+    name: 'cluster-name2',
+    server: 'https://server-name2',
+  },
+  currentContext: true,
+};
+
+beforeEach(() => {
+  vi.resetAllMocks();
+  vi.mocked(window.getConfigurationValue<boolean>).mockResolvedValue(true);
```
feloy: I would also add a test for non experimental mode, to be sure that the result is always true

feloy (approved):

cdrage (approved):
LGTM.

Only concern is lack of any comments in the code, but it's a nit / not blocking.

gastoner:
@cdrage comments added

cdrage (approved):
Much better! Thank you. I understand it a bit more what's happening.

PR #11516, opened 2025-03-04
chore: remove unused kube paths from pods page

deboer-tim:
### What does this PR do?

Two more places where kind is always 'podman' so we can clean up the Kubernetes logic.

### Screenshot / video of UI

N/A

Part of #8819.

### How to test this PR?

Just PR check.


cdrage (approved):
Tested and saw no changes (intended!) LGTM

PR #11519, opened 2025-03-04
chore: clean up PodInfoUI kind

deboer-tim:
### What does this PR do?

The kind property in PodInfoUI can only be 'podman', so remove it.

Likewise, PodGroupInfoTypeUI is only used by ProviderInfo and can only be 'podman' too.

ProviderInfo could probably remove all Kubernetes support at this point, but since the 'kubernetes' string is already used in the file and the file is used in lots of other places I'm just doing the basic removal of the type here.

### Screenshot / video of UI

N/A

Part of #8819.

### How to test this PR?

Just code review.


cdrage (approved):
LGTM

SoniaSandler (approved):
LGTM

PR #11549, opened 2025-03-06
chore: removed no reactive reassign eslint rule

### LINKED ISSUE #11510: Enable svelte/no-reactive-reassign and update impacted files
https://github.com/podman-desktop/podman-desktop/blob/8dd7b6de9f8e3292cb1648548689c3bbfbfed18b/eslint.config.mjs#L272

gastoner:
### What does this PR do?
Enables reactive reassign eslint rule and applying changes

### Screenshot / video of UI
Hopefuly none :D 

Closes #11510 

### How to test this PR?
PR checks

### DIFF
--- a/packages/renderer/src/lib/container/ContainerEmptyScreen.svelte
+++ b/packages/renderer/src/lib/container/ContainerEmptyScreen.svelte
@@ -8,6 +8,7 @@ export let runningOnly: boolean;
 export let stoppedOnly: boolean;
 const helloImage: string = 'quay.io/podman/hello:latest';
 
+let inProgress = false;
 $: inProgress = false;
 $: title = getTitle(runningOnly, stoppedOnly);
 $: messageCommandLine = getMessageCommandLine(stoppedOnly);
--- a/packages/renderer/src/lib/onboarding/Onboarding.svelte
+++ b/packages/renderer/src/lib/onboarding/Onboarding.svelte
@@ -63,6 +63,7 @@ $: onboardingItems = onboardings;
 let activeStep: ActiveOnboardingStep;
 let activeStepContent: OnboardingStepItem[][];
 
+let executing = false;
 $: executing = false;
 let globalContext: ContextUI;
 let displayCancelSetup = false;
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionDetailsLogs.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionDetailsLogs.svelte
@@ -19,7 +19,8 @@ export let setNoLogs: () => void;
 export let noLog: boolean;
 let logsTerminal: Terminal;
 
-$: noLogs = !!noLog;
+$: noLog = !!noLog;
+let noLogs = !!noLog;
 
 // Log
 let logsXtermDiv: HTMLDivElement;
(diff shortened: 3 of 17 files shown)


feloy:
I find difficult to check if the changes are valid, as svelte 4 logic around reactivity is quite opaque, compared to the one in svelte 5.

Do you think it would be the opportunity to port the impacted components to svelte 5?

gastoner:
> I find difficult to check if the changes are valid, as svelte 4 logic around reactivity is quite opaque, compared to the one in svelte 5.
> 
> Do you think it would be the opportunity to port the impacted components to svelte 5?

Sure I can try, but it would be probably separate PR for each file ~17 PRs 🤯

jeffmaury:
I think we can merge this as change are quite simple 
For the Svelte 5 migration I have this on my todo list will create the issues

jeffmaury (approved):
LGTM

packages/ui/src/lib/table/Table.svelte
```diff
@@ -39,7 +39,7 @@ $: selectedItemsNumber = row.info.selectable
   : 0;
 
 // do we need to unselect all checkboxes if we don't have all items being selected ?
-$: selectedAllCheckboxes = row.info.selectable
+let selectedAllCheckboxes = row.info.selectable
```
feloy: This breaks the behaviour described in the comment above: if you select all the checkboxes individually, the top 'all' checkbox should be automatically checked. With this change, this does not happen anymore

packages/renderer/src/lib/preferences/PreferencesConnectionDetailsLogs.svelte
```diff
@@ -19,7 +19,7 @@ export let setNoLogs: () => void;
 export let noLog: boolean;
 let logsTerminal: Terminal;
 
-$: noLogs = !!noLog;
+let noLogs = !!noLog;
```
feloy: It's not straightforward to me that this does not change the behaviour

feloy (changes requested):

benoitf (commented):
I think this PR focus should only be on declaring the variables being used in the reactive block.

- don't do replacement of reactive without reactive

packages/renderer/src/lib/onboarding/Onboarding.svelte
```diff
@@ -63,7 +63,7 @@ $: onboardingItems = onboardings;
 let activeStep: ActiveOnboardingStep;
 let activeStepContent: OnboardingStepItem[][];
 
-$: executing = false;
+let executing = false;
```
benoitf: it's changing the behavior as it won't reset executing to false if some attributes have changed

packages/renderer/src/lib/container/ContainerEmptyScreen.svelte
```diff
@@ -8,7 +8,7 @@ export let runningOnly: boolean;
 export let stoppedOnly: boolean;
 const helloImage: string = 'quay.io/podman/hello:latest';
 
-$: inProgress = false;
+let inProgress = false;
```
benoitf, line 11: should be two instructions

gastoner:
@feloy @benoitf can you PTAL?

benoitf (approved):

feloy (approved):

PR #11555, opened 2025-03-06
chore: kind extension - do not deploy ingress when using config yaml

### LINKED ISSUE #11469: Config file and form options conflict in Kind cluster creation
### Bug description

I've found a bug with the new feature: the kind config file. When specifying it, the message indicates that other options will be ignored. However, if the name of the kind cluster in the form doesn't match the name used in the config file, we get an error. Despite this, the cluster is created with the name specified in the config, and it is functional. 

Image

I've also tried changing other options, such as disabling the ingress controller.  As I understand it, this option should be ignored when defining a config file, but it isn't.

### Operating system

Fedora

### Installation Method

None

### Version

1.17.0

### Steps to reproduce

Use the following Kind config file:
```

kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: app-1-cluster
```

Try to create a cluster using the config file.
You should see the warning: `By specifying a config file, all other options will be ignored.`
If the cluster name in the form isn't changed to match the cluster name in the config file (app-1-cluster), an error will appear.
Check that the Kind cluster is active and functional on the resources page. 

### Relevant log output

```shell

```

### Additional context

_No response_

cdrage:
chore: kind extension - do not deploy ingress when using config yaml

### What does this PR do?

When providing a configuration yaml, ingress should be ignored as
everything else in the forms should be ignored.

### Screenshot / video of UI

https://github.com/user-attachments/assets/2360fa7c-87c3-4579-a3f3-4a7fb1bd9901

Closes https://github.com/podman-desktop/podman-desktop/issues/11469

### How to test this PR?

1. Deploy with yaml:

```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: app-1-cluster
```

2. See it no longer calls ingress, etc. Deploys with no errors.

### DIFF
--- a/extensions/kind/src/create-cluster.spec.ts
+++ b/extensions/kind/src/create-cluster.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023-2024 Red Hat, Inc.
+ * Copyright (C) 2023-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@ vi.mock('node:fs', () => ({
     mkdtemp: vi.fn(),
     rm: vi.fn(),
   },
+  readFileSync: vi.fn(),
 }));
 
 vi.mock('@podman-desktop/api', async () => {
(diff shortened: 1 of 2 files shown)


gastoner (approved):
LGTM on Windows

jeffmaury:
I have a big concern about this one as deploying the ingress controller and configuring the cluster are 2 different concerns to me: how can you define an ingress in a Kind configuration cluster

feloy (approved):

cdrage:
> I have a big concern about this one as deploying the ingress controller and configuring the cluster are 2 different concerns to me: how can you define an ingress in a Kind configuration cluster

From the docs, it recommends deploying it separate when using a config: 

https://kind.sigs.k8s.io/docs/user/ingress/

Disabling deploying ingress when a config is specified makes sense in order to make sure that say.. a company with a corporate environment gives a `kind` config to all employees, it's good to not deploy an ingress by accident that's been blocked to pull images from / extra features not needed.

I'm open for discussion though either way!

If we do still want to deploy ingress controller, we should output that ingress toggle will be the only one that still works in the form even after you supply the config.

benoitf (approved):

jeffmaury:
So from my POV I don't understand why we can't deploy an ingress controller when a config file is given: the other field are greyed in that case because they are part of the config file but ingress is processed after the cluster has been created.

cdrage:
> So from my POV I don't understand why we can't deploy an ingress controller when a config file is given: the other field are greyed in that case because they are part of the config file but ingress is processed after the cluster has been created.

We could add an indicator underneath the Ingress options that it'll still be deployed?

What are your opinions @benoitf / @feloy

amisskii:
I agree with Jeff. I think deploying the ingress controller should remain, even if we use a config file, as we can't configure this directly in the config file.

feloy:
> I agree with Jeff. I think deploying the ingress controller should remain, even if we use a config file, as we can't configure this directly in the config file.

I also second that. Ingress install being an option on top of the cluster install itself, it should be available independently of the cluster install method

cdrage:
Thanks for the feedback!

I will mark as draft and modify the ingress deployment method to deploy to the namespace within the configuration file if Ingress deployment has been selected

PR #11556, opened 2025-03-06
chore: indicate in provider widget if update is available

### LINKED ISSUE #11426: Provider update available on statusbar

When I have an update available for Podman (and other resources, kind etc) that status is not shown on the statusbar.

Image

Show the "upgrade" icon on the statusbar item for the provider as per the UX design. Ignore the fact that the provider in this screenshot is an extension.

Image

_No response_

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
This PR adds an indication if there is an available update in the ProviderWidget.

Currently, this will only show up for Podman, since most other extensions (especially CLI tools) only include the update information in the registered CLI tool and not the registered provider. https://github.com/podman-desktop/podman-desktop/issues/11557

### Screenshot / video of UI
Screenshot 2025-03-21 at 12 25 06 PM

Closes https://github.com/podman-desktop/podman-desktop/issues/11426

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/statusbar/ProviderWidget.spec.ts
+++ b/packages/renderer/src/lib/statusbar/ProviderWidget.spec.ts
@@ -130,3 +130,9 @@ test('class props should be propagated to button', async () => {
   const widget = getByRole('button', { name: 'provider1' });
   expect(widget).toHaveClass('potatoes');
 });
+
+test('Epect tooltip to show Update available text if the provider has an update', () => {
+  providerMock.updateInfo = { version: '1.1.0' };
+  render(ProviderWidget, { entry: providerMock });
+  expect(screen.getByText('Update available')).toBeInTheDocument();
+});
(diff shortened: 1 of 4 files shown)


SoniaSandler:
@MariaLeonova @ekidneyrh @benoitf 
here are the updated screenshots:
Screenshot from 2025-03-20 09-07-07
Screenshot from 2025-03-20 09-06-51
I think that based on the UX discussion, we can start with having the update available text show once at the end without the update icon in the tooltip, but just in the statusbar (purple area) like in the first screenshot

MariaLeonova:
@SoniaSandler the first one 
image
 looks good without the icon. I wonder if it makes sense to put the update info on top and add some bottom padding to it, so it stands out some more?

SoniaSandler:
@MariaLeonova something like this?
Screenshot from 2025-03-20 12-00-05

We could also do 
Screenshot from 2025-03-20 12-05-22

MariaLeonova:
> @MariaLeonova something like this? Screenshot from 2025-03-20 12-00-05

This looks great to me!

benoitf (approved):

feloy (approved):

Issue #10081, opened 2024-11-22
Test Kubernetes auth executable is in the path


feloy:
### Is your enhancement related to a problem? Please describe

Kubernetes clusters from Cloud providers generally need an executable installed on the user system, to negotiate auth tokens.

It can happen that users add the path of these executables to the PATH **for shell sessions**, but this change is not applicable to Podman desktop.

Because of this, Podman desktop is not able to get new tokens and clusters are not accessible.

### Describe the solution you'd like

It may be interesting that Podman Desktop check if the command defined for authentication is accessible from the Podman context, and if not, display a warning in the Settings > Kubernetes page

### Additional context

See issue https://github.com/podman-desktop/extension-kubernetes-contexts/issues/398

damaestro:
A perfect example of this is needing the `aws` command to get a user token.

On MacOS users will likely install via brew: `brew install awscli`

This results in `aws` being installed at `/opt/homebrew/bin/aws` which is usually in the users `PATH`. However, it's not for Podman Desktop.

A short term fix is to give the full command location in your `~/.kube/config` after generating it with `aws eks update-config`.

feloy:
should fix https://github.com/podman-desktop/podman-desktop/issues/10479

PR #11569, opened 2025-03-07
feat: check if user auth exec command is in the path

feloy:

### What does this PR do?

Check if user auth exec command is in the path.

This is the check part only, sending the data to the front will be done in a follow-up PR

### Screenshot / video of UI

Backend only

Part of #10081 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-preflight-checker.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-preflight-checker.ts
@@ -1,0 +1,112 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import path from 'node:path';
+
+import { isWindows } from '/@/util.js';
+
+import type { Exec } from '../util/exec.js';
+import type { KubeConfigSingleContext } from './kubeconfig-single-context.js';
+
+export interface PreflightCheckContextResults {
+  contextName: string;
+  results: PreflightCheckResult[];
+}
+
+export interface PreflightCheckResult {
+  description: string;
+}
+
+export class ContextsPreflightChecker {
+  #exec: Exec;
+
+  #registry = new PreflightCheckRegistry();
+
+  constructor(exec: Exec) {
+    this.#exec = exec;
+  }
+
+  async check(contextName: string, config: KubeConfigSingleContext): Promise<void> {
+    // reset all results for this context
+    this.#registry.reset(contextName);
+    await this.frelightCheckUserExecCommand(contextName, config);
+  }
+
+  getResults(): PreflightCheckContextResults[] {
+    return this.#registry.getAll();
+  }
+
+  private isBasename(command: string): boolean {
+    return path.basename(command) === command;
+  }
+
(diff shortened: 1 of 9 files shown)


packages/main/src/plugin/kubernetes/contexts-preflight-checker.ts
```diff
@@ -0,0 +1,112 @@
...
+export interface PreflightCheckContextResults {
+  contextName: string;
+  results: PreflightCheckResult[];
+}
+
+export interface PreflightCheckResult {
+  description: string;
+}
+
+export class ContextsPreflightChecker {
```
axel7083, line 35: Will this class manage more than the [kubeconfig.v1/#ExecConfig](https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#ExecConfig) ?

Is it planned to run the command defined in the `ExecConfig` later on to refresh the connections?

axel7083 (commented):
I must admit the PRs was not totally clear, and it got better after I discovered that the kubeconfig could contain an optional `exec` field, containing an [kubeconfig.v1/#ExecConfig](https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#ExecConfig).

This config must contain a command, with some args and envs, that should be executed to be able to update the kubeconfig is, that all correct ?

I questioned if we want to run at some point this command? (in the future?) because if we do, we should probably have a dedicated class handling all of this, and no a generic ContextsPreflightChecker (I guess?)

## Global remark (not this PR specific)

Can we avoid adding more and more files at the root directory of the kubernetes folder ?

image

I can't even see them all now :(

feloy:
> I must admit the PRs was not totally clear, and it got better after I discovered that the kubeconfig could contain an optional `exec` field, containing an [kubeconfig.v1/#ExecConfig](https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#ExecConfig).
> 
> This config must contain a command, with some args and envs, that should be executed to be able to update the kubeconfig is, that all correct ?
> 

Sorry for the lack of context. You are right, this is related to KubeConfig containing users having an exec section.

The Kubernetes clients need to obtain authentication tokens to be able to connect to kube api, and kubeconfig supports authentication plugins (exec, file, gcp, etc, you can see implementations of such plugins in https://github.com/kubernetes-client/javascript/tree/main/src, in files `*_auth.ts`).

These plugins are responsible for providing a fresh token to the client, using the information provided in the `user.plugin` section of the user (`user.exec` for the exec plugin). The kubeconfig file is not updated, the token is passed to the client then forgotten (except in client's cache).

For an exec plugin, command/args/env is provided to the client, which is responsible for executing this command, to obtain a token in response to the command. This is handled by the kubernetes-client library, and from PD point of view, this is executed every time a request is done (this is not done at time of parsing the kubeconfig file, but when following requests are executed). The library maintains a cache (in memory), so the command is in reality called for the first connection attempt.


> I questioned if we want to run at some point this command? (in the future?) because if we do, we should probably have a dedicated class handling all of this, and no a generic ContextsPreflightChecker (I guess?)
> 

I don't think so. There is some logic around calling this command, which is handled by the library, in the exec plugin.

packages/main/src/plugin/kubernetes/contexts-preflight-checker.ts
feloy: The idea of this class is to run any kind of check, to analyze the kubeconfig and detect some problems on it, to be able to give hints to the user.

axel7083:
>  This is handled by the kubernetes-client library,

@feloy ohhhh! That's interesting, thanks for the clarification, this was the element I was curious about. So follow-up question: why do we need to check for `exec` config if this is automate by the client library?

feloy:
> > This is handled by the kubernetes-client library,
> 
> @feloy ohhhh! That's interesting, thanks for the clarification, this was the element I was curious about. So follow-up question: why do we need to check for `exec` config if this is automate by the client library?

We could, effectively, but the difficult part is how to catch the error, as it is raised by the library during the *first* connection attempt, which can happen at any time (you can look at the init code of the KubernetesClient class, many calls are executed at this moment: fetchAPIGroups, etc).

In the medium term, if we are able to cleanup the Kubernetes-related code in this direction, I think that the first connection attempt would always be the health check. When we are at this term, we could handle the error correctly

feloy:
But in the other hand, if it fails, it will fail for all calls, including the health check. So it should be enough to catch the error during the health check.

I'll have a try in a separate PR

feloy:
closing in favor of #11608

PR #11598, opened 2025-03-10
feat: added experimental features feedback notification

### LINKED ISSUE #10228: Experimental features feedback notification

gastoner:
### What does this PR do?
Adds experimental features feedback notification

WORK IN PROGRESS

- [x] Add this feature to kubernetes experimental
- [x] Add this to pin/unpin providers when available https://github.com/podman-desktop/podman-desktop/pull/11973
- [x] Update the mechanism of thumbs up/down to not show in each message box

### Screenshot / video of UI
TODO

Closes #10228 

### How to test this PR?
Enable experimental features and try clinking e.g. on clear all in new task manager/ Pin or unpin provider, open kubernetes page or kubernetes troubleshooting page/ click on progress bar in statusbar
- The dialog will be visible either instantly, or after few minutes (depending on a feature), but should take more than 5 minutes


package.json
```diff
@@ -142,8 +142,8 @@
     "@types/tar": "^6.1.13",
     "@types/tar-fs": "^2.0.4",
     "@types/validator": "^13.12.3",
-    "@typescript-eslint/eslint-plugin": "^8.29.0",
-    "@typescript-eslint/parser": "^8.29.0",
+    "@typescript-eslint/eslint-plugin": "^8.29.1",
+    "@typescript-eslint/parser": "^8.29.1",
```
axel7083: bad rebase?

packages/renderer/src/lib/statusbar/PinActions.svelte
```diff
@@ -44,10 +45,12 @@ function onWindowClick(e: Event): void {
 
 function unpin(providerId: string): void {
   window.unpinStatusBar(providerId).catch(console.error);
+  setTimeout(() => showFeedbackDialog('statusbarProviders.showProviders'), 60000);
 }
 
 function pin(providerId: string): void {
   window.pinStatusBar(providerId).catch(console.error);
+  setTimeout(() => showFeedbackDialog('statusbarProviders.showProviders'), 60000);
```
axel7083: This is acceptable for testing but not for a real implementation.

What I think would be having a dedicated class (let's name it ExperimentalFeedback) in  the backend. The PinRegistry would have to be update to add an EventEmitter. 

The ExperimentalFeedback class would register a listener to the PinRegistry, and would therefore receive events when a user pin/unpin, based on this, and on other information (like => how longs does the user had the experimental feature on? did we ask feedback recently for something else? etc.)

gastoner: Sure, that would work for pining, but we are trying to show the dialog e.g. when a user opens Kubernetes page? Do you think we can use the same approach? 

Now the user needs to have the feature on (can be a few months/days/years), but user doesn't try out the feature at all it won't show the dialog at all

axel7083: navigation registry should also have an event emitter I guess, and instead of using router.goto, the frontend should use the navigationAPI

gastoner:
Closing in favor of https://github.com/podman-desktop/podman-desktop/pull/11992

PR #11602, opened 2025-03-10
chore: remove usage of any

benoitf:
### What does this PR do?
remove usage of any

related to https://github.com/podman-desktop/podman-desktop/issues/10600

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/command-registry.ts
+++ b/packages/main/src/plugin/command-registry.ts
@@ -33,10 +33,8 @@ export interface RawCommand {
 }
 
 export interface CommandHandler {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  callback: any;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  thisArg: any;
+  callback: (...args: unknown[]) => unknown;
+  thisArg: unknown;
 }
 
 export class CommandRegistry {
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/command-registry.ts
```diff
@@ -68,9 +66,8 @@ export class CommandRegistry {
     });
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  async executeCommand<T = unknown>(commandId: string, ...args: any[]): Promise<T> {
-    const telemetryOptions = { commandId: commandId };
+  async executeCommand(commandId: string, ...args: unknown[]): Promise<unknown> {
```
jeffmaury, line 69: That may be break callers, no ? Or is the below change the only one ?

benoitf: this change is mostly internal so it's not breaking/impacting callers

benoitf: the signature in extension-api.d.ts (external view) is not changed

packages/main/src/plugin/command-registry.ts
```diff
@@ -54,7 +52,7 @@ export class CommandRegistry {
   private commandPaletteCommands = new Map<string, RawCommand[]>();
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  registerCommand(command: string, callback: (...args: any[]) => any, thisArg?: any): Disposable {
+  registerCommand(command: string, callback: (...args: any[]) => any, thisArg?: unknown): Disposable {
```
gastoner, line 55: `callback: (...args: any[]) => any` => `callback: (...args: unknown[]) => unknown` ?

gastoner (commented):
Also is changing `return` to `await` necessary in this PR?

PR #11645, opened 2025-03-12
chore: add missing Type entry for dockerode

benoitf:
### What does this PR do?
dockerode type has introduced Type
add it inside the unit test

https://github.com/podman-desktop/podman-desktop/pull/11632

### How to test this PR?

pr check 💚 


jeffmaury (approved):
LGTM.

No impact on code ? (as it seems Type is now mandatory ?

benoitf:
no impact on the code as the data is always returned by the provider engine

issue was only in the mock of the data

and we don't expose dockerode type externally

Issue #5776, opened 2024-01-31
Podman compose setup fails on Podman Desktop for windows


sockejr:
### Bug description

I tried to execute compose setup on the GUI via Settings/Resources compose. After asking for admin credentials Docker Desktop shows the screen: Failed Installing Compose and a Button Try again.
Any hints for troubleshooting are appreciated.

### Operating system

Windows10

### Installation Method

Installer from website/GitHub releases

### Version

Podman Desktop 1.7.0 / 1.7.1 , Podman 4.8.0/ 4.9.2

### Steps to reproduce

as above

### Relevant log output

```shell
where to find log output ?
```

franoisD:
same issue..

ZPOMI:
same issue on Podman Desktop for Windows(v1.7.1)

ZPOMI:
same issue on Podman Desktop for Windows(v1.7.1)

sockejr:
Any hints on how to troubleshoot are welcome

odockal:
@sockejr @ZPOMI @franoisD thanks for reporting.  
If you open developer tools in Podman Desktop (CTRL + SHIFT + I), open console, there could be an error when executing the workflow, can you please past it here. Or alternatively, open troubleshooting page and check Logs (PD status bar light bulb icon -> Logs).

jeffmaury:
Cannot reproduce on my Win11Pro machine

ZPOMI:
> @sockejr @ZPOMI @franoisD thanks for reporting. If you open developer tools in Podman Desktop (CTRL + SHIFT + I), open console, there could be an error when executing the workflow, can you please past it here. Or alternatively, open troubleshooting page and check Logs (PD status bar light bulb icon -> Logs).

Thank you very much for your work. I'm sorry for not replying to this issue in time.
I also tried to launch the developer console to check the logs when the exception occurred, but when I tried again it was successful and could not reproduce the situation.
Since the DNS is not stable where I was at the time of the problem, I suspect it was a network issue, maybe you can test it with a tool like fiddler or charles, I remember there are similar tools in the developer console to simulate bad network.
Or if there is a local log that can tell me the location, I can try to find and provide relevant files.

sockejr:
[...]

PR #11651, opened 2025-03-13
fix: removed compose /usr/local/bin in PATH check to align with kubectl

### LINKED ISSUE #10840: Compose was installed but is not on a system path please install manually
### Bug description

Installation of compose with also passing right credentials shows a warning, which is not right.

https://github.com/user-attachments/assets/9b591c45-bc2a-4b8a-9047-41d45252a82f

The same process for kubectl works good. I can install it with passing right creds and it is under `/usr/local/bin` and no warning.

### Operating system

Linux, Mac OS

### Installation Method

None

### Version

1.16.0

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

No error in console log.

bmahabirbu:
### What does this PR do?

This PR removes the composed PATH check for mac and flatpack because neither inherits the global system PATH for an accurate check.

Fixes https://github.com/podman-desktop/podman-desktop/issues/10840

### How to test this PR?

pnpm build
then run the build on macos or flatpack

### DIFF
--- a/extensions/compose/src/cli-run.ts
+++ b/extensions/compose/src/cli-run.ts
@@ -89,7 +89,7 @@ export async function installBinaryToSystem(binaryPath: string, binaryName: stri
     // Use admin privileges / ask for password for copying to /usr/local/bin
     await extensionApi.process.exec(command, args, { isAdmin: true });
     console.log(`Successfully installed '${binaryName}' binary.`);
-    if (!process.env.PATH?.includes(destinationFolder)) {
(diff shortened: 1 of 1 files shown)


jeffmaury (changes requested):
I think it should be undone only for MacOS and flatpak as it was done to fix an issue reported by user: #5776

jeffmaury (approved):
LGTM

benoitf:
@Mergifyio backport 1.17.x

benoitf (commented):
AFAIK an extension should not be aware that it runs or not with flatpak, so checking for FLATPAK_ID while it works is an anti-pattern as it links the extension to the underlying system while there is an abstraction

PR #11659, opened 2025-03-13
chore: handle merge_group event as well for pr-check GH action

benoitf:
### What does this PR do?
in case of merge_group, we don't receive pull_request event but a merge_group event
handle this usecase

### How to test this PR?

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -330,7 +330,7 @@ jobs:
 
       # Enables multi-architecture support for building multi-arch manifests
       - name: Install QEMU for Multi-Arch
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-check.yaml
```diff
@@ -330,7 +330,7 @@ jobs:
 
       # Enables multi-architecture support for building multi-arch manifests
       - name: Install QEMU for Multi-Arch
-        run: | 
+        run: |
```
benoitf:
```suggestion
        run: | 
```

jeffmaury (approved):
LGTM

Issue #11153, opened 2025-02-14
Doc: Add the details about how to uninstall Podman Desktop


shipsing:
### Is your enhancement related to a problem? Please describe

Users might want to unistall the PD application based on their development needs. We do not have any section that covers the uninstalling details in the doc. 

### Describe the solution you'd like

Add a dedicated section that covers end-to-end flow to uninstall the PD application.

PR #11689, opened 2025-03-17
docs(website): added an uninstall section in the documentation

shipsing:
### What does this PR do?
Adds an uninstall section in the documentation
### Screenshot / video of UI
PR preview: https://67d90c174ce0b4d592eca2ee--podman-desktop-pr.netlify.app/docs/uninstall

https://github.com/podman-desktop/podman-desktop/issues/11153

### How to test this PR?


website/docs/uninstall/windows-uninstall.md
```diff
@@ -0,0 +1,96 @@
...
+
+  :::note
+
+  You can find the installed binaries and plugins for Podman Desktop at `~/.local/share/containers/podman-desktop/extensions-storage/`, and you may delete them if necessary.
+
+  :::
+
+## Uninstall Podman
+
+You can delete all pods, containers, and images by removing the Podman machine.
```
axel7083: We may have, for some rare user multiple machines. To remove **all** machine, we may use `podman machine reset` instead of removing them one by one (we do not need to stop the machine when using `reset` command)

```
$: podman machine reset -h
Remove all machines

Description:
  Remove all machines, configurations, data, and cached image
```

axel7083 (approved):
Great addition! Thanks for the change :rocket:

cdrage (changes requested):
IMHO similar to rest of documentation, we should have all of this on one page and use tabs for Linux / macOS / Windows, rather than separate pages..

On the fence though and would like other people's opinions.

Other pages are condensed to tabs, but install page is already 3 separate docs.

cc @Firewall

shipsing:
Thank you, Charlie, for your suggestion. If the uninstall content does not increase in size in the future, then having a single uninstall page provides a neat and clear view. You can take another look at the content.

Firewall (approved):
Good job!

and I also prefer the content all on 1 page

cdrage (approved):

PR #11340, opened 2025-02-25
chore(electron): enable web security

axel7083:
### What does this PR do?

Enable web security for Browser Window as recommended by electron security guidelines.

### What issues does this PR fix or reference?

Fixes https://github.com/podman-desktop/podman-desktop/issues/11339

### How to test this PR?

- [x] Pipeline should be green (I guess?)

**Manually**

- `compile:current` and try podman desktop with extensions etc. everything should be working 
- try in dev mode

PR #11696, opened 2025-03-17
feat(main): use media protocol for font-family

### LINKED ISSUE #11671: "A" icon showing for Kubernetes on status bar.
### Bug description

Image

Shows "A" instead of Kubernetes logo

### Operating system

mac m1

### Installation Method

None

### Version

1.17.0

### Steps to reproduce

1. Run `pnpm watch` or compile the latest 
2. Check that k8s is running
3. See that status bar on the bottom left uses A icon instead of k8s logo.

### Relevant log output

```shell

```

### Additional context

_No response_

axel7083:
### What does this PR do?

Following https://github.com/podman-desktop/podman-desktop/pull/11340 resources with url `file://` could not be loaded (security restrictions).

I did not notice the issue with the font-family (oupsi sorry). 

### Solution proposal (short)

Instead of reverting the problematic PR, I would suggest to keep the security restriction and add a custom `media` protocol. So instead of using `file://<path-to-font>` we would use `media://<path-to-font>`, and we manually handle the media protocol, to avoid having to disable the web security.  (ref https://github.com/electron/electron/issues/23393#issuecomment-2379103959)

Instead of blindly allowing all resources using the media protocol, we can restrict to a set of extensions (here only allow `.woff2`. 

### Screenshot / video of UI

image

Fixes https://github.com/podman-desktop/podman-desktop/issues/11671

### How to test this PR?

### DIFF
--- a/packages/main/src/index.ts
+++ b/packages/main/src/index.ts
@@ -17,7 +17,7 @@
  ***********************************************************************/
 import dns from 'node:dns';
 
-import { app, ipcMain, Menu, Tray } from 'electron';
+import { app, ipcMain, Menu, net, protocol, Tray } from 'electron';
 
 import { createNewWindow, restoreWindow } from '/@/mainWindow.js';
 import type { ExtensionLoader } from '/@/plugin/extension/extension-loader.js';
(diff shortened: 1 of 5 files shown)


axel7083:
Kinda old, would still be great, but let's close until we come back on it

vancura:
Related: #12485, which is now part of #12709, but that's only design of course.

PR #11702, opened 2025-03-17
fix(renderer): create container from existing with multiple connections

### LINKED ISSUE #11697: bug(CreateContainerFromExistingImage): cannot select container engines when multiple are available
### Bug description

The dropdown is not working because the parent component is using svelte5.

Image

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. have several machines started
2. go to container page
3. select `Create`
4. select `from existing`
5. assert empty input is visible
6. assert cannot interact with input

### Relevant log output

```shell

```

### Additional context

Root cause https://github.com/podman-desktop/podman-desktop/issues/10734

axel7083:
### What does this PR do?

Following https://github.com/podman-desktop/podman-desktop/pull/11700, fixes the `CreateContainerFromExisting.svelte` page when having multiple connections

### Screenshot / video of UI

https://github.com/user-attachments/assets/6bd0ad56-d2dc-4d83-9192-dac7a4d68cdf

Fixes https://github.com/podman-desktop/podman-desktop/issues/11697

Required for
- https://github.com/podman-desktop/podman-desktop/pull/11693 (current sprint)

### How to test this PR?

**Manually**

1. assert - have at least 2 connections 
2. Go to _containers_ page 
3. Click on `Create`
4. Click on `Existing Image`
5. Select any image and run it
6. Repeat but change for another connections
7. assert the container is running on the appropriate machine

### DIFF
--- a/packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
+++ b/packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
@@ -165,6 +165,7 @@ async function gotoManageRegistries(): Promise<void> {
 }
 
 onMount(() => {
+  // Select default connection
   if (!selectedProviderConnection) {
     selectedProviderConnection = providerConnections.length > 0 ? providerConnections[0] : undefined;
   }
@@ -303,6 +304,9 @@ async function buildContainerFromImage(): Promise<void> {
 }
 
 async function searchFunction(value: string): Promise<void> {
+  // do not search for images if no connection is selected
+  if (!selectedProviderConnection) return;
+
   value = value.trim();
   const localImagesValues = await searchLocalImages(value);
   const remoteImagesValues = await searchImages(value);
--- a/packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
+++ b/packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
@@ -371,9 +371,9 @@ function onContainerConnectionChange(): void {
           resultItems={values}
           compare={sortResults}
           onChange={async (s: string): Promise<void> => {
-          validateImageName(s);
-          await resolveShortname();
-          await searchLatestTag();
+            validateImageName(s);
+            await resolveShortname();
+            await searchLatestTag();
         }}
           onEnter={onEnterOperation}
           disabled={pullFinished || pullInProgress}
--- a/packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
+++ b/packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
@@ -1,11 +1,12 @@
 <script lang="ts">
 import { faArrowCircleDown, faCircleCheck, faCog, faTriangleExclamation } from '@fortawesome/free-solid-svg-icons';
-import { Button, Checkbox, Dropdown, ErrorMessage, Tooltip } from '@podman-desktop/ui-svelte';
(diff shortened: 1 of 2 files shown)


SoniaSandler:
issue: When opening the `Select an image` page for the first time, **sometimes** all local images show up, regardless of which container engine shows up as the default in the dropdown

https://github.com/user-attachments/assets/ea80bfc3-ff6d-487e-9f02-71ce3600cb1b

axel7083:
Thanks @SoniaSandler for reporting the issue! In fact yes the `window.listImages` was called before the onMount was complete.

https://github.com/podman-desktop/podman-desktop/blob/eeb6c1fc7734790a26095ed55b382ee2eeca25dc/packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte#L167-L172

This was a problem because the Typeahead component was triggering the  search, before the default provider was selected.

And the `window.listImages` return **all images** if you do not provide a provider.

The code is fixed by adding a safeguard, as shown bellow

https://github.com/podman-desktop/podman-desktop/blob/eeb6c1fc7734790a26095ed55b382ee2eeca25dc/packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte#L306-L308

This avoid to call the listImages when we do not have finish selected the default provider.

## Testing

I added a unit test to ensure that we don't call twice the listImages (once without provider, and once with a provider).

Here is a video showing how to test that

https://github.com/user-attachments/assets/6bd0ad56-d2dc-4d83-9192-dac7a4d68cdf

**Steps to follow**

1. assert **have two machines running** (need to have two connections)
2. assert **prune all images, local registries must be empty for both** (easier to test)
3. pull image `nginx` for connection A
4. pull image `hello-world` for connection B
5. go to create container
6. assert only selected connection show local images

packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
```diff
@@ -343,23 +362,25 @@ async function onEnterOperation(): Promise<void> {
...
+          name="imageName"
+          placeholder="Select or enter an image to run"
+          onInputChange={searchFunction}
+          resultItems={values}
+          compare={sortResults}
+          onChange={async (s: string): Promise<void> => {
           validateImageName(s);
           await resolveShortname();
           await searchLatestTag();
         }}
```
SoniaSandler, line 377: nit: shouldn't this part be indented? (inside the onChange)

SoniaSandler (approved):
LGTM, I haven't been able to reproduce the problem again. Thanks for the fix!

packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
axel7083: probably, biome is not able to parse this, that's why we don't get any warning

packages/renderer/src/lib/container/CreateContainerFromExistingImage.svelte
```diff
@@ -343,23 +362,25 @@ async function onEnterOperation(): Promise<void> {
...
+          id="imageName"
+          name="imageName"
+          placeholder="Select or enter an image to run"
+          onInputChange={searchFunction}
+          resultItems={values}
+          compare={sortResults}
+          onChange={async (s: string): Promise<void> => {
           validateImageName(s);
           await resolveShortname();
           await searchLatestTag();
```
axel7083:
```suggestion
            validateImageName(s);
            await resolveShortname();
            await searchLatestTag();
```

PR #11717, opened 2025-03-18
fix: remove docker compatibility from statusbar

### LINKED ISSUE #11215: Remove Docker compatibily from the statusbar

During the Epic that this issue is part of we will add better in context messages for users to configure and set up their Docker Comptibility. As such the item in the statusbar no longer be needed. In the past it was only added because we needed a quick place to put it. 

Remove Docker Compaitibility item from the statusbar completely

dgolovin:
### What does this PR do?

PR removes docker compatibility item from status bar.

### Screenshot / video of UI

image

Fix #11215 

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -1420,19 +1420,6 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
     const compatibilityEnabled = socketCompatibilityMode.isEnabled();
     extensionApi.context.setValue(PODMAN_DOCKER_COMPAT_ENABLE_KEY, compatibilityEnabled);
 
-    // Create a status bar item to show the status of compatibility mode as well as
-    // create a command so when you can disable / enable compatibility mode
-    const statusBarItem = extensionApi.window.createStatusBarItem();
-    statusBarItem.text = 'Docker Compatibility';
-    statusBarItem.command = 'podman.socketCompatibilityMode';
-
-    // Use tooltip text from class
-    statusBarItem.tooltip = socketCompatibilityMode.tooltipText();
-
-    statusBarItem.iconClass = 'fa fa-plug';
-    statusBarItem.show();
-    extensionContext.subscriptions.push(statusBarItem);
-
     // Create a modal dialog to ask the user if they want to enable or disable compatibility mode
     const command = extensionApi.commands.registerCommand('podman.socketCompatibilityMode', async () => {
       // Manually check to see if the socket is disguised (this will be called when pressing the status bar item)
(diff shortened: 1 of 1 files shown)


extensions/podman/packages/extension/src/extension.ts
```diff
@@ -1420,19 +1420,6 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
...
-
-    // Use tooltip text from class
-    statusBarItem.tooltip = socketCompatibilityMode.tooltipText();
-
-    statusBarItem.iconClass = 'fa fa-plug';
-    statusBarItem.show();
-    extensionContext.subscriptions.push(statusBarItem);
-
     // Create a modal dialog to ask the user if they want to enable or disable compatibility mode
     const command = extensionApi.commands.registerCommand('podman.socketCompatibilityMode', async () => {
```
benoitf, line 1424: question: if we remove the status bar entry calling the command 'podman.socketCompatibilityMode'
is that 'podman.socketCompatibilityMode' command is called from other places ? if not command should be removed

benoitf:
@codecov-ai-reviewer review

ghost:
No changes requiring review at this time.

extensions/podman/packages/extension/src/extension.ts
dgolovin: There is Docker compatibility settings and onDidChangeConfiguration listener calls this method indirectly.

dgolovin: It seems not used anywhere, I'll verify it.

dgolovin: The command is still available from through command palette and triggers the same workflow. If the plan is to use this command from somewhere else it should stay. 
CC: @Firewall

Firewall: I don't think it is needed to remove it from the command palette.

Firewall (approved):
👍 Works as expected to me.

extensions/podman/packages/extension/src/extension.ts
cdrage: That command is still used in other places such as the Podman Mac helper notification.

benoitf (approved):

extensions/podman/packages/extension/src/extension.ts
simonrey1: In my opinion, we should delete this command because it is only available in the command palette and it is unclear what is the difference with the Preferences > Docker Compatibility page.

Screenshot 2025-11-14 at 11 46 16
Screenshot 2025-11-14 at 11 46 30

simonrey1: Update: this command is still used here, I did not find this occurence before:

Screenshot 2025-11-24 at 10 00 09


extensions/podman/packages/extension/src/utils/notifications.ts l.161

PR #11722, opened 2025-03-18
chore: kind extension - use configuration name when deploying ingress

### LINKED ISSUE #11469: Config file and form options conflict in Kind cluster creation

cdrage:
chore: kind extension - use configuration name when deploying ingress

### What does this PR do?

When providing a configuration file and deploying an ingress, we should
be using the name from the configuration file regardless of other
options added above.

Closes https://github.com/podman-desktop/podman-desktop/issues/11469

### How to test this PR?

1. Deploy with yaml:

```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: app-1-cluster
```

2. See that your contour ingress is added to `kind-app-1-cluster`. You
   can check by doing `kubectl get pods -A` and see that contour has
   been deployed.

### DIFF
--- a/extensions/kind/src/create-cluster.spec.ts
+++ b/extensions/kind/src/create-cluster.spec.ts
@@ -29,6 +29,7 @@ import { getKindPath, getMemTotalInfo } from './util';
 vi.mock('node:fs', () => ({
   promises: {
     writeFile: vi.fn(),
+    readFile: vi.fn(),
     mkdtemp: vi.fn(),
     rm: vi.fn(),
   },
@@ -87,7 +88,7 @@ test('expect error is cli returns non zero exit code', async () => {
 
 test('expect cluster to be created', async () => {
   vi.mocked(getKindPath).mockReturnValue('/kind/path');
-  (extensionApi.process.exec as Mock).mockReturnValue({} as extensionApi.RunResult);
+  (extensionApi.process.exec as Mock).mockResolvedValue({} as extensionApi.RunResult);
   await createCluster({}, '', telemetryLoggerMock);
   expect(telemetryLogUsageMock).toHaveBeenNthCalledWith(
     1,
@@ -103,7 +104,7 @@ test('expect cluster to be created', async () => {
 });
 
 test('expect cluster to be created using config file', async () => {
-  (extensionApi.process.exec as Mock).mockReturnValue({} as extensionApi.RunResult);
+  vi.mocked(extensionApi.process.exec).mockResolvedValue({} as extensionApi.RunResult);
   const logger = {
     log: vi.fn(),
     error: vi.fn(),
@@ -120,7 +121,7 @@ test('expect cluster to be created using config file', async () => {
 });
 
 test('expect cluster to not call setupIngressController function when supplying config file and ingress is set to no', async () => {
-  (extensionApi.process.exec as Mock).mockReturnValue({} as extensionApi.RunResult);
+  vi.mocked(extensionApi.process.exec).mockResolvedValue({} as extensionApi.RunResult);
   const logger = {
     log: vi.fn(),
     error: vi.fn(),
@@ -143,7 +144,7 @@ test('expect cluster to not call setupIngressController function when supplying 
 });
 
 test('expect cluster to call ingress controller setup when ingress is set to yes AND a config is supplied', async () => {
-  (extensionApi.process.exec as Mock).mockReturnValue({} as extensionApi.RunResult);
+  vi.mocked(extensionApi.process.exec).mockResolvedValue({} as extensionApi.RunResult);
   const logger = {
     log: vi.fn(),
     error: vi.fn(),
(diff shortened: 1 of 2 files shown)


benoitf:
@codecov-ai-reviewer review

amisskii:
After the previous discussion under PR: https://github.com/podman-desktop/podman-desktop/pull/11555, I think we need to determine which parameters should be ignored when using a kind config file. In particular, I believe the  Provider Type parameter should not be ignored, as it is not part of the kind config file

cdrage:
> After the previous discussion under PR: #11555, I think we need to determine which parameters should be ignored when using a kind config file. In particular, I believe the Provider Type parameter should not be ignored, as it is not part of the kind config file

Just an update from our slack discussion:

- Created a UX issue for determining what is used vs not used when providing a config file: https://github.com/podman-desktop/podman-desktop/issues/11728
- Added feature request to respect the Provider Type option when deploying with kind: https://github.com/podman-desktop/podman-desktop/issues/11729

extensions/kind/src/create-cluster.spec.ts
```diff
@@ -118,6 +119,104 @@ test('expect cluster to be created using config file', async () => {
   expect(extensionApi.kubernetes.createResources).not.toBeCalled();
 });
 
+test('expect cluster to not call setupIngressController function when supplying config file and ingress is set to no', async () => {
+  (extensionApi.process.exec as Mock).mockReturnValue({} as extensionApi.RunResult);
```
benoitf: vi.mocked(extensionApi.process.exec).mockReturnValue...

extensions/kind/src/create-cluster.spec.ts
```diff
@@ -118,6 +119,104 @@ test('expect cluster to be created using config file', async () => {
...
+    expect.objectContaining({ provider: 'docker' }),
+  );
+  expect(telemetryLogErrorMock).not.toBeCalled();
+
+  // Expect create resources to NOT be called
+  expect(extensionApi.kubernetes.createResources).not.toBeCalled();
+});
+
+test('expect cluster to call ingress controller setup when ingress is set to yes AND a config is supplied', async () => {
+  (extensionApi.process.exec as Mock).mockReturnValue({} as extensionApi.RunResult);
```
benoitf: vi.mocked(...)

extensions/kind/src/create-cluster.spec.ts
```diff
@@ -118,6 +119,104 @@ test('expect cluster to be created using config file', async () => {
...
+  // Expect create resources to be called
+  expect(extensionApi.kubernetes.createResources).toBeCalled();
+
+  // Expect createResources to be called with `kind-kind` as the namespace since we are using a "fake"
+  // config file and there was no name supplied, so it should be using the default
+  expect(extensionApi.kubernetes.createResources).toBeCalledWith('kind-kind', expect.anything());
+});
+
+test('expect cluster to use "name: foobar" within the yaml file when supplying a config file', async () => {
+  (extensionApi.process.exec as Mock).mockReturnValue({} as extensionApi.RunResult);
```
benoitf: vi.mocked(...)

extensions/kind/src/create-cluster.ts
```diff
@@ -232,3 +240,22 @@ export async function createCluster(
     await fs.promises.rm(tmpDirectory, { recursive: true });
   }
 }
+
+// Function reads a path name, opens the yaml file and returns "name" from the kind configuration file
+// if no name is provided, we just use 'kind' which is the default.
+function getClusterNameFromConfigFile(configFilePath: string): string {
+  const configFile = fs.readFileSync(configFilePath, 'utf8');
```
benoitf: issue: should not use sync call but readFile from fs/promises

gastoner (approved):
LGTM on Windows 
+ 1 to coments from Florent

cdrage:
@benoitf Updated in a new commit, switched to using `vi.mocked` for all calls in the test suite (noticed others weren't converted yet either).

Also using readFile instead of readSyncFile.

extensions/kind/src/create-cluster.spec.ts
```diff
@@ -86,7 +88,7 @@ test('expect error is cli returns non zero exit code', async () => {
 
 test('expect cluster to be created', async () => {
   vi.mocked(getKindPath).mockReturnValue('/kind/path');
-  (extensionApi.process.exec as Mock).mockReturnValue({} as extensionApi.RunResult);
+  (extensionApi.process.exec as Mock).mockResolvedValue({} as extensionApi.RunResult);
```
benoitf: vi.mocked(extensionApi.process.exec)

benoitf (approved):
nit: there is still one not replaced

cdrage:
> nit: there is still one not replaced

Added in a new commit, thanks for the reviews!

benoitf:
https://github.com/Mergifyio backport 1.17.x

PR #11726, opened 2025-03-18
chore: enable svelte/require-store-reactive-access rule

### LINKED ISSUE #11513: Enable svelte/require-store-reactive-access rule and update impacted files
https://github.com/podman-desktop/podman-desktop/blob/8dd7b6de9f8e3292cb1648548689c3bbfbfed18b/eslint.config.mjs#L275

cdrage:
chore: enable svelte/require-store-reactive-access rule

### What does this PR do?

Enables the rule by removing the `off` line in eslint.config.mjs 

### Screenshot / video of UI

N/A

Closes https://github.com/podman-desktop/podman-desktop/issues/11513

### How to test this PR?

N/A

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -269,7 +269,6 @@ export default [
       'sonarjs/no-alphabetical-sort': 'off',
       'svelte/no-useless-children-snippet': 'off',
       'svelte/require-each-key': 'off',
-      'svelte/require-store-reactive-access': 'off',
       'svelte/no-reactive-literals': 'error',
     },
   },
--- a/packages/renderer/src/lib/ui/NavItem.svelte
+++ b/packages/renderer/src/lib/ui/NavItem.svelte
@@ -26,7 +26,7 @@ const navItems: Writable<number> = getContext('nav-items');
 $: tooltipText = counter ? `${tooltip} (${counter})` : tooltip;
 
 onMount(() => {
-  inSection = navItems !== undefined;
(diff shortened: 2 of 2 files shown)


cdrage:
@benoitf 

I get the same result either removing `'svelte/require-store-reactive-access': 'off'` or switching it to `'svelte/require-store-reactive-access': 'error',`

What's the correct way?

I've noticed `'svelte/no-reactive-literals': 'error',` in the list, shouldn't that be removed rather than switch to `error`?

axel7083:
> I get the same result either removing 'svelte/require-store-reactive-access': 'off' or switching it to 'svelte/require-store-reactive-access': 'error',

depends, if the rule is in the recommended rules, it is included, but some rules may be omitted, so they need to be defined explicitely

cdrage:
> > I get the same result either removing 'svelte/require-store-reactive-access': 'off' or switching it to 'svelte/require-store-reactive-access': 'error',
> 
> depends, if the rule is in the recommended rules, it is included, but some rules may be omitted, so they need to be defined explicitely

Ahhh thanks @axel7083 I see that https://sveltejs.github.io/eslint-plugin-svelte/rules/require-store-reactive-access/ says: `This rule is included in "plugin:svelte/recommended".` so I'll leave it like that.

benoitf:
@axel7083 already replied 👍

gastoner (approved):

benoitf (approved):

PR #11750, opened 2025-03-19
test: new cli e2e tests

### LINKED ISSUE #10696: CLI Tools Page E2E Test implementation
CLI tools page offers a functionality that allow the user to install and maintain some tools like compose, kind, etc.

Implement E2E test case that will verify that the functionality work as intended.

cbr7:
### What does this PR do?
Adds e2e tests for cli tools page tool installation/upgrade/downgrade/uninstall

https://github.com/podman-desktop/podman-desktop/issues/10696#issuecomment-2720287586


tests/playwright/src/model/pages/cli-tools-page.ts
```diff
@@ -97,11 +100,109 @@ export class CLIToolsPage extends SettingsPage {
         console.log(`Dialog for tool ${toolName} was not visible. Proceeding.`);
       }
 
+      await playExpect
+        .poll(async () => await this.getCurrentToolVersion(toolName), { timeout: timeout })
+        .toContain(version);
```
odockal, line 105: If  the version parameter is empty we should not check if version contains `''`.

cbr7:
https://github.com/podman-desktop/e2e/actions/runs/13945610647 run from azure cicd.

tests/playwright/src/specs/cli-tools-smoke.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+import { isMac } from '../utility/platform';
+import { waitForPodmanMachineStartup } from '../utility/wait';
+
+let settingsBar: SettingsBar;
+let cliToolsPage: CLIToolsPage;
+
+test.beforeAll(async ({ runner, page, welcomePage, navigationBar }) => {
+  runner.setVideoAndTraceName('cli-tools-e2e');
+  await welcomePage.handleWelcomePage(true);
+  await waitForPodmanMachineStartup(page);
```
odockal, line 36: we do not need machine to be up, right?

odockal (changes requested):
Just one comment. Also, we will create follow up to tackle also compse, kubectl, right?

tests/playwright/src/specs/cli-tools-smoke.spec.ts
cbr7: I don't think so, no, can probably be removed.

cbr7: @odockal a followup issue can certainly be created if the goal is to test the tools themselves, but if the goal is to test the install/uninstall functionality adding more tools is not required, in fact I would argue this should be avoided, running the tests for all tools on the page will very quickly exhaust github rate limit for the day.

So please let me know if you want this, because expanding the tests to run for the other tools should ve fairly easy by parametrizing the tool name as we do in other suites.

odockal: Good point. We still would need to address problems that happens for particular tools like compose, or kubectl. So as I am for to have test for one tool, it would be great to have a way to test others as well. We do not have to do it at once and always. How about we can influence that tool is installed via env. var. ? Let's say default is always ie. compose or kubectl, but then, if we set the particular name, that one tool will be installed, or there will be option to test all.
```
const tools = process.env.CLI_TOOLS;
const testedTools = tools === 'all' ? ['kind','docker-compose','kubectl'] : tools !== undefined ? [tools] : [''docker-compose];
for (let tool in testedTools) {
  test(`Testing installation of ${tool}`, () => {
   ...
})
...
}
```
Roughly speaking of course. We could also change which tool to test depending on external circumstances, like day of the week. In any case, if this scenario, will run tens of times a day in any case, on pr check, windows nightly, internally. I am not sure if we want to make it a smoke.

odockal (approved):

PR #11604, opened 2025-03-10
test: k8s services e2e test

amisskii:
### What does this PR do?
Adds a new e2e test for K8s services.

### What issues does this PR fix or reference?

https://github.com/podman-desktop/podman-desktop/issues/10698

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature

PR #11752, opened 2025-03-19
test: k8s ingress e2e test

### LINKED ISSUE #10695: Implement E2E test scenario to cover Routes Kubernetes Page
Implement E2E test case to cover scenario where we verify Kubernetes Routes Page functionality.

amisskii:

### What does this PR do?

Adds k8s Ingress resource e2e test

https://github.com/podman-desktop/podman-desktop/issues/10695

### How to test this PR?

check the workflow run: https://github.com/podman-desktop/podman-desktop/actions/runs/14130113327

### DIFF
--- a/tests/playwright/resources/kubernetes/test-ingress-resource.yaml
+++ b/tests/playwright/resources/kubernetes/test-ingress-resource.yaml
@@ -16,25 +16,19 @@
 #  * SPDX-License-Identifier: Apache-2.0
 #  ***********************************************************************/
 
-apiVersion: apps/v1
-kind: Deployment
+apiVersion: networking.k8s.io/v1
+kind: Ingress
 metadata:
-  name: test-deployment-resource
-  labels:
-    app: test-deployment-resource
+  name: test-ingress-resource
 spec:
-  replicas: 3
-  selector:
-    matchLabels:
-      app: test-deployment-resource
-  template:
-    metadata:
-      labels:
-        app: test-deployment-resource
-    spec:
-      containers:
-      - name: test-deployment-resource
-        image: ghcr.io/podmandesktop-ci/nginx
-        ports:
-        - containerPort: 80
- 
+  rules:
+  - http:
+      paths:
+      - pathType: Prefix
+        path: /
+        backend:
+          service:
+            name: test-service-resource
+            port:
+              number: 8080
+  
--- a/tests/playwright/src/model/pages/details-page.ts
+++ b/tests/playwright/src/model/pages/details-page.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2024 Red Hat, Inc.
+ * Copyright (C) 2024-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/tests/playwright/src/specs/kubernetes-networking.spec.ts
+++ b/tests/playwright/src/specs/kubernetes-networking.spec.ts
@@ -26,11 +26,10 @@ import { createKindCluster, deleteCluster } from '../utility/cluster-operations'
 import { test } from '../utility/fixtures';
 import {
   checkKubernetesResourceState,
-  configurePortForwarding,
   createKubernetesResource,
   deleteKubernetesResource,
+  monitorPodStatusInClusterContainer,
   verifyLocalPortResponse,
-  verifyPortForwardingConfiguration,
 } from '../utility/kubernetes';
 import { ensureCliInstalled } from '../utility/operations';
 import { waitForPodmanMachineStartup } from '../utility/wait';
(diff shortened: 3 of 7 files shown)


amisskii:
windows workflow run https://github.com/podman-desktop/e2e/actions/runs/139501986
linux workflow run: https://github.com/podman-desktop/podman-desktop/actions/runs/13950186558

tests/playwright/resources/kubernetes/test-ingress-resource.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: networking.k8s.io/v1
```
odockal, line 19: copyright?

odockal:
@amisskii In general I think the PR is quite good. Windows run is not available (404). It would be great that you describe the scenario is performing so that @cdrage or @jeffmaury can confirm the validity of the test case.

amisskii:
The test scenario is designed to reproduce a basic Kubernetes networking workflow, which includes a deployment, a service that forwards all requests to the deployment's pods (the service E2E test was implemented in this PR: https://github.com/podman-desktop/podman-desktop/pull/11604), and an ingress resource that manages external access to the service within the cluster.

This PR focuses on two main objectives:
**Ingress page verification**: Ensuring the ingress resource is listed on the page and checking its status.

**Ingress functionality**: Verifying that the ClusterIP service is accessible after applying the ingress resource, and confirming that the application is available on `localhost:9090`. Port `9090` is by default set to expose `port 80` on which ingress resource is running. 
 To prevent test instability, a check for the ingress controller pods has been introduced. This check repeatedly queries the status of the Contour ingress controller pods and waits until they are ready. This step also helps detect potential issues with the ingress controller.
 
 references: https://kind.sigs.k8s.io/docs/user/ingress/ https://spacelift.io/blog/kubernetes-service

tests/playwright/src/specs/kubernetes-networking.spec.ts
```diff
@@ -39,25 +38,27 @@ const CLUSTER_NAME: string = 'kind-cluster';
 const CLUSTER_CREATION_TIMEOUT: number = 300_000;
 const KIND_NODE: string = `${CLUSTER_NAME}-control-plane`;
 const RESOURCE_NAME: string = 'kind';
-const KUBERNETES_CONTEXT = `kind-${CLUSTER_NAME}`;
-const KUBERNETES_NAMESPACE = 'default';
-const DEPLOYMENT_NAME = 'test-deployment-resource';
-const SERVICE_NAME = 'test-service-resource';
+const KUBERNETES_CONTEXT: string = `kind-${CLUSTER_NAME}`;
```
jeffmaury, line 41: Why do you need to add type ?

amisskii: It's just the pattern I follow when defining constants

amisskii:
The latest workflow runs: 
Windows https://github.com/podman-desktop/e2e/actions/runs/14130122726
Linux https://github.com/podman-desktop/podman-desktop/actions/runs/14130113327

tests/playwright/src/model/pages/details-page.ts
```diff
@@ -58,4 +60,25 @@ export abstract class DetailsPage extends BasePage {
       return this;
     });
   }
+
+  async executeCommandInTerminal(command: string): Promise<void> {
+    await playExpect(this.terminal).toBeVisible();
```
odockal: You can add terminal tab activation into this function as it cannot be performed without it.

tests/playwright/src/model/pages/details-page.ts
```diff
@@ -58,4 +60,25 @@ export abstract class DetailsPage extends BasePage {
...
+    await this.terminal.click();
+    await this.page.keyboard.insertText(command);
+    await this.page.keyboard.press('Enter');
+  }
+
+  async retrieveTerminalLog(): Promise<string[]> {
+    await playExpect(this.terminal).toBeVisible();
```
odockal: Same as above.

tests/playwright/src/utility/kubernetes.ts
```diff
@@ -268,3 +268,36 @@ export async function waitForKubernetesResourceAvailability(
...
+  podName: string,
+  podStatus: string,
+  replicas: string,
+  timeout: number = 160_000,
+): Promise<void> {
+  const navigationBar = new NavigationBar(page);
+  const containersPage = await navigationBar.openContainers();
+  await playExpect.poll(async () => containersPage.getContainerRowByName(containerName)).toBeTruthy();
+  const detailsPage = await containersPage.openContainersDetails(containerName);
+  await detailsPage.activateTab('Terminal');
```
odockal: You can move the tab switching into executecommand function.

tests/playwright/src/model/pages/details-page.ts
```diff
@@ -58,4 +60,25 @@ export abstract class DetailsPage extends BasePage {
...
+    // Wait for a short while to ensure that the terminal output is fully loaded
+    await this.page.waitForTimeout(2000);
+    const terminalOutputRows = await this.terminal.locator('div:has(span)').all();
```
odockal: this locator is very fragile. We need to improve aria in the page itself. Please open follow up issue for this.

odockal (changes requested):

tests/playwright/src/model/pages/details-page.ts
amisskii: https://github.com/podman-desktop/podman-desktop/issues/11953

cbr7: I don't think this is required, using the following selectors should work, at least they do in other pages.
image

amisskii: @cbr7 But then, I think this map will iterate through all empty divs 

```
await Promise.all(
      terminalOutputRows.map(async row => {
        log.push((await row.textContent()) ?? '');
      }),
    );
```

cbr7: Why specifically would you need that?

tests/playwright/src/model/pages/details-page.ts
```diff
@@ -46,6 +47,7 @@ export abstract class DetailsPage extends BasePage {
     this.closeButton = this.breadcrumb.getByRole('button', { name: 'Close' });
     this.backLink = this.breadcrumb.getByRole('link', { name: 'Back' });
     this.pageName = this.breadcrumb.getByRole('region', { name: 'Page Name' });
+    this.terminal = this.tabContent.locator('.xterm-screen');
```
cbr7: I actually heavily disagree that this should be here at all, DetailsPage is an abstract class that is inherited in many places, not all of those places have a terminal, this implementation could generate errors in other test suites.

e.g. terminal is not a feature of image details page, pods details page or volume details page.

cbr7 (changes requested):
Functionality should not be implemented in a parent class if not common to all children classes.

tests/playwright/src/model/pages/details-page.ts
amisskii: To capture the terminal output as an array, where each entry represents one row in the terminal, so that I can make assertions for specific rows such as verifying the presence of a pod's name, status, and replicas in the output

cbr7: that can be done with `await playExpect(terminalContent).toContainText(textToFindHere)` you need not map or get the rows one by one, that is simply not scalable since the terminal could very well have very many rows.

amisskii: But I don't think it's a problem to use `this.terminalContent = this.tabContent.locator('.xterm-rows').locator('div:has(span)').all()`, especially if we can implement an aria-label for the div elements containing the span (with text inside).

cbr7: It's not a problem in the sense that yes it could work, but why do you need to? It's simply not a use case that makes sense. There is no reason to parse the rows one by one, in fact it's very wasteful resources wise and you will run into problems if the terminal is big enough, regardless of if you are using aria-label or xpath-ing your way to the span element.

amisskii: In this particular case, I want to parse the log output into rows to verify that pods with a specific name have the expected status and replicas. My proposal is to move the functionality from `executeCommandInTerminal` and `retrieveTerminalLog` to the function `monitorPodStatusInClusterContainer` in `utility/kubernetes.ts`. 
image

cbr7: I still don't see what is preventing you from doing that using the toContainText as above, does `await playExpect(terminalContent).toContainText("<podNameHere> <expectedCloneCountHere> <expectedPodStateHere>");` not work?

tests/playwright/src/utility/kubernetes.ts
```diff
@@ -254,3 +254,45 @@ export async function verifyLocalPortResponse(forwardAddress: string, responseMe
...
 }
+
+export async function monitorPodStatusInClusterContainer(
+  page: Page,
+  containerName: string,
+  command: string,
+  timeout: number = 160_000,
+): Promise<void> {
+  const navigationBar = new NavigationBar(page);
+  const containersPage = await navigationBar.openContainers();
```
cbr7, line 265: Please assert that navigation to Containers Page is complete after this line.

cbr7 (approved):

odockal (approved):

Issue #11274, opened 2025-02-20
Support for switching Kubernetes namespace


deboer-tim:
### Is your enhancement related to a problem? Please describe

Users need a way to view and easily change the Kubernetes namespace from every relevant Kubernetes page.

### Describe the solution you'd like

Implement the design from #11273. The selection will not modify Kube config. It needs to be consistent across pages (e.g. if you change the namespace in Pods and then switch to Services, or go to Images and back, you should still be in that namespace), but does not need to be persisted across restarts.

PR #11753, opened 2025-03-19
chore: api to change kubernetes namespace

deboer-tim:
### What does this PR do?

Adds new function to allow the user to change the Kubernetes namespace on the current context and updates state, but without modifying the kubeconfig. Logic is similar to what happens in the refresh() method, and exposed to renderer.

Existing function on the TestKubernetesClient renamed due to conflict and test confirms that the namespace is updated and event fired without modifying kubeconfig.

This will support a UI allowing the user to quickly switch namespaces from Kubernetes resource pages.

### Screenshot / video of UI

N/A

First part of #11274.

### How to test this PR?

UI will be updated in future PR, just PR checks and code review for now.

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -2648,6 +2648,10 @@ export class PluginSystem {
       return kubernetesClient.getCurrentNamespace();
     });
 
+    this.ipcHandle('kubernetes-client:setCurrentNamespace', async (_listener, namespace: string): Promise<void> => {
+      return kubernetesClient.setCurrentNamespace(namespace);
+    });
+
     this.ipcHandle(
       'kubernetes-client:deleteContext',
       async (_listener, contextName: string): Promise<KubernetesContext[]> => {
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
@@ -117,7 +117,7 @@ class TestKubernetesClient extends KubernetesClient {
     return super.createWatchObject();
   }
 
-  public setCurrentNamespace(namespace: string): void {
+  public setInitialNamespace(namespace: string): void {
     this.currentNamespace = namespace;
   }
 
@@ -258,7 +258,7 @@ class TestKubernetesClient extends KubernetesClient {
 function createTestClient(namespace?: string): TestKubernetesClient {
   const client = new TestKubernetesClient(apiSender, configurationRegistry, fileSystemMonitoring, telemetry);
   if (namespace) {
-    client.setCurrentNamespace(namespace);
+    client.setInitialNamespace(namespace);
   }
   return client;
 }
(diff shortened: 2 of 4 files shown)


feloy:
Reading the code, the changes seem safe to me for all the experimental mode, as you are calling its entrypoint with the in-memory modified kubeconfig (`contextsState.update(newConfig)`).

In kubernetes-client, there are many places where `this.kubeConfig` is accessed directly, and as it is not modified in `setCurrentNamespace`, I expect there will be problems in these cases.

Still in kubernetes-client, there is a problem in one place  where the kubeconfig file is read again from file (in `syncResources`), instead of using the one saved. 

To make it safer, I would instead call `refresh` from `setCurrentNamespace`, and add an optional `kubeconfig` parameter to `refresh`, which would be considered instead of the file when specified (not covering the `syncResources` problem).

Thinking again about this, I'm not very comfortable modifying the current context by changing its namespace. I would expect that a new context is created by duplicating the existing one, and setting the desired namespace for this new one, and switching to this new one.

deboer-tim:
I received notification for the original text so didn't see this until now. My concern is that we don't want to 'change the context' or affect the kubeconfig file either so I'm not sure how we'd duplicate it, but I'll take a look at this and using refresh() tomorrow.

In case anyone wants to test, I've put up a draft PR with the UI for this in #11756.

jeffmaury:
I am a little bit concerned about having the change done only in memory as it introduces a difference of behavior compared to other tools (VScode extension,....)

jeffmaury (commented):
See https://github.com/podman-desktop/podman-desktop/pull/11753#issuecomment-2739463227

feloy:
> I am a little bit concerned about having the change done only in memory as it introduces a difference of behavior compared to other tools (VScode extension,....)

For me, it is an equivalent of the flag `--namespace` of `kubectl`, which places the command in a different namespace, without having to change the current namespace

deboer-tim:
> > I am a little bit concerned about having the change done only in memory as it introduces a difference of behavior compared to other tools (VScode extension,....)
> 
> For me, it is an equivalent of the flag `--namespace` of `kubectl`, which places the command in a different namespace, without having to change the current namespace

Ditto, and I confirmed that other kube tools do this too.

FWIW I tried the other two options first: modifying kubeconfig directly, or 'modifying kubeconfig but not saving'. From my user perspective the current approach felt simpler and more consistent with kubectl, a quick switch without making anything permanent. I didn't want the 'default namespace' in kubeconfig to change, and if kubeconfig was changed in memory (Settings > Kubernetes) but not saved it was even more confusing.

feloy:
> > > I am a little bit concerned about having the change done only in memory as it introduces a difference of behavior compared to other tools (VScode extension,....)
> > 
> > 
> > For me, it is an equivalent of the flag `--namespace` of `kubectl`, which places the command in a different namespace, without having to change the current namespace
> 
> Ditto, and I confirmed that other kube tools do this too.
> 
> FWIW I tried the other two options first: modifying kubeconfig directly, or 'modifying kubeconfig but not saving'. From my user perspective the current approach felt simpler and more consistent with kubectl, a quick switch without making anything permanent. I didn't want the 'default namespace' in kubeconfig to change, and if kubeconfig was changed in memory (Settings > Kubernetes) but not saved it was even more confusing.

OK, I agree with this idea. 

My concern is that changing the kubeconfig in memory this way, and updating the informers with it (and not the `kubernetes-client.kubeconfig`, will change the lists of resources, until when you change back to the original namespace. But at the same time, if you are in the resources list (let's say pods list), and you create a pod, it will be created on the original namespace I guess.

deboer-tim:
> But at the same time, if you are in the resources list (let's say pods list), and you create a pod, it will be created on the original namespace I guess.

When you apply yaml, we use the _current_ namespace if there isn't already one set on the resource, so:
- user connects with namespace A in kubeconfig
- user switches to namespace B
- apply yaml with resource set to namespace C -> goes to namespace C
- apply yaml with no namespace in resource -> goes to namespace B (not A)

I think this behaviour is correct / what the user would expect.

feloy:
> > But at the same time, if you are in the resources list (let's say pods list), and you create a pod, it will be created on the original namespace I guess.
> 
> When you apply yaml, we use the _current_ namespace if there isn't already one set on the resource, so:
> 
> * user connects with namespace A in kubeconfig
> * user switches to namespace B
> * apply yaml with resource set to namespace C -> goes to namespace C
> * apply yaml with no namespace in resource -> goes to namespace B (not A)
> 
> I think this behaviour is correct / what the user would expect.

OK, I didn't realise by reading the code that the `this.currentNamespace` was used, and not the complete `this.kubeconfig`.

LGTM

feloy (approved):

deboer-tim:
> OK, I didn't realise by reading the code that the `this.currentNamespace` was used, and not the complete `this.kubeconfig`.
> 
> LGTM

Thanks. FWIW I've been looking through your comments from yesterday. There is overlap with refresh() but I couldn't find any refactoring that would share the right bits without making it more complicated for both paths.

I agree on syncResources() - it needs a copy of the kubeconfig b/c it could be applying to a different context, but safer and better to use an in-memory copy instead of re-reading from file. I'll open another PR to fix that.

gastoner (approved):
Codewise LGTM

Issue #11744, opened 2025-03-19
Introduce new resource type VirtualMachine


jeffmaury:
### Is your enhancement related to a problem? Please describe

A new resource type should be added to the core for handing virtual machines. This may also concern SSH capability

### Describe the solution you'd like

Extend core and backend to support such resource

### Steps

- [x] add `registerVmProviderConnection` method to `provider-impl` (#11755)
- [x] add `setVmProviderConnectionFactory` method to `provider-impl`
- [x] add methods to access vm providers through `window....` methods

feloy:
closing as done

PR #11755, opened 2025-03-19
feat: register VM provider connection

feloy:

### What does this PR do?

Add a method `registerVmProviderConnection` to `provider-impl` to register VM provider connection (to be exposed later in `extension-api.t.ts` in `Provider` interface)

### Screenshot / video of UI

N/A

Part of #11744 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/provider-registry.ts
+++ b/packages/main/src/plugin/provider-registry.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2022-2024 Red Hat, Inc.
+ * Copyright (C) 2022-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


benoitf (approved):
nit: the sourcery comment looks valid

benoitf (commented):
nit: before merging, new tests need to be added

packages/main/src/plugin/provider-registry.ts
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
```
axel7083, line 1: remark (non-blocking): the `provider-registry.ts` is 1400 lines long, this will become harder and harder to maintain, and more and more complex to test.

The `provider-registry.spec` is 1700 lines long.

When introducing new changes to files like the container registry, and the provider registry, we might want to question ourself, if we should not split the logic over smaller files.

axel7083 (approved):
LGTM codewise, I still have some reserve for making such big file even bigger

PR #11771, opened 2025-03-20
refactor(frontend): using string as key for proxy dropdown

### LINKED ISSUE #10734: bug(Dropdown): using state as value break the selection

axel7083:
### What does this PR do?

To be able to fixes https://github.com/podman-desktop/podman-desktop/issues/10734 we need to change all usage of the dropdown component to use a string as a key.

The `PreferencesProxiesRendering.svelte` is using number, so adding a Map<ProxyState, string> to get a corresponding unique label, and use it as key.

### Screenshot / video of UI

N/A

Required for

### How to test this PR?

> :information_source: I added tests to cover `PreferencesProxiesRendering`

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.spec.ts
@@ -1,0 +1,119 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+
+import { Dropdown } from '@podman-desktop/ui-svelte';
+import { fireEvent, render } from '@testing-library/svelte';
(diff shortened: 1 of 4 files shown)


deboer-tim (approved):
I think I'd marginally prefer we kept binding to a value and then converted/applied it in `updateProxySettings()`  🤷🏻‍♂️ but code is fine. 👍🏼

I'm curious if there's a better pattern we can use when values aren't strings (maybe ObjectDropdown that takes the objects and a string conversion function?) but probably not worth it for the number of times we have this.

axel7083:
> I think I'd marginally prefer we kept binding to a value and then converted/applied it in `updateProxySettings()` 🤷🏻‍♂️ but code is fine. 👍🏼

Yeah, I would  also prefer, but I wanted to minimise the code changes

gastoner (approved):
Codewise LGTM

PR #11778, opened 2025-03-20
fix(renderer): status-bar min height

axel7083:
### What does this PR do?

While working on pin / unpin I noticed that the current status bar height is not defined anywhere, but a consequence of the font and icon of the entries.

The `ProviderWidget` has a height of 26 pixel, when we remove all of them we fallback on the previous 24, with a flickering. (See video bellow)

### Screenshot / video of UI

https://github.com/user-attachments/assets/e88fb85b-ba88-4e2d-a828-b7043bb3c054

Required for

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/help/HelpMenu.svelte
+++ b/packages/renderer/src/lib/help/HelpMenu.svelte
@@ -5,7 +5,7 @@ let dropDownHeight: number;
 let dropDownWidth: number;
 let dropDownElement: HTMLElement;
 
-const STATUS_BAR_HEIGHT = 24;
+const STATUS_BAR_HEIGHT = 26;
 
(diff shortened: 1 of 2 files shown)


deboer-tim (changes requested):
I don't have a problem making the status bar 26px high, but looking at _why_ feels like we're just sweeping debt under the rug. The provider widgets currently have extra padding, a different font size, aren't vertically aligned with the other items, and shouldn't use Button.

Raising the height is just masking this one issue. We should fix the underlying cause with the widget instead, and if UX prefers a taller bar or bigger font we can do that (but consistently) too.

axel7083:
Closing in favour of https://github.com/podman-desktop/podman-desktop/pull/11781

Issue #11508, opened 2025-03-04
Enable svelte/require-each-key rule and update impacted files


jeffmaury:
https://github.com/podman-desktop/podman-desktop/blob/8dd7b6de9f8e3292cb1648548689c3bbfbfed18b/eslint.config.mjs#L271

PR #11793, opened 2025-03-21
chore: enable svelte/require-each-key rule- part 1

SoniaSandler:

### What does this PR do?

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/dashboard/PreflightChecks.svelte
+++ b/packages/renderer/src/lib/dashboard/PreflightChecks.svelte
@@ -12,7 +12,7 @@ async function openLink(url: string): Promise<void> {
 
 {#if preflightChecks.length > 0}
   <div class="flex flex-col w-full p-2 rounded-lg bg-[var(--pd-invert-content-card-bg)]">
-    {#each preflightChecks as preCheck (preCheck)}
+    {#each preflightChecks as preCheck, index (`${preCheck.name}-${index}`)}
       <div class="flex flex-col">
         <div class="mb-4 flex flex-row">
           {#if preCheck.successful === undefined}
@@ -37,7 +37,7 @@ async function openLink(url: string): Promise<void> {
           {/if}
           {#if preCheck.docLinks}
             See:
-            {#each preCheck.docLinks as link (link)}
+            {#each preCheck.docLinks as link, index (`${link.url}-${index}`)}
               <Link aria-label="precheck-link" on:click={async (): Promise<void> => await openLink(link.url)}>{link.title}</Link>
             {/each}
           {/if}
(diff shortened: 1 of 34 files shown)


benoitf:
FYI @SoniaSandler  if you want to get PR checks, you need to apply the label 'area/ci' on draft PR

SoniaSandler:
@benoitf I opened a few draft PR for this issue so that I could see what was done so far and since I'm not sure about some of the keys I chose to use. I'll probably run the CI label on those PRs closer to when I'll have them finished on my side

packages/renderer/src/lib/dashboard/PreflightChecks.svelte
```diff
@@ -37,7 +37,7 @@ async function openLink(url: string): Promise<void> {
           {/if}
           {#if preCheck.docLinks}
             See:
-            {#each preCheck.docLinks as link}
+            {#each preCheck.docLinks as link (link)}
```
SoniaSandler: @benoitf for cases like this where we have an object but no clear unique key that we can use (`{
    title: string;
    url: string;
    group?: string;
  }`) , should we just use the object itself as the key? the index explicitly? or assume that the title is unique for each link (even though there is no indication that the title is unique here)

benoitf: I think using the index + link might be better in that case

```
{#each preCheck.docLinks as link, index (`${link}-${index}`)}
```

SoniaSandler: @benoitf but since the link is an object, isn't it already unique and can be used as a key?

benoitf: sorry , `link.url`

SoniaSandler: I've run into this problem in a lot of files impacted by this change, so what should be the approach to such cases?
Using the object itself, the index, or a combination of some single property and the index?

benoitf: I think combination is better

feloy: Could we add an entry in the code guidelines for these each-keys? The documentation gives an example for some straightforward cases, but in many other cases, it is difficult to know what's the best to use.

axel7083: I agree with @feloy, this is not straightforward, and we could ask the Svelte team (maybe they have a discord/slack..) for recommendation, I don't find much things online 🥲

SoniaSandler:
@benoitf for all the combination keys in this PR, leave them as combination or change them to just index keys based on the discussion from the team meeting today?

benoitf:
@SoniaSandler up to you. If you want to be consistent with your proposal, let's replace everything by indexes. If you're okay with your current work, keep it like this

benoitf (approved):

SoniaSandler:
@benoitf I think that for straightforward cases like id, unique names, map keys, etc. we should use those properties as the keys, but for other cases use the index instead of combination to keep it more simple and consistent

feloy (approved):
LGTM
and many thanks on the clarification of what to do in  non-straightforward cases

PR #11795, opened 2025-03-21
chore: enable svelte/require-each-key rule - part 3

SoniaSandler:

### What does this PR do?

### How to test this PR?


packages/renderer/src/lib/troubleshooting/TroubleshootingGatherLogs.svelte
```diff
@@ -36,7 +36,7 @@ async function saveLogsAsZip(): Promise<void> {
   {#if logs.length > 0}
     <div class="h-full overflow-auto p-2 bg-[var(--pd-invert-content-card-bg)] mt-3">
       <ul aria-label="logs">
-        {#each logs as log}
+        {#each logs as log (log)}
```
SoniaSandler: @benoitf since this log is just a string, could we theoretically have two logs that are the same, causing an error here?
would it be better to use (explicitly) the index as the key?

feloy (approved):

benoitf (approved):

PR #11804, opened 2025-03-24
refactor(ProviderWidget): split component in two

axel7083:
### What does this PR do?

The `ProviderWidget.svelte` component has two main element, the tooltip and the button. In https://github.com/podman-desktop/podman-desktop/issues/9950 I need to have a menu with the provider button **and** a check icon.

image

To be able to do that, I don't need the full `ProviderWidget` comonent, but only the button logic.

In a follow-up PR I will add a slot to be able to add the `check` icon (needed for the pin / unpin indication)

### Screenshot / video of UI

N/A

Required for

### How to test this PR?

cc @SoniaSandler since you worked a lot of this component

### DIFF
--- a/packages/renderer/src/lib/statusbar/ProviderButton.spec.ts
+++ b/packages/renderer/src/lib/statusbar/ProviderButton.spec.ts
@@ -1,0 +1,69 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import '@testing-library/jest-dom/vitest';
+
+import type { ProviderStatus } from '@podman-desktop/api';
+import { render } from '@testing-library/svelte';
+import { beforeEach, expect, test, vi } from 'vitest';
+
+import IconImage from '/@/lib/appearance/IconImage.svelte';
+import ProviderButton from '/@/lib/statusbar/ProviderButton.svelte';
+import type { ProviderInfo } from '/@api/provider-info';
+
+vi.mock(import('/@/lib/statusbar/ProviderWidgetStatus.svelte'));
+vi.mock(import('/@/lib/appearance/IconImage.svelte'));
+
+const PROVIDER_MOCK = {
+  name: 'provider1',
+  containerConnections: [],
+  kubernetesConnections: [],
+  status: 'ready' as ProviderStatus,
+  images: {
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/statusbar/ProviderButton.svelte
```diff
@@ -0,0 +1,31 @@
...
+import ProviderWidgetStatus from '/@/lib/statusbar/ProviderWidgetStatus.svelte';
+import type { ProviderInfo } from '/@api/provider-info';
+
+interface Props {
+  provider: ProviderInfo;
+  onclick: () => void;
+  class?: string;
+}
+
+let { provider, onclick, class: className }: Props = $props();
```
benoitf, line 12: this is not a valid comment for svelte v5, it was for svelte v4

packages/renderer/src/lib/statusbar/ProviderButton.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import type { ProviderStatus } from '@podman-desktop/api';
+import { render } from '@testing-library/svelte';
+import { beforeEach, expect, test, vi } from 'vitest';
+
+import IconImage from '/@/lib/appearance/IconImage.svelte';
+import ProviderButton from '/@/lib/statusbar/ProviderButton.svelte';
+import type { ProviderInfo } from '/@api/provider-info';
+
+vi.mock(import('/@/lib/statusbar/ProviderWidgetStatus.svelte'));
```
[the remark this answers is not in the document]
benoitf: it ensures file is always there in case of file being moved

packages/renderer/src/lib/statusbar/ProviderButton.svelte
```diff
@@ -0,0 +1,28 @@
...
+</script>
+
+<button
+  on:click={onclick}
+  class="px-1 py-px flex flex-row h-full items-center gap-1 min-w-fit hover:bg-[var(--pd-statusbar-hover-bg)] hover:cursor-pointer relative {className}"
+  aria-label={provider.name}>
+  {#if provider.containerConnections.length > 0 || provider.kubernetesConnections.length > 0 || provider.status }
```
benoitf, line 19: should it handle vmConnections introduced by https://github.com/podman-desktop/podman-desktop/commit/e5115d5293609c89f3e37371d5937ed2252c376e ?

axel7083: currently not doing it, probably yes?

But this line is kinda weird

https://github.com/podman-desktop/podman-desktop/blob/e5115d5293609c89f3e37371d5937ed2252c376e/packages/renderer/src/lib/statusbar/ProviderWidget.svelte#L52

condition `entry.containerConnections.length > 0 || entry.kubernetesConnections.length > 0 || entry.status` is always true, as entry.status is always a non-empty string?

benoitf:
@axel7083 please avoid to delete comments, you could just hide them

packages/renderer/src/lib/statusbar/ProviderButton.svelte
benoitf: cc @SoniaSandler

SoniaSandler: If we know that it's for sure non-empty string, we could remove this check, and maybe do the same for the provider name below

axel7083: Can be a follow-up PR

benoitf (approved):
need the follow-up

SoniaSandler (approved):

PR #11809, opened 2025-03-24
refactor: provider-registry:create(Container|Kubernetes)ProviderConnection

feloy:

### What does this PR do?

Refactor provider-registry:createContainerProviderConnection and provider-registry:createKubernetesProviderConnection by introducing a `withTaskConnection`, to separate task related and provided related code.

### Screenshot / video of UI

N/A

Part of #11744 

### How to test this PR?

Tests have been added before refactoring, and are still green

### DIFF
--- a/packages/main/src/plugin/index.spec.ts
+++ b/packages/main/src/plugin/index.spec.ts
@@ -433,12 +433,21 @@ describe.each<{
     methodName: 'createKubernetesProviderConnection',
   },
 ])('$handler', async ({ handler, methodName }) => {
-  test('createTask is called', async () => {
-    vi.spyOn(TaskManager.prototype, 'createTask');
+  let originalTask: Task;
+
+  beforeEach(() => {
+    originalTask = {
+      status: 'in-progress',
+      error: '',
+    } as unknown as Task;
+    vi.spyOn(TaskManager.prototype, 'createTask').mockReturnValue(originalTask);
     vi.spyOn(NavigationManager.prototype, 'navigateToProviderTask');
     vi.spyOn(ProviderRegistry.prototype, 'getProviderInfo').mockReturnValue({
       name: 'provider1',
     } as ProviderInfo);
+  });
+
+  test('createTask is called', async () => {
     await pluginSystem.initExtensions(new Emitter<ConfigurationRegistry>());
     const handle = handlers.get(handler);
     expect(handle).not.equal(undefined);
@@ -451,6 +460,8 @@ describe.each<{
     }
     expect(params.title).toEqual(`Creating provider1 provider`);
     expect(params.action?.name).toEqual(`Open task`);
+
+    // check that action.execute passed to createTask is calling navigateToProviderTask
     const execute = params.action?.execute;
     expect(execute).toBeDefined();
     if (!execute) {
@@ -462,15 +473,6 @@ describe.each<{
   });
 
   test(`${methodName} is called and is resolved`, async () => {
-    const originalTask = {
-      status: 'in-progress',
-      error: '',
-    } as unknown as Task;
-    vi.spyOn(TaskManager.prototype, 'createTask').mockReturnValue(originalTask);
-    vi.spyOn(NavigationManager.prototype, 'navigateToProviderTask');
-    vi.spyOn(ProviderRegistry.prototype, 'getProviderInfo').mockReturnValue({
-      name: 'provider1',
-    } as ProviderInfo);
     vi.spyOn(ProviderRegistry.prototype, methodName).mockResolvedValue();
     const onEndMock = vi.fn();
     const errorMock = vi.fn();
@@ -490,15 +492,6 @@ describe.each<{
   });
 
   test(`${methodName} is called and is rejected`, async () => {
-    const originalTask = {
-      status: 'in-progress',
-      error: '',
-    } as unknown as Task;
-    vi.spyOn(TaskManager.prototype, 'createTask').mockReturnValue(originalTask);
-    vi.spyOn(NavigationManager.prototype, 'navigateToProviderTask');
-    vi.spyOn(ProviderRegistry.prototype, 'getProviderInfo').mockReturnValue({
-      name: 'provider1',
-    } as ProviderInfo);
     const rejectError = new Error('an error');
     vi.spyOn(ProviderRegistry.prototype, methodName).mockRejectedValue(rejectError);
     const onEndMock = vi.fn();
--- a/packages/main/src/plugin/index.spec.ts
+++ b/packages/main/src/plugin/index.spec.ts
@@ -23,11 +23,12 @@ import { tmpdir } from 'node:os';
 import type { PullEvent } from '@podman-desktop/api';
 import type { WebContents } from 'electron';
 import { app, BrowserWindow, clipboard, ipcMain, shell } from 'electron';
-import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
+import { beforeAll, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import { ExtensionLoader } from '/@/plugin/extension/extension-loader.js';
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/index.spec.ts
```diff
@@ -413,3 +419,104 @@ test('ipcMain.handle returns caught error as objects message property if it is n
...
+    expect(params.title).toEqual(`Creating provider1 provider`);
+    expect(params.action?.name).toEqual(`Open task`);
+    const execute = params.action?.execute;
+    expect(execute).toBeDefined();
+    if (!execute) {
+      throw new Error('execute should be defined');
+    }
+    execute(new TaskImpl('task1id', 'task1name'));
+    expect(NavigationManager.prototype.navigateToProviderTask).toHaveBeenCalledOnce();
+    expect(NavigationManager.prototype.navigateToProviderTask).toHaveBeenCalledWith('internal1', 'task1');
```
axel7083: Those tests are very dense, pretty hard to read

packages/main/src/plugin/index.spec.ts
```diff
@@ -413,3 +419,104 @@ test('ipcMain.handle returns caught error as objects message property if it is n
...
+  });
+
+  test(`${methodName} is called and is resolved`, async () => {
+    const originalTask = {
+      status: 'in-progress',
+      error: '',
+    } as unknown as Task;
+    vi.spyOn(TaskManager.prototype, 'createTask').mockReturnValue(originalTask);
+    vi.spyOn(NavigationManager.prototype, 'navigateToProviderTask');
+    vi.spyOn(ProviderRegistry.prototype, 'getProviderInfo').mockReturnValue({
```
axel7083: please factorise the same three line of code in a beforeEach in the describe you added

packages/main/src/plugin/index.ts
```diff
@@ -2320,6 +2320,47 @@ export class PluginSystem {
       },
     );
 
+    const withTaskConnection = async (options: {
```
axel7083: Can't we have this in a dedicated files ? like something utils?

The `index.ts` is pretty big, and this help reducing its complexity, but would be even better in a dedicated files

feloy: This is not easy to extract, as it contains many dependencies to local objects (this.getLogHandler, cancellationTokenRegistry, taskManager)

packages/main/src/plugin/index.spec.ts
feloy: added some comments

packages/main/src/plugin/index.ts
axel7083: I don't know... this feel like this is just making it hard to test, and make the index file more complex again :crying_cat_face: 

I draft something ~ the cancellationTokenRegistry and taskManager are just class, so they can be passed down, the `getLogHandler` can be binded to keep the same behavior, your tests are passing with this

https://github.com/axel7083/podman-desktop/tree/refactor/extract-task-connection-logic

feloy: I'm not sure I want to externalize this function. It is specialized to these 2 ipcHandles only (3 when createVmProviderConnection will be added), and is not generic enough to be used for other places where there is a similar pattern (the ones introduced in https://github.com/podman-desktop/podman-desktop/pull/8144).

My idea is to have this small refactor limited to these 2/3 ipcHandles, as a step towards refactoring all the ipcHandles using this pattern. But I'm not sure I want to refactor everything now. @benoitf WDYT?

benoitf: on my end, I prefer to have only class/methods, no inline arrow functions

so new TaskConnection(.....).someThing()

I think it's ok to have something that works only for ProviderConnection objects being externalized

feloy: I've added a class based on Axel's draft

packages/main/src/plugin/util/task-connection-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type containerDesktopAPI from '@podman-desktop/api';
+
+import type { CancellationTokenRegistry } from '/@/plugin/cancellation-token-registry.js';
+import type { LoggerWithEnd } from '/@/plugin/index.js';
+import type { TaskManager } from '/@/plugin/tasks/task-manager.js';
+
+interface Dependencies {
```
benoitf: the interface name does not seem obvious to me

TaskConnectionUtilsParams ?

packages/main/src/plugin/util/task-connection-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  getLogHandler(channel: string, loggerId: string): LoggerWithEnd;
+  cancellationTokenRegistry: CancellationTokenRegistry;
+  taskManager: TaskManager;
+}
+
+export interface Options {
+  loggerId: string;
+  tokenId: number | undefined;
```
benoitf:
```suggestion
  tokenId?: number;
```

packages/main/src/plugin/util/task-connection-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  title: string;
+  navigateToTask: () => Promise<void>;
+  execute: (logger: LoggerWithEnd, token: containerDesktopAPI.CancellationToken | undefined) => Promise<void>;
```
benoitf:
```suggestion
  execute: (logger: LoggerWithEnd, token?: containerDesktopAPI.CancellationToken) => Promise<void>;
```

packages/main/src/plugin/util/task-connection-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { LoggerWithEnd } from '/@/plugin/index.js';
+import type { TaskManager } from '/@/plugin/tasks/task-manager.js';
+
+interface Dependencies {
+  getLogHandler(channel: string, loggerId: string): LoggerWithEnd;
+  cancellationTokenRegistry: CancellationTokenRegistry;
+  taskManager: TaskManager;
+}
+
+export interface Options {
```
benoitf: the same here

TaskOptions ?

packages/main/src/plugin/util/task-connection-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+        return result;
+      })
+      .catch((err: unknown) => {
+        task.error = options.executeErrorMsg(err);
+        logger.error(err);
+        throw err;
+      })
+      .finally(() => {
+        logger.onEnd();
+      });
```
benoitf: as we're inside an async method, not sure why we're using chaining vs

```ts
try {
... await options.execute(logger, token);

} catch () {

} finally {}
```

packages/main/src/plugin/index.ts
```diff
@@ -2330,40 +2337,16 @@ export class PluginSystem {
...
-          action: {
-            name: 'Open task',
-            execute: () => {
-              navigationManager
-                .navigateToProviderTask(internalProviderId, taskId)
-                .catch((err: unknown) => console.error(err));
-            },
-          },
+          navigateToTask: () => navigationManager.navigateToProviderTask(internalProviderId, taskId),
+          execute: (logger: LoggerWithEnd, token: containerDesktopAPI.CancellationToken | undefined) =>
```
```suggestion
          execute: (logger: LoggerWithEnd, token?: containerDesktopAPI.CancellationToken) =>
```

packages/main/src/plugin/index.ts
```diff
@@ -2388,40 +2371,16 @@ export class PluginSystem {
...
```
```suggestion
          execute: (logger: LoggerWithEnd, token?: containerDesktopAPI.CancellationToken) =>
```

benoitf (approved):
thanks for the changes

feloy:
@axel7083 could you PTAL?

axel7083 (approved):
LGTM

PR #11830, opened 2025-03-25
chore: migrated link component to svelte5

### LINKED ISSUE #11580: Migrate UI component Link to Svelte 5
https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/link/Link.svelte#L1

gastoner:
### What does this PR do?
Migrates Link component to svelte5 syntax

Closes #11580 

### How to test this PR?
PR checks

### DIFF
--- a/packages/ui/src/lib/link/Link.svelte
+++ b/packages/ui/src/lib/link/Link.svelte
@@ -1,19 +1,29 @@
 <script lang="ts">
 import type { IconDefinition } from '@fortawesome/free-regular-svg-icons';
-import { onMount, type Snippet } from 'svelte';
+import { createEventDispatcher, onMount, type Snippet } from 'svelte';
 import Fa from 'svelte-fa';
 
 import { isFontAwesomeIcon } from '../utils/icon-utils';
+
+const dispatch = createEventDispatcher<{ click: undefined }>();
 
 interface Props {
   icon?: IconDefinition;
   class?: string;
   children: Snippet;
-  onclick: () => void;
+  onclick?: () => void;
   'aria-label'?: string;
 }
-// eslint-disable-next-line svelte/no-unused-props
-let { icon = undefined, class: classes = '', 'aria-label': ariaLabel = '', children, onclick }: Props = $props();
+
+let {
+  icon = undefined,
+  class: classes = '',
+  'aria-label': ariaLabel,
+  children,
+  onclick = (): void => {
+    dispatch('click');
+  },
+}: Props = $props();
 
 let iconType: string | undefined = $state(undefined);
 
@@ -33,7 +43,7 @@ onMount(() => {
   class="text-[var(--pd-link)] hover:bg-[var(--pd-link-hover-bg)] transition-all rounded-[4px] p-0.5 no-underline cursor-pointer {classes}"
   {onclick}
   role="link"
-  aria-label={ariaLabel}>
+  aria-label={ariaLabel ?? ''}>
   {#if icon}
     <span class="flex flex-row space-x-2">
       {#if iconType === 'fa'}
--- a/packages/ui/src/lib/link/Link.svelte
+++ b/packages/ui/src/lib/link/Link.svelte
@@ -1,18 +1,22 @@
 <script lang="ts">
-import { createEventDispatcher, onMount } from 'svelte';
+import type { IconDefinition } from '@fortawesome/free-regular-svg-icons';
+import { onMount, type Snippet } from 'svelte';
+import type { HTMLAttributes } from 'svelte/elements';
 import Fa from 'svelte-fa';
 
 import { isFontAwesomeIcon } from '../utils/icon-utils';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let icon: any = undefined;
+interface Props extends HTMLAttributes<HTMLElement> {
+  icon?: IconDefinition;
+  class?: string;
+  children: Snippet;
+  onclick: () => void;
+  ariaLabel?: string;
+}
 
-let iconType: string | undefined = undefined;
+let { icon = undefined, class: classes = '', 'aria-label': ariaLabel = '', children, onclick }: Props = $props();
 
-const dispatch = createEventDispatcher<{ click: undefined }>();
-export let onclick: () => void = () => {
-  dispatch('click');
-};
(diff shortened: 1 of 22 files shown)


axel7083 (approved):
LGTM, from my understanding using `on:click` will now raise an typecheck error, so this is breaking change, but okay, as this is easily catchable

gastoner:
> LGTM, from my understanding using `on:click` will now raise an typecheck error, so this is breaking change, but okay, as this is easily catchable

Yep, I'll also wait for some feedback on the `aria-label` stuff by adding this interface to additional file so we could use in other files too

deboer-tim:
> I'll also wait for some feedback on the `aria-label` stuff by adding this interface to additional file so we could use in other files too

@gastoner @SoniaSandler aria-label issue is just a linter failure, right? What if we just do this for now?
```
interface Props {
  'aria-label'?: string;
}
// eslint-disable-next-line svelte/no-unused-props
let {
  'aria-label': ariaLabel,
}: Props = $props();
```
The eslist-disable is unfortunate, but that's one line to remove later on (assuming they fix this) vs adding lots of unused props via HTMLAttributes, or using a new shared / duplicated interface just to get around the problem.

deboer-tim (commented):
As per the standup call I opened an issue for the linting problem: https://github.com/sveltejs/svelte/issues/15646. I had trouble getting a small test case and found that it doesn't complain if aria-label is the only prop, which makes it seem even more likely to me that this a minor linting bug, but we will see.

Changes LGTM but unfortunately need to be rebased now.

benoitf:
wondering if the repository to file bugs is not then https://github.com/sveltejs/svelte-eslint-parser/issues ?

deboer-tim:
> wondering if the repository to file bugs is not then https://github.com/sveltejs/svelte-eslint-parser/issues ?

I wasn't sure - only a few issues there and I found a few linting issues open in sveltejs/svelte too. 🤷🏻‍♂️ Much more activity in the latter so I figured they'd transfer it or tell me, if you're pretty sure I can move it.

gastoner:
@deboer-tim @benoitf PR rebased!

deboer-tim (approved):
Thanks for the updates, @gastoner. LGTM 👍🏼

Approving, but as per the standup yesterday please hold off on merging until we get confirmation this is a bug (in which case we merge and open an issue to remove svelte/no-unused-props later) or decide on another path.

deboer-tim:
Received confirmation that the issue is a bug, fixed in eslint-plugin-svelte 3.5.0 (just released today, we are on 3.4.0).

@benoitf do you prefer to hold this until dependabot update (removing eslint-disable comment), fast-track the update ourselves then merge, or merge this and open issue to track removing the comment later?

benoitf:
I would say, hold until we merge the update of eslint-plugin-svelte to 3.5.0

packages/ui/src/lib/link/Link.svelte
```diff
@@ -1,18 +1,21 @@
...
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let icon: any = undefined;
+interface Props {
+  icon?: IconDefinition;
+  class?: string;
+  children: Snippet;
+  onclick: () => void;
+  'aria-label'?: string;
+}
+// eslint-disable-next-line svelte/no-unused-props
```
benoitf: I merged https://github.com/podman-desktop/podman-desktop/pull/11959

so try to rebase and remove this comment

gastoner: Still got the same error :(

deboer-tim: I can't tell what order you tried things, and did you remember to `pnpm install` again? :) This branch appears to have picked up the eslint-plugin-svelte fix, and with that I'm not seeing the error anymore. I've tried this branch locally and could remove the `// eslint-disable`, or in #11930.

packages/ui/src/lib/link/Link.svelte
```diff
@@ -1,18 +1,21 @@
...
+  'aria-label'?: string;
+}
 
-let iconType: string | undefined = undefined;
+let { icon = undefined, class: classes = '', 'aria-label': ariaLabel = '', children, onclick }: Props = $props();
 
-const dispatch = createEventDispatcher<{ click: undefined }>();
-export let onclick: () => void = () => {
-  dispatch('click');
-};
```
benoitf, line 15: issue: removing the dispatcher will break all components

as intermediate step we introduced both the callback and the dispatcher (if callback is provided, the dispatch event is not thrown)

removing that part will break all existing apps using Link

I think we should keep this approach for a while, the time to migrate in apps the move to the callback rather than the event

benoitf (changes requested):

packages/ui/src/lib/link/Link.svelte
```diff
@@ -1,18 +1,31 @@
...
-import { createEventDispatcher, onMount } from 'svelte';
+import type { IconDefinition } from '@fortawesome/free-regular-svg-icons';
+import { createEventDispatcher, onMount, type Snippet } from 'svelte';
 import Fa from 'svelte-fa';
 
 import { isFontAwesomeIcon } from '../utils/icon-utils';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let icon: any = undefined;
+const dispatch = createEventDispatcher();
```
benoitf: wondering why `<{ click: undefined }>();` has been removed ?

benoitf (approved):

gastoner:
@deboer-tim @benoitf The linter is still failing :( , it should be rebased

benoitf:
@gastoner from my understanding allowUnusedNestedProperties should be added into the linter configuration

benoitf:
or then the patch https://github.com/sveltejs/eslint-plugin-svelte/pull/1178 is not working, I have not checked

benoitf:
so it works with the alias but not with a default value


```svelte

interface Props {
  'aria-label'?: string;
}

const {
  'aria-label': ariaLabel,
}: Props = $props();
```

but not with

```svelte
const {
  'aria-label': ariaLabel = '',
}: Props = $props();
```

gastoner:
@benoitf Updated!

benoitf (approved):

PR #11832, opened 2025-03-25
chore: updated tsconfig and fixed typecheck

### LINKED ISSUE #11746: include typescript scripts in the typecheck check

we have some scripts in extensions that are using typescript but they're not checked in the typecheck command

so invalid types can be used there and there is no report

enhance typecheck to include these scripts

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds scripts to tsconfig for typecheck (kind and podman extensions only)

Closes #11746 

### How to test this PR?
PR checks should be ✅ 

### DIFF
--- a/extensions/kind/scripts/download.ts
+++ b/extensions/kind/scripts/download.ts
@@ -62,6 +62,7 @@ async function download(tagVersion: string, repoPath: string, fileName: string):
   });
   let buffer;
 
+  if (!manifests.data.content) return;
   if (manifests.data.encoding && manifests.data.encoding === 'base64') {
     buffer = Buffer.from(manifests.data.content, 'base64');
   } else {
--- a/extensions/kind/scripts/download.ts
+++ b/extensions/kind/scripts/download.ts
@@ -62,7 +62,10 @@ async function download(tagVersion: string, repoPath: string, fileName: string):
   });
   let buffer;
 
-  if (!manifests.data.content) return;
+  if (!manifests.data.content) {
+    throw new Error('No content in manifests data');
+  }
+
   if (manifests.data.encoding && manifests.data.encoding === 'base64') {
     buffer = Buffer.from(manifests.data.content, 'base64');
   } else {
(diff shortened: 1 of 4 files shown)


extensions/kind/scripts/download.ts
```diff
@@ -62,6 +62,7 @@ async function download(tagVersion: string, repoPath: string, fileName: string):
   });
   let buffer;
 
+  if (!manifests.data.content) return;
```
axel7083: if this is an expected case, we should throw an error instead of simply return, like if this is something that should never happen, better to throw

extensions/podman/packages/extension/scripts/podman-download.ts
```diff
@@ -29,6 +29,28 @@ export enum DiskType {
...
+    architecture: string;
+  };
+};
+
+type ResponseLayers = {
+  digest: string;
+  size: number;
+};
+
+type ResponseJSON = {
```
axel7083: Please specify in the name what reponse json this is about, and probably better to name it `type ManifestsResponse`  (with an S) we can assume this is json.

For the others, no need to prefix Response, let's have ManifestLayers and Manifest.

axel7083: And could we have some comments ? like saying where the request come from to clarify what are the types

axel7083 (commented):
Left some comments

axel7083 (approved):
LGTM 🚀  thanks for the change!

gastoner:
@benoitf @cdrage can you PTAL?

extensions/podman/packages/extension/scripts/podman-download.ts
```diff
@@ -288,17 +311,18 @@ export class Podman5DownloadMachineOS {
...
+  async getManifest(manifestUrl: string): Promise<ManifestsResponse> {
     const response = await fetch(manifestUrl, {
       method: 'GET',
       headers: {
         'docker-distribution-api-version': 'registry/2.0',
         Accept: 'application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json',
       },
     });
-    return response.json();
+    return response.json() as unknown as ManifestsResponse;
```
benoitf, line 325: hello, how can you assert that it's matching ManifestResponse ?

benoitf (approved):
LGTM as step0 but I think we should check that the data is valid before asserting it's corresponding to this spec

PR #11603, opened 2025-03-10
chore(deps-dev): bump electron from 34.3.0 to 35.0.1 and to Node.js v22

benoitf:
### What does this PR do?
Bump to Node.js v22 and then Electron as well

restore the work done by @gastoner in https://github.com/podman-desktop/podman-desktop/pull/10369

### What issues does this PR fix or reference?

fixes https://github.com/podman-desktop/podman-desktop/issues/11511

### How to test this PR?

- [x] Tests are covering the bug fix or the new feature

PR #11838, opened 2025-03-25
fix: name column alignment

### LINKED ISSUE #11831: Table name column alignment issues
### Bug description

Last week I was confused how the Kubernetes name column had alignment issues when we had copied it from other existing columns. I opened and fixed #11731.

This week I had enough other content that I see our Pod and Container name columns have the same problem:

Image
Image

I traced it back to the upgrade to Electron 35 and Node 22: #11603.

Fix is easy enough, just trying to check if this was an expected change, presumably in Electron 35.

### Operating system

macOS 15.3

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Moving to Electron 35 changed the default alignment of text to center, which only causes UI change when the parent component is wider than the text. So far I've only found impact to two name columns. I can't find the 'breaking' change listed anywhere, but it is easy enough to fix with text-left for these two columns.

### Screenshot / video of UI

Before:

Screenshot 2025-03-24 at 12 10 55 PM

Screenshot 2025-03-24 at 12 15 13 PM

After:

Screenshot 2025-03-24 at 12 14 04 PM

Screenshot 2025-03-24 at 12 15 23 PM

Fixes #11831.

### How to test this PR?

See examples, the issue will only show up when you have names shorter/longer than other text in the column.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerColumnName.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerColumnName.spec.ts
@@ -86,6 +86,10 @@ test('Expect simple column styling - pod', async () => {
 
   const text = screen.getByText(`${pod.name} (pod)`);
   expect(text).toBeInTheDocument();
+  expect(text).toHaveClass('overflow-hidden');
+  expect(text).toHaveClass('text-ellipsis');
+
+  expect(text.parentElement).toHaveClass('text-left');
 });
 
 test('Expect simple column styling - compose', async () => {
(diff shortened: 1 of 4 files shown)


benoitf (approved):
are you sure it's related to Electron 35 ?

deboer-tim:
> are you sure it's related to Electron 35 ?

I narrowed it down to the PR #11603 for Electron 35 and Node 22, and assuming it's not the latter. I couldn't find anything related in the Electron release notes, nor the Chromium 33 or 34 releases that were pulled in. Since I can't find anything I don't know for sure (maybe it's even a bug they'll fix at some point?) but at least this change is safe either way.

gastoner (approved):
Codewise LGTM

PR #11840, opened 2025-03-25
chore: enable svelte/require-each-key rule - part 5 - image

SoniaSandler:

### What does this PR do?

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -354,7 +354,7 @@ async function abortBuild(): Promise<void> {
     <div hidden={buildImageInfo?.buildRunning}>
       <label for="inputKey" class="block mb-2 font-semibold text-[var(--pd-content-card-header-text)]"
         >Build arguments</label>
-      {#each buildArgs as buildArg, index (buildArg)}
+      {#each buildArgs as buildArg, index (index)}
         <div class="flex flex-row items-center space-x-2 mb-2">
           <Input bind:value={buildArg.key} name="inputKey" placeholder="Key" class="grow" required />
           <Input bind:value={buildArg.value} placeholder="Value" class="grow" required />
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -354,7 +354,7 @@ async function abortBuild(): Promise<void> {
     <div hidden={buildImageInfo?.buildRunning}>
       <label for="inputKey" class="block mb-2 font-semibold text-[var(--pd-content-card-header-text)]"
         >Build arguments</label>
-      {#each buildArgs as buildArg, index}
+      {#each buildArgs as buildArg, index (buildArg)}
         <div class="flex flex-row items-center space-x-2 mb-2">
           <Input bind:value={buildArg.key} name="inputKey" placeholder="Key" class="grow" required />
           <Input bind:value={buildArg.value} placeholder="Value" class="grow" required />
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
@@ -108,7 +108,7 @@ onMount(() => {
   <div class="flex flex-row grow w-full mt-2 items-end">
     <div class="flex">
       {#if badges.length > 0}
-        {#each badges as badge}
+        {#each badges as badge, index (`${badge}-${index}`)}
           <div
             class="text-[var(--pd-content-card-text)] border-[var(--pd-content-card-border-selected)] border text-sm font-medium me-2 px-2.5 py-0.5 rounded-xl">
             {badge}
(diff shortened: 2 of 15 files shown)


packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -354,7 +354,7 @@ async function abortBuild(): Promise<void> {
     <div hidden={buildImageInfo?.buildRunning}>
       <label for="inputKey" class="block mb-2 font-semibold text-[var(--pd-content-card-header-text)]"
         >Build arguments</label>
-      {#each buildArgs as buildArg, index}
+      {#each buildArgs as buildArg, index (buildArg)}
```
SoniaSandler: since buildArg is `{
    key: string;
    value: string;
}`, and both properties are bound to input fields below, any input will change them and cause the entire each key to change.
IIUC, the only options here are to either use the object itself or the index (explicitly)

packages/renderer/src/lib/image/ImportContainersImages.svelte
```diff
@@ -127,7 +127,7 @@ async function importContainers(): Promise<void> {
         <div class="flex flex-col w-2/4 mr-2.5">Image Name when importing (e.g quay.io/podman/hello)</div>
       </div>
     {/if}
-    {#each containersToImport as containerToImport, index}
+    {#each containersToImport as containerToImport, index (containerToImport)}
```
SoniaSandler: > properties are bound to input fields below, any input will change them and cause the entire each key to change.
IIUC, the only options here are to either use the object itself or the index (explicitly)

Same case here

gastoner (approved):
Codewise LGTM

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
axel7083: This will _unlikely_ scenario cause a big issue :/

image

we should default to index then :(

packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
```diff
@@ -108,7 +108,7 @@ onMount(() => {
   <div class="flex flex-row grow w-full mt-2 items-end">
     <div class="flex">
       {#if badges.length > 0}
-        {#each badges as badge}
+        {#each badges as badge, index (`${badge}-${index}`)}
```
axel7083: we should try to have a primitive, and serializing is expensive, doesn't badge have any unique property?

packages/renderer/src/lib/image/BuildImageFromContainerfileCards.svelte
```diff
@@ -133,7 +133,7 @@ function addCard(item: { value: string }): void {
 
 <div class="flex flex-col" role="region" aria-label="Build Platform Options">
   <div class="flex flex-row gap-x-4 gap-y-4 flex-wrap">
-    {#each sortedCards as card}
+    {#each sortedCards as card, index (`${card.title}-${index}`)}
```
axel7083: In this scenario, I would say, we should probably add a unique `id` to the card interface, and makes it unique (and ensure it is unique through a test)

packages/renderer/src/lib/image/BuildImageFromContainerfileCards.svelte
```diff
@@ -156,7 +156,7 @@ function addCard(item: { value: string }): void {
   {:else}
     <div class="flex flex-col pt-4">
       <div class="flex flex-row gap-x-4 flex-wrap gap-y-4">
-        {#each advancedCards as card}
+        {#each advancedCards as card, index (`${card.title}-${index}`)}
```
axel7083: same as https://github.com/podman-desktop/podman-desktop/pull/11840/files#r2013616332

axel7083 (commented):
I am not sure if it is _recomended_ to use the index in the each key :/

Probably need to be addressed in a call

packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
SoniaSandler: @axel7083 the badge is a string

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
SoniaSandler: @axel7083 I'm not sure if the issue you are referencing is duplicate keys since I tried it with the same values and I didn't get a duplicate key error. So unless there are 2 objects in this array that reference exactly the same object, it shouldn't be a problem, even with the same properties as every object (reference) is unique, but I can still use the index as the key since everything here inside is reactive

packages/renderer/src/lib/image/BuildImageFromContainerfileCard.svelte
axel7083: can badge be duplicate ? like can we have the same badge ?

SoniaSandler: It's possible. As far as I know, there is nothing that guarantees uniqueness for badge. This is why it's used with index to make the key unique

benoitf (approved):

cdrage (approved):

PR #11853, opened 2025-03-26
chore: remove traces

benoitf:
### What does this PR do?
when working on Carousel component (for migrating it to svelte v5) I noticed several unexpected output being displayed in the console. Removing them.

### How to test this PR?


deboer-tim (approved):
Noticed this yesterday afternoon while running other tests - thank you!

PR #11860, opened 2025-03-26
fix(dropdown): ensure onChange value are non-undefined

axel7083:
### What does this PR do?

Final PR related for https://github.com/podman-desktop/podman-desktop/issues/10734

Last typecheck failing in https://github.com/podman-desktop/podman-desktop/pull/11693 is about the tests added in https://github.com/podman-desktop/podman-desktop/pull/11771

### Screenshot / video of UI

N/A

Required for

### How to test this PR?

- [x] Pipeline should be :green_circle:

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesProxiesRendering.spec.ts
@@ -20,7 +20,7 @@ import '@testing-library/jest-dom/vitest';
 
 import { Dropdown } from '@podman-desktop/ui-svelte';
 import { fireEvent, render } from '@testing-library/svelte';
-import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { assert, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import PreferencesProxiesRendering from '/@/lib/preferences/PreferencesProxiesRendering.svelte';
 import { PROXY_LABELS } from '/@/lib/preferences/proxy-state-labels';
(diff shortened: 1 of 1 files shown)


deboer-tim (approved):

benoitf (approved):

gastoner (approved):
Question: 
Why do we use `assert`? Can't we use e.g. `expect(label).toBeDefined()` or so

axel7083:
> Question: Why do we use `assert`? Can't we use e.g. `expect(label).toBeDefined()` or so

@gastoner expect does not sure type safety after it, assert does

```
let foo: string | undefined = /* Get something possibly undefined*/
expect(foo).toBeDefined();

foo => still string | undefined 
```

Before I was doing the following
```
let foo: string | undefined = /* Get something possibly undefined*/
if(!foo) throw new Error('undefined');

foo => foo is string (typecheck happy)
```

@benoitf was the first one to use assert, and it is much better than using the if throw error

gastoner:
> > Question: Why do we use `assert`? Can't we use e.g. `expect(label).toBeDefined()` or so
> 
> @gastoner expect does not sure type safety after it, assert does
> 
> ```
> let foo: string | undefined = /* Get something possibly undefined*/
> expect(foo).toBeDefined();
> 
> foo => still string | undefined 
> ```
> 
> Before I was doing the following
> 
> ```
> let foo: string | undefined = /* Get something possibly undefined*/
> if(!foo) throw new Error('undefined');
> 
> foo => foo is string (typecheck happy)
> ```
> 
> @benoitf was the first one to use assert, and it is much better than using the if throw error

But won't the `expect(foo).toBeDefined();` fail if it is undefined? so after expect is always string although it is not explicitly written with if statement. 
I agree that the assert that will show you the message is nice

axel7083:
> I agree that the assert that will show you the message is nice

The problem is not runtime, it is static analysis, the onChange method take a string as argument, not a `string | undefined` therefore if you use expect().toBeDefined() at runtime you know it will be a string, but typescript don't know that, you need to use assert to tell the type checker, that the value is a string

PR #11862, opened 2025-03-26
feat(ProviderButton): adding left slot

axel7083:
### What does this PR do?

Following https://github.com/podman-desktop/podman-desktop/pull/11804 I need to add a left slot to be able to add a :heavy_check_mark: icon as per design (see bellow)

image

We can see that left to the ProviderButton, I need to add a check icon, but I need it to be part of the button, so adding a left slot, so the parent component can just provide a snippet.

### Screenshot / video of UI

N/A

Required for

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/statusbar/ProviderButton.svelte
+++ b/packages/renderer/src/lib/statusbar/ProviderButton.svelte
@@ -19,9 +19,7 @@ let { provider, onclick, left, class: className }: Props = $props();
   on:click={onclick}
   class="px-1 py-px flex flex-row h-full items-center gap-1 min-w-fit hover:bg-[var(--pd-statusbar-hover-bg)] hover:cursor-pointer relative {className}"
   aria-label={provider.name}>
-  {#if left}
-    {@render left()}
-  {/if}
+  {@render left?.()}
   {#if provider.containerConnections.length > 0 || provider.kubernetesConnections.length > 0 || provider.status }
     <ProviderWidgetStatus entry={provider} />
   {/if}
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/statusbar/ProviderButton.svelte
```diff
@@ -7,15 +9,19 @@ interface Props {
...
+let { provider, onclick, left, class: className }: Props = $props();
 </script>
 
 <button
   on:click={onclick}
   class="px-1 py-px flex flex-row h-full items-center gap-1 min-w-fit hover:bg-[var(--pd-statusbar-hover-bg)] hover:cursor-pointer relative {className}"
   aria-label={provider.name}>
+  {#if left}
+    {@render left()}
+  {/if}
```
benoitf: question:

does: {@render left?.()} works ? (no need of the if block)

benoitf (approved):

packages/renderer/src/lib/statusbar/ProviderButton.svelte
deboer-tim: Yes it does, e.g. we use in Dropdown, places where children is optional. Should be changed.

axel7083:
```suggestion
  {@render left?.()}
```

axel7083: Nice!! Didn't know, fixed

PR #11868, opened 2025-03-26
fix: wait for kubeconfig change while creating kind cluster

### LINKED ISSUE #11813: Kind cluster deployment failing with a 'No Active Cluster' message
### Bug description

Kind cluster deployment is failing with a 'No active cluster' message.

Image

### Operating system

Fedora

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

-  Delete the local kubeconfig.
- Try to deploy a new Kind cluster with an ingress controller.
- Observe the error: "No active cluster".
- Check if the Kind cluster is operable—it should be.

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

The Kind extension was exec-ing the cli to create a cluster and then immediately creating the ingress by applying yaml. A recent fix to main/kubernetes-client to correctly use the cached version of kubeconfig (#11779) causes kind to fail because this doesn't give the file watcher enough time to be notified about the new cluster in kubeconfig.

This fix just waits for an update to kubeconfig between the two steps.

This still isn't 100% bulletproof because there could be some other process touching kubeconfig at exactly the same time, or there could be a timing issue because kubernetes-client notifies as it is re-reading kubeconfig instead of afterwards. However, this seems robust enough and with these changes my tests have gone from always failing to always passing.

### Screenshot / video of UI

N/A

Fixes #11813.

### How to test this PR?

Stop any running Kind clusters to avoid conflict.
Confirm creating a Kind cluster works normally.
Repeat a few times.

### DIFF
--- a/extensions/kind/src/create-cluster.spec.ts
+++ b/extensions/kind/src/create-cluster.spec.ts
@@ -42,6 +42,7 @@ vi.mock('@podman-desktop/api', async () => {
     kubernetes: {
       createResources: vi.fn(),
       getKubeconfig: vi.fn().mockReturnValue({ path: '/some/path' }),
+      onDidUpdateKubeconfig: vi.fn(),
     },
     provider: {
       getContainerConnections: vi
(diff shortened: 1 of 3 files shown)


feloy (approved):

amisskii (changes requested):
Tested. K8s e2e tests failed https://github.com/podman-desktop/podman-desktop/actions/runs/14101414022. The error message :  `Error: Error: Failed to create kind cluster. No active cluster!`

deboer-tim:
The only race condition I'm aware of is that we notify before rereading kubeconfig instead of afterward. I've confirmed that the only other listeners aren't affected (mostly kube-context extension, which reads the kubeconfig itself so is unaffected) and switched the order of this, please give it another try.

@feloy in case you want to check the last commit.

amisskii (approved):
The latest test has passed (same link). Approving.

Issue #11748, opened 2025-03-19
Expose new resource type virtual machine in Podman Desktop API


jeffmaury:
### Is your enhancement related to a problem? Please describe

The Podman Desktop API should be extended to expose the new virtual machine resource type and corresponding actions: create, delete, start,stop, ssh into

### Describe the solution you'd like

Extends the Podman Desktop API and implement it into the core/backend

PR #11880, opened 2025-03-27
feat: startProvider works for VM connections

feloy:
### What does this PR do?

Make startProvier work for providers with VM connections

### Screenshot / video of UI

Backend only

Part of #11744 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/provider-registry.ts
+++ b/packages/main/src/plugin/provider-registry.ts
@@ -604,12 +604,14 @@ export class ProviderRegistry {
     }
 
     const provider = this.getMatchingProvider(providerInternalId);
-    let connection: ContainerProviderConnection | KubernetesProviderConnection | undefined;
+    let connection: ContainerProviderConnection | KubernetesProviderConnection | VmProviderConnection | undefined;
 
     if (provider.containerConnections && provider.containerConnections.length > 0) {
       connection = provider.containerConnections[0];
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/provider-registry.ts
```diff
@@ -604,12 +604,14 @@ export class ProviderRegistry {
     }
 
     const provider = this.getMatchingProvider(providerInternalId);
-    let connection: ContainerProviderConnection | KubernetesProviderConnection | undefined;
+    let connection: ContainerProviderConnection | KubernetesProviderConnection | VmProviderConnection | undefined;
```
benoitf, line 607: not for this PR but like follow-up: add a union type for `ContainerProviderConnection | KubernetesProviderConnection | VmProviderConnection`

benoitf (approved):

packages/main/src/plugin/provider-registry.ts
feloy: for the moment `ContainerProviderConnection` and `KubernetesProviderConnection` are defined in `extension-api.d.ts` but `VmProviderConnection` is still defined internally (before https://github.com/podman-desktop/podman-desktop/issues/11748). The union type will be defined when we move the definition of `VmProviderConnection` in `extension-api.d.ts` (cf @SoniaSandler)

axel7083 (approved):
LGTM

Issue #11579, opened 2025-03-07
Migrate UI component layouts to Svelte 5


jeffmaury:
https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/layouts/DetailsPage.svelte#L1

https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/layouts/FormPage.svelte#L1

https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/layouts/NavPage.svelte#L1

https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/layouts/Page.svelte#L1

rujutashinde:
@feloy do you think this issue can be worked on for release 1.19.0? It is marked as On Hold, Do we know if the underlying issues are resolved?

feloy:
@rujutashinde yes this is ok to work on this issue

PR #11889, opened 2025-03-27
refactor: component DetailsPage to svelte 5

feloy:
### What does this PR do?

Refactor component DetailsPage to svelte 5

### Screenshot / video of UI

N/A

Part of #11579 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/DetailsPage.svelte
+++ b/packages/renderer/src/lib/ui/DetailsPage.svelte
@@ -1,12 +1,34 @@
 <script lang="ts">
 import { DetailsPage } from '@podman-desktop/ui-svelte';
+import type { Snippet } from 'svelte';
 import { router } from 'tinro';
 
 import { currentPage, lastPage } from '../../stores/breadcrumb';
 
-export let title: string;
-export let titleDetail: string | undefined = undefined;
-export let subtitle: string | undefined = undefined;
+interface Props {
+  title: string;
+  titleDetail?: string;
+  subtitle?: string;
+
+  contentSnippet?: Snippet;
+  tabsSnippet?: Snippet;
+  iconSnippet?: Snippet;
+  subtitleSnippet?: Snippet;
+  actionsSnippet?: Snippet;
+  detailSnippet?: Snippet;
+}
+
+const {
+  title,
+  titleDetail = undefined,
+  subtitle = undefined,
+  contentSnippet: localContentSnippet,
+  tabsSnippet: localTabsSnippet,
+  iconSnippet: localIconSnippet,
+  subtitleSnippet: localSubtitleSnippet,
+  actionsSnippet: localActionsSnippet,
+  detailSnippet: localDetailSnippet,
+}: Props = $props();
 
 export function close(): void {
   router.goto($lastPage.path);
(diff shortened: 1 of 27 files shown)


feloy:
> The changes look good, but it would be nice to see a follow-up PR that removes the old slot syntax from the files that were changed.

On which changed file do you see remaining slots?

feloy:
> The changes look good, but it would be nice to see a follow-up PR that removes the old slot syntax from the files that were changed.

@sourcery-ai On which changed file do you see remaining slots?

packages/renderer/src/lib/ui/DetailsPage.svelte
```diff
@@ -1,12 +1,34 @@
...
+  subtitleSnippet?: Snippet;
+  actionsSnippet?: Snippet;
+  detailSnippet?: Snippet;
+}
+
+const {
+  title,
+  titleDetail = undefined,
+  subtitle = undefined,
+  contentSnippet: localContentSnippet,
```
jeffmaury, line 25: Wondering why we need different names here ?

jeffmaury (approved):
LGTM

packages/renderer/src/lib/ui/DetailsPage.svelte
feloy: This is to avoid conflict in:
```
{#snippet iconSnippet()}{@render localIconSnippet?.()}{/snippet}
```

Or we would have the same name for the snippet of the current component and the snippet of the sub-component

packages/ui/src/lib/layouts/DetailsPage.svelte
```diff
@@ -1,14 +1,41 @@
...
-export let onbreadcrumbClick: () => void = () => {};
+interface Props {
+  title: string;
+  titleDetail?: string;
+  subtitle?: string;
+  breadcrumbLeftPart?: string;
+  breadcrumbRightPart?: string;
+  breadcrumbTitle?: string;
+  onclose?: () => void;
+  onbreadcrumbClick?: () => void;
```
deboer-tim, line 14: These were actually inconsistent before (onclose vs onbreadcrumbClick), maybe change this to 'onbreadcrumb'?

deboer-tim (approved):
I'm not a fan of renaming the snippets, I think it's obvious enough as-is and would prefer to have:
`{#snippet icon()}`
over:
`{#snippet iconSnippet()}`

Likewise, I think this would be a good change to remove the 'click' inconsistency as noted.

These are preferences though and the PR is good, so I'm approving. 👍🏼

feloy:
> I'm not a fan of renaming the snippets, I think it's obvious enough as-is and would prefer to have: `{#snippet icon()}` over: `{#snippet iconSnippet()}`
> 
> Likewise, I think this would be a good change to remove the 'click' inconsistency as noted.
> 
> These are preferences though and the PR is good, so I'm approving. 👍🏼

As there is a conflict in subtitle / subtitleSnippet, I prefer to keep Snippet suffix everywhere, until we agree on a pattern

axel7083:
Renaming the slots break backward compatibility - when updating the ui package, I am forced to change all the slots to snippets.

PR #11892, opened 2025-03-27
chore: add alert for mac podman socket failures

### LINKED ISSUE #11429: Notifcation when Podman extension cannot redirect the `/var/run/docker.sock`

When we are unable to redirect the `/var/run/docker.sock` to Podman there is no clear indication for that in Podman Desktop. We can improve that by adding a notification to let the user know. 

When the Podman extension tries to redirect the `/var/run/docker.sock` and fails notify the user through a message in the task manager.

Message: Failed to redirect `/var/run/docker.sock` to Podman
Type: Error

When the user clicks on this, send them to the Docker compatibility page. 

_No response_

### Additional context

_No response_

cdrage:
chore: add alert for mac podman socket failures

### What does this PR do?

* Adds an error message when we've detected a non-disguised socket when
  the "Enable" button has already been pressed
* This is only checked on initial startup and does not perform any
  checks when restarting the machine, etc. This is intentional (for
  now).
* Only on Mac (since we use docker compatibility exclusively there).

### Screenshot / video of UI

Screenshot 2025-03-31 at 9 55 02 AM

Closes https://github.com/podman-desktop/podman-desktop/issues/11429

### How to test this PR?

1. Press "Enable docker compatibility" in PD.
2. Close PD
3. Open Docker Desktop + let it take over the socket
4. Open PD, should get the error notification on startup due to the
   socket "not being disguised".

### DIFF
--- a/extensions/podman/packages/extension/src/extension.spec.ts
+++ b/extensions/podman/packages/extension/src/extension.spec.ts
@@ -42,6 +42,7 @@ import type { UpdateCheck } from './utils/podman-install';
 import { PodmanInstall } from './utils/podman-install';
 import { getAssetsFolder, LIBKRUN_LABEL, LoggerDelegator, VMTYPE } from './utils/util';
 import * as util from './utils/util';
+import { isDisguisedPodman } from './utils/warnings';
 
 const config: Configuration = {
   get: () => {
@@ -3213,6 +3214,114 @@ test('activate and autostart should not duplicate machines ', async () => {
   // should be only 1 but we allow some more calls (if there is not a check to check during the autostart it would be 100+ calls)
   expect(podmanMachineListCalls).toBeLessThan(5);
   expect(promiseAutoStart).toBeDefined();
+});
+
+describe('macOS: tests for notifying if disguised podman socket fails / passes', () => {
+  let contextMock: extensionApi.ExtensionContext;
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+
+    contextMock = {
+      subscriptions: [],
+      secrets: {
+        delete: vi.fn(),
+        get: vi.fn(),
+        onDidChange: vi.fn(),
+        store: vi.fn(),
+      },
+    } as unknown as extensionApi.ExtensionContext;
+
+    // Mock the get compatibility functionality.
+    // we will always assuming it's "disabled" for the tests
+    vi.spyOn(compatibilityModeLib, 'getSocketCompatibility').mockReturnValue({
+      isEnabled: vi.fn().mockReturnValue(true),
+      enable: vi.fn().mockReturnValue(true),
+      disable: vi.fn(),
+      details: '',
+      tooltipText: (): string => {
+        return '';
+      },
+    });
+
+    vi.mock('./utils/warnings');
+  });
+
+  test('when isDisguisedPodman is true, error message should NOT be shown', async () => {
+    // macOS only
+    vi.mocked(extensionApi.env).isMac = true;
+    vi.mocked(extensionApi.env).isWindows = false;
+    vi.mocked(extensionApi.env).isLinux = false;
+
+    // Mock "isDisguisedPodman" to return true to indicate a failed socket
+    vi.mocked(isDisguisedPodman).mockImplementation(async () => true);
+
+    const api = await extension.activate(contextMock);
+    expect(api).toBeDefined();
+
+    // Check that isDisguisedPodman was called
+    expect(isDisguisedPodman).toBeCalled();
+
+    // Check that the error message is NOT shown
+    expect(extensionApi.window.showNotification).not.toBeCalled();
+  });
+
+  test('when isDisguisedPodman is false, error message should be shown', async () => {
+    // macOS only
+    vi.mocked(extensionApi.env).isMac = true;
+    vi.mocked(extensionApi.env).isWindows = false;
+    vi.mocked(extensionApi.env).isLinux = false;
+
+    // Mock "isDisguisedPodman" to return false to indicate a failed socket
+    vi.mocked(isDisguisedPodman).mockImplementation(async () => false);
+
+    const api = await extension.activate(contextMock);
+    expect(api).toBeDefined();
+
+    // Check that isDisguisedPodman was called
+    expect(isDisguisedPodman).toBeCalled();
+
+    // Check that the error message is shown
+    expect(extensionApi.window.showNotification).toBeCalledWith({
+      title: 'Docker socket is not disguised correctly',
+      body: 'The Docker socket (/var/run/docker.sock) is not being properly disguised by Podman. This could potentially cause docker-compatible tools to fail. Please disable any conflicting tools and re-enable Docker Compatibility.',
+      highlight: true,
+      markdownActions:
+        ':button[Docker compatibility settings]{href=/preferences/docker-compatibility title="Docker Compatibility settings"}',
+      silent: true,
+      type: 'error',
+    });
+  });
+
+  test('do not show error message OR call function if on linux', async () => {
+    // linux
+    vi.mocked(extensionApi.env).isMac = false;
+    vi.mocked(extensionApi.env).isWindows = false;
+    vi.mocked(extensionApi.env).isLinux = true;
+
+    const api = await extension.activate(contextMock);
+    expect(api).toBeDefined();
+
+    // Expect that isDisguisedPodman was NOT called
+    expect(isDisguisedPodman).not.toBeCalled();
+
+    expect(extensionApi.window.showNotification).not.toBeCalled();
+  });
+
+  test('do not show error message OR call function if on windows', async () => {
+    // windows
+    vi.mocked(extensionApi.env).isMac = false;
+    vi.mocked(extensionApi.env).isWindows = true;
+    vi.mocked(extensionApi.env).isLinux = false;
+
+    const api = await extension.activate(contextMock);
+    expect(api).toBeDefined();
+
+    // Expect that isDisguisedPodman was NOT called
+    expect(isDisguisedPodman).not.toBeCalled();
+
+    expect(extensionApi.window.showNotification).not.toBeCalled();
+  });
 });
 
 describe('podman-mac-helper tests', () => {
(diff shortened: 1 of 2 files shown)


benoitf:
hello, I'm not sure it should be a popup, I expect to be a notification in the dashboard like "Podman needs to be setup"

so you see the message when you're in the dashboard but it's far less intrusive

cdrage:
> hello, I'm not sure it should be a popup, I expect to be a notification in the dashboard like "Podman needs to be setup"
> 
> so you see the message when you're in the dashboard but it's far less intrusive

My mistake! I misread the original issue. I'll change it to task manager error / rather than showError.

cdrage:
So need your opinion @Firewall 

In the issue it says to use "task manager notification".

I encountered two problems where we're unable to do a proper error reporting, see:

https://github.com/podman-desktop/podman-desktop/issues/11900
https://github.com/podman-desktop/podman-desktop/issues/11899

So the result when doing:

    extensionApi.window.showNotification({
      title: `Compatibility mode error`,
      body: `The Docker socket is not disguised as a Podman socket. Please check any tools that may be overriding /var/run/docker.sock and re-enable compatibility mode.`, 
      type: 'error',
      highlight: true,
    })
    
Results in:
Screenshot 2025-03-28 at 8 54 47 AM

Which does not give any info / not even appearing as an error.

So unfortunatley we have to choose between:


1. Normal error notification:

image


2. Similar notification to podman-mac-helper:

Screenshot 2025-03-28 at 8 57 48 AM

3. Or we can work on issues:

https://github.com/podman-desktop/podman-desktop/issues/11900
https://github.com/podman-desktop/podman-desktop/issues/11899

And have it in task manager like you asked. But we'll have to fix the two above issues first.

Let me know what we should do.

Firewall:
I would go with `2. Similar notification to podman-mac-helper:` then. The normal notification error gives it a little bit too much priority because it totally interupts everything, I don't think this notification is that important.

Firewall:
But! the message from the popup/dialog was much better. Can you update it to that one (`Podman isn't running in Docker compatibility...`)

cdrage:
@benoitf @Firewall 

I've updated it so the notification is now on the dashboard (rather than task manager as discussed above) and updated the PR!

Ready for review 🚀 

Screenshot 2025-03-31 at 9 55 02 AM

Firewall:
Hey @cdrage, I checkout out the branch and the error message shows up correctly. Awesome! I think there is some padding missing left&right around the text on the button. You can see it on your screenshot too. 
image


Other than that, I really like the use of the word "disguised" as it explains well what we are doing. However on the page we take the users for the solution we don't really use that word. I have a feeling users are going to click on the button and then be lost which setting to change on that page. WDYT about changing the text on the Docker compatibility for the `Third-Party Docker Tool Compatibility` to 

```
On startup try to disguise the Docker socket (/var/run/docker.sock) with Podman. This will allow tools that rely on this socket to work with Podman. 
```
and also keeping the conditional text to enable/disable like there is now.
```
Disabling this option will prevent third-party Docker-compatible tools to use Podman. [Disable]
```

cdrage:
> Hey @cdrage, I checkout out the branch and the error message shows up correctly. Awesome! I think there is some padding missing left&right around the text on the button. You can see it on your screenshot too. image
> 
> Other than that, I really like the use of the word "disguised" as it explains well what we are doing. However on the page we take the users for the solution we don't really use that word. I have a feeling users are going to click on the button and then be lost which setting to change on that page. WDYT about changing the text on the Docker compatibility for the `Third-Party Docker Tool Compatibility` to
> 
> ```
> On startup try to disguise the Docker socket (/var/run/docker.sock) with Podman. This will allow tools that rely on this socket to work with Podman. 
> ```
> 
> and also keeping the conditional text to enable/disable like there is now.
> 
> ```
> Disabling this option will prevent third-party Docker-compatible tools to use Podman. [Disable]
> ```

I agree on both!

I've opened up issues for both. The first is a minor UI/UX bug: https://github.com/podman-desktop/podman-desktop/issues/11994

With regards to updating the other section, that'd be outside the scope of this PR, but I've opened up https://github.com/podman-desktop/podman-desktop/issues/11995 to get other people's opinions on it too.

feloy (approved):
LGTM codewise

Firewall:
> I agree on both!
> 
> I've opened up issues for both. The first is a minor UI/UX bug: #11994
> 
> With regards to updating the other section, that'd be outside the scope of this PR, but I've opened up #11995 to get other people's opinions on it too.

Ok 👍

cdrage:
Rebased and ready for re-review!

ping @axel7083

benoitf (changes requested):
I started a podman machine while having something using the socket 

And I get for that a warning in the CLI:

```
Starting machine "another"

This machine is currently configured in rootless mode. If your containers
require root permissions (e.g. ports < 1024), or if you run into compatibility
issues with non-podman clients, you can switch using the following command:

	podman machine set --rootful another

API forwarding listening on: /var/folders/_s/rlbw3qg15vx0j6vr4l7kd1hm0000gn/T/podman/another-api.sock

Another process was listening on the default Docker API socket address.
You can still connect Docker API clients by setting DOCKER_HOST using the
following command in your terminal session:

        export DOCKER_HOST='unix:///var/folders/_s/rlbw3qg15vx0j6vr4l7kd1hm0000gn/T/podman/another-api.sock'

Machine "another" started successfully
```

but no warning in Podman Desktop

benoitf:
note: I had to disable and re-enable the podman extension to be notified

benoitf:
also if I fixed the problem, stopping the process using the socket, and restarted the podman machine

I still have the notification and I'm redirected to the page where it correctly says that podman is listening behind the socket, so I'm a user I would be lost


https://github.com/user-attachments/assets/4ad5e2bc-5a1a-4276-826b-1e0f112e6fdc

cdrage:
> I started a podman machine while having something using the socket
> 
> And I get for that a warning in the CLI:
> 
> ```
> Starting machine "another"
> 
> This machine is currently configured in rootless mode. If your containers
> require root permissions (e.g. ports < 1024), or if you run into compatibility
> issues with non-podman clients, you can switch using the following command:
> 
> 	podman machine set --rootful another
> 
> API forwarding listening on: /var/folders/_s/rlbw3qg15vx0j6vr4l7kd1hm0000gn/T/podman/another-api.sock
> 
> Another process was listening on the default Docker API socket address.
> You can still connect Docker API clients by setting DOCKER_HOST using the
> following command in your terminal session:
> 
>         export DOCKER_HOST='unix:///var/folders/_s/rlbw3qg15vx0j6vr4l7kd1hm0000gn/T/podman/another-api.sock'
> 
> Machine "another" started successfully
> ```
> 
> but no warning in Podman Desktop

In the commit message I added:
"This is only checked on initial startup and does not perform any
checks when restarting the machine, etc. This is intentional (for
now)."

I found that having it always check every time the machine updated, was causing a slow-down and sometimes false positives (socket "detection" disappeared, or computer was put on standby and turned on again, causing errors).

I've opened up an issue to have it reactively show: https://github.com/podman-desktop/podman-desktop/issues/12188 that we can follow up on.

Right now it only does the detection / alert on extension startup.

benoitf:
ok but then at extension startup it's also reporting false messages

I fixed the issue and I still have the message so as a user I would think it's not working

benoitf:
basically two worklows will fail and this is what people might do:

1. start first podman desktop and then docker desktop: in this case: no notification while you expect to see one.
2. stop docker desktop and restart your podman machine: notification can still be there while you fixed the issue

instead of doing it every 5s, you can run it after a container engine provider has been added or removed

cdrage:
> basically two worklows will fail and this is what people might do:
> 
> 1. start first podman desktop and then docker desktop: in this case: no notification while you expect to see one.
> 2. stop docker desktop and restart your podman machine: notification can still be there while you fixed the issue
> 
> instead of doing it every 5s, you can run it after a container engine provider has been added or removed

I've updated the code to check and notify when doing `doUpdateMachines` https://github.com/cdrage/podman-desktop/blob/56d9714acb27156569731c7c429d7f4dfdfb548b/extensions/podman/packages/extension/src/extension.ts#L261-L262

Similar to the podman mac setup mac helper alert).

Can you please try again and see if it gets dismissed on your podman machine restart / no issues?

The only problem is that SOMETIMES our check fails (https://github.com/cdrage/podman-desktop/blob/56d9714acb27156569731c7c429d7f4dfdfb548b/extensions/podman/packages/extension/src/utils/warnings.ts#L29-L30) but that's outside the scope of this PR. 

We need some better compatibility checks in the future.

cdrage:
> basically two worklows will fail and this is what people might do:
> 
> 1. start first podman desktop and then docker desktop: in this case: no notification while you expect to see one.
> 2. stop docker desktop and restart your podman machine: notification can still be there while you fixed the issue
> 
> instead of doing it every 5s, you can run it after a container engine provider has been added or removed

I've updated the code to check and notify when doing `doUpdateMachines` https://github.com/cdrage/podman-desktop/blob/56d9714acb27156569731c7c429d7f4dfdfb548b/extensions/podman/packages/extension/src/extension.ts#L261-L262

Similar to the podman mac setup mac helper alert).

Can you please try again and see if it gets dismissed on your podman machine restart / no issues?

The code is based on https://github.com/cdrage/podman-desktop/blob/56d9714acb27156569731c7c429d7f4dfdfb548b/extensions/podman/packages/extension/src/utils/warnings.ts#L29-L30 which checks if it's "disguised" or not.

Whether or not this needs improvement in the future I'm unsure of. I was able to have the notification dismissed on restart on my machine, but I did get some false positives in the past I'm unable to replicate again.

benoitf (approved):
now it has detected the case

for one, it seems it took a very long time to detect but at least it was ok

cdrage:
> now it has detected the case
> 
> for one, it seems it took a very long time to detect but at least it was ok

I noticed that too and checked the logs, for me it was `podman machine` was taking a while to properly restart / go to running.

It's most likely because I have a ton of images right now within it and I'm not using Rosetta.

cdrage:
Enabling auto-merge as two LGTM's and confirmed passing tests.

Issue #11679, opened 2025-03-17
Context dropdown doesn't select


DavDeDev:
### Bug description

I am trying to select colima context, but everytime i navigate to container and then back to settings it goes back to default

### Operating system

macOS 24.3.0 - arm64

### Installation Method

None

### Version

1.17.0

### Relevant log output

```shell

```

### Additional context

**Enabled Extensions**
- podman-desktop.compose
- podman-desktop.docker
- podman-desktop.kind
- podman-desktop.kube-context
- podman-desktop.kubectl-cli
- podman-desktop.lima
- podman-desktop.podman
- podman-desktop.registries

danivilla9:
Hello! We need more info in order to reproduce your issue. Can you elaborate on the following topics?
- How are you generating the colima context? We've never worked with that one before
- I suppose that the context dropdown you are referring to is the one that appears after clicking on the "Current kubernetes context" button in the status bar, is that right?
- What do you mean by "default" context?
- Is it possible for you to make a video showing the unexpected behavior? 
Thanks in advance

DavDeDev:
Before this video, I have run `docker context set colima` therefore it defaults to Colima every time I navigate another tab.
https://github.com/user-attachments/assets/6509faa3-6e09-4be1-b2ed-b8b17c28debf

danivilla9:
Hello, just managed to reproduce what you experimented, it's definitely a bug on PD's side. Bear in mind that this is an experimental feature and this is somewhat expected. The dev team will tackle it when they had the chance, thanks for reporting it!

Note that this issue is not related exclusively to colima. I've tried with two default docker contexts and I experienced the same problem.

danivilla9:
Detailed steps to reproduce the issue:
1-Install docker, it should start running and create the default context
2-Create a new context, through cli for example: `docker context create my-test-context`
3-Enable the "Docker Compatibility" experimental feature
4-Go to Settings/Docker Compatibility
5-Select a context other than the already selected using the dropdown menu
6-Move to a different page
7-Go back to the previous page, you should see that the context you changed to now is not selected and the previous one is being selected

PR #11895, opened 2025-03-28
refactor: extract code to a method + use async

benoitf:
### What does this PR do?
Extract code to a method and use async rather than promise.then()

prior to https://github.com/podman-desktop/podman-desktop/issues/11679 fix

### How to test this PR?


benoitf:
> Consider renaming refreshContextList to initializeContextList to better reflect its purpose.

it'll be called later (for the fix of the related issue) to refresh the list, this is why I called it refresh

deboer-tim (approved):

PR #11901, opened 2025-03-28
chore: props interface

deboer-tim:
### What does this PR do?

We decided to use a Props interface for Svelte 5 component exports, but we had a few changes before that and a few slipped in since. This just cleans it up to be consistent and make sure we (i.e. me) don't accidentally copy/paste.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Just code review.

### DIFF
--- a/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
+++ b/packages/renderer/src/lib/image/ImageDetailsFiles.svelte
@@ -11,7 +11,10 @@ import FilesystemLayerView from './FilesystemLayerView.svelte';
 import { type ImageFilesystemLayerUI, toImageFilesystemLayerUIs } from './imageDetailsFiles';
 import ImageDetailsFilesLayers from './ImageDetailsFilesLayers.svelte';
 
-let { imageInfo }: { imageInfo: ImageInfo | undefined } = $props();
+interface Props {
+  imageInfo: ImageInfo | undefined;
+}
+let { imageInfo }: Props = $props();
 
 let imageLayers = $state<ImageFilesystemLayers>();
 let selectedLayer = $state<ImageFilesystemLayerUI>();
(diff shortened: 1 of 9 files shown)


packages/renderer/src/lib/image/ImageDetailsFiles.svelte
```diff
@@ -11,7 +11,10 @@ import FilesystemLayerView from './FilesystemLayerView.svelte';
 import { type ImageFilesystemLayerUI, toImageFilesystemLayerUIs } from './imageDetailsFiles';
 import ImageDetailsFilesLayers from './ImageDetailsFilesLayers.svelte';
 
-let { imageInfo }: { imageInfo: ImageInfo | undefined } = $props();
+interface Props {
+  imageInfo: ImageInfo | undefined;
```
benoitf:
```suggestion
  imageInfo?: ImageInfo;
```

packages/renderer/src/lib/ui/ProviderInfoCircle.svelte
```diff
@@ -1,7 +1,10 @@
 <script lang="ts">
 import { providerColors } from './ProviderInfoCircle';
 
-let { type }: { type: 'kubernetes' | 'podman' | 'docker' | undefined } = $props();
+interface Props {
+  type: 'kubernetes' | 'podman' | 'docker' | undefined;
```
benoitf:
```suggestion
  type?: 'kubernetes' | 'podman' | 'docker';
```

benoitf (commented):
some nits

packages/renderer/src/lib/image/ImageDetailsFiles.svelte
deboer-tim: Sure, props fixed in new commit.

packages/renderer/src/lib/ui/ProviderInfoCircle.svelte
deboer-tim: Props fixed in new commit.

benoitf (approved):

PR #11930, opened 2025-03-31
chore: migrate tooltip to svelte 5

### LINKED ISSUE #11587: Migrate UI component tooltip to Svelte 5
https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/tooltip/Tooltip.svelte#L1

https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/tooltip/TooltipSpec.svelte#L1

deboer-tim:
### What does this PR do?

Migrates the tooltip component to Svelte 5.

The original Tooltip had both a property and a slot with the same name, 'tip'. This isn't possible in Svelte 5 so I changed the slot to a snippet 'tipSnippet'.

There was a typo in PreferenceCliTool, but it was clearly expecting the component to support the aria-label property. I've added that to avoid the error, but this means we have the same linting issue as the Link PR, will hold off merging until we have a direction on that.

### Screenshot / video of UI

N/A

Fixes #11587.

### How to test this PR?

Just code review.


axel7083 (commented):
Yeah, we cannot shadow component props and snippets (https://svelte.dev/docs/svelte/compiler-errors#snippet_shadowing_prop) 

I am wondering instead of renaming the slot (from tip to tipSnippet) we should not rename the tip (string) to something like `title`? would make sense as the button is title.

I can't find much documentation online, on what a _existing_ solution would be 🤔 

- https://www.shadcn-svelte.com/docs/components/tooltip
- https://svelte.dev/playground/dd6754a2ad0547c5b1c1ea37c0293fef?version=5.25.3
- https://flowbite-svelte.com/docs/components/tooltip

So many different terms used, `trigger`, `title` etc :/

React has a different approach (using anchors, so not really usefull to use https://react-tooltip.com/docs/getting-started)

packages/ui/src/lib/tooltip/Tooltip.svelte
```diff
@@ -50,22 +50,46 @@
...
+  topRight?: boolean;
+  right?: boolean;
+  bottom?: boolean;
+  bottomLeft?: boolean;
+  bottomRight?: boolean;
+  left?: boolean;
+  class?: string;
+  'aria-label'?: string;
+  tipSnippet?: Snippet;
+  children?: Snippet;
```
feloy: One is suffixed with `Snippet`, not the other. 
In https://github.com/podman-desktop/podman-desktop/pull/11889, I took the side of suffixing all snippets, I'm not sure if this is what we want, but IMO we'll need to agree on a way to avoid having duplicates between Snippets and other props, as this may happen frequently

deboer-tim:
> I am wondering instead of renaming the slot (from tip to tipSnippet) we should not rename the tip (string) to something like `title`? would make sense as the button is title.

I agree we could probably pick a better name for the property, and out of the options I'd probably go for something like title or text. OTOH someone keeps reminding me '1 PR 1 issue' :) and renaming the 30+ consumers vs the 2 that use the snippet feels more than svelte migration?

> In https://github.com/podman-desktop/podman-desktop/pull/11889, I took the side of suffixing all snippets, I'm not sure if this is what we want, but IMO we'll need to agree on a way to avoid having duplicates between Snippets and other props, as this may happen frequently

I could be wrong, but I can't recall any other places where we overloaded like this in Svelte 4.

I don't like the idea of suffixing all snippets (and the default child snippet `children` doesn't do this) but I think it makes sense or is more readable in some cases.

deboer-tim:
Rebased to pick up linter fix.

benoitf:
looks like this PR needs a rebase (and reviews)

deboer-tim:
Rebased, still looking for review from @feloy and @axel7083.

axel7083 (approved):

packages/ui/src/lib/tooltip/Tooltip.svelte
benoitf: cc @deboer-tim : this comment is unanswered

deboer-tim: Sorry, it was answered below in a comment instead of inline: I generally don't like suffixing all snippets (readability; 'children' doesn't do it), plus it breaks compatibility. However since the tip property was overloaded in Svelte 4 as both a simple property and a slot I thought we were forced to rename one of them in this case (and better to break the 2 users vs 30).

Except of course we don't. I took another look at this and it's still possible for the property to be string | Snippet. I've applied this in a new commit, and was able to merge in the two extra TooltipTest.* files that were needed for Svelte 4.

The test for string vs Snippet is taken from here: https://github.com/sveltejs/svelte/issues/9774#issuecomment-2266343301. It's quite weak in this case, but overall I think this is a better, cleaner path and avoids migration.

packages/ui/src/lib/tooltip/Tooltip.spec.ts
```diff
@@ -19,13 +19,43 @@
...
+  expect(element).toHaveClass('border-[1px]');
+});
+
+test('Expect basic slot styling', async () => {
+  render(Tooltip, {
+    tip: createRawSnippet(() => {
+      return {
+        render: (): string => 'test',
+      };
+    }),
```
axel7083, line 47: if you are not looking for the `test` string, you can probably just use `tip: vi.fn()`

packages/ui/src/lib/tooltip/Tooltip.spec.ts
```diff
@@ -40,10 +70,49 @@ test('tooltip is not empty string when tooltipHidden value false', async () => {
...
+}
+
+createTest({ left: true }, 'left');
+createTest({ right: true }, 'right');
+createTest({ bottom: true }, 'bottom');
+createTest({ top: true }, 'top');
+createTest({ topLeft: true }, 'topLeft', 'top-left');
+createTest({ topRight: true }, 'topRight', 'top-right');
+createTest({ bottomLeft: true }, 'bottomLeft', 'bottom-left');
+createTest({ bottomRight: true }, 'bottomRight', 'bottom-right');
```
axel7083: does not seems to match our current testing practice 

You can take a look at test.each documentation https://vitest.dev/api/#test-each

axel7083 (changes requested):
Tests in packages/ui/src/lib/tooltip/Tooltip.spec.ts feel a bit weird to me

packages/ui/src/lib/tooltip/Tooltip.spec.ts
deboer-tim: Although you don't need the full / real snippet, these two tests are specifically testing the snippet code path and fall back to string tip with vi.fn(). So, could use something else, but didn't think it was worth mocking Snippet.

packages/ui/src/lib/tooltip/Tooltip.spec.ts
deboer-tim: Tests were unchanged, just moved from other file. Updated to use test.each via new commit.

packages/ui/src/lib/tooltip/Tooltip.svelte
```diff
@@ -50,22 +50,47 @@
...
+  bottomRight = false,
+  left = false,
+  class: className = '',
+  'aria-label': ariaLabel,
+  children = undefined,
+}: Props = $props();
+
+function isSnippet(obj: string | Snippet): obj is Snippet {
+  return obj.length === 1;
+}
```
axel7083, line 88: This is a blocker for me, this is not officially supported and therefore could break in a near future - we should not allow to send anything else that a Snippet as children.

benoitf:
moving to draft as not updated since a long time

benoitf:
closing as not updated

PR #11961, opened 2025-04-01
chore: migrate status icon to svelte 5

### LINKED ISSUE #11584: Migrate UI component StatusIcon to Svelte 5
https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/statusIcon/StatusIcon.svelte#L1

deboer-tim:
### What does this PR do?

Migrate the status icon component to Svelte 5.

svelte:component has been removed. In order to support it being string | Component (and needing capitalization in the second case), I've used the technique provided by the Svelte team here: https://github.com/sveltejs/svelte/issues/12668#issuecomment-2563451764

Typing of icon could be improved but deferring that as it is a bigger change and
not related to Svelte migration.

### Screenshot / video of UI

N/A

Fixes #11584.

### How to test this PR?

Check status icons in various screens.

### DIFF
--- a/packages/ui/src/lib/progress/Spinner.svelte
+++ b/packages/ui/src/lib/progress/Spinner.svelte
@@ -4,15 +4,15 @@ interface Props {
   class?: string;
   style?: string;
 }
-let { size = '2em', class: classProp, style: styleProp }: Props = $props();
+let { size = '2em', class: className, style }: Props = $props();
 </script>
 
 <i
   role="progressbar"
   aria-label="Loading"
   aria-busy="true"
-  class="flex justify-center items-center {classProp}"
-  style={styleProp}>
+  class="flex justify-center items-center {className}"
+  style={style}>
   <svg width={size} height={size} viewBox="0 0 100 100" role="img">
     <defs>
       <linearGradient id="grad1" x1="0%" y1="0%" x2="0%" y2="100%">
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM

Suggestion: remove usage of any

deboer-tim:
> Suggestion: remove usage of any

I tried briefly but typing changes cause a big ripple to most consumers, some of which cause other issues since they are on Svelte 4 types. We should do it, but since it is not strictly Svelte 5 migration I couldn't justify the big change here. Plus, will be easier later when most consumers are on the same Svelte version.

packages/ui/src/lib/progress/Spinner.svelte
```diff
@@ -1,13 +1,18 @@
 <script lang="ts">
-export let size = '2em';
+interface Props {
+  size?: string;
+  class?: string;
+  style?: string;
+}
+let { size = '2em', class: classProp, style: styleProp }: Props = $props();
```
axel7083:
```suggestion
let { size = '2em', class: className, style: styleProp }: Props = $props();
```

We seems to started using className is other places, might need to add it to CODE_GUIDELINES, or discuss if there is different opinion

axel7083: Do we need to rename style? class is a JS keyword so make sense, but slyle is okay?

deboer-tim: Both good points, I've renamed to className and removed the style renaming.

benoitf (approved):

PR #11964, opened 2025-04-01
chore: support for changing namespace in experimental mode

deboer-tim:
### What does this PR do?

Add support for experimental mode reachable status to the namespace dropdown.

I'm still seeing some reactivity issues in experimental mode - e.g.:
- switching to a namespace with no resources doesn't clear the current resources, but namespace appears set since switching pages or after leaving and coming back everything looks correct.
- badge still says connection lost after reconnection (but namespace reappears and changing it triggers refresh).

AFAIK these are still somewhat expected in experimental mode; when testing you may need to play around to confirm this is working vs other issue.

### Screenshot / video of UI

N/A

Experimental part of #11274.

### How to test this PR?

Confirm no regressions / able to use namespace dropdown in regular Kubernetes mode.
Enable Kubernetes experimental mode and restart, confirm this now works as well.

### DIFF
--- a/packages/renderer/src/lib/kube/NamespaceDropdown.svelte
+++ b/packages/renderer/src/lib/kube/NamespaceDropdown.svelte
@@ -1,14 +1,36 @@
 <script lang="ts">
 import { Dropdown } from '@podman-desktop/ui-svelte';
+import { onMount } from 'svelte';
 
+import { kubernetesContextsHealths } from '/@/stores/kubernetes-context-health';
 import { kubernetesCurrentContextState } from '/@/stores/kubernetes-contexts-state';
 
+let isExperimental: boolean = $state(false);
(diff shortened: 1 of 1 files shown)


feloy (approved):
LGTM

The problem of the list not being refreshed when jumping to a namespace with no resource is clearly a problem on the `KubernetesObjectsList` component not supporting this use case, but not related to this PR

deboer-tim:
Rebased on main to pick up test fixes, no change.

cdrage (approved):
LGTM! Tested on experimental mode and refreshed all the namespaces perfectly.

No concerns / comments regarding code side.


https://github.com/user-attachments/assets/e9a6bb3c-0f23-4814-9817-0a70ebb6a855

PR #11972, opened 2025-04-02
refactor: migrate FormPage component to svelte 5

feloy:
### What does this PR do?

Migrate FormPage component to svelte 5

EngineFormPage (using FormPage) will be migrated in a follow-up PR

### Screenshot / video of UI

N/A

Part of #11579 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/ui/EngineFormPage.svelte
+++ b/packages/renderer/src/lib/ui/EngineFormPage.svelte
@@ -8,10 +8,14 @@ export let showEmptyScreen: boolean = false;
 </script>
 
 <FormPage title={title} inProgress={inProgress}>
-  <slot slot="icon" name="icon" />
-  <slot slot="actions" name="actions" />
-  <slot slot="tabs" name="tabs" />
-  <slot slot="content">
+  {#snippet icon()}
+  <slot name="icon" />
+  {/snippet}
+  {#snippet actions()}
+  <slot name="actions" />
+  {/snippet}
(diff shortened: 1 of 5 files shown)


jeffmaury (approved):
LGTM but description does not seem to be update

feloy:
> LGTM but description does not seem to be update

Not sure which part of the description. If it is about EngineFormPage, it has been updated to use snippets, but still needs to be migrated to svelte 5 itself (declaring snippets)

packages/renderer/src/lib/ui/EngineFormPage.svelte
```diff
@@ -8,10 +8,14 @@ export let showEmptyScreen: boolean = false;
 </script>
 
 <FormPage title={title} inProgress={inProgress}>
-  <slot slot="icon" name="icon" />
-  <slot slot="actions" name="actions" />
-  <slot slot="tabs" name="tabs" />
-  <slot slot="content">
+  {#snippet icon()}
+  <slot name="icon" />
```
axel7083, line 12: question: is it necessary to keep the slot in it? like is this for backward compatibility?

feloy: This slot is necessary, as this component is still in svelte 4.

When migrated to svelte 5, this part will be:

```
{#snippet icon()}{@render localIcon?.()}{/snippet}
```

axel7083 (approved):

PR #11976, opened 2025-04-02
fix(PinRegistry): listeners and VM support

### LINKED ISSUE #11974: bug(PinRegistry): do not notify when connections are registered / unregistered
### Bug description

When I wrote the PinRegistry, I did not try to create new kubernetes, or podman machine, so I missed this part. I used the following events (thinking it was enough), but it was not.

https://github.com/podman-desktop/podman-desktop/blob/89a4a47f0a22d942f5b80b6f12dee72093ec358b/packages/main/src/plugin/statusbar/pin-registry.ts#L125-L126

I need to use the following events `onDidRegister(Container | Kubernetes | vm)Connection` and `onDidUnregister(Container | Kubernetes | vm)Connection`

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

Noticed while testing https://github.com/podman-desktop/podman-desktop/pull/11973 and creating new kubernetes cluster

### Relevant log output

```shell

```

### Additional context

_No response_

axel7083:
### What does this PR do?

While testing https://github.com/podman-desktop/podman-desktop/pull/11973 I noticed that the pin registry was not updating when a new connection is registered, or one is unregistered.

- This PR fixes that, by registering to all register, unregister event.
- It also add support for VM providers (can be in a separate PR, but this a one liner, and I also add the register, unregister event related to VM so... make sense to update all)

### Screenshot / video of UI

N/A

Fixes https://github.com/podman-desktop/podman-desktop/issues/11974

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/statusbar/pin-registry.ts
+++ b/packages/main/src/plugin/statusbar/pin-registry.ts
@@ -68,7 +68,12 @@ export class PinRegistry implements IDisposable {
   public getOptions(): Array<PinOption> {
     return this.providers
       .getProviderInfos()
-      .filter(provider => provider.containerConnections.length > 0 || provider.kubernetesConnections.length > 0)
+      .filter(
+        provider =>
+          provider.containerConnections.length > 0 ||
+          provider.kubernetesConnections.length > 0 ||
+          provider.vmConnections.length > 0,
+      )
       .map(provider => ({
         value: provider.id,
         label: provider.name,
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/statusbar/pin-registry.ts
```diff
@@ -121,8 +126,15 @@ export class PinRegistry implements IDisposable {
       this.notify();
     }
 
-    // notify if container connection / kubernetes connection changed
-    this.#disposables.push(this.providers.onDidUpdateContainerConnection(this.notify.bind(this)));
-    this.#disposables.push(this.providers.onDidUpdateKubernetesConnection(this.notify.bind(this)));
```
SoniaSandler, line 126: question (nit): would using `onDidUpdate` in addition to `onDidRegister` and `onDidUnregister` in this case be redundant/not useful?

SoniaSandler (approved):
LGTM

packages/main/src/plugin/statusbar/pin-registry.ts
axel7083: onDidUpdate is triggered on status update (from my understanding) so it would not change the content of the PinRegistry options, as we only keep the providerId 🤔 .

PR #11979, opened 2025-04-02
chore(dockerode): bump to 3.3.37

### LINKED ISSUE #11977: Update BuildImageOptions#pull type to match Dockerode and Podman definition

Following https://github.com/podman-desktop/podman-desktop/pull/11969, seems that the `pull` property is now a boolean

https://github.com/podman-desktop/podman-desktop/blob/08f1e32c5aeb648fcfa3cb39a73c2fabb46082de/packages/main/src/plugin/container-registry.ts#L2593

We define it as a string | undefined on our internal api(image-info.ts#`BuildImageOptions`)

https://github.com/podman-desktop/podman-desktop/blob/107ddf645dd70bb808e4d1b774cf1a504aa92762/packages/api/src/image-info.ts#L88

PR making the change is https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71978

We should follow podman values as a reference

Reading the documentation of podman libpodapi also makes it being a boolean (ref https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImageBuildLibpod)

image

_No response_

### Additional context

_No response_

axel7083:
### What does this PR do?

Supersede https://github.com/podman-desktop/podman-desktop/pull/11969 and update the pull property of the `BuildImageOptions ` both in our internal api and extension api.

We don't seems to use this option anywhere, the `window.buildImage` do not provide it, and we just omit it when we call the dockerode from the renderer, only the extension-api may be using it.

Fixes https://github.com/podman-desktop/podman-desktop/issues/11977

### How to test this PR?

- typecheck and tests should be ✅

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -3510,7 +3510,7 @@ declare module '@podman-desktop/api' {
     /**
      * Attempt to pull the image even if an older image exists locally.
      */
-    pull?: boolean;
+    pull?: boolean | string;
 
     /**
      * Default: true
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -3510,7 +3510,7 @@ declare module '@podman-desktop/api' {
     /**
      * Attempt to pull the image even if an older image exists locally.
      */
-    pull?: string;
+    pull?: boolean;
 
     /**
      * Default: true
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "@rollup/plugin-typescript": "^12.1.2",
     "@tsconfig/strictest": "^2.0.5",
     "@types/adm-zip": "^0.5.7",
-    "@types/dockerode": "^3.3.36",
+    "@types/dockerode": "^3.3.37",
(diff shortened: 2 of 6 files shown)


jeffmaury (commented):
Seems to be a breaking change ?

axel7083:
> Seems to be a breaking change ?

Yeahh, I don't know how we can deal with that :/

packages/extension-api/src/extension-api.d.ts
```diff
@@ -3510,7 +3510,7 @@ declare module '@podman-desktop/api' {
     /**
      * Attempt to pull the image even if an older image exists locally.
      */
-    pull?: string;
+    pull?: boolean;
```
benoitf: for now needs to be `string | boolean` there

and later if a string we convert it to true or false

benoitf (changes requested):

packages/extension-api/src/extension-api.d.ts
axel7083: I updated the extension loader to properly convert potential string to boolean

gastoner (approved):
Codewise LGTM

benoitf (approved):

PR #11981, opened 2025-04-02
refactor: component EngineFormPage to svelte 5

feloy:

### What does this PR do?

Refactor component EngineFormPage to svelte 5

### Screenshot / video of UI

N/A

Part of #11579 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/container/ContainerExport.svelte
+++ b/packages/renderer/src/lib/container/ContainerExport.svelte
@@ -87,42 +87,46 @@ async function exportContainer(): Promise<void> {
 
 {#if container}
   <EngineFormPage title="Export container {container.name}">
-    <svelte:fragment slot="icon">
+    {#snippet icon()}
       <i class="fas fa-download fa-2x" aria-hidden="true"></i>
-    </svelte:fragment>
-
-    <div slot="content" class="space-y-2">
-      <div>
-        <label for="modalSelectTarget" class="block mb-2 text-sm font-medium text-[var(--pd-content-card-header-text)]"
-          >Export to:</label>
-        <div class="flex w-full">
-          <Input
-            class="grow mr-2"
-            name={container.id}
-            readonly
-            value={outputTarget}
-            id="input-export-container-name"
-            aria-invalid={invalidFolder} />
-          <Button
-            on:click={selectFolderPath}
-            title="Open dialog to select the output file"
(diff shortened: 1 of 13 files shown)


packages/renderer/src/lib/container/ContainerExport.svelte
```diff
@@ -87,42 +87,46 @@ async function exportContainer(): Promise<void> {
...
-          Export Container
-        </Button>
-        <div aria-label="createError">
-          {#if exportedError}
-            <ErrorMessage class="py-2 text-sm" error={exportedError} />
-          {/if}
+    {/snippet}
+      
+    {#snippet content()}
+      {#if container}
```
axel7083, line 95: Why do we duplicate the if check, if we have it at line 88

feloy:
```
podman-desktop/packages/renderer/src/lib/container/ContainerExport.svelte:102:23
Error: 'container' is possibly 'undefined'. (ts)
                class="grow mr-2"
                name={container.id}
                readonly


====================================
svelte-check found 1 error and 0 warnings in 1 file
 ELIFECYCLE  Command failed with exit code 1.
```

I suppose that, as a snippet intended to be executed somewhere else, `ts` cannot guarantee that `container` will remain defined there. I'm not sure why it was not the case for `<slot>`...

jeffmaury (approved):
LGTM

packages/renderer/src/lib/container/ContainerExport.svelte
axel7083: okay, I will approve, but feels like a bug, it should be able to capture the context :/

axel7083 (approved):
LGTM 🚀

packages/renderer/src/lib/container/ContainerExport.svelte
feloy: ok, seems that we need to pass parameters to the snippet to fix the problem (https://github.com/sveltejs/language-tools/issues/1832#issuecomment-2260430766)

I'll fix the problem, and fix what I have done in another PR already merged (https://github.com/podman-desktop/podman-desktop/pull/11889)

feloy: the result in https://github.com/podman-desktop/podman-desktop/pull/12037 is not better, I'll merge this PR as is

Issue #10224, opened 2024-12-04
UX: Reminder notification for sharing feedback on experimental feature


slemeur:
### UX Description

When a user will have activated an experimental feature, we would like to encourage the user to share their feedback about their experience with this new capability.

We will need to design a feedback reminder notification.
The notification must allow the user to share a "lazy feedback" (a simple vote +1 / -1, or on a range from 0 to 5) and propose to the user to join the discussion once the lazy feedback has been shared.

Along with this issue we would like to suggest a solution for:
- When to display the notification to the end user
- How to handle the case, where the user will have multiple experimental features activated
- How do we avoid not sending again the notification

### Request type

Application mockups/designs

### Primary Contact

@slemeur

MariaLeonova:
PTAL at [this flow](https://miro.com/app/board/uXjVLC-mQjE=/?moveToWidget=3458764614889339750&cot=14), does it make sense?

![Image](https://github.com/user-attachments/assets/4b25ae20-7e97-41c9-893d-fa7ff07c7329)

My main question is: can we track the feature usage? How can we tell if it has been 'used'? Can there also be a permanent way to give feedback inside a feature?

MariaLeonova:
So if we were to follow the existing format of notifications, we could do a notifications with:
1. thumbs up or thumbs down
2. smiley 1-5 
3. with or without a field to provide feedback 

Current UI has a 1-5 range and we want to prompt them to join discussion on GitHub, so I suggest this design: 

![Image](https://github.com/user-attachments/assets/dd8e547a-b8d2-44fc-be4b-91c053ee42d6)

The `Send feedback` button only becomes active after the user picks a smiley. 

One thing I'd like to consider: if they are not ready to give feedback at this point, can we offer them an option to defer the notification to later? That would mean having one more option, which might be too many buttons to have on one notification. I'll leave it here for your consideration as an option:

![Image](https://github.com/user-attachments/assets/1f2b4d8a-0c63-4d2f-818a-81bbb7e867ff)

Another option would be to show them this message 3-4 times at intervals of a couple days and if they keep deferring, give an option to not show the message anymore: 

![Image](https://github.com/user-attachments/assets/cc59bf89-d6fd-4b32-a888-15901ca97d9b)

If there are multiple notifications, one can consider also adding them to a notification drawer, similar to the Tasks UI:

[...]

PR #11988, opened 2025-04-02
chore: added overflow visible property to modal component

### LINKED ISSUE #10228: Experimental features feedback notification

gastoner:
### What does this PR do?
Adds overflow visible property to modal component

Closes https://github.com/podman-desktop/podman-desktop/issues/10228

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/modal/Modal.svelte
+++ b/packages/ui/src/lib/modal/Modal.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { createEventDispatcher, onDestroy, type Snippet } from 'svelte';
+import { createEventDispatcher, onDestroy } from 'svelte';
 
 import { tabWithinParent } from '../utils/dialog-utils';
 
@@ -9,8 +9,8 @@ interface Props {
   name?: string;
   top?: boolean;
   ignoreFocusOut?: boolean;
+  overflowVisible?: boolean;
   onclose?: () => void;
-  children?: Snippet;
 }
 
 let modal: HTMLDivElement;
@@ -18,10 +18,10 @@ let {
   name = 'drop-down-dialog',
   top = false,
   ignoreFocusOut = false,
+  overflowVisible = false,
   onclose = (): void => {
     dispatch('close');
   },
-  children,
 }: Props = $props();
 
 const handle_keydown = (e: KeyboardEvent): void => {
@@ -66,11 +66,11 @@ function handleMousedown(e: MouseEvent): void {
   <div
     class:translate-y-[-5%]={!top}
     class:my-[32px]={top}
-    class="bg-[var(--pd-modal-bg)] z-50 rounded-xl overflow-auto w-[calc(200vw-4em)] h-fit max-w-[42em] max-h-[calc(100vh-4em)] border-[1px] border-[var(--pd-modal-border)]"
+    class="bg-[var(--pd-modal-bg)] z-50 rounded-xl {overflowVisible ? 'overflow-visible' : 'overflow-auto'} w-[calc(200vw-4em)] h-fit max-w-[42em] max-h-[calc(100vh-4em)] border-[1px] border-[var(--pd-modal-border)]"
     role="dialog"
     aria-label={name}
     aria-modal="true"
     bind:this={modal}>
-    {@render children?.()}
+    <slot />
   </div>
 </div>
--- a/packages/ui/src/lib/modal/Modal.svelte
+++ b/packages/ui/src/lib/modal/Modal.svelte
@@ -9,7 +9,6 @@ interface Props {
   name?: string;
   top?: boolean;
   ignoreFocusOut?: boolean;
-  overflow?: 'auto' | 'visible';
   onclose?: () => void;
   children?: Snippet;
 }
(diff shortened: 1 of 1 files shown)


deboer-tim:
I know this is still in draft and I don't immediately know what a better solution is (maybe button bar contributed separately?), but it is ugly for the client to have to decide whether it will be too big for the dialog and require scrolling, or overflow outside of the modal boundaries. If we ever change modal styling or size, anything using this could start overflowing and tests wouldn't fail either.

gastoner:
> I know this is still in draft and I don't immediately know what a better solution is (maybe button bar contributed separately?), but it is ugly for the client to have to decide whether it will be too big for the dialog and require scrolling, or overflow outside of the modal boundaries. If we ever change modal styling or size, anything using this could start overflowing and tests wouldn't fail either.

Without this change is the dropdown cutted of I think. I don't think it could overflow, since I'm explicitly setting the variable for this. Without this change I'm not sure how should I do this design https://github.com/podman-desktop/podman-desktop/issues/10224#issuecomment-2636084884

deboer-tim:
IMHO it's probably better to always allow overflow (remove the auto and don't add the property to control it) and ensure that all current users of Modal were not reliant on this / handle their own overflow. I took a quick look and there are 5 users. Dialog and QuickPick are good, the other three (feedback form and provider dialogs) should be extending Dialog instead of Modal directly. So: either we fix that now and then this change is simple, or you have to make sure the overflow/scrolling is handled correctly in those three.

gastoner:
> IMHO it's probably better to always allow overflow (remove the auto and don't add the property to control it) and ensure that all current users of Modal were not reliant on this / handle their own overflow. I took a quick look and there are 5 users. Dialog and QuickPick are good, the other three (feedback form and provider dialogs) should be extending Dialog instead of Modal directly. So: either we fix that now and then this change is simple, or you have to make sure the overflow/scrolling is handled correctly in those three.

@deboer-tim those looks LGTM with `overflow-visible` in Modal component (not sure how to check provider dialogs) 
image


[Screencast_20250407_074845.webm](https://github.com/user-attachments/assets/0ac2118b-a151-47bf-8fea-29101b9d1585)

gastoner:
@axel7083 @cdrage @benoitf can you PTAL?

packages/ui/src/lib/modal/Modal.svelte
```diff
@@ -66,11 +66,11 @@ function handleMousedown(e: MouseEvent): void {
...
     class:translate-y-[-5%]={!top}
     class:my-[32px]={top}
-    class="bg-[var(--pd-modal-bg)] z-50 rounded-xl overflow-auto w-[calc(200vw-4em)] h-fit max-w-[42em] max-h-[calc(100vh-4em)] border-[1px] border-[var(--pd-modal-border)]"
+    class="bg-[var(--pd-modal-bg)] z-50 rounded-xl {overflowVisible ? 'overflow-visible' : 'overflow-auto'} w-[calc(200vw-4em)] h-fit max-w-[42em] max-h-[calc(100vh-4em)] border-[1px] border-[var(--pd-modal-border)]"
     role="dialog"
     aria-label={name}
     aria-modal="true"
     bind:this={modal}>
-    {@render children?.()}
+    <slot />
```
axel7083: I don't understand the change children => slot

axel7083: does not match the PR title

packages/ui/src/lib/modal/Modal.svelte
```diff
@@ -9,19 +9,19 @@ interface Props {
...
   onclose?: () => void;
-  children?: Snippet;
 }
 
 let modal: HTMLDivElement;
 let {
   name = 'drop-down-dialog',
   top = false,
   ignoreFocusOut = false,
+  overflowVisible = false,
```
axel7083: I feel like having something like `overflow?: 'auto' | 'visible'` is more resilient: if we need to add change, we can just add a value to the allowed type

packages/ui/src/lib/modal/Modal.svelte
gastoner: What? How did this happen  :D

axel7083 (approved):

packages/ui/src/lib/modal/Modal.svelte
deboer-tim: FWIW I would still remove entirely and keep the API contract/code simple.

axel7083: I don't have strong opinion on this, so either way is okay for me (removing or use what the current commit is)

gastoner:
@deboer-tim when changed to explocit overflow-visible I was not able to see any overfloating 

> > IMHO it's probably better to always allow overflow (remove the auto and don't add the property to control it) and ensure that all current users of Modal were not reliant on this / handle their own overflow. I took a quick look and there are 5 users. Dialog and QuickPick are good, the other three (feedback form and provider dialogs) should be extending Dialog instead of Modal directly. So: either we fix that now and then this change is simple, or you have to make sure the overflow/scrolling is handled correctly in those three.
> 
> @deboer-tim those looks LGTM with `overflow-visible` in Modal component (not sure how to check provider dialogs) image
> Screencast_20250407_074845.webm

Should I change to `overflow-visible` then?

gastoner:
@deboer-tim @axel7083 switched from auto to visible

axel7083 (approved):

deboer-tim (approved):

PR #11990, opened 2025-04-02
chore: adds openExternal command to registered commands

gastoner:
### What does this PR do?
Registers openWebsite command  

Part of https://github.com/podman-desktop/podman-desktop/issues/10228

### How to test this PR?


gastoner:
@feloy @benoitf @dgolovin can you PTAL?

benoitf (changes requested):
hello,

first question: what would do openWebsite on the frontend side ? (it's not implemented in this PR)


also, I'm not sure to follow,
we register a command in the backend, we send the data to the frontend and I suspect the frontend will call again the backend to open the link ?

next, we do have `openExternal` in the extension API

so I would register openExternal accepting an URI object as well

and I would keep all the logic on the backend

command openExternal would call shell.openExternal

gastoner:
> hello,
> 
> first question: what would do openWebsite on the frontend side ? (it's not implemented in this PR)
> 
> also, I'm not sure to follow, we register a command in the backend, we send the data to the frontend and I suspect the frontend will call again the backend to open the link ?
> 
> next, we do have `openExternal` in the extension API
> 
> so I would register openExternal accepting an URI object as well
> 
> and I would keep all the logic on the backend
> 
> command openExternal would call shell.openExternal

I'f I would use shell.openExternal it will open the website 
And not open additional dialog like is on the image in the issue (not sure if this is a problem or not )
image

But I agree that calling F->B->F->B is wrong (I didn't realize that)

benoitf:
@gastoner FYI the dialog "Open External Website" is automatically displayed when you're using shell.openExternal on an unknown/non-approved URL

gastoner:
> @gastoner FYI the dialog "Open External Website" is automatically displayed when you're using shell.openExternal on an unknown/non-approved URL

Ohh, thank you for clarifying

benoitf (changes requested):
the command should be openExternal and it should not be part of feedback-form

the command has no relationship to feedback
also I don't understand why the command would handle an array of links

it should be one call = one link to open

gastoner:
@benoitf , @dgolovin , @feloy can you PTAL?

benoitf (changes requested):
as reported earlier, it needs to be openExternal, provided by a sub-component so it's easily testable (not from index.ts) and there is a missing unit test

benoitf (changes requested):
it's also missing type for arg (that should be URI)

packages/main/src/plugin/commands-init.ts
```diff
@@ -106,6 +109,14 @@ export class CommandsInit implements IDisposable {
         },
       ),
     );
+
+    this.#disposables.push(
+      commandRegistry.registerCommand('openWebsite', async (arg: Uri) => {
```
axel7083: why the command is called openWebsite, when it can be named openExternal ?

axel7083 (approved):
LGTM

benoitf (approved):
thanks for the changes

PR #11991, opened 2025-04-02
chore: added selected option if dropdown was selected

gastoner:
### What does this PR do?
Adds selected `dropdownIndex: number | undefined` to differentiate between options if dropdown was selected
Adds test to test this feature
And fixing tests (adding option to other tests), otherwise is linter/typecheck failing

Part of https://github.com/podman-desktop/podman-desktop/issues/10228

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -161,7 +161,6 @@ import { ExtensionDevelopmentFolders } from './extension/extension-development-f
 import { ExtensionsUpdater } from './extension/updater/extensions-updater.js';
 import { Featured } from './featured/featured.js';
 import type { FeaturedExtension } from './featured/featured-api.js';
-import { FeedbackForm } from './feedback-form.js';
 import { FeedbackHandler } from './feedback-handler.js';
 import { FilesystemMonitoring } from './filesystem-monitoring.js';
 import { IconRegistry } from './icon-registry.js';
@@ -174,7 +173,7 @@ import { KubernetesClient } from './kubernetes/kubernetes-client.js';
 import { downloadGuideList } from './learning-center/learning-center.js';
 import { LearningCenterInit } from './learning-center-init.js';
 import { LibpodApiInit } from './libpod-api-enable/libpod-api-init.js';
-import type { MessageBoxOptions, MessageBoxReturnValue, RemindOption } from './message-box.js';
+import type { MessageBoxOptions, MessageBoxReturnValue } from './message-box.js';
 import { MessageBox } from './message-box.js';
 import { NavigationItemsInit } from './navigation-items-init.js';
 import { OnboardingRegistry } from './onboarding-registry.js';
@@ -725,8 +724,6 @@ export class PluginSystem {
     await this.extensionLoader.init();
 
     const feedback = new FeedbackHandler(this.extensionLoader);
-    const feedbackForm = new FeedbackForm(commandRegistry, apiSender);
-    feedbackForm.init();
 
     const extensionsCatalog = new ExtensionsCatalog(certificates, proxy, configurationRegistry, apiSender);
     extensionsCatalog.init();
@@ -1851,7 +1848,7 @@ export class PluginSystem {
 
     this.ipcHandle(
       'showMessageBox:onSelect',
-      async (_listener, id: number, index: number | undefined, option: RemindOption | undefined): Promise<void> => {
+      async (_listener, id: number, index: number | undefined, option: number | undefined): Promise<void> => {
         return messageBox.onDidSelectButton(id, index, option);
       },
     );
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -161,6 +161,7 @@ import { ExtensionDevelopmentFolders } from './extension/extension-development-f
 import { ExtensionsUpdater } from './extension/updater/extensions-updater.js';
 import { Featured } from './featured/featured.js';
 import type { FeaturedExtension } from './featured/featured-api.js';
+import { FeedbackForm } from './feedback-form.js';
 import { FeedbackHandler } from './feedback-handler.js';
 import { FilesystemMonitoring } from './filesystem-monitoring.js';
 import { IconRegistry } from './icon-registry.js';
@@ -173,7 +174,7 @@ import { KubernetesClient } from './kubernetes/kubernetes-client.js';
 import { downloadGuideList } from './learning-center/learning-center.js';
 import { LearningCenterInit } from './learning-center-init.js';
 import { LibpodApiInit } from './libpod-api-enable/libpod-api-init.js';
-import type { MessageBoxOptions, MessageBoxReturnValue } from './message-box.js';
+import type { MessageBoxOptions, MessageBoxReturnValue, RemindOption } from './message-box.js';
 import { MessageBox } from './message-box.js';
 import { NavigationItemsInit } from './navigation-items-init.js';
 import { OnboardingRegistry } from './onboarding-registry.js';
@@ -724,6 +725,8 @@ export class PluginSystem {
     await this.extensionLoader.init();
 
     const feedback = new FeedbackHandler(this.extensionLoader);
+    const feedbackForm = new FeedbackForm(commandRegistry, apiSender);
+    feedbackForm.init();
 
     const extensionsCatalog = new ExtensionsCatalog(certificates, proxy, configurationRegistry, apiSender);
     extensionsCatalog.init();
@@ -1848,8 +1851,8 @@ export class PluginSystem {
 
     this.ipcHandle(
       'showMessageBox:onSelect',
-      async (_listener, id: number, index: number | undefined): Promise<void> => {
-        return messageBox.onDidSelectButton(id, index);
+      async (_listener, id: number, index: number | undefined, option: RemindOption | undefined): Promise<void> => {
+        return messageBox.onDidSelectButton(id, index, option);
       },
     );
 
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -1848,8 +1848,8 @@ export class PluginSystem {
 
     this.ipcHandle(
       'showMessageBox:onSelect',
-      async (_listener, id: number, index: number | undefined, option?: number): Promise<void> => {
-        return messageBox.onDidSelectButton(id, index, option);
+      async (_listener, id: number, index: number | undefined, dropdownIndex?: number): Promise<void> => {
+        return messageBox.onDidSelectButton(id, index, dropdownIndex);
       },
     );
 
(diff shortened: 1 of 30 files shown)


deboer-tim:
The issue this fixes is already closed, so might need a new reference / context.

I'm not totally convinced that remind options should be a core feature for all message boxes, or maybe it should be more generic so you could provide your own options? 🤷🏻‍♂️ Either way, would be nice if the options were optional, in part so that all the tests don't need to be updated.

gastoner:
> The issue this fixes is already closed, so might need a new reference / context.
> 
> I'm not totally convinced that remind options should be a core feature for all message boxes, or maybe it should be more generic so you could provide your own options? 🤷🏻‍♂️ Either way, would be nice if the options were optional, in part so that all the tests don't need to be updated.

For the issue, what is this closing (should be probably `part of`) was opened like 5 PRs  (I've merged one of them and the issue has closed :D). I'll try to make it optional not to be forced to update all the tests

packages/main/src/plugin/message-box.ts
```diff
@@ -15,11 +15,27 @@
...
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
+import type { IconDefinition } from '/@api/icon-info.js';
+
 import type { ApiSenderType } from './api.js';
 import { Deferred } from './util/deferred.js';
 
 type DialogType = 'none' | 'info' | 'error' | 'question' | 'warning';
 
+export type RemindOption = 'Remind me tomorrow' | 'Remind me in 2 days' | `Don't show again`;
```
deboer-tim: I don't see this used, and IMHO shouldn't be part of the message box API - we may want some consistency between remind options in experimental features, but (e.g.) updates could have different times.

(if it did stay, 'Don't show again' just needs regular quotes like the other two)

packages/main/src/plugin/message-box.ts
```diff
@@ -15,11 +15,27 @@
...
+
+export interface DropdownType {
+  heading: string;
+  buttons: string[];
+}
+
+export interface IconButtonType {
+  label: string;
+  icon: IconDefinition;
```
deboer-tim: I couldn't find this being used by the message box below? Not sure if this is unrelated or incomplete feature.

gastoner: @deboer-tim both of those is used in https://github.com/podman-desktop/podman-desktop/pull/11992/files 
But it is true that it should be probably in [packages/renderer/src/stores/feedbackForm.ts](https://github.com/podman-desktop/podman-desktop/pull/11992/files#diff-faf7e5e0623d3f6679a0c52fbc4276d0e6a37a30d5933e54d5267d7580bbfc33) (another PR)

packages/main/src/plugin/index.ts
```diff
@@ -1848,8 +1848,8 @@ export class PluginSystem {
 
     this.ipcHandle(
       'showMessageBox:onSelect',
-      async (_listener, id: number, index: number | undefined): Promise<void> => {
-        return messageBox.onDidSelectButton(id, index);
+      async (_listener, id: number, index: number | undefined, option: number | undefined): Promise<void> => {
```
```suggestion
      async (_listener, id: number, index: number | undefined, option?: number): Promise<void> => {
```

packages/main/src/plugin/message-box.ts
benoitf: agreed, it can't be part of the messageBox

benoitf (commented):
the RemindOption type should be defined where it's used (for sure it's not in message-box)

the other parts are almost LGTM

packages/main/src/plugin/message-box.ts
```diff
@@ -93,7 +115,22 @@ export class MessageBox {
     return deferred.promise;
   }
 
-  async showDialog(type: DialogType, title: string, message: string, items: string[]): Promise<string | undefined> {
+  isDropdownType(response: ButtonsType | undefined): response is DropdownType {
```
benoitf:
```suggestion
  isDropdownType(response?: ButtonsType): response is DropdownType {
```

packages/main/src/plugin/message-box.ts
```diff
@@ -102,14 +139,22 @@ export class MessageBox {
...
 
     return undefined;
   }
 
   // this method is called by the frontend when the user selected a button
-  async onDidSelectButton(id: number, selectedIndex: number | undefined): Promise<void> {
+  async onDidSelectButton(
+    id: number,
+    selectedIndex: number | undefined,
+    selectedOption: number | undefined,
```
benoitf:
```suggestion
    selectedIndex?: number,
    selectedOption?: number,
```

packages/preload/src/index.ts
```diff
@@ -1771,8 +1771,12 @@ export function initExposure(): void {
   );
   contextBridge.exposeInMainWorld(
     'sendShowMessageBoxOnSelect',
-    async (messageBoxId: number, selectedIndex: number | undefined): Promise<void> => {
-      return ipcInvoke('showMessageBox:onSelect', messageBoxId, selectedIndex);
+    async (
+      messageBoxId: number,
+      selectedIndex: number | undefined,
+      selectedOption: number | undefined,
```
benoitf:
```suggestion
      selectedIndex?: number,
      selectedOption?: number,
```

packages/renderer/src/lib/dialogs/MessageBox.svelte
```diff
@@ -86,14 +86,14 @@ function cleanup(): void {
...
+async function clickButton(index?: number, option?: number): Promise<void> {
   cleanup();
-  await window.sendShowMessageBoxOnSelect(currentId, index);
+  await window.sendShowMessageBoxOnSelect(currentId, index, option);
 }
 
 async function onClose(): Promise<void> {
   cleanup();
-  await window.sendShowMessageBoxOnSelect(currentId, cancelId >= 0 ? cancelId : undefined);
+  await window.sendShowMessageBoxOnSelect(currentId, cancelId >= 0 ? cancelId : undefined, undefined);
```
benoitf: optional field should just be omitted (so no need to give undefined)

gastoner:
@benoitf PR updated

packages/preload/src/index.ts
```diff
@@ -1771,8 +1771,8 @@ export function initExposure(): void {
   );
   contextBridge.exposeInMainWorld(
     'sendShowMessageBoxOnSelect',
-    async (messageBoxId: number, selectedIndex: number | undefined): Promise<void> => {
-      return ipcInvoke('showMessageBox:onSelect', messageBoxId, selectedIndex);
+    async (messageBoxId: number, selectedIndex?: number, selectedOption?: number): Promise<void> => {
```
axel7083: question (non-blocking): sorry I am late to the party, I am not to understand the difference between `selectedIndex` and `selectedOption`? Could you explain me quickly (and/or) add a comment ?

gastoner: Sure, the selectedIndex is currently used to detect which option was selected e.g. if I have Yes and No buttons in MesasgeBox and I select Yes I'll get e.g. 0 

But when you will add a Dropdown button with some options (as we are doing here), you will have buttons Yes, [Maybe, Foo, Bar], no. So you need to differentiate between Maybe, Foo, Bar => `selectedOption`. Whens selecting Foo, you will get selectedIndex = 1 and selectedOption = 1

packages/main/src/plugin/message-box.ts
```diff
@@ -52,10 +66,15 @@ export interface MessageBoxOptions {
    * The (optional) index of the button to be used to cancel the dialog.
    */
   cancelId?: number;
+  /**
+   * An additional (optional) markdown description
+   */
+  footerMarkdownDescription?: string;
```
deboer-tim: In the draft PR showing where this is going, the message property allows markdown (👍🏼) but we also have the detail property (which could be markdown too). Why do we need three different content properties and how would this be any different from the other two?

packages/main/src/plugin/message-box.ts
```diff
@@ -15,11 +15,25 @@
...
 import { Deferred } from './util/deferred.js';
 
 type DialogType = 'none' | 'info' | 'error' | 'question' | 'warning';
 
+export interface DropdownType {
+  heading: string;
+  buttons: string[];
+}
+
+export interface IconButtonType {
```
deboer-tim: Missing tests for IconButtonType. Also not clear to me how / what this would be used for as I couldn't relate it to anything in the design.

deboer-tim (commented):
Sorry for the slow review, a few comments added.

Breaking the issue into smaller PRs (and having a draft with the final) is great, but I had a really hard time mapping what is here to the requirements of the design w/o the MessageBox implementation here, especially since the design introduces a new button type that's inconsistent with the buttons on the Dashboard 😞 and how I was expecting it to work. It would be nice to describe a bit more here about what's in / what's not, or how this would be used.

I couldn't find any similar dropdowns that also closed the dialog / submitted the form they are on. I don't want to 'reopen the design', but if that's the intention I wonder if we should find a different way to do it. 🤷🏼‍♂️

packages/main/src/plugin/message-box.ts
gastoner: Not sure onestly, I can't find where it is used either :D Probbaly some leftover that I forgot to remove, I'll do it now

gastoner:
> Sorry for the slow review, a few comments added.
> 
> Breaking the issue into smaller PRs (and having a draft with the final) is great, but I had a really hard time mapping what is here to the requirements of the design w/o the MessageBox implementation here, especially since the design introduces a new button type that's inconsistent with the buttons on the Dashboard 😞 and how I was expecting it to work. It would be nice to describe a bit more here about what's in / what's not, or how this would be used.
> 
> I couldn't find any similar dropdowns that also closed the dialog / submitted the form they are on. I don't want to 'reopen the design', but if that's the intention I wonder if we should find a different way to do it. 🤷🏼‍♂️

Personally, I would say that we are just adding support for dropdown type into the message box so we can have more opinions (when there are 3+ options, the message box looks quite full, in my opinion—I think update message box it is). But I would also like to get the issue merged and iterate later since it has been open for a while  (this PR has been open for a month and counting).

packages/main/src/plugin/message-box.ts
gastoner: I can probably make it work with `detail` (after conversion to Markdown) or `message` field https://github.com/podman-desktop/podman-desktop/pull/11992/files#diff-dea85c4842773a489e054a56c9410e142094665feb91933767e5b3a4644d5066L127-R147 should I do this change then?

benoitf (commented):
overall LGTM

IMHO:
I would rename `selectedOption` to `dropdownIndex` to ease the logic
And I would move the introduction of footerMarkdownDescription to a separate PR as it looks unrelated

gastoner:
> overall LGTM
> 
> IMHO: I would rename `selectedOption` to `dropdownIndex` to ease the logic And I would move the introduction of footerMarkdownDescription to a separate PR as it looks unrelated

What about ditching the `footerMarkdownDescription` in favor of `detail`?

benoitf (approved):

deboer-tim (approved):

PR #11992, opened 2025-04-02
chore: added experimental feedback logic

### LINKED ISSUE #10228: Experimental features feedback notification

gastoner:
### What does this PR do?
Adds experimental feedback logic for showing dialog
In `settings.json` file: 
can look like this:
```
    "kubernetes.statesExperimental": {
    remindAt: 1748321012505,
    disabled: false
  }, "statusbarProviders.showProviders": {
    remindAt: 1748321012505,
    disabled: true
  }
```
Where: 
1. I have disabled features tasks.Manager / tasks.Statusbar => should be not in settings
2. statusbarProviders.showProviders is disabled => means that user clicked in `dont show again` and does not want to be notified with this feature (into disabled are added all features that are at given time enabled -> here was kubernetes.statesExperimental enabled after 'don't show again' was clicked)
3. 1748321012505 - some timestamp when user should be notified - during the first startup/enabling the feature timestamp is set to 2 days from now . When this timestamp is lower than the current timestamp, the dialog is shown during sthe tart of an app (only ONE dialog should be visible) 

### Screenshot / video of UI

image

Closes https://github.com/podman-desktop/podman-desktop/issues/10228

### How to test this PR?
1. Enable experimental feature -> in settings.json you should see new item with timestamps(should be your date + 2 days)
2. edit the timestamp so it will show during next startup e.g. going from 17..... to 16....
3. You should see feedback form with dropdown 

Click on the buttons in the dialog, those should:
a. Share Feedback on GitHub -> should open website + the feature should disappear from settings.json
b. Remind me tomorrow -> should update its timestamp with your date + 1 day 
c. Remind me in 2 days-> should update its timestamp with your date + 2 day 
d. Don't show again will add ALL the features (currently enabled) to `disabled` -> this way we can diferenciate between disabled feature and don't show again option.

Try to go through the dialog options & disable/enable features => Use 1,2,3 steps to get to "default state" to see the dialog


axel7083 (commented):
Following today discussion, here are some of my thought 

1. We should not try to detect when a user is using the experimental feature

This would take too much custom code for each experimental feature, when creating experimental feature, we probably just want to set a flag, not having to add hook all over the code to catch when the user it using it.

Same when moving from experimental to stable, I don't want to cleanup all the code doing the catch logic.

2. Using persistent timestamp

We cannot rely on the stores in the frontend, there are not persisted over reboot, therefore anything in them would be discarded.

We should store a timestamp a user is enabling an experimental feature in the configuration registry, so it is persisted.

3. Do not spam the user

We should decide when we should send a notification feedback to the user, after how long. 

Some following questions, what happen if we have several experimental feature enabled since a long time and we want to get feedback? should we ask for each of them? ask only for one (preferred) ? pick random?

Should we ask at startup ? Is there a risk that this interfere with the Update dialog? I don't want to have the update dialog popup and instantly have the feedback popup

package.json
```diff
@@ -180,6 +180,7 @@
   },
   "dependencies": {
     "@docker/extension-api-client-types": "0.3.4",
+    "@fortawesome/free-solid-svg-icons": "^6.7.2",
```
axel7083: surprising?

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,192 @@
...
+          }
+          // Option from Dropdown was selected
+          else if (response.response === 0 && response.dropdownIndex) {
+            // Default is "Dont show again"
+            let remindInDays = MAX_NUMBER;
+            const selectedOption = options[response.dropdownIndex];
+            if (selectedOption === 'Remind me tomorrow') remindInDays = 1;
+            else if (selectedOption === 'Remind me in 2 days') remindInDays = 2;
+            this.setTimestamp(key, remindInDays);
+          }
```
axel7083: If the user is saying don't show again, we should probably save somewhere that he does not want to give feedback for ALL experimental feature, not just one. 

I don't know if we should have a dedicated configuration, but having special value is not really nice, I would rather see in the `settings.json` something like `feedback.experimental: 'disabled'` than really big numbers without meaning

---

You should use a switch for the options, including all options, for readability, and safeness

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,192 @@
...
+          properties: {
+            ['timestamp']: {
+              description: 'Notification fimestamp for each experimental feature',
+              type: 'object',
+              hidden: true,
+            },
+          },
+        },
+      ]),
+      this.#configurationRegistry.onDidChangeConfiguration(event => {
```
axel7083: this produce a disposable?

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,192 @@
...
+      const configurationValue = this.#configurationRegistry
+        .getConfiguration(splittedName[0])
+        .get<boolean>(splittedName[1]);
+      const feature = `${splittedName[0]}.${splittedName[1]}`;
+      if (configurationValue) this.setTimestamp(feature, 2);
+      else this.setTimestamp(feature, undefined);
+    }
+  }
+
+  formatName(id: string): string {
```
axel7083:
```suggestion
  protected formatName(id: string): string {
```

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,192 @@
...
+    if (splittedName.length >= 2 && splittedName[1]) {
+      const configurationValue = this.#configurationRegistry
+        .getConfiguration(splittedName[0])
+        .get<boolean>(splittedName[1]);
+      const feature = `${splittedName[0]}.${splittedName[1]}`;
+      if (configurationValue) this.setTimestamp(feature, 2);
+      else this.setTimestamp(feature, undefined);
+    }
+  }
+
```
axel7083: please add a comment mention what this function is doing with example

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,192 @@
...
+  formatName(id: string): string {
+    return id
+      .split('.')
+      .map(part => part.replace(/([a-z])([A-Z])/g, '$1 $2'))
+      .join(' ');
+  }
+
+  async showFeedbackDialog(): Promise<void> {
```
axel7083:
```suggestion
  protected async showFeedbackDialog(): Promise<void> {
```

package.json
gastoner: I need to add some icon to message box, and it is not available in plugins :(

packages/main/src/plugin/experimental-feature-feedback-form.ts
gastoner: This really reflect to `null` in settings (just updated the desc^^)

package.json
axel7083: separate PR please 🙏

packages/main/src/plugin/experimental-feature-feedback-form.ts
gastoner: `ConfigurationRegistry.onDidChangeConfiguration: (listener: (e: IConfigurationChangeEvent) => unknown, thisArgs?: unknown, disposables?: DisposableGroup) => IDisposable` Yess

axel7083: then should be added to `#disposables` 👍

package.json
gastoner: https://github.com/podman-desktop/podman-desktop/pull/12641 created

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,216 @@
...
+          message: `We are testing something new!\n\nHow's your experience so far with [${featureName}](${featureGitHubLink})? Let us know on GitHub!`,
+          type: `info`,
+          buttons: [
+            {
+              heading: 'Remind me later',
+              buttons: options,
+            },
+            {
+              label: 'Share Feedback on GitHub',
+              icon: faArrowUpRightFromSquare,
```
axel7083: Whaaaatt, how does this works, how is this supported? Feel odd to me that we use svg library in the backend, sorry for https://github.com/podman-desktop/podman-desktop/pull/12641 but this feel super hacky?

I don't really want to import in the backend a ui library...

gastoner: Well here you specify buttons it can be just "yes", "no", Dropdown with multiple options, or button with some extra icon like here: https://github.com/podman-desktop/podman-desktop/issues/10224#issuecomment-2636084884

axel7083: This does seem a bit complicated cc @benoitf for guideline on this matter

benoitf (commented):
this PR is being stuck on review but I think it's b/c it's doing too many things at once

it brings changes in experimental features and change messageBox impmentation

I would expect one PR that is changing messageBox + dialog type

and another one that will use the component (because anyway if the previous PR is not accepted this one won't be validated)

gastoner:
@benoitf created https://github.com/podman-desktop/podman-desktop/pull/13049

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,222 @@
...
+import { shell } from 'electron';
+
+import type { IDisposable } from '/@api/disposable.js';
+
+import type { ConfigurationRegistry } from './configuration-registry.js';
+import type { MessageBox } from './message-box.js';
+
+// Pseudo infinity number for "don't show again" option
+// Aprox Wednesday, June 6, 2255
+const MAX_NUMBER: number = Number.MAX_SAFE_INTEGER;
```
SoniaSandler: I'm not sure I understand why we need to use a number for the "don't show again" option

SoniaSandler (commented):
> If the user is saying don't show again, we should probably save somewhere that he does not want to give feedback for ALL experimental feature, not just one.

I agree with @axel7083's comment about adding a way to do that "don't show again" will apply to everything. If it's possible, maybe you could do separate options to disable the feedback for a specific feature and for all of them

packages/main/src/plugin/experimental-feature-feedback-form.ts
gastoner: Updated, now we are using the 'disabled' instead of "infinity number"

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * Updates timestamp for given experimental feature
+   * @param feature in format feature.name
+   * @param days timeout in days, undefined -> disabled, 'disabled' -> "Don't show again" was selected
+   */
+  protected setTimestamp(feature: string, days: Timestamp): void {
+    let date: Timestamp = days;
+    if (typeof days === 'number') {
+      date = new Date(new Date().getTime() + days * 24 * 60 * 60 * 1000).getTime();
```
benoitf: suggest: please declare a constant for * 24 * 60 * 60 * 1000

also write numbers using `1_000` format for numbers greater than 1000

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+    // update configuration
+    this.#timestamps.set(feature, date);
+    this.save().catch(console.error);
```
benoitf: it should be prefixed with the problem we have in that case for example let say the error we got is 'Error stream closed' it'll display 'stream closed' but we may have no idea it's coming from this place ?

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+      const configurationValue = this.#configurationRegistry
+        .getConfiguration(splittedName[0])
+        .get<boolean>(splittedName[1]);
+      if (configurationValue) this.setTimestamp(configurationName, 2);
+      else this.setTimestamp(configurationName, undefined);
+    }
+  }
+
+  /**
+   * Formats id to better readable one
```
benoitf: instead of saying 'better readable one' it should say what it does like 'replace dots by spaces' and then adds uppercase on each sequence

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+  protected formatName(id: string): string {
+    return id
+      .split('.')
+      .map(part => part.replace(/([a-z])([A-Z])/g, '$1 $2'))
+      .join(' ');
+  }
+
+  /**
+   * Getter method for timestamps
+   * @returns timestamps
```
benoitf: IMHO it's not very helpful as comments, so either drop it or add real value

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+   */
+  protected getTimestampsMap(): Map<string, Timestamp> {
+    return this.#timestamps;
+  }
+
+  /**
+   * Goes through each enabled experimenatl feature and shows dialog if current timestamp is greater than stored value
```
benoitf:
```suggestion
   * Goes through each enabled experimental feature and shows dialog if current timestamp is greater than stored value
```

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+            this.setTimestamp(key, undefined);
+          }
+          // Option from Dropdown was selected
+          else if (response.response === 0 && typeof response.dropdownIndex === 'number') {
+            switch (options[response.dropdownIndex]) {
+              case 'Remind me tomorrow':
+                this.setTimestamp(key, 1);
+                break;
+              case 'Remind me in 2 days':
+                this.setTimestamp(key, 2);
```
benoitf: if not covered it should be covered, each option should be tested, maybe 2 days is not working ?

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+import type { IDisposable } from '/@api/disposable.js';
+
+import type { ConfigurationRegistry } from './configuration-registry.js';
+import type { MessageBox } from './message-box.js';
+
+export type Timestamp = number | undefined | 'disabled';
+
+type RemindOption = 'Remind me tomorrow' | 'Remind me in 2 days' | `Don't show again`;
+
+export class ExperimentalFeatureFeedbackForm {
```
benoitf: issue: it should now contains the @injectable() decorator

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+  #disposables: IDisposable[] = [];
+  #configuration: containerDesktopAPI.Configuration | undefined;
+  #timestamps: Map<string, Timestamp> = new Map();
+  #experimentalFeatures: Set<string> = new Set([]);
+  readonly #configurationRegistry: ConfigurationRegistry;
```
benoitf: issue: it should be the interface, IConfigurationRegistry

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+  constructor(
+    private configurationRegistry: ConfigurationRegistry,
+    private messageBox: MessageBox,
```
benoitf: issue: missing `@inject` decorators

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+  ) {
+    this.#configurationRegistry = this.configurationRegistry;
+  }
+
+  protected async save(): Promise<void> {
+    if (!this.#configuration) throw new Error('missing configuration object: cannot save');
```
benoitf: issue: it's not really clear for which part it's an issue.

IMHO add a prefix like `Error while trying to save the experimental mode settings` (or any other meaningful message)

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,225 @@
...
+        },
+      ]),
+      this.#configurationRegistry.onDidChangeConfiguration(event => {
+        // If configuration changed and if is experimenal feature
+        if (this.#experimentalFeatures.has(event.key) && typeof event.value === 'boolean') {
+          // If we detect change, check if the timestamp for a feature does not have disabled value
+          // -> was previously selected "Don't show again"
+          if (this.#timestamps.has(event.key) && this.#timestamps.get(event.key) === 'disabled') {
+            return;
+          }
```
benoitf: it looks like it's not tested

benoitf (changes requested):

gastoner:
@benoitf Fixed

packages/main/src/plugin/index.ts
```diff
@@ -738,6 +739,10 @@ export class PluginSystem {
 
     await this.setupSecurityRestrictionsOnLinks(messageBox);
 
+    container.bind<ExperimentalFeatureFeedbackForm>(ExperimentalFeatureFeedbackForm).toSelf().inSingletonScope();
+    const experimentalFeatureFeedbackForm = new ExperimentalFeatureFeedbackForm(configurationRegistry, messageBox);
```
benoitf: it should not use new keyword but perform something like:
(not tested)
```suggestion
    const experimentalFeatureFeedbackForm = container.get<ExperimentalFeatureFeedbackForm>(ExperimentalFeatureFeedbackForm);
```

benoitf:
I'm not feeling confident to merge it the same day of the release, as there won't any much usage of this.

Due to timestamps/ latetimers I have the feeling we may have forgotten some corner cases. And if it's become annoying for the user we would have multiple reports.

I think this is the kind of stuff that needs to be merged at the beginning of the release cycle, not close to the release date.

benoitf (commented):
I've noticed it's storing entries in my config file as `features.timestamp` but IMHO it is not self explanatory

like, what is this timestamp for ? last enablement time, last disabled time, ?

why is it called feature while it's for experimental stuff ?

and I'm finding more confusing that timestamp could be 'disabled' 

I think the properties being stored should be reviewed as it won't scale

If we want to introduce new fields it won't be possible

also I would expect the timestamp would stay in all case (like when was the time where it has been disabled for example) ?

and why a new top level entry features.timestamp

gastoner:
> I've noticed it's storing entries in my config file as `features.timestamp` but IMHO it is not self explanatory
> 
> like, what is this timestamp for ? last enablement time, last disabled time, ?
> 
> why is it called feature while it's for experimental stuff ?
> 
> and I'm finding more confusing that timestamp could be 'disabled'
> 
> I think the properties being stored should be reviewed as it won't scale
> 
> If we want to introduce new fields it won't be possible
> 
> also I would expect the timestamp would stay in all case (like when was the time where it has been disabled for example) ?
> 
> and why a new top level entry features.timestamp

We can merge this after release but I think this should be decided by @rujutashinde @Firewall @slemeur 

Originaly I was thinking about having this as a 3 part like features.experimental, but I'm now not sure why I did not add the experimental in here. I can rename it to e.g. experimentalFeatures.notifyAt 

And in this case I would change 'disabled' -> 'never'
I'm not sure what would be best approach for storing the timestamps/never we need to catch 3 cases:
1. store timestamp when we should show dialog
2. how to store "don't show again" option
3. how to store "disabled" experimental feature - I'm not sure if we want to keep the timestamp in here, we could and then check the feature if it is enabled. This would mean that if we select e.g. "tomorrow" and then disable the feature we wont get the dialog until we enable the feature again

benoitf:
> And in this case I would change 'disabled' -> 'never'
>I'm not sure what would be best approach for storing the timestamps/never we need to catch 3 cases:

> store timestamp when we should show dialog
> how to store "don't show again" option
> how to store "disabled" experimental feature - I'm not sure if we want to keep the timestamp in here, we could and then 
> check the feature if it is enabled. This would mean that if we select e.g. "tomorrow" and then disable the feature we wont
> get the dialog until we enable the feature again

to me there is something unclear there

if the experimental feature is disabled, why should I store a timestamp ? there is already a configuration setting for enabling/disabling the experimental feature. So it makes me feel that we're storing two things at different scopes ?

if we need to keep something maybe

```json
{
experimentalFeatures:
    {
          disabled: ['feat1', 'feat2']
          remindAt: {
                 "feat1": 123456
          }
}
```

or something more scalable:

``json
{
  "experimentalFeatures": {
      "feat1": {
        "enabled": false,
        "remindAt": 1234567890
      },
      "feat2": {
        "enabled": false
      },
      "feat3": {
        "enabled": true
      }
  }
}
```

it would also mean to merge/store experimental features under the same top level entry (because for now there is a top level entry for each experimental flag)

gastoner:
@benoitf I've updated the persistent storage mechanism => renaming, added disabled array
Not sure if we want to keep the remindAt when is the feature disabled from settings OR disabled by selecting "don't remind"?

gastoner:
I got an idea for improvement: 

We could add gifs for each feature 

[Screencast_20250811_141355.webm](https://github.com/user-attachments/assets/5a25ce9b-a982-432d-9133-70ec2321096d)

cc @Firewall @vancura

Firewall:
Excellent idea @gastoner! Normally I would be hesitant to add gifs to the product but since the nature of these is that they are temporary together with the experimental features, I think this is a great reminder of what the feature is that we are asking feedback on.

vancura:
This is so awesome! Great improvement! I am unsure about GIFs, though; they are not very well compressed. Considering we're in modern Electron, we can use a better format like WebM. This could also be automated via `ffmpeg`.

jiridostal (commented):
I like the idea of adding a gif/or any other animation for the features, user usually don't remember names of features :-)

jiridostal:
@gastoner I see the code change for showing images, but no images are added yet - is that on purpose?

vancura:
FYI, we worked together on the video file format, and we'll use WebM here. It's _much_ smaller than a GIF of the same resolution and framerate (and has more than 256 colors).

gastoner:
@jiridostal yes this is expected, there should be visible either the thumbs up/down as a fallback or videos after 
1. https://github.com/podman-desktop/podman-desktop/pull/13562
2. https://github.com/podman-desktop/podman-desktop/pull/13563
3. https://github.com/podman-desktop/podman-desktop/pull/13567
4. https://github.com/podman-desktop/podman-desktop/pull/13565 
are merged (each PR is one feature video + relevant changes)

jiridostal (approved):
Awesome!

gastoner:
@jiridostal @slemeur @Firewall I've added telemetry for this feature, so we can track on which option users clicks

benoitf (approved):

PR #12026, opened 2025-04-03
chore(vitest): setup workspace

### LINKED ISSUE #11985: Codecov spamming warning on `*.svelte` files
### Bug description

We are uploading our code coverage reports to the codecov service, and this is nice for pure typescript file as it allows us to catch branch or places where tests do not cover the code.

On svelte file, it is a bit crazy crazy, and add tons of annotations when opening a pull request, making it _a bit annoying_ to review. 

When reviewing we cannot hide all annotations at once, we need to hide them by file, making

https://github.com/user-attachments/assets/e4aaf80b-7b59-4e10-b5f7-01088ea256ff

| With annotations | Without |
| --- | --- |
| Image | Image |

A pretty simple change is not easy to review with the annotations, therefore (I personally) hide them after looking at what they are saying. 

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

The screenshots are from https://github.com/podman-desktop/podman-desktop/pull/11981/files

### Relevant log output

```shell

```

### Potential solutions

- We could ignore the `.svelte` files as a quick solution[^1]
- We could try to improve the coverage, and investigate why the problems occurs: but this would probably take time
- We could try to contact codecov teams, to have feedback on it 

[^1]: https://docs.codecov.com/docs/ignoring-paths

axel7083:
### What does this PR do?

POCing around the vitest workspace[^1] to try to address the coverage issue on svelte files.

### Notes

- Adding a building step to the workflow, to avoid having a build step hidden inside the `pnpm test:unit`
- Adding new entry in scripts in `package.json`, called `test:unit:coverage`.
- Updating the entry `test:unit` to use vitest workspace, instead of doing individual run.
- Updating invididual `test:*` to use `vitest --project` option
- Updating&Adding `vitest.config.js` to use `defineProject`[^1]
- Removing unused `vitest.config.js` at root of the repository, seems to be configured for e2e? but no usage ? was leading to issue with vitest, had to be removed
- Creating `vitest.config.unit.ts` to have a dedicated config for unit test. (We could imagine having the same for e2e) but today they seems to be launched through `npx playwright` 
- Removing `vitest-shared-extensions.config.js` as shared config should be put inside the `vitest.config.unit.ts` (and it is)

[^1]: https://vitest.dev/guide/workspace

### Screenshot / video of UI

Explore [Codecov PR report](https://app.codecov.io/github/podman-desktop/podman-desktop/pull/12026)

Hopefully fixes https://github.com/podman-desktop/podman-desktop/issues/11985#issuecomment-2773329781
Heavily inspired from https://github.com/podman-desktop/extension-podman-quadlet/pull/407

Discussion with Vitest contributors on the subjet (performance, and workspace management) https://discord.com/channels/917386801235247114/926890821364621342/1357452269775224993

### How to test this PR?

- [x] Pipeline should be green
- [x] Codecov should show some nice files


benoitf (commented):
first, please try to migrate to typescript as a separate PR (or don't move to typescript at all)
else they'll be too many changes

axel7083:
> first, please try to migrate to typescript as a separate PR (or don't move to typescript at all) else they'll be too many changes

Yeah I was not planning to do a single PR don't worry ahah, this is like experimentation :)

axel7083:
First results are promising, with the vitest workspace, we gain a 20% coverage, going deeper into the analysis show  that it properly recognize the svelte coverage

image

Let's take the `Label.svelte` example

| This PR | main |
| --- | --- |
| image | image |

Having the vitest workspace properly configured remove the weird uncovered line, which should be ignored.

axel7083:
I had to increase the memory usage of node, because I was getting `JavaScript heap out of memory`. We should open an issue on their side to notify this issue.

axel7083:
After some investigation (thanks to the vitest team on discord 👍 ) I could narrow down the problem and found out the crash was happening when the vitest process the `uncovered files` 

vitest-uncovered-files

axel7083:
image

axel7083:
kinda big PR, we could organise a dedicated call to explore why some decision has to be made ~

benoitf:
I like the workspace but they're too many unrelated changes that are not directly part of using a workspace and a lot is opinionated changes so it needs to be specific change.

I think you should proceed with small chunks
- remove unused file 
- rename file to ts extension
- etc.

unrelated changes: building the project before running tests, for example why test:unit:coverage (and not coverage by default is out of the scope of this PR)

changes like 
```
 '@podman-desktop/api': path.resolve('../../', '__mocks__/@podman-desktop/api.js'),
 '@podman-desktop/api': resolve(WORKSPACE_ROOT, '__mocks__/@podman-desktop/api.js'),
```

are also unrelated

so I would like to see very specific changes.

benoitf (commented):
as a side note, I don't see why we have so many files vite and vitest (AFAIK they can be merged) and I guess we don't need one for each extension in case of workspace

axel7083:
Supersede by https://github.com/podman-desktop/podman-desktop/pull/12111

PR #12033, opened 2025-04-04
chore(test): handle exception thrown from race condition

cbr7:
### What does this PR do?
Fixes some flakyness from faulty race condition handling.


odockal (approved):
LGTM

Issue #12034, opened 2025-04-04
Type narrowing is lost in svelte 5 snippet


feloy:
### Bug description

With svelte 4 / slots, type narrowing was preserved in `slot`s:

For example in `packages/renderer/src/lib/container/ContainerExport.svelte`:
```
{#if container}
  <EngineFormPage title="Export container {container.name}">
    <div slot="content" class="space-y-2">
      {container.id} <- this is ok, as container has been narrowed to non-undefined by englobing `if`
    </div>
  </EngineFormPage>
{/if}
```

With Svelte 5 / snippets:

```
{#if container}
  <EngineFormPage title="Export container {container.name}">
    {#snippet content()}
      {container.id} <- 'container' is possibly 'undefined'.ts(18048)
    {/snippet>
  </EngeinFormPage>
{/if}
```

### Operating system

macos

### Installation Method

None

### Version

next (development version)

### Relevant log output

```shell

```

### Additional context

Introduced by https://github.com/podman-desktop/podman-desktop/pull/11889

odockal:
@feloy Hey, I have assigned you since you have draft PR, feel free to update project information as you like, just fyi.

feloy:
closing, as the solution of passing the value as a parameter is not better: https://github.com/podman-desktop/podman-desktop/pull/12037

PR #12036, opened 2025-04-04
fix: remove unnecessary assignment

feloy:
### What does this PR do?

Remove unnecessary assignment.

I suppose that this assignment has been added originally to force svelte to react on the change of the value on the parent component (ConfigMapDetails) when the value is modified in the child (ConfigMapSecretsActions): the child modifies the status of the configmap to DELETING, so the icon is changed to a spinner during the deletion. (@cdrage @deboer-tim I think you worked on the early versions of this page, do you confirm?)

With the current version of svelte, this assignment is not necessary to obtain the reaction. I've added a timeout to check that, without the assignment, the icon status is correctly changed to a spinner.

The second commit (setting timeout) will be removed if we agree on the fix, and I'll make the changes to all the other Kubernetes pages.

This change may seem completely off topic, but is helpful for fixing https://github.com/podman-desktop/podman-desktop/issues/12034 (with #12037) , related to #11579 

I have seen that is assignment has not been added for `CronJob` and `Job`, which makes me think they are not necessary anymore.

I added a comment for the podman pods, where this assignement is still necessary, because the `PodDetails` component is still in svelte 4. I tested to port it to svelte 5, and this removes the need of the assignment (I didn't include this change in this PR)

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretActions.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretActions.svelte
@@ -15,11 +15,13 @@ const configmapSecretUtils = new ConfigMapSecretUtils();
 async function deleteConfigMapSecret(): Promise<void> {
   configMapSecret.status = 'DELETING';
 
-  if (configmapSecretUtils.isSecret(configMapSecret)) {
-    await window.kubernetesDeleteSecret(configMapSecret.name);
-  } else {
-    await window.kubernetesDeleteConfigMap((configMapSecret as ConfigMapSecretUI).name);
-  }
+  setTimeout(() => {
+    if (configmapSecretUtils.isSecret(configMapSecret)) {
+      window.kubernetesDeleteSecret(configMapSecret.name);
+    } else {
+      window.kubernetesDeleteConfigMap((configMapSecret as ConfigMapSecretUI).name);
+    }
+  }, 5_000);
 }
 </script>
 
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretActions.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretActions.svelte
@@ -15,13 +15,11 @@ const configmapSecretUtils = new ConfigMapSecretUtils();
 async function deleteConfigMapSecret(): Promise<void> {
   configMapSecret.status = 'DELETING';
 
-  setTimeout(() => {
-    if (configmapSecretUtils.isSecret(configMapSecret)) {
-      window.kubernetesDeleteSecret(configMapSecret.name);
-    } else {
-      window.kubernetesDeleteConfigMap((configMapSecret as ConfigMapSecretUI).name);
-    }
-  }, 5_000);
+  if (configmapSecretUtils.isSecret(configMapSecret)) {
+    await window.kubernetesDeleteSecret(configMapSecret.name);
+  } else {
+    await window.kubernetesDeleteConfigMap((configMapSecret as ConfigMapSecretUI).name);
+  }
 }
 </script>
 
--- a/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretActions.svelte
+++ b/packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretActions.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 import { faTrash } from '@fortawesome/free-solid-svg-icons';
-import { createEventDispatcher } from 'svelte';
 
 import { withConfirmation } from '/@/lib/dialogs/messagebox-utils';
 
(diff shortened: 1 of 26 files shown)


packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretActions.svelte
```diff
@@ -11,22 +10,18 @@ import type { ConfigMapSecretUI } from './ConfigMapSecretUI';
...
 
-const dispatch = createEventDispatcher<{ update: ConfigMapSecretUI }>();
-export let onUpdate: (update: ConfigMapSecretUI) => void = update => {
-  dispatch('update', update);
-};
-
 const configmapSecretUtils = new ConfigMapSecretUtils();
 
 async function deleteConfigMapSecret(): Promise<void> {
   configMapSecret.status = 'DELETING';
```
axel7083, line 16: I am not sure to understand the full question here, but if what I understand is right, we want to know what to do when the child component want to update a props?

Here I feel like this is not what we should be doing, kubernetes resources lifecycle and status should be provided by the backend? 

I feel like setting `configMapSecret.status = 'DELETING';` is not the right way, the frontend should not update the resource. We should have a single soruce of truth, even when deleting

When we call the backend, the resource should get updated (from the cluster) with a deletionTimestamp (https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/).

So the frontend should no play with trying to change the state, it should be stateless and only display what the backend is providing.

If we have a deletion timestamp => we display a loading spinner, or make the status being derived from this

We already seems to be using this timestamp anyway in some places for pods and PVC

https://github.com/podman-desktop/podman-desktop/blob/2e0042840ccd55e1da2ac794c17ce78cef6304a0/packages/renderer/src/lib/kube/pods/pod-utils.ts#L54

https://github.com/podman-desktop/podman-desktop/blob/2e0042840ccd55e1da2ac794c17ce78cef6304a0/packages/renderer/src/lib/pvc/pvc-utils.ts#L34-L35

Sorry if this is does not respond to your question, but I think here, the ConfigMapSecretActions should never change the resource itself

I also don't understand the setTimeout

axel7083 (commented):
left a comment 👍

packages/renderer/src/lib/configmaps-secrets/ConfigMapSecretActions.svelte
feloy: Yes, I agree with your point that this status should come from the backend. But this is not this line which is blocking me from working on #12034.

My concern is the `configMap = configMap` in the `ConfigMapDetails` component, which I would like to remove. If later we also fix the problem you raise, this assignment will of course be even less necessary, so it doesn't hurt to remove it first.

If I understand correctly, this assignment was necessary to force svelte reactivity, but it is not anymore with current version of svelte. The `setTimeout` has been temporarily added to be able to see that the status icon is changed to a spinner in the meantime 
 as expected (because deleting a configmap is quite immediate). Not really necessary for configmap I agree, but I'll make the same changes in all other resources, which make take more time to be deleted. I'll remove the setTimeout when everyone is convinced it works as expected

axel7083: maybe you can ask the svelte team on their disucord, I went there to have clarification, I don't really have an opinion / solution for the problem you mention, but they would probably have some relevant opinion?

axel7083: > they like to have a reproducer on svelte playground

feloy: In fact, there is no problem on Svelte side. This PR is about removing this `configMap = configMap` which was necessary (I suppose) with old version of Svelte, but not necessary anymore with newer versions.

By removing it, all is working correctly if I don't miss anything. The reason I raised this discussion is to be sure I do not miss a reason why this assignment was present. 

For me, it is only for changing the icon to a Spinner when the user deletes the configmap, and this is still working after the change.

axel7083: Okay, again, I missunderstood, I thought the problem was https://github.com/podman-desktop/podman-desktop/issues/12034

feloy: No, the problem raised in #12034 should be fixed by passing the container (or other data) as parameter to the snippet: see #12037 . And this is there, in #12037 that this assignment `configMap = configMap` is quite frustrating...

deboer-tim: I haven't read the entire discussion, because you guys have written a lot :) , but just dropping in to confirm that lines like `configMap = configMap` were absolutely required in Svelte 4 and as far as I can tell no longer required with the reactivity changes. And yes, it was primarily to make sure when there was a change from the backend like start/stop status or something being deleted, the UI would react.

I think I've already cleaned up one of these in another Table somewhere, and fine to remove others as we go as long as you double-check the updates. I'm assuming we'd do a full pass of cleanup as/after Table goes to Svelte 5.

benoitf (approved):

deboer-tim (approved):
Sweaty palms deleting something I know was critically important in Svelte 4 :), but tested again and working fine.

Issue #11573, opened 2025-03-07
Migrate UI component Button to Svelte 5


jeffmaury:
https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/button/Button.svelte#L1

axel7083:
This is not simple ~

jeffmaury:
> This is not simple ~

I thought you liked complex stuff 😄

PR #12043, opened 2025-04-04
chore: migrate close button to svelte 5

deboer-tim:
### What does this PR do?

Migrates the CloseButton component to Svelte 5, primarily just migrating the on:click.

TaskManager was using a title property that didn't exist, I've removed it (vs adding a new property) because it doesn't seem necessary.

### Screenshot / video of UI

N/A

Part of #11573.

### How to test this PR?

Click some close buttons!

### DIFF
--- a/packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
+++ b/packages/renderer/src/lib/dashboard/ReleaseNotesBox.svelte
@@ -80,7 +80,7 @@ onDestroy(async () => {
           <p class="text-[var(--pd-content-card-header-text)] font-bold text-xl ml-2">
             {notesInfo?.title ?? ''}
           </p>
-          <CloseButton on:click={onClose} />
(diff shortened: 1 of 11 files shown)


benoitf (changes requested):
it's a breaking change

There should be a dispatcher event in case if onclick callback is not submitted to allow smooth migration

gastoner (commented):
I was doing something similar in https://github.com/podman-desktop/podman-desktop/pull/11830/files So there is not required change from `on:click` -> `onclick`

gastoner (commented):
Otherwise LGTM

axel7083 (approved):

deboer-tim:
Guess I forgot to hit Comment the other day - @benoitf Sorry, I didn't realize we were keeping compatibility as much as possible, new commit adds the same fallback as Link for on:click.

benoitf (approved):
thanks

PR #12055, opened 2025-04-07
chore: migration of checkbox to svelte 5

### LINKED ISSUE #11575: Migrate UI component Checkbox to Svelte 5
https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/checkbox/Checkbox.svelte#L1

bmahabirbu:
### What does this PR do?

Migrate checkbox to svelte 5, keeping some legacy code like the dispatcher 

fixes https://github.com/podman-desktop/podman-desktop/issues/11575

### How to test this PR?

the code tests should test this

### DIFF
--- a/packages/ui/src/lib/checkbox/Checkbox.svelte
+++ b/packages/ui/src/lib/checkbox/Checkbox.svelte
@@ -1,33 +1,53 @@
 <script lang="ts">
 import { faSquare as faOutlineSquare } from '@fortawesome/free-regular-svg-icons';
 import { faCheckSquare, faMinusSquare, faSquare } from '@fortawesome/free-solid-svg-icons';
-import { createEventDispatcher } from 'svelte';
+import { createEventDispatcher, type Snippet } from 'svelte';
 import Fa from 'svelte-fa';
 
-export let checked = false;
-export let disabled = false;
-export let indeterminate = false;
-export let disabledTooltip = '';
-export let title = '';
-export let id: string | undefined = undefined;
-export let name: string | undefined = undefined;
-export let required = false;
-
 const dispatch = createEventDispatcher<{ click: boolean }>();
+
+interface Props {
+  onclick?: (checked: boolean) => void;
+  children?: Snippet;
+  class?: string;
+  checked?: boolean;
+  disabled?: boolean;
+  indeterminate?: boolean;
+  disabledTooltip?: string;
+  title?: string;
+  id?: string;
+  name?: string;
+  required?: boolean;
+}
+
+let {
+  onclick,
+  children,
+  class: className = '',
+  checked = $bindable(),
+  disabled = false,
+  indeterminate = false,
+  disabledTooltip = '',
+  title = '',
+  id,
+  name,
+  required = false,
+}: Props = $props();
+
 const faSize = '1.33x';
 
-function onClick(
-  event: MouseEvent & {
-    currentTarget: EventTarget & HTMLInputElement;
-  },
-): void {
+function handleClick(event: MouseEvent & { currentTarget: EventTarget & HTMLInputElement }): void {
   const checked = event.currentTarget.checked;
-  dispatch('click', checked);
+  if (onclick) {
+    onclick(checked);
+  } else {
+    dispatch('click', checked);
+  }
 }
 </script>
 
-<label class="flex flex-row items-center {$$props.class ?? ''}">
-  <div class="relative p-2 self-start" class:mt-0.5={$$slots.default} class:mr-1={$$slots.default}>
+<label class="flex flex-row items-center {className}">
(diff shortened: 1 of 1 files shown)


packages/ui/src/lib/checkbox/Checkbox.svelte
```diff
@@ -1,17 +1,39 @@
...
-export let title = '';
-export let id: string | undefined = undefined;
-export let name: string | undefined = undefined;
-export let required = false;
+interface Props {
+  onclick?: (checked: boolean) => void;
+  children?: Snippet;
+  defSnippet?: Snippet;
+  class?: string;
+  'aria-label'?: string;
```
gastoner: I think that `defSnippet` should be called `default` as it is on original code. This way you would have to change the prosps in every place where is Checkbox used. 

Also not sure with adding the aria-label (I would say next PR)

axel7083: I feel like this is a breaking change (changing the slot name)

bmahabirbu: Thanks all! I  updated the code but it seems I'm getting errors for checked with bind:checked. Should checked be $bindable()?

deboer-tim: Yes (already discussed and code changed, just noting here).

Note @gastoner 's comment on aria-label - if there are no clients currently expecting it then we shouldn't add new property here (PR is just for Svelte migration).

packages/ui/src/lib/checkbox/Checkbox.svelte
```diff
@@ -1,33 +1,49 @@
...
+  disabledTooltip?: string;
+  title?: string;
+  id?: string;
+  name?: string;
+  required?: boolean;
 }
+
+let {
+  onclick = (): void => {
+    checked = !checked;
```
deboer-tim: I think I know _why_ you did this (annoying to figure out how to get the current checked state / gets updated after this event is fired), but currently `checked` is toggled here _and_ bound to the input. i.e. user clicks, we toggle the state which sets it on the input at the same time the input is trying to change the value and svelte syncing it back to the property.

In practice I'm guessing this doesn't cause any problems because Svelte and input are both smart enough to avoid infinite loops and ignore whichever direction loses the race and is setting a value to it's current value. However, it's odd / confusing and I think we need to pick one way to bind the value.

packages/ui/src/lib/checkbox/Checkbox.svelte
```diff
@@ -1,33 +1,49 @@
...
 
-function onClick(
-  event: MouseEvent & {
-    currentTarget: EventTarget & HTMLInputElement;
-  },
-): void {
-  const checked = event.currentTarget.checked;
-  dispatch('click', checked);
+interface Props {
+  onclick?: () => void;
```
deboer-tim: The old on:click dispatch event included the current checked state, but the new onclick function doesn't. Not sure if this is intentional?

deboer-tim (changes requested):
Looking much better, just a few comments.

In most of the other Svelte 5 migration PRs we've tried to keep forced migration at a minimum, but we've also migrated code like the on:click -> onclick throughout renderer to make sure it works and there's less to do when we eventually drop Svelte 4 support.

I'm fine if that's a second PR, but then this one shouldn't say 'fixes' and you should be far enough along that you've tried a few and know that your approach to onclick is going to work.

packages/ui/src/lib/checkbox/Checkbox.svelte
bmahabirbu: Yes that makes sense! Ive added back the original function to handle the mouseevent and call dispatch if onclick isn't passed from the parent. It works properly now.

packages/ui/src/lib/checkbox/Checkbox.svelte
bmahabirbu: Fixed! thanks you

packages/ui/src/lib/checkbox/Checkbox.svelte
```diff
@@ -1,33 +1,53 @@
...
+let {
+  onclick,
+  children,
+  class: className = '',
+  checked = $bindable(),
+  disabled = false,
+  indeterminate = false,
+  disabledTooltip = '',
+  title = '',
+  id = undefined,
```
deboer-tim: FYI you shouldn't need these `= undefined` since they are optional props above, but that's fine.

deboer-tim (approved):
Thanks for fixing the issues! Tested and tried to scan everything again, LGTM.

PR #12060, opened 2025-04-07
chore(workflows): adding build step before test:unit

axel7083:
### What does this PR do?

Following https://github.com/podman-desktop/podman-desktop/pull/12059, making the build step part of the workflows.

To avoid backtracking which package need what package to be able to run the tests, let's build everything. This is needed to be able to define the vitest workspace.

Required for https://github.com/podman-desktop/podman-desktop/pull/12026

Need rebase after
- https://github.com/podman-desktop/podman-desktop/pull/12110

### How to test this PR?

- PIpeline should be :heavy_check_mark:


gastoner (approved):
LGTM Codewise

package.json
```diff
@@ -69,8 +69,8 @@
     "test:extensions:kubectl-cli": "vitest run -r extensions/kubectl-cli --passWithNoTests --coverage ",
     "test:renderer": "pnpm run build:preload:types && pnpm run build:ui && vitest -c packages/renderer/vite.config.js run packages/renderer --passWithNoTests --coverage",
     "test:ui": "vitest --config packages/ui/vite.config.js run --passWithNoTests --coverage",
-    "test:tools": "vitest run tools/src --passWithNoTests --coverage",
-    "test:scripts:stylesheet": "vitest run ./scripts/generate-stylesheet.spec.ts --passWithNoTests",
+    "test:tools": "vitest run -r tools/src --passWithNoTests --coverage",
+    "test:scripts:stylesheet": "vitest run -r ./scripts/generate-stylesheet.spec.ts --passWithNoTests",
```
axel7083: not using the `--root` option is very problematic with such large project, the vitest coverage is parsing ALL the file in the project, leading to Java out of Memory.

Anyway this is a tmp fix as https://github.com/podman-desktop/podman-desktop/pull/12111 those option are not needed anymore

benoitf (commented):
I don't get why we need to build the project before running test
it looks like we're spending time for nothing

axel7083:
> I don't get why we need to build the project before running test it looks like we're spending time for nothing

@benoitf today when we run `pnpm test:renderer` we build some packages 

https://github.com/podman-desktop/podman-desktop/blob/1196ba7797c1856c42535c0e586fbb216efa98aa/package.json#L70

The following packages `pnpm run build:preload:types && pnpm run build:ui` **need** to be built to allow the renderer package to use the library.

I agree with the _we lose time_, but from a vitest perspective, when we are at the workspace, we cannot run a command before ag given project (E.g. renderer) is tested.

So we either hardcode in `test:unit` that we have two package to build before, or we build everything.

benoitf (changes requested):
the idea was that you clone the project, modify the source code and run the tests and it works

here you've moved some logic to the CI github scripts which do not apply on the configuration of a user

also building everything is not required as basically we need only the types & ui package

unit tests need to work OOTB for newcomers without requiring to launch a build task

axel7083:
Will keep the hardcoded then 👍

Issue #11557, opened 2025-03-06
Add update information to Providers


SoniaSandler:
### Is your enhancement related to a problem? Please describe

As part of https://github.com/podman-desktop/podman-desktop/issues/11426, I found out that providers created by CLI tools extensions don't have update information. So, for example, Kind will show that an update is available in its CLI tool in the app, but it will not show up in the provider created by the kind extension.

### Describe the solution you'd like

To be able to show that there is an update available in the provider widget, we need to add the available update information to the providers created by the different extensions.
An example can be found in the Podman extension:
https://github.com/podman-desktop/podman-desktop/blob/9018d2f073da1ca5ed57d89edc493013542b9c53/extensions/podman/packages/extension/src/extension.ts#L1027-L1044

axel7083:
Do we also plan to add `registerInstallation` for the provider ? we seems to duplicate a lot of logic between the providers and the CLI

https://github.com/podman-desktop/podman-desktop/blob/c0990a328819be05f6c528db21e7b3f90768186b/packages/extension-api/src/extension-api.d.ts#L670

I don't know what is the right path here, we have two logic, maybe we should for the provider widget, get the cli tools registered, and use that.. I don't know

SoniaSandler:
@axel7083 I missed `registerInstallation` for Minkube and Kind (tried to replicate what's going on in the Podman extension provider).
One issue with CLI tools is that all the installation and updates go through the CLI tools page. This ticket is so that we could display if there is an update available in the provider widgets. 
Right now, there won't be update functionality from the statusbar (just showing that an update is available), but I'm pretty sure there was some discussion about adding more actions to the statubar provider widgets, so it might be something that we will do in the future.

deboer-tim:
Ditto to @axel7083's concern. A lot of providers are just going to be binaries, we should encourage that they are registered as CLI tools, and they only need to provide one code path / metadata for updating/version availability, etc. Updates accessible from multiple places is good, but we don't want them to have to implement things twice or have different things reported depending on where you're looking in the UI.

I'm less familiar with this code so this may make little sense, but even something like matching the API so you can provide the same object to both would be valuable.

SoniaSandler:
> Ditto to [@axel7083](https://github.com/axel7083)'s concern. A lot of providers are just going to be binaries, we should encourage that they are registered as CLI tools, and they only need to provide one code path / metadata for updating/version availability, etc.

[...]

PR #12068, opened 2025-04-07
chore: check if a provider is also a CLI tool

SoniaSandler:

Related to https://github.com/podman-desktop/podman-desktop/issues/11557

### How to test this PR?


SoniaSandler:
@deboer-tim @axel7083 based on the discussion from https://github.com/podman-desktop/podman-desktop/issues/11557, I've come up with this idea that requires no code addition in the extensions and uses only existing information. As both the providers and the CLI tools are created by the same extensions, we can find the provider's respective CLI tool info by comparing the extension ids. This way we can use the CLI tool's existing update info and operations for the provider.
However, I'm not sure if it something that should/can be done, using CLI tool info for providers

PR #11989, opened 2025-04-02
feat: experimental feature with GitHub reactions

axel7083:
### Screenshot / video of UI

image

image

### What issues does this PR fix or reference?

Fixes https://github.com/podman-desktop/podman-desktop/issues/10227

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature

Issue #10227, opened 2024-12-04
Feedback mechanism for experimental features


slemeur:
### Is your feature request related to a problem? Please describe

This issue will be the implementation of the feedback mechanism for experimental features corresponding to the mockup in issue #10223 

### Describe the solution you'd like

This issue will be the implementation of the feedback mechanism for experimental features corresponding to the mockup in issue #10223

axel7083:
Let's look at the API

Let's take the following discussion as example https://github.com/podman-desktop/podman-desktop/discussions/10533

The api is https://api.github.com/repos/podman-desktop/podman-desktop/discussions/10533 and the content is the follow (trunked)

````json
{
...
  "reactions": {
  "url": "https://api.github.com/repos/podman-desktop/podman-desktop/discussions/10533/reactions",
  "total_count": 0,
  "+1": 0,
  "-1": 0,
  "laugh": 0,
  "hooray": 0,
  "confused": 0,
  "heart": 0,
  "rocket": 0,
  "eyes": 0
  },
}
````

We would need a way to only fetch those information or refresh those when the user open the experimental page, so we avoid calling repeatably the api.

I also want a way to disable that, a user should be able to decide

axel7083:
Moving to another milestone, as there is still so much discussion needed for the technical solution (ref https://github.com/podman-desktop/podman-desktop/pull/12069#pullrequestreview-2748860213)

axel7083:
Okay following on-going discussion on this feature, this is feasible but would add lot of works for a feature that would be display only.

More info about the technical solution that was brought up in https://github.com/podman-desktop/podman-desktop/pull/12069#pullrequestreview-2748860213

PR #12069, opened 2025-04-07
feat(main): creating GitHub client class

axel7083:
### What does this PR do?

To split https://github.com/podman-desktop/podman-desktop/pull/11989 and ease review, let's see each logic individually.

The issue https://github.com/podman-desktop/podman-desktop/issues/10227 wants a logic, to display the reactions for the discussions linked to each experimental feature.

We need
- GitHub client (handling the interaction with GitHub for us)
- Reaction Registry: holding the reactions (avoid calling GitHub every time, some caching)
- Some frontend logic

This PR adds the GitHub client, it need to handle some stuff: the rate limits, to avoid calling every X seconds, we can add a _middleware_ logic wrapping each request. In each api call, GitHub return some headers with the rate limit remaining[^1]. 

Using the Octokit[^2] library (that we already use in several places) we can implement the logic above, and pass its instance to anything that would need it: here in a follow-up PR, the _Reaction Registry_

[^1]: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api
[^2]: https://www.npmjs.com/package/@octokit/rest

Part of https://github.com/podman-desktop/podman-desktop/issues/10227

### How to test this PR?

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,7 +177,8 @@
     "validator": "^13.15.0",
     "vite": "^6.2.5",
     "vitest": "^3.1.1",
-    "xvfb-maybe": "^0.2.1"
+    "xvfb-maybe": "^0.2.1",
+    "@octokit/rest": "^21.1.1"
   },
   "dependencies": {
     "@docker/extension-api-client-types": "0.3.4",
--- a/packages/main/src/plugin/github/github-client.spec.ts
+++ b/packages/main/src/plugin/github/github-client.spec.ts
@@ -1,0 +1,183 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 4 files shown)


axel7083:
> Attention: Patch coverage is `93.10345%` with `4 lines` in your changes missing coverage. Please review.

https://github.com/podman-desktop/podman-desktop/pull/12069/commits/9c2413c546399133050c3a8171e4aa38533c4a7d should have added the missing branch coverage to get :100:

benoitf (changes requested):
there should be no GitHub client in Podman Desktop.

why ?

if it's for gathering feedback data, Podman Desktop should not query GitHub for that, as it won't work for most of people due to default GH limits

Then, in case we would need to query, it should use the github auth provider, and with the default limit, there should not be any problem

so we would not need this.

Podman Desktop, like for releases notes, should not query directly the API, but an aggregated content that is exposed on `https://*.podman-desktop.io/*` so there is no GH limit at all

axel7083:
> there should be no GitHub client in Podman Desktop.

I don't have a problem in going into another direction, however I must admit I would have been happier knowing this before doing the full implementation :cry: .

I reached out during a stand-up and had a discussion with @deboer-tim, we might need to review how we discuss technical solution as this prove that we have some area where we could improve the way we communicate and choose the right solution that fit our need.

PR #12074, opened 2025-04-08
chore: Continuation of the isUnixLike() crusade

arrowd:
### What does this PR do?

This PR adds more `isUnixLike()` usage to make Podman Desktop work on FreeBSD.

### DIFF
--- a/extensions/compose/src/detect.spec.ts
+++ b/extensions/compose/src/detect.spec.ts
@@ -33,6 +33,7 @@ const osMock: OS = {
   isWindows: vi.fn(),
   isLinux: vi.fn(),
   isMac: vi.fn(),
+  isUnixLike: vi.fn(),
 };
 
 let detect: Detect;
@@ -49,7 +50,7 @@ vi.mock('@podman-desktop/api', async () => {
       exec: vi.fn(),
     },
     env: {
-      isLinux: false,
+      isUnixLike: false,
       isWindows: false,
       isMac: false,
     },
@@ -131,8 +132,8 @@ describe('Check getDockerComposePath uses proper tooling by platform', () => {
     );
   });
 
-  test('linux should use which', async () => {
-    (extensionApi.env.isLinux as boolean) = true;
+  test('unix-like OSes should use which', async () => {
+    (extensionApi.env.isUnixLike as boolean) = true;
     (extensionApi.env.isWindows as boolean) = false;
     (extensionApi.env.isMac as boolean) = false;
 
@@ -143,7 +144,7 @@ describe('Check getDockerComposePath uses proper tooling by platform', () => {
   test('mac should use which', async () => {
     (extensionApi.env.isMac as boolean) = true;
     (extensionApi.env.isWindows as boolean) = false;
-    (extensionApi.env.isLinux as boolean) = false;
+    (extensionApi.env.isUnixLike as boolean) = false;
 
     await detect.getDockerComposePath('docker-compose');
     expect(extensionApi.process.exec).toHaveBeenCalledWith('which', ['docker-compose']);
@@ -152,7 +153,7 @@ describe('Check getDockerComposePath uses proper tooling by platform', () => {
   test('windows should use which', async () => {
     (extensionApi.env.isWindows as boolean) = true;
     (extensionApi.env.isMac as boolean) = false;
-    (extensionApi.env.isLinux as boolean) = false;
+    (extensionApi.env.isUnixLike as boolean) = false;
 
     await detect.getDockerComposePath('docker-compose');
     expect(extensionApi.process.exec).toHaveBeenCalledWith('where.exe', ['docker-compose']);
@@ -189,7 +190,7 @@ describe('parseVersion', () => {
 
 describe('Check default socket path', async () => {
   test('linux', async () => {
-    (osMock.isLinux as Mock).mockReturnValue(true);
+    (osMock.isUnixLike as Mock).mockReturnValue(true);
     (osMock.isMac as Mock).mockReturnValue(false);
     (osMock.isWindows as Mock).mockReturnValue(false);
     const result = detect.getSocketPath();
@@ -197,7 +198,7 @@ describe('Check default socket path', async () => {
   });
 
   test('macOS', async () => {
-    (osMock.isLinux as Mock).mockReturnValue(false);
+    (osMock.isUnixLike as Mock).mockReturnValue(false);
     (osMock.isMac as Mock).mockReturnValue(true);
     (osMock.isWindows as Mock).mockReturnValue(false);
     const result = detect.getSocketPath();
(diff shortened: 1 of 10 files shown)


arrowd:
I'd love to see this included in the upcoming 1.18.0

cdrage (approved):
I was able to test this on Linux (Fedora 42) and it worked fine :). So I can confirm that this works.

The code looks straight forward and mostly a find+replace of isLinux. So no problems here from me, it LGTM!

Out of curiosity, what version of FreeBSD have you been testing against?

arrowd:
I'm running a bleeding edge 15-CURRENT, but it shouldn't change much wrt to Podman Desktop

arrowd:
Can we get this in?

benoitf (changes requested):
hello, thanks for the contrib

the problem I see is that we'll have two exported functions

isLinux and isUnixLike

meaning that all existing extensions would call isLinux then it won't work on FreeBSD

and I'm not sure people will think "ok this time I use isUnixLike instead of isLinux"

at the end, from what I see, we do the same code

I'm wondering then, if we should not have isLinux returning true on freebsd

so, it would break only on a platform where we really need to do things specific to Freebsd that differs to Linux behavior

and in that case I'm not sure we have enough usecases, and for that the extension could query directly the OS calls

so I think, it would be easier/simpler to have isLinux being true on Freebsd (I understand the craziness of this kind of assertion) but I think it will work in the spirit of the checks people might do.

arrowd:
> meaning that all existing extensions would call isLinux then it won't work on FreeBSD

Well, right now the whole Podman Desktop isn't really usable on FreeBSD, which is exactly what I'm trying to fix. I may as well come to fix such extensions too.

> and I'm not sure people will think "ok this time I use isUnixLike instead of isLinux"

Why? It doesn't take a senior-level software engineer or a PhD CS degree to discern between a generic POSIX functionality and something Linux-specific. I'm sure some people will start using `isUnixLike` if it becomes available in the API, but even if not, they'd stick with `isLinux` and I will fix it later.

> I'm wondering then, if we should not have isLinux returning true on freebsd

I don't think it is a good idea, because it will start triggering code paths that might not intended to be run on FreeBSD. Actually, in this very PR I left out some `isLinux` code, which was Linux-specific for real: https://github.com/podman-desktop/podman-desktop/blob/0d5cfb8532cca202d7f37faba1ae1d19a7046b6c/extensions/podman/packages/extension/src/extension.ts#L1673

With all my FreeBSD porting experience I can say that debugging cases when you actually get to execute some Linux-specific code (like, reading `/proc`) is much harder to identify and debug than cases when the whole piece didn't get executed at all.

Another argument for `isUnixLike` is that it allows me to see which places I already checked. Once I finish going over the code base all `isLinux` calls that would left be the real cases of Linux-specific code.

Finally, going the `isLinux = isFreeBSD` route feels like a movement backwards - I already introduced `isUnixLike` in some places. We'd have to revert them. So, I still insist on the `isUnixLike` approach, but the final word is of course yours.

cdrage:
> > meaning that all existing extensions would call isLinux then it won't work on FreeBSD
> 
> Well, right now the whole Podman Desktop isn't really usable on FreeBSD, which is exactly what I'm trying to fix. I may as well come to fix such extensions too.
> 
> > and I'm not sure people will think "ok this time I use isUnixLike instead of isLinux"
> 
> Why? It doesn't take a senior-level software engineer or a PhD CS degree to discern between a generic POSIX functionality and something Linux-specific. I'm sure some people will start using `isUnixLike` if it becomes available in the API, but even if not, they'd stick with `isLinux` and I will fix it later.
> 
> > I'm wondering then, if we should not have isLinux returning true on freebsd
> 
> I don't think it is a good idea, because it will start triggering code paths that might not intended to be run on FreeBSD. Actually, in this very PR I left out some `isLinux` code, which was Linux-specific for real:
> 
> https://github.com/podman-desktop/podman-desktop/blob/0d5cfb8532cca202d7f37faba1ae1d19a7046b6c/extensions/podman/packages/extension/src/extension.ts#L1673
> 
> With all my FreeBSD porting experience I can say that debugging cases when you actually get to execute some Linux-specific code (like, reading `/proc`) is much harder to identify and debug than cases when the whole piece didn't get executed at all.
> 
> Another argument for `isUnixLike` is that it allows me to see which places I already checked. Once I finish going over the code base all `isLinux` calls that would left be the real cases of Linux-specific code.
> 
> Finally, going the `isLinux = isFreeBSD` route feels like a movement backwards - I already introduced `isUnixLike` in some places. We'd have to revert them. So, I still insist on the `isUnixLike` approach, but the final word is of course yours.

I agree that we should have something at least for FreeBSD, even if it's a "isFreeBSD" functionality within the code. 

Although FreeBSD is not officially supported by Podman Desktop, having some form of functionality to differentiate between linux / bsd distros would be beneficial for community support.

Perhaps as a compromise @benoitf we can have isLinux as well as isFreeBSD? Or some other alternative.

arrowd:
Gentle ping.

arrowd:
Is this PR rejected?

benoitf:
> Is this PR rejected?

Hello @arrowd and thanks for your contribution. There was a discussion and we won't expose the isUnixLike to extensions.

The thing we can do is to make isLinux returning true even if it's FreeBSD.

Why? because most of the extensions writers will always use the kind of switch "if isWindows do this else if macOS do this and if Linux do this"

we should have made a isUnix rather than isLinux but now it's too late to change that.

so most of the time the extension will never implement the case of "isFreeBSD". And then most of the time the extension will never handle the case.

I think 90% of the isLinux code works in extensions it it's using FreeBSD. Then we can inside the extension do a specific handle for FreeBSD.

arrowd:
> Why? because most of the extensions writers will always use the kind of switch "if isWindows do this else if macOS do this and if Linux do this"

I still don't get it why would they, but whelp. This is sad.

dfr:
> > Is this PR rejected?
> 
> Hello @arrowd and thanks for your contribution. There was a discussion and we won't expose the isUnixLike to extensions.
> 
> The thing we can do is to make isLinux returning true even if it's FreeBSD.
> 
> Why? because most of the extensions writers will always use the kind of switch "if isWindows do this else if macOS do this and if Linux do this"
> 
> we should have made a isUnix rather than isLinux but now it's too late to change that.
> 
> so most of the time the extension will never implement the case of "isFreeBSD". And then most of the time the extension will never handle the case.
> 
> I think 90% of the isLinux code works in extensions it it's using FreeBSD. Then we can inside the extension do a specific handle for FreeBSD.

Having isLinux return true on FreeBSD doesn't feel right to me. It will work for the common ground between the two systems but doesn't allow for the cases where it makes sense to have different code paths on FreeBSD and Linux. Having something like isUnixLike (isPosix?) as @arrowd suggested seems clearer to me and should make it easier to have a working Podman Desktop on FreeBSD to go with the working Podman CLI and native FreeBSD containers.

Making isLinux return true on FreeBSD is likely to be confusing for maintainers of both systems (IMO). Perhaps we could discuss this, e.g. in a Podman Community Meeting?

PR #12076, opened 2025-04-08
test(ui): adding svelte legacy usage for button

axel7083:
### What does this PR do?

Before starting the migration of the `Button` component to svelte5 (required for https://github.com/podman-desktop/podman-desktop/issues/11573).

Let's add some tests that ensure that legacy component are still able to use it properly. The `ButtonSvelteLegacyTest.svelte` component use the button and test for default slot usage, and on click with legacy synthax (`on:click`).

Adding this will ensure we can still use the component from svelte4 and before, when the component will be migrated to svelte5.

### How to test this PR?


gastoner (approved):

benoitf (approved):
nice approach.

maybe use `Svelte4Test` rather than `SvelteLegacy` because let say Svelte6 is there, legacy is less meaningful

PR #12077, opened 2025-04-08
fix(ui): race condition in ImageDetailsCheck tests

### LINKED ISSUE #12078: bug(ui): race condition in `ImageDetailsCheck.svelte`
### Bug description

Dependabot tried to update svelte (https://github.com/podman-desktop/podman-desktop/pull/12071) but one test is failing.

They don't provide much info about the change, but anyway, looking deep inside the `ImageDetailsCheck` we can see the issue. 

Calling twice the `handleAbort` function will call twice the cancel token; which is not the expected behavior. (as per test file).

https://github.com/podman-desktop/podman-desktop/blob/bb1c3c0fbddcbabd2d5950d2939f6e920bc0d40f/packages/renderer/src/lib/image/ImageDetailsCheck.svelte#L117-L130

As shown above, we set `cancellableTokenId = 0; ` at the end, after two async method. Therefore if we call twice the method, we may have a race condition.

### Operating system

Fedora 41

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

See https://github.com/podman-desktop/podman-desktop/pull/12071

### Relevant log output

```shell
Error: AssertionError: expected "spy" to be called 1 times, but got 2 times
 ❯ src/lib/image/ImageDetailsCheck.spec.ts:204:26
```

### Additional context

_No response_

axel7083:
### What does this PR do?

While dependabot tried to update svelte, we got an error on `ImageDetailsCheck` tests. This is because of a race condition, where svelte better manage the unmount of components, and therefore make our test fails because it was depending on the _order_  svelte was destroying component

Required for https://github.com/podman-desktop/podman-desktop/pull/12071 to pass

Fixes https://github.com/podman-desktop/podman-desktop/issues/12078

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/ImageDetailsCheck.spec.ts
+++ b/packages/renderer/src/lib/image/ImageDetailsCheck.spec.ts
@@ -35,7 +35,7 @@ beforeAll(() => {
   Object.defineProperty(window, 'getCancellableTokenSource', { value: getCancellableTokenSourceMock });
   getCancellableTokenSourceMock.mockImplementation(() => tokenID);
   Object.defineProperty(window, 'imageCheck', { value: imageCheckMock });
-  Object.defineProperty(window, 'cancelToken', { value: cancelTokenSpy });
+  Object.defineProperty(window, 'cancelToken', { value: cancelTokenSpy.mockResolvedValue(undefined) });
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM

PR #12086, opened 2025-04-08
fix(Button): binding usage and event type

axel7083:
### What does this PR do?

In svelte4 there is **no** typecheck for using `bind:<prop>` but in svelte5 we will get an error if we try to bind a non-bindable prop.

This PR remove all the unnecessary `bind:` for the Button component as they have so effect. But they make the typecheck fails in https://github.com/podman-desktop/podman-desktop/pull/12079 . It also update the event type used for on:click.

### Screenshot / video of UI

N/A

Required for https://github.com/podman-desktop/podman-desktop/issues/11573

### How to test this PR?

- Pipeline should be :heavy_check_mark:


jeffmaury (approved):
LGTM

deboer-tim (approved):

PR #12103, opened 2025-04-09
refactor(extensions): simplify vitest config

axel7083:
### What does this PR do?

Incremental change for https://github.com/podman-desktop/podman-desktop/pull/12026: removing vitest.config.js file as requested in https://github.com/podman-desktop/podman-desktop/pull/12026#pullrequestreview-2743427312 

### Notes

- Removing import of `vitest-shared-extensions.config` in `extensions/*` folder

### Screenshot / video of UI

N/A

### How to test this PR?

- Pipeline should be ✅

PR #12111, opened 2025-04-09
feat(vitest): configure workspace

### LINKED ISSUE #11985: Codecov spamming warning on `*.svelte` files

axel7083:
### What does this PR do?

Configure vitest workspace (finally 👏 )

Requires rebase after
- ~~https://github.com/podman-desktop/podman-desktop/pull/12060~~
- https://github.com/podman-desktop/podman-desktop/pull/12110

Follow-up on
- https://github.com/podman-desktop/podman-desktop/pull/12108
- https://github.com/podman-desktop/podman-desktop/pull/12104
- https://github.com/podman-desktop/podman-desktop/pull/12103

Fixes https://github.com/podman-desktop/podman-desktop/pull/12026

### How to test this PR?

- Pipeline should be ✅ and codecov VERY happy

> https://app.codecov.io/gh/podman-desktop/podman-desktop/pull/12111/tree (74.56% coverage)


vitest.config.js
```diff
@@ -0,0 +1,55 @@
...
+  '**/dist/**',
+  '**/.{cache,git,idea,output,temp,cdix}/**',
+  '**/*.cdix/**',
+];
+
+/**
+ * vitest workspace configuration for unit tests
+ */
+export default defineConfig({
+  test: {
```
benoitf, line 31: should it be there where we add the retry to 3 ?

package.json
```diff
@@ -54,25 +54,25 @@
...
+    "test:preload-webview": "vitest --run --project=preload-webview",
+    "test:extensions": "vitest --run --project=extensions --passWithNoTests && pnpm run test:extensions:compose && pnpm run test:extensions:kind && pnpm run test:extensions:docker && pnpm run test:extensions:lima && pnpm run test:extensions:kube && pnpm run test:extensions:podman && pnpm run test:extensions:registries && pnpm run test:extensions:kubectl-cli",
+    "test:extensions:kind": "vitest --run --project=kind --passWithNoTests",
+    "test:extensions:compose": "vitest --run --project=compose --passWithNoTests",
+    "test:extensions:docker": "vitest --run --project=docker --passWithNoTests",
+    "test:extensions:kube": "vitest --run --project=kube-context --passWithNoTests ",
+    "test:extensions:lima": "vitest --run --project=lima --passWithNoTests ",
+    "test:extensions:podman": "vitest --run --project=podman --passWithNoTests ",
+    "test:extensions:registries": "vitest --run --project=registries --passWithNoTests ",
+    "test:extensions:kubectl-cli": "vitest --run --project=kubectl-cli --passWithNoTests",
```
benoitf, line 69: as a follow-up, I'm wondering if we should not name project with `extension/` prefix

like `--project=extension/kind`

vitest.config.js
axel7083: if we add it here, we retry 3 times for all the tests, where the previous config was only for the renderer and ui package

package.json
```diff
@@ -54,25 +54,25 @@
     "test:e2e:remote:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- npx playwright test tests/playwright/src/special-specs/podman-remote/ -g @podman-remote",
     "test:e2e:update": "pnpm run test:e2e:build && pnpm run test:e2e:update:run",
     "test:e2e:update:run": "xvfb-maybe --auto-servernum --server-args='-screen 0 1280x960x24' -- npx playwright test tests/playwright/src/special-specs/installation/ -g @update-install",
-    "test:main": "vitest run -r packages/main --passWithNoTests --coverage",
-    "test:preload": "vitest run -r packages/preload --passWithNoTests --coverage",
-    "test:preload-docker-extension": "vitest run -r packages/preload-docker-extension --passWithNoTests --coverage",
-    "test:preload-webview": "vitest run -r packages/preload-webview --coverage",
-    "test:extensions": "vitest run -r extensions --passWithNoTests --coverage && pnpm run test:extensions:compose && pnpm run test:extensions:kind && pnpm run test:extensions:docker && pnpm run test:extensions:lima && pnpm run test:extensions:kube && pnpm run test:extensions:podman && pnpm run test:extensions:registries && pnpm run test:extensions:kubectl-cli",
```
benoitf, line 61: wondering if it should be:
`pnpm -r --filter './extensions/**' test`

so it's in //

package.json
axel7083: The project name is defined as the following

> For project configurations defined with glob syntax, Vitest will default to using the "name" property in the nearest package.json file or, if none exists, the folder name.[^1]

So you can't have extension/kind except if you change the `extension/kind/package.json` property name to `extension/kind`.

[^1]: https://vitest.dev/guide/workspace

package.json
axel7083: not sure to understand the synthax

package.json
benoitf: well, you can set the `name` property of vitest

https://vitest.dev/config/#name

```js
  test: {
    name: 'extension/kind',
```

package.json
benoitf: it's to run the command `test` in all the folders matching '**/extensions/*'

benoitf: (idea behind would be to find a way to reduce the entries in the package.json)
where for example we could run `pnpm test:extension <extension-name>`

axel7083: Maybe we could make the `--project=extensions/*` working (I don't if they support it, but we could add it anyway)

package.json
axel7083: Oh yeah, didn't think it would work

benoitf (approved):
I noticed I commented with follow-up comments but forgot to approve the pr

PR #12129, opened 2025-04-10
test(e2e): add env vars to keep traces and videos on pass 

### LINKED ISSUE #12058: Keep the recording of the ui stress e2e test for better overview

I was looking at ui-stress test results and I think we should store the recording to better understand the how test performs. We might be missing some aspects of how test/app performs.

Store more test artifacts by using `KEEP_VIDEOS_ON_PASS` and `KEEP_TRACES_ON_PASS` env. vars.

_No response_

### Additional context

_No response_

danivilla9:
(in ui stress test)

### What does this PR do?
Adds KEEP_TRACES_ON_PASS and KEEP_VIDEOS_ON_PASS to ui-stress-test test case
https://github.com/podman-desktop/podman-desktop/issues/12058 

### How to test this PR?
Run `pnpm test:e2e:ui-stress`

### DIFF
--- a/tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
+++ b/tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
@@ -23,6 +23,8 @@ console.log(`numberOfObjects => ${numberOfObjects}`);
 
 test.beforeAll(async ({ runner, welcomePage, page }) => {
   runner.setVideoAndTraceName('ui-stress-e2e');
+  process.env.KEEP_TRACES_ON_PASS = 'true';
+  process.env.KEEP_VIDEOS_ON_PASS = 'true';
   await welcomePage.handleWelcomePage(true);
   await waitForPodmanMachineStartup(page);
 });


tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
```diff
@@ -23,6 +23,8 @@ console.log(`numberOfObjects => ${numberOfObjects}`);
 
 test.beforeAll(async ({ runner, welcomePage, page }) => {
   runner.setVideoAndTraceName('ui-stress-e2e');
+  process.env.KEEP_TRACES_ON_PASS = 'true';
+  process.env.KEEP_VIDEOS_ON_PASS = 'true';
```
cbr7, line 27: Are you use this works? I wasn't aware that you can use `process.env` as a setter also, I thought it can only read the values from the system environment variables.

cbr7 (changes requested):
I don't think you can use process.env as a setter.

tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
cbr7: Why not add the env vars to the stress test github workflow? The runner will pick up the values from there.

danivilla9: I saw that it was done like this on the podman-machine-rootless spec file: https://github.com/podman-desktop/podman-desktop/blob/fd095835aedac266de3cedd3d0b10a72fee5d190/tests/playwright/src/specs/podman-machine-rootless.spec.ts#L49 
So I figured that it would be the way to go. But from what I've searched you might be right, I think defining it here might be too late. The other one should be moved too, though

cbr7: If it's like that in some other place that needs to be changed too.

odockal: True, cannot we have this options as part of runner options?

cbr7: @odockal this option specifically probably not, but there is no reason why a parallel configurable option couldn't be created that would mirror the env var functionality.

danivilla9:
Moved the env var change to the workflow file instead from the spec file here: https://github.com/podman-desktop/e2e/pull/352 closing this pr.

Issue #11741, opened 2025-03-19
review clomonitor result and add missing items


benoitf:
### Is your enhancement related to a problem? Please describe

Podman Desktop is now listed on clomonitor

needs to review missing items and add them

### Describe the solution you'd like

Look at https://clomonitor.io/docs/topics/checks/ how the missing elements are detected and then add the missing items

benoitf:
score has been improved a lot, will open follow-up dedicated issues

PR #12138, opened 2025-04-10
chore: add security insights file

benoitf:
### What does this PR do?
Introduce a security insights file
https://github.com/ossf/security-insights-spec

#11741 

### How to test this PR?


benoitf:
validated the file against the cue schema https://github.com/ossf/security-insights-spec/blob/main/schema.cue

benoitf:
draft as it seems it's working only with the v1 spec file, not the latest v2

deboer-tim (approved):

PR #12139, opened 2025-04-10
fix: image details navigation

### LINKED ISSUE #12120: Image navigation is broken
### Bug description

Using the API to navigate to an image details is broken. There's an exception in the log and the page isn't opened.

### Operating system

macOS 15.4

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

Use the extension navigation API or within renderer use code like this from somewhere like ImageColumnName.openDetails():
```
handleNavigation({
      page: NavigationPage.IMAGE,
      parameters: { id: image.id, engineId: image.engineId, tag: image.base64RepoTag },
    });
```

### Relevant log output

```shell
Uncaught ReferenceError: Buffer is not defined
    at handleNavigation (navigation.ts:81:76)
    at openDetails (ImageColumnName.svelte:17:5)
    at HTMLButtonElement.<anonymous> (ImageColumnName.svelte:26:64)
    at chunk-F7W2M2KK.js?v=bccfaac4:2891:51
    at without_reactive_context (chunk-F7W2M2KK.js?v=bccfaac4:2843:12)
    at HTMLButtonElement.target_handler (chunk-F7W2M2KK.js?v=bccfaac4:2890:14)
```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Navigating to an image (image details) was failing because Buffer was not imported. Tests worked because they run on Node where buffer exists by default.

Secondary failure because the default tab is the /summary and we weren't adding it to the URL (and don't auto-forward). Test was checking for wrong URL, fixed.

### Screenshot / video of UI

N/A

Fixes #12120.

### How to test this PR?

I'll do a follow-up PR to use this API from the Images page, but for now just code review.

### DIFF
--- a/packages/renderer/src/navigation.spec.ts
+++ b/packages/renderer/src/navigation.spec.ts
@@ -114,7 +114,7 @@ test(`Test navigationHandle for ${NavigationPage.IMAGE}`, () => {
     parameters: { id: '123', engineId: 'dummyEngineId', tag: 'dummyTag' },
   });
 
-  expect(vi.mocked(router.goto)).toHaveBeenCalledWith('/images/123/dummyEngineId/ZHVtbXlUYWc=');
+  expect(vi.mocked(router.goto)).toHaveBeenCalledWith('/images/123/dummyEngineId/ZHVtbXlUYWc=/summary');
 });
 
 test(`Test navigationHandle for ${NavigationPage.ONBOARDING}`, () => {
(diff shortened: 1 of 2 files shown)


axel7083 (approved):

cdrage (approved):

deboer-tim:
Force pushed (for no real reason), only change was adding //eslint-disable and fixing the import. I hadn't noticed but linting changed the import from buffer to node:buffer, which caused the tests to fail.

PR #12170, opened 2025-04-11
chore: image and manifest details navigation

deboer-tim:
### What does this PR do?

Two prior PRs fixed navigation to image details and added navigation to manifests. This just switches the links from the Images list Name column to use the navigation instead of directly using the router - minor cleanup and confirms the nav API works.

Fixed tests to match, including repotag that was inconsistent with the name and tag, and checking where manifests links to.

### Screenshot / video of UI

N/A

Follow up from #12139, #12143.

### How to test this PR?

Go to the Images page and click on an image and a manifest (can use `podman manifest create` if you don't have one) to confirm they both still go to details.

### DIFF
--- a/packages/renderer/src/lib/image/ImageColumnName.svelte
+++ b/packages/renderer/src/lib/image/ImageColumnName.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
-import { router } from 'tinro';
+import { handleNavigation } from '/@/navigation';
+import { NavigationPage } from '/@api/navigation-page';
 
 import Badge from '../ui/Badge.svelte';
 import type { ImageInfoUI } from './ImageInfoUI';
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/image/ImageColumnName.svelte
```diff
@@ -8,9 +9,15 @@ export let object: ImageInfoUI;
...
+    handleNavigation({
+      page: NavigationPage.MANIFEST,
+      parameters: { id: image.id, engineId: image.engineId, tag: image.name + ':' + image.tag },
+    });
   } else {
-    router.goto(`/images/${image.id}/${image.engineId}/${image.base64RepoTag}/summary`);
+    handleNavigation({
+      page: NavigationPage.IMAGE,
+      parameters: { id: image.id, engineId: image.engineId, tag: image.name + ':' + image.tag },
+    });
```
feloy: I suppose that the `parameters` field is expected to always be the same (now and in the future), we could have the switch in the `page` field only

feloy (approved):

packages/renderer/src/lib/image/ImageColumnName.svelte
deboer-tim: Yeah, I wouldn't expect them to ever differ and that is much simpler - change pushed.

PR #12175, opened 2025-04-11
chore: use duration column for volume age

deboer-tim:
### What does this PR do?

Switches the VolumeList from using a volume's age (recalculated via timers) to using its created time and a duration column (which recalculates itself).

This allows all the timers to be removed from the list. Age should probably be removed too, but it is still used by the details page.

### Screenshot / video of UI

N/A

Re-found while building the bootc Images list, was originally raised during #4972.

Part of #12176.

### How to test this PR?

Create some volumes, confirm Age column still works.

### DIFF
--- a/packages/renderer/src/lib/volume/VolumesList.svelte
+++ b/packages/renderer/src/lib/volume/VolumesList.svelte
@@ -6,6 +6,7 @@ import {
   NavPage,
   Table,
   TableColumn,
+  TableDurationColumn,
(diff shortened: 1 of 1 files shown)


deboer-tim:
Rebased since it had been 5 days. @cdrage @jeffmaury could I get a review?

axel7083 (approved):

benoitf (approved):

Issue #11586, opened 2025-03-07
Migrate UI component table to Svelte 5


jeffmaury:
https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/table/DurationColumn.svelte#L1

https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/table/SimpleColumn.svelte#L1

https://github.com/podman-desktop/podman-desktop/blob/ce53e6c5bd49e87ed204b4275510079557f03a92/packages/ui/src/lib/table/Table.svelte#L1

jeffmaury:
Replaced by #13892

PR #12194, opened 2025-04-14
chore: migrate empty screen to svelte 5

### LINKED ISSUE #12145: Remove default empty screen message of 'Message'

The default empty screen message (e.g., shown if you use a customer upper control like here: https://github.com/podman-desktop/extension-bootc/pull/1526#pullrequestreview-2756901418) is the string 'Message'.

The empty message should be blank, since you have the choice of custom upper component, message, or both. (i.e. message should truly be optional)

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Migrates the empty screen to Svelte 5 using patterns already established in other components. Updated usage in /renderer to Svelte 5 syntax.

One minor cleanup to default message to fix second issue.

### Screenshot / video of UI

N/A

First half of #11586.
Fixes #12145.

### How to test this PR?

Check empty screens to confirm behaviour.

### DIFF
--- a/packages/ui/src/lib/screen/EmptyScreen.svelte
+++ b/packages/ui/src/lib/screen/EmptyScreen.svelte
@@ -1,26 +1,48 @@
 <script lang="ts">
 import { faPaste } from '@fortawesome/free-solid-svg-icons';
-import { createEventDispatcher, onMount } from 'svelte';
+import { createEventDispatcher, onMount, type Snippet } from 'svelte';
 import Fa from 'svelte-fa';
 
 import Button from '../button/Button.svelte';
 import { isFontAwesomeIcon } from '../utils/icon-utils';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let icon: any;
-export let title = 'No title';
-export let message = 'Message';
-export let detail = '';
-export let commandline = '';
-export let hidden = false;
+interface Props {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  icon: any;
+  title?: string;
+  message?: string;
+  detail?: string;
+  commandline?: string;
+  hidden?: boolean;
+  class?: string;
+  style?: string;
+  'aria-label'?: string;
+  upperContent?: Snippet;
+  children?: Snippet;
+  onClick?: (text: string) => void;
+}
 
-let fontAwesomeIcon = false;
-let processed = false;
+let {
+  icon = undefined,
+  title = 'No title',
+  message = '',
+  detail = '',
+  commandline = '',
+  hidden = false,
+  class: className,
+  style,
+  'aria-label': ariaLabel,
+  upperContent,
+  children,
+  onClick = (text): void => {
+    dispatch('click', text);
+  },
+}: Props = $props();
+
+let fontAwesomeIcon = $state(false);
+let processed = $state(false);
 
 const dispatch = createEventDispatcher<{ click: string }>();
-export let onClick: (text: string) => void = text => {
-  dispatch('click', text);
-};
 
 onMount(() => {
   if (isFontAwesomeIcon(icon)) {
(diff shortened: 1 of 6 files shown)


packages/ui/src/lib/screen/EmptyScreen.svelte
```diff
@@ -36,26 +58,27 @@ function handleClick(): void {
   }
 }
 
-let copyTextDivElement: HTMLDivElement;
+let copyTextDivElement = $state<HTMLDivElement>();
 </script>
 
 <div
-  class="flex flex-row w-full h-full justify-center {$$props.class ?? ''}"
+  class="flex flex-row w-full h-full justify-center {className}"
```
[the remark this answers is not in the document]
gastoner: @deboer-tim probably relevant? since the className is undefined now => add default value or this change

packages/ui/src/lib/screen/EmptyScreen.svelte
```diff
@@ -73,9 +96,9 @@ let copyTextDivElement: HTMLDivElement;
           ><Fa class="h-5 w-5 cursor-pointer text-xl text-[var(--pd-button-primary-bg)]" icon={faPaste} /></Button>
       </div>
     {/if}
-    {#if $$slots}
+    {#if children}
       <div class="py-2">
-        <slot />
+        {@render children()}
```
gastoner, line 101: `{@render children?.()}` ?

packages/ui/src/lib/screen/EmptyScreen.svelte
```diff
@@ -1,26 +1,48 @@
...
+  message?: string;
+  detail?: string;
+  commandline?: string;
+  hidden?: boolean;
+  class?: string;
+  style?: string;
+  'aria-label'?: string;
+  upperContent?: Snippet;
+  children?: Snippet;
+  onClick?: (text: string) => void;
```
axel7083: onclick ? let's be consistent with how svelte do the naming ? (I don't like it, but maybe better to have a single way)

deboer-tim: I agree it should be consistent and use the Svelte 5 pattern... but this was the Svelte 5 name picked in previous PR and we're trying to avoid forced migration? 🤷🏻‍♂️ Not sure what to do in cases like this as PR could get rejected both ways.

packages/ui/src/lib/screen/EmptyScreen.svelte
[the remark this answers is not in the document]
deboer-tim: I'll test again, but Svelte 5 seems to have a new check and this now works.

packages/ui/src/lib/screen/EmptyScreen.svelte
axel7083: if the `onClick` is for backward compatiblity, let's keep it

packages/ui/src/lib/screen/EmptyScreen.svelte
deboer-tim: See `{#if children}` two lines above. We don't want the div either if no children, so no need for `?.`

packages/ui/src/lib/screen/EmptyScreen.svelte
benoitf: for backward compliance, we need to dispatch an event which is already the case

but yes, the callback should be named with the same logic than svelte --> onclick

benoitf (commented):
rename onClick to onclick

deboer-tim:
Renamed onClick to onclick to match the Svelte 5 pattern. (it's a 'breaking' change since onClick has been there since July [https://github.com/podman-desktop/podman-desktop/commit/43db9b0a3f2b0d836619f4483da9ddfac863df35#diff-9dada4e04e82aec5969e59d18efdc17d1afdac3d264cef8ef54b67dbe93a273a], but I agree this is better/consistent and unlikely to actually break clients)

axel7083 (approved):

deboer-tim:
Fixed test and rebased (1h ago). @gastoner can you give a quick review?

gastoner (approved):
Codewise LGTM

PR #12207, opened 2025-04-15
refactor: moved registering commands to common file

gastoner:
### What does this PR do?
Moves register commands to additional file 

### Screenshot / video of UI
N/A

Required for https://github.com/podman-desktop/podman-desktop/pull/11990

### How to test this PR?
Pipeline should be ✅ 

### DIFF
--- a/packages/main/src/plugin/commands-init.ts
+++ b/packages/main/src/plugin/commands-init.ts
@@ -1,0 +1,134 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/commands-init.ts
```diff
@@ -0,0 +1,134 @@
+/**********************************************************************
```
axel7083, line 1: I like the idea, and probably good for now, but I would argue that the commands should be registered by the class using it. 

E.g.

- `feedback` command, should probably be registered by `FeedbackHandler`
- `help` I don't know, kinda random
- `troubleshooting` by the navigation registry
- `pullImage` by the container registry... but the container registry is already a very big file... :/

I don't know, I would approve for now, as this is better that what we currently have, but might need to discuss architecture in the backend

packages/main/src/plugin/index.ts
```diff
@@ -95,7 +95,6 @@ import type { KubeContext } from '/@api/kubernetes-context.js';
 import type { ContextHealth } from '/@api/kubernetes-contexts-healths.js';
```
axel7083, line 95: No test to remove in the index.spec.ts?

axel7083 (approved):

benoitf (approved):
fix the typo reported before merging

gastoner:
@benoitf updated

Issue #12187, opened 2025-04-14
bug(ui): cannot render snippet using html synthax


axel7083:
### Bug description

Svelte added a validation step when rendering component in https://github.com/sveltejs/svelte/pull/15521

The problem is the following, you cannot render [Snippet](https://svelte.dev/docs/svelte/snippet) using html synthax, you need to use `@render`.

In our codebase, we have the StatusIcon which take `any` as argument. And we provide sometime a snippet, sometime a Component.

https://github.com/podman-desktop/podman-desktop/blob/7f77b5ccd4695865314715a4d0ffee0db0e8adc7/packages/ui/src/lib/statusIcon/StatusIcon.svelte#L10

https://github.com/podman-desktop/podman-desktop/blob/7f77b5ccd4695865314715a4d0ffee0db0e8adc7/packages/ui/src/lib/statusIcon/StatusIcon.svelte#L37-L38

Related PR failing
- https://github.com/podman-desktop/podman-desktop/pull/12180
- https://github.com/podman-desktop/extension-podman-quadlet/pull/493

### Operating system

Window 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. latest svelte version
2. try to render a snippet using `<Component/>` synthax

### Relevant log output

```shell

```

axel7083:
The problem is coming from trying to use `render` function with a snippet in a test

https://github.com/podman-desktop/podman-desktop/blob/85ad2b0bcc43e77e77eec35edce4fda8596ca3fe/packages/renderer/src/lib/statusbar/PinActions.spec.ts#L204

I commented on https://github.com/testing-library/svelte-testing-library/issues/284#issuecomment-2801148669

axel7083:
More discussion on related issues can be found in 
- https://github.com/podman-desktop/podman-desktop/pull/11930#discussion_r2035574390
- https://github.com/sveltejs/svelte/issues/9774

axel7083:
Upstream issue on svelte https://github.com/sveltejs/svelte/issues/15760

deboer-tim:
FWIW I had no idea it was possible (let alone already used by an extension!) to pass a Svelte v5 snippet to a v4 `<svelte:component>`. Sorry for breakage.

[...]

PR #12210, opened 2025-04-15
fix(StatusIcon): restrict icon type

axel7083:
### What does this PR do?

On the Quadlet Extension I provided a snippet as icon, but this is invalid, and not a supported use case: [bug(ui): cannot render snippet using html synthax](https://github.com/podman-desktop/podman-desktop/issues/12187)

To avoid repeating this error in the future, remove the any, and replace with two allowed type `string` and `Component`.

### How to test this PR?


axel7083:
Too many tests failing, need to come back later, closing for now

PR #12230, opened 2025-04-16
chore: shared kubernetes status column

deboer-tim:
### What does this PR do?

Creates a sharable Status column that works for all KubernetesObjectUIs, and uses it for Pods. Follow-up PR would update all the other Kubernetes *List.svelte files to use column/Status.svelte and delete all (now unused) *ColumnStatus files.

I've checked and cannot find any way to reuse the logic directly from KubernetesIcon (you can't set a property on a component, not create it in a way to pass as a component to StatusIcon), so the mapping from kind to icon is moved to kube-icon-utils.ts and then reused from there in column/Status.

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

Part of #11035.

### How to test this PR?

Confirm no change to status icons on Kubernetes > Pods page.

- [x] Tests are covering the bug fix or the new feature

PR #12232, opened 2025-04-16
fix(icons): update Props interfaces

axel7083:
### What does this PR do?

When migrating the `CronJobIcon.svelte` and `JobIcon.svelte` component to Svelte5, the props where define with mandatory properties (size & solid) this was an error, as those value have default value provided, so they should be optional.

This is important, as those component have the wrong ComponentProps type then.

Required for
- https://github.com/podman-desktop/podman-desktop/pull/12210
- https://github.com/podman-desktop/podman-desktop/pull/12230

### How to test this PR?

- Pipeline should be :heavy_check_mark:

### DIFF
--- a/packages/renderer/src/lib/images/CronJobIcon.svelte
+++ b/packages/renderer/src/lib/images/CronJobIcon.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 interface Props {
-  size: string;
-  solid: boolean;
+  size?: string;
+  solid?: boolean;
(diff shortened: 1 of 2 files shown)


gastoner (commented):
Shouldnt be just `size` optional? 
Or please add (probably explicit `false` to solid), now I would say that solid is `undefined` although in the conditions below is the outcome the same

gastoner (approved):

Issue #12206, opened 2025-04-15
Docs: Add troubleshooting section for Podman machine stuck in starting


Firewall:
### Is your enhancement related to a problem? Please describe

In some cases the podman machine gets stuck in a starting state. 

Provide documentation to help users fix this.

### Describe the solution you'd like

Include 1 new troubleshooting step:
1. Podman machine sometimes gets stuck on starting and provide the steps to get out of that situation.
2. Steps:
3. Completely stop all podman related processes
4. Stop and remove the podman machine podman machine stop podman machine rm
5. Reset the machine podman machine reset
6. Create a new machine using the Podman desktop UI.

rujutashinde:
fixed with #12247

PR #12247, opened 2025-04-17
docs(website): added a troubleshooting section

shipsing:
### What does this PR do?
- Adds a troubleshooting section: Podman machine gets stuck in the starting phase
https://github.com/podman-desktop/podman-desktop/issues/12206

### How to test this PR?

### DIFF
--- a/website/docs/troubleshooting/troubleshooting-podman.md
+++ b/website/docs/troubleshooting/troubleshooting-podman.md
@@ -60,7 +60,7 @@ Podman Desktop might fail creating a Podman machine.
 
 #### Issue
 
-Podman Desktop might fail starting a Podman machine.
+Podman Desktop might fail starting a Podman machine. On the **Settings > Resources** page, the machine does not switch to the running phase for quite some time.
 
 #### Workaround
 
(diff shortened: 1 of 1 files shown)


website/docs/troubleshooting/troubleshooting-podman.md
shipsing, line 1: Doesn't the [previous section](https://podman-desktop.io/docs/troubleshooting/troubleshooting-podman#podman-desktop-fails-starting-a-podman-machine) also discuss the same issue? If so, we should keep only one section for the user's reference.

cdrage: Actually yes. I agree with you that it should be the same section.

Maybe we update that linked section to include information about using `podman machine reset`?

cdrage (changes requested):
See comment (I agree with you!)

Firewall (approved):
Thanks. Totally makes sense to improve the existing section

shipsing:
@cdrage, Could you please verify the PR and confirm? Thanks

cdrage (approved):

cdrage:
> @cdrage, Could you please verify the PR and confirm? Thanks

Confirmed and thanks for fixing it up! LGTM!

PR #12248, opened 2025-04-17
chore(test): new get extension version method

amisskii:

### What does this PR do?
Adds new method to retrieve extension versions from extension cards.

### DIFF
--- a/packages/renderer/src/lib/extensions/InstalledExtensionCardRight.svelte
+++ b/packages/renderer/src/lib/extensions/InstalledExtensionCardRight.svelte
@@ -18,7 +18,7 @@ export let extension: CombinedExtensionInfoUI;
     <div>
       {extension.removable ? '' : 'Podman Desktop built-in extension'}
     </div>
-    <div>
+    <div aria-label="Version">
       {#if extension.version}
         v{extension.version}
       {/if}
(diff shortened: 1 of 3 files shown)


benoitf (approved):
LGTM

packages/renderer/src/lib/extensions/InstalledExtensionCardRight.svelte
```diff
@@ -18,7 +18,7 @@ export let extension: CombinedExtensionInfoUI;
     <div>
       {extension.removable ? '' : 'Podman Desktop built-in extension'}
     </div>
-    <div>
+    <div aria-label="Version">
```
odockal, line 21: add missing component test.

tests/playwright/src/model/pages/extensions-page.ts
```diff
@@ -108,4 +108,10 @@ export class ExtensionsPage {
     const extension = this.content.getByRole('region', { name: label });
     return (await extension.count()) > 0;
   }
+
+  public async getInstalledExtensionVersion(name: string, label: string): Promise<string> {
+    const extensionCard = await this.getInstalledExtension(name, label);
+    const version = extensionCard.rightActions.getByLabel('Version');
+    return (await version.textContent()) ?? '';
```
odockal: You should change the return value to `Promise<string | undefined>`.
Then you should check if the element count is greater than 0 and then return `innerText` if it is non-empty, `undefined` otherwise.

Or you can simply define the locator in ExtensionCard and assert using `toHaveText`. etc.

odockal (changes requested):

cdrage (approved):
Code-wise LGTM after @odockal 's suggestions are added.

odockal (approved):
LGTM

Issue #7397, opened 2024-05-31
Ephemeral data management


deboer-tim:
### Is your enhancement related to a problem? Please describe

We have several places in the code where we store ephemeral data like open user terminals, or output from tasks. Currently, each place that requires it creates and stores its own data objects and comes up with its own way to clear the data when it is no longer needed. This takes a larger development effort each time we need this pattern somewhere, and is error prone, both directly via new code and chances of data never getting cleaned up.

### Describe the solution you'd like

Create a pattern or API for storing data. It must support typed data objects, and be easy to tie cleanup to the task framework or other lifecycles - e.g. when a task is deleted have the data associated with it (like console output) automatically get cleaned up as well.

jeffmaury:
Still valid

PR #12261, opened 2025-04-19
fix: make image build cancellable

### LINKED ISSUE #10080: No way to cancel build, start another one

dgolovin:
### What does this PR do?

0. Supports cancellation and restart the build
1. Preserves build history (form field values and terminal output)
2. Restores build status when called from task manager (form field values and terminal output)
3. Removes stored status for finished builds when task is removed form task manager
4. Supports all kinds of transition to the build image page and restores build state

### Screenshot / video of UI

https://github.com/user-attachments/assets/91bfa398-8ebf-499d-92be-de352f96081f

Fix #10080

### How to test this PR?

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -178,6 +178,9 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
         <Route path="/images/existing-image-create-container" breadcrumb="Select image" >
           <CreateContainerFromExistingImage />
         </Route>
+        <Route path="/images/build" breadcrumb="Build an Image" let:meta>
+          <BuildImageFromContainerfile taskId={+meta.query.taskId}/>
+        </Route>
         <Route path="/images/:id/:engineId" breadcrumb="Images" let:meta navigationHint="root">
           <ImagesList searchTerm={meta.params.id} imageEngineId={meta.params.engineId} />
         </Route>
@@ -200,9 +203,6 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
             imageID={meta.params.id}
             engineId={decodeURI(meta.params.engineId)}
             base64RepoTag={meta.params.base64RepoTag} />
-        </Route>
-        <Route path="/images/build" breadcrumb="Build an Image">
-          <BuildImageFromContainerfile />
         </Route>
         <Route path="/images/pull" breadcrumb="Pull an Image">
           <PullImage />
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -1,68 +1,85 @@
 <script lang="ts">
-/* eslint-disable import/no-duplicates */
+/* eslint-disable no-useless-escape */
 // https://github.com/import-js/eslint-plugin-import/issues/1479
 import { faCube, faMinusCircle, faPlusCircle } from '@fortawesome/free-solid-svg-icons';
-import type { OpenDialogOptions } from '@podman-desktop/api';
+import { type OpenDialogOptions } from '@podman-desktop/api';
 import { Button, Input } from '@podman-desktop/ui-svelte';
-import type { Terminal } from '@xterm/xterm';
-import { onDestroy, onMount } from 'svelte';
-import { get } from 'svelte/store';
+import { onDestroy } from 'svelte';
+import { get, type Unsubscriber } from 'svelte/store';
 
 import ContainerConnectionDropdown from '/@/lib/forms/ContainerConnectionDropdown.svelte';
 import FileInput from '/@/lib/ui/FileInput.svelte';
 import { handleNavigation } from '/@/navigation';
-import { type BuildImageInfo, buildImagesInfo } from '/@/stores/build-images';
+import {
+  type BuildImageInfo,
+  buildImagesInfo,
+  cloneBuildImageInfo,
+  createDefaultBuildImageInfo,
+  getNextTaskId,
+  lastUpdatedTaskId,
+} from '/@/stores/build-images';
 import { NavigationPage } from '/@api/navigation-page';
-/* eslint-enable import/no-duplicates */
-import type { ProviderContainerConnectionInfo, ProviderInfo } from '/@api/provider-info';
+import type { ProviderInfo } from '/@api/provider-info';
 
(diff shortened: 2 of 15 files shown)


packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -30,39 +30,50 @@ import {
...
+$: hasInvalidFields =
+  !buildImageInfo.containerFilePath ||
+  !buildImageInfo.containerBuildContextDirectory ||
+  (platforms.length > 1 && !buildImageInfo.containerImageName) ||
+  platforms.length === 0 ||
+  !buildImageInfo.selectedProvider;
+
+$: if (taskId && taskId !== buildImageInfo.taskId) {
+  if (buildImageInfo?.buildImageKey) {
+    //} && !buildImageInfo.buildFinished) {
```
feloy: you can remove this commented code

feloy (approved):
tested on mac, can come back to a previous image build screen after a cancel, with correct info in the form

packages/renderer/src/App.svelte
```diff
@@ -178,6 +178,12 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
         <Route path="/images/existing-image-create-container" breadcrumb="Select image" >
           <CreateContainerFromExistingImage />
         </Route>
+        <Route path="/images/build" breadcrumb="Build an Image">
+          <BuildImageFromContainerfile />
+        </Route>
+        <Route path="/operations/images/build/:taskId" breadcrumb="Build an Image" let:meta>
+          <BuildImageFromContainerfile taskId={+meta.params.taskId}/>
```
axel7083: question: not sure to understand the `+` prefixing the `meta.params.taskId`?

packages/renderer/src/App.svelte
```diff
@@ -178,6 +178,12 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
         <Route path="/images/existing-image-create-container" breadcrumb="Select image" >
           <CreateContainerFromExistingImage />
         </Route>
+        <Route path="/images/build" breadcrumb="Build an Image">
+          <BuildImageFromContainerfile />
+        </Route>
+        <Route path="/operations/images/build/:taskId" breadcrumb="Build an Image" let:meta>
```
axel7083: remark (blocking): I am not sure to understand why need two routes ? we could use instead a query parameter, so we have a single root, and have the taskId `string | undefined`.

packages/renderer/src/stores/build-images.ts
```diff
@@ -16,13 +16,51 @@
...
+    containerFilePath: '',
+    containerBuildContextDirectory: '',
+    containerBuildPlatform: '',
+    buildArgs: [{ key: '', value: '' }],
+  };
 }
 
-// current build key
-export const buildImagesInfo: Writable<BuildImageInfo> = writable();
+export const buildImagesInfo: Writable<Map<number, BuildImageInfo>> = writable(new Map());
```
axel7083, line 80: remark: I would rather have the store in a dedicated files like `build-image-info-store` or something with an explicit name, otherwise with the content of the file, and the store being at the bottom of the file, it get lost

packages/renderer/src/App.svelte
dgolovin: It converts string to number.

https://stackoverflow.com/questions/12862624/whats-the-fastest-way-to-convert-string-to-number-in-javascript

packages/renderer/src/App.svelte
dgolovin: resolved

packages/renderer/src/stores/build-images.ts
dgolovin: The file was not introduced by this fix, it should be follow up issue probably.

packages/renderer/src/lib/image/ImagesList.svelte
```diff
@@ -165,7 +165,7 @@ onDestroy(() => {
 });
 
 function gotoBuildImage(): void {
-  router.goto('/images/build');
+  router.goto('/images/build?taskId=');
```
axel7083: you should not need to specify the taskId parameter here, not having it makes it undefined

packages/renderer/src/stores/build-images.ts
axel7083, line 80: I don't know how future proof this is, I don't want to block this PR, because this is already how we do things in for the providers form etc.

But this is a memory leak, we never cleanup, we just store the data, if someone never restart podman desktop the build logs are stored forever, this is a problem..

 maybe we need some discussion and maybe think again about https://github.com/podman-desktop/podman-desktop/issues/7397

cc @benoitf, I am interested about your opinion on this matter

dgolovin: This should be stored until it is accessible from Task Manager, so if someone decide to check what happened it specific build it would always get build image form restored to the saved state. 

I was going to implement clean up it up when Task removed from Task Manager and therefore the stored data for the build is no longer accessible from Podman Desktop UI.

axel7083: > I was going to implement clean up it up when Task removed from Task Manager and therefore the stored data for the build is no longer accessible from Podman Desktop UI.

yeah, but my point would be, maybe this should not be stored in the frontend, but in the backend, I don't know what is the best practice here.

If the data is stored in the backend, we could add a `cleanup / dispose` event to  the tasks, so when the tasks is cleaned / removed, we can removed attached data.

ScrewTSW (approved):
Tested on Linux, seems to be working fine

ScrewTSW:
tested on Mac M2 Pro and Windows 10 as well, AOK

axel7083 (approved):

Issue #12107, opened 2025-04-09
Update the createPodmanMachineFromCli utility function for e2e tests to consider environment configuration


odockal:
### Is your enhancement related to a problem? Please describe

Right now  the code creates a rootful machine without considering the test. env. configuration:
```
    try {
      // eslint-disable-next-line sonarjs/no-os-command-from-path
      execSync('podman machine init --rootful');
    } catch (error) {
      if (error instanceof Error && error.message.includes('VM already exists')) {
        console.log(`Podman machine already exists, skipping creation.`);
      }
    }
```

### Describe the solution you'd like

We need to change the code to be aligned with the test. env. it is being run in. Ie. running rootless windows 10 configuration would require to recreate a rootless machine. Something like:
```
    try {
      // eslint-disable-next-line sonarjs/no-os-command-from-path
      const mode = process.env.PODMAN_ROOTFUL === '0' ? '' : '--rootful';
      execSync(`podman machine init ${mode}`);
    } catch (error) {
      if (error instanceof Error && error.message.includes('VM already exists')) {
        console.log(`Podman machine already exists, skipping creation.`);
      }
    }
```

PR #12322, opened 2025-04-24
chore(test): try to create and start machine before waiting

cbr7:
### What does this PR do?
Tries to create/start podman machine before waiting for it.

### DIFF
--- a/tests/playwright/src/utility/wait.ts
+++ b/tests/playwright/src/utility/wait.ts
@@ -20,6 +20,7 @@ import type { Page } from '@playwright/test';
 import test, { expect as playExpect } from '@playwright/test';
 
 import { NavigationBar } from '../model/workbench/navigation';
+import { createPodmanMachineFromCLI } from './operations';
 
 export async function wait(
   waitFunction: () => Promise<boolean>,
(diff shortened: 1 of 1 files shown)


tests/playwright/src/utility/wait.ts
```diff
@@ -116,6 +117,8 @@ export async function delay(ms: number): Promise<void> {
 
 export async function waitForPodmanMachineStartup(page: Page, timeoutOut = 30_000): Promise<void> {
   return test.step('Wait for Podman machine to be running', async () => {
+    await createPodmanMachineFromCLI();
```
odockal, line 120: This function creates rootful machine by default. It does not reflect any of the configurational parameters like, rootless, user mode networking, provider, etc. Feel free to include that, we have issue for it: https://github.com/podman-desktop/podman-desktop/issues/12107.

odockal (dismissed):

tests/playwright/src/utility/wait.ts
cbr7: @odockal I mean I agree, but does not seem to be in scope of this PR, that is a separate enhancement, this PR addresses the issue that currently on cicd, especially on macos pipelines the podman machine seems to be stopped/not started.

I've assigned myself to that issue and put it for next sprint, since there is no real urgency with it.

ScrewTSW (approved):

PR #12342, opened 2025-04-28
chore(test): don't create machine on linux

cbr7:
### What does this PR do?
Test framework will no longer try to create podman machine on linux


odockal (approved):
LGTm.

Issue #12035, opened 2025-04-04
Crashing when updating on Windows


pboyadjian:
### Bug description

The application crashes during an internal update.

### Operating system

Windows 11

### Installation Method

Installer from website/GitHub releases

### Version

1.17.0

### Steps to reproduce

![Image](https://github.com/user-attachments/assets/f3ec89b3-f650-44ac-9f50-e55771bcd4bc)

### Relevant log output

```shell

```

axel7083:
## Description

Updating from Podman Desktop 1.17.2 to 1.18.0 lead to a crash on my system.

https://github.com/user-attachments/assets/d2ad16f9-5469-4970-8ff2-eb1bce5ee1f1

## Logs

```
Installing Podman Desktop startup file at C:\Users\axels\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup\podman-desktop.vbs using C:\Users\axels\AppData\local\Programs\podman-desktop\Podman Desktop.exe location.
Checking for update
Starting http server to handle webviews on port 44000
System ready. Loading extensions...
PluginSystem: received dom-ready event from the UI
Delayed startup, flushing 8 events
Error while fetching API groups: FetchError: request to https://127.0.0.1:58265/apis failed, reason: connect ECONNREFUSED 127.0.0.1:58265
Error while fetching API groups: FetchError: request to https://127.0.0.1:58265/apis failed, reason: connect ECONNREFUSED 127.0.0.1:58265
Fetched https://registry.podman-desktop.io/api/extensions.json in 286.22979999999984ms
Extension (podman-desktop.compose) not activated because it is disabled
Extension (podman-desktop.docker) not activated because it is disabled
Extension (podman-desktop.kind) not activated because it is disabled
Extension (podman-desktop.kube-context) not activated because it is disabled
Extension (podman-desktop.kubectl-cli) not activated because it is disabled
Extension (podman-desktop.lima) not activated because it is disabled
Extension (podman-desktop.podman) not activated because it is disabled
Extension (podman-desktop.registries) not activated because it is disabled
Extension (podman-desktop.layers-explorer) not activated because it is disabled
Extension (podman-desktop.minikube) not activated because it is disabled
PluginSystem: initialization done.
Skipping update for extension podman-desktop.layers-explorer because installed version 0.2.0 is greater or equal to latest available version 0.2.0
Skipping update for extension podman-desktop.minikube because installed version 0.4.0 is greater or equal to latest available version 0.4.0
[...]

Issue #10725, opened 2025-01-20
Crash on Update


GamerPanda99:
### Bug description

Whenever the notification for an update pops up, and Iclick the update button the application hangs an crashes.

### Operating system

Windows 11 Pro - x64

### Installation Method

Installer from website/GitHub releases

### Version

1.15.0

### Steps to reproduce

- Update notification pops up
- Click on Update
- Application hangs
- application crashes

### Relevant log output

```shell

```

### Additional context

**Enabled Extensions**
- podman-desktop.compose
- podman-desktop.docker
- podman-desktop.kind
- podman-desktop.kube-context
- podman-desktop.kubectl-cli
- podman-desktop.podman
- podman-desktop.registries

benoitf:
hello @GamerPanda99 was it with the version 1.15.0 ?

or what's your current version of Podman Desktop (you can find it in the status bar)

GamerPanda99:
Hi @benoitf It ahappened with the last two versions, so when updating from 1.14.0 updating to 1.14.2, and updating from 1.14.2 updating to 1.15.0

benoitf:
thank @GamerPanda99 for the report

we have tests checking the update on our side and they're working, so to be able to investigate on how to fix/reproduce the issue, is it for your for example to install again 1.14.2 and when updating to 1.15.0, before clicking on the update, bringing the devtools console. The shortcut on Windows Alt+Shift+I

![Image](https://github.com/user-attachments/assets/9ec7d09e-b221-4eb5-a6c9-0a3e5e73a128)

and then if you could send the log being displayed on this console (hopefully before a crash)

thanks

odockal:
@benoitf @GamerPanda99 I managed to reproduce the issue on my machine (Win10Pro). The thing is that the update crash can still happen for PD 1.14.0 (and above probably) if some (unknown) steps are taken. It might that the cause is in the Podman Desktop configuration folder, (`~/.local/share/containers/podman-desktop`). I do not know how yet. 

The folder contains extensions data, settings, etc. If I remove the folder, and start up PD (1.14.0) again, the update starts to work again. I am still investigating, though...

@GamerPanda99 Do you have any other installed extensions, like minikube for example? Or just these built-in ones?

odockal:
I managed to narrow down the issue... yet, not into a very clear state.
I can somehow reproduce, the condition that is same on windows 10 atm, is the **minikube 0.4.0** extension is installed.
[...]

PR #12353, opened 2025-04-29
fix(electron-updater): disable differential download

axel7083:
### What does this PR do?

`electron-update` package uses differential download for the update. This is a problem because it consume a lot of memory to make a difference between the previous blockmap and the new one.

**We are pretty short in memory** today with Podman Desktop. This is not a _nice_ solution, as this may induce slight bigger download time (we download the full file instead of partial) but **this fixes the memory issue**.

Being able to download the update without crash in a longer time is better than nothing.

Should fixes the following issues
- https://github.com/podman-desktop/podman-desktop/issues/12035
- https://github.com/podman-desktop/podman-desktop/issues/10725

### How to test this PR?

1. checkout https://github.com/podman-desktop/podman-desktop/tree/1.17.x
2. merge this PR changes in the 1.17.x branch
3. `pnpm install`
4. `pnpm run compile:current`
5. execute `dist/*-unpacked/Podman Desktop.exe`
6. assert podman desktop starts
7. assert an update is available
8. start the update workflow

### DIFF
--- a/packages/main/src/plugin/updater.spec.ts
+++ b/packages/main/src/plugin/updater.spec.ts
@@ -31,8 +31,9 @@ import type { MessageBox } from '/@/plugin/message-box.js';
 import type { StatusBarRegistry } from '/@/plugin/statusbar/statusbar-registry.js';
 import type { Task } from '/@/plugin/tasks/tasks.js';
 import { Disposable } from '/@/plugin/types/disposable.js';
-import { Updater } from '/@/plugin/updater.js';
+import { DIFFERENTIAL_DOWNLOAD_PLATFORM_SUPPORT, Updater } from '/@/plugin/updater.js';
 import * as util from '/@/util.js';
+import { isLinux, isMac, isWindows } from '/@/util.js';
 
 import type { ApiSenderType } from './api.js';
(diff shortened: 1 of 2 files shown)


odockal (approved):
Built from main, changed version in package json to 1.17.0, applied the changes, build+compiled, tested, works on Windows 10 while without patch it crashes. I also think that there is missing very annoying glitch that blocked all UI while the blockmap was computed.

benoitf:
I would enable it only on Windows ( I don't recall errors on macOS)
And a hidden flag to change the behaviour in case it has unexpected drawbacks

amisskii (approved):
Tested on mac. It works fine. LGTM

axel7083:
> I would enable it only on Windows ( I don't recall errors on macOS) And a hidden flag to change the behaviour in case it has unexpected drawbacks

@benoitf I updated the code and added a new configuration property `preferences.update.disableDifferentialDownload`:

- defaulting to `true` on window
- false on other platforms

benoitf (approved):

axel7083:
@Mergifyio backport 1.18.x

PR #12357, opened 2025-04-29
fix: introduce fontId property to ProviderImages interface

### LINKED ISSUE #12190: Support provider icons and use in the statusbar

We allow extensions to register vector icons via woff files and use these in various places, e.g. when adding their own statusbar items or customizing container or image icons.

However, when they register a provider the only option is to provide png icons. These icons are great for most use, but do not match our design / style when we add providers to the status bar. As a result, the provider has to choose between a correct color icon or one that balances the look between color and monochrome.

Allow providers to reference an icon in addition to the pngs. Built-in extensions like Kind should register an icon.

When provided, these should be used in the status bar provider widgets.

_No response_

### Additional context

_No response_

dgolovin:
### What does this PR do?

fix: introduce fontId property to ProviderImages interface
`iconClass` property references the icon defined
by extension in package.json using `icons` extension point.

```
...
    "icons": {
      "iconId": {
        "description": "Extension defined icon",
        "default": {
          "fontPath": "icon-file.woff2",
          "fontCharacter": "\\e900"
        }
      }
    }
...
```

### Screenshot / video of UI

Before fix

image

After fix

image

See PR for sandbox extension here https://github.com/redhat-developer/podman-desktop-sandbox-ext/pull/511.

Fix #12190.

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -661,8 +661,16 @@ declare module '@podman-desktop/api' {
 
   export type ProviderLinks = Link;
 
+  /**
+   * An interface to provide information about provider's images
+   * to use on Dashboard, in Resource Settings and Status Bar.
+   */
   export interface ProviderImages {
-    icon?: string | { light: string; dark: string };
+    /**
+     * fontId is a monochrome icon ID to use in Status Bar. The icon should be defined using
+     * `icons` extension point (see details here https://podman-desktop.io/docs/extensions/developing/adding-icons).
+     */
+    icon?: string | { light: string; dark: string; fontId?: string };
     logo?: string | { light: string; dark: string };
   }
 
--- a/packages/renderer/src/lib/statusbar/ProviderButton.spec.ts
+++ b/packages/renderer/src/lib/statusbar/ProviderButton.spec.ts
@@ -18,15 +18,14 @@
 import '@testing-library/jest-dom/vitest';
 
 import type { ProviderStatus } from '@podman-desktop/api';
-import { render } from '@testing-library/svelte';
+import { render, screen } from '@testing-library/svelte';
+import { tick } from 'svelte';
 import { beforeEach, expect, test, vi } from 'vitest';
 
-import IconImage from '/@/lib/appearance/IconImage.svelte';
 import ProviderButton from '/@/lib/statusbar/ProviderButton.svelte';
 import ProviderButtonStatus from '/@/lib/statusbar/ProviderButtonStatus.svelte';
 import type { ProviderInfo } from '/@api/provider-info';
 
-vi.mock(import('/@/lib/appearance/IconImage.svelte'));
 vi.mock(import('/@/lib/statusbar/ProviderButtonStatus.svelte'));
 
 const PROVIDER_MOCK = {
@@ -44,7 +43,7 @@ beforeEach(() => {
   vi.resetAllMocks();
 });
 
-test('class props should be propagated to button', async () => {
+test('class props should be propagated to button', () => {
   const { getByRole } = render(ProviderButton, {
     provider: PROVIDER_MOCK,
     onclick: vi.fn(),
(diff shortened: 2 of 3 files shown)


gastoner (approved):
Codewise LGTM

axel7083 (changes requested):
Nice enhancement !

Just the naming (`iconClass`) and the format `${<icon>}` is a bit weird to me..

Let's break it down: we want to _attach_ to our providers image an icon defined in the `contributes#icons`.

Since we try to target an icon, maybe we should use `iconId`. Looking at the VSCode documentation (https://code.visualstudio.com/api/references/contribution-points#contributes.icons) this seems pretty close of how they define it.

We could then simply use the icon id that we use (without using `"${<icon>}"`

We should also add some documentation above the new optional `iconId` property, and probably a validation when loading the extension, we should check that the iconId exists otherwise `console.warn`. The documentation should point to https://podman-desktop.io/docs/extensions/developing/adding-icons

packages/extension-api/src/extension-api.d.ts
```diff
@@ -664,6 +664,7 @@ declare module '@podman-desktop/api' {
   export interface ProviderImages {
     icon?: string | { light: string; dark: string };
     logo?: string | { light: string; dark: string };
+    iconClass?: string;
```
axel7083: I don't see the new changes applied ? issue with the rebase / force push ?

dgolovin: should  be visible now

dgolovin:
Added documentation with link.

axel7083 (approved):
LGTM 🚀 , thanks for all the changes and the documentation !

packages/extension-api/src/extension-api.d.ts
```diff
@@ -661,9 +661,20 @@ declare module '@podman-desktop/api' {
...
   export interface ProviderImages {
     icon?: string | { light: string; dark: string };
     logo?: string | { light: string; dark: string };
+    /**
+     * A monochrome icon ID to use in Status Bar. The icon should be defined using
+     * `icons` extension point (see details here https://podman-desktop.io/docs/extensions/developing/adding-icons).
+     *
+     * If it is undefined then the image defined by icon property is used instead.
+     */
+    iconId?: string;
```
benoitf: what's the difference between icon and iconId

I'm not sure a newcomer would understand

benoitf: also could the id be used directly in `icon` field ?

we should be able to know if the id exists or not

dgolovin: The difference is one is an icon ID from icons contribution and another is relative path to an image file.

We can let to use image ID in image field, but that would further complicate the code IMO.

Icon now is defined as:

```
icon?: string | { light: string; dark: string };
```

Extensions are just using icon as a string when registering a provider instance.

For Sandbox Extension specifically I need to define a monochrome icon for the status bar and a color icon for the rest of the UI (dashboard and settings).

What do you have in mind for reusing icon property above to avoid adding new iconId property and to avoid possible compatibility issues?

deboer-tim: `icon` = full colour icon. `iconId` = woff (monochrome vector) icon. They would be used in different circumstances so we need both / they are independent. Agreed this is potentially confusing - but don't have a better suggestion atm.

benoitf: we do know what is provided

A custom font icon use the pattern `${my-icon}` like `${kind-icon}`

so if it's using this format it'll be searched as a font icon, if not using `${}` then it's a classic icon

benoitf: if it's to provide 3 items

Then it should be in the string | { light: string; dark: string };

Like light, dark, and then a 3rd one being the fontId

dgolovin: @benoitf do you mean this?
```
{
  light: string,
  dark: string, 
  id: string,
}
```

because if I just use `${iconId}` in place of icon, icon.light or icon.dark and recognize it is an id by `${}` presence, I would hit the same issue when in sandbox where I cannot have the same Red Hat logo for dark and light themes and mono icon for sandbox Provider status bar item.

benoitf: yes, the option `icon : string | {light: string; dark: string; fontId?: string}` feels more natural to me as either you choose a default icon or then you can specify a dedicated icon for light mode, dark mode or a font id to be used (also using the word font emphasis that the usage will be when we need to display text using a font)

rather than having both `icon` and `iconId` at the root level

benoitf (approved):
thanks for this last minute change

Issue #10993, opened 2025-02-05
Correct the definition of vale styles


shipsing:
### Is your enhancement related to a problem? Please describe

All the Red Hat styles for vale are defined [here](https://github.com/redhat-documentation/vale-at-red-hat/tree/main/.vale/styles/RedHat). Though most of the rules are useful for consistency purposes, we might need to modify some of the styles as per our project. For example, the rule defined [here](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/HeadingPunctuation.yml) says `Do not use end punctuation in headings`. If we include a question mark in the heading, this style will trigger a suggestion. As per Red Hat style guide, we usually do not include a heading in question format. But, this is not applicable to the Podman Desktop project. 

<img width="790" alt="Image" src="https://github.com/user-attachments/assets/001154d7-842b-42ae-8123-2f70eb1d3f3e" />

**The styles that might need customization** include:

- [HeadingPunctuation.yml](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/HeadingPunctuation.yml) -- The style suggests to not include .,?,!, in the heading title. If we want to use a heading title in question format (eg, What are Quadlets?), then we might need to tweak the style. 
- [Symbols.yml](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/Symbols.yml) -- The style suggests excluding some symbols (&, !, +) from use in documentation. we might tweak them as per our requirements, if needed.
- [DoNotUseTerms.yml](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/DoNotUseTerms.yml) -- The style suggests to not use certain terms in the documentation. we might tweak them as per our requirements, if needed.

**The styles that can be excluded**:
- [ReleaseNotes.yml](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/ReleaseNotes.yml) -- The style is specifically designed for OpenShift RNs.
- [GitLinks.yml](https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/GitLinks.yml) -- The style suggests to include link to any GitHub repo only if it is approved by the legal team (not applicable for PD) 

[...]

PR #12386, opened 2025-05-05
chore: remove vale tooling

benoitf:
### What does this PR do?
remove vale usage
each time we change the documentation, we've tons of comments about invalid links
like:

```
[vale] reported by reviewdog 🐶
[PodmanDesktop.Links] Consider starting the link with https:// or /, rather than ](img/creating-playground.png)

Raw Output:
{"message": "[PodmanDesktop.Links] Consider starting the link with https:// or /, rather than ](img/creating-playground.png)", "location": {"path": "website/docs/ai-lab/create-playground.md", "range": {"start": {"line": 26, "column": 25}}}, "severity": "WARNING"}
```

it could be re-introduced later if https://github.com/podman-desktop/podman-desktop/issues/10993 is complete

### How to test this PR?


axel7083 (approved):
LGTM, thanks for the cleanup 👍

vancura:
I am sorry that you think Vale is noisy. I totally understand it, but it's a really good tool (with its quirks, like the fact that it's CLI-based and not an NPM package). As someone commented above, I think it would be better to tweak its settings rather than remove it altogether.

Can we please revisit the Vale issue later when we have the terminology and branding document? Vale would greatly help keep the wording consistent. There's also a non-zero chance another tool would come before, so let's see :)

benoitf:
it can be brought back with a smaller subset, but for example to detect some typo/terminology we do have Sourcery and Copilot

PR #12397, opened 2025-05-06
chore(e2e): fix skip on kind tests

### LINKED ISSUE #351: Kind tests are not being skipped when they are supposed to
After issue https://github.com/podman-desktop/e2e/issues/315 was merged, kind tests were supposed to be skipped under certain circumstances. But as of the latest runs: https://github.com/podman-desktop/e2e/actions/runs/14244946146 this is not true. The logic for the skip seems correct, so it's likely a problem while setting up the values for the variables used in the check.

danivilla9:
### What does this PR do?
Fixes https://github.com/podman-desktop/e2e/issues/351

https://github.com/podman-desktop/e2e/issues/351

### How to test this PR?
Run `pnpm run test:e2e:k8s` on a windows machine. Kubernetes tests should be skipped

### DIFF
--- a/tests/playwright/src/specs/deploy-to-kubernetes.spec.ts
+++ b/tests/playwright/src/specs/deploy-to-kubernetes.spec.ts
@@ -41,7 +41,7 @@ const CONTAINER_START_PARAMS: ContainerInteractiveParams = {
 const skipKindInstallation = process.env.SKIP_KIND_INSTALL === 'true';
 const providerTypeGHA = process.env.KIND_PROVIDER_GHA ?? '';
 
-test.skip(!canRunKindTests, `This test can't run on a windows rootless machine`);
+test.skip(!canRunKindTests(), `This test can't run on a windows rootless machine`);
 
 test.beforeAll(async ({ runner, welcomePage, page, navigationBar }) => {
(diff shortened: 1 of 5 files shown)


danivilla9:
~I've been debugging this skip for a while now, and I'm still unsure why the original implementation does not work. There are several other examples on the testing suite that pass a function to the skip condition, and they supposedly work fine. I must be missing something, but if not invoked (adding parentheses), in my experience the test will never be skipped. The values used in the function are getting correctly evaluated, but it's like if the skip method sees having a function in the parameter as "true" independently of what it would output.~

With this setup I confirmed locally that the tests are skipped when expected to.

EDIT: No other tests use the implementation explained in the comment, I saw something that wasn't there.

odockal (approved):
LGTM

ScrewTSW:
@danivilla9 now it passed

@odockal although, I see quite a few of the failures https://github.com/podman-desktop/podman-desktop/actions/runs/14860210842/job/41729732580?pr=12397#step:10:1827
It seems like the tests are quitting and re-launching Podman Desktop (???)

cbr7:
@ScrewTSW every test suite is supposed to launch Podman Desktop and at the end close it.

PR #12398, opened 2025-05-06
chore(npm-registry): configure provenance for published packages

axel7083:
### What does this PR do?

Some of our public packages do not provide the [provenance](https://docs.npmjs.com/generating-provenance-statements) and the repository. (This seems to be pretty common ref https://github.com/search?q=path%3A.github+pnpm+--provenance&type=code)

> ℹ️ The provenance is a nice banner here is an example:
> image

⚠️ I had to add the `write` permission, this is a requirement from the npm tooling (https://docs.npmjs.com/generating-provenance-statements#prerequisites)

#### `@podman-desktop/podman-extension-api`[^1]

[^1]: https://www.npmjs.com/package/@podman-desktop/podman-extension-api

image

#### `@podman-desktop/ui-svelte`[^2]

[^2]: https://www.npmjs.com/package/@podman-desktop/ui-svelte

image

### How to test this PR?

### DIFF
--- a/.github/workflows/next-build.yaml
+++ b/.github/workflows/next-build.yaml
@@ -175,6 +175,8 @@ jobs:
     needs: [tag, release]
     name: Publish
     runs-on: ubuntu-24.04
+    permissions:
+      id-token: write
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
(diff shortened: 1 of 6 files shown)


benoitf (commented):
if we add provenance in the package.json, do we need to provide that in the command line ? or same opposite question: if we provide that in the command line, do we need to add that in the package.json file

axel7083:
> if we add provenance in the package.json, do we need to provide that in the command line ? or same opposite question: if we provide that in the command line, do we need to add that in the package.json file

Technically only one is needed, but same as `--access`, we have it in many places, so I thought that we could just add it everywhere, for consistency.

The `repository` is a **hard** requirements[^1], I can make a dedicated PR, if this is too much changes.

[^1]: https://docs.npmjs.com/generating-provenance-statements#prerequisites

benoitf (approved):

PR #12407, opened 2025-05-07
chore(test): accept degraded state as valid

cbr7:
### What does this PR do?
attempts to fix the pod degraded state in e2e test flakyness


axel7083 (approved):
LGTM

PR #12411, opened 2025-05-07
fix: display VM status

### LINKED ISSUE #93: Starting a second VM fails silently
Macadam now supports multiple VMs. You can create these within the extension and start/stop one at a time without issues.

If you try to start a second one when one is already running, nothing happens - the UI flashes (to 'starting'?) and the VM is stopped again. You can see a macadam error in the logs, but it isn't reported in the UI.

Macadam may eventually support starting multiple VMs at the same time so this particular error might not occur anymore, but I'd view this as 'command line errors should be visible in the UI' in case the VM fails to start for any reason.

feloy:

### What does this PR do?

When an error occurs during VM start/stop, the status should be displayed near the status of the VM, like it is displayed for container providers.

Fixes https://github.com/redhat-developer/podman-desktop-rhel-ext/issues/93

### How to test this PR?


cdrage (approved):
Tested against the RHEL VM extension and it worked well! Now displays the status / no longer silently fails.

I love how tests are 4x as much as the original code change, awesome tests!

LGTM

PR #12415, opened 2025-05-07
feat: added editing of contexts

### LINKED ISSUE #11283: Editing Kubernetes contexts

Users should be able to do simple editing of the Kubernetes contexts.

Implement the design from #11281. Users should be able to create new contexts or editing existing ones, changing the name or namespace but using the same cluster or user as contexts already in the config.

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds editing of name and namespace

There should be now 4 buttons (delete context, duplicate context, sometimes set as current) and new edit button which should allow to update name and namespace - choosing cluster and user will be another PR - https://github.com/podman-desktop/podman-desktop/pull/12445

### Screenshot / video of UI

[Screencast_20250507_140626.webm](https://github.com/user-attachments/assets/f5c57fb7-51ce-4e88-a594-530965354c5a) 

 just without duplicating

Part of #11283 
Required for https://github.com/podman-desktop/podman-desktop/pull/12445

### How to test this PR?
1. Create context (minikube/kind/etc...)
2. Go to preferences -> kubernes context
3. Try to edit context

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.ts
@@ -411,20 +411,22 @@ export class KubernetesClient {
     if (!originalContext) throw new Error('Context name was not found in kube config');
 
     const namespaceField = newContextNamespace !== '' ? { namespace: newContextNamespace } : {};
-    delete (originalContext as { namespace?: string }).namespace;
+
+    const editedContext = {
+      ...originalContext,
+      name: newContextName,
+      ...namespaceField,
+    };
+
+    if (newContextNamespace === '') {
+      delete editedContext.namespace;
+    }
 
     newConfig.loadFromOptions({
       clusters: this.kubeConfig.clusters,
       users: this.kubeConfig.users,
       currentContext: this.currentContextName === contextName ? newContextName : this.kubeConfig.currentContext,
-      contexts: [
-        {
-          ...originalContext,
-          name: newContextName,
-          ...namespaceField,
-        },
-        ...newContexts,
-      ],
+      contexts: [editedContext, ...newContexts],
     });
 
     await this.saveKubeConfig(newConfig);
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.ts
@@ -403,6 +403,35 @@ export class KubernetesClient {
     this.apiSender.send('kubernetes-context-update');
   }
 
+  async updateContext(contextName: string, newContextName: string, newContextNamespace: string): Promise<void> {
+    const newConfig = new KubeConfig();
+
+    const originalContext = this.kubeConfig.contexts.find(context => context.name === contextName);
+    const newContexts = this.kubeConfig.contexts.filter(ctx => ctx.name !== contextName);
+    if (!originalContext) return;
+
+    newConfig.loadFromOptions({
+      clusters: this.kubeConfig.clusters,
+      users: this.kubeConfig.users,
+      currentContext: this.currentContextName === contextName ? newContextName : this.kubeConfig.currentContext,
+      contexts: [
+        {
+          ...originalContext,
+          name: newContextName,
+          namespace: newContextNamespace,
+        },
+        ...newContexts,
+      ],
+    });
+
+    await this.saveKubeConfig(newConfig);
+    // the config is saved back only if saving the file succeeds
+    this.kubeConfig = newConfig;
+    // We send an update event here, even if another one will be sent after the file change is detected,
+    // because that one can get some time to be sent (as cluster connectivity will be tested)
+    this.apiSender.send('kubernetes-context-update');
+  }
+
   async deleteContext(contextName: string): Promise<Context[]> {
     const previousContexts = this.kubeConfig.contexts;
     const newContexts = this.kubeConfig.contexts.filter(ctx => ctx.name !== contextName);
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.ts
@@ -408,7 +408,10 @@ export class KubernetesClient {
 
     const originalContext = this.kubeConfig.contexts.find(context => context.name === contextName);
     const newContexts = this.kubeConfig.contexts.filter(ctx => ctx.name !== contextName);
-    if (!originalContext) return;
+    if (!originalContext) throw new Error('Context name was not found in kube config');
+
+    const namespaceField = newContextNamespace !== '' ? { namespace: newContextNamespace } : {};
+    delete (originalContext as { namespace?: string }).namespace;
 
     newConfig.loadFromOptions({
       clusters: this.kubeConfig.clusters,
@@ -418,7 +421,7 @@ export class KubernetesClient {
         {
           ...originalContext,
           name: newContextName,
-          namespace: newContextNamespace,
+          ...namespaceField,
         },
         ...newContexts,
       ],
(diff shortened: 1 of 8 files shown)


gastoner:
@deboer-tim @Firewall I've created first iteration for editing contexts can you PTAL?

Firewall:
Hey @gastoner, great job so far! Works really well so far. 

Couple of minor suggestions:
1. When I click edit on a context but I haven't changed anything. The `This context name already exists in /Users/mdemytte/.kube/config` shouldn't show up yet. This should only show when a change was made 
image

2. ~~Can we look into adding a tooltip to the button that explains the action? It might not always be clear for everyone what each action does.~~ Whilst taking the screenshot I realized there is a tooltip is just takes REALLY LONG to load. Maybe we can shorten that time? 
image

deboer-tim:
Tested this, nice work. 👍🏼

A few comments:
- I also ran into the same editing validation issue as Matt - name shouldn't conflict with itself. :)
- @Firewall the tooltip timeout is global, not specific to this PR. It is kind of long, if you want it fixed let's open a new issue.
- I noticed an odd connection issue: if you connect to another context and then edit or duplicate anything else, the connection is dropped. e.g. if you have contexts A B C, and A is current context, connect to B and then edit C: B is no longer connected. IMHO this shouldn't happen, maybe talk to Philippe to see if this is expected, improved by the experimental Kube work, or there's something we can do to avoid it.
- The edit dialog has 'Context Name', 'Contest Namespace'. This is redundant and already in the title, should just be 'Name' and 'Namespace'.
- I wouldn't say 'closes #11283' as we also need the ability to change the cluster or user. These two would just be dropdowns of cluster/user from other contexts though (i.e. they aren't 'editable', you can just pick from existing/known ones in other contexts). I'd do this separately though, i.e. change to 'Part of #11283' and then do second PR later to add those.

gastoner:
@Firewall @deboer-tim updated based on your comments & added tests

Firewall:
@gastoner Changing the name works as expected now! However...When I add a namespace the save button does not become active: 

image

gastoner:
@Firewall @deboer-tim updated based on your comments & added tests

> @gastoner Changing the name works as expected now! However...When I add a namespace the save button does not become active:
> 
> image

Can you try it again please? I've updated the PR

gastoner:
@Firewall @deboer-tim I've added editing users and clusters using dropdown. Just wanted to ask if it is ok, when opening the dropdown the dialog is always quite small to fit the content so you would need to scroll in here which is not ideal, but it is designed like that.

gastoner:
> @Firewall @deboer-tim I've added editing users and clusters using dropdown. Just wanted to ask if it is ok, when opening the dropdown the dialog is always quite small to fit the content so you would need to scroll in here which is not ideal, but it is designed like that.

Would add this change/commit in a separate PR

MariaLeonova (commented):
Looks great with the updates!

Firewall (approved):
Works perfectly now (non-code wise). Well done!

packages/main/src/plugin/kubernetes/kubernetes-client.ts
```diff
@@ -365,6 +365,72 @@ export class KubernetesClient {
     return kubeContexts;
   }
 
+  async duplicateContext(contextName: string): Promise<Context[]> {
```
feloy: This is only one big function with side effects (the contexts will be changed in place). 

I think it would be more testable and maintainable if we have 'pure' (without any side effect) functions as most as possible, for example a functio. which returns the new context, then another which adds this context to a list of contexts, then a last ones which updates the current contexts.

feloy (commented):
This PR is adding independent things: 
- editing name/namespace of an existing context
- duplicating a context

I think it would be better to have two separate PRs, one for each new feature

packages/main/src/plugin/kubernetes/kubernetes-client.ts
gastoner: Created https://github.com/podman-desktop/podman-desktop/pull/12463

gastoner:
@Firewall @feloy @MariaLeonova @deboer-tim Can you PTAL? I've rebased the PR with duplication  feature

MariaLeonova (commented):
Can you please specify what has changed compared to the initial screencast?

gastoner:
@MariaLeonova There should be now 4 buttons (delete context, duplicate context, sometimes set as current) and new edit button which should allow to update name and namespace - choosing cluster and user will be another PR - https://github.com/podman-desktop/podman-desktop/pull/12445

packages/main/src/plugin/kubernetes/kubernetes-client.ts
```diff
@@ -403,6 +403,35 @@ export class KubernetesClient {
     this.apiSender.send('kubernetes-context-update');
   }
 
+  async updateContext(contextName: string, newContextName: string, newContextNamespace: string): Promise<void> {
+    const newConfig = new KubeConfig();
+
+    const originalContext = this.kubeConfig.contexts.find(context => context.name === contextName);
+    const newContexts = this.kubeConfig.contexts.filter(ctx => ctx.name !== contextName);
+    if (!originalContext) return;
```
feloy: You may want to raise an error instead, and you can do it before computing `newContexts`

feloy (commented):
There is a specific case when we don't want to specify a namespace for a context.
With this UI, we can set the namespace as empty, but in the resulting config file, it will be set as `''`. It would be preferable to unset the namespace field instead:

What I get:
```
  - name: kind-kind-cluster-1-2
    context:
      cluster: kind-kind-cluster-1
      name: kind-kind-cluster-1-2
      user: kind-kind-cluster-1
      namespace: ''
```

What I expect:
```
  - name: kind-kind-cluster-1-2
    context:
      cluster: kind-kind-cluster-1
      name: kind-kind-cluster-1-2
      user: kind-kind-cluster-1
```

feloy:
> I noticed an odd connection issue: if you connect to another context and then edit or duplicate anything else, the connection is dropped. e.g. if you have contexts A B C, and A is current context, connect to B and then edit C: B is no longer connected. IMHO this shouldn't happen, maybe talk to Philippe to see if this is expected, improved by the experimental Kube work, or there's something we can do to avoid it.

This is expected in non-experimental mode (and would be quite complex to fix), and, according to my tests, this is fixed in experimental mode

gastoner:
@feloy updated with your suggestions

packages/main/src/plugin/kubernetes/kubernetes-client.ts
```diff
@@ -408,7 +408,10 @@ export class KubernetesClient {
 
     const originalContext = this.kubeConfig.contexts.find(context => context.name === contextName);
     const newContexts = this.kubeConfig.contexts.filter(ctx => ctx.name !== contextName);
-    if (!originalContext) return;
+    if (!originalContext) throw new Error('Context name was not found in kube config');
+
+    const namespaceField = newContextNamespace !== '' ? { namespace: newContextNamespace } : {};
+    delete (originalContext as { namespace?: string }).namespace;
```
feloy: I find this quite complex.

Why do you think about something like:

```
const editedContext = {
{
   ...originalContext,
   name: newContextName,
   ...namespaceField,
 };

if (editedContext.namespace === '') {
  delete editedContext.namespace;
}

newConfig.loadFromOptions({
[...]
      contexts: [
        ...newContexts,
        editedContext,
      ],
```

gastoner:
@feloy fixed, can you PTAL?

feloy (approved):
Thanks for the changes.
Great feature, thanks for your work on this!
LGTM

PR #10148, opened 2024-11-28
fix: compose installation via onboarding and cli tools

dgolovin:
### What does this PR do?

Fix compose onboarding install and CLI Tools preferences page install to work together.

### Screenshot / video of UI

https://github.com/user-attachments/assets/94cbe680-ee64-42e9-9cae-192f9e7c0ae5

Onboarding status updates after installing and removing compose on CLI Pref Page

https://github.com/user-attachments/assets/7ae56a6c-d435-487a-b321-ee9fa496378a

### What issues does this PR fix or reference?

Fix #9763.

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature

PR #12426, opened 2025-05-08
fix: addressed uninstall version error for kubectl installed via onboard

### LINKED ISSUE #11379: No Version detected error in CLI tools Page when trying to uninstall Kubectl that was installed via onboarding
### Bug description

I am testing a PR for removing cli binaries from system path and I noticed that if I install kubectl via onboarding and then tries to uninstall it from CLI Page, I got `no version detected` error.

This seems very similar to https://github.com/podman-desktop/podman-desktop/issues/9763 which was just referring to compose.

### Operating system

Linux, flatpak

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

See steps to reproduce in linked issue. Except for kubectl.
1. Start up podman desktop with fresh env. (remove `~/.local/share/containers/podman-desktop`) and make sure to wipe out kubectl from system beforehand
2. In global onboarding download kubectl on install it system-wide
3. Go To Cli tools
4. try to uninstall kubectl
5. assert: successfully removed
Actual: error

### Relevant log output

```shell

```

### Additional context

_No response_

bmahabirbu:
### What does this PR do?

This pr addresses the issue of kubectl version not being updated after onboard, which caused a no version detected error when trying to uninstall kubectl in the same session

Thanks to @odockal for linking a similar issue and @dgolovin for reference for a similar fix! In the long run, maybe we should align kubectl and compose extension.ts. Here is the reference PR https://github.com/podman-desktop/podman-desktop/pull/10148

fixes https://github.com/podman-desktop/podman-desktop/issues/11379

### How to test this PR?

Remove kubectl from host to simulate onboarding
i.e for windows 

go to
C:\Users\username\.local\share\containers\podman-desktop\extensions-storage
remove podman-desktop.kubectl-cli

go to
C:\Users\username\AppData\Local\Microsoft\WindowsApps
remove kubectl.exe 

Go to extension settings click on the cog to start the onboard download
Then go to cli settings and uninstall kubectl

Kubectl should be removed without error

### DIFF
--- a/extensions/kubectl-cli/src/extension.ts
+++ b/extensions/kubectl-cli/src/extension.ts
@@ -65,8 +65,14 @@ const kubectlCliDisplayName = 'kubectl';
 const kubectlCliDescription = `A command line tool for communicating with a Kubernetes cluster's control plane, using the Kubernetes API.\n\nMore information: [kubernetes.io](https://kubernetes.io/docs/reference/kubectl/)`;
 const imageLocation = './icon.png';
 
+let binaryVersion: string | undefined;
+let binaryPath: string | undefined;
+
 export async function activate(extensionContext: extensionApi.ExtensionContext): Promise<void> {
   initTelemetryLogger();
+
+  binaryVersion = undefined;
+  binaryPath = undefined;
 
   // Check kubectl binary has been downloaded and update both
   // the configuration setting and the context accordingly
@@ -137,8 +143,16 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
         // We are all done, so we can set the context value to false / downloaded to true
         extensionApi.context.setValue('kubectlIsNotDownloaded', false, 'onboarding');
+        // update the cli version
+        binaryVersion = kubectlVersionMetadata?.tag.slice(1);
+        if (!binaryVersion) {
+          return;
+        }
+        binaryPath = getSystemBinaryPath(kubectlCliName);
+        // update the version and the path
         kubectlCliTool?.updateVersion({
-          version: kubectlVersionMetadata.tag.slice(1),
+          version: binaryVersion,
+          path: binaryPath,
           installationSource: 'extension',
         });
         downloaded = true;
@@ -240,9 +254,6 @@ export function getStorageKubectlPath(extensionContext: extensionApi.ExtensionCo
 }
 
 export async function findKubeCtl(extensionContext: extensionApi.ExtensionContext): Promise<CliFinder> {
-  let binaryVersion: string | undefined;
-  let binaryPath: string | undefined;
-
   // Retrieve the version of the binary by running exec with --client
   try {
     const result = await extensionApi.process.exec(kubectlExecutableName, [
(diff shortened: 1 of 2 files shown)


extensions/kubectl-cli/src/extension.ts
```diff
@@ -65,9 +65,15 @@ const kubectlCliDisplayName = 'kubectl';
...
 const imageLocation = './icon.png';
 
+let binaryVersion: string | undefined;
+let binaryPath: string | undefined;
+
 export async function activate(extensionContext: extensionApi.ExtensionContext): Promise<void> {
   initTelemetryLogger();
 
+  binaryVersion = undefined;
+  binaryPath = undefined;
```
cdrage: no need for this? first time I've seen Sorcery AI be useful haha, but I don't think we need to re-initialize these here.

extensions/kubectl-cli/src/extension.ts
```diff
@@ -137,8 +143,16 @@ export async function activate(extensionContext: extensionApi.ExtensionContext):
 
         // We are all done, so we can set the context value to false / downloaded to true
         extensionApi.context.setValue('kubectlIsNotDownloaded', false, 'onboarding');
+        // update the cli version
+        binaryVersion = kubectlVersionMetadata?.tag.slice(1);
```
cdrage: is the slice(1) reliable enough to not cause issues?

cdrage (changes requested):
Tested this on macOS and it solves the issue, although I was unable to test it on Windows.

There's also missing tests for extension.ts

extensions/kubectl-cli/src/extension.ts
bmahabirbu: Good question! I think it will be ok, as the logic was in the previous version. I only changed how the logic was saved

prev previous 
` kubectlCliTool?.updateVersion({
          version: kubectlVersionMetadata.tag.slice(1),`

extensions/kubectl-cli/src/extension.ts
bmahabirbu: Good catch!

bmahabirbu:
updated PR to fix tests!

extensions/kubectl-cli/src/extension.ts
```diff
@@ -65,9 +65,16 @@ const kubectlCliDisplayName = 'kubectl';
 const kubectlCliDescription = `A command line tool for communicating with a Kubernetes cluster's control plane, using the Kubernetes API.\n\nMore information: [kubernetes.io](https://kubernetes.io/docs/reference/kubectl/)`;
 const imageLocation = './icon.png';
 
+interface KubectlState {
+  version?: string;
+  path?: string;
+}
+
+// State machine for version and path
+export const vp_state: KubectlState = {};
```
cdrage: Change to using camelCase similar to our other constants :) 

I'm also curious why we set this as a constant, but we do updates such as `vp_state.version = kubectlVersionMetadata?.tag.slice(1);`?

cdrage (changes requested):
Tests pass and I'm able to correctly uninstall.

Thanks for the fixes!

extensions/kubectl-cli/src/extension.ts
bmahabirbu: Thanks Charlie!

cdrage (approved):

PR #12432, opened 2025-05-09
feat: added new footer markdown property to message box

gastoner:
### What does this PR do?
Adds additional markdown property (e.g. for buttons) to message box

Follow up of https://github.com/podman-desktop/podman-desktop/pull/11991
Based on comment from Florent: https://github.com/podman-desktop/podman-desktop/pull/11991#pullrequestreview-2814558493

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/dialogs/MessageBox.spec.ts
+++ b/packages/renderer/src/lib/dialogs/MessageBox.spec.ts
@@ -67,8 +67,8 @@ describe('MessageBox', () => {
     expect(message).toBeInTheDocument();
     const detail = await screen.findByText(messageBoxOptions.detail ?? '');
     expect(detail).toBeInTheDocument();
-    const footerMarkdownDescription = await screen.findByText(messageBoxOptions.footerMarkdownDescription ?? '');
-    expect(footerMarkdownDescription).toBeInTheDocument();
+    const footerMarkdownDescription = await screen.findAllByText(messageBoxOptions.footerMarkdownDescription ?? '');
+    expect(footerMarkdownDescription[1]).toBeInTheDocument();
     const button1 = await screen.findByText(messageBoxOptions.buttons?.[0] ?? '');
     expect(button1).toBeInTheDocument();
     const button2 = await screen.findByText(messageBoxOptions.buttons?.[1] ?? '');
--- a/packages/renderer/src/lib/dialogs/MessageBox.spec.ts
+++ b/packages/renderer/src/lib/dialogs/MessageBox.spec.ts
@@ -49,6 +49,7 @@ describe('MessageBox', () => {
       title: 'My custom title',
       message: 'My message',
       detail: 'A more detailed message',
+      footerMarkdownDescription: 'Footer message in markdown',
       buttons: ['OK', 'Not OK'],
     };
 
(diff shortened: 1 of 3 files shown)


axel7083 (commented):
Could you add a screenshot of what it would look like ? I have some trouble understanding it

gastoner:
> Could you add a screenshot of what it would look like ? I have some trouble understanding it

It should be the thub up/down https://github.com/podman-desktop/podman-desktop/issues/10224#issuecomment-2636084884 Basicaly https://github.com/podman-desktop/podman-desktop/pull/11598 (quite outdated, but the thumb-up/down should be same)

packages/renderer/src/lib/dialogs/MessageBox.svelte
```diff
@@ -17,6 +18,7 @@ let type: string | undefined;
 let cancelId = -1;
 let defaultId: number;
 let buttonOrder: number[];
+let footerMarkdownDescription: string | undefined;
```
axel7083, line 21: this is not a slot (not matching PR title)

benoitf (approved):

benoitf:
FYI tests are failing

deboer-tim:
As per my comment in the other PR, it's not clear to me why message boxes need three customizable sections (message, footerMarkdown, and detail). Why aren't two enough or why couldn't detail use markdown instead?

gastoner:
> As per my comment in the other PR, it's not clear to me why message boxes need three customizable sections (message, footerMarkdown, and detail). Why aren't two enough or why couldn't detail use markdown instead?

I would use the detail slot with Markdown, but I would like to keep the content in `footerMarkdownDescription` aligned to center, the detail and message are aligned to left I think.

packages/renderer/src/lib/dialogs/MessageBox.spec.ts
```diff
@@ -66,6 +67,8 @@ describe('MessageBox', () => {
     expect(message).toBeInTheDocument();
     const detail = await screen.findByText(messageBoxOptions.detail ?? '');
     expect(detail).toBeInTheDocument();
+    const footerMarkdownDescription = await screen.findAllByText(messageBoxOptions.footerMarkdownDescription ?? '');
+    expect(footerMarkdownDescription[1]).toBeInTheDocument();
```
feloy, line 71: Why is `findAllByText` used, instead of `findByText`, and looking at the second value in the `footerMarkdownDescription` array?

gastoner: I think it is due to this: https://github.com/podman-desktop/podman-desktop/blob/main/packages/renderer/src/lib/markdown/Markdown.svelte#L147C1-L154C11
If I use just `findByText` it will get the content 2x

deboer-tim:
> > As per my comment in the other PR, it's not clear to me why message boxes need three customizable sections (message, footerMarkdown, and detail). Why aren't two enough or why couldn't detail use markdown instead?
> 
> I would use the detail slot with Markdown, but I would like to keep the content in `footerMarkdownDescription` aligned to center, the detail and message are aligned to left I think.

Using this logic we'd be creating a fourth property if we needed left-aligned markdown, or a fifth if we needed something slightly more/different than what's supported by markdown. It just feels like a one-off to support a specific case vs a clean api that could be used for other things.

If there wasn't the main/renderer API boundary in the way I'd say this should really be a fragment instead. Since it is there, maybe detail should be markdown and add a new detail-alignment property instead?

benoitf:
I think guys, you're doing a little bit of over-engineering

it's not exposed to the UI library or the extensions so it can iterate as much as we want.

PR #12433, opened 2025-05-09
chore: publish to cloudflare (in addition to netlify)

benoitf:
### What does this PR do?
suggested to use "View Deployment" commented in the PR rather than a status check by @axel7083 

the action should create on its own a "view deployment" entry

### How to test this PR?

🤞 

### DIFF
--- a/.github/workflows/publish-website-pr-cloudflare.yaml
+++ b/.github/workflows/publish-website-pr-cloudflare.yaml
@@ -1,0 +1,50 @@
+#
+# Copyright (C) 2025 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


gastoner (approved):
Codewise LGTM

Tested by :crossed_fingers:

axel7083 (dismissed):
I like the exploration, but I don't think we need to upload twice the website for each PR.
 (let's save the planet and some bandwidth) 
I was talking about the cloudflare as they have an integration with GitHub deployment, but we could mimick the same with netifly using some GitHub actions. (Here is an example https://github.com/axel7083/extension-podman-quadlet/pull/2).

benoitf:
@axel7083 the idea is to drop netlify and use only this method as the GH action is doing all the stuff.

but as usual, introduce new method, see if it works, then drop the other way

axel7083:
> but as usual, introduce new method, see if it works, then drop the other way

Should we see the deployment on this PR? Or should we see it once this is merged in main?

benoitf:
Only in main

It's because it's 2 jobs : one in pr mode without tokens
One being executed as a dependency with token access

axel7083 (approved):

PR #12445, opened 2025-05-12
feat: added switching users and clusters

### LINKED ISSUE #11283: Editing Kubernetes contexts

gastoner:
### What does this PR do?
Adds editing of users and clusters

- [x] Needs rebase after https://github.com/podman-desktop/podman-desktop/pull/12415
Closes #11283 

### How to test this PR?
1. Create context (minikube/kind/etc...)
2. Go to preferences -> kubernes context
3. Try to change user and cluster

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -27,6 +27,7 @@ import type {
   Cluster,
   Context as KubernetesContext,
   KubernetesObject,
+  User,
   V1ConfigMap,
   V1CronJob,
   V1Deployment,
@@ -2579,6 +2580,10 @@ export class PluginSystem {
       return kubernetesClient.getClusters();
     });
 
+    this.ipcHandle('kubernetes-client:getUsers', async (): Promise<User[]> => {
+      return kubernetesClient.getUsers();
+    });
+
     this.ipcHandle('kubernetes-client:getCurrentNamespace', async (): Promise<string | undefined> => {
       return kubernetesClient.getCurrentNamespace();
     });
@@ -2599,8 +2604,21 @@ export class PluginSystem {
     });
     this.ipcHandle(
       'kubernetes-client:updateContext',
-      async (_listener, contextName: string, newContextName: string, newContextNamespace: string): Promise<void> => {
-        return kubernetesClient.updateContext(contextName, newContextName, newContextNamespace);
+      async (
+        _listener,
+        contextName: string,
+        newContextName: string,
+        newContextNamespace: string,
+        newContextCluster: string,
+        newContextUser: string,
+      ): Promise<void> => {
+        return kubernetesClient.updateContext(
+          contextName,
+          newContextName,
+          newContextNamespace,
+          newContextCluster,
+          newContextUser,
+        );
       },
     );
 
--- a/packages/main/src/plugin/kubernetes/kubernetes-client-kubeconfig.spec.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client-kubeconfig.spec.ts
@@ -38,9 +38,6 @@ class TestKubernetesClient extends KubernetesClient {
   public setContexts(contexts: Context[]): void {
     this.kubeConfig.contexts = contexts;
   }
-  public getUsers(): User[] {
-    return this.kubeConfig.users;
-  }
   public setCurrentContext(name: string): void {
     this.currentContextName = name;
   }
@@ -201,7 +198,7 @@ describe('context tests', () => {
       throw new Error('originalContexts[0].name should be defined');
     }
 
-    await client.updateContext(originalContexts[0].name, 'new-name', 'new-namespace');
+    await client.updateContext(originalContexts[0].name, 'new-name', 'new-namespace', '', '');
     const contexts = client.getContexts();
     expect(contexts.length).toBe(2);
     expect(client.getContexts().length).toBe(2);
@@ -221,7 +218,7 @@ describe('context tests', () => {
       throw new Error('originalContexts[0].name should be defined');
     }
 
-    await client.updateContext(originalContexts[0].name, originalContexts[0].name, '');
+    await client.updateContext(originalContexts[0].name, originalContexts[0].name, '', '', '');
     const contexts = client.getContexts();
     expect(contexts.length).toBe(2);
     expect(client.getContexts().length).toBe(2);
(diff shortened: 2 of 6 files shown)


gastoner:
@deboer-tim @Firewall I've added editing users and clusters in this PR - is the follow up for https://github.com/podman-desktop/podman-desktop/pull/12415

feloy:
I'm not sure this feature is really useful, at least in this state (and is it really a demand in the issue https://github.com/podman-desktop/podman-desktop/issues/11283?).

In a kubeconfig, we have a "flat" list of users and clusters, which are reused in contexts. 

A specific user of the user's list, even if in the "common" list of users, is only meaningful for a specific cluster. 

To illustrate, in the screenshot below, I have a list of users including `kind-cluster-cluster-1` and `sandbox-user-pyk9l`, but only the first one is really usable for the cluster `kind-kind-cluster-1` selected above.

edit-users-clusters

gastoner:
> I'm not sure this feature is really useful, at least in this state (and is it really a demand in the issue #11283?).
> 
> In a kubeconfig, we have a "flat" list of users and clusters, which are reused in contexts.
> 
> A specific user of the user's list, even if in the "common" list of users, is only meaningful for a specific cluster.
> 
> To illustrate, in the screenshot below, I have a list of users including `kind-cluster-cluster-1` and `sandbox-user-pyk9l`, but only the first one is really usable for the cluster `kind-kind-cluster-1` selected above.
> 
> edit-users-clusters

Tim has mentioned this in his comment in name/namespace editing PR https://github.com/podman-desktop/podman-desktop/pull/12415#issuecomment-2863573281

gastoner:
cc @deboer-tim

deboer-tim:
Looks pretty good to me, just a few comments:
1. Re @feloy's comment. I agree, ideally it wouldn't include any user that we know won't work with the given cluster. But is there no way to configure the same user on two clusters? And even though this is ordered vertically, you don't necessarily always do them in order or expect the user list to change when you change the cluster.
So I think this would come down to either reactive validation or some new UI design, which I think would be good but personally don't think it needs to hold up the basic function in this PR.

3. I don't have a problem with exposing `getUsers()` to renderer, but can't we just use something like`$kubernetesContexts.map(c => c.user)`? Might make it easier to implement the previous comment too.

4. The first time I tried the dropdown I got this:
Screenshot 2025-05-20 at 3 28 19 PM
Yes, I can scroll down to see the rest, but it's not a great initial experience. I'd suggest either making the dropdown appear over top of the buttons, or if we know how many rows the dropdown will be (and it is a small, fixed number..) adding some artificial padding below.

gastoner:
> Looks pretty good to me, just a few comments:
> 
>     1. Re @feloy's comment. I agree, ideally it wouldn't include any user that we know won't work with the given cluster. But is there no way to configure the same user on two clusters? And even though this is ordered vertically, you don't necessarily always do them in order or expect the user list to change when you change the cluster.
>        So I think this would come down to either reactive validation or some new UI design, which I think would be good but personally don't think it needs to hold up the basic function in this PR.
> 
>     2. I don't have a problem with exposing `getUsers()` to renderer, but can't we just use something like`$kubernetesContexts.map(c => c.user)`? Might make it easier to implement the previous comment too.
> 
>     3. The first time I tried the dropdown I got this:
> 
> 
> Screenshot 2025-05-20 at 3 28 19 PM Yes, I can scroll down to see the rest, but it's not a great initial experience. I'd suggest either making the dropdown appear over top of the buttons, or if we know how many rows the dropdown will be (and it is a small, fixed number..) adding some artificial padding below.

1. I can add some checks for it 
2. I think this will just get the list of currently used users => if there is an unused user you wont be able to switch to him
3. I was thinking the same thing, on the other hand I feel like the name/namespace are the "main" options so I would expect to have them on top, I'll try add the padding first

gastoner:
I've tried to add dynamically padding, but there would be quite a lot of refactoring, I would say and adding additional props to the Dropdown component. So I went with an easier way of just changing the order of elements: 
image

deboer-tim:
I appreciate thinking out of the box 😀 but order should stay consistent with the page with name first. Namespace can still be last, I can't tell if that's enough space for the dropdown when you have multiple users.

gastoner:
I've tried to add dynamically padding, but there would be quite a lot of refactoring, I would say and adding additional props to the Dropdown component. So I went with an easier way of just changing the order of elements: 
image


> I appreciate thinking out of the box 😀 but order should stay consistent with the page with name first. Namespace can still be last, I can't tell if that's enough space for the dropdown when you have multiple users.

I think that it will good for aprox 3 rows

gastoner:
For the user the scroll bar will be visible from 4 and more users. 
When having 3 or less users the dropdown will fit the dialog
image
cc @deboer-tim @feloy

benoitf:
hello, should it be flagged as draft PR for now ?

gastoner:
> hello, should it be flagged as draft PR for now ?

More like a decision if to close this / merge it and close the issue @deboer-tim is the author of the issue

deboer-tim (approved):
Sorry for delay, I am good with this. If there are a lot of contexts the scroll is still unfortunate, but the order now matches and the function worked in testing.

cdrage (changes requested):
When reviewing the functionality I was unable to save anything :(

There's zero error logs or any output when trying to save.

This was tested on my mac M1. 

I rebased against main as well for the latest code / did `pnpm install` / etc. Maybe there was a change in another commit after rebasing that may have made it be unable to save?

https://github.com/user-attachments/assets/71a656f0-65a8-467f-87e5-0881a69b4598

gastoner:
@deboer-tim @cdrage I've updated one if statement, now it should be fine, can you please try again?

cdrage (changes requested):
@gastoner Still not working :(

I WAS however able to edit my "second" context (already deleted in the video).

But not the main active one. Is there somehow an if statement preventing saving because it's the "primary" context being used or something?


https://github.com/user-attachments/assets/1e69be95-0c24-46f3-b1a4-b53960ece5a3

gastoner:
@cdrage hopefuly this time :crossed_fingers:

cdrage (commented):
Tested thoroughly against my personal cluster and it works great!

Thank you for the changes :).

Looked through the code as well and LGTM, great job on the tests too.


https://github.com/user-attachments/assets/d777d33c-2b34-40d8-a36b-193f49370e3c

cdrage (approved):

deboer-tim (approved):

Firewall (approved):
Great job! 🚀

Issue #12215, opened 2025-04-15
GitHub actions: ensure permissions field is defined


benoitf:
### Is your enhancement related to a problem? Please describe

see https://clomonitor.io/projects/cncf/podman-desktop#podman-desktop_security

the check `Token permissions` should be 💚 to be able to close this issues

### Describe the solution you'd like

add the field on all the workflows as it's a best practice

axel7083:
![Image](https://github.com/user-attachments/assets/827c674e-11ed-45f9-b3e7-ce708ebb3c20)

Closing as resolved 🥳

PR #12455, opened 2025-05-12
chore(next-build): define explicit permissions

axel7083:
### What does this PR do?

Define explicitly the permissions the `next-build` workflow has access. We don't need to write anything on podman-desktop/podman-desktop. The workflow uses `PODMAN_DESKTOP_BOT_TOKEN`, `NPM_AUTH_TOKEN` and `SEGMENT_WRITE_KEY` secrets, I can't find a specific permissions to use secrets[^1]

[^1]: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions 

### How to test this PR?

- After merging, next build pipeline should be 💚

### DIFF
--- a/.github/workflows/next-build.yaml
+++ b/.github/workflows/next-build.yaml
@@ -27,6 +27,9 @@ env:
   GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
   DEBUG: electron-builder
 
+permissions:
+  contents: read
+
 jobs:
 
   tag:


feloy (approved):
Referring to another example (https://github.com/redhat-developer/podman-desktop-rhel-ext/blob/main/.github/workflows/build-next.yaml), where `packages: write` is necessary for deploying to `ghcr.io`, there is only the `contents: read` permission remaining. LGTM

cdrage (approved):

benoitf:
I'm not sure it will be enough as it's publishing files to another repository but let's try and add the missing permissions on demand

axel7083:
🤞

axel7083:
Check the logs in https://github.com/podman-desktop/podman-desktop/actions/runs/14994412770

benoitf:
Maybe we're good as anyway it's using another token

axel7083:
> Maybe we're good as anyway it's using another token

This is what I think about, because we uses secret for the token, and using secrets don't need specific permission

PR #12470, opened 2025-05-13
chore(test): parameterize runner close timeout

amisskii:

### What does this PR do?

This PR makes the runner close timeout configurable. This is useful in scenarios where shutdown times may vary due to different system configurations or runtime conditions.


ScrewTSW (approved):
LGTM

PR #12473, opened 2025-05-13
chore: make the file items in preferences clearable

### LINKED ISSUE #11904: cannot select the image path because it only allow normal files
### Bug description

Image

Image

Here I select a wrong image path 'README', then Create, then I know here should be a valid image file.
Then I want to reset the setting, But there's no way.

I have to skip this setting and try it later.

### Operating system

Windows 11

### Installation Method

### Version

1.17.0

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
Changes the field of the image path in the Podman configuration to not be read only so that users are able to remove a file if they selected one and write/paste the path themselves

### Screenshot / video of UI

https://github.com/user-attachments/assets/652db41a-2f24-45e3-b633-dffd11ef22eb

Fixes https://github.com/podman-desktop/podman-desktop/issues/11904

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/inputs/Input.svelte
+++ b/packages/ui/src/lib/inputs/Input.svelte
@@ -112,9 +112,10 @@ async function onClear(): Promise<void> {
     {#if clearable}
       <button
         class="px-0.5 cursor-pointer text-[color:var(--pd-input-field-icon)] group-hover:text-[color:var(--pd-input-field-hover-icon)] group-focus-within:text-[color:var(--pd-input-field-focused-icon)]"
-        class:hidden={!value || readonly || disabled}
+        class:hidden={!value || disabled}
         aria-label="clear"
-        onclick={onClear}>
+        onclick={onClear}
+        type="button">
         <Fa icon={faXmark} />
       </button>
     {/if}
--- a/extensions/podman/packages/extension/package.json
+++ b/extensions/podman/packages/extension/package.json
@@ -151,8 +151,7 @@
           "type": "string",
           "format": "file",
           "scope": "ContainerProviderConnectionFactory",
-          "description": "Image Path (Optional)",
-          "readonly": false
+          "description": "Image Path (Optional)"
         },
         "podman.factory.machine.image-uri": {
           "type": "string",
--- a/extensions/podman/packages/extension/package.json
+++ b/extensions/podman/packages/extension/package.json
@@ -151,7 +151,8 @@
           "type": "string",
           "format": "file",
           "scope": "ContainerProviderConnectionFactory",
-          "description": "Image Path (Optional)"
+          "description": "Image Path (Optional)",
+          "readonly": false
         },
         "podman.factory.machine.image-uri": {
           "type": "string",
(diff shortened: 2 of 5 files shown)


benoitf:
non-binding: hello, one issue that I see is that you can enter an invalid path while clicking on the button you were sure to have a valid path.
Maybe I would expect to see only a way to clear the line but not entering any string

SoniaSandler:
@benoitf I could add a reset button/icon like we have in the preferences page to reset the input field

benoitf:
@SoniaSandler I think we have the property `clearable` on the Input component

so we could also have that property enabled on FileInput (optional or by default)

benoitf:
anyway I'm not against the current proposal. I feel like it's too wide open but people may like it (but allowing to clear the selection with a single icon/click would still be valuable)

odockal:
> non-binding: hello, one issue that I see is that you can enter an invalid path while clicking on the button you were sure to have a valid path. Maybe I would expect to see only a way to clear the line but not entering any string

I agree with you, but it is quite limiting the usage of the input fields (users cannot copy/paste), making it hard to test (unless we hack around with changing the attribute value on the fly). The best would be to have a specific validation for particular input field types, imho.

benoitf:
I can copy paste, I click on the button, it's opening a dialog box, from there I can enter my path clicking on '/' character

SoniaSandler:
@odockal @benoitf making this field editable could be additional discussion, but for now we can make this field at least clearable so that users could delete their input right there instead of having to exit the form

benoitf:
unit tests are failing

packages/ui/src/lib/inputs/Input.svelte
```diff
@@ -79,8 +79,9 @@ async function onClear(): Promise<void> {
     {#if clearable}
       <button
         class="px-0.5 cursor-pointer text-[color:var(--pd-input-field-icon)] group-hover:text-[color:var(--pd-input-field-hover-icon)] group-focus-within:text-[color:var(--pd-input-field-focused-icon)]"
-        class:hidden={!value || readonly || disabled}
+        class:hidden={!value || disabled}
```
axel7083, line 115: this is not really logic, we should not be able to clear the value if we are readonly, I get why you did that, but maybe the full input should not be readonly?

SoniaSandler: @axel7083 I understand what you're saying, but as mentioned in another comment, making it not readonly would be too broad. In this particular case (FileInput), the field might be readonly, but it's editable, so users can change this value and they should be able to clear this value too.
If we feel this topic might need more discussion, I can add it to the bi-weekly meeting
@benoitf what do you think?

benoitf: FileInput is composed of a button and an input, so I should be able to clear it if I used the button to select something

about
> we should not be able to clear the value if we are readonly

we should be not be able to clear if clearable is false, but if I say to the component : allow to clear then it's up to my responsibility to say if I want to clear it.

so to be the `|| readonly` check is something the component should not check as it's up to the consumer to decide.

benoitf (approved):

axel7083 (approved):

PR #12519, opened 2025-05-17
feat(extension): add search by description in extension list

omertuc:
### What does this PR do?

I tried searching for the bootc extension by typing `bootc` in the search box, but I couldn't find it. I realized that the search only looks at the display name of the extension, which doesn't contain the word bootc.

I changed the search logic to include the description of the extension in the search. Now, when I type `bootc`, it matches the extension with the display name "Bootable Container" because the description mentions bootc.

Searching only by name seems a bit strict, especially considering the currently small amount of extensions.

Ideally in the future, extensions will be able to provide a list of keywords associated with them so we can filter against those as well, instead of something arbitrary like the description.

### Screenshot / video of UI

Before

https://github.com/user-attachments/assets/700f255e-c148-4883-99e6-2fc0b884afda

After

https://github.com/user-attachments/assets/cde0b8af-6a28-4b5f-b86e-291c5c300831

None

### How to test this PR?

Couldn't find where I should write the tests for this - if needed please let me know and I'll edit the PR

### DIFF
--- a/packages/renderer/src/lib/extensions/ExtensionList.spec.ts
+++ b/packages/renderer/src/lib/extensions/ExtensionList.spec.ts
@@ -18,7 +18,7 @@
 
 import '@testing-library/jest-dom/vitest';
 
-import { fireEvent, render, screen } from '@testing-library/svelte';
+import { cleanup, fireEvent, render, screen } from '@testing-library/svelte';
 import { beforeEach, expect, test, vi } from 'vitest';
 
 import { type CombinedExtensionInfoUI } from '/@/stores/all-installed-extensions';
@@ -35,7 +35,7 @@ beforeEach(() => {
 export const aFakeExtension: CatalogExtension = {
   id: 'idAInstalled',
   publisherName: 'FooPublisher',
-  shortDescription: 'this is short A',
+  shortDescription: 'this is short A. The word bar appears here but not in the title',
   publisherDisplayName: 'Foo Publisher',
   extensionName: 'a-extension',
   displayName: 'A Extension',
@@ -88,6 +88,7 @@ const combined: CombinedExtensionInfoUI[] = [
   {
     id: 'idAInstalled',
     displayName: 'A installed Extension',
+    description: 'The word bar appears here but not in the title',
     removable: true,
     state: 'started',
   },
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/extensions/ExtensionList.svelte
```diff
@@ -29,7 +29,9 @@ const filteredInstalledExtensions: Readable<CombinedExtensionInfoUI[]> = derived
   [combinedInstalledExtensions, combinedInstalledExtensionSearchPattern],
   ([$combinedInstalledExtensions, $combinedInstalledExtensionSearchPattern]) => {
     return $combinedInstalledExtensions.filter(extension => {
-      return extension.displayName.toLowerCase().includes($combinedInstalledExtensionSearchPattern.toLowerCase());
+      return `${extension.displayName} ${extension.description}`
```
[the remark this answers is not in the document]
omertuc: I don't trust my typescript skills enuogh to say whether this is a good suggestion or not, so if a human can confirm this is good I will do that

benoitf (commented):
Hello @omertuc could you add a unit test covering this enhancement, thanks

omertuc:
See description, I'm not quite sure where to. Could you point at the right file?

benoitf:
@omertuc usually when there is a .svelte file there is a counterpart .spec.ts file

https://github.com/podman-desktop/podman-desktop/blob/65912d1c13055a531c8db3093dede603684887e4/packages/renderer/src/lib/extensions/ExtensionList.spec.ts in your case with unit tests

omertuc:
Thank you for the pointer, I've pushed some tests

packages/renderer/src/lib/extensions/ExtensionList.spec.ts
```diff
@@ -169,3 +170,45 @@ test('Expect to see empty screens on both pages', async () => {
...
+  expect(myExtension1).toBeInTheDocument();
+
+  // second extension should not be there as only in catalog (not installed) and doesn't have "bar" in the description
+  const extensionIdB = screen.queryByRole('group', { name: 'B Extension' });
+  expect(extensionIdB).not.toBeInTheDocument();
+
+  cleanup();
+
+  // Change the search
+  render(ExtensionList, { searchTerm: 'foo' });
```
axel7083, line 190: Instead of using cleanup to _cleanup_ the `screen` you can directly use the output of the render function

```suggestion
  // Change the search
  const { queryByRole } = render(ExtensionList, { searchTerm: 'foo' });
  
  queryByRole(...)
```

omertuc: I couldn't get that to work - render() writes to some global document, and queryByRole finds leftover stuff from the beginning of the test

axel7083 (approved):
LGTM, thanks for the contribution!

cdrage (approved):
Awesome contribution! Tested it out and was able to now see bootc come up :)

The code LGTM. I do have similar doubts as @axel7083 regarding the test, but to me it feels like a non-blocker. The tests already cover the scenario well.

omertuc:
> I do have similar doubts as @axel7083 regarding the test

I'll be happy to know how to do it right without cleanup. Couldn't get it to work myself.

omertuc:
Is there anything left to do to get this to merge?

benoitf (approved):
thanks @omertuc

PR #12559, opened 2025-05-20
fix: remove unused breadcrumbTitle from Page and related components

### LINKED ISSUE #12558: breadcrumbTitle in Page.svelte is not used
### Bug description

https://github.com/podman-desktop/podman-desktop/blob/main/packages/ui/src/lib/layouts/Page.svelte#L13

breadcrumbTitle in Page.svelte is not used.

We should remove the property, and all usages of it

### Operating system

all

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

feloy:

### What does this PR do?

Remove unused breadcrumbTitle from Page

I looked old versions (back to 1.10), and I cannot see it used in older versions.

Fixes #12558 

Needed by #11579 

### How to test this PR?


axel7083 (approved):

benoitf (approved):
breaking change for consumers but here we're breaking a lot (for Snippets name for example) etc

should we introduce a MIGRATION.md file for such changes from version to version

(not mandatory for this PR)

feloy:
> breaking change for consumers but here we're breaking a lot (for Snippets name for example) etc
> 
> should we introduce a MIGRATION.md file for such changes from version to version
> 
> (not mandatory for this PR)

Yes, I was wondering how to indicate these breaking changes to the user.

I'll create a MIGRATION.md file as part of this PR

benoitf (commented):
I think the MIGRATION.md should be part of the packages/ui folder

Let say I'm a user of the library if I go there
https://www.npmjs.com/package/@podman-desktop/ui-svelte

this is where I would expect to see the link to the file maybe

I think MIGRATION.md at the root level will confuse people wanted to migrate from previous versions of Podman Desktop

benoitf (approved):
thx

PR #12582, opened 2025-05-21
fix: clear containers logs permanently

### LINKED ISSUE #11782: Clearing logs only clears them temporarily
### Bug description

The logs reappear after clearing them and swapping container details tabs back and forth.

### Operating system

Windows 11

### Installation Method

### Version

next (development version)

### Steps to reproduce

1. Create a container
2. Go to container details and Logs
3. Click on clear logs, the logs disappear
4. Click on a different tab (Summary/Inspect/..)
5. Go back to Logs
6. Expect: logs remain cleared
    Actual result: the same logs are back

### Relevant log output

```shell

```

### Additional context

https://github.com/user-attachments/assets/504d6443-178e-458a-aad7-9c73fbeec3fa

feloy:
### What does this PR do?

- Clear container logs permanently
- Change the "Clear logs" button with a "Restore Logs" button when logs are cleared
 
### Screenshot / video of UI

https://github.com/user-attachments/assets/b2c853a3-ffc9-4912-99e8-b6f11efc9aeb

Fixes #11782 

### How to test this PR?

- Clear logs
- go to other pages and come back, logs should still be cleared
- add some logs, only the new logs should appear
- clear again
- ...
- Check that after several clears, restoring the logs restore all the logs

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
@@ -99,7 +99,6 @@ function afterTerminalInit(): void {
   const { setRevert } = mount(ContainerDetailsLogsClear, {
     target: xtermElement,
     props: {
-      terminal: logsTerminal,
       onclear: async () => {
         if (logsLength === skip) {
           setSkip(0);
@@ -114,13 +113,11 @@ function afterTerminalInit(): void {
 }
 
 function setSkip(n: number): void {
-  const storeValue = $containerLogsCleared;
   if (n) {
-    storeValue.set(getKeyValue(container.id, container.engineId), n);
+    $containerLogsCleared.set(getKeyValue(container.id, container.engineId), n);
   } else {
-    storeValue.delete(getKeyValue(container.id, container.engineId));
+    $containerLogsCleared.delete(getKeyValue(container.id, container.engineId));
   }
-  containerLogsCleared.set(storeValue);
 }
 
 function clearTerminal(): void {
--- a/packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
@@ -4,6 +4,9 @@ import '@xterm/xterm/css/xterm.css';
 import { EmptyScreen } from '@podman-desktop/ui-svelte';
 import type { Terminal } from '@xterm/xterm';
 import { mount, onDestroy, onMount } from 'svelte';
+import type { Unsubscriber } from 'svelte/store';
+
+import { containerLogsCleared, getKeyValue } from '/@/stores/container-logs-cleared';
 
 import { isMultiplexedLog } from '../stream/stream-utils';
 import NoLogIcon from '../ui/NoLogIcon.svelte';
@@ -24,7 +27,7 @@ $: {
     refContainer &&
     (refContainer.id !== container.id || (refContainer.state !== container.state && container.state !== 'EXITED'))
   ) {
-    logsTerminal?.clear();
+    clearTerminal();
     fetchContainerLogs().catch((err: unknown) => console.error(`Error fetching container logs ${container.id}`, err));
   }
   refContainer = container;
@@ -33,17 +36,32 @@ let terminalParentDiv: HTMLDivElement;
 
 let logsTerminal: Terminal;
 
+let logsLength = 0;
+let skip = 0;
+let cancellableTokenId: number | undefined = undefined;
+let containerLogsClearedUnsubscriber: Unsubscriber | undefined = undefined;
+
+let containerDetailsLogsClearSetRevert: ((revert: boolean) => void) | undefined = undefined;
+
+$: {
+  if (skip > 0 && logsLength === skip) {
+    containerDetailsLogsClearSetRevert?.(true);
+  } else {
+    containerDetailsLogsClearSetRevert?.(false);
+  }
+}
+
 function callback(name: string, data: string): void {
   if (name === 'first-message') {
     noLogs = false;
     // clear on the first message
-    logsTerminal?.clear();
+    clearTerminal();
   } else if (name === 'data') {
     noLogs = false;
     if (isMultiplexedLog(data)) {
-      logsTerminal?.write(data.substring(8) + '\r');
+      addDataToTerminal(data.substring(8));
     } else {
-      logsTerminal?.write(data + '\r');
+      addDataToTerminal(data);
     }
   }
   if (!noLogs) {
@@ -52,8 +70,17 @@ function callback(name: string, data: string): void {
 }
 
 async function fetchContainerLogs(): Promise<void> {
+  containerLogsClearedUnsubscriber?.();
+  containerLogsClearedUnsubscriber = containerLogsCleared.subscribe(cleared => {
+    skip = cleared.get(getKeyValue(container.id, container.engineId)) ?? 0;
+  });
+
+  if (cancellableTokenId) {
+    await window.cancelToken(cancellableTokenId);
+  }
+  cancellableTokenId = await window.getCancellableTokenSource();
   // grab logs of the container
-  await window.logsContainer({ engineId: container.engineId, containerId: container.id, callback });
+  await window.logsContainer({ engineId: container.engineId, containerId: container.id, callback, cancellableTokenId });
 }
 
 function afterTerminalInit(): void {
@@ -61,20 +88,55 @@ function afterTerminalInit(): void {
   let xtermElement = terminalParentDiv.querySelector('.xterm');
   xtermElement ??= terminalParentDiv;
   // add svelte component using this xterm element
-  mount(ContainerDetailsLogsClear, {
+  const { setRevert } = mount(ContainerDetailsLogsClear, {
     target: xtermElement,
     props: {
       terminal: logsTerminal,
+      onclear: async () => {
+        if (logsLength === skip) {
+          setSkip(0);
+        } else {
+          setSkip(logsLength);
+        }
+        await fetchContainerLogs();
+      },
     },
   });
+  containerDetailsLogsClearSetRevert = setRevert;
+}
+
+function setSkip(n: number): void {
+  const storeValue = $containerLogsCleared;
+  if (n) {
+    storeValue.set(getKeyValue(container.id, container.engineId), n);
+  } else {
+    storeValue.delete(getKeyValue(container.id, container.engineId));
+  }
+  containerLogsCleared.set(storeValue);
+}
+
+function clearTerminal(): void {
+  logsTerminal?.clear();
+  logsLength = 0;
+}
+
+function addDataToTerminal(data: string): void {
+  if (logsLength >= skip) {
+    logsTerminal?.write(data + '\r');
+  }
+  logsLength += data.length + 1;
 }
 
 onMount(async () => {
   await fetchContainerLogs();
 });
 
-onDestroy(() => {
+onDestroy(async () => {
   logsTerminal?.dispose();
+  containerLogsClearedUnsubscriber?.();
+  if (cancellableTokenId) {
+    await window.cancelToken(cancellableTokenId);
+  }
 });
 </script>
 
--- a/packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
+++ b/packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
@@ -14,15 +14,27 @@ import TerminalWindow from '../ui/TerminalWindow.svelte';
 import ContainerDetailsLogsClear from './ContainerDetailsLogsClear.svelte';
 import type { ContainerInfoUI } from './ContainerInfoUI';
 
-export let container: ContainerInfoUI;
+const {
+  container,
+}: {
+  container: ContainerInfoUI;
+} = $props();
 
-// Log
-let refContainer: ContainerInfoUI;
 // logs has been initialized
-let noLogs = true;
+let noLogs = $state(true);
+// the terminal displaying the logs
+let logsTerminal = $state<Terminal>();
+
+// the length of logs received
+let logsLength = $state(0);
+// the length of logs to skip displaying
+let skip = $state(0);
+
+// the container to get the logs from
+let refContainer: ContainerInfoUI;
 
 // need to refresh logs when container is switched or state changes
-$: {
+$effect(() => {
   if (
     refContainer &&
     (refContainer.id !== container.id || (refContainer.state !== container.state && container.state !== 'EXITED'))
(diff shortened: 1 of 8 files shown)


packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
```diff
@@ -52,29 +70,73 @@ function callback(name: string, data: string): void {
...
+}
+
+function setSkip(n: number): void {
+  const storeValue = $containerLogsCleared;
+  if (n) {
+    storeValue.set(getKeyValue(container.id, container.engineId), n);
+  } else {
+    storeValue.delete(getKeyValue(container.id, container.engineId));
+  }
+  containerLogsCleared.set(storeValue);
```
SoniaSandler: Question (non-blocking): Isn't this redundant as the map has been updated in the if/else section?

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
```diff
@@ -52,29 +70,73 @@ function callback(name: string, data: string): void {
 }
 
 async function fetchContainerLogs(): Promise<void> {
+  containerLogsClearedUnsubscriber?.();
+  containerLogsClearedUnsubscriber = containerLogsCleared.subscribe(cleared => {
```
SoniaSandler, line 79: Question (non-blocking): is there a reason to use `subscribe` and `unsubscribe` over `$containerLogsCleared.get` in this case or it's just a preference?

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
```diff
@@ -33,17 +36,32 @@ let terminalParentDiv: HTMLDivElement;
 
 let logsTerminal: Terminal;
 
+let logsLength = 0;
+let skip = 0;
+let cancellableTokenId: number | undefined = undefined;
```
SoniaSandler: Question (non-blocking): how does this `cancellableTokenId` affect the new changes?

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
```diff
@@ -52,29 +70,73 @@ function callback(name: string, data: string): void {
...
 function afterTerminalInit(): void {
   // mount the svelte5 component to the terminal xterm element
   let xtermElement = terminalParentDiv.querySelector('.xterm');
   xtermElement ??= terminalParentDiv;
   // add svelte component using this xterm element
-  mount(ContainerDetailsLogsClear, {
+  const { setRevert } = mount(ContainerDetailsLogsClear, {
     target: xtermElement,
     props: {
       terminal: logsTerminal,
```
SoniaSandler: This line (67/94) can be removed since this prop was removed from `ContainerDetailsLogsClear`

SoniaSandler (approved):
LGTM, works as expected. I just added a few questions to better understand the new code

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
```diff
@@ -33,17 +36,32 @@ let terminalParentDiv: HTMLDivElement;
...
   if (name === 'first-message') {
     noLogs = false;
     // clear on the first message
-    logsTerminal?.clear();
+    clearTerminal();
   } else if (name === 'data') {
     noLogs = false;
     if (isMultiplexedLog(data)) {
-      logsTerminal?.write(data.substring(8) + '\r');
+      addDataToTerminal(data.substring(8));
```
axel7083, line 67: why removing the `\r`?

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
```diff
@@ -33,17 +36,32 @@ let terminalParentDiv: HTMLDivElement;
...
 let logsTerminal: Terminal;
 
+let logsLength = 0;
+let skip = 0;
+let cancellableTokenId: number | undefined = undefined;
+let containerLogsClearedUnsubscriber: Unsubscriber | undefined = undefined;
+
+let containerDetailsLogsClearSetRevert: ((revert: boolean) => void) | undefined = undefined;
+
+$: {
```
axel7083: can't we just migrate to svelte5 instead of introducing svelte4 reactivity pattern?

axel7083 (changes requested):
Using a store in the frontend to store the size of the logs cleared seems hacky and introduce a memory leak at runtime, as the store is never cleared.

This really feel like this a backend problem, and maybe we should explore using the libpod api arguments, such as `since` ref https://docs.podman.io/en/v4.8.1/_static/api.html?version=v3.2#tag/containers/operation/ContainerLogsLibpod.

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
feloy: It is in `addDataToTerminal`

feloy:
> Using a store in the frontend to store the size of the logs cleared seems hacky and introduce a memory leak at runtime, as the store is never cleared.
> 

If we want to make the clear permanent, we need to store the information somewhere. But I agree that we need to clear the entries for containers being deleted at least.

> This really feel like this a backend problem, and maybe we should explore using the libpod api arguments, such as `since` ref https://docs.podman.io/en/v4.8.1/_static/api.html?version=v3.2#tag/containers/operation/ContainerLogsLibpod.

I'm not sure that it is a backend problem, as this clearing is clearing the display of the logs, not the logs themselves: if we do this on the backend, the logs for a container will be cleared for all callers of `containerLogs`, but other callers may want to have a view of the logs not related with the `ContainerDetailsLogs` view (for example extensions also have access to this method).

Using `since` can also be a solution. As it based on timestamps, we need to be sure we are really in sync

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
feloy: yes you are right, it is not necessary. I'll simplify. Thanks

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
feloy: As `containerLogsCleared` is set only once (in `setSkip`), we could compute the `skip` value at this time, and get rid of this `subscribe`. But I would prefer to keep the reactive pattern, to modify the value of the store in one or several places (in `setSkip` only for the moment) and react on the change of its value in another place (in the subscribe)

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
feloy: this is used to cancel the call to `window.logsContainer`, or the callback function is still called when we quit the page or when we make a new call to `fetchContainerLogs`.

packages/renderer/src/lib/container/ContainerDetailsLogs.svelte
feloy: yes, good catch, thanks

axel7083:
> I'm not sure that it is a backend problem, as this clearing is clearing the display of the logs, not the logs themselves

If the user request to only have the logs since X, we do not send the full logs, to hide it in the frontend, this would waste CPU and memory, the frontend should only, as you mention display the logs, IMHO not having complicated logic to only show portion of them.

> if we do this on the backend, the logs for a container will be cleared for all callers of containerLogs,

what is containerLogs ? I am not sure to follow your logic here. When the `ContainerDetailsLog.svelte` component mount, it uses `window.logsContainer`

https://github.com/podman-desktop/podman-desktop/blob/08f1e32c5aeb648fcfa3cb39a73c2fabb46082de/packages/renderer/src/lib/container/ContainerDetailsLogs.svelte#L56

which implemention uses the container registry to logs the container given the arguments.

https://github.com/podman-desktop/podman-desktop/blob/56230a5db2e22a96943635a0e712ef912409e27d/packages/main/src/plugin/extension/extension-loader.ts#L1091-L1093

which itself make a libpod call as visible bellow

https://github.com/podman-desktop/podman-desktop/blob/08f1e32c5aeb648fcfa3cb39a73c2fabb46082de/packages/main/src/plugin/container-registry.ts#L1686-L1692

>  (for example extensions also have access to this method).

Every caller would have the full logs, sorted depending on their parameters

Each caller have a dedicated connection, meaning if we pass some parameters, it would be unique for each caller, therefore would not have side effect.

> Using since can also be a solution. As it based on timestamps, we need to be sure we are really in sync

Other method could be to use `timestamp: true` in the logs payload, so every logs have a timestamp allowing us to have fine-grained control, but yet, I still believe that the frontend should not store the full backend if we only want to display part of it :/

feloy:
> > I'm not sure that it is a backend problem, as this clearing is clearing the display of the logs, not the logs themselves
> 
> If the user request to only have the logs since X, we do not send the full logs, to hide it in the frontend, this would waste CPU and memory, the frontend should only, as you mention display the logs, IMHO not having complicated logic to only show portion of them.
> 
> > if we do this on the backend, the logs for a container will be cleared for all callers of containerLogs,
> 
> what is containerLogs ? I am not sure to follow your logic here. When the `ContainerDetailsLog.svelte` component mount, it uses `window.logsContainer`
> 

Sorry, not `containerLogs` but `logContainers` as you are mentioning

feloy:
> Every caller would have the full logs, sorted depending on their parameters

> Each caller have a dedicated connection, meaning if we pass some parameters, it would be unique for each caller, therefore would not have side effect.

If the *skip* information (timestamp or length to skip, it's the same problem) is stored in the backend, the caller can only ask for the container's logs. And the same filtering will be done in the backend, no matter which caller asked the logs.

axel7083:
> If the _skip_ information (timestamp or length to skip, it's the same problem) is stored in the backend, the caller can only ask for the container's logs. And the same filtering will be done in the backend, no matter which caller asked the logs.

There is probably a missunderstanding here, each caller would have to specify its own parameters, and the backend would not have to do any filtering, it just pass the parameters to the libpod request.

Let's look at how some tools are doing something similar, as I am curious to understand how they manage similar issues.

## Grafana

The frontend of Grafana is making a query providing a `from` and `until` parameters, and have a pulling update mechanism to get update.

image

check it out https://play.grafana.org/

## Loki by Grafana

They use a similar system for logs, query based, using timestamp

ref https://grafana.com/docs/loki/latest/query/template_functions/#timestamp

## Open Lens

Open Lens allow users to filter logs based on several settings, for example showing the timestamp or not, where it makes a new connection to the backend, itself making a new request on the updated settings.

image


## Lens

Lens is more sophisticated allowing us to define a timestamp from which we want the logs, it just cancel the previous follow connection, and make a new one with the appropriate arguments to only show after the given timestamp.

---

Just showing that those tools usually don't put the all logs into the frontend to filter it on it, and usually just forward from the producer, to the frontend with only what the user need to see. And this is this pattern I believe should be applied.

We have the libpod call, we can add the parameters we might need, no need to do additional filtering in the backend, that would have side effects, each call is unique for each caller, we don't share or reuse them anyway

feloy:
> > If the _skip_ information (timestamp or length to skip, it's the same problem) is stored in the backend, the caller can only ask for the container's logs. And the same filtering will be done in the backend, no matter which caller asked the logs.
> 
> There is probably a missunderstanding here, each caller would have to specify its own parameters, and the backend would not have to do any filtering, it just pass the parameters to the libpod request.
> 

Yes, you raised two points:
- having the skip information stored in the backend instead of the frontend
- storing the timestamp of when the user clicked Clear, to be able to pass this value to libpod (instead of storing the length of the logs at the time the user clicked Clear)

I feel that you discussing the second one, while I'm discussing the first one

axel7083:
> having the skip information stored in the backend instead of the frontend

yes I better see you point, I don't have much idea how to solve that then, storing in the frontend still feel wrong, because we save something when the user is no longer on the terminal...

I feel like the terminals state should not be kept outside the terminal, and only something similar to vscode, jetbrain etc. where terminals are always here, would be logic enough to keep this information, but this is not related to the PR, yet I question the need to add yet another store in the renderer

feloy:
For the second point, I agree that using `since` is better. I was afraid of time sync between container and Podman Desktop, but as the `since` param also accepts a duration (`42m`to ask for logs since 42m ago), this removes the problem of time sync. I'll do the change (and store the time the user asked the clear), so only the new logs are transferred from libpod to frontend.

feloy:
@vancura what is your opinion about the changes on the Clear button?

This PR makes the button change to a Restore button when the user presses Clear (see video demo at 0:30), and the button is changed again to Clear when new logs are displayed, so the user can clear the logs again.

One problem is that it is difficult to restore the logs when there is an uninterrupted flow of logs (see beginning of video demo).

Another solution would be to display the Restore button alongside the Clear button, when the logs have been cleared at least once.

vancura:
I like the concept, just a couple of ideas:

You're right that these two buttons should be side-by-side. But don't hide the button if it's unavailable, that would change their positions (interactive elements must stay at the same position all the time). Disabling (and making them semi-transparent) would solve the problem.

Additionally:

## Button placement and states

- Place the `Clear` button first, and the `Restore` button second.
- Align both icons at the bottom right of the container, since both actions are potentially destructive. We'll keep the top right area ready for standard buttons like copy to clipboard (not the scope of this Issue).
- Make sure that the `Clear` button is disabled when there is nothing to clear. The video makes it hard to see if you have already implemented this.
- For disabled states, use both semi-transparency and reduced contrast to make it more obvious (following WCAG guidelines).


## Icons and sizing

- I am not a fan of the icons (`faEraser` and `faTrashRestore`). Can you please use `faTrash` and `faUndo`? They're more universally recognized.
- Please ensure the size is the same as in the toolbar; the icons now look huge.
- Ensure sufficient touch target size (minimum `44pt`) for touch interfaces (I know we're on desktops, but you never know when we expand to tablets; better safe than sorry later).


## Accessibility and keyboard support

- Make sure it's possible to use both actions without a pointing device. If we have these keyboard shortcuts available, I'd suggest `Ctrl+L` / `⌘L` for Clear (common in terminals) and `Ctrl+Shift+L` / `⇧⌘L` for Restore. If not, let's discuss what we can do.
- Add proper `tabIndex`es and ARIA labels.
- Include `title` attributes for hover tooltips that also document the keyboard shortcuts. Use the format I described above (symbols, not words like CMD); I will work on the review of macOS keyboard shortcut strings later, but why not make sure this follows the HIG?
- Add `aria-pressed` state for toggle-like behavior.
- Consider an `aria-live` region to announce state changes to screen readers.
- Ensure focus remains on the active button after the action.


## User feedback and safety

- Consider adding a confirmation dialog for the `Clear` action since it's destructive. It depends on how destructive we consider this action.
- Add visual feedback (e.g., brief animation or toast notification) when clearing/restoring.
- Given the challenge with continuous log flow making restoration difficult (as shown in the video), consider adding a timestamp marker showing `Logs cleared at [time]` in the Terminal.
- Perhaps we could add a `Pause` button in the future to stop log flow temporarily for easier restoration, but that's probably out of scope for this PR.


Thank you for your work! It's already pretty good :)

/c @MariaLeonova

feloy:
closing for the moment, as the solution I had in mind does not work

feloy:
Thanks @vancura for your feedback. I was focused on the technical solution to filter logs efficiently (see #12626), and missed your response. I'm closing this issue for the moment as the technical solution I had in mind for filtering is not working. I'll need to come back to this issue when I find another solution, and I'll be happy to use your feedback.

vancura:
I am always ready and happy to nitpick! You know where to find me :)

PR #12625, opened 2025-05-26
chore(workflows): set permission for downloads-count.yaml

axel7083:
### What does this PR do?

The downloads count only read and publish to segment using a secret. It does not require any other permission.

### DIFF
--- a/.github/workflows/downloads-count.yaml
+++ b/.github/workflows/downloads-count.yaml
@@ -49,8 +49,8 @@ jobs:
             let totalLinux = 0;
 
             // count the total number of downloads for each release for mac,windows,linux assests
-            releases.forEach((release) => { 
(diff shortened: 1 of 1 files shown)


.github/workflows/downloads-count.yaml
```diff
@@ -68,7 +71,7 @@ jobs:
             core.exportVariable('totalMac', totalMac);
             core.exportVariable('totalWin', totalWin);
             core.exportVariable('totalLinux', totalLinux);
-        
+
```
axel7083, line 74: my linter forcing me

benoitf: separate in two commits then, one with formatter fix and one with the change

benoitf (approved):

PR #12624, opened 2025-05-26
feat: make window.logsContainer cancellable

feloy:
Signed-off-by: Philippe Martin <phmartin@redhat.com>

### What does this PR do?

Make window.logsContainer cancellable

### What issues does this PR fix or reference?

Needed for #11782 

### How to test this PR?

Unit tests included

- [x] Tests are covering the bug fix or the new feature

PR #12626, opened 2025-05-26
feat: add sinceDurationInSeconds to window.logsContainer

feloy:

### What does this PR do?

Add sinceDurationInSeconds to window.logsContainer, to be able to get most recent containetr's logs 

Needed for #11782

### How to test this PR?

Unit tests included

### DIFF
--- a/packages/main/src/plugin/container-registry.ts
+++ b/packages/main/src/plugin/container-registry.ts
@@ -1679,17 +1679,22 @@ export class ContainerProviderRegistry {
     id: string;
     callback: (name: string, data: string) => void;
     abortController?: AbortController;
+    sinceDurationInSeconds?: number;
   }): Promise<void> {
     let telemetryOptions = {};
     let firstMessage = true;
     const container = this.getMatchingContainer(logsParams.engineId, logsParams.id);
+    const apiLogsParams: Dockerode.ContainerLogsOptions & { follow: true } = {
+      follow: true,
+      stdout: true,
+      stderr: true,
+      abortSignal: logsParams.abortController?.signal,
+    };
+    if (logsParams.sinceDurationInSeconds) {
+      apiLogsParams.since = `${logsParams.sinceDurationInSeconds}s`;
+    }
     container
-      .logs({
-        follow: true,
-        stdout: true,
-        stderr: true,
-        abortSignal: logsParams.abortController?.signal,
-      })
+      .logs(apiLogsParams)
       .then(containerStream => {
         containerStream.on('end', () => {
           logsParams.callback('end', '');
--- a/packages/main/src/plugin/container-registry.spec.ts
+++ b/packages/main/src/plugin/container-registry.spec.ts
@@ -30,7 +30,7 @@ import { http, HttpResponse } from 'msw';
 import { setupServer, type SetupServerApi } from 'msw/node';
 import type { Headers, PackOptions } from 'tar-fs';
 import * as tarstream from 'tar-stream';
-import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { afterEach, assert, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import type { ApiSenderType } from '/@/plugin/api.js';
 import type { Certificates } from '/@/plugin/certificates.js';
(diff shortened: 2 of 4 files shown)


packages/main/src/plugin/container-registry.ts
```diff
@@ -1679,17 +1679,22 @@ export class ContainerProviderRegistry {
     id: string;
     callback: (name: string, data: string) => void;
     abortController?: AbortController;
+    sinceDurationInSeconds?: number;
```
axel7083, line 1682: why using such complex name, just use since, and document the property, if we don't plan to add since in ms, let's just use the same name as the libpod api

packages/main/src/plugin/container-registry.ts
```diff
@@ -1679,17 +1679,22 @@ export class ContainerProviderRegistry {
...
     let firstMessage = true;
     const container = this.getMatchingContainer(logsParams.engineId, logsParams.id);
+    const apiLogsParams: Dockerode.ContainerLogsOptions & { follow: true } = {
+      follow: true,
+      stdout: true,
+      stderr: true,
+      abortSignal: logsParams.abortController?.signal,
+    };
+    if (logsParams.sinceDurationInSeconds) {
+      apiLogsParams.since = `${logsParams.sinceDurationInSeconds}s`;
```
axel7083, line 1694: follow-up of https://github.com/podman-desktop/podman-desktop/pull/12626/files#r2106896446, we should not contraint to seconds, why can't we just expose the since like other parameters? why forcing the input to be a number in seconds?

axel7083: > Only return logs since this time, as a UNIX timestamp

From https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerLogsLibpod I should be able to pass a unix timestamp?

benoitf: it's not Libpod but compat API (but it's also a unix timestamp)
https://docs.podman.io/en/latest/_static/api.html#tag/containers-(compat)/operation/ContainerLogs

feloy: The parameter `since` of libpod accepts `number | string | undefined` and I cannot find a complete documentation of the different values it can accept.

1. With this PR, I propose to disambiguate, by accepting only one kind of value, a duration, and pass it as a string like shown in the `podman logs` example (in seconds for the PR, instead of in minutes in the example):

```
Examples:
  podman logs --follow=true --since 10m ctrID
```

2. I'm not comfortable passing a unix timestamp, because it is difficult to know if the caller and the libpod will be always synchronised. Passing a duration resolves this problem of synchronisation

benoitf:
probably we would need sooner than later also `until` and `timestamps` in addition to since
https://docs.docker.com/reference/api/engine/version/v1.49/#tag/Container/operation/ContainerLogs

and for sure the value provided is not correct, it's not seconds

packages/main/src/plugin/container-registry.ts
benoitf: the duration is not what the API is expecting, let's keep it close to the API

also what synchronization, b/c the API call is remote so in any case you need to provide the timestamp from Podman Desktop

feloy: The compat API doc for `since` looks incomplete to me when indicating only the timestamp.

The doc for the `docker container logs --since` specifies (https://docs.docker.com/reference/cli/docker/container/logs/):

```
The --since option shows only the container logs generated after a given date. You can specify the date as an RFC 3339 date, a UNIX timestamp, or a Go duration string (e.g. 1m30s, 3h)
```

And `podman container logs --since` (https://docs.podman.io/en/v5.0.1/markdown/podman-logs.1.html#since-timestamp):

```
Show logs since TIMESTAMP. The --since option can be Unix timestamps, date formatted timestamps, or Go duration strings (e.g. 10m, 1h30m) computed relative to the client machine’s time. Supported formats for date formatted time stamps include RFC3339Nano, RFC3339, 2006-01-02T15:04:05, 2006-01-02T15:04:05.999999999, 2006-01-02Z07:00, and 2006-01-02.


```

I know that these docs are not 100% related to the API, but I tested in https://github.com/podman-desktop/podman-desktop/pull/12624 the format `since=70s` like proposed in this PR, and this format is correctly accepted.

feloy: tested with Docker provider, this syntax is not recognized in this case (`strconv.ParseInt('...s'): invalid syntax

axel7083: > The compat API doc for since looks incomplete to me when indicating only the timestamp.

I tried providing `since` and `until` both with compact and libpod api with a podman machine.

image

Works fine to me

benoitf: I tried with `docker` CLI, if I use `--since 42m` it's providing the unix timestamp on the REST API


```
GET /v1.41/containers/ac3295a7b693b304dc1d2689608023e35f1307cd6d86452d7cea5b3cf03035a8/logs?since=1748258328&stderr=1&stdout=1&tail=all HTTP/1.1\r
```

and using `libpod` (not compat API)
```
GET /v5.5.0/libpod/containers/727c9186b31ec98bca45d0c246dbdeaa95580a8e57b5b438b653eac932cf476f/logs?follow=false&since=2025-05-26T13%3A21%3A44%2B02%3A00&stderr=true&stdout=true&tail=-1&timestamps=false&until=0001-01-01T00%3A00%3A00Z HTTP/1.1\r
```


so CLI uses minutes or seconds but when it's executing calls, it's giving an absolute date

Issue #12591, opened 2025-05-22
configuration-changed event is triggered frequently


feloy:
### Bug description

Investigating in #11882, I have seen that the `configuration-changed` event is triggered frequently by the podman extension to update the resources usage of the machines (https://github.com/podman-desktop/podman-desktop/blob/e4586a3537a05c3e27d1818670120cc968970994/extensions/podman/packages/extension/src/extension.ts#L634-L639)

Some stores are registering this `configuration-changed` event, specifically the `recommendedRegistries` store for #11882, which tries to get the catalog every time. When the catalog can be fetched a first time, it is put in cache and the cache is used for every future `configuration-changed` event, but when the catalog is never read (because of connection error in case of #11882), the connection is tried again for every `configuration-changed` event

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

Be sure to have at least one podman machine, and the podman extension active.

Make the download of the catalog fail (for example by replacing the lines https://github.com/podman-desktop/podman-desktop/blob/e4586a3537a05c3e27d1818670120cc968970994/packages/main/src/plugin/extension/catalog/extensions-catalog.ts#L83-L87 with `throw new Error('fake error')`)

and check in the logs, `Unable to fetch the available extensions: fake error` will be displayed every x seconds.

### Relevant log output

```shell

```

feloy:
As discussed in team meeting, it would be helpful to optimize both:
- support fine-grained configuration events
- fetching the catalog when it is not available (cache the error, and leverage network config events provided by electron)

odockal:
@feloy It feels more like enhancement to me. Feel free to change it back if you disagree.

PR #12636, opened 2025-05-26
feat: support fine-grained window events for EventStore

feloy:

### What does this PR do?

Support fine-grained window events for EventStore

Part of #12591 

### How to test this PR?

Unit tests included 

### DIFF
--- a/packages/main/src/plugin/configuration-impl.ts
+++ b/packages/main/src/plugin/configuration-impl.ts
@@ -21,6 +21,7 @@ import type * as containerDesktopAPI from '@podman-desktop/api';
 import { CONFIGURATION_DEFAULT_SCOPE } from '/@api/configuration/constants.js';
 
 import type { ApiSenderType } from './api.js';
+import type { IConfigurationChangeEvent } from './configuration-registry.js';
 
 /**
  * Local view of the configuration values for a given scope
@@ -68,21 +69,26 @@ export class ConfigurationImpl implements containerDesktopAPI.Configuration {
 
   async update(section: string, value: unknown): Promise<void> {
     const localKey = this.getLocalKey(section);
-
     // now look if we have this value
     const localView = this.getLocalView();
+
+    const eventArg: IConfigurationChangeEvent = {
+      key: localKey,
+      value,
+      scope: this.scope,
+    };
 
     // remove the value if undefined
     if (value === undefined) {
       if (localView[localKey]) {
         delete localView[section];
         delete this[localKey];
-        this.apiSender.send('configuration-changed');
+        this.apiSender.send('configuration-changed', eventArg);
       }
     } else {
       localView[localKey] = value;
       this[section] = value;
-      this.apiSender.send('configuration-changed');
+      this.apiSender.send('configuration-changed', eventArg);
     }
     // call only for default scope to save
     this.updateCallback(section, this.scope);
--- a/packages/main/src/plugin/configuration-impl.ts
+++ b/packages/main/src/plugin/configuration-impl.ts
@@ -72,10 +72,9 @@ export class ConfigurationImpl implements containerDesktopAPI.Configuration {
     // now look if we have this value
     const localView = this.getLocalView();
 
-    const eventArg: IConfigurationChangeEvent = {
+    const eventArg: Omit<IConfigurationChangeEvent, 'scope'> = {
       key: localKey,
       value,
-      scope: this.scope,
     };
 
     // remove the value if undefined
--- a/packages/main/src/plugin/configuration-impl.ts
+++ b/packages/main/src/plugin/configuration-impl.ts
@@ -72,7 +72,7 @@ export class ConfigurationImpl implements containerDesktopAPI.Configuration {
     // now look if we have this value
     const localView = this.getLocalView();
 
-    const eventArg: Omit<IConfigurationChangeEvent, 'scope'> = {
+    const configurationChangedEvent: Omit<IConfigurationChangeEvent, 'scope'> = {
       key: localKey,
       value,
     };
@@ -82,12 +82,12 @@ export class ConfigurationImpl implements containerDesktopAPI.Configuration {
       if (localView[localKey]) {
         delete localView[section];
         delete this[localKey];
-        this.apiSender.send('configuration-changed', eventArg);
+        this.apiSender.send('configuration-changed', configurationChangedEvent);
       }
     } else {
       localView[localKey] = value;
       this[section] = value;
-      this.apiSender.send('configuration-changed', eventArg);
+      this.apiSender.send('configuration-changed', configurationChangedEvent);
     }
     // call only for default scope to save
     this.updateCallback(section, this.scope);
--- a/packages/renderer/src/stores/event-store.ts
+++ b/packages/renderer/src/stores/event-store.ts
@@ -93,7 +93,7 @@ export class EventStore<T> {
 
     // The list of window events and listeners to listen for to trigger an update, for example:
     // Window event: 'extension-started'
-    // Window event with key: 'configuration-changed:extensions.ignoreRecommendations'
+    // Window event with keys: 'configuration-changed:extensions.ignoreRecommendations,ignoreBannerRecommendations'
     // Window listener: 'extensions-already-started''
     private windowEvents: string[],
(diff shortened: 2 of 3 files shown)


packages/renderer/src/stores/event-store.ts
```diff
@@ -20,6 +20,7 @@ import humanizeDuration from 'humanize-duration';
 import type { Component } from 'svelte';
 import type { Writable } from 'svelte/store';
 
+import { isConfigurationChangedArg } from '../../../main/src/plugin/configuration-impl';
```
axel7083: this is a problem; we cannot import function from the main package into the renderer

feloy: @benoitf What is the recommended way to do this? 

I can see this pattern used a few times (for example https://github.com/podman-desktop/podman-desktop/blob/f99f2ab1c5ea845ad0a7101a837b7b8022ca2184/packages/renderer/src/stores/recommendedRegistries.ts#L21) and there is no `shared` package

feloy: @benoitf what would be the appropriate pattern here? I would like to enforce the type when passing the value from `main`, and would like to check the type from the `renderer`. Should I separate the interface definition (in main) and the `isConfigurationChangedArg` (in renderer), or is there a `shared` module where we can place both of them?

benoitf: you can call a function from renderer to main, but it should be executed on the main side 'exposeInMainWorld'

benoitf: and not be like a helper function you import directly to the renderer part

feloy: I don't think this method applies here, as the function is used to narrow the type by returning a type predicate, but the `exposeInMainWorld` expects to use Promises, which seem not compatible (we cannot have `Promise<value is MyType>`)

axel7083: You can move the function to the renderer package, if this need to be in both, should be duplicated

feloy: Thanks. This is what I've finally done, see latest commit

packages/main/src/plugin/configuration-impl.ts
```diff
@@ -22,6 +22,14 @@ import { CONFIGURATION_DEFAULT_SCOPE } from '/@api/configuration/constants.js';
 
 import type { ApiSenderType } from './api.js';
 
+export interface ConfigurationChangedArg {
```
axel7083: Can't we reuse `IConfigurationChangeEvent` or at least copy it ? would be compliant with our extensions get notified

https://github.com/podman-desktop/podman-desktop/blob/80973f33fc6ea17bcf3be847b69b3b836f74cf5f/packages/main/src/plugin/configuration-registry.ts#L45-L49

Moreover, if this is something shared between the main and renderer, should be inside `packages/api`

packages/main/src/plugin/configuration-impl.ts
```diff
@@ -68,21 +72,24 @@ export class ConfigurationImpl implements containerDesktopAPI.Configuration {
 
   async update(section: string, value: unknown): Promise<void> {
     const localKey = this.getLocalKey(section);
-
     // now look if we have this value
     const localView = this.getLocalView();
 
+    const eventArg: ConfigurationChangedArg = {
```
benoitf: I'm not sure to follow the pattern `xxxArg`

AFAIK we're sending an event there, so `ConfigurationChangeEvent` might be more appropriated ?

benoitf: I would expect the value be part of the event as well ?

packages/renderer/src/stores/event-store.ts
```diff
@@ -263,6 +265,20 @@ export class EventStore<T> {
...
+          isConfigurationChangedArg(args) &&
+          this.configurationSections.includes(args.section)
+        ) {
+          update('configuration-changed', [args]).catch((error: unknown) => {
+            console.error(`Failed to update ${this.name}`, error);
+          });
+        }
+      });
+    }
+
```
benoitf: I would favor to reuse an existing pattern from `windowEvents`

so instead of `configuration-changed` I can specify for example `configuration-changed:<key>`


`configuration-changed:extensions.ignoreRecommendations`

benoitf: so if other functions have keys or specific value it could be more generic rather than adding a new entry just to handle a specific event

it would be `NameOfEvent:Key`

feloy: I'm not sure what you have in mind for matching the `key`?

Should the apiSender send `('NameOfEvent:Key')`? Or should it send `('NameOfEvent', { key: 'Key', ... })`?

benoitf: I think the later one

```
'NameOfEvent', { key: 'Key', ... }
```

if I take

```
 this.apiSender.send('configuration-changed', eventArg);
```

if there is a `eventKey` or `key` property inside the event, then I can use nameOfTheEvent:key to be triggered only if key is received

feloy: OK, I like this pattern, I will implement it.
Thanks

packages/main/src/plugin/configuration-impl.ts
feloy: The problem I can see with `ConfigurationChangeEvent` is that is contains a mandatory scope, which can be a complex structure (`ContainerProviderConnection` for example), which cannot be serialised to be sent to the renderer. The best solution I can think of is to send an empty string as scope (and document it)

benoitf: use Omit ? https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

packages/renderer/src/stores/event-store.ts
```diff
@@ -255,11 +267,20 @@ export class EventStore<T> {
...
+          if (args.length === 1 && isWithStringKey(args[0]) && key === args[0].key) {
+            update(eventNameWithOptionalKey, args).catch((error: unknown) => {
+              console.error(`Failed to update ${this.name}`, error);
+            });
+          }
+        } else {
+          update(eventNameWithOptionalKey, args).catch((error: unknown) => {
+            console.error(`Failed to update ${this.name}`, error);
+          });
+        }
```
feloy: Here, we are potentially registering several listeners for the same eventName (if we pass `eventName:key1`, `eventName:key2`, `eventName:key3` to `windowEvents`).

If this is problematic, I can regroup all window events with a common eventName, and start a single listener

packages/main/src/plugin/configuration-impl.ts
```diff
@@ -68,21 +69,25 @@ export class ConfigurationImpl implements containerDesktopAPI.Configuration {
 
   async update(section: string, value: unknown): Promise<void> {
     const localKey = this.getLocalKey(section);
-
     // now look if we have this value
     const localView = this.getLocalView();
 
+    const eventArg: Omit<IConfigurationChangeEvent, 'scope'> = {
```
benoitf: maybe rename `eventArg` by `configurationChangedEvent`

packages/renderer/src/stores/event-store.ts
```diff
@@ -43,6 +43,14 @@ interface EventStoreInfoEvent {
   humanDuration?: number;
 }
 
+interface WithStringKey {
```
benoitf:
```suggestion
interface KeyEvent {
```

(or EventKey)

packages/renderer/src/stores/event-store.ts
benoitf: could use eventName:key1,key2,key3 ? (assuming a key can't have any comma)

benoitf (approved):
LGTM except the variable name / interface name that I might find confusing but others could say it's ok so not blocking

non-blocking: for several keys, it could use comma-separated list but could be follow-up

feloy:
> could use eventName:key1,key2,key3 ? (assuming a key can't have any comma)

I added the support for this

feloy:
@axel7083 @benoitf could you please take another look?

benoitf (approved):
thanks for the update

PR #12641, opened 2025-05-27
chore: add fontawesome to dependencies

gastoner:
### What does this PR do?
Adds fontawesome to dependencies

### How to test this PR?

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,6 +180,7 @@
   },
   "dependencies": {
     "@docker/extension-api-client-types": "0.3.4",
+    "@fortawesome/free-solid-svg-icons": "^6.7.2",
     "@kubernetes/client-node": "^1.1.2",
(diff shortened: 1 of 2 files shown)


axel7083 (commented):
Shouldn't be added to https://github.com/podman-desktop/podman-desktop/blob/main/packages/renderer/package.json instead of the package at the root?

gastoner:
> Shouldn't be added to https://github.com/podman-desktop/podman-desktop/blob/main/packages/renderer/package.json instead of the package at the root?

"@fortawesome/free-solid-svg-icons": "^6.7.2", it is already there, but since I'm importing it in backend it is not there :( 
https://github.com/podman-desktop/podman-desktop/blob/main/packages/renderer/package.json#L18

gastoner:
Closing since there are other ways how to do this

PR #12646, opened 2025-05-27
feat: add button to start all containers in bulk

### LINKED ISSUE #12312: Desktop: button to start all containers

I use kind, and it needs to start 3 containers, too bad we need to start them each time

Add a button "start all" which start the selected containers (we can only delete them today)

_No response_

### Additional context

_No response_

MarsKubeX:
### What does this PR do?

This PR adds a new feature in order to have the possibility to run containers in bulk. If there are containers already running selected, the button will just apply to those who are not running.

### Screenshot / video of UI
Captura de pantalla 2025-05-27 a las 13 33 08

Captura de pantalla 2025-05-27 a las 13 30 02

closes #12312 

### How to test this PR?

1. Run podman desktop
2. Go to containers sections
3. Create some pods and containers
4. Select multiple containers, running and not running.
5. Click in the bulk run button at the top of the section.
6. After clicking those containers and pots that were not running should run.

### DIFF
--- a/packages/renderer/src/lib/container/ContainerList.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerList.spec.ts
@@ -877,3 +877,93 @@ test('Expect user confirmation to pop up when preferences require', async () => 
   expect(window.showMessageBox).toHaveBeenCalledTimes(2);
   await vi.waitFor(() => expect(window.deleteContainer).toHaveBeenCalled());
 });
+
+test('Try to run pods in bulk', async () => {
+  vi.mocked(window.startPod).mockClear();
+  vi.mocked(window.startContainer).mockClear();
+  vi.mocked(window.listContainers).mockResolvedValue([]);
+
+  window.dispatchEvent(new CustomEvent('extensions-already-started'));
+  window.dispatchEvent(new CustomEvent('provider-lifecycle-change'));
+  window.dispatchEvent(new CustomEvent('tray:update-provider'));
+
+  // wait for the store to be cleared
+  while (get(containersInfos).length !== 0) {
+    await new Promise(resolve => setTimeout(resolve, 250));
+  }
+
+  const podId = 'pod-id3';
+  const containerId = 'sha256:56789012345';
+
+  const singleContainer = {
+    Id: containerId,
+    Image: 'sha256:567',
+    Names: ['foo'],
+    Status: 'Stopped',
+    engineId: 'podman',
+    engineName: 'podman',
+    ImageID: 'dummy-image-id',
+  };
+
+  // one single container and a container as part of a pod
+  const mockedContainers = [
+    singleContainer as ContainerInfo,
+    {
+      Id: 'sha256:7897891234567890123',
+      Image: 'sha256:345',
+      Names: ['container-in-pod'],
+      Status: 'Stopped',
+      pod: {
+        name: 'my-pod3',
+        id: podId,
+        status: 'Stopped',
+      },
+      engineId: 'podman',
+      engineName: 'podman',
+      ImageID: 'dummy-image-id',
+    } as ContainerInfo,
+  ];
+
+  vi.mocked(window.listContainers).mockResolvedValue(mockedContainers);
+
+  window.dispatchEvent(new CustomEvent('extensions-already-started'));
+  window.dispatchEvent(new CustomEvent('provider-lifecycle-change'));
+  window.dispatchEvent(new CustomEvent('tray:update-provider'));
+
+  // wait until the store is populated
+  while (get(containersInfos).length === 0) {
+    await new Promise(resolve => setTimeout(resolve, 250));
+  }
+
+  await waitRender({});
+
+  // select the pod checkbox
+  const checkboxes = screen.getAllByRole('checkbox', { name: 'Toggle container' });
+  expect(checkboxes[0]).toBeInTheDocument();
+  expect(checkboxes[1]).toBeInTheDocument();
+  await fireEvent.click(checkboxes[0]);
+  await fireEvent.click(checkboxes[1]);
+
+  // click on the bulk run button
+  const runBulkButton = screen.getByRole('button', { name: 'Run selected containers and pods' });
+  expect(runBulkButton).toBeInTheDocument();
+  await fireEvent.click(runBulkButton);
+
+  // wait until startPodMock is called
+  while (vi.mocked(window.startPod).mock.calls.length === 0) {
+    await new Promise(resolve => setTimeout(resolve, 100));
+  }
+
+  // wait until startContainerMock is called
+  while (vi.mocked(window.startContainer).mock.calls.length === 0) {
+    await new Promise(resolve => setTimeout(resolve, 100));
+  }
+
+  // expect that the pod is running
+  expect(window.startPod).toHaveBeenCalledWith('podman', podId);
+  expect(window.startPod).toHaveBeenCalledOnce();
+
+  // expect that the container is running
+  expect(window.startContainer).toHaveBeenCalledWith('podman', containerId);
+  expect(window.startContainer).toHaveBeenCalledOnce();
+});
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -1,10 +1,11 @@
 <script lang="ts">
-import { faPlusCircle, faTrash } from '@fortawesome/free-solid-svg-icons';
+import { faPlay, faPlusCircle, faTrash } from '@fortawesome/free-solid-svg-icons';
 import {
   Button,
   FilteredEmptyScreen,
   NavPage,
   Table,
+  type Table as TableComponentType,
```
axel7083: question: I am not sure to understand why we need to import the type? Why can't we just use the Table import L7?

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -130,6 +137,65 @@ async function deleteSelectedContainers(): Promise<void> {
...
+  if (selectedContainers.length > 0) {
+    await Promise.all(
+      selectedContainers.map(async container => {
+        if (container.state === 'RUNNING') {
+          return; // skip already running containers
+        }
+        container.actionInProgress = true;
+        // reset error when starting task
+        container.actionError = '';
+        containerGroups = [...containerGroups];
```
axel7083, line 182: question: why do you need to make a shallow copy?

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -300,7 +366,7 @@ function setStoppedFilter(): void {
 }
 
 let selectedItemsNumber: number;
-let table: Table;
+let table: TableComponentType;
```
axel7083: question: follow-up to my first remark, I don't think this is necessary

packages/renderer/src/lib/container/ContainerList.svelte
axel7083: Maybe we should migrate the component to svelte5

packages/renderer/src/lib/container/ContainerList.svelte
MarsKubeX: There's a reference to the table component defined within ContainerList. To ensure it's correctly typed, I added a specific type definition to differentiate it from the regular component import.

axel7083: The component is the type itself, I don't think there is a difference, especially this is not related to this PR.

packages/renderer/src/lib/container/ContainerList.svelte
MarsKubeX: If I'm not wrong, this will make the containerGroups array reactive within the Table component, so this last one can update its reference to this object. From what I understand from Svelte, it only tracks changes when the array reference changes, not when you mutate its items. The shallow copy ensures this reactivity.

packages/renderer/src/lib/container/ContainerList.svelte
MarsKubeX: Removed reference to Table component, so this type has been removed too.

packages/renderer/src/lib/container/ContainerList.svelte
MarsKubeX: Removed

axel7083 (approved):

packages/renderer/src/lib/container/ContainerList.spec.ts
```diff
@@ -877,3 +877,93 @@ test('Expect user confirmation to pop up when preferences require', async () =>
   expect(window.showMessageBox).toHaveBeenCalledTimes(2);
   await vi.waitFor(() => expect(window.deleteContainer).toHaveBeenCalled());
 });
+
+test('Try to run pods in bulk', async () => {
+  vi.mocked(window.startPod).mockClear();
+  vi.mocked(window.startContainer).mockClear();
```
benoitf, line 883: question: why do we need these two mockClear instructions ? we're at the beginning of the test so everything should be "cleared"

if not, it means that beforeAll/beforeEach is not properly setup and should be fixed (resetAllMocks /  restoreAllMocks())

packages/renderer/src/lib/container/ContainerList.spec.ts
```diff
@@ -877,3 +877,93 @@ test('Expect user confirmation to pop up when preferences require', async () =>
...
+
+  window.dispatchEvent(new CustomEvent('extensions-already-started'));
+  window.dispatchEvent(new CustomEvent('provider-lifecycle-change'));
+  window.dispatchEvent(new CustomEvent('tray:update-provider'));
+
+  // wait for the store to be cleared
+  while (get(containersInfos).length !== 0) {
+    await new Promise(resolve => setTimeout(resolve, 250));
+  }
+
```
benoitf, line 894: use waitFor

packages/renderer/src/lib/container/ContainerList.spec.ts
```diff
@@ -877,3 +877,93 @@ test('Expect user confirmation to pop up when preferences require', async () =>
...
+  vi.mocked(window.listContainers).mockResolvedValue(mockedContainers);
+
+  window.dispatchEvent(new CustomEvent('extensions-already-started'));
+  window.dispatchEvent(new CustomEvent('provider-lifecycle-change'));
+  window.dispatchEvent(new CustomEvent('tray:update-provider'));
+
+  // wait until the store is populated
+  while (get(containersInfos).length === 0) {
+    await new Promise(resolve => setTimeout(resolve, 250));
+  }
```
benoitf, line 936: use waitFor

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -130,6 +136,67 @@ async function deleteSelectedContainers(): Promise<void> {
...
+        }
+        container.actionInProgress = true;
+        // reset error when starting task
+        container.actionError = '';
+        containerGroups = [...containerGroups];
+        try {
+          await window.startContainer(container.engineId, container.id);
+          container.state = 'RUNNING';
+        } catch (e) {
+          console.log('error while runnings container', e);
```
benoitf, line 187: do we need this log as the error is wrapped into the actionError ?

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -427,7 +501,6 @@ $: containersAndGroups = containerGroups.map(group =>
     <div class="flex min-w-full h-full">
       <Table
         kind="container"
-        bind:this={table}
```
benoitf, line 430: why this line is being removed ?

benoitf:
it looks it has been merged without replying my questions ?

packages/renderer/src/lib/container/ContainerList.svelte
MarsKubeX: Checking the startContainer definition, I see that the error is just being thrown, so I think is not a bad idea to handle the error here with the log.

packages/renderer/src/lib/container/ContainerList.svelte
MarsKubeX: This reference wasn't being used, and as I had some linter issues with it I removed it. As is not PR related, next time I'll make sure to create another issue

packages/renderer/src/lib/container/ContainerList.spec.ts
MarsKubeX: Added an issue to refactor this. #12713

packages/renderer/src/lib/container/ContainerList.spec.ts
MarsKubeX: Added an issue to refactor this. #12711

packages/renderer/src/lib/container/ContainerList.spec.ts
MarsKubeX: Added an issue to refactor this. #12711

MarsKubeX:
> it looks it has been merged without replying my questions ?

@benoitf Sorry, I accidentally merged it without checking those comments. I have Already created some issues to tackle them.

PR #12648, opened 2025-05-27
test(e2e): redhat extension pack installation test case

### LINKED ISSUE #11551: Automate the installation of Red Hat Extension Pack extension

We should cover the scenario of installing the RH extension pack in podman-desktop and verify presence and state of all included extensions.

Test would install the RH ext. pack extension, waits until it finishes, and the searches for installed extensions - all should be present and in active state.

Automate the scenario to be run on all platforms.

_No response_

### Additional context

_No response_

danivilla9:
### What does this PR do?
Adds a new test case that installs the 'RedHat Extension Pack' extension and checks that all expected extensions are installed alongside it.

#11551 

### How to test this PR?
Run `pnpm test:e2e:all`

### DIFF
--- a/tests/playwright/src/specs/rh-extension-pack-installation.spec.ts
+++ b/tests/playwright/src/specs/rh-extension-pack-installation.spec.ts
@@ -1,0 +1,154 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Page } from '@playwright/test';
+import { expect as playExpect, test } from '@playwright/test';
+
+import { extensionsPackExtension, extensionsRHPackList } from '../model/core/extensions';
+import { ExtensionCatalogCardPage } from '../model/pages/extension-catalog-card-page';
+import { ExtensionsPage } from '../model/pages/extensions-page';
+import { WelcomePage } from '../model/pages/welcome-page';
+import { NavigationBar } from '../model/workbench/navigation';
+import { Runner } from '../runner/podman-desktop-runner';
+
+const DISABLED = 'DISABLED';
+const ACTIVE = 'ACTIVE';
+const DOWNLOADABLE = 'DOWNLOADABLE';
+
+const RH_EXTENSION_PACK_NAME = extensionsPackExtension.extensionName;
+const RH_EXTENSION_PACK_FULL_NAME = extensionsPackExtension.extensionFullName;
+const RH_EXTENSION_PACK_LABEL = extensionsPackExtension.extensionLabel;
+const RH_EXTENSION_PACK_FULL_LABEL = extensionsPackExtension.extensionFullLabel;
+const RH_EXTENSION_PACK_INSTALL_BUTTON_LABEL = `Install ${RH_EXTENSION_PACK_FULL_LABEL}`;
+
+let pdRunner: Runner;
+let page: Page;
+let navigationBar: NavigationBar;
+
+test.describe
+  .serial(`Red Hat Extension pack extension installation`, () => {
+    test.beforeAll(async () => {
+      pdRunner = await Runner.getInstance();
+      page = pdRunner.getPage();
+      pdRunner.setVideoAndTraceName('rh-pack-extension-installation-e2e');
+
+      const welcomePage = new WelcomePage(page);
+      await welcomePage.handleWelcomePage(true);
+      navigationBar = new NavigationBar(page);
+    });
+
+    test.afterAll(async () => {
+      await pdRunner.close();
+    });
+
+    test('Install RH Pack extension through Extensions Catalog', async () => {
+      test.setTimeout(200000);
+
+      await navigationBar.openExtensions();
+      const extensionsPage = new ExtensionsPage(page);
+      await playExpect(extensionsPage.heading).toBeVisible();
+
+      await extensionsPage.openCatalogTab();
+      const extensionCatalog = new ExtensionCatalogCardPage(page, RH_EXTENSION_PACK_NAME);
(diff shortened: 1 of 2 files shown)


tests/playwright/src/specs/rh-extension-pack-installation.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+      await navigationBar.openExtensions();
+      const extensionsPage = new ExtensionsPage(page);
+      await playExpect(extensionsPage.heading).toBeVisible();
+
+      await extensionsPage.openCatalogTab();
+      const extensionCatalog = new ExtensionCatalogCardPage(page, RH_EXTENSION_PACK_NAME);
+      await playExpect(extensionCatalog.parent).toBeVisible();
+
+      await playExpect.poll(async () => await extensionCatalog.isInstalled()).toBeFalsy();
+      await extensionCatalog.install(180000);
```
odockal: use `180_000` pls

tests/playwright/src/specs/rh-extension-pack-installation.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+      await welcomePage.handleWelcomePage(true);
+      navigationBar = new NavigationBar(page);
+    });
+
+    test.afterAll(async () => {
+      await pdRunner.close();
+    });
+
+    test('Install RH Pack extension through Extensions Catalog', async () => {
+      test.setTimeout(200000);
```
odockal: update notation of the number.

tests/playwright/src/specs/rh-extension-pack-installation.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+            const errorTab = extensionPage.tabs.getByRole('button', { name: 'Error' });
+            // we would like to propagate the error's stack trace into test failure message
+            let stackTrace = '';
+            if ((await errorTab.count()) > 0) {
+              stackTrace = await errorTab.innerText();
+            }
+            await playExpect(errorTab, `Error Tab was present with stackTrace: ${stackTrace}`).not.toBeVisible();
+          });
+
+          test.describe
```
odockal: Pls, remove this describe. Looks unnecessary.

tests/playwright/src/specs/rh-extension-pack-installation.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+                await extensionPage.disableExtension();
+                await playExpect(extensionPage.status).toHaveText(DISABLED);
+
+                await extensionPage.enableExtension();
+                await playExpect(extensionPage.status).toHaveText(ACTIVE);
+              });
+            });
+        });
+    }
+    test.describe
```
odockal: PLs, remove this describe, it is part of one above.

tests/playwright/src/specs/rh-extension-pack-installation.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+import { expect as playExpect, test } from '@playwright/test';
+
+import { extensionsPackExtension, extensionsRHPackList } from '../model/core/extensions';
+import { ExtensionCatalogCardPage } from '../model/pages/extension-catalog-card-page';
+import { ExtensionsPage } from '../model/pages/extensions-page';
+import { WelcomePage } from '../model/pages/welcome-page';
+import { NavigationBar } from '../model/workbench/navigation';
+import { Runner } from '../runner/podman-desktop-runner';
+
+const DISABLED = 'DISABLED';
```
odockal: I think that these constants can be extracted into `model/core/states.ts` module. They can be reused across tests.

odockal: please open a follow up issue to switch to using the constants/enums for the states from there in other tests as well.

tests/playwright/src/specs/rh-extension-pack-installation.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+const ACTIVE = 'ACTIVE';
+const DOWNLOADABLE = 'DOWNLOADABLE';
+
+const RH_EXTENSION_PACK_NAME = extensionsPackExtension.extensionName;
```
odockal: not really necessary I think.

odockal (changes requested):

odockal (approved):
LGTM.

benoitf (changes requested):
why would we have code for 3rd party extensions inside the main repository ?

danivilla9:
> why would we have code for 3rd party extensions inside the main repository ?

@benoitf The contents of this test case are reproduced during release testing, we are just automating the process. A full-on E2E test case which checks that every UI asset is in its expected place and that the extensions' functionalities behave as they should, would not belong here but we assessed that checking for proper extension installation does.

Also here is another example of extension tests in the main repo: https://github.com/podman-desktop/podman-desktop/blob/main/tests/playwright/src/specs/extension-installation-smoke.spec.ts this one includes developer sandbox, openshift local, openshift checker, and docker (unsure if any of these is considered 3rd party)

benoitf:
Podman Desktop repository should only contain code of what is inside this repository.

Test for Red Hat extension pack belongs to the redhat extension pack repository

> Also here is another example of extension tests in the main repo: https://github.com/podman-desktop/podman-desktop/blob/main/tests/playwright/src/specs/extension-installation-smoke.spec.ts this one includes developer sandbox, openshift local, openshift checker, and docker (unsure if any of these is considered 3rd party)

it should be moved out. it's not belonging to this repository. It should be part of each extension.

odockal:
@benoitf Podman Desktop provides the extensions catalog, we are testing if we can install extension via catalog in this test. We should then reduce what we are testing in `extension-installation-smoke`. So perhaps moving these tests there...

benoitf:
> we are testing if we can install extension via catalog in this test.

it should consist of one (or more) of extension being part of podman-desktop organization, not others

danivilla9:
Hmmm. I agree with your assessment, but it results in a bit of a mess for QE. I've just checked and there is no E2E infrastructure in the redhat-pack-ext repo, and makes me wonder if it's worth setting it up just for this really simple test case (because the extension's "further functionalities" should be tested in their respective extension's repos). If not, the alternative of manually testing it ad infinitum is not great either.

Adding a new workflow is itself a bit of a liability too because, with the amount of extensions with individual E2E infrastructure, it takes more time to run the automated tests and waiting for them to be executed than it takes to manually test the features, rendering automating them pointless. It's not that adding a new one will be a make-or-break situation, but it highlights the existing issue. We've discussed solutions to this problem, like the possibility of an openshift pipeline that triggers all of the extensions' workflows and unifies their results in a single point, but we haven't studied how feasible and/or worth this would actually be.

danivilla9:
Closing the PR, issue has been moved to https://github.com/redhat-developer/podman-desktop-redhat-pack-ext/issues/14

PR #12650, opened 2025-05-27
chore: bump svelte to 5.28.3

### LINKED ISSUE #12649: Bumping svelte to version >= 5.28.3 fails
### Bug description

Many renderer tests are failing

### Operating system

mac

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

I'm able to fix first tests by adding `await tick()` at specific places in failing tests. I will check all the other tests

feloy:
### What does this PR do?

Adds tick() in specific places in tests, to make them pass with svelte 5.28.3

Among the changes in 5.28.3 (https://github.com/sveltejs/svelte/releases/tag/svelte%405.28.3), I can only see the optimisation introduced by https://github.com/sveltejs/svelte/pull/15895 to be the reason of this change in behaviour (timeout not giving the hand to the renderer anymore, now needing to call tick for that).

Fixes #12649 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/appearance/IconImage.spec.ts
+++ b/packages/renderer/src/lib/appearance/IconImage.spec.ts
@@ -20,6 +20,7 @@
 import '@testing-library/jest-dom/vitest';
 
 import { render } from '@testing-library/svelte';
+import { tick } from 'svelte';
 import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
 import { AppearanceSettings } from '../../../../main/src/plugin/appearance-settings';
@@ -78,6 +79,7 @@ test('Expect valid source and alt text with light mode', async () => {
 
   // wait for image to be loaded
   await new Promise(resolve => setTimeout(resolve, 200));
+  await tick();
 
   // grab image element
   const imageElement = image.getByRole('img');
@@ -99,6 +101,7 @@ test('Expect no alt attribute if missing and default image', async () => {
 
   // wait for image to be loaded
   await new Promise(resolve => setTimeout(resolve, 200));
+  await tick();
 
   // grab image element
   const imageElement = image.getByRole('img');
@@ -116,7 +119,7 @@ test('Expect string as image', async () => {
 
   // wait for image to be loaded
   await new Promise(resolve => setTimeout(resolve, 200));
-
+  await tick();
   // grab image element
   const imageElement = image.getByRole('img');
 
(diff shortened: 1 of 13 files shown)


cdrage (approved):
That is so strange that we now have to add `await tick()` to a lot of tests...

This LGTM, but I feel as though there should be an issue opened upstream because it's so odd it won't "waitFor" correctly now.

Or maybe we add to our code guidelines for testing that `await tick()` may be useful to force a render?

packages/renderer/src/lib/appearance/IconImage.spec.ts
```diff
@@ -78,6 +79,7 @@ test('Expect valid source and alt text with light mode', async () => {
 
   // wait for image to be loaded
   await new Promise(resolve => setTimeout(resolve, 200));
+  await tick();
```
axel7083: having this both really feel like a bad practice, and maybe we should use `vi.waitFor` instead

packages/renderer/src/lib/image/ImageDetailsCheck.spec.ts
```diff
@@ -112,7 +113,8 @@ test('expect to cancel when clicking the Cancel button', async () => {
     await fireEvent.click(abortBtn);
   });
 
-  await vi.waitFor(() => {
+  await vi.waitFor(async () => {
+    await tick();
```
axel7083: super weird that this is needed, since we are in a waitFor

feloy:
@axel7083 I have made changes, PTAL

benoitf (approved):

benoitf:
@feloy could you please fix the pnpm-lock conflict


then, I think it can be merged as changes are only in `.spec.ts` files

axel7083:
> then, I think it can be merged as changes are only in `.spec.ts` files

I was discussing with @feloy about the fixes provided, and after he investigated, the problems are related to side effect between tests, that `tick` seems to be avoiding.

So I suggested we might fix those test in dedicated PR, making https://github.com/podman-desktop/podman-desktop/issues/12649 and epic.

benoitf:
the problem is that waiting for the update, we're not updating the svelte version so we're not seeing new regressions easily

axel7083:
> the problem is that waiting for the update, we're not updating the svelte version so we're not seeing new regressions easily

IMO introducing hacky / partial solution might have more consequences on the maintenance / readability of our tests than waiting a bit for fixing them, but at the end this is up to you, I am not blocking the update 👍

PR #12661, opened 2025-05-28
feat(button): added support for class image in button component

gastoner:
### What does this PR do?
Adds support for showing image in button using class when calling from backend 
From frontend it can be shown using ImageDefiniction, but from backend there is no such way => class definition 

### Screenshot / video of UI
E.g.:
image

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/button/Button.svelte
+++ b/packages/ui/src/lib/button/Button.svelte
@@ -42,7 +42,11 @@ let {
   children,
 }: Props = $props();
 
-let iconType: string = $derived(isFontAwesomeIcon(icon) ? 'fa' : 'unknown');
+let iconType: string = $derived.by(() => {
+  if (isFontAwesomeIcon(icon)) return 'fa';
+  else if (typeof icon === 'string') return 'string';
+  else return 'unknown';
+});
 
 let classes = $derived.by(() => {
   let result: string = '';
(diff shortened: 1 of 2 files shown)


deboer-tim (changes requested):
We currently have different levels of icon support in Button, SettingsNavItem, DropdownMenu, Link, StatusIcon, and ListItemButtonIcon 🤕, and this has become even more confusing with the Svelte 5 typing changes.

So I totally support this change, but I would recommend that we only do it by creating some reusable/sharable icon type/component first, vs diverging Button from the others. Even if it is only used within Button to start this would begin to fix the problem vs compounding it.

packages/ui/src/lib/button/Button.svelte
```diff
@@ -42,7 +42,11 @@ let {
   children,
 }: Props = $props();
 
-let iconType: string = $derived(isFontAwesomeIcon(icon) ? 'fa' : 'unknown');
+let iconType: string = $derived.by(() => {
+  if (isFontAwesomeIcon(icon)) return 'fa';
+  else if (typeof icon === 'string') return 'string';
```
benoitf, line 47: I think it might be confusing to have the 'string' type 
Maybe 'custom' ? Or 'font' or something else

but it's internal so can be changed any time

benoitf (approved):

gastoner:
Closing in favour of https://github.com/podman-desktop/podman-desktop/pull/12678

PR #12677, opened 2025-05-29
feat(ui): created universal icon

gastoner:
### What does this PR do?
Adds universal icon to Podman Desktop UI lib

We are using 3 different types of rendering images: 
1. Using Fa
2. Using classes
3. Rendering component

All over the PD repo there are those 3 options randomly used, this Component will replace all those places by this component 

Required for https://github.com/podman-desktop/podman-desktop/pull/12661

Can be tried out in https://github.com/podman-desktop/podman-desktop/pull/12678

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/icons/Icon.svelte
+++ b/packages/ui/src/lib/icons/Icon.svelte
@@ -1,0 +1,40 @@
+<script lang="ts">
+import type { IconDefinition } from '@fortawesome/free-solid-svg-icons';
+import type { Component } from 'svelte';
+import { Fa, type IconSize } from 'svelte-fa';
+
+import { isFontAwesomeIcon, isFontAwesomeSize } from '../utils/icon-utils';
+
+interface Props {
+  icon: IconDefinition | Component | string;
+  size?: IconSize | number | string;
+  class?: string;
+  color?: string;
+  'aria-label'?: string;
+  'aria-hidden'?: string;
+  role?: string;
+  solid?: boolean;
+}
+
+let {
+  icon,
+  size,
+  class: className,
+  color,
+  'aria-label': ariaLabel,
+  'aria-hidden': ariaHidden,
+  role = 'img',
+  solid = true,
+}: Props = $props();
+
+const IconComponent = icon;
+</script>
+
+
+{#if isFontAwesomeIcon(icon) && (typeof size === 'undefined' || isFontAwesomeSize(size))}
+    <Fa {icon} {size} class={className} {color}/>
+{:else if typeof icon === 'string' && (icon.startsWith('fas fa-') || icon.startsWith('far fa-'))}
+    <i class={`${icon} ${size} ${className}`} aria-label={ariaLabel} aria-hidden={ariaHidden ? Boolean(ariaHidden) : false} {role}></i>
+{:else if IconComponent && typeof IconComponent !== 'string' && !isFontAwesomeIcon(IconComponent)}
+    <IconComponent class={className} {size} {role} {solid}/>
+{/if}
--- a/packages/ui/src/lib/icons/Icon.svelte
+++ b/packages/ui/src/lib/icons/Icon.svelte
@@ -9,32 +9,19 @@ interface Props {
   icon: IconDefinition | Component | string;
   size?: IconSize | number | string;
   class?: string;
-  color?: string;
-  'aria-label'?: string;
-  'aria-hidden'?: string;
-  role?: string;
-  solid?: boolean;
 }
 
-let {
-  icon,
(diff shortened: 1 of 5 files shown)


packages/ui/src/lib/icons/Icon.svelte
```diff
@@ -0,0 +1,40 @@
...
+  solid = true,
+}: Props = $props();
+
+const IconComponent = icon;
+</script>
+
+
+{#if isFontAwesomeIcon(icon) && (typeof size === 'undefined' || isFontAwesomeSize(size))}
+    <Fa {icon} {size} class={className} {color}/>
+{:else if typeof icon === 'string' && (icon.startsWith('fas fa-') || icon.startsWith('far fa-'))}
```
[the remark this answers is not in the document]
gastoner: I think we uses only fas- and far-

packages/ui/src/lib/icons/Icon.svelte
```diff
@@ -0,0 +1,40 @@
...
+    <i class={`${icon} ${size} ${className}`} aria-label={ariaLabel} aria-hidden={ariaHidden ? Boolean(ariaHidden) : false} {role}></i>
```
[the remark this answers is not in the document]
gastoner: Any tips for this? I dont like merging all into class but I'm not aware of any different approach :(

gastoner:
+ I think that we can drop the color, because black should be default one

gastoner:
@deboer-tim @benoitf  can you PTAL?

benoitf:
@gastoner I haven't reviewed as I didn't see any unit tests

gastoner:
@benoitf I've added tests

deboer-tim (commented):
Sorry for the slow review. I found going through this difficult for the same reason that I think it's necessary: I spent most of the time trying to understand where/why we use `<I>` with size, color with Fa, or why aria-label and role are only on one of the options.

On one hand this is just exposing the ugly bits of what we're already doing today; OTOH putting it in a shared component will just propagate some of the odd behaviour or make it harder to clean up later. My preference would be to start cleaner, or maybe not even exposing from the UI library to start.

It's ok for some of the unique properties like solid to not apply to all options, but generic props like aria-label or role should either apply to all of the options or be removed. It's ok if this doesn't support every single icon property we'll need in the UI library, better to get the core cases clean and we can always add or review if we really need all the other options later.

gastoner:
@deboer-tim I've removed unnecesary props so it contains only core options
- I've left the `role` since it should already be present in `Fa`

packages/ui/src/lib/icons/Icon.svelte
```diff
@@ -0,0 +1,27 @@
...
+const role = 'img';
+const IconComponent = icon;
+</script>
+
+
+{#if isFontAwesomeIcon(icon) && (typeof size === 'undefined' || isFontAwesomeSize(size))}
+    <Fa {icon} {size} class={className}/>
+{:else if typeof icon === 'string' && (icon.startsWith('fas fa-') || icon.startsWith('far fa-'))}
+    <i class={`${icon} ${size} ${className}`} {role}></i>
+{:else if IconComponent && typeof IconComponent !== 'string' && !isFontAwesomeIcon(IconComponent)}
```
deboer-tim: I'm assuming the ifs here are just to get the typing correct? If it's possible I would keep the top level ifs simple and then use nested ifs to check the size and string, i.e.:
```
{#if isFontAwesomeIcon(icon)}
  // nested if for size
   <Fa ...
{:else if typeof icon === 'string'}
 // nested if for startsWiths
  <I class...
{:else}
  <span {role}><IconCom
{/if}
```

deboer-tim (approved):
One little suggestion noted inline. The other minor things I don't like are things we're already doing in the codebase, and at least getting them in one place like this will consolidate it and make it easier to keep things consistent and change later! Thanks.

packages/ui/src/lib/icons/Icon.svelte
gastoner: I've updated this, but the else statement looks quite funny :D I need to do the checks here otherwise it wont let me render the component :(

deboer-tim: I was hoping it would mean you don't need line 30 anymore since the previous 2 if statements would reduce the type signature? It's ok though, the other sub-ids are really dependent properties and I think it reads a little cleaner now.

gastoner: Svelte check is yalling :( if I remove this line

axel7083 (approved):

PR #12678, opened 2025-05-29
refactor(ui): use new icon component in ui svelte lib

gastoner:
### What does this PR do?
Uses new UI svelte icon component in PD lib

This PR will be rebased after is new icon component merged

### How to test this PR?
There should be no visual changes

### DIFF
--- a/packages/renderer/src/lib/actions/ContributionActions.spec.ts
+++ b/packages/renderer/src/lib/actions/ContributionActions.spec.ts
@@ -255,15 +255,16 @@ test('Expect custom icon on the contributed action', async () => {
       {
         command: 'dummy.command',
         title: 'dummy-title',
-        icon: '${dummyIcon}',
+        icon: 'fas fa-${dummyIcon}',
       },
     ],
     onError: () => {},
     dropdownMenu: true,
   });
 
-  const iconItem = screen.getByRole('img', { name: 'dummy-title' });
+  const iconItem = screen.getByRole('img', { hidden: true });
   expect(iconItem).toBeInTheDocument();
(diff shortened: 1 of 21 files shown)


axel7083 (approved):
LGTM

deboer-tim (commented):
For cases like Button and DropDownMenuItem this looks great - and I'd love to take it one step further (future PR fine) where we'd just use Icon directly. i.e. no if statement, whatever icon supports is what's exposed by the component; every component is consistent.

For cases like changing `<Fa icon={faXmark} />` to `<Icon icon={faXmark} />` I'm not against it, but I don't see the point - it's a hardcoded icon so we don't need the flexibility of Icon / to change every Fa to Icon.

gastoner:
> For cases like Button and DropDownMenuItem this looks great - and I'd love to take it one step further (future PR fine) where we'd just use Icon directly. i.e. no if statement, whatever icon supports is what's exposed by the component; every component is consistent.
> 
> For cases like changing `<Fa icon={faXmark} />` to `<Icon icon={faXmark} />` I'm not against it, but I don't see the point - it's a hardcoded icon so we don't need the flexibility of Icon / to change every Fa to Icon.

Yes I agree, I would like to replace all the icons with just `Icon` I'm just not sure how to set some size e.g. for Fa icon and class icon since in the class icon I feel we are using a lot some classes for styling etc.

I would also like to have this change consistent so all images would be displayed idealy with just `Icon` component although in places where is only used Fa it is not required

jeffmaury (approved):
LGTM

jeffmaury (approved):
LGTM

PR #12682, opened 2025-05-29
fix: avoid shell prompt duplication in container's terminal when switching between tabs

### LINKED ISSUE #11952: Container terminal tab appends a new line at the end of the previous line when changing tabs
### Bug description

Currently if the users switches to and from the Terminal tab in Container Details page each time the user navigates to Terminal will add a new begin of line to the end of the previous one.

https://github.com/user-attachments/assets/94c121b1-44ed-4cd4-a9c5-64d1c72c9a7e

### Operating system

macos

### Installation Method

### Version

next (development version)

### Steps to reproduce

1. Open container details page for a running container.
2. Go to Terminal tab.
3. Go to one of the other tabs available.
4. Navigate back to Terminal.
5. Do Steps 2 and 3 in a cycle and notice that the beginning of line is appended to the end of the previous line in the Terminal tab.

### Relevant log output

```shell

```

### Additional context

_No response_

dgolovin:
### What does this PR do?

Fix adds a removal of prompt from the end of stored terminal. It is added automatically after contend of the terminal is restored.

### Screenshot / video of UI

https://github.com/user-attachments/assets/5ff6ccb3-9e0b-4702-b449-71d752490304

Fix #11952.

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
@@ -19,7 +19,6 @@
 import '@testing-library/jest-dom/vitest';
 
 import { render, screen, waitFor } from '@testing-library/svelte';
-import { tick } from 'svelte';
 import { get } from 'svelte/store';
 import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
```diff
@@ -152,7 +136,7 @@ test('terminal active/ restarts connection after stopping and starting a contain
         onEnd();
       }, 500);
       // return a callback id
-      return Promise.resolve(sendCallbackId);
+      return sendCallbackId;
```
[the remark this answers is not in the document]
dgolovin: If you return a non-Promise value, it is automatically wrapped in a Promise if function is marked as `async`.

packages/renderer/src/lib/container/ContainerDetailsTerminal.svelte
```diff
@@ -148,8 +148,24 @@ onMount(async () => {
   await executeShellIntoContainer();
 });
 
+function removePromptFromTheEnd(terminal: string): string {
```
cdrage, line 151: Any way we could possibly have this in a separate file so it can be used for other Terminal sections? (pods, kubernetes, etc.)

dgolovin: Good point, but right now it is only client for terminals store and I assume this is the only place where the function can be used right now. When time comes we might pull it out.

cdrage: ACK. That makes sense and you're right. It's the only place using stores vs Kubernetes implementation.

gastoner (approved):
LGTM on Linux

benoitf (commented):
can we avoid to have the prompt being created ?

I mean, here we're removing the prompt character from the text, but could it be avoided to have another prompt character ?

dgolovin:
> can we avoid to have the prompt being created ?
> 
> I mean, here we're removing the prompt character from the text, but could it be avoided to have another prompt character ?

prompt string comes from container through Dockerode execStream.onData and ends on ui side in receiveDataCallback. There is no options to suppress sending prompt back after dialing in remote container shell or after terminal.write when previous terminal session output is restored.

I can ignore first receiveDataCallback like this PR https://github.com/podman-desktop/podman-desktop/pull/12725, but that is even more complex and cannot be reused easily.

dgolovin:
Ignoring callback version #12725 were merged instead. Closing this PR.

PR #12689, opened 2025-05-30
chore(e2e): add scrollintoviewifneeded to ui-stress-test

danivilla9:
Signed-off-by: Daniel Villanueva <davillan@redhat.com>

### What does this PR do?
Adds the scrollIntoViewIfNeeded() functionality to the ui-stress-test spec file so that, if the execution of test case fails, the problematic row appears on the video afterwards.

[#362](https://github.com/podman-desktop/e2e/issues/362)

### How to test this PR?
Run `pnpm test:e2e:ui-stress`

### DIFF
--- a/tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
+++ b/tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
@@ -43,10 +43,7 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
     //count images => 1 original image + (1 tagged * numberOfObjects) + 1 localhost/podman-pause from pods = numberOfObjects + 2
     await playExpect.poll(async () => await images.countRowsFromTable(), { timeout: 10_000 }).toBe(numberOfObjects + 2);
     for (let imgNum = 1; imgNum <= numberOfObjects; imgNum++) {
-      const imgName = `localhost/my-image-${imgNum}`;
-      await playExpect.poll(async () => await images.waitForRowToExists(imgName), { timeout: 10_000 }).toBeTruthy();
-
-      await scrollRowIntoView(images, imgName);
+      await findRowAndScroll(images, `localhost/my-image-${imgNum}`);
     }
   });
 
@@ -60,12 +57,7 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
       .poll(async () => await containers.countRowsFromTable(), { timeout: 10_000 })
       .toBe(3 * numberOfObjects);
     for (let containerNum = 1; containerNum <= numberOfObjects; containerNum++) {
-      const containerName = `my-container-${containerNum}`;
-      await playExpect
-        .poll(async () => await containers.waitForRowToExists(containerName), { timeout: 0 })
-        .toBeTruthy();
-
-      await scrollRowIntoView(containers, containerName);
+      await findRowAndScroll(containers, `my-container-${containerNum}`);
     }
   });
 
@@ -77,19 +69,15 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
     //count pods => 1 manually created * numberOfObjects = numberOfObjects
     await playExpect.poll(async () => await pods.countRowsFromTable(), { timeout: 10_000 }).toBe(numberOfObjects);
     for (let podNum = 1; podNum <= numberOfObjects; podNum++) {
-      const podName = `my-pod-${podNum}`;
-      await playExpect.poll(async () => await pods.waitForRowToExists(podName), { timeout: 0 }).toBeTruthy();
-
-      await scrollRowIntoView(pods, podName);
+      await findRowAndScroll(pods, `my-pod-${podNum}`);
     }
   });
 });
 
-async function scrollRowIntoView(page: MainPage, rowName: string): Promise<void> {
+async function findRowAndScroll(page: MainPage, rowName: string): Promise<void> {
   const rowLocator = await page.getRowByName(rowName);
   if (rowLocator === undefined) {
     throw Error(`Row: '${rowName}' does not exist`);
   }
   await rowLocator.scrollIntoViewIfNeeded();
-  await playExpect(rowLocator).toBeVisible();
 }
--- a/tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
+++ b/tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
@@ -41,6 +41,9 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
     //count images => 1 original image + (1 tagged * numberOfObjects) + 1 localhost/podman-pause from pods = numberOfObjects + 2
     await playExpect.poll(async () => await images.countRowsFromTable(), { timeout: 10_000 }).toBe(numberOfObjects + 2);
     for (let imgNum = 1; imgNum <= numberOfObjects; imgNum++) {
+      await playExpect
+        .poll(async () => await images.waitForRowToExists(`localhost/my-image-${imgNum}`), { timeout: 0 })
+        .toBeTruthy();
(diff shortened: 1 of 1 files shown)


tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
```diff
@@ -71,7 +77,19 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
...
 });
+
+async function scrollRowIntoView(page: MainPage, rowName: string): Promise<void> {
+  const rowLocator = await page.getRowByName(rowName);
+  if (rowLocator === undefined) {
+    throw Error(`Row: '${rowName}' does not exist`);
+  }
+  await rowLocator.scrollIntoViewIfNeeded();
+  await playExpect(rowLocator).toBeVisible();
+}
```
cbr7: I don't really understand why this method is needed? Why not call scrollIntoView method directly from PW object? This method doesn't make much sense to me.

`await playExpect(rowLocator).toBeVisible()` will return true even without scrollIntoView being ran or still in execution.

tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
```diff
@@ -41,9 +43,10 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
     //count images => 1 original image + (1 tagged * numberOfObjects) + 1 localhost/podman-pause from pods = numberOfObjects + 2
     await playExpect.poll(async () => await images.countRowsFromTable(), { timeout: 10_000 }).toBe(numberOfObjects + 2);
     for (let imgNum = 1; imgNum <= numberOfObjects; imgNum++) {
-      await playExpect
-        .poll(async () => await images.waitForRowToExists(`localhost/my-image-${imgNum}`), { timeout: 0 })
-        .toBeTruthy();
+      const imgName = `localhost/my-image-${imgNum}`;
+      await playExpect.poll(async () => await images.waitForRowToExists(imgName), { timeout: 10_000 }).toBeTruthy();
```
cbr7:
```suggestion
      await playExpect.poll(async () => await images.waitForRowToExists(imgName), { timeout: 0 }).toBeTruthy();
```

cbr7 (changes requested):
I'm not sure I understand the point of these changes.

tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
odockal: I would say @cbr7 is right in the first point, that we do not need the method it self, just call it over locator.

but the other point, we actually need scrollIntoView since when there was a failure, we were not able to find out if there is any UI problem since the page was stuck at the top in the video.

cbr7: @odockal I sort of understood what the use case is, but my second point is that the toBeVisible() is not an adequate assert for this, since regardless of if the row is in view or not it's going to return true if the row is in the DOM, so it can be removed completely without any consequence.

odockal: I agree.

danivilla9: @cbr7 Okay, now I understand that toBeVisible() is pointless in this case, but I can't see why using a method isn't correct in order to not repeat code, getRowByName() returns a Promise<Locator|undefined> so I can't scrollIntoViewIfNeeded() directly onto it, I have to check that it's not undefined (and it would be the very same operation for the three test cases). I can define the row as 'page.getByRole(...)' but doesn't that defeat the purpose of having a method that searches for rows with a given name in a page?

cbr7: @danivilla9 you are already waiting for the row to exist in the previous assert, as such there will not be a situation where that object will be undefined and reach that point, because it would fail before it.

danivilla9: @cbr7 Yes, but waitForRowToExists() does not return a Locator, it returns a boolean, and I need to get the locator from somewhere else in order to invoke scrollIntoViewIfNeeded() in the row. But I think I understand your logic, I've pushed a different approach, countRowsFromTable waits for the rows to load, so I can access the locator directly with getRowByName (but with an extracted method still, because there is repeated code otherwise)

cbr7: So use optional chaining after the row assert with `row?.scrollIntoViewIfNeeded()`

danivilla9: You are absolutely right. I think the new version should be correct then.

tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
```diff
@@ -41,9 +41,8 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
     //count images => 1 original image + (1 tagged * numberOfObjects) + 1 localhost/podman-pause from pods = numberOfObjects + 2
     await playExpect.poll(async () => await images.countRowsFromTable(), { timeout: 10_000 }).toBe(numberOfObjects + 2);
     for (let imgNum = 1; imgNum <= numberOfObjects; imgNum++) {
-      await playExpect
-        .poll(async () => await images.waitForRowToExists(`localhost/my-image-${imgNum}`), { timeout: 0 })
-        .toBeTruthy();
```
cbr7: I'm not sure that you should be removing these asserts, you should still wait for the rows to exist in the list, why did you remove them? The getRowByName() method returns instantly, so if you run it right after you created a image/container/pod it's possible(even likely) that it will return undefined since the UI did not have time to process the request.

Please revert the assert/waiter deletions from all three places.

tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
```diff
@@ -57,9 +56,8 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
       .poll(async () => await containers.countRowsFromTable(), { timeout: 10_000 })
       .toBe(3 * numberOfObjects);
     for (let containerNum = 1; containerNum <= numberOfObjects; containerNum++) {
-      await playExpect
-        .poll(async () => await containers.waitForRowToExists(`my-container-${containerNum}`), { timeout: 0 })
-        .toBeTruthy();
```
cbr7: Same as above.

tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
```diff
@@ -71,7 +69,8 @@ test.describe.serial('Verification of UI handling lots of objects', { tag: ['@ui
     //count pods => 1 manually created * numberOfObjects = numberOfObjects
     await playExpect.poll(async () => await pods.countRowsFromTable(), { timeout: 10_000 }).toBe(numberOfObjects);
     for (let podNum = 1; podNum <= numberOfObjects; podNum++) {
-      await playExpect.poll(async () => await pods.waitForRowToExists(`my-pod-${podNum}`), { timeout: 0 }).toBeTruthy();
```
cbr7, line 74: Same as above.

tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
danivilla9: I thought that countRowsFromTable() would wait for the rows to be there before any checks, so I found them redundant. I've restored them

cbr7: No, countRowsFromTable() does not do that, what it does is wait for the table to exist, meaning that the header row is displayed, but that does not mean that individual entries in the table are awaited.

cbr7 (approved):

Issue #12703, opened 2025-06-02
Add e2e test utilities for browser authentication


odockal:
### Is your enhancement related to a problem? Please describe

In the other repository, we have a functionality that can be shared across multiple e2e tests.

### Describe the solution you'd like

Generalize and share the authentication and other functionality to work with chromium browser from playwright e2e tests when pd is being run.

PR #12704, opened 2025-06-02
chore(test): share e2e tests authentication functionality

odockal:
### What does this PR do?
Adds multiple e2e utility functions to allow switching from pd app to browser, logging in using browser, etc.

#12703 

### How to test this PR?


cbr7 (approved):

tests/playwright/src/utility/auth-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+    chromePage = pagesTitles.find(p => p.title.includes(expectedTitle))?.page;
+    if (chromePage) {
+      break;
+    }
+  }
+
+  if (!chromePage) {
+    console.error(`No page found with title: ${expectedTitle}`);
+  }
+  return chromePage;
```
benoitf: there is no waitFor method instead of a manual loop ?

tests/playwright/src/utility/auth-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export async function performBrowserLogin(page: Page, username: string, pass: string, path: string): Promise<void> {
+  console.log(`Performing browser login...`);
+  await playExpect(page).toHaveTitle(/Log In/);
+  await playExpect(page.getByRole('heading', { name: 'Log in to your Red Hat' })).toBeVisible();
+  console.log(`We are on the login RH Login page...`);
```
benoitf: how this method is a browser login method ? it's named Red Hat there

tests/playwright/src/utility/auth-utils.ts
odockal: Improved a little.

tests/playwright/src/utility/auth-utils.ts
odockal: Updated.

benoitf: I still don't get why would we have RH login page in Podman Desktop ?

odockal: Because we need to share it across multiple extensions.

odockal: @benoitf take a look pls., now it is more general for use.

benoitf (approved):

PR #12714, opened 2025-06-02
fix: clear caches for all resources when one informer is offline

### LINKED ISSUE #12465: Active nodes count on k8s dashboard only refreshes to correct values after context changes
### Bug description

It seems that currently the value displayed in the `Active nodes` column from the k8s dashboard only shows correct values after switching the k8s context.

https://github.com/user-attachments/assets/afb5b32d-1340-4b04-ab4e-17f594d8fcf3

### Operating system

macOS

### Installation Method

### Version

1.18.0

### Steps to reproduce

1. Start a minikube/kind/minc cluster from the Settings -> Resources page.
2. Switch k8s context to the newly created context.
3. Go to k8s dashboard and see that under Nodes the values will be 1 for active and 1 for total.
4. Go back to Settings -> Resources and stop the previously created cluster.
5. Go back to the k8s dashboard.
6. Notice that the dashboard still says 1 active and 1 total nodes.
7. Switch k8s context in PD to some other unrelated k8s context and then switch back to the previous context.
8. Notice that after the k8s context switching now 0 active clusters will be displayed on the k8s dashboard.

### Relevant log output

```shell

```

### Additional context

_No response_

feloy:
### What does this PR do?

Clear kubernetes caches when an informer is offline. 

The effect is to reset all counts to zero for all resources of the context for which at least one informer is offline.

Fixes #12465 

### How to test this PR?

See steps to reproduce in #12465 

### DIFF
--- a/packages/main/src/plugin/kubernetes/contexts-manager-experimental.spec.ts
+++ b/packages/main/src/plugin/kubernetes/contexts-manager-experimental.spec.ts
@@ -18,7 +18,8 @@
 
 import type { Cluster, KubernetesObject, ObjectCache } from '@kubernetes/client-node';
 import { KubeConfig } from '@kubernetes/client-node';
-import { beforeEach, describe, expect, test, vi } from 'vitest';
+import type { Event } from '@podman-desktop/api';
+import { assert, beforeEach, describe, expect, test, vi } from 'vitest';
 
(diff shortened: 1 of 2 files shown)


gastoner (approved):
LGTM on Linux
After stopping cluster the active will go from 1->0

axel7083 (approved):

PR #12715, opened 2025-06-02
chore(optimize): telemetry - remove createProvider for createdProviders

### LINKED ISSUE #12634: Telemetry: Optimize `createProvider` event

This event gets fired every time a provider is started. This is leading to millions of events. 10% of all our events. Look at how we can reduce it. 

**Idea:** 

```
User123:
provider.podman: enabled
provider.docker: enabled
```

**Idea 2:** 

`createdProviders: podnan, kind`

That way we need to change our queries to `where createdProviders contains "kind"`

_No response_

### Additional context

_No response_

cdrage:
chore(optimize): telemetry - remove createProvider for createdProviders

### What does this PR do?

* Use the new telemetry call createdProviders which sends a list of all
  the providers as well as the information (id, name, status, version)
  similar to the previous telemetry call.
* Above call is done after the loading of extensions.
* Removes createProvider call since it is being called multiple times on
  start

### Screenshot / video of UI

N/A

Closes https://github.com/podman-desktop/podman-desktop/issues/12634

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extension/extension-loader.ts
+++ b/packages/main/src/plugin/extension/extension-loader.ts
@@ -431,11 +431,34 @@ export class ExtensionLoader {
     // now we have all extensions, we can load them
     await this.loadExtensions(analyzedExtensions);
 
+    // Send telemetry of all the providers which were activated by the extensions
+    try {
+      await this.sendTelemetryProviderInfos();
+    } catch (error: unknown) {
+      // We do not care about the error as we never want it to fail the loading of extensions
+      console.error('Error while sending telemetry for created providers', error);
+    }
+
     // handle the reload extensions callback
     this.extensionWatcher.onNeedToReloadExtension(extension => {
       this.reloadExtension(extension, false).catch((error: unknown) => {
         console.error('error while reloading extension', error);
       });
+    });
+  }
+
+  // Send telemetry of all the providers which were activated by the extensions
+  // Payload ex. { "id": "provider.podman", "name": "Podman", "status": "started", "version": "5.0.0" }
+  async sendTelemetryProviderInfos(): Promise<void> {
+    const providerInfos = this.providerRegistry.getProviderInfos();
+    const providersDetailed = providerInfos.map(p => ({
+      id: p.id,
+      name: p.name,
+      status: p.status,
+      version: p.version ?? 'unknown',
+    }));
+    this.telemetry.track('createdProviders', {
+      providers: providersDetailed,
     });
   }
 
--- a/packages/main/src/plugin/provider-registry.ts
+++ b/packages/main/src/plugin/provider-registry.ts
@@ -212,14 +212,6 @@ export class ProviderRegistry {
     this.count++;
     this.providers.set(id, providerImpl);
     this.listeners.forEach(listener => listener('provider:create', this.toProviderInfo(providerImpl)));
-    const trackOpts: { name: string; status: string; version?: string } = {
-      name: providerOptions.name,
-      status: providerOptions.status.toString(),
-    };
-    if (providerOptions.version) {
-      trackOpts.version = providerOptions.version;
-    }
-    this.telemetryService.track('createProvider', trackOpts);
     this.apiSender.send('provider-create', id);
     providerImpl.onDidUpdateVersion(() => this.apiSender.send('provider:update-version'));
     return providerImpl;
--- a/packages/main/src/plugin/extension/extension-loader.spec.ts
+++ b/packages/main/src/plugin/extension/extension-loader.spec.ts
@@ -35,6 +35,7 @@ import { ExtensionLoaderSettings } from '/@api/extension-loader-settings.js';
 import type { BuildImageOptions as InternalBuildImageOptions } from '/@api/image-info.js';
 import { NavigationPage } from '/@api/navigation-page.js';
 import type { OnboardingInfo } from '/@api/onboarding.js';
+import type { ProviderInfo } from '/@api/provider-info.js';
 import type { WebviewInfo } from '/@api/webview-info.js';
 
 import { getBase64Image } from '../../util.js';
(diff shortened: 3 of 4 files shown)


cdrage:
cc @Firewall and @benoitf 

This goes with option B on the original issue.

Instead of multiple calls, we just do one after all the extensions activate.

The example payload being sent to amplitude (tested on my own system) is:

```
[
  {
    id: 'podman',
    name: 'Podman',
    status: 'installed',
    version: '5.5.0'
  },
  {
    id: 'kubectl',
    name: 'kubectl',
    status: 'unknown',
    version: 'unknown'
  },
  { id: 'kind', name: 'Kind', status: 'unknown', version: 'unknown' },
  {
    id: 'Compose',
    name: 'Compose',
    status: 'unknown',
    version: 'unknown'
  },
  {
    id: 'bootc',
    name: 'Bootc VMs',
    status: 'unknown',
    version: 'unknown'
  }
]
```

FYI the status `unknown` is normal / intentional, as that's how it is in amplitude right now.

cdrage:
cc @Firewall and @benoitf 

This goes with option B on the original issue.

Instead of multiple calls, we just do one after all the extensions activate.

The example payload being sent to amplitude (tested on my own system) is:

```
[
  {
    id: 'podman',
    name: 'Podman',
    status: 'installed',
    version: '5.5.0'
  },
  {
    id: 'kubectl',
    name: 'kubectl',
    status: 'unknown',
    version: 'unknown'
  },
  { id: 'kind', name: 'Kind', status: 'unknown', version: 'unknown' },
  {
    id: 'Compose',
    name: 'Compose',
    status: 'unknown',
    version: 'unknown'
  },
  {
    id: 'bootc',
    name: 'Bootc VMs',
    status: 'unknown',
    version: 'unknown'
  }
]
```

FYI the status `unknown` is normal / intentional, as that's how it is in our internal code.

packages/main/src/plugin/extension/extension-loader.ts
```diff
@@ -431,6 +431,29 @@ export class ExtensionLoader {
     // now we have all extensions, we can load them
     await this.loadExtensions(analyzedExtensions);
 
+    // START OF TELEMETRY
```
benoitf: suggestion: please remove this kind of comments

// START
block
// END

usually these comments will be lost/moved across refactoring

replace it by a method that will manage the telemetry like

this.sendTelemetryProviderInfos()

packages/main/src/plugin/extension/extension-loader.ts
```diff
@@ -431,6 +431,29 @@ export class ExtensionLoader {
...
+      // Payload ex. { "id": "provider.podman", "name": "Podman", "status": "started", "version": "5.0.0" }
+      const providerInfos = this.providerRegistry.getProviderInfos();
+      const providersDetailed = providerInfos.map(p => ({
+        id: p.id,
+        name: p.name,
+        status: p.status,
+        version: p.version ?? 'unknown',
+      }));
+      this.telemetry.track('createdProviders', {
+        providers: providersDetailed,
```
benoitf: did you test on Amplitude that you can do graphs like what is the most created providers

what is the version of provider named "podman", etc ?

cdrage: I was shown by @Firewall that we can use the formula function to utilize the data / graph it.

Can you confirm @Firewall that providing info like this is fine on Amplitude?

cdrage:
@benoitf Updated! 

Will have to confirm with @Firewall that we can graph this fine / send the data over to the test amplitude to use.

packages/main/src/plugin/extension/extension-loader.ts
Firewall: You can't parse json in ampltitude if that is what you are asking. There are Formulas you can do https://amplitude.com/docs/analytics/charts/event-segmentation/event-segmentation-custom-formulas. When I mentioned that it was more in regards to backwards compatibility. We can do `SUM(UNIQUES(oldEvent), UNIQUES(newEvent))`

It looks like if we configure it in amplitude we can recieve an array and handle it accordingly. We need to change property type to array in the settings UI. https://amplitude.com/docs/analytics/charts/cart-analysis You are sending an Array of providers, right?

Is there a way we can send test events to try that? On the Development Branch of Amplitude

cdrage:
Screenshot 2025-06-03 at 10 07 07 AM

It has been successfully pushed to the dev environment.

The data shows up as an array of json entries that can be used for charting.

I'm unfamiliar with the Formula portion + getting use of the JSON data. Can you confirm on amplitude @Firewall that this works fine?

Screenshot 2025-06-03 at 10 08 31 AM

cdrage:
Confirmed on Slack that this propagates correctly to Amplitude / new tab is opening in order for us to process data!

cc @Firewall

Firewall:
Yup! 
image

Firewall (approved):
I can cofirm that the test events were recieved correctly in Amplitude and I was able to chart them.

benoitf (changes requested):
providers are register asynchronously

so it can't be sent after loading extensions

packages/main/src/plugin/provider-registry.ts
```diff
@@ -212,14 +212,6 @@ export class ProviderRegistry {
     this.count++;
     this.providers.set(id, providerImpl);
     this.listeners.forEach(listener => listener('provider:create', this.toProviderInfo(providerImpl)));
```
cdrage, line 214: Maybe @benoitf we batch / collect all the providers doing this call and do a batch call 30 seconds after startup or something similar? Send it out asynchronously / in the background.

WDYT

cdrage:
> providers are register asynchronously
> 
> so it can't be sent after loading extensions

I've commented a suggestion what we could do. Perhaps instead of in extension-loader, we have it in provider-registry and send a combined / bulk telemetry call?

cdrage:
Closing as we are going to look at alternative ways to implement as well as @benoitf taking up the issues for telemetry optimization.

Big thanks @benoitf for taking this issue up.

Thank you @Firewall for figuring out with me with regards to Amplitude / being able to pass an object!

PR #12728, opened 2025-06-03
fix: wrong margin in the update available modal

### LINKED ISSUE #9896: "Update now" button in "Update Available Now" modal is too close to window border
### UX Description

As can be seen from the screenshot below, the "Update now" button is touching the border of the modal window, maybe it should be a bit to the left?

Image

### Request type

UX analysis/suggestions for improvement

### Primary Contact

_No response_

### Deadline for request

_No response_

MarsKubeX:
### What does this PR do?

Remove unused div causing incorrect margin in 'Update Available' modal action buttons.

### Screenshot / video of UI

Captura de pantalla 2025-06-03 a las 15 01 08

Fixes #9896 

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/dialogs/Dialog.svelte
+++ b/packages/renderer/src/lib/dialogs/Dialog.svelte
@@ -25,9 +25,9 @@ export let onclose: () => void = () => {
 
   <div class="px-5 py-5 mt-2 flex flex-row w-full space-x-5">
     {#if $$slots.validation}
-    <div class="grow">
-      <slot name="validation" />
-    </div>
+      <div class="grow">
+        <slot name="validation" />
+      </div>
     {/if}
     <slot name="buttons" />
   </div>
--- a/packages/renderer/src/lib/dialogs/Dialog.svelte
+++ b/packages/renderer/src/lib/dialogs/Dialog.svelte
@@ -24,9 +24,11 @@ export let onclose: () => void = () => {
   </div>
 
   <div class="px-5 py-5 mt-2 flex flex-row w-full space-x-5">
+    {#if $$slots.validation}
     <div class="grow">
       <slot name="validation" />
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/dialogs/Dialog.svelte
```diff
@@ -24,9 +24,11 @@ export let onclose: () => void = () => {
   </div>
 
   <div class="px-5 py-5 mt-2 flex flex-row w-full space-x-5">
+    {#if $$slots.validation}
```
axel7083, line 27: question: who is providing the validation slot?

axel7083 (approved):
LGTM, might be nice to show `before / after` screenshots in the PR, so we better understand the change.

Testing such thing is kinda tricky, so on my side I am fine with skipping it, since this is pretty trivial.

packages/renderer/src/lib/dialogs/Dialog.svelte
benoitf, line 27: I think there is a formatting issue (div/slot) should be shifted for the lines below)

I have the feeling that the proposal is a workaround. I mean, what happens if there is validation slot, the issue will still be there no ?

so should it reduce the text of some buttons, handle the fix when there is validation turned on, make the default width longer ?

MarsKubeX: Fixed the wrong formatting. 
The parent component (Dialog) caused the layout issue. The content is displayed via the MessageBox component (which isn't modified here) and it simply shows messages, so it doesn't use the validation slot. If MessageBox ever needs validation support, I can add width then, but this fix aligns with its current use case.

cdrage (approved):

benoitf (approved):

PR #12729, opened 2025-06-03
feat: allow provider update from resources page

### LINKED ISSUE #12189: Ability to update providers from the Resources page

After #12112 you can see when there is an update available to a provider via the status bar. Unfortunately, clicking on it brings you to the Settings > Resources page where you can't see or apply updates.

Add the ability to see when an update is available and apply it, similar to the CLI Tools page (or Dashboard).

This is covered in #12012 as well, but since that is a larger redesign this may be deliverable as one step.

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
Adds the ability to update providers from the resources page by adding the `ProviderUpdateButton` used in the dashboard page.

- Possible follow up issue: some providers like Podman and OpenShift Local have preflight check that show up on the dashboard if there is a problem with them. Will we include them in the resource page if we remove the providers from the dashboard page? and if so, how will we show them?

### Screenshot / video of UI
Screenshot From 2025-06-03 09-53-28

Closes https://github.com/podman-desktop/podman-desktop/issues/12189

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesResourcesRendering.spec.ts
@@ -808,3 +808,21 @@ test('Expect to see the no resource message when there is no providers', async (
   const panel = screen.getByLabelText('no-resource-panel');
   expect(panel).toBeInTheDocument();
 });
+
+test('Expect update button to show up when an update is available to a new version', async () => {
+  vi.mocked(window.runUpdatePreflightChecks).mockResolvedValue(true);
+  vi.mocked(window.updateProvider).mockResolvedValue([{ name: 'podman', status: true }]);
+  providerInfos.set([providerInfo]);
+  const component = render(PreferencesResourcesRendering, {});
+  let updateButton = screen.queryByText('Update to');
+  expect(updateButton).not.toBeInTheDocument();
+
+  providerInfos.set([{ ...providerInfo, version: '0.0.1', updateInfo: { version: '0.0.2' } }]);
+  await component.rerender({});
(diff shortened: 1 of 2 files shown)


benoitf (commented):
I thought I commented but it seems I did not click on the submit review button

it seems that there is a lot of changes regarding the classes just to include the button later

IMHO I would have expected a change around the layout first (b/c there are no easy tests)

then we can compare if it's changing something or not

and after that, you can add the lines

```
{#if provider.updateInfo?.version && provider.version !== provider.updateInfo?.version}
                    <ProviderUpdateButton onPreflightChecks={(checks): CheckStatus[] => (preflightChecks = checks)} provider={provider} />
                  {/if}
```

that will include the button

SoniaSandler:
> IMHO I would have expected a change around the layout first (b/c there are no easy tests)

@benoitf In a separate PR?

SoniaSandler:
@benoitf I removed most of the changes to the classes from the PR. One thing that I did notice when adding the button is that there might be cases where the divider for some of the cards is not aligned with the rest of them in certain screen sizes. It's expected due the fact that the size of the left column is set to `class="min-w-[170px] max-w-[200px]"`.
If we're okay with this design-wise, there is no need to add the other classes I originally did
Screenshot 2025-06-13 at 12 41 54 AM

benoitf (approved):

Firewall:
Great job on this @SoniaSandler! Just tried it out and it works perfectly! 🚀

benoitf (approved):

dgolovin (approved):
Code works and LGTM.

PR #12741, opened 2025-06-04
fix: error during extension install overlaps with other components

### LINKED ISSUE #10898: Fix Error layout from the extension installation in the extension's catalog
### Bug description

If there is an error during extension installation, its not properly aligned/framed. please see pics:

Image

Image

### Operating system

Windows 10

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. Install multiple extension at once in a fast manner.
2. Minikube, AI Lab, bootc, RH Extensions Pack (this one is important), Image Layers, etc.
3. This is not expected workflow but it should show the errors.

I would expect error, that is especially long, to be better shown withing the extension catalog card and in the details.

### Relevant log output

```shell

```

### Additional context

We might consider better handling of parallel installation of the extension, but it is still not a standard process of installing extensions, imho.

jiridostal:
### What does this PR do?

When an error occurs during the installation of an extension, the error message is currently misaligned and can overlap with other UI components, especially when the message is lengthy. 

### Screenshot / video of UI

Before:
image
After:
image

Fixes #10898

### How to test this PR?
Install multiple extension at once in a fast manner.

1. Minikube, AI Lab, bootc, RH Extensions Pack (this one is important), Image Layers, etc.
2. Should show the errors.

The error display has been moved outside of the <Button\> component. This ensures that the tooltip's hover behavior is independent of the button

### DIFF
--- a/packages/renderer/src/lib/featured/FeaturedExtensionDownload.svelte
+++ b/packages/renderer/src/lib/featured/FeaturedExtensionDownload.svelte
@@ -71,6 +71,7 @@ async function installExtension(): Promise<void> {
 }
 </script>
 
+<ErrorMessage icon wrapMessage class="-top-[15px] right-0 absolute" error={errorInstall}/>
 <button
   aria-label="Install {extension.id} Extension"
   on:click={installExtension}
@@ -89,7 +90,4 @@ async function installExtension(): Promise<void> {
     class:hidden={!installInProgress}
     class="absolute -top-[15px] right-0 text-[var(--pd-action-button-spinner)]"
     style="font-size: 8px">{percentage}</span>
-  <div class:hidden={!errorInstall} class="absolute w-56 -top-[25px] right-0" style="font-size: 8px">
-    <ErrorMessage error={errorInstall} />
-  </div>
 </button>


packages/renderer/src/lib/featured/FeaturedExtensionDownload.svelte
```diff
@@ -71,6 +71,7 @@ async function installExtension(): Promise<void> {
 }
 </script>
 
+<ErrorMessage icon wrapMessage class="-top-[15px] -right-0 absolute" error={errorInstall}/>
```
axel7083: question (non-blocking): the classes seems a bit odd to me? compare to the one bellow you added a `-` before `right-0`, is there any reason?

axel7083 (approved):
Tested and LGTM, (easier way is just to unplug internet) before clicking on download to test

> I test would be appreciated, but this is kinda tricky to test :/ we could check that the ErrorMessage component is called with the `icon` prop, but this is kinda implementation specific

packages/renderer/src/lib/featured/FeaturedExtensionDownload.svelte
jiridostal: That must be a left-over from alignment experiments. Fixing the negative-zero value to a positive-zero :-)

MarsKubeX (approved):
LGTM, just take a look to the pending check

PR #12743, opened 2025-06-04
chore: track all changes from the extension

benoitf:
### What does this PR do?
now that it's no longer tracking initial files on the watcher from a previous PR we can track all changes once we create the watcher. It allows to detect more cases where for example vite will remove a folder or recreate only one file, etc

### How to test this PR?

there are unit tests but you can try to launch one extension in development mode and do changes while having the watch mode, and it should always refresh/reload the extension and be in a consistent state (and not miss some updates)

### DIFF
--- a/packages/main/src/plugin/extension/extension-watcher.spec.ts
+++ b/packages/main/src/plugin/extension/extension-watcher.spec.ts
@@ -59,6 +59,8 @@ const fakeActivatedExtension = {
 } as unknown as ActivatedExtension;
 
 const fileSystemOnDidChangeMock = vi.fn();
+const fileSystemOnDidCreateMock = vi.fn();
+const fileSystemOnDidDeleteMock = vi.fn();
 const fileSystemDisposeMock = vi.fn();
 
 beforeEach(() => {
@@ -68,6 +70,8 @@ beforeEach(() => {
   // mock createFileSystemWatcher
   vi.mocked(fileSystemMonitoring.createFileSystemWatcher).mockReturnValue({
     onDidChange: fileSystemOnDidChangeMock,
+    onDidCreate: fileSystemOnDidCreateMock,
+    onDidDelete: fileSystemOnDidDeleteMock,
     dispose: fileSystemDisposeMock,
   } as unknown as FileSystemWatcher);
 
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/extension/extension-watcher.spec.ts
```diff
@@ -93,6 +109,18 @@ describe('monitor', () => {
...
     await vi.waitFor(() => expect(reloadSpy).toHaveBeenCalled());
+
+    // clear the calls on reloadSpy
+    reloadSpy.mockClear();
+
+    // check with onDidDelete
+    const onDidDeleteCallback = fileSystemOnDidDeleteMock.mock.calls[0]?.[0];
+    expect(onDidDeleteCallback).toBeDefined();
+    // call the callback
+    onDidDeleteCallback({ fsPath: 'fakePath' });
```
axel7083, line 120: just curious

how are you able to call `onDidDeleteCallback` directly without using `onDidDeleteCallback?` when at line 117, you use the `?.[0]` operator?

what is the type of onDidDeleteCallback ? should be something like `( () => void ) | undefined` ?

axel7083: usually I use `assert` instead of expect(<>).tobeDefined in similar scenario

axel7083 (approved):
LGTM

packages/main/src/plugin/extension/extension-watcher.ts
```diff
@@ -95,6 +95,17 @@ export class ExtensionWatcher implements IDisposable {
...
+    extensionWatcher.onDidCreate(async event => {
+      callback(event);
+    });
+    extensionWatcher.onDidDelete(async event => {
+      callback(event);
+    });
+
+    extensionWatcher.onDidChange(async event => {
+      callback(event);
     });
```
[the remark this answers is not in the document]
axel7083: this is such an ugly code

packages/main/src/plugin/extension/extension-watcher.spec.ts
benoitf: > how are you able to call onDidDeleteCallback directly without using onDidDeleteCallback? when at line 117, you use the ?.[0] operator?

b/c it returns any

packages/main/src/plugin/extension/extension-watcher.ts
```diff
@@ -95,6 +95,17 @@ export class ExtensionWatcher implements IDisposable {
       if (!hasTsConfig || !isInSource) {
         this.reloadExtension(extension);
       }
+    };
+
+    extensionWatcher.onDidCreate(async event => {
```
axel7083: question: why using async here?

packages/main/src/plugin/extension/extension-watcher.ts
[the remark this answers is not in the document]
axel7083:
```ts
    extensionWatcher.onDidCreate(callback.bind(this));
    extensionWatcher.onDidDelete(callback.bind(this));
    extensionWatcher.onDidChange(callback.bind(this));
```

would be more pleasing if we really are looking to optimize something like that

packages/main/src/plugin/extension/extension-watcher.ts
benoitf: I think I should remove it, I left the existing line that was already 
`extensionWatcher.onDidChange(async event => {` (and copy/paste it)

MarsKubeX:
LGTM, but didn't test it

packages/main/src/plugin/extension/extension-watcher.spec.ts
axel7083: Oh, yeah, that's poorly typed

packages/main/src/plugin/extension/extension-watcher.ts
benoitf: removed

PR #12766, opened 2025-06-07
style: update font loading and fallback in custom CSS

### LINKED ISSUE #12613: Website: Manrope font is TTF (slow)
The Podman Desktop website currently loads the Manrope font from `static/fonts/manrope.ttf`. Websites should avoid using TTF fonts.

See https://github.com/podman-desktop/podman-desktop/blob/main/website/src/css/custom.css#L10-L15

Using TTF files for web fonts is problematic due to several technical and performance issues. They lack compression, resulting in significantly larger file sizes than modern web font formats, often 2- 3× larger than WOFF2 equivalents, directly impacting page load times and user experience. TTF files also have limited browser optimization support and don't benefit from the advanced compression algorithms available in newer formats.

## Key pain points

- Poor loading performance due to large file sizes
- Increased bandwidth consumption, especially problematic for mobile users
- Lack of browser-specific optimizations
- No built-in font subsetting capabilities
- Higher server costs due to larger asset sizes

1. **Convert to WOFF2** - Modern browsers (95%+ support) benefit from superior compression
2. **Implement font subsetting** - Include only the characters/glyphs actually used on your site
3. **Use font-display: swap** - Prevent invisible text during font load
4. **Consider system font stacks** - Leverage OS-native fonts for critical text to eliminate network requests entirely

WOFF2 should be your primary format, with WOFF as a fallback for older browsers.

For maximum performance, let's evaluate whether custom fonts are necessary, or if we can use system fonts like `-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif`.

Note: The font is also on [Google Fonts](https://fonts.google.com/specimen/Manrope).

statickidz:
# What does this PR do?

This PR improves the website's font loading performance and compatibility by:
- Adding `font-display: swap` for improved font loading performance
- Updating font source to include WOFF2 format for better compression
- Enhancing font-family fallback stack for broader compatibility
- Adding font subset for Latin characters to reduce file size

# Screenshot / video of UI

No visual changes to the UI.

# What issues does this PR fix or reference?

- Closes #12613 

# How to test this PR?

1. Build and run the website locally:

2. Verify the following:
   - Website loads without any font-related console errors
   - Fonts load correctly across different browsers
   - No visual regressions in the website's appearance
   - Check network tab to confirm the new WOFF2 font file is being loaded
   

vancura:
Works perfectly! Thanks!

vancura (approved):

website/src/css/custom.css
```diff
@@ -6,17 +6,22 @@
...
-/* manrope variable font */
+/* Manrope latin subset font */
 @font-face {
   font-family: 'Manrope';
   font-style: normal;
   font-weight: normal;
-  src: url('../../static/fonts/manrope.ttf') format('truetype');
+  font-display: swap;
+  src: url('../../static/fonts/manrope-subset-latin.woff2') format('woff2');
+  unicode-range:
```
cdrage, line 35: I'm unfamiliar with regards to the unicode-range being added haha, do you mind adding a small comment to above here so we know what this does?

cdrage (approved):
Checking this locally (I had to approve the workflows running), this LGTM.

I'm assuming that the binary files were also downloaded straight from https://fonts.google.com/specimen/Manrope ?

This LGTM in terms of fallback.

website/src/css/custom.css
```diff
@@ -6,17 +6,22 @@
...
+    U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+0304, U+0308, U+0329, U+2000-206F, U+20AC,
+    U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
 }
 
 /* You can override the default Infima variables here. */
 :root {
-  --ifm-font-family-base: 'Manrope', sans-serif;
+  --ifm-font-family-base: 'Manrope', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu,
```
cdrage: I had to Google WHY it's in this particular order.. but I see that it's a common design trope to organize it this way. 

LGTM!

website/src/css/custom.css
statickidz: Thanks for the review @cdrage!

Yes, the font is from google fonts, the `unicode-range` property tells the browser which characters this font should be used for, in our case, we are using it to create a subset that only includes Latin characters and common symbols, I've added much more explained on the code comment!

website/src/css/custom.css
statickidz: Yes, it is a pretty common use and very boilerplate code!

cdrage (approved):
LGTM thank you for the explanation as well as the added comment!

statickidz:
@cdrage @vancura it seems like the Argos step is failing because some texts are not the same as prod, can we manually skip the validation?

website/src/css/custom.css
```diff
@@ -6,17 +6,41 @@
...
+    Cantarell, 'Helvetica Neue', sans-serif;
```
benoitf: I don't see why we would need this change as the font is embedded

statickidz: I added this to explicitly define the fallback stack to ensure better consistency across browsers if Manrope fails to load for some reason or the network is slow to prevent layout shifts and so

vancura: There are a couple of reasons to implement this, including consistency and best practices. However, one stands out: when Apple devices operate in higher-security mode (Lockdown Mode), no fonts are loaded to prevent code injection. Additionally, some users block fonts from loading altogether. In these situations, if the embedded Manrope font is unavailable, the next best option will be loaded instead of the browser's default sans-serif font.

benoitf: @vancura in lockdown mode, javascript is disabled as well so you won't get something nice (for example the hamburger menu won't work, meaning you'll be stuck on the page.


so IMHO instead of just copying/pasting some lines, there is need for a good reason. Here it's just plate of noodles.

benoitf: google.com only has `Arial, sans-serif`
youtube.com  `Roboto, Arial, sans-serif`

benoitf (changes requested):
`--ifm-font-family-base` shouldn't be changed

also it's unrelated to the scope of the issue

website/src/css/custom.css
vancura: @benoitf Actually, JS isn't completely disabled in iOS Lockdown Mode. Only JIT compilation is disabled, making JS run slower but still functional. So the hamburger menu would still work.

However, web fonts are blocked in Lockdown Mode for security reasons, which is exactly why this font fallback stack is valuable. Users get a functional site with proper typography instead of falling back to generic `sans-serif`. This change addresses a real security/UX scenario without adding complexity to the codebase.

But you raised a good point about the site when JS is disabled: that actually highlights a critical issue with our website that we should address eventually. If users can't navigate our site when JS is disabled (hamburger menu not working, being "stuck on the page"), that's a significant accessibility and usability problem beyond just font loading. Users might have JS disabled for various reasons: security concerns, slow connections, assistive technologies, or corporate firewalls.

While this font fallback change is a small step toward better resilience, your comment suggests we should have a broader conversation about progressive enhancement and ensuring our site remains functional without JS. That's a separate but important discussion for the future.

For now, this font stack change still provides value for the scenarios I mentioned, regardless of the JS situation.

benoitf: Chrome (or other browser) will pickup the right font from the Operating System.

benoitf: it's better to defer to browsers rather than having an unmaintained list of fonts

vancura: @benoitf I understand the concern, but there's a technical distinction here:

Generic `sans-serif` behavior:
- Chrome defaults to Arial on Windows, Helvetica on macOS
- Firefox might default to Liberation Sans or DejaVu Sans
- Different browsers = different fonts on the same OS
- Not necessarily the actual system font

System font stack behavior:
- `-apple-system` explicitly requests San Francisco (Apple's actual system font)
- `Segoe UI` explicitly requests Windows' actual system font
- Results in consistent, native typography across all browsers

This isn't an "unmaintained list" – it's the industry standard system font stack used by GitHub, Bootstrap, and most modern web frameworks. These are OS-maintained system fonts, not custom web fonts that could become outdated.

The goal is getting the actual system font (San Francisco, Segoe UI, etc.) rather than each browser's interpretation of what "sans-serif" should be.

You can test this: compare how `font-family: sans-serif` vs `font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', system-ui, sans-serif` renders across different browsers on the same machine - you'll see the difference in consistency.

benoitf: >This isn't an "unmaintained list" 

it is because it would be part of the source code of this website. So it's unmaintained.

For the website, we're using Infima (docusaurus default) and TailwindCSS. So it should leverage these frameworks and reuse all defaults, not try to do our own stuff there. (not trying to reinvent the wheel)

statickidz: all right, it seems there is something here to open a new issue and discuss it, let's keep it as it was for now

statickidz:
> `--ifm-font-family-base` shouldn't be changed
> 
> also it's unrelated to the scope of the issue

it should be restored now!

benoitf (approved):

benoitf:
thanks @statickidz !

PR #10500, opened 2024-12-31
fix: cleanup log files before starting the app

benoitf:
### What does this PR do?
macOS pList files are appending data to previous log files. Ensure to clean (truncate) the log file before starting the application. It ensures the log files won't grow indefinitely 

### What issues does this PR fix or reference?

fixes https://github.com/podman-desktop/podman-desktop/issues/10120

### How to test this PR?

on macOS only:
using a production binary installed in /Applications folder, disable and enable again the startup on Login in preferences
logout from your current macOS session and login

check that the log files have been wiped truncated before the app is starting

- [x] Tests are covering the bug fix or the new feature

the file was not tested, added unit tests

PR #12768, opened 2025-06-08
fix: minimize on startup plist for macOS

### LINKED ISSUE #12562: The dashboard always opens at startup
### Bug description

The dashboard always opens when the system starts. The setting "Minimize on login" does not affect the behavior.

### Operating system

macOS 15.5 (Apple silicon)

### Installation Method

### Version

1.18.0

### Steps to reproduce

Just install the package from the official site, enable the "Minimize on login" setting and restart the system.

### Relevant log output

```shell

```

### Additional context

_No response_

cblecker:
### What does this PR do?
This feature broke due to #10500, as the minimize flag was not being actually passed to the podman desktop process, but bash.

fixes #12562

### How to test this PR?

### DIFF
--- a/packages/main/src/system/macos-startup.ts
+++ b/packages/main/src/system/macos-startup.ts
@@ -64,7 +64,7 @@ export class MacosStartup {
     // so we check the configuration within the function
     const preferencesConfig = this.configurationRegistry.getConfiguration('preferences');
     const minimize = preferencesConfig.get<boolean>('login.minimize');
-    const minimizeSettings = minimize ? '<string>--minimize</string>' : '';
+    const minimizeSettings = minimize ? '--minimize' : '';
 
     // comes from a volume ? do nothing
     if (this.podmanDesktopBinaryPath.startsWith('/Volumes/')) {
@@ -76,8 +76,7 @@ export class MacosStartup {
     const stderrPath = `${app.getPath('home')}/Library/Logs/Podman Desktop/launchd-stderr.log`;
 
     // write the file
-    const plistContent =
-      `<?xml version="1.0" encoding="UTF-8"?>
+    const plistContent = `<?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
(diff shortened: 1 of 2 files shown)


MarsKubeX (approved):
LGTM (not tested)

benoitf (approved):

benoitf (commented):
thanks @cblecker could you add a unit test

packages/main/src/system/macos-startup.ts
```diff
@@ -87,10 +86,7 @@ export class MacosStartup {
     <array>
         <string>/bin/bash</string>
         <string>-c</string>
-        <string>/usr/bin/truncate -s 0 '${stdoutPath}'; /usr/bin/truncate -s 0 '${stderrPath}'; '${this.podmanDesktopBinaryPath}'</string>
-        ` +
-      minimizeSettings +
-      `
+        <string>/usr/bin/truncate -s 0 '${stdoutPath}'; /usr/bin/truncate -s 0 '${stderrPath}'; '${this.podmanDesktopBinaryPath}' ${minimizeSettings}</string>
```
dgolovin, line 89: This change breaks formatting in file and triggers change in line 79 above. Line 67 should be the only change.

dgolovin (changes requested):

packages/main/src/system/macos-startup.ts
cblecker: The cause of the bug is that the `--minimize` is in a separate `<string></string>` array item, so bash interprets it as an argument to itself, rather than an argument to the podman desktop binary. It needs to be inside the same `<string></string>` as the binary path.

If I only made the L67 change, the resulting plist would be:

```
    <array>
        <string>/bin/bash</string>
        <string>-c</string>
        <string>/usr/bin/truncate -s 0 '~/Library/Logs/Podman Desktop/launchd-stdout.log'; /usr/bin/truncate -s 0 '~/Library/Logs/Podman Desktop/launchd-stderr.log'; '/Applications/Podman Desktop.app/Contents/MacOS/Podman Desktop'</string>
        --minimize
    </array>
```

This would be a malformed plist XML file, as the `--minimize` wouldn't be inside a string at all. Open to other suggestions at how to better format this, but I believe your suggestion as I understand it would not work.

cblecker:
@benoitf I would appreciate some assistance to get a test written for this, as I don't write much in typescript and may not have the cycles to learn in a timely manner. Was just trying to fix the functional bug. Please feel free to carry the PR if that's easier (credit not needed, just trying to make podman desktop easier to use :) )

packages/main/src/system/macos-startup.ts
dgolovin: My fault.

dgolovin (approved):

dgolovin:
> @benoitf I would appreciate some assistance to get a test written for this, as I don't write much in typescript and may not have the cycles to learn in a timely manner. Was just trying to fix the functional bug. Please feel free to carry the PR if that's easier (credit not needed, just trying to make podman desktop easier to use :) )

@cblecker Not sure why I cannot see your your fork when trying to send PR. Here is my commit with possible test https://github.com/podman-desktop/podman-desktop/commit/86279dfd902787b478bbd53b210ff8f3dacc28b0

cblecker:
Thank you @dgolovin ! Yeah, that test seems to work like a charm

benoitf:
~~@cblecker it looks like the test is missing (the one suggested by Denis~~

now I see it, nevermind

benoitf:
thanks @cblecker

PR #12777, opened 2025-06-09
fix: flaky test in BuildImageFromContainerfile.spec.ts

### LINKED ISSUE #10405: flaky test in `ContainerDetailsTerminal.spec.ts:236:55` on macos

dgolovin:
The fix removes explicit 100000 timeout for test method and a 'magic' takskId used in test.

It turned out the taskId was incorrect for first try and only passed for a consecutive retry.
The fix makes it to work without retry. 

### What does this PR do?

Eliminates one of the flaky tests.

Fix #10405.

### How to test this PR?

comment out `retry:3,` from `test` section of `vite.config.js` configuration and run it with command below:

```
npx vitest --run --project=renderer BuildImageFromContainerfile.spec.ts
```

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.spec.ts
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.spec.ts
@@ -26,7 +26,7 @@ import { get } from 'svelte/store';
 import { beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
 import BuildImageFromContainerfile from '/@/lib/image/BuildImageFromContainerfile.svelte';
-import { buildImagesInfo } from '/@/stores/build-images';
+import { buildImagesInfo, getNextTaskId } from '/@/stores/build-images';
 import { providerInfos } from '/@/stores/providers';
 import { recommendedRegistries } from '/@/stores/recommendedRegistries';
 import type { ProviderContainerConnectionInfo, ProviderInfo } from '/@api/provider-info';
@@ -554,7 +554,8 @@ test('Expect build page to load a state for a build requested by taskId', async 
   expect(buildButton).toBeVisible();
   expect(cancelButton).not.toBeVisible();
 
-  await rendering.rerender({ taskId: 9 });
+  //
+  await rendering.rerender({ taskId: getNextTaskId() - 2 });
 
   await waitFor(() => {
     expect(containerFilePath).not.toBeVisible();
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/image/BuildImageFromContainerfile.spec.ts
```diff
@@ -554,7 +554,8 @@ test('Expect build page to load a state for a build requested by taskId', async
   expect(buildButton).toBeVisible();
   expect(cancelButton).not.toBeVisible();
 
-  await rendering.rerender({ taskId: 9 });
+  //
+  await rendering.rerender({ taskId: getNextTaskId() - 2 });
```
jeffmaury, line 558: What is the rationale for this - 2 ?

dgolovin: Current counter for taskId cannot be read from outside of the store module. Lets call it `idBeforeTest1`. TaskId increments with every started build.
The test runs one build (id = `idBeforeTest1` + 1), then another build (id = `idBeforeTest1` + 2). Then it needs to load state of the first build but there is only function to get next ID in sequence and it gives `idBeforeTest1`+ 3. So to get taskId for first build we have to subtract 2.

jeffmaury (approved):
LGTM

benoitf (approved):

PR #12781, opened 2025-06-10
feat: allow changing between rootless and rootful connection for an existing Podman machine

### LINKED ISSUE #12498: Ability to switch rootless/rootful mode for a podman machine from the UI

AFAIK it's not possible.

I can use a switch when I create it but not be able to switch once it's created

add a toggle so I can change the status ?

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
This PR adds the option to switch between rootless or rootful connections on an existing Podman machine from the UI by exposing the rootful value in the editable properties of Podman machines. (works only on macOS as it is the only OS where the connection is editable)

- The rootful property was hidden so it would not show up on the provider card as it messes up the layout:
Screenshot 2025-06-09 at 9 14 41 PM

A possible follow up could be adding a way to indicate if the machine is rootful or rootless.

### Screenshot / video of UI

https://github.com/user-attachments/assets/ad5347e6-89fd-4f62-98bd-73234450fc25

Closes [#12498](https://github.com/podman-desktop/podman-desktop/issues/12498)

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/package.json
+++ b/extensions/podman/packages/extension/package.json
@@ -112,6 +112,14 @@
           "description": "Disk size usage",
           "readonly": true
         },
+        "podman.machine.rootful": {
+          "type": "boolean",
+          "default": true,
+          "scope": "ContainerConnection",
+          "description": "Machine with root privileges",
+          "when": "podman.isRootfulMachineInitSupported == true",
+          "hidden": true
+        },
         "podman.factory.machine.name": {
           "type": "string",
           "default": "podman-machine-default",
--- a/extensions/podman/packages/extension/package.json
+++ b/extensions/podman/packages/extension/package.json
@@ -117,7 +117,7 @@
           "default": true,
           "scope": "ContainerConnection",
           "description": "Machine with root privileges",
-          "when": "podman.isRootfulMachineInitSupported == true",
+          "when": "podman.podmanMachineEditRootfulSupported",
           "hidden": true
         },
         "podman.factory.machine.name": {
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -763,13 +763,19 @@ export async function registerProviderFor(
   socketPath: string,
 ): Promise<void> {
   const isHyperVMachine = extensionApi.env.isWindows && machineInfo.vmType === VMTYPE.HYPERV;
+
   const isEditMemorySupported = extensionApi.env.isMac || isHyperVMachine;
   const isEditCPUSupported = extensionApi.env.isMac || isHyperVMachine;
   const isEditDiskSizeSupported = extensionApi.env.isMac;
+  const isEditRootfulSupported = extensionApi.env.isMac || extensionApi.env.isWindows;
+
+  const isEditSupported =
+    isEditMemorySupported || isEditCPUSupported || isEditDiskSizeSupported || isEditRootfulSupported;
 
   extensionApi.context.setValue(PODMAN_MACHINE_EDIT_MEMORY, isEditMemorySupported);
   extensionApi.context.setValue(PODMAN_MACHINE_EDIT_CPU, isEditCPUSupported);
   extensionApi.context.setValue(PODMAN_MACHINE_EDIT_DISK_SIZE, isEditDiskSizeSupported);
+  extensionApi.context.setValue(PODMAN_MACHINE_EDIT_ROOTFUL, isEditRootfulSupported);
 
   const lifecycle: extensionApi.ProviderConnectionLifecycle = {
     start: async (context, logger): Promise<void> => {
@@ -784,8 +790,8 @@ export async function registerProviderFor(
       });
     },
   };
-  //support edit only on MacOS and HyperV machines as Podman WSL is nop and generates errors
-  if (isEditMemorySupported || isEditCPUSupported || isEditDiskSizeSupported) {
+  // support edit only on MacOS and Windows with limited editing capabilities for HyperV and WSL machines
+  if (isEditSupported) {
     lifecycle.edit = async (context, params, logger, _token): Promise<void> => {
       let effective = false;
       const args = ['machine', 'set', machineInfo.name];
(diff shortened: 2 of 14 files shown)


extensions/podman/packages/extension/package.json
```diff
@@ -109,6 +109,14 @@
           "description": "Disk size usage",
           "readonly": true
         },
+        "podman.machine.rootful": {
+          "type": "boolean",
+          "default": true,
+          "scope": "ContainerConnection",
+          "description": "Machine with root privileges",
+          "when": "podman.isRootfulMachineInitSupported == true",
```
axel7083: question: I am wondering how does this would work on Linux and with remote connection? This seems to have a generic scope

SoniaSandler: @axel7083 the editing capability of podman machines from Podman Desktop is limited to only MacOS in the podman extension
```
//support edit only on MacOS as Podman WSL is nop and generates errors
  if (extensionApi.env.isMac)
```
so nothing should show up on Linux, but I'll have to check for remote connections

axel7083: On windows, you can use `podman machine set --rootful=false`, if this PR only work on mac, this has a pretty limited scope :/ as we can change the rootfulness of a machine on windows too, however, you need to stop before changing the property

SoniaSandler: @axel7083 the edit action stops the machine before applying the changes, and then starts it again after the changes have been applied. I don't have a windows machine to check if it causes any problems, but I can update the PR and someone with windows could check it

SoniaSandler: @benoitf should the rootful property be editable on Windows from Podman Desktop? 
It looks like other properties can be edited only on mac and I don't have a Windows machine to check if it works for rootful/rootless changes from within Podman Desktop

danivilla9: Hello @SoniaSandler, I've just checked your branch on a windows machine. Running `podman machine set --rootful=false [machineName]` seems to update the label on the summary tab properly (you have to refresh the page though) and the machine is able to be managed from PD seamlessly. As you pointed out the edit feature is not available in Windows, but imo it would be great to have it even if it's only for the rootful/rootless switch.

Feel free to ping me if you need more checks on windows

SoniaSandler: @danivilla9 thanks!

SoniaSandler: @danivilla9 I updated the PR. Now on Windows, you should see in the resources page the option to edit existing Podman connections. When you go to the edit page, there should only be the rootless/rootful option. I'm not sure if there is a different behavior between WSL and HyperV for this operation, so maybe both of them should be check with this change.

danivilla9 (approved):
@SoniaSandler Hello! Just finished checking your updates on WSL and HyperV. Everything seems to work fine as you described in both providers, the only thing of note is that PD seems to be a bit slow to pick up on the rootless/rootful changes, so sometimes if you go to the details of the machine or to the edit page again, you see an unexpected state. Not sure if there is anything to do about it, or even if that's something relevant, but I wanted to point it out. I made a recording:
[Screencast From 2025-06-19 12-38-59.webm](https://github.com/user-attachments/assets/dc987afe-9bf1-44df-a1a4-de461944be24)

axel7083 (commented):
Tried it on Window 11 works as expected.

question: the task created when updating is `Creating <machine> provider` is there a way to have a better message when updating?

image

could be another PR/issue

extensions/podman/packages/extension/package.json
```diff
@@ -100,14 +104,24 @@
...
+          "when": "isMac"
         },
         "podman.machine.diskSizeUsage": {
           "type": "number",
           "format": "diskSizeUsage",
           "scope": "ContainerConnection",
           "description": "Disk size usage",
-          "readonly": true
+          "readonly": true,
+          "when": "isMac"
```
axel7083: suggestion: For diskusage, memory, and CPU, might be better to have this in a dedicated PR

axel7083 (approved):
LGTM

cdrage:
Question! Isn't it required to restart the machine for rootful status to take effect?  Or did that change recently with upstream podman?

SoniaSandler:
@cdrage you're right, and it's already done in the existing code when editing a podman machine:
```
if (effective) {
        const state = podmanMachinesStatuses.get(machineInfo.name);
        try {
          if (state === 'started') {
            await lifecycle.stop?.(context, logger);
          }
          await execPodman(args, machineInfo.vmType, {
            logger: new LoggerDelegator(context, logger),
          });
        } finally {
          if (state === 'started') {
            await lifecycle.start?.(context, logger);
          }
        }
      }
```

cdrage:
> @cdrage you're right, and it's already done in the existing code when editing a podman machine:
> 
> ```
> if (effective) {
>         const state = podmanMachinesStatuses.get(machineInfo.name);
>         try {
>           if (state === 'started') {
>             await lifecycle.stop?.(context, logger);
>           }
>           await execPodman(args, machineInfo.vmType, {
>             logger: new LoggerDelegator(context, logger),
>           });
>         } finally {
>           if (state === 'started') {
>             await lifecycle.start?.(context, logger);
>           }
>         }
>       }
> ```

Ah thank you! I saw in the initial video on the PR it doesn't show the podman machine stopping / starting again? Unless it did it really fast.

SoniaSandler:
@cdrage I just moved to the images page right after the edit, but you can see in the statusbar provider tooltip in the video that the machine is starting again

cdrage (approved):
Testing again, the code LGTM :) Was able to switch between rootless and rootful on an existing machine.

Great work!

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -843,6 +864,10 @@ export async function registerProviderFor(
   await containerConfiguration.update('machine.memoryUsage', machineInfo.memoryUsage);
   await containerConfiguration.update('machine.diskSizeUsage', machineInfo.diskUsage);
 
+  if (isRootful !== undefined) {
```
axel7083: `isRootfulMachine` only return a boolean, can't be undefined

axel7083 (approved):
Tried on Windows 11 with two podman machine (WSL) - LGTM

PR #12786, opened 2025-06-10
chore: set libkrun as default provider on mac

### LINKED ISSUE #11453: make libkrun the default when creating a podman machine

switch from applehv to libkrun as default

I don't see any regression for now and it will still be a user choice 

set the containers.conf default provider as well if it's the first time or if there is no other machine existing
So it'll be in sync with what the cli is seeing

keep it like this but then people wanting to try tu run AI models need to redo machines afterwards etc.

### Additional context

_No response_

gastoner:
### What does this PR do?
Sets libkrun as default provider on mac

### Screenshot / video of UI

https://github.com/user-attachments/assets/590624ba-65a1-4ea9-9769-70fa7718363b

Closes #11453 

### How to test this PR?
Try to create machine

### DIFF
--- a/extensions/podman/packages/extension/src/checks/macos-checks.ts
+++ b/extensions/podman/packages/extension/src/checks/macos-checks.ts
@@ -100,3 +100,37 @@ export class MacPodmanInstallCheck extends BaseCheck {
     });
   }
 }
+
+export class MacKrunkitPodmanMachineCreationCheck extends BaseCheck {
+  title = 'Krunkit Installation';
+  async execute(): Promise<extensionApi.CheckResult> {
+    // we need to check if brew is installed to avoid unexpected error
+    try {
+      await extensionApi.process.exec('which', ['brew']);
+    } catch (err) {
+      // brew is not installed so do not check if krunkit has been installed with brew
+      return this.createSuccessfulResult();
+    }
+
+    // brew is installed, check if libKrun has been installed with brew
+    try {
+      const result = await extensionApi.process.exec('brew', ['list', '--verbose', 'krunkit'], {
+        env: { HOMEBREW_NO_AUTO_UPDATE: '1', HOMEBREW_NO_ANALYTICS: '1' },
+      });
+      if (!result.stderr) {
+        return this.createSuccessfulResult();
+      }
+    } catch (e: unknown) {
+      console.error(`Krunkit check failed with error: ${e}`);
+    }
+
+    return this.createFailureResult({
+      description:
+        'You have krunkit installed with "brew", run "brew update && brew upgrade krunkit" to install new version',
+      docLinks: {
+        url: 'https://docs.brew.sh/Manpage#upgrade-options-outdated_formulaoutdated_cask-',
+        title: 'Brew Documentation',
+      },
+    });
+  }
+}
--- a/extensions/podman/packages/extension/src/checks/macos-checks.ts
+++ b/extensions/podman/packages/extension/src/checks/macos-checks.ts
@@ -126,7 +126,7 @@ export class MacKrunkitPodmanMachineCreationCheck extends BaseCheck {
 
     return this.createFailureResult({
       description:
-        'You have krunkit installed with "brew", run "brew update && brew upgrade krunkit" to install new version',
+        'You have krunkit installed with "brew". Run "brew update && brew upgrade krunkit" to install new version',
       docLinks: {
         url: 'https://docs.brew.sh/Manpage#upgrade-options-outdated_formulaoutdated_cask-',
         title: 'Brew Documentation',
--- a/extensions/podman/packages/extension/src/checks/macos-checks.ts
+++ b/extensions/podman/packages/extension/src/checks/macos-checks.ts
@@ -126,7 +126,7 @@ export class MacKrunkitPodmanMachineCreationCheck extends BaseCheck {
 
     return this.createFailureResult({
       description:
-        'You have krunkit installed with "brew". Run "brew update && brew upgrade krunkit" to install new version',
+        'You have krunkit installed with "brew". Run "brew update && brew tap slp/krunkit && brew upgrade krunkit" to install new version',
       docLinks: {
         url: 'https://docs.brew.sh/Manpage#upgrade-options-outdated_formulaoutdated_cask-',
         title: 'Brew Documentation',
--- a/extensions/podman/packages/extension/src/extension.spec.ts
+++ b/extensions/podman/packages/extension/src/extension.spec.ts
@@ -754,6 +754,7 @@ describe.each([
 
 test.each([
   { architecture: 'arm64', expectedProvider: VMTYPE.LIBKRUN },
+  { architecture: 'arm64', expectedProvider: VMTYPE.APPLEHV },
   { architecture: 'x64', expectedProvider: VMTYPE.APPLEHV },
 ])('verify create on mac from settings on %s', async ({ architecture, expectedProvider }) => {
   vi.mocked(extensionApi.env).isMac = true;
--- a/extensions/podman/packages/extension/src/extension.spec.ts
+++ b/extensions/podman/packages/extension/src/extension.spec.ts
@@ -386,7 +386,7 @@ afterEach(async () => {
 });
 
 describe.each([
-  { os: 'macos', expectedProvider: VMTYPE.APPLEHV },
+  { os: 'macos', expectedProvider: VMTYPE.LIBKRUN },
   { os: 'windows-wsl', expectedProvider: VMTYPE.WSL },
   { os: 'windows-hyperv', expectedProvider: VMTYPE.HYPERV },
 ])('verify create on %s', ({ os, expectedProvider }) => {
--- a/extensions/podman/packages/extension/src/extension.spec.ts
+++ b/extensions/podman/packages/extension/src/extension.spec.ts
@@ -111,11 +111,12 @@ const machineInfo: extension.MachineInfo = {
   identityPath: '/path/to/key',
 };
 
+const updateMachineProviderSettingsMock = vi.fn();
 const podmanConfiguration = {
   registryConfiguration: {
     getPlaybookScriptPath: vi.fn(),
   },
-  updateMachineProviderSettings: vi.fn(),
+  updateMachineProviderSettings: updateMachineProviderSettingsMock,
 } as unknown as PodmanConfiguration;
 
 const machineDefaultName = 'podman-machine-default';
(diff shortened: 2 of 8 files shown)


gastoner:
@benoitf when discussing this issue with @jiridostal, he raised a point that there might be an issue if the user does not have libkrun & vrunkit. Both of which need to be installed before creating VrunKit-based machine

I suggest either: 
1. Switch the description to markdown and provide links to vrunkit and libkrun dependencies
2. Add a note, or error with some detection check that the user does not have vrunkit and libkrun installed (would be visible by default after opening the form if the user does not have them installed, both of those tools)

jeffmaury:
> @benoitf when discussing this issue with @jiridostal, he raised a point that there might be an issue if the user does not have libkrun & vrunkit. Both of which need to be installed before creating VrunKit-based machine
> 
> I suggest either:
> 
>     1. Switch the description to markdown and provide links to vrunkit and libkrun dependencies
> 
>     2. Add a note, or error with some detection check that the user does not have vrunkit and libkrun installed (would be visible by default after opening the form if the user does not have them installed, both of those tools)

We can perhaps add a pre check that issue a warning if vfkit and libkrun are not available and ask them to install Podman from installer and then remove the libkrun option

odockal:
This change will also impact a creating a machine from Dashboard, although there is no option to choose a provider. So if the user will not have libkrun/krunkit installed, it will fail for him.

benoitf (commented):
need to assert that krunkit binary is there

if first machine being created, the containers.conf file should be updated with libkrun so `podman machine list` will return/list that machine

SoniaSandler (commented):
Since I have brew, but I didn't install krunkit through it, I still get a warning on the machine creation page. 
However, I was able to create and start a libkrun machine because I installed Podman through the installer, so I feel like the warning might be a bit confusing in this case. 
Maybe the binary check could run only if Podman was installed using brew since it comes packaged with krunkit from the installer, or also check the Podman installation path for the binary.

gastoner:
@SoniaSandler should be updated 🤞

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -1857,6 +1858,19 @@ export async function connectionAuditor(items: extensionApi.AuditRequestItems):
     });
   }
 
+  const krunKitCheck = new MacKrunkitPodmanMachineCreationCheck().execute();
```
SoniaSandler: I get an error in the console 
```
main ↪️ [podman] Krunkit check failed with error: Command execution failed with exit code 1: Command execution failed with exit code 1
``` 
because this check is executed before checking if `podmanCheck` is successful

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -1857,6 +1858,19 @@ export async function connectionAuditor(items: extensionApi.AuditRequestItems):
...
+  const podmanCheck = new MacPodmanInstallCheck().execute();
+  // If is podman check not successful => installed with brew (does not contain krunkit & libkrun)
+  if (!(await podmanCheck).successful) {
+    // Podman is installed using installer from github (contains krunkit & libkrun)
```
SoniaSandler: nit: not sure what this comment means (L1865) since we go into the if block if Podman was installed with brew and not the installer

extensions/podman/packages/extension/src/checks/macos-checks.ts
```diff
@@ -100,3 +100,37 @@ export class MacPodmanInstallCheck extends BaseCheck {
...
+      if (!result.stderr) {
+        return this.createSuccessfulResult();
+      }
+    } catch (e: unknown) {
+      console.error(`Krunkit check failed with error: ${e}`);
+    }
+
+    return this.createFailureResult({
+      description:
+        'You have krunkit installed with "brew", run "brew update && brew upgrade krunkit" to install new version',
```
SoniaSandler: Doesn't getting here mean that there is no krunkit installed with brew (based on L120)?

SoniaSandler (approved):
LGTM, works as expected with Podman installed from the installer. Thanks for the changes! 
(added a small question about the wording of the failure result)

extensions/podman/packages/extension/src/checks/macos-checks.ts
gastoner: Your right :D I've updated the message

extensions/podman/packages/extension/src/checks/macos-checks.ts
```diff
@@ -100,3 +100,37 @@ export class MacPodmanInstallCheck extends BaseCheck {
...
+      if (!result.stderr) {
+        return this.createSuccessfulResult();
+      }
+    } catch (e: unknown) {
+      console.error(`Krunkit check failed with error: ${e}`);
+    }
+
+    return this.createFailureResult({
+      description:
+        'You dont have krunkit installed, run "brew update && brew install krunkit" to install latest version',
```
dgolovin:
```suggestion
        'You dont have krunkit installed. Run "brew update && brew install krunkit" to install latest version',
```

For consistency with message below.

dgolovin (changes requested):
Works for me, test should be added to cover new code.

odockal:
How about creating a machine from Dashboard on intel-based mac machines? This way it is gonna fail with error for the user since libkrun does not work there, no?

benoitf:
how does it work on mac/Intel ? as libkrun is not supported there

cbr7:
I have some concerns about this PR, specifically it's not obvious to me that the changes in the default value from `applehv` to `libkrun` are being propagated to podman CLI in order to maintain alignment between PD and Podman. I looked at the code in the PR and it's not clear to me where this propagation takes place and I would like to be sure that it does take place becayse otherwise changing the default value will cause podman machines to become invisible in Podman CLI since by default `applehv` will still be the default value in CLI.

gastoner:
@dgolovin Tests added

gastoner:
@benoitf @cbr7 I've added if statement that if the platform is `amd64` we will use `applev` no matter what. 
I'm just not sure what we should do with the CLI visibility, export some variable to show those machines instead? This does not seem right to me (what if is the user creating libkrun machine after he has created 4 applev machines)

cbr7:
@gastoner I agree that that is a concern also, however that is also the current behavior, only in reverse, with `applehv` as the default. In https://github.com/podman-desktop/podman-desktop/issues/11453 @benoitf describes that `containers.conf` file should be modified or created if it doesn't already exist with provider setting to `libkrun`, this would at least keep alignment with what PD is doing, even though yes in that case any existing `applehv` machines would become somewhat invisible in turn.

gastoner:
@benoitf @cbr7 should be updated, just I cant verify on intel mac :(

rujutashinde:
@cbr7 @odockal will you be able to verify this fix on  `intel mac` so we can merge this?

odockal:
> @cbr7 @odockal will you be able to verify this fix on `intel mac` so we can merge this?

I am looking at intel mac.

cbr7:
@gastoner @odockal @rujutashinde I've checked out this PR and the changes are mostly ok, however I think there is one usecase that has been overlooked. As far as I can see if the user creates a podman machine using the `Initialize and start` button from the dashboard, the podman machine created will always be applehv, regardless of what setting is written to the `containers.conf` file and additionally no changes are propagated to the file  either.

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -2165,7 +2178,13 @@ export async function createMachine(
 
   let provider: string | undefined;
   if (params['podman.factory.machine.provider'] && typeof params['podman.factory.machine.provider'] === 'string') {
-    provider = getProviderByLabel(params['podman.factory.machine.provider']);
+    if (os.arch() === 'amd64') {
+      // Intel machine
+      provider = VMTYPE.APPLEHV;
```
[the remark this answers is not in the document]
odockal: This seems like a valid point. Documentation mentions `x64` as on of the values. Not even `amd64`.

odockal (changes requested):
Just one comment I can think of. Tested on Intel Mac. Just apple hv options are available.

gastoner:
Now we should create libkrun machine on arm64 and applehv on x86 machines

gastoner:
@cbr7 @odockal Updated

cbr7:
@gastoner LGTM

cbr7 (approved):

gastoner:
@odockal @dgolovin can you PTAL?

extensions/podman/packages/extension/src/extension.spec.ts
```diff
@@ -750,6 +752,95 @@ describe.each([
...
+    stdout: `podman version 5.4.0`,
+  } as extensionApi.RunResult);
+
+  await extension.createMachine(
+    {
+      'podman.factory.machine.cpus': '2',
+      'podman.factory.machine.image': 'path',
+      'podman.factory.machine.memory': '1048000000', // 1048MB = 999.45MiB
+      'podman.factory.machine.diskSize': '250000000000', // 250GB = 232.83GiB
+      'podman.factory.machine.provider': expectedProvider,
```
odockal, line 772: this is the only line that differs within these two parametrized test cases. Can we make it a function and just change the argument?

extensions/podman/packages/extension/src/extension.spec.ts
```diff
@@ -750,6 +752,95 @@ describe.each([
   });
 });
 
+test.each([
+  { architecture: 'arm64', expectedProvider: VMTYPE.LIBKRUN },
```
odockal, line 756: Can we expand on the parameters to also contain `arm64 + applehv` ?

extensions/podman/packages/extension/src/utils/podman-configuration.spec.ts
```diff
@@ -169,6 +170,52 @@ describe('isRosettaEnabled', () => {
...
+  vi.spyOn(fs.promises, 'readFile').mockResolvedValue('');
+  vi.spyOn(podmanConfiguration, 'readContainersConfigFile').mockResolvedValue('[machine]\nprovider = "applehv"');
+
+  vi.spyOn(fs, 'existsSync').mockReturnValue(true);
+
+  await podmanConfiguration.updateMachineProviderSettings(VMTYPE.LIBKRUN);
+
+  expect(fs.promises.writeFile).toHaveBeenCalledWith(
+    podmanConfiguration.getContainersFileLocation(),
+    // should not contain provider
```
odockal: typo? should contain.

odockal (approved):
LGTM. 
nice to have, could be follow up: Just small changes mentioned in comments.

gastoner:
@odockal I've applied some suggestions, but I wasn't able to connect those 2 tests from because there wold be required some ifs in order to check that from dashboard we can create only applehv on x64 and libkrun on arm64 since there is no option to chose

gastoner:
@dgolovin Can you PTAL? Need aproval from you in order to merge this

benoitf:
wondering if it should not be merged post 1.20.0 or delaying 1.20.0

I don't think we have enough usage data to merge it the same day of a release

extensions/podman/packages/extension/src/checks/macos-checks.ts
cfergeau: Is krunkit in the main brew repos? Upstream doc uses an external tap https://github.com/containers/krunkit/?tab=readme-ov-file#installation
```
$ brew tap slp/krunkit
$ brew install krunkit
```

gastoner: I've added `brew tap slp/krunkit`

dgolovin (approved):

gastoner:
Wait until 1.20 is relased 

DO NOT merge now

gastoner:
Merging since 1.20 is out

PR #12789, opened 2025-06-10
chore(e2e): change expected img num in stress test depending on OS

danivilla9:
### What does this PR do?
This PR fixes the problem with the expected number of images in the ui-stress-test spec file by changing the expected amount depending on the operating system it's been ran

https://github.com/podman-desktop/e2e/issues/362

### How to test this PR?
Run the windows stress test workflow: https://github.com/podman-desktop/e2e/actions/workflows/podman-desktop-e2e-stress-ui-windows.yaml

### DIFF
--- a/tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
+++ b/tests/playwright/src/special-specs/ui-stress/ui-stress.spec.ts
@@ -16,6 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 import { expect as playExpect, test } from '../../utility/fixtures';
+import { isLinux } from '../../utility/platform';
 import { waitForPodmanMachineStartup } from '../../utility/wait';
 
 const numberOfObjects = Number(process.env.OBJECT_NUM) || 100;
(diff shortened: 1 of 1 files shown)


danivilla9:
Finally found out why the [Podman Desktop E2E UI Stress Test (Windows) ](https://github.com/podman-desktop/e2e/actions/workflows/podman-desktop-e2e-stress-ui-windows.yaml) has been failing recently. The image test case expects that when creating one or more pods an additional image (`localhost/podman-pause`) gets pulled as well, factoring its existence when counting the number of images in the UI but, turns out that on windows that it's not the case anymore. I've tried on my Fedora too and it doesn't behave like that either, it's only on the ubuntu runner from the workflow. This behavior depends on the `containers.conf` file and setup, as some OS distributions have built-in pod pause mechanisms that remove the necessity of pulling the `localhost/podman-pause` image (unlike the ubuntu-24.04 runner it seems).

So, this would be a solution, but I am not happy with it so I would like to get some feedback. There are other options that I think should fix the issue, but I am unsure about how to implement them or if they are good:
- Remove the row count, just expect the named elements (would work, but does not seem ideal)
- Find out a set of runners for all OS that behave all the same (either all pull `localhost/podman-pause` or none of them do) (not great, the default setup has changed once already, and it could happen in the future too)
- Setup the machines to behave all the same (really unsure about how to do it)

danivilla9:
As discussed in the meeting, the approach of this solution has been deemed adequate.

benoitf (approved):

PR #12805, opened 2025-06-11
fix: refresh providers when vm connections are registered/unregistered

### LINKED ISSUE #122: Second VM not appearing if first is started
See the demo below, when a first VM exists and is running, and the user creates a second VM, this second VM does not appear, before the first VM is stopped.

Later in the demo, if the first VM is stopped, and a second VM is created, this second VM appears immediately.

https://github.com/user-attachments/assets/af86d5bc-0c4c-4131-91bf-4d5be7f4c5a7

feloy:

### What does this PR do?

Refresh VM providers when vm connections are registered/unregistered

Fixes https://github.com/redhat-developer/podman-desktop-rhel-ext/issues/122

### How to test this PR?

See https://github.com/redhat-developer/podman-desktop-rhel-ext/issues/122 for steps to reproduce

### DIFF
--- a/packages/renderer/src/stores/providers.spec.ts
+++ b/packages/renderer/src/stores/providers.spec.ts
@@ -90,9 +90,7 @@ test.each([
   await callback();
 
   // wait listContainersMock is called
-  while (getProviderInfosMock.mock.calls.length === 0) {
-    await new Promise(resolve => setTimeout(resolve, 10));
-  }
+  await vi.waitFor(() => expect(getProviderInfosMock).toHaveBeenCalled());
 
   // now get list
   const providerListResult = get(providerInfos);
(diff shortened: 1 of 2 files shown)


benoitf (approved):
LGTM

could you please add a test where we see that it's being refreshed if register or unregister events are sent

packages/renderer/src/stores/providers.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    } as unknown as ProviderInfo,
+  ]);
+
+  // send event
+  const callback = callbacks.get(eventName);
+  expect(callback).toBeDefined();
+  await callback();
+
+  // wait listContainersMock is called
+  while (getProviderInfosMock.mock.calls.length === 0) {
```
SoniaSandler: could we use waitFor here?

feloy: yes, you are right, I changed. Thanks

SoniaSandler (approved):
LGTM

PR #12809, opened 2025-06-11
feat: support routing to experimental features page from podman-desktop:// protocol

### LINKED ISSUE #11851: Add links to experimental features in the release notes on the dashboard

When we announce new features in the release notes we also want to provide a link to users so that they can enable  these features. See mockup:
Image

Clicking on the `Enable in Settings >` link takes you to the experimental features page.

SoniaSandler:
### What does this PR do?
This PR adds routing to the experimental features page in Podman Desktop when using the podman-desktop:// protocol in order to be able to open the experimental page from the release notes both in the website and the app

Closes https://github.com/podman-desktop/podman-desktop/issues/11851

### How to test this PR?
1.  Build Podman Desktop locally and make sure that the app build is working
2. In the browser, enter podman-desktop://preferences/experimental
5. The app (when already open) should navigate to to experimental features page

### DIFF
--- a/packages/renderer/package.json
+++ b/packages/renderer/package.json
@@ -23,7 +23,6 @@
     "@testing-library/svelte": "^5.2.8",
     "@testing-library/user-event": "^14.6.1",
     "@tsconfig/svelte": "^5.0.4",
-    "@types/he": "^1.2.3",
     "@types/js-yaml": "^4.0.9",
     "@types/validator": "^13.15.2",
     "@typescript-eslint/eslint-plugin": "^8.39.1",
--- a/packages/renderer/package.json
+++ b/packages/renderer/package.json
@@ -23,6 +23,7 @@
     "@testing-library/svelte": "^5.2.8",
     "@testing-library/user-event": "^14.6.1",
     "@tsconfig/svelte": "^5.0.4",
+    "@types/he": "^1.2.3",
     "@types/js-yaml": "^4.0.9",
     "@types/validator": "^13.15.2",
     "@typescript-eslint/eslint-plugin": "^8.39.1",
--- a/packages/renderer/src/lib/markdown/Markdown.svelte
+++ b/packages/renderer/src/lib/markdown/Markdown.svelte
@@ -100,24 +100,13 @@ onMount(() => {
     htmlExtensions: [directiveHtml({ button, image, link, warnings })],
   });
 
-  console.log(
-    decode(
-      micromark(text, {
-        extensions: [directive()],
-        htmlExtensions: [directiveHtml({ button })],
-      }),
-    ),
-  );
-
   // remove href values in each anchor using # for links
   // and set the attribute data-pd-jump-in-page
   const parser = new DOMParser();
   const doc = parser.parseFromString(decode(html), 'text/html');
   const links = doc.querySelectorAll('a');
   links.forEach(link => {
-    console.log(link);
     const currentHref = link.getAttribute('href');
-    console.log(currentHref);
     // remove and replace href attribute if matching
     if (currentHref?.startsWith('#')) {
       // get current value of href
--- a/packages/renderer/src/lib/markdown/Markdown.svelte
+++ b/packages/renderer/src/lib/markdown/Markdown.svelte
@@ -82,12 +82,6 @@ $: markdown
     }))
   : undefined;
 
-function decode(htmlString: string): string {
-  let textArea = document.createElement('textarea');
-  textArea.innerHTML = htmlString;
-  return textArea.value;
-}
-
 onMount(() => {
   if (markdown) {
     text = markdown;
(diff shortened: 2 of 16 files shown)


packages/renderer/package.json
```diff
@@ -23,12 +23,14 @@
...
     "@testing-library/user-event": "^14.6.1",
     "@tsconfig/svelte": "^5.0.4",
+    "@types/he": "^1.2.3",
     "@types/js-yaml": "^4.0.9",
     "@types/validator": "^13.15.1",
     "@typescript-eslint/eslint-plugin": "^8.34.0",
     "@xterm/addon-serialize": "^0.13.0",
     "eslint-plugin-svelte": "^3.9.2",
     "filesize": "^10.1.6",
+    "he": "^1.2.0",
```
benoitf: it's unmaintained since 7 years, we can't use that library

benoitf: can't we use `decodeURIComponent` ?

packages/renderer/src/lib/markdown/Markdown.svelte
```diff
@@ -45,6 +45,7 @@ UI guidelines -->
 </style>
 
 <script lang="ts">
+import { decode } from 'he';
```
benoitf: decodeURIComponent is not working ?


also I'm not sure why we would need to change Markdown component ?

SoniaSandler: I wasn't aware of `decodeURIComponent` so I'll check if it works.
The issue that I came across was since we need the link to work both in the app and on the website, using `[]()` for links didn't work for the podman-desktop:// protocol in the website because it was treated as a relative link to the website, so I used `<a></a>`. However, the `<a></a>` tags were encoded in the markdown before processing the link which caused them not to show up as a link in the release notes banner.

SoniaSandler: It looks like decodeURIComponent doesn't work with HTML entities, but I'll check other maintained libraries

benoitf: does it work with something like
```
function decode(str) {
  const parser = new DOMParser();
  const doc = parser.parseFromString(str, 'text/html');
  return doc.documentElement.textContent;
}
```

SoniaSandler: It works partially since it does decode things like `&lt;`, but as we return the text content, the other tags disappear.
However, this function I found works:
```
function decode(str: string) {
  let txt = document.createElement("textarea");
  txt.innerHTML = str;
  return txt.value;
}
```

axel7083 (commented):
I am surprised by the title 'chore' ahah, this seems more like a feature

packages/renderer/src/navigation.ts
```diff
@@ -130,5 +130,8 @@ export const handleNavigation = (request: InferredNavigationRequest<NavigationPa
     case NavigationPage.EDIT_CONTAINER_CONNECTION:
       router.goto(`/preferences/container-connection/edit/${request.parameters.provider}/${request.parameters.name}`);
       break;
+    case NavigationPage.EXPERIMENTAL_FEATURES:
+      router.goto('/preferences/experimental');
+      break;
```
axel7083, line 135: please add a test in `navigation.spec.ts`

SoniaSandler:
@axel7083 I updated the title :) and added more tests

axel7083 (commented):
I know this PR is kinda old, and sorry for getting back this late (PTO, and forgot about it)

Anyway, I am not sure to understand the markdown thing, so I could approve the part about the navigation experimental alone, if you want to split the PR

SoniaSandler:
@axel7083 yes, I think I'll split this PR into navigation - markdown. I'll ping you once it's done

axel7083 (dismissed):
On windows 11 it is not working -

I followed the steps, here is a breakdown

1. run `pnpm compile:current`
2. execute `dist/podman-desktop-1.21.0-next-setup-x64.exe`
3. open in browser `podman-desktop://preferences/experimental` : nothing is happing, podman desktop is starting, but nothing
4. tried again with podman desktop opened: nothing is happening
5. tried with `podman-desktop:extension/headlamp.Headlamp` and nothing too, so probably broken. 

---

To check the issue, I downloaded podman destkop `v1.20.2` from the website. I can confirm that I am able to open `podman-desktop:extension/headlamp.Headlamp` and it redirects to the extension page, so this PR seems to introduce a regression

SoniaSandler:
@axel7083 in the logs of the podman desktop version that is not working, do you have a log line like `url ${url} does not start with podman-desktop:extension/ or podman-desktop:experimental, skipping.`?

SoniaSandler:
It worked for me on a Windows machine

https://github.com/user-attachments/assets/9ed40169-e6fe-4a5b-adb9-9f2a6d62e154

jeffmaury:
Works for me only if the application is already started but not if the app is not started

SoniaSandler:
@jeffmaury is it the same behavior like previous versions?
Note: on mac the navigation also works only if the application has already started

jeffmaury:
> @jeffmaury is it the same behavior like previous versions? Note: on mac the navigation also works only if the application has already started

Same behavior with 1.20.2

MarsKubeX (approved):
LGTM Tested in MacOS following the steps for it and working fine.

bmahabirbu:
Tested on Windows 11 as well, works without a hitch

bmahabirbu (approved):

PR #12810, opened 2025-06-11
fix: avoid to send status field when patching a resource

### LINKED ISSUE #12751: Cannot edit non-immutable fields in Kubernetes resources
### Bug description

After merging [commit 11d9330](https://github.com/podman-desktop/podman-desktop/commit/11d93306274ba9d76081f3d78c0e6baf7c4b8b7a), we encountered an error when attempting to edit non-immutable fields in Kubernetes resources. Fields that should be editable — such as the replicas field in a Deployment resource — now result in an error when edited. This behavior is unexpected, as these fields are not immutable by design.

Image

### Operating system

Fedora, macOs, Windows

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

- Create a Kind Kubernetes cluster.
- Create a Deployment resource in the cluster.
- Go to resource details.
- Go to the Kube tab.
- Attempt to edit the replicas field of the deployment.

### Relevant log output

```shell

```

### Additional context

_No response_

benoitf:
### What does this PR do?
status can contain fields like conditions having date, and then the kube library has issues with parsing the date (we used to remove the fields with date)

here, remove the status field before sending to the patch command

fixes https://github.com/podman-desktop/podman-desktop/issues/12751

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
+++ b/packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
@@ -2348,6 +2348,66 @@ test('test sync resources was called with no resourceVersion, uid, selfLink, or 
   );
 });
 
+test('test sync resources was called with no status being passed through', async () => {
+  const client = createTestClient('default');
+  const context = 'test-context';
+  const namespace = 'default';
+  const manifests = [
+    {
+      apiVersion: 'v1',
+      kind: 'Pod',
+      metadata: {
+        name: 'test-pod',
+        resourceVersion: '123',
+        uid: 'uid123',
+        selfLink: '/api/v1/namespaces/default/pods/test-pod',
+        creationTimestamp: new Date(42),
+      },
+      status: {
+        phase: 'Running',
+        conditions: [
+          {
+            type: 'Ready',
+            status: 'True',
+          },
+        ],
+      },
+    },
+  ];
+
+  const mockedPatch = vi.fn();
+
+  makeApiClientMock.mockReturnValue({
+    read: vi.fn(),
+    create: vi.fn(),
+    patch: mockedPatch,
+  });
+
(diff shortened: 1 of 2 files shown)


axel7083 (approved):
LGTM Codewise

packages/main/src/plugin/kubernetes/kubernetes-client.spec.ts
```diff
@@ -2348,6 +2348,66 @@ test('test sync resources was called with no resourceVersion, uid, selfLink, or
...
+    read: vi.fn(),
+    create: vi.fn(),
+    patch: mockedPatch,
+  });
+
+  // Call the syncResources method with 'create' action
+  await client.syncResources(context, manifests, 'apply', namespace);
+
+  // Expect patch method to have been called
+  expect(mockedPatch).toHaveBeenCalled();
```
dgolovin: nit: toHaveBeenCalledWith() check only should be enough IMO.

```suggestion
```

dgolovin (approved):
LGTM

feloy:
I have a vague remembering that the problem may be caused by the dates being in a wrong format. Do you think it could be a solution?

deboer-tim (approved):
Tested fine, code LGTM.

> I have a vague remembering that the problem may be caused by the dates being in a wrong format. Do you think it could be a solution?

It might work, but it's not worth it - this is simpler and there's no reason for us to include the status (or any live state), as it will be ignored by the cluster.

benoitf:
> I have a vague remembering that the problem may be caused by the dates being in a wrong format. Do you think it could be a solution?

This is what I was saying in the body of the PR

But as we're excluding already all fields having a date, then I continue the same approach. Also status field should not be changed most of the time so it was solving two things in a row

feloy (approved):

PR #12818, opened 2025-06-11
chore: correctly type guard e.target for micromark listener

### LINKED ISSUE #12211: Compose's Onboarding Setup version selector does not do anything
### Bug description

When trying to select a version for Compose on the Onboarding setup, if you click on the link nothing happens.

https://github.com/user-attachments/assets/01a457c4-d233-4584-8dc4-c9adf380a16b

### Operating system

Windows 10

### Installation Method

### Version

1.18.0

### Steps to reproduce

- Enable onboarding workflow by removing C:\Users\[youruser]\.local\share\containers folder
- Open Podman Desktop, go to Compose's step in the onboarding workflow
- Try to select a version in the "Want to download a different version?" link 

### Relevant log output

```shell

```

### Additional context

Inspecting the link shows that there is no address to point to. The issue has been reproduced in Mac and Linux too.

cdrage:
chore: correctly type guard e.target for micromark listener

### What does this PR do?

This PR refactors the `micromark-listener-handler` file that handles the
"opening a command" implementation for onboarding.

When this was refactored as the result of https://github.com/podman-desktop/podman-desktop/pull/11038
`eventTarget` was using type-narrowing manually, which does not allow
accessibility to properties like dataset, disabled, etc which is needed
for the "command" functionality to work.

This PR fixes that by safely checking / type guarding e.target.

### Screenshot / video of UI

Should now work:

https://github.com/user-attachments/assets/3e8a7836-2cf0-4663-866e-f98ee3fa0d12

Closes https://github.com/podman-desktop/podman-desktop/issues/12211

### How to test this PR?

1. Start onboarding
2. Go to Compose or Kubectl
3. Click the "Want to download a different version?" button and see that
   it works correctly now with no errors.

### DIFF
--- a/packages/renderer/src/lib/markdown/micromark-listener-handler.ts
+++ b/packages/renderer/src/lib/markdown/micromark-listener-handler.ts
@@ -19,6 +19,8 @@
 import { executeButtonCommand } from './component/micromark-button';
 import { executeExpandableToggle } from './component/micromark-expandable-section';
 
+// createListener "listens" to button clicks and executes the appropriate command or action based on the button's dataset attributes.
+// for example.. "command" will execute the extensions associated command correctly.
 export function createListener(
   inProgressMarkdownCommandExecutionCallback: (
     command: string,
@@ -27,39 +29,21 @@ export function createListener(
   ) => void,
 ): EventListener {
   return (e: Event): void => {
-    let eventTarget: EventTarget;
-
-    if (e.target && typeof e.target === 'object') {
-      eventTarget = e.target;
-    } else {
+    // If the target is NOT an HTMLElement we should immediately return, this is because we listen to ALL events, but we
+    // are only interested in events that are triggered by HTML elements.
+    if (!(e.target instanceof HTMLElement)) {
       return;
     }
 
-    // Retrieve the command and expandable within the dataset
-    let command: string | undefined;
-    let args: string | undefined;
-    let expandable: string | undefined;
+    // Retrieve the command and "expandable" within the dataset
+    const command = e.target.dataset?.command;
+    const expandable = e.target.dataset?.expandable;
 
-    if ('dataset' in eventTarget && eventTarget.dataset && typeof eventTarget.dataset === 'object') {
-      const targetDataset = eventTarget.dataset;
-      if ('command' in targetDataset && typeof targetDataset.command === 'string') {
-        command = targetDataset.command;
-      }
-      // The targetDataset.args is a JS object but represented as a string
-      // typeof targetDataset.args === 'object' => false
-      // console.log(targetDataset.args) => [object Object]
-      if ('args' in targetDataset && targetDataset.args && typeof targetDataset.args === 'string') {
-        args = targetDataset.args;
-      }
-      if ('expandable' in targetDataset && typeof targetDataset.expandable === 'string') {
-        expandable = targetDataset.expandable;
-      }
-    }
-
+    // BUTTON CHECK: 'data-pd-jump-in-page'
     // if the user click on a a href link containing data-pd-jump-in-page attribute
-    if (eventTarget instanceof HTMLAnchorElement) {
+    if (e.target instanceof HTMLAnchorElement) {
       // get a matching attribute ?
-      const hrefId = eventTarget.getAttribute('data-pd-jump-in-page');
+      const hrefId = e.target.getAttribute('data-pd-jump-in-page');
 
       // get a linked ID
       if (hrefId) {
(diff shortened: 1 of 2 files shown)


cdrage:
If someone could please test the "WSL setup" buttons that you click during your first onboarding sequence on Windows that'll be great! 

I tested the code functionality on Linux and macOS, but not Windows (having issues with Windows 11 that "upgraded" my system).

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -76,44 +67,47 @@ export function createListener(
...
         window
           .executeCommand(command, args)
           .then(value => inProgressMarkdownCommandExecutionCallback(command, 'successful', value))
           .catch((reason: unknown) => inProgressMarkdownCommandExecutionCallback(command, 'failed', reason))
           .finally(() => {
-            targetButton.disabled = false;
-            if (targetButton.firstChild && targetButton.firstChild instanceof HTMLElement) {
-              targetButton.firstChild.style.display = 'none';
+            (e.target as HTMLButtonElement).disabled = false;
+            if (e.target instanceof HTMLElement && e.target.firstChild instanceof HTMLElement) {
```
dgolovin, line 107: HTMButtonElement inherited from HTMLElement. Considering it is already checked above (line 90) it is instance of HTMLButtonElement this instanceof is not required.

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -76,44 +67,47 @@ export function createListener(
...
+      if (e.target instanceof HTMLButtonElement && !e.target.disabled) {
         // If the command exists and the button is not disabled, we execute the command
         // we'll also be updating the inProgressMarkdownCommandExecutionCallback so we have
         // real-time updates on the button
         inProgressMarkdownCommandExecutionCallback(command, 'starting');
-        targetButton.disabled = true;
-        if (targetButton.firstChild && targetButton.firstChild instanceof HTMLElement) {
-          targetButton.firstChild.style.display = 'inline-block';
+        e.target.disabled = true;
+        if (e.target instanceof HTMLElement && e.target.firstChild instanceof HTMLElement) {
```
dgolovin, line 97: HTMLButtonElement extends HTMLElement. If e.target is an instance of HTMLButtonElement it is also an instance of HTMLElement. So thins instanceof check not needed IMO.

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -27,39 +29,28 @@ export function createListener(
...
-        args = targetDataset.args;
-      }
-      if ('expandable' in targetDataset && typeof targetDataset.expandable === 'string') {
-        expandable = targetDataset.expandable;
-      }
+    // Args are passed in as a string in the dataset, so they cannot be json / must be string.
+    let args: string | undefined;
+    const dataset = e.target.dataset;
+    if ('args' in dataset && dataset.args && typeof dataset.args === 'string') {
+      args = dataset.args;
```
dgolovin: `args` is used only inside `if (command)` block, so it should be moved there.

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
```diff
@@ -27,39 +29,28 @@ export function createListener(
...
 ): EventListener {
   return (e: Event): void => {
-    let eventTarget: EventTarget;
-
-    if (e.target && typeof e.target === 'object') {
-      eventTarget = e.target;
-    } else {
+    // If the target is NOT an HTMLElement we should immediately return, this is because we listen to ALL events, but we
+    // are only interested in events that are triggered by HTML elements.
+    if (!(e.target instanceof HTMLElement)) {
```
dgolovin, line 34: nit: Having temporary var for e.target seems to be a good idea to avoid reference through event instance `e.target`.

dgolovin (changes requested):

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
cdrage: I did have that before, but then the code got a lot more complicated :( so keeping it like this for now.

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
cdrage: Still need this in place, or I get a typescript linting error.

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
cdrage: Linting error when removed, I get `Property 'style' does not exist on type 'ChildNode'`, so would need to keep this in.

packages/renderer/src/lib/markdown/micromark-listener-handler.ts
cdrage: good catch! I've moved it to before the if block

dgolovin (approved):
The code LGTM, but I have not tested it.

jeffmaury (approved):
LGTM

cdrage:
@jeffmaury Can you confirm WSL2 onboarding setup isn't effected? That's the only one I cannot test and I know it uses the micromark listener :( 

Tried the other parts (compose and kubectl) and it's all fine. Just need confirmation on that part.

jeffmaury:
> @jeffmaury Can you confirm WSL2 onboarding setup isn't effected? That's the only one I cannot test and I know it uses the micromark listener :(
> 
> Tried the other parts (compose and kubectl) and it's all fine. Just need confirmation on that part.

Confirm the WSL onboarding is working well with this PR

cdrage:
> > @jeffmaury Can you confirm WSL2 onboarding setup isn't effected? That's the only one I cannot test and I know it uses the micromark listener :(
> > Tried the other parts (compose and kubectl) and it's all fine. Just need confirmation on that part.
> 
> Confirm the WSL onboarding is working well with this PR

Thank you! Merging now :)

PR #12839, opened 2025-06-13
chore(test): install ingress controller on Kind cluster only if needed

### LINKED ISSUE #12816: Install ingress controller on Kind clusters only when needed

Currently, our Kubernetes test suites use a setup where a Kind cluster is created with an Ingress controller installed by default. This can lead to problems: if the Ingress controller has an issue, it can block other test suites, even those that do not depend on the Ingress controller,  because the Kind cluster setup fails.

We should update our tests so that the Ingress controller is installed only when explicitly required. 

_No response_

### Additional context

_No response_

amisskii:
### What does this PR do?
Skips ingress controller installation on Kind clusters if it is not required by tests. This avoids potential issues with the ingress controller and improves k8s e2e test reliability by preventing failures during cluster setup.
https://github.com/podman-desktop/podman-desktop/issues/12816

### How to test this PR?

https://github.com/podman-desktop/e2e/actions/runs/15593572158

### DIFF
--- a/tests/playwright/src/model/pages/create-kind-cluster-page.ts
+++ b/tests/playwright/src/model/pages/create-kind-cluster-page.ts
@@ -53,34 +53,30 @@ export class CreateKindClusterPage extends CreateClusterBasePage {
     this.warning = this.page.getByRole('alert', { name: 'warning' });
   }
 
-  public async createClusterDefault(clusterName: string = 'kind-cluster', timeout?: number): Promise<void> {
-    return test.step('Create default cluster', async () => {
-      await fillTextbox(this.clusterNameField, clusterName);
+  private async validateKindClusterDefaultSettings(): Promise<void> {
+    return test.step('Validate kind cluster default settings', async () => {
+      await playExpect(this.configFileInput).toBeEmpty();
+      await playExpect(this.clusterNameField).toHaveValue('kind-cluster');
       await playExpect(this.providerType).toHaveText('podman');
       await playExpect(this.httpPort).toHaveValue('9090');
       await playExpect(this.httpsPort).toHaveValue('9443');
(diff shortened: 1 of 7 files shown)


danivilla9 (approved):
LGTM

tests/playwright/src/model/pages/create-kind-cluster-page.ts
```diff
@@ -91,22 +87,15 @@ export class CreateKindClusterPage extends CreateClusterBasePage {
...
-              exact: true,
-            });
-            await playExpect(providerTypeButton).toBeEnabled();
-            await providerTypeButton.click();
-          }
-          await playExpect(this.providerType).toHaveText(providerType);
-        } catch {
-          throw new Error(`Provider type '${providerType}' doesn't exist`);
-        }
+      if (providerType && providerType !== 'podman') {
```
odockal, line 90: We should stick to previous construct. 
If the providerType argument is provided, then we need to verify the drop down has its value. No matter if it is podman or not.

odockal (changes requested):

tests/playwright/src/model/pages/create-kind-cluster-page.ts
amisskii: The verification of whether the drop down has a default value, which is podman, is now handled in the `validateClusterDefaultSettings() ` function ^, so we don't need to check it again here.

amisskii: If we try to use a provider type that doesn't exist, Playwright will throw an error on line 96 because it won't find a locator for that provider type. So a try catch block there isn't necessary.

odockal (approved):
LGTM.

cbr7 (approved):

PR #69, opened 2022-03-30
feat: Refactor provider and connections

benoitf:
Now we can register one provider with optional lifecycle.
And on the provider, we can add container or kubernetes connections

Also try to decouple some callbacks by using listeners/events

Fix https://github.com/containers/desktop/issues/62
Fix https://github.com/containers/desktop/issues/36

Change-Id: Iaf828daa70778e7743974742fdc5d46a9d17d6e3
Signed-off-by: Florent Benoit <fbenoit@redhat.com>

PR #12841, opened 2025-06-13
chore: addProviderMenuItem - add provider if it does not exist

### LINKED ISSUE #4624: registerProviderMenuItem displays a `temp` menu entry
### Bug description

I'm creating an extension, which adds a provider menu item in the tray, with the following code.

A `temp` entry is created in the tray, containing the provider's menu entry.

tray-provider

```
  // Create a provider with an example name, ID and icon
  const provider = extensionApi.provider.createProvider({

    name: 'Image Checker',
    id: 'image-checker',
    status: 'unknown',
    images: {
      icon: './icon.png',
      logo: './icon.png',
    },
  });
  extensionContext.subscriptions.push(provider);
  
  const providerMenuItem = extensionApi.tray.registerProviderMenuItem('image-checker', {
    id: "image-checker-menu1",
    type: 'normal',
    label: "Image checker 2",    
  });
  extensionContext.subscriptions.push(providerMenuItem);

  const myCommand = extensionApi.commands.registerCommand('image-checker-menu1', async (newContext: string) => {
    console.log('tray menu item clicked');
  });
  extensionContext.subscriptions.push(myCommand);
}
```

### Operating system

Linux Fedora

### Installation Method

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

### LINKED ISSUE #6915: Openshift local system tray commands not responding
### Bug description

WE have an issue reported on slack:

> having a wired issue with podman desktop openshift local feature
 podman desktop doesn't respond to any of the below commands
I need to manually execute the CRC commands.
image.png

image (4)

### Operating system

Mac OS

### Installation Method

Other

### Version

1.9.0

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

cdrage:
### What does this PR do?

**Context information:**

addProviderMenuItem is an API call that is used to "group" tray menu
items together to one provider.

There is a problem however... The API call depends that we registered the
provider with registerLifecycle in the API... since the tray menu calls
"addProviderLifecycleListener".

However, "registerLifecycle" was improved and had no longer been
required in 2022 for extensions:
https://github.com/podman-desktop/podman-desktop/pull/69

registerLifecycle is also NOT used at all for external extensions
(public GitHub search reveals... nothing.)

We also cannot "pre-emptively" give provider information since tray is
initialized very early:
https://github.com/cdrage/podman-desktop/blob/6f591c25b97c7ad6d8794699bf1f15cd5ff92325/packages/main/src/index.ts#L141-L142

As our API is designed to always be backwards compatible, in the case
that we use "addProviderMenuItem" call, we will add the provider to the
list (menuProviderItems).

**This PR does the following:**
* Removes "temp" sections as they will no longer be needed. Provider
  information should always be passed in, if there is no provider
  information (worst case scenario), we console.error out.
* Keep backwards compatibility for any non-public extensions still using
  registerLifecycle and addProviderMenuItem API calls.
* Passed in provider information when doing addProviderMenuItem (makes
  sense right?) in the scenario that the extension calls
  addProviderMenuItem WITHOUT doing `registerLifecycle` first in their
  extension
* Added lots of tests to cover all scenarios.

### Screenshot / video of UI

https://github.com/user-attachments/assets/e77da468-50d3-47f6-b117-118e7f6ff821

Closes https://github.com/podman-desktop/podman-desktop/issues/6915
Closes https://github.com/podman-desktop/podman-desktop/issues/4624

### How to test this PR?

1. Download the openshift local extension
2. Install
3. Right click on the podman desktop icon, see that OpenShift Local is
   there now (no longer "temp").

### DIFF
--- a/packages/main/src/tray-menu.spec.ts
+++ b/packages/main/src/tray-menu.spec.ts
@@ -163,3 +163,98 @@ test('Tray click trigger is only added on Windows devices', () => {
   trayMenu = new TrayMenu(tray, animatedTray);
   expect(onSpy).not.toHaveBeenCalledWith('click', expect.any(Function));
 });
+
+test('with a provider already found, should push the addProviderMenuItem fine', async () => {
+  const onSpy = vi.spyOn(ipcMain, 'on');
+  const menuBuild = vi.spyOn(Menu, 'buildFromTemplate');
+
+  trayMenu = new TrayMenu(tray, animatedTray);
+
+  // Add a "provider" so that the menu already finds the provider
+  trayMenu.addProviderItems({
+    id: 'testId',
+    name: 'TestProviderAlreadyHere',
+    internalId: 'internalId',
+  } as ProviderInfo);
+
+  // Do a mock call which should add the menu item (testId matches the provider added above)
+  onSpy.mock.calls[0]?.[1](undefined as unknown as Electron.IpcMainEvent, {
+    providerId: 'testId',
+    menuItem: { id: 'itemId', label: 'Foobar' },
+  });
+
+  // Expect that tray menu was called with TestProviderAlreadyHere
+  // as well as the actual add-provider-menu-item event happened.
+  expect(onSpy).toHaveBeenCalledWith('tray:add-provider-menu-item', expect.any(Function));
+  expect(menuBuild).toHaveBeenCalled();
+  expect(menuBuild.mock.lastCall?.[0]?.[0]?.label).toBe('TestProviderAlreadyHere');
+
+  // Check that the menu item has been successfully added to the provider's menu
+  expect(
+    (menuBuild.mock.lastCall?.[0]?.[0]?.submenu as Array<MenuItemConstructorOptions>)?.filter(
+      it => it.label === 'Foobar',
+    ),
+  ).toHaveLength(1);
+
+  // Expect that we have "TestProv" somewhere in the menu (not submenu)
+  const providerMenu = (menuBuild.mock.lastCall?.[0] as Array<MenuItemConstructorOptions>).find(
+    it => it.label === 'TestProviderAlreadyHere',
+  );
+  expect(providerMenu).toBeDefined();
+});
+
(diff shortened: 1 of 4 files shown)


gastoner (approved):
LGTM on Linux

axel7083 (approved):
I wanted to try the testing step, and ended-up opening https://github.com/crc-org/crc-extension/issues/721 (oupsi).

Anyway, tried the PR, I can confirm I do not see any regression for managing Podman Machines (at least)

packages/main/src/tray-menu.spec.ts
```diff
@@ -163,3 +163,92 @@ test('Tray click trigger is only added on Windows devices', () => {
...
+
+  // Add a "provider" so that the menu already finds the provider
+  trayMenu.addProviderItems({
+    id: 'testId',
+    name: 'TestProviderAlreadyHere',
+    internalId: 'internalId',
+  } as ProviderInfo);
+
+  // Do a mock call which should add the menu item (testId matches the provider added above)
+  onSpy.mock.calls[0]?.[1](undefined as unknown as Electron.IpcMainEvent, {
```
axel7083, line 181: suggestion: I usually would rather have a check ensuring the mock has been called usually 

```suggestion
  expect(onSpy).toHaveBeenCalledOnce();
  const listener = onSpy.mock.calls[0]?.[1]
  assert(listener);
  
  listener(undefined as unknown as Electron.IpcMainEvent, {
```

axel7083: Otherwise, let's say the onSpy has never been called, you will not get an error at this line, making it harder to debug

cdrage:
@axel7083 added some extra checks that would help! you are right, it was missing some expect(onSpy) calls. I added some menuBuild checks as well.

cdrage:
Thanks for the testing and reviews @axel7083 and @gastoner !

PR #12871, opened 2025-06-17
chore: expose a method from extension loader: ensureExtensionIsEnabled

benoitf:
### What does this PR do?
expose a method from extension loader

relates to #8616 

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -2070,6 +2070,13 @@ export class PluginSystem {
     );
 
     this.ipcHandle(
+      'extension-loader:ensureExtensionIsEnabled',
+      async (_listener: Electron.IpcMainInvokeEvent, extensionId: string): Promise<void> => {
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/index.ts
```diff
@@ -2069,6 +2069,13 @@ export class PluginSystem {
       },
     );
 
+    this.ipcHandle(
+      'extension-loader:ensureExtensionIsEnabled',
+      async (_listener: Electron.IpcMainInvokeEvent, extensionId: string): Promise<void> => {
+        return this.extensionLoader.ensureExtensionIsEnabled(extensionId);
+      },
```
axel7083, line 2076: question: return void? it throw an error if the extension is not enabled?

benoitf: if extension is not enabled it does nothing.

This method clears the id from disabled extension (if disabled) else it's a no-op

axel7083 (approved):

MarsKubeX (approved):
LGTM

PR #12874, opened 2025-06-17
test(e2e): include podman provider check after machine creation

### LINKED ISSUE #11187: Include the check for provider and mode (rootless/rootful) when running the e2e tests for podman machine

Right now we do not have a verification of the provider type for the machine present in podman desktop. Same applies for the machine mode (rootful and rootless).

Include a check and add parameter that will be taking a specific virtualization provider into account when creating Podman Machine via Podman Desktop.

It requires to look for env. var. used in CI (CONTAINERS_MACHINE_PROVIDER) and use them when creating podman machine.

Also, we should look for possible default setup (env. vars.) we use in CI (ie. PODMAN_ROOTFUL, PODMAN_NETWORKING) and make sure to check for the podman machine mode (rootful, rootless, etc.). Probably using a call with `podman machine ls --format json` or similar.

Same applies when manipulating the podman machine in e2e tests, after the test, we should have the same machine as on beginning. 

_No response_

### Additional context

_No response_

danivilla9:
### What does this PR do?
This PR adds a check after creating a podman machine that verifies if the proper provider has been assigned during machine creation.
[#11187](https://github.com/podman-desktop/podman-desktop/issues/11187)

### DIFF
--- a/tests/playwright/src/model/core/types.ts
+++ b/tests/playwright/src/model/core/types.ts
@@ -84,3 +84,12 @@ export const KubernetesResourceAttributes: Record<KubernetesResources, string[]>
   ],
   [KubernetesResources.Jobs]: ['Selected', 'Status', 'Name', 'Conditions', 'Completions', 'Age', 'Actions'],
 };
+
+export enum PodmanVirtualizationProviders {
+  WSL = 'Wsl',
+  HyperV = 'Hyperv',
+  AppleHV = 'Apple HyperVisor',
+  LibKrun = 'default GPU enabled (LibKrun)',
+  Qemu = 'Qemu',
+  Native = '', //not a real provider, used for 'Connection Type' check in Resources page of Linux machines
+}
--- a/tests/playwright/src/model/pages/forms/machine-creation-form.ts
+++ b/tests/playwright/src/model/pages/forms/machine-creation-form.ts
@@ -20,7 +20,10 @@ import type { Locator, Page } from '@playwright/test';
 import test, { expect as playExpect } from '@playwright/test';
 
 import { isWindows } from '/@/utility/platform';
+import { getDefaultVirtualizationProvider } from '/@/utility/provider';
 
+import { DropdownComponent } from '../../components/dropdown-component';
+import type { PodmanVirtualizationProviders } from '../../core/types';
 import { BasePage } from '../base-page';
 
 export class MachineCreationForm extends BasePage {
@@ -33,6 +36,7 @@ export class MachineCreationForm extends BasePage {
   readonly podmanMachineDiskSize: Locator;
   readonly rootPriviledgesCheckbox: Locator;
   readonly userModeNetworkingCheckbox: Locator;
+  readonly providerTypeDropdown: DropdownComponent;
   readonly startNowCheckbox: Locator;
   readonly createMachineButton: Locator;
 
@@ -57,6 +61,8 @@ export class MachineCreationForm extends BasePage {
     this.userModeNetworkingCheckbox = this.podmanMachineConfiguration.getByRole('checkbox', {
       name: 'User mode networking',
     });
+    this.providerTypeDropdown = new DropdownComponent(page, 'Provider Type');
+
     this.startNowCheckbox = this.podmanMachineConfiguration.getByRole('checkbox', { name: 'Start the machine now' });
     this.createMachineButton = this.podmanMachineConfiguration.getByRole('button', { name: 'Create' });
   }
@@ -67,13 +73,15 @@ export class MachineCreationForm extends BasePage {
       isRootful = true,
       enableUserNet = false,
       startNow = true,
+      virtualizationProvider,
     }: {
       isRootful?: boolean;
       enableUserNet?: boolean;
       startNow?: boolean;
+      virtualizationProvider?: PodmanVirtualizationProviders;
     } = {},
   ): Promise<void> {
-    return test.step(`Create Podman Machine: ${machineName} with settings ${isRootful}, ${enableUserNet} and ${startNow}`, async () => {
+    return test.step(`Create Podman Machine '${machineName}' with settings: ${isRootful ? 'rootful' : 'rootless'}, ${enableUserNet ? 'usernet enabled' : 'usernet disabled'}, ${startNow ? 'startnow enabled' : 'startnow disabled'}${virtualizationProvider ? ', and ' + virtualizationProvider : ''}`, async () => {
       await playExpect(this.podmanMachineConfiguration).toBeVisible({
         timeout: 10_000,
       });
@@ -83,6 +91,9 @@ export class MachineCreationForm extends BasePage {
       await this.ensureCheckboxState(isRootful, this.rootPriviledgesCheckbox);
       if (isWindows) {
         await this.ensureCheckboxState(enableUserNet, this.userModeNetworkingCheckbox);
+      }
+      if (virtualizationProvider) {
+        await this.specifyVirtualizationProvider(virtualizationProvider);
       }
       await this.ensureCheckboxState(startNow, this.startNowCheckbox);
 
@@ -96,6 +107,32 @@ export class MachineCreationForm extends BasePage {
       if (desiredState !== (await checkbox.isChecked())) {
         await checkbox.locator('..').click();
         playExpect(await checkbox.isChecked()).toBe(desiredState);
+      }
+    });
+  }
+
+  /**
+   * Specifies the virtualization provider for a Podman machine during the creation process.
+   * This method selects the specified virtualization provider from the dropdown if it differs from the default.
+   * If no provider is specified or it matches the default, no action is taken.
+   *
+   * @param virtualizationProvider - The virtualization provider to select (e.g., PodmanVirtualizationProviders.WSL, PodmanVirtualizationProviders.HyperV, etc.), or undefined to use default
+   * @returns A Promise that resolves when the provider selection is complete
+   * @throws Will throw an error if the provider dropdown is not accessible or the specified provider is not available
+   */
+  async specifyVirtualizationProvider(
+    virtualizationProvider: PodmanVirtualizationProviders | undefined,
+  ): Promise<void> {
+    return test.step(`Set Podman Provider to ${virtualizationProvider ?? getDefaultVirtualizationProvider()}`, async () => {
+      if (!virtualizationProvider) return;
+      // Only select if the dropdown is actually present/visible
+      if (!(await this.providerTypeDropdown.getContainer().isVisible())) return;
+      await this.providerTypeDropdown.waitForReady();
+      // Compare by the hidden input value (case-insensitive)
+      const currentValue = (await this.providerTypeDropdown.getCurrentValue()).toLowerCase();
+      if (virtualizationProvider.toLowerCase() !== currentValue) {
+        await this.providerTypeDropdown.selectOption(virtualizationProvider, virtualizationProvider, false);
+        await this.providerTypeDropdown.verifyState(virtualizationProvider, virtualizationProvider);
       }
     });
   }
--- a/tests/playwright/src/specs/kubernetes-status-bar-providers.spec.ts
+++ b/tests/playwright/src/specs/kubernetes-status-bar-providers.spec.ts
@@ -17,6 +17,7 @@
  ***********************************************************************/
 
 import { ResourceElementState } from '../model/core/states';
+import { PodmanVirtualizationProviders } from '../model/core/types';
 import { CreateMachinePage } from '../model/pages/create-machine-page';
 import { ResourceConnectionCardPage } from '../model/pages/resource-connection-card-page';
 import { ResourcesPage } from '../model/pages/resources-page';
@@ -29,8 +30,10 @@ import {
   ensureCliInstalled,
   handleConfirmationDialog,
   setStatusBarProvidersFeature,
+  verifyVirtualizationProvider,
 } from '../utility/operations';
 import { isLinux } from '../utility/platform';
+import { getDefaultVirtualizationProvider, getVirtualizationProvider } from '../utility/provider';
 import { waitForPodmanMachineStartup, waitUntil } from '../utility/wait';
 
 const podmanProviderName = 'Podman';
@@ -95,6 +98,10 @@ test.describe.serial('Status bar providers feature verification', { tag: '@k8s_e
   });
   test('Create and delete new resource and verify providers updated', async ({ page, statusBar }) => {
     test.skip(isLinux, 'Not creating new machine on Linux');
+    test.skip(
+      getVirtualizationProvider() === PodmanVirtualizationProviders.HyperV,
+      'Podman Desktop is not able to have 2 HyperV machines running at the same time',
+    );
     test.setTimeout(350_000);
 
     //Create a new machine
@@ -103,7 +110,11 @@ test.describe.serial('Status bar providers feature verification', { tag: '@k8s_e
     const podmanResources = new ResourceConnectionCardPage(page, 'podman');
     await podmanResources.createButton.click();
     const createMachinePage = new CreateMachinePage(page);
-    await createMachinePage.createMachine(newMachineName, { startNow: true, setAsDefault: false });
+    await createMachinePage.createMachine(newMachineName, {
+      startNow: true,
+      setAsDefault: false,
+      virtualizationProvider: getVirtualizationProvider(),
+    });
 
     const machineCard = new ResourceConnectionCardPage(page, 'podman', newMachineName);
     playExpect(await machineCard.doesResourceElementExist()).toBeTruthy();
@@ -112,6 +123,8 @@ test.describe.serial('Status bar providers feature verification', { tag: '@k8s_e
         (await machineCard.resourceElementConnectionStatus.innerText()).includes(ResourceElementState.Running),
       { timeout: 30_000, sendError: true },
     );
+
+    await verifyVirtualizationProvider(machineCard, getVirtualizationProvider() ?? getDefaultVirtualizationProvider());
 
     playExpect(await statusBar.isProviderResourceRunning(podmanProviderName, newMachineName)).toBeTruthy();
 
--- a/tests/playwright/src/specs/z-podman-machine-onboarding.spec.ts
+++ b/tests/playwright/src/specs/z-podman-machine-onboarding.spec.ts
@@ -21,6 +21,7 @@ import * as os from 'node:os';
 import type { Locator, Page } from '@playwright/test';
 
 import { ResourceElementState } from '../model/core/states';
+import { PodmanVirtualizationProviders } from '../model/core/types';
 import type { DashboardPage } from '../model/pages/dashboard-page';
 import { PodmanMachineDetails } from '../model/pages/podman-machine-details-page';
 import { PodmanOnboardingPage } from '../model/pages/podman-onboarding-page';
@@ -30,6 +31,7 @@ import type { SettingsBar } from '../model/pages/settings-bar';
 import { expect as playExpect, test } from '../utility/fixtures';
 import { createPodmanMachineFromCLI, deletePodmanMachine, resetPodmanMachinesFromCLI } from '../utility/operations';
 import { isLinux } from '../utility/platform';
+import { getDefaultVirtualizationProvider, getVirtualizationProvider } from '../utility/provider';
 import { waitForPodmanMachineStartup } from '../utility/wait';
 
 const PODMAN_MACHINE_STARTUP_TIMEOUT: number = 360_000;
@@ -141,10 +143,14 @@ test.describe
           await playExpect(podmanOnboardingPage.machineCreationForm.rootPriviledgesCheckbox).toBeChecked();
           await playExpect(podmanOnboardingPage.machineCreationForm.startNowCheckbox).toBeChecked();
 
+          await podmanOnboardingPage.machineCreationForm.specifyVirtualizationProvider(getVirtualizationProvider());
+
           if (os.platform() === 'win32') {
-            await playExpect(podmanOnboardingPage.machineCreationForm.userModeNetworkingCheckbox).not.toBeChecked({
-              timeout: 30_000,
-            });
+            if (getVirtualizationProvider() !== PodmanVirtualizationProviders.HyperV) {
+              await playExpect(podmanOnboardingPage.machineCreationForm.userModeNetworkingCheckbox).not.toBeChecked({
+                timeout: 30_000,
+              });
+            }
           } else {
             await playExpect(podmanOnboardingPage.machineCreationForm.podmanMachineCPUs).toBeVisible({
               timeout: 30_000,
@@ -187,6 +193,10 @@ test.describe
                 PODMAN_MACHINE_NAME,
               );
               await playExpect(resourcesPodmanConnections.providerConnections).toBeVisible({ timeout: 10_000 });
+              await playExpect(resourcesPodmanConnections.connectionType).toBeVisible({ timeout: 10_000 });
+              await playExpect(resourcesPodmanConnections.connectionType).toHaveText(
+                getVirtualizationProvider() ?? getDefaultVirtualizationProvider(),
+              );
               await playExpect(resourcesPodmanConnections.resourceElement).toBeVisible({ timeout: 20_000 });
               await playExpect(resourcesPodmanConnections.resourceElementDetailsButton).toBeVisible();
               await resourcesPodmanConnections.resourceElementDetailsButton.click();
--- a/tests/playwright/src/specs/z-podman-machine-resources.spec.ts
+++ b/tests/playwright/src/specs/z-podman-machine-resources.spec.ts
@@ -20,6 +20,7 @@ import type { Locator } from '@playwright/test';
 
 import { ResourceElementActions } from '../model/core/operations';
 import { ResourceElementState } from '../model/core/states';
+import { PodmanVirtualizationProviders } from '../model/core/types';
 import { CreateMachinePage } from '../model/pages/create-machine-page';
 import { ResourceConnectionCardPage } from '../model/pages/resource-connection-card-page';
 import { ResourcesPage } from '../model/pages/resources-page';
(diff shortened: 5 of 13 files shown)


danivilla9:
I've ran the updated test in the GHA workflows successfully:
- WSL: https://github.com/podman-desktop/e2e/actions/runs/16342987696
- HyperV: https://github.com/podman-desktop/e2e/actions/runs/16346406735

The 'connectionType' parameter is correctly passed from the workflow input to the azure machine, and to the test case, selecting the specified connection type during the podman machine creation step, as you can see here:
```
Create Podman Machine 'podman-machine-rootless' with settings: rootless, usernet disabled, startnow disabled, and Wsl— tests/playwright/src/model/pages/forms/machine-creation-form.ts:86
   85 |   ): Promise<void> {
   86 |     return test.step(`Create Podman Machine '${machineName}' with settings: ${isRootful ? 'rootful' : 'rootless'}, ${enableUserNet ? 'usernet enabled' : 'usernet disabled'}, ${startNow ? 'startnow enabled' : 'startnow disabled'}${connectionType ? ', and ' + connectionType : ''}`, async () => {
```
```
Create Podman Machine 'podman-machine-rootless' with settings: rootless, usernet enabled, startnow disabled, and Hyperv— tests/playwright/src/model/pages/forms/machine-creation-form.ts:86
   85 |   ): Promise<void> {
   86 |     return test.step(`Create Podman Machine '${machineName}' with settings: ${isRootful ? 'rootful' : 'rootless'}, ${enableUserNet ? 'usernet enabled' : 'usernet disabled'}, ${startNow ? 'startnow enabled' : 'startnow disabled'}${connectionType ? ', and ' + connectionType : ''}`, async () => {
      |                 ^
   87 |       await playExpect(this.podmanMachineConfiguration).toBeVisible({
```

I had to increase the machine creation timeout because it was struggling to complete the creation process on the HyperV machines. 

Initially I intended to add this functionality to the `podman-machine-resources.spec.ts` file, but there is an issue with HyperV in this test case because it's not possible to have two HyperV machines running at the same time, so it will always fail if it tries to do so. For that reason, I moved it to `podman-machine-tests.spec.ts`

tests/playwright/src/model/core/types.ts
```diff
@@ -82,3 +82,8 @@ export const KubernetesResourceAttributes: Record<KubernetesResources, string[]>
   ],
   [KubernetesResources.Jobs]: ['Selected', 'Status', 'Name', 'Conditions', 'Completions', 'Age', 'Actions'],
 };
+
+export enum PodmanConnectionTypes {
```
odockal: `PodmanVirtualizationProviders` might be more precise ?
also, we know about others - `applehv` and `libkrun` on Mac and `qemu` on Linux - please look at exact naming in PD.

tests/playwright/src/model/pages/forms/machine-creation-form.ts
```diff
@@ -57,6 +61,10 @@ export class MachineCreationForm extends BasePage {
     this.userModeNetworkingCheckbox = this.podmanMachineConfiguration.getByRole('checkbox', {
       name: 'User mode networking',
     });
+    this.providerTypeDiv = this.podmanMachineConfiguration.getByLabel('Provider Type');
+    this.providerTypeWslOption = this.providerTypeDiv.getByRole('button', { name: 'wsl' });
```
odockal: Feels to me that we are missing better aria properties defined on PD side.

tests/playwright/src/model/pages/forms/machine-creation-form.ts
```diff
@@ -33,6 +34,9 @@ export class MachineCreationForm extends BasePage {
   readonly podmanMachineDiskSize: Locator;
   readonly rootPriviledgesCheckbox: Locator;
   readonly userModeNetworkingCheckbox: Locator;
+  readonly providerTypeDiv: Locator;
```
odockal: Shouldn't we have only a dropdown locator? and then a function to change the dropdown value based on input value?

tests/playwright/src/specs/podman-machine-tests.spec.ts
```diff
@@ -146,13 +147,19 @@ test.describe
         await resourcesPage.goToCreateNewResourcePage(RESOURCE_NAME);
       });
 
+      const connectionType = {
+        wsl: PodmanConnectionTypes.WSL,
+        hyperv: PodmanConnectionTypes.HyperV,
+      }[process.env.CONTAINERS_MACHINE_PROVIDER?.toLowerCase() ?? ''];
```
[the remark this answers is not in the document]
odockal: @danivilla9 The suggestion is reasonable.

tests/playwright/src/model/pages/forms/machine-creation-form.ts
```diff
@@ -67,22 +75,36 @@ export class MachineCreationForm extends BasePage {
...
       });
       await this.podmanMachineName.clear();
       await this.podmanMachineName.fill(machineName);
 
       await this.ensureCheckboxState(isRootful, this.rootPriviledgesCheckbox);
+
       if (isWindows) {
         await this.ensureCheckboxState(enableUserNet, this.userModeNetworkingCheckbox);
+        if (connectionType === PodmanConnectionTypes.HyperV) {
+          await playExpect(this.providerTypeWslOption).toBeVisible({ timeout: 10_000 });
```
odockal: This is very fixed workflow how to switch the dropdown. Look at `create-kind-cluster-page.ts` which also has similar action handled in simpler way.

tests/playwright/src/specs/podman-machine-tests.spec.ts
```diff
@@ -146,13 +147,19 @@ test.describe
         await resourcesPage.goToCreateNewResourcePage(RESOURCE_NAME);
       });
 
+      const connectionType = {
```
odockal: Also the constant can be defined outside of the describe, since it wont change through the whole test.

odockal (changes requested):
Please amend the commit message to a simpler description. During squashing the commit you should keep just minimum of the commits messages.

In this PR we are only updating one test, but there are more tests that would benefit from this additional test check, it is gonna be follow up?

I am missing also mac part here, also follow up perhaps?

On every system/architecture, we know the default virtualization provider. In combination with test env. (`CONTAINERS_MACHINE_PROVIDER`), we could be able to reliably define the initial value, right? That is something I am missing in the code.

tests/playwright/src/utility/provider.ts
```diff
@@ -0,0 +1,42 @@
...
+    ? PodmanVirtualizationProviders.WSL
+    : envProvider === 'hyperv'
+      ? PodmanVirtualizationProviders.HyperV
+      : envProvider === 'applehv'
+        ? PodmanVirtualizationProviders.AppleHV
+        : envProvider === 'libkrun'
+          ? PodmanVirtualizationProviders.LibKrun
+          : envProvider === 'qemu'
+            ? PodmanVirtualizationProviders.Qemu
+            : undefined;
```
cbr7: I'm honestly confused by this implementation, what is this supposed to solve?

tests/playwright/src/utility/provider.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function getDefaultVirtualizationProvider(): PodmanVirtualizationProviders {
+  if (isWindows) {
+    return PodmanVirtualizationProviders.WSL;
+  } else if (isMac) {
+    return PodmanVirtualizationProviders.AppleHV;
+  } else return PodmanVirtualizationProviders.Qemu;
```
cbr7: On which platform is qemu default provider?

tests/playwright/src/specs/z-podman-machine-resources.spec.ts
```diff
@@ -152,14 +154,20 @@ for (const { PODMAN_MACHINE_NAME, MACHINE_VISIBLE_NAME, isRoot, userNet } of mac
         await podmanResources.createButton.click();
 
         const createMachinePage = new CreateMachinePage(page);
+        //Hyperv can't have 2 podman machines at the same time, in that case, change to default one
+        const virtualizationProviderToUse =
+          virtualizationProvider === PodmanVirtualizationProviders.HyperV
+            ? getDefaultVirtualizationProvider()
+            : (virtualizationProvider ?? getDefaultVirtualizationProvider());
```
cbr7: I dont know what this does, why if virtualizationProvider === HyperV does this return getDefaultVirtualizationProvider? Wouldnt that return WSL in this context?

tests/playwright/src/model/pages/resources-page.ts
```diff
@@ -47,4 +50,29 @@ export class ResourcesPage extends SettingsPage {
...
+    virtualizationProvider: PodmanVirtualizationProviders,
+  ): Promise<void> {
+    return test.step(`Verify Podman Provider is ${virtualizationProvider}`, async () => {
+      await playExpect(this.heading).toBeVisible();
+      const machineCard = new ResourceConnectionCardPage(this.page, 'podman', machineName);
+      playExpect(await machineCard.doesResourceElementExist()).toBeTruthy();
+
+      await machineCard.resourceElement.getByLabel('Connection Type').scrollIntoViewIfNeeded();
+      playExpect(await machineCard.resourceElement.getByLabel('Connection Type').innerText()).toContain(
+        virtualizationProvider,
```
cbr7: Why not add this locator to the relevant POM? there is no reason to define it just in one method.

tests/playwright/src/model/pages/forms/machine-creation-form.ts
```diff
@@ -84,6 +95,25 @@ export class MachineCreationForm extends BasePage {
...
+          exact: true,
+        });
+        await playExpect(providerTypeButton).toBeVisible({ timeout: 10_000 });
+        await providerTypeButton.scrollIntoViewIfNeeded();
+        await providerTypeButton.click();
+
+        const newProviderTypeButton = this.podmanMachineConfiguration.getByRole('button', {
+          name: virtualizationProvider,
+          exact: true,
+        });
```
cbr7: Please construct a method in this POM which returns the custom locator for the provided params, there is no need to have these locators constructed here only.

tests/playwright/src/utility/provider.ts
danivilla9: I figured that it would be the case for Linux, but after some thought I think that the default option in that platform is running Podman natively so I have to change that

tests/playwright/src/utility/provider.ts
danivilla9: We have two strings to compare, the podman provider specified when running the tests and the podman provider that appears in the Resources/Podman/Podman Machine. I want to map the value from the specified envvar to a value of an enum in order to use it later in the tests. Maybe it's a bit of a stretch but I decided to evaluate them in lowercase because in general it's a good practice and I didn't want to run into issues during test execution because of a capital letter or something. 

But I agree that the implementation is not great, I'm changing it right now.

tests/playwright/src/specs/z-podman-machine-resources.spec.ts
danivilla9: Yes, it's on purpose. When I tried this test using hyperv I ran into an error because the test tries to run two podman machines at the same time, but that's not possible for hyperv (I think this is expected), so the test fails. So, in short:
- Is the provider hyperv? => Use the default one (unless something weird happened, wsl)
- Is the provider undefined? => Use the default one (depends on plantform)

Maybe the implementation would be clearer if it specified WSL to be used? I just did not want to hardcode it just in case, as the same result is achieved in a safer way.

cbr7: I'm not sure I understand? If the provider is set to hyperV, why would I want it to use WSL? Maybe I'm not understanding the problem this is solving, but it's certainly counterintuitive.

danivilla9: Hmmm maybe it would be more intuitive to just skip the test entirely if hyperv is defined, given that it will never work

tests/playwright/src/model/pages/forms/machine-creation-form.ts
danivilla9: I went for this approach for a couple of reasons:
- There is no difference between clicking on the div that _contains_ the button that opens the dropdown and the actual button that opens the dropdown, but the ways to find the former are way shorter and cleaner than the latter: `aria-label="Provider Type"` vs `id="input-standard-podman.factory.machine.win.provider" name="podman.factory.machine.win.provider"`
- The dropdown buttons are inside the div too, so in my head it's more understandable this way
- Given that the dropdown will only be opened when the default option is changed, there will never be a conflict in finding the locator of the other option by name (there will be always only one)

tests/playwright/src/model/pages/forms/machine-creation-form.ts
```diff
@@ -84,6 +102,15 @@ export class MachineCreationForm extends BasePage {
       if (isWindows) {
         await this.ensureCheckboxState(enableUserNet, this.userModeNetworkingCheckbox);
       }
+      if (virtualizationProvider && virtualizationProvider !== getDefaultVirtualizationProvider()) {
+        await playExpect(this.providerTypeDiv).toBeVisible({ timeout: 10_000 });
+        playExpect(this.providerTypeDiv.innerText()).toContain(getDefaultVirtualizationProvider());
```
odockal: instead of getting innerText and asserting over the value, you should simply call `expect(providerTypeDiv).toContainText(...)` which evaluates the element in multiple iterations.

tests/playwright/src/model/pages/forms/machine-creation-form.ts
```diff
@@ -84,6 +102,15 @@ export class MachineCreationForm extends BasePage {
...
+        await this.providerTypeDiv.scrollIntoViewIfNeeded();
+        await this.providerTypeDiv.click();
+        await playExpect(this.providerTypeDropdownOption).toBeVisible({ timeout: 10_000 });
+        await this.providerTypeDropdownOption.click();
+        playExpect(this.providerTypeDiv.innerText()).toContain(virtualizationProvider);
```
odockal: same as above.

tests/playwright/src/model/pages/resources-page.ts
```diff
@@ -47,4 +50,27 @@ export class ResourcesPage extends SettingsPage {
...
+   * @throws Will throw an error if the expected virtualization provider is not found or doesn't match
+   */
+  public async verifyVirtualizationProvider(
+    machineName: string,
+    virtualizationProvider: PodmanVirtualizationProviders,
+  ): Promise<void> {
+    return test.step(`Verify Podman Provider is ${virtualizationProvider}`, async () => {
+      await playExpect(this.heading).toBeVisible();
+      const machineCard = new ResourceConnectionCardPage(this.page, 'podman', machineName);
+      playExpect(await machineCard.doesResourceElementExist()).toBeTruthy();
```
odockal: use poll for more robust check.

tests/playwright/src/model/pages/resources-page.ts
```diff
@@ -47,4 +50,27 @@ export class ResourcesPage extends SettingsPage {
...
+
+      await machineCard.connectionType.scrollIntoViewIfNeeded();
+      playExpect(await machineCard.connectionType.innerText()).toContain(virtualizationProvider);
```
odockal: prefer `toHaveText` or `toContainText`.

tests/playwright/src/utility/provider.ts
```diff
@@ -0,0 +1,46 @@
...
+  return envProvider ? PROVIDER_MAP[envProvider?.toLowerCase()] : undefined;
+}
+
+export function getDefaultVirtualizationProvider(): PodmanVirtualizationProviders {
+  if (isWindows) {
+    return PodmanVirtualizationProviders.WSL;
+  }
+
+  if (isMac) {
+    return PodmanVirtualizationProviders.AppleHV;
```
odockal: On mac the default provider is libkrun starting with 1.21.0.

odockal (changes requested):

danivilla9:
I was able to run successfully three of the updated tests on the windows workflows:
- WSL: https://github.com/podman-desktop/e2e/actions/runs/17320846708
- Hyperv: https://github.com/podman-desktop/e2e/actions/runs/17377112643

The kubernetes one ran successfully here: https://github.com/podman-desktop/e2e/actions/runs/17399448643

Missing testing it on mac ci

tests/playwright/src/model/pages/resources-page.ts
```diff
@@ -48,4 +51,27 @@ export class ResourcesPage extends SettingsPage {
...
+  /**
+   * Verifies that a Podman machine has the specified virtualization provider type.
+   * This method checks that the machine card exists and displays the correct connection type.
+   *
+   * @param machineName - The name of the machine to verify
+   * @param virtualizationProvider - The expected virtualization provider type (e.g., PodmanVirtualizationProviders.WSL, PodmanVirtualizationProviders.QEMU, PodmanVirtualizationProviders.HyperV)
+   * @returns A Promise that resolves when the verification is complete
+   * @throws Will throw an error if the expected virtualization provider is not found or doesn't match
+   */
+  public async verifyVirtualizationProvider(
```
odockal: This function does not fit here. It should be in utilities. you are using hardcoded `podman` region to get ResourceConnectionCardPage object. If you look at tests, there is usually the same initialization after your call, initializing the object twice (ie.: https://github.com/podman-desktop/podman-desktop/pull/12874/commits/c2dcf7643c984288bd6ad66eeb74a9a0775411f0#diff-2552cd838286a6b2b3a13d7380c6e70d92924f23e7e2b9d6706d998645f3aaa8R122). 
You should use that object as an argument of a check in function defined in right place.

tests/playwright/src/model/pages/forms/machine-creation-form.ts
odockal: @danivilla9 you still did not answered this comment.

tests/playwright/src/specs/z-podman-machine-resources.spec.ts
```diff
@@ -160,14 +177,19 @@ for (const { PODMAN_MACHINE_NAME, MACHINE_VISIBLE_NAME, isRoot, userNet } of mac
...
         const createMachinePage = new CreateMachinePage(page);
+
         const resourcePage = await createMachinePage.createMachine(PODMAN_MACHINE_NAME, {
           isRootful: isRoot,
           enableUserNet: userNet,
           setAsDefault: false,
           startNow: false,
+          virtualizationProvider: getVirtualizationProvider(),
         });
+        await resourcePage.verifyVirtualizationProvider(
```
odockal: same as above.

tests/playwright/src/specs/z-podman-machine-tests.spec.ts
```diff
@@ -158,12 +159,16 @@ test.describe
...
           enableUserNet: true,
           startNow: false,
+          virtualizationProvider: getVirtualizationProvider(),
         });
         await playExpect(podmanMachineCreatePage.goBackButton).toBeEnabled({ timeout: 180_000 });
         await podmanMachineCreatePage.goBackButton.click();
       });
 
-      await playExpect(resourcesPage.heading).toBeVisible();
+      await resourcesPage.verifyVirtualizationProvider(
```
odockal: same as above.

tests/playwright/src/utility/provider.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { isMac, isWindows } from '..';
```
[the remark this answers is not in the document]
odockal: @danivilla9 This needs to be updated to correct import.

tests/playwright/src/utility/provider.ts
```diff
@@ -0,0 +1,46 @@
+/**********************************************************************
```
odockal, line 1: You need to add file on index.

tests/playwright/src/specs/z-podman-machine-onboarding.spec.ts
```diff
@@ -141,7 +143,9 @@ test.describe
           await playExpect(podmanOnboardingPage.machineCreationForm.rootPriviledgesCheckbox).toBeChecked();
           await playExpect(podmanOnboardingPage.machineCreationForm.startNowCheckbox).toBeChecked();
 
-          if (os.platform() === 'win32') {
+          await podmanOnboardingPage.specifyVirtualizationProvider(getVirtualizationProvider());
```
odockal: same as above.

odockal (changes requested):

tests/playwright/src/utility/operations.ts
```diff
@@ -503,3 +506,64 @@ export async function readFileInVolumeFromCLI(volumeName: string, fileName: stri
...
+ *
+ * @param resourceConnectionCardPage - The resource connection card page to verify
+ * @param machineName - The name of the machine to verify
+ * @param virtualizationProvider - The expected virtualization provider type (e.g., PodmanVirtualizationProviders.WSL, PodmanVirtualizationProviders.HyperV...)
+ * @returns A Promise that resolves when the verification is complete
+ * @throws Will throw an error if the expected virtualization provider is not found or doesn't match
+ */
+export async function verifyVirtualizationProvider(
+  resourceConnectionCardPage: ResourceConnectionCardPage,
+  machineName: string,
```
odockal: ResourceConnectionCardPage might already contain a machine name. It is on of the param you can use new initializing object.

tests/playwright/src/utility/operations.ts
```diff
@@ -503,3 +506,64 @@ export async function readFileInVolumeFromCLI(volumeName: string, fileName: stri
...
+ * Specifies the virtualization provider for a Podman machine during creation through the onboarding page.
+ * This method selects the specified virtualization provider from the dropdown if it differs from the default.
+ * If no provider is specified or it matches the default, no action is taken.
+ *
+ * @param machineCreationForm - The machine creation form page object containing UI elements
+ * @param virtualizationProvider - The virtualization provider to select (e.g., PodmanVirtualizationProviders.WSL, PodmanVirtualizationProviders.HyperV, etc.), or undefined to use default
+ * @returns A Promise that resolves when the provider selection is complete
+ * @throws Will throw an error if the provider dropdown is not accessible or the specified provider is not available
+ */
+export async function specifyVirtualizationProvider(
```
odockal: Why it is not in machine-creation-form class? It can be called from there, no?

tests/playwright/src/specs/z-podman-machine-resources.spec.ts
```diff
@@ -160,12 +168,19 @@ for (const { PODMAN_MACHINE_NAME, MACHINE_VISIBLE_NAME, isRoot, userNet } of mac
...
         const createMachinePage = new CreateMachinePage(page);
+
         const resourcePage = await createMachinePage.createMachine(PODMAN_MACHINE_NAME, {
           isRootful: isRoot,
           enableUserNet: userNet,
           setAsDefault: false,
           startNow: false,
+          virtualizationProvider: getVirtualizationProvider(),
         });
+        await verifyVirtualizationProvider(
```
odockal: if this would be moved under line 186, you do not have to chain the locators in ResourceConnectionCardPage again inside the function, also no need to passing machinename then.

tests/playwright/src/specs/kubernetes-status-bar-providers.spec.ts
```diff
@@ -103,7 +110,16 @@ test.describe.serial('Status bar providers feature verification', { tag: '@k8s_e
     const podmanResources = new ResourceConnectionCardPage(page, 'podman');
     await podmanResources.createButton.click();
     const createMachinePage = new CreateMachinePage(page);
-    await createMachinePage.createMachine(newMachineName, { startNow: true, setAsDefault: false });
+    await createMachinePage.createMachine(newMachineName, {
+      startNow: true,
+      setAsDefault: false,
+      virtualizationProvider: getVirtualizationProvider(),
+    });
+    await verifyVirtualizationProvider(
```
odockal: also moving under line 124 would do the job.

tests/playwright/src/utility/operations.ts
```diff
@@ -503,3 +506,64 @@ export async function readFileInVolumeFromCLI(volumeName: string, fileName: stri
...
+      const currentProviderText = await machineCreationForm.providerType.innerText();
+
+      // Only proceed if the new provider type is different from the current one
+      if (virtualizationProvider !== currentProviderText) {
+        // Click the current provider to open the dropdown
+        await machineCreationForm.providerType.scrollIntoViewIfNeeded();
+        await machineCreationForm.providerType.click();
+
+        // Select the new provider option
+        machineCreationForm.providerType = machineCreationForm.getProviderType(virtualizationProvider);
```
odockal: I think this will fails as once the selectio is opened, we see three buttons (default and two for selection options) of which two contains the same name (ie. wsl).

odockal (changes requested):

tests/playwright/src/utility/operations.ts
danivilla9: It was initially there but from a previous comment I thought you expected it to be in operations alongside verifyVirtualizationProvider. I moved it back.

odockal (approved):
LGTM. Let's see the results from the runs.

danivilla9:
Results from the runs:
- Hyperv: https://github.com/podman-desktop/e2e/actions/runs/17607469947 some are skipped because they require 2 hyperv machines at the same time and that it's not possible, but the rest looks good
- WSL: https://github.com/podman-desktop/e2e/actions/runs/17607473615/job/50021419120 on windows 10 there were no problems, but on windows 11 the job timed out, I've seen it happening in the 3 last executions, I'd say it's not related to this PR. I've re-run the job here to confirm: https://github.com/podman-desktop/e2e/actions/runs/17764830281

I've run the podman matrix on wsl here to see if the timeout persists or if it's been a fluke: https://github.com/podman-desktop/e2e/actions/runs/17616457146
The results seem quite positive, the timeout issue has not appeared in any of the 4 jobs and the tests behave as expected. The problem is that for 2 of the jobs there was a problem importing the test results.

Tried to run the mac ones but both failed to import the results. From the logs I saw that a GET quota had been exhausted, but I'm not 100% certain that's the culprit.

odockal:
@danivilla9 I think we can merge and monitor the next nightly runs to see if there is a problem and address it afterwards.

PR #12875, opened 2025-06-17
chore: notify development folder instance when extension id is loaded/removed from extension loader

benoitf:
### What does this PR do?
add notification

#8616 
### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extension/extension-development-folders.ts
+++ b/packages/main/src/plugin/extension/extension-development-folders.ts
@@ -36,6 +36,8 @@ export class ExtensionDevelopmentFolders {
 
   #extensionAnalyzer: ExtensionAnalyzer;
 
+  #externalExtensionIds: string[] = [];
+
   // event that will be fired
   #onDidUpdateDevelopmentFolders: Emitter<ExtensionDevelopmentFolderInfo[]> = new Emitter<
     ExtensionDevelopmentFolderInfo[]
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/extension/extension-development-folders.ts
```diff
@@ -147,4 +156,13 @@ export class ExtensionDevelopmentFolders {
   getDevelopmentFolders(): ExtensionDevelopmentFolderInfo[] {
     return this.#developmentFolders.map(path => ({ path }));
   }
+
+  addExternalExtensionId(extensionId: string): void {
```
axel7083, line 160: question: what would happen if I have already the extension installed (let's say from the catalog), then try to start in dev mode?

axel7083 (commented):
not sure to understand the title of the PR `notify`? Who is notified? the extension-development-folder?

Feel more like registering logic, than notification?

packages/main/src/plugin/extension/extension-development-folders.ts
benoitf: it'll tell you that you can't start in dev mode this extension. You will need to uninstall it first.

benoitf:
> not sure to understand the title of the PR notify? Who is notified? the extension-development-folder?

it notifies extension-development-folder that there are external extension id.

updated title

axel7083 (approved):

PR #12896, opened 2025-06-18
refactor(frontend): make ContainerGroupPartInfoUI id property non-nullable

### LINKED ISSUE #12842: bug(ContainerList): having two pods name the same on two different engine group them
### Bug description

Let's say I have two machines (`podman-a` and `podman-b`), let's deploy the same pod (from a kube yaml), inside the ContainerList page, the containers from both machine will be under a same pod object.

This is a problem, as it break the delete action.

### Operating system

Fedora 42

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. create a `pod.yaml` file
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-pod
spec:
  containers:
    - name: nginx-container
      image: docker.io/library/nginx:latest
      ports:
        - name: http
          containerPort: 80
```
2. deploy the pod on each machine
````
$: podman --connection="podman-a" kube play ./pod.yaml
$: podman --connection="podman-b" kube play ./pod.yaml
````
3. open ContainerList page
4. assert single pod, and two containers under it
Image

### Relevant log output

```shell

```

### Additional context

The root cause is inside the `ContainerUtils#getContainerGroups`, to be able to group containers, each of them have an optional `groupInfo: ContainerGroupPartInfoUI` property, and `ContainerGroupPartInfoUI#name` is used to group the containers, this is a problem as the name can be, as we saw above, the same between different engine.

https://github.com/podman-desktop/podman-desktop/blob/bb1c3c0fbddcbabd2d5950d2939f6e920bc0d40f/packages/renderer/src/lib/container/container-utils.ts#L235

The `ContainerGroupPartInfoUI` has different property that are optional, that should probably not

https://github.com/podman-desktop/podman-desktop/blob/eaa421aff79ec65f5e739287b6f1179e158d05f6/packages/renderer/src/lib/container/ContainerInfoUI.ts#L37-L38

Those two, should be mandatory, the only case where it has no value is when using compose, but we could default to the engineId + compose name instead of undefined. For pods the id is the id of the pod, which is unique to the engine.

---

## Solution proposal

Make the `ContainerGroupPartInfoUI#id` and `ContainerGroupPartInfoUI#engineId` non-optional, and use it to group containers.

Quick POC, prove pretty effective, using the id instead of the name

Image

Poc branch https://github.com/axel7083/podman-desktop/tree/fix/container-list/grouping

axel7083:
### What does this PR do?

Today to group containers in the `ContainersList` ui, we use the group name, which is either the compose project or the pod name. However those field are only unique per engine. Meaning we are grouping same name resources among engines.

To fix https://github.com/podman-desktop/podman-desktop/issues/12842 we need to rely on a unique value, however the `groupInfo#id` is nullable because a compose project has no unique identifier, this PR makes the id property non-nullable, and uses `{engineId}:{composeProject}` as unique id for compose groups, making it unique across engines. 

For standalone containers, no id where provided for the group, we can just use the container id, as it is unique.

### Screenshot / video of UI

No visual changes

Required for https://github.com/podman-desktop/podman-desktop/issues/12842

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/compose/ComposeDetailsLogs.spec.ts
+++ b/packages/renderer/src/lib/compose/ComposeDetailsLogs.spec.ts
@@ -65,6 +65,7 @@ const containerInfoUIMock: ContainerInfoUI = {
   groupInfo: {
     name: 'foobar',
     type: ContainerGroupInfoTypeUI.COMPOSE,
+    id: 'podman:foobar',
   },
   selected: false,
   created: 0,
--- a/packages/renderer/src/lib/compose/ComposeDetailsSummary.spec.ts
+++ b/packages/renderer/src/lib/compose/ComposeDetailsSummary.spec.ts
@@ -44,6 +44,7 @@ const fakeContainer1: ContainerInfoUI = {
   groupInfo: {
     name: 'group1',
     type: ContainerGroupInfoTypeUI.STANDALONE,
+    id: 'fakeContainerID1',
   },
   selected: false,
   created: 1234,
@@ -70,6 +71,7 @@ const fakeContainer2: ContainerInfoUI = {
   groupInfo: {
     name: 'group2',
     type: ContainerGroupInfoTypeUI.STANDALONE,
+    id: 'fakeContainerID2',
   },
   selected: false,
   created: 1234,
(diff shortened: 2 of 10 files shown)


jeffmaury (commented):
Why do you need to have the engineId in the id field where there is already an existing engineId field in the structure ?

axel7083:
> Why do you need to have the engineId in the id field where there is already an existing engineId field in the structure ?

@jeffmaury the `groupInfo#id` for pods, the id is unique across engines, for compose, there is no unique id, so for compose group, we need to find something unique, that's mixing name and engineId is probably the easier solution.

Having the id unique, ease the usage, as if it is not, we would need to be careful everytime we use the id (which is error prone)

jeffmaury (approved):
LGTM

PR #12905, opened 2025-06-19
refactor(core): move default protocol configuration to Main

### LINKED ISSUE #12812: refactor(main/src/index): move default protocol logic to `Main`

From the main/src/index.ts we should move the `whenReady` logic to the Main.

Move the following logic to Main class.

https://github.com/podman-desktop/podman-desktop/blob/6ef6bcd5d5eaeb893159f1aa77086c186eef53d2/packages/main/src/index.ts#L92-L98

_No response_

### Additional context

Parent issue https://github.com/podman-desktop/podman-desktop/issues/10042

axel7083:
### What does this PR do?

Moving the default protocol configuration from `packages/main/src/index.ts` to the Main class. I also added unit tests to cover this feature, as this was not covered.

Fixes https://github.com/podman-desktop/podman-desktop/issues/12812

### How to test this PR?

### DIFF
--- a/packages/main/src/main.spec.ts
+++ b/packages/main/src/main.spec.ts
@@ -182,8 +182,7 @@ describe('ElectronApp#on window-all-closed', () => {
 });
 
 describe('AppPlugin', () => {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const plugins: Array<new (...args: any[]) => AppPlugin> = [DefaultProtocolClient];
+  const plugins: Array<AppPlugin> = [DefaultProtocolClient.prototype];
 
   let promiseWithResolvers: PromiseWithResolvers<void>;
 
@@ -204,8 +203,7 @@ describe('AppPlugin', () => {
   test('expect AppPlugin#onReady to be called when ElectronApp#whenReady resolved', async () => {
     // expect each plugin to have been instantiated
     plugins.forEach(plugin => {
-      expect(plugin).toHaveBeenCalledOnce();
-      expect(plugin.prototype.onReady).not.toHaveBeenCalled();
+      expect(plugin.onReady).not.toHaveBeenCalled();
     });
 
     // simulate on ready
@@ -214,7 +212,7 @@ describe('AppPlugin', () => {
     await vi.waitFor(() => {
       // expect each plugin to have been instantiated
       plugins.forEach(plugin => {
-        expect(plugin.prototype.onReady).toHaveBeenCalled();
+        expect(plugin.onReady).toHaveBeenCalled();
       });
     });
   });
@@ -222,8 +220,7 @@ describe('AppPlugin', () => {
   test('expect AppPlugin#dispose to be called when ElectronApp#on("before-quit") is emitted ', async () => {
     // expect each plugin to have been instantiated
     plugins.forEach(plugin => {
-      expect(plugin).toHaveBeenCalledOnce();
-      expect(plugin.prototype.dispose).not.toHaveBeenCalled();
+      expect(plugin.dispose).not.toHaveBeenCalled();
     });
 
     const listener = getElectronAppListener('before-quit');
@@ -232,7 +229,7 @@ describe('AppPlugin', () => {
     await vi.waitFor(() => {
       // expect each plugin to be disposed
       plugins.forEach(plugin => {
-        expect(plugin.prototype.dispose).toHaveBeenCalled();
+        expect(plugin.dispose).toHaveBeenCalled();
       });
     });
   });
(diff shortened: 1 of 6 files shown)


packages/main/src/main.ts
```diff
@@ -101,6 +101,37 @@ export class Main {
      * Setup {@link ElectronApp.on} listeners
      */
     this.app.on('window-all-closed', this.onWindowAllClosed.bind(this));
+
+    /**
+     * Register {@link Main#whenReady} for ready event
+     */
+    this.app.whenReady().then(this.whenReady.bind(this)).catch(console.error);
```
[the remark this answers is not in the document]
axel7083: This is a refactor, and should not introduce new logic, the existing logic inside the `index.ts` does not handle error for whenReady promise.

packages/main/src/main.ts
```diff
@@ -101,6 +101,37 @@ export class Main {
...
+  protected configureDefaultProtocolClient(): void {
+    if (!this.app.isReady()) throw new Error('app is not ready');
+
+    if (!import.meta.env.PROD) return;
+
+    if (isWindows()) {
+      this.app.setAsDefaultProtocolClient('podman-desktop', process.execPath, process.argv);
+    } else {
+      this.app.setAsDefaultProtocolClient('podman-desktop');
+    }
```
benoitf: IMHO this should be moved to its own class

axel7083: No problem 👍

axel7083:
@benoitf following our discussion, introducing the interface `AppPlugin`, it exposes an `onReady` method, and is disposable.

In the `Main` class, we have a readonly list of `AppPlugin` created inside the Main constructor, when electron resolve the `whenReady()` promise, all AppPlugin#onReady are called.

For completeness, the plugins are disposed when Electron Emit `before-quit`, we already use the `before-quit` event to stop all extensions

https://github.com/podman-desktop/podman-desktop/blob/04872aa4ff77d81c2012c0e44ab653d57f2b05e6/packages/main/src/index.ts#L57-L63

This logic should be moved later on, to match the `AppPlugin` pattern

packages/main/src/main.spec.ts
```diff
@@ -172,3 +180,60 @@ describe('ElectronApp#on window-all-closed', () => {
     expect(ELECTRON_APP_MOCK.quit).toHaveBeenCalledOnce();
   });
 });
+
+describe('AppPlugin', () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const plugins: Array<new (...args: any[]) => AppPlugin> = [DefaultProtocolClient];
```
benoitf: can we avoid any there ?

benoitf (commented):
LGTM except the `any` type that should be removed and replaced by something else

packages/main/src/main.spec.ts
axel7083: I tried my best :( if you put `unknown` it complain that the arguments do not match.. If I remove everything it says that argument do not match :/

packages/main/src/plugin/app-ready/default-protocol-client.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  expect(ELECTRON_APP_MOCK.setAsDefaultProtocolClient).toHaveBeenCalledWith('podman-desktop');
+});
+
+test('on windows ElectronApp#setAsDefaultProtocolClient should be called with process args', async () => {
+  vi.mocked(isWindows).mockReturnValue(true);
+
+  await plugin.onReady();
+
+  expect(ELECTRON_APP_MOCK.setAsDefaultProtocolClient).toHaveBeenCalledOnce();
+  expect(ELECTRON_APP_MOCK.setAsDefaultProtocolClient).toHaveBeenCalledWith('podman-desktop', '/foo/bar', [
```
MarsKubeX, line 75: Non-blocking: If you think these two parameters ('podman-desktop' and '/foo/bar') might be reused by other tests in this spec in the future, it could be a good idea to define them as global constants for this file.

packages/main/src/main.spec.ts
axel7083: found a work-around by using the prototype, not really a big fan, bug avoid using the any

benoitf (approved):

packages/main/src/plugin/app-ready/default-protocol-client.spec.ts
axel7083: I would say, that the first argument `podman-desktop` is hardcoded in the logic, and would need to be exported from there, so I would say, if the file grow, then why not, for now, it is not planned.

PR #12906, opened 2025-06-19
chore(podman): remove promisify usage for `node:fs` function

axel7083:
### What does this PR do?

While working on https://github.com/podman-desktop/podman-desktop/issues/12021 I noticed that we were using promisify on `node:fs` function, I don't want to duplicate those line, so replacing those by their corresponding `node:fs/promise` implementation.

Required for https://github.com/podman-desktop/podman-desktop/issues/12021

### How to test this PR?

- [x] Existing tests ensure no regression


benoitf (approved):

benoitf:
is there is a rule to ban promisify ? :-)

axel7083:
> is there is a rule to ban promisify ? :-)

it may be used for other library, but yeah would be nice to ban promisify for `node:fs`

PR #12914, opened 2025-06-19
fix: change separators for fine grained events

### LINKED ISSUE #12912: pin/unpin does not work
### Bug description

https://github.com/user-attachments/assets/dada4c49-c985-4182-92ec-f728667d0a14

The regression has been introduced with https://github.com/podman-desktop/podman-desktop/pull/12636

This new feature expects that events names do not contain columns (:). But it not the case for a few events, especially:
```
PIN_OPTIONS_UPDATE = 'statusbar:pin:update',
```

### Operating system

mac

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

feloy:

### What does this PR do?

Change separators for fine grained events in stores

### Screenshot / video of UI

N/A

Fixes #12912 

### How to test this PR?

Check that pin/unpin works correctly again, and check that fine grained events are still supported (see #12591)


packages/renderer/src/stores/event-store.spec.ts
```diff
@@ -87,7 +87,7 @@ test.each<{
   },
   {
     name: 'window event with key matches',
-    listenedWindowEvent: 'my-event:my-key',
+    listenedWindowEvent: 'my-event>my-key',
```
benoitf: use the constants there ?

axel7083 (commented):
> This new feature expects that events names do not contain columns (:). But it not the case for a few events, especially:

Why `:` are not supported ?

feloy:
> > This new feature expects that events names do not contain columns (:). But it not the case for a few events, especially:
> 
> Why `:` are not supported ?

When using `:` , events names containing this character are not interpreted correctly. For example the event name `statusbar:pin:update` is considered as the event name `statusbar`, the rest being the "sub-event"

benoitf (approved):

PR #12915, opened 2025-06-19
fix(frontend): group containers by groupId

### LINKED ISSUE #12842: bug(ContainerList): having two pods name the same on two different engine group them

axel7083:
### What does this PR do?

To group containers (compose, pods) before the `GroupInfo#name` was used, but this is not unique accros engines, this PR changes to use the `GroupInfo#id`, which is now a mandatory and unique field thanks to https://github.com/podman-desktop/podman-desktop/pull/12896

### Screenshot / video of UI

| Before | After |
| --- | --- |
| image | image |

Fixes https://github.com/podman-desktop/podman-desktop/issues/12842

### How to test this PR?

**Testing manually**

1. assert have two podman connection (have two machines)
2. create a `pod.yaml` file
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-pod
spec:
  containers:
    - name: nginx-container
      image: docker.io/library/nginx:latest
      ports:
        - name: http
          containerPort: 80
```
3. deploy the pod on each machine
````
$: podman --connection="podman-a" kube play ./pod.yaml
$: podman --connection="podman-b" kube play ./pod.yaml
````
4. open ContainerList page
5. assert you have two pods

### DIFF
--- a/packages/renderer/src/lib/container/container-utils.spec.ts
+++ b/packages/renderer/src/lib/container/container-utils.spec.ts
@@ -112,10 +112,42 @@ test('container group status should be running when all compose containers are r
     containerUtils.getContainerInfoUI(containerInfo),
     containerUtils.getContainerInfoUI(containerInfo2),
   ]);
+  expect(groups).toHaveLength(1);
   const group = groups[0];
   expect(group.name).toBe(groupName);
   expect(group.type).toBe(ContainerGroupInfoTypeUI.COMPOSE);
   expect(group.status).toBe('RUNNING');
+});
+
+test('same group name but different engine is should results in two groups', async () => {
+  const groupName = 'compose-group';
+  const containerInfo = {
+    Id: 'container1',
+    Image: 'docker.io/kindest/node:foobar',
+    Labels: { 'com.docker.compose.project': groupName },
+    Names: ['container1'],
+    State: 'RUNNING',
+    ImageID: 'sha256:dummy-sha256',
+    engineId: 'docker',
+  } as unknown as ContainerInfo;
+
+  const groups = containerUtils.getContainerGroups([
+    containerUtils.getContainerInfoUI({
+      ...containerInfo,
+      engineId: 'podman-1',
+    }),
+    containerUtils.getContainerInfoUI({
+      ...containerInfo,
+      engineId: 'podman-2',
+    }),
+  ]);
+  expect(groups).toHaveLength(2);
+  const [groupA, groupB] = groups;
+
+  expect(groupA.name).toBe(groupName);
+  expect(groupB.name).toBe(groupName);
+
+  expect(groupA.id).not.eq(groupB.id);
 });
 
 test('container group status should be stopped when any compose container is stopped', async () => {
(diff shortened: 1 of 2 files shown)


axel7083:
Mmmm new problem :p

https://github.com/user-attachments/assets/dc946d7a-5c66-4401-8ee0-c78fca456ac3

--- 

The Table uses the `object.name` as id for the collapsed, which is problematic, as name is not unique!! cc @deboer-tim 

https://github.com/podman-desktop/podman-desktop/blob/74f54f5d9523326798eb838e95c90024b8686256/packages/ui/src/lib/table/Table.svelte#L254

gastoner:
On my linux machine I'm able to start only one machine at the time => I had to change the name so it don't collapse at the same time (as on your video).

So after fixing the thing above (I think it is only on Windows where you can have multiple machines running at one point?)
LGTM

benoitf (approved):

PR #12919, opened 2025-06-20
fix(patch): patched kubernetes/client-node package

### LINKED ISSUE #12747: Validation error shown when creating kind cluster with Contour ingress controller
### Bug description

After merging commit [11d9330](https://github.com/podman-desktop/podman-desktop/commit/11d93306274ba9d76081f3d78c0e6baf7c4b8b7a), we are encountering a validation error message when attempting to create a kind cluster with an ingress controller. Despite the error, the cluster appears to be created successfully and remains functional.

Observed error: 

`Error: Error: Failed to create kind cluster. CustomResourceDefinition.apiextensions.k8s.io "contourdeployments.projectcontour.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[contour].properties[deployment].properties[strategy].properties[rollingUpdate].properties[maxSurge].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[contour].properties[deployment].properties[strategy].properties[rollingUpdate].properties[maxUnavailable].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[envoy].properties[daemonSet].properties[updateStrategy].properties[rollingUpdate].properties[maxSurge].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[envoy].properties[daemonSet].properties[updateStrategy].properties[rollingUpdate].properties[maxUnavailable].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[envoy].properties[deployment].properties[strategy].properties[rollingUpdate].properties[maxSurge].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[envoy].properties[deployment].properties[strategy].properties[rollingUpdate].properties[maxUnavailable].type: Required value: must not be empty for specified object fields]`

Image

### Operating system

Fedora, macOS, Windows

### Installation Method

None

### Version

next (development version)

### Relevant log output

```shell

```

### Additional context

_No response_

gastoner:
### What does this PR do?
Patches kubernetes/client-node package with revert of https://github.com/kubernetes-client/javascript/pull/2400/files

This PR also uses a parser from `yaml` lib, which supports 1.1 and 1.2 version of yaml (kubernetes lib is internally using `js-yaml` which supports only 1.2 version), which represents the octal number (`0` in 1.1 vs `0o` in `1.2`).
This way we are forcing the parsing (Added 1.1 header) of an octal number to a decimal number which is now done correctly by using `yaml` lib

Closes #12747 

### How to test this PR?
Try to create kind cluster


feloy:
basde on this comment https://github.com/kubernetes-client/javascript/issues/2483#issuecomment-2990436313
It may be interesting to test patching the `Objectserialier.ts` file with:

```
- instance[attributeType.baseName] = ObjectSerializer.serialize(data[attributeType.name], attributeType.type, attributeType.format);
+ instance[attributeType.baseName] = ObjectSerializer.serialize(data[attributeType.baseName], attributeType.type, attributeType.format);
```

gastoner:
@feloy updated

feloy:
@gastoner , finally it was not the good direction. I just pushed a PR on the library's repo (https://github.com/kubernetes-client/javascript/pull/2520). If merged, we will need to replace calls to `yaml.parseAllDocuments` with calls to `kubernetesClient.loadAllYaml`

jeffmaury:
> @gastoner , finally it was not the good direction. I just pushed a PR on the library's repo ([kubernetes-client/javascript#2520](https://github.com/kubernetes-client/javascript/pull/2520)). If merged, we will need to replace calls to `yaml.parseAllDocuments` with calls to `kubernetesClient.loadAllYaml`

We need to have this fixed for 1.20 so we should continue work based on this PR as the library may not be updated by next week

jeffmaury (commented):
Does not work for me:

image

gastoner:
@jeffmaury this is the correct output based on the discussion with @feloy

extensions/kind/src/create-cluster.ts
```diff
@@ -61,11 +62,8 @@ function getTags(tags: Tags): Tags {
 }
 
 export async function setupIngressController(clusterName: string): Promise<void> {
-  const manifests = parseAllDocuments(ingressManifests, { customTags: getTags });
-  await extensionApi.kubernetes.createResources(
-    'kind-' + clusterName,
-    manifests.map(manifest => manifest.toJSON()),
-  );
+  const manifests = loadAllYaml(ingressManifests, { customTags: getTags });
```
feloy: you can remove the options with `customTags`, as it is not supported by this new function `loadAllYaml`

feloy:
> @jeffmaury this is the correct output based on the discussion with @feloy

At least the first problem is resolved.

For this problem of invalid value `0644`, the problem is because we cannot handle specific cases like we were doing before with `customTags: getTags` (or we need to find a way to do it).

As a first workaround, I suppose we can change the value `0644` in the manifest with the corresponding decimal value `420`

gastoner:
@jeffmaury If I remember correctly, this commit will fix it without any error https://github.com/podman-desktop/podman-desktop/commit/01fad95df419823356d4387b8b0dad588c7c6f61

jeffmaury:
> @jeffmaury If I remember correctly, this commit will fix it without any error [01fad95](https://github.com/podman-desktop/podman-desktop/commit/01fad95df419823356d4387b8b0dad588c7c6f61)

Still getting an error when I try to create a Kind cluster

jeffmaury (commented):
See https://github.com/podman-desktop/podman-desktop/pull/12919#issuecomment-3017969773

feloy:
see https://github.com/podman-desktop/podman-desktop/pull/12919#issuecomment-3012735736 on how to fix this problem of `invalid value 644`

gastoner:
@feloy @jeffmaury PR Updated!

jeffmaury (commented):
I can now create the cluster with an ingress controller but the YAML file is one from the contour project and should not be updated or we should also update it after the download has been finished (see package.json of the extension)

feloy:
Context: https://github.com/kubernetes-client/javascript/issues/2539 (`0644` used in the contour manifest is a notation compatible with YAML 1.1 only, not 1.2, and the kubernetes library is using a parser compatible with YAML 1.2 only)

As we discussed, the best solution I can think of is to suppose the YAML manifest in entry is in version 1.1 (the one supported by Kubernetes), and convert it in a format compatible with 1.2 (at least), for example:


```
// Add 
// %YAML 1.1 
// ---
// at the beginning of the manifest to parse

const yaml1 = yaml.parse(manifest);
const yml = k8s.loadYaml<k8s.KubernetesObject>(yaml.stringify(yaml1));
```

extensions/kind/src/create-cluster.ts
```diff
@@ -46,24 +45,10 @@ export function getKindClusterConfig(
...
-      }
-    }
-  }
-  return tags;
-}
-
 export async function setupIngressController(clusterName: string): Promise<void> {
-  const manifests = loadAllYaml(ingressManifests, { customTags: getTags });
-  await extensionApi.kubernetes.createResources('kind-' + clusterName, manifests);
+  const parsedManifest = parseAllDocuments(ingressManifests);
```
feloy: let's make the conversion in a separate function (convertYamlFrom11to12(string): string), so we can just remove the call to the function when the issue is fixed (and the intent will be clearer)

gastoner:
@jeffmaury fixed

extensions/kind/src/create-cluster.ts
```diff
@@ -45,13 +45,18 @@ export function getKindClusterConfig(
...
+${manifest}
 `;
+  const parsedManifest = parseAllDocuments(manifest1);
+  return stringify(parsedManifest);
+}
 
+export async function setupIngressController(clusterName: string): Promise<void> {
+  const manifest = convertYamlFrom11to12(ingressManifests);
   const parsedManifest = parseAllDocuments(manifest);
   const yml = loadAllYaml(stringify(parsedManifest));
```
feloy: You are doing the conversion twice:

```
  const manifest = convertYamlFrom11to12(ingressManifests);
  const yml = loadAllYaml(manifest);
```

jeffmaury (changes requested):
Seems broken for me: while the cluster got created and no error reported, I cannot find the contour namespace and no related kubernetes resources

extensions/kind/src/create-cluster.ts
```diff
@@ -45,6 +46,15 @@ export function getKindClusterConfig(
   });
 }
 
+function convertYamlFrom11to12(manifest: string): string {
+  const manifest1 = `
+%YAML 1.1 
+${manifest}
+`;
+  const parsedManifest = parseAllDocuments(manifest1, { customTags: getTags });
+  return stringify(parsedManifest);
```
feloy: `stringify`ing an array of documents returns an array, not documents separated by `---`. You will have to stringify each document separately, then join them with `---`

gastoner: Should be fixed,  but I was forced to remove the header (I was not able to make it work with it, because it was added to each document and then it was causing some issues)

feloy (approved):
tested on Mac, LGTM

thanks for working on this isuue

jeffmaury (approved):
LGTM

Issue #11111, opened 2025-02-12
bug(Table): svelte warning `ownership_invalid_mutation`


axel7083:
### Bug description

While working on https://github.com/podman-desktop/podman-desktop/issues/10763 I ran the tests of `PodsLists.spec` and I noticed the following warnings

```
[svelte] ownership_invalid_mutation
podman-desktop/packages/ui/dist/table/Table.svelte mutated a value owned by src/lib/pod/PodsList.svelte. This is strongly discouraged. Consider passing values to child components with `bind:`, or use a callback instead
```

The Table component has a small issue, it update the `data` props to set the `selected` property on the item. This is useful as the parent component can determine which item is selected.

In our PodsList component we **do not** `bind` the data

https://github.com/podman-desktop/podman-desktop/blob/bb1c3c0fbddcbabd2d5950d2939f6e920bc0d40f/packages/renderer/src/lib/pod/PodsList.svelte#L236

But in the Table component we update it several times

1. we sort it

https://github.com/podman-desktop/podman-desktop/blob/df281f281cd7a4aa533b99405c0f3107dc9ca8fa/packages/ui/src/lib/table/Table.svelte#L116

2. We set the selectect property when toggle all

https://github.com/podman-desktop/podman-desktop/blob/df281f281cd7a4aa533b99405c0f3107dc9ca8fa/packages/ui/src/lib/table/Table.svelte#L58

3. We set indivudual item selected

https://github.com/podman-desktop/podman-desktop/blob/df281f281cd7a4aa533b99405c0f3107dc9ca8fa/packages/ui/src/lib/table/Table.svelte#L164

### Operating system

Linux, Fedora 41

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. Run the `PodsList.spec`

### Relevant log output

```shell

```

MarsKubeX:
Hey @axel7083 It seems that this warning is no longer being shown. I've also checked for warnings while the app was running related to the Table.component and I haven't found any.

PR #12921, opened 2025-06-20
refactor(ui): migrate Table component to Svelte5

### LINKED ISSUE #11586: Migrate UI component table to Svelte 5

axel7083:
### What does this PR do?

TableV5 and Table are not really compatible. 

Pretty complicated PR, the table component has several hard dependency on Svelte4 behavior, which is no longer supported in Svelte5 and create warnings.

E.g. 
- https://github.com/podman-desktop/podman-desktop/issues/11111

---

My solution

1. The data should be readonly, the Table component should never change the props.
> The reason are the children, the parent provide a row, with a children(object: T): Array<T> method, and those field cannot be edited, as it will lead to a svelte warning (https://github.com/podman-desktop/podman-desktop/issues/11111)
3. Instead of having a `selected` & `collapsed` property on the object, we exposed two new props `selected: Set<string>` and `collapsed: Set<string>` holding the values selected & collapsed.
4. The current Table use `object.name`
https://github.com/podman-desktop/podman-desktop/blob/4d352bc99b5bce7a850eeac3a00916d95226a21e/packages/ui/src/lib/table/Table.svelte#L27
This is a big problem as name are not unique, so they cannot be used as unique key to determine if something is selected of collapsed
5. introducing mandatory `key(object: T): string` prop, the parent should provide a method to get a unique key for a given object, this is needed for svelte each key, `selected` and `collapsed` sets.
6. adding optional `label(object: T): string` prop to set the aria-label of the row.

### Screenshot / video of UI

No visual changes

Fixes https://github.com/podman-desktop/podman-desktop/issues/11586

### How to test this PR?

- Not yet

cc @deboer-tim


benoitf (changes requested):
we should not have duplicated items or V5 components

axel7083:
> we should not have duplicated items or V5 components

Updated, to migrate everything

axel7083:
Superseded by https://github.com/podman-desktop/podman-desktop/pull/13718

PR #12933, opened 2025-06-23
refactor(types): adding generics to Table

axel7083:
### What does this PR do?

Removing some `any` from Table component, which improve typing and safety of the Table component. This will ease the migration to Svelte V5.

Part of https://github.com/podman-desktop/podman-desktop/issues/11586

Required for
- https://github.com/podman-desktop/podman-desktop/pull/12921

Need rebase after
- https://github.com/podman-desktop/podman-desktop/pull/12934

### How to test this PR?

- Pipeline should be ✅

### DIFF
--- a/packages/ui/src/lib/table/Table.svelte
+++ b/packages/ui/src/lib/table/Table.svelte
@@ -4,7 +4,7 @@
 }
 </style>
 
-<script lang="ts">
+<script lang="ts" generics="T extends { selected?: boolean; name?: string }">
 /* eslint-disable import/no-duplicates */
 // https://github.com/import-js/eslint-plugin-import/issues/1479
 import { faChevronDown, faChevronRight } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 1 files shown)


packages/ui/src/lib/table/Table.svelte
```diff
@@ -17,10 +17,9 @@ import type { Column, Row } from './table';
 
 export let kind: string;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let columns: Column<any>[];
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export let row: Row<any>;
-export let data: { selected?: boolean; name?: string }[];
+export let columns: Column<T, any>[];
```
axel7083, line 20: I cannot remove this any because of the way we define the Columns. I did my best, but we really allow any type as `RenderType`, it default to `T`, but we can define any value, using `unknown` lead to issue, as unknown is incompatible with `T`, same for `never`.

https://github.com/podman-desktop/podman-desktop/blob/b132d30046cd484d77233a8c69e0bb02fc3ba27b/packages/ui/src/lib/table/table.ts#L90

packages/ui/src/lib/table/Table.svelte
```diff
@@ -43,9 +42,16 @@ let selectedAllCheckboxes = false;
...
-      .map(object => row.info.children?.(object))
-      .flat()
+      .reduce(
+        (accumulator, current) => {
+          const children = row.info.children?.(current);
+          if (children) {
+            accumulator.push(...children);
+          }
+          return accumulator;
+        },
```
benoitf: I don't see how it's related to generics

axel7083: It was typed to any, and a typecheck error was raised, I can do that in a separate PR, if you think this is preferable

axel7083: Resolved by https://github.com/podman-desktop/podman-desktop/pull/12967

benoitf (approved):
👍

Issue #10462, opened 2024-12-21
Podman Desktop not checking VirtualMachinePlatform state properly


jski:
### Bug description

I have WSL2 installed, and an Ubuntu distribution running. The WMI check for the VirtualMachinePlatform shows it's enabled, and no matter how many times I try to reenable and restart my machine, no joy.

```
(Get-WmiObject -Query "Select * from Win32_OptionalFeature where InstallState = '1'").Name | select-string VirtualMachinePlatform

VirtualMachinePlatform
```

### Operating system

Windows 11 Pro - x64

### Installation Method

Winget (Windows)

### Version

1.14.1

### Steps to reproduce

Install Podman desktop with winget
Go through guided setup for installing Podman

### Relevant log output

```shell

```

### Additional context

**Enabled Extensions**
- podman-desktop.compose
- podman-desktop.docker
- podman-desktop.kind
- podman-desktop.kube-context
- podman-desktop.kubectl-cli
- podman-desktop.lima
- podman-desktop.podman
- podman-desktop.registries

jski:
Note: after downloading the standalone Podman installer, I could install version 5.3.1 with no issues. Afterwards, Podman Desktop detected the install and went about setting up a machine no problem. So hopefully that can help you track down the issue in the desktop installer compatibility checking.

Also, fwiw the autogenerated bug report above says I'm on v1.14.1, but the Podman Desktop app itself reports v1.15.0.

ScrewTSW:
@jski Hello. I've tested a clean install over winget on Windows 10, installed 1.15.0, installed Podman 5.3.1 via onboarding, no issues there, created a machine, ssh-d into it and tested engine version. everything seems to be working as intended.

We have not verified the bug on Windows 11 yet. Could someone from devs have a look possibly?

phillies:
Can confirm on Win11 with Podman Desktop 1.15.0, after installation the setup podman wizard failed with Virtual Machine Platform not enabled. Manually downloading and installing Podman 5.3.2 and restarting Podman Desktop works.

jeffmaury:
@jski @phillies I tried to reproduce on my Win11Pro laptop with no luck. Can you give the output of:

(Get-WmiObject -Query "Select * from Win32_OptionalFeature where InstallState = '1'").Name

phillies:
> [@jski](https://github.com/jski) [@phillies](https://github.com/phillies) I tried to reproduce on my Win11Pro laptop with no luck. Can you give the output of:
[...]

PR #12937, opened 2025-06-23
fix: send virtual machine check info to telemetry in case of error

jeffmaury:
Relates to #10462

### What does this PR do?

Send info to telemetry when virtual machine check is negative or errored

### Screenshot / video of UI

N/A

#10462

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/utils/powershell.ts
+++ b/extensions/podman/packages/extension/src/utils/powershell.ts
@@ -63,7 +63,7 @@ class PowerShell5Client implements PowerShellClient {
       telemetry['killed'] = execError.killed;
     }
     const telemetryLogger = getTelemetryLogger();
-    telemetryLogger?.logError('check.isVirtualMachineAvailable', telemetry);
+    telemetryLogger?.logUsage('check.isVirtualMachineAvailable', telemetry);
     return false;
   }
 
--- a/extensions/podman/packages/extension/src/utils/powershell.ts
+++ b/extensions/podman/packages/extension/src/utils/powershell.ts
@@ -51,6 +51,14 @@ class PowerShell5Client implements PowerShellClient {
     } catch (err) {
       // ignore error, this means that VirtualMachinePlatform not enabled
       telemetry['error'] = err;
+      const execError = err as extensionApi.RunError;
+      telemetry['message'] = execError.message;
+      telemetry['exitCode'] = execError.exitCode;
+      telemetry['command'] = execError.command;
+      telemetry['stdout'] = execError.stdout;
+      telemetry['stderr'] = execError.stderr;
+      telemetry['cancelled'] = execError.cancelled;
+      telemetry['killed'] = execError.killed;
     }
     const telemetryLogger = extensionApi.env.createTelemetryLogger();
     telemetryLogger.logError('check.isVirtualMachineAvailable', telemetry);
--- a/extensions/podman/packages/extension/src/utils/powershell.ts
+++ b/extensions/podman/packages/extension/src/utils/powershell.ts
@@ -17,8 +17,8 @@
  ***********************************************************************/
 import * as extensionApi from '@podman-desktop/api';
 
-export async function getPowerShellClient(): Promise<PowerShellClient> {
-  return new PowerShell5Client();
+export async function getPowerShellClient(telemetryLogger: extensionApi.TelemetryLogger): Promise<PowerShellClient> {
+  return new PowerShell5Client(telemetryLogger);
 }
 
 export interface PowerShellClient {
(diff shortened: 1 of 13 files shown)


extensions/podman/packages/extension/src/utils/powershell.ts
```diff
@@ -32,21 +32,28 @@ export interface PowerShellClient {
...
-      if (res.indexOf('VirtualMachinePlatform') >= 0) {
+      telemetry['command'] = res.command;
+      telemetry['stdout'] = res.stdout;
+      telemetry['stderr'] = res.stderr;
+      if (res.stdout.indexOf('VirtualMachinePlatform') >= 0) {
         return true;
       }
     } catch (err) {
       // ignore error, this means that VirtualMachinePlatform not enabled
+      telemetry['error'] = err;
```
axel7083, line 55: When the process exit with non-zero exit code, the exception raised is `ResultError` it has some field that would be nice to extract like `exitCode`, `stderr` and `stdin`

https://github.com/podman-desktop/podman-desktop/blob/1a08b7ba22a9a29bf6963ad66c2efa9de0c8e0b9/packages/extension-api/src/extension-api.d.ts#L4407

extensions/podman/packages/extension/src/utils/powershell.ts
```diff
@@ -32,21 +32,28 @@ export interface PowerShellClient {
...
     }
+    const telemetryLogger = extensionApi.env.createTelemetryLogger();
+    telemetryLogger.logError('check.isVirtualMachineAvailable', telemetry);
```
axel7083: question: I think we use always `logInfo` even when logging error, as this makes two different event in amplitude cc @benoitf for comfirmation

extensions/podman/packages/extension/src/utils/powershell.ts
```diff
@@ -32,21 +32,28 @@ export interface PowerShellClient {
...
+      telemetry['stdout'] = res.stdout;
+      telemetry['stderr'] = res.stderr;
+      if (res.stdout.indexOf('VirtualMachinePlatform') >= 0) {
         return true;
       }
     } catch (err) {
       // ignore error, this means that VirtualMachinePlatform not enabled
+      telemetry['error'] = err;
     }
+    const telemetryLogger = extensionApi.env.createTelemetryLogger();
```
axel7083: Do we really need to re-create a telemetry logger? we already created one, should be nice to reuse it

https://github.com/podman-desktop/podman-desktop/blob/6e8fc0a385ab359ea187870a82783551714f9deb/extensions/podman/packages/extension/src/extension.ts#L1142

jeffmaury: Fixed by dfd65b9ac4fbd02862b78fa3e8dbfbc37cdd962f

extensions/podman/packages/extension/src/utils/powershell.ts
jeffmaury: Fixed by f914bd14e8b7e5c1a27e25c276585dd4215c4380

extensions/podman/packages/extension/src/utils/powershell.ts
```diff
@@ -17,6 +17,8 @@
  ***********************************************************************/
 import * as extensionApi from '@podman-desktop/api';
 
+import { getTelemetryLogger } from '../extension';
```
benoitf: I think having this kind of dependency is an anti-pattern

sub-code should not import something from the extension

extension is importing the sub-code

extensions/podman/packages/extension/src/utils/powershell.ts
```diff
@@ -31,22 +31,38 @@ export interface PowerShellClient {
...
+      telemetry['command'] = res.command;
+      telemetry['stdout'] = res.stdout;
+      telemetry['stderr'] = res.stderr;
+      if (res.stdout.indexOf('VirtualMachinePlatform') >= 0) {
         return true;
       }
     } catch (err) {
       // ignore error, this means that VirtualMachinePlatform not enabled
+      telemetry['error'] = err;
+      const execError = err as extensionApi.RunError;
```
axel7083, line 56: We should probably not assume this is a RunError, without checking it

Check `isRunError` in [podman-desktop/extension-podman-quadlet#packages/backend/src/utils/run-error.ts](https://github.com/podman-desktop/extension-podman-quadlet/blob/44db38b2f8a8ec56fd9e619d94d025c53c2d0e7e/packages/backend/src/utils/run-error.ts#L20)

axel7083 (approved):
LGTM, nit pick the error casting to RunError

extensions/podman/packages/extension/src/utils/powershell.ts
SoniaSandler: +1

SoniaSandler (approved):

PR #12939, opened 2025-06-23
chore: update when field for podman machine editable properties

### LINKED ISSUE #12943: Add Podman machine editing option to HyperV machines

Right now, editing an existing Podman machine is possible only on mac since "Podman WSL is nop and generates errors". 
However, it came up in a discussion related to https://github.com/podman-desktop/podman-desktop/pull/12781 that HyperV machines should also support some editing options (see https://github.com/podman-desktop/podman-desktop/pull/12939#issuecomment-2996918502).

Expose Podman machine CPU and memory editing in HyperV machines

_No response_

### Additional context

_No response_

SoniaSandler:

### What does this PR do?
Right now, editing a podman machine is only possible on macOS. However, as part of https://github.com/podman-desktop/podman-desktop/pull/12781, the rootful status is also editable on both macOS and Windows, so we need to indicate which properties are editable only on macOS.
On HyperV, CPU and memory should be editable, and on macOS, all 3 properties are editable

Related to https://github.com/podman-desktop/podman-desktop/pull/12781
Closes https://github.com/podman-desktop/podman-desktop/issues/12943

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -879,7 +879,7 @@ export async function registerProviderFor(
       });
     },
   };
-  //support edit only on MacOS as Podman WSL is nop and generates errors
+  //support edit only on MacOS and HyperV machines as Podman WSL is nop and generates errors
   if (isEditMemorySupported || isEditCPUSupported || isEditDiskSizeSupported) {
     lifecycle.edit = async (context, params, logger, _token): Promise<void> => {
       let effective = false;
--- a/extensions/podman/packages/extension/src/extension.ts
+++ b/extensions/podman/packages/extension/src/extension.ts
@@ -857,9 +857,9 @@ export async function registerProviderFor(
   machineInfo: MachineInfo,
   socketPath: string,
 ): Promise<void> {
-  const hyperVEnabled = await isHyperVEnabled();
-  const isEditMemorySupported = extensionApi.env.isMac || hyperVEnabled;
-  const isEditCPUSupported = extensionApi.env.isMac || hyperVEnabled;
+  const isHyperVMachine = extensionApi.env.isWindows && machineInfo.vmType === VMTYPE.HYPERV;
+  const isEditMemorySupported = extensionApi.env.isMac || isHyperVMachine;
+  const isEditCPUSupported = extensionApi.env.isMac || isHyperVMachine;
   const isEditDiskSizeSupported = extensionApi.env.isMac;
 
   extensionApi.context.setValue(PODMAN_MACHINE_EDIT_MEMORY, isEditMemorySupported);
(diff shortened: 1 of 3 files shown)


axel7083 (approved):
Curious to know if this is really only working on Mac, or would work for windows HyperV (I know it is not working for WSL, but not sure abotu hyperV)

cc @jeffmaury

jeffmaury:
> Curious to know if this is really only working on Mac, or would work for windows HyperV (I know it is not working for WSL, but not sure abotu hyperV)
> 
> cc @jeffmaury

CPU and memory should be exposed for HyperV

SoniaSandler:
@axel7083 @jeffmaury right now the podman machine editing option is only registered for macOS

jeffmaury:
> @axel7083 @jeffmaury right now the podman machine editing option is only registered for macOS

So we should create a follow up issue as I would guess HyperV support was added after edit and that was missed

benoitf (commented):
I would not add isMac as we know it's a limited factor

for example we can have a property dedicated set in the code

let say for example `podman.machine.editCpu`

and then in the code if macOS we set this property to true (or if Hyper-V)
it'll avoid to change the properties in package.json later

also it's easier to have the logic in the code rather in the package.json as it's easier to provide unit tests

SoniaSandler:
@axel7083 @cdrage @feloy Could you PTAL at this PR? thanks!

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -871,7 +880,7 @@ export async function registerProviderFor(
     },
   };
   //support edit only on MacOS as Podman WSL is nop and generates errors
-  if (extensionApi.env.isMac) {
+  if (extensionApi.env.isMac || hyperVEnabled) {
```
feloy: What is the relation between this test and the flags above (isEditMemorySupported, isEditCPUSupported, isEditDiskSizeSupported)? If it's an OR on all of them, I think it would be better to explicitely use this OR instead, so we can change the flags without having to change here also

SoniaSandler: @feloy I'm not sure I understand your suggestion. Do you mean moving the set isEditMemorySupported, isEditCPUSupported, isEditDiskSizeSupported inside this if? or using them (with OR) as the if statement?

feloy: If this is the correct test, I suggest to change with

```
if (isEditMemorySupported || isEditCPUSupported || isEditDiskSizeSupported)
```

so we can change later how these three conditions are computed, without having to change here

feloy (approved):
lgtm code-wise

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -871,18 +880,18 @@ export async function registerProviderFor(
     },
   };
   //support edit only on MacOS as Podman WSL is nop and generates errors
```
jeffmaury: Maybe this comment should be updated as well

jeffmaury (approved):
LGTM

PR #12946, opened 2025-06-24
fix: reduce github rate limit errors for compose install

dgolovin:
### What does this PR do?

The fix using @memoize decorator to cache github responses eliminating most of rate limit related errors.
The only drawback of this approach is when check for update is implemented one would see new release with a delay.

### Screenshot / video of UI

https://github.com/user-attachments/assets/e425633c-5511-4ecf-a4ac-c9ce2134ee17

### How to test this PR?

1. Pull the branch and run it with `pnpm watch`. 
2. Go to CLI Tools settings and try request docker-compose install
3. After seeing the quick picks of compose versions cancel it and repeat step 2

With branch there should be no rate limit errors, because result is cached for an hour (could have define in preference settings and could be shorter of course). 

### DIFF
--- a/extensions/compose/package.json
+++ b/extensions/compose/package.json
@@ -169,7 +169,8 @@
   "dependencies": {
     "@octokit/rest": "^22.0.0",
     "mustache": "^4.2.0",
-    "shell-path": "^3.0.0"
+    "shell-path": "^3.0.0",
+    "typescript-memoize": "^1.1.1"
   },
   "devDependencies": {
     "@podman-desktop/api": "workspace:*",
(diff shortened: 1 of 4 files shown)


jeffmaury (approved):
LGTM

benoitf (changes requested):
1. this dependency is unmaintained 

body is speaking about "memoize" but package.json imports "typescript-memoize" which is not updated since 2022 while memoize was updated this year

2. AFAIK we should never introduce a cache without a way to clear it (could be a command)

3. configuration of the timeout should be a property

also I'm unsure on the overall benefit

in an enterprise environment (sharing IP, I guess the ip is already reaching the limit, so it won't help)
and on a default environment, I'm not sure we reach the limit ?

anyway I'm not against the caching

dgolovin:
Overall benefit - less rate limit related errors. I am just going to use it locally while implementing CLI tools related features.
Closing.

PR #12953, opened 2025-06-24
chore: allow to remove extension in dev mode without error

benoitf:
### What does this PR do?
in development mode we can remove the extension but we don't remove the folder

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extension/extension-loader.ts
+++ b/packages/main/src/plugin/extension/extension-loader.ts
@@ -1773,7 +1773,7 @@ export class ExtensionLoader {
       // delete the path
       if (extension.removable) {
         await fs.promises.rm(extension.path, { recursive: true, force: true });
-      } else {
+      } else if (!extension.devMode) {
         throw new Error(`Extension ${extensionId} is not removable`);
       }
       this.analyzedExtensions.delete(extensionId);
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/extension/extension-loader.ts
```diff
@@ -1773,7 +1773,7 @@ export class ExtensionLoader {
       // delete the path
       if (extension.removable) {
         await fs.promises.rm(extension.path, { recursive: true, force: true });
-      } else {
+      } else if (!extension.devMode) {
         throw new Error(`Extension ${extensionId} is not removable`);
```
axel7083, line 1777: remark: we lose the fallback here ? shouldn't we have instead `if (extension.removable && !extension.devMode)` at L1774 ?

What are the condition where `removable` is false ? because maybe removable should be false when we are in `devMode`

axel7083 (approved):

benoitf:
>remark: we lose the fallback here ? shouldn't we have instead if (extension.removable && !extension.devMode) at >L1774 ?

>What are the condition where removable is false ? because maybe removable should be false when we are in >devMode

@axel7083 I didn't see the message earlier

so removable is only flagged from extensions inside the local copy (b/c built-in extensions or in dev extensions should not be removed from the filesystem)

there is no need of the fallback

PR #12979, opened 2025-06-26
feat(table): adding accessibility labels to collapse button

axel7083:
### What does this PR do?

To be able to migrate the Table component to Svelte v5 I wanted to properly cover all the table feature to ensure no regression. The collapse action does not seems to be covered by tests.

To be able to easily test it, I need to add some recommended attributes, to properly test it. Such as [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded).

### Screenshot / video of UI

image
image

Required for
- https://github.com/podman-desktop/podman-desktop/pull/12921

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/table/Table.spec.ts
+++ b/packages/ui/src/lib/table/Table.spec.ts
@@ -445,11 +445,11 @@ describe('Table#collapsed', () => {
     });
 
     const fooRow = getByRole('row', { name: 'foo' });
-    const fooExpandBtn = within(fooRow).getByRole('button', { name: 'Expand' });
+    const fooExpandBtn = within(fooRow).getByRole('button', { name: 'Expand Row' });
     expect(fooExpandBtn).toHaveAttribute('aria-expanded', 'false');
 
     const barRow = getByRole('row', { name: 'bar' });
-    const barCollapseBtn = within(barRow).getByRole('button', { name: 'Collapse' });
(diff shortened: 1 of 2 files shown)


jeffmaury (commented):
Not sure why title is not enough to test the state

axel7083:
> Not sure why title is not enough to test the state

@jeffmaury I need the title to getByRole the button, and I need the `area-expanded` to determine if the row is collapsed or not

vancura:
I think it would be better to expand the wording a bit (sorry about the pun):
- Collapse row
- Expand row

Thoughts?

axel7083:
> I think it would be better to expand the wording a bit (sorry about the pun):
> 
> * Collapse row
> * Expand row
> 
> Thoughts?

Not much opinion about it, if you think this is better, I can update

vancura:
It's better to explain the target of the action, not just the action itself. The truth is this is a matter of personal taste. We need somebody to take care of wording, but that's not me :)

MarsKubeX (approved):
LGTM and I think it is a good idea to have the `aria-expanded` for accessibility

jeffmaury:
As the title reflects the state I think the other property is extra but I'm OK

jeffmaury (approved):
LGTM

PR #12991, opened 2025-06-27
chore(test): navigation takes longer on cicd

cbr7:
### What does this PR do?
In the overnight test runs some tests failed due to default timeout expiring while test was executing. This PR increases some timesouts to allow for a slower cicd environment.

### DIFF
--- a/tests/playwright/src/specs/image-push-to-registry-smoke.spec.ts
+++ b/tests/playwright/src/specs/image-push-to-registry-smoke.spec.ts
@@ -75,7 +75,7 @@ test.describe.serial('Push image to container registry', { tag: '@smoke' }, () =
     await playExpect(imagesPage.heading).toBeVisible();
 
     await playExpect
-      .poll(async () => await imagesPage.waitForRowToExists(helloContainer, 15_000), { timeout: 0 })
+      .poll(async () => imagesPage.waitForRowToExists(helloContainer, 30_000), { timeout: 0 })
       .toBeTruthy();
   });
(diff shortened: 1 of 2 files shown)


tests/playwright/src/specs/image-push-to-registry-smoke.spec.ts
```diff
@@ -75,7 +75,7 @@ test.describe.serial('Push image to container registry', { tag: '@smoke' }, () =
     await playExpect(imagesPage.heading).toBeVisible();
 
     await playExpect
-      .poll(async () => await imagesPage.waitForRowToExists(helloContainer, 15_000), { timeout: 0 })
+      .poll(async () => imagesPage.waitForRowToExists(helloContainer, 30_000), { timeout: 0 })
```
odockal, line 78: I do not think you should remove await here.

odockal (dismissed):

tests/playwright/src/specs/image-push-to-registry-smoke.spec.ts
cbr7: why? removing the await is the technically correct way of writing this, though in reality they are equivalent. The `await` keyword is redundant in that location.

benoitf: as I understand, poll can return a Promise so no need to await in the function, the poll function should wait for it

benoitf (approved):

PR #12992, opened 2025-06-27
chore: move disposable group to API package

benoitf:
### What does this PR do?
for inversify, we bind the type so it'll change some imports from the type to the object
as renderer files are importing some files, changing the import from 'type' scope to default scope bring Node.js dependencies to the renderer side and make some loading failures

the idea is to move all interfaces imported (or transitively imported from the rendered side to the API package

### How to test this PR?


axel7083 (approved):

MarsKubeX (approved):
LGTM

PR #12994, opened 2025-06-27
feat(table): adding key props

axel7083:
### What does this PR do?

Today to determine if a given group of containers (compose, pods) we use the `object.name` property as a key. But this has some issue, for example as explained in https://github.com/podman-desktop/podman-desktop/pull/12915#issuecomment-2988601245 the name can be shared (we can have a pod on one podman machine named the same as another pod in another podman machine.)

Therefore, we should be able to provide a custom `key` function, taking the object and returning a unique string, that will be used as unique key, to determine if something should be collapsed or not.

### Screenshot / video of UI

N/A

Required for https://github.com/podman-desktop/podman-desktop/issues/12842

### How to test this PR?

### DIFF
--- a/packages/ui/src/lib/table/Table.spec.ts
+++ b/packages/ui/src/lib/table/Table.spec.ts
@@ -410,6 +410,7 @@ test('Expect table is scoped for css manipulation', async () => {
 
 describe('Table#collapsed', () => {
   interface Item {
+    id: string;
     name?: string;
   }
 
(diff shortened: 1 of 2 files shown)


benoitf (approved):

benoitf (changes requested):
~~looks like the link is `https://www.youtube.com/playlist?list=PLwQSOOPzuslql836gEAqF66FAYLm5sno-`~~ wrong PR

benoitf (approved):

Issue #12640, opened 2025-05-27
introduce dependency injection in the main side


benoitf:
### Is your enhancement related to a problem? Please describe

we have all xxxxRegistry objects in the main part to handle all requests

the idea is to bring dependency injection framework, and then replace all new xxxxRegistry(registry1, registry2, etc)

this is a step 1: use constructor dependency injection, so small changes are introduced (annotations)

after a while we could move to property injection that avoid to change constructor parameters each time but it's not the goal of this first issue.

also, it should not try to replace all the instances at once, but we introduce it and see how it goes.

### Describe the solution you'd like

Use for example dependency injection for ExtensionLoader or few other objects requiring a lot of injection.

use `inversify` as the DI framework.

PR #12996, opened 2025-06-27
refactor: move event to API side

benoitf:
### What does this PR do?
avoid to import stuff from renderer to main package

(required for inversify)
#12640 

### How to test this PR?

should be 💚 

### DIFF
--- a/packages/api/src/event.ts
+++ b/packages/api/src/event.ts
@@ -22,7 +22,7 @@ import type { DisposableGroup } from './disposable-group.js';
 /**
  * Represents a typed event.
  *
- * @param listener The listener function will be call when the event happens.
+ * @param listener The listener function will be called when the event happens.
  * @param thisArgs The 'this' which will be used when calling the event listener.
  * @param disposables An array to which a {{IDisposable}} will be added.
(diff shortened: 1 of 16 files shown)


packages/api/src/event.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Represents a typed event.
+ *
+ * @param listener The listener function will be called when the event happens.
+ * @param thisArgs The 'this' which will be used when calling the event listener.
+ * @param disposables An array to which a {{IDisposable}} will be added.
+ * @return a disposable to remove the listener again.
+ */
+export type Event<T> = (listener: (e: T) => unknown, thisArgs?: unknown, disposables?: DisposableGroup) => IDisposable;
```
axel7083, line 30: Since this is exposed to both renderer & main, maybe we should rename it to differentiate from existing [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event)

axel7083: (follow up issue)

axel7083 (approved):

PR #13008, opened 2025-06-28
chore: wrap extensions in navigation bar into scrollable section

### LINKED ISSUE #12234: Can't reach bottom sidebar items without expanding app (no scrollbar for sidebar)
### Bug description

After installing multiple extensions, the number of sidebar items increases, and the bottom buttons (Accounts, Settings) are unreachable without expanding the application. The mouse wheel for scrolling doesn't work, and there's no scrollbar either.

### Paris F2F

The side taskbar becomes filled when too many extensions are active, which can cause key elements like the settings icon to disappear.

### Operating system

Windows 11

### Installation Method

### Version

1.18.0

### Additional context

Image

https://github.com/user-attachments/assets/9bb359df-2a58-4bf0-9999-212869df23fe

vzhukovs:
### What does this PR do?

Based on the comments from https://github.com/podman-desktop/podman-desktop/issues/6774#issuecomment-2984024027:

> I am not sure scrolling would help. From top of my head, there are two options we have:
> 
> either we collapse icons to be vertically smaller when there's no space,
> or bundle icons together in groups (similarly to what Discord does in the sidebar).
> I am afraid scrolling won't help much, as it would always make some icons offscreen. Well, that's what would happen with 2. as well.
> 
> Collapsing (making the icons smaller) won't help on the long run either, we'd just delay the problem again, we'd just buy some time (by buying space).

there’s a proposal to introduce an <ins>interim solution</ins> that relies on showing a scrollable section while we discuss the best way to display multiple extensions in the navigation area. On the next step, the navigation bar will be refactored with the upcoming requirements to display more extensions that it can fit.

Be sure that PR: https://github.com/podman-desktop/podman-desktop/pull/13009 is merged before this PR.

### Screenshot / video of UI

scrollable_navbar

#12234 

### How to test this PR?

Combine the changes from the current PR with the changes from this PR: [link to PR]. Install more than three extensions, then reduce the window height and verify that the Accounts and Settings buttons remain fixed in place while the list of installed extensions becomes vertically scrollable.

### DIFF
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -37,29 +37,48 @@ onDidChangeConfiguration.addEventListener(NAV_BAR_LAYOUT, onDidChangeConfigurati
 
 let minNavbarWidth = $derived(iconWithTitle ? 'min-w-fit' : 'min-w-leftnavbar');
 
+let scrollEl: HTMLDivElement;
+let resizeObserver: ResizeObserver;
+let showTopShadow = $state(false);
+let showBottomShadow = $state(false);
+
+let thumbTop = $state(0);
+let thumbHeight = $state(20);
+let hasScroll = $state(false);
+
 onMount(async () => {
   const commandRegistry = new CommandRegistry();
   commandRegistry.init();
   iconWithTitle = (await window.getConfigurationValue(NAV_BAR_LAYOUT)) === AppearanceSettings.IconAndTitle;
   await tick();
-  handleScroll();
+  resizeObserver = new ResizeObserver(handleScroll);
+  if (scrollEl) resizeObserver.observe(scrollEl);
 });
 
 onDestroy(() => {
   onDidChangeConfiguration.removeEventListener(NAV_BAR_LAYOUT, onDidChangeConfigurationCallback);
+  resizeObserver?.disconnect();
 });
 
-let scrollEl: HTMLDivElement;
-let showTopShadow = $state(false);
-let showBottomShadow = $state(false);
-
 function handleScroll(): void {
-  const scrollTop = scrollEl.scrollTop;
-  const scrollHeight = scrollEl.scrollHeight;
-  const clientHeight = scrollEl.clientHeight;
+  const { scrollTop, scrollHeight, clientHeight } = scrollEl;
+
+  hasScroll = scrollHeight > clientHeight + 1;
+
+  if (!hasScroll) {
+    thumbTop = 0;
+    thumbHeight = 0;
+    showTopShadow = false;
+    showBottomShadow = false;
+    return;
+  }
 
   showTopShadow = scrollTop > 5;
   showBottomShadow = scrollTop + clientHeight < scrollHeight - 5;
+
+  const ratio = clientHeight / scrollHeight;
+  thumbHeight = Math.max(clientHeight * ratio, 18);
+  thumbTop = scrollTop * ratio;
 }
 
 function handleClick(): void {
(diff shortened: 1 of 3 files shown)


MariaLeonova:
@vzhukovs thank you, this looks amazing!

vancura:
This looks great, but there's a problem with showing the user that they can actually scroll. From what I see, there's no way to figure out there's something behind the edge of the scroll area without scrolling first. There's also no scrollbar, making it impossible to see the scrolling position, and potentially the length of the scroll area (based on the ratio of background vs. thumb height).

But there would be an easy solution: even just a thin line (but always visible) would fix both issues. We can't rely on "native" scrollbars (in terms of Chromium-native) to work for us, as Windows has a huge scrollbar, so we need a custom solution.

Could you please try to implement a simple line that shows the position and length of the complete scroll? Something like this:

image

Mind that the thumb height should be based on the ratio of the whole scroll area height and the currently visible viewport. There must be a minimum thumb height (approximately 44pt). If the viewport is fully visible, hide the thumb (and disable scrolling, which you already do).

Worth considering: a subtle elastic scroll effect when scrolling is enabled.

Thoughts?

vancura:
Also, the thumb area focus area needs to be slightly larger, so it's easy to click (and drag).

image

There seems to be no way to control the scroll position with just the keyboard (some people have no pointing device or can't use one), but that's a stretch goal now. The whole application is somewhat unusable without a mouse or trackpad, so these people can't use it anyway. We need to address this issue in general, but it's outside the scope of this task.

vancura (changes requested):
Would it be possible to implement the changes I proposed, please?

vzhukovs:
> Would it be possible to implement the changes I proposed, please?

I think, yes. Working on the enhancement, that you proposed (showing thin scrollbar).

gastoner (commented):
On mac I'm able to scroll to right/left and all the extensions will "move out of the window" 


https://github.com/user-attachments/assets/8a362b8f-7dc2-441a-860c-80032660b544



Also when I have just minimal height of the PD I cant really see anything

image

Maybe when having height for the extensions less than height of the extension -> add to scrollable the `Extension` option too?

vzhukovs:
> Would it be possible to implement the changes I proposed, please?

I've prepared a new update to the PR, which adds a tiny scroll for visual representation.

scroll_area_modified

vzhukovs:
> On mac I'm able to scroll to right/left and all the extensions will "move out of the window"
> 
>  Screen.Recording.2025-07-03.at.17.03.18.mov 
> 
> Also when I have just minimal height of the PD I cant really see anything
> 
> image
> Maybe when having height for the extensions less than height of the extension -> add to scrollable the `Extension` option too?

Horizontal scroll appears because changes from https://github.com/podman-desktop/podman-desktop/pull/13009 is not included, to fully test the current PR it is better to checkout to https://github.com/podman-desktop/podman-desktop/pull/13009 and apply patch from the current PR: https://patch-diff.githubusercontent.com/raw/podman-desktop/podman-desktop/pull/13008.patch

vzhukovs:
Side note: fixing related unit tests, will be included in separate commit.

vancura (approved):

packages/renderer/src/AppNavigation.svelte
```diff
@@ -37,16 +37,50 @@ onDidChangeConfiguration.addEventListener(NAV_BAR_LAYOUT, onDidChangeConfigurati
...
+
+let thumbTop = $state(0);
+let thumbHeight = $state(20);
+let hasScroll = $state(false);
+
 onMount(async () => {
   const commandRegistry = new CommandRegistry();
   commandRegistry.init();
   iconWithTitle = (await window.getConfigurationValue(NAV_BAR_LAYOUT)) === AppearanceSettings.IconAndTitle;
+  await tick();
```
benoitf, line 53: there should be no tick in the code (ok in tests but not in the code)

packages/renderer/src/AppNavigation.svelte
```diff
@@ -84,21 +118,47 @@ function onDidChangeConfigurationCallback(e: Event): void {
...
     </div>
   </NavItem>
+
   {#each $navigationRegistry as navigationRegistryItem, index (index)}
     {#if navigationRegistryItem.items && navigationRegistryItem.type === 'group'}
       <!-- This is a group, list all items from the entry -->
-      {#each navigationRegistryItem.items as item, index (index)}
-        <NavRegistryEntry entry={item} bind:meta={meta} iconWithTitle={iconWithTitle} />
-      {/each}
+      <div class="relative min-h-0 flex-1">
```
benoitf, line 125: too much code here, there should be a wrapper

packages/renderer/src/AppNavigation.svelte
```diff
@@ -84,21 +118,47 @@ function onDidChangeConfigurationCallback(e: Event): void {
...
     {:else if navigationRegistryItem.type === 'entry' || navigationRegistryItem.type === 'submenu'}
       <NavRegistryEntry entry={navigationRegistryItem} bind:meta={meta} iconWithTitle={iconWithTitle} />
     {/if}
   {/each}
 
-  <div class="grow"></div>
-
   <div bind:this={outsideWindow}>
-    <NavItem href="/accounts" tooltip="" bind:meta={meta} onClick={(event): void => authActions?.onButtonClick(event)}>
+    <NavItem href="/accounts" tooltip="" bind:meta={meta} onClick={authActions?.onButtonClick}>
```
benoitf, line 161: this change is unrelated ?

benoitf (changes requested):

benoitf:
closing as not updated for a while

MariaLeonova:
@benoitf I think this is still needed?

benoitf:
@MariaLeonova issue is still open, but a PR without activity in 5 months, I'm closing it to reduce the number of PR being active/in review

vancura:
Just so you know, I've been working on the new navbar proto, where I implement this too (and other features). It's almost done, so we'll then see how to fix our navbar in phases. Stay tuned.

Issue #10850, opened 2025-01-27
Context Menu opened over container row is partially hidden when opened to the top of the Containers Page


odockal:
### Bug description

![Image](https://github.com/user-attachments/assets/2a70dc70-3404-4941-94a1-8585f93ba826)

### Operating system

Linux

### Installation Method

Other

### Version

next (development version)

### Relevant log output

```shell

```

odockal:
Looks like a regression of https://github.com/podman-desktop/podman-desktop/issues/7321.

odockal:
Interestingly I cannot reproduce on Mac OS.

axel7083:
Following some discussion today 

## Problem

Narrowing down the issue: the table component is using overflow scroll, a _it need to be scrollable_, however, the parent component, and many over them are also using some kind of overflow hidden, making the task difficult. 

## Generic solution

One solution to overcome this overflow problem would be to create a dedicated component managing **all context menus** putting it on the root directory, absolute, with index-z very high

This solution would have some nice advantages: consistency between all podman-desktop, as we would have a method to call to display it, like `window.showContextMenu(...)`, and it would be nice and appear where we need.

Big drawback: changing all code everywhere :( we would need to update all our current context menus, to use it.

benoitf:
isn't it easier to use `fixed` position rather than `absolute` ?

I mean today it looks like the context menu is moving when we scroll the table and AFAIK I'm not sure this is the expected behaviour.

deboer-tim:
IIRC this is at least the third time we've had this bug open. I can't find the first, but the second was #7321.

I would like to improve the table/nav page overflow-hidden path, but ultimately I think the solution outlined above is the only one that is guaranteed to work in all situations: essentially handling context menus the same way that we do the command palette or message boxes, where the actual component is higher up in the stack.

Related to this is #8336 and the fact that our native menus can extend _outside_ of the Podman Desktop window, but our current styled menus get cut off if they're too big. Whether or not we do anything about #8336 IMHO we should fix this at the same time (i.e. use separate window, not just z-order).

axel7083:
I can't find a way to fix without changing the full implementation => closing for now as discussed with other people - if anyone interested, please re-open

PR #13009, opened 2025-06-28
chore: refactor tooltip component to avoid render component inside container viewport

### LINKED ISSUE #12234: Can't reach bottom sidebar items without expanding app (no scrollbar for sidebar)

vzhukovs:
### What does this PR do?

Side PR  needed for https://github.com/podman-desktop/podman-desktop/pull/13008
The current PR need to be merged before https://github.com/podman-desktop/podman-desktop/pull/13008

A situation may arise where the component that needs a tooltip is inside a scrollable area or an element with overflow settings. In this case, the tooltip would be hidden from the user. To prevent this, we should render the tooltip as a separate component at the root of the HTML document and display it above all other elements using absolute positioning.

Our current behavior for displaying the tooltip:

With the new changes, the tooltip should appear above all UI elements.

Part of #12234 

### How to test this PR?

Ensure that behaviour of rendering tooltips is not changed on the UI.

### DIFF
--- a/packages/ui/src/lib/tooltip/TestTooltipSlotWrapper.svelte
+++ b/packages/ui/src/lib/tooltip/TestTooltipSlotWrapper.svelte
@@ -1,0 +1,8 @@
+<script>
+import Tooltip from './Tooltip.svelte';
+</script>
+
+<Tooltip>
+  <span slot="tip">slot-tip-content</span>
+  <button data-testid="trigger">trigger</button>
+</Tooltip>
--- a/packages/ui/src/lib/tooltip/Tooltip.spec.ts
+++ b/packages/ui/src/lib/tooltip/Tooltip.spec.ts
@@ -56,9 +56,7 @@ async function computePos(directionProps: Record<string, boolean>): Promise<{ to
   const outer = screen.getByText('pos').parentElement as HTMLElement;
 
   // Complain because we overwrite readonly method – this is intentional for the unit test.
-  // @ts-expect-error overriding for test stub
   wrapper.getBoundingClientRect = (): DOMRect => WRAPPER_RECT;
-  // @ts-expect-error overriding for test stub
   outer.getBoundingClientRect = (): DOMRect => TIP_RECT;
 
   await fireEvent.mouseEnter(wrapper);
--- a/packages/ui/src/lib/tooltip/Tooltip.spec.ts
+++ b/packages/ui/src/lib/tooltip/Tooltip.spec.ts
@@ -263,7 +263,7 @@ test('adds extra classes from `class` prop', () => {
     },
   });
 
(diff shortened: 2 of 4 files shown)


vancura:
This is awesome, thank you so much for the implementation. My only worry is this:

CleanShot 2025-06-30 at 10 50 50@2x

This tooltip doesn't make any sense; it doesn't bring any new information. Instead, it just duplicates the name of the component. I know this is out of the scope of this issue, but where is the tooltip coming from? Do component developers specify the tooltip string to be the same as the title one? That is one of the topics we should write in the UX guidelines for extension developers (that we're planning to write), so that they understand duplicating content is not right.

We can prevent this by detecting if the tooltip title matches the title. In that case, the tooltip shouldn't be displayed.

Thoughts?

/c @MariaLeonova

vzhukovs:
> I know this is out of the scope of this issue, but where is the tooltip coming from?

Usually, any UI component can be wrapped in the code with the following construction:

```
<Tooltip tip="visible text">
  <Component></Component> or html
<Tooltip>
```

this is how it works in PD.

vancura (approved):
LGTM

benoitf:
unit tests are failing

gastoner (commented):
LGTM on Mac after fixing the tests
Also to the future: is good idea having 8 boolean properties each one for one position what about having one property e.g. `position` with custom types which will be one of those 8?

benoitf (changes requested):
way too many changes there

there is no explanation on the current problem, what is being done, etc

axel7083 (changes requested):
The problem is similar to https://github.com/podman-desktop/podman-desktop/issues/10850 and might be interesting to find a common solution, maybe explore solution like defining an api, and custom component with position size at the root of the document

I understand that it would be nice to have this fix, but from a Svelte component perspective, we should not create and remove dynamically document nodes

feloy (changes requested):
(waiting for changes before reviewing)

dgolovin:
Probably related to #13460 too.

vancura:
I believe native tooltips (#13460) would fix this automatically.

vzhukovs:
Closed, because similar PR is being merged https://github.com/podman-desktop/podman-desktop/pull/14246

PR #13019, opened 2025-06-30
chore(test): clean hanging podman machine in e2e test

cbr7:
### What does this PR do?
One of the podman machine e2e tests causes a cascade failure of several test suites when it fails in a specific way because it leaves the podman machine hanging in a bad state.


danivilla9 (approved):
LGTM

PR #13022, opened 2025-06-30
chore(test): improve auth-utility playwright codebase

### LINKED ISSUE #13021: Improving playwright browser authentication utility code

While I was working on https://github.com/redhat-developer/podman-desktop-redhat-account-ext/issues/759, I noticed that there is a need for more debugging option while running playwright in external browser. Also, some other functionality is need like handling the cookies iframe.

Improve the `auth-utility` module code.

_No response_

### Additional context

_No response_

odockal:
### What does this PR do?
Improves authentication utilities and adds option for debugging. Adds function to handle cookie in iframe.
#13021 

### How to test this PR?

### DIFF
--- a/tests/playwright/src/utility/auth-utils.ts
+++ b/tests/playwright/src/utility/auth-utils.ts
@@ -106,6 +106,7 @@ export async function startChromium(port: string, tracesPath: string, args: stri
   return browserLaunch;
 }
 
+// to be deprecated, use getEntryFromConsoleLogs instead
 export async function getEntryFromLogs(
   page: Page,
   filter: RegExp,
--- a/tests/playwright/src/utility/auth-utils.ts
+++ b/tests/playwright/src/utility/auth-utils.ts
@@ -21,8 +21,6 @@ import { join } from 'node:path';
 import type { Browser, Locator, Page } from '@playwright/test';
 import { chromium, expect as playExpect } from '@playwright/test';
 
-import { TroubleshootingPage } from '../model/pages/troubleshooting-page';
-import { StatusBar } from '../model/workbench/status-bar';
 import { waitUntil } from './wait';
 
 export type ConfirmInputValue = {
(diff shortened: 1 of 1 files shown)


tests/playwright/src/utility/auth-utils.ts
```diff
@@ -51,29 +53,42 @@ export async function performBrowserLogin(
...
+    await page.screenshot({ path: join(path, 'screenshots', 'after_username_page.png'), type: 'png', fullPage: true });
+  }
   // password
   await playExpect(passwordAction.inputLocator).toBeVisible();
   await passwordAction.inputLocator.fill(passwordAction.inputValue);
   await playExpect(passwordAction.confirmLocator).toBeEnabled();
   await passwordAction.confirmLocator.click();
+  if (path) {
+    await page.screenshot({ path: join(path, 'screenshots', 'after_password_page.png'), type: 'png', fullPage: true });
+  }
```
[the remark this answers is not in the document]
odockal: `screenshot` call handles folder creation. No need to do it on my own.

tests/playwright/src/utility/auth-utils.ts
```diff
@@ -51,29 +53,42 @@ export async function performBrowserLogin(
...
   passwordAction: ConfirmInputValue,
   postLoginAction: (myPage: Page) => Promise<void>,
+  options: {
+    screenshotsPath: string | undefined;
+  } = { screenshotsPath: '' },
 ): Promise<void> {
   console.log(`Performing browser login...`);
+  const path = options.screenshotsPath;
+  if (path) {
+    await page.screenshot({ path: join(path, 'screenshots', 'initial_page.png'), type: 'png', fullPage: true });
```
cbr7, line 61: I'm not sure I understand what the point of these screenshots are.

odockal: To see what is happening in the browser. I was able to find out we got cookies iframe, error when logging in, and also a redirection notification...

tests/playwright/src/utility/auth-utils.ts
```diff
@@ -106,16 +122,59 @@ export async function getEntryFromLogs(
...
+  await waitUntil(
+    async () => {
+      try {
+        await playExpect(logLine).toBeVisible({ timeout: 3000 });
+        return true;
+      } catch (error: unknown) {
+        console.log(`Log line is not visible yet: ${error}`);
+        await troublePage.openGatherLogs();
+        await troublePage.openLogs();
+        return false;
```
cbr7: I'm not completely sure what this is supposed to do, but I don't think it's correct, if for no other reason than if the assert inside the try catch block fails multiple times it will be getting the logs from the catch block multiple times.

what is this code supposed to do, I think it can be simplified.

tests/playwright/src/utility/auth-utils.ts
```diff
@@ -106,16 +122,59 @@ export async function getEntryFromLogs(
...
+
+// function is dedicated to verify if some locator exists, depending on external circumstances
+export async function checkLocatorExistence(locator: Locator, timeout = 5000): Promise<boolean> {
+  try {
+    await locator.waitFor({ state: 'visible', timeout });
+  } catch (error: unknown) {
+    console.log(`Locator not found: ${error}`);
+    return false;
+  }
+  return true;
```
cbr7: Existance should be checked using `locator.count()` method call.

cbr7 (changes requested):
Left some comments

tests/playwright/src/utility/auth-utils.ts
odockal: When Troubleshooting -> OpenLogs tab is opened, it shows static content of the app logs at that moment. Since we want sso url entry, and if the url entry is not in the logs, one cannot simply wait for element to appear due to its static nature, to refresh the logs we need to switch to another tab and back.

TBH I do not like it either, but I cannot think of other way to do it.

odockal: Myabe If we could access console logs totally omitting UI, then it would be much easier.

tests/playwright/src/utility/auth-utils.ts
odockal: I am personally using tracing, but that is done from browser object and it would be very limiting if I go up to a browser level. Maybe in next turn? wdyt?

tests/playwright/src/utility/auth-utils.ts
odockal: I have tried different approach. using waitForEvent from page object to check for the console. Should be much more reliable. I also replaced former implementation but kept function not to break extension tests using old function call.

cbr7 (approved):

PR #13036, opened 2025-07-01
chore: add the rest of missing tray icon assets and fix loading

### LINKED ISSUE #9360: Tray Icon Color - make it dynamic according to OS theme mode (light/dark)

Thanks!

_No response_

### Additional context

Image

vancura:
### What does this PR do?

This PR enhances the tray icon handling system to properly support multiple resolutions (including retina displays) and improves error detection and recovery. The changes include:

- Added proper multi-resolution support for tray icons (1x and 2x/retina displays)
- Implemented robust error handling with fallback mechanisms when icon files are missing
- Enhanced the icon loading system to use `nativeImage` with proper representation handling
- Fixed asset path resolution for both development and production environments
- Added comprehensive logging for debugging icon loading issues
- Cleaned up obsolete SVG icon files that were never used in the project
- Added missing tray icon assets in both standard and high-resolution formats

Fixes #9360 

### How to test this PR?

1. Test tray icon display on both standard and high-DPI/retina displays
2. Switch between light and dark system themes to verify icon visibility
3. Test all tray icon states: initialized (empty), ready, error, and updating (animated)
4. Verify that missing icon files are gracefully handled with fallback icons
5. Check console logs for any icon loading warnings or errors
6. Test in both development and production builds to ensure asset paths work correctly


benoitf (changes requested):
this PR tries to do too many things at the same time

please split in small individual chunks

here it's impossible to check for a regression or focus a test on a particular thing

benoitf (dismissed):
the svg files should stay. it's the source file for the png file export

vancura:
> the svg files should stay. it's the source file for the png file export

I manually edited the icon PNGs in Photoshop, making the SVGs unnecessary. I can add them back, but since the final icons are created pixel by pixel, is it worth it? If someone wants to change the icons, they will need to edit them again, and these SVGs won't help them in any way.

> this PR tries to do too many things at the same time
> please split in small individual chunks
> here it's impossible to check for a regression or focus a test on a particular thing

The tray icon system was fundamentally broken due to multiple interdependent issues:

1. **Missing icon assets:** The codebase was trying to load PNG files that didn't exist. The system had SVG files, but the code was hardcoded to use .png extensions. This caused silent failures where icons wouldn't load (I added the rest of the files, so that should be fixed).

2. **No retina/HiDPI support:** The code did not handle high-resolution displays properly. It was missing `@2x.png` variants for retina screens, causing blurry icons on modern displays.

3. **Platform-specific logic failures:** The icon selection logic for different platforms and themes was flawed – it would try to load non-existent file combinations.

4. **No error handling:** When icons failed to load, the application would show a broken UI without any fallback mechanism.


Why it's hard to separate these changes:

1. Asset dependency chain
  - The new `createTrayImage()` method requires both 1x and 2x PNG files to exist.
  - Without adding the missing PNG files first, the new code would fail immediately.
  - The old code using `getIconPath()` directly with `setImage()` couldn't handle missing files.

2. Error handling requires fallback icons
  - The error handling logic needs the "empty" icon variants as fallbacks.
  - Without adding these specific icon files, the error handling would cascade into more failures.
  - Lines 184-198 in `tray-animate-icon.ts` specifically look for fallback icons.

3. Animation logic depends on complete icon sets
  - The animation cycles through step0-step3 icons.
  - Each step needs Dark/Template/default variants for different themes.
  - Missing any variant would break the animation loop.
  - The error handling in `animateTrayIcon()` (lines 84-92) stops the animation if any icon is missing.

5. Development vs production path resolution
  - The path resolution fix (lines 58-62) is critical because icons were in different locations in dev vs prod.
  - Without this fix, even if all icons existed, they wouldn't be found in development.
  - This affects ALL icon loading, not just specific ones.

If we tried to split this:
- PR 1 (just add icons): The app would still crash because the code expects SVGs
- PR 2 (just change to PNG loading): Would fail because PNGs don't exist yet
- PR 3 (just add error handling): Would mask the real problems without fixing them
- PR 4 (just add retina support): Would fail on missing 2x variants

Each change depends on the others to create a working system. The icon loading was completely broken before, and these changes together create the first working implementation.

vancura:
@benoitf: Please tell me what the next step should be. I can adjust this PR and start fresh:

1. This PR: Just keep the previously missing assets (makes the current code work, remove all code changes)
2. New PR: Add error handling
3. New PR: Add retina/HiDPI support
4. New PR: Fix path resolution (so that it works in all environments)

Each would build on the previous, but can be tested individually.

Sounds good?

benoitf:
yes it seems easier to review / smaller changes

vancura:
@benoitf This PR now contains only image changes.

packages/main/src/assets/tray-iconTemplateDark.svg
```diff
@@ -1,7 +0,0 @@
-<svg fill="none" version="1.1" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
```
benoitf, line 1: this file is deleted but it should be updated ?

vancura: There are always only three files for each state (`tray-icon*`, where `*` is following):
- (nothing) – light theme (black icon),
- dark – dark theme (white icon),
- template – macOS template icon.

Hence, the `templateDark` icon is not needed.

gastoner (commented):
On Linux KDE with light theme I see:  
image

Also might be a silly question, how can I get the PD into other  `initialized (empty), error, and updating (animated)` states?

vancura:
Right, this is to be expected. The icons are now loading, but the loading code is not implemented yet (it was, but the commit was reverted). For now, this is just icon files. There's also #13046 and another PR that I have uncommitted that fixes the icon loading.

vancura:
Also, if you [see the diff](https://github.com/podman-desktop/podman-desktop/pull/13036/files), you'll see that all the existing files are correctly either white, or dark (or template). I haven't changed the file's appearance other than sharpening the pixels. And added new files.

What you see now is that the errors in icon loading started to manifest.

benoitf (commented):
I'm not fully understanding this PR

I would expect only replacement of current files, while here there are new files that are being added

then, you could have something dedicated about some 'missing assets' and explaining why they're missing

but on this PR, I would expect to see only the update of current files with their updated version, no new file should pop up

vancura:
The problem is that the files were wrong. Many files were missing (e.g. HiDPI versions), some redundant (dark template ones), and the work was not just replacing the icons but cleaning them up.

vancura:
I explained it in a different comment: the problem with the tray icons is not just that the images are missing (or redundant), but also that the loading is wrong. There are bugs that you don't see manifest because the files were missing. And now you can finally see the issues because the icons are there. To add oil to the fire, you asked me to separate PRs into a couple, so that what we have now are partial PRs that need another.

I would say we should stop whatever is here and reconsider my overall change that worked well, but was contained not just in images, but also updated HiDPI code, and error detection.

All of us want to have updated icons. But what was not obvious before is that the loading is broken as well. The missing icons made it not manifest at all.

vancura:
To summarize, if you consider my work to be limited to updating icon files, I am done here. The loading is broken, but that's out of the scope of my work.

benoitf:
after discussion, there should probably have:

- 1st PR: update svg and png of current files (you're changing some icon details) (can be checked on any OS)
- 2nd PR: add hiRes/ @2x images for retina display (can be checked on a retina display looking if it's blury/detailed)
- 3rd PR: add missing 'light' icons (original issue) (should be checked on Windows as macOS is using templates)

daedalus-codes:
This is still what I see on Windows in dark mode, _even_ with the preference for the tray icon set to light. 

image

I realize this PR is closed, however, I found this PR because it finally annoyed me enough to search for this as bug.  It's the closest thing to resolving it and I haven't found any follow up PRs to address this issue.  The preference setting for tray icon color is irrelevant if it will not work.  Honestly, the tray icon light/dark setting should have an auto mode to use the OS/Windowing environment's light/dark theme setting.  

Occasionally this does annoy the heck out of me since using Podman, but I finally enough of it this morning for me to go on the hunt.  

@vancura Thank you for putting in the effort to resolves this!  I'm willing to assist in splitting this PR as @benoitf mentioned if that would bring this home.

> after discussion, there should probably have:
> 
> * 1st PR: update svg and png of current files (you're changing some icon details) (can be checked on any OS)
> * 2nd PR: add hiRes/ @2x images for retina display (can be checked on a retina display looking if it's blury/detailed)
> * 3rd PR: add missing 'light' icons (original issue) (should be checked on Windows as macOS is using templates)

vancura:
Hi @daedalus-codes, thanks for the feedback. Please, do you have Podman Desktop running/overridden to use the light theme, while you're in the Windows dark theme?

daedalus-codes:
This is my setting - Windows is also set to Dark Mode:

image

image

image

vancura:
@daedalus-codes ^ (#14732)

PR #13046, opened 2025-07-02
feat: update asset folder retrieval logic based on environment

### LINKED ISSUE #9360: Tray Icon Color - make it dynamic according to OS theme mode (light/dark)

vancura:
### What does this PR do?

This PR resolves the asset path resolution for tray icons to ensure they work correctly in all environments (development and production).

**Enhanced `getAssetsFolder()` method**: Now uses environment-aware path resolution:
- **Production**: Uses `app.getAppPath()` to locate assets within the app bundle
- **Development**: Uses `process.cwd()` to locate assets in the project root directory

**Problem solved:** Previously, the tray icon asset loading would fail in development environments because it only used `app.getAppPath()`, which doesn't point to the correct location during development. This caused the tray icons to not display properly when running the application in development mode.

Fixes #9360

### How to test this PR?

**Manual testing:**

1. **Development environment**: Run the application in development mode and verify that the tray icons display correctly
2. **Production environment**: Build and run the production version and verify that tray icons still work

**Automated testing:**
- [x] Unit tests verify `getAssetsFolder()` returns correct paths for both production and development modes
- [x] Tests mock electron's `app.getAppPath()` and `process.cwd()` to simulate both environments
- [x] Tests validate that the path resolution logic works as expected

### DIFF
--- a/packages/main/src/tray-animate-icon.spec.ts
+++ b/packages/main/src/tray-animate-icon.spec.ts
@@ -42,6 +42,7 @@ vi.mock('electron', async () => {
     },
     nativeTheme: {
       on: vi.fn(),
+      shouldUseDarkColors: false,
     },
   };
 });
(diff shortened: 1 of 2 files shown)


vancura:
@benoitf From what I see, the smoke E2E check failed due to a timeout. What can I do now?

benoitf:
on my end I can currently see the macOS tray icons in development mode so I'm not sure about what it's supposed to be fixing

benoitf:
> @benoitf From what I see, the smoke E2E check failed due to a timeout. What can I do now?

job can be restarted (if it's a flaky issue)

vancura:
> on my end I can currently see the macOS tray icons in development mode so I'm not sure about what it's supposed to be fixing

This PR just fixed the folder retrieval logic. The other PR (coming) is waiting for this change. You wanted to separate it into smaller tasks that are easier to review.

benoitf:
> Problem solved: Previously, the tray icon asset loading would fail in development environments because it only used app.getAppPath(), which doesn't point to the correct location during development. This caused the tray icons to not display properly when running the application in development mode.


how to check that currently it's not working as expected ?

because this PR says there is an issue but I don't see any difference

vancura:
See my reply in [#13036](https://github.com/podman-desktop/podman-desktop/pull/13036#issuecomment-3028194182). The thing is that this commit doesn't have the icons changed in there, because we wanted to split the PR into many.

vancura:
Closing the PR, but I will reopen it differently.

PR #13049, opened 2025-07-02
feat: adds dropdown option to message box

gastoner:
### What does this PR do?
Adds dropdown menu to message box

### How to test this PR?
CI

### DIFF
--- a/packages/api/src/dialog.ts
+++ b/packages/api/src/dialog.ts
@@ -16,12 +16,17 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+export interface IconButtonType {
+  label: string;
+  icon: string;
+}
+
 export interface DropdownType {
   heading: string;
   buttons: string[];
 }
 
-export type ButtonsType = string | DropdownType;
+export type ButtonsType = string | DropdownType | IconButtonType;
 
 /**
  * Options to configure the behavior of the message box UI.
--- a/packages/api/src/dialog.ts
+++ b/packages/api/src/dialog.ts
@@ -16,14 +16,20 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-export interface IconButtonType {
-  label: string;
-  icon: string;
+interface BaseButtonType {
+  type: 'iconButton' | 'dropdownButton';
 }
 
-export interface DropdownType {
+export interface IconButtonType extends BaseButtonType {
+  label: string;
+  icon: string;
+  type: 'iconButton';
+}
+
+export interface DropdownType extends BaseButtonType {
   heading: string;
   buttons: string[];
+  type: 'dropdownButton';
 }
 
 export type ButtonsType = string | DropdownType | IconButtonType;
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/dialogs/MessageBox.svelte
```diff
@@ -145,6 +151,24 @@ function getButtonType(b: boolean): ButtonType {
...
+              onChange={async (option): Promise<void> => {
+                let optionIndex: number | undefined = dropdownButtons.buttons.indexOf(option);
+                if (optionIndex === -1) {
+                  optionIndex = undefined;
+                }
+                await clickButton(i, optionIndex);
+              }}
+              options={dropdownButtons.buttons.map(button => ({label: button, value: button}))}/>
+          {:else}
+            {@const button = buttons[i] as IconButtonType}
```
benoitf, line 169: how can you cast directly as IconButtonType or DropdownType

I think there are some missing  typing ?

should we have `buttons[i].type` ? and if it's DropdownType then you can cast, etc

benoitf (commented):
I think the direct cast is very fragile as after that there is no extra check

packages/renderer/src/lib/dialogs/MessageBox.svelte
gastoner: Because only DropdownType has a heading property, if it does not have it is IconButtonType then, but I can probably add type and specify it here explicitly

gastoner: I've added the type property

packages/api/src/dialog.ts
```diff
@@ -16,12 +16,21 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-export interface DropdownType {
+interface BaseButtonType {
+  type: 'iconButton' | 'dropdownButton';
+}
+
+export interface IconButtonType extends BaseButtonType {
+  label: string;
```
benoitf, line 24: it's missing `type: 'iconButton'` here and `type: 'dropdownButton'` in interface DropdownType

benoitf (commented):
missing enforcement in interfaces definition + unit tests failing

benoitf:
but I would say we're close to be merged (only 2 minor easy fixes)

gastoner:
@benoitf updated

packages/api/src/dialog.ts
```diff
@@ -19,11 +19,13 @@
 export interface IconButtonType {
   label: string;
   icon: string;
+  type?: 'iconButton';
```
benoitf: question: why would it be optional field ?

gastoner: Because I would have to fill the type in here explicitly
So If I set it to iconButton and in another file (original PR) need to set it again to iconButton because it is required.
Maybe set it to readonly ?

benoitf: I think you need to keep previous commit where parent has the two options like 'icon' | 'dropdown' but in the dedicated interface then you only keep one option

benoitf: so from the global interface there is a choice but if you create a iconType you can't have anything else than type = 'icon'

benoitf (approved):

PR #13058, opened 2025-07-03
chore(tray): add missing tray icon assets

vancura:
### What issues does this PR fix or reference?

This PR adds the rest of the missing tray icon asset files:

- Adding fixed and previously missing SVG files, completing the set of source assets.
- Adding previously missing PNG files (the rest of the light- and dark-theme icons, including their HiDPI/retina resolutions.

There are no changes to existing files; this PR only adds new files that are necessary for the system to display icons correctly in all themes, including those for high-resolution displays.

### What issues does this PR fix or reference?

Fixes #9360

### How to test this PR?

1. Observe the diff with the GitHub diff tool.
2. Review that there are no file changes, only new SVG and PNG files.
3. Verify that icons are either light- or dark-theme based, considering their filenames.

- [ ] Tests are covering the bug fix or the new feature

There are no tests; this PR contains only new asset files (SVGs and PNGs).

PR #13057, opened 2025-07-03
chore(tray): update existing tray icons

vancura:
### What issues does this PR fix or reference?

- Updates existing tray PNG icons (sharpening, pixel-fitting).
- Updates existing tray SVG icons (updates color for better compatibility and centers the glyph in `tray-icon-error*.svg`).
- Removes obsolete (unused) `tray-iconTemplateDark.svg` (template images are just templates, not dark or light).

### What issues does this PR fix or reference?

Fixes #9360

### How to test this PR?

1. Compare the files one by one using GitHub's diff feature.
2. Verify that the colors match (white icons should remain white, and black icons should be black).
3. Observe the pixel sharpening and cleanup of the icons. The icons should no longer be blurry.

- [ ] Tests are covering the bug fix or the new feature

There are no tests. This PR contains only asset changes.

PR #13059, opened 2025-07-03
fix(tray): enhance tray icon handling and improve asset path resolution (WIP)

### LINKED ISSUE #9360: Tray Icon Color - make it dynamic according to OS theme mode (light/dark)

vancura:
### What does this PR do?

This PR fixes the tray icon loading and improves the error handling:

- Updated `getAssetsFolder` method to differentiate between production and development environments for asset paths.
- Added error handling in `animateTrayIcon` to prevent continuous failures during animation.
- Improved icon loading logic in `updateIcon` to utilize `createTrayImage` for better image management.

**Note: This PR is a draft; it's waiting for the following PRs (which add the missing files that are needed for testing):**

- #13057
- #13058

Fixes #9360

### How to test this PR?

TBD


vancura:
image

gastoner:
On KDE I can see: 
image (2)

image (3)

With dark theme it seems OK, just light theme on KDE is not visible really :(

vancura:
Thanks for the feedback, @gastoner. I need to install KDE to see where this is.

vancura:
@nichjones1 Can we postpone this task, please? I need to get my Linux installation working to test this, and I am afraid I won't be able to do it this sprint.

axel7083:
cc @vancura I can see `WIP` in the title, can we have this as a draft to avoid notifying reviewers, thanks. 

If you need the CI, you can add the `area/ci` label.

vancura:
Thanks, I know it's noisy, sorry about that, but that was because I needed bots to review my work. I didn't know about the `area/ci` label, amazing, thanks!

vancura:
Hey @benoitf, I've been investigating the tray icon behavior in https://github.com/podman-desktop/podman-desktop/blob/main/packages/main/src/tray-animate-icon.ts#L85-L97 and discovered some inconsistencies in how we handle theming across platforms.

The existing implementation always uses dark-theme icons (white ink) on Linux, which doesn't account for the fact that KDE system trays can be either light or dark-themed. While this works correctly for GNOME (which uses a consistently dark tray), it creates visual inconsistencies in KDE environments.

**Key Problem with `nativeTheme.shouldUseDarkColors`:**

I've observed that the current logic has a fundamental flaw – the tray icon appearance is being influenced by Podman Desktop's internal theme preference rather than the actual system theme. Specifically:

- The tray icon changes based on the theme selected in PD preferences (system/dark/light).
- However, the tray icon should reflect the **system's** theme, not the application's theme preference.
- This creates a disconnect where users might have a light system theme but see a dark tray icon because PD defaults to dark mode.

**Platform Behavior:**

- **macOS**: Works correctly due to template icons that automatically adapt.
- **Linux (KDE)**: Problematic - doesn't respect system tray theme.
- **Windows**: Unclear why this appears to work correctly.

**Proposed Solution:**

We need to decouple the tray icon theme detection from Podman Desktop's internal theme settings. The tray icon should exclusively follow the system's light/dark preference, regardless of what theme is configured within the application.

What's your recommendation for the best approach to achieve proper system theme detection for tray icons across platforms, please?

vancura:
We're closing this PR to take a different approach. After a thorough technical discussion with @benoitf, we've decided to close this PR and pursue a different solution for the tray icon color issues.

The existing tray icon implementation actually works correctly on most platforms:

- **macOS**: Uses template icons that automatically adapt to the menu bar theme
- **Windows**: Works correctly
- **Linux GNOME**: Uses white icons consistently, which works well with GNOME's dark tray
- **Linux KDE**: This is where the issue manifests

The core issue isn't with our system theme detection logic, but rather with our approach:

1. **Incorrect theme source**: Our current logic uses Podman Desktop's internal theme setting to determine tray icon color, rather than the actual system tray theme. This means:
   - When Podman Desktop starts in dark mode **→ white tray icon**
   - When the user switches Podman Desktop to light mode **→ tray icon changes to black**
   - This happens regardless of what the actual system tray looks like

2. **Platform complexity**: Linux desktop environments handle tray icons very differently:
   - **GNOME**: Consistent approach, always uses dark tray area
   - **KDE/Plasma**: Highly customizable – users can have a light system theme but a dark tray area, or vice versa
   - Electron's `nativeTheme.shouldUseDarkColors` API can be unreliable on Linux

While this PR does improve asset loading and error handling, the fundamental approach of trying to automatically detect the "correct" tray icon color has several issues:

1. **API limitations**: Electron's theme detection API is known to be buggy on Linux, especially KDE
2. **User customization**: KDE users often customize tray appearance independently of overall system theme
3. **Maintenance burden**: Trying to handle all edge cases of Linux desktop environment configurations adds complexity

PR #13060, opened 2025-07-03
fix: do not search full path executable

### LINKED ISSUE #12240: Docker compose binary not removed from installed location on windows
### Bug description

I installed docker-compose via CLI. Then If I try to remove it, all looks good, but the binary is still present on the expected path. 

And there is also an info entry in the console log:
```
[compose] Error: Invalid pattern in specified in "path:pattern"
```

### Operating system

Windows 10

### Installation Method

None

### Version

1.18.0

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

feloy:

### What does this PR do?

A `where.exe` / `which` is executed to find the executable to delete, but the path passed to the command is already the full path.

With the `which` command, it is harmless as this command works correctly with a full path, but `where.exe` raises an error `Invalid pattern in specified in "path:pattern"` in this case.

This PR removes the where/which part, and directly deletes the file at the given full path

### Screenshot / video of UI

N/A

Fixes #12240 

### How to test this PR?

On Windows, try to uninstall docker-compose / kind / kubectl

### DIFF
--- a/extensions/compose/src/extension.spec.ts
+++ b/extensions/compose/src/extension.spec.ts
@@ -446,7 +446,6 @@ describe('registerCLITool', () => {
 
     await installer?.doUninstall({} as unknown as Logger);
     expect(fs.promises.unlink).toHaveBeenNthCalledWith(1, 'storage-path');
-    expect(extensionApi.process.exec).toHaveBeenCalledWith('which', ['system-wide-path']);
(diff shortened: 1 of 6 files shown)


SoniaSandler (approved):
LGTM codewise, not tested on Windows

jeffmaury (approved):
LGTM

Issue #12633, opened 2025-05-26
Telemetry: Optimize `activateExtension` event


Firewall:
### Is your enhancement related to a problem? Please describe

This event gets fired every time an extension is started. This is leading to millions of events. **23%** of all our events.

### Describe the solution you'd like

Reduce the amount of events that are sent.

**Idea:** 
Can we add the extensions the users has as a property to the user. 

```
User123:
extension.kind: enabled
extension.quadlets: enabled
```

That we we would only need to send an event when it changes, which would significantly reduce the events. 

**Idea 2:** 
Change the event into a comma seperated string after the app has started.

`activatedExtensions: kind, quadlets`

That way we need to change our queries to `where activatedExtensions contains "kind"`
It would reduce the amount of events we send significantly but it would still create an event every time the app is started so it would be less effect than the above.

### Describe alternatives you've considered

- Removing the event, but this is reallly useful for us to understand which extensions users are installing using.

benoitf:
only a side note:

even if we do changes, it'll done only for new versions.

All previous versions of Podman Desktop would still send data

somehow, data should be backward compliant

or we drop all the previous events at the sources / discard them

and we have new events only for upcoming new versions

Firewall:
> All previous versions of Podman Desktop would still send data

Is there a way to prevent that? I think that will always happen, right? We might consider blocking the old event in amplitude so now new data is sent? It might be relevant for https://github.com/podman-desktop/podman-desktop/issues/12635 where we send loads of extra events 

> somehow, data should be backward compliant
> or we drop all the previous events at the sources / discard them
> and we have new events only for upcoming new versions

We can [see the charts that use the this event](https://app.amplitude.com/data/redhat/Podman%20Desktop/events/main/latest/activateExtension?view=All&eventsTab=Events&tab=USED%2520BY&propertyValidityFilter=All%2520Properties) here we can update the query to handle both.

cdrage:
After the meeting / discussion we'll be:
* Removing activateExtension in favour of a new one that sends extensions telemetry after startup, once.
* Eventually blocking activateExtension in Segment / Amplitude for users not upgrading yet.
* New one named something like "activatedExtensionsOnStartup" that will fire after all extensions try to start
* Push comma-deliminated (or array depends on implementation, see main issue) of what extensions are "enabled" on startup / use has installed by default

cc @Firewall please correct if I got anything wrong!

benoitf:
[...]

PR #13063, opened 2025-07-03
chore: add an aggregate method for telemetry

benoitf:
### What does this PR do?
this method will send all data as an array of object as long as the method is called for a given event it will delay the sending of the data

example: If I try to propagate one event with {foo: 'bar'} and another one with {foo: 'baz'}, it will collect those
and it'll sent a single event with `[{foo: 'bar'}, {foo: 'baz'}]` data after a delay

related to https://github.com/podman-desktop/podman-desktop/issues/12634 and https://github.com/podman-desktop/podman-desktop/issues/12633

### How to test this PR?

unit tests added

### DIFF
--- a/packages/main/src/plugin/telemetry/telemetry.ts
+++ b/packages/main/src/plugin/telemetry/telemetry.ts
@@ -64,6 +64,8 @@ export type EventType =
  */
 @injectable()
 export class Telemetry {
+  public static readonly DEFAULT_DELAY_AGGREGATE = 10_000; // 10 seconds
+
   private static readonly SEGMENT_KEY = 'Mhl7GXADk5M1vG6r9FXztbCqWRQY8XPy';
 
   private cachedTelemetrySettings: TelemetryRule[] | undefined;
@@ -353,7 +355,7 @@ export class Telemetry {
    * @param eventProperties all properties that will be aggregated into an array and be sent after the delay if there is no other event received during this time window
    * @param delay the delay in milliseconds before sending the aggregated event, default is 10 seconds
    */
-  aggregateTrack(event: EventType, eventProperties?: unknown, delay: number = 10_000): void {
+  aggregateTrack(event: EventType, eventProperties?: unknown, delay: number = Telemetry.DEFAULT_DELAY_AGGREGATE): void {
     // skip event ?
     if (this.shouldDropEvent(event)) {
       return;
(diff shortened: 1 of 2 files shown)


benoitf:
cc @Firewall 

sending an array of objects will work as expected ?

packages/main/src/plugin/telemetry/telemetry.ts
```diff
@@ -345,6 +347,68 @@ export class Telemetry {
...
+      this.internalTrack(event, propertiesToSend).catch((err: unknown) => {
+        console.warn(`Error sending aggregated event: ${event}`, err);
+      });
+      // remove the timeout from the map
+      this.aggregateTimeoutEvents.delete(event);
+    }, delay);
+
+    // sets a new timeout for this event name
+    this.aggregateTimeoutEvents.set(event, { timeout, properties: propertiesToSend });
+  }
```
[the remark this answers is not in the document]
benoitf: only few event (also it's sent after a while so it's never growing to infinite values)

benoitf:
it looks like the data on amplitude/segment are OK

Firewall (approved):
Tested, and events are coming into amplitude correctly, and aggregated & split. Great job 🚀

jeffmaury (approved):
LGTM

packages/main/src/plugin/telemetry/telemetry.ts
```diff
@@ -345,6 +347,69 @@ export class Telemetry {
     return dropIt;
   }
 
+  /**
+   * will cumulate the data for the given the event and send it to the telemetry system.
+   * @param event the event to send after the given delay
+   * @param eventProperties all properties that will be aggregated into an array and be sent after the delay if there is no other event received during this time window
+   * @param delay the delay in milliseconds before sending the aggregated event, default is 10 seconds
+   */
+  aggregateTrack(event: EventType, eventProperties?: unknown, delay: number = 10_000): void {
```
MarsKubeX: Hi @benoitf. Could this default delay of 10_000 be extracted to a shared constant? This would allow both the implementation and tests to reference a single source of truth, preventing test failures if the default value changes.

benoitf: @MarsKubeX fixed

MarsKubeX (approved):

PR #13064, opened 2025-07-03
chore: use the new aggregate method to track createProviders

### LINKED ISSUE #12634: Telemetry: Optimize `createProvider` event

benoitf:
### What does this PR do?
use the aggregated report for createProvider event (now becoming createProviders)

depends on :
 - [x] https://github.com/podman-desktop/podman-desktop/pull/13063
 
fixes https://github.com/podman-desktop/podman-desktop/issues/12634

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/provider-registry.spec.ts
+++ b/packages/main/src/plugin/provider-registry.spec.ts
@@ -78,6 +78,7 @@ class TestProviderRegistry extends ProviderRegistry {
 
 const telemetry: Telemetry = {
   track: vi.fn(),
+  aggregateTrack: vi.fn(),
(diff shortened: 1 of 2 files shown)


benoitf:
rebased as the required PR has been merged

PR #13071, opened 2025-07-04
chore: use aggregating way to report activateExtension event

### LINKED ISSUE #12633: Telemetry: Optimize `activateExtension` event

benoitf:
### What does this PR do?

instead of sending like 12 events in a row, send only one event from time to time that has aggregated the previous values

PR will be rebased when the dependent PR has been merged (separate commit)

depends on :
 - [x] https://github.com/podman-desktop/podman-desktop/pull/13063

fixes #12633 

### How to test this PR?

you can monitor/debug the event being sent

### DIFF
--- a/packages/main/src/plugin/extension/extension-loader.spec.ts
+++ b/packages/main/src/plugin/extension/extension-loader.spec.ts
@@ -203,7 +203,7 @@ const onboardingRegistry: OnboardingRegistry = {
 } as unknown as OnboardingRegistry;
 
 const telemetryTrackMock = vi.fn();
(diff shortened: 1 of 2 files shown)


benoitf:
rebased as the required PR has been merged

MarsKubeX (approved):
LGTM

Issue #10077, opened 2024-11-22
Building image with a capital in the name hangs


deboer-tim:
### Bug description

I inadvertently called an image 'gpuTest' when building, and it causes the build to hang for 5 minutes before erroring out because capitals are not allowed in repo names. We should validate for this, or fail fast.

### Operating system

macOS 15.1

### Installation Method

None

### Version

1.14.1

### Relevant log output

```shell

```

cbr7:
I just tried to reproduce this and it does still happen in 1.17.1, however for me there is no freeze/hangup, the build fails right away with the following error: 

<img width="1136" alt="Image" src="https://github.com/user-attachments/assets/5db56fcd-1127-410f-a460-df5cff8365fc" />

simonrey1:
I also reproduce the error on 1.19 with the build failing right away.

An idea would be to check the input to be lowercase.

PR #13086, opened 2025-07-07
chore(renderer): Migrate BuildImageFromContainerfile to svelte5

simonrey1:
### What does this PR do?

Refactor BuildImageFromContainerfile to svelte5

Referencing #10077, migrating to svelte5 before addressing the issue.


packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -28,59 +28,19 @@ import { type BuildImageCallback, disconnectUI, eventCollect, reconnectUI, start
 import BuildImageFromContainerfileCards from './BuildImageFromContainerfileCards.svelte';
 import RecommendedRegistry from './RecommendedRegistry.svelte';
 
-export let taskId: number | undefined;
-let buildImageInfo: BuildImageInfo = createDefaultBuildImageInfo();
-let providers: ProviderInfo[] = [];
+interface Props {
+  taskId: number | undefined;
```
benoitf:
```suggestion
  taskId?: number;
```

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -328,6 +288,57 @@ async function abortBuild(): Promise<void> {
...
+let providerConnections = $derived(
+  providers
+    .map(provider => provider.containerConnections)
+    .flat()
+    .filter(providerContainerConnection => providerContainerConnection.status === 'started'),
+);
+let selectedProvider = $derived(providerConnections.length > 0 ? providerConnections[0] : undefined);
+$effect(() => {
+  buildImageInfo.selectedProvider = selectedProvider;
+});
```
benoitf: question: do you think it's possible to merge the 3 effects ? b/c it may be hard to follow the logic across 3 functions


can we also put more stuff into a derived function

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -328,6 +288,57 @@ async function abortBuild(): Promise<void> {
...
+    // select the parent directory of the file as default
+    buildImageInfo.containerBuildContextDirectory = containerFilePath.replace(/\\/g, '/').replace(/\/[^\/]*$/, '');
+  }
+});
+let containerImageName = $derived(buildImageInfo.containerImageName);
+let providerConnections = $derived(
+  providers
+    .map(provider => provider.containerConnections)
+    .flat()
+    .filter(providerContainerConnection => providerContainerConnection.status === 'started'),
```
benoitf: suggestion: looks like now we prefer reduce over map/flat/filter

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
simonrey1: Thanks, I prefer this syntax as well (I used the migration script from svelte)

simonrey1: 7bef3eebc41e344d18079fb29d07cc7c2236e03f

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
simonrey1: 4d114293933ecc8f860ce24290aac7b10fce53a2

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
simonrey1: 7fb2e155987ba576ad010d25adda109b6352c36d

Svelte migration script has detected side effects in the effect block, but I will check if some other things can be derived.

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -28,59 +28,19 @@ import { type BuildImageCallback, disconnectUI, eventCollect, reconnectUI, start
...
-export let taskId: number | undefined;
-let buildImageInfo: BuildImageInfo = createDefaultBuildImageInfo();
-let providers: ProviderInfo[] = [];
+interface Props {
+  taskId?: number;
+}
+
+let { taskId = $bindable() }: Props = $props();
+let buildImageInfo: BuildImageInfo = $state(createDefaultBuildImageInfo());
+let providers: ProviderInfo[] = $state([]);
```
axel7083: suggestion: there seems to be a simplification possible prior (separate PR?), we currently seems to do some magic trick to get the store value

https://github.com/podman-desktop/podman-desktop/blob/7fb2e155987ba576ad010d25adda109b6352c36d/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte#L265-L267

we can get rid of `providers` and use `$providerInfos` instead directly; no need to manage the unsubscribe object, this is managed by svelte when using the `$` synthax

axel7083 (commented):
Nice work, the svelte migration tool is something we should consider more often

I noted one thing, that could be addressed separately, but would really be nice to have

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
simonrey1: 039f4e76cfb15d18bcffd7204fdecf36d5f90fe7

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -328,6 +282,55 @@ async function abortBuild(): Promise<void> {
   buildImageInfo.buildRunning = false;
   buildImageInfo.buildFinished = true;
 }
+
+let platforms = $derived(buildImageInfo.containerBuildPlatform ? buildImageInfo.containerBuildPlatform.split(',') : []);
+let containerFilePath = $derived(buildImageInfo.containerFilePath);
```
axel7083: question: why declaring a new derived state? `buildImageInfo` is already a state, so, using `buildImageInfo.containerFilePath` in other derived state should work directly?

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -328,6 +282,55 @@ async function abortBuild(): Promise<void> {
...
+let containerBuildContextDirectory = $derived(buildImageInfo.containerBuildContextDirectory);
```
axel7083: Same remark as https://github.com/podman-desktop/podman-desktop/pull/13086/files#r2191810259

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -328,6 +282,55 @@ async function abortBuild(): Promise<void> {
...
+
+let containerImageName = $derived(buildImageInfo.containerImageName);
```
axel7083: same remark as https://github.com/podman-desktop/podman-desktop/pull/13086/files#r2191810259

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
simonrey1: Because the migation script did so but I will simplify thanks :)

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
simonrey1: 48ea51a943f5986f5d84678cc8889f4e47568621

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
simonrey1: 48ea51a943f5986f5d84678cc8889f4e47568621

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
simonrey1: 48ea51a943f5986f5d84678cc8889f4e47568621

axel7083 (approved):
LGTM, great job, this was a pretty complicated component!

axel7083:
FYI we usually like to have two reviewers for some PR, depending on the changes (this was not a large change, so one LGTM is usually enough - more on our review process in [CONTRIBUTING.md#review-process](https://github.com/podman-desktop/podman-desktop/blob/main/CONTRIBUTING.md#review-process))

However, since another person was involved in the review, this is usually nice to have all the person giving their approval to finally merge, if you get an approval and the other person is not responding for some time you could ping or renew the request review using the `Re-request review` button

image

PR #13087, opened 2025-07-07
fix: reconnect logger handler only once

### LINKED ISSUE #12430: Logs of cancelled tasks like creating RHEL VM or Kind Cluster keep going on forever
### Bug description

When cancelling the creation of a RHEL VM or a Kind cluster and going back to the creation form page from the task manager, the logs keep showing up with the same message

### Operating system

macOS 15.1.1 M1

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

1. Create a RHEL VM or a Kind cluster and cancel the action before it's done
2. Exit the creation form page
3. Go back to the creation form page from the task manager by clicking on `go to task >`
4. Check the logs in that page 

### Relevant log output

```shell

```

### Additional context

https://github.com/user-attachments/assets/4edd77b7-5177-4473-84c3-aef809376147

https://github.com/user-attachments/assets/f41a9050-f3d6-47e0-b578-0c53c79907d7

SoniaSandler:

### What does this PR do?
This PR ensures that the reconnectUI function is called only once.
By using the `terminalLogs` value inside the effect block, any time some text is added to the terminal (especially from the reconnectUI function), the `terminalLogs` logs object is considered changed, and the effect block is executed, leading to unending logs.
Even using a derived value such as `let logsTerminalExists: boolean = $derived(logsTerminal !== undefined)` inside the effect (`if (logsTerminalExists && loggerHandlerKey)`), still leads to more and more logs as it depends on `logsTerminal`

Fixes https://github.com/podman-desktop/podman-desktop/issues/12430

### How to test this PR?

1. Create a RHEL VM or a Kind cluster and cancel the action before it's done
2. Exit the creation form page
3. Go back to the creation form page from the task manager by clicking on go to task >
4. Assert that there is only one copy of the logs and that nothing more is added

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
@@ -100,9 +100,10 @@ $effect(() => {
   if (!reconnectedUI && logsTerminal && loggerHandlerKey) {
     try {
       reconnectUI(loggerHandlerKey, getLoggerHandler());
-      reconnectedUI = true;
     } catch (error) {
       console.error('error while reconnecting', error);
+    } finally {
+      reconnectedUI = true;
     }
   }
 });
--- a/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
@@ -93,12 +93,14 @@ let contextsUnsubscribe: Unsubscriber;
 const buttonLabel = connectionInfo ? 'Update' : 'Create';
 const operationLabel = connectionInfo ? 'Update' : 'Creation';
 let loggerHandlerKey: symbol | undefined = undefined;
+let reconnectedUI: boolean = false;
 
 // reconnect the logger handler
 $effect(() => {
-  if (logsTerminal && loggerHandlerKey) {
+  if (!reconnectedUI && logsTerminal && loggerHandlerKey) {
     try {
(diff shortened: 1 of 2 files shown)


simonrey1:
I have reproduced the bug on default and not on your PR, looks great!

What happens if the `reconnectedUI` is set to true but we go in another task, will it be restored to false?

packages/renderer/src/lib/preferences/PreferencesConnectionCreationOrEditRendering.svelte
```diff
@@ -93,12 +93,14 @@ let contextsUnsubscribe: Unsubscriber;
...
 let loggerHandlerKey: symbol | undefined = undefined;
+let reconnectedUI: boolean = false;
 
 // reconnect the logger handler
 $effect(() => {
-  if (logsTerminal && loggerHandlerKey) {
+  if (!reconnectedUI && logsTerminal && loggerHandlerKey) {
     try {
       reconnectUI(loggerHandlerKey, getLoggerHandler());
+      reconnectedUI = true;
```
axel7083: suggestion: Could we have the reconnectedUI in the finally ? the problem may occurs, if the reconnectedUI throw an error, we would be in a weird state, having it in the finally would probably avoid that

SoniaSandler:
> What happens if the `reconnectedUI` is set to true but we go in another task, will it be restored to false?

@eqqe Yes, I checked manually and it looks like the page of the other task is a new rendering of this form page, so the `reconnectedUI` var is initialized again with `false`

cdrage (approved):
Looks great when there's more code in the tests vs the fix :)

LGTM!

Tried it and was unable to reproduce the bug on mac M1.

gastoner (approved):
LGTM on Linux

PR #13103, opened 2025-07-08
fix: display an error to the user using uppercase in image name

### LINKED ISSUE #10077: Building image with a capital in the name hangs

simonrey1:
### What does this PR do?

Displays an error to the user using uppercase in image name

### Screenshot / video of UI

Screenshot 2025-07-08 at 11 17 12

Fixes #10077

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -335,7 +335,7 @@ let hasInvalidFields = $derived(
     !buildImageInfo.containerBuildContextDirectory ||
     (platforms.length > 1 && !buildImageInfo.containerImageName) ||
     platforms.length === 0 ||
-    !errorContainerImageName ||
+    !!errorContainerImageName ||
     !selectedProvider,
 );
 </script>
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -324,19 +324,20 @@ $effect(() => {
   buildImageInfo.selectedProvider = selectedProvider;
 });
 
-let isImageNameLowercase = $derived(
-  buildImageInfo.containerImageName === buildImageInfo.containerImageName.toLowerCase(),
+let errorContainerImageName = $derived(
+  buildImageInfo.containerImageName === buildImageInfo.containerImageName.toLowerCase()
+    ? undefined
+    : 'Image name should be lowercase',
 );
+
 let hasInvalidFields = $derived(
   !buildImageInfo.containerFilePath ||
     !buildImageInfo.containerBuildContextDirectory ||
     (platforms.length > 1 && !buildImageInfo.containerImageName) ||
     platforms.length === 0 ||
-    !isImageNameLowercase ||
+    !errorContainerImageName ||
     !selectedProvider,
 );
-
-let errorContainerImageName = $derived(isImageNameLowercase ? undefined : 'Image name should be lowercase');
 </script>
 
 <EngineFormPage
--- a/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
+++ b/packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
@@ -379,7 +379,7 @@ let hasInvalidFields = $derived(
           name="containerImageName"
           id="containerImageName"
           placeholder="Image name (e.g. quay.io/namespace/my-custom-image)"
-          error={isImageNameLowercase ? void 0 : 'Image name should be lowercase'}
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -374,6 +379,7 @@ let hasInvalidFields = $derived(
           name="containerImageName"
           id="containerImageName"
           placeholder="Image name (e.g. quay.io/namespace/my-custom-image)"
+          error={isImageNameLowercase ? void 0 : 'Image name should be lowercase'}
```
benoitf:
```suggestion
          error={isImageNameLowercase ? undefined : 'Image name should be lowercase'}
```

simonrey1: I can do this change but I was wondering whether the `no-undefined` rule of ESLint would be relevant in this project ([https://eslint.org/docs/latest/rules/no-undefined](https://eslint.org/docs/latest/rules/no-undefined))

benoitf: we're using undefined everywhere (as default value if not defined)

but this project disable the usage of `null`

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -323,11 +323,16 @@ $effect(() => {
   }
   buildImageInfo.selectedProvider = selectedProvider;
 });
+
+let isImageNameLowercase = $derived(
+  buildImageInfo.containerImageName === buildImageInfo.containerImageName.toLowerCase(),
+);
```
axel7083: question (non-blocking): is there other condition to the name of an image ? is it alphanumerical character only? 

My point is, should we make it maybe more generic like, having several checks?

```svelte
let nameInputError: string | undefined = $dervied.by(() => {
 if( /* condition 1 */) return 'error condition 1';
 return undefined;
});
```

axel7083 (commented):
See https://github.com/podman-desktop/podman-desktop/pull/13103/files#r2192449345

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
benoitf: AFAIK also it's not possible to shadow undefined in our project as well due to strict mode

simonrey1: 6ac74d239abd26c51a2a53ec1c0b525b71a1d2b2

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
benoitf: I think we're mostly using this pattern (logic outside of the fields)

so I think it'll more readable yes if we only pass/give the error field and then avoid `? : undefined: 'error message'`

simonrey1: The validation seems to be done here in podman but it may depends on the provider
https://github.com/containers/podman/blob/aab1029b4fdc022528880a01565960c7400ad439/pkg/api/handlers/utils/images.go#L44

simonrey1: @benoitf I moved the logic in this commit cdf88d81566c2ae06372f7c387abd5ff3972c485

benoitf (approved):

packages/renderer/src/lib/image/BuildImageFromContainerfile.svelte
```diff
@@ -323,13 +323,20 @@ $effect(() => {
...
 let hasInvalidFields = $derived(
   !buildImageInfo.containerFilePath ||
     !buildImageInfo.containerBuildContextDirectory ||
     (platforms.length > 1 && !buildImageInfo.containerImageName) ||
     platforms.length === 0 ||
+    !isImageNameLowercase ||
```
axel7083: suggestion: we can probably remove `isImageNameLowercase` and have only `errorContainerImageName` as error mean invalid


```suggestion
    !errorContainerImageName ||
```

axel7083 (approved):
LGTM

axel7083 (approved):

vzhukovs:
@Mergifyio backport 1.20.x

Issue #12701, opened 2025-06-02
refactor(electron-builder): use official azure signing logic


axel7083:
### Is your enhancement related to a problem? Please describe

Following https://github.com/podman-desktop/podman-desktop/pull/12351 a new signing option aim to replace our custom logic for azure signing for electron-builder.

### Describe the solution you'd like

We should use the new azure option signing (ref https://www.electron.build/code-signing-win.html#using-azure-trusted-signing-beta)

benoitf:
I see that I'm assigned to the issue but currently we're using Azure keyvault, not Azure Trusted Signing so it's not a "deprecated API" usage, it would imply to move away from Digicerts where our certificates are.

it can't be done right now, it's not something that can be fixed from the Podman Desktop codebase yet

axel7083:
> it can't be done right now, it's not something that can be fixed from the Podman Desktop codebase yet

@benoitf we might update the deprecation indication I added, if you want I can make the change, we can also close this issue as not planned

benoitf:
yes probably need to update it as it's not really deprecated for now on our end

benoitf:
closing as there is no short plan to move away from Digicerts for signing the binaries

will update the notice

PR #13110, opened 2025-07-08
chore: update notice for azure code signing

benoitf:
### What does this PR do?
there is no plan to move away from Digicerts so method should not be deprecated

and keyvault != Trusted Signing so we can't use new electron builder methods

related to https://github.com/podman-desktop/podman-desktop/issues/12701

### How to test this PR?

### DIFF
--- a/.electron-builder.config.cjs
+++ b/.electron-builder.config.cjs
@@ -148,14 +148,10 @@ const config = {
       },
     ],
     /**
-     * TODO: replace with {@link import('electron-builder').Configuration#win#azureSignOptions}
-     * references:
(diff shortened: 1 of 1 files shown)


feloy (approved):

jeffmaury (approved):
LGTM

axel7083 (approved):
LGTM


question: should we open an upstream issue for electron-builder to support our use case ? is it something that may be interesting to share ?

benoitf:
you mean something like https://github.com/electron-userland/electron-builder/issues/7605 ?

axel7083:
> you mean something like [electron-userland/electron-builder#7605](https://github.com/electron-userland/electron-builder/issues/7605) ?

Probably

Issue #11434, opened 2025-02-27
Removing `*Sync` operation in Podman Desktop


axel7083:
### Is your enhancement related to a problem? Please describe

Sync operation such as `readFileSync`, `mkdirSync`, `rmSync` etc. are IO blocking, meaning they are blocking all the application when executing.

Those operation, mostly very fast (such as `existsSync`), adds up, and may be one of causes of the increasing time of Podman Desktop.

### Describe the solution you'd like

On the node documentation they provide some recommendations.

### Avoid race condition

We often do the following in our codebase 

https://github.com/podman-desktop/podman-desktop/blob/6c90794506b48feead9f10875275b8da240c2e95/packages/main/src/util.ts#L47-L48

https://github.com/podman-desktop/podman-desktop/blob/9076680fda276d029b26930a1c2681b8dd852211/extensions/docker/src/docker-context-handler.ts#L52-L56

etc.

We check if the file exist before reading it or interacting with it, but the node documentations say we should avoid doing that, and instead handle the error of the open, readFile etc.

> Do not use fs.access() to check for the accessibility of a file before calling fs.open(), fs.readFile() or fs.writeFile(). Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.[^1]

When this is not a problem with `existSync`, we may face issue switching to another method, so there is no _pure replacement solution here_

[^1]: https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback

### Additional context

There is a eslint plugin [eslint-plugin-n](https://github.com/eslint-community/eslint-plugin-n) for nodes, and one called [no-sync](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-sync.md) exists, that we could enforce progressively.

axel7083:
Interesting to take a look at the internal implementation (in C) of `existsSync`[^1]

```c
static void ExistsSync(const FunctionCallbackInfo<Value>& args) {
  Environment* env = Environment::GetCurrent(args);
  Isolate* isolate = env->isolate();
  CHECK_GE(args.Length(), 1);

  BufferValue path(isolate, args[0]);
  CHECK_NOT_NULL(*path);
  ToNamespacedPath(env, &path);
  THROW_IF_INSUFFICIENT_PERMISSIONS(
      env, permission::PermissionScope::kFileSystemRead, path.ToStringView());

  uv_fs_t req;
[...]

PR #13138, opened 2025-07-10
perf: remove exists call and handle error

simonrey1:
### What does this PR do?

Boosts performance by removing an unnecessary file exists call. This change also includes enhanced error handling for improved stability.

reference #11434 

This is the first in a series of multiple PRs addressing this area.

### DIFF
--- a/packages/api/src/is-no-entry-exception.ts
+++ b/packages/api/src/is-no-entry-exception.ts
@@ -1,0 +1,3 @@
+export function isNoEntryException(err: unknown): err is NodeJS.ErrnoException {
+  return err instanceof Error && 'code' in err && err.code === 'ENOENT';
+}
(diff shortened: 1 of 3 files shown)


jeffmaury (commented):
any is to be discarded in the project so err should be typed as unknown and test for checking code should be updated accordingly

simonrey1:
I see the error in the PR checks: 

`Error: packages/main/src/util.ts(55,9): error TS18046: 'error' is of type 'unknown'.`

I am missing anything in my VsCode config to see this error?

simonrey1:
The code I have added is similar to 

`isNoEntryException` from `kubernetes` plugin, I will try to refactor.

benoitf (commented):
not sure about performance boost over readability of the code

benoitf (changes requested):
looks like an anti pattern

Issue #13147, opened 2025-07-10
Add test for the monitoring loop of podman extension


simonrey1:
### Is your enhancement related to a problem? Please describe

Add test for the monitoring loop of podman extension

### Describe the solution you'd like

Add test for the monitoring loop of podman extension

PR #13153, opened 2025-07-10
fix(extensions/podman): avoid infinite loop when deactivating extension

### LINKED ISSUE #13149: Fix the podman extension monitoring loop to avoid infinite loop when deactivated
### Bug description

Fix the podman extension monitoring loop to avoid infinite loop when deactivated

### Operating system

all

### Installation Method

None

### Version

1.19.2

### Steps to reproduce

Exit the loop when deactivated

### Relevant log output

```shell

```

### Additional context

_No response_

simonrey1:
### What does this PR do?

It breaks the loop when deactivating the podman extension to avoid infinite loop. 

Like in docker extension:

Screenshot 2025-07-10 at 17 46 26

Fixes https://github.com/podman-desktop/podman-desktop/issues/13149

### How to test this PR?

I added a test and check that it was in error with the infinite loop if reverting the code of this PR.

### DIFF
--- a/extensions/podman/packages/extension/src/extension.spec.ts
+++ b/extensions/podman/packages/extension/src/extension.spec.ts
@@ -2712,15 +2712,7 @@ test('activate function returns an api implementation', async () => {
   vi.spyOn(PodmanInstall.prototype, 'checkForUpdate').mockResolvedValue({
     hasUpdate: false,
   } as unknown as UpdateCheck);
-  const contextMock = {
-    subscriptions: [],
-    secrets: {
-      delete: vi.fn(),
-      get: vi.fn(),
-      onDidChange: vi.fn(),
-      store: vi.fn(),
-    },
-  } as unknown as extensionApi.ExtensionContext;
+  const contextMock = getContextMock();
   const api = await extension.activate(contextMock);
   expect(api).toBeDefined();
   expect(typeof api.exec).toBe('function');
@@ -2730,15 +2722,7 @@ function mockExtensionForAuditTests() {
   vi.spyOn(PodmanInstall.prototype, 'checkForUpdate').mockResolvedValue({
     hasUpdate: false,
   } as unknown as UpdateCheck);
-  const contextMock = {
-    subscriptions: [],
-    secrets: {
-      delete: vi.fn(),
-      get: vi.fn(),
-      onDidChange: vi.fn(),
-      store: vi.fn(),
-    },
-  } as unknown as extensionApi.ExtensionContext;
+  const contextMock = getContextMock();
   vi.spyOn(provider, 'setContainerProviderConnectionFactory');
   return contextMock;
 }
@@ -2775,16 +2759,7 @@ test('activate on mac register commands for setting compatibility moide ', async
   vi.spyOn(PodmanInstall.prototype, 'checkForUpdate').mockResolvedValue({
     hasUpdate: false,
   } as unknown as UpdateCheck);
-  const contextMock = {
-    subscriptions: [],
-    secrets: {
-      delete: vi.fn(),
-      get: vi.fn(),
-      onDidChange: vi.fn(),
-      store: vi.fn(),
-    },
-  } as unknown as extensionApi.ExtensionContext;
-
+  const contextMock = getContextMock();
   // mock getSocketCompatibility
   const disableMock = vi.fn();
   const enableMock = vi.fn();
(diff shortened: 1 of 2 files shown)


axel7083 (commented):
unit test?

simonrey1:
> unit test?

It is done in the next sub issue https://github.com/podman-desktop/podman-desktop/issues/13147

https://github.com/podman-desktop/podman-desktop/pull/13172/files#diff-b260000c81f05bbed928d650b5b138a2361bbbdaeffce0bcb41c65a990763c03R3488

Edit: Sorry, the UT was related to the doMonitorProvider function and not monitorProvider.

axel7083:
> It is done in the next sub issue https://github.com/podman-desktop/podman-desktop/issues/13147

I would say, when changing something, you can include the unit tests with it, I would prefer to see the change and the corresponding test at the same time, than in separate PR

If you want to test the full function, this does not depends on the change introduced in this PR, so you may do that either before or after, but here as you fix something, you should have a regression test covering the specific change you made :+1:

simonrey1:
> > It is done in the next sub issue #13147
> 
> I would say, when changing something, you can include the unit tests with it, I would prefer to see the change and the corresponding test at the same time, than in separate PR
> 
> If you want to test the full function, this does not depends on the change introduced in this PR, so you may do that either before or after, but here as you fix something, you should have a regression test covering the specific change you made 👍

Sorry, the UT was related to the doMonitorProvider function and not monitorProvider.

I don't really know how to test this function because of the 8s sleep.

It was left untested in the past.

axel7083:
> I don't really know how to test this function because of the 8s sleep.
> 
> It was left untested in the past.

You should look at https://vitest.dev/guide/mocking.html#timers to better understand how to cover things with setTimeout

axel7083:
Weirdly enough, I can't see any files changes in the GitHub files tab

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -2317,3 +2317,7 @@ export function updateWSLHyperVEnabledContextValue(value: boolean): void {
     extensionApi.context.setValue(WSL_HYPERV_ENABLED_KEY, value);
   }
 }
+
+export function setStopLoop(value: boolean): void {
+  stopLoop = value;
```
axel7083: maybe the stopLoop should be reset in deactivate method

I am not a big fan of exposing a method for test purpose only :/

axel7083 (commented):
https://github.com/podman-desktop/podman-desktop/pull/13153/files#r2215571950

extensions/podman/packages/extension/src/extension.ts
simonrey1: I understand but I don't want to change the implementation of deactivate for test purpose

simonrey1: Oops sorry I thought adding stop loop in deactivate.

You meant, calling deactivate from the test, I can do it!

axel7083: When we deactivate an extension we want to dispose its resources, and stop its activity, when calling deactivate I expect the stopLoop to be true, and when reactivating I expect it to be set to false

simonrey1: Adding stopLoop = false in activate to make this change possible
https://github.com/podman-desktop/podman-desktop/pull/13275

extensions/podman/packages/extension/src/extension.spec.ts
```diff
@@ -385,6 +384,7 @@ afterEach(async () => {
   console.error = originalConsoleError;
   await extension.deactivate();
   vi.useRealTimers();
+  vi.resetAllMocks();
```
axel7083: We usually reset all mocks at the beginning of the beforeEach

extensions/podman/packages/extension/src/extension.spec.ts
```diff
@@ -3601,3 +3552,44 @@ describe('Check notify podman setup', () => {
...
+    expect(mockDoMonitorProvider).toHaveBeenCalledTimes(3);
+
+    await extension.deactivate();
+
+    await vi.runAllTimersAsync();
+    expect(mockDoMonitorProvider).toHaveBeenCalledTimes(3);
+  });
+});
+
+function getContextMock() {
```
gastoner, line 3585: I feel like help/setup functions should be on top, so we know what it is doing

gastoner (approved):
Codewise LGTM

extensions/podman/packages/extension/src/extension.spec.ts
simonrey1: I think this test file is too long to remember what was at the top, but I will take it into consideration when splitting it thanks

PR #13173, opened 2025-07-11
fix(ci): perform update of ubuntu packages earlier

simonrey1:
### What does this PR do?
Fixed problem with packages installation on ubuntu.
[[#13174 ](https://github.com/podman-desktop/podman-desktop/pull/13175)](https://github.com/podman-desktop/podman-desktop/issues/13174)

### How to test this PR?

### DIFF
--- a/.github/workflows/e2e-kubernetes-main.yaml
+++ b/.github/workflows/e2e-kubernetes-main.yaml
@@ -76,16 +76,16 @@ jobs:
       - name: Update podman
         run: |
           echo "ubuntu version from kubic repository to install podman we need (v5)"
-          ubuntu_version='23.10'
+          ubuntu_version='24.04'
           echo "Add unstable kubic repo into list of available sources and get the repo key"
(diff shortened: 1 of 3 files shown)


odockal:
@eqqe thanks for looking into this.
But removing these deps will probably break podman/criu installation. 
It might be fixed in https://github.com/podman-desktop/podman-desktop/pull/13175.

simonrey1:
@odockal indeed removing the deps as suggested in the chat did not work, I moved the update above but you also did in this PR

odockal:
> @odockal indeed removing the deps as suggested in the chat did not work, I moved the update above but you also did in this PR

It might be ok, to just move update. But using newer ubuntu 24.04 repos from kubic results in podman being 4.9.3. Instead of 5.2.5 which is from older repo. So older ubuntu version has newer podman version... :/

simonrey1:
Closing in favor of #13175

Issue #13174, opened 2025-07-11
Installation of dependency packages for podman installed from kubic repos requires updated ubuntu repos


odockal:
### Bug description

We got an error when installing some of the deps required for podman to be update on ubuntu.

```
 ubuntu version from kubic repository to install podman we need (v5)
Add unstable kubic repo into list of available sources and get the repo key
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).

100  1093  100  1093    0     0   2684      0 --:--:-- --:--:-- --:--:--  2685
OK
install necessary dependencies for criu package which is not part of 23.10
E: Failed to fetch mirror+file:/etc/apt/apt-mirrors.txt/pool/main/p/protobuf/libprotobuf32t64_3.21.12-8.2ubuntu0.1_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Failed to fetch mirror+file:/etc/apt/apt-mirrors.txt/pool/universe/p/protobuf/python3-protobuf_3.21.12-8.2ubuntu0.1_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

### Operating system

Ubuntu

### Installation Method

None

### Version

next (development version)

### Relevant log output

```shell

```

PR #13175, opened 2025-07-11
fix(ci): update ubuntu repositories before adding kubic repos

odockal:
### What does this PR do?
Fixed problem with packages installation on ubuntu.

#13174 

### How to test this PR?

### DIFF
--- a/.github/workflows/e2e-kubernetes-main.yaml
+++ b/.github/workflows/e2e-kubernetes-main.yaml
@@ -75,8 +75,8 @@ jobs:
 
       - name: Update podman
         run: |
-          # update using apt-get to get latest mirrors, etc
-          sudo apt-get update -qq
+          echo "Updating all dependencies..."
+          sudo apt-get update --allow-insecure-repositories -qq
           echo "ubuntu version from kubic repository to install podman we need (v5)"
           ubuntu_version='23.10'
           echo "Add unstable kubic repo into list of available sources and get the repo key"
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-kubernetes-main.yaml
```diff
@@ -75,8 +75,11 @@ jobs:
 
       - name: Update podman
         run: |
+          # update using apt-get to get latest mirrors, etc
+          sudo apt-get update -qq
           echo "ubuntu version from kubic repository to install podman we need (v5)"
           ubuntu_version='23.10'
+          sudo apt-get update
```
benoitf: issue: remove this line

odockal: nice catch, mistake. thx.

odockal:
@benoitf Looks like unit tests are quite unstable these days...

.github/workflows/e2e-kubernetes-main.yaml
simonrey1: You have the same issue in the other two yamls I think

odockal: One update was enough after all. so now I have move it up, should be good.

.github/workflows/e2e-kubernetes-main.yaml
```diff
@@ -75,6 +75,8 @@ jobs:
 
       - name: Update podman
         run: |
+          echo "Updating all dependencies..."
+          sudo apt-get update --allow-insecure-repositories -qq
```
simonrey1: Why did you add --allow-insecure-repositories?

axel7083: same question

odockal: Was in pr check whole time. I do not recall why. Not sure  that it is necessary since it worked fine here. note for future, leave more comments...

benoitf (approved):

.github/workflows/e2e-kubernetes-main.yaml
```diff
@@ -75,6 +75,8 @@ jobs:
 
       - name: Update podman
         run: |
+          echo "Updating all dependencies..."
```
benoitf: should be update database of packages or indexes of packages

.github/workflows/pr-check.yaml
```diff
@@ -321,6 +321,8 @@ jobs:
 
       - name: Update podman to 5.x
         run: |
+          echo "Updating database of packages..."
+          sudo apt-get update --allow-insecure-repositories -qq
```
simonrey1: In the other repositories I have done that before 


`echo "install necessary dependencies for criu package which is not part of ${ubuntu_version}"`

and the podman version displays 5

PR #13176, opened 2025-07-11
feat: added a warning when macOS user has multiple podman installations

### LINKED ISSUE #12888: podman desktop fails to use Homebrew's Podman in some situation
### Bug description

On macOS, I had a brain scratching experience with Podman installed by homebrew and Podman desktop.
Despite my efforts, Podman Desktop always failed to discover Podman and asked to set it up. And the setup would eventually fail I think at the VM creation time.

It turned out that for some reason I had a `/opt/podman` installation with v4 and `/opt/homebrew/Cellars/podman` with version 5
So my v5 images created from the command line would fail to be loaded by the /opt/podman version 4 with a 125 error code.

What tricked me is that I did not know about this /opt/podman version. It seems Podman Desktop prefers that path over the homebrew one and does not notify me of that multiple install conundrum with a warning. Some `which -a` is how I ended up finding the issue.

I classified as bug but could be enhancement.

### Operating system

macOS Sequoia 15.2

### Installation Method

None

### Version

1.17.2

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

MarsKubeX:
### What does this PR do?

When a macOS user has multiple Podman installations, a warning notification is displayed (as shown in the image below). However, if a custom binary path is configured in the Podman extension settings, this notification is suppressed.

### Screenshot / video of UI

Captura de pantalla 2025-07-08 a las 12 17 37

close #12888

### How to test this PR?
1. Install Podman with the official installer
2. Install Podman with Homebrew
3. Open Podman Desktop
4. The notification should appear as shown in the image above


extensions/podman/packages/extension/src/utils/podman-cli.ts
```diff
@@ -23,15 +25,14 @@ export function getInstallationPath(): string | undefined {
   const env = process.env;
   if (extensionApi.env.isWindows) {
     return `c:\\Program Files\\RedHat\\Podman;${env.PATH}`;
-  } else if (extensionApi.env.isMac) {
```
simonrey1: Else if is still relevant I think, to avoid checking the new if

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -794,7 +794,30 @@ function shouldNotifyQemuMachinesWithV5(installedPodman: InstalledPodman): boole
   return false;
 }
 
-async function monitorProvider(provider: extensionApi.Provider): Promise<void> {
+export async function getMultiplePodmanInstallationsMacosWarnings(
```
axel7083: question: can we have it outside of the `extension.ts`? like in a utils file?

The `extensions/podman/packages/extension/src/extension.ts` file has 2.3k lines.

And the corresponding spec file is also pretty long.

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -794,7 +794,30 @@ function shouldNotifyQemuMachinesWithV5(installedPodman: InstalledPodman): boole
...
+  installedPodman: InstalledPodman | undefined,
+): Promise<extensionApi.ProviderInformation[]> {
+  const warnings: extensionApi.ProviderInformation[] = [];
+
+  // Check for multiple Podman installations on macOS
+  if (extensionApi.env.isMac && installedPodman) {
```
axel7083: suggestion (non-blocking): I usually find early return more practical than big if block

extensions/podman/packages/extension/src/utils/podman-cli.ts
```diff
@@ -15,6 +15,8 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
+import * as fs from 'node:fs';
```
axel7083: question: why using `*` import ? usually recommended to only import what you need

extensions/podman/packages/extension/src/utils/podman-cli.ts
```diff
@@ -68,3 +69,33 @@ export async function getPodmanInstallation(): Promise<InstalledPodman | undefin
...
 }
+
+// Checks if there ara more than one version of podman installed in macos
+export async function isMultiplePodmanInstalledinMacos(): Promise<boolean> {
+  let isBrewInstalled = false;
+  let isDmgInstalled = false;
+
+  // Checks if custom binary path is set. If so, we don't need to check for multiple installations.
+  const customPath = getCustomBinaryPath();
+  console.log('customPath', customPath);
```
axel7083: question: to remove ?

extensions/podman/packages/extension/src/utils/podman-cli.ts
```diff
@@ -68,3 +69,33 @@
     return undefined;
   }
 }
+
+// Checks if there ara more than one version of podman installed in macos
+export async function isMultiplePodmanInstalledinMacos(): Promise<boolean> {
```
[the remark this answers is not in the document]
simonrey1: I like this refactoring also if you could test it :)

extensions/podman/packages/extension/src/extension.ts
MarsKubeX: I have moved this to util.ts. Also, for the tests, I've moved some of them to util.spec.ts and others to a new file podman-cli.spec.ts

extensions/podman/packages/extension/src/utils/podman-cli.ts
MarsKubeX: changed :)

extensions/podman/packages/extension/src/utils/podman-cli.ts
MarsKubeX: removed

extensions/podman/packages/extension/src/utils/podman-cli.ts
```diff
@@ -68,3 +71,20 @@ export async function getPodmanInstallation(): Promise<InstalledPodman | undefin
...
+
+// Checks if there are more than one version of podman installed in macos
+export async function isMultiplePodmanInstalledinMacos(): Promise<boolean> {
+  // Checks if custom binary path is set. If so, we don't need to check for multiple installations.
+  if (getCustomBinaryPath()) {
+    return false;
+  }
+  // Check if Podman is installed via Homebrew
+  try {
+    await extensionApi.process.exec('brew', ['list', '--verbose', 'podman'], {
```
dgolovin: It is not only macOS issue because on Windows I can install podman-cli package with chocolaty and that would be podman.exe in different location.

There is faster option to check for multiple installation that would eliminate brew, cocho and existsSync calls IMO. 

`where podman` for windows and `which podman` for everything else.

It would fail if there is no podman installed or return a line with full path for each podman executable in PATH. That would cover multiple platforms and all kinds of podman installations.

dgolovin (changes requested):
It is not only macOS issue. It is out of scope of the issue, but I think it worth to fix it.

extensions/podman/packages/extension/src/utils/util.ts
```diff
@@ -170,3 +170,31 @@ export function getProviderByLabel(label: string): string {
...
+    return warnings;
+  }
+  // Check for multiple Podman installations on macOS
+  try {
+    const hasMultiplePodmanInstallations = await isMultiplePodmanInstalledinMacos();
+    if (hasMultiplePodmanInstallations) {
+      warnings.push({
+        name: 'Multiple Podman installations detected',
+        details:
+          'You have Podman installed via both Homebrew and the official installer. This may cause conflicts. Consider removing one installation to avoid issues.',
```
dgolovin: Platform independent message would be

```suggestion
          'You have multiple Podman installations. This may cause conflicts. Consider leaving one installation or configure custom binary path in the Podman extension settings to avoid issues.',
```

extensions/podman/packages/extension/src/utils/util.ts
```diff
@@ -170,3 +170,31 @@ export function getProviderByLabel(label: string): string {
...
+export async function getMultiplePodmanInstallationsMacosWarnings(
+  installedPodman: InstalledPodman | undefined,
+): Promise<extensionApi.ProviderInformation[]> {
+  const warnings: extensionApi.ProviderInformation[] = [];
+  if (!extensionApi.env.isMac || !installedPodman) {
+    return warnings;
+  }
+  // Check for multiple Podman installations on macOS
+  try {
+    const hasMultiplePodmanInstallations = await isMultiplePodmanInstalledinMacos();
```
dgolovin: It looks like isMultiplePodmanInstalledInMacos does not throw exceptions, so try catch is not required.

dgolovin (changes requested):

extensions/podman/packages/extension/src/utils/podman-cli.ts
MarsKubeX: Changed now to check also windows and linux. For linux and MacOS I have used type `type -a podman,` because `which podman` returns only the first result.

MarsKubeX:
@dgolovin I have changed the code to check also windows and linux.

dgolovin (approved):
Tested on mac. LGTM.

nit: I did not realize `which podman` would return only first found command, but `which -a podman` would return all of them without `podman is` prefix compare to `type -a podman`.  `which` is also external command not a `built-in` shell command. 

What I found `which -a podman` gives the same output on linux, but with aliases printed on top. `--skip-alias` removes aliases part on linux, but is not supported on macOS. 

Is there any other reason to prefer `type` over `which`?

MarsKubeX:
@dgolovin thanks for the comment. I've chosen `type -a` because it was built-in, but is true that for this case I can use both.

extensions/podman/packages/extension/src/utils/podman-cli.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+vi.mock('@podman-desktop/api', () => {
+  return {
+    configuration: {
+      getConfiguration: (): Configuration => config,
+    },
+    process: {
+      exec: vi.fn(),
+    },
+    env: {
+      isWindows: false, // Default to Linux/macOS
```
simonrey1: please choose one isMac or isLinux to true, even if the code checks that it is not windows, it is not clear what we are testing here

extensions/podman/packages/extension/src/utils/util.ts
```diff
@@ -170,3 +170,27 @@ export function getProviderByLabel(label: string): string {
...
+ * Checks for multiple Podman installations and returns a warning if found.
+ * @param installedPodman - The installed Podman version.
+ * @returns A promise that resolves to an array of ProviderInformation objects.
+ */
+export async function getMultiplePodmanInstallationsWarnings(
+  installedPodman: InstalledPodman | undefined,
+): Promise<extensionApi.ProviderInformation[]> {
+  const warnings: extensionApi.ProviderInformation[] = [];
+  if (!installedPodman) {
+    return warnings;
```
simonrey1: suggestion: `return []`

extensions/podman/packages/extension/src/utils/util.ts
```diff
@@ -170,3 +170,27 @@ export function getProviderByLabel(label: string): string {
...
+
+/**
+ * Checks for multiple Podman installations and returns a warning if found.
+ * @param installedPodman - The installed Podman version.
+ * @returns A promise that resolves to an array of ProviderInformation objects.
+ */
+export async function getMultiplePodmanInstallationsWarnings(
+  installedPodman: InstalledPodman | undefined,
+): Promise<extensionApi.ProviderInformation[]> {
+  const warnings: extensionApi.ProviderInformation[] = [];
```
simonrey1: suggestion: to remove

extensions/podman/packages/extension/src/utils/util.ts
```diff
@@ -170,3 +170,27 @@ export function getProviderByLabel(label: string): string {
...
+  // Check for multiple Podman installations
+  const hasMultiplePodmanInstallations = await isMultiplePodmanInstalled();
+  if (hasMultiplePodmanInstallations) {
+    warnings.push({
+      name: 'Multiple Podman installations detected',
+      details:
+        'You have multiple Podman installations. This may cause conflicts. Consider leaving one installation or configure custom binary path in the Podman extension settings to avoid issues.',
+    });
+  }
+  return warnings;
```
simonrey1: suggestion: `return []`

extensions/podman/packages/extension/src/utils/util.ts
```diff
@@ -170,3 +170,27 @@ export function getProviderByLabel(label: string): string {
...
+  installedPodman: InstalledPodman | undefined,
+): Promise<extensionApi.ProviderInformation[]> {
+  const warnings: extensionApi.ProviderInformation[] = [];
+  if (!installedPodman) {
+    return warnings;
+  }
+  // Check for multiple Podman installations
+  const hasMultiplePodmanInstallations = await isMultiplePodmanInstalled();
+  if (hasMultiplePodmanInstallations) {
+    warnings.push({
```
simonrey1: suggestion: `
const warning: extensionApi.ProviderInformation = {...};
return [warning];`

extensions/podman/packages/extension/src/utils/podman-cli.ts
```diff
@@ -68,3 +103,14 @@ export async function getPodmanInstallation(): Promise<InstalledPodman | undefin
...
+
+// Checks if there are more than one version of podman installed (Windows, macOS, Linux)
+export async function isMultiplePodmanInstalled(): Promise<boolean> {
+  // Checks if custom binary path is set. If so, we don't need to check for multiple installations.
+  if (getCustomBinaryPath()) {
+    return false;
+  }
+  const installations = await findPodmanInstallations();
+  // If there are 2 or more unique installations, return true
+  return installations.length > 1;
```
simonrey1, line 110: Why do we extract the exact paths with complex `replace` and `trim`calls but never use it? Should we just check if there are multiple lines?

MarsKubeX: You're right, thanks for the comment.

extensions/podman/packages/extension/src/extension.ts
```diff
@@ -702,6 +703,10 @@ export async function doMonitorProvider(provider: extensionApi.Provider): Promis
     const installedPodman = await getPodmanInstallation();
     provider.updateDetectionChecks(getDetectionChecks(installedPodman));
 
+    // get warnings for multiple Podman installations
```
simonrey1: IMHO this comment is not useful because it is exactly the function name

extensions/podman/packages/extension/src/utils/podman-cli.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    configuration: {
+      getConfiguration: (): Configuration => config,
+    },
+    process: {
+      exec: vi.fn(),
+    },
+    env: {
+      isWindows: false,
+      isMac: true,
+      isLinux: true,
```
simonrey1: remark: isMac and isLinux at the same time should not be possible

extensions/podman/packages/extension/src/utils/podman-cli.ts
```diff
@@ -15,10 +15,40 @@
...
+    let result: extensionApi.RunResult;
+    if (extensionApi.env.isWindows) {
+      // Windows: Use 'where podman' command
+      result = await extensionApi.process.exec('where', ['podman']);
+    } else {
+      // Unix/macOS: Use 'type -a podman' command
+      result = await extensionApi.process.exec('sh', ['-c', 'type -a podman']);
+    }
+
+    // Remove duplicates and return array of unique lines (installations)
```
simonrey1, line 38: Is it possible that these commands return duplicates?

MarsKubeX: Yes. I've logged this result and had duplicates.

simonrey1 (approved):
LGTM, not tested on my machine

Issue #11657, opened 2025-03-13
Create a GitHub Action that installs or updates Podman


odockal:
### Is your enhancement related to a problem? Please describe

Idea is to prepare a single github-action in its dedicated repository ie. `podman-action` under `containers` org. which installs/upgrades the podman on the gha runners and allows user to start up podman machine.

### Describe the solution you'd like

Create a repository `containers/podman-action`) and prepare a github action that would
* Installs podman (from setup.exe file) on windows
* installs/upgrades the podman if installed (usually the case for ubuntu runners)
* initialize and starts the podman machine (windows)

Parametrization could offers to configure the podman machine, pick up podman version, etc.

PR #13178, opened 2025-07-11
test(ci): check major podman version is 5

simonrey1:
### What does this PR do?

This PR checks the podman version is the major 5 in the CI when updating it in ubuntu.
Because we have seen that in version 24.04 of Ubuntu today the major version was 4, so the goal of the task
"Update podman to 5.x" was not obtained. 

### How to test this PR?


benoitf (commented):
IMHO I would stop doing copy/paste and bring https://github.com/podman-desktop/podman-desktop/issues/11657

Issue #12452, opened 2025-05-12
Store GitHub stars count on website push rather than dynamically.


cdrage:
### Is your enhancement related to a problem? Please describe

Follow up from: https://github.com/podman-desktop/podman-desktop/issues/12118

### Describe the solution you'd like

Instead of getting the GitHub stars on each load (website may rate limit requests to GitHub), we should instead get the star count at build time and store it (`src/data/stars.json`) to be used.

### Additional context

Thanks @axel7083 for the suggestion here: https://github.com/podman-desktop/podman-desktop/pull/12429#pullrequestreview-2832197876

PR #13182, opened 2025-07-11
fix(website): client-side use of the GitHub REST API should be avoided

### LINKED ISSUE #10554: artifact links from download section if GitHub API limit is reached

today, it's using the GitHub API to query the files to be downloaded (latest version, etc)

but if you're on a shared network or are reaching the GitHub API limit you won't be able to grab/download the file

1. avoid to use GitHub API on client side (for example download the latest version from a different file) (like a JSON file hosted on the website)
1. or in case of failure, redirect to the GitHub release page
1. create the links offline (when building the website). It means that a new version of the site would need to be published in that case to upload the download links

_No response_

### Additional context

_No response_

simonrey1:
### What does this PR do?

This PR introduces a docusaurus plugin to download the latest release information while building the website instead of doing it client side to avoid limitation in GitHub REST API, especially for clients sharing an IP (companies...).

Fixes #10554

### How to test this PR?


website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,206 @@
...
+// Type definition for the relevant parts of the GitHub release API response
+interface GitHubRelease {
+  tag_name?: string;
+  name: string;
+  assets: GitHubAsset[];
+}
+
+// --- Data type definitions for each OS ---
+
+export interface LinuxDownloadData {
```
axel7083: suggestion(non-blocking): I like abstraction, just to mention that we could have an interface DownloadData with a field version, and the Linux & Windows & MacOs could extend it

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,206 @@
...
+  }
+  return asset.browser_download_url;
+}
+
+// The main plugin function, now strongly typed
+export default async function githubReleasePlugin(): Promise<Plugin<AllDownloadsData | null>> {
+  return {
+    name: 'docusaurus-plugin-github-release',
+
+    async loadContent(): Promise<AllDownloadsData | null> {
```
axel7083: Question: why having null as return value accepted? It seems that you throw Error in most cases anyway

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,206 @@
...
+      console.log('Fetching latest GitHub release for podman-desktop...');
```
axel7083: suggestion: to remove?

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,206 @@
...
+
+      try {
+        const response = await fetch('https://api.github.com/repos/containers/podman-desktop/releases/latest');
+        if (!response.ok) {
+          throw new Error(`Failed to fetch latest release: ${response.statusText}`);
+        }
+        const releaseData = (await response.json()) as GitHubRelease;
```
axel7083: Wondering about the Octokit usage instead of raw request to the API directly

axel7083 (commented):
Waou ! Great job, gave some nitpick feedback, will do a better review/ test Monday

website/src/plugins/github-release-plugin.ts
simonrey1: 772ad719a26c59564f7e3c649f07028d56735ec4

website/src/plugins/github-release-plugin.ts
simonrey1: This console.log is displayed only at build time for the website so I could leave it here to help debug the CI in case GitHub request does not work.

website/src/plugins/github-release-plugin.ts
simonrey1: Thanks for the suggestion.
I can give it a try and maybe I could remove these interfaces if the result is already typed
Screenshot 2025-07-15 at 09 18 35

simonrey1:
In this PR, I have kept the the search for assets in the list. But I was wondering if we could just fetch the latest version and build the assets name directly.

So instead of: 
Screenshot 2025-07-15 at 09 19 52

We would do a string interpolation:
`podman-desktop-${version}.flatpak` 

@axel7083 after reading this PR, what do you think?

website/src/plugins/github-release-plugin.ts
simonrey1: 581c8c4e1b671546e37a18b8fde4555be6354e0f

website/src/plugins/github-release-plugin.ts
simonrey1: 803318cc6812ec293005392a6b0756cc14a67cf6

Thanks! I removed it and removed the duplicated interface.

website/package.json
```diff
@@ -25,8 +25,8 @@
     "@docusaurus/core": "^3.7.0",
     "@docusaurus/plugin-client-redirects": "^3.7.0",
     "@docusaurus/preset-classic": "^3.7.0",
-    "@docusaurus/theme-mermaid": "^3.7.0",
```
benoitf: if you want to sort the file, please do a separate PR

else this line should not be touched

website/src/pages/downloads/linux.tsx
```diff
@@ -1,81 +1,22 @@
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import { usePluginData } from '@docusaurus/useGlobalData';
 import { faLinux } from '@fortawesome/free-brands-svg-icons';
 import { faDownload, faPaste, faTerminal } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import TailWindThemeSelector from '@site/src/components/TailWindThemeSelector';
 import { TelemetryLink } from '@site/src/components/TelemetryLink';
+import type { GlobalData } from '@site/src/plugins/github-release-plugin'; // Adjust path if needed
```
benoitf:
```suggestion
import type { GlobalData } from '@site/src/plugins/github-release-plugin';
```

why this comment ? it looks like a copy/paste from a tutorial or AI generated stuff

website/src/pages/downloads/macos.tsx
```diff
@@ -1,112 +1,21 @@
...
 export function MacOSDownloads(): JSX.Element {
-  const [downloadData, setDownloadData] = useState({
-    version: '',
-    universal: '',
-    x64: '',
-    arm64: '',
-    airgapsetupX64: '',
-    airgapsetupArm64: '',
-  });
+  const { macosDownloads } = usePluginData('docusaurus-plugin-github-release') as GlobalData;
```
benoitf: IMHO the correct way would be to use generics
```js
const { macosDownloads } = usePluginData<Something>('docusaurus-plugin-github-release')
```

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+  binaryArm64: string;
+  setupX64: string;
+  setupArm64: string;
+  airgapsetupX64: string;
+  airgapsetupArm64: string;
+}
+
+// --- Combined data types ---
+
+export interface GlobalData {
```
benoitf: GlobalData does not like meaningful as a wording

it's for example metrics or data from GitHub releases

something like GitHubReleaseData looks more helpful to understand what we could get

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+  linuxDownloads: LinuxDownloadData;
+  macosDownloads: MacosDownloadData;
+  windowsDownloads: WindowsDownloadData;
```
benoitf: I would opt in for a struct like

```yaml
{
release: {
   windows:
   linux:
   macOS:
  }
}
```

website/src/plugins/github-release-plugin.ts
benoitf: as we will need for https://github.com/podman-desktop/podman-desktop/issues/12452 the stars metrics

so I would avoid to use "release" in the name

```json
GitHubMetadata : {
   latestRelease : {
   }
   stars: 123
}
```

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// The main plugin function, now strongly typed
+export default async function githubReleasePlugin(): Promise<Plugin<GlobalData>> {
+  return {
+    name: 'docusaurus-plugin-github-release',
+
+    async loadContent(): Promise<GlobalData> {
+      console.log('Fetching latest GitHub release for podman-desktop using Octokit...');
```
benoitf:
```suggestion
```

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
+import type { Plugin } from '@docusaurus/types';
```
benoitf: rename the plug-in by dropping the name "release" from it as it'll probably be used for https://github.com/podman-desktop/podman-desktop/issues/12452

something more generic about stuff we can get from GitHub
github-metrics-plugin, github-fetch-data-plugin, github-metadata-plugin, etc.

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+      try {
+        const { data: releaseData } = await octokit.rest.repos.getLatestRelease({
+          owner: 'containers',
+          repo: 'podman-desktop',
+        });
+
+        const { tag_name, assets } = releaseData;
+
+        if (!tag_name) {
+          throw new Error(`No tag name for latest release`);
```
benoitf: what happen in that case, the website is not built ?

benoitf: it should say latest release of what

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+    name: 'docusaurus-plugin-github-release',
+
+    async loadContent(): Promise<GlobalData> {
+      console.log('Fetching latest GitHub release for podman-desktop using Octokit...');
+
+      const octokit = new Octokit();
+
+      try {
+        const { data: releaseData } = await octokit.rest.repos.getLatestRelease({
+          owner: 'containers',
```
benoitf: the owner is not correct, we're now in podman-desktop org (there is still an alias from the old ownership but new code should use the new org name)

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+export default async function githubReleasePlugin(): Promise<Plugin<GlobalData>> {
+  return {
+    name: 'docusaurus-plugin-github-release',
+
+    async loadContent(): Promise<GlobalData> {
+      console.log('Fetching latest GitHub release for podman-desktop using Octokit...');
+
+      const octokit = new Octokit();
+
+      try {
```
benoitf: fetching release should be a dedicated method (easier to have unit tests)

and easier when we would plug something like fetchStars() for https://github.com/podman-desktop/podman-desktop/issues/12452

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+        // Helper function to find an asset or throw an error
+        function findAssetOrThrow(predicate: (asset: (typeof assets)[0]) => boolean, assetName: string): string {
+          const asset = assets.find(predicate);
+          if (!asset) {
+            throw new Error(`Required asset not found: ${assetName}`);
+          }
+          return asset.browser_download_url;
+        }
+
+        // --- Linux ---
```
benoitf: what is the format of this comment ?

```js
// --- Linux ---
```

should it only be
```
// Linux
```

or something more descriptive

```js
// handle Linux download links
```

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+        );
+
+        const linuxDownloads: LinuxDownloadData = {
+          version,
+          flatpak: flatpakUrl,
+          arm64: linuxArmUrl,
+          amd64: linuxAmdUrl,
+        };
+
+        // --- macOS ---
```
benoitf: change scheme for the comment's name

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+        const macosDownloads: MacosDownloadData = {
+          version,
+          universal: macosUniversalUrl,
+          x64: macosX64Url,
+          arm64: macosArm64Url,
+          airgapsetupX64: macosAirgapX64Url,
+          airgapsetupArm64: macosAirgapArm64Url,
+        };
+
+        // --- Windows ---
```
benoitf: change scheme of comment

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+          version,
+          setupX64: windowsSetupX64Url,
+          setupArm64: windowsSetupArm64Url,
+          binaryX64: windowsPortableX64Url,
+          binaryArm64: windowsPortableArm64Url,
+          airgapsetupX64: windowsAirgapSetupX64Url,
+          airgapsetupArm64: windowsAirgapSetupArm64Url,
+        };
+
+        console.log(`Successfully fetched version ${version} for all platforms.`);
```
benoitf:
```suggestion
```

we don't care, if it's not successful it should have thrown an error

benoitf (dismissed):
dropped some comments

blocker: plugin require unit tests (for example https://mswjs.io/ is used in unit tests to emulate calls to a remote REST server)

website/src/plugins/github-release-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+export interface WindowsDownloadData extends DownloadData {
+  binaryX64: string;
+  binaryArm64: string;
+  setupX64: string;
+  setupArm64: string;
+  airgapsetupX64: string;
+  airgapsetupArm64: string;
+}
+
+// --- Combined data types ---
```
benoitf: either keep with something helpful or drop the comment (and drop this style with --- )

simonrey1: 87bc4220e48345790ff376b31a7af30f9f2f089e

website/src/plugins/github-release-plugin.ts
simonrey1: https://github.com/podman-desktop/podman-desktop/commit/87bc4220e48345790ff376b31a7af30f9f2f089e

website/src/plugins/github-release-plugin.ts
simonrey1: https://github.com/podman-desktop/podman-desktop/commit/87bc4220e48345790ff376b31a7af30f9f2f089e

website/src/plugins/github-release-plugin.ts
simonrey1: https://github.com/podman-desktop/podman-desktop/commit/87bc4220e48345790ff376b31a7af30f9f2f089e

website/src/plugins/github-release-plugin.ts
simonrey1: 8ab060dc48e6798d9f35c154d53fffa4b68c5719 and fb35d3f42bba654969972b76307a43cf3f6df321

website/package.json
simonrey1: 9f54f8f82cf8914f4af26691509e6d44e850b0e8

website/src/plugins/github-release-plugin.ts
simonrey1: fb35d3f42bba654969972b76307a43cf3f6df321

website/src/plugins/github-release-plugin.ts
simonrey1: Screenshot 2025-07-16 at 09 08 45

Yes it makes the command `pnpm run website:build` fail, so team can investigate

simonrey1: ce2c4de9d7a4fc75a8becbf543561327e8c029c6

Failed to retrieve tag name for the latest Podman Desktop release from GitHub. The 'tag_name' field was missing in the release data.

website/src/plugins/github-release-plugin.ts
simonrey1: 8cd0623bf779d4f782f000f72d47286dbc9ff3a5

website/src/plugins/github-release-plugin.ts
simonrey1: it was removed eventually

website/src/plugins/github-release-plugin.ts
simonrey1: I have moved it to a separate class and file, it can be extended to add a method to retrieve stars.
I will add unit test of this class

website/src/plugins/github-release-plugin.ts
benoitf: hello, adding commits links is really nice but FYI I'm fine if you just flag conversation as resolved, no need to point me each commit with a new comment, I can see them anyway in the commit history/diff if it can save you some time.

website/src/pages/downloads/linux.tsx
```diff
@@ -1,81 +1,22 @@
...
 
 export function LinuxDownloads(): JSX.Element {
-  const [downloadData, setDownloadData] = useState({
-    version: '',
-    arm64: '',
-    amd64: '',
-    flatpak: '',
-  });
+  // Use the generic version of usePluginData for full type safety
+  const { linuxDownloads } = usePluginData('docusaurus-plugin-github-metadata') as GitHubMetadata;
```
benoitf:
```suggestion
  // Use the generic version of usePluginData for full type safety
  const { linuxDownloads } = usePluginData('docusaurus-plugin-github-metadata') as GitHubMetadata;
```

I commented somewhere else but it should not use `as GithubMetadata` but the generics but really use it 
also the comment should be dropped as the comment is not doing what is expected

```ts
  const { linuxDownloads } = usePluginData<GitHubMetadata>('docusaurus-plugin-github-metadata');
```

website/src/plugins/GitHubMetadata.ts
```diff
@@ -0,0 +1,32 @@
+interface DownloadData {
```
benoitf: filename should all be lowercase `github-metadata.ts`

website/src/plugins/GitHubMetadata.ts
```diff
@@ -0,0 +1,32 @@
...
+  setupArm64: string;
+  airgapsetupX64: string;
+  airgapsetupArm64: string;
+}
+
+export interface GitHubMetadata {
+  linuxDownloads: LinuxDownloadData;
+  macosDownloads: MacosDownloadData;
+  windowsDownloads: WindowsDownloadData;
+}
```
benoitf: I commented previously it should contains sub-fields like `latestRelease` then attributes `linux`, `windows` etc

website/src/plugins/GitHubService.ts
```diff
@@ -0,0 +1,135 @@
+import { Octokit } from '@octokit/rest';
```
benoitf: filename: github-service.ts

website/src/pages/downloads/macos.tsx
```diff
@@ -1,112 +1,21 @@
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import { usePluginData } from '@docusaurus/useGlobalData';
```
axel7083, line 20: Nice method; didn't know it exists, from the top of memory the Storybook plugin generate a JSON file, and the related components imports it, but this is nice

axel7083:
> In this PR, I have kept the the search for assets in the list. But I was wondering if we could just fetch the latest version and build the assets name directly.
> [...]
> We would do a string interpolation: `podman-desktop-${version}.flatpak`
> 
> @axel7083 after reading this PR, what do you think?

The name of the assets are always the same, we could directly do the string interpolation instead of the search, and throw an error if they are no available

axel7083:
> > In this PR, I have kept the the search for assets in the list. But I was wondering if we could just fetch the latest version and build the assets name directly.
> > [...]
> > We would do a string interpolation: `podman-desktop-${version}.flatpak`
> > @axel7083 after reading this PR, what do you think?
> 
> The name of the assets are always the same, we could directly do the string interpolation instead of the search, and throw an error if they are no available

This can be a separate PR, as this is already pretty big

website/src/plugins/GitHubMetadata.ts
simonrey1: b9d4af8730e6096489b836aebfbc90b05c09e9fc

website/src/plugins/GitHubService.ts
simonrey1: b9d4af8730e6096489b836aebfbc90b05c09e9fc

website/src/pages/downloads/linux.tsx
simonrey1: Yes the comment is wrong, I tried to do this at first too but I have an error that I am unable to solve:

Screenshot 2025-07-16 at 10 00 33

website/src/plugins/test-resourcecs/mock-release-data.ts
```diff
@@ -0,0 +1,1718 @@
+export const mockReleaseData = {
```
axel7083: question: if there is no type, can't we just have this in a json file please?

website/src/plugins/github-service.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import type { GitHubMetadata } from './github-metadata';
+
+export class GitHubService {
+  private octokit: Octokit;
+  private owner: string;
+  private repo: string;
+
+  constructor(owner: string, repo: string) {
+    this.octokit = new Octokit();
```
axel7083, line 29: suggestion: Maybe we should be able to pass the Octokit as a parameter, so we can jsut mock it without having it to mock the real request ? or have it as an optional constructor parameter so tests can specify it

website/src/plugins/github-service.ts
```diff
@@ -0,0 +1,113 @@
+import { Octokit } from '@octokit/rest';
```
axel7083, line 19: missing copyright

website/src/plugins/github-service.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+            'https://github.com/podman-desktop/podman-desktop/releases/download/v1.19.2/podman-desktop-1.19.2-setup-x64.exe',
+        },
+      },
+    };
+
+    const metadata = await service.getLatestReleaseMetadata();
+    expect(metadata).toEqual(expectedMetadata);
+  });
+
+  it('should throw an error if tag_name is missing', async () => {
```
axel7083: remark: we use `test` instead of `it` in most of the codebase

website/src/plugins/github-service.spec.ts
```diff
@@ -0,0 +1,123 @@
+import { http, HttpResponse } from 'msw';
```
axel7083, line 19: missing copyright

website/src/plugins/github-metadata.ts
```diff
@@ -0,0 +1,25 @@
+export interface GitHubMetadata {
```
axel7083, line 19: missing copyright

axel7083 (commented):
Left some comments

website/src/plugins/test-resourcecs/mock-release-data.ts
simonrey1: I added the type, let me know if you still prefer a json file

website/src/plugins/github-service.ts
simonrey1: I was following Florent's comment here: https://github.com/podman-desktop/podman-desktop/pull/13182#pullrequestreview-3022059612 but I agree that both solutions work

axel7083: cc @benoitf IMO we should use provide the octokit client through the constructor, it would avoid adding `msw` as a dependency, and only work with mocks instead of having to mock networks call, and would simplify the tests

website/src/plugins/test-resourcecs/mock-release-data.ts
axel7083: I like the typed data

website/src/plugins/github-service.ts
benoitf: it's always the tricky question of mocks and usage

Here it's an easy option to mock the data and still use the library that gets the data (Octokit)

vs

you mock everything

At the end, you can have unit tests that pass but it doesn't work at all, so I like somehow to validate the real calls made by Octokit

benoitf: if we update Octokit library I would assume one test is checking that the value returned by the upgrade will still pass while mocking its behavior

axel7083 (approved):

website/src/plugins/github-service.ts
axel7083: I don't have a strong opinion on that, so approving as it is

website/src/pages/downloads/windows.tsx
```diff
@@ -1,112 +1,42 @@
...
-    binaryX64: '',
-    setupX64: '',
-    airgapsetupX64: '',
-    binaryArm64: '',
-    setupArm64: '',
-    airgapsetupArm64: '',
-  });
+  const {
+    latestRelease: { windows, version },
+  } = usePluginData('docusaurus-plugin-github-metadata') as GitHubMetadata;
```
benoitf, line 34: usePluginData with generics rather than 'cast' with as

website/src/plugins/github-service.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+
+const server = setupServer(
+  http.get(URL, () => {
+    return HttpResponse.json(mockReleaseData);
+  }),
+);
+
+describe('GitHubService', () => {
+  const owner = 'podman-desktop';
+  const repo = 'podman-desktop';
```
benoitf: I think it should be constants because 'podman-desktop' for org and 'podman-desktop' for repository name are like duplicated tons of times across these new files

let say tomorrow I change the org name, I should not have to change at multiple places

website/src/plugins/github-metadata-plugin.ts
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Plugin } from '@docusaurus/types';
+
+import type { GitHubMetadata } from './github-metadata';
+import { GitHubService } from './github-service';
+
+export default async function githubMetadataPlugin(): Promise<Plugin<GitHubMetadata>> {
+  const githubService = new GitHubService('podman-desktop', 'podman-desktop');
```
benoitf: IMHO it could reuse the stuff that is defined for the project there

https://github.com/podman-desktop/podman-desktop/blob/main/package.json#L4

as this is where we declare the repository of Podman Desktop

website/src/plugins/test-resourcecs/mock-release-data.ts
```diff
@@ -0,0 +1,1692 @@
+/**********************************************************************
```
benoitf: the folder name "test-resourcecs"

I would guess there is typo in the folder name ?

website/vitest.config.js
```diff
@@ -19,7 +19,7 @@
 const config = {
   test: {
     environment: 'node',
-    include: ['*.{test,spec}.?(c|m)[jt]s?(x)'],
+    include: ['**/*.{test,spec}.?(c|m)[jt]s?(x)'],
```
benoitf: you could add this change as a separate PR (merged quicly)

website/src/pages/downloads/windows.tsx
simonrey1: https://github.com/podman-desktop/podman-desktop/pull/13182#discussion_r2209600497

Let's continue the discussion in this comment if possible. I tried to use the generics but it does not compile.

website/src/plugins/test-resourcecs/mock-release-data.ts
simonrey1: Thanks for your attention :) It was a typo, I changed it to test/resources because test-resources was in the .gitignore

website/vitest.config.js
simonrey1: https://github.com/podman-desktop/podman-desktop/pull/13230/files

website/src/plugins/github-metadata-plugin.ts
simonrey1: Interesting but Octokit requires owner and repo to be parsed so I don't know if we can reuse without a hack?

website/src/pages/downloads/windows.tsx
benoitf: ah ignore my comment if they don't provide generics, I thought they were

website/src/plugins/github-metadata-plugin.ts
benoitf: https://www.npmjs.com/package/hosted-git-info

simonrey1: Ok I gave it a try but I am having trouble to read the repository from package.json.
In npm it is suggested that we can access these variables in process.env.npm_package_*
But here maybe with pnpm it does not work, so I would have to open a file to read those contants it seems expansive to me.

simonrey1: Screenshot 2025-07-16 at 16 09 57

benoitf: import rootPackage from '../..I don't know how many ../../../package.json' with { type: 'json' };

then fromUrl(rootPackage.repository)

simonrey1: https://github.com/podman-desktop/podman-desktop/pull/13232

website/src/plugins/github-metadata-plugin.ts
```diff
@@ -18,11 +18,12 @@
 
 import type { Plugin } from '@docusaurus/types';
 
+import { GITHUB_OWNER, GITHUB_REPOSITORY } from '../../../packages/api/src/repository-infos';
```
benoitf, line 21: I would suggest `/@api/....` import (if missing can be added separately)

avoiding relative imports ensure we don't try to import code that we don't want to use (for example importing a server file from a client file)

simonrey1: Yes i wanted to use that but it is not configured in the tsconfig.json, so it should be done in a different PR.

I hope it is not blocking for this one :)

benoitf: IMHO it's a blocker

simonrey1: I tried to use the path alias but I could not manage to make it work with docusaurus build.

I tried to override the webpack config in docusaurus plugin but the problem is happening at the module load so it's too late.

I tried to pass JITI_ALIAS env variable to define path aliases in jiti used by docusaurus but it did not work either.

website/src/plugins/github-metadata-plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Plugin } from '@docusaurus/types';
+
+import { GITHUB_OWNER, GITHUB_REPOSITORY } from '../../../packages/api/src/repository-infos';
```
jeffmaury, line 21: Why not using /@api/ here ?

website/src/plugins/github-service.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+import { setupServer } from 'msw/node';
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, test } from 'vitest';
+
+import { GITHUB_OWNER, GITHUB_REPOSITORY } from '/@api/repository-infos';
+
+import type { GitHubMetadata } from './github-metadata';
+import { GitHubService } from './github-service';
+import { mockReleaseData } from './test/resources/mock-release-data';
+
+const URL = 'https://api.github.com/repos/podman-desktop/podman-desktop/releases/latest';
```
jeffmaury: Wondering why this is not using the GITHUB_OWNER and GITHUB_REPOSITORY here ?

jeffmaury (commented):
See comments

website/src/plugins/github-metadata-plugin.ts
simonrey1: We are discussing it here: https://github.com/podman-desktop/podman-desktop/pull/13182#discussion_r2212868679

But the reason is that it is a "docusaurus" plugin and this module does not handle path aliases it seems. I have searched for hours.

cdrage: @jeffmaury I am encountering similar issues too trying to upgrade to 3.8.0 docusaurus and path aliases not being followed correctly... 

```
(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
✓ built in 5.06s
info => Output directory: /Users/cdrage/syncthing/dev/electron/podman-desktop/storybook/storybook-static

> docs@0.0.0 docusaurus /Users/cdrage/syncthing/dev/electron/podman-desktop/website
> docusaurus "clear"

node:internal/modules/cjs/loader:1225
  const err = new Error(message);
              ^

Error: Cannot find module 'image-size/fromFile'
Require stack:
- /Users/cdrage/syn
```

So same as @eqqe have to use `../../../` instead which is annoying, but required at the moment.

I have a feeling it's the way that they are managing modules within docusaurus as there's other issues on different repos opened regarding it: https://github.com/search?q=Error%3A+Cannot+find+module+%27image-size%2FfromFile%27&type=issues

jeffmaury (approved):
LGTM

PR #13192, opened 2025-07-14
feat: update notification style on dashboard

### LINKED ISSUE #13106: Implement updated design of notifcations on the dashboard

The notifications on the dashboard don't follow the same look and feel as notications we added later. This task is to update the notifications on the dashboard to look the same as on in the latest extensions (e.g. AI Lab)

Implement the new design as found in Penpot: https://design.penpot.app/#/workspace?team-id=938b71b0-2619-11ec-bdd0-f35c6ae458e9&file-id=971a4169-b56c-80d0-8002-5409c8c033db&page-id=91b5f35b-ecd5-8079-8006-2db8432426fc&layout=layers

Image

_No response_

### Additional context

_No response_

jiridostal:
### What does this PR do?
Implements new design of notifications on dashboard for consistency reasons.
Implements notification variant of info, warn and error.

Before:
image
After:
image

### How to test this PR?
To test all notification types, I'd recommend crafting custom 

Info, warn and error look like following:
image

Closes https://github.com/podman-desktop/podman-desktop/issues/13106

### DIFF
--- a/packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
+++ b/packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
@@ -17,6 +17,7 @@
  ***********************************************************************/
 import '@testing-library/jest-dom/vitest';
 
+import { faCircleExclamation, faCircleInfo, faExclamationTriangle } from '@fortawesome/free-solid-svg-icons';
 import { fireEvent, render, screen } from '@testing-library/svelte';
 import { beforeAll, expect, test, vi } from 'vitest';
 
@@ -55,4 +56,70 @@ test('Expect notification card to show notification title, description and close
   await fireEvent.click(deleteButton);
 
   expect(removeNotificationMock).toBeCalled();
+});
+
+test('Test info notification style and icon', () => {
+  const notification: NotificationCard = {
+    id: 1,
+    extensionId: 'extension',
+    title: 'Info notification title',
+    body: 'Info notification description',
+    type: 'info',
+  };
+  const { getAllByRole } = render(NotificationCardItem, {
+    notification,
+  });
+  const icon = getAllByRole('img', { hidden: true })[0];
+  // check icon shape
+  const pdIconPath = icon.querySelector('path')?.getAttribute('d');
+  const faIconPath = faCircleInfo.icon[4]; // index 4 is the actual path as per FA IconDefinition
+  expect(pdIconPath).toBe(faIconPath);
+  // check icon color
+  expect(icon).toHaveClass('text-[var(--pd-state-info)]');
+  // check region top border
+  expect(screen.getByRole('region', { name: 'id: 1' })).toHaveClass('border-[var(--pd-state-info)]');
+});
+
+test('Test warning notification style and icon', () => {
+  const notification: NotificationCard = {
+    id: 1,
+    extensionId: 'extension',
+    title: 'Warning notification title',
+    body: 'Warning notification description',
+    type: 'warn',
+  };
+  const { getAllByRole } = render(NotificationCardItem, {
+    notification,
+  });
+  const icon = getAllByRole('img', { hidden: true })[0];
+  // check icon shape
+  const pdIconPath = icon.querySelector('path')?.getAttribute('d');
+  const faIconPath = faExclamationTriangle.icon[4]; // index 4 is the actual path as per FA IconDefinition
+  expect(pdIconPath).toBe(faIconPath);
+  // check icon color
+  expect(icon).toHaveClass('text-[var(--pd-state-warning)]');
+  // check region top border
+  expect(screen.getByRole('region', { name: 'id: 1' })).toHaveClass('border-[var(--pd-state-warning)]');
+});
+
+test('Test error notification style and icon', () => {
+  const notification: NotificationCard = {
+    id: 1,
+    extensionId: 'extension',
+    title: 'Error notification title',
+    body: 'Error notification description',
+    type: 'error',
+  };
+  const { getAllByRole } = render(NotificationCardItem, {
+    notification,
+  });
+  const icon = getAllByRole('img', { hidden: true })[0];
+  // check icon shape
+  const pdIconPath = icon.querySelector('path')?.getAttribute('d');
+  const faIconPath = faCircleExclamation.icon[4]; // index 4 is the actual path as per FA IconDefinition
+  expect(pdIconPath).toBe(faIconPath);
+  // check icon color
+  expect(icon).toHaveClass('text-[var(--pd-state-error)]');
+  // check region top border
+  expect(screen.getByRole('region', { name: 'id: 1' })).toHaveClass('border-[var(--pd-state-error)]');
 });
--- a/packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
+++ b/packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
@@ -8,10 +8,6 @@ import Markdown from '../markdown/Markdown.svelte';
 
 export let notification: NotificationCard;
 
-async function removeNotification(id: number): Promise<void> {
-  await window.removeNotification(id);
-}
-
 const notificationStyleMap = {
   info: {
     borderColor: 'border-[var(--pd-state-info)]',
@@ -61,7 +57,7 @@ const notificationStyle = notificationStyleMap[notification.type];
     <div class="text-[var(--pd-content-card-carousel-card-header-text)]">
       <button
         class="p-1 hover:bg-[var(--pd-button-close-hover-bg)] hover:bg-opacity-10 transition-all rounded-[4px]"
-        on:click={(): Promise<void> => removeNotification(notification.id)}
+        on:click={(): Promise<void> => window.removeNotification(notification.id)}
         aria-label={`Delete notification ${notification.id}`}>
         <Icon icon={faXmark} />
       </button>
--- a/packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
+++ b/packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
-import { faCircleInfo, faXmark } from '@fortawesome/free-solid-svg-icons';
-import Fa from 'svelte-fa';
+import { faCircleExclamation, faCircleInfo, faExclamationTriangle, faXmark } from '@fortawesome/free-solid-svg-icons';
+import { Icon } from '@podman-desktop/ui-svelte/icons';
 
 import type { NotificationCard } from '/@api/notification';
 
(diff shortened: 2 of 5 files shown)


packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -11,35 +11,55 @@ export let notification: NotificationCard;
...
+      <div class="flex mt-2 self-center">
+        <Markdown markdown={notification.markdownActions} />
+      </div>
+    {/if}
+    <div class="text-[var(--pd-content-card-carousel-card-header-text)]">
+      <button
+        class="p-1 hover:bg-[var(--pd-button-close-hover-bg)] hover:bg-opacity-10 transition-all rounded-[4px]"
+        on:click={(): Promise<void> => removeNotification(notification.id)}
+        aria-label={`Delete notification ${notification.id}`}>
+        <Icon size="1x" icon={faXmark} />
```
gastoner: I think that 1x should be implicit?

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -11,35 +11,55 @@ export let notification: NotificationCard;
 async function removeNotification(id: number): Promise<void> {
   await window.removeNotification(id);
 }
+
+let notificationStyleMap = {
```
gastoner: `const` instead of `let`, since we are not updating this?

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -11,35 +11,55 @@ export let notification: NotificationCard;
...
+    color: '[var(--pd-state-warning)]',
+    icon: faExclamationTriangle,
+  },
+  error: {
+    color: '[var(--pd-state-error)]',
+    icon: faCircleExclamation,
+  },
+};
+
+let { [notification.type]: notificationStyle } = notificationStyleMap;
```
gastoner: Same here?

gastoner (commented):
Otherwise codewise LGTM, 
Just cant verify on Linux :(  I'll do it on Mac

packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
```diff
@@ -56,3 +57,69 @@ test('Expect notification card to show notification title, description and close
...
+    body: 'Warning notification description',
+    type: 'warn',
+  };
+  render(NotificationCardItem, {
+    notification,
+  });
+  const icon = screen.getAllByRole('img', { hidden: true })[0];
+  // check icon shape
+  const pdIconPath = icon.querySelector('path')?.getAttribute('d');
+  const faIconPath = faExclamationTriangle.icon[4]; // index 4 is the actual path as per FA IconDefinition
```
benoitf, line 97: could we use icon function from the core or something to avoid to have to go in array's index ?

packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
```diff
@@ -56,3 +57,69 @@ test('Expect notification card to show notification title, description and close
...
+    body: 'Error notification description',
+    type: 'error',
+  };
+  render(NotificationCardItem, {
+    notification,
+  });
+  const icon = screen.getAllByRole('img', { hidden: true })[0];
+  // check icon shape
+  const pdIconPath = icon.querySelector('path')?.getAttribute('d');
+  const faIconPath = faCircleExclamation.icon[4]; // index 4 is the actual path as per FA IconDefinition
```
benoitf, line 119: same remark there

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
benoitf: question: should we use $derived state rather than some map access ?

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -11,35 +11,55 @@ export let notification: NotificationCard;
...
   role="region"
   aria-label="id: {notification.id}">
-  <div class="flex flex-row">
-    <div class="mr-3">
-      {#if notification.type === 'info'}
-        <Fa size="1.5x" class="text-[var(--pd-notification-dot)]" icon={faCircleInfo} />
-      {/if}
+  <div class="flex flex-row space-x-3">
+    <div class="flex">
+        <Icon size="1.5x" icon={notificationStyle.icon} class="text-{notificationStyle.color}" />
```
benoitf: not sure it's a working pattern

packages/renderer/src/lib/dashboard/NotificationsBox.spec.ts
```diff
@@ -29,7 +29,7 @@ test('Expect notifications box to be hidden if there are no notifications in the
   notificationQueue.set([]);
   render(NotificationsBox);
 
-  const noNotificationsDiv = screen.queryByLabelText('Notifications');
+  const noNotificationsDiv = screen.queryByLabelText('Notifications Box');
```
benoitf: this could go as a separate PR as it does not seem to be related to the changes

gastoner:
https://github.com/user-attachments/assets/a4d1c701-6c53-4cdb-ba21-c691b8b6b460

1. I think the buttons should have same on:hover effect as other buttons (light+dark mode)
2. The text in light mode seems to me like purple-ish? I think it should be something like back-ish (but this one might be just a feeling), in dark mode seems fine

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
jiridostal: @benoitf I'm open to change it, but given that the map is constant, I believe it's more performant than $derived, especially when we don't need to react to any changes. But I'm happy to discuss

jiridostal:
@gastoner Buttons are provided externally, this issue with different hover color was there even with old notifications, will have a look if it makes sense to be fixed within this PR or needs to be fixed somewhere else

re:text-color - I took the variable from https://github.com/containers/podman-desktop-extension-ai-lab/blob/b8583e962f98419d9c3027fe2d61d26c0273ab11/packages/frontend/src/lib/notification/GPUEnabledMachine.svelte#L15 to keep the consistency with AI Lab

packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
jiridostal: Unfortunately I'm not able to come up with a better solution I always endup accessing an array by index. Are there any examples within our repos how to do it?

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
jiridostal: I took an inspiration from https://github.com/containers/podman-desktop-extension-ai-lab/blob/b8583e962f98419d9c3027fe2d61d26c0273ab11/packages/frontend/src/lib/notification/GPUEnabledMachine.svelte#L15 to maintain some consistency in between notifications.

simonrey1: This will not work with Tailwind: `class="text-{notificationStyle.color}"`

Screenshot 2025-07-23 at 10 58 24
https://tailwindcss.com/docs/detecting-classes-in-source-files

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -11,35 +11,55 @@ export let notification: NotificationCard;
...
+    icon: faCircleExclamation,
+  },
+};
+
+const { [notification.type]: notificationStyle } = notificationStyleMap;
 </script>
 
 <div
-  class="relative bg-[var(--pd-content-card-carousel-card-bg)] w-full py-4 px-5 rounded-md"
+  class="bg-[var(--pd-content-card-bg)] text-[var(--pd-content-card-text)] p-4 w-full border-{notificationStyle.color} border-t-[3px]"
```
simonrey1: The border style will not be recognized by tailwind.

 border-{notificationStyle.color} 
 
 
Screenshot 2025-07-23 at 10 58 24
https://tailwindcss.com/docs/detecting-classes-in-source-files

packages/renderer/src/lib/markdown/Markdown.spec.ts
```diff
@@ -53,7 +53,7 @@ describe('Custom button', () => {
     const markdownContent = screen.getByRole('region', { name: 'markdown-content' });
     expect(markdownContent).toBeInTheDocument();
     expect(markdownContent).toContainHTML(
-      '<a class="px-4 py-[6px] rounded-[4px] text-white! text-[13px] whitespace-nowrap bg-purple-600 hover:bg-purple-500 no-underline!">Name of the button</a>',
+      '<a class="px-4 py-[6px] rounded-[4px] text-white! text-[13px] whitespace-nowrap bg-purple-600 hover:bg-purple-500! no-underline!">Name of the button</a>',
```
simonrey1, line 56: Please don't use the important modifier

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
jiridostal: @eqqe Pushed updated code, hope Tailwind is able to process it this way

gastoner (approved):
LGTM

packages/renderer/src/lib/markdown/Markdown.spec.ts
gastoner: In this file it has to be done this way (there are a lot of `!` in this file). We are defining global styles in `Markdown.svelte`. So either this or huge refactor. Because this "button" is used not that mutch and the global style specified is for "link buttons" e.g. when creating kind cluster on creation page

gastoner:
@eqqe can you PTAL?

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -11,35 +11,58 @@ export let notification: NotificationCard;
 async function removeNotification(id: number): Promise<void> {
```
simonrey1: Suggestion: call `window.removeNotification` directly on the button

packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
```diff
@@ -56,3 +57,69 @@ test('Expect notification card to show notification title, description and close
...
+    id: 1,
+    extensionId: 'extension',
+    title: 'Info notification title',
+    body: 'Info notification description',
+    type: 'info',
+  };
+  render(NotificationCardItem, {
+    notification,
+  });
+  const icon = screen.getAllByRole('img', { hidden: true })[0];
```
simonrey1: Suggestion use `const { getAllByRole } = render` instead

simonrey1: Using getAllByRole and taking the first element can be a bit fragile, especially since the order of icons might change or more icons could be added later.

Maybe we could use this solution (and merge it in another PR):
https://stackoverflow.com/questions/62587372/query-font-awesome-element-in-react-testing-library

packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
```diff
@@ -56,3 +57,69 @@ test('Expect notification card to show notification title, description and close
...
+  // check icon shape
+  const pdIconPath = icon.querySelector('path')?.getAttribute('d');
+  const faIconPath = faCircleInfo.icon[4]; // index 4 is the actual path as per FA IconDefinition
+  expect(pdIconPath).toBe(faIconPath);
```
simonrey1, line 76: IMHO we should trust the font awesome rendering. So it won't be necessary to import the icon and check that paths match.

gastoner: We were updating the icons not that long ago because we had like 3 different options to show an icon (using classes, FA, and component rendering). Each component was using a different type, which was super problematic. This way, we can use just this `Icon` component and paste into it one of those options

simonrey1: But I don't understand why we should check the equality of the svg paths?

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -11,35 +11,58 @@ export let notification: NotificationCard;
...
+    icon: faExclamationTriangle,
+  },
+  error: {
+    borderColor: 'border-[var(--pd-state-error)]',
+    iconColor: 'text-[var(--pd-state-error)]',
+    icon: faCircleExclamation,
+  },
+};
+
+const { [notification.type]: notificationStyle } = notificationStyleMap;
```
simonrey1: Suggestion: `const notificationStyle = notificationStyleMap[notification.type] || notificationStyleMap.info;`

During my tests, I set "warning" instead of "warn" and an exception was thrown (no notification at all displayed). Defaulting to info helps in my case.

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -11,35 +11,58 @@ export let notification: NotificationCard;
...
-      <div class="text-[var(--pd-content-card-carousel-card-header-text)] font-bold" aria-label="Notification title">
-        {notification.title}
+    <div class="flex flex-col space-y-2 flex-1">
+      <div class="flex flex-row items-center justify-between">
+        <span class="font-medium" aria-label="Notification title">
+          {notification.title}
+        </span>
       </div>
-      <div class="text-[var(--pd-content-card-carousel-card-text)]" aria-label="Notification description">
+      <span aria-label="Notification description">
```
simonrey1: Suggestion: 

Hide this span when notification does not have a body. Otherwise it leaves a 8px margin underneath the title and the notification text is not vertically centered.

```
{#if notification.body}
<span aria-label="Notification description">
    <Markdown markdown={notification.body} />
</span>
{/if}
```
Before:
Screenshot 2025-07-25 at 11 40 54

After: 
Screenshot 2025-07-25 at 11 46 29

packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
jiridostal: During the test run the icon already has a generated class so comparing just "icon names" didn't do the trick.

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
jiridostal: It doesn't pass the typecheck for me
image
Anyways, I think it's better that it does not let you do it than default to something you didn't want initially.

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
jiridostal: Good point, I'll update it

packages/renderer/src/lib/dashboard/NotificationCardItem.spec.ts
jiridostal: I'll create a new PR for this after this

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
simonrey1: Sorry to bother again but could we go with 
`const notificationStyle = notificationStyleMap[notification.type];`
IMHO it is easier to read than:
`const { [notification.type]: notificationStyle } = notificationStyleMap;`

simonrey1 (approved):

PR #13210, opened 2025-07-15
fix(ci): disable some unstable tests

### LINKED ISSUE #13166: bug(pr-check): large number of `JavaScript heap out of memory`
### Bug description

I saw an increasing number of error in the pr-check, specially in the unit tests

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

https://github.com/podman-desktop/podman-desktop/actions/runs/16211288764/job/45772360580
https://github.com/podman-desktop/podman-desktop/actions/runs/16199172319/job/45775115482

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

simonrey1:
### What does this PR do?

Disable some unstable tests waiting to fix out of memory while loading extensions

Temporary fixes #13166

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/index.spec.ts
+++ b/packages/main/src/plugin/index.spec.ts
@@ -334,7 +334,7 @@ test('configurationRegistry propagated', async () => {
 const pushImageHandlerId = 'container-provider-registry:pushImage';
 const pushImageHandlerOnDataEvent = `${pushImageHandlerId}-onData`;
 
-test('push image command sends onData message with callbackId, event name and data, mark task as success on end event', async () => {
+test.skip('push image command sends onData message with callbackId, event name and data, mark task as success on end event', async () => {
   const createTaskSpy = vi.spyOn(TaskManager.prototype, 'createTask');
   await pluginSystem.initExtensions(new Emitter<ConfigurationRegistry>());
   const handle = handlers.get(pushImageHandlerId);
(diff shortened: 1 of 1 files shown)


benoitf (changes requested):
I don't see the rationale to skip them

simonrey1:
What can we do to improve these tests memory-wise?

axel7083:
> What can we do to improve these tests memory-wise?

We should probably make the classes implement `IDisposable`, and properly handle the dispose of all resources when running the tests, (afterEach / beforeEach).

We seems to have a lot of `Map` / `Set` that seems to be never cleared, in addition I am suspecting the Inversify package to be a bit problematic, as the tests mentionned are linked to injectable classes

axel7083 (commented):
https://github.com/podman-desktop/podman-desktop/pull/13210#issuecomment-3077793715

benoitf:
my point is also that it's always working on Linux (no memory leak there ? ), so why disable it. In the past we disabled it for custom operating system.
so even if it's not launched on a dedicated platform, some tests are still independent of the OS so if they're only enabled on Linux it ensures CI will detect a regression

jeffmaury (commented):
Maybe can we investigate if initExtensions could be called only once ?

benoitf:
let me try to cleanup the inversify container instance

@jeffmaury well in the default way, it's called only once. Only unit tests are calling it multiple times.

benoitf:
opened https://github.com/podman-desktop/podman-desktop/pull/13237

simonrey1:
> my point is also that it's always working on Linux (no memory leak there ? ), so why disable it. In the past we disabled it for custom operating system. so even if it's not launched on a dedicated platform, some tests are still independent of the OS so if they're only enabled on Linux it ensures CI will detect a regression

@benoitf 
No it fails sometimes on the Ubuntu worker too, here for instance it has a `JavaScript heap out of memory` on one of the runs https://github.com/podman-desktop/podman-desktop/actions/runs/16319375971/job/46092969801

Screenshot 2025-07-17 at 09 59 59

PR #13248, opened 2025-07-17
chore(lint): update Map to SvelteMap

simonrey1:
### What does this PR do?

This PR fixes the update of the linter.

Try to help update of PR #13239

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
+++ b/packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
@@ -133,7 +133,7 @@ onMount(() => {
       return;
     }
     podCreation = value;
-    const mapPortPrivate = new SvelteMap<number, string[]>();
+    const mapPortPrivate = new Map<number, string[]>();
     podCreation.containers.forEach(container => {
       container.ports.forEach(port => {
         mapPortExposed.set(port.PublicPort, {
--- a/packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
+++ b/packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
@@ -2,6 +2,7 @@
 import { Button, Checkbox, ErrorMessage, Input, StatusIcon } from '@podman-desktop/ui-svelte';
 import { ContainerIcon } from '@podman-desktop/ui-svelte/icons';
 import { onDestroy, onMount } from 'svelte';
+import { SvelteMap } from 'svelte/reactivity';
 import type { Unsubscriber } from 'svelte/store';
 import { router } from 'tinro';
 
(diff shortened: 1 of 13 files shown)


packages/renderer/src/lib/ui/ProviderResultPage.svelte
```diff
@@ -95,7 +96,7 @@ const filtered = $derived(
 
 function onProviderChecked(id: string, checked: boolean): void {
   selectedProviders.set(id, checked);
-  selectedProviders = new Map(selectedProviders);
+  selectedProviders = new SvelteMap(selectedProviders);
```
simonrey1, line 99: This one I am not sure because selectedProviders is a $state of Map, and svelte would not let me do a $state of SvelteMap and I am not sure what should I do?

packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
```diff
@@ -132,7 +133,7 @@ onMount(() => {
       return;
     }
     podCreation = value;
-    const mapPortPrivate = new Map<number, string[]>();
+    const mapPortPrivate = new SvelteMap<number, string[]>();
```
axel7083: Not sure that this is expected here, as this does not seems to be used in a reactive statement - but rather as a temporary storage

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
```diff
@@ -184,7 +185,7 @@ onMount(async () => {
   });
 
   onboardingsUnsubscribe = onboardingList.subscribe(onboardingItems => {
-    extensionOnboardingEnablement = new Map<string, string>();
+    extensionOnboardingEnablement = new SvelteMap<string, string>();
```
axel7083, line 188: remark (non-blocking): seems like we would need to migrate to svelte5 as this is kinda ugly to see a reactive SvelteMap inside a store subscribe, and finally updating another variable, this could be a $derived

packages/renderer/src/lib/ui/ProviderResultPage.svelte
axel7083: we should avoid nested reactivity I guess

packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
simonrey1: I will rollback it, you are right no need to change this one for this update

axel7083 (approved):

packages/renderer/src/lib/preferences/PreferencesResourcesRendering.svelte
simonrey1: It seems that it is needed for the update of the linter to pass

https://github.com/podman-desktop/podman-desktop/actions/runs/16347956584/job/46186675317?pr=13239

/home/runner/work/podman-desktop/podman-desktop/packages/renderer/src/lib/pod/PodCreateFromContainers.svelte
Error:   136:28  error  Found a mutable instance of the built-in Map class. Use SvelteMap instead  svelte/prefer-svelte-reactivity

PR #13249, opened 2025-07-17
test: add e2e test for kube play YAML build option

### LINKED ISSUE #12212: Add test case to cover a build option when using play Kube Yaml functionality

New build option was introduced in 1.18.0.

See https://github.com/podman-desktop/podman-desktop/pull/10801 for details on the feature.

Implement a new or add a test scenario to cover the functionality of the build option when "playing" a kube yaml.

_No response_

### Additional context

_No response_

amisskii:
### What does this PR do?
adds e2e test for kube play YAML build option

https://github.com/podman-desktop/podman-desktop/issues/12212

### How to test this PR?

https://github.com/podman-desktop/e2e/actions/runs/16344239443

### DIFF
--- a/tests/playwright/resources/kube/foobar/Containerfile
+++ b/tests/playwright/resources/kube/foobar/Containerfile
@@ -1,0 +1,2 @@
+FROM docker.io/library/nginx:alpine
+ 
--- a/tests/playwright/resources/kube/foobar/Containerfile
+++ b/tests/playwright/resources/kube/foobar/Containerfile
@@ -1,2 +1,1 @@
 FROM docker.io/library/nginx:alpine
- 
--- a/tests/playwright/resources/kube/play-yaml-build-test.yaml
+++ b/tests/playwright/resources/kube/play-yaml-build-test.yaml
@@ -6,6 +6,8 @@ spec:
   containers:
     - name: container
       image: foobar
+      imagePullPolicy: Never
       ports:
         - name: http
           containerPort: 80
+ 
--- a/tests/playwright/resources/kube/play-yaml-build-test.yaml
+++ b/tests/playwright/resources/kube/play-yaml-build-test.yaml
@@ -9,4 +9,3 @@ spec:
       ports:
         - name: http
           containerPort: 80
- 
--- a/tests/playwright/resources/kube/play-yaml-build-test.yaml
+++ b/tests/playwright/resources/kube/play-yaml-build-test.yaml
@@ -10,4 +10,3 @@ spec:
       ports:
         - name: http
           containerPort: 80
- 
--- a/tests/playwright/resources/kube/play-yaml-build-test.yaml
+++ b/tests/playwright/resources/kube/play-yaml-build-test.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: play-yaml-build-test
+spec:
+  containers:
+    - name: container
+      image: foobar
+      ports:
+        - name: http
+          containerPort: 80
+ 
--- a/tests/playwright/src/model/core/states.ts
+++ b/tests/playwright/src/model/core/states.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -68,4 +68,9 @@ export enum ExtensionState {
   Running = 'RUNNING',
   NotInstalled = 'NOT-INSTALLED',
   Downloadable = 'DOWNLOADABLE',
+}
+
+export enum ImageState {
+  Used = 'USED',
+  Unused = 'UNUSED',
 }
--- a/tests/playwright/src/model/pages/containers-page.ts
+++ b/tests/playwright/src/model/pages/containers-page.ts
@@ -175,4 +175,13 @@ export class ContainersPage extends MainPage {
       return this;
     });
   }
+
+  async getContainerImage(name: string): Promise<string> {
+    const container = await this.getContainerRowByName(name);
+    const image = await container?.getByRole('cell').nth(5).textContent();
+    if (image) {
+      return image;
+    }
+    return '';
+  }
 }
--- a/tests/playwright/src/model/pages/play-kube-yaml-page.ts
+++ b/tests/playwright/src/model/pages/play-kube-yaml-page.ts
@@ -33,6 +33,7 @@ export class PlayKubeYamlPage extends BasePage {
   readonly kubernetesContext: Locator;
   readonly kubernetesNamespaces: Locator;
   readonly alertMessage: Locator;
+  readonly buildCheckbox: Locator;
 
   constructor(page: Page) {
     super(page);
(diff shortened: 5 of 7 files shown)


tests/playwright/src/model/pages/play-kube-yaml-page.ts
```diff
@@ -99,6 +102,12 @@ export class PlayKubeYamlPage extends BasePage {
       }
 
       await this.yamlPathInput.fill(pathToYaml);
+      await playExpect(this.buildCheckbox).not.toBeChecked();
+      await playExpect(this.buildCheckbox).toBeEnabled();
```
cbr7, line 106: I'm not sure that a checkbox has this as a valid state, are you sure this works?

tests/playwright/src/specs/play-yaml-build-smoke.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    await playExpect.poll(async () => await podDetails.getState(), { timeout: 15_000 }).toBe(PodState.Running);
+  });
+  test('Verify that the deployed pod container uses the localhost image', async ({ navigationBar }) => {
+    const imagesPage = await navigationBar.openImages();
+    await playExpect(imagesPage.heading).toBeVisible();
+    await playExpect.poll(async () => await imagesPage.getCurrentStatusOfImage(LOCAL_IMAGE_NAME)).toBe('USED');
+
+    const containersPage = await navigationBar.openContainers();
+    await playExpect(containersPage.heading).toBeVisible();
+    playExpect(await containersPage.getContainerImage(CONTAINER_NAME)).toBe(CONTAINER_IMAGE);
```
cbr7:
```suggestion
    await playExpect.poll(async () => containersPage.getContainerImage(CONTAINER_NAME)).toBe(CONTAINER_IMAGE);
```

tests/playwright/src/model/pages/play-kube-yaml-page.ts
amisskii: Yes, it should work

tests/playwright/resources/kube/foobar/Containerfile
```diff
@@ -0,0 +1,2 @@
+FROM docker.io/library/nginx:alpine
+
```
ScrewTSW: is there an extra space on the line?

tests/playwright/resources/kube/play-yaml-build-test.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: play-yaml-build-test
+spec:
+  containers:
+    - name: container
+      image: foobar
+      ports:
+        - name: http
+          containerPort: 80
+
```
ScrewTSW: is there an extra space on the line?

tests/playwright/src/model/core/states.ts
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2025 Red Hat, Inc.
```
ScrewTSW, line 2: @odockal how do we handle copyright in our tests? should we do a general update of the copyrights?

tests/playwright/src/model/pages/containers-page.ts
```diff
@@ -175,4 +175,13 @@ export class ContainersPage extends MainPage {
       return this;
     });
   }
+
+  async getContainerImage(name: string): Promise<string> {
+    const container = await this.getContainerRowByName(name);
+    const image = await container?.getByRole('cell').nth(5).textContent();
```
ScrewTSW, line 181: don't really like the magic number here. is there no ARIA for this? might be worth committing that as a PR to PD

tests/playwright/src/model/pages/play-kube-yaml-page.ts
ScrewTSW: I agree with @cbr7 , is there any action performed in-between those two checks? if yes, doesn't seem like the proper way to track that change, especially with default timeouts

ScrewTSW (changes requested):

tests/playwright/src/model/pages/play-kube-yaml-page.ts
amisskii: @ScrewTSW I don't see any reason why I can't use these two checks: one asserts that the checkbox is not checked, and the other asserts that it's enabled to be checked — just like a button is enabled to be clicked

tests/playwright/src/model/pages/containers-page.ts
amisskii: There is no aria-label for specific resource cells in the tables, so this is the most straightforward method to locate them. I’ve created a follow-up issue for adding aria-labels: https://github.com/podman-desktop/podman-desktop/issues/13291

tests/playwright/src/model/core/states.ts
odockal: @ScrewTSW No. We update the copyright if we are updating the file.

tests/playwright/src/model/pages/containers-page.ts
odockal: Good path!

cbr7 (approved):

tests/playwright/src/model/pages/play-kube-yaml-page.ts
odockal: @cbr7 Is your concern regarding the way the checkbox is implemented with overlapping div around actual checkbox ?

cbr7: not per say, only in so far as a div does not really have much use for an Enabled status, but if Anton checked and it works, it should be fine I think. I have no strong feeling on this either way.

odockal: Ah, it is actually parent that is defined as this locator.

tests/playwright/resources/kube/play-yaml-build-test.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: play-yaml-build-test
+spec:
+  containers:
+    - name: container
+      image: foobar
+      ports:
```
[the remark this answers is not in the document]
odockal: @amisskii I think might not be bad idea, wdyt?

tests/playwright/src/model/pages/pods-page.ts
```diff
@@ -46,7 +46,7 @@ export class PodsPage extends MainPage {
       if (podRow === undefined) {
         throw Error(`Pod: ${name} does not exist`);
       }
-      await podRow.getByText(name).click();
+      await podRow.getByText(name, { exact: true }).click();
```
odockal: We are searching for a button with this text here. We can use `getByRole('button', ...)`.

ScrewTSW (approved):
LGTM

odockal (changes requested):

tests/playwright/src/model/pages/pods-page.ts
amisskii: I've refactored this locator to match the HTML structure of the page. However, we can't use ('button', {name: name}) here, as the button doesn't contain the pod name as text, there's a div element inside it that holds the name.

odockal (approved):
LGTM.

PR #13236, opened 2025-07-16
refactor(main): make all injectable disposable

axel7083:
### What does this PR do?

Trying to fix https://github.com/podman-desktop/podman-desktop/issues/13166 by making all injectable IDisposable and marking the dispose method as `preDestroy` (called when unbind)

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature

PR #13256, opened 2025-07-17
feat(ExtensionLoader): implements AsyncDisposable

axel7083:
### What does this PR do?

Split of https://github.com/podman-desktop/podman-desktop/pull/13236 - Making the ExtensionLoader AsyncDisposable, to stop all extensions when called.

Adding the `@preDestroy` inversify annotation to tell inversify to call this method when unbinding this object. In the follow-up PR, updating the `src/plugin/index.spec` to unbind all singleton after each test.

Related to https://github.com/podman-desktop/podman-desktop/issues/13166
Part of https://github.com/podman-desktop/podman-desktop/issues/13257

### How to test this PR?

### DIFF
--- a/packages/main/src/plugin/extension/extension-loader.ts
+++ b/packages/main/src/plugin/extension/extension-loader.ts
@@ -22,7 +22,7 @@ import * as path from 'node:path';
 import type * as containerDesktopAPI from '@podman-desktop/api';
 import AdmZip from 'adm-zip';
 import { app, clipboard as electronClipboard } from 'electron';
-import { inject, injectable } from 'inversify';
+import { inject, injectable, preDestroy } from 'inversify';
 
 import { ColorRegistry } from '/@/plugin/color-registry.js';
 import {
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/extension/extension-loader.ts
```diff
@@ -221,6 +221,21 @@ export class ExtensionLoader {
     this.extensionDevelopmentFolder.onNeedToLoadExension(extension => this.loadExtension(extension));
   }
 
+  @preDestroy()
+  async [Symbol.asyncDispose](): Promise<void> {
```
axel7083, line 225: We cannot use a synchronous disposable, as the deactivate of extensions is async, we don't want to use dangling promise, to properly respect the lifecycle, we should use the AsyncDisposable typescript object

MarsKubeX (approved):
code LGTM

gastoner (approved):
Codewise LGTM

PR #13262, opened 2025-07-17
test(packages/renderer): make test run faster

simonrey1:
### What does this PR do?

I changed the timers value in this test to make it pass in 79ms (on my Mac) instead of 2 200ms.
This is to make the CI faster.

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.spec.ts
+++ b/packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.spec.ts
@@ -43,15 +43,15 @@ test('Check cleanupProviders is called and button is in progress', async () => {
   const cleanupButton = screen.getByRole('button', { name: 'Cleanup' });
   expect(cleanupButton).toBeInTheDocument();
 
-  // mock the cleanup as waiting for 2 seconds
-  cleanupProvidersMock.mockResolvedValue(new Promise(resolve => setTimeout(resolve, 2000)));
+  // mock the cleanup as waiting for 10ms
+  cleanupProvidersMock.mockResolvedValue(new Promise(resolve => setTimeout(resolve, 10)));
 
   // click on the cleanup button
   expect(cleanupButton).toBeEnabled();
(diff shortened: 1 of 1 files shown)


benoitf:
not sure about the last part of the title "to avoid blocking ci"

IMHO 2s won't block the CI

simonrey1:
> not sure about the last part of the title "to avoid blocking ci"
> 
> IMHO 2s won't block the CI

It is not blocking but it will improve by approximately 0.3% the speed of every unit test run in the CI.

benoitf:
I would call it misleading title

in case we don't hit the cache, the delay penalty is much more than 2s
2s can be a network latency, download dependency delay, etc.

so I would just stick to "make the test faster"

benoitf (approved):
I think it should be migrated to waitFor logic for some parts (as we're doing now in new tests)

jeffmaury (approved):
LGTM

packages/renderer/src/lib/troubleshooting/TroubleshootingRepairCleanup.spec.ts
```diff
@@ -43,24 +43,24 @@ test('Check cleanupProviders is called and button is in progress', async () => {
...
-  cleanupProvidersMock.mockResolvedValue(new Promise(resolve => setTimeout(resolve, 2000)));
+  // mock the cleanup as waiting for 10ms
+  cleanupProvidersMock.mockResolvedValue(new Promise(resolve => setTimeout(resolve, 10)));
 
   // click on the cleanup button
   expect(cleanupButton).toBeEnabled();
   await fireEvent.click(cleanupButton);
 
   // wait next tick
-  await new Promise(resolve => setTimeout(resolve, 100));
```
axel7083, line 54: Can't the test be refactored to check using a waitFor to wait for a valid condition instead of waiting with a timeout ?

simonrey1: Oops sorry I merged without seing your comment, I will take a look thanks

PR #13263, opened 2025-07-17
test(packages/renderer): make test run faster

simonrey1:
### What does this PR do?

I changed the  value in this test to make it pass in 79ms (on my Mac) instead of 2 200ms.
This is to make the CI faster.

### How to test this PR?


axel7083 (approved):

packages/renderer/src/lib/kube/pods/terminal/KubernetesTerminal.spec.ts
```diff
@@ -82,9 +82,9 @@ test('Test should render the terminal and being able to reconnect', async () =>
 
   render(KubernetesTerminal, { podName: 'podName', containerName: 'containerName' });
 
-  await new Promise(resolve => setTimeout(resolve, 1000));
```
axel7083, line 85: Honestly, IMO setTimeout should be forbidden in spec file

axel7083 (approved):

PR #13273, opened 2025-07-18
test: speed up some vitest tests 

simonrey1:
### What does this PR do?

In the vitest tests, I am trying to remove the setTimeout that are not needed or replace them with a `waitFor` to speed up their execution in the CI.

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.spec.ts
+++ b/packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.spec.ts
@@ -235,7 +235,7 @@ test('Expects images.icon option to be used when no themes are present', async (
   const icon = await waitFor(() => {
     return screen.getByRole('img', { name: `${testProvidersInfoWithSessionRequests[0].displayName}` });
   });
-    
+
   expect(icon).toBeInTheDocument();
   expect(icon).toHaveAttribute('src', './icon.png');
 });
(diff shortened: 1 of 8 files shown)


packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.spec.ts
```diff
@@ -232,12 +232,11 @@ test('Expects images.icon option to be used when no themes are present', async (
...
-  await new Promise(resolve => setTimeout(resolve, 200));
-
-  const icon = screen.getByRole('img', { name: `${testProvidersInfoWithSessionRequests[0].displayName}` });
-  expect(icon).toBeInTheDocument();
-  expect(icon).toHaveAttribute('src', './icon.png');
+  await waitFor(() => {
+    const icon = screen.getByRole('img', { name: `${testProvidersInfoWithSessionRequests[0].displayName}` });
+    expect(icon).toBeInTheDocument();
+    expect(icon).toHaveAttribute('src', './icon.png');
+  });
```
axel7083: suggestion: a nice synthax I like to use, is only keeping inside the wait synthax inside the waitFor


```suggestion
  const icon = await waitFor(() => {
    return screen.getByRole('img', { name: `${testProvidersInfoWithSessionRequests[0].displayName}` });
  });
    
  expect(icon).toBeInTheDocument();
  expect(icon).toHaveAttribute('src', './icon.png');
```

while reading, we clearly understand what we are waiting for, here (I guess,) that the img is available

packages/renderer/src/lib/preferences/PreferencesAuthenticationProvidersRendering.spec.ts
```diff
@@ -261,10 +260,9 @@ test('Expects images.icon.dark option to be used when theme is dark', async () =
   configMock.mockReturnValue('dark');
   render(PreferencesAuthenticationProvidersRendering, {});
 
-  // wait for image to be loaded
-  await new Promise(resolve => setTimeout(resolve, 200));
-
-  const icon = screen.getByRole('img', { name: `${testProvidersInfoWithSessionRequests[0].displayName}` });
-  expect(icon).toBeInTheDocument();
-  expect(icon).toHaveAttribute('src', './icon-dark.png');
+  await waitFor(() => {
```
axel7083: suggestion(non-blocking): same remark as https://github.com/podman-desktop/podman-desktop/pull/13273/files#r2215566014

axel7083 (approved):
LGTM, great work, love to see such refactor

PR #13275, opened 2025-07-18
fix(extension/podman): ensure loop of monitoring can be started at activation of extension

simonrey1:
### What does this PR do?

This PR is made to prepare another PR where the monitoring loop is tested. The stop loop is activated in the deactivate method, but it is not desactivated in the activate method. So the test cannot be run as the global stop loop can be modified from another test.

https://github.com/podman-desktop/podman-desktop/pull/13153/files/ec5af1256af8c61821c3b277af26f5c553cba34d#r2215571950

### How to test this PR?

### DIFF
--- a/extensions/podman/packages/extension/src/extension.spec.ts
+++ b/extensions/podman/packages/extension/src/extension.spec.ts
@@ -3300,16 +3300,21 @@ test('activate and autostart should not duplicate machines ', async () => {
   // call the autostart method
   const promiseAutoStart = autoStartMethod?.start();
 
-  // call 100 times monitorMachines
-  for (let i = 0; i < 100; i++) {
-    extension.monitorMachines(provider, podmanConfiguration).catch(() => {});
-  }
+  vi.useFakeTimers();
+
+  const promises = Array.from({ length: 100 }).map(() => extension.monitorMachines(provider, podmanConfiguration));
 
   await promiseAutoStart;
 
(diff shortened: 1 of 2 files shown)


axel7083 (approved):

axel7083 (commented):
Some tests are failing

simonrey1:
> Some tests are failing

I am investigating! It seems like there are side effects between tests in the extension.spec.ts file because they run individually.

extensions/podman/packages/extension/src/extension.spec.ts
```diff
@@ -3300,16 +3300,21 @@ test('activate and autostart should not duplicate machines ', async () => {
...
 
   // should be only 1 but we allow some more calls (if there is not a check to check during the autostart it would be 100+ calls)
   expect(podmanMachineListCalls).toBeLessThan(5);
   expect(promiseAutoStart).toBeDefined();
+
+  await vi.advanceTimersByTimeAsync(5000);
+
+  await Promise.allSettled(promises);
+
+  vi.useRealTimers();
```
axel7083: We usually recommend to call vi.useRealTimers in the afterEach, to ensure that it is always called, even if the test is failing

axel7083 (commented):
https://github.com/podman-desktop/podman-desktop/pull/13275/files/d9cb2e729354714f16ce7eceee40f7cb1c9a81ae..44c87969267f1589bdcc0b7458cae49fbfcbf9be#r2216036279

axel7083 (approved):

PR #13289, opened 2025-07-21
fix(ci): update CRIU download URL

simonrey1:
### What does this PR do?

The `cz.archive.ubuntu.com` mirror for CRIU is currently unavailable, causing CI failures. This commit updates the download URL to use `archive.ubuntu.com` directly, which is a more reliable and generally available mirror.

This change ensures that the Podman update process in CI environments can reliably download and install the `criu` package, preventing build failures related to unreachable repositories.


gastoner (approved):
Codewise LGTM, 

I was able to download the `deb` from both archives on my machine

simonrey1:
> Codewise LGTM,
> 
> I was able to download the `deb` from both archives on my machine

It seems that the server is up again from here too, but let's merge this in case it happens again :)

PR #13295, opened 2025-07-21
fix(extensions/kind): dispose update version on uninstall

### LINKED ISSUE #13128: Update button for particular CLI tool in Resources Page is not removed after the cli tool was uninstalled
### Bug description

For example Kind. 

https://github.com/user-attachments/assets/97fd06ed-a6c5-4176-a030-0c25672c00ea

### Operating system

Linux

### Installation Method

None

### Version

1.20.0

### Steps to reproduce

1. I install a kind (latest) via CLI tools Page
2. Downgrade kind to `latest-1`
3. check the update button available in resources
4. Uninstall Kind from CLI tools
5. Update button from Resources disappear
Actual state: button still there, throws an error when clicked

### Relevant log output

```shell
Uncaught (in promise) Error: Cannot update kind. No cli tool installed.
	Promise.then		
o	@	ProviderUpdateButton.svelte:42
await in o		
click	@	ProviderUpdateButton.svelte:59
```

### Additional context

_No response_

simonrey1:
### What does this PR do?

This PR cleans the update data for the kind extension when it is uninstalled, to avoid showing a ghost "update" button.

Fixes #13128 

### How to test this PR?
See issue #13128 

### DIFF
--- a/extensions/kind/src/extension.spec.ts
+++ b/extensions/kind/src/extension.spec.ts
@@ -356,6 +356,14 @@ describe('cli#update', () => {
 
     expect(disposeMock).toHaveBeenCalled();
   });
+
+  test('uninstall updatable version should dispose update', async () => {
+    vi.mocked(KindInstaller.prototype.getLatestVersionAsset).mockResolvedValue(mockV1Release);
+
+    await (await getCliToolInstaller()).doUninstall({} as unknown as extensionApi.Logger);
+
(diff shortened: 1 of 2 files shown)


cdrage (changes requested):
Code LGTM, but we should have a test to test that dispose() was called on uninstall.

cdrage (approved):

Issue #13095, opened 2025-07-08
Explore options how to centralize experimental configuration


gastoner:
### Is your enhancement related to a problem? Please describe

Currently we have in our configuration experimental features like stored like: 
```
  "dockerCompatibility.enabled": true,
  "kubernetes.statesExperimental": true,
  "tasks.StatusBar": true,
  "tasks.Manager": true,
```

In the parent issue we are trying to store for each experimental feature 2 other properties (timestamp/remindAt and enabled/disabled)

But this way we have 2 places in the configuration where are we string information about this features

### Describe the solution you'd like

We would like to do something like this: 

```
"experimentalFeatures": [
    "kubernetes.statesExperimental": {
         remindAt: '1748321012505',
         enabled: true //dont notify option flag
     },
    "statusbarProviders.showProviders": {
         remindAt: '1748321012505',
         enabled: false //dont notify option flag
     },
    //another exp feature 1 and 2 are disabled from experimental settings
  },
]
```

This way we have centralized experimental features, and we can add/remove properties in the future if needed
This could be done by extending ExperimentalConfigurationRegistry by ConfigurationRegistry. Where could we add methods for getting those properties in the given feature

The problem is that if we do it this way there would have been some changes with showing these settings e.g. currently  "kubernetes.statesExperimental" is shown under "Kubernetes"

gastoner:
There might be a solution to use something like: 
```
    "kubernetes.statesExperimental": {
         remindAt: '1748321012505',
         enabled: true //dont notify option flag
     },
    "statusbarProviders.showProviders": {
         remindAt: '1748321012505',
         enabled: false //dont notify option flag
     },
    //another exp feature 1 and 2 are disabled from experimental settings
```

gastoner:
Closing this since we are already implementing those changes in comment above ^^

PR #13316, opened 2025-07-23
feat: added experimental configuration manager

gastoner:
Assisted-by: Cursor

### What does this PR do?
Adds experimental configuration manager 

In further PRs I'm using just isExperimentalConfigurationEnabled and updateExoerimentalConfigurationValue functions, other are either unused (for future?), or helper functions

Required for https://github.com/podman-desktop/podman-desktop/issues/13095

### How to test this PR?
Unit tests 

1. checkout to https://github.com/podman-desktop/podman-desktop/pull/11992 
2. enable some features (not all of them)
3. go to "Experimental page", open `~/.local/share/containers/podman-desktop/configuration/settings.json`
4. click on enable all on top of the screen, you should see that the not enabled features should have `{}` value
PS: the other (previously enabled features) will probably have values like: 
```
some.feature: {
  remindAt: TIMESTAMP
  disabled: false
}
``` 
=> This is another PR (https://github.com/podman-desktop/podman-desktop/pull/13317)

### DIFF
--- a/packages/main/src/plugin/experimental-configuration-manager.ts
+++ b/packages/main/src/plugin/experimental-configuration-manager.ts
@@ -32,6 +32,9 @@ export class ExperimentalConfigurationManager {
    */
   parseKey(key: string): { section: string; property: string } {
     const dotIndex = key.indexOf('.');
+    // If is the '.' not found
+    // if is the key '.property'
+    // If is the key 'section.'
     if (dotIndex === -1 || dotIndex === 0 || dotIndex === key.length - 1) {
       return { section: '', property: '' };
     }
@@ -53,12 +56,7 @@ export class ExperimentalConfigurationManager {
     config: unknown,
     scope?: containerDesktopAPI.ConfigurationScope | containerDesktopAPI.ConfigurationScope[],
   ): Promise<void> {
-    // HACK: when setting `{}` as value we need to stringify and parse the svelte state
-    let settings = config;
-    if (typeof config === 'object') {
-      settings = JSON.parse(JSON.stringify(config));
-    }
-    await this.configurationRegistry.updateConfigurationValue(key, settings, scope);
+    await this.configurationRegistry.updateConfigurationValue(key, config, scope);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/main/src/plugin/experimental-configuration-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+   * @param key - Full configuration key (e.g., 'kubernetes.statesExperimental')
+   * @param config - Configuration object for the feature. If undefined, disables the feature. If object (empty or with properties), enables the feature.
+   * @param scope - Configuration scope
+   */
+  async updateExperimentalConfigurationValue(
+    key: string,
+    config: unknown,
+    scope?: containerDesktopAPI.ConfigurationScope | containerDesktopAPI.ConfigurationScope[],
+  ): Promise<void> {
+    // HACK: when setting `{}` as value we need to stringify and parse the svelte state
```
cdrage: HACK seems kind of odd, maybe we supply an issue # or something? Odd that you have to do this, but not const settings = {};

packages/main/src/plugin/experimental-configuration-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+  constructor(@inject(ConfigurationRegistry) private configurationRegistry: ConfigurationRegistry) {}
+
+  /**
+   * Parse a configuration key into section and property
+   * @param key - Full configuration key (e.g., 'kubernetes.statesExperimental')
+   * @returns Object with section and property
+   */
+  parseKey(key: string): { section: string; property: string } {
+    const dotIndex = key.indexOf('.');
+    if (dotIndex === -1 || dotIndex === 0 || dotIndex === key.length - 1) {
```
cdrage, line 38: I find this a bit odd to understand, maybe we add comments or simplify this?

cdrage (approved):
Minor comments, otherwise I tested this out against https://github.com/podman-desktop/podman-desktop/pull/11992 and it worked well! 

Approving as my comments are non-blockers.

packages/main/src/plugin/experimental-configuration-manager.ts
gastoner: Looks like this hack is no longer needed (For some reason when I was setting the directly `{}` it used to fail for some reason and this was fixing it). But now it looks like it is no longer the case :D

packages/main/src/plugin/experimental-configuration-manager.ts
gastoner: Added

jeffmaury (approved):
LGTM

PR #13317, opened 2025-07-23
feat: adds experimental features feedback handler

gastoner:
### What does this PR do?
Adds Experimental features feedback manager
In another PR will be added showDialog function that takes care of showing the feedback dialog so user could chose a notify option

### How to test this PR?
Enable some experimental settings (in settings should be e.g. "some.feature: true")
Start Podman Desktop
The config should change to 
```
some.feature: {
  remindAt: TIMESTAMP
  disabled: false
}
```

### DIFF
--- a/packages/main/src/plugin/experimental-feature-feedback-form.ts
+++ b/packages/main/src/plugin/experimental-feature-feedback-form.ts
@@ -1,0 +1,178 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { inject, injectable } from 'inversify';
+
+import { IConfigurationRegistry } from '/@api/configuration/models.js';
+import type { IDisposable } from '/@api/disposable.js';
+
+import { ConfigurationRegistry } from './configuration-registry.js';
+
+export type Timestamp = number | undefined;
+
+const DAYS_TO_MS = 24 * 60 * 60 * 1_000;
+
+@injectable()
+export class ExperimentalFeatureFeedbackForm {
+  #disposables: IDisposable[] = [];
+  protected timestamps: Map<string, Timestamp> = new Map();
+  protected disabled: string[] = [];
+  protected experimentalFeatures: Set<string> = new Set([]);
+  readonly #configurationRegistry: IConfigurationRegistry;
+  constructor(
+    @inject(ConfigurationRegistry)
+    private configurationRegistry: ConfigurationRegistry,
+  ) {
+    this.#configurationRegistry = this.configurationRegistry;
+  }
+
+  protected async save(key: string): Promise<void> {
+    // If timestamp does not exist, the feature is not enabled
+    const timestamp: Timestamp = this.timestamps.get(key);
+    let disabled: boolean | undefined = this.disabled.includes(key);
+    if (!timestamp) {
+      // Timestamp does not exist we need to set disabled manually, otherwise we would end up with:
+      // feature {
+      //  "disabled" : false
+      // }
+      disabled = undefined;
+    }
+    const conf =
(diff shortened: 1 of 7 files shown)


packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,178 @@
...
+import { inject, injectable } from 'inversify';
+
+import { IConfigurationRegistry } from '/@api/configuration/models.js';
+import type { IDisposable } from '/@api/disposable.js';
+
+import { ConfigurationRegistry } from './configuration-registry.js';
+
+export type Timestamp = number | undefined;
+
+const DAYS_TO_MS = 24 * 60 * 60 * 1_000;
```
MarsKubeX: Non blocking: I think maybe we should add this constant in specific constants file for the plugin folder. I've checked and in the telemetry.ts file in the plugin folder there is another use of this constant:
` if (diff > 24 * 60 * 60 * 1000) {
              this.lastTimeEvents.set(eventName, now);
            } else {
              dropIt = true;
            }
          }`

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Updates timestamp for given experimental feature
+   * @param feature in format feature.name
+   * @param days timeout in days, undefined -> disabled, 'disabled' -> "Don't show again" was selected
+   */
+  protected setTimestamp(feature: string, days: Timestamp): void {
+    let date: Timestamp = days;
+    if (typeof days === 'number') {
+      date = new Date(new Date().getTime() + days * DAYS_TO_MS).getTime();
```
MarsKubeX: Non blocking: You could maybe consider to put this line like:

`date = Date.now() + (days * DAYS_TO_MS)`

I think it is a bit clearer.

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,178 @@
...
+      else this.setTimestamp(configurationName, undefined);
+    }
+  }
+
+  /**
+   * Replaces dots by spaces and adds uppercase on each sequence
+   * @param id in format feature.name
+   * @returns nicely formatted name e.g. feature Name
+   */
+  protected formatName(id: string): string {
```
MarsKubeX: Could this function be in an util file?

gastoner:
@MarsKubeX I've updated the PR moving the constant into another constant file could be an another PR

packages/main/src/plugin/experimental-feature-feedback-form.ts
```diff
@@ -0,0 +1,166 @@
...
+import type { IDisposable } from '/@api/disposable.js';
+
+import { ConfigurationRegistry } from './configuration-registry.js';
+
+export type Timestamp = number | undefined;
+
+const DAYS_TO_MS = 24 * 60 * 60 * 1_000;
+
+@injectable()
+export class ExperimentalFeatureFeedbackForm {
```
jeffmaury: I found the name misleading xxxForm is more related to a UI component so what is the exact purpose of this component ?

gastoner: In another PR there will be added a function that will be showing the form/dialog and further process the selected options from the user. But I can rename it e.g. to **-manager. I'm not good at naming :D

MarsKubeX (approved):
LGTM

jeffmaury (commented):
Still questionning the naming.

Also info about a feature is stored in several places wouldn't it be more logical and clear to have it store at a single place under a single object

packages/main/src/plugin/experimental-feature-feedback-handler.ts
```diff
@@ -0,0 +1,150 @@
...
+          disabled: optionDisabled,
+        };
+
+        this.experimentalFeatures.set(configurationKey, configuration);
+      } else {
+        // when started for first time, we need to set the timetamps for each experimental feature
+        this.setReminder(configurationKey);
+      }
+    }
+    // When are all features set, show dialog
```
cdrage, line 103: Lingering comment? Shows comment but no code underneath to show dialog?

cdrage (changes requested):
Tested and works well, no issues on the practical side (via the instructions on the PR).

One comment though, I feel as though there is a lingering comment? Is it suppose to "show" the notification in this PR?

packages/main/src/plugin/experimental-feature-feedback-handler.ts
gastoner: This is expected, the function itself will be added in another PR. Florent wanted to split this as backend part and "frontend" part

cdrage (approved):
LGTM from my side after curating through the code (excellent tests by the way!)

packages/main/src/plugin/experimental-feature-feedback-handler.ts
cdrage: ACK

PR #13325, opened 2025-07-24
fix: remove not needed timeout around registerCLITool in compose extension activation

dgolovin:
### What does this PR do?

PR removes setTimeout call that triggers async function registerCLITool call. setTimeout is not required because async function call without wait does not hold exit from extension's activation function.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?


gastoner (approved):
Codewise LGTM

PR #13339, opened 2025-07-24
chore(ci): enable prereleases in update e2e tests on PR check

### LINKED ISSUE #13338: Improve update e2e tests in PR check with possibility to use prerelease version

As part of investigation done on how we could improve the update e2e tests, we can add parameters to change the target update and add option to our workflows to allow prereleases. 

Add input parameters to change `owner` and `repo` properties in `dist/win-unpacked/resources` if they are present.

Add switch that adds manually `autoUpdater.allowPrerelease=true` into `updater.ts` module to test prereleases during update e2e tests.

_No response_

### Additional context

_No response_

odockal:
### What does this PR do?
Update e2e tests that runs on pr-check (if triggered) to use `testing-prereleases` github pre-release. This way we can see if electron-builder/updater could break updating of particular update.

Uses inline code insertion to allow prereleases in `updater.ts`.

#13338

### How to test this PR?

All update e2e tests should be passing on pr-check.

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -431,8 +431,12 @@ jobs:
   win-update-e2e-test:
     name: win update e2e tests
     needs: detect_pnpm_changes
+    strategy:
+      fail-fast: false
+      matrix: 
+        os: [windows-2025]
     if: contains(github.event.pull_request.labels.*.name, 'area/update') || needs.detect_pnpm_changes.outputs.pnpm_lock_changed == 'true'
-    runs-on: windows-2025
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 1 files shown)


.github/workflows/daily-testing-build.yaml
```diff
@@ -134,17 +134,17 @@ jobs:
       - name: Run Build on Ubuntu
         if: ${{ matrix.os == 'ubuntu-24.04' }}
         timeout-minutes: 40
-        run: pnpm compile:next --linux tar.gz
```
gastoner: Question: Why do we remove this format?

.github/workflows/pr-check.yaml
```diff
@@ -431,8 +431,12 @@ jobs:
   win-update-e2e-test:
     name: win update e2e tests
     needs: detect_pnpm_changes
+    strategy:
+      fail-fast: false
+      matrix: 
+        os: [windows-2025, windows-11-arm]
```
gastoner: Is it possible to do something similar on Mac and Linux?

.github/workflows/pr-check.yaml
```diff
@@ -452,20 +456,34 @@ jobs:
...
+          $runnerArch=$env:RUNNER_ARCH
+          $unpackedPath = "dist/win-unpacked"
+          if ($runnerArch -eq 'ARM64') {
+            $unpackedPath = "dist/win-arm64-unpacked"
+          }
+          $path=("./$unpackedPath/Podman Desktop.exe" | resolve-path).ProviderPath
           echo $path
           echo ("PODMAN_DESKTOP_BINARY=" + $path) >> $env:GITHUB_ENV
+          echo "Replace app-update.yml owner and repo to a prerelease"
+          # (Get-Content $unpackedPath/resources/app-update.yml).Replace('owner: podman-desktop', 'owner: odockal') | Set-Content $unpackedPath/resources/app-update.yml
```
gastoner: Is this change required? When creating pre-relases those ale already present in PD repo, why cant we use those?

gastoner (commented):
I probably dont uderstand how this works in details but I find it weird that we are setting `autoUpdater.allowPrerelease=true;` in two places
1. using replace 
2. directly updating `updater.ts` file 
So I feel like one of them is not required?

.github/workflows/pr-check.yaml
odockal: Plenty of commented out code... still draft ;)

.github/workflows/pr-check.yaml
odockal: we do not provide update on linux. It is done via flathub itself. Flatpak and binary means are not supported: 
https://github.com/podman-desktop/podman-desktop/blob/main/packages/main/src/plugin/updater.ts#L461. 
which does not mean it would not work, right?

odockal: I will be having two other issues to expand tests on windows-arm (I will remove it from this PR) and mac intel.

.github/workflows/daily-testing-build.yaml
odockal: This is commit from other PR, which should be merge (https://github.com/podman-desktop/podman-desktop/pull/13341) before this one can be. After we merge the other, in some time (assuming I did not make a mistake) we will have more artifacts in testing-prereleases repository, which will be used as update target since `podman-desktop/prereleases` releases are produced often and causes failure when we try to run update when tag exists but artifacts are not uploaded yet.

odockal:
> I probably dont uderstand how this works in details but I find it weird that we are setting `autoUpdater.allowPrerelease=true;` in two places
> 
>     1. using replace
> 
>     2. directly updating `updater.ts` file
>        So I feel like one of them is not required?



> I probably dont uderstand how this works in details but I find it weird that we are setting `autoUpdater.allowPrerelease=true;` in two places
> 
>     1. using replace
> 
>     2. directly updating `updater.ts` file
>        So I feel like one of them is not required?

Correct, only one method is now at place. Inline insertion.

.github/workflows/pr-check.yaml
```diff
@@ -516,29 +533,44 @@ jobs:
...
 
-      - name: Build Podman Desktop locally with electron updater included
+      - name: Build Podman Desktop locally and allow prereleases on update
         env:
           ELECTRON_ENABLE_INSPECT: true
         run: |
+          echo "Allow Prerelease when updating, setting it inline in the code"
+          sed -i '' '/autoUpdater.autoDownload = false;/a \
+          autoUpdater.allowPrerelease = true; \
+          ' packages/main/src/plugin/updater.ts
```
gastoner, line 551: Cant we use same code for editing the file?
Here we are using `sed` and above is `Get-Content` and `Replace`

gastoner (commented):
Otherwise LGTM

.github/workflows/pr-check.yaml
odockal: Mac has default bash shell, windows has default powershell for a shell... I do not like it either, tbh. We would need to change the shell for particular runner manually, ie. using cygwin action. Then it would be possible. But since it is not default, I would stick to using what is there already, because we cannot make it default globally (I think).

danivilla9 (approved):
LGTM

cbr7 (approved):

gastoner (approved):
LGTM codewise

Issue #9579, opened 2024-10-24
Review UX/UI for the toast notifications


slemeur:
### UX Description

We are introducing the toast notifications as an experimental capability in https://github.com/containers/podman-desktop/pull/9488 

The goal of this ticket, would be to review the UX and the UI of the toast and provide a list of enhancements we could accomplish, in particular:
- Behaviors for corner cases: length of the title, length of the description
- Styling consistency: border for the toast? drop shadow? font styles, loader consistency
- Toast Interaction: possibility to close the notification
- Interactions with the Task Manager: Should clicking on the task manager release and remove the toast

### Request type

UX analysis/suggestions for improvement

### Primary Contact

@slemeur 

### Deadline for request

Sprint 265

benoitf:
- z-index order (notification can be displayed on top of other elements or behind some of them)
 
- appearance: always on, disappear on success, disappear after few seconds, reappear on success

- duplication of information between status-bar, task manager and toasts

ekidneyrh:
current: 
![Image](https://github.com/user-attachments/assets/f2438ae9-164f-4367-bc39-e1efd297284c)

**Purpose**:  I would suggest renaming the notification to "Task started" or "Job initiated" to avoid repeating "Building image." The toast should focus on notifying users of task initiation and provide three concise lines of task details.

**Size**: Limit the size to fit typical task descriptions & cap longer task descriptions to avoid clutter - anything longer than the example. 

**Style**: Use the same font we usually use, with a bold title and regular body, and update the loader to match the status bar spinner.

**Duration and Interactions**: Display the toast for 3–5 seconds, allowing dismissal via a close button. Clicking it should navigate to the task source and close the toast.

proposed:
![Image](https://github.com/user-attachments/assets/ed357efe-2e89-4940-81ee-cbbff9ebbc4a)

slemeur:
Thanks for the feedback @ekidneyrh.

I don't think the title of the task should be "Task started" or "Job initiated". While I agree with the repetition, it also creates another repetitive effect when the user will have multiple tasks started at the same time. 

For the other item:
- **Size:** Limiting the size to fit the toast notification might not always be possible. What is your suggestion when this case happen? Should we append `...` at the end of the message, display a tooltip over the toast?

- **Style:**  Shouldn't the description of the text be a lighter color? Also, is the position of the progress at the right place?

- **Duration and Interactions:** ACK

benoitf:
[...]

PR #13355, opened 2025-07-25
chore: update toast notification style

### LINKED ISSUE #11151: Implementation of new styles for the Task Notifications

In this task, we would like to improve the UI and UX of the tasks notifications.

UX has been done in this validated ticket: https://github.com/podman-desktop/podman-desktop/issues/9579#issuecomment-2604550568

We would like to implement this new style. 

_No response_

### Additional context

_No response_

SoniaSandler:
### What does this PR do?
Update the style of the task notification toast based on https://github.com/podman-desktop/podman-desktop/issues/9579#issuecomment-2604550568:

### Screenshot / video of UI
Screenshot 2025-08-01 at 10 02 00 AM

Closes https://github.com/podman-desktop/podman-desktop/issues/11151

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/toast/ToastCustomUi.svelte
+++ b/packages/renderer/src/lib/toast/ToastCustomUi.svelte
@@ -16,7 +16,7 @@ let { toastId, taskInfo, onpop = (): void => {} }: Props = $props();
 
 const taskStatus: { [taskInfo.status]: string } = {
   failure: 'Error',
-  canceled: 'Cancelled',
+  canceled: 'Canceled',
 };
 
 function hideToast(): void {
@@ -35,7 +35,7 @@ const closeAction = (): void => {
 >
   <div class="flex flex-row gap-1">
     <div
-      class="mr-1 text-[var(--pd-state-info)] w-fit h-fit self-center w-[2em]"
+      class="mr-1 text-[var(--pd-state-info)] w-fit h-fit self-center"
       role="status"
       aria-label={taskInfo.status}
     >
@@ -46,13 +46,16 @@ const closeAction = (): void => {
       {:else if taskInfo.status === 'canceled'}
         <Fa icon={faCircleXmark} class="text-[var(--pd-state-error)] fa-xl"/>
       {:else if taskInfo.status === 'failure'}
-        <Fa icon={faTriangleExclamation } class=" fa fa-exclamation-triangle text-[var(--pd-state-warning)] fa-xl" />
+        <Fa icon={faTriangleExclamation } class="text-[var(--pd-state-warning)] fa-xl" />
       {/if}
     </div>
 
-    <p class="text-base text-ellipsis text-[var(--pd-modal-text)] line-clamp-3 h-full overflow-hidden break-all">
+    <div class="text-base text-ellipsis text-[var(--pd-card-text)] line-clamp-3 h-full overflow-hidden break-all">
       {taskStatus[taskInfo.status] ?? ''} {taskInfo.name}
-    </p>
+      {#if taskInfo.error}
+        <p class="text-[var(--pd-content-text)]">{taskInfo.error}</p>
+      {/if}
+    </div>
   </div>
 
   <div class="flex flex-none whitespace-nowrap flex-col self-start w-fit">
--- a/packages/renderer/src/lib/toast/ToastCustomUi.svelte
+++ b/packages/renderer/src/lib/toast/ToastCustomUi.svelte
@@ -33,14 +33,14 @@ const closeAction = (): void => {
   class="flex flex-nowrap min-h-10 cursor-default max-h-30 max-w-[var(--toastWidth)] flex-row p-2 border-[var(--pd-content-divider)] border rounded-md bg-[var(--pd-modal-bg)] gap-2 justify-between text-base"
   title={taskInfo.name}
 >
-  <div class="flex flex-row gap-1">
+  <div class="flex flex-row gap-1 items-center">
     <div
       class="mr-1 text-[var(--pd-state-info)] w-fit h-fit self-center"
       role="status"
       aria-label={taskInfo.status}
     >
       {#if taskInfo.status === 'in-progress'}
-        <Spinner size="2em"/>
+        <Spinner size="1.5em"/>
       {:else if taskInfo.status === 'success'}
         <Fa icon={faCheckCircle} class="text-[var(--pd-state-success)] fa-xl"/>
       {:else if taskInfo.status === 'canceled'}
@@ -50,7 +50,7 @@ const closeAction = (): void => {
       {/if}
     </div>
 
-    <div class="text-base text-ellipsis text-[var(--pd-card-text)] line-clamp-3 h-full overflow-hidden break-all">
+    <div class="text-base text-ellipsis text-[var(--pd-card-text)] line-clamp-3 h-fit overflow-hidden break-all">
       {taskStatus[taskInfo.status] ?? ''} {taskInfo.name}
       {#if taskInfo.error}
         <p class="text-[var(--pd-content-text)]">{taskInfo.error}</p>
--- a/packages/renderer/src/lib/toast/ToastCustomUi.svelte
+++ b/packages/renderer/src/lib/toast/ToastCustomUi.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
-import { faCheckCircle, faCircleXmark } from '@fortawesome/free-regular-svg-icons';
-import { faTriangleExclamation } from '@fortawesome/free-solid-svg-icons';
-import { CloseButton, Link, Spinner } from '@podman-desktop/ui-svelte';
+import { faCheckCircle, faCircleXmark, faTriangleExclamation } from '@fortawesome/free-solid-svg-icons';
+import { CloseButton, Spinner } from '@podman-desktop/ui-svelte';
 import { toast } from '@zerodevx/svelte-toast';
 import Fa from 'svelte-fa';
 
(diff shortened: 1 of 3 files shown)


SoniaSandler:
@MariaLeonova @slemeur A few of notes about this PR:
- Since it was not included in the design ticket, the action link was removed from the toast
- The blue progress line cannot be removed unless we change the toast to not disappear after certain time
- One of the icons was not available to use in the suggested style, so I changed the icons to a style that was available for all of them
- Task names are one string and anything can be written in them, so it's difficult (and maybe even impossible) to have the "title" as one text color and the "description" as another one as show in the design ticket since there is no clear separation between them in the task name. To achieve that, we'll probably have to change the task API to have task title and description separately, but I think it will be a bit too much change just for different colors in the toasts.
- The was no discussion about tasks with error messages, so I used the suggested gray[700] for error messages
- The line limit for the toasts is 3 line, we can increase it or just not use it at all.

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -22,58 +27,39 @@ const closeAction = (): void => {
...
+  <div class="flex flex-row gap-1">
     <div
-      class="mr-1 text-[var(--pd-state-info)]"
+      class="mr-1 text-[var(--pd-state-info)] w-fit h-fit self-center"
       role="status"
       aria-label={taskInfo.status}
     >
       {#if taskInfo.status === 'in-progress'}
-        <Spinner size="1em"/>
+        <Spinner size="2em"/>
```
gastoner: I feel like this make the spinner too big compared to other icons

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -22,58 +27,39 @@ const closeAction = (): void => {
...
       {:else if taskInfo.status === 'success'}
-        <Fa icon={faCheckCircle} />
+        <Fa icon={faCheckCircle} class="text-[var(--pd-state-success)] fa-xl"/>
+      {:else if taskInfo.status === 'canceled'}
+        <Fa icon={faCircleXmark} class="text-[var(--pd-state-error)] fa-xl"/>
       {:else if taskInfo.status === 'failure'}
-        <Fa icon={faCircleExclamation} class="text-[var(--pd-state-error)]" />
+        <Fa icon={faTriangleExclamation } class="text-[var(--pd-state-warning)] fa-xl" />
```
gastoner, line 49: When I try to create e.g. same kind cluster 2 times I see error notification, but with yellow triangle, I would expect to be red as previously

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -22,58 +27,39 @@ const closeAction = (): void => {
...
-  await window.executeTask(taskInfo.id);
-};
 </script>
 
 <div
-  class="flex min-h-10 cursor-default max-h-30 max-w-[var(--toastWidth)] flex-col p-2 border-[var(--pd-button-tab-border-selected)] border rounded-md bg-[var(--pd-modal-bg)]"
+  class="flex flex-nowrap min-h-10 cursor-default max-h-30 max-w-[var(--toastWidth)] flex-row p-2 border-[var(--pd-content-divider)] border rounded-md bg-[var(--pd-modal-bg)] gap-2 justify-between text-base"
   title={taskInfo.name}
 >
-  <div class="mb-1 flex flex-row items-center">
```
gastoner, line 36: Is possible to make the title of the text e.g. when having success :heavy_check_mark: the icon seems to be lower compared to text, same with the spinner

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -22,58 +27,39 @@ const closeAction = (): void => {
...
-
-const executeAction = async (): Promise<void> => {
-  hideToast();
-  await window.executeTask(taskInfo.id);
-};
 </script>
 
 <div
-  class="flex min-h-10 cursor-default max-h-30 max-w-[var(--toastWidth)] flex-col p-2 border-[var(--pd-button-tab-border-selected)] border rounded-md bg-[var(--pd-modal-bg)]"
+  class="flex flex-nowrap min-h-10 cursor-default max-h-30 max-w-[var(--toastWidth)] flex-row p-2 border-[var(--pd-content-divider)] border rounded-md bg-[var(--pd-modal-bg)] gap-2 justify-between text-base"
```
gastoner: Not sure which text was previously there, but I would make it smaller as it was previously, now it seems too large for me

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -22,58 +27,39 @@ const closeAction = (): void => {
...
-    {/if}
+
+  <div class="flex flex-none whitespace-nowrap flex-col self-start w-fit">
+    <CloseButton class="text-[var(--pd-modal-text)]" onclick={closeAction} />
   </div>
-  {#if taskInfo.action}
-    <div class="text-right text-xs text-[var(--pd-content-text)]">
-      <Link onclick={executeAction}>{taskInfo.action}</Link>
-    </div>
-  {/if}
```
gastoner, line 78: Do we want to really remove this?

gastoner (commented):
Just color/size preferences

packages/renderer/src/lib/toast/ToastCustomUi.svelte
SoniaSandler: That was the design in the design ticket, but I agree with you that red is better for errors. I'll wait for some more input and feedback from UX to see if I should change it

packages/renderer/src/lib/toast/ToastCustomUi.svelte
SoniaSandler: Based on the design ticket, the text should be size `text-base`

packages/renderer/src/lib/toast/ToastCustomUi.svelte
SoniaSandler: >Since it was not included in the design ticket, the action link was removed from the toast

I'm waiting to get some feedback from Masha or Stevan since it wasn't included in the design ticket to see if we want to keep it or remove it

gastoner: Sure, lets wait for some feedback from them

gastoner:
Thanks for the changes with size/ alignment, looks much better now :rocket:

vancura:
Hey, thanks for the changes, much welcomed!

We (@MariaLeonova) discussed this on Monday, and we couldn't find the initial design that this was built from. Masha hasn't seen this before. Do you have any design you used, please?

Also, please remember to always mention the design team (@MariaLeonova or me) or assign the `UX/UI Issue` label. It's challenging to be aware of these changes if we're not notified.

I have two minor design requests (in a minute), and I hope they won't be too difficult to implement.

vancura:
Great work, @SoniaSandler, on getting this implemented! I have a few design tweaks that will help polish this feature. Don't worry – you're doing a great job!

_Note about screenshots: before state always on top, after lower._

- [x] **Progress bar thickness**: Could we try making the progress bar a bit thinner? Currently, it's `6pt`, but `3pt` would give us more breathing room in the layout. Here's what I'm thinking:

image

---

- [x] **Icon alignment**: Would it be possible to top-align the icon? This will help with the visual balance:

image

---

- [x] **Text balance**: Let's try using Tailwind's `text-balance` (`text-wrap: balance;`) for the label. This should help with text distribution – worth a shot!

image

---

- [x] **Toast readability**: I noticed the original state is getting a bit cramped and hard to read. This is a common challenge with toast notifications! Here's what we're dealing with:

image

A few things that would help:
- **Word wrapping**: Can we prevent words from breaking mid-word (like "tryi-ng")? Keeping words intact makes it much more readable.
- **Text truncation**: Instead of ending with "Fai...", would it be possible to show the full text?
- **Toast height**: To accommodate the complete text, the toast might need to be a bit taller.

Here's the goal we're aiming for:

image

---

I know this seems like a lot of feedback, but you're handling complex UI challenges really well! Please don't hesitate to ask questions if anything isn't clear. I'm happy to help troubleshoot if you run into any tricky parts, just ping me.

Thanks for your patience with all these details.

vancura (changes requested):
Please try to implement my comments from above. Thank you!

SoniaSandler:
Hi @vancura, thanks for the feedback! Here is the original design https://github.com/podman-desktop/podman-desktop/issues/9579#issuecomment-2604550568.
As for the progress bar, I'm not sure if it something that we can change since it is built in from the external toast module that we are using. Maybe we could use a separate css style definition for that progress bar class, but I'll have to check if it works.
I also had a few questions/notes about the current UX/UI implementation compared to the original design. Could you let me know what you think about those? https://github.com/podman-desktop/podman-desktop/pull/13355#issuecomment-3128174219

vancura:
> As for the progress bar, I'm not sure if it something that we can change since it is built in from the external toast module that we are using. Maybe we could use a separate css style definition for that progress bar class, but I'll have to check if it works.

If it can't be done, no worries! It's a detail. We'll work on these components soon anyway, and if the progress bar is defined elsewhere, we'll take a look at it too (not a scope of this ticket, of course).

Re your questions from above:

> Since it was not included in the design ticket, the action link was removed from the toast

That makes sense. In the end, it's not possible to activate the toast when the user has no pointing devices. Toasts are not accessible, as they require vision – and if there's a link, it's impossible to use it for people with visual or motion impairments.

> The blue progress line cannot be removed unless we change the toast to not disappear after certain time

That's a topic for discussion that we should have. /c @MariaLeonova 

> One of the icons was not available to use in the suggested style, so I changed the icons to a style that was available for all of them

Thanks! This works well.

> The was no discussion about tasks with error messages, so I used the suggested gray[700] for error messages

It works well. The icon is what matters in this case.

> The line limit for the toasts is 3 line, we can increase it or just not use it at all.

Can we use more lines based on the message length?

SoniaSandler:
@vancura 
> Can we use more lines based on the message length?

I can remove the line limit and just make the toast height fit the whole text

SoniaSandler:
@vancura Also, based on the design ticket, the error icon is a yellow exclamation triangle, which I feel (based on color) is more suitable for warning. However, the color red is already used to indicate canceled tasks. What do you think?

packages/renderer/src/lib/toast/ToastCustomUi.svelte
```diff
@@ -22,58 +27,39 @@ const closeAction = (): void => {
...
-
-const executeAction = async (): Promise<void> => {
-  hideToast();
-  await window.executeTask(taskInfo.id);
-};
 </script>
 
 <div
-  class="flex min-h-10 cursor-default max-h-30 max-w-[var(--toastWidth)] flex-col p-2 border-[var(--pd-button-tab-border-selected)] border rounded-md bg-[var(--pd-modal-bg)]"
+  class="flex flex-nowrap min-h-10 cursor-default max-h-50 max-w-[var(--toastWidth)] flex-row p-2 border-[var(--pd-content-divider)] border rounded-md bg-[var(--pd-modal-bg)] gap-2 justify-between text-base"
```
SoniaSandler: @vancura I increased the max height to accommodate more text (for example the kind error). However, we never know how much text will need to be displayed in the toast.
Do you think it would be better to just remove the toast's height limit, increase it more, or leave it as it is now (fits the kind error)?

SoniaSandler:
@vancura I have applied the requested changes (the progress bar height too :)) and updated the screenshots with the latest changes. Please let me know if there are other changes that should be done

gastoner:
@SoniaSandler Looking great, I would maybe update the Error notification with red and smaller description? cc @vancura

jiridostal (changes requested):
Looking clean now! The error should be red and use different icon, an example of the mismatch can be seen here:
image

packages/renderer/src/lib/toast/ToastCustomUi.svelte
vancura: It's better to be safe than sorry. `max-h-50` looks reasonable to me, I hope we never cross that limit, that would be a sign of other problems :)

vancura:
> Also, based on the design ticket, the error icon is a yellow exclamation triangle, which I feel (based on color) is more suitable for warning. However, the color red is already used to indicate canceled tasks. What do you think?

Please use the error icon @jiridostal pointed at, that one is precisely what we need.

Thanks a lot for the changes! The component now looks perfect :)

vancura:
One thing, please – are we 100% sure there's nothing that could be clicked in the toast label? (that would be inaccessible, anyway). If there's nothing that could be clicked, add `select-none pointer-events-auto cursor-default` to the container. That prevents text selection and ensures the toast blocks mouse events from reaching elements behind it. We don't want the users to select text in the toast; that would be useless.

SoniaSandler:
> Please use the error icon @jiridostal pointed at, that one is precisely what we need.

@vancura This is similar to the one that is used for canceled tasks (red circle with x). If I use the red circle with exclamation mark for error, should I leave the red circle with x mark for canceled or maybe switch to the yellow exclamation triangle?

gastoner (approved):
LGTM, just still missing the "open task" cc @vancura

jiridostal (approved):

SoniaSandler:
> > Since it was not included in the design ticket, the action link was removed from the toast
> 
> That makes sense. In the end, it's not possible to activate the toast when the user has no pointing devices. Toasts are not accessible, as they require vision – and if there's a link, it's impossible to use it for people with visual or motion impairments.
> 
@gastoner

vancura:
@SoniaSandler: @gastoner is right, we still need the task button. In that case, we need to be careful about accessibility – the button needs to be well-clickable, with a large enough hit area.

I am still worried about the accessibility of the toast in general; it's impossible to click that button without a pointing device. In that case, I think it would be good to show something like:

image

This would make the toast accessible to everyone. I am not sure about the keyboard shortcut, though, but it can really be just temporary while the toast is displayed.

Thoughts?

SoniaSandler:
@vancura I checked the shortcut with those keys, and by the time I'm able to press all of them, the toast is less than a second away from disappearing. Maybe we could use only one key for the task (`t` for task)? Although I'm not sure if it is acceptable to do a keyboard shortcut with only a letter key

vancura:
Works for me, but I think @benoitf would have some details about the problem.

SoniaSandler:
@vancura if we have multiple toasts at the same time with the task links, wouldn't using the keyboard shortcut once "click" all the task links? How can we distinguish between them with the shortcut if for example I want to use only the second toast's link using the shortcut?

vancura:
Let's discuss this on the UX call this week (tomorrow, 2025-08-07).

vancura:
> If we have multiple toasts at the same time with the task links, wouldn't using the keyboard shortcut once "click" all the task links? How can we distinguish between them with the shortcut if for example I want to use only the second toast's link using the shortcut?

You have an excellent point. Unfortunately, toasts are not very accessible, and this clearly shows why. There's no clean way to disambiguate which toast's link should be activated. We should abandon the keyboard shortcut idea for toast links.

Instead, let's focus on making the toasts purely informational and direct users to a dedicated task management area for interactive actions. This follows accessibility best practices – toasts should announce information, not require interaction.

Please disregard my entire request to add the keyboard shortcuts. This won't work within the scope of this task. We (as @MariaLeonova and I) will discuss this when we get to the toast component again; hopefully, we'll figure out something.

SoniaSandler:
@vancura so for now we won't include the task link in the toast? 
> Please disregard my entire request to add the keyboard shortcuts. This won't work within the scope of this task. We (as @MariaLeonova and I) will discuss this when we get to the toast component again; hopefully, we'll figure out something.

Should I put this on hold for now or merge what I have?

vancura:
No no, the link could be there (even if it's not accessible); it's just keyboard shortcuts that we can't mention there.

vancura (approved):
I forgot about this, sorry! LGTM!

PR #13366, opened 2025-07-28
fix(core): update installed extensions actions to svelte 5

### LINKED ISSUE #13365: Migrate installed extensions actions to svelte 5
### Bug description

Migrate installed extensions actions to svelte 5

`packages/renderer/src/lib/extensions/InstalledExtensionActions.svelte`

### Operating system

all

### Installation Method

None

### Version

1.19.2

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

simonrey1:
### What does this PR do?

This PR migrates the update installed extensions component to svelte 5

Fixes https://github.com/podman-desktop/podman-desktop/issues/13365

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/extensions/InstalledExtensionActions.svelte
+++ b/packages/renderer/src/lib/extensions/InstalledExtensionActions.svelte
@@ -4,10 +4,15 @@ import type { CombinedExtensionInfoUI } from '/@/stores/all-installed-extensions
 import InstalledExtensionCardLeftLifecycle from './InstalledExtensionCardLeftLifecycle.svelte';
 import InstalledExtensionCardLeftOnboardingAndProperties from './InstalledExtensionCardLeftOnboardingAndProperties.svelte';
 
-export let extension: CombinedExtensionInfoUI;
+interface Props {
+  extension: CombinedExtensionInfoUI;
+  className: string;
+}
+
+let { extension, className }: Props = $props();
 </script>
 
-<div class="text-center flex-row flex {$$props.class}">
-  <InstalledExtensionCardLeftLifecycle extension={extension} />
-  <InstalledExtensionCardLeftOnboardingAndProperties extension={extension} />
+<div class="text-center flex-row flex {className}">
+  <InstalledExtensionCardLeftLifecycle {extension} />
+  <InstalledExtensionCardLeftOnboardingAndProperties {extension} />
 </div>
--- a/packages/renderer/src/lib/extensions/InstalledExtensionActions.svelte
+++ b/packages/renderer/src/lib/extensions/InstalledExtensionActions.svelte
@@ -6,7 +6,7 @@ import InstalledExtensionCardLeftOnboardingAndProperties from './InstalledExtens
 
 interface Props {
   extension: CombinedExtensionInfoUI;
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/extensions/InstalledExtensionActions.svelte
```diff
@@ -4,10 +4,15 @@ import type { CombinedExtensionInfoUI } from '/@/stores/all-installed-extensions
 import InstalledExtensionCardLeftLifecycle from './InstalledExtensionCardLeftLifecycle.svelte';
 import InstalledExtensionCardLeftOnboardingAndProperties from './InstalledExtensionCardLeftOnboardingAndProperties.svelte';
 
-export let extension: CombinedExtensionInfoUI;
+interface Props {
+  extension: CombinedExtensionInfoUI;
+  className: string;
```
gastoner: I think you should be able to do a similar thing, and thanks to this change, you should be able to have changes only in this file? 

https://github.com/podman-desktop/podman-desktop/blob/5ac757f2c70127da7e242536f08429337dc854f7/packages/ui/src/lib/screen/EmptyScreen.svelte#L32

gastoner (changes requested):

packages/renderer/src/lib/extensions/InstalledExtensionActions.svelte
simonrey1: You are right, by using `class: className` we can make it work by only changing this file.
However this is confusing in my opinion because class is reserved. 
So I prefer to use a unique className attribute rather than renaming the attribute when entering the component.

gastoner: Sure, but AFAIK we are doing this in all other files so I would stick with the convention

simonrey1: e350ae634419dd044b8b737899b8450d39cc9d72

gastoner (approved):

MarsKubeX (approved):

jeffmaury (approved):
LGTM

packages/renderer/src/lib/extensions/InstalledExtensionActions.svelte
```diff
@@ -4,10 +4,15 @@ import type { CombinedExtensionInfoUI } from '/@/stores/all-installed-extensions
 import InstalledExtensionCardLeftLifecycle from './InstalledExtensionCardLeftLifecycle.svelte';
 import InstalledExtensionCardLeftOnboardingAndProperties from './InstalledExtensionCardLeftOnboardingAndProperties.svelte';
 
-export let extension: CombinedExtensionInfoUI;
+interface Props {
+  extension: CombinedExtensionInfoUI;
+  class: string;
```
axel7083: generally class attribute is optional, even if for our usage we always have one, we may still have it optional to match HTML component synthax

simonrey1: e179a73fc57b81325eccfbf1951f03b95e31e98a

PR #13380, opened 2025-07-28
chore: allow passing title to Icon component

### LINKED ISSUE #13370: add optional title attribute to Icon for testing purposes
_Originally posted by @eqqe in https://github.com/podman-desktop/podman-desktop/pull/13192#discussion_r2230561437_

Icon could benefit from having a title attribute so referencing it from tests will be more straightforward.

jiridostal:
the name is self-descriptive

The main reason for this is accessibility and a possiblity to properly locate icons within tests.

Closes: #13370

### DIFF
--- a/packages/ui/src/lib/icons/Icon.svelte
+++ b/packages/ui/src/lib/icons/Icon.svelte
@@ -9,9 +9,10 @@ interface Props {
   icon: IconDefinition | Component | string;
   size?: IconSize | number | string;
   class?: string;
+  title?: string;
 }
 
-let { icon, size, class: className }: Props = $props();
+let { icon, size, class: className, title }: Props = $props();
 
 const role = 'img';
 const IconComponent = icon;
@@ -20,16 +21,16 @@ const IconComponent = icon;
 
 {#if isFontAwesomeIcon(icon)}
     {#if typeof size === 'undefined' || isFontAwesomeSize(size)}
-        <Fa {icon} {size} class={className}/>
+        <Fa {icon} {size} class={className} {title}/>
     {/if}
 {:else if typeof icon === 'string'}
     <!-- fas fa- and far fa- for Font awesome icons -->
     <!-- -icon for extension icons e.g. 'kind-icon' -->
     {#if icon.startsWith('fas fa-') || icon.startsWith('far fa-') || icon.endsWith('-icon')}
-        <i class={`${icon} ${size} ${className}`} {role}></i>
+        <i class={`${icon} ${size} ${className}`} {role} {title}></i>
     {/if}
 {:else}
     {#if IconComponent && typeof IconComponent !== 'string' && !isFontAwesomeIcon(IconComponent)}
-        <span {role}><IconComponent class={className} {size}/></span>
+        <span {role} {title}><IconComponent class={className} {size}/></span>
     {/if}
 {/if}
--- a/packages/ui/src/lib/icons/Icon.spec.ts
+++ b/packages/ui/src/lib/icons/Icon.spec.ts
@@ -65,6 +65,13 @@ describe('font awesome', () => {
     const path = img.querySelector('path');
     expect(path).toHaveAttribute('fill', 'currentColor');
   });
+
+  test('icon should have title', () => {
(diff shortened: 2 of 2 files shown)


gastoner (approved):
LGTM codewise

packages/ui/src/lib/icons/Icon.svelte
```diff
@@ -20,16 +21,16 @@ const IconComponent = icon;
 
 {#if isFontAwesomeIcon(icon)}
     {#if typeof size === 'undefined' || isFontAwesomeSize(size)}
-        <Fa {icon} {size} class={className}/>
+        <Fa {icon} {size} class={className} {title}/>
```
simonrey1, line 24: Suggestion: use font-awesome icon name for the title. It would gain us time so that we do not have to add the title manually.

simonrey1:
Your solution works but it requires to add a title manually.

I would like the icon name to be used automatically so that it can be used in tests direclty:

```svelte
<script lang="ts">
import type { IconDefinition } from '@fortawesome/free-solid-svg-icons';
import { Fa, type IconSize } from 'svelte-fa';

import { isFontAwesomeIcon, isFontAwesomeSize } from '../utils/icon-utils';

interface Props {
  icon: IconDefinition | Component | string;
  size?: IconSize | number | string;
  class?: string;
}

let { icon, size, class: className }: Props = $props();

const role = 'img';
const IconComponent = icon;

</script>


{#if isFontAwesomeIcon(icon)}
    {#if typeof size === 'undefined' || isFontAwesomeSize(size)}
        <Fa {icon} {size} class={className} title={icon.iconName}/>
    {/if}
{:else if typeof icon === 'string'}
    <!-- fas fa- and far fa- for Font awesome icons -->
    <!-- -icon for extension icons e.g. 'kind-icon' -->
    {#if icon.startsWith('fas fa-') || icon.startsWith('far fa-') || icon.endsWith('-icon')}
        <i class={`${icon} ${size} ${className}`} {role} title={icon}></i>
    {/if}
{:else}
    {#if IconComponent && typeof IconComponent !== 'string' && !isFontAwesomeIcon(IconComponent)}
        <span {role}><IconComponent class={className} {size}/></span>
    {/if}
{/if}
```

jiridostal:
@eqqe while I agree with you, I think for the accessibility reason this may not be the best idea. I have another code change that uses the title for notification icons like this 
image
If we used the default icon name, the tooltip would be showing internal name of the icon, which is not very useful. 🤷 
Maybe using the provided one and default to the name if it was not provided? Not sure...

simonrey1:
I understand that using the title attribute with technical data could be misleading in general. If it is displayed in the tooltip, or even if it is read with accessibility tools. So I'm ok to keep this solution but to me it does not really solve the testing issue.

I have also noticed that the issue linked in this PR is not the right one. I think it should be #13370.

axel7083 (approved):

jiridostal:
> I understand that using the title attribute with technical data could be misleading in general. If it is displayed in the tooltip, or even if it is read with accessibility tools. So I'm ok to keep this solution but to me it does not really solve the testing issue.
> 
> I have also noticed that the issue linked in this PR is not the right one. I think it should be #13370.

Oh thats a good catch, thanks!

packages/ui/src/lib/icons/Icon.svelte
simonrey1: Closing this because a tooltip with the technical name will be displayed.
Screenshot 2025-07-29 at 09 01 04
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/title

vancura:
FYI, Tooltips are a user-facing feature and should display information that users care about (i.e., information they may want to read), not what developers need (for that, we have the Inspector).

Tooltip accessibility is a topic in its own right, and a whole can of worms. One issue on the list: they rely on a pointing device, which some people don't/can't use. Podman Desktop, in general, is very much pointer-based, so we have a long way to go to reach an accessible state, and tooltips are one of the things we need to work on (not within the scope of this issue).

PR #13382, opened 2025-07-28
chore(renderer/dialog): prepare dialog component migration to Svelte 5

### LINKED ISSUE #13381: Prepare dialog component migration to Svelte 5
### Bug description

Remove legacy createEventDispatcher in Dialog component to prepare for the migration to Svelte 5

### Operating system

all

### Installation Method

None

### Version

1.19.2

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

simonrey1:
### What does this PR do?

Remove legacy createEventDispatcher in Dialog component to prepare for the migration to Svelte 5

Fixes https://github.com/podman-desktop/podman-desktop/issues/13381

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/dialogs/Dialog.svelte
+++ b/packages/renderer/src/lib/dialogs/Dialog.svelte
@@ -3,10 +3,10 @@ import { CloseButton, Modal } from '@podman-desktop/ui-svelte';
 
 export let title: string;
 
-export let onclose: () => void;
+export let onclose: (() => void) | undefined = undefined;
 </script>
 
-<Modal name={title} on:close={onclose}>
+<Modal name={title} onclose={onclose}>
   <div class="flex items-center justify-between pl-4 pr-3 py-3 space-x-2 text-[var(--pd-modal-header-text)]">
     <slot name="icon" />
     <h1 class="grow text-lg font-bold capitalize">{title}</h1>
--- a/packages/renderer/src/lib/dialogs/Dialog.svelte
+++ b/packages/renderer/src/lib/dialogs/Dialog.svelte
@@ -1,14 +1,9 @@
 <script lang="ts">
 import { CloseButton, Modal } from '@podman-desktop/ui-svelte';
-import { createEventDispatcher } from 'svelte';
 
 export let title: string;
 
-const dispatch = createEventDispatcher();
-
-export let onclose: () => void = () => {
-  dispatch('close');
-};
+export let onclose: () => void;
 </script>
 
 <Modal name={title} on:close={onclose}>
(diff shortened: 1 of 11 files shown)


packages/renderer/src/lib/dialogs/Dialog.svelte
```diff
@@ -1,14 +1,9 @@
...
-import { createEventDispatcher } from 'svelte';
 
 export let title: string;
 
-const dispatch = createEventDispatcher();
-
-export let onclose: () => void = () => {
-  dispatch('close');
-};
+export let onclose: () => void;
```
feloy: To keep the same behaviour as other component supporting `onclose` (FormPage, DetailsPage, etc), I would define an empty function as default value

packages/renderer/src/lib/dialogs/DialogSpec.svelte
```diff
@@ -2,7 +2,7 @@
 import Dialog from './Dialog.svelte';
 </script>
 
-<Dialog title="Test dialog">
+<Dialog title="Test dialog" onclose={(): void => {}}>
```
feloy: This change would not be needed with previous change

packages/renderer/src/lib/dialogs/Dialog.svelte
simonrey1: I was wondering because it was always implemented appart from a unit test, so I thought maybe it should be compulsory. But I can provide the default value no problem.

MarsKubeX (approved):
LGTM

packages/renderer/src/lib/dialogs/Dialog.svelte
axel7083: Here the onclose function has a default implementation because we have a logic to handle, IMHO we should not define an empty function that does nothing as default, we should make the property optional, and use `onclose?.()` when using it.

axel7083 (approved):

packages/renderer/src/lib/dialogs/DialogSpec.svelte
simonrey1: @feloy I rollbacked this but made a slighly different decision for the other comment. Let me know if it is ok for you.

feloy (approved):

feloy:
I wonder afterwards if having `onclose` optional with no defined default value is problematic or not. In other words, is having `<Modal name={title} onclose={onclose}>` problematic if `onclose` is undefined?

Do we want to have `onclose` mandatory, as @eqqe proposed first, or do we want to be sure that the `undefined` case is handled correctly?

@axel7083 wdyt?

simonrey1:
I am ok to revert 71fc03655427783459f449ea438a123d50694563 that followed https://github.com/podman-desktop/podman-desktop/pull/13382#discussion_r2237038173 in another PR, it is implemented everywhere already so no issue for me :)

axel7083:
No strong opinion after re-reading

PR #13399, opened 2025-07-29
test(renderer): try fix instability by clearing mock when unmounting

simonrey1:
### What does this PR do?

Try to fix UT instability by waiting for second call. Before this, we would wait for at least one call but it was already done above.

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
+++ b/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
@@ -263,7 +263,7 @@ test('prompt is not duplicated after restoring terminal from containerTerminals 
   let renderObject = render(ContainerDetailsTerminal, { container, screenReaderMode: true });
 
   // wait shellInContainerMock is called
-  await waitFor(() => expect(shellInContainerMock).toHaveBeenCalled());
+  await waitFor(() => expect(shellInContainerMock).toHaveBeenCalledOnce());
 
   // write some data on the terminal
   onDataCallback(Buffer.from('prompt$ \nhello\nworld\nprompt$ '));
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts
```diff
@@ -286,8 +286,8 @@ test('prompt is not duplicated after restoring terminal from containerTerminals
   // terminal.write
   renderObject = render(ContainerDetailsTerminal, { container, screenReaderMode: true });
 
-  // wait shellInContainerMock is called
-  await waitFor(() => expect(shellInContainerMock).toHaveBeenCalled());
+  // wait shellInContainerMock is called a second time
+  await waitFor(() => expect(shellInContainerMock).toHaveBeenCalledTimes(2));
```
axel7083: question: maybe we are not suppose to call this function twice?

simonrey1: There may be an optimization to do on the component side but currently it is called twice and the test is passing with this

axel7083: Okay we seems to unmount the component manually

https://github.com/podman-desktop/podman-desktop/blob/b60ecdc25ca2bc7d33432081e4bb6842ed80a712/packages/renderer/src/lib/container/ContainerDetailsTerminal.spec.ts#L282-L283

We may instead need to reset the mock, calling `vi.mocked(shellInContainerMock).mockClear();` after the unmount, as we seems to be confusing which call we are watiting for.

I would rather have a fix number, like `.haveBeenCalledOnce()`

axel7083 (approved):

odockal:
On the first spot, there seems to be less unit tests failures. But let's monitor the situation through another days.

PR #13422, opened 2025-07-30
chore(test): add new e2e test for bulk start action

### LINKED ISSUE #13346: Automate the bulk start of marked containers

New option to bulk start the containers was added in https://github.com/podman-desktop/podman-desktop/pull/12646.

Add e2e test to cover the scenario using playwright.

_No response_

### Additional context

_No response_

cbr7:
### What does this PR do?
Creates new e2e test to check the new bulk action start all containers feature.

https://github.com/podman-desktop/podman-desktop/issues/13346

### DIFF
--- a/tests/playwright/src/model/pages/containers-page.ts
+++ b/tests/playwright/src/model/pages/containers-page.ts
@@ -154,9 +154,8 @@ export class ContainersPage extends MainPage {
 
   async checkAllContainers(): Promise<void> {
     return test.step('Checks all containers', async () => {
-      let containersTable;
       try {
-        containersTable = await this.getTable();
+        const containersTable = await this.getTable();
         await playExpect(containersTable).toBeVisible();
         const controlRow = containersTable.getByRole('row').first();
         await playExpect(controlRow).toBeAttached();
--- a/tests/playwright/src/model/pages/containers-page.ts
+++ b/tests/playwright/src/model/pages/containers-page.ts
@@ -30,6 +30,7 @@ export class ContainersPage extends MainPage {
   readonly createContainerButton: Locator;
   readonly playKubernetesYAMLButton: Locator;
   readonly pruneConfirmationButton: Locator;
+  readonly runAllContainersButton: Locator;
 
   constructor(page: Page) {
     super(page, 'containers');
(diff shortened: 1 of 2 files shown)


danivilla9 (changes requested):
LGTM, but I think that the comments from the AI review should be addressed. The first adds consistency to the tests and is a good practice, while the second is a typo.

danivilla9 (approved):

tests/playwright/src/model/pages/containers-page.ts
```diff
@@ -45,6 +46,7 @@ export class ContainersPage extends MainPage {
     this.pruneConfirmationButton = this.page.getByRole('button', {
       name: 'Yes',
     });
+    this.runAllContainersButton = this.page.getByLabel('Run selected containers and pods');
```
odockal, line 49: please prefer using `getByRole`.

tests/playwright/src/model/pages/containers-page.ts
```diff
@@ -150,6 +152,31 @@ export class ContainersPage extends MainPage {
...
+        const containersTable = await this.getTable();
+        await playExpect(containersTable).toBeVisible();
+        const controlRow = containersTable.getByRole('row').first();
+        await playExpect(controlRow).toBeAttached();
+        const checkboxColumnHeader = controlRow.getByRole('columnheader').nth(1);
+        await playExpect(checkboxColumnHeader).toBeAttached();
+        const containersToggle = checkboxColumnHeader.getByTitle('Toggle all');
+        await playExpect(containersToggle).toBeAttached();
+
+        if ((await containersToggle.innerHTML()).includes('pd-input-checkbox-unchecked')) {
```
odockal, line 167: pufff. that is nasty. Though I do not see other way. On the other hand, this looks like something that all main pages (images, pods, etc) would make use of, can we extract it?

odockal (dismissed):

tests/playwright/src/model/pages/containers-page.ts
cbr7: There is no reason for that, that is why we have aria-labels, in order to use getByLabel, otherwise there is no point in having the field.

tests/playwright/src/model/pages/containers-page.ts
cbr7: I'm not sure about this, I just saw that that is how the uncheckAllContainers method was already implemented and just reversed it, I agree that it looks ugly, but it works very well.

If this is how it works in all pages, extracting the logic at a higher level might make sense.

serbangeorge-m (approved):

cbr7:
Created issue https://github.com/podman-desktop/podman-desktop/issues/13444 to track the checkbox handling at a higher level.

PR #13227, opened 2025-07-16
refactor: centralized experimental features

gastoner:
### What does this PR do?
Changes logic from storing boolean value to object/undefined in order to make work 
https://github.com/podman-desktop/podman-desktop/pull/13223
https://github.com/podman-desktop/podman-desktop/pull/13224
https://github.com/podman-desktop/podman-desktop/pull/13225
https://github.com/podman-desktop/podman-desktop/pull/13226

Uses new experimental configuration manager https://github.com/podman-desktop/podman-desktop/pull/13316

More info in https://github.com/podman-desktop/podman-desktop/issues/13095

### What issues does this PR fix or reference?
Closes #13096 
Needs rebase after https://github.com/podman-desktop/podman-desktop/pull/13316

THOSE PRs MUST be merged together:
https://github.com/podman-desktop/podman-desktop/pull/13225
https://github.com/podman-desktop/podman-desktop/pull/13224
https://github.com/podman-desktop/podman-desktop/pull/13223
https://github.com/podman-desktop/podman-desktop/pull/13226
https://github.com/podman-desktop/podman-desktop/pull/13227

### How to test this PR?

- [x] Tests are covering the bug fix or the new feature

PR #13446, opened 2025-07-31
fix: updating of configuration with undefined value

gastoner:
### What does this PR do?
Updates index from `section` to `localKey`
Previously the index was wrong and when trying to set undefined the value was not set

Required for https://github.com/podman-desktop/podman-desktop/pull/13227

### How to test this PR?
Unit tests 
OR 
https://github.com/podman-desktop/podman-desktop/pull/11992 
Where you need to go to experimental page and try to enable/disable some features with this change it should remove it from settings.json.
You can also try to switch back to "section" to see that it wont remove the configuration

### DIFF
--- a/packages/main/src/plugin/configuration-impl.spec.ts
+++ b/packages/main/src/plugin/configuration-impl.spec.ts
@@ -92,30 +92,3 @@ test('Uses localKey when deleting values', async () => {
     value: undefined,
   });
 });
-
-test.fails('Uses original update with section when deleting values', async () => {
-  // Test with globalSection to ensure localKey != section
-  const sendMock = vi.fn();
-  const map = new Map<string, { [key: string]: unknown }>();
-  const config = new TestConfigurationImpl({ send: sendMock } as unknown as ApiSenderType, vi.fn(), map, 'prefix');
-
-  // Set value
-  await config.originalUpdate('key', 'value');
-  // stored with section
-  expect(config['key']).toBe('value');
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/configuration-impl.spec.ts
```diff
@@ -46,3 +63,59 @@ test('Should callback on update with configuration key', async () => {
...
+
+  // THIS WILL FAIL, actualy it will be "value" and not undefined as expected
+  expect(config['key']).toBeUndefined();
+
+  // deleted using localKey
+  expect(config['prefix.key']).toBeUndefined();
+
+  // THIS WILL FAIL, actualy it will be "value" and not undefined as expected
+  expect(config.get('key')).toBeUndefined();
+});
```
feloy: I don't think you need this.

The test you are writing ('Uses localKey when deleting values') must fail without the change in the production code, and pass with the change. This is enough to demo the necessity of the change (and prevent regressions)

gastoner: Yes it does fail without mine changes :heavy_check_mark: 

I can remove this test for sure, just wanted to explicitly show that it is not working

simonrey1 (approved):

MarsKubeX (approved):
LGTM

gastoner:
Do not merge
@benoitf can you PTAL?

benoitf (approved):

Issue #13390, opened 2025-07-29
Update UI component container list to Svelte 5


simonrey1:
### Bug description

Update UI component container list to Svelte 5

### Operating system

all

### Installation Method

None

### Version

1.19.2

### Relevant log output

```shell

```

PR #2491, opened 2023-05-15
fix: container list is not keeping expand/collapse status

benoitf:
### What does this PR do?
use a store to keep the value across page change

### Screenshot/screencast of this PR

N/A

### What issues does this PR fix or reference?

fixes https://github.com/containers/podman-desktop/issues/1011

### How to test this PR?

Use test case of the issue

Change-Id: If79f201978b0ff02bac67e4aa93fe2d422e4810f

Issue #1011, opened 2022-12-12
Lists are not keeping expand/collapse status


slemeur:
### Bug description

Each time a list is getting "refreshed" the collapse/expand state of the items in the list are getting lost. 

See this one for example:

https://user-images.githubusercontent.com/1636769/206998930-95f17509-213d-4326-bca0-7cd9015c8349.mp4

### Operating system

Mac OS 12.6.1

### Version

0.10.0

### Steps to reproduce

**Case1: action on items from the list**
1. Start a Pod with multiple containers
2. Collapse the Pod from the list of containers
3. Start a container (or delete/stop one)
--> The collapsed pod is getting expand again

**Case2: switch to another screen of the app**
1. Start a Pod with multiple containers
2. Collapse the Pod from the list of containers
3. Go to the Pod list
4. Get back to the list of containers
--> The collapsed pod is getting expand again

slemeur:
I'm reopening this one as the navigation issue is back with 0.14.1


https://user-images.githubusercontent.com/1636769/233312328-818620f1-fca2-486c-bc2c-810529d24b12.mp4

simonrey1:
Reopening as I still see the issue on main branch

https://github.com/user-attachments/assets/ca942d2e-c2ec-4065-bad5-d9ecf4bf70d9

ZaLiTHkA:
I was just about to log a new issue for something similar, but then I found this one. not sure if what I found warrants a new issue or if it should be resolved along with this one... so let me start here first.

after updating to `1.20.2` just now, I noticed that collapsing and then expanding a container group seems to mess with the row layout.. I _don't think_ this was an issue in previous builds, but it might be as I'm honestly not sure when it started.

https://github.com/user-attachments/assets/9ebfc13c-3751-4464-9308-c25ea21f2e1b

should I rather log this separately, or would this be resolved as part of fixing this issue here?

slemeur:
Thanks @ZaLiTHkA for reporting the two issues. 

For the other issue you are experiencing, there is also an existing issue here: https://github.com/podman-desktop/podman-desktop/issues/13287 

Thanks a lot for jumping in the thread, super helpful!

ZaLiTHkA:
> For the other issue you are experiencing, there is also an existing issue here: [#13287](https://github.com/podman-desktop/podman-desktop/issues/13287)

ah, thank you.. I probably didn't find that because I wasn't searching for "shrinking" or "expanding". yay for English. (:

[...]

PR #13391, opened 2025-07-29
feat(renderer): update ContainerList to Svelte 5 

simonrey1:
### What does this PR do?

Update UI component container list to Svelte 5

Waiting for https://github.com/podman-desktop/podman-desktop/pull/13449 to be merged first

### What issues does this PR fix or reference?

Fixes #13390

### How to test this PR?

- [ ] Tests are covering the bug fix or the new feature

PR #13449, opened 2025-07-31
chore(renderer/containerList): remove unused code to restore group expansion

simonrey1:
### What does this PR do?

This code was added in PR #2491 by @benoitf but I don't think it is working anymore, and I would like to remove it to ensure the migration for #13390 works.

The issue https://github.com/podman-desktop/podman-desktop/issues/1011 still seems to exist with or without this code.

### How to test this PR?


benoitf (commented):
hello, I'm not sure to follow the goal. It was an ask to keep the state when moving screens
Here you remove the feature rather than fixing the issue ?

simonrey1:
I tried to move screen and the state is not persisted on the main branch. So I think we should remove this code and reopen the issue.

https://github.com/user-attachments/assets/ca942d2e-c2ec-4065-bad5-d9ecf4bf70d9

gastoner:
Hello, same thing as Florent. Wouldn't it be better to try to fix the issue rather than remove the code?

simonrey1:
> Hello, same thing as Florent. Wouldn't it be better to try to fix the issue rather than remove the code?

It's definitely preferable to fix the issue rather than remove the code. However, I'll be out of office for the next twenty days, and this is currently blocking a component migration.

The blocked Pull Request is #13391, and the underlying issue, #1011 has been a reccurring problem for about three years now, with several attempts at resolution.

The good news is that we don't have to migrate the existing code. We can address the issue by implementing the solution after the migration to Svelte 5 Runes mode. The new logic for derived stores in this mode could actually help to find a fix for #1011

gastoner (commented):
I think https://github.com/podman-desktop/podman-desktop/issues/1011 should be fixed after fix provided by Axel in https://github.com/podman-desktop/podman-desktop/pull/13496

gastoner (commented):
But if it is a blocker for you I agree that we could merge this so you could move forward

axel7083 (commented):
If the code is not working, I am okay with removing it, but we need an assigned follow-up issue to be sure that this get re-implemented

simonrey1:
I reopened issue https://github.com/podman-desktop/podman-desktop/issues/1011 3 weeks ago, and Florent was assigned but we could see if someone else can work on it in this sprint?

cc @jeffmaury

simonrey1:
The old issue has been closed in favor of a new issue to be assigned for the next sprint https://github.com/podman-desktop/podman-desktop/issues/13699, can we merge this PR @axel7083 @gastoner?

axel7083 (approved):

PR #13461, opened 2025-08-01
fix: verify the ports requested for kind cluster are free

### LINKED ISSUE #12405: When creating a kind cluster with default values, http port for kind control plane should be detected dynamically
### Bug description

On RHEL for example, I am trying to create a kind cluster with default parameters: http port: 9090, https: 9443. But by default, rhel has 9090 port number dedicated to a RHEL web console. We should detect that prior to setting a default port number for Kind cluster create page.

### Operating system

Linux, Rhel

### Installation Method

None

### Version

1.18.0

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

dgolovin:
### What does this PR do?

PR checks if ports requested for kind cluster is free and show auditor error message.

### Screenshot / video of UI

https://github.com/user-attachments/assets/b65bf794-cb98-4ce8-8e46-317f47c5c5f2

Fix #12405.

### How to test this PR?

Create two kind clusters. Creating second cluster should trigger request to use different free ports.

- [x] Tests are covering the bug fix or the new feature (TBD)

### DIFF
--- a/extensions/kind/src/create-cluster.ts
+++ b/extensions/kind/src/create-cluster.ts
@@ -96,6 +96,25 @@ export async function connectionAuditor(provider: string, items: AuditRequestIte
     });
   }
 
+  const httpPort = items['kind.cluster.creation.http.port'];
+  const freeHttpPort = await extensionApi.net.getFreePort(httpPort);
+  if (httpPort !== freeHttpPort) {
+    records.push({
+      type: 'error',
+      record: `HTTP Port ${httpPort} is not available. Please use next available port ${freeHttpPort}.`,
+    });
+  }
+
+  const httpsPort = items['kind.cluster.creation.https.port'];
+  const freeHttpsPort = await extensionApi.net.getFreePort(httpsPort);
+
+  if (httpsPort !== freeHttpsPort) {
+    records.push({
+      type: 'error',
+      record: `HTTPS Port ${httpsPort} is not available. Please use next available port ${freeHttpsPort}.`,
+    });
+  }
+
   const providerSocket = extensionApi.provider
     .getContainerConnections()
     .find(connection => connection.connection.type === provider);
@@ -111,7 +130,6 @@ export async function connectionAuditor(provider: string, items: AuditRequestIte
       record: 'It is recommend to install Kind on a virtual machine with at least 6GB of memory.',
     });
   }
-
   return auditResult;
 }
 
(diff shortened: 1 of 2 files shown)


extensions/kind/src/create-cluster.ts
```diff
@@ -29,6 +29,7 @@ import { parseAllDocuments, stringify } from 'yaml';
 
 import ingressManifests from '/@/resources/contour.yaml?raw';
 
+import { getFreePort } from '../../../packages/main/src/plugin/util/port';
```
dgolovin: @benoitf is there a better way to reuse port.ts? Should I use available npm package instead?

simonrey1: Maybe you could add it using a separate PR to `packages/main/src/plugin/extension/extension-loader.ts`, and the use the `extensionApi` to call it?

jeffmaury (commented):
Can't we use the diagnostic feature to report port that are already in use ?

dgolovin:
> Can't we use the diagnostic feature to report port that are already in use ?

@jeffmaury do you mean do it throug autditItems callback, then let user to update it manulay?

jeffmaury:
Yes

extensions/kind/src/create-cluster.ts
dgolovin: I am not sure it should be part of extensionApi. I'd rather use relative path or external package. I'll stick to relative path for now.

dgolovin:
@jeffmaury @eqqe  I updated screen-cast. Please review.

jiridostal:
1. Should we disable the Create button when the user runs into the occupied port and the error is shown on top?
2. The final error message has a different design, but this should be another PR as it's not related to this issue.

dgolovin:
> 1. Should we disable the Create button when the user runs into the occupied port and the error is shown on top?

@jiridostal The audit error does block creating kind instance, it just does not visible in screen recording

image

dgolovin:
> 2\. The final error message has a different design, but this should be another PR as it's not related to this issue.

yes, audit and instance creation errors look differently. I'll open issue for that.

feloy (changes requested):
I don't think that using freePort from main works. Having it shared with extension-api looks a good idea to me, as this code is already duplicated in other extensions

gastoner (approved):
Works fine on Linux

cdrage (approved):
I ran into zero issues when testing this out, implementation wise LGTM.

When using `getFreePort` it worked correctly.

I believe `getFreePort` was fixed recently by @bmahabirbu here: https://github.com/podman-desktop/podman-desktop/pull/13481 regarding the free ports for localhost + 0.0.0.0

LGTM my side.

dgolovin:
@feloy I added new PR to introduce new net module with port detection related functions see #13625.

extensions/kind/src/create-cluster.ts
dgolovin: @eqqe added new net module with getFreePort function (see #13625).

feloy (approved):

PR #13474, opened 2025-08-01
chore: switched to experimental manager function

gastoner:
### What does this PR do?
Switches to isExperientalFeature instead of getConfigurationValue

### How to test this PR?


simonrey1 (approved):
Tested on this branch, the ON / OFF `Replace the current task manager widget by the new one` switch still seems to work for me

PR #13481, opened 2025-08-01
fix: addressed multi port forwarding having the same port

### LINKED ISSUE #13279: Kubernetes: Port forwarding no longer works, always picks local port 5000
### Bug description

When you open up multiple port forwarding, it's always picking local port 5000 no matter what.

Image

### Operating system

mac m1

### Installation Method

None

### Version

1.19.2

### Steps to reproduce

1. Port forward multiple pods / services
2. Check Port Forwarding page
3. See they are all 5000 local port

### Relevant log output

```shell

```

### Additional context

_No response_

bmahabirbu:
The issue was that in port.ts it was checking 0.0.0.0 instead of localhost. This mismatch should have led to an error out but the way the promise was handled the error wasnt properly shown to the ui. Nevertheless fixing this issue solves the problem!

### What does this PR do?

see above

fixes https://github.com/podman-desktop/podman-desktop/issues/13279

### How to test this PR?

You need a kind instance running

[pod.txt](https://github.com/user-attachments/files/21555497/pod.txt)

download docker.io/library/nginx
create a container from that image
create a pod using that image
generate the kube from that pod
Add additional ports like in the txt example above but it needs to be a .yaml file (you can use the example above just change the name I think)
click on generate kube play and load up the .yaml file that you created from the generate kube output
Go to pods under the kubernetes extension click on the name of the pod go to summary tab scroll down and click on the port fowarding do it multiple times for each port
it should work!

### DIFF
--- a/packages/main/src/plugin/util/port.ts
+++ b/packages/main/src/plugin/util/port.ts
@@ -86,6 +86,7 @@ export async function isFreePort(port: number): Promise<boolean> {
 
   const intfs = getIPv4Interfaces();
 
+  await isFreeAddressPort('localhost', port);
   await isFreeAddressPort('0.0.0.0', port);
   await Promise.all(intfs.map(intf => isFreeAddressPort(intf, port)));
 


SoniaSandler (approved):
LGTM, tested and it works as expected.

PR #13499, opened 2025-08-04
fix: do not re-write registiries to auth.json if the values have not changed

### LINKED ISSUE #13061: Cannot build an image on Mac OS with Error: failed to parse "X-Registry-Config" header when there is invalid registry
### Bug description

When building an image on podman machine (affects both libkrun and appleHV) on Mac OS, I got an error:
```
error in buildImage Error: (HTTP code 400) unexpected - failed to parse "X-Registry-Config" header
```

### Operating system

Mac OS 14

### Installation Method

None

### Version

next (development version)

### Steps to reproduce (Updated)

1. Podman 5.5.2, Podman Desktop nightly
2. Start up podman machine (any virt. provider)
3. Add invalid registry (add valid registry first, then open `~/.config/containers/auth.json` and adjust the auth token to invalidate it)
4. Restart Podman Desktop
5. Try to build an image
6. Image is built
Actual result: error building an image

### Relevant log output

```shell
main ↪️ error in buildImage Error: (HTTP code 400) unexpected - failed to parse "X-Registry-Config" header for /build?dockerfile=test-containerfile&t=build-image-test&platform=linux%2Farm64&remote=&q=&rm=&forcerm=&memory=&memswap=&cpushares=&cpusetcpus=&cpuperiod=&cpuquota=&shmsize=&squash=&networkmode=&target=&outputs=&nocache=&buildargs=%7B%7D: failed to decode "X-Registry-Config": illegal base64 data at input byte 135
    at /Users/podmanqe/pd-e2e/podman-desktop/packages/main/dist/index.cjs:20190:19
    at handler (/Users/podmanqe/pd-e2e/podman-desktop/packages/main/dist/index.cjs:20228:13)
    at IncomingMessage. (/Users/podmanqe/pd-e2e/podman-desktop/packages/main/dist/index.cjs:20218:11)
    at IncomingMessage.emit (node:events:530:35)
    at endReadableNT (node:internal/streams/readable:1698:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
```

### Additional context

I can build the same image from cli: `podman build -t myimage -f path/to/my/containerfile .`

SoniaSandler:

### What does this PR do?

More info about the root of this bug and possible solutions can be found here https://github.com/podman-desktop/podman-desktop/issues/13061#issuecomment-3145694137

This fix doesn't solve the error, which is expected in this case of an invalid base64 auth value, but it makes sure to have a similar behavior between Podman and Podman Desktop.

With this fix, Podman Desktop doesn't re-write auth value that haven't changed to `auth.json`, so now the invalid base64 value stays, and the error also keeps showing up with Podman and Podman Desktop. 
Without this fix, Podman Desktop would re-encode the values to base64 and re-write them to `auth.json`, so values that were invalid base64 before, are now valid when Podman uses them.

There are other possible solutions, but some of them might lead to different Podman and Podman Desktop behavior in this case.

Fixes https://github.com/podman-desktop/podman-desktop/issues/13061

### How to test this PR?

1. Podman 5.5.2, Podman Desktop nightly
2. Start up podman machine (any virt. provider)
3. Add invalid registry (add valid registry first, then open ~/.config/containers/auth.json and adjust the auth token to invalidate it by removing one char)
4. Restart Podman Desktop
5. Try to build an image
Expected: an error pops up both in Podman Desktop and Podman

### DIFF
--- a/extensions/podman/packages/extension/src/utils/registry-setup.spec.ts
+++ b/extensions/podman/packages/extension/src/utils/registry-setup.spec.ts
@@ -187,15 +187,15 @@ test('should work with JSON auth file and alias', async () => {
 
 test('should send a warning in console if registry auth value is invalid', async () => {
   // mock the existSync
-  const existSyncSpy = vi.spyOn(fs, 'existsSync');
-  existSyncSpy.mockReturnValue(true);
+  const existSyncMock = vi.mocked(fs.existsSync);
+  existSyncMock.mockReturnValue(true);
 
   // mock the readFile
-  const readFileSpy = vi.spyOn(fs, 'readFile') as unknown as MockedFunction<ReadFileType>;
+  const readFileMock = vi.mocked(fs.readFile) as unknown as MockedFunction<ReadFileType>;
   const auth = Buffer.from('user:password').toString('base64');
   const invalidAuth = Buffer.from('userpassword').toString('base64');
 
-  readFileSpy.mockImplementation(
+  readFileMock.mockImplementation(
     (_path: string, _encoding: string, callback: (err: Error | undefined, data: string | Buffer) => void) => {
       // mock the error
 
@@ -219,7 +219,7 @@ test('should send a warning in console if registry auth value is invalid', async
   await registrySetup.updateRegistries();
 
   // expect read with the correct file
-  expect(readFileSpy).toHaveBeenCalledWith(authJsonLocation, 'utf-8', expect.anything());
+  expect(readFileMock).toHaveBeenCalledWith(authJsonLocation, 'utf-8', expect.anything());
   expect(consoleWarnMock).toHaveBeenCalledWith('Invalid auth value for myinvalidregistry.io');
 });
 
@@ -254,13 +254,13 @@ test.each([
   'do not write existing registries that did not change values to auth.json',
   async ({ fileAuth, registeredRegistry, timesCalled }) => {
     // mock the existSync
-    const existSyncSpy = vi.spyOn(fs, 'existsSync');
-    existSyncSpy.mockReturnValue(true);
+    const existSyncMock = vi.mocked(fs.existsSync);
+    existSyncMock.mockReturnValue(true);
 
     // mock the readFile
-    const readFileSpy = vi.spyOn(fs, 'readFile') as unknown as MockedFunction<ReadFileType>;
+    const readFileMock = vi.mocked(fs.readFile) as unknown as MockedFunction<ReadFileType>;
 
-    readFileSpy.mockImplementation(
+    readFileMock.mockImplementation(
       (_path: string, _encoding: string, callback: (err: Error | undefined, data: string | Buffer) => void) => {
         // mock the error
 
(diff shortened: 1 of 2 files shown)


bmahabirbu:
Tested and verified the functionality is what as its described e.i "doesn't re-write auth value that haven't changed to auth.json".

I verified that the error happens when editing the auth key including the podman desktop console output error `[main] [podman] Invalid auth value for docker.io`

I also verified a successful build when the auth is correct

And finally verified if the auth value isn't re-encoded before this patch

Codewise looks good!

extensions/podman/packages/extension/src/utils/registry-setup.ts
```diff
@@ -109,16 +113,31 @@ export class RegistrySetup {
...
 
-        await this.writeAuthFile(JSON.stringify(authFile, undefined, 8));
+        // if the registry already exists in the file, check if it has the same value as the registered registry
+        if (authFile.auths[registry.serverUrl]) {
+          const decoded = Buffer.from(authFile.auths[registry.serverUrl].auth, 'base64').toString();
+
+          // split the decoded string into username and password separated by :
+          const [username, secret] = decoded.split(':');
+
+          // do not encode the registry again if the values are not different from the ones in the file
```
bmahabirbu: Small non-blocker: I think to be more clear this comment can be written like "Only encode if values has changed from what's stored in the auth file"

extensions/podman/packages/extension/src/utils/registry-setup.spec.ts
```diff
@@ -162,3 +184,121 @@ test('should work with JSON auth file and alias', async () => {
   // expect read with the correct file
   expect(readFileSpy).toHaveBeenCalledWith(authJsonLocation, 'utf-8', expect.anything());
 });
+
+test('should send a warning in console if registry auth value is invalid', async () => {
+  // mock the existSync
+  const existSyncSpy = vi.spyOn(fs, 'existsSync');
```
axel7083: suggestion: as we have the node:fs package fully mock 

https://github.com/podman-desktop/podman-desktop/blob/541fbf15d20fd2008ac1c05466f3d9e968059fe0/extensions/podman/packages/extension/src/utils/registry-setup.spec.ts#L46

we may use `vi.mocked` instead for all vi.spyOn

extensions/podman/packages/extension/src/utils/registry-setup.ts
```diff
@@ -109,16 +113,31 @@ export class RegistrySetup {
...
+          encode = !(username === registry.username && secret === registry.secret);
+        }
+
+        if (encode) {
+          authFile.auths[registry.serverUrl] = {
+            auth: Buffer.from(`${registry.username}:${registry.secret}`).toString('base64'),
```
axel7083, line 135: I am wondering if it would not be easier to do the encoding, and compare the encoded strings, instead of decoding and comparing individual component?

SoniaSandler: @axel7083 most of (if not all) the invalid base64 values are decoded to some username (abc123) and undefined secret since they don't include `:` when decoded. So if we encode those values, we will get different encoded string since it will be `encode('${abc123}:${undefined}')`.
~~Another option could be to check if secret is undefined and the encoded username is the same as the original encoded auth value~~
The above option also don't work since based on the `toString` documentation `If encoding is 'utf8' and a byte sequence in the input is not valid UTF-8, then each invalid byte is replaced with the replacement character U+FFFD`. So if we encode the back the invalid username, we can't be sure if its encoded value the same as the auth value even though no change had been made by the user in Podman Desktop

bmahabirbu (approved):

axel7083 (approved):
LGTM codewise

PR #13506, opened 2025-08-05
chore: add custom icon and iconColor property to notifications 

### LINKED ISSUE #13371: Allow passing custom icon to notifications

In cases such as the one in the screenshot, custom icon is passed (Docker in this case). Currently there are just three icons to choose from - info, warn, error.

Image

Have a way to pass an Icon to notification.

_No response_

### Additional context

_No response_

jiridostal:
### What does this PR do?

SSIA.

Some mockups on notifications use custom icons, sometimes with a different color from the default set of "info" "warn" "error". 

Closes https://github.com/podman-desktop/podman-desktop/issues/13371

### DIFF
--- a/packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
+++ b/packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { faCircleExclamation, faCircleInfo, faExclamationTriangle, faXmark } from '@fortawesome/free-solid-svg-icons';
+import { faXmark } from '@fortawesome/free-solid-svg-icons';
 import { Icon } from '@podman-desktop/ui-svelte/icons';
 
 import type { NotificationCard } from '/@api/notification';
@@ -12,17 +12,17 @@ const notificationStyleMap = {
   info: {
     borderColor: 'border-[var(--pd-state-info)]',
     iconColor: 'text-[var(--pd-state-info)]',
-    icon: faCircleInfo,
+    icon: 'fas fa-info-circle',
   },
   warn: {
     borderColor: 'border-[var(--pd-state-warning)]',
     iconColor: 'text-[var(--pd-state-warning)]',
-    icon: faExclamationTriangle,
+    icon: 'fas fa-exclamation-triangle',
   },
   error: {
     borderColor: 'border-[var(--pd-state-error)]',
     iconColor: 'text-[var(--pd-state-error)]',
-    icon: faCircleExclamation,
+    icon: 'fas fa-circle-exclamation',
   },
 };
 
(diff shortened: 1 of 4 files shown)


bmahabirbu:
codewise looks good! Question does this PR aim to standardize the notification icon and color?

jiridostal:
@bmahabirbu Kinda. It's related to https://github.com/podman-desktop/podman-desktop/issues/13106. First part (the linked issue) was to introduce updated theme for info, warn and error, but if you look the mocks it uses custom icon in some cases (ie. the Docker icon). This PR adds possibility to do it. Where to use it exactly will need to be discussed with UX, but then it should be easy.

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
```diff
@@ -35,7 +35,11 @@ const notificationStyle = notificationStyleMap[notification.type];
   aria-label="id: {notification.id}">
   <div class="flex flex-row space-x-3">
     <div class="flex">
-        <Icon size="1.5x" icon={notificationStyle.icon} class={notificationStyle.iconColor} title={'Notification icon - ' + notification.type}/>
+      <Icon 
+        icon={notification.icon ?? notificationStyle.icon} 
+        class={(notification.iconColor ?? notificationStyle.iconColor) + ' text-xl'}
```
gastoner: Can we also change this to `${notification.iconColor ?? notificationStyle.iconColor} text-xl`` ?

gastoner (commented):
Thinking if wouldn't be beneficial to add also `iconSize? :string` that would default to `text-xl`? But not sure if there would be some cases where it would be good to have.

I would also like to have the support for all Icon types => which might not be ideal do through our API 
but for this, I think there are needed some additional opinions

packages/renderer/src/lib/dashboard/NotificationCardItem.svelte
jiridostal: Updated.

jiridostal:
@gastoner I had the same thought about the iconSize, but I came to a conclusion that it will break the visual consistency when used, so I decided not to add it. 

As for the object Icons - I tried to add support for it to the API but I ended up with a messy code and I didn't like it :-D Maybe I'll give it another try today.

jiridostal:
@gastoner I have added more types to the internal API, it works fine with other code changes. One thing I was not able to do is to allow usage with extension-api - I would either need to recreate the IconDefinition and Componet (from Svelte) from scratch or use type "any", which may need to be discussed.

gastoner (approved):
LGTM, I was looking into this too and we don't have an fa package in our extensions at all so at least `IconDefinition` would be redundant in the API

bmahabirbu (approved):

PR #13532, opened 2025-08-08
chore: introduce github labeler

jiridostal:
### What does this PR do?
This adds a workflow for PRs which matches paths/extensions for files and based on that assign corresponding label. Currently this tries to detect any visual changes and website changes for UX awareness, but could be updated with more logic to provide more robust labeling.

Don't merge before correct tokens are set-up.

Let's discuss ideas or improvements as part of this PR.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,32 @@
+#
+# Copyright (C) 2025 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Visual changes
+"area/ui":
+  - changed-files:
+    - any-glob-to-any-file:
+      - "packages/renderer/**/*"
+      - "packages/ui/**/*"
+      - "storybook/**/*"
+      - "**/*.svelte"
+      - "**/*.css"
+
+# Website changes
+"area/website":
+  - changed-files:
+    - any-glob-to-any-file:
+      - "website/**/*"
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -1,0 +1,36 @@
+#
+# Copyright (C) 2025 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,15 @@
+# Visual changes
```
benoitf, line 18: missing: copyright headers

.github/labeler.yml
```diff
@@ -0,0 +1,15 @@
...
+      - "packages/ui/**/*"
+      - "storybook/**/*"
+      - "**/*.svelte"
+      - "**/*.css"
+
+# Website changes
+"area/website":
+  - changed-files:
+    - any-glob-to-any-file:
+      - "website/**/*"
```
benoitf: missing: EOL

.github/workflows/labeler.yaml
```diff
@@ -0,0 +1,37 @@
+#
+# Copyright (C) 2022-2024 Red Hat, Inc.
```
benoitf:
```suggestion
# Copyright (C) 2025 Red Hat, Inc.
```

.github/workflows/labeler.yaml
```diff
@@ -0,0 +1,37 @@
...
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  labeler:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Labeler
+        uses: actions/labeler@v5.0.0
```
benoitf: issue: need to use sha, not a rolling tag

.github/workflows/labeler.yaml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          repo-token: "${{ secrets.LABELER_TOKEN }}"
```
benoitf:
```suggestion
          repo-token: "${{ secrets.PODMAN_DESKTOP_BOT_TOKEN }}"
```

.github/workflows/labeler.yaml
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: "PR Labeler"
+
+on:
+  - pull_request_target
```
benoitf: I suspect some events are missing

benoitf (changes requested):

.github/workflows/labeler.yaml
benoitf: related to https://github.com/podman-desktop/podman-desktop/blob/main/.github/workflows/pr-check.yaml#L23

default one are ok for this usecase ?

.github/workflows/labeler.yaml
```diff
@@ -0,0 +1,37 @@
+#
```
benoitf: suggestion: rename it pr-labeler (pr- prefix) so we know it's done on PR

.github/labeler.yml
```diff
@@ -0,0 +1,32 @@
+#
+# Copyright (C) 2022-2025 Red Hat, Inc.
```
benoitf:
```suggestion
# Copyright (C) 2025 Red Hat, Inc.
```

.github/workflows/pr-labeler.yaml
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: "PR Labeler"
+
+on:
+  merge_group:
+  pull_request:
```
benoitf: it needs only be with pull_request_target as we want to be able to read the token

benoitf (approved):
let's try

PR #13548, opened 2025-08-11
Fix onboarding ui e2e test getting blocked by extension loading

### LINKED ISSUE #13047: Onboarding initial tiles (Podman, Compose, Kubectl) are not rendered until all built-in extensions are activated
### Bug description

When we encountered an issue with e2e tests described in https://github.com/podman-desktop/e2e/issues/372, I noticed that the problem of the e2e tests was that is was waiting for Start Onboarding button. But it did no appear in 20 seconds. I noticed that kind extension activation timed out on windows on GHA CI. This was the reason the test failed, but the problem is that we are waiting for such event for extension we should not have. 

### Operating system

Windows Server 2022

### Installation Method

None

### Version

1.19.2

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

bmahabirbu:
### What does this PR do?

The e2e test for welcome-page.ts had a timeout of 20 seconds which conflicts with each extension timeout being 20 seconds. If an extension used the full timeout it would crash the e2e test. Added a new timeout of 60s to have up to 3 extensions use the max time before the error

Fixes https://github.com/podman-desktop/podman-desktop/issues/13047

### How to test this PR?

in extensions/kind/src/extension.ts 
add a timout like this in line 588

  // Add a 20s delay to simulate a slow activation
  await new Promise(resolve => setTimeout(resolve, 20_000));
  console.log('Kind extension simulate slow activation');

then run 

pnpm run test:e2e:build
pnpm exec playwright test tests/playwright/src/specs/welcome-page-smoke.spec.ts --headed

### DIFF
--- a/tests/playwright/src/model/pages/welcome-page.ts
+++ b/tests/playwright/src/model/pages/welcome-page.ts
@@ -53,7 +53,9 @@ export class WelcomePage extends BasePage {
       const isUpdateTest = test.info().tags.includes('@update-install');
 
       if (!isUpdateTest) {
-        await playExpect(this.startOnboarding).toBeEnabled({ timeout: 20_000 });
+        // Extensions load sequentially (faster speed but can block ui)
+        // Each extension timeout is 20s use 60s to provide a decent bufffer
+        await playExpect(this.startOnboarding).toBeEnabled({ timeout: 60_000 });
       }
 
       if (await this.telemetryConsent.isChecked()) {


bmahabirbu:
Need to add tests as well

axel7083 (changes requested):
I also share the issue that not only on onboarding the containers takes forever to appear when starting Podman Desktop; however

Last time we change the way extensions were loaded, we run tests over different computer, measuring startup time etc.

On some PC, there is some clear difference of startup time on low end PC, where running them in sequence is faster than loading in parallel.

I would be interested to see comparison before merge. Do we have proper telemetry on starting all extensions ? Do we telemetry to measure the impact of such change ?  

---

Ref previous change https://github.com/podman-desktop/podman-desktop/pull/7332

Maybe we can check if the previous change has an impact (using telemetry)

---

I have some intuition that the slowing starting issue of extensions is realated

- too many blocking IO
- too many things run in the extension's activate function
- too many things in the main thread

bmahabirbu:
Thanks @axel7083 for the review! I had a feeling that the extension loading was sequential. I agree with your perspective here and now think we should keep loading sync. 

This bug is to do with onboarding e2e waits for the onboard button for 20 seconds but each extensions has a 20 second timeout. So if one extension takes too long it will crash the e2e test. An easier approach would be to just lower the timeout for activating each extension to be within the larger wait time for the onboard timeout.

Do you know where I can find the testing data for the extension loading? It would be interesting to see the slowest load speed so I could adjust the timeout better. 

Thanks a bunch!

bmahabirbu:
Ive updated the pr with a new approach ive kept the sequential loading but prioritized the loading of the extensions needed for onboarding first! that way it doesnt block the ui

bmahabirbu:
@cdrage mind taking a look at this one too please!

axel7083:
> Thanks @axel7083 for the review! I had a feeling that the extension loading was sequential. I agree with your perspective here and now think we should keep loading sync.
> 
> This bug is to do with onboarding e2e waits for the onboard button for 20 seconds but each extensions has a 20 second timeout. So if one extension takes too long it will crash the e2e test. An easier approach would be to just lower the timeout for activating each extension to be within the larger wait time for the onboard timeout.

Yes, I understand the issue, but this is kinda tricky, loading them all at the same time feel patchy, why are they taking 20 seconds ? is there an issue during the test ?

The timeout has been set to 20s for low end PC, it was 10s before, but the podman extension is very complex, and has a lot of IO making it very slow to start on some hardware.

> Do you know where I can find the testing data for the extension loading? It would be interesting to see the slowest load speed so I could adjust the timeout better.

You may explore [Extension activation dashboard](https://app.amplitude.com/analytics/redhat/dashboard/cdf25xda?source=copy+url), chart related to activate telemetry, we may need to add other telemetry event or graphs

bmahabirbu:
@axel7083 I have updated this pr with a new approach that doesn't require async please take a look at my latest commit.

The test was failing here  https://github.com/orgs/podman-desktop/projects/4/views/3?pane=issue&itemId=117365770&issue=podman-desktop%7Ce2e%7C372 because the test was waiting for kind extension to activate. (I think the issue was on windows gha ci and maybe kind wasnt activated) 

but Onboard shouldn't wait for an extension that it doesn't depend on.

Before, onboard would wait for the complete extension activate cycle to complete. Now it only waits for the core extension podman compose and kubectl. Since the built in extensions dont have any dependencies its safe to move podman compose and kubectl first in the topological ordering. 

So now those 3 activate first, onboard checks if those 3 are activated and then moves on. If an extension does take a long time to activate it now wont block onboard but depending on where it is in the order will block the ones below, which is just the nature of loading sequentially

packages/main/src/plugin/extension/extension-loader.ts
```diff
@@ -590,8 +588,11 @@ export class ExtensionLoader implements AsyncDisposable {
     const sorted: AnalyzedExtension[] = [];
     const explored = new Map<string, boolean>();
 
-    // visit all unvisited nodes
-    analyzedExtensions.forEach(node => {
+    const criticalExtensionIds = ['podman-desktop.podman', 'podman-desktop.compose', 'podman-desktop.kubectl-cli'];
```
cdrage: const file? Also may open up a can of worms of what we define as critical vs non critical.

cdrage (changes requested):
Missing tests unfortunatley.

I think IMHO this is a slippery slope to introduce more bugs.

For example, if we were to add more onboarding (like adding an extension, or modifying a current one), we have to remember what's critical vs non-critical.

Same for the "kubectl-cli" extension. I believe we're moving that "out" of PD core too.

Maybe a different approach would be a timeout instead, similar to your comment before @bmahabirbu and @axel7083 "An easier approach would be to just lower the timeout for activating each extension to be within the larger wait time for the onboard timeout."

benoitf (changes requested):
there should not have any dependency to extension names from the app

bmahabirbu:
that's some good points @cdrage. I'll circle back and work on adjusting the timeouts instead for this pr thanks!

packages/main/src/plugin/extension/extension-loader.ts
bmahabirbu: Yeah that makes perfect sense from what you said before

bmahabirbu:
Ive updated the pr to increase the welcome-page.ts timeout which fixes the immediate issue

I'd like to revisit this issue, I think there is some optimization to be done on extension activation ordering. E.x We can check the extension package.json to look for onboarding to load that first (avoids hardcoding names) but as @cdrage said it can introduce bugs for a small time save

Thanks all!

odockal (approved):

cdrage (changes requested):
IMHO.

We should create separate, new PR with this fix because there's been too much discussion in here (and there was previously a different way of tackling the problem).

The commit message also has no description, and is out of date vs the  GitHub description.

Screenshot 2025-08-13 at 11 37 35 AM

 Sorry to be a pain!

bmahabirbu:
Not a problem at all! Thanks, @cdrage for the advice! Will do!

PR #13552, opened 2025-08-11
feat: added support for masking passwords

### LINKED ISSUE #13476: prompt box behaviour not handling secrets as expected
### Bug description

As per [discussion](https://github.com/podman-desktop/podman-desktop/discussions/13455l)

Seems I've found a bug in the way secrets are handled when a user is prompted for s secret, the promptbox is provided correctly but the secret is not obfuscated whilst being entered.

Expected behaviour would be 

```Typescript
      const password = await podmanDesktopAPI.window.showInputBox({
        prompt: 'Enter the password for the user',
        placeHolder: 'Password',
        password: true, // Mask the input for security
      });
```

the **password:true** setting should enable obfuscation of any text while it is being entered

### Operating system

macOS Sequoia (ARM)

### Installation Method

Other

### Version

1.20.0

### Steps to reproduce

**clone** Podman desktop repo into Documents folder on device. create new basic extension within the extensions folder a clone of <git clone https://github.com/containers/podman-desktop-extension-minimal-template/> will do.

make extension.ts look like this

```Typescript
import type { ExtensionContext } from '@podman-desktop/api';
import * as extensionApi from '@podman-desktop/api';

/**
 * Below is the "typical" extension.ts file that is used to activate and deactivate the extension.
 * this file as well as package.json are the two main files that are required to develop a Podman Desktop extension.
 */

// Initialize the activation of the extension.
export async function activate(extensionContext: ExtensionContext): Promise<void> {
  console.log('starting hello world extension');

  // Create a dialog that says "Hello World" with the extensionApi on load
  extensionApi.window.showInformationMessage('Hello World from the Podman Desktop Extension!');

  // Show an input box to get the user's password
  const password = await extensionApi.window.showInputBox({
    title: 'Enter your password',
    prompt: 'Please enter your password',
    password: true,
    ignoreFocusOut: true,
  });

  if (password !== undefined) {
    console.log('Password entered by user (hidden for security reasons)');
    // You can handle the password here as needed
  } else {
    console.log('User cancelled the password input.');
  }
}

// Deactivate the extension
export async function deactivate(): Promise<void> {
  console.log('stopping hello world extension');
}
```

Add a promptbox and test it in podmandesktop, when entering password it should be obfuscated. (Again as per my understanding if Password : True is set it infers that the input should be treated as a secret and hence should not be captured as Palintext.

### Relevant log output

```shell

```

### Additional context

Image

gastoner:
### What does this PR do?
Adds support to QuickPickInput to be able to mask password when passord is set to true

### Screenshot / video of UI
Old:

image

Current:

image

Closes #13476 

### How to test this PR?
Unit tests 

Steps in the issue

### DIFF
--- a/packages/renderer/src/lib/dialogs/QuickPickInput.spec.ts
+++ b/packages/renderer/src/lib/dialogs/QuickPickInput.spec.ts
@@ -208,6 +208,33 @@ describe('QuickPickInput', () => {
     expect(area).toBeInTheDocument();
   });
 
+  test.each([
+    { password: true, type: 'password' },
+    { password: false, type: 'text' },
+  ])('Expect that input is displayed correctly with type %s', async ({ password, type }) => {
+    const idRequest = 123;
+
+    const inputBoxOptions: InputBoxOptions = {
+      password: password,
+      validate: false,
+      placeHolder: 'Some placeholder',
+      prompt: 'Enter a value',
+      id: idRequest,
+    };
+
+    receiveFunctionMock.mockImplementation((message: string, callback: (options: InputBoxOptions) => void) => {
+      if (message === 'showInputBox:add') {
+        callback(inputBoxOptions);
+      }
+    });
+
+    render(QuickPickInput, {});
+
+    const area = await screen.findByPlaceholderText('Some placeholder');
+    expect(area).toBeInTheDocument();
+    expect(area).toHaveAttribute('type', type);
+  });
+
   test('Expect that filtering works', async () => {
     const idRequest = 123;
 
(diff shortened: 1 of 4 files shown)


packages/renderer/src/lib/dialogs/QuickPickInput.spec.ts
```diff
@@ -208,6 +208,33 @@ describe('QuickPickInput', () => {
...
+      if (message === 'showInputBox:add') {
+        callback(inputBoxOptions);
+      }
+    });
+
+    render(QuickPickInput, {});
+
+    const area = await screen.findByPlaceholderText('Some placeholder');
+    expect(area).toBeInTheDocument();
+    expect(area).toHaveAttribute('type', type);
```
axel7083:
```suggestion
    const { getByPlaceholderText } = render(QuickPickInput, {});

    const area = getByPlaceholderText('Some placeholder');
    expect(area).toBeInstanceOf(HTMLInputElement);
    expect(area).toHaveAttribute('type', type);
```

suggestion(non-blocking): personal preference
- to use result from render
- use `get` instead of `find`
- verify instance

axel7083 (approved):
LGTM, worked great 

Image

vancura (approved):
LGTM!

PR #13589, opened 2025-08-13
fix: increase onboarding timeout to allow extensions to load

### LINKED ISSUE #13047: Onboarding initial tiles (Podman, Compose, Kubectl) are not rendered until all built-in extensions are activated

bmahabirbu:
The e2e tests were failing because the tests only waited 20s for the start onboard button to appear. Each extension loads sequentially and can block ui for up to 20s (max activation time). Increased the timeout to 60s to provide slow or failing extensions a longer buffer before test fails.

### What does this PR do?

see above

Fixes https://github.com/podman-desktop/podman-desktop/issues/13047

### How to test this PR?

in extensions/kind/src/extension.ts 
add a timout like this in line 588

  // Add a 20s delay to simulate a slow activation
  await new Promise(resolve => setTimeout(resolve, 20_000));
  console.log('Kind extension simulate slow activation');

then run 

pnpm run test:e2e:build
pnpm exec playwright test tests/playwright/src/specs/welcome-page-smoke.spec.ts --headed

- [x ] Tests are covering the bug fix or the new feature

### DIFF
--- a/tests/playwright/src/model/pages/welcome-page.ts
+++ b/tests/playwright/src/model/pages/welcome-page.ts
@@ -53,7 +53,9 @@ export class WelcomePage extends BasePage {
       const isUpdateTest = test.info().tags.includes('@update-install');
 
       if (!isUpdateTest) {
-        await playExpect(this.startOnboarding).toBeEnabled({ timeout: 20_000 });
+        // Extensions load sequentially (faster speed but can block ui)
+        // Each extension timeout is 20s use 60s to provide a decent bufffer
+        await playExpect(this.startOnboarding).toBeEnabled({ timeout: 60_000 });
       }
 
       if (await this.telemetryConsent.isChecked()) {


bmahabirbu:
See https://github.com/podman-desktop/podman-desktop/pull/13548 for more info

cdrage (approved):
LGTM! Thank you for opening the new PR + updating commit message.

PR #13625, opened 2025-08-18
feat: add net module to expose free port detection function

dgolovin:
### What does this PR do?

Add new net module to Podman Desktop API with:
* getFreePort

Related to #12405.

### How to test this PR?

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -4823,23 +4823,10 @@ declare module '@podman-desktop/api' {
 
   export namespace net {
     /**
-     * Finds a free port starting from the specified port.
+     * Finds a free port starting from the specified port and returns it.
      * @param port The starting port number to search for a free port.
      */
     export function getFreePort(port: number): Promise<number>;
-
-    /**
-     * Finds a continuous range of free ports of the specified size.
-     * @param rangeSize The number of ports in the continuous range.
-     * @return A string in a format like "3000-3005".
-     */
-    export function getFreePortRange(rangeSize: number): Promise<string>;
-
-    /**
-     * Checks if the specified port is free.
-     * @param port The port number to check.
-     */
-    export function isFreePort(port: number): Promise<boolean>;
   }
 
   export namespace navigation {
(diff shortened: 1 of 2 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -4813,6 +4813,27 @@ declare module '@podman-desktop/api' {
...
+     * @param port The starting port number to search for a free port.
+     */
+    export function getFreePort(port: number): Promise<number>;
+
+    /**
+     * Finds a continuous range of free ports of the specified size.
+     * @param rangeSize The number of ports in the continuous range.
+     * @return A string in a format like "3000-3005".
+     */
+    export function getFreePortRange(rangeSize: number): Promise<string>;
```
axel7083: remark: I am not sure about exposing `getFreePortRange` where it is just a loop over `getFreePort`?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -4813,6 +4813,27 @@ declare module '@podman-desktop/api' {
...
+
+    /**
+     * Checks if the specified port is free.
+     * @param port The port number to check.
+     */
+    export function isFreePort(port: number): Promise<boolean>;
```
axel7083: I understand this is a copy past of the existing naming of the function imported from utils, and I am not an english native speaker but would `isPortFree` be more semantically correct ?

cc @SoniaSandler for native english advise

dgolovin: @axel7083 I also put isPortFree looks better for me, that what I used first and then switched to copied version.

packages/extension-api/src/extension-api.d.ts
dgolovin: @axel7083 I don't mind to remove it, because all I need in related issues is getFreePort. It can be added later if needed. The same about isFreePort below.

packages/extension-api/src/extension-api.d.ts
axel7083: Let's use `isPortFree` if we both agree!

packages/extension-api/src/extension-api.d.ts
feloy: I agree, let's export only what is needed

dgolovin:
@feloy @axel7083 updated the PR to expose the only required function.

axel7083 (approved):
LGTM

feloy (approved):

feloy (approved):

PR #13629, opened 2025-08-18
feat: adds tabs to global searchbar 

### LINKED ISSUE #13627: Add tab experience to Searchbar in titlebar

Add tabs for searchbar so we could pre-filter options 

Image

e.g. something like:

Image

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds tabs to searchbar so we can easily pre-filter options 

### Screenshot / video of UI

https://github.com/user-attachments/assets/2743582c-500b-4a99-9fc4-f045c6fdc37a

Closes #13627 

### How to test this PR?
Enable searchbar and try out different tabs (so far there is no content in them other than commands)


vancura:
Hey @gastoner, I have some feedback to share as well.

CleanShot 2025-08-18 at 13 34 26@2x

1. Please make the keyboard div horizontal padding larger (`px-2`, if I recall correctly from our call). Could you make sure they match the design in the footer and follow @MariaLeonova's comps?
2. For the close button, please make sure the hit area covers the whole input field height, and the width accommodates it. Ensure the button uses the correct border radius so that when the hover effect is visible, it blends in nicely. Follow Masha's design (the button is inside the box).
3. Consider the display of symbols like `>` here; if prefixing the prompt with them doesn't change the scope of found results, it doesn't make any sense to display them here.

vancura:
Also, please consider a good use of keyboard shortcuts, particularly `tab` / `shift+tab` for navigating between tab view.

MariaLeonova (commented):
Looks very good! One suggestion that I have -  do we need the special icons for the docs/resources tabs? I would suggest just the search icon + placeholder text (i.e. search documentation, blogs, tuts):
Screenshot 2025-08-19 at 10 28 09
Screenshot 2025-08-19 at 10 28 17

gastoner:
@MariaLeonova we probably dont, I was just feeling that it would be good to differentiate

MariaLeonova:
> @MariaLeonova we probably dont, I was just feeling that it would be good to differentiate

My thought was that for commands it's something that they type that shows up in the search bar (i.e. > symbol). And for other tabs it's not, there it's just searching. Helper text would be more beneficial in this case, especially in the case of searching across Podman Desktop. We can't really use the word resources there, so the bar could let them know what the searchable assets are.

vancura (changes requested):
Looks good.

- [x] When the panel is started with the `F1` OR `>` keyboard shortcuts, it should go straight to the _Commands_ view.
- [x] When the panel is started with the `⌘K` / `CTRL + K` keyboard shortcuts, it should go straight to the _Documentations_ view.
- [x] When the panel is started with the `⌘F` / `CTRL + F` keyboard shortcuts, it should go straight to the _Go to_ view.

gastoner:
@MariaLeonova @vancura also I assume I should add `>` next to `F1` in tabs on top right? 
It was not on the mockups but we have it in the footer 
Screenshot_20250820_165104

MariaLeonova:
> @MariaLeonova @vancura also I assume I should add `>` next to `F1` in tabs on top right? It was not on the mockups but we have it in the footer 

Yes, and it should be in the mocks, as well. I think it's here: https://github.com/podman-desktop/podman-desktop/pull/13629#pullrequestreview-3131286224

Also at some point we discussed not duplicating the shortcuts in the bottom bar, so there (if needed) we only have the mention of Algolia. So iy would look like this: 
Screenshot 2025-08-21 at 9 18 58


The latest tab designs can be seen [here](https://www.figma.com/design/IfJYttlkkly5wNiD28sMY9/Search-Flow?node-id=219-2952).

vancura (changes requested):
Hey, it's pretty good, but it's me, the nitpicker here :)

- [ ] There's a problem with the keyboard focus. See my video (and sorry about the microphone volume, I'll record better next time):

https://github.com/user-attachments/assets/4e81201f-fcb9-477b-b15c-1c8caab63214

- [ ] It's impossible to focus the _All_ tab with a general keyboard shortcut. Perhaps we should have one?

image

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -26,12 +26,35 @@ interface Props {
...
+    return faChevronRight;
+  } else {
+    return faMagnifyingGlass;
+  }
+});
+
+let isMac: boolean = $state(false);
+let modifier: string = $derived(isMac ? '⌘ ' : 'Ctrl ');
+let searchOptions: SearchOption[] = $derived([
+  { text: 'All', shortCut: [`${modifier}A`] },
```
vancura: This is actually `⌘⇧P` (`CTRL+SHIFT+P`) and not `A`. The label is wrong.

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -26,12 +26,35 @@ interface Props {
...
+  if (searchOptionsSelectedIndex === 1) {
+    // Commands
+    return faChevronRight;
+  } else {
+    return faMagnifyingGlass;
+  }
+});
+
+let isMac: boolean = $state(false);
+let modifier: string = $derived(isMac ? '⌘ ' : 'Ctrl ');
```
vancura: Please write:
* macOS: `⌘` (without space, following the macOS HIG)
* other: `Ctrl+` (with `+` and without space after the `+`); see the other uses of `Ctrl` strings in this PR.

vancura (changes requested):

vancura (approved):
Works perfectly, thanks for the hard work!

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -74,23 +113,38 @@ $effect(() => {
...
     e.preventDefault();
     return;
   } else if (e.key === ESCAPE_KEY) {
     // here we toggle the display
     hideCommandPallete();
     e.preventDefault();
     return;
+  } else if (e.ctrlKey || e.metaKey) {
+    if (e.shiftKey && e.key.toLowerCase() === 'p') {
+      searchOptionsSelectedIndex = 0;
```
benoitf, line 132: I don't see prevent default call for these new shortcuts ?

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -26,15 +29,40 @@ interface Props {
...
+    return faMagnifyingGlass;
+  }
+});
+
+let isMac: boolean = $state(false);
+let modifierC: string = $derived(isMac ? '⌘' : 'Ctrl+');
+let modifierS: string = $derived(isMac ? '⇧' : 'Shift+');
+let searchOptions: SearchOption[] = $derived([
+  { text: 'All', shortCut: [`${modifierC}${modifierS}P`] },
+  { text: 'Commands', shortCut: ['F1', '>'] },
```
benoitf: F1 is used at several places no in the code ( and others)
Probably need to be a constant to avoid any further change that will only change one path

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -237,6 +323,16 @@ async function onAction(): Promise<void> {
             </li>
           {/each}
         </ul>
+
+        {#if filteredCommandInfoItems.length === 0}
+          <div class='flex grow items-center flex-col gap-2 py-4'>
+            <Icon icon={isDarkTheme ? NotFoundDarkIcon : NotFoundLightIcon} />
```
benoitf: should it be a single icon object with a theme property ?
b/c I don't see in which case we would use just one icon file

gastoner: I was asking the same thing, but those are unique (the eyes are in both cases black as they should be)

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -43,11 +72,21 @@ let filteredCommandInfoItems: CommandInfo[] = $derived(
...
+  const platform = await window.getOsPlatform();
+
+  isMac = platform === 'darwin';
+});
 
 onMount(() => {
   contextsUnsubscribe = context.subscribe(value => {
     globalContext = value;
   });
+  themeUnsubscribe = isDark.subscribe(value => {
```
simonrey1: suggestion: it seems that we are in a Runes component, so it should be possible to use $isDark directly without subscribe in the onMount

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -43,11 +72,21 @@ let filteredCommandInfoItems: CommandInfo[] = $derived(
 );
 
 let contextsUnsubscribe: Unsubscriber;
+let themeUnsubscribe: Unsubscriber;
+
+onMount(async () => {
```
simonrey1, line 73: Duplicated onMount method with line 83

packages/renderer/src/lib/dialogs/CommandPalette.svelte
simonrey1:
```svelte

<script lang="ts">
import { faChevronRight, faMagnifyingGlass } from '@fortawesome/free-solid-svg-icons';
import { Button, Input } from '@podman-desktop/ui-svelte';
import { Icon } from '@podman-desktop/ui-svelte/icons';
import { onDestroy, onMount, tick } from 'svelte';
import type { Unsubscriber } from 'svelte/store';

import { isDark } from '/@/stores/appearance';
import { commandsInfos } from '/@/stores/commands';
import { context } from '/@/stores/context';
import type { CommandInfo } from '/@api/command-info';

import type { ContextUI } from '../context/context';
import ArrowDownIcon from '../images/ArrowDownIcon.svelte';
import ArrowUpIcon from '../images/ArrowUpIcon.svelte';
import EnterIcon from '../images/EnterIcon.svelte';
import NotFoundDarkIcon from '../images/NotFoundDarkIcon.svelte';
import NotFoundLightIcon from '../images/NotFoundLightIcon.svelte';
import { isPropertyValidInContext } from '../preferences/Util';

const ENTER_KEY = 'Enter';
const ESCAPE_KEY = 'Escape';
const ARROW_DOWN_KEY = 'ArrowDown';
const ARROW_UP_KEY = 'ArrowUp';
const TAB_KEY = 'Tab';

interface Props {
  display?: boolean;
  onclose?: () => void;
}

interface SearchOption {
  text: string;
  shortCut?: string[];
}

let { display = false, onclose }: Props = $props();

let outerDiv: HTMLDivElement | undefined = $state(undefined);
let inputElement: HTMLInputElement | undefined = $state(undefined);
let inputValue: string | undefined = $state('');
let scrollElements: HTMLLIElement[] = $state([]);
let searchIcon = $derived.by(() => {
  if (searchOptionsSelectedIndex === 1) {
    // Commands
    return faChevronRight;
  } else {
    return faMagnifyingGlass;
  }
});

let isMac: boolean = $state(false);
let modifierC: string = $derived(isMac ? '⌘' : 'Ctrl+');
let modifierS: string = $derived(isMac ? '⇧' : 'Shift+');
let searchOptions: SearchOption[] = $derived([
  { text: 'All', shortCut: [`${modifierC}${modifierS}P`] },
  { text: 'Commands', shortCut: ['F1', '>'] },
  { text: 'Documentation', shortCut: [`${modifierC}K`] },
  { text: 'Go to', shortCut: [`${modifierC}F`] },
]);
let searchOptionsSelectedIndex: number = $state(0);

let commandInfoItems: CommandInfo[] = $state([]);
let globalContext: ContextUI;

let filteredCommandInfoItems: CommandInfo[] = $derived(
  commandInfoItems
    .filter(property => isPropertyValidInContext(property.enablement, globalContext))
    .filter(item => (inputValue ? item.title?.toLowerCase().includes(inputValue.toLowerCase()) : true)),
);

let contextsUnsubscribe: Unsubscriber;

onMount(async () => {
  const platform = await window.getOsPlatform();

  isMac = platform === 'darwin';
});

onMount(() => {
  contextsUnsubscribe = context.subscribe(value => {
    globalContext = value;
  });
  // subscribe to the commands
  return commandsInfos.subscribe(infos => {
    commandInfoItems = infos;
  });
});

onDestroy(() => {
  contextsUnsubscribe?.();
});

// Focus the input when the command palette becomes visible
$effect(() => {
  if (display && inputElement) {
    tick()
      .then(() => {
        inputElement?.focus();
      })
      .catch((error: unknown) => {
        console.error('Unable to focus input box', error);
      });
  }
});

let selectedFilteredIndex = $state(0);
let selectedIndex = 0;

function displaySearchBar(): void {
  // clear the input value
  inputValue = '';
  selectedIndex = 0;
  // toggle the display
  display = true;
}

async function handleKeydown(e: KeyboardEvent): Promise<void> {
  // toggle display using F1 or ESC keys
  if (e.key === 'F1' || e.key === '>') {
    selectedFilteredIndex = 0;
    searchOptionsSelectedIndex = 1;
    displaySearchBar();
    e.preventDefault();
    return;
  } else if (e.key === ESCAPE_KEY) {
    // here we toggle the display
    hideCommandPallete();
    e.preventDefault();
    return;
  } else if (e.ctrlKey || e.metaKey) {
    if (e.shiftKey && e.key.toLowerCase() === 'p') {
      searchOptionsSelectedIndex = 0;
      displaySearchBar();
    } else if (e.key.toLowerCase() === 'k') {
      searchOptionsSelectedIndex = 2;
      displaySearchBar();
    } else if (e.key.toLowerCase() === 'f') {
      searchOptionsSelectedIndex = 3;
      displaySearchBar();
    }
  }

  // for other keys, only check if it's being displayed
  if (!display) {
    return;
  }

  // no items, abort
  if (filteredCommandInfoItems.length === 0) {
    return;
  }

  if (e.key === ARROW_DOWN_KEY) {
    // if down key is pressed, move the index
    selectedFilteredIndex++;
    if (selectedFilteredIndex >= filteredCommandInfoItems.length) {
      selectedFilteredIndex = 0;
    }

    selectedIndex = commandInfoItems.indexOf(filteredCommandInfoItems[selectedFilteredIndex]);

    scrollElements[selectedFilteredIndex].scrollIntoView({
      behavior: 'smooth',
      block: 'nearest',
      inline: 'start',
    });
    e.preventDefault();
  } else if (e.key === ARROW_UP_KEY) {
    // if up key is pressed, move the index
    selectedFilteredIndex--;
    if (selectedFilteredIndex < 0) {
      selectedFilteredIndex = filteredCommandInfoItems.length - 1;
    }
    selectedIndex = commandInfoItems.indexOf(filteredCommandInfoItems[selectedFilteredIndex]);
    scrollElements[selectedFilteredIndex].scrollIntoView({
      behavior: 'smooth',
      block: 'nearest',
      inline: 'start',
    });
    e.preventDefault();
  } else if (e.key === ENTER_KEY) {
    // hide the command palette
    hideCommandPallete();

    selectedIndex = commandInfoItems.indexOf(filteredCommandInfoItems[selectedFilteredIndex]);
    await executeCommand(selectedIndex);
    e.preventDefault();
  } else if (e.key === TAB_KEY) {
    switchSearchOption(e.shiftKey ? -1 : 1);
    e.preventDefault();
  }
}

function switchSearchOption(direction: 1 | -1): void {
  const searchOptionsLength = searchOptions.length;
  const offset = direction === 1 ? 0 : searchOptionsLength;
  searchOptionsSelectedIndex = (searchOptionsSelectedIndex + direction + offset) % searchOptionsLength;
}

async function executeCommand(index: number): Promise<void> {
  // get command id
  const commandId = commandInfoItems[index].id;
  // execute the command
  try {
    await window.executeCommand(commandId);
  } catch (error) {
    console.error('error executing command', error);
  }
}

function handleMousedown(e: MouseEvent): void {
  if (!display) {
    return;
  }

  if (outerDiv && !e.defaultPrevented && e.target instanceof Node && !outerDiv.contains(e.target)) {
    hideCommandPallete();
  }
}

function hideCommandPallete(): void {
  display = false;
  onclose?.();
}

async function clickOnItem(index: number): Promise<void> {
  // hide the command palette
  hideCommandPallete();

  // select the index from the cursor
  selectedIndex = commandInfoItems.indexOf(filteredCommandInfoItems[index]);
  await executeCommand(selectedIndex);
}

async function onInputChange(): Promise<void> {
  // in case of quick pick, filter the items
  selectedFilteredIndex = 0;
  if (filteredCommandInfoItems.length > 0) {
    selectedIndex = commandInfoItems.indexOf(filteredCommandInfoItems[selectedFilteredIndex]);
  }
}

async function onAction(): Promise<void> {
  // When input is cleared, refocus it
  tick()
    .then(() => {
      inputElement?.focus();
    })
    .catch((error: unknown) => {
      console.error('Unable to focus input box', error);
    });
}
</script>

<svelte:window on:keydown={handleKeydown} on:mousedown={handleMousedown} />

{#if display}
  <div class="fixed top-0 left-0 right-0 bottom-0 bg-[var(--pd-modal-fade)] opacity-60 h-full z-50" style='-webkit-app-region: none;'></div>

  <div class="absolute m-auto left-0 right-0 z-50">
    <div class="flex justify-center items-center mt-1">
      <div
        bind:this={outerDiv}
        class="bg-[var(--pd-content-card-bg)] w-[700px] max-h-fit shadow-lg p-2 rounded-sm shadow-[var(--pd-input-field-stroke)] text-base">
        <div class="w-full flex flex-row gap-2 items-center">
          <Input
            aria-label='Command palette command input'
            bind:value={inputValue}
            bind:element={inputElement}
            clearable={true}
            on:input={onInputChange}
            on:action={onAction}
            class="px-1 w-full text-[var(--pd-input-field-focused-text)] bg-[var(--pd-input-field-focused-bg)] border border-[var(--pd-input-field-stroke)] focus:outline-hidden" >
            {#snippet left()}
              <Icon icon={searchIcon} class="pr-1"/>
            {/snippet}
        </Input>
        </div>

        <div class="flex flex-row m-2">
          {#each searchOptions as searchOption, index (index)}
            <Button
              type="tab"
              class="focus:outline-hidden"
              on:click={(): void => {
                searchOptionsSelectedIndex = index;
              }}
              selected={searchOptionsSelectedIndex === index}>
              <div class="flex items-center gap-2">
                {#if searchOption.shortCut}
                  {#each searchOption.shortCut as shortCut, i (i)}
                    <div class='bg-[var(--pd-search-bar-nav-button)] rounded-sm px-0.5 shadow-sm shadow-b-1'>
                      {shortCut}
                    </div>
                  {/each}
                {/if}
                {searchOption.text}
              </div>
            </Button>
          {/each}
        </div>
        <ul class="max-h-[50vh] overflow-y-auto flex flex-col mt-1">
          {#each filteredCommandInfoItems as item, i (item.id)}
            <li class="flex w-full flex-row" bind:this={scrollElements[i]} aria-label={item.id}>
              <button
                onclick={(): Promise<void> => clickOnItem(i)}
                class="text-[var(--pd-dropdown-item-text)] text-left relative w-full rounded-sm {i === selectedFilteredIndex
                  ? 'bg-[var(--pd-modal-dropdown-highlight)] selected'
                  : 'hover:bg-[var(--pd-dropdown-bg)]'}  px-1">
                  <div class="flex flex-row w-full max-w-[700px] truncate">
                    <div class="text-base py-[2pt]">{item.title}</div>
                  </div>
                </div>
              </button>
            </li>
          {/each}
        </ul>

        {#if filteredCommandInfoItems.length === 0}
          <div class='flex grow items-center flex-col gap-2 py-4'>
            <Icon icon={$isDark ? NotFoundDarkIcon : NotFoundLightIcon} />
            <div class='text-lg font-bold'>No results matching '{inputValue}' found</div>
            <div class='text-md'>Not what you expected? Double-check your spelling or try searching for:</div>
            <Button icon={faChevronRight} type='link' onclick={(): void => {console.log('Variables clicked');}}>Variables</Button>
          </div>
        {/if}

        <div class="border-[var(--pd-global-nav-bg-border)] border-t-[1px] flex flex-row items-center px-3 pt-2 gap-4 text-sm">
          <span class="flex items-center gap-2 text-[var(--pd-button-tab-text)] border-[var(--pd-button-tab-border-selected)]">
            <Icon icon={EnterIcon} class="bg-[var(--pd-action-button-bg)] rounded-sm p-1.5" size='2.2em'/>
            To select
          </span>
          <div class="flex items-center gap-2 text-[var(--pd-button-tab-text)] border-[var(--pd-button-tab-border-selected)]">
            <Icon icon={ArrowUpIcon} class="bg-[var(--pd-action-button-bg)] rounded-sm p-1.5" size='2.2em'/>
            <Icon icon={ArrowDownIcon} class="bg-[var(--pd-action-button-bg)] rounded-sm p-1.5" size='2.2em'/>
            To navigate
          </div>
          <div class="flex items-center gap-2 text-[var(--pd-button-tab-text)] border-[var(--pd-button-tab-border-selected)]">
            <div class='bg-[var(--pd-action-button-bg)] rounded-sm text-base px-1 py-0.5'>esc</div>
            To close
          </div>
        </div>
      </div>
    </div>
  </div>
{/if}


```

packages/renderer/src/lib/dialogs/CommandPalette.svelte
benoitf: yes but a single NotFoundIcon could use the isDark property automatically no ?

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -74,23 +103,35 @@ $effect(() => {
...
 let selectedIndex = 0;
 
+function displaySearchBar(): void {
+  // clear the input value
+  inputValue = '';
+  selectedIndex = 0;
+  // toggle the display
+  display = true;
+}
+
```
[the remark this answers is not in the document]
simonrey1: + 1

packages/renderer/src/lib/dialogs/CommandPalette.svelte
gastoner: I think that if I use just $isDark, it will get only the current value. So If you try to change the preference, the Image will be same

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -138,9 +196,18 @@ async function handleKeydown(e: KeyboardEvent): Promise<void> {
     selectedIndex = commandInfoItems.indexOf(filteredCommandInfoItems[selectedFilteredIndex]);
     await executeCommand(selectedIndex);
     e.preventDefault();
+  } else if (e.key === TAB_KEY) {
+    switchSearchOption(e.shiftKey ? -1 : 1);
+    e.preventDefault();
   }
 }
 
+function switchSearchOption(direction: 1 | -1): void {
```
simonrey1, line 197: Good implementation 👍

packages/renderer/src/lib/dialogs/CommandPalette.svelte
gastoner: There was a problem with the return method. I think I could connect those 2 when I create a new unsubscriber for `commandsInfos`. But I didn't want to do this refactoring

packages/renderer/src/lib/dialogs/CommandPalette.svelte
```diff
@@ -237,6 +309,16 @@ async function onAction(): Promise<void> {
             </li>
           {/each}
         </ul>
+
+        {#if filteredCommandInfoItems.length === 0}
+          <div class='flex grow items-center flex-col gap-2 py-4'>
+            <Icon icon={faMagnifyingGlass} size="5x"/>
+            <div class='text-lg font-bold'>No results matching '{inputValue}' found</div>
+            <div class='text-md'>Not what you expected? Double-check your spelling or try searching for:</div>
+            <Button icon={faChevronRight} type='link' onclick={(): void => {console.log('Variables clicked');}}>Variables</Button>
```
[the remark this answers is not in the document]
simonrey1: + 1

simonrey1 (approved):
Tested on Mac some minor comments on code
But great work! Thanks for the improvement on the command palette. 
Can't wait to see those tabs implemented 👍

packages/renderer/src/lib/dialogs/CommandPalette.svelte
simonrey1: Ok I tried and indeed there was a problem with the return. But I am still confused how Svelte handles multiple onMount

packages/renderer/src/lib/dialogs/CommandPalette.svelte
simonrey1: Edit: when we switch light / dark with the menu open, the image stays the same yes with this code

simonrey1: But we have the same issue with your current implemntation, so I am still pushing to use $isDark directly

benoitf (commented):
as isDarkTheme is only used to select the proper icon

move isDarkTheme to a NotFoundIcon

and inside either load the svg for dark or light theme so it's transparent when consuming the icon

it could also be a separate PR so it's quickly merged

gastoner:
@eqqe https://github.com/podman-desktop/podman-desktop/pull/13661 here is an another PR that adds content into docs tab if you want to take a look

packages/renderer/src/lib/images/NotFoundIcon.svelte
```diff
@@ -0,0 +1,28 @@
...
+    isDarkTheme = value;
+  });
+});
+
+onDestroy(() => {
+  unsubscribe?.();
+});
+</script>
+
+{#if isDarkTheme}
```
benoitf: use $isDark and remove the subscribe/unsubscribe stuff to use the store value in reactive mode directly

benoitf (approved):
LGTM

Issue #13392, opened 2025-07-29
Migrate image components to Svelte 5


simonrey1:
### Bug description

Migrate components under `packages/renderer/src/lib/image` to Svelte 5

### Operating system

all

### Installation Method

None

### Version

1.19.2

### Relevant log output

```shell

```

PR #13706, opened 2025-08-26
fix(renderer): update LoadImages and RecommendedRegistry to svelte 5

simonrey1:
### What does this PR do?

Update LoadImages and RecommendedRegistry to svelte 5

Partial fix of #13392

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/image/LoadImages.svelte
+++ b/packages/renderer/src/lib/image/LoadImages.svelte
@@ -13,16 +13,15 @@ import type { ProviderContainerConnectionInfo, ProviderInfo } from '/@api/provid
 
 import EngineFormPage from '../ui/EngineFormPage.svelte';
 
-let archivesToLoad: string[] = [];
-let loadError: string = '';
+let archivesToLoad = $state<string[]>([]);
+let loadError = $state<string>('');
 
 let providers: ProviderInfo[] = [];
-let providerConnections: ProviderContainerConnectionInfo[] = [];
-let selectedProvider: ProviderContainerConnectionInfo | undefined = undefined;
-let inProgress = false;
+let providerConnections = $state<ProviderContainerConnectionInfo[]>([]);
+let selectedProvider = $state<ProviderContainerConnectionInfo>();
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/image/LoadImages.svelte
```diff
@@ -104,9 +103,9 @@ async function loadImages(): Promise<void> {
         <div class="flex flex-col grow pl-2">Image Archives</div>
       </div>
     {/if}
-    {#each archivesToLoad as archiveToLoad, index (index)}
+    {#each archivesToLoad as _, index (index)}
       <div class="flex flex-row justify-center w-full py-1">
-        <Input bind:value={archiveToLoad} aria-label="archive path" readonly={true} />
+        <Input bind:value={archivesToLoad[index]} aria-label="archive path" readonly={true} />
```
axel7083, line 108: this is weird, we should definitely not do that

simonrey1: Screenshot 2025-08-26 at 12 01 29

simonrey1: https://svelte.dev/docs/svelte/compiler-errors#each_item_invalid_assignment

axel7083: omg, this is so not nice, but okay

axel7083, line 108:
```suggestion
        <!-- using index is required (see https://svelte.dev/docs/svelte/compiler-errors#each_item_invalid_assignment) -->
        <Input bind:value={archivesToLoad[index]} aria-label="archive path" readonly={true} />
```

axel7083 (approved):
LGTM, but as this is pretty uncommon synthax, would be nice to have a comment

simonrey1:
> LGTM, but as this is pretty uncommon synthax, would be nice to have a comment

I found many occurrences of this while updating to Svelte 5, should I add this comment to each occurrence?

PR #13709, opened 2025-08-26
chore(test): update cli tools tests for new behavior

cbr7:
### What does this PR do?
Update cli tools e2e tests to check for the latest changes in behavior.

### DIFF
--- a/tests/playwright/src/model/pages/cli-tools-page.ts
+++ b/tests/playwright/src/model/pages/cli-tools-page.ts
@@ -89,17 +89,10 @@ export class CLIToolsPage extends SettingsPage {
     });
   }
 
-  public async installTool(toolName: string, version: string = '', timeout = 60_000): Promise<this> {
+  public async installTool(toolName: string, timeout = 60_000): Promise<this> {
     return test.step(`Install ${toolName}`, async () => {
       await playExpect(this.getInstallButton(toolName)).toBeEnabled();
       await this.getInstallButton(toolName).click();
-      await playExpect(this.dropDownDialog).toBeVisible();
-      if (!version) {
-        version = await this.getLatestVersionNumber();
-      }
-
-      await playExpect(this.getVersionSelectionButton(version)).toBeEnabled();
-      await this.getVersionSelectionButton(version).click();
 
       const confirmationDialog = this.page.getByRole('dialog', { name: toolName });
       try {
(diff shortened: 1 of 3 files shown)


tests/playwright/src/model/pages/cli-tools-page.ts
```diff
@@ -89,42 +89,10 @@ export class CLIToolsPage extends SettingsPage {
     });
   }
 
-  public async installTool(toolName: string, version: string = '', timeout = 60_000): Promise<this> {
+  public async installTool(toolName: string, timeout = 60_000): Promise<this> {
```
odockal, line 92: We are breaking contract here. In next iteration we will have option to install particular version back. I would be for keeping api intact and throw an error when version is used now in this function call.

tests/playwright/src/utility/operations.ts
```diff
@@ -296,20 +296,15 @@ export async function getVolumeNameForContainer(page: Page, containerName: strin
   });
 }
 
-export async function ensureCliInstalled(
-  page: Page,
-  resourceName: string,
-  timeout = 60_000,
-  version = '',
-): Promise<void> {
+export async function ensureCliInstalled(page: Page, resourceName: string, timeout = 60_000): Promise<void> {
```
odockal, line 299: change of api is not necessary. I would keep version argument and throw an error if version argument is passed when not empty.

odockal (changes requested):

tests/playwright/src/model/pages/cli-tools-page.ts
cbr7: I've changed all the places that would matter, which in reality was nowhere, there is no point in implementing something for some future change that might come six months from now, when the version selection will be implemented, we will updated accordingly, but certainly not now since it no longer applies.

tests/playwright/src/utility/operations.ts
cbr7: I'm sorry but no, keeping version as a param when it's not used is not only bad policy but in fact would trigger the linter rules, any param that is not used much be removed.

tests/playwright/src/model/pages/cli-tools-page.ts
odockal: You need to check if it was not used elsewhere already, minikube has usage that wont be affected, but other extensions might be.

odockal (approved):
LGTM.

Note: Check the usage in other extensions and fix if necessary.

PR #13717, opened 2025-08-26
fix: update pnpm for stability

simonrey1:
### What does this PR do?

Some PR are failing at the first pnpm step, let's try to update to see if it stabilises it.

NB: I have used the `pnpm self-update` command that removes the SHA. Let me know if you see an important reason to re add it.

### How to test this PR?

### DIFF
--- a/package.json
+++ b/package.json
@@ -236,5 +236,5 @@
       "@kubernetes/client-node": "patches/@kubernetes__client-node.patch"
     }
   },
-  "packageManager": "pnpm@10.6.2+sha512.47870716bea1572b53df34ad8647b42962bc790ce2bf4562ba0f643237d7302a3d6a8ecef9e4bdfc01d23af1969aa90485d4cebb0b9638fa5ef1daef656f6c1b"
+  "packageManager": "pnpm@10.15.0"
 }


cbr7 (approved):

benoitf:
The sha needs to be used. It is an issue to remove it
Please add it back

simonrey1:
> The sha needs to be used. It is an issue to remove it Please add it back

I can't find the reason online, and it seems to be working, why is it an issue?

benoitf:
to pinpoint immutable things. Tags can be overwritten

simonrey1:
> to pinpoint immutable things. Tags can be overwritten

I agree that using a SHA is the best way to ensure immutability since tags can be overwritten. However, the `pnpm self-update` command is the official and recommended way to manage this field, and it doesn't add a SHA.

jeffmaury:
> > to pinpoint immutable things. Tags can be overwritten
> 
> I agree that using a SHA is the best way to ensure immutability since tags can be overwritten. However, the `pnpm self-update` command is the official and recommended way to manage this field, and it doesn't add a SHA.

+1 on keeping the SHA. `corepack up` will do the trick correctly

PR #13718, opened 2025-08-26
feat(renderer): introduce TableSvelte5

### LINKED ISSUE #13892: feat(renderer): create TableSvelte5 component

axel7083:
### What does this PR do?

Part of https://github.com/podman-desktop/podman-desktop/issues/13889

### Notes

:information_source: We cannot bind to the `select` property, as children are given through a function in the RowTable object. Therefore, the parent should provide a `Set<string>` representing the selected items.

:information_source: To determine properly the _id_ / _key_ of an item, the parent should provide a `key(item: T): string` function, this is important for Svelte each mechanism. This function is also used to properly identify which items are selected or collapsed.

:information_source: The current table is trying to access the `name` property of the item to use as aria-label, this is annoying as not all item will have a name property, therefore adding a `label(item: T): string` function. The parent is responsible of determining the name of each item.

:warning: A simple migration of the existing table was not possible as the central mechanism was not compatible with Svelte 5, a partial rewrite was required.

Requires rebase after 
- https://github.com/podman-desktop/podman-desktop/pull/13893
- https://github.com/podman-desktop/podman-desktop/pull/14034
- https://github.com/podman-desktop/podman-desktop/pull/13903

Related to https://github.com/podman-desktop/podman-desktop/issues/11586

Fixes https://github.com/podman-desktop/podman-desktop/issues/13892

Required for:
- https://github.com/podman-desktop/podman-desktop/pull/14178
- https://github.com/podman-desktop/podman-desktop/pull/14185
- https://github.com/podman-desktop/podman-desktop/pull/14186
- https://github.com/podman-desktop/podman-desktop/pull/14187
- https://github.com/podman-desktop/podman-desktop/pull/14188
- https://github.com/podman-desktop/podman-desktop/pull/14189
- https://github.com/podman-desktop/podman-desktop/pull/14190

### How to test this PR?

### DIFF
--- a/packages/renderer/src/lib/table/TableSvelte5.spec.ts
+++ b/packages/renderer/src/lib/table/TableSvelte5.spec.ts
@@ -279,8 +279,44 @@ describe('collapsed', () => {
       const row = getByRole('row', { name: label(person) });
 
       // Ensure the button has aria-expanded="false"
-      const collapseBtn = within(row).getByRole('button');
+      const collapseBtn = within(row).getByRole('button', { name: 'Expand Row' });
       expect(collapseBtn).toHaveAttribute('aria-expanded', 'false');
+    },
+  );
+
+  test.each<Person>(PARENTS.filter(person => (ADVANCED_ROW.info.children?.(person) ?? []).length > 0))(
+    'collapsed and expand row $id should properly update collapsed prop',
+    async person => {
+      const collapsed = new SvelteSet<string>();
+      assert(collapsed.size === 0, 'collapsed set should be empty');
+
+      const { getByRole } = render(TableSvelte5<Person>, {
+        kind: 'persons',
+        data: PARENTS,
+        row: ADVANCED_ROW,
+        columns: COLUMNS,
+        key,
+        label,
+        // Specify collapsed set
+        collapsed: collapsed,
+      });
+
+      const row = getByRole('row', { name: label(person) });
+      const collapseBtn = within(row).getByRole('button', { name: 'Collapse Row' });
+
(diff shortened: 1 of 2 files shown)


benoitf (changes requested):
hello, I still don't get why we would need to introduce a new dedicated component for the support of svelte5

benoitf:
I would assume that the table is a migration of table to svelte5, not introducing other unrelated changes

> The current table is trying to access the name property of the item to use as aria-label, this is annoying as not all item will have a name property, therefore adding a label(item: T): string function. The parent is responsible of determining the name of each item.

not needed for svelte5 update

>  To determine properly the id / key of an item, the parent should provide a key(item: T): string function, this is important for Svelte each mechanism. This function is also used to properly identify which items are selected or collapsed.

not required

benoitf:
> ⚠️ A simple migration of the existing table was not possible as the central mechanism was not compatible with Svelte 5, a partial rewrite was required.

you decided to force the usage of label and keys but it's unrelated to svelte5

deboer-tim:
If we have to introduce a new component, there's a non-zero chance that by the time we've ported everything (extension use too?) and are ready to delete the 'old' Table component we won't feel like renaming all the new users. So maybe we should pick something less specific than 'TableSvelte5'?

For a similar reason I wonder if it's better if we just look at this as a new component: kicked off due to Svelte 5, and as minimal migration as possible, but using what we've learned to improve it at the same time? That might reduce the strict comparison and give a reason for it being 'PDTable' or 'ObjectTable' or 'AlexTable' 😆 instead.

Issue #12742, opened 2025-06-04
Change extension-installation-smoke E2E to use only podman-desktop extensions


danivilla9:
### Is your enhancement related to a problem? Please describe

The [extension-installation-smoke](https://github.com/podman-desktop/podman-desktop/blob/main/tests/playwright/src/specs/extension-installation-smoke.spec.ts) test file includes references to third party extensions.

### Describe the solution you'd like

Remove the third party extensions from the test case and add podman desktop ones.

odockal:
PR was incomplete and reverted, it needs to be redone.

PR #13752, opened 2025-08-29
chore(test): swap external extensions on extension-installation-smoke for internal ones

danivilla9:
### What does this PR do?
Solves #12742 
#12742 

### How to test this PR?


danivilla9:
There are two remaining internal extensions (kreate, kubernetes-dashboard) that have not been added to the test because of its smoke nature and the fact that they were not defined in the file, but I can define and add them if deemed necessary.

ScrewTSW (approved):

PR #13759, opened 2025-09-01
chore(test): add handling for different resource name from resource card

### LINKED ISSUE #13425: Fix the usage of label in e2e tests when clicking on Create New button on resources page
### Bug description

We have this code:
```
  public async goToCreateNewResourcePage(resourceLabel: string): Promise<void> {
    if (!(await this.resourceCardIsVisible(resourceLabel))) {
      throw new Error(`Resource card with label ${resourceLabel} is not available`);
    }

    await this.resourceCardLocatorGenerator(resourceLabel)
      .getByRole('button', { name: `Create new ${resourceLabel}` })
      .click();
  }
```
but it seems that it is not right, for example it does not work for Developer Sandbox card, that does not have `resourceLabel` value (`redhat.sandbox`) the same as suffix in `Create New xxx`, where `xxx` is `Developer Sandbox`. I am not sure if it ever worked. We need to fix it to accept two params.

### Operating system

any

### Installation Method

None

### Version

1.19.2

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

cbr7:
### What does this PR do?
Adds handling in test framework for resource card to differ from resource name.

https://github.com/podman-desktop/podman-desktop/issues/13425

### DIFF
--- a/tests/playwright/src/model/pages/resources-page.ts
+++ b/tests/playwright/src/model/pages/resources-page.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2023-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


danivilla9 (approved):
LGTM

odockal (approved):

PR #13705, opened 2025-08-26
[wip] feat: added configurable columns

gastoner:
### What does this PR do?
Adds configurable columns to tables

This PR combines 4 or 5 PR together (also implements changing the order for the dashboard)

### What issues does this PR fix or reference?
Closes #12186 

### How to test this PR?
Go to images/containers/pods/volumes - click on gear icon and modify your columns

- [x] Tests are covering the bug fix or the new feature

PR #13760, opened 2025-09-01
chore: fixed comparing default value of object config properties

gastoner:
### What does this PR do?
Fixes comparing configuration objects with default values

Required for https://github.com/podman-desktop/podman-desktop/pull/13705

### How to test this PR?
You can try https://github.com/podman-desktop/podman-desktop/pull/13705 -> change some columns -> should be saved in settings.json -> then restore to default -> should disappear form config === this works 

### DIFF
--- a/packages/main/src/plugin/configuration-registry.ts
+++ b/packages/main/src/plugin/configuration-registry.ts
@@ -18,6 +18,7 @@
 
 import * as fs from 'node:fs';
 import * as path from 'node:path';
+import { isDeepStrictEqual } from 'node:util';
 
 import type * as containerDesktopAPI from '@podman-desktop/api';
 import { inject, injectable } from 'inversify';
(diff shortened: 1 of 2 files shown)


packages/main/src/plugin/configuration-registry.ts
```diff
@@ -275,7 +275,7 @@ export class ConfigurationRegistry implements IConfigurationRegistry {
     const cloneConfig = { ...this.configurationValues.get(CONFIGURATION_DEFAULT_SCOPE) };
     // for each key being already the default value, remove the entry
     Object.keys(cloneConfig)
-      .filter(key => cloneConfig[key] === this.configurationProperties[key]?.default)
+      .filter(key => JSON.stringify(cloneConfig[key]) === JSON.stringify(this.configurationProperties[key]?.default))
```
benoitf: would it work using https://nodejs.org/api/util.html#utilisdeepstrictequalval1-val2 ?

gastoner: Yess looks like it works with this, updated

benoitf (approved):

PR #13768, opened 2025-09-02
feat: added support for layout manager in DashboardPage

### LINKED ISSUE #13773: Dashboard - Allow users to hide certain sections.

As a user I want hide certain sections on the dashboard (Learning center, etc etc). I also want to be able to re-enable them at a future date.

These settings should persist. When I restart the app they are still hidden.

[Proposed UX here ](https://github.com/podman-desktop/podman-desktop/issues/12637#issuecomment-3223477360)

Image

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds support for layout manager 
Implementation should be +- same as in Table component, but with this we need to integrate new dashboard-reistry 

### Screenshot / video of UI

[Screencast_20251008_153012.webm](https://github.com/user-attachments/assets/1f0c916f-6599-4fd5-94c0-ffe00d69a4b1)

Closes https://github.com/podman-desktop/podman-desktop/issues/13773

Needs rebase after
- [x] https://github.com/podman-desktop/podman-desktop/pull/13771 
- [x] https://github.com/podman-desktop/podman-desktop/pull/13772
- [x] https://github.com/podman-desktop/podman-desktop/pull/13770

### How to test this PR?
Do to dashboard and try to reshuffle with sections / hide/ show them etc (should be persistent)

### DIFF
--- a/packages/renderer/src/lib/dashboard/DashboardPage.svelte
+++ b/packages/renderer/src/lib/dashboard/DashboardPage.svelte
@@ -1,18 +1,213 @@
 <script lang="ts">
-import { NavPage } from '@podman-desktop/ui-svelte';
+import { ListOrganizer, type ListOrganizerItem, NavPage, tablePersistence } from '@podman-desktop/ui-svelte';
+import { onMount } from 'svelte';
+import { SvelteMap } from 'svelte/reactivity';
 
-import { dashboardPageRegistry } from '/@/stores/dashboard/dashboard-page-registry.svelte';
-
+import {
+  convertFromListOrganizerItems,
+  dashboardPageRegistry,
+  type DashboardPageRegistryEntry,
+  defaultSectionNames,
+  setupDashboardPageRegistry,
+} from '../../stores/dashboard/dashboard-page-registry.svelte';
 import NotificationsBox from './NotificationsBox.svelte';
+
+// Dashboard section configuration managed by dashboard page registry
+let dashboardSections = $state<ListOrganizerItem[]>([]);
+// eslint-disable-next-line svelte/no-unnecessary-state-wrap
+let dashboardOrdering = $state(new SvelteMap<string, number>());
+let isInitialized = false;
+let isLoading = false;
+
+let orderedSections = $derived(getOrderedDashboardSections());
+let updatedEntries = $derived(convertFromListOrganizerItems(orderedSections, dashboardPageRegistry.entries));
+
+// Initialize default dashboard configuration
+function getDefaultDashboardItems(): ListOrganizerItem[] {
+  return dashboardPageRegistry.entries.map(entry => ({
+    id: entry.id,
+    label: entry.id,
+    enabled: true,
+    originalOrder: entry.originalOrder,
+  }));
+}
+
+// Initialize dashboard configuration
+async function initializeDashboard(): Promise<void> {
+  if (isInitialized || isLoading) return;
+
+  isLoading = true;
+  try {
+    if (dashboardPageRegistry.entries.length > 0) {
+      const loadedItems = await loadDashboardConfiguration();
+      dashboardSections = loadedItems;
+    }
+    isInitialized = true;
+  } catch (error: unknown) {
+    console.error('Failed to load dashboard configuration:', error);
+    // Fallback to default configuration
+    dashboardSections = getDefaultDashboardItems();
+    isInitialized = true;
+  } finally {
+    isLoading = false;
+  }
+}
+
+// Load configuration from settings (like Table.svelte does)
+async function loadDashboardConfiguration(): Promise<ListOrganizerItem[]> {
+  if (tablePersistence.storage) {
+    const loadedItems = await tablePersistence.storage.load('dashboard', defaultSectionNames);
+
+    if (loadedItems.length > 0) {
+      // Ensure loaded items have proper originalOrder from defaults if missing
+      const defaultItems = getDefaultDashboardItems();
+      const items = loadedItems.map((item: ListOrganizerItem) => ({
+        ...item,
+        originalOrder: item.originalOrder ?? defaultItems.find(d => d.id === item.id)?.originalOrder ?? 0,
+      }));
+
+      // Build ordering map from loaded items
+      // Check if items are in a different order than their original order
+      const isReordered = items.some((item: ListOrganizerItem, index: number) => item.originalOrder !== index);
+      if (isReordered) {
+        const ordering = new SvelteMap<string, number>();
+        items.forEach((item: ListOrganizerItem, index: number) => {
+          ordering.set(item.id, index);
+        });
+        dashboardOrdering = ordering;
+      } else {
+        dashboardOrdering.clear();
+      }
+
+      return items;
+    }
+  }
+  return getDefaultDashboardItems();
+}
+
+// Get ordered dashboard sections based on current ordering
+function getOrderedDashboardSections(): ListOrganizerItem[] {
+  if (dashboardOrdering.size === 0) {
+    return dashboardSections;
+  }
+  return [...dashboardSections].toSorted((a, b) => {
+    const aOrder = dashboardOrdering.get(a.id) ?? a.originalOrder;
+    const bOrder = dashboardOrdering.get(b.id) ?? b.originalOrder;
+    return aOrder - bOrder;
+  });
+}
+
+// Save configuration (like Table.svelte does)
+async function saveDashboardConfiguration(): Promise<void> {
+  if (tablePersistence.storage) {
+    const orderedItems = getOrderedDashboardSections();
+    const serializableItems = orderedItems.map(item => ({
+      id: item.id,
+      label: item.label,
+      enabled: item.enabled,
+      originalOrder: item.originalOrder,
+    }));
+
+    await tablePersistence.storage.save('dashboard', serializableItems);
+  }
+}
+
+// Initialize dashboard on mount
+onMount(async () => {
+  await initializeDashboard();
+});
+
+// Save configuration whenever dashboardSections or ordering changes
+$effect(() => {
+  if (isInitialized && dashboardSections.length > 0) {
+    dashboardPageRegistry.entries = updatedEntries;
+  }
+});
+
+// Reset function for dashboard layout
+async function resetDashboardLayout(): Promise<void> {
+  if (tablePersistence.storage) {
+    try {
+      // Reset using the persistence callbacks (clears saved config)
+      dashboardSections = await tablePersistence.storage.reset('dashboard', defaultSectionNames);
+      dashboardOrdering.clear();
+
+      // Reset the registry to default state
+      setupDashboardPageRegistry();
+
+      // Ensure the registry reflects the reset state
+      if (dashboardPageRegistry.entries.length > 0) {
+        const updatedEntries = convertFromListOrganizerItems(dashboardSections, dashboardPageRegistry.entries);
+        dashboardPageRegistry.entries = updatedEntries;
+      }
+    } catch (error: unknown) {
+      console.error('Failed to reset dashboard layout:', error);
+    }
+  }
+}
+
+// Handle dashboard order changes from ListOrganizer
+function handleDashboardOrderChange(newOrdering: SvelteMap<string, number>): void {
+  dashboardOrdering = new SvelteMap(newOrdering);
+
+  // Save configuration immediately when order changes
+  saveDashboardConfiguration().catch((error: unknown) => {
+    console.error('Failed to save dashboard configuration after order change:', error);
+  });
+}
+
+// Handle dashboard section toggle changes from LayoutEditor
+function handleDashboardToggle(itemId: string, enabled: boolean): void {
+  dashboardSections = dashboardSections.map(item => (item.id === itemId ? { ...item, enabled } : item));
+
+  // Save configuration immediately when toggle changes
+  saveDashboardConfiguration().catch((error: unknown) => {
+    console.error('Failed to save dashboard configuration after toggle change:', error);
+  });
+}
+
+// Filter and sort dashboard registry items based on LayoutEditor configuration
+let sortedDashboardRegistry = $derived.by(() => {
+  // Get ordered sections inline to ensure reactivity
+  const orderedSections =
+    dashboardOrdering.size === 0
+      ? [...dashboardSections]
+      : [...dashboardSections].toSorted((a, b) => {
+          const aOrder = dashboardOrdering.get(a.id) ?? a.originalOrder;
+          const bOrder = dashboardOrdering.get(b.id) ?? b.originalOrder;
+          return aOrder - bOrder;
+        });
+
+  const result = orderedSections
+    .filter(section => section.enabled)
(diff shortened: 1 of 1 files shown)


jiridostal (changes requested):
To me it seems Release Notes have different handling? It does not accurately reflect the position set in the dropdown.
Steps to reproduce:

1. Swap first to items
2. Release notes do not move

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -20,6 +21,39 @@ import ReleaseNotesBox from './ReleaseNotesBox.svelte';
 
 const providerInitContexts = new SvelteMap<string, InitializationContext>();
 
+// Dashboard section configuration managed by backend
+let dashboardSections: LayoutEditItem[] = [];
+let isInitialized = false;
+
+// Available dashboard sections
+const availableSections = ['release-notes', 'extension-banners', 'learning-center', 'providers'];
```
benoitf: hello, I don't see why we would hardcode sections.

I would guess that sections are registered by themselves. So adding a section would register automatically the item.

Here we're kind of duplicating the section names.

jiridostal:
It works as expected, I did not see any bug or inconsistent behavior, so from my POV it's ready.

One thing that should be addressed soon is the handling of release notes, as now we have two different places where we store information about showing release notes. It would be great to manage showing/hiding it from this component and drop the other place (from release info dialog) so you don't run into following situation.

https://github.com/user-attachments/assets/ec263ee4-b97e-4d18-9c1f-6c1156799156

jiridostal (approved):

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,213 @@
...
+import { onMount } from 'svelte';
+import { SvelteMap } from 'svelte/reactivity';
 
+import {
+  convertFromListOrganizerItems,
+  dashboardPageRegistry,
+  type DashboardPageRegistryEntry,
+  defaultSectionNames,
+  setupDashboardPageRegistry,
+} from '../../stores/dashboard/dashboard-page-registry.svelte';
```
simonrey1: suggestion(non-blocking): you can use the path alias here

```ts

import {
  convertFromListOrganizerItems,
  dashboardPageRegistry,
  type DashboardPageRegistryEntry,
  defaultSectionNames,
  setupDashboardPageRegistry,
} from '/@/stores/dashboard/dashboard-page-registry.svelte';

```

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,213 @@
...
+
+  // Save configuration immediately when toggle changes
+  saveDashboardConfiguration().catch((error: unknown) => {
+    console.error('Failed to save dashboard configuration after toggle change:', error);
+  });
+}
+
+// Filter and sort dashboard registry items based on LayoutEditor configuration
+let sortedDashboardRegistry = $derived.by(() => {
+  // Get ordered sections inline to ensure reactivity
```
simonrey1: question: why is this needed to "Get ordered sections inline to ensure reactivity"?

I tried to remove this 

```ts

 const orderedSections =
    dashboardOrdering.size === 0
      ? [...dashboardSections]
      : [...dashboardSections].toSorted((a, b) => {
          const aOrder = dashboardOrdering.get(a.id) ?? a.originalOrder;
          const bOrder = dashboardOrdering.get(b.id) ?? b.originalOrder;
          return aOrder - bOrder;
        });

```

and it seems to be reactive when testing manually: I can still hide, show, move and reset layout.

gastoner: Removed

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Updated

packages/renderer/src/lib/dashboard/DashboardPage.svelte
simonrey1: The comment could be removed as well :)

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,205 @@
...
+  dashboardSections = dashboardSections.map(item => (item.id === itemId ? { ...item, enabled } : item));
+
+  // Save configuration immediately when toggle changes
+  saveDashboardConfiguration().catch((error: unknown) => {
+    console.error('Failed to save dashboard configuration after toggle change:', error);
+  });
+}
+
+// Filter and sort dashboard registry items based on LayoutEditor configuration
+let sortedDashboardRegistry = $derived.by(() => {
```
simonrey1: nitpick: you can use $derived instead of $derived.by here

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,205 @@
...
+}
+
+// Initialize dashboard on mount
+onMount(async () => {
+  await initializeDashboard();
+});
+
+// Save configuration whenever dashboardSections or ordering changes
+$effect(() => {
+  if (isInitialized && dashboardSections.length > 0) {
```
simonrey1: issue: when adding a log above this if, it feels like we are going in this effect a lot

this could hurt performance

I will try to investigate why.

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,205 @@
...
+  defaultSectionNames,
+  setupDashboardPageRegistry,
+} from '/@/stores/dashboard/dashboard-page-registry.svelte';
 
 import NotificationsBox from './NotificationsBox.svelte';
+
+// Dashboard section configuration managed by dashboard page registry
+let dashboardSections = $state<ListOrganizerItem[]>([]);
+// eslint-disable-next-line svelte/no-unnecessary-state-wrap
+let dashboardOrdering = $state(new SvelteMap<string, number>());
```
axel7083:
```suggestion
let dashboardOrdering = new SvelteMap<string, number>();
```

question: do we really need to wrap it in a state? I don't think so 🤔

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,205 @@
...
+let isInitialized = false;
+let isLoading = false;
```
axel7083: question: not states?

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,205 @@
...
+          ordering.set(item.id, index);
+        });
+        dashboardOrdering = ordering;
+      } else {
+        dashboardOrdering.clear();
+      }
+
+      return items;
+    }
+  }
```
axel7083: nitpick(non-blocking): prefer early return instead of large if block 


```ts
  if (!tablePersistence.storage)  return;
  
  [...]
  }
```

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,205 @@
...
+  return getDefaultDashboardItems();
+}
+
+// Get ordered dashboard sections based on current ordering
+function getOrderedDashboardSections(): ListOrganizerItem[] {
+  if (dashboardOrdering.size === 0) {
+    return dashboardSections;
+  }
+  return [...dashboardSections].toSorted((a, b) => {
```
axel7083: question(blocking): do you need to spread ? since you are using the `toSorted` you are doing the copy twice, once with `[...dashboardSections]` and once with toSorted

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,205 @@
...
+// Initialize dashboard on mount
+onMount(async () => {
+  await initializeDashboard();
+});
+
+// Save configuration whenever dashboardSections or ordering changes
+$effect(() => {
+  if (isInitialized && dashboardSections.length > 0) {
+    dashboardPageRegistry.entries = updatedEntries;
+  }
```
axel7083: suggestion: we may need to find a way to get the callback when the configuration change instead of relying on effect ?

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,205 @@
...
+});
+
+// Reset function for dashboard layout
+async function resetDashboardLayout(): Promise<void> {
+  if (tablePersistence.storage) {
```
axel7083: suggestion(non-blocking): prefer early return

```ts
if (!tablePersistence.storage) return;

[...]
```

axel7083 (commented):
Tested on Windows, works as expected !

Left some comments

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Done

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: I have removed effect

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: I have removed effect

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Removed spread

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Well without it when you reshuffle it wont update :(

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: With state it wont load for some reason haha ???

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,195 @@
...
+  
   {#snippet content()}
   <div class="flex flex-col min-w-full h-full bg-[var(--pd-content-bg)] py-5">
     <div class="min-w-full flex-1">
       <NotificationsBox />
       <div class="px-5 space-y-5 h-full">
-        {#each dashboardPageRegistry.entries as dashboardRegistryItem (dashboardRegistryItem.id)}
+        {#each sortedDashboardRegistry as dashboardRegistryItem (dashboardRegistryItem.id)}
           <dashboardRegistryItem.component />
         {/each}
```
axel7083, line 185: suggestion(blocking): we should use the right synthax, we should use uppercase;


```suggestion
        {#each sortedDashboardRegistry as { id, component: Composant} (dashboardRegistryItem.id)}
          {@const Component = dashboardRegistryItem.component}
          <Component />
        {/each}
```

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,195 @@
...
+  defaultSectionNames,
+  setupDashboardPageRegistry,
+} from '/@/stores/dashboard/dashboard-page-registry.svelte';
 
 import NotificationsBox from './NotificationsBox.svelte';
+
+// Dashboard section configuration managed by dashboard page registry
+let dashboardSections = $state<ListOrganizerItem[]>([]);
+// eslint-disable-next-line svelte/no-unnecessary-state-wrap
+let dashboardOrdering = $state(new SvelteMap<string, number>());
```
axel7083: The SvelteMap does not provide reactivity on the assignment, since in other lines we re-assign the variable we lose the reactivity (that's how wrapping in a state is working) but this is not the right way

```suggestion
const dashboardOrdering = new SvelteMap<string, number>();
```

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,195 @@
...
+let isInitialized = false;
+let isLoading = false;
```
axel7083: suggestion: we may need to group them and have a single one `loading: boolean`

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,195 @@
...
+
+    // Build ordering map from loaded items
+    // Check if items are in a different order than their original order
+    const isReordered = items.some((item: ListOrganizerItem, index: number) => item.originalOrder !== index);
+    if (isReordered) {
+      const ordering = new SvelteMap<string, number>();
+      items.forEach((item: ListOrganizerItem, index: number) => {
+        ordering.set(item.id, index);
+      });
+      dashboardOrdering = ordering;
```
axel7083: As per https://github.com/podman-desktop/podman-desktop/pull/13768/files#r2478242711 we should not re-assign




```suggestion
      items.forEach((item: ListOrganizerItem, index: number) => {
        dashboardOrdering.set(item.id, index);
      });
```

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -1,18 +1,195 @@
...
+      dashboardPageRegistry.entries = convertFromListOrganizerItems(dashboardSections, dashboardPageRegistry.entries);
+    }
+  } catch (error: unknown) {
+    console.error(`Failed to reset dashboard layout: ${error}`);
+  }
+}
+
+// Handle dashboard order changes from ListOrganizer
+function handleDashboardOrderChange(newOrdering: SvelteMap<string, number>): void {
+  dashboardOrdering = new SvelteMap(newOrdering);
```
axel7083:
```suggestion
  dashboardOrdering.clear();
  newOrdering.entries().forEach(([id, order]) => {
    dashboardOrdering.set(id, order);
  });
```

gastoner:
@axel7083 updated

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Should be fixed

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Should be fixed

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Removed both of them

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Should be fixed

axel7083 (approved):
LGTM ! Tested on windows, and work like a charm 🚀

PR #13769, opened 2025-09-02
chore: added support for layout manager component

gastoner:
### What does this PR do?
Adds support for newly added layout manager into table component

### What issues does this PR fix or reference?
Closes https://github.com/podman-desktop/podman-desktop/issues/12186

Needs rebase after
- [x] https://github.com/podman-desktop/podman-desktop/pull/13771 
- [x] https://github.com/podman-desktop/podman-desktop/pull/13772

### How to test this PR?
Whole feature can be tried out in https://github.com/podman-desktop/podman-desktop/pull/13705

- [x] Tests are covering the bug fix or the new feature

PR #13770, opened 2025-09-02
feat: added callbacks to table component for layout manager

gastoner:
### What does this PR do?
Adds callbacks to container/image/pod/volume tables to enable layout manager

### Screenshot / video of UI

[Screencast_20251008_093936.webm](https://github.com/user-attachments/assets/d8a96363-83a1-4d84-a02b-98ae68fadf38)

Needs rebase after
- [x] https://github.com/podman-desktop/podman-desktop/pull/13771 
- [x] https://github.com/podman-desktop/podman-desktop/pull/13772
- [x] https://github.com/podman-desktop/podman-desktop/pull/13769

### How to test this PR?
Try to edit table columns


packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -49,6 +50,19 @@ interface Props {
...
+  const availableColumns = columns.map(col => col.title);
+  return (await window.loadTableConfig('container', availableColumns)) ?? [];
+}
+async function saveContainerLayout(items: LayoutEditItem[]): Promise<void> {
+  await window.saveTableConfig('container', items);
+}
+async function resetContainerLayout(): Promise<LayoutEditItem[]> {
+  const availableColumns = columns.map(col => col.title);
+  return (await window.resetTableConfig('container', availableColumns)) ?? [];
+}
```
benoitf: AFAIK it's not scaling well.

why would we need to call these methods there ?

it should be handled by the table component, we should give the name of the table and say we want persistence or layout option

benoitf (changes requested):
page should not handle the persistence of the table, it's the custom table that should do it

packages/renderer/src/lib/container/ContainerList.svelte
gastoner: @benoitf I think we cant call window* from UI lib (svelte:check is yalling at me :D ) so AFAIK this is the only option? how to call window methods from UI lib?

benoitf: @gastoner you're absolutely correct and what we did in the past is to have a 'Podman Desktop' component handling the window stuff reusing an agnostic UI component

so for example Table in Podman Desktop has a field that is the name of the table and do all the persistence stuff


meaning that ContainerList, VolumeList, WhateverList in Podman Desktop shouldn't bother to call any window.* methods, it'll be done by the Podman Desktop component itself

packages/renderer/src/lib/container/ContainerList.svelte
```diff
@@ -393,7 +398,7 @@ function key(item: ContainerGroupInfoUI | ContainerInfoUI): string {
       <div class="inline-flex space-x-2">
         <Button
           on:click={(): Promise<void> =>
-           runSelectedContainers()}
+            runSelectedContainers()}
```
MarsKubeX: Hi @gastoner! is this formatting  correct?

gastoner:
@benoitf @MarsKubeX @dgolovin can you PTAL?

MarsKubeX (approved):
LGTM codewise. Just a question, is it necessary to name the component PodmanDesktopStoragePersist instead of only StoragePersist?

gastoner:
@MarsKubeX I've added the video, you should be able to edit the tables in all podman resources now (should be persisted).

MarsKubeX:
> @MarsKubeX I've added the video, you should be able to edit the tables in all podman resources now (should be persisted).

@gastoner I've just tested it and it's working good :). Nice job

gastoner:
@benoitf can you PTAL

packages/renderer/src/lib/dashboard/DashboardPage.svelte
```diff
@@ -56,51 +56,49 @@ function getInitializationContext(id: string): InitializationContext {
             <ProviderReady provider={providerReady} />
```
axel7083, line 56: I don't understand those changes in this PR?

axel7083 (approved):
Hey, following our discussion I approve this PR;

However, I have a strong reserve on the implementation, the part I don't really agree on is already merged and has been approved by two reviewers https://github.com/podman-desktop/podman-desktop/pull/13769 so this is why I don't request further change, but yet here is my two cent.

The new property `enableLayoutConfiguration` is having a flaw, if the user of the library did not initialized with the following code

```
import { tablePersistence } from '@podman-desktop/ui-svelte';

// Initialize table persistence callbacks immediately
tablePersistence.storage = new PodmanDesktopStoragePersist();
```

In a global place, the property `enableLayoutConfiguration` will have no effect; this is not an uncommon pattern to have code in the global scope, but here this is could lead to some problem when using the Table.

Let's say an extension want to use the table, and add `enableLayoutConfiguration` to have the configuration column, this will have no effect, leading to confusion.

I would suggest instead to have the storage as property of the table, so you are sure when passing down the storage property that it will work.

The code would look like to something like the following

```diff
<Table 
-  enableLayoutConfiguration={true}
+  layoutManager={storage}
/>
```

The storage could then be a global a shared state in the application without needed to be initialized at the global scope.

Moreover, this would allow to remove the tablePersistence exported state, that feel a bit odd to me.

cc @deboer-tim as he worked on the table

packages/renderer/src/lib/dashboard/DashboardPage.svelte
gastoner: Need to remove those changes :D

benoitf (approved):

